Issue #55033, opened 2023-07-15
Type hole: A function with const generic does not infer the type if it is a spread parameter (tuple)


Refzlund:
# Bug Report

### 🔎 Search Terms

`is:issue is:open function generic tuple const`
`is:issue is:open in:title tuple const`

Might solve; https://github.com/microsoft/TypeScript/issues/54254 `<T extends [a: string, b: number]>(...args: T)`

### 🕗 Version & Regression Information

- This is a overlooked type feature
- This has been an immediate effect of including a new feature: `const` in function generics
- This is the behaviour in every version I tried, and I reviewed the FAQ for entries about `const`

### ⏯ Playground Link

[...]

Issue #53398, opened 2023-03-20
Inferred ...args should not be readonly


jakebailey:
# Bug Report

### 🔎 Search Terms

readonly parameter inferred tuple array args variadic

### 🕗 Version & Regression Information

- This is the behavior in every version I tried
### ⏯ Playground Link

[Playground Link](https://www.staging-typescript.org/play?ts=5.0.2#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXzCgggDFUAeAFXhAA8MRVgBneOKYPCAT3ilW4BtALoA+ABSwA5swBc8SgBokqeeIB0m6XIUBKeAF5R8AG44swXfLMWA3AFgAUE9CRYCMHmYZ4WVAAdkDHl2TlQeeG8YPykRB2dHQmIycT9AjGUNLRgZfSN4AG8neBK+HOZ40vgAemr4AD0AficAX10nDsdXaDgkNExcfCTSVAA5PABxRhBosElykJAOLl4omJFlRFV4LPVtReXw1Yxo1FjhPOMbS2tzYHinYbJx1CnUGaw5tKDMzT3yy6FYqlbSVUq1BrNRxtJxAA)

### 💻 Code

```ts
declare function callFn<T extends readonly any[]>(args: T, fn: (...args: T) => void): void;

declare const input: readonly string[];

callFn(input, (...args) => {
    args; // readonly string[] ?
})

declare function callFnNonGeneric(args: readonly string[], fn: (...args: readonly string[]) => void): void;

callFnNonGeneric(input, (...args) => {
    args; // readonly string[] ?
})
```

### 🙁 Actual behavior

`...args` is readonly.

### 🙂 Expected behavior

`...args` is not readonly.

When this function is called, the input will be a brand new array which can be modified without affecting the caller:

```
> const fn = (...args) => { args[0] = "oops" }
undefined
> const arr = ["some", "values"]
undefined
> fn(...arr)
undefined
> arr
[ 'some', 'values' ]
```

Split out of https://github.com/microsoft/TypeScript/pull/53258#issuecomment-1470970487

The "fix" here is to strip readonly from variadic args (at the top level). This also means the fix in #53258 can be reverted (as the real fix can be more general).

fatcerberus:
Hmm, but `args` is typed as `T` in this example, which must necessarily be instantiated with a readonly array type, and `args` is required required to be the same type as `input`… so how would this work?

jakebailey:
[...]

PR #53258, opened 2023-03-14
Ensure rest type for source parameter is readonly in relations

jakebailey:
Fixes #53255
See also https://github.com/DefinitelyTyped/DefinitelyTyped/pull/64739

Given:

```ts
declare function each<T extends ReadonlyArray<any>>(cases: ReadonlyArray<T>): (fn: (...args: T) => any) => void;

const cases = [
    [1, '1'],
    [2, '2'],
] as const;

const eacher = each(cases);

// Error.
eacher((a, b) => {
});

// No error?
eacher((...args) => {
    const [a, b] = args;
});
```

We used to say:

```
Argument of type '(a: 1 | 2, b: "1" | "2") => void' is not assignable to parameter of type '(...args: readonly [1, "1"] | readonly [2, "2"]) => any'.
  Types of parameters 'a' and 'args' are incompatible.
    Type 'readonly [1, "1"] | readonly [2, "2"]' is not assignable to type '[a: 1 | 2, b: "1" | "2"]'.
      The type 'readonly [1, "1"]' is 'readonly' and cannot be assigned to the mutable type '[a: 1 | 2, b: "1" | "2"]'.(2345)
```

Here, the "source" type is `[a: 1 | 2, b: "1" | "2"]`, but that can't be assigned to the target as the target is readonly. But, these are the rest arguments, so there's no way for this function to observe the read-only ness of the parameter. (Note: these are parameters, so directionality is swapped.)

I'm not sure if this is a perfect fix, but one way to "solve" this is to just ensure the tuple/array we cook up for the source is readonly, that way it's always compatible. This fixes the issue, but does change some error messages because they _do_ observe the intermediary type here.

Maybe the "better" thing to do here would be to strip readonly-ness from both `sourceType` and `targetType` if they were produced via `getRestTypeAtPosition`? I haven't exactly found "the way" to do that yet, e.g. no helper named "remove readonly from tuple or array".

Issue #53255, opened 2023-03-14
Parameters created from tuple are treated as mutable instead of immutable, unless written as "...args"


jakebailey:
# Bug Report

### 🔎 Search Terms

const array tuple parameter spread

### 🕗 Version & Regression Information

- This is the behavior in every version I tried,.

### ⏯ Playground Link

[Playground Link](https://www.typescriptlang.org/play?ts=4.9.5#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXxCjAAsAeAFQIA8MRVgBneAJUODwgE8BBGGKTqSipOAPlEAKMFAYgGALhZsOPPgIqiAlIomJUOgHRHYAcwXxym+AF5R8YZyu34ANxxZgAbgBQ3sHgYMeGlZJmt4AG1veBjIgEYAGngAcjjkgF0E6NiIgCYk5NyMrPT7Jn9UQJ8-AKDCEhAYGwIiYikZOU0fAHpu2NiAPQB+X174AFE+HBh5bzHeE2QAWzognER4DE4ABwRkiShFOPgAH3h8+AAjRQAiOJvT+Bvcm6c7Nw9k+CwmVBwgmQMLAmVBQS4QBAYHDwbawKArWhNdabHZ7CRGAymcxwKDsVBceJJO43UpnHF4gl5IkvdJvewiZIGOZ9CyopjI2H8BGNJjJKBfYTAFJYgVwb7oHBLWHYcEgJljWLkVEpckqQlPe6k+Cq-GcSIXZ4kr4-eB-AEMIEgsEQzbQra7FIRQ7wY5nC7XDUPM6G9KM5n9GLkYiQ5XJHWUxKe33fXnhzgC+jBYRmq4IQHA1AgIVQzbB+BLZAYa0hh3JJ1HR7u273R4+xkSXIAZgALABWTSjPoAdWIetgkLzpmWqyYsjoZXzheLI289WDMAkBySlzpAG9sjEoD0WcMN1dtzFdwBfLq+WetRqLjFYtd7hX9Xf9CqBSJQZelcJYnz9Lf++CP2JLgPf8RhPHwgA)

### 💻 Code

```ts
declare function each<T extends ReadonlyArray<any>>(cases: ReadonlyArray<T>): (fn: (...args: T) => any) => void;

const cases = [
    [1, '1'],
    [2, '2'],
] as const;

const eacher = each(cases);

// Error.
eacher((a, b) => {
});

// No error?
[...]

PR #55034, opened 2023-07-15
Fixed inferred const rest types in signatures

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/55033

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -23925,7 +23925,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             callback(getTypeAtPosition(source, i), getTypeAtPosition(target, i));
         }
         if (targetRestType) {
-            callback(getRestTypeAtPosition(source, paramCount), targetRestType);
+            callback(getRestTypeAtPosition(source, paramCount, /*readonly*/ isConstTypeVariable(targetRestType)), targetRestType);
         }
     }
 
@@ -34928,7 +34928,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         return undefined;
     }
 
-    function getRestTypeAtPosition(source: Signature, pos: number): Type {
+    function getRestTypeAtPosition(source: Signature, pos: number, readonly?: boolean): Type {
         const parameterCount = getParameterCount(source);
         const minArgumentCount = getMinArgumentCount(source);
         const restType = getEffectiveRestType(source);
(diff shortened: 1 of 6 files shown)


ssalbdivad:
Pretty hyped for this one. I've run into so many issues trying to constraint narrowed array types without breaking inference. Couldn't really tell from the linked issue if this will fix a lot of them, but I have faith 🙏

jakebailey:
After messing with this same code for #53255, #53258 (note a similar `getRestTypeAtPosition` change), #53398, I'm like 50% afraid of this 😄 

@typescript-bot test this
@typescript-bot test top100
@typescript-bot run dt
@typescript-bot pack this

Andarist:
> After messing with this same code for https://github.com/microsoft/TypeScript/issues/53255, https://github.com/microsoft/TypeScript/pull/53258 (note a similar getRestTypeAtPosition change), https://github.com/microsoft/TypeScript/issues/53398, I'm like 50% afraid of this 😄

When learning about this issue I immediately opened your PR to check out if it's related anyhow to this issue. I think that at the end of the day, it isn't and that this PR shouldn't actually impact anybody negatively. This only makes those rests readonly when the type variable is `const` - given that `const` is pretty new I don't expect any accidental breakage here.

sandersn:
The bug is marked Working As Intended now, so should we close this PR?

Andarist:
The reported issue had some issues in it (it didnt use a readonly constraint). So im not sure if „working as intended” was applied to only what was oroginally reported or if it was applied to everything that was discussed there. I think that this particular thing that I fixed here still has some merit so i’d appreciate if this would get reviewed

ahejlsberg:
I like where this is going, but I think we also need the changes in #55229. For example, the following is broken by this PR without the additional changes:

```ts
declare function invoke<const T extends unknown[]>(f: (...args: T) => void, ...args: T): T;

invoke((a: string, b: number) => {}, "hello", 123);
```

I know this isn't a particularly likely break, but the inference failure because of non-readonly constraints is just confounding to most users and hard to justify (even if I've tried).

Andarist:
> I know this isn't a particularly likely break, but the inference failure because of non-readonly constraints is just confounding to most users and hard to justify (even if I've tried).

ye, this tends to be pretty confusing - would it be worth adding some diagnostics that would prompt users to add the `reaodnly` modifier to their constraints? Might not be feasible when constraints are defined using aliases but catching common cases when they are declared directly in the signature could still help a lot 🤔

ahejlsberg:
> would it be worth adding some diagnostics that would prompt users to add the reaodnly modifier to their constraints?

I'm hoping there's no need for that with #55229 since inference now adjusts itself based on constraints.

Issue #52670, opened 2023-02-08
Bloomberg feedback for 5.0 Beta


dragomirtitian:
By @molisani & @dragomirtitian 

We are in the process of evaluating the impact of upgrading to TypeScript 5.0 on our codebase. Here are some preliminary findings.

This is shaping up to be a medium impact release. The following table lists changes that affected our codebase, and where appropriate a link to a separate Issue:

| # |                   Change                              |      Affects     | Release notes | Packages affected| Reported as                                                         |
|---|-------------------------------------------------------|------------------|---------------|------------------|---------------------------------------------------------------------|
|  1| Union with any does not reduce to any                 | Type Checking    | Not Announced | ~6%              | [#52568](https://github.com/microsoft/TypeScript/issues/52568)      |
|  2| No implicit coercions in relational operators         | Type Checking    | [Yes](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0-beta/#forbidden-implicit-coercions-in-relational-operators)           | ~2%              |             |
|  3| Type inference for declared enum member assignment    | Type Checking    | Not Announced | ~2%              | [#52531](https://github.com/microsoft/TypeScript/issues/52531)      |
|  4| Circular constraints don’t type check sometimes       | Type Checking    | Not Announced | ~1%              | [#52570](https://github.com/microsoft/TypeScript/issues/52570)      |
|  5| `instanceof` narrows to `never` on else branch            | Type Checking    | Not Announced |  1               | [#52571](https://github.com/microsoft/TypeScript/issues/52571)      |
|  6| Nested callbacks are not contextually typed           | Type Checking    | Not Announced |  1               | [#52575](https://github.com/microsoft/TypeScript/issues/52575)      |
|  7| Declarations no longer include default type parameters| Declaration Emit | Not Announced | ~1%              |                                                                     |
|  8| Misc type checking improvements                       | Type Checking    | Not Announced |  2               |                                                                     |

## Union with `any` does not reduce to `any`

Generally `T | any` will reduce to any regardless of the origin of such a union. In TS 5.0 this is not the case for unions originating from a conditional type such as below. 

**Reported as ~~[#52568](https://github.com/microsoft/TypeScript/issues/52568)~~ (fixed)**

First appeared in `5.0.0-dev.20221208`. A cursory look at merged PRs doesn’t reveal an immediate culprit. 

```ts
type Spec = any extends object ? any : string;

type WithSpec<T extends number> = T

type R = WithSpec<Spec>
[...]

PR #55052, opened 2023-07-17
Support custom 'Symbol.hasInstance' methods when narrowing 'instanceof'

### LINKED ISSUE #17360: Use `Symbol.hasInstance` for `instanceof` narrowing, include constructor as part of class structural type
My proposal:

- For `instanceof` expressions, if `Symbol.hasInstance` exists on the RHS, narrow the expression according to the method's return value.
- For classes, implicitly add a type-level `constructor` prototype property which is `typeof classConstructor`, if such a property was not defined in the class body.

```ts
interface Foo { constructor: typeof Foo }
declare const Foo: {
    [Symbol.hasInstance](x: any): x is Foo
}

declare const Bar: {
    [Symbol.hasInstance](x: any): x is number
}

declare const foo: any
declare const bar: any

if (foo instanceof Foo) {
    // foo should be inferred as a Foo
}

if (bar instanceof Bar) {
    // bar should be inferred as a number
}
```

The rationale of this is that you can then (via compiler flag) by default define `Symbol.hasInstance` on constructors to type check against their return type (like for `new () => T`, defining `[Symbol.hasInstance](x: any): x is T`). This would mostly mitigate a certain unsoundness issue with classes, making the workaround a bit more difficult to do:

```ts
class Foo {
    bar: number = 1;
}

const foo1: Foo = new Foo() // Legal
const foo2: Foo = {bar: 2}  // No longer legal

// Making these illegal are out of scope of this proposal
const foo3: Foo = {
    constructor: Foo;
    bar: 2,
}
const foo4: Foo = new class {
    bar: number = 1
}
```

-----

Notes:

- This has no effect on emit.
- This has no effect on JS compatibility.
- This will create new type errors, and thus will be moderately breaking.
- This brings type semantics more in line with JS runtime semantics.
- This *shouldn't* have a substantial effect on compiler performance, since the default `Symbol.hasInstance` is equivalent to legacy `instanceof`.
- This would largely fix the premise of #17344 and would reduce the need for nominal typing (#202) in a common use case.

### LINKED ISSUE #39064: Types defining `Symbol.hasInstance` should be eligible to use with the `instanceof` operator
**TypeScript Version:**  3.9.4

As specified here, https://tc39.es/ecma262/#sec-instanceofoperator, it should be sufficient for an object to define a function for the `Symbol.hasInstance` well-known symbol in order to be used as the right-hand side of the `instanceof` operator.

It seems like this was intended to be implemented based on these comments: https://github.com/microsoft/TypeScript/pull/12728#issuecomment-269801430
but for whatever reason that didn't happen it in the PR

**Search Terms:** Symbol.hasInstance

**Code**

```ts
({}) instanceof {[Symbol.hasInstance]: () => true}
// Running this in the browser evaluates to `true`
```

**Expected behavior:**
No type-errors

**Actual behavior:**
```ts
({}) instanceof {[Symbol.hasInstance]: () => true}
                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
!!! error TS2359: The right-hand side of an 'instanceof' expression must be of type 'any' or of a type assignable to the 'Function' interface type.
```

**Playground Link:** [Playground Link](https://www.typescriptlang.org/play/?ssl=1&ssc=1&pln=1&pc=51#code/BQbwvglABAlgdgZwC4EM4GMCmB7AZlEAbQGUBPAWwCNsAbAOgAsUEBJRVDTAXQC4phoAXgB8UJACcArpjBA)

**Related Issues:**

rbuckton:
This makes our `instanceof` narrowing more closely align to Steps 2-3 of the [InstanceofOperator](https://tc39.es/ecma262/#sec-instanceofoperator) algorithm in the ECMA-262 specification by allowing any object type in the right-hand side of `instanceof` as long as it has a valid `[Symbol.hasInstance]` method. If such a method exists and returns a type predicate, that predicate can then be used when narrowing `instanceof`. 

```ts
interface Point { x: number, y: number };

const Point = {
  create(x: number, y: number): Point {
    return { x, y };
  },
  [Symbol.hasInstance](value: unknown): value is Point {
    return typeof value === "object" && value !== null &&
      "x" in value && typeof (value as Point).x === "number" &&
      "y" in value && typeof (value as Point).y === "number";
  }
};

interface Line { ... }

declare const p: Point | Line;
if (p instanceof Point) {
  p; // Point
} else {
  p; // Line
}
```

For a `[Symbol.hasInstance]` method to be considered during `instanceof` narrowing, it must have a type predicate as its return type. 

In addition, when the left-hand side of `instanceof` is considered as an argument to `[Symbol.hasInstance]`, it must be assignable to the first parameter. This does not, however, weaken our current requirement that the left-hand side be an object type (or a union containing an object type), despite the fact this is not currently a requirement of ECMAScript.

This is intended to address rather complicated workarounds employed by, for example, `engine262` which uses `[Symbol.hasInstance]` as a runtime type guard, but needed to introduce a non-creatable abstract class so that TypeScript would understand the type guard: https://github.com/engine262/engine262/blob/435e533f8bb2f66961e3c16fe67150a0bba745bb/src/completion.mts#L70

Fixes #17360
Fixes #39064
Related #32080
Related #52670

NOTE: This will need to wait for 5.3 at the earliest.


gabritto:
@typescript-bot pack this

src/compiler/checker.ts
```diff
@@ -36458,9 +36534,50 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+                    // must check the expression as if it were a call to `right[Symbol.hasInstance](left)1. The call to
+                    // `getResolvedSignature`, below, will check that leftType is assignable to the type of the first
+                    // parameter.
+                    const syntheticCall = createSyntheticHasInstanceMethodCall(left, right, leftType, hasInstanceMethodType);
+                    const returnType = getReturnTypeOfSignature(getResolvedSignature(syntheticCall));
+
+                    // We also verify that the return type of the `[Symbol.hasInstance]` method is assignable to
+                    // `boolean`. According to the spec, the runtime will actually perform `ToBoolean` on the result,
+                    // but this is more type-safe.
+                    checkTypeAssignableTo(returnType, booleanType, right, Diagnostics.An_object_s_Symbol_hasInstance_method_must_return_a_boolean_value_for_it_to_be_used_on_the_right_hand_side_of_an_instanceof_expression);
```
gabritto, line 36565: If we we check if the return type is assignable to `bool` here, why don't we check the return type in the implementations of `[Symbol.hasInstance]` methods?

src/compiler/checker.ts
```diff
@@ -36458,9 +36534,50 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+        if (!isTypeAny(rightType)) {
+            // if rightType is an object type with a custom `[Symbol.hasInstance]` method, then it is potentially
+            // valid on the right-hand side of the `instanceof` operator. This allows normal `object` types to
+            // participate in `instanceof`, as per Step 2 of https://tc39.es/ecma262/#sec-instanceofoperator.
+            const hasInstanceMethodType = getSymbolHasInstanceMethodOfObjectType(rightType);
+            if (hasInstanceMethodType) {
+                // avoid a complex check for every `instanceof` when the `[Symbol.hasInstance]` method has a single
+                // call signature that neither restricts nor narrows (via type predicate) the LHS value, e.g.
+                // `(value: unknown) => boolean`.
+                const cache = hasInstanceMethodType as HasInstanceMethodType;
```
gabritto, line 36546: I'm curious about how you know this optimization is worth it.

src/compiler/checker.ts
```diff
@@ -33153,7 +33177,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         return createDiagnosticForNodeArray(getSourceFileOfNode(node), typeArguments, Diagnostics.Expected_0_type_arguments_but_got_1, belowArgCount === -Infinity ? aboveArgCount : belowArgCount, argCount);
     }
 
-    function resolveCall(node: CallLikeExpression, signatures: readonly Signature[], candidatesOutArray: Signature[] | undefined, checkMode: CheckMode, callChainFlags: SignatureFlags, headMessage?: DiagnosticMessage): Signature {
+    function resolveCall(node: CallLikeExpression, signatures: readonly Signature[], candidatesOutArray: Signature[] | undefined, checkMode: CheckMode, callChainFlags: SignatureFlags, headMessageCallback?: () => DiagnosticMessage | undefined): Signature {
```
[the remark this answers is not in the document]
rbuckton: It's an optimization to avoid the work performed by `isSyntheticHasInstanceMethodCall` for every call expression, since this only needs to be computed in the case of an error.

src/compiler/checker.ts
rbuckton: We have no way to validate a `[Symbol.hasInstance]` method on a given object without prompting, and its implementation only matters when it is going to be used as part of `instanceof`, just as we don't validate that a given `[Symbol.iterator]()` method is valid until it is used in places where a valid iterable is expected.

src/compiler/checker.ts
rbuckton: At the moment, purely based on experience, though in general I would validate this assumption through benchmarking. Without a cache we would be performing a number of additional checks over and over again against the default implementation of `[Symbol.hasInstance](value)` on `Function` for every occurrence of `instanceof` in a given codebase, and I'd like to avoid performing additional work against existing code.

This does incur additional polymorphism in `Type`, but `Type` is already megamorphic and I'd prefer to work on decreasing `Type` polymorphism in a separate PR.

rbuckton:
@typescript-bot perf test
@typescript-bot run dt
@typescript-bot test this

gabritto (approved):

rbuckton:
@typescript-bot pack this

jakebailey:
Leaving a random comment to test the new perf infra on this PR.

sandersn (changes requested):

src/compiler/checker.ts
```diff
@@ -34307,7 +34331,12 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             return resolveErrorCall(node);
         }
 
-        return resolveCall(node, callSignatures, candidatesOutArray, checkMode, callChainFlags);
+        // If the call expression is a synthetic call to a `[Symbol.hasInstance]` method then we will produce a head
+        // message when reporting diagnostics that explains how we got to `right[Symbol.hasInstance](left)` from
+        // `left instanceof right`, as it pertains to "Argument" related messages reported for the call.
+        return resolveCall(node, callSignatures, candidatesOutArray, checkMode, callChainFlags, () => isSyntheticHasInstanceMethodCall(node) ?
```
sandersn, line 34337: is there another way to conditionally issue this error? I guess you'd have to check each time you issue the error, but I would guess it would only have to be in a couple of places, not the number of places that had to change to make this a callback in the current commit.

src/compiler/checker.ts
```diff
@@ -46679,6 +46795,21 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             case SyntaxKind.ImportKeyword:
             case SyntaxKind.NewKeyword:
                 return isMetaProperty(node.parent) ? checkMetaPropertyKeyword(node.parent).symbol : undefined;
+            case SyntaxKind.InstanceOfKeyword:
+                if (isBinaryExpression(node.parent)) {
+                    const type = getTypeOfExpression(node.parent.right);
+                    const hasInstanceMethodType = getSymbolHasInstanceMethodOfObjectType(type);
+                    // const hasInstancePropertyName = getPropertyNameForKnownSymbolName("hasInstance");
```
sandersn, line 46802: these comments look out of date

src/compiler/checker.ts
```diff
@@ -46679,6 +46795,21 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             case SyntaxKind.ImportKeyword:
             case SyntaxKind.NewKeyword:
                 return isMetaProperty(node.parent) ? checkMetaPropertyKeyword(node.parent).symbol : undefined;
+            case SyntaxKind.InstanceOfKeyword:
```
sandersn, line 46798: 'getSymbolAtLocation' is more and more of a misnomer -- this has become 'getUsefulLinkAtLocation'. Which is useful I guess.

tests/cases/compiler/controlFlowInstanceofWithSymbolHasInstance.ts
```diff
@@ -0,0 +1,118 @@
...
+interface PromiseConstructor {
+    [Symbol.hasInstance](value: any): value is Promise<any>;
+}
+
+interface SetConstructor {
+    [Symbol.hasInstance](value: any): value is Set<any>;
+}
+
+function f1(s: Set<string> | Set<number>) {
+    s = new Set<number>();
```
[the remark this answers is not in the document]
rbuckton: This is to make sure using a custom `[Symbol.hasInstance]` method with an appropriate type guard _doesn't_ break existing cases.

tests/cases/compiler/controlFlowInstanceofWithSymbolHasInstance.ts
```diff
@@ -0,0 +1,118 @@
...
+        never
+    ) {
+        return Function.prototype[Symbol.hasInstance].call(this, value);
+    }
+}
+class B extends A { b: string }
+class C extends A { c: string }
+
+function foo(x: A | undefined) {
+    x;  // A | undefined
```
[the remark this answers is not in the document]
rbuckton: That is sort of the point. I want to make sure the behavior of control flow narrowing is identical when using a custom `[Symbol.hasInstance]` with an appropriate type predicate.

src/compiler/diagnosticMessages.json
```diff
@@ -3663,6 +3663,18 @@
...
     },
+    "The left-hand side of an 'instanceof' expression must be assignable to the first argument of the right-hand side's '[Symbol.hasInstance]' method.": {
+        "category": "Error",
+        "code": 2855
+    },
+    "An object's '[Symbol.hasInstance]' method must return a boolean value for it to be used on the right-hand side of an 'instanceof' expression.": {
+        "category": "Error",
+        "code": 2856
+    },
+    "The right-hand side of an 'instanceof' expression must be either of type 'any', an object type with a '[Symbol.hasInstance]()' method, or a type assignable to the 'Function' interface type.": {
```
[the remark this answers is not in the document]
rbuckton: > 1. I'd rather add hasInstance at the end of the list.
> 2. This is an existing problem, but the Function entry should mention classes somehow. Maybe:
>    "a type assignable to the type 'Function', such as a class,"

How about:

> The right-hand side of an 'instanceof' expression must be either of type 'any', a class, function, or other type assignable to the 'Function' interface type, or an object type with a '[Symbol.hasInstance]()' method.

src/services/goToDefinition.ts
```diff
@@ -102,7 +102,7 @@ import {
     TypeChecker,
     TypeFlags,
     TypeReference,
-    unescapeLeadingUnderscores,
+    unescapeLeadingUnderscores
```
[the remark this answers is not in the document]
rbuckton: Probably a result of Organize Imports

tests/baselines/reference/typeGuardsWithInstanceOfBySymbolHasInstance.types
```diff
@@ -0,0 +1,572 @@
...
+>obj9 : A | E1
+
+if (obj9 instanceof E) { // narrowed to E1 | E2
+>obj9 instanceof E : boolean
+>obj9 : A | E1
+>E : EConstructor
+
+    obj9.foo;
+>obj9.foo : string
+>obj9 : E1
```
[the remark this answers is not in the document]
rbuckton: Both the comments here and for the same line in `typeGuardsWithInstanceOfByConstructorSignature.ts` are incorrect.

src/compiler/diagnosticMessages.json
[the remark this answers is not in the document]
sandersn: Yes, that's much better.

src/compiler/checker.ts
rbuckton: `resolveCall` already allowed you to pass in a `headMessage`, which was the way you would conditionally choose to append this message. I changed this to a callback so as to lazily compute the head message rather than do eager, expensive work up front when there may be no error to report.

I could pass in an _additional_ `boolean` parameter that moves this computation to the point before we issue an error, or add a bitflag to `CheckMode` possibly.

rbuckton: Actually, I don't need an extra parameter. I can just perform the check prior to reporting errors in absence of a more specific `headMessage`.

sandersn (approved):

src/compiler/checker.ts
```diff
@@ -28031,7 +28035,22 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             return getNarrowedType(type, instanceType, assumeTrue, /*checkDerived*/ true);
         }
 
-        function getInstanceType(constructorType: Type) {
+        function getInstanceType(constructorType: Type, left: Expression, leftType: Type, right: Expression) {
+            // if rightType is an object type with a custom `[Symbol.hasInstance]` method, and that method has a type
```
ahejlsberg, line 28039: Shouldn't `rightType` be `constructorType`?

src/compiler/checker.ts
```diff
@@ -28031,7 +28035,22 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+            // in `instanceof`, as per Step 2 of https://tc39.es/ecma262/#sec-instanceofoperator.
+            const hasInstanceMethodType = getSymbolHasInstanceMethodOfObjectType(constructorType);
+            if (hasInstanceMethodType) {
+                const syntheticCall = createSyntheticHasInstanceMethodCall(left, leftType, right, constructorType, hasInstanceMethodType);
+                const signature = getEffectsSignature(syntheticCall);
+                const predicate = signature && getTypePredicateOfSignature(signature);
+                if (predicate && predicate.kind === TypePredicateKind.Identifier && predicate.parameterIndex === 0) {
+                    return predicate.type;
+                }
+                if (!isTypeDerivedFrom(constructorType, globalFunctionType)) {
```
ahejlsberg, line 28050: When does this happen?

src/compiler/checker.ts
```diff
@@ -28031,7 +28035,22 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
         }
 
-        function getInstanceType(constructorType: Type) {
+        function getInstanceType(constructorType: Type, left: Expression, leftType: Type, right: Expression) {
+            // if rightType is an object type with a custom `[Symbol.hasInstance]` method, and that method has a type
+            // predicate, use the type predicate to perform narrowing. This allows normal `object` types to participate
+            // in `instanceof`, as per Step 2 of https://tc39.es/ecma262/#sec-instanceofoperator.
+            const hasInstanceMethodType = getSymbolHasInstanceMethodOfObjectType(constructorType);
+            if (hasInstanceMethodType) {
+                const syntheticCall = createSyntheticHasInstanceMethodCall(left, leftType, right, constructorType, hasInstanceMethodType);
```
ahejlsberg, line 28044: I'm surprised at all the complexity with synthetic method calls and so on. Isn't it just a matter of finding the `hasInstance` method, checking that it has a type predicate return type, and fishing the target type out of that. Why do we need synthetic calls and overload resolution?

src/compiler/checker.ts
```diff
@@ -33873,6 +33895,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         // If candidate is undefined, it means that no candidates had a suitable arity. In that case,
         // skip the checkApplicableSignature check.
         if (reportErrors) {
+            if (!headMessage && isCallExpression(node) && isSyntheticHasInstanceMethodCall(node)) {
```
ahejlsberg, line 33898: This all seems like complexity that follows from the synthetic call business. Do we even need to do call and overload resolution in the first place?

ahejlsberg (commented):
This seems a lot more complex than I had expected. I'm not quite understanding why we need synthetic method calls and potentially overload resolution to figure this out.

src/compiler/checker.ts
rbuckton: This is necessary to cover the case originally on original line 28019 where we would have previously exited early: https://github.com/microsoft/TypeScript/pull/55052/files/befa293a880c91110ad57b6d34e924363e8ace85#diff-d9ab6589e714c71e657f601cf30ff51dfc607fc98419bf72e04f6b0fa92cc4b8L28019

Instead, we're now performing this check on each constituent of the union, as not all constituents are necessarily going to be function types.

src/compiler/checker.ts
rbuckton: It would be perfectly reasonable to assume that a `[Symbol.hasInstance]()` method could have overloads, just as any function or method with a type predicate. The synthetic call gives us the ability to leverage normal call resolution to be consistent with how the transformed representation would be handled:

```js
x instanceof y
// is equiv. to
y[Symbol.hasInstance](x)
```

src/compiler/checker.ts
rbuckton: It seems to me that we should, otherwise we would be drastically inconsistent with how the transposed representation works, as I mentioned in [another comment](https://github.com/microsoft/TypeScript/pull/55052#discussion_r1333852020).

rbuckton: We also use the synthetic expression mechanism to properly type check tagged templates, decorators, and spread arguments where synthesizing a transposed representation makes type checking far easier than one off mechanisms.

rbuckton: The synthetic call is actually far less complicated than the mechanism we currently employ for "call-like expressions" like tagged templates, decorators, and Jsx elements, which depend on threading those expressions through `getResolvedSignature`, `resolveSignature`, and their own independent `resolveTaggedTemplateExpression`/`resolveDecorator`/`resolveJsxOpeningLikeElement` functions, as well as all of the complexity behind `getEffectiveCallArguments` et al. In the long term, it might actually be cleaner for us to generate synthetic `CallExpression` nodes for those expressions instead, and remove all of the complicated branching we do there.

src/compiler/checker.ts
```diff
@@ -37108,9 +37182,50 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+        if (!isTypeAny(rightType)) {
+            // if rightType is an object type with a custom `[Symbol.hasInstance]` method, then it is potentially
+            // valid on the right-hand side of the `instanceof` operator. This allows normal `object` types to
+            // participate in `instanceof`, as per Step 2 of https://tc39.es/ecma262/#sec-instanceofoperator.
+            const hasInstanceMethodType = getSymbolHasInstanceMethodOfObjectType(rightType);
+            if (hasInstanceMethodType) {
+                // avoid a complex check for every `instanceof` when the `[Symbol.hasInstance]` method has a single
+                // call signature that neither restricts nor narrows (via type predicate) the LHS value, e.g.
+                // `(value: unknown) => boolean`.
+                const cache = hasInstanceMethodType as HasInstanceMethodType;
```
rbuckton, line 37194: Most of the complexity here is a result of a performance optimization to avoid paying the cost of looking for `[Symbol.hasInstance]` methods when they either can't exist (because the symbol isn't defined), or have no meaningful impact (because their definition does not restrict the input argument and performs no narrowing).

rbuckton:
The most recent commits address feedback from an offline discussion with @ahejlsberg. @ahejlsberg, can you take another look?

src/compiler/checker.ts
rbuckton: This has been switched over to use the `getResolvedSignature` logic.

src/compiler/checker.ts
```diff
@@ -34838,6 +34879,43 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+                }
+
+                if (callSignatures.length) {
+                    return resolveCall(node, callSignatures, candidatesOutArray, checkMode, SignatureFlags.None);
+                }
+            }
+            // NOTE: do not raise error if right is unknown as related error was already reported
+            else if (!(typeHasCallOrConstructSignatures(rightType) || isTypeSubtypeOf(rightType, globalFunctionType))) {
+                // Do not indicate that `[Symbol.hasInstance]` is a valid option if it's not known to be present on `SymbolConstructor`.
+                const message = hasGlobalSymbolHasInstanceProperty() ?
```
ahejlsberg, line 34908: Do we really care about this distinction. I think it is fine to always report the message that includes `hasInstance`. That would also allow you to get rid of the `hasGlobalSymbolHasInstanceProperty` and the associated cache variable, which only seems to be used here.

ahejlsberg (approved):
Suggest simplifying the error reporting per comments, otherwise looks good.

src/compiler/checker.ts
rbuckton: I included the distinction because without it, there wasn't a clear correlation between `instanceof` and `Symbol.hasInstance` if you weren't familiar with the spec machinery.

rbuckton:

jakebailey:
I have no idea why the pick failed; I was going to say that it failed a the push stage but it got pushed above: https://github.com/typescript-bot/TypeScript/commit/97bf30b22d91a816a8195c6d2fa175897fb52871

Probably some weird GH outage :|

(I am working on replacing this pick task anyway.)

rbuckton:
It doesn't matter too much. Daniel said he'd just update from `main` anyways.

PR #55053, opened 2023-07-17
Fix CJS export of typedef and class w/latebound names

### LINKED ISSUE #53967: Unrelated JSDoc annotation causes class to not be assignable to itself or iterable
# Bug Report

### 🔎 Search Terms

jsdoc annotation iterable self assignable

- This changed between versions 4.2 and 4.3

### 💻 Code

Here's a compiler repro that reliably works.

```ts
// @lib: es2017
// @allowJs: true
// @checkJs: true
// @noEmit: true

// @filename: index.js
const LazySet = require("./LazySet");

/** @type {LazySet<string>} */
const stringSet = new LazySet();

stringSet.addAll(stringSet);


// @filename: LazySet.js
// Comment out this JSDoc, and note that the errors index.js go away.
/**
 * @typedef {Object} SomeObject
 */
void 0;

/**
 * @template T
 */
class LazySet {
    /**
     * @param {Iterable<T>=} iterable
     */
    constructor(iterable) {
        /** @type {Set<T>} */
        this._set = new Set(iterable);
    }

    /**
     * @param {Iterable<T> | LazySet<T>} iterable
     * @returns {LazySet<T>}
     */
    addAll(iterable) {
        return this;
    }

    [Symbol.iterator]() {
        return this._set[Symbol.iterator]();
    }
}

module.exports = LazySet;
```

### 🙁 Actual behavior

https://github.com/webpack/webpack/pull/16957#discussion_r1174262622

```
Argument of type 'LazySet<string>' is not assignable to parameter of type 'Iterable<string> | LazySet<string>'.
  Property '[Symbol.iterator]' is missing in type 'import("/home/jabaile/work/webpack-repro/lib/LazySet")<string>' but required in type 'import("/home/jabaile/work/webpack-repro/lib/LazySet")<string>'.ts(2345)
LazySet.js(27, 2): '[Symbol.iterator]' is declared here.
```

Note that `LazySet` is not assignable to `LazySet`, which is weird. Changing the signature of `addAll` to just be `Iterable<T>` is equivalent.

Also, if you actually construct this example in the editor, editing `index.js` will make the error go away temporarily, until tsserver is restarted, or `LazySet.js` is edited. Very odd.

### 🙂 Expected behavior

No errors.

@alexander-akait @TheLarkInn

sandersn:
Fixes #53967

The problem is that the weird, ad-hoc CJS merge needs to happen at two different times.  The first is its existing location in getCommonJsExportEquals. The second is getResolvedMembersOrExportsOfSymbol, which combines normal, early-bound symbols with late-bound symbols.

The CJS merge happens before the addition of late-bound symbols, so a merge will result in the loss of [Symbol.iterable] from the bug. But manually merging the resolvedExports/Members during the CJS merge also doesn't work, because it prevents the module's resolvedExports/Members from being filled in later.

Instead, this PR delays part of the CJS merge to
getResolvedMembersOrExportsOfSymbol. Unfortunately, this requires retrieving the original, unmerged symbols--since their resolvedExports/Members are already cached. I do this by looking at the symbols for each of the merged symbol's declarations, which feels wrong.

Please suggest alternate ways to do this. I don't think I understand the late-binding machinery well enough, so it's likely there's a cleaner way.


weswigham (approved):
> I don't think I understand the late-binding machinery well enough, so it's likely there's a cleaner way.

The cleaner way is probably breaking the cycle between late-binding and cjs export merging, I think by making it so module members bind directly into the `export=` symbol if it's present (which probably requires a 2-pass process), rather than the containing module - basically moving what `getCommonJsExportEquals` does up into the binder phase. There's probably a bunch of knock-on effects to this, though, since I know the js declaration emitter, at least, already does a bunch of hacks (or at least stuff similar to this where it goes back and looks at the original unmerged symbols) to work around the odd symbol structure `getCommonJsExportEquals` creates.

In any case, this is probably fine - it's just piling a similar workaround into an already ugly space in the compiler - but maybe @rbuckton has more thoughts from the late bound member angle?

PR #48091, opened 2022-03-02
Preserve the homomorphism of inlined mapped types in declaration emit

weswigham:
Fixes #46655 - declaration emit now preserves the homomorphism of mapped types by emitting a conditional type with an `infer` type parameter when needed. So instead of, eg, `{ [K in keyof T[any]]: T[any][K]; }` we now emit `T[any] extends infer T_1 ? { [K in keyof T_1]: T[any][K]; } : never`.

PR #55054, opened 2023-07-17
Declaration emit for inlined mapped types preserves modifier-preserving behavior

### LINKED ISSUE #55002: Emitted mapped type in declarations loses modifiers
# Bug Report

### 🔎 Search Terms

modifiers declaration dts emit mapped type omit pick

- This is the behavior in every version I tried (at least since 4.7)

### ⏯ Playground Link

The repro case involves multiple files, 2 packages and declaration emit so I had to prepare a repository with it:
https://github.com/Andarist/mapped-type-emit-loses-modifiers-bug

### 💻 Code

```ts
// packages/pkg-a/src/other.ts

// how Omit from lib is defined
type OmitReal<T, K extends keyof any> = Pick<T, Exclude<keyof T, K>>;
// what we see when we hover it
type OmitUnveiled<T, K extends string | number | symbol> = {
  [P in Exclude<keyof T, K>]: T[P];
};

export function test1<T, K extends string>(obj: T, k: K): OmitReal<T, K> {
  return {} as any;
}

export function test2<T, K extends string>(obj: T, k: K): OmitUnveiled<T, K> {
  return {} as any;
}

// packages/pkg-a/src/index.ts

import { test1, test2 } from "./other";

export function wrappedTest1<T, K extends string>(obj: T, k: K) {
  return test1(obj, k);
}

export function wrappedTest2<T, K extends string>(obj: T, k: K) {
  return test2(obj, k);
}

export type Obj = {
  a: number;
  readonly foo: string;
};

export const processedInternally1 = wrappedTest1({} as Obj, "a");
export const processedInternally2 = wrappedTest2({} as Obj, "a");

// packages/pkg-b/src/index.ts

import { wrappedTest1, wrappedTest2, Obj } from "pkg-a";

export const processedExternally1 = wrappedTest1({} as Obj, "a");
export const processedExternally2 = wrappedTest2({} as Obj, "a");
```

### 🙁 Actual behavior

The mapped type starts to lose modifiers when it's emitted even though modifiers are retained when the same thing is used within the package.

This can be seen in the emitted `.d.ts` for each of those packages. Both are using the same function and the same arguments but yet the type emitted in both places is different:

In `pkg-a` we can see [this](https://github.com/Andarist/mapped-type-emit-loses-modifiers-bug/blob/main/packages/pkg-a/types/index.d.ts#L7-L12):
```ts
export declare const processedInternally1: {
    readonly foo: string;
};
export declare const processedInternally2: {
    foo: string;
};
```

In `pkb-b` we can see [this](https://github.com/Andarist/mapped-type-emit-loses-modifiers-bug/blob/aa088bcba80b4c5b5d2f52b122ffed52f814fd4c/packages/pkg-b/types/index.d.ts#L1-L6):
```ts
export declare const processedExternally1: {
    foo: string;
};
export declare const processedExternally2: {
    foo: string;
};
```

It all starts with the fact that `OmitReal` and `OmitUnveiled` are not exported from `packages/pkg-a/src/other.ts` so `packages/pkg-a/types/index.d.ts` has to inline them in its declarations and both end up being emitted the same ([here](https://github.com/Andarist/mapped-type-emit-loses-modifiers-bug/blob/main/packages/pkg-a/types/index.d.ts#L1-L2)):
```ts
export declare function wrappedTest1<T, K extends string>(obj: T, k: K): { [P in Exclude<keyof T, K>]: T[P]; };
export declare function wrappedTest2<T, K extends string>(obj: T, k: K): { [P in Exclude<keyof T, K>]: T[P]; };
```

Their original declarations have **different** traits when it comes to preserving property modifiers though and thus this is a bug.

### 🙂 Expected behavior

I would expect the emitted declaration to behave in the same way as the original one.

weswigham:
Fixes #55002

This is somewhat similar to our custom declaration emit that wraps mapped types to preserve homomorphic-ness (circa #48091), if needs be, but is a slightly different (more verbose) wrapper to preserve just the modifier-preserving behavior of the emitted mapped type.


src/compiler/checker.ts
```diff
@@ -6639,6 +6639,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 const questionToken = type.declaration.questionToken ? factory.createToken(type.declaration.questionToken.kind) as QuestionToken | PlusToken | MinusToken : undefined;
                 let appropriateConstraintTypeNode: TypeNode;
                 let newTypeVariable: TypeReferenceNode | undefined;
+                // If the mapped type isn't `keyof` constraint-declared, _but_ still has modifiers preserved, and its' naive instantiation won't preserve modifiers because its' constraint isn't `keyof` constrained, we have work to do
```
sandersn, line 6642:
```suggestion
                // If the mapped type isn't `keyof` constraint-declared, _but_ still has modifiers preserved, and its naive instantiation won't preserve modifiers because its constraint isn't `keyof` constrained, we have work to do
```

src/compiler/checker.ts
```diff
@@ -6639,6 +6639,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 const questionToken = type.declaration.questionToken ? factory.createToken(type.declaration.questionToken.kind) as QuestionToken | PlusToken | MinusToken : undefined;
                 let appropriateConstraintTypeNode: TypeNode;
                 let newTypeVariable: TypeReferenceNode | undefined;
+                // If the mapped type isn't `keyof` constraint-declared, _but_ still has modifiers preserved, and its' naive instantiation won't preserve modifiers because its' constraint isn't `keyof` constrained, we have work to do
+                const needsModifierPreservingWrapper = !isMappedTypeWithKeyofConstraintDeclaration(type) && !(getModifiersTypeFromMappedType(type).flags & TypeFlags.Unknown) && context.flags & NodeBuilderFlags.GenerateNamesForShadowedTypeParams
+                    && !(getConstraintTypeFromMappedType(type).flags & TypeFlags.TypeParameter && getConstraintOfTypeParameter(getConstraintTypeFromMappedType(type))?.flags! & TypeFlags.Index);
```
sandersn, line 6644: dprint when (this is hard to read with 2 predicates per line, each of which are so long that they wrap on my screen)

src/compiler/checker.ts
```diff
@@ -6670,6 +6681,18 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                         factory.createKeywordTypeNode(SyntaxKind.NeverKeyword)
                     );
                 }
+                else if (needsModifierPreservingWrapper) {
+                    // and step 3: once the mapped type is reconstructed, create a `ConstraintType extends infer T_1 extends keyof ModifiersType ? {[K in T_1]: Template} : never`
+                    // subtly different from the `keyof` constraint case, by including the `keyof` constraint on the `infer` type parameter, it doesn't rely on the constraint type being itself
+                    // constrained to a `keyof` type to preserve it's modifier-preserving behavior. This is all basically because we preserve modifiers for a wider set of mapped types than
```
sandersn, line 6687:
```suggestion
                    // constrained to a `keyof` type to preserve its modifier-preserving behavior. This is all basically because we preserve modifiers for a wider set of mapped types than
```

double-checking: is the effect of this PR that we now preserve modifiers for these mapped types, and didn't before?

sandersn (approved):
Looks right, just a few formatting nits.

src/compiler/checker.ts
Andarist: without reading the code of the fix itself - yes, the issue that I reported is specifically about the fact that modifiers are lost but they shouldn't be so the fix for that issue should preserve them

Andarist:
IIUC the test case from my issue is meant to preserve modifiers but is not meant to make the given mapped type distributive. This is handled correctly by this fix. Is there an extra test case to be written here that would test that both are preserved? OTOH, probably when distributivity is preserved, it also implies preserving the modifiers so in practice. We don't have 4 combinations but rather just 3 (nothing is preserved, modifiers are preserved, both are preserved).

weswigham:
> Is there an extra test case to be written here that would test that both are preserved? OTOH, probably when distributivity is preserved, it also implies preserving the modifiers so in practice. We don't have 4 combinations but rather just 3 (nothing is preserved, modifiers are preserved, both are preserved).

Distributivity of mapped types comes from homomorphic-ness (which also preserves modifiers), which we have a separate emit form to preserve when inlining breaks it, which already has its' own tests. You're correct that there's 3 relevant states (distributive and modifiers preserved, only modifiers preserved, nothing preserved) and that this PR covers the missing one (only modifiers preserved)~

PR #55031, opened 2023-07-15
fix(54992): Negative constant enums return number when toString is called

a-tarasyuk:
Fixes #54992

PR #55065, opened 2023-07-18
Wrap inline negative const enum with parens

### LINKED ISSUE #54992: Negative constant enums return number when toString is called
### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.1.6#code/MYewdgzgLgBApmArgWxgFTtAok1BvAKBmJgEEYBeGARgBoiSAhSmAWmoIF8CDRIQANnAB0AkAHMAFA2JQAngAc4IAGYwAylABOASzBSM2XMNIBKeiRjylq9Jig4UJ4VBCbd+yaYLfe4CIIiYlIyVorKau56BvaOyMKM5qHWEXZGTowubtrRXj5AA

### 💻 Code

```ts
const enum TestEnum {
    A = 1,
    B = -1
}

// "string",  "string" 
console.log(
    typeof String(TestEnum.A),
    typeof TestEnum.A.toString()
)

// "string",  "number" 
console.log(
    typeof String(TestEnum.B),
    typeof TestEnum.B.toString()
)
```

### 🙁 Actual behavior

Constant enums which use negative numbers will return type of `number` when `.toString()` is called.

### 🙂 Expected behavior

A `string` type should be returned

Andarist:
supersedes https://github.com/microsoft/TypeScript/pull/55031
fixes #54992

cc @a-tarasyuk @jakebailey


tests/baselines/reference/castExpressionParentheses.js
```diff
@@ -76,9 +76,9 @@ a;
 a[0];
 a.b["0"];
 a().x;
-1..foo;
-1..foo;
-1.0.foo;
+(1).foo;
+(1.).foo;
```
Andarist, line 80: It might be worth re-using the original style from the input here (something that is the explicit intention of some related lines that I was tweaking) and I can fix this particular line by doing that.

tests/baselines/reference/destructuringControlFlow.js
```diff
@@ -73,8 +73,8 @@ function f3(obj) {
 function f4() {
     var _a, _b;
     var x;
-    (x = 0..x); // Error
-    (x = 0["x"]); // Error
+    (x = (0).x); // Error
+    (x = (0)["x"]); // Error
```
Andarist, line 77: This isn't incorrect but those extra parentheses could be avoided if I'd detect the access style~

tests/baselines/reference/constEnumToStringWithComments.js
```diff
@@ -25,15 +25,15 @@ let c1 = Foo["C"].toString();
...
 var y1 = 0.5 /* Foo["Y"] */.toString();
-var z0 = 2 /* Foo.Z */.toString();
-var z1 = 2 /* Foo["Z"] */.toString();
-var a0 = -1 /* Foo.A */.toString();
-var a1 = -1 /* Foo["A"] */.toString();
-var b0 = -1.5 /* Foo.B */.toString();
-var b1 = -1.5 /* Foo["B"] */.toString();
-var c0 = -1 /* Foo.C */.toString();
-var c1 = -1 /* Foo["C"] */.toString();
+var z0 = (2 /* Foo.Z */).toString();
```
Andarist, line 32: `2.toString()` is invalid but `2/**/.toString` is so those parenthesis could also be avoided here, OTOH I'm not really sure if it's worth avoiding them here 🤷‍♂️

src/compiler/emitter.ts
```diff
@@ -3055,17 +3052,7 @@ export function createPrinter(printerOptions: PrinterOptions = {}, handlers: Pri
...
-                && !stringContains(text, String.fromCharCode(CharacterCodes.E))
-                && !stringContains(text, String.fromCharCode(CharacterCodes.e));
-        }
-        else if (isAccessExpression(expression)) {
-            // check if constant enum value is integer
-            const constantValue = getConstantValue(expression);
-            // isFinite handles cases when constantValue is undefined
-            return typeof constantValue === "number" && isFinite(constantValue)
-                && Math.floor(constantValue) === constantValue;
+            return !isNumericLiteralTextSafeForPropertyAccess(expression, text);
```
Andarist, line 3055: I need to recheck when this line is actually hit - if the team's preference would be to just replace the "dot dot" strategy with parentheses then likely this whole function here could be just removed. Thoughts?

tests/baselines/reference/castExpressionParentheses.js
```diff
@@ -76,9 +76,9 @@ a;
 a[0];
 a.b["0"];
 a().x;
-1..foo;
-1..foo;
-1.0.foo;
+(1).foo;
```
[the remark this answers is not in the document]
jakebailey: I'm torn; I think that adding these parens is a little overzealous; my main comparison is esbuild which instead inserts a space before the dot, which is roughly equivalent to a double dot, but it seems like there's no particular consistency in the ecosystem besides double-dot or space being one character fewer for the same thing.

Of course, esbuild _always_ includes the comment too unless you specify `minify=true` (the highest minification), and tsc isn't a minifier, so it's not like it's ridiculous to work differently. esbuild will also convert `(1).toString()` to `1 .toString()` so who knows.

Example: [esbuild playground](https://jakebailey.dev/esbuild-playground/#v3=PTAEAEDMEsBsFMB2BDAtvAXKAxge0TAOYB0AVgM74BQA3laKCKAESrSLSQCezWALgCcArvCoBfKlSZQ4SNJlDsAJvAAexPuVVU8icn1BIhqUAFFQdBgwCCoALygAfgEYANPSsAhe6AC0bjwYAYR9HAApnUABqUGcASncrUAARH18I6NAAJgTAsx9I31iABncJKlNiaw1cAGVBdkIwuIBuCuJPGvqBRua2yqCuhsQm1vbkoZ6RvslTAG1ma2YAXUnesfnmTxW16Y2FoJ2+OuHR-oXko5Ops4qF0yvu9bbJHXx9Q0RjUAAxXFwLHkABoFYqlPIATR8xWIAFZElYAFo+LLEBE2NIBJLeBz+OHo0AhXHOYjiSQIAyqYo+P64YhA3a3CmgVSRBy0hZAx6nGbMrjU9n-YgQxm8+AGLhs37-BYQ7k3MUGABeAuldMRorGzKVUo5zER8ueVGZyFVtKqmraJt1MsWhr2VvFoAARmahZ1jk8Hcanc6bbgFttVp6eVqndg3XTBiGFWGDNh-Qd7bc3npcAhiLBcE0InFNXEgA)

jakebailey:
Probably @rbuckton has the most well-formed opinion here.

tests/baselines/reference/castExpressionParentheses.js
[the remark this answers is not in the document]
rbuckton: I'd rather stick with the `..` emit wherever possible and only parenthesize when it is absolutely necessary, i.e., when the number itself contains `-` or if the left-hand-side of a property access/element access is a unary `+`/`-` whose `expression` is a numeric literal.

rbuckton:
@Andarist: I can put up a PR for this that introduces parens in far fewer places, but if you'd rather do it the simplest approach would be to fix `substituteConstantValue` to avoid returning negative numeric literals and instead return parenthesized, prefix unary wrapped numeric literals.

We can reduce unnecessary parentheses even more if _don't_ always parenthesize the prefix unary in `substituteConstantValue` and we instead modify `substitutePropertyAccessExpression` and `substituteElementAccessExpression` to run substitutions early on the node's `expression` so that we can trigger the parenthesizer during substitution. Something like this:

```ts
function substitutePropertyAccessExpression(node: PropertyAccessExpression) {
    if (isAccessExpression(node.expression)) {
        node = isPropertyAccessChain(node) ?
            factory.updatePropertyAccessChain(node, substituteExpression(node.expression), node.questionDotToken, node.name):
            factory.updatePropertyAccessExpression(node, substituteExpression(node.expression), node.name);
    }
    return substituteConstantValue(node);
}
```

Note that the second approach doesn't cover other MemberExpressions like NewExpression or TaggedTemplateExpression. We could consider supporting those cases, though you're not likely to see `new E.Foo()` or `` E.Foo`bar` `` (where `E` is an enum) in the wild and it not already be an error.

Andarist:
@rbuckton i specifically didnt fix this while doing const substitution because that felt like a targeted fix. I can imagine that a custom TS transformer could do a similar substitution and then it wouldnt benefit from the fix.

Andarist:
> @Andarist: I can put up a PR for this that introduces parens in far fewer places, but if you'd rather do it

Usually, I prefer to do changes on my own. It's always a good learning experience 😉

jakebailey (approved):
After the most recent changes, the output looks good to me.

sandersn:
@jakebailey Do you want to wait for @rbuckton to sign off too, or is it good to go now?

jakebailey:
As far as I can tell, this PR matches what Ron suggested ("fix substituteConstantValue to avoid returning negative numeric literals and instead return parenthesized, prefix unary wrapped numeric literals"), so, I'd say so.

PR #55068, opened 2023-07-18
Skip @link at start of jsdoc types

### LINKED ISSUE #54650: @link tag inside a JSDoc type expression doesn't show parse error in TS
# Bug Report

### 🔎 Search Terms

tsdoc, `@throws`, `@link`, tags

This changed between versions 4.9.5 and 5.0.4

### ⏯ Playground Link

[Playground link with relevant code](https://www.typescriptlang.org/play?ts=5.0.4#code/MYGwhgzhAEBiD28CiAnF8XQKYA8AuWAdgCYyrqYDeAvgLABQokMAQmCuRtvkadJ1Tr0GAegBUYhtDHQAClhQAzDAFsYAIxABXdeoUA6KTKPQAAngAW6AO4xKpkAEtCAaziIB1aI8XRl8aAswAAdgoghDemkzSxs7B2c3Ng40DC8faHV2QJCwwgijEQZFLUJgPEd4QkztXQUACgBKaBogA)

### 💻 Code

```ts
class FooError extends Error {}
class BarError extends Error {}

/**
 * Performs blubber.
 * 
 * @throws {@link FooError} if foo happens.
 * @throws {@link BarError} if bar happens.
 */
function blubber() {}
```

### 🙁 Actual behavior

The `@link` tags after the `@throw` tags are parsed incorrectly, leading to this garbled tooltip:

image

### 🙂 Expected behavior

The `@link` tags after the `@throw` tags are resolved correctly.

sandersn:
Fixes #54650

Only the code in parseJSDocTypeExpression changes; the other diffs are moves of functions to outer scopes.

### DIFF
--- a/src/compiler/parser.ts
+++ b/src/compiler/parser.ts
@@ -8620,15 +8620,17 @@ namespace Parser {
             let pos = getNodePos();
             const hasBrace = (mayOmitBraces ? parseOptional : parseExpected)(SyntaxKind.OpenBraceToken);
             if (token() === SyntaxKind.AtToken) {
-                    // skip @link since it's a common mistake
-                    if(tokenIsIdentifierOrKeyword(nextTokenJSDoc())) {
-                        const kind = scanner.getTokenValue();
-                        if (isJSDocLinkTag(kind)) {
-                            nextTokenJSDoc();
-                            skipWhitespace();
-                        }
-                        pos = getNodePos();
+                // skip @link since it's a common mistake -- other tags will still error
+                const atStart = scanner.getTokenStart()
+                const atEnd = scanner.getTokenEnd()
+                if (tokenIsIdentifierOrKeyword(nextTokenJSDoc())) {
+                    if (!isJSDocLinkTag(scanner.getTokenValue())) {
+                        parseErrorAt(atStart, atEnd, Diagnostics.Type_expected);
                     }
+                    nextTokenJSDoc();
+                    skipWhitespace();
+                    pos = getNodePos();
+                }
             }
             const type = doInsideOfContext(NodeFlags.JSDoc, parseJSDocType);
             if (!mayOmitBraces || hasBrace) {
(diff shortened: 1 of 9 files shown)


src/compiler/parser.ts
```diff
@@ -8617,8 +8617,19 @@ namespace Parser {
 
         // Parses out a JSDoc type expression.
         export function parseJSDocTypeExpression(mayOmitBraces?: boolean): JSDocTypeExpression {
-            const pos = getNodePos();
+            let pos = getNodePos();
             const hasBrace = (mayOmitBraces ? parseOptional : parseExpected)(SyntaxKind.OpenBraceToken);
+            if (token() === SyntaxKind.AtToken) {
+                    // skip @link since it's a common mistake
+                    if(tokenIsIdentifierOrKeyword(nextTokenJSDoc())) {
```
jakebailey, line 8624:
```suggestion
                    if (tokenIsIdentifierOrKeyword(nextTokenJSDoc())) {
```

"dprint when"

src/compiler/parser.ts
```diff
@@ -8617,8 +8617,19 @@ namespace Parser {
 
         // Parses out a JSDoc type expression.
         export function parseJSDocTypeExpression(mayOmitBraces?: boolean): JSDocTypeExpression {
-            const pos = getNodePos();
+            let pos = getNodePos();
             const hasBrace = (mayOmitBraces ? parseOptional : parseExpected)(SyntaxKind.OpenBraceToken);
+            if (token() === SyntaxKind.AtToken) {
+                    // skip @link since it's a common mistake
```
jakebailey, line 8623: I'm assuming this skips anything, not just `@link`?

src/compiler/parser.ts
sandersn: This is *after* I made an attempt at formatting. I'm getting pre-emptively lazy.

jakebailey: keep it up, that way everyone will say yes to formatting!

src/compiler/parser.ts
sandersn: Yes, it's parse failure have `@throws {@anything type}` -- you'll just see a different parse error than before because the `@` will be skipped. I should probably add a test to show that off, since it will apply to checked JS files too.

jakebailey: Yeah, I just saw the comment above and it didn't line up with the condition, so wanted to make sure it was intentional.

PR #55072, opened 2023-07-18
feat: add flip operator refactoring

### LINKED ISSUE #55071: Add flip operator refactoring
# Suggestion

## 🔍 Search Terms

flip operator, flip operators, refactoring

## ✅ Viability Checklist

* [ x This could be implemented without emitting different JS based on the types of the expressions
* [ x This isn't a runtime feature (e.g. library functionality, non-ECMAScript syntax with JavaScript output, new syntax sugar for JS, etc.)

## ⭐ Suggestion

Implement flip operator refactoring for swapping left and right operators and update operand if needed

## 📃 Motivating Example
Before:
```ts
const a = 1 < 2 
```
After
```ts
const a = 2 > 1
```

Ilanaya:
Resolves #55071

### DIFF
--- a/src/compiler/diagnosticMessages.json
+++ b/src/compiler/diagnosticMessages.json
@@ -7864,5 +7864,9 @@
     "'await using' statements cannot be used inside a class static block.": {
         "category": "Error",
         "code": 18054
+    },
+    "Flip operator": {
+        "category": "Message",
+        "code": 18055
     }
 }
--- a/src/services/_namespaces/ts.refactor.ts
+++ b/src/services/_namespaces/ts.refactor.ts
@@ -4,6 +4,7 @@ export * from "../refactorProvider";
 export * from "../refactors/convertExport";
 export * from "../refactors/convertImport";
 export * from "../refactors/extractType";
+export * from "../refactors/flipOperator";
 export * from "../refactors/helpers";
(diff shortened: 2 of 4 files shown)


maIIady:

zardoy (commented):
1. Asterisk asterisk token
2. Move tokens that stays the same to set for easier review

suica:
Different people may expect the flipping to work differently. Someone may expect `<` to be flipped as `>=`.
Could that be a problem?

maIIady:
> Different people may expect the flipping to work differently. Someone may expect `<` to be flipped as `>=`. Could that be a problem?

Flipping `x < y` to `y >= x` may change the expression evaluation result, so I think 99% of developers wouldn't expect this refactoring to work that way.

suica:
> > Different people may expect the flipping to work differently. Someone may expect `<` to be flipped as `>=`. Could that be a problem?
> 
> Flipping `x < y` to `y >= x` may change the expression evaluation result, so I think 99% of developers wouldn't expect this refactoring to work that way.

Sorry, I didn't notice that the operands are also exchanged. XD

sandersn:
I want #55071 be accepted before reviewing the code. Otherwise the design might significantly change what code is needed.

Issue #55087, opened 2023-07-20
Export default of type emits invalid code


EliLichtblau:
# Bug Report

### 🔎 Search Terms
Export default type
Export default type reference error

### 🕗 Version & Regression Information
This has always been the behavior

### ⏯ Playground Link
[Playground link](https://www.typescriptlang.org/play?target=99&moduleResolution=99&module=199&ts=5.1.6#code/JYWwDg9gTgLgBDAnmApnA3gZUQOxgQwA8BpYHAEwF84AzKCEOAIiVQGcBjKYMGJgKH4pCkWHHIoa+AK4AbeNjxFSFIA)

### 💻 Code

```ts
import type {SyntaxKind} from "typescript"

export default SyntaxKind
```
Emits ESM
```js

export default SyntaxKind
```
Emits CommonJs
```js
"use strict";
Object.defineProperty(exports, "__esModule", { value: true });
exports.default = SyntaxKind;
```

which causes a reference error.

### 🙁 Actual behavior

Emitted code errors at runtime with reference error

### 🙂 Expected behavior

If I am default exporting a type I'd like that to be dropped from the emitted code as to not error. 

Since this is old behavior I'm guessing this might have something to do with globals that I do not understand? Any clarification on the issue is also super appreciated.

EliLichtblau:
[Notably this errors ](https://www.typescriptlang.org/play?moduleResolution=99&target=99&module=199&ts=5.1.6#code/JYWwDg9gTgLgBDAnmApnA3gZUQOxgQwA8BpYHAEwF84AzKCEOAIiVQGcBjKYMGJgKH4pCkWHHIoa+AK4AbeFlwESZKkA)

fatcerberus:
`import type` doesn't mean what you think it does: https://github.com/microsoft/TypeScript/issues/54990#issuecomment-1634508462

If there's a bug here, it's that the `export default` isn't being flagged as an error at compile time.

EliLichtblau:
Thanks for the import type issue! I also agree that it should fail at compile time! 

I assumed it wasn't emitted at all - not sure if this is useful to the decision making process but essentially what "bit me" from this behavior was.

Old codebase didn't properly annotate type imports and exports (as well importEqualsDeclaration use).
(Sorry if annotate is the wrong word I should mean isTypeOnly=true on the import/export specifiers)
I wrote a script to fix these cases.
A file that used to be a module now only exported types so webpack dropped it - (at least this is my understanding).
[...]

PR #55097, opened 2023-07-21
Error on export default of type: issue https://github.com/microsoft/TypeScript/issues/55087

EliLichtblau:
Fixes: [#55087](https://github.com/microsoft/TypeScript/issues/55087)
This changes the behavior of TS to the specified behavior in [issue](https://github.com/microsoft/TypeScript/issues/55087).

Export default of types become errors - i.e
```ts
type foo = 3
export default foo
```
This is because if it does not error - typescript emits invalid code.


EliLichtblau:

EliLichtblau:
Fixes: [#55087](https://github.com/microsoft/TypeScript/issues/55087)

- trying to trigger link issue

fatcerberus:
FWIW I think it's okay to default-export types in general, as long as they *only* have a typespace meaning--the problem is when you default-export something that has a valuespace meaning (like a class) which has been imported via `import type`.

EliLichtblau:
If this is the only condition - I think you can just convert 
```ts
 if (!isIllegalExportDefaultInCJS && !(node.flags & NodeFlags.Ambient) && compilerOptions.verbatimModuleSyntax && getTypeOnlyAliasDeclaration(sym, SymbolFlags.Value)) {
```
to 
```ts
if (!(node.flags & NodeFlags.Ambient) && getTypeOnlyAliasDeclaration(sym, SymbolFlags.Value)) {
```

the reason I find this a little awkward is checking with ts-fork-typechecker but emitting with ts loader transpileOnly is pretty common and you can pass the typechecking stage and still emit invalid code as the following code:
```ts
ts.transpileModule(`
import {type Node} from "typescript"
export default Node
`, {})
```
will give you
```ts
export default Node
```
even though Node is an interface. I can understand why you (the TS team) would say this is improper use and shouldn't be supported though.

With that said, my take is: 
* There should be a flag which disables all export default of things (need a better word here sorry) imported in a typeOnly context as this can emit invalid code with transpileModule calls
* The check on export default of import types should not live behind the verbatim module syntax flag and should be default behavior 


But I will implement whatever you think is best

The best solution IMO is to not change compile time behavior at all and for emit to just work but based on your comments in the issue I am unsure if this is possible. I know less than little about the emit phase so I don't have any idea here.

andrewbranch (changes requested):
I think the expected behavior is

- an `export default` of a type-only import never emits anything
- it’s an error only in `isolatedModule` (ts1205)

andrewbranch:
Fixed the checker behavior, still need to fix the emit

andrewbranch:
Ok, ready.

tests/baselines/reference/exportDefault.js
```diff
@@ -41,7 +41,6 @@ exports.A = A;
 //// [b.js]
 "use strict";
 Object.defineProperty(exports, "__esModule", { value: true });
-exports.default = types;
```
jakebailey, line 44: Oops!

jakebailey (approved):
LGTM but should get the extended test run.

jakebailey:
@typescript-bot test top200
@typescript-bot run dt

tests/baselines/reference/exportDefault.js
andrewbranch: Yeppp

andrewbranch:
These errors from the top repos look possibly wrong, especially the playwright one. Need to take a closer look.

andrewbranch:
@typescript-bot test top200

jakebailey:
You can ignore that playwright one, that is currently happening on every PR and we just haven't looked at why.

andrewbranch:
Huh? The playwright one is clearly caused by this and fixed in the latest commit.

jakebailey:
Apologies, I mixed up the "P" libraries and confused playwright for puppeteer (which you can see on other PRs like https://github.com/microsoft/TypeScript/pull/56429#issuecomment-1814887414). Ignore me!

PR #54820, opened 2023-06-28
dprint the codebase

jakebailey:
Be aware that elements of this stack are generated (like my modules PR); I'm going to be (intermittently) force pushing to keep it up to date as effectively any change to `main` causes a conflict, so apologies on that front.

---

This PR adds a [dprint](https://dprint.dev/) config to our repo. dprint (which is what `deno fmt` is running) is "yet another" code formatter in the same vein as prettier and others, with the main benefits of:

- Fast enough to format `checker.ts`, even on save, without being annoying.
- Configurable enough to format in this repo's "unique" code style (specifically, brace position after `if`, `else`, `catch`, etc).

This also enables us to remove effectively all eslint-based style checks, which is both a performance and ergonomics improvement. With all of the redundant style-related lint rules gone, linting the repo is now 7% faster!

The best way to view this PR is to look at the individual commits. This diff is large and makes GitHub mad, so consider viewing it github.dev or cloning it locally to play around. You'll want to install the dprint VS Code extension, which is automatically recommended per `.vscode/extensions.json`, and update `settings.json` to enable it.

---

Outstanding TODOs:

- [x] Add CI stuff to verify that PRs are formatted.
- [ ] Figure out if we need some sort of bot action to format people's PRs.
- [ ] Figure out if we need to run dprint as part of a local test/lint run.
	- As far as I'm aware, `dprint check` prints a diff; a list of files or a silent mode would work a lot better here. But then again, people should just run `dprint` religiously, or better yet, on save. It's fast.
- [x] Drop dprint sorting of imports in favor of simple-import-sort
- [x] Make some final style decisions
	- [x] Decide on a line length. This PR uses an arbitrarily high value of 1000, but many on the team have expressed their desire to lower this to something that can actually fit vertically with two columns, e.g. 120.
		- Immediate feedback from everyone is to just keep this at 1000 and we can lower it later.
	- [x] Decide on an operator position. This PR uses "maintain" and allows them in either position. A majority of the code puts operators on the right like in:
		```ts
		if (
		    (someCondition && otherCondition) ||
		    somethingElse()
		)
		```
        But I personally find it much, much easier to read:
		```ts
		if (
		    (someCondition && otherCondition)
		    || somethingElse()
		)
		```
		- Immediate feedback is to stick with "maintain" to allow either form, and we can tweak it later.
	- [x] Decide whether or not we are formatting JSON files.
[...]

PR #55106, opened 2023-07-21
Reorganize eslint config rule ordering

jakebailey:
I've been wanting to do this to make #54820 easier to deal with; I have to bulk delete the formatting-related rules and ensure they're handled, but it's tricky when I have to fix merge conflicts in the file or accidentally miss ones that are in fact formatting-related.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -30,10 +30,33 @@
         "/coverage/**"
     ],
     "rules": {
-        "max-statements-per-line": ["error", { "max": 1 }],
-        "simple-import-sort/imports": "error",
-        "simple-import-sort/exports": "error",
+        // eslint
+        "dot-notation": "error",
+        "eqeqeq": "error",
+        "no-caller": "error",
+        "no-eval": "error",
+        "no-extra-bind": "error",
+        "no-new-func": "error",
+        "no-new-wrappers": "error",
+        "no-return-await": "error",
+        "no-restricted-globals": [
+            "error",
+            { "name": "setTimeout" },
+            { "name": "clearTimeout" },
+            { "name": "setInterval" },
+            { "name": "clearInterval" },
+            { "name": "setImmediate" },
+            { "name": "clearImmediate" }
+        ],
+        "no-template-curly-in-string": "error",
+        "no-throw-literal": "error",
+        "no-undef-init": "error",
+        "no-var": "error",
+        "object-shorthand": "error",
+        "prefer-const": "error",
+        "prefer-object-spread": "error",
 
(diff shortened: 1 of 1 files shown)


.eslintrc.json
```diff
@@ -30,10 +30,33 @@
         "/coverage/**"
     ],
     "rules": {
-        "max-statements-per-line": ["error", { "max": 1 }],
-        "simple-import-sort/imports": "error",
-        "simple-import-sort/exports": "error",
+        // eslint
+        "dot-notation": "error",
```
jakebailey, line 34: This one may actually be a formatting rule, but neither dprint nor prettier do conversion from `foo["bar"]` to `foo.bar` or vice-versa.

.eslintrc.json
```diff
@@ -77,65 +99,52 @@
...
-        "no-throw-literal": "error",
         "no-trailing-spaces": "error",
-        "no-undef-init": "error",
-        "no-var": "error",
-        "object-shorthand": "error",
-        "prefer-const": "error",
-        "prefer-object-spread": "error",
         "quote-props": ["error", "consistent-as-needed"],
         "space-in-parens": "error",
-        "unicode-bom": ["error", "never"]
```
jakebailey, line 138: Snuck this in, but, this is just wholesale removed. I apparently missed this one in my review of Josh's PR as this one is already enabled by default in eslint.

jakebailey:
Just to show that there is no change besides removing the BOM rule, I ran this silly snippet before and after and this is the diff:

```js
console.log(
  JSON.stringify(
    Object.fromEntries(
      Object.entries(
        require("jsonc-parser").parse(
          require("fs").readFileSync(".eslintrc.json", "utf8"),
        ).rules,
      ).sort(([a], [b]) => (a < b ? -1 : 1)),
    ),
    undefined,
    4,
  ),
);
```

```diff
diff --git a/./before.json b/./after.json
index eac790eeaa..1208f39f8a 100644
--- a/./before.json
+++ b/./after.json
@@ -225,9 +225,5 @@
     ],
     "simple-import-sort/exports": "error",
     "simple-import-sort/imports": "error",
-    "space-in-parens": "error",
-    "unicode-bom": [
-        "error",
-        "never"
-    ]
+    "space-in-parens": "error"
 }
```

.eslintrc.json
jakebailey: Ugh, I'm wrong; I misread the page and the default is "never", but it's not enabled by default, so I will restore this.

sandersn (approved):

Issue #55091, opened 2023-07-20
Enum value `1 / 0` incorrectly transformed if there is `Infinity` declared in scope


nicolo-ribaudo:
# Bug Report

### 🔎 Search Terms

infinity enum variable let shadow

### 🕗 Version & Regression Information

Every version I tested

### ⏯ Playground Link

[Playground link with relevant code](https://www.typescriptlang.org/play?#code/N4KABGA2CmAuYEkB2AzAlktsCeYC8YAzANwjhjRICuAtmAIJigQQAa+YAjGAPRgAM5AL5kIAYwD2SAM4SYAOkgSA5gAp681gEoQQoA)

### 💻 Code

```ts
{
  let Infinity = 3;

  enum A {
    X = 1 / 0
  }

  console.log(A.X)
}
```

### 🙁 Actual behavior

```js
{
    let Infinity = 3;
    let A;
    (function (A) {
        A[A["X"] = Infinity] = "X";
    })(A || (A = {}));
    console.log(A.X);
}
```

and logs `3`

### 🙂 Expected behavior

It should log the result of `1 / 0`, which is `Infinity`.

Somewhat related to https://github.com/microsoft/TypeScript/issues/54981.

nmain:
This also appears to apply to `NaN`?

```typescript
  enum A {
    X = 1 / 0,
    B = 0 / 0,
  }
```

nicolo-ribaudo:
Yes 👍

ravin00:
This confusion comes because you're trying to write the global infinity value which will lead to unexpected behaviour.

`let Infinity = 3;`
You're assigning the value 3 to infinity and it will no longer represent the special value. Insted it will refer to the number 3.

`enum A {
  X = 1 / 0
}`

1 / 0 normally show the infinity. But since you have declared the value 3 to the infinity. The x value is will be 3 no the expected infinity.

To get the expected output remove the redefinition of the infinity.
`enum A {
  X = 1 / 0
}

console.log(A.X);  // This will log Infinity
`


This will log correctly infinity as expected

PR #55018, opened 2023-07-14
Correctly error on enum member values referring to identifiers shadowing `Infinity` and `NaN`

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/54981

PR #55107, opened 2023-07-22
Don't emit enum members as evaluated `Infinity`/`NaN` when their values are shadowed

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/55091

### DIFF
--- a/src/compiler/transformers/ts.ts
+++ b/src/compiler/transformers/ts.ts
@@ -1889,7 +1889,23 @@ export function transformTypeScript(context: TransformationContext) {
     function transformEnumMemberDeclarationValue(member: EnumMember): Expression {
         const value = resolver.getConstantValue(member);
         if (value !== undefined) {
-            return typeof value === "string" ? factory.createStringLiteral(value) : factory.createNumericLiteral(value);
+            if (typeof value === "string") {
+                return factory.createStringLiteral(value);
+            }
+            if (Number.isNaN(value)) {
+                return resolver.isNameReferencingGlobalValueAtLocation("NaN", member)
+                    ? factory.createIdentifier("NaN")
+                    : factory.createBinaryExpression(factory.createNumericLiteral(0), SyntaxKind.SlashToken, factory.createNumericLiteral(0));
+            }
+            if (!isFinite(value)) {
+                const infinity = resolver.isNameReferencingGlobalValueAtLocation("Infinity", member)
+                    ? factory.createIdentifier("Infinity")
+                    : factory.createBinaryExpression(factory.createNumericLiteral(1), SyntaxKind.SlashToken, factory.createNumericLiteral(0));
+                return value < 0
+                    ? factory.createPrefixUnaryExpression(SyntaxKind.MinusToken, infinity)
+                    : infinity;
+            }
+            return factory.createNumericLiteral(value);
         }
         else {
             enableSubstitutionForNonQualifiedEnumMembers();
--- a/tests/baselines/reference/enumShadowedInfinityNaN2.js
+++ b/tests/baselines/reference/enumShadowedInfinityNaN2.js
@@ -1,0 +1,28 @@
+//// [tests/cases/conformance/enums/enumShadowedInfinityNaN2.ts] ////
+
+//// [enumShadowedInfinityNaN2.ts]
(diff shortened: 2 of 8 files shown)


src/compiler/transformers/ts.ts
```diff
@@ -1889,7 +1889,23 @@ export function transformTypeScript(context: TransformationContext) {
     function transformEnumMemberDeclarationValue(member: EnumMember): Expression {
         const value = resolver.getConstantValue(member);
         if (value !== undefined) {
-            return typeof value === "string" ? factory.createStringLiteral(value) : factory.createNumericLiteral(value);
+            if (typeof value === "string") {
+                return factory.createStringLiteral(value);
+            }
+            if (Number.isNaN(value)) {
```
Andarist, line 1895: alternatively, I could just not use the constant value in those cases and fallback to the other branch - that would likely work. But I feel that `NaN` and `Infinity` should not be passed to `factory.createNumericLiteral` so some changes to this branch of code here would have to be done anyway. I also like that those evaluated, yet shadowed, `Infinity` and `NaN` are normalized~ by the proposed changes.

fatcerberus:
Just for completeness’ sake, how does this interact with `const enum`?

Andarist:
> Just for completeness’ sake, how does this interact with const enum?

`const enum` members have to be finite. TS still inlines `Infinity` and `NaN` in the emit's output despite the fact that it errors on them. This won't fix it in those instances but since the input code is considered to be invalid, it's likely OK to ignore this bug in those emits.

src/compiler/transformers/ts.ts
```diff
@@ -1889,7 +1889,22 @@ export function transformTypeScript(context: TransformationContext) {
...
-            return typeof value === "string" ? factory.createStringLiteral(value) : factory.createNumericLiteral(value);
+            if (typeof value === "string") {
+                return factory.createStringLiteral(value);
+            }
+            if (Number.isNaN(value)) {
+                return resolver.isNameReferencingGlobalValueAtLocation("NaN", member)
+                    ? factory.createIdentifier("NaN")
+                    : factory.createBinaryExpression(factory.createNumericLiteral(0), SyntaxKind.SlashToken, factory.createNumericLiteral(0));
+            }
+            if (!isFinite(value)) {
```
DanielRosenwasser, line 1900: What do you think of this alternative way to write it?

```ts
if (!isFinite(value) {
    let result: Node = resolver.isNameReferencingGlobalValueAtLocation("Infinity", member) ?
        factory.createIdentifier("Infinity") :
        factory.createBinaryExpression(factory.createNumericLiteral(0), SyntaxKind.SlashToken, factory.createNumericLiteral(0));
    if (value < 0) {
        result = factory.createPrefixUnaryExpression(SyntaxKind.MinusToken, result);
    }
    return result;
}
```

DanielRosenwasser: I guess that would emit as `-(1/0)`

Andarist: > What do you think of this alternative way to write it?

That's what I started with :p
https://github.com/microsoft/TypeScript/pull/55107/commits/ed2cc5717d0f956d90592af943141449b4b6f05d

> I guess that would emit as -(1/0)?

and why I changed it to the current version 😉

tests/baselines/reference/enumShadowedInfinityNaN2.js
```diff
@@ -0,0 +1,28 @@
+//// [tests/cases/conformance/enums/enumShadowedInfinityNaN2.ts] ////
+
+//// [enumShadowedInfinityNaN2.ts]
+// repro https://github.com/microsoft/TypeScript/issues/55091
+
+let Infinity = 3;
```
sandersn, line 6: Are there any more convincing global names that could get shadowed than Infinity and NaN? It's a stretch for people to create local names `Infinity` or `NaN`.

Andarist: I think those are the only 2 that can be produced by the „constant evaluation”

> It's a stretch for people to create local names Infinity or NaN.

cant disagree :p the issue was labeles as a bug and not a wontfix, it’s similar to the merged: https://github.com/microsoft/TypeScript/pull/55018

Issue #55075, opened 2023-07-19
Declaration for object with method named "new" incorrectly emitted with construct signature


jcalz:
# Bug Report

### 🔎 Search Terms
declaration emit; method named "new"; construct signature

### 🕗 Version & Regression Information

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about declaration emit, construct signatures
(note: some of this actually changed between TS3.7 and TS3.8, but the behavior in TS3.7 still has the main problem)

### ⏯ Playground Link

[Playground link with relevant code](https://www.typescriptlang.org/play?ts=5.2.0-dev.20230718#code/KYDwDg9gTgLgBAYwgOwM7wIZwLxwN4BQccywA7gBQgBcJArgLYBGwUAlPnFMDHVMnBBwA1HACMcAL4FpoSLEQp0cJjnxE4AIlJlNVWskYt2nbr36CR4qTIJzo8JGkdrCxANrbymgLr76zKwceFw8fAJCohLS0gQA9HFwACoAygDMAHQAHMLxAFRw9goAJsAIADYY3IrOcBi0bsQk5P6GgewGRqwA3DK9RfClFVXANcpMDRrEOq1dHQHGvZL94A5wQ5XVTsoIk03NlDQLQZ3tS71weXEE8YmpmQDsALT5haslZZuj25h7TTNHNrGNinRZ9OzvQafEZjeATdT7Ly6WbtEHHKDnCHyKHDLZKRx-DxI3wo4Ggnrg4hXAhAA)

### 💻 Code

```ts
export const a = {
  new(x: number) { return x + 1 }
}
export const b = {
  "new"(x: number) { return x + 1 }
}
export const c = {
  ["new"](x: number) { return x + 1 }
}
```

### 🙁 Actual behavior

The .d.ts file emitted is:
```ts
export declare const a: {
    new(x: number): number;
};
export declare const b: {
    new(x: number): number;
};
export declare const c: {
    new(x: number): number;
};
```
a construct signature that the actual object doesn't have.  (For TS 3.7 and below, `b` and `c` are emitted as-is with quotes/brackets so there's no problem there.  But `a` is still wrong.) If you consume this declaration weird things happen.

### 🙂 Expected behavior

Probably for this case `new` should be emitted as either `"new"` or `["new"]` or something for all three cases:
```ts
export declare const a: {
    "new"(x: number): number;
};
```

[...]

PR #55109, opened 2023-07-22
Fixed issue with method called `new` being emitted as construct signature

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/55075


tests/baselines/reference/objectTypesIdentityWithConstructSignatures2.types
```diff
@@ -32,8 +32,8 @@ var a: { new(x: Date): string }
 >x : Date
 
 var b = { new(x: RegExp) { return ''; } }; // not a construct signature, function called new
->b : { new(x: RegExp): string; }
->{ new(x: RegExp) { return ''; } } : { new(x: RegExp): string; }
+>b : { "new"(x: RegExp): string; }
```
jakebailey, line 35: This doesn't seem right... aren't all of these cases changed explicitly construct signatures that shouldn't be quoted?

Andarist: This one definitely isnt wrong. Notice how this is inferred~ from an object literal and that cant even have a construct signature - it’s just a method called `new`

jakebailey: Agh, yeah. I think I skimmed this PR on my phone on the weekend and had it in my head that there was something going on.

jakebailey:
@typescript-bot test this
@typescript-bot test top100
@typescript-bot run dt

jakebailey:
@typescript-bot pack this

jakebailey (approved):
LGTM, though maybe @rbuckton may have some other trick that would be preferred over this particular fix.

Andarist:
@jakebailey could you rephrase? from what I can tell both types baselines and tooltips are changed (and fixed) by this change

jakebailey:
My question was "should this fix be in declaration emit", but I half convinced myself the answer was "no" because of tooltips and baselines.

Of course, the fix could be applied to both `declarations.ts` and `typeToTypeNode` or whatever that thing is called.

jakebailey:
On second thought, this doesn't feel right; I think I found a better fix via `getPropertyNameNodeForSymbol` / `createPropertyNameNodeForIdentifierOrLiteral`.

Would you rather I send a new PR for this or push to your branch? (I am keeping the credit for the tests, of course; I just cloned this PR and then swapped the fix.)

jakebailey:
I don't want to stomp on your branch, I'll send a different PR.

jakebailey:
See: #55750

PR #55127, opened 2023-07-24
Replace handle on requestId in cancelAnimationFrame.

fosemberg:
"requestId" is more understandable than "handle".
On mdn, it is the "request id" that is used, not the "handle":

> A long integer value, the request id, that uniquely identifies the entry in the callback list.

https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame#return_value

### DIFF
--- a/src/lib/dom.generated.d.ts
+++ b/src/lib/dom.generated.d.ts
@@ -2511,7 +2511,7 @@ declare var AnimationEvent: {
 };
 
 interface AnimationFrameProvider {
-    cancelAnimationFrame(handle: number): void;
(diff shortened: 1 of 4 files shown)


fosemberg:

fosemberg:
https://github.com/microsoft/TypeScript/issues/55128

jakebailey:
The bot task failed to close this properly (my oops), but these files are generated from https://github.com/microsoft/TypeScript-DOM-lib-generator and so can't be modified directly here. Any change like this has to be discussed on that repo and sent to that repo.

PR #55130, opened 2023-07-24
Distribute index over mapped types when normalizing types

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/54892

### DIFF
--- a/src/compiler/types.ts
+++ b/src/compiler/types.ts
@@ -6624,6 +6624,7 @@ export interface IndexedAccessType extends InstantiableType {
     constraint?: Type;
     simplifiedForReading?: Type;
     simplifiedForWriting?: Type;
+    simplifiedForReadingD?: Type;
 }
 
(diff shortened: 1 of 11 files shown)


src/compiler/types.ts
```diff
@@ -6555,6 +6555,7 @@ export interface IndexedAccessType extends InstantiableType {
     constraint?: Type;
     simplifiedForReading?: Type;
     simplifiedForWriting?: Type;
+    simplifiedForReadingD?: Type;
```
weswigham, line 6558: Maybe `D` -> `Distributive` here.

weswigham (changes requested):
Hm, not a huge fan of a conditional check like this in `getSimplifiedType` - this may be a case where this should basically be a new relationship rule, since it's both only being done inside relationship checking, and only being done on the source side type. All that seems to indicate this equivalency would be better recognized as a new branch when comparing types. We might need to look at the pre-simplified types to make it happen, though.

PR #54693, opened 2023-06-18
Updated typescript-eslint to v6

JoshuaKGoldberg:
Applies the following strategy for updating the linter:

1. Bumping the version to the latest beta _(technically `rc`, but there's no meaningful difference here)_
2. Extended from our two recommended starting non-type-checked configs: [`recommended`](https://v6--typescript-eslint.netlify.app/linting/configs#recommended) and [`stylistic`](https://v6--typescript-eslint.netlify.app/linting/configs#stylistic)
3. Removed any existing rule configurations that are redundant when extending those configs
4. Disabled any rules with existing violations that seemed intentional, under a _"not applicable here"_ comment
5. Disabled any rules with existing violations that I wasn't sure the intentionality behind, under a _"Todo: do we want these?"_ comment
6. Manually fixed up remaining lint complaints

Explanations for the changes are inline. I realize the mild hubris on my end of posting lint recommendations on the TypeScript codebase itself - seeking input as to whether we're going the wrong way on any of them. 😄 

For the _Todo_ rules, I'm interested in your opinions on whether you'd want each of them or not in the TypeScript codebase.

Fixes #54692

PR #49119, opened 2022-05-15
Improve intersection reduction and CFA for truthy, equality, and typeof checks

ahejlsberg:
This PR introduces a number of changes affecting control flow analysis of truthy, equality, and typeof checks involving type `unknown` and unconstrained type variables in `--strictNullChecks` mode. Key to these changes is the fact that `{}`, the empty object type literal, is a supertype of all types except `null` and `undefined`. Thus, `{} | null | undefined` is effectively equivalent to `unknown`, and for an arbitrary type `T`, the intersection `T & {}` represents the non-nullable form of `T`.

The PR introduces the following new behaviors:

* An unconstrained type parameter is no longer assignable to `{}`.
* The predefined type `NonNullable<T>` is now an alias for `T & {}`.
* An intersection `T & {}`, where `T` is non-generic and not `null` or `undefined`, reduces to just `T` (`null & {}` and `undefined & {}` already reduce to `never`). See below for exceptions for `string & {}`, `number & {}`, and `bigint & {}`.
* An intersection `undefined & void` reduces to just `undefined`.
* Any type is assignable to a union type that contains `{}`, `null`, and `undefined`.
* In control flow analysis, `unknown` behaves similarly to the union type `{} | null | undefined`.
* In control flow analysis of truthy checks, generic types are intersected with `{}` in the true branch.
* In control flow analysis of equality comparisons with `null` or `undefined`, generic types are intersected with `{}`, `{} | null`, or `{} | undefined` in the false branch.
* In control flow analysis of `typeof x === "object"` expressions, generic types are intersected with `object` and/or `null` in the true branch (`typeof` checks for other types already produce intersections in a similar manner).

Some examples:

```ts
type T1 = {} & string;  // string
type T2 = {} & 'a';  // 'a'
type T3 = {} & object;  // object
type T4 = {} & { x: number };  // { x: number }
type T5 = {} & null;  // never
type T6 = {} & undefined;  // never
type T7 = undefined & void;  // undefined

function f1(u: unknown) {
    let x1: {} = u;  // Error
    let x2: {} | null | undefined = u;  // Ok
    let x3: {} | { x: string } | null | undefined = u;  // Ok
}

function f2(x: unknown) {
    if (x) {
        x;  // {}
    }
    else {
        x;  // unknown
    }
}

function f3<T>(x: T) {
    if (x) {
        x;  // T & {}
    }
    else {
        x;  // T
    }
}

function f4(x: unknown) {
    if (x !== undefined) {
        x;  // {} | null
    }
[...]

PR #55133, opened 2023-07-24
Enable @typescript-eslint/ban-types

jakebailey:
Follow-up to #54693.

All of the things it catches are harmless, realistically, though I agree that `as Function` was the wrong cast as that'd let us accidentally `new` it or something.

I definitely disagree with the default of banning `{}`; that's a very useful type (especially post #49119, a number I have memorized at this point). Annoyingly, violating the rule that bans `{}` or `Object` says to use `NonNullable<unknown>` which is exactly `{}`.

The fact that `Symbol` is mistaken as the global `Symbol` I believe is a bug in ts-eslint, ~~though I don't think it's reported so I'll do that shortly.~~ https://github.com/typescript-eslint/typescript-eslint/issues/7306

Disabling those two leaves the rule to only ban the uppercase primitive wrapper types (good) and `Object` (probably good), so it seems like this rule is fine to keep enabled.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -84,7 +84,9 @@
             {
                 "extendDefaults": true,
                 "types": {
-                    "Symbol": false, // This is theoretically good, but ts-eslint appears to mistake our declaration of Symbol for the global Symbol type.
+                    // This is theoretically good, but ts-eslint appears to mistake our declaration of Symbol for the global Symbol type.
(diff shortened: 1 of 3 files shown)


fatcerberus:
> `as Function` was the wrong cast as that'd let us accidentally `new` it or something

why isn’t `Function` treated as a top type similar to `(...args: never) => unknown`?

sheetalkamat (approved):

PR #55136, opened 2023-07-24
Triage and fix some eslint rules

jakebailey:
Follow-up to #54693.

This PR triages the rules that should "obviously" be disabled or fixed.

This caught a case where I had not noticed a stray `debugger` line in a PR.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -58,6 +58,7 @@
         "unicode-bom": ["error", "never"],
 
         // Enabled in eslint:recommended, but not applicable here
+        "no-extra-boolean-cast": "off",
         "no-case-declarations": "off",
         "no-cond-assign": "off",
         "no-constant-condition": ["error", { "checkLoops": false }],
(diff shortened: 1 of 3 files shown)


.eslintrc.json
```diff
@@ -79,22 +86,16 @@
         "@typescript-eslint/no-namespace": "off",
         "@typescript-eslint/no-non-null-asserted-optional-chain": "off",
         "@typescript-eslint/no-var-requires": "off",
+        "@typescript-eslint/no-empty-interface": "off",
```
jakebailey, line 89: This one deserves an explanation; while it's definitely bad to have empty interfaces, we seem to have quite a few cases which are "valid" in that they do things like rebinding another interface with new type variables (many in `types`), providing an interface that is augmented elsewhere (`SourceFile`), and so on.

sandersn (approved):

jakebailey:
After #55137, I've also added no-extra-boolean-cast, which similarly is too strict to work.

PR #55137, opened 2023-07-24
Enable eslint rule no-extra-boolean-cast

jakebailey:
Follow-up to #54693.

This is probably more for discussion; these extra `!!` are redundant and not obviously wrong, but, it's probably a fine cleanup.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -87,7 +87,6 @@
         "no-constant-condition": "off",
         "no-control-regex": "off",
         "no-debugger": "off",
-        "no-extra-boolean-cast": "off",
         "no-inner-declarations": "off",
         "no-useless-escape": "off",
         "prefer-rest-params": "off",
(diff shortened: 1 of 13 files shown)


jakebailey:
Aha, well, clearly I expected `--fix` to not cause any new errors, and yet here we are; seems like this lint rule isn't always accurate.

fatcerberus:
Looking at the diff I see a few places where it does stuff like `!!foo.bar ? foo.bar() : defVal` - I like the `!!` there because it makes it clear you’re checking for method existence and didn’t just accidentally forget the parens on a method call.

Also I thought TS itself had a check that flags stuff like `if (obj.method) {}` as an error (”Did you mean to call this function?”), likewise for promises without an `await`, so these might not always be redundant in a way that a linter can detect.

jakebailey:
I don't disagree; I'll probably just disable this rule instead.

We do have those checks, though they only apply in certain circumstances.

jakebailey:
I'm opting to disable this in #55136.

PR #55138, opened 2023-07-24
Enable eslint rule no-useless-escape

jakebailey:
Follow-up to #54693.

This actually caught a small number of oopsies where we wrote `\r\p` but meant `\r\np`


DanielRosenwasser (approved):
I hated reviewing this.

jakebailey:
Sending it was slightly better; at least I had a quick fix but no `--fix` because there are _two_ quick fixes, the "preserve behavior" button or the "actually I meant `\\p`" button.

sheetalkamat (approved):

src/testRunner/unittests/tsserver/cachingFileSystemInformation.ts
```diff
@@ -397,7 +397,7 @@ describe("unittests:: tsserver:: CachingFileSystemInformation:: tsserverProjectS
     "typescript"
   },
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "test": "echo "Error: no test specified" && exit 1"
```
jakebailey, line 400: Here's one I got wrong!

jakebailey: Fixed now. Didn't actually affect the test, apparently.

src/testRunner/unittests/base64.ts
```diff
@@ -10,7 +10,7 @@ describe("unittests:: base64", () => {
                 "日本語",
                 "🐱",
                 "\x00\x01",
-                "\t\n\r\\\"\'\u0062",
+                "\t\n\r\\\"'\u0062",
```
jakebailey, line 13: On second look, I'm not sure if this intended to write `'` or `\\'`. But, I'm not sure what "escaping" there is to check when roundtripping base64 anyway...

src/testRunner/unittests/convertToBase64.ts
```diff
@@ -13,7 +13,7 @@ describe("unittests:: convertToBase64", () => {
         });
 
         it("Converts escape sequences correctly", () => {
-            runTest("\t\n\r\\\"\'\u0062");
+            runTest("\t\n\r\\\"'\u0062");
```
jakebailey, line 16: Similarly here.

Issue #55129, opened 2023-07-24
keyof collapses the expression when used in a mapped type


Mad-Kat:
# Bug Report

### 🔎 Search Terms

- keyof union generic
- keyof union
- keyof

### 🕗 Version & Regression Information

- Since  4.1.5 (Playground test where mapping works)

### ⏯ Playground Link

[Playground link with relevant code](https://www.typescriptlang.org/play?ts=5.1.6#code/C4TwDgpgBAYgTgVwJbCgXgFBSgHygby22IDsBDAWwgC4oAiMsMAGwjoG4jioBjAe2Z84tOnAgATDlygBfInkLds5KiIBGZcuSlLeAoSJARmggO47ic7AunLKNekM0BzNp139Bwx3BdvpMpwYoJBQAEoQAM4IzMAAjAA8ACpQEAAewBAk4pEEKg6RwHBIJM64UCQIFGoQcOx6XrSFxaXlldW1sgB86AREANoA0hAgUCVQKWS5AAYAJPjDIP0A5PnLALoy1POLK55CGzLT67QIJADWJHymJBiBUMHg0BHRsQBMyakZWTl59k1FEplPDtGp1BoGKDNIFtKpg7q9c4jPgAMz62CGIzGJAmUCmUDmCxGKzWm22RKWy32cEOx1OFyuNzuGAwAHpWVCABZ8GLiKA1eiMFg0MSScp0DRaMjUIwma50cVOUo0JWuOiPUJJKLxRHItEvGLxBLwZDAHrs4CcpC5UxCc6RIjs7gAPQA-CynZFubz+dAGExWNRRQq8BLNOGZcYzCGfH5qKq2BroFrCm9egb3sbEChzazLdaoOI+FESMtULa4Oc8dkoABJZYUCp8VDRMRQUyc0bUAAijo5xDdLOwGCAA)

### 💻 Code

```ts
type Fruit =
  | {
      name: "apple";
      color: "red";
    }
  | {
      name: "banana";
      color: "yellow";
    }
  | {
      name: "orange";
      color: "orange";
    };

type Result1<T extends {name: string | number; color: string | number }> = {
  [Key in T as `${Key['name']}:${Key['color']}`]: unknown
}; 
type Result2<T extends {name: string | number; color: string | number }> = keyof {
  [Key in T as `${Key['name']}:${Key['color']}`]: unknown
}

// should be "apple:red" | "banana:yellow" | "orange:orange"
[...]

PR #55140, opened 2023-07-24
Defer index types on remapping mapped types

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/55129


src/compiler/checker.ts
```diff
@@ -38913,7 +38922,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         // Check if the index type is assignable to 'keyof T' for the object type.
         const objectType = (type as IndexedAccessType).objectType;
         const indexType = (type as IndexedAccessType).indexType;
-        if (isTypeAssignableTo(indexType, getIndexType(objectType, IndexFlags.None))) {
+        if (isTypeAssignableTo(indexType, getIndexType(objectType, IndexFlags.NoRemappingMappedTypeDeferral))) {
```
Andarist, line 38925: I kinda expected that I would have to pass this at more locations. Given the fact that this is the only call site using this... maybe I just should remove this `IndexFlags.NoRemappingMappedTypeDeferral` and perform the targeted simplification of `keyof RemappingMappedType` here to pass it to the `isTypeAssignableTo`.

weswigham: That would probably be better - `IndexFlags` originally only existed to support the `keyofStringsOnly` flag, and its' other uses have kind of become a convenience. With `keyofStringsOnly` likely to go away soonish, it'd be nice if we could stop having all these quiet variant index types.

src/compiler/checker.ts
```diff
@@ -1295,6 +1295,12 @@ const enum MappedTypeModifiers {
     ExcludeOptional = 1 << 3,
 }
 
+const enum MappedTypeNameTypeKind {
+    None = 0,
+    Filtering = 1 << 0,
+    Remapping = 1 << 1,
```
weswigham, line 1301: Why is this a flags enum? I can't see anywhere we actually combine multiple flags - in fact, it seems to me like a mapped type can only be one, given the implementation of `getMappedTypeNameTypeKind`.

weswigham (changes requested):

src/compiler/checker.ts
Andarist: I had some composite values here before but removed them at the end since I didn't actually use them anywhere. Refactored this now to be a non-flag enum

src/compiler/checker.ts
Andarist: pushed out this change

weswigham (approved):

weswigham:
@typescript-bot test this
@typescript-bot run dt
@typescript-bot test top100

sandersn:
@typescript-bot run dt

Issue #47693, opened 2022-02-02
Support 'go to def' and interacting with inlay hint components


mjbvz:
**TypeScript Version**: 4.6

**Search Terms**
- inlay hints
- `provideInlayHints`

---

VS Code recently updated our inlay hints to support tagging ranges within each hint with a location (file + position). This lets us perform operations such as go to definition on individual symbols within an inlay hint

You can see the VS Code side of this API here: https://github.com/microsoft/vscode/blob/516bef8deca4a3dc8544de536eb16d5e88ad8a83/src/vscode-dts/vscode.proposed.inlayHints.d.ts#L40

## Proposal
I propose that TS also starts returning symbol locations in its inlay hint response. This would let users hover and click on parts of an inlay hint to interact with it

Here's what the protocol change would look like:

```ts
interface InlayHintItem {
     // Today this is just a string
     text: string | readonly InlayHintDisplayPart[]:
  
     ....
}

interface InlayHintDisplayPart {
   // Displayed text in the ui
   text: string;

    // Definition of the symbol. 
    // TODO: Does this need to be an array or do we always have a single primary location we can use?
    span?: FileSpan;
}
```

---

As an example, consider the TS:

```ts
class A { }
class B { }

declare const Foo: A | [B];

const a = Foo;
```

With variable type inlay hints enabled, we currently see:

![Screen Shot 2022-02-01 at 6 02 10 PM](https://user-images.githubusercontent.com/12821956/152081958-5ac31bce-8909-4039-99f4-bc72b4970d32.png)

With the new proposal, we would want to return unique `InlayHintDisplayPart` elements for `A` and `B`. The response would look something like:

```
[Trace  - 02:03:33.592] <semantic> Response received: provideInlayHints (3969). Request took 2 ms. Success: true 
Result: [
    {
        "text": [
            {
                "text": ": "
            }, {
                "text": "A",
                "span": {
                    "file": "main.ts",
                    "start": { line: 1, column: 1 }, // made up location
                    "end": { line: 1, column: 99 },
                }
            }, {
                "text": " | ["
            }, {
                "text": "B",
                "span": {
                    "file": "main.ts",
                    "start": { line: 2, column: 1 },
                    "end": { line: 2, column: 99 },
                }
            }, {
                "text": "]"
            },
        ],
        "position": {
            "line": 6,
            "offset": 8
        },
        "kind": "Type",
        "whitespaceBefore": true
    }
]
```

/cc @jrieken

[...]

PR #55141, opened 2023-07-24
Interactive type inlay hints

MariaSolOs:
Second iteration of #47693


src/services/inlayHints.ts
```diff
@@ -223,13 +247,14 @@ export function provideInlayHints(context: InlayHintsContext): InlayHint[] {
             return;
         }
 
-        const typeDisplayString = printTypeInSingleLine(declarationType);
-        if (typeDisplayString) {
-            const isVariableNameMatchesType = preferences.includeInlayVariableTypeHintsWhenTypeMatchesName === false && equateStringsCaseInsensitive(decl.name.getText(), typeDisplayString);
+        const hint = typeToInlayHintParts(declarationType);
```
DanielRosenwasser, line 250: Let's call these `hints`

src/services/inlayHints.ts
```diff
@@ -421,6 +444,258 @@ export function provideInlayHints(context: InlayHintsContext): InlayHint[] {
...
+                case SyntaxKind.NeverKeyword:
+                case SyntaxKind.NumberKeyword:
+                case SyntaxKind.ObjectKeyword:
+                case SyntaxKind.StringKeyword:
+                case SyntaxKind.SymbolKeyword:
+                case SyntaxKind.UndefinedKeyword:
+                case SyntaxKind.UnknownKeyword:
+                case SyntaxKind.VoidKeyword:
+                case SyntaxKind.ThisType:
+                    parts.push({ text: tokenToString(node.kind)! });
```
DanielRosenwasser, line 484: Instead of this very-likely-to-not-be-updated list, why not just move this out of the switch and write the following?

```ts
const tokenString = tokenToString(node.kind);
if (tokenString !== undefined) {
     parts.push({ text: tokenString });
     return;
}
```

src/services/inlayHints.ts
```diff
@@ -421,6 +444,258 @@ export function provideInlayHints(context: InlayHintsContext): InlayHint[] {
...
+            return printTypeInSingleLine(type);
+        }
+
+        const flags = NodeBuilderFlags.IgnoreErrors | TypeFormatFlags.AllowUniqueESSymbolType | TypeFormatFlags.UseAliasDefinedOutsideCurrentScope;
+        const typeNode = checker.typeToTypeNode(type, /*enclosingDeclaration*/ undefined, flags);
+        Debug.assertIsDefined(typeNode, "should always get typenode");
+
+        const parts: InlayHintDisplayPart[] = [];
+        visitDisplayPart(typeNode);
+        function visitDisplayPart(node: Node) {
```
DanielRosenwasser, line 458: nit: you're not visiting display parts, you're visiting nodes to aggregate display parts

src/services/inlayHints.ts
```diff
@@ -421,6 +444,258 @@ export function provideInlayHints(context: InlayHintsContext): InlayHint[] {
...
+                case SyntaxKind.ConstructorType:
+                    const constructorType = node as ConstructorTypeNode;
+                    parts.push({ text: "new " });
+                    if (constructorType.typeParameters) {
+                        parts.push({ text: "<" });
+                        visitDisplayPartList(constructorType.typeParameters, ",");
+                        parts.push({ text: ">" });
+                    }
+                    // TODO: Parameters.
+                    parts.push({ text: " => " });
```
DanielRosenwasser, line 538: Instead of creating objects on the fly, I'm pretty sure we have display part functions that construct these. They have the benefit that they can include a display part kind as well if I recall correctly.

On that note, I think spaces are typically their own display part.

src/services/inlayHints.ts
MariaSolOs: Genius.

src/services/inlayHints.ts
MariaSolOs: Yeah I just didn't want to call it `visitor` to differentiate it from the one at the top of `provideInlayHints`. Suggestions of a better name?

src/services/inlayHints.ts
MariaSolOs: Yeah but those are for `SymbolDisplayPart`s, and I have `InlayHintDisplayPart`s here.

MariaSolOs:
@jakebailey should I rebase this PR to include the new formatting setup?

jakebailey:
As long as CI has run after the formatting was merged in main, it'll run, as the merge base is always the HEAD of main. In the checks: 

image

So, you should be fine unless it's failing for other reasons and not locally (so you need to pull to get the same state).

src/services/inlayHints.ts
```diff
@@ -423,6 +451,308 @@ export function provideInlayHints(context: InlayHintsContext): InlayHint[] {
         });
     }
 
+    function typeToInlayHintParts(type: Type): InlayHintDisplayPart[] | string {
+        if (!shouldUseInteractiveInlayHints(preferences)) {
+            return printTypeInSingleLine(type);
+        }
+
+        // TODO: No truncation?
```
MariaSolOs, line 459: Marked as a `TODO` since I'm not sure of how we want to deal with truncation here. A lazy solution would be to transform the display parts into a string and then check its length. If it's above the limit, then I return the truncated string instead.

tests/baselines/reference/inlayHintsInteractiveMultifile1.baseline
```diff
@@ -0,0 +1,128 @@
...
+{
+  "text": "",
+  "displayParts": [
+    {
+      "text": ": "
+    },
+    {
+      "text": "Promise",
+      "span": {
+        "start": -1,
```
MariaSolOs, line 12: @jakebailey here's an example of the baseline change I made.

src/services/inlayHints.ts
MariaSolOs: Resolving this as we've decided to not have a truncation limit for now and see what happens.

src/services/inlayHints.ts
jakebailey: `visitForDisplayParts`?

src/services/inlayHints.ts
```diff
@@ -422,6 +450,307 @@ export function provideInlayHints(context: InlayHintsContext): InlayHint[] {
...
+                    visitDisplayPart(propertySignature.name);
+                    if (propertySignature.questionToken) {
+                        parts.push({ text: "?" });
+                    }
+                    if (propertySignature.type) {
+                        parts.push({ text: ": " });
+                        visitDisplayPart(propertySignature.type);
+                    }
+                    break;
+            }
```
jakebailey, line 739: Are there any nodes left at the end of this switch case? If not, can we make it exhaustive such that we can `assertNever` or Debug fail if unexpected, such that we'll never forget to add here?

src/services/inlayHints.ts
```diff
@@ -422,6 +450,307 @@ export function provideInlayHints(context: InlayHintsContext): InlayHint[] {
...
+            return printTypeInSingleLine(type);
+        }
+
+        const flags = NodeBuilderFlags.IgnoreErrors | NodeBuilderFlags.AllowUniqueESSymbolType | NodeBuilderFlags.UseAliasDefinedOutsideCurrentScope;
+        const typeNode = checker.typeToTypeNode(type, /*enclosingDeclaration*/ undefined, flags);
+        Debug.assertIsDefined(typeNode, "should always get typenode");
+
+        const parts: InlayHintDisplayPart[] = [];
+        visitDisplayPart(typeNode);
+
```
jakebailey, line 464: Can you move `return parts` up to here? That way there's no unexpected code below the long function declarations.

src/services/inlayHints.ts
```diff
@@ -736,6 +737,8 @@ export function provideInlayHints(context: InlayHintsContext): InlayHint[] {
                         visitDisplayPart(propertySignature.type);
                     }
                     break;
+                default:
+                    Debug.fail(`Unhandled node kind: ${node.kind}`);
```
jakebailey, line 741: There's actually a helper for this one which will print better.
```suggestion
                    Debug.failBadSyntaxKind(node);
```

RyanCavanaugh:
@jakebailey can you review the latest commit? Thanks!

jakebailey:
The changes look good, I was just waiting on https://github.com/microsoft/TypeScript/pull/55141#discussion_r1329331792 as that was in my review comments from the last batch.

jakebailey (approved):
Of course I don't really mind the name either way.

jakebailey (approved):

Gawd71730:
Hints

Issue #54841, opened 2023-06-30
Spread operator in JSX causes loss of type information for inline functions defined outside the spread operator


kpervin:
# Bug Report
When defining conditional types on a JSX component, using a spread operator causes loss of type information on inline functions while removing the spread operator resolves the issue.

### 🔎 Search Terms
- "jsx spread"
- "TS7006"

### ⏯ Playground Link

[Playground link with relevant code](https://codesandbox.io/p/sandbox/typescript-error-sandbox-zj4633?file=%2Fsrc%2FExample.tsx%3A73%2C32)

### 💻 Code

Given the following types and JSX:

```tsx
/**
 * Either all properties are included, or none of them should be
 * @example
 * type Ex = AllOrNone<{ search: string; onSearchChange: (value: string) => void}>
 *
 * // Ok
 * const t1: Ex = {
 *   search: "hello",
 *   onSearchChange: (value) => console.log(value)
 * }
 *
 * // Error, needs onSearchChange
 * const t2: Ex = {
 *   search: "hello",
 * }
 *
 * // Error, needs search
 * const t3: Ex = {
 *   onSearchChange: (value) => console.log(value)
 * }
 */
export type AllOrNone<T> = T | { [K in keyof T]?: never };


export type OrderHistoryTableProps = {
  data: OrderRow[];
  noDataText?: string;
} & SortParams &
  FilterParams &
  SearchParams;

type SortParams = AllOrNone<{
  sorting: SortingState;
  onSortChange: (state: SortingState) => void;
}>;

type FilterParams = AllOrNone<{
  filter: ColumnFiltersState;
  onFilterChange: (state: ColumnFiltersState) => void;
}>;

type SearchParams = AllOrNone<{
  search: string;
  onSearchChange: (value: string) => void;
}>;

function OrderHistoryTable({
  data,
  noDataText = "No orders to display!",
  onFilterChange,
  onSortChange,
  sorting,
  filter,
  search,
  onSearchChange,
}: OrderHistoryTableProps): JSX.Element {
  ...
}

/* sorting, filter, and search are correct types */
<OrderHistoryTable
  data={rowData}
  {...{ sorting, filter, search }}
  onSortChange={onSortChange}
  onFilterChange={setFilter}
  onSearchChange={(value) => setSearch(value)} // TS7006: Parameter 'value' implicitly has an 'any' type.
/>
```

However, removing the spread operator resolves the issue:

```tsx
<OrderHistoryTable
  data={rowData}
[...]

PR #55144, opened 2023-07-25
Discriminate JSX attributes types using spread object literals

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/54841


tests/cases/compiler/contextuallyTypedJsxAttributeUsingSpreadObjLiteral.tsx
Andarist, line 1: This still won't work:
```ts
<OrderHistoryTable
  data={rowData}
  {...{ search: '' }}
  {...{ onSearchChange: value => {} }}
/>
```
but that's both bizarre and perhaps a different issue altogether (if it's an issue at all)

PR #55147, opened 2023-07-25
Update shared.ts

Harshitahusts:
introduced Rxjs for reducing the complexity of the code also install Rxjs before testing this by npm install rxjs in your command terminal

Fixes #


Harshitahusts:
tried my best to change and run ease on my system

Harshitahusts:
I have read the cla before the contribution and agree to all required agreements

jakebailey:
TypeScript is dependency-free; we're not going to depend on rxjs here, sorry.

PR #55152, opened 2023-07-25
Skip parent error when reporting excess property checks

RyanCavanaugh:
When we report an excess property error, the current error chain looks like this

```
    Argument of type '{ z: number; a: number; }' is not assignable to parameter of type 'T'.
      Object literal may only specify known properties, and 'z' does not exist in type 'T'.
```

As frequently noted by @fatcerberus, this gives the misleading impression that object types are sealed, since the way we printback a fresh object literal type is indistinguishable from a regular object type. It's also a waste of space since the top errorline prints back `T` when it's already in the EPC line, and the object literal source type is always anonymous (therefore long).

This PR skips the parent line when reporting EPC errors, so you don't see the "x is not assignable to y" part. The implementation is set up to handle the tricky case
```ts
declare function fn(a: { x: string }): void;
declare function fn(a: { y: string }): void;
fn({ z: 3, a: 3 });
```
where we want to print back two EPC errors:

```
// BEFORE
  Overload 1 of 2, '(a: { x: string; }): void', gave the following error.
    Argument of type '{ z: number; a: number; }' is not assignable to parameter of type '{ x: string; }'.
      Object literal may only specify known properties, and 'z' does not exist in type '{ x: string; }'.
  Overload 2 of 2, '(a: { y: string; }): void', gave the following error.
    Argument of type '{ z: number; a: number; }' is not assignable to parameter of type '{ y: string; }'.
      Object literal may only specify known properties, and 'z' does not exist in type '{ y: string; }'.

// AFTER
  Overload 1 of 2, '(a: { x: string; }): void', gave the following error.
    Object literal may only specify known properties, and 'z' does not exist in type '{ x: string; }'.
  Overload 2 of 2, '(a: { y: string; }): void', gave the following error.
    Object literal may only specify known properties, and 'z' does not exist in type '{ y: string; }'.
```


src/compiler/checker.ts
```diff
@@ -20354,6 +20354,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         let overrideNextErrorInfo = 0; // How many `reportRelationError` calls should be skipped in the elaboration pyramid
         let lastSkippedInfo: [Type, Type] | undefined;
         let incompatibleStack: DiagnosticAndArguments[] | undefined;
+        let skipParentCounter = 0;
```
RyanCavanaugh, line 20357: This currently never goes higher than 1, but if we ever had the notion of an EPC->EPC nested error, that would be how it could happen. Open to changing it to a boolean if wanted.

RyanCavanaugh:
@typescript-bot pack this

DanielRosenwasser:
Let's make sure this example doesn't regress and always reports on the discriminant.

```ts
export type Blah =
    | { type: "foo", abc: string }
    | { type: "bar", xyz: number, extra: any };

declare function thing(blah: Blah): void;

let foo = "foo";
thing({
    type: foo,
    abc: "hello!"
});

let bar = "bar";
thing({
    type: bar,
    xyz: 123,
    extra: 123,
});

File: tests/baselines/reference/arrayCast.errors.txt [hunk 2]
```diff
@@ -9,8 +8,7 @@ arrayCast.ts(3,23): error TS2352: Conversion of type '{ foo: string; }[]' to typ
     <{ id: number; }[]>[{ foo: "s" }];
                           ~~~
 !!! error TS2352: Conversion of type '{ foo: string; }[]' to type '{ id: number; }[]' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
-!!! error TS2352:   Type '{ foo: string; }' is not comparable to type '{ id: number; }'.
-!!! error TS2352:     Object literal may only specify known properties, and 'foo' does not exist in type '{ id: number; }'.
+!!! error TS2352:   Object literal may only specify known properties, and 'foo' does not exist in type '{ id: number; }'.
```
DanielRosenwasser, line 11: This error should never have been issued in the first place. It seems like a bug in our element-wise elaboration logic. Notice how [in the following example](https://www.typescriptlang.org/play?#code/FDA2FMBcAIHsGsCGBPaBeaBvaAzWsAuaAImOgF9pEBnLaASwBMiBGAJgGYKBuYAej7Qhw4QD9xEyVOlT+ggMKwAdgDdwAJ2r1lcHNEjIADuGgBybHkLRqkdfSUBzbhVP7Y+oyfMNm0dh2dyVwBbFGgAIxNEaGD6G0R4E0iAY0QAV2oTJXB6SAALDQ9jazScHHpk+nAlSFBUWDV1UERDWgB3XLz9Arh8jQA6aABJPXy46DaaBhrqyG0lRFAAGmhk5UaYPuhwAA9DdXBqLR1Id1M0pXglWDalV3LNSH65YQAFdVhjdQMzJldx2JHewOaZFLwWfBEGx2RyBVzhNIwA4ARzS9AOjFBBmK3iYrE4cOeYCgE0QiBgGAA2hCrKQKABdKi0bB4vycCiU+m8AQiXlicQvRSqDTHJS6MFmGlQ2zAwKc1ynCW43z+OX0kJhSJUGJxSAJJLgVIZLI5PrqCXUUrlSqzOpwRrNVoTTrdEywM2DEbdcaTWj2SCzeaLFZrYXfV3bPYHIEnM4XK43O64dE2Z48oQAFU8ktwkOsMthLgYtGuMDWwUMiHUiHCEDcSpZKoJLjTghEAHlwgArQ0wUC5DSLGJhZR26jGSo4VAJ27QfafDRzQ4rRBKTGmSyuRiwQ7QUuR3VY7PK-EBFvAIA) one of these is has a *missing* property error and the other is an *excess* property error:

```ts
let okay = { foo: "" } as { id: 123 };
//         ~~~~~~~~~~~~~~~~~~~~~~~~~~
// Conversion of type '{ foo: string; }' to type '{ id: 123; }' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
//   Property 'id' is missing in type '{ foo: string; }' but required in type '{ id: 123; }'.

let waat = [{ foo: "" }] as { id: 123 }[];
//            ~~~
// Conversion of type '{ foo: string; }[]' to type '{ id: 123; }[]' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
//   Type '{ foo: string; }' is not comparable to type '{ id: 123; }'.
//     Object literal may only specify known properties, and 'foo' does not exist in type '{ id: 123; }'.
```

DanielRosenwasser: Opened up https://github.com/microsoft/TypeScript/issues/55167

jakebailey:
@typescript-bot pack this

tests/baselines/reference/arrayCast.errors.txt
fatcerberus: Yeah, I wouldn’t ever expect an EPC error on a *type assertion*.

RobertSandiford:
Suggestions on wording from a user who was previously confused by it

Priorities:
1) Indicate that the error is a sanity check, not a type violation.
2) Indicate that the object literal can be assigned to that variable, albeit not directly.
3) Show the user how to assign it, if there is a preferred way.

`Warning: Property 'z' does not exist in type 'T'. Is this a mistake? To continue, cast the object literal to type 'T' or 'typeof var/Parameters<typeof func>[n]*', or save it to a new variable, before assignment.`

* Delete as appropriate. Do you have access to the variable/function name and arg pos?

It's a bit tricky to give clear instructions on how to assign it given the many scenarios. Simpler option might be:

`Warning: Property 'z' does not exist in type 'T'. Is this a mistake? Can't assign this object literal directly, to continue, use a typecast, assign to an intermediate variable, or use an IIFE.`

Really the important thing isnt how to assign but that the user understands that it is possible, to avoid misundersranding types.

RobertSandiford:
@RyanCavanaugh Yeah I generally agree. No issues with going ahead with something along those lines.

I based my suggestion on the type casting error which is similar in feeling, but with a suggestion on how to fix it.
`Conversion of type '{ c: number; }' to type 'O' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  Property 'a' is missing in type '{ c: number; }' but required in type 'O'.(2352)`

I think in the situation where you want to add properties, you are creating an object that conforms to some type standard, that is probably defined externally.

For simply object creation, you could extend this type with an intersection to creating your own expanded type.
Alternatively you could use `satisfies` and let TS infer the type of the new object. I was surprised to find that satisfies checks are subject to EPC too, I expect them to be treated like type cast, but this can be worked around by pushing the `satisfies` check into the next statement.

The other scenario that I can think of is abstract properties, but the way that this is handled is currently very permissive, as previously discussed, so no EPC checking here.

[Playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAKlC8UDeAoK6oEMBcUB2ArgLYBGEATigL4ooD0dUEAHmJngCZQDuAlsAAsovPMAoBnCAGNgvAPZ4UoSFACqCWFABkyNBhK5CpCtVpSF44FhxqNqDFlwBGADR70BqACZTKc3ksoEhI7d0coVzDPHyoscVsGfDkrcUwAMwgXJKtMDgA3XkkOWkSCSShU2XE03ghxPwsrKSlQhxtIh2ioRIpyOUpYysKautgzRqgOLkR7DHa3TtwYlCmKzCqR+JggA)

So the only scenario where EPC would come into play here is the `satisfies` approach.

Just thought I would lay that out incase it helps anyone.

src/compiler/checker.ts
weswigham, line 20357: This needs to be captured in `captureErrorCalculationState` and reset in `resetErrorInfo` so it's set/unset when a speculative relationship elaboration pyramid is discarded. AFAIK, you'd need to do something like trigger EPC inside a conditional type or indexed access type where we end up using the constraint check rather than the normal check for validity for it to matter (and the bug would look something like us discarding elaboration layers for non-EPC errors). Definitely niche, but those functions are just there for capturing/restoring error pyramid state, so should also track this new variable.

Also, maybe a comment on how this differs from the nearby `overrideNextErrorInfo` number. (`overrideNextErrorInfo` skips the next N things in the pyramid, while when set this skips the N prior things in the pyramid.)

weswigham (changes requested):

fatcerberus:
> this gives the misleading impression that object types are sealed, since the way we printback a fresh object literal type is indistinguishable from a regular object type.

[Case in point.](https://github.com/microsoft/TypeScript/issues/55277)  In general, people see "X is not assignable to Y" and expect that to apply in all contexts.  That issue, incidentally, also highlights another way the current error is misleading: it says "**object literal** may only specify known properties", but many defect reports *still* involve object literals, just without the contextual typing required for EPC to work.

PR #55153, opened 2023-07-25
Convert issue templates to issue forms

### LINKED ISSUE #54996: [DOCS] Use issue forms instead of issue templates
# Suggestion

The current templates are the old version of the issue templates. We can convert the issue templates to issue forms for easier management and better issue creation.
Many projects and repositories have been using it, and they are beneficial. It appears there are only a few templates to convert to issue forms (if no new ones are added).

## 🔍 Search Terms

`issue forms`, `issue templates`

## ✅ Viability Checklist

## ⭐ Suggestion

Convert the issue templates to issue forms.

## 📃 Motivating Example

I'm not 100% sure this fits this issue, but this would improve the issue management and make it easier for maintainers to check the issues and for users to report and fill in the required fields.

## 💻 Use Cases

1. **What do you want to use this for?**

This will be used by many people who create issues and will make everything easier and better.

2. **What shortcomings exist with the current approaches?**

There are no required fields or use of advanced form settings, which makes it very limited when creating issues or creating new types of templates/forms.

3. **What workarounds are you using in the meantime?**

Issue templates which is the current way of creating issues (some templates are using issue forms now, though).

Panquesito7:
Things added/changed:

- Convert issue templates to issue forms
  - Fixes #54996


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,93 @@
+name: "Bug report"
+description: "Create a report to help us improve TypeScript"
+title: "[BUG] <description>"
```
jakebailey, line 3: I don't know if we should have `[BUG]` in the title of all bugs; can we just leave this blank?

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,93 @@
...
+        "What search terms did you use when trying to find an existing bug report?
+
+        List them here so people in the future can find this one more easily."
+    validations:
+      required: true
+  - type: textarea
+    id: version_info
+    attributes:
+      label: "🕗 Version & Regression Information"
+      description: "
```
jakebailey, line 22: Can this just be a `|` string rather than having to quote things?

Similarly for all other multi-line strings like this.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,93 @@
...
+
+        Note: The TypeScript Playground can be used to try older versions of TypeScript.
+
+        Please keep and fill in the line that best applies.
+        "
+      value: |
+        - This is a crash
+        - This changed between versions ______ and _______
+        - This is the behavior in every version I tried, and I reviewed the FAQ for entries about _________
+        - I was unable to test this on prior versions because _______
```
jakebailey, line 42: Is there a better way to do this that uses checkboxes? I guess maybe you can't have checkboxes with extra info...

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,93 @@
...
+        //  - The code sample is short. Nearly all TypeScript bugs can be demonstrated in 20-30 lines of code!
+        //  - It doesn't use external libraries. These are often issues with the type definitions rather than TypeScript bugs.
+        //  - The incorrectness of the behavior is readily apparent from reading the sample.
+        // Reports are slower to investigate if:
+        //  - We have to pare too much extraneous code.
+        //  - We have to clone a large repo and validate that the problem isn't elsewhere.
+        //  - The sample is confusing or doesn't clearly demonstrate what's wrong.
+      render: TypeScript
+    validations:
+      required: false
```
jakebailey, line 79: Probably this should be required.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,93 @@
...
+        "
+      value: |
+        - This is a crash
+        - This changed between versions ______ and _______
+        - This is the behavior in every version I tried, and I reviewed the FAQ for entries about _________
+        - I was unable to test this on prior versions because _______
+      render: Markdown
+    validations:
+      required: true
+  - type: input
```
jakebailey, line 46: I'm curious how this will end up rendering; these links are notoriously long, so hopefully GH can do better here in the final output.

.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -0,0 +1,70 @@
+name: "Feature request"
+description: "Suggest an idea"
+title: "[FEATURE] <description>"
```
jakebailey, line 3: Similarly here, I don't think we need to have a convention for titles.

.github/ISSUE_TEMPLATE/lib_change.yml
```diff
@@ -0,0 +1,57 @@
+name: "Library change"
+description: "Fix or improve issues with built-in type definitions like `lib.dom.d.ts`, `lib.es6.d.ts`, etc."
+title: "[LIB CHANGE] <description>"
```
jakebailey, line 3: Also here; I don't think we have a convention for this.

.github/ISSUE_TEMPLATE/lib_change.yml
```diff
@@ -0,0 +1,57 @@
...
+          You can diagnose further by running `tsc` with `--listFilesOnly` or `--showConfig`.
+          Conversely, if you are seeing built-in methods you expect to *not* see, check your 'lib' setting
+
+          or review your dependencies for lib/reference directives that might be polluting
+          your global scope. This is common when using the 'node' type library. See https://github.com/microsoft/TypeScript/issues/40184
+
+          Replace the text below with the proper data.
+      value: "My compilation *target* is `ES2015` and my *lib* is `the default`."
+      render: Markdown
+  - type: input
```
jakebailey, line 36: Probably this should be a textarea; it may be the case that an issue involves multiple APIs.

.github/ISSUE_TEMPLATE/lib_change.yml
```diff
@@ -0,0 +1,57 @@
...
+          Try setting `lib: "es2020"` and checking whether the type you want is present.
+
+          You can diagnose further by running `tsc` with `--listFilesOnly` or `--showConfig`.
+          Conversely, if you are seeing built-in methods you expect to *not* see, check your 'lib' setting
+
+          or review your dependencies for lib/reference directives that might be polluting
+          your global scope. This is common when using the 'node' type library. See https://github.com/microsoft/TypeScript/issues/40184
+
+          Replace the text below with the proper data.
+      value: "My compilation *target* is `ES2015` and my *lib* is `the default`."
```
jakebailey, line 34: Nit, but maybe this should be two inputs or something; I thikn what we're asking for is for people to say what lib/target?

.github/ISSUE_TEMPLATE/lib_change.yml
```diff
@@ -0,0 +1,57 @@
...
+      required: true
+  - type: textarea
+    id: sample_code
+    attributes:
+      label: "Sample Code"
+      description: "What's some code using this that should work, but doesn't?"
+      render: TypeScript
+    validations:
+      required: true
+  - type: input
```
jakebailey, line 51: Similarly here, I think this could be a textarea.

jakebailey (commented):
One thing I'm wondering generally is if `textarea` with `display: TypeScript` will be "too strict" for some boxes; I know that people may want to file issues with multiple code samples, or file issues that don't quite fit the template but are still bugs, for example.

Really hard to tell without a preview of how these render 🙁

.github/ISSUE_TEMPLATE/bug_report.yml
Panquesito7: I'm not sure this would work as a checkbox, as this requires input from the user.

.github/ISSUE_TEMPLATE/bug_report.yml
Panquesito7: Not all bug reports are related to that, so not sure (like the issue about the issue forms for example) about that. If wanted, I can still mark it as required, though.

.github/ISSUE_TEMPLATE/lib_change.yml
fatcerberus: He's saying there should be inputs for the specific value of `target` and `lib` instead of just asking the user to "replace the text below" a la old-fashioned templates

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,90 @@
...
+    validations:
+      required: true
+  - type: textarea
+    id: version_info
+    attributes:
+      label: "🕗 Version & Regression Information"
+      description: |
+        When did you start seeing this bug occur?
+
+        \"Bugs\" that have existed in TS for a long time are very likely to be FAQs; refer to
```
jakebailey, line 24: This doesn't need to be escaped anymore, as far as I'm aware; in fact, all of the escapes should be able to go away.

.github/ISSUE_TEMPLATE/lib_change.yml
jakebailey: Yes, I meant having two input boxes for each given that's the only two pieces of info we ask for here and they're required.

jakebailey:
I've gone and pushed the contents of this PR to a repo so we can preview it: https://github.com/jakebailey/ts-issue-template-testing/issues/new/choose

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,90 @@
...
+    validations:
+      required: false
+  - type: textarea
+    id: code
+    attributes:
+      label: 💻 Code
+      description: |
+        Please post the relevant code sample here as well.
+        This code and the Playground code should be the same, do not use separate examples.
+      placeholder: |
```
jakebailey, line 65: Testing this one out in a real form, a placeholder doesn't look like it works here. The text shows up, but since it's just a placeholder, writing anything means that this goes away.

What this should be is part of the description.

https://github.com/jakebailey/ts-issue-template-testing/issues/new?assignees=&labels=&projects=&template=bug_report.yml

Panquesito7: If we use `value`, the text will be kept, thus allowing the user to modify it. I'll use that.

jakebailey: But this isn't text that needs to be updated; it's a list of requirements/suggestions for what the box will contain. I don't think there's any reason for it to be a part of the input itself.

Panquesito7: Hmm, alright, sorry for the confusion 😅 Will add it to the description.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,92 @@
...
+
+        ```ts
+        // We can quickly address your report if:
+        //  - The code sample is short. Nearly all TypeScript bugs can be demonstrated in 20-30 lines of code!
+        //  - It doesn't use external libraries. These are often issues with the type definitions rather than TypeScript bugs.
+        //  - The incorrectness of the behavior is readily apparent from reading the sample.
+        // Reports are slower to investigate if:
+        //  - We have to pare too much extraneous code.
+        //  - We have to clone a large repo and validate that the problem isn't elsewhere.
+        //  - The sample is confusing or doesn't clearly demonstrate what's wrong.
```
jakebailey, line 74: This can just be regular text; it doesn't need to be a code block. If you drop the `// ` prefix, it should render okay.

Right now, the text is very small: https://github.com/jakebailey/ts-issue-template-testing/issues/new?assignees=&labels=&projects=&template=bug_report.yml

.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -0,0 +1,68 @@
+name: "Feature request"
+description: "Suggest an idea"
+body:
+  - type: markdown
+    attributes:
+      value: "Please fill in each section completely. Thank you!"
+  - type: textarea
```
jakebailey, line 7: I just noticed that the issue template has this as an `input`; maybe that's better? Or, maybe the issue template should use a text area instead?

.github/ISSUE_TEMPLATE/lib_change.yml
```diff
@@ -0,0 +1,62 @@
...
+      description: "What's your compilation target (e.g.: `ES2015`)?"
+    validations:
+      required: true
+  - type: textarea
+    id: current_lib
+    attributes:
+      label: "⚙ Library"
+      description: "What's the current library you're using?"
+    validations:
+      required: true
```
jakebailey, line 40: These should both just be `input`, not textarea; the strings will be small.

.github/ISSUE_TEMPLATE/bug_report.yml
fatcerberus: Ironically it's bigger than the surrounding text 😅, but yeah, it doesn't look that great as a codeblock.

image

Panquesito7: I'll remove the code block and the `//`.

.github/ISSUE_TEMPLATE/feature_request.yml
jakebailey: bug_report.yml has this same field as `type: input`, while this one has it as `type: textarea`; I'm just trying to decide which one is better. I'm leaning towards both being `textarea`.

Panquesito7: Ah, yes, I noticed that. Yeah, I prefer that one, too.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,92 @@
...
+
+        Note: The TypeScript Playground can be used to try older versions of TypeScript.
+
+        Please keep and fill in the line that best applies.
+      value: |
+        - This is a crash
+        - This changed between versions ______ and _______
+        - This is the behavior in every version I tried, and I reviewed the FAQ for entries about _________
+        - I was unable to test this on prior versions because _______
+      render: Markdown
```
jakebailey, line 41: This line appears to cause problems; this apparently means it turns into a markdown code block in the issue. See https://github.com/jakebailey/ts-issue-template-testing/issues/1

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,92 @@
...
+
+          - The code sample is short. Nearly all TypeScript bugs can be demonstrated in 20-30 lines of code!
+          - It doesn't use external libraries. These are often issues with the type definitions rather than TypeScript bugs.
+          - The incorrectness of the behavior is readily apparent from reading the sample.
+
+        Reports are slower to investigate if:
+          - We have to pare too much extraneous code.
+          - We have to clone a large repo and validate that the problem isn't elsewhere.
+          - The sample is confusing or doesn't clearly demonstrate what's wrong.
+      render: Markdown
```
jakebailey, line 76: This one should also be deleted.

.github/ISSUE_TEMPLATE/bug_report.yml
jakebailey: Checking https://github.com/jakebailey/ts-issue-template-testing/issues/1, it is small in the form but gets pasted as a huge link here.

I wonder if GH has some special thing that can make links, but I doubt it.

.github/ISSUE_TEMPLATE/bug_report.yml
jakebailey: Hm, I might be wrong here, my second attempt was fine. https://github.com/jakebailey/ts-issue-template-testing/issues/4

Maybe I accidentally stuck bad backticks?

jakebailey: No, it's getting formatted as code. Let me make a change and see if it helps live.

jakebailey: Yeah, works after being removed https://github.com/jakebailey/ts-issue-template-testing/issues/5

jakebailey:
```suggestion
```

.github/ISSUE_TEMPLATE/bug_report.yml
jakebailey:
```suggestion
```

jakebailey (approved):

Panquesito7:
Sorry for not addressing the reviews before, I was busy. Thank you, though! 🙂

jakebailey:
All good, thanks for the PR!

fatcerberus:
Since this was merged, I've noticed several issues where people have written a summary of the issue in the `Search Terms` section instead of actual search terms.  So I checked out the form and I think I understand why:

image

An unhinted required multiline field immediately below the title 100% looks like a "describe your bug here" field.  Especially since the instructional text describing what to do is so small, it's easy to dismiss as boilerplate and skim past.

jakebailey:
Yeah, I've noticed it too. Maybe a placeholder would help. But, I'm not sure what to say if people are unwilling to read... 😢

fatcerberus:
> But, I'm not sure what to say if people are unwilling to read... 😢

Yeah.  IMO it would help if GitHub didn't render the interstitial text in small font and gray.  I myself am guilty of skimming past such text in forms because it's usually just boilerplate explaining what's already obvious from the shape and position of the fields what each is for.

PR #53193, opened 2023-03-10
Consistently use '...args' for diagnostic args

jakebailey:
I noticed this was inconsistent and somewhat annoying when I was working on the deprecation errors. I had to make a lot of changes to add a single arg, when the actual core infrastructure for diagnostics supports arbitrary arguments.

Along the way, this shows that we have accidentally been using escaped names in some error messages because some of these functions had their args typed as `any`.

But, no changes to any baselines.

Hopefully this isn't any slower given the bottommost level of this chain already uses variadic args.

Issue #498, opened 2014-08-21
Optimize rest arguments.


mbebenita:
The TypeScript compiler generates the following code for rest arguments:

```
function foo(x, y, z, ...rest) {
  return;
}

function foo(x, y, z) {
  var rest = [];
  for (var i = 0; i < arguments.length - 3; i++) {
    rest[i] = arguments[i + 3];
  }
  return;
}
```

Although `rest` is unused in the body of the function, JS engines can't easily dead code eliminate the allocation of the `rest` array and the subsequent for loop due to the possibility of Array.prototype being mutated:

```
Object.defineProperty(Array.prototype, 42, { set: function() { alert(this[41]); }});
```

It's much easier for the TypeScript compiler to avoid emitting this code when it is not necessary.

See more details:

https://bugzilla.mozilla.org/show_bug.cgi?id=1056446

begincalendar:
As asked by @andy-ms ([here](https://github.com/Microsoft/TypeScript/pull/19330#issuecomment-351089321)), what are the use-cases for having unused rest parameters?

RyanCavanaugh:
This doesn't really seem needed - the overload trick is sufficient, and most people are not downleveling rest args anymore.

PR #55181, opened 2023-07-28
Enable eslint rules prefer-rest-params and prefer-spread

jakebailey:
This is somewhat involved as our diagnostic handling was all `arguments` based.

I somewhat recently tried to make this consistent in #53193 but apparently didn't go all the way down to the bottom, which this PR necessarily has to do.

There are some places where I chose to just ignore things. I'm unsure if there are better ways to write things but most of those uses also already disabled a lint for using non-arrow functions.

It's possible that there's a perf factor here but I can't seem to observe one, and the callers to the diagnostics stuff already use spread.


jakebailey:

sandersn (approved):
Seems good as long as there's no performance difference.

jakebailey:
@typescript-bot new perf test this faster

jakebailey:
Now that I think harder, I wonder if this was written this way to work around #498, which no longer applies given our emit target is no longer ancient.

jakebailey:
> Now that I think harder, I wonder if this was written this way to work around #498, which no longer applies given our emit target is no longer ancient.

Aha, yes, I think I'm on the right track in that thinking: [Playground Link](https://www.typescriptlang.org/play?target=1#code/CYUwxgNghgTiAEYD2A7AzgF3gMyTAtlBgLIhppQDmIAXPFCgJ4DcAUK6JLAhowA4JiSYAFcIIAEpkkEERgCWqABJJM8ALz0mbTtDjxeA+ABF5VFKoVhS5Kgk0MW7EAA8+eLNhEowC1AZgoMBAACgALSzohUXEpNBk5RRQVTAAaeHwyCmo6U3NLeWssu3SAOnLYSjQ6RwBtAF0ASjoANyR5YDZXdxhPb18kgKDQiMwo4TFJaVk-ZMtm+DaO+ABvVngN+FGMUoxA4IBCENwCIhtskFKoPj4IRhDvUGx5FBBgdMqRTJQMNEbGtgAX2cbg8OH6syGwQATOFIvBopM4glZikMOlMrYciYzJQLJhCucSvBylcYFUakwGgslsBVutNttdvsQEcToQSMVqCFMRcyhVyX8AaxAUA)

```ts
export function trace(host: ModuleResolutionHost, message: DiagnosticMessage, ...args: any[]): void;
export function trace(host: ModuleResolutionHost): void {
    host.trace!(formatMessage.apply(undefined, arguments));
}

export function trace2(host: ModuleResolutionHost, message: DiagnosticMessage, ...args: any[]): void {
    host.trace!(formatMessage(message, ...args));
}
```

is in ES5

```ts
export function trace(host) {
    host.trace(formatMessage.apply(undefined, arguments));
}
export function trace2(host, message) {
    var args = [];
    for (var _i = 2; _i < arguments.length; _i++) {
        args[_i - 2] = arguments[_i];
    }
    host.trace(formatMessage.apply(void 0, __spreadArray([message], args, false)));
}
```

But post-modules, we're not emitting in ES5 anymore. This gives me more confidence in this change.

jakebailey:
And, perf is nice and clean.

Issue #54811, opened 2023-06-28
Missing computed properties on expando functions 


dragomirtitian:
# Bug Report

### 🔎 Search Terms

expando functions computed properties

### 🕗 Version & Regression Information

- This is the behavior in every version I tried, and I reviewed the FAQ 

### ⏯ Playground Link

[Playground link with relevant code](https://www.typescriptlang.org/play?ts=5.1.3#code/MYewdgzgLgBMMF4YCIDCyCwAoApgDwAcQAnWAMwFcxgoBLcGAEx2ABsYAKAShgG8BfbAHohTFqwB0ZECEQpMWZmwDayAELIAunIAM2Ja2XBtSPVmwWsQA)

### 💻 Code

```ts
const c = "C"
export function decl () {}
// decl.foo = ""
decl["B"] = 0
decl[c] = 0
```

### 🙁 Actual behavior

Properties "B"  and "C" are not present in the generated declarations unless `decl.foo` is uncommented. 

### 🙂 Expected behavior

Properties "B"  and "C" should be present in the generated declaration regardless of the presence of `foo`.

dragomirtitian:
Found this while trying to emulate the behavior in isolated declarations

aditya1:
Can I work on this one? I’m interviewing for a company, which wants me to contribute to some open source project in the form of PR? I’m looking to contribute for 2-4 hours.

WilczynskiT:
@aditya1 You don’t have to ask anyone ;) Just fork the repo, fix the bug, and then open a pull request

dragomirtitian:
> Can I work on this one? I’m interviewing for a company, which wants me to contribute to some open source project in the form of PR? I’m looking to contribute for 2-4 hours.

@aditya1  I would look in `isExpandoFunctionDeclaration` the check there seems different from what is happening in the declaration emitter at https://github.com/microsoft/TypeScript/blob/dd3e81a6752401f1656e2d02354bf35c3baac3ea/src/compiler/transformers/declarations.ts#L1528

PR #54726, opened 2023-06-21
Fixed expando functions with symbol-only properties

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/54220

PR #55183, opened 2023-07-28
Fixed declaration emit for expando properties on function declarations declared using element access expressions

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/54811


src/compiler/utilities.ts
```diff
@@ -10338,3 +10338,8 @@ export function getPropertyNameFromType(type: StringLiteralType | NumberLiteralT
     }
     return Debug.fail();
 }
+
+/** @internal */
+export function isExpandoPropertyDeclarationCandidate(declaration: Declaration | undefined): declaration is PropertyAccessExpression | ElementAccessExpression | BinaryExpression {
+    return !!declaration && !!(isPropertyAccessExpression(declaration) || isElementAccessExpression(declaration) || isBinaryExpression(declaration));
```
sandersn, line 10344: why is the second `!!` needed?

src/compiler/utilities.ts
```diff
@@ -10338,3 +10338,8 @@ export function getPropertyNameFromType(type: StringLiteralType | NumberLiteralT
     }
     return Debug.fail();
 }
+
+/** @internal */
+export function isExpandoPropertyDeclarationCandidate(declaration: Declaration | undefined): declaration is PropertyAccessExpression | ElementAccessExpression | BinaryExpression {
```
sandersn, line 10343:
```suggestion
export function isExpandoPropertyDeclaration(declaration: Declaration | undefined): declaration is PropertyAccessExpression | ElementAccessExpression | BinaryExpression {
```

unless it's already taken, in which case I'd prefer the name `isExpandoPropertyDeclarationLocation`

tests/cases/compiler/declarationEmitLateBoundAssignments2.ts
```diff
@@ -0,0 +1,19 @@
+// @strict: true
+// @declaration: true
+// @target: es6
+
+// repro from https://github.com/microsoft/TypeScript/issues/54811
+
+const c = "C"
+
+export function decl () {}
+decl["B"] = 'foo'
```
sandersn, line 10: what does the BinaryExpression case look like? If there's not a test for it here or somewhere else, please add one.

sandersn (changes requested):
Looks good: I had one question, one name change, and one test case request.

src/compiler/utilities.ts
Andarist: it's not 😅

tests/cases/compiler/declarationEmitLateBoundAssignments2.ts
Andarist: It's the `decl2` case. This is the baseline diff without the `isBinaryExpression` check:

<details>
<summary>baseline diff</summary>

```diff
diff --git a/tests/baselines/reference/declarationEmitLateBoundAssignments.js b/tests/baselines/reference/declarationEmitLateBoundAssignments.js
index b76db45754..763511ec0d 100644
--- a/tests/baselines/reference/declarationEmitLateBoundAssignments.js
+++ b/tests/baselines/reference/declarationEmitLateBoundAssignments.js
@@ -38,5 +38,4 @@ const a = foo[dashStrMem];
 export declare function foo(): void;
 export declare namespace foo {
     var bar: number;
-    var strMemName: string;
 }
diff --git a/tests/baselines/reference/declarationEmitLateBoundAssignments2.js b/tests/baselines/reference/declarationEmitLateBoundAssignments2.js
index 4ec90c2606..b2d3efc2a8 100644
--- a/tests/baselines/reference/declarationEmitLateBoundAssignments2.js
+++ b/tests/baselines/reference/declarationEmitLateBoundAssignments2.js
@@ -37,9 +37,6 @@ export declare namespace decl {
     var B: string;
 }
 export declare function decl2(): void;
-export declare namespace decl2 {
-    var C: number;
-}
 export declare const arrow: {
     (): void;
     B: string;
```
</details>

weswigham (changes requested):
Can we just add some test cases for expando element access properties with non-identifier names (numbers, strings with spaces, unique symbols)? This looks good, I just wanna make sure we don't try and emit `{ a b: string }` or something equally silly.

tests/baselines/reference/declarationEmitLateBoundAssignments2.js
```diff
@@ -0,0 +1,186 @@
...
+export declare function decl6(): void;
+export declare namespace decl6 { }
+export declare function decl7(): void;
+export declare namespace decl7 { }
+export declare function decl8(): void;
+export declare namespace decl8 { }
+export declare function decl9(): void;
+export declare namespace decl9 { }
+export declare function decl10(): void;
+export declare namespace decl10 { }
```
Andarist, line 146: those don't work because function declarations are emitted as namespaces and non-identifiers can't be exported from namespaces, see: https://github.com/microsoft/TypeScript/issues/55190

I could make an effort to avoid those empty namespaces. The issue itself stands though. I think it's a slight problem that function declarations and function expressions have such a subtle and silent (!) difference in their capabilities to emit expando members

Andarist:
> Can we just add some test cases for expando element access properties with non-identifier names (numbers, strings with spaces, unique symbols)? 

Nice, this helped me find the issue in `isStringNamed`. I didn't add any tests related to symbols because I'm pretty sure that symbols are already covered OK (I fixed myself a couple of issues related to expando properties with symbol names, such as https://github.com/microsoft/TypeScript/pull/54726 and https://github.com/microsoft/TypeScript/pull/55357 )

weswigham (approved):
Sorry for the delay on coming back to this PR~

PR #43882, opened 2021-04-29
More effective call arguments/prototype

sandersn:
Today, call resolution starts by expanding spread arguments with a tuple type like so:

```
f(string, [number, string], string) ==> f(string, number, string, string)
```

However, all other spread arguments are required to match a rest parameter. This PR adds two new expansions:

1. Spreads of tuple unions expand into a union of each tuple member:

```
f(string, [string] | [number, number], string) ==> (f, string, number | string, number | undefined, string)
```

2. Spreads of arrays expand into T | undefined for each member T, except for the last, which is T to match the required rest parameter:

```
declare function f(x,y,z,...rest)
f(string, string[], number) ==> f(string, string | undefined, string | undefined, string, number)
```

This is a prototype to see whether the idea works and discover what complications arise. The biggest assumption, which I still need to prove,  is that even though the expansions don't cover all permutations of calls, they will be correct if and only if all permutations are correct.

For example, the first example

```
f(string, [string] | [number, number], string)
```

really has two possible expansions

```
f(string, string, string)
f(string, number, number, string)
```

and `f(string, string | number, number | undefined, string)` is only an approximation.

Because it's a prototype, it's missing a lot of pieces:

- Array expansion isn't safe without strict null checks, because it essentially asserts that the array *always* has enough elements, instead of *possibly* having enough elements.
- Neither expansion should happen when the target parameter is a rest, since that's already allowed.
- Tuple union expansion only works with required and optional elements, but might possibly work with rest ones in trailing position.
- The existing tuple expansion has special index elaborations that need to replace or augment the current new array expansion errors.
- All of the above need a lot of tests.

Based on branch reform-arity-errors from #43855.

Fixes #42508

PR #55185, opened 2023-07-28
Allow calls ending with unions of tuples

### LINKED ISSUE #42508: Error when spreading a union of tuples in a call
# Bug Report

### 🔎 Search Terms

destructure array typescript union

I used the playground to test different versions and it looks like this has always occured.

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about "Type Systems behavior", "Type Guards", "Functions". 
 
### ⏯ Playground Link

[Playground link with relevant code](https://www.typescriptlang.org/play?ts=4.2.0-dev.20210126#code/MYewdgzgLgBLC8MDaAWANAJgAwF0YEMIZRIoAoE6ODGRVNGbBgZlwKMvM7mduQEYsDQXkLIwAVwC2AIwCmAJwaTZivAB9x0+Upgqdy7WrJkA9KZgB1EAoDWEMmDkB3GABF8UOQAoAlADoIOSgACRAJBQhvfxioX0cXd08fAKDQ8Mjo2Ix4k3MYAFEADwAHOWAvABMYfgBaFAIFAHNpOTAoCAYZCVgmkFgsGBsYKRs5fwTXDy8-QOCwiKiY-yhmXyA
)

### 💻 Code

```ts
const t = [4,20] as const
const t2 = [4, 20, 30] as const
const t3 = [10, 10] as [number, number] | [number, number, number]

// Works
new Date().setHours(...t)
new Date().setHours(...t2)

// Error: Expected 1-4 arguments, but got 0 or more.
new Date().setHours(...t3)
```

### 🙁 Actual behavior

Error: Expected 1-4 arguments, but got 0 or more.

The array type passed as a length of 2 or 3 so the message "got 0 arguments" is wrong.

### 🙂 Expected behavior

I expect the compiler should be able to recognize that the array type can only have length 2 or 3 and would not throw an error.

sandersn:
Previously, only calls with tuples were allowed, not unions of tuples. For example, this already works:

```
f(string, ...[number, string], string) ==> f(string, number, string, string)
```

But this does not:

```
f(string, ...([string] | [number, number])) ==> (f, string, string | number, number | undefined)
```

This PR allows union types like these *as the last argument*. It's possible to allow them anywhere, but quite a bit more complicated. And the code is already complicated enough: getEffectiveCallArguments now needs to return the minimum number of arguments as well as the argument list (which is the maximum number of arguments).

Also, this transformation should not happen when the tuple union argument is passed to a tuple union rest parameter: `[number] | [string]` is assignable to `[number] | [string]` but the transformed `number | string` is not. Checking for this requires passing around even more information.

Bottom line: I'm not sure the complexity is worth the new code. However, if this is a good idea, there are 3 things that need to be cleaned up:

1. More precise error messages--maybe. Right now error messages are reported in terms of the maximum number of arguments:
2. Allow tuples with trailing `...T` types. Not done or tested, but straightfoward.
3. Find a more elegant way to return the minimum number of arguments.

Fixes #42508
Supercedes #43882, but does less than half of that PR, in less than half the code.

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -28807,7 +28807,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         }
         const iife = getImmediatelyInvokedFunctionExpression(func);
         if (iife && iife.arguments) {
-            const args = getEffectiveCallArguments(iife);
+            const [args,] = getEffectiveCallArguments(iife);
             const indexOfParameter = func.parameters.indexOf(parameter);
             if (parameter.dotDotDotToken) {
                 return getSpreadArgumentType(args, indexOfParameter, args.length, anyType, /*context*/ undefined, CheckMode.Normal);
@@ -29002,7 +29002,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
 
     // In a typed function call, an argument or substitution expression is contextually typed by the type of the corresponding parameter.
     function getContextualTypeForArgument(callTarget: CallLikeExpression, arg: Expression): Type | undefined {
-        const args = getEffectiveCallArguments(callTarget);
+        const [args,] = getEffectiveCallArguments(callTarget);
         const argIndex = args.indexOf(arg); // -1 for e.g. the expression of a CallExpression, or the tag of a TaggedTemplateExpression
         return argIndex === -1 ? undefined : getContextualTypeForArgumentAtIndex(callTarget, argIndex);
     }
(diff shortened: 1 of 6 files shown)


sandersn:
After discussing this fix in the design meeting, we decided to not take this PR. It's too ad-hoc for the complexity of its code.

If/when we try to reform overload resolution's interaction with type parameter inference and contextual typing, we should address #42508 at the same time. Specifically, #42508 needs the ability to roll back inferences, same as overload type parameter inference. Then the fix here would be to make tuples generate multiple argument lists instead of trying to produce a single argument list. Each argument list can be applied against each overload (or a single signature if there's only one.)

fatcerberus:
> Each argument list can be applied against each overload (or a single signature if there's only one.)

Did somebody say "quadratic"? 🚎

sandersn:
Yes they did! We'll cross that bridge when we come to it. (One argument is: we check lots of other things that are quadratic because people ask us to, and expect them to change their types if compilation is too slow.)

PR #55193, opened 2023-07-29
error on duplicate symbols in object literals

### LINKED ISSUE #55163: TS1117 doesn't catch duplicate Symbols
# Bug Report

> error TS1117: An object literal cannot have multiple properties with the same name in strict mode.

I believe that TS1117 should catch duplicate Symbols in strict mode, or alternatively a new rule should be made to catch duplicate Symbols in an object literal.

### 🔎 Search Terms

symbol
ts1117

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about __symbols__

### ⏯ Playground Link

https://www.typescriptlang.org/play?#code/FASwdgLgpgTgZgQwMZQAQHkAOAVAnpqAZ1QG9hVUYoEATAezABtdUoBHAVwUYC5UOwITmkK4AtgCM6jYAF9gwJA0IQMmPljwFiAXlLkK7Lr1QBlcVMYA6OHRgAKAORHujgJRzUCYpvxEFSmAqqBAAFiCEAMJ2VEgQzABiCCCMuvoUhpzcfI5wIDAqjgA0BpnGOYRQgTTFpagA2s5ZjI4AujmM3hCOcgHKqmERACJ0RJBJKQBCHBCmoXQcjDSoemQZDVhWLoztqLn5hSXr9Zvbu46dhb0KAPQ3qFJhqJgw4KoXXY5eYMsMzKihBAANzQDDQLzoBBgEFwwDuqCUMFiqgkUEBQJAdiKqHAiORIT8ALexFsMBCoREFmkrAAHggxJhGFA+kFpFArIw6ABzeyDKIxKrxXATVJbZoeRTKNkc7m88KEEZjCAi6azeaLGgnTBi4ytNxAA

### 💻 Code

```ts

interface OpTypes {
  readonly equal: unique symbol
}

const Op: OpTypes = {
   equal: Symbol.for('equal')
} as OpTypes

const thisCorrectlyFails = {
    equal: 'first',
    equal: 'second',
    ['equal']: 'last'
}

const thisDoesntFailButShould = {
    [Op.equal]: 'first',
    [Op.equal]: 'last'
}


// both print 'last' and only have one property
// correct behavior, incorrect type hints for the symbol example

console.log(thisCorrectlyFails.equal)

console.log(thisDoesntFailButShould[Op.equal])
```

### 🙁 Actual behavior

TS1117 didn't catch the error in the Symbol example

### 🙂 Expected behavior

For the duplicate Symbol to trigger TS1117

## Context

We use Sequelize as an ORM, and the basic operations (`AND`, `OR` etc) are all `symbols`. We recently fixed a bug in our system where we were overriding an operator that TS didn't catch

https://sequelize.org/docs/v6/core-concepts/model-querying-basics/#operators

Zzzen:
Fixes #55163

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -49131,8 +49131,11 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
 
     function getEffectivePropertyNameForPropertyNameNode(node: PropertyName) {
         const name = getPropertyNameForPropertyNameNode(node);
-        return name ? name :
-            isComputedPropertyName(node) && isEntityNameExpression(node.expression) ? tryGetNameFromType(getTypeOfExpression(node.expression)) : undefined;
+        return name
+            ? name
+            : isComputedPropertyName(node)
+            ? tryGetNameFromType(getTypeOfExpression(node.expression))
+            : undefined;
     }
 
     function getCombinedModifierFlagsCached(node: Declaration) {
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -49131,11 +49131,8 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
 
     function getEffectivePropertyNameForPropertyNameNode(node: PropertyName) {
         const name = getPropertyNameForPropertyNameNode(node);
-        return name
-            ? name
-            : isComputedPropertyName(node)
-            ? tryGetNameFromType(getTypeOfExpression(node.expression))
-            : undefined;
+        return name ? name :
+            isComputedPropertyName(node) ? tryGetNameFromType(getTypeOfExpression(node.expression)) : undefined;
     }
 
     function getCombinedModifierFlagsCached(node: Declaration) {
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -25490,30 +25490,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             isEntityNameExpression(node.argumentExpression) ? tryGetNameFromEntityNameExpression(node.argumentExpression) : undefined;
     }
 
-    function tryGetNameFromEntityNameExpression(node: EntityNameOrEntityNameExpression) {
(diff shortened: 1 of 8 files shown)


src/compiler/checker.ts
```diff
@@ -49131,8 +49131,11 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
 
     function getEffectivePropertyNameForPropertyNameNode(node: PropertyName) {
         const name = getPropertyNameForPropertyNameNode(node);
-        return name ? name :
-            isComputedPropertyName(node) && isEntityNameExpression(node.expression) ? tryGetNameFromType(getTypeOfExpression(node.expression)) : undefined;
+        return name
```
jakebailey, line 49134: This formatting change doesn't seem required; can you just flip this back to the way it was and just drop the condition?

fatcerberus: Wait, I'm confused, what formatting change?  It's just replacing a whole nested ternary with `return name;` (unless GitHub is hiding part of the diff...)

jakebailey: Sorry, I commented on the first line; probably I should have selected the entire snippet but that would have been the entire diff in the file which felt egregious.

sandersn (commented):
This is plausible to me. @weswigham do you see any problems with relying exclusively on the type to get the name of a computed property? I *think* that's what we do elsewhere in the checker.

weswigham (approved):

weswigham:
@typescript-bot test this
@typescript-bot run dt
@typescript-bot test top100
@typescript-bot new perf test this

sandersn:
- The DT error is correct, and seems to expose a misunderstanding of how computed properties work.
- The error in xstate is correct, although it looks like the value is a shim with a manually annotated type copied from the built-in value. Fix is probably to cast one of the values to any or something. (@Andarist you might want to know about this failure.)
- No performance difference.

sandersn (approved):

Andarist:
@sandersn thanks for the heads up - luckily this code is already gone/restructured on our `next` branch so this won't affect us at all :)

Issue #55201, opened 2023-07-29
Array.forEach incorrectly narrows the inferred type when elements are annotated with union type


shogunsea:
# Bug Report

### 🔎 Search Terms
Array.forEach
union type 

### 🕗 Version & Regression Information
all versions

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about array and foreach

### ⏯ Playground Link

[Playground link with relevant **code](https://www.typescriptlang.org/play?ts=5.2.0-beta#code/JYOwLgpgTgZghgYwgAgGIHt0DU4BsCuKA3gFDLIBuehAXMgM5hSgDmA3CQL4kmiSyIUAIThQcBYmUrUIdRsxDspIOAFtZDJqw7de4aPCTIAClHQAHeslLkqE+nRFiZAbQC6yAD5pM4wu50eCAAPc3QoMGQYfBAEMGB0ECjMAAoiaXtkTjpTC3oASmspKAgwfCgku0J6ADoYcIBRRAALFJSqiEKAXgA+DMJ8jnIAemHkGIR0VXVwZDBmlHrcXHQAd1Y59AYIFHmUACMIZrgKBKhkdBhkGtU4c2QpUeQSsor+iFrb8zaO7r7fnRAA)**

### 💻 Code

```ts
interface FooValue {
  value: string;
}

interface BarValue {
  value: string;
  name: string;
}

interface Props {
  values: BarValue[] | FooValue[];
}

export function foo({ values }: Props) {
  return values.forEach((value) => value);
}
```

### 🙁 Actual behavior
The inferred type of `value` is simply `FooValue`:
<img width="582" alt="image" src="https://github.com/microsoft/TypeScript/assets/2238944/d400d767-6bfc-4443-a82c-9057c15c8d44">

### 🙂 Expected behavior
But really it should be `FooValue | BarValue`, which is correct when using `.map`:
<img width="645" alt="image" src="https://github.com/microsoft/TypeScript/assets/2238944/dbf985f5-4f02-441b-84ff-36d65d6564f3">

I suspect this issue happens because `forEach` isn't annotated with generic return type `U` while `map` is:
https://github.com/microsoft/TypeScript/blob/250065e55b8f2bb212cb9d19b191cfeb15184ee0/src/lib/es5.d.ts#L1235-L1246

In fact after changing `forEach` to `forEach<U>` fixes this problem locally:
[...]

PR #16223, opened 2017-06-02
Array cleanup

mhegazy:
This change unifies ReadOnlyArray and Array interface methods; it also removes generic signatures from both interfaces that were only used to propagate the type of `thisArg` parameter. This reverts the changes in https://github.com/Microsoft/TypeScript/pull/12784, and unfixes #12548; the removal of the generic signatures saves a good chunk of time of building the compiler and some 30 K symbols created.

before
```sh
> node built\local\tsc.js --diagnostics --p src\compiler\tsconfig.json
Files:            34
Lines:         90626
Nodes:        443548
Identifiers:  167732
Symbols:       93831
Types:         37620
Memory used: 390453K
I/O read:      0.00s
I/O write:     0.05s
Parse time:    0.88s
Bind time:     0.86s
Check time:    4.31s
Emit time:     3.11s
Total time:    9.16s

```

after: 
```sh
>node built\local\tsc.js --diagnostics --p src\compiler\tsconfig.json
Files:            34
Lines:         90421
Nodes:        438317
Identifiers:  165477
Symbols:       63565
Types:         27751
Memory used: 376816K
I/O read:      0.00s
I/O write:     0.05s
Parse time:    0.88s
Bind time:     0.97s
Check time:    3.69s
Emit time:     2.34s
Total time:    7.87s
```

Also fixes #13840

PR #55202, opened 2023-07-29
Update Array#forEach: use a generic type parameter to ensure elements are inferred correctly

shogunsea:
Fixes https://github.com/microsoft/TypeScript/issues/55201

### DIFF
--- a/src/lib/es5.d.ts
+++ b/src/lib/es5.d.ts
@@ -1237,7 +1237,7 @@ interface ReadonlyArray<T> {
      * @param callbackfn  A function that accepts up to three arguments. forEach calls the callbackfn function one time for each element in the array.
      * @param thisArg  An object to which the this keyword can refer in the callbackfn function. If thisArg is omitted, undefined is used as the this value.
      */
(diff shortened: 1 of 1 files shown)


src/lib/es5.d.ts
```diff
@@ -1237,7 +1237,7 @@ interface ReadonlyArray<T> {
      * @param callbackfn  A function that accepts up to three arguments. forEach calls the callbackfn function one time for each element in the array.
      * @param thisArg  An object to which the this keyword can refer in the callbackfn function. If thisArg is omitted, undefined is used as the this value.
      */
-    forEach(callbackfn: (value: T, index: number, array: readonly T[]) => void, thisArg?: any): void;
+    forEach<U>(callbackfn: (value: T, index: number, array: readonly T[]) => void, thisArg?: any): void;
```
MartinJohns, line 1240: The type argument isn't used anywhere.

MartinJohns:
See #16223, where the generic argument for `forEach` was explicitly removed again for performance reasons.

Issue #54939, opened 2023-07-09
[PERFORMANCE]: TS takes several seconds to validate types resulting in poor DX


mhevery:
# Bug Report

### 🕗 Version & Regression Information

- This is a slow type verification leading to bad DX.

### ⏯ Playground Link

[Playground link with relevant code](https://stackblitz.com/edit/stackblitz-starters-vgeb6t?description=&file=index.ts,package.json&title=json-server%20Starter)

### 💻 Code

```ts
// We can quickly address your report if:
//  - The code sample is short. Nearly all TypeScript bugs can be demonstrated in 20-30 lines of code!
//  - It doesn't use external libraries. These are often issues with the type definitions rather than TypeScript bugs.
//  - The incorrectness of the behavior is readily apparent from reading the sample.
// Reports are slower to investigate if:
//  - We have to pare too much extraneous code.
//  - We have to clone a large repo and validate that the problem isn't elsewhere.
//  - The sample is confusing or doesn't clearly demonstrate what's wrong.
```

I understand that this issue uses an external library, `drizzle` which is the source of performance problems. However, I think it would be useful for the TS team to see how people use types in real world and to see if either the `drizzle` team can fix the types to be better performant or the TS team can do something to better handle the use cases used by `drizzle`.

```typescript
import { integer, sqliteTable, text } from 'drizzle-orm/sqlite-core';

/**
 * This is a reproduction of slow TypeScript type validation with Drizzle.
 *
 * Steps to reproduce:
 * 1. Change "EDIT THIS" line and change `2` to `"2"` (add qoutes around the number)
 *    and save the file
 * 2. Because we have changed number to string the TS will now underline the line as
 *    being invalid.
 * 3. Go back and forth between the `2` and `"2"` and notice that there is a several
 *    second delay everytime you change source code and TS either has to add or remove
 *    the red underline. (TS Error)
 * 4. Once you establish a beseline feel to how long it takes TS to add/removo an error,
 *    comment out the `delayCount`/`latencyCount` block and repeat the steps
 * 5. Notice that now the add/remove error is instant.
 */

const x: number = 2; // EDIT THIS

export const edgeTable = sqliteTable('edges', {
  id: integer('id').primaryKey(),
  manifestHash: text('manifest_hash').notNull(),
  from: text('from'),
  to: text('to').notNull(),
  interaction: integer('interaction').notNull(),

  // COMMENT OUT THE LINES BELOW
  //////////////////////////////
  // START

[...]

PR #55224, opened 2023-08-01
Improve performance of maybe stack in recursiveTypeRelatedTo

jakebailey:
For #54939

I did a pprof profile of #54939 and found that a suspicious amount of time was being spent in `recursiveTypeRelatedTo` directly (self time).

image

The profile didn't really point to anything except for the loops over `maybeKeys`. I noticed that they were all operations where we could be looping over and over again when a set + stack could make every operation constant time.

So, I did that, and the test case from #54939 improved quite a bit:

```
Benchmark 1: old
  Time (mean ± σ):      3.115 s ±  0.067 s    [User: 4.403 s, System: 0.124 s]
  Range (min … max):    3.018 s …  3.196 s    10 runs
 
Benchmark 2: new
  Time (mean ± σ):      2.072 s ±  0.050 s    [User: 3.355 s, System: 0.135 s]
  Range (min … max):    1.985 s …  2.150 s    10 runs
 
Summary
  'new' ran
    1.50 ± 0.05 times faster than 'old'
```

On an even more inflated version of that test, I tripled the number of properties, and:

```
Benchmark 1: old
  Time (mean ± σ):      7.369 s ±  0.178 s    [User: 8.785 s, System: 0.225 s]
  Range (min … max):    7.136 s …  7.685 s    10 runs
 
Benchmark 2: new
  Time (mean ± σ):      3.437 s ±  0.073 s    [User: 4.913 s, System: 0.167 s]
  Range (min … max):    3.363 s …  3.566 s    10 runs
 
Summary
  'new' ran
    2.14 ± 0.07 times faster than 'old'
```

The current version of this PR doesn't appear to regress our regular test suites (some get better).


jakebailey:
@typescript-bot test this
@typescript-bot test top100
@typescript-bot run dt
@typescript-bot pack this

jakebailey:
Hm, slower...

src/compiler/core.ts
```diff
@@ -2882,3 +2882,45 @@ export function isNodeLikeSystem(): boolean {
...
+export interface StackSet<T extends {}> {
+    has(value: T): boolean;
+    push(value: T): void;
+    pop(): T;
+    get size(): number;
+}
+
+/** @internal */
+export function createStackSet<T extends {}>(): StackSet<T> {
+    const refs = new Map<T, number>();
```
jakebailey, line 2897: The refcount is overkill; the stack cannot contain duplicates.

jakebailey:

src/compiler/checker.ts
```diff
@@ -21301,18 +21304,37 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+                            }
+                        }
+                    }
+                    else {
+                        while (true) {
+                            const popped = maybeKeys.pop();
+                            if (popped === id) {
+                                break;
+                            }
                         }
```
jakebailey, line 21321: Probably this pattern should be a helper. Still, it's a WIP

jakebailey:

jakebailey:
Latest perf test run shows no change / maybe an improvement in xstate, while still preserving the major speedup in the drizzle example. It's rerunning again but that was just a cleanup which should still be as fast.

I'm testing this on some other repos that aren't yet in our test suite e.g. modern mui now.

jakebailey:
Alright, seems like it's still only neutral or perf positive (or noise). Very cool.

modern mui did not improve from what I can tell, though that's super noisy and hard to test anything but the largest improvements, which this is not.

jakebailey:
@typescript-bot pack this

src/compiler/core.ts
```diff
@@ -2882,3 +2882,56 @@ export function isNodeLikeSystem(): boolean {
...
+    // See: https://github.com/microsoft/TypeScript/issues/52924
+    /* eslint-disable no-var */
+    var set = new Set<T>();
+    var stack: T[] = [];
+    var end = 0;
+    /* eslint-enable no-var */
+
+    return {
+        has(value) {
+            return set.has(value);
```
jakebailey, line 2919: A part of me thinks that we could use just the array with linear searching up until a limit, rather than always using the set, but this PR appears to be an improvement regardless of that speculative optimization.

src/compiler/core.ts
```diff
@@ -2882,3 +2882,56 @@ export function isNodeLikeSystem(): boolean {
...
+     * Note that a value is popped, _then_ the callback is evaluated, meaning
+     * that if you return true to stop iteration, the value that the callback
+     * returned true for will still have been popped.
+     */
+    popAll(callback: (v: T) => boolean): void;
+}
+
+/** @internal */
+export function createStackSet<T extends {}>(): StackSet<T> {
+    // Why var? It avoids TDZ checks in the runtime which can be costly.
```
jakebailey, line 2909: No clue if this actually matters, may go test this elsewhere.

src/compiler/core.ts
```diff
@@ -2882,3 +2882,56 @@ export function isNodeLikeSystem(): boolean {
...
+
+
+/** @internal */
+export interface StackSet<T extends {}> {
+    /**
+     * Returns true if the value has already been pushed.
+     */
+    has(value: T): boolean;
+    /**
+     * Pushes to the stack. The value pushed must not have been pushed before.
```
DanielRosenwasser, line 2894: Risky! Name the method something that makes this clearer if you're really intent on doing this instead of just introducing a local `Set`.

src/compiler/checker.ts
```diff
@@ -21301,18 +21303,24 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     if (result === Ternary.True || result === Ternary.Maybe) {
                         // If result is definitely true, record all maybe keys as having succeeded. Also, record Ternary.Maybe
                         // results as having succeeded once we reach depth 0, but never record Ternary.Unknown results.
-                        for (let i = maybeStart; i < maybeCount; i++) {
-                            relation.set(maybeKeys[i], RelationComparisonResult.Succeeded | propagatingVarianceFlags);
-                        }
+                        maybeKeys.popAll(v => {
```
DanielRosenwasser, line 21306: Any idea how many of these typically get popped at a time?

src/compiler/core.ts
jakebailey: Can you come up with a better name? I tried over and over and did not come up with anything helpful.

I could just not have this and go back to loops.

src/compiler/checker.ts
jakebailey: No but I can see if I can collect that info.

src/compiler/core.ts
jakebailey: Oops, sorry, I thought this comment was on the `pop` line.

I could add the debug checks back and see if they make it slower.

jakebailey:

src/compiler/checker.ts
jakebailey: The latest version no longer requires a call each time, though in this case it's required. What optimization were you thinking of?

DanielRosenwasser: If you typically have very long strings of pushes and pops, you could even get rid of the explicit array and just iterate over the keys of the map.

*Ideally*, we could iterate from the end and bail out and get rid of the array entirely. See https://github.com/tc39/notes/blob/67ba44cc70f3338a2b8dcecd5eebec26ba5e9a40/meetings/2019-07/july-23.md#symbolreverse

src/compiler/core.ts
jakebailey: Tested and it doesn't seem like using a list is actually faster even when checking one thing. That or my microbenchmark is bad.

src/compiler/core.ts
jakebailey: Debug assert appears to be fine.

src/compiler/core.ts
jakebailey: Seems like it might matter, but the noise is a little difficult to get past.

ahejlsberg:
I'm certain @weswigham had a PR to this effect some time ago, but now I can't find it.

src/compiler/core.ts
```diff
@@ -2882,3 +2882,57 @@ export function isNodeLikeSystem(): boolean {
         && !(process as any).browser
         && typeof module === "object";
 }
+
+
+/** @internal */
+export interface StackSet<T> {
```
DanielRosenwasser, line 2888: Nit: maybe just put this in `utilities` unless we ever need it elsewhere.

jakebailey: Sure, I honestly don't know what the difference is between any of these files anymore.

DanielRosenwasser (approved):

jakebailey:

src/compiler/checker.ts
```diff
@@ -21313,20 +21319,33 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
             else {
                 // A false result goes straight into global cache (when something is false under
                 // assumptions it will also be false without assumptions)
                 relation.set(id, (reportErrors ? RelationComparisonResult.Reported : 0) | RelationComparisonResult.Failed | propagatingVarianceFlags);
-                maybeCount = maybeStart;
+                resetMaybeStack();
             }
             return result;
+
+            function resetMaybeStack(markAllAsSucceeded = false) {
```
DanielRosenwasser, line 21340: I wouldn't even make the parameter optional - there are 3 call-sites

jakebailey:
@typescript-bot pack this

DanielRosenwasser (approved):

Issue #2185, opened 2015-03-02
Declaring a var inside a loop with a let of the same name gives no error


JsonFreeman:
``` ts
for (let v; ; ) { var v; }
for (let v in []) { var v; }
for (let v of []) { var v; }
```

These all should be errors per the ES6 spec sections 13.6.3.1 and 13.6.4.1:

It is a Syntax Error if any element of the BoundNames of LexicalDeclaration also occurs in the
VarDeclaredNames of Statement.

mhegazy:
The var scope is not the for statement, it is the enclosing function. so i do not think this is an error

JsonFreeman:
That's true, but the spec still says it's an error.

mhegazy:
@bterlson is this really intended to be an error?

bterlson:
Yes, generally speaking it is an error for a var decl to hoist past a lexical decl.

JsonFreeman:
@bterlson, does that include blocks as well?
```ts
{
   let x;
   {
      var x;
   }
}
```

bterlson:
@JsonFreeman Yes, believe this is covered by the static semantics in [13.1.1](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-block-static-semantics-early-errors).

DanielRosenwasser:
Thanks @Andarist!

Andarist:
Closing 4-digit issues feels good 🤣

PR #55228, opened 2023-08-01
Error on shadowed var declarations without initializers

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/2185

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -41020,12 +41020,8 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             return;
         }
 
-        // skip variable declarations that don't have initializers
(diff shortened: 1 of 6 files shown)


jakebailey:
@typescript-bot test this
@typescript-bot test top100

DanielRosenwasser (approved):
Can you add a comment at the top of the function that reflects the spec text with respect to loops? See https://tc39.es/ecma262/multipage/ecmascript-language-statements-and-declarations.html#sec-for-statement-static-semantics-early-errors

Andarist:
@DanielRosenwasser is it worth adding if that's not a loop-specific rule? The same~ is said about blocks (and I already have tests for this here):
https://tc39.es/ecma262/multipage/ecmascript-language-statements-and-declarations.html#sec-block-static-semantics-early-errors

PR #55229, opened 2023-08-01
Only infer readonly tuples for `const` type parameters when constraints permit

### LINKED ISSUE #51931: `const` doesn't work with `extends`
# Bug Report

I'm not sure if this is intended behavior. If this is not a bug, I'd suggest supporting this as a feature.

### 🔎 Search Terms

const, args, generic args, extends

v5.0.0

### ⏯ Playground Link

[Playground link with relevant code](https://www.typescriptlang.org/play?ts=5.0.0-dev.20221217#code/CYUwxgNghgTiAEAzArgOzAFwJYHtVNQB4w8BnDeAFXhAA8MRVhT4pUBPAbQF0A+ACgB0w2AHNSALioBKKZQDcAKEUlU5VvAC8BfgCIougDRIoEUiGlKA9Ffh34APQD8ioA)

### 💻 Code

```ts
declare function fn<const T extends any[]>(...args: T): T;

const a = fn("a", false);
```

### 🙁 Actual behavior

The return type is `any[]`

### 🙂 Expected behavior

The return type should be `["a", false]`

ahejlsberg:
Expands on #55034 to interpret `const` modifier on type parameters to mean "as const as possible without violating constraints". For example:

```ts
declare function fa1<const T extends unknown[]>(args: T): T;
declare function fa2<const T extends readonly unknown[]>(args: T): T;

fa1(["hello", 42]);  // ["hello", 42]
fa2(["hello", 42]);  // readonly ["hello", 42]
```

Previously the call to `fa1` would infer `unknown[]` because the compiler would first produce the candidate `readonly ["hello", 42]`, but then realize that isn't assignable to `unknown[]` and therefore go with the `unknown[]` constraint. With the changes in this PR, the compiler appropriately adjusts its inference to match the constraint.

Fixes #51931.

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -30045,7 +30045,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             return createTupleType(elementTypes, elementFlags);
         }
         if (forceTuple || inConstContext || inTupleContext) {
-            return createArrayLiteralType(createTupleType(elementTypes, elementFlags, /*readonly*/ inConstContext));
+            return createArrayLiteralType(createTupleType(elementTypes, elementFlags, /*readonly*/ inConstContext && !(contextualType && isMutableArrayOrTuple(contextualType))));
         }
         return createArrayLiteralType(createArrayType(elementTypes.length ?
             getUnionType(sameMap(elementTypes, (t, i) => elementFlags[i] & ElementFlags.Variadic ? getIndexedAccessTypeOrUndefined(t, numberType) || anyType : t), UnionReduction.Subtype) :
@@ -32623,7 +32623,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 names.push((arg as SyntheticExpression).tupleNameSource!);
             }
         }
-        return createTupleType(types, flags, inConstContext, length(names) === length(types) ? names : undefined);
+        return createTupleType(types, flags, inConstContext && !isMutableArrayOrTuple(getBaseConstraintOrType(restType)), length(names) === length(types) ? names : undefined);
     }
 
     function checkTypeArguments(signature: Signature, typeArgumentNodes: readonly TypeNode[], reportErrors: boolean, headMessage?: DiagnosticMessage): Type[] | undefined {
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -23266,8 +23266,8 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     }
 
     function isMutableArrayLikeType(type: Type): boolean {
-        // A type is array-like if it is a reference to the global Array or global ReadonlyArray type,
-        // or if it is not the undefined or null type and if it is assignable to ReadonlyArray<any>
+        // A type is mutable-array-like if it is a reference to the global Array type, or if it is not the
+        // undefined or null type and if it is assignable to Array<any>
         return isMutableArrayOrTuple(type) || !(type.flags & TypeFlags.Nullable) && isTypeAssignableTo(type, anyArrayType);
     }
 
@@ -23931,7 +23931,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             callback(getTypeAtPosition(source, i), getTypeAtPosition(target, i));
         }
         if (targetRestType) {
-            callback(getRestTypeAtPosition(source, paramCount, /*readonly*/ isConstTypeVariable(targetRestType) && !isMutableArrayLikeType(targetRestType)), targetRestType);
+            callback(getRestTypeAtPosition(source, paramCount, /*readonly*/ isConstTypeVariable(targetRestType) && !someType(targetRestType, isMutableArrayLikeType)), targetRestType);
         }
     }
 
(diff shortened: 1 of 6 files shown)


tests/cases/conformance/types/typeParameters/typeParameterLists/typeParameterConstModifiers.ts
```diff
@@ -100,3 +100,75 @@ type NotEmptyMapped<T extends Record<string, any>> = keyof T extends never ? nev
...
+
+declare function fc1<const T extends unknown[]>(f: (...args: T) => void, ...args: T): T;
+declare function fc2<const T extends readonly unknown[]>(f: (...args: T) => void, ...args: T): T;
+
+fc1((a: string, b: number) => {}, "hello", 42);
+fc2((a: string, b: number) => {}, "hello", 42);
+
+declare function fn1<const T extends { foo: unknown[] }[]>(...args: T): T;
+
+fn1({ foo: ["hello", 123] }, { foo: [true]});
```
Andarist, line 174: maybe worth adding some weirdo test based on reverse mapped types as well?
```ts
declare function invoke<
  const T extends readonly { key: string }[],
  const T2 extends {
    [K in keyof T]: T[K]["key"];
  }
>(f: (...args: T2) => void, ...args: T): T2;

const res = invoke((a, b) => {}, { key: "A" }, { key: "B" });
```

This works fine with the changes here but it's more esoteric and could not work if different internal checks would be used or something

src/compiler/checker.ts
```diff
@@ -23925,7 +23925,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             callback(getTypeAtPosition(source, i), getTypeAtPosition(target, i));
         }
         if (targetRestType) {
-            callback(getRestTypeAtPosition(source, paramCount), targetRestType);
+            callback(getRestTypeAtPosition(source, paramCount, /*readonly*/ isConstTypeVariable(targetRestType) && !isMutableArrayOrTuple(getBaseConstraintOrType(targetRestType))), targetRestType);
```
Andarist, line 23928: IIUC this won't work correctly with union constraints, for example:
```ts
declare function test<const T extends number[] | string[]>(
  f: (...args: T) => void
): T;

const res = test((a: string, b: string) => {});
```

ahejlsberg: Yeah, this should probably use `some(...)` to distribute over unions.

tests/cases/conformance/types/typeParameters/typeParameterLists/typeParameterConstModifiers.ts
ahejlsberg: Not so sure about this, seems like an orthogonal thing.

ahejlsberg:
@typescript-bot test this
@typescript-bot user test this inline
@typescript-bot run dt
@typescript-bot test top100

ahejlsberg:
Tests and performance look good. This should be ready to merge.

Andarist:
For what it’s worth - LGTM 😉

RyanCavanaugh (approved):

weswigham (approved):

src/compiler/checker.ts
```diff
@@ -30045,7 +30051,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             return createTupleType(elementTypes, elementFlags);
         }
         if (forceTuple || inConstContext || inTupleContext) {
-            return createArrayLiteralType(createTupleType(elementTypes, elementFlags, /*readonly*/ inConstContext));
+            return createArrayLiteralType(createTupleType(elementTypes, elementFlags, /*readonly*/ inConstContext && !(contextualType && isMutableArrayLikeType(contextualType))));
```
Andarist, line 30054: This strategy doesn't seem to cover for a case like this:
```ts
declare function test_mix<const T extends [number, number] | readonly [string, string]>(a: T): T

// actual: [number, number] | readonly [string, string];
// expected: [1, 2]
const mix1 = test_mix([1, 2]); 
```

ahejlsberg:
@typescript-bot test this
@typescript-bot user test this inline
@typescript-bot run dt
@typescript-bot test top100

ahejlsberg:
Tests and performance are unaffected except for the new errors in `vercel/swr`. The new errors occur in circumstances similar to

```ts
declare function foo<T extends unknown[] | readonly unknown[]>(args: T): T;

foo(["hello", "world"] as const);  // ["hello", "world"]
```

We now infer a mutable array type when an `[...] as const` array literal is contextually typed by a type parameter with a constraint that includes a mutable array type. Previously we'd always infer a readonly array type (which often would fail to be assignable to the constraint). The mutable array type is a more specific type than the readonly array type, so this shouldn't really cause issues. However, the `vercel/swr` code checks for _exact_ type matches, thus _any_ change is a break.

The simple fix is to change the `vercel/swr` code [here](https://github.com/vercel/swr/blob/main/_internal/src/types.ts#L298) to just

```ts
type ArgumentsTuple = readonly [any, ...unknown[]]
```

The mutable tuple in the current definition is redundant anyway.

rauchg:
Thanks @ahejlsberg!

t7yang:
but the `const` keyword would be a little bit confuse.

on RHS

```ts
const foo = [1] as const; // apply readonly
```


on LHS (I consider function parameter as LHS)

```ts
function bar<const T extends unknown[]>(a: T): T // not apply readonly
```


but if we use `satisfies` instead of `const` on LHS, semantics will be more consistent.

```ts
const foo = [1] satisfies unknown[]; // doesn't apply readonly

function bar<satisfies T extends unknown[]>(a: T): T // doesn't apply readonly

const foo = [1] as const satisfies unknown[]>; // okay, now apply readonly

function bar<satisfies T extends readonly unknown[]>(a: T): T // okay, now apply readonly
```

shuding:
That type was updated in `vercel/swr` v2.2.2. Thanks @ahejlsberg (and @Andarist for opening the PR)!

jcalz:
Oh, this is great! It's too bad that this didn't make it into the [TS5.3 Release Notes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-3.html).

Issue #3229, opened 2015-05-20
'default' gets incorrect completion info when a default export is present


DanielRosenwasser:
![image](https://cloud.githubusercontent.com/assets/972891/7717057/b7abe7e4-fe4f-11e4-86bb-318e06548e10.png)

DanielRosenwasser:
Thank you @Andarist!

PR #55231, opened 2023-08-01
Do not offer completions for default module member symbol

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/3229

### DIFF
--- a/tests/cases/fourslash/completionsDefaultKeywordWhenDefaultExportAvailable.ts
+++ b/tests/cases/fourslash/completionsDefaultKeywordWhenDefaultExportAvailable.ts
@@ -1,0 +1,16 @@
+/// <reference path="fourslash.ts" />
+
+// https://github.com/microsoft/TypeScript/issues/3229
+
+// @filename: index.ts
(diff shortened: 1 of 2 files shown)


tests/cases/fourslash/completionsDefaultKeywordWhenDefaultExportAvailable.ts
```diff
@@ -0,0 +1,16 @@
...
+
+// @filename: index.ts
+
+//// export default function () {}
+//// def/*1*/
+
+verify.completions(
+    { 
+        marker: "1", 
+        includes: [{ name: "default", kind: "keyword", sortText: completion.SortText.GlobalsOrKeywords }],
```
Andarist, line 13: I'd prefer to test this through smth like `exact: completion.globals` but I wasn't able to figure out quickly which one should I use here.

DanielRosenwasser (approved):
As soon as I saw the title I was pretty sure it was for an issue I filed about 8 years ago. 😄

Issue #54298, opened 2023-05-17
TypeScript 5.2 Iteration Plan


DanielRosenwasser:
This document outlines our focused tasks for TypeScript 5.2. It minimally indicates intent to investigate tasks or contribute to an implementation. Nothing is set in stone, but we will strive to complete these tasks in a reasonable timeframe.

Date       | Event
-----------|-------------------------
2023-05-30 | TypeScript 5.1 Release
2023-06-23 | Create 5.2 Beta (5.2.0) Build for Testing
2023-06-27 | **TypeScript 5.2 Beta Release**
2023-08-04 | Create 5.2 RC (5.2.1) Build for Testing
2023-08-08 | **TypeScript 5.2 RC Release**
2023-08-18 | Create 5.2 Final (5.2.2) Build for Testing
2023-08-22 | **TypeScript 5.2 Final Release** 🚀

```mermaid
gantt
    dateFormat  YYYY-MM-DD
    TypeScript 5.1 Stabilization Period : 2023-05-15, 2023-05-30
    TypeScript 5.2 Beta Development : 2023-05-15, 2023-06-27
    TypeScript 5.2 RC Development : 2023-06-27, 2023-08-04
    TypeScript 5.2 Stabilization Period : 2023-08-04, 2023-08-22
todayMarker stroke-width:5px,stroke:#0f0,opacity:0.5
```

# Compiler and Language

* [Decorator Metadata](https://github.com/microsoft/TypeScript/issues/53461)
* [`using` Declarations](https://github.com/microsoft/TypeScript/issues/52955)
* [`Array` method signature merging](https://github.com/microsoft/TypeScript/pull/53489)
* [Investigate Type Argument Placeholders](https://github.com/microsoft/TypeScript/pull/26349)
* [Investigate Plugins for File-Watching](https://github.com/microsoft/TypeScript/pull/51074)
* [Investigate Custom Transformers in `tsc`](https://github.com/microsoft/TypeScript/issues/54276)
* [Investigate Function Implementations Returning Indexed Access and Conditional Types](https://github.com/microsoft/TypeScript/issues/33014)
* [Investigate Interop Across TypeScript and Bundlers for Default Imports](https://github.com/microsoft/TypeScript/issues/54102)
* [`lib.d.ts` Updates](https://github.com/microsoft/TypeScript/issues/54280)

# Editor and Language Service

* [Refactor to Inline Variable](https://github.com/microsoft/TypeScript/issues/18459)
[...]

PR #55233, opened 2023-08-01
Fix an off-by-one error in getBaseDirectoriesFromRootDirs

### LINKED ISSUE #55232: TSServer import path autocomplete reads incorrect directories when a file is not a direct child of a root directory
# Bug Report

My project has the following (simplified) structure:
```
app/
  build/
    dts/
      a.d.ts
      a.graphql.d.ts # Codegened from a.graphql
      tmp/
        b.d.ts
        b.graphql.d.ts # Codegened from b.graphql
  src/
    a.ts
    a.graphql
    tmp/
      b.ts
      b.graphql
```
where `rootDirs` is configured to read from both `src/` and `build/dts`. In other words, I'm codegening some extra `.d.ts` files during build time which I can then import in my typescript code.

When tsserver computes autocomplete for an import path in the `a.ts` file, it correctly returns the relevant files:

Screenshot 2023-08-01 at 21 41 06

However, when this is done inside of a directory (in this example one called `tmp/`, but this works in general), **tsserver reads the `/tmp` system directory instead of the `src/tmp` folder**:

Screenshot 2023-08-01 at 21 44 57

TSServer is still searching the current script directory (in this case `src/tmp`), but instead of searching `build/dts/tmp` it then also searches `/tmp`. This is probably why this bug went unnoticed for such a long time - it mostly affects imports of "extra files" that are not in source already.

I traced this behaviour to the following code:
https://github.com/microsoft/TypeScript/blob/b87e3018226262ea195a56abfb000a80b92c8f8c/src/services/stringCompletions.ts#L633

If `scriptDirectory` is the same as `rootDirectory`, `relativeDirectory` is computed to `""`, and `relativeDirectory` is correctly appended to the root directories. This results in TSServer correctly searching both `src/` and `build/dst`.

However, if `scriptDirectory` (in our example `/Volumes/git/app/src/tmp`) is a child of `rootDirectory` (`/Volumes/git/app/src`), this line will result in `relativeDirectory` being `/tmp` instead of `/tmp`, causing `combinePaths` a few lines down to output `/tmp`. **This results in TSServer incorrectly searching `src/tmp` and `/tmp` (instead of `build/dts/tmp`)**. This is effectively an off-by-one error because we're accidentally grabbing a leading slash.

### 🔎 Search Terms

tsserver, getPathCompletionsAtPosition, import path, rootDirs

As far as I can tell from the code this bug has always existed.

PhoebeSzmucer:
Fixes #55232 

I couldn't find where this is tested (if it even is). If someone can help me out with that I can write a test.


sheetalkamat:
You would need to add test in `src/testRunner/unittests/tsserver/autoImportProvider.ts` following example of one of the test case there.

PhoebeSzmucer:
> You would need to add test in src/testRunner/unittests/tsserver/autoImportProvider.ts following example of one of the test case there.

Isn't that for auto-import and not path completion? Or do these go through the same code path?

PhoebeSzmucer:
I added a throw statement to `getBaseDirectoriesFromRootDirs` and the only test that failed was `tests/cases/fourslash/completionForStringLiteralRelativeImport4.ts`

PhoebeSzmucer:
@microsoft-github-policy-service agree [company="Palantir"]

PhoebeSzmucer:
@microsoft-github-policy-service agree company="Palantir"

PhoebeSzmucer:
@sheetalkamat I added a fourslash test that failed before and succeeds now

sheetalkamat (changes requested):

src/services/stringCompletions.ts
```diff
@@ -630,7 +630,7 @@ function getBaseDirectoriesFromRootDirs(rootDirs: string[], basePath: string, sc
 
     // Determine the path to the directory containing the script relative to the root directory it is contained within
     const relativeDirectory = firstDefined(rootDirs, rootDirectory =>
-        containsPath(rootDirectory, scriptDirectory, basePath, ignoreCase) ? scriptDirectory.substr(rootDirectory.length) : undefined)!; // TODO: GH#18217
```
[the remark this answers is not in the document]
PhoebeSzmucer: Thank you for spotting. I fixed the issue and added more tests that test the behaviour both with and without the trailing slash.

PhoebeSzmucer:
@sheetalkamat any chance this can make it to the 5.2 release?

jakebailey:
5.2 was branched off a while ago and main is now targeting 5.3; see #54298. Given this bug isn't a regression from 5.1, I wouldn't expect to see it backported unless it's causing major ecosystem pain.

src/services/stringCompletions.ts
```diff
@@ -650,7 +650,10 @@ function getBaseDirectoriesFromRootDirs(rootDirs: string[], basePath: string, sc
     rootDirs = rootDirs.map(rootDirectory => normalizePath(isRootedDiskPath(rootDirectory) ? rootDirectory : combinePaths(basePath, rootDirectory)));
```
sheetalkamat, line 650:
```suggestion
    rootDirs = rootDirs.map(rootDirectory => ensureTrailingDirectorySeparator(normalizePath(isRootedDiskPath(rootDirectory) ? rootDirectory : combinePaths(basePath, rootDirectory))));
```

src/services/stringCompletions.ts
```diff
@@ -650,7 +650,10 @@ function getBaseDirectoriesFromRootDirs(rootDirs: string[], basePath: string, sc
     rootDirs = rootDirs.map(rootDirectory => normalizePath(isRootedDiskPath(rootDirectory) ? rootDirectory : combinePaths(basePath, rootDirectory)));
 
     // Determine the path to the directory containing the script relative to the root directory it is contained within
-    const relativeDirectory = firstDefined(rootDirs, rootDirectory => containsPath(rootDirectory, scriptDirectory, basePath, ignoreCase) ? scriptDirectory.substr(rootDirectory.length) : undefined)!; // TODO: GH#18217
+    const relativeDirectory = firstDefined(rootDirs, rootDirectory => {
```
sheetalkamat, line 653:
```suggestion
    const relativeDirectory = firstDefined(rootDirs, rootDirectory => containsPath(rootDirectory, scriptDirectory, basePath, ignoreCase) ? scriptDirectory.substr(rootDirectory.length) : undefined)!; // TODO: GH#18217
```

sheetalkamat (changes requested):

src/services/stringCompletions.ts
```diff
@@ -647,14 +647,14 @@ function getSupportedExtensionsForModuleResolution(compilerOptions: CompilerOpti
...
-    rootDirs = rootDirs.map(rootDirectory => normalizePath(isRootedDiskPath(rootDirectory) ? rootDirectory : combinePaths(basePath, rootDirectory)));
+    rootDirs = rootDirs.map(rootDirectory => ensureTrailingDirectorySeparator(normalizePath(isRootedDiskPath(rootDirectory) ? rootDirectory : combinePaths(basePath, rootDirectory))));
 
     // Determine the path to the directory containing the script relative to the root directory it is contained within
     const relativeDirectory = firstDefined(rootDirs, rootDirectory => containsPath(rootDirectory, scriptDirectory, basePath, ignoreCase) ? scriptDirectory.substr(rootDirectory.length) : undefined)!; // TODO: GH#18217
 
     // Now find a path for each potential directory that is to be merged with the one containing the script
     return deduplicate<string>(
-        [...rootDirs.map(rootDirectory => combinePaths(rootDirectory, relativeDirectory)), scriptDirectory],
+        [...rootDirs.map(rootDirectory => combinePaths(rootDirectory, relativeDirectory)), scriptDirectory].map(baseDir => removeTrailingDirectorySeparator(baseDir)),
```
PhoebeSzmucer, line 657: I had to also add `removeTrailingDirectorySeparator` here because otherwise some tests got broken due to duplicated entries with mismatched trailing slashes. It's probably a good idea to return consistent results anyway.

sheetalkamat (approved):

Issue #55251, opened 2023-08-03
Quick info doesn't show index signature information when it's declared on the interface with a base type


Andarist:
# Bug Report

### 🔎 Search Terms

quick info index signature base type interface

### 🕗 Version & Regression Information

- This is the behavior in every version I tried

### ⏯ Playground Link

[Playground link with relevant code](https://www.typescriptlang.org/play?target=1&ts=5.2.0-dev.20230802#code/JYOwLgpgTgZghgYwgAgILIN4ChnINoDWAXMgM5hSgDmAuiSAK4C2ARtANxYC+WWAJhAQAbOFBQIA9iHLI4JVJzgA6SDIC8yAIwAGdsgD0+5AEcGwBAQCSIGBJIAKUAIAeASjR5ylELXrM2ULygkLCIKAAKmDzB0PBIyABCyBDOkCB8pMiR2LiEJF7UdMiMrBzc-IIiYsiS0mDILCQJnCwqEOpaugZGpuZWNnbF-tBYQA)

### 💻 Code

```ts
interface A {
  [k: string]: number;
}

declare const a: A;
a.test = 10; // quickInfo: (index) A[string]: number

interface P {}
interface B extends P {
  [k: string]: number;
}
declare const b: B;
b.test = 10; // quickInfo: number
```

### 🙁 Actual behavior

The first quick info has an index symbol information in it but the second on doesn't.

### 🙂 Expected behavior

I expect both to disp;ay roughly the same information

PR #55252, opened 2023-08-03
Transfer all declared symbols onto the type from the interface that extends another

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/55251


src/compiler/checker.ts
```diff
@@ -12583,7 +12583,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         const baseTypes = getBaseTypes(source);
         if (baseTypes.length) {
             if (source.symbol && members === getMembersOfSymbol(source.symbol)) {
-                members = createSymbolTable(source.declaredProperties);
+                members = createSymbolTable(arrayFrom(members.values()));
```
jakebailey, line 12586: Huh, `createSymbolTable` probably should take `Iterable<Symbol>` given its implementation. I guess we've never given it anything but arrays.

DanielRosenwasser:

DanielRosenwasser:

jakebailey:
@typescript-bot test top100
@typescript-bot run dt

jakebailey:
Aha, that looks like a real failure.

Andarist:
Huh, this was fun! The inherited property member from the abstract class had the same name as own type parameter - so `addInheritedMembers` skipped it because the symbol of that name already existed in `members`. I fixed this by excluding type parameters from this "transfer".

src/compiler/checker.ts
```diff
@@ -12829,7 +12829,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         const baseTypes = getBaseTypes(source);
         if (baseTypes.length) {
             if (source.symbol && members === getMembersOfSymbol(source.symbol)) {
-                members = createSymbolTable(source.declaredProperties);
+                members = createSymbolTable(members.values(), SymbolFlags.TypeParameter);
```
Andarist, line 12832: arguably this could be changed to smth like:
```suggestion
                members = createSymbolTable(source.declaredProperties);
                const indexSymbol = members.get(InternalSymbolName.Index);
                if (indexSymbol) {
                    members.set(InternalSymbolName.Index, indexSymbol);
                }
```

but i like the current version better (if only there is no issue with it rn). It's easy to forget forwarding some of those special members and the current version ensures that all of them **are** forwarded.

jakebailey: I'm definitely liking the change less now that `createSymbolTable` is doing filtering; I think I'd prefer `createSymbolTable` not changing at all and instead writing the explicit code with a comment if needed, which I think is must more common.

Andarist: done

jakebailey:
@typescript-bot test top200
@typescript-bot run dt

jakebailey:
Hm, I guess this broke things, somehow.

Andarist:
@jakebailey ooops, sorry for that! I was juggling PRs and forgot to set `members` to what I created there. This time I actually re-run the tests locally so it should be OK 🙈

jakebailey:
@typescript-bot test top200
@typescript-bot run dt

jakebailey:
Well, thanks for testing that our infrastructure actually works and it's not just always saying "looks good" 😄

Andarist:
Let's say that was exactly my intention!

jakebailey (approved):

jakebailey:

PR #55262, opened 2023-08-04
Fix class name references

### LINKED ISSUE #54607: Incorrect transpilation of class name references in static contexts
# Bug Report

### 🔎 Search Terms

class static in:title

- This is the behavior in every version I tried (note: static blocks were introduced in TS 4.4)

### ⏯ Playground Link

[Playground link with relevant code](https://www.typescriptlang.org/play?#code/MYGwhgzhAEBiD29oG8BQ1oQC5iwS2BWmHgDsJ4QBTAOhHgHMAKLACzwgBo5EBKaAL7pMOfIQAe0ALzQm-KQD4iJcpVr1mbDtwTx+QoSuzRKAE13Se8ANyomFyNDCkAnvOikAriBC2zumnE5WyA)

### 💻 Code

```ts
class Foo {
  static { console.log(this, Foo) }
  static x = () => { console.log(this, Foo) }
}
const oldFoo = Foo;
(Foo as any) = null;
oldFoo.x();
```

### 🙁 Actual behavior

If you click **Run**, the code that TypeScript generates prints something like this:

```
class Foo {} class Foo {}
class Foo {} null
```

The code that TypeScript generates looks like this:

```js
"use strict";
var _a;
class Foo {
}
_a = Foo;
(() => {
    console.log(_a, Foo);
})();
Foo.x = () => { console.log(_a, Foo); };
const oldFoo = Foo;
Foo = null;
oldFoo.x();
```

### 🙂 Expected behavior

If you run this code natively, it prints something like this:

```
class Foo {} class Foo {}
class Foo {} class Foo {}
```

I expected TypeScript to generate something like this instead:

```js
"use strict";
var _a;
class Foo {
}
_a = Foo;
(() => {
    console.log(_a, _a);
})();
_a.x = () => { console.log(_a, _a); };
const oldFoo = Foo;
Foo = null;
oldFoo.x();
```

Class declarations in JavaScript generate an outer binding name, which turns out to be mutable (and can be reassigned). Separately, JavaScript also generates an immutable inner binding name that is only in scope within the class body. Referencing the class name within the class body gives the inner immutable binding name, and referencing the class name outside the class body gives the mutable outer binding name.

Presumably the `_a` that TypeScript is generating is this immutable inner binding name for code within the class body that needs to be moved out of the class body during syntax lowering. However, TypeScript isn't using the `_a` variable in all places where the immutable inner binding name should be referenced instead of the outer binding name. This leads to bugs in the case where the outer binding name is reassigned.

I discovered this bug in TypeScript because I'm trying to figure out how to handle this in esbuild, and as part of that I'm investigating how TypeScript transpiles code that does this.

rbuckton:
This fixes our emit for references to a class name inside of a class body when down-level emit moves the member containing the reference outside of the class body. This is necessary to preserve the "double bind" semantics of class names, where references to a class name inside of a class body are not affected when the reference to the class *outside* of the class body is changed.

We generally emit members outside of a class body when downleveling static fields, class static blocks, and private methods/accessors.

Fixes #54607

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -846,7 +846,6 @@ import {
     NodeCheckFlags,
     NodeFlags,
     nodeHasName,
-    nodeIsDecorated,
     nodeIsMissing,
     nodeIsPresent,
     nodeIsSynthesized,
@@ -1070,7 +1069,7 @@ import {
     WhileStatement,
     WideningContext,
     WithStatement,
-    YieldExpression,
+    YieldExpression
 } from "./_namespaces/ts";
 import * as moduleSpecifiers from "./_namespaces/ts.moduleSpecifiers";
 import * as performance from "./_namespaces/ts.performance";
@@ -27964,37 +27963,19 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
 
         let declaration = localOrExportSymbol.valueDeclaration;
         if (declaration && localOrExportSymbol.flags & SymbolFlags.Class) {
-            // Due to the emit for class decorators, any reference to the class from inside of the class body
-            // must instead be rewritten to point to a temporary variable to avoid issues with the double-bind
-            // behavior of class names in ES6.
-            if (declaration.kind === SyntaxKind.ClassDeclaration
-                && nodeIsDecorated(legacyDecorators, declaration as ClassDeclaration)) {
-                let container = getContainingClass(node);
-                while (container !== undefined) {
-                    if (container === declaration && container.name !== node) {
-                        getNodeLinks(declaration).flags |= NodeCheckFlags.ClassWithConstructorReference;
-                        getNodeLinks(node).flags |= NodeCheckFlags.ConstructorReferenceInClass;
-                        break;
-                    }
-
-                    container = getContainingClass(container);
-                }
-            }
-            else if (declaration.kind === SyntaxKind.ClassExpression) {
-                // When we emit a class expression with static members that contain a reference
-                // to the constructor in the initializer, we will need to substitute that
-                // binding with an alias as the class name is not in scope.
+            // When we downlevel classes we may emit some code outside of the class body. Due to the fact the
+            // class name is double-bound, we must ensure we mark references to the class name so that we can
+            // emit an alias to the class later.
+            if (isClassLike(declaration) && declaration.name !== node) {
                 let container = getThisContainer(node, /*includeArrowFunctions*/ false, /*includeClassComputedPropertyName*/ false);
-                while (container.kind !== SyntaxKind.SourceFile) {
-                    if (container.parent === declaration) {
-                        if (isPropertyDeclaration(container) && isStatic(container) || isClassStaticBlockDeclaration(container)) {
-                            getNodeLinks(declaration).flags |= NodeCheckFlags.ClassWithConstructorReference;
-                            getNodeLinks(node).flags |= NodeCheckFlags.ConstructorReferenceInClass;
-                        }
-                        break;
-                    }
-
+                while (container.kind !== SyntaxKind.SourceFile && container.parent !== declaration) {
                     container = getThisContainer(container, /*includeArrowFunctions*/ false, /*includeClassComputedPropertyName*/ false);
+                }
+
+                if (container.kind !== SyntaxKind.SourceFile) {
+                    getNodeLinks(declaration).flags |= NodeCheckFlags.ContainsConstructorReference;
+                    getNodeLinks(container).flags |= NodeCheckFlags.ContainsConstructorReference;
+                    getNodeLinks(node).flags |= NodeCheckFlags.ConstructorReference;
                 }
             }
         }
--- a/src/compiler/types.ts
+++ b/src/compiler/types.ts
@@ -6022,12 +6022,10 @@ export const enum NodeCheckFlags {
     ContainsCapturedBlockScopeBinding        = 1 << 13,  // Part of a loop that contains block scoped variable captured in closure
     CapturedBlockScopedBinding               = 1 << 14,  // Block-scoped binding that is captured in some function
     BlockScopedBindingInLoop                 = 1 << 15,  // Block-scoped binding with declaration nested inside iteration statement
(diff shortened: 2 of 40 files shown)


src/compiler/checker.ts
```diff
@@ -1070,7 +1069,7 @@ import {
     WhileStatement,
     WideningContext,
     WithStatement,
-    YieldExpression,
+    YieldExpression
```
jakebailey, line 1072: Do you know what you ran which did this? Was it organize imports? Something else? Hoping to get to the bottom of these so we can be consistent (or, I guess formatting will prevent that).

rbuckton: Probably organize imports.

jakebailey: Hm, just tested running organize on `checker.ts` and this didn't happen, even when reordering things, really odd.

jakebailey:
(Abusing this PR slightly to make sure RWC is working; this PR touches emit so it should produce a diff)

@typescript-bot test this

src/compiler/checker.ts
jakebailey, line 1072:
```suggestion
    YieldExpression,
```

Well, in any case, probably best to leave it until we can figure out what's happening. This is of course a nit (still looking at the PR)

src/compiler/types.ts
```diff
@@ -6022,12 +6022,10 @@ export const enum NodeCheckFlags {
     ContainsCapturedBlockScopeBinding        = 1 << 13,  // Part of a loop that contains block scoped variable captured in closure
     CapturedBlockScopedBinding               = 1 << 14,  // Block-scoped binding that is captured in some function
     BlockScopedBindingInLoop                 = 1 << 15,  // Block-scoped binding with declaration nested inside iteration statement
-    ClassWithBodyScopedClassBinding          = 1 << 16,  // Decorated class that contains a binding to itself inside of the class body.
-    BodyScopedClassBinding                   = 1 << 17,  // Binding to a decorated class inside of the class's body.
```
jakebailey, line 6026: Should we shift the other bits down or is it no big deal to leave them empty until later?

jakebailey (approved):
Seems correct to me though it's a little hard for me to be sure staring at the emit output 😄

src/compiler/checker.ts
rbuckton: I think it was a result of running Organize Imports to remove an unused import. I just checked locally and that does remove the trailing `,`.

src/compiler/types.ts
sandersn: this is a driveby cleanup, right? I don't see any removal of BodyScopedClassBinding in the rest of the PR.

src/compiler/checker.ts
```diff
@@ -27964,38 +27963,20 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
-                while (container.kind !== SyntaxKind.SourceFile) {
-                    if (container.parent === declaration) {
-                        if (isPropertyDeclaration(container) && isStatic(container) || isClassStaticBlockDeclaration(container)) {
-                            getNodeLinks(declaration).flags |= NodeCheckFlags.ClassWithConstructorReference;
-                            getNodeLinks(node).flags |= NodeCheckFlags.ConstructorReferenceInClass;
-                        }
-                        break;
-                    }
-
+                while (container.kind !== SyntaxKind.SourceFile && container.parent !== declaration) {
```
sandersn, line 27971: as I read it, this `while` loop is semantically equivalent to the old one. Am I missing anything?

src/compiler/checker.ts
```diff
@@ -27964,38 +27963,20 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
 
         let declaration = localOrExportSymbol.valueDeclaration;
         if (declaration && localOrExportSymbol.flags & SymbolFlags.Class) {
-            // Due to the emit for class decorators, any reference to the class from inside of the class body
```
sandersn, line 27967: To check my understanding: before, the _a reference was only generated for ClassExpressions and decorated classes. Now it happens for any class declaration/expression that references itself in its body. Is that correct?

tests/baselines/reference/decoratedClassExportsCommonJS1.js
```diff
@@ -24,5 +24,5 @@ let Testing123 = Testing123_1 = class Testing123 {
 exports.Testing123 = Testing123;
 Testing123.prop1 = Testing123_1.prop0;
 exports.Testing123 = Testing123 = Testing123_1 = __decorate([
-    Something({ v: () => Testing123_1 })
+    Something({ v: () => Testing123 })
```
sandersn, line 27: I don't understand how this diff happened. I assume it's because of the first block in the checker getting deleted, but I thought the edit to the second block would cause the reference to still get generated as Testing123_1.

sandersn (approved):
Seems good, although I have a few questions. And the emit changes Testing123_1 -> Testing123 are unexpected to me, so worth double-checking.

src/compiler/types.ts
rbuckton: All references to `BodyScopedClassBinding` were removed in a prior PR, but I neglected to remove the flag. I can fix up the bits used as well.

src/compiler/checker.ts
rbuckton: No, they are equivalent, but I feel this is a bit easier to read and reason over.

src/compiler/checker.ts
rbuckton: We won't always generate the reference, but now we will always set the flag. It's just that the cases where we need the flag set were no longer restricted to the cases mentioned in the comment.

tests/baselines/reference/decoratedClassExportsCommonJS1.js
rbuckton: `Testing123_1` represents the class binding *inside* the class body, however a class decorator is evaluated *outside* of the class body, so it was pointing to the wrong reference prior to this change.

src/compiler/checker.ts
sandersn: Agree, the new one is much easier to read.

PR #55264, opened 2023-08-04
fix(55258): JSDoc render with @param Object properties

### LINKED ISSUE #55258: JSDoc render with @param Object properties

- VS Code Version: 
- OS Version: 

Steps to Reproduce:

1. Go to vscode.dev
2. Paste the following block of code
```js
/**
 * @param {Object} options the args object
 * @param {number} options.alpha first number
 * @param {number} options.bravo second number
 * @param {Function} callback the callback function
 * @returns {number}
 */
function addNumbersFromObject(options, callback = null) {
    if (!callback) return options.alpha + options.bravo;
    return callback();
}
```
3. Hover the name of the function `addNumbersFromObject` with your mouse and wait until JSDoc render is displayed

Expected:

The render should display the Object properties of `options` variable like the following image.

image

Actual:

The render does not display the Object properties of `options` variable.

Tricky solution:

Just replace the type of `options` variable in the JSDoc by `any` or `*` and it will display the Object properties in the render.

However, it will lose autocompletion inside code function when i'll try the access properties of `options` variable and I don't want to lose this feature.

In order to have autocompletion, we need to use `Object` type for `options` variable in JSDoc like the following image.

image

Recap:

I wish to have at the same time fully JSDoc and autocompletion on an Object variable where I know some of its properties.

Thanks !

a-tarasyuk:
Fixes #55258

### DIFF
--- a/src/services/jsDoc.ts
+++ b/src/services/jsDoc.ts
@@ -44,6 +44,8 @@ import {
     isIdentifier,
     isJSDoc,
     isJSDocParameterTag,
+    isJSDocPropertyLikeTag,
+    isJSDocTypeLiteral,
     isWhiteSpaceSingleLine,
     JSDoc,
     JSDocAugmentsTag,
@@ -246,6 +248,12 @@ export function getJsDocTagsFromDeclarations(declarations?: Declaration[], check
         }
         for (const tag of tags) {
             infos.push({ name: tag.tagName.text, text: getCommentDisplayParts(tag, checker) });
+
+            if (isJSDocPropertyLikeTag(tag) && tag.isNameFirst && tag.typeExpression && isJSDocTypeLiteral(tag.typeExpression.type)) {
+                forEach(tag.typeExpression.type.jsDocPropertyTags, propTag => {
+                    infos.push({ name: propTag.tagName.text, text: getCommentDisplayParts(propTag, checker) });
+                });
+            }
         }
     });
     return infos;
--- a/tests/baselines/reference/quickInfoJsDocTags12.baseline
+++ b/tests/baselines/reference/quickInfoJsDocTags12.baseline
@@ -1,0 +1,189 @@
+=== /tests/cases/fourslash/quickInfoJsDocTags12.ts ===
+// /**
+//  * @param {Object} options the args object
+//  * @param {number} options.a first number
+//  * @param {number} options.b second number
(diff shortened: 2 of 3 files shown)


src/services/jsDoc.ts
```diff
@@ -246,6 +248,12 @@ export function getJsDocTagsFromDeclarations(declarations?: Declaration[], check
         }
         for (const tag of tags) {
             infos.push({ name: tag.tagName.text, text: getCommentDisplayParts(tag, checker) });
+
+            if (isJSDocPropertyLikeTag(tag) && tag.isNameFirst && tag.typeExpression && isJSDocTypeLiteral(tag.typeExpression.type)) {
+                forEach(tag.typeExpression.type.jsDocPropertyTags, propTag => {
```
sandersn, line 253: does this improve or change the way `@typedef` displays?
`@callback` and (I think) `@overload` have similar nested tag structure, too. Do they have similar problems displaying quickinfo? They might not since they all exist as their own comments.

sandersn (changes requested):
Needs a test of `@typedef`, or at least a quick check of how it displays now.

If typedef improves, it could be worthwhile to follow up to see if callback could be improved too.

src/services/jsDoc.ts
a-tarasyuk: @sandersn I think this should not impact tags that have their own comments. Does the test [jsDocTypedefQuickInfo1](https://github.com/microsoft/TypeScript/blob/main/tests/cases/fourslash/jsDocTypedefQuickInfo1.ts) cover the display aspects for the typedef? Should I expand it to include additional cases?

sandersn: That test doesn't really cover the expanded display; it doesn't display the tags for the typedef at all because it's requesting the type for a parameter.

On the other hand, quick info display for typedefs themselves is *decent* already -- but it doesn't show the individual tags. For example

```js
/**
  * @typedef {Object} Opts
  *                   ^^^
  * @property {string} x hello
  * @property {string=} y hup
  * @property {string} [z] hap
  * @property {string} [w="hi"] hip
  *
*/
```

Shows the TS syntax for `Opts`, which is useful, but it would be nice to also show the property tags themselves.
*However*, that's a different feature from this change, and I'd like to get this one in on its own.

sandersn (approved):

Issue #33298, opened 2019-09-07
NumericLiteral nodes with value 9007199254740993 claim to have .text = 9007199254740992


JoshuaKGoldberg:
**TypeScript Version:**  3.7.0-dev.20190907

**Search Terms:** numeric literal max value 9007199254740992 9007199254740993 

**Code**

```ts
const ts = require("typescript");
const sourceFile = ts.createSourceFile("temp.ts", "9007199254740993", 99);
const literal = sourceFile.statements[0].expression;

// '9007199254740992'
console.log("Actual .text:", literal.text);

// '9007199254740993'
console.log("Expected .text:", literal.getText(sourceFile));
```

**Expected behavior:**

`.text` should be the direct textual representation of the literal's value, right?

**Actual behavior:**

...some kind of rounding behavior related to being above `Number.MAX_SAFE_INTEGER`?

**Playground Link:** https://astexplorer.net/#/gist/8ee21b9bb7973f0d0118658496cec5ec/9dcea811268d39e678732414b53d608e37747a63 - since this is an AST parser issue

**Related Issues:** Discovered while writing a codefix for #29863. 😆

JoshuaKGoldberg:
Did a little bit of digging. This happens in `checkBigIntSuffix` (in `scanner.ts`):

https://github.com/microsoft/TypeScript/blob/73a2146583a774d322cfda8531fb22f34b6cc734/src/compiler/scanner.ts#L1478

Removing that `+` fixes the issue. Happy to send a PR if this is desirable!

lowr:
I found this issue while fixing #46195. Looking at the explanation for the `text` property of `LiteralLikeNode` ([link](https://github.com/microsoft/TypeScript/blob/4b794fe1dd0d184d3f8f17e94d8187eace57c91e/src/compiler/types.ts#L2127-L2131)), it seems like it is actually working as intended. I'm not sure it’s always been like so since when this issue was opened, though.

Also it's arguable that the property name `text` is confusing as one would expect it to be the node's textual representation in the source code as OP does.

RyanCavanaugh:
Agreed this is operating as documented.

PR #55268, opened 2023-08-04
Disallow numeric literals with negative numbers

Andarist:
This would prevent issues like [this one](https://github.com/microsoft/TypeScript/issues/54992) (that was fixed by https://github.com/microsoft/TypeScript/pull/55065 ). 

Since a negative number is not a numeric literal, I think it's worth disallowing such AST nodes from being created.


src/compiler/transformers/generators.ts
```diff
@@ -2528,7 +2528,7 @@ export function transformGenerators(context: TransformationContext): (x: SourceF
                 labelExpressions = [];
             }
 
-            const expression = factory.createNumericLiteral(-1);
+            const expression = factory.createNumericLiteral(Infinity);
```
Andarist, line 2531: This shows how `createNumericLiteral` is still a somewhat weird thing. `Infinity` is allowed because parsed numbers are "normalized" to text here:
https://github.com/microsoft/TypeScript/blob/83f3abda153f0f567e296a3c75cdca37c7eb065d/src/compiler/scanner.ts#L1765-L1773

So a sufficiently long number in the input can be stringified to `Infinity` 🤷‍♂️ I find it weird an there is [an issue](https://github.com/microsoft/TypeScript/issues/33298) about it that is labeled as a bug. While it doesn't mention `Infinity` case the issue there is basically related to the same "normalization". 

This issue is certainly fixable but I'm not 100% sure if you'd really want to fix it.

src/compiler/transformers/declarations.ts
```diff
@@ -1765,7 +1765,14 @@ export function transformDeclarations(context: TransformationContext) {
                     if (shouldStripInternal(m)) return;
                     // Rewrite enum values to their constants, if available
                     const constValue = resolver.getConstantValue(m);
-                    return preserveJsDoc(factory.updateEnumMember(m, m.name, constValue !== undefined ? typeof constValue === "string" ? factory.createStringLiteral(constValue) : factory.createNumericLiteral(constValue) : undefined), m);
+                    const newInitializer = typeof constValue === "undefined"
+                        ? undefined
```
rbuckton, line 1769:
```suggestion
                    const newInitializer = constValue === undefined
                        ? undefined

```

src/compiler/transformers/generators.ts
rbuckton: Don't change this to `Infinity` since we really should consider that invalid as well. I'd just use a large value that we're unlikely to actually hit in a generator, something like `2 ** 32 - 1` (max uint32). The only reason it's set to `-1` is to make it obvious something went wrong in tests.

tests/baselines/reference/binaryIntegerLiteral.types
```diff
@@ -15,7 +15,7 @@ var bin3 = 0B1111111111111111111111111111111111111111111111110100101010000001011
...
...111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111101001010100000010111110001111111111 : number
```
rbuckton, line 18: I really hope no one is relying on this.

rbuckton (changes requested):

rbuckton:
@typescript-bot perf test
@typescript-bot run dt
@typescript-bot test this

src/compiler/transformers/generators.ts
Andarist: > Don't change this to Infinity

Ye, I committed it this way specifically to add this PR comment 😅 

> since we really should consider that invalid as well.

Strong agree, I had versions of this work that forbid this as well but decided to roll them back and only start the conversation about this for now. I'm open to following up with further changes but it seems that the mentioned "normalization" of parsed numeric literals is relied upon in a couple of places. They can be adjusted but I'm not sure if this is something you'd like to do. In a way, it would be easiest to just drop this whole normalization altogether and just cast parsed text to string in those other places (to produce correct resolved property names etc).

> I'd just use a large value that we're unlikely to actually hit in a generator, something like 2 ** 32 - 1 (max uint32).

pushed out a change with `Number.MAX_SAFE_INTEGER`

rbuckton (approved):

PR #55269, opened 2023-08-04
Organize type imports

### LINKED ISSUE #52820: Support customization of where type imports are placed in organize/sort imports
Filing this so I don't forget about it; after #52115, we allow (though, do not fully expose support for) the ability to customize the sort order used by organize/sort/auto imports.

The intent is that we are able to match commonly used linters / formatters, leading towards fewer errors and edits needed to reach the "final state" of a file. e.g., instead of inserting an import at the end, then needing a linter to move it, we could just put it in the right place. Or, if you accidentally (or intentionally!) run Organize Imports, it should be possible to configure it such that it doesn't do anything.

This is mostly working already, however, our sort always puts type imports at the end. This differs from `eslint-plugin-simple-import-sort` and `dprint`, which ignore the `type` keyword and sort things that way (https://github.com/lydell/eslint-plugin-simple-import-sort/issues/124).

We should be able to add a new configuration which controls this behavior.

I have half a change for this sitting in a branch already, but it still needs some work.

iisaduan:
- Adds a configuration to change where named type imports are sorted. `organizeImportsTypeOrder` can be specified as `"first"`,  `"last"`, or `"inline"`, with default behavior being type imports are sorted last.
- Autoimports will place the new named import at the end of the list of named imports unless the imports are sorted as specified
- Fixes a bug in `verifyImportFixAtPosition` (function used for testing)

fixes #52820


src/server/protocol.ts
```diff
@@ -3598,6 +3598,13 @@ export interface UserPreferences {
      * Default: `false`
      */
     readonly organizeImportsCaseFirst?: "upper" | "lower" | false;
+    /**
+     * Indicates where named type-only imports should sort. The default order ("inline") sorts imports regardless of if imports are
+     * type-only.
+     *
+     * Default: `inline`
```
jakebailey, line 3605: The current default is effectively "last"; for backwards compat I'm thinking we should probaby keep that?

iisaduan: That makes sense to me, and is inline with the past changes to organize imports

jakebailey: Of course, changing the default to inline was what I was doing when I was attempting to make this change, was very helpful in debugging.

jakebailey:
This same code should also be affecting auto-imports; do we have enough test coverage for that? I'm not seeing any of those tests change which is suspicious as either it means that it's not working, or we're missing tests for auto-imports with type imports.

tests/cases/fourslash/autoImportTypeImport1.ts
```diff
@@ -0,0 +1,41 @@
...
+    { organizeImportsTypeOrder: "inline" }
+);
+
+verify.importFixAtPosition([
+`import { A, D, type C, type B } from './foo';
+const b: B | C;
+console.log(A, D);`],
+    /*errorCode*/ undefined,
+    { organizeImportsTypeOrder: "last" }
+);
```
andrewbranch, line 33: These both seem to have the wrong behavior, right? In each case, prior to the import, the specifiers are correctly sorted according to the preference, but afterwards, they’re not?

jakebailey: Yeah, noticed this too; I assumed this was just a starting point for further improvements.

iisaduan: `inline` looks alright to me since `A, D, C` is unsorted, but I agree (and am looking into) why `last` is funky. I wasn't sure if `first` should have inserted `type B` at the beginning.

iisaduan: I also should note that the previous behavior for auto-imports (which should be equivalent to `last`) also inserted `type B` at the end

andrewbranch: I think `first` could insert type-only specifiers into unsorted lists at the beginning, but I understand why it didn’t. You’re right that `inline` was also unsorted. Just `last` looks suspicious here.

src/services/organizeImports.ts
```diff
@@ -721,23 +728,77 @@ class ImportSpecifierSortingCache implements MemoizeCache<[readonly ImportSpecif
...
+            return SortKind.None;
+        }
+        return typeCaseSensitivity & regularCaseSensitivity;
+    }
+
+    // else inline
     return detectSortCaseSensitivity(specifiers, specifier => specifier.name.text, collateCaseSensitive, collateCaseInsensitive);
 }, new ImportSpecifierSortingCache());
 
+function getCategorizedSpecifiers(specifiers: readonly ImportSpecifier[]) {
```
andrewbranch, line 779: This can probably replaced by the existing `groupBy` helper

tests/cases/fourslash/autoImportTypeImport3.ts
```diff
@@ -0,0 +1,38 @@
...
+// console.log(A, D);`],
+//     /*errorCode*/ undefined,
+//     { organizeImportsIgnoreCase: true, organizeImportsTypeOrder : "inline" });
+
+// verify.importFixAtPosition([
+// `import { A, type B, type C, D } from './foo';
+// const b: B | C;
+// console.log(A, D);`],
+//     /*errorCode*/ undefined,
+//     { organizeImportsIgnoreCase: true, organizeImportsTypeOrder : "first" });
```
andrewbranch, line 38: Did you mean to delete these or uncomment them?

tests/cases/fourslash/importNameCodeFix_importType10.ts
```diff
@@ -0,0 +1,27 @@
...
+
+goTo.marker("");
+verify.importFixAtPosition([
+`import {
+    type SomeInterface,
+    SomePig,
+} from "./exports.js";
+new SomePig`],
+    /*errorCode*/ undefined,
+    { organizeImportsTypeOrder: "inline" }
```
andrewbranch, line 26: It might be more interesting to make this one `last`, since that would require swapping the order of import specifiers after `type` gets removed from `SomePig`

andrewbranch (commented):
Looks pretty good to me! Just a couple notes

tests/cases/fourslash/autoImportTypeImport3.ts
iisaduan: Ahhh thanks!

tests/cases/fourslash/importNameCodeFix_importType10.ts
iisaduan: `tests/cases/fourslash/importNameCodeFix_importType7.ts` does this same test case (since last is the default behavior). I can add a note in that test that states that

iisaduan: Though, now that I look at [importFixAtPosition](https://github.com/microsoft/TypeScript/blob/5a355093bb8c7d006966c3eb3aaf976cf4958517/src/harness/fourslashImpl.ts#L3523), it undoes the codefix so that we could make them into one test

src/server/protocol.ts
```diff
@@ -3643,6 +3643,13 @@ export interface UserPreferences {
      * Default: `false`
      */
     readonly organizeImportsCaseFirst?: "upper" | "lower" | false;
+    /**
+     * Indicates where named type-only imports should sort. "inline" sorts imports without regard to if the import is
```
andrewbranch, line 3647: When you add the VS Code option for this, make sure to say “type-only named imports” or “type-only import specifiers” since “type-only imports” could also refer to the relative order of these:

```ts
import type a from "./a";
import b from "./b";
```

which IIUC is unaffected by this preference.

andrewbranch (approved):
Nice work!

tests/cases/fourslash/autoImportTypeImport2.ts
```diff
@@ -0,0 +1,41 @@
...
+verify.importFixAtPosition([
+`import { A, type B, type C, D } from './foo';
+const b: B | C;
+console.log(A, D);`],
+    /*errorCode*/ undefined,
+    { organizeImportsTypeOrder: "inline" }
+);
+
+verify.importFixAtPosition([
+`import { A, type C, D, type B } from './foo';
```
jakebailey, line 28: I am also confused here, wouldn't this put it last? Or is it just because it's not currently sorted?

jakebailey (commented):
Some of the orderings seem weird, but is that because they aren't sorted?

tests/cases/fourslash/autoImportTypeImport1.ts
```diff
@@ -0,0 +1,41 @@
...
+
+// @Filename: /test.ts
+//// import { A, D, type C } from './foo';
+//// const b: B/**/ | C;
+//// console.log(A, D);
+
+goTo.marker("");
+
+verify.importFixAtPosition([
+`import { A, D, type C, type B } from './foo';
```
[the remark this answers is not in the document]
iisaduan: Since it's adding `type B` and it's determined that the types are unsorted (since it specified `inline`), it adds it on to the end. If the imports were unsorted before, it would always add onto the end, otherwise it might add the import somewhere weird in the middle, since we're not changing the order of the rest of the imports

tests/cases/fourslash/autoImportTypeImport2.ts
iisaduan: Yup, since it's not currently sorted

tests/cases/fourslash/autoImportTypeImport1.ts
```diff
@@ -0,0 +1,41 @@
...
+verify.importFixAtPosition([
+`import { A, D, type B, type C } from './foo';
+const b: B | C;
+console.log(A, D);`],
+    /*errorCode*/ undefined,
+    { organizeImportsTypeOrder: "last" }
+);
+
+verify.importFixAtPosition([
+`import { A, D, type C, type B } from './foo';
```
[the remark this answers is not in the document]
jakebailey: Ah, this is "first", so it's unsorted. Gotcha. I feel like these would benefit from a comment on each saying which are unsorted and which are sorted, but maybe those would go out of date...

tests/cases/fourslash/autoImportTypeImport4.ts
```diff
@@ -0,0 +1,111 @@
...
+goTo.marker("0");
+verify.importFixAtPosition([
+    `import { A, B, C, a } from "./exports1";\na;\nb;`,
+    `import { A, B, C, b } from "./exports1";\na;\nb;`,
+],
+    /*errorCode*/ undefined,
+    { organizeImportsTypeOrder : "last" });
+verify.importFixAtPosition([
+    `import { a, A, B, C } from "./exports1";\na;\nb;`,
+    `import { A, b, B, C } from "./exports1";\na;\nb;`
```
jakebailey, line 49: Just so I'm clear, these are alternatives such that both are acceptable? For this position, shouldn't it be deterministic and one always wins?

jakebailey (approved):
I've stared at the tests and I think everything is in order.

Note that the other new import sorting options are not actually exposed in VS Code yet (you have to use the "unstable" field IIRC, see `settings.template.json`), so I don't think we need to do that quite yet.

One thing I'm not sure I'm seeing is detection, in that we detect that imports are sorted "inline" and then use it. Am I missing that (there's a lot of tests...) or is that coming later?

tests/cases/fourslash/autoImportTypeImport4.ts
iisaduan: There's two results because there are two different codefixes that show up, one adds `a`, and the other adds `b`

jakebailey: Ohh I cannot read. Thanks.

iisaduan:
@andrewbranch 
> Note that the other new import sorting options are not actually exposed in VS Code yet (you have to use the "unstable" field IIRC, see `settings.template.json`), so I don't think we need to do that quite yet.
is `organizeImportsIgnoreCase` exposed to VScode?

tests/cases/fourslash/autoImportTypeImport1.ts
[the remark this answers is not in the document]
iisaduan: I'll work on adding some comments since it sounds like a good idea to make less people confused in the future. If the comments go out of date, then the tests will break (and hopefully whoever breaks the tests also updates the comments :0 )

andrewbranch:
I thought it was, but I guess it’s not.

jakebailey:
Yep, that's my understanding.

jakebailey (approved):

Issue #51440, opened 2022-11-08
Use ESM for our executables (tsc, tsserver, typingsInstaller)


jakebailey:
Our libraries are all UMD-ish, because they can be used in Node or in the browser. I didn't change this for the module transform.

However, most of our executables are CJS, as they are only intended to be run within Node. This includes tsc, typingsInstaller, and watchGuard. An exception is tsserver, which contains web code for vscode.dev.

tsc and typingsInstaller both share a lot of code; if they only run on Node anyway, we could potentially raise our minimum Node version to node 12, and ship them as ESM via esbuild's split ESM bundling. If we get vscode.dev off of tsserver (and onto tsserverlibrary), then all three can share code.

In my testing, this reduces our package size by an additional 7 MB, which is a great win.

See also #27891, #32949.

fatcerberus:
> if they only run on Node anyway, we could potentially raise our minimum Node version to node 12, and ship them as ESM

Isn't Node 14 the earliest version that supports ESM without a command-line flag?

jakebailey:
> Isn't Node 14 the earliest version that supports ESM without a command-line flag?

Nope, it's enabled by default in Node 12.17+.

jakebailey:
Another totally radical idea would be to just ship ESM _now_ with separate endpoints. It wouldn't behave how people want, as in no tree shaking, but it wouldn't be that much code. tsserver already pulls pretty much everything in anyway, so this wouldn't be all that much extra.

We'd just still have the dual package hazard, and have to get the weirdness of export mappings and self references.

lemanschik:
just my 5 cent as i am working on that since ages shipping ESM only is totally as that can get easy converted to CJS but thats never the case for CJS also browser support for ESM is more complet then the one for anything else :) also this would finaly open that we can drop our endless forks for other engines then NodeJS. it takes 100 of man years only to workaround your build processes. same for vscode.

lemanschik:
see: https://github.com/microsoft/vscode-loader/issues/46

lemanschik:
maybe you want to play around with:

- http://evmar.github.io/webtreemap/ <= typescript bloat demo
- https://github.com/evmar/webtreemap they use ur project as example

lemanschik:
@jakebailey i have a working ESM build in split of of Typescript i can do a PR but i am using Rollup and it is blazing fast and would exactly align with the Dweb and web 4.0 agenda

DanielRosenwasser:
[...]

Issue #27891, opened 2018-10-13
Reduce typescript package size


pauldraper:
## Search Terms

size, bloat, install

## Suggestion

The typescript package is large, and it only getting larger.

![screenshot from 2018-10-13 11-12-06](https://user-images.githubusercontent.com/1403271/46908127-43799000-cedb-11e8-9a3c-3b8bdb262ecc.png)

Version 3.1.3 is a whopping 40MB.

## Use Cases

TypeScript is used in many contexts.

A TypeScript formatter (e.g. [prettier](https://github.com/prettier/prettier)) does not need an entire compiler. It only needs a parser. And 45MB scripted parser is *orders of magnitude* larger than one would normally expect. (For reference, the installed npm package for Esprima -- the most compatible and compliant ES parser in the ecosystem -- is a mere 0.3MB.)

### Solution 1: Split up packages

* typescript (existing package; depends on typescript-compiler, typescript-parser, typescript-server)
* typescript-compiler (depends on typescript-parser)
* typescript-parser
* typescript-server (depends on typescript-compiler)

Optionally, there could be separate packages for typescript-config and typescript-i18n.

### Solution 2: Don't duplicate code

There is a lot of code duplication between

* lib/typescriptServices.js
* lib/typescriptServices.js
* lib/tsserver.js
* lib/tsserverlibrary.js

Don't duplicate the code.

## Checklist

My suggestion meets these guidelines:
* [x] This wouldn't be a breaking change in existing TypeScript / JavaScript code
* [x] This wouldn't change the runtime behavior of existing JavaScript code
* [x] This could be implemented without emitting different JS based on the types of the expressions
* [x] This isn't a runtime feature (e.g. new expression-level syntax)

pauldraper:
Interesting reading, thanks.

> TypeScript [2.9.0] has doubled in size since v2.0.0 - now 35 MB

It was closed in #25901, released in 3.1.1, which was 40MB. :slightly_frowning_face: 

---

It won't be hard at all to shrink the package size. For example, `lib/tsserver.js` and `lib/tsserverlibrary.js` are 98% identical.

```
$ du -b node_modules/typescript/lib/tsserver.js node_modules/typescript/lib/tsserverlibrary.js
7290127	node_modules/typescript/lib/tsserver.js
7308140	node_modules/typescript/lib/tsserverlibrary.js
[...]

Issue #46153, opened 2021-10-01
tsc -b incorrectly succeeds when there is a problem introduced by a transitive dependency


samreid:
# Bug Report

### 🕗 Version & Regression Information

- This changed between versions 3.6.0-dev.20190621 (good) and 3.6.0-dev.20190622 (bad)
The incorrect behavior was observed in many versions since then, including:
  * 4.4.2
  * 4.4.3
  * 4.5.0-dev.20210928

### ⏯ Playground Link

Reproducing the problem requires transitive project references, and running `tsc -b` from the command line, so I don't believe it can be done in the TypeScript playground.

### 💻 Code

My team observed that `tsc -b` can incorrectly succeed when a problem is introduced by a transitive dependency.  In fact, `tsc -b` can give two different results (type check passes or type check fails) for the same codebase.  We created a self-contained reproducible example to demonstrate the problem.  This creates 3 projects using project references, where project A depends on project B and that, in turn, depends on project C.  There is one TypeScript file in project A, none in B and one in C.

#### Step 1: Create the following 5 files, distributed into directories `projectA/`, `projectB/` and `projectC/`.

```ts
// projectA/A.ts
import C from '../projectC/C';

class A extends C {
  constructor() {
    super( {} );
    this.name.arbitraryFakeMethod();
  }
}
```

```json
// projectA/tsconfig.json
{
  "references": [
    {
      "path": "../projectB"
    }
  ],
  "compilerOptions": {
    "module": "commonjs",
    "target": "es5",
    "composite": true,
    "outDir": "./dist/",
    "incremental": true,
    "allowJs": true
  },
  "include": [
    "A.ts"
  ]
}
```

```json
// projectB/tsconfig.json
{
  "references": [
    {
      "path": "../projectC"
    }
  ],
  "compilerOptions": {
    "module": "commonjs",
    "target": "es5",
    "composite": true,
    "allowJs": true,
    "outDir": "./dist/",
    "incremental": true
  }
}
```

```ts
// projectC/C.ts
class C {
  name: string;
  constructor( options ) {
  }
}

export default C;
```

```json
// projectC/tsconfig.json
{
  "compilerOptions": {
    "module": "commonjs",
    "target": "es5",
    "composite": true,
    "allowJs": true,
    "outDir": "./dist/",
    "incremental": true
  },
  "include": [
    "C.ts"
  ]
}
```

#### Step 2: clean and build project A

[...]

Issue #30608, opened 2019-03-27
Pull required references from referenced projects.


rjamesnw:
## Search Terms

Inherit references from referenced project.
Pull required references from referenced projects.

## Suggestion

I have projects that have references to other projects, but one other project has references to another project it is dependent on.  That seems to be completely ignored, and is causing compiler errors.  It forces me to have to know all the references of the referenced projects and pull those also, which I should not have to do.

## Use Cases

This is pretty obvious.  In a typical C# project in Visual Studio (for example), referencing a project resolves other references from that referenced project, and so on.  It is not good practice to force projects to pull direct dependencies that are referenced indirectly by 3rd parties.

## Examples
_tsconfig.json_
```
{
  "references": [
    { "path": "3rdPartyProject" } // 3rdPartyProjectFile.d.ts
  ],
```
_3rdPartyProject/tsconfig.json_
```
{
  "references": [
    { "path": "Another3rdPartyProject" } // Another3rdPartyProjectFile.d.ts
  ],
```

This setup only outputs `3rdPartyProjectFile.d.ts`, as required by the first referenced projects.  Intellisense fails, however, because the references in `3rdPartyProject/tsconfig.json` are not included, thus `Another3rdPartyProjectFile.d.ts` is missing.

## Checklist

My suggestion meets these guidelines:

* [x] This wouldn't be a breaking change in existing TypeScript/JavaScript code
* [x] This wouldn't change the runtime behavior of existing JavaScript code
* [x] This could be implemented without emitting different JS based on the types of the expressions
* [x] This isn't a runtime feature (e.g. library functionality, non-ECMAScript syntax with JavaScript output, etc.)
* [x] This feature would agree with the rest of [TypeScript's Design Goals](https://github.com/Microsoft/TypeScript/wiki/TypeScript-Design-Goals).

RyanCavanaugh:
Not having references be transitive was an intentional design decision when we wrote project references - the idea being that if an intermediary layer intended to be an opaque wrapper over an upstream project, that should be possible.

That said, I think it's reasonable to consider an opt-in per-reference setting to pull in its (recursive?) dependencies.

Griffork:
I think there's another problem here because I have the same error but want a different outcome:

Based on how I had expected typescript's tsconfig's `references` to work I'd expect prerequisite projects (in my case my Common or Client projects) to be built on their own, and then their declaration files to be ingested by my dependant project (Server) during build. 

[...]

PR #55273, opened 2023-08-04
Move tsserverlibrary.js to typescript.js, make tsserverlibrary.js a shim

jakebailey:
The thinking here is:

- External users (ts-eslint, others) are wanting to use `ProjectService` without the problems of importing typescript twice.
- Moving `ProjectService` into `typescript.js` is effectively moving the entire server namespace into `typescript.js`.
- At that point, why even have `tsserverlibrary.js`?

Before this change, the two bundles loaded in roughly:

- `typescript.js` - 206.30ms (± 0.30%)
- `tsserverlibrary.js` - 223.68ms (± 0.21%)

So, this only adds about 20ms of extra time to importers. But, it does increase the size of `typescript.js` by 7%: https://github.com/microsoft/TypeScript/actions/runs/5814577872#summary-15764542922

In the process of this, we also noticed that we can in the future:

- Just make tsserver a thin wrapper around typescript.js. The two bundles contain exactly the same code except that tsserver adds a node host, which we probably want to export. That'd save even more space at no perf cost.
	- We could make tsc a wrapper around typescript.js too, technically, although that'd probably end up being slower and is probably not worth it.
	- But IMO these would probably be better off being improved via #51440.
- Maybe other optimizations?

cc @JoshuaKGoldberg @bradzacher 

For #27891


src/testRunner/unittests/publicApi.ts
```diff
@@ -19,21 +18,6 @@ describe("unittests:: Public APIs", () => {
         it("should be acknowledged when they change", () => {
             Harness.Baseline.runBaseline(api, fileContent, { PrintDiff: true });
         });
-
-        it("should compile", () => {
```
jakebailey, line 23: We actually already test this by nature of the new public API compiler tests I added, which disable skipLibCheck to test the actual public API. This change could be pulled out into another PR but it's needed for this PR because this test assumes that our d.ts files are self-contained, which is no longer true.

Alternatively, I can just make `tsserverlibrary.d.ts` a copy of `typescript.d.ts` and it'd be fine.

sheetalkamat: is is possible to create symlinks for tsserverlibrary.js and tsserverLibrary,d,ts to point to typescript.js and typescript.d.ts instead

jakebailey: Right now `tsserverlibrary` just reexports `typescript`, it's just that it's not self contained so this one particular test breaks (but isn't needed anyway).

In terms of the package itself, we can't really use symlinks as those are not universally supported on Windows or by package managers themselves. There are ways that people store packages which can't contain symlinks at all (e.g., zips, used in yarn, for example).

Herebyfile.mjs
```diff
@@ -403,28 +403,54 @@ export const watchMin = task({
...
+// TODO(jakebailey): how can we make this backwards compatible?
+// Does anyone actually load tsserverlibrary into the browser directly?
+// Do we have to create a <script>?
+const lsslJs = `
+if (typeof module !== "undefined" && module.exports) {
+    module.exports = require("./typescript.js");
+}
+else if (typeof importScripts !== "undefined") {
+    importScripts("./typescript.js");
+}
```
jakebailey, line 415: I'm not totally sure if this is a good idea in general; I'm guessing it's possible for this line to make it into someone else's bundle which will start importing things if it happens to be a worker... but maybe not if they're correctly emulating CJS?

I'm half wondering if it'd be better to just not do this line at all and leave it as a module reexport, and just wait for feedback that says we need to do better.

jakebailey: But, I somewhat expect that nobody is actually loading this bundle into a browser at all.

sheetalkamat (approved):

src/typescript/tsconfig.json
```diff
@@ -6,7 +6,7 @@
         { "path": "../compiler" },
         { "path": "../jsTyping" },
         { "path": "../services" },
-        { "path": "../deprecatedCompat" }
```
sheetalkamat, line 9: Is this intentional to drop drepcated compat from typescript.js  ?

jakebailey: It's already in the `server` tsconfig / project, since it's needed for both `tsserver` and `tsserverlibrary`; to make this PR I actually just deleted all of the contents of `typescript` and moved `tsserverlibrary` here, but git unfortunately can't show that!

jakebailey: They aren't? But, tsserverlibrary's been set up this way since I merged modules and has been fine; what breaks if this isn't done? If it's about the code itself, it's imported in the `server` namespace so gets pulled in there.

jakebailey: Removing `built` and doing `npx tsc -b ./src/typescript`, I see:

image

Then `hereby services` gives me a file with:

image

So, it seems like it all works?

sheetalkamat: i thinjk if you change something in deprecatedCompat, it wont build tsserver library since its not direct dependency unless server needs to be built too. (which is what may be is intended) but just pointing it out. ( i think it made much more difference with prepend, with modules i dont know)

jakebailey: Build as in typecheck or as in write out files? I guess dts emit may matter for this...

I'm surprised in that I think a lot of monorepo projects aren't just bulk including all transitive dependencies in their tsconfigs as that would get unweildy.

sheetalkamat: #46153 and #30608

jakebailey: Aha! Thanks!

I'll go figure out which transitive things we're missing.

jakebailey: I pulled this out into https://github.com/microsoft/TypeScript/pull/55339, which I'd like to merge first.

jakebailey: Given this isn't a regression (all I did was move `tsserverlibrary` and rename it), this shouldn't block merging of this PR, right? So the above isn't exactly a prereq?

jakebailey:
@typescript-bot pack this

Herebyfile.mjs
```diff
@@ -404,27 +404,50 @@ export const watchMin = task({
...
+`;
+
+const lsslDts = `
+import ts = require("./typescript.js");
+export = ts;
+`;
+
+const lsslDtsInternal = `
+import ts = require("./typescript.internal.js");
```
jakebailey, line 424: This is a little unfortunate for anyone who is attempting to pull these internal types, e.g. byots/ts-expose-internals but the fix would be quite simple for them; just use `typescript.internal.d.ts` for both.

jakebailey:
@JoshuaKGoldberg you may beat me too this, but I think we should try including the above package in ts-eslint and see if it compiles, to make sure my new d.ts works.

That and I need to also test vscode, which similarly imports this but also augments it IIRC. (I'll do that Monday)

jakebailey:
Actually, took me like 2 minutes to test in vscode, and it worked even with their augmentation, so it probably is fine everywhere else.

weswigham (approved):

jakebailey:
Meant to ping @mjbvz too; I don't think that this will affect VS Code at all; I think both of these files are preserved in your pruned package.

PR #54734, opened 2023-06-21
Interactable parameter inlay hints

MariaSolOs:
Part of #47693. This first iteration adds "go to definition" support for parameter inlay hints.

PR #55274, opened 2023-08-04
Modify interactive inlay hints API to be more backwards compatible

jakebailey:
This tweaks the API in #54734 to make it such that it's less likely to break consumers and require changes such as https://github.com/microsoft/monaco-editor/pull/4100.

Unfortunately, this will require another vscode change to modify https://github.com/microsoft/vscode/pull/189344/files.

### DIFF
--- a/src/services/inlayHints.ts
+++ b/src/services/inlayHints.ts
@@ -159,9 +159,11 @@ export function provideInlayHints(context: InlayHintsContext): InlayHint[] {
     }
 
     function addParameterHints(text: string, parameter: Identifier, position: number, isFirstVariadicArgument: boolean, sourceFile: SourceFile | undefined) {
-        let hintText: string | InlayHintDisplayPart[] = `${isFirstVariadicArgument ? "..." : ""}${text}`;
+        let hintText = `${isFirstVariadicArgument ? "..." : ""}${text}`;
+        let displayParts: InlayHintDisplayPart[] | undefined;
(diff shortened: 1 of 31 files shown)


DanielRosenwasser (approved):

DanielRosenwasser:
CC @mjbvz

src/services/inlayHints.ts
```diff
@@ -159,9 +159,11 @@ export function provideInlayHints(context: InlayHintsContext): InlayHint[] {
...
-        let hintText: string | InlayHintDisplayPart[] = `${isFirstVariadicArgument ? "..." : ""}${text}`;
+        let hintText = `${isFirstVariadicArgument ? "..." : ""}${text}`;
+        let displayParts: InlayHintDisplayPart[] | undefined;
         if (shouldUseInteractiveInlayHints(preferences)) {
-            hintText = [getNodeDisplayPart(hintText, parameter, sourceFile!), { text: ":" }];
+            displayParts = [getNodeDisplayPart(hintText, parameter, sourceFile!), { text: ":" }];
+            hintText = "";
         }
         else {
             hintText += ":";
```
jakebailey, line 169: If we don't want to leave this blank, we'd need to drop the reassignment in the first branch, and then pull the code out of the second branch to make it unconditional. I.e., add the prop only when requested and everything else stays the same.

Happy to do whatever. Not sure which is better.

MariaSolOs (approved):

PR #55278, opened 2023-08-05
Fix CFA of async immediately invoked function expressions

### LINKED ISSUE #30625: Type narrowing on object properties lost in async IIFE
**TypeScript Version:**  3.4.0-dev.20190327

**Search Terms:**
narrowing iife control flow

**Code**

```ts
declare const x: string | undefined;
declare const y: { z: string | undefined };
function needsString(it: string) { return it; }

function a() {
  if (!x) {
    throw new Error("Missing x");
  }
  const res1 = (() => needsString(x))(); // all good, per #8849
  const res2 = (async () => needsString(x))();  // ditto

  if (!y.z) {
    throw new Error("Missing z.")
  }

  const res3 = (() => needsString(y.z))();  // still good
  const res4 = (async () => needsString(y.z))();  // now things blow up  
}
```

**Expected behavior:**
The call to `needsString` in the expression for res4 succeeds, like the one in `res2` and `res3`.

**Actual behavior:**
Error for `needsString(y.z)`: Argument of type 'string | undefined' is not assignable to parameter of type 'string'. 

**Playground Link:** https://www.typescriptlang.org/play/index.html#src=declare%20const%20x%3A%20string%20%7C%20undefined%3B%0Adeclare%20const%20y%3A%20%7B%20z%3A%20string%20%7C%20undefined%20%7D%3B%0Afunction%20needsString(it%3A%20string)%20%7B%20return%20it%3B%20%7D%0A%0Afunction%20a()%20%7B%0A%20%20if%20(!x)%20%7B%0A%20%20%20%20throw%20new%20Error(%22Missing%20x%22)%3B%0A%20%20%7D%0A%20%20const%20res1%20%3D%20(()%20%3D%3E%20needsString(x))()%3B%20%2F%2F%20all%20good%2C%20per%20%238849%0A%20%20const%20res2%20%3D%20(async%20()%20%3D%3E%20needsString(x))()%3B%20%20%2F%2F%20ditto%0A%0A%20%20if%20(!y.z)%20%7B%0A%20%20%20%20throw%20new%20Error(%22Missing%20z.%22)%0A%20%20%7D%0A%0A%20%20const%20res3%20%3D%20(()%20%3D%3E%20needsString(y.z))()%3B%20%20%2F%2F%20still%20good%0A%20%20const%20res4%20%3D%20(async%20()%20%3D%3E%20needsString(y.z))()%3B%20%20%2F%2F%20now%20things%20blow%20up%20%20%0A%7D

**Related Issues:**
https://github.com/Microsoft/TypeScript/pull/8849, which I think was supposed to fix IIFE issues like this, but appears to have missed a case (or I'm missing something).

### LINKED ISSUE #38929: Inconsistent assignment analysis for async arrow IIFE's regarding class property declarations (sugar vs constructor assigning)
**TypeScript Version:**  4.0.0-dev.202000519

**Search Terms:** async IIFE Property is used before being assigned class declaration 

**Code**

This works fine:
```ts
class MyClass {
	public foo = 0;

	constructor() {
		(async () => this.foo + 5)();
	}
}
```

This, on the other hand, errors:
```ts
class MyClass {
	public foo: number;

	constructor() {
		this.foo = 0;
		(async () => this.foo + 5)(); // errors?
	}
}
```

These two code snippets should be treated the same because the first snippet is sugar for the second.

**Expected behavior:** `(async () => this.foo + 5)()` should not have the error "Property 'foo' is used before being assigned."

**Actual behavior:** It errors

**Playground Link:** [Here](https://www.typescriptlang.org/play/#code/MYGwhgzhAECyCeBhcVoG8CwAoAkABwFcAjEAS2GgDMB7a6AXmgAYBubbHYagOwgBcATgWB9qAgBQBKdBxzjI8bhSkMAfND4ALUhAB0NOgGpoAVklS2uAL7YbWbKEgwEyJwCYZuQiXJVaALmhuAgBbIgBTAUsOLl5BYVEJaUxcHC0dfVoGZkscOQUlaBV6dXS9A2hjMwtZWXFi0u1yrKrzSRZoAHpO6AB3MQBrGEpSbnCOOysgA)

**Related Issues:**

Zzzen:
Fixes #38929
Fixes #30625

### DIFF
--- a/src/compiler/utilities.ts
+++ b/src/compiler/utilities.ts
@@ -3087,6 +3087,14 @@ export function getImmediatelyInvokedFunctionExpression(func: Node): CallExpress
 }
 
 /** @internal */
+export function isAwaited(node: Node): boolean {
+    while(node.kind === SyntaxKind.ParenthesizedExpression) {
+        node = node.parent;
+    }
+    return node.kind === SyntaxKind.AwaitExpression;
+}
+
+/** @internal */
 export function isSuperOrSuperProperty(node: Node): node is SuperExpression | SuperProperty {
     return node.kind === SyntaxKind.SuperKeyword
         || isSuperProperty(node);
--- a/src/compiler/binder.ts
+++ b/src/compiler/binder.ts
@@ -166,6 +166,7 @@ import {
     isGlobalScopeAugmentation,
     isIdentifier,
     isIdentifierName,
+    isAwaited,
     isInJSFile,
     isInTopLevelContext,
     isJSDocConstructSignature,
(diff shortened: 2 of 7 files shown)


src/compiler/utilities.ts
```diff
@@ -3086,6 +3086,14 @@ export function getImmediatelyInvokedFunctionExpression(func: Node): CallExpress
     }
 }
 
+/** @internal */
+export function isAwaited(node: Node): boolean {
+    while(node.kind === SyntaxKind.ParenthesizedExpression) {
```
gabritto, line 3091: I think you should use the `walkUpOuterExpressions` function from `src/compiler/factory/utilities.ts` instead.

gabritto:
@typescript-bot pack this

gabritto:
If I understand correctly, what this PR does is that, for a "floating async" IIFE, the control flow nodes inside it are going to be a continuation from the control flow nodes immediately before it. I'll defer to @rbuckton to decide if that makes sense.

It might also be a good idea to add a test like the one suggested here: https://github.com/microsoft/TypeScript/issues/30625#issuecomment-477667719

Zzzen:
Tests added. In my opinion, https://github.com/microsoft/TypeScript/issues/30625#issuecomment-477667719 should error because the property is not initialized after the object is constructed

```ts
class Foo {
  prop: number;

  constructor() {
    (async () => (this.prop = await 1))();
  }
}

new Foo().prop.toString();  // ooops, Cannot read properties of undefined (reading 'toString') 
```

Issue #8775, opened 2016-05-24
typeof operator fails on use-before-defined block scoped variable.


sccolbert:
**TypeScript Version:** 

1.8.0

**Code**

``` ts
interface IThing<T> {
    owner: T;
}

var foo = {
    one: null as IThing<typeof foo>, // okay
}

let bar = {
    two: null as IThing<typeof bar>, // fail
}
```

**Expected behavior:**

The `let` definition should compile, since the `typeof` operator has no runtime effect.

**Actual behavior:** 

The `let` definition fails with a "used before declaration" error.

markwongsk:
Would this be a suitable bug for a first-timer?

RyanCavanaugh:
I imagine it will be pretty easy. There are lots of other places where we special-case use of an identifier inside a `typeof` so there are some examples to work from.

markwongsk:
I think this is a special-casing of `typeof`. For example, the example above works for a `var foo` declaration, but not a `let bar` declaration. Other manifestations of this problem include:
```
var baz = { 
    one: baz, // valid
};

let ack = { 
    one: ack, // fails
};

var a = [ a ]; // valid
let b = [ b ]; // fails
```

I think the reason for the differences is the following lines in `checker.ts` (revision 6e4dd752a161b745ef5b98e39b910ffe51632114, line 890):
```
                if (meaning & SymbolFlags.BlockScopedVariable) {
                    const exportOrLocalSymbol = getExportSymbolOfValueSymbolIfExported(result);
                    if (exportOrLocalSymbol.flags & SymbolFlags.BlockScopedVariable) {
                        checkResolvedBlockScopedVariable(exportOrLocalSymbol, errorLocation);
                    }
                }
```
Since the `let` causes the variable to be block-scoped, this eventually triggers the check `isImmediatelyUsedInInitializerOfBlockScopedVariable`. At this point, I'm wondering if this was an explicit design decision. It seems like under the ECMAScript specification, `baz` and `a` are valid whereas `ack` and `b` are not, which leads me to want to conclude that the current implementation is correct.

RyanCavanaugh:
The runtime behavior is mostly irrelevant to how the type system can behave. There's not anything preventing us from figuring what's going on in a block like this
```ts
let a: typeof b;
let b: string;
```

markwongsk:
[...]

PR #55283, opened 2023-08-06
Fixed a redundant used before defined error

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/8775


tests/cases/compiler/noUsedBeforeDefinedErrorInTypeContext.ts
```diff
@@ -0,0 +1,23 @@
...
+    one: {} as IThing<typeof foo>,
+}
+
+let baz = {
+    two: {} as IThing<typeof bar>,
+}
+
+let bar = {
+    three: {} as IThing<typeof bar>,
+}
```
Andarist, line 19: this isn't actually super useful as the result of this is an implicit any but that's a separate error category and the reported "no used before defined" was still redundant

jakebailey:
@typescript-bot test this
@typescript-bot test top100
@typescript-bot run dt

tests/cases/compiler/noUsedBeforeDefinedErrorInTypeContext.ts
sandersn: can you set `@strict: true` to show that the no implicit any errors are there?

sandersn (approved):

PR #55290, opened 2023-08-07
Allow `readonly` with `accessor`

### LINKED ISSUE #55289: Allow `readonly` with `accessor`
### 🔍 Search Terms

accessor
decorator
readonly

### ✅ Viability Checklist

### ⭐ Suggestion

Add `readonly` as a modifier for `accessor` fields.

### 📃 Motivating Example

Suppose you have an injection framework with `@provides`, `@consumes`, and `@inject`. Consider the example

```ts
class A {
  @provides(Number)
  readonly value: number;

  @inject([Number])
  accessor b = new B();

  constructor(value: number) {
    this.value = value;
  }
}

class B {
  @consumes(Number)
  readonly value: number;
}
```

Note because `@provides` is a field decorator, it cannot know when `value` is assigned unless it's in the initializer. As a consequence, `b` will not have `value` injected since `@provides` doesn't know when to reinject. This is fixed if you do

```ts
class A {
  @provides(Number)
  readonly accessor value: number;

  @inject([Number])
  accessor b = new B();

  constructor(value: number) {
    this.value = value;
  }
}

class B {
  @consumes(Number)
  readonly value: number;
}
```

### 💻 Use Cases

Accessor declarations that shouldn't be modified after being used in the constructor.
You can only mark a field with `@readonly` using JSDoc and just hope that other developers don't touch the field.
`@readonly` using JSDoc

jrandolf:
Fixes #55289

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -47436,9 +47436,6 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                         if (flags & ModifierFlags.Accessor) {
                             return grammarErrorOnNode(modifier, Diagnostics._0_modifier_already_seen, "accessor");
                         }
-                        else if (flags & ModifierFlags.Readonly) {
-                            return grammarErrorOnNode(modifier, Diagnostics._0_modifier_cannot_be_used_with_1_modifier, "accessor", "readonly");
-                        }
                         else if (flags & ModifierFlags.Ambient) {
                             return grammarErrorOnNode(modifier, Diagnostics._0_modifier_cannot_be_used_with_1_modifier, "accessor", "declare");
                         }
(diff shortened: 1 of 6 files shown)


jrandolf:

sandersn:
I couldn't tell from the issue whether it has an accepted solution. Is allowing `readonly` with `accessor` actually what we want to do? I think we need to agree on that back on the original issue.

jrandolf:
@sandersn Sorry Nathan, it seems the issue is on some kind of backlog. I'm not familiar with the process in TS, so I've just been waiting for some conclusion on your teams end (or whoever is responsible for triaging the issue above).

jakebailey:
The milestone isn't a problem, it's that it's not marked as "experience enhancement" i.e. it requires some discussion.

Issue #10564, opened 2016-08-26
Strange boolean-discriminant narrowing with strictNullChecks off


gcnew:
**TypeScript Version:**  nightly (2.1.0-dev.20160826)

**Code**

``` ts
type Result = { success: true }
            | { success: false, error: string }

function handleError(res: Result) {
    if (res.success === true) {
        return;
    }

    res.error; // OK
}

function handleError2(res: Result) {
    if (res.success !== false) {
        return;
    }

    res.error; // OK
}

function handleError3(res: Result) {
    if (res.success) {
        return;
    }

    res.error; // Property 'error' does not exist on type 'Result'
               // but should be OK
}
```

**Expected behavior:**
All three cases should behave the same.

RyanCavanaugh:
This is correct. We don't know that `res.success` isn't `undefined` or `null`, which would mean it's possibly not `false` and therefore not `{ success: false, error: string }`

gcnew:
@RyanCavanaugh While you are technically right, when strictNullChecks is off `null` and `undefined` are not modelled at all and such differences don't surface in other situations. For example `handleError` has exactly the same semantics but compiles just fine. I think the behaviour should be consistent or at least the error message should be more descriptive.

Kingwl:
what is the expected behavior?

RyanCavanaugh:
@DanielRosenwasser I have the same question

kevinder:
> This is correct. We don't know that res.success isn't undefined or null, which would mean it's possibly not false and therefore not { success: false, error: string }

I'm wondering why the `if (res.success === true) {` example above should compile then, because couldn't `success` still be undefined or null?

haltman-at:
Hey, I just wanted to point out that this issue now has additional relevance in TypeScript 3.6 due to the addition of the `Generator` type and the new `IteratorResult` type.  I was writing some code with generators, expecting I could use `while(!result.done)` to loop over the case where the generator was not done, but found in fact I had to use `while(result.done === false)`.  Might be a reason to revisit this?

Bernd-L:
I have experienced similar behaviour, and the following code helped me to work around this limitation:

(Assuming we have `foo` with the discriminating boolean-property "`discBool`" and the optional property "`optionalProperty`" which is only present in one of the two sub-types of `foo`'s super-type)

```TypeScript
switch (foo.discBool) {
  case false:
    doSomething(foo.optionalProperty);
}
```

while this code didn't work:

```TypeScript
[...]

PR #55291, opened 2023-08-07
Remove `Falsy` from `TypeFacts.Truthy`

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/10564

⚠️ this only affects `strictNullChecks: false`

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -1205,7 +1205,7 @@ export const enum TypeFacts {
     FalseStrictFacts = BaseBooleanStrictFacts | Falsy,
     FalseFacts = BaseBooleanFacts,
     TrueStrictFacts = BaseBooleanStrictFacts | Truthy,
-    TrueFacts = BaseBooleanFacts | Truthy,
+    TrueFacts = (BaseBooleanFacts & ~Falsy) | Truthy,
     SymbolStrictFacts = TypeofEQSymbol | TypeofNEString | TypeofNENumber | TypeofNEBigInt | TypeofNEBoolean | TypeofNEObject | TypeofNEFunction | TypeofNEHostObject | NEUndefined | NENull | NEUndefinedOrNull | Truthy,
     SymbolFacts = SymbolStrictFacts | EQUndefined | EQNull | EQUndefinedOrNull | Falsy,
     ObjectStrictFacts = TypeofEQObject | TypeofEQHostObject | TypeofNEString | TypeofNENumber | TypeofNEBigInt | TypeofNEBoolean | TypeofNESymbol | TypeofNEFunction | NEUndefined | NENull | NEUndefinedOrNull | Truthy,
--- a/tests/baselines/reference/conditionalOperatorConditionIsBooleanType.types
+++ b/tests/baselines/reference/conditionalOperatorConditionIsBooleanType.types
@@ -112,7 +112,7 @@ true ? exprString1 : exprBoolean1; // union
 //Cond is a boolean type expression
 !true ? exprAny1 : exprAny2;
 >!true ? exprAny1 : exprAny2 : any
->!true : boolean
+>!true : false
 >true : true
 >exprAny1 : any
 >exprAny2 : any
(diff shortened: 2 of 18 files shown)


src/compiler/checker.ts
```diff
@@ -1205,7 +1205,7 @@ export const enum TypeFacts {
     FalseStrictFacts = BaseBooleanStrictFacts | Falsy,
     FalseFacts = BaseBooleanFacts,
     TrueStrictFacts = BaseBooleanStrictFacts | Truthy,
-    TrueFacts = BaseBooleanFacts | Truthy,
+    TrueFacts = (BaseBooleanFacts & ~Falsy) | Truthy,
```
Andarist, line 1208: if this is too broad of a change then perhaps the fix (if it's even desired today) could be moved closer to this `narrowType` callback:
https://github.dev/microsoft/TypeScript/blob/e936eb13d2900f21d79553c32a704307c7ad03dd/src/compiler/checker.ts#L27121

tests/baselines/reference/conditionalOperatorConditionIsBooleanType.types
```diff
@@ -286,7 +286,7 @@ var resultIsString3 = null === undefined ? exprString1 : exprString2;
 var resultIsObject3 = true || false ? exprIsObject1 : exprIsObject2;
 >resultIsObject3 : Object
 >true || false ? exprIsObject1 : exprIsObject2 : Object
->true || false : boolean
+>true || false : true
```
RyanCavanaugh, line 289: This is the wrong result under `strictNullChecks: false`

tests/baselines/reference/controlFlowWhileStatement.types
```diff
@@ -158,7 +158,7 @@ function f() {
             break;
         }
         if (cond) {
->cond : true
+>cond : never
```
RyanCavanaugh, line 161: !

tests/baselines/reference/conditionalOperatorConditionIsBooleanType.types
Andarist: Could u ELI5 how to reason about `strictNullChecks: false` in situations like this?

RyanCavanaugh: Any value can be falsy due to being `null` / `undefined`

Andarist: I know that from the assignability PoV but that doesnt help me explain your expected result here. Shouldnt intrinsic value like `true` be non-falsy? 

Where is the line if the referenced issue got labeled as a bug?

RyanCavanaugh: A value of type `true` might be inhabited by `null`. Obviously the expression `true` itself isn't, but that's not really under analysis

Andarist: If the compiler has to always account for the fact that the value might be inhabited by `null` then wouldn't it mean that one of the examples from the referenced issue shouldn't typecheck (and it does typecheck today)?
```ts
type Result = { success: true }
            | { success: false, error: string }

function handleError2(res: Result) {
    if (res.success !== false) {
        return;
    }

    res.error; // OK, but shouldn't be?
}
```

I always thought that `strictNullChecks: false` means that things can be `null`/`undefined` at runtime but that the type-checker is meant to ignore this fact completely.

Andarist: Either way - I pushed out an improved (but not cleaned up, code-wise) version of the fix for the linked issue. Please take a look if those semantics look any better.

fatcerberus: > I always thought that strictNullChecks: false means that things can be null/undefined at runtime but that the type-checker is meant to ignore this fact completely.

This is a point of discussion that comes up sometimes on Discord - `strictNullChecks: false` is basically a completely different type checker.  The flag does a lot more than its name would suggest, changing inferences, narrowing, etc. in sometimes very unexpected ways, and all of it is more or less by design.

DanielRosenwasser:
@typescript-bot pack this
@typescript-bot test this
@typescript-bot test top300
@typescript-bot run dt

sandersn:
To help with PR housekeeping, I'm going to close this PR since it's pretty old now.
strictNullChecks: false is already a mess of inconsistent semantics that fewer and fewer people use. We definitely shouldn't change them.

PR #55299, opened 2023-08-07
Switch devcontainer to use Node.js 18.

DanielRosenwasser:
Also moved certain fields to `customizations`.


jakebailey:
I would honestly recommend using 20 for stack restarting; it was removed in 15 and readded in 19 hence why Volta was pinned back so long.

PR #55319, opened 2023-08-09
Don't allow mixed-enum-operand binary ops.

RyanCavanaugh:
This is a hacky implementation; do not merge

PR #55320, opened 2023-08-09
Use correct Node.flags instead of Node.kind in `isInsideAwaitableBody`

RyanCavanaugh:
Found with #55319


jakebailey (approved):
I'm not sure if I'm surprised this doesn't change any tests, or not surprised at all.

PR #53394, opened 2023-03-20
export TypingsInstaller from tsserverlibrary

zkat:
Fixes #53414 

This exports interfaces necessary for external implementers of ITypingsInstaller and extenders of TypingsInstaller.

PR #55326, opened 2023-08-09
Make tsserver and typingsInstaller thin wrappers around public API

jakebailey:
If you check out the tsconfigs for `tsserver` and `typescript` (previously `tsserverlibrary`), you'll note that they're the same! `typescript` is just a plain reexport for public use, meaning that `tsserver` is a strict superset of the public API and could be implemented as a bundle which just imports the API. We already do this for vscode.dev, which adds a web host; `tsserver` is what adds the node host. The same thing is true for `typingsInstaller` as well; we exported it in the public API in #53394.

This PR instead makes those projects import from the public API project. Then at bundle time, this import is replaced with a reference to a local `./typescript.js`. (In `--no-bundle`, this isn't done, and the original import still works.)

After this, we're down to only having copies of our code in `typescript.js` and `tsc.js`, where the latter is a significantly different bundle for performance reasons.

# Package size report

## Overall package size

| - | - | - | - | - |
| Packed | 5.51 MiB | 3.76 MiB | -1.74 MiB | -31.66% |
| Unpacked | 30.18 MiB | 20.36 MiB | -9.82 MiB | -32.55% |

## Files

| - | - | - | - | - |
| `lib/tsserver.js` | 8.19 MiB | 24.64 KiB | -8.16 MiB | -99.71% |
| `lib/typingsInstaller.js` | 1.67 MiB | 10.47 KiB | -1.66 MiB | -99.39% |


jakebailey:
Random comment so I can manually run the new perf tests.

jakebailey:
@typescript-bot pack this

jakebailey:
No perf impact, nice. (Was not expecting any, but still.)

weswigham (commented):
Should we do a `const {...} = ts.server;` at the top of the changed files, to keep the diff minimal and the accesses more static? Though I guess that won't work for types and we never added support for `import {...} = ts.server` iirc.

jakebailey:
Sure, let me give that a try, though I think most of these are actually types that change? Not sure if I can named import using an `import =` like that to grab the namespace...

weswigham (approved):
Either way this is fine, it'd probably just be nice if we didn't have to do the long `ts.server` everywhere in these files and could retain the short direct references.

jakebailey:
I couldn't figure out how to get named imports from a nested object like that, but was able to do something slightly different by reexporting the namespaces through the older files, which keeps the diff minimal.

This won't really work for ESM but that's another day.

jakebailey:
@typescript-bot pack this

jakebailey:
I can see from the benchmark that the last commit made tsserver 6% slower to start. I confirmed that locally too. Probably the extra copy of the server object that gets made.

I'm going to revert it and stick with the version that uses `ts.server.*` instead to prevent that slowdown, and come back to this later to see if I've missed a better way to write this.

Issue #18924, opened 2017-10-03
Allow Extract Constant in expression-bodied arrow functions


amcasey:
```ts
const f = () => 1 + [#|(2 + 3)|] + 4;
```

Expected:
```ts
const f = () => {
    const newLocal = (2 + 3);
    return 1 + newLocal + 4;
};
```

We can detect this scenario, but `ChangeTracker` doesn't support overlapping transformations (i.e. substituting `newLocal` into the expression and then putting the result into a return statement).

Issue #55348, opened 2023-08-13
"Add braces to arrow function" refactor triples some comments


Andarist:
### 🔎 Search Terms

refactor braces arrow comments

### 🕗 Version & Regression Information

- This is the behavior in every version I tried

### ⏯ Playground Link

https://www.typescriptlang.org/play?#code/MYewdgzgLgBFCm0YF4YAoCWAuGYCuAtgEbwBOAlCgHwwBQMMA9IzBCAfDKAQfTBjADUMAIwAGANxA

### 💻 Code

```ts
const test = (i: number) => 
  // some comm
  i + 10;
```

### 🙁 Actual behavior

```ts
const test = (i: number) => 
  // some comm
  {
        /* some comm*/
        return
            // some comm
            i + 10;
    };
```

Notice that this change also introduces an ASI problem.

### 🙂 Expected behavior

I think the comment should stick to the return statement so smth like:
```ts
const test = (i: number) => {
    // some comm
    return i + 10;            
};
```

PR #55349, opened 2023-08-13
Support extract to constant in blockless arrow functions

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/55323
closes https://github.com/microsoft/TypeScript/issues/18924


src/services/refactors/addOrRemoveBracesToArrowFunction.ts
```diff
@@ -89,6 +90,14 @@ function getRefactorActionsToRemoveFunctionBraces(context: RefactorContext): rea
     return emptyArray;
 }
 
+/** @internal */
+export function addBracesToArrowFunction(file: SourceFile, returnExpression: Expression, extraStatements: readonly Statement[] = []) {
```
Andarist, line 94: I think that exporting this broke `APILibCheck` test, cause now I get this baseline change:
<details>
<summary>baseline change: `tests/baselines/reference/APILibCheck.errors.txt`</summary>

```
tsserverlibrary.internal.d.ts(24934,143): error TS2307: Cannot find module '../_namespaces/ts' or its corresponding type declarations.
tsserverlibrary.internal.d.ts(25046,139): error TS2307: Cannot find module '../_namespaces/ts' or its corresponding type declarations.
typescript.internal.d.ts(19586,143): error TS2307: Cannot find module '../_namespaces/ts' or its corresponding type declarations.
typescript.internal.d.ts(19698,139): error TS2307: Cannot find module '../_namespaces/ts' or its corresponding type declarations.


==== index.ts (0 errors) ====
    import ts = require("typescript");
    import tsInternal = require("typescript-internal");
    import tsserverlibrary = require("tsserverlibrary");
    import tsserverlibraryInternal = require("tsserverlibrary-internal");
    
==== node_modules/typescript/package.json (0 errors) ====
    {
        "name": "typescript",
        "types": "/.ts/typescript.d.ts"
    }
    
==== node_modules/typescript-internal/package.json (0 errors) ====
    {
        "name": "typescript-internal",
        "types": "/.ts/typescript.internal.d.ts"
    }
    
==== node_modules/tsserverlibrary/package.json (0 errors) ====
    {
        "name": "tsserverlibrary",
        "types": "/.ts/tsserverlibrary.d.ts"
    }
    
==== node_modules/tsserverlibrary-internal/package.json (0 errors) ====
    {
        "name": "tsserverlibrary-internal",
        "types": "/.ts/tsserverlibrary.internal.d.ts"
    }
    
``` 
</details>

I'm not sure how to fix this though. My gut tells me that @jakebailey might know :P

src/services/refactors/extractSymbol.ts
```diff
@@ -1369,7 +1368,28 @@ function extractConstantInScope(
...
+        const newVariableDeclaration = factory.createVariableDeclaration(localNameText, /*exclamationToken*/ undefined, variableType, getSynthesizedDeepClone(initializer));
+        const newVariableStatement = factory.createVariableStatement(
+            /*modifiers*/ undefined,
+            factory.createVariableDeclarationList([newVariableDeclaration], NodeFlags.Const));
+
+        const returnExpression = visitNode(
+            scope.body,
+            function visitor(n: Node): VisitResult<Node> {
+                if (n === node) {
+                    return localReference;
```
Andarist, line 1383: I think that I should also copy comments from `n` to `localReference` here but [issues like this](https://github.com/microsoft/TypeScript/issues/55348) makes me believe that this might be slightly more complicated than just calling `copyComments(n, localReference)`. Any tips?

src/services/refactors/addOrRemoveBracesToArrowFunction.ts
jakebailey: You need to explicitly type the return type. Check the line of the output where the error is and you'll see the bad node from our very naive dts bundler.

Andarist: Thanks for the tip! I’ll apply the suggestion soon. I wonder though - shouldnt this just get stripped out entirely anyway since its marked as internal?

jakebailey: The bundler produces both public and internal outputs, since there are packages like byots and ts-expose-internals which need that. Before modules, they'd just build and disable stripInternal, but post modules, you have to use the bundler. But since the bundler is very naive, if we didn't verify that it worked with the internal exports, those users would end up with randomly broken files that could only be fixed upstream here.

DanielRosenwasser:
Can you add a test for an object literal returned from an arrow function (the case I just ran into)?

```ts
export interface Point {
    x: number;
    y: number;
}

export function double(points: Point[]) {
    return points.map(p => ({ x: p.x * 2, y: p.y * 2}));
}
```

Andarist:
@DanielRosenwasser sure thing, just pushed this extra test case out :)

PR #55360, opened 2023-08-14
Add additional information section to the bug report template

Andarist:
cc @jakebailey


jakebailey:
This seems okay to me, though the old template didn't have this section, right?

Andarist:
Ye, it didn’t but it was less awkward to add some extra info on the bottom because the whole thing was editable by the reported right from the start

jakebailey (approved):

Issue #55363, opened 2023-08-14
"Remove 'type' from import" produces invalid syntax


lgarron:
### 🔎 Search Terms

"Remove 'type' from import", IntelliSense, autocomplete, suggestion

### 🕗 Version & Regression Information

I'm seeing this in VSCode, with all extensions disabled:

Bundled TypeScript version: v5.1.0

Version: 1.81.0
Commit: 6445d93c81ebe42c4cbd7a60712e0b17d9463e97
Date: 2023-08-02T12:38:28.722Z (1 wk ago)
Electron: 22.3.18
ElectronBuildId: 22689846
Chromium: 108.0.5359.215
Node.js: 16.17.1
V8: 10.8.168.25-electron.0
OS: Darwin arm64 22.6.0

### ⏯ Playground Link

N/A

### 💻 Code

Use the following files:

```ts
// main.ts
import { TwistyAlgEditor, type TwistyPlayer } from "./other-file";

new TwistyPlayer();

// other-file.ts
export class TwistyAlgEditor {}
export class TwistyPlayer {}
```

Then attempt to select the `Remove 'type' from import of 'TwistyPlaver' from "cubing/twisty"` suggestion for `new TwistyPlayer`:

<img width="945" alt="Screenshot 2023-08-14 at 19 26 33" src="https://github.com/microsoft/TypeScript/assets/248078/e8f3a9a6-8484-4664-95b0-da8cae37d31d">

### 🙁 Actual behavior

The code is rewritten to:

```ts
import { TwistyAlgEditorTwistyPlayer,  } from "./other-file";

new TwistyPlayer();
```

### 🙂 Expected behavior

The code is rewritten to:

```ts
import { TwistyAlgEditor, TwistyPlayer } from "./other-file";

new TwistyPlayer();
```

lgarron:
Thanks for the fix! 😄

PR #55365, opened 2023-08-14
Fixed an issue with type-only import promoting

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/55363


src/services/codefixes/importFixes.ts
```diff
@@ -1329,15 +1329,17 @@ function promoteFromTypeOnly(changes: textChanges.ChangeTracker, aliasDeclaratio
             if (aliasDeclaration.isTypeOnly) {
                 const sortKind = OrganizeImports.detectImportSpecifierSorting(aliasDeclaration.parent.elements, preferences);
                 if (aliasDeclaration.parent.elements.length > 1 && sortKind) {
-                    changes.delete(sourceFile, aliasDeclaration);
                     const newSpecifier = factory.updateImportSpecifier(aliasDeclaration, /*isTypeOnly*/ false, aliasDeclaration.propertyName, aliasDeclaration.name);
                     const comparer = OrganizeImports.getOrganizeImportsComparer(preferences, sortKind === SortKind.CaseInsensitive);
```
Andarist, line 1333: I think that there is an argument to be made here that this codefix shouldn't change the position of the promoted import. This is a separate action. I left this as-is though, as I assume that this was implemented this way for a reason.

fatcerberus: I suppose it might be annoying to users if they have their imports already sorted such that type-only imports are grouped separately from regular ones; promoting a type import will leave it in the wrong position in the list.  Particularly if the user is relying on auto-imports which automatically puts them in the correct place, it'd be frustrating to have this one code fix that requires triggering Organize Imports manually when in every other case it's automatic.

jakebailey: The reason that it moves in the list currently is that TS's organize/auto-imports always sorts `type` imports at end, but otherwise they're in order. Of course, via #55269 and #52820, this will become even more configurable. So effectively, any operation that does anything to imports (adding, removing, changing modifiers) has to go through some sort of sorting code to determine whether or not the user opted into sorting something in a particular way. Thankfully, the main goal of that configurability is enabling people to stick type imports inline such that these modifier changes don't affect sorting at all, which is nice.

Andarist: > So effectively, any operation that does anything to imports (adding, removing, changing modifiers) has to go through some sort of sorting code to determine whether or not the user opted into sorting something in a particular way. 

Do you happen to know if renaming exported symbols also sorts the relevant imports that use them?

jakebailey: I'm pretty sure it doesn't, no; my understanding is that rename is more of an "in place transform" and it doesn't even check that the name you're giving it is syntactically valid.

iisaduan (approved):
I found a similar issue while working on #55269, and my fix was almost identical. Thanks!

jakebailey (approved):

Issue #55364, opened 2023-08-14
Template literal type resolution


Treen11:
### 🔎 Search Terms

mapped type template literal

### 🕗 Version & Regression Information

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about current version

### ⏯ Playground Link

https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgCoE8AOECydPIDeAUMsnAFzIDkAgtacgEZXUBCDAvscQCYQIANnCgoEAexABnMMhhUAPKmQQAHpBC8pNONWQAfGk2oA+ABSqqAAwAkhVJysBKZAF4TaLLnwBtVAF0eMC9kADEARjdkJRV1CE1tal0DI1MLazsHZzcPDGw8TD9AiWlZGHCqCKiYZAB6WuQAdwALdGRecQhtMGbgbWgocSgAfiCQ0IAmKJi1DS0dPUNqY3NLZFt7Rxd3T3zfDaz-Ooa4TSbW9s7tAFs4AGtQAHNkHpRRMABXKBAXkLgXiDXTDCSDIQTASBQOCCOTAVTICGjEoyOQTSpTVxyIA

### 💻 Code

The unexpected error is marked by a comment in the code bellow. A second comment highlights a case, where this issue does not occur, even though the two cases should be identical, as far as i understand. Thank you @mkantor for the minimalistic example. The issue has been discussed on the TS discord in various channels. The original discussion was an "ask for help" here - https://discord.com/channels/508357248330760243/1139519033809575966/1139519033809575966

```ts
interface TypeMap {
  a: 'A'
  b: 'B'
}

declare const f: <T extends 'a' | 'b'>(x: `${T}`) => TypeMap[T]

type F1 = <T extends 'a' | 'b'>(x: `${T}`) => TypeMap[T]
const f1: F1 = f // why does this error?

type F2 = <T extends 'a' | 'b'>(x: `${T}`) => TypeMap[`${T}`] // and why does making the return type a template literal fix it?
const f2: F2 = f
```

### 🙁 Actual behavior

On line 9 occurs - 
```
Type '<T extends "a" | "b">(x: `${T}`) => TypeMap[T]' is not assignable to type 'F1'.
  Type 'TypeMap[`${T}`]' is not assignable to type 'TypeMap[T]'.
    Type 'TypeMap[`${T}`]' is not assignable to type 'never'.
      Type 'string' is not assignable to type 'never'.
[...]

PR #55371, opened 2023-08-14
Reduce template literal types with a single placeholder and no extra texts

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/55364
fixes https://github.com/microsoft/TypeScript/issues/56659


gabritto:
@typescript-bot run DT
@typescript-bot user test this
@typescript-bot test top100
@typescript-bot perf test this

gabritto:
@typescript-bot run DT
@typescript-bot test top100

src/compiler/checker.ts
```diff
@@ -22051,6 +22051,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     // For example, `foo-${number}` is related to `foo-${string}` even though number isn't related to string.
                     instantiateType(source, reportUnreliableMapper);
                 }
+                if (isTemplateLiteralTypeWithSinglePlaceholderAndNoExtraTexts(target) && isTypeAssignableTo(source, stringType) && isTypeAssignableTo(target.types[0], stringType)) {
```
gabritto, line 22054: Why do we need to check that `source` is assignable to `string`?

src/compiler/checker.ts
```diff
@@ -22094,6 +22097,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             }
             else if (sourceFlags & TypeFlags.TemplateLiteral && !(targetFlags & TypeFlags.Object)) {
                 if (!(targetFlags & TypeFlags.TemplateLiteral)) {
+                    if (isTemplateLiteralTypeWithSinglePlaceholderAndNoExtraTexts(source) && isTypeAssignableTo(source.types[0], stringType) && isTypeAssignableTo(target, stringType)) {
```
gabritto, line 22100: Same here, why do we need to check that `target` is assignable to `string`? I'm sure I'm just missing something 😅

src/compiler/checker.ts
Andarist: My reasoning is that we can relate `T` and `${T}` only if both are actually strings. Template literal placeholders might be non-strings so without that we'd accidentally allow matching a string with a number and stuff like that.

gabritto:
@typescript-bot pack this

src/compiler/checker.ts
gabritto: Ok, so we have source `${S}` and target `T`, and I can see why we check that `S` <: `string`. But in that case, isn't `${S}` the same as `S`?

src/compiler/checker.ts
```diff
@@ -22051,6 +22051,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     // For example, `foo-${number}` is related to `foo-${string}` even though number isn't related to string.
                     instantiateType(source, reportUnreliableMapper);
                 }
```
gabritto, line 22053: This is not specific to the PR, but how do we reach this point of the code in the example in the bug, when the return types are written without using a template literal type, e.g. `TypeMap[T0]`?
Also, this function in the new test:
```
function f3<T3 extends "a" | "b">(x: T3) {
    const test1: `${T3}` = x    
    const test2: T3 = "" as `${T3}`;    
}
```
Already works without those changes. So I guess I'm having a hard time figuring out how the bug arises and how changes in the PR are related to the bug, because it seems we already have some simplification going on for template literal types that look like `${T}`.

Andarist: > This is not specific to the PR, but how do we reach this point of the code in the example in the bug, when the return types are written without using a template literal type, e.g. TypeMap[T0]?

This is an interesting thing! And I don't really understand its details in depth 😅 

The source signature here goes through `instantiateSignatureInContextOf` and its result is being compared against the target. The template literal type gets inferred there through `inferToTemplateLiteralType` that gets invoked through `applyToParameterTypes`.

At that point, inference source and inference target are both template literal types. `inferTypesFromTemplateLiteralType` assesses that texts are the same and "converts" the types in the source using `getStringLikeTypeForType`. That wraps back `T` into a template literal type since it's not string-like.

So we end up with a template literal type as a single match returned by `inferTypesFromTemplateLiteralType` and the target (in `inferToTemplateLiteralType`) is still a template literal type as well. The match represents a placeholder match though and we loop through the types in the target, "unwrapping" `T` from the template literal type there and thus we end up inferring a template literal type (the placeholder match) into `T`.

---

> Already works without those changes. So I guess I'm having a hard time figuring out how the bug arises and how changes in the PR are related to the bug because it seems we already have some simplification going on for template literal types that look like ${T}.

Both of those assignments **don't** work right now: [TS playground](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABMAzAHgCosQUwB5Q5gAmAzogEQCGFiAPpQEYUB8AFHgFyJYCUiAbwBQiUYggJSURISkBGbgAMAJAKwBfRYgC8iPGMQixEsFJk4pAJm5YdlWlXIq1KTQG4DQ9UA).

gabritto: Yeah, I mixed up my playgrounds again 😅
Ok, it seems to me like we're perhaps doing some unnecessary wrapping. Maybe we could check if a type parameter's constraint is string like in `getStringLikeTypeForType` and not wrap it in that case? If we did that, would we still need the changes in this PR?

Again, I'm not sure if I'm missing something and there's a reason why we do things the way they are right now, but I find it confusing.

Andarist: > If we did that, would we still need the changes in this PR?
 
I'm not entirely sure why the wrapping happens there but yes - if we'd remove it for some cases then it's likely that the issue at hand would be fixed. I considered that initially but I chose those other changes because they are also shipping what we have mentioned above:
```ts
function f3<T3 extends "a" | "b">(x: T3) {
    const test1: `${T3}` = x    
    const test2: T3 = "" as `${T3}`;    
}
```

So if you'd like me to explore that other route - that would "break" this test case.

gabritto: Ok, so I think this PR is probably good then. Maybe we should have both changes. In any case, I still have that question about whether ``` `${S}` ``` is the same as `S` when `S` is assignable to string...

weswigham (changes requested):
While this might be a true observation - `${T}` is identical to `T` when `T` is only string-like - this isn't how we should implement it, imo. Rather than messing with type relations, this should be true on _construction_, eg, inside `getTemplateLiteralType` we should check for and produce this simplification, just so the simplifiable type identity never even exists inside the checker. A "late" simplification like this we should generally only do if it's too costly/circularity inducing to check upfront, but I don't think that'll be the case here (and, broadly speaking, could be done in `getSimplifiedType` rather than as a new relationship rule).

As an aside that won't really matter once that's done, since I do think we can do that, there's also an inference issue here, which is probably the better, broader fix for the original problem (which can probably be rewritten to have non-string constraints).  When we see `${T}` and infer to `${U}`, the texts all match - ideally I'd think we want to infer `T` to `U` as a better match, no template wrapper required.

Andarist:
I pushed out the requested changes

gabritto:
@typescript-bot run DT
@typescript-bot test top200
@typescript-bot pack this

src/compiler/checker.ts
```diff
@@ -17937,6 +17937,16 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         if (contains(types, wildcardType)) {
             return wildcardType;
         }
+        if (
+            texts.length === 2 && texts[0] === "" && texts[1] === ""
+            // literals (including string enums) are stringified below
+            && !(types[0].flags & TypeFlags.Literal)
+            // infer T extends StringLike can't be unwrapped eagerly
```
gabritto, line 17944: Do you have an example of when this matters? I.e. an example of why we shouldn't unwrapp `infer T`?

gabritto:
@Andarist do you intend to fix the inference problem as well?
Also, I'm wondering if inference is the only place where we construct types like `${T}` where `T` is stringlike... If it is, fixing inference would mean this PR is not needed, but otherwise this PR is still needed.

src/compiler/checker.ts
Andarist: it's mainly because of how string enums might be "stringified" by template literal types
```ts
const enum StringLiteralEnum { Zero = "0", True = "true", False = "false", Undefined = "undefined", Null = "null" }

type TStringLiteralEnum0 = "0" extends `${infer T extends StringLiteralEnum}` ? T : never;
//   ^? type TStringLiteralEnum0 = StringLiteralEnum.Zero
type TStringLiteralEnum0Unwrapped = "0" extends infer T extends StringLiteralEnum ? T : never;
//   ^? type TStringLiteralEnum0Unwrapped = never
```

Andarist:
> @Andarist do you intend to fix the inference problem as well? (see https://github.com/microsoft/TypeScript/issues/56659#issuecomment-1910964556).

I can investigate this

> Also, I'm wondering if inference is the only place where we construct types like ${T} where T is stringlike... If it is, fixing inference would mean this PR is not needed, but otherwise this PR is still needed.

Users might also construct such types and it makes sense to allow those:
```ts
function f3<T3 extends "a" | "b">(x: T3) {
    const test1: `${T3}` = x    
    const test2: T3 = "" as `${T3}`;    
}
```

To allow this we need to either do this simplification or mess with the relationship checking (what I had here before).

gabritto (approved):

gabritto:
> > @Andarist do you intend to fix the inference problem as well? (see [#56659 (comment)](https://github.com/microsoft/TypeScript/issues/56659#issuecomment-1910964556)).
> 
> I can investigate this
> 
> > Also, I'm wondering if inference is the only place where we construct types like ${T} where T is stringlike... If it is, fixing inference would mean this PR is not needed, but otherwise this PR is still needed.
> 
> Users might also construct such types and it makes sense to allow those:
> 
> ```ts
> function f3<T3 extends "a" | "b">(x: T3) {
>     const test1: `${T3}` = x    
>     const test2: T3 = "" as `${T3}`;    
> }
> ```
> 
> To allow this we need to either do this simplification or mess with the relationship checking (what I had here before).

Right, I remember now. I think you mentioned that before but I had forgotten, sorry.

gabritto:
@Andarist can you update this with latest main so I can merge?

Andarist:
@gabritto with pleasure! done :)

ahejlsberg:
This PR is effectively a breaking change as demonstrated in #58687. I'm not sure I like it.

Andarist:
Note that this PR started as a rule in type relationships, it was requested to make it unwrap those eagerly instead

gabritto:
I'll revert this.

Andarist:
@gabritto i have proposed fix in the works, i should be able to post it soon

Issue #55369, opened 2023-08-14
New Template Literal Flags are confused for `ContainsInvalidEscape` causing tagged template expressions to be unnecessarily transpiled


engelsdamien:
### 🔎 Search Terms

In [`processTaggedTemplateExpression`](https://github.com/microsoft/TypeScript/blob/main/src/compiler/transformers/taggedTemplate.ts#L55-L57) the code checks whether the tagged template expression using the [`hasInvalidEscape` function](
https://github.com/microsoft/TypeScript/blob/main/src/compiler/utilities.ts#L5805). This was done in https://github.com/microsoft/TypeScript/issues/12700 to support a new feature in tagged template literals allowing for invalid escape sequences.

Unfortunately, the function was written in a way that relies on implicit from `number` -> `boolean` instead of masking the `templateFlags` field:

```typescript
/** @internal */
export function hasInvalidEscape(template: TemplateLiteral): boolean {
    return template && !!(isNoSubstitutionTemplateLiteral(template)
        ? template.templateFlags
        : (template.head.templateFlags || some(template.templateSpans, span => !!span.literal.templateFlags)));
}
```

This means the code assumes any template literal with non-zero flags contains an invalid escape. While this was true in the past, [new flags were introduced in 5.1](https://github.com/microsoft/TypeScript/blob/main/src/compiler/types.ts#L2828) breaking the logic of the code. 

Now all tagged templates containing unicode escape sequences now get transpiled when they don't need to, causing a code size increase.

### 🕗 Version & Regression Information

This changed between versions 5.0.4 and 5.1.6

v5.0.4: https://www.typescriptlang.org/play?target=3&jsx=0&module=1&ts=5.0.4#code/GYVwdgxgLglg9mABMOcA8AVAfACgIYBciGAlAN4BOAplCBUngNwC+AUBAgM5SJ6IC8yVAAMAOiAAMUicMbsuPAEYChcYXllA 

v5.1.6: https://www.typescriptlang.org/play?target=3&jsx=0&module=1&ts=5.1.6#code/GYVwdgxgLglg9mABMOcA8AVAfACgIYBciGAlAN4BOAplCBUngNwC+AUBAgM5SJ6IC8yVAAMAOiAAMUicMbsuPAEYChcYXllA

### ⏯ Playground Link

[...]

PR #51837, opened 2022-12-09
fix(396): (Dis)allow Octal and Bad Escape Sequences in String and (Tagged) Template Literals

graphemecluster:
#23801 makes bad escape sequences in tagged template allowable, but only in certain cases. #41030 aims to fix this but was closed.
In contrast, octal escape sequences in string literals (as in #396) should be disallowed.

Fully fixes #12700
Fixes #396
Fixes #39038
Fixes #39715
Fixes #42887
Fixes #39563 (#39698 attempts to fix this but was closed)

This PR currently disallow octal literals and escape sequences **regardless of the target, strict mode or not** as the Team is looking for simplicity (Refer to https://github.com/microsoft/TypeScript/issues/396#issuecomment-853564421).

PR #55373, opened 2023-08-14
Fixed `hasInvalidEscape` implementation

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/55369
regression from https://github.com/microsoft/TypeScript/pull/51837 , cc @graphemecluster


src/compiler/utilities.ts
```diff
@@ -5804,8 +5804,8 @@ function escapeTemplateSubstitution(str: string): string {
 /** @internal */
 export function hasInvalidEscape(template: TemplateLiteral): boolean {
     return template && !!(isNoSubstitutionTemplateLiteral(template)
-        ? template.templateFlags
-        : (template.head.templateFlags || some(template.templateSpans, span => !!span.literal.templateFlags)));
+        ? !!((template.templateFlags || 0) & TokenFlags.ContainsInvalidEscape)
```
jakebailey, line 5807: This seems super noisy; can we pull this repeated bit out into a function?

Andarist: ye, i had such a plan - pushed it out now 👍

graphemecluster:
Oh, great that you spotted it – I overlooked that method.

jakebailey:
I'm thinking this one deserves a backport; @DanielRosenwasser thoughts?

jakebailey:
Hm, no, this was in 5.1; I misremembered which release we're currently on. Though, maybe it's worth doing anyway.

jakebailey (approved):

jakebailey:
I'm also silly and didn't realize that the result of this bug is just "the generated code is worse", so, nevermind altogether then 😄

PR #55332, opened 2023-08-10
Organize/consolidate inlay hint tests

MariaSolOs:
To make #55141 easier to test, I first wanted to reorganize the existing test suite into a smaller set of files with more descriptive file names.

PR #55379, opened 2023-08-14
Fix unused baseline failure on main

jakebailey:
#55332 was merged while still red, so main is failing as this baseline is unused.


MariaSolOs:
My bad, I was about to revert the changes. I'm so sorry, I didn't see the failure and I thought I wouldn't be able to merge without all tests passing.

MariaSolOs (approved):

jakebailey:
We unfortunately don't have that particular flag enabled, though with the new GitHub protection rules that just came out of beta, maybe we can enable some without enabling the "undesirable" ones.

PR #53674, opened 2023-04-05
Fix crash in parseJsxChild found by fuzzing

jakebailey:
Found this case in fuzzing.

This fix is basically the same as #52450. I didn't change this location because I thought it couldn't be hit, but, I was wrong!

The crash already occurs before 5.0, but the assert moved from one place to another in #52667. So this isn't a 5.0 regression, per se.

PR #52450, opened 2023-01-27
Prevent valid JSX from being seen as the start of a generic arrow function, fix crashes

jakebailey:
Fixes #52449

The arrow function lookahead was too overzealous in determining what was "definitely" an arrow function. Switching a "true" to an "unknown" result fixes the issue.

~But, I am suspicious of this because there is no error saying that the variable `const` isn't found (unlike `T`).~

PR #55381, opened 2023-08-15
Prevent detached diagnostics from running off the end of the file

### LINKED ISSUE #55296: `attachFileToDiagnostics`: Debug Failure crash
### 🔎 Search Terms

Hi TypeScript team,

In an ongoing effort to secure the JS/TS ecosystem, we integrated the TypeScript project into [Google's OSS-Fuzz project recently](https://github.com/google/oss-fuzz/tree/92cbef824dc66722e3e50eca0069e2461f6d6b91/projects/typescript) with our JS/TS fuzzer [Jazzer.js](https://github.com/CodeIntelligenceTesting/jazzer.js).

Looking at the reports, it seems we were able to find another reproducer for what seems to be related to #46364, #44051, and #43832. 

This bug has been found when building TypeScript from the main branch and running the `fuzz_compiler.js` [link](https://github.com/google/oss-fuzz/blob/92cbef824dc66722e3e50eca0069e2461f6d6b91/projects/typescript/fuzz_compiler.js) on it. 

I attached a PoC below that isolates the issue to the call to `ts.createSourceFile(fileName, fileContent, ts.ScriptTarget.Latest);`.

Happy to answer any questions and also discuss the used APIs.

- This is a crash

### ⏯ Playground Link

_No response_

### 💻 Code

Here's a quick JS reproducer.

```js
const fs = require("fs");
const ts = require("typescript");

const ignored = [
  "is not a function"
];


class StringConsumer {
  constructor(filename) {
    if (filename) {

      this.data = fs.readFileSync(filename); // read file into byte buffer
    } else {
      this.data = Buffer.from([
        0x75, 0x40, 0x00, 0x6d, 0xc0, 0xc0, 0xbc, 0x28,
        0xfb, 0x70, 0xdb, 0xdb, 0xfb, 0x6d, 0x00, 0xdb,
        0x6f, 0x72, 0x74, 0x28, 0x40, 0xbc, 0xdb, 0xdb,
        0x69, 0x6d, 0x70, 0x6f, 0x72, 0x74, 0x28, 0x6f,
        0x62, 0x6a, 0x75, 0x32, 0x63, 0x37, 0x37, 0x2c
      ]);
    }
    this.dataPtr = 0; // start index
    this._remainingBytes = this.data.length;
  }

  consumeString(maxLength, encoding = "ascii") {
    if (maxLength < 0) throw new Error("maxLength must be non-negative");
    let result;
    const arrayLength = Math.min(maxLength, this._remainingBytes);

    result = this.data.toString(encoding, this.dataPtr, this.dataPtr + arrayLength);

    this.dataPtr += arrayLength;
    this._remainingBytes -= arrayLength;

    return result;
  }
}

try {
  const provider = new StringConsumer();
  const fileName = provider.consumeString(10) + ".ts";
  const filePath = provider.consumeString(10) + "/" + fileName;
  const fileContent = provider.consumeString(100);
  console.log("fileName", fileName);
  console.log("filePath", filePath);
  console.log("fileContent", fileContent);

  ts.createSourceFile(fileName, fileContent, ts.ScriptTarget.Latest);
} catch (error) {
  if (!ignoredError(error)) {
    throw error;
  }
}

function ignoredError(error) {
  return !!ignored.find((message) => error.message.indexOf(message) !== -1);
}

```

### 🙁 Actual behavior

```shell
node poc.js
fileName u@m@@<({p.ts
filePath [[{m[ort(/u@m@@<({p.ts
fileContent @<[[import(obju2c77,
/Users/ci/Git/work/oss-fuzz-onboarding-projects/TypeScript/poc.js:53
    throw error;
    ^

Error: Debug Failure. Expected 21 <= 20
    at attachFileToDiagnostic (/Users/ci/Git/work/oss-fuzz-onboarding-projects/TypeScript/node_modules/typescript/lib/typescript.js:17846:17)
    at attachFileToDiagnostics (/Users/ci/Git/work/oss-fuzz-onboarding-projects/TypeScript/node_modules/typescript/lib/typescript.js:17858:36)
    at parseSourceFileWorker (/Users/ci/Git/work/oss-fuzz-onboarding-projects/TypeScript/node_modules/typescript/lib/typescript.js:28877:41)
    at Object.parseSourceFile (/Users/ci/Git/work/oss-fuzz-onboarding-projects/TypeScript/node_modules/typescript/lib/typescript.js:28700:26)
    at Object.createSourceFile (/Users/ci/Git/work/oss-fuzz-onboarding-projects/TypeScript/node_modules/typescript/lib/typescript.js:27885:23)
    at Object.<anonymous> (/Users/ci/Git/work/oss-fuzz-onboarding-projects/TypeScript/poc.js:50:6)
    at Module._compile (node:internal/modules/cjs/loader:1254:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1308:10)
    at Module.load (node:internal/modules/cjs/loader:1117:32)
    at Module._load (node:internal/modules/cjs/loader:958:12)

Node.js v18.16.0
```

### 🙂 Expected behavior

I assume not crash the Node.js runtime as is with the debug assertion?

jakebailey:
Fixes #55296

This is similar to #52450 and #53674; we can create diagnostics with hardcoded range lengths, but if the parse error starts at the end of the file, we'll crash.

I could band-aid this by special casing the one crash, but as you can see in the change, we have quite a few other places which assume it's okay to use a `length=1` diagnostic, when it's possible that it may not work.

So, just set the length to the maximum available (or zero) when this happens.

To make this more obvious, I've restructured things a little so that we check the ranges earlier. The stack trace in #55296 happens late because the diagnostics weren't checked until they were finally attached to the file, after parse was complete. It's possible that I could find a different way to do this that doesn't look so noisy, but I feel like this is more robust.

Realistically, I could just move `createDetachedDiagnostic` into `parser.ts`; it's only used there so that'd avoid some parameters. But, all of its related helpers are still in this utilities file.

### DIFF
--- a/src/compiler/parser.ts
+++ b/src/compiler/parser.ts
@@ -1803,7 +1803,7 @@ namespace Parser {
         return sourceFile;
 
         function reportPragmaDiagnostic(pos: number, end: number, diagnostic: DiagnosticMessage) {
-            parseDiagnostics.push(createDetachedDiagnostic(fileName, pos, end, diagnostic));
+            parseDiagnostics.push(createDetachedDiagnostic(fileName, sourceText, pos, end, diagnostic));
         }
     }
 
@@ -2118,7 +2118,7 @@ namespace Parser {
         const lastError = lastOrUndefined(parseDiagnostics);
         let result: DiagnosticWithDetachedLocation | undefined;
         if (!lastError || start !== lastError.start) {
-            result = createDetachedDiagnostic(fileName, start, length, message, ...args);
+            result = createDetachedDiagnostic(fileName, sourceText, start, length, message, ...args);
             parseDiagnostics.push(result);
         }
 
@@ -2470,7 +2470,7 @@ namespace Parser {
         if (lastError) {
             addRelatedInfo(
                 lastError,
-                createDetachedDiagnostic(fileName, openPosition, 1, Diagnostics.The_parser_expected_to_find_a_1_to_match_the_0_token_here, tokenToString(openKind), tokenToString(closeKind))
+                createDetachedDiagnostic(fileName, sourceText, openPosition, 1, Diagnostics.The_parser_expected_to_find_a_1_to_match_the_0_token_here, tokenToString(openKind), tokenToString(closeKind))
             );
         }
     }
@@ -4486,7 +4486,7 @@ namespace Parser {
             if (lastError && lastError.code === Diagnostics._0_expected.code) {
                 addRelatedInfo(
                     lastError,
-                    createDetachedDiagnostic(fileName, openBracePosition, 1, Diagnostics.The_parser_expected_to_find_a_1_to_match_the_0_token_here, "{", "}")
+                    createDetachedDiagnostic(fileName, sourceText, openBracePosition, 1, Diagnostics.The_parser_expected_to_find_a_1_to_match_the_0_token_here, "{", "}")
                 );
             }
         }
@@ -8326,7 +8326,7 @@ namespace Parser {
                 if (lastError && lastError.code === Diagnostics._0_expected.code) {
                     addRelatedInfo(
                         lastError,
-                        createDetachedDiagnostic(fileName, openBracePosition, 1, Diagnostics.The_parser_expected_to_find_a_1_to_match_the_0_token_here, "{", "}")
+                        createDetachedDiagnostic(fileName, sourceText, openBracePosition, 1, Diagnostics.The_parser_expected_to_find_a_1_to_match_the_0_token_here, "{", "}")
                     );
                 }
             }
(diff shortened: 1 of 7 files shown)


src/compiler/utilities.ts
```diff
@@ -8152,8 +8149,12 @@ export function getLocaleSpecificMessage(message: DiagnosticMessage) {
 }
 
 /** @internal */
-export function createDetachedDiagnostic(fileName: string, start: number, length: number, message: DiagnosticMessage, ...args: DiagnosticArguments): DiagnosticWithDetachedLocation {
-    assertDiagnosticLocation(/*file*/ undefined, start, length);
+export function createDetachedDiagnostic(fileName: string, sourceText: string, start: number, length: number, message: DiagnosticMessage, ...args: DiagnosticArguments): DiagnosticWithDetachedLocation {
+    if ((start + length) > sourceText.length) {
+        length = sourceText.length - start;
+    }
```
jakebailey, line 8155: This is the actual fix.

src/compiler/parser.ts
```diff
@@ -2470,7 +2470,7 @@ namespace Parser {
         if (lastError) {
             addRelatedInfo(
                 lastError,
-                createDetachedDiagnostic(fileName, openPosition, 1, Diagnostics.The_parser_expected_to_find_a_1_to_match_the_0_token_here, tokenToString(openKind), tokenToString(closeKind))
+                createDetachedDiagnostic(fileName, sourceText, openPosition, 1, Diagnostics.The_parser_expected_to_find_a_1_to_match_the_0_token_here, tokenToString(openKind), tokenToString(closeKind))
```
jakebailey, line 2473: This set of three detached diagnostics are the trouble; `openPosition` can totally be at the end of the file.

andrewbranch (approved):

Issue #55383, opened 2023-08-15
Generic indexed access on a substitution type with an array constraint resolves incorrectly


Andarist:
### 🔎 Search Terms

indexed access substitution type array constraint

### 🕗 Version & Regression Information

- This is the behavior in every version I tried

### ⏯ Playground Link

https://www.typescriptlang.org/play?#code/C4TwDgpgBAMglsCAnAhgGwCrmgXigZ2CTgDsBzKAHyhIFcBbAI2SqkYHt20IUSBuAFChIUAGrpaEAOIRgiJAB5xaSVhEQAHohIATfLATJ0a3AfnHsAPih4AFAEob15auyCh2KAGUAxgAsIehQbb39AlABBJFQQVl8AoIB5RgArCB9gVnhzTDcPEXjwqJiQ4pQQBUKk1PTM6irI6PKsw1RcyEtBYWgG5LSMkIBvASgoAG0AawgQAC4CIlIyAF053pqB+rCgstjqbKN2iEEAX3duqABJEgAzZBcIfAAxJHZ6BsqoTW09UISUazwXk+Wggun0+zaJhGUAA-GIJNJZPJKpZoatgd99A0dtC4cNRgTxgBpKCkKBTEDsa7eFaXG53BFPF5vLYoSpjIlLTrQ0bHNHeDGgn5rfrAXFQfGE4mkkjk6ZUmlzK63JD3Jmvd5eDlcwSEvmjOYkCAAN2QgiAA

### 💻 Code

```ts
type LiteralType = string | number | boolean;
type ValueGetter<ValueType extends LiteralType = LiteralType> = () => ValueType;

type Schema = SchemaArray | SchemaObject | LiteralType;
type SchemaArray = Array<SchemaObject | SchemaArray | LiteralType>;
type SchemaObject = {
  [key: string]: SchemaObject | SchemaArray | LiteralType;
};

type InferValuesFromSchema<S extends Schema> = S extends LiteralType
  ? ValueGetter<S>
  : S extends SchemaArray
  ? {
      [K in keyof S]: InferValuesFromSchema<S[K]>;
    }
  : S extends SchemaObject
  ? {
      [K in keyof S]: InferValuesFromSchema<S[K]>;
    }
  : never;
```

### 🙁 Actual behavior

an error is reported:
```
Type 'S[K]' does not satisfy the constraint 'Schema'.
  Type 'S[keyof S]' is not assignable to type 'Schema'.
[...]

PR #55386, opened 2023-08-15
Add numeric constraints to type parameters of mapped types with narrowed down array constraints

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/55383

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -15580,7 +15580,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 if (getTypeParameterFromMappedType(mappedType) === getActualTypeVariable(type)) {
                     const typeParameter = getHomomorphicTypeVariable(mappedType);
                     if (typeParameter) {
-                        const constraint = getConstraintOfTypeParameter(typeParameter);
+                        const constraint = getConstraintOfTypeParameter(getConditionalFlowTypeOfType(typeParameter, parent));
                         if (constraint && everyType(constraint, isArrayOrTupleType)) {
                             constraints = append(constraints, getUnionType([numberType, numericStringType]));
                         }
--- a/tests/baselines/reference/numericStringLiteralTypes2.symbols
+++ b/tests/baselines/reference/numericStringLiteralTypes2.symbols
@@ -1,0 +1,76 @@
+//// [tests/cases/conformance/types/literal/numericStringLiteralTypes2.ts] ////
+
+=== numericStringLiteralTypes2.ts ===
+// https://github.com/microsoft/TypeScript/issues/55383
+
+type LiteralType = string | number | boolean;
+>LiteralType : Symbol(LiteralType, Decl(numericStringLiteralTypes2.ts, 0, 0))
+
+type ValueGetter<ValueType extends LiteralType = LiteralType> = () => ValueType;
+>ValueGetter : Symbol(ValueGetter, Decl(numericStringLiteralTypes2.ts, 2, 45))
+>ValueType : Symbol(ValueType, Decl(numericStringLiteralTypes2.ts, 3, 17))
+>LiteralType : Symbol(LiteralType, Decl(numericStringLiteralTypes2.ts, 0, 0))
+>LiteralType : Symbol(LiteralType, Decl(numericStringLiteralTypes2.ts, 0, 0))
+>ValueType : Symbol(ValueType, Decl(numericStringLiteralTypes2.ts, 3, 17))
+
+type Schema = SchemaArray | SchemaObject | LiteralType;
+>Schema : Symbol(Schema, Decl(numericStringLiteralTypes2.ts, 3, 80))
+>SchemaArray : Symbol(SchemaArray, Decl(numericStringLiteralTypes2.ts, 5, 55))
+>SchemaObject : Symbol(SchemaObject, Decl(numericStringLiteralTypes2.ts, 6, 67))
+>LiteralType : Symbol(LiteralType, Decl(numericStringLiteralTypes2.ts, 0, 0))
+
+type SchemaArray = Array<SchemaObject | SchemaArray | LiteralType>;
+>SchemaArray : Symbol(SchemaArray, Decl(numericStringLiteralTypes2.ts, 5, 55))
+>Array : Symbol(Array, Decl(lib.es5.d.ts, --, --), Decl(lib.es5.d.ts, --, --))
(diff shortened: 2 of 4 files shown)


weswigham (dismissed):

weswigham:
@typescript-bot test this
@typescript-bot run dt
@typescript-bot test top100

Andarist:
@jakebailey could you build a playground for this one?

jakebailey:
@typescript-bot pack this

Andarist:
Got the repro case from Vue [here](https://www.staging-typescript.org/play?ts=5.3.0-pr-55386-10#code/C4TwDgpgBAwg9gGwRAxsAlnAdgFXBAZygF4oBJYCAJwEMAjZeJVDbIgHygHUIaBrJsjSYsBANwBYAFDTQkcpVoMIgliKKkAsjTAAeGlhAAaKAZAA+KJwDKEYPsPnJUudB79VwtiW68+2vTMTM0tOdz5bexDnaQATVAQaKmgUNmAoACUIADNrEABbOkQALigAVyx0AEcy6AICooRnCAAPMDgqdPiUROSoVNF0jJoAdzzCkvLKmrqGxBipVvbOqHQsRWyaFGgs7N0cHxCoAG9pKCgANxoEWtKcZ3OAegAqZ7Pz56g8eVj0bOzqBB1ugaMAOlBsAgQAA6d5QT48KBYCAQWJQYAAC3QRDBUDo0DWUDAZQY6BQUFi0OARDoZS62AA5OkRgZ0uh0riCBi4CNymBVlhyAARACicM+NDpcFS+TAyEoJgIcCgI2gZQI0BoRKo6CulCg40aAoIlBolPFj3eAG1dobEABdUrAKi1ZwAX2kcQSSRSaQNGOuCB5u20VD41FKFWqtSg9QmTU9izaHQ5+H9geDOX2h0cPl2+0sADITlArdYA0hM9lQ+GqPaAPxOl3QN0LVxQACqWBGtDA+ZwllIB1alCwsSI5YzI3zawBVCgADVzO964v3ncoCOgePMlnZ9RF8upOdV12ezpbehZchdEv153u73L9eIAW22mAEI0DXfQg+E06lgADmVhImUhQHpwRSILwWALGsGxbDsORnr2X7oAgv5EMcHoyFI3S9L6gzppW068MIFwQDWEZTNGszxu+8ioReOTWFecqvgOPjDi0o7ju8nAAGIVF4WACbAiBCKwuD4AQ4lfj+snibsykoY+OjoZhslWuGIBwNku7ZMxYCaVh9riccpbDGMcwIA2TYxrhJ5fPePF8UQWRmpCIAAIJULQIAOBYK4llaADSApQLp+lfI6D7nn2rHsTeODhfalhOVAG5bmORBwHQABWLBQMWlllhWQZkVsGCUdRdaNkiECUfOmWrqcx7nJ1VoAAqRdFBk4HFvU5TucZGquqXdfaWXxU+WaTelDydZldyMdAxkAHKEJQsS7AQBbcZuvHbh5ORQBNM3Gc+HFvl6PQ+lA2QidJUDJNVuqcUd7kQoVLDmAAFMASRAXYdwAJSlJt22ontt14fdBATiSfkBQdI1ECjNCBVx7XnIkJrdeqGKopGY45GsqKgfc7z48AthJCgxOxKT8TZBTaKcNTeHnMSXL-eyED5ODJxwuiWIENCtOE1ylOkAL+RLecyTAGUVCCgQZSQFQ0K8xi-OUPkYNLbh7xrPELQAPLZP9GoMxiIrIPkQLAHcJjZFQcD5GQZMtA1WDgfi86cBUrPsxDYEQfOuOdZi2KS9+dO8FQjOy7GSeMw7gvO4rr12Kr6ua9Q0Jm60Vs2+n9uO87bse17PtG+8uEm1IAwmrGyP+VjPjIry1gdwF-0ACwmAArCYABsDct36+UalQlFoqQb0URANv91jU-SBrdCYzCuv-QA7FPs-UAvOtE-9ACcU9AA). I will have to find time to investigate what exactly is happening here and how this could be remedied.

Andarist:
Yes, I'll convert this to a temporary draft and ping you back when I manage to undraft it. I didn't yet have time to investigate this break.

Andarist:
The above change is "correct" (well, it's consistent...). This is just a new situation in which this thing can happen today: https://github.com/microsoft/TypeScript/issues/59260

jakebailey:
@typescript-bot test it

(been a bit)

Andarist:
The above error is the same as was reported before. I think this code was just accidentally allowed in Vue types, see my previous comment [here](https://github.com/microsoft/TypeScript/pull/55386#issuecomment-2226398610)

src/compiler/checker.ts
```diff
@@ -16633,7 +16633,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 if (getTypeParameterFromMappedType(mappedType) === getActualTypeVariable(type)) {
                     const typeParameter = getHomomorphicTypeVariable(mappedType);
                     if (typeParameter) {
-                        const constraint = getConstraintOfTypeParameter(typeParameter);
+                        const constraint = getConstraintOfTypeParameter(getConditionalFlowTypeOfType(typeParameter, parent));
```
jakebailey, line 16636: I will note that this call is at least a bit cursed; `getConditionalFlowTypeOfType` returns `Type`, not `TypeParameter`.

Andarist: yeee, this is really cursed 😬 since the result of this function was accepted as a valid type to `getConstraintOfTypeParameter`... I didn't really recheck what's happening here.

I'm changing this to a draft as I also noticed now that it needs extra consideration as to how it interacts with `getResolvedApparentTypeOfMappedType`

Andarist: I'm not yet sure what to do about it, especially in light of https://github.com/microsoft/TypeScript/pull/57801 that only treats mapped types ar arrays/tuples when every constituent of an intersection is array/tuple.

In this situation the constraint of the substitution type created by `getConditionalFlowTypeOfType` is smth like `Schema & SchemaArray` (so `(SchemaArray | SchemaObject | LiteralType) & SchemaArray`). The user's intention is clearly to treat the narrowed-down type as an array/tuple but this intersection type doesn't quite guarantee us that. Even though often an intersection like this is meant to "filter" it isn't exactly that.

I'll have to think about this one more but in the meantime I realized that there is a bug here that can be fixed separately: https://github.com/microsoft/TypeScript/pull/60084

Issue #10122, opened 2016-08-03
Inconsistent multiple export behavior for functions vs vars


evmar:
typescript@2.1.0-dev.20160803 

``` js
export type A = string;
function A() {}
export {A}

type B = string;
var B: void;
export {B};
```

The compiler accepts this module, and if you import it, you'll find that A is only exported as a type, not a value, but that B is both a type and value.

Note that two lines (1 and 3) both attempt to export A.  If you make the B block try to do the same thing (add `export` before the first declaration of B) the compiler rejects the module with: "Individual declarations in merged declaration 'B' must be all exported or all local".

I suspect the right resolution of this is that it should reject the double export of A as well in this module.

DanielRosenwasser:
Yup, and I agree that we should disallow the double export.

Andarist:
@DanielRosenwasser @jakebailey I'm not sure if this should be considered a bug today. I re-examined the `Individual_declarations_in_merged_declaration_0_must_be_all_exported_or_all_local_2395` and I think that this error only targets merged declarations that **share** the type/value space. When the spaces are different then the error is never raised *today* and the type gets exported from the module. I'm not saying this is perfect but that seems to be the intention of the code and what the test suite encodes today - so if anything the rules could be refined but that could also be a breaking change for people relying on the fact that they can only export the type without exporting the value of the same local name (and I've seen this in some repositories)

PR #50853, opened 2022-09-19
Fully resolve aliases when checking symbol flags

andrewbranch:
Fixes #44896
Fixes #50455 

I actually don’t think this is a breaking change in any meaningful way. I do have [one test](https://github.com/microsoft/TypeScript/pull/50853/files#diff-557cc72f2b8a2675ba8da5ec553f983b98f1ca3dd6062e429438421743fd9fee) that exhibits new errors, but the scenario is extremely contrived and didn’t exactly work before. Mostly, this removes false errors and false import elisions. The most visible and understandable example is the one from #44896:

```ts
// @filename: constants.ts
export const COFFEE = 0;
export const TEA = 1;


// @filename: drink.ts
export type Drink = 0 | 1;
export * as Drink from "./constants";


// @filename: index.ts
import { Drink } from "./drink";
// 👀 BEFORE: 'Drink' only refers to a type, but is being used as a value here
// 👍 AFTER: No error
const x: Drink = Drink.TEA;
```

PR #55387, opened 2023-08-15
Add extra tests for type and value symbol merging

Andarist:
this (indirectly~) closes https://github.com/microsoft/TypeScript/issues/10122 , the behavior itself was changed by https://github.com/microsoft/TypeScript/pull/50853 and this PR just "documents" those cases

### DIFF
--- a/tests/cases/conformance/externalModules/typeValueMerge1.ts
+++ b/tests/cases/conformance/externalModules/typeValueMerge1.ts
@@ -1,0 +1,17 @@
+// @noEmit: true
+
+// @Filename: other.ts
+export type A = string;
+function A() {}
+export { A };
+
+export type B = string;
+var B = 10;
+export { B };
+
+// @Filename: main.ts
+import { A, B } from "./other";
+
+A();
+
+export const C = B;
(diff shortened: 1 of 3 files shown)


tests/cases/conformance/externalModules/typeValueMerge1.ts
```diff
@@ -0,0 +1,17 @@
+// @noEmit: true
+
+// @Filename: other.ts
+export type A = string;
+function A() {}
+export { A };
```
Andarist, line 6: this resolves to a value in this situation so it's kinda equivalent to:
```ts
export type A = string;
export function A() {}
```
and this was always~ allowed.

Andarist: Note that in the past this was considered to be ambiguous and one could get an error when writing code like this ("Individual declarations in merged declaration 'B' must be all exported or all local"). In the referenced issue it was observed that the variant with exported function declaration **didn't** report an error (at least back in 2016 :P). @DanielRosenwasser even stated that it should error. However, since the behavior was changed in 4.9 by https://github.com/microsoft/TypeScript/pull/50853 I think that the proper resolution nowadays is to accept the new behavior and just commit those tests.

weswigham (approved):
This is just a new test, so I see no reason not to merge~

PR #55389, opened 2023-08-15
If outDir is not specified, dont need to include json files explicitly in the config since they will not be emitted

### LINKED ISSUE #55366: `include` patterns need to specifiy `.json` extension in composite projects despite `resolveJsonModule: true`
### 🔎 Search Terms

include composite resolveJsonModule

Reopening of https://github.com/microsoft/TypeScript/issues/25636 which doesn't seem to have fixed all problems reported in that issue

This is the behavior in every version I tried (up until 4.1)

### ⏯ Playground Link

https://github.com/eps1lon/ts-references-json

### 💻 Code

```ts
import data from "./data.json";

export default { data };
```

```json
{
  "extends": "../tsconfig.base.json",
  "compilerOptions": {
    "composite": true
  },
  "include": ["./**/*"]
}

```

### 🙁 Actual behavior

```bash
$ tsc --build tsconfig.json
feature-two/index.ts:1:18 - error TS6307: File '/Users/sebastian.silbermann/ts-references-json/feature-two/data.json' is not listed within the file list of project '/Users/sebastian.silbermann/ts-references-json/feature-two/tsconfig.json'. Projects must list all files or use an 'include' pattern.

1 import data from "./data.json";
                   ~~~~~~~~~~~~~


Found 1 error.
```

Explicitly adding `.json` to `include` like https://github.com/microsoft/TypeScript/issues/25636#issuecomment-627111031 suggested works also but is unintuitive since the extension is only required in composite projects.

### 🙂 Expected behavior

Compiles just fine without having to add another `include` pattern using `.json`.

sheetalkamat:
Fixes #55366


andrewbranch (approved):
I think this is not related to this change, but it reminds me of something else I’d like to revisit in the future: the behavior when importing `.json` files outside of `rootDir`. It’s pretty common to import your own package.json file to read the version string for CLI help, for example, and as long as your `rootDir` and `outDir` are siblings, it’s no problem to allow that import to resolve for type checking purposes, outside of `rootDir`, without emitting anything.

sheetalkamat:
@andrewbranch just before going on leave i had also thought about this and was going to experiment with this. Let me see if its easy or moderate effort to look into that as i think i had written some test cases for it.

src/compiler/utilities.ts
```diff
@@ -6382,8 +6384,17 @@ export function sourceFileMayBeEmitted(sourceFile: SourceFile, host: SourceFileM
...
-    return !!options.outDir || !!outFile(options);
+    // Emit json file if outFile is specified
+    if (outFile(options)) return true;
+    // Json file is not emitted if outDir is not specified
+    if (!options.outDir) return false;
+    // Otherwise if rootDir or composite config file, we know common sourceDir and can check if file would be emitted in same location
+    if (options.rootDir || (options.composite && options.configFilePath)) {
+        const commonDir = getNormalizedAbsolutePath(getCommonSourceDirectory(options, () => [], host.getCurrentDirectory(), host.getCanonicalFileName), host.getCurrentDirectory());
+        const outputPath = getSourceFilePathInNewDirWorker(sourceFile.fileName, options.outDir, host.getCurrentDirectory(), commonDir, host.getCanonicalFileName);
+        if (comparePaths(sourceFile.fileName, outputPath, host.getCurrentDirectory(), !host.useCaseSensitiveFileNames()) === Comparison.EqualTo) return false;
```
andrewbranch, line 6395: ❤️ Such a nice improvement!

PR #54442, opened 2023-05-29
Remove unused diagnostic 2346

Andarist:
I noticed that this code is never executed

PR #55392, opened 2023-08-16
Remove an old `Debug.assert` call

Andarist:
https://github.com/microsoft/TypeScript/pull/54442#issuecomment-1571284526

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -33236,7 +33236,6 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         const candidates = candidatesOutArray || [];
         // reorderCandidates fills up the candidates array directly
         reorderCandidates(signatures, candidates, callChainFlags);
-        Debug.assert(candidates.length, "Revert #54442 and add a testcase with whatever triggered this");
 
         const args = getEffectiveCallArguments(node);
 


src/compiler/checker.ts
```diff
@@ -33236,7 +33236,6 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         const candidates = candidatesOutArray || [];
         // reorderCandidates fills up the candidates array directly
         reorderCandidates(signatures, candidates, callChainFlags);
-        Debug.assert(candidates.length, "Revert #54442 and add a testcase with whatever triggered this");
```
jakebailey, line 33239: Maybe at this point we just remove the message and leave the assert? Would this check here be generally helpful to prevent future bugs or aid debugging?

fatcerberus:
So I just noticed this PR and, well, [it looks like you dodged a bullet here](https://github.com/microsoft/TypeScript/issues/55499) 😅

Andarist:
It would fail on the assert in `getCandidateForOverloadFailure`. 

When fixing that issue and reexamining all of those call hierarchies... I came to the conclusion that it would be best to leave the assert in. However, I would move it to the top of the function and assert if the non-empty signatures were received by this function.

All the existing calls leading to this are kind of trying to make sure that this invariant is met and those calls can always produce a **better** error message than the one that could be raised here.

jakebailey (changes requested):
So, should this PR move the assert? The bug still stands to be fixed but I believe there's a PR already open for that.

michaelfaith:
We're hitting up against this issue.  Any movement on this fix by chance?

jakebailey:
The point of this issue is not to "fix" an issue, but to remove a debug assert that shouldn't ever fire. If it's firing for you, please file an issue.

michaelfaith:
I see.  Thanks for clarifying.  We did start seeing this error after upgrading from 5.1 to 5.2, and it's still present in 5.4.  So I'll have to work on creating a repro repo for a new issue.  It'll be tricky though, since our codebase is quite large.

Issue #55391, opened 2023-08-16
Keep argument name of setter in declaration from JavaScript


regseb:
### 🔎 Search Terms

setter, set, argument, parameter, arg, declaration, javascript

### 🕗 Version & Regression Information

This is the behavior in every version I tried, and I reviewed the FAQ for entries about _setter_.

### ⏯ Playground Link

https://www.typescriptlang.org/play?filetype=js#code/MYGwhgzhAEBiD29oG8BQ0PQPQCofs2h2gCEwAnAOgM30KOgAEAHCsAWxQgBdyBLAHYBzAL7QARmABepadXo4sNaBACm3CRQAUkqQEoUywsHgCI8EKsoh4QndL0BuZSNSugA

### 💻 Code

```JavaScript
class Foo {
    /**
     * Bar.
     *
     * @param {string} baz Baz.
     */
    set bar(baz) {
        console.log(baz);
    }
}
```

### 🙁 Actual behavior

In the generated _.D.TS_ , the argument name has been renamed from `baz` to `arg`.

```TypeScript
declare class Foo {
    /**
     * Bar.
     *
     * @param {string} baz Baz.
     */
    set bar(arg: string);
}
```

### 🙂 Expected behavior

In the generated _.D.TS_, the argument name must keep the name `baz`.

```TypeScript
declare class Foo {
    /**
     * Bar.
     *
     * @param {string} baz Baz.
     */
    set bar(baz: string);
}
```

PR #55393, opened 2023-08-16
Fixed serialization of param names in setters from JS files

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/55391


MartinJohns:
The @Andarist was lightning fast again. ⚡

src/compiler/checker.ts
```diff
@@ -9595,13 +9595,21 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+                            });
+                            const setterSignature = setter && getSignatureFromDeclaration(setter);
                             result.push(setTextRange(factory.createSetAccessorDeclaration(
                                 factory.createModifiersFromModifierFlags(flag),
                                 name,
                                 [factory.createParameterDeclaration(
                                     /*modifiers*/ undefined,
                                     /*dotDotDotToken*/ undefined,
-                                    "arg",
+                                   (setterSignature?.parameters[0].valueDeclaration as BindingName | undefined) || "arg",
```
jakebailey, line 9612:
```suggestion
                                    (setterSignature?.parameters[0].valueDeclaration as BindingName | undefined) || "arg",
```

(dprint when?)

src/compiler/checker.ts
```diff
@@ -9595,13 +9595,21 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     if (p.flags & SymbolFlags.Accessor && useAccessors) {
                         const result: AccessorDeclaration[] = [];
                         if (p.flags & SymbolFlags.SetAccessor) {
+                            const setter = p.declarations?.find((d): d is SignatureDeclaration => {
```
jakebailey, line 9598: If a setter has an overloaded signature, what happens here?

Andarist: I'm not even sure how to write such a thing - but perhaps it's possible somehow when the setter is defined using `Object.defineProperty` call in JS file? Do you have an example of an overloaded setter at hand?

jakebailey: Yeah, you're right, there's no setter that could be written that could make use of any extra signatures, since you only have one parameter and no return anyway.

src/compiler/checker.ts
jakebailey, line 9612: Also, is this using the parameter's node directly? Probably, you need to be doing some cloning and such, e.g. what is done in `symbolToParameterDeclaration`. Noticing there that it has a bit of logic specifically for binding names, too.

jakebailey (changes requested):

src/compiler/checker.ts
Andarist: thanks for the tip, i reused `symbolToParameterDeclaration` here and added a test case to cover for what it fixes (stripping initializers from binding patterns)

src/compiler/checker.ts
```diff
@@ -9779,14 +9779,22 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+                                }
+                                const id = getNameOfDeclaration(d);
+                                return !!id && isIdentifier(id) && idText(id) === "set";
+                            });
+                            const paramSymbol = setter && getSignatureFromDeclaration(setter)?.parameters[0];
                             result.push(setTextRange(
                                 factory.createSetAccessorDeclaration(
                                     factory.createModifiersFromModifierFlags(flag),
                                     name,
                                     [factory.createParameterDeclaration(
```
Andarist, line 9794: This whole call could just be replaced with `symbolToParameterDeclaration(paramSymbol, context)` but types don't line up here. `p.declarations` is nullish - how safe, in general, it is to assume that nodes here have `declarations`, or `valueDeclaration` and similar things? I could just use a non-nullish assertion but I'm not sure if that's a good idea here - especially since this code was working without relying on those things before.

jakebailey: By the time it gets here, is it possible for it to be `undefined` at all? e.g. could you assert it?

jakebailey: Hm, the answer seems like "no", and it's also incorrect for setters like:

```
    /** @param {string=} value */
    set p1(value: string | undefined);
```

As the code "correctly" handles the `=` and then makes it optional, when that's invalid. Though that could be fixed.

Maybe just a new helper that uses a fixed `symbolToParameterDeclaration` if the symbol exists, otherwise falls back to a generated name?

Andarist: > Hm, the answer seems like "no"

No - it isn't possible for it to be `undefined` here. Or no - I can't assert it here? 😅 

> As the code "correctly" handles the = and then makes it optional, when that's invalid. Though that could be fixed.

Could you rephrase?

jakebailey: No as in a test fails if it's asserted, as it can be undefined.

I think it's probably clearer if you make the change and see the test change; there's a d.ts file which turns out invalid if you just use the other function as is.

src/compiler/checker.ts
```diff
@@ -9779,17 +9779,36 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     if (p.flags & SymbolFlags.Accessor && useAccessors) {
                         const result: AccessorDeclaration[] = [];
                         if (p.flags & SymbolFlags.SetAccessor) {
+                            let setter: Declaration | undefined;
+
+                            if (p.declarations) {
+                                for (const d of p.declarations) {
```
jakebailey, line 9785: I think you can actually write this as a `forEach`, which will stop iterating and return the first truthy value the callback does.

Andarist: thanks for the tip :P it wasn't meant to be re-reviewed yet but you were fast as always :)

it seems that `symbolToParameterDeclaration` can't be reused in full after all but I'll re-investigate this tomorrow and ping you when I'm done with it

tests/baselines/reference/jsDeclarationsGetterSetter.js
```diff
@@ -106,19 +122,22 @@ export class B {
...
-    set x(arg: number);
+    set x(_arg: number);
 }
 export class F {
-    set x(arg: number);
+    set x(_arg: number);
     get x(): number;
 }
+export class G {
+    set x(args: number);
```
Andarist, line 142: this could be generated as `args_0` (or smth like that) - but I don't think it's worth the added code complexity

src/compiler/checker.ts
```diff
@@ -7594,39 +7594,23 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+                return getDeclarationOfKind<JSDocParameterTag>(parameterSymbol, SyntaxKind.JSDocParameterTag);
             }
-            const parameterTypeNode = serializeTypeForDeclaration(context, parameterType, parameterSymbol, context.enclosingDeclaration, privateSymbolVisitor, bundledImports);
+        }
 
-            const modifiers = !(context.flags & NodeBuilderFlags.OmitParameterModifiers) && preserveModifierFlags && parameterDeclaration && canHaveModifiers(parameterDeclaration) ? map(getModifiers(parameterDeclaration), factory.cloneNode) : undefined;
-            const isRest = parameterDeclaration && isRestParameter(parameterDeclaration) || getCheckFlags(parameterSymbol) & CheckFlags.RestParameter;
-            const dotDotDotToken = isRest ? factory.createToken(SyntaxKind.DotDotDotToken) : undefined;
-            const name = parameterDeclaration ? parameterDeclaration.name ?
+        function parameterToParameterDeclarationName(parameterSymbol: Symbol, parameterDeclaration: ParameterDeclaration | JSDocParameterTag | undefined, context: NodeBuilderContext) {
```
Andarist, line 7607: I decided to extract some useful bits to separate helpers instead of reusing `symbolToParameterDeclaration`. To reuse `symbolToParameterDeclaration` in full I'd have to introduce some extra parameters there (like `preserveType`, `preserveQuestionToken` and I don't think it's worth it. And even with that, I'd have to check if it would for bizarre setters declared using rest params and potentially put even more work into supporting that.

src/compiler/checker.ts
```diff
@@ -7652,6 +7636,33 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             }
         }
 
+        function symbolToParameterDeclaration(parameterSymbol: Symbol, context: NodeBuilderContext, preserveModifierFlags?: boolean, privateSymbolVisitor?: (s: Symbol) => void, bundledImports?: boolean): ParameterDeclaration {
```
jakebailey, line 7639: This got moved; is there any chance you could move it back to where it was for a better diff?

Andarist: I didn't move it but I prepended it with 2 functions that I extracted from it. Git decided that the other parts were not moved, I could juggle this around to force a different diff but logically it makes sense to me that `parameterToParameterDeclarationName` comes before `symbolToParameterDeclaration`

jakebailey: Funny, I often put my helpers below the function that uses it, so you read the high level thing first... 😄

jakebailey: I guess it differed enough that git forgot, very annoying 🙁

fatcerberus: > Funny, I often put my helpers below the function that uses it, so you read the high level thing first...

It's weird, I'm okay with this but on the other hand stuff like

```ts
interface I {
    foo: Foo;
    bar: Bar;
}
interface Foo {
    /* ... */
}
interface Bar {
    /* ... */
}
```

annoys me to no end.

src/compiler/checker.ts
```diff
@@ -9779,14 +9790,31 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+                                        const id = getNameOfDeclaration(propDecl);
+                                        if (!!id && isIdentifier(id) && idText(id) === "set") {
+                                            return propDecl;
+                                        }
+                                    });
+                                }
+                            });
+
+                            Debug.assert(setter && isFunctionLikeDeclaration(setter));
+                            const paramSymbol: Symbol | undefined = getSignatureFromDeclaration(setter).parameters[0];
```
Andarist, line 9808: This annotation isn't enforced, it acts as a documentation hint~. The problem is that indexed access here won't return `| undefined` and TypeScript prefers the expression type here over the declared type (it's a known design decision). In practice, this array might be empty and I handle it in the implementation to make the truthiness check less weird I decided to annotate this here.

src/compiler/checker.ts
```diff
@@ -9779,14 +9790,31 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+
                             result.push(setTextRange(
                                 factory.createSetAccessorDeclaration(
                                     factory.createModifiersFromModifierFlags(flag),
                                     name,
                                     [factory.createParameterDeclaration(
                                         /*modifiers*/ undefined,
                                         /*dotDotDotToken*/ undefined,
-                                        "arg",
+                                        paramSymbol ? parameterToParameterDeclarationName(paramSymbol, getEffectiveParameterDeclaration(paramSymbol), context) : "value",
```
Andarist, line 9817: I decided to use `"value"` as a new fallback here to make it more consistent with: https://github.com/microsoft/TypeScript/blob/202eff75d015f55204eba9d55028f64fdcbfdcc6/src/compiler/transformers/declarations.ts#L897-L903

Andarist, line 9817: I decided to use `"value"` as a new fallback here to make it more consistent with: https://github.com/microsoft/TypeScript/blob/202eff75d015f55204eba9d55028f64fdcbfdcc6/src/compiler/transformers/declarations.ts#L897-L903

src/compiler/checker.ts
```diff
@@ -7594,39 +7594,23 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
             }
-
-            let parameterType = getTypeOfSymbol(parameterSymbol);
-            if (parameterDeclaration && isRequiredInitializedParameter(parameterDeclaration)) {
-                parameterType = getOptionalType(parameterType);
+            if (!isTransientSymbol(parameterSymbol)) {
+                return getDeclarationOfKind<JSDocParameterTag>(parameterSymbol, SyntaxKind.JSDocParameterTag);
             }
-            const parameterTypeNode = serializeTypeForDeclaration(context, parameterType, parameterSymbol, context.enclosingDeclaration, privateSymbolVisitor, bundledImports);
+        }
```
jakebailey, line 7605: Moderately weird to me that this doesn't explicitly return `undefined` but I guess we don't have that flag enabled.

src/compiler/checker.ts
jakebailey: Reviewing the diff, the change is definitely if you move the functions around...

src/compiler/checker.ts
jakebailey, line 7607: Nit: I would put the context as the first parameter.

jakebailey (commented):
Seems okay to me so far, but the diff definitely becomes 10x more reviewable when the functions are reordered (and I did that when reading it on github.dev for that reason)

Andarist:
@jakebailey reordered this, let me know how it feels now

src/compiler/checker.ts
Andarist: I would actually do the same but I have chosen to be consistent with the existing `symbolToParameterDeclaration` which has this signature:
```ts
(local function) symbolToParameterDeclaration(parameterSymbol: Symbol, context: NodeBuilderContext, preserveModifierFlags?: boolean, privateSymbolVisitor?: ((s: Symbol) => void) | undefined, bundledImports?: boolean): ParameterDeclaration
```

I can do the change to both if you prefer that, just let me know.

Andarist: Hm, actually - I probably can't change `symbolToParameterDeclaration` since it's available on the `TypeChecker` and it's not marked as `@internal`.

jakebailey: Hm, I guess I had focused on the other functions which have it the other way, but I guess it doesn't really matter (it was just a nit).

jakebailey (approved):
LGTM; https://github.dev/microsoft/TypeScript/pull/55393/files shows the diff a lot better still even than github at this point. Wish github could use that algorithm somehow :(

Andarist:
pleasure doing business with you

Issue #55394, opened 2023-08-16
Type inference changed for callbacks with optional parameters


h-joo:
### 🔎 Search Terms

optional, inference, callback

### 🕗 Version & Regression Information

- Behaviour changed between versions 5.0 and 5.1

### ⏯ Playground Link

https://www.typescriptlang.org/play?jsx=0&module=1&ts=5.1.3#code/KYDwDg9gTgLgBAMwK4DsDGMCWEWMwGxmCgB4AVAPgAowpgATTNAQyIC44qA3Z-JYDmQA0cTCnqgOKJAFsARsQCUcALwU4ciBHzBmKRYLgBvAL4AoM2hwBneM1V5CxKjQjXgAfimyFUZWrgYKH5FAG4LKxRbDQcEAiIoF0h3LzhpeWIRAH1U9N9-dSCQ0KA

### 💻 Code

```ts
export function filter<T>(predicate: (value: T, index: number) => boolean): T {}
const a = filter((pose?: number) => true);
```

### 🙁 Actual behavior

`a` gets inferred as `number`

### 🙂 Expected behavior

`a` gets inferred as `number|undefined` (this is the case in 5.0.4)

I'm not certain if this change is caused by a bug or intentional, but I couldn't find anything from the [announcement](https://devblogs.microsoft.com/typescript/announcing-typescript-5-1/)  at least that made this change, so rather asking for clarification on whether this is an intended behavior

### Additional information about the issue

The original definition of `filter` comes from rxjs (simplified in the example for reporting purpose) : https://github.com/ReactiveX/rxjs/blob/2947583bb33e97f3db9e6d9f6cea70c62a173060/src/internal/operators/filter.ts#L6

Andarist:
Changed in `5.1.0-dev.20230317`, this is what has been added in this version: [diff](https://github.com/microsoft/TypeScript/compare/7f292bf2a19aa14ed69a55e646111af9533d8f1c...ba42ad36467506f81a4f77d536354c4cb90c32d9). So likely the difference comes from https://github.com/microsoft/TypeScript/pull/52609 . 

I **think** this is a bug. Perhaps it was always there in `inferFromAnnotatedParameters` or maybe there is a specific reason why optional parameters behave differently there. I see that the inference source that is used is just `getEffectiveTypeAnnotationNode(declaration)` there and that the optional marker is ignored. I'll investigate this later more thoroughly.

fatcerberus:
[...]

PR #55397, opened 2023-08-16
Fixed contravariant inferences from annotated optional parameters

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/55394

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -35120,7 +35120,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             if (declaration.type) {
                 const typeNode = getEffectiveTypeAnnotationNode(declaration);
                 if (typeNode) {
-                    inferTypes(inferenceContext.inferences, getTypeFromTypeNode(typeNode), getTypeAtPosition(context, i));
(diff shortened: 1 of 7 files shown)


src/compiler/checker.ts
```diff
@@ -35120,7 +35120,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             if (declaration.type) {
                 const typeNode = getEffectiveTypeAnnotationNode(declaration);
                 if (typeNode) {
-                    inferTypes(inferenceContext.inferences, getTypeFromTypeNode(typeNode), getTypeAtPosition(context, i));
+                    const source = addOptionality(getTypeFromTypeNode(typeNode), /*isProperty*/ false, hasQuestionToken(declaration));
```
weswigham, line 35123: `hasQuestionToken(declaration)` isn't going to do the right thing for jsdoc optional parameters - instead, use the catch-all `isOptionalDeclaration`.

weswigham (changes requested):

src/compiler/checker.ts
Andarist: 👍 pushed out the appropriate change

weswigham (approved):

Issue #52041, opened 2022-12-28
Types for global isNaN and Number.isNaN are swapped


ImRodry:
# Bug Report

### 🔎 Search Terms

isNaN

### 🕗 Version & Regression Information

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about 
This issue was mainly introduced by https://github.com/microsoft/TypeScript/pull/24436

### ⏯ Playground Link

[Playground link with relevant code](https://www.typescriptlang.org/play?#code/JYZwcghmAUBECMsCUACA9GlAlApgFwFcAnAOxBQDMIAbEHFAIwLxVBIHIWIBjbnABzw4AJowCeKACoBlALAAoMAQC2DHEQB0oSDATJ0mAIK0A9iiL5iZSjTqNmrcjz6CR4qXPkLtUOBAbc+hjYlqTkeEQE9EwsbJwozgJCogwSMgpKqupa4L6w-oGowbiEYTa00Q6gCbxJbqkeQA)

### 💻 Code

```ts
isNaN("1") // Returns false but isn't accepted by TS
Number.isNaN("1") // Also returns false but is accepted by TS

isNaN("abc") // Returns true but isn't accepted by TS
Number.isNaN("abc") // Returns false but is accepted by TS
```

### 🙁 Actual behavior

global isNaN only takes a number as a parameter when it should take unknown because, as documented by MDN, "The isNaN() function answers the question "is the input functionally equivalent to [NaN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NaN) when used in a number context".", whereas Number.isNaN "provides a convenient way to check for equality with NaN" but "doesn't force-convert the parameter to a number". That being said I believe the global isNaN should accept an unknown parameter, whereas Number.isNaN should only accept a number.

### 🙂 Expected behavior

isNaN accepts a number and Number.isNaN accepts unknown

fatcerberus:
Duplicate of https://github.com/microsoft/TypeScript/issues/37449 and many, many others.

ImRodry:
This issue also focuses on the fact that Number.isNaN accepts unknown when that doesn’t really make any sense. If the global isNaN discussion is done then ignore that ig but I think that part should still be considered (see the linked pr)

fatcerberus:
I thought the discussion in https://github.com/microsoft/TypeScript/pull/24436 (which you linked in the OP) made it pretty clear why the `Number` functions accept any type.

[...]

Issue #37449, opened 2020-03-18
isNaN parameter should be any (or unknown?)


victorz:
According to the spec (see references below), `isNaN()` should be able to take any kind of value, which would be coerced into a number (using something like `Number(value)`, which seems to make sense from my tests).

Forcing the parameter to be `number` breaks legacy code which relies on being able to pass string arguments to `isNaN()`.

```ts
function f(value: string) {
  if (isNaN(value)) { // compilation error, but should be OK
    ...
  }
}
```

Removing the `string` type from the `value` parameter admittedly fixes the compilation error, but breaks the "no implicity any" rule.

References:

- Spec: https://tc39.es/ecma262/#sec-isnan-number
- Type declaration: https://github.com/microsoft/TypeScript/blob/f8083d291719c7671a876578234edea0ee33d624/lib/lib.es5.d.ts#L53

MartinJohns:
Please respect the issue templates for [bug reports](https://raw.githubusercontent.com/microsoft/TypeScript/master/.github/ISSUE_TEMPLATE/Bug_report.md) and [feature requests](https://raw.githubusercontent.com/microsoft/TypeScript/master/.github/ISSUE_TEMPLATE/Feature_request.md).

---

Both issue templates urge you to first search for existing issues and asks for information (including the search terms you have used to search for existing issues).

This is a duplicate of:

- #3947
- #4002
- #11414
- #12302
- #13227
- #15173
- #16427
- #19153
- #21065
- #23464
- #29607
- #34609

Search terms used: `isNaN`

victorz:
Huh. Fair enough. I see your point in alternate solutions.

However, the original reason for closing the first report of this issue is incorrect. https://github.com/microsoft/TypeScript/pull/3947#issuecomment-123558773

> Reading more the ecmascript doc the argument is specified as Number.

"Number" in the spec is not the type, confusingly enough. It's the parameter name. That becomes evident when you read the algorithm for isNaN().

Regarding searching for issues and issue templates: I simply found the line of code I was looking for in the GitHub project file browser, clicked the little ellipsis next to the line number, and selected "Reference this in a new issue." There was no issue template. Sorry for that, I suppose. I typed "isNaN" in the title but none of the "similar issues" were the same as this one in my opinion. I don't mean to get defensive but I tried and sometimes mistakes happen.

Either way, I still feel like this is in violation of the spec, and that `Number.isNaN()` should be the one whose parameter should have type `number`.

[...]

PR #55398, opened 2023-08-16
es5.d.ts isNan argument type unknown

MayCXC:
Fixes #

### DIFF
--- a/src/lib/es5.d.ts
+++ b/src/lib/es5.d.ts
@@ -30,10 +30,10 @@ declare function parseInt(string: string, radix?: number): number;
 declare function parseFloat(string: string): number;
 
 /**
- * Returns a Boolean value that indicates whether a value is the reserved value NaN (not a number).
- * @param number A numeric value.
+ * Returns a Boolean value that indicates whether a value is the reserved value NaN (not a number) after being coerced to a number.
+ * @param value The value to be tested.
  */
-declare function isNaN(number: number): boolean;
+declare function isNaN(value: unknown): boolean;
 
 /**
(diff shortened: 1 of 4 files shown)


jakebailey:
Not sure if you searched before proposing this, but this change is incorrect.

- https://github.com/microsoft/TypeScript/issues/52041#issuecomment-1371181542
- https://github.com/microsoft/TypeScript/issues/37449#issuecomment-600551704
- https://stackoverflow.com/questions/41750390/what-does-all-legal-javascript-is-legal-typescript-mean

RyanCavanaugh:
Closing PR with no approved issue

MayCXC:
Ah, I was using issues search incorrectly, as were some of the others. However, I do think that both the issue reporters and issue closers are making misinterpretations of the ECMA spec. Number.isNaN and isNaN are both intended to be used on non-number values and should be typed accordingly, for the same reason that both === and == are valid operators in typescript. An example/counterexample is that isNan("NaN") is true while Number.isNan("NaN") is false, not because isNan is coercing "NaN" incorrectly, but because it is meant to coerce "NaN" as per the spec. I can see the benefit to code clarity of making ts users do isNaN(Number("NaN")) rather than rely on ECMA's ToNumber algorithm, but it would be more appropriate to make your own standard library to address those concerns rather than misrepresent ECMA's.

RyanCavanaugh:
It's extremely intentional that the lib doesn't 1-to-1 match the spec. See ["what does all JS is TS mean?"](https://stackoverflow.com/questions/41750390/what-does-all-legal-javascript-is-legal-typescript-mean)

PR #52609, opened 2023-02-04
Collect contravariant-only inference candidates from shorter annotated functions

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/52580

The third time's a charm

PR #55400, opened 2023-08-16
Infer from parameters annotated using JSDoc

Andarist:


tests/cases/compiler/contravariantOnlyInferenceFromAnnotatedFunctionJs.ts
```diff
@@ -0,0 +1,30 @@
+// @strict: true
```
Andarist, line 1: this reuses the existing test case for a TS file: https://github.com/microsoft/TypeScript/pull/52609/files#diff-85ed3085ca9f5d4c893196581ea0785e5d6a5aca5ae449afe499fb4d4d159728

tests/cases/compiler/contravariantOnlyInferenceFromAnnotatedFunctionJs.ts
Andarist, line 1: this is an existing test case, just rewritten to JSDocs, see: https://github.com/microsoft/TypeScript/pull/52609/files#diff-85ed3085ca9f5d4c893196581ea0785e5d6a5aca5ae449afe499fb4d4d159728

src/compiler/checker.ts
```diff
@@ -35117,11 +35117,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         const len = signature.parameters.length - (signatureHasRestParameter(signature) ? 1 : 0);
         for (let i = 0; i < len; i++) {
             const declaration = signature.parameters[i].valueDeclaration as ParameterDeclaration;
-            if (declaration.type) {
```
Andarist, line 35120: I noticed that this check is somewhat redundant because `getEffectiveTypeAnnotationNode` already resolves this and that it actually prevents from this function to work on functions annotated using JSDoc

DanielRosenwasser:

sandersn (approved):

PR #55403, opened 2023-08-16
Enforce CRLF in TS/JS files

jakebailey:
For better or for worse, we format all JS/TS in this repo with `CRLF` line endings. ~~I swear that this used to be caught by an ESLint rule, but I cannot figure out where.~~ These `lib.d.ts` files were not CRLF as we did not lint `lib.d.ts`. But, we do now format them!

Unfortunately, vscode's formatting system does not notice line ending changes, so hitting save on these files doesn't do anything.


.dprint.jsonc
```diff
@@ -4,6 +4,7 @@
     "newLineKind": "auto",
```
jakebailey, line 4: I could change this but that feels like a bit much as it'll redo quite a few things, actually. Maybe that's okay and I should just override this in general?

jakebailey:
Ugh, this is my fault; in the dprint PR, I removed `"linebreak-style": ["error", "windows"]` as I had put it into the "style" section, but forgot to update the dprint PR to enforce CRLF, or just leave that rule behind.

andrewbranch (approved):

jakebailey:
I will note that I normally have `"files.eol": "\r\n",` set in `settings.json`, so it's theoretically better to set this for all files we format, but I don't think we care about it besides the ts/js given that's all we enforced before.

PR #55406, opened 2023-08-16
CodeMapper support

zkat:
CodeMappers are a new language service feature that's designed to work as a deterministic aid to code generators like Copilot. The endpoint takes some arbitrary, generated code, some "focus locations" (generally things like cursor location, current selected method scope range, etc), and tries to "place" the AI-generated code in "the right place". While deterministic, it's fairly heuristic.

This version is able to also target "things that look like they match", like functions with the same name, or if statements with the same condition and actually _replace_ their definitions/bodies, instead of appending them to the given scope.

The intention is to make it easier to go from asking an AI assistant like copilot "give me code to do this", to code that actually fits where it's supposed to. Having the language service do this lets us make more informed decisions than what a simple diff codemapper (the default) might be able to do.

### DIFF
--- a/src/harness/fourslashImpl.ts
+++ b/src/harness/fourslashImpl.ts
@@ -4515,16 +4515,10 @@ export class TestState {
         ranges: Range[],
         changes: string[] = [],
     ): void {
-        let fileName: string | undefined;
-        const focusLocations = ranges.map(({ fileName: fn, pos, end }) => {
-            if (!fileName) {
-                fileName = fn;
-            }
+        const fileName = this.activeFile.fileName;
+        const focusLocations = ranges.map(({ pos, end }) => {
             return [{ start: pos, length: end - pos }];
         });
-        if (!fileName) {
-            throw new Error("No ranges passed in, or something went wrong.");
-        }
         let before = this.getFileContent(fileName);
         focusLocations.sort((a, b) => a[0].start - b[0].start);
         for (const subLoc of focusLocations) {
@@ -4557,7 +4551,7 @@ export class TestState {
 // === ORIGINAL ===
 ${before}
 // === INCOMING CHANGES ===
-${changes}
+${changes.join("\n// ---\n")}
 // === MAPPED ===
 ${after}`;
         this.baseline("mapCode", baseline, ".mapCode.ts");
--- a/src/harness/fourslashImpl.ts
+++ b/src/harness/fourslashImpl.ts
@@ -4512,13 +4512,15 @@ export class TestState {
     }
 
     public baselineMapCode(
-        ranges: Range[],
+        ranges: Range[][],
         changes: string[] = [],
     ): void {
         const fileName = this.activeFile.fileName;
-        const focusLocations = ranges.map(({ pos, end }) => {
-            return [{ start: pos, length: end - pos }];
-        });
+        const focusLocations = ranges.map(r =>
+            r.map(({ pos, end }) => {
+                return { start: pos, length: end - pos };
+            })
+        );
         let before = this.getFileContent(fileName);
         const edits = this.languageService.mapCode(
             fileName,
@@ -4529,6 +4531,9 @@ export class TestState {
             {},
         );
         this.applyChanges(edits);
+        focusLocations.forEach(r => {
+            r.sort((a, b) => a.start - b.start);
+        })
         focusLocations.sort((a, b) => a[0].start - b[0].start);
         for (const subLoc of focusLocations) {
             for (const { start, length } of subLoc) {
--- a/src/harness/fourslashImpl.ts
+++ b/src/harness/fourslashImpl.ts
@@ -4512,16 +4512,42 @@ export class TestState {
     }
 
     public baselineMapCode(
-        { fileName, pos }: Range,
+        ranges: Range[],
         changesFilename: string,
     ): void {
         const changes = this.getFileContent(changesFilename);
-        const beforeContents = this.getFileContent(fileName);
-        const before = beforeContents.slice(0, pos) + "[||]" + beforeContents.slice(pos);
+        let fileName: string | undefined;
+        const focusLocations = ranges.map(({ fileName: fn, pos, end }) => {
+            if (!fileName) {
+                fileName = fn;
+            }
+            return [{ start: pos, length: end - pos }];
+        });
+        if (!fileName) {
+            throw new Error("No ranges passed in, or something went wrong.");
+        }
+        let before = this.getFileContent(fileName);
+        focusLocations.sort((a, b) => a[0].start - b[0].start);
+        for (const subLoc of focusLocations) {
+            for (const { start, length } of subLoc) {
+                let offset = 0;
+                for (const sl2 of focusLocations) {
+                    for (const { start: s2, length: l2 } of sl2) {
+                        if (s2 < start) {
+                            offset += 4;
+                            if ((s2 + l2) > start) {
+                                offset -= 2;
+                            }
+                        }
+                    }
+                }
+                before = before.slice(0, start + offset) + "[|" + before.slice(start + offset, start + offset + length) + "|]" + before.slice(start + offset + length);
+            }
+        }
         const edits = this.languageService.mapCode(
             fileName,
             [changes],
-            [[{ start: pos, length: 1 }]],
+            focusLocations,
             this.formatCodeSettings,
             {},
         );
--- a/src/harness/fourslashImpl.ts
+++ b/src/harness/fourslashImpl.ts
@@ -4520,6 +4520,15 @@ export class TestState {
             return [{ start: pos, length: end - pos }];
         });
         let before = this.getFileContent(fileName);
+        const edits = this.languageService.mapCode(
+            fileName,
+            // We trim the leading whitespace stuff just so our test cases can be more readable.
+            changes,
+            focusLocations,
+            this.formatCodeSettings,
+            {},
+        );
+        this.applyChanges(edits);
         focusLocations.sort((a, b) => a[0].start - b[0].start);
         for (const subLoc of focusLocations) {
             for (const { start, length } of subLoc) {
@@ -4537,15 +4546,6 @@ export class TestState {
                 before = before.slice(0, start + offset) + "[|" + before.slice(start + offset, start + offset + length) + "|]" + before.slice(start + offset + length);
             }
         }
-        const edits = this.languageService.mapCode(
-            fileName,
-            // We trim the leading whitespace stuff just so our test cases can be more readable.
-            changes,
-            focusLocations,
-            this.formatCodeSettings,
-            {},
-        );
-        this.applyChanges(edits);
         const after = this.getFileContent(fileName);
         const baseline = `
 // === ORIGINAL ===
--- a/src/harness/fourslashImpl.ts
+++ b/src/harness/fourslashImpl.ts
@@ -4510,6 +4510,32 @@ export class TestState {
 
         this.verifyCurrentFileContent(newFileContent);
     }
+
+    public baselineMapCode(
+        { fileName, pos }: Range,
+        changesFilename = "/incomingChanges",
+    ): void {
+        const changes = this.getFileContent(changesFilename);
+        const beforeContents = this.getFileContent(fileName);
+        const before = beforeContents.slice(0, pos) + "[||]" + beforeContents.slice(pos);
+        const edits = this.languageService.mapCode(
+            fileName,
+            [changes],
+            [[{ start: pos, length: 1 }]],
+            this.formatCodeSettings,
+            {},
+        );
+        this.applyChanges(edits);
+        const after = this.getFileContent(fileName);
+        const baseline = `
+// === ORIGINAL ===
+${before}
+// === INCOMING CHANGES ===
+${changes}
+// === MAPPED ===
+${after}`;
+        this.baseline("mapCode", baseline, ".mapCode.ts");
+    }
 }
 
 function updateTextRangeForTextChanges({ pos, end }: ts.TextRange, textChanges: readonly ts.TextChange[]): ts.TextRange {
--- a/src/harness/fourslashImpl.ts
+++ b/src/harness/fourslashImpl.ts
@@ -4513,9 +4513,8 @@ export class TestState {
 
     public baselineMapCode(
         ranges: Range[],
-        changesFilename: string,
+        changes: string[] = [],
     ): void {
-        const changes = this.getFileContent(changesFilename);
         let fileName: string | undefined;
         const focusLocations = ranges.map(({ fileName: fn, pos, end }) => {
             if (!fileName) {
@@ -4546,7 +4545,8 @@ export class TestState {
         }
         const edits = this.languageService.mapCode(
             fileName,
-            [changes],
+            // We trim the leading whitespace stuff just so our test cases can be more readable.
+            changes,
             focusLocations,
             this.formatCodeSettings,
             {},
--- a/src/harness/fourslashImpl.ts
+++ b/src/harness/fourslashImpl.ts
@@ -4513,7 +4513,7 @@ export class TestState {
 
     public baselineMapCode(
         { fileName, pos }: Range,
-        changesFilename = "/incomingChanges",
+        changesFilename: string,
     ): void {
         const changes = this.getFileContent(changesFilename);
         const beforeContents = this.getFileContent(fileName);
--- a/src/harness/fourslashImpl.ts
+++ b/src/harness/fourslashImpl.ts
@@ -4533,7 +4533,7 @@ export class TestState {
         this.applyChanges(edits);
         focusLocations.forEach(r => {
             r.sort((a, b) => a.start - b.start);
-        })
+        });
         focusLocations.sort((a, b) => a[0].start - b[0].start);
         for (const subLoc of focusLocations) {
             for (const { start, length } of subLoc) {
--- a/src/harness/fourslashInterfaceImpl.ts
+++ b/src/harness/fourslashInterfaceImpl.ts
@@ -240,7 +240,7 @@ export class VerifyNegatable {
         this.state.uncommentSelection(newFileContent);
     }
 
-    public baselineMapCode(ranges: FourSlash.Range[], changes: string[] = []): void {
+    public baselineMapCode(ranges: FourSlash.Range[][], changes: string[] = []): void {
         this.state.baselineMapCode(ranges, changes);
     }
 }
--- a/src/harness/fourslashInterfaceImpl.ts
+++ b/src/harness/fourslashInterfaceImpl.ts
@@ -240,8 +240,8 @@ export class VerifyNegatable {
         this.state.uncommentSelection(newFileContent);
     }
 
-    public baselineMapCode(range: FourSlash.Range, changesFilename: string): void {
-        this.state.baselineMapCode(range, changesFilename);
+    public baselineMapCode(ranges: FourSlash.Range[], changesFilename: string): void {
+        this.state.baselineMapCode(ranges, changesFilename);
     }
 }
 
--- a/src/harness/fourslashInterfaceImpl.ts
+++ b/src/harness/fourslashInterfaceImpl.ts
@@ -240,7 +240,7 @@ export class VerifyNegatable {
         this.state.uncommentSelection(newFileContent);
     }
 
-    public baselineMapCode(range: FourSlash.Range, changesFilename?: string): void {
+    public baselineMapCode(range: FourSlash.Range, changesFilename: string): void {
         this.state.baselineMapCode(range, changesFilename);
     }
 }
--- a/src/harness/fourslashInterfaceImpl.ts
+++ b/src/harness/fourslashInterfaceImpl.ts
@@ -240,8 +240,8 @@ export class VerifyNegatable {
         this.state.uncommentSelection(newFileContent);
     }
 
-    public baselineMapCode(ranges: FourSlash.Range[], changesFilename: string): void {
-        this.state.baselineMapCode(ranges, changesFilename);
+    public baselineMapCode(ranges: FourSlash.Range[], changes: string[] = []): void {
+        this.state.baselineMapCode(ranges, changes);
     }
 }
 
--- a/src/harness/fourslashInterfaceImpl.ts
+++ b/src/harness/fourslashInterfaceImpl.ts
@@ -239,6 +239,10 @@ export class VerifyNegatable {
     public uncommentSelection(newFileContent: string) {
         this.state.uncommentSelection(newFileContent);
     }
+
+    public baselineMapCode(range: FourSlash.Range, changesFilename?: string): void {
+        this.state.baselineMapCode(range, changesFilename);
+    }
 }
 
 export class Verify extends VerifyNegatable {
--- a/src/server/protocol.ts
+++ b/src/server/protocol.ts
@@ -200,6 +200,7 @@ export const enum CommandTypes {
     ProvideCallHierarchyOutgoingCalls = "provideCallHierarchyOutgoingCalls",
     ProvideInlayHints = "provideInlayHints",
     WatchChange = "watchChange",
+    MapCode = "mapCode",
 }
 
 /**
@@ -2340,6 +2341,38 @@ export interface InlayHintItemDisplayPart {
 
 export interface InlayHintsResponse extends Response {
     body?: InlayHintItem[];
+}
+
+export interface MapCodeRequestArgs  extends FileRequestArgs {
+    /**
+     * The files and changes to try and apply/map.
+     */
+    mapping: MapCodeRequestDocumentMapping;
+}
+
+export interface MapCodeRequestDocumentMapping {
+    /**
+     * The specific code to map/insert/replace in the file.
+     */
+    contents: string[];
+
+    /**
+     * Areas of "focus" to inform the code mapper with. For example, cursor
+     * location, current selection, viewport, etc. Nested arrays denote
+     * priority: toplevel arrays are more important than inner arrays, and
+     * inner array priorities are based on items within that array. Items
+     * earlier in the arrays have higher priority.
+     */
+    focusLocations?: TextSpan[][];
+}
+
+export interface MapCodeRequest extends FileRequest {
+    command: CommandTypes.MapCode;
+    arguments: MapCodeRequestArgs;
+}
+
+export interface MapCodeResponse extends Response {
+    body: readonly FileCodeEdits[];
 }
 
 /**
--- a/src/server/protocol.ts
+++ b/src/server/protocol.ts
@@ -2343,7 +2343,7 @@ export interface InlayHintsResponse extends Response {
     body?: InlayHintItem[];
 }
 
-export interface MapCodeRequestArgs  extends FileRequestArgs {
+export interface MapCodeRequestArgs extends FileRequestArgs {
     /**
      * The files and changes to try and apply/map.
      */
--- a/src/server/session.ts
+++ b/src/server/session.ts
@@ -1906,6 +1906,26 @@ export class Session<TMessage = string> implements EventSender {
         });
     }
 
+    private mapCode(args: protocol.MapCodeRequestArgs): protocol.FileCodeEdits[] {
+        const formatOptions = this.getHostFormatOptions();
+        const preferences = this.getHostPreferences();
+        const { file, languageService } = this.getFileAndLanguageServiceForSyntacticOperation(args);
+        const scriptInfo = this.projectService.getScriptInfoForNormalizedPath(file)!;
+        const focusLocations = args.mapping.focusLocations?.map(spans => {
+            return spans.map(loc => {
+                const start = scriptInfo.lineOffsetToPosition(loc.start.line, loc.start.offset);
+                const end = scriptInfo.lineOffsetToPosition(loc.end.line, loc.end.offset);
+                return {
+                    start,
+                    length: end - start,
+                };
+            });
+        });
+
+        const changes = languageService.mapCode(file, args.mapping.contents, focusLocations, formatOptions, preferences);
+        return this.mapTextChangesToCodeEdits(changes);
+    }
+
     private setCompilerOptionsForInferredProjects(args: protocol.SetCompilerOptionsForInferredProjectsArgs): void {
         this.projectService.setCompilerOptionsForInferredProjects(args.options, args.projectRootPath);
     }
@@ -3609,6 +3629,9 @@ export class Session<TMessage = string> implements EventSender {
         },
         [protocol.CommandTypes.ProvideInlayHints]: (request: protocol.InlayHintsRequest) => {
             return this.requiredResponse(this.provideInlayHints(request.arguments));
+        },
+        [protocol.CommandTypes.MapCode]: (request: protocol.MapCodeRequest) => {
+            return this.requiredResponse(this.mapCode(request.arguments));
         },
     }));
 
--- a/src/services/mapCode.ts
+++ b/src/services/mapCode.ts
@@ -3,7 +3,6 @@ import {
     ClassElement,
     ClassLikeDeclaration,
     createSourceFile,
-    emptyArray,
     FileTextChanges,
     find,
     findAncestor,
--- a/src/services/mapCode.ts
+++ b/src/services/mapCode.ts
@@ -1,0 +1,330 @@
+import {
+    Block,
+    ClassElement,
+    ClassLikeDeclaration,
+    createSourceFile,
+    emptyArray,
+    FileTextChanges,
+    find,
+    findAncestor,
+    findLast,
+    flatten,
+    forEach,
+    formatting,
+    getTokenAtPosition,
+    isBlock,
+    isClassElement,
+    isClassLike,
+    isForInOrOfStatement,
+    isForStatement,
+    isIfStatement,
+    isInterfaceDeclaration,
+    isLabeledStatement,
+    isNamedDeclaration,
+    isSourceFile,
+    isTypeElement,
+    isWhileStatement,
+    LanguageServiceHost,
+    Mutable,
+    Node,
+    NodeArray,
+    or,
+    ScriptTarget,
+    some,
+    SourceFile,
+    Statement,
+    SyntaxKind,
+    textChanges,
+    TextSpan,
+    TypeElement,
+    UserPreferences,
+} from "./_namespaces/ts";
+import { ChangeTracker } from "./textChanges.js";
+
+/** @internal */
+export function mapCode(
+    sourceFile: SourceFile,
+    contents: string[],
+    focusLocations: TextSpan[][] | undefined,
+    host: LanguageServiceHost,
+    formatContext: formatting.FormatContext,
+    preferences: UserPreferences,
+): FileTextChanges[] {
+    try {
+        return textChanges.ChangeTracker.with(
+            { host, formatContext, preferences },
+            changeTracker => {
+                const parsed = contents.map(parse);
+                const flattenedLocations = focusLocations && flatten(focusLocations);
+                for (const nodes of parsed) {
+                    placeNodeGroup(
+                        sourceFile,
+                        changeTracker,
+                        nodes,
+                        flattenedLocations,
+                    );
+                }
+            },
+        );
+    }
+    catch (e) {
+        host.error?.(`mapCode: ${e}`);
+        return emptyArray;
+    }
+}
+
+/**
+ * Tries to parse something into either "top-level" statements, or into blocks
+ * of class-context definitions.
+ */
+function parse(content: string): NodeArray<Node> {
+    // We're going to speculatively parse different kinds of contexts to see
+    // which one makes the most sense, and grab the NodeArray from there. Do
+    // this as lazily as possible.
+    const nodeKinds = [
+        {
+            parse: () =>
+                createSourceFile(
+                    "__mapcode_content_nodes.ts",
+                    content,
+                    ScriptTarget.Latest,
+                    /*setParentNodes*/ true,
+                ),
+            body: (sf: SourceFile) => sf.statements,
+        },
+        {
+            parse: () =>
+                createSourceFile(
+                    "__mapcode_class_content_nodes.ts",
+                    `class __class {\n${content}\n}`,
+                    ScriptTarget.Latest,
+                    /*setParentNodes*/ true,
+                ),
+            body: (cw: SourceFile) => (cw.statements[0] as ClassLikeDeclaration).members,
+        },
+    ];
+
+    const parsedNodes = [];
+    for (const { parse, body } of nodeKinds) {
+        const sourceFile = parse();
+        const bod = body(sourceFile);
+        if (bod.length && sourceFile.parseDiagnostics.length === 0) {
+            // If we run into a case with no parse errors, this is likely the right kind.
+            return bod;
+        }
+        // We only want to keep the ones that have some kind of body.
+        else if (bod.length) {
+            // Otherwise, we'll need to look at others.
+            parsedNodes.push({ sourceFile, body: bod });
+        }
+    }
+    // Heuristic: fewer errors = more likely to be the right kind.
+    const { body } = parsedNodes.sort(
+        (a, b) =>
+            a.sourceFile.parseDiagnostics.length -
+            b.sourceFile.parseDiagnostics.length,
+    )[0];
+    return body;
+}
+
+function placeNodeGroup(
+    originalFile: SourceFile,
+    changeTracker: ChangeTracker,
+    changes: NodeArray<Node>,
+    focusLocations?: TextSpan[],
+) {
+    if (isClassElement(changes[0]) || isTypeElement(changes[0])) {
+        placeClassNodeGroup(
+            originalFile,
+            changeTracker,
+            changes as NodeArray<ClassElement>,
+            focusLocations,
+        );
+    }
+    else {
+        placeStatements(
+            originalFile,
+            changeTracker,
+            changes as NodeArray<Statement>,
+            focusLocations,
+        );
+    }
+}
+
+function placeClassNodeGroup(
+    originalFile: SourceFile,
+    changeTracker: ChangeTracker,
+    changes: NodeArray<ClassElement> | NodeArray<TypeElement>,
+    focusLocations?: TextSpan[],
+) {
+    let classOrInterface;
+    if (!focusLocations || !focusLocations.length) {
+        classOrInterface = find(originalFile.statements, or(isClassLike, isInterfaceDeclaration));
+    }
+    else {
+        classOrInterface = forEach(focusLocations, location =>
+            findAncestor(
+                getTokenAtPosition(originalFile, location.start),
+                or(isClassLike, isInterfaceDeclaration),
+            ));
+    }
+
+    if (!classOrInterface) {
+        throw new Error(
+            "Failed to find a class or interface to map the given code into.",
+        );
+    }
+
+    const firstMatch = classOrInterface.members.find(member => changes.some(change => matchNode(change, member)));
+    if (firstMatch) {
+        // can't be undefined here, since we know we have at least one match.
+        const lastMatch = findLast(
+            classOrInterface.members as NodeArray<ClassElement | TypeElement>,
+            member => changes.some(change => matchNode(change, member)),
+        )!;
+
+        forEach(changes, wipeNode);
+        changeTracker.replaceNodeRangeWithNodes(
+            originalFile,
+            firstMatch,
+            lastMatch,
+            changes,
+        );
+        return;
+    }
+
+    forEach(changes, wipeNode);
+    changeTracker.insertNodesAfter(
+        originalFile,
+        classOrInterface.members[classOrInterface.members.length - 1],
+        changes,
+    );
+}
+
+function placeStatements(
+    originalFile: SourceFile,
+    changeTracker: ChangeTracker,
+    changes: NodeArray<Statement>,
+    focusLocations?: TextSpan[],
+) {
+    if (!focusLocations?.length) {
+        changeTracker.insertNodesAtEndOfFile(
+            originalFile,
+            changes,
+            /*blankLineBetween*/ false,
+        );
+        return;
+    }
+
+    for (const location of focusLocations) {
+        const scope = findAncestor(
+            getTokenAtPosition(originalFile, location.start),
+            (block): block is Block | SourceFile =>
+                or(isBlock, isSourceFile)(block) &&
+                some(block.statements, origStmt => changes.some(newStmt => matchNode(newStmt, origStmt))),
+        );
+        if (scope) {
+            const start = scope.statements.find(stmt => changes.some(node => matchNode(node, stmt)));
+            if (start) {
+                // Can't be undefined here, since we know we have at least one match.
+                const end = findLast(scope.statements, stmt => changes.some(node => matchNode(node, stmt)))!;
+                forEach(changes, wipeNode);
+                changeTracker.replaceNodeRangeWithNodes(
+                    originalFile,
+                    start,
+                    end,
+                    changes,
+                );
+                return;
+            }
+        }
+    }
+
+    let scopeStatements: NodeArray<Statement> = originalFile.statements;
+    for (const location of focusLocations) {
+        const block = findAncestor(
+            getTokenAtPosition(originalFile, location.start),
+            isBlock,
+        );
+        if (block) {
+            scopeStatements = block.statements;
+            break;
+        }
+    }
+    forEach(changes, wipeNode);
+    changeTracker.insertNodesAfter(
+        originalFile,
+        scopeStatements[scopeStatements.length - 1],
+        changes,
+    );
+}
+
+function matchNode(a: Node, b: Node): boolean {
+    if (a.kind !== b.kind) {
+        return false;
+    }
+
+    if (a.kind === SyntaxKind.Constructor) {
+        return a.kind === b.kind;
+    }
+
+    if (isNamedDeclaration(a) && isNamedDeclaration(b)) {
+        return a.name.getText() === b.name.getText();
+    }
+
+    if (isIfStatement(a) && isIfStatement(b)) {
+        return (
+            a.expression.getText() === b.expression.getText()
+        );
+    }
+
+    if (isWhileStatement(a) && isWhileStatement(b)) {
+        return (
+            a.expression.getText() ===
+                b.expression.getText()
+        );
+    }
+
+    if (isForStatement(a) && isForStatement(b)) {
+        return (
+            a.initializer?.getText() ===
+                b.initializer?.getText() &&
+            a.incrementor?.getText() ===
+                b.incrementor?.getText() &&
+            a.condition?.getText() === b.condition?.getText()
+        );
+    }
+
+    if (isForInOrOfStatement(a) && isForInOrOfStatement(b)) {
+        return (
+            a.expression.getText() ===
+                b.expression.getText() &&
+            a.initializer.getText() ===
+                b.initializer.getText()
+        );
+    }
+
+    if (isLabeledStatement(a) && isLabeledStatement(b)) {
+        // If we're actually labeling/naming something, we should be a bit
+        // more lenient about when we match, so we don't care what the actual
+        // related statement is: we just replace.
+        return a.label.getText() === b.label.getText();
+    }
+
+    if (a.getText() === b.getText()) {
+        return true;
+    }
+
+    return false;
+}
+
+function wipeNode(node: Mutable<Node>) {
+    resetNodePositions(node);
+    node.parent = undefined!;
+}
+
+function resetNodePositions(node: Mutable<Node>) {
+    node.pos = -1;
+    node.end = -1;
+    node.forEachChild(resetNodePositions);
+}
--- a/src/services/mapCode.ts
+++ b/src/services/mapCode.ts
@@ -38,7 +38,7 @@ import {
     TextSpan,
     TypeElement,
     UserPreferences,
-} from "./_namespaces/ts";
+} from "./_namespaces/ts.js";
 import { ChangeTracker } from "./textChanges.js";
 
 /** @internal */
--- a/src/services/mapCode.ts
+++ b/src/services/mapCode.ts
@@ -50,27 +50,21 @@ export function mapCode(
     formatContext: formatting.FormatContext,
     preferences: UserPreferences,
 ): FileTextChanges[] {
-    try {
-        return textChanges.ChangeTracker.with(
-            { host, formatContext, preferences },
-            changeTracker => {
-                const parsed = contents.map(parse);
-                const flattenedLocations = focusLocations && flatten(focusLocations);
-                for (const nodes of parsed) {
-                    placeNodeGroup(
-                        sourceFile,
-                        changeTracker,
-                        nodes,
-                        flattenedLocations,
-                    );
-                }
-            },
-        );
-    }
-    catch (e) {
-        host.error?.(`mapCode: ${e}`);
-        return emptyArray;
-    }
+    return textChanges.ChangeTracker.with(
+        { host, formatContext, preferences },
+        changeTracker => {
+            const parsed = contents.map(parse);
+            const flattenedLocations = focusLocations && flatten(focusLocations);
+            for (const nodes of parsed) {
+                placeNodeGroup(
+                    sourceFile,
+                    changeTracker,
+                    nodes,
+                    flattenedLocations,
+                );
+            }
+        },
+    );
 }
 
 /**
--- a/src/services/mapCode.ts
+++ b/src/services/mapCode.ts
@@ -28,7 +28,6 @@ import {
     Node,
     NodeArray,
     or,
-    ScriptTarget,
     some,
     SourceFile,
     Statement,
--- a/src/services/mapCode.ts
+++ b/src/services/mapCode.ts
@@ -52,7 +52,7 @@ export function mapCode(
     return textChanges.ChangeTracker.with(
         { host, formatContext, preferences },
         changeTracker => {
-            const parsed = contents.map(parse);
+            const parsed = contents.map(c => parse(sourceFile, c));
             const flattenedLocations = focusLocations && flatten(focusLocations);
             for (const nodes of parsed) {
                 placeNodeGroup(
@@ -70,7 +70,7 @@ export function mapCode(
  * Tries to parse something into either "top-level" statements, or into blocks
  * of class-context definitions.
  */
-function parse(content: string): NodeArray<Node> {
+function parse(sourceFile: SourceFile, content: string): NodeArray<Node> {
     // We're going to speculatively parse different kinds of contexts to see
     // which one makes the most sense, and grab the NodeArray from there. Do
     // this as lazily as possible.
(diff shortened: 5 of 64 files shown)


fatcerberus:
> putting this up to facilitate discussion

It would help if you explained what this even is.

jakebailey:
Kat is a team member, no need to be hasty 🙂

zkat:
@fatcerberus sorry for the lack of clarity. We're working internally on a new codeaction-like feature, and I wanted to put up this PR to help coordinate between my team (the VS side, mainly, with some tsserver), and the VS Code team implementing this on their end. We'll have more details when more details are actually relevant to share :)

src/server/protocol.ts
```diff
@@ -2700,6 +2701,42 @@ export interface InlayHintsResponse extends Response {
     body?: InlayHintItem[];
 }
 
+export interface MapCodeRequestArgs {
+    /// The files and changes to try and apply/map.
```
MariaSolOs, line 2705: Ehhh why the 3 slashes? It seems like documentation in this file uses JSDoc-y comments `/** */`.

src/server/protocol.ts
```diff
@@ -2700,6 +2701,42 @@ export interface InlayHintsResponse extends Response {
     body?: InlayHintItem[];
 }
 
+export interface MapCodeRequestArgs {
+    /// The files and changes to try and apply/map.
+    mappings: MapCodeRequestDocumentMapping[];
+
+    /// Edits to apply to the current workspace before performing the mapping.
```
MariaSolOs, line 2708: I don't think TSServer uses the concept of "workspace".

src/server/protocol.ts
```diff
@@ -2700,6 +2701,42 @@ export interface InlayHintsResponse extends Response {
...
+export interface MapCodeRequestArgs {
+    /// The files and changes to try and apply/map.
+    mappings: MapCodeRequestDocumentMapping[];
+
+    /// Edits to apply to the current workspace before performing the mapping.
+    updates?: FileCodeEdits[];
+}
+
+export interface MapCodeRequestDocumentMapping {
+    /// The file for the request (absolute pathname required). Null/undefined
```
MariaSolOs, line 2713:
```suggestion
    /// The file for the request (absolute pathname required). undefined

```

src/server/protocol.ts
```diff
@@ -2700,6 +2701,42 @@ export interface InlayHintsResponse extends Response {
...
+    /// if specific file is unknown.
+    file?: string;
+
+    /// Optional name of project that contains file
+    projectFileName?: string;
+
+    /// The specific code to map/insert/replace in the file.
+    contents: string[];
```
MariaSolOs, line 2721: What if you want to remove code? What would `contents` be?

src/services/mapCode.ts
```diff
@@ -0,0 +1,366 @@
...
+        sysLog(`mapCode: ${e}`);
+        return [];
+    }
+}
+
+/**
+ * Tries to parse something into either "top-level" statements, or into blocks
+ * of class-context definitions.
+ * @param content
+ * @returns
```
MariaSolOs, line 77:
```suggestion
```

src/services/mapCode.ts
```diff
@@ -0,0 +1,366 @@
...
+ */
+function parse(content: string): NodeArray<Node> {
+    // We're going to speculatively parse different kinds of contexts to see
+    // which one makes the most sense, and grab the NodeArray from there. Do
+    // this as lazily as possible.
+    const nodeKinds = [
+        { parse: () => createSourceFile("__mapcode_content_nodes.ts", content, ScriptTarget.Latest, /*setParentNodes*/ true), body: (sf: SourceFile) => sf.statements },
+        { parse: () => createSourceFile("__mapcode_class_content_nodes.ts", `class __class {\n${content}\n}`, ScriptTarget.Latest, /*setParentNodes*/ true), body: (cw: SourceFile) => (cw.statements[0] as ClassLikeDeclaration).members },
```
MariaSolOs, line 85: I think the line ending to use depends on the project configuration.

src/services/types.ts
```diff
@@ -1124,6 +1126,12 @@ export interface EditorSettings {
     trimTrailingWhitespace?: boolean;
 }
 
+// export interface PrioritizedSpan {
```
MariaSolOs, line 1129: To be removed?

src/server/protocol.ts
zkat: there's no specific facility for _removing_ code. If copilot wants to remove code, it simply includes the larger scope, with the code removed from it.

src/services/mapCode.ts
zkat: that doesn't really matter in this case: as long as it's at all recognized as a line ending, it'll be fine. We throw away the surrounding class anyway.

sandersn:
@zkat is there an associated PR to vscode?

src/services/mapCode.ts
```diff
@@ -0,0 +1,365 @@
...
+    _sourceMapper: SourceMapper,
+    _updates?: FileTextChanges[],
+): FileTextChanges[] {
+    try {
+        return textChanges.ChangeTracker.with(
+            { host, formatContext, preferences },
+            changeTracker => {
+                mappings.forEach(({ sourceFile, contents, focusLocations }) => {
+                    if (!sourceFile) {
+                        // TODO(@zkat): uhhh... do something about file-less mappings. Not supported for now.
```
MariaSolOs, line 56: If this is still an unknown, why do we want to still support an undefined `file` in `MapCodeRequestDocumentMapping`?

src/services/mapCode.ts
```diff
@@ -0,0 +1,365 @@
...
+    catch (e) {
+        sysLog(`mapCode: ${e}`);
+        return [];
+    }
+}
+
+/**
+ * Tries to parse something into either "top-level" statements, or into blocks
+ * of class-context definitions.
+ * @param content
```
MariaSolOs, line 76:
```suggestion
```

No point in keeping this + the type annotation.

sheetalkamat (changes requested):
I havent yet gotten around to actual mapCode.ts

src/services/services.ts
```diff
@@ -3090,6 +3092,23 @@ export function createLanguageService(
         return InlayHints.provideInlayHints(getInlayHintsContext(sourceFile, span, preferences));
     }
 
+    function mapCode(mappings: MapCodeDocumentMapping[], formatOptions: FormatCodeSettings, preferences: UserPreferences = emptyOptions, updates?: FileTextChanges[]): FileTextChanges[] {
+        synchronizeHostData();
```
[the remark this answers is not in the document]
zkat: I plan on using the typeChecker later, it's just not being used in this iteration. I did originally plan on using it from the start. Do I just remove this line for now to make it a syntax operation?

src/services/mapCode.ts
```diff
@@ -0,0 +1,365 @@
...
+                return;
+            }
+        }
+    }
+
+    // If we have no frame of reference, we'll just insert at the end of the
+    // nearest statements scope, or, if we can't find a block, at the top
+    // level of the source file.
+    let scopeStatements: NodeArray<Statement> = file.statements;
+    top: for (const locationGroup of focusLocations) {
```
sandersn, line 291: This requires focusLocations to be flattened. The fallback to original isn't needed because findAncestor will walk to the root, which is the SourceFile `original`. So I put a `!` but maybe `?? original` would be better
```suggestion
    const scopeStatements = focusLocations.find(loc => findAncestor(getNodeAtPosition(original, loc.start), isBlock))!.statements;
```

src/services/mapCode.ts
```diff
@@ -0,0 +1,365 @@
...
+    else {
+        let node: Node | undefined;
+        top: for (const locationGroup of focusLocations) {
+            for (const location of locationGroup) {
+                node = findAncestor(getNodeAtPosition(file, location.start), node => {
+                    return isClassLike(node) || node.kind === SyntaxKind.InterfaceDeclaration;
+                });
+                if (node && node.kind === SyntaxKind.InterfaceDeclaration) {
+                    classOrInterface = node as InterfaceDeclaration;
+                    break top;
```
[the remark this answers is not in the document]
sandersn: I recommend flattening focusLocations earlier since the arrays are always iterated in a double-nested way.

src/services/mapCode.ts
zkat: It's part of the protocol that VS/VSCode send already, but since we're doing fuzzy heuristics, I think it's ok for us to accept it in the protocol, but ignore it for now.

src/services/mapCode.ts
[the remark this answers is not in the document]
zkat: 😎😎😎

src/services/mapCode.ts
```diff
@@ -0,0 +1,365 @@
...
+            if (block) {
+                scopeStatements = (block as Block).statements;
+                break top;
+            }
+        }
+    }
+    changeTracker.insertNodesAfter(file, scopeStatements[scopeStatements.length - 1], nodes);
+}
+
+function getNodeAtPosition(sourceFile: SourceFile, position: number): Node {
```
navya9singh, line 303: I believe this function exists in program.ts. Can we move it to utilities.ts to avoid repeating the code?

src/services/mapCode.ts
```diff
@@ -0,0 +1,365 @@
...
+            }
+            else if (stmt.kind === SyntaxKind.InterfaceDeclaration) {
+                classOrInterface = stmt as InterfaceDeclaration;
+                break;
+            }
+        }
+    }
+    else {
+        let node: Node | undefined;
+        top: for (const locationGroup of focusLocations) {
```
sandersn, line 144: requires focusLocations to be flattened:
```suggestion
        classOrInterface = focusLocations.find(loc => findAncestor(getNodeAtPosition(original, loc.start), node => isClassLike(node) || isInterfaceDeclaration(node)));
```

src/services/mapCode.ts
```diff
@@ -0,0 +1,365 @@
...
+
+    if (a.getText() === b.getText()) {
+        return true;
+    }
+
+    // TODO: Maybe match by other characteristics?
+    return false;
+}
+
+function findAncestor(node: Node, f: (node: Node) => boolean): Node | undefined {
```
navya9singh, line 358: I saw @andrewbranch point this out in another pr, there is an existing utility function `findAncestor()` in utilitiesPublic.ts that can be used here.

src/services/mapCode.ts
```diff
@@ -0,0 +1,365 @@
...
+                classEls.push(node);
+            }
+            else if (isTypeElement(node)) {
+                typeEls.push(node);
+            }
+        }
+        if (classEls.length || typeEls.length) {
+            if (classOrInterface.members.length === 0) {
+                if (classEls.length) {
+                    const newNode = { ...classOrInterface, members: factory.createNodeArray(classEls) };
```
sandersn, line 186: what if `typeEls.length > 0` here? Is that impossible because type members and class members can't occur in the same class?

sandersn (changes requested):
I wasn't looking at other comments coming in so probably some of them are redundant.

src/server/protocol.ts
```diff
@@ -2700,6 +2701,54 @@ export interface InlayHintsResponse extends Response {
...
+     */
+    projectFileName?: string;
+
+    /**
+     * The specific code to map/insert/replace in the file.
+     */
+    contents: string[];
+
+    /**
+     * Areas of "focus" to inform the code mapper with. For example, cursor
```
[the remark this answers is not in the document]
zkat: no, they're arbitrary "just kinda look in this area", with no specific role identifiers, even if in practice, this is currently "current scope" + "cursor position"

sandersn:
re vscode PR: I forgot that this is probably going to be in VS first. Is there a PR there? Maybe you can drop a link in Teams if there's no public way to link it.

src/server/protocol.ts
```diff
@@ -2700,6 +2701,54 @@ export interface InlayHintsResponse extends Response {
...
+export interface MapCodeRequestArgs {
+    /**
+     * The files and changes to try and apply/map.
+     */
+    mappings: MapCodeRequestDocumentMapping[];
+
+    /**
+     * Edits to apply before performing the mapping.
+     */
+    updates?: FileCodeEdits[];
```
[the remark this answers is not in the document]
zkat: They're coming from the editor. They're meant to be used for refinement, where VS has a temporary patch that isn't really applied to the textbuffer, but still wants the next step of the chat refinement to use what the current patch is meant to be.

This is basically only for VS, too, and isn't even used (yet), because it's hard, but VS Code doesn't need it at all because it just applies changes immediately.

src/services/mapCode.ts
```diff
@@ -0,0 +1,365 @@
...
+    TypeElement,
+    UserPreferences,
+} from "./_namespaces/ts";
+import {
+    ChangeTracker,
+} from "./textChanges";
+
+/** @internal */
+export function mapCode(
+    mappings: { sourceFile?: SourceFile; contents: string[]; focusLocations?: TextSpan[][]; }[],
```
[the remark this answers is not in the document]
zkat: image

src/services/mapCode.ts
```diff
@@ -0,0 +1,365 @@
...
+    _program: Program,
```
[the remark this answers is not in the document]
zkat: Placeholders because I plan on using them later when mapCode gets fancier, with speculative patch application and such.

src/services/mapCode.ts
```diff
@@ -0,0 +1,365 @@
...
+/**
+ * Tries to parse something into either "top-level" statements, or into blocks
+ * of class-context definitions.
+ * @param content
+ */
+function parse(content: string): NodeArray<Node> {
+    // We're going to speculatively parse different kinds of contexts to see
+    // which one makes the most sense, and grab the NodeArray from there. Do
+    // this as lazily as possible.
+    const nodeKinds = [
```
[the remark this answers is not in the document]
zkat: I might add things like enum (which used to be there), type declaration, etc contexts in the future.

src/services/mapCode.ts
```diff
@@ -0,0 +1,365 @@
...
+    //    to expand our scope.
+    // 2. Get nodes for each focusLocation by finding the specific node that
+    //    contains it, and making a comparison with the first node type.
+    // 3. Figure out how much of a range we're going to replace or insert by
+    //    using first and last node from `nodes`. Use node types and names to
+    //    match ranges.
+    if (isClassElement(nodes[0])) {
+        placeClassNodeGroup(file, changeTracker, nodes as NodeArray<ClassElement>, focusLocations);
+    }
+    else if (isTypeElement(nodes[0])) {
```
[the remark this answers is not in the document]
zkat: huh. Apparently not.

src/services/mapCode.ts
```diff
@@ -0,0 +1,365 @@
...
+                // "injected" before it.
+                //
+                // We do this by scanning the scope for all nodes that have a
+                // match in the incoming map request code, then sorting and
+                // picking first and last by position, and that will be our
+                // range for replacement.
+                const matches = (scope as Block).statements.filter(stmt => nodes.some(node => matchNode(node, stmt)));
+                matches.sort((a, b) => a.pos - b.pos);
+                changeTracker.replaceNodeRangeWithNodes(file, matches[0], matches[matches.length - 1], nodes);
+                // はい、できました！
```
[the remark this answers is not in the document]
zkat: Yes, we did it! (I can remove it if you want, I was just being silly)

src/services/mapCode.ts
```diff
@@ -0,0 +1,365 @@
...
+                throw new Error("Nothing to append");
+            }
+        }
+    }
+}
+
+function placeStatements(file: SourceFile, changeTracker: ChangeTracker, nodes: NodeArray<Statement>, focusLocations?: TextSpan[][]) {
+    // If there's an empty or null focusLocation, we just shove everything at
+    // the end of the file and call it a day.
+    if (!focusLocations || !focusLocations.length) {
```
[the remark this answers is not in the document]
zkat: no this seems better :)

src/services/mapCode.ts
```diff
@@ -0,0 +1,365 @@
...
+            }
+            else {
+                throw new Error("Nothing to append");
+            }
+        }
+    }
+}
+
+function placeStatements(file: SourceFile, changeTracker: ChangeTracker, nodes: NodeArray<Statement>, focusLocations?: TextSpan[][]) {
+    // If there's an empty or null focusLocation, we just shove everything at
```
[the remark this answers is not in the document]
zkat: Eventually, the plan is for CodeMappers to start returning code that doesn't belong in any particular file, or code that's meant for multiple files but isn't necessarily meant to be placed anywhere in particular.

We just don't do this _right now_.

src/services/mapCode.ts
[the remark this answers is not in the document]
zkat: done.

src/services/mapCode.ts
```diff
@@ -0,0 +1,365 @@
...
+                classOrInterface = stmt as InterfaceDeclaration;
+                break;
+            }
+        }
+    }
+    else {
+        let node: Node | undefined;
+        top: for (const locationGroup of focusLocations) {
+            for (const location of locationGroup) {
+                node = findAncestor(getNodeAtPosition(file, location.start), node => {
```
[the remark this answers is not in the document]
zkat: location is from the original source, yes.

src/server/protocol.ts
```diff
@@ -174,6 +174,7 @@ export const enum CommandTypes {
     ProvideCallHierarchyOutgoingCalls = "provideCallHierarchyOutgoingCalls",
     ProvideInlayHints = "provideInlayHints",
     WatchChange = "watchChange",
+    MapCode = "mapCode",
```
sheetalkamat, line 177: In general we have `-Full` requests for VS .. i am not sure if do that any more but something to think about.

src/services/mapCode.ts
```diff
@@ -0,0 +1,365 @@
...
+        return !!(aName && aName === bName);
+    }
+
+    if (a.getText() === b.getText()) {
+        return true;
+    }
+
+    // TODO: Maybe match by other characteristics?
+    return false;
+}
```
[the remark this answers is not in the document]
zkat: I'll change the getText, but I'm going to keep the "long" version because I think it's clearer about what the match cases are :)

src/services/mapCode.ts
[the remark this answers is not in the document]
zkat: apparently there was already a `getNodeAtPosition`!

src/services/services.ts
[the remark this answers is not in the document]
sheetalkamat: I think so.. you would also want to make sure that `project.getLanguageService(/*ensureSynchronized*/ false)` is called for this operation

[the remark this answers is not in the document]
sheetalkamat: Pls also add tests and its infrastructure eg see how gthere are fourslash tests for `getEdits` or code fix etc ..

src/services/mapCode.ts
```diff
@@ -0,0 +1,365 @@
...
+                // have `baz` in the original be the end of our replacement
+                // range. `other` should be left intact, while `qux` should be
+                // "injected" before it.
+                //
+                // We do this by scanning the scope for all nodes that have a
+                // match in the incoming map request code, then sorting and
+                // picking first and last by position, and that will be our
+                // range for replacement.
+                const matches = (scope as Block).statements.filter(stmt => nodes.some(node => matchNode(node, stmt)));
+                matches.sort((a, b) => a.pos - b.pos);
```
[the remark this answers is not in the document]
zkat: They won't be if, say, Copilot reordered two functions. At least if I'm thinking about it right.

tests/baselines/reference/api/typescript.d.ts
```diff
@@ -10659,6 +10701,7 @@ declare namespace ts {
         commentSelection(fileName: string, textRange: TextRange): TextChange[];
         uncommentSelection(fileName: string, textRange: TextRange): TextChange[];
         getSupportedCodeFixes(fileName?: string): readonly string[];
+        mapCode(mappings: MapCodeDocumentMapping[], formatOptions: FormatCodeSettings, preferences: UserPreferences, updates?: FileTextChanges[]): FileTextChanges[];
```
sheetalkamat, line 10704: These should be internal for now since this is not completely hashed out from the looks of it
Everything except protocol changes should be internal

src/services/types.ts
```diff
@@ -1501,6 +1503,12 @@ export interface OutliningSpan {
     kind: OutliningSpanKind;
 }
 
+export interface MapCodeDocumentMapping {
+    fileName?: string;
+    focusLocations?: TextSpan[][];
+    contents: string[];
```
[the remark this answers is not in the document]
zkat: I'd rather not change the protocol any further if I can avoid it, since that's already being used by a number of teams and has pretty high overhead for alignment.

src/services/mapCode.ts
```diff
@@ -0,0 +1,365 @@
...
+            return name.escapedText as string;
+        case SyntaxKind.JsxNamespacedName:
+            return name.namespace.text + ":" + name.name.text;
+        default:
+            return name.getText();
+    }
+}
+
+function matchNode(a: Node, b: Node): boolean {
+    if (a.kind !== b.kind) {
```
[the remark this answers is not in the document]
zkat: I intend to add more matching heuristics in the future, which go beyond name or exact text match. I just didn't work on them for this version.

src/services/types.ts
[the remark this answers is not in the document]
sheetalkamat: Who is using this without changes in tsserver..

src/server/protocol.ts
MariaSolOs: Well now even in VS we're not using the `-full` variations anymore because of our migration to LSP.

zkat: This is a single request shared with VS Code. It's intended to be LSP-ish as-is.

src/server/protocol.ts
```diff
@@ -2700,6 +2701,54 @@ export interface InlayHintsResponse extends Response {
...
+
+    /**
+     * Edits to apply before performing the mapping.
+     */
+    updates?: FileCodeEdits[];
+}
+
+export interface MapCodeRequestDocumentMapping {
+    /**
+     * The file for the request (absolute pathname required). `undefined`
```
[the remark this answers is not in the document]
zkat: `FileRequestArgs` and `NavToRequestArgs` both do, and... I'm not actually sure why they would need to be absolute. I don't exactly remember why I put this comment here (it's been a while), but I see that it matches other places where file-related requests are being made, so...

src/server/session.ts
```diff
@@ -1905,6 +1906,71 @@ export class Session<TMessage = string> implements EventSender {
...
     }
 
+    private mapCode(args: protocol.MapCodeRequestArgs): protocol.FileCodeEdits[] {
+        const formatOptions = this.getHostFormatOptions();
+        const preferences = this.getHostPreferences();
+        const projects = new Map<Project, MapCodeDocumentMapping[]>();
+        args.mappings.forEach(mapping => {
+            if (!mapping.file) {
+                return { contents: mapping.contents };
+            }
```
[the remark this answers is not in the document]
zkat: It's not. This is leftover code from when this was a `.map()`. Fixed.

src/services/types.ts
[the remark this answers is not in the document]
zkat: C# is already using the protocol for their own codemapper, so is Razor. Other languages are following suit.

src/services/mapCode.ts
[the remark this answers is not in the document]
sandersn: argh, foiled by eslint once again!

(we should maybe switch to the array-simple setting for that rule, but this isn't the place for it)

src/services/types.ts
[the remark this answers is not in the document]
sandersn: For me, changing it in mapCode.ts is the important part because it makes the code easier to read.

src/services/mapCode.ts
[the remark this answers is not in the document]
sandersn: either way is fine with me.

src/services/mapCode.ts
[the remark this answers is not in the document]
sandersn: I got very worried about this early in my review before I figured out which structures were original and which were generated. *In theory* it would be nice to mark the distinction in the type system, but I don't think it's practical.

src/services/mapCode.ts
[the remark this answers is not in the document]
iisaduan: Since the code deletes all of the original matched statements, the order that any corresponding statements generate in copilot shouldn't matter since the copilot content just gets "pasted" in. I also had to think about this a couple times.

Instead of filtering, since we only use the first and last match, we could just `find()` and `findLast()` through `scope.statements`?

src/server/protocol.ts
zkat: I'd like to keep the naming scheme close to what the protocol is, and that's already in place and in use.

src/server/protocol.ts
```diff
@@ -2700,6 +2701,54 @@ export interface InlayHintsResponse extends Response {
...
+    contents: string[];
+
+    /**
+     * Areas of "focus" to inform the code mapper with. For example, cursor
+     * location, current selection, viewport, etc. Nested arrays denote
+     * priority: toplevel arrays are more important than inner arrays, and
+     * inner array priorities are based on items within that array. Items
+     * earlier in the arrays have higher priority.
+     */
+    focusLocations?: FileSpan[][];
```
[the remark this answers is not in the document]
zkat: this is ordered by `flatten()`-ing the nested array. We might do something fancier in the future, but this is sufficient for now.

src/services/types.ts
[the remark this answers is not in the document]
zkat: I think we should keep the name, because consistency is more important to me in this case.

src/server/session.ts
```diff
@@ -1904,6 +1905,71 @@ export class Session<TMessage = string> implements EventSender {
...
     }
 
+    private mapCode(args: protocol.MapCodeRequestArgs): protocol.FileCodeEdits[] {
+        const formatOptions = this.getHostFormatOptions();
+        const preferences = this.getHostPreferences();
+        const projects = new Map<Project, MapCodeDocumentMapping[]>();
+        args.mappings.forEach(mapping => {
+            if (!mapping.file) {
+                // TODO(@zkat): We don't support mapping file-less code yet.
+                return;
```
MariaSolOs, line 1915: Should this be a `continue` instead so that other mappings are still processed?

src/services/mapCode.ts
```diff
@@ -0,0 +1,480 @@
...
+    _updates?: FileTextChanges[],
+): FileTextChanges[] {
+    try {
+        return textChanges.ChangeTracker.with(
+            { host, formatContext, preferences },
+            changeTracker => {
+                mappings.forEach(({ sourceFile, contents, focusLocations }) => {
+                    if (!sourceFile) {
+                        // TODO(@zkat): uhhh... do something about file-less mappings. Not supported for now.
+                        return;
```
MariaSolOs, line 69: Same comment as before: Should we make this a `continue`?

src/services/mapCode.ts
```diff
@@ -0,0 +1,480 @@
...
+    }
+
+    return false;
+}
+
+function wipeNode(node: Mutable<Node>) {
+    resetNodePositions(node);
+    while (node && node.kind !== SyntaxKind.SourceFile) {
+        if (node.parent.kind === SyntaxKind.SourceFile) {
+            node.parent = undefined!;
```
MariaSolOs, line 469: What does `undefined!` do/mean?

src/server/session.ts
```diff
@@ -1904,6 +1905,71 @@ export class Session<TMessage = string> implements EventSender {
         });
     }
 
+    private mapCode(args: protocol.MapCodeRequestArgs): protocol.FileCodeEdits[] {
+        const formatOptions = this.getHostFormatOptions();
+        const preferences = this.getHostPreferences();
```
MariaSolOs, line 1910: Is there a test for checking that mapped code has the right project formatting settings applied?

src/server/session.ts
zkat: That's a `forEach`. The others will get processed, no? Did I miss something?

src/services/mapCode.ts
zkat: ditto: it's a `forEach`?

src/services/mapCode.ts
zkat: we simply choose to prioritize classEls.

sheetalkamat:
can you please update the description of what this change is suppose to do, what are the scenarios supported etc before we can review this

zkat:
@sheetalkamat done. lmk here or on Teams if you have any Qs

src/services/mapCode.ts
MariaSolOs: Awesome, TIL!

src/server/session.ts
MariaSolOs: Oh duh nvm.

src/services/mapCode.ts
```diff
@@ -0,0 +1,484 @@
...
+            // ```
+            //
+            // `bar` will be the beginning of our range, and we want to
+            // have `baz` in the original be the end of our replacement
+            // range. `other` should be left intact, while `qux` should be
+            // "injected" before it.
+            //
+            // We do this by scanning the scope for all nodes that have a
+            // match in the incoming map request code, then sorting and
+            // picking first and last by position, and that will be our
```
sandersn, line 367:
```suggestion
            // We do this by scanning the scope for the first and last
            // nodes that have a match in the incoming map request code,
            // and that will be our range for replacement.
```

src/services/mapCode.ts
```diff
@@ -0,0 +1,484 @@
...
+            // range for replacement.
+            const start = (scope as Block).statements.find(stmt => changes.some(node => matchNode(node, stmt)));
+            const end = findLast((scope as Block).statements, stmt => changes.some(node => matchNode(node, stmt)));
+            // This should generally never be the case at this point, but just in case...
```
sandersn, line 371: what should never be the case? `!start || !end` ?

src/services/mapCode.ts
```diff
@@ -0,0 +1,484 @@
...
+    if (isForInOrOfStatement(a)) {
+        return (
+            a.expression.getText() ===
+                (b as ForInStatement).expression.getText() &&
+            a.initializer.getText() ===
+                (b as ForInStatement).initializer.getText()
+        );
+    }
+
+    if (isLabeledStatement(a)) {
```
sandersn, line 455: this needs a test

src/services/mapCode.ts
```diff
@@ -0,0 +1,484 @@
...
+    }
+    forEach(changes, wipeNode);
+    changeTracker.insertNodesAfter(
+        originalFile,
+        scopeStatements[scopeStatements.length - 1],
+        changes,
+    );
+}
+
+function matchNode(a: Node, b: Node): boolean {
```
sandersn, line 408: This is a lot more careful and precise than the existing diff-matching code, I think. Is it going to cause a lot more results to be insertions instead of diffs? Is that a good thing?

My first thoughts are: Yes and Yes. I think I'd prefer a duplicate thing to be inserted cleanly next to the old thing instead of a replacement overwriting it badly.

zkat: No, and Yes: more precise matching increases the odds that we'll find a "reasonably accurate" match, and thus be able to replace existing code. The thing about this is that, a lot of times, it won't actually insert anything _next to_ where you intended: it'll shove it right in the middle of a definition, deeply nested and indented where it makes absolutely no sense, which is something we want to avoid.

src/services/mapCode.ts
```diff
@@ -0,0 +1,484 @@
...
+    }
+
+    if (a.getText() === b.getText()) {
+        return true;
+    }
+
+    return false;
+}
+
+function wipeNode(node: Mutable<Node>) {
```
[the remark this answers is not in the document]
zkat: I didn't either, but it took me a lot of digging to figure out how 😂 (being able to do this ended up being key to getting anything to work at all)

src/harness/fourslashInterfaceImpl.ts
```diff
@@ -239,6 +239,10 @@ export class VerifyNegatable {
     public uncommentSelection(newFileContent: string) {
         this.state.uncommentSelection(newFileContent);
     }
+
+    public mapCode(newFileContent: Record<string, string>, mappings: ts.MapCodeDocumentMapping[], updates?: ts.FileTextChanges[]): void {
+        this.state.mapCode(newFileContent, mappings, updates);
```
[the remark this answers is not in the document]
zkat: I don't understand how you're picturing the `=====`-separated sections working. Is there another test/set of tests that do this kind of thing? I don't see any tests that do anything like this.

src/services/mapCode.ts
```diff
@@ -0,0 +1,484 @@
...
+
+function placeNodeGroup(
+    originalFile: SourceFile,
+    changeTracker: ChangeTracker,
+    changes: NodeArray<Node>,
+    focusLocations?: TextSpan[],
+) {
+    // 1. Grab the first node, see what kind it is, to see whether we'll need
+    //    to expand our scope.
+    // 2. Get nodes for each focusLocation by finding the specific node that
```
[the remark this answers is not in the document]
zkat: I think I'm just going to delete the comments, yeah. They're probably adding more confusion than anything else.

src/services/mapCode.ts
```diff
@@ -0,0 +1,484 @@
...
+    // 2. If we have focusLocations, go over each one and:
+    //   a. Find the nearest class or interface declaration that contains the focusLocation.
+    let classOrInterface:
+        | ClassLikeDeclaration
+        | InterfaceDeclaration
+        | undefined;
+    if (!focusLocations || !focusLocations.length) {
+        classOrInterface = originalFile.statements.find(
+            stmt => isClassLike(stmt) || isInterfaceDeclaration(stmt),
+        ) as ClassLikeDeclaration | InterfaceDeclaration | undefined;
```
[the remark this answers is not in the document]
zkat: I couldn't find what you're referring to?

src/services/mapCode.ts
```diff
@@ -0,0 +1,484 @@
...
+            // `bar` will be the beginning of our range, and we want to
+            // have `baz` in the original be the end of our replacement
+            // range. `other` should be left intact, while `qux` should be
+            // "injected" before it.
+            //
+            // We do this by scanning the scope for all nodes that have a
+            // match in the incoming map request code, then sorting and
+            // picking first and last by position, and that will be our
+            // range for replacement.
+            const start = (scope as Block).statements.find(stmt => changes.some(node => matchNode(node, stmt)));
```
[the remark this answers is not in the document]
zkat: findAncestor only returns a Node. I'm not sure how to get around this?

src/services/mapCode.ts
zkat: correct. I'll clarify the comment.

src/services/mapCode.ts
```diff
@@ -0,0 +1,484 @@
...
+    }
+
+    if (a.kind === SyntaxKind.Constructor) {
+        return a.kind === b.kind;
+    }
+
+    if (isNamedDeclaration(a) && isNamedDeclaration(b)) {
+        const aName = a.name.getText();
+        const bName = b.name.getText();
+        return !!(aName && aName === bName);
```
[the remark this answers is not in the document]
zkat: `[""]() { return "I'm a method"; }` is an example.

But I guess in that case, we'd want them to match? idk, I feel weird about it. I'll just remove that requirement and let empties match each other.

src/harness/fourslashInterfaceImpl.ts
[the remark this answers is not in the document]
iisaduan: In `fourslashImpl.ts`, most of the test functions with `baseline` in them (and there are some tests that generate baselines that aren't explicitly labeled that way) will have some sort of formatting trick in them. `baselineGoToDefinition` or `baselineRename` could be good to look at for an idea, and I think I did something similar for the linked editing as well.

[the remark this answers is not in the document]
iisaduan: I haven't looked at the tests so far yet, but doing a baseline does make it easier in the long run, especially if you're adding a lot of tests, or there's a lot of output that you want to check, so that you don't have to worry about manually producing the correct test output for every one

src/services/mapCode.ts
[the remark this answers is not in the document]
sandersn: I vaguely remember using factory.update* functions instead, but I don't know if that's applicable here or not.

src/services/mapCode.ts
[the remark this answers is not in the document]
sandersn: I found it in corePublic.ts:162:

`export function find<T, U extends T>(array: readonly T[] | undefined, predicate: (element: T, index: number) => element is U, startIndex?: number): U | undefined;`

I misread this as using that function instead of Array.prototype.find. But, using `find` would mean that `classOrInterface = find(originalFile.statements, or(isClassLike, isInterfaceDeclaration));` would work

src/services/mapCode.ts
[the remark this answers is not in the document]
sandersn: This overload looks like it should propagate out the `isBlock(block) || isSourceFile(block)` check:
`export function findAncestor<T extends Node>(node: Node | undefined, callback: (element: Node) => element is T): T | undefined;`

At least, *inside* findAncestor's callback, `block` is narrowed correctly. If the first overload gets called, I think that narrowing should propagate out to `scope`. If you get rid of `scope`'s type annotation, does it get the type `SourceFile | Block`?

sandersn (changes requested):
I think that classes that have both class members and type members will drop type members in the current code. 

It would be nice to have a test for this too. Places where you can have mixed type/value declarations, like an abstract class, should show this off.

src/services/mapCode.ts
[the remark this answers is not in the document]
zkat: those seem to require that you know what kind of node you're editing, no? And the issue here is that I'm working with arbitrary nodes.

src/services/mapCode.ts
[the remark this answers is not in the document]
zkat: I still had to annotate stuff a bit but I think it's better now.

src/services/mapCode.ts
```diff
@@ -195,101 +182,86 @@ function placeClassNodeGroup(
...
-                changeTracker.insertNodesAfter(
+        // Otherwise, we insert this node at the end of the class/interface.
+        else if (isClassElement(newMember)) {
+            classEls.push(newMember);
+        }
+        else if (isTypeElement(newMember)) {
+            typeEls.push(newMember);
+        }
+    }
+    if (classEls.length || typeEls.length) {
```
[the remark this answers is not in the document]
zkat: done

zkat:
There, baseline tests in place. I didn't do the detailed diff thing because it got too gnarly and I didn't want to spend any more time on it. I'm happy enough with these.

zkat:
@sandersn @sheetalkamat @iisaduan can y'all take another look and lmk if there's anything lefty you'd like me to do?

tests/baselines/reference/mapCodeNestedForInsertion.mapCode.ts
```diff
@@ -0,0 +1,33 @@
...
+for (let j = 0; j < 10; j++) {
+  console.log("goodbye");
+  console.log("world");
+}
+
+// === MAPPED ===
+function foo() {
+    const x: number = 1;
+    const y: number = 2;
+    for (let i = 0; i < 10; i++) {
```
sandersn, line 24: same, although you could argue that you never want the LLM to arbitrarily change variable names

tests/baselines/reference/mapCodeNestedIfInsertion.mapCode.ts
```diff
@@ -0,0 +1,33 @@
...
+    const y: number = 2;
+    if (x === y) [||]{
+        console.log("hello");
+        console.log("you");
+    }
+    return 1;
+}
+
+// === INCOMING CHANGES ===
+if (y === x) {
```
sandersn, line 15: same; I think swapping the comparison order should be allowed.

sandersn (approved):
I think this is in a good place to ship.
1. Your baseline format is a lot better than the ones I've written.
2. I think we'll need fancier matching of `if`, `for`, `while`, etc, to allow slightly non-matching expressions. But that's a lot of work and should be done later in my opinion.
3. One small question about an array copy in a findLast call.

src/services/mapCode.ts
```diff
@@ -192,37 +186,30 @@ function placeClassNodeGroup(
...
-                    classOrInterface.members.length - 1
-                ],
-                factory.createNodeArray(appendedEls),
-            );
-        }
+    const firstMatch = classOrInterface.members.find(member => changes.some(change => matchNode(change, member)));
+    if (firstMatch) {
+        // can't be undefined here, since we know we have at least one match.
+        const lastMatch = findLast(
+            [...classOrInterface.members],
```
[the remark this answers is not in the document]
zkat: it's so I can use `findLast` :)

sheetalkamat (changes requested):

src/services/mapCode.ts
```diff
@@ -0,0 +1,341 @@
...
+    if (a.getText() === b.getText()) {
+        return true;
+    }
+
+    return false;
+}
+
+function wipeNode(node: Mutable<Node>) {
+    resetNodePositions(node);
+    node.parent = undefined!;
```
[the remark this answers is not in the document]
zkat: this is from a SourceFile, but the SourceFile is "fake". I need to wipe it/replace its parent so that node manipulation/insertion stuff doesn't get confused, so this is necessary.

src/services/mapCode.ts
```diff
@@ -0,0 +1,341 @@
...
+        throw new Error(
+            "Failed to find a class or interface to map the given code into.",
+        );
+    }
+
+    const firstMatch = classOrInterface.members.find(member => changes.some(change => matchNode(change, member)));
+    if (firstMatch) {
+        // can't be undefined here, since we know we have at least one match.
+        const lastMatch = findLast(
+            [...classOrInterface.members],
```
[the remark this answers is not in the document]
zkat: because `findLast` doesn't take a `NodeArray`.

src/server/session.ts
```diff
@@ -1904,6 +1905,71 @@ export class Session<TMessage = string> implements EventSender {
...
 
+    private mapCode(args: protocol.MapCodeRequestArgs): protocol.FileCodeEdits[] {
+        const formatOptions = this.getHostFormatOptions();
+        const preferences = this.getHostPreferences();
+        const projects = new Map<Project, MapCodeDocumentMapping[]>();
+        args.mappings.forEach(mapping => {
+            if (!mapping.file) {
+                // TODO(@zkat): We don't support mapping file-less code yet.
+                return;
+            }
```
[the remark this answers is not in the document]
zkat: The protocol isn't going to change, because it's externally defined and already being used by various partner teams, and we simply need to do something basic when this happens. CodeMappers isn't meant to be a fully reliable system. "Best effort" is good enough, and thus, I consider this better than doing nothing at all when a file is not present.

src/server/session.ts
```diff
@@ -1904,6 +1905,71 @@ export class Session<TMessage = string> implements EventSender {
...
+            const { file, project } = this.getFileAndProjectWorker(mapping.file, mapping.projectFileName);
```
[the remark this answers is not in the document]
zkat: This is how the protocol is defined (it's a protocol already being used by multiple teams), and the intention is for code mappers to be able to have a global view if they want it. In our case, we're not taking advantage of that yet, but we still have to stick to the protocol and how it feeds us information.

src/services/services.ts
```diff
@@ -3148,6 +3150,23 @@ export function createLanguageService(
...
```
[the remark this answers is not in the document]
zkat: what makes something a "syntax operation"?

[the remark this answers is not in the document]
zkat: actually, bringing this line back: `getValidSourceFIle` breaks without it.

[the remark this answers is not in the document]
sheetalkamat: You are suppose to use `const sourceFile = syntaxTreeCache.getCurrentSourceFile(fileName)`;

src/server/session.ts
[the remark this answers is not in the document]
sheetalkamat: It might be defined protocol else where but we are writing protocol for tsserver and its not defined. So we should be careful about what we add as we need to maintain it.

src/server/session.ts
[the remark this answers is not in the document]
sheetalkamat: Yes but tsserver protocol needs to ber per file and on your side you could be iterating instead.

src/services/mapCode.ts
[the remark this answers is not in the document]
sheetalkamat: Why. it extends ` ReadonlyArray<T>` so should be ok

sheetalkamat (changes requested):

src/services/mapCode.ts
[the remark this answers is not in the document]
zkat: Unfortunately, it is not. I don't really know why.

src/server/session.ts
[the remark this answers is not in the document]
zkat: The plan for codemappers is to eventually be able to have multi-file views into the project and operate on multiple of them at a time. tsserver already does things that involve multiple files (various code actions) so I don't understand the problem.

As you say elsewhere, if we're going to maintain a protocol, we should make sure it's what we need, and one that's restricted to one file will only need to be deprecated and replaced later when that part is implemented.

src/server/session.ts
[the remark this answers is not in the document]
zkat: As above: we should be careful to understand what the plan is with this protocol, and that will include a "global" view, and code that's coming in without a target file, which may be mapped to any file we decide to map to.

Establishing a one-file-only protocol now means we'll need to throw away the public api later and add a second one just because we didn't want to do it this time around.

src/services/types.ts
```diff
@@ -1571,6 +1573,13 @@ export interface OutliningSpan {
     kind: OutliningSpanKind;
 }
 
+/** @internal */
+export interface MapCodeDocumentMapping {
+    fileName: string;
```
sheetalkamat, line 1578:
```suggestion
```
So that its inline with other service APIs

sheetalkamat: In fact you could get rid of this and instead add these as parameters to API as well

src/services/mapCode.ts
```diff
@@ -0,0 +1,330 @@
...
+                        nodes,
+                        flattenedLocations,
+                    );
+                }
+            },
+        );
+    }
+    catch (e) {
+        host.error?.(`mapCode: ${e}`);
+        return [];
```
sheetalkamat, line 73: Use `emptyArray` in places of sending `[]` everywhere

sheetalkamat (changes requested):

src/services/mapCode.ts
```diff
@@ -0,0 +1,330 @@
...
+                        sourceFile,
+                        changeTracker,
+                        nodes,
+                        flattenedLocations,
+                    );
+                }
+            },
+        );
+    }
+    catch (e) {
```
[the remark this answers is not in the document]
zkat: I'll add a test for this. This currently happens when trying to place something inside a class, but not finding a class.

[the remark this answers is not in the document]
sheetalkamat: what kind of exception. We prefer not to throw exception and return undefined and handle accordingly instead of try catch all over for perf reasons.

src/services/types.ts
```diff
@@ -1588,6 +1590,13 @@ export interface OutliningSpan {
     kind: OutliningSpanKind;
 }
 
+/** @internal */
+export interface MapCodeDocumentMapping {
+    fileName: string;
+    focusLocations?: TextSpan[][];
+    contents: string[];
+}
```
sheetalkamat, line 1598:
```suggestion
```

tests/baselines/reference/api/typescript.d.ts
```diff
@@ -1769,6 +1770,46 @@ declare namespace ts {
             export interface InlayHintsResponse extends Response {
                 body?: InlayHintItem[];
             }
+            export interface MapCodeRequestArgs {
```
sheetalkamat, line 1773: Need to update this baseline

src/services/mapCode.ts
```diff
@@ -0,0 +1,330 @@
...
+    ScriptTarget,
+    some,
+    SourceFile,
+    Statement,
+    SyntaxKind,
+    textChanges,
+    TextSpan,
+    TypeElement,
+    UserPreferences,
+} from "./_namespaces/ts";
```
sheetalkamat, line 41: Fix lint errors

src/harness/fourslashImpl.ts
```diff
@@ -4510,6 +4510,32 @@ export class TestState {
 
         this.verifyCurrentFileContent(newFileContent);
     }
+
+    public baselineMapCode(
+        { fileName, pos }: Range,
```
sheetalkamat, line 4515: You should be using pos and end both right since thats the focus location marked using range. 
We need test cases where there are multiple focus locations with different lengths and as mentioned in API with different priorities too.

src/harness/fourslashImpl.ts
```diff
@@ -4510,6 +4510,32 @@ export class TestState {
 
         this.verifyCurrentFileContent(newFileContent);
     }
+
+    public baselineMapCode(
+        { fileName, pos }: Range,
+        changesFilename = "/incomingChanges",
```
sheetalkamat, line 4516: please make this mandatory and use extensions like ".ts" or "js"

src/services/mapCode.ts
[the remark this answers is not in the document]
sheetalkamat: Please provide test case name, its hard to know after the squash commit.

sheetalkamat (changes requested):
We recommend that PRs should squash commits after requesting review. Please add more commits instead of squashing as it becomes difficult to track history. We merge PR as a squash commit so it wont affect the merge commit. But squashing commits between reviews makes it difficult to review.

src/harness/fourslashImpl.ts
zkat: the current code mapper only uses the start location. I can add a test for multiple focusLocations, though.

zkat: I've added a test to cover this, `mapCodeReplaceUsingRange`

src/harness/fourslashImpl.ts
```diff
@@ -4510,6 +4510,58 @@ export class TestState {
...
+            if (!fileName) {
+                fileName = fn;
+            }
+            return [{ start: pos, length: end - pos }];
+        });
+        if (!fileName) {
+            throw new Error("No ranges passed in, or something went wrong.");
+        }
+        let before = this.getFileContent(fileName);
+        focusLocations.sort((a, b) => a[0].start - b[0].start);
```
sheetalkamat, line 4530: Is this requirement of API ? Does that need to be in the API jsdoc then?

src/services/mapCode.ts
```diff
@@ -0,0 +1,329 @@
...
+ */
+function parse(content: string): NodeArray<Node> {
+    // We're going to speculatively parse different kinds of contexts to see
+    // which one makes the most sense, and grab the NodeArray from there. Do
+    // this as lazily as possible.
+    const nodeKinds = [
+        {
+            parse: () =>
+                createSourceFile(
+                    "__mapcode_content_nodes.ts",
```
sheetalkamat, line 88: Doesnt this need to have same extension as sourceFile passed in as well as scriptTarget same as sourceFile  to ensure they are compatible

src/harness/fourslashImpl.ts
```diff
@@ -4510,6 +4510,58 @@ export class TestState {
...
         this.verifyCurrentFileContent(newFileContent);
     }
+
+    public baselineMapCode(
+        ranges: Range[],
+        changesFilename: string,
+    ): void {
+        const changes = this.getFileContent(changesFilename);
+        let fileName: string | undefined;
+        const focusLocations = ranges.map(({ fileName: fn, pos, end }) => {
```
sheetalkamat, line 4520: instead of sorting these ranges here can you make sure tests pass in required focus locations and they are passed as is to LS so that tests can be more explicit on order they are sending the ranges in.

sheetalkamat: Also the ranges need to be optional and the here should take a file name so that you can test when there are no focus locations (they are marked optional) - Again test for the same pls.

src/services/mapCode.ts
```diff
@@ -0,0 +1,329 @@
...
+/** @internal */
+export function mapCode(
+    sourceFile: SourceFile,
+    contents: string[],
+    focusLocations: TextSpan[][] | undefined,
+    host: LanguageServiceHost,
+    formatContext: formatting.FormatContext,
+    preferences: UserPreferences,
+): FileTextChanges[] {
+    try {
```
sheetalkamat, line 53: Please remove this try catch now that there aren't any exceptions. We don't want to do this as there are other things you need to worry about if catching exception like cancellation token etc and LS shouldn't be in charge of doing that. If exception happens the tsserver normal exception handling should come in play

tests/cases/fourslash/fourslash.ts
```diff
@@ -212,6 +212,20 @@ declare namespace FourSlashInterface {
...
+    export interface FileTextChanges {
+        fileName: string;
+        textChanges: readonly TextChange[];
+        isNewFile?: boolean;
+    }
+    export interface MapCodeDocumentMapping {
+        fileName?: string;
+        focusLocations?: TextSpan[][];
+        contents: string[];
+    }
```
sheetalkamat, line 228: why are these here ?

tests/cases/fourslash/mapCodeClassInvalidClassMember.ts
```diff
@@ -0,0 +1,13 @@
+///<reference path="fourslash.ts"/>
+
+// @Filename: /incomingChanges.ts
+//// if (false) {
+////     return "hello";
+//// }
+////
```
sheetalkamat, line 7: This should not be a file but input to test since problem is that this will be part of the program and you are not looking for that. So it should be verify.baselineMapCode(fileName, focusLocations, changes) 

Then add some test cases where fileName is `index.js` instead of `index.ts`

src/services/types.ts
```diff
@@ -683,6 +683,8 @@ export interface LanguageService {
 
     getSupportedCodeFixes(fileName?: string): readonly string[];
 
+    /** @internal */ mapCode(fileName: string, contents: string[], focusLocations: TextSpan[][] | undefined, formatOptions: FormatCodeSettings, preferences: UserPreferences, updates?: FileTextChanges[]): readonly FileTextChanges[];
```
sheetalkamat, line 686:
```suggestion
    /** @internal */ mapCode(fileName: string, contents: string[], focusLocations: TextSpan[][] | undefined, formatOptions: FormatCodeSettings, preferences: UserPreferences): readonly FileTextChanges[];
```

src/services/services.ts
```diff
@@ -3165,6 +3166,17 @@ export function createLanguageService(
         return InlayHints.provideInlayHints(getInlayHintsContext(sourceFile, span, preferences));
     }
 
+    function mapCode(sourceFile: string, contents: string[], focusLocations: TextSpan[][] | undefined, formatOptions: FormatCodeSettings, preferences: UserPreferences = emptyOptions): FileTextChanges[] {
```
sheetalkamat, line 3169:
```suggestion
    function mapCode(sourceFile: string, contents: string[], focusLocations: TextSpan[][] | undefined, formatOptions: FormatCodeSettings, preferences: UserPreferences): FileTextChanges[] {
```

sheetalkamat (changes requested):

src/harness/fourslashImpl.ts
zkat: No. This is necessary to render the baselines properly (as in, figure out the right padding while writing all the `[|` and `|]` things out). It is irrelevant to the API.

src/services/mapCode.ts
zkat: nope. The node erasure we do later makes sure this isn't an issue.

tests/cases/fourslash/mapCodeClassInvalidClassMember.ts
zkat: So you want me to write an inline string with code instead of doing the fourslash facilities? I was originally doing what you're asking, and was asked to move to doing this instead.

src/harness/fourslashImpl.ts
sheetalkamat: Then please move that for displaying and not for passing it to the API. API should be called with all the input parameters as is. The baselining code can do all sorts of sorting

tests/cases/fourslash/mapCodeClassInvalidClassMember.ts
sheetalkamat: I dont understand. Doesnt LS API take sourceFile, focus locations and text? 

I would think that it translates to verify(index.ts, focus ranges, text that you have right now in incoming.ts) is that not it?  I am saying the text of incoming.ts shouldnt be on the vfs in the test but passed to mapCode API as text itself. 

What i meant by different extension is to have the focus locations in ".js" for example

src/services/mapCode.ts
sheetalkamat: Hmm wierd it just seems not right to be to use random target and ".ts" extension to parse a file.. it feels like it should use same scanner and parser as input sourceFile's target and language version

tests/cases/fourslash/fourslash.ts
zkat: leftovers

src/harness/fourslashImpl.ts
zkat: I'll write a test with no ranges, but the sorting is not for testing functionality, but for properly outputting the baselines. It's irrelevant to the test.

src/services/mapCode.ts
```diff
@@ -0,0 +1,322 @@
...
+            return bod;
+        }
+        // We only want to keep the ones that have some kind of body.
+        else if (bod.length) {
+            // Otherwise, we'll need to look at others.
+            parsedNodes.push({ sourceFile, body: bod });
+        }
+    }
+    // Heuristic: fewer errors = more likely to be the right kind.
+    const { body } = parsedNodes.sort(
```
sheetalkamat, line 115: We are using parseDiagnostics here for heuristics so i strongly feel we should be using ScriptTarget and LanguageVersion of sourceFile for the temp sourcefiles that are created.

src/harness/fourslashImpl.ts
```diff
@@ -4510,6 +4510,52 @@ export class TestState {
 
         this.verifyCurrentFileContent(newFileContent);
     }
+
+    public baselineMapCode(
+        ranges: Range[],
```
sheetalkamat, line 4515: Please take Range[][] here instead

Add tests where we are passing [[Range0, range1], [range2]] 
Add tests when active file name is "index.js"

sheetalkamat (changes requested):

sheetalkamat (approved):

PR #55418, opened 2023-08-17
Improve CI baseline check

jakebailey:
This check previously didn't consider the diff kinds when printing and considered them all to be "unused" even when they're actually missing or modified. This modifies these checks to better distinguish the kinds of baseline changes needed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -240,7 +240,7 @@ jobs:
       - name: Self build
         run: npx hereby build-src --built
 
-  unused-baselines:
+  baselines:
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -258,11 +258,22 @@ jobs:
...
+              echo "$2:"
+              git diff --staged --name-only --diff-filter=$1
+            fi
+          }
+
+          if ! git diff --staged --exit-code --quiet; then
+            print_diff ACR "Missing baselines"
+            print_diff MTUXB "Modified baselines"
+            print_diff D "Unused baselines"
+            exit 1
```
MariaSolOs, line 278: Out of curiosity, why `exit 1` here?

.github/workflows/ci.yml
```diff
@@ -258,11 +258,22 @@ jobs:
...
-      - name: Check for unused baselines
+      - name: Check baselines
         run: |
-          if ! git diff --exit-code --quiet; then
-            echo "Unused baselines:"
-            git diff --exit-code --name-only
+          function print_diff() {
+            if ! git diff --staged --exit-code --quiet --diff-filter=$1; then
+              echo "$2:"
+              git diff --staged --name-only --diff-filter=$1
```
MariaSolOs, line 270: Are all `diff-filter`s single word? With the current use of `print_diff` there's no need to quote the expansion, but do you think it would make sense to quote it to avoid possible issues in the future?

.github/workflows/ci.yml
jakebailey: If we don't do this, the step won't fail.

.github/workflows/ci.yml
jakebailey: They're restricted to only the single letters I have written here, concatenated together. This should be plenty future proof as this is a git feature.

MariaSolOs (approved):

PR #55419, opened 2023-08-17
Make formatting settings in settings template apply to all filetypes

jakebailey:
If you have the GitHub Actions extension installed, the filetype for these files stops being yaml and starts being "github-actions-workflow", which prevents formatting. Add it to the template in case people want to format those in the editor.

### DIFF
--- a/.vscode/settings.template.json
+++ b/.vscode/settings.template.json
@@ -4,7 +4,7 @@
     // To use the locally built compiler, after 'npm run build':
     // "typescript.tsdk": "built/local"
 
-    "[typescript][javascript][yaml]": {
+    "[typescript][javascript][yaml][github-actions-workflow]": {
         "editor.formatOnSave": true,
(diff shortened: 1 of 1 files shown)


jakebailey:
Though honestly, I'm considering whether or not we should just set the default formatter in general to dprint, which should work fine.

jakebailey:
I've switched this to just set dprint as the default for everything; I'm fairly certain this is fine because dprint will just ignore any file that it doesn't know. I haven't been able to cause VS Code to complain or misformat anything in my testing.

andrewbranch (approved):

andrewbranch:
`hereby sync-vscode-settings-template`

jakebailey:
That'd sure be cool, and technically possible as we already depend on `json-parser` which is just VS Code's JSON parser/editor pulled out to npm.

PR #55426, opened 2023-08-17
Remove lib.d.ts line info in inlay hint baselines

MariaSolOs:
Had some issues in #55141 because of including line information from `lib.d.ts` files, so this PR attempts to make those baselines more robust.

### DIFF
--- a/src/harness/fourslashImpl.ts
+++ b/src/harness/fourslashImpl.ts
@@ -839,6 +839,15 @@ export class TestState {
         const fileName = this.activeFile.fileName;
         const hints = this.languageService.provideInlayHints(fileName, span, preferences);
         const annotations = ts.map(hints.sort(sortHints), hint => {
+            if (hint.displayParts) {
+                hint.displayParts = ts.map(hint.displayParts, part => {
+                    if (part.file && /lib(?:.*)\.d\.ts$/.test(part.file)) {
+                        part.span!.start = -1;
+                    }
+                    return part;
+                });
+            }
+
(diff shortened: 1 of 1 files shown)


jakebailey:
No test change; is this because your other PR is what introduces them? Was hoping to see an example that shows it working but otherwise it's probably fine.

src/harness/fourslashImpl.ts
```diff
@@ -839,6 +839,15 @@ export class TestState {
         const fileName = this.activeFile.fileName;
         const hints = this.languageService.provideInlayHints(fileName, span, preferences);
         const annotations = ts.map(hints.sort(sortHints), hint => {
+            if (hint.displayParts) {
+                hint.displayParts = ts.map(hint.displayParts, part => {
+                    if (part.file && /lib(?:.*)\.d\.ts$/.test(part.file)) {
+                        part.span!.start = -1;
```
jakebailey, line 845: `span` is definitely defined?

MariaSolOs:
> No test change; is this because your other PR is what introduces them? Was hoping to see an example that shows it working but otherwise it's probably fine.

Yep. My type hints PR will have the changes. I already tried cherry picking this locally and it worked.

src/harness/fourslashImpl.ts
MariaSolOs: Yes. If the file is defined, the span definitely is.

jakebailey (approved):

PR #53417, opened 2023-03-21
Allow getter and setter types to be unrelated

andrewbranch:
Fixes #43662

The one thing to note/discuss here is that to create unrelated accessors, both the getter and setter must have explicit annotations, because the setter’s type annotation will be used for an unannotated getter return type:

```ts
// Error:
const unannotatedGetter = {
  get p() {
    return 0;
//  ^^^^^^^^ Type 'number' is not assignable to type 'string'
  },
  set p(value: string) {
    // ...
  },
};

// Ok:
const bothAnnotated = {
  get p(): number {
    return 0;
  },
  set p(value: string) {
    // ...
  }
};
```

I think this is probably better than making it too easy to accidentally create unrelated accessor types, but perhaps it’s not obvious from the error message how that unrelated accessor types are possible.

PR #55442, opened 2023-08-20
Fixed declaration emit of object literals withs divergent accessors

Andarist:
cc @andrewbranch as divergent setters became allowed since https://github.com/microsoft/TypeScript/pull/53417


tests/baselines/reference/declarationEmitObjectLiteralAccessorsJs1.js
```diff
@@ -0,0 +1,80 @@
...
+    set x(a: number);
+};
+export namespace obj3 {
+    let x_1: string;
+    export { x_1 as x };
+}
+export namespace obj4 {
+    let x_2: number;
+    export { x_2 as x };
+}
```
Andarist, line 80: We might notice that comments are only preserved here in `obj2`. There is probably a complication related to this here since JSDoc comments here will always refer to `@param` and `@returns` and those don't make sense for variable declarations - so should the comment be reconstructed/trimmed in such a case? 

Either way, I think that's a separate issue and I would prefer not to work on it here.

src/compiler/checker.ts
```diff
@@ -7246,6 +7246,30 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
             context.enclosingDeclaration = saveEnclosingDeclaration;
             context.approximateLength += symbolName(propertySymbol).length + 1;
+
+            if (propertySymbol.flags & SymbolFlags.Accessor) {
+                const writeType = getWriteTypeOfSymbol(propertySymbol);
+                if (propertyType !== writeType) {
+                    const getterDeclaration = getDeclarationOfKind<GetAccessorDeclaration>(propertySymbol, SyntaxKind.GetAccessor)!;
+                    const getterSignature = getSignatureFromDeclaration(getterDeclaration);
+                    typeElements.push(
+                        setCommentRange(
```
Andarist, line 7256: Those two are using direct calls to `setCommentRange` instead of the existing `preserveCommentsOn` helper. The only added benefit of `preserveCommentsOn` is that it handles `JSDocPropertyTag`s. However, those only matter when serializing typedefs, as far as I can tell. We are serializing an "inferred" type of the object literal here - its property declarations can't be sourced from `@property`, right?

andrewbranch:
@typescript-bot pack this

src/compiler/checker.ts
```diff
@@ -9720,7 +9744,15 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
-                    every(getPropertiesOfType(typeToSerialize), p => isIdentifierText(symbolName(p), languageVersion));
+                    every(getPropertiesOfType(typeToSerialize), p => {
+                        if (!isIdentifierText(symbolName(p), languageVersion)) {
+                            return false;
+                        }
+                        if (!(p.flags & SymbolFlags.Accessor)) {
+                            return true;
+                        }
+                        return getNonMissingTypeOfSymbol(p) === getWriteTypeOfSymbol(p);
+                    });
```
andrewbranch, line 9755: What’s a test exercises this new code?

Andarist: IIRC this is executed by the JS-related test case: https://github.com/microsoft/TypeScript/pull/55442/files#diff-1624f807bc31e4fbb911a4979b0f28a8c7a415c3f743d047af32cecf23b58ffa

jakebailey:
Uhh, I guess I broke perf.

jakebailey:

andrewbranch (approved):
Looks good as long as perf comes back clean

tests/baselines/reference/declarationEmitObjectLiteralAccessorsJs1.js
```diff
@@ -0,0 +1,80 @@
...
+     */
+    get x(): string;
+    /**
+     * my awesome setter
+     * @param {number} a
+     */
+    set x(a: number);
+};
+export namespace obj3 {
+    let x_1: string;
```
weswigham, line 74: This is going to change because I merged your other jsdoc declaration emit PR, right? I think this is ready to go once it's sync'd, though - perf did come back clean, after all.

weswigham (approved):

tests/baselines/reference/declarationEmitObjectLiteralAccessorsJs1.js
Andarist: Yep, this was impacted by that other PR. I just synced with `main` so this should now be mergeable.

Issue #55439, opened 2023-08-20
type gets widened on declaration emit when inferred in destructuring context


h-joo:
### 🔎 Search Terms

declaration, widened, d.ts, destructuring

### 🕗 Version & Regression Information

- This is the behavior in every version I tried, and I reviewed the FAQ for entries

### ⏯ Playground Link

https://www.typescriptlang.org/play?jsx=0&module=1&ts=5.1.6#code/GYVwdgxgLglg9mABMOcAUBKAXIg3gTxwEYBfPAKESsQCcBTKEGpXRQxIxE878ugDwAOcGlEQQEAZzGt8iALyIADFwXJUmANxA

### 💻 Code

```ts
function foo(): {y: 1} {
    return { y: 1 }
}

export const { y = 0 } = foo();
```

### 🙁 Actual behavior

`y` is `number` in `d.ts` files, although the language service says `y` is `1|0`

### 🙂 Expected behavior

`y` is `1|0` in `d.ts` files.

### Additional information about the issue

Without the `= 0` assignment `y` properly becomes type `1` in d.ts

MartinJohns:
Workaround: `0 as const`.

dragomirtitian:
@Andarist I found a very similar issue. In the code below `x` is typed in ts as `0 | 1`, but is typed as `number` in declarations.
```ts
export const x = Math.random() ? 1 : 0 // 0 | 1
```
[Playground](https://www.typescriptlang.org/play?ts=5.2.2#code/KYDwDg9gTgLgBAYwgOwM7xHAvHAsgQxgAsA6KfZAEwgFsAKASjgH44BGOALjgAYBYAFAB6IXDHiJcAHrNBggJYAzOnExZ1cAExMA3oImCAvkA)

Will your fix  work for this example as well ?

Andarist:
Yes, I think so - what Jake commented on here is basically the same as your case above: https://github.com/microsoft/TypeScript/pull/55445/files#r1300377161

dragomirtitian:
@Andarist Oh. Yeah. Forgot about that. Thanks 😅

PR #55445, opened 2023-08-21
Do not widen literal types when emitting declarations

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/55439

### DIFF
--- a/tests/baselines/reference/ambientConstLiterals.js
+++ b/tests/baselines/reference/ambientConstLiterals.js
@@ -74,5 +74,5 @@ declare const c9: {
 declare const c10: number[];
 declare const c11: string;
 declare const c12: number;
-declare const c13: string;
-declare const c14: number;
+declare const c13: "abc" | "def";
+declare const c14: 123 | 456;
--- a/tests/baselines/reference/strictFunctionTypes1.js
+++ b/tests/baselines/reference/strictFunctionTypes1.js
@@ -82,7 +82,7 @@ declare function fs(x: string): void;
 declare function fx(f: (x: "def") => void): void;
 declare const x1: (x: string) => void;
 declare const x2 = "abc";
-declare const x3: string;
+declare const x3: "def" | "abc";
 declare const x4: Func<string>;
 declare const never: never;
 declare const x10: string;
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -47230,7 +47230,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         // Get type of the symbol if this is the valid symbol otherwise get type at location
         const symbol = getSymbolOfDeclaration(declaration);
         let type = symbol && !(symbol.flags & (SymbolFlags.TypeLiteral | SymbolFlags.Signature))
-            ? getWidenedLiteralType(getTypeOfSymbol(symbol))
+            ? getTypeOfSymbol(symbol)
             : errorType;
         if (
             type.flags & TypeFlags.UniqueESSymbol &&
--- a/tests/baselines/reference/declarationEmitBindingPatterns2.js
+++ b/tests/baselines/reference/declarationEmitBindingPatterns2.js
@@ -1,0 +1,16 @@
+//// [tests/cases/compiler/declarationEmitBindingPatterns2.ts] ////
+
+//// [declarationEmitBindingPatterns2.ts]
+// https://github.com/microsoft/TypeScript/issues/55439
+
+function foo(): { y: 1 } {
+  return { y: 1 };
+}
+
+export const { y = 0 } = foo();
+
+
+
+
+//// [declarationEmitBindingPatterns2.d.ts]
+export declare const y: 1 | 0;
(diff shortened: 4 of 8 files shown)


tests/baselines/reference/strictFunctionTypes1.js
```diff
@@ -82,7 +82,7 @@ declare function fs(x: string): void;
 declare function fx(f: (x: "def") => void): void;
 declare const x1: (x: string) => void;
 declare const x2 = "abc";
-declare const x3: string;
+declare const x3: "def" | "abc";
```
Andarist, line 85: I understand that there is this hidden widening thing encoded in the type and that it actually might widen when assigned to a mutable declaration. I think though that most people are not aware of this and it's *more* surprising that a type widened in the emitted declaration - people never audit them until something goes wrong.

The current behavior is also surprising in some cases. Let's expand on this very case here:
```ts
declare function f3<T>(obj: T, f1: (x: T) => void, f2: (f: (x: T) => void) => void): T;
declare function fo(x: Object): void;
declare function fx(f: (x: "def") => void): void;

const x3 = f3("abc", fo, fx);  // "abc" | "def"

let other: typeof x3 = x3
other = 'other' // error
```

As we can see we can't assign `'other'` to `other` but yet in the emitted declaration file we can see this:
```ts
declare const x3: string;
declare let other: typeof x3;
```

It means that in practice we can end up with an error locally that wouldn't get reported if the same code would get used through the declaration file. I find that problematic. 

The same could be said about the widening thing because "locally" an assignment would widen but if we change the declaration emit then the equivalent assignment would not widen since the widening behavior can't be encoded in the declaration file. I think though that's a smaller problem and a better tradeoff.

tests/baselines/reference/ambientConstLiterals.js
```diff
@@ -74,5 +74,5 @@ declare const c9: {
 declare const c10: number[];
 declare const c11: string;
 declare const c12: number;
-declare const c13: string;
-declare const c14: number;
+declare const c13: "abc" | "def";
+declare const c14: 123 | 456;
```
jakebailey, line 78: Interestingly I think this case came up during our talks about `isolatedDeclarations`. The code for these two is:

```ts
const c13 = Math.random() > 0.5 ? "abc" : "def";
const c14 = Math.random() > 0.5 ? 123 : 456;
```

@RyanCavanaugh @DanielRosenwasser

jakebailey: Oh, I just realized who reported the bug, it is directly related!

jakebailey:
I don't know if this will actually net any test changes but:

@typescript-bot test top200
@typescript-bot run dt
@typescript-bot pack this

jakebailey:
@typescript-bot run dt

jakebailey (approved):
This seems okay and what we are trying to go for to make `isolatedDeclarations` easier, correct?

tests/baselines/reference/strictFunctionTypes1.js
weswigham: > and it's more surprising that a type widened in the emitted declaration - people never audit them until something goes wrong.

Correct! And in most cases in declaration emit, the way we handle this is by preserving the literal initializer, rather than serializing a type annotation - that way the declaration has the same widening behavior as the original source. Eg, if you write

```ts
export const six = 6;
```
that is also, verbatim, the declaration emit. Unfortunately, right now this only covers simple cases of widening types - a heritage from our older, much simpler string-based declaration emitter.

IMO, rather than making an equally incorrect declaration in the other direction (preserving the literal types, rather than the widened type), we need to, once again, preserve the initializer to preserve the literal-widening behavior. In this case, that should mean making initializers with the same behavior.

Take the original issue:
```ts
function foo(): {y: 1} {
    return { y: 1 }
}

export const { y = 0 } = foo();
```

If we serialize it to
```ts
declare const _expr: {y: 1};
export declare const { y = 0 } = _expr;
```
`y` now behaves identically in the declaration file as it does in the original source file. Do note, this means that we'll need to allow identifiers in initializers in declaration files - something currently an error - but I think that's fine, especially since, despite the error, old versions of TS will still typecheck it correctly.

weswigham (changes requested):

jakebailey (changes requested):
While talking to @weswigham, I thought of this example: 

```ts
// @filename: a.ts
export const value = Math.random() > 0.5 ? "foo" : "bar";

// @filename: b.ts
import * as a from "./a";
export const x1 = a.value;
export let x2 = a.value;
```

When you run tsc from this PR, you get:

```ts
// @filename: a.d.ts
export declare const value: "foo" | "bar";

// @filename: b.d.ts
export declare const x1: "foo" | "bar";
export declare let x2: string;
```

Now, remove `a.ts` to simulate using it externally (or probably even in build mode), then rerun tsc, and you get:

```ts
// @filename: b.d.ts
export declare const x1: "foo" | "bar";
export declare let x2: "foo" | "bar";
```

That seems pretty bad. Not sure how to proceed at this point.

Though, weren't we banning ternaries in `isolatedDeclarations` anyway?

Andarist:
I could implement @weswigham's suggestion but - like @jakebailey mentioned - it's unclear what to do about ternaries. I'd prefer to work on the further improvements to this PR in one go so I'm waiting for further feedback/decision about this.

> Though, weren't we banning ternaries in isolatedDeclarations anyway?

I think that the behavior in question is confusing even outside of this mode. Perhaps it should be considered first without taking this mode into account. I imagine that a solution for this issue would automatically make `isolatedDeclarations` able to reflect the changes though.

jakebailey:
FWIW I think this PR can be closed, just because I don't think we're going to be able to fix this by just not widening; we need some other mechanism to do this.

PR #55446, opened 2023-08-21
Allow rest elements to follow other rest elements in tuple types

Andarist:
I think that this restriction is somewhat artificial and doesn't bring a lot of value. Those are allowed and work OK:
```ts
type Ok1 = [...Array<number>, ...Array<string>]

type Alias = string[]
type Ok2 = [...number[], ...Alias]
```

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -16307,6 +16307,23 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         return false;
     }
 
+    function isTupleNormalizable(node: TupleTypeNode) {
+        let hasSeenRest = false;
+        for (const element of node.elements) {
+            const flags = getTupleElementFlags(element);
+            if (flags & ElementFlags.Variadic) {
+                return true;
+            }
+            if (flags & ElementFlags.Rest) {
+                if (hasSeenRest) {
+                    return true;
+                }
+                hasSeenRest = true;
+            }
+        }
+        return false;
+    }
+
     function getTypeFromArrayOrTupleTypeNode(node: ArrayTypeNode | TupleTypeNode): Type {
         const links = getNodeLinks(node);
         if (!links.resolvedType) {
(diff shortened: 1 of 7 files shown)


ssalbdivad:
Thoughts on going the other direction and erroring in both cases?

`[...number[], ...string[]]` does imply more than just `(number | string)[]`. Barring a robust sequential representation and some advanced narrowing to make it usable, it feels safer just to error rather than implicitly weaken the constraints.

Andarist:
I dont rly think that multiple variadic elements in a tuple are usable. Even trailing non-variadic elements are already quirky to handle at runtime

ssalbdivad:
@Andarist I'm not suggesting `A` and `B` not continue to work as they do currently, just that `C` not be implicitly weakened to a union of its elements.

Whether TS supports it or not (and I understand why doing so doesn't seem worthwhile), `[...number[], ...string[]]` has a distinct meaning as a type that is stricter than `(number | string)[]`. I don't see a good reason to allow definitions like that that can't actually be enforced over requiring the weaker form directly.

Andarist:
That still would be a breaking change for cases like this:
```ts
type JoinArrays<T1 extends readonly unknown[], T2 extends readonly unknown[]> = [...T1, ...T2]
type Result1 = JoinArrays<string[], number[]> // (string | number)[]
```

This behavior is rather useful and we can't disallow `JoinArrays` being a thing since very useful things would get disallowed:
```ts
type JoinArrays<T1 extends readonly unknown[], T2 extends readonly unknown[]> = [...T1, ...T2]
type Result2 = JoinArrays<[number, number], [boolean]> // [number, number, boolean]
```

We have *variadic* elements here (and not rest ones). Technically it's trivial~ to allow this while disallowing multiple rests. Syntactically they both look the same though. I find it a hard sell for only one of them to work. If the generic variant with variadic elements works and we can "join" array types using it then IMHO it only makes sense for the non-generic case to work in the same way.

Andarist:
superseded by https://github.com/microsoft/TypeScript/pull/57031/commits/70715eae3ba4ded0ea7f4f0b3792aeae35944c8f

Issue #55203, opened 2023-07-30
Wrong return type for union of overloaded functions


maledies:
# Bug Report

### 🔎 Search Terms

union, overload, function, return type

### 🕗 Version & Regression Information

- This is the behavior in every version I tried (4.4, 4.8, 5.0, 5.1, 5.2beta)

### ⏯ Playground Link

[Playground link with relevant code](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgMJwDYbgIwxAHgDk4BbFCAD0hABMBnZesKUAcwD5kBvAKAEgAFAEoAXMgAGAEm4lyAX2QB3YGAAWAewCuYZMFoSBg-eJBbSOaGMky5ERSvV6Dveb160ICbFBQINIMzIMOLoWLj4BABEAIJRXAA+aJjYeIRRAELxvP6Bur70Whi6ALzBggCMAEwAzMLIAPQNBUVgjKDIscqqas5RCZnd6tq6+lFAA)

### 💻 Code

```ts
interface Callable<Name extends string> {
	(): `${Name} without id`
	(id: number): `${Name} with id`
}

declare const f: Callable<"A"> | Callable<"B">
const result = f(123) //results in "A with id"|"B without id"
```

### 🙁 Actual behavior

`result` is inferred as `"A with id"|"B without id"`.

`"B with id"` is completely missing and `"B without id"` should not appear since we used a parameter.

### 🙂 Expected behavior

`result` is `"A with id"|"B with id"`.

MartinJohns:
I believe it's the same as: https://github.com/microsoft/TypeScript/issues/55101#issuecomment-1646025879

> The contextual type for the return type of the function expression in this case comes from the first overload, since it's the first one that appears to match on arity.

The first overload that matches the arity is the first one, without any arguments (providing too many arguments is fine).

You can fix your code by changing the order of overloads from most specific to least specific:

```typescript
interface Callable<Name extends string> {
	(id: number): `${Name} with id`
	(): `${Name} without id`
}
```

RyanCavanaugh:
There's no contextual type or widening here. Definitely better to write it with the overloads in the other order, though

PR #55447, opened 2023-08-21
Prefer non-partial signature match in non-generic case

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/55203


weswigham:
@typescript-bot test this
@typescript-bot run dt
@typescript-bot test top100

weswigham (approved):
I think this is good, but since it affects overload resolution (admittedly in an edge case when we're talking about union overloads), it might have knock-on effects in the wild, so we'll have to make sure the extended test suites come back OK.

PR #53961, opened 2023-04-22
Fix double error on invalid delete of readonly property

galabra:
Fixes #53082

Whenever we detect there's an error of type `The_operand_of_a_delete_operator_cannot_be_a_read_only_property`, we don't need to check for the `The_operand_of_a_delete_operator_must_be_optional` error.

PR #55449, opened 2023-08-21
Fix double error on invalid delete of readonly property

### LINKED ISSUE #53082: Double error on invalid delete of readonly property
# Bug Report

### 🔎 Search Terms

operand delete operator double diagnostic error

- This changed between versions 3.9 and 4.0

Before 4.0, we only gave one error here. In 4.0 and above, we give two errors.

### ⏯ Playground Link

[Playground Link](https://www.typescriptlang.org/play?ts=4.9.5#code/MYGwhgzhAEBiD29oG8C+AoAJgUxNgLtnIgHQB2YAttgNxA)

### 💻 Code

```ts
class Foo {}
delete Foo.name;
```

### 🙁 Actual behavior

Two errors

### 🙂 Expected behavior

One error

shinichy:
Fixes #53082. This PR added a unit test based on https://github.com/microsoft/TypeScript/pull/53961.


shinichy:

jakebailey:
@typescript-bot test this
@typescript-bot test top100
@typescript-bot run dt
@typescript-bot pack this

src/compiler/checker.ts
```diff
@@ -36680,7 +36680,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             if (isReadonlySymbol(symbol)) {
                 error(expr, Diagnostics.The_operand_of_a_delete_operator_cannot_be_a_read_only_property);
             }
-            checkDeleteExpressionMustBeOptional(expr, symbol);
+            else {
+                checkDeleteExpressionMustBeOptional(expr, symbol);
```
jakebailey, line 36684: I'm somewhat surprised by this; if we ever add code here that calls `getTypeOfSymbol(symbol)` again, then the bug will reappear, so I'm somewhat wondering if this is a band-aid...

shinichy: Hmm, what do you think of changing `checkDeleteExpressionMustBeOptional` to a function that only checks the error condition similar to `isReadonlySymbol`? Then the code is more consistent and can prevent the bug from reappearing.

```
        if (symbol) {
            if (isReadonlySymbol(symbol)) {
                error(expr, Diagnostics.The_operand_of_a_delete_operator_cannot_be_a_read_only_property);
            }
            else if (isNonOptionalSymbol(symbol)) {
                error(expr, Diagnostics.The_operand_of_a_delete_operator_must_be_optional);
            }
        }
        return booleanType;
    }

    function isNonOptionalSymbol(symbol: Symbol): boolean {
        const type = getTypeOfSymbol(symbol);
        if (
            strictNullChecks &&
            !(type.flags & (TypeFlags.AnyOrUnknown | TypeFlags.Never)) &&
            !(exactOptionalPropertyTypes ? symbol.flags & SymbolFlags.Optional : getTypeFacts(type) & TypeFacts.IsUndefined)
        ) {
            return true;
        }

        return false;
    }
```

jakebailey: That's just the same code refactored to look slightly differently; the thing I'm talking about is the weird part of the bug where we don't see this double error when running `tsc` but do in tsserver. I would expect that a fix for this would include some sort of analysis as to why the two situation differ.

shinichy: I see. I misunderstood. We can see this double error in `tsc`. I guess the reason why some people don't see it is missing configuration in tsc.

```
❯ cat double_error.ts
class Foo {}
delete Foo.name;

❯ npx tsc --strict --lib es2015 double_error.ts
double_error.ts:2:8 - error TS2704: The operand of a 'delete' operator cannot be a read-only property.

2 delete Foo.name;
         ~~~~~~~~

double_error.ts:2:8 - error TS2790: The operand of a 'delete' operator must be optional.

2 delete Foo.name;
         ~~~~~~~~


Found 2 errors in the same file, starting at: double_error.ts:2
```

> The operand of a 'delete' operator cannot be a read-only property.

To see this error, we need the type declaration for Function interface that has `name` property. It's defined in https://github.com/Microsoft/TypeScript/blob/main/src/lib/es2015.core.d.ts#L73.

> The operand of a 'delete' operator must be optional.

To see this error, `strictNullChecks` needs to be turned on.

jakebailey: Ah, so it is reproducible in tsc; that changes things!

jakebailey: In that case, this definitely couldn't be reproduced via a compiler test?

shinichy: Yes, I changed the unit test to a compiler test. Would you please take a look?

jakebailey: I've confirmed that the compiler test does show the issue; very surprised as I swear this didn't work for me originally. Oh well, thanks!

jakebailey (approved):

Issue #44939, opened 2021-07-08
Dynamically named tuples


cursorsdottsx:
# Suggestion

TypeScript's type system is extremely expansive and powerful, but one thing that I cannot seem to do is give tuples dynamic names. Eg:

```ts
type Test00 = [["name"]: number];
```

Or even:

```ts
type Test01 = [`name`: number];
```

Just like objects, we should be able to use bracket syntax to indicate that the expression inside should be used as the name.

We can then proceed to use other types to "calculate" the names.

## 🔍 Search Terms

- tuple names
- dynamic tuple names
- changing tuple names
- editing tuple names
- calculated tuple names

## ✅ Viability Checklist

My suggestion meets these guidelines:

* [x] This wouldn't be a breaking change in existing TypeScript/JavaScript code
* [x] This wouldn't change the runtime behavior of existing JavaScript code
* [x] This could be implemented without emitting different JS based on the types of the expressions
* [x] This isn't a runtime feature (e.g. library functionality, non-ECMAScript syntax with JavaScript output, new syntax sugar for JS, etc.)
* [x] This feature would agree with the rest of [TypeScript's Design Goals](https://github.com/Microsoft/TypeScript/wiki/TypeScript-Design-Goals).

*I believe this one meets the last guideline because of goals 2 and 4.*

## ⭐ Suggestion

Dynamically/calculated/inserted tuple names instead of static ones.

## 📃 Motivating Example

Consider the following example (ripped from a project of mine):

```ts
type Components = -1 | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12 | 13 | 14 | 15 | 16 | 17 | 18 | 19 | 20 | 21 | 22 | 23 | 24 | 25 | 26;

type Decrement<X extends number> = [-1, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25][X];

type CreateArray<T, L extends number> = L extends -1 ? [] : [T, ...CreateArray<T, Decrement<L>>];

type CanonicalAxisNames = ["", "x", "y", "z", "w", "v", "u", "t", "s", "r", "q", "p", "o", "n", "m", "l", "k", "j", "i", "h", "g", "f", "e", "d", "c", "b", "a"];

[...]

PR #55452, opened 2023-08-21
Instantiable tuple labels

Andarist:
closes https://github.com/microsoft/TypeScript/issues/44939

A somewhat big drawback of this feature is that an instantiated label is only usable if it instantiates to a string literal type. This means that instantiations that end up being template literal types and even unions of strings are just discarded. I'm open to discussing other possible designs.

I think it's a reasonable tradeoff since this feature is primarily an editor experience improvement. Labels don't have any effect on type checking itself.

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -7004,11 +7004,12 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                             for (let i = 0; i < tupleConstituentNodes.length; i++) {
                                 const flags = (type.target as TupleType).elementFlags[i];
                                 const labeledElementDeclaration = labeledElementDeclarations?.[i];
+                                const label = labeledElementDeclaration && getTupleElementLabel(type as TupleTypeReference, labeledElementDeclaration);
 
-                                if (labeledElementDeclaration) {
+                                if (label) {
                                     tupleConstituentNodes[i] = factory.createNamedTupleMember(
                                         flags & ElementFlags.Variable ? factory.createToken(SyntaxKind.DotDotDotToken) : undefined,
-                                        factory.createIdentifier(unescapeLeadingUnderscores(getTupleElementLabel(labeledElementDeclaration))),
+                                        factory.createIdentifier(unescapeLeadingUnderscores(label)),
                                         flags & ElementFlags.Optional ? factory.createToken(SyntaxKind.QuestionToken) : undefined,
                                         flags & ElementFlags.Rest ? factory.createArrayTypeNode(tupleConstituentNodes[i]) :
                                             tupleConstituentNodes[i],
@@ -12954,7 +12955,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         function getUniqAssociatedNamesFromTupleType(type: TupleTypeReference, restName: __String) {
             const associatedNamesMap = new Map<__String, number>();
             return map(type.target.labeledElementDeclarations, (labeledElement, i) => {
-                const name = getTupleElementLabel(labeledElement, i, restName);
+                const name = getTupleElementLabel(type, labeledElement, i, restName);
                 const prevCounter = associatedNamesMap.get(name);
                 if (prevCounter === undefined) {
                     associatedNamesMap.set(name, 1);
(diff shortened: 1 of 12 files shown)


src/compiler/checker.ts
```diff
@@ -35345,13 +35357,13 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
         if (isTupleType(restType)) {
-            const associatedNames = ((restType as TypeReference).target as TupleType).labeledElementDeclarations;
+            const associatedNames = restType.target.labeledElementDeclarations;
             const index = pos - paramCount;
             const associatedName = associatedNames?.[index];
             const isRestTupleElement = !!associatedName?.dotDotDotToken;
 
-            if (associatedName) {
-                Debug.assert(isIdentifier(associatedName.name));
+            // TODO: figure out the non-identifier case here
```
Andarist, line 35365: @MariaSolOs what kind of inlay hint interactivity is expected out of a parameter name?

jakebailey:
@typescript-bot pack this

src/compiler/checker.ts
MariaSolOs: The expected behaviour is to be able to click on an argument of a function call and have that be a "go to definition" to the respective parameter in the function signature.

DanielRosenwasser:
@typescript-bot pack this

Andarist:
@DanielRosenwasser this is still a draft that I put out since I wanted to share this with some people who asked for this feature. 

Since then I tested this more and it's not working correctly **yet** for some recursive cases etc. I have some work in progress locally that fixes some of those issues. While investigating them I learned a bunch of new things about how type references, mappers, etc are implemented under the hood 😉 

Those learnings make it *harder* to implement this feature - or to rephrase: I've hoped for a simpler implementation but at the end of the day I can't avoid instantiating those labels *sooner*. This raises some data structure/internal API questions. Since those things are also consumed programmatically I'd like to ensure that an implementation design won't be too disruptive for those consumers. I have some ideas/some open questions related to that. If the team would be open to discussing those concerns before I sit down to implement an enhanced proof of concept - I would highly appreciate it. It would be also quite nice if the team could comment on general level of your interest in this feature :)

DanielRosenwasser:
Yeah, I think that's why I packed it - I want to gauge appetite/interest. Of course, even if we're interested, a big part of this will hinge on the complexity/maintainability/limitations of the implementation.

DanielRosenwasser:
Actually, I'm having a hard time imagining a ***reasonable*** implementation of `InsertTupleNames`. I think you have some very basic first-order types that are being used as labels, but I still don't quite get the motivation - what people want to write, how feasible/practical it is, and whether it's worth it.

Andarist:
I can't truly comment on `InsertTupleNames` right now - would have to read that issue more carefully. The motivating example that I'd like to make work is close to something like this ([TS playground](https://www.staging-typescript.org/play?checkJs=true&ts=5.3.0-pr-55452-4#code/C4TwDgpgBAKgjANigXigbQGYHssC4oDOwATgJYB2A5gDRQBGAhsfnaZRcALoBQokscAOwp0AOnHwEtRgC985AK4BbOhGI8A9Bqg6AegH5uvcNBhYAcgyUQAJjAVgANhAA83HTACCrAApMrEMBqBFAQAB5B5DYhxBAMNljkjiBQAN5Q5AH4RGRUANxQfBD4CuQA1uRYAO7kUAC+aJzU7lAAShAECo7AIeGR0VAM5CCNIo3NAHwiXr7+1kHEvREQUTFxCUkpaC0UGGpQABLrocuraRlZhCQUlAVFJeWVNfXNOuKiu-swDKSOJ-1reKJZLnTLWbLXfKFEwPCrVWoNJrcHg6fSwCwBOwOZxuHR42A-RyvfFicTtTrdAi0AAGABJUkd4mgAORgiDMzh1an4Rk2FlFDkovETFr4cldHpGIptDoS6YY6xYpyuFpaKAGFrbfHpNn4ZnYLDMu4wq65SgvFo6HWXZmMYhG6GQFhsDgW7UXcFQW0MGQO+4ZZSqYhuqA8EVAA)):

```ts
type ToNamedTuple<
  TAbiParameters extends readonly { name: string; type: unknown }[],
  Results extends any[] = [],
> = TAbiParameters extends readonly [
  infer Head extends { name: string; type: unknown },
  ...infer Tail extends readonly { name: string; type: unknown }[],
]
  ? ToNamedTuple<
      Tail,
      [...Results, `${Head['name']}`: Head['type']]
    >
  : Results

type Result = ToNamedTuple<
  [
    { name: 'foo'; type: string },
    { name: 'bar'; type: bigint },
    { name: 'baz'; type: number },
  ]
>
```

This is definitely implementable, just a matter of juggling type parameters and things. 

In the beginning, I thought that `.mapper` would always be there and that it would be able to resolve all type parameters. That would allow me to have those labels pretty lazily evaluated. That's not how it works though. 

Then I tried to store a global mapping between the types and mappers but that didn't pan out because it turned out that types are heavily cached backed on their type arguments. So that made the labels to "leak" between tuple instantiations with the same element types.

There is no good caching unit that would allow me to keep labels/mappers around - other than the type itself. Currently labels of tuples are `Node`s (well, `Identifier`s). Changing that `labeledElementDeclarations`' type to `(Node | Type)[]` would be bizarre and I discarded this option right away.

Types created based on template literal types at labels positions should be *somehow* included in the type's ID. This way new types could be created based on different label instantiations and things would "just work". This has an obvious drawback of creating more tuple types with the same element types. I don't think that's a big issue from the memory PoV but the checker should learn how to compare such tuple tupes effectively - labels are purely decorative and they shouldn't affect the assignability rules between same element tuples.

Alternatively, wrapper types could be introduced with `.target` that would reference the underlying tuple type that would hold the actual element types and the wrapper would only be used to store the instantiated labels. I'm not yet sure how I feel about it. On one level it feels like a complication but it has some potential of making the implementation simpler at the expense of having one extra type of a type to care about. Note that I already found a potential use case for an alternative "tuple type" in the past - currently reversed mapped tuple types have subtle differences from reverse mapped object types since they are instantiated at different times. Introducing a `ReversedTupleType` (or something akin to that) could potentially fix this by making it possible to defer their instantiations so they could behave in the same way as object ones.

rexfordessilfie:
Got some other motivating examples for you (potentially). I ran into a similar issue where I think I could benefit from the changes in this PR and from something such as `InsertTupleNames` - as I understand it. 

Here are the examples:

1. Retyping the parameters of a function. I think I have an immediate use case for this [here](https://github.com/rexfordessilfie/ts-wrappers/pull/21). 

    ```ts
    type Result = RetypeParameters<[a: string, b: number], [number, string]> // [a: number, b: string]
    ```

2. Named tuple to Object Type. Kind of the reverse of your example.
    ```ts
    type Result = ToObject<[a: string, b: number]> // { a: string, b: number }
    ```

3. Using `infer` over named tuples. (Not sure if this will be addressed by your PR, but my thinking is this would allow for the above two to be easy to implement?)
    ```ts
    type InferNames<NamedTuple extends any[]> = 
        NamedTuple extends [infer Name: any, ...infer Rest] ? 
        [Name, ...InferNames<Rest>] : never 
    ```

What do you think?

> Apologies if this is not the right avenue to post any of these. Happy to move over to a new issue, or to the original issue your PR is solving. Thanks for driving this forward!

sentialx:
Looking forward to this feature

PR #55472, opened 2023-08-22
Avoid trying to emit anonymous classish/expando functions as assignments

### LINKED ISSUE #55172: tsc with allowJs reports TS9005 error for anonymous constructor functions
# Bug Report

Declaring a constructor function like so:
```ts
const SomeConstructor = function() {
	this.x = 1;
};
```

leads to typescript reporting error "TS9005":  Declaration emit for this file requires using private name 'SomeConstructor'. An explicit type annotation may unblock declaration emit.
However the practically equivalent using a named function:
```ts
function SomeConstructor () {
	this.x = 1;
};
```
works fine.

It seems to me that the TS compiler should not care which of these notations is used. But even if the difference is intentional, the error message generated seems incorrect or unhelpful and misleading at least. Also the hint to add type annotations seems inappropriate; no amount of explicit type annotations will make this error go away.

Similar errors get generated when using the following constructs:
```
module.exports = {
	SomeConstructor: function () {
		this.x = 1;
	}
};
```
or 
```
module.exports = function () {
	this.x = 1;
};
```
My tsconfig.json:
```
{
	"include": ["*.js"],
	"compilerOptions": {
		"allowJs": true,
		"declaration": true,
		"outDir": "dist",
		"skipLibCheck": true
	}
}
```
### 🔎 Search Terms
typescript TS9005
Declaration emit for this file requires using private name
typescript allowJs error 9005 const function class declaration
typescript allowJs error 9005 anonymous function class declaration

This also happens in the Nightly version in the playground:
https://www.typescriptlang.org/play?strict=false&noImplicitAny=false&strictFunctionTypes=false&strictPropertyInitialization=false&strictBindCallApply=false&noImplicitThis=false&noImplicitReturns=false&alwaysStrict=false&ts=5.2.0-dev.20230727&filetype=js&strictNullChecks=false#code/MYewdgzgLgBAyiAtgUwMLmgJwK7CiTGAXhgDNsw8BLcGACgEoYBvAKAEgoALKiAOgAexGAEYA3KwC+EoA

It does not happen with versions 3.3.3-3.6.3: 
https://www.typescriptlang.org/play?strict=false&noImplicitAny=false&strictFunctionTypes=false&strictPropertyInitialization=false&strictBindCallApply=false&noImplicitThis=false&noImplicitReturns=false&alwaysStrict=false&ts=3.6.3&filetype=js&strictNullChecks=false#code/MYewdgzgLgBAyiAtgUwMLmgJwK7CiTGAXhgDNsw8BLcGACgEoYBvAKAEgoALKiAOgAexGAEYA3KwC+EoA

However this versions also dont generate any typing information (.D.TS).

It seems to start happening with version 3.7.5:
https://www.typescriptlang.org/play?strict=false&noImplicitAny=false&strictFunctionTypes=false&strictPropertyInitialization=false&strictBindCallApply=false&noImplicitThis=false&noImplicitReturns=false&alwaysStrict=false&ts=3.7.5&filetype=js&strictNullChecks=false#code/MYewdgzgLgBAyiAtgUwMLmgJwK7CiTGAXhgDNsw8BLcGACgEoYBvAKAEgoALKiAOgAexGAEYA3KwC+EoA

- This changed between versions 3.6.3 and 3.7.5

### ⏯ Playground Link

[Playground link with relevant code](https://www.typescriptlang.org/play?strict=false&noImplicitAny=false&strictFunctionTypes=false&strictPropertyInitialization=false&strictBindCallApply=false&noImplicitThis=false&noImplicitReturns=false&alwaysStrict=false&ts=5.2.0-dev.20230727&filetype=js&strictNullChecks=false#code/MYewdgzgLgBAyiAtgUwMLmgJwK7CiTGAXhgDNsw8BLcGACgEoYBvAKAEgoALKiAOgAexGAEYA3KwC+EoA)

### 💻 Code

```ts
const SomeConstructor = function() {
	this.x = 1;
};
```

### 🙁 Actual behavior

tsc reports error "TS9005":  Declaration emit for this file requires using private name 'SomeConstructor'. An explicit type annotation may unblock declaration emit.

This seems wrong because the equivalent code
```ts
function SomeConstructor () {
	this.x = 1;
};
```
works fine and it seems that there shouldn't be any relevant differences between these notations for the TS compiler.
### 🙂 Expected behavior

tsc should be happy

weswigham:
Fixes #55172


src/compiler/checker.ts
```diff
@@ -8831,7 +8833,18 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
                         const localName = getInternalSymbolName(symbol, symbolName);
-                        if (!(symbol.flags & SymbolFlags.Function) && isTypeRepresentableAsFunctionNamespaceMerge(type, symbol)) {
+                        if (type.symbol && type.symbol !== symbol && type.symbol.flags & SymbolFlags.Function && some(type.symbol.declarations, isFunctionExpressionOrArrowFunction) && (type.symbol.members?.size || type.symbol.exports?.size)) {
+                            // assignment of a anonymous expando/class-like function, the func/ns/merge branch below won't trigger,
+                            // and the assignment form has to reference the unreachable anonymous type so will error.
+                            // Instead, serialzie the type's symbol, but with the current symbol's name, rather than the anonymous one.
+                            if (!context.remappedSymbolRefernces) {
+                                context.remappedSymbolRefernces = new Map();
+                            }
+                            context.remappedSymbolRefernces.set(getSymbolId(type.symbol), symbol); // save name remapping as local name for target symbol
```
weswigham, line 8843: This is to handle the not-anonymous-but-still-an-expression case where you could use that local name to refer to the type of the expando-thing as a whole. Because it'd be odd if just adding a name to the function expression made this transform stop working.

src/compiler/checker.ts
```diff
@@ -8525,11 +8525,13 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 ...oldcontext,
                 usedSymbolNames: new Set(oldcontext.usedSymbolNames),
                 remappedSymbolNames: new Map(),
+                remappedSymbolRefernces: new Map(oldcontext.remappedSymbolRefernces?.entries()),
```
sandersn, line 8528:
```suggestion
                remappedSymbolReferences: new Map(oldcontext.remappedSymbolReferences?.entries()),

```

src/compiler/checker.ts
```diff
@@ -8525,11 +8525,13 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 ...oldcontext,
                 usedSymbolNames: new Set(oldcontext.usedSymbolNames),
                 remappedSymbolNames: new Map(),
+                remappedSymbolRefernces: new Map(oldcontext.remappedSymbolRefernces?.entries()),
                 tracker: undefined!,
             };
             const tracker: SymbolTracker = {
                 ...oldcontext.tracker.inner,
                 trackSymbol: (sym, decl, meaning) => {
+                    if (context.remappedSymbolNames?.has(getSymbolId(sym))) return false; // If the context has a remapped name for the symbol, it *should* mean its been made visible
```
sandersn, line 8534:
```suggestion
                    if (context.remappedSymbolNames?.has(getSymbolId(sym))) return false; // If the context has a remapped name for the symbol, it *should* mean it's been made visible
```

src/compiler/checker.ts
```diff
@@ -8831,7 +8833,18 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     else {
                         const type = getTypeOfSymbol(symbol);
                         const localName = getInternalSymbolName(symbol, symbolName);
-                        if (!(symbol.flags & SymbolFlags.Function) && isTypeRepresentableAsFunctionNamespaceMerge(type, symbol)) {
+                        if (type.symbol && type.symbol !== symbol && type.symbol.flags & SymbolFlags.Function && some(type.symbol.declarations, isFunctionExpressionOrArrowFunction) && (type.symbol.members?.size || type.symbol.exports?.size)) {
+                            // assignment of a anonymous expando/class-like function, the func/ns/merge branch below won't trigger,
+                            // and the assignment form has to reference the unreachable anonymous type so will error.
+                            // Instead, serialzie the type's symbol, but with the current symbol's name, rather than the anonymous one.
```
sandersn, line 8839:
```suggestion
                            // Instead, serialize the type's symbol, but with the current symbol's name, rather than the anonymous one.

```

src/compiler/checker.ts
sandersn: Is this is for cases like

```js
const f = function g() { }
f.p = 1
```

Does remappedSymbolReferences record `g`? When would `g` show up in the declaration emit?

Edit: Looking at the test cases, it's actually `g -> f` so that when a constructor function `g` returns `g`, it serialises as `f`.

sandersn (approved):

PR #55479, opened 2023-08-23
fix(chsTranslation): do not translate the word "rest" into Chinese

shispt:
The word "rest" should not be translated in the item `A_rest_element_cannot_have_a_property_name_2566`, otherwise it will be really confusing.

I have checked the other translations in the same file and the translation in [cht translation](https://github.com/microsoft/TypeScript/blob/main/src/loc/lcl/cht/diagnosticMessages/diagnosticMessages.generated.json.lcl#L688), they both did not translate the word "rest".

### DIFF
--- a/src/loc/lcl/chs/diagnosticMessages/diagnosticMessages.generated.json.lcl
+++ b/src/loc/lcl/chs/diagnosticMessages/diagnosticMessages.generated.json.lcl
@@ -685,7 +685,7 @@
         <Str Cat="Text">
           <Val><![CDATA[A rest element cannot have a property name.]]></Val>
           <Tgt Cat="Text" Stat="Loc" Orig="New">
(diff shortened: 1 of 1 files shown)


sandersn:
Unfortunately, Microsoft fired so many of its translators this year that we don't have budget for fixing anything but critical mistranslations. 💀 It might be possible to switch over to community translations, although I'd have to learn whether that would be possible.

In any case, I think `rest` **should** be translated here: it's a jargon word in English, like `using` or `let`, with only a weak relation to its usual meaning. It may feel more natural to keep it in English for a programmer who already speaks English, but it will cause a lot of work for a programmer who doesn't speak English. They'll have to memorise it as a new, independent term.

That said, it's quite possible that there is a better translation to Chinese. I don't speak Chinese so I don't have any ideas.

PR #55482, opened 2023-08-23
perf: replace String and Array indexOf method calls with includes method call

Havunen:
replace String and Array indexOf method calls with includes method call when return value was used only to check that given item exists


RyanCavanaugh:

jakebailey:
All research I've done on this has shown that `includes` is actually slightly slower (fraction of a percent) than `indexOf` at best in modern v8. But, it is obviously clearer. I doubt it will matter for this codebase.

RyanCavanaugh (approved):
I triple-checked this for correctness and it looks right. This seems cleaner

src/compiler/core.ts
```diff
@@ -2486,7 +2486,7 @@ export function tryRemoveSuffix(str: string, suffix: string): string | undefined
 
 /** @internal */
 export function stringContains(str: string, substring: string): boolean {
-    return str.indexOf(substring) !== -1;
+    return str.includes(substring);
```
jakebailey, line 2489: Eventually it'd be nice to inline this; it has like 40 other calls via this function.

jakebailey (approved):
Given no real perf change, seems good to me, though I would not complain if you used this PR to replace `stringContains` with `.includes` everywhere.

Havunen:
> Given no real perf change, seems good to me, though I would not complain if you used this PR to replace stringContains with .includes everywhere.

Done

src/compiler/checker.ts
```diff
@@ -30962,7 +30961,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     }
 
     function isHyphenatedJsxName(name: string | __String) {
-        return stringContains(name as string, "-");
+        return (name as string).includes("-");
```
jakebailey, line 30964:
```suggestion
        return name.includes("-");
```

The type assertion isn't needed anymore; it was likely only here as the function only accepted `string`, but the method will be present on both `string` and `__String`.

jakebailey:
Doing a regex search of `\.indexOf\(.*(>|<|<=|>=|===|!==)` shows that there are a few more of these that could be fixed up too. (Some false positives, and some that are `Array`, but mostly fixable).

Not sure what your final goal is but in any case this seems fine to me now and we can replace more later, of course.

jakebailey:
`\.indexOf\(.*(< 0|=== -1|!== -1|>= 0)` is probably a better regex.

src/compiler/checker.ts
Havunen: I tried it, but it does not compile:

image

Havunen:
> \.indexOf\(.*(< 0|=== -1|!== -1|>= 0) is probably a better regex.

Ok, I went through it and there are some edge cases where indexOf second parameter is used and that cannot be easily replaced with includes check. 

I also didn't want to touch the test files, maybe its better to go through them separately. 

Anyways, if this change is planned to be merged it would be nice to do it before conflicts emerge

jakebailey:
> Ok, I went through it and there are some edge cases where indexOf second parameter is used and that cannot be easily replaced with includes check.

Hm, `includes` has the same second parameter, so I would have expect it to work. But that's fine; I would have taken this regardless.

src/compiler/checker.ts
jakebailey: Agh, yes, I had some editor delay. Shame.

Havunen:
oh right, I didn't know that :)

jakebailey (approved):
I manually reviewed all of the changes again, and also diffed this against eslint with eslint-plugin-unicorn's `prefer-includes` rule, and they all line up. Thanks!

Issue #50699, opened 2022-09-09
Do not check in the `transpileModule` API


rricard:
# Suggestion

## 🔍 Search Terms

-   `transpileModule`
-   `skip typechecking`
-   `no typechecking`

Semi-related issues found:

-   Skip typechecking; only emit (support --transpileOnly in tsc, re-open of #4176) #29651
-   [FEATURE REQUEST] Please supply a minified version with only transpilation functionality #9245
-   Allow transformers to run without type-checking #41986

## ✅ Viability Checklist

My suggestion meets these guidelines:

-   [x] This wouldn't be a breaking change in existing TypeScript/JavaScript code
-   [x] This wouldn't change the runtime behavior of existing JavaScript code
-   [x] This could be implemented without emitting different JS based on the types of the expressions
-   [x] This isn't a runtime feature (e.g. library functionality, non-ECMAScript syntax with JavaScript output, new syntax sugar for JS, etc.)
-   [x] This feature would agree with the rest of [TypeScript's Design Goals](https://github.com/Microsoft/TypeScript/wiki/TypeScript-Design-Goals).

## ⭐ Suggestion

Modify the `transpileModule` API to mostly do the following:

-   Parses a source string
-   Runs the corresponding transforms to get to the target ES version
-   Emits a resulting string with a sourceMap string

The API itself would be unchanged but would be guaranteed to not hit the type-checker if a certain set of compatible compiler options is set. (deoptimizing options TBD)

Achieving this without modifying TypeScript is currently not possible with the current TypeScript API: there is currently no way to just apply the transforms without running the checker as a `Program` has to be created in the process.

Alternatively, if changing `transpileModule` is not possible we would be interested in exposing more minimal API primitives so we could build up this API ourselves:

-   A public `Printer` API that can issue sourcemaps
-   A way to run transforms on TS ASTs without the need to build up a `Program` (and run the checker...)

## 📃 Motivating Example

In our build system, for certain workflows, we introduced a new build mode that runs `transpileModule` instead of compiling and type-checking the full project. `transpileModule` is run on a per-file basis as we are reading files into our build stream, meaning there are no inter-file ordering dependencies.

<details>
<summary>Approximative `transpileModule` API usage</summary>

```ts
const result = ts.transpileModule(tsFileStr, {
    compilerOptions, // <- with an ESNext target
    reportDiagnostics: true,
    moduleName: tsFilePath,
    fileName: tsFilePath,
});
if (result.diagnostics) {
[...]

PR #55484, opened 2023-08-23
perf: reduce GC pressure by hoisting script target features object

Havunen:
This PR reduces GC pressure by hoisting ScriptTargetFeatures instead of creating that object over and over again.

image

### DIFF
--- a/src/compiler/utilities.ts
+++ b/src/compiler/utilities.ts
@@ -1282,7 +1282,7 @@ export function getInternalEmitFlags(node: Node): InternalEmitFlags {
 export type ScriptTargetFeatures = ReadonlyMap<string, ReadonlyMap<string, string[]>>;
 
 /** @internal */
-export const getScriptTargetFeatures = /* @__PURE__ */ memoize(() => new Map(Object.entries({
+export const getScriptTargetFeatures: () => ScriptTargetFeatures = /* @__PURE__ */ memoize(() => new Map(Object.entries({
     Array: new Map(Object.entries({
         es2015: [
             "find",
@@ -1675,7 +1675,7 @@ export const getScriptTargetFeatures = /* @__PURE__ */ memoize(() => new Map(Obj
             "cause",
         ],
     })),
-})) as ScriptTargetFeatures);
+})));
 
 /** @internal */
 export const enum GetLiteralTextFlags {
--- a/src/compiler/utilities.ts
+++ b/src/compiler/utilities.ts
@@ -386,6 +386,7 @@ import {
     mapDefined,
     MapLike,
     MemberName,
+    memoize,
     MetaProperty,
     MethodDeclaration,
     MethodSignature,
@@ -1281,7 +1282,7 @@ export function getInternalEmitFlags(node: Node): InternalEmitFlags {
 export type ScriptTargetFeatures = ReadonlyMap<string, ReadonlyMap<string, string[]>>;
 
 /** @internal */
-export const AllScriptTargetFeatures = new Map(Object.entries({
+export const getScriptTargetFeatures = /* @__PURE__ */ memoize(() => new Map(Object.entries({
     Array: new Map(Object.entries({
         es2015: [
             "find",
@@ -1674,7 +1675,7 @@ export const AllScriptTargetFeatures = new Map(Object.entries({
             "cause",
         ],
     })),
-})) as ScriptTargetFeatures;
+})) as ScriptTargetFeatures);
 
 /** @internal */
 export const enum GetLiteralTextFlags {
--- a/src/compiler/utilities.ts
+++ b/src/compiler/utilities.ts
@@ -1281,402 +1281,400 @@ export function getInternalEmitFlags(node: Node): InternalEmitFlags {
 export type ScriptTargetFeatures = ReadonlyMap<string, ReadonlyMap<string, string[]>>;
 
 /** @internal */
-export function getScriptTargetFeatures(): ScriptTargetFeatures {
-    return new Map(Object.entries({
-        Array: new Map(Object.entries({
-            es2015: [
-                "find",
-                "findIndex",
-                "fill",
-                "copyWithin",
-                "entries",
-                "keys",
-                "values",
-            ],
-            es2016: [
-                "includes",
-            ],
-            es2019: [
-                "flat",
-                "flatMap",
-            ],
-            es2022: [
-                "at",
-            ],
-            es2023: [
-                "findLastIndex",
-                "findLast",
-            ],
-        })),
-        Iterator: new Map(Object.entries({
-            es2015: emptyArray,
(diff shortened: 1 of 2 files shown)


src/compiler/utilities.ts
```diff
@@ -1281,402 +1281,400 @@ export function getInternalEmitFlags(node: Node): InternalEmitFlags {
...
-            ],
-        })),
-        Error: new Map(Object.entries({
-            es2022: [
-                "cause",
-            ],
-        })),
-    }));
-}
+export const AllScriptTargetFeatures = new Map(Object.entries({
```
Havunen, line 1284: There is a risk that some code mutates the values in this Map object, but its typed as ReadOnlyMap. I also tested that Object.freeze unfortunately does not guard Maps

src/compiler/utilities.ts
```diff
@@ -1281,402 +1281,400 @@ export function getInternalEmitFlags(node: Node): InternalEmitFlags {
 export type ScriptTargetFeatures = ReadonlyMap<string, ReadonlyMap<string, string[]>>;
 
 /** @internal */
-export function getScriptTargetFeatures(): ScriptTargetFeatures {
```
jakebailey, line 1284: Instead of turning this into a global that's always calculated, you probably should instead write this as:

```ts
export const getScriptTargetFeatures = memoize(() => new Map({ ... }))
```

Then it will still be lazy, but only calculated once.

jakebailey (changes requested):
See above; also, general note, but we typically only uppercase things that are types.

DanielRosenwasser:

jakebailey:
What are you testing where you're seeing this be costly? Is this likely to be shown in `tsc` or `tsserver` benchmarks?

Havunen:
> What are you testing where you're seeing this be costly? Is this likely to be shown in tsc or tsserver benchmarks?

I'm trying to make our application build faster. Its webpack5 vue2 typescript app

src/compiler/utilities.ts
jakebailey: Actually, you'll also want to write:

```ts
export const getScriptTargetFeatures = /* @__PURE__ */ memoize(() => new Map({ ... }))
```

See similar calls which I made lazy e.g. `createPrinterWithDefaults`. This is mainly for tree shaking reasons as the function declaration doesn't have side effects but the memoize call does.

DanielRosenwasser:
For a single compilation it's surprising that we even end up doing this more than once. This is only supposed to occur as part of error recovery for an unresolved type. Do you have a lot of unchecked JavaScript code or something?

src/compiler/utilities.ts
Havunen: good points, I will change it

Havunen: done

Havunen:
> For a single compilation it's surprising that we even end up doing this more than once. This is only supposed to occur as part of error recovery for an unresolved type. Do you have a lot of unchecked JavaScript code or something?

Here is the call stack:

image

src/compiler/utilities.ts
```diff
@@ -1281,402 +1282,400 @@ export function getInternalEmitFlags(node: Node): InternalEmitFlags {
...
+            "findLastIndex",
+            "findLast",
+        ],
+    })),
+    Error: new Map(Object.entries({
+        es2022: [
+            "cause",
+        ],
+    })),
+})) as ScriptTargetFeatures);
```
jakebailey, line 1678: Instead of casting, you probably just want to write `(): ScriptTargetFeatures => new Map()`

jakebailey:
I'm certainly confused as to why this is appearing in your call stack; it's trying to produce a nice message for IArguments not existing which could only happen because your code is not typechecking. IArguments is in `lib.es5.d.ts` so that's extra strange and implies there's something not quite right in your build.

src/compiler/utilities.ts
Havunen: ok, I removed the casting

jakebailey:
I'm sure you've already seen, but you'll want to `npm run format` to get that check passing. That'll probably make the diff better too.

jakebailey (approved):
This PR seems fine now, but I do very much question why this is helping you at all (per above).

Havunen:
It seems to have something to do with `fork-ts-checker-webpack-plugin` (v7.3.0) when I remove that plugin from the webpack pipeline this method does not get called as often anymore.

jakebailey:
Is it checking with the wrong tsconfig/compilerOptions or something? Missing `IArguments` is extremely suspect.

jakebailey:
You know, I can see how this may be a thing; this is coming out of `transpileModule`, so in that fork plugin, it's the side of the plugin that's doing emit, not checking. But if it's passing in the "wrong" options (or, if `transpileModule` doesn't give the "right" options under the assumption that it doesn't matter), but some aspect of the checker is still running, then we may get a bunch of errors that will be thrown away.

jakebailey:
Yeah, `transpileModule` implies `noLib`, so it seems to me like these errors are inevitable?

Of course, this makes me wonder if we could do a lot better about not requesting / doing work for diagnostics during `transpileModule`, which is roughly #50699 (specifically note https://github.com/microsoft/TypeScript/issues/50699#issuecomment-1244128250).

@andrewbranch for interest 😄 

In any case, this seems like a totally good optimization regardless; I can see this being checked repeatedly in the editor, for example.

RyanCavanaugh:
We could probably get a net perf win in `transpileModule` by providing a minimal shim lib just to reduce the number of errors reported in nontrivial files

src/compiler/utilities.ts
```diff
@@ -1281,8 +1282,8 @@ export function getInternalEmitFlags(node: Node): InternalEmitFlags {
 export type ScriptTargetFeatures = ReadonlyMap<string, ReadonlyMap<string, string[]>>;
 
 /** @internal */
-export function getScriptTargetFeatures(): ScriptTargetFeatures {
-    return new Map(Object.entries({
+export const getScriptTargetFeatures: () => ScriptTargetFeatures = /* @__PURE__ */ memoize(() =>
```
DanielRosenwasser, line 1285: I don't think we use this annotation anywhere else - any reason you felt we needed it?

DanielRosenwasser (approved):

src/compiler/utilities.ts
jakebailey: I had suggested sticking it on the closure inside of the memoize call but they put it here instead (which is roughly equivalent, I guess). Mainly it's that this type is an aliased ReadonlyMap. Not sure why it's named, per se

jakebailey: I just tacked a small change onto the PR to move the annotation to the other place; wanting to merge this!

PR #55485, opened 2023-08-23
Fix --clean not cleaning declaration files 

### LINKED ISSUE #54389: tsc --build tsx - emits d.ts, but doesn't --clean it, for included subfolders
# Bug Report

### 🔎 Search Terms

tsx build artefact clean
tsc --build --clean bug
type declaration d.ts
Cannot write file ... because it would overwrite input file.

tsc --version
Version 5.0.4

node --version
v18.15.0

### 💻 Code

Steps to reproduce:

```shell
mkdir tsx-clean-bug
cd tsx-clean-bug
npm init
npm install -d typescript
mkdir -p src/file
touch src/file/file.tsx
touch tsconfig.json
```

file.tsx
```ts
export const whatever = 'ok';
```

tsconfig.json
```json
{
  "compilerOptions": {
    "declaration": true
  },
  "include": ["src/**/*.tsx", "src/**/*.ts"]
}
```

### 🙁 Actual behavior

Then compile it:
`npx tsc`

```shell
> tree src
src
└── file
    ├── file.d.ts
    ├── file.js
    └── file.tsx
```

Looks fine, try to clean build artefacts:
`npx tsc --build --clean`

```shell
> tree src
src
└── file
    ├── file.d.ts
    └── file.tsx
```

`file.d.ts` - stays in place :(

And if we try to compile sources again:

```shell
> npx tsc
error TS5055: Cannot write file '/Users/g.krikun/dev/tsx-clean-bug/src/file/file.d.ts' because it would overwrite input file.

Found 1 error.
```

### 🙂 Expected behavior

Clean build artefacts:
`npx tsc --build --clean`

Should remove produced `.d.ts` declaration file:

```shell
> tree src
src
└── file
    └── file.tsx
```

Should compiles without error

piotrzarycki:
Fixes #54389

### DIFF
--- a/src/compiler/tsbuildPublic.ts
+++ b/src/compiler/tsbuildPublic.ts
@@ -44,6 +44,8 @@ import {
     emptyArray,
     ExitStatus,
     ExtendedConfigCacheEntry,
+    Extension,
+    fileExtensionIsOneOf,
     FileWatcher,
     FileWatcherCallback,
     findIndex,
@@ -2244,7 +2246,7 @@ function cleanWorker<T extends BuilderProgram>(state: SolutionBuilderState<T>, p
         const inputFileNames = new Set(parsed.fileNames.map(f => toPath(state, f)));
         for (const output of outputs) {
             // If output name is same as input file name, do not delete and ignore the error
-            if (inputFileNames.has(toPath(state, output))) continue;
+            if (inputFileNames.has(toPath(state, output)) && !fileExtensionIsOneOf(output, [Extension.Dts, Extension.Dmts, Extension.Dcts])) continue;
             if (host.fileExists(output)) {
                 if (filesToDelete) {
                     filesToDelete.push(output);


src/compiler/tsbuildPublic.ts
```diff
@@ -2244,7 +2246,7 @@ function cleanWorker<T extends BuilderProgram>(state: SolutionBuilderState<T>, p
         const inputFileNames = new Set(parsed.fileNames.map(f => toPath(state, f)));
         for (const output of outputs) {
             // If output name is same as input file name, do not delete and ignore the error
-            if (inputFileNames.has(toPath(state, output))) continue;
+            if (inputFileNames.has(toPath(state, output)) && !fileExtensionIsOneOf(output, [Extension.Dts, Extension.Dmts, Extension.Dcts])) continue;
```
sheetalkamat, line 2249: This seems like incorrect fix. This would delete d.ts files even if they are matched as input file by incorrect configuration. 

I looked at the issue and it looks like the configuration is incorrect. 

```
> tsc --explainFiles
src/file/file.tsx
  Matched by include pattern 'src/**/*.tsx' in 'tsconfig.json'
src/file/file.d.ts
  Matched by include pattern 'src/**/*.ts' in 'tsconfig.json'
```

Changing config fixes the error:
```json
{
  "compilerOptions": {
    "declaration": true
  },
  "include": ["src/**/*.tsx", "src/**/*.ts"],
  "exclude": ["src/**/*.d.ts"]
}
```

sheetalkamat (changes requested):

PR #55490, opened 2023-08-24
blahhhh

dargaCode:
Fixes #

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
 
 # TypeScript
 
+MICROSOFT IS THE DUMBEST COMPANY
+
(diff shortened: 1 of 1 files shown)


RyanCavanaugh:
Hacktoberfest seems to start earlier every year

dargaCode:
I was expecting to show my students that there would be a bunch of requirements and actions and reviews that had to be completed before the PR could be merged, maybe some Code Owners auto-reviewer-assigning, but the `This PR doesn't have any linked issues.` bot was 10x better. 

Also though sorry for spamming real humans with my garbage here, I was caught up in what I was explaining and didn't think of that part as I should have 😬

PR #55492, opened 2023-08-24
Fixed a regression in serializing setters with non-function declarations in JS files

Andarist:
fixes a regression from https://github.com/microsoft/TypeScript/pull/55393 , cc @jakebailey

### DIFF
--- a/tests/baselines/reference/jsDeclarationsGetterSetter.js
+++ b/tests/baselines/reference/jsDeclarationsGetterSetter.js
@@ -63,6 +63,63 @@ Object.defineProperty(H.prototype, "x", {
 });
 
 
+export class I {}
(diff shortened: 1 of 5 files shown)


tests/baselines/reference/jsDeclarationsGetterSetter.js
```diff
@@ -154,3 +267,18 @@ export class G {
...
+}
+export class K {
+    set x(value: number);
+}
+export class L {
+    set x(value: any);
+}
+export class M {
+    set x(value: any);
+}
```
Andarist, line 284: This one could create a "unified" signature of
```suggestion
export class M {
    set x(value: boolean);
}
```
but that's tangent and not a regression.

jakebailey (approved):

Issue #8559, opened 2016-05-11
ES6 modules cannot use "await" as an identifier


martine:
According to http://www.ecma-international.org/ecma-262/6.0/#sec-future-reserved-words
"await is only treated as a FutureReservedWord when Module is the goal symbol of the syntactic grammar."

This code should be illegal I think:
http://www.typescriptlang.org/play/#src=import%20'foo'%3B%0A%0Alet%20await%20%3D%203%3B

Came up when discussing TS semantics with @ajklein (v8 ES6 module expert).

yortus:
Also note the qualification in [section 10.2](http://www.ecma-international.org/ecma-262/6.0/#sec-types-of-source-code):

> *Module code* is source text that is code that is provided as a *ModuleBody*. It is the code that is directly evaluated when a module is initialized. The module code of a particular module does not include any source text that is parsed as part of a nested *FunctionDeclaration*, *FunctionExpression*, *GeneratorDeclaration*, *GeneratorExpression*, *MethodDefinition*, *ArrowFunction*, *ClassDeclaration*, or *ClassExpression*.

So the OP's playground example is definitely not allowed, but given the above quote I *think* the following is allowed in an ES6 module:

```ts
export function foo() {
    let await = 3;
}
```

bterlson:
@yortus, `await` is a reserved whenever the syntactic goal symbol is _Module_ which is when parsing the entire module script. Thus, `await` is never allowed as an identifier in a module. The _Module_ goal symbol is a distinct concept from "module code". See [5.1.1](https://tc39.github.io/ecma262/#sec-context-free-grammars).

yortus:
@bterlson thanks for the clarification. So we are losing another indentifier to keywordland :disappointed:  Pity it couldn't be treated as a contextual keyword like `async`.

RyanCavanaugh:
(this is AI output)
## Unable to Reproduce ❌

I attempted to reproduce this issue but could not confirm it still exists.

Test case used:
**test.ts:**
```typescript
import 'foo';

let await = 3; // This should result in a syntax error because 'await' is a reserved keyword in ES6 module code.

export function foo() {
    let await = 3; // This is allowed in a function scope, but the identifier 'await' in module scope should not be.
}
```


**Command:** `tsc -p . --module es6 --noImplicitAny`
**TypeScript Version:** Version 5.9.0-dev.20250716
**Exit Code:** 2

<details>
<summary>Compiler Output</summary>

[...]

PR #55503, opened 2023-08-24
Always error on an attempt to use `await` as an identifier in modules

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/8559

### DIFF
--- a/src/compiler/binder.ts
+++ b/src/compiler/binder.ts
@@ -167,7 +167,6 @@ import {
     isIdentifier,
     isIdentifierName,
     isInJSFile,
-    isInTopLevelContext,
     isJSDocConstructSignature,
     isJSDocEnumTag,
     isJSDocTemplateTag,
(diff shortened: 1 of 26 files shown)


src/compiler/binder.ts
```diff
@@ -2457,13 +2456,8 @@ function createBinder(): (file: SourceFile, options: CompilerOptions) => void {
...
                 file.bindDiagnostics.push(createDiagnosticForNode(node, getStrictModeIdentifierMessage(node), declarationNameToString(node)));
             }
-            else if (originalKeywordKind === SyntaxKind.AwaitKeyword) {
-                if (isExternalModule(file) && isInTopLevelContext(node)) {
-                    file.bindDiagnostics.push(createDiagnosticForNode(node, Diagnostics.Identifier_expected_0_is_a_reserved_word_at_the_top_level_of_a_module, declarationNameToString(node)));
-                }
-                else if (node.flags & NodeFlags.AwaitContext) {
-                    file.bindDiagnostics.push(createDiagnosticForNode(node, Diagnostics.Identifier_expected_0_is_a_reserved_word_that_cannot_be_used_here, declarationNameToString(node)));
-                }
+            else if (originalKeywordKind === SyntaxKind.AwaitKeyword && (node.flags & NodeFlags.AwaitContext || isExternalModule(file))) {
```
Andarist, line 2459: Perhaps this would deserve a new diagnostic. The other one (with `node.flags & NodeFlags.AwaitContext`) didn't have a specific diagnostic meant for it so I skipped adding one here as well. 

I think though that it could be nice to add a specific diagnostic here (and maybe add a specific one for that other case as well while at it). If you agree - should this be a complete new diagnostic code or could the existing code be reused? Probably a new one makes more sense.

DanielRosenwasser: I believe technically this is allowed in non-strict CJS, but I'd rather not make weird exceptions.

DanielRosenwasser:
@typescript-bot pack this
@typescript-bot test this
@typescript-bot test top100
@typescript-bot run dt

src/compiler/binder.ts
Andarist: yes, it only technically has to be disallowed in modules 😉 Isn't allowing it in TS files with script targets a weird exception already? 😅

Andarist: It's not allowed in non-strict CJS, it's **always** allowed there. The actual rule is that this is disallowed in modules - not that it's disallowed in strict mode 🤷

Issue #55499, opened 2023-08-24
Crash in compiler when using `extends null`, `super()`, and re-opening class with `interface`


rbuckton:
### 🔎 Search Terms

extends null, crash

### 🕗 Version & Regression Information

- This is a crash
- This is new since 5.2

### ⏯ Playground Link

https://www.typescriptlang.org/play?lib=true&ts=5.2.1-rc#code/JYOwLgpgTgZghgYwgAgEJwM4oN4CgC+uuCANphsgMLIQAekIAJhSAK4knJ7I-IID2IDGCisEYflAAUASi65ei5BlYAHaLIDcC3oUKhIsRCmp0GzNJhz4gA

### 💻 Code

```ts
interface Base {
}

class C extends null {
    constructor() {
        super(); // (1)
    }
}
interface C extends Base {} // (2)
```

### 🙁 Actual behavior

Crash when both `class C extends null` and `interface C extends Base` are used. If you comment out (1), then there is no crash and no error (expected). If you comment out (2), there is no crash and a diagnostic is reported at (1) (expected). The combination of both results in an error.

### 🙂 Expected behavior

The diagnostic message `A constructor cannot contain a 'super' call when its class extends 'null'.` should be printed, just as it does when removing the `interface C extends Base {}` line.

### Additional information about the issue

The crash has the following stack trace:

```
Error: Debug Failure. False expression: Revert #54442 and add a testcase with whatever triggered this
    at resolveCall (...\tsc.js:70670:11)
    at resolveCallExpression (...\tsc.js:71041:18)
    at resolveSignature (...\tsc.js:71486:16)
    at getResolvedSignature (...\tsc.js:71506:18)
    at checkCallExpression (...\tsc.js:71618:23)
    at checkExpressionWorker (...\tsc.js:74800:16)
    at checkExpression (...\tsc.js:74706:32)
    at checkExpressionStatement (...\tsc.js:77336:5)
    at checkSourceElementWorker (...\tsc.js:80280:16)
    at checkSourceElement (...\tsc.js:80140:7)
    ...
```

Related: #54442

Andarist:
I'll investigate this and provide a fix/test case/whatever is needed here.

fatcerberus:
@Andarist I guess Ryan was right in #54442 after all and the diagnostic really *was* reachable 😄

Andarist:
yeah 😅 I made a wrong assumption that this holds true:
[...]

PR #55506, opened 2023-08-24
Consistently return `errorType` for super expressions in classes that extend `null`

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/55499


src/compiler/checker.ts
```diff
@@ -39195,7 +39198,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             // TS 1.0 spec (April 2014): 8.3.2
             // Constructors of classes with no extends clause may not contain super calls, whereas
             // constructors of derived classes must contain at least one super call somewhere in their function body.
-            const containingClassDecl = node.parent as ClassDeclaration;
```
Andarist, line 39198: this cast was "unsafe" and actually redundant, `node.parent` is already types as `ClassLikeDeclaration`

src/compiler/checker.ts
```diff
@@ -12030,7 +12030,6 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     /**
      * The base constructor of a class can resolve to
      * * undefinedType if the class has no extends clause,
-     * * unknownType if an error occurred during resolution of the extends expression,
```
Andarist, line 12033: Based on my analysis and the existing test suite - this can't actually be returned. Perhaps `checkExpression` might return `unknownType` but that would fail the `isConstructorType` check and thus it would lead to returning `errorType`

src/compiler/checker.ts
```diff
@@ -29143,6 +29142,10 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             return errorType;
         }
 
+        if (classDeclarationExtendsNull(classLikeDeclaration)) {
+            return isCallExpression ? errorType : nullWideningType;
```
Andarist, line 29146: The error itself for calls is already raised by `checkConstructorDeclarationDiagnostics`. I thought about moving it here but I concluded that leaving it there is nice because it only errors on the very first super call in the constructor. In here, we'd be erroring on all calls.

tests/baselines/reference/classExtendsNull3.errors.txt
Andarist, line 1: This is a bonus fix - currently no error is raised here and error type (aka `any`) leaks without being noticed. I'm not entirely sure if this should **always** be an error since I don't really understand what the "reopening class with interface" is meant to do and what behaviors are associated with it.

src/compiler/checker.ts
```diff
@@ -29143,6 +29142,10 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             return errorType;
         }
 
+        if (classDeclarationExtendsNull(classLikeDeclaration)) {
+            return isCallExpression ? errorType : nullWideningType;
+        }
+
         const classType = getDeclaredTypeOfSymbol(getSymbolOfDeclaration(classLikeDeclaration)) as InterfaceType;
         const baseClassType = classType && getBaseTypes(classType)[0];
```
Andarist, line 29150: In a normal situation, there would be no base class in the case of `extends null`. `resolveBaseTypesOfClass` returns an empty array if the `baseConstructorType` there is not an object (while `null` is a valid return value of `getBaseConstructorTypeOfClass`)

We don't have direct access to this resolution though, all that we currently have are those base types here and that's roughly equivalent of:
```ts
type BaseTypeOfClassArr = [ConstructorType?]
type BaseTypesOfInterfaces = [...Type[]]
type BaseTypes = [...BaseTypeOfClassArr, ...BaseTypesOfInterfaces]
```

There is always at most one base type of class but it's also optional. If it's there then it's always the first item in the array and that's why `[0]` is used here. So the problem was caused here by the fact that this array was not empty and yet the first item in the array was not a base type of class. That allowed the code to continue here, returning a value that didn't have any construct signatures (`null`).

tests/baselines/reference/classExtendsNull3.errors.txt
Andarist: It's worth noting that this **isn't** a crash:
```ts
class C extends null {
    static method() {
        console.log('undefined', super.whatever)
    }
}
C.method()
```

So technically a non-call super expression with a base constructor null could resolve to something like `Record<PropertyKey, undefined>` but I honestly don't think it's worth it 😉

src/compiler/checker.ts
jakebailey: Should this say `errorType`, then?

jakebailey (approved):

jakebailey:
Almost forgot

@typescript-bot test top200
@typescript-bot run dt

jakebailey:
The heck is with that xstate perf

jakebailey:

Andarist:
> The heck is with that xstate perf

Not sure but I like it 🤣

src/compiler/checker.ts
Andarist: Ye, not sure why I removed it instead of adjusting it. I just brought it back 😉

jakebailey (approved):

PR #55512, opened 2023-08-25
Commit to parsing a module declaration when `declare module` is met

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/11167

If this issue is no longer meant to be fixed - feel free to close this PR and that issue 😉

### DIFF
--- a/src/compiler/parser.ts
+++ b/src/compiler/parser.ts
@@ -7163,9 +7163,9 @@ namespace Parser {
                     if (scanner.hasPrecedingLineBreak()) {
                         return false;
                     }
-                    if (previousToken === SyntaxKind.DeclareKeyword && token() === SyntaxKind.TypeKeyword) {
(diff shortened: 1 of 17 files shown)


fatcerberus:
I thought this was a good point: https://github.com/microsoft/TypeScript/issues/11167#issuecomment-256502354

Andarist:
@fatcerberus this was meant to still be an error, similarly to what is reported here:
```ts
declare type
Z = {} // Line break not permitted here.(1142)
```

I forgot to push out that extra change though, thanks for the ping!

fatcerberus:
@Andarist thanks, I wasn’t sure because you say this “fixes #11167” but that issue actually asks for this to *not* be an error.

Andarist:
a more accurate wording here would be "this can be used as a resolution of XXX" :p

DanielRosenwasser:

PR #55515, opened 2023-08-25
Clean up helpers which are always present in ES2020

jakebailey:
As of TS 5.1, we require Node 14.17 / ES2020. This PR removes our conditional use of:

- `String.prototype.trim`
- `String.prototype.trimEnd`
- `String.prototype.trimStart`
- `String.prototype.codePointAt`
- `String.prototype.localeCompare`
- `String.prototype.toLocaleUpperCase`
- `Intl`

We've used `.trim()` in quite a few places. We already unconditionally use `Intl` after the import organization PR.

It also removes `padLeft` and `padRight`, which could have been replaced by `String` methods a while ago but were not.

### DIFF
--- a/src/compiler/core.ts
+++ b/src/compiler/core.ts
@@ -2240,11 +2240,7 @@ export function getStringComparer(ignoreCase?: boolean) {
  * Creates a string comparer for use with string collation in the UI.
  */
 const createUIStringComparer = (() => {
-    let defaultComparer: Comparer<string> | undefined;
-    let enUSComparer: Comparer<string> | undefined;
-
-    const stringComparerFactory = getStringComparerFactory();
-    return createStringComparer;
+    return createIntlCollatorStringComparer;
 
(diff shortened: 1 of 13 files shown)


jakebailey:
@typescript-bot perf test this full

DanielRosenwasser:
If you're open to it, can you also remove uses of `padLeft`?

jakebailey:
Yes, great idea; not sure why I forgot those.

DanielRosenwasser (approved):

src/compiler/core.ts
```diff
@@ -2261,70 +2256,6 @@ const createUIStringComparer = (() => {
         const comparer = new Intl.Collator(locale, { usage: "sort", sensitivity: "variant" }).compare;
         return (a, b) => compareWithCallback(a, b, comparer);
```
jakebailey, line 2257: Another cleanup here would be to remove this fallback; IIRC you can get all of the way there with only collator options (though this fallback method is not uncommon; it's what eslint-plugin-simple-import-sort does).

PR #55517, opened 2023-08-26
Fix good lints given by eslint-plugin-unicorn

jakebailey:
Not final; I'm not totally sure I want to add this dependency (it's kinda big).


scripts/errorCheck.mjs
```diff
@@ -27,7 +27,7 @@ async function checkErrorBaselines() {
             let g;
             while (g = errRegex.exec(baseline)) {
                 const errCode = +g[1];
-                const msg = keys.filter(k => messages[k].code === errCode)[0];
```
fatcerberus, line 30: This just seems like a weird pattern in general.  Particularly since `filter` doesn't short-circuit.

jakebailey: I don't disagree, I didn't write it, but I am fixing it 😅

src/compiler/commandLineParser.ts
```diff
@@ -2615,7 +2613,7 @@ export function getCompilerOptionsDiffValue(options: CompilerOptions, newLine: s
     return getOverwrittenDefaultOptions();
 
     function makePadding(paddingLength: number): string {
-        return Array(paddingLength + 1).join(" ");
+        return new Array(paddingLength + 1).join(" ");
```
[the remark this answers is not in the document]
jakebailey: Yeah: https://github.com/microsoft/TypeScript/pull/55515#issuecomment-1695115855

jakebailey:

Issue #5551, opened 2015-11-06
Same references from different (but not found) modules


jkillian:
If you have code like the following:

``` typescript
import { resolve as resolveUrl } from 'url';
import { resolve } from 'path';
```

`getDocumentHighlights` marks both references of `resolve` as the same if the modules 'url' and 'path' can't be found. However, if 'url' and 'path' can be found, it correctly differentiates between the two.

I know this is somewhat of an edge-case, but is this intended behavior? Or is it a bug?

mhegazy:
they both given the same symbol (`unknownSymbol`), the find all references code will think they are because they also have the same name. we could exclude the unknownSymbol from this logic. but would not say it is a priority.

jakebailey:
Thanks for checking; I bisected and this was fixed in https://github.com/microsoft/TypeScript/commit/e8c3d548eb66c020b02cfa135b492f9480ef639c, but it's not clear to me that this actually has a test and it'd be nice to have one via a fourslash baseline or something.

PR #13643, opened 2017-01-23
Support find-all-references for default exports

andy-ms:
Fixes #11551

PR #55519, opened 2023-08-26
Add a test for finding all references of same-named imports from two missing modules

Andarist:
closes https://github.com/microsoft/TypeScript/issues/5551

@jakebailey [bisected the actual fix](https://github.com/microsoft/TypeScript/issues/5551#issuecomment-1662690796) to https://github.com/microsoft/TypeScript/pull/13643

### DIFF
--- a/tests/baselines/reference/findAllRefsMissingModulesOverlappingSpecifiers.baseline.jsonc
+++ b/tests/baselines/reference/findAllRefsMissingModulesOverlappingSpecifiers.baseline.jsonc
@@ -1,0 +1,43 @@
+// === findAllReferences ===
+// === /tests/cases/fourslash/findAllRefsMissingModulesOverlappingSpecifiers.ts ===
+// // https://github.com/microsoft/TypeScript/issues/5551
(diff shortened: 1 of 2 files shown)


tests/baselines/reference/findAllRefsMissingModulesOverlappingSpecifiers.baseline.jsonc
```diff
@@ -0,0 +1,43 @@
+// === findAllReferences ===
+// === /tests/cases/fourslash/findAllRefsMissingModulesOverlappingSpecifiers.ts ===
+// // https://github.com/microsoft/TypeScript/issues/5551
+// import { resolve/*FIND ALL REFS*/ as resolveUrl } from "idontcare";
```
Andarist, line 4: It's worth noting that this one just doesn't find any references while the other one finds the reference at the location of the request. I think it's fine - the most important thing is that both locations are not treated as references to the same thing.

jakebailey (approved):

PR #55522, opened 2023-08-26
Add an extra test for `as const satisfies` with a mutable array

Andarist:
a small, test-only, followup to: https://github.com/microsoft/TypeScript/pull/55229


DanielRosenwasser:
Love that you added this test. Can you add the other ordering of `satisfies ... as const`?
```

Andarist:
`as const` can only be applied on literals so the only thing that we'd get out of this test is that it errors properly with:
> A 'const' assertions can only be applied to references to enum members, or string, number, boolean, array, or object literals.ts(1355)

DanielRosenwasser:
Hm - then I can go either way on it.

DanielRosenwasser (approved):

PR #55529, opened 2023-08-27
fix(55494): Invalid declaration with computed property using imported symbol

### LINKED ISSUE #55494: Invalid declaration with computed property using imported symbol
### 🔎 Search Terms

computed property declaration destructure binding

### ⏯ Playground Link

https://www.typescriptlang.org/dev/bug-workbench/?lib=lib.es2015.d.ts&ts=5.1.6#code/PTAEAEBMFMGMBsCGAnRAXAlgewHYC5Q1kBXaAKBAngwCMDoBnANgrHADMN5odEBbaAVi400AB5oAdGgZlxAByzI0oYTgYqA0tACeoALygAyjr40s8ABQBKANysInbrwEEMOGGOmyMfRctAAb1BtPQBfUHZkLD5QACJJYDVRCTj7BSUVdmIcWExcUAB3DDQACwBhEXE0S2CAbVCAXQIAN0R4UlAwgkQcHWsCDWR3AHMgslBJyeRoNGJkHFA2juh7MLIgA

### 💻 Code

```ts
// @declaration: true
// @lib: es6
// @filename: context.ts
export const Key = Symbol();
// @filename: index.ts
import { Key } from "./context";
export function withContext({ [Key]: value }: any): string {
     return value;
}
```

### 🙁 Actual behavior

In `index.d.ts` the import of `Key` is removed  leading to an invalid declaration file

```ts
// index.d.ts
export declare function withContext({ [Key]: value }: any): string;
```

### 🙂 Expected behavior

In `index.d.ts` the import of `Key` should be kept

```ts
// index.d.ts
import { Key } from "./context";
export declare function withContext({ [Key]: value }: any): string;
```

Seems like `filterBindingPatternInitializersAndRenamings` in `declarations.ts` does not call `checkEntityNameVisibility` to mark the import as used in the declaration.

a-tarasyuk:
Fixes #55494

### DIFF
--- a/src/compiler/transformers/declarations.ts
+++ b/src/compiler/transformers/declarations.ts
@@ -100,6 +100,7 @@ import {
     isBindingPattern,
     isClassDeclaration,
     isClassElement,
+    isComputedPropertyName,
     isDeclaration,
     isElementAccessExpression,
     isEntityName,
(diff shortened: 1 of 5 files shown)


weswigham:
Thanks!

Issue #55526, opened 2023-08-27
Assigning property values square bracket access and exactOptionalPropertyTypes enabled has inconsistent behaviour


aswinsvijay:
### 🔎 Search Terms

exact optional
indexed access
indexed assignment
square bracket access

### 🕗 Version & Regression Information

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about exact optional property types

### ⏯ Playground Link

https://www.typescriptlang.org/play?exactOptionalPropertyTypes=true&ts=5.2.2#code/N4KABGAuCeAOCmYDKB7AtvSALAlgOwHMwBeMUCCAQwH4AuMAZ0gCd8CBucMAX064GMUeJmAAe9VBmxsSZXiC4AzAK55+kHELCUGDHATwAVFAGl40ABQBrc-RvQUi5Oky5CAGjAA3SgBtl8PRMrIRgAD5gqgAm8Ir48FEAlGRcEKIA2vYAurI+-vDsFGAA9MVg8MzMKMwKRaIAdJS5fgGFRRCl5ZXVqWLpAOSU-Tmkea1FnRVVNRDctdq6+kam5haD-Z7RsfFJnHMg5FBwiJKuMqSHVHSMLGycRQBG18F3XPICQiLizlJuRBfvCAqNQaLQ6PQGYxmSz2OzmRw-M4ebwtQI3EJECJbOJ4BLJS59bLNfJtDplPAoLrTeZpRrE8btEplKY9OoDIYjFEkibM7ozHjzcFLKGrdabPAxHF4vZAA
### 💻 Code

```ts
{
  type Something = {
    a?: string;
  };

  const x: Something = {};

  function assignToKey(key: keyof Something, value: string | undefined) {
    x[key] = value;    // error

    x.a = value;       // error
    x['a'] = value;    // error
  }

  assignToKey('a', undefined);
}

{
  type Something = {
    a?: string;
    b?: string;
  };

  const x: Something = {};

  function assignToKey(key: keyof Something, value: string | undefined) {
    x[key] = value;    // no error

    x.a = value;       // error
    x['a'] = value;    // error
  }

  assignToKey('a', undefined);
}
```

### 🙁 Actual behavior

Assigning `undefined` to properties when using variable as key shows no error, where the type of the variable is a union of strings.
Accessing a particular property and assigning `undefined` shows error.
If the variable used as key name has the type as a single string, it shows error.

### 🙂 Expected behavior

[...]

PR #55531, opened 2023-08-27
Fixed an issue with assignments to dynamic properties with `exactOptionalPropertyTypes`

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/55526

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -32200,7 +32200,10 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         // accessor, or optional method.
         const assignmentKind = getAssignmentTargetKind(node);
         if (assignmentKind === AssignmentKind.Definite) {
-            return removeMissingType(propType, !!(prop && prop.flags & SymbolFlags.Optional));
+            // since this is used as an assignment target we always want to remove the missing type
+            // usually this is done conditionally based on the `prop.flags`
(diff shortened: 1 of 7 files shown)


src/compiler/checker.ts
```diff
@@ -32200,7 +32200,10 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         // accessor, or optional method.
         const assignmentKind = getAssignmentTargetKind(node);
         if (assignmentKind === AssignmentKind.Definite) {
-            return removeMissingType(propType, !!(prop && prop.flags & SymbolFlags.Optional));
+            // since this is used as an assignment target we always want to remove the missing type
+            // usually this is done conditionally based on the `prop.flags`
+            // the prop might not be available here though if the assignment is done to a dynamic property
+            return removeMissingType(propType, /*isOptional*/ true);
```
Andarist, line 32206: This makes `isOptional` parameter somewhat misleading here.

The idea is that if we have `missingType` in the `propType` here we should just always remove it. It could only be obtained here from that access expression itself (I think?) so if we write to that access expression it should be always safe to just try to remove it - without even checking if the prop was optional or not. 

To fix the off-putting usage of this parameter I could:
- rename the parameter
- leave it as is since there is an inline comment that mentions why it's done this way
- inline `removeMissingType` here

PR #55547, opened 2023-08-28
Infer `unknown` at value positions of objects inferred from `keyof T`

Andarist:
I noticed this when investigating a completion problem. Using `any` here might "accidentally leak" (which happens in the changed baselines) without the user noticing. I think it would be great to change this to `unknown` unless there is a strong reason why it should stay this way.

Note that this function was introduced by @sandersn in https://github.com/microsoft/TypeScript/pull/19227 . This was in pre(historic)-3.0 times when `unknown` wasn't even a thing in the language yet 😉 So back then `unknown` wasn't even an option.

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -24526,7 +24526,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             isTypeParameterAtTopLevel(getReturnTypeOfSignature(signature), typeParameter);
     }
 
-    /** Create an object with properties named in the string literal type. Every property has type `any` */
+    /** Create an object with properties named in the string literal type. Every property has type `unknown` */
     function createEmptyObjectTypeFromStringLiteral(type: Type) {
         const members = createSymbolTable();
         forEachType(type, t => {
@@ -24535,7 +24535,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             }
             const name = escapeLeadingUnderscores((t as StringLiteralType).value);
             const literalProp = createSymbol(SymbolFlags.Property, name);
-            literalProp.links.type = anyType;
+            literalProp.links.type = unknownType;
             if (t.symbol) {
                 literalProp.declarations = t.symbol.declarations;
                 literalProp.valueDeclaration = t.symbol.valueDeclaration;
--- a/tests/baselines/reference/inferObjectTypeFromStringLiteralToKeyof.types
+++ b/tests/baselines/reference/inferObjectTypeFromStringLiteralToKeyof.types
@@ -14,8 +14,8 @@ declare var two: "a" | "d";
 >two : "a" | "d"
 
 const x = inference1(two);
->x : { a: any; d: any; }
->inference1(two) : { a: any; d: any; }
+>x : { a: unknown; d: unknown; }
+>inference1(two) : { a: unknown; d: unknown; }
 >inference1 : <T>(name: keyof T) => T
 >two : "a" | "d"
 
--- a/tests/baselines/reference/keyofInferenceIntersectsResults.types
+++ b/tests/baselines/reference/keyofInferenceIntersectsResults.types
@@ -27,15 +27,15 @@ const a = foo<X>('a', 'b'); // compiles cleanly
 >'b' : "b"
 
 const b = foo('a', 'b');    // also clean
->b : { a: any; } & { b: any; }
->foo('a', 'b') : { a: any; } & { b: any; }
+>b : { a: unknown; } & { b: unknown; }
+>foo('a', 'b') : { a: unknown; } & { b: unknown; }
 >foo : <T = X>(x: keyof T, y: keyof T) => T
 >'a' : "a"
 >'b' : "b"
 
 const c = bar('a', 'b');    // still clean
->c : { a: any; } & { b: any; }
->bar('a', 'b') : { a: any; } & { b: any; }
+>c : { a: unknown; } & { b: unknown; }
+>bar('a', 'b') : { a: unknown; } & { b: unknown; }
 >bar : <T>(x: keyof T, y: keyof T) => T
 >'a' : "a"
 >'b' : "b"


weswigham (approved):

weswigham:
@DanielRosenwasser I think this is a pretty good break - should we just pull it in now, or should we wait until a different part of the release cycle?

DanielRosenwasser:
@typescript-bot pack this
@typescript-bot test this
@typescript-bot test top300
@typescript-bot run dt

Andarist:
A minimal repro case of the above problem: [TS playground](https://www.staging-typescript.org/play?ts=5.3.0-pr-55547-8#code/CYUwxgNghgTiAEAzArgOzAFwJYHtX1ES1RAGEcBbABzxFQwB4AFGHKgZ3hAA8M7hOAJXA4YwBuwwxiAcwA08KKgCeAPlUAKALAAoePHYgMyKgC54Gqqw7mWbdgEp4AXlXwAbjizA5u-W2w8dgB+cwBvP314K3tQiwBrEGUcRHg7DgcAbQBdAG5IgF9fHQdzS2t2WwqnV0UVfJ1dQmIyShoSem09eAYAFS5efk5JaVQZTQB9GJt4MPgKdhlzfoKatwju-ThjGHwNNdmChv0iyI2o6IrzTIAiBZkb7OKT4odcoA). It looks fixable on the Vue's side. I think that it actually somewhat proves how this is an improvement because the return type of [this function](https://github.com/vuejs/core/blob/b775b71c788499ec7ee58bc2cf4cd04ed388e072/packages/runtime-core/src/apiDefineComponent.ts#L115) utilizes the inferred type of `Props`. So even though this setup function tries to constraint `Props` to `{ msg: T }` where `T extends string` the inferred `{ msg: any }` leaks~ and the resulting function is callable with incompatible prop types.

sandersn:
@weswigham would you be able to talk about it at the Friday meeting?

DanielRosenwasser:
I guess the only thing I might say is that we could change this behavior based on `noImplicitAny`, but tests do come up clean.

sandersn:
That's fine with me if we have a plan to fix Vue.

Andarist:
I dug into this case deeper and I think there are two things that could be done about this and both are quite independent from this. They could become prerequisites for this PR.

One thing about the example above is that a pretty low priority inference wins over something better that could be inferred from the generic signature. I’ve opened this issue about it: https://github.com/microsoft/TypeScript/issues/56931 and an experimental fix for it: https://github.com/microsoft/TypeScript/pull/56939 . The extended community test suite didn’t fail there for Vue so likely the combination of that PR and this PR here wouldnt either.

The second issue is that by inferring an object with properties typed as `unknown` the inference might fail the constraint check. `T` could be constrained to `Record<string, number>` (or something else).

Now… is that a problem? Inferring anything here is quite unsound **unless** there are some other things at play (like a return context or smth). `unknown` is the safest option because it assumes that the function might create an object with those properties in its implementation (at least when those keys are passed at argument position). We can’t safely say anything about its values.

Going further… if a union of keys is passed not within an array or something… we can’t even say that those properties are required. If they are passed within an array we similarly can’t guarantee that this array contain all of the keys that we can see in the type system. So really we can only say that all properties are optional.

The question is - what are the use cases for this inference? Why it was introduced? Are those use cases covered today by more sound alternatives?

OTOH, similar unsound situations can always be achieved when supplying type arguments manually (and likely in other situations too). So perhaps the mental model is that it’s not the type system’s responsibility to ensure soundness here but rather the function/signature’s author.

If the type system is meant to allow this regardless of constraints then `unknown` is not the best fit here. Note that the potential constraint failure is not a new problem:

```ts
declare function test1<T>(keys: keyof T): T;
const result1 = test1("" as "a" | "b"); // { a: any; b: any; }

declare function test2<T extends Record<string, never>>(keys: keyof T): T;
const result2 = test2("" as "a" | "b"); // Record<string, never>
```

This case is far-fetched but conceptually the point stands - the better fit would be the constraint of each potential property.

As @weswigham noted [here](https://github.com/microsoft/TypeScript/pull/56300#discussion_r1385736269), it’s not enough to call `getBaseConstraintOfType` in a situation like this because it wouldn't properly work with constraints depending on other type parameters (like `Record<string, U>`). Since this fake object created based on the keys is created during the main inference pass, it might be enough to obtain `getConstraintOfType`, leave `U` in it and hope that it gets instantiated down the road.

I'm not sure if that would actually happen though (I will try after we establish here how this whole thing should work). If not then I'd say that this could utilize the reverse typed machinery. Instead of creating an anonymous object here, we could create a reverse mapped type (with a fake source object) and depend on the fix from https://github.com/microsoft/TypeScript/pull/56300 to get correct (and individual) constraints for each property.

jfet97:
> One thing about the example above is that a pretty low priority inference wins over something better that could be inferred from the generic signature. I’ve opened this issue about it: https://github.com/microsoft/TypeScript/issues/56931 and an experimental fix for it: https://github.com/microsoft/TypeScript/pull/56939 . The extended community test suite didn’t fail there for Vue so likely the combination of that PR and this PR here wouldnt either.

@Andarist  I tested your minimal repro with #56939 (here is [the playground](https://tsplay.dev/mAL1kN)):

```ts
declare function defineComponent<Props extends Record<string, any>>(
  setup: (props: Props) => void,
  options?: {
    props?: (keyof Props)[];
  },
): (props: Props) => any;

defineComponent(
  <T extends string>(_props: { msg: T }) => {
    return () => {};
  },
  {
    props: ["msg"],
  },
);
```

The inferred type of the call to `defineComponent` is:

```ts
function defineComponent<{
    msg: T;
}>(setup: (props: {
    msg: T;
}) => void, options?: {
    props?: "msg"[] | undefined;
} | undefined): <T extends string>(props: {
    msg: T;
}) => any
```

Therefore the `Props` type parameter of `defineComponent` has been unified with `{ msg: T }`, where `T` is the type parameter of the passed `setup` callback. Seems good to me :D

jfet97:
@Andarist 
Thinking about it better though, I think there might be a problem. With #56939 the following would be allowed, but it shouldn't because `"test"` has not been explicitly declared as runtime prop:

```ts
defineComponent(
  <T extends string>(_props: { msg: T, test: T }) => {
    return () => {};
  },
  {
    props: ["msg"],
  },
);
```

If you see [the documentation](https://vuejs.org/api/general.html#function-signature):
image

it seems that they want to force you to manually declare the props inside the array before allowing you to use them as arguments of the `setup` function. In fact, currently, the snippet above produces an error.

So it seems that the fact the `props` array has somehow an higher priority is the behavior they desire.

Andarist:
Yeah, I understand that - in general - this other PR would change the behavior for Vue today. It's great to know **how** it would affect them negatively though, that I didn't know, and thanks for that! 

Though, their whole signature is somewhat dubious and I don't find it to be a compelling use case that would prove that this other PR is worse than what we have today.

Their docs mention:
> The main use case for this signature is with TypeScript (and in particular with TSX), as it supports generics:

But if we actually, try to utilize this then I'm not quite sure about what kind of generics support we are talking about:
```ts
import { defineComponent, ref } from "vue";
import { h } from "@vue/runtime-dom";

const Comp = defineComponent(
  <T extends string | number>(props: { msg: T; list: T[] }) => {
    const count = ref(0);

    return () => {
      return h("div", count.value);
    };
  },
  {
    props: ["msg", "list"],
  },
);

Comp;
// ^? const Comp: (props: { msg: any; list: any; } & {}) => any
```

Maybe I don't understand how their compiler uses this but TS-wise this leaves me quite confused since this "generics support" leaves the output totally unsafe.

I also think that their API choice for this leaves some nasty runtime holes here and that they could do better. If they need to declare required keys~ then they really should use an object for that and not an array. An array typed as `Array<keyof Foo>` might contain *some* of `Foo`'s keys or even none of them. So while this today sort of works at type level - it isn't overly correct anyway. That's exactly why, with my proposed inference change, the array with a subset of keys is accepted.

If they would use an object to declare those props then it would work with my change **and** it would improve the runtime safety. The location of the error would change but I don't think that's a big deal ([TS playground](https://www.staging-typescript.org/play?ts=5.4.0-pr-56939-8#code/CYUwxgNghgTiAEAzArgOzAFwJYHtX1ES1RAGEcBbABzxFQwB4AFGHKgZ3hAA8M7hOAJXA4YwBuwwxiAcwA08KKgCeAPlUAKALAAoePHYgMyKgC54Gqqw7mWbdgEp4AXlXwAbjizA5u-W2w8dgB+cwBvP314K3tQ+GEwUXEAaxBlHER4Ow4FNGTUHAB3VFVIgF9fHQdzS2t2WzqnV0UVAG5dXUTUSXg4dmQIDBcCECIScmpaem09eAYAFS5efk5JaVQZTQB9GJt4MPgKdhlzeYUILElTgG0AXXgyprcI2f04Yxh8DSf9svbZiqRF5RaJ1cLwAD0EK4MFYMF6IBoMD4wHgAAsQHB4IUMfgCkMoBAIKD7IosaBILAQKiMXBIiCjid4FJkCBKiD9FD4BcrsyYKzIdC0IkKBQ6EMMDgDCAEBgMfBiIhMXBUTI6JisGADFgZKgoB8QPSHuzjboHP9dH0BhhdFyAHrBIA)):
```ts
declare function defineComponent<Props extends Record<string, any>>(
  setup: (props: Props) => void,
  options?: {
    props?: Record<keyof Props, unknown>
  },
): (props: Props) => any;

const result = defineComponent(
  <T extends string>(_props: { msg: T, list: T[] }) => {
    return () => {};
  },
  {
    props: { // error reported here when not all props are declared here
      msg: true,
      // list: true // uncomment to see the inferred generic signature
    },
  },
);
```

> @Andarist Thanks to https://github.com/microsoft/TypeScript/pull/55811 , if https://github.com/microsoft/TypeScript/pull/56939 lands, they could use a definition similar to the following one to error out when someone destructures a property not declared in the props array:

Nice ❤️

sandersn:
I'm trying to understand what state this PR is in. Here's my attempted summary:
1. @weswigham likes the basic idea
2. The implementation is not quite right.
3. It could be improved if two other PRs are merged.
4. Vue currently breaks with this PR, and I can't tell if it's indicative of lots of other breaks. But it seems like it's improvement in semantics and the Vue code is suspicious.

On (3), I'm not sure whether it should actually be "another approach would work better if two other PRs are merged". If so, we should close this one. Even if this PR is a good start, it might not be worth keeping around unless those other two PRs are worth taking too.

On (4), if it breaks lots of code, I think it's too minor an improvement to be worthwhile. If not a lot breaks, it could be fine.

Issue #48194, opened 2022-03-09
Add option for navto to exclude built-in libs and files from node_modules


mjbvz:
# Feature request

### 🔎 Search Terms
- navto
- go to symbol in workspace
- tsserver

### Request
Currently `navto` includes all known symbols in a project. This include both symbols from the stdlib and from `node_modules`. However the most common use case for `navto` in my experience is jumping to another symbol within my code. 

I propose we add new options to `navto` that let us exclude:

- Symbols from the standard library
- Any symbols from `node_modules`

This could be exposed as either a single new option on the request that's called something like `justMyCode`, or as two separate options for more fine grained control. In either case, we would expose the option(s) to users in the VS Code settings UI

Extracted https://github.com/microsoft/vscode/issues/46718 as the most common use case there seems to be excluding files under `node_modules`

RyanCavanaugh:
Let's talk next week about how this might be handled

PR #55550, opened 2023-08-28
Return all symbols in navto for empty string pattern 

### LINKED ISSUE #55483: navto returns no results for empty strings
### 🔍 Search Terms

- navto
- Go to symbol

Not a regression

### ⏯ Playground Link

_No response_

### 💻 Code

In a workspace, request `navTo` with an empty search string

```
2023-08-23 10:18:38.078 [trace] <semantic> Sending request: navto (98). Response expected: yes. Current queue length: 0 {
    "searchValue": "",
    "maxResultCount": 256
}
```

### 🙁 Actual behavior

No results are returned

### 🙂 Expected behavior

We should return a list of all symbols in the workspace

_No response_

gabritto:
Fixes #55483.

### DIFF
--- a/src/harness/fourslashImpl.ts
+++ b/src/harness/fourslashImpl.ts
@@ -3766,7 +3766,8 @@ export class TestState {
 
     public verifyNavigateTo(options: readonly FourSlashInterface.VerifyNavigateToOptions[]): void {
         for (const { pattern, expected, fileName } of options) {
-            const items = this.languageService.getNavigateToItems(pattern, /*maxResultCount*/ undefined, fileName);
+            const file = fileName && this.findFile(fileName).fileName;
+            const items = this.languageService.getNavigateToItems(pattern, /*maxResultCount*/ undefined, file);
             this.assertObjectsEqual(
                 items,
                 expected.map((e): ts.NavigateToItem => ({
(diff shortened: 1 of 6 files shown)


DanielRosenwasser:
Does navto do an after-the-fact filter today? Or does it filter as it aggregates symbols? Also curious if you've tried this on slightly bigger codebases like VS Code or TypeScript itself.

andrewbranch (approved):

gabritto:
> Does navto do an after-the-fact filter today? Or does it filter as it aggregates symbols? Also curious if you've tried this on slightly bigger codebases like VS Code or TypeScript itself (for both _Go To Symbol in Editor_ and _Go To Symbol in Workspace_).

Do you mean if vscode is going to filter the navto results TS Server sends? If so, it doesn't. I'll wait to merge this until we decide on the changes needed on the editor side.

gabritto:
We might also want to consider doing https://github.com/microsoft/TypeScript/issues/48194, but I don't know what was discussed/decided about it the last time it was brought up.

gabritto:

gabritto:
> @gabritto The results of the perf run you requested are in!
> 
> Here they are:

I just remembered the benchmark sends a navto request with some pre-defined string... So the perf tests are not relevant here. But I tested this locally and there isn't a delay on the vscode codebase, but for the TS codebase there's a ~1.5s delay between the command and the results showing up.

DanielRosenwasser:
No, I was asking if it's the case that running navTo with `"getTypeFrom"` will walk across the whole project and then filter the symbols based on `getTypeFrom`, or if the filtering is done during the walk itself.

gabritto:
> No, I was asking if it's the case that running navTo with `"getTypeFrom"` will walk across the whole project and then filter the symbols based on `getTypeFrom`, or if the filtering is done during the walk itself.

It'll walk across all files in all the projects, collect all named declarations (if they haven't been computed yet), and then look at every named declaration to see if it matches `"getTypeFrom"`.

Issue #55545, opened 2023-08-28
Wrong code suggestion when using "Mapped Types" as function parameter


lNaymonl:
Type: **Bug**

Extension version: 5.3.20230827

Version: 1.81.1 (user setup)
Commit: 6c3e3dba23e8fadc360aed75ce363ba185c49794
Date: 2023-08-09T22:22:42.175Z
Electron: 22.3.18
ElectronBuildId: 22689846
Chromium: 108.0.5359.215
Node.js: 16.17.1
V8: 10.8.168.25-electron.0
OS: Windows_NT x64 10.0.19045

Problem:
The Codesuggestion shows "1", "2", "3", "4" event though `a` is "valA".  When hovering it also shows that `b` has type "1" | "2". And when i select "3" or "4" it gives me the following error: 
`Argument of type '"3"' is not assignable to parameter of type '"1" | "2"'.`

![Screenshot (7)](https://github.com/microsoft/vscode-typescript-next/assets/65818963/d12aad72-a72e-4c03-bb11-239877e25b36)

Expected:
The Codesuggestion should only display"1" and "2".

Source Code:
```typescript
enum myEnum {
  valA = "valA",
  valB = "valB",
}

interface myEnumParamMapping {
  ["valA"]: "1" | "2";
  ["valB"]: "3" | "4";
}

function myFunction<K extends keyof typeof myEnum>(a: K, b: myEnumParamMapping[K]) {
  console.log({ a, b });
}

myFunction("valA", "2");
```

13OnTheCode:
![微信截图_20230829014357](https://github.com/microsoft/TypeScript/assets/137921275/fb3e5e34-5e0c-47f7-9411-5e4c1d7bef41)

@lNaymonl  My TypeScript version is 5.2.2, and when I tried to reproduce your issue, everything seems to be working fine, unless I missed something.

13OnTheCode:
@andrewbranch 

```json
"compilerOptions": {
    "strict": true,
    "noEmit": true,
    "allowJs": true,
    "skipLibCheck": true,
    "isolatedModules": true,
    "esModuleInterop": true,
    "strictNullChecks": true,
    "resolveJsonModule": true,
    "useDefineForClassFields": true,
    "noUncheckedIndexedAccess": true,
    "allowImportingTsExtensions": true,
    "allowSyntheticDefaultImports": true,
    "forceConsistentCasingInFileNames": true
}
```

andrewbranch:
[...]

PR #53996, opened 2023-04-24
Fixed some string literal argument completions depending on resolved signature

Andarist:

PR #55552, opened 2023-08-28
Fixed a string completions regression when requested directly in argument position

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/55545 (a regression from https://github.com/microsoft/TypeScript/pull/53996 )

it turns out that reusing `wildcardType` for this wasn't ideal 😉 cc @andrewbranch


andrewbranch (approved):
Thank you!

Andarist:
@andrewbranch technically, this was a regression from 5.1. It's not rly critical but perhaps it's worth cherry-picking to 5.2?

andrewbranch:
That’s fine with me. @DanielRosenwasser?

PR #55554, opened 2023-08-28
Add friendly name to intrinsics in debug helpers

jakebailey:
We create multiple intrinsic types with the same intrinsic name, but that means that the debug helpers can't distinguish between them at all which is frustrating while debugging. I frequently have to do things like `type === wildcardType` to see which of the like ~~4~~ 5 `any` intrinsics we have.

This PR adds a debug name, which will be included in parens when present, like:

image

On intrinsic creation, the checker throws if you create an intrinsic with the same name more than once, to make sure we never forget.

@Andarist


jakebailey:
Doubt that this will actually change perf as the number of these we create is limited, but:

jakebailey:
Well, I guess this is a good opportunity to see what happens when perf fails at the TS build stage. Retrying:

Andarist:
😍

src/compiler/debug.ts
```diff
@@ -638,7 +638,7 @@ export namespace Debug {
                         this.flags & TypeFlags.BigIntLiteral ? `LiteralType ${(this as BigIntLiteralType).value.negative ? "-" : ""}${(this as BigIntLiteralType).value.base10Value}n` :
                         this.flags & TypeFlags.UniqueESSymbol ? "UniqueESSymbolType" :
                         this.flags & TypeFlags.Enum ? "EnumType" :
-                        this.flags & TypeFlags.Intrinsic ? `IntrinsicType ${(this as IntrinsicType).intrinsicName}` :
+                        this.flags & TypeFlags.Intrinsic ? `IntrinsicType ${(this as IntrinsicType).intrinsicName}${(this as IntrinsicType).debugIntrinsicName ? ` (${(this as IntrinsicType).debugIntrinsicName})` : ""}` :
```
Andarist, line 641: How hard it would be to make `__debugTypeToString()` print this too?

jakebailey:
Noticed that my example wasn't perfect; should really check intrinsics first as "nullable type" isn't that helpful:

image

src/compiler/debug.ts
jakebailey: Hm, the problem is that `__debugTypeToString` goes through `typeToString`, where you kinda want to see what the user would see. But it could potentially be done.

Andarist: Maybe `TypeFormatFlags.Debug` could be added so it could be passed down to `typeToString` by `__debugTypeToString`? Those flags are not all exactly limited to formatting-concerns right now ;p

jakebailey: That was my first thought but then I looked at it and got scared trying to decide if I had to mirror the flag in `NodeBuilderFlags` and also if there was even room in the limited number of bits we have + if I'm allowed to overlap... But I think it's doable.

jakebailey: Yeah, it has to be in `NodeBuilderFlags`, which is full effectively full, and then to be a part of `NodeBuilderFlagsMask` so it can be the same in both flags enums, it can't be internal. Not great.

Andarist: Perhaps not worth it then right now. What is nice about `typeToString` though is that it's recursive and those types might sometimes appear at some nested positions.

weswigham (approved):

sheetalkamat (approved):

PR #55556, opened 2023-08-29
Prevents auto import in module: "none"

### LINKED ISSUE #55256: typescript should never auto-import into files with `"module": "none"`
# Bug Report

I have a mixture of module and non-module code.
The problem occurs when I edit non-module files (tsconfig: `"module": "none"`).

Sometimes when using auto-completion (in VisualStudio 2022) typescript finds a "matching" name from the module world and adds a line like
```ts
import { HmiHelper } from './helper/HmiHelper.js';
```
into the first line of my file. 
But as the current file was configured as `"module": "none"` this `import` pushes typescript into modules mode and the current code is invalid.

### 🔎 Search Terms

intellisense
auto-import
modules

This bug is many years old and persists in Typescript 5.1.

hardikkoul:
Fixes #55256


andrewbranch:
Looks like the test is failing 😕

hardikkoul:
> Looks like the test is failing 😕

yeah , just saw that , checking it out.

hardikkoul:
Hey @andrewbranch , as i was not much familiar with the fourslash testing framework, so had to do some research and with some hit and tries the tests are passing now, Please review and verify once if some things need to be changed.

tests/cases/fourslash/autoImportNonModules.ts
```diff
@@ -0,0 +1,27 @@
...
+//// Thing/**/
+
+// @Filename: /src/dirA/thing2A.ts
+//// export class Thing2A {}
+
+verify.completions({
+  marker: "",
+  excludes: ["Thing2A"],
+  preferences: {
+    includeCompletionsForModuleExports: false
```
andrewbranch, line 25: This has to be true for it to be testing your code. This is the equivalent of just turning auto-imports off in VS Code, so it has nothing to do with `module: none`

andrewbranch (changes requested):

tests/cases/fourslash/autoImportNonModules.ts
hardikkoul: done

tests/cases/fourslash/autoImportNonModules.ts
```diff
@@ -3,24 +3,20 @@
...
 //// export class Thing1A {}
 //// Thing/**/
 
-// @Filename: /src/dirA/thing2A.ts
-//// export class Thing2A {}
-
 verify.completions({
   marker: "",
-  excludes: ["Thing2A"],
+  includes: ["!Thing1A"],
```
andrewbranch, line 19: 🤔 What is this?

hardikkoul: > 🤔 What is this?

haha just trying the same thing to exclude it by including it with "!" , read it somewhere if exclude is giving error try including with "!".

hardikkoul:
@andrewbranch , please review and assist me if more changes required as im not good at writing tests

hardikkoul:
> It turns out the behavior was already working as intended, but it was untested, so I’ve added two tests. `--module none` only errors when `--target` is below es2015, and auto-imports were already prevented in that case, and not in the others. (Like I said, `--module none` is pretty broken and possibly meaningless.) Thanks for your patience!

ohhh , thats a relief to know that it was working already, anyway im analyzing the test you wrote , really helpful :) thanks for your help !!

andrewbranch:
I would say, “working” here means auto-imports agrees with whether or not `--module none` allows imports to be written. Whether `--module none` itself is “working”... 🤷 There have been lot of ideas about what `--module none` _could_ mean if we wanted to try to make it more consistent:

- You're never allowed to write module constructs (what you assumed, I think)
- You're not allowed to write modules, but you can use dynamic `import()` calls
- You're not allowed to write modules, but `import type` imports would be allowed and wouldn't make the file a module
- Any module constructs you write are allowed, but never transformed to a particular output module format

I think potentially any of these might be on the table for the future, we just haven’t had a reason to pull the trigger on any of them. So the current state is... pretty weird.

hardikkoul:
> I would say, “working” here means auto-imports agrees with whether or not `--module none` allows imports to be written. Whether `--module none` itself is “working”... 🤷 There have been lot of ideas about what `--module none` _could_ mean if we wanted to try to make it more consistent:
> 
>     * You're never allowed to write module constructs (what you assumed, I think)
> 
>     * You're not allowed to write modules, but you can use dynamic `import()` calls
> 
>     * You're not allowed to write modules, but `import type` imports would be allowed and wouldn't make the file a module
> 
>     * Any module constructs you write are allowed, but never transformed to a particular output module format
> 
> 
> I think potentially any of these might be on the table for the future, we just haven’t had a reason to pull the trigger on any of them. So the current state is... pretty weird.

omg this is a lot, hahaha and yes i did indeed assume that its possible to write module constructs , current state as it seems is actually weird ngl , hahaha

hardikkoul:
@andrewbranch  , did i just contribute to TypeScript? :D , kind of i guess XD

HolgerJeromin:
> It turns out the behavior was already working as intended,

I do not understand. With typescript ts5.2.2 I still get bogus, file breaking autosuggestions in my `modules:none` code:
image

Was that changed after branching ts5.2?
At least THIS PR does not fix my #55256

andrewbranch:
If you accept the import, do you get an error?

PR #55558, opened 2023-08-29
Allow of in `await using` declarations in for-of loops

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/55555


DanielRosenwasser (approved):

tests/baselines/reference/awaitUsingDeclarationsInForAwaitOf.2.errors.txt
```diff
@@ -0,0 +1,15 @@
...
+awaitUsingDeclarationsInForAwaitOf.2.ts(4,32): error TS2448: Block-scoped variable 'of' used before its declaration.
+
+
+==== awaitUsingDeclarationsInForAwaitOf.2.ts (2 errors) ====
+    // https://github.com/microsoft/TypeScript/issues/55555
+    
+    {
+      for await (await using of of of) {};
+          ~~~~~
+!!! error TS1431: 'for await' loops are only allowed at the top level of a file when that file is a module, but this file has no imports or exports. Consider adding an empty 'export {}' to make this file a module.
```
rbuckton, line 11: Can we put this in an async context, like an `async function` so that we don't have this spurious error in `errors.txt`?

rbuckton (changes requested):

tests/baselines/reference/awaitUsingDeclarationsInForAwaitOf.2.errors.txt
Andarist: I just pushed out the requested change. Could you take another look at this?

rbuckton (approved):

jakebailey:
Note that 5.3 is cut; not sure if you intended for this to be in 5.3 or not.

rbuckton:
I did. This was a bugfix for a grammar error.

evanw:
With this change, `for (await using of x) ;` is [an error](https://www.typescriptlang.org/play?target=99&ts=5.3.1-rc#code/GYewTgBAFAhg7jAlgFwgVwM6IHYHMIjAQAeAlBANwBQApsQA7ioDeAvlUA) but `for await (await using of x) ;` is now [not an error](https://www.typescriptlang.org/play?target=99&ts=5.3.1-rc#code/GYewTgBAhg7lCWAXCAKWDkFcDO8B2A5hCMBAB4CUEA3AFACmZADuMgN4C+tQA) (a change in behavior from TS 5.2). In addition, the emit for `for await (await using of x) ;` has an extra space in it which implies to me that TypeScript's internal AST contains an empty variable declaration list, which would be invalid. Babel treats both as a syntax error (see [this](https://babel.dev/repl#?browsers=chrome%20111&build=&builtIns=false&corejs=3.21&spec=false&loose=false&code_lz=GYewTgBAFAhg7jAlgFwgVwM6IHYHMIjAQAeAlBANwBQQA&debug=false&forceAllTransforms=false&modules=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Creact%2Cstage-3%2Ctypescript&prettier=false&targets=&version=7.23.3&externalPlugins=%40babel%2Fplugin-syntax-explicit-resource-management%407.20.0&assumptions=%7B%7D) and [this](https://babel.dev/repl#?browsers=chrome%20111&build=&builtIns=false&corejs=3.21&spec=false&loose=false&code_lz=GYewTgBAhg7lCWAXCAKWDkFcDO8B2A5hCMBAB4CUEA3AFBA&debug=false&forceAllTransforms=false&modules=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Creact%2Cstage-3%2Ctypescript&prettier=false&targets=&version=7.23.3&externalPlugins=%40babel%2Fplugin-syntax-explicit-resource-management%407.20.0&assumptions=%7B%7D)), which also leads me to believe that this is a bug in TypeScript.

Was that an intentional aspect of this change or was that a regression?

Andarist:
@evanw thanks for this information - I'll take a look at this soon.

PR #55565, opened 2023-08-29
String#matchAll should return iterable of RegExpExecArray (fixes #36788)

### LINKED ISSUE #36788: String#matchAll should return iterable of RegExpExecArray instead of RegExpMatchArray
**TypeScript Version:**  3.7 (issue is with lib typing) 

**Search Terms:** regex, matchall, es2020 

**Code**

```ts
const matches = Array.from('xxx'.matchAll(/x/g));
matches[0].index; // TS reports index may be undefined, when it should always be defined
```

Currently `matchAll` returns an iterable of `RegExpMatchArray`. However, the `index` and `input` properties on `RegExpMatchArray` are optional because `String#match` returns a match object without those if the regex has the global flag (https://github.com/microsoft/TypeScript/issues/35157).

`matchAll` on the other hand doesn't have the same behavior, so it should use `RegExpExecArray` where both of those properties are non-optional (or maybe create a `RegExpMatchAllArray` type) 

``` js 
> Array.from('xxx'.matchAll(/x/g)) 
[
  [ 'x', index: 0, input: 'xxx', groups: undefined ],
  [ 'x', index: 1, input: 'xxx', groups: undefined ],
  [ 'x', index: 2, input: 'xxx', groups: undefined ]
]
> Array.from('xxx'.matchAll(/a/g)) 
[]
> Array.from('aaa'.matchAll(/x/g)) 
[]
``` 

**Expected behavior:**

`input` and `index` on the match objects are not optional 

**Actual behavior:**

TS reports that `input` and `index` may be optional 

**Playground Link:** (Can't use playground since lib cannot be configured) 

**Related Issues:**  https://github.com/microsoft/TypeScript/pull/30936 introduced `String#matchAll`

nihil-admirari:
Fixes #36788.

`matchAll` accepts only global `RegExp`s and thus, if it didn't throw, it always returns `RegExpExecArray` with non-optional `index` and `input`.

Reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/matchAll#return_value.


nihil-admirari:

jakebailey:
@typescript-bot test this
@typescript-bot test top100
@typescript-bot run dt

nihil-admirari:
@jakebailey, sorry, I'm new here. Where `string.prototype.matchall-tests.ts` is located?

jakebailey:
That's over in https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/string.prototype.matchall

This is somewhat an expected failure because that other package duplicated the type, so is going to need typesVersions to fix.

nihil-admirari:
I've made a pull request to [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/66581) but it's failing the tests and it'll continue to fail unless this pull request is merged.

DanielRosenwasser (approved):

DanielRosenwasser:
I think we can take this in post-beta. Thoughts @sandersn?

sandersn:
It seems quite safe for a change to the stdlib, since it's only in es2020 and above.
I'm OK with merging it if you're OK with shipping a point release to revert it if needed!

sandersn (approved):

Issue #55540, opened 2023-08-28
Tsserver doesn't discover all files on OpenBSD


nbonfils:
### 🔎 Search Terms

"tsserver openbsd", "file 'node_modules/@types/react/index.d.ts' is not a module.", "typescript file is not a module", "typescript file is not a module openbsd", "typescript TS2306 openbsd", "typescript TS2306 Eglot", "typescript TS2306 lsp", "tsserver Changing to watchFile", "typescript lsp Changing to watchFile", "typescript emacs eglot openbsd", "tsserver syslog Changing to watchFile", "eglot typescript problem", "create react app typescript lsp", "cannot find name 'document'. do you need to change your target library? try changing the 'lib' compiler option to include 'dom'."

### 🕗 Version & Regression Information

I was unable to test this on prior versions because I don't believe this to be an issue related to the version of typescript but simply an issue linked with the OS. And I found this issue as soon as I started using the [typescript-language-server](https://github.com/typescript-language-server/typescript-language-server).

## Case study: Create React App

To test a similar setup on both Linux and OpenBSD, I setup a barebone CRA project with typescript template, I just added the files in a git repo (so that emacs sees the whole project), otherwise there is no modification, just running `npx create-react-app ts-test --template typescript` should get you the exact same setup.

The code per say works and compile on both OSes, but when using the [typescript-language-server](https://github.com/typescript-language-server/typescript-language-server) on OpenBSD, the tsserver sends wrong diagnostics about this snipet.

```ts
import React from 'react';
import ReactDOM from 'react-dom/client';
import './index.css';
import App from './App';
import reportWebVitals from './reportWebVitals';

const root = ReactDOM.createRoot(
  document.getElementById('root') as HTMLElement
);
root.render(
  <React.StrictMode>
    <App />
  </React.StrictMode>
);

// If you want to start measuring performance in your app, pass a function
// to log results (for example: reportWebVitals(console.log))
// or send to an analytics endpoint. Learn more: https://bit.ly/CRA-vitals
reportWebVitals();
```

### 🙁 Actual behavior

So I've been trying to setup a typescript dev environment on my OpenBSD box with Emacs+Eglot and latest [typescript-language-server](https://github.com/typescript-language-server/typescript-language-server). (It may also be worth noting I use node v18.17.1)
So the server start and Emacs is able to communicate with it, without any issues, it even manages to send some diagnostics and display them.

## Problem

[...]

PR #55567, opened 2023-08-29
Always perform inode watching unless on Windows.

DanielRosenwasser:
Experiment motivated by #55540.


DanielRosenwasser:
@typescript-bot pack this

PR #55580, opened 2023-08-30
Update dprint to 0.41 and bump plugins

jakebailey:
This is done by:

```
$ dprint config update
$ dprint fmt
```

(This is super handy; you can also `dprint add` to add new plugins.)

Only one change was produced; @dsherret I don't think this is actually intentional, given the only thing that could have happened was that dprint's parser was updated to a new version? (I have no preference on this one, though I don't usually see dprint _removing_ a newline like this.)


src/services/utilities.ts
```diff
@@ -4042,14 +4042,12 @@ function getSymbolParentOrFail(symbol: Symbol) {
...
-                    const { expression } = d as any;
-                    return (inJS ? "[JS]" : "") + kind + (expression ? ` (expression: ${Debug.formatSyntaxKind(expression.kind)})` : "");
-                }).join(", ")
-            }.`,
+            `Declarations: ${symbol.declarations?.map(d => {
+                const kind = Debug.formatSyntaxKind(d.kind);
+                const inJS = isInJSFile(d);
+                const { expression } = d as any;
+                return (inJS ? "[JS]" : "") + kind + (expression ? ` (expression: ${Debug.formatSyntaxKind(expression.kind)})` : "");
+            }).join(", ")}.`,
```
dsherret, line 4050: I looked into it. Essentially swc changed its representation of optional chaining to be more like babel and I think that change in the AST caused this change where it will try to be more "inline". It seems in this scenario, changing the call expr to have an optional chain in it will now causes different behaviour unlike before. I will fix this tomorrow.

dsherret: By the way, something I've wanted to do for a while is have certain codebases in a test suite in order to catch regressions before a release. I haven't gotten around to it yet, but when I add it then I will add typescript's repo https://github.com/dprint/dprint-plugin-typescript/issues/13

dsherret: @jakebailey I fixed it now and did a release.

jakebailey: Thank you!!

sheetalkamat (approved):

PR #55576, opened 2023-08-30
Discriminate types with divergent read/write property types

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/55532#issuecomment-1694744665
fixes https://github.com/microsoft/TypeScript/issues/55566

The cases related to optional members and EOPT regressed from https://github.com/microsoft/TypeScript/pull/54777
The divergent accessors case has never worked but I feel like it should and this fixes it as well

cc @jakebailey

PR #54777, opened 2023-06-26
Display write type for property accesses in write locations

Andarist:
closes https://github.com/microsoft/TypeScript/issues/53305

PR #55585, opened 2023-08-31
Fixed element access expression writes for divergent write types

Andarist:
I noticed this problem yesterday when working on https://github.com/microsoft/TypeScript/pull/55576

The problem is that those two should always behave in the same way and they should use the same write type:
```ts
prop.foo = value
prop['foo'] = value
```

### DIFF
--- a/src/compiler/utilities.ts
+++ b/src/compiler/utilities.ts
@@ -7239,8 +7239,8 @@ export function isRightSideOfQualifiedNameOrPropertyAccess(node: Node) {
 
 /** @internal */
 export function isRightSideOfAccessExpression(node: Node) {
-    return isPropertyAccessExpression(node.parent) && node.parent.name === node
-        || isElementAccessExpression(node.parent) && node.parent.argumentExpression === node;
+    return !!node.parent && (isPropertyAccessExpression(node.parent) && node.parent.name === node
+        || isElementAccessExpression(node.parent) && node.parent.argumentExpression === node);
 }
 
 /** @internal */
--- a/tests/cases/compiler/divergentAccessorsTypes8.ts
+++ b/tests/cases/compiler/divergentAccessorsTypes8.ts
@@ -1,0 +1,153 @@
+// @strict: true
+// @lib: esnext, dom
+// @noEmit: true
+
+export {}
+
+interface Serializer {
+  set value(v: string | number | boolean);
+  get value(): string;
+}
+declare let box: Serializer;
+const v = box['value']
+box['value'] = true;
+box['value'] = 42;
+box['value'] = "hello";
+
+interface Element {
+    get style(): CSSStyleDeclaration;
+    set style(cssText: string);
+}
+
+declare const element: Element;
+element['style'] = "color: red";
+element['style'] = element.style;
+
+class One {
+  get prop1(): string {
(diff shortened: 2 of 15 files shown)


src/compiler/utilities.ts
```diff
@@ -7239,8 +7239,8 @@ export function isRightSideOfQualifiedNameOrPropertyAccess(node: Node) {
 
 /** @internal */
 export function isRightSideOfAccessExpression(node: Node) {
-    return isPropertyAccessExpression(node.parent) && node.parent.name === node
-        || isElementAccessExpression(node.parent) && node.parent.argumentExpression === node;
+    return !!node.parent && (isPropertyAccessExpression(node.parent) && node.parent.name === node
```
Andarist, line 7242: Other utils like this one do not have a guard like this so it looks a little bit off. However, `.parent` is actually nullable - a `SourceFile` doesn't have a parent. I could guard this before calling this function but personally, I prefer to have it here.

tests/cases/fourslash/quickInfoOnElementAccessInWriteLocation1.ts
Andarist, line 1: those tests are borrowed from https://github.com/microsoft/TypeScript/pull/54777 , just adjusted to use the element access expressions

tests/cases/compiler/divergentAccessorsTypes8.ts
Andarist, line 1: most of the test cases here (but not all) are borrowed from `tests/cases/compiler/divergentAccessorsTypes1-6.ts`

tests/cases/compiler/divergentAccessorsTypes8.ts
```diff
@@ -0,0 +1,153 @@
...
+  }
+  set prop4(s: string | boolean) {}
+}
+
+declare const u1: One | Two;
+
+u1['prop1'] = 42;
+u1['prop1'] = "hello";
+
+u1['prop2'] = 42;
```
Andarist, line 72: I think that this one should error. See the playground [here](https://www.typescriptlang.org/play?ts=5.3.0-dev.20230831#code/MYGwhgzhAEDyB2BTaBvAUNaBzRAXaADgE4D2BATABQCUAXNBLkQJbxaoabRF4CuR8aACIhAbk4BfThDyFSFShHqMWbaAB9o8XgFsARoiLVUUqWlCQYAFQDuJDphz5iZKnQZNW7dF258BwmKS0rIuCkoeqljGKKZoaAAmiBY80MAk8IzQGYj0CIjiOQB0YeTQALzQACzk4onJ4Knpmfi4dvS2JOJtJCXyZZU1dUkpyM1ZvPDMGXlIGtCd4pPT8H2uFdW1aEA)

I re-verified all of the tests here and everything here behaves in the same way as it behaves for regular property access expressions. So I think that this issue is a separate one and I will try to prepare a follow up PR to fix it.

Andarist:
I don't think this will be affected by this PR

jakebailey:
@typescript-bot test top200
@typescript-bot run dt

jakebailey:
@typescript-bot run dt

jakebailey (approved):

PR #55597, opened 2023-09-01
Experiment: make fetch a var, not a function

andrewbranch:
Testing whether this will break a lot of stuff if we do this for https://github.com/DefinitelyTyped/DefinitelyTyped/issues/60924


andrewbranch:
@typescript-bot run dt

jakebailey:
@typescript-bot test top200

andrewbranch:
I don’t think this will be necessary.

PR #55598, opened 2023-09-01
[experiment] Don't smuggle deprecated out of parser, use jsdoc

jakebailey:
This is potentially part of #52921; this eager evaluation of `@deprecated` is the only reason that JSDoc parsing couldn't be lazy, but it's possible that this may be slow.


jakebailey:

jakebailey:
Ouch.

PR #55600, opened 2023-09-01
Provide Syntax Checking for Regular Expressions

### LINKED ISSUE #3432: Compiler should error when encountering invalid regular expressions
In [`tests/cases/conformance/parser/ecmascript5/RegressionTests/parser579071.ts`](https://github.com/Microsoft/TypeScript/blob/master/tests/cases/conformance/parser/ecmascript5/RegressionTests/parser579071.ts) we have an invalid regex:

``` TypeScript
var x = /fo(o/;
```

This is not valid JavaScript, but we don't give any errors. It would be helpful to let users know when their regular expressions are invalid.

### LINKED ISSUE #54744: Provide Basic Syntax Check for Regular Expression Literals
**Edit:** Since someone is complaining about the failure to follow issue templates, I copied the content to https://github.com/microsoft/TypeScript/issues/3432#issuecomment-1604431899. Please continue the discussion in that issue.

<details><summary>Original content</summary>

Previously worked on #51837, I found that TypeScript gives almost no syntax errors for regular expressions. I would like to file a PR about it, thus I am opening this issue for easier tracking.

But before I work on the issue, I would like to hear some opinions from the Team. I am not sure, but I suppose the job of providing syntax check should not be shirked to linters right?

Something I would like to do are:
- Check for duplicated or unknown flags
- Check for unbalanced parentheses, which is the most common mistake people make
- Check for invalid escapes
  But this should be done only for RegExps with `u` or `v` flag, i.e. in `UnicodeMode`, that means if we encounter a `u` or `v` flag we will need to rescan the whole RegExp again (!!) (i.e. redoing what is done in the current `reScanSlashToken` method)
  And to check for invalid [`DecimalEscape`s and k`<GroupName>`s](https://tc39.es/ecma262/#prod-AtomEscape) we will also need to count the number of capture groups and record the names of all named capture groups along the way.

Am I doing too much or too less? I know doing too much may cause serious performace regressions (well, luckily regular expression literals are not that common compared with string literals). It should be better than doing nothing after all though.

</details>

graphemecluster:
Finally there!
---

This PR provides syntax checks to regular expressions. It can also be considered a follow-up of #51837 where escapes in RegExp are still not being addressed.
It introduces 36 new diagnostic messages for errors not previously covered.
Although [this comment](https://github.com/microsoft/TypeScript/issues/3432#issuecomment-1604513075) suggested that errors are better thrown regardless of flags to avoid rescanning, the parsing behavior of the new Unicode Sets (v) mode in character class is totally different from that of the non-v modes. So I just reuse the old code to quickly skip to the end of the regex for recording the flags beforehand.

Besides the [Set Notation](https://github.com/tc39/proposal-regexp-v-flag) support, the PR also includes the following Stage 3 proposals:
- [Duplicate Named Capturing Groups](https://github.com/tc39/proposal-duplicate-named-capturing-groups) (_Update: reached Stage 4 per the 101<sup>st</sup> meeting of TC39_)
- [Regular Expression Pattern Modifiers](https://github.com/tc39/proposal-regexp-modifiers)

~~Additionally, some test cases from [test262](https://github.com/tc39/test262) are added to make up both this PR and #51837.~~
Reverted due to excessive amount of files, but may be unreverted per request.

Fixes #3432 (Fixes #54744)

---
Follow-up Proposal
---

This PR provides a basis for making further changes to achieve additional type safety for the `RegExp` interface.
A part of them is given below:

```ts
// types.ts

export interface RegularExpressionLiteral extends LiteralExpression {
    readonly kind: SyntaxKind.RegularExpressionLiteral;
    readonly regExpBody: string;
    readonly regExpFlags: RegularExpressionFlags;
    readonly regExpFlagsText: string;
    readonly capturingGroups: RegularExpressionCapturingGroup[]; // capturingGroups[0] is the whole pattern
    readonly capturingGroupSpecifiers: Map<string, RegularExpressionCapturingGroup[]>; // Duplicate groups with the same name are allowed as long as they are mutually exclusive, thus an array
}

export interface RegularExpressionCapturingGroup extends ReadonlyTextRange {
    readonly pattern: string[]; // format similar to the first argument of a template tag function
    readonly isPossiblyUndefined: boolean;
}

// es5.d.ts

type CapturingGroupsArray = [string, ...(string | undefined)[]];
type NamedCapturingGroupsObject = { [name: string]: string | undefined } | undefined;

interface RegExp<
    CapturingGroups extends CapturingGroupsArray = CapturingGroupsArray,
    NamedCapturingGroups extends NamedCapturingGroupsObject = undefined
> {
    exec(string: string): RegExpExecArray<NamedCapturingGroups> & CapturingGroups | null;
    // ...
}

interface String {
    // ...

    match<
        CapturingGroups extends CapturingGroupsArray = CapturingGroupsArray,
        NamedCapturingGroups extends NamedCapturingGroupsObject = undefined
    >(regexp: string | RegExp<CapturingGroups, NamedCapturingGroups>): RegExpMatchArray | RegExpExecArray<NamedCapturingGroups> & CapturingGroups | null;

    replace(searchValue: string | RegExp, replaceValue: string): string; // untouched

    replace<
        CapturingGroups extends CapturingGroupsArray = CapturingGroupsArray,
        NamedCapturingGroups extends NamedCapturingGroupsObject = undefined
    >(
        searchValue: string | RegExp<CapturingGroups, NamedCapturingGroups>,
        replacer: (...capturingGroups: CapturingGroups, offset: number, input: string, groups: NamedCapturingGroups) => string
    ): string;

    // Buggy until GH#45972 is fixed
    // Seems like we should have a es2018.string.d.ts for the additional `groups` argument? But another overload is confusing to both the type checker and the users

    // Same for `replaceAll` in es2021.string.d.ts

    // ...
}

type RegExpMatchArray = [string, ...string[]];

interface RegExpExecArray<NamedCapturingGroups extends NamedCapturingGroupsObject = undefined> {
    index: number;
    input: string;
}

// es2018.regexp.d.ts

interface RegExpExecArray<NamedCapturingGroups extends NamedCapturingGroupsObject = undefined> {
    groups: NamedCapturingGroups;
}
```

### Usage

```ts
/^(?<date>\d{4}-\d{2}-\d{2})[Tt ](?<time>\d{2}:\d{2}:\d{2})$/u.exec("2024-01-01T00:00:00")
// Produces
{
    kind: SyntaxKind.RegularExpressionLiteral,
    regExpBody: "^(?<date>\\d{4}-\\d{2}-\\d{2})[Tt ](?<time>\\d{2}:\\d{2}:\\d{2})$",
    regExpFlags: RegularExpressionFlags.Unicode,
    regExpFlagsText: "u",
    capturingGroups: [
        { pos: 1, end: 60, pattern: ["", "-", "-", ":", ":", ""], isPossiblyUndefined: false },
        { pos: 2, end: 28, pattern: ["", "-", "-", ""], isPossiblyUndefined: false },
        { pos: 33, end: 59, pattern: ["", ":", ":", ""], isPossiblyUndefined: false },
    ],
    capturingGroupSpecifiers: Map {
        date: [{ pos: 2, end: 28, pattern: ["", "-", "-", ""], isPossiblyUndefined: false }],
        time: [{ pos: 33, end: 59, pattern: ["", ":", ":", ""], isPossiblyUndefined: false }],
    },
}
// Which is transformed to types in checker.ts, thus:
RegExp<
    /*CapturingGroups*/ [`${string}-${string}-${string}:${string}:${string}`, `${string}-${string}-${string}`, `${string}:${string}:${string}`],
    /*NamedCapturingGroups*/ {
        date: `${string}-${string}-${string}`;
        time: `${string}:${string}:${string}`;
    }
>.exec(string: string): {
    index: number;
    input: string;
    groups: {
        date: `${string}-${string}-${string}`;
        time: `${string}:${string}:${string}`;
    };
} & [`${string}-${string}-${string}:${string}:${string}`, `${string}-${string}-${string}`, `${string}:${string}:${string}`] | null;
```

No more runtime errors on `match.groups.foo.length` and `match[3].length`!


src/compiler/scanner.ts
```diff
@@ -2373,15 +2428,965 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
...
+                    case CharacterCodes.bar:
+                        pos++;
+                        return String.fromCharCode(ch);
+                    default:
+                        if (pos >= end || isLineBreak(ch)) {
+                            error(Diagnostics.Undetermined_character_escape, pos - 1, 1, ch);
+                            return "\\";
+                        }
+                        pos--;
+                        return scanEscapeSequence(/*shouldEmitInvalidEscapeError*/ unicodeMode, /*isRegularExpression*/ true);
```
graphemecluster, line 2796: I don't know if we should always emit those invalid escape errors despite of flags (though that's actually what were intended originally), since it seems like a bad practice to use things like octal escape sequences in character class (currently those outside character class are not fell back from capturing group references to escape sequences even in non-Unicode mode either).

graphemecluster:
The fix to self-check seems exactly like how it *should* be, but is too common to say anything about it 😅

src/compiler/diagnosticMessages.json
```diff
@@ -5294,7 +5442,7 @@
         "category": "Error",
         "code": 6258
     },
-      "Found 1 error in {0}": {
+    "Found 1 error in {0}": {
```
graphemecluster, line 5445: Curious why it isn't formatted previously. Shouldn't it be already formatted in the dprint PR?

src/compiler/scanner.ts
```diff
@@ -1066,15 +1122,8 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
...
-    function error(message: DiagnosticMessage, errPos: number = pos, length?: number, arg0?: any): void {
-        if (onError) {
-            const oldPos = pos;
-            pos = errPos;
-            onError(message, length || 0, arg0);
-            pos = oldPos;
-        }
+    function error(message: DiagnosticMessage, errPos = pos, length = 0, ...args: DiagnosticArguments): DiagnosticWithDetachedLocation | undefined {
+        return onError?.(message, errPos, length, ...args);
     }
```
graphemecluster, line 1127: I wonder if this refactor would cause serious compatibility issue, as what the TypeScript Bot warns. Nevertheless provided that an extra argument makes the code neat and tidy, setting the variable back and forth sounds ridiculous to me.

graphemecluster:
The fix to self-check looks exactly like how it *should* be, but is too common among the community to say anything about it 😅

src/compiler/scanner.ts
graphemecluster, line 1127: I wonder if this refactor would cause serious compatibility issue, as what the TypeScript Bot warns. Nevertheless provided that an extra argument makes the code neat and tidy, setting the variable back and forth sounds ridiculous to me.

graphemecluster:
I am not sure if there are conflicts between the licenses of each test262 test file and this repo. If that's the case, I will remove that commit.

src/compiler/diagnosticMessages.json
jakebailey: No, I haven't formatted the JSON yet; there are some downsides.

jakebailey:
> I am not sure if there are conflicts between the licenses of each test262 test file and this repo. If that's the case, I will remove that commit.

You just have to include the license, we have it in other places pretty sure. We don't distribute the tests so it's theoretically not a problem.

jakebailey:
@typescript-bot test top100
@typescript-bot run dt
@typescript-bot pack this

src/compiler/diagnosticMessages.json
jakebailey: Also, this formatting is unrelated to the PR; I would not include it.

src/compiler/types.ts
```diff
@@ -6909,10 +6924,8 @@ export interface Diagnostic extends DiagnosticRelatedInformation {
     /** @internal */ skippedOn?: keyof CompilerOptions;
 }
 
-/** @internal */
 export type DiagnosticArguments = (string | number)[];
```
jakebailey, line 6927: This stuff definitely should not become public.

src/compiler/utilities.ts
```diff
@@ -9788,7 +9788,7 @@ export function isValidBigIntString(s: string, roundTripOnly: boolean): boolean
     if (s === "") return false;
     const scanner = createScanner(ScriptTarget.ESNext, /*skipTrivia*/ false);
     let success = true;
-    scanner.setOnError(() => success = false);
+    scanner.setOnError(() => void (success = false));
```
jakebailey, line 9791: What's going on here?

src/compiler/utilities.ts
```diff
@@ -8205,7 +8205,7 @@ export function setObjectAllocator(alloc: ObjectAllocator) {
 
 /** @internal */
 export function formatStringFromArgs(text: string, args: DiagnosticArguments): string {
-    return text.replace(/{(\d+)}/g, (_match, index: string) => "" + Debug.checkDefined(args[+index]));
+    return text.replace(/\{(\d+)\}/g, (_match, index: string) => "" + Debug.checkDefined(args[+index]));
```
jakebailey, line 8208: Why does this have to change? Isn't this legal (but suspicious) syntax?

src/harness/fourslashImpl.ts
```diff
@@ -2699,7 +2699,7 @@ export class TestState {
                         },
                     });
                 }
-                else if (prevChar === " " && /A-Za-z_/.test(ch)) {
+                else if (prevChar === " " && /[A-Za-z_]/.test(ch)) {
```
jakebailey, line 2702: This is suspicious, yes, but isn't the old code legal?

jakebailey (commented):
I have no idea if this is something we actually want, but just leaving a couple of comments before I forget...

src/compiler/utilities.ts
graphemecluster: Yes, this special case is currently not handled because I consider it a bad practice to not escape the character despite being in non-Unicode mode, just like what we are doing so now for octal escape.
But as I just commented I understand that this case seems to be common in codebases. Yet, I don’t want to totally silence the error, since if someone writes /a{1,3/ it’s very likely that the user forgets the closing brace.
So, a compromise would be to error if there are digits directly following the opening brace. Does it sound right?

jakebailey: I feel like there are lint rules which are intended to catch this...

src/compiler/types.ts
graphemecluster: OK, let’s keep `any` for the `ErrorCallback` then

tests/baselines/reference/api/typescript.d.ts
```diff
@@ -8721,7 +8747,7 @@ declare namespace ts {
     function isIdentifierStart(ch: number, languageVersion: ScriptTarget | undefined): boolean;
     function isIdentifierPart(ch: number, languageVersion: ScriptTarget | undefined, identifierVariant?: LanguageVariant): boolean;
     function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean, languageVariant?: LanguageVariant, textInitial?: string, onError?: ErrorCallback, start?: number, length?: number): Scanner;
-    type ErrorCallback = (message: DiagnosticMessage, length: number, arg0?: any) => void;
+    type ErrorCallback = (message: DiagnosticMessage, pos: number, length: number, ...args: DiagnosticArguments) => DiagnosticWithDetachedLocation | undefined;
```
jakebailey, line 8750: I just noticed this; yeah we do not want to be breaking this API.

jakebailey: In general I don't think this change should contain any API changes at all.

src/compiler/diagnosticMessages.json
graphemecluster: Actually I didn’t notice that until I opened the PR. I will revert it. I don’t know why it’s being formatted on saving. You might want to review the dprint config.

jakebailey: The config works fine on my machine; we don't even have the JSON plugin enabled, so you may want to check what setting you have and if it matches the template.

src/compiler/utilities.ts
graphemecluster:
```ts
scanner.setOnError(() => {
    success = false;
});
```
is what I intended; writing this way was probably because both that I did a search of `void (` and found usage in the code, and I wanted to keep the original (inline) code style. I suppose y'all have a strong preference on the multiline style?

tests/baselines/reference/api/typescript.d.ts
graphemecluster: OK, I will revert it.

src/compiler/utilities.ts
graphemecluster: (Indeed, and this defeats the whole point of working on it…)

graphemecluster:
@jakebailey I just removed all the unwanted changes (including the unnecessary refactoring). Please ignore what you have previously reviewed since they are all reverted.

graphemecluster:
Actually I now think that the test cases from test262 itsn’t that helpful, and they just lengthen the time necessary for review.
I am not sure how much tolerance there is for changes unrelated to the goal of the PR (like whether I should include trivia mistakes discovered during working on it), but if that’s supposed to be low, the commit of adding test262 tests should probably be removed since a significant part of it is actually related to #51837 (or even outside of it).

tests/baselines/reference/api/typescript.d.ts
```diff
@@ -5445,6 +5445,19 @@ declare namespace ts {
     interface RegularExpressionLiteral extends LiteralExpression {
         readonly kind: SyntaxKind.RegularExpressionLiteral;
     }
+    enum RegularExpressionFlags {
```
jakebailey, line 5448: This enum is unused, which implies it shouldn't be public.

graphemecluster: That's because originally I actually intended to add a few properties to `RegularExpressionLiteral`:
```ts
export interface RegularExpressionLiteral extends LiteralExpression {
    readonly kind: SyntaxKind.RegularExpressionLiteral;
    readonly regExpBody: string;
    readonly regExpFlags: RegularExpressionFlags;
    readonly regExpFlagsText: string;
}
```
I guess I'll defer the enum exposure to the next PR about the follow-up proposal then. I will mark the enum internal later, before the PR is ready to ship.

DanielRosenwasser:
@typescript-bot test top200
@typescript-bot pack this

RyanCavanaugh:
It seems like a lot of the regexes this is flagging as wrong are fine (they at least execute without error), e.g. `/^[\w-\s.,\t\n]+$/` gets flagged due to the `\w-\s` sequence, but the engine handles that correctly (it behaves the same as `-\w\s`)

graphemecluster:
@RyanCavanaugh That's [Annex B](https://tc39.es/ecma262/#sec-patterns-static-semantics-early-errors-annexb), which means not all engines are happy with it, e.g. engine262. And that's misleading; I suppose this alone justify it.

graphemecluster:
I guess there’re too many files that makes one hesitant to check them due to time cost. If that’s the case I am going to remove the test cases from test262 since the test file I wrote already covers >90% of them (while conversely more than half of the cases in my file are not covered by them).
The remaining changes to the files are mostly due to combining two files with the same content with two different targets into one.
Anyway I will resolve the conflicts after the next review.

DanielRosenwasser:
@typescript-bot test top300
@typescript-bot pack this

src/compiler/scanner.ts
```diff
@@ -388,6 +416,22 @@ export function stringToToken(s: string): SyntaxKind | undefined {
...
+export function regularExpressionFlagToCharacter(f: RegularExpressionFlags): string | undefined {
+    return regExpFlagChars[f];
+}
+
+/** @internal */
+export function characterToRegularExpressionFlag(c: string): RegularExpressionFlags | undefined {
+    return charToRegExpFlag.get(c);
+}
+
+function regularExpressionFlagToFirstAvailableLanguageVersion(f: RegularExpressionFlags): ScriptTarget | undefined {
```
DanielRosenwasser, line 431: I would just inline usages of this function.

jakebailey: This pattern isn't actually new; see the token code above.

src/compiler/types.ts
```diff
@@ -7350,7 +7366,7 @@ export const enum ScriptKind {
 // NOTE: We must reevaluate the target for upcoming features when each successive TC39 edition is ratified in
 //       June of each year. This includes changes to `LanguageFeatureMinimumTarget`, `ScriptTarget`,
 //       transformers/esnext.ts, commandLineParser.ts, and the contents of each lib/esnext.*.d.ts file.
-export const enum ScriptTarget {
+export enum ScriptTarget {
```
jakebailey, line 7369: This is unrelated and should be reverted; the const enum here is a perf optimization when using tsc emit.

src/compiler/scanner.ts
```diff
@@ -2384,15 +2444,979 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
...
+                            return;
+                        default:
+                            scanSourceCharacter();
+                            isPreviousTermQuantifiable = true;
+                            break;
+                    }
+                }
+            }
+
+            function scanPatternModifiers(currFlags = RegularExpressionFlags.None): RegularExpressionFlags {
```
DanielRosenwasser, line 2709: There's only 2 usages of this function, don't use a default.

src/compiler/scanner.ts
DanielRosenwasser: That one is exported and used pretty frequently, right?

I haven't seen where `regularExpressionFlagToCharacter` and `characterToRegularExpressionFlag` are used yet, but at least this one is only used in this file.

src/compiler/types.ts
graphemecluster: This is required to get the name of the target for ts1501. Perhaps there’s a better way to do so?

jakebailey: Yeah, you definitely do not want to use the reverse mapping for anything except debugging in the TS repo. I'm not sure what we typically use to do this sort of message. I think we just hardcode them in the diagnostics but that doesn't sound fun. But we shouldn't use the reverse mapping.

src/compiler/scanner.ts
```diff
@@ -1580,42 +1626,58 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
             case CharacterCodes.paragraphSeparator:
                 return "";
             default:
+                if (isRegularExpression && (shouldEmitInvalidEscapeError || isIdentifierPart(ch, languageVersion))) {
+                    error(Diagnostics._0_is_not_a_valid_character_escape, pos - 2, 2, String.fromCharCode(ch));
```
DanielRosenwasser, line 1630: Is this going to cause issues if `ch` is an unprintable character?

src/compiler/scanner.ts
```diff
@@ -2343,8 +2399,12 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
 
     function reScanSlashToken(): SyntaxKind {
         if (token === SyntaxKind.SlashToken || token === SyntaxKind.SlashEqualsToken) {
+            // Quickly get to the end of regex such that we know the flags
             let p = tokenStart + 1;
             let inEscape = false;
+            // Although nested character class is allowed in Unicode Sets mode,
+            // an unescaped slash is nevertheless invalid even in character class in Unicode mode.
+            // Thus we can simply ignore nested character class in the first pass.
```
DanielRosenwasser, line 2407: I guess what's happening is you're giving up the opportunity for a better error in cases like this:

```
var x = /[[\x00-\xff]&&\/]/v;
```

Could you at least add an example into the comment?

DanielRosenwasser: It would be relatively easy to recover better if you just swapped `inCharacterClass` to be a stack counter.

src/compiler/scanner.ts
```diff
@@ -2384,15 +2444,979 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
...
+                            if (isUnterminated && !isInGroup) {
+                                // Assume what starting from the character to be outside of the regex
+                                return;
+                            }
+                            if (unicodeMode) {
+                                error(Diagnostics.Unexpected_0_Did_you_mean_to_escape_it_with_backslash, pos, 1, String.fromCharCode(ch));
+                            }
+                            pos++;
+                            isPreviousTermQuantifiable = true;
+                            break;
```
DanielRosenwasser, line 2697: Don't you need to make sure that these all match up with the matching opener?

src/compiler/scanner.ts
```diff
@@ -553,8 +597,13 @@ function isHexDigit(ch: number): boolean {
     return isDigit(ch) || ch >= CharacterCodes.A && ch <= CharacterCodes.F || ch >= CharacterCodes.a && ch <= CharacterCodes.f;
 }
 
-function isCodePoint(code: number): boolean {
-    return code <= 0x10FFFF;
+function isASCIILetter(ch: number): boolean {
```
DanielRosenwasser, line 600: Nit: `isAsciiCharacter` or `isASCIICharacter`

src/compiler/scanner.ts
```diff
@@ -2384,15 +2444,979 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
...
+        function scanRegularExpressionWorker(text: string, end: number, regExpFlags: RegularExpressionFlags, isUnterminated: boolean) {
+            const unicodeMode = !!(regExpFlags & RegularExpressionFlags.UnicodeMode);
+            const unicodeSetsMode = !!(regExpFlags & RegularExpressionFlags.UnicodeSets);
+            let mayContainStrings = false;
+
+            let numberOfCapturingGroups = 0;
+            const groupSpecifiers = new Set<string>();
+            const groupNameReferences: (TextRange & { name: string; })[] = [];
+            const decimalEscapes: (TextRange & { value: number; })[] = [];
+            const namedCapturingGroups: Set<string>[] = [];
```
DanielRosenwasser, line 2497:
```suggestion
            /** All named capture groups defined in the regex. */
            const groupSpecifiers = new Set<string>();
            /** All referenced to named capture groups in the regex. */
            const groupNameReferences: (TextRange & { name: string; })[] = [];
            /** All numeric backreferences within the regex. */
            const decimalEscapes: (TextRange & { value: number; })[] = [];
            /** A stack of scopes for named capture groups. */
            const namedCapturingGroups: Set<string>[] = [];
```

src/compiler/scanner.ts
graphemecluster: But this really just cover letters

src/compiler/scanner.ts
```diff
@@ -2384,15 +2444,979 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
...
+                    scanAlternative(isInGroup);
+                    namedCapturingGroups.pop();
+                    if (text.charCodeAt(pos) !== CharacterCodes.bar) {
+                        return;
+                    }
+                    pos++;
+                }
+            }
+
+            // Alternative ::= Term*
```
rbuckton, line 2512: I'd suggest that if we're going to include the grammar in a comment that we use the `grammarkdown`-format grammar from the spec itself as it makes it easier to compare against the spec.

src/compiler/scanner.ts
```diff
@@ -2384,15 +2444,979 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
...
+            let mayContainStrings = false;
+
+            let numberOfCapturingGroups = 0;
+            const groupSpecifiers = new Set<string>();
+            const groupNameReferences: (TextRange & { name: string; })[] = [];
+            const decimalEscapes: (TextRange & { value: number; })[] = [];
+            const namedCapturingGroups: Set<string>[] = [];
+
+            // Disjunction ::= Alternative ('|' Alternative)*
+            function scanDisjunction(isInGroup: boolean) {
```
rbuckton, line 2500: Do we need to nest these functions? All of these nested functions inside `scanRegularExpressionWorker` are closures and thus are essentially new objects allocated every time this function is called. These, along with `scanRegularExpressionWorker` itself should be lifted out of `reScanSlashToken()` if possible as it's a lot of unnecessary overhead.

src/compiler/parser.ts
```diff
@@ -2399,7 +2405,7 @@ namespace Parser {
         }
 
         // The user alternatively might have misspelled or forgotten to add a space after a common keyword.
-        const suggestion = getSpellingSuggestion(expressionText, viableKeywordSuggestions, n => n) ?? getSpaceSuggestion(expressionText);
+        const suggestion = getSpellingSuggestion(expressionText, viableKeywordSuggestions, identity) ?? getSpaceSuggestion(expressionText);
```
rbuckton, line 2408: This change seems to be unrelated to the rest of the PR.

rbuckton (changes requested):
I have some concerns about how the code is structured. It seems to put the responsibilities for parsing the regular expression grammar in the scanner, rather than in the parser where it belongs. 

If you're goal is to extend this capability further in the future, then we're eventually going to want to use AST nodes rather than a bespoke API associated with `RegularExpressionLiteral` so that we can eventually leverage other language service features, e.g., for renaming groups, etc. Keeping the actual parsing logic in the parser would facilitate that change.

Also, since `createScanner` is part of our public API, it's entirely likely that someone is using it purely to scan portions of a source file. Performing a full parse of a RegExp during scan seems like unnecessary overhead for those scenarios.

src/compiler/scanner.ts
```diff
@@ -2384,15 +2444,979 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
...
+                pos = p;
+            }
+            tokenStart = saveTokenPos;
+            tokenFlags = saveTokenFlags;
             tokenValue = text.substring(tokenStart, pos);
             token = SyntaxKind.RegularExpressionLiteral;
         }
         return token;
+
+        function scanRegularExpressionWorker(text: string, end: number, regExpFlags: RegularExpressionFlags, isUnterminated: boolean) {
```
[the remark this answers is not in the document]
rbuckton: I realize this is a big change, so we can potentially consider doing this in a follow-up PR.

src/compiler/parser.ts
```diff
@@ -2143,7 +2145,11 @@ namespace Parser {
         // Don't report another error if it would just be at the same position as the last error.
         const lastError = lastOrUndefined(parseDiagnostics);
         let result: DiagnosticWithDetachedLocation | undefined;
-        if (!lastError || start !== lastError.start) {
+        if (message.category === DiagnosticCategory.Message && lastError && start === lastError.start && length === lastError.length) {
```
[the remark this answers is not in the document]
graphemecluster: This is for providing spelling suggestion in `scanCharacterClassEscape`. Otherwise we will have to add another method to the `Scanner` interface other than `setOnError`

src/compiler/parser.ts
graphemecluster: This is because I did the same in `scanCharacterClassEscape`

src/compiler/scanner.ts
graphemecluster: Ah, OK. This is essentially `/[\p{ASCII}&&\p{Letter}]/v` though

src/compiler/types.ts
jakebailey: Per @sheetalkamat, it looks like the thing to do is:

```ts
forEachEntry(targetOptionDeclaration.type, (value, key) => value === getEmitScriptTarget(options) ? key : undefined)
```

It would be nice to pull that code out into a utility function as we have copied that in at least two places.

src/compiler/scanner.ts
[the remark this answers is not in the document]
graphemecluster: But [Ryan once said](https://github.com/microsoft/TypeScript/issues/3432#issuecomment-1612083581) he has some reservations about allocating a notable amount of objects due to memory regression

src/compiler/scanner.ts
```diff
@@ -2384,15 +2444,979 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
...
+                        error(Diagnostics.This_regular_expression_flag_is_only_available_when_targeting_0_or_later, p, 1, ScriptTarget[availableFrom]);
+                    }
+                }
                 p++;
             }
-            pos = p;
+            pos = tokenStart + 1;
+            const saveTokenPos = tokenStart;
+            const saveTokenFlags = tokenFlags;
+            scanRegularExpressionWorker(text.slice(0, endOfBody), endOfBody, regExpFlags, isUnterminated);
```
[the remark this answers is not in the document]
graphemecluster: I did this in order to prevent characters beyond the end of the regex from being accidentally grabbed. `text` without any slicing with adequate `pos <= endOfRegexBody` checks is actually fine.

src/compiler/scanner.ts
graphemecluster: I was just aiming to group the functions logically, otherwise we will need to prefix all functions with `scanRegExp` or move them to a new file instead. I’ll lift them out right before the PR is ready to ship for easier reviewing.

src/compiler/scanner.ts
graphemecluster: `scanAlternative` itself suggests that the opener is an open parenthesis.

src/compiler/scanner.ts
```diff
@@ -2384,15 +2444,979 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
...
+                            if (isInGroup) {
+                                return;
+                            }
+                        // falls through
+                        case CharacterCodes.closeBracket:
+                        case CharacterCodes.closeBrace:
+                            if (isUnterminated && !isInGroup) {
+                                // Assume what starting from the character to be outside of the regex
+                                return;
+                            }
```
graphemecluster, line 2691: @rbuckton
> Performing a full parse of a RegExp during scan seems like unnecessary overhead for those scenarios.

Unfortunately `scanRegularExpressionWorker` may change the end position of a regex due to this part. See the change in `src/testRunner/unittests/incrementalParser.ts` for an example.

(`!isInGroup` indicates we are currently at the outermost level, all open parentheses are closed)

src/compiler/scanner.ts
graphemecluster: I actually did some sort of “elaboration” of the spec such as inlining intermediate productions instead of merely copying, aiming to make it more easily understandable at a glance
Both have their own advantages I believe

src/compiler/scanner.ts
graphemecluster: I once tried it but I can’t remember the rationale as to why I didn’t adopt it. I just tried it again and it really seems to give better error messages.

graphemecluster: I know why now. `/[[]/` is a totally valid regex. If we did that it would break all expressions like `/[[]/.test("[")`

src/compiler/scanner.ts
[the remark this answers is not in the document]
rbuckton: I'd prefer the checks over the extra string slice as that can add up to a fair bit of overhead if you're parsing a bunch of regex literals near the end of a large file. We need to be doing bounds checks anyways, so I'd rather them be against `endOfBody` than `text.length`.

src/compiler/scanner.ts
graphemecluster: I’ll try to take this out independently when I move the worker to the parser.

src/compiler/scanner.ts
graphemecluster: For the example you gave, `[/[[\x00-\xff]&&/]/v` is already a valid JavaScript expression if `v` is defined.

src/compiler/types.ts
graphemecluster: I’ve done the refactor for you. (Better use reverse mapping though, I think)

graphemecluster: But then all the target names have become lowercase and `ES2015` is now `es6`. Are they desirable?

jakebailey: > I’ve done the refactor for you. (Better use reverse mapping though, I think)

The reverse mapping isn't guaranteed to be stable if two enum values end up with the same value, but maybe that's alright. In any case, we just can't turn this into a non-const enum without consequence. It may be possible to extract `formatEnum` out of `ts.Debug`, but that would require other changes and I think can come later.

> But then all the target names have become lowercase and ES2015 is now es6. Are they desirable?

I don't feel strongly, but @DanielRosenwasser might. This does theoretically match other diagnostics (e.g. see the `Parse empty options of --target.js` baseline). But, we're definitely inconsistent.

The `es6` thing is the worst bit, and that comes down to the order of the array. We don't ever refer to to it as `es6` in any diagnostic, only `es2015`, so that's not great. Perhaps we need to iterate the list in reverse. Or, just hardcode a switch case like `getDefaultLibFileName`.

src/compiler/scanner.ts
graphemecluster: P.S. IMO `isDigit` should be renamed to `isASCIIDigit` too

src/compiler/types.ts
graphemecluster: How about just swapping the order and changing the case in `targetOptionDeclaration`, and add `.toLowerCase()` to other usages as appropriate?

graphemecluster: Oh sorry, what I just meant to say was the `makeReverseMap` method in the scanner (I just realized this is a local function, perhaps we should move it to utilities)

jakebailey: If by `makeReverseMap`, you mean that something that reads the enum object itself to build the mapping, I wouldn't want to do that. We have `ts.Debug.formatEnum` which does that, but we don't use it for anything other than debugging or debug failures because it's not guaranteed to be human readable or consistent, just "good enough" for us to read when debugging or viewing stack traces.

DanielRosenwasser:

src/compiler/types.ts
```diff
@@ -7429,6 +7445,7 @@ export const enum ScriptTarget {
     ES2021 = 8,
     ES2022 = 9,
     ES2023 = 10,
+    ES2024 = 11,
```
jakebailey, line 7448: This actually shouldn't be added yet; 2024 isn't finalized. Anything that refers to ES2024 should either be removed or point to ESNext

src/compiler/types.ts
graphemecluster: Ah, I mean that on `targetOptionDeclaration.type`

src/compiler/types.ts
jakebailey: Pushed a change for this (we're trying to get this in for 5.5!)

jakebailey: Afterwards, we should move this enum to `LanguageFeatureMinimumTarget`, which is the canonical place to store language feature values so we know what to update when ratification happens.

DanielRosenwasser:
So talking to a few people on our team, we still want to do some refactoring here and possibly move the logic to the parser or another component. On top of that, there may be a noticeable perf impact from this PR; however, we want this in for TypeScript 5.5 beta, and we're going to try to optimize after merging. If there is a considerable overhead, we will have to evaluate if this feature justifies the cost; but for now we are eager to push ahead on this and will merge.

Thank you for the contribution and all the work you've put into this PR!

graphemecluster:
I have resolved some of the conversations since this is merged.
And a big thank you for recognizing my efforts on the PR. It really means a lot to me!
I am pleased to work on it further during my free time.

Issue #55437, opened 2023-08-20
Unreachable returns badly break return type inference


rotu:
### 🔎 Search Terms

unreachable any return

### 🕗 Version & Regression Information

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about `let`

### ⏯ Playground Link

https://www.typescriptlang.org/play?target=7&jsx=0&ts=5.2.0-beta&allowUnreachableCode=true#code/GYVwdgxgLglg9mABMAFASgN6IPTcQBwCc58BTQgGwE9FDThy6ATRAQwGdEwQBbAI3IAoAJAVSURAA8A3FMQBeRAEZZdKCEJIZI3IgBypSRIowwpREzilOYOBNbAG0WuI1IoVMolMM6kUgBcIpIKiABEwHBwYYIAvoKgkLAIiADm6Fi6PoykLBxsYFQiYhIycooqLuqaUtI6eAZGiCZmiADWMBQUnGpuiB5e2X4QgSK9NdrxidDwSAAWGTh4Q8xsnKyFY641KvX6hsam5h1dPdvunuZDpP5BouKIVLI0Farnj3XxQA

### 💻 Code

```ts
function f(){ // properly referred as number
	let x; x = 1; return x;
	// Next line does not affect return type inference:
	x = "foo"
}
function g(){ // inferred as any
	let x; x = 1; return x;
	// Next line kills return type inference:
	return x;
}
function h(){ // inferred as any
	return 1;
	// Next line kills return type inference:
	let y; y = 1; return y;
}
```

### 🙁 Actual behavior

The types are inferred as:

```ts
declare function f(): number;
declare function g(): any;
declare function h(): any;
```

The case of `f` demonstrates that return types based on assignment to mutable variables *can* work. The unreachable assignment to `x` after return has no effect on the return type.

In the case of `g`, the return type should be unaffected - `x` is already inferred to be `number` and an extra return does not change the type of `x`.

In the case of `h`, the return type should be unaffected as well. The second return can be inferred to be `number`.

### 🙂 Expected behavior

The types should be inferred as:
```ts
declare function f(): number;
declare function g(): number;
declare function h(): number;
```

Either unreachable return statements should not affect return type or `return` should not stop type inference.

### Additional information about the issue

possibly related to #40393
likely related to #26914

Disabling `allowUnreachableCode` does not alleviate these issues.

fatcerberus:
[...]

PR #55601, opened 2023-09-01
Avoid inferring return and yield types from unreachable statements

Andarist:
closes https://github.com/microsoft/TypeScript/issues/55437

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -36286,6 +36286,10 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         const nextTypes: Type[] = [];
         const isAsync = (getFunctionFlags(func) & FunctionFlags.Async) !== 0;
         forEachYieldExpression(func.body as Block, yieldExpression => {
+            const statement = findAncestor(yieldExpression, isStatement)!;
+            if (canHaveFlowNode(statement) && !statement.flowNode && !compilerOptions.allowUnreachableCode) {
+                return;
+            }
             const yieldExpressionType = yieldExpression.expression ? checkExpression(yieldExpression.expression, checkMode) : undefinedWideningType;
             pushIfUnique(yieldTypes, getYieldedTypeOfYieldExpression(yieldExpression, yieldExpressionType, anyType, isAsync));
             let nextType: Type | undefined;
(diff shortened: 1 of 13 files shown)


tannerlinsley:
I needed this just the other day.

jakebailey:
@typescript-bot test top200
@typescript-bot run dt
@typescript-bot pack this

DanielRosenwasser:
This doesn't really work for assertion functions right?

Can you add a few tests around that?

```ts
function throws(): never {
    throw new Error();
}

export function foo() {
    throws();
    return 42;
}

export function bar() {
    var x;
    x = 1;
    if (Math.random()) {
        throws();
        return x;
    }
    x = 2;
    return x;
}
```

Andarist:
Should I just codify the current results in the test suite or would you expect those cases to work differently? It would be possible to achieve different results for them, right? I'm not sure if that's totally desirable from your PoV though.

src/compiler/checker.ts
```diff
@@ -36286,6 +36286,10 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         const nextTypes: Type[] = [];
         const isAsync = (getFunctionFlags(func) & FunctionFlags.Async) !== 0;
         forEachYieldExpression(func.body as Block, yieldExpression => {
+            const statement = findAncestor(yieldExpression, isStatement)!;
+            if (canHaveFlowNode(statement) && !statement.flowNode && !compilerOptions.allowUnreachableCode) {
```
jakebailey, line 36290: Out of curiousity, why flag this on `allowUnreachableCode`? My impression is that `allowUnreachableCode` is more like a lint rule that errors on the unreachable code, and not something that should impact type analysis.

Andarist: I honestly have no idea cause I've written it half a year ago 😅 

If the tests would come roughly unchanged without this condition (which is likely) then maybe I just have tried to be conservative when it comes to the impact of this change. Taking a look at how `allowUnreachableCode` is defined in the docs now - you are right that this flag should mostly just affect error reporting.

jakebailey:
@typescript-bot test it
@typescript-bot pack this

Andarist:
The above break is expected. This is the `Thrower`:
```ts
      function Thrower() {
        throw new Promise(() => {
          // Never resolve
        })

        // @ts-expect-error
        return null
      }
```

When considering the **unreachable** return statement, its type becomes `() => null` and that's a valid JSX component function. With this change, it's now `() => void` and that's not a valid component function. 

Function declarations and method declarations outside of object literals are not allowed to infer `never` as their return type (see `mayReturnNever` and its callers). So everything checks out. I'll fix the code in that repository - the usage of `// @ts-expect-error` already shows this is trying to work around things. The proper fix is to just annotate the return type with `never` and remove that unreachable code.

PR #55602, opened 2023-09-01
Run dprint on json

jakebailey:
Followup to the dprint change; I'd like to do this though one day I do hope that https://github.com/dprint/dprint-plugin-json/issues/10 and https://github.com/dprint/dprint-plugin-json/issues/14 are solved (but it's hard to say when `.json` files can also be treated as jsonc).


andrewbranch (approved):

src/compiler/diagnosticMessages.json
```diff
@@ -5294,7 +5294,7 @@
         "category": "Error",
         "code": 6258
     },
-      "Found 1 error in {0}": {
```
sandersn, line 5297: how did only 4 lines change here?

sandersn (approved):
This is fine with me. Are there good reasons not to do it?

jakebailey:
> This is fine with me. Are there good reasons not to do it?

The only reservation I have is about it removing trailing commas in files which support them, but I've been using this for a pretty long time in my own projects without much pain anyway. So, I think no.

PR #55604, opened 2023-09-01
minor changes on readme

dan-arnaiz:
Fixes #


jakebailey:
Thanks for the PR, but I don't these are actually improvements. (Also note that we don't accept PRs for small wording tweaks in general.)

PR #55605, opened 2023-09-01
Add option to exclude library symbols from navTo results

### LINKED ISSUE #48194: Add option for navto to exclude built-in libs and files from node_modules

gabritto:
Fixes #48194.

I'll follow up with an equivalent setting in vscode.

### DIFF
--- a/src/compiler/types.ts
+++ b/src/compiler/types.ts
@@ -9936,6 +9936,7 @@ export interface UserPreferences {
     readonly organizeImportsNumericCollation?: boolean;
     readonly organizeImportsAccentCollation?: boolean;
     readonly organizeImportsCaseFirst?: "upper" | "lower" | false;
+    readonly excludeExternalFileSymbols?: boolean;
 }
 
 /** Represents a bigint literal value without requiring bigint support */
--- a/src/compiler/types.ts
+++ b/src/compiler/types.ts
@@ -9936,7 +9936,7 @@ export interface UserPreferences {
     readonly organizeImportsNumericCollation?: boolean;
     readonly organizeImportsAccentCollation?: boolean;
     readonly organizeImportsCaseFirst?: "upper" | "lower" | false;
-    readonly excludeExternalFileSymbols?: boolean;
+    readonly excludeLibrarySymbols?: boolean;
 }
 
 /** Represents a bigint literal value without requiring bigint support */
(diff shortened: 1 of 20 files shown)


src/harness/client.ts
```diff
@@ -579,14 +589,13 @@ export class SessionClient implements LanguageService {
             const providePrefixAndSuffixTextForRename = typeof preferences === "boolean" ? preferences : preferences?.providePrefixAndSuffixTextForRename;
             const quotePreference = typeof preferences === "boolean" ? undefined : preferences?.quotePreference;
             if (providePrefixAndSuffixTextForRename !== undefined || quotePreference !== undefined) {
+                const oldPreferences = this.preferences;
```
gabritto, line 592: The other diffs in the file were just a drive-by fix to how we restore preferences for the harness client. I think we don't have tests that observe this though.

src/compiler/types.ts
```diff
@@ -9936,6 +9936,7 @@ export interface UserPreferences {
     readonly organizeImportsNumericCollation?: boolean;
     readonly organizeImportsAccentCollation?: boolean;
     readonly organizeImportsCaseFirst?: "upper" | "lower" | false;
+    readonly excludeLibrarySymbols?: boolean;
```
andrewbranch, line 9939: I think the name of this setting needs to mention the command/feature it affects.

src/harness/client.ts
```diff
@@ -314,15 +314,25 @@ export class SessionClient implements LanguageService {
         return notImplemented();
     }
 
-    getNavigateToItems(searchValue: string): NavigateToItem[] {
+    getNavigateToItems(searchValue: string, maxResultCount: number, file: string | undefined, _excludeDtsFiles: boolean | undefined, excludeLibFiles: boolean | undefined): NavigateToItem[] {
         const args: protocol.NavtoRequestArgs = {
             searchValue,
-            file: this.host.getScriptFileNames()[0],
+            file,
+            currentFileOnly: !!file,
```
andrewbranch, line 321: One thought: if I have `excludeLibrarySymbols` enabled, but then I manually open `node_modules/typescript/lib/lib.dom.d.ts` and use ⌘⇧O (symbol search in current file), I still want to see results from the current file.

andrewbranch (approved):

PR #55613, opened 2023-09-03
Fix issue with optional chaining and type inference in type guard

### LINKED ISSUE #34974: Optional chaining, proper infer in type guard
**TypeScript Version:**  3.7.0

**Search Terms:** optional chaining

**Code**

```ts
type X = {
  y?: {
    z?: string
  }
}
const x: X = {
  y: {
  }
}
// type guard
function isNotNull<A>(x: A): x is NonNullable<A> {
  return x!== null && x!== undefined;
}
// function which I want to call in the result of the expression
function title(str: string) {
  return str.length > 0 ? 'Dear ' + str : 'Dear nobody';
}

isNotNull(x?.y?.z) ? title(x.y.z) : null // type error - object is possibly undefined
```
The code fix is by adding additional checks or additional temporary variables:
```ts
(x?.y?.z && isNotNull(x.y.z)) ? title(x.y.z) : null
// or
const tmp = x?.y?.z
isNotNull(tmp) ? title(tmp) : null
```
**Expected behavior:**
TypeScript is able to infer that optional chaining was used as an argument, what means that typeguard is checking the whole chain.

**Actual behavior:**
TypeScript needs a code change in order to understand that optional chaining already checked other values from being null | undefined.

Zzzen:
Fixes #34974

### DIFF
--- a/tests/cases/compiler/typePredicatesOptionalChaining2.ts
+++ b/tests/cases/compiler/typePredicatesOptionalChaining2.ts
@@ -1,0 +1,15 @@
+// @strict: true
+
+type Person = { name: string; }
+
+const getName1 = (person?: Person): string => {
+  return typeof person?.name === 'string' ? person?.name : '';
+};
+
+const isString = (value: any): value is string => {
+  return typeof value === 'string';
+};
+
+const getName2 = (person?: Person): string => {
+  return isString(person?.name) ? person?.name : '';
+};
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -26109,7 +26109,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     function hasMatchingArgument(expression: CallExpression | NewExpression, reference: Node) {
         if (expression.arguments) {
             for (const argument of expression.arguments) {
-                if (isOrContainsMatchingReference(reference, argument)) {
+                if (isOrContainsMatchingReference(reference, argument) || optionalChainContainsReference(argument, reference)) {
                     return true;
                 }
             }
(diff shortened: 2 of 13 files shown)


jakebailey:
@typescript-bot test top100
@typescript-bot run dt

tests/cases/compiler/typePredicatesOptionalChaining2.ts
```diff
@@ -0,0 +1,15 @@
...
+const getName1 = (person?: Person): string => {
+  return typeof person?.name === 'string' ? person?.name : '';
+};
+
+const isString = (value: any): value is string => {
+  return typeof value === 'string';
+};
+
+const getName2 = (person?: Person): string => {
+  return isString(person?.name) ? person?.name : '';
```
Andarist, line 14: Shouldn't this be changed to this?
```suggestion
  return isString(person?.name) ? person.name : '';
```

tests/cases/compiler/typePredicatesOptionalChaining2.ts
```diff
@@ -0,0 +1,15 @@
+// @strict: true
+
+type Person = { name: string; }
+
+const getName1 = (person?: Person): string => {
+  return typeof person?.name === 'string' ? person?.name : '';
```
Andarist, line 6: similarly here:
```suggestion
  return typeof person?.name === 'string' ? person.name : '';
```

tests/cases/compiler/typePredicatesOptionalChaining2.ts
Zzzen: Both should be fine. If I recall correctly, optional chaining does not alter the type of the expression it is applied to.

Andarist: Ye, right - i thought that previously the error was also raised within the truthy branch of the conditional expression but it seems that this part was already working ok.

fatcerberus:
It must be getting too close to spooky season because I read the issue title as "Fix issue with ... type *gourd*" 🎃

weswigham (approved):

Issue #55574, opened 2023-08-30
Displayed parameter types when invoking a union of functions are inconsistent with behavior


ssalbdivad:
### 🔎 Search Terms

function parameter display signature helper inconsistent union

### 🕗 Version & Regression Information

This is the behavior in every version I tried

### ⏯ Playground Link

https://www.typescriptlang.org/play?ssl=27&ssc=1&pln=1&pc=1#code/CYUwxgNghgTiAEYD2A7AzgF3kgDhglqlBGgFwBQ8V8APvABT1QD8p8A3vFGxjAK4IAvgEp4AXgB88PigDWKJAHcUwytTqMARqw7xNPfkNGTpchctXlQkWAmTosuAkQgARfGhzQAnm3pqqWABzAAY-Tm54XgF4QXgAMl19KMNY0ToZUAAzfBQQYHJjKRl5JRRycgB6SvgAFQALD3g0fCCUKAw+OHgQGBgkGDR4RXqQFHgFLBwoNDR8rnHg8idCdrcPLyhvektq+HdPH2bW9s7u+lzkPvAMVT2VlzQmGFDwrgMBAG5YhKSPkG+IlopmyuXywjYJXM5T2ACEQPUoAA3QgwBhXOBgW5VGoPNZPYIhHT0CL-QG-TjJaIAtKQsxlZZ4VbEJ6WazQbr2TDwOCYNgBYGMFhsUkpGKCAA08AAdLLeRg2JgYLkggBtAC6RVMpQsAo09G0Ir0-zS4mK9N1OP2GyOLTaHS6CAuKAxNylk3wYAQ8vgTUmXCxfGIEG8XB5IG5IAgIAAtmNsXtnq8GKLqZ84olKWTTRkUKC8sApfLFbwVRqtSVUMorfDESiBuiBpiEzUk0S3pE0xm-mKaSIpbLpcXmqWUGrNWbtVXyvKSe9e1KqalJfAAERZJBIVdS1eaWCr1RAA

### 💻 Code

```ts
declare const optionals:
    | ((a?: { a: true }) => unknown)
    | ((b?: { b: true }) => unknown)

declare const optionalDisplay: (
    arg0: ({ a: true } & { b: true }) | undefined
) => unknown

// This signature errors when not passed an arg
optionalDisplay()

// Display signature (incorrect)
// optionals(arg0: ({ a: true; } & { b: true; }) | undefined): unknown
// Behavior (correct)
[...]

PR #55625, opened 2023-09-04
Fixed display of optional and rest parameters created as result of combining union parameters

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/55574

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -13201,12 +13201,13 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             const paramSymbol = createSymbol(
                 SymbolFlags.FunctionScopedVariable | (isOptional && !isRestParam ? SymbolFlags.Optional : 0),
                 paramName || `arg${i}` as __String,
+                isRestParam ? CheckFlags.RestParameter : isOptional ? CheckFlags.OptionalParameter : 0,
             );
             paramSymbol.links.type = isRestParam ? createArrayType(unionParamType) : unionParamType;
             params[i] = paramSymbol;
         }
         if (needsExtraRestElement) {
-            const restParamSymbol = createSymbol(SymbolFlags.FunctionScopedVariable, "args" as __String);
+            const restParamSymbol = createSymbol(SymbolFlags.FunctionScopedVariable, "args" as __String, CheckFlags.RestParameter);
             restParamSymbol.links.type = createArrayType(getTypeAtPosition(shorter, longestCount));
             if (shorter === right) {
                 restParamSymbol.links.type = instantiateType(restParamSymbol.links.type, mapper);
--- a/tests/cases/fourslash/completionEntryForUnionMethod.ts
+++ b/tests/cases/fourslash/completionEntryForUnionMethod.ts
@@ -3,7 +3,7 @@
 ////var y: Array<string>|Array<number>;
 ////y.map/**/(
 
-const text = "(method) Array<T>.map<unknown>(callbackfn: ((value: string, index: number, array: string[]) => unknown) & ((value: number, index: number, array: number[]) => unknown), thisArg: any): unknown[]";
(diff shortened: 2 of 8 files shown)


tests/cases/fourslash/completionEntryForUnionMethod.ts
```diff
@@ -3,7 +3,7 @@
 ////var y: Array<string>|Array<number>;
 ////y.map/**/(
 
-const text = "(method) Array<T>.map<unknown>(callbackfn: ((value: string, index: number, array: string[]) => unknown) & ((value: number, index: number, array: number[]) => unknown), thisArg: any): unknown[]";
+const text = "(method) Array<T>.map<unknown>(callbackfn: ((value: string, index: number, array: string[]) => unknown) & ((value: number, index: number, array: number[]) => unknown), thisArg?: any): unknown[]";
```
Andarist, line 6: From what I can tell - this is the correct change. It matches the display of the same method accessed on a single array (this one accesses it on a union of arrays):
```ts
var y: Array<string>
y.map(
// (method) Array<string>.map<unknown>(callbackfn: (value: string, index: number, array: string[]) => unknown, thisArg?: any): unknown[]
```

src/compiler/checker.ts
```diff
@@ -13201,12 +13201,13 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
                 SymbolFlags.FunctionScopedVariable | (isOptional && !isRestParam ? SymbolFlags.Optional : 0),
                 paramName || `arg${i}` as __String,
+                isRestParam ? CheckFlags.RestParameter : isOptional ? CheckFlags.OptionalParameter : 0,
             );
             paramSymbol.links.type = isRestParam ? createArrayType(unionParamType) : unionParamType;
             params[i] = paramSymbol;
         }
         if (needsExtraRestElement) {
-            const restParamSymbol = createSymbol(SymbolFlags.FunctionScopedVariable, "args" as __String);
+            const restParamSymbol = createSymbol(SymbolFlags.FunctionScopedVariable, "args" as __String, CheckFlags.RestParameter);
```
Andarist, line 13210: to give context: those are later checked by `symbolToParameterDeclaration` here: https://github.dev/microsoft/TypeScript/blob/5d04803196899fab535103a5114bc9bf8d04dbd9/src/compiler/checker.ts#L7615-L7618

src/compiler/checker.ts
```diff
@@ -13201,12 +13201,13 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             const paramSymbol = createSymbol(
                 SymbolFlags.FunctionScopedVariable | (isOptional && !isRestParam ? SymbolFlags.Optional : 0),
                 paramName || `arg${i}` as __String,
+                isRestParam ? CheckFlags.RestParameter : isOptional ? CheckFlags.OptionalParameter : 0,
```
jakebailey, line 13204: `0` should probably be `undefined` or `CheckFlags.None`.

jakebailey:
@typescript-bot test top100
@typescript-bot run dt

src/compiler/checker.ts
Andarist: I followed the existing "convention" here with that `0` :p the first argument here is passed in this very way

jakebailey: Er, well, that's `SymbolFlags`, but yeah that also doesn't use `None` for some reason, boo

Andarist: I can adjust both places if you'd prefer that

jakebailey: Eh, it's fine; commenting out `.None` shows 5 other places we used zero and none where we used `None`.

jakebailey (approved):
This seems correct to me.

sandersn:
@jakebailey is this ready to merge now?

jakebailey:
I'd say so but I typically don't merge PRs that aren't assigned to me just in case 😅

weswigham (approved):

PR #55638, opened 2023-09-05
Track recursive homomorphic mapped types by the symbol of their target

### LINKED ISSUE #55535: Assignability false positive for recursive mapped types, even when fully instantiated
### 🔎 Search Terms

recursive mapped types, assignability, depth check, deeply nested

If this is in fact a bug, it looks like it was introduced in v4.5

- This can cause a runtime error when using incompatible values whose types are derived from recursive mapped types with more than 2 levels of nesting because the compiler seems to consider them equal (even when the derived type is fully instantiated (not infinite)).
- This changed between versions v4.4 and v4.5

### ⏯ Playground Link

https://tsplay.dev/WoM1gN

### 💻 Code

```typescript
type Id<T> = { [K in keyof T]: Id<T[K]> }

type HasNumeric = Id<{ x: { y: { z: number } } }>
//   ^? type HasNumeric = { x: { y: { z: number } } }

type HasString = Id<{ x: { y: { z: string } } }>
//   ^? type HasString = { x: { y: { z: string } } }

const toDollarsAndCents = (_: HasNumeric): string => _.x.y.z.toFixed(2)

declare const illegalValue: HasString
//            ^? const illegalValue:  { x: { y: { z: string } } }
 
/** 😵 */
toDollarsAndCents(illegalValue) 
```

### 🙁 Actual behavior

I passed a value known by the compiler to be of type `{ x: { y: { z: string } } }` to a function expecting `{ x: { y: { z: number } } }` and did not get a compile time error warning me about it.

### 🙂 Expected behavior

I expect to get a compile time error when I pass a value known by the compiler to be of type `{ x: { y: { z: string } } }` to a function expecting `{ x: { y: { z: number } } }`

Hey team! First of all, thank you for your work on this project. I have used TypeScript every day for years, and since the release of v5 in particular, saw a noticeable increase in my productivity at work :)

Regarding this issue, after further testing, it seems pretty trivial to reproduce with any recursive mapped type more than 2 levels deep (even with a "no-op" transformation, as is the case with `Id`).

Even if the type were not fully instantiated, the assignability rules in this case seem to work exactly backwards from what I'd expect, which makes me think this might not be the intended behavior.

This issue might be related to #46599, which fits given that the playground's type-checker [does complain in v4.4](https://tsplay.dev/mqyqrm)

Let me know if there's anything you'd like me to elaborate on!

ahejlsberg:
With this PR, when a homomorphic mapped type is applied to a type with a symbol, we use the symbol of that type as the recursion identity. This is a better strategy than using the symbol of the mapped type, which doesn't work well for recursive mapped types because they end up being considered deeply nested after just three levels.

```ts
type Id<T> = { [K in keyof T]: Id<T[K]> };

type Foo1 = Id<{ x: { y: { z: { a: { b: { c: number } } } } } }>;
type Foo2 = Id<{ x: { y: { z: { a: { b: { c: string } } } } } }>;

declare const foo1: Foo1;
const foo2: Foo2 = foo1;  // Now errors, previously didn't
```

Note that this PR isn't a general solution to the problem of distinguishing finite and infinite types, rather it is a targeted fix for situations where mapped types are applied to object types that clearly originate in distinct declarations.

Fixes #55535.

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -23484,10 +23484,18 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 // unique AST node.
                 return (type as TypeReference).node!;
             }
-            if (type.symbol && !(getObjectFlags(type) & ObjectFlags.Anonymous && type.symbol.flags & SymbolFlags.Class)) {
-                // We track all object types that have an associated symbol (representing the origin of the type), but
-                // exclude the static side of classes from this check since it shares its symbol with the instance side.
-                return type.symbol;
+            if (type.symbol) {
+                // We track object types that have a symbol by that symbol (representing the origin of the type).
+                if (getObjectFlags(type) & ObjectFlags.Mapped) {
+                    // When a homomorphic mapped type is applied to a type with a symbol, we use the symbol of that
+                    // type as the recursion identity. This is a better strategy than using the symbol of the mapped
+                    // type, which doesn't work well for recursive mapped types.
+                    type = getMappedTargetWithSymbol(type);
+                }
+                if (!(getObjectFlags(type) & ObjectFlags.Anonymous && type.symbol.flags & SymbolFlags.Class)) {
+                    // We exclude the static side of a class since it shares its symbol with the instance side.
+                    return type.symbol;
+                }
             }
             if (isTupleType(type)) {
                 return type.target;
@@ -23509,6 +23517,14 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             return (type as ConditionalType).root;
         }
         return type;
+    }
+
+    function getMappedTargetWithSymbol(type: Type) {
+        let target = type;
+        while ((getObjectFlags(target) & ObjectFlags.InstantiatedMapped) === ObjectFlags.InstantiatedMapped && isMappedTypeWithKeyofConstraintDeclaration(target as MappedType)) {
+            target = getModifiersTypeFromMappedType(target as MappedType);
+        }
+        return target.symbol ? target : type;
     }
 
     function isPropertyIdenticalTo(sourceProp: Symbol, targetProp: Symbol): boolean {
--- a/src/compiler/types.ts
+++ b/src/compiler/types.ts
@@ -6204,6 +6204,8 @@ export const enum ObjectFlags {
     RequiresWidening = ContainsWideningType | ContainsObjectOrArrayLiteral,
     /** @internal */
     PropagatingFlags = ContainsWideningType | ContainsObjectOrArrayLiteral | NonInferrableType,
+    /** @internal */
+    InstantiatedMapped = Mapped | Instantiated,
     // Object flags that uniquely identify the kind of ObjectType
     /** @internal */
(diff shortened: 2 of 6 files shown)


ahejlsberg:
@typescript-bot test this
@typescript-bot user test this inline
@typescript-bot run dt
@typescript-bot test top100

Andarist:
This is awesome, thank you! ❤️

ssalbdivad:
This is massive, thanks so much for the quick turnaround 😍

ahrjarrett:
This looks great. Did not expect to see this picked up so quickly :)

Thank you!

weswigham (approved):
Feels maybe a bit odd that `Readonly<Foo>`, `Partial<Foo>`, and `WhateverMapper<Foo>` all have the same recursion identity, but it's probably better than every `Partial` having the same recursion identity. Feels like it'd be better to use both the target symbol and the modifiers type symbol as an aggregate identity for mapped types like this, if possible - that way if you have mutually recursive mapped types mapping over the same type (eg, you have a self referential type that is deep mapped in a way that alternates between applying readonly and partial or the like), they all get the full 3-repition depth.

ahejlsberg:
The DT run has new errors in the `raphael` package. They're caused by deeper exploration of mapped types that now occurs because we no longer cut off recursion after three levels of nesting of *any* instantiation of a particular mapped type. I did a bit of experimenting, and it looks like the issue can be remedied by placing an `in out` (i.e. invariant) annotation on the type parameter of the `RaphaelStatic` interface:

```
export interface RaphaelStatic<
    in out TTechnology extends RaphaelTechnology = "SVG" | "VML"
    > {
        // ...
}
```

Also, two new errors in the `vscode` project. These look like true type mismatches that previously weren't discovered because we only checked three levels of nesting.

jakebailey:
`in out` can't really be used on DT without duplicating the entire package to support old versions; DT still works back to TS 4.4.

weswigham:
@typescript-bot pack this

DanielRosenwasser:
One thing from the design meeting notes - include a test case around something like

```ts
type Expand<T> = {
  [K in keyof T]: Expand<Expand<T[K]>>
};
```

ahejlsberg:
Ok, a better fix for the new `raphael` errors is to modify the recursive `RaphaelPaperPluginRegistry` type to:

```ts
export type RaphaelPaperPluginRegistry<
    TTechnology extends RaphaelTechnology = "SVG" | "VML",
    T = RaphaelPaper<TTechnology>> = {
        /**
         * Either the paper plugin method or a new namespace with methods.
         */
        [P in keyof T]: RaphaelPaperPluginMethodOrRegistry<TTechnology, T[P]>;
    };

type RaphaelPaperPluginMethodOrRegistry<TTechnology extends RaphaelTechnology, T> =
    T extends (...args: any) => any
        ? RaphaelPaperPluginMethod<TTechnology, Parameters<T>, ReturnType<T>>
        : RaphaelPaperPluginRegistry<TTechnology, T>;
```

Extracting the conditional type into a separate aliased type allows the compiler to compute variance information for the type arguments, which in turns makes it much cheaper to relate instantiations. This change is backwards compatible and has positive impact on type instantiation counts even before this PR.

jakebailey:
Is there anything blocking this? Just, sending the above to DT?

PR #55655, opened 2023-09-06
remove typescript

DarrenBaldwin07:
"Fully recognize that TypeScript offers some people some advantages, but to my eyes, the benefits are evident in this PR. The code not only reads much better, it’s also freed of the type wrangling and gymnastics needed to please the TS compiler.
We write all our client side code at 37signals now in pure JavaScript and the same too with any internal libraries. This is going to bring that in line.
All the love and appreciation to contributors who would prefer TypeScript. This is one of those debates where arguments aren’t likely to move anyone’s fundamental position, so I won’t attempt to do that." -- DHH


DanielRosenwasser:
I only unlocked this to react to the PR. 😄

PR #55676, opened 2023-09-08
Remove TypeScript

DominoTree:
By all accounts, TypeScript has been a big success for Microsoft. I've seen loads of people sparkle with joy from dousing JavaScript with explicit types that can be checked by a compiler. But I've never been a fan. Not after [giving it five minutes](https://signalvnoise.com/posts/3124-give-it-five-minutes), not after giving it five years. So it's with great pleasure that I can announce [we're dropping TypeScript](https://github.com/hotwired/turbo/pull/971) from the next big release of Turbo 8.

The fact is that I actually rather like JavaScript. I'd go so far as to say it's my second favorite language after Ruby. Yes, a distant second, but a second none the less. This wasn't always the case. But after we got proper classes in JavaScript, and all the other improvements that flowed since ES6, it's become a real joy to write.


jakebailey:
Duplicate of #55655

PR #50779, opened 2022-09-14
Partially revert #41044, restoring parameter destructurings in d.ts files

jakebailey:
Fixes #50707 (at least to the extent that we have the pre-4.8 behavior).

In a followup PR, I can remove these destructurings/bindings/whatevertheyrecalleds in the declaration emitter.

Note that the new errors added in #41044 remain. I guess the PR had two somewhat-related changes in it?

PR #55678, opened 2023-09-08
Bring back renamings in emitted d.ts files

### LINKED ISSUE #55654: Binding element alias is kept even though it is unused in declaration files
### 🔎 Search Terms

- This changed between versions 4.7 and 4.8

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.2.2#code/PTAEhRyUFgCgVAygCwPYFcA2ATUA7FALqAEYCmoA1qQA5ECGuOAlgM6gBOpAtigG6lZYAMzS4AxgSYpcoFqkxYAcoQDSpGqAAUAbzx0upAFyg6GJnTYBfY7tz6jsguya4A5qEsBKUNtih-0DCWsLDwgDLkoWAAgmYWoKygaCwC8TIESOQi4pLSJGhE+EQuoOnkLEyudgRonJGI8th4hCTkVLSpoFikYhh07HQ5uGxxTEXDMkxc1BjcpLgEA1IyXQtMGHUsAJ5cxCgYAHSsAEqkQqSc4ikJSSkA7khMYkigT90UbJxnF2Kk45v351I+2EogkS1kDSUqnU1AAQvkAJIsNTtHR6AzGUzmKw2dEOFhOFzuLy+GABDikarsGRYiywYIwOrgOoAVWSzBknCpaU21HImjE6EafNA1D69gI5zYBF5vxMbFupAwGG8LBQ8SIciFODIlBoBBB2XBWrQWAwag0aLsGJMsRxPjxxgJzjcHk8xhlfJQQlt2J8fgCXJqNLt9KAA

### 💻 Code

```ts
// ✔ 
// Should not be kept and is removed
function shouldNotKeep ({ name: alias }: { name: string }) {
    
}

// ❌
// Alias is used in the function but not in the signature
// Should not be kept in declarations as it is an implementation detail
// symbol.isReferenced is used which checks references anywhere.
function shouldNotKeepButIsKept ({ name: alias }: { name: string }){
    return alias
}

// ✔
// Used in return type (could pe parameters types as well) so it should be kept
function shoudlKeep ({ name: alias }: { name: string }): typeof alias {
    return alias
}
```

### 🙁 Actual behavior

`alias` is kept in declaration files for `shouldNotKeepButIsKept`

### 🙂 Expected behavior

Either `alias` is removed from declaration file for `shouldNotKeepButIsKept`, or `alias` is also kept for `shouldNotKeep`.

Seems like declaration emit is using `isReferenced` of the symbol. But this tracks usage of the symbol anywhere in the scope, we would need usage of the symbol in the function signature.

jakebailey:
Fixes #55654 (with the unfortunate solution)

This isn't ideal. I think the direction we want to go is to not emit renamings if we don't have to, but the current implementation uses `isReferenced` which unfortunately considers references other than those that affect the signatures of the functions. So if you ever actually _use_ the parameter you've renamed, then it will be kept, even if that isn't required. In the real world, that'd of course be nearly _all_ instances, because there's usually no reason to destructure a parameter and then not use it. Of course our tests have a lot of this because we typically don't care about what's inside a function, or we don't have a body at all!

This with #50779 effectively reverts all of #41044 _except_ the new error. You'll notice that this PR's baseline changes don't contain any new errors, which is good. This is because the error only happens when the signature _doesn't_ have an annotation, e.g.:

```ts
declare function foo({ a: string }): void; // Error
declare function bar({ a: string }: { a: string }): void; // Okay!
```

The former can't happen in emitted `d.ts` files; we're always going to fully annotate the output.

Maybe there's a solution that would instead attempt to figure out if the name is ever actually used elsewhere, but I'm not sure how to do that. I feel like we probably have that logic somewhere for other reasons? In which case, I'm happy to redo the PR.

### DIFF
--- a/src/compiler/transformers/declarations.ts
+++ b/src/compiler/transformers/declarations.ts
@@ -116,8 +116,6 @@ import {
     isFunctionDeclaration,
     isFunctionLike,
     isGlobalScopeAugmentation,
-    isIdentifier,
-    isIdentifierANonContextualKeyword,
     isIdentifierText,
     isImportDeclaration,
     isImportEqualsDeclaration,
@@ -684,7 +682,7 @@ export function transformDeclarations(context: TransformationContext) {
         return ret;
     }
 
-    function filterBindingPatternInitializersAndRenamings(name: BindingName) {
+    function filterBindingPatternInitializers(name: BindingName) {
         if (name.kind === SyntaxKind.Identifier) {
             return name;
         }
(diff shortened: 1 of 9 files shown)


andrewbranch:
I can’t see from the conversation in #55654 why we need to do anything here 🤔

jakebailey:
No, you're right; we could "do nothing" and it would be fine; the existing code only removes things that are definitely unreferenced. If we want to do this right with usage tracking, that's #55683.

PR #55683, opened 2023-09-08
Perform usage analysis on refenced binding aliases in function signatures.

### LINKED ISSUE #55654: Binding element alias is kept even though it is unused in declaration files

dragomirtitian:
Fix #55654 by marking  binding aliases that need to be preserved. 

This is done in a two step process:

1. We first remove all existing aliases (assuming they will not be used as this is the common case)
2. After the signature types are processed in the declaration emit, if any alias is marked as  being used (should be rare), then we go  though the parameters again to add back the referenced aliases.

Fixes #55654


src/compiler/transformers/declarations.ts
```diff
@@ -298,6 +301,7 @@ export function transformDeclarations(context: TransformationContext) {
     let lateStatementReplacementMap: Map<NodeId, VisitResult<LateVisibilityPaintedStatement | ExportAssignment | undefined>>;
     let suppressNewDiagnosticContexts: boolean;
     let exportedModulesFromDeclarationEmit: Symbol[] | undefined;
+    const usedBindingElementAliases = new Map<NodeId, Map<NodeId, BindingName>>();
```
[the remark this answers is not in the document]
dragomirtitian: I was debating if I should use that or node directly. I'll switch it over

src/compiler/transformers/declarations.ts
```diff
@@ -684,7 +699,7 @@ export function transformDeclarations(context: TransformationContext) {
         return ret;
     }
 
-    function filterBindingPatternInitializersAndRenamings(name: BindingName) {
+    function filterBindingPatternInitializers(name: BindingName) {
```
jakebailey, line 702: This function still modifies renamings so I would probably leave the name as-is.

src/compiler/transformers/declarations.ts
```diff
@@ -715,16 +731,80 @@ export function transformDeclarations(context: TransformationContext) {
...
                 elem.dotDotDotToken,
                 elem.propertyName,
-                filterBindingPatternInitializersAndRenamings(elem.name),
+                filterBindingPatternInitializers(elem.name),
                 shouldPrintWithInitializer(elem) ? elem.initializer : undefined,
             );
         }
     }
 
+    function ensureBindingAliasesInParameterList<T extends any[], N extends Node, P extends number>(original: Node, updater: (...p: T) => N, parametersIndex: P, args: T & Record<P, NodeArray<ParameterDeclaration>>) {
```
jakebailey, line 745: I'm not sure I like this signature; I'm not aware of this pattern existing in our code and it seems easy to mess up. Can this not just be a switch case on the input node type, or a function you only call on the parameter list, etc?

dragomirtitian: Fixed

jakebailey:
The focus of this is `isolatedDelarations` where that emitter is currently not emitting renamed type variables and so the two differ where `tsc`'s current emit is the suboptimal one.

dragomirtitian:
@andrewbranch Most of the 1.5k are tests 😅. The actual implementation is about 350 lines, most of which is just a switch to call the appropriate node factory to recreate the node.

As for CPU cycles, that was one of my worries too. As far as I can tell the only extra CPU cycles that are spent in the common case of not using the type alias in the signature is a call to `isBindingElement(declaration)` in the checker and a lookup on empty map in the declaration emitter, so hopefully no real impact.

While it's not a design goal to 100% reproduce TS declaration files with an external emitter, it would be ideal to limit cases where the two differ. This would be one example of a difference taht we can eliminate without. Always keeping the alias in is also a way to achieve this.

andrewbranch:
I guess I’m surprised that the more complicated behavior is the one we’re pursuing if the goal is consistency between tsc and a third-party `isolatedDeclarations` emitter.

dragomirtitian:
@andrewbranch Most of the machinery to track references is needed for tracking the usage of variables and imports anyway. This extra analysis for binding elements shouldn't make a third-party emitter much more complicated IMO.

Someone obviously cared enough to implement the removal of aliases, even if it did have some bugs, fixing the bugs seems the better way forward. They are just unused implementation details that should probably not be exposed unless absolutely necessary.

jakebailey:
This PR theoretically looks good to me, though the repetition in the d.ts emitter makes me feel like there must be a better way still that I'm missing here. Maybe @weswigham or @rbuckton know a better trick.

weswigham (commented):
Does this handle a binding element merged with a `var` well? eg,
```ts
function a({e1: t1}: {e1: {}})/*: typeof t1*/ {
    var t1: {} = 1;
    return t1;
}
```

dragomirtitian:
> Does this handle a binding element merged with a `var` well? eg,
> 
> ```ts
> function a({e1: t1}: {e1: {}})/*: typeof t1*/ {
>     var t1: {} = 1;
>     return t1;
> }
> ```

That is already an error: `Return type of exported function has or is using private name 't1'`. This PR does not attempt to change that. My only goal here was to not keep the binding alias if it was not used in declarations.

jakebailey:
So, as far as I can tell, this looks good to me, though I'm not a super fan of the repetition I don't think there's much of a way to do better. @weswigham do you have any other feedback here? I'm hoping to get the list of differences for isolatedDeclarations down to as small as possible (I think think this is the only one left??)

weswigham (approved):
Hmm, this looks fine to me, implementation-wise. I think I'd have a minor preference for merging the nodes from `bindingElementToMakeVisible` and `aliasesToMakeVisible` into a single `nodesToMakeVisible` list, but it's not strictly required if nobody else feels strongly about it.

jakebailey:
> Hmm, this looks fine to me, implementation-wise. I think I'd have a minor preference for merging the nodes from `bindingElementToMakeVisible` and `aliasesToMakeVisible` into a single `nodesToMakeVisible` list, but it's not strictly required if nobody else feels strongly about it.

As in, make it `nodesToMakeVisible?: (LateVisibilityPaintedStatement | BindingElement)[];` and then check the kind?

weswigham:
> As in, make it nodesToMakeVisible?: (LateVisibilityPaintedStatement | BindingElement)[]; and then check the kind?

Aye.

jakebailey:
I think the awkward thing is that there's only one `bindingElementToMakeVisible`... unless this is an indicator that the PR is missing something because there could be more than one like aliases?

dragomirtitian:
> Hmm, this looks fine to me, implementation-wise. I think I'd have a minor preference for merging the nodes from `bindingElementToMakeVisible` and `aliasesToMakeVisible` into a single `nodesToMakeVisible` list, but it's not strictly required if nobody else feels strongly about it.


I could merge those, the problem is that will change the common case to something like:

```ts
if (symbolAccessibilityResult.aliasesToMakeVisible) {
    if(!lateMarkedStatements) {
        lateMarkedStatements = []
    }
    for (const ref of symbolAccessibilityResult.aliasesToMakeVisible) {
        if(ref.kind === SyntaxKind.BindingElement) {
            const bindingElement = ref;
            const parameter = findAncestor(bindingElement, isParameter);
            Debug.assert(parameter !== undefined);
            const parent = getOriginalNode(parameter.parent);
            let aliases = usedBindingElementAliases.get(parent);
            if (!aliases) {
                usedBindingElementAliases.set(parent, aliases = new Map());
            }
            aliases.set(getOriginalNode(bindingElement), bindingElement.name);
        } else {
            pushIfUnique(lateMarkedStatements, ref);
        }
    }
}
```


Which would mean some extra work filtering out the binding elements first time `lateMarkedStatements` is assigned as well as using `pushIfUnique` that first time. Previously it could just do `if(!lateMarkedStatements) lateMarkedStatements = symbolAccessibilityResult.aliasesToMakeVisible`). Not sure if this significant. I would guess not.

> I think the awkward thing is that there's only one `bindingElementToMakeVisible`... unless this is an indicator that the PR is missing something because there could be more than one like aliases?

@jakebailey I don't think there is any way to have multiple binding elements to make visible if the code is valid. You can get it if with duplicate identifiers but I don't think that is a case we need to think to deeply about as it's a syntax error (`function foo({a: v, b: v}: {a: number, b: string}): typeof v`)

jakebailey:
I think I'd personally prefer to leave it as is, in that case. I don't feel strongly, but it does seem like an extra loop to find one thing...

jakebailey (approved):

jakebailey:
I'll quick run the extended suite, though I expect it to not really do anything since it can't detect emit changes.

@typescript-bot test top200
@typescript-bot run dt

jakebailey:
Just to note it, this is being reverted in #57020.

PR #32131, opened 2019-06-27
Accurate Array.prototype.flat definition

Bannerets:
Fixes #29604 (and #29741).

Examples:

```typescript
const a0 = [1, [2]].flat() // const a0: number[]
const a1 = [[1], 'a'].flat(1) // const a1: (string | number)[]
const n: number = 1
const a2 = [1, [2]].flat(n) // const a2: (number | number[])[]
const b0 = [[1], ['a']].flat(1) // const b0: (string | number)[]
const c0 = [
  [[1]],
  [[2]]
].flat(2) // const c0: number[]
const arr = [[[1, [2, [3, [4]]]]]]
const d0 = arr.flat(2) // const d0: (number | (number | (number | number[])[])[])[]
const d1 = arr.flat(n) // const d1: (number | number[] | (number | number[])[] | (number | (number | number[])[])[] | (number | (number | (number | number[])[])[])[] | (number | (number | (number | number[])[])[])[][])[]
```

The old declaration returns `any[]` for all cases except `c0`, `d0` and supports `depth` only up to 7 (this one supports up to 20).

PR #55686, opened 2023-09-08
Unlimited Array.prototype.flat depth

### LINKED ISSUE #55685: Array.prototype.flat limited depth to 21
### ⚙ Compilation target

ES2019

### ⚙ Library

es2019

Array.prototype.flat definition is supporting `depth` only up to 21.  Over 21 return wrong type.

### Sample Code

```TypeScript
const foo = [ [ ['string' ] ] ]
const flatFoo = foo.flat(22) // string[][][] should be string[]
```

### Documentation Link

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat

PKrasinski:
Fixes #55685

The old `Array.prototype.flat` definition supports `depth` only up to 21.  Over 21 return wrong type.
This solution allow call flat method with unlimited depth parameter.

```ts
const foo = [ [ ['string' ] ] ]
const flatFoo = foo.flat(1000000) // string[]
```
Performance (checked with [this code](https://github.com/microsoft/TypeScript/pull/32131#issuecomment-604784171)) is the same or better in some cases. 

The only limitation of this solution is that for arrays deeper then `999` levels `unknown[]` type is returned. 

```ts
[
  [
    [
       // 997 levels
    ]
  ]
].flat(1000) // unknown[]
```


PKrasinski:

MartinJohns:
Introducing new global types with generic names and increasing the complexity of the types enormously for a highly questionable use case. :-/

PKrasinski:
Yep, probably you are right. I was just looking for the way how to type safe recurrent functions so I checked how it looks like in flat method declaration and I was a bit surprised that solution using hard-coded array and returns wrong type when the depth is higher. So I decided to write my own implementation just for fun 😄  I could avoid new global types in this solution. I also know that probably I solved the problem that nobody had.

DanielRosenwasser:
I think for reasons discussed here and elsewhere, I would encourage people who need this to include their own local augmentations to the global scope.

PR #55690, opened 2023-09-08
Fix file matching with tsx and dts of same name are included by include patterns

sheetalkamat:
Fixes https://github.com/microsoft/TypeScript/issues/54389

### DIFF
--- a/src/compiler/commandLineParser.ts
+++ b/src/compiler/commandLineParser.ts
@@ -3849,7 +3849,10 @@ function hasFileWithHigherPriorityExtension(file: string, literalFiles: Map<stri
         return false;
     }
     for (const ext of extensionGroup) {
-        if (fileExtensionIs(file, ext)) {
+        // d.ts files match with .ts extension and with case sensitive sorting the file order for same files with ts tsx and dts extension is
+        // d.ts, .ts, .tsx in that order so we need to handle tsx and dts of same same name case here and in remove files with same extensions
+        // So dont match .d.ts files with .ts extension
+        if (fileExtensionIs(file, ext) && (ext !== Extension.Ts || !fileExtensionIs(file, Extension.Dts))) {
             return false;
         }
         const higherPriorityPath = keyMapper(changeExtension(file, ext));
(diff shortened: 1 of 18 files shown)


src/testRunner/unittests/config/matchFiles.ts
```diff
@@ -141,6 +141,26 @@ const caseSensitiveOrderingDiffersWithCaseHost = new fakes.ParseConfigHost(
     }),
 );
 
+const caseInsensiiveHostWithSameFileNamesWithDifferentExtensions = new fakes.ParseConfigHost(
```
jakebailey, line 144: Typo here

```suggestion
const caseInsensitiveHostWithSameFileNamesWithDifferentExtensions = new fakes.ParseConfigHost(
```

jakebailey:
@typescript-bot test top200

src/compiler/commandLineParser.ts
```diff
@@ -3849,7 +3849,10 @@ function hasFileWithHigherPriorityExtension(file: string, literalFiles: Map<stri
         return false;
     }
     for (const ext of extensionGroup) {
-        if (fileExtensionIs(file, ext)) {
+        // d.ts files match with .ts extension and with case sensitive sorting the file order for same files with ts tsx and dts extension is
+        // d.ts, .ts, .tsx in that order so we need to handle tsx and dts of same same name case here and in remove files with same extensions
+        // So dont match .d.ts files with .ts extension
+        if (fileExtensionIs(file, ext) && (ext !== Extension.Ts || !fileExtensionIs(file, Extension.Dts))) {
```
jakebailey, line 3855: There's the same code in the function below this one; does it also need to change?

(I don't fully understand this change so I'll probably defer to someone else)

sheetalkamat: No since if you look at ts order its `.ts, .tsx, .d.ts` so in this loop we first look at `.ts` In the below loop we iterate over in reverse order to remove file with the lower priority extensions so that part is will match `.d.ts` first and hence correct.

sheetalkamat:
Ping @weswigham @andrewbranch

andrewbranch (approved):

Issue #55693, opened 2023-09-09
Spreading a named tuple slice results in the names being in the wrong position


danny-may:
### 🔎 Search Terms

named tuple (label:bug)

### 🕗 Version & Regression Information

- This was introduced in 5.2.2 as prior versions would discard the element names

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.2.2#code/JYOwLgpgTgZghgYwgAgLIFcA2ZgGkICeqcADgDz4EDOyEAHpCACY1QRxMD2ImBy6IANYhOAdxABtALoAaZADU4mdBAB8yAN4BYAFDJ9yAOYQwFQrQYRmNAOInKVM9VWqAFADpPgwgC5klAEo-OzAAJQgqLFMHOUo5RWU1XQBfXTACEhQQhycaekYWZDYObl5+IRFxaXUAXn9CPMtrZAlPd1AYaGRwgFs4UDkOroB9KWRdA2QAfnrqZAAfZGyGsl7+kHUJgz8HAG5dNIysk3DI7DIASSYLAtZ2Lh4+AWExSVlkAEEoQ0bbpfsVldVPElCpap9vr8rIUruM9AYZgkVAtykwIDBQBAmFt9H5YfloTRWp4vj85G0hlBuhEwFIcZNpsgLpAoHAAEaYCBkYnuU5gEGJKSqemTPwgCAAN2g+x0ugQ3CoYGQdD8GGweEIxHIEmATAAjH5FVBQIZBkwAEyGsDGkCGd6cNkAKwgCDA4JAWEwAEIZfKQIrkLq9cg6gBydkIUPIOA0I0m30KpV9MAIAAWERDyvcxjAriDAV2cKAA

### 💻 Code

```ts
interface MultiKeyMap<Keys extends readonly unknown[], Value> {
    get<Key extends GetKeys<Keys>>(...key: Key): GetResult<Keys, Key, Value>
}
type GetKeys<Keys extends readonly unknown[]> = Keys extends [...infer Remain, infer _] 
    ? Keys | GetKeys<Remain> 
    : Keys;

type GetResult<Id extends readonly unknown[], Args extends GetKeys<Id>, Value> = Args extends Id 
    ? Value | undefined
    : Id extends [...Args, ...infer Rest]
        ? Iterable<[...Rest, Value]>
        : never;

const x: MultiKeyMap<[id1: string, id2: string], object> = null!;
const id1 = 'abc' as string;
[...]

PR #55695, opened 2023-09-09
Fixed cache key computation for tuple target types with partially named members

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/55693

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -16392,10 +16392,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             // [...X[]] is equivalent to just X[]
             return readonly ? globalReadonlyArrayType : globalArrayType;
         }
-        const memberIds = mapDefined(namedMemberDeclarations, node => node ? getNodeId(node) : undefined);
         const key = map(elementFlags, f => f & ElementFlags.Required ? "#" : f & ElementFlags.Optional ? "?" : f & ElementFlags.Rest ? "." : "*").join() +
             (readonly ? "R" : "") +
-            (memberIds.length ? "," + memberIds.join(",") : "");
+            (some(namedMemberDeclarations, node => !!node) ? "," + map(namedMemberDeclarations, node => node ? getNodeId(node) : "_").join(",") : "");
         let type = tupleTypes.get(key);
         if (!type) {
             tupleTypes.set(key, type = createTupleTargetType(elementFlags, readonly, namedMemberDeclarations));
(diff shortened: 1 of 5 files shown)


src/compiler/checker.ts
```diff
@@ -16392,10 +16392,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             // [...X[]] is equivalent to just X[]
             return readonly ? globalReadonlyArrayType : globalArrayType;
         }
-        const memberIds = mapDefined(namedMemberDeclarations, node => node ? getNodeId(node) : undefined);
```
Andarist, line 16395: this would previously compute the same cache key segment for `[undefined, labelNode]` and `[labelNode, undefined]`

src/compiler/checker.ts
```diff
@@ -16392,10 +16392,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             // [...X[]] is equivalent to just X[]
             return readonly ? globalReadonlyArrayType : globalArrayType;
         }
-        const memberIds = mapDefined(namedMemberDeclarations, node => node ? getNodeId(node) : undefined);
         const key = map(elementFlags, f => f & ElementFlags.Required ? "#" : f & ElementFlags.Optional ? "?" : f & ElementFlags.Rest ? "." : "*").join() +
             (readonly ? "R" : "") +
-            (memberIds.length ? "," + memberIds.join(",") : "");
+            (some(namedMemberDeclarations, node => !!node) ? "," + map(namedMemberDeclarations, node => node ? getNodeId(node) : "_").join(",") : "");
```
jakebailey, line 16397: Is there a reason to precheck rather than just doing the map and letting it give the empty string via returning nothing?

Andarist: Hmm, not rly - this only keeps the cache key shorter if there are no named members.

jakebailey: Oh, I misread this and didn't notice that this was adding a conditional prefix. Just seemed less nice to double iterate but it's fine.

Andarist: I debugged this down in a moment between things so I didn't exactly give it an extended thought at that moment. It was previously added conditionally and I didn't spend a second thinking if this would still be needed after things change. 

Since you called it out, I re-evaluated it and I think it is, indeed, redundant now. So I would be more than OK with dropping this. Unlabeled tuples are probably more common than labeled ones though but I doubt that the cache key size would impact perf in a significant way.

jakebailey: All good, I'd just leave it as it's not a regression.

jakebailey:
@typescript-bot test top100
@typescript-bot run dt

jakebailey (approved):

jakebailey:
Maybe this might be worth a backport given this can end up breaking the new feature in 5.2? @DanielRosenwasser

PR #55696, opened 2023-09-10
Allow more things to be evaluated in enum members with template expression initializers

Andarist:
@dragomirtitian has noticed this inconsistency~ ([TS playground](https://www.typescriptlang.org/play?ts=5.2.2#code/MYewdgzgLgBAGjAvDABgEgN5QE4FcCmAvijAIYQyiRQDcAUPmLgLYwCCMGMdMv8SqTAEZiAGh58AmgPRY8RFOMJA)):
```ts
const X = `${true}` as const; // "true"
enum A { 
    X = `${1}`, // "1"
    Y = `${true}`, // Type 'string' is not assignable to type 'number' as required for computed enum member values.(18033)
}
```

Note that even with changes in this PR there will still be inconsistencies like:
```ts
const test1 = `${1 + 1}` as const; // `${number}`

export enum Test2 {
  A = `${1 + 1}`, // "2"
}
```

It's easy to fix this but I'm not sure if it's desirable so I left it out of this PR for now.

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -44426,8 +44426,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     function computeConstantValue(member: EnumMember): string | number | undefined {
         const isConstEnum = isEnumConst(member.parent);
         const initializer = member.initializer!;
-        const value = evaluate(initializer, member);
-        if (value !== undefined) {
+        const evaluated = evaluate(initializer, member);
+        if (evaluated) {
+            const { value } = evaluated;
             if (isConstEnum && typeof value === "number" && !isFinite(value)) {
                 error(
                     initializer,
@@ -44436,6 +44437,11 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                         Diagnostics.const_enum_member_initializer_was_evaluated_to_a_non_finite_value,
                 );
             }
+            else if (typeof value !== "string" && typeof value !== "number") {
+                checkTypeAssignableTo(checkExpression(initializer), numberType, initializer, Diagnostics.Type_0_is_not_assignable_to_type_1_as_required_for_computed_enum_member_values);
+                return undefined;
+            }
+            return value;
         }
         else if (isConstEnum) {
             error(initializer, Diagnostics.const_enum_member_initializers_must_be_constant_expressions);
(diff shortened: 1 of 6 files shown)


src/compiler/checker.ts
```diff
@@ -44446,77 +44452,125 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         else {
             checkTypeAssignableTo(checkExpression(initializer), numberType, initializer, Diagnostics.Type_0_is_not_assignable_to_type_1_as_required_for_computed_enum_member_values);
         }
-        return value;
+        return undefined;
     }
 
-    function evaluate(expr: Expression, location?: Declaration): string | number | undefined {
+    function evaluate(expr: Expression, location?: Declaration): { value: string | number | bigint | boolean | null | undefined; } | undefined {
```
Andarist, line 44458: I'm open to using a different strategy between differentiating evaluated `undefined` and lack of the evaluation result.

jakebailey: Could always use a Symbol as a sentinel value.

src/compiler/checker.ts
```diff
@@ -44446,77 +44452,125 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+                            return { value: left - right };
                         case SyntaxKind.PercentToken:
-                            return left % right;
+                            return { value: left % right };
                         case SyntaxKind.AsteriskAsteriskToken:
-                            return left ** right;
+                            return { value: left ** right };
+                    }
+                }
+                else if (typeof left === "bigint" && typeof right === "bigint") {
```
Andarist, line 44520: This is somewhat redundant since it doesn't really close "gap" between the template literal types simplification and this evaluation here. I feel like it would be weird for those 2 to behave differently though:
```ts
enum T {
  A = `${1 +1 }`,
  B = `${1n +1n }`
}
```

On top of that, it mostly repeats the number-oriented case but I don't know how to implement this in a better way since I can't force TS to treat two different identifiers to be of the same type. It always complains about `number | bigint` in this scenario (for good reasons).

tests/baselines/reference/enumConstantMemberWithTemplateLiterals.errors.txt
```diff
@@ -50,4 +50,10 @@ enumConstantMemberWithTemplateLiterals.ts(28,15): error TS2363: The right-hand s
         b = `1` + `1`,
         c = "2" + `1`
     }
+    
+    declare enum T8 {
+      X = `${1}`,
+      Y = `${true}`,
+      Z = `pre${!false}middle${"foo" + true}`,
```
Andarist, line 57: If the PR gets softly accepted - I should add more test cases here. I didn't want to spend time on it yet in case it gets rejected

src/compiler/checker.ts
```diff
@@ -44446,77 +44452,125 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+                return { value: (expr as StringLiteralLike).text };
+            case SyntaxKind.TemplateExpression: {
+                const value = evaluateTemplateExpression(expr as TemplateExpression, location);
+                if (typeof value === "undefined") {
+                    break;
+                }
+                return { value };
+            }
             case SyntaxKind.NumericLiteral:
-                checkGrammarNumericLiteral(expr as NumericLiteral);
```
Andarist, line 44512: I believe this was redundant here since it should already be checked elsewhere and it's not exactly the responsibility of this evaluation to raise errors like this.

tangent: I can't see any test case for this diagnostic
> Numeric literals with absolute values equal to 2^53 or greater are too large to be represented accurately as integers.(80008)

src/compiler/checker.ts
Andarist: This actually crossed my mind :D but I wasn't sure if this should just be `return Symbol()` or if I should introduce some global symbol or what. Code style - I prefer the current approach but it might be nice to avoid introducing super disposable objects like this in the compiler.

jakebailey: Disposable objects is what I'm thinking is going to bite this; I'd just make a const local or something. But maybe it doesn't actually matter.

jakebailey: I doubt it matters so is fine to leave as-is; we have worse things like helpers which almost always return empty arrays of 1 item (boo)

DanielRosenwasser:
@typescript-bot pack this

DanielRosenwasser:
The biggest thing we dislike is actually the error message you get today. The temptation to expand the functionality actually comes from that. It's awkward to be in a spot where you could put `true` in a string literal, but not as the initializer of an enum member; but we also don't want to allow `true` as the initializer of an enum member.

Ideally we'd just have an error message like "This string could not be evaluated at compile-time because it contains unsupported expressions." This might require trafficking around some extra information at each step describing the type of the expression, or it could be some sort of best-effort thing for certain kinds of expression forms.

sandersn:
Is this direction worth pursuing? It sounds like what @DanielRosenwasser recommends is to improve error reporting on unsupported expressions rather than expanding the expressions that can be used.
If not, I'd like to close the PR to help with housekeeping.

PR #49208, opened 2022-05-22
fix(46246): Extending Class/Interface With a Primitive Throws a More Concise Error Report

maross3:
fixes #46246 

### Example
Class:
```TS
class foo extends number {} 
```
Error:
```A class cannot extend a primitive type like 'number'. A class can only extend named object types.```

Interface:
```TS
interface bar extends number { }
```
Error:
```An interface cannot extend a primitive type like 'number'. An interface can only extend named object types.```

### Notes:
* gulp `runtests` passed on current branch.
* I updated a few unit tests to test for the error. I did not feel new test cases were necessary, as the errors were well covered.
* Function responsible for the ambiguous error: `checkAndReportErrorForUsingTypeAsValue`
* Function to check heritage clause: `checkAndReportErrorForExtendingPrimitiveType`

## Description
The approach was to check for extension of primitives directly before the ambiguous error message was thrown. This lead to making a new function.

I didn't add the code to the function `checkAndReportErrorForUsingTypeAsValue` because the intended error checking functionality was separate from evaluating the *value* for error checking. 
 
The new function `checkAndReportErrorForExtendingPrimitiveType` now handles checking heritage clause related errors in regards to classes extending primitives.

In the code, we check if the current errorLocation's expected heritage clause was, in fact, a `SyntaxKind.HeritageClause`. A bitmask was used to check against the expected heritage clause: `errorLocation.parent.parent.kind`. If it's a primitive name and a heritage clause, we throw a more descriptive error.

PR #46357, opened 2021-10-14
Fix46246

HarsheetKakar:
Fixes #46246 

The new error message is : `An interface cannot extend a primitive type like '{0}'; an interface can only extend named types and classes`

where {0} has the primitive type name e.g. any, number, string etc.

Please give feedback as this is my first PR on this repo.

Issue #46246, opened 2021-10-06
Unhelpful error message when `interface ... extends number`


Mr-Andersen:
# Bug Report

### 🔎 Search Terms

interface extends only refers to a type

### 🕗 Version & Regression Information

This is the behavior in every version I tried, and I reviewed the FAQ for entries about interface extension

### ⏯ Playground Link

[Playground link with relevant code](https://www.typescriptlang.org/play?ts=4.5.0-dev.20211006#code/KYDwDg9gTgLgBASwHY2FAZgQwMbDgWWAGcjMBzYASQBM5RUlqi4kBXAWwCM04BvAXyA)

### 💻 Code

```ts
export interface MessageId extends number {}
```

### 🙁 Actual behavior

error TS2693: 'number' only refers to a type, but is being used as a value here.

### 🙂 Expected behavior

Informative error about why an interface cannot extend `number` (or `string` -- the error is the same besides the type name)

younes-io:
I would think of a message like: `an interface cannot extend a primitive type like 'number'; an interface can only extend named types and classes`

What do you think?

HarsheetKakar:
I was actually working on it, sorry I forgot to mention that earlier

charan1998:
Anyone working on this? Or can I take this up?

JoelFickson:
Was this issue closed? I wanna take it.

bentongxyz:
Hi there, if this is not being actively worked on right now, may I give this a try?
Thanks!

JCalmCrasher:
Hi, @RyanCavanaugh I'm new to TypeScript; I'll like to give this a try.

Thanks.

maross3:
I opened a draft PR for this issue: https://github.com/microsoft/TypeScript/pull/49208

PR #55712, opened 2023-09-11
Add conditions to check for class extension or implementation of prim…

qwefgh90:
fixes unresolved issues that discussed on https://github.com/microsoft/TypeScript/pull/49208.

#### Class implementing the primitive:

```typescript
class C implements number { }
class C2 implements string { }
class C3 implements boolean { }
```

Actual behavior:
```
'number' only refers to a type, but is being used as a value here.
'string' only refers to a type, but is being used as a value here.
'boolean' only refers to a type, but is being used as a value here.
```

Improvement:
```
A class cannot implement a primitive type like 'number'; a class can only implement interfaces.
A class cannot implement a primitive type like 'string'; a class can only implement interfaces.
A class cannot implement a primitive type like 'boolean'; a class can only implement interfaces.
```

#### Class extending the primitive:

```typescript
class C extends number { }
class C2 extends string { }
class C3 extends boolean { }
```

Actual behavior:
```
'number' only refers to a type, but is being used as a value here.
'string' only refers to a type, but is being used as a value here.
'boolean' only refers to a type, but is being used as a value here.
```

Improvement:
```
A class cannot extend a primitive type like 'number'; a class can only extend named types and classes.
A class cannot extend a primitive type like 'string'; a class can only extend named types and classes.
A class cannot extend a primitive type like 'boolean'; a class can only extend named types and classes.
```

####  Notes:
- `gulp runtests` passed on current branch.
- The errors related to `implement` weren't covered by current unit tests. I added a new test `tests\cases\compiler\classImplementsPrimitive.ts` to test classes implementing primitive types. 
- I updated a few unit tests to test for the error.
- I made additional changes for some minor typos.

### Description
PR https://github.com/microsoft/TypeScript/pull/49208 was initially created to address issue https://github.com/microsoft/TypeScript/issues/46246 along with some related problems.
 
However, that issue https://github.com/microsoft/TypeScript/issues/46246 has been resolved by a different PR. So, PR https://github.com/microsoft/TypeScript/pull/49208 wasn't merged and It was closed while leaving some issues that had been newly found and discussed in this PR, specifically related to incorrect error messages where classes extends or implements primitive types, as shown in the examples above.

To address these issue, I have made changes to the `checkAndReportErrorForUsingTypeAsValue` function, which has been responsible for these kinds of errors since https://github.com/microsoft/TypeScript/pull/46357 was merged.

I have added new functions, `isExtendedByClass` and `isImplementedByClass` which determine if `errorLocation` 's grandparent contains `SyntaxKind.ImplementsKeyword` or `SyntaxKind.ExtendsKeyword`. If a condition evaluates to true, the relevant error message is emitted.

I would appreciate any feedback!


DanielRosenwasser:
The errors are no longer exactly correct. They should be

```
An interface cannot extend a primitive type like '{0}'. It can only extend other named object types.

A class cannot implement a primitive type like '{0}'. It can only extend other named object types.

A class cannot extend a primitive type like '{0}'. Classes can only extend constructable values.
```

src/compiler/diagnosticMessages.json
```diff
@@ -3663,6 +3663,14 @@
         "category": "Error",
         "code": 2854
     },
+    "A class cannot extend a primitive type like '{0}'. Classes can only extend constructable values.": {
+        "category": "Error",
+        "code": 2855
+    },
+    "A class cannot implement a primitive type like '{0}'. It can only extend other named object types.": {
```
sandersn, line 3670: Should this message refer to 'implement' both times? Seems like it should but I'm not sure.

```suggestion
    "A class cannot implement a primitive type like '{0}'. It can only implement other named object types.": {
```

src/compiler/checker.ts
```diff
@@ -3790,6 +3796,28 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         return false;
     }
 
+    function isExtendedByClass(node: Node): boolean {
```
sandersn, line 3799: Instead of creating two new functions, can you instead rename `isExtendedByInterface` to `isInheritedByClassOrInterface` and add parameters `heritageKind: SyntaxKind, containerKind: SyntaxKind` for calculating `isExtending` and `isInterface`, respectively?

tests/cases/compiler/classImplementsPrimitive.ts
```diff
@@ -0,0 +1,17 @@
+// classes cannot implement primitives
+
+class C implements number { }
+class C2 implements string { }
+class C3 implements boolean { }
+class C4 implements Void  { }
```
sandersn, line 6: no need for this test; there's no type named Void.

tests/cases/compiler/classImplementsPrimitive.ts
```diff
@@ -0,0 +1,17 @@
+// classes cannot implement primitives
+
+class C implements number { }
+class C2 implements string { }
+class C3 implements boolean { }
+class C4 implements Void  { }
+class C4a implements void {}
+class C5 implements Null { }
```
sandersn, line 8: same

tests/cases/compiler/classImplementsPrimitive.ts
```diff
@@ -0,0 +1,17 @@
...
+
+class C implements number { }
+class C2 implements string { }
+class C3 implements boolean { }
+class C4 implements Void  { }
+class C4a implements void {}
+class C5 implements Null { }
+class C5a implements null { }
+class C6 implements undefined { }
+class C7 implements Undefined { }
```
sandersn, line 11: same

tests/cases/compiler/classImplementsPrimitive.ts
```diff
@@ -0,0 +1,17 @@
+// classes cannot implement primitives
+
+class C implements number { }
+class C2 implements string { }
+class C3 implements boolean { }
+class C4 implements Void  { }
+class C4a implements void {}
+class C5 implements Null { }
+class C5a implements null { }
+class C6 implements undefined { }
```
sandersn, line 10: testing null and undefined doesn't feel relevant to this PR since the new errors don't apply. Can you grep for existing tests that have 'implements null/undefined'? If there aren't any, feel free to leave them here; otherwise, remove them.

tests/cases/compiler/classImplementsPrimitive.ts
```diff
@@ -0,0 +1,17 @@
...
+class C5 implements Null { }
+class C5a implements null { }
+class C6 implements undefined { }
+class C7 implements Undefined { }
+
+enum E { A }
+class C8 implements E { }
+
+class C4a implements void {}
+class C5a implements null { }
```
sandersn, line 17: same comment for these; it's nice to have these tests, but if they're already in the test suite, you don't need to add them.

sandersn (changes requested):

src/compiler/diagnosticMessages.json
qwefgh90: You're right, that looks a typo. 'implement' would be better in this context.

tests/cases/compiler/classImplementsPrimitive.ts
qwefgh90: Thanks for catching that.

tests/cases/compiler/classImplementsPrimitive.ts
qwefgh90: There aren't both. However, I would remove these cases because these tests aren't relevant to this PR as you said.

tests/cases/compiler/classImplementsPrimitive.ts
qwefgh90: I tried to look up the case of class implementing enum. However there isn't any result found. this time, I would remove these cases because these tests aren't relevant to this PR

src/compiler/checker.ts
qwefgh90: I agree that the `isExtendedByInterface`, `isExtendedByClass`, and `isImplementedByClass` functions are similar. I tried to merge them into a single function called `isInheritedByClassOrInterface`, but realized it is redundant because it simply returns true or false after checking if the node contains a HeritageClause and a Class or Interface. I couldn't decide which error message to print with single boolean value returned by `isInheritedByClassOrInterface`.

Therefore, I think it's best to move all of the logic for checking these conditions into the `checkAndReportErrorForUsingTypeAsValue` function.
You can find changes in a next commit.

andrewbranch (approved):
Thank you!

andrewbranch:
Sorry, looks like you’ll need to `git merge main` and fix the diagnostic message conflicts. Someone probably took one of the numbers you were using 🙈

qwefgh90:
Thank you for letting me know. I did `git rebase main` to put my changes on the top of the latest commit. I stored old branch in case I made a mistake.

Is it okay to `git rebase main`? I just fixed error numbers during the rebase.

sandersn (approved):

PR #55714, opened 2023-09-11
Use `...args: never` in builtin conditional type helpers

### LINKED ISSUE #55667: `ReturnType` and `InstanceType` don't work for functions/constructors with `never` in their parameters
### 🔎 Search Terms

ReturnType InstanceType any never

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about `ReturnType` and `InstanceType`.

### ⏯ Playground Link

https://www.typescriptlang.org/play#code/C4TwDgpgBA+gjFAvFAShYBXATgOwCrgQA8AFAHQUCGWA5gM4BcUOEAbhFgJRIB8UA5ADMA9sP48AUAHopUOQD0A-BImhIsAExIoASRx1glHAGMIBSERYB3KOSq1GzNh26I+Q0eOmyFioA

### 💻 Code

```ts
type _1 = ReturnType<(...args: never) => 'foo'>
//   ^? type _1 = any

type _2 = InstanceType<new (...args: never) => 'foo'>
//   ^? type _2 = any
```

### 🙁 Actual behavior

`ReturnType` and `InstanceType` always evaluate to `any` when given a function/constructor whose parameters are `...args: never` (or `...args: never[]`, `a: never, b: string`, etc).

### 🙂 Expected behavior

`ReturnType` and `InstanceType` should return the actual return/instance type.

Changing the parameter constraint in the conditions of these types from `any` to `never` seems to fix the issue:

```ts
type ReturnType<T extends (...args: any) => any> = T extends (...args: never) => infer R ? R : any

type InstanceType<T extends abstract new (...args: any) => any> = T extends abstract new (...args: never) => infer R ? R : any
```

I don't think this will cause any regressions, and did some spot-checking to validate that, but haven't run the entire test suite or anything like that yet.

### LINKED ISSUE #56919: Counterexample to current ReturnType logic with type parameters
### 🔎 Search Terms

"ReturnType inference any"

 This is the behavior in every version I tried

### ⏯ Playground Link

https://www.typescriptlang.org/play?#code/C4TwDgpgBAKgDFAvFAShYBXATgOxuCAHkJgD4AKYALlgEolSoBrCEAewDNZSBuAWABQAeiFQxAPQD8gwaEhQAogA9IAY2AQAJmky5885C3ZcMOJjjYB3HP2GiJkoA

### 💻 Code

```ts
type T0 = ReturnType<<T>(t: T) => keyof T>; // any

type ExpectedReturnType = keyof unknown; // never
```

### 🙁 Actual behavior

`ReturnType` fails and returns `any`

### 🙂 Expected behavior

Given the fact that the upper bound of the type parameter `T` is `unknown`, I'd expect `keyof unknown` as result, i.e. `never`.

This is somewhat related to #55667 and it comes from the attempts made to understand why #55714 failed to pass all the test cases.

The problem seems to be located in `compareSignaturesRelated`, called by `getConditionalType` to resolve the `ReturnType`'s inner conditional type. We have that `source = <T>(t: T): keyof T` is compared against `target = (...args: any): never`. This `never` is, in fact, the correct return type that was successfully inferred.

Because `source` has a type parameter, `instantiateSignatureInContextOf(source, target, ...)` comes into play and we get a new `source`: `(t: any): string | number | symbol`. I'm not 100% sure about the logic of  `instantiateSignatureInContextOf`, but it's worth noting that `target` is the second argument of the call above and the type of `target`'s parameter is `any`. Therefore it seems that, from the `source` point of view, `T` becomes just `any` (and `keyof any` is exactly `string | number | symbol`).

We now have `source = (t: any): string | number | symbol` to be compared against `target = (...args: any): never`. This comparison fails because of the return types: `string | number | symbol` is not assignable to `never`.

&nbsp;

The example in this issue may seem intentionally crafted, and in fact it is, but the point was to highlight the same problem explained by [this comment](https://github.com/microsoft/TypeScript/pull/55714#issuecomment-1732597254). There we have a similar issue with the `source` we get from `instantiateSignatureInContextOf`: `any` has been substituted with `never` inside `ReturnType` definition, therefore if `source` has a type parameter it will be instantiated with `never` and sometimes things don't end well.

This probably undermines the possibility of having an all-encompassing definition for `ReturnType` if the current under-the-hood logic doesn't change. However you type the rest parameter inside it, I suppose you can always find a counterexample that uses a not-yet-instantiated type parameter to make its logic fail. I think the same thing can be said for `InstanceType`.

Andarist:
closes https://github.com/microsoft/TypeScript/issues/55667 , cc @andrewbranch 
fixes https://github.com/microsoft/TypeScript/issues/56919


jakebailey:
@typescript-bot test top200
@typescript-bot run dt
@typescript-bot pack this

DanielRosenwasser:
Why not `InstanceType`?

src/lib/es5.d.ts
andrewbranch, line 1: Should `InstanceType` have been updated in this file?

Andarist:
I missed the `InstanceType` with my multi-cursor edit :v thanks for catching that.

andrewbranch:
@typescript-bot test top200

Andarist:
Extracted (not exactly minimal yet) repro from `vuetify`: [TS playground](https://www.typescriptlang.org/play?ts=5.2.2#code/C4TwDgpgBAkgZgQQHYgDwBUA0UCa2ByAfFALxQAMUEAHsBEgCYDOUAjFAGRTpQD8uUAFxR8AbgBQoSFAAiAQ2BzSUAEoQAxgHsATg1RNg2gJZIA5tgCuSANZJNAdySEJU6DIhw5FgDbAAYnLqwDpo6MRkABRg2ppgTMLyigCUpMQ8AD5QSD7eUJlWDB4mEAwu4NAAwpoAtmCaSPTAAPIARgBWGsAACjFxTWDARvVMqF3KiXLhUADe4lBQANoA0lAmUNYQIJpwUF0AusI9saPLe8SZ2d7eEgC+EiZ02p7q0Edg-YPDGMpyKNgyyjCMzmUFcvEOvXQ5Qw51B2gs0AuOQk820EAAjhYjGiGOCoC1NJpvBBfiioIVPD5gHiAZl3JTfAEgiFUDJYZdcvlGEUGgw8lBNO1OmSAG5ybxGBgKHS8CJi7wI4RWWwOJBJYQEokkpC3STlXa9ACyEGAAAtNHosNwqkgDPDmdofigpgt0HsqLR6MxFiCIhEAHSBuTaUzxKC-EApEjECMpLkU4oMTDiPYg-izeaZrIQexQCLq63DQwWB1krP54ToMuZtFyBj1bwgKDRTTBVyVm12kvBbTVm4g4QNEUQXt66RvTvFh3fMgR8IgzIZrPZ3MBoMhsMRhZ7As8LjTO4g+b9+aLo+ZivcPsLg2xY1mi0wsrjyHQoFkCdF+09mH8z+2qcfzCbdn1eXoMH+QEpjeD4hltCDZFhN4oUgJ8x2gGAkDgEdkLfF03Q9OhGBYBYOVTeZ+AjKAAHpqKyHJODhBEoHsTQfD5TwjFyYJViwkcB0WAiaCI70FmmUFykHBjMmLaAbnIvgaLohAWHQABlVYmCYZjTWAYA4kEWjTCMM0LBaf0tGqajDSMdQYiYbZgGolCIDUuyjAGaijC0hEmGo1gABYAA4ACYAE5z1oqAHOqCBzVzAADVoOiCSdvx0BKWNNeheOw7QcSgOAYmqGY813KAbnxDQaggFgEojBLIrohKACFCWJX40u7DKspykw8oKoqalvMAuodCJ7BM00Rvvc0GBSFpqtiuq2q1X5GqzCMBNdd1hK9Ejks6Mae35cT2ygQ7Uq-brHXktNVGq3R9EMExzHDZ1tqEz1iMWVaOqQY6dFOiTIGEP7tUB26FP4TV-s+3bvtExIIEh4HzuR1G7oo2QFAgeHCP2xYIn6kcoAAVSSbc0ckvMScdCmqaxxSyYJn6Meuh1z34ZH+UwvLcNQsnCHPYQ+Zw19BeF+ZhB21nRLeVA6agAA1bAlbZaGoGVOxHDllhla52BEBQVBVZV-4pczYQDelq9xHQ3Zg0GcUlk2EZ31KhYxjWDYth2N08W11UKokB2EDASBGHpLxfFQEEeD2n6qlqepGku7peiYWDhmTeYAUT70uidowXbdmFkymJcvd49ZNm2bg9gAWkDmwdaQPXZC9hTsfQLuRe4LuO7ULQnrtV7LFb1VLexppqhMjAu+wAAiVwl-5JeKRj4Al+Ifdzyzc6BYgVBjRDCBo6pBf9n+LvCGcffM03qlhFP0xz48Ler72G-9nv5cKv7kuf+h8JbH1fu-BkwAv4-zOH-f+5IP7PygOAi+sde7X07r-asx5bih1cMgkcb9UFQKtGyZQQddYFxYDIe6YttBHxhAJfA9R8A5DkC0YkqA6EMLCLCGQodCjqG8MGaAWgAJQGqHIDYys1IQGJEEOCNpDBEjeGGOOedEG+BYFQx22hnbeFdiAEYQCmIQGVuKRU9ErjVk8N4JgZiLEQCktY888pLEcjJDcSuNwK4RBBE-LRNJNHABYPGHkJRkxRmIOHSODBiHqOBAfeEDiFROKsdcc8tj7HmNSc4jJWY3FpI8SCHxIJiFMHEPfB26BarAHYGQNQwALDaCQC5VArh66SOkbI+RnwAb1GUd4VRVT8H4BzI05prS3wdzXP6YMoZBwQGHNoKJ70QBTATojFgfj5iBjmRuRZyzxCrKVioe6ZzbYRlAtwWpIVlBjPsBMlpbSOk7C6WYnpnRFEDJiEMzO98gA)

andrewbranch:
That’s... weird

jfet97:
[Here](https://www.typescriptlang.org/play?ts=5.2.2#code/C4TwDgpgBAchDuAlCwCuAnAdgFXBAPNlBAB7ASYAmAzlABQB0TAhugObUBcUmEAbhHQBKKAF4AfFGaYQk0VCKlyVWnQBQUKEwasO3XgPRqREqAEtMAM0FREGqAH5b97tJABuNWtCQFEasBithAAxgD26JT4AegWbAA0UKiYANaYYfCY4l4+0ABiYWH4ACLEZBQ0fgFyUADe9gDaAArmmFApECBhllXAALoO3ABE8AAWzOSGQ2XKlaVOAIxQ3AAMagC+nt54CitByGhYuJD4JTMVtNj+wOJ0wNzFJpIFRcXi2ZoA9J9QAHoOOR22CW8jgSBQGBweFOpSUF16t3uUEeYmehRK73s3z+DiAA) is a 17 loc repro.

andrewbranch:
I think this is exposing an inference/assignability bug, not a logical consequence of switching the parameter type in the conditional extends type. I notice that if you swap `infer R` in `NewReturnType` for `any` or `unknown`, T1 starts taking the true branch of the conditional. So it seems like `infer R` is making the input type fail the `extends` assignability check, not `...args: never`. That will need to be fixed first or as part of this PR.

jfet97:
@andrewbranch I'll tell you what I found out after some debugging of the code in [the playground](https://www.typescriptlang.org/play?ts=5.2.2#code/C4TwDgpgBAchDuAlCwCuAnAdgFXBAPNlBAB7ASYAmAzlABQB0TAhugObUBcUmEAbhHQBKKAF4AfFGaYQk0VCKlyVWnQBQUKEwasO3XgPRqREqAEtMAM0FREGqAH5b97tJABuNWtCQFEasBithAAxgD26JT4AegWbAA0UKiYANaYYfCY4l4+0ABiYWH4ACLEZBQ0fgFyUADe9gDaAArmmFApECBhllXAALoO3ABE8AAWzOSGQ2XKlaVOAIxQ3AAMagC+nt54CitByGhYuJD4JTMVtNj+wOJ0wNzFJpIFRcXi2ZoA9J9QAHoOOR22CW8jgSBQGBweFOpSUF16t3uUEeYmehRK73s3z+DiAA).

Inside `getConditionalType`, when we are applying the two versions of the conditional to `<D extends Record<string, unknown>>(t: D) => Foo<D>` we have:

1. With the classic definition of `ReturnType`: photo_2023-09-24_17-03-28

1. With the modified defition, the one that uses `never` as the argument's type: 
photo_2023-09-24_17-05-05

In both cases `R` is correctly inferred as `Foo<Record<string, unknown>>`, but what is failing is the `isTypeAssignableTo` check between `checkType` and `inferredExtendsType`. What is that check for?

> Return falseType for a definitely false extends check. We check an instantiations of the two types with type parameters mapped to the wildcard type, the most permissive instantiations possible (the wildcard type is assignable to and from all types). If those are not related, then no instantiations will be and we can just return the false branch type.

This is exactly what happens, because the `never` version always returns what is present in the false branch in this specific playground.

So when `compareSignaturesRelated` comes into play because of `isTypeAssignableTo` we have:

1. `any` version:
1.1 `source` is `<D extends Record<string, unknown>>(t: D): Foo<D>`
1.2 `target` is `(...args: any): Foo<Record<string, unknown>>`
1.3 `source` becomes `(t: any): Foo<any>` because of `instantiateSignatureInContextOf`
1.4 So it seems all fine and dandy :)

2. `never` version:
2.1 `source` is `<D extends Record<string, unknown>>(t: D): Foo<D>`
2.2 `target` is `(...args: never): Foo<Record<string, unknown>>`
2.3 `source` becomes `(t: never): Foo<never>` because of `instantiateSignatureInContextOf`
2.4 When params are checked `sourceType = [t: never]` and `targetType = never` are compared, and this comparison seems to fail

andrewbranch (changes requested):
Marking this as not currently mergeable based on my previous comment

Andarist:
I plan to get back to this soon-ish but I have a backlog of PR reviews to address right now so it will take some time 😉

Andarist:
@jakebailey @andrewbranch could you rerun test suite here? :)

andrewbranch:
@typescript-bot test top200

jfet97:
Hi @andrewbranch , do you think this change can become part of TS `5.4`? The assignability problem has been solved by changing the definitions again (arguments/return type are being inferred even when not needed, then discarded), and there are no more failed tests.

andrewbranch:
Why are the `infer _` needed?

jfet97:
> Why are the `infer _` needed?

#56919 and [this comment](https://github.com/microsoft/TypeScript/pull/55714#issuecomment-1732597254) explain what happens if you don't use it.

andrewbranch:
@typescript-bot perf test

sandersn:
I'm trying to understand what state this PR is in. Here's my attempted summary:
- the original fix does indeed fix the bug with never-typed parameters causing ReturnType to produce `any`.
- but in one corner case, it hits another bug, #56919 
- necessitating an `infer _` workaround.

And we left off reviewing it at this point, without deciding whether `infer _` is an OK workaround, or whether this PR should wait until #56919 is fixed.

Is that accurate?

jfet97:
> I'm trying to understand what state this PR is in. Here's my attempted summary:
> 
> - the original fix does indeed fix the bug with never-typed parameters causing ReturnType to produce `any`.
> 
> - but in one corner case, it hits another bug, #56919 
> 
> - necessitating an `infer _` workaround.
> 
> 
> 
> And we left off reviewing it at this point, without deciding whether `infer _` is an OK workaround, or whether this PR should wait until #56919 is fixed.
> 
> 
> 
> Is that accurate?

The infer trick was really promising, because with a concrete type (like any or never) it seemed to me that you can always come up with something that fails the extends check.

Andarist:
This is not a good approach given the type parameters might be invariant and if a type parameter of a generic signature gets instantiated with `never` then usually this breaks things. 

Fixing the original issue would, at the very least, require addressing https://github.com/microsoft/TypeScript/issues/62712

Issue #55525, opened 2023-08-27
Typescript issue with includes in tsconfig


alainbelez:
### 🔎 Search Terms

tsconfig includes

### 🕗 Version & Regression Information

Version: 5.2.2/ 5.1.6

### 💻 Code

![image](https://github.com/microsoft/TypeScript/assets/6648110/68d6a5b8-b813-4d29-a5ec-de505e70f847)

### 🙁 Actual behavior

I created a new file that is included in the includes in tsconfig, but it keeps on nagging me that the file 

`File '/home/path/to/file/resources/js/views/TestNewView.vue' is not listed within the file list of project '/home/path/to/file/tsconfig.app.json'. Projects must list all files or use an 'include' pattern`

### 🙂 Expected behavior

there shouldn't be any error since the pattern is included already

### Additional information about the issue

 restarting vscode seems to fix the issue, but will occur again for other new files

andrewbranch:
Does this happen for `.ts` files or just for `.vue` files?

alainbelez:
it seems to only happen with `.vue` files only

sheetalkamat:
Can you please share tsserver verbose log for the same. 

You can upload logs by following these steps:

1. Set `"typescript.tsserver.log": "verbose"`
2. Restart vscode or vscode insiders
3. Reproduce the problem
4. In vscode or vscode insiders, run the `TypeScript: Open TS Server log` command
5. This should open a large file called `tsserver.log`. Please share this file.

alainbelez:
here you go. thanks

[tsserver.log](https://github.com/microsoft/TypeScript/files/12564461/tsserver.log)

PR #55716, opened 2023-09-11
Fix issue with wildcard with supported extensions when plugins add external files and override getScriptKind to add script kinds for the additional extensions

sheetalkamat:
When file is added to directory watched for wild cards, we check if the file is of supported extension. Here we dont want to match against include pattern because in most cases it doesnt include extension and we would re-evaluate root file names too many times if we did that. But with this if plugin adds extra file extension support (eg vue plugin) we ignore those files. This adds additional check to see if the script kind for the file is supported.

Fixes https://github.com/microsoft/TypeScript/issues/55525

### DIFF
--- a/src/compiler/watchUtilities.ts
+++ b/src/compiler/watchUtilities.ts
@@ -20,9 +20,11 @@ import {
     FileWatcherCallback,
     FileWatcherEventKind,
     find,
+    getAllowJSCompilerOption,
     getBaseFileName,
     getDirectoryPath,
     getNormalizedAbsolutePath,
+    getResolveJsonModule,
     hasExtension,
     identity,
     insertSorted,
@@ -44,6 +46,7 @@ import {
     removeIgnoredPath,
     returnNoopFileWatcher,
     returnTrue,
+    ScriptKind,
     setSysLog,
     SortedArray,
     SortedReadonlyArray,
@@ -563,6 +566,7 @@ export interface IsIgnoredFileFromWildCardWatchingInput {
     useCaseSensitiveFileNames: boolean;
     writeLog: (s: string) => void;
     toPath: (fileName: string) => Path;
+    getScriptKind?: (fileName: string) => ScriptKind;
 }
 /** @internal */
 export function isIgnoredFileFromWildCardWatching({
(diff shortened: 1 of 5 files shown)


src/server/project.ts
```diff
@@ -663,7 +663,7 @@ export abstract class Project implements LanguageServiceHost, ModuleResolutionHo
     }
 
     getScriptKind(fileName: string) {
-        const info = this.getOrCreateScriptInfoAndAttachToProject(fileName);
+        const info = this.projectService.getScriptInfoForPath(this.toPath(fileName));
```
sheetalkamat, line 666: This ensures that we arent adding script info and attaching to project when say watch invokes with say "file.js"

src/compiler/watchUtilities.ts
```diff
@@ -634,6 +643,25 @@ export function isIgnoredFileFromWildCardWatching({
             builderProgram.getState().fileInfos.has(file) :
             !!find(program as readonly string[], rootFile => toPath(rootFile) === file);
     }
+
+    function isSupportedScriptKind() {
+        if (!getScriptKind) return false;
```
jakebailey, line 648: Should this fall back to `getScriptKindFromFileName` or something? If it's not provided, it seems like this returning false would make `isIgnoredFileFromWildCardWatching` always return true where it wouldn't have before?

(Doing that locally, nothing changed, so, maybe I don't know what I'm talking about)

sheetalkamat: File extensions are always checked in is supported source name so it should have worked there so this is fallback so extensions’s additional answers where file is treated as ts/js etc are handled

jakebailey (approved):

johnsoncodehk:
Thanks for this fix! I think there is a very specific edge case, when the TS plugin proxy getScriptKind may cause this line to crash.

https://github.com/microsoft/TypeScript/blob/61a96b1641abe24c4adc3633eb936df89eb991f2/src/server/project.ts#L1516

Here are the conditions to reproduce:

1. Clone https://github.com/withastro/astro then execute `pnpm install`
2. Install the [Astro extension](https://marketplace.visualstudio.com/items?itemName=astro-build.astro-vscode) in VSCode
3. Open https://github.com/withastro/astro/blob/main/packages/astro/client.d.ts
4. Execute the `Organize Imports` command, then VSCode reports: `<semantic> TypeScript Server Error (5.3.0-beta) Debug Failure. False expression. Error: Debug Failure. False expression. at InferredProject2.updateGraphWorker` ([tsserver.log](https://github.com/microsoft/TypeScript/files/12841790/tsserver.log))

If you need me to open an issue, please let me know!

sheetalkamat:
This looks like issue with one of the plugin in your chain where LanguageService used program, that was not updated through Project and hence the state of project is out of sync with LS

PR #54242, opened 2023-05-13
feat(53656): Add support for the updated import attributes proposal

a-tarasyuk:
Fixes #53656

PR #55725, opened 2023-09-12
Support resolution-mode overrides in type-only constructs in all moduleResolution modes

### LINKED ISSUE #55579: Improve portability of resolution-mode assertions automatically emitted in `nodenext`
In the last week, I’ve gotten two separate requests for help with solving an issue where a resolution-mode assertion in a triple-slash reference:

```ts
/// <reference types="chai" resolution-mode="require" />
```

was automatically inserted into a declaration file. The implementation file was an ES module that referenced a global defined in `chai`, which only ships as a CommonJS module. Presumably, the resolution-mode assertion gets added in declaration emit to increase the likelihood that, if the user installs a version of `chai` that contains separate ESM and CJS types, the ESM-format .d.ts file here will still resolve to the CJS types, as it did when it was created by the library’s compilation. (Note that if the user does have the same version of `chai` that the library had during compilation, the resolution-mode assertion will not actually be needed.)

The reason this was an issue for the two folks who reached out to me was that some of their users  who consumed these declaration files were using `--moduleResolution bundler` or maybe even `--moduleResolution node10`, and consequently got the error:

```
error TS1452: 'resolution-mode' assertions are only supported when `moduleResolution` is `node16` or `nodenext`.
```

I feel pretty confident that non-node16/nodenext modes could simply ignore these assertions when they occur in declaration files. While we can’t guarantee that users in other modes will resolve these references the same way the library did, this is no different from any `import` statement the user wrote, or indeed any `import()` type automatically inserted by the declaration emitter.

Maybe better, we could also consider supporting resolution-mode assertions and the way they affect conditional package.json `"exports"` lookups in every `moduleResolution` mode in type reference directive lookups. They don’t reflect a runtime module resolution feature, so it feels like all we should care about is doing our best to resolve to the same global types in every `moduleResolution` mode.

It should be noted that in the meantime, a workaround for library authors is to manually write a `/// <reference types="chai" />` with no resolution-mode assertion in the implementation file that generated the reference that caused the problem.

andrewbranch:
This PR + the passage of time fixes #55579

Affects the following syntaxes:

- triple-slash references: `<reference types="foo" resolution-mode="require" />`
- `import()` types: `type Foo = typeof import("foo", { assert: { "resolution-mode": "import" } })`
- type-only imports: `import type foo from "foo" assert { "resolution-mode": "require" }`

(The latter two are still errors outside of nightly due to the use of import assertions, but these examples will be replaceable with import attributes after #54242.) The semantics are changed as follows in `moduleResolution` modes:

- `node16`/`nodenext`: no changes
- `bundler` and `node10`:
  - no longer errors
  - the presence of an override enables `imports`/`exports` resolution even if disabled for normal imports via `node10` or `resolvePackageJsonImports`/`resolvePackageJsonExports`, since these are intended to signal information about the runtime module resolver, and a resolution-mode override is a TypeScript-only construct
  - the presence of an override affects matching conditions for `imports`/`exports` resolution
- `classic`:
  - no longer errors

Notes:

- There’s a tension between wanting all TS-only resolutions to behave similarly to each other across `moduleResolution` modes and wanting those resolutions to behave similarly to their non-TS-only resolutions in the same `moduleResolution` mode. While it sounds reasonable to propose “triple-slash references should resolve the same in every `moduleResolution` mode,” it gets uglier when thinking about type-only imports:

  ```ts
  import type {} from "./foo" assert { "resolution-mode": "import" }
  ```

  This is an error in `nodenext` since ESM-mode resolutions require file extensions (`"./foo.js"`). It makes sense for this resolution to behave how a runtime import would. But in `--moduleResolution bundler`, file extensions are never required, so by the same logic, this import should work under that mode. So, I made the choice to vary the resolution algorithm for each of the supporting syntaxes only by forcing support for `imports`/`exports` and setting the appropriate conditions—in other words, `"resolution-mode": "import"` doesn’t imply _Node.js_ ESM resolution outside of `node16`/`nodenext`.

- `classic` doesn’t resolve node_modules in the first place (except for `@types`) so it doesn’t seem very compelling to try to add support for resolution-mode assertions there, so I opted to just remove the error and ignore them.


src/compiler/moduleNameResolver.ts
```diff
@@ -1411,13 +1416,13 @@ export function resolveModuleName(moduleName: string, containingFile: string, co
...
             case ModuleResolutionKind.Node10:
-                result = nodeModuleNameResolver(moduleName, containingFile, compilerOptions, host, cache, redirectedReference);
+                result = nodeModuleNameResolver(moduleName, containingFile, compilerOptions, host, cache, redirectedReference, resolutionMode);
                 break;
             case ModuleResolutionKind.Classic:
                 result = classicNameResolver(moduleName, containingFile, compilerOptions, host, cache, redirectedReference);
                 break;
             case ModuleResolutionKind.Bundler:
-                result = bundlerModuleNameResolver(moduleName, containingFile, compilerOptions, host, cache, redirectedReference);
+                result = bundlerModuleNameResolver(moduleName, containingFile, compilerOptions, host, cache, redirectedReference, resolutionMode);
```
andrewbranch, line 1425: In `node16`/`nodenext`, this `resolutionMode` is always set. In `bundler` and `node10`, it’s only set when the syntax has an explicit override.

It feels like it would be much nicer to have access to the syntax that triggered the request, or at least a serialization of all the import attributes, as part of this API. I can imagine other resolution-affecting options in import attributes in the future. But this change was possible without an API change. Open to doing something different though.

src/compiler/moduleNameResolver.ts
```diff
@@ -1750,20 +1762,35 @@ export function nodeModuleNameResolver(moduleName: string, containingFile: strin
...
+    features: NodeResolutionFeatures,
+    moduleName: string,
+    containingDirectory: string,
+    compilerOptions: CompilerOptions,
+    host: ModuleResolutionHost,
+    cache: ModuleResolutionCache | undefined,
+    extensions: Extensions,
+    isConfigLookup: boolean,
+    redirectedReference: ResolvedProjectReference | undefined,
+    overrideResolutionMode: ResolutionMode,
```
weswigham, line 1783: Maybe instead of `overrideResolutionMode`, we let the caller pass in a custom `conditions` array? Since generating that is all the resolution mode is actually used for, looks like (the features flags handle most other things you'd think the resolution mode may be used for). That'll be more flexible for API consumers, too.

andrewbranch: I’m fine with that as the API for `nodeModuleNameResolverWorker` (and maybe even the non-`worker` entrypoints), but it doesn’t feel right for `resolveModuleName`. What do you think about `attributes?: { "resolution-mode"?: "import" | "require" }` as a parameter? We could also make it an internal overload of the public functions if we’re not sure.

weswigham: Doesn't `resolveModuleName` already have a `resolutionMode` parameter? Would that replace it?

src/compiler/moduleNameResolver.ts
```diff
@@ -1750,20 +1772,38 @@ export function nodeModuleNameResolver(moduleName: string, containingFile: strin
...
+    features: NodeResolutionFeatures,
+    moduleName: string,
+    containingDirectory: string,
+    compilerOptions: CompilerOptions,
+    host: ModuleResolutionHost,
+    cache: ModuleResolutionCache | undefined,
+    extensions: Extensions,
+    isConfigLookup: boolean,
+    redirectedReference: ResolvedProjectReference | undefined,
+    conditions?: readonly string[],
```
sheetalkamat, line 1794: Can we make this non optional in all the workers/Internal methods or signatures to ensure we are not ignoring any params anywhere.

src/compiler/moduleNameResolver.ts
```diff
@@ -1774,19 +1814,18 @@ function nodeModuleNameResolverWorker(features: NodeResolutionFeatures, moduleNa
...
+        conditions: conditions ?? emptyArray,
         requestContainingDirectory: containingDirectory,
         reportDiagnostic: diag => void diagnostics.push(diag),
         isConfigLookup,
         candidateIsFromPackageJsonField: false,
     };
-
-    if (traceEnabled && moduleResolutionSupportsPackageJsonExportsAndImports(getEmitModuleResolutionKind(compilerOptions))) {
-        trace(host, Diagnostics.Resolving_in_0_mode_with_conditions_1, features & NodeResolutionFeatures.EsmMode ? "ESM" : "CJS", conditions.map(c => `'${c}'`).join(", "));
+    if (traceEnabled && moduleResolutionSupportsPackageJsonExportsAndImports(moduleResolution)) {
```
sheetalkamat, line 1823: Is this enough or do we need to check if conditions array is nonempty as well ?

andrewbranch: This is correct since `default` always matches, rather than being a real condition. Also, imports/exports don’t have to specify conditions, they can just point subpaths to files unconditionally.

src/compiler/moduleNameResolver.ts
andrewbranch: The exported functions with internal overloads already have many optional parameters in their public overload, so this won’t do anything there. Anywhere we call them with fewer parameters would just use the public overload. But I double checked them and they’re only called in tests.

sheetalkamat (approved):

src/compiler/moduleNameResolver.ts
```diff
@@ -1411,13 +1430,13 @@ export function resolveModuleName(moduleName: string, containingFile: string, co
                 result = nodeNextModuleNameResolver(moduleName, containingFile, compilerOptions, host, cache, redirectedReference, resolutionMode);
                 break;
             case ModuleResolutionKind.Node10:
-                result = nodeModuleNameResolver(moduleName, containingFile, compilerOptions, host, cache, redirectedReference);
+                result = nodeModuleNameResolver(moduleName, containingFile, compilerOptions, host, cache, redirectedReference, resolutionMode ? getConditions(compilerOptions, resolutionMode === ModuleKind.ESNext) : undefined);
```
weswigham, line 1433: Do we want to consider pushing this conditions lookup out to the caller? Or is it too useful a shorthand to take a `resolutionMode` directly on `resolveModuleName` as meaning a lookup type override for node10/bundler? I'm just thinking that if we push it out to the caller, we can pass the `conditions` argument through to all resolvers (and not just the `node10` and `bundler` ones), which'd be nice for API consumers of the resolver who are running funky custom import/import condition structures, but might be bad for our existing resolution cache APIs (namely, they'd have to go from being mode-aware to being condition-aware). It's probably a good compromise to leave it as-is, messing with the resolution cache's API is a lot of surface area to change just to make an edgecase API use nicer.

weswigham (approved):

Ciornenichii (commented):
Ciornenichii

src/compiler/moduleNameResolver.ts
```diff
@@ -556,18 +556,21 @@ export function resolveTypeReferenceDirective(typeReferenceDirectiveName: string
 
     const failedLookupLocations: string[] = [];
     const affectingLocations: string[] = [];
+    // Allow type reference directives to opt into `exports` resolution in any resolution mode
+    // when a `resolution-mode` override is present.
     let features = getNodeResolutionFeatures(options);
-    // Unlike `import` statements, whose mode-calculating APIs are all guaranteed to return `undefined` if we're in an un-mode-ed module resolution
-    // setting, type references will return their target mode regardless of options because of how the parser works, so we guard against the mode being
```
Ciornenichii, line 561: Paypal

Ciornenichii: Paypal.com

PR #55727, opened 2023-09-13
Avoid unnecessary resolution-mode assertion in declaration emit

andrewbranch:
Fixes the bug part of #55579, where the manual triple-slash reference workaround doesn’t work for @types/node.

When a type reference directive is processed, we also record what additional files the resolved types bring in via `<reference path="..." />` directives. This way, when you reference a global declared in `node_modules/@types/node/globals.d.ts`, we can see that it actually came in via a `<reference types="node" />`. But instead of recording this entry with the resolution-mode that the _types_ reference had, we recorded it with the resolution-mode that the _file_ reference had, which is irrelevant (and could even lead to some totally incorrect results with some manufactured bizarre cases). All we need to know is _how did this file enter the program?_ which is the type reference directive with whatever mode it had.

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -47735,11 +47735,14 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         function addReferencedFilesToTypeDirective(file: SourceFile, key: string, mode: ResolutionMode) {
             if (fileToDirective.has(file.path)) return;
             fileToDirective.set(file.path, [key, mode]);
-            for (const { fileName, resolutionMode } of file.referencedFiles) {
+            for (const { fileName } of file.referencedFiles) {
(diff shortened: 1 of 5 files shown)


andrewbranch:
@mpodwysocki I was never able to repro a case where the declaration emitter included a _duplicate_ triple-slash directive, so I can’t be sure whether this fixes the whole problem. Would you mind trying this build out on your repro?

@typescript-bot pack this

sheetalkamat (approved):

weswigham (approved):

mpodwysocki:
> @mpodwysocki I was never able to repro a case where the declaration emitter included a _duplicate_ triple-slash directive, so I can’t be sure whether this fixes the whole problem. Would you mind trying this build out on your repro?
> 
> @typescript-bot pack this

@andrewbranch it works perfectly, thank you!

PR #55728, opened 2023-09-13
Remove stableSort, rename sort to toSorted

jakebailey:
See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#sort_stability

Also, `toSorted` seems to be sightly slower than `arr.slice().sort()`; was hoping to also rename `sort` to `toSorted` and conditionally use it, but probably not worth it yet.


jakebailey:

fatcerberus:
> perf test this bun

yeah i'm sorry to tell you but your buns are performing poorly.  they're probably overcooked so people hate them

DanielRosenwasser (approved):

src/compiler/core.ts
```diff
@@ -1122,17 +1122,6 @@ export function* arrayReverseIterator<T>(array: readonly T[]) {
     }
 }
 
-/**
- * Stable sort of an array. Elements equal to each other maintain their relative position in the array.
- *
- * @internal
- */
-export function stableSort<T>(array: readonly T[], comparer: Comparer<T>): SortedReadonlyArray<T> {
```
jakebailey, line 1130: I should note that there's a behavior change with this PR; sort is:

```
export function sort<T>(array: readonly T[], comparer?: Comparer<T>): SortedReadonlyArray<T> {
    return (array.length === 0 ? array : array.slice().sort(comparer)) as SortedReadonlyArray<T>;
}
```

So... if you happen to sort an empty array, you don't make a copy! This is scary but nothing fails either way and I don't know how to feel about that. This feels like a footgun that kinda makes me scared of this PR, but no more scared than I already am about using `sort` at all.

DanielRosenwasser: Does anything break if you replace it with

```ts
export function sort<T>(array: readonly T[], comparer?: Comparer<T>): SortedReadonlyArray<T> {
    return (array.length === 0 ? Object.freeze([]) : array.slice().sort(comparer)) as SortedReadonlyArray<T>;
}
```

jakebailey: I'll check, though that does make me want to freeze `emptyArray` and then return that instead.

jakebailey: At that point though, there's not really a reason not to slice the array if we know it's empty...

jakebailey: Oops, sorry, no, your code fails lots of tests. Your code freezes the input array so breaks callers who are just sorting lists they have.

jakebailey: So now that I realize that the return result is readonly, I'm no longer concerned about the safety of this PR. I will send another PR which tries out freezing things, though, as if its perf is fine that's nice to prevent oopses.

src/compiler/core.ts
```diff
@@ -1112,7 +1112,7 @@ function stableSortIndices<T>(array: readonly T[], indices: number[], comparer:
  * @internal
  */
 export function sort<T>(array: readonly T[], comparer?: Comparer<T>): SortedReadonlyArray<T> {
```
jakebailey, line 1114: If I rename this to `toSorted`, would that be preferred? This is really just a polyfill for `toSorted` that appears to be slightly faster (for now).

DanielRosenwasser: sure

src/compiler/core.ts
```diff
@@ -835,7 +835,7 @@ export function sortAndDeduplicate<T>(array: readonly string[]): SortedReadonlyA
 export function sortAndDeduplicate<T>(array: readonly T[], comparer: Comparer<T>, equalityComparer?: EqualityComparer<T>): SortedReadonlyArray<T>;
 /** @internal */
 export function sortAndDeduplicate<T>(array: readonly T[], comparer?: Comparer<T>, equalityComparer?: EqualityComparer<T>): SortedReadonlyArray<T> {
-    return deduplicateSorted(sort(array, comparer), equalityComparer || comparer || compareStringsCaseSensitive as any as Comparer<T>);
+    return deduplicateSorted(toSorted(array, comparer), equalityComparer || comparer || compareStringsCaseSensitive as any as Comparer<T>);
```
sheetalkamat, line 838: I have always wondered, given `deduplicateSorted` used only by `sortAndDeduplicate` can we mark it to make readyOnly to ReadWrite array when length is not zero and work on that array itself instead of creating new array.

jakebailey:

jakebailey:
https://github.com/mozilla/rhino/issues/1151 is now closed; rhino now has a stable sort. Is this safe to merge or should we wait a while?

PR #55739, opened 2023-09-13
Make JSDoc skipping public, plus more configurable

jakebailey:
For #52959.

This is a follow-up to #52921.

This changes the way we do the new JSDoc skipping, instead replacing it with a publicly-available option which lets you control how much info you want. Specifically, you can specify:

- `JSDocParsingMode.ParseAll`, the default, which is to keep parsing JSDoc and including it in the tree. This is the default used by `tsserver`, `services`, etc, where we definitely need all of that info.
- `JSDocParsingMode.ParseForTypeErrors`, which only parses JSDoc tags which have a _semantic_ meaning to the checker which impact error calculation, that is, any JSDoc in a non-TS/TSX file, or any JSDoc containing `@see` or `@link`. This is the new default when calling from `tsc`, and would be a good default for anyone orchestrating TS builds, e.g. heft, nx.
- `JSDocParsingMode.ParseForTypeInfo`, if you only _query_ type info, but never care about its errors. This parses JSDoc in JS files, but doesn't in TS files at all (as they are never used for type info there). This is good for something like ts-eslint with type info (which does not surface errors).
- `JSDocParsingMode.ParseNone`, which always skips JSDoc, no matter what. This is good for those who _only_ want the parser, e.g. formatters like prettier, ts-eslint without type information, `transpileModule`, maybe others. Set this at your own risk.

To use this when calling `createSourceFile`, set `jsDocParsingMode` on the create options:

```ts
const sourceFile = ts.createSourceFile(
    filename,
    content,
    { languageVersion: ts.ScriptTarget.ESNext, jsDocParsingMode: ts.JSDocParsingMode.ParseNone },
);
```

To use this when constructing a host, set `jsDocParsingMode`:

```ts
const host = ts.createCompilerHost(options);
host.jsDocParsingMode = ts.JSDocParsingMode.ParseForTypeInfo;
// or
const host = ts.createIncrementalCompilerHost(options);
host.jsDocParsingMode = ts.JSDocParsingMode.ParseForTypeErrors;
// or
const host = ts.createWatchCompilerHost(...);
host.jsDocParsingMode = ts.JSDocParsingMode.ParseForTypeErrors;
```

FYI @fisker @bradzacher @JoshuaKGoldberg @JamesHenry @dmichon-msft


jakebailey:
Bleh, there's that dprint ternary annoyance again that Sheetal noted (https://github.com/microsoft/TypeScript/actions/runs/6178742090/job/16772520512?pr=55739), maybe that's https://github.com/dprint/dprint-plugin-typescript/pull/450 ?

bradzacher:
Copying from the other PR

> I think my main concern is that people will want to read out the .jsDoc properties by asking typescript-eslint for the original AST nodes (e.g. #55139), and then find that the info was never populated, or similar. I also wanted to try and make .jsDoc lazy and only create it when accessed (since we have the text anyway), but since JSDoc is bound by the binder, you can't really do that.


I think that this is going to be the rare case, if it even exists across the ecosystem.
If this is a configurable thing then we can easily set it up to be configurable on our side too - eg a user can pass `parserOptions.tsJSdoc` or something in their eslint config if a plugin needs it (and plugins can add it to their recommended config).

The vast majority of the ecosystem would just get faster-by-default lint runs which would be a big win for everyone!

jakebailey:
On that front, I probably need to figure out a way to plumb this through ProjectService too.

jakebailey:
Marking as ready for review, though I need to double check that I have plumbed this up through `createProgram` and so on.

jakebailey:
It turns out that I missed a lot of places that should plumb this. This will be tricky. ☹️

src/services/types.ts
```diff
@@ -427,6 +428,8 @@ export interface LanguageServiceHost extends GetEffectiveTypeRootsHost, MinimalR
     getParsedCommandLine?(fileName: string): ParsedCommandLine | undefined;
     /** @internal */ onReleaseParsedCommandLine?(configFileName: string, oldResolvedRef: ResolvedProjectReference | undefined, optionOptions: CompilerOptions): void;
     /** @internal */ getIncompleteCompletionsCache?(): IncompleteCompletionsCache;
+
+    getJSDocParsingKind?(): JSDocParsingKind | undefined;
```
jakebailey, line 432: After doing more plumbing I feel like I should leave this as a plain property rather than a getter; theoretically someone should only be setting this once.

sheetalkamat: Agree this should be property. Also I think you want this to be passed when creating documentRegistry.  Otherwise there is mismatch between these that you need to handle for releasing the documents.

jakebailey: Yep, working on that one now! It's pretty hard to plumb when it's an optional :(

src/executeCommandLine/executeCommandLine.ts
```diff
@@ -871,8 +872,8 @@ function performBuild(
         reportDiagnostic,
         createBuilderStatusReporter(sys, shouldBePretty(sys, buildOptions)),
         createReportErrorSummary(sys, buildOptions),
+        jsDocParsingMode,
     );
-    buildHost.skipNonSemanticJSDocParsing = skipNonSemanticJSDocParsing;
```
jakebailey, line 875: Through the process of ensuring I plumbed this option properly everywhere, this (and other assignments) turned into a parameter on the create function. I'm not sure which is better, but I feel like providing it up front is clearer and matches what we do for `createIncrementalCompilerHost`, `createCompilerHostWorker`, etc, and is likely more discoverable than having to patch the host afterwards. But, I'm not picky here.

tests/baselines/reference/api/typescript.d.ts
```diff
@@ -10086,8 +10096,8 @@ declare namespace ts {
...
      */
     function createBuilderStatusReporter(system: System, pretty?: boolean): DiagnosticReporter;
-    function createSolutionBuilderHost<T extends BuilderProgram = EmitAndSemanticDiagnosticsBuilderProgram>(system?: System, createProgram?: CreateProgram<T>, reportDiagnostic?: DiagnosticReporter, reportSolutionBuilderStatus?: DiagnosticReporter, reportErrorSummary?: ReportEmitErrorSummary): SolutionBuilderHost<T>;
-    function createSolutionBuilderWithWatchHost<T extends BuilderProgram = EmitAndSemanticDiagnosticsBuilderProgram>(system?: System, createProgram?: CreateProgram<T>, reportDiagnostic?: DiagnosticReporter, reportSolutionBuilderStatus?: DiagnosticReporter, reportWatchStatus?: WatchStatusReporter): SolutionBuilderWithWatchHost<T>;
+    function createSolutionBuilderHost<T extends BuilderProgram = EmitAndSemanticDiagnosticsBuilderProgram>(system?: System, createProgram?: CreateProgram<T>, reportDiagnostic?: DiagnosticReporter, reportSolutionBuilderStatus?: DiagnosticReporter, reportErrorSummary?: ReportEmitErrorSummary, jsDocParsingMode?: JSDocParsingMode): SolutionBuilderHost<T>;
```
jakebailey, line 10099: This signature is ridiculous.

src/compiler/tsbuildPublic.ts
```diff
@@ -302,8 +303,8 @@ export function createBuilderStatusReporter(system: System, pretty?: boolean): D
     };
 }
 
-function createSolutionBuilderHostBase<T extends BuilderProgram>(system: System, createProgram: CreateProgram<T> | undefined, reportDiagnostic?: DiagnosticReporter, reportSolutionBuilderStatus?: DiagnosticReporter) {
-    const host = createProgramHost(system, createProgram) as SolutionBuilderHostBase<T>;
+function createSolutionBuilderHostBase<T extends BuilderProgram>(system: System, createProgram: CreateProgram<T> | undefined, reportDiagnostic?: DiagnosticReporter, reportSolutionBuilderStatus?: DiagnosticReporter, jsDocParsingMode?: JSDocParsingMode) {
```
sheetalkamat, line 306: I think we shouldnt change these apis and nor should we add it to createProgramHost gien that its going be just property on host that can be set by API user if need arises.

src/compiler/program.ts
```diff
@@ -393,16 +394,16 @@ export function computeCommonSourceDirectoryOfFilenames(fileNames: readonly stri
     return getPathFromPathComponents(commonPathComponents);
 }
 
-export function createCompilerHost(options: CompilerOptions, setParentNodes?: boolean): CompilerHost {
-    return createCompilerHostWorker(options, setParentNodes);
+export function createCompilerHost(options: CompilerOptions, setParentNodes?: boolean, jsDocParsingMode?: JSDocParsingMode): CompilerHost {
```
sheetalkamat, line 397: I dont think we should be threading in this to defaults. User of this API can always call thius method and override getSourceFile on their own .. there is nothing special here.

src/compiler/watch.ts
```diff
@@ -829,7 +830,7 @@ export function setGetSourceFileAsHashVersioned(compilerHost: CompilerHost) {
  *
  * @internal
  */
-export function createProgramHost<T extends BuilderProgram = EmitAndSemanticDiagnosticsBuilderProgram>(system: System, createProgram: CreateProgram<T> | undefined): ProgramHost<T> {
+export function createProgramHost<T extends BuilderProgram = EmitAndSemanticDiagnosticsBuilderProgram>(system: System, createProgram: CreateProgram<T> | undefined, jsDocParsingMode: JSDocParsingMode | undefined): ProgramHost<T> {
```
sheetalkamat, line 833: we should not override this. User can set the property

src/compiler/watch.ts
```diff
@@ -851,15 +852,16 @@ export function createProgramHost<T extends BuilderProgram = EmitAndSemanticDiag
...
         now: maybeBind(system, system.now),
+        jsDocParsingMode,
     };
 }
 
 /**
  * Creates the watch compiler host that can be extended with config file or root file names and options host
  */
-function createWatchCompilerHost<T extends BuilderProgram = EmitAndSemanticDiagnosticsBuilderProgram>(system = sys, createProgram: CreateProgram<T> | undefined, reportDiagnostic: DiagnosticReporter, reportWatchStatus?: WatchStatusReporter): WatchCompilerHost<T> {
+function createWatchCompilerHost<T extends BuilderProgram = EmitAndSemanticDiagnosticsBuilderProgram>(system = sys, createProgram: CreateProgram<T> | undefined, reportDiagnostic: DiagnosticReporter, reportWatchStatus?: WatchStatusReporter, jsDocParsingMode?: JSDocParsingMode): WatchCompilerHost<T> {
```
sheetalkamat, line 862: Again should not have overload. user can call this and then set host.jsDocParsingMode

src/compiler/watchPublic.ts
```diff
@@ -118,11 +119,8 @@ export function readBuilderProgram(compilerOptions: CompilerOptions, host: ReadB
     return createBuilderProgramUsingProgramBuildInfo(buildInfo, buildInfoPath, host);
 }
 
-export function createIncrementalCompilerHost(options: CompilerOptions, system?: System): CompilerHost;
-/** @internal */
-export function createIncrementalCompilerHost(options: CompilerOptions, system?: System, skipNonSemanticJSDocParsing?: boolean): CompilerHost;
-export function createIncrementalCompilerHost(options: CompilerOptions, system = sys, skipNonSemanticJSDocParsing?: boolean): CompilerHost {
-    const host = createCompilerHostWorker(options, /*setParentNodes*/ undefined, skipNonSemanticJSDocParsing, system);
+export function createIncrementalCompilerHost(options: CompilerOptions, system = sys, jsDocParsingMode?: JSDocParsingMode): CompilerHost {
```
sheetalkamat, line 122: Again i dont think we should be exposing this externally unless we find users for the API.

src/compiler/watchPublic.ts
```diff
@@ -367,9 +360,9 @@ export interface WatchOfFilesAndCompilerOptions<T> extends Watch<T> {
...
+export function createWatchCompilerHost<T extends BuilderProgram>(configFileName: string, optionsToExtend: CompilerOptions | undefined, system: System, createProgram?: CreateProgram<T>, reportDiagnostic?: DiagnosticReporter, reportWatchStatus?: WatchStatusReporter, watchOptionsToExtend?: WatchOptions, extraFileExtensions?: readonly FileExtensionInfo[], jsDocParsingMode?: JSDocParsingMode): WatchCompilerHostOfConfigFile<T>;
+export function createWatchCompilerHost<T extends BuilderProgram>(rootFiles: string[], options: CompilerOptions, system: System, createProgram?: CreateProgram<T>, reportDiagnostic?: DiagnosticReporter, reportWatchStatus?: WatchStatusReporter, projectReferences?: readonly ProjectReference[], watchOptions?: WatchOptions, jsDocParsingMode?: JSDocParsingMode): WatchCompilerHostOfFilesAndCompilerOptions<T>;
```
sheetalkamat, line 364: Dont think we should do these API changes

src/services/services.ts
```diff
@@ -1362,7 +1363,7 @@ class SyntaxTreeCache {
                 ),
                 setExternalModuleIndicator: getSetExternalModuleIndicator(this.host.getCompilationSettings()),
             };
-            sourceFile = createLanguageServiceSourceFile(fileName, scriptSnapshot, options, version, /*setNodeParents*/ true, scriptKind);
+            sourceFile = createLanguageServiceSourceFile(fileName, scriptSnapshot, options, version, /*setNodeParents*/ true, scriptKind, this.host.jsDocParsingMode);
```
sheetalkamat, line 1366: Should this be "SkipAll" because this is syntax only file?

src/services/types.ts
```diff
@@ -427,6 +428,8 @@ export interface LanguageServiceHost extends GetEffectiveTypeRootsHost, MinimalR
     getParsedCommandLine?(fileName: string): ParsedCommandLine | undefined;
     /** @internal */ onReleaseParsedCommandLine?(configFileName: string, oldResolvedRef: ResolvedProjectReference | undefined, optionOptions: CompilerOptions): void;
     /** @internal */ getIncompleteCompletionsCache?(): IncompleteCompletionsCache;
+
+    jsDocParsingMode?: JSDocParsingMode;
```
sheetalkamat, line 432: Do we really need this apart from syntax where it seems like it should be "SkipAll"

sheetalkamat (changes requested):

src/compiler/tsbuildPublic.ts
jakebailey: Sure, I'll drop it where I can. I was not sure where we drew the line on parameters to constructors versus setting properties after the fact.

src/compiler/program.ts
jakebailey: I think that not providing this parameter will make it very inconvenient; this function and `createCompilerHostWorker` are what call `createGetSourceFile`, which is internal and non-trivial, and then wrap it again in another closure. Setting the JSDoc parsing mode is a lot like setting a default `setParentNodes` which this function already does.

jakebailey: See: https://github.com/typescript-eslint/typescript-eslint/blob/7826910d8c6ba4d2d40b8d55fcf7fd39fe290d88/packages/typescript-estree/src/create-program/useProvidedPrograms.ts#L84

src/compiler/watchPublic.ts
jakebailey: My same commment about `createCompilerHost` applies here; an example user is rushstack's `heft` which is another user I've been meaning to talk to about this: https://github.com/octogonz/rushstack/blob/15680b899a06c614c32088b02940bf676596bcf2/apps/heft/src/plugins/TypeScriptPlugin/TypeScriptBuilder.ts#L1079

tests/baselines/reference/api/typescript.d.ts
RyanCavanaugh: We changed `createProgram` to take an options bag when its parameter list got too long; can do that here too if you're inclined

src/services/types.ts
jakebailey: We do, because if someone is using the ProjectService (like ts-eslint), they need to pass ProjectService the option, which then passes it to its LanguageService instances, which then are the host for their own compilations. At least, that's what I'm seeing.

src/compiler/tsbuildPublic.ts
jakebailey: This one has been removed.

src/compiler/watch.ts
jakebailey: This one has been removed.

src/compiler/watch.ts
jakebailey: This one has been removed.

src/compiler/watchPublic.ts
jakebailey: This one has been removed.

src/services/services.ts
jakebailey: Good point, I think it may be able to!

src/compiler/watchPublic.ts
jakebailey: (they use both `createCompilerHost` and the incremental version)

src/services/services.ts
jakebailey: Hm, it fails tests; I remember now that we have syntactic classifications for JSDoc nodes, so that would break by default which we don't want.

tests/baselines/reference/api/typescript.d.ts
jakebailey: Not in this PR, anyway. This is harder than `createProgram` because that one used to accept an array as its first parameter, which made it very easy to distinguish the overloads, but this one takes `System`, which is also `object`.

src/executeCommandLine/executeCommandLine.ts
jakebailey: Okay, per PR feedback, these are back to being assignments.

src/executeCommandLine/executeCommandLine.ts
```diff
@@ -1014,6 +1015,7 @@ function createWatchOfConfigFile(
     updateWatchCompilationHost(system, cb, watchCompilerHost);
     watchCompilerHost.configFileParsingResult = configParseResult;
     watchCompilerHost.extendedConfigCache = extendedConfigCache;
+    watchCompilerHost.jsDocParsingMode = jsDocParsingMode;
```
jakebailey, line 1018: This (and one below) were missed in my last PR; I only found this by making the parameter required and bubbling that up to the top until I got everything compiling, then made it all optional again.

src/services/services.ts
jakebailey: So that being said, I think it's best to leave this as in the PR; I thought about reverting it back to the default (e.g. skip node), but my feeling is that anyone who configures this downstream will understand the implications. They can expect many other things to also be broken if they skip more than they need to function.

src/compiler/program.ts
jakebailey: Generally speaking I think this means that anywhere we can expect to find `setParentNodes`, we probably need to be expecting `jsDocParsingMode`.

src/compiler/watchPublic.ts
jakebailey: I'm also kinda shocked that this function doesn't also provide `setParentNodes`.

jakebailey: In fact, I'm inclined to add them just to make sure we end up with the params in the same order...

src/services/types.ts
sheetalkamat: No.. this does not pass to LS .. it will use from document registry

src/services/services.ts
sheetalkamat: I feel like we should leave this non configurable for now. And remove LS.jsDocParsingMode as its not used

src/services/types.ts
jakebailey: Ah, well, I just tried to remove it, and it just where `SyntaxTreeCache` gets its value. I suppose we could make this internal at that point.

src/services/services.ts
jakebailey: I guess I could remove it; I can actually already make it non-configurable by marking `jsDocParsingMode` internal on `LanguageServiceHost` (this thread turned out to be redundant with the other one). Which would you prefer?

src/executeCommandLine/executeCommandLine.ts
sheetalkamat: is this not handled by updateWatchCompilationHost above ?

jakebailey: Oh, they are, oops! I'll remove them.

src/compiler/types.ts
```diff
@@ -9850,6 +9850,12 @@ export const commentPragmas = {
     },
 } as const;
 
+export const enum JSDocParsingMode {
+    KeepAll = 0,
+    KeepSemanticOnly = 1,
+    SkipAll = 2,
```
jakebailey, line 9856: I feel like "Keep" may be a bad prefix here; "Keep" makes me feel like we're always parsing then _dropping_ from the AST, not always skipping. Perhaps it'd be better to have:

- `All`
- `SkipNonSemantic`
- `SkipAll`

? Anyone else want to paint the bikeshed?

sandersn: I like your new proposal, or

```
JSDoc.ParseAll
JSDoc.SkipNonSemantic
JSDoc.SkipAll
```

(assuming that JSDoc isn't already taken by another value -- I guess `JSDocMode` or `JSDocParsing` are OK too)

jakebailey: JSDoc is taken by the namespace. I was also trying to avoid it being "too short" because it _technically_ doesn't control JSDoc parsing under the hood at all. It's more like... `JSDocSkipMode`?

sandersn: `Semantic` is also a little fuzzy. Maybe `SkipNonChecked` or `ParseCheckedTagsOnly`

jakebailey: "Semantic" is the name we spitballed a day or two ago, and what I referred to it as before this PR. I am of course happy to come to a consensus on any new name because we'll be stuck with it forever!

jakebailey:
@typescript-bot pack this

jakebailey:
I plan to do some perf testing when I get a chance (next week...)

bradzacher:
Spitballing how we'd probably use this:

for all non-type-aware parses we'd use `JSDocParsingMode.SkipAll`.

for type-aware parses we'd use `JSDocParsingMode.KeepSemanticOnly`.
TBH - ideally we'd use `JSDocParsingMode.SkipAll` for TS files and `JSDocParsingMode.KeepSemanticOnly` for JS files so we can get the best of both worlds (i.e. we would be able to do some type-aware linting on JS files via the JSDoc but TS files would be super fast as they don't need any JSDoc).

Question: is it possible to enable that granularity? Or do you think "one value per host" is probably best for implementation simplicity's sake?

jakebailey:
I feel like there's some other value I could add to the option to make that work. Or, turn this back into a method and pass it the filename, but that sounds like overkill.

bradzacher:
I was thinking the same thing - a host option method that takes a filename would be the perfect solution but I do think it's getting towards the over-engineered territory.

I think it'd be worth testing the performance of the parser though. Test it on a big repo with skip and with semantic only and record the difference.
Even though it's complex it may just be worth it to improve the end-user experience.

I'm all for KISS but if we can shave a few seconds off a lint run it's defs worth it.

JamesHenry:
Oh this is cool! Thanks for the heads up @jakebailey

jakebailey:
It occurred to me that this could actually be 4 levels, if you think about it a different way:

- "All" - I need all JSDoc tags, e.g. editor, defaults.
- "I query semantic errors" - tsc, build orchestrators, so what KeepSemantic is right now. This parses all JSDoc in js files but only see/link in TS
- "I query type information" - ts-eslint with type information - JSDoc in js files, but not in TS files. We explicitly never use JSDoc for type info from TS files.
- "I don't care at all" - prettier, ts-eslint without type info, what SkipAll is now.

That seems like it'd cover all of the bases without a callback. Thoughts?

---

Also, does anyone use the Deprecated tag? Without parsing JSDoc, you won't get that anymore.

JoshuaKGoldberg:
For explicitness, just confirming: untyped lint rules such as `@typescript-eslint/no-unused-vars` would use _"I query semantic errors"_, right?

I'm surprised that only `@link` and `@see` are needed, but can't find evidence in any of the linked issues/PRs across eslint-plugin-jsdoc and typescript-eslint of other tags being needed...

Linking for easier reference: https://github.com/typescript-eslint/typescript-eslint/issues/5017 / https://github.com/typescript-eslint/typescript-eslint/issues/6967 

> Also, does anyone use the Deprecated tag? Without parsing JSDoc, you won't get that anymore.

Is https://github.com/gund/eslint-plugin-deprecation what you're looking for?

bradzacher:
That is actually a really good way to think about this, yea. I like that separation.

---

> Also, does anyone use the Deprecated tag? Without parsing JSDoc, you won't get that anymore.

I know of one rule (which we ourselves use in ts-eslint purely because it helps us with upstream TS deprecations)

https://github.com/gund/eslint-plugin-deprecation/blob/master/src/rules/deprecation.ts#L311-L318

Currently that uses it but i don't know if there's an explicit need to? If there was a simple method of grabbing the JSDoc comment node for a symbol then the rule would super easily be able to just do `if (comment.value.includes("@deprecated))` - no need for it to have complete JSDoc parsing semantics.

bradzacher:
> untyped lint rules such as @typescript-eslint/no-unused-vars would use "I query semantic errors", right?

No. Because they are purely ast-only rules.
No ast-only rules make any assumptions about even having access to the TS AST or parser services and operate wholly in the ESTree AST land.

---

> but can't find evidence in any of the linked issues/PRs across eslint-plugin-jsdoc and typescript-eslint of other tags being 

This is the whole point of this! Most rules - even type-aware ones - don't use any JSDoc information in TS files.

We never added APIs for it and people didn't want to go via the TS AST to get it. TBH there's also not really any usecases for it aside from the `@deprecated` tag.
`eslint-plugin-jsdoc` has its own JSDoc parser, ofc, because they are built to operate outside the TS world.

We have purposely never supported JSDoc in any of our rules because for the longest time non-type-aware rules weren't able to even access the TS AST. It's only a more recent change that we've exposed the TS AST to non-type-aware rules.

jakebailey:
Updated the PR to introduce one more mode, per the above description.

@typescript-bot pack this

JoshuaKGoldberg:
> TBH there's also not really any usecases for it aside from the @deprecated tag.

Maybe I'm cycling here: TypeScript itself supports `@link` to mark something as used [[playground with `@link`](https://www.typescriptlang.org/play?#code/C4TwDgpgBAaghgGwK4QKoDskGcIBMoC8UmAtgEYQBOA3AFC2iSyIqo75GkU30D0AVP1pR+UAN4ABBAEt0Aa2bI07AL7D+vWhAAeYAPaVgUAGZJ0AY2DS96KAj0BzABQA6N3EoOsALihm56HoA7ugA2gC6AJTiwlBx5jZYeggQLvbOHl6RdCpAA)]. So from my perspective, `@link` _is_ an AST-only concern and part of the TS AST.

src/compiler/types.ts
```diff
@@ -9850,6 +9850,13 @@ export const commentPragmas = {
     },
 } as const;
 
+export const enum JSDocParsingMode {
+    ParseAll,
+    ParseForTypeErrors,
```
jakebailey, line 9855: I haven't confirmed it yet, but I also think that you need to use this mode if you want to do emit, as `see/link` may impact import preservation. But, I might be wrong there and it'll require testing. It's possible that it doesn't matter, after all, I was able to make `transpileModule` use `ParseNone` without failures.

bradzacher:
> Maybe I'm cycling here: TypeScript itself supports @link to mark something as used [playground with @link]. So from my perspective, @link is an AST-only concern and part of the TS AST.

It depends on how you look at it, really.

You're thinking from the perspective of "typescript uses `@see` and `@link` within its `noUnusedLocals` checks, so ts-eslint should too".

But that's an insider take.
Most people don't use these JSDoc tags. TBH most people probably don't fully understand how to write them. (Even more TBH most codebases don't even write JSDoc and instead just dump some prose in a block comment)
So the real-world usecases of "I'm writing TS and I have an import that's only used within `@see`/`@link`" are really, really rare.

Additionally really the only usecase from a linting perspective is unused variable checking. The LSP also has the "go to definition" and "render link in hover text" usecases - but not for linting.

So pragmatically thinking - if barely anyone writes TS code that's only referenced inside JSDoc, then us building the complicated support for it into `no-unused-vars` (whilst correct in terms of "TS parity") isn't really worth it. And we've seen this in terms of the TS team's usecase being the first ask for this feature!

There's probably a little chicken/egg here - most tools don't support it well so most users don't learn/use it, so tools don't support it...... But it does seem like the ecosystem would be better off having faster parses, rather than slower parses to allow us to satisfy a <1% usecase that has other (sloooowww) workarounds.

/rant

jakebailey:
> There's probably a little chicken/egg here - most tools don't support it well so most users don't learn/use it, so tools don't support it...... But it does seem like the ecosystem would be better off having faster parses, rather than slower parses to allow us to satisfy a <1% usecase that has other (sloooowww) workarounds.

I haven't actually been able to benchmark ts-eslint with these options quite yet, but actually checking for see/link in the comment is pretty quick (string slicing is effectively free, as is a regex test on that slice), and if it's really that rare, I don't expect it to actually hurt to keep parsing. I personally don't see people starting to use it more just because a linter starts to support it.

I feel like the benefit of _not_ using the JSDoc plugin when TS can give it "for free" feels better, especially given the JSDoc plugins perf + the fact that the way you get it is to effectively run an unrelated rule and then _disable_ error reporting to get the unused marks as a side effect. But, I of course have an opinion on this already 😄

src/compiler/types.ts
jakebailey: Thankfully I believe this isn't true, I've tested `@link` and referencing an import doesn't mean the import is kept.

src/compiler/watchPublic.ts
jakebailey: My dreams here have been dashed a bit because `createCompilerHost` doesn't take `system` but `createIncrementalCompilerHost` does. I feel like that should be fixed too.

src/compiler/program.ts
sheetalkamat: Hmm.. Generally i like to go with conservative approach and avoid having to add additional surface area that we have to keep maintaining. This is ok i guess though , user can just create source file from file text easily to override but ok.

src/compiler/watchPublic.ts
sheetalkamat: I would keep createCompilerHost as is and add jsdoc thing like you have. As i said there this doesnt make me happy but its ok.

src/services/types.ts
sheetalkamat: We should remove this instead of making internal,

src/services/services.ts
sheetalkamat: We should make this non configurable and use whichever value is correct here.

jakebailey: Alrighty, I'll drop it!

src/compiler/types.ts
jakebailey: I've renamed things further; check the PR description. They're now close to your suggestion with the nuance about errors/no errors.

src/compiler/types.ts
```diff
@@ -9850,6 +9850,13 @@ export const commentPragmas = {
     },
 } as const;
 
+export const enum JSDocParsingMode {
+    ParseAll,
+    ParseNone,
+    ParseForTypeErrors,
+    ParseForTypeInfo,
```
jakebailey, line 9857: Maybe this should be `ParseforTypeQueries`? Any opinions?

src/compiler/parser.ts
```diff
@@ -1324,10 +1325,7 @@ function setExternalModuleIndicator(sourceFile: SourceFile) {
     sourceFile.externalModuleIndicator = isFileProbablyExternalModule(sourceFile);
 }
 
-export function createSourceFile(fileName: string, sourceText: string, languageVersionOrOptions: ScriptTarget | CreateSourceFileOptions, setParentNodes?: boolean, scriptKind?: ScriptKind): SourceFile;
-/** @internal */
-export function createSourceFile(fileName: string, sourceText: string, languageVersionOrOptions: ScriptTarget | CreateSourceFileOptions, setParentNodes?: boolean, scriptKind?: ScriptKind, skipNonSemanticJSDoc?: boolean): SourceFile;
-export function createSourceFile(fileName: string, sourceText: string, languageVersionOrOptions: ScriptTarget | CreateSourceFileOptions, setParentNodes = false, scriptKind?: ScriptKind, skipNonSemanticJSDoc?: boolean): SourceFile {
+export function createSourceFile(fileName: string, sourceText: string, languageVersionOrOptions: ScriptTarget | CreateSourceFileOptions, setParentNodes = false, scriptKind?: ScriptKind, jsDocParsingMode?: JSDocParsingMode): SourceFile {
```
weswigham, line 1328: Since we're changing it anyway, should we just fold `jsDocParsingMode` into `CreateSourceFileOptions`?

jakebailey: My original PR worked that way, yeah, I just found it a little irritating to have to `{ ...options, jsDocParsingMode }` all over the place in order to set it. Let me recheck how bad that is.

jakebailey: The difference is that `languageVersionOrOptions` is basically always set by the caller, whereas `setParentNodes` is always provided by the host more or less like `jsDocParsingMode`. But I also don't like how many functions have like 50 parameters 😄

src/compiler/program.ts
weswigham: If the option is fed in via the `CreateSourceFileOptions`, rather than a top-level field, callers will already have the opportunity to request it on a per-file basis when making the file, rather than needing to set it on the host as a whole.

src/compiler/scanner.ts
```diff
@@ -2288,6 +2289,28 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
...
+    function shouldParseJSDoc() {
+        switch (jsDocParsingMode) {
+            case JSDocParsingMode.ParseAll:
+                return true;
+            case JSDocParsingMode.ParseNone:
+                return false;
+        }
+
+        if (scriptKind !== ScriptKind.TS && scriptKind !== ScriptKind.TSX) {
+            // If outside of TS, we need JSDoc to get any type info.
```
weswigham, line 2301: `ScriptKind.JSON`? AFAIK none of the comments we may read in a JSON document can modify the expression types. `ScriptKind.External` is probably also safe to bail on.

jakebailey: I don't think we ever actually make it to this check at all for those file types, so I felt safe not checking for them (and choosing the backwards-compatible option). But, I'll verify.

src/compiler/scanner.ts
```diff
@@ -2003,7 +2004,7 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
                             }
                         }
 
-                        if (isJSDoc && (!skipNonSemanticJSDoc || semanticJSDocTagRegEx.test(text.slice(fullStartPos, pos)))) {
+                        if (isJSDoc && shouldParseJSDoc()) {
```
sandersn, line 2007: 👍🏼

src/services/documentRegistry.ts
```diff
@@ -309,7 +310,7 @@ export function createDocumentRegistryInternal(useCaseSensitiveFileNames?: boole
 
         if (!entry) {
             // Have never seen this file with these settings.  Create a new source file for it.
-            const sourceFile = createLanguageServiceSourceFile(fileName, scriptSnapshot, sourceFileOptions, version, /*setNodeParents*/ false, scriptKind);
+            const sourceFile = createLanguageServiceSourceFile(fileName, scriptSnapshot, sourceFileOptions, version, /*setNodeParents*/ false, scriptKind, jsDocParsingMode);
             if (externalCache) {
                 externalCache.setDocument(keyWithMode, path, sourceFile);
```
weswigham, line 315: `keyWithMode` should probably also consider `jsDocParsingMode`, so each different jsdoc parse mode gets a different LS document registry cache entry (rather than them randomly overwriting/taking priority over one another). `getDocumentRegistryBucketKeyWithMode` probably needs to be adjusted to take a full `CreateSourceFileOptions` object and work from that (assuming it gets a `jsDocParsingMode` member).

src/compiler/parser.ts
jakebailey: Hm, this gets very hairy. Providing an object instead of a version causes behavior changes, so I can't override it for everything. And if I don't override it and instead attempt to figure out the call chain up to who passes in the object or version in the first place, I'm forced to change the behavior and/or plumb it in an awkward, more-error-prone way...

bradzacher:
> string slicing is effectively free, as is a regex test on that slice

In isolation yeah but across an entire codebase doing these to every single comment string adds up to a non-trivial amount.

`no-unused-vars` is already one of the slower rules we have because it does so much work in order to do nothing most of the time - so we do want to be careful at the same time. Even adding 1ms per file adds up - at Canva just 1ms per file would add 60s to our lint run!

But also that massive codebase only has ~500 usages of `@see`/`@link` - and a lot of them aren't variable names, but are instead URLs.

I wouldn't be surprised if most non-library codebases are in exactly the same boat.

src/services/documentRegistry.ts
jakebailey: I was going to say that each registry has its own setting, so it shouldn't matter, but then I noticed `externalCache`, oops. Will fix.

src/compiler/parser.ts
weswigham: > Providing an object instead of a version causes behavior changes

Whu - it does? AFAIK `{ languageVersion: ts.ScriptTarget.Latest }` and `ts.ScriptTarget.Latest` as parameters should behave identically. What changes?

jakebailey: `setExternalModuleIndicator` and `impliedNodeFormat` differ, as of the PR where you introduced that stuff. It's possible that it doesn't matter but I think there's a reason we left the old signature behind, and it's still used heavily in services.

src/compiler/scanner.ts
jakebailey: Flipping this to be "=== JS, === JSX" doesn't change any tests, but truthfully I don't know when the others are even used. I would feel safest with the conservative I have now, but am not strongly attached if someone knows better.

src/services/documentRegistry.ts
sheetalkamat: You dont need to add jsDocParsingMode to key since externalCache is internal and operates with same jsDocParsingMode. So just a comment in editorServices should suffice and not add to the key

jakebailey: Ah, yeah, you're right; it's tied to the ProjectService which has a fixed JSDocParsingMode. I'll revert it.

sheetalkamat (approved):

jakebailey:
At this point, the only outstanding questions are:

- Should this be on `CreateSourceFileOptions` instead of a parameter?
- Are these modes enough, or does this need to be a host function so people can override this at-will?
- Are the modes named well?

Do we think that this PR is good enough as-is, or should I try and pull this discussion into a meeting?

sheetalkamat:
For the createSourceFileOptions, i prefer thing to be on host and as a separate parameter since its kind of similar to setParentNodes. And that also allows us to not have to pass "jsDocMode" in layers further down and have to worry about its value changes in different caches etc

jakebailey:
I will note that it is certainly possible to have it not as a parameter and stick it on the options bag, and then just override it everywhere I currently pass the flags in, it's just that doing so will create a bunch of extra objects and potentially change behavior if the caller didn't provide an object. I should test that "for sure", though.

src/compiler/parser.ts
jakebailey: I think I'm just totally wrong, apologies!!

https://github.com/microsoft/TypeScript/commit/12f59036b948188126e117ca7510703452462f4d shows that this would theoretically work, and I can use the current mechanism to set it as a default instead and not plumb it all the way out.

jakebailey:
Yeah, here is a commit which would move the option off params and onto the options bag by changing all callers to set it as the default: https://github.com/microsoft/TypeScript/commit/12f59036b948188126e117ca7510703452462f4d

I still prefer the parameter in this particular instance, though.

jakebailey:
Per design meeting feedback, which effectively said "we should stop making parameter count higher", I have moved this onto `CreateSourceFileOptions` + the `CompilerHost` suggestion from my WIP branch (plus more code to make that work). This leaves the signature of `createCompilerHost` unchanged, and one must set the prop after the fact like the other hosts.

I'm not totally confident that I got all of the places, though; I am going to locally remove the "old" signature without the options bag and see if I've forgotten something. Really, I think we should deprecate the old signature and go full-options-bag.

src/compiler/program.ts
```diff
@@ -482,7 +480,7 @@ export function createCompilerHostWorker(
     const newLine = getNewLineCharacter(options);
     const realpath = system.realpath && ((path: string) => system.realpath!(path));
     const compilerHost: CompilerHost = {
-        getSourceFile: createGetSourceFile(fileName => compilerHost.readFile(fileName), () => options, setParentNodes, skipNonSemanticJSDocParsing),
+        getSourceFile: createGetSourceFile(fileName => compilerHost.readFile(fileName), () => options, setParentNodes),
```
jakebailey, line 483: It's possible this isn't what you meant, @sheetalkamat; maybe you meant to make `createGetSourceFile` instead call the host, kinda like readFile?

jakebailey:

jakebailey:
I effectively removed support for `ScriptTarget` up the chain, and I believe I have plumbed everything correctly. If I've gone overboard and there's some way to simplify this, I'm totally happy to change it. Just want to get it in...

sheetalkamat (approved):

src/compiler/program.ts
sheetalkamat: Your current change is what i meant in the commit comment before. So all good.

jakebailey: Yay, thank you!!

jakebailey:
FYI @nonara; in `executeCommandLine.ts`, there's a new global `defaultJSDocParsingMode`; in `tsc` here that's set to `ParseForTypeErrors`, but it's likely that someone's plugin will actually need this to be reset to `ParseAll` in order to get all of the JSDoc nodes they expect. e.g. I suspect typia reads JSDoc, somehow. Probably others. Just something to watch out for!

src/compiler/types.ts
```diff
@@ -9850,6 +9852,34 @@ export const commentPragmas = {
...
+     * Never parse JSDoc comments, mo matter the file type.
+     */
+    ParseNone,
+    /**
+     * Parse only JSDoc comments which are needed to provide correct type errors.
+     *
+     * This will always parse JSDoc in non-TS files, but only parse JSDoc comments
+     * containing `@see` and `@link` in TS files.
+     */
+    ParseForTypeErrors,
```
jakebailey, line 9872: Last minute bikeshedding; `ParseForErrors`? Technically, this covers grammar checks.

jakebailey: But, not actually sure we do grammar checks which relate to JSDoc.

nonara:
Thanks for the heads up!

src/compiler/types.ts
sandersn: Not in TS files, no. Just in JS files.

jakebailey:
Alright, seems like everything is alright; I'm going to merge so we can get this in for 5.3. Thank you everyone for the feedback, hopefully it works out (and tell me if it doesn't).

samchon:
Thanks for detailed documentation.

PR #55740, opened 2023-09-14
Fix playright browser installs due to upstream breaking change

jakebailey:
See: https://github.com/microsoft/playwright/releases/tag/v1.38.0 and https://github.com/microsoft/TypeScript/actions/runs/6178823659/job/16773312341

The reason we didn't depend directly on `playwright` was because it'd automatically download multiple browsers. The latest version no longer does this, which breaks our workflow. But since that was the whole reason we didn't directly depend on it, we can just depend on it directly and get rid of the awkward CI scripts which needed to install it after the fact.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,9 +97,6 @@ jobs:
           check-latest: true
       - run: npm ci
 
-      - name: Adding playwright
-        run: npm install --no-save --no-package-lock playwright
-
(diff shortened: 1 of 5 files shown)


jakebailey:
Note that this does increase local package install size by like 7MB. I could leave the conditional import in there if we really wanted by reverting the last commit.

andrewbranch (approved):

Issue #55301, opened 2023-08-08
How and where to add some utility types


Daggron:
### Acknowledgement

- [X] I acknowledge that issues using this template may be closed without further explanation at the maintainer's discretion.

### Comment

I just want to ask few questions w.r.t utility types in typescript

Right now whenever we have to get a union of object values we have to do something like this

```ts

const stages = {
   LOADING : 'LOADING'
   ERROR: 'ERROR'
   SUCCESS: 'SUCCESS'
} as const 

type Stages = typeof stages[keyof typeof stages]
```

this becomes redundant in code base and can be improved with utility types for e.g.

```ts
type ObjectValues<T extends Record<any, any>> =  T[keyof T];
```

Willing to add this as a utility type provided by TS, but stuck on where to add this `es5.d.ts` is one place where I can write it but not sure if thats the correct place and do I need to add it somewhere else or not.

MartinJohns:
Nowhere. See https://github.com/microsoft/TypeScript/issues/39522#issuecomment-656201388:

> We've opted to not include utility type aliases in the lib unless they're required for declaration emit purposes.

Daggron:
> Nowhere. See [#39522 (comment)](https://github.com/microsoft/TypeScript/issues/39522#issuecomment-656201388):
> 
> > We've opted to not include utility type aliases in the lib unless they're required for declaration emit purposes.

Lol, time to create my own utility library then 😂

MartinJohns:
You're free to do so. Or use one of the many many out there.

Issue #39522, opened 2020-07-09
Add 'Nullable' built-in utility type


wessberg:
I've made a [PR](https://github.com/microsoft/TypeScript/pull/39523)

## Search Terms

`Nullable`

## Suggestion

TypeScript already comes with a `NonNullable` type, defined in `lib.es5.d.ts`:

```typescript
/**
 * Exclude null and undefined from T
 */
type NonNullable<T> = T extends null | undefined ? never : T;
```

I propose that its inverse, `Nullable`, is added to the built-in type utilities.

I'm aware that your general position in terms of built-in type utilities is that you don't extend them unless strictly necessary, but I _do_ think this is an important addition for the following reason:

The built-in type utilities generally offer inverse types. For example, `Partial` is the inverse of `Required`. And, `Pick` is the inverse of `Omit`. Because of this, to me it naturally follows that `NonNullable` should have an inverse called `Nullable`.

## Use Cases

For the following examples, _nullable_ is defined as `null` or `undefined`, thereby following TypeScripts definition from the `NonNullable` utility type.

A few examples that comes to mind:

- Required keys with nullable values.
- Function arguments that may have a nullable value.

Additionally, it may help with simplifying built-in type definitions:

For example in `lib.es5.d.ts`, the `catch` and `then` declarations for the `Promise` interface can be simplified:

```typescript
interface Promise<T> {
    then<TResult1 = T, TResult2 = never>(onfulfilled?: Nullable<(value: T) => TResult1 | PromiseLike<TResult1>>, onrejected?: Nullable<(reason: any) => TResult2 | PromiseLike<TResult2>>): Promise<TResult1 | TResult2>;
    catch<TResult = never>(onrejected?: Nullable<(reason: any) => TResult | PromiseLike<TResult>>): Promise<T | TResult>;
}
```

There are several types, particularly in `lib.dom.d.ts` that could also be simplified by leveraging the `Nullable` type.

## Examples

As part of an `InterfaceDeclaration`:

```typescript
interface Foo {
    prop: Nullable<string>;
}
```

As a optional parameter to a function that may also accept `null`:

```typescript
function validateUserInput (text: Nullable<string>): void {}
```

## Checklist

My suggestion meets these guidelines:

* [x] This wouldn't be a breaking change in existing TypeScript/JavaScript code
* [x] This wouldn't change the runtime behavior of existing JavaScript code
* [x] This could be implemented without emitting different JS based on the types of the expressions
[...]

PR #55746, opened 2023-09-14
Add 'not a request for a utility type' to the feature issue template

jakebailey:
Please do bikeshed the wording, but I think we should probably have a checklist item for this given a non-trivial amount of asks for this which we immediately decline unless it's paired with some other need.

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.yml
+++ b/.github/ISSUE_TEMPLATE/feature_request.yml
@@ -40,6 +40,8 @@ body:
           required: true
         - label: This isn't a runtime feature (e.g. library functionality, non-ECMAScript syntax with JavaScript output, new syntax sugar for JS, etc.)
           required: true
+        - label: This isn't a request to add a new utility type
+          required: true
         - label: 'This feature would agree with the rest of our Design Goals: https://github.com/Microsoft/TypeScript/wiki/TypeScript-Design-Goals'
           required: true
   - type: textarea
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -40,6 +40,8 @@ body:
           required: true
         - label: This isn't a runtime feature (e.g. library functionality, non-ECMAScript syntax with JavaScript output, new syntax sugar for JS, etc.)
           required: true
+        - label: This isn't a request to add a new utility type
```
RyanCavanaugh, line 43: I think we need a FAQ entry to link to. Thoughts?

jakebailey: https://github.com/microsoft/TypeScript/issues/39522#issuecomment-656201388 or https://github.com/microsoft/TypeScript/issues/55301? They aren't FAQs per say. Otherwise, the wiki somewhere?

fatcerberus:
I probably laughed harder at the title of this PR than I want to admit. 😄  It's definitely true, asks for new utility types are right up there with 9998 dupes and bug reports about type narrowing not working for indexed access.

And yeah, there should probably be a contributor FAQ for this kind of thing (assuming there isn't already).  It's like how people aren't supposed to submit typo fixes, it doesn't really seem to be documented except in other issues and *maybe* the issue template, so the usual recourse is just to dupe to a random rejected issue of the same type, which isn't really an ideal way to communicate policy.

.github/ISSUE_TEMPLATE/feature_request.yml
RyanCavanaugh: Wrote https://github.com/microsoft/TypeScript/wiki/No-New-Utility-Types

jakebailey:
```suggestion
        - label: This isn't a request to add a new utility type: https://github.com/microsoft/TypeScript/wiki/No-New-Utility-Types
```

jakebailey: Added to the template, PTAL

PR #55750, opened 2023-09-15
Ensure generated property names for methods named "new" are quoted

### LINKED ISSUE #55075: Declaration for object with method named "new" incorrectly emitted with construct signature

jakebailey:
Fixes #55075
Closes #55109

This effectively replaces #55109, instead using a fix in `createPropertyNameNodeForIdentifierOrLiteral` instead of the emitter. If you look at the last commit, you'll see that the only change to TS emit was removed, which I believe to be correct as the code was able to be written that way in the first place so should be legal.

While I was here, I made the parameters to `createPropertyNameNodeForIdentifierOrLiteral` required, which showed that there was an existing bug for this in a code fix. Likely I need a test for this. It also shows that this call doesn't respect `stringNamed` either, which unfortunately requires some type info I didn't look into getting quite yet.

cc @Andarist


src/compiler/checker.ts
```diff
@@ -8213,16 +8213,17 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
                 return fromNameType;
             }
             const rawName = unescapeLeadingUnderscores(symbol.escapedName);
-            return createPropertyNameNodeForIdentifierOrLiteral(rawName, getEmitScriptTarget(compilerOptions), singleQuote, stringNamed);
+            return createPropertyNameNodeForIdentifierOrLiteral(rawName, getEmitScriptTarget(compilerOptions), singleQuote, stringNamed, isMethod);
         }
 
         // See getNameForSymbolFromNameType for a stringy equivalent
-        function getPropertyNameNodeForSymbolFromNameType(symbol: Symbol, context: NodeBuilderContext, singleQuote?: boolean, stringNamed?: boolean) {
+        function getPropertyNameNodeForSymbolFromNameType(symbol: Symbol, context: NodeBuilderContext, singleQuote: boolean, stringNamed: boolean, isMethod: boolean) {
```
RyanCavanaugh, line 8226: Idle observation: Lint rule for functions with optional parameters which are never not supplied? 🤔

RyanCavanaugh (approved):

src/compiler/checker.ts
jakebailey: Probably better as some external pass, as ts-eslint doesn't really do stuff cross-project. Technically possible with the new ProjectService thing, maybe.

PR #55751, opened 2023-09-15
Require Node 16.14+ and ES2021+

jakebailey:
Node 16 just went EOL, making Node 14 doubly EOL. For the last few releases, we've "officially" supported only the latest EOL release, which Node 16 now is. Node 16's deprecation is "early" upstream due to its OpenSSL dependency going EOL (and nobody wants to maintain a crypto lib on their own). Whether or not that means we should wait on this PR until when Node 16 "would have" gone EOL if it had not been for the OpenSSL thing, I'm not sure.

- Node 16.14 was chosen as it's the last version of Node 16 which added new syntax, namely, import assertions. Of course, those are now defunct in ECMAScript in favor of `with`, so, oops. Otherwise I would have chosen 16.11 for class static blocks.
- ES2021 is chosen per the mapping to Node 16 here: https://github.com/microsoft/TypeScript/wiki/Node-Target-Mapping

Notably, this change gives us native `?.`, which is a nice for code size. (Maybe for perf, too?)

## Overall package size

| - | - | - | - | - |
| Packed | 3.77 MiB | 3.75 MiB | -16.24 KiB | -0.42% |
| Unpacked | 20.37 MiB | 20.27 MiB | -98.30 KiB | -0.47% |

## Files

| - | - | - | - | - |
| `lib/tsc.js` | 5.57 MiB | 5.53 MiB | -40.00 KiB | -0.70% |
| `lib/tsserver.js` | 24.64 KiB | 24.42 KiB | -222.00 B | -0.88% |
| `lib/typescript.js` | 8.17 MiB | 8.11 MiB | -58.07 KiB | -0.69% |
| `lib/typingsInstaller.js` | 10.47 KiB | 10.46 KiB | -13.00 B | -0.12% |


jakebailey:
@typescript-bot perf test full
@typescript-bot perf test bun
@typescript-bot perf test vscode

jakebailey:

ehoogeveen-medweb:
FWIW, according to https://kangax.github.io/compat-table/es2016plus/ Node 16.11+ actually supports all ES2022 features (aside from the `d` / `hasIndices` regexp flag not showing up in the flags, but the latest versions of Node, Chrome and Firefox all seem to agree on that). IIRC some features were backported to version 16.

jakebailey:
I pulled this info from https://github.com/evanw/esbuild/blob/main/internal/compat/js_table.go which combines many different sources, choosing the maximum Node 16 there, which is v16.14 for import assertions. But I doubt we need those, so yes, 16.11 would probably also be fine (mentioned that in the description).

ehoogeveen-medweb:
I was more commenting on the choice of `ES2021` as it appears that `ES2022` could also work :)

jakebailey:
Ah, yeah. It's half awkward because we recommend Node16=ES2021 on our wiki. That and we need to assume the correct `lib`, but maybe that's ES2022 too?

jakebailey:

Issue #53699, opened 2023-04-08
should be able to run typescript under `--frozen-intrinsics`


loynoir:
# Bug Report

### 🔎 Search Terms

- `--frozen-intrinsics`
- `Error.stackTraceLimit`
- prototype protection

### 🕗 Version & Regression Information

- This is a crash
- This changed between versions ______ and _______
- This is the behavior in every version I tried, and I reviewed the FAQ for entries about _________
- I was unable to test this on prior versions because _______

### ⏯ Playground Link

[Playground link with relevant code](https://www.typescriptlang.org/play?#code/PTAEFkE9QYwewCYFNQHM5IM6gBZIE5JA)

### 💻 Code

```ts
import ts from 'typescript'
console.log(ts.sys.fileExists("reproduce.mjs"))
```

### 🙁 Actual behavior

```txt
$ node --frozen-intrinsics reproduce.mjs 
(node:239937) ExperimentalWarning: The --frozen-intrinsics flag is experimental
(Use `node --trace-warnings ...` to show where the warning was created)
/path/to/node_modules/typescript/lib/typescript.js:6530
            Error.stackTraceLimit = 0;
                                  ^

TypeError <Object <Object <[Object: null prototype] {}>>>: Cannot assign to read only property 'stackTraceLimit' of function 'function Error() { [native code] }'
    at fileSystemEntryExists (/path/to/node_modules/typescript/lib/typescript.js:6530:35)
    at fileExists (/path/to/node_modules/typescript/lib/typescript.js:6551:20)
    at isFileSystemCaseSensitive (/path/to/node_modules/typescript/lib/typescript.js:6397:21)
    at getNodeSystem (/path/to/node_modules/typescript/lib/typescript.js:6189:45)
    at /path/to/node_modules/typescript/lib/typescript.js:6603:18
    at src/compiler/sys.ts (/path/to/node_modules/typescript/lib/typescript.js:6609:9)
    at __init (/path/to/node_modules/typescript/lib/typescript.js:22:58)
    at src/compiler/_namespaces/ts.ts (/path/to/node_modules/typescript/lib/typescript.js:125073:7)
    at __init (/path/to/node_modules/typescript/lib/typescript.js:22:58)
    at src/typescript/_namespaces/ts.ts (/path/to/node_modules/typescript/lib/typescript.js:169775:7)

Node.js v18.15.0
```

### 🙂 Expected behavior

```txt
$ node reproduce.mjs 
true
[...]

PR #55755, opened 2023-09-15
Freeze emptyArray, make emptyMap and emptySet immutable

jakebailey:
We'll see if this ends up slow, but it theoretically prevents us from making mistakes and modifying these sentinel values.


jakebailey:

jakebailey:
oof

src/compiler/core.ts
```diff
@@ -13,11 +13,27 @@ import {
...
 /** @internal */
-export const emptyMap: ReadonlyMap<never, never> = new Map<never, never>();
+export const emptyMap: ReadonlyMap<never, never> = createReadonlyMap();
 /** @internal */
-export const emptySet: ReadonlySet<never> = new Set<never>();
+export const emptySet: ReadonlySet<never> = createReadonlySet();
+
+function createReadonlyMap<K, V>(): ReadonlyMap<K, V> {
+    const map: ReadonlyMap<K, V> & Partial<Map<K, V>> = new Map();
+    map.clear = notImplemented;
```
Jack-Works, line 24: use Object.defineProperty or you'll hit the override mistake

jakebailey: As in someone setting it back to another method?

Jack-Works: override mistake is a tc39 term: https://web.archive.org/web/20141230041441/http://wiki.ecmascript.org/doku.php?id=strawman:fixing_override_mistake

in some environment it might be frozen

fatcerberus: I don’t think that’s an official term.  It’s just one person thinking it was a mistake/bug in the specification and calling it the “override mistake” as a result.

Jack-Works: this is well-known in a subset of tc39 members (especially secure ecmascript people). for example, this issue also mentioned this term https://github.com/tc39/proposal-iterator-helpers/issues/286#issue-1876367091

jakebailey: I'm still not sure what this has to do with overwriting a specific instance's properties. We have other such constructs like MultiMap which do this.

Jack-Works: So maybe it's already throws under a lockdown environment (where all built-in prototypes are frozen). I hope you can use defineProperty, and at least be aware there is a thing that might cause problems. thanks!

Jack-Works: oh... I see that TS is already not working with it, but I do not see a statement that the support is rejected by the team

jakebailey: Of those I'm pretty sure only #53699 is related (where tsc can run), but that one is still just talking about modifying some sort of global prototype or something. Can someone please explain to me why modifying a _specific instance of Map_ is invalid?

Jack-Works: Try this:

```js
"use strict"
Object.freeze(Map.prototype)
var x = new Map
x.set = () => {} // TypeError
```

jakebailey: Thanks, that's the example I was looking for. Probably when I get bored I should just add a new CI job which tests with frozen-intrinsics or something.

Jack-Works: thanks very much!

jakebailey: Hm, I guess `--frozen-intrinsics` does not ban this particular code. The only thing it flags in the codebase is our touching of `Error`, and I'll also point out that if tacking things onto `Map` is bad, then `NodeArray` surely is _very_ bad because it's `Array` with extra properties.

Jack-Works: extra property is fine, assign a property that is already in the prototype chain and writable: false is not fine. 
you can see https://github.com/tc39/how-we-work/pull/137/files

the code example i have given before also shows that the current code your writing will throw a type error in the frozen mode.

```js
"use strict"
Object.freeze(Map.prototype)
var x = new Map
x.set = () => {} // TypeError
```

use defineProperty and all problems gone!!

jakebailey: I see, though I still can't just make a CI job with `--frozen-intrinsics` and have this code error.

image

michaelficarra: @jakebailey It seems that `--frozen-intrinsics` doesn't make non-writable data properties, it makes them magic accessor properties.

```
> Object.getOwnPropertyDescriptor(Map.prototype, 'set')
Object <[Object: null prototype] {}> {
  get: [Function: getter] Function <Function <[Object: null prototype] {}>> {
    value: [Function: set] Function <Function <Complex prototype>>
  },
  set: [Function: setter] Function <Function <[Object: null prototype] {}>>,
  enumerable: false,
  configurable: false
}
> console.log(Object.getOwnPropertyDescriptor(Map.prototype, 'set').set.toString())
function setter(newValue) {
        if (obj === this) {
          // eslint-disable-next-line no-restricted-syntax
          throw new TypeError(
            `Cannot assign to read only property '${prop}' of object '${obj}'`,
          );
        }
        if (ObjectPrototypeHasOwnProperty(this, prop)) {
          this[prop] = newValue;
        } else {
          ObjectDefineProperty(this, prop, {
            __proto__: null,
            value: newValue,
            writable: true,
            enumerable: true,
            configurable: true,
          });
        }
      }
>
```

Issue #55778, opened 2023-09-18
Const type parameter isn't inferred as const if it's part of an intersection


Andarist:
### 🔎 Search Terms

const type parameter inference intersection

### 🕗 Version & Regression Information

- This is the behavior in every version I tried

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.3.0-dev.20230918#code/JYOwLgpgTgZghgYwgAgMIHsQ2AcwDwAqAjMhAB6QgAmAzsgN7JgCeADhAFzI1hSg7IAvgD4GAKGTIArjQgBJMF2IBuMYLFiqEBABs4UFDCkgEYYJiYQeeCcmKkKEanUYt2XHnxADBAGlsImDx2GFi4DpS0aJjY+MTC-sIAFIFhOFyMrFDoVFJIBAAW-EokgsgAZCExuACUSqGxqmKpwQY0UjpgyAC8ljxJ9LZZOXkQhcUMZXAutpJunMgARDDo6Iuqkn62MvKK4pJzbAvLq4v+m+cRUHBcRAAM-oI1qkA

### 💻 Code

```ts
interface Config<T1 extends { type: string }> {
  useIt: T1;
}

declare function test<
  T1 extends { type: string },
  const TConfig extends Config<T1>,
>(config: { produceThing: T1 } & TConfig): TConfig;

const result = test({
  produceThing: {} as {
    type: "foo";
  },
  useIt: {
    type: "foo",
  },
  extra: 10,
});
```

### 🙁 Actual behavior

We can observe an EPC error here which indicates that `TConfig` failed to be inferred.

### 🙂 Expected behavior

I would expect this to be inferred successfully. 

### Additional information about the issue

Such an inference candidate is gathered for `TConfig`: `{ produceThing: { type: "foo"; }; useIt: { type: string; }; extra: number; }`. As we can see, `useIt` refers to a `string` type and not to a string literal type. This, in turn, fails the `compareTypes` check since the constraint is `Config<{ type: 'foo' }>` and thus the constraint is assigned as the inferred type.

PR #55779, opened 2023-09-18
Use const context in intersections containing const type variables

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/55778


jakebailey:
This feels almost too simple... I feel like I remember the original PR not doing this on purpose. @ahejlsberg 

@typescript-bot test top200
@typescript-bot run dt
@typescript-bot pack this

jakebailey:
I will note that perf/DT don't really have const type params, and they don't have high usage yet, so it's hard to say if we'll get any feedback for this within the PR now. Probably we should just merge it and then wait for someone to come up with a bad case.

PR #53356, opened 2023-03-19
Allowed tuples to have both named and anonymous members

JoshuaKGoldberg:
Fixes #52853

Previously, the `labeledElementDeclarations` property on `TupleType`s was all-or-nothing for having named vs. anonymous members. Now that it's possible for them to mix-and-match, this PR pulls logic from `getParameterNameAtPosition` into `getTupleElementLabel` for filling in a name (<code>\`${restParameterName}_${index}\`</code>).

Co-authored-by: Mateusz Burzyński <mateuszburzynski@gmail.com>

PR #55789, opened 2023-09-19
Preserve partially typed tuple names in more places

jakebailey:
#53356 introduced the ability to have partially named tuples, but while working on another bug, I noticed that there were a few places that the PR missed that should also be preserving names. This fixes those cases, with tests.

There are still other calls to `createTupleType` which are missing names, but they all appear to be unable to generate them (anywhere that generates a tuple from a binding, or are "noramlizing" a tuple which can create more than one element for the same name).

### DIFF
--- a/tests/baselines/reference/restTuplesFromContextualTypes.types
+++ b/tests/baselines/reference/restTuplesFromContextualTypes.types
@@ -265,8 +265,8 @@ f3((a, b, c) => {})
 f3((...x) => {})
 >f3((...x) => {}) : void
 >f3 : (cb: (x: number, args_0: boolean, ...args_1: string[]) => void) => void
->(...x) => {} : (x_0: number, x_1: boolean, ...x_2: string[]) => void
->x : [number, boolean, ...string[]]
+>(...x) => {} : (x: number, x_1: boolean, ...x_2: string[]) => void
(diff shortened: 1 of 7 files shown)


tests/baselines/reference/restTuplesFromContextualTypes.types
```diff
@@ -265,8 +265,8 @@ f3((a, b, c) => {})
 f3((...x) => {})
 >f3((...x) => {}) : void
 >f3 : (cb: (x: number, args_0: boolean, ...args_1: string[]) => void) => void
->(...x) => {} : (x_0: number, x_1: boolean, ...x_2: string[]) => void
->x : [number, boolean, ...string[]]
+>(...x) => {} : (x: number, x_1: boolean, ...x_2: string[]) => void
```
weswigham, line 268: This change is a little funky - `x` isn;t consistent with `x_1`, but not a dealbreaker.

weswigham (approved):

tests/baselines/reference/restTuplesFromContextualTypes.types
jakebailey: `x` is just the name given in the original type, the rest are generated.

PR #54785, opened 2023-06-26
Resolution testing

sheetalkamat:
Fixes #

PR #55790, opened 2023-09-19
Source files not affected by all module resolution options

sheetalkamat:
Stops taking dependency on all module resolution options when sharing sourceFiles and instead moves `resolvedModules` and `resolvedTypeReferenceDirectiveNames` from `sourceFile` to `Program` so it is safe by program to copy the resolutions across.

This came up as part of testing with #54785 where `axion` and some more (which i dont recollect) didnt pass incremental resolution check. This is shown in ddaf1c7 specifically (https://github.com/microsoft/TypeScript/commit/ddaf1c77fc29e91cc8a8a5c45dc1cb002d0ee061#diff-215d7c39ab1082f80d4c3a4f1c8bd9dd1c9ccb1c6450192bbaf409b8e5376579R140) where the resolved module on sourceFile changes and results in issues program referencing other resolution.
The issue here lies in fact that "as type" resolution takes into account typeRoots and if config had it its good and used as a key for affecting resolution but if user didnt specify it depends on "config file location" if present otherwise "currentDirectory" . This is issue because you could include either by types or default types (matching in both projects) a type thats in parent directories `node_modules/@types` but then that type could include `types` directive of say `node` which in both projects will be resolved say from config location which is different and if those directories have their own `node_modules/@types/node` then the location which is picked will depend on which project is picked first
Investigated few options for this
- adding default type roots as key but that mean resolutions cannot be shared between two projects with config file in different directories which is very common.
- On top of above option adding only present `node_modules/@types`  in the key but that means on each `npm install` we would need to update module resolutions from scratch as its change in "key" and parse "sourceFiles" so that doesnt seem practical
-  The current change where decoupling sourceFiles and module resolution storage, This doesnt fix module resolution cache sharing across projects (which is only in tsbuild for now which doesnt share the caches between watch events and in future we will share in tsserver but that one will watch failed lookup locations so we could somehow use that so probably can be handled in some way without having to invalidate everything)

### DIFF
--- a/src/compiler/utilities.ts
+++ b/src/compiler/utilities.ts
@@ -79,7 +79,6 @@ import {
     contains,
     containsPath,
     createGetCanonicalFileName,
-    createModeAwareCache,
     createMultiMap,
     createScanner,
     createTextSpan,
@@ -745,34 +744,6 @@ export function getFullWidth(node: Node) {
 }
 
 /** @internal */
-export function getResolvedModule(sourceFile: SourceFile | undefined, moduleNameText: string, mode: ResolutionMode): ResolvedModuleFull | undefined {
-    return sourceFile?.resolvedModules?.get(moduleNameText, mode)?.resolvedModule;
-}
-
-/** @internal */
-export function setResolvedModule(sourceFile: SourceFile, moduleNameText: string, resolvedModule: ResolvedModuleWithFailedLookupLocations, mode: ResolutionMode): void {
-    if (!sourceFile.resolvedModules) {
-        sourceFile.resolvedModules = createModeAwareCache();
-    }
-
-    sourceFile.resolvedModules.set(moduleNameText, mode, resolvedModule);
-}
-
-/** @internal */
-export function setResolvedTypeReferenceDirective(sourceFile: SourceFile, typeReferenceDirectiveName: string, resolvedTypeReferenceDirective: ResolvedTypeReferenceDirectiveWithFailedLookupLocations, mode: ResolutionMode): void {
-    if (!sourceFile.resolvedTypeReferenceDirectiveNames) {
-        sourceFile.resolvedTypeReferenceDirectiveNames = createModeAwareCache();
-    }
-
-    sourceFile.resolvedTypeReferenceDirectiveNames.set(typeReferenceDirectiveName, mode, resolvedTypeReferenceDirective);
-}
-
-/** @internal */
-export function getResolvedTypeReferenceDirective(sourceFile: SourceFile | undefined, typeReferenceDirectiveName: string, mode: ResolutionMode): ResolvedTypeReferenceDirective | undefined {
-    return sourceFile?.resolvedTypeReferenceDirectiveNames?.get(typeReferenceDirectiveName, mode)?.resolvedTypeReferenceDirective;
-}
-
-/** @internal */
 export function projectReferenceIsEqualTo(oldRef: ProjectReference, newRef: ProjectReference) {
     return oldRef.path === newRef.path &&
         !oldRef.prepend === !newRef.prepend &&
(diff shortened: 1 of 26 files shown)


sheetalkamat:
@typescript-bot test this
@typescript-bot test top100

sheetalkamat:
@typescript-bot test this
@typescript-bot test top100

DanielRosenwasser:
Can you give some more details on what this change is regarding? I can see baselines changing, but I'm not entirely sure what was wrong before.

sheetalkamat:
@DanielRosenwasser Was just in process of writing

tests/baselines/reference/nodeColonModuleResolution.trace.json
```diff
@@ -1,4 +1,5 @@
 [
+    "Module 'ph' was resolved as locally declared ambient module in file '/a/b/node_modules/@types/node/ph.d.ts'.",
```
sheetalkamat, line 2: This is something to do with how we compile the program twice. which i dont understand clearly but when i run the scenario as "tsc" this trace is expected and was missing before.

tests/baselines/reference/tsserver/documentRegistry/Caches-the-source-file-if-script-info-is-orphan,-and-orphan-script-info-changes.js
```diff
@@ -69,7 +69,7 @@ Info seq  [hh:mm:ss:mss] Open files:
 Info seq  [hh:mm:ss:mss] 	FileName: /user/username/projects/myproject/index.ts ProjectRootPath: undefined
 Info seq  [hh:mm:ss:mss] 		Projects: /user/username/projects/myproject/tsconfig.json
 DocumentRegistry::
-  Key:: undefined|undefined|undefined|undefined|false|undefined|undefined|undefined|undefined|undefined|undefined|undefined|undefined|undefined|undefined|undefined|undefined|undefined|undefined|undefined|undefined|undefined|undefined|undefined
+  Key:: undefined|undefined|undefined|false|undefined|undefined|undefined|undefined|undefined|undefined
```
sheetalkamat, line 72: This just shortens the key because now many of the module resolution affecting options are not accounted in sourceFile sharing

jakebailey:
Just to confirm my understanding, this moves properties off of `SourceFile` onto `Program`, such that the `SourceFile` can be shared more and has fewer props, and the diff is just all of the old places now calling the `Program` to get the same info, without any other behavior changes?

src/compiler/utilities.ts
```diff
@@ -744,34 +743,6 @@ export function getFullWidth(node: Node) {
     return node.end - node.pos;
 }
 
-/** @internal */
-export function getResolvedModule(sourceFile: SourceFile | undefined, moduleNameText: string, mode: ResolutionMode): ResolvedModuleFull | undefined {
```
jakebailey, line 748: I'm honestly a little nervous to remove these APIs even though they're internal; people appear to be using it, including Roblox's TS-to-Lua converter and @andrewbranch's arethetypeswrong: https://github.com/search?q=%2Fts%5C.getResolvedModule%2F&type=code

jakebailey: It also seems to be the case that (at least for ATTW) that some people don't actually have Programs at this point, either...

sheetalkamat: With this change this can not work without giving program so i dont see what we can do here. Was internal for a purpose there

jakebailey: Roblox appears to have a `Program` at that point, so that one is at least fixable (and a few other smaller examples). `dukat` doesn't in all uses, nor does ATTW.

andrewbranch: The properties being retrieved here would not exist if there were not a Program already. In my case the Program is somewhere else but I can get to it. In fact, I think what I’m doing becomes easier if it’s moved to the Program.

jakebailey: Ah, you're totally right, I didn't realize! Consider this resolved 😄

jakebailey (approved):

andrewbranch:
It seems like this could actually inflate memory usage (you can see a couple modest increases in some of the perf tests) when there are many programs with identical `moduleResolution`-related settings and some common dependencies. If the resolution-affecting options are the same between programs, those cases already had optimal SourceFile reuse but will now duplicate the resolutions (or at least maps of pointers to them) across each Program. Would it be valuable to move the resolutions not just to Program but to a shared, properly-keyed data structure similar to `DocumentRegistry`?

sheetalkamat:
I was thinking as a next step we could try and use module resolution cache to get the result instead of storing it in the program but that needs some thinking( since in most cases resolution already exists in the cache)

storing on document registry would need so comparisons and invalidation logic that would lead to similar issue as I am trying to fix here (resolving type reference of auto type from different locations when two projects don’t specify type roots )

andrewbranch (approved):

PR #12589, opened 2016-11-30
Mapped type modifier inference

ahejlsberg:
This PR improves our handling of readonly and optional modifiers in mapped type inference. When inferring to a homomorphic (structure preserving) mapped type `{ readonly [P in keyof T]: X }`, no `readonly` modifiers will be present in the type inferred for `T` (because it is known that the mapping will add it). Likewise, when inferring to a mapped type `{ [P in keyof T]?: X }`, no `?` modifers will be present in the type inferred for `T`. In effect, readonly and optional modifiers in the target are used to indicate that those modifiers should be *stripped* in the inferred type.

```ts
declare function validate<T>(obj: { [P in keyof T]?: T[P] }): T;
declare function clone<T>(obj: { readonly [P in keyof T]: T[P] }): T;
declare function validateAndClone<T>(obj: { readonly [P in keyof T]?: T[P] }): T;

type Foo = {
    a?: number;
    readonly b: string;
}

function test(foo: Foo) {
    let x = validate(foo);  // { a: number, readonly b: string }
    let y = clone(foo);  // { a?: number, b: string }
    let z = validateAndClone(foo);  // { a: number, b: string }
}
```

PR #55794, opened 2023-09-20
Fixed const reverse mapped types themselves to be treated as const

Andarist:


tests/cases/conformance/types/typeParameters/typeParameterLists/typeParameterConstModifiersReverseMappedTypes.ts
```diff
@@ -0,0 +1,41 @@
+// @strict: true
+// @noEmit: true
+
+declare function test1<const T>(obj: {
```
Andarist, line 4: The first 3 test cases are testing objects and how `readonly`, `-readonly`, and lack of them impact the `T` and the fresh object that is created by mapping over `T`.

Those objects today would be constified for the most part but their properties wouldn't always be marked as `readonly` when they should be. When `readonly` modifier is used on the mapped type then that result (`typeof obj`) **has** `readonly` properties but the `const` type parameter itself doesn't have them. This is consistent with https://github.com/microsoft/TypeScript/pull/12589 . You might also want to recheck when/how `readonly` is preserved/stripped on members of the reverse mapped type [here](https://github.dev/microsoft/TypeScript/blob/2738fbf1e05b9429d03b5956ba0413d92f5b993c/src/compiler/checker.ts#L13491)

tests/cases/conformance/types/typeParameters/typeParameterLists/typeParameterConstModifiersReverseMappedTypes.ts
```diff
@@ -0,0 +1,41 @@
...
+  nested: {
+    nestedProp: "bar",
+  },
+});
+
+declare function test4<const T extends readonly unknown[]>(arr: {
+  [K in keyof T]: T[K];
+}): T;
+
+const result4 = test4(["1", 2]);
```
Andarist, line 41: This is an especially important part of the fix. Currently `T` is inferred as `(2 | "1")[]` but it really should be `readonly ["1", 2]`

tests/cases/conformance/types/typeParameters/typeParameterLists/typeParameterConstModifiersReverseMappedTypes.ts
Andarist, line 1: There are also some other test cases that I'd like to add here but I'm not entirely sure what should be the expected results for them
1. `const T, P extends keyof T` - this might iterate over `P` which might be the subset of `keyof T`. Should we still assign constness to `T` in such a case? See the playground [TS playground](https://www.typescriptlang.org/play?ts=5.3.0-dev.20230920#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXwxAGcMBWAHjDxPgBUAaeABXhAA9DVgj4BrEAJ45EdAHwAKALAAoePBwAjAFYAueAG8ZcuQG0A0vCz4mAXTXr4ANygRkINbX0n4AXwDcW1-U-8BRNUw6Jt7SAJRqOgwEAgAOIMLyyiYe0jJUqDRwRMgQZPAAvATEZFKyGp5yUOZWNnZqAERQ9V4V8ArV1rb28PUKzS4hcgOeOo31jL31wTKhKWnUGPBZORgAbAVFJKSlcppllR213RagIDENTa4t+22HXdWn5z19V8Nlb7pjE33TYXPSQA). Note that it's not really possible to gather `P` across from different mapped types and somehow different mapped type to build up a "combined" reverse mapped type: [TS playground](https://www.typescriptlang.org/play?ts=5.3.0-dev.20230920#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXwxAGcMA2AHjDxPgBUAaeABXhAA9DVgj4BrEAJ45EdAHwAKALAAoePBwAjAFYAueAG8ZcuQG0A0vCz4mAXQD8a9fABuUCMhBra+k-AC+Abi3v63xUoAmS29dAyNmc0sbOwcnF3cvWR9vfgEiNSYdE19pAEo1HQYCAQAHEGF5ZRNEmSpUGjgiZAgMAHZ4AF4CYjIpJM0kuSgo23tHeAAiKAnkwfgFEZjxiYUZtxy5de8B7XgwRbG1CbA1je1gA9jJ4FPvLaSdKYnGFYnsmVya6SA)
2. `K in keyof T | "extra"` - what should happen with a non-homomorphic case like this? [TS playground](https://www.typescriptlang.org/play?ts=5.3.0-dev.20230920&ssl=2&ssc=4&pln=2&pc=26#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXwxAGcMB2AHjDxPgBUA+AChwCMArALngG8BYAKHjwA2gGl4WfAGsQATxyI68AD7wARCAAeGGFDUBdLmimocAd1QBuAQF8AlF2G0ANAVkAHEAvis2+6-wCVKg0cETIEBgAHPAAvATEZIx8gvDuMDjuXGqIODhqzgJCqIkgwFwpQsWlwAAKGVnqLLAFRfA2halaOlDZzQBerfz2AUA)

weswigham: 1. Probably? I'm not a huge fan of overuse of `const` type variables, but if they're in-use, there a strong signal that const-ness is desired *somewhere* related to that type parameter. The workflow is usually "does this do what I want? No? Add a `const`.", so the `const` normally only gets added by people if it's meant to be meaningful to inference.
2. Eh, I'd leave that as whatever it falls out as from being not homomorphic. It'll get unified if and when we ever adjust the logic to consider something like that as homomorphic-enough to get the homomorphic type variable treatment.

weswigham (approved):

weswigham:
@typescript-bot test this
@typescript-bot run dt
@typescript-bot test top100

PR #55799, opened 2023-09-20
Add pre-commit dprint hook, fix post-checkout hook

jakebailey:
This adds a pre-commit hook that will format staged files before committing.

This also fixes the `post-checkout` hook, which I'm pretty sure has been broken this whole time (which to me means that it's not very useful).

### DIFF
--- a/scripts/hooks/post-checkout
+++ b/scripts/hooks/post-checkout
@@ -1,2 +1,2 @@
 #!/bin/sh
-npm run npx hereby -- generate-diagnostics
+npx hereby -- generate-diagnostics
--- a/scripts/hooks/pre-commit
+++ b/scripts/hooks/pre-commit
@@ -1,0 +1,2 @@
+#!/bin/sh
+npx dprint fmt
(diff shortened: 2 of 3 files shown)


scripts/hooks/post-checkout
```diff
@@ -1,2 +1,2 @@
 #!/bin/sh
-npm run npx hereby -- generate-diagnostics
+npx hereby -- generate-diagnostics
```
weswigham, line 2: Maybe
```suggestion
npx hereby -- generate-diagnostics || true
```
So the checkout itself always actually works even if the post-checkout fails?

weswigham (approved):

scripts/hooks/post-checkout
jakebailey: Ah, yeah, both of these need that. That and I wanted to avoid `npx` because it has a very very slow startup time compared to what it's running.

jakebailey: Updated, and now it uses `node_modules/.bin/...`, which in my testing works fine on Windows (this all is run via git-bash).

jakebailey: Like wow

image

Issue #55795, opened 2023-09-20
Inconsistent literals inference in reverse mapped types


jfet97:
### 🔎 Search Terms

reverse mapped type literal inference

### 🕗 Version & Regression Information

This is the behavior in every version I tried.

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.2.2#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXwxAGcMAeAKHngBV4QAPQ1YI+Ab3gAcYdOAueCRhZUAcwDc8VMULABHaSRDAACj36CMwsfAC+egDTkAfAAocAIwBW8+AG0A0vBHwA1iACeORDQC6A6kdfPQBKAPFycjA8Eng4ImQIDHgAXgIZUzZKLnUBACJEHBw8oypFWXlsspllNV58i1gS7N1SukYYKAaoAC9m3RCI8njEjAiAenGAPQB+IA

### 💻 Code

```ts
declare function test<
  T extends { prop: string; nested: { nestedProp: string } },
>(obj: { [K in keyof T]: T[K] }): T;

const result = test({
  prop: "foo",
  nested: {
    nestedProp: "bar",
  },
  extra: "baz",
});
```

### 🙁 Actual behavior

The type parameter `T`, that is the type of `result` too, is inferred as:

```ts
{
  prop: "foo";
  nested: {
      nestedProp: string;
  };
  extra: string;
}
``` 

But it's unusual that `prop` is inferred as `"foo"` instead of `string`, especially since there are no `as` assertions being used and `T` is not a `const` type parameter.

### 🙂 Expected behavior

`T` inferred as:

```ts
{
  prop: string;
  nested: {
      nestedProp: string;
  };
  extra: string;
}
```

Andarist:
I think that the current behavior is expected (although somewhat surprising at first).

The basic rule is that a literal type is preserved when the inference target is a type variable and when it has a primitive constraint. For example:
```ts
declare function test1<T>(obj: { prop: T }): T
test1({ prop: 'foo' }) // string

declare function test2<T extends string>(obj: { prop: T }): T
test2({ prop: 'foo' }) // "foo"
```

[...]

PR #55807, opened 2023-09-21
Add a test for a property of reverse mapped type having a primitive type

Andarist:
closes https://github.com/microsoft/TypeScript/issues/55795

Please verify that this is actually the expected result. I think it is and explained why [here](https://github.com/microsoft/TypeScript/issues/55795#issuecomment-1727197771). 

I'm pretty sure that there is no test case like this in the codebase since I made an experiment to to obtain an apparent type of substituted mapped type property. Only 4 tests failed and none of them looked like this.


weswigham:
Thanks for the test - you explaination looks right to me!

Issue #55645, opened 2023-09-06
Intersected type variable with a recursive constraint fails to infer


Andarist:
### 🔎 Search Terms

intersection type variable parameter inference recursive constraint

### 🕗 Version & Regression Information

- This is the behavior in every version I tried

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.3.0-dev.20230906#code/C4TwDgpgBAysCGwIGED2A7AZgSwOYB4AVAQQGNhsMoIAPJdAEwGcongAnbdXAPigF4oAbwBQUaug4gA-AC4oJcpXRjWCJEzlQAShFKp2DfG07cANLHUoMOAoooYePANwiAvq5GhIlxBBikABYQALbwAsKqbH6a8rr6hsYcXLgWcH4BwWEu7p4MegA28OzQmACu6EpUpCV+ALLwQVwQ+Kr2ytR0EIwsJilmbWhYeJ30zL5ImaHwAzwAFPrDuPKEQ7ZQAGQT1ktEZA7oPACU8gDa7RgWqzZ4ALqei2xQXJgQ7CUMAArF8CEsgjUIPVGoFmnNROJulJ5AAiTCoVAwgbiaIaeQQ8RQeDo1SYiTQqAwgBGxSRuKgbmRFIGbiOnhebw+33YvyYrgA9OyoAA9aQiIA

### 💻 Code

```ts
type StateConfig<TAction extends string> = {
  entry?: TAction
  states?: Record<string, StateConfig<TAction>>;
};

type StateSchema = {
  states?: Record<string, StateSchema>;
};

declare function createMachine<
  TAction extends string,
  TConfig extends StateSchema,
>(config: TConfig & StateConfig<TAction>): [TAction, TConfig];

const inferredParams = createMachine({
  entry: "foo",
  states: {
    a: {
      entry: "bar",
    },
  },
});

inferredParams; // ["foo" | "bar", StateSchema]
```

### 🙁 Actual behavior

`TConfig` gets inferred as its constraint

### 🙂 Expected behavior

I'd expect `TConfig` to infer successfully 

### Additional information about the issue

[...]

Issue #55644, opened 2023-09-06
Surprising excess property check with recursive generic constraint


Andarist:
### 🔎 Search Terms

excess property check recursive generic contraint type variable inference

### 🕗 Version & Regression Information

- This is the behavior in every version I tried

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.3.0-dev.20230906#code/C4TwDgpgBAysCGwIwMYAsIFt5QLxQG8AoKKAZwSTIH4AuKAJQhQHsAnAEwB4K2BLAHYBzADSxKydFngA+ANxEAvgqIdmAG3htoAMwCuAlMD4sBUFNsQQAsvHSCIXACoBhUzr5CoEAB5IBHGTiVqgY2DIAFCTm7p70rrGiRACU8W4CHkIqFhBWtvYCEBHEpBACwGwg9ADaBFCgkPQARDosLE1QigC6ItEUVmT0JaRQ8EPRI97llTV1DRDNAEZaHd29k4rrnb2KyQpAA

### 💻 Code

```ts
type StateSchema = {
  states?: Record<string, StateSchema>;
};

declare function createMachine<TConfig extends StateSchema>(
  config: TConfig,
): TConfig;

createMachine({
  entry: [{ type: "foo" }],
  states: {
    a: {
      entry: [{ type: "bar" }],
    },
  },
});
```

### 🙁 Actual behavior

At the nested `entry` property the error is raised:
```
Object literal may only specify known properties, and 'entry' does not exist in type 'StateSchema'.(2353)
```
but it isn't raised at the `entry` property at the root of this object.

### 🙂 Expected behavior

I'd expect no excess property error to be raised with a recursive constraint like this.

### Additional information about the issue

This particular case is somewhat easily fixable by adding `[k: string]: unknown` to `StateSchema`. I still find the reported behavior to be a problem though.

andrewbranch:
[...]

PR #55811, opened 2023-09-21
Reverse mapped types with intersection constraint

### LINKED ISSUE #55927: Support for reverse mapped types with intersection constraint
### 🔍 Search Terms

"reverse mapped types", "intersection"

### ✅ Viability Checklist

### ⭐ Suggestion

I'd like to see reverse mapped types capabilities enhanced to support an intersection constraint, providing users with the ability to enable EPC (Excess Property Checking) on type parameters inference sites.

Reverse mapped types are a powerful tool. Currently TypeScript is able to reverse three kind of mapped types:

1. a homomorphic mapped type `{ [P in keyof T]: X }`
1. a mapped type `{ [P in K]: X }`, where `K` is a type parameter
1. a mapped type with an union constraint, as long as the union contains one of the constraints of the previous cases

This feature request asks support for a fourth case: a mapped type with an intersection constraint, as long as the intersection contains one of the constraints of the previous cases.

While the union constraint could be seen as a way to force the presence of some properties, the intersection one could be seen as a way to prevent the presence of extra properties.

May be closed by #55811.

### 📃 Motivating Example

In this example, taken from #12936, the type parameter `U` is the one that gets inferred by reversing the mapped type that has an intersection constraint: `keyof U & keyof T`. It should be inferred as `{ x: 1; y: "y"; z: string }` or, even better, as `{ x: 1; y: "y"; }` (with `z` stripped away because it wouldn't get through the application of the mapped type anyway). Either way, the application of the mapped type on the just inferred `U` imposes the absence of extra properties with respect to the ones declared in `T = {x: number, y: string}`.

```ts
const checkType = <T>() => <U extends T>(value: { [K in keyof U & keyof T]: U[K] }) => value;

const checked = checkType<{x: number, y: string}>()({
  x: 1,
  y: "y",
  z: "z", // <-- Object literal may only specify known properties, and 'z' does not exist in type '{ x: 1; y: "y"; }'.
});
```

### 💻 Use Cases

It's a way to enable EPC on type parameters when it's really needed. [Motivating comment](https://github.com/microsoft/TypeScript/issues/55645#issuecomment-1718359249).

1. What shortcomings exist with current approaches/workarounds?
One of the main workaround is the use of type functions like the following:
```ts
type NoExtra<T> = {
  [K in keyof T]: K extends keyof MyRefType ? T[K] : never
}
```
But then you get error messages like `Type 'number' is not assignable to type 'never'` that are bad for DX, expecially if compared with EPC ones.

What's good about this proposal, and the related PR, is the fact that it lets the user enable EPC on type parameters without fundamentally change how TS treats type parameters inference. It's just a nice side effect of enhancing reverse mapped types' capabilities.

jfet97:
## TLDR

This PR enhances reverse mapped types capabilities providing users with the ability to enable EPC (Excess Property Checking) on type parameters inference sites.

<h2 id="background">Background</h2>

This PR has come up trying to find a way to solve some problems discussed in #55645 and #55644, which are a simplification of what's going on in [this PR from XState](https://github.com/statelyai/xstate/pull/4262). In those cases the usual type parameters inference behaviour is not enough, mainly because it disables EPC (Excess Property Checking).

FIxes #55927

## Idea

Reverse mapped types are a powerful tool. Currently TypeScript is able to reverse three kind of mapped types:

1. a homomorphic mapped type `{ [P in keyof T]: X }`
1. a mapped type `{ [P in K]: X }`, where `K` is a type parameter
1. a mapped type with an union constraint, as long as the union contains one of the constraints of the previous cases

This PR adds a fourth case: a mapped type with an intersection constraint, as long as the intersection contains one of the constraints of the previous cases.

While the union constraint could be seen as a way to force the presence of some properties, the intersection one could be seen as a way to prevent the presence of extra properties.

### Example N. 1

Here the type parameter `T` is inferred by reversing the mapped type, and in this example in is inferred as `{ bar: 1; baz: "a"; }`. In the initial version of this PR `T` would have been inferred as `{ bar: 1; baz: "a"; extra: number }` in this erroneous case, but it seemed good to strip `extra` away because it wouldn't have survived the action of the mapped type anyway.

In both cases the properties of the parameter `s` are constrained to be exactly the ones already declared in `Foo`, thanks to the now supported intersection. In this way EPC comes into play if the argument contains extra properties:

```ts
interface Foo {
  bar: number;
  baz: string;
}

declare function useFoo<T extends Foo>(s: { [K in keyof T & keyof Foo]: T[K] } ): T

useFoo({ 
  bar: 1,
  baz: 'a',
  extra: 123, // <-- Object literal may only specify known properties, and 'extra' does not exist in type '{ bar: 1; baz: "a"; }'.
})
```

### Example N. 2

In the following example, taken from #12936, is the type parameter `U` the one that gets inferred by reversing the mapped type. It is inferred as `{ x: 1; y: "y"; }` because we strip `z: string` away from `U`. But even if we didn't, the application of the mapped type on the just inferred `U` imposes the absence of extra properties with respect to the ones declared in `T = {x: number, y: string}`.

```ts
const checkType = <T>() => <U extends T>(value: { [K in keyof U & keyof T]: U[K] }) => value;

const checked = checkType<{x: number, y: string}>()({
  x: 1,
  y: "y",
  z: "z", // <-- Object literal may only specify known properties, and 'z' does not exist in type '{ x: 1; y: "y"; }'.
});
```

## Tests

Two tests files have been added:

1. `reverseMappedTypeIntersectionConstraint`, in which the main capabilities brought by this PR have been tested, although I must admit I don't think there are enough test cases
1. `reverseMappedTypeLimitedConstraint`, in which the stripping behaviour on the inferred type parameters has been written down, but it is somehow subsumed by the previous test and for this reason it might be deleted altogether

## A note on EPC

What's good about this PR is the fact that it lets the user enable EPC on type parameters without fundamentally change how TS treats type parameters inference. It's just a nice side effect of enhancing reverse mapped types' capabilities.

## Remarks

This PR may help with some cases presented in #12936 as well.

#55794 could help improving what happens when a `const` type parameter is used, because there are some inconsistencies with `readonly` modifiers. 

## Acknowledgements

Huge thanks to @Andarist  for always having my back 😄.


jfet97:

sandersn:
@jfet97 can you open a new bug with your proposal where we can discuss it? For complex issues like this the PR itself is usually pretty short, but if it sits around for long the code gets stale.

jfet97:
Done 😄, it's #55927 @sandersn

weswigham:
This looks pretty good, but since it's definitely pretty high on the break risk scale (since it's intentionally adding new errors), let me run the extended tests and say that we'll probably want to merge this after we cut the 5.3 rc, so it can be in the nightly for awhile.

@typescript-bot test this
@typescript-bot run dt
@typescript-bot test top100

jfet97:
@weswigham thanks for your review, I'm glad you liked it!

To be honest I didn't expect any particular problems because:
1. reverse mapped types are sort of a niche/advanced feature
2. if you try to reverse a mapped type with an intersection constraint now the result is pretty useless, so I find it hard to believe that it ever occurred to someone to try 😂

I believe the tests not finding any errors supports my opinion, but there is no rush to add this feature of course. I'm just sharing my thoughts.

weswigham (approved):

src/compiler/checker.ts
```diff
@@ -13523,14 +13523,34 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         return instantiateType(instantiable, createTypeMapper([type.indexType, type.objectType], [getNumberLiteralType(0), createTupleType([replacement])]));
     }
 
+    function getLimitedConstraint(type: ReverseMappedType) {
```
RyanCavanaugh, line 13526: I'd really like a comment explaining a "limited constraint" is, maybe with an example or two. Otherwise LGTM

Andarist:
@jakebailey would you be so kind and create a TS playground for this one? :)

jakebailey:
@typescript-bot pack this

jfet97:
Another example can be found on [this playground](https://www.typescriptlang.org/play?ts=5.4.0-dev.20231201#code/C4TwDgpgBAysCGwIGED2A7AZgSwOZQF4oBvAKCignWACcQB+ALigGdbt1cBuc1hJFkygAlCAGNUNACYAeNjQ64ANLH4oMOXAD4eAXx6lQkVYggwxACwgBbeIRK82pwc1ETpc9pxVxT5q7Y6pPqkpFLiADbwNNCYAK7oYsDYGFBiMaYAsvCWHBAyACpoWHiUAB5I6FIsJkjFmloAFBIluMxkFADaANJQHFAA1hAgqJhQRRqlAGSDw6O16q0AuswTrT1LegCUq-V4Bi1saZP4ROkQWTkWeY0dfM7tvBTw7fcCr2LtulC6ukpPUAARl9-hQ-rwIBUaC8oABGABM-10WwOJx4AHp0VAAHr0ULhMRRGJQeKJZKpc6XXLoCAAeQisjWmnKlWqCz22maJ0eXV6-SGIzGTLwK3GHI2212aNCh2Ax1a9Kk9kpSGy1LpDNujjULB5FCgMOI4P1QJBvGNLOhzARKkxBoiEVQAHcIErALwbgCI94IomUnRUYrG40hAA) by @Andarist , to be compared with [the 5.3.2 one](https://www.typescriptlang.org/play?ts=5.3.2#code/C4TwDgpgBAysCGwIGED2A7AZgSwOZQF4oBvAKCignWACcQB+ALigGdbt1cBuc1hJFkygAlCAGNUNACYAeNjQ64ANLH4oMOXAD4eAXx6lQkVYggwxACwgBbeIRK82pwc1ETpc9pxVxT5q7Y6pPqkpFLiADbwNNCYAK7oYsDYGFBiMaYAsvCWHBAyACpoWHiUAB5I6FIsJkjFmloAFBIluMxkFADaANJQHFAA1hAgqJhQRRqlAGSDw6O16q0AuswTrT1LegCUq-V4Bi1saZP4ROkQWTkWeY0dfM7tvBTw7bpKT1AARq-vFG+8EAqNBeUAAjAAmd66LYHE48AD08KgAD16KFwmIojEoPFEslUudLrl0BAAPIRWRrTTlSrVBZ7bTNE6PLq9fpDEZjKl4FbjBkbba7OGhQ7AY6tclSeyEpDZYlkim3RxqFgsihQEHEf7qr4-XjamnA5gQlSIjURCKoADuEClgF4NwBEe8EYSKTpKEUjUaQgA). The latter fails to infer completely in the first scenario.

PR #49627, opened 2022-06-22
Ensure parameters are in scope when converting parameter/return types to type nodes

jakebailey:
Fixes #48783

PR #55820, opened 2023-09-21
Preserve original type parameter names more often when shadowed

### LINKED ISSUE #55575: Arrow function parameters using an alias and intersection are expanded inappropriately in `.d.ts`
### 🔎 Search Terms

arrow function type alias parameters expansion intersection

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about declarations and aliases.

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.3.0-dev.20230830#code/JYWwDg9gTgLgBAbzgBShMBnOBfOAzNEOAIigFMBDAYxgFoMyAbMm4gbgCgOyAPSWOFQgA7DPADCAVzEQQAZSYsYARjgBeOAB4A8mBjARAGgB8ACh4AuFGkw69B4YfwVGDY3ABkibAEp17hGxuPmh4PElhGgc4KRl5RRoAJjt9IzNLa3QMFIcnPBc3T28-QKA

### 💻 Code

```ts
import { Props } from "react-select";

export const CustomSelect1 = <Option,>(x: Props<Option, false> & {}) => {}
export function CustomSelect2<Option,>(x: Props<Option, false> & {}) {}
```

### 🙁 Actual behavior

The output `.d.ts` incorrectly expands the imported `Props` alias. Note that it has incorrectly resolved the default parameter. 

```ts
import { Props } from "react-select";
export declare const CustomSelect1: <Option>(x: Omit<import("react-select/base/dist/react-select-base.cjs").PublicBaseSelectProps<Option, false, import("react-select").GroupBase<Option>>, "inputValue" | "menuIsOpen" | "onChange" | "onInputChange" | "onMenuOpen" | "onMenuClose" | "value"> & Partial<import("react-select/base/dist/react-select-base.cjs").PublicBaseSelectProps<Option, false, import("react-select").GroupBase<Option>>> & import("react-select/dist/declarations/src/useStateManager").StateManagerAdditionalProps<Option>) => void;
export declare function CustomSelect2<Option>(x: Props<Option, false> & {}): void;
```

### 🙂 Expected behavior

The imported type alias `Props` _must not_ be expanded, if the upstream package changes its defaults this will cause confusion because dependencies relying on the defaults will not be able to use the new default.

```ts
import { Props } from "react-select";
export declare const CustomSelect1: <Option>(x: Props<Option, false> & {}) => void;
export declare function CustomSelect2<Option>(x: Props<Option, false> & {}): void;
```

This problem became _much_ worse with the advent of 4.2.3, with prior version of typescript only mistakenly expanding the default argument.

Additionally this causes difficult-to-reproduce issues producing errors like

> The inferred type of 'CustomSelect' cannot be named without a reference to '../../../node_modules/react-select/dist/declarations/src/useStateManager'. This is likely not portable. A type annotation is necessary.ts(2742)

because the expanded alias uses types that aren't imported.

### LINKED ISSUE #55653: Type parameters are renamed in declarations
### 🔎 Search Terms

type parameters declaration rename

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.2.2#code/MYGwhgzhAECC0G9oF8CwAoAbmATtARtALzQA8sANAHwAUAHgFxwCUxV0dQA

### 💻 Code

```ts
class A { }
var b = <A,>(x: A) => x
```

### 🙁 Actual behavior

Declaration file type for `b` is `declare var b: <A_1>(x: A_1) => A_1;`

### 🙂 Expected behavior

Declaration file type for `b` should be `declare var b: <A>(x: A) => A;` as written in source code. There is no reason to rename the type parameter

_No response_

jakebailey:
Fixes #55653
Fixes #55575

This turned out to be extremely related to #49627. Our original mechanism to solve the problem for type parameters was to be very conservative and rename when we think there might be a problem.

In this PR, I extend the mechanism from #49627 to the type parameters, introducing a second symbol table and fake scope to hold them, as we can't merge the symbols that share the same name without modifying them.

This turns out to have a positive impact on some existing baselines; see https://github.com/microsoft/TypeScript/pull/55820/commits/83275d2c18cf41fea0a22680a0d6d743bc8c9ebd for the actual baseline diff.


jakebailey:
@typescript-bot test top100
@typescript-bot run dt
@typescript-bot pack this

src/compiler/types.ts
```diff
@@ -5987,7 +5987,7 @@ export interface NodeLinks {
     decoratorSignature?: Signature;     // Signature for decorator as if invoked by the runtime.
     spreadIndices?: { first: number | undefined, last: number | undefined }; // Indices of first and last spread elements in array literal
     parameterInitializerContainsUndefined?: boolean; // True if this is a parameter declaration whose type annotation contains "undefined".
-    fakeScopeForSignatureDeclaration?: boolean; // True if this is a fake scope injected into an enclosing declaration chain.
+    fakeScopeForSignatureDeclaration?: "params" | "typeParams"; // If present, this is a fake scope injected into an enclosing declaration chain.
```
jakebailey, line 5990: Not sure if this should be an enum, I just used a string here as the least annoying way to get two truthy values.

fatcerberus:
> Preserve shadowed type parameter names

Looking at #55653, technically it's the *other* type that's being shadowed by the type parameter.  If it were the other way around then the current behavior might actually be correct :wink:

fatcerberus:
Here's a test, in case this isn't already covered:
```ts
class A { }
const a = new A();
const b = <A,>(x: A) => a
```

The type parameter still needs to be renamed here.

Also here's a fun bit of chicanery:
image

jakebailey:
This PR makes the work, check the declaration tab here: https://www.staging-typescript.org/play?ts=5.3.0-pr-55820-7#code/MYGwhgzhAECC0G9oF8BQwD2A7CAXaY0AvNFgKYDucAFAJQDc62e0ARsdADywA0AfNQAeALji1ifAkA

That's actually the test case I added first.

Unfortunately hover doesn't set the flag from: https://github.com/microsoft/TypeScript/pull/55821 I don't think. I've been wondering if we should just drop this flag, so we are consistent...

fatcerberus:
> `declare const b: <A>(x: A) => globalThis.A;`

This is a legal declaration?  I didn't know `globalThis` existed in type-space.

Also if the file is a module the output is... questionable.  I'd rather just have the type parameter be renamed here.
```ts
export declare class A {
}
export declare const a: A;
export declare const b: <A>(x: A) => import("./input").A;
```

jakebailey:
Write:

```ts
class A { }
const a = new A();
function b<A,>(x: A) { return a; }
```

And you'll see that we already emit `globalThis.A`; the only reason we don't now is because declaration emit and _signature to node_ emit are different.

fatcerberus:
Hmm, you're right, thanks.  That self-import in the case of a module still bugs me *a lot*, though.

tests/baselines/reference/declarationEmitTypeParameterNameShadowedInternally.js
```diff
@@ -0,0 +1,22 @@
...
+exports.foo = void 0;
+var foo = function (x) {
+    var inner = function (y) { return [x, y]; };
+    return inner;
+};
+exports.foo = foo;
+
+
+//// [declarationEmitTypeParameterNameShadowedInternally.d.ts]
+export declare const foo: <T>(x: T) => <T_1>(y: T_1) => readonly [T, T_1];
```
jakebailey, line 22: Actually I thought this test had failed, but actually, I don't think it did fail; this matches main, I think...

tests/baselines/reference/declarationEmitTypeParameterNameShadowedInternally.types
```diff
@@ -0,0 +1,21 @@
...
+>inner : <T>(y: T) => readonly [T, T]
+><T,>(y: T) => [x, y] as const : <T>(y: T) => readonly [T, T]
+>y : T
+>[x, y] as const : readonly [T, T]
+>[x, y] : readonly [T, T]
+>x : T
+>y : T
+
+	return inner;
+>inner : <T>(y: T) => readonly [T, T]
```
jakebailey, line 19: Here you can see that the tooltip is very wrong because we don't do the right renaming there...

jakebailey: Nothing new, though:
image

weswigham (approved):
I know I've said it elsewhere - Not a huge fan of the double-scope thing (or basically rebuilding the signature's `.locals` table by hand in two parts), but this isn't *much* worse than what we're already doing, and fixing that is... another issue (ungh, the reading through this, my own PR, and the original PR adding this codepath made me aware of some problems in caching caused by this approach that we really aughta fix - we're caching results between unrelated queries). This is a meaningful improvement for now (and technically will hide the problem a bit - two cache nodes means you're slightly less likely to hit the incorrect cache key issue), so we should probably take it. Once it's in, I'll make a followup with how I'd prefer it to be structured, ignoring any perf issues that causes, and once the perf of that is made acceptable via other changes I've been looking into, we can try that out for size. This all just feels.... too complicated and bespoke, for what little it does.

jakebailey:
Dunno why CI isn't running...

PR #55822, opened 2023-09-22
Partially revert PR 55723 to fix OOM in testing

jakebailey:
#55723 fixes an actual bug in our testing infrastructure, but doubles memory usage on my machine, causing OOMs where I hit the max memory limit of my system and the OS starts killing processes.

This PR effectively reverts that change. I hate this, but I hate OOMing my machine more and #55723 was not supposed to have caused any behavior changes.

### DIFF
--- a/src/harness/fakesHosts.ts
+++ b/src/harness/fakesHosts.ts
@@ -366,7 +366,12 @@ export class CompilerHost implements ts.CompilerHost {
         // reused across multiple tests. In that case, we cache the SourceFile we parse
         // so that it can be reused across multiple tests to avoid the cost of
         // repeatedly parsing the same file over and over (such as lib.d.ts).
-        const cacheKey = this.vfs.shadowRoot && `SourceFile[languageVersionOrOptions=${languageVersionOrOptions !== undefined ? JSON.stringify(languageVersionOrOptions) : undefined},setParentNodes=${this._setParentNodes}]`;
+
+        // TODO(jakebailey): the below is totally wrong; languageVersionOrOptions can be an object,
+        // and so any options bag will be keyed as "[object Object]", and we'll incorrectly share
+        // SourceFiles parsed with different options. But fixing this doesn't expose any bugs and
+        // doubles the memory usage of a test run, so I'm leaving it for now.
+        const cacheKey = this.vfs.shadowRoot && `SourceFile[languageVersionOrOptions=${languageVersionOrOptions},setParentNodes=${this._setParentNodes}]`;
         if (cacheKey) {
             const meta = this.vfs.filemeta(canonicalFileName);
             const sourceFileFromMetadata = meta.get(cacheKey) as ts.SourceFile | undefined;


andrewbranch:
What's the correct fix? What's the lifetime of this cache?

jakebailey:
I have no idea what the correct fix is; it seems like this cache lives forever. It's not a simple Map either so I nope'd out last night, but I'll stare at it harder.

andrewbranch:
I guess the idea is to cache mostly lib.d.ts files but it attempts to cache literally everything, failing only because of the broken key bug you fixed. Failing something very simple like only caching files in certain directories or with lib-like file names, maybe a simple MRU cache would do it.

andrewbranch (approved):

jakebailey:
> I guess the idea is to cache mostly lib.d.ts files but it attempts to cache literally everything, failing only because of the broken key bug you fixed. Failing something very simple like only caching files in certain directories or with lib-like file names, maybe a simple MRU cache would do it.

The problem is that it actually succeeds and returns the same AST even when the options change... but miraculously it doesn't break anything, probably the files in shadow roots don't actually change on these option differences? They sure would depending on the JSDoc mode :(

Now that TS Congress is over I'm going to look at this harder before doing the pure revert. But it's possible Sheetal will save the day with some insight faster than me 😄

andrewbranch:
I think it’s not broken because of the extra `getFullText() === content` check. So stuff in the cache is being overwritten and then not used when it doesn’t match.

jakebailey:
Gating this on "only cache if it's a lib.d.ts file" still produces the problem, unfortunately. I see no reason why this memory should be sticking around, either; the compiler tests explicitly try and drop the only reference to this stuff.

jakebailey:
Should I just merge this as-is to get memory usage back to "normal"? I still want to figure out how to do this right but I definitely didn't intend this when I sent the original PR :(

sheetalkamat:
Vfs is caching only typescript lib files because those are the only ones on the shadow root and available. So I think for those files it doesn't matter if they account for script target or not since the result is going to be same?

jakebailey:
Unfortunately WSL's kernel is too old for me to properly measure this, but watching htop's total system memory usage by eye, I can see:

- With the original code before my original PR, my memory usage goes from 2.87GB to 17.3GB
- After my original PR (main), I see 3.19GB grow to 31.3GB
- With main but only allowing caching on `lib.d.ts` files, I see 3.18GB grow to 33.0GB

I've so far not been able to find any way to fix this while keeping an accurate cache key.

> Vfs is caching only typescript lib files because those are the only ones on the shadow root and available. So I think for those files it doesn't matter if they account for script target or not since the result is going to be same?

I observed it caching files that aren't `lib.d.ts` files, unfortunately. And even so, lib.d.ts files can be parsed with differing options (e.g. JSDocParsingMode), so reusing them may not actually be safe.

jakebailey:
I'm just going to merge this for now as it's effectively a revert; the memory usage is really hurting me while working on other other changes :(

PR #55831, opened 2023-09-22
fix(55823) - Remove Braces From Arrow function fails when using satisfies 

### LINKED ISSUE #55823: Remove Braces From Arrow function fails when using satisfies
### 🔎 Search Terms

remove braces from arrow function satisfies

The "Remove Braces From Arrow Function" code action generates invalid JS when the function returns an object that features a `satisfies` declaration.

Before:
```ts
() => {
  return {} satisfies any
}
```

After:
```ts
() => {} satisfies any
```

Expected:
```ts
() => ({} satisfies any)
```

### ⏯ Playground Link

_No response_

### 💻 Code

_No response_

### 🙁 Actual behavior

Missing parens.

### 🙂 Expected behavior

Correct syntax.

_No response_

a-tarasyuk:
Fixes #55823

### DIFF
--- a/src/services/utilities.ts
+++ b/src/services/utilities.ts
@@ -226,6 +226,7 @@ import {
     isRequireVariableStatement,
     isRightSideOfQualifiedNameOrPropertyAccess,
     isRootedDiskPath,
+    isSatisfiesExpression,
     isSetAccessorDeclaration,
     isSourceFile,
     isSourceFileJS,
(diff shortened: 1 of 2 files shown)


jakebailey (approved):

jakebailey:
Of course this makes me wonder if we should sit down one day and find all uses of `isAsExpression` and verify that they don't also need an `isSatisfiesExpression` 😄

PR #55836, opened 2023-09-22
Make unchangedTextChangeRange const

jakebailey:
Noticed this while working close to this declaration. A bit of bisecting shows that it was originally `var` and was switched to `let` in a bulk PR many many many years ago, but probably should have been const.


jakebailey:
After this PR, the only mutable export we have is `sys`.

jakebailey:
This could also technically be `/* @__PURE__ */` but esbuild doesn't detect that it's not used in tsc. I blame the barrels.

DanielRosenwasser (approved):

PR #55842, opened 2023-09-23
fix(55841) - Array/Tuple indexed by `never` behaving like indexed by `number`

### LINKED ISSUE #55841: `never` as keyof tuples/arrays behaves the same as `number`
### 🔎 Search Terms

tuple keyof never, tuple[never], array keyof never

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about this issue

Versions I tried:
- 5.2.2
- nightly (playground)

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.3.0-dev.20230923#code/C4TwDgpgBAglC8UDeBfA2gOwgNwgJwF0oB6YqQUHIotc8AoUSKAIQWSgEMAuKAZ2DwCWGAOZR01fEVIUqOfPXDQAwqzQFMcwiTKUJdBtAAiqvoJHq9UsoBlyXvyHCFjAKKtTDtRppWot9yKdoADFWAAp-UQAfKgBXAFsAI3wASk9LKG1fOzMo2MT8IA

### 💻 Code

```ts
type A = {}[never] // ✅ never
type B = { a: string }[never] // ✅ never
type C = [][never] // ✅ never
type D = [string][never] // ❌ string
type E = string[][never] // ❌ string
type F = (string | number)[][never]  // ❌ string | number
```

### 🙁 Actual behavior

After using a `never` key on tuples/arrays you get back the entirety of the tuple/array elements type.

### 🙂 Expected behavior

After using a `never` key on tuples/arrays you get back `never`, as when dealing with objects.

Not sure if this has historical reasons to behave like it does, but couldn't find any specific info about it.

jfet97:
Fixes #55841 by doing the check before the one about index signatures.

Closed because: even if `{ a: 42 }[never]` is already computed as `never`, something like `Record<string, boolean>[never]` is `boolean`, not `never`. So the current behaviour of arrays/tuples is not consistent with the one of object literals, but is consistent with other cases.

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -17783,6 +17783,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             if (objectType.flags & (TypeFlags.Any | TypeFlags.Never)) {
                 return objectType;
             }
+            if (indexType.flags & TypeFlags.Never) {
+                return neverType;
+            }
             // If no index signature is applicable, we default to the string index signature. In effect, this means the string
             // index signature applies even when accessing with a symbol-like type.
             const indexInfo = getApplicableIndexInfo(objectType, indexType) || getIndexInfoOfType(objectType, stringType);
@@ -17813,9 +17816,6 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     return getUnionType([indexInfo.type, missingType]);
                 }
                 return indexInfo.type;
-            }
-            if (indexType.flags & TypeFlags.Never) {
-                return neverType;
             }
             if (isJSLiteralType(objectType)) {
                 return anyType;
--- a/tests/baselines/reference/indexIntoArrayTupleObjectWithNever.js
+++ b/tests/baselines/reference/indexIntoArrayTupleObjectWithNever.js
@@ -1,0 +1,10 @@
+//// [tests/cases/compiler/indexIntoArrayTupleObjectWithNever.ts] ////
+
+//// [indexIntoArrayTupleObjectWithNever.ts]
+type A = { a: 42 }[never]
+
+type B = [42][never]
+
+type C = Array<42>[never]
+
+//// [indexIntoArrayTupleObjectWithNever.js]
--- a/tests/baselines/reference/indexIntoArrayTupleObjectWithNever.symbols
+++ b/tests/baselines/reference/indexIntoArrayTupleObjectWithNever.symbols
@@ -1,0 +1,14 @@
+//// [tests/cases/compiler/indexIntoArrayTupleObjectWithNever.ts] ////
+
+=== indexIntoArrayTupleObjectWithNever.ts ===
+type A = { a: 42 }[never]
+>A : Symbol(A, Decl(indexIntoArrayTupleObjectWithNever.ts, 0, 0))
+>a : Symbol(a, Decl(indexIntoArrayTupleObjectWithNever.ts, 0, 10))
+
+type B = [42][never]
+>B : Symbol(B, Decl(indexIntoArrayTupleObjectWithNever.ts, 0, 25))
+
+type C = Array<42>[never]
+>C : Symbol(C, Decl(indexIntoArrayTupleObjectWithNever.ts, 2, 20))
+>Array : Symbol(Array, Decl(lib.es5.d.ts, --, --), Decl(lib.es5.d.ts, --, --))
+
--- a/tests/baselines/reference/indexIntoArrayTupleObjectWithNever.types
+++ b/tests/baselines/reference/indexIntoArrayTupleObjectWithNever.types
@@ -1,0 +1,13 @@
+//// [tests/cases/compiler/indexIntoArrayTupleObjectWithNever.ts] ////
+
+=== indexIntoArrayTupleObjectWithNever.ts ===
+type A = { a: 42 }[never]
+>A : never
+>a : 42
+
+type B = [42][never]
+>B : never
+
+type C = Array<42>[never]
+>C : never
+
--- a/tests/cases/compiler/indexIntoArrayTupleObjectWithNever.ts
+++ b/tests/cases/compiler/indexIntoArrayTupleObjectWithNever.ts
@@ -1,0 +1,5 @@
+type A = { a: 42 }[never]
+
+type B = [42][never]
+
+type C = Array<42>[never]
--- a/tests/cases/fourslash/quickInfoNestedGenericCalls.ts
+++ b/tests/cases/fourslash/quickInfoNestedGenericCalls.ts
@@ -14,9 +14,9 @@ verify.quickInfoAt("1", `const m: <"foo">(s: {
 
 // the exact generic type params are not important in this test (they could change with changes to the inference algorithm)
 // it's important though that they both display the same types
-verify.quickInfoAt("2", `const $: <unknown, string>(s: string) => {
-    $: unknown;
+verify.quickInfoAt("2", `const $: <never, never>(s: never) => {
+    $: never;
 }`);
-verify.quickInfoAt("3", `const $: <unknown, string>(s: string) => {
-    $: unknown;
+verify.quickInfoAt("3", `const $: <never, never>(s: never) => {
+    $: never;
 }`);


DanielRosenwasser:
@typescript-bot pack this
@typescript-bot test this
@typescript-bot test top100
@typescript-bot test tsserver top300
@typescript-bot run dt

DanielRosenwasser:
@typescript-bot test top300

gabritto:
@typescript-bot test top300

gabritto:
@typescript-bot run DT

src/compiler/checker.ts
```diff
@@ -18193,6 +18193,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             if (objectType.flags & (TypeFlags.Any | TypeFlags.Never)) {
                 return objectType;
             }
+            if (indexType.flags & TypeFlags.Never) {
```
Andarist, line 18196: XState failures would likely get fixed if you'd bring back the other piece of code and change this one to:
```suggestion
            if (indexType === neverType) {
```

XState relies on propagating `silentNeverType`. Previously `[silentNeverType][silentNeverType]` was evaluated to `silentNeverType` but with this change its `neverType`. Note that for this special never type (and there are some more of them in the compiler) you should go through the index signature lookup and not return `indexType` here directly (that just happens to match what indexing with `silentNeverType` evaluates to in our case).

Andarist:
I think it would also be best to forbid "simplifying" the `never` `indexType` to `string` at the top of `getIndexedAccessTypeOrUndefined`. That's what makes those 2 to behave differently:
```ts
type A = Record<string, boolean>[never] // boolean
type B = Record<`--${string}`, boolean>[never] // never
```

src/compiler/checker.ts
jfet97: @Andarist can you expand further on how/why XState relies on propagating `silentNeverType`?

P.S. I implemented this check as you suggested :D

jfet97: P.P.S. Currently I've not bringed back the other check after the index signature one, do you really think it is needed or I can ask to rerun the tests?

Andarist: > @Andarist can you expand further on how/why XState relies on propagating silentNeverType?

To be honest, I have no idea. It's a magic internal type and I never had time to explore how this works.

> P.P.S. Currently I've not bringed back the other check after the index signature one, do you really think it is needed or I can ask to rerun the tests?

I'd check what happens now with `type Result = {}[never]`. It should be `never` but I have a feeling that it might be `any` right now or an error type (a special kind of `any`).

Andarist:
@jfet97 the case above is still weirdly inconsistent right now.

In general, I'm also not sure if this is the right change - but that's really something that the TS should comment on. 

What does indexing a type mean? What does it have to check? The argument is to be made that it has to check if `IndexType` is assignable to `keyof ObjectType`. Since `never` is assignable to just everything... shouldn't it always be allowed to index with `never`? 

Perhaps a different change should be made that would make `{ a: 42 }[never]` valid? I think it would make sense to either allow it at all times or "disallow" it at all times (what this PR tries to accomplish at the moment). The fact that index infos are consulted after missing a concrete prop match is kind of an implementation detail (one that is very much observable though).

gabritto:
@typescript-bot pack this

src/compiler/checker.ts
```diff
@@ -18196,6 +18196,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             if (objectType.flags & (TypeFlags.Any | TypeFlags.Never)) {
                 return objectType;
             }
+            if (indexType === neverType) {
```
gabritto, line 18199: I think you can try doing something like we have in https://github.dev/microsoft/TypeScript/blob/3de2cdce4d62fab6c8fb9e8a3f78eb33125d6cd5/src/compiler/checker.ts#L17574-L17575,
and do something like `if (indexType.flags & TypeFlags.Never) { return indexType === silentNeverType ? silentNeverType : neverType; }`

jfet97: @gabritto may I ask you to run the top 300 tests agaib? I'd like to see if the current version is enough or if we need to handle the `silentNeverType` as you have suggested here. I checked `type Result = {}[never]` plus some friends in the playground and I haven't found any hiccup.

Andarist: The issue has arised because the meaning of this has changed `[silentNeverType][silentNeverType]`. It's true that the suggested change would keep the old behavior and keep "propagating" the `silentNeverType` but there are two other scenarios to consider here:
- what should be the evaluation of `[T][silentNeverType]`? Is it `silentNeverType` or `T`? It's hard to reason about `silentNeverType` because it's an internal type so we'd have to deeply understand how it is used to determine which is better. I proposed the `indexType === neverType` check specifically to avoid this - erring on the safer side to avoid breaking changes related to those complex scenarios.
- there are also other special never types internally. I thought that perhaps it would be better to keep the old behavior for them too but for example this one is quite scary and I don't think it's the right behavior:
```ts
const obj = {
  arr: [],
};

const objWithIndex: Record<string, boolean> = {};

const el = obj.arr[0];
const result = objWithIndex[el];
//    ^? const result: boolean
```

I'm not sure why in the example above `implicitNeverType[]` gets created and not `[]`.

jfet97:
@Andarist We are of the same mind on this, I feel TS is quite inconsistent in these circumstances and I'm not sure this is the right change too. I re-opened this PR just because the correspondent issue has been accepted as a bug. You are right on the fact that this PR is not exactly about arrays and tuples and it is related with index signatures.

OTOH I suspect that making `{ a: 42 }[never]` valid wouldn't be the right thing to do either. Some folks activly exploit this behaviour, [like @danvk does here with the `LooseKey` type](https://effectivetypescript.com/2020/12/09/gentips-2-intersect/).

> What does indexing a type mean?

42, of course :laughing:.

gabritto:
@typescript-bot test top300

gabritto:
Top tests all pass, but I do find [this example](https://github.com/microsoft/TypeScript/pull/55842#discussion_r1490703349) kinda bad and it probably should be `never` instead of `boolean`?

src/compiler/checker.ts
Andarist: I'm not sure what I was thinking about in the comment above. `type Result = {}[never]` is still `never` here because that `never` is just a regular `never` (so with the proposed change it just enters the branch that you have moved).

src/compiler/checker.ts
```diff
@@ -18196,6 +18196,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             if (objectType.flags & (TypeFlags.Any | TypeFlags.Never)) {
                 return objectType;
             }
+            if (indexType === neverType || indexType === implicitNeverType) {
```
Andarist, line 18199: I'd probably go with this:
```suggestion
            if (indexType.flags & TypeFlags.Never && indexType !== silentNeverType) {
```

jfet97: In that way we would change the indexing behaviour of `unreachableNeverType` too. Are you sure? XD

Andarist: Am I sure? Not at all. It seems reasonable to treat that the same way as the regular never here but who knows?

jfet97: @Andarist 

I tried the following:
```ts
const obj = {
  arr: [],
};

const objWithIndex: Record<string, boolean> = {};

function testUnreachableNeverType(smt: number | string) {
  if(typeof smt === "number") {}
  else if(typeof smt === "string") {}
  else {
    const result = objWithIndex[smt]
  }
}
```

My supposition was that the type of `smt` is `unreachableNeverType`, therefore the type of `result` should have been `boolean` without handling `unreachableNeverType` in that `if` condition, but it is already `never` 🧐

I probably have to come up with a better example to try to have an inconsistency with `unreachableNeverType` too.

jfet97:
@gabritto I think now it's fixed. What do you think?

gabritto:
@typescript-bot pack this

gabritto:
@typescript-bot run DT
@typescript-bot test top300

jfet97:
@gabritto I think this error is right: `pgSettingsKey` has type `undefined`, therefore inside `if (pgSubscriberKey != null) {` it has type `never` and indexing with `never` now produces `never`.

jfet97:
About the meaning of indexing with `never`: if we have a record/object/tuple/array/whatever indexed by something that has type `never` I'm already in an absurd situation, like the default case of an exhaustive `switch`. I think we could apply the _ex falso quodlibet principle_, therefore whatever resulting type would be correct. Propagating the `never` itself it's just an option, that probably makes some sense.

Issue #55847, opened 2023-09-24
Intrinsic string mapping type with type parameters is incorrectly resolved in conditional types


sankantsu:
# Bug Report

### 🔎 Search Terms

lowercase, uppercase, intrinsic, conditional type

### 🕗 Version & Regression Information

- This changed between versions 5.0.4 and 5.1.0-beta.
- 5.2.2 and nightly still have the same problem.
- I also reviewed the FAQ for entries about lowercase, uppercase, intrinsic and conditional types.

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.1.6#code/C4TwDgpgBAggPAZQHxQLxQDIHsDuEBOAxgIYDOEiAZKcPgJYB2A5ihAB7AQMAmpUARADMsWflAD8UAIxQAXFAAMAKAD0KqFAB64paEhQAQohTpseImQoJqtRiyjtOPPgANBAEgDeNeswC+LhLScoqq6lo6SgA2EMBQbFLy8EIi-CbBag74+Fj40bHxAEzyRimi6TKZBDn4QA

### 💻 Code

```ts
type A<S> = Lowercase<S&string> extends "foo" ? 1 : 0
//  ^?  type A<S> = 0
type B<S> = Lowercase<S&string> extends `f${string}` ? 1 : 0
//  ^? type B<S> = 0

let x1: A<"foo"> = 1 // error
let x2: B<"foo"> = 1 // error
```

### 🙁 Actual behavior

`A<"foo"> = 0` and `B<"foo"> = 0`.
Actually, these conditional branches are resolved at the time of type alias definition.
Then, any instantiation of `A` and `B` is `0`

### 🙂 Expected behavior

Both `A<"foo">` and `B<"foo">` should be `1`.
Conditional types in A and B should not be resolved during type alias definition, but should be deferred until an instantiation.

### Related

#52102 is similar.
This issue is about incorrect "simplification" involving intrinsic string mapping types.

### More info

I did some investigation in codebase.
It looks like that the cause of the problem is handling of string mapping type in `getGenericObjectFlags` or `isPatternLiteralPlaceholderType` which was introduced in #52836 .

https://github.dev/microsoft/TypeScript/blob/97147915ab667a52e31ac743843786cbc9049559/src/compiler/checker.ts#L17967

MartinJohns:
Workaround:

```typescript
type A<S extends string> = Lowercase<S> extends "foo" ? 1 : 0
[...]

PR #55856, opened 2023-09-25
Fixed an issue with string mappings over generic intersections not being deferred in conditional types

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/55847

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -17906,16 +17906,19 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             accessNode;
     }
 
-    function isPatternLiteralPlaceholderType(type: Type): boolean {
+    function isPatternLiteralPlaceholderType(type: Type, ignoreGenericIntersections = false): boolean {
         if (type.flags & TypeFlags.Intersection) {
-            return some((type as IntersectionType).types, t => !!(t.flags & (TypeFlags.Literal | TypeFlags.Null | TypeFlags.Undefined)) || isPatternLiteralPlaceholderType(t));
+            if (ignoreGenericIntersections && isGenericType(type)) {
+                return false;
+            }
+            return some((type as IntersectionType).types, t => !!(t.flags & (TypeFlags.Literal | TypeFlags.Nullable)) || isPatternLiteralPlaceholderType(t, ignoreGenericIntersections));
         }
-        return !!(type.flags & (TypeFlags.Any | TypeFlags.String | TypeFlags.Number | TypeFlags.BigInt)) || isPatternLiteralType(type);
+        return !!(type.flags & (TypeFlags.Any | TypeFlags.String | TypeFlags.Number | TypeFlags.BigInt)) || isPatternLiteralType(type, ignoreGenericIntersections);
     }
 
-    function isPatternLiteralType(type: Type) {
-        return !!(type.flags & TypeFlags.TemplateLiteral) && every((type as TemplateLiteralType).types, isPatternLiteralPlaceholderType) ||
-            !!(type.flags & TypeFlags.StringMapping) && everyContainedType((type as StringMappingType).type, isPatternLiteralPlaceholderType);
+    function isPatternLiteralType(type: Type, ignoreGenericIntersections = false) {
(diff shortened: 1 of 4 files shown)


jakebailey:
This looks good to me (having touched this recently), but:

@typescript-bot test top200
@typescript-bot run dt
@typescript-bot pack this

sankantsu:
Thank you for creating PR!

It seems still fails in some (very artificial) examples...

```
type E<S> = Lowercase<`f${S & string}` & `${S & string}f`>
>E : Lowercase<`f${S & string}` & `${S & string}f`>

type F = E<""> extends "f" ? 1 : 0
>F : 1

type G<S> = E<S> extends "f" ? 1 : 0
>G : 0

let x5: G<""> = 1;
>x5 : 0
>1 : 1
```

`E<"">` is `"f"`, so `G<"">` should be `1`

Andarist:
Ah, ye - I see why it might fail. I'll try to improve the fix in a moment to handle more crazy scenarios like this recursively.

src/compiler/checker.ts
```diff
@@ -17946,7 +17949,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             return (type as SubstitutionType).objectFlags & ObjectFlags.IsGenericType;
         }
         return (type.flags & TypeFlags.InstantiableNonPrimitive || isGenericMappedType(type) || isGenericTupleType(type) ? ObjectFlags.IsGenericObjectType : 0) |
-            (type.flags & (TypeFlags.InstantiableNonPrimitive | TypeFlags.Index | TypeFlags.StringMapping) && !isPatternLiteralType(type) ? ObjectFlags.IsGenericIndexType : 0);
+            (type.flags & (TypeFlags.InstantiableNonPrimitive | TypeFlags.Index | TypeFlags.StringMapping) && !isPatternLiteralType(type, /*ignoreGenericIntersections*/ true) ? ObjectFlags.IsGenericIndexType : 0);
```
Andarist, line 17952: alternatives that I considered:
- using other names for this parameter (`deferGenerics`, `deferGenericIntersections`?)
- writing "wrapper functions" for `isPatternLiteralType` and `isPatternLiteralPlaceholderType`, something like `isGenericPatternLiteralType` and `isGenericPatternLiteralPlaceholderType`. I'd still have to handle common concerns in those new functions though - so I concluded that an extra parameter in the existing ones makes sense the most to me.

src/compiler/checker.ts
```diff
@@ -17906,16 +17906,19 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             accessNode;
     }
 
-    function isPatternLiteralPlaceholderType(type: Type): boolean {
+    function isPatternLiteralPlaceholderType(type: Type, ignoreGenericIntersections = false): boolean {
         if (type.flags & TypeFlags.Intersection) {
-            return some((type as IntersectionType).types, t => !!(t.flags & (TypeFlags.Literal | TypeFlags.Null | TypeFlags.Undefined)) || isPatternLiteralPlaceholderType(t));
```
ahejlsberg, line 17911: I think this entire fix can be simplified to simply adding a check that the type isn't generic:

```ts
return !isGenericType(type) && some(...);
```

Generic types should never be classified as placeholders since upon instantiation they may become something completely different.

ahejlsberg (changes requested):

src/compiler/checker.ts
Andarist: Oh, that's cool! Thanks for the tip. I was worried that `addSpans` would fail if I did something like this. It seems that (somewhat confusingly) `isGenericIndexType` already returns `true` for an intersection like this so `addSpans` is covered.

ahejlsberg (approved):

jakebailey:
LGTM but just going to double check:

@typescript-bot test top200
@typescript-bot run dt

jakebailey (approved):

sandersn:
@jakebailey @ahejlsberg is this a fix intended for 5.3, or should we wait until 5.4?

jakebailey:
Probably meant for 5.3, not sure why we didn't merge it as we both approved

Issue #55858, opened 2023-09-25
JSX is still not transformed to `React.createElement` when inside spread with target >=`ES2018`


bob-difronzo:
### 🔎 Search Terms

JSX, spread, React.createElement

### 🕗 Version & Regression Information

Issue is first seen in `5.1.6`, and forward to `5.2.2` and the current nightly version.

### ⏯ Playground Link

https://www.typescriptlang.org/play?target=5&ts=5.2.2#code/DwYQ9gtgDgBATgSwOYAsAuBeA3sAJggNwD4APYAen2IF8YsA6RrGAdzjADskAuGPQ0hSpFq1ckQCwAKCA

### 💻 Code

```tsx
<Comp right={<div>x</div>} {...{ wrong: <div>x</div>}}/>
```

### 🙁 Actual behavior

```js
React.createElement(Comp, { right: React.createElement("div", null, "x"), wrong: <div>x</div> });
```

### 🙂 Expected behavior

```js
React.createElement(Comp, { right: React.createElement("div", null, "x"), wrong: React.createElement("div", null, "x") });
```

### Additional information about the issue

The same bug seems to have been fixed in #54425, but only when the target is `ES2017` or below. When target is set to `ES2018` or above the issue still exists.

Andarist:
This one is also related: https://github.com/microsoft/TypeScript/pull/55008 . Both of the mentioned fixes touched `transformJsxAttributesToExpression` but the fix likely also needs to be introduced to `transformJsxAttributesToProps`. I'll put up a PR for this soon.

PR #55859, opened 2023-09-25
Ensure all JSX spread properties get visited in ES2018+

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/55858


src/compiler/transformers/jsx.ts
```diff
@@ -444,7 +446,7 @@ export function transformJsx(context: TransformationContext): (x: SourceFile | B
 
     function transformJsxSpreadAttributeToProps(node: JsxSpreadAttribute) {
         if (isObjectLiteralExpression(node.expression) && !hasProto(node.expression)) {
-            return node.expression.properties;
+            return sameMap(node.expression.properties, p => Debug.checkDefined(visitNode(p, visitor, isObjectLiteralElementLike)));
```
jakebailey, line 449: This seems reasonable, but I can't help but notice that `sameMap` is never used in `**/transformers/**`, so I get a bad feeling that there's some other canonical method for doing this...

Andarist: Using this helper wasn't super intentional on my side. I just knew that there is such an util function and since this was returned as-is here before I decided to use it here (as they were returned as-is returning the same array shouldn't be the problem). 

I can change this to a regular `map` though - the outcome would be the same.

sandersn:
@jakebailey This looks fine to me too. Should it go into 5.3?

jakebailey:
Probably, though the bot apparently forgot to update the PR and do the right assignment so I'd be happier if @rbuckton took a quick look.

jakebailey:
Actually, looking at how this is used, it's not like this is being passed to anything scary; it's always going to be flattened by `transformJsxAttributesToProps` so this efficiency doesn't even matter. So, LGTM and I definitely want this in 5.3.

Issue #55832, opened 2023-09-22
Generated d.ts includes implicit any for recursive types


CraigMacomber:
### 🔎 Search Terms

recursive any noImplicitAny d.ts

### 🕗 Version & Regression Information

- This is the behavior in every version I tried (3.3 through 5.2), and I reviewed the FAQ for entries about anything related to this.

### ⏯ Playground Link

[playground link](https://www.typescriptlang.org/play?#code/MYewdgzgLgBAhjAvDAFASiQPngbgFAD0BMA4gKZhkBOcUZEAXDACZnAA2cVZMokscJuizwwAT3xEYAZQASIAK7tmMAOYVqtMk1YcuPPtHhCMibFDEAHMiABmuPEA)

### 💻 Code

```ts
const a = () => a;
```

### 🙁 Actual behavior

Any is implicitly introduced in the d.ts:
```typescript
declare const a: () => any;
```

Note that the local type checking (not using the d.ts) correctly handles this recursive type without introducing "any", and builds fine with noImplicitAny enabled.

### 🙂 Expected behavior

Either generate something without implicitly introducing `any`, or generate an error if `noImplicitAny` is enabled.
In this case, the below code generation for the d.ts would work:
```typescript
declare const a: () => typeof a;
```

fatcerberus:
I’m *more* surprised this doesn’t have a circularity error, since `a` *genuinely* refers to itself in its own initializer and the fully expanded type of `a` is `() => () => () => ...` ad infinitum (I don’t think TS generally infers `typeof` types)

CraigMacomber:
> I’m _more_ surprised this doesn’t have a circularity error, since `a` _genuinely_ refers to itself in its own initializer and the fully expanded type of `a` is therefore `() => () => () => ...` ad infinitum (I don’t think TS generally infers `typeof` types)

I just found a workaround which actually both avoids the need for inferring a `typeof` type in the d.ts, which for some reason also generates a `typeof` type in the d.ts. Well, two redundant workarounds for the price of one is still a workaround :)

```typescript
/**
 * Interface which carries the runtime and compile type data (from the generic type paramater) in a member.
 * This is also a constructor so that instances of it can be extended as classes.
 * Using classes in this way allows introducing a named type and a named value at the same time, helping keep the runtime and compile time information together and easy to refer to un a uniform way.
[...]

PR #55860, opened 2023-09-25
Try emitting an error when export would be emitted with different unwidened type

jakebailey:
This is a very very bad attempt at making the problems we've found in #55445 into check errors. I guarantee that this PR is wrong, you can see very strange things happening because observing `isDeclarationVisible` of each declaration passed to `checkVariableLikeDeclaration` seems to cause a side effect (???), and my check is poorly copied from `declarations.ts` so has some improper edge cases.

In any case, it still catches everything we mentioned in #55445 so it's somewhat interesting.


jakebailey:
@typescript-bot test top200

jakebailey:
Reading live logs, this seems to be firing on `private` things, probably because those are "preserved" in `d.ts` emit but with types removed (but this test PR doesn't model that).

jakebailey:
@typescript-bot test top100

jakebailey:
A bunch of the above declarations are still ones that shouldn't be visible, so I've clearly still done something wrong.

jakebailey:
@typescript-bot test top100

jakebailey:
@typescript-bot test top200

jakebailey:

DanielRosenwasser:
@jakebailey you mentioned examples where this doesn't just solve a divergence of declaration emit, and maybe I missed the specifics in the design meeting. Can you write up/show what this solves more explicitly?

jakebailey:
Yes, I am building an example repo now!

jakebailey:
Example: https://github.com/jakebailey/dts-emit-widened-unwidened-examples

Just to summarize @weswigham, this is arguably a limitation of declaration emit, and we could alternatively fix this by introducing syntax into `d.ts` emit that emulates ternaries, destructuring, `??`, etc, which all have this problem. This is the method we currently take via emitting initializers (and why this PR skips those as they are already handled). Fixing it in declaration emit is somewhat preferable because it doesn't shift this problem off to users.

fatcerberus:
Just guessing based on the title of this PR, but is https://github.com/microsoft/TypeScript/issues/55832 related? (it's a case where declaration emit produces a different type--in particular `any` (!)--than what's inferred locally)

jakebailey:
@typescript-bot pack this

jakebailey:
> Just guessing based on the title of this PR, but is #55832 related? (it's a case where declaration emit produces a different type--in particular `any` (!)--than what's inferred locally)

In a way, yes, it's similar to this and #55901.

PR #55864, opened 2023-09-25
Do not erase signature constraints when calculating variance

### LINKED ISSUE #55671: Instantiations of constrained generic signatures allow all assignments
### 🔎 Search Terms

variance generic signature measurement

This has been our behavior forever as far as I can tell - in fact, I've seen this pattern used once or twice to work around other bugs, even though the behavior is in no way safe.

### ⏯ Playground Link

https://www.typescriptlang.org/play?#code/C4TwDgpgBAYgdgHgCoD4oF4oIKpQgD2AjgBMBnKVACnwC4psBKDNbAbgCgPRIoBhAPZwAxgCcIRAIwYsuAkVIU4AVwC2AIwiiUNekxYNOPaIJHiiAJhk48hYuShlgogJZwA5jroNm6Vpw4AM2URYBchKEDkSQAaSgsdYEl6eGiUOOALFMQkBLjhZP4hMQkIWKhhLKKzUotmAG8OKGaoAHpWvAA3LRBHN3cAG2ghYcCoYAALCDJoAEMyMhd3OFViYAoyCYFlAZI8UVEBUS4W8elMTLY2js3t3f3D0Tj1ZWAoEgFpuAByYCaWzIyJJXdqOLY7PZaR7PV7vT5kH5-f7NAoySqcU6VNGSAKnJLYjEtVEXHEcAC+QA

### 💻 Code

```ts
type Fn<T> = <U extends T>(x: U) => U;

type Concrete1 = <U extends number>(x: U) => U;
type Concrete2 = <U extends string>(x: U) => U;

function f<T1, T2>(t1: Fn<T1>, t2: Fn<T2>, c1: Concrete1, c2: Concrete2) {
    // every single one of these assignments should error

    t1 = t2; // should error, but doesn't
    t2 = t1; // should error, but doesn't

    c1 = c2;
    c2 = c1;

    t1 = c1;
    c1 = t1;
}
```

### 🙁 Actual behavior

Missing two errors on the generic relationships.

### 🙂 Expected behavior

All of the assignments are errors, as the generic relationship should be even stricter than the non-generic one.

The issue lies in our variance analysis - we currently incorrectly measure `T` as _independent_, which is quite different from the _invariant_ it should adopt (via `U`'s invariant usages). This discrepancy arises because of some oddities in how we do generic signature relationships.

weswigham:
Fixes #55671

As noted in the linked issue, this is, unfortunately, probably in-use to mask a bunch of real issues (or as a workaround for other issues) in the wild, since it was an easy to make a type that would "accept" any assignments between instances of the same alias, even though once you did a structural comparison it only accepted near-exact matches. Even in our test suite it comes up a couple times in some examples we've pulled from the wild - I expect DT and top100 to have more.

### DIFF
--- a/tests/baselines/reference/api/typescript.d.ts
+++ b/tests/baselines/reference/api/typescript.d.ts
@@ -4145,10 +4145,10 @@ declare namespace ts {
     interface MapLike<T> {
         [index: string]: T;
     }
-    interface SortedReadonlyArray<T> extends ReadonlyArray<T> {
+    interface SortedReadonlyArray<out T> extends ReadonlyArray<T> {
         " __sortedArrayBrand": any;
     }
-    interface SortedArray<T> extends Array<T> {
+    interface SortedArray<out T> extends Array<T> {
         " __sortedArrayBrand": any;
     }
     type Path = string & {


weswigham:
@typescript-bot pack this
@typescript-bot test this
@typescript-bot run dt

tests/baselines/reference/complexRecursiveCollections.errors.txt
```diff
@@ -339,6 +347,15 @@ immutable.ts(391,22): error TS2430: Interface 'Set<T>' incorrectly extends inter
         export function Set<T>(): Seq.Set<T>;
         export function Set<T>(collection: Iterable<T>): Seq.Set<T>;
         export interface Set<T> extends Seq<never, T>, Collection.Set<T> {
+                         ~~~
+!!! error TS2430: Interface 'Set<T>' incorrectly extends interface 'Seq<never, T>'.
+!!! error TS2430:   The types of 'map(...).filter(...).concat(...).toMap().toSeq().toOrderedMap().keySeq().toMap().mapKeys(...).toSeq().toOrderedMap().toSet().union' are incompatible between these types.
```
andrewbranch, line 352: 😵

weswigham: We go hard at finding the incompatibility, don't we. 😆

Guess there's an `any` somewhere in this type definition that should be a `never` (or vice-versa, depending on intent).

weswigham:
DT reported as "a lot of problems", but it's just a lot of packages depending on immutable (which has the same extends error as in the test here - so I should probably figure out where an `any` needs to be a `never` (or vice-versa) and fix it), and dependencies on `fp-ts` (which, just like with the user suite, is an array subtype we now measure as invariant instead of covariant (which you can work around with an `out` annotation on the type parameter, as I've done in our own API). There's also a few instances of the `propTypes` circularity that's now found in the tests (usually via `styled-components`, which is where the test comes from), too. Not really sure what the workaround for that one should be yet, since the types are quite complicated. And lastly, there's some dependencies on a package called [`clownface`](https://www.npmjs.com/package/clownface), which also is detected as having an incorrect extends now (albeit much less deeply nested than `immutable`!). There's a few others, too (which are usually array or collection subtyping issues caused by more accurate variance calculation - zod and bluebird come to mind), but those are the ones that infect a lot of packages on DT by proxy.

I think this is a change we can take, once we get those big offenders fixed. Though the array subtype issue might warrant some special-casing in code, even though you can work around the issue with an annotation; arrays are forced to be covariant, whereas if we structurally inspect them (like we do their subtypes), they're measured as invariant (which is more correct! just.... harder to use). It's probably worth detecting array subtypes and short-circuiting their variance calculations, so subtypes have the same hardcoded-covariant behavior array itself does. That change can probably stand on its' own, too, since it's likely performance-positive w.r.t. variance calculation for array subtypes (unlike this change, which can only be bad for perf, since it forces us to make more comparisons with a stricter variance result).

fatcerberus:
It’s interesting to me that method bivariance alone isn’t enough to keep arrays covariant (though it does help with user-defined arraylikes, as pointed out by Ryan in the array subtype covariance PR)

PR #54121, opened 2023-05-03
Include source node inferences in string literal completions

rbuckton:
This adopts the proposed fix mentioned in https://github.com/microsoft/TypeScript/issues/49680#issuecomment-1199933521. This is a different approach than the one taken by @Andarist in #52997, as that PR attempted to pre-filter results.

Fixes #49680
Supersedes #52997

PR #55867, opened 2023-09-25
Remove `CheckMode.IsForStringLiteralArgumentCompletions`

Andarist:
It's just an internal refactor. 

I initially removed this special check mode almost entirely in https://github.com/microsoft/TypeScript/pull/53996 . It was brought up in https://github.com/microsoft/TypeScript/pull/54121 . I managed to remove it while reviewing this PR (see [here](https://github.com/microsoft/TypeScript/pull/54121#issuecomment-1704739925)) but I concluded then that this change should be proposed as a follow up - so here it is :)

### DIFF
--- a/tests/cases/fourslash/stringCompletionsFromGenericConditionalTypesUsingTemplateLiteralTypes.ts
+++ b/tests/cases/fourslash/stringCompletionsFromGenericConditionalTypesUsingTemplateLiteralTypes.ts
@@ -28,4 +28,5 @@
 //// const shallowExpression = parse("foo|/*ts*/")
 //// const nestedExpression = parse({ prop: "foo|/*ts2*/" })
 
-verify.completions({ marker: ["ts", "ts2"], exact: ["foo|foo", "foo|bar", "foo|baz"] });
+verify.completions({ marker: ["ts"], exact: ["foo", "bar", "baz", "foo|foo", "foo|bar", "foo|baz"] });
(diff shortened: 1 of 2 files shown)


tests/cases/fourslash/stringCompletionsFromGenericConditionalTypesUsingTemplateLiteralTypes.ts
```diff
@@ -28,4 +28,5 @@
 //// const shallowExpression = parse("foo|/*ts*/")
 //// const nestedExpression = parse({ prop: "foo|/*ts2*/" })
 
-verify.completions({ marker: ["ts", "ts2"], exact: ["foo|foo", "foo|bar", "foo|baz"] });
+verify.completions({ marker: ["ts"], exact: ["foo", "bar", "baz", "foo|foo", "foo|bar", "foo|baz"] });
```
Andarist, line 31: This matches Ron's initial expectations from https://github.com/microsoft/TypeScript/pull/54121 .

The other case, ideally, should provide the same completions but currently, the ones based on the contextual types and the ones based on `getCandidateSignaturesForStringLiteralCompletions` work quite differently (and even more so with changes from https://github.com/microsoft/TypeScript/pull/54121 ). I plan to close the gap between both a little bit but I'm leaving that for future PRs.

jakebailey (approved):
Seems good to me as a cleanup; I the fewer check modes the better 😄

DanielRosenwasser:
@typescript-bot pack this
@typescript-bot test top100

PR #55868, opened 2023-09-25
Remove shims.ts and related infra

jakebailey:
I'm sure there's some additional historical context I'm missing, but my understanding is that these shims were largely added to bridge the gap in the days before VS used `tsserver` (given references to COM objects and native code). This code is shipped to users, but is entirely unused. It would have been used by the global factory functions we used to set, but those were removed during the module conversion.

I personally find the LS shims annoying as adding new LS methods means adding code (mostly boilerplate) to make the shim work. But, this code is entirely unreferenced other than in tests.

As far as I can tell, the only changes to the fourslash shim tests for the past many many years have been related to other changes where we changed the fourslash API or modified things like token classification or completions (which are tested in these shims too, on top of other tests).

I feel like it's probably safe to remove these and stop shipping this code to users via our "everything is exported but hidden" API design.


sandersn (approved):
It is certainly suspicious that I've never had to interact with the shim layer or heard of anyone having to interact with it. So I think this is a good idea, but @sheetalkamat should definitely sign off first.

jakebailey:
Updated to remove the unused baselines (forgot to check that); I did a quick skim and all of the tests I'm removing appear to be tests copied and pasted from other fourslash tests but into the shims directory, just to get coverage of those operations though the shims, which this PR removes.

sheetalkamat:
I have forgotten the context of shims.ts.. possibly @RyanCavanaugh or @rbuckton remember? i have vague feeling that it has to do with one of the VS version support but i can't remember which one and if we really need it.

rbuckton:
If I recall correctly, this was related to interacting with VS when we were still being loaded in process in Visual Studio via jscript. I don't know if it's still in use, or if we still support the editions of VS that needed this.

jakebailey:
I believe that if VS used this, I broke it in https://github.com/microsoft/TypeScript/commit/b9e24eed8f1c801bbb80ab5e229e2b18029708b1, which dropped `TypeScriptServicesFactory` from the global scope. JSTS doesn't contain any references to these shims as far as I've been able to search for. Potentially @joj knows (or knows someone who knows), but I expect this to be okay to do.

joj:
@zkat or @MariaSolOs may know. We did an internal update yesterday and everything seemed ok in smoke testing, though.

jakebailey:
Well, this PR isn't merged, so I assume you wouldn't have tested out this particular PR. But I'm like 99.99% certain this is safe.

sheetalkamat (approved):

jakebailey:
For posterity:

# Package size report

## Overall package size

| - | - | - | - | - |
| Packed | 5.49 MiB | 5.48 MiB | -11.75 KiB | -0.21% |
| Unpacked | 30.51 MiB | 30.44 MiB | -70.74 KiB | -0.23% |

## Files

| - | - | - | - | - |
| `lib/tsserver.js` | 8.14 MiB | 8.11 MiB | -32.72 KiB | -0.39% |
| `lib/typescript.js` | 8.61 MiB | 8.57 MiB | -38.03 KiB | -0.43% |

"it's something"

PR #55880, opened 2023-09-26
Array subtypes inherit special array variance

weswigham:


weswigham:
@typescript-bot pack this
@typescript-bot test this
@typescript-bot run dt

fatcerberus:
Wait, `Array` is special-cased to be covariant?  That kind of shoots down the argument that method parameters are bivariant just “to ensure arrays continue to relate covariantly”…

RyanCavanaugh:
> That kind of shoots down the argument that method parameters are bivariant just “to ensure arrays continue to relate covariantly”

No, it's still exactly that -- without method bivariance, you can't make arrayish things that are covariant, which means types like `Omit<Array<T>, "sort">` would be measured as invariant
```ts
interface Arrayish1<out T> {
    indexOf(arg: T): number;
    [n: number]: T;
}

interface Arrayish2<out T> { // <-- error
    indexOf: (arg: T) => number;
    [n: number]: T;
}
```

DanielRosenwasser:
@typescript-bot test top300

sandersn:
@weswigham did the results indicate that this is worth pursuing? It is worth keeping open?

Issue #28013, opened 2018-10-20
Global variable of non-null type, never assigned, no error if used in function


fatcerberus:
The issue title is kind of awkward, I know.  Basically TypeScript allows a global variable without `undefined` in its type signature to be accessed inside of a function even if it's not provably assigned, even with `strictNullChecks` enabled.  In other words:

**TypeScript Version:**  3.2.0-dev.20181020

**Search Terms:** global undefined function error

**Code**

```ts
let pig: { eatiness: number };

function init()
{
    pig = { eatiness: 812 };
}

function feedPig()
{
    ++pig.eatiness;
}

// init();  // <-- oops! forgot to call this
feedPig();  // runtime error!
```

**Expected behavior:**
The TypeScript compiler should produce an error since `pig` is not provably assigned before first access, `strictNullChecks` is enabled and its type signature says it can't be `undefined`.

**Actual behavior:**
No compile-time error; program throws a TypeError at runtime.

**Playground Link:** 
https://www.typescriptlang.org/play/index.html#src=let%20pig%3A%20%7B%0D%0A%20%20%20%20eatiness%3A%20number%3B%0D%0A%7D%3B%0D%0A%0D%0Afunction%20init()%0D%0A%7B%0D%0A%20%20%20%20pig%20%3D%20%7B%20eatiness%3A%20812%20%7D%0D%0A%7D%0D%0A%0D%0Afunction%20feedPig()%0D%0A%7B%0D%0A%20%20%20%20%2B%2Bpig.eatiness%3B%0D%0A%7D%0D%0A%0D%0A%2F%2F%20init()%3B%20%20%2F%2F%20%3C--%20oops!%20forgot%20to%20call%20this%0D%0AfeedPig()%3B%20%20%2F%2F%20error!%0D%0A

**Related Issues:** 
#23305

mattmccutchen:
Related to #23305.

fatcerberus:
I note that if the uninitialized variable is accessed in the same scope it’s declared in (global in the example) then TS *does* produce an object-not-assigned error.  It’s only when moving the first access to a different scope that the problem arises.

ahejlsberg:
These are effects of design limitations in the control flow analyzer. See discussion in #9998.

ahejlsberg:
These are effects of design limitations in the control flow analyzer. See discussion in #9998.

fatcerberus:
[...]

PR #55887, opened 2023-09-27
error on variables that are used but never initialized

### LINKED ISSUE #23305: Uninitialized variables work around strictNullChecks (follow-up to #13884)
This is a follow-up to #13884, where @RyanCavanaugh asked me to open a new issue.

**tl;dr:** Now that there is `strictPropertyInitialization`, it seems that there should also be an analogous `strictLocalInitialization`.

**TypeScript Version:**  2.8

**Search Terms:** uninitialized local variables, strict local variables, strictLocalInitialization, strictPropertyInitialization, strictNullChecks

**Code**

```typescript
let x: number[];  // No definite assignment assertion.
// x.toString();  // Compile-time error. Yay!
function foo() {
  x.toString();  // No compile-time error, but throws an exception. :(
}
foo();
```

Slightly more interesting:

```typescript
function fooFactory() {
  let x: number;  // Inaccessible from anywhere but `foo()` below.
  return {
    foo() {
      // Guaranteed to throw an exception since `x` can't change.
      return x.toString();
    }
  };
}
```

**Expected behavior:**

If an imaginary `--strictLocalInitialization` flag is enabled, these snippets should each result in a compile-time error.

To resolve such an error, one would need to:

A) Initialize the variable in a way that the TypeScript compiler understands (e.g. immediately, or before referencing it in a closure).

-OR-

B) Use a [definite assignment assertion](https://blogs.msdn.microsoft.com/typescript/2018/01/31/announcing-typescript-2-7/#definite-assignment-assertions). It seems that this issue would come up naturally when discussing definite assignment assertions for local variables. Instead, the example provided in the blog feels contrived and wouldn't be a problem if it weren't for control flow analysis limitations.

-OR-

C) Explicitly mark the variable as potentially `undefined`, forcing callers to acknowledge that it may not be set.

**Actual behavior:**

No compile-time errors, but obvious run-time errors.

**Playground Link:**

[Link](https://www.typescriptlang.org/play/#src=function%20fooFactory()%20%7B%0D%0A%20%20let%20x%3A%20number%3B%0D%0A%20%20return%20%7B%0D%0A%20%20%20%20foo()%20%7B%0D%0A%20%20%20%20%20%20%2F%2F%20%60x%60%20isn't%20initialized!%0D%0A%20%20%20%20%20%20%2F%2F%20Run-time%20error!%0D%0A%20%20%20%20%20%20return%20x.toString()%3B%0D%0A%20%20%20%20%7D%0D%0A%20%20%7D%3B%0D%0A%7D%0D%0A) (enable `strictNullChecks`)

**Related Issues:**

#13884, #9998

---

Understandably, this may not be as simple as `strictPropertyInitialization`, but I feel it needs a discussion none the less. The `fooFactory()` example above is analogous to this `Foo` class, which correctly emits a compile-time error under `strictPropertyInitialization`:

```typescript
class Foo {
  private x: number; // Compile-time error!
  foo() {
    // `x` isn't initialized! But it's OK, compile-time error above.
    return this.x.toString();
  }
}
```

Zzzen:
Fixes #23305


src/compiler/checker.ts
```diff
@@ -2147,7 +2147,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     var deferredGlobalClassAccessorDecoratorResultType: GenericType | undefined;
     var deferredGlobalClassFieldDecoratorContextType: GenericType | undefined;
 
-    var allPotentiallyUnusedIdentifiers = new Map<Path, PotentiallyUnusedIdentifier[]>(); // key is file name
+    var allPotentiallyUnusedOrUninitializedIdentifiers = new Map<Path, PotentiallyUnusedIdentifier[]>(); // key is file name
```
Zzzen, line 2150: Will `allLocalIdentifiers` be a better variable name?

src/compiler/types.ts
```diff
@@ -5765,6 +5765,7 @@ export interface Symbol {
     /** @internal */ exportSymbol?: Symbol; // Exported symbol associated with this symbol
     /** @internal */ constEnumOnlyModule: boolean | undefined; // True if module contains only const enums or other modules with only const enums
     /** @internal */ isReferenced?: SymbolFlags; // True if the symbol is referenced elsewhere. Keeps track of the meaning of a reference in case a symbol is both a type parameter and parameter.
+    /** @internal */ isInitialized?: boolean; // True if the symbol is ever initialized.
```
DanielRosenwasser, line 5768: If this is set in the checker, it probably shouldn't be directly set on the `Symbol` (and should be set on `SymbolLinks` instead). Maybe someone else has some thoughts.

DanielRosenwasser:
@typescript-bot pack this
@typescript-bot test this
@typescript-bot test top300

src/compiler/types.ts
DanielRosenwasser: Alternatively, can you just use `isAssigned` (defined below) which is set in the binder? You'll need to change some existing code to proactively check for whether the symbol also corresponds to a parameter.

jakebailey:
Hm, perf result indicates that adding this prop has moved Symbol into a new size class, given the memory increase.

src/compiler/types.ts
jakebailey: Based on perf, this is making the Symbol allocated in the next size class up, which is undesirable. So probably on links.

DanielRosenwasser:
Can't we just reuse `isAssigned`?

src/compiler/types.ts
DanielRosenwasser: Can't we just reuse `isAssigned`?

jakebailey: Probably!

DanielRosenwasser:
I spent a bit looking at the breaks and decided to take a break.

There was one example of a TODO which is questionable - I didn't quite understand that.

https://github.com/immich-app/immich/blob/0a22e64799435a6ddf668733fab82da542e495f9/server/src/immich/api-v1/asset/asset.service.spec.ts#L89

For the most part, [in the top 300 TypeScript projects on GitHub](https://github.com/microsoft/TypeScript/pull/55887#issuecomment-1738315709), the only real false negatives are variables that get assigned in a `for`/`in` or `for`/`of` loop. That seems like a bug and I'd like to rerun the tests once that's fixed.

https://github.com/snabbdom/snabbdom/blob/420fa78abe98440d24e2c5af2f683e040409e0a6/src/modules/class.ts#L8-L18

https://github.com/snabbdom/snabbdom/blob/420fa78abe98440d24e2c5af2f683e040409e0a6/src/modules/eventlisteners.ts#L56-L67

https://github.com/snabbdom/snabbdom/blob/420fa78abe98440d24e2c5af2f683e040409e0a6/src/modules/style.ts#L29-L40

https://github.com/snabbdom/snabbdom/blob/420fa78abe98440d24e2c5af2f683e040409e0a6/src/modules/style.ts#L70-L74

https://github.com/microsoft/playwright/blob/c8143748e6af90df485dd46410915033b1001b73/utils/generate_types/test/test.ts#L164-L173

https://github.com/microsoft/vscode/blob/c965b118a67c0f22895d2af899736d0e0aa82ffd/extensions/ipynb/src/helper.ts#L44-L58

https://github.com/vuejs/core/blob/b8fc18c0b23be9a77b05dc41ed452a87a0becf82/packages/compiler-core/src/parse.ts#L122-L123

https://github.com/vuejs/core/blob/b8fc18c0b23be9a77b05dc41ed452a87a0becf82/packages/template-explorer/src/index.ts#L122-L123

https://github.com/vuetifyjs/vuetify/blob/3832f737c030d2f8a12b004473879f28689dc75e/packages/vuetify/src/components/VGrid/VCol.ts#L120-L121

https://github.com/vuetifyjs/vuetify/blob/3832f737c030d2f8a12b004473879f28689dc75e/packages/vuetify/src/components/VGrid/VRow.ts#L126-L127

fatcerberus:
Just to be clear: "False negative" in this context means "this should be an error but isn't", correct?

I'm a little confused by the `for..in` examples - why would you write `let foo; for (foo in whatever) ...` instead of just `for (let foo in whatever) ...`, unless you're intending to use the variable outside the loop.

src/compiler/types.ts
Zzzen: Updated. It's totally fine to reuse `isAssigned`.

tests/baselines/reference/findAllRefsObjectBindingElementPropertyName06.baseline.jsonc
```diff
@@ -13,7 +13,7 @@
 // for (<|var { [|property1|]: p1 } of elems|>) {
 // }
 // var p2;
-// for (<|{ [|{| isWriteAccess: true |}property1|] : p2 } of elems|>) {
+// for (<|{ [|property1|] : p2 } of elems|>) {
```
Zzzen, line 16: Could anyone please explain the meaning of "isWriteAccess" in this context? Additionally, I'm curious to know if this qualifies as a regression.

Zzzen:
>For the most part, https://github.com/microsoft/TypeScript/pull/55887#issuecomment-1738315709, the only real false negatives are variables that get assigned in a for/in or for/of loop. That seems like a bug and I'd like to rerun the tests once that's fixed.

It has been fixed. I'm also curious why they were written in this manner.

tests/baselines/reference/findAllRefsObjectBindingElementPropertyName06.baseline.jsonc
DanielRosenwasser: in find-all-references, we can indicate that certain references are writes so that editors can filter reads and writes out, or display them differently. Technically this is correct as `property1` s not being written to - `p2` is.

DanielRosenwasser: Though I wonder if `p2` has `isWriteAccess` on it.

DanielRosenwasser:
> Just to be clear: "False negative" in this context means "this should be an error but isn't", correct?

Sorry, I meant false positives. I gotta stop trying to sound smarter than I am.

DanielRosenwasser:
I'm a little bit concerned that we aren't guarding previous accesses of `isAssigned` in some way - but it looks like the binder didn't need to be changed *at all*. Were accesses of `isAssigned` already guarded by something like `symbol.flags & SymbolFlags.Parameter`?

@typescript-bot pack this
@typescript-bot test top300

Zzzen:
>I'm a little bit concerned that we aren't guarding previous accesses of isAssigned in some way - but it looks like the binder didn't need to be changed at all. Were accesses of isAssigned already guarded by something like symbol.flags & SymbolFlags.Parameter?

After conducting some research, it appears that the `isAssigned` flag is not referenced in the binder. Moreover, all the reads have been appropriately guarded, as demonstrated in the examples below:

```ts
parent.kind === SyntaxKind.Parameter && isSymbolAssigned(symbol)

isParameter && !isSymbolAssigned(localOrExportSymbol)

isParameterOrCatchClauseVariable(symbol) && !isSymbolAssigned(symbol)
```

DanielRosenwasser:
@typescript-bot test top800

gabritto:
@typescript-bot pack this

gabritto (changes requested):
There seems to be a false error in the extended test results (see [eclipse/che errors here](https://github.com/microsoft/TypeScript/pull/55887#issuecomment-2040892107)) that looks like this:

```ts
function ff() {
  let _;
  let rest: {}; // error but should not

  [_, ...rest] = bar();
}

declare function bar(): [number, ...string[]];
```

Please add a test and fix.

src/compiler/types.ts
```diff
@@ -5788,6 +5788,7 @@ export interface Symbol {
     /** @internal */ isReferenced?: SymbolFlags; // True if the symbol is referenced elsewhere. Keeps track of the meaning of a reference in case a symbol is both a type parameter and parameter.
     /** @internal */ lastAssignmentPos?: number; // Source position of last node that assigns value to symbol
     /** @internal */ isReplaceableByMethod?: boolean; // Can this Javascript class property be replaced by a method symbol?
+    /** @internal */ isAssigned?: boolean;  // True if the symbol is a parameter with assignments or a variable that is assigned elsewhere
```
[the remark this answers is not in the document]
Zzzen: Good news: `lastAssignmentPos` can be reused for this PR and it also fixes https://github.com/microsoft/TypeScript/pull/55887#pullrequestreview-2178523134
Bad news: We can't check whether it is write-only assignment.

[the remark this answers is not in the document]
Zzzen: added a flag `isDefinitelyAssigned`

DanielRosenwasser:
@typescript-bot test it

src/compiler/types.ts
```diff
@@ -5936,6 +5936,7 @@ export interface Symbol {
     /** @internal */ constEnumOnlyModule: boolean | undefined; // True if module contains only const enums or other modules with only const enums
     /** @internal */ isReferenced?: SymbolFlags; // True if the symbol is referenced elsewhere. Keeps track of the meaning of a reference in case a symbol is both a type parameter and parameter.
     /** @internal */ lastAssignmentPos?: number; // Source position of last node that assigns value to symbol
+    /** @internal */ isDefinitelyAssigned?: boolean; // True if the symbol is assigned definitely
```
DanielRosenwasser, line 5939: I think there's a different use of terminology here. In most interpretations, definitely-assigned implies that a variable is *always* assigned to by a specific point.

Here, it feels like the meaning is "some assignment definitely exists".

src/compiler/checker.ts
```diff
@@ -29501,6 +29501,11 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         return !isPastLastAssignment(symbol, /*location*/ undefined);
     }
 
+    // Check if a parameter or catch variable is assigned definitely
+    function isSymbolAssignedDefinitely(symbol: Symbol) {
+        return symbol.isDefinitelyAssigned || (isSymbolAssigned(symbol) && symbol.isDefinitelyAssigned);
```
DanielRosenwasser, line 29506: Isn't the right thing here

```suggestion
        return symbol.isDefinitelyAssigned ?? (isSymbolAssigned(symbol) && symbol.isDefinitelyAssigned);
```

Otherwise, you will recalculate `isSymbolAssigned` every time you want to run `isSymbolDefinitelyAssigned` in the case where a variable is never assigned to.

src/compiler/checker.ts
```diff
@@ -43201,10 +43220,15 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 case SyntaxKind.MethodDeclaration:
                 case SyntaxKind.GetAccessor:
                 case SyntaxKind.SetAccessor:
-                    if (node.body) { // Don't report unused parameters in overloads
-                        checkUnusedLocalsAndParameters(node, addDiagnostic);
+                    if (checkUnused) {
+                        if (node.body) { // Don't report unused parameters in overloads
```
DanielRosenwasser, line 43224: While we're here:

```suggestion
                        // Only report unused parameters on the implementation, not overloads.
                        if (node.body) {
```

src/compiler/types.ts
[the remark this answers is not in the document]
DanielRosenwasser: What do you mean by write-only assignment? Like `a = b` as opposed to `a ??= b`?

Zzzen:
>What do you mean by write-only assignment? Like a = b as opposed to a ??= b?

Yes, definite assignment would be a more concise description.

src/compiler/checker.ts
Zzzen: fixed

src/compiler/types.ts
Zzzen: 🤔  what about `hasDefiniteAssignment`

src/compiler/types.ts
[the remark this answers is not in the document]
Zzzen: Yes, definite assignment would be a more concise description.

gabritto:
@typescript-bot test it

gabritto:
@typescript-bot pack this

src/compiler/checker.ts
```diff
@@ -29504,6 +29504,11 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         return !isPastLastAssignment(symbol, /*location*/ undefined);
     }
 
+    // Check if a parameter or catch variable is assigned definitely
```
gabritto, line 29507: This comment (and the one for `isSymbolAssigned`) is not accurate, I think, because we also check if mutable local variables are assigned definitely.

src/compiler/checker.ts
```diff
@@ -43521,6 +43551,28 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         });
     }
 
+    function checkUninitializedLocals(nodeWithLocals: HasLocals) {
+        nodeWithLocals.locals!.forEach(local => {
+            if (!(local.flags & SymbolFlags.Variable) || !local.isReferenced || isSymbolAssignedDefinitely(local)) {
```
gabritto, line 43556: I think you should also check if the variable is constant, i.e. use `isConstantVariable`, otherwise we can get two errors for an uninitialized const variable, and `isSymbolAssigned` and `isSymbolAssignedDefinitely` are not intended to be used with constant variables.

Test case:
```ts
function f() {
    const x: number; // Two similar errors here
    function bar() {
        let y = x;
    }
}
```

gabritto (changes requested):

gabritto:
I noticed the new error is not present for global variables. So this PR won't fix #28013. I believe the reason is that the variable could be initialized in a different file.

Zzzen:
Yes, global variables are ignored intentionally.

gabritto:
@typescript-bot pack this

gabritto (changes requested):
We discussed this at design meeting today, and we like this change, but there are two things we feel need to be addressed before we can merge this:

1. Move the error to the usages of the never-initialized variable, instead of its declaration.
Having the error on the usages instead of declaration would make the new error more consistent with the existing one (error ts2454).
In terms of implementation, this can be achieved by moving the checks inside `checkUninitializedLocals` to `checkIdentifier`, maybe right after we do our checks for the existing error: https://github.dev/microsoft/TypeScript/blob/451388cfd5b0d086c5f69b9ff5675b650feef623/src/compiler/checker.ts#L30431

2. Don't introduce a new symbol property.
For performance reasons (see [this](https://github.com/microsoft/TypeScript/pull/55887#issuecomment-1738211339)), it would be best to reuse `lastAssignmentPos` somehow to indicate that a variable was "definitely" assigned. One idea would be to change the sign of the numbers to indicate that, i.e. making `lastAssignmentPos` a negative number to indicate it (and `0` if we already computed `lastAssignmentPos` but the symbol is never assigned).

Also left a minor comment on `checkUninitializedLocals`.

src/compiler/checker.ts
```diff
@@ -43521,6 +43551,28 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
     }
 
+    function checkUninitializedLocals(nodeWithLocals: HasLocals) {
+        nodeWithLocals.locals!.forEach(local => {
+            if (!(local.flags & SymbolFlags.Variable) || !local.isReferenced || isSymbolAssignedDefinitely(local)) {
+                return;
+            }
+            const declaration = local.valueDeclaration;
+            if (
+                !declaration || !isVariableDeclaration(declaration) || !(getCombinedNodeFlagsCached(declaration) & NodeFlags.Let) ||
```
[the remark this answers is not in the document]
Zzzen: `var` variables are not considered as mutable local variables. Do you know why?

[playground](https://www.typescriptlang.org/play/?useDefineForClassFields=true&target=9&module=1&noImplicitUseStrict=true&ts=5.6.0-dev.20240727#code/GYVwdgxgLglg9mABMAFASkQbwFCL4gGwFMpEBDALkQGcoAnGMAc0QB9FwATI4RozgNy58ANzJ1EAIyq0GzNhzDdeYfkOF4yiALyIA5AEY9GqTv0AmYybokQdJOh0A+LCfyIICWh5n1GLXTIhd3cAelDEABUATwAHIn1ZfwUuHj5OPUQYakQwOFIyamoYJjAySWJEKDgquIS9JOY9ADo3fHDEAHUAC2jEAH5BtrxPMG9OXzkAqWD8AF9sOaA)

```ts
function f() {
    let a: string | undefined;
    var b: string | undefined;

    a = '1'
    b = '2'

    return () => {
        const c: string = a;
        // Type 'string | undefined' is not assignable to type 'string'.
        // Why ???
        const d: string = b;
    }
}```

**Comment by Zzzen** [comment]
Changes have been implemented as discussed. Could you run the performance test suite to ensure everything is running smoothly?

**Comment by gabritto** [comment]
@typescript-bot test it

**Comment by gabritto** [comment]
@typescript-bot pack this

**Review by gabritto — CHANGES_REQUESTED** [review]
Left some comments for minor fix/refactor. The change to reuse `lastAssignmentPos` seems to have worked and performance numbers are better now 😊

File: src/compiler/checker.ts [hunk 11]
```diff
@@ -29557,12 +29565,19 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
-                        const referencingFunction = findAncestor(node, isFunctionOrSourceFile);
-                        const declaringFunction = findAncestor(symbol.valueDeclaration, isFunctionOrSourceFile);
-                        symbol.lastAssignmentPos = referencingFunction === declaringFunction ? extendAssignmentPosition(node, symbol.valueDeclaration!) : Number.MAX_VALUE;
+                    const isDefiniteAssignment = assigmentTarget === AssignmentKind.Definite || (symbol.lastAssignmentPos !== undefined && symbol.lastAssignmentPos < 0);
+                    if (isParameterOrMutableLocalVariable(symbol)) {
+                        if (symbol.lastAssignmentPos === undefined || Math.abs(symbol.lastAssignmentPos) !== Number.MAX_VALUE) {
+                            const referencingFunction = findAncestor(node, isFunctionOrSourceFile);
+                            const declaringFunction = findAncestor(symbol.valueDeclaration, isFunctionOrSourceFile);
+                            symbol.lastAssignmentPos = referencingFunction === declaringFunction ? extendAssignmentPosition(node, symbol.valueDeclaration!) : Number.MAX_VALUE;
```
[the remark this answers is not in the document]
Zzzen: We shouldn't needs a sign multiplier here since we are already checking and correcting the symbol of `symbol.lastAssignmentPos` in [the next 'if' statement](https://github.com/microsoft/TypeScript/pull/55887/files#diff-d9ab6589e714c71e657f601cf30ff51dfc607fc98419bf72e04f6b0fa92cc4b8R29578-R29580). This logic cannot be merged into the current 'if' statement because we might encounter a non-definite assignment first and then a definite assignment later.

src/compiler/checker.ts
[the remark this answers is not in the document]
Zzzen: Are Monaco and VSCode the same test suite? I added support for validating 'Var' in this merge request, but it seems like the performance has not been affected.

src/compiler/checker.ts
```diff
@@ -30405,7 +30422,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         // We only look for uninitialized variables in strict null checking mode, and only when we can analyze
         // the entire control flow graph from the variable's declaration (i.e. when the flow container and
         // declaration container are the same).
-        const assumeInitialized = isParameter || isAlias || isOuterVariable || isSpreadDestructuringAssignmentTarget || isModuleExports || isSameScopedBindingElement(node, declaration) ||
+        const assumeInitialized = isParameter || isAlias ||
+            (isOuterVariable && !((immediateDeclaration && isVariableDeclaration(immediateDeclaration) && !immediateDeclaration.initializer && !immediateDeclaration.exclamationToken && isMutableLocalVariableDeclaration(symbol, immediateDeclaration)) && !isSymbolAssignedDefinitely(symbol))) ||
```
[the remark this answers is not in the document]
Zzzen: Haha, that was my initial thought as well, but I'm not keen on the idea of having to perform additional evaluations even when isParameter or isAlias is true, although this may not impact the overall performance.

https://github.com/microsoft/TypeScript/pull/55887/commits/bf4bb51b23d29470930781bf682d52b30e35ea9a

gabritto:
@typescript-bot pack this

src/compiler/checker.ts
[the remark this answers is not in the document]
gabritto: I don't think it would matter that much, the checks being performed should be very cheap, but it's up to you.

src/compiler/checker.ts
```diff
@@ -29622,13 +29638,13 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
-            isVariableDeclaration(declaration) && (isCatchClause(declaration.parent) || isMutableLocalVariableDeclaration(declaration))
+            isVariableDeclaration(declaration) && (isCatchClause(declaration.parent) || isMutableLocalVariableDeclaration(symbol, declaration))
         );
     }
 
-    function isMutableLocalVariableDeclaration(declaration: VariableDeclaration) {
+    function isMutableLocalVariableDeclaration(symbol: Symbol, declaration: VariableDeclaration) {
         // Return true if symbol is a non-exported and non-global `let` variable
-        return !!(declaration.parent.flags & NodeFlags.Let) && !(
+        return !!(symbol.flags & SymbolFlags.FunctionScopedVariable || (declaration.parent.flags & NodeFlags.Let)) && !(
```
gabritto, line 29647: I got an answer and the reason we exclude `var` variables from being considered mutable local variables is because they have different scoping rules we would have to consider when doing the analysis we do for parameters or mutable local variables. For instance, I think `extendAssignmentPos` would have to account for an assignment that occurs before the variable declaration, since `var`s are hoisted.

So I think we can just revert this change for now and not do the new analysis for `var` variables, otherwise we'd have to possibly update every other place that calls `isParameterOrMutableLocalVariable` to account for `var`. You can leave the existing `var` tests in place though.

gabritto (changes requested):

src/compiler/checker.ts
Zzzen: reverted

src/compiler/checker.ts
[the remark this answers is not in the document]
Zzzen: It has been relocated to isNeverInitialized, resulting in cleaner code now.

src/compiler/checker.ts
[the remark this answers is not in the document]
Zzzen: Indeed, `hasDefiniteAssignment` is a preferable option.

tests/baselines/reference/unusedLocalsInMethod4.errors.txt
```diff
@@ -0,0 +1,191 @@
...
+        let x2: number[] | null; // should error
+        let x3: number[] | undefined; // should not error
+        let x4: number[] | undefined | null; // should not error
+        let x5!: number[]; // should not error
+        let x6: any; // should not error
+        let x7: unknown; // should not error
+        let x8: T; // should error
+        let x9: NonNull; // should error
+        var x10: NonNull; // should error
+        var x11: NonNull; // should not error
```
gabritto, line 30: `x11` should probably be a `let` variable if you want to test the case where the variable is definitely assigned inside another function.

tests/baselines/reference/unusedLocalsInMethod4.errors.txt
```diff
@@ -0,0 +1,191 @@
...
+        let x1: number[]; // should error
+        let x2: number[] | null; // should error
+        let x3: number[] | undefined; // should not error
+        let x4: number[] | undefined | null; // should not error
+        let x5!: number[]; // should not error
+        let x6: any; // should not error
+        let x7: unknown; // should not error
+        let x8: T; // should error
+        let x9: NonNull; // should error
+        var x10: NonNull; // should error
```
gabritto, line 29: This `// should error` is out of date since we're not erroring on `var`s anymore.

gabritto:
@typescript-bot test it

DanielRosenwasser:
@typescript-bot pack this

src/compiler/checker.ts
```diff
@@ -29557,12 +29565,19 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
-                        symbol.lastAssignmentPos = referencingFunction === declaringFunction ? extendAssignmentPosition(node, symbol.valueDeclaration!) : Number.MAX_VALUE;
+                    const hasDefiniteAssignment = assigmentTarget === AssignmentKind.Definite || (symbol.lastAssignmentPos !== undefined && symbol.lastAssignmentPos < 0);
+                    if (isParameterOrMutableLocalVariable(symbol)) {
+                        if (symbol.lastAssignmentPos === undefined || Math.abs(symbol.lastAssignmentPos) !== Number.MAX_VALUE) {
+                            const referencingFunction = findAncestor(node, isFunctionOrSourceFile);
+                            const declaringFunction = findAncestor(symbol.valueDeclaration, isFunctionOrSourceFile);
+                            symbol.lastAssignmentPos = referencingFunction === declaringFunction ? extendAssignmentPosition(node, symbol.valueDeclaration!) : Number.MAX_VALUE;
+                        }
+                        if (hasDefiniteAssignment && symbol.lastAssignmentPos > 0) {
+                            symbol.lastAssignmentPos = -symbol.lastAssignmentPos;
```
DanielRosenwasser, line 29579:
```suggestion
                            symbol.lastAssignmentPos *= -1;
```

tests/baselines/reference/narrowingPastLastAssignment.errors.txt
```diff
@@ -160,4 +161,14 @@ narrowingPastLastAssignment.ts(90,20): error TS7005: Variable 'x' implicitly has
         }
         values.forEach(v => foo.push(v));
     }
-    
\ No newline at end of file
+    
+    function f13() {
+        var foo: string | undefined;
```
DanielRosenwasser, line 166: I guess the subtlety here is that this is `var` and I missed that. So if you add a comment to the original test like

```
// Test for captured 'var' declaration (as opposed to parameters, let, const).
```

that might be helpful.

src/compiler/transformers/es2015.ts
```diff
@@ -523,7 +523,7 @@ export function transformES2015(context: TransformationContext): (x: SourceFile
      * They are persisted between each SourceFile transformation and should not
      * be reset.
      */
-    let enabledSubstitutions: ES2015SubstitutionFlags;
+    let enabledSubstitutions!: ES2015SubstitutionFlags;
```
DanielRosenwasser, line 526: It feels like all of these should just be initialized with `0` or some equivalent.

DanielRosenwasser (dismissed):
Apart from my nits I think this is ready. Excited to get this into 5.7!

src/compiler/transformers/es2015.ts
DanielRosenwasser: CC @rbuckton

Zzzen: Added a `None` flag

tests/baselines/reference/narrowingPastLastAssignment.errors.txt
Zzzen: Comment added

gabritto (approved):

Issue #55884, opened 2023-09-26
No error on `super.` for inferred class properties in JavaScript


DanielRosenwasser:
In https://github.com/microsoft/TypeScript/pull/54056 we added checks for class fields; however, we don't issue the same error in JavaScript for properties inferred from assignments.

```js
// @ts-check

class YaddaBase {
    constructor() {
        this.roots = "hi";
    }
}

class DerivedYadda extends YaddaBase {
    get rootTests() {
        return super.roots; // no error?
    }
}

console.log(new DerivedYadda().rootTests); // prints undefined!
```

_Originally posted by @DanielRosenwasser in https://github.com/microsoft/TypeScript/issues/55883#issuecomment-1736293678_

PR #54056, opened 2023-04-28
fix: class field is not accessible via super

Jack-Works:
fixes #54054
fixes #35314
fixes #54900

PR #55892, opened 2023-09-27
Error on class fields accesses through `super` in JS files

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/55884
follow up to https://github.com/microsoft/TypeScript/pull/54056

### DIFF
--- a/src/compiler/utilitiesPublic.ts
+++ b/src/compiler/utilitiesPublic.ts
@@ -102,6 +102,7 @@ import {
     isArrowFunction,
     isAssignmentExpression,
     isBinaryExpression,
+    isBindableStaticAccessExpression,
     isBindableStaticElementAccessExpression,
     isBindingElement,
     isBlock,
@@ -154,6 +155,7 @@ import {
     isPropertyAccessExpression,
     isPropertyAssignment,
     isPropertyDeclaration,
+    isPrototypeAccess,
     isRootedDiskPath,
     isSourceFile,
     isStringLiteral,
@@ -1708,7 +1710,7 @@ export function isAutoAccessorPropertyDeclaration(node: Node): node is AutoAcces
 /** @internal */
 export function isClassFieldAndNotAutoAccessor(node: Declaration): boolean {
     if (isInJSFile(node) && isExpandoPropertyDeclaration(node)) {
-        return true;
+        return !isBindableStaticAccessExpression(node) || !isPrototypeAccess(node.expression);
     }
     return node.parent && isClassLike(node.parent) && isPropertyDeclaration(node) && !hasAccessorModifier(node);
 }
--- a/src/compiler/utilitiesPublic.ts
+++ b/src/compiler/utilitiesPublic.ts
@@ -111,6 +111,7 @@ import {
     isClassStaticBlockDeclaration,
     isDecorator,
     isElementAccessExpression,
+    isExpandoPropertyDeclaration,
     isExportAssignment,
     isExportDeclaration,
     isExportSpecifier,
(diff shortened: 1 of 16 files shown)


tests/baselines/reference/classFieldSuperNotAccessibleJs.types
```diff
@@ -0,0 +1,85 @@
...
+>foo : number | undefined
+>10 : 10
+    }
+}
+
+class DerivedYadda extends YaddaBase {
+>DerivedYadda : DerivedYadda
+>YaddaBase : YaddaBase
+
+    get rootTests() {
```
Andarist, line 52: That is a separate issue - but I wonder... shouldn't such an invalid super expression result in an `errorType`? 🤔

tests/baselines/reference/classFieldSuperNotAccessibleJs.types
```diff
@@ -0,0 +1,85 @@
...
+>rootTests : string
+
+        return super.roots;
+>super.roots : string
```
Andarist, line 56: That is a separate issue - but I wonder... shouldn't such an invalid super expression result in an `errorType`? 🤔

tests/baselines/reference/classFieldSuperNotAccessibleJs.errors.txt
```diff
@@ -0,0 +1,47 @@
...
+                         ~~~
+!!! error TS2855: Class field 'foo' defined by the parent class is not accessible in the child class via super.
+        }
+        get justPropTests() {
+            return super.justProp;
+                         ~~~~~~~~
+!!! error TS2855: Class field 'justProp' defined by the parent class is not accessible in the child class via super.
+        }
+        get literalElementAccessTests() {
+            return super.literalElementAccess;
```
DanielRosenwasser, line 42: Can you add a test for `super["literalElementAccess"]` as well?

Also, can you just add a test for `super.b`? I bet we have no test for `super.` on an `accessor` declaration at all.

tests/baselines/reference/classFieldSuperNotAccessibleJs.types
DanielRosenwasser: No, I think it's "clear" what the user was doing, so `errorType` becomes unnecessarily infectious.

DanielRosenwasser:
Can you add a test for this case in JavaScript as well and confirm it works?

```js
// @checkJs: true

// @filename: foo.js
export class C {
  static blah1 = 123;
}
C.blah2 = 456;

export class D extends C {
  static {
    console.log(super.blah1);
    console.log(super.blah2);
  }
}
```

along with this:

```js
// @checkJs: true

// @filename: foo.js
class C {
  constructor() {
    this.foo = () => {
      console.log("called arrow")
    }
  }
  foo() {
    console.log("called method")
  }
}

class D extends C {
  foo() {
    console.log("SUPER:");
    super.foo();
    console.log("THIS:");
    this.foo();
  }
}

const obj = new D();
obj.foo();
D.prototype.foo.call(obj);
```

tests/baselines/reference/classFieldSuperNotAccessibleJs.errors.txt
Andarist: > Can you add a test for super["literalElementAccess"] as well?

Good catch. It doesn't work correctly today ([TS playground](https://www.typescriptlang.org/play?ts=5.3.0-dev.20230927#code/MYGwhgzhAEAq0G8BQ1oDMCWBTEATaAvNABQCUhAfIgL4DcS1SokMsATNFgB4AuWAdrlaIU6MiNSoIAVwAOWAE4A6TDlxl6k6DPkKA2gCJVeAwF0NoxoyA)) even in TS files. For that reason - I'd say that this is a separate issue. I can look into it as well but I'd prefer to explore it separately.

> Also, can you just add a test for super.b? I bet we have no test for super. on an accessor declaration at all.

Sure thing, just added it.

Andarist:
> Can you add a test for this case in JavaScript as well and confirm it works?

Added both - I had to fix one of them 😉

jakebailey:
@typescript-bot test top100

jakebailey (approved):

src/compiler/utilitiesPublic.ts
```diff
@@ -1705,7 +1709,10 @@ export function isAutoAccessorPropertyDeclaration(node: Node): node is AutoAcces
 }
 
 /** @internal */
-export function isClassFieldAndNotAutoAccessor(node: Node): boolean {
+export function isClassFieldAndNotAutoAccessor(node: Declaration): boolean {
```
DanielRosenwasser, line 1712:
```suggestion
export function isClassInstanceProperty(node: Declaration): boolean {
```

DanielRosenwasser:
Let's get this in for 5.3 beta and do the rename after.

PR #51455, opened 2022-11-08
Convert debug namespace into a module, direct import

jakebailey:
For #51441.

This is best viewed without whitespace: https://github.com/microsoft/TypeScript/pull/51455/files?w=1

All of the real changes are in `src/compiler/debug.ts`.

## Overall package size

|  | Before | After | Diff | Diff (percent) |
| - | - | - | - | - |
| Packed | 4.00 MiB | 4.00 MiB | -3.92 KiB | -0.10% |
| Unpacked | 21.47 MiB | 21.44 MiB | -32.89 KiB | -0.15% |

## Files

|  | Before | After | Diff | Diff (percent) |
| - | - | - | - | - |
| `lib/tsc.js` | 5.82 MiB | 5.80 MiB | -21.63 KiB | -0.36% |
| `lib/tsserver.js` | 25.94 KiB | 25.96 KiB | +26.00 B | +0.10% |
| `lib/typescript.js` | 8.55 MiB | 8.54 MiB | -11.29 KiB | -0.13% |

PR #55894, opened 2023-09-27
Convert Diagnostics to a module, direct import

jakebailey:
This is the same as #51455 but for the diagnostics module.

~I was hoping this would tree shake out a bunch in `tsc.js`, but `esbuild` appears to not notice that a load of diagnostics (370 or so) are unused in `tsc` and can be removed. I need to report that upstream (could be https://github.com/evanw/esbuild/issues/3256?), though maybe if I just @evanw he might be interested~ 😅 

See discussion about the tree shaking; I've committed a workaround which has helped.

So far, the change is:

## Overall package size

| - | - | - | - | - |
| Packed | 3.77 MiB | 3.79 MiB | +21.74 KiB | +0.56% |
| Unpacked | 20.37 MiB | 20.53 MiB | +162.50 KiB | +0.78% |

## Files

| - | - | - | - | - |
| `lib/tsc.js` | 5.57 MiB | 5.49 MiB | -86.12 KiB | -1.51% |
| `lib/typescript.js` | 8.17 MiB | 8.41 MiB | +248.62 KiB | +2.97% |


jakebailey:

evanw:
I took a look and I know why this happens. Here is one of the diagnostics as an example:

```ts
export const Show_diagnostic_information: DiagnosticMessage =
  /* @__PURE__ */ diag(
    6149,
    DiagnosticCategory.Message,
    "Show_diagnostic_information_6149",
    "Show diagnostic information.");
```

The problem is that while the call to `diag` is marked as pure, that only neutralizes the side effects from the call itself. It does not affect any side effects from any of the arguments, which still take place. In particular, esbuild does not know whether the dynamic property access `DiagnosticCategory.Message` has side effects or not at the stage in the pipeline where esbuild is making these side effect decisions. This is because `DiagnosticCategory` is defined in a separate file and esbuild processes files in parallel.

One hack to fix this could be to move (or copy) the definition of `DiagnosticCategory` into that file. Another hack to fix this could be to wrap each diagnostic in an IIFE so that the arguments are included in the scope of `/* @__PURE__ */`. Obviously it would be best for esbuild to be able to determine that this is side-effect free itself without any code changes, but that is not necessarily straightforward on esbuild's end.

jakebailey:
`DiagnosticCategory` is an enum, though; is it possible that the information of "this is an enum we're giong to inline" could be used to determine that the operation is side effect free? IIRC enum info is collected at parse time, so I think it's known at that point...

jakebailey:
Redeclaring it locally does work, though; I'll commit that workaround for now. Thanks for the suggestion!

evanw:
The problem is a combination of:

* Side effect decisions are currently made independently per file using just the syntactic information available to the parser for that file
* The enum definition for `DiagnosticCategory` is in another file so the parser of `diagnosticInformationMap.generated.ts` doesn't have any information about `DiagnosticCategory` as it determines side effects for `diagnosticInformationMap.generated.ts`

This is an esbuild problem of course and not a problem inherent to all bundlers. You could imagine another implementation of esbuild that defers all optimization decisions until link time to handle things like this. That's not how esbuild currently works, however.

jakebailey:
Ah, gotcha. Thanks for the explainer.

jakebailey:

evanw:
I have a bit of a weird bug report for you: #55915. I found it while trying to help you out here. I tried making an incompatible definition of `DiagnosticCategory` locally to sanity-check that the TypeScript compiler would flag it as an error but I was surprised to find that it didn't. It seems to be based on whether the two incompatible enum definitions are named the same thing or not in their respective files. Seems like a bug to me? So I figured I should report it. Hopefully it's a useful error report. Feel free to close if it's not though.

jakebailey:
Thanks for that; I'm sure it's a bug! TS only "recently" started complaining about relating enums with compatible values, so I'm sure there are just more edge cases there.

jakebailey:
The reason the package gets larger is thanks to this, which is needed in the public API (unfortunately people use this object even though it's internal so I'm afraid to remove it):

image

I had a WIP esbuild change which would have helped improve this; I need to resurrect that change and see if it's still workable. Basically, I had a version of `__export` which allowed you to specify ahead of time which values are `const`, though I don't quite know how I even had that working given `__export` is called _before_ the module props are defined...

jakebailey:
Perf positive, that's nice. Bun seems to be slightly slower but unsure if that's real as it's so tiny.

jakebailey:
Going to close this; we're moving more in the direction of being ESM, so there's no point in trying to make the diagnostics tree shakable in this way since we'd want to share the code.

Issue #42360, opened 2021-01-15
Private check can be bypassed by ElementAccessExpression


Jack-Works:
# Bug Report

### 🔎 Search Terms

bypass private element access

### 🕗 Version & Regression Information

- I tested it on the master branch

### 💻 Code

```ts
class X { private x: number }
new X()['x']
```

### 🙁 Actual behavior

No error

### 🙂 Expected behavior

Type Error

MartinJohns:
It's intentional: https://github.com/microsoft/TypeScript/issues/12821#issuecomment-266249583

ajafff:
Duplicate of #19335

Some use it as a feature, e.g. to test private state. Some abuse it.

I have written a lint rule that applies the same rules of property access to element access: https://github.com/fimbullinter/wotan/blob/master/packages/mimir/docs/no-restricted-property-access.md

PR #55903, opened 2023-09-28
fix(55899): Indexing/element access on super avoids instance property checks

### LINKED ISSUE #55899: Indexing/element access on `super` avoids instance property checks
Discovered in conversation at https://github.com/microsoft/TypeScript/pull/55892#discussion_r1339658106

```ts
class C {
    yadda = () => {};
}

class D extends C {
    badda() {
        super["yadda"]();
    }
}

new D().badda()
```

Currently there is no error for this code; however, if you switched it to a property access, you'd get the error introduced at https://github.com/microsoft/TypeScript/pull/54056.

```ts
class C {
    yadda = () => {};
}

class D extends C {
    badda() {
        super.yadda(); // ❌ errors, as expected
    }
}
```

a-tarasyuk:
Fixes #55899


DanielRosenwasser:
If my memory serves me right, indexed accesses might've been an escape hatch to opt out of visibility modifiers. Maybe @RyanCavanaugh remembers better.

That said, let's see how this works

@typescript-bot test top200

DanielRosenwasser (approved):

DanielRosenwasser:
Oh my

jakebailey:
So I guess a lot of people accidentally stumbled on this hole and used it to violate accessible rules?

Jack-Works:
yeah, this has been repeatedly found as a "bug" but actually is a feature (https://github.com/microsoft/TypeScript/issues/42360).

maybe the team should add some comment in the checker, or emit a Message (hint) in IDE that this is violating visibility so people won't repeatedly report/fix this.

jakebailey:
Personally, I don't think this is a good hole to have. If someone wants to violate visibility rules, they can `as any`. I have never thought to try and skip past visibility via `this["..."]`. The list of breaks isn't _that_ long.

DanielRosenwasser:
I don't think we can remove the accessibility checks without a broader discussion. If we want to reduce the scope to just `super["..."]` for instance property checks, that's fine, but we would need to bring this to a design meeting

weswigham:
> Personally, I don't think this is a good hole to have.

> I don't think we can remove the accessibility checks without a broader discussion.

We could always put it behind a strictness option if we're worried about it being a break - it's pretty in-line with the existing indexed-access relaxing option, [suppressImplicitAnyIndexErrors](https://www.typescriptlang.org/tsconfig/#suppressImplicitAnyIndexErrors), though admittedly not implicit-`any`-related, so maybe we shouldn't bundle it into that option specifically.

sandersn:
> We could always put it behind a strictness option

If the option joins the `--strict` family, then we would need some additional way to disable it for just test files, since otherwise-strict code bases might still want to use it for tests. Actually, that would be nice whether or not it's part of the --strict.

weswigham:
@DanielRosenwasser didn't we bring this up in a design meeting? Did we come to a conclusion on if we should merge this as-is or flag it as a strictness improvement?

a-tarasyuk:
@DanielRosenwasser @RyanCavanaugh @weswigham @sandersn This PR has been open for a while., should these changes be under the `suppressImplicitAnyIndexErrors` option, or should the current approach remain unchanged?

a-tarasyuk:
@DanielRosenwasser @RyanCavanaugh @weswigham @sandersn This PR has been open for a while., should these changes be under the `suppressImplicitAnyIndexErrors` option, or should the current approach remain unchanged?

DanielRosenwasser:
This didn't come up at a design meeting yet - we should discuss it.

That said, the affected test has a comment explaining the behavior:

```ts
    // Indexed access expressions have always permitted access to private and protected members.
    // For consistency we also permit such access in indexed access types.
```

So it probably makes more sense to have under a `--strict` flag. What the flag name is, I don't know.

a-tarasyuk:
> This didn't come up at a design meeting yet - we should discuss it.

@DanielRosenwasser thank you for your feedback. I’d appreciate it if you could add it to the _design meeting queue_ when you get a chance.

Issue #55907, opened 2023-09-29
Template string wrongly infers type from a default value used in destructuring assignment


LasAtt:
### 🔎 Search Terms

"template string", "default value", "infer type", "destructuring"

### 🕗 Version & Regression Information

This changed between versions `5.1.6` and `5.2.2`. The bug is also present in `5.3.0-dev.20230928`.

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.3.0-dev.20230928#code/C4TwDgpgBACghgJzgWwM5QLxQN4CgpQBucANgK4QD8AXFKsAgJYB2A5lAD5TNnIBGEBLgC+uXADMyzAMbBGAe2ZQIADxRgSEABRhEKCMEGpa8JGgCUOfFGmL6OIqQqYoAcgCMAJgDMrqMJddMwMjawQDMgQlAAMAEmxicghhaMwMLFdvABYAVlcRXCA

### 💻 Code

```ts
type Params = {
  value?: string | number
}

function example(parameters: Params) {
  const { value = '123' } = parameters
  return `${value}` === '345' // <--- Error: This comparison appears to be unintentional because the types '"123"' and '"345"' have no overlap.(2367)
}
```

### 🙁 Actual behavior

The template string expression, `${value}`, is assigned type `"123"`, based on a default value assigned to `value` in a destructing assignment.

### 🙂 Expected behavior

The expression should instead be assigned a type of `string`. The field `value` is correctly assigned a type of `string | number`, so any template string using `value` should not be restricted to the default value used in destructuring.

Andarist:
I'm almost certain that this is a fall out of https://github.com/microsoft/TypeScript/pull/53907 and it's definitely a bug. I'll look into this soon.

LasAtt:
> I'm almost certain that this is a fall out of #53907 and it's definitely a bug. I'll look into this soon.

Thank you for looking into it!

Andarist:
It's actually an old bug but this one is a new manifestation of it. [TS playground](https://www.typescriptlang.org/play?ts=5.0.4#code/C4TwDgpgBACghgJzgWwM5QLxQN4CgpQBucANgK4QD8AXFKsAgJYB2A5lAD5TNnIBGEBAG5cAXxG4AJhADGJRNBkB7ZvShhEKCMEGpa8JGhHLVwHEVIVMUAEQBGAEwBmG1FHWNh7bokQeyKABRHHwoHTUsYnIIABoxXCA)

PR #55910, opened 2023-09-29
Fixed na issue with evaluator using binding element initializers as const values

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/55907

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -44674,7 +44674,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                             return location ? evaluateEnumMember(expr, symbol, location) : getEnumMemberValue(symbol.valueDeclaration as EnumMember);
                         }
                         if (isConstantVariable(symbol)) {
-                            const declaration = symbol.valueDeclaration as VariableDeclaration | undefined;
+                            const declaration = symbol.valueDeclaration as VariableDeclaration | BindingElement | undefined;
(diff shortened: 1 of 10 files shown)


src/compiler/checker.ts
```diff
@@ -44675,7 +44675,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                         }
                         if (isConstantVariable(symbol)) {
                             const declaration = symbol.valueDeclaration as VariableDeclaration | undefined;
-                            if (declaration && !declaration.type && declaration.initializer && (!location || declaration !== location && isBlockScopedNameDeclaredBeforeUse(declaration, location))) {
+                            if (declaration?.kind === SyntaxKind.VariableDeclaration && !declaration.type && declaration.initializer && (!location || declaration !== location && isBlockScopedNameDeclaredBeforeUse(declaration, location))) {
```
Andarist, line 44678: alternatively, I could check if the parent is not a `BindingElement`

jakebailey: Would that break on parameters? I guess not, since `isConstantVariable` won't be true.

jakebailey: I'll also note that this feels weird because `declaration` is `VariableDeclaration` which _must_ have `kind === SyntaxKind.VariableDeclaration`.

Andarist: > I'll also note that this feels weird because declaration is VariableDeclaration which must have kind === SyntaxKind.VariableDeclaration.

Good catch! I didn't notice it. It is `VariableDeclaration` just because of the cast - so I'll make that cast more accurate in a second :)

jakebailey:
@typescript-bot test top100
@typescript-bot run dt

jakebailey (approved):

PR #55913, opened 2023-09-29
Update dprint-typescript to 0.88.1, reformat

jakebailey:
This fixes a formatting bug when using multi-line strings.

### DIFF
--- a/.dprint.jsonc
+++ b/.dprint.jsonc
@@ -51,7 +51,7 @@
     ],
     // Note: if adding new languages, make sure settings.template.json is updated too.
     "plugins": [
-        "https://plugins.dprint.dev/typescript-0.87.1.wasm",
+        "https://plugins.dprint.dev/typescript-0.88.1.wasm",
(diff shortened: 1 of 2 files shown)


sheetalkamat (approved):
Why is this update not auto like other dependencies ?

jakebailey:
I could make the update job do that, yeah. It'd just have to also commit the new formatting, and I was not sure if that was desirable or not given there _could_ be dprint bugs that come up.

sheetalkamat:
It could do auto update when no change is needed in code and create PR with update of dprint and code  otherwise?

jakebailey:
Yeah, I'll see what I can do there. That should be possible.

jakebailey:
I think I'll be able to reuse the workflow code from my WIP cherry-pick pipeline replacement workflow (hopefully done soon, doing like 50 things at once) Both workflows are situations where we want to push to a specific branch and then send a PR if there isn't one yet.

For now though, I'm just going to merge this.

PR #55923, opened 2023-09-30
Speed up GHA workflows, use consistent bot token, etc

jakebailey:
An amalgamation of various fixups and improvements:

- Use the bot's token everywhere rather than `contents: write`. When using `contents: write`, GHA workflows won't run when pushed as GHA tries to prevent loops by poisoning the token. A PAT works around this.
- Don't run tests in `sync-branch`. This is slightly risky given baselines might fail, but I don't think we've seen that happen and @DanielRosenwasser reports that this delay is annoying.
- Update `package-lock.json` when changing `package.json`'s version. This is harmless but generally annoying that `package-lock` is not up to date on release branches. Honestly, I'm shocked that `npm ci` is okay with that state.
- Remove `update-lkg.yml`, which could only ever update LKG on main, which I removed a while back.
- Use a "blobless" clone rather than a full clone when we need history; this mode downloads the commit info, but none of the file contents until they are actually needed. This brings a full clone of our repo in CI down from 4m to 1m30m or so. See: https://github.blog/2020-12-21-get-up-to-speed-with-partial-clone-and-shallow-clone/ https://github.com/actions/checkout/pull/1396#issuecomment-1741638179


.github/workflows/update-package-lock.yaml
```diff
@@ -46,7 +43,7 @@ jobs:
             echo "No change."
           else
             npm test
-            npx hereby lkg
+            npx hereby LKG
```
DanielRosenwasser, line 46: What's the point of the LKG? Just as a test to see that it works?

.github/workflows/accept-baselines-fix-lints.yaml
```diff
@@ -16,11 +16,10 @@ jobs:
   build:
     runs-on: ubuntu-latest
 
-    permissions:
-      contents: write
-
     steps:
       - uses: actions/checkout@v3
```
DanielRosenwasser, line 20: do we want this to be v4 also?

DanielRosenwasser:
Thanks for sending this! I don't know the nuances of `permissions.contents: "write"` so I can't sign off right away, but doing the blobless clone is a great idea for these.

.github/workflows/update-package-lock.yaml
jakebailey: E.g. the LKG script silently broke because a dep updated, or something. Otherwise we wouldn't know until we attempted to do a release or nightly or something.

.github/workflows/accept-baselines-fix-lints.yaml
jakebailey: Surprisingly, this doesn't do a full fetch, so I don't need to bump it.

I wanted to bump all of the versions but I'd need to update every file's version specifiers which felt noisy. Probably, we should enable dependabot but just for GHA. But either of those I'd rather in a different PR.

jakebailey:
> Thanks for sending this! I don't know the nuances of `permissions.contents: "write"` so I can't sign off right away

This should be safe; we already use that in the `package-lock` workflow.

jakebailey:
Any other concerns?

.github/workflows/update-package-lock.yaml
jakebailey: (this particular change is just to make all of the workflows have the same casing; silly, I know)

RyanCavanaugh (approved):

DanielRosenwasser (approved):

Issue #52345, opened 2023-01-20
Performance regression from #48044


epmatsw:
# Bug Report

It seems like a fairly significant performance regression was introduced in this https://github.com/microsoft/TypeScript/pull/48044, which was shipped in TS 4.7. In TS 4.6, it takes ~30 seconds to run `yarn tsc` in our repo, and the reported strict subtype cache size is 24_011. In TS 4.7, it takes 75+ seconds to run `yarn tsc` (~2.5x longer than previous), and the reported strict subtype cache size is 122_778 (~5x larger than previous).

I verified that it was that specific PR that regressed the performance by taking a locally-installed 4.6.4 and applying the changes from that PR directly.

Another symptom is that several types (mostly spreads of objects into others) are now reporting as "Expression produces a union type that is too complex to represent."

### 🔎 Search Terms

`Expression produces a union type that is too complex to represent`, `strict subtype cache size`

### 🕗 Version & Regression Information

- Regressed in https://github.com/microsoft/TypeScript/pull/48044
- This changed between versions 4.6.4 and 4.7.3

|Version|Time|Strict Subtype Cache Size|
|-------|-----|--------------------------|
|4.6.4  | 33s | 24011                                  |
|4.7.3  | 75s | 122778                                 |
|4.9.4  | 77s | 122869                                |
|5.0.0-dev.20230120 | 78s | 122869 |

### 💻 Code

I haven't been able to narrow down a minimal reproduction of this issue, but if someone points me in the right direction I'm happy to work on it.

### 🙁 Actual behavior

Running `yarn tsc --noEmit --extendedDiagnostics` takes ~75 seconds in our repo, with a Strict subtype cache size of 122778.

### 🙂 Expected behavior

Running `yarn tsc --noEmit --extendedDiagnostics` takes ~30 seconds in our repo, with a Strict subtype cache size of 24011.

DanielRosenwasser:
Would you be able to use our analyze-trace tool?

https://github.com/microsoft/typescript-analyze-trace

It should highlight specific hot spots that might occur as of TS 4.7. The [usage instructions](https://github.com/microsoft/typescript-analyze-trace#usage) are a little better now. Let me know if you need any more info using it.

epmatsw:
Not public, unfortunately. Sorry about the delay on the trace, I’ll try to get something going tomorrow.

epmatsw:
Dang, thought I had a reproduction but apparently not. Probably not gonna get something until Monday now :(

[...]

PR #55926, opened 2023-09-30
Bail early in getNarrowedTypeWorker if type is candidate

jakebailey:
This technically handles https://github.com/microsoft/TypeScript/issues/52345#issuecomment-1737898885 and maybe #55948, bringing it back to <=4.7 levels, but I think it's a little silly to narrow something to itself. But, probably cheap to check anyhow.

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -28163,6 +28163,12 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             if (type.flags & TypeFlags.AnyOrUnknown) {
                 return candidate;
             }
+
+            // TODO(jakebailey): should this just be type === candidate?
+            if (isTypeIdenticalTo(type, candidate)) {
+                return candidate;
+            }
+
             // We first attempt to filter the current type, narrowing constituents as appropriate and removing
             // constituents that are unrelated to the candidate.
             const isRelated = checkDerived ? isTypeDerivedFrom : isTypeSubtypeOf;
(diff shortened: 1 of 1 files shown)


jakebailey:
@typescript-bot pack this

jakebailey:

jakebailey:
I still feel like this might be slightly silly, but it seems really cheap and quick to just check for cases where we're trying to narrow something to itself and bail early without doing all of the other work.

gabritto:
The change makes sense, but where in type checking the motivating case https://github.com/epmatsw/tsdemo/blob/main/slow.ts do we call `getNarrowedType` with `type` and `candidate` being the same?

jakebailey:
I think the example has been updated; it used to be narrowing something to itself.

gabritto:
> I think the example has been updated; it used to be narrowing something to itself.

Do you have an example then?

jakebailey:
The repro repo has it if you go back a few commits: https://github.com/epmatsw/tsdemo/tree/b2a2af937e492819045ea4e880a56236494c2147

Something like:

```ts
function test(base: SpecificString) {
  if (isSpecificString(base)) {
    return base;
  }
}
```

Where `SpecificString` is a large union of literals.

gabritto:
Ah, so it is an explicit narrowing to itself. I don't know how common it is, it doesn't seem like something you want to be doing, but then again you might be programming defensively and you want to validate at run time that something has a type, so I can see a case for writing this kind of code.
I think I agree that, as long as it's cheap to check and doesn't make anything slower, there's not a lot of harm in doing that.

gabritto (approved):

weswigham (approved):
Don't know if you're still waiting on my review for this, but this can be a reminder to merge this at some point~

jakebailey:
I was, kinda; mainly I was trying to decide if this was even worthwhile if it's so rare and silly.

PR #55941, opened 2023-10-02
Fixed an issue with reverse mapped types inference when single type variable is left after inferring from matching types

Andarist:


src/compiler/checker.ts
```diff
@@ -25078,7 +25078,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     target = getIntersectionType(targets);
                 }
             }
-            else if (target.flags & (TypeFlags.IndexedAccess | TypeFlags.Substitution)) {
+            if (target.flags & (TypeFlags.IndexedAccess | TypeFlags.Substitution)) {
```
Andarist, line 25081: preceding logic (such as the one responsible for inferring from matching types in unions [here](https://github.dev/microsoft/TypeScript/blob/747250aca4d4f188b5c2bcfdf58a1c8d98741eb9/src/compiler/checker.ts#L25044-L25054)) could "reduce" the target in various ways - potentially to leaving indexed access or substitution in the `target` (when the target doesn't start as such). For that reason, we always need to `getActualTypeVariable` here as it should be called on the "reduced" target and not based on the pre-reduction flags (which sort of happened when this condition was used as just a fallback branch)

weswigham (approved):
Obligatory "Does this have a matching issue?", but this looks good to me. Honestly surprising if we haven't gotten any issue reports about inference failure that boil down to this, since it doesn't seem *that* hard to trigger.

PR #55943, opened 2023-10-02
Don't widen unique symbol for different declarations

### LINKED ISSUE #55901: unique symbol lost on assignment to const despite type assertion
### 🔎 Search Terms

unique symbol widening

### ⏯ Playground Link

[Playground Link](https://www.typescriptlang.org/play?#code/N4KABBYMYPYHYGcAuYCuYC8YBEBBbA3OJLIigIYCMmaY5CYSAngA4CmMAZrQPQ9hwYKAO4BLACZs4bccQilkdAEw10fAULBjJ02ZDAAbNhQDMqug2bsuvfoJESpMuYeN0ALOfXansgL4gIKD6CijoWADKTAC2AEYwBgAUAJREIfCKVOb0jKwc3Gr8PrpggLwbgKU7LqHKXkWOui5GptmWeTaFWvUyZZX6TR61nTrOfkA)

### 💻 Code

```ts
const u = Symbol();
const a7 = u as typeof u // widened  to symbol
```

### 🙁 Actual behavior

`a7` is `symbol`

### 🙂 Expected behavior

`a7` has `typeof u`

This issue is a huge problem for isolated declarations as it breaks the assumption that in an expression `const o = a as T` we can always use `T` as the type of  `o`. This is also inconsistent with the way similar expressions work for string/number literal types ([ex](https://www.typescriptlang.org/play?#code/N4KABBYMYPYHYGcAuYCuYC8YBEBBbA3OJLIigIYCMmaY5CYSAngA4CmMAZrQPQ9hwYKAO4BLACZs4bccQilkdAEw10fAULBjJ02ZDAAbNhQDMqug2bsuvfoJESpMuYeN0ALOfXansgL4gIKD6CijoWADKTAC2AEYwBgAUAJREIfCKVOb0jKwc3Gr8PrpggLwbgKU7LqHKXkWOui5GptmWeTaFWvUyZZX6TR61nTrOfkA))

jakebailey:
Fixes #55901

This is mainly for discussion for now. It appears as though this case is needed in some cases, e.g. in `tests/baselines/reference/uniqueSymbolsDeclarations.errors.txt`.

```
uniqueSymbolsDeclarations.ts(104,7): error TS2527: The inferred type of 'constInitToLReadonlyType' references an inaccessible 'unique symbol' type. A type annotation is necessary.
uniqueSymbolsDeclarations.ts(105,7): error TS2527: The inferred type of 'constInitToLReadonlyNestedType' references an inaccessible 'unique symbol' type. A type annotation is necessary.
```

Which is what @DanielRosenwasser hinted at.


tests/baselines/reference/uniqueSymbolsDeclarations.errors.txt
```diff
@@ -0,0 +1,247 @@
+uniqueSymbolsDeclarations.ts(104,7): error TS2527: The inferred type of 'constInitToLReadonlyType' references an inaccessible 'unique symbol' type. A type annotation is necessary.
+uniqueSymbolsDeclarations.ts(105,7): error TS2527: The inferred type of 'constInitToLReadonlyNestedType' references an inaccessible 'unique symbol' type. A type annotation is necessary.
```
jakebailey, line 2: These are the cases this breaks an would need to be solved differently; perhaps we can detect this through visibility instead.

tests/baselines/reference/uniqueSymbols.types
```diff
@@ -524,10 +524,10 @@ const o2 = {
 >s : unique symbol
 
     method5(p = s) { return p; },
->method5 : (p?: symbol) => symbol
->p : symbol
+>method5 : (p?: unique symbol) => symbol
```
jakebailey, line 527: These kinds of changes don't make me happy and probably indicate that this change is probably not good as-is.

RyanCavanaugh:
@typescript-bot pack this

jakebailey:
referencing https://github.com/microsoft/TypeScript/pull/55860 before I delete this branch

Issue #55945, opened 2023-10-02
Inlay type hints don't work for bigint literals


DanielRosenwasser:
```ts
const yadda: 1n = 1n;

export function helper(b = yadda) {
}
```

When using inlay type hints for parameters (using these settings in VS Code)

```json5
"editor.inlayHints.enabled": "on",
"typescript.inlayHints.parameterTypes.enabled": true,
```

I get the following stack trace

```
Error: Debug Failure. Unexpected node.
Node BigIntLiteral was unexpected.
    at visitForDisplayParts (.../node_modules/typescript/lib/tsserver.js:164854:17)
    at visitForDisplayParts (.../node_modules/typescript/lib/tsserver.js:164801:11)
    at typeToInlayHintParts (.../node_modules/typescript/lib/tsserver.js:164582:5)
    at getParameterDeclarationTypeHints (.../node_modules/typescript/lib/tsserver.js:164547:12)
    at visitFunctionLikeForParameterType (.../node_modules/typescript/lib/tsserver.js:164531:25)
    at visitor (.../node_modules/typescript/lib/tsserver.js:164321:9)
    at visitNodes (.../node_modules/typescript/lib/tsserver.js:30187:22)
    at forEachChildInSourceFile (.../node_modules/typescript/lib/tsserver.js:30400:12)
    at forEachChild (.../node_modules/typescript/lib/tsserver.js:30696:35)
    at visitor (.../node_modules/typescript/lib/tsserver.js:164327:12)
    at Object.provideInlayHints (.../node_modules/typescript/lib/tsserver.js:164288:3)
    at Object.provideInlayHints2 [as provideInlayHints] (.../node_modules/typescript/lib/tsserver.js:145334:34)
    at IpcIOSession.provideInlayHints (.../node_modules/typescript/lib/tsserver.js:183679:48)
    at provideInlayHints (.../node_modules/typescript/lib/tsserver.js:182838:43)
    at .../node_modules/typescript/lib/tsserver.js:184857:69
    at IpcIOSession.executeWithRequestId (.../node_modules/typescript/lib/tsserver.js:184849:14)
    at IpcIOSession.executeCommand (.../node_modules/typescript/lib/tsserver.js:184857:29)
    at IpcIOSession.onMessage (.../node_modules/typescript/lib/tsserver.js:184899:51)
    at process.<anonymous> (.../node_modules/typescript/lib/tsserver.js:186480:14)
    at process.emit (node:events:513:28)
[...]

PR #55947, opened 2023-10-02
fix(55945): Inlay hints don't work for bigint literals

a-tarasyuk:
Fixes https://github.com/microsoft/TypeScript/issues/55945

### DIFF
--- a/src/services/inlayHints.ts
+++ b/src/services/inlayHints.ts
@@ -2,6 +2,7 @@ import {
     __String,
     ArrayTypeNode,
     ArrowFunction,
+    BigIntLiteral,
     CallExpression,
     ConditionalTypeNode,
     ConstructorTypeNode,
@@ -735,6 +736,9 @@ export function provideInlayHints(context: InlayHintsContext): InlayHint[] {
                         parts.push({ text: ": " });
                         visitForDisplayParts(propertySignature.type);
                     }
+                    break;
+                case SyntaxKind.BigIntLiteral:
+                    parts.push({ text: (node as BigIntLiteral).text });
                     break;
                 default:
                     Debug.failBadSyntaxKind(node);
--- a/src/services/inlayHints.ts
+++ b/src/services/inlayHints.ts
@@ -756,7 +756,7 @@ export function provideInlayHints(context: InlayHintsContext): InlayHint[] {
         function getLiteralText(node: LiteralExpression) {
             if (isStringLiteral(node)) {
                 const text = escapeString(node.text);
-                return quotePreference === QuotePreference.Single ? `'${text}'` : `"${text}"`
+                return quotePreference === QuotePreference.Single ? `'${text}'` : `"${text}"`;
             }
             return node.text;
         }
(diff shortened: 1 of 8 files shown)


DanielRosenwasser (approved):

MariaSolOs:
I would prefer adding an extra check before the switch to handle literal nodes:
```ts
if (tokenString) {
  parts.push({ text: tokenString });
  return;
}

if (isLiteralExpression(node)) {
  parts.push({ text: node.text });
  return;
}

switch (node.kind) { ...
```

MariaSolOs (approved):

src/services/inlayHints.ts
```diff
@@ -473,6 +472,11 @@ export function provideInlayHints(context: InlayHintsContext): InlayHint[] {
                 return;
             }
 
+            if (isLiteralExpression(node)) {
+                parts.push({ text: isStringLiteral(node) ? `"${node.text}"` : node.text });
```
DanielRosenwasser, line 476: This has to be escaped technically. Can you add a test for the string literals `'"'` and `"\\"`?

DanielRosenwasser (changes requested):

src/services/inlayHints.ts
MariaSolOs: Would it be an overkill to use the quote kind from `UserPreferences` here?

DanielRosenwasser: That's probably the ideal thing to do.

Sorry for not making this an easy fix @a-tarasyuk. 😄

a-tarasyuk: @MariaSolOs Is this how you wanted to use `quotePreference` https://github.com/microsoft/TypeScript/pull/55947/commits/9700893341028bdc53989df7c6623b3c0b4af10c?

tests/baselines/reference/inlayHintsQuotePreference2.baseline
```diff
@@ -0,0 +1,33 @@
+export function fn(a = a1, b = b1) {}
+                    ^
+{
+  "text": "",
+  "displayParts": [
+    {
+      "text": ": "
+    },
+    {
+      "text": "'''"
```
MariaSolOs, line 10: Hmmm using the quote preference here means that this is no longer a single quote literal, and I wonder if this might end up confusing people in some cases. @DanielRosenwasser wdyt about this?

src/services/inlayHints.ts
MariaSolOs: @a-tarasyuk It is. I just have a couple of concerns and I want to check in with @DanielRosenwasser first. Sorry for all the trouble!

tests/baselines/reference/inlayHintsQuotePreference2.baseline
DanielRosenwasser: You need to still use `escapeString`

MariaSolOs: @a-tarasyuk I think you need to escape the string after quoting it. Notice how this is still `'''` which is not a single quote literal.

MariaSolOs (approved):

MariaSolOs:
LGTM! I'll merge this now since you've addressed @DanielRosenwasser's (and my) suggestions.

Thanks again for the fix @a-tarasyuk (and your patience 😅)

PR #55985, opened 2023-10-04
Recognize `export * as default` as a syntactic default export for synthetic default detection

### LINKED ISSUE #55829: `export * as default` in declaration file loses members
### 🔎 Search Terms

Reexport, re-export, default, namespace, missing type

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about modules

### ⏯ Playground Link

https://www.typescriptlang.org/dev/bug-workbench/?moduleResolution=100&emitDeclarationOnly=true&ts=5.2.2#code/PTAEAEFsHsBMFcA2BTAXKZBnAdsgHgC4CwAUCBDAigEpbSLwECW026ARvNrCgE6mly4AGZMU2AIaQ0oAgAteyZADpYygplL4ADtF4FZAT23JQAOSnJYoALygADAG4BZMCLHJJ09AQDu0VXVNV1BQsPCI0AA9KMEwSITQxRgAN1N5UwByVUzQCW5ZOVNkXl49UABzaCw83wlDUC08XX1QACo8zFBYZGEJJANhMshQACJlYHlFZFHnEjiIUXFLdFYVDVImSBaDbC6h6BHxyf9Z0gJjUwBBRCYJLrs95QtpWEcw8gBRamoAeWpQAAmABsAE4ACzoF5YbQSADGWVGJ2go1ycnuoGw0AwzT0BCsoGkkHYJVAmWhsGyCzCUQA-EA

### 💻 Code

```ts repro
// @module: esnext
// @moduleResolution: bundler

// @filename: three.d.ts
export type Named = 0;

// @filename: two.d.ts
//                ^^
//                WORKAROUND2: Remove the '.d' and the error goes away 
export type * as default from "./three";

// @filename: one.ts
import type ns from "./two";
type Alias = ns.Named;   // <- error ts(2694)
```

### 🙁 Actual behavior

`ERROR 2694: Namespace '"/two"' has no exported member 'Named'.`  The namespace is empty.

### 🙂 Expected behavior

The exported type `Named` should be retained in the re-exported namespace.

The issue is not specific to using type-only re-exports.  `export * as default` [also exhibits the problem.](https://www.typescriptlang.org/dev/bug-workbench/?moduleResolution=99&module=199&emitDeclarationOnly=true&ts=5.2.2#code/PTAEAEDMEsBsFMB2BDAtvAXKALgCwE7zwB0AJsdgM4CwAUKfAMazKGiMD2il2oAcmniksABgDcdeAA8ADh3y8A3v0GlQAXwm06ICDAQp0WbAHcOZCjVrS5C0ACpQySqAaRkAV1i9I+DqlAAImJgPEJ4QK0dMCg4JEEsLhIqOmhUW15uUF9-IJDTDki6bmIBdFIxIA)

The error goes away if we make any of the following changes/workarounds.

### 1. Re-export via a non-default name

The problem only occurs when using a default export to expose the namespace.

- Fails: `export type * as default from "./three";`
- Succeeds:  `export type * as foo from "./three";`

### 2. Manually re-export

Desugaring the re-export fixes the problem.

- Fails: `export type * as default from "./three";`
- Succeeds:  `import type * as NS from "./three"; export type {NS as default}`

### 3. Re-export in a regular `*.ts` file

The problem only exists when the re-export is in a `*.d.ts` file.

```ts repro
// @filename: two.ts  <-- this works fine!
export type * as default from "./three";
```

(We discovered the problem when performing post-build type-checking on the emitted declaration files.)

### 4. Using a different `"moduleResolution"` option

Not in the workbench, but when recreating in a local project the issue did not occur when `moduleResolution` was defaulting to `'classic'`

andrewbranch:
Fixes #55829

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -4022,7 +4022,10 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     }
 
     function isSyntacticDefault(node: Node) {
-        return ((isExportAssignment(node) && !node.isExportEquals) || hasSyntacticModifier(node, ModifierFlags.Default) || isExportSpecifier(node));
+        return ((isExportAssignment(node) && !node.isExportEquals)
+            || hasSyntacticModifier(node, ModifierFlags.Default)
+            || isExportSpecifier(node)
+            || isNamespaceExport(node));
     }
 
     function getUsageModeForExpression(usage: Expression) {
--- a/tests/cases/conformance/es2020/modules/exportAsNamespace5.ts
+++ b/tests/cases/conformance/es2020/modules/exportAsNamespace5.ts
@@ -3,6 +3,7 @@
 
 // @filename: three.d.ts
 export type Named = 0;
+declare const Named: 0;
 
 // @filename: two.d.ts
 export * as default from "./three";
@@ -10,3 +11,4 @@ export * as default from "./three";
 // @filename: one.ts
 import ns from "./two";
 type Alias = ns.Named;
+ns.Named;
--- a/tests/cases/conformance/es2020/modules/exportAsNamespace5.ts
+++ b/tests/cases/conformance/es2020/modules/exportAsNamespace5.ts
@@ -1,0 +1,12 @@
+// @module: esnext
+// @moduleResolution: bundler
+
+// @filename: three.d.ts
+export type Named = 0;
+
+// @filename: two.d.ts
+export * as default from "./three";
+
+// @filename: one.ts
+import ns from "./two";
+type Alias = ns.Named;
--- a/tests/baselines/reference/exportAsNamespace5.js
+++ b/tests/baselines/reference/exportAsNamespace5.js
@@ -2,6 +2,7 @@
 
 //// [three.d.ts]
 export type Named = 0;
+declare const Named: 0;
 
 //// [two.d.ts]
 export * as default from "./three";
@@ -9,7 +10,9 @@ export * as default from "./three";
 //// [one.ts]
 import ns from "./two";
 type Alias = ns.Named;
+ns.Named;
 
 
 //// [one.js]
-export {};
+import ns from "./two";
+ns.Named;
--- a/tests/baselines/reference/exportAsNamespace5.js
+++ b/tests/baselines/reference/exportAsNamespace5.js
@@ -1,0 +1,15 @@
+//// [tests/cases/conformance/es2020/modules/exportAsNamespace5.ts] ////
+
+//// [three.d.ts]
+export type Named = 0;
+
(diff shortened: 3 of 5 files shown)


src/compiler/checker.ts
```diff
@@ -4022,7 +4022,10 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     }
 
     function isSyntacticDefault(node: Node) {
-        return ((isExportAssignment(node) && !node.isExportEquals) || hasSyntacticModifier(node, ModifierFlags.Default) || isExportSpecifier(node));
+        return ((isExportAssignment(node) && !node.isExportEquals)
+            || hasSyntacticModifier(node, ModifierFlags.Default)
+            || isExportSpecifier(node)
+            || isNamespaceExport(node));
```
DanielRosenwasser, line 4028: Is the following covered under these rules?

```ts
import * as yadda from "./yadda";
export { yadda as default };
```

andrewbranch: Yes (`isExportSpecifier`)

andrewbranch: This function is only called when the symbol attached to the declaration was reached by the name `default`, which is why these don’t check specifically check for defaultness.

andrewbranch: It’s intended to distinguish between these ES module syntaxes and something like

```ts
export = {
  options: {},
  default: {},
}
```

DanielRosenwasser (approved):

DanielRosenwasser:
@typescript-bot test top200

tests/cases/conformance/es2020/modules/exportAsNamespace5.ts
```diff
@@ -0,0 +1,12 @@
+// @module: esnext
+// @moduleResolution: bundler
+
+// @filename: three.d.ts
+export type Named = 0;
```
robpalme, line 5: Please could we also test the value land namespace properties. Either in this test case or as a separate one.

https://www.typescriptlang.org/dev/bug-workbench/?moduleResolution=99&module=199&emitDeclarationOnly=true&ts=5.2.2#code/PTAEAEDMEsBsFMB2BDAtvAXKALgCwE7zwB0AJsdgM4CwAUKfAMazKGiMD2il2oAcmniksABgDcdeAA8ADh3y8A3v0GlQAXwm06ICDAQp0WbAHcOZCjVrS5C0ACpQySqAaRkAV1i9I+DqlAAImJgPEJ4QK0dMCg4JEEsLhIqOmhUW15uUF9-IJDTDki6bmIBdFIxIA

robpalme (commented):
LGTM - thank you for the quick fix!

PR #55991, opened 2023-10-05
Fix fallthrough and default in new switch-true narrowing

### LINKED ISSUE #55986: Narrowing switch(true) doesnt work for default case and is wrong for fallthrough cases in 5.3
### 🔎 Search Terms

switch, fallthrough

- This is a crash No
- This changed between versions 5.2 and 5.3

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.3.0-dev.20231004#code/C4TwDgpgBAZg9nKBeKByAhqqAfNAjLXVAY1QCgzi4A7AZ2FmTUynVtgQtoHcBLYYgAsAFMABOAVwgBKKAG8yUJVGJtoMZEhQZUALkXLVtdZu0F9ywzVpwANhAB0tuAHNhMaQcsB6b5f9QAHoA-F7KeGIQ6ADWYQAmEDDoErbAFv5UdHaOzm4eYUq+AZYhZAC+QA

### 💻 Code

```ts
type foo = 'a' | 'b' | 'c'

const f = 'a' as foo

switch(true) {
    case f === 'a':
    case f === 'b':
      console.log(f)
      //          ^?  'b'
      break
    default:
      console.log(f)
      //          ^?  'a' | 'b' | 'c'
}
```

### 🙁 Actual behavior

The fallthrough case has wrong type `b`.
The default case is not narrowed.

### 🙂 Expected behavior

In the fallthrough the type should be a union of `a | b`.
The default case should be `c`.

Related to https://github.com/microsoft/TypeScript/commit/70b7de11fbadb16d30f6770332f544c44134eb82

jakebailey:
Fixes #55986

This fixes narrowing in fallthrough/default cases, narrowing the type properly against `assumeTrue` for the other cases. It now behaves like if statements do.

(We could totally also do `switch (false)` but that seems like a bad move.)


jakebailey:
@typescript-bot test top200
@typescript-bot pack this

src/compiler/checker.ts
```diff
@@ -28040,6 +28036,20 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             return getUnionType(map(clauseWitnesses, text => text ? narrowTypeByTypeName(type, text) : neverType));
         }
 
+        function narrowTypeBySwitchOnTrue(type: Type, switchStatement: SwitchStatement, clauseStart: number, clauseEnd: number): Type {
+            // Equal start and end denotes implicit fallthrough; undefined marks explicit default clause.
+            const defaultIndex = findIndex(switchStatement.caseBlock.clauses, clause => clause.kind === SyntaxKind.DefaultClause);
```
DanielRosenwasser, line 28041:
```suggestion
            const defaultIndex = findIndex(switchStatement.caseBlock.clauses, isDefaultClause);

```

src/compiler/checker.ts
```diff
@@ -28040,6 +28036,20 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             return getUnionType(map(clauseWitnesses, text => text ? narrowTypeByTypeName(type, text) : neverType));
         }
 
+        function narrowTypeBySwitchOnTrue(type: Type, switchStatement: SwitchStatement, clauseStart: number, clauseEnd: number): Type {
+            // Equal start and end denotes implicit fallthrough; undefined marks explicit default clause.
```
DanielRosenwasser, line 28040: What is undefined here?

src/compiler/checker.ts
```diff
@@ -28040,6 +28036,20 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             return getUnionType(map(clauseWitnesses, text => text ? narrowTypeByTypeName(type, text) : neverType));
         }
 
+        function narrowTypeBySwitchOnTrue(type: Type, switchStatement: SwitchStatement, clauseStart: number, clauseEnd: number): Type {
+            // Equal start and end denotes implicit fallthrough; undefined marks explicit default clause.
+            const defaultIndex = findIndex(switchStatement.caseBlock.clauses, clause => clause.kind === SyntaxKind.DefaultClause);
+            const hasDefaultClause = clauseStart === clauseEnd || (defaultIndex >= clauseStart && defaultIndex < clauseEnd);
+            if (hasDefaultClause) {
+                // Here we'd need to find the opposite of all of the other cases, but without control flow,
```
DanielRosenwasser, line 28044: It's not just the opposite because you can fall through to a default. For example

```ts
declare const f: 'a' | 'b' | 'c';

switch (true) {
  case f === "a":
  default:
    f;
  case f === "b":
}
```

Ideally we would know that `f` is `"a" | "c"`.

src/compiler/checker.ts
```diff
@@ -28040,6 +28036,20 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+            const defaultIndex = findIndex(switchStatement.caseBlock.clauses, clause => clause.kind === SyntaxKind.DefaultClause);
+            const hasDefaultClause = clauseStart === clauseEnd || (defaultIndex >= clauseStart && defaultIndex < clauseEnd);
+            if (hasDefaultClause) {
+                // Here we'd need to find the opposite of all of the other cases, but without control flow,
+                // the best we can do is to not narrow at all.
+                return type;
+            }
+            // In the non-default cause we create a union of the type narrowed by each of the listed cases.
+            const clauses = switchStatement.caseBlock.clauses.slice(clauseStart, clauseEnd);
+            return getUnionType(map(clauses, clause => clause && clause.kind === SyntaxKind.CaseClause ? narrowType(type, clause.expression, /*assumeTrue*/ true) : neverType));
```
DanielRosenwasser, line 28050: When would `clause` not be truthy?

DanielRosenwasser (changes requested):
Can we have a test of a `default` preceding a `case`?

Additionally, do we have any tests for case clauses narrowing subsequent case clauses?

```ts
type Shape =
    | { kind: "circle", radius: number }
    | { kind: "square", sideLength: number }

function wat(shape: Shape) {
    switch (true) {
        case shape.kind === "circle":
            return Math.PI * shape.radius ** 2;
        case shape.kind === "circle": // should error
    }

    if (shape.kind === "circle") {
        return Math.PI * shape.radius ** 2;
    }
    else if (shape.kind === "circle") {
        //         ~~~~
        // Property 'kind' does not exist on type 'never'.
    }
}
```

src/compiler/checker.ts
jakebailey: Never, unless we supported `switch (false)`

src/compiler/checker.ts
jakebailey: Dunno. Copied and pasted from above

jakebailey:
@andarist feel free to tackle this if you know the right answer here

src/compiler/checker.ts
DanielRosenwasser: I don't understand, these are each syntax nodes, right?

jakebailey: Oh, you're talking about `clause && clause.kind` and not `assumeTrue`.

Yes, I can undo this; I just copied and pasted from the original code which could have had it be falsy.

src/compiler/checker.ts
jakebailey: This is just the code from the `switch (typeof x)` function above. I can change it but it'll be different.

jakebailey:
I am surely wrong, we totally create branch labels for case blocks and everything.

jakebailey:
> Additionally, do we have any tests for case clauses narrowing subsequent case clauses?

I don't think your code would error in the way you expect; e.g. see what we do for typeof now: [Playground Link](https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEYD2A7AzgF3gDwFzxQFcBbAIxBngB95CVQAzASxRGAFgAoLtAdyYxgAFvAAUGAJ4AHEEgY4AlPADeXLvA2IoaBACI6jFm1251m89gDcZjQHpbGgHoB+G+fik4UANbXunczBtPQMQZlZgEzcLP3N7J1cA9w1PEB8-NyCdeF0iMgoopOSrN3j4F2jNVPS1Iq1s-Xowo0jTOpjSh3LE5JSvX1rA4Jy88hhC3pxYzTKK9qr+vwBfLiA)

jakebailey:
@typescript-bot pack this

jakebailey:
I'm still not confident but I'd appreciate people trying the playground from the most recent change and try and break it.

Clearly, this is not my area.

jakebailey:
Notably it feels suspicious that I haven't explicitly handled fallthrough, so I bet there's another edge case I'm missing where there needs to be CFA from one case group into another?

src/compiler/checker.ts
```diff
@@ -28040,6 +28036,27 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             return getUnionType(map(clauseWitnesses, text => text ? narrowTypeByTypeName(type, text) : neverType));
         }
 
+        function narrowTypeBySwitchOnTrue(type: Type, switchStatement: SwitchStatement, clauseStart: number, clauseEnd: number): Type {
+            const clauses = switchStatement.caseBlock.clauses.slice(clauseStart, clauseEnd);
+            const defaultIndex = findIndex(switchStatement.caseBlock.clauses, clause => clause.kind === SyntaxKind.DefaultClause);
+            const hasDefaultClause = clauseStart === clauseEnd || (defaultIndex >= clauseStart && defaultIndex < clauseEnd);
+            if (hasDefaultClause) {
+                // In the default case, we need to assume that all of the other cases were false,
+                // but also that any of the other cases in this block are true.
```
DanielRosenwasser, line 28045:
```suggestion
                // In the default case, it's possible that we came from:
                // 1. a set of fallthrough cases, where their conditions might be true
                // 2. no case succeeding, in which case all conditions might be false
```

DanielRosenwasser: Actually...rereading it, is it the case that we gain no information from considering the fallthrough cases? Since you have to assume they could be both true and false?

jakebailey: Yeah, that's the edge case I'm guessing at in https://github.com/microsoft/TypeScript/pull/55991#issuecomment-1749364932, I need to write a test that can show that possibility.

Not to mention that you could have a broad case and then narrow within the case and return, leaving only part of the type to fallthough...

jakebailey:
@typescript-bot pack this

jakebailey:
Okay, another better version, but probably still not there. Please break it...

src/compiler/checker.ts
jakebailey: In thinking harder, I realized that what this function should do is just "what do I know about this thing based on only the case itself?"; you don't need to worry about control flow, fallthrough, etc, here at all, because CFA will combine this info with the other cases later.

So in reality, the bug being fixed is "multiple cases were not handled" and "default case was not handled".

jakebailey:
@typescript-bot test top200

src/compiler/checker.ts
```diff
@@ -28040,6 +28036,33 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+                const clausesBefore = switchStatement.caseBlock.clauses.slice(0, clauseStart);
+                const clausesAfter = switchStatement.caseBlock.clauses.slice(clauseEnd);
+
+                const before = narrowTypeForTrueClauses(type, clausesBefore);
+                const after = narrowTypeForTrueClauses(type, clausesAfter);
+                const other = getUnionType([before, after]);
+
+                const typeNotOther = filterType(type, t => !isTypeSubsetOf(t, other));
+
+                return getUnionType([clausesType, typeNotOther]);
```
jakebailey, line 28057: I have no doubt that there's a more efficient way to do it, but I'm not sure how much it matters.

jakebailey:
@typescript-bot test top200

DanielRosenwasser:
@typescript-bot pack this

src/compiler/checker.ts
```diff
@@ -28040,6 +28036,33 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+                // that aren't covered by the other cases.
+
+                const clausesBefore = switchStatement.caseBlock.clauses.slice(0, clauseStart);
+                const clausesAfter = switchStatement.caseBlock.clauses.slice(clauseEnd);
+
+                const before = narrowTypeForTrueClauses(type, clausesBefore);
+                const after = narrowTypeForTrueClauses(type, clausesAfter);
+                const other = getUnionType([before, after]);
+
+                const typeNotOther = filterType(type, t => !isTypeSubsetOf(t, other));
```
DanielRosenwasser, line 28055: Suspicious- as mentioned, look towards `getNarrowedType`

DanielRosenwasser:
<details><summary>

Seems to work correctly from a series of combinatons of `default`s in various positions and fallthroughs ([playground link](https://www.staging-typescript.org/play?ts=5.3.0-pr-55991-37#code/JYOwLgpgTgZghgYwgAgILIN4FgBQz-IDWoAJgFzIBEclA3LgcnAApQD2ADgM4VdhSgA5vRwBfXLlCRYiFACFMDAsRDkqAIzpL861px7I+AkMNzick8NHhJkAYUV5lpCpQRan+BHu69+QkXMJHDAATw4UAFlQgBVwlABeNGQAH2QFNLsRXBgAVxAEMGA2EGRgLlQACgAPCmi4iIBKR0YuAHdgMAQAC2RK-lyIZuxPRmQSCHhcgBswMm0xxgQSvmQQCAA3aAp1rahkJOqRRa84LhRqgDoVEgOEpOpKedGT5CuWdm5j14J1KAg4IRvosEGcLtdSHcHppnj8CFddJ8uMCfn8AUCFkswW8IaooVR3LC4TjvEiUa80YCUUFRu1Oj0+gMhi0ThMprMiXDliBVrttmtNtADm9yQRQeccTd8Y8KCNiTiPvpRYtKRiXgRzK9xeCpfdoU8WXCET5kZiTqrlchNSdtZLIXqCQa5cSrqSlWaVf8qR6rQtrfgFnSur1+lBBsMfWy4DM5j7GO8TZbGBafbarrr7honnH4ZdEe71Z70Za07jbg63NnC2NXYmczovWrGDTWh1g4yw8znYsozHOT9ubzBVAdsPhUdU9j0-bMzL6wq69XGAB6ZfIeDTWbddi5QTdScS6d4isw+fGsnz1fruCbsDbti7-fV0sZh6Es+XN1fP3BGkIaZnFw6TYt2yDqGCoR+MYgjCpQHh-gBXBAQAItAwBbCQACMyAQNUkCqEBcggQsahGEIsHwcE-6AcgqECBhABMOF4RABHARKoGkf4JgUSibB3vyZE8Q8lEWDgeQFEUJTINRSGhJU4HnBQRHnAANIYbAALYQHR6EQGoukYdhaSGfpDERrSbYMqG4aGmMtqKSgoB8HABQQGwMC0WhRn9hSYKXCQl5rhuW47nukaTNGHLzo5eYQT+YmtvSIZMhZWrYlwWk6d5+llDyYCuUgHlefR+mYb5JyZdppkkAFMWNiWGVZTVeUuW5xU1QxFWLFV2WlbVgVLr8jYReysZDV4KxgAKeyjnswq9TV1LBEl7Y2V2B4oItOW3M5BXtZ5NXlfO239XVE3IFeIV3mFT7pRKp16bt+WFe5h07V1J3NTt53yqqo1ReN8qDtNfIjjNQpJI9GHLYlBBBtZqV2YwvbRRdIMQ+DYMLd9-WNQ9uNPa1+1Fe9-XHRd0P6QFSbw4TGFxecoS0w26KbRp1U7cTr0dR93VjFTA0sxzfVPZc-HdNAsOBlZKWdmlNpNZz-XcwdJVPRT8qCzTX3K2LjnM-VbPVqjQPEhjYNzZDItLezguq6T6uMfzrT09TJDC9rEtSwlMvJR2tmgViBN6xhDtvU7ZUu3Tofu57bu1Qbwv-c+Sui2He082TT2fZTCc63nse1d7UCWqb0f4NrHu6+n1Ml8LFvDlb+xQwn0tiEAA))

</summary>
<p>

```ts
interface A {
    kind: "a";
    aProps: string;
}

interface B {
    kind: "b";
    bProps: string;
}

interface C {
    kind: "c";
    cProps: string;
}


type MyType = A | B | C;

function isA(x: MyType) {
    switch (true) {
        default:
            const never: never = x;
        case x.kind === "a":
            x.aProps;
            break;
        case x.kind === "b":
            x.bProps;
            break;
        case x.kind === "c":
            x.cProps;
            break;
    }

    switch (true) {
        default:
            const never: never = x;
        case x.kind === "a": {
            x.aProps;
            break;
        }
        case x.kind === "b": {
            x.bProps;
            break;
        }
        case x.kind === "c": {
            x.cProps;
            break;
        }
    }
    
    switch (true) {
        default:
            x.aProps;
            break;
        case x.kind === "b":
            x.bProps;
            break;
        case x.kind === "c":
            x.cProps;
            break;
    }

    switch (true) {
        default:
            const never: never = x;
        case x.kind === "a":
            x.aProps;
            // fallthrough
        case x.kind === "b":
            x.bProps;
            // fallthrough
        case x.kind === "c":
            x.cProps;
    }

}

class Base {
    basey: string = "";
}

class Derived1 extends Base {
    d: string = "";
}

class Derived2 extends Base {
    d: string = "";
    other: string = "";
}

function classy(base: Base, someDerived: Derived1 | Derived2) {
    switch (true) {
        case base instanceof Derived1:
            base.d
            // fallthrough
        default:
            base.basey
    }

    switch (true) {
        case someDerived instanceof Derived1:
            someDerived.d
            break;
        case someDerived instanceof Derived2:
            someDerived.d
            break
        default:
            const never: never = someDerived;
    }

    switch (true) {
        case someDerived instanceof Derived1:
            someDerived.d
            // fallthrough
        case someDerived instanceof Derived2:
            someDerived.d
            break
        default:
            const never: never = someDerived;
    }

    switch (true) {
        default:
            const never: never = someDerived;
        case someDerived instanceof Derived1:
            someDerived.d;
            someDerived.basey;
            break
        case someDerived instanceof Derived2:
            someDerived.d;
            someDerived.other;
    }

    switch (true) {
        case someDerived instanceof Derived1:
            someDerived.d;
            someDerived.basey;
            break
        default:
            const never: never = someDerived;
        case someDerived instanceof Derived2:
            someDerived.d;
            someDerived.other;
    }

    switch (true) {
        case someDerived instanceof Derived1:
            someDerived.d;
            someDerived.basey;
            break
        case someDerived instanceof Derived2:
            someDerived.d;
            someDerived.other;
        default:
            someDerived.d;
            someDerived.other;
            const never: never = someDerived;
    }

}
```

</p>
</details>

jakebailey:
Since it seems like it's working, I've rebased such that the fix and its test changes are all together for easier viewing.

src/compiler/checker.ts
jakebailey: This comment was previously on the line for `filterType`, but `getNarrowedType` just calls `filterType` like I do here.

jakebailey:
The above shows that this PR is still not fully ready: [Playground Link](https://www.staging-typescript.org/play?ts=5.3.0-pr-55991-37#code/GYVwdgxgLglg9mABMOcAUEEBMCMAuRAI1QBsBTAQzABpFMwsAmA4ucqgSkQG8BYAKERDEAZwDuMKBAAWiNFABOIMlz6DhGiBRFk62fAI1Hh9XAG5DxowHprQgHoB+S1ZPZGF9a6G2Hzr96ECpQA1p4uRlo6egzMEa6mOJ7ewr6ITvFWph6ZxmkZAa5BoeGFwlhkwBQgJFB4uZH6ySmI+f4tMUzNKW0NGsUUYREAvgIRid2tduntbrGTvfzDQA)

RyanCavanaugh:
The effect on `this` is bizarre:
```ts
interface IProps {
    one: boolean;
}

class Foo {
    mine: string = "";

    myMethod(x: IProps) {
        const { one } = x;
        switch (true) {
            case one:
                break;
            default:
                let x = this.mine;
        }
    }
}
```

Andarist:
I'm currently away and just peeking here. The Ryan's case above is broken because:
```ts
clausesType // never
hasDefaultClause // true
before // this
after // never
other // this
```

Since `other` is `this` then `typeNotOther` becomes `never` and the result is a union of that and the `clausesType` (so basically `never | never`). I don't believe this is specific anyhow to `this` type.

jakebailey:
I think I figured it out... Will update shortly. I'm sure it'll be wrong in some all new way, though 😄

jakebailey:
@typescript-bot test top200
@typescript-bot pack this

src/compiler/checker.ts
jakebailey: No longer relevant in the latest iteration.

src/compiler/checker.ts
```diff
@@ -28040,26 +28040,31 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         }
 
         function narrowTypeBySwitchOnTrue(type: Type, switchStatement: SwitchStatement, clauseStart: number, clauseEnd: number): Type {
-            const clauses = switchStatement.caseBlock.clauses.slice(clauseStart, clauseEnd);
-            const clausesType = narrowTypeForTrueClauses(type, clauses);
             const defaultIndex = findIndex(switchStatement.caseBlock.clauses, clause => clause.kind === SyntaxKind.DefaultClause);
             const hasDefaultClause = clauseStart === clauseEnd || (defaultIndex >= clauseStart && defaultIndex < clauseEnd);
```
DanielRosenwasser, line 28044: Reading this in the below code was confusing for me because it's not about the `switch` having a default, it's about the given range including a default.

src/compiler/checker.ts
```diff
@@ -28040,26 +28040,31 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
-                const clausesAfter = switchStatement.caseBlock.clauses.slice(clauseEnd);
 
-                const before = narrowTypeForTrueClauses(type, clausesBefore);
-                const after = narrowTypeForTrueClauses(type, clausesAfter);
-                const other = getUnionType([before, after]);
-
-                const typeNotOther = filterType(type, t => !isTypeSubsetOf(t, other));
+            // First, narrow away all of the cases that preceded this set of cases.
+            const clausesBefore = switchStatement.caseBlock.clauses.slice(0, clauseStart);
+            for (const clause of clausesBefore) {
```
DanielRosenwasser, line 28048: Might as well turn this into a traditional loop

src/compiler/checker.ts
jakebailey: This is just the terminology used above by the `typeof` narrowing. I can change it, but it will be inconsistent.

src/compiler/checker.ts
jakebailey: Yay, time for some off by one bugs

DanielRosenwasser: I mean you don't have to, but it'd be nice to avoid the slices.

jakebailey: I figured someone ask for this in review, so I already had it stashed. It really irritates me that engines can't optimize away the copy when the result is never written to...

src/compiler/checker.ts
```diff
@@ -28043,6 +28039,36 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+                    const clause = switchStatement.caseBlock.clauses[i];
+                    if (clause.kind === SyntaxKind.CaseClause) {
+                        type = narrowType(type, clause.expression, /*assumeTrue*/ false);
+                    }
+                }
+                return type;
+            }
+
+            // Now, narrow based on the cases in this set.
+            const clauses = switchStatement.caseBlock.clauses.slice(clauseStart, clauseEnd);
```
jakebailey, line 28068: I could theoretically try and make this directly iterate instead of slicing and mapping, if that's desirable. We already do a lot of slicing/mapping/etc in narrowing that isn't really needed, so this isn't _new_.

jakebailey: (see also `narrowTypeBySwitchOnDiscriminant` which just slices and doesn't try and be optimal)

src/compiler/checker.ts
```diff
@@ -28043,6 +28039,36 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             return getUnionType(map(clauseWitnesses, text => text ? narrowTypeByTypeName(type, text) : neverType));
         }
 
+        function narrowTypeBySwitchOnTrue(type: Type, switchStatement: SwitchStatement, clauseStart: number, clauseEnd: number): Type {
+            const defaultIndex = findIndex(switchStatement.caseBlock.clauses, clause => clause.kind === SyntaxKind.DefaultClause);
```
DanielRosenwasser, line 28043: The only time `switchStatement` is used is to get `caseBlock` which is only used for `clauses`. Maybe we just pass that in or destructure that out?

DanielRosenwasser: I don't really know what `clauseStart === clauseEnd` means, but you could also avoid the `findIndex` by folding the walk into the first walk, and then you'd be able to avoid the range check, right?

You are basically walking across every `switch/case` per group of clauses (or per clause) which is N^2, right?

jakebailey: Like the other stuff, this is just me doing what the other switch case narrowings do. I can do whatever but it'll look different when it's possible that all of them could be improved.

DanielRosenwasser: Fair enough, we should come back to these at another PR.

DanielRosenwasser (approved):

src/compiler/checker.ts
jakebailey: Yeah, having now touched this, I do think there's a bunch to improve here and elsewhere. It's extremely likely that this just hasn't been touched in a while (when do we ever add new narrowings???)

jakebailey: Do you want me to leave a TODO behind?

jakebailey:
Just to note it, in talking to Daniel, it seems like there's some follow up stuff which would be nice to look at, e.g. if we just CFA'd between cases in switches, then it's likely that all of this code (and probably the other code for switch narrowing) would get to be deleted as we wouldn't have to re-encode that behavior into the checker (which we already do even before `switch (true)` for other narrowings, it's just that `switch (true)` is extra interesting).

PR #55992, opened 2023-10-05
Use xl runners for test/lint

jakebailey:
Not sure if these work, but they seem to be used in other repos in the org, so maybe they exist.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ defaults:
 
 jobs:
   test:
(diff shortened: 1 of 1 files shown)


jakebailey:
Hm, I guess only GitHub themselves have this runner type; I was bamboozled by https://github.com/search?q=%2Fruns-on%3A+ubuntu-latest-%2F+lang%3Ayaml+owner%3Amicrosoft&type=code. But maybe we have other ones available.

PR #56001, opened 2023-10-05
Allow valid contextual keyword completions even when an auto-import by the same name exists

### LINKED ISSUE #55607: Missing `type` keyword completion causes auto-import of `type` symbol
With `@types/node` installed, start typing `type` at the top level in any TS file 😱 

image

andrewbranch:
See code comment and linked issue comments for description.

Fixes #55607

### DIFF
--- a/src/services/completions.ts
+++ b/src/services/completions.ts
@@ -1329,7 +1329,11 @@ function completionInfoFromData(
 
     if (keywordFilters !== KeywordCompletionFilters.None) {
         for (const keywordEntry of getKeywordCompletions(keywordFilters, !insideJsDocTagTypeExpression && isSourceFileJS(sourceFile))) {
-            if (isTypeOnlyLocation && isTypeKeyword(stringToToken(keywordEntry.name)!) || !uniqueNames.has(keywordEntry.name)) {
+            if (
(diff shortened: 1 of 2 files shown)


src/services/completions.ts
```diff
@@ -5829,3 +5833,31 @@ function toUpperCharCode(charCode: number) {
...
+ * may be the beginning of a type alias declaration (keyword completion), or
+ * it may be the beginning of
+ *
+ * ```ts
+ * import { type } from "os";
+ * type() === "Darwin" ? doSomething() : doSomethingElse();
+ * ```
+ */
+function isContextualKeywordInAutoImportableExpressionSpace(keyword: string) {
+    return keyword === "abstract" ||
```
jakebailey, line 5856: I think canonically we'd use a switch for this (faster at runtime I assume).

gabritto (approved):

src/services/completions.ts
andrewbranch: Didn’t we move the visitor API off a `switch` and onto a lookup table for performance reasons?

andrewbranch: Anyway, this list seems small enough and is unlikely to grow

jakebailey: No, that was for a different reason IIRC, the function was so big that it didn't optimize with so many cases handled inside of it

PR #56004, opened 2023-10-06
Improve constraints of conditional types applied to constrained type variables

### LINKED ISSUE #30152: Conditional types are incorrectly narrowed
**TypeScript Version:**  3.1.6, 3.2.1, the current playground (3.3?), and next as of Feb 28 

**Search Terms:**
conditional types are incorrectly 

**Code**
```ts
interface A { foo(); }
interface B { bar(); }

function test1<T extends A>(x: T, y: T extends B ? number : string) {
    if (typeof y == 'string') {
        y;
    } else {
        y; // never ?
    }
    const newY: string | number = y;
    newY;  // just string
}

function test2<T extends A>(x: T, y: T extends B ? string : number) {
    if (typeof y == 'string') {
        y; // never ?
    } else {
        y; 
    }
    const newY: string | number = y;
    newY;  // just number 
}
```

**Expected behavior:**
`T extends B ? string : number` should either be left unchanged, or rounded up to `string|number`: I think the issue stems from incorrect inference that `T extends B` is false given `T extends A` (while they're just unrelated interfaces that have a non-empty intersection). The test case below is as far as I've managed to reduce the problem.

**Actual behavior:**
The `T extends A` constraint seems to make TS guess `T extends B` is always false, and so the `a?b:c` type behaves as `c`. 

**Playground Link:** ([playground](https://www.typescriptlang.org/play/index.html#src=interface%20A%20%7B%20foo()%3B%20%7D%0D%0Ainterface%20B%20%7B%20bar()%3B%20%7D%0D%0A%0D%0Afunction%20test1%3CT%20extends%20A%3E(x%3A%20T%2C%20y%3A%20T%20extends%20B%20%3F%20number%20%3A%20string)%20%7B%0D%0A%20%20%20%20if%20(typeof%20y%20%3D%3D%20'string')%20%7B%0D%0A%20%20%20%20%20%20%20%20y%3B%0D%0A%20%20%20%20%7D%20else%20%7B%0D%0A%20%20%20%20%20%20%20%20y%3B%20%2F%2F%20never%20%3F%0D%0A%20%20%20%20%7D%0D%0A%20%20%20%20%0D%0A%20%20%20%20const%20newY%3A%20string%20%7C%20number%20%3D%20y%3B%0D%0A%20%20%20%20newY%3B%20%20%2F%2F%20just%20string%0D%0A%7D%0D%0A%0D%0Afunction%20test2%3CT%20extends%20A%3E(x%3A%20T%2C%20y%3A%20T%20extends%20B%20%3F%20string%20%3A%20number)%20%7B%0D%0A%20%20%20%20if%20(typeof%20y%20%3D%3D%20'string')%20%7B%0D%0A%20%20%20%20%20%20%20%20y%3B%20%2F%2F%20never%20%3F%0D%0A%20%20%20%20%7D%20else%20%7B%0D%0A%20%20%20%20%20%20%20%20y%3B%20%0D%0A%20%20%20%20%7D%0D%0A%20%20%20%20const%20newY%3A%20string%20%7C%20number%20%3D%20y%3B%0D%0A%20%20%20%20newY%3B%20%20%2F%2F%20just%20number%20%0D%0A%7D%0D%0A%0D%0Aclass%20test3%3CT%20extends%20A%3E%20%7B%0D%0A%20%20%20%20test3_f(x%3A%20T%2C%20y%3A%20T%20extends%20B%20%3F%20string%20%3A%20number)%20%7B%0D%0A%20%20%20%20%20%20%20%20if%20(typeof%20y%20%3D%3D%20'string')%20%7B%0D%0A%20%20%20%20%20%20%20%20%20%20%20%20y%3B%20%2F%2F%20never%20%3F%0D%0A%20%20%20%20%20%20%20%20%7D%20else%20%7B%0D%0A%20%20%20%20%20%20%20%20%20%20%20%20y%3B%20%0D%0A%20%20%20%20%20%20%20%20%7D%0D%0A%20%20%20%20%20%20%20%20const%20newY%3A%20string%20%7C%20number%20%3D%20y%3B%0D%0A%20%20%20%20%20%20%20%20newY%3B%20%20%2F%2F%20just%20number%20%0D%0A%20%20%20%20%7D%0D%0A%7D))

**Related Issues:**
https://github.com/Microsoft/TypeScript/issues/29939 looks slightly similar, but I don't see the same constraints when playing around with my example, so I'm not sure it's the same.

ahejlsberg:
With this PR we compute more accurate constraints for distributive conditional types applied to constrained type variables. For example:

```ts
type IsArray<T> = T extends any[] ? true : false;

function f1<U extends object>(x: IsArray<U>) {
    let t: true = x;   // Error
    let f: false = x;  // Error, but previously wasn't
}
```

Previously, we'd obtain the constraint of `IsArray<U>` by applying the distributive conditional type to the constraint of `U`, which resolves to `false` (because `object` doesn't extend `any[]`). But really, it ought to be `boolean` (i.e. `true | false`) since clearly something constrained to `object` both could or could not be an array.

A type variable represents any possible type within its constraint. So, given a type parameter `T` with the constraint `C`, the constraint of `T extends X ? A : B` is

* `A` when `C` is known to always extend `X`,
* `B` when `C` is known to never extend `X`, or
* `A | B` when `C` possibly extends `X`.

We previously didn't consider the third possibility. With this PR, we choose the third outcome when `C` is not assignable to `X`, but `X` (or, if `X` is a union type, some constituent of `X`) is assignable to `C`.

Fixes #30152.


ahejlsberg:
@typescript-bot test top100
@typescript-bot run dt

gabritto:
@typescript-bot pack this

ahejlsberg:
@typescript-bot test top100
@typescript-bot run dt

ahejlsberg:
Latest commit changes checking for the "possibly extends" case to use a reverse assignability check. This check is not quite as permissible as a comparability check, meaning we'll less often hit the "possibly extends" case where the constraint is a union. Still, since constraint types are typically close-to-top types (like `object` or `{}`), it isn't likely to matter. And, certainly, performance with the reverse assignability test is much more acceptable.

src/compiler/checker.ts
```diff
@@ -18323,7 +18323,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 // then no instantiations will be and we can just return the false branch type.
                 if (!(inferredExtendsType.flags & TypeFlags.AnyOrUnknown) && (checkType.flags & TypeFlags.Any || !isTypeAssignableTo(getPermissiveInstantiation(checkType), getPermissiveInstantiation(inferredExtendsType)))) {
                     // Return union of trueType and falseType for 'any' since it matches anything
-                    if (checkType.flags & TypeFlags.Any) {
+                    if (checkType.flags & TypeFlags.Any || forConstraint && isTypeAssignableTo(getPermissiveInstantiation(inferredExtendsType), getPermissiveInstantiation(checkType))) {
```
weswigham, line 18326: While this is probably a bit better than the often-incorrect result we give now, so maybe we're just fine with that being that, isn't a check in this style kind of fundamentally flawed? For example, a `type Cond<T extends {a: string} | {b: string}> = T extends {c: string} ? true : false`. `T`'s constraint isn't related in any way to the check type of `{c: string}`, yet the constraint should probably still be the union of both branches, as an instantiation of `T` might satisfy the check type, or might not, while here we're still skipping the `true` branch type just because `{a: string}` isn't assignable to `{c: string}`. And comparability wouldn't have been any better. As @andrewbranch said - the only case where we can definitively say only one branch is taken would seem to require checking for an empty intersection between the check-type-constraint and the extends type (and, as you mentioned, ideally tighter empty intersection logic so that has higher accuracy).

How different are the results it if we just replace this assignability check with a `intersectTypes(getPermissiveInstantiation(inferredExtendsType), getPermissiveInstantiation(checkType)) !== neverType`? Are the bad changes any worse?

weswigham (approved):

ahejlsberg:
@typescript-bot test top100
@typescript-bot run dt

MichaelMitchell-at:
I don't know if this is intentional or a regression but the following code used to typecheck, but no longer does with this PR:
```ts
export type DeepReadonly<T> = T extends boolean | string | number | null | undefined
    ? T
    : T extends ReadonlyArray<infer Item>
        ? ReadonlyArray<DeepReadonly<Item>>
        : {readonly [K in keyof T]: DeepReadonly<T[K]>}

type Shape = {x: number};

export function getKey<T extends Partial<Shape>, K extends keyof Shape>(shape: DeepReadonly<T>, key: K) {
    return shape[key];
}
```

[Playground with TS 5.2](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBDAnmYcAixhgErAIYAmEAdgDaIA8AKgHxwC8cVcoMwxBAznAEYQSl8xOAB84nGFACWxAOai4xAK4BbHsCgLlpUgqUdgAMxnACAKDiW4AfiYWrALiYsQbDt1yES5AIJQoeJQyhhpwAJJsKjT2Vla2nkRkiH4BlBhYCd6UEcBR0bGxTgDeUPiJ5HAA2gDScDJwANbAiBCGTAC6Tuk4ZVnUNe00AL5mZkgocADKABZ4E4xFIE7KahpDANyjoJCwcIb6AMYwUiRwssAw1c3ULm5ccAAKeLBSeKQUM3PANAA0cLWsdj3JotNqfFA0AAUnFmKC6mB6XiS1F+jWaTmqAEo4EUYnBSjAlFBhDCvpUQe1NkMgA)
[Playground with this PR](https://www.staging-typescript.org/play?ts=5.3.0-pr-56004-10#code/KYDwDg9gTgLgBDAnmYcAixhgErAIYAmEAdgDaIA8AKgHxwC8cVcoMwxBAznAEYQSl8xOAB84nGFACWxAOai4xAK4BbHsCgLlpUgqUdgAMxnACAKDiW4AfiYWrALiYsQbDt1yES5AIJQoeJQyhhpwAJJsKjT2Vla2nkRkiH4BlBhYCd6UEcBR0bGxTgDeUPiJ5HAA2gDScDJwANbAiBCGTAC6Tuk4ZVnUNe00AL5mZkgocADKABZ4E4xFIE7KahpDANyjoJCwcIb6AMYwUiRwssAw1c3ULm5ccAAKeLBSeKQUM3PANAA0cLWsdj3JotNqfFA0AAUnFmKC6mB6XiS1F+jWaTmqAEo4EUYnBSjAlFBhDCvpUQe1NkMgA)

ahejlsberg:
@typescript-bot pack this

ahejlsberg:
@MichaelMitchell-at That was an earlier copy of the PR. It works with [a current build](https://www.staging-typescript.org/play?ts=5.3.0-pr-56004-36#code/KYDwDg9gTgLgBDAnmYcAixhgErAIYAmEAdgDaIA8AKgHxwC8cVcoMwxBAznAEYQSl8xOAB84nGFACWxAOai4xAK4BbHsCgLlpUgqUdgAMxnACAKDiW4AfiYWrALiYsQbDt1yES5AIJQoeJQyhhpwAJJsKjT2Vla2nkRkiH4BlBhYCd6UEcBR0bGxTgDeUPiJ5HAA2gDScDJwANbAiBCGTAC6Tuk4ZVnUNe00AL5mZkgocADKABZ4E4xFIE7KahpDANyjoJCwcIb6AMYwUiRwssAw1c3ULm5ccAAKeLBSeKQUM3PANAA0cLWsdj3JotNqfFA0AAUnFmKC6mB6XiS1F+jWaTmqAEo4EUYnBSjAlFBhDCvpUQe1NkMgA).

MichaelMitchell-at:
I tried the current build in our code base and ran into a couple of errors that I'm having trouble creating a minimal reproduction that I can share. In general they seem to arise when generics are involved together with complex conditional types. I wonder if I'm having trouble producing a minimal repro because in the actual code we're hitting one of the depth limits that exists in the type checker.


Unrelated, but just mentioning while I have your attention with the hope y'all will keep in mind, is on nightly I'm encountering  a new `The inferred type of this node exceeds the maximum length the compiler will serialize` error. It's not really actionable on our side; the object is created by composing a bunch of schema objects from our in-house [Zod](https://zod.dev/?id=basic-usage)-like library. One of the reasons to use the schema library is to avoid needing to write out the type by hand. It would probably only help us temporarily if the max length were increased; A more long-term approach might be some ability to give users more control over when type aliases are used/expanded in types when serialized/displayed.

voxpelli:
As an FYI in case anyone follows this issue, it turned out that this caused a regression in `type-fest` and I'm still debugging why and/or how to fix, but want to mention here in the meanwhile, might file an issue later: https://github.com/sindresorhus/type-fest/issues/784

src/compiler/checker.ts
```diff
@@ -22303,25 +22307,23 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
                 }
+                // conditionals aren't related to one another via distributive constraint as it is much too inaccurate and allows way
+                // more assignments than are desirable (since it maps the source check type to its constraint, it loses information).
+                const distributiveConstraint = !(targetFlags & TypeFlags.Conditional) && hasNonCircularBaseConstraint(source) ? getConstraintOfDistributiveConditionalType(source as ConditionalType) : undefined;
+                if (distributiveConstraint) {
+                    resetErrorInfo(saveErrorInfo);
+                    if (result = isRelatedTo(distributiveConstraint, target, RecursionFlags.Source, reportErrors)) {
+                        return result;
+                    }
+                }
```
LukeAbby, line 22326: y

PR #47388, opened 2022-01-11
Better reuse of package.json cache, module resolution cache, and package.json auto import filter

andrewbranch:
I made a test project that contained a lot of dependencies with export maps to ensure the new code paths were being hit. The slowdown from main to #47092 can largely be explained by the fact that we’re finding more things to auto-import, as the baseline case completely ignored those export maps. The cache reuse pays off in `collectAutoImports`, where previously we had to read package.jsons that have already been read and cached. Note that `getExportInfoMap` and `collectAutoImports` are components of `completionInfo`, which is the “bottom line” perf number for the server-side work of getting completions.

|                   | dev.20220109 | #47092 | This PR |
|-------------------|-------------:|--------:|-------:|
| createAutoImportProviderProgram | 354 ms | 395 ms | 408 ms |
| AutoImportProvider files | 6 | 30 | 30 |
| getExportInfoMap | 34 ms | 48 ms | 44 ms |
| collectAutoImports | 89 ms | 98 ms | 76 ms |
| completionInfo | 246 ms | 245 ms | 230 ms |
| completion entries | 3075 | 3149 | 3149 |

PR #56016, opened 2023-10-06
noDts project resolutions verification and updates for incremental behaviour

sheetalkamat:
When resolving module specifier for finding implementation file, we were using module resolution cache which means that it was possible to reuse existing module resolution and add more failed lookup locations to it. Which means that when we release that resolution as part of program update/close, we will try to release more locations than being watched.
Fixed this by marking module resolution cache as read-only after construction of program, which indicates module resolution algorithm to not update the cache or resolution. (instead of freezing internal maps, went ahead with at usage mode approach to avoid having to deal with invalidation logic eg for package json info before program construction) 1b6c9a8

Also updated some of the other `noDts` project related things:
- Project gets updates root files only if the requested file changes 2d40a69
- Add `isBackgroundProject` so we can ensure all the updates where we need to skip things for background project are handled correctly fe7f068
- Earlier if the js file was not part of project it would use it only if it was at any point used and cached in services. Instead of that, now it's always added to project to search for implementation from. 0efe4ee

Added coverage for some of the missing cases like 
- when module specifier resolved from file is not already in the file: d552636
- file is added to project after project construction for the first time based on imports present in existing file 43720b9
- repeated query shouldn't change project: 5ef80a9
- resolution is reused that was done from some sibling folder 8538100
- incremental verification for `resolveSingleModuleNameWithoutWatching` to ensure it does not impact state: a49de0e


jakebailey:
I'm not entirely sure how these types are used in public, but is the readonly flag something that should be exported so that external consumers know to use the API in the right way, or is it unlikely to matter?

sheetalkamat:
I think we only care about resolution cache we created and are using so i am not sure if it matters. But also while investigating i was wondering `Project::getModuleResolutionCache` should have been internal and i was going to mark it that way and forgot. Will do that.

tests/baselines/reference/api/typescript.d.ts
```diff
@@ -3354,7 +3354,6 @@ declare namespace ts {
             readFile(fileName: string): string | undefined;
             writeFile(fileName: string, content: string): void;
             fileExists(file: string): boolean;
-            getModuleResolutionCache(): ModuleResolutionCache | undefined;
```
jakebailey, line 3357: Hm, https://github.com/marko-js/language-server/blob/6829c580d16eb1c6128c90a73af36fa4efb64d08/packages/language-server/src/ts-plugin/index.ts#L59 uses it, but that may be the only example I found, so maybe it's fine.

sheetalkamat: They shouldnt have been. Gitlens went in loops but couldnt figure out when and why this was exposed but from looks of it, it should not be exposed at all

jakebailey: It was added in https://github.com/microsoft/TypeScript/pull/47388, it seems. @andrewbranch

jakebailey: Wait, but it's internal there. Hm, one sec.

src/compiler/moduleNameResolver.ts
```diff
@@ -257,13 +259,24 @@ function createResolvedModuleWithFailedLookupLocations(
...
     resultFromCache: ResolvedModuleWithFailedLookupLocations | undefined,
+    cache: ModuleResolutionCache | NonRelativeModuleNameResolutionCache | undefined,
     legacyResult?: string,
 ): ResolvedModuleWithFailedLookupLocations {
     if (resultFromCache) {
-        resultFromCache.failedLookupLocations = updateResolutionField(resultFromCache.failedLookupLocations, failedLookupLocations);
-        resultFromCache.affectingLocations = updateResolutionField(resultFromCache.affectingLocations, affectingLocations);
-        resultFromCache.resolutionDiagnostics = updateResolutionField(resultFromCache.resolutionDiagnostics, diagnostics);
-        return resultFromCache;
+        if (!cache?.isReadonly) {
```
andrewbranch, line 266: My only concern is it seems somewhat easy to miss this check if someone were adding new code that interacts with the cache, and it seems like there still are no compile-time or runtime checks preventing you from mutating the contents at a time when it could cause problems. Would it be worth changing `ResolvedModuleWithFailedLookupLocations` etc. to have readonly properties that can only be changed by a method on the cache, which could check `isReadonly` to ensure it’s not being violated?

tests/baselines/reference/api/typescript.d.ts
andrewbranch: It’s internal on `LanguageServiceHost`, but not on `Project` 😕

jakebailey: Ah, yeah, I just discovered that. I guess it should just be dropped?

jakebailey: (To be clear I'm not trying to actually blame anyone, moreso just trying to do the forensics so we know if there was a reason to have it 😅)

src/compiler/moduleNameResolver.ts
sheetalkamat: i can do that, but frankly I dont see point in that since I think we want this method to be that one. Which ensures the resolutionWithFailedLookupLocation is created correctly.

andrewbranch (approved):

Issue #56018, opened 2023-10-07
Incorrect inference of mapped substitution type in conditional involving a union of arrays (regression)


geoffreytools:
### 🔎 Search Terms

"inference", "conditional type", "substitution type", "mapped type", "type constraint", "array", "union type"

### 🕗 Version & Regression Information

This changed between versions 5.0.4 and 5.1.6

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.1.6#code/C4TwDgpgBAShB2BDAthAJlAvFAThRaA9vADYhQAUA3lANoDWAXFAAo6GQ6gDSEIAuswDOwHAEt4AcygBfKAB8oAV3hoIAMwnoAlLX4AofaEhQAYoUIAeACpQIAD2AI0Q3PiKlyFNhwhcQvOSKKmqa8Dp6CrAIKOgAfFj6UMlQtg5Oqq5wSKgYAPxQAOIQwIFClnAAjkpieGg2cQnMVTV1DYbG0MWlfOUwdo7OWTG5Cdg0tNxQElD0fITqsIKz84swk-yyhgD028pCiJIQRuDQAIJYZhaWtADkiLd5-Am7dABEiG8GnVAAQpfmKy0GiIPLMW4AI1usmeUFetA+XyAA

### 💻 Code

```ts
type Renamed = readonly ({ [k: PropertyKey]: string } | undefined)[]

type Foo<T extends readonly (PropertyKey | undefined)[] | Renamed> =
    T extends Renamed ? GetKeys<Required<T>> : Required<T>
//                              ~~~~~~~~~~
// Type 'Required<T>' does not satisfy the constraint 'Renamed'

type GetKeys<R extends Renamed> = { [K in keyof R]: keyof R[K] }

// usage
type A = Foo<['a'?]> // ["a"]
type B = Foo<[{ a?: 'b' }]> // ["a"]
```

### 🙁 Actual behavior

TS fails to see that `Required<T extends Renamed>` extends `Renamed`.

### 🙂 Expected behavior

It should compile.

### Additional information about the issue

Substitute `readonly (PropertyKey | undefined)[]` with something that is not an array and the problem disappears. 

I get the same error when I do `GetKeys<Id<T>>` with `type Id<T> = { [K in keyof T]: T[K] }`.

`Extract<Required<T>, Required<Renamed>>` silences the error, but `Extract<Required<T>, Renamed>` does not.

Andarist:
[...]

PR #56027, opened 2023-10-08
Fixed a relationship check when apparent type of mapped type expands to a union

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/56018

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -22373,7 +22373,11 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 }
                 const sourceIsPrimitive = !!(sourceFlags & TypeFlags.Primitive);
                 if (relation !== identityRelation) {
+                    const originalSource = source;
                     source = getApparentType(source);
+                    if (!(originalSource.flags & TypeFlags.Union) && source.flags & TypeFlags.Union && (result = unionOrIntersectionRelatedTo(source, target, reportErrors, intersectionState))) {
+                        return result;
+                    }
                     sourceFlags = source.flags;
                 }
                 else if (isGenericMappedType(source)) {
--- a/tests/baselines/reference/intersectionSatisfiesConstraint.types
+++ b/tests/baselines/reference/intersectionSatisfiesConstraint.types
@@ -29,9 +29,9 @@ const myFirstFunction = <T extends FirstInterface | SecondInterface>(param1: T) 
 >3 : 3
 
     mySecondFunction(newParam)
->mySecondFunction(newParam) : { commonProperty: number; otherProperty: number; }
+>mySecondFunction(newParam) : T & { otherProperty: number; }
 >mySecondFunction : <T extends { commonProperty: number; otherProperty: number; }>(newParam: T) => T
->newParam : (FirstInterface | SecondInterface) & { otherProperty: number; }
+>newParam : T & { otherProperty: number; }
 }
 
 const mySecondFunction = <T extends { commonProperty: number, otherProperty: number }>(newParam: T) => {
--- a/tests/baselines/reference/keyofAndIndexedAccessErrors.errors.txt
+++ b/tests/baselines/reference/keyofAndIndexedAccessErrors.errors.txt
@@ -42,8 +42,10 @@ keyofAndIndexedAccessErrors.ts(103,9): error TS2322: Type 'Extract<keyof T, stri
   'Extract<keyof T, string>' is assignable to the constraint of type 'K', but 'K' could be instantiated with a different subtype of constraint 'string'.
     Type 'string & keyof T' is not assignable to type 'K'.
       'string & keyof T' is assignable to the constraint of type 'K', but 'K' could be instantiated with a different subtype of constraint 'string'.
-        Type 'string' is not assignable to type 'K'.
-          'string' is assignable to the constraint of type 'K', but 'K' could be instantiated with a different subtype of constraint 'string'.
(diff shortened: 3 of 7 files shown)


src/compiler/checker.ts
```diff
@@ -22373,7 +22373,11 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 }
                 const sourceIsPrimitive = !!(sourceFlags & TypeFlags.Primitive);
                 if (relation !== identityRelation) {
+                    const originalSource = source;
                     source = getApparentType(source);
+                    if (!(originalSource.flags & TypeFlags.Union) && source.flags & TypeFlags.Union && (result = unionOrIntersectionRelatedTo(source, target, reportErrors, intersectionState))) {
```
Andarist, line 22378: I'm not fond of how this check is so targeted. I also tried to make the least amount of work required here so I chosen to use `unionOrIntersectionRelatedTo` but perhaps I should go full circle and use `isRelatedTo`?

I think it's pretty unlikely for `getApparentType` to end up with some other type of a type that would require us to go full circle but I'm definitely not 100% sure of that.

Maybe a better fix would be to obtain the `getApparentTypeOfMappedType` sooner in this function. The only thing that it does is that it instantiates mapped types with `everyType(constraint, isArrayOrTupleType)`. And that's the thing that might "expand" it to a union.

weswigham: `unionOrIntersectionRelatedTo` is going to skip the cache (it basically only exists to skip the cache for small unions and intersections), so yeah, probably go full circle and reenter `isRelatedTo`, so it can make that cache or no-cache call. Honestly, you can probably generalize this to `const oldSource = source; source = getApparentType(source); if (oldSource !== source) return isRelatedTo(source, target, reportErrors, intersectionState)` - probably. There may be a perf reason to only limit the reentry to union apparent types, but until that's shown, the even more general approach is probably preferable.

weswigham (changes requested):

src/compiler/checker.ts
Andarist: None of those diffs worked:
<details>
<summary>diff 1</summary>

```diff
diff --git a/src/compiler/checker.ts b/src/compiler/checker.ts
index 3573d5b466..e63d2810cb 100644
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -22447,12 +22447,10 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 }
                 const sourceIsPrimitive = !!(sourceFlags & TypeFlags.Primitive);
                 if (relation !== identityRelation) {
-                    const originalSource = source;
-                    source = getApparentType(source);
-                    if (!(originalSource.flags & TypeFlags.Union) && source.flags & TypeFlags.Union && (result = unionOrIntersectionRelatedTo(source, target, reportErrors, intersectionState))) {
-                        return result;
+                    const apparentSource = getApparentType(source);
+                    if (apparentSource !== source) {
+                        return isRelatedTo(apparentSource, target, RecursionFlags.Both, reportErrors);
                     }
-                    sourceFlags = source.flags;
                 }
                 else if (isGenericMappedType(source)) {
                     return Ternary.False;
```
</details>

<details>
<summary>diff 2</summary>

```diff
diff --git a/src/compiler/checker.ts b/src/compiler/checker.ts
index 3573d5b466..8670df0345 100644
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -22447,12 +22447,10 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 }
                 const sourceIsPrimitive = !!(sourceFlags & TypeFlags.Primitive);
                 if (relation !== identityRelation) {
-                    const originalSource = source;
-                    source = getApparentType(source);
-                    if (!(originalSource.flags & TypeFlags.Union) && source.flags & TypeFlags.Union && (result = unionOrIntersectionRelatedTo(source, target, reportErrors, intersectionState))) {
+                    const apparentSource = getApparentType(source);
+                    if (apparentSource !== source && (result = isRelatedTo(apparentSource, target, RecursionFlags.Both, reportErrors))) {
                         return result;
                     }
-                    sourceFlags = source.flags;
                 }
                 else if (isGenericMappedType(source)) {
                     return Ternary.False;
```
</details>

<details>
<summary>diff 3</summary>

```diff
diff --git a/src/compiler/checker.ts b/src/compiler/checker.ts
index 3573d5b466..db0c7abbc6 100644
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -22447,11 +22447,11 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 }
                 const sourceIsPrimitive = !!(sourceFlags & TypeFlags.Primitive);
                 if (relation !== identityRelation) {
-                    const originalSource = source;
-                    source = getApparentType(source);
-                    if (!(originalSource.flags & TypeFlags.Union) && source.flags & TypeFlags.Union && (result = unionOrIntersectionRelatedTo(source, target, reportErrors, intersectionState))) {
+                    const apparentSource = getApparentType(source);
+                    if (apparentSource !== source && (result = isRelatedTo(apparentSource, target, RecursionFlags.Both, reportErrors))) {
                         return result;
                     }
+                    source = apparentSource;
                     sourceFlags = source.flags;
                 }
                 else if (isGenericMappedType(source)) {
```
</details>

I'll have to look later at what actually happens when `isRelatedTo` gets called here though since I just blindly tried to call it here ;p

PR #56034, opened 2023-10-08
fix(55994): Type-check Import Attributes in static imports

### LINKED ISSUE #55994: Type-check Import Attributes in static imports
### 🔍 Search Terms

import attributes, import assertions

### ✅ Viability Checklist

### ⭐ Suggestion

[TypeScript 5.3 Beta](https://devblogs.microsoft.com/typescript/announcing-typescript-5-3-beta/) supports the latest updates to the [import attributes](https://github.com/tc39/proposal-import-attributes) proposal, which is great to see!

It's possible to augment the global `ImportAttributes` type with host-specific attributes. However, auto-complete and type-checking of import attributes seems to only work with dynamic `import()`. It would be nice if this could also work for attributes in static declarations.

### 📃 Motivating Example

https://www.typescriptlang.org/play?target=99&ts=5.3.0-dev.20231002#code/CYUwxgNghgTiAEBzCB7ARlC8DeBYAUPPAJYB2ALiDAGZRgLwCSAtgA4ozkCC55MxaAK6UAzjgJEi5AJ6sQALngAiAFYiUpJRPgBfAnvwFibDuXgAqeFDGkx1GCmbLqKFEvgB3YuQAWOeDJyikqkKOQAtGoa7joA3AQAbijEwPC28Yb4xuycABRKLm4ANOKEnt4+inhlUrIKyqERUZraBjoAlBkEQA 

```typescript
declare global {
    interface ImportAttributes {
        type: "json";
    }
}

import * as ns from "foo" with { type: "not-json" }; // no error
void ns;

import("foo", {
    with: {
        type: "not-json", // Type '"not-json"' is not assignable to type '"json"'. (2322)
    },
});
```

### 💻 Use Cases

https://github.com/tc39/proposal-import-attributes#motivation

In addition, we have an internal use case in Bloomberg to use Import Attributes as an annotation to achieve lazy-loading by deferring module evaluation. This is an early implementation of the [TC39 proposal for Deferring Module Evaluation](https://github.com/tc39/proposal-defer-import-eval) that annotates the static import.

It would be a better developer experience if auto-complete was available for Import Attributes in the static form.

It is possible to write a static import declaration using unknown/invalid attributes.

None.

a-tarasyuk:
Fixes #55994

### DIFF
--- a/src/compiler/types.ts
+++ b/src/compiler/types.ts
@@ -5971,6 +5971,7 @@ export const enum InternalSymbolName {
     Default = "default", // Default export symbol (technically not wholly internal, but included here for usability)
     This = "this",
     InstantiationExpression = "__instantiationExpression", // Instantiation expressions
+    ImportAttributes = "__importAttributes",
 }
 
 /**
--- a/tests/baselines/reference/completionListForImportAttributes.baseline
+++ b/tests/baselines/reference/completionListForImportAttributes.baseline
@@ -6,6 +6,7 @@
 //         "resolution-mode": "import"
 //     }
 // }
+// const str = "hello";
 // 
 // import * as t1 from "./a" with {  };
 //                                  ^
@@ -14,7 +15,7 @@
 // | (property) ImportAttributes.type: "json"
 // | ----------------------------------------------------------------------
 // import * as t2 from "./a" with { type: "" };
-//                                         
+//                                         ^
 // | ----------------------------------------------------------------------
 // | json
 // | ----------------------------------------------------------------------
@@ -23,12 +24,18 @@
 // | ----------------------------------------------------------------------
 // | (property) ImportAttributes["resolution-mode"]: "import"
 // | ----------------------------------------------------------------------
+// import * as t4 from "./a" with { type:  };
+//                                        ^
+// | ----------------------------------------------------------------------
+// | "json"
+// | (property) ImportAttributes["resolution-mode"]: "import"
+// | ----------------------------------------------------------------------
 
 [
   {
     "marker": {
       "fileName": "/tests/cases/fourslash/./b.ts",
-      "position": 150,
+      "position": 171,
       "name": "1"
     },
     "item": {
@@ -145,7 +152,7 @@
   {
     "marker": {
       "fileName": "/tests/cases/fourslash/./b.ts",
-      "position": 194,
+      "position": 215,
       "name": "2"
     },
     "item": {
@@ -153,7 +160,7 @@
       "isMemberCompletion": false,
       "isNewIdentifierLocation": false,
       "optionalReplacementSpan": {
-        "start": 194,
+        "start": 215,
         "length": 0
       },
       "entries": [
@@ -163,7 +170,7 @@
           "kind": "string",
           "sortText": "11",
           "replacementSpan": {
-            "start": 194,
+            "start": 215,
             "length": 0
           },
           "displayParts": [
@@ -179,7 +186,7 @@
   {
     "marker": {
       "fileName": "/tests/cases/fourslash/./b.ts",
-      "position": 246,
+      "position": 267,
       "name": "3"
     },
     "item": {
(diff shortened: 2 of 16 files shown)


src/compiler/types.ts
```diff
@@ -3650,14 +3651,14 @@ export type AssertClause = ImportAttributes;
 
 export type ImportAttributeName = Identifier | StringLiteral;
 
-export interface ImportAttribute extends Node {
+export interface ImportAttribute extends Declaration {
```
jakebailey, line 3654: Making import attributes extend `Declaration` feels wrong; what was the thinking behind these?

a-tarasyuk: @jakebailey This is one of my questions, you beat me with your review ;). Is it okay to handle types with TransientSymbol (and them in createAnonymousType), especially in cases like completion? I was thinking about using contextual type to get a list of available attributes. Alternatively, is it okay to simply extend `getTypeAtLocation` to handle that case for services?

jakebailey:
LGTM, but I would like a second opinion.

@typescript-bot test top200
@typescript-bot run dt

@typescript-bot pack this

tests/baselines/reference/completionListForImportAttributes.baseline
```diff
@@ -0,0 +1,247 @@
...
+//                                  ^
+// | ----------------------------------------------------------------------
+// | (property) ImportAttributes["resolution-mode"]: "import"
+// | (property) ImportAttributes.type: "json"
+// | ----------------------------------------------------------------------
+// import * as t2 from "./a" with { type: "" };
+//                                         
+// | ----------------------------------------------------------------------
+// | json
+// | ----------------------------------------------------------------------
```
andrewbranch, line 20: The `^` is missing from the baseline generator 🤔

tests/baselines/reference/importAssertionNonstring.errors.txt
```diff
@@ -16,14 +32,26 @@ mod.mts(11,66): error TS2837: Import assertion values must be string literal exp
...
+!!! error TS2322:   Property 'field' is incompatible with index signature.
+!!! error TS2322:     Type 'string[]' is not assignable to type 'string'.
                                                        ~~~~~
 !!! error TS2837: Import assertion values must be string literal expressions.
     
     import * as thing5 from "./mod.mjs" assert {field: { a: 0 }};
+                                        ~~~~~~~~~~~~~~~~~~~~~~~~
+!!! error TS2322: Type '{ field: { a: number; }; }' is not assignable to type 'ImportAttributes'.
+!!! error TS2322:   Property 'field' is incompatible with index signature.
+!!! error TS2322:     Type '{ a: number; }' is not assignable to type 'string'.
```
andrewbranch, line 54: It’s worth considering the pros and cons of applying the checking to import assertions as well as import attributes. My instinct was to suggest leaving them alone with no checking, but I don’t know that I want to give people a reason _not_ to migrate to import attributes.

tests/cases/fourslash/completionListForImportAttributes.ts
```diff
@@ -0,0 +1,20 @@
...
+////        type: "json",
+////        "resolution-mode": "import"
+////    }
+////}
+////
+////import * as t1 from "./a" with { /*1*/ };
+////import * as t2 from "./a" with { type: "/*2*/" };
+////import * as t3 from "./a" with { type: "json", /*3*/ };
+
+verify.baselineCompletions();
```
andrewbranch, line 20: It might be nice to show that locals/globals are excluded when the contextual type allows any `string`, since values have to be literals, not references:

```ts
declare global {
  interface ImportAttributes {
    type: string
  }
}

const str = "hello";
import * as t4 from "./a" with { type: /* exclude 'str' */ };
```

andrewbranch (approved):

tests/baselines/reference/importAssertionNonstring.errors.txt
a-tarasyuk: @andrewbranch Thanks for the feedback. Is it okay to skip checking _import assertions_, or should I change `getGlobalImportAttributesType` to handle both the deprecated `ImportAssertions` and `ImportAttributes`?

andrewbranch: I don’t think the type of import attributes should be affected by the `ImportAssertions` type—I was just wondering whether new errors should appear at all on import _assertions_. I’m not sure, so I raised this comment for other team members to consider while reviewing. I’m not suggesting any change at this point 👍

a-tarasyuk: oke, got it.

andrewbranch (approved):

tests/baselines/reference/importAssertionNonstring.errors.txt
andrewbranch: Conclusion from design meeting was this is fine the way it is.

a-tarasyuk: @andrewbranch Oke, thanks for the feedback.

jakebailey (approved):

andrewbranch:
Other note from the design meeting: was it intentional that we didn't mark the `ImportAssertions` global type as `@deprecated`? We think perhaps we just forgot.

andrewbranch:
Thanks @a-tarasyuk!

Issue #56030, opened 2023-10-08
Switch not narrowed when there is an extra pair of braces. 


JeanMeche:
### 🔎 Search Terms

* switch
* narrowing

### 🕗 Version & Regression Information

This is not a regression, it never worked. 

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.2.2#code/JYOwLgpgTgZghgYwgAgEJwM4oN4ChnJgCeADhAFzIDkMA9rVcgD7UBGcUVuAvrrqJFiIUAMXrIIAD0ggAJhjSYc+QqQrU6DANwrNlDGCigA5jt79w0eEkVQJ0iHIXosyPAWJlKVdpx0FfShAAVwBbVmgzPhhgkAQwYFoQZGMIMAA1AAoASkoxWmZbNxUCKDTgqGSQgBtqgEIo3AQkg2QAN2QAXhS0rOydfhhMzLaAOk8IbK7O7p8OKin3dvG1Hj5cAHoN5AB1WigAawxcDAB3YDAEAAtkEZWyReQt5AAeGc6VBCU2efIS5d8-gIyFYZTgBwGBC+rho9Cof2By00QICYIha022wAIrQIAoQLQwMhTvsDidzpcbsMxhNso9nm8Zp9vnNOAjgWNAf9QRBwZDkNCULCGOyCGNkdy0WYgA

### 💻 Code

```ts
interface Base {
  type: 'foo' | 'bar'
}

interface Foo extends Base {
  type: 'foo';
  foo: string;
}

interface Bar extends Base {
  type: 'bar';
  bar: number;
}

function getV(): Foo | Bar {
    return null!;
}

const v = getV();

if((v.type) === 'bar') {
  v.type
}


// Works
switch (v.type) { // <===
  case 'bar':
    v.bar;
    break;

  case 'foo':
    v.foo;
    break;
}

// Does not work
switch ((v.type)) { // <===
  case 'bar':
    v.bar; // KO
    break;

  case 'foo':
    v.foo; // KO
    break;
}
```

### 🙁 Actual behavior

With extra braces in the `switch`, narrowing is broken. 

```
switch ((v.type)) { // <=== extra braces
  case 'bar':
    console.log(v.bar);
    break;

  case 'foo':
    console.log(v.foo);
    break;
}
```

### 🙂 Expected behavior

`switch ((v.type))` should behave as `switch (v.type)`. 

There is no such issue on a `if` condition.

PR #56020, opened 2023-10-07
Ignore awaited self tail calls when collecting the return type of an async function

Andarist:
closes https://github.com/microsoft/TypeScript/issues/55973
expands https://github.com/microsoft/TypeScript/pull/53995

cc @RyanCavanaugh

PR #56035, opened 2023-10-08
Fixed type narrowing in switch statements with parenthesized expressions

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/56030

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -27446,7 +27446,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         }
 
         function getTypeAtSwitchClause(flow: FlowSwitchClause): FlowType {
-            const expr = flow.switchStatement.expression;
+            const expr = skipParentheses(flow.switchStatement.expression);
             const flowType = getTypeAtFlowNode(flow.antecedent);
             let type = getTypeFromFlowType(flowType);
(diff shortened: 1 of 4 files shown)


JoostK:
Ah, you beat me to it :) I would have thought it was something like this. Thanks for providing the fix (and the _incredible_ work you do for TypeScript, really)

Andarist:
Thanks :)

jakebailey (approved):

jakebailey:
I merged this, but I should have run the tests. In light of #56020, is there a problem here in JS code which casts the argument to a `switch`, e.g.:

```js
switch (/** @type { "a" | "b" } */ (foo)) {
	case "a":
		return;
	default: {
		const x: "b": foo
	}
}
```

jakebailey:
Ah, yes, you should be using:

```
skipParentheses(node, /*excludeJSDocTypeAssertions*/ true)
```

jakebailey:
Hm, I can't seem to make a test case where this fails. I thought this would show a difference but it does not:

```ts
// @strict: true
// @noEmit: true
// @checkJs: true
// @allowJs: true

// @filename: index.js

let value = "";

switch (/** @type {"foo" | "bar"} */ (value)) {
  case "bar":
    value;
    break;

  case "foo":
    value;
    break;

  case "invalid":
    value;
    break;
}
```

PR #56046, opened 2023-10-09
Reuse more input nodes when they contain local type parameter references

### LINKED ISSUE #55575: Arrow function parameters using an alias and intersection are expanded inappropriately in `.d.ts`

weswigham:
Fixes #55575

Mostly, this just required adding local type parameters into the fake scopes we gin up during node building, and then ensuring type parameter node generation always went through the potentially-shadowed-type-parameter-name generation codepath (since we can no longer rely on their global accessibility to determine if their name is OK to use, since while it may be accessible via a fake scope, it might require a different name than it originally had due to shadowing).

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -8406,7 +8406,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     includePrivateSymbol?.(sym);
                 }
                 if (isIdentifier(node)) {
-                    const name = sym.flags & SymbolFlags.TypeParameter ? typeParameterToName(getDeclaredTypeOfSymbol(sym), context) : factory.cloneNode(node);
+                    const name = sym.flags & SymbolFlags.TypeParameter && context.flags & NodeBuilderFlags.GenerateNamesForShadowedTypeParams ? typeParameterToName(getDeclaredTypeOfSymbol(sym), context) : factory.cloneNode(node);
                     name.symbol = sym; // for quickinfo, which uses identifier symbol information
                     return { introducesError, node: setEmitFlags(setOriginalNode(name, node), EmitFlags.NoAsciiEscaping) };
                 }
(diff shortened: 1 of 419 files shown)


jakebailey:
This looks nearly the same as #55820, so maybe there was some issue duplication there? One thing that I realized that I needed was two fake scopes, since the same name can be both in type space and value space, but the scope can only hold one of them.

If you pull in my tests, do they work / fix the followup problem I discovered? I'm hoping there's a hybrid of the two PRs which avoids the pitfalls.

jakebailey:
It is surreal but very reassuring that we came up with effectively the same fix.

jakebailey:
I merged #55820; I didn't have it auto-close this PR though as I believe there some more reuse things in this PR that weren't covered in mine? (I know you're redoing more stuff anyhow so I'll leave it to you.)

weswigham:
We redid this (and actually merged it) recently as part of the node builder improvement work. 💖

PR #56047, opened 2023-10-09
pass program update level to getExternalFiles in tsserver plugins so plugins can make decision about their cache based on that

sheetalkamat:
Before this change, we called getExternalFiles on every project update (and some) To get correct behaviour, plugin users had to either watch things by themselves or inefficient way of refreshing this every call or just dont handle the updtes well
This passes the program update level to getExternalFiles so plugin users can make decisions based on that.
Fixes https://github.com/microsoft/TypeScript/issues/56044

### DIFF
--- a/src/compiler/watchUtilities.ts
+++ b/src/compiler/watchUtilities.ts
@@ -365,12 +365,17 @@ export function createCachedDirectoryStructureHost(host: DirectoryStructureHost,
     }
 }
 
-/** @internal */
-export enum ConfigFileProgramReloadLevel {
-    None,
-    /** Update the file name list from the disk */
-    Partial,
-    /** Reload completely by re-reading contents of config file from disk and updating program */
+export enum ProgramUpdateLevel {
+    /** Program is updated with same root file names and options */
+    Update,
+    /** Loads program after updating root file names from the disk */
+    RootNamesAndUpdate,
+    /**
+     * Loads program completely, including:
+     *   re-reading contents of config file from disk
+     *   calculating root file names for the program
+     *   Updating the program
+     */
     Full,
 }
 
--- a/src/compiler/watchUtilities.ts
+++ b/src/compiler/watchUtilities.ts
@@ -372,10 +372,11 @@ export enum ProgramUpdateLevel {
     RootNamesAndUpdate,
     /**
      * Loads program completely, including:
-     *   re-reading contents of config file from disk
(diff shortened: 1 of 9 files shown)


src/compiler/watchUtilities.ts
```diff
@@ -365,12 +365,17 @@ export function createCachedDirectoryStructureHost(host: DirectoryStructureHost,
     }
 }
 
-/** @internal */
-export enum ConfigFileProgramReloadLevel {
-    None,
-    /** Update the file name list from the disk */
-    Partial,
-    /** Reload completely by re-reading contents of config file from disk and updating program */
+export enum ProgramUpdateLevel {
```
sheetalkamat, line 368: Just renamed this and exposed as Public API for use in `getExternalFiles`

src/server/project.ts
```diff
@@ -1892,10 +1892,10 @@ export abstract class Project implements LanguageServiceHost, ModuleResolutionHo
             }
             // compute and return the difference
             const lastReportedFileNames = this.lastReportedFileNames;
-            const externalFiles = this.getExternalFiles().map((f): protocol.FileWithProjectReferenceRedirectInfo => ({
+            const externalFiles = this.externalFiles?.map((f): protocol.FileWithProjectReferenceRedirectInfo => ({
```
sheetalkamat, line 1895: this shouldnt have been querying things again from modules but should have been returning the current state in the first place. so just made that change

src/compiler/watchUtilities.ts
```diff
@@ -365,12 +365,17 @@ export function createCachedDirectoryStructureHost(host: DirectoryStructureHost,
...
+    /** Program is updated with same root file names and options */
+    Update,
+    /** Loads program after updating root file names from the disk */
+    RootNamesAndUpdate,
+    /**
+     * Loads program completely, including:
+     *   re-reading contents of config file from disk
+     *   calculating root file names for the program
+     *   Updating the program
+     */
```
andrewbranch, line 378: Nit: depending on the markdown-ish interpretation of the editor, these lines may have been rendered separated only by a space, which would be confusing. This will either make it parse as a real list or at least separate the items with a dash.

```suggestion
     * Loads program completely, including:
     *  - re-reading contents of config file from disk
     *  - calculating root file names for the program
     *  - Updating the program
     */

```

andrewbranch (approved):

PR #56048, opened 2023-10-09
Do not narrow using JSDoc type assertion nodes

Andarist:
spotted by @jakebailey [here](https://github.com/microsoft/TypeScript/pull/56035#issuecomment-1753571444)

### DIFF
--- a/src/compiler/binder.ts
+++ b/src/compiler/binder.ts
@@ -173,6 +173,7 @@ import {
     isJSDocEnumTag,
     isJSDocTemplateTag,
     isJSDocTypeAlias,
+    isJSDocTypeAssertion,
     isJsonSourceFile,
     isJsxNamespacedName,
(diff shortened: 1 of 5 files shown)


jakebailey:
Just to further describe things, `switch (value as "a" | "b")` or `switch (value satisfies "a" | "b")` don't allow narrowing to work today, but `switch (/** @type {"a" | "b"} */ (value))` _does_, because the existing code thinks it can skip parens safely. But, that type assertion in JS should behave just like `as/satisfies`.

jakebailey:
@typescript-bot test top200
@typescript-bot run dt
@typescript-bot pack this

jakebailey (approved):
This seems correct to me. Technically a breaking change but not one I expect anyone has ever encountered.

sandersn (approved):

PR #55870, opened 2023-09-26
Update timezone to allow for an optional undefined

aar9nk:
Fixes https://github.com/microsoft/TypeScript/issues/55869
Attempted fix for issue [55869](https://github.com/microsoft/TypeScript/issues/55869) that I created.
I believe this is somewhat common as it causes an actual error if chaining options. I'm not certain on the procedures when creating a PR here, I have followed the README and believe I have a successful contribution here but if there's any questions please reach out.

PR #56056, opened 2023-10-10
Fix type of `timeZone` so it can be `undefined`

### LINKED ISSUE #56057: The type of `timeZone` could be `undefined` but it's marked as a `string` only instead
### ⚙ Compilation target

es5

### ⚙ Library

Intl

The type of `Intl.DateTimeFormat.prototype.resolvedOptions().timeZone` is incorrectly set to only `string` where it could also be `undefined`.

### Sample Code

```TypeScript
// TypeScript should raise an error in the next line since the return value can be `undefined`
const getTimeZone = (): string => {
  return new Intl.DateTimeFormat().resolvedOptions().timeZone;
};
```

### Documentation Link

See possible values for the `timeZone` property in [MDN `resolvedOptions()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/resolvedOptions#description), where `undefined` is included.

luciomartinez:
Fixes #56057

I feel like a test could be added, but I'm having a hard time finding the appropriate test suit where this should be, so I can use some help with it. Hence, I'm setting it as _Ready for review_...

### DIFF
--- a/src/lib/es5.d.ts
+++ b/src/lib/es5.d.ts
@@ -4463,7 +4463,7 @@ declare namespace Intl {
         locale: string;
         calendar: string;
         numberingSystem: string;
-        timeZone: string;
+        timeZone?: string;
         hour12?: boolean;
         weekday?: string;
         era?: string;


fatcerberus:
A PR to change this was already closed: https://github.com/microsoft/TypeScript/pull/55870

PR #51214, opened 2022-10-18
fix(51202): Circular instantiation expression crashing IDEs

a-tarasyuk:
Fixes #51202
Fixes #52757

PR #56064, opened 2023-10-11
Ensure instantiation expressions have symbols, preventing crash in signature relations

### LINKED ISSUE #55810: Cannot read properties of undefined (reading 'parameters') in reportIncompatibleCallSignatureReturn
### 🔎 Search Terms

reportIncompatibleCallSignatureReturn
signatureRelatedTo

- This is a crash
- Tried downgrading to 4.x but the crash happened again
- Nightly crashes as well

### ⏯ Playground Link

_No response_

### 💻 Code

I tracked down the crash to these lines:
```ts
import * as results from 'ts-results';
import { Result as _Result } from 'ts-results';

// @ts-expect-error weird exports
const { Ok, Err } = results.default ? results.default : results;

export type Ok<T = void> = typeof results.Ok<T>;
export type Err<T = string> = typeof results.Err<T>;
```

Commenting the exports out, would fix the crash. However, I tried to reproduce it in a minimal example and couldnt get it to crash

From what I debugged, the crash happens because in `signaturesRelatedTo` source and target are not from the same length.

It loops over the `targetSignatures` and passes `sourceSignatues[i]` into `reportIncompatibleCallSignatureReturn` which is `undefined`

Hence the error `Cannot read properties of undefined (reading 'parameters')`

I am sorry that I cannot provide more info

### 🙁 Actual behavior

```
TypeError: Cannot read properties of undefined (reading 'parameters')
    at reportIncompatibleCallSignatureReturn (/home/repos/foo/node_modules/typescript/lib/tsc.js:62105:16)
    at signaturesRelatedTo (/home/repos/foo/node_modules/typescript/lib/tsc.js:62027:13)
    at structuredTypeRelatedToWorker (/home/repos/foo/node_modules/typescript/lib/tsc.js:61488:24)
    at structuredTypeRelatedTo (/home/repos/foo/node_modules/typescript/lib/tsc.js:60946:21)
    at recursiveTypeRelatedTo (/home/repos/foo/node_modules/typescript/lib/tsc.js:60899:19)
    at isRelatedTo (/home/repos/foo/node_modules/typescript/lib/tsc.js:60344:122)
    at someTypeRelatedToType (/home/repos/foo/node_modules/typescript/lib/tsc.js:60657:25)
    at unionOrIntersectionRelatedTo (/home/repos/foo/node_modules/typescript/lib/tsc.js:60554:14)
    at structuredTypeRelatedToWorker (/home/repos/foo/node_modules/typescript/lib/tsc.js:61104:23)
    at structuredTypeRelatedTo (/home/repos/foo/node_modules/typescript/lib/tsc.js:60946:21)
```

### 🙂 Expected behavior

`yarn tsc` should run and not crash

I am using node v19.8.1 but I know it also happens on node v20.

jakebailey:
Fixes #55810

~I am still not 100% sure why the two differ in length, but my understanding is roughly that the combo of a type alias and instantiation expression lead to the potential for two of the types to have the same symbol, but have different observed members. This leads the source to have a signature, but the target to not, but the current code assumes that if the two symbols are the same that they'll have the same number of signatures. But, that's not the case here. Skipping this particular path when the lengths differ leads to the correct result, and I think that's okay.~

The above is just all wrong, the main underlying problem is that instantiation expressions don't get symbols, which violates assumptions the checker makes elsewhere.

Notably, this PR is effectively a revert of #51214 then fixed again with a different fix that isn't so targeted toward type printing.


tests/baselines/reference/aliasInstantiationExpressionGenericIntersectionNoCrash.errors.txt
```diff
@@ -0,0 +1,23 @@
...
+    }
+    
+    declare const Err: typeof ErrImpl & (<T>() => T);
+    
+    type ErrAlias<U> = typeof Err<U>;
+    
+    declare const e: ErrAlias<number>;
+    e as ErrAlias<string>;
+    ~~~~~~~~~~~~~~~~~~~~~
+!!! error TS2352: Conversion of type 'typeof Err<U> & typeof Err<U>' to type 'typeof Err<U> & typeof Err<U>' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
```
gabritto, line 19: what the hell is this error message? 😕

jakebailey: Yes, exactly. I don't get it. I have debugged for hours trying to figure out _why_ these types when printed don't have their type parameters mapped. The mapper is there, the types are compared, queried, etc, correctly, but when you actually look at them, the type vars don't get replaced!

gabritto: it's also weird to me that we have `'typeof Err<U> & typeof Err<U>' in the first place... how is that not just `typeof Err<U>`?

jakebailey: welcome to hell!

jakebailey: Consider this a plea of "please someone else be nerd sniped by this too"

src/compiler/checker.ts
```diff
@@ -22993,8 +22993,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             const sourceObjectFlags = getObjectFlags(source);
             const targetObjectFlags = getObjectFlags(target);
             if (
-                sourceObjectFlags & ObjectFlags.Instantiated && targetObjectFlags & ObjectFlags.Instantiated && source.symbol === target.symbol ||
```
gabritto, line 22996: yet another question: the comment below says "We have instantiations of the same anonymous type", meaning that if we have two types with the same `symbol`, they're instantiations of the same type (presumably the type represented by the symbol). is that not true then, or is the actual bug that there are two different types with the same symbol?
I know you said in the PR description that "the combo of a type alias and instantiation expression lead to the potential for two of the types to have the same symbol, but have different observed members", but how can that be possible?

jakebailey: In this case, they are both instantiations of the same symbol because it's a type alias. But I agree that it's mega confusing, because surely I can invent a type alias whose final type is different depending on its type parameter, and yet we don't crash today. So I think my explanation is wrong and I still need to look at this harder.

jakebailey: Like, I feel like it's specifically "type alias pointing to an instantiation expression" that is the problem here, as though there's one extra level deep that was forgotten somewhere, and I just haven't found it.

Or like, this code needs to explicitly exempt instantiation expressions specifically _because_ they're going to have the same symbol, but other things don't. But then, do aliases of aliases not work today?

RyanCavanaugh:
@typescript-bot pack this

tests/baselines/reference/aliasInstantiationExpressionGenericIntersectionNoCrash.errors.txt
weswigham: I think it's a bug in the printer implementation for instantiation expression types in `createAnonymousTypeNode` - namely, it goes back to the declared type when deciding what to print for an instantiation expression on a type query node:
```ts
                    const isInstantiationExpressionType = !!(getObjectFlags(type) & ObjectFlags.InstantiationExpressionType);
                    if (isInstantiationExpressionType) {
                        const instantiationExpressionType = type as InstantiationExpressionType;
                        if (isTypeQueryNode(instantiationExpressionType.node)) {
                            const typeNode = serializeExistingTypeNode(context, instantiationExpressionType.node);
                            if (typeNode) {
                                return typeNode;
                            }
                        }
                        if (context.visitedTypes?.has(typeId)) {
                            return createElidedInformationPlaceholder(context);
                        }
                        return visitAndTransformType(type, createTypeNodeFromObjectType);
                    }
```
usually we use `serializeTypeForDeclaration` and not `serializeExistingTypeNode` directly because it checks that the existing node is actually the same type as the one we're looking at.

weswigham (commented):
> I am still not 100% sure why the two differ in length

IIRC, instantiation expressions filter signature lists to only those applicable for the number of type arguments supplied by the instantiation expression. So that case would be intentional, at least - signature relationship logic probably just didn't get updated to handle this case quite correctly. In the test, though, I wouldn't be able to tell you why those type have differing signature counts. It definitely looks like there could be at most one.

tests/baselines/reference/aliasInstantiationExpressionGenericIntersectionNoCrash.errors.txt
jakebailey: Aha, so dropping that `typeof` special case and letting it walk shows more info:

```
aliasInstantiationExpressionGenericIntersectionNoCrash.ts(10,1): error TS2352: Conversion of type '{ new (): ErrImpl<number>; prototype: ErrImpl<any>; } & (() => number)' to type '{ new (): ErrImpl<string>; prototype: ErrImpl<any>; } & (() => string)' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  Type '{ new (): ErrImpl<number>; prototype: ErrImpl<any>; } & (() => number)' is not comparable to type '{ new (): ErrImpl<string>; prototype: ErrImpl<any>; }'.
    Type 'ErrImpl<number>' is not comparable to type 'ErrImpl<string>'.
      Type 'number' is not comparable to type 'string'.
```

It's checking the two intersections against each other, so it goes down into one branch.

jakebailey: Looking at the changes here, I think that removing this special case is the right thing to do, honestly. The way instantiation expressions seem to work is different than type references; the type arguments that it was instantiated with just aren't preserved at all.

It makes me feel like instantiation expressions should instead be some form of TypeReference; all of the places that check InstantiationExpression appear to also look for Reference too...

jakebailey:
Okay, I have a clearer example:

```ts
declare class Class<T> {
  x: T;
}

declare function fn<T>(): T;


type ClassAlias<T> = typeof Class<T>;
type FnAlias<T> = typeof fn<T>;

type Wat<T> = ClassAlias<T> & FnAlias<T>;


declare const wat: Wat<number>;
wat as Wat<string>;
```

This code also crashes today. When we do the relation check, we do the whole "compare all of the bits and bobs in the two intersections", and in doing so encounter this:

image

Now that we can see past the `typeof`, we can clearly see that we are comparing something that has call signatures with something that doesn't. But, they're both marked as `Instantiated|InstantiationExpressionType`, so this is true:

```
sourceObjectFlags & ObjectFlags.Instantiated && targetObjectFlags & ObjectFlags.Instantiated && source.symbol === target.symbol
```

But they're also both Anonymous, so have no symbol, and we think they're _equal_.

I think the actual fix is not necessarily about the lengths (a debug assert should be left for that), but that this code doesn't know that you can have two instantiated anonymous types, which I believe only became possible when instantiation expressions were added.

That or, maybe these _shouldn't_ be marked as Instantiated? That seems weird; they are produced in `instantiateAnonymousType` after all.

jakebailey:
Actually, adding an assert to `instantiateAnonymousType` to check that we always have a symbol points out that maybe not having a symbol is the true strange thing here.

jakebailey:
Well, that's funny. Make these have symbols, the bugs go away, and all of the special casing for instantiated types without symbols goes away too.

jakebailey:
@typescript-bot test top200
@typescript-bot run dt
@typescript-bot pack this

jakebailey:
Also, I felt like I had seen this before and this PR effectively just undoes #51214, which in hindsight was too targeted and incorrect due to the aforementioned reuse of the declaration node (even though the type can be instantiated). We just didn't have an example yet where that mattered, and this PR introduces those tests (type aliases!).

src/compiler/checker.ts
```diff
@@ -6867,20 +6867,6 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
-                            if (typeNode) {
-                                return typeNode;
-                            }
-                        }
-                        if (context.visitedTypes?.has(typeId)) {
-                            return createElidedInformationPlaceholder(context);
-                        }
-                        return visitAndTransformType(type, createTypeNodeFromObjectType);
-                    }
                     // Anonymous types without a symbol are never circular.
```
gabritto, line 6870: If I understood this correctly, now that anonymous types that are instantiated always have a symbol, then the invariant stated in this comment holds again and we can get rid of the special code here, right?

And the invariant that "Anonymous types without a symbol are never circular" comes from the fact that you couldn't instantiate something that didn't have a name/declaration and therefore a symbol before instantiation expressions existed?

src/compiler/checker.ts
```diff
@@ -35559,7 +35547,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     hasSignatures ||= resolved.callSignatures.length !== 0 || resolved.constructSignatures.length !== 0;
                     hasApplicableSignature ||= callSignatures.length !== 0 || constructSignatures.length !== 0;
                     if (callSignatures !== resolved.callSignatures || constructSignatures !== resolved.constructSignatures) {
-                        const result = createAnonymousType(/*symbol*/ undefined, resolved.members, callSignatures, constructSignatures, resolved.indexInfos) as ResolvedType & InstantiationExpressionType;
+                        const result = createAnonymousType(createSymbol(SymbolFlags.None, InternalSymbolName.InstantiationExpression), resolved.members, callSignatures, constructSignatures, resolved.indexInfos) as ResolvedType & InstantiationExpressionType;
```
gabritto, line 35550: That seems like a correct change to re-establish symbol invariants we had before instantiation expressions, but it makes me wonder: when does a type have or not have a symbol? and when should it have a symbol?

jakebailey: Honestly, I'm not totally sure. It seems like many types don't have symbols; I tried adding the assert to createAnonymousType and it broke everything.

src/compiler/checker.ts
```diff
@@ -23000,6 +22987,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 // method). Simply do a pairwise comparison of the signatures in the two signature lists instead
                 // of the much more expensive N * M comparison matrix we explore below. We erase type parameters
                 // as they are known to always be the same.
+                Debug.assertEqual(sourceSignatures.length, targetSignatures.length);
```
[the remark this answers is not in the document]
jakebailey: They can actually be undefined; lots of anonymous types don't have symbols. More or less it seems like symbols are only around if the user actually wrote the type out or it's named or something. My understanding here is poor.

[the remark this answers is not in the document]
jakebailey: Adding `Debug.assertIsDefined(source.symbol);` here fails 297 tests, FWIW

src/compiler/checker.ts
jakebailey: That's my understanding, yeah.

src/compiler/checker.ts
jakebailey: One thing I'm wondering here is if there needs to be some sort of cache here to make sure we always use the same symbol for identical `exprType` + `type` combos. But obviously the code here works...

ahejlsberg: I haven't looked at this PR too closely, but wondering why you aren't just using the symbol already associated with `type` here? I mean, the anonymous type being created is just a specialization of that type, but it originates in the same declaration (which is what the symbol represents).

jakebailey: Good question; let me try and see what happens. I swear I tried that but I have forgotten the result.

jakebailey: That does seem to work (as in, fixing the crash), though it does create some strange baseline differences (forgive the screenshot):

image

Now the instantiation of the `typeof ErrImpl` appears to disappear.

And stuff like:

image

Where `<T>` is dropped.

src/compiler/checker.ts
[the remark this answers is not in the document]
gabritto: In case I wasn't clear, I didn't mean an assert, just an extra check in the `if` condition that wraps this, since the comment inside it says "We have instantiations of the same anonymous type" and assumes so, but if two instantiated types both have symbol `undefined`, it doesn't seem to me that that's a valid assumption?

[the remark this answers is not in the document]
jakebailey: Yeah, theoretically two symbol-less anonymous symbols could be compared here, but the intent of the code seems to be explicitly that this path is only for things from the same symbol.

src/compiler/checker.ts
```diff
@@ -6867,20 +6867,6 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
-                            const typeNode = serializeExistingTypeNode(context, instantiationExpressionType.node);
-                            if (typeNode) {
-                                return typeNode;
-                            }
-                        }
-                        if (context.visitedTypes?.has(typeId)) {
-                            return createElidedInformationPlaceholder(context);
-                        }
-                        return visitAndTransformType(type, createTypeNodeFromObjectType);
-                    }
```
weswigham, line 6883:
```ts
                    const isInstantiationExpressionType = !!(getObjectFlags(type) & ObjectFlags.InstantiationExpressionType);
                    if (isInstantiationExpressionType) {
                        const instantiationExpressionType = type as InstantiationExpressionType;
                        const existing = getEffectiveTypeAnnotationNode(instantiationExpressionType.node)!;
                        if (isTypeQueryNode(instantiationExpressionType.node) && getTypeFromTypeNode(existing) === type && existingTypeNodeIsNotReferenceOrIsReferenceWithCompatibleTypeArgumentCount(existing, type)) {
                            const typeNode = serializeExistingTypeNode(context, instantiationExpressionType.node);
                            if (typeNode) {
                                return typeNode;
                            }
                        }
                        if (context.visitedTypes?.has(typeId)) {
                            return createElidedInformationPlaceholder(context);
                        }
                        return visitAndTransformType(type, createTypeNodeFromObjectType);
                    }
```
Should allow still reusing the `typeof symbolName<Thing>` input node when that node isn't affected by further instantiation. Like in the `typeof Array<number>` case.

weswigham (approved):
This is definitely more generally correct - but for the sake of nicer quickinfo, we might still want some capability to printback an instantiation expression type as the instantiation expression where possible, rather than always the expanded instantiated instantiation expression type. In particular, (and I think I mentioned this on the original PR at some point but it never made it in), the deleted code that tried to write out the instantiation expression type needs to sanity check the type it's building a bit more, which is what `serializeTypeForDeclaration` (vs `serializeExistingTypeNode`) does (basically, just checking that the node being reused, `instantiationExpressionType.node`, actually is for the same type as the type being printed!).

src/compiler/checker.ts
jakebailey: Cool, I'll try it!

jakebailey: Well, this will have to be put up into the other branch now that they have symbols but I think it should work.

jakebailey: Seems like `getEffectiveTypeAnnotationNode` is always returning `undefined`; is there a bit I'm missing?

weswigham: Yeah, the edit I made a few seconds after posting swapping it to just be `getTypeFromTypeNode` (which may need to be conditioned on the `isTypeQueryNode` check). 😜

jakebailey: Thanks, that did make the baselines no longer change for any test in this PR. Is `getTypeFromTypeNode(existing) === type` enough, though? I'm thinking of a case where the type comes from another file's instantiation expression but the node shouldn't be reused in another file.

But, removing the check (to see what it does) does show that it's doing something.

PR #56070, opened 2023-10-11
Use sameMap in mapType to avoid allocation when possible

jakebailey:
Noticed this while working on a perf regression.

This function always allocates a new array. But if the mapping function doesn't change the array, we could avoid it. Elsewhere, we use `sameMap` for this kind of thing, but presumably we didn't in `mapType` as mappers can _remove_ types by returning `undefined`. To keep behavior, this now calls `compact`, which also returns the original array when possible.


jakebailey:

jakebailey:
Well, that's not super conclusive.

jakebailey:
@typescript-bot perf test

jakebailey:

jakebailey:
I guess this doesn't help.

PR #56072, opened 2023-10-11
Shortcut relations in getNarrowedTypeWorker when comparing literals

jakebailey:
For #55948

When narrowing large unions of literals, we do four different relation calls which appear to be quite costly. By checking ahead of time that we're going to relate two literals, we can skip this work by comparing the values directly and only once (as the result shouldn't be order dependent).

This leads to a significant speedup on #55948:

```
Benchmark 1: main
  Time (mean ± σ):     12.456 s ±  0.851 s    [User: 12.929 s, System: 0.059 s]
  Range (min … max):   11.219 s … 13.428 s    10 runs
 
Benchmark 2: PR
  Time (mean ± σ):      2.973 s ±  0.053 s    [User: 3.434 s, System: 0.057 s]
  Range (min … max):    2.906 s …  3.073 s    10 runs
 
Summary
  'PR' ran
    4.19 ± 0.30 times faster than 'main'
```

This PR passes all of the tests, but I think my implementation is probably wrong for some edge cases. ~Like, I don't validate that the literal types are the same between the two, just their values, which is wrong. But, doesn't break?~ Fixed that edge case.


jakebailey:
@typescript-bot test top100
@typescript-bot run dt
@typescript-bot pack this

jakebailey:

src/compiler/checker.ts
```diff
@@ -28240,7 +28240,29 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
                         t => isTypeDerivedFrom(t, c) ? t : isTypeDerivedFrom(c, t) ? c : neverType :
-                        t => isTypeStrictSubtypeOf(t, c) ? t : isTypeStrictSubtypeOf(c, t) ? c : isTypeSubtypeOf(t, c) ? t : isTypeSubtypeOf(c, t) ? c : neverType,
+                        t => {
+                            // WAS: isTypeStrictSubtypeOf(t, c) ? t : isTypeStrictSubtypeOf(c, t) ? c : isTypeSubtypeOf(t, c) ? t : isTypeSubtypeOf(c, t) ? c : neverType;
+
+                            // Avoid going through relations for comparison between literals of the same types;
+                            // if they're not equal then they're not related and we can skip doing a bunch of work.
+                            if (t.flags & TypeFlags.Literal && (t.flags & TypeFlags.Literal) === (c.flags & TypeFlags.Literal)) {
+                                if (t.flags & TypeFlags.BooleanLiteral) {
+                                    return t === c ? t : neverType;
```
weswigham, line 28250: `true` and `false` have fresh and un-fresh versions (and since this is control flow, you can probably witness both, since that's the point of having two) - so you can't *quite* only check `===` equality - there's 2 true's and 2 false's.

```suggestion
                                    return ((t === trueType || t === regularTrueType) ? (c === trueType || c === regularTrueType) : (c === falseType || c === regularFalseType)) ? t : neverType;

```

Actually, it's somewhat surprising we don't have an `isSameBooleanLiteralValue` function lying around, but I guess it just doesn't come up too often?

jakebailey: Ah, yes, I forgot about those. I should really pull this into a function.

jakebailey: We probably don't need the helper, because the first thing isRelatedTo does is to strip freshness.

jakebailey: Also... as long as I don't allow this code to work when two enum literals are being compared, this entire optimization works when brought into `isTypeRelatedTo`. I might commit that to run perf on it quick as that's kinda interesting. But, I suspect that it will actually hurt the case I'm trying to fix because it's still going to make 4 calls to relations even though it doesn't need to.

It's kinda like relations should say something about both the positive and negative case, e.g. answer stuff like "A and B are not related, and also B is not related to A so no need to call me later".

jakebailey:
@typescript-bot test top200

jakebailey:
Hm, not very good, probably because if we're already willing to pay the cost to do relations, it's not going to be very expensive to go through simpleTypeRelated or whatever. But in the other place, we're guaranteed to call relation _4 times_ when the types aren't related, which is expensive on large unions.

jakebailey:

sandersn:
@jakebailey did the results indicate that this is worth pursuing? It is worth keeping open?
A quick glance at the latest perf run didn't show anything substantial.

jakebailey:
You can see the relevant benchmark in the original post; our current test suite doesn't have a case like this.

PR #56075, opened 2023-10-12
Some of the test refactoring and readable baselining

sheetalkamat:
Pulled out from #56074

### DIFF
--- a/src/jsTyping/jsTyping.ts
+++ b/src/jsTyping/jsTyping.ts
@@ -30,6 +30,9 @@ import {
     Version,
     versionMajorMinor,
 } from "./_namespaces/ts";
+import {
+    stringifyIndented,
+} from "./_namespaces/ts.server";
 
 export interface TypingResolutionHost {
     directoryExists(path: string): boolean;
@@ -236,7 +239,7 @@ export function discoverTypings(
         }
     });
     const result = { cachedTypingPaths, newTypingNames, filesToWatch };
-    if (log) log(`Result: ${JSON.stringify(result)}`);
+    if (log) log(`Finished typings discovery:${stringifyIndented(result)}`);
     return result;
 
     function addInferredTyping(typingName: string) {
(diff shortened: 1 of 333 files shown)


tests/baselines/reference/api/typescript.d.ts
```diff
@@ -3450,7 +3450,6 @@ declare namespace ts {
             getLanguageService(): never;
             getHostForAutoImportProvider(): never;
             getProjectReferences(): readonly ts.ProjectReference[] | undefined;
-            getTypeAcquisition(): TypeAcquisition;
```
jakebailey, line 3453: Is this an intentional change?

jakebailey (commented):
Largely this looks good to me, though I just have the one comment about the API change, which doesn't feel intentional in a test-only change.

tests/baselines/reference/api/typescript.d.ts
sheetalkamat: It isnt test only change but i had done this as part of https://github.com/microsoft/TypeScript/pull/56074 first and since it has no impact, left it there

src/jsTyping/jsTyping.ts
```diff
@@ -211,37 +214,37 @@ export function discoverTypings(
         );
         addInferredTypings(module, "Inferred typings from unresolved imports");
     }
+    // Remove typings that the user has added to the exclude list
```
sheetalkamat, line 217: This is order change and shouldnt have any impact as such on functionality a359540

src/server/project.ts
```diff
@@ -2640,10 +2640,6 @@ export class AutoImportProviderProject extends Project {
         return PackageJsonAutoImportPreference.Off;
     }
 
-    override getTypeAcquisition(): TypeAcquisition {
```
sheetalkamat, line 2643: This shouldnt  be needed since dedault typAcquisition is always disabled

jakebailey (approved):

PR #56079, opened 2023-10-12
fix: add `Intl.getCanonicalLocales` type definition

### LINKED ISSUE #29129: [i18n] Intl getCanonicalLocales / RelativeTimeFormat / ListFormat > 2339 error
**TypeScript Version:**  3.2.2

**Search Terms:**
Intl.getCanonicalLocales TS TypeScript Error 2339 undefined i18n

**Code**

```ts
Intl.getCanonicalLocales('EN-US');
```

**Expected behavior:**
Code works, and returns ["en-US"]
https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/Intl/getCanonicalLocales

**Actual behavior:**
TS Error 2339, the property getCanonicalLocales does not exists on Intl

kazupon:
Fixes #29129

The issue to resolve this PR is not `Backlog` milestone.
`Intl.getCanonicalLocales` type definition should be included in #45647, but unfortunately it could not do.

Even now, the `Intl.getCanonicalLocales` type is not resolved.
We can check it in the playground.
https://www.typescriptlang.org/play?ssl=1&ssc=6&pln=1&pc=25#code/JIOwLgNgdA5gpmAwgQxAexASwMbIgGTVwjgGcg

## about `Intl.getCanonicalLocales`
- MDN: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/getCanonicalLocales
- ecma402 spec: https://tc39.es/ecma402/#sec-intl.getcanonicallocales

## why do i add as `es2016.intl` 
`Intl.getCanonicalLocales` is supported since 2016, so some proejcts using typescript might still use es2016.
I've added `es2016.intl.d.ts` for compatibility and defined in it.

## Parameters and return types of `Intl.getCanonicalLocales`
`UnicodeBCP47LocaleIdentifier` is not defined at `es2020.intl.d.ts`. In the past, [A comment](https://github.com/microsoft/TypeScript/pull/39662#issuecomment-852194127) of #39662 has tried to support `UnicodeBCP47LocaleIdentifier` by defining it in es2016.
However, if we do that, it will result in a destructive change for some users, as this comment states.

`UnicodeBCP47LocaleIdentifier` is currently an alias for `string`. So, as a compromise, the parameters and return value of `Intl.getCanonicalLocales` are defined as string. by that way, we can keep compatibility.


kazupon:

sandersn:
It took a lot of digging, but I found proof that this was indeed added in 2016: The first page of https://402.ecma-international.org/3.0/ mentions it explicitly.

sandersn (approved):

PR #56086, opened 2023-10-12
Fix bug template line breaks, which do not follow normal markdown behavior

jakebailey:
I didn't realize it, but the markdown used in the bug template is apparently rendered using the same markdown processing as issues, which is to say that newlines are preserved even if they are part of the same paragraph. This is different than regular markdown files where paragraphs are joined. Annoying difference.

Anyway, I didn't intend for it to look like this:

image

The lines were supposed to join.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -26,14 +26,11 @@ body:
       description: |
         When did you start seeing this bug occur?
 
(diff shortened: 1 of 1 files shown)


jakebailey:
Had I prechecked on my fork I would have seen this and fixed it in the previous PR, I'm just gonna quick merge this to de-gross the bug template.

fatcerberus:
@jakebailey You might want to check the other templates too.  They’re not quite as bad as this, but still a bit icky with line breaks in odd places.

Issue #49171, opened 2022-05-19
Portability error issued when reachable source file is redirected to an unreachable target


jordanmbell:
# Bug Report

### 🔎 Search Terms

resolution, symlinks, references, incorrect dependency

### 🕗 Version & Regression Information

I have tested typescript@4.0.2 and typescript@next and both experience this.

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about module resolution

### ⏯ Playground Link

I can't give a playground link as this issue revolves around dependencies

### 💻 Code

I have created a repo demonstrating this here: https://github.com/jordanmbell/ts-resolution-issue

It is difficult to demonstrate this issue in a few lines because it revolves around dependencies, but the errors are reported here.

```
import { BAD_TYPE, OTHER_BAD } from "fake-dep";

export const bad = (arg: BAD_TYPE) => {
  return arg;
};

const other_bad: OTHER_BAD = 1;

console.log(bad);
```
### Screenshot of the issue

Here we see that there are two errors reported, one saying:
`The inferred type of 'bad' cannot be named without a reference to 'subpackage-a/node_modules/fake-dep/dist/BadType'. This is likely not portable. A type annotation is necessary.ts(2742)
`
and one saying:
`Type 'number' is not assignable to type 'string'.ts(2322)`

However from the screenshots of the dependency inside that package, it would seem this error should not be a problem, as `OTHER_BAD` has type `number`, and there is a direct reference to `BAD_TYPE`. Strangely though, typescript tells us it is actually using the version of `fake-dep` from a completely different subpackage.

![image](https://user-images.githubusercontent.com/50465383/169198524-7c565726-ff18-4fcd-8a4f-6c22dcf361f9.png)

![image](https://user-images.githubusercontent.com/50465383/169204970-ebb3ad9f-1be3-4afa-8579-84a9e4fc5a55.png)

I believe the problem is that typescript is caching dependencies when it first finds them, and uses the same located version of that dependency when it finds it used later, even if that dependency should not be reachable. This repo demonstrates that inside of `subpackage-d`, which has an index file that imports `fake-dep` and creates some variables. `tsc` reports a problem though, that `other_bad` should be a `string`, and that `bad cannot be named without a reference to 'subpackage-a/node_modules/fake-dep/dist/BadType'`. These errors occur because rather than `subpackage-d` using the version of `fake-dep` found inside its `node_modules`, it uses the version found inside of `subpackage-a`'s `node_modules`.

[...]

PR #56087, opened 2023-10-12
Use symbols of type aliases when emitting declarations

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/49171

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -5665,6 +5665,12 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             if (getSymbolIfSameReference(exported, symbol)) {
                 return exported;
             }
+            if (symbol.flags & SymbolFlags.TypeAlias) {
+                const aliasSymbol = getDeclaredTypeOfTypeAlias(exported).aliasSymbol;
+                if (aliasSymbol && getSymbolIfSameReference(aliasSymbol, symbol)) {
+                    return exported;
+                }
+            }
         });
     }
 
--- a/tests/baselines/reference/declarationEmitUsingTypeAlias1.js
+++ b/tests/baselines/reference/declarationEmitUsingTypeAlias1.js
@@ -1,0 +1,47 @@
+//// [tests/cases/compiler/declarationEmitUsingTypeAlias1.ts] ////
+
+//// [inner.d.ts]
+export declare type Other = { other: string };
+export declare type SomeType = { arg: Other };
+
+//// [index.d.ts]
+export type OtherType = import('./inner').Other;
+export type SomeType = import('./inner').SomeType;
+
+//// [package.json]
+{
+  "name": "some-dep",
+  "exports": {
+    ".": "./dist/index.js"
+  }
+}
+
+//// [index.ts]
+import { SomeType } from "some-dep";
+
+export const foo = (thing: SomeType) => {
+  return thing;
+};
+
+export const bar = (thing: SomeType) => {
+  return thing.arg;
+};
+
+//// [index.js]
+"use strict";
+Object.defineProperty(exports, "__esModule", { value: true });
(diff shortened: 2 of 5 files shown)


src/compiler/checker.ts
```diff
@@ -5665,6 +5665,12 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             if (getSymbolIfSameReference(exported, symbol)) {
                 return exported;
             }
+            if (symbol.flags & SymbolFlags.TypeAlias && exported.declarations?.find(isTypeAlias)) {
```
jakebailey, line 5668: I think we canonically use our own `find` helper here to avoid the `this` penalty; might be important if this is hot code.

Andarist: I have literally copied this bit from `getDeclaredTypeOfTypeAlias` :P
https://github.dev/microsoft/TypeScript/blob/cf8763b6c3193deefd095c86ce79921f23de307b/src/compiler/checker.ts#L12436

Should I switch both call sites to ur custom `find`?

jakebailey: Hah, then nevermind.

sandersn: >  to avoid the `this` penalty
The original reason was that `find` wasn't defined in old runtimes. What is the `this` penalty? Maybe that's a good replacement reason.

jakebailey: IIRC functions passed to `find` and other helpers like it are bound such that `this` in them points to the array.

src/compiler/checker.ts
```diff
@@ -5665,6 +5665,12 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             if (getSymbolIfSameReference(exported, symbol)) {
                 return exported;
             }
+            if (symbol.flags & SymbolFlags.TypeAlias && exported.declarations?.find(isTypeAlias)) {
+                const aliasSymbol = getDeclaredTypeOfTypeAlias(exported).aliasSymbol;
+                if (aliasSymbol && getSymbolIfSameReference(aliasSymbol, symbol)) {
```
weswigham, line 5670: Shouldn't this be a change in the logic of `getSymbolIfSameReference` instead? To make a type alias that points directly at another type count as a "same reference" of that type? It's been reused in one or two other places, but it basically only exists to answer "should we consider these two symbols as ultimately referring to the same thing", and mostly just for the usages here, in this function. In particular, only doing this check here means 1. the fastpath lookup will fail for type aliases that pass this check, so they always hit the slow path and 2. an `export =`'d type alias will *never* be considered equal to the target symbol.

weswigham (changes requested):

src/compiler/checker.ts
```diff
@@ -5672,6 +5672,12 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
      * Checks if two symbols, through aliasing and/or merging, refer to the same thing
      */
     function getSymbolIfSameReference(s1: Symbol, s2: Symbol) {
+        if (s1.flags & SymbolFlags.TypeAlias && s2.declarations?.find(isTypeAlias)) {
```
weswigham, line 5675: Maybe extract this into another helper function? Seems a bit repetitive.

Andarist: This checks both `s1` and `s2` and I don't have any good name for a helper function that would accept both as arguments or other ideas on how to improve this. Everything I have tried locally felt clunky to me.

weswigham (approved):

weswigham:
@typescript-bot test this
@typescript-bot run dt
@typescript-bot test top100

jakebailey:
I've sent a revert for this in #57849; this PR caused two bugs and a perf regression, so I think it's safest to back it out as it was just to fix a bug that has existed seemingly forever.

Issue #56082, opened 2023-10-12
Array of objects union - autocompletion & intellisense not working properly


ivodolenc:
### 🔎 Search Terms

"array of objects union", "array union narrowing", "union narrowing", "object union operator in"

### 🕗 Version & Regression Information

TypeScript: 5.2.2

### ⏯ Playground Link

https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgKLisCBnZBvAKGWOQHoAqcok5c5ABQBsBXAc1F1BgHsA6f6iXKlByAA4t2IbAH4AXMgBKEBNygATADzYwmEKwA0SlWq069h5MxABrENwDuIAHzPRFKjVrIA8szBi-shcfAJewqLc-oFg8sjmoKzulKLeAGJqALZwYMEgPPy8qRE0PFDZsQoJ+gQAvgQEoJCwiCgAkiAxyBAAHpAg6rjouli4hDQexcgdXSGFxSI0oDFVIzX1jeDQ8EjIACoAnmI43X0QA0MYo-jJnjR0h8ec+aFF4YskYEc4qxZ1DV9jmgrjgfGIwMBuNJkABeaadIIAH3232wDQIqmkuW44Mh0gUw0woNxUOwAG0ALqw5Bk0TjLzEZb+BQAcl4pGwUAQpFA6l6vDA2BZRiiAWZyDZpHUwB0PIG-MyACshaJagY6alPqjWezOdzATgBUKjGUKqycJkWaqCBSgA

### 💻 Code

[...]

PR #56088, opened 2023-10-12
Improve quick info display on properties belonging to all union contextual type members

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/56082

### DIFF
--- a/src/services/services.ts
+++ b/src/services/services.ts
@@ -3282,7 +3282,9 @@ export function getPropertySymbolsFromContextualType(node: ObjectLiteralElementW
         // Bad discriminant -- do again without discriminating
         return mapDefined(contextualType.types, t => t.getProperty(name));
     }
-    return discriminatedPropertySymbols;
+    // by eliminating duplicates we might even end up with a single symbol
+    // that helps with displaying better quick infos on properties of union types
+    return Array.from(new Set(discriminatedPropertySymbols));
 }
 
 function isArgumentOfElementAccessExpression(node: Node) {
--- a/src/services/services.ts
+++ b/src/services/services.ts
@@ -3284,7 +3284,7 @@ export function getPropertySymbolsFromContextualType(node: ObjectLiteralElementW
     }
     // by eliminating duplicates we might even end up with a single symbol
     // that helps with displaying better quick infos on properties of union types
-    return Array.from(new Set(discriminatedPropertySymbols));
+    return deduplicate(discriminatedPropertySymbols, equateValues);
 }
 
 function isArgumentOfElementAccessExpression(node: Node) {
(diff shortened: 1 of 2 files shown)


src/services/services.ts
```diff
@@ -3282,7 +3282,9 @@ export function getPropertySymbolsFromContextualType(node: ObjectLiteralElementW
         // Bad discriminant -- do again without discriminating
         return mapDefined(contextualType.types, t => t.getProperty(name));
     }
-    return discriminatedPropertySymbols;
+    // by eliminating duplicates we might even end up with a single symbol
+    // that helps with displaying better quick infos on properties of union types
+    return Array.from(new Set(discriminatedPropertySymbols));
```
jakebailey, line 3287: I think we technically have a `dedupe` helper but this is probably fine.

Andarist: oh, cool! I was looking for one but couldn't find it 😅 I'll switch to it in a sec

jakebailey: (we technically _also_ have `arrayFrom` which is faster than `Array.from` but whatever)

src/services/services.ts
```diff
@@ -3282,7 +3282,9 @@ export function getPropertySymbolsFromContextualType(node: ObjectLiteralElementW
         // Bad discriminant -- do again without discriminating
         return mapDefined(contextualType.types, t => t.getProperty(name));
     }
-    return discriminatedPropertySymbols;
+    // by eliminating duplicates we might even end up with a single symbol
+    // that helps with displaying better quick infos on properties of union types
+    return deduplicate(discriminatedPropertySymbols, equateValues);
```
RyanCavanaugh, line 3287: I don't understand why this "works" -- which symbol gets discarded, and how do we know the one that's being kept is the one with doc comments attached?

jakebailey: They're all the same symbol, so wouldn't all of that info be the same? It's just that we're q	querying potentially the same info from multiple angles and don't dedupe?

(Why don't we do this in the above mapDefined case, too?)

RyanCavanaugh: It seems like the deduping should be happening upstream of this, or that deduping should already be a prevalent pattern in this file - surely this isn't the only place where we might end up with multiple instances of the same symbol in the list?

Andarist: This works because only a single symbol is allowed by the caller here:
https://github.com/microsoft/TypeScript/blob/b4787652d281188da1d61fe9cc3a17589f20c9fc/src/services/services.ts#L3255-L3258

We can't dedupe this too soon because it would interfere with `.length`-based checks like this one:
https://github.com/microsoft/TypeScript/blob/b4787652d281188da1d61fe9cc3a17589f20c9fc/src/services/services.ts#L3277

I agree though that perhaps this deduping should also happen here:
https://github.com/microsoft/TypeScript/blob/b4787652d281188da1d61fe9cc3a17589f20c9fc/src/services/services.ts#L3283

but it would be cool to figure out a test case for it before making the change.

RyanCavanaugh (approved):
I don't object to this as-is but a future refactor would be nice

sandersn (approved):
I agree with Ryan that this area needs some cleanup.

PR #56090, opened 2023-10-12
Add user preference for preferring type-only auto imports

### LINKED ISSUE #55675: Support a user preference for preferring the type keyword in auto-imports
### 🔍 Search Terms

import type auto-import

### ✅ Viability Checklist

### ⭐ Suggestion

When an import is generated automatically (by the language server?), I'd like to be able to have it generate a `type` import if the imported symbol is used in the file only as a type. Probably, this preference would have options for whether the type import should be of the `import type` or the `import { type ... }` form.

### 📃 Motivating Example

See below

### 💻 Use Cases

Some codebases prefer to use type-only imports whenever possible to indicate to readers that an import isn't actually used as a value. Meanwhile, other codebases prefer the brevity of `import` without `type`. 

Currently, a number of TS compiler settings influence the auto-imports that the language server generates, but some of those settings are hard to turn on in existing codebases (e.g., `verbatimModuleSyntax`, because of how it effects interop) and others are deprecated. Moreover, none of them are actually designed to apply a stylistic preference across the board — they only lead to `type` imports being generated when it's absolutely necessary for interop, so they won't generate type-only imports as often as a user of this preference would want.

@andrewbranch made the case for adding preference in #54664

> The flags that affect type-only auto imports... affect things in precisely the way that the compiler requires, minus bugs. None of them are specifically designed to control auto-imports. It sounds like an antipattern to use compiler options that are supposed to control real things in order to arrive at an editor behavior that matches a particular stylistic preference, so I would much rather give you a user preference.

Really, this issue is just a duplicate of #54664, because the original author closed that issue for unrelated reasons.

andrewbranch:
Closes #55675

This makes the auto-imports behavior of `verbatimModuleSyntax` available as an editor preference, independent of compiler options. Stylistic preference for type-only imports has been a popular demand since the syntax was introduced, and compiler options which may or may not be appropriate for a user’s actual compilation have long been adopted for their side effects on auto-import behavior. This setting allows users to separate their stylistic preference from picking the right compiler options for their actual use case.

### DIFF
--- a/src/services/codefixes/importFixes.ts
+++ b/src/services/codefixes/importFixes.ts
@@ -389,6 +389,7 @@ function createImportAdderWorker(sourceFile: SourceFile, program: Program, useAu
                 namedImports && arrayFrom(namedImports.entries(), ([name, addAsTypeOnly]) => ({ addAsTypeOnly, name })),
                 namespaceLikeImport,
                 compilerOptions,
+                preferences,
             );
             newDeclarations = combine(newDeclarations, declarations);
         });
@@ -1348,6 +1349,7 @@ function codeActionForFixWorker(
                     namedImports,
                     namespaceLikeImport,
                     program.getCompilerOptions(),
+                    preferences,
                 ),
                 /*blankLineBetween*/ true,
                 preferences,
@@ -1505,7 +1507,7 @@ function doAddExistingFix(
         const newSpecifiers = stableSort(
             namedImports.map(namedImport =>
                 factory.createImportSpecifier(
-                    (!clause.isTypeOnly || promoteFromTypeOnly) && needsTypeOnly(namedImport),
+                    (!clause.isTypeOnly || promoteFromTypeOnly) && shouldUseTypeOnly(namedImport, preferences),
                     /*propertyName*/ undefined,
                     factory.createIdentifier(namedImport.name),
                 )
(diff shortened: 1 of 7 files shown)


sheetalkamat (approved):

src/services/codefixes/importFixes.ts
```diff
@@ -1604,32 +1606,37 @@ function needsTypeOnly({ addAsTypeOnly }: { addAsTypeOnly: AddAsTypeOnly; }): bo
...
+    preferences: UserPreferences,
 ): AnyImportSyntax | readonly AnyImportSyntax[] {
     const quotedModuleSpecifier = makeStringLiteral(moduleSpecifier, quotePreference);
     let statements: AnyImportSyntax | readonly AnyImportSyntax[] | undefined;
     if (defaultImport !== undefined || namedImports?.length) {
         // `verbatimModuleSyntax` should prefer top-level `import type` -
         // even though it's not an error, it would add unnecessary runtime emit.
         const topLevelTypeOnly = (!defaultImport || needsTypeOnly(defaultImport)) && every(namedImports, needsTypeOnly) ||
-            compilerOptions.verbatimModuleSyntax &&
+            (compilerOptions.verbatimModuleSyntax || preferences.preferTypeOnlyAutoImports) &&
```
jakebailey, line 1628: Can this use `shouldUseTypeOnly` somehow?

jakebailey (approved):
Seems good to me, though I can't help but feel like we're going in circles a bit 😅 

I guess anyone who actually wants to enforce this needs `verbatimModuleSyntax` and/or the two ts-eslint rules, and this at least helps the editor do the right thing, more or less exactly like my proposal for how we should better handle import sorting.

jakebailey:
(of course note that the PR is failing, some missing baselines)

johnnyreilly:
This is amazing!! I will be updating my blog post on the topic: https://johnnyreilly.com/typescript-5-importsnotusedasvalues-error-eslint-consistent-type-imports

I'll confess to being possibly unhealthily interested in being exact in the type (ha!) of my imports.  Every time a type import is used, an angel gets their wings.

Issue #55021, opened 2023-07-14
Module resolution using nodenext and a wildcard in package.json exports


mikearnaldi:
# Bug Report

### 🔎 Search Terms

Nodenext wildcard, type cannot be named

### 🕗 Version & Regression Information
The bug only appears when using nodenext for module resolution

### ⏯ Playground Link

Cloneable repo: https://github.com/geoffreytools/effect-test-tsc

### 💻 Code

```ts
import { TaggedClass } from '@effect/data/Data';
// import '@effect/data/Equal';

export class Foo extends TaggedClass('Foo')<{}> {}
```

### 🙁 Actual behavior

```
The inferred type of 'Foo' cannot be named without a reference to '../node_modules/@effect/data/Equal'. This is likely not portable. A type annotation is necessary.
```

### 🙂 Expected behavior

To compile correctly, without using nodenext it works fine and the same issue doesn't apply to other transitive imports such as Hash that are identically exported. All goes away the very moment there is either an explicit import of `Equal` or when adding to the library package.json the following:

```ts
{
  "exports": {
    "./Equal": {
      "import": {
        "types": "./Equal.d.ts",
        "default": "./mjs/Equal.mjs"
      }
    }
  }
}
```

Note that the library uses a wildcard `./*` as each import map to a file and this is perfectly resolvable by node

datner:
This impacts all multi-module/multi-entry modules..
Can someone take a look at this? it's really tedious trying to build in esm with this bug

andrewbranch:
I think https://github.com/microsoft/TypeScript/issues/56107 was a duplicate of this—see my analysis there.

mikearnaldi:
> I think #56107 was a duplicate of this—see my analysis there.
> 
> > This impacts all multi-module/multi-entry modules.
> 
> #56107 was specific to a re-export pattern used by Effect; I haven’t seen other libraries affected by this. @datner, you may have been running into something else entirely. If anyone is having issues with this outside of Effect, please provide a repro so we can see if there’s something different going on.

What's specific to Effect here? we are just re-exporting a module

schickling:
I've created a minimal reproduction of this issue using Cloudflare Workers types that successfully triggers the TS2742 error:

https://github.com/schickling-repros/2025-01-typescript-portable-type-issue

## The Error
```bash
[...]

PR #56100, opened 2023-10-13
Emit declarations using alias chains

Andarist:


weswigham:
Obligatory: Is there a linked issue?

AFAIK this is sensible, but probably expensive - `getAliasForSymbolInContainer` is already costly, since it's an exhaustive traversal of all symbols in scope looking for aliases that refer to the target symbol - making it recursive is going to increase the worst case cost exponentially without some very smart caching. And, perhaps moreover, `getAliasForSymbolInContainer` is basically just a duplicated tiny slice of the inner loop of `getAccessibleSymbolChain` without all the recursion, safety, or caching rigamarole, to try and prevent the need for the (still expensive) `getAccessibleSymbolChain` call. By making it return a chain like this, isn't it basically just a copy of `getAccessibleSymbolChain` that preferences aliases, without all the aforementioned safety valves? Maybe it'd just be better to modify the `useExternalAliasingOnly` parameter into an alias-preference-flags parameter so `getAccessibleSymbolChain` is suitable for use here directly?

I say that, but I'm also actively working on a replacement for `getAccessibleSymbolChain` itself that's a bit less.... bad. Exhaustive scope traversal on lookup is not a good thing (there are many scopes with many symbols), and I'd like to remove it; while we don't have the perf numbers for it, `getAccessibleSymbolChain` has basically organically grown in function since TS 1.8 or earlier, making it the oldest part of the declaration emit pipeline that's never seen a major refactor, and its' implementation is pretty unoptimized (but with 5+ years of edge cases written into it, hgggg) - it's come up as a hotspot in specific examples multiple times in the past, but we've always worked around the issue by adding limiters higher up in the declaration emit or checking process.

I'd also say "we want to check the perf on this, at least", but declaration emit performance is still a bit of a blind spot in our perf suite. It's usually only lightly exercised by the error messages generated by the perf tests. 😭

weswigham (changes requested):
Yeah, definitely see if this can be integrated into the normal `getAccessibleSymbolChain` function. It has some inefficiencies that I'm looking into fixing, but so does this solution as-written, so better to keep them in one place.

fubhy:
Also linked (atm. unclear if this fixes it though): https://github.com/microsoft/TypeScript/issues/55021

PR #56101, opened 2023-10-13
Specialized parsing error for apparent JSX tag in .ts file

### LINKED ISSUE #29375: Suggest renaming file from .ts to .tsx if needed?
## Search Terms

ts tsx react jsx rename file errors

## Suggestion

When adding JSX elements (React) to a `.ts` file rather than a `.tsx` file, error messages aren't clear that it should be a `.tsx` file. If it's clear the user is intending to write JSX code, such as having multiple JSX-like components and/or components that extend `React.Component`, can a code fix be given (on the file? on individual `<>`-related parsing errors?) to rename to `.tsx`?

## Use Cases

There's been some buzz on Twitter lately about how easy it is to forget to rename to `.tsx`. Some users are apparently even used to writing JSX in `.js`, so the need to use `.tsx` is a bit confusing for them.

## Examples

Given this `file.ts`:

```ts
import * as React from "react";

export class ComponentClass extends React.Component {
    public render() {
        return <span />;
        // index.ts(5,22): error TS1005: '>' expected.
        // index.ts(5,23): error TS1161: Unterminated regular expression literal.
    }
}

export const ComponentArrow = () => <span />;
// index.ts(11,43): error TS1005: '>' expected.
// index.ts(11,44): error TS1161: Unterminated regular expression literal.
```

Would it be possible to special-case just from these errors to suggest to rename the file? Alternately, if we just have the `ComponentArrow` errors, would it still be possible to suggest to rename the file?

## Checklist

JoshuaKGoldberg:
Adds a specialized diagnostic for the case of a `/>` (slash and greater-than-sigh) after an identifier in a non-JSX-mode file.

```plaintext
JSX tags are not permitted in `.ts` files. Did you mean to change the file extension to `.tsx`?
```

I'm not at all confident in this PR's approach. It feels _weird_ to work with a "`previousNode`" and then skip through tokens. Would love to know if there's a cleaner way of doing things, please!

This also doesn't capture every possible syntax variant that could appear to be from JSX syntax in a non-JSX file. I couldn't figure out any clean way to get more cases that came to mind.

Fixes #29375

### DIFF
--- a/src/compiler/parser.ts
+++ b/src/compiler/parser.ts
@@ -162,6 +162,7 @@ import {
     isStringOrNumericLiteralLike,
     isTaggedTemplateExpression,
     isTemplateLiteralKind,
+    isTypeAssertionExpression,
     isTypeReferenceNode,
     IterationStatement,
     JSDoc,
@@ -5558,9 +5559,9 @@ namespace Parser {
         );
     }
 
-    function parseBinaryExpressionOrHigher(precedence: OperatorPrecedence): Expression {
+    function parseBinaryExpressionOrHigher(precedence: OperatorPrecedence, previousNode?: Node): Expression {
         const pos = getNodePos();
-        const leftOperand = parseUnaryExpressionOrHigher();
+        const leftOperand = parseUnaryExpressionOrHigher(previousNode);
         return parseBinaryExpressionRest(precedence, leftOperand, pos);
     }
 
@@ -5626,7 +5627,7 @@ namespace Parser {
                 }
             }
             else {
-                leftOperand = makeBinaryExpression(leftOperand, parseTokenNode(), parseBinaryExpressionOrHigher(newPrecedence), pos);
+                leftOperand = makeBinaryExpression(leftOperand, parseTokenNode(), parseBinaryExpressionOrHigher(newPrecedence, leftOperand), pos);
             }
         }
 
@@ -5698,7 +5699,7 @@ namespace Parser {
      *      1) UnaryExpression[?Yield]
      *      2) UpdateExpression[?Yield] ** ExponentiationExpression[?Yield]
      */
-    function parseUnaryExpressionOrHigher(): UnaryExpression | BinaryExpression {
+    function parseUnaryExpressionOrHigher(previousNode?: Node): UnaryExpression | BinaryExpression {
         /**
          * ES7 UpdateExpression:
          *      1) LeftHandSideExpression[?Yield]
@@ -5709,7 +5710,7 @@ namespace Parser {
          */
         if (isUpdateExpression()) {
             const pos = getNodePos();
-            const updateExpression = parseUpdateExpression();
+            const updateExpression = parseUpdateExpression(previousNode);
             return token() === SyntaxKind.AsteriskAsteriskToken ?
                 parseBinaryExpressionRest(getBinaryOperatorPrecedence(token()), updateExpression, pos) as BinaryExpression :
                 updateExpression;
@@ -5835,7 +5836,7 @@ namespace Parser {
      *      5) --LeftHandSideExpression[?yield]
      * In TypeScript (2), (3) are parsed as PostfixUnaryExpression. (4), (5) are parsed as PrefixUnaryExpression
      */
-    function parseUpdateExpression(): UpdateExpression {
+    function parseUpdateExpression(previousNode?: Node): UpdateExpression {
         if (token() === SyntaxKind.PlusPlusToken || token() === SyntaxKind.MinusMinusToken) {
             const pos = getNodePos();
             return finishNode(factory.createPrefixUnaryExpression(token() as PrefixUnaryOperator, nextTokenAnd(parseLeftHandSideExpressionOrHigher)), pos);
@@ -5845,7 +5846,7 @@ namespace Parser {
             return parseJsxElementOrSelfClosingElementOrFragment(/*inExpressionContext*/ true);
         }
 
-        const expression = parseLeftHandSideExpressionOrHigher();
+        const expression = parseLeftHandSideExpressionOrHigher(previousNode);
 
         Debug.assert(isLeftHandSideExpression(expression));
         if ((token() === SyntaxKind.PlusPlusToken || token() === SyntaxKind.MinusMinusToken) && !scanner.hasPrecedingLineBreak()) {
@@ -5857,7 +5858,7 @@ namespace Parser {
         return expression;
     }
 
-    function parseLeftHandSideExpressionOrHigher(): LeftHandSideExpression {
+    function parseLeftHandSideExpressionOrHigher(previousNode?: Node): LeftHandSideExpression {
         // Original Ecma:
         // LeftHandSideExpression: See 11.2
         //      NewExpression
@@ -5909,11 +5910,11 @@ namespace Parser {
                 sourceFlags |= NodeFlags.PossiblyContainsImportMeta;
             }
             else {
-                expression = parseMemberExpressionOrHigher();
+                expression = parseMemberExpressionOrHigher(previousNode);
             }
         }
         else {
-            expression = token() === SyntaxKind.SuperKeyword ? parseSuperExpression() : parseMemberExpressionOrHigher();
+            expression = token() === SyntaxKind.SuperKeyword ? parseSuperExpression() : parseMemberExpressionOrHigher(previousNode);
         }
 
         // Now, we *may* be complete.  However, we might have consumed the start of a
@@ -5922,7 +5923,7 @@ namespace Parser {
         return parseCallExpressionRest(pos, expression);
     }
 
-    function parseMemberExpressionOrHigher(): MemberExpression {
+    function parseMemberExpressionOrHigher(previousNode?: Node): MemberExpression {
         // Note: to make our lives simpler, we decompose the NewExpression productions and
         // place ObjectCreationExpression and FunctionExpression into PrimaryExpression.
         // like so:
@@ -5971,7 +5972,7 @@ namespace Parser {
         // Because CallExpression and MemberExpression are left recursive, we need to bottom out
         // of the recursion immediately.  So we parse out a primary expression to start with.
         const pos = getNodePos();
-        const expression = parsePrimaryExpression();
+        const expression = parsePrimaryExpression(previousNode);
         return parseMemberExpressionRest(pos, expression, /*allowOptionalChain*/ true);
     }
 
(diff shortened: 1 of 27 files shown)


RyanCavanaugh:
A good fix might not exist. We really don't know something's syntactically invalid until way later; this code *is* legal (!):
```ts
declare const div: any, foo: any;
type div = {};
const a = <div>foo</div>/;
```

src/compiler/parser.ts
```diff
@@ -6327,8 +6328,18 @@ namespace Parser {
...
         parseExpected(SyntaxKind.LessThanToken);
+        const posLessThan = getNodePos();
         const type = parseType();
-        parseExpected(SyntaxKind.GreaterThanToken);
+
+        // <Type />: specialized diagnostic for a JSX element in a .ts file - and skip some excess reporting
+        if (token() === SyntaxKind.SlashToken && scanner.scan() === SyntaxKind.GreaterThanToken) {
+            nextToken();
+            parseErrorAt(posLessThan - 1, getNodePos(), Diagnostics.JSX_tags_are_not_permitted_in_ts_files_Did_you_mean_to_change_the_file_extension_to_tsx, tokenToString(SyntaxKind.GreaterThanToken));
+        }
```
JoshuaKGoldberg, line 6338: I'd originally had it return here, so it doesn't try to `parseSimpleUnaryExpression`. But I couldn't figure out how to make an empty identifier that obeyed node ordering rules. See failures from up till [fd5adf7](https://github.com/microsoft/TypeScript/pull/56101/commits/fd5adf73cfd89c33c2e891f4519cf445aa774b78).

sandersn (changes requested):
There are parts I like and parts I dislike in this change. The change in parsePrimaryExpression is pretty invasive whereas the parseTypeAssertion is simple. How much of the benefit comes from the parseTypeAssertion change?

I also don't like the fact that old errors are still around, and that the new error spans the entire tag. I guess it doesn't *really* matter what the span is, since the important thing is for the "why not JSX?" message to be early in the list of error messages.

I almost wish we could add this message to an existing error that appears with JSX, only in cases where we can be sure that the cause really is a tag. Would that be a better approach? I'm not sure any of the existing errors have enough context to decide that they were *really, truly* caused by a JSX tag.

jakebailey:
I'm not totally sure how I feel about threading the previous node down into the parser like this, honestly. I'm not sure of a better way, though, besides passing down a boolean that's true if it's a type assertion or something.

src/compiler/parser.ts
```diff
@@ -5558,9 +5559,9 @@ namespace Parser {
         );
     }
 
-    function parseBinaryExpressionOrHigher(precedence: OperatorPrecedence): Expression {
+    function parseBinaryExpressionOrHigher(precedence: OperatorPrecedence, previousNode?: Node): Expression {
```
JoshuaKGoldberg, line 5562: From https://github.com/microsoft/TypeScript/pull/56101#issuecomment-1819801432:

> 
I'm not totally sure how I feel about threading the previous node down into the parser like this, honestly. I'm not sure of a better way, though, besides passing down a boolean that's true if it's a type assertion or something.

I'll just go ahead and say this: I hate it. Adding in the first concept of a "previous" node to a parser for a specialized error message makes me unhappy.

An alternative could be to keep a variable alongside `var currentToken: SyntaxKind` & co. I dislike that ... more, honestly. But it's an option. 🤷

src/compiler/parser.ts
```diff
@@ -5698,7 +5699,7 @@ namespace Parser {
      *      1) UnaryExpression[?Yield]
      *      2) UpdateExpression[?Yield] ** ExponentiationExpression[?Yield]
      */
-    function parseUnaryExpressionOrHigher(): UnaryExpression | BinaryExpression {
+    function parseUnaryExpressionOrHigher(previousNode?: Node): UnaryExpression | BinaryExpression {
```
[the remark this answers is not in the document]
JoshuaKGoldberg: It wouldn't be a 100% straightforward "replace the `?` with `| undefined` then add `/*previousNode*/ undefined` to function calls because of a `nextTokenAnd(parseLeftHandSideExpressionOrHigher)`. But it's 13 one-line changes otherwise: https://github.com/JoshuaKGoldberg/TypeScript/commit/2f6b35e5b9bca1c887d6b0a54d7558dc1a6b9dfc

tests/baselines/reference/nonTsxExtensionOpeningClosingTagMultiline.errors.txt
```diff
@@ -0,0 +1,23 @@
...
+                           ~~~~~~
+                            ~~~~
+!!! error TS2304: Cannot find name 'span'.
+        yippee
+    ~~~~~~~~~~
+        ~~~~~~
+!!! error TS2304: Cannot find name 'yippee'.
+    </span>;
+    ~~~~~~~
+!!! error TS1465: JSX tags are not permitted in `.ts` files. Did you mean to change the file extension to `.tsx`?
```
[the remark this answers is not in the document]
JoshuaKGoldberg: It's the entire expression, yes.

src/compiler/parser.ts
JoshuaKGoldberg: Hard to say. I would personally assume rather low given that most JSX element children are often one of:

* More JSX
* Plain text: will often be parsed mostly as identifiers and/or odd punctuation characters
* `{}` blocks containing a function call or some other odd code - which is sometimes valid non-JSX

src/compiler/parser.ts
```diff
@@ -6601,6 +6611,25 @@ namespace Parser {
             case SyntaxKind.NewKeyword:
                 return parseNewExpressionOrNewDotTarget();
             case SyntaxKind.SlashToken:
+                // JSX inside a .ts (not .tsx) file gets a specialized error...
```
JoshuaKGoldberg, line 6614: From https://github.com/microsoft/TypeScript/pull/56101#pullrequestreview-1735462101:

> The change in parsePrimaryExpression is pretty invasive whereas the parseTypeAssertion is simple. How much of the benefit comes from the parseTypeAssertion change?

Hard to say what balance comes from parsing self-closing tags (`parseTypeAssertion`) vs. tags with children (`parsePrimaryExpression`). My _intuition_ is that tags with children (the one we find more invasive) is _probably_ much more common - especially in complex pages where the error messages without this change are convoluted. 😕

JoshuaKGoldberg, line 6614: From https://github.com/microsoft/TypeScript/pull/56101#pullrequestreview-1735462101:

> I almost wish we could add this message to an existing error that appears with JSX, only in cases where we can be sure that the cause really is a tag. Would that be a better approach? I'm not sure any of the existing errors have enough context to decide that they were really, truly caused by a JSX tag.

I looked into that at first but couldn't find any good path forward 😞. JSX syntax issues tend to cause multiple seemingly unrelated parsing errors. If you have a good path I would be _very_ happy to go down it.

src/compiler/parser.ts
```diff
@@ -6327,7 +6328,16 @@ namespace Parser {
...
         const pos = getNodePos();
         parseExpected(SyntaxKind.LessThanToken);
+        const posLessThan = getNodePos();
         const type = parseType();
+
+        // <Type />: specialized diagnostic for a JSX element in a .ts file - and skip some excess reporting
+        if (token() === SyntaxKind.SlashToken && scanner.scan() === SyntaxKind.GreaterThanToken) {
+            nextToken();
+            parseErrorAt(posLessThan - 1, getNodePos(), Diagnostics.JSX_tags_are_not_permitted_in_ts_files_Did_you_mean_for_the_file_extension_to_be_tsx, tokenToString(SyntaxKind.GreaterThanToken));
+            return finishNode(factory.createTypeAssertion(type, createMissingNode(SyntaxKind.Identifier, /*reportAtCurrentPosition*/ false)), pos);
```
JoshuaKGoldberg, line 6338: From https://github.com/microsoft/TypeScript/pull/56101#pullrequestreview-1735462101:

> I also don't like the fact that old errors are still around, and that the new error spans the entire tag. I guess it doesn't really matter what the span is, since the important thing is for the "why not JSX?" message to be early in the list of error messages.

Agreed! Even after adding `createMissingNode` (thanks for that!) there is still typically a _`Cannot find name '{0}'.`_ reported on the type assertion. I tried replacing the `factory.createTypeAssertion` with a `missingNode` but that produced _`Error: Debug Failure. False expression: Invalid token`_ in the tests.

src/compiler/parser.ts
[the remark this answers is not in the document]
sandersn: Thanks. The main thing this points out to me is that there is only one narrow code path that preserves the previous node, and a bunch of others that could, but don't. I guess nothing else needs this information, but it still seems wasteful to wire it up for one usage.

src/compiler/parser.ts
sandersn: There should be a TypeAssertion somewhere earlier in the parser tree -- but we don't know something has gone wrong until much later, when we see an unexpected `/` that doesn't end up parsing as a regex.

What if you are extra suspicious of TypeAssertionExpressions? Maybe after parsePrimaryExpression you can do a speculative check for illegal stuff. That should at least let you detect small things like `<x>y</x>`, and probably also illegal things like `<x>y z</x>`. In theory that would then let you do some error recovery to skip past the closing tag, but that sounds quite difficult.

src/compiler/parser.ts
```diff
@@ -6327,7 +6328,16 @@ namespace Parser {
...
+        }
+
         parseExpected(SyntaxKind.GreaterThanToken);
         const expression = parseSimpleUnaryExpression();
```
sandersn, line 6342: Right after this, you'd want to check `if (token() === SyntaxKind.Identifier)` and `tryParse` some function that combines `parseJSXChildren` and `parseJSXClosingElement`. If that succeeds, you'll need to rebuild the previously parsed nodes above into an opening element (somehow??), issue your error, and return a JSX element. *In theory* the function you `tryParse` could be extracted from the first `if` body of `parseJsxElementOrSelfClosingElementOrFragment`, although it may do too much error recovery to be usable in a possibly-erroneous position.

Aside: Edge's AI-suggestion features are getting more aggressive but boy howdy was it confused by that paragraph.

Anyway, `if (token() === SyntaxKind.LessThanToken)` here, you'd instead `tryParse(parseJSXClosingElement)`, which covers the `<x>y</x>` case.

I'm not yet sure how to reparse `<x>y` as JSX opening/child token. Maybe there's somewhere else in the parser that does that, or maybe you can create some error fields for TypeAssertion and just cram the body/close token there. Like I said, TypeAssertions are rarely used at this point, so making its structure weird or inefficient is not a huge problem.

JoshuaKGoldberg:
Sorry to not have a good update here - this is a tricky issue and I don't have the bandwidth to focus on it now. 😞 

If someone else could push this forward, that'd be great.

PR #56105, opened 2023-10-13
Use contextual type to determine 'this' when determining member visibility

### LINKED ISSUE #56094: Class members are accessible or inaccessible on `this` depending on how `this` is defined
### 🔎 Search Terms

class members protected only accessible

- This is a crash
- This is the behavior in every version I tried, and I reviewed the FAQ for entries about __something or other_

### ⏯ Playground Link

https://www.typescriptlang.org/play?#code/MYGwhgzhAEBiD29oG8BQ1oAcBO8AuApsIQCbQBmAlgSGQLzQDkARmNo+tMPJgJ4CEALmgQ82SgDsA5p24TR2AK7F42ABQBKFJwzNJJAMI9eAOmBgQINXgAWlCBp3Q9Ew8YBMZi1dv3HGAF9UINRyRQliSngJZ30jPms7CGEEeC00DF8IM2NoBiyTKhoSYNRUPF5MAmgAIX1JKVhwyOi86ET7FMQtOgA+aAA3eEoS1DlRWNd43ndhOtcGpoi8KJiGMOXVzW1MpJy+NoKi2mCgA

### 💻 Code

```ts
class Foo {
  protected field = 'bar'
  copy!: string
  constructor() {
    bindCopy.call(this)
    bindCopy2.call(this)
  }
}

function bindCopy(this: Foo) {
  this.copy = this.field
}

type BindingFunction = (this: Foo) => void

const bindCopy2: BindingFunction = function() {
  this.copy = this.field
}
```

### 🙁 Actual behavior

Second definition (`bindCopy2`) throws:
```
Property 'field' is protected and only accessible within class 'Foo' and its subclasses.
```

### 🙂 Expected behavior

Both definitions have the same `this` binding. One is simply abstracted into a type. Both should pass similarly.

I recently wanted to refactor some functions I had to bind values in a class's constructor. Note that while this code looks simple, it should be noted that the actual functions CANNOT be methods on the class instance, because they function as "mixins" on multiple classes.

I naively assumed I could just define a type for each function, and thus eliminate the `this` declaration in each binding function. However, to my surprise, moving the type of `this` to a separate `type` definition caused the `protected` error to be thrown in the function.

Because the function is identical, and the value of `this` is identical, it is not intuitive why one form is allowed and the other isn't.

RyanCavanaugh:
Fixes #56094


jakebailey:
@typescript-bot test top200
@typescript-bot run dt
@typescript-bot pack this

src/compiler/checker.ts
```diff
@@ -32041,10 +32041,22 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
-        if (thisType && thisType.flags & TypeFlags.TypeParameter) {
-            thisType = getConstraintOfTypeParameter(thisType as TypeParameter);
+        if (thisType) {
+            // 2. The constraint of a type parameter used for an explicit 'this' parameter
+            if (thisType.flags & TypeFlags.TypeParameter) {
+                thisType = getConstraintOfTypeParameter(thisType as TypeParameter);
+            }
+        }
+        else {
+            // 3. This 'this' parameter of a contextual type
```
sandersn, line 32055:
```suggestion
            // 3. The 'this' parameter of a contextual type
```

sandersn (approved):

RyanCavanaugh:
@typescript-bot test this

PR #56117, opened 2023-10-16
Make "extract to constant" find and replace all similar expressions

### LINKED ISSUE #35349: Extract Constant Refactoring: Add way to extract all instances of selected expression
From https://github.com/microsoft/vscode/issues/85507

## Search Terms
- refactor
- extract 
- extract constant / extract function

## Suggestion
For the code:

```js
function doStuff() {
    console.log('hello')
    console.log('hi')
    console.log('hello')
}
```

If I run `extract constant on the string `'hello'`, it'd be nice if we offered a way to extract all instances of that expression in the current scope. This would turn the code into:

```js
function doStuff() {
    const text = 'hello'
    console.log(text)
    console.log('hi')
    console.log(text)
}
```

## Use Cases
- Help reducing duplicated code 
- Speed up code cleanup

mariusGundersen:
The goal of this pull-request is for "extract ..." refactoring to find all similar expressions and replaces them all with the new variable created.

For example, it would be nice to turn the following:

```ts
function topFiveSomething(a: SomeObject){
  if(a.something.something.length < 5){
    return a.something.something;
  } else {
    return a.something.something.slice(0, 5);
  }
}
```

Into this:

```ts
function topFiveSomething(a: SomeObject){
  const something = a.something.something;
  if(something.length < 5){
    return something;
  } else {
    return something.slice(0, 5);
  }
}
```

But the current "extract to constant in enclosing scope" will only find and replace on instance of `a.something.something`. This PR makes it find all of them and replace all of them. 

This pr makes the extract constant refactorings replace all matching values, as long as they are fairly simple expressions. 

Fixes #35349

### DIFF
--- a/src/services/refactors/extractSymbol.ts
+++ b/src/services/refactors/extractSymbol.ts
@@ -170,7 +170,7 @@ import {
     visitEachChild,
     visitNode,
     visitNodes,
-    VisitResult
+    VisitResult,
 } from "../_namespaces/ts";
 import {
     refactorKindBeginsWith,
@@ -1477,7 +1477,7 @@ function extractConstantInScope(
                 if (isLiteralExpression(a) && isLiteralExpression(b)) {
                     return a.text === b.text;
                 }
-                else if(isBooleanLiteral(a) && isBooleanLiteral(b)){
+                else if (isBooleanLiteral(a) && isBooleanLiteral(b)) {
                     // must be true here, since we check if a.kind !== b.kind above
                     return true;
                 }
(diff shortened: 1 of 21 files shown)


mariusGundersen:

sandersn:
The original bug hasn't been accepted, it looks to me because there are open questions and related design discussions that need to happen.
In particular, the boundary for what kinds of expressions should be included. This PR has an answer, but I'd like to see it summarised in English and agreed on the original bug.

I don't want to review this PR until the bug is accepted because design changes could change the code quite a bit.

(I'm also curious if any other IDEs besides Jetbrains ones have this feature but that's not very important.)

sandersn:
This PR has been open for quite some time with open design questions, so I'm going to close it.

Issue #56121, opened 2023-10-16
TypeScript reports a cycle when using a function with an Expression as body


cpojer:
### 🔎 Search Terms

initializer cycle implicit any

### 🕗 Version & Regression Information

- This is the behavior in every version I tried.

### ⏯ Playground Link

https://www.typescriptlang.org/play?#code/C4TwDgpgBAEg9gdwDYFEBOa5oMIEMlIBGuAxgNZQC8UAFAM5wCuAdgCYCSrAXFM4wLaEIaADRRhmNDxZlmiZgEoqAPigA3OAEtWAbgBQezc2DCAZqWjxkAeTDBNcZnSgBvPVA9RHYJLhASsAH4eK1QMLDwCYnIoAB8oFlYIUyMIXT0AXwMSXzpnUNd3TxJHOmA0RhJgLBowRkIkTRIoNAhcVkckEC87BycQxCRbe1KlFyzPL2YSCBpmXH4IHgByFiQ4cmWxU2YeGiVKVQ1tMaLJkqc4JAgAOnWAczmFiG3FfUmsrL0LsqgjMtw02g1GYEAQsEGNDcHm8vn84SktAOqn+wEBMxujhmNFWzHWmzE+xUrgyCkyCn02VKwD+TjRQIATFReGCIcgoUVYX4AoiiYdCpNWsBGGhmLSAYzMUCcWsNmQtkjieMKUVPirvtTxfSZgBmZmg8GhDkw5g+bkIvbI9RaVha9EQHVS7G4-HywlW5Xkyk-Gmo+0AFn1rKN0KmZvhkktSrOdqB-qdsxdcoVfNUyveUDVOiAA

### 💻 Code

```ts
type HowlErrorCallback = (soundId: number, error: unknown) => void;

interface HowlOptions {
    onplayerror?: HowlErrorCallback | undefined;
}

class Howl {
    constructor(public readonly options: HowlOptions) {}
    once(name: 'unlock', fn: () => void) {
      console.log(name, fn);
    }
}

const instance = new Howl({
  onplayerror: () => instance.once('unlock', () => {})
});


const instance2 = new Howl({
  onplayerror: () => {
    return instance2.once('unlock', () => {});
  }
});

const instance3 = new Howl({
  onplayerror: () => void instance3.once('unlock', () => {})
});

const instance4 = new Howl({
  onplayerror: () => {
    instance4.once('unlock', () => {});
  }
});
```

### 🙁 Actual behavior

[...]

PR #56123, opened 2023-10-16
Avoid reporting a cycle when contextual return type is `any`, `unknown` or `void`

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/56121

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -36874,7 +36874,8 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     }
                 }
                 if (contextualSignature && !getReturnTypeFromAnnotation(node) && !signature.resolvedReturnType) {
-                    const returnType = getReturnTypeFromBody(node, checkMode);
+                    const contextualReturnType = getContextualReturnType(node, /*contextFlags*/ undefined);
+                    const returnType = contextualReturnType && contextualReturnType.flags & (TypeFlags.AnyOrUnknown | TypeFlags.Void) ? contextualReturnType : getReturnTypeFromBody(node, checkMode);
                     if (!signature.resolvedReturnType) {
                         signature.resolvedReturnType = returnType;
                     }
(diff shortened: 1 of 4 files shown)


src/compiler/checker.ts
```diff
@@ -36874,7 +36874,8 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     }
                 }
                 if (contextualSignature && !getReturnTypeFromAnnotation(node) && !signature.resolvedReturnType) {
-                    const returnType = getReturnTypeFromBody(node, checkMode);
+                    const contextualReturnType = getContextualReturnType(node, /*contextFlags*/ undefined);
+                    const returnType = contextualReturnType && contextualReturnType.flags & (TypeFlags.AnyOrUnknown | TypeFlags.Void) ? contextualReturnType : getReturnTypeFromBody(node, checkMode);
```
Andarist, line 36878: This has a somewhat unwanted side effect - those types become now assigned to `signature.resolvedReturnType`. This shouldn't (usually?) be observable by the user but might be observable by the API users.

There is also a chance that this could break some convoluted scenarios with named functions - when the inner closure has access to a function that I've altered here and calls it directly~. For that reason... perhaps those added lines should skip named functions.

RyanCavanaugh:
@typescript-bot pack this

Andarist:
I mentioned the named functions problems in the comment [here](https://github.com/microsoft/TypeScript/pull/56123#discussion_r1360585935). I could fine-tune this further to only skip those when functions are anonymous. The question is - is there any merit in the approach used here at all?

RyanCavanaugh:
I don't think it's workable. A contextual type is usually an indication that the final result won't be observable, but it's not a guarantee - in generics, nested assignments, and `satisfies`, there will be a contextual type but the type of the expression can also end up bound to a symbol.

Computing the contextual type can cause also new circularity errors where one wasn't there previously.

PR #56130, opened 2023-10-16
More reliable mechanism for `_this`/`_super` simplification

### LINKED ISSUE #55637: Incorrect ES5 transpilation for 'this' usage in lambda as a constructor default parameter
### 🔎 Search Terms

ES5, this alias

- This changed between versions 5.1.6 and 5.2.2

### ⏯ Playground Link

https://www.typescriptlang.org/play?target=1&ts=5.2.2#code/MYGwhgzhAEDqBOYAOSCm9oG8BQ1rAHsA7CAF3gFdhSD4AKXPaANzBAEsATMU1aAXmh0KJMADNUALiwBfAJQCAfNFIALdhAB0rDt17DREuQpxMWbLj1R1M8gNyMZ2RknjtWvc7qt1SATzQAVUMpaDAiPzlpACMCAhBUcKxGPHhUUgp4IhVKVAc8JydsIlQAdzhEFHQ6OQcgA

### 💻 Code

```ts repro
// @target: es5
// @showEmit: true
class Wrapper {
  constructor(
    validate = (unsafe: {}) => this.validate(unsafe)) {
    validate({});
  }

  private validate(typeUnsafe: any): boolean {
    return true;
  }
}

new Wrapper();
```

Target: ES5

### 🙁 Actual behavior

```
Cannot read properties of undefined (reading 'validate') 
```

### 🙂 Expected behavior

No error (as with untranspiled version or with TS5.1.6)

_No response_

### LINKED ISSUE #55646: Incorrect ES5 transpilation for try/catch + super
### 🔎 Search Terms

ES5, this alias

- This changed between versions 5.1.6 and 5.2.2

### ⏯ Playground Link

https://www.typescriptlang.org/play?target=1&ts=5.2.2#code/IYIwzgLgTsDGEAJYBthjAgYge2wg3gFAJLYB2kUArvNlABRgAW2VyAJgCpNTYDuALgQhcyAKbAyASgLESCAJYAzBIxZsuPfjKLz5ELXwRkxRgKJReDAOQAFcWjEIoY6AE9rUgNxySAXwQxZDAnXT1SCmxxADpkbABzemsAeQBpTx89Pzls7MIUNAwAIWAoQIAPCDEydgwcPDDYckoaCDp6HV8Ed1lwhDAqAAcxBmgqMW8ugNhgCFgmVTEhKjIAazJ+aV7wpsiYuMTrCyshawQAakWENARjuilogFsxdGB4iczwgHov51coNxdEgDYYMJTAYIfKY5Qh5ExGEoMbxAA

### 💻 Code

```ts repro
// @target: es5
// @showEmit: true

abstract class Foo {
  constructor(shouldThrow: boolean) {
    if (shouldThrow) {
      throw new Error('Please retry');
    } else {
      console.log('OK');
    }
  }
}

class Bar extends Foo {
  constructor() {
    try {
      super(true);
    } catch (e: unknown) {
      console.log('Error: ' + (e as Error).message);
      // retry
      super(false);
    }
  }
}

new Bar();
```

Target: ES5

### 🙁 Actual behavior

```
_this is not defined 
```

### 🙂 Expected behavior

No error

In TS5.2.2 output `var _this = this;` has been removed, but `_this` is still being used.

rbuckton:
This changes the somewhat fragile way we handled the simplification of constructor body statements when transforming `super()` that became even more fragile in TS 5.2. These simplifications were intended to slim down transformed constructor bodies when a temporary variable like `_this` may not be necessary:

```js
// given
class C extends Base {
  constructor() {
    super();
  }
}

// without simplification
var C = (function (_super) {
    function C() {
        var _this = this;
        _this = _super.call(this) || this;
        return _this;
    }
    return C;
})(Base);

// with simplificaiton
var C = (function (_super) {
    function C() {
        return _super.call(this) || this;
    }
    return C;
})(Base);
```

However, when `using` was introduced in TS 5.2, the simplification mechanism became more complicated as the `super()` statements we would normally look for would be moved inside of a `try..catch..finally` block if the constructor contained a `using` declaration.

Rather than try to handle specific `super()` simplifications when initially visiting the constructor body, this PR instead performs these simplifications in a second pass after the initial transformation completes. This results in a mechanism that is more reliable than the prior approach as it only performs these simplifications with complete knowledge of the transformed result..

Fixes #55646
Fixes #55637


rbuckton:
@typescript-bot perf test

tests/baselines/reference/callWithSpread.js
```diff
@@ -136,8 +136,7 @@ var D = /** @class */ (function (_super) {
     __extends(D, _super);
     function D() {
         var _this = _super.call(this, 1, 2) || this;
```
jakebailey, line 138: Is it intentional to leave this unused variable behind? I guess it's harmless?

jakebailey (approved):

tests/baselines/reference/callWithSpread.js
rbuckton: Yes and no. It's intentional in that the conditions for the simplification that elides `var _this =` isn't met. We could change the rules, in the future if necessary, but I don't want to spend too much time processing these simplifications during emit, which are primarily only to improve readability when debugging unmapped outputs, so I've generally opted for more coarse-grained rules.

rbuckton:
I'm going to make a small update to make `updatePropertyName` public to align with https://github.com/microsoft/TypeScript/pull/56194

jakebailey (approved):

PR #56131, opened 2023-10-16
fix54035, extractType now allows parts of union and intersection types to be extracted

### LINKED ISSUE #54035: Extract type alias doesn't extract entire selected type
# Bug Report

### 🔎 Search Terms
- refactor 
- refactoring

5.1.0-dev.20230426

### ⏯ Playground Link

### 💻 Repo

```
type a = { a: string } | { b: string } | { c: string };
```

Select `{ b: string } | { c: string }` and run `extract type alias`

### 🙁 Actual behavior

```ts
type NewType = {
	b: string;
};

type a = { a: string } | NewType | { c: string };
```

### 🙂 Expected behavior

```ts
type NewType =  { b: string } | { c: string }

type a = { a: string } | NewType;
```

iisaduan:
Fixes #54035

### DIFF
--- a/src/services/refactors/extractType.ts
+++ b/src/services/refactors/extractType.ts
@@ -61,6 +61,7 @@ import {
     SymbolFlags,
     textChanges,
     TextRange,
+    toArray,
     TypeChecker,
     TypeElement,
     TypeNode,
@@ -237,18 +238,13 @@ function flattenTypeLiteralNodeReference(checker: TypeChecker, selection: TypeNo
     return undefined;
 }
 
-function rangeContainsSkipTrivia(r1: TextRange | TextRange[], node: Node | Node[], file: SourceFile): boolean {
-    const r1Range = arrayToNode(r1);
-    const nodeRange = arrayToNode(node);
-    return rangeContainsStartEnd(r1Range, skipTrivia(file.text, nodeRange.pos), nodeRange.end);
-
-    function arrayToNode(rangeOrArray: TextRange | TextRange[]): TextRange {
-        return isArray(rangeOrArray) ? { pos: rangeOrArray[0].pos, end: rangeOrArray[rangeOrArray.length - 1].end } : rangeOrArray;
-    }
+function rangeContainsSkipTrivia(r1: TextRange, node: TextRange, file: SourceFile): boolean {
+    return rangeContainsStartEnd(r1, skipTrivia(file.text, node.pos), node.end);
 }
 
 function collectTypeParameters(checker: TypeChecker, selection: TypeNode | TypeNode[], enclosingNode: Node, file: SourceFile): TypeParameterDeclaration[] | undefined {
     const result: TypeParameterDeclaration[] = [];
+    const selectionRange = isArray(selection) ? { pos: selection[0].pos, end: selection[selection.length - 1].end } : selection;
     if (isArray(selection)) {
         selection.forEach(t => {
             if (!visitor(t)) return undefined;
@@ -266,11 +262,11 @@ function collectTypeParameters(checker: TypeChecker, selection: TypeNode | TypeN
                     if (isTypeParameterDeclaration(decl) && decl.getSourceFile() === file) {
                         // skip extraction if the type node is in the range of the type parameter declaration.
                         // function foo<T extends { a?: /**/T }>(): void;
-                        if (decl.name.escapedText === typeName.escapedText && rangeContainsSkipTrivia(decl, selection, file)) {
+                        if (decl.name.escapedText === typeName.escapedText && rangeContainsSkipTrivia(decl, selectionRange, file)) {
                             return true;
                         }
 
-                        if (rangeContainsSkipTrivia(enclosingNode, decl, file) && !rangeContainsSkipTrivia(selection, decl, file)) {
+                        if (rangeContainsSkipTrivia(enclosingNode, decl, file) && !rangeContainsSkipTrivia(selectionRange, decl, file)) {
                             pushIfUnique(result, decl);
                             break;
                         }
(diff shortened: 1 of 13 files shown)


src/services/refactors/extractType.ts
```diff
@@ -302,17 +350,33 @@ function doInterfaceChange(changes: textChanges.ChangeTracker, file: SourceFile,
...
     changes.insertNodeBefore(file, enclosingNode, ignoreSourceNewlines(newTypeNode), /*blankLineBetween*/ true);
-    changes.replaceNode(file, selection, factory.createTypeReferenceNode(name, typeParameters.map(id => factory.createTypeReferenceNode(id.name, /*typeArguments*/ undefined))), { leadingTriviaOption: textChanges.LeadingTriviaOption.Exclude, trailingTriviaOption: textChanges.TrailingTriviaOption.ExcludeWhitespace });
+
+    if (isArray(selection)) {
+        changes.replaceNodeRange(file, selection[0], selection[selection.length - 1], factory.createTypeReferenceNode(name, typeParameters.map(id => factory.createTypeReferenceNode(id.name, /*typeArguments*/ undefined))), { leadingTriviaOption: textChanges.LeadingTriviaOption.Exclude, trailingTriviaOption: textChanges.TrailingTriviaOption.ExcludeWhitespace });
+    }
+    else {
+        changes.replaceNode(file, selection, factory.createTypeReferenceNode(name, typeParameters.map(id => factory.createTypeReferenceNode(id.name, /*typeArguments*/ undefined))), { leadingTriviaOption: textChanges.LeadingTriviaOption.Exclude, trailingTriviaOption: textChanges.TrailingTriviaOption.ExcludeWhitespace });
+    }
```
andrewbranch, line 359: Same here, can probably use `replaceNodeRange` for both cases and cut down on the duplication

src/services/refactors/extractType.ts
```diff
@@ -302,17 +350,33 @@ function doInterfaceChange(changes: textChanges.ChangeTracker, file: SourceFile,
...
+    if (isArray(selection)) {
+        selection.forEach(typeNode => {
+            setEmitFlags(typeNode, EmitFlags.NoComments | EmitFlags.NoNestedComments);
+        });
+        newTypeValue = isUnionTypeNode(selection[0].parent) ? factory.createUnionTypeNode(selection) : factory.createIntersectionTypeNode(selection);
+    }
+    else {
+        setEmitFlags(selection, EmitFlags.NoComments | EmitFlags.NoNestedComments);
+        newTypeValue = selection;
+    }
```
andrewbranch, line 375: Possibly same, you could `toArray(selection).forEach(...)`

src/services/refactors/extractType.ts
```diff
@@ -339,7 +403,13 @@ function doTypedefChange(changes: textChanges.ChangeTracker, context: RefactorCo
...
         changes.insertNodeBefore(file, enclosingNode, jsDoc, /*blankLineBetween*/ true);
     }
-    changes.replaceNode(file, selection, factory.createTypeReferenceNode(name, typeParameters.map(id => factory.createTypeReferenceNode(id.name, /*typeArguments*/ undefined))));
+
+    if (isArray(selection)) {
+        changes.replaceNodeRange(file, selection[0], selection[selection.length - 1], factory.createTypeReferenceNode(name, typeParameters.map(id => factory.createTypeReferenceNode(id.name, /*typeArguments*/ undefined))));
+    }
+    else {
+        changes.replaceNode(file, selection, factory.createTypeReferenceNode(name, typeParameters.map(id => factory.createTypeReferenceNode(id.name, /*typeArguments*/ undefined))));
+    }
```
andrewbranch, line 412: Same 😄

src/services/refactors/extractType.ts
```diff
@@ -205,21 +228,33 @@ function flattenTypeLiteralNodeReference(checker: TypeChecker, node: TypeNode |
...
-function rangeContainsSkipTrivia(r1: TextRange, node: Node, file: SourceFile): boolean {
-    return rangeContainsStartEnd(r1, skipTrivia(file.text, node.pos), node.end);
+function rangeContainsSkipTrivia(r1: TextRange | TextRange[], node: Node | Node[], file: SourceFile): boolean {
+    const r1Range = arrayToNode(r1);
+    const nodeRange = arrayToNode(node);
+    return rangeContainsStartEnd(r1Range, skipTrivia(file.text, nodeRange.pos), nodeRange.end);
+
+    function arrayToNode(rangeOrArray: TextRange | TextRange[]): TextRange {
+        return isArray(rangeOrArray) ? { pos: rangeOrArray[0].pos, end: rangeOrArray[rangeOrArray.length - 1].end } : rangeOrArray;
+    }
```
andrewbranch, line 247: Are you sure these changes ended up being needed? I don’t see any change to the caller. It looks like it’s always just passing a `TextRange`.

andrewbranch (changes requested):

src/services/refactors/extractType.ts
iisaduan: In `visitor` in `collectTypeParameters`, this function was called with a reference to an un-narrowed `selection` as `r1`

iisaduan: After some more thought, this code will only ever be converting `selection` into a `TextRange`, it might be better to do that once, and pass that range into `rangeContainsSkipTrivia`, as I don't see why `node` needs to be a `Node`

gabritto:
@typescript-bot pack this

gabritto:
What happens if the selection doesn't fully enclose the types in a union?
I tried out these cases:
```ts
type a = { a: string /*1*/} | { b: string }/*2*/ | { c: string };
```
and it becomes this:
```ts
type NewType = {
  a: string;
};

type a = NewType | { b: string } | { c: string };
```
and
```ts
type a = { a: st/*1*/ring } | { b: string }/*2*/ | { c: string };
```
which becomes this:
```ts
type NewType = string;

type a = { a: NewType } | { b: string } | { c: string };
```

Those surprised me a bit, but tbh I don't know how much we care about this or how much people are annoyed or not by this behavior.

andrewbranch:
When @navya9singh implemented partial selection ranges for move to file, I think the logic was basically that the selection shrinks over whitespace and otherwise grows to encompass a valid set of nodes to move. It would be nice to have similar behavior between all selection-based refactors. So for example

```ts
type a = { a: string }/*1*/ | { b: string } | /*2*/{ c: string }; // extract b
type a = { a: string /*1*/} | { b: string }/*2*/ | { c: string }; // extract a and b
type a = { a: st/*1*/ring } | { b: string }/*2*/ | { c: string }; // extract a and b
type a = { a: st/*1*/ring }/*2*/ | { b: string } | { c: string }; // extract a
type a = { a: st/*1*/ring /*2*/} | { b: string } | { c: string }; // extract string
```

is probably what I’d expect

iisaduan:
The original behavior didn't check for a lot of these imperfect selection cases, and as a result there is some weird behavior. For example 
```
type A = /*1*/{ a: string } & { b: string } & { c: string }/*2*/ ;  // extracts a, b, c
type A = {/*1*/ a: string } & { b: string } & { c: string }/*2*/ ;  // extracts a
type A = { a: string } /*1*/& { b: string } & { c: string } ;/*2*/  // extracts a, b, c
type A = { a: st/*1*/ring }/*2*.; // extracts string
```
are all things that happen with the original behavior.

>Those surprised me a bit, but tbh I don't know how much we care about this or how much people are annoyed or not by this behavior.

> It would be nice to have similar behavior between all selection-based refactors. 

I was also surprised by this behavior, but I was also surprised that this partial type extraction bug wasn't noticed sooner. I think it would be really nice to have the behavior do be less finicky for the user and be more consistent across refactors, but again, not sure how much it's been noticed.

gabritto:
We might want to check with Matt, since he opened the original issue, to get a sense of how much this partial selection behavior matters.

iisaduan:
After discussion with Matt today, and taking another look at the code that picks the nodes from the selection, I think that small improvements such as 
``` 
// type A = { a: str/*1*/ing } | { b: string } | { c: string }/*2*/ ;  // extracts string
```
becoming
``` 
// type A = { a: str/*1*/ing } | { b: string } | { c: string }/*2*/ ;  // extracts a, b, c
```
will require adjustments to the way I did the bug fix here. The existing code assumes the start of the selection is well placed, and my fix uses that assumption.

iisaduan:
The most recent change only addresses the behavior discussed above, and doesn't require such a big change like I had originally expected.

src/services/refactors/extractType.ts
```diff
@@ -205,21 +233,28 @@ function flattenTypeLiteralNodeReference(checker: TypeChecker, node: TypeNode |
...
     return rangeContainsStartEnd(r1, skipTrivia(file.text, node.pos), node.end);
 }
 
-function collectTypeParameters(checker: TypeChecker, selection: TypeNode, enclosingNode: Node, file: SourceFile): TypeParameterDeclaration[] | undefined {
+function collectTypeParameters(checker: TypeChecker, selection: TypeNode | TypeNode[], enclosingNode: Node, file: SourceFile): TypeParameterDeclaration[] | undefined {
     const result: TypeParameterDeclaration[] = [];
+    const selectionRange = isArray(selection) ? { pos: selection[0].pos, end: selection[selection.length - 1].end } : selection;
+    if (isArray(selection)) {
+        selection.forEach(t => {
+            if (!visitor(t)) return undefined;
```
gabritto, line 254: I'm not sure I understand what's going on here: if `visitor` returns true, that means we want to skip collection of type parameters and so we want `collectTypeParameters` to return `undefined`, right?
It seems to me we're not doing that anymore here when `selection` is an array, since inside this `if` we're always returning the `result` array.

gabritto (approved):
Left one small question, but otherwise looks good to me

gabritto (approved):

PR #56132, opened 2023-10-16
During fourslash server test baseline tsserver log file so that project updates and watches can be baselined and reasoned about

sheetalkamat:
I forgot to cherry pick this in #55814 from #54785 

This baselines tsserver long when running fourslash-server tests.
What this helps with is that project updates that happen get baselined and helps with debugging (eg Have failing test as part of c2dbfe4 which i can investigate easily with this)
The other thing it helps with is the watches that get created: We would be more mindful of watches that get created as part of particular scenario and handle those.


jakebailey:
Wow, that's a lot of baselines 😅 

Do you have an example of something that's caught by this?

Should we be removing or deprecating the old methods for testing things that aren't baselines?

sheetalkamat:
Its 170 baselines for 170 fourslash server tests. (others are just adding name of the test performed to the baseline)

Eg https://github.com/microsoft/TypeScript/blob/aa3c470b6b3d4b7ae0973f179c0ed64b6b391b25/tests/baselines/reference/tsserver/fourslashServer/goToSource16_callbackParamDifferentFile.js#L268C1-L270C36 @andrewbranch  and myself discussed before if we should really be watching these files and stuff for noDts project. (we think we shouldnt be i havent gotten around to looking into this)

sheetalkamat:
> Should we be removing or deprecating the old methods for testing things that aren't baselines?

Its already done for the references tests.. this is just making all the baselining commands be able to aggregate rather than just few selected and through special command.

jakebailey (approved):

jakebailey (approved):

PR #56150, opened 2023-10-18
Properly handle AsyncGeneratorUnwrapYieldResumption

### LINKED ISSUE #54768: Lowered async generator `return` function doesn't resolve promise argument
# Bug Report

### 🔎 Search Terms

async generator return promise

- This is the behavior in every version available in the playground

### ⏯ Playground Link

[Playground link with relevant code](https://www.typescriptlang.org/play?#code/PQgEB4CcFMDNpgOwMbVAGwJYCMC8AiaAZ0WgA8AXfUYAPgCgBDIgTxVFgFcULMB7RACoOiABQBKUAG96oUC0zR0AE2EBtAIwBdWaBgVOkRKADM9AL5NW7Lj37GKxChOm7kAohVCYvuERLcPPnRoADp0PgBzUUYAd0YfbwpQ0koJcUDEImCwiOi4hK8fUP1DMQAFSD4AW0wiMJhs9AA3aFEAJnEui3pHTwCgA)

### 💻 Code

```ts
/// <reference lib="esnext" />
async function* fn() {
  yield* [1]
  return 3
}
async function test() {
  const it = fn()
  console.log(await it.next())
  console.log(await it.return(Promise.resolve(2)))
}
test()
```

### 🙁 Actual behavior

When transformed by TypeScript to ES2017, this code prints the following:

```
{value: 1, done: false}
{value: Promise, done: true}
```

### 🙂 Expected behavior

When run natively, this code prints the following:

```
{value: 1, done: false}
{value: 2, done: true}
```

I discovered this bug when implementing async generator transformation for esbuild. I was [running relevant test262 tests](https://github.com/evanw/esbuild/blob/d7c028d7cb56750ad00fece941a2b9c1917694ca/scripts/test262-async.js) through both TypeScript and esbuild's transformations and comparing the differences. There are other TypeScript-only test failures as well but they mostly involve weird uses of the API where getters throw errors, which doesn't seem like something that would come up in practice. So I'm only reporting this bug because this seems like more of an important correctness issue. Let me know if you'd like to know about the other test failures too.

rbuckton:
This updates the `__asyncGenerator` helper to more correctly observe [`AsyncGeneratorUnwrapYieldResumption`](https://tc39.es/ecma262/#sec-asyncgeneratorunwrapyieldresumption) with regards to the `return` method of an async generator.

Fixes #54768

### DIFF
--- a/src/compiler/factory/emitHelpers.ts
+++ b/src/compiler/factory/emitHelpers.ts
@@ -854,17 +854,18 @@ export const asyncGeneratorHelper: UnscopedEmitHelper = {
     scoped: false,
     dependencies: [awaitHelper],
     text: `
-            var __asyncGenerator = (this && this.__asyncGenerator) || function (thisArg, _arguments, generator) {
-                if (!Symbol.asyncIterator) throw new TypeError("Symbol.asyncIterator is not defined.");
-                var g = generator.apply(thisArg, _arguments || []), i, q = [];
-                return i = {}, verb("next"), verb("throw"), verb("return"), i[Symbol.asyncIterator] = function () { return this; }, i;
-                function verb(n) { if (g[n]) i[n] = function (v) { return new Promise(function (a, b) { q.push([n, v, a, b]) > 1 || resume(n, v); }); }; }
-                function resume(n, v) { try { step(g[n](v)); } catch (e) { settle(q[0][3], e); } }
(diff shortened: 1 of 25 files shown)


rbuckton:
An update to tslib will follow shortly.

jakebailey (approved):

PR #10749, opened 2016-09-07
Allow Infinity and NaN to be used as an Enum property identifier

fabiancook:
Fixes #6153

Issue #6347, opened 2016-01-04
Suggestion Backlog Slog, 1/4/2016


RyanCavanaugh:
First slog of 2016 :tada:

Some old ones :older_man: 
* #1166 Allow `import` of identifiers
* #1778 Allow type aliases of strings to be index signature parameters
* #1796 Reduce error noise when class incorrectly implements interface
* #2175 Generic type argument defaults
* #2310 Use the 'void functions are constructors' in assignability (as they are in `new` expression checking)
* #2607 Provide a way for decorator functions to be generic on their targets/values

Some new ones :baby:
* #6326 Allow functions returning `Foo|void` to lack a `return` statement
* #6199 Contextually type `y` in `x, y` and `x && y`
* #6154 Allow properties to override prototype methods
* #6153 Allow numeric-named (e.g. `Infinity`) keys in `const enum`s
* #5850 Allow `Array#filter`'s predicate to return any value (not just `boolean`)
* #4669 Reconsider abstract properties

RyanCavanaugh:
* #1166 Allow `import` of identifiers
  * Allow, but enforce that all meanings of the identifier come from the same place
  * :+1: - the first symbol meaning we resolve is the one used
* #1778 Allow type aliases of strings to be index signature parameters
  * :-1: this implies semantics we don't have yet. Rethink when if/we have 'fresh' types
* #1796 Reduce error noise when class incorrectly implements interface
  * Rejected due to implementation complexity
* #2175 Generic type argument defaults
  * Anders to investigate
* #2310 Use the 'void functions are constructors' in assignability (as they are in `new` expression checking)
  * :-1: ES6 is the future; current behavior in `new` invocations is an ugly hack
* #2607 Provide a way for decorator functions to be generic on their targets/values
  * Get @rbuckton to explain this one

* #6326 Allow functions returning `Foo|void` to lack a `return` statement
  * :+1:
* #6199 Contextually type `y` in `x, y` and `x && y`
  * :+1:
* #6154 Allow properties to override prototype methods
  * Still no answers on why this error exists.
  * :+1: pending check-in with @bterlson
* #6153 Allow numeric-named (e.g. `Infinity`) keys in `const enum`s
  * :+1: Ignore a leading minus and the first character must be a digit for this to be illegal
* #5850 Allow `Array#filter`'s predicate to return any value (not just `boolean`)
  * Check all lib.d.ts uses of `boolean` and see if they're actually truthiness checks
  * Use `any` instead
[...]

Issue #6153, opened 2015-12-18
const enum should allow "Infinity" as a valid name


speigg:
```typescript
	declare const enum CameraDeviceFocusMode {
		Normal = 0,
		TriggerAuto = 1,
		ContinuousAuto = 2,
		Infinity = 3, // <- error: An enum cannot have a numeric name
		Macro = 4
	}
```

RyanCavanaugh:
This is intentional because `Infinity` is also a valid enum *value*; we disallow any name that might be a value because the lookup object only works if no names are values and no values are names.

Recommended fix is to just use `Infinite` instead of `Infinity`

speigg:
What is the lookup object exactly? Using the name `Infinity` as a property key is completely valid in javascript by the way.

RyanCavanaugh:
I missed that this enum was `const`. We don't need to enforce this rule for those.

RyanCavanaugh:
(Lookup object, see http://stackoverflow.com/a/28818850/1704166)

speigg:
Not enforcing that rule for `const` sounds great.

RyanCavanaugh:
Approved even for non-`const` enums.

The suggested implementation was to skip any leading `-`, then see if the next (or first) character is a digit. Only names matching this rule should be subject to restriction. This should allow `Infinity` and `NaN` but still disallow names like `5`.

DanielRosenwasser:
So what's the new rule now for numeric names? Same as before but make a special exception for "Infinity" and "NaN"?

Also @jmatthiesen

RyanCavanaugh:
> Same as before but make a special exception for "Infinity" and "NaN"?

Correct (as well as `-Infinity`)

YuichiNukiyama:
@RyanCavanaugh [A JavaScript identifier must start with a letter, underscore (_), or dollar sign ($);](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Grammar_and_types#Variables). So,  I think `-Infinity` is invalid.

RyanCavanaugh:
Enum members need not be identifiers - they can be quoted much like the property names in object literals

jbondc:
Reported here too #3572

PR #56161, opened 2023-10-20
Disallow `Infinity`, `-Infinity` and `NaN` as an enum key name

### LINKED ISSUE #48956: Enum with special numbers overwrites own value
# Bug Report

### 🔎 Search Terms

Enum number computed overwrite Infinity NaN

I noticed on 4.6.2, but using TS Playground am able to reproduce on earliest possible version (3.3.3). Also still occurs on nightly build.

### ⏯ Playground Link

[Playground link with relevant code](https://www.typescriptlang.org/play?ts=4.6.2#code/PQKhCgAIUgVALAlgZ0gUwHYFcC2kDuA9gE4DWqAhqmgB4AOaAxgC5oAmUIw4mukAovSas2kAN5RIUgGIB5WZAC8kAOQAzQoRUAaSVIBCAQQBKSyAEZdUqSsgBFAKqzY-ACKqzKx87crwAX3BwRkIMZEIAGzQAOgjCAHMACkEGFnZouQVFbNUNLQBKAG5IYGBIZmIsNGDQ8KjYhOShNLZoo1Ns5XMikrKKqpqwyJi4pJThdgBtW28XdxUAXSUcryc5lR7S8srqwbqRxvGWjPll5RUAZUIcNEhCZng0YkhkCsQMeI3irbUKCOQ0NpIBQMKJmABPBjIRjERB0ZgECjEMJ7YYNMbNERtExnSAAJk2ZV+-0BwNB5UhaGhsPhiORyCCoAg0DgSGo2DwbEIVMgGHuBBIpGB1Ex7CsMBBogoLCwfwi4LuADcnvhYaxUIhmKhCPgMJBFX8qshopxuCEwgjCAAjABWwgAMpqnn8zGJ-IUeBzIABha50LAicSSACSGDU701CuUofDGEjVkgADkKImzMnEwmVABaPHRPF4gAMABYiwA2YsATjxAFYCwBmcx1tBZ8yl2zKXP54tlys1+uN5uthNbWAXSAAdSRYUgGmeDxQ+qeyEQoUgiStAd5-NnwIiEUXxGXtWA5vD8WQ+SBG4Rr0Qe8gIRwdDvVMkW19OBuGEtm+YhEgxBoDghDKnScYfJIKiTNadosJAsi2sICztnciEsI6rDEH8CbYPeyi4REASovUoyJB+-pYjGEYQriVFxhC+S9DOfwAsRBxJORAbpHRka4ioPEQhsTH9LciRoMqeoPIQWDxPAkCag+foRBQ7wXkE5r7OiZF+lxrTpri6aMT8LHVBpaKkZxWL6Z0qjpkJWwiWu4mYOU8DSbJ8kIo+dDKap+TqbU5mNJZUxQTBwjwWhzDIUsNnhehTpYRERlEiZbFaSFbDTNBUWRbB0WLFINlhblCH5chKXbFUTkSa57lyQp3m+WE-npRZOlYgRuIEZVxKsWZJHBR16RdcVBH2X0Ow1S5UkyQ1XlKSpLVAA)

### 💻 Code

```ts
enum Computed {
  Infinity = Infinity,
  NaN = NaN
}
```

### 🙁 Actual behavior

Typescript generates the following code for that enum using a "lookup object":
```ts
var Computed;
(function (Computed) {
    Computed[Computed["Infinity"] = Infinity] = "Infinity";
    Computed[Computed["NaN"] = NaN] = "NaN";
})(Computed || (Computed = {}));
```

As a result, the Computed enum actually has values of Strings, rather than the "computed" numbers as expected
```ts
console.log(Computed);
// { Infinity: 'Infinity', NaN: 'NaN' }
```

Note the "lookup object" behavior itself is not unexpected, but the side effect of overwriting keys is unexpected.

### 🙂 Expected behavior

Typescript should not overwrite an existing key with a new value. Most likely that would mean disabling the "lookup values" for computed.

```ts
console.log(Computed);
// { Infinity: Infinity, NaN: NaN }
```

### Extra Discussion

This does not impact `const enum` because values like `Infinity` are considered "computed".

It is also possible to do something like
```ts
enum Computed {
  '[object Object]' = {},
  null = null
}
```
with similar results. 
The latest versions of typescript warn at this behavior, but it is possible to compile and I don't think it warns on _all_ versions. But I omitted from core issue for that reason.

Similarly
```ts
enum Overwrite {
  123 = 'FOO',
  BAR = 123,
}
```
has similar behavior (when compiled) but TS does warn about numeric names (I believe this warning occurs in all versions).

It does not occur when using string enums, as those never write "lookup objects"
```ts
enum NoOverwrite {
  FOO = 'SOME VAL',
  BAR = 'FOO',
  BAZ = 123
}
console.log(NoOverwrite);
// { 123: "BAZ", FOO: "SOME VAL", BAR: "FOO", BAZ: 123 }
```

Effectively this issue happens whenever the `String(val)` (generally `.toString()` internally) returns a key in the Enum.

Possible solutions:
* Don't emit "lookup object"
  * I'm going to assume the lookup object behavior is intentional/beneficial for literal numbers, but perhaps not for computed... Although values like `1 + 2` are "computed" but are resolved at compile time. So maybe "computed" where the value is anything but a number literal.
* Forbid special "numeric" names
  * Don't allow `NaN`, `Infinity` or `-Infinity` as an enum key name. There may be others I am not aware of right now... but I was not able to reproduce error with exponential notation like `Number.EPSILON` ("An enum member cannot have a numeric name.")
* Don't allow computed values that are non numeric literals
  * Throw compile-time error when value is something like Infinity, but allow literals.
    ```ts
    enum Allowed {
        FOO = 1 + 2
    }
    enum NotAllowed {
        BAR = 1 / 0
    }
    ```

magic-akari:
- Fixes #48956

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -44729,7 +44729,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         }
         else {
             const text = getTextOfPropertyName(member.name);
-            if (isNumericLiteralName(text) && !isInfinityOrNaNString(text)) {
+            if (isNumericLiteralName(text)) {
                 error(member.name, Diagnostics.An_enum_member_cannot_have_a_numeric_name);
             }
         }
--- a/tests/baselines/reference/enumWithInfinityProperty.errors.txt
+++ b/tests/baselines/reference/enumWithInfinityProperty.errors.txt
@@ -1,0 +1,10 @@
+enumWithInfinityProperty.ts(2,5): error TS2452: An enum member cannot have a numeric name.
+
+
+==== enumWithInfinityProperty.ts (1 errors) ====
+    enum A {
+        Infinity = 1
+        ~~~~~~~~
+!!! error TS2452: An enum member cannot have a numeric name.
+    }
+    
--- a/tests/baselines/reference/enumWithNaNProperty.errors.txt
+++ b/tests/baselines/reference/enumWithNaNProperty.errors.txt
@@ -1,0 +1,10 @@
+enumWithNaNProperty.ts(2,5): error TS2452: An enum member cannot have a numeric name.
+
+
+==== enumWithNaNProperty.ts (1 errors) ====
+    enum A {
+        NaN = 1
+        ~~~
+!!! error TS2452: An enum member cannot have a numeric name.
+    }
+    
--- a/tests/baselines/reference/enumWithNegativeInfinityProperty.errors.txt
+++ b/tests/baselines/reference/enumWithNegativeInfinityProperty.errors.txt
@@ -1,0 +1,10 @@
+enumWithNegativeInfinityProperty.ts(2,5): error TS2452: An enum member cannot have a numeric name.
+
+
+==== enumWithNegativeInfinityProperty.ts (1 errors) ====
+    enum A {
+        "-Infinity" = 1
+        ~~~~~~~~~~~
+!!! error TS2452: An enum member cannot have a numeric name.
+    }
+    


sandersn (approved):

jakebailey:
I can't recall; was banning this actually something we wanted? Can't remember what we said in the design meeting, or maybe we didn't talk about it at all?

sandersn:
My memory is that we briefly discussed this and said "yes of course you can't do that because of the lookup object", but I could be wrong.

We did discuss Infinity *types* a couple of weeks after your comment, but that's the only mention of Infinity and NaN in the design notes. And then a couple of days after that, we discussed enum assignability. So it sounds like my memory smudged the two discussions together and in fact we never did.

I still think this is a good thing to forbid since the emit is bad and it's unlikely to be something people want to write.

jakebailey:
My impression was that this PR was good, as they're not good enum names, but that the recent `infinity` chat was about the whole `1e999` thing allowing a half-functional backdoor into gaining an infinity type, which we banned but then backed out and decided that we were instead going to look into introducing those literal types into the type system instead.

jakebailey (approved):
Yep.

JoostK:
@DanielRosenwasser This change  is missing from breaking changes section in the [the blog post](https://devblogs.microsoft.com/typescript/announcing-typescript-5-4-beta/#breaking-changes), perhaps it can be included.

DanielRosenwasser:
I'm surprised we merged this because the logic was explicitly there, and was added 8ish years ago

- https://github.com/microsoft/TypeScript/issues/6153
- https://github.com/microsoft/TypeScript/issues/6347
- https://github.com/microsoft/TypeScript/pull/10749

Given that there's no linked design meeting notes, in the future let's have a more intentional discussion before merging a change like this.

PR #56165, opened 2023-10-20
Less template literal and string literal reduction

### LINKED ISSUE #56081: Memory Leak / Infinite Recursion. TS hangs, TSC never finishes, VSCode and Intellisense dies.
### 🔎 Search Terms

Typescript Hangs, Memory Leak, VS Code, Prisma, React-Hook-Form, 

- This is a crash

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about hanging. 

Tested on latest 5.2.2 and 5.3.0-dev-20231012. typescript@next 

### ⏯ Playground Link

https://codesandbox.io/p/sandbox/musing-turing-78vp6j?file=%2Fapp%2Fpage.tsx%3A6%2C33

### 💻 Code

```ts
import { useForm } from "react-hook-form";
import { Prisma } from "@prisma/client";

export default function Home() {
  const form = useForm<Prisma.UserCreateInput>();
  ...
```
This will hang everything. Typescript won't tsc and intellisense will just sit with "loading" no matter what you hover. 

meanwhile this works fine: 

```ts
import { useForm } from "react-hook-form";
import { Prisma } from "@prisma/client";

export default function Home() {
  const form = useForm<>();
  ...
  ```

`tsc --noEmit --extendedDiagnostics` will complete in like 3 seconds with this. 

### 🙁 Actual behavior

Typescript won't complete a `tsc --noEmit` and intellisense will just sit with "loading" no matter what you hover over. I have seen it once before just set everything to `any`

With the generated types in the first example - i let it run for 1 hour 59mins on a M1 mac and no output. 

You can see the full code example with a minimal repro on the codesandbox link. It also hangs it's intellisense in CSB and won't tsc. 
CPU also goes to 100% on both local and CSB 

### 🙂 Expected behavior

Doesn't hang typescript, operates just like if you put any other type in there say 
```ts
type SomeType = {name: string} 
const form = useForm<SomeType>() 
```

Another report by someone else here: 

https://stackoverflow.com/questions/76932600/high-cpu-usage-when-using-some-prisma-types-with-react-hook-form

It seems to be anything that comes off the Prisma.* types

PranavSenthilnathan:
Checking perf impact of the naive change.

Fixes #56081


PranavSenthilnathan:
@typescript-bot test this

jakebailey:
@typescript-bot test this

I don't think you're on the author list so that doesn't work, though maybe it's some other list that does it.

jakebailey:
What test case are you looking at for this? https://github.com/microsoft/TypeScript/issues/52345?

PranavSenthilnathan:
Updated comment. It's #56081

PranavSenthilnathan:
So this is not the solution since this was an intentional feature introduced in https://github.com/microsoft/TypeScript/pull/41276. But I think there should probably be some middle ground between reducing eagerly during union creation and lazily when subtype reduction is performed. Maybe just doing subtype reduction when checking variable declaration nodes?

jakebailey:
The main problem is that if we don't always reduce, then two unions that are actually the same thing can be present at once, which means you can't fast check for equality and such (maybe some other invariants).

PR #38839, opened 2020-05-29
Type guard by deep property

ShuiRuTian:
Fixes #32399 
Fixes #18758

support `if` clause,  `switch` clause and `??` to narrow type by nest property for 
`Typeof`, `Truthiness`, `Discriminant`

---

There some limitations described here https://github.com/microsoft/TypeScript/pull/38839#issuecomment-902383213

And we need to do more things described here as following PR https://github.com/microsoft/TypeScript/pull/38839#issuecomment-973769448

PR #56173, opened 2023-10-22
Add Support for Using Aliased Discriminants in Conditional Statements

### LINKED ISSUE #55577: Type inference fails when using combined boolean variable in if condition
### 🔎 Search Terms

"typescript destructuring losses types", "typescript destructuring losses type object is possibly undefined", "typescript type alias destructuring loses type", "typescript inference with union types"

### ⏯ Playground Link

https://www.typescriptlang.org/play?#code/C4TwDgpgBAqgzhAigVwgJxAJQnZAbYAHgBUA+KAXigG8AoKBqASzgGVkBjDnOALigBmAQzwIA3PUYATIcCH9kAOykQBTRRCkSAvlAA+NSQxbsuPfsDSoJjKDLn9itbRNoClHYEwD2iqMgQUdBAACgBKfngkVAxsXAJCRWQAWwAjdHI6WzQIYGQ0PyzbRhNObjg+QREEABojYvt5f2VVdU062xdnWloOXzhgGjtZJsaARhrmNjLzKdNyuDGoXSoA6ODwiT7FAaHG-kaAJknSswr+U4XD5cp-QJjQsK3+wephh3ehAGYT6bPKy48L43Vb3DZPHoAekhUAAkoMAO7eNAAazgUCE6IgAA9IJ5NLQmAIoCFARUlgAyClzGYVa5Umn-L5hQy2cYAOmA3gAorjfBBFF4RJt6kdOTy+RpBUxhRC2SMvuLeWB+dLZToettdkIcvMeLcyYsoAzDfTqYavq5oXDBlJvDhFAByRHIlFQAAW6AgkwR7qYeGgTGAjvRACsAoMhFA+ml1LIfH5vMSQ7SEHBHYTiSEdRA9RUWUUGByucrVUK8CL5XJDkrJQLy5XpArayqpQ2IdogA

### 💻 Code

```ts
type UseQueryResult<T> = {
    isSuccess: false;
    data: undefined;
} | {
    isSuccess: true;
    data: T
};

function useQuery(): UseQueryResult<number> {
    return {
        isSuccess: false,
        data: undefined,
    };
}

const { data: data1, isSuccess: isSuccess1 } = useQuery();
const { data: data2, isSuccess: isSuccess2 } = useQuery();
const { data: data3, isSuccess: isSuccess3 } = useQuery();

// It works as expected
if (isSuccess1 && isSuccess2 && isSuccess3) {
    data1.toExponential();
    data2.toExponential();
    data3.toExponential();
}

const areSuccess = isSuccess1 && isSuccess2 && isSuccess3;

// It doesn't work here, while it's just a combination of 'successes'
if (areSuccess) {
    data1.toExponential();
    data2.toExponential();
    data3.toExponential();
}
```

### 🙁 Actual behavior

The issue occurs when using the combined boolean variable `areSuccess` in the second if condition. Although `isSuccess1`, `isSuccess2`, and `isSuccess3` are all true, TypeScript cannot narrow down the types of `data1`, `data2`, and `data3` within the if block, leading to a type error

### 🙂 Expected behavior

`areSuccess` is a combination of the individual `isSuccess` variables, it should correctly infer the types of `data1`, `data2`, and `data3` as the specific data type (number) defined for the queries. This should allow the code to call the toExponential method on each data variable without any issues.

_No response_

Zzzen:
Fixes #55577

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -27162,7 +27162,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 return isConstantReference((node as AccessExpression).expression) && isReadonlySymbol(getNodeLinks(node).resolvedSymbol || unknownSymbol);
             case SyntaxKind.ObjectBindingPattern:
             case SyntaxKind.ArrayBindingPattern:
-                return isVariableDeclaration(node.parent) && isVarConstLike(node.parent);
+                return isBindingElement(node.parent) ? isConstantReference(node.parent.parent) : isVariableDeclaration(node.parent) && isVarConstLike(node.parent);
         }
         return false;
     }
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -27162,7 +27162,8 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 return isConstantReference((node as AccessExpression).expression) && isReadonlySymbol(getNodeLinks(node).resolvedSymbol || unknownSymbol);
             case SyntaxKind.ObjectBindingPattern:
             case SyntaxKind.ArrayBindingPattern:
-                return isBindingElement(node.parent) ? isConstantReference(node.parent.parent) : isVariableDeclaration(node.parent) && isVarConstLike(node.parent);
+                const rootDeclaration = getRootDeclaration(node.parent);
+                return isVariableDeclaration(rootDeclaration) && isVarConstLike(rootDeclaration);
         }
         return false;
     }
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -27160,6 +27160,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             case SyntaxKind.ElementAccessExpression:
                 // The resolvedSymbol property is initialized by checkPropertyAccess or checkElementAccess before we get here.
                 return isConstantReference((node as AccessExpression).expression) && isReadonlySymbol(getNodeLinks(node).resolvedSymbol || unknownSymbol);
+            case SyntaxKind.ObjectBindingPattern:
+            case SyntaxKind.ArrayBindingPattern:
+                return isVariableDeclaration(node.parent) && isVarConstLike(node.parent);
         }
         return false;
     }
--- a/tests/baselines/reference/controlFlowAliasedDiscriminants.errors.txt
+++ b/tests/baselines/reference/controlFlowAliasedDiscriminants.errors.txt
@@ -2,9 +2,13 @@ controlFlowAliasedDiscriminants.ts(39,9): error TS18048: 'data1' is possibly 'un
 controlFlowAliasedDiscriminants.ts(40,9): error TS18048: 'data2' is possibly 'undefined'.
 controlFlowAliasedDiscriminants.ts(65,9): error TS18048: 'bar2' is possibly 'undefined'.
 controlFlowAliasedDiscriminants.ts(66,9): error TS18048: 'bar3' is possibly 'undefined'.
+controlFlowAliasedDiscriminants.ts(86,14): error TS1360: Type 'string | number' does not satisfy the expected type 'string'.
+  Type 'number' is not assignable to type 'string'.
+controlFlowAliasedDiscriminants.ts(98,19): error TS1360: Type 'string | number' does not satisfy the expected type 'string'.
+  Type 'number' is not assignable to type 'string'.
 
 
-==== controlFlowAliasedDiscriminants.ts (4 errors) ====
+==== controlFlowAliasedDiscriminants.ts (6 errors) ====
     type UseQueryResult<T> = {
         isSuccess: false;
         data: undefined;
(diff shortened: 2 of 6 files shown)


src/compiler/checker.ts
```diff
@@ -27160,6 +27160,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             case SyntaxKind.ElementAccessExpression:
                 // The resolvedSymbol property is initialized by checkPropertyAccess or checkElementAccess before we get here.
                 return isConstantReference((node as AccessExpression).expression) && isReadonlySymbol(getNodeLinks(node).resolvedSymbol || unknownSymbol);
+            case SyntaxKind.ObjectBindingPattern:
+            case SyntaxKind.ArrayBindingPattern:
+                return isVariableDeclaration(node.parent) && isVarConstLike(node.parent);
```
Andarist, line 27165: shouldn't this lead to calling `isConstantReference` recursively (perhaps accompanied by some other checks but still)? What about nested binding patterns (`const { a: { b: b2 } } = ...`)?

Zzzen: Thank you for bringing this to my attention! I was surprised to discover that TypeScript currently lacks support for nested binding patterns during cross symbol analysis. I don't know if it is intended though.

[playground](https://www.typescriptlang.org/play?useDefineForClassFields=true&target=9&module=1&noImplicitUseStrict=true&ts=5.3.0-dev.20231029#code/C4TwDgpgBAchDOwIBMoF4oG8BQU9VEgC4oByRAJwEsA7Ac1IG5d8KEwSd9upkBDYHxKVadFngC+2CVAA+WcQXAQSpGgFcAtgCMIFJorbwOCnvn6CSGnXubcpU7F3wAbCMChGwAQhJxEKHb4AMYA9jSIWJ7snLwCfFASADRKkIno0cZBeFQAZlAAFITQaKVkIvSkAJSmZnGCUPACVPC5VAiNwNT0jHgA9H2NABah6i6oNKEeehShFIpS3HmFXgB0xehl5F2i1bVma2sWCU3ALW0dFXTZidLYQA)

```ts
type Nested = {
    type: 'string';
    resp: {
        data: string
    }
} | {
    type: 'number';
    resp: {
        data: number;
    }
}

{
    let resp!: Nested;
    const { resp: { data }, type } = resp;
    if (type === 'string') {
        data satisfies string;  // should not error
    }
    if (resp.type === 'string') {
        resp.resp.data satisfies string;
    }
}
```

Andarist: This sounds like https://github.com/microsoft/TypeScript/pull/38839#issuecomment-1160929515

I was more thinking about cases like this:
```ts
type Nested =
  | {
      resp: {
        type: "string";
        data: string;
      };
    }
  | {
      resp: {
        type: "number";
        data: number;
      };
    };

declare const nested: Nested;

const {
  resp: { type, data },
} = nested;
if (type === "string") {
  data satisfies string; // should be OK
}
```

This keeps the dependency between the discriminant and another property "at the same level" but at the same time both are coming from a nested binding pattern.

src/compiler/checker.ts
```diff
@@ -27160,6 +27160,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             case SyntaxKind.ElementAccessExpression:
                 // The resolvedSymbol property is initialized by checkPropertyAccess or checkElementAccess before we get here.
                 return isConstantReference((node as AccessExpression).expression) && isReadonlySymbol(getNodeLinks(node).resolvedSymbol || unknownSymbol);
+            case SyntaxKind.ObjectBindingPattern:
+            case SyntaxKind.ArrayBindingPattern:
+                return isBindingElement(node.parent) ? isConstantReference(node.parent.parent) : isVariableDeclaration(node.parent) && isVarConstLike(node.parent);
```
Andarist, line 27165: an alternative solution for this could look smth like this:
```suggestion
                const rootDeclaration = getRootDeclaration(node.parent);
                return isVariableDeclaration(rootDeclaration) && isVarConstLike(rootDeclaration);
```

Andarist: I believe that this should work as well: [TS playground](https://www.typescriptlang.org/play?ts=5.4.0-dev.20231104#code/C4TwDgpgBAqgzhAigVwgJxAJQnZAbYAHgBUA+KAXgCgooAfKAbxttYEs4BlZAYx5zgAuKADMAhngQBuFqygATMcDHDkAO3kQRbNRHky5AXxYNmc2h258Bw4GlQHzi5cOKPahmVRHqewNgD2alDICCjoIAAUAJTC8EioGNi4BIRqyAC2AEbo5GZQaBDAyGjB+RZcvPxwQqISCAA0sgpKKiEaWjp6TR4yxlQ8QXDATC0uY2IAjA1QllU2s5XWNZNQhpQhYYlR0V4+an6BwcA4wJGME8LOUzNzy7V31XCrhnFbEcn4ROnZudFMLEGamGo2uV1aACZbksnsJHgIIWsNqEEhEYo4gSCLmCJgBmaFWWGLQkCXFIiibVEYdFUQFDEZiQokmobeErKAAMg5xPmNURXJ591xjjYIigkUZEGZcH+5QmkwAdMAAgBRAAeYCCEDU-gk6KgAHoDVA4AALAL4eRQAIAa2a1whStVGq1OrYet2huNZoteCttvtrVxTvVmt0bo9Ui9JvNlutdo8VH6QA) but the problem is that we are dealing with a pseudo-reference here. At this point, we can't easily check if the original symbol for which we are performing the narrowing is a non-reassigned parameter.

src/compiler/checker.ts
Andarist: I think that I might have accidentally fixed this in [this PR](https://github.com/microsoft/TypeScript/pull/56306). When I sat down to it I totally forgot about the conversation here 😅

src/compiler/checker.ts
Zzzen: Wish there was a 'const' modifier available for parameters and we could just check that!

Andarist: This should make that possible: https://github.com/microsoft/TypeScript/pull/56313

gabritto:
@typescript-bot run DT
@typescript-bot test top100

gabritto (approved):

Issue #56177, opened 2023-10-23
Don't suggest private fields in contexts where they are out of scope


Jamesernator:
### 🔎 Search Terms

private fields suggestions

### 🕗 Version & Regression Information

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about private fields

### ⏯ Playground Link

https://www.typescriptlang.org/play?#code/C4TwDgpgBAogHgQwLZgDYQJIDsCWwoC8UA3gFBQVQBmA9jQFxRYCuSARhAE4A0pAvgG5SpUJFiIU6ADI4A1tCLxkaaAB9xy9NjxDSAY1QIAzkY2ToZSlCPAEwHHqg4jS8wAoAbglTMIjBFggAJSMXj7QzmYqJORWlJwQwMycWFAAxLQ0TqlhvkJWfMJWNnYO1Jw0SG4QEioy8oyudXIQIVHoMXGUOFRQbk3oAHTOAxDVtdItQUGdXXEJSSlQNZoQ9RD5c4VzUAvJqVgQAO7tYyvm60GbFIWxFAkIACY0WKgg6ZmMLOxculZ6LxsnGYemANE4bmI1DoUD4jQmmFwwBmljmwAAFs5BhkYURMtdYfxhACsDZlgjCKdBlQKlVURQAPQMqAAFU471EOCwAHMoAAiKh8qABR7WCDQDF2KAYrgRUwIazMbnciA2HAvaGcD50fhXUhAA

### 💻 Code

```ts
type ExampleInit = {
    foo: number,
};

type ExampleLike = Example | ExampleInit;

class Example {
    static isExample(value: any): value is Example {
        return #foo in value;
    }

    static from(exampleLike: ExampleLike): Example {
        if (Example.isExample(exampleLike)) {
            return exampleLike;
        }
        return new Example(exampleLike);
    }

    readonly #foo: number;

    constructor({ foo }: ExampleInit) {
        this.#foo = foo;
    }
}

const example = Example.from({
    // Try typing "f" and see that there is a suggestion for #foo
});
```

### 🙁 Actual behavior

We get suggestions for private fields:

![Screenshot from 2023-10-23 15-45-37](https://github.com/microsoft/TypeScript/assets/8518303/98788e34-ce81-495a-ac9f-24fcd4c164ae)

This is particularly annoying as the private fields tend to be suggested first, so tab-completion tends to produce broken code.

### 🙂 Expected behavior

[...]

PR #56183, opened 2023-10-23
Filter out classes with private members from the contextual types for object literal nodes

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/56177

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -30204,6 +30204,18 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         );
     }
 
+    function filterContextualTypeForLiteralExpressionOfObject(node: Node, contextualType: Type) {
+        if (!isLiteralExpressionOfObject(node)) {
+            return contextualType;
+        }
+        return filterType(contextualType, t => {
+            if (!t.symbol || !(t.symbol.flags & SymbolFlags.Class)) {
+                return true;
+            }
+            return every(getPropertiesOfType(t), p => !p.valueDeclaration || !isNamedDeclaration(p.valueDeclaration) || (!isPrivateIdentifier(p.valueDeclaration.name) && !(getDeclarationModifierFlagsFromSymbol(p) & ModifierFlags.NonPublicAccessibilityModifier)));
+        });
+    }
+
     // Return the contextual type for a given expression node. During overload resolution, a contextual type may temporarily
     // be "pushed" onto a node using the contextualType property.
     function getApparentTypeOfContextualType(node: Expression | MethodDeclaration, contextFlags: ContextFlags | undefined): Type | undefined {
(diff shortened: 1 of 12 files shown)


src/services/completions.ts
```diff
@@ -5384,7 +5385,7 @@ function getApparentProperties(type: Type, node: ObjectLiteralExpression | JsxAt
 }
 
 function containsNonPublicProperties(props: Symbol[]) {
-    return some(props, p => !!(getDeclarationModifierFlagsFromSymbol(p) & ModifierFlags.NonPublicAccessibilityModifier));
+    return some(props, p => !!(getDeclarationModifierFlagsFromSymbol(p) & ModifierFlags.NonPublicAccessibilityModifier) || !!p.valueDeclaration && isNamedDeclaration(p.valueDeclaration) && isPrivateIdentifier(p.valueDeclaration.name));
```
Andarist, line 5388: likely this alone (or something close to it) would fix the completion problem (`private foo` is already filtered out from completions, only `#foo` is not) but that wouldn't fix the `contextuallyTypedParametersPositionIncludesClassWithPrivateMember` test cases

src/compiler/checker.ts
```diff
@@ -30270,34 +30282,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         return type;
     }
 
-    /**
```
jakebailey, line 30273: Bah, I guess git doesn't understand what happened here.

src/compiler/checker.ts
```diff
@@ -30270,34 +30282,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
-            // We cannot answer semantic questions within a with block, do not proceed any further
-            return undefined;
-        }
-        // Cached contextual types are obtained with no ContextFlags, so we can only consult them for
-        // requests with no ContextFlags.
-        const index = findContextualNode(node, /*includeCaches*/ !contextFlags);
-        if (index >= 0) {
-            return contextualTypes[index];
-        }
+    function getContextualTypeFromParent(node: Expression, contextFlags: ContextFlags | undefined): Type | undefined {
```
jakebailey, line 30285: Not `getContextualTypeWorker` with the real function above it?

Andarist: do u suggest a simple rename of this functions + putting this *after* the `getContextualType`?

jakebailey: Yeah, I think that would be clearer.

Andarist: done

jakebailey: Thanks; sorry to be picky! The diff seems easier to understand now.

jakebailey:
@typescript-bot test top200
@typescript-bot run dt

@typescript-bot pack this

jakebailey:
There's a crash above, but not sure if it's due to this PR or not.

Andarist:
I really doubt that this crash is related to this PR - but who knows? Could you rerun the extended test suite here before I start investigating this?

jakebailey:
@typescript-bot test top200

tests/cases/compiler/contextuallyTypedParametersPositionIncludesClassWithPrivateMember2.ts
```diff
@@ -0,0 +1,17 @@
...
+
+interface FooLike {
+  doStuff: (cb: (arg: number) => void) => void;
+}
+
+declare function useIt(arg: Foo | FooLike): void;
+
+useIt({
+  doStuff: (arg) => {},
+});
```
jakebailey, line 17: Do you mind adding the cases for just `Foo` and `FooLike`? like:

```ts
declare function useIt2(arg: Foo): void;

useIt2({
  doStuff: (arg) => {},
});

declare function useIt3(arg: FooLike): void;

useIt3({
  doStuff: (arg) => {},
});
```

Andarist: Sure thing, my dear friend - I just pushed them out :)

src/compiler/checker.ts
```diff
@@ -38810,8 +38830,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     }
 
     function checkExpressionWithContextualType(node: Expression, contextualType: Type, inferenceContext: InferenceContext | undefined, checkMode: CheckMode): Type {
+        const filteredContextualType = filterContextualTypeForLiteralExpressionOfObject(node, contextualType);
```
jakebailey, line 38833: Not just `contextualType = filterContextualTypeForLiteralExpressionOfObject(node, contextualType)`?

jakebailey (commented):
I'm a little bit suspicious of the fix itself from the PoV of just adding another layer of workers and then filtering, as compared to doing this more deeply in `getContextualType`, but I can't quite put my finger on what is better.

Andarist:
> I'm a little bit suspicious of the fix itself from the PoV of just adding another layer of workers and then filtering, as compared to doing this more deeply in getContextualType, but I can't quite put my finger on what is better.

it has been a while since I was working on this. I vaguely remember similar concerns that I had. IIRC, I do this filtering here because I can't access the information about being within "literal expression object" context at a deeper level. At the end of the day, the same operation would have to be done deeper - we get a set of values and we need to reject some. So this level works as OK as any other (as long as it doesn't introduce problems, ofc).

src/compiler/checker.ts
Andarist: I can do that, no problem. I was probably just trying to avoid parameter reassignment at the time.

Andarist:
> I'm a little bit suspicious of the fix itself from the PoV of just adding another layer of workers and then filtering, as compared to doing this more deeply in getContextualType, but I can't quite put my finger on what is better.

I think you were right. From what I can tell, this can just easily be put into `discriminateTypeByDiscriminableItems` - I pushed out this change.

PR #56190, opened 2023-10-23
Explicitly declare type=commonjs in package.json

jakebailey:
Node.js is experimenting with `.js` files having detection applied to them when `package.json` doesn't list a type; this detection appears to have a performance impact for bundles as large as ours, so it may be best to just explicitly declare that our package treats `.js` files as `commonjs`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
         "type": "git",
         "url": "https://github.com/Microsoft/TypeScript.git"
     },
+    "type": "commonjs",
     "main": "./lib/typescript.js",
     "typings": "./lib/typescript.d.ts",
     "bin": {


RyanCavanaugh (approved):

jakebailey:
Apparently this change had an effect on esbuild's behavior, for the worse.


| - | - | - | - | - |
| Packed | 5.49 MiB | 5.56 MiB | +71.00 KiB | +1.26% |
| Unpacked | 30.49 MiB | 31.03 MiB | +546.64 KiB | +1.75% |


| - | - | - | - | - |
| `lib/tsc.js` | 5.55 MiB | 5.55 MiB | +75.00 B | +0.00% |
| `lib/tsserver.js` | 8.12 MiB | 8.52 MiB | +407.05 KiB | +4.89% |
| `lib/typescript.js` | 8.59 MiB | 8.72 MiB | +139.49 KiB | +1.59% |
| `package.json` | 3.44 KiB | 3.46 KiB | +24.00 B | +0.68% |

This wasn't intended at all. Will send a revert.

PR #56193, opened 2023-10-23
Expose TypeChecker.getMergedSymbol to public (#56192)

### LINKED ISSUE #56192: Expose `getMergedSymbol` in type checker to public
### 🔍 Search Terms

"type checker", "getMergedSymbol", "merged symbol"

### ✅ Viability Checklist

### ⭐ Suggestion

While using the compiler's API in some cases (I wasn't able to find a small reproduction yet) the type checker might return "merged symbol" that it seems is not possible to resolve to the original symbol as yet. I noticed that there is an API but it is marked as internal. Is there any particular reason why this API isn't exposed yet and we should use something else to resolve merged symbols?

I also noticed that dtsBundler script from the compiler's repo is [using it via `any` hack](https://github.com/microsoft/TypeScript/blob/d9d027d0c4f5044631a0c078166b8d01f41d28b5/scripts/dtsBundler.mjs#L260-L262) which means that there is a room for an external usage of this API?

If that's fine to expose this API I'm happy to make a PR.

### 📃 Motivating Example

This feature allows to resolve merged symbols? I don't know to be honest 😄 

### 💻 Use Cases

1. What do you want to use this for? I'm maintaining https://github.com/timocov/dts-bundle-generator and I've got [an issue report](https://github.com/timocov/dts-bundle-generator/issues/265), after debugging I figured out that the tool is getting different symbols for the same node depends on where you asking the type checker to resolve it.
1. What shortcomings exist with current approaches? Its not possible to resolve some symbols to the original versions and it makes impossible to perform some operations.
1. What workarounds are you using in the meantime? No workarounds as yet, but probably will use the same `as any` hack

timocov:
Fixes #56192

I haven't updated [dtsBundler.mjs](https://github.com/microsoft/TypeScript/blob/d9d027d0c4f5044631a0c078166b8d01f41d28b5/scripts/dtsBundler.mjs#L260-L262) as it seems relies on the compiler from npm rather than local - but please let me know if you want me to update it as well.

### DIFF
--- a/src/compiler/types.ts
+++ b/src/compiler/types.ts
@@ -5043,7 +5043,7 @@ export interface TypeChecker {
     isUndefinedSymbol(symbol: Symbol): boolean;
     isArgumentsSymbol(symbol: Symbol): boolean;
     isUnknownSymbol(symbol: Symbol): boolean;
-    /** @internal */ getMergedSymbol(symbol: Symbol): Symbol;
+    getMergedSymbol(symbol: Symbol): Symbol;
 
     getConstantValue(node: EnumMember | PropertyAccessExpression | ElementAccessExpression): string | number | undefined;
     isValidPropertyAccess(node: PropertyAccessExpression | QualifiedName | ImportTypeNode, propertyName: string): boolean;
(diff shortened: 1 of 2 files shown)


jakebailey (approved):
Per design meeting, this looked okay to add.

PR #56197, opened 2023-10-24
Added special-case error for checking instanceof an arrow function

### LINKED ISSUE #56143: Unsound error while check if object instance of another
### 🔎 Search Terms

unsound, instanceof, instance, object, types, error

- This is unsound error
- All versions in playground have this bug

### ⏯ Playground Link

https://www.typescriptlang.org/play?#code/MYewdgziA2CmB00QHMAUAVAngB1gUQCcCQCACAS0gBcBDMYWEAM1NVQEpSBeAPlIGZ27IA

### 💻 Code

```ts
console.log(TypeError instanceof (() => 3));
```

### 🙁 Actual behavior

TypeScript successfully transpiles the code

### 🙂 Expected behavior

Throws error because the right-side operand has not prototype

_No response_

JoshuaKGoldberg:
Fixes #56143.

Adds a special-cased error for known arrow functions being the right-hand-side of an `instanceof` expression.

I'd wanted to also capture the case of `instanceof Object.create(null)`, but that type is `any`. Perhaps a followup issue for another day to produce a non-`any` type in that case.


RyanCavanaugh:
I'm starting to doubt why I didn't just Won't Fix the linked issue. Who would ever do this?

fatcerberus:
> I'm starting to doubt why I didn't just Won't Fix the linked issue.

I was wondering the same thing - I was fully expecting you to dismiss it as a case of "you tried to break it and succeeded" and was *quite* surprised when you marked it as a bug instead.  Especially since, as I hinted and then you confirmed [here](https://github.com/microsoft/TypeScript/issues/56143#issuecomment-1779543682), TS can't tell the difference at the type level between the arrow function and non-arrow function case.

sandersn:
So, I guess we don't want this after all? The issue is no longer marked Bug, at least.

JoshuaKGoldberg:
Closing to keep the PR queue small.

PR #56198, opened 2023-10-24
Fix incorrect flags in modifierFlagsCache

### LINKED ISSUE #42189: getSyntacticModifierFlags return value affected by getEffectiveModifierFlags
# Bug Report

I don't know if this already causes a user-facing bug, but it might in the future.

### 🔎 Search Terms

syntactic effective modifierflags

The API was introduced in #38403 by @rbuckton on 12th May 2020, I don't know which TS version that is

### 💻 Code

```ts
class C {
  /** @private */
  prop = 1;
}
```

The code snippets below assume that `node: ts.PropertyDeclaration` of property `prop`

### 🙁 Actual behavior

Note: I'm not talking about user code using compiler API as these APIs are not public. I'm talking about the compiler implementation itself.

If the above code is a TS file:

* call `ts.getSyntacticModifierFlags(node)`, return `ts.ModifierFlags.None` (expected)
* call `ts.getEffectiveModifierFlags(node)`, return `ts.ModifierFlags.None` (expected)
* call `ts.getSyntacticModifierFlags(node)` again, returns `ts.ModifierFlags.None` (expected)

If the above code is a JS file:

* call `ts.getSyntacticModifierFlags(node)`, return `ts.ModifierFlags.None` (expected)
* call `ts.getEffectiveModifierFlags(node)`, return `ts.ModifierFlags.Private` (expected)
* call `ts.getSyntacticModifierFlags(node)` again, returns `ts.ModifierFlags.Private` (**not** expected)

(see https://runkit.com/ajafff/5ff204e2ac3559001a1407f7)

This is because both use the same property to cache the result. `getEffectiveModifierFlags` adds modifiers from JSDoc to the cached result. Calling `getSyntacticModifierFlags` afterwards cannot remove the modifiers from JSDoc, so it simply returns them too.

### 🙂 Expected behavior

Looking at the description of #38403, the intended use of `getSyntacticModifierFlags` is: "get me the syntactic modifiers and **don't** include those from JSDoc". Calling `getEffectiveModifierFlags` should not have a side-effect on `getSyntacticModifierFlags`.

Also there's `getEffectiveModifierFlagsAlwaysIncludeJSDoc` introduced in #38523 by @Kingwl. This affects TS files, making the original change in #38403 useless depending on execution order: 
* call `ts.getSyntacticModifierFlags(node)`, return `ts.ModifierFlags.None` (expected)
* call `ts.getEffectiveModifierFlags(node)`, return `ts.ModifierFlags.None` (expected)
* call `getEffectiveModifierFlagsAlwaysIncludeJSDoc`, return `ts.ModifierFlags.Private` (expected)
* call `ts.getSyntacticModifierFlags(node)` again, returns `ts.ModifierFlags.Private` (**not** expected)
* call `ts.getEffectiveModifierFlags(node)` again, returns `ts.ModifierFlags.Private` (**not** expected)

rbuckton:
Due to the overlap between syntactic and JSDoc modifiers, there exists a case where what we report as cached syntactic modifiers can change in a JS file. To avoid this conflict, this PR modifies the caching behavior for modifier flags to give JSDoc modifiers their own cache value when they overlap with a syntactic modifier. The cache-only modifiers are then converted back to what we normally produce for those modifiers so as not to break existing callers.

Fixes #42189


rbuckton:
@typescript-bot perf test

rbuckton:
@typescript-bot run dt
@typescript-bot test this

rbuckton:
(from the original issue)
> I don't know if this already causes a user-facing bug, but it might in the future.

Turns out there was a bug in `fixSpelling` that this change made apparent as we were requesting only a syntactic `override` for a code-fix when we should have been requesting the effective `override`.

rbuckton:
:( I was hoping the perf cost would be negligible, but that doesn't seem to be the case.

jakebailey:
Where does the slowdown come from? A bit hard to tell at a glance. Are we calling `getJSDocModifierFlagsNoCache` twice somehow?

One interesting bit is that the `Deprecated` modifier isn't used at all to determine whether or not a symbol is deprecated or not (in fact it's entirely unreferenced); that instead is done directly in the parser through a goofy backdoor (which I once tried to remove in an effort to make JSDoc parsing fully lazy). Maybe that fact doesn't help given people may be querying modifiers for that info externally?

rbuckton:
It's also being too aggressive in collecting JSDoc modifiers in a TS file.

rbuckton:
@typescript-bot perf test

rbuckton:
much better

src/compiler/types.ts
```diff
@@ -851,25 +851,45 @@ export const enum NodeFlags {
...
+    Const =              1 << 12, // Const enum
+    In =                 1 << 13, // Contravariance modifier
+    Out =                1 << 14, // Covariance modifier
+    Decorator =          1 << 15, // Contains a decorator.
+
+    // JSDoc-only modifiers
+    Deprecated =         1 << 16, // Deprecated tag.
+
+    // Cache-only JSDoc-modifiers. Should match order of Syntactic/JSDoc modifiers, above.
+    /** @internal */ JSDocPublic = 1 << 23, // if this value changes, `adjustJSDocModifierFlags` must change accordingly
```
jakebailey, line 879: `adjustJSDocModifierFlags` doesn't exist anymore; is there something that needs to be updated?

Do these flags turn back into the public ones for public consumption somewhere?

rbuckton: Sorry, that should be `selectEffectiveModifierFlags`, and yes that's what `selectEffectiveModifierFlags` does.

jakebailey: Aha, yes! Thanks!

jakebailey:
This seems good to me, though is it intended that DT still fail? It seems odd that we'd hard error on a JSDoc comment in TS code like this.

rbuckton:
> This seems good to me, though is it intended that DT still fail? It seems odd that we'd hard error on a JSDoc comment in TS code like this.

That failure may be related to the change in the first approach, I need to retest to see if it still occurs in the new approach.

jakebailey (approved):
LGTM, I don't _think_ this is breaking given nobody should be depending on the exact values anyway.

I'm half worried that this PR appears to entirely fill up the entirety `ModifierFlags`'s bitwise space, but I'm not sure how we could do this different other than to have another flag prop or something.

rbuckton:
> LGTM, I don't _think_ this is breaking given nobody should be depending on the exact values anyway.
> 
> I'm half worried that this PR appears to entirely fill up the entirety `ModifierFlags`'s bitwise space, but I'm not sure how we could do this different other than to have another flag prop or something.

There are still quite a few bits remaining, and we don't add new modifiers all that often. I'm trying to conserve memory as much as possible here, so I'd rather wait to expand into a second property when it becomes absolutely necessary.

PR #56201, opened 2023-10-24
Fixed types of properties of contextual filtering mapped types

Andarist:
This is based on the [contextualPropertyOfGenericMappedType](https://github.com/microsoft/TypeScript/blob/7f482f9e47110ba40eaafe6b097c6780192a8e0b/tests/cases/compiler/contextualPropertyOfGenericMappedType.ts) test case - just with the added filtering `.nameType`

fixes https://github.com/microsoft/TypeScript/issues/56881


RyanCavanaugh:
@typescript-bot test this
@typescript-bot test top100
@typescript-bot run dt
@typescript-bot pack this

tests/cases/compiler/contextualPropertyOfGenericFilteringMappedType.ts
```diff
@@ -0,0 +1,33 @@
...
+  },
+  {
+    foo: (value, key) => {},
+    bar: (value, key) => {},
+  },
+);
+
+declare function f2<T extends object>(
+  data: T,
+  handlers: { [P in keyof T as T[P] extends string ? P : never]: (value: T[P], prop: P) => void },
```
julienandreu, line 22: *Question:* Is there any reason you're using the `extends string` here instead of `PropertyKey` ?

Andarist: it's meant to actually filter something, so I intentionally filter `T[P]` that match some subset of potential types. Note that I'm checking against `T[P]` and not just `P` - but even if the check would be against `P`, I could still want to filter the keys so `PropertyKey` isn't what I'd use here

andrewbranch (approved):

jakebailey:
This also makes me wonder if this is also a bug:

```diff
- const keysRemapped = !!target.declaration.nameType;
+ const keysRemapped = getMappedTypeNameTypeKind(target) === MappedTypeNameTypeKind.Remapping;
```

(Or maybe other places where `nameType` is used)

PR #56211, opened 2023-10-24
Hash-pin GitHub Actions, use dependabot to update them

### LINKED ISSUE #55216: Hash-pin workflow dependencies, keep them updated with Dependabot
# Suggestion

## 🔍 Search Terms
github action, hash-pin, dependabot

## ✅ Viability Checklist

## ⭐ Suggestion

When developing with CI workflows, it's common to version-pin dependencies (i.e. `actions/checkout@v3`). However, version tags are mutable, so a malicious attacker could overwrite a version tag to point to a malicious or vulnerable commit instead.

Pinning workflow dependencies by hash ensures the dependency is immutable and its behavior is guaranteed.

These dependencies can be kept up-to-date with dependabot. Whenever a new version of an Action is released, you'll receive a PR updating both its hash and the version comment (see [this repo](https://github.com/pnacht/logr/pulls) as an example).

I'll send a PR pinning the dependencies and adding dependabot along with this issue.

## 📃 Motivating Example

The `new-release-branch.yml` workflow runs with `contents: write` permissions, and the workflow commits directly to the repository. That workflow (like any others) relies on the `actions/setup-node@v3` Action.

If an attacker manages to take over the `actions/setup-node` Action, they could release a malicious `@v3` version of the Action which would run whenever `new-release-branch` runs. They could use it to add their own malicious code to the next release based on that branch.

However, if TypeScript instead uses `actions/setup-node@[hash]`, the Action's behavior is guaranteed: we will always run the same exact version each time.

## 💻 Use Cases

\#N/A

pnacht:
Fixes #55216.

This PR hash-pins almost all GitHub Actions. The only exception is `microsoft/TypeScript-Twoslash-Repro-Action`, which has no tags or release branches, and must therefore remain `@master`.

The PR also adds Dependabot to keep the Action hashes (and associated version comments) up-to-date. It is configured to send a single monthly PR updating all Actions at once. If you'd rather change the update cadence (weekly, for example) or create smaller logical groups of Actions, let me know and I'll update the config.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,15 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    groups:
+      github-actions:
+        patterns:
+          - "*"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,11 +5,11 @@
 
 version: 2
 updates:
-  - package-ecosystem: "github-actions"
-    directory: "/"
+  - package-ecosystem: 'github-actions'
+    directory: '/'
     schedule:
-      interval: "monthly"
+      interval: 'monthly'
     groups:
       github-actions:
         patterns:
-          - "*"
+          - '*'
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,7 +8,7 @@ updates:
   - package-ecosystem: 'github-actions'
     directory: '/'
     schedule:
-      interval: 'monthly'
+      interval: 'weekly'
     groups:
       github-actions:
         patterns:
--- a/.github/workflows/accept-baselines-fix-lints.yaml
+++ b/.github/workflows/accept-baselines-fix-lints.yaml
@@ -17,10 +17,10 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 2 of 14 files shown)


jakebailey:
I think I'd prefer weekly; monthly seems a little too spread out.

Overall I think the worst thing about pinning is how it's a pain to update locally if you need some feature "now"; I'm not aware of any tool one can run that updates things to this format (which dependabot and renovate both use). I pin actions on my repos and use renovate, but I'm also happy to have loads of PR noise that I don't wish to bestow on the team 😄

jakebailey:
As for `TypeScript-Twoslash-Repro-Action`, can it not pin at `<hash> # master`?

pnacht:
> I think I'd prefer weekly; monthly seems a little too spread out.

Done!

> As for `TypeScript-Twoslash-Repro-Action`, can it not pin at `<hash> # master`? Then again, that's code we own, and it'd be a pain to not get updates there in short order when we change it.

Yes, absolutely. Done!

I was just concerned it'd create too much noise. But given that we're using grouped PRs for all Actions, it actually isn't a problem, especially since the Action's repo isn't very active.

jakebailey:
Let's stick with pinned for now, and if it's a problem it's easy to go back. I agree that the other repo isn't extremely trafficked at the moment though there's some recent stuff I wanted to change there.

jakebailey (approved):

diogoteles08:
> Overall I think the worst thing about pinning is how it's a pain to update locally if you need some feature "now"; I'm not aware of any tool one can run that updates things to this format (which dependabot and renovate both use).

I'm not 100% sure I understood your complain, but renovate bot have a feature that "triggers" it to run again and check for new updates "now". Is that what you're looking for? See an example of this trigger at [this issue](https://github.com/diogoteles08/testing-dependabot/issues/13) of one testing repo of mine. The trigger is at the bottom.

PR #56212, opened 2023-10-24
Fix(55658): Move to file and new file syntax errors with overload functions

### LINKED ISSUE #55658: Move to new file produces syntax errors with overloaded functions
### 🔎 Search Terms

"move to new file", "new file syntax error", "to new file error", "move to file overloads"

- I first noticed this on a nightly build of 5.1.6. 
- It is still occurring on the latest nightly build, `5.3.0-dev.20230905`
- I can't say I ever encountered this bug when using this feature pre-5.0
- I can't provide a reproduction in the playground because this depends on the mounted file system code actions, which are not available in Monaco

### ⏯ Playground Link

_No response_

### 💻 Code

```ts
export function CanBeHeldWeakly<V>(v: V): v is WeakKey<V>;
export function CanBeHeldWeakly<V extends symbol>(v: V): v is V;
export function CanBeHeldWeakly<V extends object>(v: V): v is V;
export function CanBeHeldWeakly(v: unknown): v is object | symbol;
export function CanBeHeldWeakly(v: unknown): v is object | symbol {
  foo();
  return (v != null && (typeof v === "object" || typeof v === "function")) || (
    typeof v === "symbol" && Symbol.keyFor(v) === undefined
  );
}

export function foo(): "bar";
export function foo(bar: "bar"): "bar";
export function foo(bar?: "bar"): "bar" {
  return bar() === "foo" ? "bar" : dependent();
}

export function bar(): "foo" {
  return "foo";
  CanBeHeldWeakly(1);
}

export function dependent() {
  if (CanBeHeldWeakly(globalThis)) {
    return foo();
  }
}
```

### 🙁 Actual behavior

If I select the CanBeHeldWeakly function and apply the code action to move it to a new file, the automatic import generated for the `foo` function on which it depends produces a syntax error. It should only import one identifier - `foo` - from the file `./test1.ts`. As you can see in the screen recording below, it instead imports `foo` three times (one for each of its overload signatures).

As an experiment, I then moved foo to a new file, and the issue seemed to fix itself in that instance. 

It did not occur when I tried only moving `foo` to another file and not `CanBeHeldWeakly`.

https://github.com/microsoft/TypeScript/assets/11234104/90914ac3-1d99-4afa-8159-c51f6d8e4242

### 🙂 Expected behavior

I'd expect it to generate a single import for `foo` and all of its overloads.

_No response_

navya9singh:
Fixes https://github.com/microsoft/TypeScript/issues/55658
Importing overload functions caused this issue. Fixed it by ensuring that a symbol's declaration for an overload function was being imported only once.

### DIFF
--- a/src/services/refactors/moveToFile.ts
+++ b/src/services/refactors/moveToFile.ts
@@ -348,7 +348,7 @@ function getTargetFileImportsAndAddExportInOldFile(
                 if (hasSyntacticModifier(decl, ModifierFlags.Default)) {
                     oldFileDefault = name;
                 }
-                else {
+                else if (!oldFileNamedImports.includes (name.text)) {
                     oldFileNamedImports.push(name.text);
                 }
             }
--- a/src/services/refactors/moveToFile.ts
+++ b/src/services/refactors/moveToFile.ts
@@ -348,7 +348,7 @@ function getTargetFileImportsAndAddExportInOldFile(
                 if (hasSyntacticModifier(decl, ModifierFlags.Default)) {
                     oldFileDefault = name;
                 }
-                else if (!oldFileNamedImports.includes (name.text)) {
+                else if (!oldFileNamedImports.includes(name.text)) {
                     oldFileNamedImports.push(name.text);
                 }
             }
--- a/src/services/refactors/moveToNewFile.ts
+++ b/src/services/refactors/moveToNewFile.ts
@@ -176,7 +176,7 @@ function getNewFileImportsAndAddExportInOldFile(
             if (hasSyntacticModifier(decl, ModifierFlags.Default)) {
                 oldFileDefault = name;
             }
-            else if (!oldFileNamedImports.includes (name.text)) {
+            else if (!oldFileNamedImports.includes(name.text)) {
                 oldFileNamedImports.push(name.text);
             }
         }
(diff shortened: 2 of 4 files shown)


src/services/refactors/moveToFile.ts
```diff
@@ -348,7 +348,7 @@ function getTargetFileImportsAndAddExportInOldFile(
                 if (hasSyntacticModifier(decl, ModifierFlags.Default)) {
                     oldFileDefault = name;
                 }
-                else {
+                else if (!oldFileNamedImports.includes (name.text)) {
```
RyanCavanaugh, line 351:
```suggestion
                else if (!oldFileNamedImports.includes(name.text)) {

```

src/services/refactors/moveToNewFile.ts
```diff
@@ -176,7 +176,7 @@ function getNewFileImportsAndAddExportInOldFile(
             if (hasSyntacticModifier(decl, ModifierFlags.Default)) {
                 oldFileDefault = name;
             }
-            else {
+            else if (!oldFileNamedImports.includes (name.text)) {
```
RyanCavanaugh, line 179: Let's deduplicate this code

RyanCavanaugh (changes requested):

PR #56220, opened 2023-10-25
Better completion for property access of computed properties

### LINKED ISSUE #56096: Intellisense provides different suggestions on autocomplete when using indexed access or dotted access
### 🔎 Search Terms

"autocomplete suggestions intellisense differences between indexed access and dotted access"
"indexed access vs dotted access suggested properties"

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about

I tested it in TS Playground with Nightly and TS 5.2.

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.3.0-dev.20231013#code/PQKhCgAIUhxB5AggGQFyQOYEsBuBTAO0gEMjCBXAW0gHsAzSABxoGcWsAjAGz0j3wIAXFgBpIAd17jSgyIJpQYxNlgxEWNSr36FZjYgE8uNYgBM5Bxr3mQNW2g0EALbQOEA6RZC8AVF5AAnPBZZFicaci5zDDxZYkxCPACsAGNIACIAUTd0pkNjMxIVNTxTT2hIAHU8MWdCMTCIqMgOXmJuaxpILQCYlppnOX8QgPIUwXIgli9Sc1aSDsgsIjpyAnGsGgIWIeJZchZeGIIk1J2uLABraycsHa3eVuNxLxsaDkFiZaHefSMTcwhYgBQTLDCQOgBTQ-PhuCxWcogYDgcCmPApLjA3irdagrYQrBcQRJADCB3klGyugACvkAQAxGgBKlCAA8Pj4AA9iQRTDsSdSWcIAHwACjoTKF6B8AEp0PAOIcAjh2jxWQKhSwALLERgAbR8AF1hSjQBAKlqsJRUpcwYFOQArHY2Yg4GhYcxoqy8wgpAxLSjMEGKZHgQSWXgKpUqjqsgAawsgAF5IABvcAASBY5EVKWSrVFcsgUaSMbVCfAAF8ANymsCQZCxADkO1a2CISJRFGoGrcO3TGZJyAAkiSANLJjIYm3pESZ6nIRAATUn6UYmIMs8zJMQADkF8vVylSOvDFuMwel8Pd7BV6eDGDz9TEABVADKmTvxAOeHPmQASv+8D-quSRQgEf67gAIpkUGgT6pjnsgSBQW+PiIP+HIpukBSmG+nwgue-6ID4mQkgAEnusCfthAR7OiTikDE56VIgw4+Net7YdIWCggQGDnh+mRjpxq6HHgtr8YJmTCbBYl4BJpTnhxWqZC+1JQSRNEZKCWjkIwpj0eeABq8DIC+qkUVR2npG6XBUAxTG-nOGZQS+xEcfAu5WTeNmmJMeybOsjH8c5maIFBUE+MR46rmYpiCHRKSXH+WpvlxGR4JQLACS5l4AEKIOOAD6+WZLA17FZRyCkXB2H3hwxDJfleDtuRxBEkpLloRhPjFSpmTFYgxlsYu+XIDZQIgj4Vp4IgKqEqqYWDl5pG7n1PWkcVPnUXVU5bDygj4fRJIhTEiEua+UHDvAxWXsV-7DrA5F9UVACKL7Dv+cnYd+pibNSG7-qoTiCIgKQAI7kFgQQXZmm1vsV6maaRYmfMIL4GUZc6VqawCQLS-xmDsjH4C0ClEEEp4pKUEJMjCVNQgAtMehxiBw5CyHU-qmF0BADN0ezEgEgQg7I-PiJ4yzC3QTW8AKOqMFYpiammmZ6r2ugsO4mRpbAhroGslwSwQ1beBmGuCn27huR5N3eZRvkG5ARsm7WFua0I2umeZlmO9RzuuzQ4im+rnseANyNaYHBDG8Hoce1bWs64BwEx3HIfu5bmruB5ZH+5k6du2HSde+4EVRTFY5F-HWfh9rJKrZk63FZtg07bBNeZyXOc9Zh-XDqpQ0jcOY0TV3CfZ9bBVFWOpXlZV1W1RPdelx4CNIxp0eG7HxeJznV03Xdi5Lg9T0vUNJIfV9nc7xnk-1+4QkiTeK899bz+3y7u+11WdbmjAPwdwliiAcJAFIEQAhsxhCwYg9h6BeC8AAAxQV4f86ImSmFZJkTkGJyBonVGvUB1wDD0EgArXUytNTCjEEKE0FQUFIJDCicMVhyHUhfAQSglDSiqxTAOPUE5vg4LwQQ+uYgSFkIoUrXhfZhTOyFLWGsXZORBi5hGdhmpFaTmkVQvskAABk7DOHcJkSrPstZwB4FUUydRbCBQlmVEtPhasLZCKIJIhg9dFbO0cWWPAhCtG6kEcaJRlizReEbIIFs4Cgj0VoJMWgHAHTolkOIXiTgEki1MYwMEXhWiCEkIQPgBAqCQBITsWYSToxLW1sw1E6JMRBHAVsEIsJdB+NqegBxipSzOIsf-Lwu5g60G9DCaWeAuAXEONsXg+NvzyAgYGHgeIiBaFKbQIgzhgHWMYFMdg+IOA83dPxfozhUCIIYagioOghCdI6HUq5TCKgzF5JcmAjCvC3MEPcngLA9TpHSIaZBKCAAF7zvAVD8NoVRqTaZTEiLIYBzg9hLC5i4f0jRIjmD2TQHAHprDiC6P9OgdAki6C8NmXpwgwEpHCKkYIkBRTEhCLTdJgxdxiy4P6fE8BxiQAAIwAGYZT1O+b84IgyKj5RfD4MQWAGDDkgLzEgKQaZsCGFCcgGBMm80EMScwTV1XOi6PEXFVgQT+hRUinY-NQjapiKy0wc4Ki8RAYEDBagsAAC9abKEgCqC4hreQSEmVweEpREShnFb0pxDz3ANSapcFqbUOoGvAMAfGkBs05tzXmyAAA9AA-FKwBtwdh3HldsYkZgxBBAgV631ZRIDQuaeJbKchTVKyxEIbl5NNl9oeMqhVZKgjrBuHsLw1odWyHmHUcpeAeWOH8LijolB0DmqSOGBdBhKnNPrTQRtfqdgktHboKNXY3ASu1voH8Gas35sfdm4tLCNFjkXSweAdAhTXsnJ4iNZCY01IebWTNuaX2luLAwCBkxoFGuCOwU5zgtU6pKWUip7qQiEnDYGj0F7wAQOrZATkk5RRAb6Q8qeycZ4lTKhVXcVUUDLxlKBh9haS0EdabIf0KYyNXtjf4-5j9nzvkLix+94Gi1AA

### 💻 Code

```ts
/**
 * Mimicking rxjs to avoid dependency import
 */

type Observable<X> = {
	subscribe(): Observable<X>
};

enum CPEvents {
	CLICK = "click",
	PLAY = "play",
	CANPLAY = "canplay",
	PLAYING = "playing",
	PAUSE = "pause",
	ERROR = "error",
	ENDED = "ended",
	LOADSTART = "loadStart",
	RATECHANGE = "ratechange",
	WAITING = "waiting",
	SEEKING = "seeking",
	SEEKED = "seeked",
	TIMEUPDATE = "timeupdate",
	VOLUMECHANGE = "volumechange",
	DURATIONCHANGE = "durationchange",
	ADDTRACK = "addtrack",
	EMSG = "emsg",
	PLAYBACK_BEGIN_HALTED = "playbackBeginHalted",
	START_TIME_AVAILABLE = "startTimeAvailable",
	CONTENT_STATE_CHANGED = "contentStateChanged",
	AUDIO_PLAY_RIGHT_ACQUIRED = "audioPlayRightAcquired",
	STATS_UPDATE = "statsUpdate",
}

// Payloads have been replaced for the repro-case, but they do not matter right now.
interface CPMappedEvents {
	[CPEvents.EMSG]: unknown; 
	[CPEvents.DURATIONCHANGE]: unknown;
	[CPEvents.VOLUMECHANGE]: unknown;
	[CPEvents.TIMEUPDATE]: unknown;
	[CPEvents.ERROR]: unknown;
	[CPEvents.RATECHANGE]: unknown;
	[CPEvents.ADDTRACK]: unknown;
	[CPEvents.CONTENT_STATE_CHANGED]: unknown;
	[CPEvents.START_TIME_AVAILABLE]: unknown;
	[CPEvents.PLAYBACK_BEGIN_HALTED]: unknown;
	[CPEvents.STATS_UPDATE]: unknown;
	[CPEvents.AUDIO_PLAY_RIGHT_ACQUIRED]: unknown;
	[CPEvents.SEEKING]: unknown;
	[CPEvents.SEEKED]: unknown;
}

type CPUnmappedEvents = {
	[K in Exclude<CPEvents, keyof CPMappedEvents>]: Event;
};

export type CPEventsMap = CPMappedEvents & CPUnmappedEvents;

export type CPObservableEvents = {
	[K in keyof CPEventsMap]: Observable<CPEventsMap[K]>;
};

declare const eventObservables: CPObservableEvents;
```

### 🙁 Actual behavior

Intellisense (autocomplete) suggests two different subset of valid values when accessing a mapped object through dotted access and indexed access, but accessing with a dotted access to an unsuggested (available in indexed access and there fore still valid) property seems to be still valid and well-recognized.

### 🙂 Expected behavior

The suggested values should be the same if there is not a specific indexed signature.

_No response_

PranavSenthilnathan:
Improves completion to now show the qualified name of the completion.

```
const o = { A: "a" } as const;
enum e { B = "b" };
type t = { [o.A]: 1, [e.B]: 2 };
declare const x: t;
x.
// Old: completes to x[o]| and x[e]|
// New: completes to x[o.A]| and x[e.B]|
// ('|' denotes cursor position)
```

If the user sets `includeCompletionsWithInsertText` to false, we will only use the literal completion for the property (`x.a` and `x.b` in the example above).

Fixes #56096


src/services/completions.ts
```diff
@@ -3663,71 +3664,131 @@ function getCompletionData(
...
+                node = typeChecker.symbolToEntityName(nameSymbol, /*meaning*/ undefined!, contextToken, NodeBuilderFlags.UseAliasDefinedOutsideCurrentScope);
+            }
+            else {
+                // Object literals assigned as const
+                const leftMostNodeAccessExpression = typeChecker.symbolToNode(leftMostNameSymbol, /*meaning*/ undefined!, contextToken, NodeBuilderFlags.UseAliasDefinedOutsideCurrentScope);
+                type OnlyPropertyAccess = Identifier | (PropertyAccessExpression & { expression: OnlyPropertyAccess; });
+                node = createPropertyAccess(computedPropertyNameExpression as OnlyPropertyAccess);
+                function createPropertyAccess(n: OnlyPropertyAccess): Expression {
+                    if (isIdentifier(n)) {
+                        return leftMostNodeAccessExpression! as Expression; //TODO ! and cast
```
PranavSenthilnathan, line 3783: Not sure what to do here. The factory wants `Expression` but the base case here is coming from `symbolToNode` which is going to be a `Node`

jakebailey: You could `Debug.assertNode` just to make sure.

gabritto:
@typescript-bot pack this

src/services/completions.ts
```diff
@@ -3663,71 +3664,130 @@ function getCompletionData(
...
+                typeChecker.tryGetMemberInModuleExportsAndProperties(firstAccessibleSymbol.name, moduleSymbol) !== firstAccessibleSymbol
+            ) {
+                // If preferences allow insert text, add completion for [<QualifiedSymbolName>]
+                const node = preferences.includeCompletionsWithInsertText
+                    ? createComputedPropertyAccess(nameSymbol, leftMostNameSymbol, computedPropertyNameExpression)
+                    : undefined;
+
+                if (!node) {
+                    // Switch to literal symbol if user doesn't want insert text
+                    symbols[index] = symbol;
```
gabritto, line 3714: Could we move `symbols.push(nameSymbol)` to the `else` below and call `addLiteralSymbol()` here instead? I think it would be clearer this way, and also make sure we're still setting the things `addLiteralSymbol` sets (i.e. the correct sort text).

src/services/completions.ts
```diff
@@ -3663,71 +3664,130 @@ function getCompletionData(
...
             symbols.push(symbol);
         }
 
+        /** Combines a property access expression (from the completion location to the left-most name in the computed property expression)
+         * and the computed property expression itself to return an expression to access the computed property from the completion location.
+         * Assumes that the left-most symbol is accessible so there should always be a way to access the computed property symbolically. */
+        function createComputedPropertyAccess(nameSymbol: Symbol, leftMostNameSymbol: Symbol, computedPropertyNameExpression: Expression) {
+            let node: Node | undefined;
+            if (!isTransientSymbol(nameSymbol)) {
+                node = typeChecker.symbolToEntityName(nameSymbol, /*meaning*/ undefined!, contextToken, NodeBuilderFlags.UseAliasDefinedOutsideCurrentScope);
```
gabritto, line 3772: Why is `meaning` `undefined` instead of e.g. `SymbolFlags.All`?

src/services/completions.ts
```diff
@@ -3663,71 +3664,130 @@ function getCompletionData(
...
+
+        function addLiteralSymbol() {
             addSymbolOriginInfo(symbol);
             addSymbolSortInfo(symbol);
             symbols.push(symbol);
         }
 
+        /** Combines a property access expression (from the completion location to the left-most name in the computed property expression)
+         * and the computed property expression itself to return an expression to access the computed property from the completion location.
+         * Assumes that the left-most symbol is accessible so there should always be a way to access the computed property symbolically. */
```
gabritto, line 3768: That comment was a bit confusing to me. I think having a comment like the one you left above, with an example of what each parameter is and what we want to do, e.g. `x.y`, `y`'s symbol, `m`'s symbol, would help clarify it.

src/services/completions.ts
```diff
@@ -3663,71 +3664,130 @@ function getCompletionData(
...
+        function createComputedPropertyAccess(nameSymbol: Symbol, leftMostNameSymbol: Symbol, computedPropertyNameExpression: Expression) {
+            let node: Node | undefined;
+            if (!isTransientSymbol(nameSymbol)) {
+                node = typeChecker.symbolToEntityName(nameSymbol, /*meaning*/ undefined!, contextToken, NodeBuilderFlags.UseAliasDefinedOutsideCurrentScope);
+            }
+            else {
+                // Object literals assigned as const
```
gabritto, line 3775: What's this case for, exactly? I think I don't understand the comment here and how we know this from knowing that `nameSymbol` is a transient symbol.

src/services/completions.ts
```diff
@@ -3663,71 +3664,130 @@ function getCompletionData(
...
+                const leftMostNodeAccessExpression = typeChecker.symbolToNode(leftMostNameSymbol, /*meaning*/ undefined!, contextToken, NodeBuilderFlags.UseAliasDefinedOutsideCurrentScope);
```
gabritto, line 3776: Same as above on `meaning`

gabritto:
Improvement suggestion:
My understanding is that, with this PR, if we can't find an already-imported way to offer the computed property, we won't offer it in this new way and will default to the old way instead. Now, the old way already supports auto-importing symbols when possible. Could we make the new way also auto-import symbols?

Example:
```ts
// @filename: ex.ts
export type T = {
    [Sym.sym]: number;
}
export namespace Sym {
    export const sym = Symbol("unique!");
}
// @filename: index.ts
import { T } from "./ex";

declare const obj: T;
obj.|
```
Since we don't have `Sym` imported from `"ex.ts"`, we will default to offering a completion for `Sym` and auto-importing it.
We could, I imagine, suggest the new completion for `Sym.sym` and auto-import `Sym`.

src/services/completions.ts
```diff
@@ -3663,71 +3664,130 @@ function getCompletionData(
...
+                    symbolToOriginInfoMap[index] = origin;
                 }
-                addSymbolOriginInfo(symbol);
-                addSymbolSortInfo(symbol);
-                symbols.push(symbol);
             }
         }
-        else {
+
+        function addLiteralSymbol() {
```
gabritto, line 3760: Not very important, but `addLiteralSymbol` makes me think we're adding a symbol for a literal, maybe simply `addSymbol` would be better.

src/services/completions.ts
PranavSenthilnathan: This scenario goes into this branch:
```ts
const x = { a: "foo" } as const;
const y = { [x.a]: 0 };
y.|
```

It's because the symbol `a` is associated with the object and not `x` so walking up the parent symbols will just go to `__object` - so `symbolToEntityName` (the other branch) won't give a full access expression to it.

That being said, I don't know if `isTransientSymbol` is the right check here - are there transient symbols that won't run into this issue? and are there non-transient symbols that will run into this issue? I can be safe an always choose the `else` branch (get rid of the whole `if`).

src/services/completions.ts
```diff
@@ -3124,6 +3125,15 @@ function getContextualType(previousToken: Node, position: number, sourceFile: So
     }
 }
 
+/**
+ * An accessible symbol is a local in the same block or any enclosing blocks, up to and including the global scope (note imports are global).
```
PranavSenthilnathan, line 3129: I don't know if this comment is 100% accurate, but this is my understanding of what this method does. Open to change this if it's wrong.

src/services/completions.ts
gabritto: I'm not sure if all transient symbols run into this problem with `symbolToEntityName`, but I think I was confused by the comment because it sorta implied that if the symbol is transient, then we have that object literal situation, which I think is not always the case.
@weswigham might know the details of `symbolToEntityName` and `symbolToExpression`.

PR #51196, opened 2022-10-17
Filter possible contextual return types from unions for async functions and generators

Andarist:
fixes #51187
fixes #47682
fixes #41428
supersedes #47683
fixes #42439

PR #56222, opened 2023-10-25
Consider all constituents of contextual return type when resolving type of 'yield'

### LINKED ISSUE #42439: Union of Generator and AsyncGenerator introduces spurious undefined next() value path
# Bug Report

I was creating a coroutine-based interactive prompt library. It can consume either synchronous or asynchronous co-routines. 

There doesn't seem to be any runtime issue associated with this approach, (in some configurations the library harmlessly awaits on a synchronous generator.next()). 

However, any function which is declared as fulfilling a union of async+sync generators faces a compile error when a spurious `undefined` value is indicated as possible from a `yield` even though the next() call only accepts strings.

### 🔎 Search Terms

yield, undefined, typescript, asyncgenerator, generator

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about AsyncGenerator

### ⏯ Playground Link

[Playground link with relevant code](https://www.typescriptlang.org/play?target=5#code/C4TwDgpgBAyiB2BjGECOBXCSIDECGiwA9gE4hQC8UAFAJSUB8UA4lhCXsSQDwDOwJAJbwA5gBoo-IaIlThIhgG4AUMtCQoAQV4JkaTNnyFS5KnUZadSVvHadSfAfNlOZk1wpVrw0FBiyIuARcprC6fgaBRiFQAD6W4foBQcZkqohE8PySutEmAFxhSBHJeWSUUABm6EjAgpkAVDS0AN7KUFAANhDAUPB4ALbQVABEIyodAO4AFoLd1ACE-UOt7R19g8NQIIIQnQAmUCMA6tOcUIK820ToJBtDAPwjawC+ayQ9t-BQAAYAKmdgAByK7AabQLCHIiVKBg6AiTYSAAkLWWEBeP2ULyg6Uy2TwVkQZRAhW0iX8hmCJgqBN0VRqhHq8CadDaHW6vTRFTGEygMzmEEWaNW63uWx2e0OJ0BFyuIBudzRT1e70+JG+-0BINh4KgkKg0J18MRUBRaIxWJxygyWV6nSIRF47DgSGJhRKlNSoWqtSZLJFXR6Yu54zW-PmS02AY6XKoEoOR1O50u11uYuVHTeHQ+wC+vwBnG1cL18ChMOLCKGyNRmwt2Nxtq6DqdJDJrqpZHdSU9MSotKQ9N9jWabMDnM2Id54cFkZWo5jE7juwT0uTcoV6eemdVufV+a1oN1+sNFZNZtrmPrQA)

### 💻 Code

```ts
type SyncSequenceFactory = () => Generator<string, string, string>;

type AsyncSequenceFactory = () => AsyncGenerator<string, string, string>;

type SequenceFactory = SyncSequenceFactory | AsyncSequenceFactory

const syncFactory: SyncSequenceFactory = function* (){
  let name = "";
  while(!name){
    name = yield "What is your name?"
  }
  return `That's the end of the game, ${name}`
} 

const asyncFactory: AsyncSequenceFactory = async function* (){
  let name = "";
  while(!name){
    name = yield "What is your name?"
  }
  return `That's the end of the game, ${name}`
} 

const looserSyncFactory: SequenceFactory = function* (){
  let name = "";
  while(!name){
    name = yield "What is your name?"
  }
  return `That's the end of the game, ${name}`
} 

const looserAsyncFactory: SequenceFactory = async function* (){
  let name = "";
  while(!name){
    name = yield "What is your name?"
  }
  return `That's the end of the game, ${name}`
} 

```

### 🙁 Actual behavior

Compiler reports `Type 'string | undefined' is not assignable to type 'string'.`

### 🙂 Expected behavior

I don't believe this possible execution path is actually made possible by the union, since it is not possible in either of the unioned cases.

rbuckton:
This avoids an early-exit in `getIterationTypesOfIterable` when the intended use is resolving the iteration types of a contextual return type of a generator function:

```ts
let f: () => Generator<string, void, string> | AsyncGenerator<string | void | string>;
f = function*() {
  let name: string;
  name = yield ""; 
};
```

The above code is an error currently as we drop the iteration types of the first constituent when the second fails.

For other uses we continue to preserve the early exit so that we correctly error for cases such as this:

```ts
let data: string[] | null;
const [value] = data; // should still be an error when the second constituent fails.
```

Fixes #42439

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -8816,12 +8816,12 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
 
             function addExportModifier(node: Extract<HasModifiers, Statement>) {
                 const flags = (getEffectiveModifierFlags(node) | ModifierFlags.Export) & ~ModifierFlags.Ambient;
-                return factory.updateModifiers(node, flags);
+                return factory.replaceModifiers(node, flags);
             }
 
             function removeExportModifier(node: Extract<HasModifiers, Statement>) {
                 const flags = getEffectiveModifierFlags(node) & ~ModifierFlags.Export;
-                return factory.updateModifiers(node, flags);
+                return factory.replaceModifiers(node, flags);
             }
 
             function visitSymbolTable(symbolTable: SymbolTable, suppressNewPrivateContext?: boolean, propertyAsAlias?: boolean) {
@@ -9120,7 +9120,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                         newModifierFlags |= ModifierFlags.Default;
                     }
                     if (newModifierFlags) {
-                        node = factory.updateModifiers(node, newModifierFlags | getEffectiveModifierFlags(node));
+                        node = factory.replaceModifiers(node, newModifierFlags | getEffectiveModifierFlags(node));
                     }
                 }
                 results.push(node);
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -1102,6 +1102,7 @@ const enum IterationUse {
     SpreadFlag = 1 << 5,
     DestructuringFlag = 1 << 6,
     PossiblyOutOfBounds = 1 << 7,
+    ReturnType = 1 << 8,
 
     // Spread, Destructuring, Array element assignment
     Element = AllowsSyncIterablesFlag,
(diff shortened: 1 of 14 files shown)


Andarist:
[My old PR](https://github.com/microsoft/TypeScript/pull/51196) that awaits review already fixes this issue (I just [pushed out a test case](https://github.com/microsoft/TypeScript/pull/51196/commits/03c6ee4d262a5733e406251dc64599ef35ad46f6) that is being added here in this PR). It also shares all of the baseline changes that you have here.

Would you mind re-reviewing that PR and comparing the used approaches? Maybe it would be also good to pull all the other test cases that I added there here to compare their results?

rbuckton:
> [My old PR](https://github.com/microsoft/TypeScript/pull/51196) that awaits review already fixes this issue (I just [pushed out a test case](https://github.com/microsoft/TypeScript/pull/51196/commits/03c6ee4d262a5733e406251dc64599ef35ad46f6) that is being added here in this PR). It also shares all of the baseline changes that you have here.
> 
> Would you mind re-reviewing that PR and comparing the used approaches? Maybe it would be also good to pull all the other test cases that I added there here to compare their results?

The test in tests/cases/compiler/contextuallyTypeAsyncFunctionReturnTypeFromUnion.ts fails in my branch. It looks like your PR covers more cases, so I'll happily close this branch in favor of your PR.

rbuckton:
Closed in favor of #51196.

PR #56237, opened 2023-10-27
Revert "Explicitly declare type=commonjs in package.json (#56190)"

jakebailey:
This apparently had an effect on esbuild's behavior. I didn't notice it in the package size report:

| - | - | - | - | - |
| Packed | 5.49 MiB | 5.56 MiB | +71.00 KiB | +1.26% |
| Unpacked | 30.49 MiB | 31.03 MiB | +546.64 KiB | +1.75% |

| - | - | - | - | - |
| `lib/tsc.js` | 5.55 MiB | 5.55 MiB | +75.00 B | +0.00% |
| `lib/tsserver.js` | 8.12 MiB | 8.52 MiB | +407.05 KiB | +4.89% |
| `lib/typescript.js` | 8.59 MiB | 8.72 MiB | +139.49 KiB | +1.59% |
| `package.json` | 3.44 KiB | 3.46 KiB | +24.00 B | +0.68% |

I haven't investigated exactly why this happens, other than that some export objects are getting duplicated. But, I did not intend this at all, just to maybe make our package more explicit / faster before the release.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,6 @@
         "type": "git",
         "url": "https://github.com/Microsoft/TypeScript.git"
     },
-    "type": "commonjs",
     "main": "./lib/typescript.js",
     "typings": "./lib/typescript.d.ts",
     "bin": {


jakebailey:

jakebailey:
@evanw I hate to ping another time, but maybe you have a better idea what might be happening here. I know our reexports are awful, but the behavior surprised me given `type=commonjs` is implicitly true because we omit it.

jakebailey:
@DanielRosenwasser Not sure if you're actually making the release branch today, but I'd definitely want to get this revert in before then as this was totally meant to be a benign positive change.

gabritto (approved):

evanw:
@jakebailey I believe I know why this happens. Here is a demonstration:

* [With `type: commonjs`](https://esbuild.github.io/try/#YgAwLjE5LjUALS1idW5kbGUgLS1wbGF0Zm9ybT1ub2RlAABwYWNrYWdlLmpzb24AewogICJ0eXBlIjogImNvbW1vbmpzIgp9AGUAc3JjL3Rzc2VydmVyL3NlcnZlci50cwBleHBvcnQgKiBmcm9tICIuL19uYW1lc3BhY2VzL3RzIjsAAHNyYy90c3NlcnZlci9fbmFtZXNwYWNlcy90cy50cwBleHBvcnQgKiBmcm9tICIuLi8uLi9zZXJ2ZXIvX25hbWVzcGFjZXMvdHMiOwAAc3JjL3NlcnZlci9fbmFtZXNwYWNlcy90cy50cwBleHBvcnQgKiBmcm9tICIuLi8uLi9kZXByZWNhdGVkQ29tcGF0L19uYW1lc3BhY2VzL3RzIjsAAHNyYy9kZXByZWNhdGVkQ29tcGF0L19uYW1lc3BhY2VzL3RzLnRzAGV4cG9ydCAqIGZyb20gIi4uLzUuMC9pZGVudGlmaWVyUHJvcGVydGllcyI7AABzcmMvZGVwcmVjYXRlZENvbXBhdC81LjAvaWRlbnRpZmllclByb3BlcnRpZXMudHMAZGVjbGFyZSBtb2R1bGUgIi4uLy4uL2NvbXBpbGVyL3R5cGVzIiB7fQ)
* [Without `type: commonjs`](https://esbuild.github.io/try/#YgAwLjE5LjUALS1idW5kbGUgLS1wbGF0Zm9ybT1ub2RlAABwYWNrYWdlLmpzb24Aewp9AGUAc3JjL3Rzc2VydmVyL3NlcnZlci50cwBleHBvcnQgKiBmcm9tICIuL19uYW1lc3BhY2VzL3RzIjsAAHNyYy90c3NlcnZlci9fbmFtZXNwYWNlcy90cy50cwBleHBvcnQgKiBmcm9tICIuLi8uLi9zZXJ2ZXIvX25hbWVzcGFjZXMvdHMiOwAAc3JjL3NlcnZlci9fbmFtZXNwYWNlcy90cy50cwBleHBvcnQgKiBmcm9tICIuLi8uLi9kZXByZWNhdGVkQ29tcGF0L19uYW1lc3BhY2VzL3RzIjsAAHNyYy9kZXByZWNhdGVkQ29tcGF0L19uYW1lc3BhY2VzL3RzLnRzAGV4cG9ydCAqIGZyb20gIi4uLzUuMC9pZGVudGlmaWVyUHJvcGVydGllcyI7AABzcmMvZGVwcmVjYXRlZENvbXBhdC81LjAvaWRlbnRpZmllclByb3BlcnRpZXMudHMAZGVjbGFyZSBtb2R1bGUgIi4uLy4uL2NvbXBpbGVyL3R5cGVzIiB7fQ)

When esbuild sees a file without any exports, it attempts to guess what type of module it is. That logic lives [here](https://github.com/evanw/esbuild/blob/4e11b50fe3178ed0a78c077df78788d66304d379/internal/js_parser/js_parser.go#L17641-L17660):

```go
// If this module has no exports, try to determine what kind of module it
// is by looking at node's "type" field in "package.json" and/or whether
// the file extension is ".mjs"/".mts" or ".cjs"/".cts".
switch {
case p.options.moduleTypeData.Type.IsCommonJS():
  // ".cjs" or ".cts" or ("type: commonjs" and (".js" or ".jsx" or ".ts" or ".tsx"))
  exportsKind = js_ast.ExportsCommonJS

case p.options.moduleTypeData.Type.IsESM():
  // ".mjs" or ".mts" or ("type: module" and (".js" or ".jsx" or ".ts" or ".tsx"))
  exportsKind = js_ast.ExportsESM

default:
  // Treat unknown modules containing an import statement as ESM. Otherwise
  // the bundler will treat this file as CommonJS if it's imported and ESM
  // if it's not imported.
  if p.esmImportStatementKeyword.Len > 0 {
    exportsKind = js_ast.ExportsESM
  }
}
```

The file `src/deprecatedCompat/5.0/identifierProperties.ts` has no exports (there may be others too, but I haven't checked). In that case the file is considered to be a CommonJS module if `type: commonjs` is present in `package.json`, which means that imports of that file use `require()`, which means that re-exports of that file use dynamic re-exports via a property-copying loop instead of static re-exports via inheriting the re-exported properties at compile time, which results in a bigger bundle size.

If `type` is absent then esbuild guesses that the file should be considered ESM if there is an `import` statement, which there is in `src/deprecatedCompat/5.0/identifierProperties.ts`. It's true that node would default to CommonJS in this case, but esbuild can't really do that because esbuild is a bundler and bundling code that people would consider to be ESM as CommonJS instead by default can result in much worse code, which is a non-starter. So esbuild only does that if you explicitly say that files without `.m*` or `.c*` extensions should be CommonJS.

jakebailey:
I see, so to recap for my own understanding, for files without exports, setting `type=commonjs` forced us into the first case, rather than falling back to the default. I'm honestly surprised that just having an import statement wouldn't be enough to force it into ESM (for "maximum good codegen"), but I know this stuff is fickle. In our case, we kinda do want it to treat everything as ESM; with the way our build works, we actually `--define` away `require` entirely and replace it post-build, making the entire bundle entirely ESM internally (source wise), only calling `require` on Node. Probably we're unique in that we don't have external deps whatsoever.

It seems like adding `export {}` to the end of these files fixes the problem (at least in the playground). So maybe that is the fix or leads to one.

For now I'm going to just revert the change and swing back later to figure out something that will allow us to set this and also not regress the code size. I really assumed this change would be benign, so I'd rather go back to the status quo given today's the release cutoff. (Good thing I looked at packagephobia while building my perf review...)

Thanks for the help, I appreciate it!

Jack-Works:
can you rename it to `.mts` to force esbuild treat it as esm?

jakebailey:
No, that would totally break our `--no-bundle` build. Using non-ts extensions is its own can of worms.

Issue #56229, opened 2023-10-27
Invalid `.d.ts` generated from JSDoc if `module.exports` is assigned not an identifier


remcohaszing:
### 🔎 Search Terms

jsdoc namespace commonjs

### 🕗 Version & Regression Information

- This changed between versions 3.6 and 3.7 (the first version to support types in JSDoc)

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.2.2&filetype=js#code/PQKhCgAIUgBAXAngBwKYBNUDNIHlnwCWA9gHYDOUMsyATsWrUpAN7ny2GkDmAvpA3hVg4cKAjQ4yAIa1pAW1b4iZcv0EkKw8POLoArgBtUAOlQAPZMSblIAXkgAKDaoCU9gHyteAbnBA

### 💻 Code

```js
/**
 * @typedef Options
 * @property {string} opt
 */

/**
 * @param {Options} options
 */
module.exports = function loader(options) {}
```

### 🙁 Actual behavior

TypeScript emits the following `.t.ds`.

```ts
declare function _exports(options: Options): void;
export = _exports;
export type Options = {
    opt: string;
};
```

### 🙂 Expected behavior

I expect the TypeScript to emit the following:

```ts
export = loader;
/**
 * @typedef Options
 * @property {string} opt
 */
/**
 * @param {Options} options
 */
declare function loader(options: Options): void;
declare namespace loader {
    export { Options };
}
type Options = {
    opt: string;
};
```

Or when the expression is unnamed:

```ts
export = _exports;
/**
 * @typedef Options
 * @property {string} opt
 */
/**
 * @param {Options} options
 */
declare function _exports(options: Options): void;
declare namespace _exports {
    export { Options };
}
type Options = {
    opt: string;
};
```

### Additional information about the issue

I initially noticed this in the output of https://github.com/mdx-js/mdx/blob/3.0.0/packages/loader/index.cjs.

Andarist:
I don't know much about namespaces in TS so I'm not exactly sure how this should behave. Why do you believe that this should emit a namespace? I understand that this is valid: `/** @typedef {import('./input').Options} Alias */` but isn't it because all `@typedef`s are auto-exported from a JS file?

remcohaszing:
The expected behaviour as I described, matches the behaviour of when a value is declared first, then exported using an identifier. That behaviour is correct, as namespaces are the only way to export a type for a CJS file that assigns a value to `module.exports`.

[...]

PR #56243, opened 2023-10-28
Fxied declaration emit for JS files when `module.exports` is assigned a non-alias value and when it has extra type members

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/56229


sandersn (changes requested):
This seems like a good solution. I'd like @weswigham to take a look too.

src/compiler/checker.ts
```diff
@@ -9197,8 +9197,12 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             }
 
             function getNamespaceMembersForSerialization(symbol: Symbol) {
-                const exports = getExportsOfSymbol(symbol);
-                return !exports ? [] : filter(arrayFrom(exports.values()), m => isNamespaceMember(m) && isIdentifierText(m.escapedName as string, ScriptTarget.ESNext));
+                let exports = arrayFrom(getExportsOfSymbol(symbol).values());
+                const merged = getMergedSymbol(symbol);
+                if (merged !== symbol) {
+                    exports = concatenate(exports, filter(arrayFrom(getExportsOfSymbol(merged).values()), m => !(getSymbolFlags(resolveSymbol(m)) & SymbolFlags.Value)));
```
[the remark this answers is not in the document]
Andarist: > what happens if you don't concatenate+filter?

Please see the diff below:

<details>
<summary>git diff</summary>

```diff
diff --git a/src/compiler/checker.ts b/src/compiler/checker.ts
index 6ab6160af6..d9084479f3 100644
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -9268,7 +9268,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 let exports = arrayFrom(getExportsOfSymbol(symbol).values());
                 const merged = getMergedSymbol(symbol);
                 if (merged !== symbol) {
-                    exports = concatenate(exports, filter(arrayFrom(getExportsOfSymbol(merged).values()), m => !(getSymbolFlags(resolveSymbol(m)) & SymbolFlags.Value)));
+                    exports = arrayFrom(getExportsOfSymbol(merged).values());
                 }
                 return filter(exports, m => isNamespaceMember(m) && isIdentifierText(m.escapedName as string, ScriptTarget.ESNext));
             }
diff --git a/tests/baselines/reference/jsDeclarationsExportAssignedClassInstance3.js b/tests/baselines/reference/jsDeclarationsExportAssignedClassInstance3.js
index 5f23727d23..fb9234b587 100644
--- a/tests/baselines/reference/jsDeclarationsExportAssignedClassInstance3.js
+++ b/tests/baselines/reference/jsDeclarationsExportAssignedClassInstance3.js
@@ -23,5 +23,37 @@ module.exports.additional = 20;
 
 
 //// [index.d.ts]
-export let member: number;
-export let additional: 20;
+declare namespace _exports {
+    export { additional };
+}
+declare namespace _exports {
+    let member: number;
+    let additional: 20;
+}
+export = _exports;
+
+
+//// [DtsFileErrors]
+
+
+out/index.d.ts(2,14): error TS2323: Cannot redeclare exported variable 'additional'.
+out/index.d.ts(2,14): error TS2484: Export declaration conflicts with exported declaration of 'additional'.
+out/index.d.ts(6,9): error TS2323: Cannot redeclare exported variable 'additional'.
+
+
+==== out/index.d.ts (3 errors) ====
+    declare namespace _exports {
+        export { additional };
+                 ~~~~~~~~~~
+!!! error TS2323: Cannot redeclare exported variable 'additional'.
+                 ~~~~~~~~~~
+!!! error TS2484: Export declaration conflicts with exported declaration of 'additional'.
+    }
+    declare namespace _exports {
+        let member: number;
+        let additional: 20;
+            ~~~~~~~~~~
+!!! error TS2323: Cannot redeclare exported variable 'additional'.
+    }
+    export = _exports;
+    
\ No newline at end of file
diff --git a/tests/baselines/reference/jsDeclarationsExportAssignmentExpressionPlusSecondary.js b/tests/baselines/reference/jsDeclarationsExportAssignmentExpressionPlusSecondary.js
index 83bf1fb653..32b3b52524 100644
--- a/tests/baselines/reference/jsDeclarationsExportAssignmentExpressionPlusSecondary.js
+++ b/tests/baselines/reference/jsDeclarationsExportAssignmentExpressionPlusSecondary.js
@@ -31,12 +31,50 @@ module.exports.Strings = Strings;
 
 
 //// [index.d.ts]
-export namespace Strings {
+declare namespace _exports {
+    export { Strings };
+}
+declare namespace _exports {
+    export let thing: string;
+    export let also: string;
+    export namespace desc {
+        let item: string;
+    }
+    export { Strings };
+}
+export = _exports;
+declare namespace Strings {
     let a: string;
     let b: string;
 }
-export declare let thing: string;
-export declare let also: string;
-export declare namespace desc {
-    let item: string;
-}
+
+
+//// [DtsFileErrors]
+
+
+out/index.d.ts(2,14): error TS2300: Duplicate identifier 'Strings'.
+out/index.d.ts(10,14): error TS2300: Duplicate identifier 'Strings'.
+
+
+==== out/index.d.ts (2 errors) ====
+    declare namespace _exports {
+        export { Strings };
+                 ~~~~~~~
+!!! error TS2300: Duplicate identifier 'Strings'.
+    }
+    declare namespace _exports {
+        export let thing: string;
+        export let also: string;
+        export namespace desc {
+            let item: string;
+        }
+        export { Strings };
+                 ~~~~~~~
+!!! error TS2300: Duplicate identifier 'Strings'.
+    }
+    export = _exports;
+    declare namespace Strings {
+        let a: string;
+        let b: string;
+    }
+    
\ No newline at end of file
diff --git a/tests/baselines/reference/jsDeclarationsExportDoubleAssignmentInClosure.js b/tests/baselines/reference/jsDeclarationsExportDoubleAssignmentInClosure.js
index 610ded5f10..9bd4e00af5 100644
--- a/tests/baselines/reference/jsDeclarationsExportDoubleAssignmentInClosure.js
+++ b/tests/baselines/reference/jsDeclarationsExportDoubleAssignmentInClosure.js
@@ -27,9 +27,36 @@ function foo() {
 
 
 //// [index.d.ts]
+declare namespace _exports {
+    export { m as methods };
+}
 declare function _exports(o: any): any;
 declare namespace _exports {
     export { m as methods };
 }
 export = _exports;
 declare function m(): void;
+
+
+//// [DtsFileErrors]
+
+
+out/index.d.ts(2,19): error TS2300: Duplicate identifier 'methods'.
+out/index.d.ts(6,19): error TS2300: Duplicate identifier 'methods'.
+
+
+==== out/index.d.ts (2 errors) ====
+    declare namespace _exports {
+        export { m as methods };
+                      ~~~~~~~
+!!! error TS2300: Duplicate identifier 'methods'.
+    }
+    declare function _exports(o: any): any;
+    declare namespace _exports {
+        export { m as methods };
+                      ~~~~~~~
+!!! error TS2300: Duplicate identifier 'methods'.
+    }
+    export = _exports;
+    declare function m(): void;
+    
\ No newline at end of file

```
</details>

> Would it make more sense to start from merged's exports and add only specific things from symbol?

Maybe. Some time has past since I was investigating this and it is a little bit blurry to me at this point. Based on the diff above though we can't do that - we could only add **more** this way and with the diff above we already return **too much**. 

Perhaps a different fix can be figured out though.

src/compiler/checker.ts
```diff
@@ -9265,8 +9265,12 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
```
weswigham, line 9271: I think this could potentially have the same symbol in both `exports` and the merged exports through some complicated aliasing. In such a case, it'd behoove us to deduplicate them. Maybe we could just add 'em both to a `Set` and then arrayifying it? The `deduplicate` array helper is... icky - definitely not for things as potentially large as a symbol export table.

weswigham (changes requested):

src/compiler/checker.ts
Andarist: i pushed out the requested change

weswigham (approved):

weswigham:
@typescript-bot run dt
@typescript-bot test top100
@typescript-bot pack this

sandersn (approved):

Issue #56239, opened 2023-10-28
`keyof FilteringMappedType` can't compute its constraint


Andarist:
### 🔎 Search Terms

keyof index constraint filtering mapped 

### 🕗 Version & Regression Information

- This is the behavior in every version I tried

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.3.0-dev.20231027#code/C4TwDgpgBAaghgGwK4QM4B4AqA+KBeKTAbQGsIQB7AM0IF0BuAKEdEigAUAnCgNwEsAJhAEBBAMbAKnfFADejKFFScxALiXBOfAHYBzJooQVdfNVCTaS2igHdtTAL5NGO4BE5U4Y6AGVgcNwBZLwALHQgAYQptKj5dLHFJTlQoCAAPN20BFK5eQWFEqVx5RR0eCjJ1EsUlFXVMQuSiACJlMWaGBSgnRgdmITEEOE5oKgsJPmilCGAkMASJKRT0zOyoACUIMSkBdFRNHV0AGnNLaztsbAAKAH0qrq8k1HuaqCIAaSgdKDJKGgbFslaPVGqgPp1FD0HABKF5QMQjAIQYJiMLaCDqK5dRTbGJxdR+JEotGRaKxeLYmrwZBodDVV6KD5fbQ-cjUQigqBwFKfABkGi0emBckpDNqZneBjFNSMJjMAKe4KlYp6KuwlOwRy60PwuHKgkcTCAA

### 💻 Code

```ts
type Values<T> = T[keyof T];

type ProvidedActor = {
  src: string;
  logic: unknown;
};

interface StateMachineConfig<TActors extends ProvidedActor> {
  invoke: {
    src: TActors["src"];
  };
}

declare function setup<TActors extends Record<string, unknown>>(_: {
  actors: {
    [K in keyof TActors]: TActors[K];
  };
}): {
  createMachine: (
    config: StateMachineConfig<
      Values<{
        [K in keyof TActors as K & string]: {
          src: K;
          logic: TActors[K];
        };
      }>
    >,
  ) => void;
};
```

### 🙁 Actual behavior

```
Type 'Values<{ [K in keyof TActors as K & string]: { src: K; logic: TActors[K]; }; }>' does not satisfy the constraint 'ProvidedActor'.
  Types of property 'src' are incompatible.
[...]

PR #56246, opened 2023-10-28
Improve relationship checking with `keyof GenericMappedType` as the source type

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/56239


src/compiler/checker.ts
```diff
@@ -17468,6 +17468,12 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 const modifiersType = getApparentType(getModifiersTypeFromMappedType(type)); // The 'T' in 'keyof T'
                 forEachMappedTypePropertyKeyTypeAndIndexSignatureKeyType(modifiersType, TypeFlags.StringOrNumberLiteralOrUnique, !!(indexFlags & IndexFlags.StringsOnly), addMemberForKeyType);
             }
+            else if (isFilteringMappedType(type)) {
+                const modifiersIndex = getIndexType(getModifiersTypeFromMappedType(type), indexFlags, UnionReduction.None);
```
Andarist, line 17472: `UnionReduction.None` was added here to keep [this assignment](https://github.com/microsoft/TypeScript/blob/514f7e639a2a8466c075c766ee9857a30ed4e196/tests/cases/compiler/keyRemappingKeyofResult.ts#L42) allowed. 

I **think** that without it the union reduction was forgetting about `'str'` since it's a subtype of `string`. `string` is a valid input `K` but it is meant to be filtered away by `DistributiveNonIndex`. So with the union reduction on, I have ended up removing `'str'` entirely - breaking the test case.

Andarist: I'm somewhat unsure - but maybe the union reduction could still mess this up when another union gets created manually with that as one of its members? Maybe I should explore a different fix that would do the same~ instantiation but later on, when comparing some constraints?

weswigham: `UnionReduction.None` is, in some ways, a hack to make contextual typing work a little bit better (and avoid the perf overhead of subtype reduction we don't need to bother doing). The union members could, very easily, evaporate later on down the line.

IMO, it'd be better to leave this case as producing a generic `keyof` type (especially since this `nameType` could still just be filtering a homomorphic mapped type), and improve constraint-following logic inside the generic-`keyof`-over-mapped-type case in `structuredTypeRelatedToWorker` - specifically, we already have a generic-index-type-target case that does the right thing (tm) - that probably just needs to be brought up and situationally applied when the *source* is a generic index, too. Since our only special case for `keyof` on the source side right now is
```
            else if (sourceFlags & TypeFlags.Index) {
                if (result = isRelatedTo(keyofConstraintType, target, RecursionFlags.Source, reportErrors)) {
                    return result;
                }
            }
```
it could probably use some more specifics, since it's definitely true that generic `keyof`s (over mapped types) can have constraints more specific than `string | number | symbol`.

weswigham (changes requested):

src/compiler/checker.ts
Andarist: Thank you for the advice! I now reused the mentioned logic for `keyof GenericMappedType` as the target type so now the whole fix is within `structuredTypeRelatedToWorker`

weswigham: Hah, I never linked it here, but after I looked over this, I was assigned another issue that tracked back to the same thing, so I ended up writing #56742, which basically scooped this change. Do you want to sync this and at least add the new test cases? In #56742 I used much more specific carve-outs of which `keyof` types this applies to than the very general check here (and actually use a slightly different source key type).

weswigham (changes requested):
Definitely conflicts with #56742 😄

src/compiler/checker.ts
Andarist: https://github.com/microsoft/TypeScript/pull/56742 didn't fix #56239 , see the nightly playground [here](https://www.typescriptlang.org/play?ts=5.4.0-dev.20240110#code/PTAEAsBdIBwZwFwgOYEtLgK4CMB0BjAewFthjV8AnQuQgM0mABUBPGAUwGUrUZHU4cTOzjAArADYATAGYAnACgFkNu1AA1AIYAbYXAA8TAHygAvKCYBtANbsW9CwF0A3EpUdQABWoA3VABN2fwBBfEhCSjNQAG8FUFA4SnwEBMhKVAA7ZFd47UI0ZNBMDOsMwgB3DNcAX1cFTMh2SjpNfDVOSE1GgFlW8Ez2AGFCDLpUZENQ8Mo4UHYAD0aM-1nvQj9AkLCIk1j4zJ9CWxS9+ISklKYpiLhLACJE-DuXONBahWqlQPxtTUo1OjFMKoEYJdiQTAwSbbGZzRbsZazABK7CIlH8+jgaUyyAANEUSmVKkYjAAKAD6J1erWmiBir3ilgA0qBMqBbPY6BZrjNHJcebcmS94u9qgBKKnxKjsLrsXr4foZdgpUkM0BEUbjFIdWXyxVDEZjCZq+JaXQifSnM5nZmsjLsuwOK4w2aaWYsgBkqXSWT59OtAfOhSZOUDZzyBX5LuZwrD70D1SMaqMuNeYrMJkOARqriAA). I'll investigate what's the difference between our approaches in the coming days and report back.

Andarist: The reason this wasn't fixed is that `keyof { [K in keyof TActors as K & string]: { src: K; logic: TActors[K]; }; }` wasn't treated as a deferred index type (although I'd argue that it clearly is deferred). So the logic added by https://github.com/microsoft/TypeScript/pull/56742 couldn't kick in. 

I recently changed that in https://github.com/microsoft/TypeScript/pull/60528 and I think that essentially supersedes this PR so I just merged it in there now to keep the tests.

PR #56247, opened 2023-10-29
Check auto-import completion for spread assignment 

### LINKED ISSUE #56071: No import suggestion when destructuring inside object inside array
### 🔎 Search Terms

 import suggestion destructuring

This is the behavior in both 5.2 and nightly, and I reviewed the FAQ for entries about import suggestions

### ⏯ Playground Link
Linking a repository since it requires multiple file to demonstrate
https://github.com/henribru/typescript-import-suggestion-bug

### 💻 Code

foo.ts:
```
export function foo {]
```

main.ts:
```
foo
{...foo}
[{...foo}]
```

### 🙁 Actual behavior

If you place your cursor after `foo` in the first line of `main.ts` and press ctrl-enter to trigger completions, it will suggest importing `foo` from `foo.ts`. This is also the case on the second line. But on the third line, the import isn't suggested.
image
image
image

### 🙂 Expected behavior

The import should be suggested, since this is perfectly valid code and a place you might want to use an imported variable.

I tried enabling tracing and it looks like the request sent to the language server is basically the same in all three cases, but the response is different in the third case.

hanzooo:
Fixes #56071

### DIFF
--- a/src/services/completions.ts
+++ b/src/services/completions.ts
@@ -3886,7 +3886,8 @@ function getCompletionData(
         // If already typing an import statement, provide completions for it.
         if (importStatementCompletion) return true;
         // If current completion is for non-contextual Object literal shortahands, ignore auto-import symbols
-        if (isNonContextualObjectLiteral) return false;
+        // if current non-contextual Object literal is spread assignment, we may need auto-import symbols
+        if (isNonContextualObjectLiteral && !isSpreadAssignment(contextToken.parent)) return false;
         // If not already a module, must have modules enabled.
         if (!preferences.includeCompletionsForModuleExports) return false;
         // If already using ES modules, OK to continue using them.
--- a/src/services/completions.ts
+++ b/src/services/completions.ts
@@ -3415,7 +3415,6 @@ function getCompletionData(
 
     const semanticStart = timestamp();
     let completionKind = CompletionKind.None;
-    let isNonContextualObjectLiteral = false;
     let hasUnresolvedAutoImports = false;
     // This also gets mutated in nested-functions after the return
     let symbols: Symbol[] = [];
@@ -3885,9 +3884,6 @@ function getCompletionData(
     function shouldOfferImportCompletions(): boolean {
         // If already typing an import statement, provide completions for it.
         if (importStatementCompletion) return true;
-        // If current completion is for non-contextual Object literal shortahands, ignore auto-import symbols
-        // if current non-contextual Object literal is spread assignment, we may need auto-import symbols
-        if (isNonContextualObjectLiteral && !isSpreadAssignment(contextToken.parent)) return false;
         // If not already a module, must have modules enabled.
         if (!preferences.includeCompletionsForModuleExports) return false;
         // If already using ES modules, OK to continue using them.
@@ -4332,7 +4328,6 @@ function getCompletionData(
                 if (objectLikeContainer.flags & NodeFlags.InWithStatement) {
                     return GlobalsSearch.Fail;
                 }
-                isNonContextualObjectLiteral = true;
                 return GlobalsSearch.Continue;
             }
             const completionsType = typeChecker.getContextualType(objectLikeContainer, ContextFlags.Completions);
@@ -4345,7 +4340,6 @@ function getCompletionData(
             if (typeMembers.length === 0) {
                 // Edge case: If NumberIndexType exists
                 if (!hasNumberIndextype) {
-                    isNonContextualObjectLiteral = true;
                     return GlobalsSearch.Continue;
                 }
             }
--- a/tests/cases/fourslash/completionForObjectProperty.ts
+++ b/tests/cases/fourslash/completionForObjectProperty.ts
@@ -12,6 +12,9 @@
 // @Filename: /d.ts
 //// const test3 = [{...foo/*3*/}]
 
+// @Filename: /e.ts
+//// const test4 = { foo/*4*/ }
+
 verify.completions({ 
     marker: "1",
     includes: { name: "foo", source: "/a", hasAction: true, sortText: completion.SortText.AutoImportSuggestions }, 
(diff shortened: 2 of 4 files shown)


src/services/completions.ts
```diff
@@ -3886,7 +3886,8 @@ function getCompletionData(
         // If already typing an import statement, provide completions for it.
         if (importStatementCompletion) return true;
         // If current completion is for non-contextual Object literal shortahands, ignore auto-import symbols
-        if (isNonContextualObjectLiteral) return false;
+        // if current non-contextual Object literal is spread assignment, we may need auto-import symbols
```
DanielRosenwasser, line 3889:
```suggestion
        // if current non-contextual object literal is spread assignment, we may need auto-import symbols
```

src/services/completions.ts
```diff
@@ -3886,7 +3886,8 @@ function getCompletionData(
         // If already typing an import statement, provide completions for it.
         if (importStatementCompletion) return true;
         // If current completion is for non-contextual Object literal shortahands, ignore auto-import symbols
-        if (isNonContextualObjectLiteral) return false;
+        // if current non-contextual Object literal is spread assignment, we may need auto-import symbols
+        if (isNonContextualObjectLiteral && !isSpreadAssignment(contextToken.parent)) return false;
```
DanielRosenwasser, line 3890: I don't understand, how is this branch ever getting hit if the previous check for `isNonContextualObjectLiteral` being true always returns before this one?

tests/cases/fourslash/completionForSpreadAssignment.ts
```diff
@@ -0,0 +1,31 @@
...
+// @Filename: /b.ts
+//// const test = foo/*1*/
+
+// @Filename: /c.ts
+//// const test2 = {...foo/*2*/}
+
+// @Filename: /d.ts
+//// const test3 = [{...foo/*3*/}]
+
+// hereby runtests --tests=completionForSpreadAssignment
```
DanielRosenwasser, line 15:
```suggestion
```

src/services/completions.ts
sandersn, line 3889: 1. The existing comment says to ignore auto-import symbols for non-contextual object literal *shorthands*. Why? A shorthand assignment references a variable the same way that a spread assignment does. Should auto-import work for `{ foo/**/ }` as well?
2. I think the existing comment is wrong since isNonContextualObjectLiteral doesn't appear to check for shorthand assignments.
3. The two comment lines should be combined, and if I'm right about (1) and (2), something like "If current completion is for non-contextual object literal, ignore auto-import symbols for property assignments, but not spread assignments or shorthand property assignments."

sandersn (changes requested):

src/services/completions.ts
hanzooo: 1.  I found that auto-import does not work for `{ foo/**/ }`, I think it would be better if it works. But I haven't found a good way to detect this case, any suggestions?
2. Yes, I think these two comments should be combined.

hanzooo: For `{ foo/**/ }`, currentToken.parent is considered an ObjectLiteralExpression node.

sandersn: Looking back at this, I don't understand why object literals without contextual types would ignore imports at all: they may not have a contextual type, but identifiers mentioned in them might still refer to potential imports.

For the case `{ foo/**/ }` where there is a contextual type, you can guess whether `foo` is a property name or not, and if it is, you don't want to auto-import unless the syntax is clearly a shorthand assignment. I think. But that's not what the code seems to implement. 

Non-contextually typed object literals don't have any information about expected property names -- maybe that's why they're not supposed to auto-import, because it's possible that any shorthand property assignment could turn into a normal property assignment?

@andrewbranch you've spent a lot of time thinking about this. Can you help me figure out what the right thing is?

andrewbranch: I think I agree there’s no strong reason to block an auto-import on a non-contextual object property. We could try removing it.

tests/cases/fourslash/completionForObjectProperty.ts
```diff
@@ -0,0 +1,38 @@
...
+//// const test = foo/*1*/
+
+// @Filename: /c.ts
+//// const test2 = {...foo/*2*/}
+
+// @Filename: /d.ts
+//// const test3 = [{...foo/*3*/}]
+
+// @Filename: /e.ts
+//// const test4 = { foo/*4*/ }
```
sandersn, line 16: Are there tests for `{ foo: /**/ }` and `{ unrelated: foo/**/ }`? I'd expect the first *not* to import and the second to import.
If not, please add them.

sandersn (changes requested):

tests/cases/fourslash/completionForObjectProperty.ts
andrewbranch: I don’t think I understand why you wouldn’t expect to see an import on the former 🤔

hanzooo: I added some test cases. Currently, both cases will import. Do you think it meets expectations?

sandersn: Err, I really underspecified that. Sorry. What I meant was that I didn't expect `{ foo: /*5*/ }` to import `foo`, since it's a property name. It makes sense to import whatever is provided for the property value.

I guess the test case I should have asked for is `{ foo/*7*/: "unrelated" }`, but now I suspect that existing test cases cover this already.

andrewbranch (commented):
Can you add a test case for

```ts
const test7: { foo: string } = { foo/*7*/ }
```

I think we _don’t_ expect an auto-import there.

hanzooo:
> Can you add a test case for
> 
> ```ts
> const test7: { foo: string } = { foo/*7*/ }
> ```
> 
> I think we _don’t_ expect an auto-import there.

I tried this test case and it can make an auto-import. Maybe the type of `{ foo: string }` is not considered here?

tests/cases/fourslash/completionForObjectProperty.ts
hanzooo: Yeah, `{ foo/*7*/: "unrelated" }` is currently covered.

sandersn (approved):

hanzooo:
> > I tried this test case and it can make an auto-import.
> 
> Hm, I just tried it locally and I don’t get an auto-import. Can you add the test so I can see what’s going on?

Yeah, sure. Seens that the test case isn't right. I changed it, it should works now.

andrewbranch (approved):
Thanks @hanzooo!

Issue #51104, opened 2022-10-07
Solutions for Editing Issues with JSDoc in TypeScript Files


DanielRosenwasser:
I've been thinking about a set of specific problems that relate to how JSDoc functions when we edit TypeScript.

# Issues with Unbound Nodes

The first problem is that within a TypeScript file, we don't bind types within JSDoc comments. As a result, later operations in the language service may jump into a type node embedded in JSDoc, assume that some anonymous types have a symbol available, and fail. Two instances of this include https://github.com/microsoft/TypeScript/issues/50750 and https://github.com/microsoft/TypeScript/issues/51054. This isn't ideal, but luckily this isn't very apparent to the user - completion might not work for that request, but the editing experience will still work.

The reason why we don't bind is probably for two reasons - the most obvious one is to avoid doing any more work than is necessary. The other (and more important) reason is that there is some default behavior of *declaration* JSDoc comments where binding them adds a type in the current scope. A type declared by a `@typedef` or a function with a `@constructor` or `@class`. This is not such a huge point of complexity. We could easily make this behavior conditional, or add a separate binding strategy for these code paths.

# Issues with Diagnostics in JSDoc

The second problem is that within a TypeScript file, we may accidentally report errors in JSDoc comments. Some instances of this include https://github.com/microsoft/TypeScript/issues/50341 and https://github.com/microsoft/TypeScript/issues/49109.

For some background knowledge on this issue, we still parse and preserve JSDoc in our TypeScript syntax trees, and plain TypeScript types can be embedded within JSDoc comments. What I believe is happening is that a user can perform editor operations in arbitrary locations, and these editor operations will perform some semantic query in the language service. If a semantic query is performed within a JSDoc comment, it will perform some type-checking which can result in new diagnostics being added to the current file. If the editor requests diagnostics *after* that query, then the inappropriate error will stick around in addition to the usual diagnostics the type-checker will pick up. This used to not be a problem because our language service spun up two type-checkers - one for type-checking, and the other for answering questions about the program; but recently we consolidated them into one. So previously, diagnostics would still be created, they were never user-visible.

# Fixing the Binding Issue

One approach to fixing the issue of unbound anonymous types is just to guard against every possible operation in the language service. While I think that that is one possible approach, it is likely to be spotty and there will always be a few bugs we technically won't catch; plus, it would be nice if the editor operations that are erroring actually did work in JSDoc.

[...]

PR #56257, opened 2023-10-30
Fixed a crash in completions on functionlike types

Andarist:
reported [here](https://github.com/microsoft/TypeScript/issues/56253#issuecomment-1784234276)
closes https://github.com/microsoft/TypeScript/issues/56402

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -1868,7 +1868,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 const nodeLinks = getNodeLinks(node);
                 cachedResolvedSignatures.push([nodeLinks, nodeLinks.resolvedSignature] as const);
                 nodeLinks.resolvedSignature = undefined;
-                if (isFunctionLikeDeclaration(node)) {
+                if (isFunctionExpressionOrArrowFunction(node)) {
                     const symbolLinks = getSymbolLinks(getSymbolOfDeclaration(node));
                     const type = symbolLinks.type;
                     cachedTypes.push([symbolLinks, type] as const);
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -1868,7 +1868,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 const nodeLinks = getNodeLinks(node);
                 cachedResolvedSignatures.push([nodeLinks, nodeLinks.resolvedSignature] as const);
                 nodeLinks.resolvedSignature = undefined;
-                if (isFunctionLike(node)) {
+                if (isFunctionLikeDeclaration(node)) {
                     const symbolLinks = getSymbolLinks(getSymbolOfDeclaration(node));
                     const type = symbolLinks.type;
                     cachedTypes.push([symbolLinks, type] as const);
--- a/src/compiler/utilitiesPublic.ts
+++ b/src/compiler/utilitiesPublic.ts
@@ -1928,8 +1928,8 @@ export function isPropertyAccessOrQualifiedName(node: Node): node is PropertyAcc
 }
 
 /** @internal */
-export function isCallLikeOrFunctionLikeExpression(node: Node): node is CallLikeExpression | SignatureDeclaration {
-    return isCallLikeExpression(node) || isFunctionLike(node);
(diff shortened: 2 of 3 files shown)


src/compiler/utilitiesPublic.ts
```diff
@@ -1928,8 +1928,8 @@ export function isPropertyAccessOrQualifiedName(node: Node): node is PropertyAcc
 }
 
 /** @internal */
-export function isCallLikeOrFunctionLikeExpression(node: Node): node is CallLikeExpression | SignatureDeclaration {
-    return isCallLikeExpression(node) || isFunctionLike(node);
+export function isCallLikeOrFunctionLikeExpression(node: Node): node is CallLikeExpression | FunctionLikeDeclaration {
+    return isCallLikeExpression(node) || isFunctionLikeDeclaration(node);
 }
```
Andarist, line 1933: I believe that this reflects the name better and perhaps the change reflects the original intention

jakebailey: I guess I don't expect this to return true for function declarations, except those in expression positions... am I missing something?

Andarist: Yee, you are right. I limited this further to only return true for function expressions and arrows

DanielRosenwasser: Are you trying to limit to functions that can be contextually typed? What about object literal method declarations?

DanielRosenwasser:
It feels like this is partially related to https://github.com/microsoft/TypeScript/issues/51104. But I'm okay with a limited fix.

src/compiler/utilitiesPublic.ts
Andarist: > Are you trying to limit to functions that can be contextually typed?

Somewhat. IIRC I meant to limit it to expression-space functions and both of those "sets" are almost the same. I think that either would be OK here (at least for the purpose of my fix).

> What about object literal method declarations?

They are already covered by `isFunctionLikeDeclaration` check

Andarist:
> It feels like this is partially related to https://github.com/microsoft/TypeScript/issues/51104. But I'm okay with a limited fix.

Yeah, it looks very much related.

src/compiler/utilitiesPublic.ts
```diff
@@ -1928,8 +1931,8 @@ export function isPropertyAccessOrQualifiedName(node: Node): node is PropertyAcc
 }
 
 /** @internal */
-export function isCallLikeOrFunctionLikeExpression(node: Node): node is CallLikeExpression | SignatureDeclaration {
```
jakebailey, line 1931: Was this function really not used elsewhere? Super surprised, but it does seem to be true.

src/compiler/checker.ts
```diff
@@ -1868,7 +1868,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 const nodeLinks = getNodeLinks(node);
                 cachedResolvedSignatures.push([nodeLinks, nodeLinks.resolvedSignature] as const);
                 nodeLinks.resolvedSignature = undefined;
-                if (isFunctionLike(node)) {
+                if (isFunctionExpressionOrArrowFunction(node)) {
                     const symbolLinks = getSymbolLinks(getSymbolOfDeclaration(node));
```
jakebailey, line 1872: Maybe I'm wrong here, but is it the case that this could also be fixed by instead checking if the declaration has a symbol?

jakebailey (approved):

src/compiler/checker.ts
Andarist: > Maybe I'm wrong here

You are not 😉 

> but is it the case that this could also be fixed by instead checking if the declaration has a symbol?

It could have been fixed like that but the possibility of lacking a symbol isn't reflected by the types. I also just noticed that this kinda happened because I was using a function that was matching more types than I wanted so this just seemed like a slightly better fix.

jakebailey: > the possibility of lacking a symbol isn't reflected by the types

This is an ongoing struggle; I should really resurrect my PR which attempts to make it known-optional.

jakebailey:
@DanielRosenwasser This is a new crash in 5.3; I don't know that it's required pre-release, but I think this one should be backported.

jakebailey:

PR #56258, opened 2023-10-30
Fixed a crash when inferring return type of an accessor with errors in its return statement

Andarist:
fixes what has been reported [here](https://github.com/microsoft/TypeScript/issues/56253#issuecomment-1784234294)

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -7312,7 +7312,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
 
             if (propertySymbol.flags & SymbolFlags.Accessor) {
                 const writeType = getWriteTypeOfSymbol(propertySymbol);
-                if (propertyType !== writeType) {
+                if (propertyType !== writeType && !isErrorType(propertyType) && !isErrorType(writeType)) {
                     const getterDeclaration = getDeclarationOfKind<GetAccessorDeclaration>(propertySymbol, SyntaxKind.GetAccessor)!;
                     const getterSignature = getSignatureFromDeclaration(getterDeclaration);
                     typeElements.push(
--- a/tests/baselines/reference/accessorInferredReturnTypeErrorInReturnStatement.errors.txt
+++ b/tests/baselines/reference/accessorInferredReturnTypeErrorInReturnStatement.errors.txt
@@ -1,0 +1,16 @@
+accessorInferredReturnTypeErrorInReturnStatement.ts(2,7): error TS7023: 'primaryPath' implicitly has return type 'any' because it does not have a return type annotation and is referenced directly or indirectly in one of its return expressions.
+accessorInferredReturnTypeErrorInReturnStatement.ts(4,18): error TS2339: Property 'collection' does not exist on type '{ readonly primaryPath: any; }'.
+
+
+==== accessorInferredReturnTypeErrorInReturnStatement.ts (2 errors) ====
+    export var basePrototype = {
+      get primaryPath() {
+          ~~~~~~~~~~~
+!!! error TS7023: 'primaryPath' implicitly has return type 'any' because it does not have a return type annotation and is referenced directly or indirectly in one of its return expressions.
+        var _this = this;
+        return _this.collection.schema.primaryPath;
+                     ~~~~~~~~~~
+!!! error TS2339: Property 'collection' does not exist on type '{ readonly primaryPath: any; }'.
+      },  
+    };
+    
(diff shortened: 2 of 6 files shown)


src/compiler/checker.ts
```diff
@@ -7312,7 +7312,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
 
             if (propertySymbol.flags & SymbolFlags.Accessor) {
                 const writeType = getWriteTypeOfSymbol(propertySymbol);
-                if (propertyType !== writeType) {
+                if (propertyType !== writeType && !isErrorType(propertyType) && !isErrorType(writeType)) {
```
Andarist, line 7315: I feel like it might be seen as a little bit of an ad-hoc fix. The problem is that a resolved type of a position with a cycle usually is `anyType` but when we are resolving the type initially that type is returned as an error type when the cycle is detected. 

So there is some small mismatch between the return values of functions like `getTypeOfAccessors` - one that depends on the timing of the call to them. I assume that this is intentional.

I noticed that `getWriteTypeOfAccessors` could accidentally-ish return the `errorType` **and** cache it. So I tried to fix it with:
```diff
diff --git a/src/compiler/checker.ts b/src/compiler/checker.ts
index 074c3d3034..59c538e298 100644
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -11758,7 +11758,11 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 writeType = anyType;
             }
             // Absent an explicit setter type annotation we use the read type of the accessor.
-            links.writeType = writeType || getTypeOfAccessors(symbol);
+            if (!writeType) {
+                const readType = getTypeOfAccessors(symbol);
+                writeType = readType !== errorType ? readType : anyType;
+            }
+            links.writeType = writeType;
         }
         return links.writeType;
     }
```

That didn't fix the issue though because when this line (the one that I'm changing here) was hit for the first time we had a situation like this:
```ts
propertyType // errorType
writeType // anyType
```

So the mismatch was still here - it just happened sooner. Originally, the mismatch could happen later when the property type already had a chance to "settle" as `anyType` but the `writeType` was already cached as the `errorType`.

I think this fix is quite fine since `errorType` being returned while resolving the type initially is expected and when that happens we don't quite need to serialize the property as one with divergent accessors. Maybe some further fine-tuning can be done here. I imagine that maybe there is some value in cases that could be serialized as:
```ts
export declare var basePrototype: {
    get primaryPath(): string;
    set primaryPath(v: any); // coming from the error
};
```
I don't have a test case for that at hand though. This whole issue that is being fixed by this PR is a regression so it's worth fixing it sooner than later 😉 

I also think that *maybe* the patch that I posted above might still be worth pulling in since caching `errorType` here looks like something that is not intended. I don't have any test case that would prove it though.

jakebailey:
@typescript-bot test top200
@typescript-bot run dt

@typescript-bot pack this

src/compiler/checker.ts
jakebailey: > The problem is that a resolved type of a position with a cycle usually is anyType

Maybe my memory is shot, but I kinda thought that we always returned `errorType` whenever a cycle occurred? Maybe that's just in the push/pop resolution world?

DanielRosenwasser: I don't think there is anything wrong per se with letting an `errorType` propagate further. It exists to act as an `any` and to specially handle more permissively in other cases.

Andarist: Perhaps this is just somewhat inconsistent across different callers. I see that some similar ones are returning `errorType` just fine. However, for example [here](https://github.com/microsoft/TypeScript/blob/e170bc59d4ba0d335ce86f66296bec71c5018317/src/compiler/checker.ts#L15298-L15300) the cycle is detected and `errorType` is returned immediately but when we climb up the stack to the first "visitor" of this type that `errorType` is converted to `anyType` [here](https://github.com/microsoft/TypeScript/blob/e170bc59d4ba0d335ce86f66296bec71c5018317/src/compiler/checker.ts#L15328). A similar situation ("converting" `errorType` to `anyType`) can be seen in `getTypeOfAccessors` [here](https://github.com/microsoft/TypeScript/blob/e170bc59d4ba0d335ce86f66296bec71c5018317/src/compiler/checker.ts#L11769) and in `getWriteTypeOfAccessors` [here](https://github.com/microsoft/TypeScript/blob/e170bc59d4ba0d335ce86f66296bec71c5018317/src/compiler/checker.ts#L11790).

> I don't think there is anything wrong with letting an errorType propagate further (or get cached or whatever). It exists to act as an any and to specially handle more permissively in other cases.

Ye, it might not be a problem at all. I just tried to follow the pre-existing conventions in other functions.

Andarist: I'll prepare an experiment later that will just return/cache `errorType` in all of those locations that detect the cycle.

PR #56265, opened 2023-10-30
Fix declaration emit in `--moduleResolution bundler` using wrong exports conditions

### LINKED ISSUE #56261: "Inferred type ... cannot be named" after #55725
Pulled out of https://github.com/microsoft/TypeScript/issues/56251#issuecomment-1784211537

Running `tsc -p tsconfig.build.json` (via a script) in https://github.com/vuejs/core with 5.3 causes:

```

> @3.3.7 build-dts /home/jabaile/work/vuejs-core
> tsc -p tsconfig.build.json && rollup -c rollup.dts.config.js

packages/compiler-sfc/src/parse.ts:96:14 - error TS2742: The inferred type of 'parseCache' cannot be named without a reference to '../node_modules/lru-cache/dist/mjs'. This is likely not portable. A type annotation is necessary.

96 export const parseCache = createCache<SFCParseResult>()
                ~~~~~~~~~~


Found 1 error in packages/compiler-sfc/src/parse.ts:96

 ELIFECYCLE  Command failed with exit code 1.
```

This is odd, because `createCache` is defined as:

```ts
import { LRUCache } from 'lru-cache'

export function createCache<T extends {}>(
  max = 500
): Map<string, T> | LRUCache<string, T> {
  if (__GLOBAL__ || __ESM_BROWSER__) {
    return new Map<string, T>()
  }
  return new LRUCache({ max })
}
```

A bisect points to #55725, and we unfortunately didn't run the extended suite there. cc @andrewbranch 

I can fix this with an explicit annotation, but this feels weird.

The funkiness here seems to be that this is a monorepo build of all files in various `/packages/**` dirs, but the output goes into `/temp`, meaning no `d.ts` file's imports resolve properly. Before, the export would get an annotation of `import("lru-cache")`, but now it tries to emit a real import, or something along those lines.

andrewbranch:
Fixes #56261


andrewbranch:
Does this need a cherry-pick in order to make it into the RC?

src/compiler/moduleSpecifiers.ts
```diff
@@ -992,7 +992,8 @@ function tryGetModuleNameAsNodeModule({ path, isRedirect }: ModulePath, { getCan
                 // use the actual directory name, so don't look at `packageJsonContent.name` here.
                 const nodeModulesDirectoryName = packageRootPath.substring(parts.topLevelPackageNameIndex + 1);
                 const packageName = getPackageNameFromTypesPackageName(nodeModulesDirectoryName);
-                const conditions = getConditions(options, importMode === ModuleKind.ESNext);
+                const useImportCondition = importMode === ModuleKind.ESNext || (importMode !== undefined ? false : undefined);
+                const conditions = getConditions(options, useImportCondition);
```
jakebailey, line 996: This bit is duplicated (based on the chagne); should it be a helper somewhere?

jakebailey (approved):

src/compiler/moduleSpecifiers.ts
sheetalkamat: +1

sheetalkamat (approved):

jakebailey (approved):

src/compiler/moduleSpecifiers.ts
jakebailey: Nice refactor!

Issue #56270, opened 2023-10-31
Adding parentheses changes the result of type inference


Max10240:
### 🔎 Search Terms

"constraint evaluation", "type infer"

### 🕗 Version & Regression Information

- This changed between versions __4.9__ and __5.0__

In Version 5.x, parentheses changes the result of type inference;
In Version 4.9 - 4.7(The minimum version that supports `infer X extends ...`), parentheses **not** changes the result of type inference;

### ⏯ Playground Link

[...]

PR #56271, opened 2023-10-31
Defer conditional types with parenthesized multi-element tuple types in `extends` clause

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/56270


src/compiler/checker.ts
```diff
@@ -18312,7 +18313,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         return constraint && (isGenericObjectType(constraint) || isGenericIndexType(constraint)) ? cloneTypeParameter(p) : p;
     }
 
-    function isSimpleTupleType(node: TypeNode) {
+    function isSimpleTupleType(node: TypeNode): node is TupleTypeNode {
```
jakebailey, line 18316: This guard seems unsafe; it doesn't prove the negative

Andarist: I reverted this. I forgot that the negative case should also be considered here since it might have an unwanted effect on the CFA :v

ahejlsberg:
As per my comments [here](https://github.com/microsoft/TypeScript/issues/56270#issuecomment-1787371268) I don't think we want this to change.

Andarist:
Reopening on @jakebailey’s request

jakebailey:
@typescript-bot pack this

jakebailey (approved):

craigphicks:
If we were resolving from the inside out, rather than the outside in, we would also cover cases like `[...[<something>],...[<something>]]` which may resolve to simple tuples.

jakebailey:
I'm not sure if this is what you mean, but potentially a better fix would have been to strip parens around anything passed into `isSimpleTupleType` and any node it examines.

jakebailey:
Well, that can't be the actual implementation, but almost. I'll send a PR for discussion as the parens inside are still significant (and my intention was to not have that be the case).

jakebailey:
I am confusing myself; this doesn't matter; `isSimpleTupleType` looks at the elements and only cares about optional/rest/named members, which cannot be parenthesized. So maybe you meant something else?

craigphicks:
I think "out of scope" covers it.  But what I meant was an inside-to-outside order processing of nodes to types.

PR #54148, opened 2023-05-05
Use union types for all common classes of AST nodes

weswigham:
Like `Node`, `Expression`, `TypeNode`, and the like.

This allows us to make many more `switch` statements in our codebase exhaustive (a change not included in this PR, which just swaps over to unions and fixes resulting errors/lints).

This is like the `Compiler-Unions` perf test, except:
1. Totally complete, rather than just the `compiler` folder.
2. Totally up-to-date (this comment will age well 😂).

Incremental rebuilds are a rebuild-accelerating thing now, so I think this is more palatable nowadays (because a fresh build does _still_ take about 2.7x longer than `main` today does). Especially since using unions for all these types has tangible style and safety advantages (no more casts everywhere!). Immediately, this found some dead code (which is now gone) and some incorrect types (which are now fixed). In addition to the ergonomic niceties of using unions like this, I do think using them more like this would help us dogfood our own typesystem more in scenarios we consider more "extreme", since we have 363 different node kinds now (and usually growing), so the `Node` type is what we should consider a "reasonable" but large type.

There are a handful of places where I had to introduce some new casts to work around some bogus complexity errors (opened #54146 to track the issue causing it, and flagged the offending locations), but other than that this actually works pretty well without significant refactoring (other than automatically removing a ton of useless casts). The big question is really just if we, as a team, think the tradeoffs are worth it at this point or not.

PR #56274, opened 2023-10-31
[Reference] feat: switched Node to a discriminated union

JoshuaKGoldberg:
Posting as reference for now. Still has build failures. Not ready for review. 

Naming is not at all well-thought-out. Names like `NodeBase` are just what came to mind first.

The most breaking portion of this PR's change is that `Node` no longer has a `parent: Node`, because `SourceFile` has `parent: undefined`. In retrospect this probably could be split into >= two steps:

1. Creating a discriminated union, keeping the technically incorrect `parent: Node` inside `SourceFile`
2. Switching `parent: Node` to `parent: undefined` inside `SourceFile`


Jack-Works:
If you're interested, we have discussed this in the engine262 repo. https://github.com/engine262/engine262/pull/215#discussion_r1246023457

I really hope this can land, but I think there might be performance pushback.

jakebailey:
Yes; the previous iteration caused check time to triple: https://github.com/microsoft/TypeScript/pull/54148

JoshuaKGoldberg:
Closing to keep the review queue small.

PR #50377, opened 2022-08-19
Explore adding the Boolean to filter to narrow out null/undefined in array#filter

orta:
From [this tweet](https://twitter.com/mattpocockuk/status/1560594658409029633).

The main issue this attempts to solves is the design to drop null/undefined from an array in an ergonomic way by providing an override for `arr.filter(Boolean)`.

Issue #55777, opened 2023-09-18
Valid filter predicates rejected after 5.2.2 upgrade #53489


craigphicks:
### 🔎 Search Terms

 Add fallback logic for generating signatures for unions of array members #53489 

### 🕗 Version & Regression Information

- This changed between versions 5.1.6 and 5.2.2

### ⏯ Playground Link

[...]

PR #53489, opened 2023-03-24
Add fallback logic for generating signatures for unions of array members

weswigham:
Fixes #44373

With this PR, when we see a type like `(A[] | B[])["member"]` and look for signatures on it, and find none, we reconstruct `(A | B)[]["member"]` and instead look up signatures on that. This _only_ applies to array and tuple types because, well, arrays are special. We treat them as covariant when they're not, and we know their collection-like behavior is consistent. This logic does not apply to any custom subtypes of arrays or array-similar things, like `UInt8Array` - those may have less or more type parameters that both make the reconstruction difficult and potentially break our expectations on how collection-element-like the type parameters are.

This allows things like this:
```ts
// @target: es6
interface Fizz {
    id: number;
    fizz: string;
}

interface Buzz {
    id: number;
    buzz: string;
}

([] as Fizz[] | Buzz[]).filter(item => item.id < 5);
([] as Fizz[] | readonly Buzz[]).filter(item => item.id < 5); 

([] as Fizz[] | Buzz[]).find(item => item);
declare function isFizz(x: unknown): x is Fizz;
([] as Fizz[] | Buzz[]).find(isFizz);
declare function isBuzz(x: unknown): x is Buzz;
([] as Fizz[] | Buzz[]).find(isBuzz);

([] as Fizz[] | Buzz[]).every(item => item.id < 5);

([] as Fizz[] | Buzz[]).reduce(item => item);

([] as [Fizz] | readonly [Buzz?]).filter(item => item?.id < 5);
```
to all work mostly like you'd expect (with maybe the caveat that output arrays are unions of elements within a single array rather than unions of array types).

I'd say that it's _possible_ to broaden this logic to other types using variance measurement logic and some hearty assumptions about how types behave based on their variance, but I feel much better about it making this an array-only change, at least right now.

Issue #50387, opened 2022-08-21
Add BooleanConstructor as an overload to .filter to allow for easy type predicate filtering


mattpocock:
# lib Update Request

## Configuration Check

My compilation *target* is `ES2015` and my *lib* is `the default`.

## Missing / Incorrect Definition

I would love to add an overload to Array.filter which takes in a common use case of passing a `Boolean` directly to `.filter`. This would offer a simple, intuitive workaround for a common pain point with `.filter`.

This could be achieved via an overload adding `BooleanConstructor`, and a type predicate.

https://www.typescriptlang.org/play?#code/PTAEHUEsBcAtTgU1AewG6IE4BsUEMATAWACgBjFAOwGdpQ9NNQBeUAbQCIBGDgGlACulAogBmkSogIBdANykQoJaAB6AflKkJ0LKLxlkAQUZ4AngB4AKgD5QAb1JLx2HZgAUAB0xTIZPDoAuUAAhFBRsRDxKAGEqWkwBMmgUTABKIIA5KgAxPGxqCxs2OVIAX00SaFMPZCzKXPzC21ZLUEQADx1halAABlAAH1AAcmHB0EoBbGxxoRFxSQJQNQnEDCYgy3kSUgoaOmdXKWMmVgZMADpDrDdQ8MjKVO3FZXUgA

## Sample Code

```ts
const arr = ['1', undefined];

const filteredArr = arr.filter(Boolean);
```

`filteredArr` should be `string[]`, not `(string | undefined)[]`.

acutmore:
Related: https://github.com/microsoft/TypeScript/issues/16655

lf-novelt:
A bit convoluted to understand for most people, isn't it?

ghoullier:
see https://github.com/microsoft/TypeScript/pull/50377

mattpocock:
@ghoullier Aha, didn't see that @orta had already tried it.

LMK how I can help.

mattpocock:
> A bit convoluted to understand for most people, isn't it? It looks more like a JS quirks that TS was supposed to prevent than beautiful TS code

Could you clarify what you mean? Which bits are convoluted?

The point of intermediate-advanced TS is to try to make the beginner's TS life easier while staying out of the way. I'd argue that this overload does that very well. `.filter(Boolean)` is a common pattern at all levels of JS, and using it to make TS's inference smarter benefits everyone.

lf-novelt:
[...]

PR #56281, opened 2023-11-01
PR-50377-redo-filterBooleanOverload

### LINKED ISSUE #56013: Adding an overload to .filter breaks specific inference with nested functions
### 🔎 Search Terms

"filter inference overload", "filter boolean"

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about inferred types 

### ⏯ Playground Link

https://tsplay.dev/wQbrvN

### 💻 Code

```ts
type NonFalsy<T> = T extends false | 0 | "" | null | undefined | 0n
    ? never
    : T;

// Comment this out
interface Array<T> { filter(predicate: BooleanConstructor, thisArg?: any): NonFalsy<T>[]; }

const id = <T,>() => (t: T) => !!t;

['foo', 'bar'].filter(id())
//                    ^?
```

### 🙁 Actual behavior

Nested function inference works (type parameter is `string`) without the additional overload, and fails (type parameter is `unknown`) with the overload (even though the added overload isn't the one in use).

### 🙂 Expected behavior

Nested function inference should work regardless of whether an overload is added to .filter or not.

Adding the overload is described in https://github.com/microsoft/TypeScript/issues/50387 and made popular by `ts-reset`. 
It's possible the overload *order* may matter in this instance, but difficult to test when the main overloads are coming from lib definitions.

craigphicks:
Fixes #50377 - allowing `Boolean` as a predicate to Array.filter with falsey types filtered in the result.
Fixes #56013 (*) - Adding an overload to .filter breaks specific inference with nested functions
(* with qualification: The original #56013 added the filter overload externally.  The bug is only fixed for the case where the overload is added internally, as it is in this pull).

## Source Files changed:

- `src/lib/es5.d.ts`
- `src/compiler/checker.ts`

## Added Test Files

- tests/cases/compiler/arrayFilterBooleanOverload#56013WithoutExternalOverload.ts
- tests/cases/compiler/arrayFilterBooleanOverload.ts
- tests/cases/compiler/unionOfArraysBooleanFilterCall.ts
- tests/cases/compiler/arrayMapPredicates1.ts

This one was not added by this pull, but is related:
- tests/cases/fourslash/completionEntryForUnionMethod.ts

## Related issues

filter(Boolean)
#50377 Explore adding the Boolean to filter to narrow out null/undefined in array#filter (closed pull)
#50387 Add BooleanConstructor as an overload to .filter to allow for easy type predicate filtering (inDiscussion)
#30621 narrow array type with .filter(Boolean) (marked as duplicate of #16655, although #16655 is more broad than #30621)
#56013 Adding an overload to .filter breaks specific inference with nested functions (related)
#55777

Boolean as falsey check in general
#16655 Boolean() cannot be used to perform a null check (broader than filter(Boolean))
#29955 Allow Boolean() to be used to perform a null check (accepted pull, may have reverted)

#53489 Add fallback logic for generating signatures for unions of array members

## Source code changes

- `es5.d.ts`: Overload for filter with Boolean predicate is added for both [Array](https://github.com/microsoft/TypeScript/pull/56281/files?file-filters%5B%5D=.ts&show-viewed-files=false#r1382349785) and [ReadonlyArray](https://github.com/microsoft/TypeScript/pull/56281/files?file-filters%5B%5D=.ts&show-viewed-files=false#r1382349772).

- `checker.ts`
    - With new filter overloads, `getSignaturesOfStructuredType()` returns an invalid result of length 1.  The strategy to address that is to piggyback onto #53489.
    - #53489 used a carveout which included a condition that `getSignaturesOfStructuredType()` return a result of length 0.
    - In the interests of generality, [the length condition is dropped altogether](https://github.com/microsoft/TypeScript/pull/56281/files?file-filters%5B%5D=.ts&show-viewed-files=false#r1382421482), rather than just length <= 1, and includes all array methods.
    - The original #53489 carveout logic was "*Transform the type from `(A[] | B[])["member"]` to `(A | B)[]["member"]` (since we pretend array is covariant anyway)*".  This pull modifies some method return types to be a union of the return types of the members, e.g., `ReturnType<(A[])["member"]> | ReturnType<(B[])["member"]>` instead of  `ReturnType<(A|B)["member"]>`, but only in cases where the return type does not have a type parameter.  This change was in response to [this reviewer comment](https://github.com/microsoft/TypeScript/pull/56281#discussion_r1378804648). *(The parameters remain unchanged.)*

## Regressions

1. `Array<string>.map` completion is [changed](https://github.com/microsoft/TypeScript/pull/56281/files?file-filters%5B%5D=.ts&show-viewed-files=false#r1382324409) from
```
(method) Array<T>.map<unknown>(callbackfn: ((value: string, index: number, array: string[]) => unknown) & ((value: number, index: number, array: number[]) => unknown), thisArg?: any): unknown[]
```
to
```
"(method) Array<T>.map<unknown>(callbackfn: (value: string | number, index: number, array: (string | number)[]) => unknown, thisArg?: any): unknown[]"
```

2. `Array<string|number>.map(((value: string, index: number, array: string[]) => unknown) & ((value: number, index: number, array: number[]) => unknown))` is now an [error](https://github.com/microsoft/TypeScript/pull/56281/commits/e904d7772b0990a6169e3c1f8a876e9bcabcc8f0#r1382389001). (Intersection of functions predicate type.) It was previously not an error.

3. `Array<string|number>.map({((value: string, index: number, array: string[]) => unknown), ((value: number, index: number, array: number[]) => unknown)})` is now [error](). (Overload function predicate type.) It was previously not an error.

It could be argued these regressions, occuring only in the rare case where the return is of type `unknown`, are not a serious problem.
Arguable there are benefits to the new completion shown in 1.  A listing of `arrayMapPredicates1` is shown below.

```
type ArrayPredISect<S,N> = ((value: string, index: number, array: string[]) => S) & ((value: number, index: number, array: number[]) => N)

interface ArrayPredOverload <S,N> {
    (value: string, index: number, array: string[]): S,
    (value: number, index: number, array: number[]): N
}

type ArrayPredUnion<S,N> = ((value: string|number, index: number, array: (string|number)[]) => S | N)


declare const pisect1:ArrayPredISect<string,number>
declare const pisect2:ArrayPredISect<unknown,unknown>

declare const poload1:ArrayPredOverload<string,number>
declare const poload2:ArrayPredOverload<unknown,unknown>

declare const punion1:ArrayPredUnion<string,number>
declare const punion2:ArrayPredUnion<unknown,unknown>


declare const asn: string[]|number[];
asn.map(pisect1); // 5.2.2. & #56821 error
asn.map(pisect2); // #56821 error only

asn.map(poload1); // 5.2.2.  & #56821 error
asn.map(poload2); // #56821 error only.

asn.map(punion1);
asn.map(punion2);

declare const asun: (string|number)[];

asun.map(pisect1); // 5.2.2. & #56821 error
asun.map(pisect2); // 5.2.2. & #56821 error

asun.map(poload1); // 5.2.2. & #56821 error
asun.map(poload2); // 5.2.2. & #56821 error

asun.map(punion1);
asun.map(punion2);

declare const astr: string[];
astr.map(pisect1); // 5.2.2.  & #56821 error
astr.map(pisect2);

astr.map(poload1); // 5.2.2.  & #56821 error
astr.map(poload2);

astr.map(punion1);
astr.map(punion2);
```
Here the are the different kinds of Arrays: `Array<string[]|number[]>`, `Array<(string|number)[]>`, `Array<string[]>`,
being tested with two variations each of three different predicate types: `ArrayPredISect`, `ArrayPredOverload`, `ArrayPredUnion`.
The variations are predicate return types of `string`/`number` and `unknown`.

What is noticable is that:
- the errors should not be there.
- only `ArrayPredUnion` predicate type has no errors in all of its 6 combinations.

That implies that a user who is using `ArrayPredUnion` is not likely to encounter any frustrating errors,
and therefore that regression 1 listed above, offering the union predicate completion, is actually a benefit.

## Criticism

1. It could be argued that the scope of the changes in this pull go to far beyond filter Boolean overload in the title, and it has become an omnibus pull.
2. Going in the other direction, it could be argued that the scope of the changes in this pull do not go far enough, are just a patch on a patch, and the real problem lies in the contorted workarounds to pre process the candidates before they enter `chooseOverload`.  The real solution would be to have `chooseOverload` be able to handle the candidates as they are, without preprocessing, in an efficient way.

Both could be true.

Regarding 1, it would be simple to limit the scope of the changes to just the filter Boolean overload.

Regarding 2, inside `chooseOverload`, the arguments are in hand, and impose contraints on the candidates generics constraints, which can therefore be narrowed, or elimanted, or conglomerated on the fly during matching.  That's not something that can be done when the candidates are preprocessed.  That's worth doing anyway.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2651,9 +2651,9 @@
             "dev": true
         },
         "node_modules/istanbul-lib-coverage": {
-            "version": "3.2.1",
-            "resolved": "https://registry.npmjs.org/istanbul-lib-coverage/-/istanbul-lib-coverage-3.2.1.tgz",
-            "integrity": "sha512-opCrKqbthmq3SKZ10mFMQG9dk3fTa3quaOLD35kJa5ejwZHd9xAr+kLuziiZz2cG32s4lMZxNdmdcEQnTDP4+g==",
+            "version": "3.2.2",
+            "resolved": "https://registry.npmjs.org/istanbul-lib-coverage/-/istanbul-lib-coverage-3.2.2.tgz",
+            "integrity": "sha512-O8dpsF+r0WV/8MNRKfnmrtCWhuKjxrq2w+jpzBL5UZKTi2LeVWnWOmWRxFlesJONmc+wLAGvKQZEOanko0LFTg==",
             "dev": true,
             "engines": {
                 "node": ">=8"
@@ -5786,9 +5786,9 @@
             "dev": true
         },
         "istanbul-lib-coverage": {
-            "version": "3.2.1",
-            "resolved": "https://registry.npmjs.org/istanbul-lib-coverage/-/istanbul-lib-coverage-3.2.1.tgz",
-            "integrity": "sha512-opCrKqbthmq3SKZ10mFMQG9dk3fTa3quaOLD35kJa5ejwZHd9xAr+kLuziiZz2cG32s4lMZxNdmdcEQnTDP4+g==",
+            "version": "3.2.2",
+            "resolved": "https://registry.npmjs.org/istanbul-lib-coverage/-/istanbul-lib-coverage-3.2.2.tgz",
+            "integrity": "sha512-O8dpsF+r0WV/8MNRKfnmrtCWhuKjxrq2w+jpzBL5UZKTi2LeVWnWOmWRxFlesJONmc+wLAGvKQZEOanko0LFTg==",
             "dev": true
         },
         "istanbul-lib-report": {


src/compiler/checker.ts
```diff
@@ -14749,19 +14749,38 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
      * maps primitive types and type parameters are to their apparent types.
      */
     function getSignaturesOfType(type: Type, kind: SignatureKind): readonly Signature[] {
-        const result = getSignaturesOfStructuredType(getReducedApparentType(type), kind);
-        if (kind === SignatureKind.Call && !length(result) && type.flags & TypeFlags.Union) {
+        function carveoutResult(type: Type, kind: SignatureKind): readonly Signature[] | undefined {
```
craigphicks, line 14752: The carve out for arrays is raised in precedence.  Previously it only occurred after `result = getSignaturesOfStructuredType` was called, and only if `length(result)` was zero.
It squeezes the (# Overloads X # Template Params) grid of signatures to a (# Overloads X 1) set of overloads by changing, e.g. an (A[] | B[]) union of arrays to a non-union ((A|B)[])  array of a single type union.  Thus there is in increase in such changes in the tests results.

src/lib/es5.d.ts
```diff
@@ -536,6 +536,11 @@ interface BooleanConstructor {
     readonly prototype: Boolean;
 }
 
+interface BooleanConverter {
```
craigphicks, line 539: The reason for adding `BooleanConverter` is to be specific as possible, rather than forcing another function to choose which member of `BooleanConstructor` is the right one.

src/lib/es5.d.ts
```diff
@@ -1257,6 +1262,12 @@ interface ReadonlyArray<T> {
      * @param thisArg An object to which the this keyword can refer in the predicate function. If thisArg is omitted, undefined is used as the this value.
      */
     filter<S extends T>(predicate: (value: T, index: number, array: readonly T[]) => value is S, thisArg?: any): S[];
+    /**
+     * Returns the non-Falsy elements of an array
+     * @param predicate Must be exactly "Boolean"
+     * @param thisArg An object to which the this keyword can refer in the predicate function. If thisArg is omitted, undefined is used as the this value.
+     */
+    filter(predicate: BooleanConverter, thisArg?: any): (T extends false | 0 | "" | null | undefined | 0n ? never : T)[];
```
craigphicks, line 1270: Perhaps `Falsey` (`Falsy`?) could be a utility function.

tests/baselines/reference/arraySlice.types
```diff
@@ -5,7 +5,7 @@ var arr: string[] | number[];
 >arr : string[] | number[]
 
 arr.splice(1, 1);
->arr.splice(1, 1) : string[] | number[]
+>arr.splice(1, 1) : (string | number)[]
```
craigphicks, line 8: array types merged

tests/baselines/reference/bestChoiceType.types
```diff
@@ -4,7 +4,7 @@
 // Repro from #10041
 
 (''.match(/ /) || []).map(s => s.toLowerCase());
->(''.match(/ /) || []).map(s => s.toLowerCase()) : any[]
+>(''.match(/ /) || []).map(s => s.toLowerCase()) : string[]
```
craigphicks, line 7: The change from `any` to `string` is an unexpected bonus.  Probably the `any` was due to an inviable  `getSignaturesOfStructuredType` result.

tests/baselines/reference/controlFlowArrayErrors.errors.txt
```diff
@@ -77,8 +76,6 @@ controlFlowArrayErrors.ts(63,9): error TS7005: Variable 'x' implicitly has an 'a
         }
         x;           // boolean[] | (string | number)[]
         x.push(99);  // Error
```
craigphicks, line 78: The "non-evolving-array-ness" is lost as `boolean[] | (string | number)[]` becomes `(boolean | string | number)[]` in the Array.push call.

tests/baselines/reference/unionOfArraysBooleanFilterCall.types
```diff
@@ -0,0 +1,95 @@
...
+
+    member: string;
+>member : string
+}
+type Falsey = "" | 0 | false | null | undefined;
+>Falsey : false | "" | 0
+>false : false
+
+
+([] as (Fizz|Falsey)[] | (Buzz|Falsey)[]).filter(Boolean); // expect type (Fizz|Buzz)[]
```
craigphicks, line 23: new test case showing falseys are correctly filtered.

tests/baselines/reference/arrayFilterBooleanOverload.types
```diff
@@ -0,0 +1,45 @@
...
+const values2 = nullableValues.filter(new Boolean);
+>values2 : (string | null)[]
+>nullableValues.filter(new Boolean) : (string | null)[]
+>nullableValues.filter : { <S extends string | null>(predicate: (value: string | null, index: number, array: (string | null)[]) => value is S, thisArg?: any): S[]; (predicate: BooleanConverter, thisArg?: any): string[]; (predicate: (value: string | null, index: number, array: (string | null)[]) => unknown, thisArg?: any): (string | null)[]; }
+>nullableValues : (string | null)[]
+>filter : { <S extends string | null>(predicate: (value: string | null, index: number, array: (string | null)[]) => value is S, thisArg?: any): S[]; (predicate: BooleanConverter, thisArg?: any): string[]; (predicate: (value: string | null, index: number, array: (string | null)[]) => unknown, thisArg?: any): (string | null)[]; }
+>new Boolean : Boolean
+>Boolean : BooleanConstructor
+
+const arr = [0, 1, "", "foo", null] as const;
```
craigphicks, line 28: This is a case that brought up by [Ryan Cavanaugh](https://github.com/microsoft/TypeScript/pull/50377#issuecomment-1228975917) in the original #50377.

tests/baselines/reference/unionOfArraysFilterCall.types
```diff
@@ -34,11 +34,11 @@ interface Buzz {
...
 ([] as Fizz[] | readonly Buzz[]).filter(item => item.id < 5); 
 >([] as Fizz[] | readonly Buzz[]).filter(item => item.id < 5) : (Fizz | Buzz)[]
->([] as Fizz[] | readonly Buzz[]).filter : { <S extends Fizz>(predicate: (value: Fizz, index: number, array: Fizz[]) => value is S, thisArg?: any): S[]; (predicate: (value: Fizz, index: number, array: Fizz[]) => unknown, thisArg?: any): Fizz[]; } | { <S extends Buzz>(predicate: (value: Buzz, index: number, array: readonly Buzz[]) => value is S, thisArg?: any): S[]; (predicate: (value: Buzz, index: number, array: readonly Buzz[]) => unknown, thisArg?: any): Buzz[]; }
+>([] as Fizz[] | readonly Buzz[]).filter : { <S extends Fizz>(predicate: (value: Fizz, index: number, array: Fizz[]) => value is S, thisArg?: any): S[]; (predicate: BooleanConverter, thisArg?: any): Fizz[]; (predicate: (value: Fizz, index: number, array: Fizz[]) => unknown, thisArg?: any): Fizz[]; } | { <S extends Buzz>(predicate: (value: Buzz, index: number, array: readonly Buzz[]) => value is S, thisArg?: any): S[]; (predicate: BooleanConverter, thisArg?: any): Buzz[]; (predicate: (value: Buzz, index: number, array: readonly Buzz[]) => unknown, thisArg?: any): Buzz[]; }
```
craigphicks, line 37: Notice that when an Array and ReadonlyArray are merged, the Array dominated the Readonly-ness is lost.  I wonder if it better the other way round?

tests/baselines/reference/arraySlice.types
MartinJohns: Is this change correct? When you have an array that contains either numbers or strings, then splicing shouldn't suddenly result in an array that can contain numbers AND strings.

craigphicks: That's a very good question. This behavior is not new - the "carveout" I mentioned in summary of the pull was originally added in #53489 - 

> With this PR, when we see a type like `(A[] | B[])["member"]` and look for signatures on it, and find none, we reconstruct `(A | B)[]["member"]` and instead look up signatures on that. This only applies to array and tuple types because, well, arrays are special. We treat them as covariant when they're not, and we know their collection-like behavior is consistent. 

When @weswigham says "We treat them as covariant ..." I believe he is talking in the past tense - meaning that even before that pull #53489 there were already cases of such.  (to be continued .... I have to go).

jakebailey:
@typescript-bot test top200
@typescript-bot run dt
@typescript-bot pack this

craigphicks:
> Comparison Report - baseline..pr
Compiler-UnionsTSServer - node (v18.15.0, x64)
Req 5 - completionInfo
Delta  +9ms (+10.18%)

That doesn't look so good.  @jakebailey  - do you agree?
"Unions" fits the description of Arrays with Union Template Types which are processed in `getSignaturesOfType` in the carve out branch.  Currently its taking the carve out branch more than the minimum required to get to the Boolean overload to pass.  I could change that in an attempt to reduce the time. How about it?

jakebailey:
I doubt that's anything but noise.

craigphicks:
Re: top-repos suite

> error TS2322: Type '{ word: string | VimCompleteItem; }[]' is not assignable to type 'ExtendedCompleteItem[]'.
> error TS2322: Type 'string | undefined' is not assignable to type 'string'.

Almost certainly due to the transformation of union-of-arrays to array-of-union-of-types in the carve-out in `getSignaturesOfType`. Those errors would probably be eliminated by making the planned changes described in section "Possible improvements / changes" of the pull description at the top of this page.

src/compiler/checker.ts
```diff
@@ -14748,21 +14748,71 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+                // If the union is all different instantiations of a member of the global array type...
+                let memberName: __String;
+                if (everyType(reducedType, t => !!t.symbol?.parent && isArrayOrTupleSymbol(t.symbol.parent) && (!memberName ? (memberName = t.symbol.escapedName, true) : memberName === t.symbol.escapedName))) {
+                    if ((reducedType as UnionType).arrayFallbackSignatures) {
+                        return (reducedType as UnionType).arrayFallbackSignatures!;
+                    }
+                    //const numberOfOverloads = (reducedType as UnionType).types[0].symbol.declarations?.length;
+
+                    // calculate return types as union of return types over the associated type, rather than return type associated with the union of types
+                    const numTypes = (reducedType as UnionType).types.length;
```
craigphicks, line 14764: Return types in the carveout are now modified  to be "A[] | B[]" rather than "(A | B)[]".

tests/baselines/reference/unionOfArraysBooleanFilterCall.types
```diff
@@ -0,0 +1,95 @@
...
+>([] as (Fizz|Falsey)[] | (Buzz|Falsey)[]).filter(Boolean) : Fizz[] | Buzz[]
```
craigphicks, line 24: Now returns union of arrays `Fizz[] | Buzz[]` rather than `(Fizz|Buzz)[]`.  (The comment "expect type (Fizz|Buzz)[]" is out of date).

tests/baselines/reference/arrayFilterBooleanOverload.types
```diff
@@ -0,0 +1,45 @@
...
+>[0, 1, "", "foo", null] as const : readonly [0, 1, "", "foo", null]
+>[0, 1, "", "foo", null] : readonly [0, 1, "", "foo", null]
+>0 : 0
+>1 : 1
+>"" : ""
+>"foo" : "foo"
+
+const arr2 = arr.filter(Boolean); // expect ("foo" | 1)[]
+>arr2 : (1 | "foo")[]
+>arr.filter(Boolean) : (1 | "foo")[]
```
craigphicks, line 39: This is a case that was brought up by https://github.com/microsoft/TypeScript/pull/50377#issuecomment-1228975917 in the original https://github.com/microsoft/TypeScript/pull/50377.

tests/baselines/reference/unionOfArraysFilterCall.types
```diff
@@ -18,12 +18,12 @@ interface Buzz {
 }
 
 ([] as Fizz[] | Buzz[]).filter(item => item.id < 5);
->([] as Fizz[] | Buzz[]).filter(item => item.id < 5) : (Fizz | Buzz)[]
->([] as Fizz[] | Buzz[]).filter : { <S extends Fizz>(predicate: (value: Fizz, index: number, array: Fizz[]) => value is S, thisArg?: any): S[]; (predicate: (value: Fizz, index: number, array: Fizz[]) => unknown, thisArg?: any): Fizz[]; } | { <S extends Buzz>(predicate: (value: Buzz, index: number, array: Buzz[]) => value is S, thisArg?: any): S[]; (predicate: (value: Buzz, index: number, array: Buzz[]) => unknown, thisArg?: any): Buzz[]; }
+>([] as Fizz[] | Buzz[]).filter(item => item.id < 5) : Fizz[] | Buzz[]
```
craigphicks, line 21: Returns ` Fizz[] | Buzz[]` instead of `(Fizz | Buzz)[]`.

tests/baselines/reference/unionOfArraysFilterCall.types
```diff
@@ -33,12 +33,12 @@ interface Buzz {
 >5 : 5
 
 ([] as Fizz[] | readonly Buzz[]).filter(item => item.id < 5); 
->([] as Fizz[] | readonly Buzz[]).filter(item => item.id < 5) : (Fizz | Buzz)[]
->([] as Fizz[] | readonly Buzz[]).filter : { <S extends Fizz>(predicate: (value: Fizz, index: number, array: Fizz[]) => value is S, thisArg?: any): S[]; (predicate: (value: Fizz, index: number, array: Fizz[]) => unknown, thisArg?: any): Fizz[]; } | { <S extends Buzz>(predicate: (value: Buzz, index: number, array: readonly Buzz[]) => value is S, thisArg?: any): S[]; (predicate: (value: Buzz, index: number, array: readonly Buzz[]) => unknown, thisArg?: any): Buzz[]; }
+>([] as Fizz[] | readonly Buzz[]).filter(item => item.id < 5) : Fizz[] | Buzz[]
```
craigphicks, line 36: `([] as Fizz[] | readonly Buzz[]).filter(item => item.id < 5)` is now returning `Fizz[] | Buzz[]` instead of the previous ` (Fizz | Buzz)[]`, however it should appear as `Fizz[] | readonly Buzz[]`.
Notably, in the original code for `getSignaturesOfType`, which merged the return types to a single array,  there was definitely code there to prioritize "readonly" over "readable".    Whatever "lost" the readonly may be affecting both version.  Warrants investigation.  Because it is not a regression compared to the previous version, it may not be an obstacle to accepting this pull.

tests/cases/fourslash/completionEntryForUnionMethod.ts
```diff
@@ -3,7 +3,7 @@
 ////var y: Array<string>|Array<number>;
 ////y.map/**/(
 
-const text = "(method) Array<T>.map<unknown>(callbackfn: ((value: string, index: number, array: string[]) => unknown) & ((value: number, index: number, array: number[]) => unknown), thisArg?: any): unknown[]";
+const text = "(method) Array<T>.map<unknown>(callbackfn: (value: string | number, index: number, array: (string | number)[]) => unknown, thisArg?: any): unknown[]";
```
craigphicks, line 6: In the short run user are probably better off being directed to write the single function as shown in the "+" (green),
because it will result in less errors. (See test file `arrayMapPredicates1.ts`, to be uploaded soon).
In the longer tun, we need to talk about overloads and modifying `chooseOverload` to return multiple hits - but that's another story.

src/compiler/checker.ts
```diff
@@ -14748,21 +14748,70 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+                            rtset.add(returnType);
+                        });
+                        const art: Type[] = [];
+                        rtset.forEach(rt => art.push(rt));
+                        if (art.length === 0) return undefined;
+                        if (art.length === 1) return art[0];
+                        return getUnionType(art, UnionReduction.Subtype);
+                    });
+
+                    // Transform the type from `(A[] | B[])["member"]` to `(A | B)[]["member"]` (since we pretend array is covariant anyway)
```
craigphicks, line 14787: The original carveout code description - "// Transform the type from `(A[] | B[])["member"]` to `(A | B)[]["member"]` "
In this pull that action is kept for the signature parameters, but the return type is reverted the union of return types, where appropriate.

src/lib/es5.d.ts
```diff
@@ -1257,6 +1257,12 @@ interface ReadonlyArray<T> {
      * @param thisArg An object to which the this keyword can refer in the predicate function. If thisArg is omitted, undefined is used as the this value.
      */
     filter<S extends T>(predicate: (value: T, index: number, array: readonly T[]) => value is S, thisArg?: any): S[];
+    /**
+     * Returns the non-Falsy elements of an array
+     * @param predicate Must be exactly "Boolean"
+     * @param thisArg An object to which the this keyword can refer in the predicate function. If thisArg is omitted, undefined is used as the this value.
+     */
+    filter(predicate: BooleanConstructor, thisArg?: any): (T extends false | 0 | "" | null | undefined | 0n ? never : T)[];
```
craigphicks, line 1265: New overload for ReadonlyArray

src/lib/es5.d.ts
```diff
@@ -1448,6 +1454,12 @@ interface Array<T> {
      * @param thisArg An object to which the this keyword can refer in the predicate function. If thisArg is omitted, undefined is used as the this value.
      */
     filter<S extends T>(predicate: (value: T, index: number, array: T[]) => value is S, thisArg?: any): S[];
+    /**
+     * Returns the non-Falsy elements of an array
+     * @param predicate Must be exactly "Boolean"
+     * @param thisArg An object to which the this keyword can refer in the predicate function. If thisArg is omitted, undefined is used as the this value.
+     */
+    filter(predicate: BooleanConstructor, thisArg?: any): (T extends false | 0 | "" | null | undefined | 0n ? never : T)[];
```
craigphicks, line 1462: New Overload for Array

tests/baselines/reference/arrayFilterBooleanOverload#56013WithoutExternalOverload.types
```diff
@@ -19,31 +19,31 @@ const id = <T,>() => (t: T) => !!t;
...
 >['foo', 'bar'].filter(id()) : string[]
->['foo', 'bar'].filter : { <S extends string>(predicate: (value: string, index: number, array: string[]) => value is S, thisArg?: any): S[]; (predicate: (value: string, index: number, array: string[]) => unknown, thisArg?: any): string[]; }
+>['foo', 'bar'].filter : { <S extends string>(predicate: (value: string, index: number, array: string[]) => value is S, thisArg?: any): S[]; (predicate: BooleanConstructor, thisArg?: any): string[]; (predicate: (value: string, index: number, array: string[]) => unknown, thisArg?: any): string[]; }
 >['foo', 'bar'] : string[]
 >'foo' : "foo"
 >'bar' : "bar"
->filter : { <S extends string>(predicate: (value: string, index: number, array: string[]) => value is S, thisArg?: any): S[]; (predicate: (value: string, index: number, array: string[]) => unknown, thisArg?: any): string[]; }
+>filter : { <S extends string>(predicate: (value: string, index: number, array: string[]) => value is S, thisArg?: any): S[]; (predicate: BooleanConstructor, thisArg?: any): string[]; (predicate: (value: string, index: number, array: string[]) => unknown, thisArg?: any): string[]; }
 >id() : (t: string) => boolean
```
craigphicks, line 27: In #56013 (bug), `id()` was not evaluated when an external `filter(Boolean)` overload was added to the Array interface  in usercode.  Now it works (with the overload in TypeScript code).

tests/baselines/reference/arrayMapPredicates1.errors.txt
```diff
@@ -50,17 +48,29 @@ arrayMapPredicates1.ts(46,10): error TS2345: Argument of type 'ArrayPredOverload
...
-!!! error TS2345: Argument of type 'ArrayPredISect<string, number>' is not assignable to parameter of type '((value: string, index: number, array: string[]) => number) & ((value: number, index: number, array: number[]) => number)'.
-!!! error TS2345:   Type 'ArrayPredISect<string, number>' is not assignable to type '(value: string, index: number, array: string[]) => number'.
-!!! error TS2345:     Type 'string' is not assignable to type 'number'.
+!!! error TS2345: Argument of type 'ArrayPredISect<string, number>' is not assignable to parameter of type '(value: string | number, index: number, array: (string | number)[]) => number'.
+!!! error TS2345:   Types of parameters 'value' and 'value' are incompatible.
+!!! error TS2345:     Type 'string | number' is not assignable to type 'string'.
+!!! error TS2345:       Type 'number' is not assignable to type 'string'.
     asn.map(pisect2);
+            ~~~~~~~
+!!! error TS2345: Argument of type 'ArrayPredISect<unknown, unknown>' is not assignable to parameter of type '(value: string | number, index: number, array: (string | number)[]) => unknown'.
```
craigphicks, line 57: This is one of two added errors.  Note that the previous version is an error whenever the a map function output is more specific than unknown.

tests/baselines/reference/arrayMapPredicates1.errors.txt
```diff
@@ -50,17 +48,29 @@ arrayMapPredicates1.ts(46,10): error TS2345: Argument of type 'ArrayPredOverload
...
-!!! error TS2345: Argument of type 'ArrayPredOverload<string, number>' is not assignable to parameter of type '((value: string, index: number, array: string[]) => number) & ((value: number, index: number, array: number[]) => number)'.
-!!! error TS2345:   Type 'ArrayPredOverload<string, number>' is not assignable to type '(value: string, index: number, array: string[]) => number'.
-!!! error TS2345:     Type 'string' is not assignable to type 'number'.
+!!! error TS2345: Argument of type 'ArrayPredOverload<string, number>' is not assignable to parameter of type '(value: string | number, index: number, array: (string | number)[]) => number'.
+!!! error TS2345:   Types of parameters 'value' and 'value' are incompatible.
+!!! error TS2345:     Type 'string | number' is not assignable to type 'string'.
+!!! error TS2345:       Type 'number' is not assignable to type 'string'.
     asn.map(poload2);
+            ~~~~~~~
+!!! error TS2345: Argument of type 'ArrayPredOverload<unknown, unknown>' is not assignable to parameter of type '(value: string | number, index: number, array: (string | number)[]) => unknown'.
```
craigphicks, line 70: The other added error.

tests/baselines/reference/arrayMapPredicates1.errors.txt
```diff
@@ -50,17 +48,29 @@ arrayMapPredicates1.ts(46,10): error TS2345: Argument of type 'ArrayPredOverload
...
             ~~~~~~~
-!!! error TS2345: Argument of type 'ArrayPredISect<string, number>' is not assignable to parameter of type '((value: string, index: number, array: string[]) => number) & ((value: number, index: number, array: number[]) => number)'.
-!!! error TS2345:   Type 'ArrayPredISect<string, number>' is not assignable to type '(value: string, index: number, array: string[]) => number'.
-!!! error TS2345:     Type 'string' is not assignable to type 'number'.
+!!! error TS2345: Argument of type 'ArrayPredISect<string, number>' is not assignable to parameter of type '(value: string | number, index: number, array: (string | number)[]) => number'.
+!!! error TS2345:   Types of parameters 'value' and 'value' are incompatible.
+!!! error TS2345:     Type 'string | number' is not assignable to type 'string'.
+!!! error TS2345:       Type 'number' is not assignable to type 'string'.
     asn.map(pisect2);
+            ~~~~~~~
```
craigphicks, line 56: Regression

tests/baselines/reference/arrayMapPredicates1.errors.txt
craigphicks, line 70: Regression

tests/baselines/reference/completionEntryForUnionMethod.baseline
```diff
@@ -4,28 +4,28 @@
...
+// | (method) Array<T>.concat(...items: ConcatArray<string | number>[]): string[] | number[] (+1 overload)
 // | (method) Array<T>.every<S>(predicate: (value: string | number, index: number, array: (string | number)[]) => value is S, thisArg?: any): this is S[] (+1 overload)
-// | (method) Array<T>.filter<S>(predicate: (value: string | number, index: number, array: (string | number)[]) => value is S, thisArg?: any): S[] (+1 overload)
-// | (method) Array<T>.forEach(callbackfn: ((value: string, index: number, array: string[]) => void) & ((value: number, index: number, array: number[]) => void), thisArg?: any): void
-// | (method) Array<T>.indexOf(searchElement: never, fromIndex?: number): number
+// | (method) Array<T>.filter<S>(predicate: (value: string | number, index: number, array: (string | number)[]) => value is S, thisArg?: any): S[] (+2 overloads)
+// | (method) Array<T>.forEach(callbackfn: (value: string | number, index: number, array: (string | number)[]) => void, thisArg?: any): void
+// | (method) Array<T>.indexOf(searchElement: string | number, fromIndex?: number): number
 // | (method) Array<T>.join(separator?: string): string
```
craigphicks, line 12: Improvement `never` -> `string | number`

tests/baselines/reference/controlFlowArrayErrors.errors.txt
```diff
@@ -77,8 +76,6 @@ controlFlowArrayErrors.ts(63,9): error TS7005: Variable 'x' implicitly has an 'a
         }
         x;           // boolean[] | (string | number)[]
         x.push(99);  // Error
-               ~~
-!!! error TS2345: Argument of type 'number' is not assignable to parameter of type 'never'.
     }
```
craigphicks, line 80: Possible regression

tests/baselines/reference/implementArrayInterface.errors.txt
```diff
@@ -0,0 +1,45 @@
...
+        indexOf(searchElement: T, fromIndex?: number): number;
+        lastIndexOf(searchElement: T, fromIndex?: number): number;
+        every(callbackfn: (value: T, index: number, array: T[]) => boolean, thisArg?: any): boolean;
+        some(callbackfn: (value: T, index: number, array: T[]) => boolean, thisArg?: any): boolean;
+        forEach(callbackfn: (value: T, index: number, array: T[]) => void, thisArg?: any): void;
+        map<U>(callbackfn: (value: T, index: number, array: T[]) => U, thisArg?: any): U[];
+        filter(callbackfn: (value: T, index: number, array: T[]) => boolean, thisArg?: any): T[];
+        ~~~~~~
+!!! error TS2416: Property 'filter' in type 'MyArray<T>' is not assignable to the same property in base type 'T[]'.
+!!! error TS2416:   Type '(callbackfn: (value: T, index: number, array: T[]) => boolean, thisArg?: any) => T[]' is not assignable to type '{ <S extends T>(predicate: (value: T, index: number, array: T[]) => value is S, thisArg?: any): S[]; (predicate: BooleanConstructor, thisArg?: any): (T extends false | "" | 0 | 0n ? never : T)[]; (predicate: (value: T, index: number, array: T[]) => unknown, thisArg?: any): T[]; }'.
```
craigphicks, line 33: Benign error - informing change in definition of filter.

tests/baselines/reference/unionOfArraysBooleanFilterCall.types
```diff
@@ -21,30 +21,30 @@ type Falsey = "" | 0 | false | null | undefined;
...
+>([] as (Fizz|Falsey)[] | readonly (Buzz|Falsey)[]).filter(Boolean) : Fizz[] | Buzz[]
+>([] as (Fizz|Falsey)[] | readonly (Buzz|Falsey)[]).filter : { <S extends Fizz | Falsey>(predicate: (value: Fizz | Falsey, index: number, array: (Fizz | Falsey)[]) => value is S, thisArg?: any): S[]; (predicate: BooleanConstructor, thisArg?: any): Fizz[]; (predicate: (value: Fizz | Falsey, index: number, array: (Fizz | Falsey)[]) => unknown, thisArg?: any): (Fizz | Falsey)[]; } | { <S extends Buzz | Falsey>(predicate: (value: Buzz | Falsey, index: number, array: readonly (Buzz | Falsey)[]) => value is S, thisArg?: any): S[]; (predicate: BooleanConstructor, thisArg?: any): Buzz[]; (predicate: (value: Buzz | Falsey, index: number, array: readonly (Buzz | Falsey)[]) => unknown, thisArg?: any): (Buzz | Falsey)[]; }
```
craigphicks, line 34: Falsey is correctly filteredy, however the `readonly` trait is not preserved.  Not preserving `readonly` was also a defect in the previous version. so it is not a regression.  Maybe a problem in `getUnionOfTypes()`.

src/compiler/checker.ts
```diff
@@ -14748,21 +14748,70 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
-                const arrayArg = mapType(type, t => getMappedType((isReadonlyArraySymbol(t.symbol.parent) ? globalReadonlyArrayType : globalArrayType).typeParameters![0], (t as AnonymousType).mapper!));
-                const arrayType = createArrayType(arrayArg, someType(type, t => isReadonlyArraySymbol(t.symbol.parent)));
-                return (type as UnionType).arrayFallbackSignatures = getSignaturesOfType(getTypeOfPropertyOfType(arrayType, memberName!)!, kind);
-            }
-            (type as UnionType).arrayFallbackSignatures = result;
+    function getSignaturesOfType(typeIn: Type, kind: SignatureKind): readonly Signature[] {
+        const reducedType = getReducedApparentType(typeIn);
+        function carveoutResult(): readonly Signature[] | undefined {
+            if (kind === SignatureKind.Call && reducedType.flags & TypeFlags.Union) {
+                // If the union is all different instantiations of a member of the global array type...
```
craigphicks, line 14755: This carveout was originally introduced in #53480.  Previously it was only used in the case where `getSignaturesOfStructuredType()` return a length zero result: `!length(result)`.

With new filter overload,  `getSignaturesOfStructuredType()` returns an invalid result of length 1.  Therefore, in order to piggyback onto this carveout to solve the problem, the condition would have to be expanded to inlcude at least length 1, at least for the case of method `filter`.

Considering the benefits of having the type behavior of array methods predicable and uniform, the length condition was dropped altogether, and the extension was not limited to just `filter`.

src/compiler/checker.ts
```diff
@@ -14748,21 +14748,70 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+                let memberName: __String;
+                if (everyType(reducedType, t => !!t.symbol?.parent && isArrayOrTupleSymbol(t.symbol.parent) && (!memberName ? (memberName = t.symbol.escapedName, true) : memberName === t.symbol.escapedName))) {
+                    if ((reducedType as UnionType).arrayFallbackSignatures) {
+                        return (reducedType as UnionType).arrayFallbackSignatures!;
+                    }
+
+                    // calculate return types as union of return types over the associated type, rather than return type associated with the union of types
+                    const numTypes = (reducedType as UnionType).types.length;
```
craigphicks, line 14763: An additional change was made to #53489, such that return types are calculated as union of return types over the associated type, rather than return type associated with the union of types.  E.g., filter returns  "A[] | B[]" rather than "(A | B)[]".

tests/baselines/reference/arraySlice.types
craigphicks: @MartinJohns - Now returns the narrower type as you prefer.

craigphicks:
@typescript-bot test top200
@typescript-bot run dt
@typescript-bot pack this

craigphicks:
> Hey @jakebailey, the results of running the DT tests are ready. There were interesting changes:
> Branch only errors:
> 
> Error: Errors in typescript@local for external dependencies:
> ../winjs/index.d.ts(10075,9): error TS2416: Property 'filter' in type 'QueryCollection<T>' is not assignable to the same property in base type 'T[]'.
>   Type '(callbackfn: (value: T, index: number, array: T[]) => boolean, thisArg?: any) => T[]' is not assignable to type '{ <S extends T>(predicate: (value: T, index: number, array: T[]) => value is S, thisArg?: any): S[]; (predicate: BooleanConstructor, thisArg?: any): (T extends false | ... 2 more ... | 0n ? never : T)[]; (predicate: (value: T, index: number, array: T[]) => unknown, thisArg?: any): T[]; }'.
>     Type 'T[]' is not assignable to type '(T extends false | "" | 0 | 0n ? never : T)[]'.
>       Type 'T' is not assignable to type 'T extends false | "" | 0 | 0n ? never : T'.

@jakebailey Is this a fatal problem?

jakebailey:
Only team members can start these jobs.

@typescript-bot test top200
@typescript-bot run dt

@typescript-bot pack this

To be clear I'm just running this out of interest and haven't read the PR or anything.

craigphicks:
> Heya @jakebailey, I've started to run the diff-based top-repos suite on this PR at [e904d77](https://github.com/microsoft/TypeScript/commit/e904d7772b0990a6169e3c1f8a876e9bcabcc8f0). You can monitor the build [here](https://typescript.visualstudio.com/cf7ac146-d525-443c-b23c-0d58337efebc/_build/results?buildId=158492).
> 
> Update: [The results are in!](https://github.com/microsoft/TypeScript/pull/56281#issuecomment-1793592915)

Upon analysis it appears this error

```
> error TS2322: Type '{ word: string | VimCompleteItem; }[]' is not assignable to type 'ExtendedCompleteItem[]'.
```
is due to the arguments types to `map` being squished together, and nothing to with the result type.
I am going to take this into development while I attempt to fix it.

craigphicks:
In this revision, `es5.d.ts` is not modified at all.  Instead, the return type is changed dynamically in `chooseOverload` in `checker.ts`.  

Therefore, the user won't see a completion suggesting `Boolean`, unlike the last revision.  Obviously that is a disadvantage.   

The advantage is that there are less changes overall, so the pull is easier to review.

craigphicks:
Repoening.  In this revision, es5.d.ts is not modified at all. Instead, the return type is changed dynamically in chooseOverload in checker.ts. Therefore, the user won't see a completion suggesting Boolean, unlike the last revision. Obviously that is a disadvantage. The advantage is that there are less changes overall, so the pull is easier to review.

PR #56293, opened 2023-11-02
pass 'detached:true' for diff tool

rbuckton:
Properly detach subprocess when spawning diff tool for `hereby diff` so that it doesn't terminate immediately when the NodeJS process terminates.

### DIFF
--- a/scripts/build/utils.mjs
+++ b/scripts/build/utils.mjs
@@ -52,6 +52,7 @@ export async function exec(cmd, args, options = {}) {
         }
         else {
             proc.unref();
+            resolve({ exitCode: undefined });
         }
     }));
 }
--- a/scripts/build/utils.mjs
+++ b/scripts/build/utils.mjs
@@ -28,7 +28,7 @@ export async function exec(cmd, args, options = {}) {
         const { ignoreExitCode, waitForExit = true, ignoreStdout } = options;
 
         if (!options.hidePrompt) console.log(`> ${chalk.green(cmd)} ${args.join(" ")}`);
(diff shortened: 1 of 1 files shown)


scripts/build/utils.mjs
```diff
@@ -52,8 +52,6 @@ export async function exec(cmd, args, options = {}) {
         }
         else {
             proc.unref();
-            // wait a short period in order to allow the process to start successfully before Node exits.
-            setTimeout(() => resolve({ exitCode: undefined }), 100);
```
jakebailey, line 56: Doesn't someone still need to resolve this promise? Otherwise hereby will just keep running? (I haven't tested quite yet)

jakebailey (commented):
To be honest I didn't even know anyone used this and considered removing it... I just diff using a shell alias as the diff command we have in CONTRIBUTING is extremely useful elsewhere.

scripts/build/utils.mjs
rbuckton: Ah, fair point. I'll fix.

rbuckton:
> To be honest I didn't even know anyone used this and considered removing it... I just diff using a shell alias as the diff command we have in CONTRIBUTING is extremely useful elsewhere.

I use this quite regularly.

jakebailey:
Tested this out locally, and it seems like output doesn't appear: 
image

But, it also doesn't show up on `main` either...

rbuckton:
Ah, the `diff` build task is primarily designed to launch a GUI diff tool, like Beyond Compare, so it doesn't redirect standard output to the console.

jakebailey:
Oh, gotcha. Not a regression, of course, just not what I thought it was for.

jakebailey (approved):

PR #56294, opened 2023-11-02
fix(56127): TypeScript "Move to File" refactor causes self import

### LINKED ISSUE #56127: TypeScript "Move to File" refactor causes self import
Type: <b>Bug</b>

Take the two files:

#### `1.ts`
```ts
export class Bar {}
export class Baz {}
```
#### `2.ts`
```ts
import { Bar, Baz } from "./1.ts";

const Foo = {
    bar: Bar,
    baz: Baz,
}

export function GetFoo(name:keyof typeof Foo) {
    return Foo[name];

}
```

If you then try and select the `Foo` and refactor > move to file > `1.ts`, will cause `1.ts` to become:

```ts
import { Bar, Baz } from "./1.ts"; // <- bad!

export class Bar {}
export class Baz {}

export const Foo = {
    bar: Bar,
    baz: Baz,
};
```

As you can see `1.ts` is trying to import itsself, which results in the error message (which in TypeScript fashion, is borderline incomprehensible): 

```
Individual declarations in merged declaration 'Bar' must be all exported or all local.
```

VS Code version: Code - Insiders 1.84.0-insider (Universal) (6a7a661757dec1983ff05ef908a2bbb75ce841e0, 2023-10-11T05:35:38.633Z)
OS version: Darwin x64 21.6.0
Modes:

a-tarasyuk:
Fixes #56127

### DIFF
--- a/src/services/refactors/moveToFile.ts
+++ b/src/services/refactors/moveToFile.ts
@@ -206,7 +206,7 @@ function doChange(context: RefactorContext, oldFile: SourceFile, targetFile: str
     else {
         const targetSourceFile = Debug.checkDefined(program.getSourceFile(targetFile));
         const importAdder = codefix.createImportAdder(targetSourceFile, context.program, context.preferences, context.host);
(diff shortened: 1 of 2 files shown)


navya9singh (approved):

andrewbranch (approved):
Thanks @a-tarasyuk!

PR #56296, opened 2023-11-02
Move complex parameter lists of async function into downlevel generator body

### LINKED ISSUE #40410: Behavior change when lowering async function which throws during argument evaluation
**TypeScript Version:** 4.0.2

**Search Terms:** async argument throw emit lower incorrect behavior

**Code**

```ts
(async() => {
  async function f1(x, y = z) {}
  async function f2({[z]: x}) {}
  for (let f of [f1, f2]) {
    let p = f({}) // This should not throw an error
    try {
      await p
    } catch (e) {
      continue // This error is expected
    }
    throw new Error('Expected an error')
  }
})()
```

**Expected behavior:**

The code should run without crashing, both when compiled to `ESNext` and when compiled to `ES2015`.

**Actual behavior:**

The code crashes with `ReferenceError: z is not defined` when compiled to `ES2015`. This is because the generated code moves the evaluation of default arguments from inside the promise context to outside of the promise context during lowering:

```js
function f1(x, y = z) {
  return __awaiter(this, void 0, void 0, function* () { });
}
function f2({ [z]: x }) {
  return __awaiter(this, void 0, void 0, function* () { });
}
```

There are many ways to fix this, but an example correct compilation might look something like this:

```js
function f1(_0) {
  return __awaiter(this, arguments, void 0, function* (x, y = z) { });
}
function f2(_0) {
  return __awaiter(this, arguments, void 0, function* ({ [z]: x }) { });
}
```

The `_0` variables are to ensure `f1.length` and `f2.length` don't change.

I discovered this correctness issue because I replicated TypeScript's approach to async function lowering into [esbuild](https://github.com/evanw/esbuild) but it turned out to have this correctness issue.

**Playground Link:** [link](https://www.typescriptlang.org/play?target=2#code/BQQwzgngdgxsCUACAvAPkQbwFCMeaMiAZgK6wAuAlgPZTECMwAHgDSIQqIBeSGAvjjyRYxMjCq1iAJmAYA2lwC6ALkRM+vAbiLUATomAAbAKbliiakURyi9NkSmLeg3CbMAHTkVkbEAej9EABUAC0owRDAQ6hJDABNEKGozchDdagB3PDpjXXTdF0RyXQ5sXHK8DJBKD0K+RBgQchgQg2NnCtwYWiooEmN-QNDwxFz8xBHjJndjcWM4usLU9KyoYyyAUTy9YAByDenZ8nns0e3dXfhBAQ0ELCA)

**Related Issues:**

rbuckton:
This modifies our down-level emit for async functions when they contain non-trivial parameter lists that may potentially throw when bound. Prior to this change, the following would throw immediately when it should instead produce a rejected promise:

```ts
// @target: es2015
// @lib: esnext
async function f(x, y = doesNotExist) {}
f();
```

The reason for this is that we eagerly evaluate the parameter list in the down-level emit:

```js
function f(x, y = doesNotExist) {
  return __awaiter(this, void 0, void 0, function*() {});
}
```

To address this, this PR repurposes the second parameter to `__awaiter`, which was previously only used to bind a lexically-captured `arguments`, to instead be used to forward arguments for the function to the down-level generator body, producing the following emit instead:

```js
function f(x_1) {
  return __awaiter(this, arguments, void 0, function*(x, y = doesNotExist) {});
}
```

When the transformed function is not an arrow function, we merely forward the function's `arguments` and emit placeholders for each fixed parameter to preserve the function's `length`, as can be seen with `x_1` above.

Arrow functions, however, do not have their own `arguments`. In that case, we instead capture all fixed arguments, followed by a rest parameter that holds any additional arguments.

Thus,
```ts
const f = async (x, y = doesNotExist) => {};
```

now transforms to

```js
const f = (x_1, ...args_1) => __awaiter(this, [x_1, ...args_1], void 0, function*(x, y = doesNotExist) {});
```

This change means that we can no longer rely on the previous mechanism for lexical `arguments` capturing (i.e., just passing it along as the second parameter to `__awaiter`), and instead must manually capture a lexical `arguments` binding.

Thus,

```ts
function f() {
  return async (x, y = doesNotExist) => arguments.length;
}
```

now transforms to

```js
function f() {
  return (x_1, ...args_1) => {
    var arguments_1 = arguments;
    return __awaiter(this, [x_1, ...args_1], void 0, function* (x, y = doesNotExist) {
      return arguments_1.length;
    });
  };
}
```

Aside from the change to lexical `arguments` capturing, a function with a trivial parameter list does not undergo this transformation.

Thus,
```ts
async function f(x, y) {}
```

still transforms to
```js
function f(x, y) { return __awaiter(this, void 0, void 0, function*() {}); }
```

in an effort to limit the complexity of the down-level emit when it is unwarranted.

It should also be noted that this PR applies a similar change to async generators as well.

Fixes #40410


jakebailey:
Just to understand, is this backwards compatible for `tslib` helpers? Given they aren't changing, I think yes?

rbuckton:
> Just to understand, is this backwards compatible for `tslib` helpers? Given they aren't changing, I think yes?

That's correct. No change needs to be made to the helpers, so this is backwards compatible with existing `tslib` versions.

rbuckton:
@typescript-bot perf test

rbuckton:
@typescript-bot test this

jakebailey (approved):

Issue #56303, opened 2023-11-03
`ES2023` missing for `target` in docs?


thernstig:
### Acknowledgement

- [X] I acknowledge that issues using this template may be closed without further explanation at the maintainer's discretion.

### Comment

Caveat: I am not entirely sure where these faults stem from, but they have existed for some time and causes confusion for users. I am starting here.

From: https://www.typescriptlang.org/tsconfig

- `es2023` is missing for `target` (but it should exist, see https://github.com/microsoft/TypeScript/pull/49636)

If we use VS Code it reports that it cannot be used:

![image](https://github.com/microsoft/TypeScript/assets/30827238/ca77ad5a-4e32-4cbb-acd9-644e56c1271e)

But https://json.schemastore.org/tsconfig  has it.

I have no idea where the fault in VS Code comes from as it is/should be checking https://json.schemastore.org/tsconfig, and I also do not know why it was not included on From: https://www.typescriptlang.org/tsconfig.

It is possible I should split this up in more than one bug report, if this is a bug.

RyanCavanaugh:
This is a bit confusing - the error message is auto-generated from the set of possible values that the parser knows about. *Today* `esnext` and `es2023` are the same value, so it doesn't realize it needs to print them back separately since they're not long-term semantically identical.

While we're at it, we should remove `es3` from the printback list.

thernstig:
@RyanCavanaugh note there are two bugs then. The error message from the parse + the docs issue on https://www.typescriptlang.org/tsconfig.

fatcerberus:
I’ve always thought the all-lowercase variants were canonical, fwiw.  I assume the camel-case versions probably originate from the TS source since there’s an enum for it.  Though I guess ultimately the tsconfig parser doesn’t care about the capitalization.

> Today esnext and es2023 are the same value

Wait, that doesn’t seem right.  When I was working with the compiler API `ESNext` always had value 99 while all `ES20xx` were 0-based; there was never a case where one of them overlapped with `ESNext`.  Has that changed?

RyanCavanaugh:
Oh, I got messed up, ignore almost everything up there. OP is confusing `lib` and `target` -- `es2023` is a valid lib, but isn't a valid target because there's nothing different between es2022 and es2023 re: syntax

RyanCavanaugh:
New runtime methods (i.e. things that can be polyfilled) are not syntactic features.

[...]

PR #56304, opened 2023-11-03
Deprecate ES3 in commandline parser

RyanCavanaugh:
I noticed in #56303 that we didn't remove ES3 (or `moduleResolution: node`) from our help printback

### DIFF
--- a/src/executeCommandLine/executeCommandLine.ts
+++ b/src/executeCommandLine/executeCommandLine.ts
@@ -399,6 +399,7 @@ function generateOptionOutput(sys: System, option: CommandLineOption, rightAlign
                     });
                     return Object.entries(inverted)
                         .map(([, synonyms]) => synonyms.join("/"))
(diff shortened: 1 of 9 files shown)


src/executeCommandLine/executeCommandLine.ts
```diff
@@ -395,7 +395,9 @@ function generateOptionOutput(sys: System, option: CommandLineOption, rightAlign
                     // Group synonyms: es6/es2015
                     const inverted: { [value: string]: string[]; } = {};
                     option.type.forEach((value, name) => {
-                        (inverted[value] ||= []).push(name);
+                        if (!(option.deprecatedKeys && option.deprecatedKeys.has(name))) {
```
jakebailey, line 398:
```suggestion
                        if (!option.deprecatedKeys?.has(name)) {
```

Is equivalent?

jakebailey (approved):

PR #56309, opened 2023-11-04
fix inlay hints for template literal type

### LINKED ISSUE #56544: Inlay hints causing a crash on TemplateLiteralType
### 🔎 Search Terms

"Inlay hints", "TemplateLiteralType"

- This is a crash
- This changed between versions master and 5.2.2

### ⏯ Playground Link

_No response_

### 💻 Code

```ts
declare function getTemplateLiteral1(): `${string},${string}`;
const lit1 = getTemplateLiteral1();
```

### 🙁 Actual behavior

it does not display the type of `lit1`
image

### 🙂 Expected behavior

it should display the type of `lit1`
image

`TemplateLiteralType` is not supported by `visitForDisplayParts` yet.
https://github.com/microsoft/TypeScript/actions/runs/6755514395/job/18364019347?pr=56309#step:5:50

Zzzen:
fix #56544


sandersn:
Can you explain what the bug was, what changed, and how it fixes the bug? It would be nice to have a bug tracking this as well.

tests/baselines/reference/inlayHintsInteractiveVariableTypes3.baseline
```diff
@@ -0,0 +1,128 @@
...
+      "text": "`start${"
+    },
+    {
+      "text": "string"
+    },
+    {
+      "text": "}\\${,$${"
+    },
+    {
+      "text": ""
```
MariaSolOs, line 84: This looks wrong. I don't think we should generate empty display parts like these.

Zzzen: Rpdated.

Zzzen:
> Can you explain what the bug was, what changed, and how it fixes the bug? It would be nice to have a bug tracking this as well.

I've added an issue to track the bug mentioned. https://github.com/microsoft/TypeScript/issues/56544

tests/cases/fourslash/inlayHintsInteractiveVariableTypes3.ts
```diff
@@ -0,0 +1,16 @@
+/// <reference path="fourslash.ts" />
```
MariaSolOs, line 1: For smoother maintenance, could you please use a more descriptive name for this test file? E.g. `inlayHintsInteractiveTemplateLiteralTypes.ts`.

MariaSolOs (approved):

Issue #56312, opened 2023-11-05
Not all parameter assignments prevent incorrect dependent narrowings 


Andarist:
### 🔎 Search Terms

parameter symbol assignment binding pattern dependent narrowing variables

### 🕗 Version & Regression Information

- This is the behavior in every version I tried

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.4.0-dev.20231105#code/GYVwdgxgLglg9mABFApgZygCgNoA8A0iAngLoBci2AjIQEwmIA+lAzIQCwkCUiA3gFCJEMYIkwBZAIZQAFgDoATpLAATOAFtMXHgKFCiiALyJaAbkGIAvhZFiDhhyZ0WhuF4gD0HxAD0A-BbW1vxeiMqIKLiS6gAOADYoiABuknEwKogQqXHIMtKI6jAA5jJQiAmSGVBwYYgK4LDqiSgKCnAKwkiyiVlxCRmgkLAI-KgYOGyInFymQA

### 💻 Code

```ts
function test([x, y]: [1, 2] | [3, 4]) {
  if (Math.random()) {
    y = 2;
  }
  if (y === 2) {
    x
    // ^? (parameter) x: 1
  }
}

// an example valid call that might lead to a runtime error in the called function
test([3, 4]);
```

### 🙁 Actual behavior

Assignment to `y` keeps narrowing `x` as if `y` was never mutated. Effectively the binding pattern in this parameter is treated as `const` variable

### 🙂 Expected behavior

I'd expect this assignment to prevent dependent narrowing altogether.

### Additional information about the issue

Assignment to `x` actually prevents dependent narrowing in a scenario like this: [TS playground](https://www.typescriptlang.org/play?ts=5.4.0-dev.20231105#code/GYVwdgxgLglg9mABFApgZygCgNoA8A0iAngLoBci2AjIQEwmIA+lAzIQCwkCUiA3gFCJEMYIkwBZAIZQAFgDoATpLAATOAFtMXHgKFDciALyIqgxAF8zIsUSOHjtHWf3PEAejeIAegH4zlyyA). It's just that those assignments are tracked per symbol and not per the root declaration. I believe that the simplest fix would be to just mark the whole root declaration as reassigned and treat all contained variables as non-const based on that.

PR #52283, opened 2023-01-18
Deprecate isParameter, rename to isParameterDeclaration

jakebailey:
This is a follow-up to #49929, which:

- ~Renames `isParameterDeclaration` to `isPartOfParameterDeclaration`.~ Done in another PR
- Renames `isParameter` to `isParameterDeclaration` (as that is the node name).
- Adds a deprecated compat function `isParameter`.
- Renames `SyntaxKind.Parameter` to `SyntaxKind.ParameterDeclaration` and adds a deprecated duplicate (like we did with JSDoc).

I made up the timing for this deprecation; I just copied the same scheme as another helper's but I feel like this one needs a loooot of time for people to fix.

This may not be the safest thing because `isParameterDeclaration` was previously a different function, so people may use the new one while still using old TS versions. Not great.

PR #56313, opened 2023-11-05
Avoid dependent parameters narrowings if any declared symbol of the parameter is assigned to

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/56312


gabritto:
@typescript-bot run DT
@typescript-bot test top100

gabritto:
@typescript-bot pack this

src/compiler/checker.ts
```diff
@@ -28527,37 +28527,33 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
-        return symbol.isAssigned || false;
+        return getNodeLinks(rootDeclaration).someSymbolAssigned;
     }
 
     function hasParentWithAssignmentsMarked(node: Node) {
         return !!findAncestor(node.parent, node => (isFunctionLike(node) || isCatchClause(node)) && !!(getNodeLinks(node).flags & NodeCheckFlags.AssignmentsMarked));
     }
 
-    function markNodeAssignments(node: Node) {
+    function markNodeAssignments(node: Node): true | undefined {
```
gabritto, line 28546: Can we not `return` in this function, the way it was before? I think it's not really necessary and potentially confusing, since we don't use the results returned by this function.

src/compiler/checker.ts
```diff
@@ -28921,7 +28918,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         while (
             flowContainer !== declarationContainer && (flowContainer.kind === SyntaxKind.FunctionExpression ||
                 flowContainer.kind === SyntaxKind.ArrowFunction || isObjectLiteralOrClassExpressionMethodOrAccessor(flowContainer)) &&
-            (isConstantVariable(localOrExportSymbol) && type !== autoArrayType || isParameter && !isSymbolAssigned(localOrExportSymbol))
+            (isConstantVariable(localOrExportSymbol) && type !== autoArrayType || isParameter && !isSomeSymbolAssigned(rootDeclaration))
```
gabritto, line 28921: It looks like the change to `isSymbolAssigned` behavior, to aggregate `isAssigned` at the root declaration level, might not be the correct thing to do here, for this particular usage of `isSymbolAssigned`, as evidenced by this breaking change: https://github.com/microsoft/TypeScript/pull/56313#issuecomment-1841559785.

Maybe we want to keep marking symbols with `isAssigned`, and also aggregate that info at the root declaration level, since `getNarrowedTypeOfSymbol` needs the aggregate info but `checkIdentifier` needs the symbol info.

Here's a more minimal repro of the break linked above:
```ts
function ff({ a, b }: { a: string | undefined, b: () => void }) {
  if (a !== undefined) {
    b = () => {
      const x: string = a;
    }
  }
}
```

gabritto (changes requested):

src/compiler/checker.ts
```diff
@@ -28766,7 +28762,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     const contextualSignature = getContextualSignature(func);
                     if (contextualSignature && contextualSignature.parameters.length === 1 && signatureHasRestParameter(contextualSignature)) {
                         const restType = getReducedApparentType(instantiateType(getTypeOfSymbol(contextualSignature.parameters[0]), getInferenceContext(func)?.nonFixingMapper));
-                        if (restType.flags & TypeFlags.Union && everyType(restType, isTupleType) && !isSymbolAssigned(symbol)) {
+                        if (restType.flags & TypeFlags.Union && everyType(restType, isTupleType) && !isSomeSymbolAssigned(declaration)) {
```
gabritto, line 28765: I think this case here is not fixed by aggregating `isAssigned` at root declaration level, since the root declaration will be a parameter declaration, but the other dependent parameter could have been reassigned:

```ts
const test2: (...args: [1, 2] | [3, 4]) => void = (x, y) => {
  if (Math.random()) {
    y = 2;
  }
  if (y === 2) {
    x
    // ^? (parameter) x: 1
  }
}
```

src/compiler/checker.ts
Andarist: > since we don't use the results returned by this function.

It is used by `forEachChild`. This `return` acts as an early return for its iteration.

That being said, I probably need to mark both individual symbols as assigned or not together with the aggregate information on the root declaration to fix the regression found [here](https://github.com/microsoft/TypeScript/pull/56313#issuecomment-1841559785).

A minimal repro case for this regression:
```ts
declare function useRouter(): {
    push: (href: string) => void;
}

type SidebarFooterButtonProps = {
  href?: string;
  onClick?: () => void;
};

export const SidebarFooterButton = ({
  href,
  onClick,
}: SidebarFooterButtonProps): JSX.Element => {
  const router = useRouter();

  if (href !== undefined) {
    onClick = () => {
      void router.push(href);
    };
  }

  return (
    <button type="button" onClick={onClick} />
  );
};
```

We still want to narrow "const" parameters and only turn off the dependent parameter narrowing if some of the relevant symbols are reassigned - so we need to keep track of both (or well, derive the aggregate information on demand from the root declaration's symbols)

src/compiler/checker.ts
Andarist: Heh, I really need to start reading through all of the existing comments at once instead of reading through them one by one. It would save me a few minutes here 😅

src/compiler/checker.ts
Andarist: A great catch! The "dependent root declaration" might come from the context... I'll have to think about this and recheck how this kind of narrowing is handled today.

src/compiler/utilities.ts
```diff
@@ -5278,7 +5278,7 @@ export function isPushOrUnshiftIdentifier(node: Identifier) {
  *
  * @internal
  */
-export function isParameterDeclaration(node: Declaration): boolean {
+export function isParameterDeclaration(node: Declaration): node is ParameterDeclaration {
```
gabritto, line 5281: Did you forget to undo this maybe? I think despite the terribly misleading name, this function is not meant to return whether `node is ParameterDeclaration`.

gabritto (changes requested):

gabritto:
@typescript-bot test top100

src/compiler/utilities.ts
Andarist: Yeah, totally! I could swear that I have reverted this... but I guess I didn't 😬 I fixed this just now.

gabritto (approved):

src/compiler/utilities.ts
jakebailey: Yeah, this definitely is not a type guard. See also #52283.

Issue #55495, opened 2023-08-24
Quick info displays incorrect type in discriminated union type


kettanaito:
### 🔎 Search Terms

quick info, union, discriminated union, wrong info.

### 🕗 Version & Regression Information

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about 

### ⏯ Playground Link

https://www.typescriptlang.org/play?#code/C4TwDgpgBAGlC8UDeAoKUB2BDAthAXFAOQBWA9gBYZFQA+xA1lsBDfQM7ABOAlhgOZQAZMgC+KUcKgAKNMjnY8hTrwFSk49AG0GEEMu59+AXUJYMICXXnpFBYuSpE57MngAKXMmEIAjMmQANhDmEnL0qOi2uPZETCzOUVDxEADKkADGPABmPBkA0nqESFAZZACuGMCEGOU4vhBcUJrNKACUKCjSJXaEROYgALR2RAA0UFiEAIzjvoTc5RDjrkpQUwBMAMzNbVDszDzsuRDssJ0A9OdQABJkAG6NUPePAEQrEJ7eLwB0UAAqFGgAEdynkGFA+NkyHssCBTjxgERTv4giEMChLlAAO4UHjBcZ8KDZLAZYAExGnF4DF4yYmBQKnSoAE0eWCgODIXGgYC5OEO7B4DwhGBZAA89jx+NhgOUuW1vl0ejE+o5qMs3B8vD4oAtoKJdvtgIdjqcYJ0UEA

### 💻 Code

```ts
type X = {
  name: 'john' | 'kate' | string & {}
} & (
  {
  name: string & {}
  [key: string]: any
} | {
  name: 'john'
  someProp: boolean
}
  | {
    name: 'kate'
    kateSpecificKey: { count: number }
  }
)

({ name: 'any-name', a: 1, b: true, some: 123 }) satisfies X

// Hover over "someProp". The quick info says it's boolean
// while, in fact, it's "any" (falls under a more premissive index signature).
({ name: 'john', someProp: true }) satisfies X



```

### 🙁 Actual behavior

The quick info prop for `someProp` displays `boolean` as if the object matches the discriminated union. It doesn't, as you can observe from the actual type validation—`someProp` is still `any`, matching a more permissive index signature. 

### 🙂 Expected behavior

The quick info must display `someProp` correctly—as `any`. 

### Additional information about the issue

This is a purely visual issue. The type validation is performed correctly.

[...]

PR #56318, opened 2023-11-05
Fixed an issue with property type display when contextual type is a union

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/55495

### DIFF
--- a/src/services/services.ts
+++ b/src/services/services.ts
@@ -3273,12 +3273,15 @@ export function getPropertySymbolsFromContextualType(node: ObjectLiteralElementW
         return symbol ? [symbol] : emptyArray;
     }
 
-    const discriminatedPropertySymbols = mapDefined(contextualType.types, t => (isObjectLiteralExpression(node.parent) || isJsxAttributes(node.parent)) && checker.isTypeInvalidDueToUnionDiscriminant(t, node.parent) ? undefined : t.getProperty(name));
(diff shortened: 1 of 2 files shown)


src/services/services.ts
```diff
@@ -3273,12 +3273,15 @@ export function getPropertySymbolsFromContextualType(node: ObjectLiteralElementW
...
+    const filteredTypes = isObjectLiteralExpression(node.parent) || isJsxAttributes(node.parent)
+        ? filter(contextualType.types, t => !checker.isTypeInvalidDueToUnionDiscriminant(t, node.parent))
+        : contextualType.types;
+    const discriminatedPropertySymbols = mapDefined(filteredTypes, t => t.getProperty(name));
     if (unionSymbolOk && (discriminatedPropertySymbols.length === 0 || discriminatedPropertySymbols.length === contextualType.types.length)) {
         const symbol = contextualType.getProperty(name);
         if (symbol) return [symbol];
     }
-    if (discriminatedPropertySymbols.length === 0) {
+    if (!filteredTypes.length && !discriminatedPropertySymbols.length) {
```
Andarist, line 3284: The main idea behind the fix is that we don't want to go into this branch because the type satisfies its contextual type in this test case. So there is one filtered type here **but** there are no discriminated property symbols. The first type in the union doesn't have an explicit prop for `someProp` and the second one is rejected by the `isTypeInvalidDueToUnionDiscriminant` check.

andrewbranch (approved):

PR #56340, opened 2023-11-08
Fix declaration emit for JS default re-exports that resolve to modules through synthesized default exports

### LINKED ISSUE #55082: Re-exporting `default` from a dependency in JavaScript produces an invalid declaration file
# Bug Report

### 🔎 Search Terms

Re-export, default, declaration, `module.exports`, Identifier expected ts(1003), js, JavaScript.

- This affects `Bundler` since version `5.0`.
- This affects `Node16`/`NodeNext` since version `4.7`.
- This affects `System` since version `3.7`.

### ⏯ Playground Link

[Workbench Repro](https://www.typescriptlang.org/dev/bug-workbench/?checkJs=true&moduleResolution=3&module=100&emitDeclarationOnly=true&listEmittedFiles=true#code/PTAEAEGMAsFNINYCkDOAoEEC2B7AJgK4A2sAXKAHL6wCMAbBmOLoSQEqwo5EEAuAljgB25KnloNM4cZCIBDAE5yBwxhFhZ+vACLx5SlUIDyQogE80a8ADN+JIXKxlQ-IeIAeAOgBW6WO4AHHAVeUABvcWs5YlC5FFAWAF9QawUcLFAAcjlMgG4rW3tHZwC5RDkAc1gfLiE0MIAiAIV+ADdlWAbyXgUCWAAaUAbeMwDO8gaWYk7EgrtYBydyIWoAfSmSFGA5YFLyqprVRsXxobkGwYbW2AUUQSEuoYAGTxenhtmpQoXi5bWNzjbYCuDw1ND+IIhFCeOSgAC8oBoACYAMz5L7zE5-cTrfDTLY7EH+Tx4Ty8PyBYKhSDCFCxeGI1G5IA)

Switch to the “Debug” tab and look at `/index.d.ts`. `@showEmittedFile` did not work.

### 💻 Code

```ts repro
// @checkJs
// @module: Node16
// @moduleResolution: Node16
// @declaration
// @emitDeclarationOnly

// @filename: index.js
export {default as mod} from 'a';
// @filename: package.json
{"private": true, "type": "module"}
// @filename: node_modules/a/package.json
{"name": "a", "version": "0.0.0"}
// @filename: node_modules/a/index.js
exports.a = 123;
// @filename: node_modules/a/index.d.ts
export const a = 123;
```

Other cases like this:

- Indirect re-export\
  (Only the declaration file that re-exports a dependency will be invalid.)
- `index.js`: `export {default} from 'a';`\
  (Declaration file will have as default)
- `node_modules/a/index.d.ts` may be provided separately (e.g., by a `@types/` package).
- [My comment below](https://github.com/microsoft/TypeScript/issues/55082#issuecomment-1645402382)

### 🙁 Actual behavior

Generated `index.d.ts` is invalid because, instead of re-exporting `default as mod`, it re-exports `"/path/to/node_modules/a/index" as mod`. Firstly, there is no export on `a/index.js` with that identifier. Secondly, TypeScript does not even support “arbitrary module namespace identifier names” (#40594), so the resulting declaration file cannot be subsequently parsed anyway. (Errs `Identifier expected. ts(1003)`.)

### 🙂 Expected behavior

Generated `index.d.ts` should simply re-export `default as mod`, just like the JavaScript file. If `node_modules/a` is ESM, it should err that there is no export named `default`.

andrewbranch:
Fixes #55082 

I think this is too narrowly targeted—@weswigham is this on the right track?


src/compiler/checker.ts
```diff
@@ -9707,6 +9707,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                         // does not use localName because the symbol name in this case refers to the name in the exports table,
                         // which we must exactly preserve
                         const specifier = (node.parent.parent as ExportDeclaration).moduleSpecifier;
+                        if (specifier && target?.flags & SymbolFlags.ValueModule && (node as ExportSpecifier).propertyName?.escapedText === InternalSymbolName.Default) {
```
weswigham, line 9710: `target?.flags & SymbolFlags.ValueModule` is also set for namespaces (via `bindModuleDeclaration`), which we don't want to trigger this - honestly, I think the reliable indicators would be either the symbol declarations having a source file or ambient module declaration in them.

weswigham (changes requested):
You're definitely right that this feels a *little* bespoke; our ability to identify synthetic default symbols is a bit rough, since the module symbol is just returned for the `default` member. One day we should try swapping that to a generated `Alias` symbol, or at least making a bespoke symbol for the export, rather than just returning the module symbol directly.

Still, for now an approach like this one is probably fine.

andrewbranch:
> One day we should try swapping that to a generated `Alias` symbol, or at least making a bespoke symbol for the export, rather than just returning the module symbol directly.

This would make things much easier to reason about. I wonder how much memory penalty it would incur.

src/compiler/checker.ts
andrewbranch: The thing that feels weird about the line of code I wrote here is that it’s checking both the symbol structure and the original syntax. It seems like when the JS contains ESM syntax like this, the desired outcome is just to copy the syntax verbatim, but maybe there are edge cases where that’s not true? On one hand, it feels possible to do only a check of the symbol like you’re suggesting and _assert_ that the export specifier was `default`, but on the other hand, I think I could also not check the symbol _at all_ and just set the target name to `default` every time the original syntax did a `default` re-export.

andrewbranch:
@weswigham can you take a look at my latest commit?

Issue #56341, opened 2023-11-08
Duplicate property error when a computed property names come from an object destructuring with the same default value 


bradzacher:
### 🔎 Search Terms

"An object literal cannot have multiple properties with the same name" object destructure destructuring computed

### 🕗 Version & Regression Information

- This changed between versions 5.0.4 and 5.1.6

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.2.2#code/MYGwhgzhAEAqCmEAu0DeAoa0AOAnA9thAIQBcaAHgPznK4CWAdgOYA00AnjdHU8wL4BuTNAC28JAAt8AEwAUASjQiswfI2RpoFaAF5oAcgPsOewwf5mp9CADo8hCMKxZcEgK65Gyly4DaFAC65ACMrCr+HMHQAEzhvkIi-OjJ6ABm7ozASPTq0ABGYLhyDkTkqNS0SAwsJty8LPxKGL5qGiio2mZGJt0WZqVOEW5Int4tvtAB0WERWH5R5HERiakZWTl5afj4cp06+j2cffzllTzVfHVVNQLNIiNjPv5BofGR0ctYq+hAA

### 💻 Code

```ts
function bar(props: {x?: string, y?: string}) {
    const { x = '', y = ''} = props;
    return {
      [x]: 1,
      [y]: 2,
    };
}
```

```ts
function foo({ x = '', y = ''}: {x?: string, y?: string}) {
  return {
    [x]: 1,
    [y]: 2,
  };
}
```

### 🙁 Actual behavior

In the first example - the `[y]: 2` line reports `An object literal cannot have multiple properties with the same name.(1117)`

In the second example - the `[y]: 2` line reports nothing.

### 🙂 Expected behavior

There should be no reports as both variables are typed as `string`.

### Additional information about the issue

[...]

PR #56347, opened 2023-11-08
Avoid incorrect narrowings using const variables from binding elemens with literal initializers

Andarist:
spotted this while investigating why https://github.com/microsoft/TypeScript/pull/55193 fixed https://github.com/microsoft/TypeScript/issues/56341

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -26206,7 +26206,8 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         if (hasOnlyExpressionInitializer(declaration) && isBlockScopedNameDeclaredBeforeUse(declaration, node)) {
             const initializer = getEffectiveInitializer(declaration);
             if (initializer) {
-                return tryGetNameFromType(getTypeOfExpression(initializer));
+                const initializerType = isBindingPattern(declaration.parent) ? getTypeForBindingElement(declaration as BindingElement) : getTypeOfExpression(initializer);
+                return initializerType && tryGetNameFromType(initializerType);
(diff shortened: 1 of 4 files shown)


src/compiler/checker.ts
```diff
@@ -26206,7 +26206,8 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         if (hasOnlyExpressionInitializer(declaration) && isBlockScopedNameDeclaredBeforeUse(declaration, node)) {
             const initializer = getEffectiveInitializer(declaration);
             if (initializer) {
-                return tryGetNameFromType(getTypeOfExpression(initializer));
+                const initializerType = isBindingPattern(declaration.parent) ? getTypeForBindingElement(declaration as BindingElement) : getTypeOfExpression(initializer);
```
jakebailey, line 26209: I feel like I see this `isBindingPattern(declaration.parent) ? getTypeForBindingElement(declaration)` pattern over and over. I'm not sure there's a clean way to abstract it generally speaking but it seems odd to have to know to do this all over the place.

jakebailey (approved):

src/compiler/checker.ts
Andarist: Yeah, I just borrowed this from `getTypeForVariableLikeDeclaration`. It would be cool to not worry about this.

sandersn:
@jakebailey Is this ready to merge for 5.4?

jakebailey:
Yeah though I should run

@typescript-bot test top200
@typescript-bot run dt

PR #56354, opened 2023-11-09
isolatedModules error on alias merging with local value

### LINKED ISSUE #55584: imported symbol that is shadowed with a local symbol does not error
### 🔎 Search Terms

imported symbol shadowed

### ⏯ Playground Link

https://www.typescriptlang.org/play?isolatedModules=true#code/JYWwDg9gTgLgBAbzgYQuCA7Aph+BfOAMyjTgHIosBDAYxjIG44B6ZuAFQGU4soSoAznExwYAC2BCqQ4PSlwAblQA2AVywAoZVnip02XAC4UaSAfgAfOBlXLlcALzXbyhho2hIsRHABiyOAJiUgpqOkYWNi44FWUIAHchcUkYmTkY0QBPME1teH9jfzgrGztHZzs3IA

### 💻 Code

```ts
import { Component } from 'react';
let Component: Component | null = null;

import { FC } from 'react';
let FC: FC | null = null;
```

### 🙁 Actual behavior

The first case correctly errors:
```
Import declaration conflicts with local declaration of 'Component'. (2440)
```

The second case is allowed, even under `isolatedModules`.

### 🙂 Expected behavior

Both cases should error

Looking at this from two perspectives - `isolatedModules: true` and `false`

For `isolatedModules: false` - this looks like it's working as expected. In a fully type-aware build TS can tell that the local value symbol shadows an imported type symbol - which is valid - and it can handle it just fine and knows it 100% can elide the import.

However for `isolatedModules: true` - I'd expect that TS acts stricter here. Without type information it's not really safe to assume that the imported name is a type.

From TS's POV I believe that when it transpiles the code it sees the imported symbol is never used as a value (due to the local symbol), so it elides the import.
This accidentally makes the code runtime-valid so it won't crash.

I'm of the opinion here that TS should always error on this case or else it's possible to write some really cooked looking code like

```ts
import { FC } from 'react';
let FC: FC | null = null;
   FC = null;
// ^^ assigning to an import... but also not technically
```

andrewbranch:
Fixes #55584

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -45423,6 +45423,19 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     Diagnostics.Import_declaration_conflicts_with_local_declaration_of_0;
                 error(node, message, symbolToString(symbol));
             }
+            else if (node.kind !== SyntaxKind.ExportSpecifier) {
+                // Look at 'compilerOptions.isolatedModules' and not 'getIsolatedModules(...)' (which considers 'verbatimModuleSyntax')
+                // here because 'verbatimModuleSyntax' will already have an error for importing a type without 'import type'.
+                const appearsValueyToTranspiler = compilerOptions.isolatedModules && !findAncestor(node, isTypeOnlyImportOrExportDeclaration);
+                if (appearsValueyToTranspiler && symbol.flags & (SymbolFlags.Value | SymbolFlags.ExportValue)) {
+                    error(
+                        node,
+                        Diagnostics.Import_0_conflicts_with_local_value_so_must_be_declared_with_a_type_only_import_when_isolatedModules_is_enabled,
+                        symbolToString(symbol),
+                        isolatedModulesLikeFlagName,
+                    );
+                }
+            }
 
             if (
                 getIsolatedModules(compilerOptions)
--- a/src/compiler/diagnosticMessages.json
+++ b/src/compiler/diagnosticMessages.json
@@ -3711,8 +3711,12 @@
         "category": "Error",
         "code": 2864
     },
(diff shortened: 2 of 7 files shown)


andrewbranch:
@typescript-bot test top200
@typescript-bot user test
@typescript-bot perf test

andrewbranch:
@typescript-bot test top200

andrewbranch:
I’m kind of ambivalent on this. Should `isolatedModules` be slightly stricter than necessary and break these projects to help swc and eslint, or should these examples be used as evidence that transpilers can handle this pattern and close the issue as Working as Intended?

jakebailey:
Thinking about this from the `isolatedDeclarations` angle (which IIRC will require `isolatedModules`), this too would be a pattern we'd have to ban, I think, since it's the same problem but for the same symbol in type space?

andrewbranch:
I don’t think this applies to `isolatedDeclarations`, unless I’m missing something. The problematic question for single-file JS transpilers here is whether an import needs to be dropped. An `isolatedDeclarations` emitter need not ever drop an import.

jakebailey:
Ah, true; I guess I was thinking about it in terms of "can a d.ts emit reasonably understand what to export", but I guess they'd just leave things as-is.

I'm personally all for doing this break. It's only 2 out of the top 200 and the fix seems to be straightforward, right? We could rerun top500 or something to really see. That'd be fun...

@typescript-bot test top500

andrewbranch:
Yeah, the fix is just to make the imports type-only.

jakebailey:
Yeah, given the above run I _think_ this is okay.

jakebailey (approved):
I'm going to approve it, but maybe this warrants a design meeting quick topic?

PR #56356, opened 2023-11-09
Always update LS state through Project

sheetalkamat:
There have been multiple reports of issues with hitting assert at https://github.com/microsoft/TypeScript/blob/80ab111685e3e0c10d8d13cdd18d8dffa8328895/src/server/project.ts#L1518 because the program update before that didnt pass through Project's update graph and hence LS and Project go out of sync, Recent report at https://github.com/microsoft/TypeScript/pull/55716#issuecomment-1752163519

It has always been hard to reproduce this without plugin and race conditions etc. So for testing i wrote simple plugin test that fails without this fix.

### DIFF
--- a/src/testRunner/unittests/tsserver/plugins.ts
+++ b/src/testRunner/unittests/tsserver/plugins.ts
@@ -28,6 +28,8 @@ describe("unittests:: tsserver:: plugins:: loading", () => {
                     create(info: ts.server.PluginCreateInfo) {
                         info.session?.addProtocolHandler(testProtocolCommand, request => {
                             session.logger.log(`addProtocolHandler: ${jsonToReadableText(request)}`);
+                            // Assume this one needs program
+                            info.languageService.getProgram();
                             return {
                                 response: testProtocolCommandResponse,
                             };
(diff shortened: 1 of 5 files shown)


src/testRunner/unittests/tsserver/plugins.ts
```diff
@@ -101,6 +103,38 @@ describe("unittests:: tsserver:: plugins:: loading", () => {
...
+                        { name: pluginName },
+                    ],
+                },
+            }),
+        };
+
+        const { session, host } = createHostWithPlugin([aTs, tsconfig, libFile]);
+
+        openFilesForSession([aTs], session);
+        host.writeFile("/user/username/projects/project/b.ts", "const y = 10;"); // Missing file is writting so scheduling the update
```
DanielRosenwasser, line 126:
```suggestion
        // Write the missing file (referenced by 'a.ts') to schedule an update.
        host.writeFile("/user/username/projects/project/b.ts", "const y = 10;");
```

src/testRunner/unittests/tsserver/plugins.ts
```diff
@@ -101,6 +103,38 @@ describe("unittests:: tsserver:: plugins:: loading", () => {
...
+
+        session.executeCommandSeq({ // This should update LS with new program
```
DanielRosenwasser, line 128:
```suggestion
        // This should update the language service with a new program.
        session.executeCommandSeq({
```

jakebailey:
@typescript-bot pack this

@johnsoncodehk does this fix the crash you noted in https://github.com/microsoft/TypeScript/pull/55716#issuecomment-1752163519 ?

src/testRunner/unittests/tsserver/plugins.ts
```diff
@@ -101,6 +103,38 @@ describe("unittests:: tsserver:: plugins:: loading", () => {
...
+            command: testProtocolCommand,
+            arguments: testProtocolCommandRequest,
+        });
+
+        host.runQueuedTimeoutCallbacks(); // This results in program update
```
DanielRosenwasser, line 133:
```suggestion
        // This results in a program update.
        host.runQueuedTimeoutCallbacks();
```

tests/baselines/reference/tsserver/plugins/when-plugins-use-LS-to-get-program-and-update-is-pending.js
```diff
@@ -0,0 +1,318 @@
+currentDirectory:: / useCaseSensitiveFileNames: false
+Info seq  [hh:mm:ss:mss] Provided types map file "/typesMap.json" doesn't exist
```
DanielRosenwasser, line 2: Is this a problem?

sheetalkamat: No.. Its used only in some of the TI test cases and there that file is on the host

johnsoncodehk:
> @johnsoncodehk does this fix the crash you noted in [#55716 (comment)](https://github.com/microsoft/TypeScript/pull/55716#issuecomment-1752163519) ?

@jakebailey I can confirm this fix works, thanks!

jakebailey (approved):

PR #56357, opened 2023-11-09
Pr 50377 redo filter boolean overload

craigphicks:
It turns out a problem with this idea is that `BooleanConstructor` is not distinguishable from `(x:any)=>boolean`.  Meaning TypeScript doesn't check the interface prototype when the prototype is a TypeScript Symbol for a user variable. I.e., in this case `Boolean`.

Oddly, it does check the interface prototype when the prototype is a unique *ES Symbol*.

If that gets sorted out, it might be worth considering.

--------------------------------------

Fixes #50377 - when `Boolean` is used a predicate to Array.filter falsie types are filtered in the result.
(However, this version doesn't use an overload in `lib/es5.d.ts`)

## Source Files changed:

- `src/compiler/checker.ts`

## Added Test Files

- tests/baselines/reference/unionOfArraysBooleanFilterCall.js
- tests/cases/compiler/arrayFilterBooleanOverload.ts

## Related issues

filter(Boolean)
#50377 Explore adding the Boolean to filter to narrow out null/undefined in array#filter (closed pull)
#50387 Add BooleanConstructor as an overload to .filter to allow for easy type predicate filtering (inDiscussion)
#30621 narrow array type with .filter(Boolean) (marked as duplicate of #16655, although #16655 is more broad than #30621)
#56013 Adding an overload to .filter breaks specific inference with nested functions (related)
#55777

Boolean as falsey check in general
#16655 Boolean() cannot be used to perform a null check (broader than filter(Boolean))
#29955 Allow Boolean() to be used to perform a null check (accepted pull, may have reverted)

#53489 Add fallback logic for generating signatures for unions of array members

## Source code changes

- `checker.ts`
In `chooseOverload`, after `result` is selected, when the argument is of type `BooleanConstructor` and the result signature is `Array.filter`, the result signature is cloned, and its `resolvedReturnType` property is set.

## Regressions

None found with runtests

## Discussion

In this revision, es5.d.ts is not modified at all. Instead, the return type is changed dynamically in `chooseOverload` in checker.ts.
Therefore, the user won't see a completion suggesting Boolean, unlike the last revision. Obviously that is a disadvantage.

The advantage is that there are less changes overall, so the pull is easier to review. (Previously there were ~60 files changes, and that's a lot to review).

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -34120,6 +34120,11 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             result = chooseOverload(candidates, assignableRelation, isSingleNonGenericCandidate, signatureHelpTrailingComma);
         }
         if (result) {
+            const returnType = calculateSignatureReturnTypeForSpecialCases(result, args);
+            if (returnType) {
+                result = cloneSignature(result);
+                result.resolvedReturnType = returnType;
+            }
             return result;
         }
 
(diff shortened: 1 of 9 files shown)


fatcerberus:
Adding special magic for this in the compiler internals for what is otherwise a userland method call (that isn’t reflected in the method signature) feels like it opens a can of worms that the TS maintainers may not want to touch…

craigphicks:
> Adding special magic for this in the compiler internals for what is otherwise a userland method call (that isn’t reflected in the method signature) feels like it opens a can of worms that the TS maintainers may not want to touch…

I hadn't initially thought of the js standard lib as a userland method.  The compiler type processing internals do include tests for the js `Array` and `Readonly` array symbols, but not for the methods within those objects - so this would indeed be an outlier there.

craigphicks:
I believe you.

BTW - here is TypeScript compiler code where Array and Tuple methods are treated specially and rewritten:
```
            let memberName: __String;
            if (everyType(type, t => !!t.symbol?.parent && isArrayOrTupleSymbol(t.symbol.parent) && (!memberName ? (memberName = t.symbol.escapedName, true) : memberName === t.symbol.escapedName))) {
                // Transform the type from `(A[] | B[])["member"]` to `(A | B)[]["member"]` (since we pretend array is covariant anyway)
                const arrayArg = mapType(type, t => getMappedType((isReadonlyArraySymbol(t.symbol.parent) ? globalReadonlyArrayType : globalArrayType).typeParameters![0], (t as AnonymousType).mapper!));
                const arrayType = createArrayType(arrayArg, someType(type, t => isReadonlyArraySymbol(t.symbol.parent)));
                return (type as UnionType).arrayFallbackSignatures = getSignaturesOfType(getTypeOfPropertyOfType(arrayType, memberName!)!, kind);
            }
            (type as UnionType).arrayFallbackSignatures = result;
```
It is an example (and I think probably the only example) of the built-in .d.ts files being treated differently from user-authored ones.  The result of being forced into a corner, I guess.

jakebailey:
Yes, that was explicitly added in #53489.

@typescript-bot test top200
@typescript-bot run dt

@typescript-bot pack this

fatcerberus:
Emit time improved 10+% from *this*?  That *has* to be noise.

jakebailey:
It is, note the time granularity is only 0.01s and that test has enough problems that it doesn't emit and so runs in 0.08.

craigphicks:
It turns out a problem with this idea is that `BooleanConstructor` is not distinguishable from `(x:any)=>boolean`.  Meaning TypeScript doesn't check the interface prototype when the prototype is a TypeScript Symbol for a user variable. I.e., in this case `Boolean`.

Oddly, it does check the interface prototype when the prototype is a unique *ES Symbol*.

If that gets sorted out, it might be worth considering.

Issue #56352, opened 2023-11-09
Type narrowing not working with multiple adjacent case statements


johan13:
### 🔎 Search Terms

"switch", "type narrowing"

### 🕗 Version & Regression Information

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about narrowing and switch statements

### ⏯ Playground Link

https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABMOcAUBDATgcwFyIDOUWMYOiAPouACYCmwZ9tAlIgN4BQiviMwRGgCE2HOygALLHADuiMPXkBRLDKxpWAbi48+hWTCgRJQkiHrtufG4ggZC9RGIB0UOAFUADl-pYAwg70mogAvOGIAEQAgpEEAPTxiADyANJ6trZY9FAgWEgxkToZtvaOzrhunj5+gY4h4aFRAEJxiImIfuqIACoAygCMABwADAAsQwQA5GJT-ISIXnCEhDAARgA2AJ6IU3SMzLRTLiU2ZU6u7t6+AUENEZH+bR1dcFi9g6MT07Pzi8urTY7PZgBhMRRHE6ZXjnCo4KrXWp3diNKIAEWeSTSp2h2Vy+RaABpEP5EG9EBidDYAL66Gx4vIFABycCgiGixOaxNJ5MpXFpQA

### 💻 Code

```ts
function foo(arg: string | undefined) {
    if (!arg) throw new Error();

    switch (true) {
        case arg.toUpperCase() === "A": // OK
            return "A";

        case arg.toUpperCase() === "B": // error TS18048: 'arg' is possibly 'undefined'.
        case arg.toUpperCase() === "C": // error TS18048: 'arg' is possibly 'undefined'.
        case arg.toUpperCase() === "D": // OK
            return "B, C or D";
    }

    return "Not A, B, C or D";
}
```

### 🙁 Actual behavior

In two of the four case statements, the `arg` variable has the type `string | undefined` instead of `string`.

### 🙂 Expected behavior

I would expect the `arg` variable to have the same type in all four case statements.

RyanCavanaugh:
Not fixed in Nightly

cc @Andarist

Andarist:
I'll work on this, thanks for the ping.

Andarist:
[...]

PR #56358, opened 2023-11-09
Fixed narrowing in case clauses that follow other clauses that return

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/56352

### DIFF
--- a/src/compiler/binder.ts
+++ b/src/compiler/binder.ts
@@ -1689,14 +1689,12 @@ function createBinder(): (file: SourceFile, options: CompilerOptions) => void {
         const clauses = node.clauses;
         const isNarrowingSwitch = node.parent.expression.kind === SyntaxKind.TrueKeyword || isNarrowingExpression(node.parent.expression);
         let fallthroughFlow = unreachableFlow;
-        let afterPreviousClauseFlow: FlowNode | undefined;
 
         for (let i = 0; i < clauses.length; i++) {
             const clauseStart = i;
             while (!clauses[i].statements.length && i + 1 < clauses.length) {
-                if (afterPreviousClauseFlow && afterPreviousClauseFlow === unreachableFlow) {
+                if (fallthroughFlow === unreachableFlow) {
                     currentFlow = preSwitchCaseFlow!;
-                    afterPreviousClauseFlow = undefined;
                 }
                 bind(clauses[i]);
                 i++;
@@ -1707,7 +1705,7 @@ function createBinder(): (file: SourceFile, options: CompilerOptions) => void {
             currentFlow = finishFlowLabel(preCaseLabel);
             const clause = clauses[i];
             bind(clause);
-            fallthroughFlow = afterPreviousClauseFlow = currentFlow;
+            fallthroughFlow = currentFlow;
             if (!(currentFlow.flags & FlowFlags.Unreachable) && i !== clauses.length - 1 && options.noFallthroughCasesInSwitch) {
                 clause.fallthroughFlowNode = currentFlow;
             }
--- a/src/compiler/binder.ts
+++ b/src/compiler/binder.ts
@@ -1689,9 +1689,15 @@ function createBinder(): (file: SourceFile, options: CompilerOptions) => void {
         const clauses = node.clauses;
         const isNarrowingSwitch = node.parent.expression.kind === SyntaxKind.TrueKeyword || isNarrowingExpression(node.parent.expression);
         let fallthroughFlow = unreachableFlow;
+        let afterPreviousClauseFlow: FlowNode | undefined;
+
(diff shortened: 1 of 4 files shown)


jakebailey:
@typescript-bot test top200
@typescript-bot run dt

@typescript-bot pack this

src/compiler/binder.ts
```diff
@@ -1689,9 +1689,15 @@ function createBinder(): (file: SourceFile, options: CompilerOptions) => void {
         const clauses = node.clauses;
         const isNarrowingSwitch = node.parent.expression.kind === SyntaxKind.TrueKeyword || isNarrowingExpression(node.parent.expression);
         let fallthroughFlow = unreachableFlow;
+        let afterPreviousClauseFlow: FlowNode | undefined;
```
Andarist, line 1692: This can be removed with a patch like this
<details>
<summary>git diff</summary>

```diff
diff --git a/src/compiler/binder.ts b/src/compiler/binder.ts
index b3840cc157..b26af72dcc 100644
--- a/src/compiler/binder.ts
+++ b/src/compiler/binder.ts
@@ -1689,14 +1689,12 @@ function createBinder(): (file: SourceFile, options: CompilerOptions) => void {
         const clauses = node.clauses;
         const isNarrowingSwitch = node.parent.expression.kind === SyntaxKind.TrueKeyword || isNarrowingExpression(node.parent.expression);
         let fallthroughFlow = unreachableFlow;
-        let afterPreviousClauseFlow: FlowNode | undefined;
 
         for (let i = 0; i < clauses.length; i++) {
             const clauseStart = i;
             while (!clauses[i].statements.length && i + 1 < clauses.length) {
-                if (afterPreviousClauseFlow && afterPreviousClauseFlow === unreachableFlow) {
+                if (fallthroughFlow === unreachableFlow) {
                     currentFlow = preSwitchCaseFlow!;
-                    afterPreviousClauseFlow = undefined;
                 }
                 bind(clauses[i]);
                 i++;
@@ -1707,7 +1705,7 @@ function createBinder(): (file: SourceFile, options: CompilerOptions) => void {
             currentFlow = finishFlowLabel(preCaseLabel);
             const clause = clauses[i];
             bind(clause);
-            fallthroughFlow = afterPreviousClauseFlow = currentFlow;
+            fallthroughFlow = currentFlow;
             if (!(currentFlow.flags & FlowFlags.Unreachable) && i !== clauses.length - 1 && options.noFallthroughCasesInSwitch) {
                 clause.fallthroughFlowNode = currentFlow;
             }
```
</details>

but I'm not yet 100% sure if I can repeatedly "reset" the `currentFlow` to `preSwitchCaseFlow` for consecutive statementless clauses.

Andarist: I came to the conclusion that this patch should be fine. However, I found a new small bug:
```ts
function test(
  foo:
    | { kind: "a"; prop: string }
    | { kind: "b"; prop: number }
    | { kind: "c"; prop: boolean },
  bar?: {
    type: "c";
  },
) {
  if (!bar) {
    return;
  }

  switch (foo.kind) {
    case "a":
      return;
    case ((bar = undefined), "b"):
    case bar.type: // should error
      return;
  }
}
```

It's a separate issue since it doesn't work on `main` either.

jakebailey:
@typescript-bot test top200
@typescript-bot run dt

jakebailey (approved):
This _seems_ correct to me; if the fallthrough case doesn't exist, then the flow is just the pre-switch-case flow.

Maybe this is for another PR, but it also seems to me like this bit is where we should be doing the CFA between checks now, right?

Andarist:
> This seems correct to me; if the fallthrough case doesn't exist, then the flow is just the pre-switch-case flow.

A small clarification for other readers since I'm not sure if that will always be 100% clear. The fallthrough might not exist yet - since we might be iterating over the first case. Or the fallthrough from the previous case might have been computed to `unreachableFlow` in the most common cases like: `case 'prev': return; case 'next':`

> Maybe this is for another PR, but it also seems to me like this bit is where we should be doing the CFA between checks now, right?

Yeah, I'd prefer to leave this for another PR. In fact, I'm already tinkering with this locally ;p

jakebailey:
@typescript-bot pack this

jakebailey:
One edge case I'm thinking about but can't make happen is if we get an unreachable flow sometime after reassigning `fallthroughFlow`, which would then point the flow back up the top of the switch case.

jakebailey:
I suppose if that were true, then the only way to get to the `case` would be through the start flow, so maybe that's fine.

Andarist:
> I suppose if that were true, then the only way to get to the case would be through the start flow, so maybe that's fine.

That's roughly my understanding here as well.

PR #56373, opened 2023-11-12
Pr 56013

### LINKED ISSUE #56013: Adding an overload to .filter breaks specific inference with nested functions

craigphicks:
Fixes #56013

## Test files

`tests/cases/compiler/arrayFilterBooleanExternalOverload{1,2,3}.ts`

## Bugs found

1. In the course of looping through overloads in `function chooseOverload(...)`, inner signature results are cached in the global variable`nodeLinks`.  The loop iteration candidate set the cache, and subsequent loop iteration candidates read from the cache instead of calculating anew.

2. (*Possibly not a bug*) After adding some particular external overload to `Array.filter`, when the array type is an union of arrays, an error occurs, even though no error occurred when the overload was not added.  This error is not particular to `Array.filter` at all, and such behavior is the current status quo of TypeScript.  As such this might not be an official bug.

## Fixes

1. The quick and dirty fix for the given test cases is to set a flag on each iteration after the first, and read/reset that flag in `function checkExpression`.  

2.  Pull #53489 added fallback logic for generating signatures for unions of array members.  It turns out that the scope #53489 can  be fractionally extended to address the case of bug 2.

## Problems remaining

1. As currently written, fix 1 is incomplete because it won't work if multiple inner nodes have signatures calculated.  The test cases have only one.  *There is also the possibility of creating a more professional fix that could be applied to a wider range of uses: enabling the caching of temporary results in a top down manner*.

2.1.  Fix 2 could be separated out / removed from this pull.  Fix 2 sweep an immediate problem under the rug.  Unfortunately it does not extend to similar cases which are are not `Array` methods. A more general approach is required.

2.2  The display of the type of the `filter` in `arrayFilterBooleanExternalOverload{2,3}.ts` is incorrect - it shows the value without fix 2, whereas the actual values used for computation are those of fix 2.


src/compiler/checker.ts
```diff
@@ -34310,87 +34313,95 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
-                const candidate = candidates[0];
-                if (some(typeArguments) || !hasCorrectArity(node, args, candidate, signatureHelpTrailingComma)) {
-                    return undefined;
-                }
-                if (getSignatureApplicabilityError(node, args, candidate, relation, CheckMode.Normal, /*reportErrors*/ false, /*containingMessageChain*/ undefined)) {
-                    candidatesForArgumentError = [candidate];
-                    return undefined;
+            chooseOverloadRecursionLevel++; // #56013
+            chooseOverloadFlushNodesSignaturesReq[chooseOverloadRecursionLevel] = undefined;
+            const result = (() => {
```
craigphicks, line 34318: The main body of choose overload is wrapped in an Immediately Invoked Function Expression to implement the `chooseOverloadFlushNodesSignaturesReq` stack without error.

src/compiler/checker.ts
```diff
@@ -34310,87 +34313,95 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
-                return candidate;
-            }
 
-            for (let candidateIndex = 0; candidateIndex < candidates.length; candidateIndex++) {
-                const candidate = candidates[candidateIndex];
-                if (!hasCorrectTypeArgumentArity(candidate, typeArguments) || !hasCorrectArity(node, args, candidate, signatureHelpTrailingComma)) {
-                    continue;
-                }
+                for (let candidateIndex = 0; candidateIndex < candidates.length; candidateIndex++) {
+                    if (candidateIndex > 0) chooseOverloadFlushNodesSignaturesReq[chooseOverloadRecursionLevel] = new Set(); // #56013
```
craigphicks, line 34336: At the top of every loop after the first the request is set.

src/compiler/checker.ts
```diff
@@ -35292,7 +35303,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         checkGrammarTypeArguments(node, node.typeArguments);
 
         const signature = getResolvedSignature(node, /*candidatesOutArray*/ undefined, checkMode);
-        if (signature === resolvingSignature) {
+        if (signature === resolvingSignature || !signature /* #56013 */) {
```
craigphicks, line 35306: The `!signature` test was added to prevent the self-build from failing.

src/compiler/checker.ts
```diff
@@ -38932,6 +38944,16 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         const saveCurrentNode = currentNode;
         currentNode = node;
         instantiationCount = 0;
+        // #56013
+        if (node.kind === SyntaxKind.CallExpression && chooseOverloadRecursionLevel >= 0) {
+            let setOfNode: Set<Node> | undefined;
+            if (chooseOverloadRecursionLevel >= 0 && (setOfNode = chooseOverloadFlushNodesSignaturesReq[chooseOverloadRecursionLevel])) {
+                if (!setOfNode.has(node)) {
+                    getNodeLinks(node).resolvedSignature = undefined;
```
craigphicks, line 38952: The request is read and `resolvedSignature` is cleared, but only once per overload loop.

tests/baselines/reference/arrayFilterBooleanExternalOverload2.types
```diff
@@ -72,7 +72,7 @@ declare const maybe: boolean;
 >'bar' : "bar"
 >undefined : undefined
 >filter : { <S extends string | undefined>(predicate: (value: string | undefined, index: number, array: (string | undefined)[]) => value is S, thisArg?: any): S[]; (predicate: (value: string | undefined, index: number, array: (string | undefined)[]) => unknown, thisArg?: any): (string | undefined)[]; (predicate: BooleanConstructor & { prototype: unique symbol; }): string[]; }
->id() : (t: unknown) => boolean
+>id() : (t: string | undefined) => boolean
```
craigphicks, line 75: The original problem behavior specified in #56013 is fixed.

tests/baselines/reference/arrayFilterBooleanExternalOverload4.types
```diff
@@ -49,37 +49,37 @@ const t11 = (0 as any as F2<1,1>)(id(),id());
 >id : ID
```
craigphicks, line 50: The change in behavior is easiest to see in this test file.  The evaluation of `id()` got stuck in earlier overloads, resulting in incorrect overload choice behavior..

`id()` is used in multiple positions to shows that multiple nodes are being cleared for the same iteration.

src/compiler/checker.ts
```diff
@@ -35159,14 +35170,19 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             // since resolving such signature leads to resolving the potential outer signature, its arguments and thus the very same signature
             // it's possible that this inner resolution sets the resolvedSignature first.
             // In such a case we ignore the local result and reuse the correct one that was cached.
-            if (links.resolvedSignature !== resolvingSignature) {
+            // [cph 56013] This is buggy - links.resolvedSignature could be `undefined`. Unfortunately flow did not warn us.
+            // if (links.resolvedSignature !== resolvingSignature) {
+            //     result = links.resolvedSignature;
+            // }
+            if (links.resolvedSignature /* #56013 */ && links.resolvedSignature !== resolvingSignature) {
                 result = links.resolvedSignature;
```
craigphicks, line 35178: Fixed bug here - `links.resolvedSignature` is assumed to non-undefined, because flow assumes "no side effects".  But that only works if there are no side effects.  The bug that was fixed in this pull is based on eliminating a side effect between candidate iterations in `chooseOverload` - and that is done by resetting `getNodeLinks(node).resolvedSignature` to undefined.  

Perhaps the resetting in `checkExpression` could be changed to resetting to `resolvingSignature` instead of `undefined`?.   Then this extra check here would be unnecessary - but it would be worth adding a `Debug.assert(getNodeLinks(node).resolvedSignature)` immediately after the call to `resolveSignature()` to at least make clear that is a state invariant.  

However, is this logic:
```
if (links.resolvedSignature !== resolvingSignature) {
     result = links.resolvedSignature;
}
```
even necessary at all?  I will try removing it to test that question.

tests/baselines/reference/arrayFilterBooleanExternalOverload4.types
```diff
@@ -49,37 +49,37 @@ const t11 = (0 as any as F2<1,1>)(id(),id());
...
->t22 : never
->(0 as any as F2<2,2>)(id(),id()) : never
+>t22 : 99
+>(0 as any as F2<2,2>)(id(),id()) : 99
 >(0 as any as F2<2,2>) : F2<2, 2>
 >0 as any as F2<2,2> : F2<2, 2>
 >0 as any : any
 >0 : 0
->id() : (i: 1) => 1
+>id() : (i: 2) => 2
```
craigphicks, line 82: [Note that the incorrect `id()` value is not only cosmetic, it affects the final result.](https://github.com/microsoft/TypeScript/pull/56373/commits/2317daad176e8f99f2d41394a632434a0738330e#diff-8939d872f4bdf87cb549a3feb2fce124b57f681cd64b7536f2485b42654993b0)

weswigham:

src/compiler/checker.ts
```diff
@@ -38932,6 +38952,23 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+            if (chooseOverloadRecursionLevel >= 0 && (setOfNode = chooseOverloadFlushNodesSignaturesReq[chooseOverloadRecursionLevel])) {
+                if (!setOfNode.has(node)) {
+                    /**
+                     * #56013
+                     * The following is an invariant condition beneath the above call to resolveSignature:
+                     * - getNodeLinks(node).resolvedSignature is either the global constant readonly value `resolvingSignature`
+                     * - or a calculated signature.
+                     * Therefore resetting must set to `resolvingSignature` - it means "not yet calculated".
+                     */
+                    getNodeLinks(node).resolvedSignature = resolvingSignature;
```
weswigham, line 38967: I'm surprised the perf of this is OK, but it's likely because this solution is incomplete - this only partially purges the cached nodes; the _cached parameter symbol types_ on the parameter symbol links are still there! So while you'll get new signatures here, any parameters that get witnessed by multiple overload passes will still have their types locked down. Which isn't any worse than today, to be fair! It's a known issue with the signature resolution logic. IME, adding a full hierarchical cache to fully handle it and make speculative overload resolution totally unwindable was pretty bad for perf, but this approach, being partial, looks like it's OK. And if it still manages to fix a few usecases, that sounds pretty OK to me.

weswigham (changes requested):
While the extended suites run to make sure this doesn't break anything in a non-obvious way (it might, people tend to depend on odd corners of overload resolution), I'd like to request all the `#56013` refs in the comments get removed - issue refs in comments are for TODOs, not completed work (otherwise every line in the repo would have multiple issue refs). You can always check the blame to try down the "why is it like this" context later.

weswigham:
@typescript-bot test this
@typescript-bot run dt
@typescript-bot test top100

src/compiler/checker.ts
weswigham: Like, this feels odd, because it seems naive - it's basically resetting the resolved signature of *every* call expression we visit while looking at nested overload resolutions - in theory, only *some* calls should need to get reset - those which actually depend on the overload resolution in progress (which'd require a bit more tracing). I'd have figured over-invalidation like this'd manifest as terrible perf, but no - perf looks OK, broadly speaking, not more than a single percent diff.

src/compiler/checker.ts
weswigham, line 34318: Just a style thing - extract the IIFE into a named function in the scope, such as `chooseOverloadWorker` instead; makes debugging easier later for it to be named.

weswigham (changes requested):

weswigham:
Looks like this introduces some errors in `semantic-ui` on DT that are pretty pervasive. Those'll need some explanations (as expected breaking changes, with some details on how to mitigate it and accompanying changes to DT)/fixes (if it's just a bug in the PR) for this to pass muster.

craigphicks:
> Looks like this introduces some errors in `semantic-ui` on DT that are pretty pervasive. Those'll need some explanations (as expected breaking changes, with some details on how to mitigate it and accompanying changes to DT)/fixes (if it's just a bug in the PR) for this to pass muster.

I guess the only way to address these is to clone the repo and run the tests and debug - there doesn't seem anything in the test files that can easily be cut and pasted to create local tests.

There are a number of errors in `semantic-ui-sticky-tests.ts` that look this this:
```
ERROR: 38:13  expect  TypeScript@local expected type to be:
  JQuery<HTMLElement>
got:
  String
```
which definitely looks wrong.


The rest, a large number from a number of different files, seem to be of the form
```
ERROR: 37:9  expect  TypeScript@local compile error: 
  The last overload gave the following error.
     .....
```
*If* that means that an unnecessary error has been cleaned up, then this might be a worthwhile change after all.
But if it means that a truly wanted error is now missed, then it is bad.  We'll see ...

craigphicks:
> While the extended suites run to make sure this doesn't break anything in a non-obvious way (it might, people tend to depend on odd corners of overload resolution), I'd like to request all the `#56013` refs in the comments get removed - issue refs in comments are for TODOs, not completed work (otherwise every line in the repo would have multiple issue refs). You can always check the blame to track down the "why is it like this" context later.

Done

craigphicks:
@weswigham 
> Looks like this introduces some errors in `semantic-ui` on DT that are pretty pervasive. Those'll need some explanations (as expected breaking changes, with some details on how to mitigate it and accompanying changes to DT)/fixes (if it's just a bug in the PR) for this to pass muster.

It seems at least some of the `semantic-ui` error reports are pre-existing, by which I mean they can recreated with 5.3.2 and below, with modified test code.  I believe the opening this bug fix, which fixes improper caching that prevents recalculation of some generic call expressions, correctly allowed new evaluations that were previously improperly blocked - thus revealing the latent bug causing (at least some of) the `semantic-ui` reports.

I've separately posted the discovered bug on #56702, if you are interested.

PR #56384, opened 2023-11-13
Disallow whitespace before non-null assertion

jakebailey:
Prototype implementation just to run the test suite; this is wrong for comments and such and probably trailing trivia (if that is even possible).

A formatting test fails because we don't format code with syntax errors, which this of course introduces.


jakebailey:
@typescript-bot test top200
@typescript-bot run dt

tests/cases/compiler/parseNoWhitespaceBeforeExclamation.ts
```diff
@@ -0,0 +1,5 @@
+declare var value: any;
+
+value !;
+value !instanceof String;
+value/* this is a comment */! instanceof String;
```
DanielRosenwasser, line 5:
```suggestion
value !;

value !instanceof String;
value ! instanceof String
value/* this is a comment */!instanceof String;
value/* this is a comment */! instanceof String;

value !in String;
value ! in String
value/* this is a comment */!in String;
value/* this is a comment */! in String;
```

jakebailey: Sure though I'll note that none of the code here actually depends on what happens after 😄

tests/baselines/reference/parseNoWhitespaceBeforeExclamation.errors.txt
```diff
@@ -0,0 +1,42 @@
...
+!!! error TS1289: A non-null assertion may not be preceded by whitespace.
+    value ! in String
+         ~~
+!!! error TS1289: A non-null assertion may not be preceded by whitespace.
+    value/* this is a comment */!in String;
+         ~~~~~~~~~~~~~~~~~~~~~~~~
+!!! error TS1289: A non-null assertion may not be preceded by whitespace.
+    value/* this is a comment */! in String;
+         ~~~~~~~~~~~~~~~~~~~~~~~~
+!!! error TS1289: A non-null assertion may not be preceded by whitespace.
```
jakebailey, line 41: Obviously this is not the way I want to emit this error; we can discuss this later of course.

jakebailey:
Wow, really, nobody had this?

DanielRosenwasser:
> Wow, really, nobody had this?

In the top 200? Seems reasonable.

From the search @gorosgobe did over at https://github.com/tc39/proposal-negated-in-instanceof/issues/1#issuecomment-1799372434, there are only 13 results where `!in` or `!instanceof` occur.

tests/baselines/reference/parseNoWhitespaceBeforeExclamation.errors.txt
DanielRosenwasser: What's wrong with it? The inclusion of the `!` itself?

jakebailey: Complaining about a comment when there's no whitespace; I would have expected at least one tool to support some sort of interleaved comment a la `/*@__PURE__*/` or something. Maybe that's silly. If we do want to complain about comments, then my message would need to be updated (which is fine).

src/compiler/diagnosticMessages.json
```diff
@@ -951,6 +951,10 @@
         "category": "Error",
         "code": 1288
     },
+    "A non-null assertion may not be preceded by whitespace.": {
```
DanielRosenwasser, line 954: What about something like

> Whitespace and comments are not allowed between the argument and '!' operator in a non-null assertion.

jakebailey: Also shockingly, we don't use the term "whitespace" anywhere in our messages.

RyanCavanaugh: > Non-null assertion operator must immediately follow its expression

I like Daniel's too though

src/compiler/diagnosticMessages.json
```diff
@@ -1801,6 +1801,10 @@
         "category": "Error",
         "code": 1537
     },
+    "A non-null assertion must be preceded by an identifier.": {
```
DanielRosenwasser, line 1804: identifier? Isn't that misleading?

jakebailey: I guess `a[b]!`. bleh

towerofnix:
Hey, may I ask why you closed your own PR here? Just curious if there was any outside discussion or change of circumstance! (Here via https://github.com/tc39/proposal-negated-in-instanceof/issues/1#ref-pullrequest-1991581178)

jakebailey:
It didn't catch any bugs and wasn't the right implementation anyway. Just keeping my PR list clean.

PR #56387, opened 2023-11-14
Pull out parts of TI Adapter so we can test that more correctly instead of having to copy things

sheetalkamat:
This found another issue where in closing project which doesnt use TI, initialized typing installer just to do nothing

### DIFF
--- a/src/server/typingsCache.ts
+++ b/src/server/typingsCache.ts
@@ -157,7 +157,8 @@ export class TypingsCache {
     }
 
     onProjectClosed(project: Project) {
-        this.perProjectCache.delete(project.getProjectName());
-        this.installer.onProjectClosed(project);
+        if (this.perProjectCache.delete(project.getProjectName())) {
+            this.installer.onProjectClosed(project);
+        }
     }
 }
--- a/src/tsserver/nodeServer.ts
+++ b/src/tsserver/nodeServer.ts
@@ -1,8 +1,6 @@
 import * as protocol from "../server/protocol";
 import * as ts from "./_namespaces/ts";
 import {
-    ApplyCodeActionCommandResult,
-    assertType,
     CharacterCodes,
     combinePaths,
     createQueue,
@@ -12,7 +10,6 @@ import {
     FileWatcher,
     getDirectoryPath,
     getRootLength,
-    JsTyping,
     LanguageServiceMode,
     MapLike,
     noop,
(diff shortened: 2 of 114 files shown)


src/server/typingsCache.ts
```diff
@@ -157,7 +157,8 @@ export class TypingsCache {
     }
 
     onProjectClosed(project: Project) {
-        this.perProjectCache.delete(project.getProjectName());
-        this.installer.onProjectClosed(project);
+        if (this.perProjectCache.delete(project.getProjectName())) {
```
sheetalkamat, line 160: This is the fix for unnecessary queueing of closeProject request

src/tsserver/nodeServer.ts
```diff
@@ -634,155 +575,7 @@ function startNodeSession(options: StartSessionOptions, logger: Logger, cancella
...
-
-        onProjectClosed(p: Project): void {
-            this.send({ projectName: p.getProjectName(), kind: "closeProject" });
-        }
-
-        private send<T extends TypingInstallerRequestUnion>(rq: T): void {
-            this.installer.send(rq);
-        }
-
-        enqueueInstallTypingsRequest(project: Project, typeAcquisition: TypeAcquisition, unresolvedImports: SortedReadonlyArray<string>): void {
```
sheetalkamat, line 647: All of this moved to adapter

tests/baselines/reference/tsserver/autoImportProvider/Auto-importable-file-is-in-inferred-project-until-imported.js
```diff
@@ -89,12 +89,8 @@ TI:: [hh:mm:ss:mss] Got install request
       },
       "unresolvedImports": [],
       "projectRootPath": "/node_modules/@angular/forms",
-      "cachePath": "/a/data",
```
sheetalkamat, line 92: Removes unnecessary addition of cachePath that was done by test typing installer. (its same as globalcache path that will be processed at the initialization and was not set by node typings installer as well)

src/server/typingsCache.ts
sheetalkamat: Eg tests in unittests\tsserver\autoImportProvider would fail by showing creating TI when the project is closed that does not enable TI

jakebailey (approved):

PR #56395, opened 2023-11-14
ImportAttributes should go through the same emit phases when in an ImportTypeNode

### LINKED ISSUE #56394: Inferring return types with import attributes causes debug assertion.
### 🔎 Search Terms

infer return type import attributes

- This is a crash

### ⏯ Playground Link

[Playground Link](https://www.typescriptlang.org/dev/bug-workbench/?module=199&traceResolution=true#code/PTAEAEFsHsBMFcA2BTAXKAcnZHkA8AXAWAChSQIAzASxQDsBDSNUO7AfRgRQGdhq6dZACd+dWPgB0sSZAI9S+AA7RhBUAGNodHuuQ9IoALygCw+MgDcpcmHA16TFmwmc4SfWKGilDDQGsGAHNkSQArHm1SAG9QACJGZjj0OIFvOIAaeOVVeWTQWLjJYEgI-KKxCTxZMtAAX3qbEgp7WmRElkpoaEl5RTwVNVBgACoGEeBKeDoNAmptUYAjCdBKOgAKAEoC0lA90GFkAnhhOlBqSEGCddTBERKyzdI6oA)

### 💻 Code

```ts
// @module: NodeNext

// @filename: node_modules/inner/index.d.mts
export const esm = true;

// @filename: node_modules/inner/package.json
{ "name": "inner", "exports": { "./mjs": "./index.mjs" } }

// @filename: foo.ts
export /*a*/function/*b*/ fn() {
    return import("inner/mjs")
}
```

### 🙁 Actual behavior

Using infer return type on `fn` will cause a debug assertion. 

### 🙂 Expected behavior

Should infer the return type.

Found while working on isolated declaration.

It's caused by `ImportAttributes` not going through the full emit pipeline and thus not having their position assigned if they are part of an import type node.

dragomirtitian:
Fixes #56394

### DIFF
--- a/src/compiler/emitter.ts
+++ b/src/compiler/emitter.ts
@@ -223,6 +223,7 @@ import {
     isGeneratedIdentifier,
     isGeneratedPrivateIdentifier,
     isIdentifier,
+    isImportAttributes,
     isIncrementalCompilation,
     isInJsonFile,
     isInternalDeclaration,
@@ -1840,7 +1841,7 @@ export function createPrinter(printerOptions: PrinterOptions = {}, handlers: Pri
         if (hint === EmitHint.IdentifierName) return emitIdentifier(cast(node, isIdentifier));
         if (hint === EmitHint.JsxAttributeValue) return emitLiteral(cast(node, isStringLiteral), /*jsxAttributeEscape*/ true);
         if (hint === EmitHint.MappedTypeParameter) return emitMappedTypeParameter(cast(node, isTypeParameterDeclaration));
-        if (hint === EmitHint.TypeImportAttributes) return emitTypeImportAttributes(cast(node, ts.isImportAttributes));
+        if (hint === EmitHint.TypeImportAttributes) return emitTypeImportAttributes(cast(node, isImportAttributes));
         if (hint === EmitHint.EmbeddedStatement) {
             Debug.assertNode(node, isEmptyStatement);
             return emitEmptyStatement(/*isEmbeddedStatement*/ true);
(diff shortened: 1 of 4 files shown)


src/compiler/emitter.ts
```diff
@@ -4077,6 +4071,18 @@ export function createPrinter(printerOptions: PrinterOptions = {}, handlers: Pri
         writeTrailingSemicolon();
     }
 
+    function emitTypeImportAttributes(node: ImportAttributes) {
```
jakebailey, line 4074: I feel like this function name (and new enum value name) may be misleading; they don't have anything to do with type imports. Maybe this should be `emitImportTypeNodeAttributes`?

dragomirtitian: Renamed.

jakebailey (approved):

src/compiler/types.ts
```diff
@@ -8141,13 +8141,14 @@ export const enum ExternalEmitHelpers {
...
-    EmbeddedStatement,   // Emitting an embedded statement
-    JsxAttributeValue,   // Emitting a JSX attribute value
+    SourceFile,              // Emitting a SourceFile
+    Expression,              // Emitting an Expression
+    IdentifierName,          // Emitting an IdentifierName
+    MappedTypeParameter,     // Emitting a TypeParameterDeclaration inside of a MappedTypeNode
+    Unspecified,             // Emitting an otherwise unspecified node
+    EmbeddedStatement,       // Emitting an embedded statement
+    JsxAttributeValue,       // Emitting a JSX attribute value
+    ImportTypeNodeAttributes,// Emitting attributes as part of an ImportTypeNode
```
jakebailey, line 8151: Not sure how much this matters given we seem to be adding stuff here, I do wonder if new things should be marked as internal. @rbuckton probably knows better than me for this bit.

dragomirtitian:
@jakebailey This one would also ideally make it in before isolated declarations. It causes a failure in a test. Not critical but nice to have.

src/compiler/types.ts
rbuckton: I think it's fine for this to be public, as it is observable to custom transformers.

rbuckton (approved):

Issue #56266, opened 2023-10-30
Unexpected error diagnostics in Typescript 5.2.2


lramos15:
Working on https://github.com/microsoft/vscode-extension-telemetry updating all packages to latest (including typescript 5.2.2) appears to have caused unexpected red underlining telling me I don't have the node types when I have them installed. 

Attempting to install the missing types via the quick fix leads to a false expression error notification
![image](https://github.com/microsoft/vscode/assets/4544166/8bb41533-aab2-41ec-ab57-758d9498b3cf)

RyanCavanaugh:
What I did:
 * git clone the repo
 * npm install typescript@5.2
 * code .
 * Open src\node\telemetryReporter.ts
 * No errors
 * 🤷‍♂️

Repro steps please?

lramos15:
Remove `"applicationinsights": "^2.7.1"` from the package JSON and `npm install`. That seems to be what triggers it for me

RyanCavanaugh:
> but not an error regarding the built-in node packages.

Regarding the error -- you explicitly have type directive searching disabled due to `typeRoots: []` in the tsconfig. I'm guessing `applicationinsights` got it brought in in a roundabout way, but TS is doing what you told it to.
```ts
"typeRoots": []
```

RyanCavanaugh:
@sheetalkamat here's the error trace. `@types/node` is already in package.json under devdeps
```
Err 280   [15:47:21.799] Exception on executing command {
  "seq": 65,
  "type": "request",
  "command": "applyCodeActionCommand",
  "arguments": {
    "command": {
      "type": "install package",
      "file": "d:/Throwaway/vscode-extension-telemetry/src/node/telemetryReporter.ts",
      "packageName": "@types/node"
    }
  }
}:

    Debug Failure. False expression.

    Error: Debug Failure. False expression.
        at _NodeTypingsInstaller.installPackage (c:\Users\ryanca\AppData\Local\Programs\Microsoft VS Code\resources\app\extensions\node_modules\typescript\lib\tsserver.js:186191:13)
        at TypingsCache.installPackage (c:\Users\ryanca\AppData\Local\Programs\Microsoft VS Code\resources\app\extensions\node_modules\typescript\lib\tsserver.js:175906:27)
        at ConfiguredProject2.installPackage (c:\Users\ryanca\AppData\Local\Programs\Microsoft VS Code\resources\app\extensions\node_modules\typescript\lib\tsserver.js:176180:30)
[...]

PR #56400, opened 2023-11-14
Fix crash when there are multiple packages to install per code fix action

sheetalkamat:
Fixes https://github.com/microsoft/TypeScript/issues/56266

### DIFF
--- a/tests/baselines/reference/tsserver/codeFix/install-package.js
+++ b/tests/baselines/reference/tsserver/codeFix/install-package.js
@@ -1,0 +1,325 @@
+currentDirectory:: / useCaseSensitiveFileNames: false
+Info seq  [hh:mm:ss:mss] Provided types map file "/typesMap.json" doesn't exist
+Before request
+//// [/home/src/projects/project/src/file.ts]
+import * as os from "os";
+import * as https from "https";
+import * as vscode from "vscode";
+
+
+//// [/home/src/projects/project/tsconfig.json]
+{ }
+
+//// [/home/src/projects/project/node_modules/vscode/index.js]
+export const x = 10;
+
+//// [/a/lib/lib.d.ts]
+/// <reference no-default-lib="true"/>
+interface Boolean {}
+interface Function {}
+interface CallableFunction {}
+interface NewableFunction {}
+interface IArguments {}
+interface Number { toExponential: any; }
+interface Object {}
+interface RegExp {}
+interface String { charAt: any; }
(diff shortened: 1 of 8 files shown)


tests/baselines/reference/tsserver/codeFix/install-package.js
```diff
@@ -0,0 +1,462 @@
...
+          "ts2.7": "1.3.0"
+        }
+      }
+    }
+Info seq  [hh:mm:ss:mss] response:
+    {
+      "response": {
+        "changes": [],
+        "commands": [
+          {
```
sheetalkamat, line 256: Should these be deduplicated? If yes, at the end or during addition?
 (irrespective of that we do need the fix to TI adapter to handle the correct promise resolution)

jakebailey (approved):

tests/baselines/reference/tsserver/codeFix/install-package.js
jakebailey: Probably? I'm not familiar enough with this to know for sure...

sheetalkamat: May be @sandersn

andrewbranch (approved):

tests/baselines/reference/tsserver/codeFix/install-package.js
sandersn: I'm guessing that this will run two `npm installs`, at once probably, and should be deduped. It *seems* safer, and just as easy, to do it on addition, but I could be wrong. It looks like the key of the new Map might need to be based on the package name instead of a numeric id.

@zkat worked on this code much more recently than I have and might have additional thoughts.

sheetalkamat: its possible to have duplicate "npm install requests" if i clicked "fix all" command multiple times right? So  it has to be numeric id and installer handles the multiple installs for same package

sheetalkamat: The deduped result would be nice but wondering if it really matters

sandersn: It'd be equivalent to running two `npm install X` commands at once, right? If a manual test of that shows that npm handles that case on its own, I think it's fine to have ATA allow dupes.

sheetalkamat: Its does

PR #56411, opened 2023-11-15
feat(51870): Add a Quick Fix to add an additional parameter to a method or function

### LINKED ISSUE #51870: Add a Quick Fix to add an additional parameter to a method or function
See here:

image

```ts
openEditor(editor: EditorInput, didChangeActive: boolean): void {
	this.handleOpenedEditors(didChangeActive);
}
```

From: https://github.com/microsoft/vscode/blob/14f54b128dd5955db7a69e0aac760731d343e54b/src/vs/workbench/browser/parts/editor/tabsTitleControl.ts#L417

I would have expected a way to add `boolean` to the method.

a-tarasyuk:
Fixes #51870

### DIFF
--- a/src/services/codefixes/fixAddMissingParam.ts
+++ b/src/services/codefixes/fixAddMissingParam.ts
@@ -99,6 +99,7 @@ function getInfo(sourceFile: SourceFile, program: Program, pos: number): Signatu
         signature.declaration &&
         isConvertibleSignatureDeclaration(signature.declaration)
     ) {
+        // find a non-overload signature
         const declaration = (signature.declaration.body === undefined ? find(signature.declaration.symbol.declarations, d => isConvertibleSignatureDeclaration(d) && !!d.body) : signature.declaration) as ConvertibleSignatureDeclaration;
         if (declaration === undefined) {
             return undefined;
--- a/src/services/codefixes/fixAddMissingParam.ts
+++ b/src/services/codefixes/fixAddMissingParam.ts
@@ -1,0 +1,246 @@
+import {
+    append,
+    ArrowFunction,
+    declarationNameToString,
+    DiagnosticOrDiagnosticAndArguments,
+    Diagnostics,
+    factory,
+    find,
+    findAncestor,
+    first,
+    forEach,
+    FunctionDeclaration,
+    FunctionExpression,
+    FunctionLikeDeclaration,
+    getNameOfAccessExpression,
+    getNameOfDeclaration,
+    getTokenAtPosition,
+    isAccessExpression,
+    isCallExpression,
+    isIdentifier,
+    isParameter,
+    isPropertyDeclaration,
+    isSourceFileFromLibrary,
+    isVariableDeclaration,
+    last,
+    lastOrUndefined,
+    length,
+    map,
+    MethodDeclaration,
+    Node,
+    NodeBuilderFlags,
+    ParameterDeclaration,
+    Program,
+    SignatureKind,
+    sort,
+    SourceFile,
+    SyntaxKind,
+    textChanges,
+    Type,
+    TypeChecker,
+} from "../_namespaces/ts";
+import {
+    codeFixAll,
+    createCodeFixAction,
+    registerCodeFix,
+} from "../_namespaces/ts.codefix";
+
+const fixId = "addMissingParam";
+const errorCodes = [Diagnostics.Expected_0_arguments_but_got_1.code];
+
+registerCodeFix({
+    errorCodes,
+    fixIds: [fixId],
+    getCodeActions(context) {
+        const info = getInfo(context.sourceFile, context.program, context.span.start);
+        if (info === undefined) return undefined;
+
+        const changes = textChanges.ChangeTracker.with(context, t => doChange(t, context.sourceFile, info));
+        return [createCodeFixAction(fixId, changes, info.description, fixId, Diagnostics.Add_all_missing_parameters)];
+    },
+    getAllCodeActions: context =>
+        codeFixAll(context, errorCodes, (changes, diag) => {
+            const info = getInfo(context.sourceFile, context.program, diag.start);
+            if (info) {
+                doChange(changes, context.sourceFile, info);
+            }
+        }),
+});
+
+type ConvertableSignatureDeclaration =
+    | FunctionDeclaration
+    | FunctionExpression
+    | ArrowFunction
+    | MethodDeclaration;
+
+interface SignatureInfo {
+    readonly description: DiagnosticOrDiagnosticAndArguments;
+    readonly newParameters: ParameterInfo[];
+    readonly declarations: ConvertableSignatureDeclaration[];
+    readonly overload: ConvertableSignatureDeclaration | undefined;
+}
+
+interface ParameterInfo {
+    readonly pos: number;
+    readonly declaration: ParameterDeclaration;
+}
+
+function getInfo(sourceFile: SourceFile, program: Program, pos: number): SignatureInfo | undefined {
+    const token = getTokenAtPosition(sourceFile, pos);
+    const callExpression = findAncestor(token, isCallExpression);
+    if (callExpression === undefined || length(callExpression.arguments) === 0) return undefined;
+
+    const checker = program.getTypeChecker();
+    const type = checker.getTypeAtLocation(callExpression.expression);
+    const signatures = checker.getSignaturesOfType(type, SignatureKind.Call);
+    const signature = lastOrUndefined(sort(signatures, (a, b) => length(a.parameters) - length(b.parameters)));
+    if (
+        signature &&
+        signature.declaration &&
+        isConvertableSignatureDeclaration(signature.declaration)
+    ) {
+        const declaration = (signature.declaration.body === undefined ? find(signature.declaration.symbol.declarations, d => isConvertableSignatureDeclaration(d) && !!d.body) : signature.declaration) as ConvertableSignatureDeclaration;
+        if (declaration === undefined) {
+            return undefined;
+        }
+
+        isSourceFileFromLibrary(program, declaration.getSourceFile());
+
+        const overload = signature.declaration.body === undefined ? signature.declaration : undefined;
+        if (overload && length(overload.parameters) !== length(declaration.parameters)) return undefined;
+
+        const name = tryGetName(declaration);
+        if (name === undefined) return undefined;
+
+        const declarations: ConvertableSignatureDeclaration[] = append([declaration], overload);
+        const newParameters: ParameterInfo[] = [];
+        const parametersLength = length(declaration.parameters);
+        const argumentsLength = length(callExpression.arguments);
+        for (let i = 0, pos = 0, paramIndex = 0; i < argumentsLength; i++) {
+            const arg = callExpression.arguments[i];
+            const expr = isAccessExpression(arg) ? getNameOfAccessExpression(arg) : arg;
+            const type = checker.getWidenedType(checker.getBaseTypeOfLiteralType(checker.getTypeAtLocation(arg)));
+            const parameter = pos < parametersLength ? declaration.parameters[pos] : undefined;
+            if (
+                parameter &&
+                checker.isTypeAssignableTo(type, checker.getTypeAtLocation(parameter))
+            ) {
+                pos++;
+            }
+            else {
+                const newParameter = {
+                    pos: i,
+                    declaration: factory.createParameterDeclaration(
+                        /*modifiers*/ undefined,
+                        /*dotDotDotToken*/ undefined,
+                        expr && isIdentifier(expr) ? expr.text : `p${paramIndex++}`,
+                        /*questionToken*/ undefined,
+                        typeToTypeNode(checker, type, declaration),
+                        /*initializer*/ undefined,
+                    ),
+                };
+                append(newParameters, newParameter);
+            }
+        }
+        return {
+            declarations,
+            newParameters,
+            overload,
+            description: [
+                length(newParameters) > 1 ? Diagnostics.Add_missing_parameters_to_0 : Diagnostics.Add_missing_parameter_to_0,
+                declarationNameToString(name),
+            ],
+        };
+    }
+    return undefined;
+}
+
+function tryGetName(node: FunctionLikeDeclaration) {
+    const name = getNameOfDeclaration(node);
+    if (name) {
+        return name;
+    }
+
+    if (
+        isVariableDeclaration(node.parent) && isIdentifier(node.parent.name) ||
+        isPropertyDeclaration(node.parent) ||
+        isParameter(node.parent)
+    ) {
+        return node.parent.name;
+    }
+}
+
+function typeToTypeNode(checker: TypeChecker, type: Type, enclosingDeclaration: Node) {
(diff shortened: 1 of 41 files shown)


sandersn:
I'll bring this PR up at the vscode-TS weekly meeting to decide if we want to add this.

a-tarasyuk:
@sandersn Thanks

a-tarasyuk:
@sandersn Thanks for the feedback.

> This codefix sounds good. It replaces something that is tedious to do by hand. C# has this codefix. It would even be nice to have a codefix that deletes parameters when you pass too few.

Are you suggesting implementing a Quick Fix to address this case? I think `fixUnusedIdentifier` could be expanded to cover this new trigger case.

```ts
function f(a: number) {}
f(); // QF to remove a
```

> How does it handle overloads? Does it add to the longest matching one? (C# seems to do this)

Yes, the Quick Fix attempts to utilize the longest _overload declaration_.

> Does it break other callers? (C# does)

Yes, I think we can attempt to use `findAllRefs` and update other callers. Not sure if it's safe to do, though...

> Should the new parameter be optional? (C# gives codefixes for both; one is a new overload which is equivalent to optional)

It's a topic open for discussion. The safest route is to use optional parameters (it depends on the arg position), however, I think if a user calls the function with a new argument, there's an expectation that they want this parameter to be explicitly declared in the signature.

> Is the new parameter always added to the end? Can the new parameter be added in the middle based on which types match?

Yes.

> Does it offer the codefix for 'external' code -- basically, code from node_modules. (No idea how other codefixes handle this.)

Need to check, I think there's a utility to handle that.

> There is a codefix to remove unused parameters from a function declaration. Since it's from the declaration side it shouldn't interfere. But it may edit callers to remove the parameters (Need to test this -- unused parameters should never have callers in order to be 'unused', right?)

Could you clarify this question?

sandersn:
Here's an example for the last question:

```ts
function f(x: number, y: number, unused?: number) {
  return x + y
}
f(1, 2)
f(1, 2, 3)
```

`unused` is marked as unused whether or not `f(1, 2, 3)` is in the file. However, I just tested the quick fix, and it's smart enough not to offer to remove `unused` when `f(1, 2, 3)` is present. Only the "prefix `unused` with _" quick fix is offered. When `f(1, 2, 3)` is not in the file, a second quick  fix to remove `unused` is offered.

In summary, there's nothing to do here; the existing quick fix is smart enough already.

sandersn:
Most of the other questions from the design meeting don't require changing anything, as far as I can see. The only 2 open ones are
1. whether to offer two code fixes (optional/non-optional) -- I slightly prefer two since I think people will sometimes want non-optional to force themselves to update all callers.
2. how to avoid offering the quick fix when the callee comes from node_modules.

tests/cases/fourslash/codeFixAddMissingParam13.ts
```diff
@@ -0,0 +1,20 @@
...
+////}
+////f("", 1, "");
+
+verify.codeFix({
+    description: [ts.Diagnostics.Add_missing_parameter_to_0.message, "f"],
+    index: 0,
+    newFileContent:
+`function f(a: string): string;
+function f(a: string, b: number, p0: string): string;
+function f(a: string, b?: number, p0?: string): string {
```
sandersn, line 16: Looks like the rule is: Adds `p0` to both of the longest overloads, matching optionality.

tests/cases/fourslash/codeFixAddMissingParam14.ts
```diff
@@ -0,0 +1,20 @@
...
+////}
+////f("", "", 1);
+
+verify.codeFix({
+    description: [ts.Diagnostics.Add_missing_parameter_to_0.message, "f"],
+    index: 0,
+    newFileContent:
+`function f(a: string): string;
+function f(a: string, p0: string, b: number): string;
+function f(a: string, p0: string, b?: number): string {
```
sandersn, line 16: Same here, except that it's also able to match in the middle.

src/services/codefixes/fixAddMissingParam.ts
```diff
@@ -0,0 +1,248 @@
...
+                }
+                else {
+                    changes.insertNodeAtEndOfList(sourceFile, declaration.parameters, parameter);
+                }
+            });
+        }
+    });
+}
+
+function isConvertableSignatureDeclaration(node: Node): node is ConvertableSignatureDeclaration {
```
sandersn, line 209:
```suggestion
function isConvertibleSignatureDeclaration(node: Node): node is ConvertibleSignatureDeclaration {
```

src/services/codefixes/fixAddMissingParam.ts
```diff
@@ -0,0 +1,248 @@
...
+    getAllCodeActions: context =>
+        codeFixAll(context, errorCodes, (changes, diag) => {
+            const info = getInfo(context.sourceFile, context.program, diag.start);
+            if (info) {
+                doChange(changes, context.sourceFile, info);
+            }
+        }),
+});
+
+type ConvertableSignatureDeclaration =
```
sandersn, line 70:
```suggestion
type ConvertibleSignatureDeclaration =
```

src/services/codefixes/fixAddMissingParam.ts
```diff
@@ -0,0 +1,248 @@
...
+        signature &&
+        signature.declaration &&
+        isConvertableSignatureDeclaration(signature.declaration)
+    ) {
+        const declaration = (signature.declaration.body === undefined ? find(signature.declaration.symbol.declarations, d => isConvertableSignatureDeclaration(d) && !!d.body) : signature.declaration) as ConvertableSignatureDeclaration;
+        if (declaration === undefined) {
+            return undefined;
+        }
+
+        isSourceFileFromLibrary(program, declaration.getSourceFile());
```
sandersn, line 107: looks like there should be an early return based on the value of this expression (and some associated tests)

src/services/codefixes/fixAddMissingParam.ts
```diff
@@ -0,0 +1,248 @@
...
+            description: [
+                length(newParameters) > 1 ? Diagnostics.Add_missing_parameters_to_0 : Diagnostics.Add_missing_parameter_to_0,
+                declarationNameToString(name),
+            ],
+        };
+    }
+    return undefined;
+}
+
+function tryGetName(node: FunctionLikeDeclaration) {
```
sandersn, line 160: I vaguely remember something like this already existing as a utility. Something like getAssignedName maybe?

src/services/codefixes/fixAddMissingParam.ts
```diff
@@ -0,0 +1,248 @@
...
+        isPropertyDeclaration(node.parent) ||
+        isParameter(node.parent)
+    ) {
+        return node.parent.name;
+    }
+}
+
+function typeToTypeNode(checker: TypeChecker, type: Type, enclosingDeclaration: Node) {
+    const typeNode = checker.typeToTypeNode(checker.getWidenedType(type), enclosingDeclaration, NodeBuilderFlags.NoTruncation);
+    return typeNode ? typeNode : factory.createKeywordTypeNode(SyntaxKind.UnknownKeyword);
```
sandersn, line 177:
```suggestion
    return checker.typeToTypeNode(checker.getWidenedType(type), enclosingDeclaration, NodeBuilderFlags.NoTruncation)
        ?? factory.createKeywordTypeNode(SyntaxKind.UnknownKeyword);
```

src/services/codefixes/fixAddMissingParam.ts
```diff
@@ -0,0 +1,248 @@
...
+    const checker = program.getTypeChecker();
+    const type = checker.getTypeAtLocation(callExpression.expression);
+    const signatures = checker.getSignaturesOfType(type, SignatureKind.Call);
+    const signature = lastOrUndefined(sort(signatures, (a, b) => length(a.parameters) - length(b.parameters)));
+    if (
+        signature &&
+        signature.declaration &&
+        isConvertableSignatureDeclaration(signature.declaration)
+    ) {
+        const declaration = (signature.declaration.body === undefined ? find(signature.declaration.symbol.declarations, d => isConvertableSignatureDeclaration(d) && !!d.body) : signature.declaration) as ConvertableSignatureDeclaration;
```
sandersn, line 102: I'd suggest a comment in front like  `// find a non-overload signature`, or just rename `declaration` to `nonOverloadDeclaration`

src/services/codefixes/fixAddMissingParam.ts
```diff
@@ -0,0 +1,248 @@
...
+
+        isSourceFileFromLibrary(program, declaration.getSourceFile());
+
+        const overload = signature.declaration.body === undefined ? signature.declaration : undefined;
+        if (overload && length(overload.parameters) !== length(declaration.parameters)) return undefined;
+
+        const name = tryGetName(declaration);
+        if (name === undefined) return undefined;
+
+        const declarations: ConvertableSignatureDeclaration[] = append([declaration], overload);
```
sandersn, line 115: how is `append([declaration], overload)` different from `[declaration, overload]`? Does it omit `overload` when it's undefined?

tests/cases/fourslash/codeFixAddMissingParam13.ts
sandersn: Reading the code: no, it's the longest overload and the implementation signature, so long as they both have the same length. What about when there are two overloads with the same length as the implementation signature? Should they both get the new parameter? It seems little odd to me, but it seems odder to choose between them, so maybe 'both' is the right answer.

src/services/codefixes/fixAddMissingParam.ts
```diff
@@ -0,0 +1,248 @@
...
+        const newParameters: ParameterInfo[] = [];
+        const parametersLength = length(declaration.parameters);
+        const argumentsLength = length(callExpression.arguments);
+        if (parametersLength > argumentsLength) return undefined;
+
+        for (let i = 0, pos = 0, paramIndex = 0; i < argumentsLength; i++) {
```
sandersn, line 121: it's a little unfortunate that this code is, essentially, re-implementing the parameter-argument assignability checks in resolveCall. But it's probably the right approach, since exposing resolveCall for re-use isn't feasible.

sandersn (changes requested):

src/services/codefixes/fixAddMissingParam.ts
a-tarasyuk: `getAssignedName` doesn't handle `parameters` and `property declarations`.

src/services/codefixes/fixAddMissingParam.ts
a-tarasyuk: Yes. `append` allows to omit +1 conditions

tests/cases/fourslash/codeFixAddMissingParam13.ts
a-tarasyuk: The Quick Fix attempts to find and modify the longest overload and signature declaration. I thought it might seem odd to only change the overload. Would it be acceptable to modify the overload?

a-tarasyuk:
@sandersn

> whether to offer two code fixes (optional/non-optional) -- I slightly prefer two since I think people will sometimes want non-optional to force themselves to update all callers.

Should it be provided for both _overload+signature declarations_ and _signature declarations_?
```ts

function f(a: string): string;
function f(a: string, b: number): string;
function f(a: string, b?: number): string {
    return "";
}
f("", 1, "");

1. Add missing parameter to "f"
2. Add optional parameter to "f"
```

```ts
function f() {}
f("");
1. Add missing parameter to "f"
2. Add optional parameter to "f"
```

> how to avoid offering the quick fix when the callee comes from node_modules.

I came across this utility in another Quick Fix. Need to check if it works correctly, especially in the _fourslash_ context.

https://github.com/microsoft/TypeScript/blob/2c14a1c2250d0b891f743c1b6ae58ac951942206/src/services/codefixes/fixAddMissingMember.ts#L366-L368

tests/cases/fourslash/codeFixAddMissingParam13.ts
sandersn: What I meant was, what about the case when there are two overloads with the same length?

```ts
function f(a: string, b: string): string
function f(a: number, b: number): number
function f(a: number | string, b: number | string): number | string {
  return a + b
}
f(1, 2, "commentary: adding 1 and 2")
```

should add a new parameter to all 3 signatures.

(Aside: this would be a great quick fix to follow up with AI-powered rename, since it should obviously be named `comment` or `message` but Typescript has no way of calculating that.)

sandersn:
- I don't think we need to distinguish between adding a parameter to overload signatures vs implementation signatures: to work correctly, you have to add the new parameter to the implementation signature and at least one overload signature.

- I like the wording "Add missing parameter" vs "Add optional parameter".

tests/cases/fourslash/codeFixAddMissingParam13.ts
a-tarasyuk: Oke, I've added the [test](https://github.com/microsoft/TypeScript/pull/56411/files#diff-370f33a81a61409d162a7737ad0cbaf02ab3a7ef6c16746b5eaa8fc76452863cR1-R20) to cover this case

a-tarasyuk:
> I like the wording "Add missing parameter" vs "Add optional parameter".

@sandersn I've added `Add optional parameter` QF

a-tarasyuk:
> I like the wording "Add missing parameter" vs "Add optional parameter".

@sandersn I've added `Add optional parameter QF`

sandersn (changes requested):
I like the design. Just comments and suggestions on the code.

src/services/codefixes/fixAddMissingParam.ts
```diff
@@ -88,76 +127,80 @@ interface ParameterInfo {
...
-            return undefined;
-        }
 
-        if (isSourceFileFromLibrary(program, declaration.getSourceFile())) {
-            return undefined;
+    // find a non-overload signature
+    const declaration = firstDefined(signatures, s => {
+        if (s.declaration && isConvertibleSignatureDeclaration(s.declaration)) {
+            return s.declaration.body ? s.declaration :
+                find(s.declaration.symbol.declarations, d => isConvertibleSignatureDeclaration(d) && !!d.body) as ConvertibleSignatureDeclaration;
```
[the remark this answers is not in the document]
a-tarasyuk: The result of `checker.getSignaturesOfType(type, SignatureKind.Call);` does not include the implementation of overloaded declarations. I've opted to use a symbol directly from the type to retrieve all declarations, bypassing the need for `getSignaturesOfType`. Does that make sense?

src/services/codefixes/fixAddMissingParam.ts
```diff
@@ -88,76 +127,80 @@ interface ParameterInfo {
...
-    ) {
-        // find a non-overload signature
-        const declaration = (signature.declaration.body === undefined ? find(signature.declaration.symbol.declarations, d => isConvertibleSignatureDeclaration(d) && !!d.body) : signature.declaration) as ConvertibleSignatureDeclaration;
-        if (declaration === undefined) {
-            return undefined;
-        }
 
-        if (isSourceFileFromLibrary(program, declaration.getSourceFile())) {
-            return undefined;
+    // find a non-overload signature
```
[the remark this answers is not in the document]
a-tarasyuk: I've changed to use the last declaration.

sandersn (approved):

PR #56429, opened 2023-11-16
Consistently return `errorType` when detecting circularities

Andarist:
a quick experiment based on discussion/observations we had here: https://github.com/microsoft/TypeScript/pull/56258/files#r1393385776

### DIFF
--- a/tests/baselines/reference/assignmentCompatWithObjectMembers.types
+++ b/tests/baselines/reference/assignmentCompatWithObjectMembers.types
@@ -189,16 +189,16 @@ module ObjectTypes {
 >b : { foo: any; }
 
     var a2 = { foo: a2 };
->a2 : any
+>a2 : error
 >{ foo: a2 } : { foo: any; }
->foo : any
->a2 : any
+>foo : error
+>a2 : error
 
     var b2 = { foo: b2 };
->b2 : any
+>b2 : error
 >{ foo: b2 } : { foo: any; }
->foo : any
->b2 : any
+>foo : error
+>b2 : error
 
     s = t;
 >s = t : T
@@ -216,9 +216,9 @@ module ObjectTypes {
 >s2 : S2
 
     s = a2;
->s = a2 : any
+>s = a2 : error
 >s : S
->a2 : any
+>a2 : error
 
     s2 = t2;
 >s2 = t2 : T2
@@ -241,9 +241,9 @@ module ObjectTypes {
 >b : { foo: any; }
 
     s2 = a2;
->s2 = a2 : any
+>s2 = a2 : error
 >s2 : S2
->a2 : any
+>a2 : error
 
     a = b;
 >a = b : { foo: any; }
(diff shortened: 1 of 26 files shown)


tests/baselines/reference/witness.errors.txt
```diff
@@ -2,24 +2,17 @@ witness.ts(4,21): error TS2372: Parameter 'pInit' cannot reference itself.
 witness.ts(8,14): error TS2729: Property 'x' is used before its initialization.
 witness.ts(20,5): error TS2403: Subsequent variable declarations must have the same type.  Variable 'a' must be of type 'any', but here has type 'never'.
 witness.ts(28,12): error TS2695: Left side of comma operator is unused and has no side effects.
-witness.ts(29,5): error TS2403: Subsequent variable declarations must have the same type.  Variable 'co1' must be of type 'any', but here has type 'number'.
```
Andarist, line 5: it seems that this is even a small improvement since the intention of the caller was not to raise those errors when the type is the `errorType`, see: https://github.com/microsoft/TypeScript/blob/32b618c2d8f42948e7d6fa2e9b264079e30b5349/src/compiler/checker.ts#L42290-L42296

jakebailey:
Honestly I like this better

@typescript-bot test top200
@typescript-bot run dt

@typescript-bot pack this

Andarist:
> Honestly I like this better

Than the targeted `if` in https://github.com/microsoft/TypeScript/pull/56258 ? Or in general? ;p

jakebailey:
> > Honestly I like this better
> 
> Than the targeted `if` in #56258 ? Or in general? ;p

Both!

tests/baselines/reference/assignmentCompatWithObjectMembers.types
```diff
@@ -189,16 +189,16 @@ module ObjectTypes {
 >b : { foo: any; }
 
     var a2 = { foo: a2 };
->a2 : any
+>a2 : error
```
jakebailey, line 192: I am somewhat suspicious of this; this file does not appear to have an error baseline, yet we are seeing the error type in the type baselines. Is there a missing error, or do we actually need to be returning any sometimes?

jakebailey (changes requested):

tests/baselines/reference/assignmentCompatWithObjectMembers.types
Andarist: It's about this case:

```ts
// @noImplicitAny: false
module ObjectTypes {
    var a2 = { foo: a2 };
}
```

I'm not sure what should be done about this. `errorType` does behave like `any`... but I'd be worried about some accidental propagation of it since it could be infectious for consumers. OTOH, `any` is already infectious.

Returning a different value based on `noImplicitAny` would kinda defeat the goal of this PR. Unless we'd also return the same value (based on `noImplicitAny`) right away when the circularity is detected for the first time (currently `errorType` is returned from there and it's often "replaced" by `anyType` when the stack unwinds).

tests/baselines/reference/witness.errors.txt
jakebailey: Swinging back here, it seems sorta odd just from the PoV that a self referencing thing will just not have any errors. It seems like this test is mainly to "witness" the type of these expressions in errors, but maybe it's fine. Odd that the `.types` file does not show a change?

jakebailey: Will have to do some forensics but I'm not sure why we even want to ignore these anyway...

tests/baselines/reference/assignmentCompatWithObjectMembers.types
jakebailey: Looking again, I see a number of tests which have `: error` in their baselines but have no errors, so this is probably fine honestly.

Issue #54648, opened 2023-06-14
5.1.3 new "feature"? Branded types are preserved in `const` template literal


conorbrandon:
# Bug Report

### 🔎 Search Terms

branded, interpolation, 5.1.3

### 🕗 Version & Regression Information

This changed between versions 5.0.4 and 5.1.3

### ⏯ Playground Link

[Playground link with relevant code](https://www.typescriptlang.org/play?ts=5.1.3#code/C4TwDgpgBAylC8UAGBLAJgfQCQG8DOwATigHYDmUAZFDhgEaECGJaAXFAEQCSAIhwL78kAbgBQAenFRpAPQD8otBADGAG0aFoygPYkCUdOwLFyVGvSYt23PvzE69wKHgTJ02HOiFRGLhwTFJaVk5IA)

### 💻 Code

```ts
type S = `id_${string & {_brand: "ID"}}`;
//   ^? type S = `id_${string & { _brand: "ID"; }}`
declare const id: string & {_brand: "ID"};
const s = `id_${id}` as const;
//    ^? const s: `id_${string & { _brand: "ID"; }}`
```

### 🙁 Actual behavior

The intersected brand is preserved during string interpolation for variables and types!

### 🙂 Expected behavior

Pre 5.1.3, the intersected brand always got removed during string interpolation for variables and types.

This is part bug report and question. My question being in future versions, will this behavior be preserved? I wrote a clunky generic function and type to do this exact thing because I found it valuable, but I'd much rather rely on native behavior if this is going to stick around 👀

RyanCavanaugh:
I think this was (unintentional?) fallout from #48034 / #52345

jakebailey:
I think this was intentional? (I at least explicitly added it and tests for it, though in the course of fixing other stuff.)

Also @weswigham

jakebailey:
Though, I'm not sure how this should interact with as const... I guess it's a noop and that's normal?

jakebailey:
The relevant PR is https://github.com/microsoft/TypeScript/pull/52836, note that the tests contain effectively exactly what the original post contains, a branded type inside a template.

conorbrandon:
Ah, template literals, that's the word I was looking for (instead of string interpolation). Had a brain fart, my apologies.

ahejlsberg:
I'm reopening this one. The original issue, #48034, was that we're not "untagging" literal types when they occur in template literal placeholders. In other words, the expectation was that

```ts
[...]

PR #56434, opened 2023-11-16
Fix support for intersections in template literal placeholder types

### LINKED ISSUE #54177: Template Literal Types reduces string union
# Bug Report

When the template literal type is used in a union with strings that match its pattern, the union is reduced:

```ts
type union = 'downcast' | 'dataDowncast' | 'editingDowncast' | `${ string }Downcast`;
//   ^? type union = `${string}Downcast` | "downcast"
```

This is a problem when I want to use it as a parameter type, because then only "downcast" is suggested when calling the function:

```ts
function test( groupName: union ): void {}

test('') // Only "downcast" is suggested
```

As a workaround, I can use function overloading, where one signature has union with all string literals and another only the template literal type, but it would be nice to get this issue fixed:

```ts
function test2( groupName: 'downcast' | 'dataDowncast' | 'editingDowncast' ): void;
function test2( groupName: `${ string }Downcast` ): void;
function test2( groupName: string ): void {}

test2('') // All "downcast", "dataDowncast" and "editingDowncast" are suggested
```

From 4.1 (when template literal types were introduced) to 5.1.0-beta, tested in TypeScript Playground.

### ⏯ Playground Link

[Link to the TypeScript Playground](https://tsplay.dev/wXrnLm)

### 🙁 Actual behavior

Only string literals that don't match the template literal type are displayed.

### 🙂 Expected behavior

All string literals should be shown.

ahejlsberg:
Implements the changes discussed [here](https://github.com/microsoft/TypeScript/issues/54648#issuecomment-1814874112) and fixes #54177 according to our design meeting discussion.

Fixes #54177.

### DIFF
--- a/tests/cases/conformance/types/literal/templateLiteralTypesPatterns.ts
+++ b/tests/cases/conformance/types/literal/templateLiteralTypesPatterns.ts
@@ -202,9 +202,9 @@ export abstract class BB {
 }
 
 // repro from https://github.com/microsoft/TypeScript/issues/54177#issuecomment-1538436654
-function conversionTest(groupName: | "downcast" | "dataDowncast" | "editingDowncast" | `${string & {}}Downcast`) {}
+function conversionTest(groupName: | "downcast" | "dataDowncast" | "editingDowncast" | `${string}Downcast` & {}) {}
 conversionTest("testDowncast");
-function conversionTest2(groupName: | "downcast" | "dataDowncast" | "editingDowncast" | `${{} & string}Downcast`) {}
+function conversionTest2(groupName: | "downcast" | "dataDowncast" | "editingDowncast" | {} & `${string}Downcast`) {}
 conversionTest2("testDowncast");
 
 function foo(str: `${`a${string}` & `${string}a`}Test`) {}
--- a/tests/cases/conformance/types/literal/templateLiteralTypesPatterns.ts
+++ b/tests/cases/conformance/types/literal/templateLiteralTypesPatterns.ts
@@ -206,6 +206,10 @@ function conversionTest(groupName: | "downcast" | "dataDowncast" | "editingDownc
 conversionTest("testDowncast");
 function conversionTest2(groupName: | "downcast" | "dataDowncast" | "editingDowncast" | {} & `${string}Downcast`) {}
 conversionTest2("testDowncast");
+function conversionTest3(groupName: | "downcast" | "dataDowncast" | "editingDowncast" | `${string & {}}Downcast`) {}
+conversionTest3("testDowncast");
+function conversionTest4(groupName: | "downcast" | "dataDowncast" | "editingDowncast" | `${{} & string}Downcast`) {}
+conversionTest4("testDowncast");
 
 function foo(str: `${`a${string}` & `${string}a`}Test`) {}
 foo("abaTest"); // ok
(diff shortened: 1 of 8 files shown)


ahejlsberg:
@typescript-bot test top100
@typescript-bot run dt

tests/cases/conformance/types/literal/templateLiteralTypesPatterns.ts
```diff
@@ -202,9 +202,9 @@ export abstract class BB {
 }
 
 // repro from https://github.com/microsoft/TypeScript/issues/54177#issuecomment-1538436654
-function conversionTest(groupName: | "downcast" | "dataDowncast" | "editingDowncast" | `${string & {}}Downcast`) {}
+function conversionTest(groupName: | "downcast" | "dataDowncast" | "editingDowncast" | `${string}Downcast` & {}) {}
 conversionTest("testDowncast");
-function conversionTest2(groupName: | "downcast" | "dataDowncast" | "editingDowncast" | `${{} & string}Downcast`) {}
+function conversionTest2(groupName: | "downcast" | "dataDowncast" | "editingDowncast" | {} & `${string}Downcast`) {}
 conversionTest2("testDowncast");
```
Andarist, line 208: I think that those previous test cases are still somewhat valuable. Even if `string & {}` within the placeholder is not meant to enhance open-ended completions those cases are still meant to be type-checked correctly. In TS 5.1 they errored incorrectly: [TS playground](https://www.typescriptlang.org/play?ts=5.1.6#code/GYVwdgxgLglg9mABBBA3ApgJwM7zAFXWygGYAKAc0zhAAcA5AQwFt0AuRAH0QCIATOAHdIjYjy68+jKIwAiQkWIk90fGLDAV5wiKKjjuAAwAkAb2KYYmxADJEpgL4PtiqIYCU9hwCgUYDDh4hMTkPFBEUC66Yu4A3N6gkLAIyGhYuAjBUAAslNR0TKwc3PwK0frKUjJRega8qupWWmW1EiamjraIFk3OLcQeXr5pgZkRuWERNTGxQA)

```suggestion
conversionTest2("testDowncast");
function conversionTest3(groupName: | "downcast" | "dataDowncast" | "editingDowncast" | `${string & {}}Downcast`) {}
conversionTest3("testDowncast");
function conversionTest4(groupName: | "downcast" | "dataDowncast" | "editingDowncast" | `${{} & string}Downcast`) {}
conversionTest4("testDowncast");
```

ahejlsberg: Ok, I will add them.

jakebailey (approved):

DanielRosenwasser:
@typescript-bot pack this

ahejlsberg:
@jakebailey @DanielRosenwasser So, there's this beauty:

```ts
type Goo<T> = {
  [key: `abc${"def" & { tag: T }}`]: string;
}

type T0 = Goo<void>;  // Index signature key type not instantiated so type parameter leaks
```

By design we don't instantiate index signature key types (that's what mapped types are for), but that in turn means we really can't permit anything generic in tag types. But we have no guaranteed way of detecting whether an object type contains something generic.

Starting to seem like we just want to strip tags in template literal placeholders.

ahejlsberg:
Hmm, the leaking type parameter issue isn't actually anything new:

```ts
type Leak<T> = {
  [key: string & { tag: T }]: string;
}

type T0 = Leak<void>;  // Index signature key type not instantiated so type parameter leaks
```

I seem to recall us discussing this at some point in the past but can't find anything. As I said above, it's harmless, but certainly doesn't look right.

Anyway, if we decide to stay

DanielRosenwasser:

Issue #54057, opened 2023-04-28
`Debug Failure. False expression: Paths must either both be absolute or both be relative`


loynoir:
# Bug Report

- `Debug Failure. False expression: Paths must either both be absolute or both be relative`
- Dynamic import
- Conditional import
- Optional import

### 🕗 Version & Regression Information

- This is a crash
- This changed between versions __5.0.4____ and _______
- This is the behavior in every version I tried, and I reviewed the FAQ for entries about _________
- I was unable to test this on prior versions because _______

### ⏯ Playground Link

[Playground link with relevant code](https://www.typescriptlang.org/play?#code/PTAEFkE9QYwewCYFNQHM5IM6gBZIE5JA)

https://github.com/loynoir/reproduce-typescript-54057

### 💻 Code

```ts
export const mod = await (async () => {
  const x: number = 0;
  switch (x) {
    case 0:
      return await import("./case0.js");
    case 1:
      return await import("./case1.js");
    default:
      return await import("./caseFallback.js");
  }
})();
```

### 🙁 Actual behavior

```
/path/to/node_modules/typescript/lib/tsc.js:113984
      throw e;
      ^

Error: Debug Failure. False expression: Paths must either both be absolute or both be relative
    at getRelativePathFromDirectory (/path/to/node_modules/typescript/lib/tsc.js:5717:9)
    at getLocalModuleSpecifier (/path/to/node_modules/typescript/lib/tsc.js:41911:206)
    at computeModuleSpecifiers (/path/to/node_modules/typescript/lib/tsc.js:41875:21)
    at getModuleSpecifiersWithCacheInfo (/path/to/node_modules/typescript/lib/tsc.js:41831:18)
    at getModuleSpecifiers (/path/to/node_modules/typescript/lib/tsc.js:41803:10)
    at getSpecifierForModuleSymbol (/path/to/node_modules/typescript/lib/tsc.js:48272:27)
    at symbolToTypeNode (/path/to/node_modules/typescript/lib/tsc.js:48317:23)
    at createAnonymousTypeNode (/path/to/node_modules/typescript/lib/tsc.js:47333:20)
    at typeToTypeNodeWorker (/path/to/node_modules/typescript/lib/tsc.js:47178:16)
    at typeToTypeNodeHelper (/path/to/node_modules/typescript/lib/tsc.js:46958:24)

Node.js v18.15.0
```

### 🙂 Expected behavior

No error

### Workaround
```diff
[...]

PR #56439, opened 2023-11-16
When reparsing top level await dont set original node for the new source file

sheetalkamat:
Source file before reparsing top level await doesnt contain important fields like `path`, `imports` etc that are present only on completely parsed file and are set later on by the program. So while its convinient to use factory methods to recreate new source file, dont set the original node since the parent pointers are not set yet anyways and shouldnt matter which source file those nodes belong to.

Fixes https://github.com/microsoft/TypeScript/issues/54057

### DIFF
--- a/src/compiler/factory/nodeFactory.ts
+++ b/src/compiler/factory/nodeFactory.ts
@@ -133,6 +133,7 @@ import {
     hasSyntacticModifier,
     HeritageClause,
     Identifier,
+    identity,
     idText,
     IfStatement,
     ImmediatelyInvokedArrowFunction,
@@ -510,6 +511,7 @@ export function addNodeFactoryPatcher(fn: (factory: NodeFactory) => void) {
  */
 export function createNodeFactory(flags: NodeFactoryFlags, baseFactory: BaseNodeFactory): NodeFactory {
     const update = flags & NodeFactoryFlags.NoOriginalNode ? updateWithoutOriginal : updateWithOriginal;
+    const setOriginal = flags & NodeFactoryFlags.NoOriginalNode ? identity : setOriginalNode;
 
     // Lazily load the parenthesizer, node converters, and some factory methods until they are used.
     const parenthesizerRules = memoize(() => flags & NodeFactoryFlags.NoParenthesizerRules ? nullParenthesizerRules : createParenthesizerRules(factory));
@@ -6135,7 +6137,7 @@ export function createNodeFactory(flags: NodeFactoryFlags, baseFactory: BaseNode
 
     function cloneSourceFile(source: SourceFile) {
         const node = source.redirectInfo ? cloneRedirectedSourceFile(source) : cloneSourceFileWorker(source);
-        setOriginalNode(node, source);
+        setOriginal(node, source);
         return node;
     }
 
@@ -6365,7 +6367,7 @@ export function createNodeFactory(flags: NodeFactoryFlags, baseFactory: BaseNode
         const clone = createBaseIdentifier(node.escapedText) as Mutable<GeneratedIdentifier>;
         clone.flags |= node.flags & ~NodeFlags.Synthesized;
         clone.transformFlags = node.transformFlags;
-        setOriginalNode(clone, node);
+        setOriginal(clone, node);
         setIdentifierAutoGenerate(clone, { ...node.emitNode.autoGenerate });
         return clone;
     }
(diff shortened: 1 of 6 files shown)


src/compiler/parser.ts
```diff
@@ -1971,8 +1971,14 @@ namespace Parser {
...
         // If we parsed this as an external module, it may contain top-level await
         if (!isDeclarationFile && isExternalModule(sourceFile) && sourceFile.transformFlags & TransformFlags.ContainsPossibleTopLevelAwait) {
+            const oldSourceFile = sourceFile;
             sourceFile = reparseTopLevelAwait(sourceFile);
-            setFields(sourceFile);
+            if (sourceFile !== oldSourceFile) {
+                setFields(sourceFile);
+                // While we are parsing the file, we dont want to use the prev source file as original node anywhere
+                // as it wont have information that gets set on the source file later by program (like paths, computed imports etc)
+                sourceFile.original = undefined;
```
jakebailey, line 1980: At a glance, needing to do this feels weird... My understanding is that this is needed because this is the only place in the parser that uses a `factory.update*` method.

What's interesting is that the factory has a mechanism (`NoOriginalNode`) which is set for the factory the parser uses, but it's not used consistently used inside the factory (search `setOriginalNode` in `nodeFactory.ts`). Part of me thinks that the better fix is to make sure we're always respecting that flag in the factory.

rbuckton: > it's not used consistently used inside the factory (search `setOriginalNode` in `nodeFactory.ts`). Part of me thinks that the better fix is to make sure we're always respecting that flag in the factory.

I agree with @jakebailey's assessment.

jakebailey: For fun, I tried replacing all uses of `setOriginalNode` with `update`, and everything passed except a bunch of source map tests; `update` also `setRange`s so not sure if the diff was good or bad (I just thought I'd try and see what happened).

sheetalkamat: Ok there is `cloneNode` method that uses `cloneSourceFile` method and is also used by update worker. Are clone methods suppose to set original node or use `update` of the factory. I am not sure i know the protocol there to know if just removing `setOriginalNode` from `cloneSourceFile` is correct thing to do or not.

jakebailey: I don't think removing it is the right thing; moreso that any use of `setOriginalNode` should be gated on the `NoOriginalNode` flag. `update` does that automatically because it is conditionally set based on the flag (then that determines if `setOriginalNode` is called), but given `update` also updates position info, it may be sufficient to just conditionalize any calls to `setOriginalNode` on the flag, or pull it out to a helper kinda like `update`.

src/compiler/parser.ts
```diff
@@ -1971,14 +1971,8 @@ namespace Parser {
...
         if (!isDeclarationFile && isExternalModule(sourceFile) && sourceFile.transformFlags & TransformFlags.ContainsPossibleTopLevelAwait) {
-            const oldSourceFile = sourceFile;
             sourceFile = reparseTopLevelAwait(sourceFile);
-            if (sourceFile !== oldSourceFile) {
-                setFields(sourceFile);
-                // While we are parsing the file, we dont want to use the prev source file as original node anywhere
-                // as it wont have information that gets set on the source file later by program (like paths, computed imports etc)
-                sourceFile.original = undefined;
-            }
+            setFields(sourceFile);
```
rbuckton, line 1975: this should probably still be behind the condition since there's no reason to set fields again if nothing changed.

rbuckton (approved):
This looks good, aside from the single comment here. I'd also recommend running a perf test to ensure this change doesn't introduce a deopt or regression due to `setOriginal` now being polymorphic.

src/compiler/factory/nodeFactory.ts
```diff
@@ -510,6 +511,7 @@ export function addNodeFactoryPatcher(fn: (factory: NodeFactory) => void) {
  */
 export function createNodeFactory(flags: NodeFactoryFlags, baseFactory: BaseNodeFactory): NodeFactory {
     const update = flags & NodeFactoryFlags.NoOriginalNode ? updateWithoutOriginal : updateWithOriginal;
```
jakebailey, line 513: This isn't needed for this PR, but if we instead use the new `setOriginal`, we could just rename `updateWithOriginal` to `update` and remove this conditional. But, maybe that's slower for some reason.

sheetalkamat:
@typescript-bot test this
@typescript-bot test top100
@typescript-bot run dt
@typescript-bot pack this

jakebailey (approved):

sheetalkamat:

src/compiler/factory/nodeFactory.ts
sheetalkamat: Perf is good with this change so did it

Issue #27273, opened 2018-09-21
Object spread drops index signature


ajafff:
**TypeScript Version:**  3.0.1

**Search Terms:**
spread index signature, spread indexer

**Code**

```ts
declare var v: Record<string, string>;

let v2 = {...v}; // works
let v3 = {a: '', ...v}; // index signature lost
let v4 = {...v, a: ''}; // same as above
let v5 = {a: 1, ...v}; // index signature lost, should be '{a: number, [x: string]: number | string}'
let v6 = {[Number()]: 1, ...v}; // empty object type ò,Ó, should be '{[x: string]: number | string}'
```

**Expected behavior:**
Index signature exists on all object types.

**Actual behavior:**
Index signature is lost if object literal contains another property assignment. I'm not entirely sure what the type of `v6` should be.

**Playground Link:** https://agentcooper.github.io/typescript-play/#code/CYUwxgNghgTiAEA3WSBc8BK4D2NgB4BnAFxgEsA7AcwBp4TzqA+AbgCg2IRikAmeALzwA3gDpxiAL4t4AelnwA7rgDWhTtyQBmQSKjoA5AbrjRUmfPiVQAD3pkqFKMQCuceBGwkNPRABZdMQk6fXgjaTkFQigAWwQoQngoACNsRBAfJABWQNCARhMJCMtrEDtCBydXd08SOkIAC2wXCGB4ZIQDYVCKFxiOmDoAbRt0BkoqAF10Xv6QGHgAH3pSCckDTMQANkChgDk+gYAKAEpp+AL4U3NI+BAYgAdiAE94bGSAK3AeF4eEACeaABl+pNFptDphYQjMarajnWYDJYrRhUdZAA

**Related Issues:**
#16694
#16373

RyanCavanaugh:
I think it's extremely arguable one way or the other. Certainly if you introduce a non-conforming property, the original intent of the index signature is lost.

jsingerlenovo:
Object.assign is able to handle the types correctly, why can't Object Spread?

```
declare var v: Record<string, string>;

const v3 = { a: '', ...v }; // index signature lost, type is only { a: string; }
const v7 = Object.assign({ a: '' }, v);  // works, type is { a: string; } & Record<string, string>

const s3 = v3.stuff; // Property 'stuff' does not exist on type '{ a: string; }'.ts(2339)
const s7 = v7.stuff; // works fine
```

joealden:
[...]

PR #56441, opened 2023-11-17
Experiment: Produce wide index signatures from the spread of an index signature

### LINKED ISSUE #56431: Object spreading produces wrong type with non-literal keys
### 🔎 Search Terms

object spread, object rest, object spreading

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about object spreading

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.2.2#code/CYUwxgNghgTiAEYD2A7AzgF0QLngb3gG0BrEAT10xgEsUBzAXVxQFcBbAIxBngF8BYAFDJ0WKLgIlylDDXpN4VWnT7wAvPiEBIcfADkegDTaAdGbBDeQA

### 💻 Code

```ts
declare const c: { [key: string]: number }
const a: { [key: string]: string } = {
	a: '',
	...c
}

```

### 🙁 Actual behavior

It allows numbers to be in the object, when they shouldn't be allowed.

### 🙂 Expected behavior

It shouldn't allow numbers.

Because `c` has non-literal keys, in this case `string`, which could be any string, it doesn't care what's in it.

andrewbranch:
Fixes #56431
Related: https://github.com/microsoft/TypeScript/issues/27273 (does not change the behavior of dropping index signatures from the left side of a spread)

This would need a lot more test coverage if it proves viable but wanted to run the top repos first.

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -13435,6 +13435,26 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         return emptyArray;
     }
 
+    function getSpreadIndexInfos(left: Type, right: Type): IndexInfo[] {
+        const rightInfos = getIndexInfosOfType(right);
+        if (!rightInfos.length) return emptyArray;
+        const leftInfos = getIndexInfosOfType(left);
+        const leftProperties = getPropertiesOfType(left);
+        const result: IndexInfo[] = [];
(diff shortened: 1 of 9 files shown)


andrewbranch:
@typescript-bot run dt
@typescript-bot test top200

Issue #9879, opened 2016-07-21
Proposal: Type Relationship API


weswigham:
They type checker has no public API for checking type relationships or for looking up types in a given scope. This can make writing semantic-analysis informed lint rules very difficult (either because you need to replicate a lot of logic, or because you simply don't have enough information). Since [tslint enabled semantic checks](https://github.com/palantir/tslint/commit/9248e056c1a652ee13597e68da26cacf0b87e134) we now have a popular consumer who would very much like to be able to perform meaningful semantic comparisons (See [this rule](https://github.com/palantir/tslint/pull/1394#issuecomment-234384650) for instance).

Proposal
======

A small new set of public API methods on the type checker:
```ts
interface TypeChecker {
  isIdenticalTo(a: Type, b: Type): boolean; // Exposes internal identity relationship
  isSubtypeOf(a: Type, b: Type): boolean; // Exposes internal structural subtype relationship
  isAssignableTo(a: Type, b: Type): boolean; // Exposes internal assignable relationship
  isComparableTo(a: Type, b: Type): boolean; // Exposes internal comparable relationship
  isInstantiationOf(a: GenericType, b: GenericType): boolean; // While not an internal relationship, it is a straightforward semantic question to ask an answer which requires internal APIs to answer

  lookupGlobalType(name: string): Type; // Looks up a global symbol named "name" with meaning SymbolFlags.Type (Identical to getGlobalType but without the expected arity based return value) - returns unknownType on failure
  lookupTypeAt(name: string, position: Node): Type // Resolve a symbol lexically at the position specified with meaning SymbolFlags.Type (identical to resolveName, but with less levers) - returns unknownType on failure

  getAnyType(): IntrinsicType;
  getStringType(): IntrinsicType;
  getNumberType(): IntrinsicType;
  getBooleanType(): IntrinsicType;
  getVoidType(): IntrinsicType;
  getUndefinedType(): IntrinsicType;
  getNullType(): IntrinsicType;
  getESSymbolType(): IntrinsicType;
  getNeverType(): IntrinsicType;
  getUnknownType(): IntrinsicType;
}
```

While this isn't a complete set of APIs for manipulating the type checker (I've left off an API to programatically create new types for comparison), it is enough to do meaningful analysis with existing types in the compilation without resorting to brittle (or often wrong) syntax-based solutions.

weswigham:
cc: @ahejlsberg @mhegazy @rbuckton @vladima

[...]

PR #52473, opened 2023-01-28
Expose getStringLiteralType and getNumberLiteralType on the TypeChecker, plus remove /** @internal */ from several useful methods.

sstchur:
Expose the following on the TypeChecker:
* getStringType
* getStringLiteralType
* getNumberType
* getNumberLiteralType
* getBooleanType
* getFalseType
* getTrueType
* getUndefinedType
* getNullType
* isTypeAssignableTo

Fixes #50694

I opened a suggestion issue #50694 to suggest exposing `getStringLiteralType` and `getNumberLiteralType`. The rationale was that they would help avoid questionable and hacky code in an ES-Lint rule I was working on and that they would be useful to other developers using the TypeChecker. All of the other methods were technically available on the TypeChecker, but not "visibile" due to having been marked /** @Internal */.  This PR exposes the two recommended in #50694 and removes the Internal designation on several others that are useful for developers leveraging the TypeChecker.

@andrewbranch suggested that I create the PR myself, so I did :-)

I read the Contributing.md, and I'm aware of the following:

> Features (things that add new or improved functionality to TypeScript) may be accepted, but will need to first be approved (labelled ["help wanted"](https://github.com/Microsoft/TypeScript/issues?q=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22) or in the "Backlog" milestone) by a TypeScript project maintainer in the suggestion issue. Features with language design impact, or that are adequately satisfied with external tools, will not be accepted."

And I'm also aware that I don't have such a label from a TS project maintainer, but at Andrew's suggestion, I have created the PR anyway.  Please let me know if there is anything else I can do to make the PR better or the review process easier.

PR #56448, opened 2023-11-17
Make isTypeAssignableTo public on TypeChecker

jakebailey:
For #9879

See:

- https://github.com/microsoft/TypeScript/issues/56450#:~:text=Exporting%20TypeChecker%27s%20isAssignableTo%20Method
- https://github.com/microsoft/TypeScript/issues/9879#issuecomment-1758108764
- https://github.com/microsoft/TypeScript/issues/9879#issuecomment-1812679850
- https://github.com/search?q=%2F%5C.isTypeAssignableTo%2F&type=code
- https://discord.com/channels/508357248330760243/1173997445919801504

This function has been exported on the checker since TypeScript 3.7. It's meaning has not changed in that time, and many people have begun depending on it.

`ts-eslint` _doesn't_ use it as it's internal (as they're trying to do "the right thing"), but using it would fix many many lint bugs, and wouldn't require waiting for 5.4 (as stated above, it's been around since 3.7 and has seemingly worked the same).

We've previously been unhappy with exporting relation APIs because (1) we're free to add and remove relations at any time, (2) we're free to change relations, and (3) it's sometimes hard to document what they are.

But, assignability is assignability; it's unlikely that this relation is going to go away, and if we change its meaning, it's probably the case that dependents want that behavior change too to match the language. That and its pervasiveness has made it such that we probably can't remove it anyhow (as they say, "load-bearing").

So, this PR just exports it. This does not fully cover #9879, of course, but it seems worse to not do anything when we could just export this one well-defined function so many are already using unsafely.

cc @JoshuaKGoldberg


src/compiler/types.ts
```diff
@@ -5158,7 +5158,7 @@ export interface TypeChecker {
     /** @internal */ getPromiseLikeType(): Type;
     /** @internal */ getAsyncIterableType(): Type | undefined;
 
-    /** @internal */ isTypeAssignableTo(source: Type, target: Type): boolean;
+    isTypeAssignableTo(source: Type, target: Type): boolean;
```
jakebailey, line 5161: This needs a doc; please leave suggestions!

JoshuaKGoldberg (commented):
_**YES!**_ Thank you very much for getting this started!

src/compiler/types.ts
JoshuaKGoldberg: Thinking on what I'd want as a user, it's mostly just knowing what kind of caching I'd have to worry about (since that was brought up by TS folks previously) and examples of what direction the assignability runs in. Vaguely:

```ts
/**
 * Performs an assignability check, including using persistent caches.
 * @returns Whether `source` is assignable to `type`.
 * @example
 * ```ts
 * declare const literal: ts.Type; // Type of "abc"
 * declare const primitive: ts.Type; // Type of string
 * isTypeAssignableTo(literal, literal); // true
 * isTypeAssignableTo(literal, primitive); // true
 * isTypeAssignableTo(primitive, literal); // false
 * isTypeAssignableTo(primitive, primitive); // true
 * ```
```

jakebailey: Ugh, I did forget that each speculative use of this API adds to a cache permanently.

JoshuaKGoldberg: oh no

jakebailey: Then again, we provide no API by which one can create a type, so that's actually whatever, pretty sure.

bradzacher (commented):
One thing that might be important is the ability to resolve a type in the context of a file. For example most of our "promise checking" rules currently do really funky checks on the type to see if it has specific signatures.

The ability to do like "grab the global promise type and check if thing is assignable to that type" would vastly simplify our logic.

Similar for like Function (which I believe we shim right now via "has at least one call signature").

Error is another case we've used a few times where we do some shonky thing like "keep checking the inheritance chain until we find the last parent, ensure that parent is named error and it is in the TS lib file"

src/compiler/types.ts
jakebailey: Added something similar to the above.

jakebailey:
My impression from the recent thread was that this function was all that was needed to make everyone happy; is that incorrect such that we need to discuss this more?

bradzacher:
This will definitely do for a huge chunk of new cases!
There's also our existing rules which... Approximate the API in some cases by manually pulling apart types and applying some heuristic.

If there's ways to fetch "common" types or construct new types based on common types then we could vastly simplify some existing rules.

jakebailey:
#52473 added quite a few of these, but I am curious what else you're looking for.

bradzacher:
The only thing missing would be the ability to get a type from the globals, I think. Or less generally - some of the more "spec standard" types like Promise.
Really `getPromiseType` would be the biggest one for us and would help us out given how many promise rules we have.

As an example the [`no-floating-promise` rule](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/rules/no-floating-promises.ts) and [`no-misused-promises` rule](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/rules/no-misused-promises.ts) rules have their own logic (some of which is based off [`ts-api-utils.isThenable`](https://github.com/JoshuaKGoldberg/ts-api-utils/blob/main/src/types/utilities.ts#L302-L327)) to vaguely determine if a type is a promise. They are mostly correct - but it's a few hundred lines of code that could generally be reduced to `checker.isTypeAssignableTo(type, anyPromise)`.

jakebailey:
This isn't public (yet), but theoretically you could use `getAwaitedType`; if it returns something different than the input, then the input has a promise.

fatcerberus:
> But, assignability is assignability; it's unlikely that this relation is going to go away

Given that assignability is the basic primitive the type system is built on (conditional types check assignability, generic constraints are based on assignability, `A & B` means the set of values assignable to both, heck even parameter contravariance is just assignability with the polarity reversed!), I'd be *very* surprised if it went away without, you know, rearchitecting the whole language from scratch.  Which I guess is what is meant by "load-bearing"

jakebailey:
> Which I guess is what is meant by "load-bearing"

In this instance, I meant that we couldn't remove the function without breaking the ecosystem, even though it's not public.

nickmccurdy:
@jakebailey The GitHub search in your original comment has non-JavaScript files, try this: https://github.com/search?q=%2F%5C.isTypeAssignableTo%2F&type=code
https://discord.com/channels/508357248330760243/1173997445919801504

bradzacher (commented):
think-amazing

jakebailey:
> This isn't a public method (yet), but theoretically you could use `getAwaitedType`; if it returns something different than the input, then the input has a promise.

I have not tried to stick this into your codebase, but testing on [ts-ast-viewer.com](https://ts-ast-viewer.com/#code/CYUwxgNghgTiAEYD2A7AzgF3ipAFGSAtgJZogBc2AroQEYgwDcAUKJLAsuljvkaSABMlFDXox4AH3hUUoAGbEUIYC1bhocRKkzw0xQgAcIFeHxJkAPKLoMAfCzabOOrMiMRiYKBhWVzAtZi9lJmBBYglpgwSgDmdqEBVtFxANoAuqG0SEgmUCgJ0kmRxUG2MBmF8LTEsUoYDkA) shows that it seems to work:

```
> function containsPromise(type) { return type !== checker.getAwaitedType(type) }
< undefined
> [checker.typeToString(type), containsPromise(type)]
< (2) ['number', false]
> [checker.typeToString(type), containsPromise(type)]
< (2) ['number | undefined', false]
> [checker.typeToString(type), containsPromise(type)]
< (2) ['Promise<number>', true]
> [checker.typeToString(type), containsPromise(type)]
< (2) ['Promise<number> | Promise<string> | Promise<boolea…| string[]> | Promise<bigint | Promise<number[]>>', true]
```

A single equality check is almost surely better than 100 lines of code!

I actually would have thought people would have already used `getAwaitedType` externally, but apparently not. I only see it used in our codefixes. I know a few people have asked for that one already on the Discord.

sandersn (approved):
One thing: it still might need some discussion of caching.

sandersn:
Were there specific concerns we brought up in discord? I don't remember.

jakebailey:
I don't recall any, besides what I mentioned in the thread here (then went back on once I realized that you couldn't just make random types and overflow the cache).

sandersn:
:shipit:

andrewbranch (approved):

JoshuaKGoldberg:
Rainn Wilson as Dwight Schrute from The Office tearfully saying 'thank you', happily

src/compiler/types.ts
```diff
@@ -5158,7 +5158,20 @@ export interface TypeChecker {
     /** @internal */ getPromiseLikeType(): Type;
     /** @internal */ getAsyncIterableType(): Type | undefined;
 
-    /** @internal */ isTypeAssignableTo(source: Type, target: Type): boolean;
+    /**
+     * Returns true if the "source" type is assignable to the "target" type.
```
SlurpTheo, line 5162:
```suggestion
     * Returns true if the `source` type is assignable to the `target` type.
```

PR #56449, opened 2023-11-17
Elide re-exports of unresolved type-only imports

### LINKED ISSUE #56445: Invalid JavaScript output with re-exported  unresolved import types (regression)
### 🔎 Search Terms

export type invalid JavaScript

- This changed between versions 5.1 and 5.2 

### ⏯ Playground Link

[Playground Link](https://www.typescriptlang.org/play?ts=5.1.6#code/JYWwDg9gTgLgBDAnmApnA3nAyhEKAqyaAvnAGZS5wDkAdAPRKrUDcAsAFCcoAeksGbLgJE4cYuw5A)

### 💻 Code

```ts
import type { SomeType } from './type';

export { SomeType  };
```

### 🙁 Actual behavior

TypeScript will generate the following JS:

```js
export { SomeType };
```

This JS is invalid since `SomeType` is not defined. Basically the `import type` is removed, but the `export` of the type was not.

### 🙂 Expected behavior

Type should be removed from export as in 5.1

In the example above the source code has errors, so it could be argued that the output does not have to be valid. The issue was however discovered when using `transpileModule` . `transpileModule` similarly can't by design resolve the types and will preserve them in the output even though the types actually exist. 

This is going to block us at Bloomberg from fully adopting TS 5.2 as we have workflows that depend on `transpileModule`

andrewbranch:
Fixes #56445

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -47581,7 +47581,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         }
         const target = getExportSymbolOfValueSymbolIfExported(resolveAlias(symbol));
         if (target === unknownSymbol) {
-            return true;
+            return !excludeTypeOnlyValues || !getTypeOnlyAliasDeclaration(symbol);
         }
         // const enums and modules that contain only const enums are not considered values from the emit perspective
         // unless 'preserveConstEnums' option is set to true
--- a/src/testRunner/unittests/services/transpile.ts
+++ b/src/testRunner/unittests/services/transpile.ts
@@ -662,4 +662,15 @@ export * as alias from './file';`,
             testVerbatimModuleSyntax: true,
         },
     );
+
+    transpilesCorrectly(
+        "Elides re-export of type-only import",
+        `
+        import type { Foo } from "./types";
+        export { Foo };
+        `,
+        {
+            options: { compilerOptions: { module: ts.ModuleKind.ESNext, target: ts.ScriptTarget.ESNext } },
+        },
+    );
 });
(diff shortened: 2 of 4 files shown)


src/compiler/checker.ts
```diff
@@ -47581,7 +47581,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         }
         const target = getExportSymbolOfValueSymbolIfExported(resolveAlias(symbol));
         if (target === unknownSymbol) {
-            return true;
+            return !excludeTypeOnlyValues || !getTypeOnlyAliasDeclaration(symbol);
```
jakebailey, line 47584: Does this need a test for the other condition? I think the test covers the latter only.

jakebailey (approved):

src/compiler/checker.ts
andrewbranch: `!excludeTypeOnlyValues` is never covered under any test because I’m pretty sure it’s never correct to pass `false` for that. I fixed one case that forgot to pass `true` for it in https://github.com/microsoft/TypeScript/pull/55097, and the last remaining caller didn’t change any baselines when I had it pass `true`. I made a note to look at it and probably remove the parameter altogether in a follow-up PR.

andrewbranch: Oh yeah, the last place that doesn’t pass `true` is checking whether to elide an ImportEqualsDeclaration, which doesn’t strictly matter because it’s not a runtime error to `require()` a module that doesn’t export anything. But it should probably be elided anyway, so it’s likely that parameter will go away and we’ll treat it as if it’s always `true`.

jakebailey: Sounds good, thanks!

andrewbranch:

PR #56453, opened 2023-11-18
Add GHA workflow that auto LKGs on pushes to release branches

jakebailey:
This eliminates a manual step when cherry picking changes.

Ironically, I think this PR would need to be cherry-picked to the latest release branch to actually work.

I have left the other LKG pushing from other workflows in place, but I can make those only commit partial state instead... it's just tricky because we have things like version strings that get updated and whatnot.


jakebailey:
Took #57835 instead for now.

.github/workflows/lkg-on-release-push.yml
```diff
@@ -0,0 +1,52 @@
+name: Update LKG on release branch push
+
+on:
+  push:
+    branches:
+      - 'release-*'
+    branches-ignore:
+      - 'release-[0-4].*' # Predates hereby
```
jakebailey, line 8: Apparently you can't have both of these 
image

PR #48044, opened 2022-02-26
fix(48034): Get a literal string of type intersection in a template literal type

islandryu:
Fixes #48034

PR #56458, opened 2023-11-19
Strip tags in placeholders

ahejlsberg:
This PR builds on #56434 to strip object type tags in template literal placeholders. We can merge this in case we decide to go that way.

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -17696,7 +17696,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
 
         function addSpans(texts: readonly string[], types: readonly Type[]): boolean {
             for (let i = 0; i < types.length; i++) {
-                const t = types[i];
+                const t = stripObjectTypeTags(types[i]);
                 if (t.flags & (TypeFlags.Literal | TypeFlags.Null | TypeFlags.Undefined)) {
                     text += getTemplateStringForType(t) || "";
                     text += texts[i + 1];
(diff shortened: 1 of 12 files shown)


ahejlsberg:
@typescript-bot test top100
@typescript-bot run dt

jakebailey:
Given the leak problem is not new, I personally prefer #56434 unless there's some major performance wins happening. I still need to test it (though I think you said you did?)

@typescript-bot pack this

@epmatsw Is there any chance you could try this build compared to nightly on your codebase from #52345? This is not a revert of #48044 (in that it doesn't reduce everything to `string`), so I'm curious as to the perf difference. Maybe if you're bored you can also try https://github.com/microsoft/TypeScript/pull/56434#issuecomment-1817043158

epmatsw:
@jakebailey 

Doesn't seem like either has much of an effect, but it does look like something in an upcoming release did!

|Version|Time (s)|
|-------|---------|
|5.2.2|67|
|5.3.2|67|
|5.4.0-dev.20231120|47|
|#56434|49|
|#56458|49|

jakebailey:
Curious; would be interesting to bisect that using something like https://www.npmjs.com/package/every-ts, e.g.:

```console
$ every-ts bisect start
$ every-ts bisect old 5.3.2
$ every-ts bisect new 5.4.0-dev.20231120
$ every-ts tsc ...
$ every-ts bisect <old|new>
...
```

epmatsw:
5.4.0-dev.20231107 seems to be the first "fast" version

jakebailey:
Oh, right. And that actually helped on your real-world case? That's good, at least, though a little suspect to be narrowing something to itself.

Issue #56459, opened 2023-11-19
Inference from annotated parameters of context-sensitive functions is order dependent 


Andarist:
### 🔎 Search Terms

inference annotated parameters context sensitive order

### 🕗 Version & Regression Information

- This is the behavior in every version I tried

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.4.0-dev.20231118#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXwxAGcMAeAFQBp4BBagIQD4AKHAIwCsAueAbwCh48TAA8e5ANyD4UHs1m1qo8QEp4AXkbwAbjizApQtnOPx6SjGPhUZPGms069B-gF8VPXfqn9CJZgJCyvAA5Ig4OCGU0gryPCQwWKgA5hYiDlq8rtFGJvEYiSlp1FAZfNluKj5+GAHSwWERUdKmzKYJSanCliVlWTm28HHwHUXd6RqZFe5SQA

### 💻 Code

```ts
declare function test<T, A, B>(obj: {
  ctx: T;
  a: (a: A, ctx: T) => void;
  b: (b: B, ctx: T, a: A) => void;
}): void;

test({
  ctx: 'foo',
  a: (a: string, ctx) => {},
  b: (b: string, ctx, a) => {},
});

test({
  ctx: 'foo',
  b: (b: string, ctx, a) => {},
  a: (a: string, ctx) => {},
});
```

### 🙁 Actual behavior

The second call doesn't infer properly

### 🙂 Expected behavior

I'd expect the second call to be inferred as good as the first one

### Additional information about the issue

related to https://github.com/microsoft/TypeScript/issues/47599

PR #56460, opened 2023-11-19
Infer from annotated parameters of context sensitive functions in the first inference pass

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/56459

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -36023,14 +36023,18 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         return signature.parameters.length > 0 ? getTypeAtPosition(signature, 0) : fallbackType;
     }
 
-    function inferFromAnnotatedParameters(signature: Signature, context: Signature, inferenceContext: InferenceContext) {
-        const len = signature.parameters.length - (signatureHasRestParameter(signature) ? 1 : 0);
-        for (let i = 0; i < len; i++) {
-            const declaration = signature.parameters[i].valueDeclaration as ParameterDeclaration;
+    function inferFromAnnotatedParameters(node: FunctionExpression | ArrowFunction | MethodDeclaration, contextualSignature: Signature, inferenceContext: InferenceContext) {
+        if (!node.parameters.length) {
+            return;
+        }
+        const len = node.parameters.length - (last(node.parameters).dotDotDotToken ? 1 : 0);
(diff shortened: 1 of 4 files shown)


Andarist:
@jakebailey could you run the extended test suite here? :)

jakebailey:
@typescript-bot test it

jakebailey:
@typescript-bot pack this

src/compiler/checker.ts
```diff
@@ -38913,6 +38917,12 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     return links.contextFreeType = returnOnlyType;
                 }
             }
+            if (checkMode & CheckMode.Inferential) {
+                const contextualSignature = getContextualSignature(node);
+                if (contextualSignature) {
+                    inferFromAnnotatedParameters(node, contextualSignature, getInferenceContext(node)!);
+                }
+            }
```
jakebailey, line 38925: Is it not a little weird for `getInferenceContext` and `inferFromAnnotatedParameters` to get called here? I'm not finding other examples where we're doing this sort of thing outside of a function intended to contextually check (which this function is not, `contextuallyCheckFunctionExpressionOrObjectLiteralMethod` is).

Andarist: hm, this might be the first time this is called like it but the placement of those calls here is what fixes the bug conceptually - as in, for the added test cases it's important to infer *before* contextually checking those functions

Contextual check of functions can fix "requested" inference type parameters and that creates an ordering issue for https://github.com/microsoft/TypeScript/issues/56459 as the parameter type "request" can happen before `inferFromAnnotatedParameters` even has a chance to be called on the other function that comes later in the source code.

And the similar problem happens when it comes to https://github.com/microsoft/TypeScript/issues/60047 . The conditional type asks for the inferred type of the type parameter and since by that time the `inferFromAnnotatedParameters` was not called it resolves to its constraint. Then the arguments fail to be typechecked against current parameter types in `getSignatureApplicabilityError` so the algorithm returns, never reaching the second inference pass that starts to include context-sensitive parameters.

So if those issues are meant to be fixed it's absolutely necessary to call `inferFromAnnotatedParameters` within the first inference pass and that doesn't check context-sensitive functions, that's deferred to the second inference pass. I don't see how to even attempt to fix this in any other way given the known limitations around context-sensitive expressions and their associated implementation details

jakebailey:
I feel like this may have had some overlap with #60964; would be interested in a rebase of this PR.

Andarist:
@jakebailey synced with main

PR #56466, opened 2023-11-19
Fixed an issue with broken `await using` declarations in `for of` loops

Andarist:
fixes what was reported by @evanw here: https://github.com/microsoft/TypeScript/pull/55558#issuecomment-1817595357

cc @rbuckton


src/compiler/checker.ts
```diff
@@ -49476,7 +49476,6 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                         return true;
                     }
                 }
-                return false;
```
Andarist, line 49479: That's the core of the fix for what was reported. `for await` loops never had a chance to reach `checkGrammarVariableDeclarationList` call in this function. Perhaps there are some other subtle issues in this function? Shouldn't the rule of thumb be here only to use `return true` as an early return?

src/compiler/transformers/esnext.ts
```diff
@@ -302,9 +303,9 @@ export function transformESNext(context: TransformationContext): (x: SourceFile
             // before handing the shallow transformation back to the visitor for an in-depth transformation.
             const forInitializer = node.initializer;
             Debug.assertNode(forInitializer, isUsingVariableDeclarationList);
-            Debug.assert(forInitializer.declarations.length === 1, "ForInitializer may only have one declaration");
+            Debug.assert(forInitializer.declarations.length <= 1, "ForInitializer may only have one declaration");
```
Andarist, line 306: That was a crash on those broken empty declaration lists produced by the added test cases.

Other broken inputs are somewhat handled by transformers. For example:
```ts
// input

for (const x of)

// es5 output

for (var _i = 0, _1 = ; _i < _1.length; _i++) {
    var x = _1[_i];
    ;
}
```

It's not that this particular output is problem-free... it's broken but at least the transform/emit doesn't crash.

src/compiler/transformers/esnext.ts
```diff
@@ -302,9 +303,9 @@ export function transformESNext(context: TransformationContext): (x: SourceFile
             // before handing the shallow transformation back to the visitor for an in-depth transformation.
             const forInitializer = node.initializer;
             Debug.assertNode(forInitializer, isUsingVariableDeclarationList);
-            Debug.assert(forInitializer.declarations.length === 1, "ForInitializer may only have one declaration");
+            Debug.assert(forInitializer.declarations.length <= 1, "ForInitializer may only have one declaration");
 
-            const forDecl = forInitializer.declarations[0];
+            const forDecl = firstOrUndefined(forInitializer.declarations) || factory.createVariableDeclaration(factory.createUniqueName("x"));
```
Andarist, line 308: I also wonder if this whole case shouldn't be a parse error and if it shouldn't be partially handled by `createMissingNode` and/or similar mechanisms

Andarist: The errors are now roughly equivalent after my change. However, IIUC, esbuild and Babel treat this as a syntax error - whereas TS currently treats this as a grammar error (in TS we end up with `sourceFile.parseDiagnostics.length === 0`). The same is true about some other existing cases like:

```ts
for (var of of) { }
```

Note that my understanding of syntax/grammar errors is somewhat superficial so maybe I'm just misinterpreting something here.

jakebailey: But, yeah, is there a fix which instead makes a missing node instead? That feels better given the debug assert above.

Andarist: > But, yeah, is there a fix which instead makes a missing node instead?

Possibly. I only learned about `createMissingNode` after I opened this PR - so I would have to explore that route and recheck how it's used in the codebase and if that would be a good fit here.

Andarist: I rechecked this and this `.declarations` is already `isMissingList`. That's a concept only really known to the parser though right now - I could move a util for checking that to some shared module but I don't think it's worth it.

It just seems to me that this transform here has been written with extra invariants that are not present in some other very similar situations:
```ts
// @target: es5
for (var  of a) { }
```

This is also a for-of loop but the `es2015` transform doesn't require this to have exactly one declaration, see [here](https://github.com/microsoft/TypeScript/blob/d027e9619fb8ca964df3885a536a67b5f813738b/src/compiler/transformers/es2015.ts#L3067). It just gracefully grabs the first one or creates a temp variable (so basically what I'm doing here).

I think it would just be best to follow the es2015 transform here, I'll adjust the code to use the same way of creating a temp variable and gonna remove this invariant entirely.

jakebailey (approved):
I feel like this is okay, but will defer to @rbuckton.

rbuckton (approved):

PR #56467, opened 2023-11-19
Fix "Extract to type alias" not available at end of span

### LINKED ISSUE #56401: "Extract to type alias" not available at end of span
```ts
interface Yadda<T> {
    x: T
}

export let blah: Yadda/*a*/<string>/*b*/;
```

**Expected**: At both `a` and `b`, *extract to type alias* is offered as a refactoring.
**Actual**: Only `a` offers the refactoring.

babakks:
Fixes #56401 

This PR somewhat refactors the way we find `firstType` token when looking for the closest type node. To fix the original issue, we had to also consider the previous token. Now, to keep the exisiting behavior as much as possible, this is done as a second fallback.

### DIFF
--- a/src/services/refactors/extractType.ts
+++ b/src/services/refactors/extractType.ts
@@ -23,6 +23,7 @@ import {
     getRefactorContextSpan,
     getRenameLocation,
     getTokenAtPosition,
+    getTouchingToken,
     getUniqueName,
     ignoreSourceNewlines,
     isArray,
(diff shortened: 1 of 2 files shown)


navya9singh:
@typescript-bot pack this

andrewbranch:
@typescript-bot pack this _please_

iisaduan (approved):
Looks good! This PR in general allows `extractType` to be triggered on the right end of a range that can be refactored. For example, you can now also trigger the refactor in the following locations 
``` 
interface Yadda<T = boolean/**/> { x: T/**/ }
```


Note: This currently is the only test for `extractType` that tests single-cursor behavior.

andrewbranch (approved):

babakks:
@iisaduan Thanks for your comment. I can add more tests to this or a follow-up PR. Whichever you deem fit.

babakks:
@iisaduan I've just added your suggestions to the same test file.

PR #56476, opened 2023-11-21
[wip] Replace disk path operations to use fileNames instead of path

sheetalkamat:


sheetalkamat:
@gluxon can you please try this build and see if this helps

sheetalkamat:
@typescript-bot pack this

sheetalkamat:
@typescript-bot test this
@typescript-bot test top100
@typescript-bot run dt

sheetalkamat:

sheetalkamat:

src/compiler/moduleNameResolver.ts
```diff
@@ -901,11 +900,19 @@ export interface NonRelativeModuleNameResolutionCache extends NonRelativeNameRes
     getOrCreateCacheForModuleName(nonRelativeModuleName: string, mode: ResolutionMode, redirectedReference?: ResolvedProjectReference): PerModuleNameCache;
 }
 
+/** @internal */
+export interface MissingPackageJsonInfo {
+    packageDirectory: string;
+    contents: boolean;
```
andrewbranch, line 906:
```suggestion
    directoryExists: boolean;
```

?

src/compiler/program.ts
```diff
@@ -2387,7 +2383,7 @@ export function createProgram(rootNamesOrOptions: readonly string[] | CreateProg
         // If the missing file paths are now present, it can change the progam structure,
         // and hence cant reuse the structure.
         // This is same as how we dont reuse the structure if one of the file from old program is now missing
-        if (oldProgram.getMissingFilePaths().some(missingFilePath => host.fileExists(missingFilePath))) {
+        if (forEachEntry(oldProgram.getMissingFilePaths(), missingFilePath => host.fileExists(missingFilePath))) {
```
andrewbranch, line 2386:
```suggestion
        if (forEachEntry(oldProgram.getMissingFilePaths(), missingFileName => host.fileExists(missingFileName))) {
```

right?

andrewbranch (approved):
I’m sure it would be incredibly disruptive, but it kind of feels like we would benefit from a `FileName` branded string type like `Path` so that a `Path` can’t be passed to a function that needs the name of an existing file with original casing.

sheetalkamat:
> I’m sure it would be incredibly disruptive, but it kind of feels like we would benefit from a `FileName` branded string type like `Path` so that a `Path` can’t be passed to a function that needs the name of an existing file with original casing.

We have this somewhat in server with normalized path but not in absolute path and we probably want that distinction as well .. but that’s too much

src/compiler/moduleNameResolver.ts
sheetalkamat: I kept the name because it becomes easier to check for things that come out of cache with same property names but I guess I can add helper for isPackageJsonInfo ?

PR #56485, opened 2023-11-21
Make AssertEntry and AssertClause interfaces

### LINKED ISSUE #56477: [5.3] Deprecated types (AssertEntry, AssertClause) redefined as type aliases, rather than interfaces
### 🔎 Search Terms

AssertEntry, AssertClause, deprecated

- This changed between versions 5.2 and 5.3

### Background

https://github.com/microsoft/TypeScript/blob/0abfb521d52c4f3b25569b42b16bdc512905c98b/src/compiler/types.ts#L3665-L3669

In TS 5.3 import assertions were replaced with import attributes.
The old AST nodes were retained in the in the codebase with `@deprecated` markers.
However they were converted from interfaces to type aliases.

This has caused issues with @typescript-eslint because our tooling supports back to TypeScript 4.4 (before the import assertions were added).
Our types have a deep relationship with the TS AST types - so in order for us to support eslint plugin consumers on older TS versions we need to be able to reference both new and old AST nodes.

In order to do this [we resort to a bit of a hack](https://github.com/typescript-eslint/typescript-eslint/blob/75c128856b1ce05a4fec799bfa6de03b3dab03d0/packages/typescript-estree/src/ts-estree/ts-nodes.ts#L3-L15) where we leverage interface declaration merging.
In a nutshell we declare an empty interface within the typescript module. On older versions of TS without the required AST node this defines the name so that our types compile fine for the user. On newer versions of TS with the required AST node - the declaration is a no-op.

This strategy has worked well for us for quite a few years! However with the release of 5.3 we now have a unique situation where an AST node was added AND removed within our supported version ranges. This would be fine for us and our trick would work - if it weren't for the fact that the nodes were redefined as type aliases, rather than interfaces with `extends`. This means we can't declaration merge them as before and instead we generate TS errors.

When Josh added the hack for the new nodes he added a `// @ts-ignore` to our codebase to silence the TS error and (as we've just learned) the line comment is elided from the `.d.ts` output. This means that anyone on TS 5.3 will see a type error.
In testing we can't even use a jsdoc `/** @ts-ignore */` here because TS will still error on the typescript lib side.

Could we consider changing the types instead so that they export the deprecated nodes as follows:
```ts
    /** @deprecated */
    interface AssertEntry extends ImportAttribute {}
    /** @deprecated */
    interface AssertClause extends ImportAttributes {}
```

For reference - we did work with the beta and RC locally within our repo to test this out.
However this problem wasn't visible within our repo because we have `skipLibCheck: true`.
So within the `typescript-estree` package the error was suppressed by the `// @ts-ignore`, and in the consuming packages the error was suppressed by `skipLibCheck`.

Additionally we didn't have a chance to try our release in a separate repo that used TS 5.3RC.

This is something we'll need to think about in future (cc @JoshuaKGoldberg)

andrewbranch:
Fixes #56477

### DIFF
--- a/src/compiler/types.ts
+++ b/src/compiler/types.ts
@@ -858,7 +858,7 @@ export const enum ModifierFlags {
     Protected =          1 << 2,  // Property/Method
     Readonly =           1 << 3,  // Property/Method
     Override =           1 << 4,  // Override method.
-    
+
     // Syntactic-only modifiers
     Export =             1 << 5,  // Declarations
     Abstract =           1 << 6,  // Class/Method/ConstructSignature
@@ -3663,10 +3663,10 @@ export interface ImportClause extends NamedDeclaration {
 export type AssertionKey = ImportAttributeName;
 
 /** @deprecated */
-export type AssertEntry = ImportAttribute;
+export interface AssertEntry extends ImportAttribute {}
 
 /** @deprecated */
-export type AssertClause = ImportAttributes;
+export interface AssertClause extends ImportAttributes {}
 
 export type ImportAttributeName = Identifier | StringLiteral;
 
--- a/tests/baselines/reference/api/typescript.d.ts
+++ b/tests/baselines/reference/api/typescript.d.ts
@@ -6020,9 +6020,11 @@ declare namespace ts {
     /** @deprecated */
     type AssertionKey = ImportAttributeName;
     /** @deprecated */
-    type AssertEntry = ImportAttribute;
+    interface AssertEntry extends ImportAttribute {
+    }
     /** @deprecated */
-    type AssertClause = ImportAttributes;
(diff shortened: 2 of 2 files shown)


fatcerberus:
Ugh.  I get that this is to fix a real-world break, but `interface Foo extends Bar {}` is just… yuck.

jakebailey (approved):

andrewbranch:
Getting in the weeds with DefinitelyTyped has conditioned me to see `interface Foo extends SomeInterface {}` as beautiful and `type Foo = SomeInterface` as unsettling

DanielRosenwasser (approved):
TIHI

andrewbranch:
🤔 

image

PR #56487, opened 2023-11-21
Fast path for unions of string literals only

ahejlsberg:
This PR adds a fast path for relating unions of only string literal, template literal, or string mapping types to the `string` type. Should have a positive impact on performance, though not sure how much.

### DIFF
--- a/src/compiler/types.ts
+++ b/src/compiler/types.ts
@@ -6154,6 +6154,8 @@ export const enum TypeFlags {
     IncludesInstantiable = Substitution,
     /** @internal */
     NotPrimitiveUnion = Any | Unknown | Void | Never | Object | Intersection | IncludesInstantiable,
+    /** @internal */
+    NotStringLiteral = ~(StringLiteral | EnumLiteral | TemplateLiteral | StringMapping)
 }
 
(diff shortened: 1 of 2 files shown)


ahejlsberg:
@typescript-bot test top100
@typescript-bot run dt

src/compiler/types.ts
```diff
@@ -6154,6 +6154,8 @@ export const enum TypeFlags {
     IncludesInstantiable = Substitution,
     /** @internal */
     NotPrimitiveUnion = Any | Unknown | Void | Never | Object | Intersection | IncludesInstantiable,
+    /** @internal */
+    NotStringLiteral = ~(StringLiteral | EnumLiteral | TemplateLiteral | StringMapping)
```
weswigham, line 6158: `EnumLiteral` could be set for number enums, and is redundant, since `StringLiteral` is always also set on stringy-enums.

```suggestion
    NotStringLiteral = ~(StringLiteral | TemplateLiteral | StringMapping)

```

sandersn:
@ahejlsberg this was a small speedup on Compiler-Unions. Will this PR be worth porting to Corsa eventually? If so, is it worth keeping the PR open until then? Otherwise I'd like to close it.

PR #56493, opened 2023-11-21
Add a new cherry-pick GHA workflow

jakebailey:
The old Pipelines workflow has been spotty as of late. This moves the functionality into a https://github.com/actions/github-script script. Theoretically this is an improvement becuase:

- With GHA, we can do a blobless clone. Pipelines was doing a full clone each time, which was slow.
- This removes one of the two remaining Pipelines that use scripts in this repo; the only remaining one is the tarball task (which I plan to move too).
- We no longer need to use the `typescript-bot` fork to produce cherry picks; it will just make them on this repo with a specific name. This speeds things up because the bot's fork doesn't get updates so cherry-picks tend to need to push lots of extra info to the fork.
- Rerunning a cherry pick for a PR will force push over top of the old code and then update the PR info; the old task could do this but didn't say so.

Downsides:

- GHA is event based, so the trigger to run the workflow has no clue what build it started. There's a workaround for that I have yet to implement as it's quite annoying. This is not new and already happens to the sync/version tasks.
- This script is a string in a file; I made it manually in another file and made sure it typechecked, but maybe it breaks in the future. But, I really want to drop octokit as a dep of the repo.

~I still need to make a PR to the bot trigger function to point to this.~ https://github.com/microsoft/typescript-bot-test-triggerer/pull/32


.github/workflows/create-cherry-pick-pr.yml
```diff
@@ -0,0 +1,131 @@
+name: Create cherry pick PR
+
+on:
+  push:
+    branches:
+      - orta-twoslash-repros
+  workflow_dispatch:
```
jakebailey, line 7: The existing bot-triggered GHA workflows use `repository_dispatch`, but I'm like 99% sure that they are only accidentally working, given my editor says:

image

.github/workflows/create-cherry-pick-pr.yml
```diff
@@ -0,0 +1,131 @@
+name: Create cherry pick PR
+
+on:
+  push:
+    branches:
+      - orta-twoslash-repros
```
jakebailey, line 6:
```suggestion
  push:
    branches:
      - orta-twoslash-repros
```

.github/workflows/create-cherry-pick-pr.yml
jakebailey: Eh, it's just supposed to be an array. But, these two have different formats which is quite annoying.

sheetalkamat (approved):

jakebailey:
I'm going to update this PR to not delete the old script; that way I can merge it and then test it via a workflow dispatch to double check that it works.

jakebailey:
Sweet, it works.

jakebailey:

Issue #56482, opened 2023-11-21
[5.3] Type narrowing regression with property comparison to boolean false


meyfa:
### 🔎 Search Terms

narrowing, strict equals, property access, boolean false, undefined

### 🕗 Version & Regression Information

- This changed between versions 5.2 and 5.3
- This also happens in the nightly

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.3.1-rc#code/JYOwLgpgTgZghgYwgAgGIHt3IN4ChnIBGcUAXEZgDYRwi4C+uuC6IAzmMgOYRgZYBeZAAoAlOX7IAPsgCuIACYQYoCAuQCAfDnzIovWVBDIAsnDAALAHRRaC9AFsxybQAYrAVmQB+HERLk8JRsKPTI5PJKKiBqDEws7JwwmBrcvPxiuMAwIsno3lbEUBoCQkEhojoECWzo1FaU6FzCeaJxuAD0HcgAAmBsALQQAB4ADhAIYENQUOjFA8h5yMBsyKPobGzAhJQAnnKKyqoKzKy19Y3NeYUkbUA

### 💻 Code

```ts
interface Foo {
  bar: boolean
}

const getFoo = (): Foo | undefined => {
  return Math.random() > 0.5 ? { bar: false } : undefined
}

const foo = getFoo()
if (foo?.bar === false) {
  console.log(foo)
}

// @ts-expect-error - foo is possibly undefined
console.log(foo.bar)
```

### 🙁 Actual behavior

In line 13 and following, `foo` is incorrectly narrowed to just `Foo` (whereas it was previously `Foo | undefined`), although we learned nothing about it from the preceding `if` as that did not influence control flow.

```
Unused '@ts-expect-error' directive.(2578)
```

### 🙂 Expected behavior

TS should report an error when accessing `foo.bar` as `foo` can potentially be undefined.

### Additional information about the issue

I originally discovered this issue in JSX with a snippet similar to:

```tsx
function Component () {
  const foo = getFoo()
  return (
    <div>
      {foo?.bar === false && 'foo'}
      {foo.bar ? 'true' : 'false'}
    </div>
  )
}
```

There as well, the access `foo.bar` is allowed due to the preceding check, although they are completely separate.

Note that changing it to `=== true` or removing the checks entirely fixes the problem.

Andarist:
[...]

PR #56504, opened 2023-11-22
Fixed an issue in boolean comparison narrowing when the reference is an optional chain

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/56482

cc @jakebailey

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -27929,10 +27929,10 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     if (isMatchingConstructorReference(right)) {
                         return narrowTypeByConstructor(type, operator, left, assumeTrue);
                     }
-                    if (isBooleanLiteral(right)) {
+                    if (isBooleanLiteral(right) && !isAccessExpression(left)) {
                         return narrowTypeByBooleanComparison(type, left, right, operator, assumeTrue);
                     }
-                    if (isBooleanLiteral(left)) {
+                    if (isBooleanLiteral(left) && !isAccessExpression(right)) {
                         return narrowTypeByBooleanComparison(type, right, left, operator, assumeTrue);
                     }
                     break;
--- a/tests/baselines/reference/controlFlowOptionalChain3.errors.txt
+++ b/tests/baselines/reference/controlFlowOptionalChain3.errors.txt
@@ -1,0 +1,40 @@
+controlFlowOptionalChain3.tsx(30,8): error TS18048: 'foo' is possibly 'undefined'.
+
+
+==== controlFlowOptionalChain3.tsx (1 errors) ====
+    /// <reference path="/.lib/react16.d.ts" />
+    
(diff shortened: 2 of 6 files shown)


src/compiler/checker.ts
```diff
@@ -27929,10 +27929,10 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     if (isMatchingConstructorReference(right)) {
                         return narrowTypeByConstructor(type, operator, left, assumeTrue);
                     }
-                    if (isBooleanLiteral(right)) {
+                    if (isBooleanLiteral(right) && !isAccessExpression(left)) {
```
Andarist, line 27932: It turns out that the original `type` can be reassigned based on `optionalChainContainsReference`. That messed up the else branch because `narrowTypeByBooleanComparison` was operating on the wrong type.

I've also tried keeping `originalType` around and pass that to `narrowTypeByBooleanComparison`. That fixed the reported issue~ related to the wrong type in the else branch but it broke the same type within the if block. This happened because originally the intention was for that reassigned `type` to be returned through the `return` at the bottom of this function but I changed it to return `originalType` through `narrowTypeByBooleanComparison`.

So all in all, this seems like the simplest fix and reflects the intention the best. The motivation behind the original change was to improve non-access expression narrowing. Access expressions are already handled by discrimination-based narrowing etc. 

Even though it worked previously (as far as I know) for non-optional chain access expressions, this helps them avoid some redundant work too.

jakebailey:
@typescript-bot test top200
@typescript-bot run dt

@typescript-bot pack this

src/compiler/checker.ts
jakebailey: This seems fine as a quick fix for now, though I suspect that we could try and do better later.

jakebailey (approved):

jakebailey:

src/compiler/checker.ts
jakebailey: Then again, not sure what "better" even means; an expression like `(options?.a === false || options.b` is not the same as `(!options?.a || options.b)` so realistically I don't think there's any extra info gained here.

jakebailey:

jakebailey:

DanielRosenwasser:
I don't understand, isn't `isAccessExpression` too big of a hammer?

DanielRosenwasser:
Never mind, I think I understand now.

jakebailey:
> Is using `isAccessExpression` just intended to not check more deeply on the expression?

If you write an expression like `options?.a === false || options.b` or `options?.a === undefined || options.b`, it's not the same as like `!options?.a || options.b`; specifically, it doesn't provide any information about whether the equality was done on `options` or `options.a`, so we can't know if `options` was undefined or not for the following expression. Kinda makes sense; we're allowing a narrowing based on a _specific_ falsy value.

It's possible that this could work with just `==`, though, but I think the big hammer is okay.

Issue #56505, opened 2023-11-22
Contextual `never` parameter type gets precedence over the initializer's type


Andarist:
### 🔎 Search Terms

contextual parameter type initializer never

### 🕗 Version & Regression Information

- This is the behavior in every version I tried

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.4.0-dev.20231122#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXwxAGcMAeAKHngBUBhPQgDwwBpKaBZEDACx2CLwQzEKgHwASuBwxgpEjCyoA5i3gAKTIwBcNeqiat4AOlOxlRXahAA3EDADaAXQCU8ALwA+eGgDWqHAB3VE82T00GYQxdOkjmNQBbbj4BGK5efiIXXRscLGAAbnJyQhJ1dgBvdiowHDRo+AAGNioAXxb4KqoannBfABksBKwGzQxGRKhGD3gAVkbGty9O9ur4HigiAEEwMGJLDS01ZCJ7XQr4VCgk3QUlZXhWpe8K1ba2FyKgA

### 💻 Code

```ts
declare function test<
  TContext,
  TMethods extends Record<string, (ctx: TContext, ...args: never[]) => unknown>,
>(context: TContext, methods: TMethods): void;

test(
  {
    count: 0,
  },
  {
    checkLimit: (ctx, max = 500) => {},
    hasAccess: (ctx, user: { name: string }) => {},
  },
);
```

### 🙁 Actual behavior

`never` gets assigned as the type of `max` parameter and TS errors on the initializer's type mismatch:
> Type 'number' is not assignable to type 'never'.(2322)

### 🙂 Expected behavior

I'd expect the type inferred from the initializer to take precedence over `never` here. After all, we could annotate this contravariant position with `max: number = 500` and that wouldn't violate any constraints.

fatcerberus:
For what it's worth, this isn't specific to `never` and in fact is a more general issue that TS doesn't care about the initializer type at all: if the contextual parameter type is, say, `1 | 2 | 3`, then this is *also* an error when it doesn't need to be (a function that takes `number` is assignable to one that takes `1 | 2 | 3`).

Andarist:
Good point. I can probably tweak this somewhat easily by checking the assignability between the initializer's type and the contextual type - just in the right direction.

PR #56506, opened 2023-11-22
Improve contextually typed parameters with initializers

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/56505


src/compiler/checker.ts
```diff
@@ -36065,8 +36065,11 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             const declaration = parameter.valueDeclaration as ParameterDeclaration;
             if (!getEffectiveTypeAnnotationNode(declaration)) {
                 let type = tryGetTypeAtPosition(context, i);
-                if (type && type.flags & TypeFlags.Never && declaration.initializer) {
-                    type = widenTypeInferredFromInitializer(declaration, checkDeclarationInitializer(declaration, CheckMode.Normal));
+                if (type && declaration.initializer) {
+                    const initializerType = widenTypeInferredFromInitializer(declaration, checkDeclarationInitializer(declaration, CheckMode.Normal));
+                    if (!isTypeAssignableTo(initializerType, type) && isTypeAssignableTo(type, initializerType)) {
```
fatcerberus, line 36070: Hmm... what happens in the case of mutually assignable types?

Andarist: How many of them do we have? There is, of course, `any` and I can also think of bivariant methods but I can't call more on the spot.

That said - if the types are mutually assignable then the old behavior is preserver (the contextual parameter type is preferred). Can you think of any scenario in which this would be problematic?

fatcerberus: I guess you’re right.  If the types are mutually assignable then it probably doesn’t matter which one is picked and in the absence of a type annotation it makes more sense to pick the contextual type (especially since it prevents inferred `any`s from sneaking in)

Andarist: > especially since it prevents inferred anys from sneaking in

That I'm not so sure about :p

```ts
const test: (arg: any) => void = (arg = 5) => {
  arg
  // ^? (parameter) arg: any
};
```

But inferring `number` would also not be good here either. It would only give the illusion of type-safety within the function while `test` would still be callable with just anything.

fatcerberus: I specifically meant inferring `any` from a default when there’s a more sane contextual type available.

Andarist:
Minimal repro case for the above:
```ts
type CanvasDirection = "RIGHT" | "LEFT"
interface GraphActions {
  setDirection: (direction: CanvasDirection) => void;
}
declare function create<T>(config: T): void
create<GraphActions>({
  setDirection: (direction = "RIGHT") => {
    direction // should be `CanvasDirection`
  },
})
```

sandersn:
Looks to me like this fix should only apply when the initialiser type is not a subtype of the contextual type. The bug's example was `initialiser=500, context=never`, right?

Actually, `isAssignableTo("RIGHT", "RIGHT" | "LEFT")` is true -- why then does it use the initialiser type instead of the contextual type? Does the widening give `string` for the initialiser instead? Maybe widening is not right here.

sandersn (changes requested):
Need to avoid the break found by the user tests. Probably has to do with widening the initialiser type?

src/compiler/checker.ts
```diff
@@ -36221,9 +36221,16 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
             const parameter = signature.parameters[i];
-            if (!getEffectiveTypeAnnotationNode(parameter.valueDeclaration as ParameterDeclaration)) {
-                const contextualParameterType = tryGetTypeAtPosition(context, i);
-                assignParameterType(parameter, contextualParameterType);
+            const declaration = parameter.valueDeclaration as ParameterDeclaration;
+            if (!getEffectiveTypeAnnotationNode(declaration)) {
+                let type = tryGetTypeAtPosition(context, i);
+                if (type && declaration.initializer) {
+                    const initializerType = checkDeclarationInitializer(declaration, CheckMode.Normal);
+                    if (!isTypeAssignableTo(initializerType, type) && isTypeAssignableTo(type, initializerType)) {
```
gabritto, line 36229: I think you want to use the widened initializer type in this second condition (`isTypeAssignableTo(type, initializerType)`), because consider the following example:

```ts
declare function test<
  TContext,
  TMethods extends Record<string, (ctx: TContext, ...args: (1 | 2)[]) => unknown>,
>(context: TContext, methods: TMethods): void;

test(
  {
    count: 0,
  },
  {
    checkLimit: (ctx, max = 3) => {},
  },
);
```
The initializer type is `3`, I think, and the contextual type is `1 | 2`. We will still error on that case, and we still want to pick type `number` over type `1 | 2`.

tests/cases/compiler/contextuallyTypedParametersWithInitializers3.ts
```diff
@@ -0,0 +1,16 @@
...
+
+interface GraphActions {
+  setDirection: (direction: CanvasDirection) => void;
+}
+
+export declare function create<T>(config: T): void;
+
+create<GraphActions>({
+  setDirection: (direction = "RIGHT") => {
+    direction; // CanvasDirection
```
gabritto, line 14: Can you modify this test so that instead of `direction` we have something like `f(direction)` where `f` takes a `CanvasDirection`? I think it makes more clear that this would break if we picked parameter type `string` over `CanvasDirection`.

gabritto (changes requested):

tests/cases/compiler/contextuallyTypedParametersWithInitializers3.ts
Andarist: Do u mean smth like this?
```ts
type CanvasDirection = "RIGHT" | "LEFT";

interface GraphActions {
  setDirection: (setter: (direction: CanvasDirection) => void) => void;
}

export declare function create<T>(config: T): void;

create<GraphActions>({
  setDirection: (setter = (direction: string) => {}) => {
    setter;
  },
});
```

gabritto: I meant something like:

```ts
type CanvasDirection = "RIGHT" | "LEFT";

interface GraphActions {
  setDirection: (direction: CanvasDirection) => void;
}

declare function takesDirection(x: CanvasDirection): void;
export declare function create<T>(config: T): void;

create<GraphActions>({
  setDirection: (direction = "RIGHT") => {
    takesDirection(direction);
  },
});
```

src/compiler/checker.ts
Andarist: done

gabritto:
@typescript-bot run DT
@typescript-bot test top100

gabritto:
@typescript-bot pack this

gabritto:
> I was also considering using `checkExpressionWithContextualType` - maybe there is a test case to be found here that would require it? One that would, for example, involve a context-sensitive function in the initializer?

I think that's worth thinking about, but to be honest I couldn't so far think about an example, so I don't know if it matters.

gabritto (approved):

jakebailey (approved):

PR #56515, opened 2023-11-22
Reduce intersections of constrained type variables and primitive types

ahejlsberg:
With this PR we reduce intersections of constrained type variables and primitive types as follows:

* An intersection `T & P` or `P & T`, where `T`'s constraint is a subtype of `P`, is reduced to just `T`. For example, given `T extends "a" | "b"`, the intersection `T & string` is reduced to just `T`. 
* An intersection `T & P` or `P & T`, where no constituent of `T`'s constraint is a subtype of `P`, and `P` is not a subtype of `T`'s constraint, is reduced to `never`. For example, given `T extends "a" | "b"`, the intersection `T & number` is reduced to `never`.
* In a union containing intersections of a type variable and primitive types that fully cover the constraint of that type variable, the intersections are replaced with just the type variable. For example, given `T extends "a" | "b"`, the union type `T & "a" | T & "b"` is reduced to just `T`.

### DIFF
--- a/tests/baselines/reference/unknownControlFlow.types
+++ b/tests/baselines/reference/unknownControlFlow.types
@@ -819,47 +819,29 @@ function fx1<T extends unknown>(value: T & ({} | null)) {
 
 function fx2<T extends {}>(value: T & ({} | null)) {
 >fx2 : <T extends {}>(value: T & ({} | null)) => void
->value : T & ({} | null)
+>value : T
 
     if (value === 42) {
 >value === 42 : boolean
->value : T & ({} | null)
+>value : T
 >42 : 42
 
         value;  // T & {}
->value : T & ({} | null)
+>value : never
     }
     else {
         value;  // T & ({} | null)
->value : T & ({} | null)
+>value : T
     }
 }
 
 function fx3<T extends {} | undefined>(value: T & ({} | null)) {
 >fx3 : <T extends {} | undefined>(value: T & ({} | null)) => void
->value : T & ({} | null)
+>value : T & {}
 
     if (value === 42) {
 >value === 42 : boolean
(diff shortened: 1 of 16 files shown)


ahejlsberg:
@typescript-bot test top100
@typescript-bot run dt

tests/baselines/reference/unknownControlFlow.types
```diff
@@ -819,55 +819,55 @@ function fx1<T extends unknown>(value: T & ({} | null)) {
...
 
     if (value === 42) {
 >value === 42 : boolean
->value : T & ({} | null)
+>value : T
 >42 : 42
 
         value;  // T & {}
->value : T & ({} | null)
+>value : never
```
jakebailey, line 830: This seems funky; `T` could be `42` here but it's thinking that comparison is impossible and made `never`.

ahejlsberg: It's actually existing behavior that is being revealed by the reduction of `T & ({} | null)` to just `T`. We have logic somewhere in the comparable relation that considers a naked type parameter constrained to `{}` to be something that always represents an object.

gabritto:
@typescript-bot pack this

weswigham (approved):
I had to include a similar change in the `not` types PR to get them to work decently - I think this is a pretty good simplification to perform on union construction (since making the type smaller is basically always good).

The one thing I worry a bit about though, if when you reduce `T & U` to just `T`, you remove the intersection entirely, which means you remove the type an alias could have been pinned to, which could be an observable change. Hopefully it doesn't matter in practice - I'm hoping there's no `type Int<T extends number> = T & number;` style types too prevalent in the wild.

Issue #52998, opened 2023-02-27
Built-in Symbol.iterator methods return IterableIterator<T> extending Iterator<T>, not Iterator<T, void>


lionel-rowe:
# lib Update Request

The built-in `Symbol.iterator` methods return `IterableIterator<T>` extending `Iterator<T>`, not the expected `Iterator<T, void>`. For example:

```ts
const char = string[Symbol.iterator]().next().value
```

The expected type of `char` is `string | undefined`, but instead it's `any`. This isn't correct — it will never be anything other than a string (strictly a single Unicode char) or `undefined`.

This looks to be related to https://github.com/microsoft/TypeScript/issues/33353, which was closed due to backwards-compatibility concerns; however, it could presumably fixed without changing the default types of `IterableIterator`, instead changing its definition to take an extra type parameter:

```diff
- interface IterableIterator<T> extends Iterator<T> {
+ interface IterableIterator<T, TReturn = any> extends Iterator<T, TReturn> {
      [Symbol.iterator](): IterableIterator<T>;
  }
```

And then changing all the built-ins (`String`, `Array`, etc) like so:

```diff
  interface String {
      /** Iterator */
-     [Symbol.iterator](): IterableIterator<string>;
+     [Symbol.iterator](): IterableIterator<string, void>;
  }
```

Could that work as a fix and be sufficiently backward-compatible?

## Configuration Check

My compilation *target* is `ES2020` and my *lib* is `the default`.

## Missing / Incorrect Definition

Various — any built-in JS type with a `Symbol.iterator` method, such as `Array`, `String`, `Map`, `Set`, `Uint8Array`, etc.

## Sample Code

```ts
const char = string[Symbol.iterator]().next().value
// expected type: string | undefined
// actual type: any

const definitelyChar = nonEmptyString[Symbol.iterator]().next().value!
// expected type: string
// actual type: any
```

## Documentation Link

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/@@iterator

fatcerberus:
Somewhat of a tangent but `void` is not a synonym for `undefined` and shouldn’t be treated as such.  See https://github.com/microsoft/TypeScript/issues/42709.

lionel-rowe:
> Somewhat of a tangent but `void` is not a synonym for `undefined` and shouldn’t be treated as such. See #42709.

Thanks for the correction! I've fixed my original comment to be more accurate about this.

[...]

Issue #36239, opened 2020-01-16
Infer non returning function to be returning `undefined`, not `void`


sharno:
## Search Terms
non returning, undefined instead of void, undefined or unknown instead of void

## Suggestion
- Change return of non returning functions to `undefined` instead of `void`
- Accept `unknown` as a type of non returning functions
examples:
```ts
const f = () => {} // this gets inferred as () => void

const f1 = (): undefined => {} // doesn't work
// error: A function whose declared type is neither 'void' nor 'any' must return a value.(2355)

const f2 = (): unknown => {} // same error
```

## Motivations
`void` is an annoying type to deal with, has many inconsistencies and there's a ton of issues around it for its weird behavior like: #35850 , #35236 , #33420 ... etc
Non-returning functions actually return `undefined` which is a more predictable type to deal with and narrow disjunctions with and even use optional chaining with.

## Use Cases
Many use cases appear while using promises where there's a need to catch an error but at the same time not deal with `void`
```ts
const f = () => {}
async (p: Promise<A>) => {
  const a = await p.catch(f);
  // do stuff with a whose type is now A | void
};
```

## Checklist

My suggestion meets these guidelines:

> This wouldn't be a breaking change in existing TypeScript/JavaScript code
Not sure if this might break some code or not but I think any code that dealt with void should work while dealing with undefined?

* [x] This wouldn't change the runtime behavior of existing JavaScript code
* [x] This could be implemented without emitting different JS based on the types of the expressions
* [x] This isn't a runtime feature (e.g. library functionality, non-ECMAScript syntax with JavaScript output, etc.)
* [x] This feature would agree with the rest of [TypeScript's Design Goals](https://github.com/Microsoft/TypeScript/wiki/TypeScript-Design-Goals).

nmain100:
I like the current behavior.  If I write a function with no `return <expr>` in it at any point, then there's no reason for me to try to observe the return value of it, and any attempt to observe the return value is probably a mistake.  I like it when Typescript catches my mistakes.

sharno:
`undefined` is a more specialized and narrower type that `void`
My guess is If a function always returns `undefined` there's no reason to observe its return type either.

Another case is if you are using a function of type `() => void` in the `catch` part of a promise, you'll end up with a type like this: `Promise<A | void>`, at this point you care to know that your catch isn't actually returning `A` which will be hard to handle

falsandtru:
[...]

PR #56517, opened 2023-11-23
Add `BuiltInIterableIterator` for built-in iterables with `undefined` return values

lionel-rowe:
Fixes https://github.com/microsoft/TypeScript/issues/52998

Adds `BuiltInIterableIterator`, a stricter version of `IterableIterator` which is adhered to by all built-in iterables. This ensures correct types for `next().value` and for the return value of `yield*`:

```ts
declare const str: string
const iter = str[Symbol.iterator]()

// old type: any
// new type: string | undefined
const maybeChar = iter.next().value

function* generator() {
    // old type: any
    // new type: undefined
    const val = yield* [1, 2, 3]
}

// old type: Generator<number, any, undefined>
// new type: Generator<number, undefined, undefined>
function* generator2() {
    return yield* [1, 2, 3]
}

// old version: TS allows this
// new version: Property 'return' does not exist on type 'BuiltInIterableIterator<number>'.
iter.return?.()
// old version: TS allows this
// new version: Property 'throw' does not exist on type 'BuiltInIterableIterator<number>'.
iter.throw?.()
```

`IterableIterator` still remains, and should probably be preferred for usage as a parameter type to maintain backward compatibility and to allow lenience where the return value is discarded.

Currently out of scope for this PR: corresponding changes to DOM libs; any equivalent changes to async iterables.


Jack-Works:
Maybe the name `IterableIteratorWithNoReturn` is better? built-in does not definitely means it will return undefined, and user land iterators can also return undefined.

lionel-rowe:
> Maybe the name `IterableIteratorWithNoReturn` is better? built-in does not definitely means it will return undefined, and user land iterators can also return undefined.

Yeah, it'd be worth considering alternatives for the name. Some other possibles, per https://github.com/microsoft/TypeScript/issues/52998#issuecomment-1822085956:

> `GenericIterableIterator` (not to be confused with TS generics) ... `StrictIterableIterator`, `ImplicitReturnIterableIterator`, `StandardIterableIterator`

Personally I prefer the nomenclature "implicit return" over "no return", because the latter could also refer to a non-terminating function/generator (`function* noReturn() { while (true) yield 1; }`)

However, bikeshedding about names may be premature — per https://github.com/microsoft/TypeScript/issues/52998#issuecomment-1825100980, this will likely remain blocked pending https://github.com/microsoft/TypeScript/issues/36239, unless someone smarter than me can see a non-breaking way forward.

rbuckton:
To preserve backwards compatibility, in #58243 we've opted to introduce a strict-mode flag to control the expected return type for a built-in iterator. We've also adopted `BuiltinIterator` as part of introducing types to support the iterator helpers proposal, per #58222.

There is a slight conflation of terms here, however. #58222 introduces `BuiltinIterator` as a place to hang the iterator helpers methods so as not to collide with TypeScript's `Iterator`, which merely describes the minimal interface for iteration. However, all generators are also derived from `BuiltinIterator`, so we cannot specify a default return type lest we conflict with existing generator functions. In #58243 we introduced the `BuiltinIteratorReturn` type alias, which is an `intrinsic` type whose actual type is either `undefined` or `any` depending on whether the `--strictBuiltinIteratorReturn` compiler option is set. Since this is a strict-mode flag, anyone with `strict: true` in their configurations will get this new behavior automatically, but can manually opt out by setting `strictBuiltinIteratorReturn: false`. Anyone not compiling under strict mode will retain the existing behavior where the return type is `any`.

Issue #56474, opened 2023-11-20
Latter type parameters are not suggested in constraints of the preceding ones


Andarist:
### 🔎 Search Terms

type parameters constraints completions

### 🕗 Version & Regression Information

- This changed in https://github.com/microsoft/TypeScript/pull/52690

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.4.0-dev.20231120#code/GYVwdgxgLglg9mABFApgZygHgGIwE4aIoAeqYAJmogMoD0AVPbQDQ0oQLkB8AFAIYAuRLgJRWAIyHV2nAJSIA3gF8AUCqA

### 💻 Code

```ts
function test<First extends S/**/, Second>(a: First, b: Second) {}
```

### 🙁 Actual behavior

`Second` is not suggested at this marker

### 🙂 Expected behavior

I'd expect `Second` to be suggested since depending on the latter type parameters is valid (even if weird at times). 

### Additional information about the issue

This change was called out in comments of the PR that introduced this (initially it focused only on expression-space parameters) but I wonder if this is something that could be revisited. cc @zardoy

zardoy:
I didn't remember the exact [motivation](https://github.com/microsoft/TypeScript/pull/52690#issuecomment-1434979172) for filtering them out, probably I just found these completions annoying myself as I couldn't imagine a case where you would write `First extends Second, Second` instead of `First, Second extends First`, though there are pretty complex function signatures that still do this (just Vue.extend for example). Since these completions don't cause any type-errors, I think it would be more correct to revert the change for type parameters as you suggest.

Ideally checker and completion modules should use more methods in common so it's easier to filter out completions that can cause type errors to avoid such misalignments as this one.

PR #56543, opened 2023-11-26
Include all type parameters in completions within type parameters' constraints

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/56474


src/services/completions.ts
```diff
@@ -5767,20 +5771,39 @@ function isModuleSpecifierMissingOrEmpty(specifier: ModuleReference | Expression
     return !tryCast(isExternalModuleReference(specifier) ? specifier.expression : specifier, isStringLiteralLike)?.text;
 }
 
-function getVariableOrParameterDeclaration(contextToken: Node | undefined, location: Node) {
+function getClosestSymbolDeclaration(contextToken: Node | undefined, location: Node) {
```
Andarist, line 5774: since `Parameter` has a specific meaning and commonly doesn't include `TypeParameter` I decided to rename this (as this function might return both of those types)

src/services/completions.ts
```diff
@@ -2613,26 +2613,30 @@ export function getCompletionEntriesFromSymbols(
...
-                    (isParameter(variableOrParameterDeclaration) && isParameter(symbolDeclaration))
-                )
-            ) {
-                const symbolDeclarationPos = symbolDeclaration.pos;
-                const parameters = isParameter(variableOrParameterDeclaration) ? variableOrParameterDeclaration.parent.parameters :
-                    isInferTypeNode(variableOrParameterDeclaration.parent) ? undefined :
-                    variableOrParameterDeclaration.parent.typeParameters;
-                if (symbolDeclarationPos >= variableOrParameterDeclaration.pos && parameters && symbolDeclarationPos < parameters.end) {
-                    return false;
+            if (closestSymbolDeclaration && symbolDeclaration) {
```
Andarist, line 2624: The only real change here is the addition of `isInTypeParameterDefault` check. I decided to refactor this slightly though as I find handling the `parameters` and `typeParameters` cases through those separate (although similar) code paths cleaner/more readable

src/services/completions.ts
```diff
@@ -2613,26 +2613,30 @@ export function getCompletionEntriesFromSymbols(
...
+                    if (symbolDeclaration.pos >= closestSymbolDeclaration.pos && symbolDeclaration.pos < parameters.end) {
+                        return false;
+                    }
+                }
+                else if (
+                    isTypeParameterDeclaration(closestSymbolDeclaration) && isTypeParameterDeclaration(symbolDeclaration)
+                    && isInTypeParameterDefault(contextToken)
+                    && !isInferTypeNode(closestSymbolDeclaration.parent)
+                ) {
+                    const typeParameters = closestSymbolDeclaration.parent.typeParameters;
```
Andarist, line 2636: I believe that those are always available in this codepath - but I kept this as nullable to stay within the safety bounds just in case

andrewbranch (approved):
👍

Andarist:
@iisaduan would it be possible for you to take a quick look at this?

tests/cases/fourslash/completionsForAllTypeParametersInConstraints.ts
iisaduan, line 1: `function test<First extends First, Second>(a: First, b: Second) {}` and `type A1<K extends K, L> = K` are both circular definitions and are invalid code, so we should filter the current parameter out. I can't think of a situation where `First extends First` is the first part of some valid parameter and not offered by the completion for the valid case

iisaduan (changes requested):
Thanks for the reminder! Change generally looks good, but the self-completion is invalid, so we should continue filtering it out.

tests/cases/fourslash/completionsForAllTypeParametersInConstraints.ts
Andarist: Ok, I addressed this - could you take another look? :)

iisaduan (approved):
Looks good, thanks!

Issue #55382, opened 2023-08-15
Recursive reverse mapped types don't work well with mixed object and tuple structures


Andarist:
### 🔎 Search Terms

inference recursive reverse mapped type tuples arrays

### 🕗 Version & Regression Information

- This is the behavior in every version I tried

### ⏯ Playground Link

https://www.typescriptlang.org/play?#code/C4TwDgpgBAygxgCwgWwIZQLxQEoTgewCcATAHgGdhCBLAOwHMAaKAV1oGtb8B3WgPigAfKIQipi+WgBsQrDl14BtALoBuAFDrQkKABEIAMzrVg1SaQAqArAG91UKIoDSUOlHYQQ+A1AvKAXFAAFEEAlJgCFs7K4cL6RrQmZrSW0XwaAL4a6sR4UqiiUAZscKaSUHCiqMAQllAQAB41tMTksIgoqHxBuQlJkoHxxmUpVqGBFtkEtJQiEOQsUsCYFVU1QXYOqIFhEVAAjIz2UABGgZsOFTvhGAIARHdHDhlHGaHZAPQfUMAI1G2SaCJOAQGSuWgGCCENoAAxsUG2UFoLGQJyhqlO5yuUEoNAYUAyBJh6lECyWGi+UAAegB+TTTWakxbAABMK0qYnWil2twOzG5N3ud2UMU+31+-yggPBkOhUBhyNRUJUMNOLGWJhxCHwi2IMqh8sUirRhH5uLo9BFxKZSxZFO+tPpkkZ82ZAGZ2WsIBtjoieQJDsczo5-VAHspXu9NJSJQDaNA6LLYfDEcaocxg+aGCoiWqNctyNrdfrCPKU4E06bMY4s5aiSTXUs3fbqTSgA

### 💻 Code

```ts
type Schema = Record<string, unknown> | readonly unknown[];

type Definition<T> = {
  [K in keyof T]: (() => T[K]) | Definition<T[K]>;
};

declare function create<T extends Schema>(definition: Definition<T>): T;

const result = create({
  a: () => 1,
  b: {
    c: () => "",
  },
});

// this one nicely infers `{ a: number; b: { c: string } }`
result;
// ^?

const result2 = create([() => 1, [() => ""]]);

// this one infers `number[]` but it should infer `[number, [string]]`
result2;
// ^?

const result3 = create({
  a: () => 1,
  b: [() => ""],
});

[...]

PR #56555, opened 2023-11-27
Treat array literal contextually typed by homomorphic mapped types as in tuple context

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/55382


tests/cases/compiler/reverseMappedTupleContext.ts
```diff
@@ -0,0 +1,25 @@
...
+
+declare function test1<T>(arg: {
+  [K in keyof T]: T[K];
+}): T;
+const result1 = test1(["foo", 42]);
+
+declare function test2<T extends readonly unknown[]>(arg: {
+  [K in keyof T]: T[K];
+}): T;
+const result2 = test2(["foo", 42]);
```
Andarist, line 14: IMHO this alone is worth this improvement. Today we can infer tuples using mapped types but we need to awkwardly spread them ([TS playground](https://www.typescriptlang.org/play?ts=5.4.0-dev.20231126#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXwCMYcBrEVAHgBV4QAPDc4AZ3jimDwgE940TUOAO6oA2gF0AfAApYAcwBc8AN4AoePFEBpeFnxluORPCrilVbeIDcqgL4BKczbB5mGNiGbIIGAIzwAXkJiMlRpUQAiRBwcCIAaeAAWACZxexsAegyNDQA9AH5VVVBIWAQUdGw8eCEcGBJYHDRgaloGJlZ2TlQePlQBYTEpaXV4eSVRUY0AOlm1HJztXX0QQ2NTc0sbBds40fE9xxNnV3c4Lx9kwJq6huJm8KiY+KTU9NUshYLVIA)):
```ts
declare function broken<T extends readonly unknown[]>(arg: {
  [K in keyof T]: T[K];
}): T;
const result1 = broken(["foo", 42]);
//    ^? const result1: (string | number)[]

declare function workaround<T extends readonly unknown[]>(
  arg: [
    ...{
      [K in keyof T]: T[K];
    },
  ],
): T;
const result2 = workaround(["foo", 42]);
//    ^? const result2: [string, number]
```

That makes the recursive scenarios much more complicated because we can't "just" iterate over the recursive type parameter and have that tuple context assigned to array literals.

IMHO, the mapped type itself strongly indicates that the user *wants* to have the tuples inferred as they might want to type each element separately.

fatcerberus: It doesn’t infer the tuple type if `T` is marked `const`?  I thought that was the whole reason behind const type params, because people wanted to infer tuples instead of `(T | U | V)[]`

Andarist: > It doesn’t infer the tuple type if T is marked const?

It does (although not always, see this [TS playground](https://www.typescriptlang.org/play?ts=5.4.0-dev.20231126#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXwxAGcMAWAHjDxPgBV4QAPQ1YI+OKYPCAT3jQBrVDgDuqANoBdAHwAKWDABc8AN4AoePAkBpeFnyCQvHIjpSVtXVIDc6gL4BKS3fVVUNOEWQQy8ALwExGRyEgBEAIxhADTwAExSjnYA9MlaWgB6APzq6qCQsAgo6Nh4QSQArJTUGHQMzCCs7JzcqHwCqMJikrJymvAAdEOwAOZEKhrp2noG8EYmZrQWdNZ2WvbqznSu7p7EPhgVAeWHcqrwUCrhiDg4YVLwTilp6dlAA), but that is being fixed by https://github.com/microsoft/TypeScript/pull/55794

> I thought that was the whole reason behind const type params, because people wanted to infer tuples instead of (T | U | V)[]

I don't think that was the whole reason behind it. The reason was that people want to use schema-like arguments for which often strings, numbers, etc have to be kept as literal strings. The fact that tuples are also preserved is just a byproduct of this - since a tuple is a more specific type than an array, just like a string literal is a more specific type than a string type. 

Regardless... there are times when constness of the whole argument is not something that I want and preserving the "tuple context" is all that I want. I might want to infer `[string]` and not let's say `["foo"]`.

fatcerberus: > I might want to infer `[string]` and not let's say `["foo"]`.

Hmm, now that I think about this, there’s no precedent for that in the type system today, is there?  When inferring arrays *in situ* you either write `[ "foo", "bar" ]` and get `string[]` or else add an `as const` and get `readonly [ "foo", "bar" ]`.

So the upshot is that this change would add a way to infer `[ string, string ]` from an array literal via a generic and people will start writing more constrained identity functions to accomplish that (…and likely complain about having to do that 😜)

Andarist: You can already do this today - you have to use the workaround mentioned above. This PR makes this whole thing *simpler* while also making it possible (or just way easier) to write recursive types that preserve tuples like this (especially in the "mixed" scenarios when your type is supposed to iterate through both object and array types). 

I also think that the workaround above isn't that intuitive. After all an unconstrained generic mapped type when iterating over a tuple preserves its tupleness.I would also conceptually consider all array literal expressions to start as tuples~. It's the widening behavior that usually widens them to arrays unless some other hint that the tuple should be preserved is given. So I just expect that contextual mapped type to be considered as such a hint.

fatcerberus: That’s fair.  To be honest it just feels wrong that generic inference seems to have more knobs than what you can control inline.  In general, the fewer CIFs I have to write to get type inference to do what I want, the better.

Andarist:
@jakebailey would u be so kind as to create a playground for this one? :)

jakebailey:
@typescript-bot test top200
@typescript-bot run dt

@typescript-bot pack this

weswigham (approved):
We spoke about this in the design meeting finally and decided this was a good idea (tm) - hopefully it won't cause any unexpected weirdness in the broader community when merged. :)

PR #56561, opened 2023-11-27
Update dependencies to latest compatible versions

jakebailey:
We haven't done this in a bit. Most of these updates we already have via lockfile, but some are major updates that are compatible.

After this, ncu says:

```
Checking /home/jabaile/work/TypeScript/package.json
[====================] 41/41 100%

 chai   ^4.4.1  →  ^5.0.0
 chalk  ^4.1.2  →  ^5.3.0
 which  ^3.0.1  →  ^4.0.0

Run ncu --target latest -u to upgrade package.json
```

`chai` is now ESM only and breaks our use.

`chalk` would now work, because we only use it in ESM contexts. But, upgrading it actually manages to _increase_ the size of `node_modules` thanks to the pervasiveness of v4.

`which`'s minimum node version is now Node 16, but we still support and build with Node 14.


sandersn (approved):

src/cancellationToken/cancellationToken.ts
```diff
@@ -60,7 +60,7 @@ function createCancellationToken(args: string[]): ServerCancellationToken {
     }
     else {
         return {
-            isCancellationRequested: () => pipeExists(cancellationPipeName!), // TODO: GH#18217
+            isCancellationRequested: () => pipeExists(cancellationPipeName), // TODO: GH#18217
```
RyanCavanaugh, line 63:
```suggestion
            isCancellationRequested: () => pipeExists(cancellationPipeName),

```

RyanCavanaugh (approved):

src/cancellationToken/cancellationToken.ts
jakebailey: Ah, yeah, indeed; I did just this in #57099 which I pulled out to show specifically what is new in the latest TS.

sheetalkamat (approved):

PR #56565, opened 2023-11-27
fix(56562): Type with negative properties cause crash in infer return type

### LINKED ISSUE #56562: Type with negative properties cause crash in infer return type
### 🔎 Search Terms

negative numeric literals infer return type

- This is a crash

### ⏯ Playground Link

[Playground Link](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABADwBQEpEG8CwAoRQxAJwFMoRilcCi6BtAWgEYBdALkQAZ86BfRAEMAzoggJhUfHyA)

### 💻 Code

```ts
function x() {
    return {
        [-1]: 0
    } as const
}
```

### 🙁 Actual behavior

When using the refactor `Infer function return type` nothing happens. Extension host in VS Code shows a stack trace ending with:

```sh
Error: Debug Failure. False expression: Token end is child end
    at processChildNode (c:\dev\TSC\tmp\defaults-and-bindings-code-fix\node_modules\typescript\lib\tsserver.js:171847:17)
    at c:\dev\TSC\tmp\defaults-and-bindings-code-fix\node_modules\typescript\lib\tsserver.js:171778:9
```

### 🙂 Expected behavior

Should infer `{ readonly [-1]: 0 }` as it did in 

Found while working on isolated declarations.

The issue appears to be that the formatter does not deal well with negative numbers in computed properties. This was not a problem before our recent merge with master as the type printer appears to have been producing negative number literals and it now produces the correct `UnaryOperator(-, abs(value))`

The error occurs in this assertion:

```ts
            if (isToken(child)) {
                // if child node is a token, it does not impact indentation, proceed it using parent indentation scope rules
                const tokenInfo = formattingScanner.readTokenInfo(child);
                // JSX text shouldn't affect indenting
                if (child.kind !== SyntaxKind.JsxText) {
                    Debug.assert(tokenInfo.token.end === child.end, "Token end is child end");
                    consumeTokenAndAdvanceScanner(tokenInfo, node, parentDynamicIndentation, child);
                    return inheritedIndentation;
                }
            }

```

a-tarasyuk:
Fixes #56562

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -8331,7 +8331,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                         return factory.createStringLiteral(name, !!singleQuote);
                     }
                     if (isNumericLiteralName(name) && startsWith(name, "-")) {
-                        return factory.createComputedPropertyName(factory.createNumericLiteral(+name));
+                        return factory.createComputedPropertyName(+name < 0 ? factory.createPrefixUnaryExpression(SyntaxKind.MinusToken, factory.createNumericLiteral(-name)) : factory.createNumericLiteral(+name));
                     }
                     return createPropertyNameNodeForIdentifierOrLiteral(name, getEmitScriptTarget(compilerOptions), singleQuote, stringNamed, isMethod);
                 }
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -8331,7 +8331,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                         return factory.createStringLiteral(name, !!singleQuote);
                     }
                     if (isNumericLiteralName(name) && startsWith(name, "-")) {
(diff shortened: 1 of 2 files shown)


src/compiler/checker.ts
```diff
@@ -8331,7 +8331,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                         return factory.createStringLiteral(name, !!singleQuote);
                     }
                     if (isNumericLiteralName(name) && startsWith(name, "-")) {
-                        return factory.createComputedPropertyName(factory.createNumericLiteral(+name));
+                        return factory.createComputedPropertyName(+name < 0 ? factory.createPrefixUnaryExpression(SyntaxKind.MinusToken, factory.createNumericLiteral(-name)) : factory.createNumericLiteral(+name));
```
jakebailey, line 8334: Above, this branch only runs when the literal starts with `-` already; I think we could safely create the prefix unary expression unconditionally.

jakebailey (approved):

jakebailey:
@typescript-bot test top200
@typescript-bot run dt

PR #56422, opened 2023-11-15
Revert "Disallow numeric literals with negative numbers (#55268)"

jakebailey:
This reverts #55268.

Closes #56301
Updates #56272

For 5.3 it seemed too risky to take this change; there's a lot more inconsistency we've discovered at the last minute and changing it _again_ later seemed bad.

Also #56270 shows that this behavior was being depended on as TS lacks true infinity types.

PR #56570, opened 2023-11-28
Disallow negative numbers in create numeric literal (take 2)

Andarist:
reverts https://github.com/microsoft/TypeScript/pull/56422 which was a revert of https://github.com/microsoft/TypeScript/pull/55268 but this version now focuses on negative numbers alone - without making any changes to `Infinity` (and that was the only reason why the original PR was reverted), cc @jakebailey 

this would prevent issues like https://github.com/microsoft/TypeScript/issues/56562 😉


tests/cases/compiler/fakeInfinity3.ts
```diff
@@ -1,20 +1,20 @@
...
-
-    export function f(): A {
-        throw new Error()
-    }
-}
-
-export const m = X.f();
-
-export const Infinity = "oops";
+// @strict: true
```
jakebailey, line 1: What happened to these files? Some sort of line ending bug?

jakebailey:
This looks good to me besides the test diff.

IIRC our intention was to investigate actually adding an infinity/nan type, so we're leaving the status quo on `1e999` and `-1e999` for now. I believe this PR doesn't affect that.

tests/cases/compiler/fakeInfinity3.ts
Andarist: I was also wondering about it - but I didn't care as much to investigate this. I first reverted your revert (so that removed those files) and then applied your patch from https://github.com/microsoft/TypeScript/pull/56422/commits/45371af679fe83b5128fe4ee4c4b74b1ad17e7be using `git am` (that brought those files back). So maybe it somehow tripped on this combo - where the same file was both removed and added in the series of commits?

I like the approach that I used since the "commit trace" ends up being quite nice using it. The resulting diff is weird though so maybe it was not worth it. I could try to just squash all of those commits here and it's likely the problem would go away after that. Would you prefer me doing this?

Andarist:
> IIRC our intention was to investigate actually adding an infinity/nan type, so we're leaving the status quo on 1e999 and -1e999 for now. I believe this PR doesn't affect that.

I can confirm that's the intention here. Only negative finite numbers should be affected by this change, and it just forbids creating "wrong" AST nodes like that.

tests/cases/compiler/fakeInfinity3.ts
jakebailey: I suspect this is just a line endings problem, and these files can just be opened in VS Code and swapped to the other line endings (there's a button at the bottom right for it).

Andarist: You were right - this was the line-ending problem. It's kinda weird that they were changed in this whole process since I literally applied the downloaded patch directly - so I wonder at what stage this was changed... but well, the important thing is that the problem is gone now :p Thanks for the VS Code tip - I didn't know that it can convert all of the existing line endings for me.

jakebailey (approved):

Andarist:
> this matches another change made in https://github.com/microsoft/TypeScript/pull/56565 (which I am going to merge first, funnily your code would have caught this bug albeit without a test).

Yeah, that PR reminded me that I was supposed to re-propose this change :)

> Actually, one suggestion to directly negate instead of using abs;

done

src/compiler/transformers/declarations.ts
```diff
@@ -1798,7 +1798,14 @@ export function transformDeclarations(context: TransformationContext) {
                         if (shouldStripInternal(m)) return;
                         // Rewrite enum values to their constants, if available
                         const constValue = resolver.getConstantValue(m);
-                        return preserveJsDoc(factory.updateEnumMember(m, m.name, constValue !== undefined ? typeof constValue === "string" ? factory.createStringLiteral(constValue) : factory.createNumericLiteral(constValue) : undefined), m);
+                        const newInitializer = constValue === undefined
```
jakebailey, line 1801: Hate to nit one more time, but can this be reformatted into our typical switch/case form?

Andarist: hm, I don't know what you request here. Each branch~ uses a different kind of check so I can't easily use switch/case here (unless we are talking about our favorite `switch (true)` :P)

jakebailey:
```ts
const newInitializer = constValue === undefined ? undefined
    : typeof constValue === "string" ? factory.createStringLiteral(constValue)
    : constValue < 0 ? factory.createPrefixUnaryExpression(SyntaxKind.MinusToken, factory.createNumericLiteral(-constValue))
    : factory.createNumericLiteral(constValue);
```

jakebailey: Sorry, meant match/case.

jakebailey: (Or operators on the right, we're inconsistent)

Andarist: thanks for the clarification, I just pushed out the requested change

jakebailey (approved):

src/compiler/transformers/declarations.ts
jakebailey: Thanks, sorry for the churn!

PR #56580, opened 2023-11-28
Add inlay hints for IndexSignature

### LINKED ISSUE #56559: Error in 'visitForDisplayParts' during 'provideInlayHints'
This issue comes from crash dumps in telemetry. We've tried to de-duplicate issues on a best-effort basis, comparing the sequence of methods called and the command requested while ignoring line numbers.
**TypeScript version prefix**: 5.3.2
**VSCode version**: 1.84.2
**Command requested**: provideInlayHints
**Hitting sessions**: 186
**Five line hash**: bd08cd174587a4926f96e285ac9581c8
**Stack**:
```
    at visitForDisplayParts (tsserver.js:165385:17)
    at <anonymous> (tsserver.js:165393:9)
    at visitDisplayPartList (tsserver.js:165389:13)
    at visitForDisplayParts (tsserver.js:165228:13)
    at <anonymous> (tsserver.js:165393:9)
    at visitDisplayPartList (tsserver.js:165389:13)
    at visitForDisplayParts (tsserver.js:165163:13)
    at visitForDisplayParts (tsserver.js:165345:11)
    at typeToInlayHintParts (tsserver.js:165115:5)
    at visitVariableLikeDeclaration (tsserver.js:164923:23)
    at visitor (tsserver.js:164845:7)
    at visitNodes (tsserver.js:30235:22)
    at forEachChildInVariableDeclarationList (tsserver.js:30454:12)
    at forEachChild (tsserver.js:30744:35)
    at visitor (tsserver.js:164860:12)
    at visitNode2 (tsserver.js:30227:18)
    at forEachChildInVariableStatement (tsserver.js:30451:59)
    at forEachChild (tsserver.js:30744:35)
    at visitor (tsserver.js:164860:12)
    at visitNodes (tsserver.js:30235:22)
    at forEachChildInSourceFile (tsserver.js:30448:12)
    at forEachChild (tsserver.js:30744:35)
    at visitor (tsserver.js:164860:12)
    at provideInlayHints (tsserver.js:164821:3)
    at provideInlayHints2 (tsserver.js:145846:34)
    at  (tsserver.js:184199:48)
    at unknown (tsserver.js:183358:43)
    at <anonymous> (tsserver.js:185377:69)
    at  (tsserver.js:185369:14)
    at  (tsserver.js:185377:29)
    at  (tsserver.js:185419:51)
    at <anonymous> (tsserver.js:187001:14)
```

PranavSenthilnathan:
Fixes #56559


MariaSolOs (approved):

src/services/inlayHints.ts
```diff
@@ -750,6 +751,17 @@ export function provideInlayHints(context: InlayHintsContext): InlayHint[] {
                         visitForDisplayParts(node.type);
                     }
                     break;
+                case SyntaxKind.IndexSignature:
+                    Debug.assertNode(node, isIndexSignatureDeclaration);
+                    Debug.assertEqual(node.parameters.length, 1);
```
DanielRosenwasser, line 756: I'm not sure if an assertion is appropriate - what happens if you have something like

```ts
const x: { []: string }
const y = x;
```

DanielRosenwasser: Specifically, assertions shouldn't assume well-formed code, they should assert invariants of how the compiler constructs the tree. You'd probably be better off with just

```
node.parameters.forEach(visitForDisplayParts)
```

PranavSenthilnathan: I get `An index signature must have exactly one parameter.ts(1096)` for that code. Also, the parser doesn't create an index signature node for that (the type literal node just has no `member` children nodes) so this switch case isn't hit,

That being said, I also wasn't so enthusiastic about using the assert but wasn't sure about alternatives. Your suggestion is better though since this is a user error and we already report diagnostics to the user elsewhere.

I think this should be represented in the types though so we can enforce the invariant. Can we just add something like:
```ts
export interface IndexSignatureDeclaration extends SignatureDeclarationBase, ...
{
    ...
    readonly parameters: [ParameterDeclaration] & NodeArray<ParameterDeclaration>;
}
```

It's probably not that easy since this is a public API and would be breaking.

PR #56593, opened 2023-11-29
Add types for Promise.withResolvers()

### LINKED ISSUE #56483: Type definitions for `Promise.withResolvers()`
### ⚙ Compilation target

esnext

### ⚙ Library

ESNext

`Promise.withResolvers()` is missing type definitions.

It is a Stage 3 TC39 proposal that is shipping in [Chromium 119](https://chromestatus.com/feature/5810984110784512), [Firefox 121](https://bugzilla.mozilla.org/show_bug.cgi?id=1845586) and soon in other runtimes.

### Sample Code

```TypeScript
const { promise, resolve, reject } = Promise.withResolvers();
//                                           ^ Property 'withResolvers' does not exist on type 'PromiseConstructor'.(2339)
```

### Documentation Link

https://github.com/tc39/proposal-promise-with-resolvers

linusg:
Fixes #56483.


sandersn (approved):

sandersn:
Thanks for the fast turnaround @linusg

tjx666:
Why latest ts playground no this type?

https://www.typescriptlang.org/play?ts=5.4.0-dev.20231229#code/AoJw9gtglgzgpgOgO5QC4AsBKcZgDYBucIMQA

linusg:
@tjx666 the default target in the TS Playground appears to be ES2017, you have to select ESNext under the "TS Config" section to make it available.

tjx666:
@linusg @mjbvz Any way to make script in html also works?

image

src/lib/esnext.promise.d.ts
```diff
@@ -0,0 +1,17 @@
+interface PromiseWithResolvers<T> {
+    promise: Promise<T>;
+    resolve: (value: T | PromiseLike<T>) => void;
```
tophf, line 3: AFAICT `value` should be `value?`, otherwise linter complains on `resolve()` about a missing argument of type "unknown".

jakebailey: It can't be `value?` because then you could resolve _any_ promise with `undefined` accidentally by omitting the parameter.

I'm not sure how to do better other than to create two variants of `PromiseWithResolvers` and then overload `withResolvers` such that when `T` is undefined, you are allowed to have an optional parameter.

linusg: FWIW these were copied from here to match the existing resolve and reject functions passed to the Promise constructor callback:

https://github.com/microsoft/TypeScript/blob/b7f691d9f13be987e445c7878a40190ca5808691/src/lib/es2015.promise.d.ts#L13

```ts
const p = new Promise((resolve) => {
    resolve(); // Expected 1 arguments, but got 0. Did you forget to include 'void' in your type argument to 'Promise'?(2794)
});
```

The suggested fix also works here, this typechecks fine:

```ts
const { resolve } = Promise.withResolvers<void>();
resolve();
```

anion155: That's "the solution". And any other idea would be workaround. That's the official way of making `resolve` without arguments for `Promise`. You can't make `new Promise(resolve => resolve())` neither, you would need to provide `void` type parameter

tophf: I see, thanks.

FWIW, I was trying to use these definitions in plain JS, not TS. I guess I'll just edit the files locally.

PR #56594, opened 2023-11-29
Always emit `Object` type in decorator metadata for import type

### LINKED ISSUE #56530: Runtime error for emitDecoratorMetadata in an (almost) ts.transpileModule() emit
### 🔎 Search Terms

emitDecoratorMetadata, error

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about v5.2.2, v5.3.2

### ⏯ Playground Link

[Bug Workbench](https://www.typescriptlang.org/dev/bug-workbench/?importHelpers=true&experimentalDecorators=true&emitDecoratorMetadata=true&target=8&module=1&isolatedModules=true&noResolve=true&ts=5.3.2#code/PTAEAEFMA8AdIE4EsC2kB2AXAhgGwCKQDGA9gtpmQM4BQIEkKSmhp5lCAspDgCYXY6YcEiolcFSL04leAV1yRa9ESlhlMACUi54CZcJSyFkAFyhSKI+gBWBiOhIAlJeIBukGjVTqEmUADeoJgAnvCgnCEAMqL+AL6gAGYIJCigAOQAdMCh8FTpANxevMQSCJCgbtgIoLApeqHm2OghRTQwvv5EElRUESEAoopoWIE0oBB1JA0h46Ao0bEA-OaRMVSYRXFeKolIiujYaOa5Spm8mZi0PhqBoOvxSSlpWcCoKHI4AEaKhe1wt1O-QeoAAvPdYgAedByFBfRAAPjau32GCOZlA70+2B+kHOl3sHVuSCwiES2CIFQekIAKgjQDBMBheH0AIIIcghWn0gLbIA)

### 💻 Code

```ts
// @experimentalDecorators
// @emitDecoratorMetadata
// @isolatedModules
// @noResolve
// @module: commonjs

import { type MyList } from './types';

declare var property: any;

export class MyElement {
  @property
  myList?: MyList;
}

// @filename: types.d.ts
import { List } from './immutable';
export type MyList = List<number>;
```

### 🙁 Actual behavior

Using a regular emit, this produces:

```js
"use strict";
var _a;
exports.MyElement = void 0;
const tslib_1 = require("tslib");
class MyElement {
}
exports.MyElement = MyElement;
tslib_1.__decorate([
    property,
    tslib_1.__metadata("design:type", typeof (_a = typeof types_1.MyList !== "undefined" && types_1.MyList) === "function" ? _a : Object)
], MyElement.prototype, "myList", void 0);
```

Note the `types_1.MyList` reference. But `types_1` is not defined.

### 🙂 Expected behavior

Because the import for `MyList` uses `import type`, TypeScript definitely elides the import. The decorator metadata emit should therefore not check if `types_1` exists.

We're currently adopting `ts.transpileModule()`. However we're not transpiling individual files just yet. Instead we're using `ts.Program#emit` with `noResolve` and few source files for ... reasons.

- If we give TypeScript only the single file, the emit is correct: `tslib_1.__metadata("design:type", Object)`.

- If we give TypeScript all transitive dependencies, the emit is correct: `tslib_1.__metadata("design:type", Object)`

- But in this setup where TypeScript has the immediate dependency but the transitive one is missing, the emit leads to an runtime error.

I could totally understand if you don't support this kind of usage. But I'd be happy if you do :-).

frigus02:
Always emit `Object` type in legacy decorator metadata for identifiers imported with `import type`, even if the type is only partially resolved.

<details>
<summary>
Without the fix, the added test case shows the following emit:
</summary>

```js
//// [tests/cases/compiler/decoratorMetadataTypeOnlyImport.ts] ////

//// [a.ts]
import { List } from 'unknown-module';
export type MyList = List<number>;

//// [b.ts]
import { type MyList } from './a';

declare var Decorator: any;

class Foo {
  @Decorator myList?: MyList;
}


//// [a.js]
"use strict";
//// [b.js]
"use strict";
var __decorate = (this && this.__decorate) || function (decorators, target, key, desc) {
    var c = arguments.length, r = c < 3 ? target : desc === null ? desc = Object.getOwnPropertyDescriptor(target, key) : desc, d;
    if (typeof Reflect === "object" && typeof Reflect.decorate === "function") r = Reflect.decorate(decorators, target, key, desc);
    else for (var i = decorators.length - 1; i >= 0; i--) if (d = decorators[i]) r = (c < 3 ? d(r) : c > 3 ? d(target, key, r) : d(target, key)) || r;
    return c > 3 && r && Object.defineProperty(target, key, r), r;
};
var __metadata = (this && this.__metadata) || function (k, v) {
    if (typeof Reflect === "object" && typeof Reflect.metadata === "function") return Reflect.metadata(k, v);
};
var Foo = /** @class */ (function () {
    function Foo() {
    }
    var _a;
    __decorate([
        Decorator,
        __metadata("design:type", typeof (_a = typeof a_1.MyList !== "undefined" && a_1.MyList) === "function" ? _a : Object)
    ], Foo.prototype, "myList", void 0);
    return Foo;
}());
```

</details>

Fixes #56530

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -47787,28 +47787,35 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             isTypeOnly = !!rootValueSymbol?.declarations?.every(isTypeOnlyImportOrExportDeclaration);
         }
         const valueSymbol = resolveEntityName(typeName, SymbolFlags.Value, /*ignoreErrors*/ true, /*dontResolveAlias*/ true, location);
-        const resolvedSymbol = valueSymbol && valueSymbol.flags & SymbolFlags.Alias ? resolveAlias(valueSymbol) : valueSymbol;
+        const resolvedValueSymbol = valueSymbol && valueSymbol.flags & SymbolFlags.Alias ? resolveAlias(valueSymbol) : valueSymbol;
         isTypeOnly ||= !!(valueSymbol && getTypeOnlyAliasDeclaration(valueSymbol, SymbolFlags.Value));
 
         // Resolve the symbol as a type so that we can provide a more useful hint for the type serializer.
-        const typeSymbol = resolveEntityName(typeName, SymbolFlags.Type, /*ignoreErrors*/ true, /*dontResolveAlias*/ false, location);
-        if (resolvedSymbol && resolvedSymbol === typeSymbol) {
+        const typeSymbol = resolveEntityName(typeName, SymbolFlags.Type, /*ignoreErrors*/ true, /*dontResolveAlias*/ true, location);
+        const resolvedTypeSymbol = typeSymbol && typeSymbol.flags & SymbolFlags.Alias ? resolveAlias(typeSymbol) : typeSymbol;
+
+        // In case the value symbol can't be resolved (e.g. because of missing declarations), use type symbol for reachability check.
+        if (!valueSymbol) {
+            isTypeOnly ||= !!(typeSymbol && getTypeOnlyAliasDeclaration(typeSymbol, SymbolFlags.Type));
+        }
+
+        if (resolvedValueSymbol && resolvedValueSymbol === resolvedTypeSymbol) {
             const globalPromiseSymbol = getGlobalPromiseConstructorSymbol(/*reportErrors*/ false);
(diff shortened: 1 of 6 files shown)


sheetalkamat (approved):

rbuckton:
@typescript-bot pack this

rbuckton (approved):

Issue #56571, opened 2023-11-28
5.3.2 adds broken "reference" comment


rchl:
### 🔎 Search Terms

reference types error

### 🕗 Version & Regression Information

- This changed between versions 5.2.2 and 5.3.2

### 💻 Code

My code does a type import like:

```ts
import type { ModuleOptions } from '../../module/module';
```

and in 5.3.2 this adds a type reference in the compiled output:

```ts
/// <reference types="dist/module/module" />
```

which is an invalid path from the point of the user of this package and result in an error:

```
node_modules/foo-package/dist/lib/dir/file.d.ts:4:23 - error TS2688: Cannot find type definition file for 'dist/module/module'.

4 /// <reference types="dist/module/module" />
```

### 🙁 Actual behavior

Adds a broken reference:

```ts
/// <reference types="dist/module/module" />
```

### 🙂 Expected behavior

5.2.2 didn't add a reference in this case.

NullVoxPopuli:
I've noticed this, too:
https://github.com/tildeio/ember-element-helper/issues/111

mainly, here:
![image](https://github.com/microsoft/TypeScript/assets/199018/138c2d1b-dbf7-4a0f-b2bd-7dc340b3e410)

that whole reference line should not be present.
_and_ its referencing a private file that the referenced package does not expose.
the `typesVersions` of that package is:
```
  "typesVersions": {
    "*": {
      "types": [
        "types/stable"
      ],
      "types/preview": [
        "types/preview"
      ]
    }
  },
```

the main thing through, is that as a library author, I've declared `ember-source` as a peer dependency so I know that consumers will have it available and the reference isn't needed.


### AS IS

the triple slash reference behavior lends itself to libraries "breaking" during patch releases because typescript is exposing stuff it shouldn't.

rchl:
Likely caused by https://github.com/microsoft/TypeScript/issues/55636

andrewbranch:
I think you’re probably using `baseUrl` and/or `paths` and shouldn’t be. The triple-slash reference is necessary. As a library author, you’ll need to make sure you use module resolution settings that are strict and portable enough that paths in your code (whether written by you or generated, based on your settings) also resolve for users. https://www.typescriptlang.org/docs/handbook/modules/guides/choosing-compiler-options.html#im-writing-a-library

NullVoxPopuli:
@andrewbranch my tsconfig settings are already strict, and I don't use `paths` or `baseUrl`.

[...]

PR #56614, opened 2023-11-30
Don’t emit file references that are already referenced via imports

andrewbranch:
Fixes:

> Just an extra note that TS is adding:
> 
> ```ts
> /// <reference path="../../module/module.d.ts" />
> ```
> 
> even though there also exists a related type import, importing from the same file:
> 
> ```ts
> import type { ModuleOptions } from '../../module/module';
> ```
> 
> So in that case at least the reference looks redundant.

from https://github.com/microsoft/TypeScript/issues/56571#issuecomment-1830869896

### DIFF
--- a/src/compiler/transformers/declarations.ts
+++ b/src/compiler/transformers/declarations.ts
@@ -617,6 +617,11 @@ export function transformDeclarations(context: TransformationContext) {
 
         function mapReferencesIntoArray(references: FileReference[], outputFilePath: string): (file: SourceFile) => void {
             return file => {
+                if (exportedModulesFromDeclarationEmit?.includes(file.symbol)) {
+                    // Already have an import declaration resolving to this file
+                    return;
+                }
+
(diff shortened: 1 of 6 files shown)


src/compiler/transformers/declarations.ts
```diff
@@ -617,6 +617,11 @@ export function transformDeclarations(context: TransformationContext) {
 
         function mapReferencesIntoArray(references: FileReference[], outputFilePath: string): (file: SourceFile) => void {
             return file => {
+                if (exportedModulesFromDeclarationEmit?.includes(file.symbol)) {
```
andrewbranch, line 620: This is not what this array was for, and the name confuses me (these aren’t exported from this file, they’re imported 🤔), but it appears to be _exactly_ what I was going to add, so I opted not to duplicate it.

andrewbranch:

sheetalkamat (approved):

PR #56627, opened 2023-12-01
Transpile jsdoc parsing mode

sheetalkamat:
Fixes https://github.com/microsoft/TypeScript/issues/56625

### DIFF
--- a/src/services/transpile.ts
+++ b/src/services/transpile.ts
@@ -124,8 +124,6 @@ export function transpileModule(input: string, transpileOptions: TranspileOption
             setExternalModuleIndicator: getSetExternalModuleIndicator(options),
             jsDocParsingMode: transpileOptions.jsDocParsingMode !== undefined ?
                 transpileOptions.jsDocParsingMode :
-                transpileOptions.transformers ?
-                JSDocParsingMode.ParseAll :
                 JSDocParsingMode.ParseNone,
         },
     );
--- a/src/services/transpile.ts
+++ b/src/services/transpile.ts
@@ -122,9 +122,7 @@ export function transpileModule(input: string, transpileOptions: TranspileOption
             languageVersion: getEmitScriptTarget(options),
             impliedNodeFormat: getImpliedNodeFormatForFile(toPath(inputFileName, "", compilerHost.getCanonicalFileName), /*packageJsonInfoCache*/ undefined, compilerHost, options),
             setExternalModuleIndicator: getSetExternalModuleIndicator(options),
-            jsDocParsingMode: transpileOptions.jsDocParsingMode !== undefined ?
-                transpileOptions.jsDocParsingMode :
-                JSDocParsingMode.ParseNone,
+            jsDocParsingMode: transpileOptions.jsDocParsingMode ?? JSDocParsingMode.ParseNone,
         },
     );
     if (transpileOptions.moduleName) {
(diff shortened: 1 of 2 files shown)


src/services/transpile.ts
```diff
@@ -121,7 +122,11 @@ export function transpileModule(input: string, transpileOptions: TranspileOption
             languageVersion: getEmitScriptTarget(options),
             impliedNodeFormat: getImpliedNodeFormatForFile(toPath(inputFileName, "", compilerHost.getCanonicalFileName), /*packageJsonInfoCache*/ undefined, compilerHost, options),
             setExternalModuleIndicator: getSetExternalModuleIndicator(options),
-            jsDocParsingMode: JSDocParsingMode.ParseNone,
+            jsDocParsingMode: transpileOptions.jsDocParsingMode !== undefined ?
```
jakebailey, line 125: Nit, but it took me three rereads to read this code; can we write it match/case style? Or use `??` somewhere?

jakebailey:
I will be interested in a test for this; my impression was that this flag did not affect emit at all, and #56625 did not mention transformers in the original report.

src/services/transpile.ts
```diff
@@ -121,7 +122,11 @@ export function transpileModule(input: string, transpileOptions: TranspileOption
             languageVersion: getEmitScriptTarget(options),
             impliedNodeFormat: getImpliedNodeFormatForFile(toPath(inputFileName, "", compilerHost.getCanonicalFileName), /*packageJsonInfoCache*/ undefined, compilerHost, options),
             setExternalModuleIndicator: getSetExternalModuleIndicator(options),
-            jsDocParsingMode: JSDocParsingMode.ParseNone,
+            jsDocParsingMode: transpileOptions.jsDocParsingMode !== undefined ?
+                transpileOptions.jsDocParsingMode :
+                transpileOptions.transformers ?
+                JSDocParsingMode.ParseAll :
```
andrewbranch, line 128: I think this is a somewhat unintuitive dependent default. I would personally opt for the simplicity of just having the default be `ParseNone`.

sheetalkamat:
```suggestion
```

src/services/transpile.ts
```diff
@@ -121,7 +122,9 @@ export function transpileModule(input: string, transpileOptions: TranspileOption
             languageVersion: getEmitScriptTarget(options),
             impliedNodeFormat: getImpliedNodeFormatForFile(toPath(inputFileName, "", compilerHost.getCanonicalFileName), /*packageJsonInfoCache*/ undefined, compilerHost, options),
             setExternalModuleIndicator: getSetExternalModuleIndicator(options),
-            jsDocParsingMode: JSDocParsingMode.ParseNone,
+            jsDocParsingMode: transpileOptions.jsDocParsingMode !== undefined ?
+                transpileOptions.jsDocParsingMode :
+                JSDocParsingMode.ParseNone,
```
sheetalkamat, line 127:
```suggestion
            jsDocParsingMode: transpileOptions.jsDocParsingMode ?? JSDocParsingMode.ParseNone,
```

andrewbranch (approved):

andrewbranch:

src/services/transpile.ts
jakebailey: To be honest, I would have made this default to `ParseAll`; when I wrote this I had forgotten about transformers and assumed that this was pure string to string without them.

PR #56637, opened 2023-12-01
Allow function declarations with members to be typed using type tag in JSDoc

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/56615

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -10981,7 +10981,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     return getReturnTypeOfSignature(getterSignature);
                 }
             }
-            const parameterTypeOfTypeTag = getParameterTypeOfTypeTag(func, declaration);
+            const parameterTypeOfTypeTag = getParameterTypeOfTypeTag(func, declaration, /*allowMembers*/ true);
             if (parameterTypeOfTypeTag) return parameterTypeOfTypeTag;
             // Use contextual parameter type if one is available
             const type = declaration.symbol.escapedName === InternalSymbolName.This ? getContextualThisParameterType(func) : getContextuallyTypedParameterType(declaration);
@@ -15201,15 +15201,15 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         return true;
     }
 
-    function getSignatureOfTypeTag(node: SignatureDeclaration | JSDocSignature) {
+    function getSignatureOfTypeTag(node: SignatureDeclaration | JSDocSignature, allowMembers?: boolean) {
         // should be attached to a function declaration or expression
         if (!(isInJSFile(node) && isFunctionLikeDeclaration(node))) return undefined;
         const typeTag = getJSDocTypeTag(node);
-        return typeTag?.typeExpression && getSingleCallSignature(getTypeFromTypeNode(typeTag.typeExpression));
+        return typeTag?.typeExpression && getSingleCallSignature(getTypeFromTypeNode(typeTag.typeExpression), allowMembers);
     }
 
-    function getParameterTypeOfTypeTag(func: FunctionLikeDeclaration, parameter: ParameterDeclaration) {
-        const signature = getSignatureOfTypeTag(func);
+    function getParameterTypeOfTypeTag(func: FunctionLikeDeclaration, parameter: ParameterDeclaration, allowMembers?: boolean) {
+        const signature = getSignatureOfTypeTag(func, allowMembers);
         if (!signature) return undefined;
         const pos = func.parameters.indexOf(parameter);
         return parameter.dotDotDotToken ? getRestTypeAtPosition(signature, pos) : getTypeAtPosition(signature, pos);
@@ -33486,9 +33486,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             (typeArguments.length >= minTypeArgumentCount && typeArguments.length <= numTypeParameters);
     }
 
-    // If type has a single call signature and no other members, return that signature. Otherwise, return undefined.
-    function getSingleCallSignature(type: Type): Signature | undefined {
-        return getSingleSignature(type, SignatureKind.Call, /*allowMembers*/ false);
+    // By default if type has a single call signature and no other members, return that signature. Otherwise, return undefined.
+    function getSingleCallSignature(type: Type, allowMembers = false): Signature | undefined {
+        return getSingleSignature(type, SignatureKind.Call, allowMembers);
     }
 
     function getSingleCallOrConstructSignature(type: Type): Signature | undefined {
--- a/tests/baselines/reference/typeTagOnFunctionDeclaration.errors.txt
+++ b/tests/baselines/reference/typeTagOnFunctionDeclaration.errors.txt
@@ -1,0 +1,21 @@
+index.js(15,1): error TS2322: Type 'number' is not assignable to type 'string'.
+
+
+==== index.js (1 errors) ====
+    /**
+     * @type {{(arg1: string): string;}}
+    */
+    function fn(arg1) {
+        return arg1;
+    }
+    
+    /**
+     * @type {{(arg1: string): string; bar: string}}
+    */
+    function fn2(arg1) {
+        return arg1;
+    }
+    
+    fn2.bar = 42; // error
+    ~~~~~~~
+!!! error TS2322: Type 'number' is not assignable to type 'string'.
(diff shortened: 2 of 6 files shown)


src/compiler/checker.ts
```diff
@@ -15201,15 +15201,15 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         return true;
     }
 
-    function getSignatureOfTypeTag(node: SignatureDeclaration | JSDocSignature) {
+    function getSignatureOfTypeTag(node: SignatureDeclaration | JSDocSignature, allowMembers?: boolean) {
```
sandersn, line 15204: how many callers are there of these two functions? Could allowMembers be required without changing tonnes of call sites?

src/compiler/checker.ts
```diff
@@ -33486,9 +33486,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             (typeArguments.length >= minTypeArgumentCount && typeArguments.length <= numTypeParameters);
     }
 
-    // If type has a single call signature and no other members, return that signature. Otherwise, return undefined.
-    function getSingleCallSignature(type: Type): Signature | undefined {
-        return getSingleSignature(type, SignatureKind.Call, /*allowMembers*/ false);
+    // By default if type has a single call signature and no other members, return that signature. Otherwise, return undefined.
+    function getSingleCallSignature(type: Type, allowMembers = false): Signature | undefined {
```
sandersn, line 33490: same question here, although it likely has way more calls than the other two functions

sandersn (changes requested):
I think this is basically fine, I just prefer to avoid optional parameters if possible because they're too easy to forget about when just looking at call sites.

Andarist:
I made those new `allowMembers` required. By looking at all the call sites that I had to adjust, I think it's somewhat likely that some of them should also allow those members. I might try to find some time to investigate them if that would be required to merge this - but, at the moment, I have a long tail of reviews to address so I would kind of prefer not to do this.

src/compiler/checker.ts
```diff
@@ -14965,7 +14965,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             result = appendIfUnique(result, getDeclaredTypeOfTypeParameter(node.symbol));
         }
         return result?.length ? result
-            : isFunctionDeclaration(declaration) ? getSignatureOfTypeTag(declaration)?.typeParameters
+            : isFunctionDeclaration(declaration) ? getSignatureOfTypeTag(declaration, /*allowMembers*/ false)?.typeParameters
```
Andarist, line 14968:
```ts
/** @type {{(arg1: string): void;}} */
function Named() {
	this.x = 1;
};
Named.foo = 10;
```

This is currently treated as a JS class and the extra expando property is allowed too. The generated DTS file looks like this:
```ts
declare function Named(arg1: string): void;
declare class Named {
    x: number;
}
declare namespace Named {
    let foo: number;
}
```

So... it would potentially also make sense to allow members here.

src/compiler/checker.ts
```diff
@@ -10981,7 +10981,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     return getReturnTypeOfSignature(getterSignature);
                 }
             }
-            const parameterTypeOfTypeTag = getParameterTypeOfTypeTag(func, declaration);
+            const parameterTypeOfTypeTag = getParameterTypeOfTypeTag(func, declaration, /*allowMembers*/ true);
```
Andarist, line 10984: This is the call that makes the original case work.

src/compiler/checker.ts
```diff
@@ -15201,22 +15201,22 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+    function getParameterTypeOfTypeTag(func: FunctionLikeDeclaration, parameter: ParameterDeclaration, allowMembers: boolean) {
+        const signature = getSignatureOfTypeTag(func, allowMembers);
         if (!signature) return undefined;
         const pos = func.parameters.indexOf(parameter);
         return parameter.dotDotDotToken ? getRestTypeAtPosition(signature, pos) : getTypeAtPosition(signature, pos);
     }
 
     function getReturnTypeOfTypeTag(node: SignatureDeclaration | JSDocSignature) {
-        const signature = getSignatureOfTypeTag(node);
+        const signature = getSignatureOfTypeTag(node, /*allowMembers*/ false);
```
Andarist, line 15219: The only function that calls `getReturnTypeOfTypeTag` is `getReturnTypeFromAnnotation`. That one uses `getEffectiveReturnTypeNode` and roughly returns that as is if it's available. An explicit annotated return type might appear in a function with members - so members are kinda allowed in that code path. So I think they should be allowed here too.

src/compiler/checker.ts
```diff
@@ -15336,7 +15336,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 const type = signature.declaration && getEffectiveReturnTypeNode(signature.declaration);
                 let jsdocPredicate: TypePredicate | undefined;
                 if (!type) {
-                    const jsdocSignature = getSignatureOfTypeTag(signature.declaration!);
+                    const jsdocSignature = getSignatureOfTypeTag(signature.declaration!, /*allowMembers*/ false);
```
Andarist, line 15339: I can't figure out a test case for which this would matter but I don't see a reason why this would have to **not** allow members.

src/compiler/checker.ts
```diff
@@ -29493,7 +29493,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         if (thisTag && thisTag.typeExpression) {
             return getTypeFromTypeNode(thisTag.typeExpression);
         }
-        const signature = getSignatureOfTypeTag(node);
+        const signature = getSignatureOfTypeTag(node, /*allowMembers*/ false);
```
Andarist, line 29496: this should allow members to make this work:
```ts
/** @type {{(this: { name: string }, arg: boolean): void; bar: string}} */
function fn(arg) {
    this
}

fn.bar = ''
```

src/compiler/checker.ts
```diff
@@ -30954,7 +30954,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     // union type of return types from these signatures
     function getContextualSignature(node: FunctionExpression | ArrowFunction | MethodDeclaration): Signature | undefined {
         Debug.assert(node.kind !== SyntaxKind.MethodDeclaration || isObjectLiteralMethod(node));
-        const typeTagSignature = getSignatureOfTypeTag(node);
+        const typeTagSignature = getSignatureOfTypeTag(node, /*allowMembers*/ false);
```
Andarist, line 30957: I came to a realization that this is the only call to `getSignatureOfTypeTag` that has to disallow members. However, I also find this call being here slightly off. A signature annotated through a type tag is still an annotated signature in my eyes and not a contextual signature. 

I think there is a chance that resolving this could be lifted and then `getSignatureOfTypeTag` could just **always** allow members. 

We could also allow members here too. It should error on the lack of those extra members but the param annotations could be taken into consideration.

tests/baselines/reference/typeTagOnFunctionDeclaration2.js
```diff
@@ -0,0 +1,54 @@
...
+
+
+//// [index.d.ts]
+export function test1(): void;
+export namespace test1 {
+    let foo: string;
+}
+export function test2(): void;
+export namespace test2 {
+    let foo_1: any;
```
Andarist, line 52: this definitely shouldn't be `any`

tests/baselines/reference/typeTagOnFunctionDeclaration3.js
```diff
@@ -0,0 +1,62 @@
...
+
+
+//// [index.d.ts]
+export function test1(): void;
+export namespace test1 {
+    const foo: string;
+}
+export function test2(): void;
+export namespace test2 {
+    const foo_1: any;
```
Andarist, line 60: similarly, this definitely shouldn't be `any`

src/compiler/checker.ts
sandersn: 1. If this is the only place that disallows members, I do think it would be better to get the signature and then disallow them here.
2. I don't remember the situation precisely, I just remember that treating a type tag as context was a hack. It would be nice not to have it.

src/compiler/checker.ts
Andarist: This is already using `allowMember: true` with the new code - and we need that because otherwise the signature would not be retrieved at all (and thus its type parameters wouldn't be retrieved either) when `declaration` has members

src/compiler/checker.ts
Andarist: 1. I now allow members here. The "contextual" parameter types are assigned based on the type tag with members but it errors later because inline function expressions etc can't have members. I added a test case to this PR that verifies this behavior
2. Roget that. I might take a look at this sometime in the future but it's not relevant to this PR so I'd have to tackle it separately.

PR #56646, opened 2023-12-02
GeneratorFunction should not be newable

### LINKED ISSUE #56645: GeneratorFunction should not be newable
### ⚙ Compilation target

ES2015 ES2018

### ⚙ Library

generator asyncgenerator

`GeneratorFunction` and `AsyncGeneratorFunction` (`async function * () {}`) should not be newable.

`GeneratorFunction` means `function * () {}` which is only callable, while `GeneratorFunctionConstructor` means `function * () {}.constructor` which is both callable and newable.

### Sample Code

```TypeScript
declare const gen: GeneratorFunction;
new gen;// there should cause a type error, but not
```

### Documentation Link

_No response_

LongTengDao:
Fixes #56645


RyanCavanaugh:
@typescript-bot test it

PR #56649, opened 2023-12-02
feat(56634): Support for "const modifier on type parameters" with JSDoc

### LINKED ISSUE #56634: Support for "`const` modifier on type parameters" with JSDoc
### 🔍 Search Terms

"const type parameter jsdoc", "const generic jsdoc", "const template jsdoc"

### ✅ Viability Checklist

### ⭐ Suggestion

v5.0 added support for ["`const` modifier on type parameters"](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#const-type-parameters), but it is not supported with JSDoc [`@template`](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#template) tag (or at least I couldn't make it work nor find any documentation about it).

I propose adding support for the `const` prefix in the `@template`:
```javascript
/**
* @template {const Array<any>} T
*/
```

### 📃 Motivating Example

The same as "`const` modifier on type parameters" itself..

### 💻 Use Cases

...
...
I have to wrap the generic value with `/** @type {const} */(...)` every time I'm using the function

a-tarasyuk:
Fixes #56634


src/compiler/parser.ts
```diff
@@ -9711,8 +9712,13 @@ namespace Parser {
                 if (isBracketed) {
                     skipWhitespace();
                 }
-                const name = parseJSDocIdentifierName(Diagnostics.Unexpected_token_A_type_parameter_name_was_expected_without_curly_braces);
 
+                let modifiers: Modifier[] | undefined;
+                if (token() === SyntaxKind.ConstKeyword) {
+                    modifiers = append(modifiers, parseTokenNode<ConstKeyword>());
+                }
```
Andarist, line 9719: Shouldn't this do the same thing that `parseTypeParameter` already does?
```suggestion
                const modifiers = parseModifiers(/*allowDecorators*/ false, /*permitConstAsModifier*/ true);
```

The advantage of this would likely be that you'd also handle `in` and `out` modifiers.

Andarist: It would also allow parsing "incorrect" modifiers - but TS already allows this in `.ts` files. The diagnostics about incorrect modifier usage are raised later - but the parsing alone succeeds with incorrect ones. Like for example:
```ts
function test<private T>() {} // 'private' modifier cannot appear on a type parameter(1273)
```

sandersn (changes requested):
requesting changes based on @Andarist 's comments: parsing should ideally use parseModifiers and then forbid unsupported modifiers in the checker.

a-tarasyuk:
@sandersn @Andarist Thanks for the review. @sandersn Do I need to add additional tests for disallowed modifiers? If so, which ones?"

Andarist:
I think that the test you added for the disallowed `private` is already enough to cover the disallowed modifiers.

I would add some related to `in` and `out` though to ensure that those work as expected in JSDoc.

Andarist:
LGTM :)

src/compiler/checker.ts
```diff
@@ -48786,7 +48786,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                         const parent = node.parent;
                         if (
                             node.kind === SyntaxKind.TypeParameter && !(isFunctionLikeDeclaration(parent) || isClassLike(parent) || isFunctionTypeNode(parent) ||
-                                isConstructorTypeNode(parent) || isCallSignatureDeclaration(parent) || isConstructSignatureDeclaration(parent) || isMethodSignature(parent))
+                                isConstructorTypeNode(parent) || isCallSignatureDeclaration(parent) || isConstructSignatureDeclaration(parent) || isMethodSignature(parent) || isJSDocTemplateTag(parent))
```
sandersn, line 48789: typescript disallows `type X<const T> = [T]` but I don't think this disallows the equivalent

```js
/**
 * @template const T
 * @typedef {[T]} X
 */
```

src/compiler/checker.ts
```diff
@@ -49072,7 +49072,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     case SyntaxKind.OutKeyword:
                         const inOutFlag = modifier.kind === SyntaxKind.InKeyword ? ModifierFlags.In : ModifierFlags.Out;
                         const inOutText = modifier.kind === SyntaxKind.InKeyword ? "in" : "out";
-                        if (node.kind !== SyntaxKind.TypeParameter || !(isInterfaceDeclaration(node.parent) || isClassLike(node.parent) || isTypeAliasDeclaration(node.parent))) {
+                        if (node.kind !== SyntaxKind.TypeParameter || !(isInterfaceDeclaration(node.parent) || isClassLike(node.parent) || isTypeAliasDeclaration(node.parent) || isJSDocTemplateTag(node.parent))) {
```
sandersn, line 49075: typescript disallows `function f<in T>(t: T) {}` but I don't think this disallows the equivalent

```js
/**
 * @template in T
 */
function f(t: T) {}
```

sandersn (changes requested):

sandersn (approved):

PR #52807, opened 2023-02-16
Stricter relational comparisons checking `valueOf`

RyanCavanaugh:
Per discussion in #52790, this implements the following ruleset for `>`, `<`, `>=`, and `<=`:

For each operand expressions, let the "operand type" be:
* The base primitive type of expression's type, if one exists
* Otherwise, the return type of the `valueOf` property of the expression's type, if that exists
* Otherwise, the expression type

One of the following must be true:
* Either operand type is any
* Both operand types are assignable to string
* Both operand types are assignable to number

Otherwise an error is issued

PR #27926, opened 2018-10-16
Flag non-any/number/string operands on either side of comparison as errors

feloy:
Fixes #15506

Issue #15506, opened 2017-05-01
Flag boolean in comparison operator operands as errors


mhegazy:
From https://github.com/Microsoft/TypeScript/issues/15444#issuecomment-298055832

```ts
var x: boolean, y: boolean;
if (x > y) {  // suspicious!
}
```

dmichon-msft:
Well, that is the most concise way of writing:
```ts
let x: boolean, y: boolean;
if (x && !y) {
}
```

RyanCavanaugh:
Accepting PRs: Disallow non-`any`/`number`/`string` operands on either side of `<`, `>`, etc.. Should be an easy change

Kovensky:
Should `{}` also be comparable?

At least I have used it in the past to indicate a "comparable" that was not `any` (i.e. couldn't do anything with it other than compare; using `any` would allow property access and other shenanigans).

DanielRosenwasser:
> Should `{}` also be comparable?

Could you elaborate on the scenario? What was the run-time value supposed to be?

Kovensky:
In this specific case, it was a function that received a function that produces an ordered list of sorting keys; where the value of the sorting keys are anything comparable. At the time that included both `number`s and `Date` objects.

Speaking of which, are `Date` objects comparable because of spec magic, or because of `Symbol.toPrimitive` 🤔

RyanCavanaugh:
> Speaking of which, are Date objects comparable because of spec magic

*Extreme* spec magic; see https://github.com/Microsoft/TypeScript/issues/2361#issuecomment-82509802

FabianLauer:
I've just taken a stab at this issue. The conformance tests for comparison operators work as expected. However, a few tests still fail, such as:

```
    async function foo(await): Promise<void> {
                       ~~~~~
!!! error TS1138: Parameter declaration expected.
                       ~~~~~
!!! error TS2304: Cannot find name 'await'.
                            ~
!!! error TS1005: ';' expected.
                             ~
!!! error TS1128: Declaration or statement expected.
                               ~~~~~~~~~~~~~~~
                                       ~~~~
!!! error TS2532: Object is possibly 'undefined'.
                                           ~
!!! error TS1109: Expression expected.
    }
    ~
!!! error TS2365: Operator '>' cannot be applied to types 'boolean' and '{}'.
```
... or ...
```
    new <T>Foo()
    ~~~~~~~~~~~~
[...]

PR #56666, opened 2023-12-04
Fix(15506) - Disallow non-any/number/string operands on comparisons

c0sta:
Fixes #

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -20035,6 +20035,21 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         return isTypeComparableTo(type1, type2) || isTypeComparableTo(type2, type1);
     }
 
+    function checkTypeComparableWithLessOrGreaterThanOperator(valueType: Type): boolean {
+        const t = valueType.flags;
+        return !!(
+            (t & TypeFlags.Number ||
+                t & TypeFlags.NumberLiteral ||
+                t & TypeFlags.BigInt ||
+                t & TypeFlags.BigIntLiteral ||
+                t & TypeFlags.String ||
+                t & TypeFlags.StringLiteral ||
+                t & TypeFlags.Any) &&
+            !(t & TypeFlags.Boolean) &&
+            !(t & TypeFlags.Object)
+        );
+    }
+
     function checkTypeAssignableTo(source: Type, target: Type, errorNode: Node | undefined, headMessage?: DiagnosticMessage, containingMessageChain?: () => DiagnosticMessageChain | undefined, errorOutputObject?: { errors?: Diagnostic[]; }): boolean {
         return checkTypeRelatedTo(source, target, assignableRelation, errorNode, headMessage, containingMessageChain, errorOutputObject);
     }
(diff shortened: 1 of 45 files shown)


c0sta:
> @c0sta please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

c0sta:
This last commit fixes the Self Check action error by allowing compare `Objects`, `Union`, `TypeParameter`, and `Intersection`. That's because TS has some internal comparisons that use these types (already discussed [here](https://github.com/microsoft/TypeScript/pull/27926#discussion_r225654629) in the past). But I'm unsure if that's a good idea since it goes against the main issue purpose to disallow non-`any`, `number`, and `string` operands.

jakebailey:
> This last commit fixes the Self Check action error by allowing compare `Objects`, `Union`, `TypeParameter`, and `Intersection`. That's because TS has some internal comparisons that use these types (already discussed [here](https://github.com/microsoft/TypeScript/pull/27926#discussion_r225681970) in the past). But I'm unsure if that's a good idea since it appears to go against the main issue purpose to disallow non-`any`, `number`, and `string` operands.

I would expect that we would be recursing into those types to verify, and not ignore them. I'm pretty sure we have this sort of checking for other operators, so it feels weird that we are adding a whole new function to check this...

c0sta:
Just some updates, I've searched for checkers that could make this comparison but I couldn't find it, so basically what I did was start using this `isTypeAssignableTo`. I'm just struggling a little bit to trigger the error on `never and number` comparisons but I'm working on this.

The Self Check error will come back now and after fixing the `never and number` comparison I'll look at them.

This last push brings a lot from the main branch with it, not sure if I'm doing something wrong here.

jakebailey:
It seems like you incorrectly merged main, I'm afraid.

c0sta:
> It seems like you incorrectly merged main, I'm afraid.

Sorry about that, I did a reset on my [875169e](https://github.com/microsoft/TypeScript/pull/56666/commits/875169efb59de481c3b7159486c9a49ab6aeea01) commit, but if you think it's a better idea to open another PR I can do that too.
I changed my configs to use merge instead of rebase, it seems to be merging properly now.

c0sta:
So I've worked to fix some of the [self-check errors](https://github.com/microsoft/TypeScript/actions/runs/7408902335/job/20158062843?pr=56666), most of them were related to Date comparisons so I started using `.getTime()` to convert the Date objects to numbers. 

But I'm not sure how to go on this last error: ` Operator '<' cannot be applied to types 'string | number' and 'string | number'.`. Since `string` and `number` are valid types for this operand, should this union keep triggering an error? and should it be possible to compare `string` with `number`?

jakebailey:
How is this PR different from #52807?

`compareComparableValues` is also flagged by that PR for the same reason; it's unsafe but cannot actually be safe.

c0sta:
> How is this PR different from #52807?
> 
> `compareComparableValues` is also flagged by that PR for the same reason; it's unsafe but cannot actually be safe.

I don't think they are, they look the same to me. In that case, it's a better idea to just close this current PR?

jakebailey:
Maybe; #52807 did not mention #15506 but was instead in the context of better handling `valueOf` which I don't think this PR does (but likely should, as it is part of a way to make it less breaky). I don't recall where we were with #52807 but @RyanCavanaugh may remember. I'm personally still hoping that's merged.

PR #56684, opened 2023-12-06
Eopt list

### LINKED ISSUE #56682: Confusing diagnostics with exact optional property types (EOPT)
### 🔎 Search Terms

EOPT
exact optional property types
exactOptionalPropertyTypes

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about EOPT

### ⏯ Playground Link

https://www.typescriptlang.org/play?exactOptionalPropertyTypes=true&ts=5.4.0-dev.20231205#code/PTAEAEGcBcCcEsDG0Bco4FcCmAoEEsAPAQ2QHkAHaeAewDtiAbABVhoq1mgE8AVbjpDSZcOHh1ABBUAF5QAb1DEA-GjoYAtgCNOAblBa0MBHQDmAGlAAvVQoC+oO7pw4AZhjrJadUK4AUxGiSAJQKoLBY0BiwPsT6di7+ioGgHgAmWK7wdFhpllZo6ZnZuY7BzviSsKaaWHTQoDSu6AJYoADkyYV0GVk5afHtoPCQoHQ0DcSQkPCmDFqMbdA0oBTEsMQakZyNzeJt7ZJDAO7w0AAWHUSk0JTU9Eys7Jw8-ILCsNjtAHSgAML0GYZWBKNJpbKmDpFPq5IbLdDnJatUZNBFLdamSLtUYUNgcLjwLCQb54MCgJ74ngdLRDEagDQjGZmYY+fYdLqpHrFfqDAwYBoRACOGHgETSLJaEkOPz8ACYAMwAdgAnMEcEA

### 💻 Code

```ts
// @strict: true
// @exactOptionalPropertyTypes: true

type A = { a?: number; b: string, z?: {} };

function f(a: A) { return a; }

f({ a: undefined, z: undefined });
// Argument of type '{ a: undefined; }' is not assignable to parameter of type 'A' with 'exactOptionalPropertyTypes: true'. Consider adding 'undefined' to the types of the target's properties.
//  Property 'b' is missing in type '{ a: undefined; }' but required in type 'A'.(2379)

```

### 🙁 Actual behavior

Normally nested diagnostic messages are a refinement of their parents.  In this case the inner diagnostic about property `b`  is not related to the EOPT assignability issue in the head message.  We have two different errors being represented in the same chain.

### 🙂 Expected behavior

There are several reasonable behaviors here.

1. Do not change the parent message to the EOPT flavor when the inner message is not EOPT related
2. Report two separate diagnostics
3. Display both the current inner message and the EOPT mismatches at the same level

I would prefer option 2.  
I have a working version of option 3 that I can a PR for if you want to go that route ( either as a permanent solution or a stop gap until further work on EOPT is done)

```
Argument of type '{ a: undefined; z: undefined; }' is not assignable to parameter of type 'A'.
  Properties not assignable with exactOptionalPropertyTypes: a, z
  Property 'b' is missing in type '{ a: undefined; z: undefined; }' but required in type 'A'.ts(2345)
```

_No response_

typeholes:
Fixes #56682

Implements option 3 - reporting the EOPT mismatches as an entry in the diagnostic message chain at the same level

If this moves forward, should we remove the EOPT flavored diagnostic messages as the flavoring is redundant with the new chain entries?


typeholes:

tests/baselines/reference/strictOptionalProperties1.errors.txt
```diff
@@ -15,6 +15,7 @@ strictOptionalProperties1.ts(75,5): error TS2322: Type '[number, undefined, true
   Type at position 1 in source is not compatible with type at position 1 in target.
     Type 'undefined' is not assignable to type 'string'.
 strictOptionalProperties1.ts(99,7): error TS2375: Type '{ foo: undefined; bar: string; }' is not assignable to type 'InputProps' with 'exactOptionalPropertyTypes: true'. Consider adding 'undefined' to the types of the target's properties.
+  Properties not assignable with exactOptionalPropertyTypes: foo
   Types of property 'foo' are incompatible.
```
DanielRosenwasser, line 19: To be honest, this just seems confusing. I don't like the idea of adding an equal-level diagnostic when we can avoid it. Do you already know that the incompatibility is from `exactOptionalPropertyTypes`? If so, just don't report the other error.

typeholes: Thanks for looing. I agree that it is superfluous when the child is an EOPT incompatibility.  The issue is that when the child is not an EOPT incompatibility the parent getting the EOPT flavor implies that it is and does not provide any detail on the reason it is marked as EOPT.

Would this be more palatable if the list of properties were added to the parent message instead?  For example:

```
Argument of type '{ a: undefined; z: undefined; }' is not assignable to parameter of type 'A' with 'exactOptionalPropertyTypes: true'. Consider adding 'undefined' to the types of the target's properties: a, z
   Property 'b' is missing in type '{ a: undefined; }' but required in type 'A'.(2379)
```

I could also try to split this into two separate errors if that would be preferable.

The other options provide less information and seem more difficult to implement:

1. ensure the reported child is an EOPT incompatibility if there is one
   a. losing the non EOPT incompatibility in the error makes it harder when the fix is that the incorrect target type was specified
2. Do not add the EOPT flavor when the reported child is not an EOPT incompatibility
   b. I would prefer this to 1, but in is inconvenient to fix an error only to have another take its place.

src/compiler/diagnosticMessages.json
```diff
@@ -4164,7 +4160,6 @@
         "category": "Error",
         "code": 4124
     },
-
```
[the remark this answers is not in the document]
typeholes: Sorry about that, newlines have been restored and upstream main merged in

PR #56626, opened 2023-11-30
Directly copy only the index signature in addition to declared properties

RyanCavanaugh:
Fixes #56620

Partial revert of #55252

PR #56690, opened 2023-12-06
fix a bug preventing a subclass property sharing a name with a type parameter from being accessed

### LINKED ISSUE #56689: Subclass property sharing name with type parameter can't be resolved
### 🔎 Search Terms

subclass property type parameter resolution 2339

### ⏯ Playground Link

https://tsplay.dev/wEYkOw

### 💻 Code

```ts
class Leg {}

class Foo<t> extends Leg {
    t = {} as t

    // should allow this access since t was declared as a property on Foo
    foo = this.t 
}
```

### 🙁 Actual behavior

Error on accessing `t`: `Property 't' does not exist on type 'Foo<t>'.ts(2339)`

### 🙂 Expected behavior

Should allow `t` to be accessed.

_No response_

ssalbdivad:
Fixes #56689

I have very little context here so happy to adjust if there is a better way to resolve this. It just looked like it would be straightforward to resolve when I encountered so decided to give it a shot 🙏

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -13062,7 +13062,8 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 const symbolTable = createSymbolTable();
                 // copy all symbols (except type parameters), including the ones with internal names like `InternalSymbolName.Index`
                 for (const symbol of members.values()) {
-                    if (!(symbol.flags & SymbolFlags.TypeParameter)) {
+                    // include declared properties that share a name with a type parameter by checking for multiple declarations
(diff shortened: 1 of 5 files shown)


ssalbdivad:

MartinJohns:
The same issue is fixed by #56626. @RyanCavanaugh You might want to compare this solution with yours.

fatcerberus:
> Why would I want to do that, to make myself feel bad?

He meant that *Ryan* should compare the solution in #56626 to this one (to see which one is better), I think.

ssalbdivad:
@fatcerberus Ahh, well that would be a very generous appraisal of the likelihood that I have something to contribute.

PR #51787, opened 2022-12-06
Bump LKG and package.json to 5.0.0-dev.20230112

jakebailey:
- Updates `package.json`/`node_modules` to `5.0.0-dev.20230112` with an override to force its usage (even though it's a major version bump).
- Bumps LKG to this nightly to get #51152, #51771, #52123, #52180 for use in #49929.
- Adds a script (`hereby bump-lkg-to-nightly`) which automates bumping `package.json` and LKG in lockstep, until we eventually drop LKG on main.

This PR looks huge because #51699 effectively completely rewrote `tsserver.js`.

PR #56701, opened 2023-12-06
Show implied options in --showConfig

andrewbranch:
Fixes my minor annoyance from earlier today

`--showConfig` on this tsconfig:

```json
{
  "compilerOptions": {
    "module": "NodeNext"
  }
}
```

Before:

```json
{
    "compilerOptions": {
        "module": "nodenext"
    },
    "files": [
        "./main.ts",
    ]
}
```

After:

```json
{
    "compilerOptions": {
        "module": "nodenext",
        "target": "esnext",
        "moduleResolution": "nodenext",
        "moduleDetection": "force",
        "esModuleInterop": true,
        "useDefineForClassFields": true
    },
    "files": [
        "./main.ts"
    ]
}
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,7 +78,7 @@
         "playwright": "^1.38.0",
         "source-map-support": "^0.5.21",
         "tslib": "^2.5.0",
-        "typescript": "^5.3.2",
+        "typescript": "^5.4.0-dev.20231206",
         "which": "^2.0.2"
     },
     "overrides": {
--- a/package.json
+++ b/package.json
@@ -78,7 +78,7 @@
         "playwright": "^1.38.0",
         "source-map-support": "^0.5.21",
         "tslib": "^2.5.0",
-        "typescript": "^5.4.0-dev.20231206",
+        "typescript": "5.4.0-dev.20231206",
         "which": "^2.0.2"
     },
     "overrides": {
--- a/src/compiler/utilities.ts
+++ b/src/compiler/utilities.ts
@@ -8587,140 +8587,233 @@ export function getSetExternalModuleIndicator(options: CompilerOptions): (file: 
     }
 }
 
-/** @internal */
-export interface ComputedCompilerOption<N extends keyof CompilerOptions, K extends keyof CompilerOptions, R extends CompilerOptions[N]> {
-    name: N;
-    dependencies: readonly K[];
-    computeValue: (compilerOptions: Pick<CompilerOptions, N | K>) => R;
+type CompilerOptionKeys = keyof { [K in keyof CompilerOptions as string extends K ? never : K]: any; };
+function createComputedCompilerOptions<T extends Record<string, CompilerOptionKeys[]>>(
+    options: {
+        [K in keyof T & CompilerOptionKeys]: {
+            dependencies: T[K];
+            computeValue: (compilerOptions: Pick<CompilerOptions, K | T[K][number]>) => Exclude<CompilerOptions[K], undefined>;
+        };
+    },
+) {
+    return options;
 }
 
 /** @internal */
-function computedCompilerOption<N extends keyof CompilerOptions, K extends keyof CompilerOptions, R extends CompilerOptions[N]>(
-    name: N,
-    dependencies: readonly K[],
-    computeValue: (compilerOptions: Pick<CompilerOptions, N | K>) => R,
-): ComputedCompilerOption<N, K, R> {
-    return { name, dependencies, computeValue };
-}
-
-/** @internal */
-export const computedOptions = {
-    target: computedCompilerOption("target", ["module"], compilerOptions => {
-        return compilerOptions.target ??
-            ((compilerOptions.module === ModuleKind.Node16 && ScriptTarget.ES2022) ||
-                (compilerOptions.module === ModuleKind.NodeNext && ScriptTarget.ESNext) ||
-                ScriptTarget.ES5);
-    }),
-    module: computedCompilerOption("module", ["target"], (compilerOptions): ModuleKind => {
-        return typeof compilerOptions.module === "number" ?
-            compilerOptions.module :
-            computedOptions.target.computeValue(compilerOptions) >= ScriptTarget.ES2015 ? ModuleKind.ES2015 : ModuleKind.CommonJS;
-    }),
-    moduleResolution: computedCompilerOption("moduleResolution", ["module", "target"], (compilerOptions): ModuleResolutionKind => {
-        let moduleResolution = compilerOptions.moduleResolution;
-        if (moduleResolution === undefined) {
-            switch (computedOptions.module.computeValue(compilerOptions)) {
-                case ModuleKind.CommonJS:
-                    moduleResolution = ModuleResolutionKind.Node10;
-                    break;
-                case ModuleKind.Node16:
-                    moduleResolution = ModuleResolutionKind.Node16;
-                    break;
-                case ModuleKind.NodeNext:
-                    moduleResolution = ModuleResolutionKind.NodeNext;
-                    break;
-                default:
-                    moduleResolution = ModuleResolutionKind.Classic;
-                    break;
+export const computedOptions = createComputedCompilerOptions({
+    target: {
+        dependencies: ["module"],
+        computeValue: compilerOptions => {
+            return compilerOptions.target ??
+                ((compilerOptions.module === ModuleKind.Node16 && ScriptTarget.ES2022) ||
+                    (compilerOptions.module === ModuleKind.NodeNext && ScriptTarget.ESNext) ||
+                    ScriptTarget.ES5);
+        },
+    },
+    module: {
+        dependencies: ["target"],
+        computeValue: (compilerOptions): ModuleKind => {
+            return typeof compilerOptions.module === "number" ?
+                compilerOptions.module :
+                computedOptions.target.computeValue(compilerOptions) >= ScriptTarget.ES2015 ? ModuleKind.ES2015 : ModuleKind.CommonJS;
+        },
+    },
+    moduleResolution: {
+        dependencies: ["module", "target"],
+        computeValue: (compilerOptions): ModuleResolutionKind => {
+            let moduleResolution = compilerOptions.moduleResolution;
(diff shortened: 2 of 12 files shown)


andrewbranch:

src/compiler/utilities.ts
```diff
@@ -8588,51 +8588,176 @@ export function getSetExternalModuleIndicator(options: CompilerOptions): (file:
...
-                break;
-        }
-    }
-    return moduleResolution;
-}
+export const getEmitModuleResolutionKind = computedOptions.moduleResolution.computeValue;
+/** @internal */
+export const getEmitModuleDetectionKind = computedOptions.moduleDetection.computeValue;
+/** @internal */
+export const getIsolatedModules = computedOptions.isolatedModules.computeValue;
```
andrewbranch, line 8734: One advantage is that these functions now automatically say what they depend on in their parameter type. The signature of this one, for example, is `(compilerOptions: Pick<CompilerOptions, "isolatedModules" | "verbatimModuleSyntax">) => boolean`

jakebailey (approved):
This seems reasonable to me.

I know the perf doesn't seem to be affected, but I do wonder if we should lift all of these calls in the checker upward so they're all precalculated and local, rather than continually calculating them. e.g. we call `getEmitModuleResolutionKind` and `getESModuleInterop` a looot. As evidenced by this change, they all strictly depend on `compilerOptions`, so could be known immediately.

src/compiler/utilities.ts
```diff
@@ -8588,51 +8588,176 @@ export function getSetExternalModuleIndicator(options: CompilerOptions): (file:
...
+    name: N,
+    dependencies: readonly K[],
+    computeValue: (compilerOptions: Pick<CompilerOptions, N | K>) => R,
+): ComputedCompilerOption<N, K, R> {
+    return { name, dependencies, computeValue };
+}
+
+/** @internal */
+export const computedOptions = {
+    target: computedCompilerOption("target", ["module"], compilerOptions => {
```
sheetalkamat, line 8608: Add `strict` and other strict categories as well ?

sheetalkamat (changes requested):

src/compiler/utilities.ts
```diff
@@ -8587,140 +8587,233 @@ export function getSetExternalModuleIndicator(options: CompilerOptions): (file:
...
-    }),
-    module: computedCompilerOption("module", ["target"], (compilerOptions): ModuleKind => {
-        return typeof compilerOptions.module === "number" ?
-            compilerOptions.module :
-            computedOptions.target.computeValue(compilerOptions) >= ScriptTarget.ES2015 ? ModuleKind.ES2015 : ModuleKind.CommonJS;
-    }),
-    moduleResolution: computedCompilerOption("moduleResolution", ["module", "target"], (compilerOptions): ModuleResolutionKind => {
-        let moduleResolution = compilerOptions.moduleResolution;
-        if (moduleResolution === undefined) {
+export const computedOptions = createComputedCompilerOptions({
```
andrewbranch, line 8603: Thanks @Andarist for helping with this tricky inference 🌟 

This only works in recent nightlies, hence the “LKG” bump (cc @jakebailey)

jakebailey: I don't mind bumping the TS version, though I am wondering which PR was needed to make this pattern work (and of course if there's any change at all that we could avoid it 😅).

But yeah, this pattern is nicer than the previous IMO.

package.json
```diff
@@ -78,7 +78,7 @@
         "playwright": "^1.38.0",
         "source-map-support": "^0.5.21",
         "tslib": "^2.5.0",
-        "typescript": "^5.3.2",
+        "typescript": "^5.4.0-dev.20231206",
```
sheetalkamat, line 81: Is *dev* ok here ?

src/compiler/utilities.ts
```diff
@@ -8587,52 +8587,270 @@ export function getSetExternalModuleIndicator(options: CompilerOptions): (file:
...
-/** @internal */
-export function getEmitScriptTarget(compilerOptions: { module?: CompilerOptions["module"]; target?: CompilerOptions["target"]; }): ScriptTarget {
-    return compilerOptions.target ??
-        ((compilerOptions.module === ModuleKind.Node16 && ScriptTarget.ES2022) ||
-            (compilerOptions.module === ModuleKind.NodeNext && ScriptTarget.ESNext) ||
-            ScriptTarget.ES5);
+type CompilerOptionKeys = keyof { [K in keyof CompilerOptions as string extends K ? never : K]: any; };
+function createComputedCompilerOptions<T extends Record<string, CompilerOptionKeys[]>>(
+    options: {
+        [K in keyof T & CompilerOptionKeys]: {
```
Andarist, line 8593: To get rid of the ` satisfies Record<StrictOptionName, any>` at the resulting object you could do this:

```suggestion
        [K in keyof T & CompilerOptionKeys | StrictOptionName]: {
```

package.json
jakebailey: Yeah, it's okay. We previously had a dev version in #51787, for example.

I'm not totally sure about the `^`, though; not clear to me if this will be relocked every day or what (I'll test quick).

src/compiler/utilities.ts
```diff
@@ -8587,52 +8587,270 @@ export function getSetExternalModuleIndicator(options: CompilerOptions): (file:
...
+                case ModuleResolutionKind.Node16:
+                case ModuleResolutionKind.NodeNext:
+                case ModuleResolutionKind.Bundler:
+                    return true;
+            }
+            return false;
+        },
+    },
+    resolvePackageJsonImports: {
+        dependencies: ["moduleResolution", "resolvePackageJsonExports"],
```
sheetalkamat, line 8703: Do we need `resolvePackageJsonExports`  here?
Asking because eg: for `module` it doesnt include `module` in dependencies but here it does ?

src/compiler/utilities.ts
```diff
@@ -8587,52 +8587,270 @@ export function getSetExternalModuleIndicator(options: CompilerOptions): (file:
...
+                case ModuleKind.Node16:
+                case ModuleKind.NodeNext:
+                    return true;
+            }
+            return false;
+        },
+    },
+    allowSyntheticDefaultImports: {
+        dependencies: ["module", "target", "moduleResolution"],
+        computeValue: (compilerOptions): boolean => {
```
sheetalkamat, line 8674:
```suggestion
        computeValue: compilerOptions => {
```

src/compiler/utilities.ts
```diff
@@ -8587,52 +8587,270 @@ export function getSetExternalModuleIndicator(options: CompilerOptions): (file:
...
+                return compilerOptions.allowSyntheticDefaultImports;
+            }
+            return computedOptions.esModuleInterop.computeValue(compilerOptions)
+                || computedOptions.module.computeValue(compilerOptions) === ModuleKind.System
+                || computedOptions.moduleResolution.computeValue(compilerOptions) === ModuleResolutionKind.Bundler;
+        },
+    },
+    resolvePackageJsonExports: {
+        dependencies: ["moduleResolution"],
+        computeValue: (compilerOptions): boolean => {
```
sheetalkamat, line 8685:
```suggestion
        computeValue: compilerOptions => {
```

src/compiler/utilities.ts
sheetalkamat: Oh i see its exports and this is for imports.. omg.. its so hard to see the difference

src/compiler/utilities.ts
```diff
@@ -8587,52 +8587,270 @@ export function getSetExternalModuleIndicator(options: CompilerOptions): (file:
...
+                case ModuleResolutionKind.NodeNext:
+                case ModuleResolutionKind.Bundler:
+                    return true;
+            }
+            return false;
+        },
+    },
+    resolveJsonModule: {
+        dependencies: ["moduleResolution", "module", "target"],
+        computeValue: (compilerOptions): boolean => {
```
sheetalkamat, line 8723:
```suggestion
        computeValue: compilerOptions => {
```

src/compiler/utilities.ts
```diff
@@ -8587,52 +8587,270 @@ export function getSetExternalModuleIndicator(options: CompilerOptions): (file:
...
+    },
+    declaration: {
+        dependencies: ["composite"],
+        computeValue: compilerOptions => {
+            return !!(compilerOptions.declaration || compilerOptions.composite);
+        },
+    },
+    preserveConstEnums: {
+        dependencies: ["isolatedModules", "verbatimModuleSyntax"],
+        computeValue: (compilerOptions): boolean => {
```
sheetalkamat, line 8738:
```suggestion
        computeValue: compilerOptions => {
```

src/compiler/utilities.ts
```diff
@@ -8587,52 +8587,270 @@ export function getSetExternalModuleIndicator(options: CompilerOptions): (file:
...
+    },
+    incremental: {
+        dependencies: ["composite"],
+        computeValue: compilerOptions => {
+            return !!(compilerOptions.incremental || compilerOptions.composite);
+        },
+    },
+    declarationMap: {
+        dependencies: ["declaration", "composite"],
+        computeValue: (compilerOptions): boolean => {
```
sheetalkamat, line 8750:
```suggestion
        computeValue: compilerOptions => {
```

src/compiler/utilities.ts
```diff
@@ -8587,52 +8587,270 @@ export function getSetExternalModuleIndicator(options: CompilerOptions): (file:
...
+    },
+    isolatedModules: {
+        dependencies: ["verbatimModuleSyntax"],
+        computeValue: compilerOptions => {
+            return !!(compilerOptions.isolatedModules || compilerOptions.verbatimModuleSyntax);
+        },
+    },
+    esModuleInterop: {
+        dependencies: ["module", "target"],
+        computeValue: (compilerOptions): boolean => {
```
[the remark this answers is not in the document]
andrewbranch: These are necessary to avoid circular type inference of the top-level object, since the implementation references `computedOptions`

Issue #56688, opened 2023-12-06
Crash on 5.4.0-dev.20231206


ArnaudBarre:
### 🔎 Search Terms

-

### 🕗 Version & Regression Information

It's working on 5.3
Here is the stacktrace, I hope it helps, I don't really have time to disect which part my codebase triggers this. Feel free to close if this not enough to understand the regression
 
 ```
 TypeError: Cannot read properties of undefined (reading '0')
    at removeConstrainedTypeVariables (/Users/arnaud/git/carbon-calculator/node_modules/typescript/lib/tsc.js:57118:33)
    at getUnionTypeWorker (/Users/arnaud/git/carbon-calculator/node_modules/typescript/lib/tsc.js:57215:9)
    at getUnionType (/Users/arnaud/git/carbon-calculator/node_modules/typescript/lib/tsc.js:57194:12)
    at createJsxAttributesTypeFromAttributesProperty (/Users/arnaud/git/carbon-calculator/node_modules/typescript/lib/tsc.js:69288:216)
    at checkJsxAttributes (/Users/arnaud/git/carbon-calculator/node_modules/typescript/lib/tsc.js:69353:12)
    at checkExpressionWorker (/Users/arnaud/git/carbon-calculator/node_modules/typescript/lib/tsc.js:75700:16)
    at checkExpression (/Users/arnaud/git/carbon-calculator/node_modules/typescript/lib/tsc.js:75557:32)
    at checkExpressionWithContextualType (/Users/arnaud/git/carbon-calculator/node_modules/typescript/lib/tsc.js:75203:18)
    at resolveJsxOpeningLikeElement (/Users/arnaud/git/carbon-calculator/node_modules/typescript/lib/tsc.js:72230:51)
    at resolveSignature (/Users/arnaud/git/carbon-calculator/node_modules/typescript/lib/tsc.js:72303:16)
 ```

### 🙁 Actual behavior

Crash

### 🙂 Expected behavior

No crash

Andarist:
Does it crash on `5.4.0-dev.20231206` too?

Andarist:
[...]

PR #56711, opened 2023-12-08
Prevent `ObjectFlags.IdenticalBaseTypeExists` and `ObjectFlags.IsConstrainedTypeVariable` confusion

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/56688

### DIFF
--- a/tests/baselines/reference/unionReductionWithStringMappingAndIdenticalBaseTypeExistsNoCrash.symbols
+++ b/tests/baselines/reference/unionReductionWithStringMappingAndIdenticalBaseTypeExistsNoCrash.symbols
@@ -1,0 +1,275 @@
+//// [tests/cases/compiler/unionReductionWithStringMappingAndIdenticalBaseTypeExistsNoCrash.tsx] ////
+
+=== node_modules/@types/react/index.d.ts ===
+// https://github.com/microsoft/TypeScript/issues/56688
+
+
+export = React;
+>React : Symbol(React, Decl(index.d.ts, 4, 26))
+
+export as namespace React;
+>React : Symbol(React, Decl(index.d.ts, 3, 15))
+
+declare namespace React {
+>React : Symbol(React, Decl(index.d.ts, 4, 26))
+
+  type ReactNode =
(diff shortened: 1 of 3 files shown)


src/compiler/checker.ts
```diff
@@ -17005,7 +17005,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         const typeVariables: TypeVariable[] = [];
         // First collect a list of the type variables occurring in constraining intersections.
         for (const type of types) {
-            if (getObjectFlags(type) & ObjectFlags.IsConstrainedTypeVariable) {
+            if (type.flags & TypeFlags.Intersection && getObjectFlags(type) & ObjectFlags.IsConstrainedTypeVariable) {
```
Andarist, line 17008: The problem is that `ObjectFlags` reuse some values as they are reserved for certain types, like unions, intersections, substitutions, etc. So the same value is being reused between `ObjectFlags.IdenticalBaseTypeExists` and `ObjectFlags.IsConstrainedTypeVariable`. That created a problem here since the code in this branch assumes that it's dealing with an intersection based on this value - and yet it encountered an object with `ObjectFlags.IdenticalBaseTypeExists`

The union creation here starts with three types:
```
['Element', 'string', 'Capitalize<string>']
```

The `Element` is the one that has `ObjectFlags.IdenticalBaseTypeExists` while `Capitalize<string>` is the one with `StringMapping|IncludesConstrainedTypeVariable` (where `TypeFlags.IncludesConstrainedTypeVariable` is defined as `= StringMapping`).

This `Capitalize<string>` gets eliminated from this union since `string` is its supertype - but it's the one that adds the `TypeFlags.IncludesConstrainedTypeVariable` to the `includes` returned by `addTypesToUnion` (that happens based on the `TypeFlags.IncludesMask`).

jakebailey: Eh, no, that's not helpful; it'd just be preserved if overlapping.

jakebailey (approved):

PR #41635, opened 2020-11-22
Un‑consolidate and fix `WeakMap` constructor overloads

ExE-Boss:
Fixes <https://github.com/microsoft/TypeScript/issues/27951> for `WeakMap`.
Fixes <https://github.com/microsoft/TypeScript/issues/23551> which regressed as a result of <https://github.com/microsoft/TypeScript/pull/28052> only changing the `Iterable`‑taking constructor overload.

PR #28052, opened 2018-10-22
Un-consolidate overloads for Map and WeakMap

andy-ms:
Fixes #27951

PR #56713, opened 2023-12-08
Un‑consolidate and fix `WeakMap` constructor overloads

### LINKED ISSUE #23551: Can't extend WeakMap in type-checked JS files
**TypeScript Version:**  2.8.1

**Search Terms:** 
extend WeakMap, WeakSet, "Base constructors must all have the same return type." 

**Code**

```js
class ColorMap extends WeakMap {
  constructor(color) {
    super();
  }
}
```

Type-checked using:
```
tsc color.js -t es6 --checkJs --allowJs --noEmit
```

**Actual behavior:**
```
color.js(1,24): error TS2510: Base constructors must all have the same return type.
```

**Notes:**
It works with a `Map` or `Set`, or in a `*.ts` file.

### LINKED ISSUE #27951: Extending `Map` with wrong number of type arguments shows no error
**TypeScript Version:**  3.2.0-dev.20181017

**Code**

```ts
class MyMap extends Map<string> {}
```

**Expected behavior:**

Error: `No base constructor has the specified number of type arguments.`

**Actual behavior:**

No error.

Issue was introduced by #25374. The issues it fixed, #23551 and #25362, seem to work fine even if I revert the change. CC @sandersn to confirm that JS class augmentation has been fixed and #25374 can be reverted.

### LINKED ISSUE #56664: Can't extend WeakMap in type-checked JS files
### 🔎 Search Terms

WeakMap constructors return type JSDoc

- This is the behavior in every version I tried.  I opened #23551 5 years ago, but it got closed in error, as it was never fixed.

### ⏯ Playground Link

https://www.typescriptlang.org/play?filetype=js#code/MYGwhgzhAEDCD2J4CcCyYAO0CmAPALtgHYAmMA6tmANbpYDeAUNNMPERPsgK7D4oAKAJTQmLFhG4ZsyYQG5m0AL6MlQA

### 💻 Code

```js
class ColorMap extends WeakMap {
  constructor() {
    super();
  }
}
```

### 🙁 Actual behavior

> color.js(1,24): error TS2510: Base constructors must all have the same return type.

I tried a few different workarounds, like adding @template generic params, or forcing types using @type JSDoc, but nothing seems to work.

### 🙂 Expected behavior

No error.

_No response_

ExE-Boss:
Fixes #56664
Fixes #27951 for `WeakMap`.
Fixes #23551 which regressed as a result of #28052 only changing the `Iterable`‑taking constructor overload.

--------------------------------------------------------------------------------

See also: #41635


src/lib/es2015.iterable.d.ts
```diff
@@ -144,7 +144,7 @@ interface MapConstructor {
 interface WeakMap<K extends WeakKey, V> {}
 
 interface WeakMapConstructor {
-    new <K extends WeakKey, V>(iterable: Iterable<readonly [K, V]>): WeakMap<K, V>;
+    new <K extends WeakKey = WeakKey, V = any>(iterable?: Iterable<readonly [K, V]> | null): WeakMap<K, V>;
```
Andarist, line 147: Wouldn't `unknown` be a better default here? It would more closely match the previous behavior (unconstrained type param is implicitly constrained to `unknown`, at least in strict mode)
```suggestion
    new <K extends WeakKey = WeakKey, V = unknown>(iterable?: Iterable<readonly [K, V]> | null): WeakMap<K, V>;

```

ExE-Boss: The other `WeakMap` constructor uses `any`:
<https://github.com/microsoft/TypeScript/blob/e3d234cfc8653d14e907ae84f918d5bd18546619/src/lib/es2015.collection.d.ts#L65-L68>

And making this change broke some tests.

tests/cases/compiler/extendingCollectionsWithCheckJs.ts
```diff
@@ -3,6 +3,7 @@
 // @noEmit: true
 // @lib: es2017
 
+// @filename: extendingCollectionsWithCheckJs.js
```
sandersn, line 6: =(

tests/cases/compiler/newMap.ts
```diff
@@ -1,2 +1,3 @@
 // @lib: es6
 new Map<string>();
+new WeakMap<object>();
```
sandersn, line 3: why do we need to support `new WeakMap<object>()` instead of `new WeakMap<object, object>()`? If only the latter were supported, we wouldn't need to add the type parameter defaults in this PR, would we?

sandersn (dismissed):
Looks reasonable, but I have one question and I'd like a second opinion from @rbuckton.

tests/cases/compiler/extendingCollectionsWithCheckJs.ts
ExE-Boss: This has to end in `.js` so that the code is actually type checked as a **JavaScript** file.

tests/cases/compiler/newMap.ts
ExE-Boss: Removing support for this would be a breaking change, as the other constructor has type parameter defaults:
https://github.com/microsoft/TypeScript/blob/b2c6a56e38bc8cca1573aeadf8c2bda74a67e8e9/src/lib/es2015.collection.d.ts#L66

jakebailey:
@typescript-bot test it

DanielRosenwasser (approved):
Don't love that there's more `any`, but most people were probably hitting the other overload anyway.

PR #56723, opened 2023-12-09
Add asynciterable libs

### LINKED ISSUE #29867: Support async iterable DOM types
## Search Terms

asynciterator asynciterable async iterable iterator

## Suggestion

Add a `dom.asynciterable` option to `--lib` to support DOM types that define a `Symbol.asyncIterator` method.

## Use Cases

The [streams spec](https://streams.spec.whatwg.org/) has [recently been updated](https://github.com/whatwg/streams/commit/bf2cac8a52664df3e6da7a48755890e87b00953a) to add a `Symbol.asyncIterator` method to `ReadableStream`, making it the first DOM type to implement the [ES2018 `AsyncIterable` interface](https://www.ecma-international.org/ecma-262/9.0/index.html#sec-asynciterable-interface).

I would like to update the type definitions for streams in [TSJS-lib-generator](https://github.com/Microsoft/TSJS-lib-generator/) to include this new async iterator method. However, this requires `Symbol.asyncIterator`, which is not available before ES2018 and thus cannot be readily used in the DOM type definitions.

Previously, a `dom.iterable` option was added to support DOM types that define a `Symbol.iterator` method. I think it would make sense to use this same strategy to support `Symbol.asyncIterator`.

## Examples
TypeScript would have (generated) type definitions in `lib/lib.dom.asynciterable.d.ts`:
```typescript
interface ReadableStream<R = any> {
  [Symbol.asyncIterator](): AsyncIterator<R>;
}
```

Users could then use a `ReadableStream` in a `for await..of` loop by compiling with `--target es2018`:
```typescript
const readable = new ReadableStream<Uint8Array>();
for await (const chunk of readable) {
  console.log(chunk);
}
```

## Checklist

### LINKED ISSUE #56360: Missing FileSystemDirectoryHandle methods in the lib.dom.d.ts
### 🔎 Search Terms

typescript, visual-studio-code, intellisense, origin-private-system, FileSystemDirectoryHandle, lib.dom.d.ts

https://stackoverflow.com/questions/77453180/missing-filesystemdirectoryhandle-keys-method-in-vscode-intellisense

The nightly version and Playground do not display methods either.

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about FileSystemDirectoryHandle methods.

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.4.0-dev.20231110#code/BQQwzgngdgxgBMAlHAvAPjgbwFBz3GAeyjABc4AnQw0gEQEsLU4QB3Ee8qEAN3oHMQpQhQB0ZESH4BTUTLqNpMYRQhIA3LnwB6bXABihADZHCrelH5wAttNIALQgBMwN+mDAX+AGjgAjAFdyMACAB1CRUlcLG0IyfypWMGkKV2B7UlJQsAAuXSdpHmlTUJTRa0IAL3oTEFERfm1pKABaAFUAZW0nQhgwbQB1aT9tAEEABQBJbX0a6Q6IMmlrBgolFQgACRAoJyNpAGI-ROSKAH0ia1Chej8azghELTwqGlXRAGtpRY1nymoFGJmqQKPRpGBfvh-m9GKIeCAjAFwb8AL6IJBAA

### 💻 Code

```ts
    const rootDir = await navigator.storage.getDirectory();
    // Following methods missing, but supports in most browsers (https://developer.mozilla.org/en-US/docs/Web/API/FileSystemDirectoryHandle#browser_compatibility)
    rootDir.keys();
    rootDir.entries();
    rootDir.values();
```

### 🙁 Actual behavior

TypeScript write: Property 'keys' does not exist on type 'FileSystemDirectoryHandle'.(2339).

### 🙂 Expected behavior

That the TypeScript check recognizes the methods as valid.

_No response_

saschanaz:
Fixes #29867
Fixes #56360
Fixes https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1639


sandersn (approved):

thomas-darling:
Sorry to revive this, but [this line in es2022.full.d.ts](https://github.com/microsoft/TypeScript/blob/main/src/lib/es2022.full.d.ts#L6C1-L6C42) indicates there should be a lib named `dom.asynciterable` - but when I try to add that to my own set of libs in my `tsconfig.json`, it says no such lib exists.

Is that a bug?

saschanaz:
Can you file a new issue with your TS version and full tsconfig?

PR #56725, opened 2023-12-10
Remove redundant `hasNonCircularBaseConstraint` check

Andarist:


src/compiler/checker.ts
```diff
@@ -22255,7 +22255,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     // This is a carve-out in comparability to essentially forbid comparing a type parameter
                     // with another type parameter unless one extends the other. (Remember: comparability is mostly bidirectional!)
                     let constraint = getConstraintOfTypeParameter(source);
-                    if (constraint && hasNonCircularBaseConstraint(source)) {
+                    if (constraint) {
```
Andarist, line 22258: this `constraint` is the result of `getConstraintOfTypeParameter` call (located just one line above it) and `getConstraintOfTypeParameter` already handles exactly this thing:
```ts
function getConstraintOfTypeParameter(typeParameter: TypeParameter): Type | undefined {
    return hasNonCircularBaseConstraint(typeParameter) ? getConstraintFromTypeParameter(typeParameter) : undefined;
}
```

So this check here is completely redundant as it rechecks what has just been checked

jakebailey (approved):

sandersn (approved):

Issue #56726, opened 2023-12-10
Non-homomorphic instantiation of a homomorphic mapped type is not spreadable when it has an array constraint


Andarist:
### 🔎 Search Terms

mapped type homomorphic instantiation array tuple constraint spread spreadable

### 🕗 Version & Regression Information

- This is the behavior in every version I tried

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.4.0-dev.20231210#code/C4TwDgpgBAEghgOwCYBsIHkzAJYHsEDOAPOgHxQC8UA3gLABQUUA2gJJTYJQDWEIuAMyjoAugC4aDJkwBucFAFcIE9GxEBuKVAC+m+roYMkEAMYo4AJ2gCFCEznxQbdgIxEtAFSgQAHsAjIBFAASqa4FkhEAAoWuJAWoADSfAA0UFZwSPgoIFCIIMwipCkMpAAUAtgQKEgEEnSMLIkcXLz8Qh7iko1M5gBG1RIEwBacAOZ60lBxDoQSzAB0S-DIaJizxB7MiUUaWgb6AJQSHnoMJvjD6RAECijAlE62Ji5lDUxgsWD1Wr1wAygJAAiSoWYZAko9aZYPBzFi-aTvKbSOSKZRQFwAJgAzJDkTo8cikfjUUpgQJcLgIQimNpCUwRIS6VpcMAABYQCw-KH9QZQIEEMLIalQmawurwqFMYnI0nooF9Swi-HMqXdfGyeRkqAjJT06SqqaM-aQ7SHPRAA

### 💻 Code

```ts
type HandleOptions<O> = {
  [I in keyof O]: {
    value: O[I];
  };
};

declare function func1<
  T extends Record<PropertyKey, readonly any[]>,
>(fields: {
  [K in keyof T]: {
    label: string;
    options: [...HandleOptions<T[K]>];
  };
}): T;

const result = func1({
  prop: {
    label: "first",
    options: [
      {
        value: 123,
      },
      {
        value: "foo",
      },
    ],
  },
  other: {
    label: "second",
    options: [
      {
        value: "bar",
      },
      {
        value: true,
      },
    ],
  },
});
```

### 🙁 Actual behavior

Error at line 12: 
> A rest element type must be an array type.(2574)

### 🙂 Expected behavior

The error should not be reported

PR #56727, opened 2023-12-10
Fixed apparent type of homomorphic mapped type with non-homomorphic instantiation

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/56726

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -2047,6 +2047,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     var noConstraintType = createAnonymousType(/*symbol*/ undefined, emptySymbols, emptyArray, emptyArray, emptyArray);
     var circularConstraintType = createAnonymousType(/*symbol*/ undefined, emptySymbols, emptyArray, emptyArray, emptyArray);
     var resolvingDefaultType = createAnonymousType(/*symbol*/ undefined, emptySymbols, emptyArray, emptyArray, emptyArray);
+    var resolvingApparentMappedType = createAnonymousType(/*symbol*/ undefined, emptySymbols, emptyArray, emptyArray, emptyArray);
 
     var markerSuperType = createTypeParameter();
     var markerSubType = createTypeParameter();
@@ -2126,7 +2127,6 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     var amalgamatedDuplicates: Map<string, DuplicateInfoForFiles> | undefined;
     var reverseMappedCache = new Map<string, Type | undefined>();
     var homomorphicMappedTypeInferenceStack: string[] = [];
-    var homomorphicMappedTypeApparentTypeStack: Type[] = [];
     var ambientModulesCache: Symbol[] | undefined;
     /**
      * List of every ambient module with a "*" wildcard.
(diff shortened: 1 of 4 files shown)


src/compiler/checker.ts
```diff
@@ -14441,11 +14442,26 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+        if (typeVariable && !mappedType.declaration.nameType) {
+            let constraint: Type | undefined;
+            if (!type.target) {
+                constraint = getConstraintOfTypeParameter(typeVariable);
+            }
+            else {
+                const modifiersConstraint = getConstraintOfType(getModifiersTypeFromMappedType(type));
+                if (modifiersConstraint) {
+                    homomorphicMappedTypeApparentTypeStack.push(type);
+                    constraint = getApparentType(modifiersConstraint);
```
Andarist, line 14459: I'm not super sure if I should be calling `getApparentType` here but then... we get to this place from `getApparentType` so it feels like it's OK to do right now.

src/compiler/checker.ts
```diff
@@ -14441,11 +14442,26 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
-            const constraint = getConstraintOfTypeParameter(typeVariable);
+        if (contains(homomorphicMappedTypeApparentTypeStack, type)) {
+            return type;
+        }
+        const mappedType = type.target as MappedType || type;
+        const typeVariable = getHomomorphicTypeVariable(mappedType);
+        if (typeVariable && !mappedType.declaration.nameType) {
+            let constraint: Type | undefined;
+            if (!type.target) {
+                constraint = getConstraintOfTypeParameter(typeVariable);
```
Andarist, line 14453: The other branch is now using `getApparentType`. If that's the correct thing to do then I feel like this probably should call that too. This would fix issues like this one:
```ts
declare function func2<A extends readonly any[], B extends A>(
  arr: A,
  fields: [
    // A rest element type must be an array type.(2574)
    ...{
      [K in keyof B]: { v: B[K] };
    },
  ],
): [A, B];

declare const tuple: [number | string, number | boolean, number | bigint];

const result = func2(tuple, [{ v: "foo" }, { v: true }, { v: 6 }]);
```

weswigham (approved):
This looks OK to me - calling `getApparentType` on part of a type while resolving an apparent type is fairly normal behavior (mapped types can just be circular, which you guard against), and using that information to see if the apparent mapped type should be array or tuple like seems reasonable to me. Still, I'd like to let @ahejlsberg get a chance to review this if he'd like, too, so once the extended tests come back (presumably clean), I'll hold off on merging this for a little bit.

weswigham:
@typescript-bot test this
@typescript-bot run dt
@typescript-bot test top100

src/compiler/checker.ts
weswigham: You probably could... or you could move the apparent type check forward and replace `everyType(constraint, isArrayOrTupleType)` with something like `everyType(constraint, isApparentArrayOrTupleType)`, provided it's not critical the instantiation here be with the apparent types (and I don't *think* the instantiation using an apparent constraint is the important part, since mapped types have lazy apparent type members). It's possible changing the should-this-be-a-mapped-array check we use *everywhere* to something that looks into the apparent types of the members of the constraint could be prudent.

jakebailey:
This PR was missing a "user test this" and broke VS Code's compiling, though I think in a good way:

```ts
node_modules/@types/sinon/index.d.ts:47:20 - error TS2370: A rest parameter must be of an array type.

47         calledWith(...args: Partial<MatchArguments<TArgs>>): boolean;
                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

node_modules/@types/sinon/index.d.ts:1120:13 - error TS2370: A rest parameter must be of an array type.

1120             ...args: Partial<MatchArguments<TArgs>>
                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

node_modules/@types/sinon/index.d.ts:1127:69 - error TS2370: A rest parameter must be of an array type.

1127         alwaysCalledWith<TArgs extends any[]>(spy: SinonSpy<TArgs>, ...args: Partial<MatchArguments<TArgs>>): void;
                                                                         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

node_modules/@types/sinon/index.d.ts:1133:68 - error TS2370: A rest parameter must be of an array type.

1133         neverCalledWith<TArgs extends any[]>(spy: SinonSpy<TArgs>, ...args: Partial<MatchArguments<TArgs>>): void;
                                                                        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


Found 4 errors in the same file, starting at: node_modules/@types/sinon/index.d.ts:47
```

Andarist:
It is slightly surprising to me since I wouldn't exactly expect some types to **stop** being spreadable with this change. Unless this is hitting some circularity case here or something. I'll reduce this to a minimal repro case later so we can assess with more confidence what happens here.

PR #56732, opened 2023-12-11
isolatedModules error on global shadowed by imported type

### LINKED ISSUE #56521: Incorrect ts.transpileModule() emit when imported type shadows type but not constructor of global symbol
### 🔎 Search Terms

transpileModule, isolatedModules

- This is the behavior in every version I tried: v5.2.2, v5.3.2, v5.4.0-dev.20231123

### ⏯ Playground Link

[Bug Workbench](https://www.typescriptlang.org/dev/bug-workbench/?isolatedModules=true&ts=5.3.2#code/PTAEAEEsGcHsBsCGAXApgEwLK3QV3qtAFBGQC2ADrAE7KgDeAIiqgL6gBm1sZoARADpgyAJ4VCfANwkOuAHYBjZJFhzOsWAApEALlDM0ASgZFQoBauh0ARqAC8oOagDu+ltoEjUiagBpQiAJkqsgAFv6B6IgihtJm1KjIuNRq1gIA5okAKuSomrFErCQgEByQBHKIZKh6ouLQAugCyMSoAB5UtKCQcmjUHIgKqG5oJmZRInpyuGTWqNRxoMG9oVMzcwumoF4+a7Pz0qxAA)

### 💻 Code

```ts
// @isolatedModules

import {Date} from "./types";

function foo(a: Date) {
  const b = new Date(a.year, a.month, a.day);
  return b.getTime();
}

// @filename: types.d.ts
export interface Date {
  day: number;
  month: number;
  year: number;
}
```

### 🙁 Actual behavior

Regular emit:

```js
function foo(a) {
  const b = new Date(a.year, a.month, a.day);
  return b.getTime();
}
```

`ts.transpileModule()` emit:

```js
import {Date} from "./types";
function foo(a) {
  const b = new Date(a.year, a.month, a.day);
  return b.getTime();
}
```

The `ts.transpileModule()` emit has a runtime error because the imported `Date` either doesn't exist or doesn't have a constructor.

### 🙂 Expected behavior

Same (or at least runtime equivalent) emit of transpileModule.

I realize this code is probably not ideal. I was honestly surprised that this type-checks. I'm happy with any solution here:

- Make this code an error in all configurations.
- Make `isolatedModules` require `import type`.
- ...

frigus02:
Fixes #56521

I took inspiration from #56354, which seems very similar.

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -735,6 +735,7 @@ import {
     isTypeNode,
     isTypeNodeKind,
     isTypeOfExpression,
+    isTypeOnlyImportDeclaration,
     isTypeOnlyImportOrExportDeclaration,
     isTypeOperatorNode,
     isTypeParameterDeclaration,
@@ -3601,6 +3602,19 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                             typeOnlyDeclaration,
                             unescapedName,
                         );
+                    }
+                }
+
+                // Look at 'compilerOptions.isolatedModules' and not 'getIsolatedModules(...)' (which considers 'verbatimModuleSyntax')
+                // here because 'verbatimModuleSyntax' will already have an error for importing a type without 'import type'.
+                if (compilerOptions.isolatedModules && result && isInExternalModule && meaning & SymbolFlags.Value) {
+                    const isGlobal = lookup(globals, name, meaning) === result;
+                    const typeSymbol = isSourceFile(lastLocation!) && lastLocation.locals && lookup(lastLocation.locals, name, SymbolFlags.Type);
+                    if (isGlobal && typeSymbol) {
+                        const importDecl = typeSymbol.declarations?.find(d => d.kind === SyntaxKind.ImportSpecifier || d.kind === SyntaxKind.ImportClause || d.kind === SyntaxKind.NamespaceImport || d.kind === SyntaxKind.ImportEqualsDeclaration);
+                        if (importDecl && !isTypeOnlyImportDeclaration(importDecl)) {
+                            error(importDecl, Diagnostics.Import_0_conflicts_with_global_value_used_in_this_file_so_must_be_declared_with_a_type_only_import_when_isolatedModules_is_enabled, unescapeLeadingUnderscores(name));
+                        }
                     }
                 }
             });
--- a/src/compiler/diagnosticMessages.json
+++ b/src/compiler/diagnosticMessages.json
@@ -3731,6 +3731,10 @@
         "category": "Error",
         "code": 2865
     },
+    "Import '{0}' conflicts with global value used in this file, so must be declared with a type-only import when 'isolatedModules' is enabled.": {
+        "category": "Error",
+        "code": 2866
+    },
 
     "Import declaration '{0}' is using private name '{1}'.": {
         "category": "Error",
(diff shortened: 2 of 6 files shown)


src/compiler/checker.ts
```diff
@@ -3603,6 +3604,19 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                         );
                     }
                 }
+
+                // Look at 'compilerOptions.isolatedModules' and not 'getIsolatedModules(...)' (which considers 'verbatimModuleSyntax')
+                // here because 'verbatimModuleSyntax' will already have an error for importing a type without 'import type'.
+                if (compilerOptions.isolatedModules && result && isInExternalModule && meaning & SymbolFlags.Value) {
+                    const isGlobal = lookup(globals, name, meaning) === result;
+                    const typeSymbol = isSourceFile(lastLocation!) && lastLocation.locals && lookup(lastLocation.locals, name, SymbolFlags.Type);
```
andrewbranch, line 3612: I don’t think this is quite right—the characteristic of the local symbol that makes it problematic isn’t that it _is a type_; it’s that it _is not a value_. In other words, there could be potential for false positives or false negatives in a few cases here:

- the imported symbol is a type and a value at the same time, like a class (I don’t think this case is possible since `isGlobal` would be false)
- the imported symbol is an uninstantiated namespace
- the imported symbol was unresolvable and so has all `SymbolFlags` (again, shouldn’t be an issue here since `isGlobal` would be false)

Of these corner cases to think about, it sounds like uninstantiated namespaces might pose a problem. Here’s a test:

```ts
// @isolatedModules: true
// @noEmit: true
// @noTypesAndSymbols: true

// @filename: ./types.ts
export namespace Date {
  export type T = any;
}

// @filename: ./bad.ts
import { Date } from './types';
new Date();
```

Maybe a lookup with `~SymbolFlags.Value` instead of `SymbolFlags.Type` would better capture the intent?

andrewbranch (changes requested):
Thanks! I think this is pretty close.

src/compiler/checker.ts
frigus02: Good catch and thanks for the explanation. Changed to `~SymbolFlags.Value` and added the test case.

frigus02:
Thanks for the quick review! I think I addressed your comments.

(Not sure about the comment resolving etiquette here. I hope me resolving them is fine?)

frigus02:
The new error currently incorrectly fires for the following Node.js code:

```ts
import { Console } from "node:console";
const c: Console = new Console();
```

That's because `console.Console` is indeed a global. But it's re-exported in the module `node:console`. Importantly, though, the global is `console.Console`, not `Console`, so the import is necessary.

Node.js typings for reference: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/420d2ceb50f728aa715da1250a2a9f27760a640c/types/node/console.d.ts#L412

Weirdly there is no error for this code:

```ts
import { Console } from "node:console";
new Console();
```

I'm going to look into that tomorrow. If you have any ideas, I'd definitely appreciate your help, @andrewbranch.

frigus02:
Aha. I think `SymbolFlags.Value` is not a single bit. I just copied the check from a few lines up. But I think I need to use `(meaning & SymbolFlags.Value) === SymbolFlags.Value` here. I'll test some more tomorrow and update the code and tests.

andrewbranch:
Oh, good catch. Yeah, I think something like that should work, or alternatively a lookup with `SymbolFlags.Type | SymbolFlags.NamespaceModule`

frigus02:
As far as I can tell this looks good now. I did not use `SymbolFlags.Type | SymbolFlags.NamespaceModule`, yet. But happy to switch to that if you prefer.

andrewbranch:
@typescript-bot test top200

andrewbranch (approved):
Thanks @frigus02!

PR #53463, opened 2023-03-23
Isolated declarations

dragomirtitian:
Fixes #47947 

## Design Goals

When adding this new flag to TypeScript, these are the design goals we had in mind:

1. No inferred type information should be used in declaration emit.
2. Declaration emit should be a syntactic transformation that adds type information present in the current statement only. (i.e. It does not use information from outside the file, such as types of globals or imports). 
3. Declaration emit can be done on a per-file basis, without information in other files.
4. An external declaration emitter tool can emit the exact same declaration file as TypeScript if the original source code does not have syntactic or semantic errors.
5. An external declaration emitter tool should be able to make a best effort to emit the same declarations as TypeScript if the original source code has semantic or syntactic errors.

Non goals:

1. Replicate the inference behavior of TypeScript. Even for trivial cases a declaration emitter should not be required to infer types from values. If values can’t be preserved in declaration files, explicit type annotations will be required.
2. Preserve the current declaration emit.

## Status

The work-in-progress branch is here which implements the flag:
  https://github.com/bloomberg/TypeScript/tree/isolated-declarations

The new standalone declaration emitter is located as a sub-package here:
  https://github.com/bloomberg/TypeScript/tree/isolated-declarations/external-declarations

There are three deliverables:

1. **Error generation** in the compiler to enforce the opt-in flag-enabled source code 
restrictions.
   - These error conditions are described below as "Restrictions".
2. A standalone **declaration emitter**
   - This implementation uses the the parser and printer from the TypeScript code base but replaces the binder and emit resolver with rewritten versions that do not depend on the type checker.
    - We should decide if this emitter should remain in the TS repository or be spinned out into its own repository. Code changes are still needed to it in either scenario. 
3. **Conformance tests** for the new declaration emit
   - These are designed to be reusable to enable the ecosystem to create compliant declaration emitters, e.g. in the esbuild and swc open-source compilers.

## Restrictions

*When `--isolatedDeclarations` is enabled, TypeScript will enforce stricter rules than usual on the user's source code.  These restrictions are necessary in order to keep the Declaration Emit to be a simple one-in-one-out file transform.*

### R1. Do not print synthetic types
This is the main restriction that enables isolated declarations to work. Type annotations will be required on: 
- Function/Method/Accessor parameters
- Function/Method/Accessor return types
- Variables 
- Public/protected class fields

The requirement only exists if the symbol makes it into the declaration file (directly by exporting or indirectly by referencing in another exported symbol)

For example these are now errors with isolated declarations:

```ts
[...]

PR #56736, opened 2023-12-11
isolatedModules errors for non-literal enum initializers

### LINKED ISSUE #56153: Certain resolution-dependent enum emit isn't correctly flagged as an error under `isolatedModules`
### 🔎 Search Terms

- isolatedModules
- enum
- cross module
- export

- This is a crash

Test on Playground with version 3.9.7 and version 5.2.2, got the same incorrect result.

### ⏯ Playground Link

[Workbench Repro](https://www.typescriptlang.org/dev/bug-workbench/?noImplicitAny=false&target=99&isolatedModules=true&ts=5.2.2#code/PTAEAEGcAsHsHcCiBbAlgFwFAgqysAbAQ3QFMATAWVnIFcDTIAuUdAJ1tOzHADNUGAOyLJSLXrFgA6dJEykAHgAdYbdKFKDayUADFJoAN6ZQp0AEFQAXlABGAAyYAvpm4R+QkWNCpB5RTJyqMgqakZ6Bk6gvGywOgBEUsASsPEA3PLKquqa2qAAQkRsRiZm+dYR0uYANKWmAMLOmEA)

### 💻 Code

```ts repro
// @showEmit
// @isolatedModules: true
// @filename: foo.ts
export enum Foo {
    A = 10
}

// @filename: index.ts
import { Foo } from "./foo";
export enum Bar {
    B = Foo.A,
    C
}

```

[Workbench Repro](https://www.typescriptlang.org/dev/bug-workbench/?noImplicitAny=false&target=99&isolatedModules=true&ts=5.2.2#code/PTAEAEGcAsHsHcCiBbAlgFwFAgqysAbAQ3QFMATAWVnIFcDTIAuUdAJ1tOzHADNUGAOyLJSLXrFgA6dJEykAHgAdYbdKFKDayUADFJoAN6ZQp0AEFQAXlABGAAyYAvpm4R+QkWNCpB5RTJyqMgqakZ6Bk6gvGywOgBEUsASsPEA3PLKquqa2qAAQkRsRiZm+dYR0uYANKWmAMLOmEA)

### 🙁 Actual behavior

Please notice the value of Bar.C

```JavaScript 
import { Foo } from "./foo";
export var Bar;
(function (Bar) {
    Bar[Bar["B"] = 10] = "B";
    Bar[Bar["C"] = 11] = "C";
})(Bar || (Bar = {}));
```

### 🙂 Expected behavior

Emit a warning or an error, and generate code that does not rely on other imports.

Example:
```JavaScript
import { Foo } from "./foo";
export var Bar;
(function(Bar) {
    Bar[Bar["B"] = Foo.A] = "B";
    Bar[Bar["C"] = void 0] = "C";
})(Bar || (Bar = {}));
```

The Bar.C is void 0, since it cannot be refered.

- https://github.com/swc-project/swc/issues/8147

frigus02:
Fixes #56153

Adds 2 new errors under `isolatedModules` only:

- (1)   When an enum initializer expression that isn't a string literal has a string type
- (2)   When an enum member without an initialization expression follows an enum member with an initialization expression that isn't a numeric literal

Ran the new checks across our code base. I wanted to check if we need to support computed values based on local identifiers. But we have so few errors, I don't think this is necessary:

| Error | Percent of TS projects failing with the new error |
| --- | --- |
| (1) identifier imported from another file |	0.0059% |
| (1) identifier local to the same file |	0.0025% |
| (2) |	0.0002% |
| sum |	0.0085% |


jakebailey:
@typescript-bot test top200
@typescript-bot run dt

@typescript-bot pack this

jakebailey:

src/compiler/checker.ts
```diff
@@ -45073,6 +45087,44 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         return value;
     }
 
+    function evaluatesToNumericLiteral(expr: Expression): boolean {
```
andrewbranch, line 45090: Nit: I think this name describes an oxymoron. Maybe `isSyntacticallyNumeric`?

andrewbranch (changes requested):
Thanks @frigus02!

Whatever we do in `evaluatesToStringLiteral` and `evaluatesToNumericLiteral` here will become the minimum requirement for all third-party compilers to implement to guide their own emit. I think what you have is pretty reasonable, but we should get the team, and other compiler maintainers, to weigh in. @nicolo-ribaudo @evanw @kdy1

src/compiler/diagnosticMessages.json
```diff
@@ -7936,5 +7936,13 @@
     "'await using' statements cannot be used inside a class static block.": {
         "category": "Error",
         "code": 18054
+    },
+    "A member initializer in a enum declaration for a string value must be a string literal when 'isolatedModules' is enabled.": {
```
andrewbranch, line 7940: If `evaluatesToStringLiteral` and `evaluatesToNumericLiteral` remain fairly complex, we should probably try to find more accurate wording for these messages.

jakebailey:
At least for `esbuild`, my understanding is that it doesn't really care about enum stuff that violates `isolatedModules`; for example, TS itself uses exported const enums and to do so disables `isolatedModules` and `esbuild` will happily inline them (which is good).

The new test cases don't fail currently in `esbuild`, so I'm not sure there's actually any work to be done outside of TS if this PR just intends to add more errors.

src/compiler/checker.ts
```diff
@@ -45073,6 +45087,44 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+                const leftIsNumeric = evaluatesToNumericLiteral(left);
+                const rightIsNumeric = evaluatesToNumericLiteral(right);
+                return (
+                    !(leftIsNumeric && rightIsNumeric) &&
+                    (evaluatesToStringLiteral(left) || leftIsNumeric) &&
+                    (evaluatesToStringLiteral(right) || rightIsNumeric) &&
+                    (expr as BinaryExpression).operatorToken.kind === SyntaxKind.PlusToken
+                );
+            case SyntaxKind.TemplateExpression:
+                return (expr as TemplateExpression).templateSpans.every(span => evaluatesToStringLiteral(span.expression));
```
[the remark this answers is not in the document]
frigus02: In this PR I wanted to implement https://github.com/microsoft/TypeScript/issues/56153#issuecomment-1773331706. I struggled with "When an enum initializer expression that isn't a string literal has a string type". What counts as "string literal" here? Just `""` or also parenthesized expressions or ...? I ended up copying the logic from the `evaluate` function (see below starting in line 45127).

The result of `evaluate` is currently used to determine if reverse mappings need to be created ([here via `getConstantValue`](https://github.com/microsoft/TypeScript/blob/01a51d2d017dd50d487835282b7e4a2a4f7300ab/src/compiler/transformers/ts.ts#L1876)). Without changing that, `evaluatesToStringLiteral` can only return `true` when `evaluate` returns a string.

I'd actually be in favor of changing that. See my comment https://github.com/microsoft/TypeScript/issues/56153#issuecomment-1849782437. I'd prefer if an enum did not generate reverse mappings for template expressions. Then we could just `return true` here.

I'm also open to just `return false` here in this PR.

src/compiler/diagnosticMessages.json
frigus02: Makes sense. Something like these maybe?

- A string member initializer in a enum declaration can't include references when 'isolatedModules' is enabled.
- A string member initializer in a enum declaration can only use constant expressions when 'isolatedModules' is enabled.

src/compiler/checker.ts
```diff
@@ -45040,11 +45042,17 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
         // occurs if the immediately preceding member is not a constant enum member.
-        if (autoValue !== undefined) {
-            return autoValue;
+        if (autoValue === undefined) {
+            error(member.name, Diagnostics.Enum_member_must_have_initializer);
+            return undefined;
         }
-        error(member.name, Diagnostics.Enum_member_must_have_initializer);
-        return undefined;
+        if (compilerOptions.isolatedModules && previous?.initializer && !evaluatesToNumericLiteral(previous.initializer)) {
```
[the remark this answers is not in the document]
frigus02: Done.

src/compiler/checker.ts
frigus02: Done.

frigus02 (commented):
Thanks for the review, Andrew and Jake. I addressed the comments.

This implementation is quite thorough. I wanted to avoid too many new errors in our code base when turning on isolatedModules, so I tried to support as much as possible. I'm happy to simplify it if you like, though.

src/compiler/checker.ts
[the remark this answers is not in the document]
andrewbranch: I think one of us isn’t understanding something specific about this `TemplateExpression` case. A “string literal” means literally something in single or double quotes, but you instead implemented “trivially syntactically determinable to evaluate to a string” which I think is fine and good. But by that definition, I don’t see why you don’t return `true` for cases like `` `foo${1}` `` or indeed even `` `foo${a}${b}${await import("c").blah}` ``. You’re guaranteed to get a string no matter what values get interpolated into the template spans. I expected just a `return true` here. Unless I’m missing something big!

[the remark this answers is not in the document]
frigus02: > I think one of us isn’t understanding something specific about this `TemplateExpression` case.

Yeah. I wouldn't be surprised if that was me. Let me start with my goal:

I want to ensure that the regular JS emit and ts.transpileModule emit are the same (or equivalent) for enums.

TS emits reverse mappings for template expression initializers if an expression inside can't be resolved to a constant string. Example [workbench](https://www.typescriptlang.org/dev/bug-workbench/?#code/PTAEAEDMEsBsFMB2BDAtvAXKAdMSB7fbAFwGcAoaVAB3wCdjQBvUAIQEEAlUAX1Ejr5UoAOS4ARsjoiA3OSQBXYQDFCzUO1ABeUAAMAJEw6ceu3uXIgIMBCnRYJUkhXgAPWg1ABjfIlKNjbVFJaRkgA):

```ts
import {BAR} from './bar';
enum Foo { A = `${BAR}` }
```

This emits `Foo["A"] = "bar";` if the `./bar` module is known. But with ts.transpileModule it emits ```Foo[Foo["A"] = `${BAR}`] = "A";```. This difference breaks a few tests in our code base.

If my understanding is correct so far, I think we have 2 options for template expressions here:

- `return false`: The more common (at least in our code base) NoSubstritionTemplateLiteral would still work. But TemplateExpressions would not be allowed under isolatedModules as enum member initializers.
- Some version of what I have in this PR.
- `return true` _and_ change TS emit so a TemplateExpression does not generate reverse mappings.

I'm open to all of those.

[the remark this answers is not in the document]
andrewbranch: Ah, nope, it was me who was confused. This makes sense. You need to ensure that the actual value of the string is statically knowable for a reverse mapping, not just that it’s going to be a string.

src/compiler/diagnosticMessages.json
frigus02: I updated the error message with the 2nd option.

frigus02:
It's probably too late to get this into 5.4 now, is it? That's alright.

<details>
<summary>
Rather than introducing 2 new functions, which duplicate some of the `evaluate` logic, we could also extend `evaluate` to return whether the evaluation was purely syntactic, e.g.:</summary>

```diff
diff --git a/src/compiler/checker.ts b/src/compiler/checker.ts
index b0ea7ea9ff..78cab2d428 100644
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -38687,7 +38687,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         if (isConstContext(node) || isTemplateLiteralContext(node) || someType(getContextualType(node, /*contextFlags*/ undefined) || unknownType, isTemplateLiteralContextualType)) {
             return getTemplateLiteralType(texts, types);
         }
-        const evaluated = node.parent.kind !== SyntaxKind.TaggedTemplateExpression && evaluateTemplateExpression(node);
+        const evaluated = node.parent.kind !== SyntaxKind.TaggedTemplateExpression && evaluateTemplateExpression(node).value;
         return evaluated ? getFreshTypeOfLiteralType(getStringLiteralType(evaluated)) : stringType;
     }
 
@@ -45010,15 +45010,17 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         if (!(nodeLinks.flags & NodeCheckFlags.EnumValuesComputed)) {
             nodeLinks.flags |= NodeCheckFlags.EnumValuesComputed;
             let autoValue: number | undefined = 0;
+            let previous: EnumMember | undefined;
             for (const member of node.members) {
-                const value = computeMemberValue(member, autoValue);
+                const value = computeMemberValue(member, autoValue, previous);
                 getNodeLinks(member).enumMemberValue = value;
                 autoValue = typeof value === "number" ? value + 1 : undefined;
+                previous = member;
             }
         }
     }
 
-    function computeMemberValue(member: EnumMember, autoValue: number | undefined) {
+    function computeMemberValue(member: EnumMember, autoValue: number | undefined, previous: EnumMember | undefined) {
         if (isComputedNonLiteralName(member.name)) {
             error(member.name, Diagnostics.Computed_property_names_are_not_allowed_in_enums);
         }
@@ -45040,26 +45042,41 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         // If the member is the first member in the enum declaration, it is assigned the value zero.
         // Otherwise, it is assigned the value of the immediately preceding member plus one, and an error
         // occurs if the immediately preceding member is not a constant enum member.
-        if (autoValue !== undefined) {
-            return autoValue;
-        }
+        if (autoValue === undefined) {
             error(member.name, Diagnostics.Enum_member_must_have_initializer);
             return undefined;
         }
+        if (getIsolatedModules(compilerOptions) && previous?.initializer) {
+            const previousResult = evaluate(previous.initializer);
+            if (typeof previousResult.value !== "number" || !previousResult.syntactic) {
+                error(
+                    member.name,
+                    Diagnostics.Enum_member_following_a_non_literal_numeric_member_must_have_an_initializer_when_isolatedModules_is_enabled,
+                );
+            }
+        }
+        return autoValue;
+    }
 
     function computeConstantValue(member: EnumMember): string | number | undefined {
         const isConstEnum = isEnumConst(member.parent);
         const initializer = member.initializer!;
-        const value = evaluate(initializer, member);
-        if (value !== undefined) {
-            if (isConstEnum && typeof value === "number" && !isFinite(value)) {
+        const result = evaluate(initializer, member);
+        if (result.value !== undefined) {
+            if (isConstEnum && typeof result.value === "number" && !isFinite(result.value)) {
                 error(
                     initializer,
-                    isNaN(value) ?
+                    isNaN(result.value) ?
                         Diagnostics.const_enum_member_initializer_was_evaluated_to_disallowed_value_NaN :
                         Diagnostics.const_enum_member_initializer_was_evaluated_to_a_non_finite_value,
                 );
             }
+            else if (getIsolatedModules(compilerOptions) && typeof result.value === "string" && !result.syntactic) {
+                error(
+                    initializer,
+                    Diagnostics.A_string_member_initializer_in_a_enum_declaration_can_only_use_constant_expressions_when_isolatedModules_is_enabled,
+                );
+            }
         }
         else if (isConstEnum) {
             error(initializer, Diagnostics.const_enum_member_initializers_must_be_constant_expressions);
@@ -45070,13 +45087,23 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         else {
             checkTypeAssignableTo(checkExpression(initializer), numberType, initializer, Diagnostics.Type_0_is_not_assignable_to_type_1_as_required_for_computed_enum_member_values);
         }
-        return value;
+        return result.value;
+    }
+
+    interface EvaluationResult<T> {
+        value: T;
+        syntactic: boolean;
     }
 
-    function evaluate(expr: Expression, location?: Declaration): string | number | undefined {
+    function evaluate(expr: Expression, location?: Declaration): EvaluationResult<string | number | undefined> {
+        let syntactic = true;
+        let value = evaluateInternal(expr, location);
+        return {value, syntactic};
+
+        function evaluateInternal(expr: Expression, location?: Declaration): string | number | undefined {
             switch (expr.kind) {
                 case SyntaxKind.PrefixUnaryExpression:
-                const value = evaluate((expr as PrefixUnaryExpression).operand, location);
+                    const value = evaluateInternal((expr as PrefixUnaryExpression).operand, location);
                     if (typeof value === "number") {
                         switch ((expr as PrefixUnaryExpression).operator) {
                             case SyntaxKind.PlusToken:
@@ -45089,8 +45116,8 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     }
                     break;
                 case SyntaxKind.BinaryExpression:
-                const left = evaluate((expr as BinaryExpression).left, location);
-                const right = evaluate((expr as BinaryExpression).right, location);
+                    const left = evaluateInternal((expr as BinaryExpression).left, location);
+                    const right = evaluateInternal((expr as BinaryExpression).right, location);
                     if (typeof left === "number" && typeof right === "number") {
                         switch ((expr as BinaryExpression).operatorToken.kind) {
                             case SyntaxKind.BarToken:
@@ -45131,13 +45158,16 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 case SyntaxKind.NoSubstitutionTemplateLiteral:
                     return (expr as StringLiteralLike).text;
                 case SyntaxKind.TemplateExpression:
-                return evaluateTemplateExpression(expr as TemplateExpression, location);
+                    const templateResult = evaluateTemplateExpression(expr as TemplateExpression, location);
+                    syntactic &&= templateResult.syntactic;
+                    return templateResult.value;
                 case SyntaxKind.NumericLiteral:
                     checkGrammarNumericLiteral(expr as NumericLiteral);
                     return +(expr as NumericLiteral).text;
                 case SyntaxKind.ParenthesizedExpression:
-                return evaluate((expr as ParenthesizedExpression).expression, location);
+                    return evaluateInternal((expr as ParenthesizedExpression).expression, location);
                 case SyntaxKind.Identifier: {
+                    syntactic = false;
                     const identifier = expr as Identifier;
                     if (isInfinityOrNaNString(identifier.escapedText) && (resolveEntityName(identifier, SymbolFlags.Value, /*ignoreErrors*/ true) === getGlobalSymbol(identifier.escapedText, SymbolFlags.Value, /*diagnostic*/ undefined))) {
                         return +(identifier.escapedText);
@@ -45145,6 +45175,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     // falls through
                 }
                 case SyntaxKind.PropertyAccessExpression:
+                    syntactic = false;
                     if (isEntityNameExpression(expr)) {
                         const symbol = resolveEntityName(expr, SymbolFlags.Value, /*ignoreErrors*/ true);
                         if (symbol) {
@@ -45154,13 +45185,14 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                             if (isConstantVariable(symbol)) {
                                 const declaration = symbol.valueDeclaration;
                                 if (declaration && isVariableDeclaration(declaration) && !declaration.type && declaration.initializer && (!location || declaration !== location && isBlockScopedNameDeclaredBeforeUse(declaration, location))) {
-                                return evaluate(declaration.initializer, declaration);
+                                    return evaluateInternal(declaration.initializer, declaration);
                                 }
                             }
                         }
                     }
                     break;
                 case SyntaxKind.ElementAccessExpression:
+                    syntactic = false;
                     const root = (expr as ElementAccessExpression).expression;
                     if (isEntityNameExpression(root) && isStringLiteralLike((expr as ElementAccessExpression).argumentExpression)) {
                         const rootSymbol = resolveEntityName(root, SymbolFlags.Value, /*ignoreErrors*/ true);
@@ -45176,6 +45208,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             }
             return undefined;
         }
+    }
 
     function evaluateEnumMember(expr: Expression, symbol: Symbol, location: Declaration) {
         const declaration = symbol.valueDeclaration;
@@ -45190,17 +45223,19 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         return getEnumMemberValue(declaration as EnumMember);
     }
 
-    function evaluateTemplateExpression(expr: TemplateExpression, location?: Declaration) {
-        let result = expr.head.text;
+    function evaluateTemplateExpression(expr: TemplateExpression, location?: Declaration): EvaluationResult<string | undefined> {
+        let value = expr.head.text;
+        let syntactic = true;
         for (const span of expr.templateSpans) {
-            const value = evaluate(span.expression, location);
-            if (value === undefined) {
-                return undefined;
+            const evaluateResult = evaluate(span.expression, location);
+            syntactic &&= evaluateResult.syntactic;
+            if (evaluateResult.value === undefined) {
+                return {value: undefined, syntactic};
             }
-            result += value;
-            result += span.literal.text;
+            value += evaluateResult.value;
+            value += span.literal.text;
         }
-        return result;
+        return {value, syntactic};
     }
 
     function checkEnumDeclaration(node: EnumDeclaration) {
```

</details>

andrewbranch:
Yeah, sorry for the delay. I was holding off close to the 5.4 beta when I noticed that there would be significant merge conflicts with #53463, but it turned out that didn’t go into the beta either.

src/compiler/checker.ts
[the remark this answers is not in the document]
andrewbranch: Coming back to this. Maybe I’m still confused.
> ```ts
> import {BAR} from './bar';
> enum Foo { A = `${BAR}` }
> ```
> This emits `Foo["A"] = "bar";` if the `./bar` module is known. But with `ts.transpileModule` it emits `Foo[Foo["A"] = `${BAR}`] = "A";`.

I think the `transpileModule` output is bugged. Only numeric enum values are supposed to get reverse mappings, and you can syntactically tell that this is not a numeric value.

Although we typically don’t care much what emits from `tsc` when errors are present, fixing the `transpileModule` output here would probably involve “fixing” the full emit in the case where `BAR` doesn’t have a string literal type:

```ts
const BAR = 2..toFixed(0);

enum Foo {
    A = `${BAR}`
}
```

This is already an error today, because “computed” enum members must be numeric, and this has a string type. But the emit includes a reverse mapping:

```js
Foo[Foo["A"] = `${BAR}`] = "A";
```

I suspect this reverse mapping is produced for the same reason it’s produced in the check-error-free code transpiled with `ts.transpileModule`. The logic must be “I can’t evaluate this out to a known string value, so it’s considered computed, so it must be numeric, so it gets a reverse mapping. Any inconsistencies in this logic will be covered by a check error.” But that’s not true when the only reason a known string value couldn’t be determined was the fact that we’re transpiling.

If we change this logic to never produce a reverse mapping when the value can’t be fully statically determined but _can_ be statically determined to be a string, it feels like we can loosen the restrictions here.

Does that make sense?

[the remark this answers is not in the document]
frigus02: Yes, I think I can follow. It'd be nice to allow the example above, because with full type info we can determine a known string value.

> If we change this logic to never produce a reverse mapping when the value can’t be fully statically determined but can be statically determined to be a string, it feels like we can loosen the restrictions here.

Do we really loosen the restrictions if we do that? I can't think of one right now.

I added a commit to the end of the PR that removes the reverse mapping in case the enum initializer is syntactically a string. It didn't change the emit of any tests. That might be expected though because most tests have full type information?

How would you like to proceed? Would you like to discuss this internally?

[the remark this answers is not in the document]
andrewbranch: > Do we really loosen the restrictions if we do that? I can't think of one right now.

Yes, as seen when the error in the enumWithNonLiteralStringInitializer test went away, because now it’s possible for `ts.transpileModule` and `tsc` to produce behaviorally equivalent emit:

```ts
// tsc:
var A;
(function (A) {
    A["a"] = "2";
})(A || (A = {}));

// transpileModule:
var A;
(function (A) {
    A["a"] = `${helpers_1.foo}`;
})(A || (A = {}));
```

We would just have to ensure that the `transpileModule` emit is updated to this.

> How would you like to proceed? Would you like to discuss this internally?

Yeah, I’ll bring this to a design meeting soon.

frigus02:
Hi @andrewbranch. Congrats to the TS 5.4 release! Do you think we can merge some version of this PR in time for 5.5? 😊

frigus02:
Thank you for #57686, Andrew. I update the PR and it's much smaller now.

src/compiler/utilities.ts
```diff
@@ -10657,3 +10657,17 @@ export function isSyntacticallyString(expr: Expression): boolean {
     }
     return false;
 }
+
+/** @internal */
+export function isSyntacticallyNumeric(expr: Expression): boolean {
```
andrewbranch, line 10662: Nit: I think the implementation of this function is correct, but the name is not. _Unlike_ what we settled on with `isSyntacticallyString`, what matters here _is_ actually that the value is evaluatable to a constant (with no extra-file info), because the emit feature being protected is the ability for the next member value to be equal to this member value plus one. So for example, the expression `+variableOfUnknownValue` is “syntactically numeric,” but does not and should not pass this function, because we don’t know the constant value.

frigus02: Good point. I suppose the "syntactic" part is still important because we don't want to allow sth like this in isolatedModules:

```ts
// a.ts
export const a = 1;
// b.ts
import {a} from './a';
enum Foo {A = a, B}
```

How about `syntacticallyEvaluatesToConstantNumber` or `isSyntacticallyNumericConstant`?

Given that this is different from `isSyntacticallyString`, would you prefer the function to live in the checker?

andrewbranch: I’m fine with either of those names, and I think if it doesn’t need to be used outside the checker, it can go in the checker.

src/compiler/diagnosticMessages.json
```diff
@@ -7976,5 +7976,13 @@
     "'await using' statements cannot be used inside a class static block.": {
         "category": "Error",
         "code": 18054
+    },
+    "A string member initializer in a enum declaration can only use constant expressions when 'isolatedModules' is enabled.": {
```
andrewbranch, line 7980: Paraphrasing a suggestion from @DanielRosenwasser:

```suggestion
    "'{0}' has a string type, but must have syntactically recognizable string syntax when 'isolatedModules' is enabled.": {
```

We could totally do a quick fix to wrap it in a template literal but I think this error is going to be so niche that I’m not sure it’s worth it.

andrewbranch:
@typescript-bot test top200

src/compiler/diagnosticMessages.json
frigus02: I like the suggested error message. But what do I substitute `{0}` with? The initializer can be a relatively complex expression. Is there a utility function to stringify that? Or should I use the enum member name or even the computed value?

andrewbranch: I forgot to include that; Daniel’s suggestion was `'EnumName.MemberName'`

frigus02: Sounds good. Do we want to handle member names that are not valid identifier? E.g.:

```ts
import {bar} from './bar';
enum Foo { ['not an identifier'] = bar }
// 'Foo.not an identifier' has a string type, ...
// 'Foo["not an identifier"]' has a string type, ...
```

frigus02: I updated the error message and ignored the non-identifier case for now, because it seems like an extremely rare case. That means it would print `'Foo.not an identifier'` in the error message.

andrewbranch (approved):
Thank you for your patience @frigus02!

jakebailey (approved):

PR #56742, opened 2023-12-11
When relating a deferred index type over a mapped type on the source side ...

### LINKED ISSUE #56718: TypeScript emits d.ts file that is invalid for complex mapped type.
### 🔎 Search Terms

d.ts emit type error

- This is a compile error when consuming a d.ts file emitting by the same version and configuration of TypeScript
- This repros in all versions I have tested (5.0.6, 5.1.6 and 5.3.2)
- Earlier version error when compiling the input file, so they could not run this scenario.

### ⏯ Playground Link

https://www.typescriptlang.org/play?module=1#code/KYDwDg9gTgLgBAMwK4DsDGMCWEV2CpAWwDEoJCBlGKTFAcwGcAeNHB+AWWEICNgoGeEDHwATQVGABDUTgA2ATzjsa9ANoBdAHwAKAJRwA3gCg4cGArDA4ABSmwAjHAC8R02bhqA1sCW04PgoQCHBcvPyCUoJhfALevhpCIijiytS0dO4ecAD8odyxDPEKGlkeAFxwKMAAbvwalTERxRoA3O4Avu1mFla29jAATC5GnjZkVrB+uIHB-Y4N8zAOauMQkxaJXe6SMEhQuADyPABWwBgAdGiSUiI6BHJyBlFLg+0dxkA

### 💻 Code

```ts
export function enumFromStrings<const Members extends readonly string[]>() {
  type Part1 = {
    [key in keyof Members as Members[key] extends string
      ? Members[key]
      : never]: Members[key];
  };
  type Part2 = { [Property in keyof Part1]: Part1[Property] };
  return Object.create(null) as Part2;
}
```

### 🙁 Actual behavior

Emits d.ts:

```ts
export declare function enumFromStrings<const Members extends readonly string[]>(): { [Property in keyof { [key in keyof Members as Members[key] extends string ? Members[key] : never]: Members[key]; }]: { [key in keyof Members as Members[key] extends string ? Members[key] : never]: Members[key]; }[Property]; };
```

Which errors:
```
Type 'Property' cannot be used to index type '{ [key in keyof Members as Members[key] extends string ? Members[key] : never]: Members[key]; }'.
```

[In playground](https://www.typescriptlang.org/play?ssl=3&ssc=1&pln=1&pc=1#code/KYDwDg9gTgLgBAE2AYwDYEMrDgMwK4B2yMAlhAXMAXgLYBiUENAyjFCQQOYDOAPMuW7wAssBoAjYFG6UQMKghlZ0CcqgCecIey4BtALoA+ABQBKAFxwA3nF0AFRmCkxNHOAGtg6iDmu3PrhQBPnCiElIy6DJhktK6Afqy8gSKWmwcnHAA-KFisdzxXomWBMAAblL6ljERher6ANxwAL5VfnVwbsG+NdJwUbnhcQlJCjLaGdmD+XXFcKUVUG29BQlNzfaOzvXrDQBQe0A)

### 🙂 Expected behavior

I expect TypeScript to either error when trying to emit a d.ts file, or generate a valid d.ts file it can consume without error.

I have a workaround for the case where I hit this, so I'm mainly reporting this incase it helps TypeScript development.

weswigham:
actually compare against the mapped type's apparent keys.

FIxes #56718 - the emit is fine (all the anonymous/local types are inlined and duplicated like you'd expect), our relationship logic was just too strict when relating deferred index types of mapped types, so in the emit, when the `keyof {big anonymous mapped type}` got duplicated into two places, we failed to recognize them as the same type.

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -22142,6 +22142,18 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             return result;
         }
 
+        function getApparentMappedTypeKeys(nameType: Type, targetType: MappedType) {
+            const modifiersType = getApparentType(getModifiersTypeFromMappedType(targetType));
+            const mappedKeys: Type[] = [];
+            forEachMappedTypePropertyKeyTypeAndIndexSignatureKeyType(
+                modifiersType,
+                TypeFlags.StringOrNumberLiteralOrUnique,
+                /*stringsOnly*/ false,
+                t => void mappedKeys.push(instantiateType(nameType, appendTypeMapping(targetType.mapper, getTypeParameterFromMappedType(targetType), t))),
+            );
+            return getUnionType(mappedKeys);
+        }
+
         function structuredTypeRelatedToWorker(source: Type, target: Type, reportErrors: boolean, intersectionState: IntersectionState, saveErrorInfo: ReturnType<typeof captureErrorCalculationState>): Ternary {
             let result: Ternary;
             let originalErrorInfo: DiagnosticMessageChain | undefined;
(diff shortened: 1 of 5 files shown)


weswigham:
@typescript-bot test this
@typescript-bot run dt
@typescript-bot test top100
@typescript-bot pack this

jakebailey:

jakebailey (approved):

src/compiler/checker.ts
```diff
@@ -22507,9 +22512,23 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 }
             }
             else if (sourceFlags & TypeFlags.Index) {
-                if (result = isRelatedTo(keyofConstraintType, target, RecursionFlags.Source, reportErrors)) {
+                const isDeferredMappedIndex = shouldDeferIndexType((source as IndexType).type, (source as IndexType).indexFlags) && getObjectFlags((source as IndexType).type) & ObjectFlags.Mapped;
+                if (result = isRelatedTo(keyofConstraintType, target, RecursionFlags.Source, reportErrors && !isDeferredMappedIndex)) {
```
gabritto, line 22516: I don't know that it matters a lot, but shouldn't `!isDeferredMappedIndex` come first to avoid calling `isRelatedTo` when the source is in fact a deferred mapped index? Or is it still useful to call `isRelatedTo` in this case?

weswigham: That's not after the call, it's *in* the call, modifying the `reportErrors` parameter to it.

PR #56743, opened 2023-12-12
fix(56737): unique symbol quick info is different in JS than TS

### LINKED ISSUE #56737: unique symbol quick info is different in JS than TS
Discovered by @8HoLoN at https://github.com/microsoft/TypeScript/issues/48231#issuecomment-1849865024

The quick info behaviour is still different for me between .js and .ts:

```ts
const typescript: unique symbol = Symbol()
typescript
// ^? prints 'const typescript: typeof typescript'
```

```js
/** @type {unique symbol} */
const javascript = Symbol();
javascript
// ^? prints `const javascript: symbol`
```

However, the tests from https://github.com/microsoft/TypeScript/pull/48270 still pass and show that the emitted .d.ts is indeed `unique symbol`, just like typescript. So it's likely that #48270 is an incomplete fix and that quick info still needs additional code.

### LINKED ISSUE #56800: `unique symbol` type annotation not respected in JS
```js
/** @type {unique symbol} */
const s = Symbol()
/** @typedef {{ [s]: boolean }} A - should be OK in JS */
/** @type {unique symbol} */
const v = Symbol()
/** @typedef {{ [key: v]: boolean }} D - should error in JS */
```

compare to TS

```ts
const t: unique symbol = Symbol()
type B = { [t]: boolean } // should be OK in TS (and is)
const u: unique symbol = Symbol()
type C = { [key: u]: boolean } // should error in TS (and does)
```

It appears as if `s` and `v` have type `symbol` not `unique symbol`.

https://www.typescriptlang.org/play?filetype=js#code/MYewdgzgLgBFBcMCuYCWBHJBTGECeAtgEYgA2MAvDAMqEmkAUAlALABQUeADjgEKUwA3jADaUALqISZLAEMwMAL4wA9CtwALEElIATGERwB5ANIxUCgCrUYDeftQRWbUJFhJEKDNlx0yA2mIyZnZOHhgAYQFhEQBrLDxPSQMQGXklVXUILR19LAAnfJB88ysbOzB9XRAsJ1DuHABRAS9MHHwg8jVNbT0YAqKSixhrW3sYatrndhUAKlmYAAEwnEFWnw76ZVmVdldoXAC-Rmc5heWG3SwAMyEYiGTpUjkFRWUAQRgAWh7cg2MzMMAFI2HYzeZLFZCdbtY7bXYucAHABuR06ITYZ0hlxud1E8USMGRj1Sz3SbxgABFvr8+gNiqUYCCYGC2ODzisrrc1mg2r5OsoAGI07K9PKFBnA8rjSZOFkqIA

a-tarasyuk:
Fixes #56737
Fixes #56800

### DIFF
--- a/src/compiler/utilities.ts
+++ b/src/compiler/utilities.ts
@@ -2685,7 +2685,7 @@ export function isCommonJsExportPropertyAssignment(node: Node) {
 export function isValidESSymbolDeclaration(node: Node): boolean {
     return (isVariableDeclaration(node) ? isVarConst(node) && isIdentifier(node.name) && isVariableDeclarationInVariableStatement(node) :
         isPropertyDeclaration(node) ? hasEffectiveReadonlyModifier(node) && hasStaticModifier(node) :
-        isPropertySignature(node) && hasEffectiveReadonlyModifier(node)) || isCommonJsExportPropertyAssignment(node);
+        isPropertySignature(node) && hasEffectiveReadonlyModifier(node)) || isCommonJsExportPropertyAssignment(node) || isJSDocTypeExpression(node);
 }
 
 /** @internal */
--- a/src/compiler/utilities.ts
+++ b/src/compiler/utilities.ts
@@ -2685,7 +2685,7 @@ export function isCommonJsExportPropertyAssignment(node: Node) {
 export function isValidESSymbolDeclaration(node: Node): boolean {
     return (isVariableDeclaration(node) ? isVarConst(node) && isIdentifier(node.name) && isVariableDeclarationInVariableStatement(node) :
         isPropertyDeclaration(node) ? hasEffectiveReadonlyModifier(node) && hasStaticModifier(node) :
-        isPropertySignature(node) && hasEffectiveReadonlyModifier(node)) || isCommonJsExportPropertyAssignment(node) || isJSDocTypeExpression(node);
+        isPropertySignature(node) && hasEffectiveReadonlyModifier(node)) || isCommonJsExportPropertyAssignment(node);
 }
 
 /** @internal */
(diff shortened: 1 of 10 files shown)


tests/baselines/reference/quickInfoUniqueSymbolJsDoc.baseline
```diff
@@ -0,0 +1,66 @@
+// === QuickInfo ===
+=== /tests/cases/fourslash/./a.js ===
+// /** @type {unique symbol} */
+// const foo = Symbol();
+// foo
+// ^^^
+// | ----------------------------------------------------------------------
+// | const foo: symbol
```
sandersn, line 8: in typescript quickinfo says `const foo: typeof foo`, not `const foo: symbol`

src/compiler/utilities.ts
```diff
@@ -2685,7 +2685,7 @@ export function isCommonJsExportPropertyAssignment(node: Node) {
 export function isValidESSymbolDeclaration(node: Node): boolean {
     return (isVariableDeclaration(node) ? isVarConst(node) && isIdentifier(node.name) && isVariableDeclarationInVariableStatement(node) :
         isPropertyDeclaration(node) ? hasEffectiveReadonlyModifier(node) && hasStaticModifier(node) :
-        isPropertySignature(node) && hasEffectiveReadonlyModifier(node)) || isCommonJsExportPropertyAssignment(node);
+        isPropertySignature(node) && hasEffectiveReadonlyModifier(node)) || isCommonJsExportPropertyAssignment(node) || isJSDocTypeExpression(node);
 }
```
sandersn, line 2689: can you add a compiler test from #56800:

```js
/** @type {unique symbol} */
const foo = Symbol()
/** @typedef {{ [foo]: boolean }} Alias */
/** @typedef {{ [key: foo] boolean }} BadAlias */
```

The first typedef should work and the second should error, with the same error as Typescript has.

sandersn (changes requested):
Looks like this half-solves the problem.

sandersn (approved):

Issue #56738, opened 2023-12-11
Abstract property specified on multiple classes loses modifier after intersection


ssalbdivad:
### 🔎 Search Terms

abstract property class intersection InstanceType typeof mixin trait

### 🕗 Version & Regression Information

- This is the behavior in every version I tried

### ⏯ Playground Link

https://tsplay.dev/mb1k9N

### 💻 Code

```ts
abstract class A {
	abstract a(): number
}

abstract class A2 {
	abstract a(): number
}

abstract class B  {
	abstract b(): number
}

declare const Base: abstract new () => A & B

// correctly has type error requiring the implementation of 'a' and 'b'
class Foo extends Base {}

declare const FalsePositiveBase: abstract new () => A & A2

// this sould be an error
// if an abstract property appears on multiple classes, it is not treated as abstract
class Bar extends FalsePositiveBase {}


// can workaround this as follows:
declare const WorkaroundBase: abstract new () => InstanceType<typeof A & typeof A2>

// now gives the correct error that 'a' is missing
class Baz extends WorkaroundBase {}
```

Seems like there must be some flag that associates the abstractness of the original class instance when it is used as a class instance like that, but there is a logic error that checks if the prop exists on other intersection members and erases the modifier, even if all of them are also abstract.

### 🙁 Actual behavior

Behavior for multiple props of the same name was inconsistent with behavior for props with differing names

### 🙂 Expected behavior

Intersecting multiple props of the same name that all of the `abstract` modifier should preserve it

PR #56751, opened 2023-12-12
Fixed an issue with intersected abstract properties not requiring to be implemented

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/56738


src/compiler/checker.ts
```diff
@@ -1946,6 +1946,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     var subtypeReductionCache = new Map<string, Type[]>();
     var decoratorContextOverrideTypeCache = new Map<string, Type>();
     var cachedTypes = new Map<string, Type>();
+    var cachedUnionOrIntersectionPropertySymbols = new Map<string, Symbol>();
```
Andarist, line 1949: maybe this could be named like `cachedSyntheticPropertySymbols`?

src/compiler/checker.ts
```diff
@@ -14597,6 +14599,11 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             }
         }
         const props = propSet ? arrayFrom(propSet.values()) : [singleProp];
+        const id = `${checkFlags & CheckFlags.ReadPartial ? 'p' : ''}${skipObjectFunctionPropertyAugment ? 's' : ''}${isUnion ? 'u' : 'i'}${getSymbolListId(props)}`;
```
Andarist, line 14602: This id is a little bit ad-hoc so it might be missing some important. To the best of my understanding (and that might not be complete here) this sounds reasonably well - unless there is just some fundamental reason why this is blatantly wrong that I missed.

src/compiler/checker.ts
```diff
@@ -14597,6 +14599,11 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             }
         }
         const props = propSet ? arrayFrom(propSet.values()) : [singleProp];
+        const id = `${checkFlags & CheckFlags.ReadPartial ? 'p' : ''}${skipObjectFunctionPropertyAugment ? 's' : ''}${isUnion ? 'u' : 'i'}${getSymbolListId(props)}`;
+        const cached = cachedUnionOrIntersectionPropertySymbols.get(id);
+        if (cached) {
+            return cached;
+        }
```
Andarist, line 14606: It's worth explaining why I'm even caching those...

The problem that it solves is that [`getTypeWithThisArgument` call](https://github.dev/microsoft/TypeScript/blob/efc9c065a2caa52c5bebd08d730eed508075a78a/src/compiler/checker.ts#L13075-L13076) for the `Bar1` creates a **new** intersection type, one that doesn't have any reference (through `.target` or similar) to the "original" intersection (the base type).

So `createUnionOrIntersectionProperty` creates 2 separate synthetic properties for `a` for both the base type and the derived type (even though they both refer to the same symbols, declarations, etc). That, in turn, fails [`derived === base` check](https://github.dev/microsoft/TypeScript/blob/efc9c065a2caa52c5bebd08d730eed508075a78a/src/compiler/checker.ts#L44697-L44700) that is responsible for checking if the property was implemented or not.

An alternative solution to this caching could be extending this `derived === base` with some extra conditions.

src/compiler/checker.ts
```diff
@@ -13033,8 +13033,16 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
-            const types = sameMap((type as IntersectionType).types, t => getTypeWithThisArgument(t, thisArgument, needApparentType));
-            return types !== (type as IntersectionType).types ? getIntersectionType(types) : type;
+            const intersectionType = type as IntersectionType;
+            const types = sameMap(intersectionType.types, t => getTypeWithThisArgument(t, thisArgument, needApparentType));
+            if (types === intersectionType.types) {
+                return type;
+            }
+            const result = getIntersectionType(types);
+            if (result.flags & TypeFlags.Intersection) {
+                (result as IntersectionType).withThisArgumentTarget = intersectionType;
```
Andarist, line 13043: The problem that it solves is that [`getTypeWithThisArgument` call](https://github.dev/microsoft/TypeScript/blob/efc9c065a2caa52c5bebd08d730eed508075a78a/src/compiler/checker.ts#L13075-L13076) for the `Bar1` creates a **new** intersection type, one that doesn't have any reference (through `.target` or similar) to the "original" intersection (the base type).

So `createUnionOrIntersectionProperty` creates 2 separate synthetic properties for `a` for both the base type and the derived type (even though they both refer to the same symbols, declarations, etc). That, in turn, fails [`derived === base` check](https://github.dev/microsoft/TypeScript/blob/efc9c065a2caa52c5bebd08d730eed508075a78a/src/compiler/checker.ts#L44697-L44700) that is responsible for checking if the property was implemented or not.

This introduces some kind of a link between those types so I can trace this information back through the symbol links.

It's not a super elegant solution but it's the best I was able to figure out. I've tried to cache those property symbols and reuse them between both of those types but that led to a lot of problems. An alternative solution would be to make the mentioned `derived === base` check smarter but I also failed to figure out how to accurately compare 2 symbols like that.

PR #50586, opened 2022-09-01
Improve binding element type inference using `CheckMode` 

babakks:
Fixes #49989 
Alternative to PR #50241.

Here, to prevent false circular-relationship on binding elements initialized with their siblings (like the code segment below), we've utilized `CheckMode` values other than `Normal` as an indicator of suppressing type-checking errors.

```ts
const [a, b = a] = [0];
// or
const [a, b = a + 1] = [0];
```

~One gotcha in this PR was to revert the updated links (i.e., `links.type`) of a symbol. Because once a circular relationship error was encountered the type of the symbol is inferred to be `any` and then somewhere in the call stack, the symbol's associated `links.type` would be assigned to that returned `any` type. So, in cases other than `CheckMode.Normal` we'd replace back the `links.type` old value.~ **(Sorry, that was not up-to-date)**

The gotcha is that I had to suppress updating the symbol's `links.type` (to `any` type) for other than `CheckMode.Normal` cases, because that would block further efforts to find the correct type for the symbol. Now, I'm thinking maybe it's better to also check if the returned type is `any` and then do the suppression. I didn't find any better approach than this, so let me know if there's one:

```ts
function getTypeOfVariableOrParameterOrProperty(symbol: Symbol, checkMode?: CheckMode): Type {
    const links = getSymbolLinks(symbol);
    let result = links.type;
    if (!result) {
        result = getTypeOfVariableOrParameterOrPropertyWorker(symbol, checkMode);
        // For a contextually typed parameter it is possible that a type has already
        // been assigned (in assignTypeToParameterAndFixTypeParameters), and we want
        // to preserve this type.
        if (!links.type && !(checkMode && checkMode !== CheckMode.Normal)) {
            links.type = result;
        }
    }
    return result;
}
```

## Further issues

Although this PR resolves common cases like the one above, but in cases like below false errors still occur. I didn't dig more to fix such issues because handling them would, in my opinion, clutter the type-checker without significant value in return due to the rarity of such usages. Please correct me if I'm wrong.

```ts
const [a, b = (() => 1 + a)()] = [0];
```

PR #56753, opened 2023-12-12
Improve binding element type inference using CheckMode (rebased)

### LINKED ISSUE #49989: Using variable defined by array destructuring assignment in default value of variables defined afterwards.
# Bug Report

### 🔎 Search Terms

Array destructuring assignment default value 7022

Exists in 3.3.3 - 4.8.0-beta
- This is the behavior in every version I tried, and I reviewed the FAQ for entries about this behaviour

### ⏯ Playground Link

[Playground link with relevant code](https://www.typescriptlang.org/play?target=8&ts=4.8.0-beta#code/MYewdgzgLgBA2gQwDQwEYwLwwTA1DARgF1N4AmIgbgFgAoUSEAGwFMA6JkAcwAoADBAC4YAEgDeCAL4pUw8akl8AlDVpA)

### 💻 Code

```ts
const [a, b = a + 1] = [2];
console.log(`a: ${a}, b: ${b}`);
```

### 🙁 Actual behavior

Types of `a` and `b` inferred as `any`, or if `noImplicitAny` enabled, shows error: `'a' implicitly has type 'any' because it does not have a type annotation and is referenced directly or indirectly in its own initializer.(7022)`

### 🙂 Expected behavior

Types of both `a` and `b` should be inferred as `number`.

babakks:
Fixes #49989
Rebased/updated version of PR #50586 

This is an updated version of my old PR (#50586), which is now rebased on top of the latest `main` branch. I've applied @weswigham's comment ([here][comment]) on this PR.

[comment]: https://github.com/microsoft/TypeScript/pull/50586#discussion_r1139190623


src/compiler/checker.ts
```diff
@@ -11634,6 +11634,10 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             if (symbol.flags & SymbolFlags.ValueModule && !(symbol.flags & SymbolFlags.Assignment)) {
                 return getTypeOfFuncClassEnumModule(symbol);
             }
+
+            if (checkMode !== CheckMode.Normal && checkMode) {
+                return anyType;
```
Andarist, line 11639: I **think** that there is a non-zero chance that [this PR](https://github.com/microsoft/TypeScript/pull/56429) will get picked up at some point (cc @jakebailey ). If the team decides to go with it then this should return `errorType`:
```suggestion
                return errorType;
```

I think that there is no harm in returning that here already now (unless proven otherwise)

n9:
@babakks Will this PR also solve https://github.com/microsoft/TypeScript/issues/49989#issuecomment-1852694486?

babakks:
@n9 Yes. I just checked the test output and it's now inferring the right type:

```
const { a, b = a } = o ?? {};
>a : number
>b : number
>a : number
>o ?? {} : { a?: number; b?: number; }
>o : { a?: number; b?: number; }
>{} : {}
```

n9:
Great! Thank you! Hopefully someone will merge it soon.

babakks:
@n9 Sorry, but seems there's a problem. Actually, the type of `a` (and `b`) should be detected as `number | undefined`, but it's inferred as `number`, which is not accurate.

Anyway, thanks for bringing this useful test case. I can use it as an extra check for the PR.

n9:
Yes, you are right. 😎

src/compiler/checker.ts
babakks: Thanks for the comment. Since, it's an error (of course, not a circularity one but still), I think you're right about changing the return value from `anyType` to `errorType`. 

Applied in the last commit.

babakks:
@n9 While working on the PR, I noticed it's not actually wrong. So, I've included your test case in the new test files.

n9:
@babakks What is not wrong?

babakks:
@n9 I wanted to make sure that the type of `a` in the code example below is inferred as `number | undefined` (not just `number`). Now it's working (note the hover says `number | undefined`):

hover

babakks:
@sandersn @Andarist I think this is ready for review now.

By the way, I always had a hard time debugging TypeScript checker code, because of its complicated nature and repeate/recursive-like calls that you couldn't just put ordinary (non-conditional) breakpoints. So, I made this simple VS Code extension that helps with setting call-stack bound breakpoints (which will only hit, when you're at a specific call-trace). I think it might be useful for you guys too. Here's the link to it:

https://marketplace.visualstudio.com/items?itemName=babakks.vscode-js-custom-breakpoint

n9:
@babakks

Hmm, I just do not understand `>a : number` notation. (Why there is no `| undefined`?)

babakks:
@n9 Yeah, that was my concern, too. But after checking other (unrelated) test cases noticed this is not necessarily wrong. That notation is used in baseline `.types` files that reflect the resolved types for a test case file.

src/compiler/checker.ts
```diff
@@ -11574,24 +11574,24 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
         if (!links.type) {
-            const type = getTypeOfVariableOrParameterOrPropertyWorker(symbol);
+            const type = getTypeOfVariableOrParameterOrPropertyWorker(symbol, checkMode);
             // For a contextually typed parameter it is possible that a type has already
             // been assigned (in assignTypeToParameterAndFixTypeParameters), and we want
             // to preserve this type. In fact, we need to _prefer_ that type, but it won't
             // be assigned until contextual typing is complete, so we need to defer in
             // cases where contextual typing may take place.
-            if (!links.type && !isParameterOfContextSensitiveSignature(symbol)) {
+            if (!links.type && !isParameterOfContextSensitiveSignature(symbol) && !(checkMode !== CheckMode.Normal && checkMode)) {
```
weswigham, line 11586:
```suggestion
            if (!links.type && !isParameterOfContextSensitiveSignature(symbol) && !checkMode) {

```

weswigham (changes requested):
Minor change and a sync with `main`, and I'll get the extended tests running for this

src/compiler/checker.ts
babakks: Applied the suggested change. Also, merged the latest `main` into the branch.

weswigham:
@typescript-bot run dt
@typescript-bot test top100
@typescript-bot pack this

weswigham (approved):

Issue #56778, opened 2023-12-14
Indexed access and object destructuring on never doesn't type error


TkDodo:
### 🔎 Search Terms

"never", "indexed access", "destruct", "property access"

### 🕗 Version & Regression Information

This is the behavior in every version I tried, and I reviewed the FAQ for entries

### ⏯ Playground Link

https://www.typescriptlang.org/play?&code/MYewdgzgLgBApgDwIYFsAOAbOAuGY4BucATjALx4CuGGMSEehJAUMwPRswACUEAtIjRxgUAcWIhSABQlDiUAJ4wA5ADMQIZTAAmIOAzAhYiAJbQY4GIqEr8RYsoB0ACgBMAZncBOAJTNEqJhwjuogrBzwCEIiJmAA5lYAFnAwEKgpJBKkycRwADQwAEaUsFA5KWZ44HDsnDz8gsKimZIwMiByiiqhWrr6VcYIZrCW1inKdiRObp6%20-sjoWADaahrKALpAA=#code/MYewdgzgLgBApgDwIYFsAOAbOAuGY4BucATjALx4CuGGMSEehJAUMwPRswACUEAtIjRxgUAcWIhSABQlDiUAJ4wA5ADMQIZTAAmIOAzAhYiAJbQY4GIqEr8RYsoB0ACgBMAZncBOAJTNEqJhwjuogrBzwCEIiJmAA5lYAFnAwEKgpJBKkycRwADQwAEaUsFA5KWZ44HDsnDz8gsKimZIwMiByiiqhWrr6VcYIZrCW1inKdiRObp6+-sjoWADaahrKALrhnACSytp0GBAgkdGlyanp8OKt5QXFZyQVBtW13LwCUU1iWW2yJF2rTQ6PTPQbDCxgKwKGwTJgOFwebx+UCQWAAbxgoRgAF9yJFAlggA

### 💻 Code

```ts
// Given an `example` variable of type `never`:

const example: never = null as never

// this errors correctly
// @ts-expect-error Property 'foo' does not exist on type 'never'.(2339)
example.foo

// expecting the same error here, but there is none
// @ts-expect-error Property 'foo' does not exist on type 'never'.(2339)
example['foo']

// I'd also expect the same error here, but there is none
// @ts-expect-error Property 'foo' does not exist on type 'never'.(2339)
const { foo } = example
```

### 🙁 Actual behavior

[...]

Issue #42999, opened 2021-02-27
Destructuring or bracket notation on `never` shouldn't be allowed


dwelle:
# Bug Report

You can destructure or use bracket notation on `never` type, but you can't access props using the dot notation. It should disallow all forms of property access.

### 🔎 Search Terms

never destructuring destructure bracket dot notation

### 🕗 Version & Regression Information

- This is the behavior in every version I tried, and I reviewed the FAQ :heavy_check_mark:
  - [3.9.7](https://www.typescriptlang.org/play?ts=3.9.7#code/MYewdgzgLgBA3jADgJxImBfGBeGAKOLAQwhjAFMA3c5ASgG4BYAKFElhTQCYd9CYSZKjVoBtAESdE4gLr0gA)
  - [4.1.5](https://www.typescriptlang.org/play?#code/MYewdgzgLgBA3jADgJxImBfGBeGAKOLAQwhjAFMA3c5ASgG4BYAKFElhTQCYd9CYSZKjVoBtAESdE4gLr0gA)
  - [Nightly](http://www.typescriptlang.org/play/?ts=Nightly#code/MYewdgzgLgBA3jADgJxImBfGBeGAKOLAQwhjAFMA3c5ASgG4BYAKFElhTQCYd9CYSZKjVoBtAESdE4gLr0gA)

### ⏯ Playground Link

[Playground link](https://www.typescriptlang.org/play?ts=4.1.5#code/MYewdgzgLgBA3jADgJxImBfGBeGAKOLAQwhjAFMA3c5ASgG4BYAKFElhTQCYd9CYSZKjVoBtAESdE4gLr0gA)

### 🙁 Actual behavior

```ts
const { prop } = ({} as never); // does not error
const prop2 = ({} as never)["prop"]; // does not error
```

### 🙂 Expected behavior

TS should not allow this, same as you can't access props using the dot notation:

```ts
const prop = ({} as never).prop; // error
```

gabritto:
In theory, when something has type `never`, it means that code is never supposed to run, because no value of type `never` exists (since `never` is the empty type). So there's theoretically nothing wrong with allowing you to do an element access or destructuring of something of type `never`, because that's also code that is never supposed to run.

[...]

PR #56780, opened 2023-12-14
Disallow more flavors of property accesses on `never` in expression space

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/42999
fixes https://github.com/microsoft/TypeScript/issues/56778

### DIFF
--- a/tests/baselines/reference/controlFlowAssignmentPatternOrder.errors.txt
+++ b/tests/baselines/reference/controlFlowAssignmentPatternOrder.errors.txt
@@ -1,0 +1,97 @@
+controlFlowAssignmentPatternOrder.ts(6,9): error TS2339: Property '1' does not exist on type 'never'.
+controlFlowAssignmentPatternOrder.ts(12,9): error TS2339: Property '0' does not exist on type 'never'.
+controlFlowAssignmentPatternOrder.ts(31,9): error TS2339: Property '1' does not exist on type 'never'.
+controlFlowAssignmentPatternOrder.ts(37,9): error TS2339: Property '0' does not exist on type 'never'.
+controlFlowAssignmentPatternOrder.ts(56,14): error TS2339: Property '1' does not exist on type 'never'.
+controlFlowAssignmentPatternOrder.ts(62,14): error TS2339: Property '0' does not exist on type 'never'.
+
+
+==== controlFlowAssignmentPatternOrder.ts (6 errors) ====
+    // https://github.com/microsoft/TypeScript/pull/41094#issuecomment-716044363
+    declare function f(): void;
+    {
+        let a: 0 | 1 = 0;
+        let b: 0 | 1 | 9;
+        [{ [(a = 1)]: b } = [9, a] as const] = [];
+            ~~~~~~~
+!!! error TS2339: Property '1' does not exist on type 'never'.
+        const bb: 0 = b;
+    }
+    {
+        let a: 0 | 1 = 1;
+        let b: 0 | 1 | 9;
+        [{ [a]: b } = [9, a = 0] as const] = [];
+            ~
+!!! error TS2339: Property '0' does not exist on type 'never'.
+        const bb: 9 = b;
+    }
+    {
+        let a: 0 | 1 = 0;
+        let b: 0 | 1 | 8 | 9;
+        [{ [(a = 1)]: b } = [9, a] as const] = [[9, 8] as const];
+        const bb: 0 | 8 = b;
+    }
+    {
+        let a: 0 | 1 = 1;
+        let b: 0 | 1 | 8 | 9;
+        [{ [a]: b } = [a = 0, 9] as const] = [[8, 9] as const];
+        const bb: 0 | 8 = b;
+    }
+    // same as above but on left of a binary expression
+    {
+        let a: 0 | 1 = 0;
+        let b: 0 | 1 | 9;
+        [{ [(a = 1)]: b } = [9, a] as const] = [], f();
+            ~~~~~~~
+!!! error TS2339: Property '1' does not exist on type 'never'.
+        const bb: 0 = b;
+    }
+    {
+        let a: 0 | 1 = 1;
+        let b: 0 | 1 | 9;
+        [{ [a]: b } = [9, a = 0] as const] = [], f();
+            ~
+!!! error TS2339: Property '0' does not exist on type 'never'.
+        const bb: 9 = b;
+    }
+    {
+        let a: 0 | 1 = 0;
+        let b: 0 | 1 | 8 | 9;
+        [{ [(a = 1)]: b } = [9, a] as const] = [[9, 8] as const], f();
+        const bb: 0 | 8 = b;
+    }
+    {
+        let a: 0 | 1 = 1;
+        let b: 0 | 1 | 8 | 9;
+        [{ [a]: b } = [a = 0, 9] as const] = [[8, 9] as const], f();
+        const bb: 0 | 8 = b;
+    }
+    // same as above but on right of a binary expression
+    {
+        let a: 0 | 1 = 0;
+        let b: 0 | 1 | 9;
+        f(), [{ [(a = 1)]: b } = [9, a] as const] = [];
+                 ~~~~~~~
+!!! error TS2339: Property '1' does not exist on type 'never'.
+        const bb: 0 = b;
+    }
+    {
+        let a: 0 | 1 = 1;
+        let b: 0 | 1 | 9;
+        f(), [{ [a]: b } = [9, a = 0] as const] = [];
+                 ~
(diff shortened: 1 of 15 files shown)


tests/baselines/reference/controlFlowAssignmentPatternOrder.errors.txt
```diff
@@ -0,0 +1,97 @@
...
+
+==== controlFlowAssignmentPatternOrder.ts (6 errors) ====
+    // https://github.com/microsoft/TypeScript/pull/41094#issuecomment-716044363
+    declare function f(): void;
+    {
+        let a: 0 | 1 = 0;
+        let b: 0 | 1 | 9;
+        [{ [(a = 1)]: b } = [9, a] as const] = [];
+            ~~~~~~~
+!!! error TS2339: Property '1' does not exist on type 'never'.
```
Andarist, line 17: I'm not entirely sure if the error here is completely desirable but it feels consistent to me with things like:
```ts
const obj = { b: '' }
let a: number
;({ a = 10 } = obj) // Property 'a' does not exist on type '{ b: string; }'.(2339)
```

What I mean is that despite the first element in this array pattern having an initializer it's the "right-side" type of the original source that is used to check for valid property accesses. 

So since we have `never[]` on the right side, its first element is actually `never` and that has no properties so disallowing those patterns here looks desirable to me.

gabritto:
@typescript-bot run DT
@typescript-bot user test this
@typescript-bot test top100
@typescript-bot perf test this

gabritto:
@typescript-bot pack this

gabritto:
@typescript-bot run DT
@typescript-bot test top100

tests/baselines/reference/controlFlowAssignmentPatternOrder.errors.txt
gabritto: This error doesn't look right to me: it says "Property '1' does not exist on type 'never'", but we should be getting property `1` of type `[9, typeof a]`, and that tuple should not have type `never`.

gabritto:
> @gabritto Here are the results of running the top-repos suite comparing `main` and `refs/pull/56780/merge`:
> 
> 
> Details
> ## [ant-design/ant-design](https://github.com/ant-design/ant-design)
> ## [backstage/backstage](https://github.com/backstage/backstage)
> ## [microsoft/vscode](https://github.com/microsoft/vscode)
> ## [tldraw/tldraw](https://github.com/tldraw/tldraw)

Of the new errors on the top repos test, most seem to be because of "defensive programming", where there is an exhaustive check on some variable followed by a branch for when it has an unexpected type. In this defensive branch, the variable has type `never`, so an element access produces an error.

Those seem to be actual errors:
https://github.com/ant-design/ant-design/blob/4704eb578b8e7caabc502a5204c2ba058ee8266c/components/tree/utils/dictUtil.ts#L22
https://github.com/ant-design/ant-design/blob/4704eb578b8e7caabc502a5204c2ba058ee8266c/components/tree/utils/dictUtil.ts#L22

gabritto:
We discussed this at design meeting today, and the conclusion was that this change is not worth the breaks it causes. As evidenced by the extended tests, most new errors are not really desirable and come from an intentional usage of `never`, i.e. something has type `never` in code that is not supposed to be reachable.
There was one error that seemed like a bug, as I pointed out above, where the code looks like `function foo(x: A & B) { ... x[key] ... }`, and `A & B` reduces to `never`.
So it seems like most times, people are using `never` intentionally and TS shouldn't error, but other times it should, and it's not immediately clear how to tell the difference.

gabritto:
We discussed this at design meeting today, and the conclusion was that this change is not worth the breaks it causes. As evidenced by the extended tests, most new errors are not really desirable and come from an intentional usage of `never`, i.e. something has type `never` in code that is not supposed to be reachable.
There was one error that seemed like a bug, as I pointed out above, where the code looks like `function foo(x: A & B) { ... x[key] ... }`, and `A & B` reduces to `never`.
So it seems like most times, people are using `never` intentionally and TS shouldn't error, but other times it should, and it's not immediately clear how to tell the difference.

jakebailey:
I unfortunately missed the meeting, but I find it a little weird that we're saying that it's okay to access any prop on never; if that were the case, then `Foo | never` should not reduce because a component of the union "has all props", but that's not what `never` does... It's the empty set of types, with which you can do nothing.

I personally feel like all of the errors this PR showed are real and that `never` is a type that's okay to have around, and that if you want to touch something that's `never` you need to cast it to something else to probe at it.

gabritto:
> It's the empty set of types, with which you can do nothing.

Since we know `never` is the empty set of values, once you know you have a value `v` of type `never`, you've reached a contradiction. And, by the principle of "from contradiction, anything follows", you can access anything on that value because anything is true about it, including statements like "`v` has property `blah`".
If we had a sound type system, at run time we'd never reach that portion of code where `v` has type `never`, so it would not matter if you tried to access a property on it.

> I unfortunately missed the meeting, but I find it a little weird that we're saying that it's okay to access any prop on never; if that were the case, then `Foo | never` should not reduce because a component of the union "has all props", but that's not what `never` does...

`Foo | never` reduces to `Foo` because you know whatever value has that type, it can't have type `never`.
But let's say we didn't reduce it and you have a value `v` of type `Foo | never`. `v` either has type `never`, in which case you can access any property on `v`, or it has type `Foo`, in which case you can only access `Foo` properties on `v`. In both of those cases, you can access the `Foo` properties on `v`. This is consistent with saying `v` has type `Foo`.

jakebailey:
> `Foo | never` reduces to `Foo` because you know whatever value has that type, it can't have type `never`.
> But let's say we didn't reduce it and you have a value `v` of type `Foo | never`. `v` either has type `never`, in which case you can access any property on `v`, or it has type `Foo`, in which case you can only access `Foo` properties on `v`. In both of those cases, you can access the `Foo` properties on `v`. This is consistent with saying `v` has type `Foo`.

Ah, that is true; I had it flipped in my mind; obviously if you have a union of two types, you can only access what they have in common.

Andarist:
I'm OK with the judgment itself but this inconsistency is killing me:
```ts
declare const foo: never

foo.prop // error ?????
foo["prop"] // ok
const { prop } = foo // ok
```

It feels just wildly inconsistent and hard to explain to people. How this difference in behavior would be explained in the docs?

I didn't yet have the time to review carefully everything reported by the extended test suite but I'd definitely call most of those failures improvements. I feel like the only use case that was said to be valid here is the defensive programming case. 

And to quote @jakebailey:
> if you want to touch something that's never you need to cast it to something else to probe at it.

It would at least give some visual indicator that something funky is going on there. Those would also often appear in the codepaths that would be clearly associated with such defensive programming fallbacks (close to thrown error messages, `console.warn`s etc). I feel like in those contexts people wouldn't mind *at all* the cast.

On the other hand, I'm quite worried - as the user - that `never` might accidentally sneak past me like this and I won't notice it at all. For instance, I doubt that [this](https://github.com/ant-design/ant-design/blob/4704eb578b8e7caabc502a5204c2ba058ee8266c/components/tree/utils/dictUtil.ts#L22) intentionally relies on `never` this way.

This isn't exactly a new risk though. `never` is assignable to everything so it's fairly easy (in the same way) to accidentally pass it to a function as an argument. That leads me to be OK with the final judgment. That inconsistency though... if this is the final judgment, shouldn't `foo.prop` be relaxed to allow this access on never?

gabritto:
> I'm OK with the judgment itself but this inconsistency is killing me:
> 
> ```ts
> declare const foo: never
> 
> foo.prop // error ?????
> foo["prop"] // ok
> const { prop } = foo // ok
> ```
> 
> It feels just wildly inconsistent and hard to explain to people. How this difference in behavior would be explained in the docs?

I'm not even sure we did that on purpose, I'll have to investigate.

> I feel like in those contexts people wouldn't mind _at all_ the cast.

I'm open to believing that, given evidence, but it could still be a burden to make this update if it's needed in too many places.

> On the other hand, I'm quite worried - as the user - that `never` might accidentally sneak past me like this and I won't notice it at all. For instance, I doubt that [this](https://github.com/ant-design/ant-design/blob/4704eb578b8e7caabc502a5204c2ba058ee8266c/components/tree/utils/dictUtil.ts#L22) intentionally relies on `never` this way.

Yeah, that was a genuine error, as I pointed out above.

> This isn't exactly a new risk though. `never` is assignable to everything so it's fairly easy (in the same way) to accidentally pass it to a function as an argument. That leads me to be OK with the final judgment. That inconsistency though... if this is the final judgment, shouldn't `foo.prop` be relaxed to allow this access on never?

I'm having this thought that `never` is kind of like an `any` in the sense that you can do almost anything with it, so I agree that it's risky.
I think this isn't necessarily a "final judgement" in that we'll never ever do it, but we aren't currently convinced this is worth the break.

Issue #56412, opened 2023-11-15
Intellisense not resolving modules when using `exports` in package.json


mjbvz:
Reposting from VSCode repo for @magnusriga. I transferred the original issue to the wrong repo and now can't transfer it back

----

Does this issue occur when all extensions are disabled?: Yes

- VS Code Version: 1.84.2
- OS Version: Windows 11 Pro

Steps to Reproduce:

1. Build any project (A) with a dependency on another project (B), where project B's package.json exports files using the `exports` field
2. Observer how Intellisense does not work
3. Insert the `main` and `types` fields and see that Intellisense starts working again

### **Intellisense Works When `main` and `types` are set:**
**Package B's `package.json`**

![image](https://github.com/microsoft/vscode/assets/38915578/c4a87fea-7ee8-4bdb-8f96-37396e5b4fdb)

**Package A's `import` (notice the Intellisense pop-up. It is possible to ctrl-click the path.)**

![image](https://github.com/microsoft/vscode/assets/38915578/46e25339-85f3-4666-8922-d640942778d6)

### **Intellisense Does Not Work When only `exports` is set:**

**Package B's `package.json` (removed `main` and `types`)**

![image](https://github.com/microsoft/vscode/assets/38915578/382f9478-6031-4214-a60c-08fa055d224b)

**Package A's `import` (not possible to crl+click to find the package, i.e. no Intellisense)**

![image](https://github.com/microsoft/vscode/assets/38915578/7b138a5c-217a-4adc-b2cf-52e7721d035c)

RyanCavanaugh:
We need an actual repro. Also, please post code, not screenshots of code.

magnusriga:
I finally got Intellisense to work and respect the `exports` field.

Here is a working repo: https://github.com/magnusriga/my-turborepo

Leaving my config below in case it is helpful to anyone else. The below is slightly different to the repo above, it uses a `src` folder in package B.

**Package B's package.json**

````json
{
  "name": "@acme/ui",
  "version": "3.0.1",
  "private": true,
  "sideEffects": [
    "**/*.css"
  ],
  "exports": {
    ".": {
      "types": "./types/index.d.ts",
      "import": "./dist/index.js"
    },
    "./*": {
      "types": "./types/*.js",
      "import": "./dist/*.js"
    }
  },
  "scripts": {
    "build": "rimraf dist && tsc -p",
    "clean": "rm -rf dist",
[...]

PR #56785, opened 2023-12-14
Add `--module preserve`, support `require` in `--moduleResolution bundler`

andrewbranch:
This PR allows users targeting Bun and bundlers to write fully-typed `import` statements and `require` calls in the same file.

## Background and motivation

When we first introduced `--moduleResolution bundler`, we asked for feedback on how often people found themselves needing to write `require` calls in bundler code. We heard that it was very rare, so we chose not to perform any module resolution for `require` in that `moduleResolution` mode. This simplified resolution of package.json conditional `"exports"` and `"imports"`—if we were performing module resolution at all, it must have been from an import statement (or dynamic import), therefore we could unconditionally match the `import` condition. This resolution strategy would only be safe and accurate if we could ensure that these imports would not be transformed into require calls before a bundler saw them—and this is exactly what ts-loader, for example, would do if it used a tsconfig.json with `"module": "commonjs"`. To avoid this problem, we required `"module": "esnext"` to be used in conjunction with `"moduleResolution": "bundler"`. This restriction also prevented usage of `import x = require("dep")` syntax entirely.

The motivation for lifting this restriction is that we want a better default module resolution setting for VS Code than `--moduleResolution node10`, which lacks support for package.json `"exports"` but is still the default today[^1]. `nodenext` is too restrictive to be a good default, since many unconfigured projects might be targeting Bun or a bundler. We tried to switch to `--moduleResolution bundler`, but had to [revert](https://github.com/microsoft/vscode/pull/197031) because JavaScript users complained that they stopped getting IntelliSense on things they `require`d.

Adding support for `require()` resolution to `--moduleResolution bundler` solves this. Now, when module resolution is initiated from an import statement, it matches the `import` package.json `"exports"` condition, and when it’s initiated from a require call, it matches `require`.

## `--module preserve`

In TypeScript files, though, require calls look like:

```ts
import x = require("dep");
```

which are not allowed in `--module esnext`. To represent the fact that Bun and most bundlers allow `import` and `require` in the same file, we need a `--module` mode that does the same, without assuming that imports will be transformed into require calls.

In `--module preserve`, ESM syntax is emitted as-written[^2], while TypeScript-specific CommonJS syntax (`import x = require("dep")` and `export = {}`) is allowed and emitted as its standard CommonJS equivalent output (`require` variable statement and `module.exports` assignment, respectively). (Although we expect most users to use this mode with `--noEmit` for use in Bun and bundlers, the way code emits, or _would_ emit if enabled, is the basis for [how TypeScript performs type checking and module resolution](https://www.typescriptlang.org/docs/handbook/modules/theory.html#the-module-output-format) on imports/requires.)

`--module preserve` implies these defaults:
- `--moduleResolution bundler`
- `--esModuleInterop`
- `--resolveJsonModule`

I also considered, and am open to, implying `--moduleDetection force`.

[^1]: Recent confusion due to bad default module resolution: #56412
[^2]: `--module preserve` does not change the default behavior of import elision. Imports will never be transformed into `require` calls, but imports that are never used in JS-emitting positions will still be dropped, as in any other mode. To prevent unused imports from being elided, use `--verbatimModuleSyntax`.


andrewbranch:
@typescript-bot test top100

andrewbranch:
@typescript-bot test top100

andrewbranch:
@typescript-bot test top200

andrewbranch:
@typescript-bot pack this

andrewbranch:
@fatcerberus

1. The former is obviously TS code, so not allowed in JS
2. In the former, `x` can resolve to a type. `const` can only declare a value.
3. The `const` version does nothing (i.e. does not trigger module resolution) in TS files since `import` is available. It gets treated like any other variable statement, looking up a global for `require`.

fatcerberus:
> The `const` version does nothing (i.e. does not trigger module resolution) in TS files since `import` is available. It gets treated like any other variable statement, looking up a global for `require`.

This I didn't know; I assumed `require()` was always special-cased to trigger module resolution wherever it appears (e.g. you might dynamically require something).

DanielRosenwasser:
> `--module preserve` does not change the default behavior of import elision.

The name "preserve" also feels off because it is implicitly CommonJS-y. I don't know if that should weigh in here.

andrewbranch:
I have no idea what you mean by "it is implicitly CommonJS-y" but I don't think it is.

robpalme:
Module modes are primarily about the output.  On the naming of `preserve`, I get that this is communicating the output is passthrough/transparent/no-transform relative to the input - but even that seems slightly off given the way `import x = require()` gets transformed.  I mildly (and probably subjectively) associate the word `preserve` with the clarity/strictness of type-deletion-only, TS=JS+types, etc.

Perhaps the more important point to convey is that, unlike existing options, this permitting a mixture of both module systems (ESM and CJS) in the output.  So maybe it could be `--module dual` or `--module mixed`?

andrewbranch:
`import x = require("foo")` of course _has_ to be transformed if we’re emitting JavaScript, since it’s not valid JavaScript syntax. But the way it gets transformed is minimal and direct, similar to stripping type annotations. While I think it’s regrettable that such syntax exists, users are supposed to recognize `import x = require("foo")` as a CommonJS require, and `export = foo` as a CommonJS `module.exports =` statement. The output of these is massaged into valid JavaScript, but it doesn’t change flavor in the way that transforming an ESM import declaration into a CommonJS require does. Even in `verbatimModuleSyntax`, `import x = require("foo")` get transformed as much as they have to, making the module syntax not _truly_ verbatim.

src/compiler/program.ts
```diff
@@ -1052,7 +1056,7 @@ export function loadWithModeAwareCache<Entry, SourceFile, ResolutionCache, Resol
     const loader = createLoader(containingFile, redirectedReference, options, host, resolutionCache);
     for (const entry of entries) {
         const name = loader.nameAndMode.getName(entry);
-        const mode = loader.nameAndMode.getMode(entry, containingSourceFile);
+        const mode = loader.nameAndMode.getMode(entry, containingSourceFile, options);
```
sheetalkamat, line 1059: I think you want to use `redirectedReference.options || options` since for imports in file from referenced project, we want to use the options of referenced project and not referencing project

sheetalkamat (changes requested):

src/compiler/program.ts
andrewbranch: I think the expected behavior is debatable. Suppose you have a bundled frontend app that references a shared library designed to work in Node.js as well as bundlers and other runtimes. The referenced project settings would probably use module settings for Node.js, since those tend to be the strictest. But when the bundled frontend app imports that project, the actual module system in play is only the bundler—it doesn’t matter what settings the referenced library was checked against. In this scenario, it makes more sense to use the options from the primary project, not the one being imported.

But you can also imagine the opposite. Suppose you have a Node.js app that references a shared library built into a single file with a bundler. In that case, the primary project (the Node.js app) would be using very strict, Node.js-specific module settings, while the referenced library might have been built with `--moduleResolution bundler` and maybe even some `customConditions`. And whatever imports are present in the source files / declaration files of the bundled library have already been baked into the output that the Node.js app will see at runtime. In this scenario, it makes more sense to use the options from the referenced project.

Both of these sound plausible, and we don’t have configuration sophisticated enough to distinguish between them. I think the former sounds more likely, which is why I chose to use the options of the primary project being compiled. However, I’m ok with switching if there’s a strong precedent to do everything with respect to the referenced project options.

sheetalkamat: When resolving imports from references projects it’s going to be .d.ts file built from referenced project but resolutions happen at original ts location with the referenced project setting…in fact that’s how it works today as otherwise the resolutions may not work 

for editor scenarios anyways we include ts file from referenced project and all the resolutions in that project follow the options of referenced project 

when I get to desk tomorrow I can link the issue where we started doing this

sheetalkamat: Also note that none of the imports are emitted so emit thing doesn’t matter

src/compiler/program.ts
```diff
@@ -1183,7 +1187,7 @@ export function getReferencedFileLocation(program: Program, ref: ReferencedFile)
     switch (kind) {
         case FileIncludeKind.Import:
             const importLiteral = getModuleNameStringLiteralAt(file, index);
-            packageId = program.getResolvedModule(file, importLiteral.text, getModeForResolutionAtIndex(file, index))?.resolvedModule?.packageId;
+            packageId = program.getResolvedModule(file, importLiteral.text, getModeForResolutionAtIndex(file, index, program.getCompilerOptions()))?.resolvedModule?.packageId;
```
sheetalkamat, line 1190: You need to handle redirected reference here as well

src/compiler/program.ts
```diff
@@ -2157,7 +2161,7 @@ export function createProgram(rootNamesOrOptions: readonly string[] | CreateProg
             const moduleName = moduleNames[i];
             // If the source file is unchanged and doesnt have invalidated resolution, reuse the module resolutions
             if (file === oldSourceFile && !hasInvalidatedResolutions(file.path)) {
-                const mode = getModeForUsageLocation(file, moduleName);
+                const mode = getModeForUsageLocation(file, moduleName, options);
```
sheetalkamat, line 2164: And here

src/compiler/program.ts
```diff
@@ -2226,7 +2230,7 @@ export function createProgram(rootNamesOrOptions: readonly string[] | CreateProg
         // If we change our policy of rechecking failed lookups on each program create,
         // we should adjust the value returned here.
         function moduleNameResolvesToAmbientModuleInNonModifiedFile(moduleName: StringLiteralLike): boolean {
-            const resolutionToFile = oldProgram?.getResolvedModule(file, moduleName.text, getModeForUsageLocation(file, moduleName))?.resolvedModule;
+            const resolutionToFile = oldProgram?.getResolvedModule(file, moduleName.text, getModeForUsageLocation(file, moduleName, oldProgram.getCompilerOptions()))?.resolvedModule;
```
sheetalkamat, line 2233: And here

src/compiler/program.ts
```diff
@@ -2525,6 +2529,7 @@ export function createProgram(rootNamesOrOptions: readonly string[] | CreateProg
                 (name, mode) => oldProgram!.getResolvedModule(newSourceFile, name, mode),
                 moduleResolutionIsEqualTo,
                 moduleResolutionNameAndModeGetter,
+                oldProgram.getCompilerOptions(),
```
sheetalkamat, line 2532: Here

src/compiler/program.ts
```diff
@@ -2538,6 +2543,7 @@ export function createProgram(rootNamesOrOptions: readonly string[] | CreateProg
                 (name, mode) => oldProgram?.getResolvedTypeReferenceDirective(newSourceFile, name, mode),
                 typeDirectiveIsEqualTo,
                 typeReferenceResolutionNameAndModeGetter,
+                oldProgram.getCompilerOptions(),
```
sheetalkamat, line 2546: Here

src/compiler/program.ts
```diff
@@ -4029,7 +4034,7 @@ export function createProgram(rootNamesOrOptions: readonly string[] | CreateProg
             for (let index = 0; index < moduleNames.length; index++) {
                 const resolution = resolutions[index].resolvedModule;
                 const moduleName = moduleNames[index].text;
-                const mode = getModeForUsageLocation(file, moduleNames[index]);
+                const mode = getModeForUsageLocation(file, moduleNames[index], optionsForFile);
```
sheetalkamat, line 4037: This should always be `getRedirectReferenceForResolution(file)?.commandline.options || options`

since optionsForFile are to handle "allowJs" etc

src/compiler/resolutionCache.ts
```diff
@@ -822,7 +822,7 @@ export function createResolutionCache(resolutionHost: ResolutionCacheHost, rootD
         const seenNamesInFile = createModeAwareCache<true>();
         for (const entry of entries) {
             const name = loader.nameAndMode.getName(entry);
-            const mode = loader.nameAndMode.getMode(entry, containingSourceFile);
+            const mode = loader.nameAndMode.getMode(entry, containingSourceFile, options);
```
sheetalkamat, line 825: Handle redirectedReference

src/compiler/resolutionCache.ts
```diff
@@ -881,7 +881,7 @@ export function createResolutionCache(resolutionHost: ResolutionCacheHost, rootD
         reusedNames?.forEach(entry =>
             seenNamesInFile.set(
                 loader.nameAndMode.getName(entry),
-                loader.nameAndMode.getMode(entry, containingSourceFile),
+                loader.nameAndMode.getMode(entry, containingSourceFile, options),
```
sheetalkamat, line 884: redirectedReference to be handled

src/services/codefixes/convertToEsModule.ts
```diff
@@ -107,7 +107,7 @@ function fixImportOfModuleExports(
     quotePreference: QuotePreference,
 ) {
     for (const moduleSpecifier of importingFile.imports) {
-        const imported = program.getResolvedModule(importingFile, moduleSpecifier.text, getModeForUsageLocation(importingFile, moduleSpecifier))?.resolvedModule;
+        const imported = program.getResolvedModule(importingFile, moduleSpecifier.text, getModeForUsageLocation(importingFile, moduleSpecifier, program.getCompilerOptions()))?.resolvedModule;
```
sheetalkamat, line 110: Handle redirectReference

src/services/codefixes/fixSpelling.ts
```diff
@@ -181,7 +181,7 @@ function convertSemanticMeaningToSymbolFlags(meaning: SemanticMeaning): SymbolFl
 function getResolvedSourceFileFromImportDeclaration(sourceFile: SourceFile, context: CodeFixContextBase, importDeclaration: ImportDeclaration): SourceFile | undefined {
     if (!importDeclaration || !isStringLiteralLike(importDeclaration.moduleSpecifier)) return undefined;
 
-    const resolvedModule = context.program.getResolvedModule(sourceFile, importDeclaration.moduleSpecifier.text, getModeForUsageLocation(sourceFile, importDeclaration.moduleSpecifier))?.resolvedModule;
+    const resolvedModule = context.program.getResolvedModule(sourceFile, importDeclaration.moduleSpecifier.text, getModeForUsageLocation(sourceFile, importDeclaration.moduleSpecifier, context.program.getCompilerOptions()))?.resolvedModule;
```
sheetalkamat, line 184: And here

src/services/codefixes/importFixes.ts
```diff
@@ -1439,7 +1440,7 @@ function promoteFromTypeOnly(
         // Change .ts extension to .js if necessary
         if (!compilerOptions.allowImportingTsExtensions) {
             const moduleSpecifier = tryGetModuleSpecifierFromDeclaration(importClause.parent);
-            const resolvedModule = moduleSpecifier && program.getResolvedModule(sourceFile, moduleSpecifier.text, getModeForUsageLocation(sourceFile, moduleSpecifier))?.resolvedModule;
+            const resolvedModule = moduleSpecifier && program.getResolvedModule(sourceFile, moduleSpecifier.text, getModeForUsageLocation(sourceFile, moduleSpecifier, compilerOptions))?.resolvedModule;
```
sheetalkamat, line 1443: And here

src/services/findAllReferences.ts
```diff
@@ -1015,7 +1015,7 @@ export namespace Core {
             if (!options.implementations && isStringLiteralLike(node)) {
                 if (isModuleSpecifierLike(node)) {
                     const fileIncludeReasons = program.getFileIncludeReasons();
-                    const referencedFileName = program.getResolvedModule(node.getSourceFile(), node.text, getModeForUsageLocation(node.getSourceFile(), node))?.resolvedModule?.resolvedFileName;
+                    const referencedFileName = program.getResolvedModule(node.getSourceFile(), node.text, getModeForUsageLocation(node.getSourceFile(), node, program.getCompilerOptions()))?.resolvedModule?.resolvedFileName;
```
sheetalkamat, line 1018: And here

src/services/getEditsForFileRename.ts
```diff
@@ -246,7 +246,7 @@ function getSourceFileToImport(
         return newFileName === undefined ? { newFileName: oldFileName, updated: false } : { newFileName, updated: true };
     }
     else {
-        const mode = getModeForUsageLocation(importingSourceFile, importLiteral);
+        const mode = getModeForUsageLocation(importingSourceFile, importLiteral, program.getCompilerOptions());
```
sheetalkamat, line 249: And here

src/services/goToDefinition.ts
```diff
@@ -180,7 +180,7 @@ export function getDefinitionAtPosition(program: Program, sourceFile: SourceFile
     if (!symbol && isModuleSpecifierLike(fallbackNode)) {
         // We couldn't resolve the module specifier as an external module, but it could
         // be that module resolution succeeded but the target was not a module.
-        const ref = program.getResolvedModule(sourceFile, fallbackNode.text, getModeForUsageLocation(sourceFile, fallbackNode))?.resolvedModule;
+        const ref = program.getResolvedModule(sourceFile, fallbackNode.text, getModeForUsageLocation(sourceFile, fallbackNode, program.getCompilerOptions()))?.resolvedModule;
```
sheetalkamat, line 183: And here

I feel like we can add internal API to program that gives the mode from sourceFile and literal and handles referenced project options

src/services/goToDefinition.ts
```diff
@@ -336,7 +336,7 @@ export function getReferenceAtPosition(sourceFile: SourceFile, position: number,
     if (sourceFile.imports.length || sourceFile.moduleAugmentations.length) {
         const node = getTouchingToken(sourceFile, position);
         let resolution: ResolvedModuleWithFailedLookupLocations | undefined;
-        if (isModuleSpecifierLike(node) && isExternalModuleNameRelative(node.text) && (resolution = program.getResolvedModule(sourceFile, node.text, getModeForUsageLocation(sourceFile, node)))) {
+        if (isModuleSpecifierLike(node) && isExternalModuleNameRelative(node.text) && (resolution = program.getResolvedModule(sourceFile, node.text, getModeForUsageLocation(sourceFile, node, program.getCompilerOptions())))) {
```
sheetalkamat, line 339: Same

src/services/refactors/moveToFile.ts
```diff
@@ -313,10 +313,11 @@ function getTargetFileImportsAndAddExportInOldFile(
             forEachImportInStatement(oldStatement, i => {
                 // Recomputing module specifier
                 const moduleSpecifier = moduleSpecifierFromImport(i);
-                const resolved = program.getResolvedModule(oldFile, moduleSpecifier.text, getModeForUsageLocation(oldFile, moduleSpecifier));
+                const compilerOptions = program.getCompilerOptions();
+                const resolved = program.getResolvedModule(oldFile, moduleSpecifier.text, getModeForUsageLocation(oldFile, moduleSpecifier, compilerOptions));
```
sheetalkamat, line 317: Same

src/services/stringCompletions.ts
```diff
@@ -583,7 +583,7 @@ function getStringLiteralCompletionsFromModuleNames(sourceFile: SourceFile, node
 
 function getStringLiteralCompletionsFromModuleNamesWorker(sourceFile: SourceFile, node: LiteralExpression, compilerOptions: CompilerOptions, host: LanguageServiceHost, typeChecker: TypeChecker, preferences: UserPreferences): readonly NameAndKind[] {
     const literalValue = normalizeSlashes(node.text);
-    const mode = isStringLiteralLike(node) ? getModeForUsageLocation(sourceFile, node) : undefined;
+    const mode = isStringLiteralLike(node) ? getModeForUsageLocation(sourceFile, node, compilerOptions) : undefined;
```
sheetalkamat, line 586: Same

src/services/suggestionDiagnostics.ts
```diff
@@ -88,7 +88,7 @@ export function computeSuggestionDiagnostics(sourceFile: SourceFile, program: Pr
             const importNode = importFromModuleSpecifier(moduleSpecifier);
             const name = importNameForConvertToDefaultImport(importNode);
             if (!name) continue;
-            const module = program.getResolvedModule(sourceFile, moduleSpecifier.text, getModeForUsageLocation(sourceFile, moduleSpecifier))?.resolvedModule;
+            const module = program.getResolvedModule(sourceFile, moduleSpecifier.text, getModeForUsageLocation(sourceFile, moduleSpecifier, program.getCompilerOptions()))?.resolvedModule;
```
sheetalkamat, line 91: Same

sheetalkamat (changes requested):

src/services/goToDefinition.ts
andrewbranch: I think it needs to be public (and maybe even deprecate the existing `getModeForUsageLocation`), otherwise this will be a footgun for API users.

sheetalkamat: In most of the cases the options supplied to this should have been the redirect options .. (eg.  CompilerHost.getSourceFile should pass correct options to the "sourceFile" creation but clearly thats incorrect today.

src/compiler/types.ts
```diff
@@ -4760,6 +4762,14 @@ export interface Program extends ScriptReferenceHost {
...
     isSourceFileFromExternalLibrary(file: SourceFile): boolean;
     isSourceFileDefaultLibrary(file: SourceFile): boolean;
+    /**
+     * Calculates the final resolution mode for a given module reference node. This is the resolution mode explicitly provided via import
+     * attributes, if present, or the syntax the usage would have if emitted to JavaScript. In `--module node16` or `nodenext`, this may
+     * depend on the file's `impliedNodeFormat`. In `--module preserve`, it depends only on the input syntax of the reference. In other
+     * `module` modes, when overriding import attributes are not provided, this function returns `undefined`, as the result would have no
+     * impact on module resolution, emit, or type checking.
+     */
+    getModeForUsageLocation(file: SourceFile, usage: StringLiteralLike): ResolutionMode;
```
andrewbranch, line 4772: FYI @sheetalkamat:

I made this API public and suggested using it instead of `ts.getModeForUsageLocation`. I think we could consider deprecating that one.

However, when replacing all of our own usage of `getModeForUsageLocation` with `program.getModeForUsageLocation`, I noticed that it was almost _always_ used in conjunction with `program.getResolvedModule`. So, I made the internal API `program.getResolvedModuleFromModuleSpecifier` above to simplify. I think we could also consider making that one public, since getting the resolution of an import is extremely useful, and the new function does so without exposing the complexity of `ResolutionMode`.

jakebailey:
> I also considered, and am open to, implying `--moduleDetection force`.

I assume that we aren't doing this, because we want this option to be usable in VS Code as the new default for random files?

src/compiler/program.ts
```diff
@@ -4443,8 +4456,8 @@ export function createProgram(rootNamesOrOptions: readonly string[] | CreateProg
             createDiagnosticForOptionName(Diagnostics.Option_0_can_only_be_used_when_moduleResolution_is_set_to_node16_nodenext_or_bundler, "customConditions");
         }
 
-        if (moduleResolution === ModuleResolutionKind.Bundler && !emitModuleKindIsNonNodeESM(moduleKind)) {
-            createOptionValueDiagnostic("moduleResolution", Diagnostics.Option_0_can_only_be_used_when_module_is_set_to_es2015_or_later, "bundler");
+        if (moduleResolution === ModuleResolutionKind.Bundler && !emitModuleKindIsNonNodeESM(moduleKind) && moduleKind !== ModuleKind.Preserve) {
+            createOptionValueDiagnostic("moduleResolution", Diagnostics.Option_0_can_only_be_used_when_module_is_set_to_preserve_or_to_es2015_or_later, "bundler");
```
jakebailey, line 4460: Do we also need to verify that `preserve` is not used with other options? (Are there other incompatible options?)

src/compiler/transformers/module/esnextAnd2015.ts
```diff
@@ -238,9 +242,27 @@ export function transformECMAScriptModule(context: TransformationContext): (x: S
         return statements;
     }
 
-    function visitExportAssignment(node: ExportAssignment): VisitResult<ExportAssignment | undefined> {
-        // Elide `export=` as it is not legal with --module ES6
-        return node.isExportEquals ? undefined : node;
+    function visitExportAssignment(node: ExportAssignment): VisitResult<ExportAssignment | ExpressionStatement | undefined> {
+        if (node.isExportEquals) {
+            if (getEmitModuleKind(compilerOptions) === ModuleKind.Preserve) {
```
jakebailey, line 247: At some level I wish this were written like this to reduce nesting:

```ts
if (!node.isExportEquals) {
	return node;
}
if (getEmitModuleKind(compilerOptions) !== ModuleKind.Preserve) {
	return undefined;
}
// big long text
```

But that's a silly nit and we aren't consistent about early returns anyway.

src/compiler/types.ts
```diff
@@ -7296,6 +7307,9 @@ export enum ModuleKind {
     // Node16+ is an amalgam of commonjs (albeit updated) and es2022+, and represents a distinct module system from es2020/esnext
     Node16 = 100,
     NodeNext = 199,
+
+    // Emit as written
+    Preserve = 200,
```
jakebailey, line 7312: ~99~ ~100~ ~199~ 200 ought to be enough for anyone 😄

andrewbranch:
> > I also considered, and am open to, implying `--moduleDetection force`.
> 
> I assume that we aren't doing this, because we want this option to be usable in VS Code as the new default for random files?

Honestly I suspect the default behavior of separate files being assumed to be part of the same global scope causes more confusion and doesn’t reflect real usage more often than not, but it deserves a separate discussion, I guess.

src/compiler/program.ts
andrewbranch: Nothing I’ve thought of. Most `module` values don’t impose any restrictions.

tests/baselines/reference/api/typescript.d.ts
```diff
@@ -9908,19 +9917,27 @@ declare namespace ts {
      * @param file File to fetch the resolution mode within
      * @param index Index into the file's complete resolution list to get the resolution of - this is a concatenation of the file's imports and module augmentations
      */
-    function getModeForResolutionAtIndex(file: SourceFile, index: number): ResolutionMode;
+    function getModeForResolutionAtIndex(file: SourceFile, index: number, compilerOptions: CompilerOptions): ResolutionMode;
```
jakebailey, line 9920: This is a breaking change; maybe it was discussed somewhere else in the PR (I can't find it), but is this intentional? It does appear to be used externally. https://github.com/search?q=%2F%5C.getModeForResolutionAtIndex%2F&type=code

tests/baselines/reference/api/typescript.d.ts
```diff
@@ -9908,19 +9917,27 @@ declare namespace ts {
...
      */
-    function getModeForUsageLocation(file: {
-        impliedNodeFormat?: ResolutionMode;
-    }, usage: StringLiteralLike): ModuleKind.CommonJS | ModuleKind.ESNext | undefined;
+    function getModeForUsageLocation(
+        file: {
+            impliedNodeFormat?: ResolutionMode;
+        },
+        usage: StringLiteralLike,
+        compilerOptions: CompilerOptions,
```
jakebailey, line 9939: Similarly here, and also used externally: https://github.com/search?q=%2F%5C.getModeForUsageLocation%2F&type=code

jakebailey (commented):
Seems good but I am wondering about the breaking API changes; the two I noted are used by some important libraries.

andrewbranch:
@jakebailey there’s no way to get a correct resolution mode answer in `--module preserve` without supplying `compilerOptions`. What do you think about making type-level breaking changes, but guarding against an undefined `compilerOptions` internally?

jakebailey:
I think making them assert with a good message would be a good idea, at the least. But, I don't know how I feel about breaking `ts-node`, `tsimp`, `rollup`, `svelte`, `rush`, `lit-analyzer`, `dt-tools`, etc...

If undefined, can we just assume some options, keeping those from breaking? Presumably before, we were able to answer without the compiler option, right?

andrewbranch:
Yes, the implementation will only be incorrect when `--module` is `preserve`.

jakebailey:
> Yes, the implementation will only be incorrect when `--module` is `preserve`. (By “guarding against an undefined `compilerOptions`” I meant “don’t throw, assume `module` is not `preserve`.”)

Even better!

tests/baselines/reference/api/typescript.d.ts
jakebailey: I think this one is still breaking after the commit.

jakebailey: Oh, the new stuff is still "breaking" at a type level, but with hidden overloads?

andrewbranch: Yep. The thought is that I want tool authors to notice immediately, but I don’t want it to crash when users bring a newer TS to an existing tool.

jakebailey: Sounds good; hopefully nobody accidentally removes that backwards compat code in a refactor but that sounds doubtful.

jakebailey (approved):
Looks good to my eye, thanks for adding the back compat code.

mrazauskas:
Perhaps this is the right place to ask. I was wondering why new `ModuleKind`s and `ModuleResolutionKind`s are getting listed only in `ts.ModuleKind` and `ts.ModuleResolutionKind`, but not in `ts.server.protocol.ModuleKind` and `ts.server.protocol.ModuleResolutionKind` enums? 

For instance, "node16" or "bundler" are missing here: https://github.com/microsoft/TypeScript/blob/55153b027490cea949d50747b4b4637029120777/src/server/protocol.ts#L3756-L3759

Similarly "node16" is missing and this PR is not adding "preserve" here: https://github.com/microsoft/TypeScript/blob/55153b027490cea949d50747b4b4637029120777/src/server/protocol.ts#L3745-L3754

andrewbranch:
I honestly had no idea those were there. I think the only time they’d be used is for clients to send the `CompilerOptionsForInferredProjects` command, which today in VS Code still uses `module: "CommonJS", moduleResolution: "Node"`, so we didn’t notice the need to add anything else. I’ll update them, thanks 👍

sheetalkamat (approved):

PR #56791, opened 2023-12-15
Reduce file system calls during test setup.

### LINKED ISSUE #56790: Improve test discovery speed
### Acknowledgement

### Comment

Running a single test or a small number of tests can be slow because the current test runner will discover all tests and then greps based on test title. The main issue appears to be the number of fs sys calls during setup. This does not happen on all systems, but on my system (Windows 10), especially in debug test discovery can take up to 6/7 seconds.

dragomirtitian:
Improvements to reduce fs calls during test discovery:

1. Use `opendirSync` to read the directory content. This reduces the need to call `fs.existsSync` and `fs.statSync`
2. Cache `enumerateTestFiles` result to avoid having to enumerating the files again
3. Delay creation of vfs for project tests (as creating the vfs also mounts real folders which also touch the file system)

Improvements in test discovery time (anecdotal):

1. Linux node 18 debug - ~1.4 seconds -> ~800ms
2. Windows 10 node 20 debug - 8s -> 2s

Performance of running all the tests does not appear to be impacted as the cost of test discovery is not a big percentage of running all tests and the work that si delayed will have to eventually be executed anyway.

Fixes #56790

### DIFF
--- a/src/harness/harnessIO.ts
+++ b/src/harness/harnessIO.ts
@@ -25,7 +25,7 @@ export interface IO {
     fileExists(fileName: string): boolean;
     directoryExists(path: string): boolean;
     deleteFile(fileName: string): void;
-    enumerateTestFiles(runner: RunnerBase): (string | FileBasedTest)[];
+    enumerateTestFiles(runner: RunnerBase): string[];
     listFiles(path: string, filter?: RegExp, options?: { recursive?: boolean; }): string[];
     log(text: string): void;
     args(): string[];
--- a/src/harness/harnessIO.ts
+++ b/src/harness/harnessIO.ts
@@ -87,19 +87,20 @@ function createNodeIO(): IO {
     function listFiles(path: string, spec: RegExp, options: { recursive?: boolean; } = {}) {
         function filesInFolder(folder: string): string[] {
             let paths: string[] = [];
+            const dir = fs.opendirSync(folder);
+            for (let entry = dir.readSync(); entry; entry = dir.readSync()) {
+                const pathToFile = pathModule.join(folder, entry.name);
 
-            for (const file of fs.readdirSync(folder)) {
-                const pathToFile = pathModule.join(folder, file);
-                if (!fs.existsSync(pathToFile)) continue; // ignore invalid symlinks
-                const stat = fs.statSync(pathToFile);
-                if (options.recursive && stat.isDirectory()) {
+                if (entry.isSymbolicLink() && !fs.existsSync(pathToFile)) continue; // ignore invalid symlinks
+
+                if (options.recursive && entry.isDirectory()) {
                     paths = paths.concat(filesInFolder(pathToFile));
                 }
-                else if (stat.isFile() && (!spec || file.match(spec))) {
+                else if (entry.isFile() && (!spec || entry.name.match(spec))) {
                     paths.push(pathToFile);
                 }
             }
-
+            dir.closeSync();
             return paths;
         }
 
--- a/src/harness/harnessIO.ts
+++ b/src/harness/harnessIO.ts
@@ -84,52 +84,26 @@ function createNodeIO(): IO {
         return runner.getTestFiles();
     }
 
-    function listFiles(path: string, spec: RegExp, options: { recursive?: boolean; } = {}) {
+    function listFiles(path: string, spec: RegExp | undefined, options: { recursive?: boolean; } = {}) {
         function filesInFolder(folder: string): string[] {
+            const { files, directories } = ts.sys.getAccessibleFileSystemEntries!(folder);
             let paths: string[] = [];
-            for (const entry of fs.readdirSync(folder, { withFileTypes: true })) {
-                const pathToFile = pathModule.join(folder, entry.name);
-
-                if (entry.isSymbolicLink() && !fs.existsSync(pathToFile)) continue; // ignore invalid symlinks
-
-                if (options.recursive && entry.isDirectory()) {
-                    paths = paths.concat(filesInFolder(pathToFile));
-                }
-                else if (entry.isFile() && (!spec || entry.name.match(spec))) {
+            for (const file of files) {
+                const pathToFile = pathModule.join(folder, file);
+                if (!spec || file.match(spec)) {
                     paths.push(pathToFile);
+                }
+            }
+            if (options.recursive) {
+                for (const dir of directories) {
+                    const pathToDir = pathModule.join(folder, dir);
+                    paths = paths.concat(filesInFolder(pathToDir));
                 }
             }
             return paths;
         }
 
         return filesInFolder(path);
-    }
-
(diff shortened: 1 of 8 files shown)


src/harness/harnessIO.ts
```diff
@@ -87,19 +87,20 @@ function createNodeIO(): IO {
     function listFiles(path: string, spec: RegExp, options: { recursive?: boolean; } = {}) {
         function filesInFolder(folder: string): string[] {
             let paths: string[] = [];
+            const dir = fs.opendirSync(folder);
```
jakebailey, line 90: Could we just use `readdirSync` with `withFileTypes: true` instead? With recursion, we'd be holding directories open the entire way down, but maybe we never go deep enough.

(I'd also say that readdir has `recursive` now, but it's only available in Node 18+)

dragomirtitian: That is a fair point. I don't think the we have a very deep directory structure in tests so it's probably fine.  I will read the whole current directory first close it and then continue reading sub directories just to be safe.

jakebailey: Is withFileTypes enough to get all of the needed info? That's effectively what loading it all at once does.

I'm not against holding them open really, the code is plenty clear and I'm sure fine as is; ignore my hold open comment (speaking aloud too much)

robpalme, line 90: Given we're optimizing for performance, does increasing the default buffer size from 32 help? That option has been available since Node 12.

dragomirtitian: I changed it over to use withFileTypes

dragomirtitian: Switched over to withFileTypes

jakebailey: Thanks; does the perf improvement change at all?

jakebailey (approved):

src/harness/harnessIO.ts
```diff
@@ -87,19 +87,18 @@ function createNodeIO(): IO {
     function listFiles(path: string, spec: RegExp, options: { recursive?: boolean; } = {}) {
         function filesInFolder(folder: string): string[] {
             let paths: string[] = [];
+            for (const entry of fs.readdirSync(folder, { withFileTypes: true })) {
```
dmichon-msft, line 90: Apply the same optimization to `getAccessibleFileSystemEntries`?

jakebailey: Yeah, this is a good idea; the implementation we have in `sys.ts` already uses `withFileTypes`.

(That being said, this whole block just duplicates functionality in `sys.ts` so I do wonder how much copy pasting we could avoid.)

src/testRunner/compilerRunner.ts
```diff
@@ -51,9 +51,10 @@ export class CompilerBaselineRunner extends RunnerBase {
         return this.testSuiteName;
     }
 
+    private loadedTestFiles: CompilerFileBasedTest[] | undefined;
     public enumerateTestFiles() {
         // see also: `enumerateTestFiles` in tests/webTestServer.ts
-        return this.enumerateFiles(this.basePath, /\.tsx?$/, { recursive: true }).map(CompilerTest.getConfigurations);
+        return this.loadedTestFiles ??= this.enumerateFiles(this.basePath, /\.tsx?$/, { recursive: true }).map(CompilerTest.getConfigurations);
```
sheetalkamat, line 57: I think these need to be released at some point as otherwise we are holding onto this data for too long (till all tests are running?)

jakebailey: It would be until the process ends I believe, but I don't think that this actually ends up holding on to too much; it's just a list of filenames, right?

sheetalkamat: Its typed :  `CompilerFileBasedTest` so not sure if its just file names or more ?

jakebailey: So, it'll include filenames, but also file contents and configurations. Honestly, I'm shocked that we do that stuff up front, but I suppose you need the file contents to know the config to then know what tests to generate...

dragomirtitian: @sheetalkamat @jakebailey The more aggressive approach would be to return just the `string[]` from `enumerateTestFiles`. I do not see an instance where the extra information returned is used. Maybe it was in the past and it is no longer needed? https://github.com/microsoft/TypeScript/commit/fd8b18244bffb349cd1bf54e31e597041422d24e

This would remove the double reading and parsing of the test files and there would not be a need to cache the result.

jakebailey: That actually seems great.

dragomirtitian: Pushed the new version.

jakebailey:
Looks good, though there's still an open thread about another function that could use `withFileTypes`.

That being said, I also don't see as great of an improvement as your benchmark; before:

```
Benchmark 1: hereby runtests -t "notarealtest" --no-typecheck --no-lint
  Time (mean ± σ):      2.265 s ±  0.012 s    [User: 2.540 s, System: 0.652 s]
  Range (min … max):    2.250 s …  2.286 s    10 runs
```

After:

```
Benchmark 1: hereby runtests -t "notarealtest" --no-typecheck --no-lint
  Time (mean ± σ):      2.093 s ±  0.039 s    [User: 2.351 s, System: 0.559 s]
  Range (min … max):    2.035 s …  2.155 s    10 runs
```

8% faster is still good, but you had seen 75% faster...

sheetalkamat (commented):
This looks ok to me except @jakebailey s suggestion on just replacing `listFiles` and `getAccessibleFileSystemEntries` to use `ts.sys.getAccessibleFileSystemEntries`

dragomirtitian:
> This looks ok to me except @jakebailey s suggestion on just replacing `listFiles` and `getAccessibleFileSystemEntries` to use `ts.sys.getAccessibleFileSystemEntries`

@sheetalkamat One small problem is that `getAccessibleFileSystemEntries` is not exposed in `sys`. I exposed it as internal. hope that is ok. I rewrote `listFiles` to use the sys version and also removed the local implementation of `getAccessibleFileSystemEntries`

sheetalkamat (approved):

src/harness/harnessIO.ts
```diff
@@ -170,7 +143,7 @@ function createNodeIO(): IO {
         getWorkspaceRoot: () => workspaceRoot,
         exit: exitCode => ts.sys.exit(exitCode),
         readDirectory: (path, extension, exclude, include, depth) => ts.sys.readDirectory(path, extension, exclude, include, depth),
-        getAccessibleFileSystemEntries,
+        getAccessibleFileSystemEntries: ts.sys.getAccessibleFileSystemEntries!,
```
jakebailey, line 146: I don't think that it matters, but every other function in this list is an arrow function into `sys`, not the actual value.

jakebailey (approved):

PR #52968, opened 2023-02-25
Add `NoInfer` intrinsic type

Andarist:
closes https://github.com/microsoft/TypeScript/issues/14829

There is still some work to be done here (type displays on hover, assignability bug, and maybe more) - but I'm pushing this out to receive an early feedback

cc @RyanCavanaugh

PR #56794, opened 2023-12-15
Add `NoInfer<T>` intrinsic represented as special substitution type

### LINKED ISSUE #14829: Suggestion: Noninferential type parameter usage
We keep getting bugs like this and I keep not finding the original so I'm making a new one so we can find it.

Keywords: ignore type parameter optional inference

### Problem

Often with generics, there will be some locations where a type parameter *should* be inferrable from usage, and *other* places where the type parameter should only be used to *enforce* typechecking. This comes up in a variety of contexts

```ts
class Animal { move }
class Dog extends Animal { woof }

function doSomething<T>(value: T, getDefault: () => T) { }
// Wanted an error here - getDefault() ought to return same type as 'value'
doSomething(new Dog(), () => new Animal());
```

```ts
declare function assertEqual<T>(actual: T, expected: T): boolean;
const g = { x: 3, y: 2 };
assertEqual(g, { x: 3 }); // Forgot y, wanted error
```

### Proposal Sketch

We should be able to mark type parameter consumption sites as being "not eligible for inference". For example, let's say we had a special global type that the compiler knew not to unwrap during inference:
```ts
type NoInfer<T> = T;
```
Then we can annotate usage sites
```ts
function doSomething<T>(value: T, getDefault: () => NoInfer<T>) { }
// Wanted an error here - getDefault() ought to return same type as 'value'
doSomething(new Dog(), () => new Animal());
```
```ts
declare function assertEqual<T>(actual: T, expected: NoInfer<T>): boolean;
const g = { x: 3, y: 2 };
assertEqual(g, { x: 3 }); // Error
```

ahejlsberg:
This PR adds official support for a `NoInfer<T>` marker type that blocks inferences to the contained type. The PR supersedes #52968, but borrows as much as possible from that PR (lots of thanks for your work, @Andarist).

The PR represents `NoInfer<T>` types as special substitution types (types with kind `TypeFlags.Substitution`) having constraint type `unknown`, a pattern that otherwise never occurs. This representation ensures that we maximally leverage our existing infrastructure for erasing marker types when appropriate, and that we minimally disrupt the tooling ecosystem (because we're not introducing a new kind of type).

Other than blocking inference, `NoInfer<T>` markers have no effect on `T`. Indeed, `T` and `NoInfer<T>` are considered identical types in all other contexts.

Some examples:

```ts
declare function foo1<T extends string>(a: T, b: NoInfer<T>): void
declare function foo2<T extends string>(a: T, b: NoInfer<T>[]): void
declare function foo3<T extends string>(a: T, b: NoInfer<T[]>): void
declare function foo4<T extends string>(a: T, b: { x: NoInfer<T> }): void
declare function foo5<T extends string>(a: T, b: NoInfer<{ x: T }>): void

foo1('foo', 'foo') // ok
foo1('foo', 'bar') // error
foo2('foo', ['bar']) // error
foo3('foo', ['bar']) // error
foo4('foo', { x: 'bar' }) // error
foo5('foo', { x: 'bar' }) // error
```

Above, inferences to the type of `b` are blocked by the `NoInfer<T>` marker, thus causing errors to be reported when the argument types differ. Without the `NoInfer<T>` markers, inference simply produces the union `"foo" | "bar"`.

Fixes #14829.


andrewbranch:
@typescript-bot pack this

src/compiler/checker.ts
```diff
@@ -1363,13 +1363,15 @@ const enum IntrinsicTypeKind {
...
     Uncapitalize,
+    NoInfer,
 }
 
 const intrinsicTypeKinds: ReadonlyMap<string, IntrinsicTypeKind> = new Map(Object.entries({
     Uppercase: IntrinsicTypeKind.Uppercase,
     Lowercase: IntrinsicTypeKind.Lowercase,
     Capitalize: IntrinsicTypeKind.Capitalize,
     Uncapitalize: IntrinsicTypeKind.Uncapitalize,
+    NoInfer: IntrinsicTypeKind.NoInfer,
```
Andarist, line 1374: Slightly related question: should `ThisType` become an intrinsic type as well (it is kinda intrinsic by nature but it's not defined as such in the typedefs)? Or is it better to not touch it at all?

Andarist: Obviously this shouldn't be touched as part of this PR - I just take this opportunity to ask a question about it ;p

ahejlsberg:
@typescript-bot test top100
@typescript-bot run dt

fatcerberus:
> Without the `NoInfer<T>` markers, inference simply produces the union `"foo" | "bar"`.

Huh, so it does... here I had thought TS never inferred cross-candidate unions, but apparently there are cases when it does?

RyanCavanaugh:
Unions can be formed under the same constrained primitive family. That's why my TS Congress talk had to disavow knowledge of primitive types.

src/compiler/checker.ts
ahejlsberg: We want to leave `ThisType` as is. An `intrinsic` declaration doesn't really say anything about the internal representation of the type, and we already have a perfectly good solution for with the current interface-based declaration.

ahejlsberg:
Tests and performance all look unaffected.

craigphicks:
> Some examples:
> 
> ```ts
> declare function foo1<T extends string>(a: T, b: NoInfer<T>): void
> declare function foo2<T extends string>(a: T, b: NoInfer<T>[]): void
> declare function foo3<T extends string>(a: T, b: NoInfer<T[]>): void
> declare function foo4<T extends string>(a: T, b: { x: NoInfer<T> }): void
> declare function foo5<T extends string>(a: T, b: NoInfer<{ x: T }>): void
> 
> foo1('foo', 'foo') // ok
> foo1('foo', 'bar') // error
> foo2('foo', ['bar']) // error
> foo3('foo', ['bar']) // error
> foo4('foo', { x: 'bar' }) // error
> foo5('foo', { x: 'bar' }) // error
> ```

All these cases can be addressed by using the constraint `T&U` as follows:
```
declare function foo1<T extends string, U>(a: T, b: T&U): void
declare function foo2<T extends string, U>(a: T, b: T&U): void
declare function foo3<T extends string, U>(a: T, b: (T&U)[]): void
declare function foo4<T extends string, U>(a: T, b: { x: T&U }): void
declare function foo5<T extends string, U>(a: T, b: { x: T&U }): void

foo1('foo', 'foo') // ok
foo1('foo', 'bar') // error
foo2('foo', ['bar']) // error
foo3('foo', ['bar']) // error
foo4('foo', { x: 'bar' }) // error
foo5('foo', { x: 'bar' }) // error
```
[playground](https://www.typescriptlang.org/play?jsx=0#code/CYUwxgNghgTiAEAzArgOzAFwJYHtVJxwEYAeAFXhAA8MRVgBneBjGLVAcwBp4BVAPgAUUAFzwyPAEZiyAMl4BKMQDccWYAChQkWAhTpseAjgBM5SjTqNmrdtz5DR4qTPlL4q9VvDQ4SNJi4+IiEAMzm1LT0TCxsnDwCwjIu8IJyigDaALrunpravnoBhsGEACwRltE2cfaJThLw0vAA3vBUrrzwAL65avk+uv4GQcYArJVR1rF2CY7JTWJtHeLyPX1eGiHEggDk27s8+4S7CvAA9OfwOADWW4REewdHkrCnF1cgMDA4MPemTxOPAyu1eMF2OQ+lG+v3+oUBOEO8BBYIhZ0u0J+f22ZQRSOWYlBb3WUK+WP+YzxPAJ8CJ4JJGLJsI0QA)

`T&U` is fairly obvious and logical constraint to add.  It is not a trick.  Do you have any stronger cases?

So far it doesn't seem to me like a bug fix, it seems like perhaps a performance optimization, because you are explicitly shutting down inference, which could only shorten processing time.

Andarist:
As the library author, I don't want to introduce an extra generic type parameter to do this - this affects my public interface. `NoInfer` that is used within the signature isn't.

craigphicks:
> As the library author, I don't want to introduce an extra generic type parameter to do this - this affects my public interface. `NoInfer` that is used within the signature isn't.

"Performance", "Convenience", "Simplification" are worthwhile qualities.  However, based on the test cases, I still wouldn't call it a bug fix.

Andarist:
The only mention of this being a "fix" (that I can find) is in the closing reference to the original issue. I don't think it's worthwhile at all to bikeshed if that reference is using "fixes XXX" or "closes XXX".

craigphicks:
The opening line of the issue being closed:  #14829 

> We keep getting bugs like this and I keep not finding the original so I'm making a new one so we can find it.

As far as I can see, more test cases, especially ones that highlight the utility of this feature, would only serve to highlight it's utility and serve as instruction on what can be done now it is there.  Again, thank you for your hard work.

fatcerberus:
> We keep getting bugs like this and I keep not finding the original so I'm making a new one so we can find it.

I read "bug" here as a synonym for "bug-tracker issue"; in other words, "People keep posting GitHub issues requesting this feature so I'm making this as the canonical issue to bikeshed it."

btw I think the `T & U` trick *only* works because the `T` there, being part of an intersection, is considered a lower-priority inference site than the one at `a`.  That seems like an implementation detail to me and I'm not sure I'd want to rely on it.

craigphicks:
> > We keep getting bugs like this and I keep not finding the original so I'm making a new one so we can find it.
> 
> I read "bug" here as a synonym for "bug-tracker issue"; in other words, "People keep posting GitHub issues requesting this feature so I'm making this as the canonical issue to bikeshed it."
> 
> btw I think the `T & U` trick _only_ works because the `T` there, being part of an intersection, is considered a lower-priority inference site than the one at `a`. That seems like an implementation detail to me and I'm not sure I'd want to rely on it.

Yes you are right, it **was** a trick, so I changed it to `extends`. The original post also had a comment [that noted that extends could be used(https://github.com/microsoft/TypeScript/issues/14829#issuecomment-288902999), and that was countered by a [different example](https://github.com/microsoft/TypeScript/issues/14829#issuecomment-289179700), but that is also easy to write as a feed forward inference generic, as appended above.

I am convinced `NoInfer` will be more convenient, simpler, and faster.  I'm curious whether there are cases it would indispensable.

craigphicks:
Without NoInfer
```
function f<T,U>(a:T, b: U){
    f(b,a); 
//  ^? function f<U, T>(a: U, b: T): void
} 
```
What will the call the internal call to `f` look like in 
```
function f<T,U>(a:T, b: NoInfer<T>){
    f(b,a); 
//  ^? function f<?>(a:?, b: ?): void
} 
```
Will it be able to infer the type for `T` in the inner call?

Does `NoInfer` get automatically stripped by 
```
type UnNoInfer<N> = N extends NoInfer<infer T> ? T : never
```
?

tests/cases/conformance/types/typeRelationships/typeInference/noInfer.ts
```diff
@@ -0,0 +1,78 @@
...
+type T02 = NoInfer<undefined>;
+type T03 = NoInfer<"foo">;
+type T04 = NoInfer<`foo${string}`>;
+type T05 = NoInfer<`foo${string}` & `${string}bar`>;
+type T06 = NoInfer<{}>;
+
+// NoInfer<T> is preserved for object types
+
+type T10 = NoInfer<string[]>;
+type T11 = NoInfer<{ x: string }>;
```
Andarist, line 17: Any particular reason why those can't be erased?

At the moment this, in part at least, creates problems with mapped types and `keyof`. This works OK:
```ts
declare const obj: NoInfer<{ a: string, b: number }>
const ok: keyof typeof obj = 'a'
const err: keyof typeof obj = 'c'
```

However, `shouldDeferIndexType` returns `true` for this type and I feel like it shouldn't - I'd expect that this might create some problems.

In a similar vein this isn't reduced eagerly:
```ts
declare const obj: NoInfer<{ a: string, b: number }>
type Mapped/**/ = { [K in keyof typeof obj]: K }
```

If we hover over `Mapped` we'll see the mapped type instead of its evaluated form (`isGenericMappedType` return `true` since `isGenericIndexType(constraint)` is `true`).

I didn't analyze this to its full extent but I think that it's very likely that this has a direct impact on:
```ts
class Animal {
  move() {}
}
class Dog extends Animal {
  woof() {}
}
class Cat extends Animal {
  meow() {}
}

type K<T> = T extends any ? { [k in keyof T]: string } : never;

export type UnionToIntersection<U> = (
  U extends any ? (k: U) => void : never
) extends (k: infer I) => void
  ? I
  : never;

function doSomething<T extends Dog | Cat>(
  getDefault: (x: any) => T,
  value: UnionToIntersection<K<NoInfer<T>>>,
) {}

doSomething((x) => new Dog(), { woof: "", move: "" });
doSomething(() => new Dog(), { woof: "", move: "" });
```

Both calls fail here while the second one succeeds without `NoInfer`. I think that both should succeed with `NoInfer` though. At the moment, the first call infers `Dog | Cat` and the second one infers `Dog` but they both should infer `Dog`

craigphicks:
@Andarist

Does your mean that your plan is that the programmer must manually strip the NoInfer in the implementation of a function with NoInfer parameters, before using it where it might be inappropriate (and presumably trigger a new type of diagnostic), i.e. using it as an argument for a parameter the will be inferred?  That would be tedious.

It might be better just to auto-strip all `NoInfer` types automatically and immediately - or to put it another way, `NoInfer` would be a transient state.  Specifically,
1.  NoInfer would only be allowed as an attribute for call parameter types, not for stand alone types.
2. Inference would take place as coded in this pull, detecting on entry to `inferFromTypes`. 
3. In the function implementation, flow would have to access a parameter's type with an extra level of indirection ensure the `NoInfer` is removed.

ahejlsberg:
@craigphicks 

> Inside function `f` what are `typeof a`, `typeof b`, and what does the inner `f` resolve to?

`typeof a` resolves to `T` and `typeof b` resolves to `NoInfer<T>`. Thus, in the inner call to `f`, the inferred type argument is `NoInfer<T>`.

> Does NoInfer get stripped by

No, the `NoInfer<infer TN>` blocks inferences to `TN`, so the conditional type resolves to `unknown`.

> It might be better just to auto-strip all NoInfer types automatically and immediately - or to put it another way, NoInfer would be a transient state, or perhaps even a token noinfer, sibling to infer.

It's not that simple. `NoInfer<T>` can occur in nested positions, e.g. `{ a: T, b: NoInfer<T> }`, so we would have to exhaustively traverse types to find all `NoInfer<T>` occurrences and erase them--which in turn would produce new structural types that cause more work for the type checker. And recursive types would be a major headache. It's simply not feasible to do at depth, and it would be odd to only do shallow erasure.

tests/cases/conformance/types/typeRelationships/typeInference/noInfer.ts
ahejlsberg: > Any particular reason why those can't be erased?

We don't have infrastructure to consistently determine if an object type contains generics. It would require deep traversal, detection of recursion, and so on. Basically the equivalent of the traversal we do in `inferTypes`. We could consider using `couldContainTypeVariables` to guide erasure of `NoInfer<T>`, but those are optimization heuristics that all of a sudden would become observable.

> At the moment this, in part at least, creates problems with mapped types and `keyof`.

Agreed, we should erase `NoInfer<T>` when it is the argument of `keyof`. I'll make that change.

craigphicks:
@ahejlsberg 

Is there any way a `NoInfer` type in flow could ever unintentionally result in a target *parameter* being inferred as as a `NoInfer`, which then creates a problem in the form of user-unexpected result or an error?

For example a function type
```
declare function<T>(a: T): (a:T):T  
```
could be resolved to `f<NoInfer<T>>(a: NoInfer<T>): NoInfer<T>`? (Which I guess would result in a diagnostic error?).

Andarist:
Could we get a new TS playground for this? It would make it easier and faster to tinker with edge cases :p

DanielRosenwasser:
I have been summoned to summon TypeScript Bot.

@typescript-bot pack this

src/compiler/checker.ts
```diff
@@ -6710,7 +6712,8 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 return visitAndTransformType(type, type => conditionalTypeToTypeNode(type as ConditionalType));
             }
             if (type.flags & TypeFlags.Substitution) {
-                return typeToTypeNodeHelper((type as SubstitutionType).baseType, context);
+                const typeNode = typeToTypeNodeHelper((type as SubstitutionType).baseType, context);
+                return isNoInferType(type) ? factory.createTypeReferenceNode("NoInfer", [typeNode]) : typeNode;
```
DanielRosenwasser, line 6716: What happens in declaration emit when something like this gets serialized?

```ts
// foo.ts
export const f: <T>(x: T, y: NoInfer<T>) => bool;

// bar.ts

import { f } from "./foo.js";

type NoInfer<T> = number;

export const g = f;
```

tests/cases/conformance/types/typeRelationships/typeInference/noInfer.ts
Andarist: The weird case from the above works now 🎉 [TS playground](https://www.staging-typescript.org/play?ts=5.4.0-pr-56794-29#code/MYGwhgzhAECCB2BLAtmE0DeBYAUNayA9gG4CmAFAJSYC+udOokMAIoQObSkAeALqfAAmMBCjSZc+AO6FCAMyq16uJlGgBhMLy58BwuElTpseAqUJTFGBg1y8AngAdS0ANIAeACoA+aAF5oTx1+IRgweHtoAH5MaABtAGtoRHhoBNJ7eUCAXQAuaAheACcUzhpofPhSMiKAblxcHkdCIu0HZ2gAVSRCeE9CAEl4fiKIUmBeRF73Tt8A8kku4L0wiOjocgT8zuo-X2JCREEK6Cqa3GoeEP1N-JS5UiLoAd39w8FFmIHFyurH+pwuDkAFd4BMpqlBIQAMqEZCkXgAC1KXmWoWgbE4AB8NFpvAtTOwESxSHIwMCQLx8uRuPlwvZXoEADSLYhoYGkbY9PqDYaPMbg6YeAByvIeRS83ilLJw1GsDRwUNh8KRpXINMZVSkGI4VCZsRk8nyACJjfqiGQTcboDRKAClXCEcj4Ox1ZrSNrMXqDbI5FbzSROdBTTa7bggA). 

> Agreed, we should erase NoInfer<T> when it is the argument of keyof

Isn't this erasure a little bit too eager now? [TS playground](https://www.staging-typescript.org/play?ts=5.4.0-pr-56794-29&ssl=7&ssc=9&pln=1&pc=1#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXwxAGcMBGAHgBV4QAPQ1YI+Ab3igH4AueUgGngAjbvABM8AL4A+ABSwA5jwDWIAJ45E8SgEoelALAAoMHhLw4RZBDLwAvAWJkZAcijPtRgPSf4v+AD0OIyNQSFgEFHRsPAcSUSoaehBGZjZOHn4hEXFpORhFeBV1TQA5HABJVEQQGCopXS0jE1QzCysMcXtCOJc3D0NvPwCgwyA)

```ts
declare function test1<T extends { a?: 1, b?: 2 }>(arg: keyof T): T
const result1 = test1('a')
//    ^? const result1: { a: any }

declare function test2<T extends { a?: 1, b?: 2 }>(arg: keyof NoInfer<T>): T
const result2 = test2('a')
//    ^? const result2: { a: any }
```

If we use `NoInfer<keyof T>` then it falls back to the constraint - shouldn't the same happen with `test2`? I see that `keyof` part here isn't wrapped with `NoInfer` but `T` is wrapped and `T` is the target of the inference here. So this creates a weird precedence in which a wrapped inference target is still inferred.

craigphicks:
My gut feeling is that the `infer` vs `noinfer` attribute belongs to the call, rather than to the argument to the call.
Therefore it would be better to specify like this:
```
declare type F<T, Arg0 extends {a: infer T, b: (x: noinfer T)=> noinfer T}> = (a: Arg0) => Arg0;
```
Perhaps the `noinfer` on the return type is not necessary, I'm not sure.
We could also put `infer` on a parent so that it's children are never searched.

```
declare type F<T, Arg0 extends {a: infer T, [Symbol.iterator]: noinfer any}> = (a: Arg0) => Arg0;
```

tests/cases/conformance/types/typeRelationships/typeInference/noInfer.ts
Andarist: inferring `T` through `keyof T` is already a weird-ish thing so perhaps this is a better demo of the problem with reverse mapped types:
```ts
declare function test1<T>(
  o1: {
    [K in keyof T]?: [T[K]];
  },
  o2: {
    [K in keyof T]: T[K];
  },
): T;
const result1 = test1({ a: [1] }, { a: 1, b: 2 });
//    ^?

declare function test2<T>(
  o1: {
    [K in keyof NoInfer<T>]?: [T[K]];
  },
  o2: {
    [K in keyof T]: T[K];
  },
): T;
const result2 = test2({ a: [1] }, { a: 1, b: 2 });
```

I'd expect the second call to succeed here. Wrapping `T` directly with `NoInfer` here is more intuitive than having to wrap whole `keyof T` - solely because `T` is the inference target and not `keyof T`.

jakebailey:
[surprised]     Jake Bailey reacted to your message:
________________________________

jakebailey:
[0]     Jake Bailey reacted to your message:
________________________________

craigphicks:
Case from 2.8 release documentation + NoInfer 
```
type Bar<T> = T extends { a: (x: infer U) => void; b: (x: NoInfer< infer U>) => void }
  ? U
  : never;
type T20 = Bar<{ a: (x: string) => void; b: (x: string) => void }>; // string
type T21 = Bar<{ a: (x: string) => void; b: (x: number) => void }>; // 5.3.2: string & number => NoInfer: never (expecting string)
```
In the NoInfer version, `T21` is `never` but was expecting `string`.

src/compiler/checker.ts
ahejlsberg: With the latest commit this will now emit `globalThis.NoInfer<T>`, similar to what we do for other global types in conflict situations.

tests/cases/conformance/types/typeRelationships/typeInference/noInfer.ts
ahejlsberg: Yeah, erasure was a bit too aggressive. I think the proper solution is to transform `keyof NoInfer<T>` into `NoInfer<keyof T>`. The latest commits implement that.

src/compiler/checker.ts
DanielRosenwasser: Do we have a test for that? Same with Uppercase etc.

ahejlsberg:
@typescript-bot pack this

src/compiler/checker.ts
ahejlsberg: We don't, at least not that I can tell.

tests/cases/conformance/types/typeRelationships/typeInference/noInfer.ts
Andarist: An interesting solution to the problem! I think it might be worth adding some of the test cases I mentioned here since based on the emit/display alone it's not quite apparent **why** such transformation happens and what could go wrong if that would be missing. I **very often** disable some parts of the compiler to learn what tests break without them to learn how things work and what I might have to support in my PRs etc. Tests like that helps me quite a bit.

For convenience, I'm putting all of them in the details here - feel free to pick up/ignore any of them:

<details>
<summary>Potential test cases</summary>

```ts
class Animal { move() {} }
class Dog extends Animal { woof() {} }
class Cat extends Animal { meow() {} }

type K<T> = T extends any ? { [k in keyof T]: string } : never;
type UnionToIntersection<U> = (U extends any ? (k: U) => void : never) extends (k: infer I) => void ? I : never;

function doSomething<T extends Dog | Cat>(getDefault: (x: any) => T, value: UnionToIntersection<K<NoInfer<T>>>) {}

doSomething((x) => new Dog(), { woof: "", move: "" });
doSomething(() => new Dog(), { woof: "", move: "" });

declare function test1<T extends { a?: 1, b?: 2 }>(arg: keyof NoInfer<T>): T
const result1 = test1('a')

declare function test2<T>(
  o1: {
    [K in keyof NoInfer<T>]?: [T[K]];
  },
  o2: {
    [K in keyof T]: T[K];
  },
): T;
const result2 = test2({ a: [1] }, { a: 1, b: 2 });
```
</details>

src/compiler/checker.ts
ahejlsberg: Test added.

andrewbranch (approved):

Xample:
I'm surprised not to see more examples where the type is infered from the returned value. Here is a (simplified) [real world example of a code](https://www.typescriptlang.org/play?target=6&ts=5.4.0-beta&ssl=31&ssc=2&pln=1&pc=1#code/PTAEBUGVQHgVgHQBYBQKBmBXAdgYwC4CWA9tqAOYCm+AagIYA2mlM4oAvKAM74BOh2cgD4AFAEoAXBFABvFKFC9qmXmRwATSugGV1oOlwgBuFAF80WPEVIVqAOUwBbAEaVeASWwMd4qdieuvLLyisqqtrSMzOJGoCCg2MSgbrzEvAA00gKgAugpuvqG-i5uoCIA7pSEvHquABZ0AG4kvGJmFjgEJGRU+A4lvACqXALk9Px0zgyUvgkBbnIouKQ8oI1RlBwR9EwzYrHxKWmZbPh0ANaUhvh1m5rodJgM+Nx8o2sbIUr4KmTruyZzCh4lBQEJ2IhUBhOtYetQdsw7MRPHleKwtjx+IJRJJQEiUW5WEJggpvr9QBotDo9AZjO1oVZuhF+oFPN5sJR8dhUbNioESaEfuFegjOcjuW4YnEwEcMtIzpdrrdQPdHs9XljyB9dvTLF0bL0WW5hqNxoRJtMuTzcXyFmhlthVv9mFsRRsrZL9tLkrxUnLThcrqAbnctGqXpj3s7KF8wn8NoCgA) I faced where it was unpossible to catch errors without the `Noinfer` keyworld :

```typescript
// TS <5.4

function getValue<T = string>(): T {
  return undefined as T;
}

function getNumberInline(): number {
  return getValue(); // no error, T in inferred as number (weird behavior)
}

function getNumberUsingVariable(): number{

const value = getValue(); // error, T takes the default string value
  return value;
}

// TS >=5.4

function getValueNoInfer<T = string>(): NoInfer<T> {
  return undefined as T;
}

function getNumberInlineNoInfer(): number {
  return getValueNoInfer(); // error, T takes the default string value
}

function getNumberUsingVariableNoInfer(): number{

const value = getValueNoInfer(); // error, T takes the default string value
  return value;
}
```

dhmk083:
It would be nice if the following pattern could work:

```ts
declare function create<T>(fn: (get: () => NoInfer<T>) => T): T

create(get => ({
    id: 123,
    getId() {
        return get().id
    }
}))
```

Right now `T` is inferred as `unknown`.

If I understand `NoInfer<T>` correctly, it should give a hint to a compiler to not try to infer `get`-parameter type and instead infer `T` from type of returned object.

Xample:
@dhmk083 shouldn't you [write](https://www.typescriptlang.org/play?ts=5.4.0-beta#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXzDigxAB4AVAPgApFUAueagcxA0eoEp4BeS+AHI4AkqkQgYFStz7xynRuQBQSsHgDOGeKy08mnWdQDeS+GfhZgjAIwAmAMwAaU+Z3DgXeCfM-4cDMgw+DpcAHSWLmYAvkpRnCqEIMQgLGzxQA):
```typescript
declare function create<T>(fn: (get: () => NoInfer<T>) => T): T

const get = ()=> ({
    id: 123,
    getId() {
        return get().id
    }
})

create(get)
```

Andarist:
@dhmk083 currently, TS can't defer this sufficiently enough, it has to assign the contextual type for `get` before it even looks into your function. To do that it instantiates the type of this parameter and all of the type parameters in its scope (and `T` is its outer type parameter).

It's probably not impossible to improve this by deferring the instantiation of those parameter types that are only observable by the return types. I don't feel that it's particularly related to `NoInfer` though.

PR #56795, opened 2023-12-15
Give suggestion for @types/bun when Bun global is not found

jakebailey:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/67827 will introduce `@types/bun`. This PR makes TypeScript aware of the global `Bun` name, suggesting the types package now that its name is settled.

I thought we had something similar for modules such that we could see `bun:` prefixed modules and also suggest the types, but I think I was wrong.

### DIFF
--- a/src/compiler/diagnosticMessages.json
+++ b/src/compiler/diagnosticMessages.json
@@ -3735,6 +3735,14 @@
         "category": "Error",
         "code": 2866
     },
+    "Cannot find name '{0}'. Do you need to install type definitions for bun? Try `npm i --save-dev @types/bun`.": {
+        "category": "Error",
+        "code": 2867
+    },
+    "Cannot find name '{0}'. Do you need to install type definitions for bun? Try `npm i --save-dev @types/bun` and then add 'bun' to the types field in your tsconfig.": {
+        "category": "Error",
+        "code": 2868
+    },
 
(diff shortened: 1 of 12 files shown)


DanielRosenwasser (approved):

DanielRosenwasser:
> I thought we had something similar for modules such that we could see `bun:` prefixed modules and also suggest the types, but I think I was wrong.

This would be great to get in - here or separately is fine.

jakebailey:
> This would be great to get in - here or separately is fine.

We don't have it for node, so I'll file an issue for module specifier based hinting.

andrewbranch (approved):

src/compiler/diagnosticMessages.json
```diff
@@ -3735,6 +3735,14 @@
         "category": "Error",
         "code": 2866
     },
+    "Cannot find name '{0}'. Do you need to install type definitions for bun? Try `npm i --save-dev @types/bun`.": {
```
colinhacks, line 3738: Could we capitalize Bun when used in prose?

jakebailey: That is entirely up to you all, I assumed lowercase but I do see that the website capitalizes it consistently, so I can fix it here.

colinhacks: Cheers, thanks.

PR #56805, opened 2023-12-16
Add types for Object.groupBy() and Map.groupBy()

### LINKED ISSUE #47171: Add Array.prototype.groupBy[toMap] to JS lib definitions.
[The proposal for `Array.prototype.groupBy` and `Array.prototype.groupByToMap`](https://github.com/tc39/proposal-array-grouping) reached stage 3 recently.

Search terms: proposal-array-grouping array groupBy ernest

# lib Update Request

With advancement into stage 3, it's time for JS engines, transpilers and toolings to start implementing support for new features and provide feedback for the proposal's advancement into stage 4.

## Configuration Check

This is so new that it needs to go into `ESNext` target.

Missing `Array.prototype.groupBy` and `Array.prototype.groupByToMap`.

## Sample Code

```js
const array = [1, 2, 3, 4, 5];

// groupBy groups items by arbitrary key.
// In this case, we're grouping by even/odd keys
array.groupBy((num, index, array) => {
  return num % 2 === 0 ? 'even': 'odd';
});

// =>  { odd: [1, 3, 5], even: [2, 4] }

// groupByToMap returns items in a Map, and is useful for grouping using
// an object key.
const odd  = { odd: true };
const even = { even: true };
array.groupByToMap((num, index, array) => {
  return num % 2 === 0 ? even: odd;
});

// =>  Map { {odd: true}: [1, 3, 5], {even: true}: [2, 4] }
```

Sample code from https://github.com/es-shims/Array.prototype.groupBy

## Documentation Link
Proposal: [proposal-array-grouping](https://github.com/tc39/proposal-array-grouping)

Shim implementation that can the code can be tested against: [es-shims/Array.prototype.groupBy](https://github.com/es-shims/Array.prototype.groupBy)

bakkot:
Fixes #47171.

The tests are pretty simple because the types are simple, but I'm happy to expand them.

Try them on [the playground](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgPICMBWEFgMID2IAzmFAK64FTIDeAUMk8gPQBUbjzybyA4lALkADsWQBbCOPTQxBGMjghkwSFDjoANikQJqAE1ABzZGAKmAFiigQw5KMoBucTeRTzLKYXGLEI+5AQXTXREAGsAOi5mXgABb3VxFUhxMQBBZVVoDW0o7h5keLhE5DCIAE8AZQhtKho0wODQhDDkAHcLYAQLduBNTWQZFRBHAjKAmGpkCEQerKTQZKliPO42FmjkI0ERACFygB4AaWmAD0gQfTEABUFhaDByo4qAGmQAFQA+AApN7nniAAuZAASTUOQgBy+Lz+zDKVRqODMUGB33mwPeb1A+ggp2BIHI0mgAEpkABeT7II4w7jE4HXYpgYAuA4AJRwBmOb3eAG0ALqfT4AbnoAF96PRQGp4EhkABZODCQgkMiUZF0TbsTj5XgCISiCRSGRQOQKJRLdRaHQIPRQQwgExmTzIGx2BzIZyudwKMBWZDeXz+Rr9ZqRP5xBJwBYpdKZcFW1YxQqRpLw6q1dUNIIh8LtTrdXr9QYoUCjcbISY0GYF+bDJapRNMdabbb6-Zcj4-WFMAHAsHZK1Qz40-JMNOIuqo9EfLGXXH4wnG0kUqkjph0+WKju8gUi8X0PQqwY+LrktBYJERVt7crfHkABjeACY3gAOPlvU7kylfg7IACsyAAPzIAA5MQ4jBKByDAqBmjFEYECgcSIosCw3AAHpAUAA).


huw:
I’d probably also credit @karlhorky and @nikeee ❤️

bakkot:
Done. Hope that's alright @karlhorky @nikeee.

DanielRosenwasser:
@typescript-bot pack this

DanielRosenwasser:
This seems good overall, though users will have to opt in to the more precise inference for keys.

I would also think to stick with the same precedent we have elsewhere for type parameter names (`K` and `T`), and I *feel* like specifying the key first might be more desirable if we ever had the ability to only provide some arguments while letting the others be inferred.

I'll let others weigh in.

DanielRosenwasser (approved):

bakkot:
> This seems good overall, though users will have to opt in to the more precise inference for keys.

What do you mean by "more precise"?

`Object.groupBy([0, 2, 8], x => x < 5 ? 'small' : 'large')` correctly infers the key type as `'small | 'large'`, which is what it ought to be.

Similarly `type Employee = { name: string, role: 'ic' | 'manager' };
Object.groupBy([] as Employee[], x => x.role);` infers the key type as `'ic' | 'manager'` - again, what it should be. (These are both in the tests.)

> I would also think to stick with the same precedent we have elsewhere for type parameter names (`K` and `T`), and I _feel_ like specifying the key first might be more desirable if we ever had the ability to only provide some arguments while letting the others be inferred.

SGTM, will push this up later if a maintainer doesn't get to it first.

src/lib/esnext.object.d.ts
```diff
@@ -0,0 +1,11 @@
+interface ObjectConstructor {
+    /**
+     * Groups members of an iterable according to the return value of the passed callback.
+     * @param items An iterable.
+     * @param keySelector A callback which will be invoked for each item in items.
+     */
+    groupBy<K extends PropertyKey, T>(
+        items: Iterable<T>,
+        keySelector: (item: T, index: number) => K,
+    ): Partial<Record<K, T[]>>;
```
sandersn, line 10: why is the return type `Partial`?

sandersn (changes requested):
Looks right, just one question about Partial.

src/lib/esnext.object.d.ts
bakkot: Because [you might not actually get all of the keys](https://github.com/microsoft/TypeScript/issues/47171#issuecomment-1848451195).

As a trivial example, consider the case that the iterable is empty. Then the resulting record will have no keys at all. Typing as `Record<K, T[]>` [would mean that](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgPICMBWEFgMID2IAzmFAK64FTIDeAUMk8gOZQHkAOAQgJ4A8AaWQQAHpBAATYsgAK7TtDC9BEXgBpkAFQB8ACkbMjwSAFtiALmQBJSFDjoANhH671ho0wDWagMoRnKigrPRMIUystTVBJMSsQclN0aABKZABeHWRBdyMUqwAlHGpJIU0tAG0AXR0AbnoAX3p6BCJSZHQ4YmAEDLQsHDAAOjYOHl49CoAOKs1RDKz5-mQAVmQAfmQAcmJTOEdHLeQrLcc4KBYILZT6+gB6O+QAKheX5DAAC2AZYg+OR0kHRQcBAIig7CgAEJ6J1ugghrt9o4hs4QCxPvUgA) e.g. `Object.groupBy(vals, x => x < 5 ? 'small' : 'large')` would be typed as _always_ having `small` and `large` keys, such that `result.small.length` would check without errors, which is misleading.

sandersn: Ah, right, I forgot that `Map<K, T[]>` has `get: (k: K) => T[] | undefined` so was confused by the difference in types.

sandersn (approved):

sandersn:
@bakkot I'll merge this after you merge from main (or rebaseline, whichever).

bakkot:
@sandersn Done.

nikelborm:
Hi, I just noticed this and think it needs some improvements, since it doesn't handle cases like these:

```typescript
function tester<K extends number | string>(k: K, index: number): K extends number ? 'numbers' : 'strings' {
    return (typeof k === 'number') ? 'numbers' : 'strings';
}


// expected numbers ✅ 
const a1 = tester(123, 123);
//    ^?  const a1: "numbers"


// expected strings ✅ 
const a2 = tester('123', 123);
//    ^?  const a1: "strings"

// expected error ✅ 
const a3 = tester(true, 123);
//    Argument of type 'boolean' is not assignable to parameter of type 'string | number'.

const basic = Object.groupBy([0, 2, 8, 'asd'], tester);
//    ^?  const basic: Partial<Record<"numbers" | "strings", (string | number)[]>>


// expected number[] | undefined ❌
basic.numbers
//     ^?  (property) numbers?: (string | number)[] | undefined


// expected string[] | undefined ❌
basic.strings
//     ^?  (property) numbers?: (string | number)[] | undefined
```

bakkot:
@nikelborm Typescript's types don't generally go for that level of type-level logic. It might be possible in theory but it would significantly complicate the types. The current types aren't wrong, just slightly imprecise.

nikelborm:
They do go, and do it pretty well

Screenshot from 2024-01-12 19-50-37

nikelborm:
It neither does "hard" logic
```typescript
function tester(k: number | string, index: number): k is string {
    return typeof k === 'string';
}

// expected {'true'?: string[], 'false'?: number[]}
const basic = Object.groupBy([0, 2, 8, 'asd'], tester);
//  Type 'boolean' is not assignable to type 'PropertyKey'.
```
nor easy (it doesn't seems hard to support just booleans)
```typescript
function tester2(k: number | string, index: number): boolean {
    return typeof k === 'string';
}

// expected Partial<Record<"true" | "false", (number | string)[]>>
const basic2 = Object.groupBy([0, 2, 8, 'asd'], tester2);
//    Type 'boolean' is not assignable to type 'PropertyKey'.
```

bakkot:
@nikelborm You're welcome to send a followup PR and see if the TS team accepts it.

nikelborm:
@sandersn what do you think about this?

huw:
Hey there 🙂 You’ve raised 2 separate issues here:

1. `groupBy` actually accepts all return types that are coercible to `PropertyKey`, and we should widen the types
2. That `groupBy` doesn’t narrow on the real-world output types of key selector functions

## `PropertyKey` coercion
You are correct to identify that, under spec, `Object.groupBy([1, 2, 3], () => true)` should technically be typed as `{ true?: number[], false?: number[] }`.

We can see this because the [array grouping proposal](https://tc39.es/proposal-array-grouping/) specifies that outputs should be coerced to property keys if possible, and it refers to [`ToPropertyKey`](https://tc39.es/ecma262/multipage/abstract-operations.html#sec-topropertykey), which specifies that booleans should be coerced to the literals ‘true’ and ‘false’.

TypeScript actually manages this natively for numeric types, which are usually used to index arrays (which are sort of objects under the hood)—that’s why `PropertyKey extends number`, even though the spec says [property keys can only be strings or Symbols](https://tc39.es/ecma262/multipage/ecmascript-data-types-and-values.html#sec-object-type). 

Although other types are coercible to property keys, TypeScript seems to have preferred simplicity/sanity/ease. Challenging this is much more philosophical and would affect a large amount of TypeScript, so at the very least you’ll have to raise a new issue. It would affect a lot more than `Object.groupBy()`.

## Narrowing key selector outputs
I am by no means a TypeScript expert, but I don’t think there’s a way to do what you’re asking for. To recap, you correctly noted that (for example) `groupBy([1, "2", 3], (item) => typeof item === "string" ? "string" : "number")` implicitly types `keySelector` as `(item: string | number) => "string" | "number"`, which doesn’t have enough type information to determine the output type from the input type. 

To fix this, you added a generic parameter to determine the output type from the input type, such as `groupBy([1, "2", 3], <K extends number | string>(item: K): K extends string ? "string" : "number" => typeof item === "string" ? "string" : "number")`. I don’t think there’s a way to achieve this _without_ generics.

What you need, concretely, is a type that can map specific `K` keys to `T` values. A generic function type can do this by itself, as we’ve seen. But the issue is that this would require higher-order generics to infer the output type from the input type within the already-genericised function. Here’s some fake code that is doing roughly what you mean:

```ts
declare function groupBy<K extends PropertyKey, KeySelector<T> extends (item: T, index: number) => K>(
    items: Iterable<genericof KeySelector>,
    keySelector: KeySelector,
): { [key in T]?: KeySelector<key>[] };
```

microsoft/typescript#1213 seems to be the canonical issue for higher-kinded types, in case you want to follow it (or someone there is looking for a use-case here).

Alternatively, you might get clever and create a mapping from `K` → `T` directly, using mapped object types:

```ts
declare function groupBy<TypeMap extends {}, Key extends keyof TypeMap = keyof TypeMap, KeySelector extends (item: TypeMap[Key], index: number) => Key = (item: TypeMap[Key], index: number) => Key>(
    items: Iterable<TypeMap[Key]>,
    keySelector: KeySelector,
): { [key in Key]?: TypeMap[key][] };

groupBy<{ "string": string; "number": number }>([1, "2", 3], <K extends number | string>(value: K): K extends string ? "string" : "number" => typeof value === "string" ? "string" : "number");
```

This actually works! But (a) it’s a hack, and TypeScript doesn’t accept hacks in the core library and (b) it requires you to always specify `TypeMap`, which isn’t possible to infer, and as noted above TypeScript doesn’t usually accept types that are complex to use in this way. Also, perhaps more importantly, is that it’s no more ergonomic than just overriding the return type of `groupBy` yourself.

[Here’s the TypeScript playground](https://www.typescriptlang.org/play?ts=5.4.0-dev.20240112#code/PTAEAoElQQwW1AcwPYEsB2jQBdmmQG4CmATgM7YzZE4AWNADicgEYA2RCM6AJqCzRJEYbNgE98LAqmQBXMgBpQAKBCg4yCuP5EAxjHk1U2AORlQ9NgwBmstjjxwauUAHcSMBuon0YfGMyyvKDGsKJKqmAsstgWyAxEtqISoZYM5rpsAajWEtj0oEysHAjWyCT4+aSSZKTE5ACUypGgAAYA8iwAVnrYAHSIgQwAQmKtIeY8ellCfPmo5q4wYgBcylOZATS26LrYMuhIQ6MAPADSoEQAHtS85gAKzAkk2GJnRGJKACoAfODKoEBIWocDIK1AkGoHnYRBOvwUAKBAGsPgBlIgcPblcHgYyccFfJQYKZXcHoWRwAQkBqgAC8P1AZwRDXB9wC+xEJwASnpyjxzt8ANoAXR+PwA3M01F9aAs3KhRBYYMQ6M4xAlQABvUAAIgoJAwiB1OP1htAAB9QOTKaQGiLxbrrVTjRBTZgLVaKVS7cKHQBfFrDGIhQ5MGB7VC6IhKVw0JHoZCuYFmLW6t1G8Hp+2Or2kF1O0jZgO6ZDoCj4GIMGIARjpRzkIzE4EF1aUOoATDqlABmYVKcAEESyIg0+k4dVEZDWUCDtjDum02lp7AGzA60AAfmXq6NoHBOoLJB1DUlLRlctcCvstGVzgKrw12r1K8NLvA6Y9h59DoPuaPJpfd1LS-Isz1lcw5RLEghD2ENCg8CMo2UEsy1iORsCrbB2zrQYG1GZtW11Tsez7CBZ2HUcGQAWSoWg+g8XhkDgcAaQZAAGPoAFZN23V89xzG0jxPKUwEgQ5rHDbAlBYZB8lAWpMVwcglRVKpxw1XEQUzQCsGAv9KN4tcPV-QSdTAohUAqB8iEmUsTFiFDKAwS4E1kRBaBDMoSDgKgDgcTzqjU9DMIgw41KWCRYzcbhsBaAAZIhTHMQwkCIdBSEjMFkNLctgpibscOOJsWzbYjQF7JRzkuG40p4cxDw9dM-nIohwTOFlGWq246vknSeOfHd133Q91zHRFAWsqcZyHGhFyXAa+K3BajOGv9j1PNQAHUiBMFUSCCEMXDUooYS4cwAAF0FQFE2Bk7zeqoIg+GMB0qiECZQCCQw5gnTzyh8-ZSwsUgiHixKUxXCRuGeuAihVVpcNkRtxhcDBrGqaxmAQKpahE0AEqSh6Xn4CRags90cjcGh9EOHyUTad4xHRRTynGRA0oy3QN3WaYtlAHYIyBxHG3bOElAua5uoeJ5SFeRnxbRDFenKLravMTT8VAQkQxJMl9LpBkzj+cbgU4MEIShGAYThH4ESBUAUSZpWsRINrFZZkhmVZdlUE5HkoP5HlsFkEh0C+Cdznd5WSFt0A2Q8JwoTISOnY9n5BTY4URTFDawD6fPolieZbJs9B7LccokWy1CKwwmIABZCrwsR2wI0qu3K0iqsltXPUExqdOama2o6iWaruXqd369MhoE50DZUe3JunFqF3mmfp502eTOdYSwcJyHQF0IRfPdbhYBeHJI1969UHc0gAFo+UCidpIkeQzTUiTYPyKhQDEOQR9z76EVK0cOCQaIMEFI7YUKNHCeBDCFfyeVsBkB5psd6AtAaHGFqMbscIJyQNVhPTUfoFYSB7hPR2U1wFECIUuah05aGQPIczaOxCeoazgASQhnhBSM1IsSa4etBIGUZnWLhPCIF8IEUSXgwi+7egXozY29s8SgnBJCUgVsOAEOkVAgRtsTaOzYS7N2qdo5e1TNAj4cEBEbikXQvhMCRSgD9LnLW4EK4kCRIofgwYAGyBUs4PAZAEi6ByBIMBvCGDjCIPUPIqAnB7n+ChXKlYYjcSXLgsQ+Cnwz20juH8I0RFUjcX8EqREO4VVAN3cePUGqWiagOYejJR4cPMB+JaG9VqmQXibZe005yzUXIZXc3St78R3nmE8QA) I used for the above.

I hope that clears things up! I sort of hate to write stuff like this because it feels very conservative and self-justifying; if you have a problem with either of these answers I absolutely support starting a constructive dialogue with the TypeScript team to try and change these things—it may be the case that this is the issue that tips the balance :)

(Also, just as a matter of etiquette, I personally wouldn’t @-mention a maintainer in the future if another commenter disagrees with your take. I have to imagine it’s pretty annoying to be called in to adjudicate things, especially when another commenter has explained the next course of action (raise a new issue), and it’s pretty likely they already have notifications on for this thread and will see it in due time. In most open-source repos it’s a pretty fast way to get your position ignored.)

slorber:
Returning Partial `Partial<Record<K, T[]>>` is quite annoying in practice

See for example:

CleanShot 2024-06-28 at 10 55 40

We now have to deal with entries with undefined values, which technically is not really possible in practice.

Isn't there a way to express a Record where "only keys are Partial", but values are always defined when the key exists?


---

I tried various things but not sure to have a good solution.

Here's my last attempt:


```ts
type SubsetRecord<K extends PropertyKey, T> = {
  [P in K]?: T;
};


declare function groupBy<K extends PropertyKey, T>(
  items: Iterable<T>,
  keySelector: (item: T, index: number) => K,
): SubsetRecord<K, T[]>


type User = {name: string,age: number, type: "a" | "b" | "c"}

const users: User[] = [{name: "Seb",age: 42, type: "a"}]

const groups = groupBy(users, u => u.type);


// ✅ entries do not have undefined 
Object.entries(groups) satisfies [string,User[]][]

// ✅ 
groups["a"]!.toSorted()

// ✅ 
groupBy(users,u => "Seb")["Seb"]!.toSorted()


// ❌ should be forbidden: value can be undefined
// @ts-expect-error
groups["a"].toSorted()

// ❌ should be forbidden because keys do not match
// @ts-expect-error
groupBy(users,u => "Seb")["blabla"] 

// ❌ should be forbidden: value can be undefined
// @ts-expect-error
groupBy(users,u => "Seb")["Seb"].toSorted()
```

ktkiiski:
> Returning Partial `Partial<Record<K, T[]>>` is quite annoying in practice
> 
> See for example:
> 
> CleanShot 2024-06-28 at 10 55 40
> 
> We now have to deal with entries with undefined values, which technically is not really possible in practice.
> 
> Isn't there a way to express a Record where "only keys are Partial", but values are always defined when the key exists?

I agree with this.

In cases where the type of the group key `K` becomes `string` (instead of a more narrow type such as `"a" | "b" | "c"`), having the `Partial` there is unnecessary and causes these issues where, according to the typing, the object values could be `undefined`, even though they never are in practice.

This all comes down to TypeScript's limits to separate if the property is there at all vs. if they exist with value `undefined`, when dealing with this kinds of types, so for now that's something that we need to live with.

However, imo the developer experience would be greatly improved if the signature of `Object.groupBy` is changed to this:

```typescript
interface ObjectConstructor {
    /**
     * Groups members of an iterable according to the return value of the passed callback.
     * @param items An iterable.
     * @param keySelector A callback which will be invoked for each item in items.
     */
    groupBy<K extends PropertyKey, T>(
        items: Iterable<T>,
        keySelector: (item: T, index: number) => K,
    ): string extends K ? Record<K, T[]> : Partial<Record<K, T[]>>; // 👈 changed return type
}
```
The change is on the return value of the method: it now conditionally returns the regular "dictionary" type, e.g. `Record<string, User[]>`, without unnecessary `Partial`.

Now, the previously mentioned unnecessary `undefined` in the value type of e.g. `Object.entries` would be gone.

This return type still matches the implementation, but is more useful in cases where the grouping key is just the broad type of `string`.

Also, with this change, it also leaves up to the developer to choose their `noUncheckedIndexedAccess` configuration depending on how they would like to deal with possibly undefined values with indexed property access.

What do you think?

bakkot:
I think having the types incorrectly suggest that every key is present is worse than having the types incorrectly suggest that keys which are present might have value `undefined`. The former will accept incorrect programs, which you discover by running into bugs at runtime and can only fix by globally enabling `noUncheckedIndexedAccess` (which affects your whole program), whereas the latter will reject correct programs, which you discover at compile-time and can fix by adding `!` or other casts to the specific cases where you need it.

So I do not think that would be a good change.

PR #56826, opened 2023-12-18
Expose getUnknownType

sstchur:
Fixes #

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -1746,6 +1746,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         getVoidType: () => voidType,
         getUndefinedType: () => undefinedType,
         getNullType: () => nullType,
+        getUnknownType: () => unknownType,
         getESSymbolType: () => esSymbolType,
         getNeverType: () => neverType,
         getOptionalType: () => optionalType,
--- a/src/compiler/types.ts
+++ b/src/compiler/types.ts
@@ -5142,6 +5142,7 @@ export interface TypeChecker {
      * is `null`. Instead, use `type.flags & TypeFlags.Null`.
      */
     getNullType(): Type;
+    getUnknownType(): Type;
     getESSymbolType(): Type;
     /**
      * Gets the intrinsic `never` type. There are multiple types that act as `never` used internally in the compiler,


jakebailey:
This makes sense, but do you have some context for this?

sstchur:
@jakebailey, Sure, I can provide context.  Jason Gore was doing some work where he was writing an ESLint rule and needed access to the `unknown` type.  I recalled that I had done the work to expose `getNumberLiteralType` and `getStringLiteralType` and it occurred to me that perhaps `getUnknownType` should have been a part of that work.  So I just quickly threw together a draft PR, thinking that it might be useful to Jason (and perhaps also logically went w/ the PR I did last year).

I think Jason has actually abandoned the rule he was working on.  That doesn't necessarily mean this isn't still a worthwhile chang to consider though, but I'll leave that to the TS team to decide.  If you think `getUnknownType` should have been exposed in this original PR (https://github.com/microsoft/TypeScript/pull/52473), then I can do the work to finish out this change as well.

If you think there's no need to expose it, we can just abandon the effort.

jakebailey:
I suspect that if it had been included, we would have accepted it given we already have `getAnyType` exported.

But, since I sent and merged #56448 which exposed `isTypeAssignableTo`, maybe it's useful. Then again, anything is assignable to `unknown`, so maybe that's not the intended use. Hence, my curiosity in how you are able to use it.

sandersn:
We haven't had time to finish reviewing this in over a year. 
Since the Corsa API will be quite different from the Strada one, does it make sense merge this? I think it's better to close it and revisit what types we expose as we work on the Corsa API.

jakebailey:
We'll likely want this both before and after. I'm still surprised we didn't have this.

sandersn:
I'm not opposed to merging it here. @sstchur do you want to mark this as Ready for review?

sstchur:
Sure thing!

sstchur:
It's been so long here, that I don't recall if there is something more I need to do?  Do I need to add tests, or otherwise update this PR?

jakebailey:
You will need to update tests, yes. Should just be `npm test` and `npx hereby baseline-accept` and commit what changed. The PR is old enough that GitHub won't run CI on it, unfortunately.

jakebailey:
I foolishly forgot that I already approved and merged #60502, which does this. So, closing as done, oops.

Issue #53362, opened 2023-03-19
Improve string split return type of first array index


zardoy:
# lib Update Request

Please, consider improving user experience for `"".split(...)` in cases when string literal is passed, which I believe most users do e.g. `split(' ')` or `split('\n')`

## Configuration Check

My compilation *target* is `ES2015` and my *lib* is `the default`.

## Missing / Incorrect Definition

https://github.com/microsoft/TypeScript/blob/b70784ef8f3ba3fd21750b9acf275dd7edcc7fe7/src/lib/es5.d.ts#L460

Essentially duplicate of https://github.com/microsoft/TypeScript/issues/49635, but with https://github.com/microsoft/TypeScript/issues/49635#issuecomment-1163047914 handled:

```ts
split<T extends string | RegExp>(separator: T, limit?: number): T extends `${string}${infer U}` ? [string, ...string] : string[];
```

## Sample Code
 
This is super inconvenient when used with destructure:

```ts
const [firstLine, ...restLine] = str.split('\n')
// firstLine is possibly undefined, need add exclamation marks everywhere or `as [string, ...string]` above
```

I also wanted to do the same for when regexp literal is passed e.g. `split(/\n\r?/)` as only `split(new RegExp(''))` can result in empty array (if I'm not mistaken), which I believe no one does, but the literal value can't be captured anyway

## Documentation Link

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split#description (note)

P.S. I remembered of this issue after https://github.com/microsoft/TypeScript/pull/49682, which significantly improved type-checking experience by using less exclamation marks

EzioMercer:
[Original](https://github.com/microsoft/TypeScript/issues/61816)

I came up with more strict overloads when `limit` is `0` and when `separator` is `string`:

```TypeScript
type Split<Source extends string, Separator extends string> = Source extends `${infer Left}${Separator}${infer Right}`
    ? [...(string extends Left ? string[] : [`${Left}`]), ...Split<Right, Separator>]
    : Source extends ""
      ? Separator extends ""
          ? []
          : [""]
      : string extends Source
        ? string[]
        : [Source];

interface String {
    split<Source extends string, Separator extends string>(
        this: Source,
        separator: Separator,
        limit?: undefined,
    ): Split<Source, Separator>;

    split(separator: string | RegExp, limit: 0): [];
}
```

[...]

Issue #49635, opened 2022-06-22
String split return type


Forlini91:
# lib Update Request

## Configuration Check

My compilation *target* is `ES2020` and my *lib* is `ES2020`.

## Missing / Incorrect Definition
Method `split` of String currently has this signature
```
split(separator: string | RegExp, limit?: number): string[];
```

This method, when called with a `limit` argument = 0, returns an empty array, but, when called without `limit` argument or with a non 0 `limit` argument, returns an array with at least one element (the initial string).

This means the signature could be changed to acknowledge at least the "missing limit argument" case, where we are sure the array contains at least one element
```
split(separator: string | RegExp): [string, ...string[]];
split(separator: string | RegExp, limit: number): string[];
```

## Sample Code
Current situation:
```
const strArr = "123".split("=");  // strArr has type string[]
const item = strArr[0];   // item has type string | undefined
```

Expected:
```
const strArr = "123".split("=");  // strArr has type [string, ...string[]]
const item = strArr[0];   // item has type string
```

I called split without "limit", so I'm sure to find at least 1 element in the returned array

## Documentation Link
https://262.ecma-international.org/11.0/#sec-string.prototype.split

Josh-Cena:
`"".split("")`

alecvision:
@RyanCavanaugh @Josh-Cena I do think there is actually a bug here, if not the one OP is pointing to. Shouldn't the return type be `(string | undefined)[]` - making the elements of type `string | undefined`? Seems the undefined case is missed by TS.

Here is the behavior I'm getting:

```
const test1 = 'test'; // type: 'test'
const splitresult1 = test1.split(':'); // type: string[]
const firstelement1 = splitresult1[0]; // type: string
const secondelement1 = splitresult1[1]; // type: string

console.log(test1) // test
console.log(splitresult1) // [ 'test' ]
console.log(firstelement1) // test
console.log(secondelement1) // undefined

const splitresult2 = "".split(""); // type: string[]
const firstelement2 = splitresult2[0]; // type: string
const secondelement2 = splitresult2[1]; // type: string

console.log(splitresult2) // []
console.log(firstelement2) // undefined
console.log(secondelement2) // undefined
```

Josh-Cena:
@alecvision If you want that behavior please turn on `noUncheckedIndexAccess`. It's a general problem with index-accessing arrays.

MicahZoltu:
[...]

PR #56834, opened 2023-12-20
feat: refine .split() return type

castarco:
Work in Progress:
- [ ] Searching if there is an appropriate related issue
  - [ ] If not, prepare a properly described issue to describe the problem
- [ ] Prepare new specific tests

### DIFF
--- a/src/lib/es2015.symbol.wellknown.d.ts
+++ b/src/lib/es2015.symbol.wellknown.d.ts
@@ -207,7 +207,8 @@ interface RegExp {
      * @param limit if not undefined, the output array is truncated so that it contains no more
      * than 'limit' elements.
      */
-    [Symbol.split](string: string, limit?: number): string[];
+    [Symbol.split](string: string, limit: 0): string[];
+    [Symbol.split](string: string, limit?: number): [string, ...string[]];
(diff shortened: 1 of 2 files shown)


castarco:

fatcerberus:
> Searching if there is an appropriate related issue

I found two:
* https://github.com/microsoft/TypeScript/issues/53362
* https://github.com/microsoft/TypeScript/issues/49635

MartinJohns:
Your changed type definitions fail with `"".split("")`.

castarco:
@MartinJohns 
> Your changed type definitions fail with "".split("").

Yes, I noticed some time later 😞 . I was thinking if it was possible to cover those corner cases with some extra function overrides... but that won't work, because there's a more generic case:

```typescript
const strValue = 'whatever'
const a = strValue
const b = strValue

// I don't know any way to assert (via the type system) a == b
// In this particular case, it is conceivable that TS would be able to
// perform some smart inferences... but sadly, a and b could be completely
// opaque in some contexts.
a.split(b)
```

So I guess I'll have to close this one.

castarco:
My new attempt: #56841 .

PR #56841, opened 2023-12-21
feat: refine .split() return type

### LINKED ISSUE #49635: String split return type

### LINKED ISSUE #53362: Improve string split return type of first array index

castarco:
**NOTE:** ⚠️ Still work in progress. ⚠️

## PR Description
This PR refines `String.split(*)` return type (plus the `RegExp`'s cousin method).

As opposed to previous attempts, it does not try to characterise the "empty array result" as a special case, but quite the opposite, it characterises "`[string, ...string[]]` result" as a special case.

With this change, we only get `[string, ...string[]]` as return type when we can statically know for sure that we are not trying to execute `''.split('')`, **and** we do not set the second parameter (`limit`) to `0`.

In other words, the separator must be a literal for this to give as a "nice" return type. If we use a variable, we'll keep the more general `string[]` return type.

This (combined with the fact that we can't specialise interfaces for specific literals) means that there are cases where we still get non-empty arrays as a result but our return type won't be able to represent that fact.

## Other Relevant Details

I introduced the types `NonEmptyStringParam<>` and `NonZeroNumberParam<>`. I suspect that these 2 new types might be not well received. I'm not sure how to make the non-global.

## Pending tasks:
- [ ] Fix formatting issues
- [ ] Fix baseline tests
- [ ] Add new tests

## What this PR tries to fix
- Fixes #53362

### DIFF
--- a/src/lib/es2015.symbol.wellknown.d.ts
+++ b/src/lib/es2015.symbol.wellknown.d.ts
@@ -162,6 +162,10 @@ interface PromiseConstructor {
     readonly [Symbol.species]: PromiseConstructor;
 }
 
+// TODO: inline at usage site
+type NonEmptyStringParam<S extends string> = S extends '' ? never : S;
+type NonZeroNumberParam<N extends number> = N extends 0 ? never : N;
+
 interface RegExp {
     /**
      * Matches a string with this regular expression, and returns an array containing the results of
@@ -207,6 +211,11 @@ interface RegExp {
      * @param limit if not undefined, the output array is truncated so that it contains no more
      * than 'limit' elements.
      */
+    [Symbol.split]<S extends string>(string: NonEmptyStringParam<S>): [string, ...string[]];
+    [Symbol.split]<S extends string, N extends number>(
+        string: NonEmptyStringParam<S>,
+        limit: NonZeroNumberParam<N>
+    ): [string, ...string[]];
     [Symbol.split](string: string, limit?: number): string[];
 }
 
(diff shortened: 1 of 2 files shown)


MartinJohns:
No idea why you decided to open another PR instead of just changing your existing.

Nothing proposed here hasn't been proposed before. See also: https://github.com/microsoft/TypeScript/issues/53362#issuecomment-1478293505

> we have a pretty low appetite for making the standard library definitions extremely complex for the sake of NUIA

I'd argue your suggested change is very complex. See also: https://github.com/microsoft/TypeScript/issues/56333#issuecomment-1802306245

castarco:
@MartinJohns Ok. 🤷🏻 . I guess there's no point on fixing the tests then. (and sorry for the duplicate PR).

PR #56845, opened 2023-12-21
Improve re-use of old program if imports changed but structure remains valid

### LINKED ISSUE #56846: Program is not completely re-used when imports change but import graph is identical
### 🔎 Search Terms

program re-use, old program, oldProgram, imports changed, graph same, reuse result

n/a

### ⏯ Playground Link

_No response_

### 💻 Code

_No response_

### 🙁 Actual behavior

Consider an incremental program using `oldProgram` in the compiler API. 

TypeScript will sometimes NOT fully re-use the program when a source file changed its imports even though the module graph remains the same. It considers structure changed. This is reasonable, but in some cases new imports may not signify that resolution results or the import graph has changed. A new import with existing imported modules may be added. e.g.

```
import {Directive} from `@angular/core`; // part of first compilation
import * as i0 from `@angular/core`; // added in the second! same module name!
```  

### 🙂 Expected behavior

TypeScript may consider deduplicating module names and would not treat the structure as invalidated. Later resolution validation as part of the re-use strategy may still bail out if discovers changes, but most of the time this is very likely the same (unless e.g. an import mode is used using `with`- this is NOT new behavior though)  

PR: https://github.com/microsoft/TypeScript/pull/56845

devversion:
Currently if a source file changes and new imports may be added for similar existing modules, TypeScript will degrade from full program re-use to only safe modules. This can be improved with a seemingly small cost of having to sort and dedupe imports.

The use case is that in Angular we are optimizing for ideal program re-use and ocasionally we will run transforms on user code that already contain Angular imports. We may want to add additional Angular imports but aren't bothering with re-using the actual existing import declaration, so a new one is added. This new one now changes the length of `file.imports` and therefore degrades the program re-use.

Angular could consider re-using the existing import, but that might be difficult if it e.g. is a default import, or just technically non-trivial without breaking exiting references etc. It seems more natural if TypeScript would not consider this as a change in structure, as clearly the module names. outgoing-edges in the graph remain the same.

Mode overrides with imports may be an exception, but be aware that the deduplication does not have an effect on the later re-use resolution validation- in which case it would fail regardless. This is matching 1:1 with what would happen right now if an import mode changes. No new behavior, except that a comment for this is now added.

Fixes #56846

### DIFF
--- a/src/compiler/program.ts
+++ b/src/compiler/program.ts
@@ -142,6 +142,7 @@ import {
     getSetExternalModuleIndicator,
     getSpellingSuggestion,
     getStrictOptionValue,
+    getStringComparer,
     getSupportedExtensions,
     getSupportedExtensionsWithJsonIfResolveJsonModule,
     getTemporaryModuleResolutionState,
@@ -290,6 +291,7 @@ import {
     skipTrivia,
     skipTypeChecking,
     some,
+    sortAndDeduplicate,
     sortAndDeduplicateDiagnostics,
     SortedReadonlyArray,
     SourceFile,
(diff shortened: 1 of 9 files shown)


sheetalkamat (changes requested):
Even though this change makes it better for this one case, it also imparts extra cost in normal scenarios where we have to now sort all imports while checking if program can be reused. In most cases, doing linear check suffices and even in the scenario this is fixing, the module resolution would still be reused.

devversion:
@sheetalkamat would you be open for accepting a PR if we can still perform the check O(n) with extra linear memory pressure at program initialization? (can be GC immediately then)

sheetalkamat:
Not really since this adds up easily with many files and imports to have to check per each program construction phase.

devversion:
Are these files analyzed concurrently though? I was under the impression, no, and that there generally aren't many imports in a file

devversion:
@sheetalkamat Feel free to close though, if you feel confident there is no solution we can take

Issue #53116, opened 2023-03-03
Offer auto-imports for all files in package.json `exports` wildcard directories


pilaoda:
My npm lib module's package.json
```
  "name": "my-lib",
  "exports": {
        ".": {
            "require": "./dist/index.js",
            "types": "./dist/index.d.ts"
        },
        "./*": {
            "require": "./dist/*.js",
            "types": "./dist/*.d.ts"
        },
        "./**/*": {
            "require": "./dist/**/*.js",
            "types": "./dist/**/*.d.ts"
        }
    },
```

An app depends on this npm module and install it.
The vscode auto import only support types in `import {A} from "my-lib"` resolve to `my-lib/dist/index.d.ts`, and could not find the symbol and type in something like `import {B} from "my-lib/utils"` resolve to `my-lib/dist/utils.d.ts`.

pilaoda:
Sorry I just learn that `./**/*` is no need and only `./*` can be enough for this situation.
But even thouth I delete the unnecessary `./**/*` pattern and keep the `./*` pattern, the auto import still not working for something like `import {B} from "my-lib/xx/utils"` which should resolve to `my-lib/dist/xx/utils.d.ts`.
If I import it manually, it will be compiled successfully.
The real problem I think is that the ts language server in editor only search index.d.ts but ignore * pattern searching in a node_modules package.

pilaoda:
More detailed example, my project has below structure:
```
- node_modules
   - my-lib
      - dist
         index.d.ts
         utils.d.ts
      package.json
- src
   main.ts
```

index.d.ts:
```
export declare const A: number;
```

utils.d.ts:
```
export declare const B: number;
```

package.json
```
{
  "name": "my-lib",
  "exports": {
        ".": {
            "require": "./dist/index.js",
            "types": "./dist/index.d.ts"
        },
        "./*": {
            "require": "./dist/*.js",
            "types": "./dist/*.d.ts"
        }
    }
}
```

During my developing in src/main.ts, If I type "A", the editor will help me auto add `import {A} from "my-lib"`. If I type "B", the editor don't know where the symbol B is, so I need to add `import {B} from "my-lib/utils"` manually.

andrewbranch:
[...]

PR #56848, opened 2023-12-21
Let AutoImportProvider find non-declaration files in wildcard exports...

andrewbranch:
...which is important for internal packages consumed by bundlers in monorepos, which typically point their package.json entries to TS implementation files.

Fixes https://github.com/microsoft/TypeScript/issues/53116#issuecomment-1866150009

### DIFF
--- a/src/compiler/moduleNameResolver.ts
+++ b/src/compiler/moduleNameResolver.ts
@@ -3,6 +3,7 @@ import {
     appendIfUnique,
     arrayIsEqualTo,
     changeAnyExtension,
+    changeFullExtension,
     CharacterCodes,
     combinePaths,
     CommandLineOption,
@@ -2290,9 +2291,7 @@ function loadEntrypointsFromExportMap(
                     extensionsToExtensionsArray(extensions),
                     /*excludes*/ undefined,
                     [
-                        isDeclarationFileName(target)
-                            ? changeAnyExtension(replaceFirstStar(target, "**/*"), ".*", supportedDeclarationExtensions, /*ignoreCase*/ true)
-                            : changeAnyExtension(replaceFirstStar(target, "**/*"), ".*"),
+                        changeFullExtension(replaceFirstStar(target, "**/*"), ".*"),
                     ],
                 ).forEach(entry => {
                     entrypoints = appendIfUnique(entrypoints, {
--- a/src/compiler/moduleNameResolver.ts
+++ b/src/compiler/moduleNameResolver.ts
@@ -41,7 +41,6 @@ import {
     GetCanonicalFileName,
     getCommonSourceDirectory,
     getCompilerOptionValue,
-    getDeclarationEmitExtensionForPath,
     getDirectoryPath,
     GetEffectiveTypeRootsHost,
     getEmitModuleKind,
--- a/src/compiler/moduleNameResolver.ts
+++ b/src/compiler/moduleNameResolver.ts
@@ -2292,8 +2292,8 @@ function loadEntrypointsFromExportMap(
                     /*excludes*/ undefined,
                     [
                         isDeclarationFileName(target)
-                            ? replaceFirstStar(target, "**/*")
-                            : changeAnyExtension(replaceFirstStar(target, "**/*"), getDeclarationEmitExtensionForPath(target)),
+                            ? changeAnyExtension(replaceFirstStar(target, "**/*"), ".*", supportedDeclarationExtensions, /*ignoreCase*/ true)
+                            : changeAnyExtension(replaceFirstStar(target, "**/*"), ".*"),
                     ],
                 ).forEach(entry => {
                     entrypoints = appendIfUnique(entrypoints, {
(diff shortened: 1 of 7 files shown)


tests/baselines/reference/tsserver/fourslashServer/autoImportProvider_wildcardExports1.js
```diff
@@ -183,9 +184,10 @@ Info seq  [hh:mm:ss:mss] FileWatcher:: Added:: WatchInfo: /node_modules/pkg/b/pa
 Info seq  [hh:mm:ss:mss] FileWatcher:: Added:: WatchInfo: /node_modules/pkg/c/package.json 2000 undefined Project: /dev/null/autoImportProviderProject1* WatchType: File location affecting resolution
 Info seq  [hh:mm:ss:mss] Finishing updateGraphWorker: Project: /dev/null/autoImportProviderProject1* Version: 1 structureChanged: true structureIsReused:: Not Elapsed:: *ms
 Info seq  [hh:mm:ss:mss] Project '/dev/null/autoImportProviderProject1*' (AutoImportProvider)
-Info seq  [hh:mm:ss:mss] 	Files (5)
+Info seq  [hh:mm:ss:mss] 	Files (6)
 	/node_modules/pkg/a/a1.d.ts Text-1 "export const a1: number;"
 	/node_modules/pkg/b/b1.d.ts Text-1 "export const b1: number;"
+	/node_modules/pkg/b/b2.d.mts Text-1 "export const NOT_REACHABLE: number;"
```
andrewbranch, line 190: I guess I could have kept the same strict extension filtering while mixing in the allowed TS implementation file extensions, keeping this unreachable file out of the AutoImportProviderProject compilation. But it doesn’t seem realistic for packages to ship declaration files that are literally unreachable, so doing the simpler thing and allowing all supported extensions seems fine. Updated the test to show that this file doesn’t result in a new auto-import.

src/compiler/moduleNameResolver.ts
```diff
@@ -2292,8 +2291,8 @@ function loadEntrypointsFromExportMap(
                     /*excludes*/ undefined,
                     [
                         isDeclarationFileName(target)
-                            ? replaceFirstStar(target, "**/*")
-                            : changeAnyExtension(replaceFirstStar(target, "**/*"), getDeclarationEmitExtensionForPath(target)),
+                            ? changeAnyExtension(replaceFirstStar(target, "**/*"), ".*", supportedDeclarationExtensions, /*ignoreCase*/ true)
```
sheetalkamat, line 2294: It seems like we are only filtering out replacing to `*.*` only if extension is of type ".d.someext.ts"
Was that intentional. I  wonder if we can just always do `changeAnyExtension(replaceFirstStar(target, "**/*"), ".*")` irrespective of extension .

andrewbranch: The problem is `changeAnyExtension(path, ".*")` only replaces from the last `.`, so that would result in `foo.d.ts` becoming `foo.d.*`, when we need it to become `foo.*` instead.

sheetalkamat: but if i locally change this to do ` changeAnyExtension(replaceFirstStar(target, "**/*"), ".*"),` none of the tests fail ?

andrewbranch: You're right. There's currently no test where the `exports` target has a declaration file extension but we find a `.ts` on disk instead. While that should arguably work, it's not very realistic.

weswigham (approved):

src/compiler/moduleNameResolver.ts
weswigham: Presumably we don't have many tests with export maps pointing explicitly at declaration files with js or implementation files alongside them that we'd prefer to lookup - it's usually the other way around, so only finding matches for `name.d.*` gets us what we want almost all the time.

andrewbranch: Yeah, exactly. Still, I running with an include of `name.d.*` feels too much like a mistake to me. I made a new helper `changeFullExtension` that recognizes and swaps all declaration file extensions (including ones for arbitrary file extensions, which the current code doesn't handle correctly), and am just trying to figure out how to push the commit I made on my host machine from my dev box 😅

src/compiler/path.ts
```diff
@@ -756,6 +757,24 @@ export function changeAnyExtension(path: string, ext: string, extensions?: strin
...
+ *
+ * ```ts
+ * changeAnyExtension("file.d.ts", ".js") === "file.d.js"
+ * changeFullExtension("file.d.ts", ".js") === "file.js"
+ * ```
+ */
+export function changeFullExtension(path: string, newExtension: string) {
+    const declarationExtension = getDeclarationFileExtension(path);
+    if (declarationExtension) {
+        return path.slice(0, path.length - declarationExtension.length) + newExtension;
```
sheetalkamat, line 773: Does this need to handle presence or absence of `"."` in new extension like in `changeAnyExtension` : `(startsWith(ext, ".") ? ext : "." + ext)`

sheetalkamat (approved):

Issue #56856, opened 2023-12-23
Declarations can't be emitted when types from inner modules are exported through separate export declarations


Andarist:
### 🔎 Search Terms

declarations portable emit modules export declaration alias aliasing

### 🕗 Version & Regression Information

- This is the behavior in every version I tried

### ⏯ Playground Link

N/A

### 💻 Code

```ts
// @strict: true
// @declaration: true
// @module: nodenext
// @moduleResolution: nodenext
// @target: esnext

// @filename: node_modules/@tanstack/vue-query/build/modern/useQuery-CPqkvEsh.d.ts
type QueryKey = ReadonlyArray<unknown>;

interface Register {}

type DefaultError = Register extends {
  defaultError: infer TError;
}
  ? TError
  : Error;

type ShouldRetryFunction<TError = DefaultError> = (
  failureCount: number,
  error: TError,
) => boolean;
type RetryValue<TError> = boolean | number | ShouldRetryFunction<TError>;

type QueryFunctionContext<
  TQueryKey extends QueryKey = QueryKey,
  TPageParam = never,
> = [TPageParam] extends [never]
  ? {
      queryKey: TQueryKey;
    }
  : {
      queryKey: TQueryKey;
      pageParam: TPageParam;
    };

type QueryFunction<
  T = unknown,
  TQueryKey extends QueryKey = QueryKey,
  TPageParam = never,
> = (context: QueryFunctionContext<TQueryKey, TPageParam>) => T | Promise<T>;

interface QueryOptions<
  TQueryFnData = unknown,
  TError = DefaultError,
  TData = TQueryFnData,
  TQueryKey extends QueryKey = QueryKey,
  TPageParam = never,
> {
  retry?: RetryValue<TError>;
  queryFn?: QueryFunction<TQueryFnData, TQueryKey, TPageParam>;
  queryKey?: TQueryKey;
  initialData?: TData;
  initialDataUpdatedAt?: number | (() => number | undefined);
}

interface QueryObserverOptions<
  TQueryFnData = unknown,
  TError = DefaultError,
  TData = TQueryFnData,
  TQueryData = TQueryFnData,
  TQueryKey extends QueryKey = QueryKey,
  TPageParam = never,
> extends QueryOptions<
    TQueryFnData,
    TError,
    TQueryData,
    TQueryKey,
    TPageParam
  > {
  enabled?: boolean;
  refetchInterval?: number;
  select?: (data: TQueryData) => TData;
}

[...]

Issue #38111, opened 2020-04-22
Declaration emit reveals paths within dependency that were not referred to in the source file


mheiber:
Context:

- Generating declaration files

Declaration emit reveals paths within dependency that were not referred to in the source file:

```ts
export declare const instance: import("dependency/internal-types").TheInterface;
```

This is a problem because changing paths within a dependency can break a dependent package.

**TypeScript Version:**  3.9.0-dev.20200212

**Search Terms:**

- [inline import declaration](https://github.com/microsoft/TypeScript/issues?q=is%3Aissue+is%3Aopen+inline+import+declaration)
- [declaration external relative path](https://github.com/microsoft/TypeScript/issues?q=is%3Aissue+is%3Aopen+declaration+external+relative+path+)

**Code**

`node_modules/@types/dependency/entrypoint.d.ts`
```ts
import { TheInterface } from "./internal-types";
export { TheInterface };
export declare function getInstance(): TheInterface;

```

`node_modules/@types/dependency/internal-types.d.ts`
```ts
export interface TheInterface {}
```

`node_modules/@types/dependency/index.d.ts`
```
// empty - doesn't matter for this example
```

`example.ts`
```ts
import { getInstance } from "dependency/entrypoint";
export const instance = getInstance();
```

`src/app/tsconfig.json`
```json
{
    "compilerOptions": {
        "declaration": true,
    }
}
```

**Expected behavior:**

`example.d.ts`
```ts
export declare const instance: import("dependency/entrypoint").TheInterface;
```

**Actual behavior:**

`example.d.ts`
```ts
export declare const instance: import("dependency/internal-types").TheInterface;
```

It seems that part of the solution could involve having the compiler avoid using relative paths in types (`import("<relativepath>").<typename>`) if the `relativepath` is outside the project.

**Repro Repo:** https://github.com/mheiber/repro-rel-import-inlining

**Related Issues:** 

- Declaration emit should not inline type definitions - #37151
- Proposal: Bundling TS module type definitions - #4433

**Issue written with the help of**: @rricard, @robpalme and @mkubilayk

mheiber:
will clean this up and reopen

rricard:
When writing up that issue we also found out a pretty interesting effect.

If you were to move the `internal-types.d.ts` into a sub-directory in the dependency and have entrypoint load it:

`node_modules/@types/dependency/entrypoint.d.ts`
```ts
import { TheInterface } from "./int/internal-types";
export { TheInterface };
[...]

PR #56857, opened 2023-12-23
Emit declarations using alternative containing modules for types exported using separate export statements

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/56856

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -5626,7 +5626,17 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             return undefined;
         }
         const containers = mapDefined(candidates, candidate => getAliasForSymbolInContainer(candidate, symbol) ? candidate : undefined);
-        return containers.length === 1 ? getWithAlternativeContainers(containers[0]) : containers;
+
+        let bestContainers: Symbol[] = [];
+        let alternativeContainers: Symbol[] = [];
+
+        for (const container of containers) {
+            const [bestMatch, ...rest] = getWithAlternativeContainers(container);
+            bestContainers = append(bestContainers, bestMatch);
+            alternativeContainers = addRange(alternativeContainers, rest);
+        }
+
+        return concatenate(bestContainers, alternativeContainers);
 
         function getWithAlternativeContainers(container: Symbol) {
             const additionalContainers = mapDefined(container.declarations, fileSymbolIfFileSymbolExportEqualsContainer);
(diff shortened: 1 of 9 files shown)


src/compiler/checker.ts
```diff
@@ -5601,6 +5601,34 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+                }
+                checkExpressionCached(d.parent.left.expression);
+                return getNodeLinks(d.parent.left.expression).resolvedSymbol;
+            }
+        });
+        if (!length(candidates)) {
+            return undefined;
+        }
+        const containers = mapDefined(candidates, candidate => getAliasForSymbolInContainer(candidate, symbol) ? candidate : undefined);
+        return containers.length === 1 ? getWithAlternativeContainers(containers[0]) : containers;
```
Andarist, line 5629: this line is ad-hoc, perhaps I should just `flatMap` this. From what I've seen the consumers of this return value assume some ordering - or well, operate on first items with priority or smth. So I'm not sure how it would behave in more complex scenarios with smth like: `[...withAlternativeContainersA, ...withAlternativeContainersB, ...withAlternativeContainersC]`

fatcerberus: Wait I’m confused; why would there only being a single element in the array affect whether you need to repackage it or not?

Andarist: I tried to explain this in the comment above. This is the only case I know how to reason about and about which i’m somewhat confident here. This is *not* the final solution - i’m seeking guidance what should be done here when there are more items in this array

weswigham: > From what I've seen the consumers of this return value assume some ordering - or well, operate on first items with priority or smth

Indeed, the results returned here are intended to be in priority order for "what is best to serialize", which is, admittedly, a bit subjective. IMO, the ordering here would probably be the direct symbols for each available container symbol, followed by all their alternative symbols, in order - so not *quite* a flat map.

weswigham (changes requested):

src/compiler/checker.ts
Andarist: I implemented the suggested solution. Could you take another look?

weswigham (approved):

acutmore:
Just a note to say that this change seems to have regressed our (Bloomberg's) workaround for #38111.

I'll put a playground together and raise a concrete issue tomorrow 🙂

PR #56859, opened 2023-12-23
Keep returned (and yielded) literal types as const when their types using `const` type variables

### LINKED ISSUE #53813: Type-level narrowing constraints (like `as const`)
# Suggestion

In userland code, you can use `as const`, but as a library author, I often want to provide the best possible experience, and when there's an opportunity for a literal value to be produced, I would like to infer that value for my users without them having to use (or remember to use) `as const`.

## 🔍 Search Terms

constant literals, Math.random, require as const, infer literals, readonly by default

## ✅ Viability Checklist

## ⭐ Suggestion

Provide an ability to narrow types at the type level (i.e. without JavaScript where you can use `as const`).

## 📃 Motivating Example

```ts
const notLiteral = () => "A";
```

A type like this produces `() => string`.  I can improve the situation by specifying `as const`, which will cause it to produce `() => "A"`.

> Note: I'm not totally sure how related this is, but I've seen `> () => Math.random() > 0.5 ? "A" : "B"` which produces `() => "A" | "B"`, and it's not clear to me what the difference is from the example above

However, in the context of a library, a library author doesn't have the option to do this in the type level.  Something for this use-case would be a nice improvement:

```ts
////// LIBRARY CODE
const UploadThingServerHelper = <ValidRoutes,>(
  route: {
    readonly [Route in keyof ValidRoutes]: {
      middleware: () => ValidRoutes[Route]; // how can a library author force a more specific type here?
      onUpload: (response: { metadata: ValidRoutes[Route] }) => void;
    };
  }
) => {};

////// END USER CODE
const FileRouter = UploadThingServerHelper({
  example: {
    middleware: () => "someValue", // and end-user must remember to put `as const` here to force it to return a literal
    onUpload: response => {
      response.metadata; // the result is that the type here is not as narrow as it could be for the best user experience
      //       ^?
    },
  },
});
```

## 💻 Use Cases

This is useful for libraries trying to provide a great user experience with great inferencing.  It should be a backwards compatible change.  There are no workarounds that I'm aware of (from the code of a library).

### LINKED ISSUE #56858: Literals in return positions are widened despite being typed using `const` type variables
### 🔎 Search Terms

literal widening const type variable primitive

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.4.0-dev.20231223#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXwxAGcMAeMPE+AFQD4AKWAcwC556BKeAXlpo7bUAUEIqoqcIsggYAjDwLEM9Tjz4BtAOSIcOTQF0OQgPTH45+AD0A-KMoZ4k6RgBMCwiRVde8AAY6cABIAbwBZKAwACwA6GChUYBwAW04AX18jUwsrWzEJYmcAZnclLzV4bV1NTLMLGxEgA

### 💻 Code

```ts
declare function test<const T>(arg: () => T): T

const result1 = test(() => ["foo"])
//    ^? const result1: readonly ["foo"]
const result2 = test(() => `foo${Math.random()}`)
//    ^? const result2: `foo${number}`
const result3 = test(() => "foo")
//    ^? const result3: string
```

### 🙁 Actual behavior

`result3` is inferred as `string`

### 🙂 Expected behavior

`result3` should be inferred as `"foo"`

_No response_

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/56858
closes https://github.com/microsoft/TypeScript/issues/53813


src/compiler/checker.ts
```diff
@@ -36952,7 +36955,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         const isAsync = (getFunctionFlags(func) & FunctionFlags.Async) !== 0;
         forEachYieldExpression(func.body as Block, yieldExpression => {
             const yieldExpressionType = yieldExpression.expression ? checkExpression(yieldExpression.expression, checkMode) : undefinedWideningType;
-            pushIfUnique(yieldTypes, getYieldedTypeOfYieldExpression(yieldExpression, yieldExpressionType, anyType, isAsync));
+            pushIfUnique(yieldTypes, getYieldedTypeOfYieldExpression(yieldExpression, yieldExpression.expression && isConstContext(yieldExpression.expression) ? getRegularTypeOfLiteralType(yieldExpressionType) : yieldExpressionType, anyType, isAsync));
```
sandersn, line 36958: This line is too long and we haven't turned on a line length yet. I'd suggest extracting the conditional in a similar way as the first diff in the PR:
```suggestion
            let yieldExpressionType = yieldExpression.expression ? checkExpression(yieldExpression.expression, checkMode) : undefinedWideningType;
            if (yieldExpression.expression && isConstContext(yieldExpression.expression)) {
                yieldExpressionType = getRegularTypeOfLiteralType(yieldExpressionType);
            }
            pushIfUnique(yieldTypes, getYieldedTypeOfYieldExpression(yieldExpression, yieldExpression.expression && isConstContext(yieldExpression.expression) ? getRegularTypeOfLiteralType(yieldExpressionType) : yieldExpressionType, anyType, isAsync));
```

sandersn (changes requested):
The change seems sensible but we should probably review it at a design meeting.

sandersn:
My notes from the design meeting: 

This improves the apparent consistency so much that it's worth taking. The implementation needs to be right of course.

To check that: 
- run all test suites, 
- test multiply-nested cases
- test overloads
- test async functions, generators, async generators

Also decide whether this is the right place to implement it; could be that it's better to put it in getWidenedType. (@weswigham brought this concern up but seemed to be believe that its current place is better.)

sandersn:
@typescript-bot test top100
@typescript-bot run dt
@typescript-bot pack this

sandersn:
It looks like we decided in a design meeting that this is a good change, but it still needs more tests:

- test multiply-nested cases
- test overloads
- test async functions, generators, async generators

@Andarist do you want to finish up this PR by adding those tests?

Andarist:
@sandersn I added some new test cases here. Please take a look and see if those are satisfactory.

sandersn (approved):

jakebailey (approved):

Issue #36401, opened 2020-01-24
Date.getMonth could return a union of all possible values


nikeee:
## Search Terms
- `getMonth()`
- `Date`
- Zero-Based Month
- Also filed on https://github.com/microsoft/TSJS-lib-generator/issues/817 (I discovered later that this repo is only for the DOM types).

## Suggestion
[According to the MDN docs](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Date/getMonth), `Date.getMonth()` always returns a number between `0` and `11` (inclusive):
> An integer number, between 0 and 11, representing the month in the given date according to local time. 0 corresponds to January, 1 to February, and so on.

Currently, the `lib.d.ts` definition states that this function returns `number`.

So I suggest du change the definition of getMonth() from:
https://github.com/microsoft/TypeScript/blob/91ffa1c752ac26882b1426fb4c9012701c1d908e/lib/lib.es5.d.ts#L750-L751

...to something like:
```TypeScript
interface Date {
    // ...
    /** Gets the month, using local time. */
    getMonth(): 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11; // (or using a type definition for this union)
    // ...
}
```

## Use Cases
This would be beneficial if user code contains something like `if (date.getMonth() == 12)`. The compiler can then infer that the entire if branch is dead code.
Also, it serves as a quick documentation to let the user know that the return value is in fact a 0-based (not 1-based) month. The number of months in a year is also unlikely to change as well as the base (0/1) of the month index that this function returns.

## Checklist
My suggestion meets these guidelines:
* [x] This wouldn't be a breaking change in existing TypeScript/JavaScript code <-- **I'm not sure about that. It *should not* be breaking, since `0 | ... | 11` is a subset of `number`.**
* [x] This wouldn't change the runtime behavior of existing JavaScript code
* [x] This could be implemented without emitting different JS based on the types of the expressions
* [x] This isn't a runtime feature (e.g. library functionality, non-ECMAScript syntax with JavaScript output, etc.)
* [x] This feature would agree with the rest of [TypeScript's Design Goals](https://github.com/Microsoft/TypeScript/wiki/TypeScript-Design-Goals).

## Remarks
This could likewise be done for `getUTCMonth()`.

[...]

Issue #15135, opened 2017-04-11
NaN, Infinity and -Infinity not accepted in number literal types


olivr70:
**TypeScript Version:**  2.2.2

**Code**

```ts
// A *self-contained* demonstration of the problem follows...
let a : 0 | 1 / compiles
let b : 0 | NaN // fails : Cannot find name 'NaN'
let b : 0 | Inifinity // fails : Cannot find name 'Inifinity'
let b : 0 | -Inifinity // fails : Cannot find name '-Inifinity'
```

**Expected behavior:**
Well, NaN, +Infinity and -Infinity are valid numbers, so I see no reason they should not be accepted

**Actual behavior:**
It fails to compile

** Benefits **
I discovered this was all discussed in #9407. @ahejlsberg and @weswigham  explicetly took position against including it, except for a compelling case...
 
Other than being the intuitive result (why on earth a perfectly valid number would not be acceptable a in a number literal type, I see several benefits : 

One application would be the Math.sign() type definition. It is currently defined as returning a ```number``` although the spec says it can only return ```0 | -0 | +1 | 1 | NaN```
This is in fact how I got into the problem : I wanted to implement a comparator function (returning -1|0|1) using Math.sign( <some difference computation> ), intuitively expecting to be limited to those 3 values (and at the time forgetting about NaN)

To continue with the Math.sign, it could be defined as follows
```ts
interface Math {
  sign(x:NaN):NaN
  sign(x:number):0|1|-1
}
```
This way, it explicitely states when the function returns NaN and be extremely useful for documentation purposes (I suppose modifying a compiler for documentation is not very strong...)

Another one I can see (although it's a quick thought) : I think it could provide very interesting handling of NaN (if we had substraction types #4183)

RyanCavanaugh:
> it explicitely states when the function returns NaN and be extremely useful for documentation purpose

Basically comes down to this. The first signature there would *only* ever be hit when writing exactly `Math.sign(NaN)`, which... why would you ever do that (or anything equivalent where you already know that a value is `NaN`).

A secondary objection is that `NaN` is not semantically a unit type the same way 0 is -- any number of *different* invalid operations produce `NaN`; this is also somewhat true of `Infinity`. `NaN`'s odd behavior with `===` would also introduce a bunch of extra logic in how we narrow.

PR #56866, opened 2023-12-24
fix(es5): type narrowing of return type

### LINKED ISSUE #56865: type narrowing for getday
### 🔎 Search Terms

getday

All versions

### ⏯ Playground Link

_No response_

### 💻 Code

_No response_

### 🙁 Actual behavior

```ts
const now = new Date();
const now_day = now.getDay(); // type "number"
```

### 🙂 Expected behavior

```ts
const now = new Date();
const now_day = now.getDay(); // type "0 | 1 | 2 | 3 | 4 | 5 | 6"
```

Using type narrowing will be return correct type and improve usability.

imwh0im:
Fixes #56865

### DIFF
--- a/src/lib/es5.d.ts
+++ b/src/lib/es5.d.ts
@@ -780,7 +780,7 @@ interface Date {
     /** Gets the day-of-the-month, using Universal Coordinated Time (UTC). */
     getUTCDate(): number;
     /** Gets the day of the week, using local time. */
-    getDay(): number;
+    getDay(): 0 | 1 | 2 | 3 | 4 | 5 | 6;
     /** Gets the day of the week using Universal Coordinated Time (UTC). */
     getUTCDay(): number;
     /** Gets the hours in a date, using local time. */


MartinJohns:
A change like this was previously rejected: https://github.com/microsoft/TypeScript/issues/36401#issuecomment-578217960

src/lib/es5.d.ts
```diff
@@ -780,7 +780,7 @@ interface Date {
     /** Gets the day-of-the-month, using Universal Coordinated Time (UTC). */
     getUTCDate(): number;
     /** Gets the day of the week, using local time. */
-    getDay(): number;
+    getDay(): 0 | 1 | 2 | 3 | 4 | 5 | 6;
```
Josh-Cena, line 783: This is incorrect. It can also be `NaN` (like every other method).

MartinJohns: Related: #15135

Issue #56771, opened 2023-12-13
Incorrect inferred function return type with destructuring assignments


byanofsky:
### 🔎 Search Terms

"return type destructure"
"destructuring assignment type"

### 🕗 Version & Regression Information

- This changed between versions 3.3.3 and 3.5.1

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.3.2#code/C4TwDgpgBASlC8UDeAoK6oEEBcy0YIENcByTE-dAXygB8oA7AVwBsW6omGATCAMwCWDCNxRUUfLgGNgAgPYMofBgB4AKlAgAPYBB4BnWAD4AFMGxqAlLg14CAJwjAm9xUhqFDagNxiUKAHoAKhQASUUIQikACyhdfWAoKU9oACMIFjkAdwAaKFCSbihPfSYAW2hgaOhHZ1c48Gg5PiVpWQU4uSh0qAADJBxkYigyEioqXoA6FAAJbIgANwh7POT9aAAWTnXDXgT7JhkXIQBzYp446IFDQjBIQntDYC7ePkJWRKqapxdFUEhOn0YFMUEEAv4AgEoABhFJQACMgSh0Lk9kcMhYICgQj4yyejSgzT6jlKLGAk0wvWKhn6gyQw1G416KCkCgSUBJHwQSgYJgGuHppHI40svk5ZIpk0I3igkKgADkust7KiIci4QAmJEw1Ho4CY7EMXGPBoAomYeHUvqCkbC5mshjsi3c5R8ukM4VUSwU3wWqUyuWKzRo1Xa2HrKAAZjDuogGKxOLxpqaLUwkat-Q9Y3tbMS-KwGpoiFd+ZtjI8hgdCVFKEwGv9sqhQeVobl4c22vCrLRcf1CaNSf+KYLGdLWZo9GYbA4XFeQhEExZueQgzTRZ5boF469vsjDblAFEQ-ZcAB5VIAK172MMYDk+n0AlSBt6U5YVNRfVn-Hn3F6QA

### 💻 Code

```ts
type R = {
    A: {
        a: 'A'
    } | null | undefined
}
function fn<T extends R>(t:T): T  {
    return {} as T;
}

/*
In each test case below, I'd assume the return type of function to be `{A: {a: 'A'}}`.
However, case 4 uses destructuring and this appears to default the return type to `R`.
*/

// Case 1
[...]

Issue #42969, opened 2021-02-25
nested destructured tuple value is `string | undefined` when second tuple item is not accessed


G-Rath:
# Bug Report

When using `noUncheckedIndexedAccess`, the type inferred by my code is affected by whether I access the second item in a tuple or not.

### 🔎 Search Terms

tuple, undefined, type, omit, noUncheckedIndexedAccess, error, index, array, infer, nested

### 🕗 Version & Regression Information

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about `noUncheckedIndexedAccess`
- I was unable to test this on prior versions because `--noUncheckedIndexedAccess` didn't exist before those versions

### ⏯ Playground Link

[Playground link with relevant code](https://www.typescriptlang.org/play?noUncheckedIndexedAccess=true&ts=4.2.0-beta#code/CYUwxgNghgTiAEYD2A7AzgF3gMxQLngB4AVeEADwxBWDXkxgEsUBzAbQF0A+ACgwOIBKeAF4u8NsQA09DE1acOAbgCwAKHXJ0WNm3IcOonCh5sA5AEYzHQao1qtaJBBAA6CEhY9yrjEgAySADuIDAAwlBoIDyCtkA)

### 💻 Code

```ts
declare const fn: <T extends string[]>(t: T) => [T, string[]];

const [[x]] = fn(['1']);

console.log(x.toLowerCase());
```

It seems that the inferred type for `fn` is:

```
const fn: <string[]>(t: string[]) => [string[], string[]]
```

Since the first item is an array rather than a tuple, items in that array have type `string | undefined`.

If I change the assignment to be:

```
const [[x], []] = fn(['1']);
```

It gets inferred as:

```
const fn: <[string]>(t: [string]) => [[string], string[]]
```

Since the first item is a tuple rather than an array, items in that tuple have type `string`.

### 🙁 Actual behavior

`x` has type `string | undefined`

### 🙂 Expected behavior

`x` has type `string`.

G-Rath:
It seems like I can resolve this by changing the `extends` type to be `string[] | [string]` without any changes in or loss of functionality.

It's very possible that this is the Correct type I should be using for typing this the most accurately, but will leave this issue open for now as it still strikes me as a little odd that accessing the second item changes the inferred type of the first item, given it's redundant in JS (which is what got me here: eslint rules telling me it's not needed).

RyanCavanaugh:
[...]

PR #49086, opened 2022-05-12
Limit type argument inference from binding patterns

andrewbranch:
Second attempt of #45719 (and #46009)

Fixes #43605
Fixes #45074
Fixes #45663

In the past, when we made inferences to type arguments in return types, we did two things with those inferences: (1) put them in the primary inference context with a low priority, and (2) put them in a secondary inference context that gets used solely for instantiating contextual types of arguments in the same call expression. The consequence of (1) is easily observable when no other inference sources are present to overwrite the low-priority return type inference made from a binding pattern:

```ts
declare function f<T>(): T;
const { foo } = f(); // T: { foo: any }
```

#45719 fixed this by preventing return type inferences from binding patterns altogether, i.e., it stopped both (1) and (2) above. This turned out to be a problem, because (2) produces very desirable behavior with array binding patterns:

```ts
declare function f<T>(cb: () => T): T;
const [e1, e2, e3] = f(() => [1, "hi", true]); T: [number, string, boolean]
```

The binding pattern provides a tuple contextual type for the array literal in the return expression, allowing `e1`, `e2`, and `e3` to have separate types rather than all having `string | number | boolean`. To preserve this desirable behavior, I tried #46009 which only prevented inferences from _object_ binding patterns. However, it was later discovered that this still wasn’t a complete fix, because it still allowed:

```ts
declare function f<T>(): T;
const [e1, e2, e3] = f(); // T: [any, any, any]
```

I ended up reverting all these attempts to reevaluate, with the conclusion

> While it _may_ be true that object binding patterns never contribute anything useful to type argument inference, it is _not_ true that array binding patterns _always_ contribute usefully to type argument inference. What we really want is for binding patterns to contribute contextual typing information to other inference sources, while never being allowed to stand on their own as inference sources.

So that’s what _this_ PR does. Rather than make any distinctions between object and array binding patterns, I make a distinction between inference passes into separate contexts (1) and (2) above. If a contextual type for the return type came from a binding pattern, we make no inferences into (1), but we always make inferences into (2) so that other arguments can benefit from that contextual type (primarily for the array → tuple case—it does happen in other cases but I haven’t found any that are useful).

PR #46009, opened 2021-09-22
Revert binding pattern inference changes, but only for tuples

andrewbranch:
Partial revert of #45719—we now let an _array_ binding pattern turn inference of an array into a tuple, which is really nice. I still think the behavior we were seeing for _object_ binding patterns was undesirable, and #45846 supports that hypothesis since all the related errors were due to a tuple turning into an array.

PR #56875, opened 2023-12-25
Fixed inferred type widening with contextual types created by binding patterns

Andarist:
@atcb concluded [here](https://github.com/microsoft/TypeScript/pull/46009#issuecomment-925391490) that:
> While it may be true that object binding patterns never contribute anything useful to type argument inference, it is not true that array binding patterns always contribute usefully to type argument inference. What we really want is for binding patterns to contribute contextual typing information to other inference sources, while never being allowed to stand on their own as inference sources.

While working on this PR here, I realized that even object binding patterns can contribute useful things to inferences. That relies on initializers' types though, such as here (a test case from [this PR](https://github.com/microsoft/TypeScript/pull/49086/files#diff-8a2c3ec6f3bb9c38f0c1749d71492360d2785030f5220eb8c1efab13ce0eb552)):
```ts
declare function id<T>(x: T): T;
const { f = (x: string) => x.length } = id({ f: x => x.charAt });
```

One extra thing that I noticed based on the issue [here](https://github.com/microsoft/TypeScript/issues/56771) is that **constraints** are still useful. When inferring from a binding pattern we often end up with an inferred type like `{ a: nonInferrableAny }` but that **doesn't** pass the constraint check in `getInferredType` in the situations like this:
```ts
declare function id<T extends { a: 'foo' | 'bar', b: 'foo' | 'bar' }>(x: T): T;
const { a } = id({ a: 'foo', b: 'foo' });
```

The constraint here requires 2 properties but the type inferred from the binding pattern has only one. This particular case isn't problematic because when the inferred type doesn't satisfy the constraint the algorithm (silently~) picks the constraint.

Let's take a look at what happens in [the reported case](https://github.com/microsoft/TypeScript/issues/56771):
```ts
type R = { A: { a: 'A' } | null | undefined };
declare function fn<T extends R>(t: T): T;
const { A } = fn({ A: { a: 'A' } })
```

The inferred type here (`{ A: nonInferrableAny; }`) does satisfy the constraint and that leads to some lossy behaviors.
1. we end up using the inferred type as contextual type and the next inference pass widens the literal string in `getWidenedLiteralLikeTypeForContextualType` to `string`
2. then the inferred `{ A: string }` **fails** the constraint check in `getInferredType`
3. so we pick up the constraint and the argument satisfies that

The problem is that now the inferred type is the constraint (and that's nullable). We could have infer much better type of `{ A: { a: "A" } }`.

This indicates that the constraint is still useful but our inferred type from the binding patterns might even be a **supertype** of the constraint (a binding pattern might only refer to **some** of the keys etc).

So the solution to this problem is to *combine* both pieces of information - the inferred type and the constraint type. Without introducing any new concepts to the compiler we can do that by intersecting them. That creates a problem without further changes since `nonInferrablyAny` is just `any` and `T & any` is just `any`. Using this is lossy for contextual typing in a similar way to the one described above.

On the other hand, `unknown` behaves almost perfectly for this purpose. It's "erasable" in intersections: `T & unknown // unknown`. The only real problem I have found with it is that intersections between objects with concrete properties and objects with index signatures don't behave as I'd like them to here. 

A concrete property always shadows over the index signature - which is especially annoying~ here when it comes to tuple/array types:
```ts
type T1 = { a: nonInferrableUnknown } & { [k: string]: 'a' | 'b' }
type T2 = [nonInferrableUnknown] & ('a' | 'b')[]
```

In both of those situations we want to avoid those concrete properties to overshadow the more useful information contained in the index signature-based constraint. Since `nonInferrableUnknown` doesn't really hold any useful information for contextual typing it seems to be OK to ignore it and to look into the index signature in such cases. That has been implemented here.

fixes https://github.com/microsoft/TypeScript/issues/56771
fixes https://github.com/microsoft/TypeScript/issues/42969


src/compiler/checker.ts
```diff
@@ -11444,10 +11444,10 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             reportImplicitAny(element, anyType);
         }
         // When we're including the pattern in the type (an indication we're obtaining a contextual type), we
-        // use a non-inferrable any type. Inference will never directly infer this type, but it is possible
+        // use a non-inferrable unknown type. Inference will never directly infer this type, but it is possible
         // to infer a type that contains it, e.g. for a binding pattern like [foo] or { foo }. In such cases,
-        // widening of the binding pattern type substitutes a regular any for the non-inferrable any.
-        return includePatternInType ? nonInferrableAnyType : anyType;
+        // widening of the binding pattern type substitutes a regular any for the non-inferrable unknown.
+        return includePatternInType ? nonInferrableUnknownType : anyType;
```
Andarist, line 11450: I'm not sure when "widening of the binding pattern type substitutes a regular any for the non-inferrable any" is happening and if this part of the comment is even relevant today.

src/compiler/checker.ts
```diff
@@ -20905,7 +20905,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     function isSimpleTypeRelatedTo(source: Type, target: Type, relation: Map<string, RelationComparisonResult>, errorReporter?: ErrorReporter) {
         const s = source.flags;
         const t = target.flags;
-        if (t & TypeFlags.Any || s & TypeFlags.Never || source === wildcardType) return true;
+        if (t & TypeFlags.Any || s & TypeFlags.Never || source === wildcardType || source === nonInferrableUnknownType) return true;
```
Andarist, line 20908: This likely isn't strictly needed - at least not for the test cases I have here now. While iterating on this, I planned to make the created pattern type pass the constraint check in `getInferredType`. This line would allow for that in **some** scenarios. Later I realized that it wouldn't allow for that in all scenarios though - and that's why I just avoid doing the constraint check in `getInferredType`. The idea is still that the `nonInferrableUnknown` should act like `any` for assignability purposes though - I just replaced its type with `unknown` to benefit from the `T & unknown // T`.

src/compiler/checker.ts
```diff
@@ -25089,8 +25108,8 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     }
 
     function tupleTypesDefinitelyUnrelated(source: TupleTypeReference, target: TupleTypeReference) {
-        return !(target.target.combinedFlags & ElementFlags.Variadic) && target.target.minLength > source.target.minLength ||
-            !target.target.hasRestElement && (source.target.hasRestElement || target.target.fixedLength < source.target.fixedLength);
+        return !source.pattern && (!(target.target.combinedFlags & ElementFlags.Variadic) && target.target.minLength > source.target.minLength ||
```
Andarist, line 25111: This allows us to infer between tuples of different shapes in those scenarios - since the tuple created by an array binding pattern might be shorter than the target. So far, I haven't managed to break this but I would advise extra caution around this when reviewing.

src/compiler/checker.ts
```diff
@@ -30279,7 +30307,16 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             else if (t.flags & TypeFlags.StructuredType) {
                 const prop = getPropertyOfType(t, name);
                 if (prop) {
-                    return isCircularMappedProperty(prop) ? undefined : removeMissingType(getTypeOfSymbol(prop), !!(prop && prop.flags & SymbolFlags.Optional));
+                    if (isCircularMappedProperty(prop)) {
+                        return undefined;
+                    }
+                    const type = getTypeOfSymbol(prop);
+                    if (type !== nonInferrableUnknownType) {
```
Andarist, line 30314: Since `nonInferrableUnknownType` doesn't hold any useful information for contextual typing I let it fallthrough to index signatures here. Note that `{ x: nonInferrableUnknownType } & { x: string }` **shouldn't** return `nonInferrableUnknownType` for `'x'` property here. So if the constraint already defines a property then index signatures won't be consulted.

src/compiler/checker.ts
```diff
@@ -30279,7 +30307,16 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+                    if (isCircularMappedProperty(prop)) {
+                        return undefined;
+                    }
+                    const type = getTypeOfSymbol(prop);
+                    if (type !== nonInferrableUnknownType) {
+                        return removeMissingType(type, !!(prop && prop.flags & SymbolFlags.Optional));
+                    }
+                    if (t.flags & TypeFlags.Intersection) {
+                        t = getIntersectionType((t as IntersectionType).types.filter(t => !t.pattern));
+                    }
```
Andarist, line 30319: I wish I knew how to do it in a better way... this is mainly to satisfy the `isTupleType(t)` below (when it should be satisfied)

src/compiler/checker.ts
```diff
@@ -1968,10 +1968,10 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     var blockedStringType = createIntrinsicType(TypeFlags.Any, "any", /*objectFlags*/ undefined, "blocked string");
     var errorType = createIntrinsicType(TypeFlags.Any, "error");
     var unresolvedType = createIntrinsicType(TypeFlags.Any, "unresolved");
-    var nonInferrableAnyType = createIntrinsicType(TypeFlags.Any, "any", ObjectFlags.ContainsWideningType, "non-inferrable");
     var intrinsicMarkerType = createIntrinsicType(TypeFlags.Any, "intrinsic");
     var unknownType = createIntrinsicType(TypeFlags.Unknown, "unknown");
     var nonNullUnknownType = createIntrinsicType(TypeFlags.Unknown, "unknown", /*objectFlags*/ undefined, "non-null");
+    var nonInferrableUnknownType = createIntrinsicType(TypeFlags.Unknown, "unknown", ObjectFlags.ContainsWideningType, "non-inferrable");
```
Andarist, line 1974: Likely `getIntersectionType` should reduce `nonInferrableUnknown & unknown` to `unknown`. Most of such reductions are performed based on `includes` coming from `addTypesToIntersection`. It would be cool to introduce `TypeFlags.IncludesNonInferrableUnknown` but those `Includes*` flags are mostly reusing (?) other existing flags and I'm not super comfortable with touching this on my own.

jakebailey:
@typescript-bot test top200
@typescript-bot run dt

@typescript-bot pack this

Issue #56786, opened 2023-12-14
switch (typeof something) isn't seen as exhaustive when something is generic


benasher44:
### 🔎 Search Terms

- typeof
- switch
- narrow
- type guard

### 🕗 Version & Regression Information

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about Type Guards

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.3.2#code/GYVwdgxgLglg9mABBAFgUwgawIIBtcA8AKgHwAUAznALZpQoxgDmAXIkQJRtGIDeAUIkQUA7jCipEZKAE8ADmjjBhNOg2Yc+gocgCGFNIgBEYENQBGaAE5GW2oRH2Gj5mE0ZRb9vQePm4cLhoumBeOj7OFDIWgWE6jr5G4AAmaMCMaMlxDk7GoJCwCNkRxnDmAFYYnnbhQlZ0IFZIVLT0jEwA3NoAvvzdQA

### 💻 Code

```ts
function checkAll<T>(something: T): T {
  switch (typeof something) {
    case "number":
    case "bigint":
    case "boolean":
    case "symbol":
    case "undefined":
    case "function":
    case "object":
      return something;
  }
}
```

### 🙁 Actual behavior

This emits the TS error: "Function lacks ending return statement and return type does not include 'undefined'.(2366)"

### 🙂 Expected behavior

No error, since all possible results of typeof are handled.

### Additional information about the issue

It seems to only be an issue when the function argument is generic. It works if I switch the return type and parameter types to `any`, but I'm not sure that's a fair comparison.

fatcerberus:
> Function lacks ending return statement and **return type does not include 'undefined'.**

Implicit returns are allowed for `any`, so that's why it's not an error in that case.  The generic is a red herring; the switch is just not seen as exhaustive, period:

```ts
function checkAll(something: unknown): unknown {
  switch (typeof something) {
    case "number":
    case "bigint":
    case "boolean":
    case "symbol":
    case "undefined":
    case "function":
    case "object":
      return something;
  }
}
```

produces the error
```
Not all code paths return a value. (7030)
```

[...]

PR #56877, opened 2023-12-26
Treat `switch (typeof x)` as exhaustive when all cases are handled for unconstrained type parameter

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/56786

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -37014,7 +37014,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             if (!witnesses) {
                 return false;
             }
-            const operandConstraint = getBaseConstraintOrType(checkExpressionCached((node.expression as TypeOfExpression).expression));
+            const operandConstraint = getBaseConstraintOfType(checkExpressionCached((node.expression as TypeOfExpression).expression)) || unknownType;
             // Get the not-equal flags for all handled cases.
             const notEqualFacts = getNotEqualFactsFromTypeofSwitch(0, 0, witnesses);
             if (operandConstraint.flags & TypeFlags.AnyOrUnknown) {
--- a/tests/baselines/reference/exhaustiveSwitchTypeofUnconstrainedTypeParameter.symbols
+++ b/tests/baselines/reference/exhaustiveSwitchTypeofUnconstrainedTypeParameter.symbols
@@ -1,0 +1,28 @@
+//// [tests/cases/compiler/exhaustiveSwitchTypeofUnconstrainedTypeParameter.ts] ////
+
+=== exhaustiveSwitchTypeofUnconstrainedTypeParameter.ts ===
+// https://github.com/microsoft/TypeScript/issues/56786
+
+function checkAll<T>(something: T): T {
+>checkAll : Symbol(checkAll, Decl(exhaustiveSwitchTypeofUnconstrainedTypeParameter.ts, 0, 0))
+>T : Symbol(T, Decl(exhaustiveSwitchTypeofUnconstrainedTypeParameter.ts, 2, 18))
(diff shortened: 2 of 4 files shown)


src/compiler/checker.ts
```diff
@@ -37014,7 +37014,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             if (!witnesses) {
                 return false;
             }
-            const operandConstraint = getBaseConstraintOrType(checkExpressionCached((node.expression as TypeOfExpression).expression));
+            const operandConstraint = getBaseConstraintOfType(checkExpressionCached((node.expression as TypeOfExpression).expression)) || unknownType;
```
Andarist, line 37017: Note that the constraint was already obtained here before so this doesn't really change that much. It just unifies how `T` and `T extends unknown` are handled here (`switch (typeof x)` was already treated as exhaustive for the latter).

On top of that, this is literally what `getTypeFactsWorker` (that is used later here) does as the very first thing. So this change doesn't even affect the other branch of the code in a meaningful way.

jakebailey:
@typescript-bot test top200
@typescript-bot run dt

@typescript-bot pack this

src/compiler/checker.ts
fatcerberus: I had to stare at this diff for a good 10 seconds before I even realized it was calling a different function. 😵‍💫  They differ by only a single character (and `r` and `f` look quite similar at small font sizes) and it feels like it would be really easy to mix these up when reading the code.

gabritto, line 37017: This change as is breaks the following case:

```ts
// error on return type because switch is not seen as exhaustive
function woo(x: string): number {
    switch (typeof x) {
        case "string":
            return 1;
    }
}
```
The problem is that `getBaseConstraintOfType` returns `undefined` for primitive types like `string`, in addition to returning `undefined` for an unconstrained type parameter, which is why we use `getBaseConstraintOrType` currently.
I think you might need to explicitly check the unconstrained type parameter case and use `unknown` as the operand constraint then.

gabritto (changes requested):

src/compiler/checker.ts
Andarist: Good catch 👍 I pushed out a fix for this - could you take a look?

src/compiler/checker.ts
```diff
@@ -37014,7 +37014,10 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             if (!witnesses) {
                 return false;
             }
-            const operandConstraint = getBaseConstraintOrType(checkExpressionCached((node.expression as TypeOfExpression).expression));
+            const type = checkExpressionCached((node.expression as TypeOfExpression).expression);
+            const operandConstraint = type.flags & (TypeFlags.Intersection | TypeFlags.Instantiable) ?
```
gabritto, line 37018: Why do we want to do this when the type is intersection or instantiable?

Andarist: Doing this for instantiables shouldn't fall into the trap that you have mentioned (with primitive types). I'm not sure about intersections though, I just followed the same logic that I found at the top of `getTypeFactsWorker`.

gabritto: I see.
I was going to say I don't know exactly if this is correct, but I realized there might be a simpler way to get what we want:
we can always say that the `typeof` switch is exhaustive if it has all possible `typeof` strings, regardless of the operand constraint.
So I think we can leave this all unchanged and just check if `(TypeFacts.AllTypeofNE & notEqualFacts) === TypeFacts.AllTypeofNE` is true, and if it is, we can return true (and otherwise we can do what we were already doing).

gabritto (changes requested):
https://github.com/microsoft/TypeScript/pull/56877#discussion_r1439968635

PR #56889, opened 2023-12-27
added cycle detection for excessivly deep types

typeholes:
Fixes #

For perf test - need to measure impact vs savings in the when cycle detection lets excessively deep errors be triggered earlier


jakebailey:

Issue #56894, opened 2023-12-29
setter inferred from union has incorrect variance


craigphicks:
### 🔎 Search Terms

setter union variance

is:issue in-title: setter (nothing relevant found)

is:issue in-title: setter union variance (nothing relevant found)

### 🕗 Version & Regression Information

confirmed 4.3.5 ~ 5.3.2 

### ⏯ Playground Link

[playground](https://www.typescriptlang.org/play?#code/FAEwpgxgNghgTmABBA9gOwM4BdEEsQBcAPACoB8AFAJQC8ZJwwAZgK5oRa7qJMBMFADwKIA3kwIUKMAgEYAPr1pl5igL6I5o8ZOm85AZiV7D6qiOCJLydNkRxENRAIB0TCvmpVgAem9X-Af4AetaYOPjCRLyUVA5kiLwAND5+gYghqGF2wjIaCXn6wKqMrOyc3Ez6FOKIOrIKSiqxmnXGRgaxZhZWmbb2jm4eVF6+aWkZNuGEiFExcQmJliljE1lwOXm8BUWMzGwcXGg8ACyCwiIYYDggUvWKANyIAOZXdtR36poXrze6HY8vHBwd7GUzmUZjSwhCgAB3gMAAtlcwHBYkJRIhprlNLxHggYCB0FAAJ7ZRDYhKPT4Y6ZbTT6PFgAlE0nrfL0qndSwuEAOPA3Kj3ZbjUK2CIzCl0xD6OZ0cmbbb+FxwR7CqH+WFwFAwlFYYmxNmS7bFIA)

### 💻 Code

```ts
declare const id:<T>()=>T

function f2(x: {f:((a:1|2)=>1|2)} | {f:((a:2|3)=>2|3)} ){
    const r = x.f(id())
//                ^ const id: <2>() => 2,
//        ^ const r: 1 | 2 | 3
}

function f3(f: ((a:1|2)=>1|2) | ((a:2|3)=>2|3) ){
    const r = f(id())
//              ^ const id: <2>() => 2,  
//        ^ const r: 1 | 2 | 3
}


function f4(x: {set d(a:1|2); get r():1|2} | {set d(a:2|3); get r():2|3} ){
//           ^ (parameter) x: { d: 1 | 2; readonly r: 1 | 2; } | { d: 2 | 3; readonly r: 2 | 3; }
    x.d = id();
//        ^ const id: <1 | 2 | 3>() => 1 | 2 | 3
[...]

PR #47674, opened 2022-01-31
Compute writeType from set accessors for union and intersection properties

andrewbranch:
Fixes #45376

Issue #18770, opened 2017-09-26
Strict variance and read-only versions of types


mattmccutchen:
I feel like we should have an open "suggestion" issue with a meaningful summary to track the work on stricter variance and read-only versions of types that the TypeScript team is considering, especially as other issues are being duped against it.  The initial design sketch and a few comments are in #17502.

mhegazy:
Here is a rough sketch to what we have discussed over the past years regarding variance checking. if we were to implement it we need at least:
- A new flag that disables the bivariance comparisons for **all** functions and methods. https://github.com/Microsoft/TypeScript/pull/18654 does this for function/constructor types but not methods.
- A way to declare parameter variance for a function. We have existing suggestions in https://github.com/Microsoft/TypeScript/issues/1394 and https://github.com/Microsoft/TypeScript/issues/10717. Though looking at the issue so far, using a `readonly`/`readwrite`/`writeonly` annotations seem to be more fitting for a structural type system. A long with that we probably need a matching `readonly` type operator that allows you to get the readonly version of a type, and that is tracked by https://github.com/Microsoft/TypeScript/issues/10725.
- A way to declare a lower bound on generic type parameter (https://github.com/Microsoft/TypeScript/issues/14520)

More discussions can be found in https://github.com/Microsoft/TypeScript/issues/18513 and https://github.com/Microsoft/TypeScript/issues/18339

RReverser:
@mhegazy If I understand correctly, does the second suggestion cover function calls that may currently modify readonly types?

In particular, got bitten before by popular "immutable" `Object.assign({}, obj, { ... })` pattern when person forgot the empty object and did `Object.assign(obj, { ... })` and TS didn't complain despite `obj` being `Readonly`.

mhegazy:
I think we can do that regardless of variance. the issue here is we do not check `readonly` flags in assignment compatibility.

```ts
type I = { p: number };

declare var i: I;
declare var roi: Readonly<I>;

i = roi; // Not an error!

roi.p = 0; // Error, p is readonly
``` 

This was tracked by https://github.com/Microsoft/TypeScript/issues/11481. The reason why we have shied from this check is it forces users to mark their parameters as readonly consistently everywhere once one of their libraries adopts a stricter readonly annotations. This might have been overblown concern, and maybe we should revisit our decision of a new `--strictReadonlyChecks` flag.

RReverser:
[...]

Issue #14150, opened 2017-02-17
Unsafe type-incompatible assignments should not be allowed


jaredru:
**TypeScript Version:**  2.1.6

**Code**

```ts
interface Immutable {
  readonly val: string;
}

interface Literal {
  val: "str";
}

interface NotNullable {
  val: string;
}

interface MutableAndNullable {
  val: string | null;
}

const immutable: Immutable = { val: "val" };
const literal: Literal = { val: "str" };
const notNullable: NotNullable = { val: "notNull" };

// should not be allowed
const mut1: MutableAndNullable = immutable; 
const mut2: MutableAndNullable = literal;
const mut3: MutableAndNullable = notNullable;

mut1.val = null;
mut2.val = null;
mut3.val = null;

// immutable, literal, and notNullable are now all in bad states
```

**Expected behavior:**
The sample should not compile, as it's not safe.  Type casts/assertions should be required to override type incompatibility errors here.  (Or, of course, cloning the object itself `const mut: MutableAndNullable = { ...immutable };`).

For comparison, Flow does _not_ allow the sample code.  

**Actual behavior:**
The sample compiles and further accesses of `immutable.val`, `literal.val`, and `notNullable.val` will likely result in exceptions.

mhegazy:
Duplicate of https://github.com/Microsoft/TypeScript/issues/13347

jaredru:
@mhegazy This issue overlaps with #13347, certainly, but it's not a duplicate.  There are more issues here than simply `readonly`.  Union and optional/nullable types have similar problems:

```ts
type A = { val: string; };
type B = { val: string | number; };

const a: A = { val: "string" };
const b: B = a;

b.val = 3;

console.log(a.val.toUpperCase()); // throws
```

mhegazy:
A literal is in the domain of string, similarly a string is in the domain or string or null. These work as intended.

jaredru:
I'm sorry I'm not making myself clear, but that's not the point.  The inner value type is not the problem.

The reference type containing those value types is the problem.  Again, I'll point out that Flow does not allow these assignments, _because they are unsafe_.

In my last example, type `A` is "in the domain of" type `B`, but the converse is not.  However, by allowing the assignment of the variable `a` to `b` without copying it, the compiler is allowing us to act on `A` as if it were a `B`.  That is bad behavior, intended or not.

mhegazy:
[...]

PR #56895, opened 2023-12-29
Fixed union/intersection write types

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/56922

cc @andrewbranch


src/compiler/checker.ts
```diff
@@ -14674,7 +14674,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         else {
             result.links.type = isUnion ? getUnionType(propTypes) : getIntersectionType(propTypes);
             if (writeTypes) {
-                result.links.writeType = isUnion ? getUnionType(writeTypes) : getIntersectionType(writeTypes);
+                result.links.writeType = isUnion ? getIntersectionType(writeTypes) : getUnionType(writeTypes);
```
Andarist, line 14677: after [the original change](https://github.com/microsoft/TypeScript/pull/47674) has been merged it [has been mentioned in the comments](https://github.com/microsoft/TypeScript/pull/47674#discussion_r935892404) that likely this should work in the way I have adjusted it here but I don't see any open issue that would be the result of this discussion there

craigphicks: 1. Currently the getters and setter seem to be converted right away to plain readonly property (for getter) and plain property (for setter).  So will write types exist for setters at this point?
2. What about the case where `if (propTypes.length > 2){ ... }` before this `else` clause?  Isn't that branch  handled later in different code which also has to be updated?

Andarist: 1. if I understand you correctly - what is displayed depends on the location. In write locations, the write type is displayed and in read locations, you get the read type. To change this the TS team would have to receive a well-motivated feature request with many examples of how it would differ from the current behavior.
2. Yes - you are right. I'll take a look at this.
3. What do you consider a plain type here? Could you rephrase this question?

craigphicks:
Although this is labelled as "fixes #56894" it does not change the behavior of the test case listed in that issue.

[See here](https://github.com/microsoft/TypeScript/pull/56915#discussion_r1438938034)

craigphicks:
An unintuive [difference in read behavior between a getter with a setter, and a getter without a setter](https://github.com/microsoft/TypeScript/pull/56915#discussion_r1438938644).

It seems to me like not treating getter and setters as functions means you end up having to creating exceptions in order to get them to behave like properties - and in the end they are neither properties or functions, but a instead a third class.

You would feel it more directly if you had to create the user documentation for it.

Andarist:
I think that part of what you mention here is likely related to [this](https://github.com/microsoft/TypeScript/issues/56894#issuecomment-1871903003) and I'd treat this as another issue. If you prefer, I can remove the reference to your issue from this PR.

craigphicks:
> I think that part of what you mention here is likely related to [this](https://github.com/microsoft/TypeScript/issues/56894#issuecomment-1871903003) and I'd treat this as another issue. If you prefer, I can remove the reference to your issue from this PR.

Of course I prefer if this pull (appropriately modified) fixes #56894!!!  
I posted #56922 which supersedes #56894 to highlight the runtime error risk.

Andarist:
I changed the referenced issue here as I don't want to fix the variance measurement for setters as part of this PR.

andrewbranch:
@typescript-bot test top200

andrewbranch:
@typescript-bot pack this

gabritto (approved):

gabritto:
As pointed out already, we should either not close #56922 with this PR or, preferably, re-open #56894 to track the setter variance computation issue.

Andarist:
I think the currently linked issue can be closed with this. From my PoV, its main focus is on those unsound writes that are fixed here. I agree that the other issue could be reopened to track the variance computation issue.

andrewbranch:
I'm surprised I can't find discussion on this, but I think we discussed this when we added the divergent accessor type feature and decided to make the implementation consistent with the (obviously incorrect) way that normal properties work. This PR does not make this an error:

```ts
class One {
    x!: string
}
class Two {
    x!: number
}
declare let u1: One | Two;
u1.x = "";
```

I would like to see this fixed everywhere, but that's of course a much bigger and more disruptive change, and I'm not sure if it makes sense to fix one without the other. Even more confusing, the logic changed here doesn't even kick in unless the property has a getter with a _different_ type. This is still not an error:

```ts
declare class OneAccessor {
    set x(value: string);
}
declare class TwoAccessor {
    set x(value: number);
}
declare let u2: OneAccessor | TwoAccessor;
u2.x = "";
```

And it's _still_ not an error when adding getters of the same type. But once one of the getters&mdash;theoretically irrelevant to this assignment&mdash;diverges from its setter, suddenly the assignment is an error? No, unfortunately, I think this is wrong on purpose.

Andarist:
> I would like to see this fixed everywhere, but that's of course a much bigger and more disruptive change, and I'm not sure if it makes sense to fix one without the other.

To confirm - is this something that you want to get fixed? I could certainly try to do this, I'm just looking for a soft confirmation that fixing this is even on the table. As you have said, it could potentially be a big and disruptive change - and I know that sometimes the team errs on the side of backwards compatibility.

andrewbranch:
I'm not sure when it was last discussed, but my impression was it's not on the table. I think the problem is that it's too slippery of a slope down to the realization that all non-`readonly` properties should be strictly invariant, which would be an unacceptably large break without a flag, and possibly too onerous even with a flag.

```ts
let obj: { x: string } | { x: number };
obj.x = ""; // Error, right?
let obj2: { x: unknown } = obj; // ...Uh oh
obj2.x = ""; // Must be ok, so previous line must be error
```

https://github.com/microsoft/TypeScript/issues/14150
https://github.com/microsoft/TypeScript/issues/18770

Issue #56900, opened 2023-12-29
Using `@satisfies` tag in JSDoc has strange effects on arrow function parameters


jespertheend:
### 🔎 Search Terms

jsdoc satisfies

### 🕗 Version & Regression Information

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about `satisfies`

### ⏯ Playground Link

https://www.typescriptlang.org/play?strictNullChecks=true&filetype=js#code/PQKhAIAEGcEMBcCW0BmiCm1wG8AUBXfRAEwC5xp4AnRAOwHMBKcAXgD5wA3AexIF9wIYACgAxt1qVwAWwCeAMXy1RSCa3CESrDtj7DhoCJHiyADumLoUOAAqwqsaenjoq0ADwnz3a3MXLVWjYBeW5uQRFgYHAY2Lj4hMSkxIA9NINo5Ky4gAluTldwfMKAA1DuEvBYWmIigCNoVwLweAALBHBEeE6sEtp0AqoS4SA

### 💻 Code

```js
/** @satisfies {(uuid: string) => void} */
const myFunction = uuid => {}

/** @typedef {Parameters<typeof myFunction>} Foo */
//                                           ^^^
//                        Hover over `Foo` and observe that it is `never`
```

### 🙁 Actual behavior

Type of `Foo` is `never`.

### 🙂 Expected behavior

Type of `Foo` should be `[uuid: string]`

### Additional information about the issue

In the playground link above I have enabled `strictNullChecks`, but when you disable this, the parameter becomes optional. I.e. `[uuid?: string]`, which is still incorrect. ([playground link](https://www.typescriptlang.org/play?filetype=js&strictNullChecks=true#code/PQKhAIAEGcEMBcCW0BmiCm1wG8AUBXfRAEwC5xp4AnRAOwHMBKcAXgD5wA3AexIF9wIYACgAxt1qVwAWwCeAMXy1RSCa3CESrDtj7DhoCJHiyADumLoUOAAqwqsaenjoq0ADwnz3a3MXLVWjYBeW5uQRFgYHAY2Lj4hMSkxIA9NINo5NiACW5OV3A8goADUO5i8FhaYkKAI2hXfPB4AAsEZpb0cFN7R2cCtqxa9HEnQtNA2AAbADogA))

[...]

PR #56907, opened 2023-12-30
Treat contextually typed functions in JS files as typed

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/56900


jakebailey:
@typescript-bot test top200
@typescript-bot run dt

@typescript-bot pack this

src/compiler/checker.ts
```diff
@@ -15107,7 +15107,8 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 isInJSFile(declaration) &&
                 isValueSignatureDeclaration(declaration) &&
                 !hasJSDocParameterTags(declaration) &&
-                !getJSDocType(declaration);
+                !getJSDocType(declaration) &&
+                !getContextualSignatureForFunctionLikeDeclaration(declaration);
```
sandersn, line 15111: Does this change the semantics of `@type`, or is it specific to `@satisfies`? For example, in

```js
/** @type {(uuid: string) => string}
const f = uuid => uuid
```

I recall that `@type` intentionally is *not* treated as a contextual signature in that case, but I could be mistaken. I believe that instead the `@type` tag is looked up and treated the same as if it were written:

```js
const f = /** @type {(uuid: string) => }*/(uuid => uuid)
```

That is, treated like a cast. Probably `@satisfies` should be the same.

sandersn (changes requested):
I think looking for a contextual signature might be incorrect unless there's somethiing inherently contextual about `@satisfies` that `@type` doesn't have.

src/compiler/checker.ts
Andarist: It changes the semantics of the second example but not the first one. `@type` above the declaration is treated as the contextual signature (`getContextualSignature` calls `getSignatureOfTypeTag` and returns that if it's found).

That said, I'm not sure if this is a bad change. This flag has one purpose - it changes how `getMinArgumentCount` gets computed. If the types of parameters are assigned in just any way, I think it's surprising that the resulting signature might get treated as "untyped" and that it might allow calls without all (seemingly) required parameters. Just take a look at the situation reported in [this thread](https://github.com/microsoft/TypeScript/issues/58580) and at the discussion happening there. While this PR here doesn't solve the problem described there with `@type` applied to parameters, the root cause is similar - despite the call being typed it's treated as untyped. This creates weird discrepancies between TS and JS files.

Andarist: An extra data point:
```ts
const f1 = /** @type {(uuid: string) => {}}*/(uuid => uuid)
f1() // error

const f2 = /** @satisfies {(uuid: string) => {}}*/(uuid => uuid)
f2() // ok, wat?
```

The reason behind the difference here is that with the cast the inner signature is untyped but the final type of `f1` is sourced from the casted type itself. Since that's a type node~ it can't be untyped. 

To give more context on the above:
- `checkAssertionWorker` computes `exprType` with an untyped signature (but its parameters are typed!)
- but it returns `getTypeFromTypeNode(type)`

This can't work like that with `@satisfies` as the `@satifies` merely provides some contextual typing - it doesn't "override" the expression type.

sandersn: Thanks, that explanation helps. I was getting side-tracked by having the jsdoc on the declaration instead of directly on the arrow.

sandersn (approved):
This makes sense to me as a place that should have been contextually typed before, but wasn't.

sandersn:
@Andarist can you merge this from main and then I'll run typescript-bot test it before merging it.

Andarist:
@sandersn done

jakebailey:
@typescript-bot test it

Andarist:
The webpack break above is correct in a sense that it brings the JS behavior closer to the TS behavior:
1. [TS playground](https://www.typescriptlang.org/play/?ts=5.8.0-dev.20250215#code/C4TwDgpgBAygxgCwgWwIZQLxQEoTgewCcATAHgGdhCBLAOwHMAaKAV1oGtb8B3WgPgDcAKCGhIOCOXwAbAG4QAwqmnSARqjjtMUABTViALiiUaDAJSY+UWfn3ChdYBEIAzDdAByRNNICy+YhZpCAAZfFRiZwV8WicAD2BSAHkwYGoY8gAVcAgrAG8hKCh6CGAUtIydMyNy9NosnOEikrLUuvIdckQUVCN4JDRqqFqM7Mh7ZtLcKTkIKqMdHUKiqAJYiASjEzomZaLCCABHFklgLaodxj3V5TUNdiNpmXklFXVNK6KLDCsbfQsAGTXRZreJnYwXBjMA7HU7nUz0b5WAAKhHwyGo5AgFEh9D4ZmEAF8RHBpKhyOQoF5CD5-IFglACkUAPpwA6oJxhCJRGJg0iZPhVRnXNaUKDScKRQjRdabKneZR0oKhSU82WJAXaJkrSatCr1BZdAaoJHCnU6g7AFiEWhQWhBaRQclO2ggJrmwmfc0tJ6zIXa81QS3W206UEbYDQo4nSjMOC3d7sU0BwNQaguXTxt73Cwp1NB0oh1i0SIuOgQYju-OEqAQaRYs35lbBm12iDcKCo9GYuY6A4zeRRgBWeGAycJBOugeJ1arK0910J7uJxKAA) - it has 2 errors
2. [JS playground](https://www.typescriptlang.org/play/?ts=5.8.0-dev.20250215&filetype=js#code/PQKhCgAIUgBAXAngBwKYBNUDNIG8BKqAxgPYBO6APAM7xkCWAdgOYA0kArowNaMkDujAHwBfSAGUiAC1QBbAIZQQwcOFARocIvIA2OgEbyi3SIWokdAN1QBhXQaPclcZPLLzZeWgxZj66Z1gyVHgOMkZqPEsSfxElFTUwQPg5ZB15FMgAeWR4ehIIgBUUVGSSzBxcXChIWuYQnLyC6gAKAEoALmzc-KKSgG4auoae5pbqaTl5LskZBU7upr60QaHIevgzC2t2rpaWtdrSRhSAD3gu7yY2Q8hggEcOVFpLumvWW+09Q2MuratbPYftwPrVam1IABeISQaL+CEAMlu+2OZwukCuLHYDyeLwxbxYEOhkAACmQSLJ6NRUDQCcwhG1BrURGIAHLkBQ6ACyJHQHB0qAAMiR5JgyDYCmj4qpwER0tRIuyyJyeXyBXgauohjAEKl0plCtqVLUAPpEYIZIUisUSk6oc7tDVgyDqOBINB4JUq3n8q2i1DiyX2+CUQqiaDGsHHWiQHTWgO2tFQp3O9YjJbUPYTObyIkw6qpsHBULhF1gN0lPDyRiIMTKFqMfk6Rm3ESg1Mbf47CEFwt3EJhRiQFpmoPndgmnHPeDsL4OYx5lN9yD0HAtOfAnu3QvFwecRgVJgYJnLyBiVA6alL0+70uMVD8UnkynU-bBcwA7GoABWxHgi97U8zxbIC4mXEQT2dNs1ggmo4hEIA) - it currently has no errors

jakebailey (approved):

Issue #5689, opened 2015-11-16
Language Service: TypeParameter.constraint is lazily calculated, but does not have an accessor function


fsoikin:
Because of this, I have to go through this awkward motion every time:

``` typescript
function getConstraint( type: ts.TypeParameter ) {
  // This call will cause the typechecker to resolve properties, as well as a bunch of other information
  // about the type (such as generic constraints), but we don't actually need its result right now.
  type.getProperties(); 

  return type.constraint;
}
```

For some members (such as "properties" - above), there is an accessor function, which I can use to be sure the data is up to date. Another category of members don't need an accessor function, because they are calculated eagerly, and so are "always there".

But there is this third category of members, which fall through the gap. One example is `TypeParameter.constraint`. Another is `InterfaceTypeWithDeclaredMembers.declaredProperties`. There are probably more, but I haven't come across them yet.

mhegazy:
Would exposing [`getConstraintOfTypeParameter`](https://github.com/Microsoft/TypeScript/blob/master/src/compiler/checker.ts#L4083-L4094) on the API address the issue?

fsoikin:
It would solve half the issue: now I can use that function to get the value while making sure it's calculated. But that is assuming that _I know which function to use_. If I was just exploring the API, I would see the `constraint` property and wouldn't think twice about just using it. It will be only after hours of debugging and googling (and possibly finding this issue) that I would gain the necessary knowledge.

So to solve the other half, the property needs to be hidden. But I'm not sure if it will break anything somewhere.

mhegazy:
Agreed. so what we need to do is 1. expose the getConstraintOfTypeParameter on the TypeChecker interface, 2, add Type.getConstraint() method on the TypeObject class, and 3, mark type.constraint as internal.

3 is a breaking change, so not sure if/when we can do it. but other ones should be doable.

mhegazy:
PRs are definitely welcomed.

RyanCavanaugh:
This was fixed via #20137

PR #56915, opened 2023-12-31
Andarist fix union intersection setters

craigphicks:
This draft was created only to analyze a couple more tests files in the context of pull #56895 

Fixes #


tests/cases/compiler/-da/divergentAccessorsTypes-56894-a.ts
```diff
@@ -0,0 +1,27 @@
...
+//        ^ const r: 1 | 2 | 3
+}
+
+
+function f4(x: {set d(a:1|2); get r():1|2} | {set d(a:2|3); get r():2|3} ){
+//           ^ (parameter) x: { d: 1 | 2; readonly r: 1 | 2; } | { d: 2 | 3; readonly r: 2 | 3; }
+    x.d = id();
+//        ^ const id: <1 | 2 | 3>() => 1 | 2 | 3      (5.3.2)
+//        ^ const id: <2>() => 2      (expecting, if a setter was treated as a function)
+    x.r;
```
craigphicks, line 25: Was expecting pull #56895 to change this behavior, because therein it say it fixes #5689.  But no change occurred as far as I can see.

tests/cases/compiler/-da/divergentAccessorsTypes-56894-b.ts
```diff
@@ -0,0 +1,44 @@
...
+    u.a = 2;
+    u.sb = 2;
+    u.gsa = 2;
+    u.a = 3;
+    u.sb = 3;
+    u.gsa = 3;
+    if (u.gsa === 1) {
+        u.gsa;
+    }
+    if (u.gc === 1) {
```
craigphicks, line 30: u.gsa===1 produces a read range error but u.gc === 1 does not.  That's not an intuitively obvious rule.

tests/cases/compiler/-da/divergentAccessorsTypes-56894-a.ts
Andarist: Getters are treated like properties so a union of getters returns the union of their return types. I think this just works as expected. Even if we treat them as functions - it still all checks out, return types are unionified:
```ts
declare const fn: (() => 1 | 2) | (() => 2 | 3)
const result = fn()
//    ^? const result: 1 | 2 | 3
```

craigphicks: My comment and the subject of #56894 is setters.  The point of the result diff is that valid domain has not narrowed by your fix. For more clarity see #56922.
```
function f4(x: {set d(a:1|2); get r():1|2} | {set d(a:2|3); get r():2|3} ){
//           ^ (parameter) x: { d: 1 | 2; readonly r: 1 | 2; } | { d: 2 | 3; readonly r: 2 | 3; }
    x.d = id();
//        ^ const id: <1 | 2 | 3>() => 1 | 2 | 3      (5.3.2)
//        ^ const id: <2>() => 2      (expecting, if a setter was treated as a function)
```

jakebailey:
#56895 is closed, so I'm going to close this draft too.

Issue #56433, opened 2023-11-16
Error from parameter assignability incorrectly reported on satisfies keyword 


ssalbdivad:
### 🔎 Search Terms

satisfies, error, reporting, span, call, function

### 🕗 Version & Regression Information

- This changed between versions 5.0 and 5.1

### ⏯ Playground Link

https://tsplay.dev/mMl61N

### 💻 Code

```ts
const f = (s: { a: true }) => {}

// I would expect the error to span:
// a) "{}" (Ideal IMO)
// b) "{} satisfies unknown" (A little messier but not wrong- this was the 5.0 behavior)

// The error is instead reported only on "satisfies" despite {} satisfying unknown
f({} satisfies unknown)
```

### 🙁 Actual behavior

(see example)

### 🙂 Expected behavior

(see example)

Andarist:
I'm pretty certain that this is a fallout from one of my PRs: https://github.com/microsoft/TypeScript/pull/53797

I think that the solution would be to "unwrap" the `errorNode` in argument positions to the `satisfies`' expression. I'll put up a PR for this later, I just need to locate all of the places that have to be updated and add the relevant test cases.

PR #56918, opened 2024-01-01
Fixed error spans for `SatisfiesExpression` check nodes

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/56433

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -20543,7 +20543,8 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             const elem = node.elements[i];
             if (isOmittedExpression(elem)) continue;
             const nameType = getNumberLiteralType(i);
-            yield { errorNode: elem, innerExpression: elem, nameType };
+            const checkNode = getEffectiveCheckNode(elem);
+            yield { errorNode: checkNode, innerExpression: checkNode, nameType };
         }
     }
 
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -698,6 +698,7 @@ import {
     isRightSideOfQualifiedNameOrPropertyAccess,
     isRightSideOfQualifiedNameOrPropertyAccessOrJSDocMemberName,
     isSameEntityName,
+    isSatisfiesExpression,
     isSetAccessor,
     isSetAccessorDeclaration,
     isShorthandAmbientModuleSymbol,
@@ -33939,6 +33940,11 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         }
     }
 
+    function getEffectiveCheckArgumentNode(argument: Expression): Expression {
+        argument = skipParentheses(argument);
+        return isSatisfiesExpression(argument) ? skipParentheses(argument.expression) : argument;
+    }
+
     function getSignatureApplicabilityError(
         node: CallLikeExpression,
         args: readonly Expression[],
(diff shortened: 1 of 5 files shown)


src/compiler/checker.ts
```diff
@@ -33939,6 +33940,11 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         }
     }
 
+    function getEffectiveCheckArgumentNode(argument: Expression): Expression {
```
Andarist, line 33943: I was looking for other solutions that wouldn't explicitly target argument nodes. However, I concluded that this is quite a specific situation because it's the only situation (that I could find) when the `SatisfiesExpression` node is **also** the location on which other errors (argument checks) are meant to be reported. Usually the "final target" (like an assignment target) of `SatisfiesExpression` is different so there is no such overlap~ like here.

fatcerberus: > Usually the "final target" (like an assignment target) of SatisfiesExpression is different so there is no such overlap~ like here.

Counterpoint:
```ts
((): null => ({}) satisfies unknown)();
```

fatcerberus: Food for thought: If the arrow function above is rewritten with braces, then the error span is correctly placed on the `return`.  For braceless arrow functions the "final target" could be the `=>` token.

Andarist: > Counterpoint:

Great example :P I'll fix this in a moment

> For braceless arrow functions the "final target" could be the => token.

I prefer to highlight smaller spans over huge ones but this one feels just too short. OTOH, an elementwise elaboration can highlight even a single character if a property is that short. I'm not sure if it would be completely apparent to everybody that highlighted `=>` is about the return value. It's not that it doesn't make sense when explained - it's just that in isolation it doesn't feel completely intuitive/unambiguous

tests/baselines/reference/typeSatisfaction_errorLocations1.errors.txt
```diff
@@ -0,0 +1,146 @@
...
+      const o = { a: true as const };
+      f(o satisfies unknown);
+        ~
+!!! error TS2345: Argument of type '[{ a: true; }]' is not assignable to parameter of type 'T'.
+!!! error TS2345:   '[{ a: true; }]' is assignable to the constraint of type 'T', but 'T' could be instantiated with a different subtype of constraint '{ a: true; }[]'.
+    }
+    
+    const tuple1: [boolean, boolean] = [true, 100 satisfies unknown];
+                                                  ~~~~~~~~~
+!!! error TS2322: Type 'number' is not assignable to type 'boolean'.
```
andrewbranch, line 87: This behavior is unchanged by this PR, but looks wrong. Did you look at this example?

Andarist: heh, I was starring at this and somehow missed the fact that it's wrong in the very same way 😅 

I just pushed out a fix for this. It is a little bit cumbersome to call `getEffectiveCheckNode` around the code to get the best behavior... I think it semantically makes a lot of sense. It is the expression of the `SatisfiesExpression` that we want to highlight. But since I have to call it at multiple places it's rather easy to miss some places where it should be called too.

So I wonder now if it wouldn't be better to flip the situation slightly. I could move this logic into `getErrorSpanForNode` and "target" the `satisfies` token in `checkSatisfiesExpressionWorker` (since that's the only place where we want to target the `satisfies` token). The problem is... that I can't easily pass a token down as `errorNode`. It's not even directly accessible in the checker. So maybe a text span could be passed down? I'm not sure if this is even feasible though since that argument is passed down through some functions and they might even try to check stuff on it. So overall it's likely not a 5min change.

Do you think the current solution with `getEffectiveCheckNode` is OK? Or should I try this other alternative?

andrewbranch: I think this is fine.

andrewbranch (approved):
Thanks!

Issue #46793, opened 2021-11-12
[transform API]: Allow passing `getEmitResolver` into `TransformationContext` and expose `getEmitResolver` from `Program`


ahnpnl:
# Suggestion

## 🔍 Search Terms

- getEmitResolver
- transform API

## ✅ Viability Checklist

My suggestion meets these guidelines:

* [x] This wouldn't be a breaking change in existing TypeScript/JavaScript code
* [x] This wouldn't change the runtime behavior of existing JavaScript code
* [x] This could be implemented without emitting different JS based on the types of the expressions
* [x] This isn't a runtime feature (e.g. library functionality, non-ECMAScript syntax with JavaScript output, new syntax sugar for JS, etc.)
* [x] This feature would agree with the rest of [TypeScript's Design Goals](https://github.com/Microsoft/TypeScript/wiki/TypeScript-Design-Goals).

## ⭐ Suggestion

`transform` API is quite nice and convenient because using this API doesn't need to invoke `Program` to compile `ts` to `js` to use custom AST transformers. However, it doesn't allow the case which requires to access to `EmitResolver`. It would be nice if `transform` API can allow users to pass `getEmitResolver` into the `TransformationContext`.

Currently, `getEmitResolver` isn't exposed for both `TransformationContext` as well as `Program` interfaces. 

## 📃 Motivating Example

An ideal example would be like this
```
export function constructorParametersDownlevelTransform(program: ts.Program): ts.TransformerFactory<ts.SourceFile> {
  const typeChecker = program.getTypeChecker();
  const reflectionHost = new TypeScriptReflectionHost(typeChecker);

  return (ctx: ts.TransformationContext) => {
    return getDownlevelDecoratorsTransform(
      typeChecker,
      reflectionHost,
      [],
      false,
      false,
      true,
    )({
      ...ctx,
      getEmitResolver: ctx.getEmitResolver()
        ? ctx.getEmitResolver
        : program.getDiagnosticsProducingTypeChecker().getEmitResolver,
    });
  };
}
```

## 💻 Use Cases

I am trying to experiment the idea:

- Use `transform` API to process AST of TypeScript source codes with some custom AST transformers.

- Use `swc` or `esbuild` to compile `ts` to `js` with the newly transformed source.

[...]

PR #56932, opened 2024-01-02
Exposed `TypeChecker.resolveName` API to public along with `SymbolFlags.All`

timocov:
Related to #46793 (doesn't actually fixes it, but partially addresses - see https://github.com/microsoft/TypeScript/issues/46793#issuecomment-1874537198)

### DIFF
--- a/src/compiler/types.ts
+++ b/src/compiler/types.ts
@@ -5238,7 +5238,7 @@ export interface TypeChecker {
      * @internal
      */
     getAllPossiblePropertiesOfTypes(type: readonly Type[]): Symbol[];
-    /** @internal */ resolveName(name: string, location: Node | undefined, meaning: SymbolFlags, excludeGlobals: boolean): Symbol | undefined;
+    resolveName(name: string, location: Node | undefined, meaning: SymbolFlags, excludeGlobals: boolean): Symbol | undefined;
     /** @internal */ getJsxNamespace(location?: Node): string;
     /** @internal */ getJsxFragmentFactory(location: Node): string | undefined;
 
@@ -5744,7 +5744,6 @@ export const enum SymbolFlags {
     Transient               = 1 << 25,  // Transient symbol (created during type check)
     Assignment              = 1 << 26,  // Assignment treated as declaration (eg `this.prop = 1`)
     ModuleExports           = 1 << 27,  // Symbol for CommonJS `module` of `module.exports`
-    /** @internal */
     All = FunctionScopedVariable | BlockScopedVariable | Property | EnumMember | Function | Class | Interface | ConstEnum | RegularEnum | ValueModule | NamespaceModule | TypeLiteral
         | ObjectLiteral | Method | Constructor | GetAccessor | SetAccessor | Signature | TypeParameter | TypeAlias | ExportValue | Alias | Prototype | ExportStar | Optional | Transient,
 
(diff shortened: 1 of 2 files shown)


src/compiler/types.ts
```diff
@@ -5744,7 +5744,6 @@ export const enum SymbolFlags {
     Transient               = 1 << 25,  // Transient symbol (created during type check)
     Assignment              = 1 << 26,  // Assignment treated as declaration (eg `this.prop = 1`)
     ModuleExports           = 1 << 27,  // Symbol for CommonJS `module` of `module.exports`
-    /** @internal */
     All = FunctionScopedVariable | BlockScopedVariable | Property | EnumMember | Function | Class | Interface | ConstEnum | RegularEnum | ValueModule | NamespaceModule | TypeLiteral
```
andrewbranch, line 5747: I wonder why `All` isn't just `-1` 🤔

andrewbranch:
@timocov is there a reason this is a draft? Were you planning to add anything else?

timocov:
Not really, I wasn't sure if this is the change you'd want to see in the compiler (asked in the linked issue but haven't got any response yet). If that's alright I can mark it as ready for review.

timocov:
@andrewbranch @rbuckton is there anything I can do to get this merged?

src/compiler/types.ts
jakebailey: We noticed that because this isn't `-1`, we're missing `Assignment` and `ModuleExports`. But, adding those does not appear to matter for our tests and are probably an oversight.

I just tried with `-1` and that also worked.

andrewbranch: IMO let’s change it to `-1` and then it won’t be out of date if we add any more flags.

jakebailey:
FWIW we did decide to expose this in today's design meeting, so it's just the `-1` thing and then this should be good to get in before the beta cutoff.

src/compiler/types.ts
```diff
@@ -5744,7 +5744,6 @@ export const enum SymbolFlags {
     Transient               = 1 << 25,  // Transient symbol (created during type check)
     Assignment              = 1 << 26,  // Assignment treated as declaration (eg `this.prop = 1`)
     ModuleExports           = 1 << 27,  // Symbol for CommonJS `module` of `module.exports`
-    /** @internal */
     All = FunctionScopedVariable | BlockScopedVariable | Property | EnumMember | Function | Class | Interface | ConstEnum | RegularEnum | ValueModule | NamespaceModule | TypeLiteral
         | ObjectLiteral | Method | Constructor | GetAccessor | SetAccessor | Signature | TypeParameter | TypeAlias | ExportValue | Alias | Prototype | ExportStar | Optional | Transient,
```
jakebailey, line 5748:
```suggestion
    All = -1,
```

tests/baselines/reference/api/typescript.d.ts
```diff
@@ -7000,6 +7001,7 @@ declare namespace ts {
         Transient = 33554432,
         Assignment = 67108864,
         ModuleExports = 134217728,
+        All = 67108863,
```
jakebailey, line 7004:
```suggestion
        All = -1,
```

jakebailey:
I just merged the suggestion to get this along; want to get this in before we cut for the beta.

jakebailey (approved):

tests/baselines/reference/api/typescript.d.ts
rbuckton: We should test this, to be certain it doesn't affect something else due to the additional flags that are now included.

jakebailey:
@typescript-bot test top200
@typescript-bot run dt

andrewbranch (approved):

jakebailey:
None of the above is new; merging. Thanks for the PR, @timocov!

Issue #56931, opened 2024-01-02
Low priority inference trumps a much better inference that can be made from a generic signature


Andarist:
### 🔎 Search Terms

inference generic signature type parameters low priority 

### 🕗 Version & Regression Information

- This is the behavior in every version I tried

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.4.0-dev.20240102#code/CYUwxgNghgTiAEAzArgOzAFwJYHtX1ES1RAGEcBbABzxFQwB4AFGHKgZ3hAA8M7hOAJXA4YwBuwwxiAcwA08KKgCeAPlUAKAFDx47EBmRUAXPA1VWHUyzbsAlPAC8q+ADccWYHJ3w22POwA-KYA3j66FrbBZgDWIMo4iPA2HHYA2gC6ANw+AL7edqbmluzWJQ7O8GgxqDgA7qg5WmABGPBw7ACMTgQgRCTk1LT02roMACpcvPycktKoMpoA+pFW8CHwFOwyppO5FS5hurpwhjD4GgfruTm6uVp2TR2dWgD0r-AAeoFaza3tIHYACYeoRiGRKDQSCMfBMpnxUAI9FJZMtVqV1pttrt4PsnIdwgCzhcriEbnlvLojsd0aY0gAiLYyekZSm4gpPQFAt4fb5aIA

### 💻 Code

```ts
declare function defineComponent<Props extends Record<string, any>>(
  setup: (props: Props) => void,
  options?: {
    props?: (keyof Props)[];
  },
): (props: Props) => unknown;

const res1 = defineComponent(
  <T extends string>(_props: { msg: T }) => {
    return () => {};
  }
);

res1
// ^? const res1: <T extends string>(props: { msg: T; }) => unknown

const res2 = defineComponent(
  <T extends string>(_props: { msg: T }) => {
    return () => {};
  },
  {
    props: ["msg"],
  },
);

res2
// ^? const res2: (props: { msg: any; }) => unknown
```

### 🙁 Actual behavior

A low-priority inference made from `keyof Props` prevents inference from the generic signature

### 🙂 Expected behavior

I'd expect both of those to infer the same result: `<T extends string>(props: { msg: T; }) => unknown`

### Additional information about the issue

There is an existential question to be asked here... what is a low-priority inference in the first place? :P Where is the line? Should every inference priority other than `InferencePriority.None` be treated as low-priority **here**?

PR #56939, opened 2024-01-03
Prefer higher-priority inferences from generic signatures over low-priority inferences from the first pass

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/56931

At the moment, I'm treating this as an experiment and my primary goal is to get some potential feedback from the test suite run (cc @jakebailey). It's not meant to be reviewed yet - unless somebody feels like it 😉 At the very least, I still have to adjust code comments.


src/compiler/checker.ts
```diff
@@ -39034,12 +39034,19 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+    function tryInferencesMerge(target: InferenceInfo[], source: InferenceInfo[]) {
+        let merged = false;
+
+        const areOverlapping = hasOverlappingInferences(target, source);
+
         for (let i = 0; i < target.length; i++) {
-            if (!hasInferenceCandidates(target[i]) && hasInferenceCandidates(source[i])) {
+            if (areOverlapping ? hasInferenceCandidates(source[i]) && hasInferenceCandidates(target[i]) && source[i].priority! < (target[i].priority! & ~InferencePriority.ReturnType) : hasInferenceCandidates(source[i])) {
                 target[i] = source[i];
+                target[i].priority = InferencePriority.None;
```
Andarist, line 39045: TODO for myself: 

this line might not be needed at all or the assigned priority should be the source priority without `InferencePriority.ReturnType` (the latter sounds like a good idea if those inferences should stay "upgradable"/mergeable). IIUC, no new inferences without `InferencePriority.ReturnType` can be made here anyway - all of those were already collected before the ones from generic signatures start to be included.

RyanCavanaugh:
@typescript-bot test this
@typescript-bot test top100
@typescript-bot run dt
@typescript-bot pack this

src/compiler/checker.ts
weswigham: Hm, yeah, `None` definitely doesn't feel correct, since `None` is actually the highest priority inference possible - and I don't think simply adopting `source`'s inference for `target` when they overlap should guarantee this is the best inference possible across all potential inference sites.

But also - this is mutating the inference object, which you should avoid, since it means if that inference is reused in multiple contexts, you'll be mangling the priority in all of them (...we probably shouldn't be reusing an inference in multiple contexts, but there are probably some edge cases we decided it was OK in that this might break the invariants of). If we can't reuse the `source` inference wholesale (because its priority already doesn't contain `ReturnType`), then we should duplicate the `source` inference object with a new priority.

weswigham (changes requested):

src/compiler/checker.ts
Andarist: I initially picked `None` here because it's the most conservative thing to do. If I'm reasoning about it correctly. Right now (on `main`) we iterate over those generic signatures (conceptually at least):
- if the inferences are overlapping then we don't try to use any of the new ones
- if the inferences are not overlapping then we use the new ones **but** that means that when we overlap again we won't pick any new ones

So my conservative change adjusts this only slightly - it only allows a single "upgrade" when inferences are overlapping (and in such a case, we only update the overlapping one, this is important - we can't assign inferences that are not overlapping).

I thought that this, both, is just the smallest change that could be done here if this direction is promising at all and that there might be an argument to be made here that in this scenario the inferences made from earlier generic signatures should be preferred (left ones having the higher priority over the right ones). The latter doesn't match how "regular" inferences are gathered but it is how it implicitly works on `main` (when overlap happens the new inferences are ignored after all, so they can't be replaced/updated).

I could certainly experiment with relaxing this behavior and allow those upgrades continuously.

---

I pushed out a change using `cloneInferenceInfo` but I'm not sure if you meant that or `cloneInferenceContext`. And perhaps it should also be applied conditionally? but then that would imply that `cloneInferenceInfo` is the right choice here since `ReturnType` is relevant to the info and not to the whole context. 

> If we can't reuse the source inference wholesale (because its priority already doesn't contain ReturnType)

Does it mean that when the inference info of the source contains `ReturnType` then it's OK to reuse it and avoid cloning?

fatcerberus:
> `None` is the highest priority inference

🎶 Everything you know is wrong
Black is white, up is down and short is long
And everything you thought was just so important doesn't matter
Everything you know is wrong
Just forget the words and sing along
All you need to understand is
Everything you know is wrong 🎶

src/compiler/checker.ts
weswigham: > Since those inferredTypeParameters are simply concatenated there is no good way of even rejecting the unused ones~ at some point.

Yeah, at some point a filtering of that list to remove type parameters that definitely aren't used and logic for figuring out if the type parameter was actually used is probably needed for usability. It'll be a bit of a challenge, though, with all the deferred inferences we can make "maybe it's used later 🤷‍♂️" is probably a somewhat common case.

PR #56946, opened 2024-01-04
With `allowJs`, process JS files found searching node_modules when they have a realpath outside node_modules

andrewbranch:
Basically, treat JS files in other monorepo packages as local files, not untyped dependencies.

Re https://github.com/microsoft/TypeScript/issues/56412#issuecomment-1875702738


src/compiler/program.ts
```diff
@@ -4039,7 +4040,7 @@ export function createProgram(rootNamesOrOptions: readonly string[] | CreateProg
 
                 const isFromNodeModulesSearch = resolution.isExternalLibraryImport;
                 const isJsFile = !resolutionExtensionIsTSOrJson(resolution.extension);
-                const isJsFileFromNodeModules = isFromNodeModulesSearch && isJsFile;
+                const isJsFileFromNodeModules = isFromNodeModulesSearch && isJsFile && (!resolution.originalPath || pathContainsNodeModules(resolution.resolvedFileName));
```
sheetalkamat, line 4043: Do we want to process this file as if it was not found in `node_modules` if yes i dont think it is enough. We increase `currentNodeModulesDepth` so we will add this file but may still elide its impoirts?

andrewbranch: I pushed some test changes that I think show it’s working.

sheetalkamat: But with `allowJS`, `maxNodeModuleJsDepth` is 2 so you would need to adjust that in the test case right ?

andrewbranch: No, it only defaults to 2 with a jsconfig.json as opposed to a tsconfig.json.

sheetalkamat (approved):

Issue #55688, opened 2023-09-08
Feedback on Standard Decorators emit size


AndrewJakubowicz:
### Acknowledgement

- [X] I acknowledge that issues using this template may be closed without further explanation at the maintainer's discretion.

### Comment

This GitHub issue contains feedback on the Standard Decorators emit size from the [Lit](https://lit.dev) team.
### Summary
It is difficult to recommend usage of Standard Decorators with the current JavaScript emit size.
### Background

Lit supplies a set of decorators that reduce the amount of boilerplate code you need to write when defining a component.

For example, the [material web button component implementation](https://github.com/material-components/material-web/blob/v1.0.0-pre.16/button/internal/button.ts), uses Lit’s `@property` decorator to annotate a field that triggers a reactive update.

### Measurement

The [Material Web](https://github.com/material-components/material-web) repository was used to test the emit size when using Standard Decorators. Note, this was done quite naively to get ballpark figures.

The following table has emitted JavaScript target as “es2022”, and no emitted JavaScript minification:

| Name              |  Size               | Gzipped        | Brotli             |
| ------------- | ------------- | ------------- | ------------- |
| Total size (experimental decorators)    | 650.66 kB | 172.29 kB | 143.65 kB |
| Total size (standard decorators)           | 871.46 kB (+34%) | 209.85 kB (+21.8) | 175.98 kB (+22.5) |

Results when minifying the emitted JavaScript with [Terser](https://terser.org/):

| Name              |  Size               | Gzipped        | Brotli             |
| ------------- | ------------- | ------------- | ------------- |
| Total (experimental decorators + terser) | 444.92 kB | 120.2 kB |  102.12 kB |
| Total (standard decorators + terser) | 520.09 kB (+16.9%) | 145.61 kB (+21.1%) | 123.78 kB (+21.2%) |

### Emit size of the Material Button implementation (without styles)

The material-web repository contains a lot of styles which are also emitted as JavaScript and incorporated in the numbers above. As a smaller benchmark, here is a single file which contains a single component.

Source code: https://github.com/material-components/material-web/blob/v1.0.0-pre.16/button/internal/button.ts

| Name              |  Size               | Gzipped        | 
| ------------- | ------------- | ------------- |
| button.js (experimental decorators)  | 5.13 kB | 1.66 kB |
| button.js (standard decorators)           | 10.8kB (+110%) | 2.53 kB (+52%) |

[...]

PR #56955, opened 2024-01-04
Stage 3 Decorators normative changes

### LINKED ISSUE #56606: Normative changes to decorator "extra initializer" evaluation order
### Acknowledgement

### Comment

Per https://github.com/pzuraq/ecma262/pull/12/files, which achieved consensus in the November 2023 TC39 meeting, the "extra initializers" for fields/auto-accessors (added by `context.addInitializer`) now need to be run independently for each field, _immediately after_ the field itself is initialized and defined on the class instance.

This results in the need to change our emit for Stage 3 decorators as follows:

Given:
```ts
class C {
  @dec a;
  @dec b;
  @dec f() { }
}
```

We currently emit (some parts elided):
```js
let C = (() => {
  ...
  return class C {
    static {
      ...
      __esDecorate(null, null, _f_decorators, { ... }, null, _instanceExtraInitializers);
      __esDecorate(null, null, _a_decorators, { ... }, _a_initializers, _instanceExtraInitializers);
      __esDecorate(null, null, _b_decorators, { ... }, _b_initializers, _instanceExtraInitializers);
      ...
    }
    a = (__runInitializers(this, _instanceExtraInitializers), __runInitializers(this, _a_initializers, void 0));
    b = __runInitializers(this, _b_initializers, void 0);
    f() { }
  };
})();
```

Here, we collect all "instance extra initializers" in a single array and evaluate them immediately after the `this` value is bound in the constructor (i.e. immediately upon entry to the constructor in a base class, or immediately after `super()` is called in a derived class). When fields are present, we evaluate the extra initializers as part of evaluating the initializer for the first field of the class.

After this change, we need to emit something like this instead:
```js
let C = (() => {
  ...
  let _a_extraInitializers = [];
  let _b_extraInitializers = [];
  return class C {
    static {
      __esDecorate(null, null, _f_decorators, { ... }, null, _instanceExtraInitializers);
      __esDecorate(null, null, _a_decorators, { ... }, _a_initializers, _a_extraInitializers);
      __esDecorate(null, null, _b_decorators, { ... }, _b_initializers, _b_extraInitializers);
    }
    constructor() {
      __runInitializers(this, _b_extraInitializers);
    }
    a = (__runInitializers(this, _instanceExtraInitializers), __runInitializers(this, _a_initializers, void 0));
    b = (__runInitializers(this, _a_extraInitializers), __runInitializers(this, _b_initializers, void 0));
    f() { }
  };
})();
```

Here, we still run the extra initializers for methods/getters/setters first, but we run the extra initializers for `a` after it is defined on the instance, but before we initialize `b`, and run the initializers for `b` at the top of the constructor, which is evaluated after `b` has been defined on the instance.

rbuckton:
This changes our Stage 3 Decorators down-level emit to align with recent normative changes to the proposal. 

Unfortunately, this will further complicate #55688, which pertains to the emit size for decorators, as this will increase overall emit size since we have to track "extra" initializers on a per-element basis.

Related pzuraq/ecma262#12
Related tc39/proposal-decorators#513
Fixes #56606


fatcerberus:
*More* normative changes in Stage 3?  They really can’t nail this one down, can they?

rbuckton:
> _More_ normative changes in Stage 3? They really can’t nail this one down, can they?

This happens with any large proposal. Runtimes generally don't start actual implementations until Stage 3, and once they do there is often feedback that requires normative changes based on implementation experience. The point of Stage 3 is to catch these issues when the rubber hits the road. So far, the changes to Decorators in Stage 3 have been relatively minor compared to other proposals of this magnitude and consist mostly of timing-related changes so that the feature is consistent and meets user expectations. Decorators is actually fairly stable in comparison, and the fact we have only had such minor changes is a testament to the time, energy, and attention to detail of everyone involved in the proposal.

fatcerberus:
I guess it seems worse because IIRC decorators were stage 3 originally, then rolled back to stage 2 for a bit, and now are stage 3 again, so the normative changes since then stand out more.

jakebailey (approved):
Based on my naive understanding, LGTM 😄 

(Needs a `dprint fmt`).

PR #56961, opened 2024-01-05
Reduce buffer to string conversions.

### LINKED ISSUE #56960: Reduce Buffer to string round trip conversions in tests
### Acknowledgement

### Comment

Typescript uses am implementation for a virtual file system that holds the data as a `Buffer`. Typescript exclusively (as far as I can tell) uses `utf8` when reading and writing files. Converting from string to buffer and back again is taking between 5-10% of test run time and can easily be avoided if the `vfs` implementation keeps the file content in the provided format (utf8 string) and only converts it as required.

dragomirtitian:
Fixes #56960 

This PR implements the suggested improvement of not converting between `Buffer` and `string` unless absolutely necessary in the virtual file system implementation used for tests.

Some anecdotal tests:

| Environment                              |       main |        PR |    Win% |
|-----------------------------------|-----------|---------|---------|
| Linux,  6 workers node 18              |  01:59     |  01:49  |   8.40% |
| Windows 10, 11 workers node 20  |  03:05     | 02:55  |    5.41% |

This also helps with isolated declarations clawing back some of the time taken by ID tests:

| Environment                              |       main |        PR |    Win% |
|-----------------------------------|-----------|---------|---------|
| Linux,  6 workers node 18              |  02:12     |  01:58  |   10.61% |
| Windows 10, 11 workers node 20  |  03:20     | 03:10  |    5.00% |

### DIFF
--- a/src/harness/vfsUtil.ts
+++ b/src/harness/vfsUtil.ts
@@ -650,14 +650,14 @@ export class FileSystem {
      * NOTE: do not rename this method as it is intended to align with the same named export of the "fs" module.
      */
     public readFileSync(path: string, encoding?: BufferEncoding | null): string | Buffer;
-    public readFileSync(path: string, encoding: BufferEncoding | null = null) { // eslint-disable-line no-null/no-null
+    public readFileSync(path: string, encoding: BufferEncoding | null | undefined = undefined) { // eslint-disable-line no-null/no-null
         const { node } = this._walk(this._resolve(path));
         if (!node) throw createIOError("ENOENT");
         if (isDirectory(node)) throw createIOError("EISDIR");
         if (!isFile(node)) throw createIOError("EBADF");
 
-        const buffer = this._getBuffer(node).slice();
-        return encoding ? buffer.toString(encoding) : buffer;
+        const fileBuffer = this._getBuffer(node, encoding ?? undefined);
+        return !fileBuffer.encoding ? fileBuffer.data.slice() : fileBuffer.data;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/harness/vfsUtil.ts
```diff
@@ -650,14 +650,14 @@ export class FileSystem {
      * NOTE: do not rename this method as it is intended to align with the same named export of the "fs" module.
      */
     public readFileSync(path: string, encoding?: BufferEncoding | null): string | Buffer;
-    public readFileSync(path: string, encoding: BufferEncoding | null = null) { // eslint-disable-line no-null/no-null
+    public readFileSync(path: string, encoding: BufferEncoding | null | undefined = undefined) { // eslint-disable-line no-null/no-null
```
jakebailey, line 653: It doesn't look like this change is still needed to make this compile.

jakebailey (approved):
This seems reasonable to me, though I am not seeing quite the same speedup on my machine.

I'm interested in @sheetalkamat's thoughts.

src/harness/vfsUtil.ts
dragomirtitian: The problem is this overload `public readFileSync(path: string, encoding?: null): Buffer;` I could change that `null` to `undefined`, but I hesitate to change the public API of vfs, although this overload does not appear to be used.

jakebailey: I don't quite understand what you mean; I reverted this line to the way it was before and it compiled.

dragomirtitian:
> This seems reasonable to me, though I am not seeing quite the same speedup on my machine.
> 

I am curios if you are seeing any speedups on your machine?

jakebailey:
Barely, if anything; I can try and get more data but I ran it a couple times and saw roughly my normal 1m45s before and after.

jakebailey:
```console
$ hyperfine --prepare 'git switch main' -n 'main' 'npm test -- --no-lint' --prepare 'git switch pr-56961' -n 'PR 56961' 'npm test -- --no-lint'
Benchmark 1: main
  Time (mean ± σ):     114.415 s ±  1.469 s    [User: 2134.875 s, System: 51.155 s]
  Range (min … max):   110.731 s … 116.508 s    10 runs
 
Benchmark 2: PR 56961
  Time (mean ± σ):     109.475 s ±  1.874 s    [User: 2032.384 s, System: 37.635 s]
  Range (min … max):   108.442 s … 114.625 s    10 runs
 
Summary
  'PR 56961' ran
    1.05 ± 0.02 times faster than 'main'
```

jakebailey:
Perhaps this has a better effect when fewer workers are in use; my machine ends up running with 19 workers.

src/harness/vfsUtil.ts
dragomirtitian: Yes. You are right. My bad. It was initially different, but it does work with the original signature now. Fixed

Issue #56691, opened 2023-12-06
Empty types and types with `never` as a key behave differently


kazatsuyu:
### 🔎 Search Terms

Mapped type
unknown

### 🕗 Version & Regression Information

- This changed between versions 4.7.4 and 4.8.4

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.4.0-dev.20231206#code/CYUwxgNghgTiAEYD2A7AzgF3gVwFw5QGsUkB3FAbgCgqMBPABwQEF4BeeAbwG0BpeAJYp4KEADcQMALr4oKOgF8qAemXx1APQD8NVfABySeAFEYMJDCrJ0WKPhTYIEeAB8CoAGZCQwV1wXsONR66vDaumqm5pbWmPAARvaOzm7YKJ7evm6sHNjBaqHhQA

### 💻 Code

```ts
declare const u: unknown;

type A = {[K in never]: any}
//   ^?

// No Error
const a: null | undefined | {} = u;
//    ^?

// Error
const b: null | undefined | A = u;
//    ^?
```

### 🙁 Actual behavior

The types `a` and `b` appear to be the same, but only the assignment to `b` causes an error.

### 🙂 Expected behavior

I'm not sure which is better, I think it's either no error or both errors.

### Additional information about the issue

Both were errors in version 4.7.4

MartinJohns:
What is that mapped type supposed to do?

jcalz:
I suggest simplifying the example to something like `type A = {[K in never]: any}` which reproduces the behavior without the added distraction of conditional types.

Relevant: #49119 and https://github.com/microsoft/TypeScript/pull/49119#discussion_r880504101

kazatsuyu:
Well, it certainly looks like the problem is not that I mapped `{}`, but that I created a type with `never` key.

RyanCavanaugh:
I suppose the steelmanned version of the issue looks like this
```ts
export  {}
declare const u: unknown;
type Point = { x: number, y: number };

declare function foo<T, K extends keyof T>(obj: T, keys: K[], rest: Omit<T, K> | null | undefined): void;
const p: Point = { x: 0, y: 0 };
foo(p, ["x", "y"], u);
```

RyanCavanaugh:
I'm not *committing* to take a fix for this since it's pretty niche, but if it's straightforward and doesn't break/slow down something else, we can take a PR

PR #56974, opened 2024-01-07
Extend `isEmptyAnonymousObjectType` with support for non-generic mapped types

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/56691

This makes non-generic mapped types with `never` constraints behave like `{}` (since it displays as `{}`). "If it looks like a duck, swims like a duck, and quacks like a duck, then it probably is a duck."


src/compiler/checker.ts
```diff
@@ -20849,7 +20849,11 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     }
 
     function isEmptyAnonymousObjectType(type: Type) {
```
Andarist, line 20851: Note that this is different from `isEmptyObjectType` and none of those reuse the other. Perhaps with this change, some sharing could be introduced. I tried that at first and a lot of tests failed - but I didn't bother digging into this right now. 

I don't really know what's the intended use case of both. Some code comments here would be highly appreciated. I could add the. However, I'd need first to learn from the team about the different goals of those two.

tests/cases/fourslash/codeFixClassImplementInterfaceAutoImports.ts
```diff
@@ -28,11 +28,10 @@ verify.codeFix({
   description: "Implement interface 'Base'",
   newFileContent:
 `import { Base } from './interface';
-import A from './types1';
 import { B, C, D } from './types2';
 
 export class C implements Base {
-    a: Readonly<A> & { kind: 'a'; };
```
Andarist, line 35: This is a somewhat unfortunate change. However, it matches the behavior of plain `A` that is `{}`. In such a case, `A` doesn't appear in the output after the refactoring. Maybe that's an issue worth raising? The intention of the user could have been to reuse types from the interface quite literally - so when they start implementing `A`'s members the `C` class automatically "receives" them.

tests/cases/fourslash/completionPreferredSuggestions1.ts
```diff
@@ -15,6 +15,6 @@
 
 verify.completions({ marker: "1", includes: ["a", "b", "c"] });
 verify.completions({ marker: "2", includes: ["0", "1", "2"], isNewIdentifierLocation: true });
-verify.completions({ marker: "3", includes: ["a", "b", "c"] });
+verify.completions({ marker: "3", exact: [] });
 verify.completions({ marker: "4", exact: [] });
-verify.completions({ marker: "5", includes: ["a", "b", "c"] });
+verify.completions({ marker: "5", exact: [] });
```
Andarist, line 20: Since `Record<never, never>` is just `{}` with this PR those are expected *here*. Those unions were reduced to just `string` and thus concrete literal strings can no longer be suggested.

Both of those changed locations obey the rule from https://github.com/microsoft/TypeScript/pull/49119 :
> For backwards compatibility, special exceptions to the T & {} type reduction rules existing for intersections written explicitly as string & {}, number & {}, and bigint & {} (as opposed to created through instantiation of a generic type T & {}).

gabritto:
@typescript-bot run DT
@typescript-bot test top200

tests/cases/fourslash/completionPreferredSuggestions1.ts
gabritto: How is it that this `string & Record<never, never>` was being preserved before? Is it because, before this PR, `Record<never, never>` was some mapped type and we couldn't do subtype reduction on the intersection?

I'm curious to see if this breaks people, since this is in our tests, but you're right, officially we only support `string & {}` explicitly written as so.

tests/cases/fourslash/codeFixClassImplementInterfaceAutoImports.ts
gabritto: I agree that it might be unfortunate, but I think this change is ok. If we want to have a mechanism for preserving those kinds of intersections in refactorings and codefixes, we need it to also work for `A & { kind: ... }`.

gabritto:
Seems like there are many breaks on the top repos test :/

tests/cases/fourslash/completionPreferredSuggestions1.ts
Andarist: > Is it because, before this PR, Record<never, never> was some mapped type and we couldn't do subtype reduction on the intersection?

IIRC - yeah, something like that. The type was not reduced to `{}` so it stayed as a mapped type and that wasn't reduced. But with this change it's treated like `{}` so intersections containing it are way more easily reducible.

gabritto (changes requested):
https://github.com/microsoft/TypeScript/pull/56974#issuecomment-1918064441

PR #56989, opened 2024-01-09
Skip parent error when reporting JSX excess property checks

Andarist:
This is just a small extension of https://github.com/microsoft/TypeScript/pull/55152 since that PR didn't touch the JSX-related codepath. cc @RyanCavanaugh

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -21711,10 +21711,10 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                                 const suggestionSymbol = getSuggestedSymbolForNonexistentJSXAttribute(propName, errorTarget);
                                 const suggestion = suggestionSymbol ? symbolToString(suggestionSymbol) : undefined;
                                 if (suggestion) {
-                                    reportError(Diagnostics.Property_0_does_not_exist_on_type_1_Did_you_mean_2, propName, typeToString(errorTarget), suggestion);
+                                    reportParentSkippedError(Diagnostics.Property_0_does_not_exist_on_type_1_Did_you_mean_2, propName, typeToString(errorTarget), suggestion);
                                 }
                                 else {
-                                    reportError(Diagnostics.Property_0_does_not_exist_on_type_1, propName, typeToString(errorTarget));
+                                    reportParentSkippedError(Diagnostics.Property_0_does_not_exist_on_type_1, propName, typeToString(errorTarget));
                                 }
                             }
                             else {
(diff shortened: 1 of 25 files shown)


src/compiler/checker.ts
```diff
@@ -21711,10 +21711,10 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                                 const suggestionSymbol = getSuggestedSymbolForNonexistentJSXAttribute(propName, errorTarget);
                                 const suggestion = suggestionSymbol ? symbolToString(suggestionSymbol) : undefined;
                                 if (suggestion) {
-                                    reportError(Diagnostics.Property_0_does_not_exist_on_type_1_Did_you_mean_2, propName, typeToString(errorTarget), suggestion);
+                                    reportParentSkippedError(Diagnostics.Property_0_does_not_exist_on_type_1_Did_you_mean_2, propName, typeToString(errorTarget), suggestion);
```
Andarist, line 21714: One thing that could also be done here is that this could use a more specific message. The one for object literals is better since it mentions object literals. Perhaps this could mention JSX attributes? Or "literal" JSX attributes?

jakebailey (approved):
Yeah, if you expand out the code right below the diff, it's exactly the same. I think this makes sense.

jakebailey:
On second thought, I think there's some value in the extra object being printed because there's nothing about the code in some of these examples where the `children` prop even comes from.

Maybe this does actually need a better special-case error.

Andarist:
To understand better what you are saying... you want to special case `children` here because it often comes from JSX children and not from an explicit prop named `children`? And now we don't see the "normalized props" that were passed to JSX

jakebailey:
I'm not sure, really; IIRC we have to manufacture a fake `children` property since that's not explicitly written out (but is just what you stick inside your element body), so I feel like it must make some sense to also be careful about how we report such an error. But, I guess this is why it's assigned to Daniel and Ryan 😄

DanielRosenwasser:
It's been a while since I was involved in this code (I remember talking to @weswigham about it a while back).

If the mismatch is regarding children, it would be helpful to report a more specific error about children. If it's clearly not, we can probably avoid the elaboration. If it's unclear (e.g. the target is a union type and not all of the constituents have `children` or something) maybe we provide the current error.

But I don't remember where/how we manufacture a property called `children` and how much the special case actually needs to be generalized.

iisaduan (commented):
The generating children is done in `createJsxAttributesTypeFromAttributesProperty` in `checker.ts` https://github.dev/microsoft/TypeScript/blob/2149a0f31a5bf1b1b9e9c2745efeaa74e95170b9/src/compiler/checker.ts#L32951. 

Based on the comments there, it also seems reasonable (and in-line with the special case jsx error messages we've added latelyish) to me to special case `children` when children isn't defined in props. But I wonder it would get complicated for error messages where  `children` only shows up in the elaboration

PR #57008, opened 2024-01-10
Detect binary files by checking first 256 code units for 0xFFFD

### LINKED ISSUE #21136: Code Helper attempts to index MPEG-TS files in project and eats 100% cpu
*TS Template added by @mjbvz*

**TypeScript Version**: 4.3.5

**Search Terms**
- video
- mpeg
- mp4
- binary

---

_From @jon-valliere on January 10, 2018 22:54_

- VSCode Version: 1.19.2
- OS Version: 10.13.2

Steps to Reproduce:

1.  Add a MPEG-TS file (e.g. somefile.ts) outside of your normal code folder but somewhere in the project so it shows up in "Explorer" in VSC
2. Look at your processes -- intellicode totally stops working and the process eats 100%

The TypeScript processor should be smart enough to be able to detect if the file is text or not.

_Copied from original issue: Microsoft/vscode#41439_

### LINKED ISSUE #56516: TSC runs out of memory when TransportStream videos are present in the path
### 🔎 Search Terms

typescript transportstream

- This is bug that causes OOM
- This behavior exists on previous versions, checked on 4.2.4

### ⏯ Playground Link

_No response_

### 💻 Code

```ts
// Your code here
```

### 🙁 Actual behavior

I had a 300MB TransportStream (`video.ts`) video in my project dir, this caused tsc to try and process it, which caused an out-of-memory exception.

### 🙂 Expected behavior

tsc should ignore non-text files with `.ts` extensions, or blacklist videos or TransportStream specifically.

_No response_

jakebailey:
Fixes #56516
Fixes #21136


jakebailey:
@typescript-bot test top200
@typescript-bot run dt

@typescript-bot pack this

sheetalkamat (approved):

sandersn (approved):
Now do cryptocurrency .ts files.

fatcerberus:
> One thing I'm wondering is if we should just search the first 80 characters or something and say it's binary if any of them are replacementCharacter. Theoretically any garbage could be in a ts file and break things, and it's not clear to me that replacementCharacter would ever legally appear in any reasonable file.

That's what I suggested here: https://github.com/microsoft/TypeScript/issues/56516#issuecomment-1828650642

I'm pretty sure the "first 80 characters" method is the way it's typically done (e.g. in text editors, to determine if they should open a hex view).  The Windows API even has a function for it IIRC.

jakebailey:
That or I could just check for `/\ufffd/`. A replacement character won't happen in good files, right?

fatcerberus:
Yeah, check the first 80 chars and if any is a replacement char, treat as binary

src/compiler/scanner.ts
```diff
@@ -1799,7 +1799,21 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
                 // position 0 because UTF-8 decode will fail and produce U+FFFD.
                 // If that happens, just issue one error and refuse to try to scan further;
                 // this is likely a binary file that cannot be parsed
-                if (ch === CharacterCodes.replacementCharacter) {
+                let isBinary = ch === CharacterCodes.replacementCharacter;
+                // See if this is an MPEG Transport Stream, where every 188th byte is "G" and the rest is garbage.
```
DanielRosenwasser, line 1803: Is an MPEG transport stream guaranteed to be a multiple of 188 bytes long?

jakebailey: Yes, unless it's malformed. But, I actually think I should dump this PR and instead just check the first 80 or 256 characters for replacementCharacter instead. Curious what people think.

DanielRosenwasser: It's probably fine, but my concern is really more the case that at any arbitrary position, when `scan` is called, you will look ahead 188 characters.

If you had the byte position and the raw content length, you could avoid doing any of this, but because we're operating on something parsed into UTF-16, I don't think you have that option.

jakebailey: This only happens at `pos == 0`.

DanielRosenwasser: Missed that, thanks!

jakebailey:
@typescript-bot test top200

DanielRosenwasser:
> > Now do cryptocurrency .ts files.
> 
> NFT types when

Ryan closed #48513 😭

jakebailey:

sheetalkamat (approved):

jakebailey:
Bot is a liar

RyanCavanaugh (approved):

jakebailey:
@typescript-bot pack this

jakebailey:
Oops, wrong comment

@typescript-bot perf test

wooorm:
@jakebailey this seems to affect JavaScript files that include a replacement character. Example: https://github.com/micromark/micromark/blob/8b08d16891f4d6801eb0c0e990bd7b87836c0826/test/io/misc/zero.js#L9.

Is this intentional? Or a bug? Should I create an issue?

Cross-reference https://github.com/micromark/micromark/pull/166

jakebailey:
Ugh, not super thrilled that people are quite literally putting the replacement character into their files instead of escaping it; that's going to be indistinguishable from actually broken files.

I'm not sure if it's "intentional" because I really did not think people would be doing that. So far the linked issues appear to be cases where it's possible to just use `\uFFFD` (which is what I expected everyone do be doing).

I guess it could deserve a dedicated issue rather than discussing on the PR. Not doing this means we have to come up with some other, new way to handle this :(

wooorm:
> it's possible to just use `\uFFFD` (which is what I expected everyone do be doing).

Well, you can always use an escape for all unicode characters.
But like. I have to test the handling of markdown. Which says that some unicode characters result in another unicode character meant for “cannot control this nicely”.
So I don’t expect linters or TS or so to not like them.

Many algos in the HTML spec also result in REPLACEMENT CHARACTER. I don’t expect “regular” people to write them a lot but there are APIs that need to generate them, so there will be test files.

In my experience with GH internals, they have some heuristic with control characters and lone surrogates to infer whether something is binary.

RyanCavanaugh:
I don't want to make TS slower for every other file in the world, or forego doing other more impactful work, for the sake of a preference in the small number of scenarios where people need to write text encoding testcases.

We struggle with this too -- people are always asking us to normalize `\r\n` in our repo, etc, but we can't do that because TS needs to be able to successfully parse both `\r\n` and `\n` in files. At the point you're working with the byte structure of text instead of the semantic interpretation of it, things are going to be slightly less convenient compared to all other text content. Them's the breaks.

PR #57020, opened 2024-01-11
Do not remove binding alias in function declarations

### LINKED ISSUE #56991: Removing binding aliases causes duplicate identifiers in declarations.
### 🔎 Search Terms

binding rename duplicate identifier

- This changed between versions 4.9 and 5.0

### ⏯ Playground Link

[Playground Link](https://www.typescriptlang.org/play?ts=5.3.3#code/KYDwDg9gTgLgBAMwK4DsDGMCWEWJQJjgAoBvOMKCMALjgEMAacymuAIzgF9ayKraUSALZtgULgEo4ZTgG4AsACggA)

### 💻 Code

```ts
export function fn2 ({ prop: a, prop: b }: { prop: number }) { };

```

### 🙁 Actual behavior

Declaration emit produces a signature with duplicate identifiers:

```ts
export declare function fn2({ prop, prop }: {
    prop: number;
}): void;
```

### 🙂 Expected behavior

Declaration emit produces a signature correct signature:

```ts
export declare function fn2({ prop }: {
    prop: number;
}): void;
```

_No response_

### LINKED ISSUE #56992: Binding element alias is kept in function expressions even though they are never used in declaration files.
### 🔎 Search Terms

### ⏯ Playground Link

[Playground Link](https://www.typescriptlang.org/play?ts=4.3.5#code/KYDwDg9gTgLgBAYwgOwM7wGbIIxwLxwAUA3nGFBGAFwBGcAvlXKeZU8gK4C2NwUDASnwA+ZgwDcAWABQM0JFiIU6OFgBM+IiwrU6jMa2pxOPPoJFwaU2dPnR4SNJmQBmTSTI7aDJtrbHuXn56ASYYAE8wYAgMSwsABmsgA)

### 💻 Code

```ts
export const fn1 = ({ prop:b }: { prop: number }) => { };

export const fn2 = ({ prop:b }: { prop: number }) => b;

export const fn3 = ({ prop:b }: { prop: number }): typeof b => 0;

```

### 🙁 Actual behavior

`b` is kept in all exported function signatures, even though it is never used. Even if we have `typeof b` in the signature of the original function expression, the type printer will remove that and use the resolved type.

```ts
export declare const fn1: ({ prop: b }: {
    prop: number;
}) => void;
export declare const fn2: ({ prop: b }: {
    prop: number;
}) => number;
export declare const fn3: ({ prop: b }: {
    prop: number;
}) => number;
```

### 🙂 Expected behavior

`b` is removed from all signatures since it will never be used. Even if the original type uses `typeof` the resulting type will resolve any typeof expressions

Related to #55654

dragomirtitian:
Revert change that removes binding aliases from declaration files. Doing this correctly implies doing a lot more work than previously anticipated. 

Fixes #56991
Fixes #56992

### DIFF
--- a/tests/baselines/reference/renamingDestructuredPropertyInFunctionType.js
+++ b/tests/baselines/reference/renamingDestructuredPropertyInFunctionType.js
@@ -101,14 +101,14 @@ type O = {
     c: number;
 };
 type F1 = (arg: number) => any;
-type F2 = ({ a }: O) => any;
-type F3 = ({ a, b, c }: O) => any;
-type F4 = ({ a }: O) => any;
-type F5 = ({ a, b, c }: O) => any;
+type F2 = ({ a: string }: O) => any;
+type F3 = ({ a: string, b, c }: O) => any;
+type F4 = ({ a: string }: O) => any;
+type F5 = ({ a: string, b, c }: O) => any;
 type F6 = ({ a: string }: {
     a: any;
 }) => typeof string;
-type F7 = ({ a, b: number }: {
+type F7 = ({ a: string, b: number }: {
     a: any;
     b: any;
 }) => typeof number;
@@ -118,14 +118,14 @@ type F8 = ({ a, b: number }: {
 }) => typeof number;
 type F9 = ([a, b, c]: [any, any, any]) => void;
 type G1 = new (arg: number) => any;
-type G2 = new ({ a }: O) => any;
-type G3 = new ({ a, b, c }: O) => any;
-type G4 = new ({ a }: O) => any;
-type G5 = new ({ a, b, c }: O) => any;
+type G2 = new ({ a: string }: O) => any;
+type G3 = new ({ a: string, b, c }: O) => any;
+type G4 = new ({ a: string }: O) => any;
+type G5 = new ({ a: string, b, c }: O) => any;
 type G6 = new ({ a: string }: {
     a: any;
 }) => typeof string;
-type G7 = new ({ a, b: number }: {
+type G7 = new ({ a: string, b: number }: {
     a: any;
     b: any;
 }) => typeof number;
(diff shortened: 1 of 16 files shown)


jakebailey (approved):
All of the examples that are broken feel like they should be fixable; I know some of this I've tackled in `typeToTypeNode`, but the same strategies probably don't work in declaration emit. I'm fine reverting this so it doesn't go out incorrectly and we can try and tackle it another day.

jakebailey:
This also partially reverts #41044 as it removes all removal; I was actually expecting this to have caused d.ts errors because #41044 also introduced an error for unused renamings... but that didn't happen?

dragomirtitian:
> All of the examples that are broken feel like they should be fixable; I know some of this I've tackled in `typeToTypeNode`, but the same strategies probably don't work in declaration emit. I'm fine reverting this so it doesn't go out incorrectly and we can try and tackle it another day.

The problems are not unsolvable for sure. Even in declaration emit I think we ca solve it. But the problem is that it will introduce a lot of perf cost. Seems like it's not worth the complexity. (I belatedly arrived at the same conclusion as @andrewbranch 😅)


> This also partially reverts #41044 as it removes all removal; I was actually expecting this to have caused d.ts errors because #41044 also introduced an error for unused renamings... but that didn't happen?

The error for unused renames is still there. I did not remove it. It only seems to pop up on some kinds of declarations for example function types `let o: ({n : a}: any) => void;`

tests/baselines/reference/renamingDestructuredPropertyInFunctionType.js
```diff
@@ -101,14 +101,14 @@ type O = {
     c: number;
 };
 type F1 = (arg: number) => any;
-type F2 = ({ a }: O) => any;
-type F3 = ({ a, b, c }: O) => any;
-type F4 = ({ a }: O) => any;
-type F5 = ({ a, b, c }: O) => any;
+type F2 = ({ a: string }: O) => any;
```
jakebailey, line 104: > The error for unused renames is still there. I did not remove it. It only seems to pop up on some kinds of declarations for example function types `let o: ({n : a}: any) => void;`

Just to point out an example from the PR, this is code that I believed would have gained an error: https://www.typescriptlang.org/play?#code/C4TwDgpgBA8lC8UDeBYAUFTUCGB+AXFAM7ABOAlgHYDmA3OllAEaGUCuAtkxKfRlgGNWnbr3QBfPulCQoAMQBMCKAAokOQiQo0o4wjACUCAHw5KIWkA

Yet, there's no error baseline. That being said, it looks like for declaration files we straight up skip checking the final list:

```
if (!node.isDeclarationFile) {
    checkPotentialUncheckedRenamedBindingElementsInTypes();
}
```

So, I think everything is fine? (Other than that I don't think it makes any sense that we still collect `potentialUnusedRenamedBindingElementsInTypes` in declaration files only to not read them later, which is not new.)

jakebailey:
After checking the above and noticing https://github.com/microsoft/TypeScript/pull/41044#issuecomment-1146271768, I'm now confident that this PR is working correctly.

PR #57027, opened 2024-01-11
Dom update 2024-01-11

sandersn:
Not many breaks from this one. Two packages have their shims break and one package assigns to a property that's not there anymore (but might still be on the browser it targets?)

Performance:
- the change makes the compiler use slightly more memory

Failures:
- webrtc in DT (because it's a shim for WebRTC*)
- pixijs incorrectly extends CanvasTextDrawingStyles -- because it was shimming letterSpacing, which was added *non*-optional.
- tldraw assigns to now-missing FontFaceDescriptors.variant, though it's "just" an excess property error and will be ignored.

Summary of changes: 
- AuthenticationExtensionClientInputs pinLength new
- BufferSource -> AllowSharedBufferSource in 18 places
- NotificationAction deleted, and several properties of NotificationOptions
- new properties in RTCEncodedAudio/VideoMetadata
- other minor additions/removals in RTC*
- WebTransportSendStreamOptions.sendOrder removes null from type
- MIDIConnectionEvent.port adds null to type
- More MDN links in jsdoc
- minor additions in CSS\*
- CSSStyleRule now extends CSSGroupRule instead of CSSRule
- CredentialsContainer.store no longer returns `Credential`, just `void`
- FontFace.variant deleted
- Gamepad.hapticActuators deleted
- New event: beforetoggle
- minor additions in HTML\*Element
- add Highlight/HighlightRegistry
- midimessage now has type MIDIMessageEvent, not just Event
- statechange now has type MIDIConnectionEvent
- icecandidateerror now has type RTCPeerConnectionIceErrorEvent
- minor additions in RTC\*
- RTCRtpSender.setParameters has new optional parameter
- new property: SVGImageElement.crossOrigin
- TextMetrics adds a few new properties
- VideoFrame constructors now contruct from AllowSharedBufferSource instead of BufferSource
- WEBGL_\* switches from GLuint to number (also switches to AllowSharedBufferSource some places)
- Window adds ondeviceorientationabsolute event


sandersn:
@typescript-bot test top100
@typescript-bot run dt
@typescript-bot pack this

BickelLukas:
> NotificationAction deleted, and several properties of NotificationOptions

@sandersn Any insight as to why these have been deleted? Looking at the documentation on mdn these properties should exist: https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerRegistration/showNotification

sandersn:
It was based on https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/1631, which updates the types to @webref/idl@3.39.1 and browser-compat-data@5.3.21

The comments on that PR says that NotificationOptions is Blink-only as of 2023-10.

src/lib/dom.generated.d.ts
```diff
@@ -1000,27 +1000,16 @@ interface NavigationPreloadState {
...
     icon?: string;
-    image?: string;
     lang?: string;
-    renotify?: boolean;
     requireInteraction?: boolean;
     silent?: boolean | null;
     tag?: string;
-    timestamp?: EpochTimeStamp;
-    vibrate?: VibratePattern;
 }
```
RulixConception, line 1013: Why were those properties removed?

dlarocque: @RulixConception 
https://github.com/microsoft/TypeScript-DOM-lib-generator?tab=readme-ov-file#when-the-type-is-missing

PR #57029, opened 2024-01-11
Take into account peerDependency versions when calculating packageId

### LINKED ISSUE #56762: pnpm peer dependency handling conflicts with source file redirection / deduplication
### 🔎 Search Terms

 error TS2741

version: latest.

### ⏯ Playground Link

https://github.com/bestlyg/component-type-checker

### 💻 Code

_No response_

### 🙁 Actual behavior

error TS2741.

### 🙂 Expected behavior

Normal compilation.

_No response_

sheetalkamat:
Fixes #56762


src/compiler/moduleNameResolver.ts
```diff
@@ -2392,6 +2396,44 @@ function getVersionPathsOfPackageJsonInfo(packageJsonInfo: PackageJsonInfo, stat
...
+    if (state.traceEnabled) trace(state.host, Diagnostics.package_json_has_a_peerDependencies_field);
+    const packageDirectory = realPath(packageJsonInfo.packageDirectory, state.host, state.traceEnabled);
+    if (packageDirectory.indexOf(".pnpm") === -1) {
+        // For now skip extra work for anything else
+        if (state.traceEnabled) trace(state.host, Diagnostics.Package_is_not_installed_using_pnpm_skipping_peerDependencies_version_resolution);
+        return undefined;
+    }
+    const nodeModules = packageDirectory.substring(0, packageDirectory.lastIndexOf("node_modules") + "node_modules".length) + directorySeparator;
+    let result = "";
+    if (state.traceEnabled) {
```
jakebailey, line 2418: If trace isn't enabled, won't this do nothing and return the empty string?

sheetalkamat: oops

jakebailey: No tests changed; were they generated before this code was added and so the fix is still workign?

sheetalkamat: This only takes place if it’s at .pnpm location and has peer dependencies.. as far as I know it’s the only test I have and it was with trace resolutions so didn’t change any baseline

jakebailey: Ah, tracing was enabled, right.

jakebailey:
@typescript-bot test top200
@typescript-bot run dt

@typescript-bot pack this

src/compiler/moduleNameResolver.ts
```diff
@@ -2392,6 +2396,42 @@ function getVersionPathsOfPackageJsonInfo(packageJsonInfo: PackageJsonInfo, stat
...
+    }
+    return packageJsonInfo.contents.peerDependencies || undefined;
+}
+
+function readPackageJsonPeerDependencies(packageJsonInfo: PackageJsonInfo, state: ModuleResolutionState): string | undefined {
+    const peerDependencies = readPackageJsonField(packageJsonInfo.contents.packageJsonContent, "peerDependencies", "object", state);
+    if (peerDependencies === undefined) return undefined;
+    if (state.traceEnabled) trace(state.host, Diagnostics.package_json_has_a_peerDependencies_field);
+    const packageDirectory = realPath(packageJsonInfo.packageDirectory, state.host, state.traceEnabled);
+    if (packageDirectory.indexOf(".pnpm") === -1) {
```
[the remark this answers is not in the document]
jakebailey: > ...? I don't think we should have anything pnpm-cache-directory-name-specific; this is even configurable within `pnpm` itself using `pnpm config set store-dir D:/arbitary-path` - it could be anywhere by any name

The "store" is the global store where pnpm stores its global cache; installed projects hardlink to those files into `node_modules/.pnpm` and that can't be customized.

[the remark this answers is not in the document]
weswigham: I believe it actually is customizable, by the related [`virtual-store-dir`](https://github.com/pnpm/pnpm/blob/d715f87b60ce79be741435ab9d4ce551f2b097c4/config/config/src/index.ts#L265) config name.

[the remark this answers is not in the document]
jakebailey: Ah, yeah. https://pnpm.io/npmrc#virtual-store-dir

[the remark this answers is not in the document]
merceyz: > Moreover, you can imagine other package managers adopting this structured symlink behavior at any time

For reference Yarn in pnpm mode uses `node_modules/.store` (symlinks) and in PnP mode `.yarn/__virtual__` (virtual file system).

sheetalkamat:
@typescript-bot test top200
@typescript-bot run dt

@typescript-bot pack this

sheetalkamat:
@weswigham @andrewbranch @jakebailey  can you please take a look again so we can get this in for 5.5 early on.

src/compiler/moduleNameResolver.ts
```diff
@@ -2400,6 +2404,37 @@ function getVersionPathsOfPackageJsonInfo(packageJsonInfo: PackageJsonInfo, stat
...
+    if (state.traceEnabled) trace(state.host, Diagnostics.package_json_has_a_peerDependencies_field);
+    const packageDirectory = realPath(packageJsonInfo.packageDirectory, state.host, state.traceEnabled);
+    const nodeModules = packageDirectory.substring(0, packageDirectory.lastIndexOf("node_modules") + "node_modules".length) + directorySeparator;
+    let result = "";
+    for (const key in peerDependencies) {
+        if (hasProperty(peerDependencies, key)) {
+            const peerPackageJson = getPackageJsonInfo(nodeModules + key, /*onlyRecordFailures*/ false, state);
+            if (peerPackageJson) {
+                const version = (peerPackageJson.contents.packageJsonContent as PackageJson).version;
+                result += `+${key}@${version}`;
```
weswigham, line 2426: Q: I could theoretically name a package on disk `typescript@5.4.1+` and munge this cache key up pretty badly (all these separators are valid filepath characters after all). Shouldn't this use, eg, `?` and `|` instead of `+` and `@`, since, at least on windows, those aren't valid path characters? Believe it or not, a `require("+typescript@5.4.1")` will work and resolve on disk, even if `npm` won't let you publish a package with those characters!

sheetalkamat: eg. if your file has import whose peer dependency is the source File for the package then short cut to directly use package name?

sheetalkamat:
```ts 
import { a } from "foo";
export functopn bar() { 
return a.bar(); // This is from package Bar@BarVersion
}
```
then while going through all imports in the file: if you find say "foo" import whose peer dependency is "Bar@Bar.version" use that package name ?

andrewbranch: Maybe that could be leveraged somehow, but getting just the package name is usually fairly easy compared to figuring out what subpath can be used to get to the export we need.

sheetalkamat: @weswigham @andrewbranch  do we still have concern about the string for peer dependency calculations.  
Do we want this as array of package name and versions for future use or just some string is ok. Should i make this internal for now so we have option to change this later? Would like to get this in sooner than later

andrewbranch: I think it should probably be internal if it’s a formatted string that only gets used as a lookup id. But if you want to go ahead and change it to be an object with name/version, that works too.

sheetalkamat:
@andrewbranch marked peerDependencies as internal for now

PR #53522, opened 2023-03-27
Avoid infinite recursion when instantiating circular inline mapped generic tuple type

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/53458

Issue #53458, opened 2023-03-23
Stack overflow error when using inline mapped tuples


danny-may:
# Bug Report

### 🔎 Search Terms
`Maximum call stack size exceeded`
`getMappedType`

### 🕗 Version & Regression Information

- This is a crash
- This changed between versions 4.6.4 and 4.7.4

### ⏯ Playground Link

[Playground link with relevant code](https://www.typescriptlang.org/play?ts=5.0.2#code/MYGwhgzhAEBiD28A8BREBTAtugdgFxnQA89cATGAJ3TDPhxAE9oBXHAax3gHccBtALoA+aAG8AsAChoM6AAcWAIxABLYNGq16TaOgzZ8EAFxjofAArQVOaO3SN4AM2hosuAgJOjoisJROuBgQWAgDc0AC+kaFSUrLySqrqwPQQeJQswHjwlAAUAHSFem6GXmaW1rb2Ti767hCepr7+tSXB5gKRkQCUYnHxsngAFioQ+cVBMAC8unWGMdKyEbGLMgrKatC0ZLndJgjIfIX5gfUANNAA5GCKwJfCfasD1HgslDY46NxwiEhHhadDBdrrd7kICoVhqNxnMCBdvM0TCC7j0FvFlpIIkA)

[This is an example of a logically identical class, with the inline mapping extracted, where the compiler doesnt crash](https://www.typescriptlang.org/play?ts=5.0.2#code/C4TwDgpgBAsghmSATAKgVzAGwgHgKLYC2EAdsAM5QQAewpSlAThHEgPYmYhRokDWJNgHcSAbQC6APigBeKAG8oogApQAliSh8IINgDMoBCMTLlxALgVQARnEaWjJiivEBuKAF9PAWABQAY0w4ckoAMTY2fCJSCipaeiYWdk5uXgFhMSkFPyhcqDA0a0w1fyhmVg4uKmjTS3hECFQMbCjjGPJJVz8cvIKikqh-DnJgRjR-YDZGAAoAOnmIGoo6hGR0LFxHdskASmzfPMOoYAALNXJZxbbTWWrrii6DvI9up9y+4tLWJGmdy3DIqJ5rMtqYADRQADkcGs-khWXkPSOZQgwDQjE0JAgQigAJwQPmoIoEOhsPhkjm81O50uS3IEMUtnsUJhcM8O0ehxevm5QA)

### 💻 Code

```ts
class Foo<Elements extends readonly unknown[]> {
[...]

PR #55774, opened 2023-09-18
Add numeric constraint to type parameter of mapped types with name type and array constraints

Andarist:
closes https://github.com/microsoft/TypeScript/issues/55762
expands on the logic introduced in https://github.com/microsoft/TypeScript/pull/48837

PR #57031, opened 2024-01-11
Revise mapped tuple type instantiation logic

### LINKED ISSUE #48856: Key in mapped variadic tuple should not be literal after the variadic part
# Bug Report

### 🔎 Search Terms

Variadic tuple mapped type keyof

### ⏯ Playground Link

[Playground link with relevant code](https://www.typescriptlang.org/play?ts=4.6.2#code/C4TwDgpgBA0hIGcA8B5KEAewIDsAmCUArjgNY4D2A7jgNoC6AfFALxQDeUtMUAljlFLwKAMygp6ALlhQAvgG4AUItCQoAQVax4yWgmAAnfgHMANFAB0V-UZzGGTJQHonUNwD0A-CvDQAQlpwiEh6hibmVhY2JgzmOEQAtgBGEAaOii5uUF6KQA)

### 💻 Code

```ts
type Keys<O extends unknown[]> = { [K in keyof O]: K };

type A = Keys<[string, ...string[]]>;
type B = Keys<[string, ...string[], number]>;
```

### 🙁 Actual behavior

`A` and `B` both have the variadic part typed as `..."1"[]`, which is wrong. `B` is further wrong because the `number` part becomes `"2"`, while in practice it would rarely be `"2"`.

### 🙂 Expected behavior

`A` should be `["0", ...number[]]`; `B` should be `["0", ...number[], number]`

### LINKED ISSUE #56888: `Readonly<[...unknown[], infer L]>` and `readonly [...unknown[], infer L]` exhibit inconsistent behavior.
### 🔎 Search Terms

`Readonly<[...unknown[], infer L]>` and `readonly [...unknown[], infer L]` exhibit inconsistent behavior.

-

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.3.2#code/C4TwDgpgBAglC8UBKECGATA9gOwDYgB4BtAOjIFdsBrbTAd2yIF0AaKSm+7JgPgFgAUFCgB6EVAB6AfkGDQkKACEEUAE5oseEFFIVqtBszYcD3QcLGSZAueGgAVAIwqiAclSu2rgEaeorgGNXJigIAA9gCGx0AGc1DRx8HQBLbAAzCFUoAAk2MhITLmYQqRyoAC4obAgAN0zzUXFpWQF5BwAmF3c-Hx6gkPDI6LiUDETCIlSMrNyofMLDJl4oUuyKqtr6oUarFraoewBmLo8vXy9+0Iio2PixrR15-SLWKCnMqAAZEq-16rrVA1LM0bK07AcACwnHrnfyXQY3EYJLTEJ6cRZsd5Zb48Fa-Sr-LYWJrWYmwALAcioXCVBbYIHiACiYUgFIg6EqgVcgiAA

### 💻 Code

```ts
type A = Readonly<[...unknown[], unknown]>
  // ^? type A = readonly [...unknown[], unknown]

type B = readonly [...unknown[], unknown]
  // ^? type B = readonly [...unknown[], unknown]

type T1 = ['a', 'b', 'c'] extends readonly [infer H, ...unknown[]] ? H : never
  // ^? type T1 = 'a'

type T2 = ['a', 'b', 'c'] extends Readonly<[infer H, ...unknown[]]> ? H : never
  // ^? type T2 = 'a'

type T3 = ['a', 'b', 'c'] extends readonly [...unknown[], infer L] ? L : never
  // ^? type T3 = 'c'

type T4 = ['a', 'b', 'c'] extends Readonly<[...unknown[], infer L]> ? L : never
  // ^? type T4 = unknown
  // Actual: unknown
  // Expected: 'c'
```

### 🙁 Actual behavior

-

### 🙂 Expected behavior

Infer the correct type.

_No response_

ahejlsberg:
This PR revises and simplifies our mapped tuple type instantiation logic. Previously, when a mapped type was applied to a variadic tuple type, we'd transform the operation into a sequence of operations on manufactured singleton tuple types. This would have surprising effects when observing property key types, as reported in #48856. For example:

```ts
type Keys<T> = { [K in keyof T]: K };
type Foo<T extends unknown[]> = Keys<[string, string, ...T, string]>;  // ["0", "0", ...Keys<T>, "0"]
```

With this PR, the property keys are more consistent:

```ts
type Keys<T> = { [K in keyof T]: K };
type Foo<T extends unknown[]> = Keys<[string, string, ...T, string]>;  // ["0", "1", ...Keys<T>, number]
```

The PR also fixes issues related to instantiation of tuples with required elements following a rest element, as reported in #56888, and it fixes the error reporting issue discussed [here](https://github.com/microsoft/TypeScript/pull/57031#discussion_r1449576651).

Finally, the PR reverts #53522. The fix in that PR is incorrect--it effectively leaks a type parameter. With this PR, the repro from #53458 now causes an excessive instantiation depth error, which should have been the behavior in the first place.

Fixes #48856.
Fixes #56888.


ahejlsberg:
@typescript-bot test top100
@typescript-bot run dt

src/compiler/checker.ts
```diff
@@ -19769,26 +19766,28 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
-        // When a tuple type is generic (i.e. when it contains variadic elements), we want to eagerly map the
-        // non-generic elements and defer mapping the generic elements. In order to facilitate this, we transform
-        // M<[A, B?, ...T, ...C[]] into [...M<[A]>, ...M<[B?]>, ...M<T>, ...M<C[]>] and then rely on tuple type
-        // normalization to resolve the non-generic parts of the resulting tuple.
+    function instantiateMappedTupleType(tupleType: TupleTypeReference, mappedType: MappedType, typeVariable: TypeVariable, mapper: TypeMapper) {
+        // We apply the mapped type's template type to each of the fixed part elements. For variadic elements, we
+        // apply the mapped type itself to the variadic element type. For other elements in the variable part of the
+        // tuple, we surround the element type with an array type and apply the mapped type to that. This ensures
+        // that we get sequential property key types ("0", "1", "2", etc.) for the fixed part of the tuple, and
+        // property key type number for the remaining elements.
```
DanielRosenwasser, line 19774: Can you keep the original example in place? Or is it too different now for some reason?

DanielRosenwasser: and as you note the example with the keys, maybe include the example from your PR description as well?

```ts
type Keys<T> = { [K in keyof T]: K };
type Foo<T extends unknown[]> = Keys<[string, string, ...T, string]>;  // ["0", "1", ...Keys<T>, number]
```

ahejlsberg: Problem is, the original transformation (as witnessed by that example) was inconsistent with regards to property keys, so not sure there's any value in keeping it.

src/compiler/checker.ts
```diff
@@ -19769,26 +19766,28 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
-            }
-            // The singleton is never a generic tuple type, so it is safe to recurse here.
-            return instantiateMappedType(mappedType, prependTypeMapping(typeVariable, singleton, mapper));
+        const fixedLength = tupleType.target.fixedLength;
+        const fixedMapper = fixedLength ? prependTypeMapping(typeVariable, tupleType, mapper) : mapper;
+        const newElementTypes = map(getElementTypes(tupleType), (type, i) => {
+            const flags = elementFlags[i];
+            return i < fixedLength ? instantiateMappedTypeTemplate(mappedType, getStringLiteralType("" + i), !!(flags & ElementFlags.Optional), fixedMapper) :
+                flags & ElementFlags.Variadic ? instantiateType(mappedType, prependTypeMapping(typeVariable, type, mapper)) :
+                getElementTypeOfArrayType(instantiateType(mappedType, prependTypeMapping(typeVariable, createArrayType(type), mapper))) || unknownType;
```
DanielRosenwasser, line 19782:
```suggestion
                getElementTypeOfArrayType(instantiateType(mappedType, prependTypeMapping(typeVariable, createArrayType(type), mapper))) ?? unknownType;
```

tests/cases/conformance/types/mapped/mappedTypesGenericTuples.ts
```diff
@@ -0,0 +1,45 @@
...
+
+// Property keys are `number` following the fixed part of a tuple
+
+type K<T> = { [P in keyof T]: P };
+type M<T> = { [P in keyof T]: T[P] };
+
+type KA = K<[string, string, boolean]>;  // ["0", "1", "2"]
+type KB = K<[string, string, ...string[], string]>;  // ["0", "1", ...number[], number]
+type KC = K<[...string[]]>;  // number[]
+type KD = K<string[]>;  // number[]
```
DanielRosenwasser, line 12: I know it's not allowed to have non-generic variable elements follow a non-generic rest, but it would be good to make sure the checker tolerates states like that

```suggestion
type KD = K<string[]>;  // number[]
type KE = K<[string, ...boolean[], number?]>;
type KE = K<[string, ...boolean[], ...number[]]>;
```

DanielRosenwasser:
Looks good, left some comments that should be quick to address.

src/compiler/checker.ts
```diff
@@ -19745,11 +19745,8 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                                 ) {
                                     return instantiateMappedArrayType(t, type, prependTypeMapping(typeVariable, t, mapper));
                                 }
-                                if (isGenericTupleType(t)) {
-                                    return instantiateMappedGenericTupleType(t, type, typeVariable, mapper);
-                                }
                                 if (isTupleType(t)) {
-                                    return instantiateMappedTupleType(t, type, prependTypeMapping(typeVariable, t, mapper));
+                                    return instantiateMappedTupleType(t, type, typeVariable, mapper);
```
Andarist, line 19749: An offtopic question if I may. I noticed (on several occasions already) that instantiating array/tuple mapped types eagerly~ leads to inconsistent behaviors with the object variants. The timing of instantiation is quite different and thus the availability of certain information tends to be different between them. 

What are your thoughts on making this more consistent? How feasible that would even be? When it comes to objects each property stays deferred for as much as possible and I imagine that this would require deferring this instantiation per tuple element.

tests/cases/conformance/types/mapped/mappedTypesGenericTuples.ts
Andarist: Since you bring this up here... this is killing me 😅 
```ts
type A = [number, ...boolean[], ...string[]] // error
type B = [number, ...Array<boolean>, ...Array<string>] // ok
```

Shouldn't those really behave the same way? IIRC I have a PR laying around somewhere that makes this error go away (by matching the behavior of the latter)

src/compiler/checker.ts
DanielRosenwasser: Yeah, I was thinking of trying to update the example, not keep it.

I guess the issue you might've run into is that we don't have great notation to describe this. The closest thing being something like:

```
We transform

M<[A, B?, ...T, ...C[]]

into

[M<{ 0: A }>[0], M<{ 1: B }>[1]?, ...M<T>, ...M<C[]>]
```

which is not *too* terrible, but I think having an example still has value.

tests/cases/conformance/types/mapped/mappedTypesGenericTuples.ts
DanielRosenwasser: Is the PR already open?

Not that it's related, but you inspired me to file something else I've put off mentioning for a while: #57034

DanielRosenwasser: It is probably not the intention that we *want* to allow that code to be honest. But I'll let Anders weigh in.

Andarist: > Is the PR already open?

sure thing :) https://github.com/microsoft/TypeScript/pull/55446

src/compiler/checker.ts
ahejlsberg: I'll add the `Keys` example. In general, we now process the fixed part of the tuple just like any other object. It's really just the elements in the variable part that get special processing.

src/compiler/checker.ts
ahejlsberg: Sure, will fix.

tests/cases/conformance/types/mapped/mappedTypesGenericTuples.ts
ahejlsberg: @DanielRosenwasser Those extra examples wouldn't really check anything related to this PR since we normalize the tuple type before the instantiation logic even sees it.

ahejlsberg: @Andarist I still think the error makes sense, but we should fix the error reporting logic to be consistent between the notations. Permitting `[...string[], ...number[]]` with no errors will just make users think we can actually check for a sequence of strings followed by a sequence of numbers, but that's not the case.

Andarist: I'd assume that people might already be relying on the fact spreading like this might "normalize" the shape, so I thought that is off the table as it would be considered a breaking change.

src/compiler/checker.ts
ahejlsberg: In general, the advantage of eager instantiation is it is simpler to implement and it allows sharing of identical instantiations. The specific reason we eagerly instantiate mapped array and tuple types is that elsewhere in the compiler we detect arrays and tuples by looking for instantiations of `Array<T>` and the corresponding synthetic tuple classes. We'd have to extend those cases to also handle mapped types applied to arrays and tuples.

tests/cases/conformance/types/mapped/mappedTypesGenericTuples.ts
ahejlsberg: Error reporting fixed with the latest commit.

tests/cases/conformance/types/tuple/variadicTuples2.ts
```diff
@@ -11,6 +11,11 @@ type V10 = [number, ...string[], ...boolean[]];  // Error
 type V11 = [number, ...string[], boolean?];  // Error
 type V12 = [number, string?, boolean];  // Error
 
+type V15 = [...string[], ...number[]];  // Error
+type V16 = [...string[], ...Array<number>];  // Error
+type V17 = [...Array<string>, ...number[]];  // Error
+type V18 = [...Array<string>, ...Array<number>];  // Error
```
Andarist, line 17: I propose some extra test cases with type aliases:
```suggestion
type Alias19 = number[];
type V19 = [...string[], ...Alias19];  // Error
type Alias20 = Array<number>;
type V20 = [...string[], ...Alias20];  // Error
```

src/compiler/checker.ts
Andarist: Thank you very much for the answer. I might want to attempt unifying this at some point - it's good to know that you don't have **strong** objections 😉 Or at least I read that from your comment ;p I understand that this might introduce some new complexity but I think it might be inevitable (as long as the goal is to have consistent behavior). When trying to do this, I will gather the test cases backing up that it's a desired change.

Andarist:
It might be worth rechecking if the recently merged in https://github.com/microsoft/TypeScript/pull/55774 doesn't throw any wrinkles into this.

jakebailey:
I checked, and this PR still passes with main merged in (unless you were thinking of something other effect).

Andarist:
Unfortunately, https://github.com/microsoft/TypeScript/pull/55774 got reverted because it weirdly clashed with another PR that was merged recently ( https://github.com/microsoft/TypeScript/pull/57091 ). However, I didn't think that a merge of those would fail any existing test cases but rather that *maybe* some new test cases should be written in light of this PR. I just had a hunch of sorts that maybe there is something to be re-checked but I didn't have the energy myself at the time to think about the actual test cases.

RyanCavanaugh (approved):

PR #57033, opened 2024-01-12
Check `autoImportExcludePatterns` against symlink paths

### LINKED ISSUE #55111: autoImportFileExcludePatterns has no effect on node modules.
# Bug Report

### 🔎 Search Terms

autoImportFileExcludePatterns

Visual Studio Code v1.80.1, Typescript v5.1

### ⏯ Playground Link

None.

### 💻 Code

```
{
    "typescript.preferences.autoImportFileExcludePatterns": [
        "**/node_modules/@xenon.js/core/**/*"
    ]
}
```

or

```
{
    "typescript.preferences.autoImportFileExcludePatterns": [
        "@xenon.js/core"
    ]
}
```

### 🙁 Actual behavior

Will still be automatically imported from the package `@xenon.js/core`.

I want to exclude the automatic import of the npm package @xenon.js/core from the project. I set autoImportFileExcludePatterns, but the automatic import is still triggered.
image
If what I add to autoImportFileExcludePatterns is a relative path, it works, but if it's an npm package, no matter what the package is.

### 🙂 Expected behavior

Will not be automatically imported from these paths.

andrewbranch:
The `autoImportExcludePatterns` entry was `@remix-run/server-runtime`, but yarn mangles the realpath of that package to `@remix-run-server-runtime-{hash}`. We need to see if some filename is actually the realpath of a symlinked directory, and check the exclude patterns against the symlink locations too.

Fixes #55111

### DIFF
--- a/src/services/exportInfoMap.ts
+++ b/src/services/exportInfoMap.ts
@@ -8,6 +8,7 @@ import {
     createMultiMap,
     Debug,
     emptyArray,
+    ensureTrailingDirectorySeparator,
     findIndex,
     firstDefined,
     forEachAncestorDirectory,
@@ -43,6 +44,7 @@ import {
     nodeModulesPathPart,
     PackageJsonImportFilter,
     Path,
+    pathContainsNodeModules,
     Program,
     skipAlias,
     skipOuterExpressions,
@@ -427,12 +429,12 @@ export function forEachExternalModuleToImportFrom(
         return pattern ? getRegexFromPattern(pattern, useCaseSensitiveFileNames) : undefined;
     });
 
-    forEachExternalModule(program.getTypeChecker(), program.getSourceFiles(), excludePatterns, (module, file) => cb(module, file, program, /*isFromPackageJson*/ false));
+    forEachExternalModule(program.getTypeChecker(), program.getSourceFiles(), excludePatterns, host, (module, file) => cb(module, file, program, /*isFromPackageJson*/ false));
     const autoImportProvider = useAutoImportProvider && host.getPackageJsonAutoImportProvider?.();
     if (autoImportProvider) {
         const start = timestamp();
         const checker = program.getTypeChecker();
-        forEachExternalModule(autoImportProvider.getTypeChecker(), autoImportProvider.getSourceFiles(), excludePatterns, (module, file) => {
+        forEachExternalModule(autoImportProvider.getTypeChecker(), autoImportProvider.getSourceFiles(), excludePatterns, host, (module, file) => {
             if (file && !program.getSourceFile(file.fileName) || !file && !checker.resolveName(module.name, /*location*/ undefined, SymbolFlags.Module, /*excludeGlobals*/ false)) {
                 // The AutoImportProvider filters files already in the main program out of its *root* files,
                 // but non-root files can still be present in both programs, and already in the export info map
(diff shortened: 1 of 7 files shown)


src/services/exportInfoMap.ts
```diff
@@ -445,15 +447,32 @@ export function forEachExternalModuleToImportFrom(
     }
 }
 
-function forEachExternalModule(checker: TypeChecker, allSourceFiles: readonly SourceFile[], excludePatterns: readonly RegExp[] | undefined, cb: (module: Symbol, sourceFile: SourceFile | undefined) => void) {
-    const isExcluded = excludePatterns && ((fileName: string) => excludePatterns.some(p => p.test(fileName)));
+function forEachExternalModule(checker: TypeChecker, allSourceFiles: readonly SourceFile[], excludePatterns: readonly RegExp[] | undefined, host: LanguageServiceHost, cb: (module: Symbol, sourceFile: SourceFile | undefined) => void) {
+    const realpathsWithSymlinks = host.getSymlinkCache?.().getSymlinkedDirectoriesByRealpath();
+    const isExcluded = excludePatterns && (({ fileName, path }: SourceFile) => {
+        if (excludePatterns.some(p => p.test(fileName))) return true;
+        if (realpathsWithSymlinks?.size && pathContainsNodeModules(fileName)) {
```
andrewbranch, line 454: This `pathContainsNodeModules` check may be too conservative. The real world cases I’ve seen all have both realpaths and symlink paths somewhere inside a directory called `node_modules`. But if there are any package manager strategies where the realpaths live in a directory that _doesn’t_ contain `/node_modules/`, this check will fail.

src/services/exportInfoMap.ts
```diff
@@ -445,15 +447,32 @@ export function forEachExternalModuleToImportFrom(
...
-function forEachExternalModule(checker: TypeChecker, allSourceFiles: readonly SourceFile[], excludePatterns: readonly RegExp[] | undefined, cb: (module: Symbol, sourceFile: SourceFile | undefined) => void) {
-    const isExcluded = excludePatterns && ((fileName: string) => excludePatterns.some(p => p.test(fileName)));
+function forEachExternalModule(checker: TypeChecker, allSourceFiles: readonly SourceFile[], excludePatterns: readonly RegExp[] | undefined, host: LanguageServiceHost, cb: (module: Symbol, sourceFile: SourceFile | undefined) => void) {
+    const realpathsWithSymlinks = host.getSymlinkCache?.().getSymlinkedDirectoriesByRealpath();
+    const isExcluded = excludePatterns && (({ fileName, path }: SourceFile) => {
+        if (excludePatterns.some(p => p.test(fileName))) return true;
+        if (realpathsWithSymlinks?.size && pathContainsNodeModules(fileName)) {
+            let dir = getDirectoryPath(fileName);
+            let dirPath = getDirectoryPath(path);
+            while (dirPath !== "/") {
```
jakebailey, line 457: I'm not sure that this will work on Windows; `getDirectoryPath` is documented as returning the drive letter at the top level, so this loop will never complete.

andrewbranch: Good catch. I forgot `forEachAncestorDirectory` is a thing.

src/services/exportInfoMap.ts
andrewbranch: Checked yarn pnp and it contains both `node_modules` and the real/original package name, so that one’s fine. I think this logic is ok.

image

src/services/exportInfoMap.ts
```diff
@@ -445,15 +447,30 @@ export function forEachExternalModuleToImportFrom(
...
 }
 
-function forEachExternalModule(checker: TypeChecker, allSourceFiles: readonly SourceFile[], excludePatterns: readonly RegExp[] | undefined, cb: (module: Symbol, sourceFile: SourceFile | undefined) => void) {
-    const isExcluded = excludePatterns && ((fileName: string) => excludePatterns.some(p => p.test(fileName)));
+function forEachExternalModule(checker: TypeChecker, allSourceFiles: readonly SourceFile[], excludePatterns: readonly RegExp[] | undefined, host: LanguageServiceHost, cb: (module: Symbol, sourceFile: SourceFile | undefined) => void) {
+    const realpathsWithSymlinks = host.getSymlinkCache?.().getSymlinkedDirectoriesByRealpath();
+    const isExcluded = excludePatterns && (({ fileName, path }: SourceFile) => {
+        if (excludePatterns.some(p => p.test(fileName))) return true;
+        if (realpathsWithSymlinks?.size && pathContainsNodeModules(fileName)) {
+            let dir = getDirectoryPath(fileName);
```
jakebailey, line 455: Just so I'm clear, this is needed because we're basically doing `forEachAncestorDirectory` on two paths at once?

Could `dir` bottom out before `path` does?

src/services/exportInfoMap.ts
```diff
@@ -445,15 +447,30 @@ export function forEachExternalModuleToImportFrom(
...
+            return forEachAncestorDirectory(getDirectoryPath(path), dirPath => {
+                const symlinks = realpathsWithSymlinks.get(ensureTrailingDirectorySeparator(dirPath));
+                if (symlinks) {
+                    return symlinks.some(s => excludePatterns.some(p => p.test(fileName.replace(dir, s))));
```
jakebailey, line 459: Is `replace` enough here? If we're on a system with `/`, then `dir` can start with `/` which could match basically anywhere in `fileName`. Maybe this is rare but does it need a `prefix + '/'` check?

src/services/exportInfoMap.ts
sheetalkamat: Doesnt npm link let your realpath be "say c:/typescript" from say "c:/myproject/node_modules/typescript" ? Or is that not what you are checking here?

andrewbranch: Hm, yes, I think that’s a case where this would fail. I’m not sure how important that is though. If you’ve bothered to npm link something you _probably_ want to be able to auto-import from it? I was hoping to have some sort of filter here for performance reasons.

src/services/exportInfoMap.ts
andrewbranch: `dirPath === toPath(dir)` (i.e., `dir` has original casing preserved, but we need a `Path` for cache lookups). So no, they’ll always have the same number of path components.

src/services/exportInfoMap.ts
andrewbranch: `dir` is necessarily a prefix of `fileName` because it’s just `getDirectoryPath(fileName)` (repeated some number of times)

src/services/exportInfoMap.ts
jakebailey: Oh, I think I'm just getting confused because this code uses both `fileName` and `path` off of SourceFile, which are the same thing but with different types?

andrewbranch: `path` is normalized to lowercase on case-insensitive file systems, whereas `fileName` retains the casing as it exists on disk. I’m not sure it actually matters here because the exclude regex is also case-insensitive on case-insensitive file systems, so I _think_ we could just use `path` and it would always work, but it felt more correct to check the pattern against the original filename, not the `Path`-ified one.

jakebailey: Oh, thanks. I legitimately didn't know.

PR #57042, opened 2024-01-12
Implement a `@nonnull` JSDoc tag

### LINKED ISSUE #23405: Support a @nonnullable JSDoc comment
This could be used in place of the non-null assertion operator, and solve #23403.

```js
while (queue.length) {
    (/** @nonnull */ queue.pop())();
}
```

Related is #23217.

DanielRosenwasser:
Fixes #23405.

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -587,6 +587,7 @@ import {
     isJSDocNameReference,
     isJSDocNode,
     isJSDocNonNullableType,
+    isJSDocNonNullExpression,
     isJSDocNullableType,
     isJSDocOptionalParameter,
     isJSDocOptionalType,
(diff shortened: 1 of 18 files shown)


NemoStein:
TS 5.5 beta is released, but this isn't merged yet... Are there any plans to ship this with 5.5 or will it be pushd forward?

codenomnom:
Any news on this one? It's been a while...

tats-u:
@sandersn You should leave a review to other members if you are too busy to review it.
About 1 year has been passed.

aweebit:
`@nonnull` is a pretty bad name for this, makes it feel like it's only about `null`, and not `undefined`.

`@nonnullable` is better as it corresponds to the `NonNullable` utility type's name.

`@nonnullish` with nullish as in [the nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) meaning exactly that, the `null` and `undefined` values, is also a good alternative.

But my favorite option is actually `/** @type {!} */`.

louwers:
I like nonnullable better, it is a bit more verbose, but JSDoc does not shy away from some verbosity.

PR #57051, opened 2024-01-14
feat(57028): Support inlining variables that are used in object literal shorthand

### LINKED ISSUE #57028: Support inlining variables that are used in object literal shorthand
### 🔎 Search Terms

inline, refactor, refactoring

Not a recession 

### ⏯ Playground Link

_No response_

### 💻 Code

```ts
const abc = 123;
const obj = { abc };
```

Try inlining `abc`

### 🙁 Actual behavior

No refactoring offered

### 🙂 Expected behavior

We should show a `inline` refactoring that results in:

```ts
const obj = { abc: 123 };
```

_No response_

a-tarasyuk:
Fixes #57028

### DIFF
--- a/src/services/refactors/inlineVariable.ts
+++ b/src/services/refactors/inlineVariable.ts
@@ -245,6 +245,10 @@ function getReplacementExpression(reference: Node, replacement: Expression) {
         return factory.createParenthesizedExpression(replacement);
     }
 
+    // Inline shorthand property assignment
+    // E.g.:
+    // const x = 1;
+    // const y = { x }; -> const y = { x: 1 };
     if (isIdentifier(reference) && isShorthandPropertyAssignment(parent)) {
         return factory.createPropertyAssignment(reference, replacement);
     }
--- a/src/services/refactors/inlineVariable.ts
+++ b/src/services/refactors/inlineVariable.ts
@@ -152,7 +152,7 @@ function getInliningInfo(file: SourceFile, startPosition: number, tryWithReferen
     }
 
     // Try finding the declaration and nodes to replace via the reference token.
-    if (tryWithReferenceToken || isShorthandPropertyAssignment(parent)) {
+    if (tryWithReferenceToken) {
         let definition = checker.resolveName(token.text, token, SymbolFlags.Value, /*excludeGlobals*/ false);
         definition = definition && checker.getMergedSymbol(definition);
 
--- a/src/services/refactors/inlineVariable.ts
+++ b/src/services/refactors/inlineVariable.ts
@@ -189,8 +189,9 @@ function isDeclarationExported(declaration: InitializedVariableDeclaration): boo
 function getReferenceNodes(declaration: InitializedVariableDeclaration, checker: TypeChecker, file: SourceFile): Identifier[] | undefined {
     const references: Identifier[] = [];
     const cannotInline = FindAllReferences.Core.eachSymbolReferenceInFile(declaration.name as Identifier, checker, file, ref => {
-        // Only inline if all references are reads. Else we might end up with an invalid scenario like:
-        // const y = x++ + 1 -> const y = 2++ + 1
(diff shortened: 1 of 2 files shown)


DanielRosenwasser (approved):

DanielRosenwasser:
@typescript-bot test tsserver top120

src/services/refactors/inlineVariable.ts
```diff
@@ -152,7 +152,7 @@ function getInliningInfo(file: SourceFile, startPosition: number, tryWithReferen
     }
 
     // Try finding the declaration and nodes to replace via the reference token.
-    if (tryWithReferenceToken) {
+    if (tryWithReferenceToken || isShorthandPropertyAssignment(parent)) {
```
MariaSolOs, line 155: Why is this condition modified? The purpose of `tryWithReferenceToken` is for a lightbulb to show up only in the declaration site, and to require explicit refactor requests in reference sites.

src/services/refactors/inlineVariable.ts
```diff
@@ -191,7 +191,7 @@ function getReferenceNodes(declaration: InitializedVariableDeclaration, checker:
     const cannotInline = FindAllReferences.Core.eachSymbolReferenceInFile(declaration.name as Identifier, checker, file, ref => {
         // Only inline if all references are reads. Else we might end up with an invalid scenario like:
         // const y = x++ + 1 -> const y = 2++ + 1
-        if (FindAllReferences.isWriteAccessForReference(ref)) {
+        if (FindAllReferences.isWriteAccessForReference(ref) && !isShorthandPropertyAssignment(ref.parent)) {
```
MariaSolOs, line 194: If changing this condition, consider updating the comment as well.

src/services/refactors/inlineVariable.ts
a-tarasyuk: Should _shorthand properties_ only be allowed to change with the `invoked` trigger reason?

MariaSolOs: I believe so. Thoughts @DanielRosenwasser?

DanielRosenwasser: That sounds right. The shorthand property is effectively considered a usage. So it should work similarly to the following case

```ts
let x = 123;
let y = {
  x: x/**/
};
```

Unless there's something you're noticing that we're missing @a-tarasyuk

src/services/refactors/inlineVariable.ts
MariaSolOs: Could you please explain why this change was needed? Does `FindAllReferences.isWriteAccessForReference` return true for `x`...

```ts
const x = 0;
const y = { x // ...here? }
```

src/services/refactors/inlineVariable.ts
```diff
@@ -244,5 +245,9 @@ function getReplacementExpression(reference: Node, replacement: Expression): Exp
         return factory.createParenthesizedExpression(replacement);
     }
 
+    if (isIdentifier(reference) && isShorthandPropertyAssignment(parent)) {
```
MariaSolOs, line 248: To keep the nice documentation convention in this file, a comment here would be nice ;)

src/services/refactors/inlineVariable.ts
a-tarasyuk: `ShorthandPropertyAssignment` is considered by `isWriteAccessForReference` as a write access reference.

https://github.com/microsoft/TypeScript/blob/6add96fd18d126d1505de5d96f0df6a86e5f2c21/src/services/findAllReferences.ts#L899-L902

https://github.com/microsoft/TypeScript/blob/6add96fd18d126d1505de5d96f0df6a86e5f2c21/src/services/findAllReferences.ts#L949

MariaSolOs (approved):

PR #57070, opened 2024-01-15
Fix unnecessary non-null assertions caught by newer ts-eslint

jakebailey:
After https://github.com/typescript-eslint/typescript-eslint/pull/8143, typescript-eslint catches even more of these unnecessary non-null assertions. Linting will fail with newer versions until this is merged (so package-lock bumps will be stopped as that checks tests/lint).

### DIFF
--- a/src/services/documentRegistry.ts
+++ b/src/services/documentRegistry.ts
@@ -327,7 +327,7 @@ export function createDocumentRegistryInternal(useCaseSensitiveFileNames?: boole
             // the script snapshot.  If so, update it appropriately.  Otherwise, we can just
             // return it as is.
             if (entry.sourceFile.version !== version) {
-                entry.sourceFile = updateLanguageServiceSourceFile(entry.sourceFile, scriptSnapshot, version, scriptSnapshot.getChangeRange(entry.sourceFile.scriptSnapshot!)); // TODO: GH#18217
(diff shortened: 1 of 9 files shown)


src/services/documentRegistry.ts
```diff
@@ -327,7 +327,7 @@ export function createDocumentRegistryInternal(useCaseSensitiveFileNames?: boole
             // the script snapshot.  If so, update it appropriately.  Otherwise, we can just
             // return it as is.
             if (entry.sourceFile.version !== version) {
-                entry.sourceFile = updateLanguageServiceSourceFile(entry.sourceFile, scriptSnapshot, version, scriptSnapshot.getChangeRange(entry.sourceFile.scriptSnapshot!)); // TODO: GH#18217
+                entry.sourceFile = updateLanguageServiceSourceFile(entry.sourceFile, scriptSnapshot, version, scriptSnapshot.getChangeRange(entry.sourceFile.scriptSnapshot)); // TODO: GH#18217
```
sheetalkamat, line 330:
```suggestion
                entry.sourceFile = updateLanguageServiceSourceFile(entry.sourceFile, scriptSnapshot, version, scriptSnapshot.getChangeRange(entry.sourceFile.scriptSnapshot));
```

sheetalkamat (approved):

src/services/documentRegistry.ts
jakebailey: Yes, thanks!

PR #57076, opened 2024-01-16
Revert "Add numeric constraint to type parameter of mapped types with…

gabritto:
… name type and array constraints (#55774)"

This reverts commit 3f8707d04f99bc0ca5bca55e745de74cadb16885 because it broke test `ramdaToolsNoInfinite2`.


jakebailey (approved):

jakebailey:
We could also just fix the test, if the change is reasonable.

gabritto:
> We could also just fix the test, if the change is reasonable.

There's an actual problem with the PR, not with the test.

Issue #56416, opened 2023-10-25
Code actions  "Move to file" and "Move to new file" are shown everywhere


jrieken:
* open [this file]( https://github.com/microsoft/vscode/blob/109e1f8d8afb754ed31317f79937a44e98d5063b/src/vs/base/common/arrays.ts#L126)
* select `data`
* select light bulb
* :bug: there is entries for "Move to file" and "Move to new file"

<img width="425" alt="Screenshot 2023-10-25 at 08 13 20" src="https://github.com/microsoft/vscode/assets/1794099/ba91dbaf-e161-44f9-a63a-b169199465c7">

mjbvz:
Some ideas:

- Only show the lightbulb (implicit code action) when on declarations
- Maybe still keep showing these actions if refactorings are explicitly requested 
- In the refactoring label, show the name of what is going to be moved. Right now it makes is seem like the loop itself will get moved

jrieken:
> Only show the lightbulb (implicit code action) when on declarations

💯 on that - showing the code action further down is noise and lowers my trust in light bulb being helpful (and not ad-space)

aiday-mar:
HI @sandersn, I would like to have a go at this PR. Could I have a pointer to place in the code where the change should be, or discuss this issue with one of your team members in a meeting? We are currently looking into displaying code actions in the light-bulb menu in VS Code on empty lines too, and currently we are seeing the `Move to file` and `Move to new file` actions appearing very often. I had initially prepared a PR to patch this issue on the VS Code side, but we think ideally this fix should come directly from TypeScript

sandersn:
@aiday-mar I believe @navya9singh and @andrewbranch have worked the most on Move To File.

PR #57080, opened 2024-01-17
Making `Move To File` Action appear less often

aiday-mar:
Fixes https://github.com/microsoft/TypeScript/issues/56416

This work is done in relation to conversation with @navya9singh. From the conversation, the following PR aims to do the following:

Add a condition in getAvailableActions() to decide when you want the refactor to not show up. The strategy to follow for now is to not offer the refactor if both ends of the initial selection range are inside a BlockLike node type (other than the source file).

### DIFF
--- a/tests/cases/fourslash/moveToFile_refactorAvailable2.ts
+++ b/tests/cases/fourslash/moveToFile_refactorAvailable2.ts
@@ -9,7 +9,7 @@
 /////*b*/
 
 goTo.select("a", "b");
-verify.not.refactorAvailable("Move to file",
+verify.refactorAvailable("Move to file",
 /*actionName*/ undefined,
 /*actionDescription*/ undefined,
 /*kind*/ undefined,


src/services/refactors/moveToFile.ts
```diff
@@ -151,6 +152,9 @@ import {
 import {
     registerRefactor,
 } from "../refactorProvider";
+import {
+    isBlockLike,
```
jakebailey, line 156: This indicates that we should be moving this helper somewhere else; refactors typically don't import each other.

DanielRosenwasser: And in fact, you can see that it reorders the refactorings offered because they get registered in a different order now.

src/services/refactors/extractSymbol.ts
```diff
@@ -2221,7 +2221,7 @@ function isExtractableExpression(node: Node): boolean {
     return true;
 }
 
-function isBlockLike(node: Node): node is BlockLike {
+export function isBlockLike(node: Node): node is BlockLike {
```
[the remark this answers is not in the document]
DanielRosenwasser: This is usually either done with an `/** @internal */` comment or by moving it to an all-internal file (like `src/services/utilities.ts`)

src/services/refactors/moveToFile.ts
```diff
@@ -167,6 +171,18 @@ registerRefactor(refactorNameForMoveToFile, {
         if (!interactiveRefactorArguments) {
             return emptyArray;
         }
+        const endPosition = context.endPosition;
+        if (endPosition) {
```
DanielRosenwasser, line 175: I would explicitly test for `undefined` - in the (weird, rare?) case where an editor sends a 0-length span at position 0, you won't get a refactoring offered here.

DanielRosenwasser (changes requested):
We generally expect a test to exhibit new behavior. Language service tests should be placed in `tests/cases/fourslash/` and you can follow similar schemes.

src/services/refactors/extractSymbol.ts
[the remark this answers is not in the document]
aiday-mar: Thank you for the review comment

src/services/refactors/moveToFile.ts
aiday-mar: Right thank you for picking up on it, 0 is also falsy

aiday-mar:
Hi thank you for the review comments. I have been working on the changes, and was looking at adding a test. More specifically I talked to @navya9singh about how tests work in the TypeScript repo and she recommended to add a test of the following type:

```
/// <reference path='fourslash.ts' />

////function f3(x, /*e*//*f*/{ y }) {
////    return { x, y: 1 };
////}

goTo.select("e", "f");
verify.not.refactorAvailable("Move to file");
```

I was debugging this test and I noticed that due to the current structure of the code, the `move to` refactoring is never made available in a testing context. More specifically by tracking the method `refactorAvailable`, it leads to `verifyRefactorAvailable`, which ultimately leads to `getApplicableRefactors` where the boolean `includeInteractiveActions` is undefined. Now because this boolean is undefined, later the method `getAvailableActions` returns an empty array through an early return statement independently of the block conditions later. In this manner, I believe, this test may not be properly testing the new behavior added in this PR. I was therefore wondering if you believe this test is good, and if not, which method I could use to write the new test? Thank you.

navya9singh:
> Hi thank you for the review comments. I have been working on the changes, and was looking at adding a test. More specifically I talked to one of your colleagues about how tests work in the TypeScript repo and this person recommended to add a test of the following type:
> 
> ```
> /// <reference path='fourslash.ts' />
> 
> ////function f3(x, /*e*//*f*/{ y }) {
> ////    return { x, y: 1 };
> ////}
> 
> goTo.select("e", "f");
> verify.not.refactorAvailable("Move to file");
> ```
> 
> I was debugging this test and I noticed that due to the current structure of the code, the `move to` refactoring is never made available in a testing context. More specifically by tracking the method `refactorAvailable`, it leads to `verifyRefactorAvailable`, which ultimately leads to `getApplicableRefactors` where the boolean `includeInteractiveActions` is undefined. Now because this boolean is undefined, later the method `getAvailableActions` returns an empty array through an early return statement independently of the block conditions later. In this manner, I believe, this test may perhaps not be testing the behavior introduced in this PR. I was therefore wondering if you believe this test is good, and if not, which method I could use to write the new test? Thank you.

You're right about the interactiveActions always being undefined in this test case and so `getAvailableActions` always returning an empty array. Looking at this I think that we need to add a function called moveToFileVerifyAction or something similar under `verifyNegatable` class in fourslash.ts to actually be able to test if this action should be available or not. Does that sound okay @andrewbranch

andrewbranch:
Is there any reason not to have `verifyRefactorAvailable` always pass `includeInteractiveActions`?

aiday-mar:
So if that suits everyone, could we just set the boolean `includeInteractiveActions` to true?

aiday-mar:
Seeing as there is a thumbs up on my last message I will make the boolean true. I am trying to construct a test that will make the condition `startNode.kind !== SyntaxKind.SourceFile && isBlockLike(startNode)` true. I have written the following test, which I though would make the condition evaluate to true, since the selection start is inside of a `module` keyword.

```
/// <reference path='fourslash.ts' />

////modu/*e*/le My.App {
////    export var appModule = angular.module("app", [
////    ]).config([() => {
////        configureStates/*1*/($stateProvider);
////    }]).run(My.App.setup);
////}/*f*/

goTo.select("e", "f");
verify.not.refactorAvailable("Move to file");
```

Upon testing however the token at the start of the selection was evaluated to be kind `ModuleKeyword` not `ModuleBlock` or `Block`. I tried several other examples, aiming to make the `isBlockLike` method return true. I was wondering if you have an example of a test with a given selection where the selection start node would be evaluated to be a Block node.

navya9singh:
I looked at extractSymbol for some reference and one simple example for this case would be :
```
 namespace ns {
    /*a*/export function fn() {

    }
    fn();
    /*b*/
}
```
In here, I believe you will have to change `startNode.kind !== SyntaxKind.SourceFile && isBlockLike(startNode)` condition to instead be a loop that starts at the parent of the startNode and traverses up the tree to check if the corresponding node is a blockLike node. Same will have to be done for the endNode, and if both conditions are true, then Move to file should not show up.

navya9singh (changes requested):

aiday-mar:
Hi @navya9singh thank you for the review comments. I made the changes requested.

aiday-mar:
Hi @andrewbranch @navya9singh if the code looks good, would it be possible to receive an approval for merging the code? On the VS Code side, we would like to get this PR merged by the end of this week.

src/services/refactors/moveToFile.ts
```diff
@@ -167,6 +170,14 @@ registerRefactor(refactorNameForMoveToFile, {
         if (!interactiveRefactorArguments) {
             return emptyArray;
         }
+        if (context.endPosition !== undefined) {
+            const file = context.file;
+            const startNodeAncestor = getBlockLikeAncestor(getTokenAtPosition(file, context.startPosition));
```
andrewbranch, line 175: There’s an existing utility for this. `findAncestor(token, isBlockLike)`

src/services/refactors/moveToFile.ts
```diff
@@ -167,6 +170,14 @@ registerRefactor(refactorNameForMoveToFile, {
         if (!interactiveRefactorArguments) {
             return emptyArray;
         }
+        if (context.endPosition !== undefined) {
+            const file = context.file;
+            const startNodeAncestor = getBlockLikeAncestor(getTokenAtPosition(file, context.startPosition));
+            const endNodeAncestor = getBlockLikeAncestor(getTokenAtPosition(file, context.endPosition));
+            if (!isSourceFile(startNodeAncestor) && !isSourceFile(endNodeAncestor)) {
+                return emptyArray;
```
andrewbranch, line 178: It would be good to leave a comment explaining in human terms exactly what the condition is and why it was implemented. I'm pretty sure I suggested this method of implementation, but I already don't remember what it was and it's a little hard to make the leap from code to intuition.

tests/cases/fourslash/moveToFile_refactorAvailable1.ts
```diff
@@ -0,0 +1,18 @@
...
+////namespace ns {
+////    /*a*/export function fn() {
+////    }
+////    fn();
+////    /*b*/
+////}
+
+goTo.select("a", "b");
+verify.not.refactorAvailable("Move to file",
+undefined,
```
andrewbranch, line 12: Please label the undefineds and trues with a comment like `/*includeInteractiveRefactors*/ true)`. This is a lint rule enforced in source code, but we can't really lint test content. But it still helps future readers understand what's being passed to this function.

andrewbranch (commented):
Looks good overall, just a few nits. Thanks!

tests/cases/fourslash/moveToFile_refactorAvailable1.ts
sheetalkamat: on that topic i think we should lint fourslash and fourslash-server tests as they contain code rather than "test case" as a code

aiday-mar:
Hi thank you for all your review comments, I have made the latest changes requested.

navya9singh (approved):

andrewbranch (approved):

mu578:
Don't you have a bitwise operator in typescript instead of switch casing on enum values? You already have a terrible coding style with your camel casing everywhere, it would need some fresh air.

DanielRosenwasser:
@mu578 I've never seen you on this repository, but here you must adhere to the [code of conduct](https://github.com/microsoft/TypeScript/blob/60f93aa83ae644092ace6d729d0f10c42715292f/CODE_OF_CONDUCT.md). You must keep feedback respectful and objective.

PR #50034, opened 2022-07-25
Make `any[]` the effective rest type when `any` is spread

Andarist:
~I think this might be a prerequisite to fixing the issue https://github.com/microsoft/TypeScript/issues/29919~ fixes https://github.com/microsoft/TypeScript/issues/29919

Why do I think this might be required?
1. the problem with that issue is that `getResolvedApparentTypeOfMappedType` for `MappedType<Parameters<FnType>>` is not recognized as an array type
2. but we know that it has to be - mapping over an array type (using a homomorphic mapped type) produces an array type and we know that `Parameters<FnType>` has to produce an array type
3. with the current logic the computed `constraint` [here](https://github.dev/microsoft/TypeScript/blob/a179e910dd818262571b46313a7918aacf67978a/src/compiler/checker.ts#L12327) is `any` and thus it doesn't pass `isArrayOrTupleType` check, the mapped type stays non-instantiated and later on can't be used as a type for the rest parameter
4. therefore it, IMHO, makes sense to "propagate" the information from `Parameters` to the mapped type in the easiest way possible - by changing the final constraint of `...T` to `any[]` when `T extends any`.

~There are still some changes to be made to the code for this to be part of the fix. I'm still struggling to tweak the logic in `getResolvedApparentTypeOfMappedType` to instantiate the mapped type using this newly adjusted constraint but that's, sort of, a problem for another day. I'd like to check with the team first if I'm on the right path and if this change makes any sense.~ This has been done in https://github.com/microsoft/TypeScript/pull/56727

PR #57091, opened 2024-01-18
Resolve apparent mapped types recursively

Andarist:
this expands on https://github.com/microsoft/TypeScript/pull/56727
I'm opening this as a draft but I'd love it if somebody could run an extended test suite on this PR in the meantime (cc @gabritto )

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -14558,9 +14558,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 constraint = getConstraintOfTypeParameter(typeVariable);
             }
             else {
-                const modifiersConstraint = getConstraintOfType(getModifiersTypeFromMappedType(type));
-                if (modifiersConstraint) {
-                    constraint = getApparentType(modifiersConstraint);
+                const modifiersType = getModifiersTypeFromMappedType(type);
+                if (modifiersType) {
+                    constraint = getApparentType(getConstraintOfType(modifiersType) || modifiersType);
                 }
             }
             if (constraint && everyType(constraint, isArrayOrTupleType)) {


src/compiler/checker.ts
```diff
@@ -14558,9 +14558,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 constraint = getConstraintOfTypeParameter(typeVariable);
             }
             else {
-                const modifiersConstraint = getConstraintOfType(getModifiersTypeFromMappedType(type));
-                if (modifiersConstraint) {
-                    constraint = getApparentType(modifiersConstraint);
+                const modifiersType = getModifiersTypeFromMappedType(type);
+                if (modifiersType) {
+                    constraint = getApparentType(getConstraintOfType(modifiersType) || modifiersType);
```
Andarist, line 14563: I'm not particularly fond of this - maybe there is a better way of doing this. This mainly tries to get the apparent type of the modifiers type when the modifiers type is a mapped type itself (like in `Mapped<Mapped2<Foo>>`). Mapped types don't return anything from `getConstraintOfType`.

Andarist:
The test failure doesn't surprise me. It would pass if another problem would get fixed - the one that is being tackled by https://github.com/microsoft/TypeScript/pull/50034

The recursive resolution here just uncovers the fact that this other thing doesn't exactly work properly - as far as I was able to investigate so far and I spent 2 nights on this.

gabritto:
@typescript-bot run DT
@typescript-bot test top100

Andarist:
I pushed out a test case that shows how this is an improvement. It fixes the VS Code's build failure that @jakebailey reported post-merge [here](https://github.com/microsoft/TypeScript/pull/56727#issuecomment-1886338688). 

It's obviously not perfect right now because `ramdaToolsNoInfinite` test case fails. As mentioned above, that would get fixed with https://github.com/microsoft/TypeScript/pull/50034 (or an alternative fix to the underlying issue).

Andarist:
superseded by https://github.com/microsoft/TypeScript/pull/57122

PR #57092, opened 2024-01-19
Revert "Fixed apparent type of homomorphic mapped type with non-homomorphic instantiation (#56727)" and add test

gabritto:
#56727 breaks @types/sinon version 10: https://github.com/microsoft/TypeScript/pull/56727#issuecomment-1886338688
This PR reverts that one, and adds a test with a simplified version of the code that broke (test thanks to @Andarist).


jakebailey (approved):

tests/cases/compiler/homomorphicMappedTypeWithNonHomomorphicInstantiationSpreadable1.ts
jakebailey, line 1: Perhaps leave this test in?

Andarist: Isn't it confusing to have tests that are kinda meant to pass but they fail?

ahejlsberg (approved):

tests/cases/compiler/homomorphicMappedTypeWithNonHomomorphicInstantiationSpreadable1.ts
DanielRosenwasser: Well the fact that the test was added meant that the behavior wasn't tested in the first place. Capturing the behavior and changes over time is useful.

gabritto: I'll remember to leave the tests in the future.

PR #57101, opened 2024-01-19
chore: deprecate Node's getFullStart()

### LINKED ISSUE #57053: Document and/or deprecate Node's getFullStart() vs. pos?
### Acknowledgement

### Comment

Coming over from https://github.com/typescript-eslint/typescript-eslint/issues/8227 -> https://discord.com/channels/508357248330760243/640177429775777792/1194002057858916483: it looks like `ts.Node` objects have both a `pos: number` property and a `getFullStart(): number` method that directly returns `this.pos`:

https://github.com/microsoft/TypeScript/blob/02f9ddf55d27716c0306ec6561baf80d42744fed/src/services/services.ts#L564-L566

Doing some quick _Find all references_ from the interfaces' definitions in VS Code:

* `Node`'s `getFullStart()` has 25 results
* `ReadonlyTextRange`s `readonly pos` has 407 results

I looked through the GitHub repo and code history but couldn't find anything justifying why both exist. My hunch is maybe that #9529 which added `getFullStart()` predates TS 2.0's `readonly` ([TS 2.0 release notes > readonly](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#read-only-properties-and-index-signatures))?

Given that `getFullStart()` incurs an extra function call, is there any reason beyond backwards compatibility to keep it around? If not, would the team be open to marking it as `@deprecated` in JSDoc to reduce ambiguity?

JoshuaKGoldberg:
Fixes #57053.

Marks the method as `@deprecated` in the `Node` interface and `TokenOrIdentifierObject` implementing class. Then replaces all find-all-references results from either to `.pos`.

Does not add the deprecation notice to the `NodeObject` class, since its `getFullStart()` also does a `this.assertHasRealPosition()`. Should that also be marked as deprecated?

### DIFF
--- a/src/services/callHierarchy.ts
+++ b/src/services/callHierarchy.ts
@@ -424,7 +424,7 @@ export function createCallHierarchyItem(program: Program, node: CallHierarchyDec
     const containerName = getCallHierarchItemContainerName(node);
     const kind = getNodeKind(node);
     const kindModifiers = getNodeModifiers(node);
-    const span = createTextSpanFromBounds(skipTrivia(sourceFile.text, node.getFullStart(), /*stopAfterLineBreak*/ false, /*stopAtComments*/ true), node.getEnd());
+    const span = createTextSpanFromBounds(skipTrivia(sourceFile.text, node.pos, /*stopAfterLineBreak*/ false, /*stopAtComments*/ true), node.getEnd());
     const selectionSpan = createTextSpanFromBounds(name.pos, name.end);
     return { file: sourceFile.fileName, kind, kindModifiers, name: name.text, containerName, span, selectionSpan };
 }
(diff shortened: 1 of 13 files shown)


jakebailey:
> Does not add the deprecation notice to the `NodeObject` class, since its `getFullStart()` also does a `this.assertHasRealPosition()`. Should that also be marked as deprecated?

Hm, this is sort of a good point. This is a message that we have seen before: https://github.com/microsoft/TypeScript/issues?q=is%3Aissue+sort%3Aupdated-desc+%22assertHasRealPosition%22

Maybe this is an indicator that we should keep it, if we're catching buggy behavior?

DanielRosenwasser:

Issue #57103, opened 2024-01-20
`this is` return type of method isn't checked when implementing


KermanX:
### 🔎 Search Terms

"this is type", "return type", "implement", "inherit", "not checked"

### 🕗 Version & Regression Information

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about "this is"

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.3.3#code/JYOwLgpgTgZghgYwgAgILIN4ChnILYRgAWA9gCYCMAFAJQBcyxwAzsi5jrsnAxZwL6cCxcgCZaDAEYkSAGwhwQAbiyCEsuM1YAhNngAO8guFbpsuAPQXkAdSIBPZCBLJoUElAD8QwqUq1MZEFOK2QAATBmAFoIAA99CAQwGKh3KGQAFXsE5AByADcSYDJctlZnMG4tYABzEDhJeUYXMGyUXOk5BRBcnxEycRpAwWDWnIzkAF5kACVCAFcoECyEgB4ASRBmMEUkFYhVsYgSGGRtAD4AbQAiYT8Ka4Bdc6wgA

### 💻 Code

```ts
interface A {
  method1(): this is {
    a: 1
  };
  method2(): boolean;
}
class B implements A {
  // Why no error?
  method1() { }

  // @ts-expect-error Type 'void' is not assignable to type 'boolean'
  method2() { }
}
```

### 🙁 Actual behavior

`B.method1` produces no error, and its return type is `void`.

### 🙂 Expected behavior

`B.method1` should be an error.

### Additional information about the issue

If `A` in the above example is a class, and `B` is extending `A`, this bug will also happen.

PR #57104, opened 2024-01-20
Update tsconfig.json

starter-asd:
issue #57103


jakebailey:
I'm not sure of your intention here, but I don't think this change is valid; these scripts must not be emitted.

Issue #57108, opened 2024-01-20
Quickinfo displays `any` in TS files within JSDoc comments


Andarist:
### 🔎 Search Terms

quickinfo jsdoc ts any 

### 🕗 Version & Regression Information

- This is the behavior in every version I tried

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.4.0-dev.20240120#code/PQKhAIAEBcE8AcCm4DeAKAlOAvAPleACYCG0xAXOAM7QBOAlgHYDmA2gLrgC+X4IwAKABmAV0YBjaPQD2jcNEQ1MlFEVIVqdJm068UA8OFqJoI2nP2HDJMpQ4AaA9wDcArgIGgIkY6fNVUVRsNGgYWDm5efmExSRk5BRoAJmUCYMpQ7Qi9J18zCydrdTt2R0MuV3dPMCg4JEDwACNiWgytFmdIvkFxWRp5RWgAZhwCZtbwAHJJ7g8gA

### 💻 Code

```ts
/** @type {() => { data: string[] }} */
function test(): { data: string[] } {
  return {
    data: [],
  };
}

/** @returns {{ data: string[] }} */
function test2(): { data: string[] } {
  return {
    data: [],
  };
}

/** @type {{ bar: string; }} */
const test3 = { bar: '' }
```

### 🙁 Actual behavior

When hovering over any of those properties in the JSDoc comments we get a useless quickinfo mentioning `any`

### 🙂 Expected behavior

I'd expect those quickinfos to not be displayed at all in TS files

fatcerberus:
That's super weird since I thought TS just didn't parse JSDoc type info at all in `.ts` files, but clearly it's done enough work to know `data` and `bar` are things that should have types.

Issue #57094, opened 2024-01-19
GetReferencesAtPosition in ObjectBindingPattern on BindingName fails


JoviDeCroock:
### 🔎 Search Terms

LSP Plugin, TSServer, ObjectBindingPattern, GetReferencesAtPosition 

### 🕗 Version & Regression Information

- This is a crash

### ⏯ Playground Link

https://github.com/0no-co/GraphQLSP/issues/187#issuecomment-1899544136

### 💻 Code

The following happens on the `data` binding-name of the second snippet

```ts
  const references = info.languageService.getReferencesAtPosition(
    source.fileName,
    node.getStart()
  );
```

```ts
 const { data, loading, error, refetch } = useQuery(Query)
```

We follow the reference of Query to the invocation, see it's an object-destructuring and follow all the binding-names to collect access patterns. I have noticed before that we had to use `node.getStart()` rather than `node.pos` else TS would not find anything, however this is just a straight up crash 😅 

### 🙁 Actual behavior

Crashes within tsserver on 

```
    TypeError: Cannot read properties of undefined (reading 'escapedName')
        at getTypeForVariableLikeDeclaration (/home/banks/Development/beep/node_modules/typescript/lib/tsserver.js:56154:39)
```

### 🙂 Expected behavior

This to follow the reference correctly.

### Additional information about the issue

I am terribly sorry that this is on a real project, I could not find good ways to reproduce issues with LSP plugins yet...

Andarist:
The problem is that a symbol is read from a declaration in JSDoc function type and those don't have symbols and it leads to a crash. I'm still working on reproducing this in isolation.

Andarist:
test case:
```ts
/// <reference path='fourslash.ts' />

// @moduleResolution: node

// @Filename: node_modules/use-query/package.json
////{
////  "name": "use-query",
////  "types": "index.d.ts"
////}
// @Filename: node_modules/use-query/index.d.ts
////declare function useQuery(): {
////  data: string[];
////};

// @Filename: node_modules/use-query/package.json
////{
////  "name": "other",
////  "types": "index.d.ts"
////}
// @Filename: node_modules/other/index.d.ts
////interface BottomSheetModalProps {
////  /**
////   * A scrollable node or normal view.
////   * @type null | (({ data: any }?) => any)
////   */
////  children: null | ({ data: any }?) => any;
////}

// @Filename: src/index.ts
////import { useQuery } from "use-query";
////const { /*1*/data } = useQuery();

[...]

PR #57110, opened 2024-01-20
Fixed crashes when looking up symbols of jsdoc nodes in TS files

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/57094
closes https://github.com/microsoft/TypeScript/issues/57108


src/services/findAllReferences.ts
```diff
@@ -1837,6 +1837,9 @@ export namespace Core {
                 }
                 // falls through I guess
             case SyntaxKind.Identifier:
+                if (node.flags & NodeFlags.JSDoc && !isInJSFile(node)) {
```
Andarist, line 1840: I need to additionally check if this is in an "any type tag" as those should be ignored whereas nodes in semantic~ jsdoc tags shouldn't (like in `@see`, `@link`, etc). I couldn't find any utility that would check something like this - I could create such and manually curate the list of those tags in it.

But there is another question: should those references in type-like tags always be ignored? Or should TS try to resolve corresponding TS types? There is always a risk that those won't match but in such a case displaying `any` (likely as a result of displaying `errorType`) could actually be OK.

An alternative fix for the crashes would be to ignore symbol-less declarations for now but I intentionally didn't go that route. It seemed like a bandaid for a deeper problem and I think handling this at a higher level would be better. I'm just not sure now what's the exact desired behavior.

cc @sandersn

tests/baselines/reference/myFindAllRefs.baseline.jsonc
```diff
@@ -0,0 +1,45 @@
+// === findAllReferences ===
```
sandersn, line 1: I think this baseline is left over from a previous test that isn't in this PR.

tests/cases/fourslash/findReferencesBindingPatternInJsdocNoCrash2.ts
```diff
@@ -0,0 +1,33 @@
...
+// @Filename: node_modules/use-query/package.json
+////{
+////  "name": "other",
+////  "types": "index.d.ts"
+////}
+// @Filename: node_modules/other/index.d.ts
+////interface BottomSheetModalProps {
+////  /**
+////   * A scrollable node or normal view.
+////   * @type null | (({ data: any }?) => any)
```
sandersn, line 24: is `null |` the only difference between NoCrash2 and 1? If so, I don't see what value it adds above NoCrash1, especially since the new code doesn't reference Typescript types really.

src/services/findAllReferences.ts
sandersn: I think you're asking whether find-all-refs should skip the references in `@see` et al? Is that right? Ideally the answer is No. Symbol resolution for those tags works outside the normal name resolution code, as I recall, via resolveJSDocMemberName. That function is called from getSymbolAtLocation, the usual entrypoint for public users of the API. You might want to call one of those two functions.

A utility to identify tags with semantics in Typescript would be useful, and I don't think it exists yet. scanner.ts does have a regex, but that's not useful for a parse tree. (and, besides, `@deprecated` should be included in the post-parser predicate, but not the scanner one.)

Please add a test either way.

sandersn (changes requested):

tests/cases/fourslash/findReferencesBindingPatternInJsdocNoCrash1.ts
```diff
@@ -0,0 +1,33 @@
...
+////{
+////  "name": "use-query",
+////  "types": "index.d.ts"
+////}
+// @Filename: node_modules/use-query/index.d.ts
+////declare function useQuery(): {
+////  data: string[];
+////};
+
+// @Filename: node_modules/use-query/package.json
```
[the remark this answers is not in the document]
Andarist: yes, it should - the package split shouldn't be required for this to be reproduced at all, it was just faster for me to keep the package structure as-is (like in the original real-world code)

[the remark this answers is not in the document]
Andarist: If u feel strongly about it, I can simplify this further or add an extra test case. I need to refresh myself on the whole situation (which I'll do soon since I plan to finish this PR) - but the file split itself might have been important for this (I'm not 100% sure about it right now though)

tests/cases/fourslash/findReferencesBindingPatternInJsdocNoCrash2.ts
Andarist: This test case exercises a braceless `@type` which was somewhat important when I started to reduce this and for some moments I couldn't figure out why the crash went away on certain reduction attempts. So I included that here to be extra sure that it's gonna be fixed in such a scenario too

src/compiler/checker.ts
```diff
@@ -11027,6 +11027,10 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         }
 
         if (isParameter(declaration)) {
+            if (!declaration.symbol) {
```
Andarist, line 11030: I decided to include this here because it's just always possible to call `checker.getTypeAtLocation` or `checker.getTypeAtLocation` and those shouldn't crash or do something overly goofy - so the safest bet is to actually be on the safe side of things here + fine-tune existing call sites to those 2 if it's absolutely necessary to avoid getting `errorType` back when it's obvious that it will be returned back

src/services/services.ts
```diff
@@ -2110,6 +2111,14 @@ export function createLanguageService(
...
         switch (node.kind) {
             case SyntaxKind.Identifier:
+                if (
+                    node.flags & NodeFlags.JSDoc && !isInJSFile(node) &&
+                    ((node.parent.kind === SyntaxKind.PropertySignature && (node.parent as PropertySignature).name === node) ||
+                        findAncestor(node, (n) => n.kind === SyntaxKind.Parameter))
+                    ) {
+                    // if we'd request type at those locations we'd get `errorType` that displays confusingly as `any`
+                    return false;
+                }
```
Andarist, line 2121: I don't like how specific this is but I concluded that some condition like this is the best solution for https://github.com/microsoft/TypeScript/issues/57108 . I'm not 100% sure if I'm missing some other node types that should be ignored here or not though. The worst case scenario quick info will continue to display `any` when it shouldn't so I feel it's fine. If you come up with any extra situations that should be handled here I can tweak this and add more tests

Andarist: Note that it's still good to display `any` [here](https://github.com/microsoft/TypeScript/blob/353ccb7688351ae33ccf6e0acb913aa30621eaf4/tests/cases/fourslash/jsDocFunctionSignatures13.ts#L10) - and that's why I have to prevent the type request conditionally instead of just ignoring return values that are error types or smth like that

tests/cases/fourslash/quickInfoInJsdocInTsFile1.ts
```diff
@@ -0,0 +1,62 @@
...
+////   return {
+////     data: [],
+////   };
+//// }
+////
+//// /** @type {{ /*3*/bar: string; }} */
+//// const test3 = { bar: '' };
+////
+//// type SomeObj = { bar: string; };
+//// /** @type {SomeObj/*4*/} */
```
Andarist, line 21: In TS files this is just ignored, in the sense that it doesn't impact types at all. So technically we could just avoid displaying anything here. I concluded though that it would actually be less useful and that it would be confusing - usually the user expects some feedback in form of a tooltip and not returning any info here could feel like something is off, like it's broken.

Since this isn't used for type-checking there is no guarantee that it "matches" the declaration below - I think it's an acceptable tradeoff. If somebody is writing JSDoc types in TS files then it's their responsibility to make them match.

Andarist: cc @jakebailey since we were talking about this a few hours ago

sandersn (approved):

sandersn:
Can you merge from main before I merge this? We've had a lot of baseline churn recently.

jakebailey:
We'll still not done, forgot one PR is left...

Andarist:
@sandersn done

Issue #57109, opened 2024-01-20
Literal tested as key of a certain type cannot be used as an index type


ghostinpeace:
### 🔎 Search Terms

"literal", "extends", "keyof", "cannot be used to index type", "conditional"

### 🕗 Version & Regression Information

Versions: ^5.0 (regression)

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.3.3&ssl=7&ssc=54&pln=1&pc=1#code/C4TwDgpgBAKlC8UDeBYAUFKBrAXFAziALYBGA9gDYA06mZwAFhAE57mUQCGAdlVAPT8oPek2bYIIAhAhF8UZhACOAVwCWigCZQAZmXGNoJFQHMowMsLCROzdAF906QbHDQA5ACIsn91ADGPNz0UCTQKvgQ2pzyPFBq3JoQAB7mbgB0ABQATACsAMwAHACU6KCQUADCCFDuWH4pwBCJ8liSZDqwUAD8sADade4AulB43BAAbiwA3AJCPCBAA

### 💻 Code

```ts
type T = {
  k: symbol,
  other: boolean, // another key seems required for the bug to appear
}

// Type '"k"' cannot be used as an index type.(2538)
type C = 'k' extends keyof T ? T['k'] : never; // any
```

### 🙁 Actual behavior

An error is produced in the positive clause of the conditional type.
The resulting type is incorrect.

### 🙂 Expected behavior

No error.

craigphicks:
Weird.  This works
```
type T = {
  k: symbol,
  other: boolean, // another key seems required for the bug to appear
}
type C<U=T> = 'k' extends keyof U ? U['k'] : never;
type X = C; // symbol
```

PR #57113, opened 2024-01-21
Avoid creating redundant substitution types in non-generic contexts

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/57109

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -16173,7 +16173,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     }
 
     function getSubstitutionType(baseType: Type, constraint: Type) {
-        return constraint.flags & TypeFlags.AnyOrUnknown || constraint === baseType || baseType.flags & TypeFlags.Any ?
+        return constraint.flags & TypeFlags.AnyOrUnknown || constraint === baseType || baseType.flags & TypeFlags.Any || !isGenericType(constraint) && !isGenericType(baseType) && isTypeAssignableTo(baseType, constraint) ?
             baseType :
             getOrCreateSubstitutionType(baseType, constraint);
     }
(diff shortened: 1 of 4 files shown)


src/compiler/checker.ts
```diff
@@ -16173,7 +16173,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     }
 
     function getSubstitutionType(baseType: Type, constraint: Type) {
-        return constraint.flags & TypeFlags.AnyOrUnknown || constraint === baseType || baseType.flags & TypeFlags.Any ?
+        return constraint.flags & TypeFlags.AnyOrUnknown || constraint === baseType || baseType.flags & TypeFlags.Any || !isGenericType(constraint) && !isGenericType(baseType) && isTypeAssignableTo(baseType, constraint) ?
```
Andarist, line 16176: alternatively, substitution types could get some improved handling in `getPropertyTypeForIndexType`

weswigham: Yeah, I think I'd prefer `getPropertyNameFromIndex` + `getPropertyTypeForIndexType` being updated to handle substitutions - the more places we recognize them as transparently their original type (if the substituted version isn't useful), the better, since it makes it easier on us to broaden the scope of how we use substitution types to support other features.

As far as this approach goes... `instantiateTypeWorker` has similar logic to this - honestly, a lot of that logic should *probably* get moved into here, so it happens on initial construction *and* instantiation, rather than just instantiation.

ahejlsberg: I agree, the right approach is to erase them where needed. Omitting the substitution type just because the constraint is assignable isn't right and will definitely have adverse effects as the constraint may be a more specific type.

Andarist: I pushed out the requested change. I decided to use a similar approach of unwrapping the substation type early in `getPropertyTypeForIndexType` like it was done in https://github.com/microsoft/TypeScript/pull/58292

jakebailey:
@typescript-bot test it

PR #43097, opened 2021-03-05
Uncalled function checks don't work with negation

jonhue:
Fixes #43096

PR #57114, opened 2024-01-21
Uncalled function checks check negation condition

### LINKED ISSUE #34815: Uncalled Function Checks don't check negative condition
```ts
function foo() {
}

if (!foo) {
    let x = 10
}
else {
}
```

**Expected**: Error on `!foo` that the condition is always true because `foo` is defined.
**Actual**: No error.

Also, any work on this should include the following test:

```ts
function foo() {

}

if (!foo) {
}
else {
    foo()
}
```

### LINKED ISSUE #43096: Uncalled function checks don't work with negation
# Bug Report

### 🔎 Search Terms

`uncalled function checks`, `negation`

### ⏯ Playground Link

[Playground link with relevant code](https://www.typescriptlang.org/play?ssl=16&ssc=2&pln=1&pc=1#code/PTAEAEGcBcCcEsDG0ByBXANhgwgCwKaIDWkAXKHGvgFDUAmhGAhrPqAGZoB2y8A9l1DxIAMT58AFAEpyAI3EZ8TLgG56jFm0QCYQyAFUuDdvC746YvuQnTQAXgB8oeX0XKpoAD6hux0+bVqeHZQCWFLDwBvalBY0BAfHiYscw5uXgFQRAJiSFBk1iY6AE9QAHc+WCJy-CxQAlZqAF9aYNCAQnDxKJi4hPTkxTo0nmh+QWzCElBIXD5MYe0AN3xYClxhLKZINmhxZtaQiU6DI3wTMwtu0Gi4+LBoDbzEbbZZ+Yxhrj5oZy0+FasOgHIA)

### 💻 Code

```ts
declare function isFoo(): boolean;
declare const isUndefinedFoo: (() => boolean) | undefined;

if (isFoo) {
    // uncalled function checks already work well here
}

if (!isFoo) {
    // uncalled function checks should cover this case too
}

if (!isUndefinedFoo) {
    // this case should not be covered
}
```

### 🙁 Actual behavior

Uncalled function checks do not work with the `!` operator.

### 🙂 Expected behavior

Unless a function may be undefined it would be very helpful to check whether it is called when it is used with the `!` operator.

I created this issue to separate a part of the implementation of #42835 resulting from issue #35584. This specific issue was inspired by https://github.com/microsoft/TypeScript/issues/35584#issuecomment-769401594.

AlexMarkUSDS:
Fixes #34815
Fixes #43096

This PR is partially a revival of https://github.com/microsoft/TypeScript/pull/43097, which had gone stale and was closed. Thank you @jonhue for pointing me off in the right direction.

### DIFF
--- a/src/compiler/moduleSpecifiers.ts
+++ b/src/compiler/moduleSpecifiers.ts
@@ -996,7 +996,7 @@ function tryGetModuleNameFromRootDirs(rootDirs: readonly string[], moduleFileNam
 }
 
 function tryGetModuleNameAsNodeModule({ path, isRedirect }: ModulePath, { getCanonicalFileName, canonicalSourceDirectory }: Info, importingSourceFile: SourceFile, host: ModuleSpecifierResolutionHost, options: CompilerOptions, userPreferences: UserPreferences, packageNameOnly?: boolean, overrideMode?: ResolutionMode): string | undefined {
-    if (host.fileExists === undefined || !host.readFile) {
+    if (!host.readFile) {
         return undefined;
     }
     const parts: NodeModulePathParts = getNodeModulePathParts(path)!;
@@ -1145,7 +1145,6 @@ function tryGetModuleNameAsNodeModule({ path, isRedirect }: ModulePath, { getCan
 }
 
 function tryGetAnyFileFromPath(host: ModuleSpecifierResolutionHost, path: string) {
-    if (host.fileExists === undefined) return;
     // We check all js, `node` and `json` extensions in addition to TS, since node module resolution would also choose those over the directory
     const extensions = flatten(getSupportedExtensions({ allowJs: true }, [{ extension: "node", isMixedContent: false }, { extension: "json", isMixedContent: false, scriptKind: ScriptKind.JSON }]));
     for (const e of extensions) {
--- a/src/compiler/moduleSpecifiers.ts
+++ b/src/compiler/moduleSpecifiers.ts
@@ -996,7 +996,7 @@ function tryGetModuleNameFromRootDirs(rootDirs: readonly string[], moduleFileNam
 }
 
 function tryGetModuleNameAsNodeModule({ path, isRedirect }: ModulePath, { getCanonicalFileName, canonicalSourceDirectory }: Info, importingSourceFile: SourceFile, host: ModuleSpecifierResolutionHost, options: CompilerOptions, userPreferences: UserPreferences, packageNameOnly?: boolean, overrideMode?: ResolutionMode): string | undefined {
-    if (!host.fileExists || !host.readFile) {
+    if (host.fileExists === undefined || !host.readFile) {
         return undefined;
     }
     const parts: NodeModulePathParts = getNodeModulePathParts(path)!;
@@ -1145,7 +1145,7 @@ function tryGetModuleNameAsNodeModule({ path, isRedirect }: ModulePath, { getCan
 }
 
 function tryGetAnyFileFromPath(host: ModuleSpecifierResolutionHost, path: string) {
-    if (!host.fileExists) return;
+    if (host.fileExists === undefined) return;
     // We check all js, `node` and `json` extensions in addition to TS, since node module resolution would also choose those over the directory
     const extensions = flatten(getSupportedExtensions({ allowJs: true }, [{ extension: "node", isMixedContent: false }, { extension: "json", isMixedContent: false, scriptKind: ScriptKind.JSON }]));
     for (const e of extensions) {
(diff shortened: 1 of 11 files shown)


AlexMarkUSDS:

src/compiler/moduleSpecifiers.ts
```diff
@@ -996,7 +996,7 @@ function tryGetModuleNameFromRootDirs(rootDirs: readonly string[], moduleFileNam
 }
 
 function tryGetModuleNameAsNodeModule({ path, isRedirect }: ModulePath, { getCanonicalFileName, canonicalSourceDirectory }: Info, importingSourceFile: SourceFile, host: ModuleSpecifierResolutionHost, options: CompilerOptions, userPreferences: UserPreferences, packageNameOnly?: boolean, overrideMode?: ResolutionMode): string | undefined {
-    if (!host.fileExists || !host.readFile) {
+    if (!host.readFile) {
```
AlexMarkUSDS, line 999: On the `ModuleSpecifierResolutionHost`, `host.fileExists` is always defined, but `host.readFile` is optionally defined. 

Searching the codebase for `host.fileExists` showed it being used multiple times without a check, so I felt safe removing this. If you think this is wrong or risky, I'm happy to replace this with 

```
if (host.fileExists === undefined || !host.readFile)
```
 or however you think the codebase is best maintained.

src/compiler/sys.ts
```diff
@@ -131,9 +131,6 @@ let pollingChunkSize = createPollingIntervalBasedLevels(defaultChunkLevels);
 export let unchangedPollThresholds = createPollingIntervalBasedLevels(defaultChunkLevels);
 
 function setCustomPollingValues(system: System) {
-    if (!system.getEnvironmentVariable) {
```
AlexMarkUSDS, line 134: I used VSCode to search for all implementations of `System` and found only the node `System` defined in this file. `getEnvironmentVariable` is always defined as

```
            getEnvironmentVariable(name: string) {
                return process.env[name] || "";
            },
```

so I felt safe removing this check -- the condition cannot currently evaluate to true.

src/services/stringCompletions.ts
```diff
@@ -1182,8 +1182,6 @@ function getCompletionEntriesFromTypings(host: LanguageServiceHost, options: Com
 }
 
 function enumerateNodeModulesVisibleToScript(host: LanguageServiceHost, scriptPath: string): readonly string[] {
-    if (!host.readFile || !host.fileExists) return emptyArray;
```
AlexMarkUSDS, line 1185: The LanguageServiceHost has a comment

```
     * Unlike `realpath and `readDirectory`, `readFile` and `fileExists` are now _required_
     * to properly acquire and setup source files under module: node16+ modes.
```

so I believe these checks have been unnecessary since https://github.com/microsoft/TypeScript/commit/94724a8c2e68a4c7e267072ca79971f317c45e4a#diff-9717eb6d935579eea3b5662088dd258d4e1e17e2a405c3f1fa431e981e01b08f

AlexMarkUSDS (commented):
My final commits show changes to the compiler caused by this change. A number of paths were revealed to be unreachable by this commit, so I've removed that code and left comments as to why I think that was safe to do. 

If I was wrong (and I am not familiar with TS internals, so I easily could be), or if that makes the scope of this PR larger than you'd like), I'm happy to go back and replace those with explicit undefined checks.

src/compiler/moduleSpecifiers.ts
jakebailey: This one I believe is correct; this same change was made in #53463.

src/services/stringCompletions.ts
jakebailey: I would git blame this; node16 is a "recent" addition and requiring those would be new, so this is likely a compatibility shim to ensure that older callers don't crash. We do this sort of thing somewhat often when making "breaking" API changes (but, still detecting older uses and trying to fix them within reason).

jakebailey:
@typescript-bot test top200
@typescript-bot run dt

@typescript-bot pack this

src/services/stringCompletions.ts
AlexMarkUSDS: It looks like it went in 6 years ago with https://github.com/microsoft/TypeScript/commit/70682b77991d02a3ba9b2438397d6bd82cf5637f#diff-10538e33f93edc9b510a4cd505ecaa718f6ab878a54b7d044a18ef10b1e36792R383 . 

Better understanding the compatibility shim, I'll undo the changes that assume `host.readFile` and `host.fileExists`. I wouldn't want an innocent change like this to break someone's module compatibility.

AlexMarkUSDS:
Hey @jakebailey -- thank you for such a quick review. I've re-added the 3 statements that may be required for back-compatibility with https://github.com/microsoft/TypeScript/pull/57114/commits/4bac8898eb591e832dcdcfcc7c93d8665e48416e.

AlexMarkUSDS:
I'm happy to close this out (or have it closed) based on the discussion of https://github.com/microsoft/TypeScript/issues/34815 -- it's a lot of false positives.

Issue #55932, opened 2023-10-01
Remapped generic array type Can't pass to rest parameters


maxzhou0:
### 🔎 Search Terms

rest parameters  tuple array generic

### 🕗 Version & Regression Information

 I tried some versions from 3.3.3 to  5.2.2 and Nightly version 。none of them fixed this problem.

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.2.2&ssl=13&ssc=119&pln=1&pc=1#code/ATAuE8AcFNgJWpANgQwMbQDwBVjQB6jQB2AJgM7ADaAdHSseALoA0wAgmwEIB8wAvMADeAWABQIEFQDSwAJbFgAa2jgA9gDNg2JgC5tMpnkIkKHYAH5gXYPuyGA3OJABfJ2OdgosBMnTQABRQAJxQAW3IcYyIySlp6RlZOXn4ACjoaEIBzcl1fVAxMKmIAVzCAI2hgllKKqqSWXgBKfh4GcAcAek65SjUlTwgYeEQCwJDw8gBGTHY4OHYATWjTOIz2hpT0+mCcvNH-WfmllmSeFrbGLs6NFDkkaFJzYOhyUGBICbDoImDgMJKb2AlWAmmADHBwVC4C8MBog28Iz8GCCoQiACYogQYmZtpldrp2hd2qdGjw0utdrl8odUeEflVItgeKTmq12g5gN1bvdHs9Xu9PmiGX8AUCQWCISFobDoPCPAqQEMfNAwihIDBSFiTLFqOtEnx+KIJJIZPJFCp1FodLp7NImJ4XAjhgBlORhZDQADCKHIWFw2NWeoSzENwDx2WpqvVmpw53ZV25dweT3YwBeQKF9N+-0B7wlWilUJQMOVNCAA

### 💻 Code

```ts
  type Replace<T extends [...any], A, B> = {
    [K in keyof T]: T[K] extends A ? B : T[K];
  };

  type ReplaceParams<T extends [...any],A,B>=(...args:Replace<[number,number],A,B>)=>any;//is ok
  type ReplaceParams1<ARRAY extends [...any],A,B>=(...args:Replace<ARRAY,A,B>)=>any;//failed A rest parameter must be of an array type.
  type ReplaceParams2<T extends (...arg:any)=>any,A,B>=(...args:Replace<Parameters<T>,A,B>)=>any; //failed A rest parameter must be of an array type.


[...]

Issue #29919, opened 2019-02-14
When trying to use mapped tuples as rest parameters error 'A rest parameter must be of an array type' given


Roaders:
**TypeScript Version:**  3.2

**Search Terms:** mapped tuples rest 

**Code**

```ts
type FuncParams<T> = T extends (...args: infer P) => any ? P : never;
type Stringify<T> = {
    [K in keyof T]: string;
};
type Optional<T> = {
    [K in keyof T]?: T[K];
};

type ThreeParamFunc = (paramOne: string, paramTwo: number, paramThree: boolean) => void;

type Params = FuncParams<ThreeParamFunc>; // [string, number, boolean]
type StringParams = Stringify<FuncParams<ThreeParamFunc>>; // [string, string, string]
type OptionalParams = Optional<FuncParams<ThreeParamFunc>>; // [string?, number?, boolean?]

function doStuff<T>(func: T, ...params: FuncParams<T>) { // works fine
}

function doOptionalStuff<T>(func: T, ...params: Optional<FuncParams<T>>) { // A rest parameter must be of an array type.
}

function doStringStuff<T>(func: T, ...params: Stringify<FuncParams<T>>) { // A rest parameter must be of an array type.
}
```

**Expected behavior:**
I should be able to use a mapped tuple as a rest param in a function

**Actual behavior:**
I get the error `A rest parameter must be of an array type.`

**Playground Link:** 
[...]

PR #57116, opened 2024-01-21
Avoid creating rest elements with `errorType` when `any` is spread

Andarist:
This isn't particularly observable as far as I know. However, it turns out that currently `[...any]` is normalized~ to `[...errorType]`. `errorType` is a special `any` and displays as such, so this went unnoticed. However, since `[...any]` is legal it doesn't make sense to produce `errorType` in this situation at all.

EDIT:// this is more observable than what I've described above. I just pushed out a test for this. This PR fixes https://github.com/microsoft/TypeScript/issues/55932 that was incorrectly~ attributed to be a duplicate of https://github.com/microsoft/TypeScript/issues/29919

fixes https://github.com/microsoft/TypeScript/issues/57389


tests/baselines/reference/argumentsSpreadRestIterables(target=es5).types
```diff
@@ -24,10 +24,10 @@ declare const itNum: Iterable<number>
 
 ;(function(a, ...rest) {})('', true, ...itNum)
 >(function(a, ...rest) {})('', true, ...itNum) : void
->(function(a, ...rest) {}) : (a: string, rest_0: boolean, ...rest_1: any[]) => void
->function(a, ...rest) {} : (a: string, rest_0: boolean, ...rest_1: any[]) => void
+>(function(a, ...rest) {}) : (a: string, rest_0: boolean, ...rest_1: Iterable<number>[]) => void
```
Andarist, line 27: This test targets `es5` where `Iterable` doesn't exist. This change shows how now an unresolved type continues to be displayed with its alias name~. Just like here:
```ts
type Test = {
  foo: Thing; // even though it's unresolved and errors, it is still displayed as `Thing`
};
```

And actually, this matches the display of those IIFEs above this one. So it turns out that this one was going through `createNormalizedTupleType` that accidentally lost this display value.

fatcerberus:
> normalized~

> incorrectly~

I still haven’t figured out how to interpret these tildes in your text

Andarist:
😅 I use this as a softener of sorts - when I'm not sure if the used term is 100% precise etc ;p Sorry for making it confusing, I'm not sure where I got it from.

sandersn:
Turns out this is a fix for a break introduced in https://github.com/microsoft/TypeScript/pull/57031 and reported a few days ago in #57389

@ahejlsberg @Andarist reported this as bug in #57389. Can you confirm and decide whether it's worth taking in 5.4?

DanielRosenwasser:
@typescript-bot test top200

gabritto (approved):

jakebailey:

DanielRosenwasser (approved):

DavidArchibald:
This is also observable with something like this:
```ts
type SomeCallback<Args extends readonly any[]> = (...args: Args) => void;

function call<Callback extends SomeCallback<any>>(x: Callback, args: Parameters<Callback>) {
    x(...args);
}
```

Maybe this is just a simpler #55932 but I only knew how to fix this particular issue (which I actually ran into) because I've been reading the TypeScript issue tracker.

PR #57122, opened 2024-01-22
Improve apparent type of mapped types

### LINKED ISSUE #29919: When trying to use mapped tuples as rest parameters error 'A rest parameter must be of an array type' given

### LINKED ISSUE #56726: Non-homomorphic instantiation of a homomorphic mapped type is not spreadable when it has an array constraint

ahejlsberg:
This PR supersedes #56727 (which was reverted). The improvements in this PR don't exhibit the issue that caused us to revert #56727.

EDIT: ~However, the PR does break one test because the improved resolution of apparent types for mapped types causes the issue reported in #29919 to surface in that test.~

~I'm going to recommend we merge #50034, which fixes #29919 and thus the broken test in this PR.~

Fixes #29919.
Fixes #56726.


ahejlsberg:
@typescript-bot test top100
@typescript-bot run dt

src/compiler/checker.ts
```diff
@@ -14542,11 +14542,15 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
     function getResolvedApparentTypeOfMappedType(type: MappedType) {
-        const typeVariable = getHomomorphicTypeVariable(type);
-        if (typeVariable && !type.declaration.nameType) {
-            const constraint = getConstraintOfTypeParameter(typeVariable);
-            if (constraint && everyType(constraint, isArrayOrTupleType)) {
-                return instantiateType(type, prependTypeMapping(typeVariable, constraint, type.mapper));
+        const target = (type.target ?? type) as MappedType;
+        const typeVariable = getHomomorphicTypeVariable(target);
+        if (typeVariable && !target.declaration.nameType) {
+            const constraint = getConstraintTypeFromMappedType(type);
```
Andarist, line 14548: All of this makes sense now... it is tricky that mapped types don't have constraints resolvable by `getBaseConstraintOfType` and such but we can reach for `getConstraintTypeFromMappedType` and use that.

src/compiler/checker.ts
```diff
@@ -14542,11 +14542,15 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+            if (constraint.flags & TypeFlags.Index) {
```
Andarist, line 14549: I wonder if this check shouldn't be replaced with an assert - it's quite a strong invariant that this branch can only be entered after `getHomomorphicTypeVariable` returns true and that requires the constraint to be an `Index` type.

Andarist:
For what it's worth - I spent some sleepless nights on this last week so I built up some good context around this issue. LGTM (pending the resolution of https://github.com/microsoft/TypeScript/pull/50034 ). 

One of the test cases from https://github.com/microsoft/TypeScript/issues/29919 requires https://github.com/microsoft/TypeScript/pull/57116 to be fixed:
```ts
type Replace<T extends [...any], A, B> = {
  [K in keyof T]: T[K] extends A ? B : T[K];
};

type ReplaceParams1<ARRAY extends [...any], A, B> = (
  ...args: Replace<ARRAY, A, B>
) => any;
```

I confirmed that only a combination of this PR here (#57122) and #57116 fixes this problem.

src/compiler/checker.ts
ahejlsberg: An assert here wouldn't be right. There many cases where an instantiation of a homomorphic mapped type doesn't have a `keyof XXX` constraint anymore. For example, the homomorphic `Readonly<T>` has the constraint type `keyof T`, but the instantiation `Readonly<{ a: string, b: string }>` has the constraint type `"a" | "b"`. What we're looking for here are instantiations where the constraint type is _still_ a `keyof XXX`. That's what wasn't handled before.

Andarist: Thank you for the explanation - it makes sense, `getHomomorphicTypeVariable` is called now with the `target` and not the `type` after all.

gabritto:
@typescript-bot pack this

tests/baselines/reference/ramdaToolsNoInfinite.errors.txt
```diff
@@ -0,0 +1,126 @@
...
+    
+        type CleanedGaps<T extends any[]> = {
+            [K in keyof T]: NonNullable<T[K]>
+        };
+    
+        type Gaps<T extends any[]> = CleanedGaps<PartialGaps<T>>;
+    
+        type Curry<F extends ((...args: any) => any)> =
+            <T extends any[]>(...args: Tools.Cast<Tools.Cast<T, Gaps<Parameters<F>>>, any[]>) =>
+                              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
gabritto, line 120: How is the change in the PR causing this error?

Andarist: I believe the underlying issue is the same as it was in my PRs. This is now resolved through layers and thus `getResolvedApparentTypeOfMappedType` sees the constraint of `Parameters<F>` - that once gets computed as `unknown[]` (by `getDefaultConstraintOfConditionalType` when restrictive instantiation bails out from `getConstraintOfDistributiveConditionalType`) and once as `any` (by `getConstraintOfDistributiveConditionalType`). So `any` can get returned and `getResolvedApparentTypeOfMappedType` can't recognize this as an array-like. So it ends up not d about it not being assignable to `readonly any[]` (that is being checked to determine if the type can be used as rest). 

IIRC the other case (when `unknown[]`) is returned makes `any[]` (added by this `Cast`) eliminated from the substitution type since, at that point, it's determined that the `baseType` satisfies the added constraint - so the `baseType` gets returned.

Andarist: So the issue is mainly in the fact that previously the substitution type wasn't simplified at all here and that `& any[]` in it (coming from its added constraint) was preserved and made the comparison against `readonly any[]` OK.

gabritto: What do you mean by "this is now resolved through layers"?
How do we go from a call to `getResolvedApparentTypeOfMappedType` to calls to `getDefaultConstraintOfConditionalType` and `getConstraintOfDistributiveConditionalType`?

Andarist: Please note that I'm partially speaking based on what I remember about debugging this last week - so some details might be slightly off.

`Gaps` is a mapped type - so we enter `getResolvedApparentTypeOfMappedType`. With this PR we recognize that an instantiation of a homomorphic mapped type is received and we end up checking its constraint and if it's an index type. It turns out it is because it's something like `keyof Parameters<...>`. So we grab the `.type` and compute its base constraint. So now we are computing the constraint of a conditional type (through `getConstraintFromConditionalType`) and, as mentioned above, this once resolves through `getDefaultConstraintOfConditionalType` (for the restrictive instantiation case) and once through `getConstraintOfDistributiveConditionalType` (for a regular case :P).

ahejlsberg:
Latest PR incorporates the changes in #50034, plus an additional change to fix this inconsistency:

```ts
function foo<T extends string[]>(
    fa: (s: string, ...args: string[]) => string,
    fb: (s: string, ...args: T) => string
) {
    const f1: (...args: any) => string = fa;  // Ok
    const f2: (...args: any[]) => string = fa;  // Error
    const f3: (...args: any) => string = fb;  // Ok
    const f4: (...args: any[]) => string = fb;  // Error
}
```

Before this PR, the assignment to `f4` was an error, which is inconsistent with `f2`. With #50034, both of `f3` and `f4` become errors, which also isn't right. With the latest commit, all of the above are now ok.

tests/baselines/reference/ramdaToolsNoInfinite.errors.txt
ahejlsberg: With latest commit this error is gone.

ahejlsberg:
@typescript-bot test top100
@typescript-bot run dt

gabritto:
@typescript-bot pack this

gabritto (approved):

ahejlsberg:
The test runs reveal inconsistencies related to indexing arrays with numeric string literals (as in #50034):

```ts
function bar<T extends string[], K extends number>() {
    type T00 = string[]["0"];  // Ok
    type T01 = string[]["0.0"];  // Error, as expected
    type T02 = string[][K | "0"];  // Error, but shouldn't be
    type T10 = T["0"];  // Error, but shouldn't be
    type T11 = T["0.0"];  // Error, as expected
    type T12 = T[K | "0"];  // Error, but shouldn't be
}
```

The errors on `T01` and `T11` are expected because the numeric string literal doesn't round-trip (i.e. when converted to number and back to string, it yields a different representation). However, `T02`, `T10`, and `T12` shouldn't error and I will fix that. Regardless of the fix, I don't think we want to include `` `${number}` `` in the index type for types that have a `number` index signature. First, it isn't quite correct because `` `${number}` `` doesn't require matching strings to round-trip, and second, it is a breaking change that I'd rather avoid. I think the best place to land is simply a targeted fix to make indexing with round-tripping numeric strings consistent as per the example above.

ahejlsberg:
@typescript-bot test top100
@typescript-bot run dt

tests/cases/compiler/assignmentToAnyArrayRestParameters.ts
```diff
@@ -12,3 +12,12 @@ function foo<T extends string[]>(
...
 }
+
+function bar<T extends string[], K extends number>() {
+    type T00 = string[]["0"];
+    type T01 = string[]["0.0"];  // Error
+    type T02 = string[][K | "0"];
+    type T10 = T["0"];
+    type T11 = T["0.0"];  // Error
+    type T12 = T[K | "0"];
+}
```
Andarist, line 23: nit: those should likely be in a different test file since they don't have anything in common with this "test title" (`assignmentToAnyArrayRestParameters`)

ahejlsberg:
Test and performance suites are all clean. This is ready to merge.

PR #57123, opened 2024-01-22
enable `@typescript-eslint/no-unused-vars`

### LINKED ISSUE #57124: enable `@typescript-eslint/no-unused-vars`
### Acknowledgement

### Comment

The ESLint config file references investigating whether `@typescript-eslint/no-unused-vars` can be enabled. I have done so in [PR 57123](https://github.com/microsoft/TypeScript/pull/57123).

abrahamguo:
Resolves the TODO comment regarding enabling the ESLint rule `@typescript-eslint/no-unused-vars`

Fixes #57124


abrahamguo:

jakebailey (changes requested):
You should see #55139 for context. We can't accept this kind of change without problems (and this is a duplicate)

jakebailey:
Well, we could of course ignore lines for JSDoc, like you've done here.

package-lock.json
```diff
@@ -49,7 +49,7 @@
                 "mocha-fivemat-progress-reporter": "^0.1.0",
                 "ms": "^2.1.3",
                 "node-fetch": "^3.3.2",
-                "playwright": "^1.41.0",
+                "playwright": "^1.41.1",
```
jakebailey, line 52: This change doesn't seem to be related and should be reverted.

package.json
jakebailey, line 1: This change doesn't seem to be related and should be reverted.

package-lock.json
abrahamguo: apologies, I always run `npm up --save` by default

abrahamguo:
Yes, I think that (at least for now) it seems like an easy task to suppress the <5 places with JSDoc issues in order to get the benefit of this rule applying throughout the whole codebase.

.eslintrc.json
```diff
@@ -105,9 +105,13 @@
                 }
             }
         ],
-
-        // Todo: For each of these, investigate whether we want to enable them ✨
-        "@typescript-eslint/no-unused-vars": "off",
+        "@typescript-eslint/no-unused-vars": [
```
jakebailey, line 108: If anything, I would probably prefer the settings on https://github.com/microsoft/TypeScript/pull/55139/files#diff-6884918dc8291219be508e05e28965b958c734def85324f3b53858ea4702090fR81, as they match TypeScript's. ignoreRestSiblings is interesting, though.

package.json
```diff
@@ -93,7 +93,7 @@
         "build:tests:notypecheck": "hereby tests --no-typecheck",
         "clean": "hereby clean",
         "gulp": "hereby",
-        "lint": "hereby lint",
+        "lint": "DEBUG=eslint:cli-engine hereby lint",
```
jakebailey, line 96: This should also be removed.

jakebailey:
Generally speaking, I prefer the settings and changes in #55139, besides the JSDoc plugin thing.

That being said, the regex should probably be changed to `^_[^_]` as your PR has pointed out that `__String` is treated as used which isn't intentional.

.eslintrc.json
```diff
@@ -105,9 +105,14 @@
...
+            "error",
+            {
+                "argsIgnorePattern": "^_([^_]|_*$)",
```
jakebailey, line 111: Hm, what is the second variant needed for? That regex is likely to be quite inefficient.

abrahamguo: - about 20 places in the codebase define arrow functions with no parameters, but use `_ => ... ` rather than `() => ...`. although now that I look at the rest of the codebase, there are many more places that use the normal `() => ...` form, so I could update those ~20ish places to match?
- These two places in `src/testRunner/parallel/shared.ts` do the same thing, but with two unused parameters. Therefore I made this generalized so that it supports a variable name made up of any number of underscores and nothing else
image

jakebailey: Oh, then that can just be `^_($|[^_)`.

jakebailey: Ignore me, I think what you write is fine; I misread it as a more complicated one that does arbitrary lookup.

src/compiler/emitter.ts
```diff
@@ -4131,7 +4131,7 @@ export function createPrinter(printerOptions: PrinterOptions = {}, handlers: Pri
         writeSpace();
         nextPos = emitTokenWithComment(SyntaxKind.AsKeyword, nextPos, writeKeyword, node);
         writeSpace();
-        nextPos = emitTokenWithComment(SyntaxKind.NamespaceKeyword, nextPos, writeKeyword, node);
```
jakebailey, line 4134: Here, I would prefer to leave this code and ignore it. If this code changes, we may not realize we didn't update the pos.

src/compiler/core.ts
```diff
@@ -286,6 +285,7 @@ export function filter<T>(array: T[], f: (x: T) => boolean): T[];
 /** @internal */
 export function filter<T, U extends T>(array: readonly T[], f: (x: T) => x is U): readonly U[];
 /** @internal */
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
jakebailey, line 288: I think we should be able to drop the type vars in this file.

abrahamguo: The reason why all these are suppressed is because TypeScript requires that all overloads of a function must accept the same type parameters. So while the type parameter may be unused for a given overload, removing it causes TS errors.

jakebailey: I don't think that's true; see also my PR https://github.com/microsoft/TypeScript/pull/55139/files#diff-8eee8edc4cdf82aaa3b0c0ba3c8c74fa80ac6d5214540a460c49fb2ee1031b25R289

jakebailey:
Another thing this PR is missing is to remove the unused variable check from our tsconfig base.

jakebailey:
We talked about this in the design meeting today and tentatively approved it; one thing that we want changed though is to remove the ignores in the plain JS files, and instead replace them with things like `import("@typescript-eslint/utils").TSESTree.Node` or similar.

jakebailey:
I also think we could remove `allowUnusedLabels` too, we're already getting `no-unused-labels` from eslint.

.eslintrc.json
```diff
@@ -105,9 +105,13 @@
...
```
jakebailey, line 111: For clarity, I do wonder if we should just have `^(_+$|_[^_])` for these, i.e. "all underscores, or starts with only one underscore".

abrahamguo: makes sense!

src/compiler/watchPublic.ts
```diff
@@ -264,6 +264,7 @@ export interface ProgramHost<T extends BuilderProgram> {
  *
  * @internal
  */
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
 export interface ProgramHost<T extends BuilderProgram> {
```
jakebailey, line 268: Can you bring in the change from https://github.com/microsoft/TypeScript/pull/55139/files#diff-ca579690d4b8ab4c83f5d18e3e8797d11a4c814ce32afe840eb3d7b797c46df5 instead? It's more typical for us to be using internal properties than declaration merging for this purpose.

jakebailey:
The scripts are currently inconsistent as to their use of typedefs or import type nodes; can you switch them all to be import type nodes? At least for the ones that aren't compound like `CallOrNewExpression`?

abrahamguo:
Sure — my thinking was to create typedefs for ones that are used more than once, since the imports are long, but I can inline the typedefs if you like that 👍🏻

jakebailey:
That makes sense, though I think having the code be more easily copy-paste-able between the rules is probably worthwhile.

Eventually we'll have https://github.com/microsoft/TypeScript/pull/57207, though. That'll be awesome.

src/compiler/factory/utilities.ts
```diff
@@ -160,6 +160,8 @@ import {
     ReadonlyKeyword,
     RelationalOperator,
     RelationalOperatorOrHigher,
+    // Will become used when explicit return types are added in declaration file during build step.
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
     SetAccessorDeclaration,
```
jakebailey, line 165: One thing I'll want to check is that this plays nicely with organizing improts / `eslint-plugin-simple-import-sort`. A good way to test is to swap a few of these import lines around and then see if this comment stays in the right place when fixing it with eslint (or using organize imports with the right settings). I'll check that now.

jakebailey (approved):
Looks good to me, thanks for bearing with all of my comments.

abrahamguo:
No worries! Thanks for the detailed review.

Would you be open to future PRs turning on other ESLint rules for the codebase?
In particular, I've been looking at `prefer-destructuring` or `logical-assignment-operators` which are both autofixable, as well as some of the rules from `@typescript-eslint/recommended-requiring-type-checking`.
I wasn't sure if there was any bigger vision one way or another.

jakebailey:
Enabling the rules that use type checking is a bit of a perf hole; I wouldn't attempt to enable any of those unless there are some that look super valuable.

Last I checked `logical-assignment-operators` has a load of false positives; I tried to enable it and fix everything but I had to sift through hundreds of changes to make sure none caused a behavior change.

There may be others that are valuable but I think it really depends.

abrahamguo:
`logical-assignment-operators` looks to just change `a = a || b` to `a ||= b` (and same with `&&` `??`), so it looks to be guaranteed correctly autofixable.

I'll look into the type-checked lint rules and maybe do some performance monitoring.

jakebailey:
Ah, I was thinking of `prefer-nullish-coalescing`.

But generally, know that it isn't a goal to enable every lint rule; the codebase is huge and it slows down linting and editor performance. Most on the team don't even run eslint in the editor for that reason. We really only want cheap rules / rules that have high value.

Issue #54018, opened 2023-04-25
Tracking issue: Named imports from CJS module incorrectly allowed in nodenext


andrewbranch:
This issue has come up a few times, most recently at https://github.com/DefinitelyTyped/DefinitelyTyped/pull/65147#issuecomment-1515134407, and I don’t think we’ve had a canonical place to explain, discuss, or track it.

## The symptom

Sometimes, when you’re writing an ES module in `--module nodenext`, you will want to import a CJS dependency, and TypeScript lets you used a named import:

```ts
import { createProgram } from "typescript";
createProgram(/* ... */);
```

but Node complains:

```
file:///project/out/main.mjs:1
import { createProgram } from "typescript";
         ^^^^^^^^^^^^^
SyntaxError: Named export 'createProgram' not found. The requested module 'typescript' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'typescript';
const { createProgram } = pkg;
```

Alternatively but equivalently, you might try to use a namespace import:

```ts
import * as ts from "typescript";
ts.createProgram(/* ... */);
```

and Node gives a much less helpful error:

```
file:///project/out/main.mjs:2
ts.createProgram();
   ^

TypeError: ts.createProgram is not a function
```

In either case, you find that you instead need to write a default import:

```ts
import ts from "typescript";
ts.createProgram(/* ... */);
```

Both TypeScript and Node are happy with this, but why did TypeScript let you write the forms that crash at runtime?

## The problem

Node uses [cjs-module-lexer](https://github.com/nodejs/cjs-module-lexer) to syntactically analyze CommonJS modules without executing them in order to turn `module.exports` properties into named exports that can be imported as named (or namespace) imports by ES modules. However, syntactic analysis has its limitations, and not all `module.exports` properties get detected on all modules. (In practice, it’s common for this to be all-or-nothing, as with the `typescript` package in its current state: it has _no_ named exports available.)

Type definitions have no way of declaring which exports/properties of a CommonJS module will be detectable by Node’s named export analysis. In other words, `typescript.d.ts` is not misrepresenting the shape of the module; it would be impossible to “fix” it for Node ESM consumers without breaking it for CJS consumers or bundlers, which typically have more relaxed interop rules. TypeScript currently assumes that _all_ declared exports/properties of a CJS module will be detectable by Node and available as named imports. This is unsound, leading to the error in the example above.

[...]

PR #57133, opened 2024-01-23
"Annotate" exported object to fix named / namespace imports of our API in Node ESM

### LINKED ISSUE #56366: TS Build api failed when using node ESM module (mjs)
### 🔎 Search Terms

ESM ts.sys undefined 

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about modules

### ⏯ Playground Link

https://github.com/stevebeauge/tsc-api-esm

### 💻 Code

```ts
import * as tsc from "typescript";

const tsSystem = tsc.sys;

console.log({ tsSystem });

```

### 🙁 Actual behavior

If I transpile this code into CJS and run `node index.cjs` works as expected (tsc.sys is defined).

If I transpile this code into ESM and run `node index.mjs` the code does not works as expected (tsc.sys is undefined).

### 🙂 Expected behavior

tsc.sys (`System`) should be available in CLI tools, whether the output is ESM or CJS.

The issue occurs when trying to create a CLI tool that require parsing tsconfig files, including potential "extends". Typescript provides an API for that.

However, my cli tool requires to be ESM because of some imports not available as CJS.

Repro steps (using pnpm but it does not matter):

````
git clone https://github.com/stevebeauge/tsc-api-esm
cd tsc-api-esm
pnpm i 
node index.cjs
node index.mjs
````

When using cjs : OK
When using ESM : KO

My investigations leds me to : https://github.com/microsoft/TypeScript/blob/80ab111685e3e0c10d8d13cdd18d8dffa8328895/src/compiler/core.ts#L2748

It populates the tsc.sys when the function "isNodeLikeSystem" returns true. However, the test expect to have `module` be defined (which is true in CJS, not in ESM).

### PR DESCRIPTION
**Pull request by jakebailey**
Our current API bundle looks like this:

```js
var ts = (() => { ... })();
if (typeof module !== "undefined" && module.exports) {
	module.exports = ts;
}
```

[cjs-module-lexer](https://github.com/nodejs/cjs-module-lexer) cannot understand this. As such, this code fails:

```js
import { createSourceFile } from "typescript";
```

```
file:///home/jabaile/work/TypeScript/complain.mjs:2
    createSourceFile,
    ^^^^^^^^^^^^^^^^
SyntaxError: Named export 'createSourceFile' not found. The requested module 'typescript' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'typescript';
```

This has been the case with TypeScript for the entire time that Node's ESM support has been around. Even before 5.0, it didn't work since our exported object was still complicated.

However, there's a footgun; this code _will not_ error:

```js
import * as ts from "typescript";
```

In TypeScript code, we'll emit helpers that fix this up. But in native ESM, this will only be an object with a `default` property! You'll write `ts.createSourceFile`, and it'll be undefined and fail later.

To fix this, we need to make `cjs-module-lexer` aware of our exports. When emitting `cjs`, `esbuild` uses a trick to "declare" the names to Node:

```js
module.exports = someComplicatedThing();
0 && (module.exports = {
	foo,
	bar,
});
```

This is dead code, but Node will still read the names.

~If we do this same thing at the bottom of our custom bundle, we can also get named imports working. Unfortunately, there's no static way within esbuild for us to know the names via the API (likely, https://github.com/evanw/esbuild/issues/3110 or https://github.com/evanw/esbuild/issues/3281; I also filed https://github.com/evanw/esbuild/issues/3607), so we have to import the bundle and grab the names. But, once we have them, we can tack the output onto the bundle.~

To fix this, the PR now instead emits CJS, then uses the header/footer (instead of just a footer + IIFE) to do something like UMD. This lets us use esbuild to make the annotations.

This now means that _both_ of these imports work!

```js
import * as ts from "typescript";
import { createSourceFile } from "typescript";

ts.createSourceFile(...);
```

Fixes #56366

Related: #54018


jakebailey:
@typescript-bot pack this

jakebailey:

fatcerberus:
> But in native ESM, this is the empty object! You'll write ts.createSourceFile, and it'll be undefined and fail later.

This feels like it should be considered a bug in Node.  IMO if it's going to error on named imports because it doesn't understand how `module.exports` is constructed, then it should also error on a wildcard import instead of just silently returning an empty object.  That's a pretty big footgun; I don't imagine TS users are the only victims of it.

jakebailey:
> > But in native ESM, this is the empty object! You'll write ts.createSourceFile, and it'll be undefined and fail later.
> 
> This feels like it should be considered a bug in Node. IMO if it's going to error on named imports because it doesn't understand how `module.exports` is constructed, then it should also error on a wildcard import instead of just silently returning an empty object. That's a pretty big footgun; I don't imagine TS users are the only victims of it.

I think I over simplified; `.default` works, so it's not the empty object.

fatcerberus:
Even so, that’s still a pretty big footgun.  Anyone successfully doing a wildcard import from a CJS module probably expects to be able to do more than just `foo.default.bar()`, else they would just default-import in the first place.

jakebailey:
One problem with this approach is that live bindings don't actually work if you use named or namespace imports from native ESM. I'm not totally sure why, as my understanding was that they worked in esbuild thanks to all of the getters and such.

I can reproduce exactly the same behavior using esbuild's CJS output, so maybe this isn't a deal breaker and I'm misunderstanding things (live bindings work for the default export). Node also says that they don't update: https://nodejs.org/api/esm.html#:~:text=Live%20binding%20updates%20or%20new%20exports%20added%20to%20module.exports%20are%20not%20detected%20for%20these%20named%20exports.

It's not like we have any actual bindings that update after the fact, besides say, sys, which is our only mutable export for "reasons".

The default export is definitely a better thing to be doing at the user level, though, but I'm not sure how to statically force users of the TS API to do so (and people _definitely_ use a star import more often than a default).

jakebailey:
During the design meeting, @andrewbranch pointed to #54018, which funnily contains the same explanation I independently described (which I am going to place on the Somerton Scale as "subconscious appropriation" 😄)

andrewbranch (approved):

RyanCavanaugh (approved):

jakebailey:
Evan pointed out that we could just flip our bundle from "global that's conditionally turned into CJS" to "CHS that's conditionally turned into a global", a la UMD. If we then switch to named exports, the annotations will be produced and the lexer will dig into the closure.

There are some gotchas but I'm going to draft this in the meantime.

jakebailey:
I used @evanw's suggestion in https://github.com/evanw/esbuild/issues/3607#issuecomment-1911523751 and pushed it here; the only gotcha is that pesky `__esModule` property, which I'm currently hacking away by replacing esbuild's helper that adds it (rather than copying the export object a second time).

The new bundle is now actually 5% faster to start up and 5% smaller (8.8MB to 8.3MB):

| - | - | - | - | - |
| Packed | 5.55 MiB | 5.54 MiB | -16.86 KiB | -0.30% |
| Unpacked | 30.85 MiB | 30.35 MiB | -512.46 KiB | -1.62% |

| - | - | - | - | - |
| `lib/typescript.js` | 8.71 MiB | 8.21 MiB | -512.46 KiB | -5.75% |

---

@typescript-bot perf test

jakebailey:
I'm undrafting this; the only concern is the hack for CJS, which I've mitigated by asserting that the replacement pattern must match exactly once.

That and the timing, given we just pushed out the beta and this does change the bundle quite a bit.

jakebailey:
@typescript-bot pack this

jakebailey:
Oof, this breaks monaco.

jakebailey:
@typescript-bot pack this

jakebailey:
Alright, it's fixed now. Arguably monaco is doing the wrong thing in their build by using `var ts` instead of the fake `module.exports` I added (my fault for not doing that when I fixed their codebase a while ago), but it's not hard to keep working in case someone else does this same thing.

src/compiler/core.ts
```diff
@@ -2749,6 +2749,5 @@ export function isNodeLikeSystem(): boolean {
     // and definitely exist.
     return typeof process !== "undefined"
         && !!process.nextTick
-        && !(process as any).browser
-        && typeof module === "object";
```
jakebailey, line 2753: In the new version, `module` is a parameter and is always provided. This check would no longer do anything. I'm not sure that it really matters that we don't have this.

andrewbranch (approved):

src/compiler/core.ts
jakebailey: Actually, I'm going to switch this back to the same definition it used to be, e.g. checking for `require`. The comment I left above about `require` stopped being correct at some point as I refined our output bundle format to rewrite `require` temporarily.

jakebailey: (I don't believe this will end up mattering in practice, but that's why we merge these sorts of things early in the cycle...)

PR #57143, opened 2024-01-23
Add missing members to array of objects in quickfix

### LINKED ISSUE #57142: Quickfix for missing properties of object in array not available
### 🔎 Search Terms

quickfix fix array missing property member

- This is the behavior in every version I tried, and I reviewed the FAQ for entries

### ⏯ Playground Link

_No response_

### 💻 Code

```ts
interface A {
    b: string;
    c: string;
}

const foo: A[] = [{}];
```

### 🙁 Actual behavior

No quickfix is available for the missing properties

### 🙂 Expected behavior

A quickfix is offered to add the properties `b` and `c` to the object in the array

I would like to provide a MR that fixes this

hborchardt:
If an object with missing properties is part of an array, then the token where the error occurs is an OpenBraceToken, which was not handled by the quickfix.

The correct type is available on the parent ObjectLiteralExpression.

Note that the ObjectLiteralExpression does not have a `text` property, so place the `identifier` string for deduplication on the InfoKind.ObjectLiteral structure.

The added test cases were failing on the main branch.

Fixes #57142

### DIFF
--- a/src/services/codefixes/fixAddMissingMember.ts
+++ b/src/services/codefixes/fixAddMissingMember.ts
@@ -187,7 +187,7 @@ registerCodeFix({
         return createCombinedCodeActions(textChanges.ChangeTracker.with(context, changes => {
             eachDiagnostic(context, errorCodes, diag => {
                 const info = getInfo(diag.file, diag.start, diag.code, checker, context.program);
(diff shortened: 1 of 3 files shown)


hborchardt:

sandersn (approved):
Looks right to me.

PR #55015, opened 2023-07-14
Add auto-import for the `package.json` `imports` field

emmatown:
Fixes #49492

This treats `package.json#imports` as roughly equivalent to `compilerOptions#paths` in terms of deciding which specifier to use, it also prefers `package.json#imports` over `compilerOptions#paths`

While doing this, I found some broken behaviour around extension replacement which was already observable with the `package.json#exports` field auto-import but is more visible with `package.json#imports` so this fixes that as well. `autoImportPackageJsonExportsSpecifierEndsInTs` shows an example of what's fixed for `package.json#exports`. That fix and test is in a seperate commit if that's easier to review

PR #57145, opened 2024-01-23
Deprioritize declaration emit paths from baseUrl containing node_modules

andrewbranch:
Fixes `error TS2742: The inferred type of 'resolveParserPlugins' cannot be named without a reference to 'packages/compiler-core/node_modules/@babel/parser/typings/babel-parser'. This is likely not portable. A type annotation is necessary.` reported in #57117. (It bisected to #55015, though it’s not clear why, and doesn’t seem like that PR directly introduced the buggy behavior.)

Also worth noting that Vue should almost certainly not be using `baseUrl` at all.

### DIFF
--- a/src/compiler/moduleSpecifiers.ts
+++ b/src/compiler/moduleSpecifiers.ts
@@ -437,7 +437,16 @@ function computeModuleSpecifiers(
                 redirectPathsSpecifiers = append(redirectPathsSpecifiers, local);
             }
             else if (pathIsBareSpecifier(local)) {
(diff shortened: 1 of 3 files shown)


sheetalkamat (approved):

jakebailey (approved):

jakebailey:
Does this need a cherry-pick or are we okay with a temporary break in the beta?

andrewbranch:
IMO the beta only exists to be a timestamp; it’s not a build anyone should use. Every beta is expected to have all the features and none of the bugfixes.

andrewbranch:

PR #57148, opened 2024-01-23
Fix isolatedModules check for export assignments

### LINKED ISSUE #57136: 5.4 breaks @types/node when using isolatedModules
### 🔎 Search Terms

isolatedModules "resolves to a type and must be marked" @types/node export import

- This changed between versions 5.3 and 5.4.0-dev.20240120

### ⏯ Playground Link

_No response_

### 💻 Code

```ts
declare module "node:events" {
    import events = require("events");
    export = events;
}
```
from https://unpkg.com/browse/@types/node@20.11.5/events.d.ts (at the end)

### 🙁 Actual behavior

```
node_modules/@types/node/events.d.ts:878:14 - error TS1291: 'events' resolves to a type and must be marked type-only in this file before re-exporting when 'isolatedModules' is enabled. Consider using 'import type' where 'events' is imported.

878     export = events;
                 ~~~~~~
```

### 🙂 Expected behavior

It should work, as `events` is not just a type but also a value

DT discussion: https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/68285

Originally found at https://github.com/babel/babel/pull/16222

andrewbranch:
Fixes #57136 

Originally added in #55097, this was missing both the non-ambient check and a check that the resolved symbol has no value meaning:

- We only issue isolatedModules errors on non-ambient code, because the goal of all isolatedModules errors to ensure that code can be transpiled to JS. Ambient code produces no JS, so it’s exempt from checks.
- These errors are only supposed to occur when some re-export (`events` in the test and linked issue) resolves to something that doesn’t really exist at runtime (i.e., something that’s only a type, or something whose value meaning comes from a type-only import/export). The point is it needs to be easy for a transpiler to tell whether the re-export should be emitted or not—if it resolves to something real, it can and must be emitted; otherwise, it must be marked as type-only somewhere in the file so it can be removed.
  
  To try to determine if `events` resolved to anything real, the implementation was only checking that the re-export resolves to something with a type meaning before issuing an error. But `events` has both a value meaning and a type meaning (since it’s a namespace that contains one value and two types). So the implementation needed to check that the resolved symbol doesn’t _also_ have a value meaning in order to issue an error.

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -46384,11 +46384,12 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     );
                 }
 
-                if (!isIllegalExportDefaultInCJS && getIsolatedModules(compilerOptions) && !(sym.flags & SymbolFlags.Value)) {
+                if (!isIllegalExportDefaultInCJS && !(node.flags & NodeFlags.Ambient) && getIsolatedModules(compilerOptions) && !(sym.flags & SymbolFlags.Value)) {
+                    const nonLocalMeanings = getSymbolFlags(sym, /*excludeTypeOnlyMeanings*/ false, /*excludeLocalMeanings*/ true);
                     if (
                         sym.flags & SymbolFlags.Alias
-                        && resolveAlias(sym) !== unknownSymbol
-                        && getSymbolFlags(sym, /*excludeTypeOnlyMeanings*/ false, /*excludeLocalMeanings*/ true) & SymbolFlags.Type
+                        && nonLocalMeanings & SymbolFlags.Type
+                        && !(nonLocalMeanings & SymbolFlags.Value)
                         && (!typeOnlyDeclaration || getSourceFileOfNode(typeOnlyDeclaration) !== getSourceFileOfNode(node))
                     ) {
                         // import { SomeType } from "./someModule";
--- a/tests/cases/compiler/isolatedModulesReExportAlias.ts
+++ b/tests/cases/compiler/isolatedModulesReExportAlias.ts
@@ -1,0 +1,25 @@
+// @isolatedModules: true
+// @noEmit: true
+// @noTypesAndSymbols: true
+
+// @Filename: /events.d.ts
+declare module "events" {
+  interface EventEmitterOptions {
+    captureRejections?: boolean;
+  }
+  class EventEmitter {
+    constructor(options?: EventEmitterOptions);
+  }
+  export = EventEmitter;
+}
+declare module "node:events" {
+  import events = require("events");
+  export = events;
+}
+
+// @Filename: /moreEvents.ts
+import events = require("events");
+export = events;
+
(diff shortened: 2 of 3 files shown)


src/compiler/checker.ts
```diff
@@ -46384,11 +46384,12 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     );
                 }
 
-                if (!isIllegalExportDefaultInCJS && getIsolatedModules(compilerOptions) && !(sym.flags & SymbolFlags.Value)) {
+                if (!isIllegalExportDefaultInCJS && !(node.flags & NodeFlags.Ambient) && getIsolatedModules(compilerOptions) && !(sym.flags & SymbolFlags.Value)) {
+                    const nonLocalMeanings = getSymbolFlags(sym, /*excludeTypeOnlyMeanings*/ false, /*excludeLocalMeanings*/ true);
                     if (
                         sym.flags & SymbolFlags.Alias
-                        && resolveAlias(sym) !== unknownSymbol
```
andrewbranch, line 46390: `unknownSymbol.flags === SymbolFlags.All`, so while this check was necessary before, it’s now redundant with `!(nonLocalMeanings & SymbolFlags.Value)`.

tests/cases/compiler/isolatedModulesReExportAlias.ts
```diff
@@ -0,0 +1,25 @@
...
+  import events = require("events");
+  export = events;
+}
+
+// @Filename: /moreEvents.ts
+import events = require("events");
+export = events;
+
+// @Filename: /boo.ts
+// Bad test runner (ignores stuff when last file contains the string r-e-q-u-i-r-e)
```
jakebailey, line 25: I am definitely curious about this one... 😄

src/compiler/checker.ts
jakebailey: They're actually equal to `All`? I thought that it was `SymbolFlags.Property` for this symbol.

jakebailey:
@typescript-bot test top200
@typescript-bot run dt

src/compiler/checker.ts
andrewbranch: Huh, oops, you’re right. At any rate, it’s still redundant with `!(nonLocalMeanings & SymbolFlags.Value)`, but I’m glad to fix that memory

tests/cases/compiler/isolatedModulesReExportAlias.ts
andrewbranch: The absolute worst https://github.com/microsoft/TypeScript/blob/25a4f9e53595fe944ada7cc747ec393ce0bc2c6e/src/testRunner/compilerRunner.ts#L238-L242

jakebailey (approved):

gabritto (approved):

andrewbranch:

PR #57158, opened 2024-01-24
Fix ReloadRequest in protocol found while updating to use session in tests

sheetalkamat:
This came up when i was trying to work with projects management and updating tests to use session instead of project service

### DIFF
--- a/src/testRunner/unittests/tsserver/projectErrors.ts
+++ b/src/testRunner/unittests/tsserver/projectErrors.ts
@@ -583,8 +583,7 @@ describe("unittests:: tsserver:: projectErrors:: dont include overwrite emit err
         const session = new TestSession(host);
         openFilesForSession([f1], session);
 
-        const projectService = session.getProjectService();
-        const projectFileName = projectService.inferredProjects[0].getProjectName();
+        const projectFileName = session.getProjectService().inferredProjects[0].getProjectName();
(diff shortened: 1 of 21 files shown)


src/testRunner/unittests/tsserver/projectErrors.ts
```diff
@@ -583,8 +583,7 @@ describe("unittests:: tsserver:: projectErrors:: dont include overwrite emit err
         const session = new TestSession(host);
         openFilesForSession([f1], session);
 
-        const projectService = session.getProjectService();
-        const projectFileName = projectService.inferredProjects[0].getProjectName();
+        const projectFileName = session.getProjectService().inferredProjects[0].getProjectName();
```
andrewbranch, line 586: So is `projectService` not safe to reuse between requests now?

andrewbranch (approved):

src/testRunner/unittests/tsserver/projectErrors.ts
sheetalkamat: It is but I changed this so it’s easier to track any other usages

jakebailey (approved):
Changing the protocol is _theoretically_ a breaking change, but isn't clear anyone actually has called this API? https://github.com/search?q=%2F%5Cbts%5C.%5Ba-zA-Z.%5D*ReloadProjects%5Cb%2F&type=code

sheetalkamat:
@jakebailey anyways the api was incorrect for anyone to use ..

PR #54477, opened 2023-05-31
Allow non-generic return types to be read from single generic call signatures

Andarist:
This PR permits circular references in some return types of generic call signatures - something that is already allowed in non-generic call signatures.

If the return type doesn't depend on type variables then the same thing can be allowed for it (including cases with **instantiated** return types not depending on type variables).

PR #57160, opened 2024-01-24
Revert #54477 but keep the tests

sandersn:
Discovered in #57117 for the 5.4 beta.

The implementation should not use `couldContainTypeVariables`--it's intended as a fast path, and should not be used in places where its unreliability can be observed.

The tests stay, but with a note added that they should pass but do not.

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -39347,20 +39347,17 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         }
     }
 
-    function getNonGenericReturnTypeOfSingleCallSignature(funcType: Type) {
+    function getReturnTypeOfSingleNonGenericCallSignature(funcType: Type) {
         const signature = getSingleCallSignature(funcType);
-        if (signature) {
-            const returnType = getReturnTypeOfSignature(signature);
-            if (!signature.typeParameters || !couldContainTypeVariables(returnType)) {
-                return returnType;
-            }
+        if (signature && !signature.typeParameters) {
+            return getReturnTypeOfSignature(signature);
         }
     }
 
     function getReturnTypeOfSingleNonGenericSignatureOfCallChain(expr: CallChain) {
         const funcType = checkExpression(expr.expression);
         const nonOptionalType = getOptionalExpressionType(funcType, expr.expression);
-        const returnType = getNonGenericReturnTypeOfSingleCallSignature(funcType);
+        const returnType = getReturnTypeOfSingleNonGenericCallSignature(funcType);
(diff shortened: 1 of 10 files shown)


tests/baselines/reference/circularReferenceInReturnType.errors.txt
DanielRosenwasser, line 1: This is a minimal repro for the break, right?

DanielRosenwasser (approved):

tests/baselines/reference/circularReferenceInReturnType.errors.txt
Andarist: yes

src/compiler/checker.ts
```diff
@@ -39347,20 +39347,17 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         }
     }
 
-    function getNonGenericReturnTypeOfSingleCallSignature(funcType: Type) {
+    function getReturnTypeOfSingleNonGenericCallSignature(funcType: Type) {
         const signature = getSingleCallSignature(funcType);
-        if (signature) {
-            const returnType = getReturnTypeOfSignature(signature);
-            if (!signature.typeParameters || !couldContainTypeVariables(returnType)) {
```
Andarist, line 39354: I try to understand how to use `couldContainTypeVariables` appropriately and I wonder how this was different from how `checkFunctionExpressionOrObjectLiteralMethod` uses it here?
```ts
        // The identityMapper object is used to indicate that function expressions are wildcards
        if (checkMode && checkMode & CheckMode.SkipContextSensitive && isContextSensitive(node)) {
            // Skip parameters, return signature with return type that retains noncontextual parts so inferences can still be drawn in an early stage
            if (!getEffectiveReturnTypeNode(node) && !hasContextSensitiveParameters(node)) {
                // Return plain anyFunctionType if there is no possibility we'll make inferences from the return type
                const contextualSignature = getContextualSignature(node);
                if (contextualSignature && couldContainTypeVariables(getReturnTypeOfSignature(contextualSignature))) {
```

Andarist: I see how this is part of the inference algorithm and the check here was kinda not - but I still fail to understand how those 2 situations are fundamentally different so it's OK to call it within one context and not OK to call it within another.

jakebailey: I'm not sure that other use is correct either...

Andarist: I also tried to repro the regression caused by this cause I wanted to add a test case for it but I failed. It has been mentioned that it was only visible in the IDE but I assume that this particular report was reported by a bot reporting `tsc`-related changes. I couldn't repro with either though.

Issue #51473, opened 2022-11-10
__esModule in ts cause breaking change


Jack-Works:
In https://github.com/microsoft/TypeScript/pull/51387 the codebase changed to use ESBuild. And now `require('typescript')` will include `__esModule: true` and that is a breaking change.

```
> require("typescript").version
'4.9.2-rc'
> require("typescript").__esModule
undefined
>
```

```
> require('typescript').version
'5.0.0-dev.20221109'
> require('typescript').__esModule
true
>
```

This crashes https://github.com/hipstersmoothie/react-docgen-typescript-plugin

```
TypeError: Cannot read properties of undefined (reading 'JsxEmit')
    at DocgenPlugin.getOptions (node_modules/@storybook/react-docgen-typescript-plugin/dist/plugin.js:246:39)
```

It's output code uses `tslib.importDefault` to import typescript.

```ts
const typescript_1 = tslib_1.__importDefault(require("typescript"));
let compilerOptions = {
    jsx: typescript_1.default.JsxEmit.React,
    //   ~~~~~~~~~~~~~~~~~~~~ ts namespace before, undefined after
    module: typescript_1.default.ModuleKind.CommonJS,
    target: typescript_1.default.ScriptTarget.Latest,
};
```

jakebailey:
Thanks for testing this so early. This is unfortunate.

Can you try with `"typescript": "npm:@typescript-deploys/pr-build@5.0.0-pr-51474-2"` and see if the issue goes away? (from #51474)

Jack-Works:
yes, this fixes the problem

PR #51474, opened 2022-11-10
Fix __importDefault when used on typescript libraries

jakebailey:
Fixes #51473

PR #57163, opened 2024-01-25
Use 'export * from' for public API instead of 'export ='

jakebailey:
In #57133, I noticed that the emit for `export =` is actually worse for our bundle by a good bit; if we use `export * from`, things are less nested and smaller. Nothing changes for the public API as far as I'm aware.

### DIFF
--- a/Herebyfile.mjs
+++ b/Herebyfile.mjs
@@ -2,6 +2,7 @@
 import {
     CancelToken,
 } from "@esfx/canceltoken";
+import assert from "assert";
 import chalk from "chalk";
 import chokidar from "chokidar";
 import esbuild from "esbuild";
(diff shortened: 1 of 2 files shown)


jakebailey:
@typescript-bot perf test

jakebailey:
Ah, of course.

```
() => ts.version succeeded as expected.
() => ts.default.version failed as expected.
() => __importDefault(ts).version unexpectedly succeeded.
() => __importDefault(ts).default.version unexpectedly failed.
() => __importStar(ts).version succeeded as expected.
() => __importStar(ts).default.version unexpectedly failed.
```

In #51474 I actually did the reverse of this PR to fix #51473 and it's been long enough that I've forgotten.

I honestly don't understand this import helper stuff. I guess I could force `__esModule` back off again via another post-processing hack.

jakebailey:
I stuck this into #57133.

Issue #57166, opened 2024-01-25
Braceless JSDoc type tags fail to parse types starting with a parenthesis


Andarist:
### 🔎 Search Terms

jsdoc brace braceless curly parens paren parenthesis typetag function

### 🕗 Version & Regression Information

- This is the behavior in every version I tried

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.4.0-dev.20240124&filetype=js#code/PQKhAIAEBcE8AcCm4AUBKcBeAfOAztAE4CWAdgObgjACwAUAGYCupAxtMQPangOkCM6cAG964cIUTQmhHgBYATAG56AX3r1QEGAmRCc+ImUrV6zNh269SCoaLrjJ02eABEDTp1cq66uprAoOCRUAENCcgAuQxIKDAMCWJNaRhZ2Lh4+AGYUcPIMe0cpGR48nz8A7WC9YXBq6IByD04G8FVwAB8ROt1GgCNw1tUMADJu+EJOeGjSJgBbPsRCNqoU1m4CcE4+gCt+LG76t2bXABpwCano-gAGNp9KoN1UWqOmzyHOw97wBoHCIajcaTabgWYLJYrUx0dakTbbHYKA6vH6uTjQAAWSzOFxB1zuqh8QA

### 💻 Code

```ts
/** @type () => string */
function fn1() {
  return 42;
}

/** @type () => string */
function fn2() {
  return "foo";
}

/** @type (arg: string) => string */
function fn3(arg) {
  return arg;
}

/** @type ({ type: 'foo' } | { type: 'bar' }) & { prop: number } */
const obj1 = { type: "foo", prop: 10 };

/** @type ({ type: 'foo' } | { type: 'bar' }) & { prop: number } */
const obj2 = { type: "other", prop: 10 };
```

### 🙁 Actual behavior

All of those raise parse errors

### 🙂 Expected behavior

I'd expect them to be parsed correctly

### Additional information about the issue

`parseJSDocTypeExpression` gets called with `mayOmitBraces` equals to `true` but it still fails to parse this:
https://github.com/microsoft/TypeScript/blob/b7f691d9f13be987e445c7878a40190ca5808691/src/compiler/parser.ts#L8708

PR #57167, opened 2024-01-25
Fixed braceless type tags with types starting with an open parenthesis

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/57166


sandersn:
@typescript-bot test top200
@typescript-bot run dt

sandersn:
@jakebailey do you remember if jsdoc parsing shows up in perf results anymore?

Having a lot of tokens was a significant cause of slowdown in JS comment parsing when I profiled. So I don't want to add tokens unless it's for a good reason.

Is there a lot of un-braced usage of types? It's only supported in TS for compatibility, since the braces are supposed to be required.

jakebailey:

jakebailey:
> @jakebailey do you remember if jsdoc parsing shows up in perf results anymore?

No, not really.

Andarist:
> Is there a lot of un-braced usage of types? It's only supported in TS for compatibility, since the braces are supposed to be required.

Right. But since they are optional in this context I'd expect this to work (there is an explicit boolean set here that allows it). I don't want this for myself - I'm not using JSDoc types anyway. It's just a problem I noticed when working on https://github.com/microsoft/TypeScript/pull/57110 and such a braceless type came there (in the original issue) from the real-world code.

sandersn (approved):
I'm not convinced that this is very valuable, but it does fix an inconsistency, and it shouldn't hurt the fast path for scanning jsdoc comment text.

sandersn:
Can you update this PR from main before I merge? There has been a lot of churn in baselines recently.

Andarist:
@sandersn done

Issue #57173, opened 2024-01-25
Unflagged used before its declaration for const enum when isolatedModules is enabled


frigus02:
### 🔎 Search Terms

isolatedModules, const enum, used before its declaration

### 🕗 Version & Regression Information

- This changed between versions v5.0.4 and v5.1.6
- I think this changed in https://github.com/microsoft/TypeScript/pull/53625

### ⏯ Playground Link

https://www.typescriptlang.org/play?isolatedModules=true&ts=5.1.6#code/FAYw9gdgzmA2CmA6WYDmAKAYmMiCCAlANzCiRQAuABPBAK4C2V2YVA3lXgDRUBCVAX2BA

### 💻 Code

```ts
console.log(Foo.A);
const enum Foo { A, B }
```

### 🙁 Actual behavior

JS emit with `isolatedModules` is enabled:

```ts
console.log(Foo.A);
var Foo;
(function (Foo) {
    Foo[Foo["A"] = 0] = "A";
    Foo[Foo["B"] = 1] = "B";
})(Foo || (Foo = {}));
```

No error.

### 🙂 Expected behavior

Expected "used before its declaration error" because the const enum is not inlined.

Or ... could TypeScript still inline the const enum since it's defined in the same file? I guess that might be too expensive?

### Additional information about the issue

I profusely apologize. I think the behavior was changed due to https://github.com/microsoft/TypeScript/issues/53019. I reported that the error should not happen if `preserveConstEnums` is enabled because the const enum is still inlined then. I didn't test `isolatedModules` then. I suspect the error was correct, it was just gated by the wrong flag.

PR #57174, opened 2024-01-25
Reissue used before declaration error for const enums in `isolatedModules`

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/57173

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -3926,7 +3926,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             }
             else {
                 Debug.assert(!!(result.flags & SymbolFlags.ConstEnum));
-                if (compilerOptions.isolatedModules) {
+                if (getIsolatedModules(compilerOptions)) {
                     diagnosticMessage = error(errorLocation, Diagnostics.Enum_0_used_before_its_declaration, declarationName);
                 }
             }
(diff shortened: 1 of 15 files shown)


src/compiler/checker.ts
```diff
@@ -3924,6 +3924,12 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             else if (result.flags & SymbolFlags.RegularEnum) {
                 diagnosticMessage = error(errorLocation, Diagnostics.Enum_0_used_before_its_declaration, declarationName);
             }
+            else {
+                Debug.assert(!!(result.flags & SymbolFlags.ConstEnum));
+                if (compilerOptions.isolatedModules) {
```
andrewbranch, line 3929:
```suggestion
                if (getIsolatedModules(compilerOptions)) {
```

andrewbranch (changes requested):

src/compiler/checker.ts
andrewbranch: I would commit this suggestion for you and approve, but GitHub can't handle this on checker.ts 🥲

Andarist: 🫠

I pushed out the change + the relevant missing test case for `verbatimModuleSyntax` :p

andrewbranch (approved):
Thanks!

PR #57176, opened 2024-01-25
Fix missing error for accessing type-only `export *` member through namespace

### LINKED ISSUE #56648: No error when referencing a value that was re-exported with `export type *`
### 🔎 Search Terms

"class export type", "class re-export type", "enum export type", "namespace import type export"

- This changed between versions 4.9.5 and 5.04 (introduction of `export type *`)

### ⏯ Playground Link

https://www.typescriptlang.org/dev/bug-workbench/?ts=5.4.0-dev.20231202&ssl=4&ssc=39&pln=4&pc=1#code/PTAEAEDMEsBsFMB2BDAtvAXKVzqIHQAuAzgFDSoAOA9gE6GgBUoyxoeh8t6AJtMp1CRa1VKADk+YBy69+ncaVIBjaomIMA5gAtqG0AF5QieAHd2iTt3h8B8fAHFdGgBQBKJSAgwEKdFhlrW04iMngADxp6UEIAT0p4JiERMUlgHT1CRVIvKDgkNExQDI1Q0giohmVYVjYnTNAAbwBfUiA

### 💻 Code

```ts
// @filename: main.ts
import * as intermediate from './intermediate'

const ghost = new intermediate.Ghost()

// @filename: intermediate.ts
export type * from './ghost'

// @filename: ghost.ts
export class Ghost {}
```

### 🙁 Actual behavior

The code compiles without errors and fails at run-time.

### 🙂 Expected behavior

A compiler error at the constructor call.

The issue only occurs when using a namespace import and `export type *`. Explicit `{ Ghost }` imports and/or re-exports are correctly recognized as type only. The issue also affects enums.

andrewbranch:
Fixes #56648

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -14908,7 +14908,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         if (type.flags & TypeFlags.Object) {
             const resolved = resolveStructuredTypeMembers(type as ObjectType);
             const symbol = resolved.members.get(name);
-            if (symbol && !includeTypeOnlyMembers && type.symbol.flags & SymbolFlags.ValueModule && getSymbolLinks(type.symbol).typeOnlyExportStarMap?.has(name)) {
+            if (symbol && !includeTypeOnlyMembers && type.symbol?.flags & SymbolFlags.ValueModule && getSymbolLinks(type.symbol).typeOnlyExportStarMap?.has(name)) {
                 // If this is the type of a module, `resolved.members.get(name)` might have effectively skipped over
                 // an `export type * from './foo'`, leaving `symbolIsValue` unable to see that the symbol is being
                 // viewed through a type-only export.
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -14908,6 +14908,12 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         if (type.flags & TypeFlags.Object) {
             const resolved = resolveStructuredTypeMembers(type as ObjectType);
             const symbol = resolved.members.get(name);
+            if (symbol && !includeTypeOnlyMembers && type.symbol.flags & SymbolFlags.ValueModule && getSymbolLinks(type.symbol).typeOnlyExportStarMap?.has(name)) {
+                // If this is the type of a module, `resolved.members.get(name)` might have effectively skipped over
+                // an `export type * from './foo'`, leaving `symbolIsValue` unable to see that the symbol is being
+                // viewed through a type-only export.
+                return undefined;
+            }
             if (symbol && symbolIsValue(symbol, includeTypeOnlyMembers)) {
                 return symbol;
             }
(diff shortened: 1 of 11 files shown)


tests/baselines/reference/exportNamespace11.errors.txt
```diff
@@ -0,0 +1,14 @@
+main.ts(2,52): error TS2339: Property 'Ghost' does not exist on type 'typeof import("intermediate")'.
+
+
+==== main.ts (1 errors) ====
+    import * as intermediate from './intermediate'
+    const ghost: intermediate.Ghost = new intermediate.Ghost()
+                                                       ~~~~~
+!!! error TS2339: Property 'Ghost' does not exist on type 'typeof import("intermediate")'.
```
andrewbranch, line 8: It would be nice to be able to give a better type-only-specific error message here, but this is consistent with existing error messages you can get with other ways of doing property access on a module with type-only exports.

src/compiler/checker.ts
```diff
@@ -14908,6 +14908,12 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         if (type.flags & TypeFlags.Object) {
             const resolved = resolveStructuredTypeMembers(type as ObjectType);
             const symbol = resolved.members.get(name);
+            if (symbol && !includeTypeOnlyMembers && type.symbol?.flags & SymbolFlags.ValueModule && getSymbolLinks(type.symbol).typeOnlyExportStarMap?.has(name)) {
```
sandersn, line 14911: I don't understand the check for `!includeTypeOnlyMembers`. If this is from a `import * as intermediate from './x.js'`, the type doesn't indicate that type-only members should be skipped. Where does this come from?

Actually, the overall intent is to skip values that show up in the type-only exports map, right? How does this code distinguish a value (which shouldn't be included) from a type (which should (I think?))? Maybe that waits until the check just below the new one.

sandersn (changes requested):
This code is likely right, but I realised I'm missing background knowledge on how type-only imports work in the first place, so I asked a couple of questions.

src/compiler/checker.ts
andrewbranch: This function gets _properties_ which need to be values by definition. The question is whether the value has to exist in the emitted JavaScript or not. The only callers that pass `includeTypeOnlyMembers: true` are type queries and import type nodes:

```ts
type T = typeof x.y.z; // 'y' may be a namespace made by `export type * from "y"`
type U = typeof import("x").y.z;
```

These are constructs that query values, but don’t rely on those values actually existing, so they can safely return value symbols that are only accessible via type-only imports/exports.

sandersn: Thanks, that example helped a lot. So the namespace 'y' can be, almost, a synthetic property that should only be visible from type space?

sandersn (approved):

src/compiler/checker.ts
andrewbranch: I didn’t show where x/y/z were declared, but the way I intended the example was that `y` itself is real (it’s the module namespace object of a file that does `export type *` but maybe has no real exports—so it would be an empty object), while `z` is the thing that acts like a “synthetic property” on `y`. But I think you understood the idea.

andrewbranch:
@typescript-bot perf test
@typescript-bot test top200
@typescript-bot run dt

PR #57183, opened 2024-01-26
Make checker into a var

jakebailey:
Noticed that this wasn't a var; this is passed in every time we make a new Type or Signature, so maybe it matters?


jakebailey:

PR #49568, opened 2022-06-15
Fix renaming of node_modules

gabritto:
Fixes #45668.
I think it also fixes #48681.

The rule I'm trying to implement is the following:
1. If we try to rename from a project, and that would cause us to rename something in `node_modules/`, don't allow that rename.
2. If we try to rename from inside a `node_modules` package, only allow renames that would not affect a different package.

I'd like some feedback on the wording in the diagnostic message.
I'll update the tests' `verify.renameInfoFails()` calls once there's a definitive diagnostic message.

Note: I had to change `getRenameInfo` to take a user preferences parameter, because I needed to know if `providePrefixAndSuffixTextForRename` (surfaced as "Use Aliases For Renames" in vscode) is set or not. Since that function was already taking as parameter another option that comes from user preferences, it made sense to me to change that. A lot of the diff is refactoring related to that change.

PR #57201, opened 2024-01-28
fix(57022): Rename of string literal property name doesn't rename all occurences

### LINKED ISSUE #57022: Rename of string literal property name doesn't rename all occurences
### 🔎 Search Terms

Rename string literal

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about rename

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.3.3#code/C4TwDgpgBACgTgezAZygXigbwLACgoFQBmCCAXFAEakA2EAhgHYDceAvnnncFsaVGwrwkqDIwCuNGlHqomIVrhIJmUAPRqoAEwQRUjBDwDmEHnAiN6AWwhbOubjORahiFOl4AiZZ4rA44tBsqhpQ5pY2fAhQABYQ5kA

### 💻 Code

```ts
type Props = {
    foo: boolean;
}

let { foo }: Props = null as any;
foo; // does not get renamed

let asd: Props = { "foo": true }; // rename foo here
```

### 🙁 Actual behavior

When putting the cursor inside the `"foo"` string prop literal and then renaming it to something else, it will not rename the `foo;` occurence. This is it how it looks like after renaming:

```ts
type Props = {
    bar: boolean;
}

let { bar }: Props = null as any;
foo;

let asd: Props = { "bar": true };
```

Note that you need to try this locally, in the playground it works. It does work if the property is not a string literal.

### 🙂 Expected behavior

All occurences are renamed as expected. This would be the expected output:
```ts
type Props = {
    bar: boolean;
}

let { bar }: Props = null as any;
bar;

let asd: Props = { "bar": true };
```

We're using string literal properties quite heavily for various reasons in `svelte2tsx` which powers the Svelte intellisense, through which I noticed this.

Zzzen:
… occurences

Fixes #57022

### DIFF
--- a/src/services/findAllReferences.ts
+++ b/src/services/findAllReferences.ts
@@ -757,7 +757,7 @@ interface PrefixAndSuffix {
     readonly suffixText?: string;
 }
 function getPrefixAndSuffixText(entry: Entry, originalNode: Node, checker: TypeChecker, quotePreference: QuotePreference): PrefixAndSuffix {
-    if (entry.kind !== EntryKind.Span && isIdentifier(originalNode)) {
+    if (entry.kind !== EntryKind.Span && (isIdentifier(originalNode) || isStringLiteralLike(originalNode))) {
         const { node, kind } = entry;
         const parent = node.parent;
         const name = originalNode.text;
--- a/tests/baselines/reference/renameStringPropertyNames2.baseline.jsonc
+++ b/tests/baselines/reference/renameStringPropertyNames2.baseline.jsonc
@@ -1,0 +1,10 @@
+// === findRenameLocations ===
+// === /tests/cases/fourslash/renameStringPropertyNames2.ts ===
+// type Props = {
+//   <|[|fooRENAME|]: boolean;|>
+// }
+// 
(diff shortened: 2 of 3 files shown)


iisaduan:
@typescript-bot pack this

iisaduan:
This PR doesn't quite do what the original bug report was asking for, however, I think it is the correct behavior that is in line with other renames that we offer. The `foo` immediately following the destructuring assignment is a separate variable than the one referred to by the string literal property.


Before this PR: rename only works when the property being renamed is not a string literal.
image
image

After this PR: `foo` is now properly assigned after renaming the property
image

Renaming the `foo` that doesn't change: the renamed variable is assigned properly
image

andrewbranch (approved):

iisaduan (approved):

iisaduan:
@gabritto pointed out that you can make the behavior match the playground behavior by setting `providePrefixAndSuffixTextForRename` (or "Use Aliases For Renames" in vscode) to false. https://github.com/microsoft/TypeScript/pull/49568#issue-1272837630

PR #53066, opened 2023-03-02
Always substitute indexed generic mapped type when getting constraint from indexed access

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/50030

PR #57202, opened 2024-01-28
Revert "Always substitute indexed generic mapped type when getting constraint from indexed access"

Andarist:
reverts my own https://github.com/microsoft/TypeScript/pull/53066 , fixes https://github.com/microsoft/TypeScript/pull/53066#issuecomment-1913384757

cc @ahejlsberg @weswigham

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -14230,7 +14230,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     }
 
     function getConstraintFromIndexedAccess(type: IndexedAccessType) {
-        if (isMappedTypeGenericIndexedAccess(type) || isGenericMappedType(type.objectType)) {
+        if (isMappedTypeGenericIndexedAccess(type)) {
             // For indexed access types of the form { [P in K]: E }[X], where K is non-generic and X is generic,
             // we substitute an instantiation of E where P is replaced with X.
             return substituteIndexedMappedType(type.objectType as MappedType, type.indexType);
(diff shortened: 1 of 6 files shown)


src/compiler/checker.ts
```diff
@@ -14230,7 +14230,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     }
 
     function getConstraintFromIndexedAccess(type: IndexedAccessType) {
-        if (isMappedTypeGenericIndexedAccess(type) || isGenericMappedType(type.objectType)) {
```
Andarist, line 14233: while my comment [here](https://github.com/microsoft/TypeScript/pull/53066#discussion_r1123593187) can be partially true - so far I couldn't find the situation in which this extra check would benefit anything (after adjustments) since `getSimplifiedIndexedAccessType` already performs this substitution when it's safe to do so

tests/cases/conformance/types/mapped/mappedTypeConstraints2.ts
```diff
@@ -57,5 +83,5 @@ type ObjectWithUnderscoredKeys<K extends string> = {
 };
 
 function genericTest<K extends string>(objectWithUnderscoredKeys: ObjectWithUnderscoredKeys<K>, key: K) {
-  const shouldBeTrue: true = objectWithUnderscoredKeys[`_${key}`];
+  const shouldBeTrue: true = objectWithUnderscoredKeys[`_${key}`]; // assignability fails here, but ideally should not
```
Andarist, line 86: Those kinds of cases could always be OK if we could determine that the template type doesn't depend on the original `K`. I don't think the compiler is equipped to check this sort of thing right now though.

weswigham:
@DanielRosenwasser I don't think this revert needs to be cherry-picked or anything, right? Since we'll sync the release branch with `main` again for the rc.

DanielRosenwasser:
Yeah, it'll be in the RC

Issue #52666, opened 2023-02-08
[Playground] two slash `@filename` not work in typescript playground


magic-akari:
# Bug Report

### 🔎 Search Terms

- playground
- two slash
- filename

### 🕗 Version & Regression Information

- This is a crash
- This changed between versions ______ and _______
- This is the behavior in every version I tried, and I reviewed the FAQ for entries about _________
- I was unable to test this on prior versions because _______

### ⏯ Playground Link
[Playground link with relevant code](https://www.typescriptlang.org/play?ts=4.9.5#code/PTAEAEDMEsBsFMB2BDAtvAXKGCC0B3aAFwAtd4APABwHsAnIgOiIGcAoS2h0AYxsRZFQJeLFg0A6vVgATUAF5QAIgCiFNFQShBdaIgDmStmxAQcSNJlB6ZlZu2iouQgN7DR4qXVmgAvtjoaVGVGYHMCYjJOeiIjPgEaBEZxfQAKETFJaRkASiA)

### 💻 Code

Code from https://github.com/microsoft/TypeScript-Website/tree/v2/packages/ts-twoslasher#import_filests

```ts
// @filename: file-with-export.ts
export const helloWorld = "Example string"

// @filename: index.ts
import { helloWorld } from "./file-with-export"
console.log(helloWorld)
```

### 🙁 Actual behavior

TS2307: Cannot find module './file-with-export' or its corresponding type declarations.

### 🙂 Expected behavior

It works.

andrewbranch:
Multiple files aren’t supported on the playground.

uloco:
@andrewbranch You should consider updating the help docs on ts-playground then, it has an extra entry regarding multi-file playgrounds: https://www.typescriptlang.org/play#handbook-15

DerGernTod:
please consider reopening this. this used to work, otherwise it wouldn't be documented. it also behaves very weird if you use it by showing a default "/input.tsx":
![image](https://user-images.githubusercontent.com/19726130/228853705-def5de70-9c1a-45bd-9468-328d237adb7a.png)

RyanCavanaugh:
> this used to work

We don't intend to support this in the playground. You can use the Bug Workbench to do multi-file repros.

jsejcksn:
> We don't intend to support this in the playground. You can use the Bug Workbench to do multi-file repros.

[...]

PR #57207, opened 2024-01-28
[WIP] feat(41825): JSDoc equivalent of import *

### LINKED ISSUE #41825: JSDoc equivalent of import *
## Search Terms

- jsdoc import
- import *
- import namespace 
- jsdoc namespace
## Suggestion

Provide a JSDoc equivalent of `import * as Sub from "./sub"` so that types & interfaces from the `./sub` module could be referenced as `Sub.Type`, `Sub.Interface`.

I have no preference in terms of actual syntax, but in my experience general expectation had been that following syntax should do just that (but it is not and I can't seem to figure out what does `Sub` results in here:

```ts
@typedef {import('./sub')} Sub
```

If above is not a viable options maybe using same `*` character could be an options:

```ts
@typedef {import('./sub').*} Sub
```

Or maybe whole new `@import` jsdoc tag. 

## Use Cases

js-ipfs (fairly large code base) has adopted TS via JSDoc syntax but dealing with large number of imported types & interfaces from other modules had been a major source of pain for following reasons:

1. Every single type requires `@typedef {import('...').Name} Alias`.
2. Importing generic types / interfaces requires repeating type parameters via `@template` tags
  - Which is error prone because when omitted it turns into `any`
  - Ends up require lot of typing
3. Changes to library layout requires updating all those typdefs all over the code base (and vscode can't help there sadly)
   - Alternative is to consolidating them in one place, but then vscode (and alike) takes multiple hops / clicks to get you to actual  type definition.

All of the above make otherwise mostly good experience to be painful. 

Current approach also has side effect of turning imported interfaces into types (see #41258) which than can't be used in `implements` syntax as they are no longer interfaces.

Given that it is possible to do `import * as Sub from "./sub"` to get a namespace of exports in TS syntax it seems like equivalent in jsdoc syntax would:

1. Allow reduced number of `@typedef {import(...).Name} Name` declarations.
2. Allow importing generic types without having to retype type parameters via `@temaplate` tags.
3. Avoid turning interfaces.

## Examples

Consider following TS code:

```ts
import { BlockEncoder, BlockDecoder } from "@multiformats/codecs"
import { DagNode } form "./node"

class DagPB implements BlockEncoder<0x70, DagNode>, BlockDecoder<0x70, DagNode> {
  async encode(node:DagNode):Promise<Uint8Array> {
     // ...
  }
  async decode(bytes:Uint8Array):Promise<DagNode> {
    // ...
  }
}
```

Same code with JSDoc syntax turns into following:

```js
/**
 * @template {number} Code
 * @template T
 * @typedef {import('@multiformats/codecs').BlockEncoder<Code, T>} BlockEncoder
 */
/**
 * @template {number} Code
 * @template T
 * @typedef {import('@multiformats/codecs').BlockDecoder<Code, T>} BlockDecoder
 */
/**
  * @typedef {import("./node").DagNode} DagNode
  */

/**
 * @implements {BlockEncoder<0x70, DagNode>}
 * @implements {BlockDecoder<0x70, DagNode>}
 */
class DagPB {
  /**
   * @param {DagNode} node
   * @returns {Promise<Uint8Array>}
   */
  async encode(node) {
     // ...
  }
  /**
   * @param {Uint8Array} bytes
   * @returns {Promise<DagNode>}
   */
  async decode(bytes) {
    // ...
  }
}
```

**Note**: Above code won't even produce desired typedefs due to #41258

And now if we had what this issue proposes it would be:

```js
/**
 * @typedef {import('@multiformats/codecs')} Codec
 * @typedef {import('./node').DagNode} DagNode
 */

/**
 * @implements {Codec.BlockEncoder<0x70, DagNode>}
 * @implements {Codec.BlockDecoder<0x70, DagNode>}
 */
class DagPB {
  /**
   * @param {DagNode} node
   * @returns {Promise<Uint8Array>}
   */
  async encode(node) {
     // ...
  }
  /**
   * @param {Uint8Array} bytes
   * @returns {Promise<DagNode>}
   */
  async decode(bytes) {
    // ...
  }
}
```
## Checklist

a-tarasyuk:
Fixes #41825
Fixes https://github.com/microsoft/TypeScript/issues/41825#issuecomment-1474529135

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -11,8 +11,8 @@ import {
     AmbientModuleDeclaration,
     and,
     AnonymousType,
+    AnyImportOrJsDocImportTypeImport,
     AnyImportOrReExport,
-    AnyImportSyntax,
     append,
     appendIfUnique,
     ArrayBindingPattern,
@@ -583,6 +583,7 @@ import {
     isJSDocCallbackTag,
     isJSDocConstructSignature,
     isJSDocFunctionType,
+    isJSDocImportTypeTag,
     isJSDocIndexSignature,
     isJSDocLinkLike,
     isJSDocMemberName,
@@ -770,6 +771,7 @@ import {
     JSDocEnumTag,
     JSDocFunctionType,
     JSDocImplementsTag,
+    JSDocImportTypeTag,
     JSDocLink,
     JSDocLinkCode,
     JSDocLinkPlain,
@@ -3942,7 +3944,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             || (n === stopAt || isFunctionLike(n) && (!getImmediatelyInvokedFunctionExpression(n) || (getFunctionFlags(n) & FunctionFlags.AsyncGenerator)) ? "quit" : false));
     }
 
-    function getAnyImportSyntax(node: Node): AnyImportSyntax | undefined {
+    function getAnyImportSyntax(node: Node): AnyImportOrJsDocImportTypeImport | undefined {
         switch (node.kind) {
             case SyntaxKind.ImportEqualsDeclaration:
                 return node as ImportEqualsDeclaration;
@@ -4280,8 +4282,8 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         }
     }
 
-    function getExternalModuleMember(node: ImportDeclaration | ExportDeclaration | VariableDeclaration, specifier: ImportOrExportSpecifier | BindingElement | PropertyAccessExpression, dontResolveAlias = false): Symbol | undefined {
-        const moduleSpecifier = getExternalModuleRequireArgument(node) || (node as ImportDeclaration | ExportDeclaration).moduleSpecifier!;
+    function getExternalModuleMember(node: ImportDeclaration | ExportDeclaration | VariableDeclaration | JSDocImportTypeTag, specifier: ImportOrExportSpecifier | BindingElement | PropertyAccessExpression, dontResolveAlias = false): Symbol | undefined {
+        const moduleSpecifier = getExternalModuleRequireArgument(node) || (node as ImportDeclaration | ExportDeclaration | JSDocImportTypeTag).moduleSpecifier!;
         const moduleSymbol = resolveExternalModuleName(node, moduleSpecifier)!; // TODO: GH#18217
         const name = !isPropertyAccessExpression(specifier) && specifier.propertyName || specifier.name;
         if (!isIdentifier(name)) {
@@ -5006,6 +5008,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             ? location
             : (isModuleDeclaration(location) ? location : location.parent && isModuleDeclaration(location.parent) && location.parent.name === location ? location.parent : undefined)?.name ||
                 (isLiteralImportTypeNode(location) ? location : undefined)?.argument.literal ||
+                (isInJSFile(location) && isJSDocImportTypeTag(location) ? location.moduleSpecifier : undefined) ||
                 (isVariableDeclaration(location) && location.initializer && isRequireCall(location.initializer, /*requireStringLiteralLikeArgument*/ true) ? location.initializer.arguments[0] : undefined) ||
                 findAncestor(location, isImportCall)?.arguments[0] ||
                 findAncestor(location, isImportDeclaration)?.moduleSpecifier ||
@@ -9720,12 +9723,13 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     case SyntaxKind.ImportClause: {
                         const generatedSpecifier = getSpecifierForModuleSymbol(target.parent || target, context); // generate specifier (even though we're reusing and existing one) for ambient module reference include side effects
                         const specifier = bundled ? factory.createStringLiteral(generatedSpecifier) : (node as ImportClause).parent.moduleSpecifier;
+                        const attributes = isImportDeclaration(node.parent) ? node.parent.attributes : undefined;
                         addResult(
                             factory.createImportDeclaration(
                                 /*modifiers*/ undefined,
                                 factory.createImportClause(/*isTypeOnly*/ false, factory.createIdentifier(localName), /*namedBindings*/ undefined),
                                 specifier,
-                                (node as ImportClause).parent.attributes,
+                                attributes,
                             ),
                             ModifierFlags.None,
                         );
@@ -9734,12 +9738,13 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     case SyntaxKind.NamespaceImport: {
                         const generatedSpecifier = getSpecifierForModuleSymbol(target.parent || target, context); // generate specifier (even though we're reusing and existing one) for ambient module reference include side effects
                         const specifier = bundled ? factory.createStringLiteral(generatedSpecifier) : (node as NamespaceImport).parent.parent.moduleSpecifier;
+                        const attributes = isImportDeclaration(node.parent) ? node.parent.attributes : undefined;
                         addResult(
                             factory.createImportDeclaration(
                                 /*modifiers*/ undefined,
                                 factory.createImportClause(/*isTypeOnly*/ false, /*name*/ undefined, factory.createNamespaceImport(factory.createIdentifier(localName))),
                                 specifier,
-                                (node as ImportClause).parent.attributes,
+                                attributes,
                             ),
                             ModifierFlags.None,
                         );
@@ -9759,6 +9764,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     case SyntaxKind.ImportSpecifier: {
                         const generatedSpecifier = getSpecifierForModuleSymbol(target.parent || target, context); // generate specifier (even though we're reusing and existing one) for ambient module reference include side effects
                         const specifier = bundled ? factory.createStringLiteral(generatedSpecifier) : (node as ImportSpecifier).parent.parent.parent.moduleSpecifier;
+                        const attributes = isImportDeclaration(node.parent.parent.parent) ? node.parent.parent.parent.attributes : undefined;
                         addResult(
                             factory.createImportDeclaration(
                                 /*modifiers*/ undefined,
@@ -9774,7 +9780,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                                     ]),
                                 ),
                                 specifier,
-                                (node as ImportSpecifier).parent.parent.parent.attributes,
+                                attributes,
                             ),
                             ModifierFlags.None,
                         );
--- a/src/compiler/parser.ts
+++ b/src/compiler/parser.ts
@@ -175,6 +175,7 @@ import {
     JSDocEnumTag,
     JSDocFunctionType,
     JSDocImplementsTag,
+    JSDocImportTypeTag,
     JSDocLink,
     JSDocLinkCode,
     JSDocLinkPlain,
@@ -1125,6 +1126,7 @@ const forEachChildTable: ForEachChildTable = {
     [SyntaxKind.JSDocReadonlyTag]: forEachChildInJSDocTag,
     [SyntaxKind.JSDocDeprecatedTag]: forEachChildInJSDocTag,
     [SyntaxKind.JSDocOverrideTag]: forEachChildInJSDocTag,
+    [SyntaxKind.JSDocImportTypeTag]: forEachChildInJSDocImportTypeTag,
     [SyntaxKind.PartiallyEmittedExpression]: forEachChildInPartiallyEmittedExpression,
 };
 
@@ -1211,6 +1213,13 @@ function forEachChildInJSDocLinkCodeOrPlain<T>(node: JSDocLink | JSDocLinkCode |
 
 function forEachChildInJSDocTag<T>(node: JSDocUnknownTag | JSDocClassTag | JSDocPublicTag | JSDocPrivateTag | JSDocProtectedTag | JSDocReadonlyTag | JSDocDeprecatedTag | JSDocOverrideTag, cbNode: (node: Node) => T | undefined, cbNodes?: (nodes: NodeArray<Node>) => T | undefined): T | undefined {
     return visitNode(cbNode, node.tagName)
+        || (typeof node.comment === "string" ? undefined : visitNodes(cbNode, cbNodes, node.comment));
+}
+
+function forEachChildInJSDocImportTypeTag<T>(node: JSDocImportTypeTag, cbNode: (node: Node) => T | undefined, cbNodes?: (nodes: NodeArray<Node>) => T | undefined): T | undefined {
+    return visitNode(cbNode, node.tagName)
+        || visitNode(cbNode, node.importClause)
+        || visitNode(cbNode, node.moduleSpecifier)
         || (typeof node.comment === "string" ? undefined : visitNodes(cbNode, cbNodes, node.comment));
 }
 
@@ -9073,6 +9082,9 @@ namespace Parser {
                     case "throws":
                         tag = parseThrowsTag(start, tagName, margin, indentText);
                         break;
+                    case "importType":
+                        tag = parseImportTypeTag(start, tagName, margin, indentText);
+                        break;
                     default:
                         tag = parseUnknownTag(start, tagName, margin, indentText);
                         break;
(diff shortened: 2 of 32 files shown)


src/compiler/parser.ts
```diff
@@ -9445,6 +9457,31 @@ namespace Parser {
...
+                    || token() === SyntaxKind.OpenBraceToken // @importType {
+                ) {
+                    importClause = parseImportClause(identifier, afterImportTypeTagPos, /*isTypeOnly*/ true);
+                    parseExpected(SyntaxKind.FromKeyword);
+                }
+
+                Debug.assert(importClause);
+
+                const moduleSpecifier = parseModuleSpecifier();
+                const comments = margin !== undefined && indentText !== undefined ? parseTrailingTagComments(start, getNodePos(), margin, indentText) : undefined;
```
DanielRosenwasser, line 9481: Do we actually need to parse comments out here? @sandersn?

DanielRosenwasser:
@typescript-bot pack this

DanielRosenwasser:
Something we should test

```js
/**
 * @importType
 * as Foo from "foo"
 */
```

and

/**
 * @importType {
 *   A,
 *   B,
 *   C,
 * } from "foo"
 */

DanielRosenwasser:
In our design meeting, I brought up some of the discussion [on Twitter](https://twitter.com/drosenwasser/status/1752435501444034986) around the tag name.

I think the team is up for just calling it `@import`.

Something that came up was the use of `type` modifiers on import specifiers.

```js
/**
 * @import { type Foo } from "foo"
 */
```

The preference was to disallow them for now, even though it might make copy/pasting from TS to JS easier. The idea was can always relax that restriction.

Another two things that I do want us to test:

1. Auto-imports - whether it works, where they get inserted in an import list, etc. This is kind of a big task
2. Declaration file generation - precisely how a `/** @import */` comment translates into a TS import within a `.d.ts` file.

a-tarasyuk:
> As raised above, multi-line should just work:

```js
/**
 * @import {
 *   A,
 *   B,
 *   C,
 * } from "foo"
 */
```

@DanielRosenwasser I think the `scanner` can handle this case. However, I think we need to change the naming for this option :). 

https://github.com/microsoft/TypeScript/blob/821b1d8fdfe5212563f18dd90c9b4d846f716432/src/compiler/scanner.ts#L1922-L1926

a-tarasyuk:
> Auto-imports - whether it works, where they get inserted in an import list, etc.

@DanielRosenwasser The existing auto-import logic will add `@param { import("./a").A }` for the following case., Should we consider changing this logic to use `@import`?

```js
// @Filename: /a.ts
export default interface A { }

// @Filename: /test.js
/**
 * @param { A/**/ } a
 */
export function f(a) { }
```

a-tarasyuk:
```js
/**
 * @import
 * as Foo from "foo"
 ^
 what is this?
 */

```

@DanielRosenwasser  It's a tricky case because the first `*` in the JSDoc is a _decoration_. So, if we handle JSDoc like other tags, the correct way to use multiline `import *` should be...

```js
/**
 * @import
 * * as Foo from "foo"
 */
```

DanielRosenwasser:
> Should we consider changing this logic to use `@import`?

I would say so - however, in the interest of keeping this PR smaller, you might want to do a follow-up PR instead. @andrewbranch might have thoughts here.

_____

Regarding the ambiguity with `*`, I am okay with either decision. My personal suggestion is to say that the next token following a JSDoc import has to occur on the same line. @sandersn might have thoughts here.

a-tarasyuk:
> Declaration file generation - precisely how a /** @import */ comment translates into a TS import within a .d.ts file.

@DanielRosenwasser These imports are currently transformed into `ImportTypeNode` (`{import("./types").A}`). Is it better to translate import tags as top-level declarations within the `checker`? https://github.com/microsoft/TypeScript/blob/c790dc1dc7ff67e619a5a60fc109b7548f171322/src/compiler/transformers/declarations.ts#L440

andrewbranch:
Since these have type-only import semantics in JSDoc, I would have assumed they would translate to type-only import declarations in the .d.ts file. I’m not sure it actually matters though.

sandersn (commented):
Not done reviewing, but I found that I wrote some requests for tests last time I looked at this and forgot to post them.

sandersn:
> Regarding the ambiguity with *, I am okay with either decision. My suggestion is to say that the next token following a JSDoc import has to occur on the same line. @sandersn might have thoughts here.

I'm also OK with any decision, and I think strict no-newline one is a good place to start *except* that it would require special-casing `parseTag`, which currently uses
```ts
const tagName = parseJSDocIdentifierName(/*message*/ undefined);
const indentText = skipWhitespaceOrAsterisk();
```

So I'd personally leave it as interpreting a leading asterisk as decoration:

```js
/**
 * @import
 * { foo, bar }
 * from "multiline.js"
 */
```

equivalent to `import { foo, bar } from 'multiline.js'`. I don't think it's worth any additional code. Just test that

```js
/**
 * @import
 * * as multiline from 'multiline.js'
 */
```

works.

src/services/completions.ts
```diff
@@ -3210,25 +3213,30 @@ function getCompletionData(
...
+                    }
+                }
+                if (!insideJsDocTagTypeExpression && isJSDocParameterTag(tag) && (nodeIsMissing(tag.name) || tag.name.pos <= position && position <= tag.name.end)) {
+                    return { kind: CompletionDataKind.JsDocParameterName, tag };
+                }
             }
         }
 
-        if (!insideJsDocTagTypeExpression) {
+        if (!insideJsDocTagTypeExpression && !insideJsDocImportTypeTag) {
```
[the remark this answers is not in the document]
a-tarasyuk: [jsdocImportTagCompletion1](https://github.com/microsoft/TypeScript/pull/57207/files#diff-f8b28d8b891d75f8b93a45cc5d1616d0e2baa765cf666d5f771f80abfad5b31d
), [jsdocImportTagCompletion2](https://github.com/microsoft/TypeScript/pull/57207/files#diff-2c07de5ce4fcf249b72f091f2106722534f38aeaf09b58bd752f92b093adb3d8) Are there any additional cases that require coverage?

src/services/stringCompletions.ts
```diff
@@ -418,6 +418,7 @@ function getStringLiteralCompletionEntries(sourceFile: SourceFile, node: StringL
         case SyntaxKind.ImportDeclaration:
         case SyntaxKind.ExportDeclaration:
         case SyntaxKind.ExternalModuleReference:
+        case SyntaxKind.JSDocImportTypeTag:
```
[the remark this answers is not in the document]
a-tarasyuk: [jsdocImportTagCompletion3](https://github.com/microsoft/TypeScript/pull/57207/files#diff-852631547cb9f0fe5fb1f7095d1cf7c31f529a825984a3d657b742bf799fa2a4)

src/compiler/parser.ts
```diff
@@ -9445,6 +9457,31 @@ namespace Parser {
...
+                    identifier = parseIdentifier();
+                }
+
+                let importClause: ImportClause | undefined;
+                if (
+                    identifier // @importType id
+                    || token() === SyntaxKind.AsteriskToken // @importType *
+                    || token() === SyntaxKind.OpenBraceToken // @importType {
+                ) {
+                    importClause = parseImportClause(identifier, afterImportTypeTagPos, /*isTypeOnly*/ true);
```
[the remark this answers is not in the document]
a-tarasyuk: [importTag7](https://github.com/microsoft/TypeScript/pull/57207/files#diff-ca1a76cda424e12411feef631c6709f7315fa7533021723f77f0071fce5d283d)

src/compiler/transformers/declarations.ts
```diff
@@ -1458,6 +1459,8 @@ export function transformDeclarations(context: TransformationContext) {
         }
         if (isDeclaration(input) && isDeclarationAndNotVisible(input)) return;
 
+        if (isJSDocImportTypeTag(input)) return;
```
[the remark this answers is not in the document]
a-tarasyuk: @sandersn https://github.com/microsoft/TypeScript/pull/57207#issuecomment-1932352470

sandersn (changes requested):
I re-reviewed the compiler part and found a couple of missing pieces. Still need to look at the services part, and double-check that tests are complete (though they look that way from my initial glance).

src/compiler/checker.ts
```diff
@@ -9726,12 +9730,13 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     case SyntaxKind.ImportClause: {
                         const generatedSpecifier = getSpecifierForModuleSymbol(target.parent || target, context); // generate specifier (even though we're reusing and existing one) for ambient module reference include side effects
                         const specifier = bundled ? factory.createStringLiteral(generatedSpecifier) : (node as ImportClause).parent.moduleSpecifier;
+                        const attributes = isImportDeclaration(node.parent) ? node.parent.attributes : undefined;
```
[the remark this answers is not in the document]
sandersn: I think I see, based on the tests: an import type gets emitted instead of an import statement + reference to aliased type.

sandersn (approved):
Looks good to me, but I want to get @weswigham 's opinion on the method of emitting import types. Would it be correct/reasonably easy to emit an import statement + references to an alias instead of retaining the current approach of emitting an import type in-place?

weswigham:
> Would it be correct/reasonably easy to emit an import statement + references to an alias instead of retaining the current approach of emitting an import type in-place?

If the statement's exposed named are bound as an `Alias`, like import declarations are, they should get used over an import type in declaration emit automatically. Actually, that this works without binding anything new right now surprises me - we must do some very special lookups for jsdoc types.

To get us to emit statements, you'd probably want to make sure we bind as `SymbolFlags.Alias`, there's a codepath in `getTargetOfAliasDeclaration` for those types, and that we actually transform (rather than uniformly elide) the jsdoc imports in declaration emit.

tests/baselines/reference/importTag15(module=es2015).js
```diff
@@ -21,4 +20,6 @@ export interface I {
 /** @import { I } from './0' with { type: "json" } */
 /** @import * as foo from './0' with { type: "json" } */
 /** @param {I} a */
-declare function f(a: import("./0").I): void;
+declare function f(a: I): void;
+import { I } from './0' with { type: "json" };
```
weswigham, line 24: ...We maybe want to transform to `import type`, since that's how we treated the tag?

a-tarasyuk: oke

DanielRosenwasser:
@typescript-bot pack this

sandersn:
@weswigham this looks ready to me now. Do you agree?

src/compiler/checker.ts
```diff
@@ -46321,7 +46333,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 return grammarErrorOnNode(node, message);
             }
 
-            if (isImportDeclaration(declaration) ? declaration.importClause?.isTypeOnly : declaration.isTypeOnly) {
+            if (isImportDeclaration(declaration) || isJSDocImportTag(declaration) ? declaration.importClause?.isTypeOnly : declaration.isTypeOnly) {
```
weswigham, line 46336: Since jsdoc imports are *always* type-only, shouldn't this error always be issued, regardless of weather the input clause had a `type` modifier or not?

a-tarasyuk: Yes, you are right. Thanks

a-tarasyuk: @weswigham could you review the latest changes? Thanks

weswigham (approved):
I think this is fine? @andrewbranch should give the type-only behavior a once-over, though.

andrewbranch:
I was skimming for a test like this and I didn’t see one, so let me know if I missed it:

```ts
// @declaration: true
// @emitDeclarationOnly: true
// @checkJs: true
// @allowJs: true

// @filename: 0.ts
export default interface Foo {}
export interface I {}

// @filename: 1.js
/** @import Foo, { I } from './0' */

/**
 * @param {Foo} a
 * @param {I} b
 */
export function foo(a, b) {}
```

Is this input syntax valid, and does it emit valid declaration file syntax? The input syntax should be valid in my opinion, but if it wasn’t handled specifically, it may emit invalid syntax on the output:

```ts
import type Foo, { I } from './0';
//     ^^^^^^^^^^^^^^^
// A type-only import can specify a default import or named bindings, but not both.(1363)
```

sandersn (approved):

andrewbranch (approved):

bcomnes:
I love this feature! I've noticed though, that if you import a type with the new syntax, and then only use that type inside of another typedef (used or unused in the js file), you get a ts6133 "is declared but its value is never read." warning, despite the imported type resolving and the new typedef being declared properly.  It only counts read state when something is assigned the type directly.

bcomnes:
Hrmmm you are right. I'm having a hard time coming up with a minimal reproducible example. I think it may be related to cycles in the type import, but also related to usage of the cycling type.

acidoxee:
My apologies if this has already been discussed, but it seems the new `@import` syntax doesn't support import attributes, especially the `resolution-mode` one that enables `@typedef` to resolve imports correctly regardless of the type of module being used.

For additional details, my codebase is mostly ESM, but some parts still have to be CJS for compatibility reasons with "old" tools. In those CJS parts, I previously imported types from the ESM part like this:

```ts
/** @typedef {import('../path/to/some/ESModule.js', { with: { 'resolution-mode': 'import' } }).MyType} MyType */
```

I have tried migrating to the new `@import` tag with an equivalent syntax, but it seems the import attribute is being ignored:

```ts
/** @import {MyType} from '../path/to/some/ESModule.js' with { 'resolution-mode': 'import' } */
```

Using the above syntax yields this error message:

> The current file is a CommonJS module whose imports will produce 'require' calls; however, the referenced file is an ECMAScript module and cannot be imported with 'require'. Consider writing a dynamic 'import("../path/to/some/ESModule.js")' call instead.
  To convert this file to an ECMAScript module, change its file extension to '.mjs', or add the field `"type": "module"` to '/CJSpart/package.json'. ts(1479)

Am I wrong, or is this simply not available yet for `@import`? Thanks 🙏

a-tarasyuk:
@acidoxee Can you create a new issue? I'll take a look at it.

acidoxee:
Sure @a-tarasyuk, here you go! https://github.com/microsoft/TypeScript/issues/58955
I hope the format of the issue is OK, I couldn't create a TS playground with several files because of https://github.com/microsoft/TypeScript/issues/52666.

zheung:
> @bcomnes I can't reproduce that behavior; here's my attempt in the playground: https://www.typescriptlang.org/play/?noUnusedLocals=true&ts=5.6.0-dev.20240605&filetype=js#code/PTAEAEDsHsFVIK4GcCmATAMtAxgQwDZIBcoALgE4IoCwAUHQJYC2ADtOaaAEQBmSXAbjp1gAKlERmbDqADeoAOrkGpFADEG+FAHkWpBtEhJQAX1A9y0Jtz5dQo4MNohQ8bFaYpI+yAHNzDAAeKMakABYooCjkluQiYGIS4KQAnizoKDxySirqmjp6BkZmCmra9sBAA
> 
> If you can show an example, could you open a new issue? Thanks!

maybe i can reproduce this in my case: 

types.d.ts
````typescript
export type TestType = {
	route: string;
	method?: string | undefined;
	handle: Function;
	upload?: boolean | undefined;
	option?: Object | undefined;
};
````
test1.js
````javascript
/** @import { TestType } from './types.d.ts' */

export default async function doSomething() {
	/** @type {TestType[]} */
	const types = [];

	return types;
}
````
test2.js
````javascript
/** @import { TestType } from './types.d.ts' */

export default {
	/** @type {TestType[]} */
	types: []
};
````

screenshot@240622-183722
screenshot@240622-183727
screenshot@240622-183736

**Comment by zheung** [comment]
For the same project as above, I found another problem: tsc doesn't seem to recognize types imported via @import correctly.

screenshot@240623-113727

**Comment by jbeaudoin11** [comment]
Hi guys, not sure if i should simply open an issue instead of writing here but is it me or it's not possible to just import all ?
```
/** @import * from "./types.ts" */
```

**Comment by DanielRosenwasser** [comment]
In general it's not best practice to comment on a closed PR or issue since people stop following as often, and creates noise in the conversation.

To answer your question, you have to create a name for the import:

```ts
/** @import * as myTypes from "./types.ts" */
//            ^^^^^^^^^^
```

JavaScript/TypeScript doesn't support an "import all the things from this module into the current scope" sort of import statement.

jbeaudoin11:
Alright, i'm really confused, i swear i used to do this all the time before ESM was a thing but maybe i'm wrong...
Maybe i assumed we could do it since we can do it with `export` 🤷‍♂️ 
Thanks !

juhort:
Does this only work with `.js` files, doesn't seem to be working with `.ts`/`.d.ts` files.

I want to link a type from another type
```ts
/** @import {BetterOmit} from "./better-omit" */

/**
 * Homomorphic version of the built-in `Pick` type.
 *
 * @see {@link BetterOmit}
 */
type BetterPick<T, K extends keyof T> = {
  [P in keyof T as K & P]: T[P];
};
```

If not this, is there some other way of achieving this?

jbeaudoin-lf:
@juhort You are trying to import from a ts file ? "./better-omit.ts" ? Just use the native typescript syntax like you're importing a module or types. This feature is only useful if you do hybrid ts in js via jsdoc.

juhort:
@jbeaudoin-lf Yeah, import normally works, but I'm not using that type anywhere apart from the JSDoc comment. Does it still make sense to import only for JSDoc, and is the recommended way of doing something like this?

jbeaudoin11:
@juhort Just use ts import in a ts file even if you're only using it in the tsdoc.

The goal of this feature is to improve the DX around jsdoc types in js files (`@typedef` isn't great). I don't see why you would use this in a ts file since you can import types natively.

bcomnes:
I have seen some oddness with node native esm, when you @import a .d.ts it actually expects a .js extension, even if there is no associated .js with that declaration file. The use case being: you want to define some types in real ts syntax but then import it into a .js file.

PR #57213, opened 2024-01-29
Switch our tsconfigs over to NodeNext

jakebailey:
This is fairly overdue, but there's no real affect here. I typically would prefer Node16, but the only way we could break is by... breaking ourselves, which will be caught via the self check.

Technically we use a bundler, but we also support a mode that runs our emit directly so that needs to be CJS.


weswigham (approved):
We could also add an explicit `"type": "commonjs"` to the `package.json` - not that it should *do* anything, but just to make it obvious to people.

But the big change here is that while it does nothing to us right now, it does *allow* us to add `.mts` esm files to the build - are we okay with that? Do we need to, like, lint against that?

jakebailey:
> We could also add an explicit `"type": "commonjs"` to the `package.json` - not that it should _do_ anything, but just to make it obvious to people.

I tried that and it broke things thanks to esbuild treating things differently :( See #56190 and #56237

It's on my TODO to figure that out.

> But the big change here is that while it does nothing to us right now, it does _allow_ us to add `.mts` esm files to the build - are we okay with that? Do we need to, like, lint against that?

If adding those files doesn't work, the tests would fail in the unbundled mode, so I don't think we need anything special. We could of course also do nothing, but it seems wrong to keep using settings we've told people to stop using,

sandersn:
@jakebailey Is this ready to merge or do you want to add the explicit commonjs to package.json?
Also I don't know dangerous this is to merge right before 5.4 RC. Any breaks would be obvious immediately, but also urgent to roll back.

jakebailey:
We definitely cannot merge type=commonjs, per the above.

This change should be safe but isn't at at all urgent so I won't try and get it in before the cut while I'm OOF

PR #57223, opened 2024-01-29
Looser import/export elision blocking in visitElidableStatement

### LINKED ISSUE #40603: Transforming ImportDeclaration or ExportDeclaration causes type specifiers to be output in js files
Type imports/exports which are regularly excluded in compiled JS output get included when `updateImportDeclaration` / `updateExportDeclaration` are called during transform.

**TypeScript Version:**  3+

**Related Issues:** 
https://github.com/microsoft/TypeScript/issues/31446 (this is the same issue, but it was closed prematurely by the OP)

The aforementioned issue details the problem well. Unfortunately, the OP closed it before it could be investigated because of finding a workaround. The workaround, however, was an internal method (`ts.updateNode`) which has been removed in TS 4.0+.

**Playground Link:**
Reproduction: https://repl.it/@nonara/ts-bug#index.ts

**Search Terms:** updateImportDeclaration, createImportDeclaration

**Code:**

`a.ts`
```ts
export type A = string;
```

`index.ts`
```ts
import { A } from './a'
export { A } from './a'
```

`transformer.ts`
```ts
function transformer(ctx: ts.TransformationContext) {
  const { factory } = ctx;
  return (s: ts.SourceFile) => ts.visitEachChild(s, visitor, ctx);

  function visitor(node: ts.Node): ts.Node {
    if (ts.isImportDeclaration(node) && ts.isStringLiteral(node.moduleSpecifier))
      return factory.updateImportDeclaration(
        node,
        node.decorators,
        node.modifiers,
        node.importClause,
        ts.createStringLiteral(node.moduleSpecifier.text)
      );

    if (ts.isExportDeclaration(node) && node.moduleSpecifier && ts.isStringLiteral(node.moduleSpecifier))
      return factory.updateExportDeclaration(
        node,
        node.decorators,
        node.modifiers,
        node.isTypeOnly,
        node.exportClause,
        ts.createStringLiteral(node.moduleSpecifier.text)
      );

    return node;
  }
}
```

**Expected Behaviour**: Should match output without transformer...

`index.js`
```js
export {};
```

**Actual Behaviour**: _(Output with transformer)_

`index.js`
```js
import { A } from "./a";
export { A } from "./a";
```

rbuckton:
This uses a slightly less restrictive elision blocking mechanism in `visitElidableStatement` such that we preserve import/export elision when a "before" transform updates an import or export declaration, so long as we are able to verify that the local bindings that we would have used to detect elision have not changed.

Fixes #40603

### DIFF
--- a/src/compiler/transformers/ts.ts
+++ b/src/compiler/transformers/ts.ts
@@ -86,7 +86,10 @@ import {
     isComputedPropertyName,
     isDecorator,
     isElementAccessExpression,
+    isEntityName,
     isEnumConst,
+    isExportAssignment,
+    isExportDeclaration,
     isExportOrDefaultModifier,
     isExportSpecifier,
     isExpression,
@@ -96,6 +99,8 @@ import {
     isHeritageClause,
     isIdentifier,
     isImportClause,
+    isImportDeclaration,
+    isImportEqualsDeclaration,
     isImportSpecifier,
     isInJSFile,
     isInstantiatedModule,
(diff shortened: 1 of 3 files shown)


src/compiler/transformers/ts.ts
```diff
@@ -425,13 +430,65 @@ export function transformTypeScript(context: TransformationContext) {
...
+     * as long as the local bindings for the declarations are unchanged.
+     */
+    function isElisionBlocked(node: ImportDeclaration | ImportEqualsDeclaration | ExportAssignment | ExportDeclaration) {
         const parsed = getParseTreeNode(node);
-        if (parsed !== node) {
-            // If the node has been transformed by a `before` transformer, perform no ellision on it
-            // As the type information we would attempt to lookup to perform ellision is potentially unavailable for the synthesized nodes
-            // We do not reuse `visitorWorker`, as the ellidable statement syntax kinds are technically unrecognized by the switch-case in `visitTypeScript`,
-            // and will trigger debug failures when debug verbosity is turned up
+        if (parsed === node || isExportAssignment(node)) {
```
jakebailey, line 443: Are export assignments never blocked? When I test `export =` of a type, it's elided: https://www.typescriptlang.org/play?module=1#code/C4TwDgpgBAYg9nKBeKBGATAZgCwG4BQ+EAHmHAE7DKwK5A

Or is there something else going on and I'm confused?

jakebailey:
I feel partially unsure because my reading of #40603 was that changes to the import path (specifier?) should stop elision; is that not the case?

rbuckton:
Per the original issue, the intent of the bug author [is to rewrite paths](https://github.com/microsoft/TypeScript/issues/40603#issuecomment-732446993). Pretty much everything we care about related to path resolution occurs during check, not emit. The main reason we block elision is that we can't be certain what the user's intent was when they changed the node. i.e., if you add a new import specifier, should we still elide or no? Since most of these elisions are based on how the _current file_ uses the import and not what the export actually is, the paths don't matter all that much.

jakebailey (approved):

andrewbranch (approved):

src/compiler/transformers/ts.ts
rbuckton: The purpose of blocking statement elision is to handle cases where a custom transform modifies the list of export bindings to catch cases like one where a user adds a new binding to an existing import, but we then just blindly elide the whole statement.

For `export =` and `export default` there's no change you can make to the binding list that would matter, since there isn't one to begin with. The only change you could make is to the expression and we don't look at the expression of a transformed ExportAssignment when determining whether to elide, we instead look at the original parse tree node that the transformed ExportAssignment replaces.

Issue #48247, opened 2022-03-14
Too narrow typing for Array.prototype.includes


kulak-at:
# Bug Report

The type for `Array.prototype.includes` seems to be too narrow making it hard to use in many use cases. Currently `Array<T>` would use `T` as a parameter for searchElement which makes it impractical when used to check against optional types:
```ts
const arr = [
    "a",
    "b",
    "c",
];

const fn = (val: string | number): val is string => {
    return arr.includes(val);
}
```

The code above should work properly but instead val is too wide typing to be passed to it.

The issue is even more prevalent when used with `as const` keyword:
```ts
const arr = [
    "a",
    "b",
    "c"
] as const;

const isInConstArray = (val: string): boolean => {
    return arr.includes(val); // Error, Argument of type 'string' is not assignable to parameter of type '"a" | "b" | "c"'.
}
```

### 🕗 Version & Regression Information

Seems to be old bug, I could reproduce it in 3.9.7 which seems to be the first to understand `.includes` syntax on the Playground.
- This is the behavior in every version I tried, and I reviewed the FAQ for entries about Array / includes

### ⏯ Playground Link

[Playground link with relevant code](https://www.typescriptlang.org/play?ts=4.6.2#code/MYewdgzgLgBAhgJwTAvDA2gKBjmAiOPAGm1zwCNjSc9g9MBdAbk01ElgEsIBJMAYXDQAgkjgBPVDAAUANzgAbAFwxoCTmADmMAD4wwAVwC25AKYIAlCvIgQC03DCoAfDADe1GAlNQDCJ4gIAHQawAoGACamEHKKFiwAvkA)

### 💻 Code

```ts
const arr = [
    "a",
    "b",
    "c"
];

const isInConstArray = (val: string | number): boolean => {
    return arr.includes(val);
}
```

### 🙁 Actual behavior

Error when passing `val` as parameter to `arr.includes`.

### 🙂 Expected behavior

`arr.includes` should accept wider type and behave like type guard instead.

MartinJohns:
Duplicate of #26255 / #14520.

whzx5byb:
Workaround:

```ts
interface ReadonlyArray<T> {
    includes<U>(this: T extends U ? this : never, searchElement: U): boolean;
}

const arr = [
    "a",
    "b",
    "c"
] as const;

declare var v1: string;
declare var v2: number;

arr.includes(v1); // OK
arr.includes(v2); // Error as expected
```

[...]

Issue #44645, opened 2021-06-18
the types 'T' and 'typeof Class' have no overlap


karikera:
# Bug Report

### 🔎 Search Terms

template class no overlap label:Bug 

### 🕗 Version & Regression Information
4.3.4 / 4.4.0-dev.20210617 / Nightly Playground

### ⏯ Playground Link

[Playground Link](https://www.typescriptlang.org/play?#code/FAYwNghgzlAEDCkawN7Fh2ATArgWzwE9YBeWABgG5gBfYegMxwDsQAXASwHtnY2BTKGwA8AFQB8ACggAuUQEoZANy4csaTLA4NY00iTKJoUeajp1gQA)

### 💻 Code

```ts

class Class {
    dummy = 0;
}


function test<T>(a:T):void{
    if (a === Class) {} // This condition will always return 'false' since the types 'T' and 'typeof Class' have no overlap.ts(2367)
}

```

### 🙁 Actual behavior
ts2367 error. `a` can be `Class`

### 🙂 Expected behavior
no error. `a` can be `Class`

RyanCavanaugh:
Not all comparisons are allowed; you can upcast either side to `unknown` if this was intentional.

karikera:
Thanks for replying.
I'm reducing it from my use cases. but it seems I reduced it too much. without knowing it's intentional work

```ts

class Class {
}

class Class2 extends Class {
    dummy = 0;
}

class Class3 {
}

function test<T extends typeof Class>(a:T):void{
    if (a === Class) {} // it's okay
    if (a === Class3) {} // it's okay
    if (a === Class2) {} // it has problem
}

const test2:typeof Class = Class2; // it's okay


```
[Playground Link](https://www.typescriptlang.org/play?#code/FAYwNghgzlAEDCkawN7AL7FEujpQCZYBTADwBdiA7AE1x1WFmdhoFcBbDgT1gF5YABgDcGLOHwIcAZkaZgAMzZUQ5AJYB7KrEpRyAHgAqJCtTo7uAB2IaFU-AD4AFBABchgJSuAbhrU00Flg1Oxd+PgE8GA9UdFgAenjg8gByOA0AawhuJhYQ2DCIyJkYlDjE5LTYTOzc5nzCiPsYAlLypLVyWAALaFhLACcNACMwYg4xUC09HWI9AldyKxs7KLhi-AJhBI7U9Kyc4CA)

on this case. it has the same problem.
[...]

PR #57230, opened 2024-01-30
add types for set methods proposal

### LINKED ISSUE #57228: Add types for stage 3 "set methods" proposal
### ⚙ Compilation target

esnext

### ⚙ Library

esnext

Set.prototype.intersection(other)
Set.prototype.union(other)
Set.prototype.difference(other)
Set.prototype.symmetricDifference(other)
Set.prototype.isSubsetOf(other)
Set.prototype.isSupersetOf(other)
Set.prototype.isDisjointFrom(other)

### Sample Code

```TypeScript
setInstance.intersection(other)
setInstance.union(other)
setInstance.difference(other)
setInstance.symmetricDifference(other)
setInstance.isSubsetOf(other)
setInstance.isSupersetOf(other)
setInstance.isDisjointFrom(other)
```

### Documentation Link

https://github.com/tc39/proposal-set-methods

Currently stage 3, but it's shipping in Safari and will hit stable Chrome in a couple weeks, so I'll be asking for stage 4 at the April meeting.

I'll go ahead and submit a PR.

bakkot:
Fixes #57228.

Set methods are stage 3, and I expect to ask for stage 4 in April.

I've used the simplest possible types here, but in some sense these are too restrictive. It's perfectly reasonable to ask if a Set of `0 | 1` is a subset of a Set of `number`, for example, or even of a Set of `1 | 2` - anything where the intersection of the types is nonempty (I guess that would be, any type `S` such that `T & S extends never` is false).

Unfortunately I don't know of a good way to express that constraint, at least not without making the types way more complex. So I've stuck with the simple thing here. But if someone has a suggestion for better types I'd happily take it.

One possibility would be to have the Set-producing types take `SetLike<T>` but the predicates take `SetLike<unknown>`. I don't know if that would be better.

Do note that this this PR introduces the `SetLike` interface.


ssalbdivad:
> I guess that would be, any type S such that T & S extends never is false

I think this concept of "overlaps with" or its opposite "is disjoint with" is actually useful quite often. For example, it would be the ideal type of the parameter in `[].includes(...)`. Currently I use this utility:

```ts
export const includes = <array extends readonly unknown[], element>(
	array: array,
	element: element & array[number] extends never ? array[number] : element
): element is element & array[number] => array.includes(element)

declare const array: string[]
declare const overlapping: string | number
declare const nonOverlapping: boolean | number

// okay
includes(array, overlapping)

// error
includes(array, nonOverlapping)
```

I wonder if the team would consider adding a builtin `Overlaps` type for cases like these and the set methods in this PR.

fatcerberus:
> Unfortunately I don't know of a good way to express that constraint, at least not without making the types way more complex. So I've stuck with the simple thing here. But if someone has a suggestion for better types I'd happily take it.

I'm not sure there's a good solution for this right now - it's more or less the same problem as #48247/#26255 and the fix is generally considered to be #14520 (lower-bounded type parameters)--though FWIW I think an "overlaps with" constraint would be more useful for this than "supertype of"

ssalbdivad:
@fatcerberus I definitely agree `overlaps` would be very useful in a number of circumstances like includes, as well as internally for a lot of these comparison errors that incorrectly assert the lack of an overlap (when really they're just checking for a downcast): 

```ts
declare const a: { a: true }

declare const b: { b: true }

// '{ a: true; }' and '{ b: true; }' have no overlap
if (a === b) {
}
```

fatcerberus:
Well, in reality any `overlaps` primitive we got would likely use the same internal check that produces the "these types don't overlap" error, so I don't think it would fix that problem by itself :wink:

ssalbdivad:
@fatcerberus Maybe it would be an opportunity to revisit that logic if it were to be published as its own utility 😅

src/lib/esnext.collection.d.ts
```diff
@@ -9,3 +9,49 @@ interface MapConstructor {
         keySelector: (item: T, index: number) => K,
     ): Map<K, T[]>;
 }
+
+interface SetLike<T> {
```
DanielRosenwasser, line 13: At first glance, this looks a lot like `ReadonlySet`. I know that these methods only need a subset of functionality, but it might be worth using that instead.

DanielRosenwasser:
@typescript-bot pack this
@typescript-bot test this
@typescript-bot test top200
@typescript-bot run dt

src/lib/esnext.collection.d.ts
bakkot: You need the `keys` method, so I don't think that works unless that gets added ReadonlySet as well. (And users shouldn't have to implement `forEach` for their own set-likes, though that isn't as big a deal since that probably doesn't come up as much.)

DanielRosenwasser:
The "overlaps" concept is what we internally call "comparability". The problem with expressing it as an intersection that produces `never` is that it is rarely possible to say an intersection of two object types is vacuous.

In other words, you can't say that `{ meow(): void } & { woof(): void }` is an impossible type to fulfill, so there is no straightforward way to disallow `catSet.intersection(dogSet)`.

src/lib/esnext.collection.d.ts
bakkot: Wait, no, `keys` is already there, just in a different file, whoops.

Still, I think it would be kind of annoying not to be able to do `set.union({ keys(){ return array.keys() }, size: array.size, has: () => {} })` or similar. Not the end of the world, but I'd prefer to keep `SetLike` unless there's a reason to switch to `ReadonlySet`.

ssalbdivad:
@DanielRosenwasser Yes, that is exactly the behavior I'd expect in a structural type system, with a result of `Set<{meow(): void; woof(): void}>`. The type could still give completions based on the base type which must be overlapped, but this kind of intersection is definitely sensible as illustrated by an alternate example like `{ poisonous: true } & { mammalian: true }` for a set of platypuses and echidnas 😄 

I'm often frustrated by TS overreaching based on a comparability check (e.g. https://github.com/microsoft/TypeScript/issues/44645), especially when the word "overlap" is used in the error message which has a specific meaning that does not match the criteria for the error.

DanielRosenwasser:
I brought this up in the design meeting, specifically focusing on

1. the types of `union`/`intersection`
2. the parameter container types.

The feedback for them was:

1. Methods like `union` and `intersection` should be generic and return a corresponding TypeScript type. `union` should return a `Set<T | U>`, and `intersection` should return a `Set<T & U>`.
1. Methods should not expect their arguments to be related to the underlying container type. This risks accidental invocations of `intersection` that are guaranteed to produce empty collections, but it has the benefit of allowing more symmetric operations.

   In other words, both `stringSet.intersection(stringOrBooleanSet)` and `stringOrBooleanSet.intersection(stringSet)` are valid invocations of each other.

   I'll note that I am personally the most opposed to this behavior, but I was the lone dissenter. 🙂
1. With respect to `SetLike`, @rbuckton raised the same points of `ReadonlySet` being overkill. I suggested `ReadonlySetLike` might make more sense. Nobody else had a strong opinion.

@bakkot, in this PR I'd suggest to just address the first two points, do whatever you feel is right for point 3, and we will bikeshed by the time we can bring this in for TypeScript 5.5.

bakkot:
WFM, thanks for the feedback. I'll get that pushed up later today.

Any opinions on `isSubsetOf` and friends? If not I'll default to making them generic, following point 2, even though (because they return booleans) this doesn't let you constrain the output.

src/lib/esnext.collection.d.ts
```diff
@@ -9,3 +9,49 @@ interface MapConstructor {
...
+     */
+    symmetricDifference(other: SetLike<T>): Set<T>;
+    /**
+     * @returns a boolean indicating whether all the elements in this Set are also in the argument.
+     */
+    isSubsetOf(other: SetLike<T>): Set<T>;
+    /**
+     * @returns a boolean indicating whether all the elements in the argument are also in this Set.
+     */
+    isSupersetOf(other: SetLike<T>): Set<T>;
```
DanielRosenwasser, line 52: These should return `boolean`s like you mentioned 😄

DanielRosenwasser:
> Any opinions on `isSubsetOf` and friends? If not I'll default to making them generic

I think the right thing for those is to take a `Set<unknown>` since the type parameter will never be observed.

src/lib/esnext.collection.d.ts
bakkot: lol I don't know how I managed that.

fatcerberus:
> In other words, both `stringSet.intersection(stringOrBooleanSet)` and `stringOrBooleanSet.intersection(stringSet)` are valid invocations of each other.

This is generally what people expect of e.g. `Array.prototype.includes` as well, and I largely concur, but to my knowledge there's no way to allow that without *also* allowing `stringSet.intersection(numberSet)` (unless we get lower-bounded or comparability constraints), correct?

bakkot:
@DanielRosenwasser Done. Specifically:

```js
union<U>(other: ReadonlySetLike<U>): Set<T | U>; // things in result are in either
intersection<U>(other: ReadonlySetLike<U>): Set<T & U>; // things in result are in both
difference<U>(other: ReadonlySetLike<U>): Set<T>; // things in result are in receiver
symmetricDifference<U>(other: ReadonlySetLike<U>): Set<T | U>; // things in result are in either
```
and then `ReadonlySetLike<unknown>` as the parameter for the three predicates.

ssalbdivad:
@fatcerberus It's possible to validate as a function parameter with an approach like the one I mentioned [in my original comment](https://github.com/microsoft/TypeScript/pull/57230#issuecomment-1916890725) which could be leveraged here, but I assume that falls into the category of "too complex for builtin types."

fatcerberus:
Hmm, I have mixed feelings about the name `ReadonlySetLike`.  While I get it--it doesn't have any of the mutating methods--it'd be weird (and maybe a bit misleading, for people reading the code) if I actually wanted to make a *mutable* setlike and had to extend `ReadonlySetLike` to get there.

Also it doesn't seem completely analogous to `ArrayLike` since the methods are defined to return a full-fledged `Set`...

RyanCavanaugh:
> but to my knowledge there's no way to allow that without also allowing...

Correct. "This is why we need `Comparable<T>`!" got brought up multiple times in the discussion 🙂

DanielRosenwasser:
If we are seriously thinking about `Comparable<T>`, why not take the more conservative option first and then introduce `Comparable`?

RyanCavanaugh:
The conservative approach still fails basic desirable things like subtyping the output or `a.intersect(b) === b.intersect(a)`

rbuckton:
> Hmm, I have mixed feelings about the name `ReadonlySetLike`. While I get it--it doesn't have any of the mutating methods--it'd be weird (and maybe a bit misleading, for people reading the code) if I actually wanted to make a _mutable_ setlike and had to extend `ReadonlySetLike` to get there.

While I'm fine with both `SetLike` and `ReadonlySetLike`, there are other names we could consider as well. In C++, the requirements for a `Container` are quite similar (i.e., has a size, is iterable), so `Container<T>` is a possible option if there's a strong negative sentiment towards `ReadonlySetLike`.

> Also it doesn't seem completely analogous to `ArrayLike` since the methods are defined to return a full-fledged `Set`...

These methods allocate and return a new `Set` no matter what object you pass as the argument so long as it has the minimum necessary API for the methods to succeed. In fact, `ArrayLike` and `SetLike` are nearly identical in their capabilities, but are tailored to their specific protocols. Both are minimal, read-only representations of their respective full implementations. `ArrayLike` has a read-only `length`, the ability to read a single element by index, and the ability to read all elements by index and length. While `SetLike` has a read-only `size`, the ability to test whether a single value is in its domain via `has()`, and the ability to read all elements via `keys()`.

fatcerberus:
Ah, I think I misinterpreted `ReadonlySetLike` as being the interface that provided the methods, which I'm guessing now isn't actually the case - that's the part I was objecting to, since they always return a real `Set`.  My objections may have been misplaced :)

ssalbdivad:
@RyanCavanaugh Huge +1 for `Comparable` as a builtin type.

While I still prefer `Overlaps` because it's easier to reason about (is it possible to create a value that satisfies both types) without knowing the heuristics in TSC, `Comparable` would still be a huge practical improvement.

For now, I'm stuck using this utility which, while neat, feels like overkill 😅 

```ts
export type isDisjoint<l, r> = l & r extends never
	? true
	: // domainOf is similar to typeof in JS, but aligned to TS keywords
	domainOf<l> & domainOf<r> extends never
	? true
	: [l, r] extends [object, object]
	? true extends valueOf<{
			[k in Extract<
				keyof l & keyof r,
				requiredKeyOf<l> | requiredKeyOf<r>
			>]: isDisjoint<l[k], r[k]>
	  }>
		? true
		: false
	: false
```

inoyakaigor:
@bakkot I'm not from TS team and just curious: wouldn't it be better if `ReadonlySetLike` interface will be like this:
```ts
type ReadonlySetLike<T> = Pick<Set<T>, 'size' | 'has' | 'keys'>
```
In any case of probable future changes of `Set` typings for `size`, `has` or `keys` ReadonlySetLike will also have it automatically

bakkot:
Sure, that would work too. I leave it up to the TS to decide which approach they'd prefer.

noinkling:
Doesn't `keys()` need to return an `Iterator` anyway, not an `Iterable`?

https://github.com/microsoft/TypeScript/blob/6ceb39551a30a3aac50eb862c3f959cfc892d478/src/lib/esnext.collection.d.ts#L17

In current Chrome:
```js
const mySetLike = {
  size: 2,
  has(a) { return a === 'foo' || a === 'bar' },
  keys() { return ['foo','bar'] }
};

// new Set(['baz']).union(mySetLike);
// Uncaught TypeError: string "next" is not a function

mySetLike.keys = () => {
  let i = 0;
  return {
    next() {
      i++;
      if (i === 1) return { value: 'foo', done: false };
      if (i === 2) return { value: 'bar', done: false };
      return { done: true };
    }
  };
};

new Set(['baz']).union(mySetLike);
// Set(3) {'baz', 'foo', 'bar'}
```
Also piggybacking off `Set` would require it to return an `IterableIterator`, which isn't necessary as you can see.

bakkot:
@noinkling Nice catch, that's an embarrassing typo given that I designed the feature 😅. Fixed.

Possibly should update the description of `keys`/`values`/`entries` on `Map` and `Set` to say "iterable iterator" instead of "iterable", but not in this PR.

inoyakaigor:
@bakkot One more question about return type for `difference` and `symmetricDifference`.
I suspect a return type for this two functions should be something like this, isn't it?
```js
difference<U>(otherSet: ReadonlySetLike<U>): Set<Exclude<T, U>> 

symmetricDifference<U>(otherSet: ReadonlySetLike<U>): Set<Exclude<T, U> | Exclude<U, T>>
```

bakkot:
Depends on how we want to make the tradeoff between precision and making the types noisier, but yes, formally that's correct.

inoyakaigor:
Any blockers* for this PR? My PR into Mobx mobxjs/mobx#3853 state manager (which mostly used for React.js) is depending on it
____
_* except failed tests_

rbuckton (changes requested):
It looks like baselines need to be updated after the most recent commit.

rbuckton:
> @bakkot One more question about return type for `difference` and `symmetricDifference`. I suspect a return type for this two functions should be something like this, isn't it?
> 
> ```js
> difference<U>(otherSet: ReadonlySetLike<U>): Set<Exclude<T, U>> 
> 
> symmetricDifference<U>(otherSet: ReadonlySetLike<U>): Set<Exclude<T, U> & Exclude<U, T>>
> ```

Those definitions would be too strict since a given instantiation of `Set` does not imply that the set itself contains all of those values, only that it *may* contain those values:

```ts
const set1 = new Set<"A" | "B">(["A", "B"]);
const set2 = new Set<"B" | "C">(["C"]);
const set3 = set1.difference(set2);
set.has("A"); // true
set.has("B"); // true
```
Even though `set2` is declared as `Set<"B" | "C">`, it does not actually contain a `"B"` value, thus `"B"` is not actually excluded from `set3`.

rbuckton:
> @bakkot I'm not from TS team and just curious: wouldn't it be better if `ReadonlySetLike` interface will be like this:
> 
> ```ts
> type ReadonlySetLike<T> = Pick<Set<T>, 'size' | 'has' | 'keys'>
> ```
> 
> In any case of probable future changes of `Set` typings for `size`, `has` or `keys` ReadonlySetLike will also have it automatically

IMO, an interface is a better option as it allows for interface merging if a polyfill is loaded that also introduces these types.

inoyakaigor:
> Those definitions would be too strict

@rbuckton so maybe wrap it in `Partial<>` ?

src/lib/esnext.collection.d.ts
```diff
@@ -9,3 +9,49 @@ interface MapConstructor {
...
+     */
+    symmetricDifference<U>(other: ReadonlySetLike<U>): Set<T | U>;
+    /**
+     * @returns a boolean indicating whether all the elements in this Set are also in the argument.
+     */
+    isSubsetOf(other: ReadonlySetLike<unknown>): boolean;
+    /**
+     * @returns a boolean indicating whether all the elements in the argument are also in this Set.
+     */
+    isSupersetOf(other: ReadonlySetLike<unknown>): boolean;
```
rbuckton, line 52: @DanielRosenwasser would overloads like the following be worthwhile?

```ts
interface Set<T> {
  isSubsetOf<U>(other: ReadonlySetLike<U>): this is Set<T & U>;
  isSupersetOf<U>(other: ReadonlySetLike<U>): other is ReadonlySetLike<T & U>;
}
```

src/lib/esnext.collection.d.ts
```diff
@@ -9,3 +9,49 @@ interface MapConstructor {
...
+     * @returns a boolean indicating whether an element with the specified value exists in the set-like or not.
+     */
+    has(value: T): boolean;
+    /**
+     * @returns the number of (unique) elements in the set-like.
+     */
+    readonly size: number;
+}
+
+interface Set<T> {
```
rbuckton, line 28: Since all of these methods are non-mutating, we might also want to add them to `ReadonlySet<T>` since `ReadonlySet<T>` serves the same purpose as what `ReadonlyArray<T>` does for `Array<T>`.

rbuckton (changes requested):

bakkot:
> It looks like baselines need to be updated after the most recent commit.

Hm. The way I've been doing that (`npm ci; npm run build; npm run test -- --no-lint; npx hereby baseline-accept`) doesn't produce any changes locally for this commit. And the diff in CI doesn't immediately look like it should be caused by the last commit, even though the one before was passing. Is there a different process I should follow here?

jakebailey:
CI always tests main + the PR; have you merged main locally?

bakkot:
Aha, thanks, that's fixed it.

src/lib/esnext.collection.d.ts
bakkot: Done.

src/lib/esnext.collection.d.ts
rbuckton: We can always change these in a follow-up PR.

rbuckton (approved):

PR #57235, opened 2024-01-30
fix(57215): Organize Imports: Incorrect combination of imports with import attributes 

### LINKED ISSUE #57215: Organize Imports: Incorrect combination of imports with import attributes
### 🔎 Search Terms

"organize imports assert", "organize imports with", "organize imports assertions", "organize imports attributes"

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about "organize imports"

### ⏯ Playground Link

Playground doesn't support organize imports as far as I can tell.

### 💻 Code

In VSCode run the organize imports command on the following code:

```ts
import { A } from "./file";
import { type B } from "./file";
import { C } from "./file" assert { type: "a" };
import { A as D } from "./file" assert { type: "b" };
import { E } from "./file" with { type: "a" };
import { A as F } from "./file" with { type: "b" };

type G = A | B | C | D | E | F;
```

### 🙁 Actual behavior

Currently, organize imports will combine all the imports to the following line:

```ts
import { A, C, A as D, E, A as F, type B } from "./file";

type G = A | B | C | D | E | F;
```

### 🙂 Expected behavior

As import attributes can change the behavior of the imported module they should stay as separate lines:
```ts
import { A, type B } from "./file";
import { C } from "./file" assert { type: "a" };
import { A as D } from "./file" assert { type: "b" };
import { E } from "./file" with { type: "a" };
import { A as F } from "./file" with { type: "b" };

type G = A | B | C | D | E | F;
```

_No response_

a-tarasyuk:
Fixes #57215

### DIFF
--- a/src/services/organizeImports.ts
+++ b/src/services/organizeImports.ts
@@ -90,7 +90,7 @@ export function organizeImports(
     const shouldCombine = shouldSort; // These are currently inseparable, but I draw a distinction for clarity and in case we add modes in the future.
     const shouldRemove = mode === OrganizeImportsMode.RemoveUnused || mode === OrganizeImportsMode.All;
     // All of the old ImportDeclarations in the file, in syntactic order.
-    const topLevelImportGroupDecls = groupByNewlineContiguous(sourceFile, sourceFile.statements.filter(isImportDeclaration));
+    const topLevelImportGroupDecls = groupTopLevelDeclarations(sourceFile, sourceFile.statements.filter(isImportDeclaration));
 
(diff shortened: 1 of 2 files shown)


DanielRosenwasser (approved):

iisaduan (changes requested):
Thanks for the eyes! However, I think the best thing to do is to close this PR as I am currently reworking organizeImports, and since the fix is incorrect, I'll submit a temporary fix instead. Currently this fix gives some weird behavior to organizeImports such as
```
import { A, } from "./a";
import { C } from "./a" assert { type: "a" };
import { b } from "./a"
```
becoming 
```
import { A, } from "./a";
import { C, b } from "./a" assert { type: "a" };
```
and 
```
import { A, } from "./a";
import { C } from "./c" assert { type: "c" };
import { b } from "./b"
```
will no longer sort.

Issue #57231, opened 2024-01-30
discriminated union type matching behaviour changed starting from 5.2.x


tusharf5:
### 🔎 Search Terms

"key order when matching object types" "discriminated unions key order"

### 🕗 Version & Regression Information

- This changed between versions 5.1 and 5.2

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.2.2#code/FAFwngDgpgBAYgewQExgXhgcgIYQgGykxgB8sEAnbAOwHMiBuUSWANSnpGwCND0sAzhACW1bAGMAFsTKZx2ChQQhMTcNBgB5EJKgV+mALbD8AaxlYA7thB7VzDQGEArgJAJDB7MIoXMAhBN7B1gYABUFTnRgGFi4sgBvGLiU2OxnZGEoanEoAC4sKAUde1TUwyL8POSylLIXNw8a2tiyAAMAMwpnYRAAfQASBMQUAF825pb2gDcOKC5eKEGE9k4eQnHJ2vblXQpl7T3xphbRrcStlPTM7NyCowUBPwArZxFbXxOWmArsKpg3BRRLQvmUzsBgOIENQ3DBZmtFgVVvN1rAMHIFEoVExIdDYVwKJwChFCfN+DAklcMlkcvlCsVpAAaZq-f5teEoxbLDkLDYTUY4qEwkAwAmcABMxMiZIwFJZlQK7LmvKWQx5qPGzKpN1p9yKFBKwAFwCAA

### 💻 Code

```ts

type Food = 'apple' | 'orange';
type Vegetable = 'spinach' | 'carrot';
type Other = 'milk' | 'water';
type Custom = 'air' | 'soil';

type  Target =
      | {
          audience: 'earth';
          meal:
            | Custom
            | `fruit_${Food}`
            | `vegetable_${Vegetable}`
            | `other_${Other}`;
        }
      | {
          audience: 'mars' | 'jupiter';
          meal: string;
        }


const vegetable: Vegetable = 'carrot';

const target: Target =  {
    audience: 'earth',
    meal: `vegetable_${vegetable}`
};

const target2: Target =  {
    meal: `vegetable_${vegetable}`,
    audience: 'earth'
};

```

<details><summary><b>Output</b></summary>

```ts
"use strict";
const vegetable = 'carrot';

const target = { // ok
    audience: 'earth',
    meal: `vegetable_${vegetable}`
};

const target2 = { // error
[...]

Issue #57199, opened 2024-01-27
Cannot use const string in enum when it's type is specified explicitly


undsoft:
### 🔎 Search Terms

enum string, enum const

### 🕗 Version & Regression Information

- This is the behavior in every 5+ version available in Playground.

### ⏯ Playground Link

https://www.typescriptlang.org/play?#code/MYewdgzgLgBAZgQwJYBsIC4YHIAqBTADyixgF5t8isBuAKFElgHcQAnAawjIsOLtrxgArgFsYARSF5oScDgCeABzwwA3rRgxKscolQQANBq28ATNxYdDtAL5A

### 💻 Code

```ts
const fails: 'Text' = 'Text';
const works = 'Text';

enum QuestionType {
  Text = fails,
  Text2 = works,
}
```

### 🙁 Actual behavior

When `fails` is used in enum, there is an error:

> Type 'string' is not assignable to type 'number' as required for computed enum member values.(18033)

### 🙂 Expected behavior

I can use `fails`, because to me it is equivalent to `works`.

### Additional information about the issue

Related to https://github.com/microsoft/TypeScript/issues/29718 ?

ahejlsberg:
Enum members may be initialized by _constant expressions_ that compute string or numeric values. Constant expressions may reference `const` variables, but only if those variables have no type annotations. When enum member initializers reference variables with type annotations, they're considered non-constants to be computed at run-time. Such computed enum members are restricted to numeric values. See #50528 for the exact rules.

RyanCavanaugh:
> Constant expressions may reference const variables, but only if those variables have no type annotations

I think this is probably the thing that sounds surprising, but presumably it's to ward off problems like this?
```ts
// Correct emit is ... ?
const fails: number | 'Text' = 'Text';
```

ahejlsberg:
> but presumably it's to ward off problems like this?

That's one reason, but another is to avoid circularities. Imagine that a type annotation depends on the enum type for which we are in the process of evaluating constant expressions.

Andarist:
Isn't @RyanCavanaugh argument already somewhat trumped by allowing this in other similar situations like this one?
```ts
// input
const one: number | "Text" = "Text";
const two = one;

// dts output
declare const one: number | "Text";
declare const two: "Text";
```

I also found a confusing error to be reported in a related situation:
```ts
const a = E.two; // A member initializer in a enum declaration cannot reference members declared after it, including members defined in other enums.(2651)
[...]

PR #57236, opened 2024-01-30
Evaluate expressions with annotated constant variables

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/57231

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -45495,7 +45495,16 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                         if (isConstantVariable(symbol)) {
                             const declaration = symbol.valueDeclaration;
                             if (declaration && isVariableDeclaration(declaration) && declaration.initializer && (!location || declaration !== location && isBlockScopedNameDeclaredBeforeUse(declaration, location))) {
-                                return evaluate(declaration.initializer, declaration);
+                                if (!declaration.type) {
+                                    return evaluate(declaration.initializer, declaration);
+                                }
+                                const declaredType = getTypeOfSymbol(symbol);
+                                const type = declaredType.flags & TypeFlags.Union ?
+                                    getAssignmentReducedType(declaredType as UnionType, getInitialType(declaration)) :
+                                    declaredType;
+                                if (type.flags & TypeFlags.StringOrNumberLiteral) {
+                                    return (type as StringLiteralType | NumberLiteralType).value;
+                                }
                             }
                         }
                     }
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -45494,7 +45494,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                         }
                         if (isConstantVariable(symbol)) {
                             const declaration = symbol.valueDeclaration;
-                            if (declaration && isVariableDeclaration(declaration) && !declaration.type && declaration.initializer && (!location || declaration !== location && isBlockScopedNameDeclaredBeforeUse(declaration, location))) {
+                            if (declaration && isVariableDeclaration(declaration) && declaration.initializer && (!location || declaration !== location && isBlockScopedNameDeclaredBeforeUse(declaration, location))) {
                                 return evaluate(declaration.initializer, declaration);
                             }
                         }
--- a/tests/baselines/reference/discriminantUsingEvaluatableTemplateExpression.symbols
+++ b/tests/baselines/reference/discriminantUsingEvaluatableTemplateExpression.symbols
@@ -37,3 +37,35 @@ foo({
   },
 });
 
+type E1 = { d: "main1-sub1"; cb: (x: string) => void };
+>E1 : Symbol(E1, Decl(discriminantUsingEvaluatableTemplateExpression.ts, 12, 3))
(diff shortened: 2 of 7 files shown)


DanielRosenwasser:
Any clue why we disallowed annotated variables previously?

ahejlsberg:
> Any clue why we disallowed annotated variables previously?

See #57199.

ahejlsberg:
This PR doesn't really fix the issue. See [here](https://github.com/microsoft/TypeScript/issues/57231#issuecomment-1918094611).

Andarist:
Hm, ye - I assumed that the `isDiscriminantProperty` was intentionally not checking this kind of thing for performance reasons. 

So this PR might not fix the referenced issue in full. Would it still make sense to have an adjusted version of it in? The mentioned enum case requires some extra test cases that would involve potential circularities but the template expression case isn't prone to that, right? The expression like this can't depend on itself (I'm sure that would already be reported as an error elsewhere 😅 )

ahejlsberg:
> Would it still make sense to have an adjusted version of it in?

Consider declarations like the following:

```ts
const someValue: number = 42;
```

The intuition here is that `someValue` represents _some_ number that I can change at will without affecting type checking because I have explicitly indicated I want it treated like it could be any number. That wouldn't be the case if constant evaluation "cheated" and looked at the value anyway.

Then what if the type annotation was `"42"`, i.e. exactly a singleton literal type. Sure, we could consider allowing that, but not clear it adds a lot of value.

Andarist:
> The intuition here is that someValue represents some number that I can change at will without affecting type checking because I have explicitly indicated I want it treated like it could be any number. That wouldn't be the case if constant evaluation "cheated" and looked at the value anyway.

That would be incorrect and it was a total oversight on my part. I somehow assumed that the type was already obtained in that conditional block and passing tests didn't make me reinspect it 😉 

I pushed out the fix for this particular issue.

> Then what if the type annotation was "42", i.e. exactly a singleton literal type? Sure, we could consider allowing that, but not clear it adds a lot of value.

I think that the main motivation behind this wouldn't be an annotation with a literal type but rather with a union type. The user might want to make sure that the initializer's type is assignable to some other type - that doesn't seem like an edge-case scenario. I think this should not affect the constant evaluation since it doesn't change the initial~ type of a variable. 

Even when we consider the literal type annotation - it doesn't bring any value and can be seen as redundant I find it unexpected that it has an effect here. Its effect could be similar to a redundant parenthesized expression - it could just be "skippable". As is, it feels like it breaks the least surprise principle. 

The union case might go slightly against your argument that an explicit indication should make it OK to change the initializer at will. However, I think that the existence of assignment-reduced types already trumps that. Initializers might already "leak" past the declared type.

Taking a closer look at constant expressions and declared types or type operators vs unions:
```ts
const v1 = "foo";
const test1 = `pre${v1}`; // "prefoo"

const v2 = "foo" as const;
const test2 = `pre${v2}`; // string

type Union = "foo" | "bar";

const v3 = "foo" satisfies Union;
const test3 = `pre${v3}`; // string

const v4: Union = "foo";
const test4 = `pre${v4}`; // string
```

We can see that the declared type situation - the one that I touched in this PR - is not that unique. However, I'd argue that all of them should work the same way and that they should return `"prefoo"` (can you get any more constant than `v2`? :P). I could make the required changes for this as part of this PR.

The examples here use template literal expressions but the same applies to enum members.

src/compiler/checker.ts
```diff
@@ -45494,8 +45494,17 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+                            if (declaration && isVariableDeclaration(declaration) && declaration.initializer && (!location || declaration !== location && isBlockScopedNameDeclaredBeforeUse(declaration, location))) {
+                                if (!declaration.type) {
+                                    return evaluate(declaration.initializer, declaration);
+                                }
+                                const declaredType = getTypeOfSymbol(symbol);
+                                const type = declaredType.flags & TypeFlags.Union ?
+                                    getAssignmentReducedType(declaredType as UnionType, getInitialType(declaration)) :
+                                    declaredType;
+                                if (type.flags & TypeFlags.StringOrNumberLiteral) {
+                                    return (type as StringLiteralType | NumberLiteralType).value;
```
Andarist, line 45506: Note that the initializer is still required for all of this to work here. I'm not sure what's the preferred way of handling potential errors in situations like this:

```ts
const str: 'foo' = 'bar'
const another = `_${str}` // "_foo" or "_bar" ?
```

I think the argument can be made both ways. I'm not sure if this matters all that much when the program errors anyway - but perhaps using the evaluated initializer in this situation would be slightly more consistent. After all, the initializer is still required for this to work - so the evaluation happens on the grounds that the expression that can be evaluated is there. It's not like this can be used as part of the evaluation: `declare const str: 'foo'`;

It's also worth noting that this code here should be expanded with support for enum members.

sandersn:
What is the status of this PR? From the comments here and on the bug it fixes, it looks to me that this doesn't fix the problem adequately. But I don't see a good fix for it elsewhere.

Is the change in 5.2 unambiguously bad? How bad? Maybe this isn't worth fixing? The broken code mentioned in the bug and here sounded like it was low-frequency and not that useful.

Andarist:
1.5 year later, I don't think the 5.2 change was a true root cause of the whole issue. The real problem predates that change. I just opened what I believe is a better solution for this: https://github.com/microsoft/TypeScript/pull/62203

PR #57250, opened 2024-01-31
fix 57215 -- add support for import attributes to OrganizeImports

### LINKED ISSUE #57215: Organize Imports: Incorrect combination of imports with import attributes

iisaduan:
This fix will prevent import statements from coalescing unless they also have the same import attributes. It does not reorder statements with the same module specifier but different attributes.

Fixes #57215

### DIFF
--- a/src/services/organizeImports.ts
+++ b/src/services/organizeImports.ts
@@ -59,6 +59,7 @@ import {
     Scanner,
     setEmitFlags,
     some,
+    sort,
     SortKind,
     SourceFile,
     stableSort,
@@ -324,7 +325,11 @@ function coalesceImportsWorker(importGroup: readonly ImportDeclaration[], compar
 
     const importGroupsByAttributes = groupBy(importGroup, decl => {
         if (decl.attributes) {
-            return decl.attributes.getFullText();
+            let attrs = decl.attributes.token + " ";
+            for (const x of sort(decl.attributes.elements)) {
+               attrs += x.name + ":" + x.value.getText() + " ";
+            }
+            return attrs;
         }
         return "";
     });
--- a/src/services/organizeImports.ts
+++ b/src/services/organizeImports.ts
@@ -327,7 +327,7 @@ function coalesceImportsWorker(importGroup: readonly ImportDeclaration[], compar
         if (decl.attributes) {
             let attrs = decl.attributes.token + " ";
             for (const x of sort(decl.attributes.elements)) {
-               attrs += x.name + ":" + x.value.getText() + " ";
+                attrs += x.name + ":" + x.value.getText() + " ";
             }
             return attrs;
         }
--- a/src/services/organizeImports.ts
+++ b/src/services/organizeImports.ts
@@ -322,96 +322,105 @@ function coalesceImportsWorker(importGroup: readonly ImportDeclaration[], compar
         return importGroup;
     }
 
-    const { importWithoutClause, typeOnlyImports, regularImports } = getCategorizedImports(importGroup);
+    const importGroupsByAttributes = groupBy(importGroup, decl => {
+        if (decl.attributes) {
+            return decl.attributes.getFullText();
+        }
+        return "";
+    });
     const coalescedImports: ImportDeclaration[] = [];
+    for (const attribute in importGroupsByAttributes){
+        const importGroupSameAttrs = importGroupsByAttributes[attribute] as ImportDeclaration[];
+        const { importWithoutClause, typeOnlyImports, regularImports } = getCategorizedImports(importGroupSameAttrs);
 
-    if (importWithoutClause) {
-        coalescedImports.push(importWithoutClause);
-    }
-
-    for (const group of [regularImports, typeOnlyImports]) {
-        const isTypeOnly = group === typeOnlyImports;
-        const { defaultImports, namespaceImports, namedImports } = group;
-        // Normally, we don't combine default and namespace imports, but it would be silly to
-        // produce two import declarations in this special case.
-        if (!isTypeOnly && defaultImports.length === 1 && namespaceImports.length === 1 && namedImports.length === 0) {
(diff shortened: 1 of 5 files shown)


src/services/organizeImports.ts
```diff
@@ -322,96 +322,105 @@ function coalesceImportsWorker(importGroup: readonly ImportDeclaration[], compar
...
-            // Add the namespace import to the existing default ImportDeclaration.
-            const defaultImport = defaultImports[0];
-            coalescedImports.push(
-                updateImportDeclarationAndClause(defaultImport, defaultImport.importClause.name, namespaceImports[0].importClause.namedBindings),
-            );
-
-            continue;
+    const importGroupsByAttributes = groupBy(importGroup, decl => {
+        if (decl.attributes) {
+            return decl.attributes.getFullText();
```
andrewbranch, line 327: Saw some discussion on Teams about this. As you were concerned about, this is sensitive to attribute order, spacing, comments, quote preference, etc. The grammar is simple enough that I think instead of doing something like emitting a normalized version of the text (which would still be sensitive to attribute order), you could pretty easily write something like

```ts
function getImportAttributesId(attributes: ImportAttributes) {
  const keyValuePairs = new Map<string, string>;
  // - iterate over attribute elements
  // - value must be a string literal expression if valid
  // - if invalid, insert something unique?
  let id = "";
  const sorted = sort(
    arrayFrom(keyValuePairs.entries()),
    ([keyA], [keyB]) => compareStringsCaseSensitive(keyA, keyB)
  );
  for (const [key, value] of sorted) {
    // value.length included in case `key` contains `:`,
    // a bit over the top but technically something like
    // this is needed
    id += `${key}:${value}:${value.length}|`;
  }
  return id;
}

File: src/services/organizeImports.ts [hunk 2]
```diff
@@ -322,96 +323,109 @@ function coalesceImportsWorker(importGroup: readonly ImportDeclaration[], compar
...
-            coalescedImports.push(
-                updateImportDeclarationAndClause(defaultImport, defaultImport.importClause.name, namespaceImports[0].importClause.namedBindings),
-            );
-
-            continue;
+    const importGroupsByAttributes = groupBy(importGroup, decl => {
+        if (decl.attributes) {
+            let attrs = decl.attributes.token + " ";
+            for (const x of sort(decl.attributes.elements)) {
+                attrs += x.name + ":" + x.value.getText() + " ";
```
andrewbranch, line 330: This is still sensitive to quote type since `getText()` includes quotes. Something like ``isStringLiteralLike(x.value) ? `"${x.value.text}"` :  x.value.getText()`` would be better—normalizing quotes without treating `"42"` and (the illegal value) `42` as the same.

tests/cases/fourslash/organizeImportsAttributes3.ts
```diff
@@ -0,0 +1,20 @@
+/// <reference path="fourslash.ts" />
+
+////import { A } from "./a";
+////import { C } from "./a" assert {      type: "a" };
+////import { Z } from "./z";
+////import { A as D } from "./a" assert    { type: "b" };
+////import { E } from "./a" assert { type: /* comment*/ "a"              };
+////import { F } from "./a" assert     {type: "a" };
+////import { Y } from "./a"   assert{ type: "b" /* comment*/};
```
andrewbranch, line 9: You could add another test showing that

```ts
import { A } from "./a" assert { foo: "foo", bar: "bar" };
import { B } from "./a" assert { bar: "bar", foo: "foo" };
```

can be combined since you sorted the elements.

iisaduan: Nice idea!

src/services/organizeImports.ts
```diff
@@ -322,96 +323,109 @@ function coalesceImportsWorker(importGroup: readonly ImportDeclaration[], compar
...
-            const defaultImport = defaultImports[0];
-            coalescedImports.push(
-                updateImportDeclarationAndClause(defaultImport, defaultImport.importClause.name, namespaceImports[0].importClause.namedBindings),
-            );
-
-            continue;
+    const importGroupsByAttributes = groupBy(importGroup, decl => {
+        if (decl.attributes) {
+            let attrs = decl.attributes.token + " ";
+            for (const x of sort(decl.attributes.elements)) {
```
andrewbranch, line 329: I’m not sure this `sort` is doing what you want. `elements` is an array of `Node` objects, which are going to get coerced to strings before sorting. Most likely every element is going to turn into `[object Object]` or `[object Node]` and the sort order will never change.

image

andrewbranch (approved):

PR #57251, opened 2024-01-31
Update .dprint.jsonc

Rishu2023:
Ensure .dprint.jsonc is up to date by changing https://plugins.dprint.dev/typescript-0.88.9.wasm to https://plugins.dprint.dev/typescript-0.88.10.wasm


jakebailey:
The right way to do this is with `dprint config update`, but I'm not sure it's worthwhile to accept a change until we actually need them. (I want to instead automate bumping it if there are no formatting changes + create an issue if it ever mismatches.)

PR #57255, opened 2024-02-01
Baseline Project Service State for easy diffing in changes

sheetalkamat:
This helps with testing my future changes as well


tests/baselines/reference/api/typescript.d.ts
```diff
@@ -3332,18 +3332,6 @@ declare namespace ts {
              * Last version that was reported.
```
jakebailey, line 3332: Unrelated to your PR, but this is somehow the first time that I've noticed that our dts emitter emits jsdoc on things marked private. Wow! (Surely that behavior isn't desirable?)

src/server/editorServices.ts
```diff
@@ -1139,6 +1139,7 @@ export class ProjectService {
     private pendingPluginEnablements?: Map<Project, Promise<BeginEnablePluginResult>[]>;
     private currentPluginEnablementPromise?: Promise<void>;
 
+    /** @internal */ baseline: (titile?: string) => void = noop;
```
jakebailey, line 1142:
```suggestion
    /** @internal */ baseline: (title?: string) => void = noop;
```

Maybe a typo?

jakebailey (approved):
I'll be honest and say that I don't quite know what kind of change this would detect, but it seems fine. 100k lines though... :D

sheetalkamat:
This helps in ensuring even though logs and order might change the state remains unchanged and its easier to detect whats going on after each request.

PR #57261, opened 2024-02-01
Use loader with extension substitution for package.json main, types, typesVersions

### LINKED ISSUE #56933: Bundler module resolution fails to resolve colocated declaration file
### 🔎 Search Terms

moduleresolution
bundler
directory
package.json 

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about "moduleResolution"

### ⏯ Playground Link

https://github.com/Methuselah96/test-ts-module-resolution

### 💻 Code

`./app/test.ts`

```ts
import { test } from '../lib';

test();
```

`./lib/package.json`

```json
{
  "main": "./cjs/index.js",
  "module": "./esm/index.js"
}
```

`./lib/cjs/index.js`

```js
export function test() {}
```
`./lib/cjs/index.d.ts`

```ts
export function test(): void;
```

`./lib/esm/index.js`

```js
export function test() {}
```
`./lib/esm/index.d.ts`

```ts
export function test(): void;
```

### 🙁 Actual behavior

```
app/test.ts:1:22 - error TS7016: Could not find a declaration file for module '../lib'. 'C:/Users/nbier/Documents/test-ts-module-resolution/lib/cjs/index.js' implicitly has an 'any' type.

1 import { test } from '../lib';
                       ~~~~~~~~


Found 1 error in app/test.ts:1
```

### 🙂 Expected behavior

Types should work.

`react-icons` use a directory structure like this. I encountered this issue while trying to switch to `"moduleResolution": "bundler"` with a dependency on `react-icons`.

andrewbranch:
Fixes #56933

Looking up a types/main/typesVersions path has always worked a little differently from other parts of module resolution.

Before:
- Try looking up the exact file name as written, e.g. `"main": "index.js"` would look for `index.js` before anything else.
- If the file exists but has an extension that doesn’t match what we were looking for, ignore it (e.g. if we only cared about declaration files, we’d discard the `.js` result). (Note that this made the previous file lookup unnecessary, since we could have known whether the request had a supported extension before hitting the file system.)
- Try the request like any other relative name, as if it were part of an `import` module specifier in a TS file.

After:
- If the file name has an TS/declaration extension, try the exact file lookup. (If this lookup fails, abort.)
- Otherwise, try replacing the extension with TS/declaration file equivalents (e.g. if it was `.js`, try `[".ts", ".tsx", ".d.ts"]`)
- Try the request like any other relative name, as if it were part of an `import` module specifier in a TS file.

This fixes an issue where all we had to go on was `"main": "index.js"`, when a JavaScript file was a valid extension to resolve, but not as good of a result as declaration file. We always need to prioritize the TS/declaration extensions over the JavaScript extension.


andrewbranch:
@typescript-bot test top200
@typescript-bot perf test

tests/baselines/reference/moduleResolution/non-relative-uses-originalPath-for-caching.js
```diff
@@ -31,8 +31,7 @@ Resolution:: {
     "/sub/dir/node_modules/@types/a/index.d.ts",
     "/sub/node_modules/a.ts",
     "/sub/node_modules/a.tsx",
-    "/sub/node_modules/a.d.ts",
-    "/sub/node_modules/a/index"
```
weswigham, line 35: We still want these paths in the failed lookup locations, yeah? They're places we need to watch for changes that could affect resolution, right? And since if a directory appears at this path, it'd potentially load something, we do actually need to monitor it?

andrewbranch: No—`/sub/node_modules/a.d.ts` is still present, but we removed `/sub/node_modules/a/index` because we wouldn’t have loaded it even if it existed, since it has no extension. `/sub/node_modules/a/index` as a directory would also not be relevant here, since the original request was for `"a"`.

weswigham: Right, it was for `"a"`, which found the package json at `/sub/node_modules/a/` and opened it up and found a `main`, which in turn is looking for `./index` (which is ultimately satisfied by `/sub/node_modules/a/index.ts`), _which could also be satisfied by a directory `/sub/node_modules/a/index` with a package.json in it_. That's the major difference between `main` resolution and exports - exports ignores package files in directories you point at because it's supposed to only look at specific files, while `main` can end up recursively looking up multiple `package.json`s that all refer to one another.

andrewbranch: The `/sub/node_modules/a/index` you see removed here was tried as a _file_ (again, for no reason—we would never load it), and still would have failed on that lookup if it existed as a directory. `/sub/node_modules/a/index.ts` _did_ exist and has higher precedence than anything inside the hypothetical `/sub/node_modules/a/index/` directory.

Only if `index.ts`, `index.tsx`, and `index.d.ts` all were missing would we then proceed into `/sub/node_modules/a/index/`, if it existed. In there, we would look for a package.json, or more index files, and you would see _those_ file names (e.g. `/sub/node_modules/a/index/package.json` appear here in `failedLookupLocations`. `failedLookupLocations` does not contain directory paths, but even if it did, it should not appear in this particular test case because we successfully resolved something at a higher priority than anything in such a directory would be.

weswigham: Oh, you're right, these are all file locations. I guess even though node would load an extensionless file at that location, we wouldn't since we don't know if it's js or not. But it feels like that aughta block resolution, regardless, rather than fallthrough.

Thinking about it, we technically support declaration files for file with arbitrary extensions - would an extensionless `index` file just a `index.d..ts` file, then?

andrewbranch: That might be something to explore. And yeah, I’ve brought up something similar in the past—there are many cases where we continue module resolution after Node.js would have stopped, often because (even before this PR, in other lookup paths) we don’t check for file existence of stuff that we know we can’t load. For example, if the request is for `"./foo.css"`, in CJS mode, we’ll look for `./foo.d.css.ts`, but if it doesn’t exist, we move onto trying `./foo.css.{ts,tsx,d.ts,js}`, even though we really shouldn’t if `/foo.css` itself exists.

weswigham (approved):
Honestly, I'm not even sure how node handles extensionless files nowadays, considering it uses the extension for format disambiguation. I *think* it still interprets extensionless as the same format as `.js`, but... honestly, the only reason you ever see it is for js shell scripts doubling as modules, which is pretty niche.

sheetalkamat (approved):

Issue #50187, opened 2022-08-04
Bring along imports when copy and pasting code across files


mjbvz:
# Feature request

From https://github.com/microsoft/vscode/issues/66236

When copy and pasting code between two TS/JS files, we should support adding the required imports on paste. For example, with the following code:

```ts
import fs from 'fs';
import http from 'http';

http.createServer((req, res) => {
    const data = fs.readFileSync('bla');
    res.write(data);
    res.end()
})
```

If I copy ` fs.readFileSync('bla')` and paste it into a new file, an import for `fs` should also be added to the top of that file:

# TS Server Protocol changes

VS Code is actively working on an API that lets extensions hook into copy and paste: https://github.com/microsoft/vscode/issues/30066

Using this api, here's what I imagine the flow between VS Code and TS server would look like:

1. User copies code

2. VS Code tells TS Server about the copied range

3. TS Server goes and writes whatever state it needs to re-create the imports. This state needs to be JSON serializable (so that TS Server can send it to VS Code). Just as an example,  this could be something like (note! just for demonstration purposes, not intended to be what this actually looks like):

    ```json
    {
        "referencedSymbols": [
             {"name": "default", "from": "node:fs"},
             {"name": "mySymbol", "from": "/full/path/to/file"}
        ]
    }
    ```

4. TS Server returns this state to the VS Code extension which attaches it to the copied text (this can happen asynchronously)

5. The user now pastes the code in another file

6. VS Code sends TS Server a request with the following:

    - The paste location
    - The saved off state from step 3 / 4 (note that we may actually have to await this data if the server was slow to return here
   

7. TS Server returns a text edit that adds the imports to the file

elliotwaite:
The [Copy With Imports](https://marketplace.visualstudio.com/items?itemName=stringham.copy-with-imports) extension seems to work pretty well at addressing this issue.

Thanks to @lukeapage for mentioning this extension in https://github.com/microsoft/vscode/issues/30066.

dantman:
The Copy With Imports extension works by overriding the copy and paste keybindings in a kind of hacky way. And having VSCode add an extension into copying introduces a delay that can result in the wrong line of code being copied if you don't wait for the copy command to be invoked.

Copy With Imports as an extension is an ok stopgap. But this should definitely be handled natively in TypeScript and VSCode.

elliotwaite:
[...]

Issue #57257, opened 2024-02-01
moveToFile does not combine imports 


iisaduan:
### 🔎 Search Terms

moveToFile, multiple auto imports, import order

### 🕗 Version & Regression Information

This is the behavior in every version I tried

### ⏯ Playground Link

No playground for moveToFile

### 💻 Code

```
// @Filename: /bar.ts
import { a, z } from './a';
const q = 0;
```
```
// @Filename: /a.ts
import { } from './other';
import type { } from './other';

export const p = 0;
export const b = 1;
export const a = 2;
export const z = 3;
[|const y = p + b;|]
```
Move the selection in `a.ts` to `bar.ts`

### 🙁 Actual behavior

A separate import statement is generated.
```
//// @Filename: /bar.ts
import { p, b } from './a';
import { a, z } from './a';
const q = 0;
const y = p + b;
```

### 🙂 Expected behavior

The named imports should combine into one statement: 
```
// @Filename: /bar.ts
import { a, b, p, z } from './a';
const q = 0;
const y = p + b;
```

### Additional information about the issue

I found this while working with auto-imports and playing with actions that generate multiple imports. The ideal order of imports should be `a, b, p, z`, as the imports in the original file were sorted.  From https://github.com/microsoft/TypeScript/issues/33839#issuecomment-642294737, and the current behavior of the `add all missing imports` codefix, `p, b` should be generated only if the original imports are not sorted.

DanielRosenwasser:
Duplicate of #54309?

iisaduan:
@DanielRosenwasser Not quite, the imports here should be added, they just generate in a weird order. Its possible the fix for both is related.

RyanCavanaugh:
Closing language service bugs related to the 6.0 implementation. For more information, see #62827

PR #57262, opened 2024-02-01
Implementing copy/paste

navya9singh:
Fixes https://github.com/microsoft/TypeScript/issues/50187


src/services/codefixes/importFixes.ts
```diff
@@ -1002,14 +1009,20 @@ function getFixInfos(context: CodeFixContextBase, errorCode: number, pos: number
...
+export function sortFixInfo(fixes: readonly (FixInfo & { fix: ImportFixWithModuleSpecifier; })[], sourceFile: SourceFile, program: Program, packageJsonImportFilter: PackageJsonImportFilter, host: LanguageServiceHost): readonly (FixInfo & { fix: ImportFixWithModuleSpecifier; })[] {
     const _toPath = (fileName: string) => toPath(fileName, host.getCurrentDirectory(), hostGetCanonicalFileName(host));
     return sort(fixes, (a, b) =>
         compareBooleans(!!a.isJsxNamespaceFix, !!b.isJsxNamespaceFix) ||
         compareValues(a.fix.kind, b.fix.kind) ||
         compareModuleSpecifiers(a.fix, b.fix, sourceFile, program, packageJsonImportFilter.allowsImportingSpecifier, _toPath));
 }
 
+function getFixInfosWithoutDiagnostic(context: CodeFixContextBase, symbolToken: Identifier,  useAutoImportProvider: boolean): readonly FixInfo[] | undefined {
+    const info = getFixesInfoForNonUMDImport(context, symbolToken, useAutoImportProvider);
```
navya9singh, line 1021: In here, do I also need to check for UMDImports using `getFixesInfoForUMDImport()` ?

src/harness/client.ts
```diff
@@ -1009,6 +1012,22 @@ export class SessionClient implements LanguageService {
         return getSupportedCodeFixes();
     }
 
+    getPostPasteImportFixes(targetFile: string, pastes: Array<{text: string; range: TextRange}>, _preferences: UserPreferences, _formatOptions: FormatCodeSettings, originalFile?: string, copyRange?: CopyRange): PostPasteImportFixes{    
+        const args = this.createFileLocationOrRangeRequestArgs(pastes[0].range, targetFile) as protocol.GetPostPasteImportFixesRequestArgs;
```
navya9singh, line 1016: I'm not sure what the range should be passed in here to `createFileLocationOrRangeRequestArgs` . Is this going to be the position that the paste is called at and passes in as a separate parameter in the api?

tests/cases/fourslash/server/postPasteImportFix_multiplePastes2.ts
```diff
@@ -0,0 +1,54 @@
...
+const a = 1;
+export const t = aa + bb + r + s;
+const u = 1;
+const c = 3;
+export const t = aa + bb + r + s;
+const u = 1;
+`
+    }, 
+    originalFile: "file1.ts",
+    copyRange: { start: { line : 3, offset: 0}, end : { line: 4, offset: 12}},
```
navya9singh, line 53: This is currently pasting same text at different cursor locations but it can be modified to paste different text at different locations as well. In that case, if the original file from which the code is copied is known, then  the `copyRange` will have to be an Array that points the correct copy range for every text. Is that something that we want to change?

src/server/protocol.ts
```diff
@@ -630,6 +631,32 @@ export interface GetMoveToRefactoringFileSuggestions extends Response {
...
+/**
+ * Request refactorings at a given position post pasting text from some other location.
+ */
+
+export interface GetPostPasteImportFixesRequest extends Request {
+    command: CommandTypes.GetPostPasteImportFixes;
+    arguments: GetPostPasteImportFixesRequestArgs;
+}
+export type DocumentRange = FileRangeRequestArgs;
+export type CopyRange  = {
```
mjbvz, line 643: Can this be a `TextSpan`? Maybe also combine it with `orignalFile` and use a `FileSpan`

src/server/protocol.ts
```diff
@@ -630,6 +631,32 @@ export interface GetMoveToRefactoringFileSuggestions extends Response {
...
+    start: Location;
+    end: Location;
+}
+export type GetPostPasteImportFixesRequestArgs = FileLocationOrRangeRequestArgs & {
+    targetFile: string, 
+    pastes: Array<{text: string; range: DocumentRange}>,
```
mjbvz, line 649: I think range here could also be a `TextSpan` since the file has already been provided

src/server/protocol.ts
```diff
@@ -630,6 +631,32 @@ export interface GetMoveToRefactoringFileSuggestions extends Response {
...
+export interface GetPostPasteImportFixesRequest extends Request {
+    command: CommandTypes.GetPostPasteImportFixes;
+    arguments: GetPostPasteImportFixesRequestArgs;
+}
+export type DocumentRange = FileRangeRequestArgs;
+export type CopyRange  = {
+    start: Location;
+    end: Location;
+}
+export type GetPostPasteImportFixesRequestArgs = FileLocationOrRangeRequestArgs & {
```
mjbvz, line 647: Maybe use `FileRequestArgs` instead and get rid of `targetFile` since that type already provides `file`

mjbvz (commented):
Just reviewed the protocol changes. Direction looks good but had a few suggestions for using different types

src/server/project.ts
```diff
@@ -2212,6 +2218,20 @@ export abstract class Project implements LanguageServiceHost, ModuleResolutionHo
         return this.noDtsResolutionProject;
     }
 
+    /** @internal */
+    updateTargetFile(rootFile: string, targetFileText: string, pastedText: string): {updatedFile: SourceFile | undefined, updatedProgram: Program | undefined, originalProgram: ts.Program | undefined} {
```
sheetalkamat, line 2222: Why does other functionality needs to be in the services and why cant it be in projectService instead?

src/services/postPasteImportFix.ts
```diff
@@ -0,0 +1,143 @@
+import { addRange, append } from "../compiler/core";
+import { SourceFile, Statement, UserPreferences, TypeChecker, AnyImportOrRequireStatement, Identifier, ModifierFlags, Program, TextRange, CancellationToken, ImportClause, ObjectBindingPattern, TypeOnlyAliasDeclaration, SymbolFlags } from "../compiler/types";
+import { hasSyntacticModifier, skipAlias } from "../compiler/utilities";
+import { codefix, Debug, factory, fileShouldUseJavaScriptRequire, forEachChild, formatting, getQuotePreference, ImportKind, insertImports, nodeSeenTracker, Symbol, SymbolExportInfo, textChanges, isIdentifier } from "./_namespaces/ts";
+import { addExportToChanges, filterImport, forEachImportInStatement, getTopLevelDeclarationStatement, getUsageInfo, isTopLevelDeclaration, makeImportOrRequire, moduleSpecifierFromImport, nameOfTopLevelDeclaration } from "./refactors/moveToFile";
+import { CodeFixContextBase, CopyRange, FileTextChanges, LanguageServiceHost, PostPasteImportFixes } from "./types";
+
+/** @internal */
+export function postPastImportFixProvider(
```
zkat, line 9:
```suggestion
export function postPasteImportFixProvider(

```

src/services/postPasteImportFixes.ts
```diff
@@ -0,0 +1,187 @@
...
+                    // generate imports
+                    importAdder.addImportsForUnknownSymbols(context, node, /*useAutoImportProvider*/ true);
+                }
+            }
+            else {
+                node.forEachChild(cb);
+            }
+        });
+        importAdder.writeFixes(changes, getQuotePreference(targetFile, preferences));
+        host.revertUpdatedFile?.(targetFile.fileName, updatedTargetFile.updatedFile.text, targetFile.text);
```
andrewbranch, line 120: I’m concerned about an unexpected error getting thrown between `updateTargetFile` and `revertUpdatedFile`, leaving the server in a corrupted state for subsequent commands. I think the code needs to be structured very carefully like

```ts
// Comment explaining what's about to happen
updateFile()
try {
  doEverything()
} finally {
  revertFile()
}
```

By that I primarily mean you need a try/finally, but I also think it may be worth a little aesthetic refactoring to make the `try` block very short like my pseudocode example, so that this whole dangerous sequence can easily be digested by someone reading the code in the future.

src/server/project.ts
```diff
@@ -2215,6 +2215,20 @@ export abstract class Project implements LanguageServiceHost, ModuleResolutionHo
...
+        this.getScriptInfo(rootFile)?.editContent(0, targetFileText.length, pastedText);
+        this.updateGraph();
+        return { updatedFile: (this.program?.getSourceFile(rootFile)), updatedProgram: this.program, originalProgram };
+    }
+
+    /** @internal */
+    revertUpdatedFile(rootFile: string, updatedText: string, originalText: string) {
+        this.getScriptInfo(rootFile)?.editContent(0, updatedText.length, originalText);
+        this.updateGraph();
+    }
```
andrewbranch, line 2230: Two observations:

1. These two methods are the same, except one returns nothing. Just `updateTargetFile` would suffice.
2. You don’t need to pass in the current text of the file; you can get that from the SourceFile or ScriptInfo. You only need to pass in the desired text.

Combining this with my previous comment about needing a try/finally, maybe it would be best to make that part of the API here, so it’s impossible to forget to revert or handle an error. You can make an API like

```ts
host.runWithTemporaryFileUpdate(sourceFile, newText, (updatedProgram, updatedFile) => {
  doEverything();
});
```

I think that would be much better than having two separate methods that must be called symmetrically to avoid catastrophic results 😅

andrewbranch (changes requested):

src/server/protocol.ts
```diff
@@ -630,6 +631,26 @@ export interface GetMoveToRefactoringFileSuggestions extends Response {
...
+ */
+
+export interface GetPostPasteImportFixesRequest extends Request {
+    command: CommandTypes.GetPostPasteImportFixes;
+    arguments: GetPostPasteImportFixesRequestArgs;
+}
+
+export type GetPostPasteImportFixesRequestArgs = FileRequestArgs & {
+    pastes: { text: string; range: TextSpan; }[];
+    copySpan?: FileSpan;
```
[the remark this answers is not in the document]
navya9singh: For the first version of copy/paste the idea was to paste the same text at multiple cursors. So yes, with the current protocol if a single copy had to be pasted at 100 cursors, the text would be duplicated 100 times in `pastes`. Whenever later we decided to allow copying different text at multiple cursors and pasting at multiple cursor locations, the protocol would have to change to `copySpan?: FileSpan[]`.
Looking at your mental model, I think that seems to make this clearer 😄 I'd be happy to change this part of the protocol if that's more useful!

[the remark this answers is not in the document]
andrewbranch: The protocol can’t really change; we can only add to it. So even if we ship an implementation that doesn’t cover all the use cases we’d eventually like to support, the protocol design needs to be somewhat future proof. (That said, from discussions so far, it didn’t sound like it will be very hard to implement a paste provider that’s faithful to the way VS Code multi-cursor paste usually works. Since this isn’t supposed to land until 5.5, I think there’s likely plenty of time to make that happen.)

src/server/project.ts
```diff
@@ -2215,6 +2215,20 @@ export abstract class Project implements LanguageServiceHost, ModuleResolutionHo
...
+        const originalProgram = this.program;
+        this.getScriptInfo(rootFile)?.editContent(0, targetFileText.length, pastedText);
+        this.updateGraph();
+        return { updatedFile: (this.program?.getSourceFile(rootFile)), updatedProgram: this.program, originalProgram };
+    }
+
+    /** @internal */
+    revertUpdatedFile(rootFile: string, updatedText: string, originalText: string) {
+        this.getScriptInfo(rootFile)?.editContent(0, updatedText.length, originalText);
+        this.updateGraph();
```
sheetalkamat, line 2229: i dont think you should do updateGraph here.,..  (that should take place whenever its needed instead)

src/services/postPasteImportFixes.ts
```diff
@@ -0,0 +1,187 @@
...
+        try {
+            importAdder.addImportFromExportedSymbol(skipAlias(symbol, checker), isValidTypeOnlyUseSite);
+        }
+        catch {
+            for (const oldStatement of originalFile.statements) {
+                forEachImportInStatement(oldStatement, i => {
+                    append(copiedOldImports, filterImport(i, factory.createStringLiteral(moduleSpecifierFromImport(i).text), name => importsToCopy.has(checker.getSymbolAtLocation(name)!)));
+                });
+            }
+        }
```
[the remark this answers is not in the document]
navya9singh: I've moved this part of the code to utilities, but the `catch` part deals with unresolved imports. So, if an import statement is written for a file that doesn't exist, this part handles that case.

src/server/project.ts
navya9singh: If I skip updateGraph from here then the file does not get back to its original form, with the original text. So to ensure that happens, I added updateGraph here.

src/services/postPasteImportFixes.ts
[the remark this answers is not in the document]
andrewbranch: The nesting still seems wrong. For every import that doesn’t resolve, it looks like you’re adding a syntactic copy of every import in the file. So if you had a file where 10 imports resolve and 2 don’t, you’d end up with 34 imports in the new file: 10 that were rewritten successfully via the ImportAdder; 12 when one fails; 12 again when the other fails. Maybe these are deduplicated down the line, but it at least appears that this is looping and synthesizing import nodes far too aggressively when things fail.

src/server/project.ts
sheetalkamat: The file will be updated at next update.

src/services/postPasteImportFixes.ts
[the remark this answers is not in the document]
navya9singh: Yeah, you're right I see how imports are getting duplicated here. I'll create an issue for this as a bug and fix it because I need to add more tests for it as well.

mjbvz (commented):
Update protocol changes look good to me

src/server/project.ts
```diff
@@ -2214,6 +2214,18 @@ export abstract class Project implements LanguageServiceHost, ModuleResolutionHo
         return this.noDtsResolutionProject;
     }
 
+    /** @internal */
+    runWithTemporaryFileUpdate(rootFile: string, pastedText: string, cb: (updatedProgram: Program | undefined, originalProgram: Program | undefined, updatedFile: SourceFile) => void) {
```
andrewbranch, line 2218: Nit: seems like `pastedText` doesn’t make sense as a parameter name here anymore

src/harness/fourslashInterfaceImpl.ts
```diff
@@ -1883,6 +1887,13 @@ export interface VerifyCodeFixAllOptions {
     preferences?: ts.UserPreferences;
 }
 
+export interface VerifyPasteEdits {
```
andrewbranch, line 1890: Is this used?

src/services/types.ts
```diff
@@ -431,7 +431,7 @@ export interface LanguageServiceHost extends GetEffectiveTypeRootsHost, MinimalR
     getParsedCommandLine?(fileName: string): ParsedCommandLine | undefined;
     /** @internal */ onReleaseParsedCommandLine?(configFileName: string, oldResolvedRef: ResolvedProjectReference | undefined, optionOptions: CompilerOptions): void;
     /** @internal */ getIncompleteCompletionsCache?(): IncompleteCompletionsCache;
-
+    /** @internal */ runWithTemporaryFileUpdate?(rootFile: string, pastedText: string, cb: (updatedProgram: Program | undefined, originalProgram: Program | undefined, updatedPastedText: SourceFile) => void): void;
```
andrewbranch, line 434: Same param name appears here

src/services/types.ts
```diff
@@ -684,6 +684,13 @@ export interface LanguageService {
     getSupportedCodeFixes(fileName?: string): readonly string[];
 
     dispose(): void;
+    getPasteEdits(
+        targetFile: string,
+        copies: { text: string; range?: { file: string; range: TextRange; }; }[],
```
andrewbranch, line 689: It might be worth extracting this object type to an interface. Then you can avoid duplicating it in the fourslash code.

src/server/session.ts
```diff
@@ -2795,6 +2796,26 @@ export class Session<TMessage = string> implements EventSender {
...
 
+    private getPasteEdits(args: protocol.GetPasteEditsRequestArgs): protocol.PasteEditsAction | undefined {
+        const { file, project } = this.getFileAndProject(args);
+        const copyFile = args.copies[0].range ? args.copies[0].range.file : undefined;
+        const result = project.getLanguageService().getPasteEdits(
+            file,
+            args.copies.map(copy => ({
+                text: copy.text,
+                copyRange: copy.range && copyFile
+                    ? { file: copy.range.file, range: this.getRange({ file: copyFile, startLine: copy.range.start.line, startOffset: copy.range.start.offset, endLine: copy.range.end.line, endOffset: copy.range.end.offset }, project.getScriptInfoForNormalizedPath(toNormalizedPath(copyFile))!) } : undefined,
```
andrewbranch, line 2807: You’re assuming, but not checking, that every copy has the same file. I think this is necessarily true from VS Code, but the protocol suggests that they could be different. We should revise the protocol to fix that, handle differing files, or debug assert that all copies have the same file.

src/services/pasteEdits.ts
```diff
@@ -0,0 +1,103 @@
...
+    targetFile: SourceFile,
+    copies: { text: string; copyRange?: { file: SourceFile; range: TextRange; }; }[],
+    pastes: TextRange[],
+    host: LanguageServiceHost,
+    preferences: UserPreferences,
+    formatContext: formatting.FormatContext,
+    cancellationToken: CancellationToken,
+    changes: textChanges.ChangeTracker,
+) {
+    const copy = copies[0];
```
andrewbranch, line 61: Have we decided to ignore multiple copy locations? There is a lot of time left before 5.5 beta so I would advocate for shipping the correct behavior from the start. If not, are we tracking a follow-up task to finish implementing the expected multi-cursor semantics?

src/services/utilities.ts
```diff
@@ -4288,3 +4302,93 @@ export function isBlockLike(node: Node): node is BlockLike {
             return false;
     }
 }
+
+/** @internal */
+export function getTargetFileImportsAndAddExportInOldFile(
```
andrewbranch, line 4307: Maybe `src/services/refactors/helpers.ts` would be a better place for this, judging by the number of new imports from `ts.refactor`.

andrewbranch (changes requested):

src/services/pasteEdits.ts
navya9singh: Yeah, the last time I spoke to @mjbvz it was decided that for the first version we can keep pasting the same text, but yes, there's definitely time right now before 5.5 beta, so I'll instead implement the correct behavior here.

src/server/protocol.ts
```diff
@@ -161,6 +161,7 @@ export const enum CommandTypes {
     GetApplicableRefactors = "getApplicableRefactors",
     GetEditsForRefactor = "getEditsForRefactor",
     GetMoveToRefactoringFileSuggestions = "getMoveToRefactoringFileSuggestions",
+    GetPasteEdits = "GetPasteEdits",
```
mjbvz, line 164: nit: lowercase to match other commands

src/server/protocol.ts
```diff
@@ -625,6 +626,29 @@ export interface GetMoveToRefactoringFileSuggestions extends Response {
...
+    copies: {
+        text: string;
+        range?: FileSpan;
+    }[];
+    pastes: TextSpan[];
+};
+export interface GetPasteEditsResponse extends Response {
+    body: PasteEditsAction;
+}
+export interface PasteEditsAction {
```
mjbvz, line 648: For the future we will likely want a `name` and/or `id` too so users can tell what type of edit is being made. We can leave that out until there are multiple providers, but means that we'll have to add a flag so older clients don't unexpectedly start using the wrong paste edit

src/services/pasteEdits.ts
andrewbranch: AFAIK multiple copies can only happen with multiple cursors in a single file, but the protocol I suggested seems to imply that it’s possible for the copy sources to be different files. I think this was a mistake in my suggestion 😕

navya9singh: I see, so what about something like:
```
export type GetPasteEditsRequestArgs = FileRequestArgs & {
    copies: {
        text: string;
        range?: TextSpan;
    }[];
    copyFromFile?: string;
    pastes: TextSpan[];
};
```
or 
```
export type GetPasteEditsRequestArgs = FileRequestArgs & {
    copiesText: string[];
    copiesFrom?:{ file: string, range: TextSpan[] };
    pastes: TextSpan[];
};
```

andrewbranch: @mjbvz any input?

mjbvz: Proposal:

```ts
export type GetPasteEditsRequestArgs = FileRequestArgs & {
    copiesFrom?:{ file: string, range: TextSpan[] };
    pastedText: string[];
    pastes: TextSpan[];
};
```

navya9singh:
This commit 7027ea31d1b5911fd46e8ea4f4c33fe87ed24611 fixes a possible error for this feature, where named imports could get duplicated if there was an import statement for the same file in the target file https://github.com/microsoft/TypeScript/issues/57257

mjbvz (approved):

src/server/protocol.ts
```diff
@@ -625,6 +626,28 @@ export interface GetMoveToRefactoringFileSuggestions extends Response {
...
+export interface GetPasteEditsRequest extends Request {
+    command: CommandTypes.GetPasteEdits;
+    arguments: GetPasteEditsRequestArgs;
+}
+
+export type GetPasteEditsRequestArgs = FileRequestArgs & {
+    pastedText: string[];
+    pasteLocations: TextSpan[];
+    copiedFrom?: { file: string; range: TextSpan[]; };
+};
```
DanielRosenwasser, line 642:
```suggestion
export interface GetPasteEditsRequestArgs extends FileRequestArgs {
    pastedText: string[];
    pasteLocations: TextSpan[];
    copiedFrom?: { file: string; range: TextSpan[]; };
}

```

src/server/protocol.ts
```diff
@@ -625,6 +626,28 @@ export interface GetMoveToRefactoringFileSuggestions extends Response {
...
+
+export interface GetPasteEditsRequest extends Request {
+    command: CommandTypes.GetPasteEdits;
+    arguments: GetPasteEditsRequestArgs;
+}
+
+export type GetPasteEditsRequestArgs = FileRequestArgs & {
+    pastedText: string[];
+    pasteLocations: TextSpan[];
+    copiedFrom?: { file: string; range: TextSpan[]; };
```
DanielRosenwasser, line 641: Can there be multiple possible files that are being copied from? (niche, I know)

src/server/protocol.ts
```diff
@@ -625,6 +626,28 @@ export interface GetMoveToRefactoringFileSuggestions extends Response {
...
+};
+export interface GetPasteEditsResponse extends Response {
+    body: PasteEditsAction;
+}
+export interface PasteEditsAction {
```
DanielRosenwasser, line 646:
```suggestion

export interface PasteEditsAction {
```

src/server/session.ts
```diff
@@ -2796,6 +2797,27 @@ export class Session<TMessage = string> implements EventSender {
...
+        const { file, project } = this.getFileAndProject(args);
+        const copyFile = args.copiedFrom ? args.copiedFrom.file : undefined;
+        const result = project.getLanguageService().getPasteEdits(
+            {
+                targetFile: file,
+                pastedText: args.pastedText,
+                pasteLocations: args.pasteLocations.map(paste => this.getRange({ file, startLine: paste.start.line, startOffset: paste.start.offset, endLine: paste.end.line, endOffset: paste.end.offset }, project.getScriptInfoForNormalizedPath(file)!)),
+                copiedFrom: args.copiedFrom && copyFile
+                    ? { file: args.copiedFrom.file, range: args.copiedFrom.range.map(copies => this.getRange({ file: copyFile, startLine: copies.start.line, startOffset: copies.start.offset, endLine: copies.end.line, endOffset: copies.end.offset }, project.getScriptInfoForNormalizedPath(toNormalizedPath(copyFile))!)) }
+                    : undefined,
```
DanielRosenwasser, line 2810: Break this into its own variable

src/server/session.ts
```diff
@@ -2796,6 +2797,27 @@ export class Session<TMessage = string> implements EventSender {
         return project.getLanguageService().getMoveToRefactoringFileSuggestions(file, this.extractPositionOrRange(args, scriptInfo), this.getPreferences(file));
     }
 
+    private getPasteEdits(args: protocol.GetPasteEditsRequestArgs): protocol.PasteEditsAction | undefined {
+        const { file, project } = this.getFileAndProject(args);
+        const copyFile = args.copiedFrom ? args.copiedFrom.file : undefined;
```
DanielRosenwasser, line 2802:
```suggestion
        const copySource = args.copiedFrom?.file;
```

src/server/session.ts
```diff
@@ -2796,6 +2797,27 @@ export class Session<TMessage = string> implements EventSender {
...
 
+    private getPasteEdits(args: protocol.GetPasteEditsRequestArgs): protocol.PasteEditsAction | undefined {
+        const { file, project } = this.getFileAndProject(args);
+        const copyFile = args.copiedFrom ? args.copiedFrom.file : undefined;
+        const result = project.getLanguageService().getPasteEdits(
+            {
+                targetFile: file,
+                pastedText: args.pastedText,
+                pasteLocations: args.pasteLocations.map(paste => this.getRange({ file, startLine: paste.start.line, startOffset: paste.start.offset, endLine: paste.end.line, endOffset: paste.end.offset }, project.getScriptInfoForNormalizedPath(file)!)),
+                copiedFrom: args.copiedFrom && copyFile
```
DanielRosenwasser, line 2808: `copyFile` will only be defined (though possibly falsy) if `args.copiedFrom` is itself defined.

src/services/codefixes/importFixes.ts
```diff
@@ -623,6 +647,30 @@ function getSingleExportInfoForSymbol(symbol: Symbol, symbolName: string, module
...
+): { computedWithoutCacheCount: number; fixes: readonly ImportFixWithModuleSpecifier[]; } {
+    const checker = program.getTypeChecker();
+    const symbolInfo = [{
+        symbol,
+        targetFlags: SymbolFlags.Value,
+        importKind,
+        targetFilePath,
+    }];
+    const existingImports = flatMap(symbolInfo, importMap.getImportsForNonExportedSymbols);
+    const addToExisting = tryAddToExistingImport(existingImports, isValidTypeOnlyUseSite, checker, program.getCompilerOptions());
```
DanielRosenwasser, line 667: Assert that this is defined so that it is easy to track later on in case something goes wrong.

src/services/codefixes/importFixes.ts
```diff
@@ -623,6 +647,30 @@ function getSingleExportInfoForSymbol(symbol: Symbol, symbolName: string, module
...
+    return {
+        computedWithoutCacheCount: 0,
```
DanielRosenwasser, line 669: Where is this used?

src/services/codefixes/importFixes.ts
```diff
@@ -823,6 +871,23 @@ function createExistingImportMap(checker: TypeChecker, importingFile: SourceFile
...
+            Debug.assertIsDefined(values);
+            for (const value of values) {
+                value.forEach(v => {
+                    if (v.kind === SyntaxKind.ImportDeclaration && cast(v.moduleSpecifier, isStringLiteral).text === targetFilePath) {
+                        fix = [{ declaration: v, importKind, targetFlags, symbol }];
+                        return fix;
+                    }
+                    else {
+                        return emptyArray;
+                    }
```
DanielRosenwasser, line 886: There's no point to returning, the return value is never witnessed and it doesn't make `forEach` stop traversing.

On that note, was it your intention to stop at the first value? Because the last value will always win out in writing to `fix` in this code.

src/services/pasteEdits.ts
```diff
@@ -0,0 +1,130 @@
...
+    const statements: Statement[] = [];
+
+    let start = 0;
+    let newText = "";
+    pasteLocations.forEach((location, i) => {
+        if (i === pasteLocations.length - 1) {
+            newText += targetFile.getText().slice(start, location.pos) + pastedText[i] + targetFile.getText().slice(location.end);
+        }
+        else {
+            newText += targetFile.getText().slice(start, location.pos) + pastedText[i];
```
DanielRosenwasser, line 76:
```suggestion
            newText += targetFile.text.slice(start, location.pos) + pastedText[i] + targetFile.text.slice(location.end);
        }
        else {
            newText += targetFile.text.slice(start, location.pos) + pastedText[i];
```

DanielRosenwasser (changes requested):

src/server/protocol.ts
andrewbranch: No—I guess you were out when we discussed this, but we redesigned the protocol again specifically to prevent the API from being able to represent that.

src/services/codefixes/importFixes.ts
```diff
@@ -959,7 +937,7 @@ function getNewImportFixes(
     let computedWithoutCacheCount = 0;
     const fixes = flatMap(exportInfo, (exportInfo, i) => {
         const checker = getChecker(exportInfo.isFromPackageJson);
-        const { computedWithoutCache, moduleSpecifiers } = getModuleSpecifiers(exportInfo.moduleSymbol, checker);
+        const { computedWithoutCache, moduleSpecifiers } = targetFileImportFrom ? { computedWithoutCache: true, moduleSpecifiers: targetFileImportFrom } : getModuleSpecifiers(exportInfo.moduleSymbol, checker);
```
andrewbranch, line 940: This `targetFileImportFrom` seems weird to me. What would it take to compute the module specifier here instead of in `moveToFile`?

src/services/codefixes/importFixes.ts
navya9singh: Removed this part now.

src/services/codefixes/importFixes.ts
navya9singh: For tests like `tests/cases/fourslash/moveToFile_existingImports5.ts` where a symbol exists in `b.ts` and is being moved to a different file, say  `a.ts`. There is a possibility that it needs to be imported by `b.ts`. This is where I was using the `targetFileImportFrom` to replace the module specifier with `a.ts` to create something like `import { Foo } from './a'`.

In `addImportFromSymbol()`, for the case where a symbol doesn't have a module specifier, I assign `modulesFileName` to  `targetFilePath`, which is `a.ts` and I can do something similar for the else case in there too to the `exportInfo`. Then in `getNewImportFixes()` I can compare if `moduleSpecifiers !== exportInfo.moduleFileName` then replace `moduleSpecifiers` with the `moduleFileName`.
Would that look better?

src/services/codefixes/importFixes.ts
navya9singh: Removed this part of the code now.

src/services/codefixes/importFixes.ts
navya9singh: Which one are you pointing at?

src/services/pasteEdits.ts
```diff
@@ -0,0 +1,123 @@
...
+    let newText = "";
+    pasteLocations.forEach((location, i) => {
+        if (i === pasteLocations.length - 1) {
+            newText += targetFile.text.slice(start, location.pos) + pastedText[i] + targetFile.text.slice(location.end);
+        }
+        else {
+            newText += targetFile.text.slice(start, location.pos) + pastedText[i];
+            start = location.end;
+        }
+    });
```
andrewbranch, line 73: I think you can avoid this if/else and `start` tracking by applying the edits in reverse order (sorted by insertion position).

src/services/pasteEdits.ts
```diff
@@ -0,0 +1,123 @@
...
+                    if (!originalProgram?.getTypeChecker().resolveName(node.text, node, SymbolFlags.All, /*excludeGlobals*/ false)) {
+                        // generate imports
+                        importAdder.addImportForUnresolvedIdentifier(context, node, /*useAutoImportProvider*/ true);
+                    }
+                }
+                else {
+                    node.forEachChild(cb);
+                }
+            });
+            importAdder.writeFixes(changes, getQuotePreference(targetFile, preferences));
```
andrewbranch, line 108: Nit: both `importAdder.writeFixes` calls can be extracted out of their if/else blocks into a single call afterwards

src/services/pasteEdits.ts
```diff
@@ -0,0 +1,123 @@
...
+                sourceFile: updatedFile,
+                program: originalProgram!,
+                cancellationToken,
+                host,
+                preferences,
+                formatContext,
+            };
+            forEachChild(updatedFile, function cb(node) {
+                if (isIdentifier(node)) {
+                    if (!originalProgram?.getTypeChecker().resolveName(node.text, node, SymbolFlags.All, /*excludeGlobals*/ false)) {
```
andrewbranch, line 99: Nit: combine nested `if` conditions. I think `node.forEachChild(cb)` at the end doesn't need to be in an `else` since it should do nothing for Identifiers (which don't have children)

src/services/pasteEdits.ts
```diff
@@ -0,0 +1,123 @@
...
+                    node.forEachChild(cb);
+                }
+            });
+            importAdder.writeFixes(changes, getQuotePreference(targetFile, preferences));
+        }
+    });
+
+    if (pastedText.length !== 1) {
+        Debug.assert(pastedText.length === pasteLocations.length);
+    }
```
andrewbranch, line 114: Maybe this should happen at the beginning? VS Code can handle this scenario, though I think it's fine if we don't. Will the extension ensure it doesn't ask us for paste edits when this is true?

navya9singh: The last time I checked with Matt, he mentioned that he wanted these edits along with the import edits that we provide, but I can remove if needed.
As for this being true, I believe that this would be checked on the vscode side, @mjbvz ?

andrewbranch: Yeah, you are supposed to apply these edits; I remember Matt clarifying that. I meant that VS Code has no problem pasting two copy locations (`pastedText.length === 2`) into three paste locations (`pasteLocations.length === 3`). What happens is that each `pasteLocation` gets `pastedText.join("\n")` inserted, but we're throwing.

src/services/pasteEdits.ts
```diff
@@ -0,0 +1,115 @@
...
+
+    const statements: Statement[] = [];
+
+    let end = targetFile.text.length - 1;
+    let newText = "";
+    for (let i = pasteLocations.length - 1; i >= 0; i--) {
+        newText = pastedText[i] + targetFile.text.slice(pasteLocations[i].end, end) + newText;
+        end = pasteLocations[i].pos;
+    }
+    newText = targetFile.text.slice(0, end) + newText;
```
andrewbranch, line 81: Here's what I meant.

```suggestion
    let newText = targetFile.text;
    for (let i = pasteLocations.length - 1; i >= 0; i--) {
        const { pos, end } = pasteLocations[i];
        newText = newText.slice(0, pos) + pastedText[i] + newText.slice(end);
    }
```

src/services/codefixes/importFixes.ts
```diff
@@ -894,7 +894,7 @@ function getSingleExportInfoForSymbol(symbol: Symbol, symbolName: string, module
 }
 
 function isFutureSymbolExportInfoArray(info: readonly SymbolExportInfo[] | readonly FutureSymbolExportInfo[]): info is readonly FutureSymbolExportInfo[] {
-    return info[0].symbol === undefined;
+    return info[0].moduleSymbol === undefined;
```
andrewbranch, line 897: This looks incorrect. `moduleSymbol` is a required non-nullable property in `FutureSymbolExportInfo`; `symbol` must be `undefined`

tests/cases/fourslash/server/pasteEdits_knownSourceFile.ts
```diff
@@ -0,0 +1,47 @@
...
+////const a = 1;
+////[|const c = a + b;
+////const t = 9;|]
+
+// @Filename: /tsconfig.json
+////{ "files": ["file1.ts", "file2.ts", "target.ts"] }
+
+const range = test.ranges();
+const t = range[0];
+format.setOption("insertSpaceAfterSemicolonInForStatements", true);
```
andrewbranch, line 22: What is this for?

src/server/protocol.ts
```diff
@@ -625,6 +626,30 @@ export interface GetMoveToRefactoringFileSuggestions extends Response {
...
+
+export interface GetPasteEditsRequest extends Request {
+    command: CommandTypes.GetPasteEdits;
+    arguments: GetPasteEditsRequestArgs;
+}
+
+export interface GetPasteEditsRequestArgs extends FileRequestArgs {
+    pastedText: string[];
+    pasteLocations: TextSpan[];
+    copiedFrom?: { file: string; range: TextSpan[]; };
```
andrewbranch, line 641: Last protocol naming suggestion 😅 

_Almost_ everything is consistent about calling a `TextSpan` a `span` or a `location` since internally, a range is something else. And the singular/plural mismatch here was really confusing when I was reading the code elsewhere.

```suggestion
    /** The source location of each `pastedText`. If present, the length of `spans` must be equal to the length of `pastedText`. */
    copiedFrom?: { file: string; spans: TextSpan[]; };
```

Additional comments would be good too.

tests/cases/fourslash/server/pasteEdits_existingImports1.ts
```diff
@@ -0,0 +1,39 @@
...
+    newFileContents: {
+        "/target.ts":
+`import { t } from "./other";
+import { t2 } from "./other2";
+import { t3 } from "./other3";
+const a = t + 1;
+const m = t3 + t2 + 1;
+const c = 10;`
+    },
+    fixId: "providePostPasteEdits"
```
andrewbranch, line 38: Is there any reason to pass this string into the `verify` call in every test?

src/testRunner/unittests/tsserver/pasteEdits.ts
```diff
@@ -0,0 +1,44 @@
...
+
+        session.executeCommandSeq<ts.server.protocol.GetPasteEditsRequest>({
+            command: ts.server.protocol.CommandTypes.GetPasteEdits,
+            arguments: {
+                file: target.path,
+                pastedText: [pastedText],
+                pasteLocations: [{ start: { line: 2, offset: 0 }, end: { line: 2, offset: 0 } }],
+            },
+        });
+        baselineTsserverLogs("pasteEdits", "adds paste edits", session);
```
sheetalkamat, line 41:
```suggestion
        verifyGetErrRequest({ session, files: [target.path] });
        baselineTsserverLogs("pasteEdits", "adds paste edits", session);
```

src/testRunner/unittests/tsserver/pasteEdits.ts
```diff
@@ -0,0 +1,44 @@
...
+        };
+        const tsconfig: File = {
+            path: "/project/tsconfig.json",
+            content: "{}",
+        };
+        const pastedText = `const q = 1;
+function e();
+const f = r + s;`;
+
+        const host = createServerHost([target, tsconfig]);
```
sheetalkamat, line 29:
```suggestion
        const host = createServerHost([target, tsconfig, libFile]);
```

tests/cases/fourslash/server/pasteEdits_knownSourceFile.ts
navya9singh: I get an extra symbol in the server tests if  I don't use this 
image

navya9singh: I see, I'll take a look at this.

tests/cases/fourslash/server/pasteEdits_existingImports1.ts
```diff
@@ -0,0 +1,38 @@
...
+
+// @Filename: /tsconfig.json
+////{ "files": ["target.ts", "other.ts", "other2.ts", "other3.ts"] }
+
+const range = test.ranges();
+format.setOption("insertSpaceAfterSemicolonInForStatements", true);
+verify.pasteEdits({
+    args: {
+        pastedText: [ `const m = t3 + t2 + 1;`],
+        pasteLocations: [{ pos: range[0].pos, end: range[0].end }],
```
sheetalkamat, line 27:
```suggestion
        pasteLocations: range,
```

Isnt it just this

sheetalkamat: oh you dont want file name ? got it

tests/cases/fourslash/server/pasteEdits_knownSourceFile.ts
navya9singh: Here, `TestState.formatCodeSetting` had  `newLineCharacter` as `\n` but `ProjectService.hostConfiguration.formatCodeOption` would have `newLineCharacter` as `\r\n`. So adding `this.setFormattingOptions(formatOptions);` in client.ts made it `\n` again.

andrewbranch (approved):

sheetalkamat (approved):

sheetalkamat:
@DanielRosenwasser @navya9singh  wasnt able to merge because it was marked as pending your review and you are oof. And the feedback seems to be addressed

PR #57266, opened 2024-02-01
Allow plugins to be loaded from package subpaths

### LINKED ISSUE #57047: TSServer Plugins with sub-packages/paths are not loaded
### 🔎 Search Terms

"language server", "because only package name is allowed plugin name", "language server plugins"

This behaviour must be going back to 4.2.2 as per #43124.

### ⏯ Playground Link

_No response_

### 💻 Code

**tsconfig.json**

```json
{
  "compilerOptions": {
    "plugins": [
      {
        "name": "package/lsp"
      }
    ]
  }
}
```

### 🙁 Actual behavior

TypeScript rejects sub-path package plugins, as per:
- https://github.com/microsoft/TypeScript/blob/23faef92703556567ddbcb9afb893f4ba638fc20/src/server/editorServices.ts#L4490-L4493
- and; https://github.com/microsoft/TypeScript/blob/23faef92703556567ddbcb9afb893f4ba638fc20/src/compiler/moduleNameResolver.ts#L2525-L2531

**tsserver log**

```
Info 11   [15:52:00.931] Enabling plugin package/subpath from candidate paths: [...]node_modules/typescript/lib/tsserver.js/../../..
Info 12   [15:52:00.931] Skipped loading plugin package/lsp because only package name is allowed plugin name
```

### 🙂 Expected behavior

Plugins are loaded as per either `package.json:exports` with subfolders specified, or via `package/sub-path/package.json` or `package/sub-path/index.js`.

Related issues: #42688 and #42688.

If my understanding is correct, TypeScript only wants to allow packages that:
- have been installed via a package manager, so that opening VSCode without installing doesn't immediately run unauthorised code
- are not relative paths or local code that's checked into a repository

Neither applies to sub-packages. The use-case here is to publish packages that both provide code _and_ an LSP package in one, without splitting this into two packages because of the assertion in TypeScript.

kitten:
Fixes #57047

When merged, this allows plugins to also be loaded from package subpaths.

**Before, `tsconfig.json:compilerOptions.plugins[]` would:**
- reject `package-plugin/sub/path`
- reject `@package-plugin/subpath`
- load `..` and `.`

**After, `tsconfig.json:compilerOptions.plugins[]` will:**
- load `package-plugin/sub/path`
- load `@package-plugin/subpath`
- reject `@package-plugin/subpath/../../malicious`
- reject `plugin/subpath/../../malicious`
- reject `..` and `.`

I noticed while looking at the tests that `.` and `..` are accepted plugin names, which is likely unintentional.


src/compiler/moduleNameResolver.ts
```diff
@@ -2531,6 +2532,9 @@ function extensionIsOk(extensions: Extensions, extension: string): boolean {
 
 /** @internal */
 export function parsePackageName(moduleName: string): { packageName: string; rest: string; } {
+    if (pathIsRelative(moduleName) || pathIsAbsolute(moduleName)) {
+        return { packageName: "", rest: moduleName };
+    }
```
sheetalkamat, line 2537:
```suggestion
```

src/server/editorServices.ts
```diff
@@ -4487,8 +4488,18 @@ export class ProjectService {
...
+
+        const parsedPackageName = parsePackageName(pluginConfigEntry.name);
+        if (
+            !parsedPackageName.packageName ||
+            normalizePath(parsedPackageName.rest) !== parsedPackageName.rest ||
+            ensurePathIsNonModuleName(parsedPackageName.rest) === parsedPackageName.rest
+        ) {
+            this.logger.info(`Skipped loading plugin ${pluginConfigEntry.name} because only package name is allowed plugin name`);
             return;
         }
```
sheetalkamat, line 4504: If we are allowing subpaths in package name, this is better test i think

```suggestion
        if (
            !pluginConfigEntry.name ||
            isExternalModuleNameRelative(pluginConfigEntry.name) ||
            /[\\/]\.\.?($|[\\/])/.test(pluginConfigEntry.name)
        ) {
            this.logger.info(`Skipped loading plugin ${pluginConfigEntry.name || JSON.stringify(pluginConfigEntry)} because only package name is allowed plugin name`);
            return;
        }
```

sheetalkamat (changes requested):

src/server/editorServices.ts
sheetalkamat: @RyanCavanaugh @DanielRosenwasser do we need to ensure that this is "extensionless" as well?

kitten: lgtm 🙌 I'll get this applied tomorrow. I didn't have a good overview of all the utilities available in the repo and wasn't sure whether a custom additional regex was acceptable, so apologies for the rough attempt 😅

re. extensions; personally, I'd disallow them to limit requires/imports to qualified `package.json:main`, `exports` entries, and `index` files only. Just to be on the safe side btw, but that might be a little controversial. Not sure

kitten:
@sheetalkamat That's all applied and test cases are looking good and pass 👍 Sorry for the delay!

sheetalkamat (approved):

PR #52897, opened 2023-02-21
Ensure that Comparers only return -1, 0, 1 so results can be equated

jakebailey:
I noticed this while working on cleaning up our (inconsistent) use of the binary search and ordered insert APIs.

The binary search API requires that you have comparisons of _exactly_ `-1 | 0 | 1`, however, we had some comparers which didn't do this thanks to a hole in our type system ([Playground Link](https://www.typescriptlang.org/play?ts=4.9.5#code/KYOwrgtgBAwg9hADgQwE4EsDOcRQN4BQUxUAMsJpgCoAWyuJAvFALQCMANESQKICOYZABsqcEsWYAGLuIDiqYMgAuwVLXpRmbAgF8CBJQE9EwWAhQLUAHioA+TVAAUyAFxQqHKACM3VAJSa9vBIaFg4ANz6ACbAAMZCaKYAZmAgsUroOFCx5ok2ts6+nj7unjkhlm7BFqr5fm5ecHBCiiCR+uU1jpxQbJL9ns7FAYxBuRjYIAB0-IIicH7tBAD0y+40FKaJUKmYyEnAAIQEnYndnn0DTsjDgVDIrN6LJ+PA57390te3o1AATM99EYTGYKqo-vkHIVSt5fCN7OwoAAfKCSZG9doxeLbFJpDJZU4KCF2aEeWEwwmqKqvVDE2z1bxNFr0dqUv7vS5fIZPO7VUKTGYCYSiQGrdZYKASgDmTSixzZHM+gxuPN+DxYT0iCp6nOVP3sAPCQA)).

Make it so that we consistently return only one of the three valid values, so comparers can always be used in all APIs.

I also noticed (but did not change) the fact that we use binary searches for things like import insertion with the assumption that the input is always sorted, but, I don't think this actually turns out to be the case all the time anymore?

PR #57267, opened 2024-02-01
Change detection to detection by diff in organizeImports 

### LINKED ISSUE #57030: organizeImports needs to be called twice to be stable
### 🔎 Search Terms

organizeImports

- I was unable to test this on prior versions because this seems to be a tooling issue.

### ⏯ Playground Link

_No response_

### 💻 Code

```ts
import {abc, Abc} from 'b';
import {
  I,
  R,
  M,
} from 'a';
```

### 🙁 Actual behavior

This needs to be organized twice in order to be stable. The first time we get

import {I, M, R} from 'a';
import {abc, Abc} from 'b';

and the second time we get

import {I, M, R} from 'a';
import {Abc, abc} from 'b';

### 🙂 Expected behavior

I'd expect

import {I, M, R} from 'a';
import {Abc, abc} from 'b';

the first time around

_No response_

iisaduan:
This PR changes the detection method in OrganizeImports to make a decision off of the sort order that will make the least changes to the file. 

If there is a tie or not enough information to make a decision, then we default to case insensitive, and for type order, `last` followed by `inline`. These defaults should line up with our previous behavior, though with this new detection method, we should be resorting to the default in fewer cases.

The new detection has also been implemented in the codefixes that use any functions that organizeImports was previously supplying.

Some other features of this PR:
- Adds type-order detection.
- Fixes [#57030](https://github.com/microsoft/TypeScript/issues/57030), so that outputs of running Organize Imports will be stable.
- Removes need to resolve SortKind. 
- Runs in about the same time as `organizeImports` did with the previous detection method. Testing on a codespace, detecting on a version of `checker.ts` that is sorted and has some imports changed to type-only, tsserver logs report that organize imports takes:
  - On nightly 20240220, runs in ~3100ms. This PR runs in ~2700ms after this [commit](https://github.com/microsoft/TypeScript/pull/57267/commits/e3ccaf166adfcffc5fe55a1815343b442377cdd7) (that changes detection method to "is-sorted").
   - On nightly 20240221, runs in ~2000ms. This PR runs in ~1900ms after this [commit](https://github.com/microsoft/TypeScript/pull/57267/commits/c5239e75b03263019707b25b1d1fa3e22d38c427).
  - (On a not-already-sorted version, runtime increases all around, due to the amount of time needed to write changes)
- Cleans up the `organizeImports.ts` file by removing unused functions, and grouping functions (especially the ones that are only used for testing)


src/compiler/core.ts
```diff
@@ -837,6 +837,46 @@ export function sortAndDeduplicate<T>(array: readonly T[], comparer?: Comparer<T
...
+ */
+export function mergeAndDeduplicateSorted<T>(array1: readonly T[], array2: readonly T[], comparer: Comparer<T>): SortedReadonlyArray<T> {
+    if (array1.length === 0) return array2 as any as SortedReadonlyArray<T>;
+    if (array2.length === 0) return array1 as any as SortedReadonlyArray<T>;
+
+    const finalArray: T[] = [];
+    let i = 0, j = 0;
+    while (i < array1.length && j < array2.length) {
+        switch (comparer(array1[i], array2[j])) {
+            case Comparison.LessThan:
```
jakebailey, line 852: This isn't something new, but I am somewhat nervous about switching on these values exactly; our Comparison type is unsound and we have cases where we actually return values that are not these three, but instead just positive/negative values. See also #52897.

Maybe it doesn't happen with what is passed to this new function (and the assert helps), of course.

iisaduan: I actually encountered that bug already while working on this function (fixed by using a better `comparer`), so I put the default option there to prevent any loose ends from happening with unsound Comparisons.

iisaduan:
@typescript-bot pack this

src/services/utilities.ts
```diff
@@ -4276,6 +4275,39 @@ export function fileShouldUseJavaScriptRequire(file: SourceFile | string, progra
     return preferRequire;
 }
 
+/**
+ * Does not perform change matching, only returns a the number of the minimum edit
+ * @internal
+ */
+export function getDiffNum<T>(s1: T[], s2: T[]): number {
```
iisaduan, line 4282: Should I keep this?

src/services/codefixes/importFixes.ts
```diff
@@ -123,7 +127,7 @@ import {
     skipAlias,
     some,
     sort,
-    SortKind,
+    // SortKind,
```
jakebailey, line 130: This looks leftover.

src/services/codefixes/importFixes.ts
```diff
@@ -1520,23 +1536,20 @@ function doAddExistingFix(
...
         );
 
         // The sorting preference computed earlier may or may not have validated that these particular
         // import specifiers are sorted. If they aren't, `getImportSpecifierInsertionIndex` will return
         // nonsense. So if there are existing specifiers, even if we know the sorting preference, we
         // need to ensure that the existing specifiers are sorted according to the preference in order
         // to do a sorted insertion.
-        const specifierSort = existingSpecifiers?.length && OrganizeImports.detectImportSpecifierSorting(existingSpecifiers, preferences);
-        if (specifierSort && !(ignoreCaseForSorting && specifierSort === SortKind.CaseSensitive)) {
+        // const specifierSort = existingSpecifiers?.length && OrganizeImports.detectImportSpecifierSorting(existingSpecifiers, preferences);
```
jakebailey, line 1547: This looks leftover too.

src/services/organizeImports.ts
```diff
@@ -88,44 +81,67 @@ export function organizeImports(
...
+    const shouldCombine = shouldSort;
     const shouldRemove = mode === OrganizeImportsMode.RemoveUnused || mode === OrganizeImportsMode.All;
+
     // All of the old ImportDeclarations in the file, in syntactic order.
-    const topLevelImportGroupDecls = groupByNewlineContiguous(sourceFile, sourceFile.statements.filter(isImportDeclaration));
+    const topLevelImportDecls = sourceFile.statements.filter(isImportDeclaration);
+    const topLevelImportGroupDecls = groupByNewlineContiguous(sourceFile, topLevelImportDecls);
 
-    const comparer = getOrganizeImportsComparerWithDetection(preferences, shouldSort ? () => detectSortingWorker(topLevelImportGroupDecls, preferences) === SortKind.CaseInsensitive : undefined);
+    const DefaultComparer = getOrganizeImportsComparer(preferences, /*ignoreCase*/ typeof preferences.organizeImportsIgnoreCase === "boolean" ? preferences.organizeImportsIgnoreCase : true);
```
jakebailey, line 93: Nit, but this isn't a type or a namespace, so I wouldn't name this in PascalCase.

src/services/organizeImports.ts
```diff
@@ -927,3 +829,114 @@ function getTopLevelExportGroups(sourceFile: SourceFile) {
...
+    const moduleSpecifiersByGroup: string[][] = [];
+    importDeclsByGroup.forEach(importGroup => {
+        // Turns importDeclsByGroup into string[][] of module specifiers by group to detect sorting on module specifiers
+        moduleSpecifiersByGroup.push(getModuleNamesFromDecls(importGroup));
+    });
+    return detectCaseSensitivityBySort(moduleSpecifiersByGroup, comparersToTest);
+}
+
+/** @internal */
+export type TypeOrder = "first" | "last" | "inline";
```
jakebailey, line 844: I would think that this is actually a part of the services types / protocol somewhere; should this be moved?

jakebailey: Actually, I'm not seeing any changes which affect the protocol or service types, which I expected given we're adding a new preference (like Ron's past PR touching this).

src/services/organizeImports.ts
```diff
@@ -927,3 +829,114 @@ function getTopLevelExportGroups(sourceFile: SourceFile) {
...
+
+/** @internal */
+export function detectNamedImportOrganizationBySort(originalGroups: readonly ImportDeclaration[], comparersToTest: Comparer<string>[], typesToTest: TypeOrder[]): { namedImportComparer: Comparer<string>; typeOrder?: "first" | "last" | "inline"; isSorted: boolean; } | undefined {
```
jakebailey, line 847: This union is repeated from above.

src/services/utilities.ts
jakebailey: Is it used? If not, I would remove it.

tests/cases/fourslash/autoImportSortCaseSensitivity1.ts
```diff
@@ -28,7 +28,7 @@
 //// d/*1*/
 
 goTo.marker("0");
-verify.importFixAtPosition([`import { A, B, C, a } from "./exports1";\na`]);
+verify.importFixAtPosition([`import { a, A, B, C } from "./exports1";\na`]);
```
jakebailey, line 31: I assume since this is ambiguous, it doesn't really matter what we choose here?

tests/cases/fourslash/autoImportTypeImport4.ts
```diff
@@ -23,27 +23,28 @@
...
-//// d/*2*/
+//// d/*2*//*2a*//*2b*//*2c*/
 
 // addition of correctly sorted type imports should not affect behavior as shown in autoImportSortCaseSensitivity1.ts
 goTo.marker("0");
 verify.importFixAtPosition([
-    `import { A, B, C, a } from "./exports1";\na;\nb;`,
-    `import { A, B, C, b } from "./exports1";\na;\nb;`,
+    `import { a, A, B, C } from "./exports1";\na;\nb;`,
+    `import { A, b, B, C } from "./exports1";\na;\nb;`,
```
jakebailey, line 43: Hm, a bunch of these default orderings are changing; is this intended, in that we are now determining this to be case insensitively sorted? I'm half worried we'll break someone's use of our organizer as the default has changed for this specific example.

tests/cases/fourslash/importNameCodeFix_importType7.ts
```diff
@@ -32,21 +34,22 @@ new SomePig`],
...
-
-verify.importFixAtPosition([
-`import {
-    type SomeInterface,
-    SomePig,
-} from "./exports.js";
-new SomePig`],
-/*errorCode*/ undefined,
-{ organizeImportsTypeOrder: "first" });
+//TODO: bugged cases
```
jakebailey, line 37: Is this resolved?

tests/cases/fourslash/organizeImportsType10.ts
```diff
@@ -0,0 +1,50 @@
...
+////    func7,
+////    func8,
+////    func9,
+////} from "foo";
+////interface Use extends Type1, Type2, Type3, Type4, Type5, Type6, Type7, Type8, Type9 {}
+////console.log(func1, func2, func3, func4, func5, func6, func7, func8, func9);
+
+verify.organizeImports(
+`import {
+    type Type1,
```
jakebailey, line 28: This is a funky one; is it the case that we're detecting that putting them first has the fewest edits?

src/services/organizeImports.ts
jakebailey: No, I'm wrong; the ordering was added in a previous commit. But this duplicate declaration does feel suspect.

tests/cases/fourslash/organizeImportsType8.ts
```diff
@@ -20,33 +20,31 @@ console.log(a, b, A, B);`,
 
 edit.replaceLine(0, 'import { type A, type a, b, B } from "foo2";');
 verify.organizeImports(
-`import { B, b, type A, type a } from "foo2";
+`import { b, B, type A, type a } from "foo2";
```
jakebailey, line 23: I was going to ask how this was determining `b, B` instead of `B, b`, but I guess it's a 50/50 split between the two given it's `A, a, b, B`?

jakebailey (commented):
I think the only concern I have left is just how we're determining the "default" sorting (no options, what most people see). There are tools which wrap our organizer which will change behavior, and I'm not totally sure if that's good or bad...

src/services/organizeImports.ts
iisaduan: I wasn't sure what the convention was. All caps?

src/services/organizeImports.ts
iisaduan: It does sound better that way. I hadn't named the type in the previous commit

src/services/organizeImports.ts
jakebailey: I'd just call it `defautlComparer`.

tests/cases/fourslash/organizeImportsType8.ts
iisaduan: The original ordering is `b, B`, and it should preserve that since it's still ordered case-insensitively

tests/cases/fourslash/organizeImportsType10.ts
iisaduan: Yup, on both (funky and few edits)

src/services/organizeImports.ts
jakebailey: `defaultComparer`. I'm great at spelling.

tests/cases/fourslash/autoImportSortCaseSensitivity1.ts
iisaduan: For this particular test, since `ignoreCase = true`, `a` and `A` should be together, the test was wrong before 🤔

tests/cases/fourslash/autoImportTypeImport4.ts
iisaduan: There used to be an issue in the way that detection would default to case-sensitive very often (#57030 for example). Also, the codefixes may have been incorrectly resolving the sort from `SortKind` returned by the original detection function, so we're seeing a lot of behavior changes since we're no longer using the previous detection function, and thus no longer needing to resolve `SortKind`

tests/baselines/reference/api/typescript.d.ts
```diff
@@ -2939,7 +2939,7 @@ declare namespace ts {
                  *
                  * Default: `last`
                  */
-                readonly organizeImportsTypeOrder?: "last" | "first" | "inline";
+                readonly organizeImportsTypeOrder?: ts.OrganizeImportsTypeOrder;
```
jakebailey, line 2942: Can this be pulled in without .ts?

tests/cases/fourslash/importNameCodeFix_importType7.ts
iisaduan: Now it is!

tests/cases/fourslash/autoImportSortCaseSensitivity1.ts
jakebailey: Oh, didn't notice that ignore case was on!

tests/cases/fourslash/autoImportTypeImport4.ts
```diff
@@ -23,27 +23,28 @@
...
 ],
     /*errorCode*/ undefined,
     { organizeImportsTypeOrder : "last" });
+goTo.marker("0a");
```
jakebailey, line 47: What is with the extra navigation here? Aren't these all at the same positions?

jakebailey:
I feel like this is looking good, though my only overarching concern is the general feeling that we're going to see a lot of change in the way people's organize imports behave if it really is the case that we're going to start preferring insensitive more than we used to. Unless I'm still totally brainfarting and that's what we were doing before.

src/services/codefixes/importFixes.ts
```diff
@@ -1400,11 +1404,21 @@ function promoteFromTypeOnly(
...
+                    if (typeof preferences.organizeImportsIgnoreCase !== "boolean" || !preferences.organizeImportsTypeOrder) {
+                        const importDecl: ImportDeclaration[] = [aliasDeclaration.parent.parent.parent];
+                        const namedImportSort = OrganizeImports.detectNamedImportOrganizationBySort(importDecl, comparersToTest, typeOrdersToTest);
+                        if (namedImportSort) {
+                            const { namedImportComparer, typeOrder } = namedImportSort;
+                            specifierComparer = OrganizeImports.getOrganizeImportsSpecifierComparer({ organizeImportsTypeOrder: typeOrder }, namedImportComparer);
+                        }
+                    }
+
+                    const insertionIndex = OrganizeImports.getImportSpecifierInsertionIndex(aliasDeclaration.parent.elements, newSpecifier, specifierComparer);
```
andrewbranch, line 1421: Would it make sense to wrap some of this up into a higher-level helper function in organizeImports.ts? It feels like the ideal case is for auto-imports to just say “I want to insert this specifier into this import; just tell me where it should go”

tests/cases/fourslash/autoImportTypeImport4.ts
iisaduan: Sorry, just saw this comment. It helps to see which of the tests fail, as opposed to just reading "test failed at marker 0" for all of the tests. I can change it back if it's going to make the test much slower?

jakebailey: No, that totally makes sense, I just haven't seen any examples of this.

Not for this PR, but this makes me wonder if we should have the fourslash harness keep track of how many times it runs each function to better indicate what it's doing, or include the line number, etc. I think some of that already happens in stack traces?

iisaduan:
@typescript-bot pack this

tests/cases/fourslash/importNameCodeFixConvertTypeOnly1.ts
```diff
@@ -9,5 +9,5 @@
 //// new B
 
 goTo.file('/b.ts');
-verify.importFixAtPosition([`import { B, type A } from './a';
+verify.importFixAtPosition([`import { type A, B } from './a';
```
jakebailey, line 12: This file doesn't have any preferences set, but we're now switching from auto inserting the type import at the end to inlining it; is that desirable / by design?

I of course prefer inline sorting as that's what other tools do, but I'm just thinking about if this will be a surprise for someone downstream like vscode.

jakebailey: I'm just half confused because `tests/cases/fourslash/importNameCodeFix_importType7.ts` appears to not detect any order at all but this example doesn't seem much different.

jakebailey:
I don't want to let this one languish; besides how the defaults may have changed, is there anything else you need feedback on? The behavior does seem fine.

tests/cases/fourslash/importNameCodeFixConvertTypeOnly1.ts
iisaduan: `importFixes` is incorrectly comparing named imports when doing type promotions, I think that other case may be correct only by chance.

iisaduan: these changes have been reverted and importfixes has been corrected!

jakebailey (approved):
LGTM I think, but would be cool if @andrewbranch looked (I think he just went out of town, oops)

PR #57281, opened 2024-02-03
Fix bug#56997 - Parenthesized SatisfiesExpressions with comments are not unwrapped consistently in emitted JS

### LINKED ISSUE #56997: Parenthesized `SatisfiesExpression`s with comments are not unwrapped consistently in emitted JS
### 🔎 Search Terms

parenthesis satisfies comment removeComments emit

### ⏯ Playground Link

https://www.typescriptlang.org/play?#code/PTAEAECcFMFsHsBu0DC9a2gOwC4GcAuUAMwEMAbPaAKAGN4s8dRTQBeUYAKnoy7ACMABhZ5QWAK6wARtEh0GTUNPagAFN16x+oYaPFTZkAJQLGzWqrUae6bYJGkxkmXOOmzSgCarNdnXp4pDgAlnjEIdDOhnKezNBWfnwOoEGh4ZHRriZxJFY2WgEiaWERUQbZ7kA

### 💻 Code

```ts
// @removeComments: false
const a = /*comm*/ 10 as number
const b = (/*comm*/ 10 as number)
const c = ((/*comm*/ 10 as number))

const d = /*comm*/ 10 satisfies number
const e = (/*comm*/ 10 satisfies number)
const f = ((/*comm*/ 10 satisfies number))
```

### 🙁 Actual behavior

Emitted `.js` file:
```ts
"use strict";
// @removeComments: false
const a = /*comm*/ 10;
const b = /*comm*/ 10;
const c = /*comm*/ 10;
const d = /*comm*/ 10;
const e = ( /*comm*/10);
const f = (( /*comm*/10));
```

### 🙂 Expected behavior

```ts
"use strict";
// @removeComments: false
const a = /*comm*/ 10;
const b = /*comm*/ 10;
const c = /*comm*/ 10;
const d = /*comm*/ 10;
const e = /*comm*/ 10;
const f = /*comm*/ 10;
```

Those are not handled by `visitParenthesizedExpression`
https://github.com/microsoft/TypeScript/blob/4557e34e7003f5dd11fd33881ae91d10451fb23a/src/compiler/transformers/ts.ts#L1641
since they are (correctly) not considered to be assertions. That skips over the `createPartiallyEmittedExpression` call that, from what I understand, would help it to be handled correctly (consistently with assertions). I'm not sure but fixing this might require some changes/additions to `OuterExpressionKinds`

jean-michelet:
Fixes #56997 

Hi, this is my first time contributing here 👋

I don't know if I've done anything relevant, but I will follow your feedback carefully.


jean-michelet:

src/compiler/factory/utilities.ts
```diff
@@ -646,10 +646,11 @@ export function isOuterExpression(node: Node, kinds = OuterExpressionKinds.All):
         case SyntaxKind.TypeAssertionExpression:
         case SyntaxKind.AsExpression:
         case SyntaxKind.ExpressionWithTypeArguments:
-        case SyntaxKind.SatisfiesExpression:
```
Andarist, line 649: So it seems like `satisfies` was treated as a type assertion here. Would all tests pass if you'd only adjust the condition in `visitParenthesizedExpression` (from `isAssertionExpression(innerExpression)` to `isAssertionExpression(innerExpression) || isSatisfiesExpression(innerExpression)`)?

If I read correctly, the current change introduces a behavior change in `checkReferenceExpression`. Could you recheck what should happen there and add the appropriate test case?

tests/cases/compiler/parenthesizedSatisfiesExpressionWithComments.ts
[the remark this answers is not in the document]
jean-michelet: Something like that?
```ts
// @strict: true
// @lib: esnext
// @target: esnext
// @removeComments: false

export class Foo {
    client = {
      getThing: () => Promise.resolve('')
    }
  
    foo(): Promise<void> {
      return (((
        /* TODO: Avoid using type assertions, please refactor. */ this.client
          .getThing() as unknown as Promise<void> 
      ) satisfies Promise<void>));
    }
  }
  ```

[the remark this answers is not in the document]
Comment (Andarist) [reply tests/cases/compiler/parenthesizedSatisfiesExpressionWithComments.ts:1]: I'd shorten this to:
```ts
// @strict: true
// @lib: esnext
// @target: esnext
// @removeComments: false

export class Foo {
  client = {
    getThing: () => Promise.resolve('')
  }

  foo() {
    return (
      /* TODO: please refactor */ this.client
        .getThing() satisfies Promise<string>);
  }
}
```

src/compiler/factory/utilities.ts
jean-michelet: Oops, I missed that...

Are you aware of an existing test that you could update or take inspiration from?

I am not sure to understand what `checkAssignmentOperatorWorker` in `checkAssignmentOperator` do, but I can imagine these kind of test?
```ts
let a = 10;
a = (a satisfies number);
a = -(a satisfies number);
a = +(a satisfies number);
a = (a satisfies number)++;
a = (a satisfies number)--;

const arr = [10]
const [b = (arr satisfies number[])[0]] = arr

const obj = { c: 10 }
const { c = (obj satisfies { c: number }).c } = obj

for (let x of (arr satisfies number[])) { x++ }

(async function () {
  for await (const num of (arr satisfies number[])) {
    console.log(num)
  }
})();

for (const property in (obj satisfies { c: number })) { 
    console.log(property)
}

// checkAssignmentOperatorWorker ?
```

src/compiler/checker.ts
```diff
@@ -37684,7 +37684,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
 
     function checkReferenceExpression(expr: Expression, invalidReferenceMessage: DiagnosticMessage, invalidOptionalChainMessage: DiagnosticMessage): boolean {
         // References are combinations of identifiers, parentheses, and property accesses.
-        const node = skipOuterExpressions(expr, OuterExpressionKinds.Assertions | OuterExpressionKinds.Parentheses);
+        const node = skipOuterExpressions(expr, OuterExpressionKinds.Assertions | OuterExpressionKinds.SatisfiesExpressions | OuterExpressionKinds.Parentheses);
```
jean-michelet, line 37687: @Andarist 

I added `OuterExpressionKinds.SatisfiesExpressions` to the skipped outer expressions.

I tracked the calls to `checkReferenceExpression` and found that several uses cases had been silently broken.
```ts
let a = 10;

// called in checkPrefixUnaryExpression
--(a satisfies number);
++(a satisfies number);

// called in checkPostfixUnaryExpression
(a satisfies number)++;
(a satisfies number)--;

// called in checkAssignmentOperatorWorker
let b: number;
(b satisfies number) = 10;

// called in checkReferenceAssignment called in checkDestructuringAssignment
let c: number;
[(c satisfies number)] = [10];

let d: number, e = 1;
({ d: (e satisfies number) } = { d: 10 });

// called in checkForInStatement
let x: string = "hello"
for ((x satisfies string) in { a: 10 }) {
  console.log(x)
}

// called in checkForOfStatement
let g = 1
for ((g satisfies number) of [10]) {
  console.log(g)
}
```

Adding `OuterExpressionKinds.SatisfiesExpressions` corrects the error, and I've added the tests to prove it.

tests/cases/compiler/referenceSatisfiesExpression.ts
```diff
@@ -0,0 +1,33 @@
+// checkReferenceExpression -> skip OuterExpressionKinds.SatisfiesExpressions
```
Andarist, line 1: I'd probably drop those comments here (in this whole file) - they refer to implementations details that might change in the future

src/compiler/transformers/ts.ts
```diff
@@ -1696,8 +1697,9 @@ export function transformTypeScript(context: TransformationContext) {
     }
 
     function visitParenthesizedExpression(node: ParenthesizedExpression): Expression {
-        const innerExpression = skipOuterExpressions(node.expression, ~OuterExpressionKinds.Assertions);
-        if (isAssertionExpression(innerExpression)) {
+        const kinds = OuterExpressionKinds.Assertions | OuterExpressionKinds.SatisfiesExpressions;
+        const innerExpression = skipOuterExpressions(node.expression, ~kinds);
```
Andarist, line 1701: nit: I think this would be slightly more in line with the overall style in the codebase
```suggestion
        const innerExpression = skipOuterExpressions(node.expression, ~(OuterExpressionKinds.Assertions | OuterExpressionKinds.SatisfiesExpressions));
```

src/compiler/checker.ts
Andarist: Great and thorough investigative work 👏

src/compiler/types.ts
```diff
@@ -8235,12 +8235,13 @@ export const enum OuterExpressionKinds {
     Parentheses = 1 << 0,
     TypeAssertions = 1 << 1,
     NonNullAssertions = 1 << 2,
-    PartiallyEmittedExpressions = 1 << 3,
+    SatisfiesExpressions = 1 << 3,
```
Andarist, line 8238: a new kind here makes sense to me 👍 it's clearer this way because `satisfies` isn't usually called an assertion

src/compiler/types.ts
```diff
@@ -8235,12 +8235,13 @@ export const enum OuterExpressionKinds {
...
-    PartiallyEmittedExpressions = 1 << 3,
+    SatisfiesExpressions = 1 << 3,
+    PartiallyEmittedExpressions = 1 << 4,
 
     Assertions = TypeAssertions | NonNullAssertions,
-    All = Parentheses | Assertions | PartiallyEmittedExpressions,
+    All = Parentheses | Assertions | SatisfiesExpressions | PartiallyEmittedExpressions,
 
-    ExcludeJSDocTypeAssertion = 1 << 4,
+    ExcludeJSDocTypeAssertion = 1 << 5,
```
jakebailey, line 8244: That we keep changing this makes me wonder if we should just move this (and `PartiallyEmittedExpressions`) up to the end of the bits or something.

src/compiler/types.ts
jakebailey: Yeah though I'm not sure why we're pushing `PartiallyEmittedExpressions` up here. I guess it's consistent with previous values but it doesn't seem strictly required.

src/compiler/types.ts
jakebailey: I think I was suggesting like:

```ts
export const enum OuterExpressionKinds {
    Parentheses = 1 << 0,
    TypeAssertions = 1 << 1,
    NonNullAssertions = 1 << 2,
    SatisfiesExpressions = 1 << 3,
    PartiallyEmittedExpressions = 1 << 4,

    Assertions = TypeAssertions | NonNullAssertions,
    All = Parentheses | Assertions | SatisfiesExpressions | PartiallyEmittedExpressions,
    ExcludeJSDocTypeAssertion = 1 << 29,
}
```

But, this isn't really a request for this to be changed. Just thinking aloud.

jean-michelet:
Thanks for the review @jakebailey @Andarist

I've edited what was requested, let me know if it looks good to you.

jakebailey:
@typescript-bot test top200
@typescript-bot run dt

jakebailey (approved):
I think this is okay, though I think I'd like a second opinion from @rbuckton.

rbuckton (changes requested):

jakebailey (approved):

rbuckton (approved):

DanielRosenwasser (approved):

PR #57291, opened 2024-02-04
feat(57240): Unimplemented abstract methods result in verbose, repetitive message

### LINKED ISSUE #57240: Unimplemented abstract methods result in verbose, repetitive message
### 🔎 Search Terms

abstract method verbose 18052

### ⏯ Playground Link

https://tsplay.dev/Wkz4lm

### 💻 Code

```ts
abstract class A {
    abstract m1() : number;
    abstract m2() : number;
}

abstract class B extends A {
    abstract bm1() : number;
    abstract bm2() : number;
}


// The following error:
/*
Non-abstract class 'C' does not implement all abstract members of 'B'(18052)
input.ts(9, 14): Non-abstract class 'C' does not implement inherited abstract member 'bm1' from class 'B'.
input.ts(10, 14): Non-abstract class 'C' does not implement inherited abstract member 'bm2' from class 'B'.
input.ts(3, 14): Non-abstract class 'C' does not implement inherited abstract member 'm1' from class 'B'.
input.ts(4, 14): Non-abstract class 'C' does not implement inherited abstract member 'm2' from class 'B'.


Would be better expressed as one of the following (either of which could optionally
include a reference to B and/or A's declaration locations):


OPTION A:
Non-abstract class 'C' does not implement inherited abstract members 'bm1', 'bm2', 'm1', 'm2' from class 'B'.

OPTION B:
Non-abstract class 'C' does not implement inherited abstract members 'bm1', 'bm2' from class 'B'.
Non-abstract class 'C' does not implement inherited abstract members 'm1', 'm2' from class 'A'.
*/
class C extends B {

}
```

### 🙁 Actual behavior

Long, redundant error message

### 🙂 Expected behavior

Concise, readable error message

The actual case in which I ran into this was significantly more verbose because the class had generic parameters whose constraints were expanded in the description.

This makes it very hard to actually see which properties are missing in a single location.

a-tarasyuk:
Fixes #57240


typeholes:
Would it be possible to reword the error message so that the list is at the end?
```
Non-abstract class 'C' does not implement inherited abstract members 'bm1', 'bm2', 'bm3', 'm1', 'm2', 'm3' from class 'B'.ts(2654)
```

Something along the lines of 
```json
    "Type '{0}' is missing the following properties from type '{1}': {2}": {
        "category": "Error",
        "code": 2739
    },
```

andrewbranch:
I like @typeholes suggestion of moving the name of the abstract class before the list of members.

a-tarasyuk:
@andrewbranch Oke, I've changed diagnostic messages...

src/compiler/diagnosticMessages.json
```diff
@@ -7973,10 +7965,6 @@
         "category": "Error",
         "code": 18051
     },
-    "Non-abstract class '{0}' does not implement all abstract members of '{1}'": {
```
andrewbranch, line 7976: Sorry, I wasn’t suggesting removing these—I think the “Non-abstract” language is important—I just wanted to avoid the error message structure @typeholes pointed out:

> ```
> Non-abstract class 'C' does not implement inherited abstract members 'bm1', 'bm2', 'bm3', 'm1', 'm2', 'm3' from class 'B'.ts(2654)
> ```

However, it looks like this message was never possible. I’m fine with the messages the way they were, or this one could be rewritten as

```
Non-abstract class '{0}' is missing implementations for the following members of '{1}: {2}.'
```

But I only suggested changes because I thought it was possible for the name of the abstract class to be buried behind a long list. Sorry for the confusion.

src/compiler/diagnosticMessages.json
```diff
@@ -2946,10 +2950,22 @@
         "category": "Error",
         "code": 2652
     },
-    "Non-abstract class expression does not implement inherited abstract member '{0}' from class '{1}'.": {
+    "Non-abstract class expression does not implement inherited abstract member {0} from class '{1}'.": {
```
andrewbranch, line 2953:
```suggestion
    "Non-abstract class expression does not implement inherited abstract member '{0}' from class '{1}'.": {
```

andrewbranch (approved):
This looks great, thanks!

PR #57293, opened 2024-02-04
Permit circular references in type arguments for object type aliases

### LINKED ISSUE #35017: Circular reference detection is not lazy enough
**TypeScript Version:**  3.8.0-dev.20191105

**Search Terms:**
[circular](https://github.com/microsoft/TypeScript/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+circular)

**Code**

```ts
type A<T> = { x: T }
type R = A<R> // FAIL

type B = { x: B } // OK
```

**Expected behavior:**
Both cases throw no error.

**Actual behavior:**
Type R is rejected due to a circular reference.

**Playground Link:**
[Link](http://www.typescriptlang.org/play/?ts=3.8.0-dev.20191105&ssl=1&ssc=1&pln=64&pc=3#code/C4TwDgpgBAggPAFQHxQLxQN5QB4C4oJQC+AUKJFAEpqxyUoD0DUAYjAJIAyJZ40AQjSx4ogolCZQA8gGkgA)

**Related Issues:** 
https://github.com/microsoft/TypeScript/pull/33050

### LINKED ISSUE #35164: Record type fail with circular reference error
**TypeScript Version:**  3.7.2 (and also nightly)

**Search Terms:** `circularly references itself`, `Record circular reference`

**Code**

```ts
type RecursiveTypeA = Record<number, RecursiveTypeA> // error: Type alias 'RecursiveTypeA' circularly references itself.

type RecursiveTypeB = RecursiveTypeB[] // works fine

type RecursiveTypeC = { [index: number]: RecursiveTypeC } // works fine
```

**Expected behavior:**
Record type should work like the third example that defines an object with number keys to the recursive value (the key type is irrelevant).

**Actual behavior:**
Only the Record type fail with circular reference

**Playground Link:** [link](http://www.typescriptlang.org/play/?ssl=5&ssc=58&pln=3&pc=1#code/C4TwDgpgBAShDGBXATgZwJYDcIBVwQEEoBeWBAe2QBMAeAO0QFsAjCZAGjKTS13wIB8AKCGhIXFBmx5IAIRISe0-LIDaAXRFjocblL6QAwgoDeUVejpUIADwBcUBizbqHuybxkRjAXyA)

**Related Issues:** couldn't find any specific related issues

### LINKED ISSUE #41164: Circular reference error when defining Record type
**TypeScript Version:**  4.0.2 (and 4.1.0-dev.20201015)

**Search Terms:** circularly, Record

**Code**

```ts
type T1 = { [key: string]: T1 }
type T2 = Record<string, T2>
```

**Expected behavior:**
No compilation errors

**Actual behavior:**
While type T1 compiles properly, type T2 reports `Type alias 'T2' circularly references itself.` even though they are inherently the same type.  This was actually found by autofixing a new @typescript-eslint rule [consistent-indexed-object-style](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/consistent-indexed-object-style.md)

**Playground Link:** https://www.typescriptlang.org/play?ts=4.0.2#code/C4TwDgpgBAKgjFAvFA3lA2gawiAXFAZ2ACcBLAOwHMBdfeKAXwFgAoUSWAJiSgCUIAxgHtiAEwA8RMlQA0XAHxA

**Related Issues:** None found

### LINKED ISSUE #57034: Cannot describe recursive type aliases like JSON in terms of generic mapped types
In the following examples, I've defined 4 different versions of a `Json` type, where they differ on their description of object types.

[Playground link here](https://www.typescriptlang.org/play?#code/PTAEDkHtQUwJzpOBnUhQcgLACgB2BDAWxmQAc8BjGUAeQCMArGcgFwElkBZPEkmAEwAqAT17UcAZWZwAljgDmoAN7ZQq2AA8SSZqGYiqAKWSQcoALwq1VgD6hkU2XMtXVtnAFcCteM5e3akJAANjB4OL42oB5BQRFqtkYmANoAunGuSqBJANKgsnYO8ikAXKCJpgC+6aAA3NhVWNggENDwiCjo2PhEpBRUdIws7Kw4fDDq4tJy+MzucFTKWFbjWnA6erxlxqYWSy4Z9jLy1W6e3nAnoAHBoeF7+6cxl+Wpl4pZANYwQqWHjiVbEygBr7VR1LANJpgACiCCQqEAMuRdQjEMiUGgMJhsZAAJSYSD4Sl8K20un0gJ2lz+x3ufiiZx8tMi1xCYUu0ViTPiFNeXIyePIBIAPNS5AAaCkAPmq4MhWGasPaoCRuBRvXRAyx7AA4jAcPBpOQuDx+MJNosrBtDNsAAqIXhraTEIW5cbMPV8VB2yAOvTZb6S8xEvlZG15UzZAHlXwVcHLTSkq0U8xUwpOEOnLyM0GgfyBVl3HOPTlFnlpDMU72+p3IIXlaUh2VAA).

In the first two, I've defined them in terms of mapped types with a concrete key of string, and an object type with a `string` index signature. TypeScript has no issues with either of those.

On the other hand, the last two are described with `Record`, or with a helper mapped type fed in a `string` key type.

```ts
// No errors ✅
namespace ObjectIsMappedTypeOnString {
    export type Json =
        | string
        | number
        | boolean
        | null
        | Json[]
        | { [K in string]: Json }
        ;
}

// No errors ✅
namespace ObjectIsIndexSignature {
    export type Json =
        | string
        | number
        | boolean
        | null
        | Json[]
        | { [key: string]: Json }
        ;
}

// Errors ❌
namespace ObjectIsRecord {
    export type Json =
        | string
        | number
        | boolean
        | null
        | Json[]
        | Record<string, Json>
        ;
}

// Error ❌
namespace ObjectIsGenericMappedType {
    type JsonProperties<K extends PropertyKey> = {
        [P in K]: Json
    };
    export type Json =
        | string
        | number
        | boolean
        | null
        | Json[]
        | JsonProperties<Json>
        ;
}
```

I would expect all of these to work.

ahejlsberg:
With this PR we lift a long standing restriction and permit type arguments for object type aliases to circularly reference themselves. For example:

```ts
type Rec = Record<string, Rec>;  // Ok, previously was error

type Json = string | number | boolean | null | Json[] | Record<string, Json>;  // Ok, previously was error

type Fun<T> = () => T;
type RecFun = Fun<RecFun>;  // Ok, previously was error

type Tup<A, B> = [A, B];
type RecTup<T> = Tup<T, RecTup<T>> | null;  // Ok, previously was error

```

The PR specifically covers circular references in type arguments for type aliases that reference the following kinds of types:

* References to generic classes
* References to generic interfaces
* Object type literals
* Function type literals
* Array type literals
* Tuple type literals

Circular references in type arguments for type aliases that reference other kinds of types still aren't permitted. For example:

```ts
type TNode<T> = { type: "node", left: T, right: T };
type TLeaf<V> = { type: "leaf", value: V };
type TTree<V> = TNode<TTree<V>> | TLeaf<V>;  // Ok

type TLevel<V, T> = TNode<T> | TLeaf<V>;  // Ok
type TTree2<V> = TLevel<V, TTree2<V>>;  // Error
```

Above, instantiations of `TNode<T>` and `TLeaf<V>` are permitted to have circular type arguments because they are type aliases for object types. However, `TLevel<V, T>` does not permit circular type arguments because it is a type alias for a union type.

Fixes #35017.
Fixes #35164.
Fixes #41164.
Fixes #57034.


ahejlsberg:
@typescript-bot test top200
@typescript-bot run dt

tests/baselines/reference/omitTypeTestErrors01.types
```diff
@@ -13,7 +13,7 @@ interface Foo {
 }
 
 export type Bar = Omit<Foo, "c">;
->Bar : { a: string; b: number; }
+>Bar : { b: number; a: string; }
```
DanielRosenwasser, line 16: Why'd this change?

ahejlsberg: It has to do with the order in which the literal types were materialized. Because we defer the `Omit` instantiations, we end up materializing the `"b"` and `"c"` literal types before we materialize `"a"`. Since `Omit` is non-homomorphic and since unions are ordered by type ID, this ends up flipping the properties.

DanielRosenwasser:
@typescript-bot pack this

DanielRosenwasser:
```ts
type Loopy1<T> = Loopy2<T>;
type Loopy2<T> = Record<string, keyof Loopy1<T>>;
```

Seems to cause a stack overflow in the playground.

jedwards1211:
@ahejlsberg can you clarify what you mean by this?

> However, TLevel<V, T> does not permit circular type arguments because it is a type alias for a union type.

Taking this at face value would mean `TTree<V>` could not be instantiated, because it's also a type alias for a union type.

Do you mean that you can't instantiate a type alias for a circular union type *if* one of the type parameters is itself a circular union type?

dgreensp:
Does this cover the basic case of aliasing a class?

```ts
class Box<T> {
  t!: T;
}

type MyBox<T> = Box<T>;

type Foo = number | Box<Foo>; // this is fine
type Bar = number | MyBox<Bar>; // error, circular reference
```

I'm hitting this problem in my code, and the alias is needed for other reasons, making this hard to work around.

What is the status of this PR?

ahejlsberg:
@dgreensp Yes, it covers that scenario. However, it's not clear the implementation approach in the PR is viable, so we're still thinking about solutions. I can say for sure it won't be in the 5.5 release.

AaronNGray:
@ahejlsberg Seems you dealt with mutually recursive types fine.

My cases from https://github.com/microsoft/TypeScript/issues/61062 pass fine !
```
type Any<T extends Object> = {
    [key:string]: T;
}; 

type Result = string | Result[] | Any<Result>;
```
```
type ResultObject = {
    [key:string]: Result;
}; 

type ResultArray<T> = T[];

type Result = string | ResultArray<Result> | ResultObject;
```
```
type Result = string | Array<Result> | Record<string, Result>;
```

AaronNGray:
@ahejlsberg Yes @DanielRosenwasser test case causes :- RangeError: Maximum call stack size exceeded

AaronNGray:
Disecting :-
```
type Loopy1<T> = Loopy2<T>;
type Loopy2<T> = Record<string, keyof Loopy1<T>>;  
```
we get the following :-
```
type Loopy1<T> = Loopy2<T>;
type KeyOfLoopy1<T> = keyof Loopy1<T>;   // error TS2456: Type alias 'KeyOfLoopy1' circularly references itself.
type Loopy2<T> = Record<string, KeyOfLoopy1<T>>;  // TS2315: Type 'KeyOfLoopy1' is not generic.
```
But not the :- RangeError: Maximum call stack size exceeded

so it looks like recursive parameterization over `keyof T` does not resolve, which probably means we just need to cover the `keyof<T>` case. @ahejlsberg I need to have a serious look at your code, it does not seem very othoganal to the existing code base but that suffers from non othoganality anyway.
Seeing as this is over a year old and AFAIK there is no other resolution, atleast no one else has offered any other forwards movement on this thread, is anyone else looking at this issue as its an incomplete type system ? @DanielRosenwasser @sandersn @weswigham @gabritto.

PR #57305, opened 2024-02-06
fix(57302): Cannot move symbols from tsx to ts files

### LINKED ISSUE #57302: Cannot move symbols from tsx to ts files
* have tsx file with some function
* select the function, invoke the "move to file" code action
* the picker doesn't allow for ts files, only tsx files

a-tarasyuk:
Fixes #57302


DanielRosenwasser (approved):

sheetalkamat (approved):

sandersn:
@DanielRosenwasser I think this makes sense for 5.4 but I want to double-check with you.

PR #57314, opened 2024-02-06
fix(57303): no space between new line and from keyword

### LINKED ISSUE #57303: Allman style in import syntax seems weird

- VS Code Version: `1.86.0`
- OS Version: `Linux Tastral 6.5.0-10022-tuxedo microsoft/vscode#26 SMP PREEMPT_DYNAMIC Thu Jan 18 02:29:42 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux` (ubuntu-based - TuxedoOS 2)

Steps to Reproduce:

1. Add an `import` line
2. Break `import` line into 2 lines from the inside of bracket
3. Format for first time, you'll see all brackets got new line and indented \
   image
4.  Format for second time, you'll see all brackets lose its indentation \
   image

I see 2 problems in here:

1. Format 2 times with 2 different results is no good
2. If the `from` keyword got new line, it always get intended. I don't think it make sense. Can we just have like this:

```ts
import
{
   ChallengeTier,
   ChallengeLevel, ChallengeData
}
from './challenge.ts'
```

This happens when I use TypeScript. I haven't tested it on JavaScript.

solasamuel:
Fixes #57303 
Changes syntax format from
```
import
{
   ChallengeTier,
   ChallengeLevel, ChallengeData
}
    from './challenge.ts'
```

to

```
import
{
   ChallengeTier,
   ChallengeLevel, ChallengeData
}
from './challenge.ts'
```

### DIFF
--- a/src/services/formatting/rules.ts
+++ b/src/services/formatting/rules.ts
@@ -235,6 +235,13 @@ export function getAllRules(): RuleSpec[] {
             [isNonJsxSameLineTokenContext],
             RuleAction.InsertSpace,
         ),
+        rule(
+            "NoSpaceBetweenFromKeywordAndLineStart",
+            SyntaxKind.NewLineTrivia,
+            SyntaxKind.FromKeyword,
+            [isNonJsxSameLineTokenContext],
+            RuleAction.DeleteSpace
+        ),
         // Treat string literals in module names as identifiers, and add a space between the literal and the opening Brace braces, e.g.: module "m2" {
         rule("SpaceAfterModuleName", SyntaxKind.StringLiteral, SyntaxKind.OpenBraceToken, [isModuleDeclContext], RuleAction.InsertSpace),
 


iisaduan (changes requested):
Thanks for the contribution! 

The formatting described in the bug is probably not the ideal for our formatter to output, as other formatting tools will re-format import statements differently. Can you also add the following test and change the rule so that the test passes?

```
///<reference path="fourslash.ts" />

//// import 
//// {
////    ChallengeTier,
////    ChallengeLevel, 
////    ChallengeData
//// }
//// from './challenge.ts'
//// import 
////    {
////        A,B,C
////    } 
////    from './abc.ts'

format.document();
verify.currentFileContentIs(
`import {
   ChallengeTier,
   ChallengeLevel,
   ChallengeData
} from './challenge.ts'
import {
   A, B, C
} from './abc.ts'`);
```

Issue #57317, opened 2024-02-07
Make `instanceof` narrow better using an anonymous class returned from a function


DavidArchibald:
### 🔎 Search Terms

Function return, anonymous class, instanceof, mixins, narrowing, 

### 🕗 Version & Regression Information

This is the behavior in every version I tried, and I reviewed the FAQ for entries about instanceof.

I used `every-ts` (kudos to Jake Bailey for making it so easy) so I'm pretty confident this is true unless some random commit fixed it temporarily.

The closest entry is this: https://github.com/Microsoft/TypeScript/wiki/FAQ#why-doesnt-x-instanceof-foo-narrow-x-to-foo

However as I understand it as long as `if (x instanceof y)` compiles, it should be narrowing `x` to basically `typeof x & InstanceType<typeof y>`. There might be some other edge cases I'm not aware of but as you'll see in the playground it seems like this isn't happening as precisely as desired with an anonymous class returned from a function.

### ⏯ Playground Link

https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABAWQIYGsCmBhANqgZwIB4AVAPgAoC4BbTAZShGGAC5FSBKRAbwFgAUIkQAnTM1FII+InyEiRNekxbBEAXkTLGzVgG4FiAL5DTgoRAQEoKVABMcsggDkQtAEaZRmu1jyEBJQAjABMAMxchoJWYDZ2jgFETKIwYADmvmj+zpQARMBwcHlRQpbWtmDuXqIAknFQqJCYvmCYAO4JToFunt6UpYJVfXUNTRCYAHQ6qgaIAPTziPXA3gSIhIgABsM1WxvrMLYEABZwILj2iF6TZTEV2nSY9TbjmBzg6GBw7UhaaFATpNRE17HQBohyIgAAyTACsiAA-Ig2p00IlnL0ahCOKiukkCCk0ukBtEYOpqE8Xo1mog0q9mnB1OjukQsd4eAJhI96NS3tMnrNgPoFksVmsDtsmgBPfaglFwWw7areLa3QTGIA

### 💻 Code

```ts
function MakeClass<T>(someStuff: T) {
  return class {
    someStuff = someStuff;
  }
}

const MadeClassNumber = MakeClass(123);
const MadeClassString = MakeClass("foo");

const numberInstance = new MadeClassNumber();
numberInstance.someStuff; // Infers as `number` as it should be.

[...]

PR #49863, opened 2022-07-11
Infer instance type arguments as constraint

molisani:
In JavaScript, there are several runtime checks that can determine if an object is an instance of a specific class. However, due to the fact that TypeScript type parameters only exist at the type level, there are limited options for how to handle these when building the class instance reference type. At the moment, there are two different approaches used by the two different kind of checks: one defaults to `any` and the other re-uses the type argument from the enclosing class. The first could be stricter and the second is possibly too strict in some cases.

The proposed change is to modify the way TypeScript handles these checks such that the type arguments are inferred as the constraint from the corresponding type parameter. This would default to `unknown` if no constraint was specified as that is the "default constraint".

### Fixes #17473 - Infer constrained generic parameters after instanceof check

For `instanceof` checks, TypeScript mirrors the underlying behavior of JavaScript and looks at the `Class.prototype` property. This property is computed dynamically in the checker and is then used to narrow the target of the `instanceof` expression. If the class in question has any type parameters, TypeScript returns a type reference where all of the type arguments are `any`. This is documented as being part of the "TypeScript 1.0 spec" so it is very long-standing behavior. The proposed change in behavior would constitute a breaking change as it would narrow some types from `any` to the constraint type or `unknown`.

```ts
class Box<T> { value: T }
Box.prototype; // changes from `Box<any>` to `Box<unknown>`
declare const box: any;
if (box instanceof Box) {
    box; // changes from `Box<any>` to `Box<unknown>`
}

class StringBox<T extends string> { value: T }
StringBox.prototype; // changes from `StringBox<any>` to `StringBox<string>`
declare const stringBox: any;
if (stringBox instanceof StringBox) {
    stringBox; // changes from `StringBox<any>` to `StringBox<string>`
}
```

### Fixes #46668 - Private field check narrows generic class too far

For `#brand in obj` checks, TypeScript can infer that if an object has a private identifier then it must be an instance of the enclosing class. If the class in question has any type parameters, TypeScript sets all of the type arguments of the inferred type to match the type arguments of `this` (see thread for potential issues). This behavior was only introduced recently via [ergonomic brand checks for private fields](https://github.com/microsoft/TypeScript/pull/44648). The proposed change in behavior is technically a breaking change, but less so than the `instanceof` check as in some cases it would loosen type arguments to match their constraint.

```ts
class PrivateBox<T> {
    #brand;
    value: T;
    copyValueFrom(other: object) {
        if (#brand in other) {
[...]

PR #57323, opened 2024-02-07
Replace only local type parameters with `any` when creating `prototype` property type of a class

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/57317
this is different from https://github.com/microsoft/TypeScript/pull/49863 - that PR likely still has a bug that I'm fixing here because it would replace the outer type parameter with its constraint

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -10701,7 +10701,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         // the type of which is an instantiation of the class type with type Any supplied as a type argument for each type parameter.
         // It is an error to explicitly declare a static property member with the name 'prototype'.
         const classType = getDeclaredTypeOfSymbol(getParentOfSymbol(prototype)!) as InterfaceType;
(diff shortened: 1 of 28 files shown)


src/compiler/checker.ts
```diff
@@ -10701,7 +10701,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         // the type of which is an instantiation of the class type with type Any supplied as a type argument for each type parameter.
         // It is an error to explicitly declare a static property member with the name 'prototype'.
         const classType = getDeclaredTypeOfSymbol(getParentOfSymbol(prototype)!) as InterfaceType;
-        return classType.typeParameters ? createTypeReference(classType as GenericType, map(classType.typeParameters, _ => anyType)) : classType;
+        return classType.localTypeParameters ? createTypeReference(classType as GenericType, concatenate(classType.outerTypeParameters, map(classType.localTypeParameters, _ => anyType))) : classType;
```
Andarist, line 10704: I wonder if this doesn't create any type parameter leak or something like this but so far I couldn't recognize any unintended side-effect like this here.

PR #57349, opened 2024-02-09
fix(55500): Incorrect error reported when using class extends null an…

### LINKED ISSUE #55500: Incorrect error reported when using `class extends null` and re-opening interface
### 🔎 Search Terms

class extends null interface

- This is the behavior in every version I tried (at least up until 3.3, which is the oldest build in the playground).

### ⏯ Playground Link

https://www.typescriptlang.org/play?lib=true&ts=5.1.6#code/JYOwLgpgTgZghgYwgAgEJwM4oN4CgC+uuCANphsgMLIQAekIAJhSAK4knJ7I-IID2IDGCisEYflAAUASi65eyQoVCRYiFNToNmaTDnxA

### 💻 Code

```ts
interface Base {
}

class C extends null {
    constructor() {
    }
}
interface C extends Base {}
```

### 🙁 Actual behavior

Reports the following error:

```
Class static side 'typeof C' incorrectly extends base class static side 'null'.
```

### 🙂 Expected behavior

No error. `inteface Base` does not describe the static side of anything, and reopening a class using something like `interface C extends Base {}` is common practice for shims.

I ran into this when implementing a class that utilizes `return` override in a constructor when combined with `class extends null` while experimenting with the Shared Structs origin trial (whose constructors produce objects with a `null` prototype). Though native `extends null` has some issues in ES2015+, it still works fine with explicit `return` override in a constructor in any engine:

```js
class C extends null {
  constructor() {
    return {};
  }
}
new C(); // works just fine.
```

Zzzen:
…d re-opening interface

Fixes #55500


tests/baselines/reference/classExtendsNull2.errors.txt
Andarist, line 1: to increase the coverage you could test this with strictNullChecks on and off (currently it's only tested with `strictNullChecks: true`)

PR #57358, opened 2024-02-09
Allow (non-assert) type predicates to narrow by discriminant

### LINKED ISSUE #45770: Narrowing discriminated union via type predicate and type assertion results in different behavior.
# Bug Report

### 🔎 Search Terms
type predicate discriminated union

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about v4.5.0-dev.20210907

### ⏯ Playground Link
[Playground link with relevant code](https://www.typescriptlang.org/play?ts=4.5.0-dev.20210907#code/CYUwxgNghgTiAEYD2A7AzgF3gMxgVwEsMAueAb3gGsCVhSByKAByYhHoF94AfcqmuvHoAjKCjFR68LrwrVaDMAAsQMGAE8pHAFDbQkWAmx4UYDAVTwCaAPIoQN7AB4AKgBp4AVXggAHhhBaNHgXAD4ACiIQAFtSd3hYGCh1UjgoYFQIdS8AbQBdAEpSKOirYM9tAmx4SNt7R3DcQgwAOnlgDxzGFjZ6DxEJcXo8hODkdAwCgvJteDmcfCI2gXgAelWhZlZ2HiFRcSHZ+aaiNY25AQYt3pk+doZ9iS1di4UhZVUNLW0uEAg0BBkI5zE6tdpnd4qNSaYELZoQ16CbrbTgvfhvAYHSTSNH3SGfTTSH66fTQOA4ExmCwoMoAERA2BoUSydgczni3j8ASCIQiJTiHkSyVSIHSmWynnyRVGAJgGGCJTKXkqaHpjJQzPUrIaoOWtE6yN6-UeQxGUDGqEwBW0uvB602PR2vExTxtiyw9sRV0dqNk6KRJuxOiAA)

### 💻 Code

```ts
declare const fruit: { kind: 'apple'} | { kind: 'banana' } | { kind: 'cherry' }

declare function isOneOf<T, U extends T>(item: T, array: readonly U[]): item is U
if (isOneOf(fruit.kind, ['apple', 'banana'] as const)) {
    fruit.kind // 'apple' | 'banana'
    fruit // { kind: 'apple'} | { kind: 'banana' } | { kind: 'cherry' }
} else {
    fruit.kind // 'cherry'
    fruit // { kind: 'apple'} | { kind: 'banana' } | { kind: 'cherry' }
}

declare function isDefinitelyOneOf<T, U extends T>(item: T, array: readonly U[]): asserts item is U
isDefinitelyOneOf(fruit.kind, ['apple', 'banana'] as const)
fruit.kind // 'apple' | 'banana'
fruit // { kind: 'apple'} | { kind: 'banana' }
```

### 🙁 Actual behavior
Type assertion does the correct thing, but the equivalent type predicate results in an odd scenario where the property is *correctly* narrowed, but the union object is not narrowed.

### 🙂 Expected behavior
Type assertion and type predicate have the same narrowing behavior, and only differ on whether the code branches or throws.

I would also expect the type predicate to either narrow neither the property nor the object or both the property and the object (like the type assertion does).

gabritto:
Fixes #45770.

In the example:

```ts
declare const fruit: { kind: 'apple'} | { kind: 'banana' } | { kind: 'cherry' }

declare function isOneOf<T, U extends T>(item: T, array: readonly U[]): item is U
if (isOneOf(fruit.kind, ['apple', 'banana'] as const)) {
    fruit.kind
    fruit
}
```

In `narrowTypeByCallExpression` when the reference to be narrowed is `fruit` and the call expression is `isOneOf(fruit.kind, ['apple', 'banana'] as const)`, we were simply not doing the narrowing because we though none of the arguments of the call matched the reference. Now, we see that argument `fruit.kind` matches `fruit`.

Note that cases like this one still won't be narrowed:
```ts
declare function isA(x: unknown): x is "a";
declare const what: { kind1: { kind2: "a" } } | { kind1: { kind2: "b" } };

if (isA(what.kind1.kind2)) {
    const a: "a" = what.kind1.kind2; // This works: `what.kind1` is narrowed
    const w: { kind1: { kind2: "a" } } = what; // This doesn't work: `what` is not narrowed
}
```
because when matching the reference `what` to the arguments of `isA(what.kind1.kind2)`, we only check one level of access expression, so we only check if `what` matches `what.kind1`, which it doesn't.
But this behavior is on par with how narrowing by discriminant property already works, including for the assert version of the example above (see, for instance, how `getCandidateDiscriminantPropertyAccess` is implemented).

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -26740,11 +26740,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     function hasMatchingArgument(expression: CallExpression | NewExpression, reference: Node) {
         if (expression.arguments) {
             for (const argument of expression.arguments) {
-                if (
-                    isOrContainsMatchingReference(reference, argument)
-                    || optionalChainContainsReference(argument, reference)
-                    || isAccessExpression(argument) && isMatchingReference(reference, argument.expression)
-                ) {
+                if (isOrContainsMatchingReference(reference, argument) || optionalChainContainsReference(argument, reference)) {
                     return true;
                 }
             }
@@ -28758,12 +28754,10 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         }
 
         function narrowTypeByCallExpression(type: Type, callExpression: CallExpression, assumeTrue: boolean): Type {
-            if (hasMatchingArgument(callExpression, reference)) {
-                const signature = assumeTrue || !isCallChain(callExpression) ? getEffectsSignature(callExpression) : undefined;
-                const predicate = signature && getTypePredicateOfSignature(signature);
-                if (predicate && (predicate.kind === TypePredicateKind.This || predicate.kind === TypePredicateKind.Identifier)) {
-                    return narrowTypeByTypePredicate(type, predicate, callExpression, assumeTrue);
-                }
+            const signature = assumeTrue || !isCallChain(callExpression) ? getEffectsSignature(callExpression) : undefined;
+            const predicate = signature && getTypePredicateOfSignature(signature);
+            if (predicate && (predicate.kind === TypePredicateKind.This || predicate.kind === TypePredicateKind.Identifier)) {
+                return narrowTypeByTypePredicate(type, predicate, callExpression, assumeTrue);
             }
             if (containsMissingType(type) && isAccessExpression(reference) && isPropertyAccessExpression(callExpression.expression)) {
                 const callAccess = callExpression.expression;
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -26740,7 +26740,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     function hasMatchingArgument(expression: CallExpression | NewExpression, reference: Node) {
         if (expression.arguments) {
             for (const argument of expression.arguments) {
-                if (isOrContainsMatchingReference(reference, argument) || optionalChainContainsReference(argument, reference)) {
+                if (isOrContainsMatchingReference(reference, argument)
+                    || optionalChainContainsReference(argument, reference)
+                    || isAccessExpression(argument) && isMatchingReference(reference, argument.expression)) {
                     return true;
                 }
             }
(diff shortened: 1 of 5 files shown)


gabritto:
@typescript-bot run DT
@typescript-bot test top200

src/compiler/checker.ts
```diff
@@ -26740,7 +26740,11 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     function hasMatchingArgument(expression: CallExpression | NewExpression, reference: Node) {
         if (expression.arguments) {
             for (const argument of expression.arguments) {
-                if (isOrContainsMatchingReference(reference, argument) || optionalChainContainsReference(argument, reference)) {
+                if (
+                    isOrContainsMatchingReference(reference, argument)
+                    || optionalChainContainsReference(argument, reference)
+                    || isAccessExpression(argument) && isMatchingReference(reference, argument.expression)
```
weswigham, line 26746: Wouldn't it be better to `getDiscriminantPropertyAccess(argument, type)` so the argument only matches if it has an actual discriminable property? AFAIK, `hasMatchingArgument` is only used in two places, so specializing it a bit more for this usecase is probably fine. Technically `getDiscriminantPropertyAccess` can also match binding patterns and identifiers aliasing discriminable properties, too, eg, 

```ts
const kind = fruid.kind;
if (isOneOf(kind, ['apple', 'banana'] as const)) {
    fruit.kind
    fruit
}
```

gabritto:
@typescript-bot pack this

src/compiler/checker.ts
gabritto: We'll end up calling `getDiscriminantPropertyAccess` twice, but I'll try doing that.

gabritto: You know what, at this rate, if we need to call `getDiscriminantPropertyAccess` to decide whether we should call `narrowTypeByTypePredicate`, which will then possibly call `getDiscriminantPropertyAccess` (and also call `isMatchingReference` anyways), then we might as well just always call `narrowTypeByTypePredicate`? I'll try that and see if it affects performance.

gabritto:
@typescript-bot run DT
@typescript-bot test top200

gabritto:
@typescript-bot pack this

gabritto:
@typescript-bot run DT
@typescript-bot test top200

gabritto:

src/compiler/checker.ts
gabritto: @weswigham I've switched to using `getCandidateDiscriminantPropertyAccess` for the above reasons.
Can you take a look again?

weswigham (approved):

PR #57361, opened 2024-02-09
Deduplicate protocol.ts content

### LINKED ISSUE #57232: Possibly missing properties in the `ts.server.protocol.CompilerOptions` interface
### 🔍 Search Terms

"server", "protocol", "CompilerOptions"

### ✅ Viability Checklist

### ⭐ Suggestion

Just wanted to draw your attention that the `ts.server.protocol.CompilerOptions` interface is missing recently added properties that are present in `ts.CompilerOptions`. Like `allowImportingTsExtensions`, `resolvePackageJsonExports`, `resolvePackageJsonImports`, `allowArbitraryExtensions `, etc. Probably there are more:

https://github.com/microsoft/TypeScript/blob/8c34b56cde00b5479fb2c85f94fa8be83e28b06a/src/server/protocol.ts#L3665-L3670
https://github.com/microsoft/TypeScript/blob/8c34b56cde00b5479fb2c85f94fa8be83e28b06a/src/server/protocol.ts#L3713-L3720

### 📃 Motivating Example

Similar to: https://github.com/microsoft/TypeScript/pull/56785#issuecomment-1900255297

@andrewbranch Could you take a look, please? Perhaps it is worth adding these too?

### 💻 Use Cases

Index signature covers them all. So this is not a problem. Only felt like those should be explicitly defined.

By the way, deprecated options like `suppressExcessPropertyErrors`, `suppressImplicitAnyIndexErrors` most probably have to be removed from `ts.server.protocol.CompilerOptions` in TypeScript 5.5.

https://github.com/microsoft/TypeScript/blob/8c34b56cde00b5479fb2c85f94fa8be83e28b06a/src/server/protocol.ts#L3724-L3729

andrewbranch:
Fixes #57232 and improves some quality of life, hopefully


src/server/protocol.ts
```diff
@@ -1,24 +1,41 @@
...
+    CompletionTriggerKind,
+    OrganizeImportsMode,
+    SemicolonPreference,
+} from "./_namespaces/ts";
+
+// These types/enums used to be defined in duplicate here and exported. They are re-exported to avoid breaking changes.
+export { ApplicableRefactorInfo, ClassificationType, CompletionsTriggerCharacter, CompletionTriggerKind, OrganizeImportsMode, RefactorTriggerReason, RenameInfoFailure, SemicolonPreference, SignatureHelpTriggerReason, SymbolDisplayPart, UserPreferences };
+
+type ChangeStringIndexSignature<T, NewStringIndexSignatureType> = { [K in keyof T]: string extends K ? NewStringIndexSignatureType : T[K]; };
+type ChangePropertyTypes<T, Substitutions extends { [K in keyof T]?: any; }> = {
```
jakebailey, line 36: These will have to be exported for the API to be complete. (At least, that's what it looked like when I was testing the bundler fix).

src/server/protocol.ts
```diff
@@ -630,66 +625,6 @@ export interface GetMoveToRefactoringFileSuggestions extends Response {
     };
 }
 
-/**
- * A set of one or more available refactoring actions, grouped under a parent refactoring.
- */
-export interface ApplicableRefactorInfo {
```
andrewbranch, line 636: This is an example of an interface that was copied exactly from services/types.ts. This can simply be imported instead of duplicated.

src/server/protocol.ts
```diff
@@ -380,27 +397,7 @@ export interface OutliningSpansRequest extends FileRequest {
...
-     * the 'Collapse to Definitions' command is invoked.
-     */
-    autoCollapse: boolean;
-
-    /**
-     * Classification of the contents of the span
-     */
-    kind: OutliningSpanKind;
-}
+export type OutliningSpan = ChangePropertyTypes<ts.OutliningSpan, { textSpan: TextSpan; hintSpan: TextSpan; }>;
```
andrewbranch, line 400: This is an example of an interface that was copied from services/types.ts with a couple changed property types. I added a utility for this purpose. I only used it in cases where the majority of properties were identical; when there were many changes, I let the near-duplicate stand. Note that this is a homomorphic mapped type so JSDoc from the source type’s properties is retained.

src/server/protocol.ts
```diff
@@ -1,24 +1,41 @@
...
     TodoCommentDescriptor,
     TypeAcquisition,
+    UserPreferences,
 } from "./_namespaces/ts";
+import {
+    ClassificationType,
+    CompletionTriggerKind,
+    OrganizeImportsMode,
+    SemicolonPreference,
+} from "./_namespaces/ts";
```
andrewbranch, line 30: These were identical copies of const enums previously duplicated in this file, which can simply be imported now. Since we have `preserveConstEnums` on, they have to be imported without `import type` and re-exported below to ensure the shape of this module doesn’t change. (I validated this manually by comparing `Object.keys(ts.server.protocol)` before and after this PR.)

src/server/protocol.ts
```diff
@@ -3478,292 +3074,53 @@ export const enum IndentStyle {
...
-    None = "None",
-    Preserve = "Preserve",
-    ReactNative = "ReactNative",
-    React = "React",
+    None = "none",
+    Preserve = "preserve",
+    ReactNative = "react-native",
+    React = "react",
+    ReactJSX = "react-jsx",
+    ReactJSXDev = "react-jsxdev",
```
andrewbranch, line 3095: All of these compiler options enums were missing values, but this one also had a value `ReactNative = "ReactNative"` which has forever been broken, as far as I can tell. When inferred or external compiler options are sent over the wire, enum-typed properties are allowed to be, for example, `ts.JsxEmit | protocol.JsxEmit`; i.e. a number or a corresponding string value. However, the conversion method simply uses the same parser as command-line or tsconfig values, which means these string values need to match the ones in the maps defined in `commandLineParser.ts`, which are kebab-cased. Uppercase/lowercase letters don’t matter, but `"ReactNative"` would not have worked since it was missing the hyphen.

I changed all of these enum values to lowercase for consistency, but `react-native` was the only one that strictly matters. This is technically a type-level breaking change.

src/server/protocol.ts
```diff
@@ -3772,44 +3129,30 @@ export const enum NewLineKind {
...
+{
+    type AssertKeysComplete<Source extends { [K in keyof Target]: any; }, Target> = Source;
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    type CopiedTypesComplete = [
+        AssertKeysComplete<typeof ModuleResolutionKind, typeof ts.ModuleResolutionKind>,
+        AssertKeysComplete<typeof ModuleKind, typeof ts.ModuleKind>,
+        AssertKeysComplete<typeof ScriptTarget, typeof ts.ScriptTarget>,
+        AssertKeysComplete<typeof JsxEmit, typeof ts.JsxEmit>,
+        AssertKeysComplete<typeof IndentStyle, typeof ts.IndentStyle>,
+    ];
```
andrewbranch, line 3157: As mentioned before, what really matters is that the stringified enums represent every value for each option that the command line parser can recognize. Those strings aren’t encoded in types anywhere else that I know of, so rather than checking that directly, these checks just make sure each enum has all the keys the source enum has, protecting us from forgetting to update the protocol when compiler options get new values. The enum values here in the protocol could still be typo’d, as was the case for `"ReactNative"`

src/server/protocol.ts
```diff
@@ -32,8 +32,10 @@ import {
 // These types/enums used to be defined in duplicate here and exported. They are re-exported to avoid breaking changes.
 export { ApplicableRefactorInfo, ClassificationType, CompletionsTriggerCharacter, CompletionTriggerKind, OrganizeImportsMode, RefactorTriggerReason, RenameInfoFailure, SemicolonPreference, SignatureHelpTriggerReason, SymbolDisplayPart, UserPreferences };
 
-type ChangeStringIndexSignature<T, NewStringIndexSignatureType> = { [K in keyof T]: string extends K ? NewStringIndexSignatureType : T[K]; };
-type ChangePropertyTypes<T, Substitutions extends { [K in keyof T]?: any; }> = {
+/** @internal */
```
jakebailey, line 35: Internal isn't enough, as they are used within the public API.

andrewbranch: Well that’s a bummer

andrewbranch: I get why that can’t be dropped from declaration emit, but why do they need to be exported?

andrewbranch: Maybe I’ll work on patching the bundler to allow this

jakebailey: The bundler is super dumb, and just walks exports and quite literally copies code over from the dts files. All it knows is when it's about to shadow another symbol of the same name. Then later, a test verifies that the code actually compiles, which only happens when the API is complete and consistent.

Theoretically we could mark some things as not being exported, but at the moment all of our namespaces are implicitly exporting everything, so we'd need to swap everything over to exports and then come up with some other tag that says to not actually export it, or something.

Or, completely redo the bundler to not be as dumb!

sheetalkamat: I think our dts emitter works the same way isnt it.. it would want changePropertyTypes to be exported and if its marked internal the resulting d.ts would not have it and it would generate d.ts that has error..

sheetalkamat: But i like this change though its not very readable..  So if we could write a script to generate these parts it might  be better.

jakebailey: > it would want changePropertyTypes to be exported and if its marked internal the resulting d.ts would not have it and it would generate d.ts that has error..

I think that we'd actually just generate a namespace with export modifiers on everything, such that the helper type was not visible; see the dts emit here: [Playground Link](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBAOwIYFtgGcxIMbDgMQgjgG8AoMuOGATzDwBEIBlCNGACwEsEBzAHgAqAPjgBeUnBAAuOILgBfANwUqAE2DYANkih4AZgFcE2GFwgI4alm2CceAkQAoZcgJSymrdtz5DhKpRwoJCwcNgW6PD6ROJWNj4OTgCMAEwAzAAsbioKQA)

The difference is that the bundler only walks the exported symbols only and assumes that the API is complete, not ever going through to find things it missed.

scripts/dtsBundler.mjs
```diff
@@ -310,6 +318,10 @@ function verifyMatchingSymbols(decl) {
             }
             const symbolInScope = findInScope(symbolOfNode.name);
             if (!symbolInScope) {
+                if (symbolOfNode.declarations?.every(d => isLocalDeclaration(d) && d.getSourceFile() === decl.getSourceFile()) && !isSelfReference(node, symbolOfNode)) {
+                    // The symbol is a local that needs to be copied into the scope.
+                    scopeStack[scopeStack.length - 1].locals.set(symbolOfNode.name, { symbol: symbolOfNode, writeTarget: isInternal ? WriteTarget.Internal : WriteTarget.Both });
+                }
                 // We didn't find the symbol in scope at all. Just allow it and we'll fail at test time.
                 return node;
```
jakebailey, line 326: Maybe not for this PR, but I do feel a bit like I should have originally just made this error.

scripts/dtsBundler.mjs
```diff
@@ -310,6 +318,10 @@ function verifyMatchingSymbols(decl) {
             }
             const symbolInScope = findInScope(symbolOfNode.name);
             if (!symbolInScope) {
+                if (symbolOfNode.declarations?.every(d => isLocalDeclaration(d) && d.getSourceFile() === decl.getSourceFile()) && !isSelfReference(node, symbolOfNode)) {
+                    // The symbol is a local that needs to be copied into the scope.
+                    scopeStack[scopeStack.length - 1].locals.set(symbolOfNode.name, { symbol: symbolOfNode, writeTarget: isInternal ? WriteTarget.Internal : WriteTarget.Both });
```
jakebailey, line 323: It seems moderately weird for this verification function to modify the locals, but not sure what's better.

andrewbranch: I feel like I considered either renaming the function or returning the set of new locals to be appended before I paused this work.

tests/baselines/reference/api/typescript.d.ts
```diff
@@ -14,93 +14,31 @@ and limitations under the License.
...
-            readonly typings: string[];
-            readonly unresolvedImports: SortedReadonlyArray<string>;
-            readonly kind: ActionSet;
-        }
-        interface WatchTypingLocations extends ProjectResponse {
-            /** if files is undefined, retain same set of watchers */
-            readonly files: readonly string[] | undefined;
-            readonly kind: ActionWatchTypingLocations;
-        }
+    export namespace server {
```
jakebailey, line 17: What declaration is outside of the server namespace such that everything in the ts namespace now gets an export modifier?

andrewbranch: There’s some stuff in the internal version, like I was showing you earlier. But, looking closer at it, the specific locals that are causing this one seem like they’re being duplicated from `ts.tracingEnabled` to `ts`, so I’ll take another look.

jakebailey: Yeah, I was just assuming that because this PR is only changing the protocol that other stuff wouldn't change as our current API is fully exported and complete.

scripts/dtsBundler.mjs
andrewbranch: @jakebailey any more thoughts on this? I agree this code isn’t great. Otherwise I think the current state is ready for review.

sheetalkamat (approved):

jakebailey (approved):
LGTM, though at some level I wonder if we should switch everything to `export` now such that future PRs which modify the top level API don't have to incur the diff change. But, another PR for that would be fine.

(I'm waffling on whether or not an export modifier is good or bad at this point; before we didn't have any so I was happy because it meant a smaller package.)

Issue #57286, opened 2024-02-04
Wrong type inference, returns the base class type instead of the child class type


denis-migdal:
### 🔎 Search Terms

None

### 🕗 Version & Regression Information

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about _________

### ⏯ Playground Link

[...]

PR #42747, opened 2021-02-10
Properly instantiate inferred constraints in conditional types

ahejlsberg:
Fixes #42636.

PR #31455, opened 2019-05-18
Fix conditional type type parameter leak

weswigham:
Fixes #31402

PR #57362, opened 2024-02-09
Instantiate earlier inferred constraints in conditional types

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/57286#issuecomment-1927920336

some historical related context can be found in https://github.com/microsoft/TypeScript/pull/31455 and https://github.com/microsoft/TypeScript/pull/42747

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -18636,11 +18636,6 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         return type;
     }
 
-    function maybeCloneTypeParameter(p: TypeParameter) {
-        const constraint = getConstraintOfTypeParameter(p);
-        return constraint && (isGenericObjectType(constraint) || isGenericIndexType(constraint)) ? cloneTypeParameter(p) : p;
-    }
-
     function isSimpleTupleType(node: TypeNode): boolean {
         return isTupleTypeNode(node) && length(node.elements) > 0 &&
             !some(node.elements, e => isOptionalTypeNode(e) || isRestTypeNode(e) || isNamedTupleMember(e) && !!(e.questionToken || e.dotDotDotToken));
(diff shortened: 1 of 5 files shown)


jakebailey:
@typescript-bot test top200
@typescript-bot run dt

src/compiler/checker.ts
```diff
@@ -18683,44 +18678,33 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
-                //    * Set the clones to both map the conditional's enclosing `mapper` and the original params
-                //    * instantiate the extends type with the clones
+                //    * combine `context.nonFixingMapper` with `mapper` so their constraints can be instantiated in the context of `mapper` (otherwise they'd only get inference context information)
                 //    * incorporate all of the component mappers into the combined mapper for the true and false members
-                // This means we have three mappers that need applying:
+                // This means we have two mappers that need applying:
                 //    * The original `mapper` used to create this conditional
-                //    * The mapper that maps the old root type parameter to the clone (`freshMapper`)
-                //    * The mapper that maps the clone to its inference result (`context.mapper`)
-                const freshParams = sameMap(root.inferTypeParameters, maybeCloneTypeParameter);
```
Andarist, line 18702: This PR - in its current form - removes the context of those fresh params.

They were introduced in [this PR](https://github.com/microsoft/TypeScript/pull/31455) to address a type parameter leak. From what I understand, the leak was caused by non-instantiated constraints of infer type parameters, and those fresh params were introduced to "wire up" the `mapper` so it would partake in the instantiation of those constraints. Those instantiations are made using `context.nonFixingMapper` so I figured out that maybe this can be simplified by combining that with the `mapper`. 

The problem with `maybeCloneTypeParameter` that was used to create those `freshParams` is that the check inside it is non-exhaustive:
```ts
isGenericObjectType(constraint) || isGenericIndexType(constraint)
```

When it comes to the added test case, this check doesn't cover `Klass<V>` - that's neither of the tested things. (when the inferred `Klass<number>` gets checked against that constraint it fails the assignability check and the inferred result gets replaced by the constraint which in turn is instantiated by the `combinedMapper`)

So the alternative fix would be to extend this check to cover more situations. That can't be done exhaustively and cheaply. The closest check to that is `couldContainTypeVariables` and that's not supposed to be used in situations like this, it's just an optimization.

So given the fact above, one extra alternative solution comes to mind... `freshParams` could always be created. I verified that it would fix the issue and that it wouldn't break any existing case.

weswigham (approved):
IIRC, I added the `maybeCloneTypeParameters` bit back when it was supposed that the type identities produced here could be bad for perf. Well, we have much better perf tests now, and _they look fine to me_, so I think this is pretty fine. Now, there may be some edge case that it effects in an outsized way, but hopefully if/when that comes up, we can come up with something nicer to make that work out, because this has always been a bit off-putting to me.

DanielRosenwasser:
@typescript-bot pack this

PR #57363, opened 2024-02-10
Add replacement release pipeline

jakebailey:
This creates a new release pipeline based on our existing release pipeline to meet new standards.

### DIFF
--- a/azure-pipelines.release.yml
+++ b/azure-pipelines.release.yml
@@ -35,23 +35,16 @@ extends:
                   npm --version
                 displayName: 'Install packageManager from package.json'
 
-              - script: |
-                  if [ -f ./package-lock.json ]; then
-                    npm ci
-                  else
-                    npm install
-                  fi
-                displayName: 'npm install'
+              - script: npm ci
+                displayName: 'npm ci'
 
-              - script: 'npm run test -- --no-lint'
-                displayName: 'Run npm run test -- --no-lint'
+              - script: 'npm test'
+                displayName: 'npm test'
 
               - script: |
-                  TASK_RUNNER=`npm list --depth 1 --package-lock-only hereby > /dev/null 2>&1 && echo hereby || echo gulp`
-
-                  npx $TASK_RUNNER LKG
-                  npx $TASK_RUNNER clean
-                  npm pack ./
+                  npx hereby LKG
+                  npx hereby clean
+                  npm pack
                 displayName: 'LKG, clean, pack'
 
               - task: CopyFiles@2
--- a/azure-pipelines.release.yml
+++ b/azure-pipelines.release.yml
@@ -15,7 +15,7 @@ extends:
   parameters:
     pool:
       name: AzurePipelines-EO
-      image: 1ESPT-Ubuntu22.04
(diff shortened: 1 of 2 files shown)


azure-pipelines.release.yml
jakebailey, line 1: I am not sold on the location of this file; would have preferred to have it in a build subdir, but hilariously we just have `scripts/build`. We already stick some files in `scripts` that aren't scripts, so maybe that is sufficient.

azure-pipelines.release.yml
```diff
@@ -0,0 +1,64 @@
...
+
+extends:
+  template: v1/1ES.Official.PipelineTemplate.yml@1esPipelines
+  parameters:
+    pool:
+      name: AzurePipelines-EO
+      image: 1ESPT-Ubuntu22.04
+      os: linux
+
+    stages:
```
jakebailey, line 21: Normally I would not use stages, but since this is a template, all of this stuff are parameters specific to the template, so I have to use them.

azure-pipelines.release.yml
```diff
@@ -0,0 +1,64 @@
...
+              - checkout: self
+                clean: true
+
+              - task: NodeTool@0
+                inputs:
+                  versionSpec: 20.x
+                displayName: 'Install Node'
+
+              - script: |
+                  which jq && npm install -g `jq -r '.packageManager' < package.json`
```
jakebailey, line 37: Not super clear what software packages are available on the machines. `jq` may not be available.

azure-pipelines.release.yml
```diff
@@ -0,0 +1,64 @@
...
+      type: git
+      name: 1ESPipelineTemplates/1ESPipelineTemplates
+      ref: refs/tags/release
+
+extends:
+  template: v1/1ES.Official.PipelineTemplate.yml@1esPipelines
+  parameters:
+    pool:
+      name: AzurePipelines-EO
+      image: 1ESPT-Ubuntu22.04
```
jakebailey, line 18: This may be better off as `1ESPT-Mariner2.0`. That definitely won't have `jq` 😄

jakebailey:
```suggestion
      image: 1ESPT-Mariner2.0
```
May as well try.

azure-pipelines.release.yml
```diff
@@ -0,0 +1,64 @@
...
+      name: 1ESPipelineTemplates/1ESPipelineTemplates
+      ref: refs/tags/release
+
+extends:
+  template: v1/1ES.Official.PipelineTemplate.yml@1esPipelines
+  parameters:
+    pool:
+      name: AzurePipelines-EO
+      image: 1ESPT-Mariner2.0
+      os: linux
```
noshlag, line 19: Is this the pipeline you want to be running on, or was this the default configuration from the migration tool?

jakebailey: I wrote this pipeline by hand; this was just recommended in the docs.

noshlag (approved):

DanielRosenwasser (approved):

PR #57364, opened 2024-02-10
fix(57348): Auto-complete on satisfies auto-imports satisfies 

### LINKED ISSUE #57348: Auto-complete on `satisfies` auto-imports `satisfies`
### 🔎 Search Terms

satisfies auto import vscode

- This is the behavior in every version I tried, and I (didn't) reviewed the FAQ for entries about it because @DanielRosenwasser told me to open an issue

### ⏯ Playground Link

_No response_

### 💻 Code

Create a file that has `export function satisfies() {}`. Then, in a different file type `const a = 1 sat` and hit enter (to select the "satisfies" auto-complete in vscode).

### 🙁 Actual behavior

The TS language server will add an `import { satisfies } from "..."`.

### 🙂 Expected behavior

It should not do it.

This is only a problem if you have a module that exports `satisfies` somewhere, but this is very common: https://www.npmjs.com/package/@types/semver does and has 21M weekly downloads.

It also happens for `as`, but  I have never seen a module exporting an `as` function so I never encountered this in practice.

a-tarasyuk:
Fixes #57348

### DIFF
--- a/src/services/completions.ts
+++ b/src/services/completions.ts
@@ -4005,7 +4005,6 @@ function getCompletionData(
                         if (!isIdentifierText(symbolName, getEmitScriptTarget(host.getCompilationSettings()))) return false;
                         if (!detailsEntryId && isStringANonContextualKeyword(symbolName)) return false;
                         if (!isTypeOnlyLocation && !importStatementCompletion && !(targetFlags & SymbolFlags.Value)) return false;
-                        if (!isTypeOnlyLocation && contextToken && isExpression(contextToken) && contains([SyntaxKind.AsKeyword, SyntaxKind.SatisfiesKeyword], stringToToken(symbolName))) return false;
                         if (isTypeOnlyLocation && !(targetFlags & (SymbolFlags.Module | SymbolFlags.Type))) return false;
 
                         // Do not try to auto-import something with a lowercase first letter for a JSX tag
@@ -5902,5 +5901,7 @@ function isContextualKeywordInAutoImportableExpressionSpace(keyword: string) {
         keyword === "declare" ||
         keyword === "module" ||
         keyword === "namespace" ||
-        keyword === "type";
+        keyword === "type" ||
+        keyword === "satisfies" ||
+        keyword === "as";
 }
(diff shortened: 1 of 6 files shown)


src/services/completions.ts
```diff
@@ -4004,7 +4005,9 @@ function getCompletionData(
                         if (!isIdentifierText(symbolName, getEmitScriptTarget(host.getCompilationSettings()))) return false;
                         if (!detailsEntryId && isStringANonContextualKeyword(symbolName)) return false;
                         if (!isTypeOnlyLocation && !importStatementCompletion && !(targetFlags & SymbolFlags.Value)) return false;
+                        if (!isTypeOnlyLocation && contextToken && isExpression(contextToken) && contains([SyntaxKind.AsKeyword, SyntaxKind.SatisfiesKeyword], stringToToken(symbolName))) return false;
```
DanielRosenwasser, line 4008: Special-casing just these two feels off - why can't we just return both the auto-import and the keyword? For example, if you have a module like this:

```ts
namespace foo {
    export function await() {

    }
}

export = foo;
```

Both the keyword and auto-import are available in other files.

a-tarasyuk: As I understand it, this _"special-casing"_ should apply in cases where the user is in a _"casting/satisfies"_ position.

```ts
1 a/**/
1 sati/**/
// should we suggest auto-import here?
```

DanielRosenwasser: I suppose that's fine - in which case, at the very least that you need to check for no line terminator before the position, right?

andrewbranch: Isn’t the keyword supposed to be present and sorted above the auto-import? Here’s `async`, for example:

image

In a position like `1 satisf/**/`, it’s true that an identifier would be invalid, so we could consider blocking _all_ auto-imports there. However, I sometimes find myself wanting completions in the middle of a messy edit:

```ts
const x = someFn bla/**/, foo); // going to backtrack and add the missing `(`
```

so I’m hesitant to block completions based on relying too much on the current state of the parse tree. It feels to me like the bug is just that the keywords don’t show up here, even though they’re valid:

image

a-tarasyuk: Screenshot 2024-02-14 at 03 42 12
Screenshot 2024-02-14 at 03 42 26

Mateusnasciment: special-casing two items might seem odd. Consider returning both auto-import and keyword suggestions to enhance completion ...

andrewbranch (approved):

andrewbranch:
Thanks @a-tarasyuk!

Issue #57356, opened 2024-02-09
Unbound type parameter leak when calling union of methods in some circumstances


jcalz:
### 🔎 Search Terms

generic type parameter, union of functions, leak unbound escaped rogue fugitive 

### 🕗 Version & Regression Information

- This is the behavior in every version I tried (back to 4.2 anyway), and I reviewed the FAQ for entries about leaky generic

### ⏯ Playground Link

[Furnished](https://www.typescriptlang.org/play?ts=5.4.0-dev.20240209#code/JYOwLgpgTgZghgYwgAgKJSgHgCoD5kDeAUMqcjJgKq4AUcAXMjdFI9gJTIC8+l7j6LNQDcRAL5FQkWIhQB5ANY58xMuTqNmGNpx7IAriAUgA9gHcQ-NBmWiJAEwgIANnCgoEJkAGcwyCAI2AAz4AD7WWACMYciKmABMuKKePn4Q8dz+AHQwzNz4EOyiAPTFagB6APxEpWqkZgAWcH7A3siUlcgAtF345UQDQA)

### 💻 Code

```ts
interface Err<T> {
    f<U>(a: (err: T) => U): Err<U>;
}
interface Ok<T> {
    f(a: (err: T) => unknown): Err<T>;
}
declare const e: Err<0> | Err<1> | Ok<2>;
const e2 = e.f(e => e);
//    ^? const e2: Err<U> | Err<2> | Err<0 | 1 | 2>
//      what is U? --> ^
```

### 🙁 Actual behavior

The type of `e2` is a union that includes `Err<U>`, where `U` is an unbound type parameter.

### 🙂 Expected behavior

I expect type parameters to stay in their scopes. As for what `Err<U>` should actually be... I think, uh, maybe it should be another copy of `Err<0 | 1 | 2>` or just omitted from the union? 

### Additional information about the issue

This is a minimal example derived from [a Stack Overflow question](https://stackoverflow.com/q/77967261/2887218).  

Not sure if #43961 and #55467 are related, or if those are different manifestations of type parameter escape.

PR #57371, opened 2024-02-11
Fixed type parameter leak in union calls with mixed type parameter presence

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/57356

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -13541,6 +13541,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         if (paramMapper) {
             result.mapper = left.compositeKind !== TypeFlags.Intersection && left.mapper && left.compositeSignatures ? combineTypeMappers(left.mapper, paramMapper) : paramMapper;
         }
+        else if (left.compositeKind !== TypeFlags.Intersection && left.mapper && left.compositeSignatures) {
+            result.mapper = left.mapper;
+        }
         return result;
     }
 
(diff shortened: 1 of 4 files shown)


src/compiler/checker.ts
```diff
@@ -13540,6 +13540,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         result.compositeSignatures = concatenate(left.compositeKind !== TypeFlags.Intersection && left.compositeSignatures || [left], [right]);
         if (paramMapper) {
```
Andarist, line 13541: the gist of the issue was that first 2 calls with type params were unionized here (so `paramMapper` was created and attached to became `left.mapper` in the next iteration) but then that composite signature was unionized with one without the type params so the `left.mapper` wasn't attached to the `result` (because `paramMapper` was not created this time)

src/compiler/checker.ts
```diff
@@ -13540,6 +13540,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         result.compositeSignatures = concatenate(left.compositeKind !== TypeFlags.Intersection && left.compositeSignatures || [left], [right]);
         if (paramMapper) {
             result.mapper = left.compositeKind !== TypeFlags.Intersection && left.mapper && left.compositeSignatures ? combineTypeMappers(left.mapper, paramMapper) : paramMapper;
+        } 
+        else if (left.compositeKind !== TypeFlags.Intersection && left.mapper && left.compositeSignatures) {
```
Andarist, line 13544: This whole function is very similar to `combineSignaturesOfIntersectionMembers` so I thought that there might be a similar problem there but so far I couldn't create a test case that would manifest an issue there.

It's also worth noting that both of those functions check `left.compositeKind` but it's never the "other one" in the existing test suite (in `combineSignaturesOfIntersectionMembers` we never witness `left.compositeKind === TypeFlags.Union` and in `combineSignaturesOfUnionMembers` we never witness `left.compositeKind === TypeFlags.Intersection`). So either some tests are missing or those branches are unused.

weswigham:
@typescript-bot run dt
@typescript-bot test top100
@typescript-bot pack this

PR #57391, opened 2024-02-12
fix(57386): Invalid use of 'eval' when defining a namespaced eval function

### LINKED ISSUE #57386: Invalid use of 'eval' when defining a namespaced eval function
### 🔎 Search Terms

eval, strict mode, namespace

### ⏯ Playground Link

https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEBzCB7ARlC8DeAoeB8IAHgA4owAu8AdlALYgDOpUYCNIlAZhAJbEc+QiJLkq8bgFcaYSnxQ0iAN0wAKAJQAueMpR9gAbmEEAvrnO4xFatnNA

### 💻 Code

```ts
declare global {
    export namespace myNamespace {
        export function eval(): void;
    }
}

export {}
```

### 🙁 Actual behavior

Invalid use of 'eval'. Modules are automatically in strict mode.

### 🙂 Expected behavior

No error

I see in https://github.com/microsoft/TypeScript/issues/3036 that this behavior is by design as defining an 'eval' function is forbidden in strict mode. However, the namespace declaration above describes runtime code more like the following, which is perfectly valid in strict mode:

```ts
const myNamespace = {
    eval() {
        // ...
    }
}
```

Would it be possible to not fail on namespaced function declarations?

a-tarasyuk:
Fixes #57386

### DIFF
--- a/src/compiler/binder.ts
+++ b/src/compiler/binder.ts
@@ -2563,7 +2563,7 @@ function createBinder(): (file: SourceFile, options: CompilerOptions) => void {
     }
 
     function checkStrictModeFunctionName(node: FunctionLikeDeclaration) {
-        if (inStrictMode) {
+        if (inStrictMode && !(node.flags & NodeFlags.Ambient)) {
             // It is a SyntaxError if the identifier eval or arguments appears within a FormalParameterList of a strict mode FunctionDeclaration or FunctionExpression (13.1))
             checkStrictModeEvalOrArguments(node, node.name);
         }
--- a/tests/baselines/reference/evalOrArgumentsInDeclarationFunctions.symbols
+++ b/tests/baselines/reference/evalOrArgumentsInDeclarationFunctions.symbols
@@ -1,0 +1,18 @@
+//// [tests/cases/compiler/evalOrArgumentsInDeclarationFunctions.ts] ////
+
+=== /a.d.ts ===
+declare global {
+>global : Symbol(global, Decl(a.d.ts, 0, 0))
+
+    export namespace ns {
+>ns : Symbol(ns, Decl(a.d.ts, 0, 16))
+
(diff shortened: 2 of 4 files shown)


src/compiler/binder.ts
```diff
@@ -2563,7 +2563,7 @@ function createBinder(): (file: SourceFile, options: CompilerOptions) => void {
     }
 
     function checkStrictModeFunctionName(node: FunctionLikeDeclaration) {
-        if (inStrictMode) {
+        if (inStrictMode && !(node.flags & NodeFlags.Ambient)) {
```
sandersn, line 2566: This will allow for un-namespaced `declare function eval(): void` too, right? Is there any legal strict-mode code that could be typed by that? I can't think of any.

Restricting this exemption to namespaces makes sense to me if there aren't any other examples with legal emit.

sandersn (changes requested):
I think this should be restricted to namespaced-contained functions, or else have a lot more tests with examples of ambient variables, functions, etc named `eval`.

src/compiler/binder.ts
a-tarasyuk: > This will allow for un-namespaced declare function eval(): void too, right?

Yes, I've added tests to cover this case

sandersn: I'd like to disallow `declare function eval(): void`. You can't write a matching `function eval() { }` or `var eval = function() { }`, can you?

RyanCavanaugh: How do _we_ declare the global `eval` if that's illegal?

sandersn: good point!

sandersn (approved):

PR #57401, opened 2024-02-13
Some of the fixes picked out from my work for file updates

sheetalkamat:
- 6a573b2bfb61c45219e79abb6b8feef884fb6a12 When trying to find the default tsconfig file for source file of d.ts through source maps, before fix we didnt use the cached config file name, with this fix it does
- 3d7061fca3cbb6c0f4e90fe2c41c3af2917d0b1c When opening external projects, we are not opening projects first before closing the unused projects so the script infos and file watches can be reused
- 6a573b2bfb61c45219e79abb6b8feef884fb6a12 We were doing unordered remove of containing projects when project was closed from script info so depending on how and when this happened the default project would be different. So fixing this to keep the order.

### DIFF
--- a/tests/baselines/reference/tsserver/externalProjects/deleting-config-file-opened-from-the-external-project-works-with-lazyConfiguredProjectsFromExternalProject.js
+++ b/tests/baselines/reference/tsserver/externalProjects/deleting-config-file-opened-from-the-external-project-works-with-lazyConfiguredProjectsFromExternalProject.js
@@ -193,9 +193,7 @@ Info seq  [hh:mm:ss:mss] request:
               "allowJs": false
             },
             "typeAcquisition": {
-              "include": [],
-              "exclude": [],
-              "enable": true
+              "include": []
             }
           }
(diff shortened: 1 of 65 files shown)


jakebailey (approved):

tests/baselines/reference/tsserver/externalProjects/deleting-config-file-opened-from-the-external-project-works-with-lazyConfiguredProjectsFromExternalProject.js
```diff
@@ -193,9 +193,7 @@ Info seq  [hh:mm:ss:mss] request:
               "allowJs": false
             },
             "typeAcquisition": {
-              "include": [],
-              "exclude": [],
-              "enable": true
+              "include": []
             }
```
sheetalkamat, line 197: This changed because the previous command incorrectly updated the request object itself so passed in request later was changed too..

PR #57403, opened 2024-02-13
Propagate outer type parameters of single signature types

### LINKED ISSUE #55467: nested generic function inference causes unbound type parameter leak
### 🔎 Search Terms

unbound generic, unbound type parameter

This is the behavior in every version I tried, and I reviewed the FAQ.

### ⏯ Playground Link

https://www.typescriptlang.org/play?jsx=0#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXwHcYoAHAHgA0A+ACgA8AueCgSiYG95Hn4BfAKFCRYCFOmx54YKBAhkAgvBB0MIVMADO8KKgCeAbQC6AGngAVWt07caAOnuwA5hqbyW8ALxVzfU-dtOLvBuTGb8-GB4GhjwECBQANaeUjIQNESkNGQW9KHuXlwspgCMLPwA9OXw1fAAegD8-EA

### 💻 Code

```ts
declare function wrap<X>(x: X): { x: X }
declare function call<A extends any[], T>(x: { x: (...args: A) => T }, ...args: A): T

const leak = call(wrap(<T>(x: T) => x), 1)
//    ^? - const leak: A[0]
```

### 🙁 Actual behavior

The type of `leak` involves an unbound type parameter.

### 🙂 Expected behavior

The type of `leak` should be `number`.

_No response_

weswigham:
Fixes #55467

This is wildly more complicated than the issue would make you think, because it uncovers a few issues in how we construct "single signature types" and how that interacts with free type variable inference.

Specifically:
* Since we project type parameters from outer signatures onto inner expression types, we need to instantiate argument expressions with the inference context mapper to complete those types. The signature mapper would overeagerly instantiate inner call expressions (one of the new tests demonstrates this), so the inference context itself must be used. This isn't *wholly unusual* - we already instantiate expression types with the inference context in a number of scenarios (at least in the first inference pass when we're looking to fix argument types, or when the expression is maybe a literal type). This just ensures we *always* do it when we go to check signature argument expression compatibility.
* Since those inner expression types may reference the type parameters of the signature, when we're within the body of a signature and also checking a call to that same signature, we check against a copy of the signature with fresh type parameters, rather than the original, so we don't confuse references to the type parameters of the signature declaration and the inferred type arguments of the nested call.
* The single-signature types we produce now capture their outer type parameters (which may come from active inference contexts), so they can be correctly instantiated, so when they contain those projected type parameters, they can be later replaced with the inference results.
* We discard inferences of the form `[A[0]]` for `A`, since "the first element of A is the first element of A" is a tautological inference that adds no useful information, and always yields `A`'s constraint if selected. This improves inferences in scenarios involving a rest parameter alongside another inference site, as in the linked issue.


weswigham:
@typescript-bot run dt
@typescript-bot test top100

weswigham:
@typescript-bot pack this

weswigham:
@typescript-bot run dt
@typescript-bot test top100

fatcerberus:
*propagate.  Sorry, I had to.

weswigham:
@typescript-bot run dt
@typescript-bot test top100

weswigham:
@typescript-bot pack this

src/compiler/checker.ts
```diff
@@ -7088,7 +7089,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
 
                 const abstractSignatures = filter(resolved.constructSignatures, signature => !!(signature.flags & SignatureFlags.Abstract));
                 if (some(abstractSignatures)) {
-                    const types = map(abstractSignatures, getOrCreateTypeFromSignature);
+                    const types = map(abstractSignatures, s => getOrCreateTypeFromSignature(s));
```
jakebailey, line 7092: Is this leftover from a previous refactor?

jakebailey: Oh, it's just the second optional param not being assignable to map's param.

jakebailey (approved):
From what I remember from your previous explanation, this seemed good to go?

PR #57409, opened 2024-02-15
Refactor workflows for new typescript-bot triggers

jakebailey:
This goes with https://github.com/microsoft/typescript-bot-test-triggerer/pull/36.

Every action now uses a workflow dispatch (not a repo dispatch; the former is newer), along with new parameters for data sent by the bot. These options must always be defined for workflows the bot calls.

The composite action `.github/actions/post-workflow-result/action.yml` is used to share code between actions for updating the status comment and posting results. All of our workflows that trigger the bot use this, except the "repros" workflow, which has its own code for stuff (which I will just not touch for now).

I'm not using a "reusable workflow" for this, as those jobs run in an entirely different unrelated context, whereas composite actions are just sorta jammed into an existing workflow (which is nice).

Most workflows don't change much, save for the cherry-pick action, which surprisingly was the only action that actually replied back.


.github/workflows/create-cherry-pick-pr.yml
```diff
@@ -109,23 +131,18 @@ jobs:
...
+            core.setOutput("comment_body", commentBody);
 
-          gh pr comment "$PR" --repo ${{ github.repository }} --body "$MESSAGE"
-        if: ${{ failure() }}
-        env:
-          PR: ${{ inputs.pr || github.event.client_payload.pr }}
-          TARGET_BRANCH: ${{ inputs.target_branch || github.event.client_payload.target_branch }}
-          REQUESTING_USER: ${{ inputs.requesting_user || github.event.client_payload.requesting_user }}
-          GH_TOKEN: ${{ secrets.TS_BOT_GITHUB_TOKEN }}
+      - uses: microsoft/TypeScript/.github/actions/post-workflow-result@main
```
jakebailey, line 139: I feel sorta gross writing this; the reason this isn't just `./.github/actions/post-workflow-result` alone is because most of these workflows don't actually clone TS or clone a random version (e.g. to make a release branch).

Maybe this is a reason why I should just make a new repo for this bot stuff instead.

.github/workflows/new-release-branch.yaml
```diff
@@ -1,8 +1,40 @@
...
 
 on:
-  repository_dispatch:
-    types: [new-release-branch]
+  workflow_dispatch:
+    inputs:
+      branch_name:
+        description: Release branch name to create
+        required: true
+        type: number
```
weswigham, line 9: Shouldn't this be `string`?

.github/workflows/set-version.yaml
```diff
@@ -1,8 +1,40 @@
...
 
 on:
-  repository_dispatch:
-    types: [set-version]
+  workflow_dispatch:
+    inputs:
+      branch_name:
+        description: Release branch name to create
+        required: true
+        type: number
```
weswigham, line 9: Here, too.

.github/workflows/sync-branch.yaml
```diff
@@ -1,13 +1,32 @@
...
 on:
-  repository_dispatch:
-    types: [sync-branch]
   workflow_dispatch:
     inputs:
       branch_name:
-        description: 'Target Branch Name'
+        description: Release branch name to create
         required: true
+        type: number
```
weswigham, line 9: Also here.

weswigham (commented):
Just something I noticed while going over the new triggering workflows.

.github/workflows/new-release-branch.yaml
jakebailey: You're totally right, copy paste error, blah

Issue #46312, opened 2021-10-11
Inferences made when trying to match an overload are carried over to matching consecutive overloads


Andarist:
# Bug Report

### 🔎 Search Terms

overload inference generic consecutive subsequent following 

the only issue that I have found that **might** be related to this is [this one](https://github.com/microsoft/TypeScript/issues/39850) but I can't assess on my own if they are the same or not

### 🕗 Version & Regression Information

This is the behavior present in all 4.x versions available on the playground, including 4.5-beta.

### ⏯ Playground Link

[...]

PR #57421, opened 2024-02-15
Add checker speculation helper, use in overload resolution

### LINKED ISSUE #12733: Regression in type inference?
**TypeScript Version:**  2.1.1, 2.1.4

**Code**

```ts
declare function func(callback: () => void): any;
declare function func(callback: (arg: number) => void): any;

func(argx => { });
```

**Expected behavior:**
Successfully compile, even if `--noImplicitAny` is set. This is the behavior we saw until `2.0.10` (which appears reasonable, the second overload of `func` allows inferring that `argx: number`).

**Actual behavior:**
`error TS7006: Parameter 'argx' implicitly has an 'any' type`

### LINKED ISSUE #13430: loss of type safety when using function overloads with callbacks
**TypeScript Version:**  2.1.4
**Code**
```ts
declare function it(f: () => void): number;
declare function it(f: (x: string) => void): string;

let r = it((x) => {x});
```

**Expected behavior:**
The inferred type of `x` inside the closure should be `string`.

**Actual behavior:**
The inferred type of `x` is `any`. Moreover, this triggers noImplicitAny errors.

It appears that the first overload somehow throws off the type inference (no issue without overloads). However, the resulting type of `r` is `string` thus the second overload was ultimately chosen. Despite that, the type of `x` is not inferred to match the one in the overload.

This example worked as expected in TS 2.0.x so it appears to be a regression.

//cc @calebegg

### LINKED ISSUE #21525: Wrong callback type inferred with multiple overloads
**TypeScript Version:**  2.8.0-dev.20180131

**Search Terms:** callback type inference, incorrect overload 

**Code**

```ts
interface TestFunction {
    <T>(input: { [key: number]: T }, callback: (value: T, key: number, collection: { [key: number]: T }) => boolean): boolean;
    <T extends object>(input: T, callback: (value: T[keyof T], key: string, collection: T) => boolean): boolean;
}

const fn: TestFunction = {} as any;
fn({ a: "a", b: "b" }, (value, key) => true);
```

**Expected behavior:** Code should compile successfully

**Actual behavior:** Compile error:

```
src\test.ts (10,32): Argument of type '(value: {}, key: number) => boolean' is not assignable to parameter of type '(value: string, key: string, collection: { a: string; b: string; c: string; }) => boolean'.
  Types of parameters 'key' and 'key' are incompatible.
    Type 'string' is not assignable to type 'number'. (2345)
```

**Related Issues:** https://github.com/DefinitelyTyped/DefinitelyTyped/issues/23317

**Notes:**

- The code above works in typescript 2.4. It seems to have broken in 2.5 and later (including 2.8).
- It appears as if it's inferring the callback type based on the first overload, then failing to match it to the second overload.
- The generic types seem necessary to reproduce the issue. I tried removing them and the issue went away.
- If I explicitly pass the generic type argument, the code compiles successfully, i.e:
    ```ts
    fn<{ a: "a", b: "b" }>({ a: "a", b: "b" }, (value, key) => true);
    ```
- If I remove the second callback parameter, the code compiles successfully, i.e:
    ```ts
    fn({ a: "a", b: "b" }, (value) => true);
    ```
- Enabling/disabling `strictFunctionTypes` does not affect the result

### LINKED ISSUE #49820: Overload is not recognized correctly in some cases
# Bug Report

### 🔎 Search Terms

generics overload

- 4.7.4(current), 4.8.0-beta(beta)

### ⏯ Playground Link

[Playground link with relevant code](https://www.typescriptlang.org/play?ts=4.7.4#code/C4TwDgpgBAggjAHgCoD4oF4oG8qkgLigCIBDOIgGigDdCkoBfAbgCg9oAhRVDbXcCISIAjclVpR6zNgNgAmXjhJzCAZ2AAnAJYA7AOaNW7KBwWYcwlVHXb9hliwBmAVx0BjYFoD2OqI53IKAAUYHCEIaGEAAwAlBho8IFUYFYRVjBycehosYTUXloAJqwu7p4+fgGoIWFQEbWx8SbcKMmpYCmEplk5MXkFxU6uHt6+-oFBLFBQkVPTUAA+dfXRPbAtMXPTS0ErUI3ZzYExFHOdW4vLHemZTbEXO2ldt4f3cVgMDm4+6jQkADbwQiJGy6PRoTBBHDsIRkSg0IRERgxVjfHS-agArhdRCg-QQurQgRCUTwiREJEMFEOAD0NKgEAAHpAPLAnDpdiRQmtMYC4FQglyXr0ShzBdymrz4AKhcDhfsUVB2ZyJYcpfy6rKoCQdCA1rFRSq4DyAdLNZ0oK4ANY6LwAdx0+up-iNq0lpo14vlBpYdIZzIgrI4yqCwlVaF5XAFYe9ir9wmcwCgWiTdq0-3+UGE0BgVB8-xAuAAFlpVFA0xmoBoIKpnP8k7ptb4IBoNF4NCGw8b3f8o3UY88nUwlS7Q+G-r3PQOm3q7s6xV2TZPoxbrbaHUPO5EFT2+2PY6wgA)

### 💻 Code

```ts
type A1<T> = { type: "a1", v: T };
type B1<T> = { type: "b1", v: T };
type A2 = { a2: string };
type B2 = { b2: string };

function fn<T>(p1: (pp1: 0) => A1<T>, p2: (pp2: A2) => 0): void;
function fn<T>(p1: (pp1: 0) => B1<T>, p2: (pp2: B2) => 0): void;
function fn<T>(
  p1:
    | ((pp1: 0) => A1<T>)
    | ((pp1: 0) => B1<T>),
  p2:
    | ((pp2: A2) => 0)
    | ((pp2: B2) => 0)
) {}

const valA1: A1<string> = ({ type: "a1", v: "" });
const valB1: B1<string> = ({ type: "b1", v: "" });

// expect A
fn((ap1) => valA1, (ap2) => 0);
fn((ap1) => valA1, (ap2: A2) => 0); 
fn((ap1) => valA1, (ap2: any) => 0);
fn((ap1) => valA1, (ap2: unknown) => 0);
fn((ap1: 0) => valA1, (ap2) => 0);
// expect B
fn((bp1) => valB1, (bp2) => 0); // but it will be A, only this will result in an error
fn((bp1) => valB1, (bp2: B2) => 0); 
fn((bp1) => valB1, (bp2: any) => 0);
fn((bp1) => valB1, (bp2: unknown) => 0);
fn((bp1: 0) => valB1, (bp2) => 0);
```

### 🙁 Actual behavior

```
  Overload 1 of 2, '(p1: (pp1: 0) => A1<string>, p2: (pp2: A2) => 0): void', gave the following error.
    Type 'B1<string>' is not assignable to type 'A1<string>'.
      Types of property 'type' are incompatible.
        Type '"b1"' is not assignable to type '"a1"'.
  Overload 2 of 2, '(p1: (pp1: 0) => B1<string>, p2: (pp2: B2) => 0): void', gave the following error.
    Argument of type '(bp2: A2) => 0' is not assignable to parameter of type '(pp2: B2) => 0'.
      Types of parameters 'bp2' and 'pp2' are incompatible.
        Property 'a2' is missing in type 'B2' but required in type 'A2'.(2769)
```

### 🙂 Expected behavior

No errors

weswigham:
This both prevents inference data from failed overloads from leaking into follow-on overloads, and allows contextual parameters to become unfixed after a failed overload and adopt a new type. This fixes a lot of longstanding design limitations.

For authors in the checker, the important changes in this PR are twofold:

1. There is now a function called `speculate` - you pass it a callback, and it executes that callback within a speculative context. if the callback returns a truthy value, the context is kept when the callback completes (as you may want to do if a speculative operation is successful), otherwise it is rewound to the state it was when the callback was first invoked. Right now, it's just used in overload resolution, but you can imagine the use of it in a handful of other places.

2. To support this, `Node`- and `Symbol`-links have been refactored into classes with decorators on the properties we should speculatively cache. The decorator itself implements all the speculative caching. It basically works by recording timestamps writes occur at, and lazily discarding writes at timestamps marked as discarded upon read.

I've narrowed the set of caches we consider "speculative" down about as much as I think I safely can. Only a handful of fields on the symbol and node `links` structures (that retain context-sensitive calculated data or diagnostic data), plus diagnostics and relationship caches (since those retain diagnostic emit data!) are checkpointed and potentially rewound on speculation boundaries. Links speculation caches are lazy, but some other global structures are still eagerly copied - in particular, relationship caches (because they contain error reporting state!) are currently eagerly copied - this is probably a location ripe for improvement in real world scenarios.

For users, since this does use the speculation helper in one big location already, this:

Fixes #49820 (Previously A2 would be fixed as a parameter type during checking of the first overload, now that is undone, and B2 can be correctly assigned on the second overload check)
Fixes #13430 (Previously, `x` would get an implicit `any` assigned during checking of the first overload - that is now discarded (the function argument arity doesn't match, so the overload ultimately fails) and the second overload is selected)
Fixes #12733 (Same as above)
Fixes #21525 (Previously the first overload would project incorrect contextual types for the second overload and fail, making the second overload fail as well)

And potentially many more - many of the issues in this vein have been classified "Design Limitation", or closed as duplicate/won't fix over the years.


src/compiler/checker.ts
```diff
@@ -34848,63 +35045,67 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 if (!hasCorrectTypeArgumentArity(candidate, typeArguments) || !hasCorrectArity(node, args, candidate, signatureHelpTrailingComma)) {
                     continue;
                 }
+                const result = speculate(() => {
```
weswigham, line 35048: This is easier to read with whitespace diffs turned off.

src/tsconfig-base.json
```diff
@@ -4,8 +4,8 @@
         "outDir": "../built/local",
 
         "pretty": true,
-        "lib": ["es2020"],
-        "target": "es2020",
+        "lib": ["es2021"],
+        "target": "es2021",
```
weswigham, line 8: This bump is just to add `WeakRef` to the lib (which is supported in `node14`+). I mention it in a comment above - this can work without weak refs, it'd just guarantee transient symbols always stick around forever, rather than maybe getting collected if they weren't actually kept around anywhere. But a transient symbol not being stores somewhere more checker-lifetime'd is probably rare, so it's *probably* OK to just not use weak refs to track transient symbols if there's an issue with that. It's more defensive than anything else.

tests/baselines/reference/contextualTypingOfTooShortOverloads.types
```diff
@@ -31,11 +31,11 @@ app.use((err: any, req, res, next) => { return; });
 >app.use : IRouterHandler<MyApp> & IRouterMatcher<MyApp>
 >app : MyApp
 >use : IRouterHandler<MyApp> & IRouterMatcher<MyApp>
->(err: any, req, res, next) => { return; } : (err: any, req: any, res: any, next: any) => void
+>(err: any, req, res, next) => { return; } : (err: any, req: Request, res: Response, next: NextFunction) => void
```
weswigham, line 34: Previously, this'd get `any` argument types fixed by an earlier overload we didn't actually pick because of an arity mismatch - this is a pretty clear improvement!

tests/baselines/reference/destructuringTuple.errors.txt
```diff
@@ -25,22 +15,9 @@ destructuringTuple.ts(11,60): error TS2769: No overload matches this call.
...
-!!! related TS6502 lib.es5.d.ts:--:--: The expected type comes from the return type of this signature.
-                                                               ~~
-!!! error TS2769: No overload matches this call.
-!!! error TS2769:   Overload 1 of 2, '(...items: ConcatArray<never>[]): never[]', gave the following error.
-!!! error TS2769:     Argument of type 'number' is not assignable to parameter of type 'ConcatArray<never>'.
-!!! error TS2769:   Overload 2 of 2, '(...items: ConcatArray<never>[]): never[]', gave the following error.
-!!! error TS2769:     Argument of type 'number' is not assignable to parameter of type 'ConcatArray<never>'.
+!!! error TS2461: Type 'unknown' is not an array type.
+                                                   ~~~~
+!!! error TS18046: 'accu' is of type 'unknown'.
```
weswigham, line 20: I'm wishy washy on this one. Previously, we'd fix the first parameter to `[]` at the end of the first overload's checking, and then witness it while checking the second overload and that'd produce these errors. Now, since both overloads fail, we fall back to the constraint on the type parameter (`unknown`), which produces different errors.

tests/baselines/reference/fixingTypeParametersRepeatedly1.types
```diff
@@ -35,7 +35,7 @@ declare function g();
 >g : { <T>(x: T, y: (p: T) => T, z: (p: T) => T): T; (): any; }
 
 g("", x => null, x => x.toLowerCase());
->g("", x => null, x => x.toLowerCase()) : any
+>g("", x => null, x => x.toLowerCase()) : string
```
weswigham, line 38: Like many thing reference sin the OP, this is just a straight improvment - previously `any` would get fixed to the second parameter's parameter during overload checking early on, now that gets discarded and we get the type from the overload we actually pick - `string`.

tests/baselines/reference/fixingTypeParametersRepeatedly2.errors.txt
```diff
@@ -24,6 +28,11 @@ fixingTypeParametersRepeatedly2.ts(17,5): error TS2403: Subsequent variable decl
     declare function bar<T>(x: T, func: (p: T) => T): T;
     declare function bar<T>(x: T, func: (p: T) => T): T;
     var result = bar(derived, d => d.toBase());
-        ~~~~~~
-!!! error TS2403: Subsequent variable declarations must have the same type.  Variable 'result' must be of type 'Derived', but here has type 'Base'.
-!!! related TS6203 fixingTypeParametersRepeatedly2.ts:11:5: 'result' was also declared here.
\ No newline at end of file
+                                   ~~~~~~~~~~
```
weswigham, line 31: It's funny - this test seems to insist the overload case here and the no overload case above should have had similar errors - but that actually wasn't true until this PR. Maybe it was a forward looking comment? 🤷 This is certainly more consistent now - both overloads produce the same `Base` is not assignable to `Derived` error.

tests/baselines/reference/overloadedConstructorFixesInferencesAppropriately.errors.txt
```diff
@@ -0,0 +1,34 @@
...
+        constructor(...args: any[]) {}
+    }
+    
+    function load(): Box<{ success: true } | ErrorResult> {
+        return null as any;
+    }
+    
+    new AsyncLoader({
+        asyncLoad: load,
+        children: result => result.success as any,
```
weswigham, line 30: `result` would previously get a type fixed on it while checking the first overload (via `asyncLoader`), which'd the inform the expression checking during the second overload. Ideally inference should still work here (since a source for the better type still exists), but this might need some more changes to make work (like fine-tuning the expression checking order). That is probably the only -0 change I've seen caused by this thus far.

tests/baselines/reference/overloadresolutionWithConstraintCheckingDeferred.errors.txt
```diff
@@ -53,60 +21,23 @@ overloadresolutionWithConstraintCheckingDeferred.ts(19,14): error TS2344: Type '
...
-!!! error TS2769:         Property 'q' is missing in type 'C' but required in type 'D'.
-!!! error TS2769:   Overload 3 of 3, '(arg: (x: B) => any): number', gave the following error.
-!!! error TS2769:     Argument of type '(x: D) => G<A>' is not assignable to parameter of type '(x: B) => any'.
-!!! error TS2769:       Types of parameters 'x' and 'x' are incompatible.
-!!! error TS2769:         Property 'q' is missing in type 'B' but required in type 'D'.
-!!! related TS6502 overloadresolutionWithConstraintCheckingDeferred.ts:10:27: The expected type comes from the return type of this signature.
-!!! related TS2728 overloadresolutionWithConstraintCheckingDeferred.ts:4:15: 'q' is declared here.
-!!! related TS2728 overloadresolutionWithConstraintCheckingDeferred.ts:4:15: 'q' is declared here.
+        ~~~~~~
+!!! error TS2322: Type 'string' is not assignable to type 'number'.
```
weswigham, line 25: Previously `x` would have the type from the first overload, since it'd get immediately fixed, now `x` gets the type form the first overload that passes, which is the _second_, since `G` is assignable to `any` and not `number`. So even though the call to `G` fails with `x` being C`, the call as a whole still works out. This is actually a place we could improve - rather than just checking the overload's validity, we could check for errors in the whole subexpression tree - but that's future work.

tests/baselines/reference/overloadsWithProvisionalErrors.types
```diff
@@ -18,8 +18,8 @@ var func: {
 func(s => ({})); // Error for no applicable overload (object type is missing a and b)
 >func(s => ({})) : never
 >func : { (s: string): number; (lambda: (s: string) => { a: number; b: number; }): string; }
->s => ({}) : (s: string) => {}
->s : string
+>s => ({}) : (s: any) => {}
+>s : any
```
weswigham, line 22: These are all places where `string` would get fixed in an earlier overload and bleed into later ones, even though checking for that overload failed.

tests/baselines/reference/parenthesizedContexualTyping1.types
```diff
@@ -130,100 +130,100 @@ var h = fun((((x => x))), ((x => x)), 10);
 
 // Ternaries in parens
 var i = fun((Math.random() < 0.5 ? x => x : x => undefined), 10);
->i : any
->fun((Math.random() < 0.5 ? x => x : x => undefined), 10) : any
+>i : unknown
```
weswigham, line 133: This one is funny - `unknown` is better here because it's the constraint of the type parameter - `any` only snuck in here before because it'd get an implicit `any` during the first overload (since the position takes a `T` and not a function in the first overload), which'd then be carried forward into the following overloads. Definitely a safety improvement, but only really noticeable for people with `noImplicitAny` off.

tests/baselines/reference/spreadOfParamsFromGeneratorMakesRequiredParams.errors.txt
```diff
@@ -1,8 +1,6 @@
-error TS2318: Cannot find global type 'IterableIterator'.
 spreadOfParamsFromGeneratorMakesRequiredParams.ts(6,1): error TS2554: Expected 2 arguments, but got 1.
 
 
-!!! error TS2318: Cannot find global type 'IterableIterator'.
```
weswigham, line 5: This error only crops up while resolving one of the overloads (the iterator one, ofc), which isn't the last one that we "accept" (though it still has an error), so it gets tossed.

tests/baselines/reference/targetTypeArgs.types
```diff
@@ -37,7 +37,7 @@ foo(function(x) { x });
 >["hello"] : string[]
 >"hello" : "hello"
 >every : { <S extends string>(predicate: (value: string, index: number, array: string[]) => value is S, thisArg?: any): this is S[]; (predicate: (value: string, index: number, array: string[]) => unknown, thisArg?: any): boolean; }
->function(v,i,a) {return true;} : (v: string, i: number, a: string[]) => true
+>function(v,i,a) {return true;} : (v: string, i: number, a: string[]) => boolean
```
weswigham, line 40: This is one of those funky places where parameter type fixing would make the return type's type contextually type itself in overload passes after the first one. That no longer occurs.

tests/baselines/reference/thisConditionalOnMethodReturnOfGenericInstance.errors.txt
```diff
@@ -0,0 +1,24 @@
...
+
+
+==== thisConditionalOnMethodReturnOfGenericInstance.ts (1 errors) ====
+    class A<T> {
+        unmeasurableUsage!: {[K in keyof T]-?: T[K]};
+    }
+    
+    class B<T> extends A<T> {
+        method(): string | (this extends C ? undefined : null) {
+                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
weswigham, line 11: Per OP: I plan to back out the change making this test error, since that fix is somewhat unrelated to this, even if it did motivate it.

tests/baselines/reference/unionOfClassCalls.types
```diff
@@ -66,7 +66,7 @@ arr.map((a: number | string, index: number) => {
 
 // This case still doesn't work because `reduce` has multiple overloads :(
 arr.reduce((acc: Array<string>, a: number | string, index: number) => { 
->arr.reduce((acc: Array<string>, a: number | string, index: number) => {     return []}, []) : never[]
+>arr.reduce((acc: Array<string>, a: number | string, index: number) => {     return []}, []) : string[]
```
weswigham, line 69: Yes, this does fix the `// This case still doesn't work because reduce has multiple overloads :(` comment 😀

Now, since every overload is done in its own speculative context, the `[]` expression no longer has its type locked in by the first overload, and it can be interpreted as desired by the later overload instead.

src/compiler/checker.ts
```diff
@@ -20878,18 +21080,6 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         }
 
         if (!(checkMode & SignatureCheckMode.IgnoreReturnTypes)) {
-            // If a signature resolution is already in-flight, skip issuing a circularity error
```
weswigham, line 20881: Note for self: This is the change I want to back out, since it doesn't impact the speculative overload resolution fixes in this PR.

Andarist:
I have confirmed locally that this PR would fix https://github.com/microsoft/TypeScript/issues/46312 too

The copy-pasteable test case can be found here:
<details>
<summary>test case </summary>

```ts
// @strict: true
// @noEmit: true

interface TypegenDisabled {
  "@@xstate/typegen": false;
}
interface TypegenEnabled {
  "@@xstate/typegen": true;
}
interface EventObject {
  type: string;
}
interface ActionObject<TEvent extends EventObject> {
  _TE: TEvent;
}

interface StateMachine<
  TEvent extends EventObject,
  TTypesMeta = TypegenDisabled,
> {
  _TE: TEvent;
  _TRTM: TTypesMeta;
}

interface MachineOptions<TEvent extends EventObject> {
  action?: ActionObject<TEvent>;
}

type MaybeTypegenMachineOptions<
  TEvent extends EventObject,
  TTypesMeta = TypegenDisabled,
> = TTypesMeta extends TypegenEnabled
  ? {
      action?: ActionObject<{ type: "WITH_TYPEGEN" }>;
    }
  : MachineOptions<TEvent>;

declare function assign<TEvent extends EventObject>(
  assignment: (ev: TEvent) => void,
): ActionObject<TEvent>;

declare function useMachine<
  TEvent extends EventObject,
  TTypesMeta extends TypegenEnabled,
>(
  getMachine: StateMachine<TEvent, TTypesMeta>,
  options: MaybeTypegenMachineOptions<TEvent, TTypesMeta>,
): void;
declare function useMachine<TEvent extends EventObject>(
  getMachine: StateMachine<TEvent>,
  options?: MachineOptions<TEvent>,
): void;

const machine = {} as StateMachine<{ type: "WITHOUT_TYPEGEN" }>;

const ret = useMachine(machine, {
  action: assign((_ev) => {
    ((_type: "WITHOUT_TYPEGEN") => null)(_ev.type);
  }),
});
```

</details>

fatcerberus:
I'm curious to see how badly this is going to impact performance.  It definitely *sounds* like a potential perf trap, at least for codebases with large quantities of overloads...

jakebailey:
https://typescript.visualstudio.com/TypeScript/_build/results?buildId=160185&view=results made mui time out; I'm going to up the max time limit of the benchmarker as apparently 60 _minutes_ was too slow, oops.

jakebailey:

weswigham:
@typescript-bot run dt

weswigham:
@typescript-bot pack this

weswigham:
@typescript-bot run dt
@typescript-bot test top100

weswigham:
Mmmm, vscode OOMing probably shouldn't report as "no difference" in the perf results with high probability. 😆 Definitely needs an indicator that the builds failed for some reason.

jakebailey:
> Mmmm, vscode OOMing probably shouldn't report as "no difference" in the perf results with high probability. 😆 Definitely needs an indicator that the builds failed for some reason.

It's low probability, p=1 which is the most bad. Happens when the perf numbers are zero. I agree that the table needs to say something about when the compilation fails (it just doesn't yet, only in logs).

RyanCavanaugh:
@typescript-bot pack this

weswigham:
Mmmm, this works, generally speaking (and there are big improvements in behavior!), but the perf cost is too high to do it this way over today's behavior (doubled checking times, much larger memory usage, ew), which feels quite bad. I'm not gonna keep this updated over time, and a future iteration is going to have to try an entirely different approach to try to get it to perform well, so I'm going to close this.

weswigham:
Note for future readers: AFAIK, it's the downlevel-decorator access-intermediated caches that are slow to touch, so maybe something closer to this becomes more viable once we have native decorators, OR, if you feel like rewriting every cache access in the codebase, that might work, too. Maybe.

Issue #57412, opened 2024-02-15
Private constructor type simplification produces incompatible type


JacobLey:
### 🔎 Search Terms

private constructor incompatible parameter types removed

### 🕗 Version & Regression Information

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about private constructor type simplification

### ⏯ Playground Link

[...]

PR #57428, opened 2024-02-17
Allow constructors of all visibility kinds to always be relatable to private constructors

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/57412


src/compiler/checker.ts
```diff
@@ -23423,8 +23423,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     }
                     return Ternary.False;
                 }
-                if (!constructorVisibilitiesAreCompatible(sourceSignatures[0], targetSignatures[0], reportErrors)) {
-                    return Ternary.False;
+                const result = relateConstructorVisibilities(sourceSignatures[0], targetSignatures[0], reportErrors)
+                if (result !== undefined) {
```
Andarist, line 23427: the idea is that if the target has a private constructor then it won't be able to call/utilize it so it doesn't matter what parameters the source might accept

src/compiler/checker.ts
```diff
@@ -23653,34 +23654,34 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             return Ternary.True;
         }
 
-        function constructorVisibilitiesAreCompatible(sourceSignature: Signature, targetSignature: Signature, reportErrors: boolean) {
+        function relateConstructorVisibilities(sourceSignature: Signature, targetSignature: Signature, reportErrors: boolean): Ternary | undefined {
```
Andarist, line 23657: Perhaps usage of `Ternary` is slightly questionable here but it's super convenient. This function now tries to give "definitive answer" to whether those signatures relate to each other or not, it's a short-circuiting of sorts. Usually, it won't be able to give such a definitive answer though and it will just return `undefined` so the caller will proceed with relating the parameters etc

fatcerberus:
If we're going to exclude private constructors from assignability checks, shouldn't we also exclude private members in general?  But then that's not sound because you might have

```ts
class Foo<T> {
    private foo: T;
}
```

and you don't want `Foo<number>` to be assignable to `Foo<string>`.  Feels like the same argument would apply to the constructor (I think?)

Andarist:
This doesn't change how `private` members affect assignability. I see your point but I always thought that `private` in TypeScript doesn't make classes nominal because of those private members themselves but rather because of how those members might be used by *other* members within a class and its subclasses. 

This is not about classical class members and the constructor is a bit special here. That being said, I would be equally fine with following the same assignability rules around private members with private constructors. This is inconsistent, both should be OK or errors:
```ts
class A {
  private constructor() {}
}
class B {
  private constructor() {}
}
class C {
  private constructor(arg: string) {}
}

const test1: typeof A = B; // ok
const test2: typeof A = C; // error
```

If we'd make this an error then we wouldn't address the OP's needs - there wouldn't be a way to express just "any constructable" in the type system. 

On top of that, the existing code comment in `constructorVisibilitiesAreCompatible` implies that it is desired to allow all constructor types to be relatable to private constructors. There are even tests for that (see [classConstructorAccessibility3](https://github.com/microsoft/TypeScript/blob/d04e3489b0d8e6bc9a8a9396a633632a5a467328/tests/cases/conformance/classes/constructorDeclarations/classConstructorAccessibility3.ts)). 

The reported problem is that this still enforces constructor parameters to be compatible and yet private constructors are stripped from their parameters in the generated declaration files. So another fix for this could be to tweak the declaration emit but I figured out that it would be worse since if the constructor is not callable by the receiver... then why would we even compare parameter types in the first place?

That being said, I have tweak this PR after your comment to not ignore the return types of such private constructors. Even if I don't quite see how one would obtain an instance of such private constructor :P

Andarist:
> Of course there, the cross-instance compatibility is less desirable...

Yeah, this is my line of thinking here. 

> and private static members would presumably already make the static-side of the class nominal

Ye, I checked that before putting this up - a private static member on the target exits the relation checking before we even get to comparing constructors.

Issue #57442, opened 2024-02-19
[ID-Prep] Explicit type annotation of const replaced with initializer in declaration emit


dragomirtitian:
_The [[ID-prep]](https://github.com/microsoft/TypeScript/issues?q=is%3Aissue+%5BID-prep%5D+in%3Atitle+) set of issues aligns Declaration Emit with the forthcoming [Isolated Declarations](https://github.com/microsoft/TypeScript/issues/47947) feature._

### 🔎 Search Terms

declarations type annotation const not preserved

### 🕗 Version & Regression Information

- This is the behavior in every version I tried

### ⏯ Playground Link

[Playground Link](https://www.typescriptlang.org/play?&code/MYewdgzgLgBAHgZhgXhgWQIZQBYDoBOGYAJiALYAUAlDAPwwBEGARsAzAFyPECmAZgwDcMGAHpRjFmxgAfbvwYBYAFBA=#code/KYOwrgtgBAolDeUBEBnA9hYUBuBDANmMElALxQCMUAvgNwCwAUAMZogoAuUwAXFBwE8ADsDQAzWAG1UGLHkLEAumSkzMOAkSSLaQA)

### 💻 Code

```ts
enum E { "some value" = 1 };
const e: typeof E["some value"] = E["some value"];
```

### 🙁 Actual behavior

`e` is emitted in declarations as `declare const e = E["some value"];`

### 🙂 Expected behavior

`e` should have in declarations the type that was manually declared

Issue #57441, opened 2024-02-19
Type of variable is widened in declaration with !strictNullChecks


dragomirtitian:
### 🔎 Search Terms

const declaration widen

### 🕗 Version & Regression Information

- This is the behavior in every version I tried

### ⏯ Playground Link

[Playground Link](https://www.typescriptlang.org/play?&code/MYewdgzgLgBAHgZhgXhgWQIZQBYDoBOGYAJiALYAUAlDAPwwBEGARsAzAFyPECmAZgwDcMGAHpRjFmxgAfbvwYBYAFBA)

### 💻 Code

```ts
declare function f3<T>(a: T, b: T): T;
const x3 = f3("abc", "def");  // "abc" | "def"
```

### 🙁 Actual behavior

`x3` is emitted as `declare const x3: string;` in declaration files but it's actually `"abc" | "def"` in source

### 🙂 Expected behavior

`x3` is emitted as `declare const x3: "abc" | "def"`

fatcerberus:
I think this behavior might be connected with automatic literal widening because
[Playground](https://www.typescriptlang.org/play?&code/MYewdgzgLgBAHgZhgXhgWQIZQBYDoBOGYAJiALYAUAlDAPwwBEGARsAzAFyPECmAZgwDcMGAHpRjFmxgAfbvwYBYAFBA=#code/CYUwxgNghgTiAEAzArgOzAFwJYHtVIGYAeAFQD4AKKALnhIBp4AjWkgSlYG4AoMPAZwzwAHgBYA+gEZ4AXkIUARFCZgFjBaEQK2nePAD0++EpUL4AH2OaFvAULHiATLPknV8KP3h9Ug9dY8vH0EdPUNjZXdLDRAtbggQewBWKRcHSR5w+AA9AH54xJEU5zkHR0yjHPygA)

```ts
declare function f3<T>(a: T, b: T): T;
const x4_1 = f3("abc", "def");  // "abc" | "def"
const x4_2 = f3("abc" as const, "def" as const);  // "abc" | "def"
let x5_1 = x4_1;
//  ^? string
let x5_2 = x4_2;
//  ^? "abc" | "def"
```

btw your playground link in the OP doesn't seem to work properly - for me it just loads whatever I had last entered into the playground.

PR #57444, opened 2024-02-19
Simplify declaration emit for literal types.

### LINKED ISSUE #57439: [ID-Prep] Preserve all triple slash directives in declarations files

dragomirtitian:
Fixes #57439, #57441, #57442


src/compiler/checker.ts
```diff
@@ -48387,8 +48387,11 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         // Get type of the symbol if this is the valid symbol otherwise get type at location
         const symbol = getSymbolOfDeclaration(declaration);
         let type = symbol && !(symbol.flags & (SymbolFlags.TypeLiteral | SymbolFlags.Signature))
-            ? getWidenedLiteralType(getTypeOfSymbol(symbol))
+            ? getTypeOfSymbol(symbol)
             : errorType;
+        if(!isReadonlySymbol(symbol)) {
```
dragomirtitian, line 48392: The lack of this test is the reason for #57441

src/compiler/checker.ts
```diff
@@ -48518,29 +48521,6 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         return undefined;
     }
 
-    function isLiteralConstDeclaration(node: VariableDeclaration | PropertyDeclaration | PropertySignature | ParameterDeclaration): boolean {
```
dragomirtitian, line 48521: No longer needed since we only use the syntactic information to determine if we want to preserve the initializer.

src/compiler/transformers/declarations.ts
```diff
@@ -707,13 +712,16 @@ export function transformDeclarations(context: TransformationContext) {
...
+        if(canHaveLiteralInitializer(node) && node.initializer && !node.type && isPrimitiveLiteralValue(node.initializer)) {
+            return isDeclarationReadonly(node) || isVariableDeclaration(node) && isVarConstLike(node);
+        }
+        return false;
     }
 
     function ensureNoInitializer(node: CanHaveLiteralInitializer) {
         if (shouldPrintWithInitializer(node)) {
-            return resolver.createLiteralConstValue(getParseTreeNode(node) as CanHaveLiteralInitializer, symbolTracker); // TODO: Make safe
+            return clonePrimitiveLiteralValue(node.initializer)
```
dragomirtitian, line 724: We clone the primitive in order to remove any source specific features. Such as `_` or hex or octal or usage of `` ` `` or `'` with strings. We could just use the original node, but if there is a new way to write a primitive literal it might not be backward compatible with older versions of TypeScript. If we don't think that is a problem, I would prefer to use the original node with its original format.

weswigham (changes requested):
This introduces bugs in incremental compilation - by not preserving the widening behavior of the literals (wherever possible), by creating an initializer whenever the type at the location is a widening type, you can create declaration files where, when you check against the declaration files, you witness the literal type, while when you check against the source file, you witness the widening type (and thus, the base type). We have open issues about other types where we don't correctly do this today (eg, for tagged templates), and this would make the problem hugely worse.

This *cannot* be a baseline change in the compiler. This absolutely has to be flagged off, if we take it at all. There is a meaningful difference between a `= 1` and `: 1` on a variable declaration - that some emit mode can't track the widening behavior of literals across declarations does not mean we should introduce buggy behavior for all users.

dragomirtitian:
@weswigham  Sorry it took a while to get back to this. 

I changed the goal to just preserve the initializer node if possible.  This should not change the meaning of the declaration. 

It should also be a safe change to make. Even older versions of TypeScript will understand all of the primitive literal syntax, even if it was previously not present in declarations.

Issue #57435, opened 2024-02-18
`Debug Failure. Expected [object Object] === [object Object]. Parameter symbol already has a cached type which differs from newly assigned type` on TS 5.3


marekdedic:
### 🔎 Search Terms

`Debug Failure. Expected [object Object] === [object Object]. Parameter symbol already has a cached type which differs from newly assigned type`

### 🕗 Version & Regression Information

- This is a crash
- This is a duplicate of #50773 - as per the instructions there, it is re-filed with a new TS version repro

### ⏯ Playground Link

https://github.com/skaut/wordpress-version-checker/tree/typescript-chached-type-issue

### 💻 Code

See https://github.com/skaut/wordpress-version-checker/tree/typescript-chached-type-issue

Run `npm ci && npm test`

Sorry for not providing a playground link, but I could only replicate this with Jest involved

### 🙁 Actual behavior

```
 FAIL  __tests__/index.test.ts
  ● Test suite failed to run

    Debug Failure. Expected [object Object] === [object Object]. Parameter symbol already has a cached type which differs from newly assigned type

    > 1 | import * as github from "@actions/github";
        | ^
      2 |
      3 | export async function testFn(): Promise<boolean> {
      4 |   const octokit = github.getOctokit("");

      at assignParameterType (node_modules/typescript/lib/typescript.js:75417:15)
      at assignContextualParameterTypes (node_modules/typescript/lib/typescript.js:75383:11)
      at contextuallyCheckFunctionExpressionOrObjectLiteralMethod (node_modules/typescript/lib/typescript.js:76107:15)
      at checkFunctionExpressionOrObjectLiteralMethod (node_modules/typescript/lib/typescript.js:76082:7)
      at checkExpressionWorker (node_modules/typescript/lib/typescript.js:77864:18)
      at checkExpression (node_modules/typescript/lib/typescript.js:77766:34)
      at maybeCheckExpression (node_modules/typescript/lib/typescript.js:76852:30)
      at BinaryExpressionStateMachine.onRight (node_modules/typescript/lib/typescript.js:76816:18)
      at Array.right (node_modules/typescript/lib/typescript.js:27972:36)
      at trampoline (node_modules/typescript/lib/typescript.js:27738:44)
      at node_modules/typescript/lib/typescript.js:76746:24
      at checkExpressionWorker (node_modules/typescript/lib/typescript.js:77889:18)
      at checkExpression (node_modules/typescript/lib/typescript.js:77766:34)
[...]

PR #57458, opened 2024-02-21
Fixed crash related to JS synthethic rest param and preceeding parameters with initializers

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/57435


tests/cases/compiler/argumentsReferenceAndParameterWithInitializer1.ts
Andarist, line 1: I have quite a bit of different variations on tuple types used as rest here. They don't test anything specific to the referenced issues (beyond what the first 2 cases test already). I was stress-testing this and trying to trigger the report of `A_required_parameter_cannot_follow_an_optional_parameter` in some specific combinations. That led me to creating: https://github.com/microsoft/TypeScript/issues/57565 

I think it still doesn't hurt to leave all of those different cases here.

tests/cases/compiler/argumentsReferenceAndParameterWithInitializer1.ts
```diff
@@ -0,0 +1,131 @@
...
+  function structuredClone (value, options = undefined) {
+    if (arguments.length === 0) {
+      throw new TypeError('missing argument')
+    }
+    return value;
+  }
+
+/** @type {(a: number, b: boolean | undefined, ...rest: string[]) => void} */
+const test1 = function(value, options = undefined) {
+  if (arguments.length === 0) {
```
Andarist, line 24: This is the only reference to `arguments` here. I think that perhaps this pattern (checking `arguments.length`) wouldn't have to make `containsArgumentsReference` to return `true`. I'm not sure if there is an appetite for this though.

src/compiler/checker.ts
```diff
@@ -12057,6 +12057,15 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         if (!links.type) {
             Debug.assertIsDefined(links.deferralParent);
             Debug.assertIsDefined(links.deferralConstituents);
+            if (links.deferralParent === neverType) {
+                const functionDecl = links.jsSyntheticRestParameterFunctionDeclaration!;
+                if (isFunctionExpressionOrArrowFunction(functionDecl)) {
+                    const contextualSignature = getContextualSignature(functionDecl);
+                    if (contextualSignature) {
+                        return getRestTypeAtPosition(contextualSignature, functionDecl.parameters.length);
```
andrewbranch, line 12065: Is it intentional that this isn’t assigned to `links.type`?

src/compiler/checker.ts
```diff
@@ -12057,6 +12057,15 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         if (!links.type) {
             Debug.assertIsDefined(links.deferralParent);
             Debug.assertIsDefined(links.deferralConstituents);
+            if (links.deferralParent === neverType) {
+                const functionDecl = links.jsSyntheticRestParameterFunctionDeclaration!;
+                if (isFunctionExpressionOrArrowFunction(functionDecl)) {
```
andrewbranch, line 12062: It kind of looks like you could check just `links.jsSyntheticRestParameterFunctionDeclaration`. Is the `neverType` actually relevant here?

Andarist:
@andrewbranch this issue has already been fixed by https://github.com/microsoft/TypeScript/pull/57580 (it doesn't come with the tests contained by this PR here - you can check them out in the current nightly though: [TS playground](https://www.typescriptlang.org/play?ts=5.5.0-dev.20240314&filetype=js#code/OQVwzgpgBGAuBOBLAxrYBuAUJg9DqAFrLAA5gBceA5orASAEYB0yA9gLY7srytisAzWDgAqATxIQAysiQlhiMGBAQwOAKwB2ACwBmddhwAqI1AACsCdADeVADasGAQzsiCigNrA48EKhDwEAAmAMIOAHYQwAC6AL5QRjiYbOFwMAh+sAHBYayRUAC8mFBQ9o4ubopMPpnZoRHQAPyNxVACIOGoiHnpvv6B9XnQABQAbi4qADRQrPLdqYVQHUEQAoiRQQCUUNatJYgCUMNO8FQg7BDhsGBMdpdUdIUFBVAADNu7JV9QdLwA7lBIgDxJIAKLwXjwYbAbhKdZUKAnM4XK7ATZ7KCxDGBLLwcJQcZ2FRYEpY3AmcyWSQ7Y7kQHnBgQeDTBh0hisVh3Jz4gA+S3CKzWG2mTFFgTgdJ88I80W2BQAfATWIggvFEsk8mlYKpYABGRbtTqweZjCYQaazY2axbLVbrYIfVoHI5I86Xa63e6PZ4vd47DG-VgAoFQEEQcGQ6GwsDwxGnN2o9GkzBk4ymCxWGlOOnhBlMllsjlc3n8wX2oIisU6yUIaWywqK0bK1UJJIpLU6gBMBo6XR660iUMJUxmc2tL1tQod-pK7dgcaoYEWMuqdhQEBYLjsx3jKOumywTvCg+GuumGXNUGABAgdgcwGmwD+rHgdiCaKwqYpGep1lp9PYRlmSgVkQKLCBuSgPlJ3LOVG2bNU201edtTgXQeyNeYoAHJlTSJS9LXmJcJwFO0NkdWdkIXYioBXMA12QDdkC3HdkXdMADxTQxvypGx-1zQD8xAwtOQgksYOFKBRSYcVYDpDwpXCKhGkrJgBKAmV6wVJUVUQjVUhQnVtAwvt8RwoczQtMcFhIstyJnKA52o5dolXddNzvViE33T9uPTXisxzPNgNA9lRMg6DSKnCspKrCVaMUqhVPUplNLgnSW3VJzUNgdQTKtMzj1w4cCOsmiJOnT5HKopEaLohimJY109w43zyX8zM-2zACgILMDwvEqLy1U2T5MS5LgrShsMr07KdQANnyrDzLwkdCPHUsyOCc93HgLYHKc2qXLcxiPO3Zr2M4r8Ot-fjgr6sLiygzbopG6setSrT4N01t9I7OBNCW-sios-CrIK8qhskuhED2ijqoM5yXnq9zmM8i6rlarj2spTq7sEkLGhE4s3vixKZXSpsfqyqicoADiBwqTxK8GiJtKHKtaQ7Tjq1z6NRprd0utq01x27upSwnibE0m5N6OtKYQ365rgABORnsJB1bSoh9m7O2n5dv2qrucXY7+dOtHzqFzGrr8sW+Il+6QKJ-qSdimT3oU2slImgmpu0qnMqQxGct1V4NZWlnR112ytuNrmap582GrOryWrtnGf0doKCZZV3Hplj3RoSn3lL9jToi+mbldpnVdX1F5DVMzXmcsmO2bj6L4dN3mTsa9GbZ87HRezwKPqlt2i+kkvvaQX2PclgPvuDv7DLgXVuyb3sCtb4r2-WmyXuGw3YYTyjEaO5G+dTq30+FkeeLxp285d6XuVlsay5UxfJqrxXqYh3+nqdC29MLAzbmDDuG0KoxRhnDA6SczbX37mnDGw9roO3HpLfO79wif1LvPcu8slKNGrkHWadcN7GTAS3KOB8yp63jj3JBfcLYD2tmxW2Isn7i1zr1N+U8P7Fy9olH+YjyFKxpqHeueVaG73oVAw+kN9ZwKNiwy+ycUHsLQUPLGmCx5dX4UJBgBdwLCJnqIsu0xEqSMAWvH49dFryOWlraOyimHd0QZo5BtEb4C0HlwjB9tDH4wEaYvBBC57whsWXOxq8VZ6kBi4iB+8lGMK7ifeB58EZpCvn41Bd90H6OwIk3UDMUlMzSWtDJx97Im1YSnAJnDvIlIcWHdWlS96gxqbHOpBtskaLyVogpOiil6MzmmYo6YSAnCcOwHYkt4geAYNEaZSRm67xyp2V4xwWSLHDtMZAQz5z5JRpbQWQS2mjwCkYieuChH4JEfFaJC9bEANXpsrC2zdR7JAosC8RyTlI1GbfS5rTM4QAAB4kBfPOawsQgA)).

So the question is: is the fix here still worth pursuing? I think it might be because the other PR fixed just one of the call hierarchies leading to this problem. I now don't have any other test case at hand to prove the necessity of this PR here though

src/compiler/checker.ts
Andarist: It's intentional in the sense that I followed what some other, similar, situation was already doing. At the time of writing this, I knew that I'm not assigning to `links.type` and I've decided not to do it... but now I'm not 100% sure why I didn't do it.

Taking a *quick* look at related functions I might have been trying to replicate he intention of the comment in `getTypeOfVariableOrParameterOrProperty`. OTOH, that still assigns `links.type` - just conditionally. But yeee, I think it was to make sure that the type set by `assignParameterType` prevails.

src/compiler/checker.ts
Andarist: Yeah, you are right - this is just checking the same thing really and I can simplify it.

src/compiler/checker.ts
andrewbranch: `getTypeOfVariableOrParameterOrProperty` only assigns conditionally because there is a `checkMode`, and we only cache types computed with `CheckMode.Normal`. This function appears to be a pure function of `symbol` and so AFAIK its type can/should be cached to prevent repeating work next time it’s asked for.

andrewbranch:
It sounds like this fix may still be worth pursuing, but a failing test case probably does need to be found or reverse engineered.

Issue #57460, opened 2024-02-21
Untyped JS module is not assignable to type even though it satisfies it


bradzacher:
### 🔎 Search Terms

Untyped JS module is not assignable to type even though it satisfies it, "Object literal may only specify known properties"

### ⏯ Playground Link

https://github.com/bradzacher/bug-repros/tree/ts-js-issue-huh

### 💻 Code

```ts
import eslintReact from 'eslint-plugin-react/configs/recommended.js';

(eslintReact satisfies {
  plugins: {
      react: {
          rules: any;
      };
  };
});
```

### 🙁 Actual behavior

There should be no errors because the type
`{ plugins: { react: { deprecatedRules: any; rules: any; }; }; }`
is assignable to
`{ plugins: { react: { rules: any; }; }; }`

### 🙂 Expected behavior

```
Type '{ plugins: { react: { deprecatedRules: any; rules: any; }; }; rules: any; readonly languageOptions: any; } & { languageOptions: any; rules: { 'react/display-name': number; 'react/jsx-key': number; 'react/jsx-no-comment-textnodes': number; ... 18 more ...; 'react/require-render-return': number; }; }' does not satisfy the expected type '{ plugins: { react: { rules: any; }; }; }'.
  The types of 'plugins.react' are incompatible between these types.
    Object literal may only specify known properties, and 'deprecatedRules' does not exist in type '{ rules: any; }'
```

### Additional information about the issue

This issue is particularly weird because it only occurs in a JS file!

For more context this is the original bug report https://github.com/typescript-eslint/typescript-eslint/issues/8522
And the original repro from the issue (the usecase we really want to fix): https://github.com/printfn/tseslint-react-repro

We have built support for ESLint flat configs via our `typescript-eslint` package and it exports a function to allow typechecking of the config array. So the repo looks like this

```js
// @filename eslint.config.js


// @ts-check

import tseslint from 'typescript-eslint';
import eslintReact from 'eslint-plugin-react/configs/recommended.js';

export default tseslint.config(
	eslintReact, // type error is on this line
);
```

Best I can tell it looks like TS is treating the `eslintReact` variable the same way it would as an inline object expression - hence it's erroring saying "too many properties".

Andarist:
This is a private repo or something: https://github.com/bradzacher/bug-repros/tree/ts-js-issue-huh

bradzacher:
oops sorry made it public

Andarist:
[...]

PR #57467, opened 2024-02-21
Propagate object flags related to object and array literals onto `module.exports` types

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/57460


gabritto:
@typescript-bot test it

gabritto (changes requested):
(1) I don't exactly understand how the code change is related to the observed behavior in the repro, if you could provide an explanation that would be very helpful.
(2) There were test changes that I also don't know how they relate to the code change, so would be nice to get an explanation for those too (even if they're bugs). Thanks!

Andarist:
1. IIRC those flags affect `isPerformingExcessPropertyChecks` 
2. the in-repo change looks like being related to [type inference for object literals](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html#improved-type-inference-for-object-literals). It looks like a correct change to me - it now benefits from this since we are correctly propagating those literal flags. As to the `webpack`'s change from [this comment](https://github.com/microsoft/TypeScript/pull/57467#issuecomment-2040374361) - I'm not sure yet. I just synced with `main`, could we rerun those tests to get a confirmation that nothing wonky is happening with that report (I'm not sure how likely that is with the current infra)? If it comes back with this error again then I'll happily investigate this (cc @jakebailey )

gabritto:
According to Jake, the extended tests use the latest main, so the not-synced PR shouldn't be a problem, but I can try running again.

gabritto:
@typescript-bot test it

jakebailey:
My impression at the moment is that the CFA change introduced some non-determinism; that same error went away in that change and is now back again.

Andarist:
Something weird is happening with those tests but I think it's fair to say that it's not related to this PR 😉

gabritto (approved):

jakebailey:
This PR was out of date and broke main.

gabritto:
> This PR was out of date and broke main.

I'll fix it now

PR #57472, opened 2024-02-21
Remove code related to prepend and fix baselines

sheetalkamat:
- Removes all the APIs and call related prepend and corresponding tests
- Fixed the tests that didnt depend on prepend
- Also fixed issue with tsserver not baselining the version as FakeMajor.FakeMinor when the callbacks were baselined by passing around the logs


sheetalkamat:
# Package size report

## Overall package size

| - | - | - | - | - |
| Packed | 5.55 MiB | 5.53 MiB | -26.18 KiB | -0.46% |
| Unpacked | 30.86 MiB | 30.72 MiB | -143.91 KiB | -0.46% |

## Files

| - | - | - | - | - |
| `lib/tsc.js` | 5.62 MiB | 5.58 MiB | -43.71 KiB | -0.76% |
| `lib/tsserver.js` | 8.24 MiB | 8.19 MiB | -45.31 KiB | -0.54% |
| `lib/typescript.d.ts` | 584.73 KiB | 580.04 KiB | -4.69 KiB | -0.80% |
| `lib/typescript.js` | 8.72 MiB | 8.67 MiB | -47.20 KiB | -0.53% |
| `lib/typingsInstaller.js` | 1.68 MiB | 1.68 MiB | -3.01 KiB | -0.17% |

src/compiler/types.ts
```diff
@@ -7098,7 +7008,10 @@ export interface ProjectReference {
     path: string;
     /** The path as the user originally wrote it */
     originalPath?: string;
-    /** True if the output of this reference should be prepended to the output of this project. Only valid for --outFile compilations */
+    /**
+     * Only present for deprecate notification
```
jakebailey, line 7012: I _think_ we can declaration merge from the deprecation helper, if that's what you mean. (See also how `src/services/types.ts` extends the compiler.)

sheetalkamat: yeah but what would it solve. we still need to check "ref.prepend" in program.ts and utilities to compare and issue errors?

jakebailey: Oh, gotcha.

DanielRosenwasser:
I feel a bit mixed - it's a little hard to read it casually (thanks GitHub), but what I want to know is:

* will users still get an error message explaining why `--prepend` no longer works?
* will the language service stop functioning correctly on projects using `--prepend`?

If the language service stops working, it means we'll be unable to browse the code of TypeScript 4.9 and earlier with TypeScript 5.5+.

jakebailey:
> If the language service stops working, it means we'll be unable to browse the code of TypeScript 4.9 and earlier with TypeScript 5.5+.

IIRC `--prepend` only affected emit, so I would think that everything would still be browsable, but I have not checked myself.

sheetalkamat:
@jakebailey said it correctly.. this affects emit and it shouldnt impact the LS at all. 
We do error that prepend is not supported but thats all. the actual functionality of prepending is removed now.

src/compiler/types.ts
```diff
@@ -7098,7 +7008,10 @@ export interface ProjectReference {
     path: string;
     /** The path as the user originally wrote it */
     originalPath?: string;
-    /** True if the output of this reference should be prepended to the output of this project. Only valid for --outFile compilations */
+    /**
+     * Only present for deprecate notification
+     * @internal
+     */
     prepend?: boolean;
```
jakebailey, line 7015: In #57527 I just marked "removed" things as deprecated but left them in, which has the benefit of letting people still compile (and us getting a strikethrough), and then in 6.0 we could just delete them wholesale. I feel like that's safer, but I'm not 100% certain we've actually talked about how we want to do this sort of thing.

src/compiler/types.ts
```diff
@@ -8882,13 +8793,6 @@ export interface NodeFactory {
 
     /** @internal */ createRedirectedSourceFile(redirectInfo: RedirectInfo): SourceFile;
 
-    /** @deprecated @internal */ createUnparsedSource(prologues: readonly UnparsedPrologue[], syntheticReferences: readonly UnparsedSyntheticReference[] | undefined, texts: readonly UnparsedSourceText[]): UnparsedSource;
```
jakebailey, line 8885: So happy to see this stuff disappear 😄

src/compiler/types.ts
sheetalkamat: They have been getting strike through for since 5.0 no ?

sheetalkamat: oh this one we forgot to deprecate before so makes sense to mark deprecate.

src/compiler/types.ts
sheetalkamat: This and bunch of bundle manipulation to make sure emit helpers and stuff like that was ordered correctly.

src/compiler/types.ts
jakebailey: Yeah, we didn't do that :(

src/compiler/commandLineParser.ts
```diff
@@ -813,6 +813,9 @@ const commandOptionsWithoutBuild: CommandLineOption[] = [
     {
         name: "verbatimModuleSyntax",
         type: "boolean",
+        affectsEmit: true,
```
RyanCavanaugh, line 816: Does this belong in its own PR? Looks unrelated

jakebailey (approved):

src/compiler/commandLineParser.ts
sheetalkamat: This was related to baseline i fixed so added it here.

sheetalkamat: As part of 1012104 and c871a27

jdalton:
@RyanCavanaugh Thank you! Is there a suggested alternative API to `createUnparsedSourceFile`?

RyanCavanaugh:
The only (intended) legal thing to do with `createUnparsedSourceFile` was to put it into a `Bundle`, which no longer accepts that. I would just use a `string` instead if you were using it to ferry around unparsed sources.

Issue #57301, opened 2024-02-05
Typescript autocomplete: extending an abstract class's async method get autocompleted incorrectly


JorensM:
Does this issue occur when all extensions are disabled?: Yes

- VS Code Version: 1.85.2
- OS Version: Windows 10

https://github.com/microsoft/vscode/assets/17122123/9cf9a4d3-130f-4fef-8aaa-a3e336583ca8

Steps to Reproduce:

1.  Create an abstract class with an abstract async method ( returns a Promise)
2.  Create another class that extends this abstract class, and try to write the method using autocomplete, by starting with typing `async`
3.  See that it gets autocompleted incorrectly

mjbvz:
Copiable code: 

```ts
abstract class AFoo {
    abstract bar(): Promise<void>;
}

class Foo extends AFoo {
    async b
}
```

mjbvz:
Here is the completion entry:

```
        {
            "name": "bar",
            "kind": "method",
            "kindModifiers": "abstract",
            "sortText": "11",
            "insertText": "async bar(): Promise<void> {\n    $0\n}",
            "filterText": "bar",
            "isSnippet": true
        },
```

No replacement range is returned

PR #52525, opened 2023-01-30
Improve class member snippet completions and keyword completions interaction

gabritto:
Fixes #52121.

This PR changes how class member snippet completions are sorted and filtered, and how they interact with class member modifiers.
The problem, as presented in the linked issue, is that class member snippet completions are currently sorted *after* keyword completions (via `sortText`), and this is not ideal when you manually request completions inside a class without having typed anything (in other words, when you'd likely want to see the overridable class members first). However, this was done so that member completions that have modifiers do not show up before the modifier keywords when you are typing the modifier.

With this PR, class member completions are sorted *before* keywords. But a class member `protected foo()` will have a filter text of `foo`, meaning that, as you're typing, that member completion will only show up if you're typing a substring of `foo` (i.e. the editor will match what you're typing to `foo`, whereas previously it would match what you're typing to the full insert text), so the member completion won't show up as you're typing `protected`.

I think in general this approach roughly matches what I've seen in other editors/languages (VS+C#, WebStorm+TS, xcode+Swift), and it makes sense to how I'd use it, but I'm looking for feedback on the UX and whether this is the best approach. 

*Case not supported:*
- `async` modifier is not inserted/allowed. This has been the case since the initial implementation of class member snippet completions. It's not clear to me if/when we should insert `async` or allow it to be inserted, open for opinions on that. I vaguely recall that, regarding object method completions, the sentiment around `async` was that we shouldn't insert it and people could add it after accepting the completion if they wanted.

*Not ideal:*
- if one of the class member completions has a class member modifier as its prefix (e.g. a property called `publicProperty`), it will be sorted *before* the modifier keyword. I'm hoping that's not common enough to be disruptive, but I might be wrong.
- if you type modifiers in the wrong order (or the wrong modifiers) for a certain class member, the class member completion will not be displayed. Again, don't know how disruptive that would be, because if your intent is to override a certain class member, you could start by typing its name (or part of its name), instead of starting by typing the member's modifiers.

**How to test this in vscode:**
- Use the packaged version of TS corresponding to this PR: https://github.com/microsoft/TypeScript/pull/52525#issuecomment-1409483281
- Use the vscode version from my [fork](https://github.com/gabritto/vscode-1) or [this PR](https://github.com/microsoft/vscode/pull/172866), otherwise the completions won't work properly because a completion entry's `filterText` is currently modified by vscode and not respected.

PR #57480, opened 2024-02-22
Include `replacementSpan` in member completions

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/57301


tests/cases/fourslash/completionsOverridingMethod6.ts
Andarist, line 1: I have confirmed manually that those were broken in the very same way as the referenced issue - the existing modifiers were not removed and since the `insertText` had them the new source text of the file was broken because they got duplicated

iisaduan:
@typescript-bot pack this

iisaduan (changes requested):
Thanks for finding the other cases that break in the same way! 

However, I don't think this is the correct fix, as when I try to use this version in the editor, it completely removes the completions from (all) the marked locations. 
image

I did by chance happen to be in an old version of TS where the first test case worked, so with some quick checking, the bug seems to be a regression bug that happened between 5.1.6 and 5.2.2.

Andarist:
Interesting, I'll take another look at this soon.

Andarist:
Thanks for testing this out in VS Code. I made the mistake of assuming that this would just work - the local tests here looked alright and this exact thing (`replacementSpan` inclusion) was mentioned as a fix for this by a VS Code team member ([here](https://github.com/microsoft/TypeScript/issues/57301#issuecomment-1927332771))

However, it turns out... that it can't work today in this form. VS Code assumes certain things about the replace ranges. So while this technically could be a fine change in TS... it just doesn't work well when integrating with VS Code. 

For starters, replace ranges **can't** span multiple lines (see the check [here](https://github.com/microsoft/vscode/blob/1f94e5cd54ce0a7bc503a3f95a3742ddc5980151/src/vs/editor/contrib/suggest/browser/suggest.ts#L115-L117)). The range computed by `getPresentModifiers` includes the leading trivia and that would usually contain the leading whitespace and the newline character after the previous member. That's an easy fix - we can just select a smaller range by using `.getStart()` over `.po`.

But that's still not good. VS Code also assumes that the replacement range contains part of the "word" that is meant to be replaced. Since the target word here is `bar` we can't get away with having the replace range to correspond to `async b`. In such a case, VS Code filters out our completion entry [here](https://github.com/microsoft/vscode/blob/1f94e5cd54ce0a7bc503a3f95a3742ddc5980151/src/vs/editor/contrib/suggest/browser/completionModel.ts#L218-L222)

I could fight this by adjusting the `label` or `filterText` but none of that looks particularly good to me.

So my next bet was to use code actions to do this. I pushed out draft changes doing that (without adjusting the test cases, I hope that it's not the problem at this stage) so you could test it out. It's far from perfect right now. The `eraseRange` still corresponds to `async b` and then when the code action removes that range it removes those extra 2 characters (one whitespace and `b`) from the final result (and not just the old modifiers). This could be adjusted - no problem. But I still find the experience here to not be the best:
1. the code actions is a separate action from the original insertion - this sometimes is observable as it takes a moment to apply the follow up code action
2. undoing/redoing doesn't work like I'd expect it to. We end up with 2 items on the stack and it feels like we should only have 1

I either didn't find the correct combination of things to do this correctly today or there is some API gap here that would have to be filled. I somewhat doubt that VS Code would be able to change their assumptions about replacement ranges right now (no support for multiline and treating that whole range as part of the target word).

Andarist:
I have implemented this using `filterText` and - from what I can tell - it works with VS Code and all. I feel like I'm abusing this API though and that it wasn't designed for this.

iisaduan:
@typescript-bot pack this

gabritto:
> I have implemented this using `filterText` and - from what I can tell - it works with VS Code and all. I feel like I'm abusing this API though and that it wasn't designed for this.

From what I remember, if we include modifiers in the `filterText`, we'll run into the same problem that #52525 was addressing, in that now if you're typing that modifier, you'll get the method completion sorted before the keyword completion. I think the `filterText` is just the name of the method on purpose, to avoid the method completion from showing up as you type a modifier.

The other course of action I can think of to fix this is to omit existing modifiers from the completion, but now if the modifier order is wrong or you need to add a modifier *before* the existing one, you can't.

It really does look like what we need is to use a `replacementSpan` to replace the existing modifiers. Maybe I should check with vscode again if that's at all possible.

Andarist:
> It really does look like what we need is to use a replacementSpan to replace the existing modifiers. Maybe I should check with vscode again if that's at all possible.

That would be the most straightforward thing to do here - on this end at least :p So I very much agree that if that's the possibility then it would be great to explore it.

iisaduan:
We discussed offline, and our current plan is to accept #57643 with the code action, leave `filterText` alone, and in the future, we may return to work on this with vscode to find a solution that doesn't require code actions/multiple user steps to undo. 

Thank you for such a thorough investigation!

PR #57484, opened 2024-02-22
Preserve parameter types for optional parameters /fields  with undefined in type and for required params with default value

### LINKED ISSUE #57483: [ID-Prep] Parameter types are not preserved.
### 🔎 Search Terms

optional parameter types not preserved

### ⏯ Playground Link

[Playground Link](https://www.typescriptlang.org/play?#code/C4TwDgpgBAsghmKBeKBvAvlAZGqBtABSgEsA7KAZ2ACcyBzAXQC4o5SQp0BYAKFElgIA8tQCqpACYQAZmQgTkgxAB8oAV0ky5C1QCIJagLaGQu3hAAeYAPbVgUAMYAbOBQpQAYtetpeUR9akVNRqDsC2ABR+-lAA9LFQACrg0KTWUiTucA4OatRwwBAANFBg1BAUENQAbhDR-mBqAEZOxA5Q5WqVyZAAcukQAPws8CrqmrKk8kX1pc2t7Z3dKf1SAEzDUACiFs5qUgA8oyLiUpPTUPpGJroAfDM8MXEJAEoV1k61UPypAyXlAEc1MRylkJBJ6FBVBoztpZo0Wm0Ou9PhAekMRggHv4AJS4bg8AnmKy2ezSDRhYiBKDSbwARiijxi8SSKSgaQyxCyOTyBWKpVBVVqsyWaJWA0xYxhWimChQGGxMVF6NWEDWLB2e0OxzEE20JSuxlMt388vQiv8LLeFA+Xx+7L+yKBIIqrHBkOhetlIpRtXRksUCqgPud5QkBDg+UMLFIRiaVTxqF4vHQQA)

### 💻 Code

```ts
type Map = {} & { [P in string]: any }
type MapOrUndefined = Map | undefined | "dummy"
export class Foo {
  constructor(
    // Type node is accurate, preserve
    public reuseTypeNode?: Map | undefined,
    public reuseTypeNode2?: Exclude<MapOrUndefined, "dummy">,
    // Resolve type node, requires adding | undefined
    public resolveType?: Map,
  ) { }
}

export function foo1(
    // Type node is accurate, preserve
    reuseTypeNode: Map | undefined = {},
    reuseTypeNode2: Exclude<MapOrUndefined, "dummy">  = {},
    // Resolve type node, requires adding | undefined
    resolveType: Map = {}, 
    requiredParam: number) {

}
```

### 🙁 Actual behavior

Parameters and fields named `reuseTypeNode*` have their types re-printed (and mangled) by the declaration transform even though the re-printed type is semantically equivalent to the source type.

### 🙂 Expected behavior

Parameters and fields named `reuseTypeNode*` should have their types preserved as they are written in source code

_No response_

dragomirtitian:
Fixes #57483


src/compiler/transformers/declarations.ts
```diff
@@ -741,46 +739,43 @@ export function transformDeclarations(context: TransformationContext) {
...
-                resolver.isOptionalUninitializedParameterProperty(node));
-        if (type && !shouldUseResolverType) {
+        const shouldAddImplicitUndefined = node.kind === SyntaxKind.Parameter && resolver.requiresAddingImplicitUndefined(node);
+        if (type && !shouldAddImplicitUndefined) {
             return visitNode(type, visitDeclarationSubtree, isTypeNode);
         }
-        if (!getParseTreeNode(node)) {
-            return type ? visitNode(type, visitDeclarationSubtree, isTypeNode) : factory.createKeywordTypeNode(SyntaxKind.AnyKeyword);
-        }
-        if (node.kind === SyntaxKind.SetAccessor) {
```
dragomirtitian, line 753: `ensureType` is never called with a `SetAccessorDeclaration`. The set accessor type does not have a return type and the parameter type is inferred independently. Removing this had 0 impact on tests.

src/compiler/transformers/declarations.ts
```diff
@@ -741,46 +739,43 @@ export function transformDeclarations(context: TransformationContext) {
...
-        if (node.kind === SyntaxKind.VariableDeclaration || node.kind === SyntaxKind.BindingElement) {
-            return cleanup(resolver.createTypeOfDeclaration(node, enclosingDeclaration, declarationEmitNodeBuilderFlags, symbolTracker));
-        }
-        if (
-            node.kind === SyntaxKind.Parameter
-            || node.kind === SyntaxKind.PropertyDeclaration
-            || node.kind === SyntaxKind.PropertySignature
-        ) {
-            if (isPropertySignature(node) || !node.initializer) return cleanup(resolver.createTypeOfDeclaration(node, enclosingDeclaration, declarationEmitNodeBuilderFlags, symbolTracker, shouldUseResolverType));
-            return cleanup(resolver.createTypeOfDeclaration(node, enclosingDeclaration, declarationEmitNodeBuilderFlags, symbolTracker, shouldUseResolverType) || resolver.createTypeOfExpression(node.initializer, enclosingDeclaration, declarationEmitNodeBuilderFlags, symbolTracker));
```
dragomirtitian, line 773: The logic here is overly complicated. Both calls to `createTypeOfDeclaration` boil down to the same arguments. Also I have not found a test that benefits from the use of `createTypeOfExpression` on the initializer and I can't think of a case when this would help. (There might be some error case, but it's not tested for and returning `any` for that seems just as appropriate.

src/compiler/transformers/declarations.ts
```diff
@@ -741,46 +739,43 @@ export function transformDeclarations(context: TransformationContext) {
...
+            case SyntaxKind.MethodSignature:
+            case SyntaxKind.MethodDeclaration:
+            case SyntaxKind.GetAccessor:
+            case SyntaxKind.CallSignature:
+                typeNode = resolver.createReturnTypeOfSignatureDeclaration(node, node, declarationEmitNodeBuilderFlags, symbolTracker);
+                break;
+            default:
+                Debug.assertNever(node);
         }
-        return cleanup(resolver.createReturnTypeOfSignatureDeclaration(node, enclosingDeclaration, declarationEmitNodeBuilderFlags, symbolTracker));
```
dragomirtitian, line 775: Removed the `cleanup` closure. Since the logic is simpler now it's easier to just assign a variable and do the same cleanup after the switch.

src/compiler/types.ts
```diff
@@ -5699,8 +5699,7 @@ export interface EmitResolver {
     isLateBound(node: Declaration): node is LateBoundDeclaration;
     collectLinkedAliases(node: Identifier, setVisibility?: boolean): Node[] | undefined;
     isImplementationOfOverload(node: SignatureDeclaration): boolean | undefined;
-    isRequiredInitializedParameter(node: ParameterDeclaration): boolean;
-    isOptionalUninitializedParameterProperty(node: ParameterDeclaration): boolean;
+    requiresAddingImplicitUndefined(node: ParameterDeclaration): boolean;
```
dragomirtitian, line 5702: `isRequiredInitializedParameter` and `isOptionalUninitializedParameterProperty` were used only once in the declaration transform in `(resolver.isRequiredInitializedParameter(node) || resolver.isOptionalUninitializedParameterProperty(node))`. I think it's worth simplifying the API and only expose a function taht tells us is the parameter requires adding an implcit `undefined` to it's type or not. 

`requiresAddingImplicitUndefined` also takes into account if the type already contains undefined and thus avoids re-printing types unless it's absolutely necessary. (Fixing #57483)

src/compiler/checker.ts
```diff
@@ -48219,6 +48219,15 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         return false;
     }
 
+    function declaredParameterTypeContainsUndefined(parameter: ParameterDeclaration) {
+        if (!parameter.type) return false;
+        const type = getTypeFromTypeNode(parameter.type);
+        return type === undefinedType || !!(type.flags & TypeFlags.Union) && !!((type as UnionType).types[0].flags & TypeFlags.Undefined);
```
jakebailey, line 48225: Can this just be `containsUndefinedType`?

dragomirtitian: Yes it can be. I didn't find it and was going off what is in the `getOptionalType` functions:

```ts
    function getOptionalType(type: Type, isProperty = false): Type {
        Debug.assert(strictNullChecks);
        const missingOrUndefined = isProperty ? undefinedOrMissingType : undefinedType;
        return type === missingOrUndefined || type.flags & TypeFlags.Union && (type as UnionType).types[0] === missingOrUndefined ? type : getUnionType([type, missingOrUndefined]);
    }
```

jakebailey:
@typescript-bot test top200

@typescript-bot pack this

tests/baselines/reference/verbatim-declarations-parameters.js
```diff
@@ -0,0 +1,74 @@
...
+    [P in string]: any;
+};
+type MapOrUndefined = Map | undefined | "dummy";
+export declare class Foo {
+    reuseTypeNode?: Map | undefined;
+    reuseTypeNode2?: Exclude<MapOrUndefined, "dummy">;
+    resolveType?: {
+        [x: string]: any;
+    } | undefined;
+    constructor(reuseTypeNode?: Map | undefined, reuseTypeNode2?: Exclude<MapOrUndefined, "dummy">, resolveType?: {
```
jakebailey, line 67: I assume this is preexisting, but why is the "resolveType" prop/parameter resolved at all, and not a simple rewrite like the other params? Seems odd to me that reuseTypeNode is _more_ complex and doesn't get rewritten, but resolveType does and is.

dragomirtitian: `resolveType` requires adding `undefined` to the type. So we can't directly use the original type as it was written, unless it already contains `undefined`

jakebailey: I see. How does this work in the context of isolatedDEclarations, where we presumably would instead want to syntactically do this? Isn't the intent to just add `| undefined` unconditionally?

dragomirtitian: I was planning to do that in one of the follow on PRs since that would arguably not be an improvement to current emit.  But I could add it as part of this PR as well if you think it's uncontroversial.

jakebailey: I feel moderately like they're tied together, but I suppose that this PR is okay as-is as it's really just saying "if the type already has undefined, don't add it", which is a simple enough thing to reason about.

jakebailey: I guess this is still an improvement as it leads to increased node reuse...

jakebailey (approved):
LGTM though I am surprised that no other baselines changed.

jakebailey:
To be clear, this doesn't actually fix #57483, though, right? As we don't have the follow-up to continue node reuse?

tests/baselines/reference/verbatim-declarations-parameters.js
dragomirtitian: It will have an effect. If we switch to always add ` | undefined` it will always reuse the existing type node (now it's only reuses it if we don't need to add `undefined` because the declared type already includes it)

dragomirtitian:
> To be clear, this doesn't actually fix #57483, though, right? As we don't have the follow-up to continue node reuse?

It reuses type nodes if we know they are correct (if they already contain `undefined`). This is better than the situation before where it never reused the declared type node (see [test output before the change](https://github.com/microsoft/TypeScript/pull/57484/commits/3e041023b9c4e50f43234c00b0c9f20cf65b50e0#diff-21e7f70761c001f6689dd65a2ad642c716ca7911fa4e407fd9c91ff44ce1e9f7R56)). A follow on PR can always reuse the type node and add `| undefined` even if the declared type already includes it.

PR #57196, opened 2024-01-27
Search ancestor and its references for default projects

sheetalkamat:
On Top of #58486

When finding default project for open file, we search tsconfig .. Before this change, we would stop at first tsconfig/jsconfig encountered and check if any of references has the file. If none of this had file we would stop and put file in inferred projects. 
This has issues when you arrange code such that your test files are next to source files and are not using project references or the default project is the project for building sources. So test files endup in inferred project and dont get their custom options. 

With this change, 

- Continue search for the tsconfig further if we couldnt find default project as if we didnt find this tsconfig (so till projectRootPath) and try to find the default project.
- Caches the config file name for ancestor lookups as well as we will be skipping to config if it doesn't contain project
- Fixed config search of ancestor lookup to not skip jsconfig if tsconfig is present

Fixes #56959

PR #57492, opened 2024-02-22
Keep scriptInfo and project alive even after file delete till next file open

sheetalkamat:
Some times when updating branches or doing `npm ci`, instead of file change event, we may get file delete followed by file create event. 
This was previously handled not so efficiently as: as soon as we get delete event say on configFile for the project, we would remove the whole project from memory and schedule update for open files, only to get file create so end up creating project again. 

With this change:
1. 784e0efed93633c1e35d510a1b229b9e96978044 When config file watched gets deleted, we defer removing/closing it and reuse if we get create event, we can just mark this project for update and remove its `defer delete` option. The deferred delete marked project is kept alive and acts as orphan project until next file open which is when we normally do `gc` on all projects and scriptInfos. Note we do something similar for inferred project so we can reuse it so its just extending that concept to configured project
2. bd101171c33b5f835bf1b3306b13d71b080e79cc For scriptInfos that are deleted, they are also kept around and watched till next file open, so we can reuse the text and version if it doesnt change. Eg `npm ci` can replace file with same text and this helps with file usage. Again these deferred removed on next file open.

Other two commits in the PR are tests before the change to cover for scenarios of file delete, recreate etc and refactoring

This should also help with #57196 when we would open more projects and configs to search for your default project

Without proper repro to measure perf reliably but in my dogfooding and from tests it seems to really help with the perf in these kind of scenarios. (No guarantee when timeout happens and how things will unfold to see if delete, create happen or not)

### DIFF
--- a/tests/baselines/reference/api/typescript.d.ts
+++ b/tests/baselines/reference/api/typescript.d.ts
@@ -2813,7 +2813,6 @@ declare namespace ts {
             addMissingFileRoot(fileName: NormalizedPath): void;
             removeFile(info: ScriptInfo, fileExists: boolean, detachFromProject: boolean): void;
             registerFileUpdate(fileName: string): void;
-            markAsDirty(): void;
             /**
              * Updates set of files that contribute to this project
              * @returns: true if set of files in the project stays the same and false - otherwise.
@@ -2858,10 +2857,8 @@ declare namespace ts {
         class AutoImportProviderProject extends Project {
             private hostProject;
             private rootFileNames;
-            isOrphan(): boolean;
             updateGraph(): boolean;
             hasRoots(): boolean;
-            markAsDirty(): void;
             getScriptFileNames(): string[];
             getLanguageService(): never;
             getHostForAutoImportProvider(): never;
@@ -2895,8 +2892,6 @@ declare namespace ts {
             getAllProjectErrors(): readonly Diagnostic[];
             setProjectErrors(projectErrors: Diagnostic[]): void;
             close(): void;
-            markAsDirty(): void;
-            isOrphan(): boolean;
             getEffectiveTypeRoots(): string[];
         }
         /**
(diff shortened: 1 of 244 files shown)


tests/baselines/reference/api/typescript.d.ts
```diff
@@ -2895,6 +2895,8 @@ declare namespace ts {
             getAllProjectErrors(): readonly Diagnostic[];
             setProjectErrors(projectErrors: Diagnostic[]): void;
             close(): void;
+            markAsDirty(): void;
```
jakebailey, line 2898: How much of this are you still wanting to be public?

tests/baselines/reference/api/typescript.d.ts
```diff
@@ -3187,6 +3189,13 @@ declare namespace ts {
             private delayUpdateSourceInfoProjects;
             private delayUpdateProjectsOfScriptInfoPath;
             private handleDeletedFile;
+            /**
+             * This function goes through all the openFiles and tries to file the config file for them.
+             * If the config file is found and it refers to existing project, it schedules the reload it for reload
+             * If there is no existing project it just opens the configured project for the config file
+             * shouldReloadProjectFor provides a way to filter out files to reload configured project for
+             */
+            private delayReloadConfiguredProjectsForFile;
```
jakebailey, line 3198: Not for this PR of course, but it sure seems like we should be dropping JSDoc on private props in dts emit...

jakebailey: https://github.com/microsoft/TypeScript/issues/58145

tests/baselines/reference/api/typescript.d.ts
sheetalkamat: This is public because for some reason Project.markAsDirty is public too ? But i can mark them internal .. not very comfortable with these methods being public at all

jakebailey: Whatever you'd prefer; I don't really know whether or not anyone needs this externally. ts-eslint's use case for these is solely just "opening files" and "getting the checker"

jakebailey:

jakebailey (approved):
This seems fine to me (though I don't know how to physically test it), though I think my main concern are API users who expect that closing a file actually frees its resources. I'm not sure if `ts-eslint` actually closes any files, though.

Issue #57487, opened 2024-02-22
Regression: Indexed Access of Partial Mapped Type with Union from Generic Function is missing `undefined`


3x071c:
### 🔎 Search Terms

generic function, partial, optional, mapped type, indexed access, undefined, bug, regression

### 🕗 Version & Regression Information

- This is a bug
- This changed between versions 4.5.5 and 4.6.2
- This is the behavior in every version I tried (all playground versions)

### ⏯ Playground Link

https://www.typescriptlang.org/play?target=99&jsx=0&pretty=true&ts=4.6.2#code/C4TwDgpgBA8gRgKygXigbwFBW1AhgLigEYAaLHOQgJgwF8BuDDAYwHsA7AZ2CgFtcwkACaFMOKAG0A0lACW7KAGsIIVgDMoABVwAnYLNwAbADzwEAPgC6hM9Mt0U6WkzZceOiJ1aGAbhACyAsKOxjIQAB7AEOxCnEoq6lCgkIn8ghBC5gAUyiCEUgCUKOZ8QRkSuZYAdMCsAMrAOvIA5lkFjAD0HVAAegC0fVAAojo6rDpQEZDMUUJQWQAGacIVKpYLcnFgrJycsnCGIFAArjEQavIZBRhAA

### 💻 Code

```ts
type Obj = {
    a: 1,
    b: 2
};

const mapped: {
    [K in keyof Partial<Obj>]: Obj[K]
} = {}

const resolveMapped = <K extends keyof typeof mapped>(key: K) => mapped[key].toString();
// ^-- Error expected (`mapped[key]` is possibly undefined)
```

### 🙁 Actual behavior

TypeScript does not recognize `mapper[key]` might be undefined and emits broken code without errors

### 🙂 Expected behavior

Correct type inference from pre-v4.6.x

RyanCavanaugh:
Isn't the bug earlier? The type `{ [K in keyof Partial<Obj>]: Obj[K] }` should be `{ a: number, b: number }` and the initialization from `{ }` should fail.

3x071c:
@RyanCavanaugh The `mapped` definition seems to work as intended (`{ a?: number, b?: number }`), as it does in all TypeScript versions since 3.3.3 (from the playground). Kinda surprised if that actually isn't expected behavior.

RyanCavanaugh:
Bisects to #47109

jcalz:
[From here](https://stackoverflow.com/a/78043776/2887218)

PR #57493, opened 2024-02-22
Add optionality to mapped type substitutions based on existence of an optional modifiers prop

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/57487

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -18514,10 +18514,20 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         return !!(getUnionType([intersectTypes(type1, type2), neverType]).flags & TypeFlags.Never);
     }
 
+    function hasOptionalModifiersProperty(mappedType: MappedType) {
+        const modifiersType = getApparentType(getModifiersTypeFromMappedType(mappedType));
+        if (!(modifiersType.flags & TypeFlags.Object)) {
+            return false;
+        }
+        const properties = resolveStructuredTypeMembers(modifiersType as ObjectType).properties;
+        return some(properties, p => !!(p.flags & SymbolFlags.Optional));
+    }
+
     function substituteIndexedMappedType(objectType: MappedType, index: Type) {
         const mapper = createTypeMapper([getTypeParameterFromMappedType(objectType)], [index]);
         const templateMapper = combineTypeMappers(objectType.mapper, mapper);
-        return instantiateType(getTemplateTypeFromMappedType(objectType.target as MappedType || objectType), templateMapper);
+        const type = instantiateType(getTemplateTypeFromMappedType(objectType.target as MappedType || objectType), templateMapper);
+        return addOptionality(type, /*isProperty*/ true, hasOptionalModifiersProperty(objectType));
     }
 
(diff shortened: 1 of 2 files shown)


src/compiler/checker.ts
```diff
@@ -18514,10 +18514,20 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         return !!(getUnionType([intersectTypes(type1, type2), neverType]).flags & TypeFlags.Never);
     }
 
+    function hasOptionalModifiersProperty(mappedType: MappedType) {
+        const modifiersType = getApparentType(getModifiersTypeFromMappedType(mappedType));
+        if (!(modifiersType.flags & TypeFlags.Object)) {
```
Andarist, line 18519: TODO: might also have to handle intersection (that's what `isWeakType` is doing) and I need to recheck behavior with unions

tests/baselines/reference/keyofAndIndexedAccess2.errors.txt
```diff
@@ -205,6 +206,9 @@ keyofAndIndexedAccess2.ts(108,5): error TS2322: Type '123' is not assignable to
     type StrictExtract<T, U> = T extends U ? U extends T ? T : never : never;
     type StrictExclude<T, U> = T extends StrictExtract<T, U> ? never : T;
     type A<T> = { [Q in { [P in keyof T]: P; }[keyof T]]: T[Q]; };
```
Andarist, line 208: I'm not sure what's the best strategy to avoid circularity here. I didn't look into it yet. I'm hitting the 🛌 but maybe somebody would chime in with some suggestions in the meantime ;p

src/compiler/checker.ts
Mateusnasciment: recheck behavior with union types later

RyanCavanaugh: @Mateusnasciment Please try to keep comments actionable and on-topic

src/compiler/checker.ts
```diff
@@ -18514,10 +18514,20 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+        const properties = resolveStructuredTypeMembers(modifiersType as ObjectType).properties;
+        return some(properties, p => !!(p.flags & SymbolFlags.Optional));
+    }
+
     function substituteIndexedMappedType(objectType: MappedType, index: Type) {
         const mapper = createTypeMapper([getTypeParameterFromMappedType(objectType)], [index]);
         const templateMapper = combineTypeMappers(objectType.mapper, mapper);
-        return instantiateType(getTemplateTypeFromMappedType(objectType.target as MappedType || objectType), templateMapper);
+        const type = instantiateType(getTemplateTypeFromMappedType(objectType.target as MappedType || objectType), templateMapper);
+        return addOptionality(type, /*isProperty*/ true, hasOptionalModifiersProperty(objectType));
```
Andarist, line 18530: TODO: this should get the list of known optional properties and check if `index` is assignable to that, if the substitution happens using a known non-optional property then the optionality shouldn't be added

ahejlsberg:
I have what I think it the right fix in #57549. The main difference is that my PR is based on computing combined modifier flags, which I think is a better approach. It avoids resolving the members, which may cause circularities as you've observed. Also, member resolution is based on what members are _known_ to be present, not which members _might_ be present.

PR #57495, opened 2024-02-22
Fix output artifact in new release pipeline

jakebailey:
No clue if it works, the docs have two different schemas for this same thing and the examples 404.

### DIFF
--- a/azure-pipelines.release.yml
+++ b/azure-pipelines.release.yml
@@ -57,8 +57,8 @@ extends:
                   Contents: 'typescript-*.tgz'
                   TargetFolder: '$(Build.ArtifactStagingDirectory)'
 
-              - task: PublishPipelineArtifact@0
-                displayName: 'Publish Pipeline Artifact'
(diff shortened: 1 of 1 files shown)


DanielRosenwasser (approved):
I see a config change, I roll the dice 🎲

azure-pipelines.release.yml
```diff
@@ -57,8 +57,8 @@ extends:
...
-              - task: PublishPipelineArtifact@0
-                displayName: 'Publish Pipeline Artifact'
-                inputs:
-                  artifactName: tgz
-                  targetPath: '$(Build.ArtifactStagingDirectory)'
+            templateContext:
+              outputs:
+                - output: pipelineArtifact
+                  path: '$(Build.ArtifactStagingDirectory)'
+                  artifact: tgz
```
noshlag: This should be targetPath and artifactName, not path and artifact

noshlag (approved):

Issue #51909, opened 2022-12-15
Feedback: 5.0 deprecations


RyanCavanaugh:
See also #51000 / #51424

In 5.0, we've marked the following configurations as deprecated:
 * `target: ES3`
 * `noImplicitUseStrict`
 * `keyofStringsOnly`
 * `suppressExcessPropertyErrors`
 * `suppressImplicitAnyIndexErrors`
 * `noStrictGenericChecks`
 * `charset`
 * `out`
 * `prepend` in project references
 * implicitly OS-specific `newLine`

These configurations will continue to "work" until TypeScript 5.5, at which point they will be removed entirely. In TypeScript 4.9.5+, 5.0, 5.1, 5.2, 5.3, and 5.4, you can specify `ignoreDeprecations: "5.0"` to silence this warning.

#### `target: ES3`

ES3 runtimes are nearly 100% extinct, so we're removing support for this as an emit target.

Fix: Use `target: ES5` as the next-closest thing along with a linter to detect any potentially-problematic ES3 constructs, or use `target: ESNext` and use a different downleveling emitter.

#### `noImplicitUseStrict`

This flag disables emitting a `"use strict";` prologue at the top of module bodies when the target is not implicitly a module. This can lead to unexpected behavior since there are subtle behavioral differences between strict and nonstrict mode.

Fix: Remove any reliance on nonstrict behavior.

#### `keyofStringsOnly`

This flag removes `symbol`-based keys from the `keyof T` type operator. This can lead to unexpected type errors when code relying on this flag interacts with code that doesn't.

Fix: Use `string & keyof T` to filter keys to those with `string` types.

#### `suppressExcessPropertyErrors`

This flag disables excess property checking. This can lead to unexpected program behavior when an optional property name is misspelled, since no error is issued.

Fix: use a type assertion to silence this in cases where you really do want an optional property, or fix the target type.

#### `suppressImplicitAnyIndexErrors`

This flag disables the error produced when indexing an object which doesn't support indexing. This was always intended to be a temporary flag, so is going away.

Fix: Add an index signature to the relevant type (or use a type assertion at the indexing location).

#### `noStrictGenericChecks`

This flag disables certain checks relating the constraints of generic types, which can lead to undesirable unsound behavior.

Fix: Correct the types, or use a type assertion.

#### `charset`

This flag hasn't done anything for a very long time.

Fix: Remove it from your config file

#### `out`

[...]

PR #57525, opened 2024-02-25
Remove target=es3

jakebailey:
For #51909

This removes ES3 related code, including `transformES5` (which does ES5 -> ES3), specific ES3 errors and language, and so on.

The actual enum value is kept; since our deprecation system detects deprecated values _after_ parsing, we can't delete it because we won't know the value is wrong.

If ES3 is set, it just turns into ES5. Doing this involved fixing how we compute the actual target, fixing some places to use the computed value, and fixing the harness (which eagerly resolved the computed value).

# Package size report

## Overall package size

| - | - | - | - | - |
| Packed | 5.56 MiB | 5.55 MiB | -6.24 KiB | -0.11% |
| Unpacked | 30.86 MiB | 30.82 MiB | -46.20 KiB | -0.15% |

## Files

| - | - | - | - | - |
| `lib/tsc.js` | 5.62 MiB | 5.61 MiB | -12.69 KiB | -0.22% |
| `lib/tsserver.js` | 8.24 MiB | 8.22 MiB | -12.76 KiB | -0.15% |
| `lib/typescript.d.ts` | 584.73 KiB | 584.76 KiB | +27.00 B | +0.00% |
| `lib/typescript.js` | 8.72 MiB | 8.70 MiB | -13.18 KiB | -0.15% |
| `lib/typingsInstaller.js` | 1.68 MiB | 1.67 MiB | -7.59 KiB | -0.44% |


tests/baselines/reference/nodeModules1(module=node16).errors.txt
```diff
@@ -15,27 +14,16 @@ index.cts(52,22): error TS1471: Module './index' cannot be imported using this c
 index.cts(59,22): error TS1471: Module './subfolder2/another' cannot be imported using this construct. The specifier only resolves to an ES module, which cannot be imported with 'require'. Use an ECMAScript import instead.
 index.cts(60,22): error TS1471: Module './subfolder2/another/' cannot be imported using this construct. The specifier only resolves to an ES module, which cannot be imported with 'require'. Use an ECMAScript import instead.
 index.cts(61,22): error TS1471: Module './subfolder2/another/index' cannot be imported using this construct. The specifier only resolves to an ES module, which cannot be imported with 'require'. Use an ECMAScript import instead.
-index.cts(75,14): error TS2712: A dynamic import call in ES5/ES3 requires the 'Promise' constructor.  Make sure you have a declaration for the 'Promise' constructor or include 'ES2015' in your '--lib' option.
```
jakebailey, line 18: It seems strange for this to be changed, but the testing harness eagerly resolved the target such that the "implicit" target for node16/nodenext was not used! Now it is.

tests/baselines/reference/nodeModules1(module=node16).js
```diff
@@ -325,51 +325,51 @@ export {x};
 Object.defineProperty(exports, "__esModule", { value: true });
 exports.x = void 0;
 // cjs format file
-var x = 1;
+const x = 1;
```
jakebailey, line 328: This is a test harness fix; this file has no target set, but sets modeule=node16, which is supposed to mean es2022.

jakebailey:
Removing the feature _added_ lines to the repo thanks to the test harness fixing. Oh well.

jakebailey:
Probably I'm going to split this PR in two to fix harness problems; this PR is noisier than it should be.

jakebailey:
With #57526 in, this is a lot clearer.

src/compiler/types.ts
```diff
@@ -7365,6 +7365,7 @@ export const enum ScriptKind {
 }
 
 export const enum ScriptTarget {
+    /** @deprecated */
```
sheetalkamat, line 7368: Should this also be marked internal ?

src/server/protocol.ts
```diff
@@ -3772,6 +3772,7 @@ export const enum NewLineKind {
 }
 
 export const enum ScriptTarget {
+    /** @deprecated */
```
sheetalkamat, line 3775: Same here?

src/compiler/program.ts
```diff
@@ -4366,10 +4366,6 @@ export function createProgram(rootNamesOrOptions: readonly string[] | CreateProg
             }
         }
 
-        if (options.useDefineForClassFields && languageVersion === ScriptTarget.ES3) {
```
sheetalkamat, line 4369: Do we have a test that still sets the target as es3 and results in option should be removed error.

sheetalkamat (changes requested):

src/compiler/types.ts
jakebailey: I said this on https://github.com/microsoft/TypeScript/pull/57472#discussion_r1503210241, but I don't think we should remove them entirely, at least not until 6.0.

sheetalkamat (approved):

src/compiler/scanner.ts
```diff
@@ -274,31 +274,6 @@ const textToToken = new Map(Object.entries({
...
-
```
RyanCavanaugh, line 301: yessssss

tests/baselines/reference/ES3For-ofTypeCheck4.errors.txt
```diff
@@ -1,10 +1,7 @@
 error TS5108: Option 'target=ES3' has been removed. Please remove it from your configuration.
-ES3For-ofTypeCheck4.ts(2,17): error TS2494: Using a string in a 'for...of' statement is only supported in ECMAScript 5 and higher.
 
 
 !!! error TS5108: Option 'target=ES3' has been removed. Please remove it from your configuration.
```
RyanCavanaugh, line 4: Is this in the follow-up PR?

jakebailey:
note to self: the website still lists es3 and even says it's the default even though it hasn't been for a while.

src/compiler/checker.ts
```diff
@@ -49039,8 +49019,6 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 return ["__classPrivateFieldSet"];
             case ExternalEmitHelpers.ClassPrivateFieldIn:
                 return ["__classPrivateFieldIn"];
-            case ExternalEmitHelpers.CreateBinding:
```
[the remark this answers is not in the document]
jakebailey: Nobody sets the flag anymore, so I don't think this code is reachable (your comment above suggests removing the flag altogether so I think it'll be fully dead). But, let me check.

There are other helpers which depend on `__createBinding`, which is why I didn't remove it entirely, but it seems like they depend on it using other means.

src/compiler/factory/emitHelpers.ts
```diff
@@ -129,6 +129,7 @@ export interface EmitHelperFactory {
     // ES2015 Generator Helpers
     createGeneratorHelper(body: FunctionExpression): Expression;
     // ES Module Helpers
+    /** @deprecated */
```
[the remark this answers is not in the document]
jakebailey: Oops, didn't realize this was internal. Will just remove it.

As for the other cleanups, I didn't feel qualified to go that far; definitely would be good as a followup.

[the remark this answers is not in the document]
jakebailey: Removed this.

src/compiler/types.ts
```diff
@@ -8152,6 +8153,7 @@ export const enum ExternalEmitHelpers {
     ClassPrivateFieldGet = 1 << 19, // __classPrivateFieldGet (used by the class private field transformation)
     ClassPrivateFieldSet = 1 << 20, // __classPrivateFieldSet (used by the class private field transformation)
     ClassPrivateFieldIn = 1 << 21,  // __classPrivateFieldIn (used by the class private field transformation)
+    /** @deprecated */
```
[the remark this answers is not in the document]
jakebailey: Removed.

src/compiler/checker.ts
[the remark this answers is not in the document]
jakebailey: The flag is now gone, so I assume that there's nothing to do here now?

src/compiler/types.ts
```diff
@@ -8152,7 +8153,6 @@ export const enum ExternalEmitHelpers {
     ClassPrivateFieldGet = 1 << 19, // __classPrivateFieldGet (used by the class private field transformation)
     ClassPrivateFieldSet = 1 << 20, // __classPrivateFieldSet (used by the class private field transformation)
     ClassPrivateFieldIn = 1 << 21,  // __classPrivateFieldIn (used by the class private field transformation)
-    CreateBinding = 1 << 22,        // __createBinding (use by the module transform for (re)exports and namespace imports)
     SetFunctionName = 1 << 23,      // __setFunctionName (used by class fields and ECMAScript decorators)
     PropKey = 1 << 24,              // __propKey (used by class fields and ECMAScript decorators)
     AddDisposableResourceAndDisposeResources = 1 << 25, // __addDisposableResource and __disposeResources (used by ESNext transformations)
```
rbuckton, line 8158: I would shift these all up so it's easier to see how much room is left in the enum:

```suggestion
    SetFunctionName = 1 << 22,      // __setFunctionName (used by class fields and ECMAScript decorators)
    PropKey = 1 << 23,              // __propKey (used by class fields and ECMAScript decorators)
    AddDisposableResourceAndDisposeResources = 1 << 24, // __addDisposableResource and __disposeResources (used by ESNext transformations)

```

rbuckton (changes requested):

src/compiler/types.ts
jakebailey: Done.

rbuckton (approved):

PR #57526, opened 2024-02-25
Ensure harness doesn't force or eagerly computes `target`, stop using raw `target` elsewhere

jakebailey:
Pulled out of #57525 to make that change less noisy.

- The test harness no longer forces a `target` in certain conditions; this had the affect of making some tests not test the implicit target controlled by the `module` option.
- The test harness no longer eagerly computes `target`; eagerly computing it had the affect of making our nice "this setting is dead, please remove" tests not work (because a change like #57525 would change `target` from ES3 into ES5 at use).
- I found a small number of places which were using `target` directly, when they should have been using the computed target.


src/harness/compilerImpl.ts
```diff
@@ -257,7 +257,6 @@ export function compileFiles(host: fakes.CompilerHost, rootFiles: string[] | und
     }
 
     // establish defaults (aligns with old harness)
-    if (compilerOptions.target === undefined && compilerOptions.module !== ts.ModuleKind.Node16 && compilerOptions.module !== ts.ModuleKind.NodeNext) compilerOptions.target = ts.ScriptTarget.ES3;
     if (compilerOptions.newLine === undefined) compilerOptions.newLine = ts.NewLineKind.CarriageReturnLineFeed;
     if (compilerOptions.skipDefaultLibCheck === undefined) compilerOptions.skipDefaultLibCheck = true;
     if (compilerOptions.noErrorTruncation === undefined) compilerOptions.noErrorTruncation = true;
```
jakebailey, line 262: I want to say that we should drop these too, but:

- If we change the default `newLine`, all baselines will change from CRLF to LF and lose git history.
- `skipDefaultLibCheck` is a perf optimization.
- `noErrorTruncation` gives us better baselines.

src/services/inlayHints.ts
```diff
@@ -350,7 +351,7 @@ export function provideInlayHints(context: InlayHintsContext): InlayHint[] {
     }
 
     function leadingCommentsContainsParameterName(node: Node, name: string) {
-        if (!isIdentifierText(name, compilerOptions.target, getLanguageVariant(file.scriptKind))) {
+        if (!isIdentifierText(name, getEmitScriptTarget(compilerOptions), getLanguageVariant(file.scriptKind))) {
```
jakebailey, line 354: It seems quite easy to make this kind of error; I am tempted to write an eslint rule which disallows the use of any CompilerOption property that we know to be computed, but that is non-trivial.

tests/baselines/reference/bundlerDirectoryModule(moduleresolution=bundler).trace.json
```diff
@@ -18,6 +18,136 @@
     "File '/lib/cjs/index.tsx' does not exist.",
     "File '/lib/cjs/index.d.ts' exists - use it as a name resolution result.",
     "======== Module name '../lib' was successfully resolved to '/lib/cjs/index.d.ts'. ========",
+    "======== Resolving module '@typescript/lib-esnext' from '/.src/__lib_node_modules_lookup_lib.esnext.d.ts__.ts'. ========",
```
jakebailey, line 21: I believe this baseline change to be correct; this test would (outside of the test harness) have an implicit target of `esnext`, but we were forcing it to be es5.

tests/baselines/reference/moduleResolutionWithoutExtension8.errors.txt
```diff
@@ -1,13 +1,8 @@
-error TS2468: Cannot find global value 'Promise'.
-/src/bar.cts(2,1): error TS2712: A dynamic import call in ES5/ES3 requires the 'Promise' constructor.  Make sure you have a declaration for the 'Promise' constructor or include 'ES2015' in your '--lib' option.
```
jakebailey, line 2: Another example of the implicit module setting now being reflected in tests; this was defaulted to ES5, but now it's esnext.

PR #57527, opened 2024-02-25
Remove 5.5 deprecated compiler options

jakebailey:
This removes all of the deprecated options listed in https://github.com/microsoft/TypeScript/issues/51909 (except es3, which was removed separately in #57525).

Like #57525, the options aren't fully removed; we still need to error on them.

# Package size report

## Overall package size

| - | - | - | - | - |
| Packed | 5.52 MiB | 5.52 MiB | -4.63 KiB | -0.08% |
| Unpacked | 30.67 MiB | 30.65 MiB | -24.41 KiB | -0.08% |

## Files

| - | - | - | - | - |
| `lib/tsc.js` | 5.57 MiB | 5.56 MiB | -7.59 KiB | -0.13% |
| `lib/tsserver.js` | 8.18 MiB | 8.17 MiB | -7.92 KiB | -0.09% |
| `lib/typescript.d.ts` | 579.90 KiB | 580.33 KiB | +441.00 B | +0.07% |
| `lib/typescript.js` | 8.66 MiB | 8.65 MiB | -8.13 KiB | -0.09% |
| `lib/typingsInstaller.js` | 1.67 MiB | 1.67 MiB | -1.21 KiB | -0.07% |


jakebailey:
@typescript-bot test top200
@typescript-bot run dt

jakebailey:
@andrewbranch The last two commits remove `importsNotUsedAsValues` and `preserveValueImports`, but breaks some auto import tests. Some of them are easy to fix, but others I can't seem to fix. Do you have any suggestions? (The CI will eventually error...)

jakebailey:
Theoretically I think this PR is ready, and I'd appreciate review where needed, but I probably won't undraft until #57525 is merged. that or, I can merge the two together and we remove everything in one go.

src/compiler/utilities.ts
```diff
@@ -6302,7 +6298,7 @@ export function getPossibleOriginalInputExtensionForExtension(path: string) {
 
 /** @internal */
 export function outFile(options: CompilerOptions) {
-    return options.outFile || options.out;
+    return options.outFile;
```
sheetalkamat, line 6301: Can we remove this helper and use `compilerOptions.outFile` directly everywhere now?

src/compiler/utilities.ts
```diff
@@ -8889,7 +8885,7 @@ export function hasJsonModuleEmitEnabled(options: CompilerOptions) {
 
 /** @internal */
 export function importNameElisionDisabled(options: CompilerOptions) {
-    return options.verbatimModuleSyntax || options.isolatedModules && options.preserveValueImports;
+    return options.verbatimModuleSyntax;
```
sheetalkamat, line 8888: Can we inline this as well

sheetalkamat (changes requested):

src/compiler/utilities.ts
jakebailey: Sure, I can inline this (and the other one).

jakebailey: Done.

src/compiler/utilities.ts
jakebailey: Done.

RyanCavanaugh (approved):

Issue #54309, opened 2023-05-18
"Move to file" generates self-import when target file already imports declaration


DanielRosenwasser:
Possibly the same as #54307.

`basics.ts`

```ts
export function getHello() {
    return "hello";
}

export function getWorld() {
    return "world";
}```

`greeting.ts`

```ts
import { getHello, getWorld } from "./basics";

export function getGreeting() {
    return getHello() + getWorld();
}
```

Extract `getWorld` into `greeting.ts` and this is what it looks like:

```ts
import { getHello } from "./basics";
import { getWorld } from "./greeting";

export function getGreeting() {
    return getHello() + getWorld();
}
export function getWorld() {
    return "world";
}
```

RyanCavanaugh:
Fixed by linked PR

PR #57530, opened 2024-02-25
fix(57451): Prevent self-imports when using the "Move to File" refactor

### LINKED ISSUE #57451: Incorrect import destination after using `Move to file`

- VS Code Version: 1.87.0-insider (Universal)
- OS Version: macOS 14.3.1

Steps to Reproduce:

1. Create files `utils.ts` and `program.ts`
2. Create and export any function inside `utils.ts`.
3. Import this function from `program.ts`.
4. Move the function into `program.ts` using the `Refactor` -> `Mote to file` action.

**Expected result**:
The function should be excluded from the import in `program.ts`.

**Actual result**:
The import statement for the function remains present, and the import destination changes from `utils` to `program` (which would not work).

eloyrobillard:
Fixes #57451

### DIFF
--- a/src/services/refactors/moveToFile.ts
+++ b/src/services/refactors/moveToFile.ts
@@ -18,6 +18,7 @@ import {
     ClassDeclaration,
     codefix,
     combinePaths,
+    Comparison,
     concatenate,
     contains,
     createModuleSpecifierResolutionHost,
@@ -52,12 +53,14 @@ import {
     getDirectoryPath,
     getLocaleSpecificMessage,
     getModifiers,
+    getNormalizedAbsolutePath,
     getPropertySymbolFromBindingElement,
     getQuotePreference,
     getRangesWhere,
     getRefactorContextSpan,
     getRelativePathFromFile,
     getSourceFileOfNode,
+    getStringComparer,
     getSynthesizedDeepClone,
     getTokenAtPosition,
     getUniqueName,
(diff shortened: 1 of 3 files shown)


eloyrobillard:

src/services/refactors/moveToFile.ts
```diff
@@ -430,6 +430,10 @@ export function updateImportsInOtherFiles(
                 deleteUnusedImports(sourceFile, importNode, changes, shouldMove); // These will be changed to imports from the new file
 
                 const pathToTargetFileWithExtension = resolvePath(getDirectoryPath(oldFile.path), targetFileName);
+
+                // no self-imports
+                if (sourceFile.fileName === pathToTargetFileWithExtension) return;
+
```
sheetalkamat, line 436:
```suggestion
                const pathToTargetFileWithExtension = resolvePath(getDirectoryPath(getNormalizedAbsolutePath(oldFile.fileName, program.getCurrentDirectory())), targetFileName);
                // no self imports
                if (getStringComparer(!program.useCaseSensitiveFileNames())(pathToTargetFileWithExtension, sourceFile.fileName) === Comparison.EqualTo) return; 

```

sheetalkamat: @navya9singh  we also need to replace usage of `sourceFile.path` in `makeImportOrReqiure` so its not using canonical path but actual file name: 
```ts
 const pathToTargetFile = resolvePath(getDirectoryPath(sourceFile.path), targetFileNameWithExtension);
```
to 
```ts
const pathToTargetFile = resolvePath(getDirectoryPath(getNormalizedAbsolutePath(sourceFile.fileName, program.getCurrentDirectory())), targetFileNameWithExtension);

```

sheetalkamat (changes requested):

src/services/refactors/moveToFile.ts
eloyrobillard: made the changes

sheetalkamat (approved):

navya9singh:
Thanks for working on this! This pr also fixes https://github.com/microsoft/TypeScript/issues/54309

navya9singh (approved):

PR #57537, opened 2024-02-25
Improve best type matching for elementwise elaborations

Andarist:


src/compiler/checker.ts
```diff
@@ -20352,16 +20353,16 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     }
 
     function getBestMatchIndexedAccessTypeOrUndefined(source: Type, target: Type, nameType: Type) {
-        const idx = getIndexedAccessTypeOrUndefined(target, nameType);
-        if (idx) {
-            return idx;
-        }
         if (target.flags & TypeFlags.Union) {
```
Andarist, line 20356: swapping the order of operations here improves some errors quite a bit in my eyes since the union discrimination happens early, for example:
```ts
const obj6: {
  prop:
    | {
        type: "foo";
        prop: string;
      }
    | {
        type: "bar";
        prop: number;
      };
} = {
  prop: {
    type: "foo",
    // current: Type 'boolean' is not assignable to type 'string | number'.
    // pr: Type 'boolean' is not assignable to type 'string'
    prop: true,
  },
};

const obj7: {
  prop:
    | {
        type: "foo";
        prop: string;
      }
    | {
        type: "bar";
        prop: number;
      };
} = {
  // current: Type '{ type: "foo"; prop: number; }' is not assignable to type '{ type: "foo"; prop: string; } | { type: "bar"; prop: number; }'.
  //  Types of property 'prop' are incompatible.
  //    Type 'number' is not assignable to type 'string'.
  prop: {
    type: "foo",
    // pr: Type 'number' is not assignable to type 'string'.
    prop: 42,
  },
};
```

src/compiler/checker.ts
```diff
@@ -21983,8 +21984,8 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 }
             }
             if (reportErrors) {
-                // Elaborate only if we can find a best matching type in the target union
```
Andarist, line 21986: I removed this `union` from here because it was misleading - the containing function handles both unions and intersections. However, from what I can tell `reportErrors: true` is only ever passed in for unions - I can revert this if requested

src/compiler/checker.ts
```diff
@@ -21983,8 +21984,8 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 }
             }
             if (reportErrors) {
-                // Elaborate only if we can find a best matching type in the target union
-                const bestMatchingType = getBestMatchingType(source, target, isRelatedTo);
+                // Elaborate only if we can find a best matching type in the target
+                const bestMatchingType = getBestMatchingType(source, target, /*matchSingleOverlappy*/ true, isRelatedTo);
```
Andarist, line 21988: Returning a single overlappy type is required here to make the non-elementwise union-based elaboration to elaborate about the "last" element in the union, see the "recovered" errors in this commit: https://github.com/microsoft/TypeScript/pull/57537/commits/e9d136a2b89a7a7c7f535dcdf0f5bfadce9396c0

src/compiler/checker.ts
```diff
@@ -50901,10 +50901,18 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         }
     }
 
-    function findBestTypeForObjectLiteral(source: Type, unionTarget: UnionOrIntersectionType) {
-        if (getObjectFlags(source) & ObjectFlags.ObjectLiteral && someType(unionTarget, isArrayLikeType)) {
-            return find(unionTarget.types, t => !isArrayLikeType(t));
```
Andarist, line 50906: The referenced bug was a somewhat funny one because optional properties always include `| undefined` in their types and `undefinedType` is always kept at the beginning of `unionType.types`. So in such a situation this was **always** returning that `undefinedType` - making it impossible to create a good elementwise elaboration.

src/compiler/checker.ts
```diff
@@ -50901,10 +50901,18 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
-    function findBestTypeForObjectLiteral(source: Type, unionTarget: UnionOrIntersectionType) {
-        if (getObjectFlags(source) & ObjectFlags.ObjectLiteral && someType(unionTarget, isArrayLikeType)) {
-            return find(unionTarget.types, t => !isArrayLikeType(t));
+    function filterTypesForObjectLiteralMatch(source: Type, target: UnionOrIntersectionType) {
+        if (getObjectFlags(source) & ObjectFlags.ObjectLiteral && target.flags & TypeFlags.Union) {
+            const filtered = filterType(target, t =>
+                !(t.flags & TypeFlags.Primitive) &&
+                !isArrayLikeType(t) &&
+                !typeHasCallOrConstructSignatures(t) &&
+                (everyContainedType(t, t => !t.symbol || !(t.symbol.flags & SymbolFlags.Class)) || !containsNonPublicProperties(getAugmentedPropertiesOfType(t))));
```
Andarist, line 50910: this is quite related to https://github.com/microsoft/TypeScript/pull/56183 - although completions and contextual types are orthogonal to elementwise elaborations. I only figured out that I should filter those nominal types here because I was working on that other PR in the past

tests/baselines/reference/jsxChildrenArrayWrongType.errors.txt
Andarist, line 1: Changed errors here might feel a little bit questionable. I feel like it's a good change and that it indirectly matches some other situations. 

Let's take a look at this one:
```ts
function test(_: { children: [string, number] | boolean[] }) {}

const children = [{}, ""] satisfies [unknown, unknown];

test({
  // Type '[{}, string]' is not assignable to type '[string, number] | boolean[]'.
  //   Type '[{}, string]' is not assignable to type '[string, number]'.
  //     Type at position 0 in source is not compatible with type at position 0 in target.
  //       Type '{}' is not assignable to type 'string'.(2322)
  children,
});
```

Regardless of the order of this union we get the same error. The tuple gets priority selection here and that's exactly what happens with my changes here. Since `getBestMatchingType` gets called before `getIndexedAccessTypeOrUndefined` I now select the tuple through `findMatchingTypeReferenceOrTypeAliasReference`.

If you feel strongly about it I can work on special-casing tuples/arrays in this algorithm.

tests/baselines/reference/indirectDiscriminantAndExcessProperty.errors.txt
Andarist, line 1: Those errors were improved based on overlapping type selection. Even though the discriminant property has the error (so we can't quite know which type it was supposed to be) - the other properties indicate the user's intention.

tests/baselines/reference/excessPropertyCheckWithUnions.errors.txt
```diff
@@ -108,9 +106,8 @@ excessPropertyCheckWithUnions.ts(112,63): error TS2322: Type 'string' is not ass
             a: "a",
             b: "b", // excess -- kind: "A"
             ~
-!!! error TS2322: Type '{ kind: "A"; n: { a: string; b: string; }; }' is not assignable to type 'AB'.
-!!! error TS2322:   Types of property 'n' are incompatible.
-!!! error TS2322:     Object literal may only specify known properties, and 'b' does not exist in type 'AN'.
+!!! error TS2353: Object literal may only specify known properties, and 'b' does not exist in type 'AN'.
```
Andarist, line 109: This feels like a straight improvement - this was supposed to look like this since https://github.com/microsoft/TypeScript/pull/55152 (cc @RyanCavanaugh - maybe it would be worth rechecking if the approach used by that PR doesn't have any other holes)

DanielRosenwasser:

tests/baselines/reference/indirectDiscriminantAndExcessProperty.errors.txt
```diff
@@ -14,7 +14,7 @@ indirectDiscriminantAndExcessProperty.ts(22,5): error TS2322: Type 'string' is n
     thing({
         type: foo1,
         ~~~~
-!!! error TS2322: Type 'string' is not assignable to type '"foo" | "bar"'.
+!!! error TS2322: Type 'string' is not assignable to type '"foo"'.
```
DanielRosenwasser, line 17: Heh - this is kind of better if you're already familiar with string literals, but I'd argue overall worse in that it doesn't give as much of a hint that `type` is really the discriminant.

tests/baselines/reference/jsxChildrenArrayWrongType.errors.txt
```diff
@@ -20,9 +20,9 @@ index.tsx(11,6): error TS2769: No overload matches this call.
          ~~~
 !!! error TS2769: No overload matches this call.
 !!! error TS2769:   Overload 2 of 2, '(props: PropsType, context: any): Foo', gave the following error.
-!!! error TS2769:     Type 'unknown' is not assignable to type 'string | boolean'.
+!!! error TS2769:     Type 'unknown' is not assignable to type 'string'.
```
DanielRosenwasser, line 23: This does seem worse overall, but it feels like JSX children should have better messages.

tests/baselines/reference/objectLiteralNormalization.errors.txt
```diff
@@ -41,10 +39,9 @@ objectLiteralNormalization.ts(49,14): error TS2322: Type '2' is not assignable t
     a2 = { a: "def" };
     a2 = {};
     a2 = { a: "def", b: 20 };  // Error
-    ~~
-!!! error TS2322: Type '{ a: string; b: number; }' is not assignable to type '{ a: number; b: number; } | { a: string; b?: undefined; } | { a?: undefined; b?: undefined; }'.
-!!! error TS2322:   Types of property 'b' are incompatible.
-!!! error TS2322:     Type 'number' is not assignable to type 'undefined'.
+                     ~
+!!! error TS2322: Type 'number' is not assignable to type 'undefined'.
+!!! related TS6500 objectLiteralNormalization.ts:11:19: The expected type comes from property 'b' which is declared here on type '{ a: number; b: number; } | { a: string; b?: undefined; } | { a?: undefined; b?: undefined; }'
```
DanielRosenwasser, line 44: Unlike the other regressions, it really feels like we've lost too much context here.

tests/baselines/reference/indirectDiscriminantAndExcessProperty.errors.txt
Andarist: Note that this isn't related to discriminants here. This case just happens to have a discriminant target.

The reason this has changed is that previously `getBestMatchIndexedAccessTypeOrUndefined` would prefer the simple match done using `getIndexedAccessTypeOrUndefined`. Since `type` is a shared property in the union target that would return early and the error with the union type would be raised. Now `getBestMatchingType` is preferred so this new error here is the product of the best overlappy type matching.

This now matches more closely the non-elementwise error reported in the very same situation, see the playground [here](https://www.typescriptlang.org/play?ts=5.5.0-dev.20240226#code/KYDwDg9gTgLgBDAnmYcBCAbAhgCzgXgCg4S4AfOAbwWWAC44AiAMwgkYBo4sAjAYwYBnGFACWAOwDmcAL7FSFakhQNGPLFE5wQiAF4NxAVwC2PYFC6gRWBlnGJZAbkKEAJsD7YoqZofF8YUQhxBBwJSQAKHmwcBkxcAEoGADcIUVdnQgxgeFYIAEYCJjzGZxgwqQjKeRJlejg8-I4a7n5VHGAMDAgAQkZCGQTMvmDhOAgeACsiqha6hkbm0laBJg6u3v65cvCIicmhwiA)

tests/baselines/reference/jsxChildrenArrayWrongType.errors.txt
Andarist: I touched on this [here](https://github.com/microsoft/TypeScript/pull/57537#discussion_r1502163633). Let me know if that makes any sense or if you think I should work more on improving this case.

tests/baselines/reference/objectLiteralNormalization.errors.txt
Andarist: Doesn't the related info contain the same information that was reported previously in the non-elementwise error? What would be your ideal outcome here? It could be nicer if the error would focus on the selected best type and not on the containing union. 

I wouldn't call this bit a regression per se - the same information is still here, it's presented differently though. And, of course, the presentation **does** matter - I'm not claiming that it doesn't.

fatcerberus: FWIW: That error is confusing.  It starts out saying `number` isn't assignable to `undefined`, which is true, but then goes on to claim that the "expected type" comes from `b` of `{ a: number; b: number; } | { a: string; b?: undefined; } | { a?: undefined; b?: undefined; }`.  Now, given that union as the target type, if the `b` in the source is the only problem... why *can't* it be a number?  So the error message kinda-sorta contradicts itself.

Andarist: It makes sense. When rereading it now I see how this is confusing - perhaps the related information should mention the complete source type or it should somehow mention the selected best type based on which this elementwise elaboration is produced.

I'll look into improving this - thanks for the feedback ❤️

Issue #57540, opened 2024-02-26
Wrong completion inside object spread with type annotation


so1ve:
### 🔎 Search Terms

spread complete

### 🕗 Version & Regression Information

- This is the behavior in every version I tried

### ⏯ Playground Link

https://www.typescriptlang.org/play?#code/MYewdgzgLgBAZiEMC8MDeMBGAuGBGGAXwG4AoU0SWTAQwCdc1SBIG3MAVwFtMBTOssxwxOPfmUIp0LNvgA0LAHTKYAelUwAKnQCWAcz38YNDlCSguABwA2vKLxgALfr1IkgA

### 💻 Code

```ts
const foo = { b: 1 };

const bar: {
	a: number;
	b: number;
} = {
	a: 1,
	... // Trigger auto complete here
};
```

### 🙁 Actual behavior

![image](https://github.com/microsoft/TypeScript/assets/58381667/b867edbe-04aa-4af7-89e8-c21cf2e77877)

And if you use this completion, the code will become `...,b`.

### 🙂 Expected behavior

Should complete `foo`, not `b`.

### Additional information about the issue

The type annotation is necessary to reproduce this issue.

PR #57542, opened 2024-02-26
Prefer global over object members completions after `...`

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/57540


jakebailey (approved):
I think this is right?

andrewbranch (approved):

PR #57549, opened 2024-02-26
Add optionality to mapped type indexed access substitutions

### LINKED ISSUE #57487: Regression: Indexed Access of Partial Mapped Type with Union from Generic Function is missing `undefined`

ahejlsberg:
Fixes #57487.

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -14040,10 +14040,14 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         return modifiers & MappedTypeModifiers.ExcludeOptional ? -1 : modifiers & MappedTypeModifiers.IncludeOptional ? 1 : 0;
     }
 
+    function getModifiersTypeOptionality(type: Type): number {
+        return type.flags & TypeFlags.Intersection ? Math.max(...map((type as IntersectionType).types, getModifiersTypeOptionality)) :
+            getObjectFlags(type) & ObjectFlags.Mapped ? getMappedTypeOptionality(type as MappedType) :
+            0;
+    }
+
     function getCombinedMappedTypeOptionality(type: MappedType): number {
-        const optionality = getMappedTypeOptionality(type);
-        const modifiersType = getModifiersTypeFromMappedType(type);
-        return optionality || (isGenericMappedType(modifiersType) ? getMappedTypeOptionality(modifiersType) : 0);
+        return getMappedTypeOptionality(type) || getModifiersTypeOptionality(getModifiersTypeFromMappedType(type));
     }
 
     function isPartialMappedType(type: Type) {
@@ -18517,7 +18521,8 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     function substituteIndexedMappedType(objectType: MappedType, index: Type) {
         const mapper = createTypeMapper([getTypeParameterFromMappedType(objectType)], [index]);
         const templateMapper = combineTypeMappers(objectType.mapper, mapper);
-        return instantiateType(getTemplateTypeFromMappedType(objectType.target as MappedType || objectType), templateMapper);
+        const instantiatedTemplateType = instantiateType(getTemplateTypeFromMappedType(objectType.target as MappedType || objectType), templateMapper);
+        return addOptionality(instantiatedTemplateType, /*isProperty*/ true, getCombinedMappedTypeOptionality(objectType) > 0);
     }
 
     function getIndexedAccessType(objectType: Type, indexType: Type, accessFlags = AccessFlags.None, accessNode?: ElementAccessExpression | IndexedAccessTypeNode | PropertyName | BindingName | SyntheticExpression, aliasSymbol?: Symbol, aliasTypeArguments?: readonly Type[]): Type {
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -14042,7 +14042,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
 
     function getModifiersTypeOptionality(type: Type): number {
         return type.flags & TypeFlags.Intersection ? Math.max(...map((type as IntersectionType).types, getModifiersTypeOptionality)) :
-            getObjectFlags(type) & ObjectFlags.Mapped ? getMappedTypeOptionality(type as MappedType) :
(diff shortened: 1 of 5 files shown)


ahejlsberg:
@typescript-bot test top200
@typescript-bot run dt

src/compiler/checker.ts
```diff
@@ -14040,10 +14040,14 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         return modifiers & MappedTypeModifiers.ExcludeOptional ? -1 : modifiers & MappedTypeModifiers.IncludeOptional ? 1 : 0;
     }
 
+    function getModifiersTypeOptionality(type: Type): number {
+        return type.flags & TypeFlags.Intersection ? Math.max(...map((type as IntersectionType).types, getModifiersTypeOptionality)) :
+            getObjectFlags(type) & ObjectFlags.Mapped ? getMappedTypeOptionality(type as MappedType) :
+            0;
+    }
+
     function getCombinedMappedTypeOptionality(type: MappedType): number {
```
Andarist, line 14049: If I'm not mistaken this currently falls short here:
```ts
type Obj = {
    a: 1,
    b: 2
};

type Identity<T> = { [K in keyof T]: T[K] };

const mapped: {
    [K in keyof Identity<Partial<Obj>>]: Obj[K]
} = {}

const resolveMapped = <K extends keyof typeof mapped>(key: K) => mapped[key].toString(); // should error but it doesnt
```

Some extra handling of homomorphic mapped types might have to be added here.

src/compiler/checker.ts
```diff
@@ -18517,7 +18521,8 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     function substituteIndexedMappedType(objectType: MappedType, index: Type) {
         const mapper = createTypeMapper([getTypeParameterFromMappedType(objectType)], [index]);
         const templateMapper = combineTypeMappers(objectType.mapper, mapper);
-        return instantiateType(getTemplateTypeFromMappedType(objectType.target as MappedType || objectType), templateMapper);
+        const instantiatedTemplateType = instantiateType(getTemplateTypeFromMappedType(objectType.target as MappedType || objectType), templateMapper);
+        return addOptionality(instantiatedTemplateType, /*isProperty*/ true, getCombinedMappedTypeOptionality(objectType) > 0);
```
Andarist, line 18525: Taking known properties into consideration would fix this inconsistency that this PR creates:
```ts
type Obj = {
  a: string;
  b: number;
};

type Obj2 = {
  b: number;
  c: boolean;
};

declare const mapped: {
  [K in keyof (Partial<Obj> & Required<Obj2>)]: number;
};

// displays the same way as `resolved` below!
mapped;
// ^? const mapped: { a?: number | undefined; b: number; c: number; }

const accessMapped = <K extends keyof Obj2>(key: K) => mapped[key].toString();

declare const resolved: { a?: number | undefined; b: number; c: number };

const accessResolved = <K extends keyof Obj2>(key: K) => resolved[key].toString();
```

src/compiler/checker.ts
ahejlsberg: Yeah, looks like we need to recurse in `getCombinedMappedTypeOptionality`.

src/compiler/checker.ts
ahejlsberg: Right, when mapping over intersections constructed from multiple mapped types the PR makes the most conservative assumption about optionality. I think that's an acceptable compromise and I'm concerned with the potential cost of examining every member possibly selected by the key.

src/compiler/checker.ts
```diff
@@ -14040,10 +14040,14 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         return modifiers & MappedTypeModifiers.ExcludeOptional ? -1 : modifiers & MappedTypeModifiers.IncludeOptional ? 1 : 0;
     }
 
+    function getModifiersTypeOptionality(type: Type): number {
```
RyanCavanaugh, line 14043: I don't understand what this is computing. What does a result of `2` mean?

RyanCavanaugh (approved):
Behavior looks good

src/compiler/checker.ts
ahejlsberg: The result is always either -1, 0, or 1. -1 means optionality is stripped (i.e. `-?`), 0 means optionality is unchanged, 1 means optionality is added (i.e. `+?`). When a homomorphic mapped type doesn't modify optionality, we (recursively) consult the optionality of the type you're mapping over to see if it strips or adds optionality.

gabritto: Should this be an enum? Then it'd be easier to tell what the numbers mean.

ahejlsberg: It could be an enum, but `number` feels more appropriate since we use the greater and less than operators to relate the combined values.

gabritto: Would a comment about what each number represents work?

Issue #57539, opened 2024-02-26
unintended escaped string for hover


Hajime-san:
### 🔎 Search Terms

"lsp"

### 🕗 Version & Regression Information

version `"typescript": "5.3.3"`

### 💻 Code

```ts
// 🦁.ts
export const foo = "bar";

// main.ts
import { bar } from './🦁.ts';
```

### 🙁 Actual behavior

The lsp message shows `Module '"./\uD83E\uDD81.ts"' has no exported member 'bar'.ts(2305)`.

### 🙂 Expected behavior

It should be `Module '"./🦁.ts"' has no exported member 'bar'.ts(2305)`.

### Additional information about the issue

<img width="524" alt="スクリーンショット 2024-02-26 10 11 39" src="https://github.com/microsoft/TypeScript/assets/41257923/7c0a935d-5f09-433c-b8af-3b785411b5bd">

Andarist:
This likely falls under the umbrella [here](https://github.com/microsoft/TypeScript/issues/40594) (it's worked on [here](https://github.com/microsoft/TypeScript/pull/49297))

fatcerberus:
@Andarist Wait, is that even related?  That looks like it's about arbitrary import names, e.g.

```ts
import { "😊" as yay } from "foo";
```

...which doesn't work at all today (i.e. it's a feature request), whereas this issue is about unicode characters in module specifiers, which *does* work today, the specifier is just not being rendered property in error messages.

Andarist:
You are totally right. I'm trying to do to much lately and it seems my circuits are overheating :v

Andarist:
My penance: https://github.com/microsoft/TypeScript/pull/57550

PR #39463, opened 2020-07-07
not escape unicode char for import path string.

ShuiRuTian:
Fixes #39027

PR #57550, opened 2024-02-26
Don't escape unicode characters in import/export paths in reported errors

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/57539

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -4324,7 +4324,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     combineValueAndTypeSymbols(symbolFromVariable, symbolFromModule) :
                     symbolFromModule || symbolFromVariable;
                 if (!symbol) {
-                    errorNoModuleMemberSymbol(moduleSymbol, targetSymbol, node, name);
+                    errorNoModuleMemberSymbol(moduleSymbol, targetSymbol, getSourceFileOfNode(node), name);
                 }
                 return symbol;
             }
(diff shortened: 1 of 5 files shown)


src/compiler/checker.ts
```diff
@@ -4324,7 +4324,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     combineValueAndTypeSymbols(symbolFromVariable, symbolFromModule) :
                     symbolFromModule || symbolFromVariable;
                 if (!symbol) {
-                    errorNoModuleMemberSymbol(moduleSymbol, targetSymbol, node, name);
+                    errorNoModuleMemberSymbol(moduleSymbol, targetSymbol, getSourceFileOfNode(node), name);
```
Andarist, line 4327: This relies on the `printer` selection in `symbolToStringWorker`. While the containing function also accepts `node.kind === SyntaxKind.VariableDeclaration`, such a node doesn't reach this line in any test case so I'm currently not 100% sure if I can just call `getSourceFileOfNode` unconditionally here. I think it's OK for imports and exports.

andrewbranch: Misrepresenting the location that we’re going to scope a symbol reference from sounds like a bad idea—the original fix in #39463 that you’re building on here feels a little too magically indirect to begin with to me. Is there any way you see to make the printer escaping behavior more directly dependent on whether a module specifier is being printed?

andrewbranch (changes requested):

PR #57465, opened 2024-02-21
Infer type predicates from function bodies using control flow analysis

danvk:
Fixes #16069
Fixes #38390
Fixes #10734
Fixes #50734
Fixes #12798

This PR uses the TypeScript's existing control flow analysis to infer type predicates for boolean-returning functions where appropriate. For example:

```ts
function isString(x: string | number) {
  return typeof x === 'string';
}
```

This currently has an inferred return type of `boolean`, but with this PR it becomes a type predicate:

image

I filed #16069 seven years ago (!) and thought it would be interesting to try and fix it. It turned out to be cleaner and simpler than I thought: only ~65 LOC in one new function. I think it's a nice win!

## How this works

A function is a candidate for an inferred type guard if:

1. It does not have an explicit return type or type predicate.
2. Its inferred return type is `boolean`.
3. It has a single `return` statement and no implicit returns (this could potentially be relaxed later).
4. It does not mutate its parameter.

If so, then the function looks something like this:

```ts
function f(p: T, p2: T2, ...) {
  // ...
  return expr;
}
```

For each parameter, this PR determine what its flow type would be in each branch if the function looked like this instead:

```ts
function f(p: T, p2: T2, ...) {
  // ...
  if (expr) {
    p1;  // trueType
  }
}
```

if `trueType != T` then we have a _candidate_ for a type predicate.

We still need to check what a `false` return value means because of the semantics of type predicates. If we have:

```ts
declare function isString(x: string | number): x is string;
```

then `x` is a `string` if this function returns `true`. But if it returns `false` then `x` _must_ be a `number`. In other words, in order to be a type predicate, a function should return `true` if _and only if_ the predicate holds.

We can test this directly by plugging in `trueType` to the synthesized `if` statement and seeing what's left in the `else` branch:

```ts
function f(p: trueType, p2: T2, ...) {
  // ...
  if (expr) {
    p1;  // trueType
  } else {
    p1;  // never?
}
```

If it's `never` then we've demonstrated the "only if" condition. 

Note that the previous versions of this PR did slightly different checks. The original version had [problems with subtypes](https://github.com/microsoft/TypeScript/pull/57465#issuecomment-1957867140) and my initial fix made more calls to `getFlowTypeOfReference` than was necessary. This version directly tests the condition that we want.

## Wins

TypeScript is now able to infer type guards in many places where it's convenient, e.g. calls to `filter`:

```ts
const nums = [12, "foo", 23, "bar"].filter(x => typeof x === 'number');
[...]

PR #57552, opened 2024-02-26
Single-expression only type predicates

RyanCavanaugh:
Variant on #57465 with tighter rules for inference:

 * Must have exactly one parameter
 * Must be an expression-bodied function, or have a body consisting of a single return with expression
 * May be a getter

Hoping this gets perf cost down to ~0% while still getting ~all of the value


RyanCavanaugh:

RyanCavanaugh:

danvk:
@RyanCavanaugh Happy to see this caught your interest 😊

My suspicion is that at least some of the slowdown in Compiler-Unions is due to the extra work that happens indirectly because of the new type guards themselves, rather than the direct new work of determining which functions are type guards.

I found this sort of "ablation test" useful in teasing these two apart:
https://github.com/danvk/TypeScript/compare/infer-type-predicate-16069...danvk:TypeScript:ablation-2024-02-26

i.e. do all the work, but don't actually infer the type predicate.

With my PR as of last week I got results that made sense: ~4% slowdown on the PR, ~1.7% slowdown on the ablation → ~2.3% of the slowdown is due to more elaborate type checking.
With the latest version of my PR the results made less sense: ~1.8% slowdown vs. ~1.75% slowdown on the ablation → almost no slowdown is due to extra type checking? But it's inferring all the same type predicates… 🤷

RyanCavanaugh:
Not just from a perf perspective, but my current take (subject to additional thinking) is that this is also a correctness  / DX improvement. Any function that does anything _but_ the type guard itself is likely to either a) have additional constraints that mean it's not really a type guard or b) only appear to guard as a sort of side effect of implementation. But I want to think about that more

RyanCavanaugh:
Newest commit no-ops the result as suggested above (for perf testing only)

RyanCavanaugh:

danvk:
Running the two versions against https://github.com/microsoft/TypeScript/pull/54148 and `tsc` itself (self-check), this is the only dropped predicate that I see (in `src/compiler/transformers/es2018.ts`):

```
    function isSuperContainer(node: Node) {
        const kind = node.kind;
        return kind === SyntaxKind.ClassDeclaration
            || kind === SyntaxKind.Constructor
            || kind === SyntaxKind.MethodDeclaration
            || kind === SyntaxKind.GetAccessor
            || kind === SyntaxKind.SetAccessor;
    }
```

Definitely a legitimate inference, and maybe a little confusing that you'd change the return type by factoring out a local variable. But not a huge loss.

RyanCavanaugh:
> From a DevX perspective it is super annoying because you can't do small stylistic refactors like you mentioned like extracting a variable. So you ended up doing stuff like chaining guard functions to work around the limitation.

I don't think TS really has this problem since you're not *forced* to infer the predicateness of the function; you can just write `isBar(x: unknown): x is Bar`.

bradzacher:
> I don't think TS really has this problem since you're not forced to infer the predicateness of the function; you can just write isBar(x: unknown): x is Bar.

That's entirely true. FWIW flow also supports [TS style type guard annotations now](https://flow.org/en/docs/types/type-guards/) (but safer than TS cos flow validates the predicate matches the refinement).

The thing that I'd be scared about as a user is "actions at a distance". Eg the small refactor of one function causes type errors in other files.

Eg if your codebase relies upon inferred return types you could end up with a really spooky action at a distance as a refined variable changes type which changes the inferred return type of a function which causes a type error.
If that inferred return type function is in a library - it could instead mean an unintended breaking API change.

RyanCavanaugh:
> The thing that I'd be scared about as a user is "actions at a distance". Eg the small refactor of one function causes type errors in other files.

That's a pretty good point and makes me think more about the expression vs declaration distinction -- function expressions are almost always "local", whereas declarations are much more likely to be exported with an inferred type. People might not realize they're exporting a type predicate and it might not be their intent.

craigphicks:
@bradzacher - *There's probably some scope for a lint rule to catch it (though it'd be a first of its kind and I'm not sure if it's fully possible) but it'd be cool if TS had more advanced rules.*

@RyanCavanaugh  - *People might not realize they're exporting a type predicate and it might not be their intent.*

I think what Brad is saying is that "inference" can be turned around to be a confirmation of the users declared `typePredicate`, to catch cases where a mismatch was not intended.  Using it that way wouldn't cause the problem that Ryan mentions.  

Brad is also saying use it like a lint rule, avoiding the performance issue, at least while editing.  

I wonder if this functionality (inverted to be a typePredicate checker, not inferr-er) would be hard to seperate out as a seperate API function for use by lint?  

Maybe hard to separate it out from a whole compile.  In that case how adding a command line flag to allow a super check, set only to coinciding with the timing that lint would be used before or afterwards anyway.  A super check could include typePredicate checking, or any similar expensive addional checking.  

Having to add `@ts-ignore` (or something) for a typePredicate that wasn't expected to pass the check would also be a useful form of documentation showing the users intention.

PR #57560, opened 2024-02-27
test only, ignore

RyanCavanaugh:
Not for merging, just lets us see what inferrable predicates are out there in the wild


RyanCavanaugh:
@typescript-bot test top200

jakebailey:
Seems like a lot of these are explicitly annotated guards?

Issue #57568, opened 2024-02-27
Design Meeting Notes, 2/27/2024


DanielRosenwasser:
# Triple-Slash Reference Emit/Preservation (motivated by `--isolatedDeclarations`)

https://github.com/microsoft/TypeScript/pull/57440

* One stance we had with `--isolatedDeclarations` is that the flag should not affect emit itself.
* One question around this: what do we do with `/// <reference >` directives?
* When doing declaration file emit, TypeScript will try to insert 
* Let's say you have a statement like `export * as fs from "fs"`.
    * TypeScript will emit a triple-slash reference to say `/// <reference types="node" />`
* One idea for `--isolatedDeclarations` is to say that *when* TypeScript would do this, it should emit an error. In other words, the user has to write it out explicitly.
* Problem with this is that TypeScript occasionally *removes* these directives as well when they're not needed.
* This all feels pretty complicated.
* Suggestion: preserve triple-slash references when they're written, never generate them.
    * Thinking we could do this in 6.0.
* The other option would be an attribute on `///` references that allows people to control whether the reference is preserved, elided, or whatever.
* Do we know how often we generate these in the top200 repos?
    * No
* Until recently, preservation wasn't perfect anyway - so maybe people aren't relying on it?
* One strong preference: what you see is what you get.
* If you inserted one of these references, you had to make sure you had the right types installed. The difference is a UX issue of what makes the error more obvious.
* Can view `///` refs as a holdover from a time long-past. Don't ascribe them more value than they deserve!
* What about the non-types references?
    * `/// <reference types="..." />`
    * `/// <reference paths="..." />`
* Do we synthesize?
    * types: currently sometimes yes.
    * paths (aka file references): hopefully no?
* Do we elide?
    * types: currently also sometimes yes.
    * paths (aka file references): hopefully no?
* Do we generate `/// <reference paths="..." />`
* So we want to make these changes. But when?
    * One idea floated was to make sure that this only kicks in within `--isolatedDeclarations`, and gets turned on for everyone in 6.0
* If we're not going to synthesize references, it might be appropriate to tell library authors that they need to preserve that?
* A little gun-shy on including this in 5.5.
    * `--verbatimReferenceDirectives`, require this in `--isolatedDeclarations`, change it in 6.0...
        * And then deprecate `--verbatimReferenceDirectives` in 6.0? 😂
    * We'd like to see how common this is.
* Nightly?
    * Nobody is going to publish on nightly!
[...]

PR #57569, opened 2024-02-27
DO NOT MERGE: Show synthesized triple slash references in diagnostics for top200 analysis

andrewbranch:
Experiment for follow-up discussions from #57440, #57568


andrewbranch:
@typescript-bot test top200

jakebailey:
Quite surprised about react showing up that much, given the files I opened all contained an import of react anyway...

dragomirtitian:
> Quite surprised about react showing up that much, given the files I opened all contained an import of react anyway...

Yes that is really surprising. For `path` references there is code that specifically excludes references to imported files. Guess for type reference directives that logic is missing:

https://github.com/microsoft/TypeScript/blob/main/src/compiler/transformers/declarations.ts#L580-L586 

```ts
if (exportedModulesFromDeclarationEmit?.includes(file.symbol)) {
    // Already have an import declaration resolving to this file
    return;
}
```

andrewbranch:
Looking at these, I don’t feel too worried about the volume of real-world breaks being very high. The only thing I might send a preemptive PR for is adding an explicit reference to `"../../env.d.ts"` to [vercel/swr](https://github.com/vercel/swr/blob/b103154bd0003861933d12d1de286d31acd2104d/src/core/index.ts#L1). That said, knowing you need to do that still feels like a footgun. swr’s tsconfig.json explicitly includes `env.d.ts`. That could be considered bad practice now, as it sets up a program graph that a library consumer would never have. Libraries should probably set their `files` array to be just their package.json entrypoints. But even that’s unrealistic, since a consumer might not load _all_ entrypoints. Ideally, you’d want to make sure your library code compiles when setting each package.json entrypoint as the lone tsconfig.json `files` entry. But nobody is going to do that.

PR #57570, opened 2024-02-28
Set `impliedNodeFormat` in every module mode

andrewbranch:
This is a starting point for substantive changes, only up in this state to test perf and breaking changes.


andrewbranch:
@typescript-bot perf test
@typescript-bot test top200

andrewbranch:
@typescript-bot perf test

andrewbranch:
@typescript-bot test top200

jakebailey:
The pipeline actually failed to push due to the text being too long; will try and fix that, but it was about to send a looot of comments: https://typescript.visualstudio.com/TypeScript/_build/results?buildId=160202&view=logs&j=15ac8a4d-b341-5815-5352-14d7ae9c5a86&t=b4a5e2cd-8727-5d20-e3d3-86822d3c46a6

andrewbranch:
@typescript-bot test top100

src/testRunner/unittests/tsc/projectReferences.ts
```diff
@@ -49,4 +49,49 @@ describe("unittests:: tsc:: projectReferences::", () => {
...
+                    references: [
+                        { path: "../lib" },
+                    ],
+                }),
+                "/app/src/local.ts": "export const local = 0;",
+                "/app/src/index.ts": `
+                    import local from "./local"; // Error
+                    import esm from "esm-package"; // Error
+                    import referencedSource from "../../lib/src/a"; // Error
+                    import referencedDeclaration from "../../lib/dist/a"; // Error
```
andrewbranch, line 92: @sheetalkamat do you think it’s worth it to fix this? These two default imports would ideally error, because we could know by using the referenced project settings that the emit was ESM. We currently don’t error because we assume that a declaration file that says `export declare const a = 0` could represent a CJS file. I’m not sure if it should be out of scope for this PR to start using referenced project settings to get more accurate information during type checking like this.

andrewbranch:
@typescript-bot test top200

andrewbranch:
@typescript-bot test top200

PR #57579, opened 2024-02-28
Ensure comments before // @Filename are not part of the following virtual test file

andrewbranch:
I wanted to see how using `JSON.parse` for package.json in module resolution would affect perf, but when I did that, many tests failed because the test package.json file had leading comments when set up like:

```ts
// Some commentary about this test

// @Filename: package.json
{}
```

The virtual file content was:

```
// Some commentary about this test

{}
```

This caught a handful of test cases that were not set up as the author intended.


tests/cases/compiler/enumAssignmentCompat6.ts
```diff
@@ -1,4 +1,4 @@
-// @filename a.ts
+// @filename: a.ts
```
andrewbranch, line 1: Apparently this didn't matter, but 😱

jakebailey (approved):

PR #57580, opened 2024-02-28
Fixed an out-of-order quick info issue with contextual rest parameter

Andarist:


tests/baselines/reference/completionDetailsOfContextSensitiveParameterNoCrash.baseline
```diff
@@ -85,7 +85,7 @@
 //         return curry(getStylingByKeys, 2)(mergedStyling, ...args);
 //                                                             ^^^^
 // | ----------------------------------------------------------------------
-// | (parameter) args: any
+// | (parameter) args: []
```
Andarist, line 88: this is the correct display that we can see when hovering at this location in the playground (after the playgrounds gets typechecked first): [TS playground](https://www.typescriptlang.org/play?ts=5.5.0-dev.20240228#code/C4TwDgpgBA+jUF4oDsIDcICcDcBYAUAQJbLBYBmAhgMbQDCArppkRACYBiDy1wRA9sgCMAHgAqQgDRQASgD4oAbwJRVUABQBKAFxRGzVp268Bw8VNly8+NRuBDdEnbOsBfYqQo16TFuy48fIIATObSYsHS8koqalq6+n5GgaahEuGRlta26vaOQs6JhgEmgqIRUVaxqrkOsDDSwMGOwYW+xcZBZulZ1XZ1PU0tzjJuBB5kmFS0eu3+naYAzGFQFauLlTE2cW0G8ymCyz1rYhu92zV5qwUJc8mlyGmZp5XZcVdwjc2rrbd79100hYXuccldBt8IrskiUuuUzvI3pc6p8oEN6o1Fo5FtCOgdHis1oi+rV8l9dKjgFj1rj9g9ypliRd+hSGmjIZkqdjaQDTOULEywQMLOi1lyabpRgR3IR8CRJtMfP9YaYACyE55nMSqzbKZnxWbKhaCdXHTXhHWg97Cnkqw7ic2rS2C62s8k-W3G5BHYFa51VfXgkUcz34tJ+14kj5s9GU6mnUMPH0ZC2RwMomPfOPcv4wr1PVNW5Fu9ktTE5w15sOE-1IlnXd1i+M43N4+niWtRjPu7MYtGqxyqxNdZM-cIIgNCsmlvviykDp3D1IO8dpqezjnl2cL7VLkKEid10kbkvikHAHdD1t0wEateuhszpuny977qHrvT2Mx5uNV-X3kyhXIt60pTc0V-ftBzffl72LR9RU5SCL2gyUxllNgIGoAAbShMGgNA8KgahfBAXQ9VsFZ5HUchjF0Y8nEQBQZGkPCiFAAB+XRkAYABbAAjLAYKoqxbD6O9LBouj6whYYmNkViWE47i+MEzA3yBFMQIkkFqNonh6KDRskO5eSWKgNjlJQVShIAu1vR0j9mUcws9OkhjgzLCDsT-aCzMU9iQC46yBNsys2y6U1gUdbU4I0fTqHogA6FK8IAcwAZ10ShkBAABtABdTR5JykAAqsnjQvU5LUswTLstywrioQBRSrrMBcNoAALfhsMwzBWTcawCAgAAPMB+EwYA0XAaAAGVQGwkg0vsxAND6ABrCAQCyjQMuAFhkDSqAAB8oCobCMugM7uEw8gSHYYqzvUfbDuOs6Lqu06oFuiB7tQNhNEKyQ+hSpL8P23RuA25B+AAd2QQqCGahR+H4gArLDgGG-BMJwvDoGoQR9qgNKIGABaQCWo6ACEQAAaW23b1D6XisDJthKeptLdDRzHeBB5ktp2+jXuW77Puun7kDuh7Ae+l6DvFj7KEuqXfv+x7gdB2r6ulmH4cRgrkfkvmsZxvHcPwqBCMwKA2bq9gueW3mMfN8Z8DGiapuI4npuofDKDIZ2jrsr1lhJfjKCuoQADYxC6iA2dd-ngBR-XYYRwXbDN3hvo1uXs7UXPppumW-sLvoQ5Wr0CAUJASOYEAREj6OIDjhOk4gFOsfT6HM+QIvVBL-Py81tgh6gEey9lgHJ+r+y65Z-U+lsMmKcW5aOEwfheJptu4-oqOY-jxPk6nt3eD75ADaz1fi7ALpdpHpBFFcSfbGAM+IAAeUwaue551fu-e+qgwbpQyn0ZwC8vTyQorYNQ+FgBMGQMRUi6h17VzpozHa0hWjqAdhzau0gwYAHoABU5DSEQM0HWEBzJFjI2wEAA). If we make an edit and quickly hover over this location we can sometimes "catch" this `any` being displayed since then the quick info request happens before computing semantic diagnostics and this is exactly what this test is doing.

src/compiler/checker.ts
```diff
@@ -49585,7 +49590,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     return grammarErrorOnNode(parameter.name, Diagnostics.A_rest_parameter_cannot_have_an_initializer);
                 }
             }
-            else if (isOptionalParameter(parameter)) {
+            else if (hasEffectiveQuestionToken(parameter)) {
```
Andarist, line 49593: `isOptionalParameter` calls `getMinArgumentCount` for parameters with initializers and that in turn calls `getTypeOfSymbol` on rest parameters. This led to `anyType` being returned by `reportCircularityError` since the circularity was caused by this `getTypeOfSymbol` called again for something that was just being computed through out-of-order request

I spent some time on this and to the best of my knowledge - this particular function (`checkGrammarParameterList`) doesn't even have to consult that type. A parameter with an initializer is always treated as required (unless it's last) - and the main purpose of checking this here is to set `seenOptionalParameter` to potentially error on a required parameter that follows an optional one. This still happens but this error is mainly based on the syntactic~ optionality so it should be OK to just call `hasEffectiveQuestionToken` here.

I think that avoiding this extra type resolution from a grammar check function should be preferred - but I didn't recheck with all of the other existing grammar checks to see if this is a hard rule today or not.

src/compiler/utilities.ts
```diff
@@ -10518,7 +10517,7 @@ export function canHaveExportModifier(node: Node): node is Extract<HasModifiers,
 }
 
 /** @internal */
-export function isOptionalJSDocPropertyLikeTag(node: Node): node is JSDocPropertyLikeTag {
+export function isOptionalJSDocPropertyLikeTag(node: Node): boolean {
```
Andarist, line 10520: If I'm not mistaken this function wasn't proving that the input is `JSDocPropertyLikeTag` - some extra checks on a variable with `JSDocPropertyLikeTag` type were done here and it could still return `false` for it. So this wasn't a correct return type annotation

jakebailey: It's certainly not the first type guard we've had which has lied for convenience's sake. But it doesn't seem like it really helped here.

jakebailey (approved):
Seems correct to me?

src/compiler/checker.ts
sandersn: checkGrammar* should in general not get types. So I think this is an improvement.

> A parameter with an initializer is always treated as required

Shouldn't it be treated as optional by default?

sandersn (approved):

sandersn:

src/compiler/checker.ts
Andarist: > Shouldn't it be treated as optional by default?

I worded this terribly here. I think I've meant that it's always treated as required when followed by a required parameter:
```ts
function fn1(a: string, b = '', c = 0) {} // b is optional
function fn2(a: string, b = '', ...args: any[]) {} // b is optional
function fn3(a: string, b = '', c: number) {} // b is required
```

jakebailey:
I broke the perf pipeline for PRs accidentally, rerunning

Issue #57582, opened 2024-02-29
Internal crash with `TypeError: Cannot read properties of undefined (reading 'parent')`


geritol:
### 🔎 Search Terms

Project works fine with typescript version `4.8.4`, when upgrading to `4.9.4` (and any further version) type checking crashes with an internal error.

### 🕗 Version & Regression Information

- This is a crash
- This changed between versions 4.8.4 and 4.9.4
- Error is present with `typescript@next` (`Version 5.5.0-dev.20240229`)

With the help of every-ts bisect I found the first bad commit to be:
```
42f9143e114c5c07f40df83ed07ffeb3cbaf2101 is the first bad commit
commit 42f9143e114c5c07f40df83ed07ffeb3cbaf2101
Author: Jack Works <jackworks@protonmail.com>
Date:   Fri Sep 30 07:13:25 2022 +0800

    feat: codefix for `for await of` (#50623)

 src/compiler/checker.ts                                | 18 +++++++++++++++++-
 src/compiler/types.ts                                  |  1 +
 src/services/codefixes/addMissingAwait.ts              |  9 +++++++++
 .../fourslash/codeFixAddMissingAwait_forAwaitOf.ts     | 17 +++++++++++++++++
 4 files changed, 44 insertions(+), 1 deletion(-)
 create mode 100644 tests/cases/fourslash/codeFixAddMissingAwait_forAwaitOf.ts
```

https://github.com/microsoft/TypeScript/pull/50623/files

I _think_ error might come from `errorNode` being undefined here:
<img width="718" alt="image" src="https://github.com/microsoft/TypeScript/assets/5496681/030c263b-7202-4f28-a11b-e28ea376dd1e">

### 🙁 Actual behavior

```
> tsc --noEmit

[...]/node_modules/typescript/lib/tsc.js:99607
                throw e;
                ^

TypeError: Cannot read properties of undefined (reading 'parent')
    at reportTypeNotIterableError ([...]/node_modules/typescript/lib/tsc.js:71494:51)
    at getIteratedTypeOrElementType ([...]/node_modules/typescript/lib/tsc.js:71168:17)
    at checkIteratedTypeOrElementType ([...]/node_modules/typescript/lib/tsc.js:71163:20)
    at getTypeOfDestructuredArrayElement [...]/node_modules/typescript/lib/tsc.js:59574:50)
    at getAssignedTypeOfArrayLiteralElement ([...]/node_modules/typescript/lib/tsc.js:59599:20)
[...]

PR #57586, opened 2024-02-29
Fixed a crash when computing flow type of destructured never

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/57582


jakebailey (approved):
Good that another error still fires for this case; I was worried it wouldn't.

PR #57589, opened 2024-03-01
[ID-Prep] PR4 - Preserve types nodes from type assertions in declaration emit

### LINKED ISSUE #57587: [ID-Prep] Preserve type nodes from type assertions in declarations

### 🔍 Search Terms

declaration emit preserve types assertions

### ✅ Viability Checklist

### ⭐ Suggestion

Preserve types from type assertions when emitting declarations.

### 📃 Motivating Example

```ts
// index.ts
export let p = null as P

// index-current.d.ts
export declare let p: {
    foo: boolean;
};

// index-proposed.d.ts
export declare let p: P;

```

### 💻 Use Cases

    An external tool can extract the type from the assertion, but then they would have different declarations. It would be great if, where possible, typescript would also preserve types as written.
    We could force users to always specify the type even when it's obvious from the assertion, but this makes for very frustrating DX

dragomirtitian:
Fixes #57587


tests/cases/compiler/hugeDeclarationOutputGetsTruncatedWithError.ts
```diff
@@ -3,4 +3,4 @@ type props = "a" | "b" | "c" | "d" | "e" | "f" | "g" | "h" | "i" | "j" | "k" | "
 
 type manyprops = `${props}${props}`;
 
-export const c = null as any as {[K in manyprops]: {[K2 in manyprops]: `${K}.${K2}`}};
\ No newline at end of file
+export const c = () => null as any as {[K in manyprops]: {[K2 in manyprops]: `${K}.${K2}`}};
```
dragomirtitian, line 6: Changed this to preserve the original intent of the test (test what happens when huge type nodes need to be printed).

tests/baselines/reference/accessorDeclarationEmitVisibilityErrors.errors.txt
```diff
@@ -1,12 +1,9 @@
...
 
 
-==== accessorDeclarationEmitVisibilityErrors.ts (2 errors) ====
+==== accessorDeclarationEmitVisibilityErrors.ts (1 errors) ====
     export class Q {
         set bet(arg: DoesNotExist) {}
                      ~~~~~~~~~~~~
 !!! error TS2304: Cannot find name 'DoesNotExist'.
-                     ~~~~~~~~~~~~
-!!! error TS4106: Parameter 'arg' of accessor has or is using private name 'DoesNotExist'.
```
jakebailey, line 11: This is unrelated to type assertions, no? Does this PR contain changes for the getter/setter stuff too?

src/compiler/transformers/declarations.ts
```diff
@@ -282,6 +296,7 @@ export function transformDeclarations(context: TransformationContext) {
     const resolver = context.getEmitResolver();
     const options = context.getCompilerOptions();
     const { noResolve, stripInternal } = options;
+    const strictNullChecks = getStrictOptionValue(options, "strictNullChecks");
```
jakebailey, line 299: Feels weird to have this detail in the transform; right now the only options that are used are ones explicitly related to declaration emit (noResolve/stripInternal).

My impression is that isolatedDeclarations is not assumed to know the compiler options; is the case that we can treat strictNullChecks as true for emit with the assumption that `undefined`/`null` are just going to be ignored by non-strict users?

dragomirtitian: There is a bit of `strictNullChecks` that needs to be taken into account. This is used here in the context of inferring optional property declaration types. If `strictNullChecks` is on we need to add `undefined` to the type otherwise we don't. Without taking this into account any local inference optional declarations will not be possible:

```ts
class X { 
	// x?: 1 | 2 | undefined; with !strictNullChecks
	// x?: 1 | 2; // with !strictNullChecks
	x? = 1 as 1 | 2 
}
```
[Playground Link](https://www.typescriptlang.org/play?#code/MYGwhgzhAEAa0G9oFgBQBIA9J6APA-AFzQCM0APtAEwXQCuAdgCYCmAZgJYMtMDc0Adw4AXABbQAhBGEAnDsGEA5OiBABhUS2ABrCGiw4CxMpSr9sgkeKmz5Sles069GAtAC8paJC+mUqAF80IA)

jakebailey: Sure but if a user doesn't have strictNullChecks enabled, aren't the two equivalent? Since the way that strictNullChecks=false works is to effectively treat null/undefined as though they don't exist.

dragomirtitian: They are usually, but if you take the declaration from a project with `!strictNullChecks` and use them in a project `exactOptionalPropertyTypes` then the missing `| undefined` will make a difference.

dragomirtitian: Thinking about it, it might actually be an improvement. Because a `!strictNullChecks` project should be fine with a `undefined`. 

I can change this to always add `| undefined` but the second problem is that it will make emit different between the case when we take the type from the checker (which does not add undefined) and when we try to get the type from the assertion. The solution would be to add the `|undefined` always form the checker as well. 

Maybe this should be a different PR though? One that always adds `|undefined` to optional properties even if `!strictNullChecks`

tests/baselines/reference/accessorDeclarationEmitVisibilityErrors.errors.txt
dragomirtitian: The declared accessor type was not visited. Fixed.

src/compiler/transformers/declarations.ts
RyanCavanaugh: > One that always adds `|undefined` to optional properties

` | undefined` is not implied by optional properties under `exactOptionalPropertyTypes`, so this would be unsafe

jakebailey: I guess I see what you mean, in that this PR is only trying to touch assertions and therefore making this code unconditional would not change any other declarations?

dragomirtitian: Exactly. Even it's final form, `!isolatedDeclarations` some types will still come from the checker, I don't think adding undefined just sometimes would be a good thing.

src/compiler/transformers/declarations.ts
```diff
@@ -725,45 +741,197 @@ export function transformDeclarations(context: TransformationContext) {
...
+                    break;
+                case SyntaxKind.BindingElement:
+                    typeNode = inferVariableLikeType(node);
+                    break;
+                case SyntaxKind.ConstructSignature:
+                case SyntaxKind.PropertySignature:
+                case SyntaxKind.MethodSignature:
+                case SyntaxKind.CallSignature:
+                    declaredType = node.type;
+                    typeNode = declaredType ? visitTypeNode(declaredType) : factory.createKeywordTypeNode(SyntaxKind.AnyKeyword);
```
rbuckton, line 800: `visitNode` allows `node` to be undefined. If you add an overload to `visitTypeNode` that allows undefined, this could simply be:

```suggestion
                    typeNode = visitTypeNode(node.type) ?? factory.createKeywordTypeNode(SyntaxKind.AnyKeyword);
```

src/compiler/transformers/declarations.ts
```diff
@@ -725,45 +741,197 @@ export function transformDeclarations(context: TransformationContext) {
...
+                    break;
+                case SyntaxKind.MethodDeclaration:
+                case SyntaxKind.FunctionDeclaration:
+                    declaredType = node.type;
+                    typeNode = declaredType ? visitTypeNode(declaredType) : inferReturnTypeOfSignatureSignature(node);
```
rbuckton, line 805: As above:
```suggestion
                    typeNode = visitTypeNode(node.type) ?? inferReturnTypeOfSignatureSignature(node);
```

src/compiler/transformers/declarations.ts
```diff
@@ -725,45 +741,197 @@ export function transformDeclarations(context: TransformationContext) {
...
-            getSymbolAccessibilityDiagnostic = oldDiag!;
+            errorNameNode = node.name;
+            errorFallbackNode = node;
+            let oldDiag;
+            if (!suppressNewDiagnosticContexts) {
+                oldDiag = getSymbolAccessibilityDiagnostic;
+                getSymbolAccessibilityDiagnostic = createGetSymbolAccessibilityDiagnosticForNode(node);
+            }
+            let typeNode: TypeNode | undefined;
+            let declaredType: TypeNode | undefined;
```
rbuckton, line 778: This only exists to hold what amounts to be a temporary variable, and only in two of the branches below. I would suggest the minor refactors below so we can remove this.

src/compiler/transformers/declarations.ts
```diff
@@ -725,45 +741,197 @@ export function transformDeclarations(context: TransformationContext) {
...
+                getSymbolAccessibilityDiagnostic = oldDiag!;
+            }
+            return typeNode;
         }
-        return typeNode ?? factory.createKeywordTypeNode(SyntaxKind.AnyKeyword);
-    }
 
+        function visitTypeNode(type: TypeNode) {
+            return visitNode(type, visitDeclarationSubtree, isTypeNode)!;
+        }
```
rbuckton, line 824:
```suggestion
        function visitTypeNode(type: TypeNode): TypeNode;
        function visitTypeNode(type: TypeNode | undefined): TypeNode | undefined;
        function visitTypeNode(type: TypeNode | undefined) {
            return visitNode(type, visitDeclarationSubtree, isTypeNode);
        }
```

src/compiler/transformers/declarations.ts
```diff
@@ -725,45 +741,197 @@ export function transformDeclarations(context: TransformationContext) {
             // Literal const declarations will have an initializer ensured rather than a type
             return;
         }
-        const shouldAddImplicitUndefined = node.kind === SyntaxKind.Parameter && resolver.requiresAddingImplicitUndefined(node);
-        if (type && !shouldAddImplicitUndefined) {
-            return visitNode(type, visitDeclarationSubtree, isTypeNode);
+        return typeFromDeclaration(node);
+
+        function inferVariableLikeType<T extends VariableLikeDeclaration>(node: T, addUndefined?: boolean) {
```
rbuckton, line 746: Do all of these functions need to close over `node` and `ignorePrivate`, or could we refactor them out of `ensureType` to avoid the overhead of creating a closure? `ensureType` is called frequently in declaration emit and closures aren't free, so I'd rather we not pay the cost of we can avoid it.

src/compiler/transformers/declarations.ts
```diff
@@ -725,45 +741,197 @@ export function transformDeclarations(context: TransformationContext) {
...
+                return canAddUndefined((node as ParenthesizedTypeNode).type);
+            }
+            if (node.kind === SyntaxKind.UnionType || node.kind === SyntaxKind.IntersectionType) {
+                return (node as UnionTypeNode | IntersectionTypeNode).types.every(canAddUndefined);
+            }
+            return false;
+        }
+        function typeFromExpression(node: Expression, requiresUndefined?: boolean): TypeNode | undefined {
+            switch (node.kind) {
+                case SyntaxKind.AsExpression:
```
rbuckton, line 868: I think you're missing a `case` here

```suggestion
                case SyntaxKind.AsExpression:
                case SyntaxKind.TypeAssertion:
```

tests/cases/compiler/verbatim-declarations-assertions.ts
rbuckton, line 1: We should have tests that use TypeScript's `<T>x` assertion syntax as well.

rbuckton (changes requested):

src/compiler/transformers/declarations.ts
dragomirtitian: Yup. Dropped it accidentally when moved code over.

tests/cases/compiler/verbatim-declarations-assertions.ts
dragomirtitian: Added

tests/baselines/reference/duplicatePropertiesInTypeAssertions01.js
```diff
@@ -10,4 +10,5 @@ var x = {};
 //// [duplicatePropertiesInTypeAssertions01.d.ts]
 declare let x: {
     a: number;
+    a: number;
```
rbuckton, line 13: This seems wrong. What caused this change?

tests/baselines/reference/duplicatePropertiesInTypeAssertions02.js
```diff
@@ -10,4 +10,5 @@ var x = {};
 //// [duplicatePropertiesInTypeAssertions02.d.ts]
 declare let x: {
     a: number;
+    a: number;
```
rbuckton, line 13: This as well.

tests/baselines/reference/duplicatePropertiesInTypeAssertions01.js
dragomirtitian: The type in the type assertion, as written in source has a duplicate property. TypeScript will error on it.

```
let x = <{a: number; a: number}>{};
```

tests/baselines/reference/duplicatePropertiesInTypeAssertions02.js
dragomirtitian: Same :

```ts
let x = {} as {a: number; a: number};
```

tests/baselines/reference/duplicatePropertiesInTypeAssertions01.js
rbuckton: Does this affect declaration emit from a JS file? The JS standard allows you to specify a property twice, but the last one wins. I'm concerned this would cause valid JS to produce an invalid declaration file that would have previously been valid, and generating declarations from JS is one of the ways we suggest users produce declaration files for DefinitelyTyped.

@jakebailey, @danielrosenwasser: what are your feelings on this?

rbuckton: In other words, does this affect

```js
const x = /** @type {{ a: number, a: number }} */(foo);
```

and is that something we should be concerned about.

rbuckton: Also, since we have tests for `<T>x` and `x as T` in TS, do we need coverage for `/** @type {T} */(x)` in JS in general?

dragomirtitian: This should not impact declaration emit from JS files. That would be something we might work on after the initial isolated declaration emit.

rbuckton: > This should not impact declaration emit from JS files.

Could you add a test to verify this?

dragomirtitian: Added a test that uses assertions in JSDoc. The type is not preserved.

rbuckton (approved):

jakebailey (approved):
I still don't feel too great about declaration emit directly knowing about strictNullChecks as I'm not sure how we are expecting third party emitters to also know this, but, I guess it's fine.

jakebailey:
I think the reason I think it's "fine" is that dependence is already present today, just that it's hidden by calls to the checker. So as a whole, this PR is just an emit improvement and not actually regressing in that way. Third party tools are already able to resolve tsconigs, I just think it's frustrating that declaration emit is sensitive to it at all (but I'm also in the camp of wanting to ditch strictNullChecks=false altogether!)

dragomirtitian:
> > I still don't feel too great about declaration emit directly knowing about strictNullChecks as I'm not sure how we are expecting third party emitters to also know this, but, I guess it's fine.
> 
> That's a good point. Aside from `strictOptionalProperties`, what would be the harm in adding `| undefined` regardless of `strictNullChecks`? In `strictNullChecks: false` we would ignore the `undefined` anyways.

My biggest issue is that it will be added *sometimes*. When the type is extracted in declarations, we will add `undefined` but when it comes from the type checker it will not be added:

```ts
type P = number | undefined
declare function getP(): P
class X {
    f1? = getP() as P; // P | undefined  with this PR
    f2? = getP(); // P
}
``` 
This seems like a strange difference that will make people open issues.

I think our options are: 
1. Always add ` | undefined` to optional fields, even if the type is inferred - A big emit change.
2. Only add it for type that come from the declaration emit - An inconsistent emit change.
3. We can just let go of this and not extract types for optional fields in declaration emit (and make it an error in ID) - Worse DX in ID
4. Allow the declaration emit to know about `strictNullChecks`.

I am not 100% we can keep `strictNullChecks` completely out of declaration emit in future PRs. When we add inference for objects, something like `let o = { n: null, u: undefined }` is typed as `{  n: any;  u: any; }` without `strictNullChecks`   and `{ n: null, u: undefined }` with it so. So the declaration emit will need to know what type to use there (unless we ban `null` and `undefined` there too)

rbuckton:
We'll discuss this PR in the design meeting on Friday and assuming there are no other concerns around `strictNullChecks`, should be able to merge then.

rbuckton:
@typescript-bot pack this

weswigham (changes requested):
IMO, this is something we could/should have done even without isolated declarations, and we've done things like this in the past. We already have machinery for cloning/reusing input type nodes in new locations _inside_ the emit resolver (usually for signatures copied from one location to another by structural inlining) - this logic should be moved into it, too, so all the guardrails around node copying are in one place (and there are a fair few). Over in `createTypeOfDeclaration` and `createReturnTypeOfSignatureDeclaration` they both defer to `nodeBuilder.typeToTypeNode` - within the node builder, we have a `serializeExistingTypeNode` (and parent function `serializeTypeForDeclaration` used extensively by the JS declaration emitter) function that adds a lot of guard rails onto this node copying process, because, well, there's a lot that needs to be checked for that currently isn't in this PR.

IMO, this logic should be exposed on the `nodeBuilder` and then used by the `EmitResolver`'s `createTypeOfDeclaration` and `createReturnTypeOfSignatureDeclaration` instead of `typeToTypeNode` directly (rather than built directly into the `declarations` transformer). That way it's easy to leverage the existing caching and safe copying mechanisms we have, and to reuse the logic in the JS declaration emitter (which, I dunno if you wanna consider testing that out of scope or not).

dragomirtitian:
Closed in favor of #57772

PR #57595, opened 2024-03-01
Accept decimal strings in Intl.NumberFormat

### LINKED ISSUE #52124: Intl.NumberFormat.prototype.format parameter type is too strict
# lib Update Request

I'm happy to open up a PR here, but wanted to open an issue first to discuss.

## Configuration Check

My compilation *target* is `ES2017` and my *lib* is `the default`.

The types for `Intl.NumberFormat.prototype.format` require a number type, even though the [TC39 spec](https://tc39.es/ecma402/#sec-number-format-functions) does not require the input parameter to be a number and will [handle string input](https://tc39.es/ecma262/#sec-tonumber).

## Sample Code

```TypeScript
// This will throw a type error because the input is a string, even though the browser will actually handle a string
new Intl.NumberFormat('en-US').format('100');
```

https://www.typescriptlang.org/play?#code/HYUw7gBAksAuA2A6AcgVwLYCMQCcBiA9jugIawAUA5CMALQCqAypQDQCUiAZkaRZQIwAGQZTYBuIA

## Documentation Link

1. https://tc39.es/ecma402/#sec-intl.numberformat.prototype.format
2. https://tc39.es/ecma402/#sec-number-format-functions
3. https://tc39.es/ecma262/#sec-tonumeric
4. https://tc39.es/ecma262/#sec-tonumber

felixfbecker:
Fixes #52124 

Note: This is **not** a coercion like in other JS APIs that also "accept" strings by coercing them, it is specifically designed to handle arbitrary-length decimals without precision loss. This was added as an explicit feature in Intl.NumberFormat v3 (which is why I added it to `es2023.intl`, which also houses `formatRange()`, which also got added in v3): https://github.com/tc39/proposal-intl-numberformat-v3#interpret-strings-as-decimals-ecma-402-334

The format matches what `${number}` accepts (plus `Infinity`):

> The general syntax for decimal strings, essentially #.#E#, is a widely understood format in computing. The specific version we use is the ECMA-262 StringNumericLiteral grammar, which also allows non-base-10 numbers like hexadecimal and binary.

Spec link: https://tc39.es/ecma402/#sec-runtime-semantics-stringintlmv

It is very common to represent monetary numbers as strings in JSON APIs to avoid any precision loss and those can currently not be passed to the `Intl.NumberFormat` API without casting `as any`. The resulting type errors currently misleads developers into using `parseFloat()` to make the type error go away, which causes precision loss.

It is supported by all browsers: https://caniuse.com/mdn-javascript_builtins_intl_numberformat_format_number_parameter-string_decimal

### DIFF
--- a/src/lib/es2023.intl.d.ts
+++ b/src/lib/es2023.intl.d.ts
@@ -28,7 +28,9 @@ declare namespace Intl {
     }
 
     interface NumberFormat {
-        formatRange(start: number | bigint, end: number | bigint): string;
-        formatRangeToParts(start: number | bigint, end: number | bigint): NumberRangeFormatPart[];
+        format(value: number | bigint | `${number}`): string;
+        formatToParts(value: number | bigint | `${number}`): NumberFormatPart[];
+        formatRange(start: number | bigint | `${number}`, end: number | bigint | `${number}`): string;
+        formatRangeToParts(start: number | bigint | `${number}`, end: number | bigint | `${number}`): NumberRangeFormatPart[];
     }
 }
--- a/src/lib/es2023.intl.d.ts
+++ b/src/lib/es2023.intl.d.ts
@@ -27,10 +27,12 @@ declare namespace Intl {
         source: "startRange" | "endRange" | "shared";
     }
 
+    type StringNumericLiteral = `${number}` | "Infinity" | "-Infinity" | "+Infinity";
+
     interface NumberFormat {
-        format(value: number | bigint | `${number}`): string;
-        formatToParts(value: number | bigint | `${number}`): NumberFormatPart[];
-        formatRange(start: number | bigint | `${number}`, end: number | bigint | `${number}`): string;
-        formatRangeToParts(start: number | bigint | `${number}`, end: number | bigint | `${number}`): NumberRangeFormatPart[];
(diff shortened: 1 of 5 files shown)


Renegade334:
👍 ([this](https://402.ecma-international.org/10.0/#sec-tointlmathematicalvalue) is the relevant feature in the spec – strings are now converted to mathematical values, rather than coerced to numbers as in previous versions)

FYI, the ECMA-402 v10 NumberFormat definitions were moved to [`es2023.intl`](https://github.com/microsoft/TypeScript/blob/877d9d316dd69ebd8253fe4c8915800415b0f455/src/lib/es2023.intl.d.ts#L30-L33) since your branch, so this will need updating.

felixfbecker:
Rebased. The new structure is nice!

Renegade334:
@felixfbecker updated tests: afc13e9e53731792be58f6b50b83c87f1db06afe

sandersn:
This makes sense to me. I'll bring the issue up at a design meeting.

sandersn:
We agreed that this makes sense. @felixfbecker would you mind bringing the PR up to date with main?

sandersn (approved):

felixfbecker:
@sandersn Awesome, I rebased onto `main`! Build is passing.

PR #52350, opened 2023-01-21
Run tests on all OSs

jakebailey:
Up until now, we only tested Linux in CI. It seems like a good idea to test all of our supported platforms and versions (especially since we get it "for free" in parallel).

PR #57607, opened 2024-03-02
Make sure virtual file system with watch behaves same way as sys/node so we have proper test coverage for symlinks

sheetalkamat:
- With node.js watching, if the folder/file thats being watched is symlink the updates to target are observed. So i added test thats common for `sys` and our `virtualFileSystemWithWatch` so they are in sync and fixed vfswith watch issue where updates were not percolated.
- Also added few additional test cases and fixed incorrect test as well .

### DIFF
--- a/src/testRunner/unittests/sys/symlinkWatching.ts
+++ b/src/testRunner/unittests/sys/symlinkWatching.ts
@@ -167,11 +167,10 @@ describe("unittests:: sys:: symlinkWatching::", () => {
         fsWatch: FsWatch<System>,
         dir: string,
         link: string,
-        isMacOs: boolean,
-        isWindows: boolean,
+        osFlavor: TestServerHostOsFlavor,
     ) {
         it(`watchDirectory using fsEvents`, async () => {
-            const tableOfEvents: FsEventsForWatchDirectory = isMacOs ?
+            const tableOfEvents: FsEventsForWatchDirectory = osFlavor === TestServerHostOsFlavor.MacOs ?
                 {
                     fileCreate: [
                         { event: "rename", fileName: "file1.ts" },
@@ -202,7 +201,7 @@ describe("unittests:: sys:: symlinkWatching::", () => {
                         { event: "rename", fileName: "file2.ts" },
                     ],
                 } :
-                isWindows ?
+                osFlavor === TestServerHostOsFlavor.Windows ?
                 {
                     fileCreate: [
                         { event: "rename", fileName: "file1.ts" },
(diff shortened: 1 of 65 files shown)


src/testRunner/unittests/sys/symlinkWatching.ts
```diff
@@ -0,0 +1,222 @@
...
+        });
+    }
+
+    function verifyWatchDirectoryUsingFsEvents<System extends ts.System>(
+        sys: System,
+        dir: string,
+        link: string,
+        fsWatch: (dir: string, cb: ts.FsWatchCallback, sys: System) => ts.FileWatcher,
+    ) {
+        it(`watchDirectory using fsEvents sheetal`, async () => {
```
sheetalkamat, line 76:
```suggestion
        it(`watchDirectory using fsEvents`, async () => {
```

src/testRunner/unittests/sys/symlinkWatching.ts
```diff
@@ -0,0 +1,222 @@
...
+    function getFileName(): (dir: string) => string {
+        return dir => `${dir}/${ts.getBaseFileName(dir)}.ts`;
+    }
+
+    describe("with ts.sys::", () => {
+        const root = ts.normalizePath(IO.joinPath(IO.getWorkspaceRoot(), "tests/baselines/symlinks"));
+        before(() => {
+            cleanup();
+            ts.sys.writeFile(`${root}/polling/file.ts`, "export const x = 10;");
+            withSwallowException(() => fs.symlinkSync(`${root}/polling`, `${root}/linkedpolling`, "junction"));
```
jakebailey, line 128: Does this test need to be using `symlink` on non-windows OSs?

sheetalkamat: i am not sure what you mean. This code becomes OS independent with the usage of symlinkSync

jakebailey: I guess it only works because all OSs besides Windows ignore the symlink type parameter. Feels sorta strange.

I will say that I'm wondering if we should get #52350 in so that we're actually running the tests on Windows; right now we're just testing Linux so can't verify that this PR is okay elsewhere.

sheetalkamat: can we do this sooner then.. i would like to get this change in so that we are consistent in the watch behaviour testing. 
I have some follow up issues that need this to be merged

jakebailey: Merged that PR, so you should be good to merge main or open/close to rerun with the new jobs.

jakebailey:
Ah, good thing I made that change; the PR fails on macos.

src/testRunner/unittests/sys/symlinkWatching.ts
```diff
@@ -0,0 +1,805 @@
...
+            getFileName(),
+        );
+
+        verifyWatchDirectoryUsingFsEvents(
+            getSys(TestServerHostOsFlavor.Windows),
+            (dir, recursive, cb, sys) => sys.fsWatchWorker(dir, recursive, cb),
+            `${root}/folder`,
+            `${root}/linked`,
+            /*isMacOs*/ false,
+            /*isWindows*/ true,
```
jakebailey, line 750: The "flavor" and the named parameters seem redundant; maybe `verifyWatchDirectoryUsingFsEvents` can just infer the bools from the flavor and construct a sys itself? (And `verifyWatchFile` too?)

jakebailey (approved):

Issue #57264, opened 2024-02-01
Adopt code actions ranges for refactorings in TS


justschen:
see API proposal here: https://github.com/microsoft/vscode/issues/199608

vs code tracking: https://github.com/microsoft/vscode/issues/204040

The API was made since often times in refactorings such as with `extract` or `move too...` it's not always clear what part of code or what part of the file the actions will apply to. By allowing contributed Code Actions to provide one or more ranges (`Range[]`), the range will be highlighted to show the user what will be changed.

All we need to provide is a `Range[] ` of affected code. For refactorings, this range would be the code in the current file that will be effected by the refactoring.

ashu8912:
Hey @justschen can I pick this one?? Would be a great start for me to learn and start my first contribution to the project.

justschen:
@RyanCavanaugh @navya9singh heads up I've done some work on the vs code / extension side for quick fixes already https://github.com/microsoft/vscode/pull/204191. If you have questions about API usage lmk 👍

navya9singh:
@justschen thanks for the update! For refactorings, do you want the `ApplicableRefactorInfo` api to have another parameter for `range?: Range[]` in the protocol? 
I looked through what `Range` looks like on the vscode side, and it seems that it has more information other than `start` and `end`. Is that something that is needed on the TS side as well?

justschen:
@navya9singh for the purposes of the highlights, we use `startLine`, `startCharacter`, `endLine`, and `endCharacter` since we convert it to the vscode `IRange` type to be used in the editor. everything else won't be checked/used so likely won't have to implement those 👍

PR #57608, opened 2024-03-02
Adopt code action ranges for refactorings

navya9singh:
This pr fixes https://github.com/microsoft/TypeScript/issues/57264 by getting ranges of text that will be affected for extract type, extract symbol, move to file and move to new file refactorings.


src/services/refactors/extractType.ts
```diff
@@ -260,14 +264,14 @@ function rangeContainsSkipTrivia(r1: TextRange, node: TextRange, file: SourceFil
...
+function collectTypeParameters(checker: TypeChecker, selection: TypeNode | TypeNode[], enclosingNode: Node, file: SourceFile): { typeParameters: TypeParameterDeclaration[] | undefined; affectedTextRange: TextRange | undefined; } {
     const result: TypeParameterDeclaration[] = [];
     const selectionArray = toArray(selection);
     const selectionRange = { pos: selectionArray[0].pos, end: selectionArray[selectionArray.length - 1].end };
     for (const t of selectionArray) {
-        if (visitor(t)) return undefined;
+        if (visitor(t)) return { typeParameters: undefined, affectedTextRange: undefined };
     }
-    return result;
+    return { typeParameters: result, affectedTextRange: selectionRange };
```
andrewbranch, line 274: Small nit: the beginning of `selectionRange` includes all the leading whitespace before the first type parameter, which may look weird to the user now that it will be visible. Not a huge deal, but we should be somewhat careful that ranges that previously existed only for programmatic use look reasonable when turned into a UI element.

src/services/refactors/moveToFile.ts
```diff
@@ -179,7 +179,8 @@ registerRefactor(refactorNameForMoveToFile, {
             }
         }
         if (context.preferences.allowTextChangesInNewFiles && statements) {
-            return [{ name: refactorNameForMoveToFile, description, actions: [moveToFileAction] }];
+            const affectedTextRange = { start: statements.all[0].pos, length: statements.all.length > 1 ? last(statements.all).end - statements.all[0].pos : statements.all[0].end - statements.all[0].pos };
```
andrewbranch, line 182: Same here with use of `.pos`; it might make more sense to use `.getStart(sourceFile)`

src/services/refactors/extractSymbol.ts
```diff
@@ -247,6 +248,7 @@ export function getRefactorActionsToExtractSymbol(context: RefactorContext): rea
                         description,
                         name: `function_scope_${i}`,
                         kind: extractFunctionAction.kind,
+                        range: { start: { line: getLineAndCharacterOfPosition(context.file, affectedTextRange.pos).line, offset: getLineAndCharacterOfPosition(context.file, affectedTextRange.pos).character }, end: { line: getLineAndCharacterOfPosition(context.file, affectedTextRange.end).line, offset: getLineAndCharacterOfPosition(context.file, affectedTextRange.end).character } },
```
andrewbranch, line 251: These lines look super long 😅

andrewbranch (approved):

src/services/refactors/extractSymbol.ts
navya9singh: Changed it 😄

mjbvz (approved):

Issue #56289, opened 2023-11-02
Default rest parameter names to destructuring names when of an anonymous tuple


JoshuaKGoldberg:
### 🔍 Search Terms

default rest parameter name destructure anonymous tuple

### ✅ Viability Checklist

- [X] This wouldn't be a breaking change in existing TypeScript/JavaScript code
- [X] This wouldn't change the runtime behavior of existing JavaScript code
- [X] This could be implemented without emitting different JS based on the types of the expressions
- [X] This isn't a runtime feature (e.g. library functionality, non-ECMAScript syntax with JavaScript output, new syntax sugar for JS, etc.)
- [X] This feature would agree with the rest of our Design Goals: https://github.com/Microsoft/TypeScript/wiki/TypeScript-Design-Goals

### ⭐ Suggestion

Right now, when calling a function with an inline destructured rest parameter whose type is an anonymous tuple, the inferred variable names will be auto-generated like `__0_0`:

```ts
function withPair(...[first, second]: [number, named: string]) { }

withPair(/* __0_0: number, named: string */ 0, "");
```

In cases where the tuple is destructured inline, how about any constant/static destructuring names being used as a fallback for anonymous tuple elements?

I.e.:

```ts
function withPair(...[first, second]: [number, named: string]) { }

withPair(/* first: number, named: string */ 0, "");
```

### 📃 Motivating Example

The pattern of immediately destructured rest parameters of a tuple is one that sometimes gets used for better type safety on parameters than generics or overloads. One downside, though, is that developers have to either explicitly add names to the tuple type _or_ suffer from poor auto-generated parameter hints.

A slightly more real-world example:

```ts
interface AppleInfo {
  color: "green" | "red";
}

interface BananaInfo {
  curvature: number;
}

type FruitAndInfo = ["apple", AppleInfo] | ["banana", BananaInfo];

function logFruitTuple(...[fruit, info]: FruitAndInfo) {
  switch (fruit) {
    case "apple":
      console.log(`My apple's color is ${info.color}.`);
      break;
    case "banana":
      console.log(`My banana's curvature is ${info.curvature}.`);
      break;
  }
}

logFruitTuple(/* __0_0: "apple", __0_1: AppleInfo */ "apple", { color: "red" });
```

With this suggested change, the parameter hints for `logFruitTuple` would instead be:

```ts
logFruitTuple(/* fruit: "apple", info: AppleInfo */ "apple", { color: "red" });
```

### 💻 Use Cases

[...]

PR #57619, opened 2024-03-03
Use destructured names as fallback labels when expanding parameters

Andarist:
closes https://github.com/microsoft/TypeScript/issues/56289


src/compiler/checker.ts
```diff
@@ -13255,23 +13255,68 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         return result;
     }
 
+    function getFallbackExpandedLabels(restType: TupleTypeReference, restDeclaration: Declaration | undefined) {
+        if (!restDeclaration) {
+            return;
+        }
+        Debug.assert(isParameter(restDeclaration));
```
jakebailey, line 13262:
```suggestion
        Debug.assertNode(restDeclaration, isParameter);
```

src/compiler/checker.ts
```diff
@@ -13255,23 +13255,68 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+        }
+        Debug.assert(isParameter(restDeclaration));
+        if (restDeclaration.name.kind !== SyntaxKind.ArrayBindingPattern) {
+            return;
+        }
+        const nameElements = restDeclaration.name.elements;
+        const elementFlags = restType.target.elementFlags;
+        const length = Math.min(nameElements.length, elementFlags.length);
+        const result: (__String | undefined)[] = [];
+        for (let i = 0; i < length; i++) {
```
jakebailey, line 13270: I can't really put my finger on why this feels like a lot of code, but it feels like this would be simpler if this just preallocated an array of undefineds and then wrote into it by index.

src/compiler/checker.ts
```diff
@@ -13349,23 +13349,60 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+        if (!restDeclaration) {
+            return;
+        }
+        Debug.assertNode(restDeclaration, isParameter);
+        if (restDeclaration.name.kind !== SyntaxKind.ArrayBindingPattern) {
+            return;
+        }
+        const nameElements = restDeclaration.name.elements;
+        const elementFlags = restType.target.elementFlags;
+        const result: (__String | undefined)[] = arrayOf(elementFlags.length, () => undefined);
```
jakebailey, line 13362: On second thought, I don't know how I feel about this, but I'm also still not coming up with a better algorithm. I was probably too nitpicky before.

src/compiler/checker.ts
```diff
@@ -13349,23 +13349,60 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
     function getExpandedParameters(sig: Signature, skipUnionExpanding?: boolean): readonly (readonly Symbol[])[] {
         if (signatureHasRestParameter(sig)) {
             const restIndex = sig.parameters.length - 1;
             const restName = sig.parameters[restIndex].escapedName;
             const restType = getTypeOfSymbol(sig.parameters[restIndex]);
+            const restDeclaration = sig.parameters[restIndex].valueDeclaration;
+
             if (isTupleType(restType)) {
-                return [expandSignatureParametersWithTupleMembers(restType, restIndex, restName)];
+                return [expandSignatureParametersWithTupleMembers(restType, restIndex, restName, getFallbackExpandedLabels(restType, restDeclaration))];
```
jakebailey, line 13395: Another thing I realized is that we're going to be calculating this fallback list every time we do this, even though it's not always used. But, I'm not sure how to make that lazy either.

src/compiler/checker.ts
Andarist: I think this might not be equivalent. The `elementFlag & ElementFlags.Variable` case doesn't immediately "stop" the iteration and I think that the top condition (`seenDotDotDot || i >= nameElements.length`) doesn't cover this case either.

I pushed out your proposed code with a tweak to address the above. I'm not sure if I find it particularly more readable but I appreciate how it avoids an extra iteration so I'm fine with it.

src/compiler/checker.ts
Andarist: ye, this is somewhat unfortunate but I don't think it's a particularly costly operation to create this so I just don't mind having this eager here

jakebailey:
@typescript-bot test it

jakebailey:
Hm, hard to say if that vscode result is just noise or not.

jakebailey:
Retesting with new benchmarks

jakebailey:
@typescript-bot pack this

DanielRosenwasser:
for

```ts
function foo(first: string, ...args: [first: number, second: number]) {
}
```

are we okay with this behavior?

image

jakebailey (changes requested):
Also, on second thought, I think that this should actually be `first, second`, not `first, named`; after all, the type could be coming from far far away, but the names from the rest are actually what's relavent for the function call...

image

jakebailey:
> are we okay with this behavior?

I'd say no; now that we have two sources of truth, we should probably try and remember which names we've already used 🙁

rbuckton:
This doesn't cover quite as many cases as #59045, though I found the fourslash tests in this PR very helpful so I've added them into #59045.

PR #57628, opened 2024-03-04
Fixed crash in go to definition related to expando classes in JS files

Andarist:
fixes a crash reported [here](https://github.com/microsoft/TypeScript/issues/57622#issuecomment-1975331863)


src/services/goToDefinition.ts
```diff
@@ -583,16 +583,20 @@ function isExpandoDeclaration(node: Declaration): boolean {
 
 function getDefinitionFromSymbol(typeChecker: TypeChecker, symbol: Symbol, node: Node, failedAliasResolution?: boolean, excludeDeclaration?: Node): DefinitionInfo[] | undefined {
     const filteredDeclarations = filter(symbol.declarations, d => d !== excludeDeclaration);
+    const signatureDefinition = getConstructSignatureDefinition() || getCallSignatureDefinition();
```
Andarist, line 586: Changes in this function are not needed. I just tweaked it while trying to understand what happens here. This still acts as a short-circuit for calls below it so I'd call it a micro-optimization and keep the change ;p

src/services/goToDefinition.ts
```diff
@@ -583,16 +583,20 @@ function isExpandoDeclaration(node: Declaration): boolean {
...
+    return map(results, declaration => createDefinitionInfo(declaration, typeChecker, symbol, node, /*unverified*/ false, failedAliasResolution));
 
     function getConstructSignatureDefinition(): DefinitionInfo[] | undefined {
         // Applicable only if we are in a new expression, or we are on a constructor declaration
         // and in either case the symbol has a construct signature definition, i.e. class
         if (symbol.flags & SymbolFlags.Class && !(symbol.flags & (SymbolFlags.Function | SymbolFlags.Variable)) && (isNewExpressionTarget(node) || node.kind === SyntaxKind.ConstructorKeyword)) {
-            const cls = find(filteredDeclarations, isClassLike) || Debug.fail("Expected declaration to have at least one class-like declaration");
-            return getSignatureDefinition(cls.members, /*selectConstructors*/ true);
+            const cls = find(filteredDeclarations, isClassLike);
+            return cls && getSignatureDefinition(cls.members, /*selectConstructors*/ true);
```
Andarist, line 599: there is a question in this case - should this be treated more like a class or more like a property? I decided that it's better to ignore class-ness here as the expando property acts as a common root for other things and as such it makes sense to prefer showing it here in my opinion.

sandersn: What is the difference in behaviour? The baselines for goToDefinitionExpandoClass2 show that the constructor is found when present, so what are we skipping here?

sandersn (changes requested):
Results look good -- but I would like to know what effect replacing an assert with a skip actually has.

src/services/goToDefinition.ts
Andarist: Without this skip it crashes while resolving what becomes `defId: 0` in `goToDefinitionExpandoClass2` with the fix. The resolved `constructor` (`defId: 1`) comes from `sigInfo` here:
https://github.com/microsoft/TypeScript/blob/9d8f812a83df0045d5e2e7757017c050986c800a/src/services/goToDefinition.ts#L239

sandersn: OK, that makes sense. I guess I thought of Test so strongly as a class in `Core.Test = class { ...` that I didn't even consider that it might not appear because it's a property. In other words, I think that `defId: 0` definitely makes sense to include.

sandersn (approved):

jakebailey (approved):

PR #57637, opened 2024-03-04
Fixed a regression related to determining argument index when spread elements are involved

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/57622#issuecomment-1975331860
likely fixes https://github.com/microsoft/TypeScript/issues/57623#issuecomment-1975372558
likely fixes https://github.com/microsoft/TypeScript/issues/57826

the issue is a regression from https://github.com/microsoft/TypeScript/pull/56372


src/services/signatureHelp.ts
```diff
@@ -485,8 +484,7 @@ function getArgumentIndex(argumentsList: Node, node: Node, checker: TypeChecker)
     // The list we got back can include commas.  In the presence of errors it may
     // also just have nodes without commas.  For example "Foo(a b c)" will have 3
     // args without commas. We want to find what index we're at.  So we count
-    // forward until we hit ourselves, only incrementing the index if it isn't a
-    // comma.
+    // forward until we hit ourselves.
```
Andarist, line 487: I removed this part of the comment because I adjusted the behavior (see `tests/cases/fourslash/signatureHelpSkippedArgs1.ts`). I find the new behavior better and it was easier for me to rewrite those loops while accommodating for that test case.

src/services/signatureHelp.ts
```diff
@@ -525,32 +534,35 @@ function getSpreadElementCount(node: SpreadElement, checker: TypeChecker) {
...
+function getArgumentCount(argumentsList: Node, checker: TypeChecker) {
     // The argument count for a list is normally the number of non-comma children it has.
     // For example, if you have "Foo(a,b)" then there will be three children of the arg
     // list 'a' '<comma>' 'b'.  So, in this case the arg count will be 2.  However, there
     // is a small subtlety.  If you have "Foo(a,)", then the child list will just have
     // 'a' '<comma>'.  So, in the case where the last child is a comma, we increase the
     // arg count by one to compensate.
-    //
-    // Note: this subtlety only applies to the last comma.  If you had "Foo(a,," then
-    // we'll have: 'a' '<comma>' '<missing>'
```
Andarist, line 537: The mention of `<missing>` was added here like 9 years ago but the missing node is not used here for years already (I checked against some 3.x versions). When dealing with `fn(,,,,)` we just get a list of comma tokens

src/services/signatureHelp.ts
```diff
@@ -525,32 +534,35 @@ function getSpreadElementCount(node: SpreadElement, checker: TypeChecker) {
     return 0;
 }
 
-function getArgumentCount(argumentsList: Node, ignoreTrailingComma: boolean, checker: TypeChecker) {
+function getArgumentCount(argumentsList: Node, checker: TypeChecker) {
```
Andarist, line 537: This function and `getArgumentIndex` are meant to be kinda in sync - I can try to merge them if there is an appetite for it.

tests/cases/fourslash/signatureHelpRestArgs2.ts
```diff
@@ -0,0 +1,18 @@
...
+// @allowJs: true
+// @checkJs: true
+
+// @filename: index.js
+
+//// const promisify = function (thisArg, fnName) {
+////     const fn = thisArg[fnName];
+////     return function () {
+////         return new Promise((resolve) => {
+////             fn.call(thisArg, ...arguments, /*1*/);
```
Andarist, line 13: Crash from https://github.com/microsoft/TypeScript/issues/57622#issuecomment-1975331860 happens in completions but it's caused by the same underlying issue. If requested I can add an extra completions-oriented test case too

src/services/signatureHelp.ts
jakebailey: I am somewhat curious how that'd look, but I also don't want you to spend a bunch of time on that if you don't want to.

src/services/signatureHelp.ts
```diff
@@ -525,32 +534,35 @@ function getSpreadElementCount(node: SpreadElement, checker: TypeChecker) {
...
-    // we'll have: 'a' '<comma>' '<missing>'
-    // That will give us 2 non-commas.  We then add one for the last comma, giving us an
-    // arg count of 3.
-    const listChildren = argumentsList.getChildren();
-
+    const args = argumentsList.getChildren();
     let argumentCount = 0;
-    for (const child of listChildren) {
+    let skipComma = false;
+    for (let pos = 0; pos < length(args); pos++) {
```
DanielRosenwasser, line 547: Why `length(args)` instead of args.length?

src/services/signatureHelp.ts
Andarist: > if you don't want to.

this tbh ;p

IIRC they are now mostly the same, one call would pass an "end index" to get the argument's index and the other one could pass `undefined` in its place to iterate through the whole thing and get the arg count (then we'd also have to fixup the index vs count difference with some `+ 1`)

src/services/signatureHelp.ts
Andarist: nNo particular reason really - I was probably just replicating some other code here. While `length()` has this additional ability to accept undefined inputs, it's already used by a lot of places where the input is defined. I'll replace this with plain `.length` in a second

src/services/signatureHelp.ts
jakebailey: Skimming, aren't these the same functions, just `getArgumentIndex` breaks out early when it finds a specific node? The count difference only matters if you actually finish the loop. It does feel like they should be just one iterator that returns early when a given node is found (or just lets it end if not provided).

Andarist: yep, I mostly ended up not unifying them because I had to switch to other work and because naming things is hard 😉 if you want them unified (which is a reasonable ask) I'll get to it some time later

jakebailey: It's pretty easy to unify them:

```ts
function getArgumentIndexOrCount(argumentsList: Node, checker: TypeChecker, node: Node | undefined) {
    const args = argumentsList.getChildren();
    let argumentIndex = 0;
    let skipComma = false;
    for (const child of args) {
        if (node && child === node) {
            if (!skipComma && child.kind === SyntaxKind.CommaToken) {
                argumentIndex++;
            }
            return argumentIndex;
        }
        if (isSpreadElement(child)) {
            argumentIndex += getSpreadElementCount(child, checker);
            skipComma = true;
            continue;
        }
        if (child.kind !== SyntaxKind.CommaToken) {
            argumentIndex++;
            skipComma = true;
            continue;
        }
        if (skipComma) {
            skipComma = false;
            continue;
        }
        argumentIndex++;
    }
    if (node) {
        return argumentIndex;
    }
    return args.length && last(args).kind === SyntaxKind.CommaToken ? argumentIndex + 1 : argumentIndex;
}
```

The harder part is reintegrating the comments, I think.

Andarist: I've used this suggestion and reintegrated the comments.

jakebailey: Thanks; I forgot to mention it but would you mind leaving the original functions behind as wrappers, just so the call sites remain the same? (A la abcdWorker funcs in the checker, for example)

jakebailey: I'm also totally happy to apply that change; I think this is something I want to suggest for the next patch release just due to the crashes.

jakebailey: Pushed that to your branch quick just to move this forward towards a backport (sorry to make you pull)

jakebailey (approved):

jakebailey:

DanielRosenwasser (approved):
I get the general gist here, and I don't see much changing in the way of existing tests which I know we spent a bunch of time on regarding picking the right overload. So I think we can take this PR.

PR #53413, opened 2023-03-21
Avoid calculating union in spread if types are identical

jakebailey:
Fixes #52345

When spreading, we remove `undefined` from the right type and union it with the left. If the two types are huge but overlap, we'll add all of the union's members together into one big list, then run subtype reduction over it, which for this test takes a long time. But, we can skip that work if we can recognize that left/right are the same. With #53406, this eliminates all of the bottlenecks in #52345.

Since this spread stuff is tricky, I generated a test which tests all combos of spreading + optional + missing to verify that I don't screw it up.

Likely, there's something more general that could be done to `getUnionType` here; it always recurses into every union and _then_ does subtype reduction. I wonder if we can deduplicate the list first (ignoring undefined-ness), and trivially bail out early where possible, but that seemed a little tricky.

Issue #48178, opened 2022-03-08
Spurious [?] circularity error in getter with spread


RyanCavanaugh:
# Bug Report

### 🔎 Search Terms

circularity getter

### 🕗 Version & Regression Information

- This is the behavior in every version I tried

### ⏯ Playground Link

[Playground link with relevant code](https://www.typescriptlang.org/play?#code/C4TwDgpgBAQg9gDygXigbwFBSgYzgO2AkIH4AuKAMTjigB9ZEMBfAbgwBMIcAbAQwBO0PPgDOwKACMK8BO159RoqjXRYoAcwgSAZjQAUASjXZsQ4AFcB+E6ewiihCsAAWAS2WKVtBrIA06nYAdCGSgVBs6swsQA)

### 💻 Code

```ts
type Box = {
  content?: Foo | Box
};
declare const b: Box;
class Foo {
  get foo() {
    return {
      content: this as Foo | Box,
      ...b
    };
  }
}
```

### 🙁 Actual behavior

Circularity error on 'foo' when this is probably deferrable

### 🙂 Expected behavior

No circularity error

Ref. #46981

Andarist:
The problem comes from within the [`checkAccessorDeclaration`](https://github.dev/microsoft/TypeScript/blob/c6fb2e10dcdf4839924a4288dae1cfd93b3c1c57/src/compiler/checker.ts#L38221) and the [`getTypeOfAccessors`](https://github.dev/microsoft/TypeScript/blob/c6fb2e10dcdf4839924a4288dae1cfd93b3c1c57/src/compiler/checker.ts#L38276) call in it. 

This starts a type resolution that ends up with a cycle because `getSpreadType` tries to create a union of the left type and the right type [here](https://github.dev/microsoft/TypeScript/blob/c6fb2e10dcdf4839924a4288dae1cfd93b3c1c57/src/compiler/checker.ts#L18114). The problem is that this is a subject of subtype union reduction and this [`find(getPropertiesOfType(source), p => isUnitType(getTypeOfSymbol(p)))`](https://github.dev/microsoft/TypeScript/blob/c6fb2e10dcdf4839924a4288dae1cfd93b3c1c57/src/compiler/checker.ts#L16196) leads to requesting the `get foo`'s type while we are still resolving it.

RyanCavanaugh:
Not an error anymore

Andarist:
[...]

PR #57642, opened 2024-03-04
Add a test case related to spreading objects

Andarist:
test case for https://github.com/microsoft/TypeScript/issues/48178
it was fixed by https://github.com/microsoft/TypeScript/pull/53413

### DIFF
--- a/tests/baselines/reference/spreadObjectNoCircular1.symbols
+++ b/tests/baselines/reference/spreadObjectNoCircular1.symbols
@@ -1,0 +1,37 @@
+//// [tests/cases/compiler/spreadObjectNoCircular1.ts] ////
+
+=== spreadObjectNoCircular1.ts ===
+type Box = {
+>Box : Symbol(Box, Decl(spreadObjectNoCircular1.ts, 0, 0))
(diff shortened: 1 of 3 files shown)


jakebailey:
#53413 shouldn't have fixed anything, so I do wonder what the heck I did...

Andarist:
@jakebailey the circularity was caused by subtype reduction - ur PR avoids it in certain cases and thus it accidentally~ fixed this issue

Andarist:
I re-checked and at the time the circularity was encountered by the subtype reduction we can see those types when we get to your optimization:
```ts
typeToString(getTypeOfSymbol(leftProp)) // Foo | Box
typeToString(rightType) // Foo | Box | undefined
```

So it makes sense that we don't get to the circularity now since u avoid that subtype reduction based on this check:
```ts
result.links.type = leftTypeWithoutUndefined === rightTypeWithoutUndefined ? leftType : getUnionType([leftType, rightTypeWithoutUndefined], UnionReduction.Subtype);
```

I'd call it - "accidental but good" :P

PR #57643, opened 2024-03-05
Properly set `hasAction` for erasing modifiers in class member completions

### LINKED ISSUE #57301: Typescript autocomplete: extending an abstract class's async method get autocompleted incorrectly

### LINKED ISSUE #57333: Method completion insertion is mangled
On the latest Insiders

1. Check out this branch https://github.com/microsoft/vscode/tree/connor4312/ts-insert-issue and go to L215 of extHostDebugService.ts
2. Type `public $` 

    image

3. Accept the completion for `$provideDebugVisualization`, and notice the insertion gets mangled 🐛 

    image

gabritto:
Fixes #57301.
Also fixes #57333.
@iisaduan pointed out that class member completions were always duplicating modifiers (any modifiers, not just `async`).

To update on my [last comment](https://github.com/microsoft/TypeScript/pull/52525#issuecomment-1441005953) on the original PR for dealing with modifiers in class member completions, the current behavior (with this fix) is:

- We only offer the completion if the already present modifiers match the modifiers the completion would insert
- The modifiers already present may be out of order, so we always add *all* the modifiers as part of the member completion entry, and send a code action to erase the present modifiers.

Example:

```ts
class Base {
    protected prop = 1;
}
class E extends Base {
    override protected |
}
```

becomes, after accepting completion for `prop`:

```ts
class Base {
    protected prop = 1;
}
class E extends Base {
    protected override prop: number;|
}
```

### DIFF
--- a/src/services/completions.ts
+++ b/src/services/completions.ts
@@ -1246,6 +1246,7 @@ function completionInfoFromData(
     const {
         symbols,
         contextToken,
+        previousToken,
         completionKind,
         isInSnippetScope,
         isNewIdentifierLocation,
@@ -1305,6 +1306,7 @@ function completionInfoFromData(
         entries,
         /*replacementToken*/ undefined,
         contextToken,
+        previousToken,
         location,
         position,
         sourceFile,
@@ -1634,6 +1636,7 @@ function createCompletionEntry(
     sortText: SortText,
     replacementToken: Node | undefined,
     contextToken: Node | undefined,
+    previousToken: Node | undefined,
     location: Node,
     position: number,
     sourceFile: SourceFile,
@@ -1757,7 +1760,19 @@ function createCompletionEntry(
         isClassLikeMemberCompletion(symbol, location, sourceFile)
     ) {
         let importAdder;
-        const memberCompletionEntry = getEntryForMemberCompletion(host, program, options, preferences, name, symbol, location, position, contextToken, formatContext);
+        const memberCompletionEntry = getEntryForMemberCompletion(
+            host,
+            program,
+            options,
+            preferences,
+            name,
+            symbol,
+            location,
+            position,
+            contextToken,
+            previousToken,
+            formatContext,
+        );
         if (memberCompletionEntry) {
             ({ insertText, filterText, isSnippet, importAdder } = memberCompletionEntry);
             if (importAdder?.hasFixes() || memberCompletionEntry.eraseRange) {
@@ -1915,6 +1930,7 @@ function getEntryForMemberCompletion(
     location: Node,
     position: number,
     contextToken: Node | undefined,
+    previousToken: Node | undefined,
     formatContext: formatting.FormatContext | undefined,
 ): { insertText: string; filterText?: string; isSnippet?: true; importAdder?: codefix.ImportAdder; eraseRange?: TextRange; } | undefined {
     const classLikeDeclaration = findAncestor(location, isClassLike);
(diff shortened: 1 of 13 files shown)


gabritto:
@Andarist

Andarist:
I kind of already had this change as part of my PR (see the commit [here](https://github.com/microsoft/TypeScript/pull/57480/commits/dc270ed67cef4c240c65215c0bd67b236474636f)). I found the code action approach to have some DX downsides (mentioned that past half point of my comment [here](https://github.com/microsoft/TypeScript/pull/57480#issuecomment-1974754550)) and thus I didn't quite follow on it to adjust the test cases, add a new diagnostic etc.

gabritto:
As I said in https://github.com/microsoft/TypeScript/pull/57480#issuecomment-1979324171:

> The other course of action I can think of to fix this is to omit existing modifiers from the completion, but now if the modifier order is wrong or you need to add a modifier before the existing one, you can't.
> It really does look like what we need is to use a replacementSpan to replace the existing modifiers. Maybe I should check with vscode again if that's at all possible.

Meanwhile, this PR fixes what I tried to implement in #52525.

tests/cases/fourslash/completionsOverridingMethod18.ts
iisaduan, line 1: Same as the marker `g` test in `tests/cases/fourslash/completionsOverridingMethod6.ts`

iisaduan (commented):
I'm still looking through the code, but the test cases look pretty good so far! I think we should add a case for https://github.com/microsoft/TypeScript/issues/57333, since this PR also fixes it.

```
//b.ts
export type C = { x: number }

export interface ExtShape {
    $arg(id: C ): Promise<number[]>;
    $returnPromise(id: number): Promise<C>;
    $return(id: number): C;
}

//test.ts
import { ExtShape } from './b';

abstract class ExtBase implements ExtShape {
    public $/**/
}
```

src/services/completions.ts
```diff
@@ -2054,11 +2070,13 @@ function getEntryForMemberCompletion(
 
 function getPresentModifiers(
     contextToken: Node | undefined,
+    previousToken: Node | undefined,
```
Andarist, line 2073: you could avoid having to pass `previousToken` here like this: https://github.com/Andarist/TypeScript/commit/4721fbdc61bb64a1f896fc5010b826dcf1b48c50

note that this commit reverts changes to `getPresentModifiers` and introduces changes (slightly adjusted) from https://github.com/microsoft/TypeScript/pull/57480 . I think that your version might have some further readability improvements but I just skipped merging your changes into this. If you like the core of the proposed change - I can incorporate your changes into this and open a PR targeting your branch

gabritto: Does that work because the case where `previousToken` is not the same as `contextToken` is the same as when `contextToken.parent.name !== contextToken`?

Andarist: I believe so, yes.

andrewbranch (approved):

iisaduan (approved):
Thank you and @Andarist for investigating and fixing this!

Issue #57620, opened 2024-03-03
Class marked as unused when used in a static block


nicolo-ribaudo:
### 🔎 Search Terms

class unused static field block

### 🕗 Version & Regression Information

- This is the behavior in every version I tried

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.5.0-dev.20240303#code/MYewdgzgLgBAQgGwK4FMoAsQigERQNwEtgUB9AaxQE8YBeGAbwF8BuAWACgE15k1NseIiVLAATigCGUFOw7AEkiBF6oMWXAWIpGnGDADEhACYBCAFwxoYwmADmcvTAAON-NJ2hIUMUmBQQMQAKSipLKCpnFBAAM1V+DSFtCmoYAB8YJDBjFBjbFGM6TOzc-OMASl0OfX1COJDU01p6RDUBTWEyUMqGJxqYDDEQAHcYMBRRgFExIeCAIgBNECQYYEkwMGwYW2h1qEIPAfRCFQUlFWNCCX8EKjnyuX6mJ2eOJ1394Cr+1oTBLRE4ikMiKQRMlmstjslVoAD5vv1VuBoDAcp0iuNRr91P9OkFse0kiJuo9EWjtAA6IyFegmUn9CRQJBiMCogGyPowVgvThOOw8ExBHqcxnM1kYE5Uuk8jivTjcWCSIoAJkcHAJiXZogkHiCc0kknuapQAA9nIFYMwgA

### 💻 Code

```ts
const BluethootDevice_key = {};
let BluethootDevice_create;
class BluethootDevice {
  #id!: string;

  private constructor(key: typeof BluethootDevice_key | undefined = undefined) {
    if (key !== BluethootDevice_key) {
      throw new Error("You cannot instantiate this class directly");
    }
  }

  static {
    BluethootDevice_create = (id: string) => {
      const device = new BluethootDevice(BluethootDevice_key);
      device.#id = id;
      return device;
    };
  }

  get id() {
    return this.#id;
  }
}

let a = 2;

BluethootDevice_create("aa");

export {}
```

### 🙁 Actual behavior

The class `BluethootDevice` is marked as unused, but it's used by the `BluethootDevice_create` function

### 🙂 Expected behavior

`BluethootDevice` should not be marked as unused

fatcerberus:
> Bluethoot

nicolo-ribaudo:
Yes I learned today that it's not spelled like that 😛

[...]

PR #57649, opened 2024-03-05
Treat self references in class static blocks as used

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/57620

### DIFF
--- a/tests/cases/compiler/noUnusedLocals_selfReferenceInClassStaticBlock.ts
+++ b/tests/cases/compiler/noUnusedLocals_selfReferenceInClassStaticBlock.ts
@@ -1,0 +1,37 @@
+// @strict: true
+// @target: esnext
+// @noUnusedLocals: true
+// @noEmit: true
+
+// https://github.com/microsoft/TypeScript/issues/57620
+
+let BluethootDevice_create;
+class BluethootDevice {
+  #id!: string;
+
+  static #allowConstruct = false;
+
+  private constructor() {
+    if (!BluethootDevice.#allowConstruct) {
+      throw new Error("You cannot instantiate this class directly");
+    }
+  }
+
+  static {
+    BluethootDevice_create = (id: string) => {
+      BluethootDevice.#allowConstruct = true;
+      const device = new BluethootDevice();
+      BluethootDevice.#allowConstruct = false;
+      device.#id = id;
+      return device;
+    };
+  }
+
+  get id() {
+    return this.#id;
+  }
+}
+
+BluethootDevice_create("aa");
+
+export {};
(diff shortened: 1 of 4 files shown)


tests/cases/compiler/noUnusedLocals_selfReferenceInClassStaticBlock.ts
```diff
@@ -0,0 +1,37 @@
+// @strict: true
+// @target: esnext
+// @noUnusedLocals: true
+// @noEmit: true
+
+// https://github.com/microsoft/TypeScript/issues/57620
+
+let BluethootDevice_create;
+class BluethootDevice {
```
nicolo-ribaudo, line 9: Happy to see that the way I always assumed Bluetooth is spelled is making it to the TS codebase :)

nicolo-ribaudo (commented):
Question without having understood the code in this PR. Is `A` marked as used or unused here?
```js
class A {
  method() {
    class B {
      static { use(A) }
    }
  }
}
```

Andarist:
I intend to treat `A`  as used here - see my reasoning [here](https://github.com/microsoft/TypeScript/pull/57649#discussion_r1512666920). I'll recheck this case later just to be sure that the current behavior matches my expectation but it really should because just *any* static block met on the path towards the target declaration (`class A` here) sets a boolean variable to true based on which I avoid reporting this error.

gabritto:
@typescript-bot test this

gabritto:
@typescript-bot test it

gabritto (approved):

gabritto:
@Andarist this probably needs a merge from main to update the baselines before I merge.

Andarist:
@gabritto done

gabritto:
@typescript-bot pack this

gabritto (changes requested):
Left a question in the comments about whether some cases should be considered unused or not. Sorry about not asking that earlier.

gabritto:
But doesn't this PR mark `Bar` and `foo` as used vs not used in current version?  https://www.staging-typescript.org/play?noUnusedLocals=true&module=1&ts=5.5.0-pr-57649-11#code/KYDwDg9gTgLgBAbwL4G4CwAoTBjANgQwGdC4AhfKRTOGuAW2BgAsIATACgEoqNa+5sEAHaF4eEgF4BBYnABiECD34q4o-DACW2Zar2CR8AJ5wp5KOl57VBwhFzAAdLggBzdkc6XrtJNVV+VipQjACuUELShN6+mIGYAGahQthawnAJily6NLZiuJLSRCQKSgj+KupaOuVB1nlwJlKZEDE+tvZOLu6ebfyBfAO0ITDhkeKWSEA

Andarist:
Ah, I thought the previous URL was already using the build from this PR. So I agree that those should be marked as unused (like on `main`) - I'll take a look into fixing this.

PR #57656, opened 2024-03-06
DO NOT MERGE: Show preserved triple slash references in diagnostics for top200 analysis 

andrewbranch:
Counterpart of #57569


andrewbranch:
@typescript-bot test top200

jakebailey:
@typescript-bot test top400

jakebailey:
The fuels-ts one emits a directive into the published package which doesn't actually exist. https://unpkg.com/browse/@fuel-ts/abi-typegen@0.76.0/dist/index.d.ts

jakebailey:
The jupyterlab one doesn't seem intentional, in that the package only dev deps on jest and there's a comment that says it's there for another reason. (I think normally such a reference would be elided if it were actually used elsewhere?)

In any case, these results are very good.

PR #57660, opened 2024-03-06
Experimental variant on #57465 with two optimizations

danvk:
This adds two optimizations to #57465:

- For the false check, if `trueType` is a large union (20+ constituents) then split it into two chunks when calling `getFlowTypeOfReference`. The first chunk has the first 10 types, the second has the remainder. If the first chunk doesn't narrow to `never` then we don't have to check the second. This is a huge win on #54148. On my laptop it reduces the time for the [most extreme check](https://github.com/microsoft/TypeScript/pull/57465#issuecomment-1974921974) from 312ms→140ms by only checking 10/112 types in the union. There are several other expensive calls on that PR that it's able to bypass as well. This has less of an effect on the Compiler-Unions benchmark.
- A fast path for functions of the form `x => x === y` (or `==`, `!==`, `!=`). These can only be type predicates if `y` is a unit type like `null` or `undefined`, but that can be expensive to determine via the usual calls to `getFlowTypeOfReference` if `x` and `y` are both large union types. In both Compiler-Unions and #54148 there are some expensive checks that this bypasses, notably `d => d === objectLiteralDeclaration`.

In local testing these help on both Compiler-Unions and #54148. The `x === y` optimization is definitely targeted at a specific case in Compiler-Unions, but it's a win on #54148 as well, so perhaps it does generalize.


jakebailey:

danvk:
I'm surprised this wasn't a win. Might be time to stop optimizing this!

PR #57663, opened 2024-03-06
Set a line length limit

jakebailey:
For discussion; this ratchets down the line length limit down probably past what we're willing to do.

You should be able to click on a specific commit then view the repo at that state, or something.

Overall change values for each level (run fresh for each level from main):

```
700  1 file changed, 5 insertions(+), 1 deletion(-)
600  2 files changed, 31 insertions(+), 4 deletions(-)
500  3 files changed, 43 insertions(+), 5 deletions(-)
400  7 files changed, 87 insertions(+), 10 deletions(-)
360  19 files changed, 269 insertions(+), 35 deletions(-)
300  51 files changed, 1155 insertions(+), 164 deletions(-)
280  62 files changed, 1702 insertions(+), 242 deletions(-)
240  106 files changed, 3882 insertions(+), 610 deletions(-)
220  145 files changed, 5917 insertions(+), 996 deletions(-)
200  189 files changed, 10034 insertions(+), 1769 deletions(-)
190  215 files changed, 12767 insertions(+), 2365 deletions(-)
180  245 files changed, 16270 insertions(+), 3094 deletions(-)
170  267 files changed, 20903 insertions(+), 4069 deletions(-)
160  301 files changed, 26924 insertions(+), 5424 deletions(-)
150  339 files changed, 35804 insertions(+), 7520 deletions(-)
140  381 files changed, 47241 insertions(+), 10088 deletions(-)
130  419 files changed, 60942 insertions(+), 13274 deletions(-)
120  453 files changed, 78208 insertions(+), 17397 deletions(-)
```

### DIFF
--- a/src/harness/evaluatorImpl.ts
+++ b/src/harness/evaluatorImpl.ts
@@ -10,7 +10,7 @@ const sourceFileJs = vpath.combine(vfs.srcFolder, "source.js");
 
 // Define a custom "Symbol" constructor to attach missing built-in symbols without
 // modifying the global "Symbol" constructor
-const FakeSymbol: SymbolConstructor = ((description?: string) => Symbol(description)) as any;
+export const FakeSymbol: SymbolConstructor = ((description?: string) => Symbol(description)) as any;
 (FakeSymbol as any).prototype = Symbol.prototype;
 for (const key of Object.getOwnPropertyNames(Symbol)) {
     Object.defineProperty(FakeSymbol, key, Object.getOwnPropertyDescriptor(Symbol, key)!);
--- a/src/lib/esnext.disposable.d.ts
+++ b/src/lib/esnext.disposable.d.ts
@@ -1,4 +1,6 @@
 /// <reference lib="es2015.symbol" />
+/// <reference lib="es2015.iterable" />
+/// <reference lib="es2018.asynciterable" />
 
 interface SymbolConstructor {
     /**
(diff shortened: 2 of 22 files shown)


andrewbranch:
Honestly I’m not going to look at the actual changes because everything I need to know to form my opinion is in my own editor. I use a 27" monitor and usually have VS Code taking up exactly half my desktop width, which gives me just over 140 characters visible without a sidebar open. I’m comfortable with 120, and anything over 140 is not very useful to me because I’d still have to scroll. I’ll be thrilled if we pick 120, 130, or 140.

jakebailey:
I am honestly unsure about 120 simply because we love to stick stuff into deep closures; most code within checker.ts is already at least 8 spaces deep, if not more. But, I would definitely normally agree that 120 is good, though even that barely doesn't fit on my screen with two columns...

andrewbranch:
I can think of a way to reduce the amount of spaces without changing the closure depth 👀

jakebailey:
absolutely not

Herebyfile.mjs
```diff
@@ -696,7 +729,10 @@ export const runTestsAndWatch = task({
             if (!token.signaled) {
                 running = true;
                 try {
-                    await runConsoleTests(testRunner, "mocha-fivemat-progress-reporter", /*runInParallel*/ false, { token, watching: true });
+                    await runConsoleTests(testRunner, "mocha-fivemat-progress-reporter", /*runInParallel*/ false, {
```
sheetalkamat, line 732: I would have expected it to be:
```ts
await runConsoleTests(
    testRunner,
    "mocha-fivemat-progress-reporter",
    /*runInParallel*/ false,
    { token, watching: true }
);
```

jakebailey: Yeah, this just comes down to the algorithm that dprint uses; it sees the first parameter is on the same line so starts by keeping it there, then gets to the end and flows the object type over instead.

jakebailey:
I have no idea why github closed this PR

PR #57665, opened 2024-03-06
Remove nonNullUnknownType

jakebailey:
As far as I can tell (and test), this `unknown` variant is no longer needed. Probably due to the changes to how unknown is more or less `{} | undefined | null`.

This effectively finishes reverting #45575; most of it was removed in #49119 (as expected). By the numbers, #45575 added the declaration of and 5 uses of `nonNullUnknownType`, #49119 removed 3 references, and this PR removes the 2 remaining references and declaration.


jakebailey:
@typescript-bot test top200
@typescript-bot run dt

fatcerberus:
> Probably due to the changes to how `unknown` is more or less `{} | undefined | null`.

Oh no - I hope it's not distributive like `boolean` :eyes:

Andarist:
> Oh no - I hope it's not distributive like boolean 👀

It's not :D although internally such a type kinda exists (`unknownUnionType`) - I don't think it's particularly special though, it's just a literal internal creation of a union that contains those 3 types

gabritto (approved):

jakebailey:
@typescript-bot pack this

jakebailey:
(I intentionally merged before the above result; I was just testing that a recent change in the error deltas repo was working.)

PR #57668, opened 2024-03-06
Remove es3 related tests

jakebailey:
Follow-up to #57525. There aren't really that many tests that appeared to be testing explicitly es3 only things.


src/harness/harnessIO.ts
```diff
@@ -1082,6 +1082,13 @@ function getVaryByStarSettingValues(varyBy: string): ReadonlyMap<string, string
     const option = ts.forEach(ts.optionDeclarations, decl => ts.equateStringsCaseInsensitive(decl.name, varyBy) ? decl : undefined);
     if (option) {
         if (typeof option.type === "object") {
+            if (option.deprecatedKeys) {
+                const map = new Map(option.type.entries());
+                for (const key of option.deprecatedKeys) {
```
RyanCavanaugh: Should we be warning or something here?

RyanCavanaugh (approved):

src/harness/harnessIO.ts
jakebailey: This was leftover from me trying to figure out why I _added_ a new baseline. Basically we have this syntax in the harness I had never seen, `*,-es3`, which says "everything but es3". But since we need to still have es3 in the map to know how to parse it (and then error on it), it's there in the list.

jakebailey: (then I forgot to unstage this change)

RyanCavanaugh: Oh, I misunderstood when this code runs. Makes sense

tests/baselines/reference/sourceMap-LineBreaks.js
```diff
@@ -3,14 +3,17 @@
...
var endsWithLineFeed = 1;
 var endsWithCarriageReturnLineFeed = 1;
-var endsWithCarriageReturn = 1;
var endsWithLineFeedCarriageReturn = 1;
-
var endsWithLineFeedCarriageReturnLineFeed = 1;
+var endsWithCarriageReturn = 1;
+var endsWithLineFeedCarriageReturn = 1;
+
+var endsWithLineFeedCarriageReturnLineFeed = 1;
```
jakebailey, line 9: I am trying so hard to get this test back the way it was, but every editor is fighting me

jakebailey: Finally, turns out it's git's fault and I had to force squash and then fix in order to not have any intermediary diffs

PR #57669, opened 2024-03-06
Hoist function exports to top of module body in CJS/AMD/UMD

### LINKED ISSUE #39853: Hoisted function declarations should be assigned to `exports` before any `require` calls
**TypeScript Version:**  4.0.0-beta

**Search Terms:**

- hoisted function
- exports hoisted function
- export hoisted function
- export hoist function

**Code**

```js
// a.js
import { b } from "./b.js";

export function a() {
    return b();
}

// b.js
import { a } from "./a.js";

// Crashes when transpiled by TypeScript:
a();

export function b() {
    return 123;
}
```

**Expected behavior:**

The `exports.*` assignment should be done immediately after `__esModule`, just like how Babel does it.

```js
// a.js
"use strict";
exports.a = a;
const b_1 = require("./b.js");
function a() {
    return b_1.b();
}

// b.js
"use strict";
exports.b = b;
const a_js_1 = require("./a.js");
// Crashes when transpiled by TypeScript:
a_js_1.a();
function b() {
    return 123;
}
```

**Actual behavior:**

The exports assignment is performed after the function declaration, which causes behaviour differences between native **ES** modules (where hoisted function declarations are initialised before any code begins executing) and transpiled **CommonJS** modules (where hoisted function declarations are added to the `exports` object after `require(…)` calls)

```js
// a.js
"use strict";
exports.a = void 0;
const b_1 = require("./b.js");
function a() {
    return b_1.b();
}
exports.a = a;

// b.js
"use strict";
exports.b = void 0;
const a_js_1 = require("./a.js");
// Crashes when transpiled by TypeScript:
a_js_1.a();
function b() {
    return 123;
}
exports.b = b;
```

**Playground Link:** [a.js](https://www.staging-typescript.org/play?module=1&ts=4.0.0-beta&useJavaScript=true#code/JYWwDg9gTgLgBAbzgIzgXzgMyhEcBEAdAPTKEBWAzvgNwBQdApgB6SxYCuAdgMYzAQucAIYAKAJSI6cGXCiMYHKEOQT6aIA), [b.js](https://www.staging-typescript.org/play?module=1&ts=4.0.0-beta&useJavaScript=true#code/JYWwDg9gTgLgBAbzgQzgXzgMyhEcBEAdAPTKEBWAzvgNwBQdxxcAwlMpQBYCmlcA7jwB2cGOyGUwwADbcAJnABGATzgAVZWG4BlAMZRgYGAC46yABQBKenW4APSLCwBXIbpjAIIxVcR04AXBQ3DDOUCIAjABMAMz0aEA)

**Related Issues:** <https://github.com/microsoft/TypeScript/pull/37093>

rbuckton:
This changes our emit for CommonJS, AMD, and UMD so that exports for exported function declarations are hoisted to the top of the transformed module body. There is a minor impact on the order of keys in the `exportedNames_1` map synthesized for SystemJS modules as a consequence of this change, but it has no impact to runtime behavior.

Fixes #39853


andrewbranch (approved):

jakebailey:
I just merge conflicted this via #57668; you'll have to merge main to delete some of these baselines. (sorry!)

tjzel:
@rbuckton Is there an option to disable this behavior while generating CommonJS?

rbuckton:
> @rbuckton Is there an option to disable this behavior while generating CommonJS? 

No, there is not. It's there a reason you would need to disable it?

tjzel:
[Reanimated Babel plugin](https://github.com/software-mansion/react-native-reanimated/tree/main/packages/react-native-reanimated/plugin) does code transformations that break hoisting - function declarations become variable declarations with assignments. Therefore with this change we also have to [move the exports to the bottom of file](https://github.com/software-mansion/react-native-reanimated/pull/6316) so the values are not exported as `undefined`. It's not really a blocker, but an inconvenience nevertheless.

PR #57670, opened 2024-03-06
Track minimum language version for various language features

### LINKED ISSUE #41241: Reevaluate checkExternalEmitHelpers() callers
**TypeScript Version:**  3.7.x-dev.201xxxxx

**Search Terms:** importHelpers

**Code**

Some callers still checks against ESNext for non-esnext stable features.

https://github.com/microsoft/TypeScript/blob/94b0e0e389f5328d1e909c4efd951dbd882829f8/src/compiler/checker.ts#L31351-L31353

**Expected behavior:** They should check against ES20xx target.

**Actual behavior:** They do against ESNext

**Playground Link:** 

**Related Issues:**

rbuckton:
This adds `LanguageFeatureMinimumTarget`, which is used to track the specific ECMAScript edition that various language features were standardized. This is intended to be a central location to check for when specific features are added to the language and provide a single place to update a feature's minimum target from `ESNext` to the edition it landed in.

In keeping with this, all of our emit helper checks have been tightened up to reference the correct version for when a given emit helper is not necessary during emit.

Fixes #41241

### DIFF
--- a/src/compiler/types.ts
+++ b/src/compiler/types.ts
@@ -7277,6 +7277,9 @@ export const enum ScriptKind {
     Deferred = 7,
 }
 
+// NOTE: We must reevaluate the target for upcoming features when each successive TC39 edition is ratified in
+//       June of each year. This includes changes to `LanguageFeatureMinimumTarget`, `ScriptTarget`,
+//       transformers/esnext.ts, commandLineParser.ts, and the contents of each lib/esnext.*.d.ts file.
 export const enum ScriptTarget {
     /** @deprecated */
     ES3 = 0,
@@ -8087,6 +8090,9 @@ export namespace LanguageFeatureMinimumTarget {
     export const ShebangComments = ScriptTarget.ESNext;
 
     // Upcoming Features
+    // NOTE: We must reevaluate the target for upcoming features when each successive TC39 edition is ratified in
+    //       June of each year. This includes changes to `LanguageFeatureMinimumTarget`, `ScriptTarget`,
+    //       transformers/esnext.ts, commandLineParser.ts, and the contents of each lib/esnext.*.d.ts file.
     export const UsingAndAwaitUsing = ScriptTarget.ESNext; // `using x = y`, `await using x = y`
     export const ClassAndClassElementDecorators = ScriptTarget.ESNext; // `@dec class C {}`, `class C { @dec m() {} }`
 }
--- a/src/compiler/types.ts
+++ b/src/compiler/types.ts
@@ -8043,58 +8043,58 @@ export type EmitHelperUniqueNameCallback = (name: string) => string;
  *
  * @internal
  */
-export namespace LanguageFeatureMinimumTarget {
+export const enum LanguageFeatureMinimumTarget {
     // ES2015 Features
-    export const Classes = ScriptTarget.ES2015;
-    export const ForOf = ScriptTarget.ES2015;
-    export const Generators = ScriptTarget.ES2015;
-    export const Iteration = ScriptTarget.ES2015;
-    export const SpreadElements = ScriptTarget.ES2015;
-    export const RestElements = ScriptTarget.ES2015;
-    export const TaggedTemplates = ScriptTarget.ES2015;
-    export const DestructuringAssignment = ScriptTarget.ES2015;
-    export const BindingPatterns = ScriptTarget.ES2015;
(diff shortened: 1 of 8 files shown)


sandersn (approved):
I like this idea, and would love for it to include a way to remember to update lib/esnext.*.d.ts. (Right now I think our intended approach is to manually check each year after the June TC39 meeting.)

src/compiler/types.ts
```diff
@@ -8035,6 +8035,62 @@ export type UniqueNameHandler = (baseName: string, checkFn?: (name: string) => b
 
 export type EmitHelperUniqueNameCallback = (name: string) => string;
 
+/**
+ * Indicates the minimum `ScriptTarget` (inclusive) after which a specific language feature is no longer transpiled.
+ *
+ * @internal
+ */
+export namespace LanguageFeatureMinimumTarget {
```
jakebailey, line 8043: Can we turn this into a const enum such that it gets inlined? I believe we should be able to reference ScriptTarget from one. Or does that fail because of the type mismatch with ScriptTarget?

(Would like to limit our use of namespaces, honestly... plus potential perf hits here)

RyanCavanaugh:
This is great

src/compiler/types.ts
RyanCavanaugh: Or just a big `const` since we don't really need the reverse lookup

rbuckton: It doesn't fail to match `ScriptTarget`, but it does produce its own type which doesn't illustrate the relationship when looking at the `.d.ts` file:

image

vs.

image

vs.

image


I actually still like `namespace` in moderation. This felt like an appropriate use for it, but I don't mind the `as const`. There's no perf difference between `namespace` and `as const`.

jakebailey: This is an internal declaration, so I don't personally feel like it's a big deal for the values to be opaque. It does make me wonder if we're missing something in declaration emit that would have reused those nodes, though.

rbuckton: Namespaces, as a concept, exist in both the standard library (`Math`, `Atomics`, `Reflect`) and the web platform (`subtle`, `performance`, etc.) so I don't agree that they should be wholly discouraged.

rbuckton:
> I like this idea, and would love for it to include a way to remember to update lib/esnext.*.d.ts. (Right now I think our intended approach is to manually check each year after the June TC39 meeting.)

The best we might be able to do is to just document this, `ScriptTarget`, and `transformESNext` with comments. We will still need to watch for the June Meeting regardless.

src/compiler/types.ts
jakebailey: In type declarations, sure, I'm just trying to avoid the runtime cost of them (e.g. const enums are inlined, modules are tree shaken).

rbuckton:
@typescript-bot perf test

src/compiler/types.ts
rbuckton: If the perf results show a significant dip, I can either make it a const enum, it just feels less useful when hover information only shows the final value like `2` instead of something meaningful like `ES2015`.

jakebailey: I see what you mean; I do wish our hover and declaration emit did better here. But, we seem to just always compute in dts emit (not sure about hover).

rbuckton: > I see what you mean; I do wish our hover and declaration emit did better here. But, we seem to just always compute in dts emit (not sure about hover).

It's arguable as to whether someone who writes `const enum E2 { A = E1.B }` would expect hover/d.ts to print the value or qualified name. With `const` variables or fields it's far clearer as to what to show. I could see maybe supporting `const enum E2 { A = E1.B as E1.B }` or something similar to explicitly opt-in to showing the qualified name, but that seems like a very niche use case.

jakebailey: I would much rather see `(enum member) TypeFlags.Unit = Enum | Literal | UniqueESSymbol | Nullable` than `(enum member) TypeFlags.Unit = 109472`. The actual computed values seem like the least important part of an enum.

rbuckton:
@typescript-bot perf test

rbuckton:
It looks like the `enum` version is slightly faster so I'll go with that. We might want to look into quick info/.d.ts emit changes for enums in a future PR.

jakebailey (approved):

PR #50535, opened 2022-08-30
Only normalize intersections that include {}

ahejlsberg:
Fixes #50515.

PR #57673, opened 2024-03-07
Fixed excess and common property checks with `NoInfer`

Andarist:
I figured out that it might be better to handle `NoInfer` at those levels because there is always a risk that `isWeakType`/`isKnownProperty` might get called outside of the `checkTypeRelatedTo`.

However, this could potentially also be fixed within `getNormalizedType` - I briefly took a look at this possible solution and it felt like it could be slightly more complicated and that it could potentially result in potential gotchas. Normally, each constituent gets normalized while the type gets broken down in the relationship check - not when the union/intersection gets first observed by it. In fact, at one moment in time normalization for intersection members was introduced by https://github.com/microsoft/TypeScript/pull/49119 and then partially reverted in https://github.com/microsoft/TypeScript/pull/50535 (the logic got adjusted to only normalize intersections containing `{}`).

If you think that it's a wrong judgement call - please let me know and I'll adjust the PR to put this logic into `getNormalizedType`. There is also a risk that normalization earlier could fix some other things. For the most part those other things are likely working correctly already because those types get normalized - just later on.

fixes https://github.com/microsoft/TypeScript/issues/57697


src/compiler/checker.ts
```diff
@@ -23756,6 +23756,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             return resolved.callSignatures.length === 0 && resolved.constructSignatures.length === 0 && resolved.indexInfos.length === 0 &&
                 resolved.properties.length > 0 && every(resolved.properties, p => !!(p.flags & SymbolFlags.Optional));
         }
+        if (isNoInferType(type)) {
```
ahejlsberg, line 23759: I think it would be better to erase all substitution types (to their base types) instead of just `NoInfer` types.

ahejlsberg (approved):
Approved, but would prefer to erase all substitution types as I suggest.

src/compiler/checker.ts
Andarist: I pushed out the requested change but without any new tests.

I've tried a bunch of things based on my intuition in an attempt to write some tests for this change and nothing - from my attempts - managed to hit those lines with a non-NoInfer substitution type. So I went to recheck if I could observe some cases like this using the existing test suite:
- `isExcessPropertyCheckTarget`/`isKnownProperty` didn't observe such substitution types at all
- `isWeakType` has observed it in [`deeplyNestedMappedTypes`](https://github.com/microsoft/TypeScript/blob/ede8ad84b1b2708a0eea234babd5bbf89fda57b8/tests/cases/compiler/deeplyNestedMappedTypes.ts) and [`conditionalTypeDoesntSpinForever`](https://github.com/microsoft/TypeScript/blob/ede8ad84b1b2708a0eea234babd5bbf89fda57b8/tests/cases/compiler/conditionalTypeDoesntSpinForever.ts). Those are complex enough that figuring out a test case based solely on them might take me a longer moment.

ahejlsberg: Yeah, it's not going to be easy to come up with a case where it matters and I'm fine going without.

jakebailey:
This PR was not up to date with main and broke baselines.

ahejlsberg:
Argh!

PR #57678, opened 2024-03-07
 [ID-Prep] PR5 - Preserve type nodes from function like expression in declaration emit

### LINKED ISSUE #57677: [ID-Prep] Preserve type nodes from function expressions in initialisers
### 🔍 Search Terms

declaration emit preserve types from functions

### ✅ Viability Checklist

### ⭐ Suggestion

Preserve types from function expressions in initializers when emitting declarations. 

### 📃 Motivating Example

```ts

// index.ts
export let fn = (p: P, p2:typeof p) => {}

// index-current.d.ts
export declare let fn: (p: P, p2: {
    foo: boolean;
}) => void;
export {};

// index-proposed.d.ts
export declare let fn: (p: P, p2: typeof p2) => void;
export {};

```

### 💻 Use Cases

    An external tool can extract the type from the function expression, but then they would have different declarations. It would be great if, where possible, typescript would also preserve types as written.
    We could force users to always specify the type even when it's obvious from the function expression, but this makes for very frustrating DX

dragomirtitian:
Fixes #57677


src/compiler/checker.ts
```diff
@@ -6329,8 +6329,25 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
         const firstIdentifier = getFirstIdentifier(entityName);
         const symbol = resolveName(enclosingDeclaration, firstIdentifier.escapedText, meaning, /*nameNotFoundMessage*/ undefined, /*nameArg*/ undefined, /*isUse*/ false);
-        if (symbol && symbol.flags & SymbolFlags.TypeParameter && meaning & SymbolFlags.Type) {
-            return { accessibility: SymbolAccessibility.Accessible };
+        if (symbol) {
+            if (symbol.flags & SymbolFlags.TypeParameter && meaning & SymbolFlags.Type) {
+                return { accessibility: SymbolAccessibility.Accessible };
+            }
+
+            // Parameters or binding elements from parameters are always visible in their enclosing function declarations
```
dragomirtitian, line 6337: The checker complained about the parameter name not being accessible (since arrow functions are not themselves marked as visible by `determineIfDeclarationIsVisible`.

src/compiler/checker.ts
```diff
@@ -48160,13 +48177,25 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             hasSyntacticModifier(parameter, ModifierFlags.ParameterPropertyModifier);
     }
 
-    function isExpandoFunctionDeclaration(node: Declaration): boolean {
-        const declaration = getParseTreeNode(node, isFunctionDeclaration);
+    function isExpandoFunctionDeclaration(node: FunctionDeclaration | VariableDeclaration): boolean {
```
dragomirtitian, line 48180: Both function and variable declarations will now be tested. We can't reliably determine the type for an expando functions variable from just it's declaration so we fall back on the type checker.

PR #57679, opened 2024-03-07
Add missing parameters from Array.toLocaleString on ES2015 libs

### LINKED ISSUE #57603: Missing arguments in "toLocaleString" method of Array
### 🔎 Search Terms

array toLocaleString arguments locale

- This is a crash

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.3.3#code/MYewdgzgLgBAhgJwXAnjAvDA2gIgBICmANkSDAOogJEAmOANDDgISs4C6AUKJCEQQDpSAcwAUiZCgFQQAGRDA4-AMpQEASzBiA5AQjaAlAaA

### 💻 Code

```ts
const array = ["Hello World", "!!!"]
console.log(array.toLocaleString('es'))
```

### 🙁 Actual behavior

Expected 0 arguments, but got 1.

### 🙂 Expected behavior

No error

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toLocaleString#syntax

neilbryson:
Fixes #57603 

Added on both `Array` and `ReadonlyArray` in es2015.core.d.ts

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toLocaleString


src/lib/es2015.core.d.ts
```diff
@@ -42,6 +42,20 @@ interface Array<T> {
...
+     * @param locales A string with a BCP 47 language tag, or an array of such strings. For the general form and
+     * interpretation of the locales argument, see
+     * [the parameter description on the Intl main page](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl#locales_argument).
+     * @param options An object with configuration properties. For numbers, see
+     * [Number.prototype.toLocaleString()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString#options);
+     * for dates, see [Date.prototype.toLocaleString()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString#options).
+     *
+     * [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toLocaleString)
+     */
+    toLocaleString(locales?: string | string[], options?: Intl.NumberFormatOptions & Intl.DateTimeFormatOptions): string;
```
sandersn, line 58: 1. Does NumberFormatOptions & DateTimeFormatOptions cover everything? I see that BigInt also has toLocaleString. Could other types have it too?
2. since this adds an overload, `locales` should not be optional. Otherwise a zero-argument call might resolve to this when it should resolve to the es5 overload.
```suggestion
    toLocaleString(locales: string | string[], options?: Intl.NumberFormatOptions & Intl.DateTimeFormatOptions): string;
```

src/lib/es2015.core.d.ts
```diff
@@ -342,6 +356,20 @@ interface ReadonlyArray<T> {
...
```
sandersn, line 372:
```suggestion
    toLocaleString(locales: string | string[], options?: Intl.NumberFormatOptions & Intl.DateTimeFormatOptions): string;
```

sandersn (changes requested):
Couple of questions, plus I need to find out what we need to do legally to use MDN text in Typescript.

src/lib/es2015.core.d.ts
neilbryson: Thanks for the comments!

1. When I checked, it looks like there are only NumberFormatOptions and DateTimeFormatOptions which is accepted by JS in all types with this method (Object, Array, Number, BigInt, TypedArray). BigInt arrays are covered by the current definitions upon testing, but I will still need to update the `toLocaleString` signature of TypedArrays. I'll send a commit for that 

2. I tried not making it optional, but it seems to stop working with single argument calls based on my test run
```
arrayToLocaleStringES2015.ts(4,11): error TS2575: No overload expects 1 arguments, but overloads do exist that expect either 0 or 2 arguments.
arrayToLocaleStringES2015.ts(10,14): error TS2575: No overload expects 1 arguments, but overloads do exist that expect either 0 or 2 arguments.
  ```
  I think they should be kept optional, what do you think? In my [baselines errors](https://github.com/microsoft/TypeScript/blob/4752316627bbd9e634c5d31778119df8fac89781/tests/baselines/reference/arrayToLocaleStringES5.errors.txt#L1), it was showing the expected behaviour for es5

Comment (sandersn) [reply src/lib/es2015.core.d.ts:46]: 1. 👍🏼 
2. The error makes it look like both `locales` and `options` are both required. But only `locales` should be required; `options` should stay optional. When I did a quick local test with a test project using es2015, all 3 calls below worked:

```ts
const a = [1,2,3]
a.toLocaleString()
a.toLocaleString('en-US')
a.toLocaleString('en-US', {style: 'currency', currency: 'USD'}) // thanks copilot
```

sandersn (changes requested):
I'd still like locales to be required and options to be optional.

src/lib/es2015.core.d.ts
neilbryson: Hello! I applied the changes to es2015

sandersn (approved):
Looks good!

PR #57681, opened 2024-03-07
Do not perserve references in declaration emit, unless preserve=true

### LINKED ISSUE #15488: <reference types="" /> is not emitted when the dependency is indirect
**TypeScript Version:**  2.3.2

This is similar to #15487.

**Code**

```ts
// deps/dep/dep.d.ts
/// <reference path="indirect.d.ts" />
```

```ts
// deps/dep/indirect.d.ts
interface Dep {
}
```

```json
// deps/dep/package.json
{
  "typings": "dep.d.ts"
}
````
----

```ts
// src/index.ts
class Src implements Dep {}
```

Compile with

```json
// src/tsconfig.json
{
    "compilerOptions": {
        "declaration": true,
        "types": [
            "dep"
        ],
        "typeRoots": [
            "../deps"
        ]
    }
}
```

**Expected behavior:**

```index.d.ts
/// <reference types="dep" />
declare class Src implements Dep {
}
```

**Actual behavior:**

```index.d.ts
declare class Src implements Dep {
}
```

### LINKED ISSUE #26948: Auto type reference directives do not lowercase the typing name
As part of #9824 all type reference directives are lower cased so type="somePackage" results in looking for type reference of `somepackage` instead.

When including the auto type directives, we include list of typings from user or get the package names from typeRoots (where if config doesn't provide it, it will be node_modules/@types)
The package names read here are not lower cased. Thus this leaves a discrepancy between these type sources included and typereference directives included resulting in confusion for packages in these auto type directives to reference other packages at same level.

Eg.
If auto type references are: "UpperCasePackage", "somepackage", and if somepackage needs to reference this `UpperCasePackage` it has no way to do so correctly.

Found while working on #26944

### LINKED ISSUE #28195: Compiling with -d should emit triple-slash lib paths
## Search Terms

Triple slash, reference, path, lib, library, types, typings, dom

## Suggestion

When compiling a *.ts* file with `declaration: true`, TSC should emit triple-slash reference paths for all `lib`s consumed by the *.ts* file.

## Problem

Say I publish typings for my package `foo` to NPM, and `foo` depends on DOM typings. I then include `foo` in my project (`npm install foo`). There are a few things that can happen:

1. If `foo`'s typings include `<reference lib="dom" />`:
  - `foo` has access to DOM
  - My project has access to DOM types (**is this sort of leakage desirable?**)

2. If my project includes `dom` (or uses the same triple-slash directive as in 1):
  - `foo` has access to DOM
  - My project has access to DOM types

3. Neither my project nor `foo` include `<reference lib="dom" />`:
  - `foo` errors wherever it uses DOM types (but I have to inspect *node_modules/foo/index.d.ts* to see those errors)
  - My project treats `foo`'s DOM types as `any`s (**should this error in `strict` mode?**)

    ```ts
    // foo.d.ts
    export function a(): HTMLInputElement

    // myProject/index.ts
    import {a} from 'foo'
    a() // any
    ```

## Solution

If `tsc -d` generated triple-slash directives for `foo/index.d.ts`, we can ensure that we're always in case (1). This is nice because if my project doesn't use `dom`, I shouldn't have to declare it for the sake of `foo` like I did in (2); it also avoids the `any` from (3).

```ts
// Before
export function a(): HTMLInputElement

// After
/// <reference lib="dom" />
export function a(): HTMLInputElement
```

An open question is how strict should TSC be? Ie. If `foo` uses `dom` but fails to reference it, should this cause a compile error in my project (and a nice message suggesting I add a reference to `dom`)?

## Checklist

* [ ] This wouldn't be a breaking change in existing TypeScript / JavaScript code
* [x] This isn't a runtime feature (e.g. new expression-level syntax)

### LINKED ISSUE #45096: lowercase filename in triple slash directives
# Bug Report
Referencing a file name in triple slash directives automatically looks for the lower case filename

### 🔎 Search Terms
- triple slash directive
- lowercase filename

have seen it since ts > 3.x

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about triple slash directives / lowercase filename

### ⏯ Playground Link
not sure how to reproduce it in the playground
This requires multiple files, and there is no easy way to verify

### 💻 Code
```
// fileOne.d.ts
...
...
```

```
// file2.d.ts
// <reference types="../fileOne.d.ts">
...
...
```

### 🙁 Actual behavior
on case sensitive OS, i got a warning from tscompiler about the casing being incorrect.

### 🙂 Expected behavior
should not get this warning

More Info:
I guess this is related to https://github.com/microsoft/TypeScript/issues/26948
and it seems this is correct behaviour.

If that is the case, it would be great to document it alongside triple slash directives or add it to FAQs
https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html

### LINKED ISSUE #48143: duplicate emit `/// <reference types="node" />`
# Bug Report

Version 4.7.0-dev.20220305

- This is a crash

### ⏯ Playground Link

### 💻 Code

```ts
// We can quickly address your report if:
//  - The code sample is short. Nearly all TypeScript bugs can be demonstrated in 20-30 lines of code!
//  - It doesn't use external libraries. These are often issues with the type definitions rather than TypeScript bugs.
//  - The incorrectness of the behavior is readily apparent from reading the sample.
// Reports are slower to investigate if:
//  - We have to pare too much extraneous code.
//  - We have to clone a large repo and validate that the problem isn't elsewhere.
//  - The sample is confusing or doesn't clearly demonstrate what's wrong.
```

### 🙁 Actual behavior

https://github.com/bluelovers/ws-iconv/commit/c039e942f8562c9fdbba4e8d50c19afc20cc687f

https://github.com/bluelovers/ws-iconv/blob/c039e942f8562c9fdbba4e8d50c19afc20cc687f/packages/%40lazy-node/readlines/lib/types.d.ts

```ts
/// <reference types="node" />
/// <reference types="node" />
/**
 * Created by user on 2020/5/29.
 */
import { PathLike } from 'fs';
export interface IOptions {
    readChunk?: number;
    newLineCharacter?: INewLineCharacter;
}
export declare const enum EnumNewLineCharacter {
    LF = 10,
    CR = 13
}
export declare type IPathLike = PathLike | number;
export declare type INewLineCharacter = string | number | EnumNewLineCharacter | (string | number | EnumNewLineCharacter)[] | Buffer;
```

### 🙂 Expected behavior

```ts
/// <reference types="node" />
/**
 * Created by user on 2020/5/29.
 */
import { PathLike } from 'fs';
export interface IOptions {
    readChunk?: number;
    newLineCharacter?: INewLineCharacter;
}
export declare const enum EnumNewLineCharacter {
    LF = 10,
    CR = 13
}
export declare type IPathLike = PathLike | number;
export declare type INewLineCharacter = string | number | EnumNewLineCharacter | (string | number | EnumNewLineCharacter)[] | Buffer;
```

### LINKED ISSUE #52110: Casing is not preserved when compiling `<reference path="...">` triple slash directives
# Bug Report

### 🔎 Search Terms

reference types

This is the behavior in every version I tried, and I reviewed the FAQ for entries about triple slash directives.

### ⏯ Playground Link

Unable to produce playground link, as this involves the use of several files.

GitHub repository: https://github.com/craxal/triple-slash-path

#### module-one/PanelWindow.d.ts
```ts
declare global {
    interface CustomWindow {
        getSomething(): any;
    }

    interface Window {
        customWindow: CustomWindow;
    }
}
```

#### module-two/Class.ts
```ts
// <reference path="../../node_modules/storage-explorer/dist/PanelWindow.d.ts" />

import * as $ from "jquery";
import { IContextMenuItem } from "storage-explorer";

export interface ICustomType {
    someProperty: string;
}

export async function doSomething(args: ICustomType): any {
    const obj = await window.customWindow.getSomething();
    return obj;
}
```

### 🙁 Actual behavior

The triple slash directive is compiled to something similar to the following (note the casing is not preserved, even if `forceConsistentCasingInFileNames` is set to `true`):

```js
// <reference types="module-one/customwindow" />
```
Because of this issue, I cannot compile on macOS. This is not an issue on Windows.

### 🙂 Expected behavior

I expect the triple slash directive in module-two/Class.ts to be compiled to something similar to the following (note the casing ought to be preserved if `forceConsistentCasingInFileNames` is set to `true`):
```js
// <reference types="module-one/CustomWindow" />
```

### LINKED ISSUE #56590: Multiple redundant tripleslash directives can be emitted
### 🔎 Search Terms

tripleslash directive modules reference

### ⏯ Playground Link

https://www.typescriptlang.org/play?target=99&moduleResolution=3&module=100&esModuleInterop=false#code/PTAEAEFsHsBMFcA2BTAXKAcnZBGAbALABQIEMCKASsgM7SLwAuAltAHbpay6EljjcAxogCGAJxEt26RmPjJixEGAA8Y5ADNk6toOShGATwAOtALwAiNtguhgAPkVFmkY9DGNQAbwCqlADIAvqAaYtCQoADk1tyo8GKIkQDcxC5uHt4AogBuyGyMmZDMjIzawaHhUTFoyLn5NMmpru6eXgBC8BpaYuVhEdHYqABGnd2NRMTIAB7pnoLsNJ6Lo+gA2n7+APwANKA5eQVFJdo7oB1dJwC6oGagq2zIAO6gGwAUkQAWJcY0qCCPAIAdEZTDRBGJmMZGKI2ABzQHuWHAYyiQyRACUlyAA

### 💻 Code

```ts
// @module: Node16
// @moduleResolution: Node16
// @declaration: true

/// <reference types="node" />

import {URL} from 'node:url';
import {EventEmitter} from 'node:events';
import {Buffer} from 'node:buffer';

export const stuff: [URL?, EventEmitter?, Buffer?] = [new URL('https://www.typescriptlang.org/play')]
```

### 🙁 Actual behavior

```ts
/// <reference types="node" resolution-mode="require"/>
/// <reference types="node" resolution-mode="require"/>
/// <reference types="node/url.js" />
/// <reference types="node/ts4.8/url.js" />
/// <reference types="node/events.js" />
/// <reference types="node/ts4.8/events.js" />
/// <reference types="node/buffer.js" />
/// <reference types="node/ts4.8/buffer.js" />
import { URL } from 'node:url';
import { EventEmitter } from 'node:events';
import { Buffer } from 'node:buffer';
export declare const stuff: [URL?, EventEmitter?, Buffer?];
```

### 🙂 Expected behavior

I'd expect only a single `/// <reference types="node" resolution-mode="require"/>` to be emitted.

Note that types acquisition and switching between ESM/CJS modes is a little bit unreliable in the TS playground so you might have to "touch" the repro to "refresh" the DTS output

### LINKED ISSUE #57439: [ID-Prep] Preserve all triple slash directives in declarations files

### LINKED ISSUE #57482: triple-slash reference added only if another triple-slash reference exists
### 🔎 Search Terms

triple-slash reference added

### ⏯ Playground Link

[Workbench](https://www.typescriptlang.org/dev/bug-workbench/?#code/PTAEAEBMFMGMBsCGAnRAXAlgewHYC5Q1kBXaAWACgQIAzDeaHRAW2gOEWByxgH1mexBgGdg0AB5FOGHDHEA6SPLTDKMtNGQ1EsaKACiyZFmQAmUAG9KoG6GFodAa1MF7yGQHMA3JQC+lSmpwOgYmVnZObj4BSCFoUWEsVjQAC09gGTlFZVUqagAeZGgaTUZdQgBPAAd4gF4AIgkpetBgAD5AsAA9UCKBADc9RAAjLEHQRFlQDDRQAHd6eFBuUEgMIthMcYX4JeGhyBhIQiwJqqrslUoJKpNZmmIcTewcUBosLAAKAEo8Q2MzD4KJ1aPRGCw2K1OIhzjkQUUSkUnno0Kcmqh5otQPsJodoMcUqVpjRCISiqAJBh7MIJtxUppesVSsiThTJKhruJbsh7o9nrg3h8fpZrLYimhiMhXjghPAAIQTGn-EymIG+IA)

### 💻 Code

```ts
// @declaration: true
// @filename: /a/node_modules/extra/index.d.ts
interface Error2 {
    stack2: string;
}

// @filename: /a/node_modules/something/index.d.ts
/// <reference types="extra" />
// ^ remove above and it will no directive will be added to app.d.ts
export function foo():Error2;

// @filename: /a/app.ts
// reference to extra will be added here if there exists another reference to extra
export function foo() {
    return null! as Error2;
}
```

### 🙁 Actual behavior

`/// <reference types="extra" />`  is added to `app.d.ts` only if another reference directive to extra already exists (in this example one from `node_modules`)

### 🙂 Expected behavior

`/// <reference types="extra" />` is always added since `extra` is required for `app.d.ts` to be valid

This was discovered while trying to understand reference directives in the context of isolated declarations.

jakebailey:
At the moment, declaration emit does a funky mixture of preserving some references, eliding unused references, adding of referenced files for ambients (like node's builtins), but also sometimes imports (react), and so on. For external implementations of `isolatedDeclarations` (and honestly, the human brain), this is either difficult to reason about or totally impossible to replicate.

This PR greatly changes the way we handle references. References are no longer preserved in the output from the input, nor are they ever added based on any analysis. If a reference is intended to be preserved, one must write `preserve="true"` to have it copied to the output file. This gives a "what you see is what you get" format, not unlike what `verbatimModuleSyntax` does for imported values.

This is technically a break, as packages which re-export from something like `@types/node` will no longer have implicitly added reference directives for those types, however this change is in line with how we handle lib.d.ts; it's assumed that the user will have actually included the environment's types into their project. Testing in prep for this change (#57569, #57656) showed that most instances of added references fit into this bucket and should continue to work (and why we are shipping this without a new flag).

This change seems to improve performance a bit (1-4% in benchmarks with dts emit enabled), likely due to the reduced bookkeeping.

The only non-obvious thing that remains is the rewriting of file references, which includes the changing of import extensions (funny) and the shifting of paths to be relative to the output paths. This specific analysis is more reasonable to implement externally as it's just a function of file paths and compiler options.

Fixes #57439 (though does the opposite by default)

It also fixes these issues by nature of us no longer synthesizing new references, or just keeping the references as written:

Fixes #15488
Fixes #26948
Fixes #45096
Fixes #48143
Fixes #52110
Fixes #56590
Fixes #57482

Then in terms of design, the direction is now "we don't generate references out of thin air anymore", in which case this also closes:

Closes #28195

Related: #56571


jakebailey:
A significant number of our tests break because they don't include the new references; is this an indicator that the default should actually be preserve=true? Or that our tests are "weird code heavy"?

jakebailey:
The errors are declaration errors in the emit, since references that were needed were not preserved by default.

src/compiler/transformers/declarations.ts
```diff
@@ -511,64 +470,47 @@ export function transformDeclarations(context: TransformationContext) {
...
-            refs.forEach(referenceVisitor);
-            emittedImports = filter(combinedStatements, isAnyImportSyntax);
+            rawReferencedFiles.forEach(referenceVisitor);
             if (isExternalModule(node) && (!resultHasExternalModuleIndicator || (needsScopeFixMarker && !resultHasScopeMarker))) {
                 combinedStatements = setTextRange(factory.createNodeArray([...combinedStatements, createEmptyExports(factory)]), combinedStatements);
             }
         }
-        const updated = factory.updateSourceFile(node, combinedStatements, /*isDeclarationFile*/ true, references, getFileReferencesForUsedTypeReferences(), node.hasNoDefaultLib, getLibReferences());
+        const updated = factory.updateSourceFile(node, combinedStatements, /*isDeclarationFile*/ true, referencedFiles, getTypeReferences(), node.hasNoDefaultLib, getLibReferences());
         updated.exportedModulesFromDeclarationEmit = exportedModulesFromDeclarationEmit;
```
jakebailey, line 497: I haven't removed this prop yet, but it seems like it probably shouldn't be needed anymore? I think this is used for build mode / incremental and now that the determination of whether or not references are included is entirely file local, I don't think it's possible for us to need to keep track of anything else anymore?

src/compiler/parser.ts
```diff
@@ -10510,19 +10510,20 @@ export function processPragmasIntoFields(context: PragmaContext, reportDiagnosti
                 const typeReferenceDirectives = context.typeReferenceDirectives;
                 const libReferenceDirectives = context.libReferenceDirectives;
                 forEach(toArray(entryOrList) as PragmaPseudoMap["reference"][], arg => {
-                    const { types, lib, path, ["resolution-mode"]: res } = arg.arguments;
+                    const { types, lib, path, ["resolution-mode"]: res, preserve: _preserve } = arg.arguments;
+                    const preserve = !!_preserve ? true : undefined;
```
jakebailey, line 10514: This seems super weird. See also #57782.

tests/baselines/reference/nodeModulesTripleSlashReferenceModeDeclarationEmit3(module=node16).js
```diff
@@ -25,15 +25,54 @@ declare global {
...
-/// <reference types="pkg" resolution-mode="require"/>
+/// <reference types="pkg" preserve="true" />
 export interface LocalInterface extends RequireInterface {
 }
+
+
+//// [DtsFileErrors]
+
+
+out/index.d.ts(2,41): error TS2304: Cannot find name 'RequireInterface'.
```
jakebailey, line 45: This test is super weird and I'm not finding any way to make this error go away, even though `pkg` is preserved?

tests/baselines/reference/nodeModulesTripleSlashReferenceModeDeclarationEmit3(module=node16).js
```diff
@@ -25,15 +25,54 @@ declare global {
...
 
 //// [index.js]
-/// <reference types="pkg" resolution-mode="require"/>
+/// <reference types="pkg" resolution-mode="require" preserve="true" />
 export {};
 
 
 //// [index.d.ts]
-/// <reference types="pkg" resolution-mode="require"/>
+/// <reference types="pkg" preserve="true" />
```
sheetalkamat, line 37: Dont you need resolution-mode in d.ts and i think thats why its causing it not to include ?

sheetalkamat (changes requested):

tests/baselines/reference/nodeModulesTripleSlashReferenceModeDeclarationEmit3(module=node16).js
jakebailey: Woah, weird. Thanks for noticing. It's written in the original code so there's something else wrong in what I changed. 

image

src/compiler/transformers/declarations.ts
```diff
@@ -216,7 +202,6 @@ import {
     visitNodes,
     VisitResult,
 } from "../_namespaces/ts";
-import * as moduleSpecifiers from "../_namespaces/ts.moduleSpecifiers";
```
jakebailey, line 205: Super excited to drop this dep.

src/compiler/emitter.ts
```diff
@@ -4185,19 +4183,23 @@ export function createPrinter(printerOptions: PrinterOptions = {}, handlers: Pri
                 writeLine();
             }
         }
+        // TODO(jakebailey): clean up preserve duplication
```
jakebailey, line 4186: Need to do something here, but I have not found something I like.

jakebailey (commented):
Marking this as ready to review; there are some TODOs:

- Should we preserve `preserve` in output declarations, or omit them as they have no meaning there?
- Various small TODOs I've left for myself

tests/baselines/reference/nodeModulesTripleSlashReferenceModeDeclarationEmit3(module=node16).js
```diff
@@ -25,15 +25,15 @@ declare global {
     "type": "module"
 }
 //// [index.ts]
-/// <reference types="pkg" resolution-mode="require"/>
+/// <reference types="pkg" resolution-mode="require" preserve="true" />
```
jakebailey, line 28: One of many examples of `preserve` appearing; maybe we should just not write these to dts files? We already selectively omit `resolution-mode` when it's redundant to the implied format of the module, so it feels like dropping `preserve` would be fine.

jakebailey: The answer from the design meeting was yes, keep it.

andrewbranch (approved):

sheetalkamat (commented):
Are you not going to remove exportedModulesFromDeclarationEmit

jakebailey:
Oh, yes, oops, forgot I had to do that in this PR.

jakebailey:
@typescript-bot test top400

jakebailey:
I never ran perf on this, and the above shows that this is a ~2% emit saving.

src/compiler/checker.ts
```diff
@@ -48891,93 +48872,6 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
-            for (const decl of symbol.declarations) {
-                const file = getSourceFileOfNode(decl);
-                if (fileToDirective.has(file.path)) {
-                    return true;
-                }
-            }
-            return false;
-        }
-
         function addReferencedFilesToTypeDirective(file: SourceFile, key: string, mode: ResolutionMode) {
```
sheetalkamat, line 48875: I think you can get rid of `addReferencedFilesToTypeDirective` as well as `fileToDirective` 
Probably `getResolvedTypeReferenceDirectives`  on `TypeCheckerHost`

jakebailey:
The above failures are ones we noted during the analysis that led up to this PR; I think most of these come down to potential oddities in the way we scan the project and run tsconfigs, as some of these files are test files that shouldn't have been loaded outside of their test projects. Not sure about the notebook thing, will look.

src/compiler/checker.ts
jakebailey: Nice, yes! Thanks!

jakebailey: Finding even more dead code now that I look harder!

jakebailey:

jakebailey:
Nice, now nearly 4% faster emit for the self benchmarks.

jakebailey:
@typescript-bot pack this

jakebailey:
Tested jupyterlab out; adding `preserve=true` fixes all problems. All of the other errors seem like oddities with how we attempt to build the project without actually invoking their build system.

sheetalkamat (approved):

PR #57684, opened 2024-03-07
Switch to eslint flat config

jakebailey:
The VS Code extension now supports this as of v3.0.10. The next version of ESLint makes a flat config required, so may as well do it now.


eslint.config.mjs
```diff
@@ -0,0 +1,259 @@
+// @ts-check
+import eslint from "@eslint/js";
+import noNull from "eslint-plugin-no-null";
+import simpleImportSort from "eslint-plugin-simple-import-sort";
```
mrazauskas, line 4: If I get it right, `eslint-plugin-no-null` and `eslint-plugin-simple-import-sort` do not ship typings. Unfortunately `// @ts-check` is another painful aspect of flat config. Type checks are broken. This also means that using  `typescript-eslint` with `ProjectService` makes linting `eslint.config.js` very troublesome.

I was trying to explain these problems in ESLint repo, but got ignore. Unfortunately support for JSON flat config got overlooked. They say it's not possible. Dear TypeScript, thanks for doing the impossible (;

jakebailey: Oh, right, the default inferred project options have `noImplicitAny=false`.

jakebailey: All plugins are now gone.

sheetalkamat (approved):
Dont have very detailed knowledge here but if things are passing i would say get this in and then deal with fallouts if any later.

Issue #56164, opened 2023-10-20
Design Meeting Notes, 10/20/2023


DanielRosenwasser:
# `isolatedModules` and Cross-File Enum Member References

https://github.com/microsoft/TypeScript/issues/56153

* "Who doesn't love enums?"
* If you have a numeric enum member in one file, and a numberic enum that references that first one in another file...
* ...and a subsequent member has no initializer...

  ```ts
  // foo.ts
  enum Foo {
      A = 123,
  }

  // bar.ts
  import { Foo } from "./foo";

  enum Bar {
      B = Foo.A,
      C
  }
  ```

  ```js
  import { Foo } from "./foo";
  export var Bar;
  (function (Bar) {
      Bar[Bar["B"] = 10] = "B";
      Bar[Bar["C"] = 11] = "C";
  })(Bar || (Bar = {}));
  ```
* ...under `isolatedModules` we don't issue an error - we don't know what this is supposed to be.
* We do when `A` is a string - and if it is, we generate a `void 0` for the initializer - but not if it's numeric (as above).
  ```ts
  // foo.ts
  enum Foo {
      A = "hello",
  }

  // bar.ts
  import { Foo } from "./foo";

  enum Bar {
      B = Foo.A,
      C
  }
  ```

  ```js
  import { Foo } from "./foo";
  export var Bar;
  (function (Bar) {
      Bar["B"] = "10";
      Bar[Bar["C"] = void 0] = "C";
  })(Bar || (Bar = {}));
  ```
* Should we?
* Probably?
* A compiler doesn't know what it should perform - numbers get a reverse-map, strings don't.
    * Well, a compiler doesn't, but runtime emit can do a dynamic check.
        * Doing a lot of those is undesirable.
        * You only have to do one, right?
        * No, each uninitialized enum member may have to perform the same check.
            * (Is this true?)
* Some philosophical discussion - should `isolatedModules` imply this dynamic behavior.
    * We won't emit these things, so how are compilers supposed to know what they should do?
    * On the other hand, there is nothing preventing compilers from making these enums "work".
* We have a few choices.
    1. Force people to do the auto-numbering themselves.

       ```ts
       // bar.ts
       import { Foo } from "./foo";

       enum Bar {
           B = Foo.A,
           C = B + 1
       }
       ```
    2. Emit a dynamic check using `typeof`.
[...]

PR #57686, opened 2024-03-08
Ensure enum members syntactically determinable to be strings do not get reverse mappings

andrewbranch:
Unblocks #56736?

In #56164, we decided that under `isolatedModules`, (from the notes) “enum members referencing another external enum must always be numeric, or explicitly converted to a string. Otherwise, it is an error.” This proposed rule has since undergone a number of interpretations on its way to an attempt to implement it. In https://github.com/microsoft/TypeScript/issues/56153#issuecomment-1773331706, Ryan stated that we should issue an error when “an enum initializer expression that isn't a string literal has a string type.” In #56736, @frigus02 worked off this statement, but interpreted “string literal” as any expression that could be evaluated to a known constant string value with single-file syntactic information only.

An invariant guiding the decision is that error-free (under `--isolatedModules`) code should produce semantically equivalent JS between `ts.transpileModule` and `tsc`. Currently, that’s violated by this test case:

```ts
// @filename: ./foo.ts
import { BAR } from './bar';
enum Foo { A = `${BAR}` }

// @filename: ./bar.ts
export const BAR = 'bar';
```

This is currently error-free, but the emit varies. `tsc`:

```ts
import { BAR } from './bar';
var Foo;
(function (Foo) {
    Foo["A"] = "bar";
})(Foo || (Foo = {}));
```

`ts.transpileModule`:

```ts
import { BAR } from './bar';
var Foo;
(function (Foo) {
    Foo[Foo["A"] = `${BAR}`] = "A";
})(Foo || (Foo = {}));
```

The problematic part isn’t the inlining of the value `"bar"`; it’s the fact that the transpiled output includes a reverse mapping when it absolutely shouldn’t.

To fix this, we either need an error under `isolatedModules` or we need to fix the `ts.transpileModule` emit. Both Ryan’s formulation of the rule proposed in #56164 and @frigus02’s expanded interpretation in #56736 make this code an error. This PR fixes the emit by determining that the initializer is definitely going to be a string without looking at any type or extra-file info.

I’m fine with adopting Ryan’s very strict rule instead of this, but it does seem like fairly low hanging fruit here. I’m less of a fan of the approach currently in #56736, since constant evaluation _feels_ like the wrong thing to control whether a reverse mapping gets created, even though that’s what actually controls the emit today.


frigus02 (commented):
Thank you, Andrew. I'm personally in favor of this approach.

src/testRunner/unittests/services/transpile.ts
```diff
@@ -673,4 +673,13 @@ export * as alias from './file';`,
             options: { compilerOptions: { module: ts.ModuleKind.ESNext, target: ts.ScriptTarget.ESNext } },
         },
     );
+
+    transpilesCorrectly(
+        "Syntactically string but non-evaulatable enum members do not get reverse mapping",
```
frigus02, line 678:
```suggestion
        "Syntactically string but non-evaluatable enum members do not get reverse mapping",
```

jakebailey:
> since constant evaluation feels like the wrong thing to control whether a reverse mapping gets created

I personally agree with this, and so _think_ I prefer this PR.

> even though that’s what actually controls the emit today.

Is there some way we can instead use `isSyntacticallyString` to do that, and we stop using constant eval entirely for the other determinations?

andrewbranch:
This PR moves in that direction, but we can't fully control emit by `isSyntacticallyString` because that would break the case

```ts
import { SomeOtherStringEnum } from './m';

enum Foo {
  A = SomeOtherStringEnum.StringValue
}
```

This works in non-isolatedModules `tsc`. #56736 will make it an error under `isolatedModules`. Basically, you're asking for the `isolatedModules` rules to apply to everyone, which might be the design choice we would make starting fresh, but would be a big breaking change now.

andrewbranch:
@typescript-bot pack this

jakebailey:
FWIW esbuild does it the same way this PR does so I am all for it: https://jakebailey.dev/esbuild-playground/#v3=PTAEAEDMEsBsFMB2BDAtvAXKAxge0TAOYB0AVgM74BQA3laKAERIAuATgJ4AKu0iL5RlgDajYHwAm8AB7EB0xgF0ANPSaRcbVMhZCm8cqkZUAvlSogIMBCnRZJMueWlVoqAA6aWoGqABCAIIASqAmoJBsuKigAOQARshsMQDcVEgArtEAYri4PqABoAC8oAAGACQ0gUEmpaFUQA

andrewbranch:
Jake and Ryan have expressed support for this. Can I have a ✅? 😄

src/compiler/utilities.ts
```diff
@@ -10639,3 +10639,23 @@ export function replaceFirstStar(s: string, replacement: string): string {
...
+export function isSyntacticallyString(expr: Expression): boolean {
+    switch (expr.kind) {
+        case SyntaxKind.BinaryExpression:
+            const left = (expr as BinaryExpression).left;
+            const right = (expr as BinaryExpression).right;
+            return (
+                (expr as BinaryExpression).operatorToken.kind === SyntaxKind.PlusToken &&
+                (isSyntacticallyString(left) || isSyntacticallyString(right))
+            );
+        case SyntaxKind.ParenthesizedExpression:
```
jakebailey, line 10653: Technically I think this might need the skip parents function for this as I think these also need to handle type casts in JS via JSDoc?

jakebailey (approved):

PR #57689, opened 2024-03-08
Disallow string as `regexp` argument to `String#{match,search}`

### LINKED ISSUE #56109: `String#match` and `String#search` types are too lenient (should not accept string as `regexp` argument)
### ⚙ Compilation target

ES5+

### ⚙ Library

ES5/any

Current definitions, which allow `string | RegExp` as the parameter named `regexp`:

https://github.com/microsoft/TypeScript/blob/e38944d8b8b5f6039e18b36574720ce337a5be1b/src/lib/es5.d.ts#L434-L438

https://github.com/microsoft/TypeScript/blob/e38944d8b8b5f6039e18b36574720ce337a5be1b/src/lib/es5.d.ts#L454-L458

### Sample Code

The following results would be unexpected if `string` was truly supported:

```ts
'$abc'.search('$') // 4
'$abc'.match('$') // ['', index: 4, input: '$abc', groups: undefined]
'abc['.search('[') // Uncaught SyntaxError: Invalid regular expression: /[/: Unterminated character class
'abc['.match('[') // Uncaught SyntaxError: Invalid regular expression: /[/: Unterminated character class
```

Strings are only supported insofar as a regex can be constructed from them (i.e. they're only marginally supported because JS is weakly typed). So `'$abc'.search('$')` is equivalent to `'$abc'.search(/$/)`, not `'$abc'.indexOf('$')` or `'$abc'.search(/\$/)`.

### Documentation Link

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/search

> [regexp](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/search#regexp)
A regular expression object, or any object that has a [Symbol.search](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/search) method.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match

> [regexp](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match#regexp)
A regular expression object, or any object that has a [Symbol.match](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/match) method.

https://tc39.es/ecma262/multipage/text-processing.html#sec-string.prototype.search

> 4. Let rx be ? [RegExpCreate](https://tc39.es/ecma262/multipage/text-processing.html#sec-regexpcreate)(regexp, undefined).

https://tc39.es/ecma262/multipage/text-processing.html#sec-string.prototype.match

> 4. Let rx be ? [RegExpCreate](https://tc39.es/ecma262/multipage/text-processing.html#sec-regexpcreate)(regexp, undefined).

lionel-rowe:
> I think we can take a PR and assess what the breaks look like
> _Originally posted by @RyanCavanaugh in https://github.com/microsoft/TypeScript/issues/56109#issuecomment-1764791186_

Fixes #56109

### DIFF
--- a/src/lib/es2020.string.d.ts
+++ b/src/lib/es2020.string.d.ts
@@ -4,7 +4,7 @@ interface String {
     /**
      * Matches a string with a regular expression, and returns an iterable of matches
      * containing the results of that search.
-     * @param regexp A variable name or string literal containing the regular expression pattern and flags.
+     * @param regexp The regular expression to match against.
      */
     matchAll(regexp: RegExp): IterableIterator<RegExpExecArray>;
 
--- a/src/lib/es5.d.ts
+++ b/src/lib/es5.d.ts
@@ -433,9 +433,9 @@ interface String {
 
     /**
      * Matches a string with a regular expression, and returns an array containing the results of that search.
-     * @param regexp A variable name or string literal containing the regular expression pattern and flags.
+     * @param regexp The regular expression to match against.
      */
-    match(regexp: string | RegExp): RegExpMatchArray | null;
+    match(regexp: RegExp): RegExpMatchArray | null;
 
     /**
      * Replaces text in a string, using a regular expression or search string.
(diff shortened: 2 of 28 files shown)


jakebailey:
@typescript-bot test top400
@typescript-bot run dt

RyanCavanaugh:
This is shady as heck https://github.com/jupyterlab/jupyterlab/blob/864754e0b4274df82ca93296fb9bf7f80ff9f710/packages/services/src/serverconnection.ts#L344

Nearly everything else looks like one of:
 * *Intentional* coercion, e.g. a regex that originates in a JSON store
 * A literal string that isn't unambiguous (`.match("foo")`)
 * A literal string that is clearly intentionally a regex

Overall this is far too many breaks in code that isn't problematic

PR #57696, opened 2024-03-08
Update 'pack this' result script to update new status comment

jakebailey:
I didn't do this originally because I wanted to change the way this build system works, but I probably won't have time for a bit. The fact that this job is the only one that doesn't update the status comment is more annoying in the meantime, so, fix that.

### DIFF
--- a/scripts/post-vsts-artifact-comment.mjs
+++ b/scripts/post-vsts-artifact-comment.mjs
@@ -5,17 +5,29 @@ import assert from "assert";
 import ado from "azure-devops-node-api";
 import fetch from "node-fetch";
 
+/**
+ * @param {string} name
+ */
+function mustGetEnv(name) {
+    const value = process.env[name];
+    assert(value, `No ${name} specified`);
+    return value;
+}
+
+const REQUESTING_USER = mustGetEnv("REQUESTING_USER");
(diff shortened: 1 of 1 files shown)


scripts/post-vsts-artifact-comment.mjs
```diff
@@ -44,26 +49,65 @@ async function main() {
...
-            owner: "Microsoft",
-            repo: "TypeScript",
-            body: `Hey @${process.env.REQUESTING_USER}, something went wrong when looking for the build artifact. ([You can check the log here](https://typescript.visualstudio.com/TypeScript/_build/index?buildId=${process.env.BUILD_BUILDID}&_a=summary)).`,
-        });
-    }
+    newComment = `Hey @${REQUESTING_USER}, something went wrong when looking for the build artifact. ([You can check the log here](https://typescript.visualstudio.com/TypeScript/_build/index?buildId=${BUILD_BUILDID}&_a=summary)).`;
+    emoji = "❌";
+}
+
+const resultsComment = await gh.issues.createComment({
```
jakebailey, line 74: This is more or less the same as https://github.com/microsoft/DefinitelyTyped-tools/blob/main/packages/dtslint-runner/src/post-results.ts#L67.

weswigham (approved):

PR #57704, opened 2024-03-08
Allow `import =` in module augmentations

### LINKED ISSUE #13175: Import aliases are not permitted in a global augmentation
Found from http://stackoverflow.com/questions/41336330/typescript-how-can-i-make-an-existing-namespace-global

```ts
export {};

declare global {
    namespace foo {
        export var a;
    }
    import bar = foo;
}
```

I believe this should yield no error, however, it gives back the error

```
Imports are not permitted in module augmentations. Consider moving them to the enclosing external module.
```

RyanCavanaugh:
Fixes #13175

AFAIK no one knows why this was ever illegal... so just delete the code I guess?


jakebailey (approved):
Do these tests show that the feature works? It seems like they're not doing anything at all, but then again the diff doesn't show the types/symbol baselines.

RyanCavanaugh:
Fixing this to still disallow `import/require` then will add a couple more tests

jakebailey:
I was too excited and couldn't help myself

jakebailey (approved):

Issue #57706, opened 2024-03-09
Unexpected type infer for function first argument with default value since 5.1


ShenQingchuan:
### 🔎 Search Terms

"type infer first argument"

### 🕗 Version & Regression Information

- [ ] This is a crash
- [x] This changed between versions **v5.0.4** and **v5.1.6**
- [ ] This changed in commit or PR _______
- [ ] This is the behavior in every version I tried, and I reviewed the FAQ for entries about _________
- [ ] I was unable to test this on prior versions because _______

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.5.0-dev.20240308#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXy1URBgEYAeAFQD4AKRVALngDEpMcYBPK6gSmaUAUKEiwEKdNjwEiJAEy96TVuwycuimgPjChIAB4AHThngYuRhGw7de8ALzxasAOalBAGnhv5zKKhcfI7UPoH6xqbmltZqGlqhTrQAdKluAM7MANqU3gFcALrBDqH5QuVgeOlmiKSOzgC26a7M1TCErvUA5AAWIBAQOF3eGCDV-oHFoQDe8AC+QpWo1T51ToTEZPSkfEIA9HvwR-AAegD8i1VmUPL1GwrbuwfHpxeXyzW3SU0t8G0d3T6AyGU3gswWSxWACM7nItoh5E9Dsdzu9oWtZJt5PREftkUdUWizGBYZtSLRaD9Whh2qhOk5ev1Bl1QeCkS9UZDiV9MQ9Kc1qbT6fBGcCWSEwfN2Si3kA

### 💻 Code

```ts
export type Factory<T> = (arg1: T, arg2: any) => any
export type Factory2<T> = (...args: [T, any]) => any

declare function infer1<T>(fn: Factory<T>): T
declare function infer2<T>(fn: Factory2<T>): T

const f1 = (msg: string = 'hello', test: any) => { }
const a1 = infer1(f1)
//    ^? const a1: string | undefined
const a2 = infer2(f1)
//    ^? const a2: string | undefined

const f2 = (msg: string = 'hello') => { }
const b = infer1(f2)
//    ^? const b: string | undefined
const b1 = infer2(f2)
//    ^? const b1: string | undefined

const c = infer1((msg: string = 'hello') => { })
//    ^? const c: string
const c2 = infer2((msg: string = 'hello') => { })
//    ^? const c2: string | undefined
```

[...]

PR #57708, opened 2024-03-09
Consider annotated parameter with an initializer as optional when inferring from it

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/57706

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -36542,7 +36542,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             const declaration = signature.parameters[i].valueDeclaration as ParameterDeclaration;
             const typeNode = getEffectiveTypeAnnotationNode(declaration);
             if (typeNode) {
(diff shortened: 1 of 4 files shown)


src/compiler/checker.ts
```diff
@@ -36542,7 +36542,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             const declaration = signature.parameters[i].valueDeclaration as ParameterDeclaration;
             const typeNode = getEffectiveTypeAnnotationNode(declaration);
             if (typeNode) {
-                const source = addOptionality(getTypeFromTypeNode(typeNode), /*isProperty*/ false, isOptionalDeclaration(declaration));
+                const source = addOptionality(getTypeFromTypeNode(typeNode), /*isProperty*/ false, hasInitializer(declaration) || isOptionalDeclaration(declaration));
```
Andarist, line 36545: this matches the logic used by `getTypeOfParameter`

jakebailey:
@typescript-bot test it

PR #57717, opened 2024-03-11
fix(57635): duplicate property name error when trying to overwrite early-bound prop with late-bound prop

### LINKED ISSUE #57635: Duplicate property name error when accessing an expando function in diffrent ways.
### 🔎 Search Terms

expando function duplicate

### ⏯ Playground Link

[Playground Link](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBAMwK4DsDGMCWEWIhACgEoAuOANwkwBM4BvAXwFgAoNHAZ3gCMBDKOAF44Acj5QRrBPgB04oXACMAJgDcU-AG1xAXQUr1LVkA)

### 💻 Code

```ts
export function foo(): void {}
const bar = 'bar'
foo.bar = 12;
foo[bar] = 12;
```

### 🙁 Actual behavior

The property access that uses the constant `bar` causes the error:  `Property 'bar' was also declared here.`

### 🙂 Expected behavior

Code should work without errors.

Seems like the late symbol does not get merged with early one.

idango10:
Fixes #57635

---

As written in the issue, I noticed there's a conflict in which the late-bound member already exists as an early-bound member.

My naive approach was to remove `earlySmybol` from the `if` statement and see which tests are failing, and I saw that `dynamicNamesErrors.ts` was the only test to fail (https://github.com/microsoft/TypeScript/blob/main/tests/cases/compiler/dynamicNamesErrors.ts#L6-L12).

I thought there's a reason for that, and that `Function` should behave differently from `Interface`, but they appear to behave similarly in that case.

In the example below, you can see that when one prop is early-bound and the second is late-bound, there's an error pointing to duplication. However, when they are both early-bound or late-bound, there are no errors ([TS Playground](https://www.typescriptlang.org/play?#code/MYewdgzgLgBMAMMC8MBEBGVBuAsAKFElmHWRgHJ1zc98BLMKAUwCcAzAQ2CZgBVEA3vhgiYAbQQBdAFwwwAVwC2AI1ZYYAeg0wAoixYgWsgCLyADgBs6wDsxhmDZ1lACeFKgDoAFACYA7OgAHACUwqLosgoqapraegZGMAAKjs5ulOQwAO4cEDAcFhAgMAAmTMAWHCxMJTAAFqxM3v4AzC2heAC++PhaMAByxawJEPSMrJzcfKRCeKLiGTJySqosNPMRy9Fr+N20eH2DMMOGo3gMzOxcPLw+MLPzEvBLUavrohLoLytqu-hAA)):

```ts
const c0 = "1";
const c1 = '1';

interface T0 {
    [c0]: number; // Error: Duplicate property '1'.(2718)
    1: number; // Error: Property '1' was also declared here.(2733)
}

// No errors
interface T1 {
    ['1']: number;
    1: number;
}

// No errors
interface T2 {
    [c0]: number;
    [c1]: number;
}
```

I think that regardless of whether we want to forbid or allow duplications in functions and interfaces, it should be consistent, and not inside `lateBindMember`.


tests/baselines/reference/dynamicNamesErrors.errors.txt
```diff
@@ -9,17 +7,13 @@ dynamicNamesErrors.ts(25,1): error TS2322: Type 'T1' is not assignable to type '
...
     const c0 = "1";
     const c1 = 1;
     
     interface T0 {
         [c0]: number;
-        ~~~~
-!!! error TS2718: Duplicate property '1'.
         1: number;
-        ~
-!!! error TS2733: Property '1' was also declared here.
```
jakebailey, line 22: I was going to say that this sure felt wrong, but your playground link does show that doing effectively this in other forms does not error. (Makes me wonder if we're missing errors; but that's more of a design question I don't have context for.)

jakebailey:
@typescript-bot test top400
@typescript-bot run dt

@typescript-bot pack this

src/compiler/checker.ts
```diff
@@ -12997,7 +12997,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 // conflicting flags.
```
sandersn, line 12997: please update this comment to read "Report an error if there's a symbol declaration with the same name and conflicting flags.", since there's no longer an error for any earlySymbol with the same name.

sandersn (changes requested):
Looks correct, although the preceding comment needs simplification.

tests/baselines/reference/dynamicNamesErrors.errors.txt
sandersn: I don't think we should allow it, but it's definitely a separate question -- and disallowing it will probably break some amount of code.

src/compiler/checker.ts
idango10: Updated 🙂

sandersn (approved):

Issue #57680, opened 2024-03-07
package.json#imports doesn't get IDE suggestions when typing the import path directly


anthonyshew:
### 🔎 Search Terms

package.json imports, package.json imports intellisense, package.json imports autocomplete, package.json import suggestions

### 🕗 Version & Regression Information

This feature released with v5.4 yesterday so not too much bisection to be done. 😄

### ⏯ Playground Link

https://github.com/anthonyshew/packagejson-imports-typescript-suggestions

### 💻 Code

I apologize for not following the standards of the template to provide a TypeScript playground. I couldn't find a way to work with imports or a package.json there. If that is possible and I missed it, please let me know so I can update appropriately.

Reproduction repository: https://github.com/anthonyshew/packagejson-imports-typescript-suggestions

package.json#import: https://github.com/anthonyshew/packagejson-imports-typescript-suggestions/blob/b33225588c9ca061807f8c1cb55c50dff66ef581/package.json#L6

Attempted inference: https://github.com/anthonyshew/packagejson-imports-typescript-suggestions/blob/b33225588c9ca061807f8c1cb55c50dff66ef581/app/page.tsx#L1

### 🙁 Actual behavior

With your cursor on the path string for the import on line 1 of `./app/page.tsx`, opening suggestions does not provide autocomplete.
![CleanShot 2024-03-07 at 11 39 38@2x](https://github.com/microsoft/TypeScript/assets/35677084/0007dd4e-8d37-41c6-8040-876356b76d0f)

### 🙂 Expected behavior

I was hoping to see a suggestion for `#components/TestComponent.tsx`. Notably, if I write it out fully myself, my editor is happy and my build compiles/bundles successfully so I'm relatively certain that I am set up correctly.

### Additional information about the issue

I did some light futzing with the `"allowImportingTsExtensions"` and `"verbatimModuleSyntax"` compiler options in my `tsconfig.json` and was noticing some un-aligned behavior. I'm not sure if this is deserving of another bug report or is related.

As an example, auto-importing seems to use `verbatimModuleSyntax` (or something like it) even when I don't have that flag enabled. Again, not certain if related.
![CleanShot 2024-03-07 at 11 43 15@2x](https://github.com/microsoft/TypeScript/assets/35677084/3a80b987-6cf8-4f00-933c-de8544a8a47a)

Andarist:
[...]

Issue #52460, opened 2023-01-27
Path completions for Node.js subpath imports


thetutlage:
# Suggestion

Node.js for a while has a way to define [import aliases (known as subpath imports)](url) in `package.json` file. The subpath imports also work fine when I compile my code using `tsc` and inside my code editor (VSCode).

However, as I type the import path in my code editor, no suggestions related to subpath imports are shown.

Given, I have the following subpath import alias inside `package.json`.

```json
{
  "imports": {
    "#controllers/*": "./app/controllers/*.js"
  }
}
```

I should be able to type `#controllers` and get suggestions. Infact, using an export symbol from one of the controller files should write the auto-import using the alias.

### `tsconfig.json`

Following is a trimmed down version of my `tsconfig.json` file.

```json
{
  "compilerOptions": {
    "target": "ESNext",
    "module": "NodeNext",
    "lib": ["ESNext"],
    "rootDir": "./",
    "outDir": "./build",
    "esModuleInterop": true,
    "skipLibCheck": true,
    "types": ["@types/node"]
  },
  "include": ["./**/*"],
  "exclude": ["./node_modules", "./build"]
}
```

## 🔍 Search Terms

- Node subpath imports
- Auto complete subpath imports
- Redirected by https://github.com/microsoft/vscode/issues/128064

## ✅ Viability Checklist

* [x] This wouldn't be a breaking change in existing TypeScript/JavaScript code
* [x] This wouldn't change the runtime behavior of existing JavaScript code
* [x] This could be implemented without emitting different JS based on the types of the expressions
* [x] This isn't a runtime feature (e.g. library functionality, non-ECMAScript syntax with JavaScript output, new syntax sugar for JS, etc.)
* [x] This feature would agree with the rest of [TypeScript's Design Goals](https://github.com/Microsoft/TypeScript/wiki/TypeScript-Design-Goals).

## ⭐ Suggestion
Since, TypeScript can understand and compile subpath imports, it will be great, if they are also suggested during import autocomplete

## 📃 Motivating Example

Have nice experience of auto-importing files with aliases.

## 💻 Use Cases

Right now, I have to either manually type the complete import statement by hand. Or type the export symbol name and update the import path to use the subpath alias.

Attaching a small screen recording of the same

https://user-images.githubusercontent.com/1706381/215099932-cf274151-6882-4a14-ba49-3b127a9231e3.mov

xobotyi:
Same here, except the IDE used.

[...]

PR #20763, opened 2017-12-18
Update harness to use single robust virtual file system for tests.

rbuckton:
To better support the lib reference PR (#15780) and the ES3 lib PR (#16077) we needed to make a number of changes to our test harness to improve and consolidate our various approaches to emulating a file system. To that end, this PR introduces a much more robust virtual file system implementation that supports additional features needed by our various test runners including:

* Symbolic links
* ~File/Directory watching~
* File stats
* Resolving paths/files to the real file system
* File/Directory metadata
* VFS shadowing (allows the use of a shared VFS where changes in a shadowed VFS do not affect the real VFS)

As of this PR, the Compiler/Conformance, Project, and FourSlash runners all now use the VFS. ~In addition all existing unit tests that used "virtualFileSystemWithWatch" have been migrated over to the single VFS implementation.~

PR #57718, opened 2024-03-11
Add string literal completions for `package.json` `imports` field

Andarist:
closes https://github.com/microsoft/TypeScript/issues/52460
closes https://github.com/microsoft/TypeScript/issues/57680
closes https://github.com/microsoft/TypeScript/issues/57777

Currently, this only has tests based on https://github.com/microsoft/TypeScript/pull/55015 but I still have to add more


src/compiler/utilities.ts
```diff
@@ -6288,6 +6290,21 @@ export function getPossibleOriginalInputExtensionForExtension(path: string) {
         [Extension.Tsx, Extension.Ts, Extension.Jsx, Extension.Js];
 }
 
+/** @internal */
+export function getPossibleOriginalInputPathWithoutChangingExt(
```
Andarist, line 6294: this is kinda a reverse of `getOutputPathWithoutChangingExt`: https://github.com/microsoft/TypeScript/blob/3fca8c842a42bb43ba5a86dcce1b7fb0eae6c7a4/src/compiler/emitter.ts#L542-L554

src/compiler/utilities.ts
```diff
@@ -6288,6 +6290,21 @@ export function getPossibleOriginalInputExtensionForExtension(path: string) {
         [Extension.Tsx, Extension.Ts, Extension.Jsx, Extension.Js];
```
Andarist, line 6290: tangent: would it make any sense to limit this to avoid returning js/jsx for `path` with ts/tsx?

src/services/stringCompletions.ts
```diff
@@ -1047,7 +1125,10 @@ function getModulesForPathsPattern(
     pattern: string,
     extensionOptions: ExtensionOptions,
     isExportsWildcard: boolean,
+    isImports: boolean,
+    compilerOptions: CompilerOptions,
     host: LanguageServiceHost,
+    moduleSpecifierResolutionHost: ModuleSpecifierResolutionHost | undefined
```
Andarist, line 1131: TODO: it was annoying to pass `moduleSpecifierResolutionHost` here from all of the callers but I probably should do that

src/services/stringCompletions.ts
```diff
@@ -1106,11 +1189,29 @@ function getModulesForPathsPattern(
...
+                    return directoryResult(getPathComponents(removeLeadingDirectorySeparator(trimmedWithPattern))[1]);
+                }
+                const { name, extension } = getFilenameWithExtensionOption(trimmedWithPattern, compilerOptions, extensionOptions, isExportsWildcard);
+                return nameAndKind(name, ScriptElementKind.scriptElement, extension);
+            }
+        });
+        matches = concatenate(matches, inputMatches);
+    }
+
+    matches = concatenate(matches, directories);
```
Andarist, line 1208: TODO: this lacks `inputDirectories`

src/services/stringCompletions.ts
```diff
@@ -1106,11 +1189,29 @@ function getModulesForPathsPattern(
     const directories = normalizedSuffix
         ? emptyArray
         : mapDefined(tryGetDirectories(host, baseDirectory), dir => dir === "node_modules" ? undefined : directoryResult(dir));
-    return [...matches, ...directories];
 
-    function trimPrefixAndSuffix(path: string): string | undefined {
+    if (inputBaseDirectory && inputBaseDirectory !== baseDirectory) {
```
Andarist, line 1193: TODO: a helper has to be introduced to deduplicate this branch with the ones that creates original `matches`

src/services/stringCompletions.ts
```diff
@@ -831,17 +840,20 @@ function addCompletionEntriesFromPaths(
...
     };
-    return addCompletionEntriesFromPathsOrExports(result, /*isExports*/ false, fragment, baseDirectory, extensionOptions, host, getOwnKeys(paths), getPatternsForKey, comparePaths);
+    return addCompletionEntriesFromPathsOrExports(result, /*isExports*/ false, /*isImports*/ false, fragment, baseDirectory, extensionOptions, compilerOptions, host, moduleSpecifierResolutionHost, getOwnKeys(paths), getPatternsForKey, comparePaths);
 }
 
 /** @returns whether `fragment` was a match for any `paths` (which should indicate whether any other path completions should be offered) */
 function addCompletionEntriesFromPathsOrExports(
     result: NameAndKindSet,
     isExports: boolean,
+    isImports: boolean,
```
Andarist, line 850: it's somewhat annoying to have those separate parameters but I need to distinguish between those 2 cases. Alternatively, I could introduce an enum

src/services/stringCompletions.ts
```diff
@@ -875,7 +887,7 @@ function addCompletionEntriesFromPathsOrExports(
             if (typeof pathPattern === "string" || matchedPath === undefined || comparePaths(key, matchedPath) !== Comparison.GreaterThan) {
                 pathResults.push({
                     matchedPattern: isMatch,
-                    results: getCompletionsForPathMapping(keyWithoutLeadingDotSlash, patterns, fragment, baseDirectory, extensionOptions, isExports && isMatch, host)
+                    results: getCompletionsForPathMapping(keyWithoutLeadingDotSlash, patterns, fragment, baseDirectory, extensionOptions, (isExports || isImports) && endsWith(keyWithoutLeadingDotSlash, '*'), isImports, compilerOptions, host, moduleSpecifierResolutionHost)
```
Andarist, line 890: this `isExportsWildcard` parameter here is a little bit surprising. It feels like it could be checked further down the callee chain - but I'm not sure yet

src/services/stringCompletions.ts
```diff
@@ -938,10 +951,41 @@ function getCompletionEntriesForNonRelativeModules(
...
+            let seenPackageScope = false;
             let ancestorLookup: (directory: string) => void | undefined = ancestor => {
                 const nodeModules = combinePaths(ancestor, "node_modules");
                 if (tryDirectoryExists(host, nodeModules)) {
-                    getCompletionEntriesForDirectoryFragment(fragment, nodeModules, extensionOptions, host, /*moduleSpecifierIsRelative*/ false, /*exclude*/ undefined, result);
+                    getCompletionEntriesForDirectoryFragment(fragment, nodeModules, extensionOptions, compilerOptions, host, /*moduleSpecifierIsRelative*/ false, /*exclude*/ undefined, result);
+                }
+                if (!seenPackageScope) {
+                    const packageFile = combinePaths(ancestor, "package.json");
+                    if (seenPackageScope = tryFileExists(host, packageFile)) {
```
Andarist, line 962: TODO: this has to be heavily deduplicated with the existing branch that handles `exports`

src/services/stringCompletions.ts
```diff
@@ -938,10 +951,41 @@ function getCompletionEntriesForNonRelativeModules(
             }
         }
         if (!foundGlobal) {
+            let seenPackageScope = false;
```
Andarist, line 954: TODO: add a test case that will show that it's needed

src/services/stringCompletions.ts
```diff
@@ -1021,20 +1093,26 @@ function getCompletionsForPathMapping(
     packageDirectory: string,
     extensionOptions: ExtensionOptions,
     isExportsWildcard: boolean,
+    isImports: boolean,
+    compilerOptions: CompilerOptions,
     host: LanguageServiceHost,
+    moduleSpecifierResolutionHost: ModuleSpecifierResolutionHost | undefined
 ): readonly NameAndKind[] {
-    if (!endsWith(path, "*")) {
```
Andarist, line 1026: This should also fix completions when `*` is somewhere in the middle of `paths`. The algorithm is largely the same between paths/exports/imports after all.

Note that nothing here fixes completions with multiple wildcards in the value - I don't want to work on that.

src/services/stringCompletions.ts
```diff
@@ -1072,9 +1153,11 @@ function getModulesForPathsPattern(
 
     const normalizedSuffix = normalizePath(parsed.suffix);
     const declarationExtension = normalizedSuffix && getDeclarationEmitExtensionForPath("_" + normalizedSuffix);
-    const matchingSuffixes = declarationExtension ? [changeExtension(normalizedSuffix, declarationExtension), normalizedSuffix] : [normalizedSuffix];
+    const inputExtension = isImports && normalizedSuffix ? getPossibleOriginalInputExtensionForExtension("_" + normalizedSuffix) : undefined;
```
Andarist, line 1156: All of the logic in this function assumes the same extensions between potential inputs and outputs and stuff. This can impact the used globs but I think it's largely fine right now - unless somebody would come up with a compelling test case that should work differently.

tests/cases/fourslash/server/importCompletions_importsMap3.ts
```diff
@@ -0,0 +1,36 @@
...
+//// import {} from "#internal//*2*/";
+
+verify.completions({
+    marker: ["1"],
+    exact: ["#internal"],
+    isNewIdentifierLocation: true,
+});
+verify.completions({
+    marker: ["2"],
+    exact: ["foo.js"],
```
Andarist, line 34: This currently fails. I took this test case from https://github.com/microsoft/TypeScript/pull/55015/files#diff-c8e8521e91d6946ce7869359a143f8cf7b65ee15703a4e04c7d73dcbaa5682a4 but I think that perhaps it should just be removed/tweaked in both places. A trailing slash is already not supported by node@>=17 and it was deprecated before that. There is little value in supporting this - in my opinion, it would be better to discourage this completely.

In addition to that, I noticed that the previous segment gets suggested in a situation like this today:
```json
{
    "name": "pkg_test",
    "exports": {
        "./test/": "./"
    }
}
```

```ts
import { bazinga } from 'pkg_test/test//*1*/'
```

At marker `1` I get `test` again, it's imho a clear bug and I can fix it here - after we agree on not supporting this old deprecated pattern at all.

Andarist:
@andrewbranch I opened this as a draft because there are some minor cleanups to be done here. I'd appreciate an early review here though - in case there is something fundamentally wrong with this.

tests/cases/fourslash/server/importCompletions_importsMap5.ts
Andarist, line 1: I need more test cases like this but with a pattern. Atm, I didn't add any special logic for `declarationDir` - this "exact" match works solely based on the fact that it's exact. It doesn't even need to check if the corresponding file exists on the disk or not so it doesn't perform any output->input mapping etc. This has to be fixed.

DanielRosenwasser:
@Andarist did you still want to pursue this? We're looking to get this done in the TS 5.7 timeframe.

Andarist:
@DanielRosenwasser yes, it would be great if @andrewbranch could give this a quick look to check if im not doing anything overly wrong so I dont spend too much time on cleaning up things that will turn out to be wrong in the end ;p

andrewbranch (commented):
Yeah, this looks on the right track to me, thanks!

Andarist:
@andrewbranch thanks for the review! I’ll try to clean this up asap

Andarist:
I have been busy the last 2 weeks, I've done some progress on this in the past days but I hope to get everything done till the end of this week 🤞

src/harness/vfsUtil.ts
```diff
@@ -1148,7 +1148,7 @@ export class FileSystem {
         for (const key of Object.keys(files)) {
             const value = normalizeFileSetEntry(files[key]);
             const path = dirname ? vpath.resolve(dirname, key) : key;
-            vpath.validate(path, vpath.ValidationFlags.Absolute);
+            vpath.validate(path, vpath.ValidationFlags.Absolute | vpath.ValidationFlags.AllowWildcard);
```
Andarist, line 1151: I have no idea if this is completely correct. I added it to make [pathCompletionsPackageJsonImportsSrcNoDistWildcard6](https://github.com/microsoft/TypeScript/blob/a80cdfaf5aee37250ae5cfae3d374f8773bd1dd4/tests/cases/fourslash/server/pathCompletionsPackageJsonImportsSrcNoDistWildcard6.ts) pass as it's using `?` in the component path component. 

I have added that test based on the existing [pathCompletionsPackageJsonExportsWildcard6](https://github.com/microsoft/TypeScript/blob/b8e4ed8aeb0b228f544c5736908c31f136a9f7e3/tests/cases/fourslash/pathCompletionsPackageJsonExportsWildcard6.ts). The difference is that the test I have added is a `fourslash/server` test and it follows a slightly different codepath.

The existing one validates here:
https://github.com/microsoft/TypeScript/blob/b8e4ed8aeb0b228f544c5736908c31f136a9f7e3/src/harness/vfsUtil.ts#L1104-L1108

And the `ValidationFlags.AllowWildcard` was added there [here](https://github.com/microsoft/TypeScript/pull/24076/files#diff-effd49d37f377e1f194f7ceaddf609b1d5eb60744ad741cf4e43e11e6878f79dR872-R873)

On the other hand, the `fourslash/server` validates here, where I'm adding this comment. This codepath was added as part of https://github.com/microsoft/TypeScript/pull/20763 - which proceeds the PR that introduced `ValidationFlags.AllowWildcard`. So perhaps this was just a harmless omission in that newer PR. I don't know why this is a flag in the first place though so 🤷‍♂️

Andarist:
@andrewbranch it's ready for re-review :)

I could also use a build of this to test it more easily in a real project, and not only in the test harness ;p cc @jakebailey

src/harness/vfsUtil.ts
andrewbranch: @rbuckton can you advise?

jakebailey:
@typescript-bot pack this

jakebailey (approved):
I think this is okay; the FS thing I think is also fine? Everything passes...

andrewbranch (approved):

PR #57724, opened 2024-03-11
Fix nullability intersections in CFA and relations

### LINKED ISSUE #57693: `null` gets accidentally eliminated when narrowing by undefined's equality
### 🔎 Search Terms

null undefined narrow narrowing reduction equality narrowby

- This is the behavior in every version I tried* 

*it worked differently in 5.1-5.3 because those had a bug in them ( https://github.com/microsoft/TypeScript/pull/57202 )

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.5.0-dev.20240308#code/C4TwDgpgBAggdiA8gIwFYQMbCgXigJUwHsAnAEwB4BnYEgSzgHMAaKAQwQD4BuAKFEhQAysDbBoeeEjSZgfXmUwAbNiWgAzAK5wsdInCgALNlUQB3OAAUSRSCVAUAKlAgAPcXDJVYCFOiycABS8UFBEMlgAXFCOzCFQYDZg0da2EPYgANIQIHEAlNGJtlB03gDWOUTqMfIM4iTqbBjQIqQQFEK4wqLinFAA3vFUEMAiYu2ZLu4QnuWV1UJBFSDRmawAbmxKmhDRQgDamQC6BVDrRHRkfAC+vLxuYKTYWjrAegZUIDqtaqlgFPFOm4PF5uuM4qFLFMQd5LKo3lsOpwIVBJsCZqDllVhHEgjQ2ntgG0kawimAqClWMtVnkBvEMPoaGctjsusZTBY-ulQIEyVQqTlaQB+BJJKj7ZZHKDRbSKdQMCBXO6hOjVQKbbYSHB4WUQeVwRW0tTATQkOB8UIAektUGAhlKLhINhIVCGRLUADphqMehBAssNiyIHl5FabXaHVRDERNEoyHAAOTYNjqdSyW2GaACaAcEDRsysMwQBPrHNKNRsMggKAYTMYCpkdiMNgMJl26BUNgAWyz9qY8VVUHVQdw2qguv1hqgxtN5vi1qgVE0ZEUcGASmrdGwS4wzUVrtC+M93rG4n9OUDmpDvFuQA

### 💻 Code

```ts
type AnyObject = Record<string, any>;
type State = AnyObject;

declare function hasOwnProperty<T extends AnyObject>(
  object: T,
  prop: PropertyKey,
): prop is keyof T;

interface Store<S = State> {
  setState<K extends keyof S>(key: K, value: S[K]): void;
}

export function syncStoreProp<
  S extends State,
  P extends Partial<S>,
  K extends keyof S,
>(store: Store<S>, props: P, key: K) {
  const value = hasOwnProperty(props, key) ? props[key] : undefined;

  if (value === undefined) return;
  // this errors
  store.setState(key, value);

  // this shouldn't affect the type anyhow, we've already checked against the same thing
  if (value === undefined) return;
  // suddently it succeeds
  store.setState(key, value);
}
```

### 🙁 Actual behavior

one of the calls errorrs while the other one doesn't

### 🙂 Expected behavior

i'd expect consistent behavior at both call sites

originally reported by @diegohaz here: https://twitter.com/diegohaz/status/1765674095293747628

ahejlsberg:
Fixes #57693.


ahejlsberg:
@typescript-bot test top200
@typescript-bot run dt

src/compiler/checker.ts
```diff
@@ -21180,6 +21186,16 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
     }
 
+    function shouldNormalizeIntersection(type: IntersectionType) {
+        let hasInstantiable = false;
+        let hasNullableOrEmpty = false;
+        for (const t of type.types) {
+            hasInstantiable ||= !!(t.flags & TypeFlags.Instantiable);
+            hasNullableOrEmpty ||= !!(t.flags & TypeFlags.Nullable) || isEmptyAnonymousObjectType(t);
+        }
+        return hasInstantiable && hasNullableOrEmpty;
```
RyanCavanaugh, line 21196: Potential to bail early from this loop
```suggestion
            hasNullableOrEmpty ||= !!(t.flags & TypeFlags.Nullable) || isEmptyAnonymousObjectType(t);
            if (hasInstantiable && hasNullableOrEmpty) return true;
        }
        return false;
```

RyanCavanaugh (approved):

Issue #49033, opened 2022-05-09
Go to definition for TypeScript literal strings


KATT:
> **$500 bounty** from my own pocket if it's done before August 2022 _(just putting a timeline so I won't be hunted about this two years from now, feel free to contact me ahead of working on it if you are curious if I'm still up for it!)_

## Feature request

Go to definition for TypeScript literal strings.

### Example

Given the below code, I would really like to be able to jump to the definition of the `'foo'`:

```ts
const obj = {
  foo: 'bar' as const,
  hello: 'world' as const,
}

function get<T, K extends keyof T>(obj: T, path: K): T[K] {
  return obj[path]
}

const val = get(obj, 'foo')
//                   👆👆👆 Is there a way of make jump to definition work here?
//                         I really want to CMD+click `'foo'` and jump to line #2
```

[**Link to TypeScript playground**](https://www.typescriptlang.org/play?ts=4.6.2#code/MYewdgzgLgBCBGArGBeGBvAsAKBjAZiCAFwwDk8AhgE5kyUQyiRQA0OeAFgKYA2vJcgHcQ1XgBM6DJuGjtsAXxw58AVzDAoAS3AwA5tygAeACqsYAaRjcAHlG5hxjANbcAniHwwTAPgAUCIikZjAADpRQnKQWAJTBANoWALoYHDDUhqrUYHBI8eGRSThK2DjM0DAAbpS8qPqGAUjmZIQgZDE4APSdeL19-f2AvBuAYjsjIzAAkoyR3Bn0MEKUbnBeALaUrjCIqquhMFAgMOLc+FpgWtq6ItTOMDwZAPxdPQOvb70T6dw1vMuLYLADjAAMIAWQAIgBqYC8LTAW4AAxaRDICPoji2Oz2QNhYG4MAAxAAmZTYcogXjcAB0Aj0fmqvA62G67xgAD0HkA).

### WebStorm Demo of the same code

In WebStorm, it actually works how I wish that VSCode worked, here's a video:

https://user-images.githubusercontent.com/459267/167470744-80bdd8af-b7fa-444a-84d5-7e3f83e7e003.mov

KATT:
Cc @RyanCavanaugh is there a way to put bug bounties on issues in order to incentivize OSS-contributors to take on tasks ahead of the core team? I'd happily put forward a few hundred $ personally for this to be solved as it's a big pain point in my API design of [tRPC](https://trpc.io/)

[...]

PR #57729, opened 2024-03-11
Use `origin` for string literal definitions

Andarist:
⚠️ this is a proof-of-concept right now. It's not a complete implementation.

`.origin` information often allows the compiler to track back how a union has been created. It's used for improved type displays by quick info and such. It's also used by some optimizations. It occurred to me that it could be used in more places - such as go to definition

Shortcomings of this:
- not every pattern is supported by `.origin`. IIRC, `T[K]` doesn't track this under any circumstances
- `keyof { 'anonymous': 'object' }` is not tracked today. Since `.origin` was first introduced for type-displaying purposes it didn't make sense to track this. It would make sense to track it now as it would be quite crucial to support this in inference contexts.
- `keyof SinglePropObject` is simplified to a string literal type as it's not a union so naturally `.origin` is not tracked in such a case. It's not impossible to track this but it would raise the memory footprint. Maybe this extra `.origin` tracking could only be done by the `tsserver`?
- "complicated" intersections/unions (such as `keyof Obj1 & keyof Obj2`) don't always track this information so it won't always work. Some of that is likely fixable. A good enough solution, oriented on a specific subset of patterns could also be an option (as in - without touching how and how often `.origin` gets created today).
- it's just about go to definition - other features such as reference finding are unaffected by this

closes https://github.com/microsoft/TypeScript/issues/49033

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -1664,6 +1664,10 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         getResolvedSignature: (node, candidatesOutArray, argumentCount) => getResolvedSignatureWorker(node, candidatesOutArray, argumentCount, CheckMode.Normal),
         getCandidateSignaturesForStringLiteralCompletions,
         getResolvedSignatureForSignatureHelp: (node, candidatesOutArray, argumentCount) => runWithoutResolvedSignatureCaching(node, () => getResolvedSignatureWorker(node, candidatesOutArray, argumentCount, CheckMode.IsForSignatureHelp)),
+        getContextualOriginTypeForStringDefinition: (node: StringLiteralLike) => {
+            const contextualType = getContextualType(node, /*contextFlags*/ undefined);
+            return contextualType && contextualType.flags & TypeFlags.Union ? (contextualType as UnionType).origin : undefined;
+        },
         getExpandedParameters,
         hasEffectiveRestParameter,
         containsArgumentsReference,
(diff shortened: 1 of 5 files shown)


sandersn:
I asked @iisaduan and @weswigham to review this but I'm not sure how ready it is. Should it still be a draft? Is it nearly complete?

Andarist:
I mention the status in the PR's description:
> I asked @iisaduan and @weswigham to review this but I'm not sure how ready it is. Should it still be a draft? Is it nearly complete?

I'm ready to commit time to this to work but for now, I'm just testing out the waters. I want to confirm this approach has merit before writing more code here

sandersn:
Makes sense. @weswigham and @iisaduan can you give an opinion about whether this is going in the right direction?

weswigham (approved):
I'm not sure we'd want to *rely* on union `.origin` being around to provide services (since any type that gets intern'd or dedupe'd for performance is liable to have an incorrect one or have it removed), but using it like this to provide a nicer, more connected result when it *is* present certainly seems OK.

Andarist:
It will lead to quite an inconsistent behavior if we don't try to improve some of the mentioned scenarios. Especially points 2 and 3 are worrying.

sandersn:
Double-checking: @weswigham this is good to go, right? Although since it's not a bug fix I guess we should wait for main to switch to 5.6 to merge it.

Andarist:
I think before landing this the team should comment on the potential inconsistent behaviors resulting from this. Users won't always understand why it works in certain situations while it won't in others. I think there is even potential in tracking more `.origin`s when the checker is in the services mode to address most of those concerns. I'm not sure how the team would feel about it though so I refrained from expanding this implementation.

Issue #57710, opened 2024-03-09
Regression performance compiling types in version 5.4.2


georgekaran:
### 🔎 Search Terms

"regression", "5.4.2", "performance"

### 🕗 Version & Regression Information

- This changed between versions 5.3.3 and 5.4.2

### 💻 Code

Extended diagnostics in version 5.4.2
```bash
Files:                          4294
Lines of Library:              40640
Lines of Definitions:         180807
Lines of TypeScript:          430513
Lines of JavaScript:               0
Lines of JSON:                     0
Lines of Other:                    0
Identifiers:                  591567
Symbols:                     1232510
Types:                       4677381
Instantiations:            395496734
Memory used:                2330550K
Assignability cache size:    3974893
Identity cache size:           14482
Subtype cache size:            37794
Strict subtype cache size:      9374
I/O Read time:                 0.70s
Parse time:                    0.96s
ResolveModule time:            0.58s
ResolveTypeReference time:     0.01s
ResolveLibrary time:           0.01s
Program time:                  2.51s
Bind time:                     0.62s
Check time:                  367.45s
printTime time:                0.00s
Emit time:                     0.00s
Total time:                  370.58s
```

Extended diagnostics in version 5.3.3
```bash
Files:                         4290
Lines of Library:             40241
Lines of Definitions:        180807
Lines of TypeScript:         430513
Lines of JavaScript:              0
Lines of JSON:                    0
Lines of Other:                   0
Identifiers:                 591320
Symbols:                    1234773
Types:                       699200
Instantiations:             2247099
Memory used:               1308962K
Assignability cache size:    223823
Identity cache size:          14479
Subtype cache size:           28353
Strict subtype cache size:     9128
I/O Read time:                0.73s
Parse time:                   1.11s
ResolveModule time:           0.44s
ResolveTypeReference time:    0.01s
ResolveLibrary time:          0.02s
Program time:                 2.60s
Bind time:                    0.76s
Check time:                  13.75s
printTime time:               0.00s
Emit time:                    0.00s
Total time:                  17.11s
```

### 🙁 Actual behavior

The type validation using `tsc --noEmit` increase more than 20 times using the version `5.4.2` comparing with the version `5.3.3`. 

In the CI is always hitting the heap memory limit:
![image](https://github.com/microsoft/TypeScript/assets/23644327/95c713b0-8938-48ac-bd3d-26861eb6d45b)

### 🙂 Expected behavior

[...]

PR #57730, opened 2024-03-11
Automatic perf stats in baselines

RyanCavanaugh:
This adds automatic baselines similar to what you get from `--extendedDiagnostics` whenever the harness detects an "interesting" amount of something happening. To reduce baseline noise from future changes, these values are rounded to metric-specific intervals, so e.g. if we go from 513 to 517 symbols, that's not going to register a baseline diff (it will print back as "500").

This should let us easily create "don't blow stuff up" tests when things like #57710 happen.

I want to make this opt-in/opt-out on a per-test base but the harness code is not well-suited to threading that info through and I want to do a follow-up refactor that cleans up the baseliner to have access to the per-test options.


src/harness/harnessIO.ts
```diff
@@ -776,12 +776,36 @@ export namespace Compiler {
...
+                perfStats.push(["Identity cache", 1000, 100, caches.identity]);
+                perfStats.push(["Assignability cache", 1000, 100, caches.assignable]);
+                perfStats.push(["Type Count", 1000, 100, checker.getTypeCount()]);
+                perfStats.push(["Instantiation count", 1500, 500, checker.getInstantiationCount()]);
+                perfStats.push(["Symbol count", 45000, 500, checker.getSymbolCount()]);
+
+                if (perfStats.some(([, threshold, , value]) => value >= threshold)) {
+                    perfLines.push(`//// Performance Stats`);
+                    for (const [name, _, rounding, value] of perfStats) {
+                        const display = (Math.round(value / rounding) * rounding).toLocaleString("en-US");
```
jakebailey, line 795: If zero, maybe we should just drop it?

DanielRosenwasser:
What happens if a test uses `noTypeAndSymbolsBaselines` or whatever the flag is called? I guess we just don't care usually?

jakebailey (approved):

Issue #57738, opened 2024-03-12
TS v5.4.2 errors on referencing a property `delete` in JSDoc `@link` or `@linkcode`


jaydenseric:
### 🔎 Search Terms

"jsdoc link keyword delete", "Identifier expected. 'delete' is a reserved word that cannot be used here. ts(1359)"

### 🕗 Version & Regression Information

- This changed between versions 5.3.3 and 5.4.2

The bug was not present in TS v5.3.3; upgrading to v5.4.2 exhibits the problem.

### ⏯ Playground Link

https://www.typescriptlang.org/play?moduleResolution=99&target=99&jsx=0&module=199&ssl=3&ssc=26&pln=1&pc=1&filetype=js#code/PTAEAEBcGcFoGMAWBTeBrAUB4AqHoBvcAGwEsA7NUAWQEMAHAOgBNljlJkBfUHYIA

### 💻 Code

```js
// @ts-check

/** {@link Map.delete} */
```

### 🙁 Actual behavior

A TypeScript error on the `delete` in the JSDoc `@link`:

```
Identifier expected. 'delete' is a reserved word that cannot be used here. ts(1359)
```

### 🙂 Expected behavior

No TypeScript error, and the `delete` in the JSDoc `@link` should have hover information.

### Additional information about the issue

Mentioned this bug here:

https://github.com/microsoft/TypeScript/issues/53111#issuecomment-1991191200

[`graphql-react`](https://www.npmjs.com/package/graphql-react) users who upgrade TypeScript in their projects to v5.4.2 suddenly start seeing TypeScript errors wherever imported `graphql-react` modules have a JSDoc `@link` referencing a property `delete`, e.g:

https://github.com/jaydenseric/graphql-react/blob/5525aca4e5ba5aad5115e56de92bf5ab4acbd2c9/cacheDelete.mjs#L11

Andarist:
Bisects to https://github.com/microsoft/TypeScript/pull/56338

jaydenseric:
This was closed accidentally by a private GitHub repo commit having the description:

```
TypeScript can’t be updated yet until this bug is fixed:

https://github.com/microsoft/TypeScript/issues/57738
```

With `fixed:` before the issue link.

PR #57739, opened 2024-03-12
Fixed referencing properties that are reserved keywords in `@link`

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/57738

### DIFF
--- a/src/compiler/parser.ts
+++ b/src/compiler/parser.ts
@@ -9227,7 +9227,7 @@ namespace Parser {
 
                     let name: EntityName | JSDocMemberName = parseIdentifierName();
                     while (parseOptional(SyntaxKind.DotToken)) {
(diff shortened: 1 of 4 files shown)


sandersn (approved):

jaydenseric:
When can this fix be released? Several TypeScript patch releases have been published since this PR was merged, but none seem to include the fix. Several open and closed source repos I'm working on can't update TypeScript until this fix is published.

DanielRosenwasser:
@jaydenseric sorry about that - it would have been ideal to include this fix in the patch since this was a regression from the last version. If there is a need for another patch, we'll try to pick it in. Otherwise, we'd like to limit the number of versions we publish.

In the meantime, I'd recommend using [the nightly TypeScript/JavaScript extension for VS Code](https://marketplace.visualstudio.com/items?itemName=ms-vscode.vscode-typescript-next) and/or installing a nightly over npm with `typescript@next`.

PR #57749, opened 2024-03-12
Report grammar errors for invalid decorator grammar

### LINKED ISSUE #55336: TS 5.0 decorators don't follow the DecoratorCallExpression grammar
### 🔎 Search Terms

DecoratorCallExpression

- This issue is specific to TypeScript 5.0+ decorators (not experimental decorators)

### ⏯ Playground Link

https://www.typescriptlang.org/play?target=99&ts=5.1.6#code/CYUwxgNghgTiAEEQBd4A8Bc8AUBKeAvAHzwDe8AnlnoSdpvFAHYUA0lWzF+xjL8AXwCwAKAACaPADoKeUZCgBnRfABmAe3VlhIoA

### 💻 Code

```ts
declare let x: () => { y: () => (x: any, y: any) => any }
@x().y()
class foo {}
```

### 🙁 Actual behavior

TypeScript allows this invalid syntax without any errors.

### 🙂 Expected behavior

Here's the grammar (taken from the spec linked in [the proposal](https://github.com/tc39/proposal-decorators)):

```
Decorator[Yield, Await] :
    @ DecoratorMemberExpression[?Yield, ?Await]
    @ DecoratorParenthesizedExpression[?Yield, ?Await]
    @ DecoratorCallExpression[?Yield, ?Await]

DecoratorMemberExpression[Yield, Await] :
    IdentifierReference[?Yield, ?Await]
    DecoratorMemberExpression[?Yield, ?Await] . IdentifierName
    DecoratorMemberExpression[?Yield, ?Await] . PrivateIdentifier

DecoratorParenthesizedExpression[Yield, Await] :
    ( Expression[+In, ?Yield, ?Await] )

DecoratorCallExpression[Yield, Await] :
    DecoratorMemberExpression[?Yield, ?Await] Arguments[?Yield, ?Await]
```

My reading of this grammar is that TypeScript should only be allowing a single `Arguments` production at the end of the decorator. So I believe TypeScript should consider this example to be a syntax error. Here's what other decorator parsers do:

* [Babel](https://babel.dev/repl#?browsers=&build=&builtIns=false&corejs=3.21&spec=false&loose=false&code_lz=CYUwxgNghgTiAEEQBd4A8Bc8AUBKeAvAHzwDe8AnlnoSdpvFAHYUA0lWzF-xjL8AXwBQAATR4AdBTxDIUAM7z4AMwD2qssKA&debug=false&forceAllTransforms=false&modules=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Cstage-0%2Cstage-1%2Cstage-2%2Cstage-3%2Ctypescript&prettier=false&targets=&version=7.22.10&externalPlugins=%40babel%2Fplugin-syntax-explicit-resource-management%407.20.0&assumptions=%7B%7D):

    ```
    /repl.ts: Leading decorators must be attached to a class declaration. (2:4)

      1 | declare let x: () => { y: () => (x: any, y: any) => any }
    > 2 | @x().y()
        |     ^
      3 | class foo {}
      4 |
    ```

* [esbuild](https://esbuild.github.io/try/#dAAwLjE5LjAALS1sb2FkZXI9dHMAZGVjbGFyZSBsZXQgeDogKCkgPT4geyB5OiAoKSA9PiAoeDogYW55LCB5OiBhbnkpID0+IGFueSB9CkB4KCkueSgpCmNsYXNzIGZvbyB7fQ):

    ```
    ✘ [ERROR] Expected "class" after decorator but found "."

        <stdin>:2:4:
          2 │ @x().y()
            ╵     ^

      The preceding decorator is here:

        <stdin>:2:0:
          2 │ @x().y()
            ╵ ^

      Decorators can only be used with class declarations.
    ```

rbuckton:
This reports a grammar error when invalid syntax is used in a decorator expression that does not match the proposed syntax for Stage 3 Decorators. In addition, this adds quick fixes to add parentheses around the expression to make it valid.

This is a somewhat aggressive error as we previously were far more flexible for decorator expressions under `--experimentalDecorators`, which could mean new errors reported in existing code. Since parse and emit do not change, and since there is a quick-fix available, I'd like to try to go ahead with this change as is. However, if we feel it is too much of break, I can change the grammar check to only apply when targeting ES decorators.

Fixes #55336


jakebailey:
@typescript-bot test top400
@typescript-bot run dt

@typescript-bot pack this

weswigham (approved):
Should we also test `@import.meta.decorator`, since we have tests for `super` and `this`?

jakebailey:
All of the user / top tests seem to be okay with this, which is good news.

rbuckton:
> Should we also test `@import.meta.decorator`, since we have tests for `super` and `this`?

I'm not sure why we would. Those tests specifically test which `this` is in scope.

rbuckton:
@typescript-bot pack this

andrewbranch (approved):

ef4:
> However, if we feel it is too much of break, I can change the grammar check to only apply when targeting ES decorators.

That would be a lot better, speaking as a framework maintainer that's trying to help shepherd a big ecosystem of community maintained code across the migration from experimental decorators to stage3 decorators. As experimental decorators are clearly an end-of-the-road feature that's maintained only for compatibility, anything you can do to not destabilize them unnecessarily would be helpful.

StNekroman:
Why this touches legacy experimental decorators , despite PR's description was about "Stage 3 Decorators" ?
Any good reason for breaking backward compatibility?

PR #57750, opened 2024-03-12
Revert "Allow (non-assert) type predicates to narrow by discriminant"

### LINKED ISSUE #57690: Narrowed type not "carried" iteratively
### 🔎 Search Terms

narrow, generic

- This changed between versions 5.3.3 and 5.4.2

### ⏯ Playground Link

https://www.typescriptlang.org/play?#code/PTAEF5NB1B5AlA0gZQpdH0CgsBMCmAxgDYCGATvqAGYCuAdoQC4CWA9vaCwM4AKZLerABGAKyJMAjAAosoULlJNSALlAMA1vTYB3egBosASjWLlXbqHhE25XAB5e5NgAd85JgE9E+T-vX0Wrr0AHwA3DgEJBRUhBzcTKAAtp4iopJqmtp6ESzUoNI8-KSCaRIyKWmSRkagAN5y8jS2BXH0CaAavqBs+QDaAESkA-4DwiOgA4QDALq1DU1NbR0E+G7kEMmpYpJ9XZ4zEYvyIIsAegD8jYvLiavrPp6W4BbFpWLl0vfutRegZcwAHT7bhffBrH6gNR9Q7XUAAXywiKwp3QoAAIrAAKLIAByABUYAhEGhMGTsKdYPRiJ4FHlqO58IwqMJ8EwdODOEwABZUABMXCSLmIlh4oB5VAA5kz3CxCKBCYJxbzxZ43P58AA3JmgHS8zgsRI8egAckStG4+FwgJtkSIZEoNAYzHYBr4AiEH2YfPs+JCsnkZlUASCekMJgUSlIFisNjsjmc6y8j38WWC4Tt0Udty2aT5mUC2XouXyhXdJU94m90kqYj5NXqcOoLWkOf2PX6QwmY270zmjeOoBz3w2L1roj5e18sMHpyalzh8mH4Ievmerw9AKYfLBEPIv3+XqYwLXu-WtWhM6aiMRQA

### 💻 Code

```ts
// === WORKS =============

declare function isPlainObject1(
  data: unknown,
): data is Record<PropertyKey, unknown>;

declare const myObj1: unknown;
if (isPlainObject1(myObj1)) {
    for (const key of ["a", "b", "c"]) {
      const deeper = myObj1[key];
      //    ^?
      const deeperKeys = isPlainObject1(deeper) ? Object.keys(deeper) : [];
    }
}

// === DOESNT WORK ================
// Only difference between the 2 impls is the generic T in the type, even when it isn't used...

declare function isPlainObject2<T>(
  data: unknown,
): data is Record<PropertyKey, unknown>;

declare const myObj2: unknown;
if (isPlainObject2(myObj2)) {
    for (const key of ["a", "b", "c"]) {
      const deeper = myObj2[key];
      //    ^?
      const deeperKeys = isPlainObject2(deeper) ? Object.keys(deeper) : [];
    }
}
```

### 🙁 Actual behavior

The `deeper` variable, which is nested deeper from the narrowed type is inferred as `any`.

### 🙂 Expected behavior

It is the result of accessing a prop on a `Record<PropertyKey, unknown>` so is expected to be `unknown`.

I maintain the Remeda utility functions library, I just got this reported here: https://github.com/remeda/remeda/issues/559 and there are additional anecdotal reports that upgrading to typescript 5.4 breaks several of our types.

I couldn't find a reasoning to this issue, including any new changes to Typescript in 5.4 which might explain it.

I'm not sure my title makes sense for this bug, as I don't fully understand what's going on here, or even how to build a truly *minimal* example.

### LINKED ISSUE #57705: TS7022 circularity regression in 5.4.2
### 🔎 Search Terms

TS7022 circularity

- This changed between versions 5.3.3 and 5.4.2

### ⏯ Playground Link

* [5.4.2](https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXywGcAeAFQD4AKANwC55SBKe6gwhgbgCgw9CN4OegCVwOGMGL8YWVAHMANPGmy55eAB94aUIlkhg8ALzwA3gF9uWRPEo5GZrvCTjbvVP3gBrEAE8C+PaOzs7untRQEMgIJjgA2j6+ALrcIQQ2lERSGDLyVBFRIIwOpk4h5mUV5kA)
* [5.3.3](https://www.typescriptlang.org/play?ts=5.3.3#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXywGcAeAFQD4AKANwC55SBKe6gwhgbgCgw9CN4OegCVwOGMGL8YWVAHMANPGmy55eAB94aUIlkhg8ALzwA3gF9uWRPEo5GZrvCTjbvVP3gBrEAE8C+PaOzs7untRQEMgIJjgA2j6+ALrcIQQ2lERSGDLyVBFRIIwOpk4h5mUV5kA)

### 💻 Code

```ts
declare function is<T>(v: T): v is T;
const o: Record<string, string> | undefined = {};
if (o) {
  for (const key in o) {
    const value = o[key];
    if (is<string>(value)) {
    }
  }
}
```

### 🙁 Actual behavior

TS7022: `value` implicitly has type `any` because it does not have a type annotation and is referenced directly or indirectly in its own initializer.

### 🙂 Expected behavior

No error, like in 5.3.3.

_No response_

gabritto:
Fixes #57690.
Fixes #57705.

The underlying problem that my original PR exposes seems to be that, when we try to narrow by type predicate, we call `getEffectsSignature`, which then in many cases results in calling `resolveSignature`, which results in type checking something we're already in the middle of type checking, resulting in circularities.

I think there isn't an immediate fix that would work for all of the reported examples or all possible new circularity cases, so I'm reverting the change for now. But in the future, we might come up with a fix for the underlying problem and we can then bring this change back.


gabritto:
Self-check is failing because of [this change](https://github.com/microsoft/TypeScript/commit/3a0869fd97ea38d2eedcda48d4c794487c345c42#diff-4dd2736e47d6b51eaf8ff275c70e53dc5f6bca718a54a2af474cd06147413f68L3512).

I guess we have to merge this PR, then after we publish a patch, update the ts dependency and also bring back that cast.

RyanCavanaugh:

weswigham (approved):

jakebailey:
This will need to be manually cherry-picked; not sure if someone's done that already.

lazytype:
> But in the future, we might come up with a fix for the underlying problem and we can then bring this change back.

Is there any chance that this fix happens before the next patch release of 5.4? It's somewhat thrashy to have to migrate code back and forth especially between just patch versions.

- TS 5.4 comes out
```diff
switch
...
default:
-  assertNever(foo.type);
+  assertNever(foo);
```

- This PR gets is released
```diff
switch
...
default:
-  assertNever(foo);
+  assertNever(foo.type);
```

- Future fix
```diff
switch
...
default:
-  assertNever(foo.type);
+  assertNever(foo);
```

jakebailey:
No, a future fix would be in 5.5 or later, hence the nebulous "in the future, we might come up with a fix". We typically just revert anything that is nontrivial to fix for patch releases. I guess that's "thrashy", but TS 5.4's only been out for two weeks and reverting this is safest.

niels-vdz:
> This will need to be manually cherry-picked; not sure if someone's done that already.

It looks like this revert was not included in v5.4.3 that was just released. Will this revert be included in the next patch release?

jakebailey:
This was definitely included in 5.4.3: https://github.com/microsoft/TypeScript/pull/57795

PR #57768, opened 2024-03-13
Allow `target`-only ES decorators

### LINKED ISSUE #53208: TypeScript 5.0 decorators don't work when `noUnusedParameters` is enabled and the context is unused
# Bug Report

In the current version of TypeScript, decorators which don't use the context do not work properly when `noUnusedParameters` is enabled. If the decorator includes a context parameter, then the compiler reports an error on the unused parameter. If the decorator does not include the context parameter, than the decorator cannot be used, as the decorator is called with one too many arguments.

### 🔎 Search Terms

`decorator, decorators`

This changed between versions 4.9 and 5.0

### ⏯ Playground Link

[Playground link with relevant code](https://www.typescriptlang.org/play?ts=5.1.0-dev.20230311&experimentalDecorators=false#code/GYVwdgxgLglg9mABAWQJ4BECmE4CcCGUeAFFPrgOaZQBciYmA7osQHTvkUDOdDAbplwBKRAF4AfInABrMHEZgRAbwCwAKESbEOMFzgAbTK31wKpTtSEBudQF916gAJosOAkVyJ1Effi5cUVABhX39EJXs1IA)

### 💻 Code

```ts
function MyDecorator(target: new (...args: never) => unknown) {
    console.log(target);
}

@MyDecorator 
class MyClass {}
```

### 🙁 Actual behavior

Code compiles without errors.

### 🙂 Expected behavior

TypeScript throws the following error:

```
Unable to resolve signature of class decorator when called as an expression.
  The runtime will invoke the decorator with 2 arguments, but the decorator expects 1.
```

rbuckton:
This allows users to write ES decorators that contain only a `target` parameter.

Fixes #53208

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -34447,7 +34447,8 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     function getDecoratorArgumentCount(node: Decorator, signature: Signature) {
         return compilerOptions.experimentalDecorators ?
             getLegacyDecoratorArgumentCount(node, signature) :
-            2;
+            // Allow the runtime to oversupply arguments to an ES decorator as long as there's at least one parameter.
+            Math.min(Math.max(getParameterCount(signature), 1), 2);
     }
 
     /**
--- a/tests/baselines/reference/esDecorators-arguments.errors.txt
+++ b/tests/baselines/reference/esDecorators-arguments.errors.txt
@@ -1,22 +1,17 @@
 esDecorators-arguments.ts(1,2): error TS1238: Unable to resolve signature of class decorator when called as an expression.
   The runtime will invoke the decorator with 2 arguments, but the decorator expects 0.
-esDecorators-arguments.ts(2,2): error TS1238: Unable to resolve signature of class decorator when called as an expression.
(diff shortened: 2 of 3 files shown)


fatcerberus:
Yeah, I was confused by this too--in general, function types with fewer parameters are assignable to ones with more (there's even a FAQ entry explaining this), so having a minimum-parameter-count check in the first place made no sense to me and I'd expect the check to just be, is it assignable to the correct function type

rbuckton:
> Is the requirement for at least one parameter (`target`) just a sanity check? It seems inconsistent with our function assignability rules.

IIRC, this is a sanity check that @DanielRosenwasser added many years ago to catch a failure to invoke a decorator factory that takes no arguments. See `isPotentiallyUncalledDecorator`.


> Yeah, I was confused by this too--in general, function types with fewer parameters are assignable to ones with more (there's even a FAQ entry explaining this), so having a minimum-parameter-count check in the first place made no sense to me and I'd expect the check to just be, is it assignable to the correct function type

We don't type check the decorator as if it were a callback, we type check it as if it were a direct invocation by the runtime. This is a very narrow fix to address a specific case within decorators. Changing the mechanics to use assignability would be a fairly extensive rewrite.

sheetalkamat (approved):

PR #57769, opened 2024-03-13
fix(57761): Invalid verbatimModuleSyntax quick fix offered when also reexporting a value

### LINKED ISSUE #57761: Invalid `verbatimModuleSyntax` quick fix offered when also reexporting a value
### 🔎 Search Terms

quick fix, verbatimModuleSyntax, value re-export 

- This changed between versions 5.1.6 and 5.2.2
- This changed in PR #54255

### ⏯ Playground Link

https://www.typescriptlang.org/play?verbatimModuleSyntax=true&ts=5.4.2#code/JYWwDg9gTgLgBAbzgQQEYGcZQIYGMYBqwApgO4A0cIxIEcAvnAGZQQhwBEUxeMHA3ACgA9MLgA9cXBik6ARwCuwXAGtmwAB7o4wAHa4ANgoAmegOZwA5KEixpATzDE4AOjeW4gGXJBg4htvwSDCOzmiYOPhEZJTUtAw+onC4bGDY3MbSdKjEBhCkAFw+NtCBcAAq2How2KgGxACqusCKxATYRsSUCujEALI0dAzMrOxcPPgCImKScBC6BvbqGjraegZ6zpbBTh6AoOQ+fgFJc5hw3X0DAExwALxnPf20Qr7+JQ5OiOWVutW1DU0tNodeJAA

### 💻 Code

```ts
// verbatimModuleSyntax: true

import { AbstractView, memo } from "react";
// ^^ two quick fixs including 'import type ...' ❌

export { type AbstractView, memo }

// compared to below:

import { TaintableUniqueValue, useMemo  } from "react";
// ^^ only fix is inline 'type' ✅

export const useMemo2 = useMemo;

export type { TaintableUniqueValue }
```

### 🙁 Actual behavior

Quick fix is offering to convert the whole import to a `import type`.

### 🙂 Expected behavior

Only offer a quick fix of marking the individual named import with `type <name>` 

This only appears to happen when the values being imported are only re-exported.

While a user is free to pick the more appropriate quick fix, it adds complexity [when running the quickfixes on mass](https://github.com/microsoft/ts-fix) on large code bases where making individual judgment calls for each fix could take too long.

a-tarasyuk:
Fixes #57761


tests/cases/fourslash/codeFixConvertToTypeOnlyImport4.ts
```diff
@@ -10,13 +10,14 @@
...
+
+verify.codeFixAvailable([
+    { description: "Use 'type I'" },
+    { description: "Remove import from './b'" }
+]);
+
 verify.codeFix({
     index: 0,
-    description: ts.Diagnostics.Use_import_type.message,
-    newFileContent: `import type { I, foo } from "./b";`
```
andrewbranch, line 16: This is a valid fix and it’s intended to be offered, because `foo` is never used in value space in `a.ts`. I think the right fix is to recognize the non-type-only re-export in the new test case as a value-space usage in the FindAllReferences API call.

andrewbranch (changes requested):

tests/cases/fourslash/codeFixConvertToTypeOnlyImport4.ts
a-tarasyuk: Oke, thanks

src/services/codefixes/convertToTypeOnlyImport.ts
```diff
@@ -104,7 +106,11 @@ function canConvertImportDeclarationForSpecifier(specifier: ImportSpecifier, sou
     const checker = program.getTypeChecker();
     for (const specifier of nonTypeOnlySpecifiers) {
         const isUsedAsValue = FindAllReferences.Core.eachSymbolReferenceInFile(specifier.name, checker, sourceFile, usage => {
-            return !isValidTypeOnlyAliasUseSite(usage);
+            if (isValidTypeOnlyAliasUseSite(usage)) {
+                return true;
```
andrewbranch, line 110: 🤔 if a usage is in a type location, `isUsedAsValue` becomes true?

src/services/codefixes/convertToTypeOnlyImport.ts
```diff
@@ -104,7 +106,11 @@ function canConvertImportDeclarationForSpecifier(specifier: ImportSpecifier, sou
     const checker = program.getTypeChecker();
     for (const specifier of nonTypeOnlySpecifiers) {
         const isUsedAsValue = FindAllReferences.Core.eachSymbolReferenceInFile(specifier.name, checker, sourceFile, usage => {
-            return !isValidTypeOnlyAliasUseSite(usage);
+            if (isValidTypeOnlyAliasUseSite(usage)) {
+                return true;
+            }
+            const symbol = checker.getSymbolAtLocation(usage);
+            return !!(symbol && skipAlias(symbol, checker).flags & SymbolFlags.Value);
```
andrewbranch, line 113: You should also check whether the alias’s value meaning resolves through a type-only import/export already, via `checker.getTypeOnlyAliasDeclaration(symbol, SymbolFlags.Value)` (you’ll need to update the signature on `TypeChecker`). If that call returns something, then this value reference isn’t really a value reference.

a-tarasyuk: Oke, I’ve changed

src/services/codefixes/convertToTypeOnlyImport.ts
```diff
@@ -104,7 +105,8 @@ function canConvertImportDeclarationForSpecifier(specifier: ImportSpecifier, sou
     const checker = program.getTypeChecker();
     for (const specifier of nonTypeOnlySpecifiers) {
         const isUsedAsValue = FindAllReferences.Core.eachSymbolReferenceInFile(specifier.name, checker, sourceFile, usage => {
-            return !isValidTypeOnlyAliasUseSite(usage);
+            const symbol = checker.getSymbolAtLocation(usage);
+            return !isValidTypeOnlyAliasUseSite(usage) || !!symbol && !checker.getTypeOnlyAliasDeclaration(symbol, SymbolFlags.Value);
```
andrewbranch, line 109: This needs a few more tests. I think you may still need to `skipAlias` and check flags like you had in the previous commit as part of this condition. It’s too complicated to guess at without test coverage though. Here are some cases, assuming the same `b.ts` and test execution.

Should use `import type` (`f` has no reified export):
```ts
// @filename: /a.ts
function f() {}
export interface I {}
export type { f };
```

```ts
// @filename: /a.ts
export interface f {}
export interface I {}
```

Should use `import { type` (`f` has reified export):
```ts
// @filename: /z.ts
export function f() {}

// @filename: /a.ts
import { f } from "./z";
export interface I {}
type f = number;
export { f };
```

a-tarasyuk: Should we have a `QF' suggestion in the last example?

andrewbranch (approved):
Thanks!

src/services/codefixes/convertToTypeOnlyImport.ts
andrewbranch: I was assuming there would still be a `b.ts` paired with the last example that imports and re-exports `f` and `I`

tests/cases/fourslash/codeFixConvertToTypeOnlyImport12.ts
```diff
@@ -8,11 +8,18 @@
...
 
-goTo.file("/b.ts");
-verify.codeFixAvailable([ ]);
+// @filename: /c.ts
+////import { I, f } from "./b";
+////export { type I, f };
+
+goTo.file("/c.ts");
+verify.codeFixAvailable([
+    { description: "Use 'import type'" },
```
andrewbranch, line 23: This looks wrong. The real function `f` is being exported in the JS emit, and this fix would change that.

andrewbranch (changes requested):

src/services/codefixes/convertToTypeOnlyImport.ts
```diff
@@ -148,3 +155,10 @@ function doChange(changes: textChanges.ChangeTracker, sourceFile: SourceFile, de
         }
     }
 }
+
+function skipAlias(checker: TypeChecker, symbol: Symbol) {
+    while (symbol.flags & SymbolFlags.Alias) {
+        symbol = checker.getAliasedSymbol(symbol);
+    }
```
andrewbranch, line 162: This would have the wrong effect if the type and value locations were swapped in my symbol merging example. What you really want is `getAllSymbolFlags` from checker.ts, which you could export as `@internal`. It also has an `includeTypeOnlyMeanings` parameter, which I think could eliminate the need to call `checker.getTypeOnlyAliasDeclaration`.

a-tarasyuk: I considered `symbolIsValue` but opted not to extend the `checker` interface.

andrewbranch (approved):

acutmore:
Thanks!

Issue #57756, opened 2024-03-12
Design Meeting Notes, 3/12/2024


DanielRosenwasser:
# Reusing Type Nodes from Assertion Expressions

https://github.com/microsoft/TypeScript/issues/57587
https://github.com/microsoft/TypeScript/pull/57589

* Gist of the PR is that when you have an `as` assertion, like `const x = foo as number`, you should be able to do declaration emit as `const x: number = foo`.
* Valid to move between a top-level type in an `as` versus an annotation.
* Last time we discussed, we liked this because it reused more nodes, but concerns were raised that this logic should live in the `TypeChecker` versus the `EmitResolver`.
* `TypeChecker` and its `NodeBuilder`s takes care of many of the concerns.
* If we want to scale out builds, relying on the `TypeChecker` for this causes problems.
    * But if we actually care about perf, why are we not committing to a fast `transpileDeclaration`?
* Current feedback we've provided is just to stick to checks and make our output conform more with an isolated declaration emitter.
    * A faster implementation can happen down the line.
    * Or an alternative emit resolver can do the logic in the PR instead of calling out to the checker.
* Also: doing this in the checker means that JS declaration emit gets better.
* A lot of PRs for a large change - asking to reorganize this has a lot of back-and-forth and risks slipping on 5.4.
* Conclusion: we will prototype a change where this work is done in the `NodeBuilder`.

# Do not preserve triple-slash references unless marked `preserve`

https://github.com/microsoft/TypeScript/issues/57655
https://github.com/microsoft/TypeScript/pull/57681

* We have a ton of test cases that depend on this stuff working.
* Have to decide for these test cases whether:
    * The new baselines are good.
    * Where we need to specify `preserve`.
    * The tests need to be fixed in some other configuration manner.
* Seeing what was affected in the top 200 mostly showed up stuff where people would have React, Node, or whatever loaded up anyway.
* Is `isolatedDeclarations` going to be compatible with `--module amd/umd`?
    * Mostly that `isolatedModules` is incompatible with `--outFile`, which those work with.
* Adding `preserve` is not the worst workaround.
* Two classes of issues:
    * Used to generated a reference, now don't.
    * References that used to be preserved, now don't.

# Spec conformance (break) on decorator syntax in `--experimentalDecorators`

https://github.com/microsoft/TypeScript/pull/57749

* We have a fairly lax sytnax
* Decorators proposal says either parenthesized expressions or dotted names with a trailing call expression.

  ```ts
  (
[...]

PR #57772, opened 2024-03-14
Look for usable type nodes in associated expressions for declaration emit

### LINKED ISSUE #57587: [ID-Prep] Preserve type nodes from type assertions in declarations

weswigham:
Per our design meeting notes (#57756), to expedite things a bit, this is a version of #57589 that (re)uses the machinery we already have for node reuse within the `NodeBuilder` in the checker.

Honestly, a worthwhile endeavor, since the test cases from the other PR let me find a bug in the existing node reuse code where we weren't reusing any nodes with a parameter or property in them (oops). The quick fix for that is responsible for most of the test whitespace changes in this PR (reused nodes don't get `EmitFlags.SingleLine` set, new nodes always do - and if that's a good thing or not is another discussion) - but I could split that short change out, if that was important for comparison (but it'd make two of the examples in `tests\cases\compiler\declarationEmitCastReusesTypeNode5.ts` not be copied - that's what made me notice the issue).

I've also normalized the new tests a bit, so all the similar ones test all the same variations, in addition to a couple extra variations. Notable differences from the original PR include this working for JS declaration emit, for `set` accessors, and for annotated required initialized parameters. I also took the opportunity to normalize all the callers of `serializeExistingTypeNode` (since I was adding another one or two) to perform the *same* sanity checking on the type (so all callers ensure the type node we're preserving actually resolves to the type we're looking at, and that if we're looking at a type reference, we don't try and reuse a type reference with less than the required number of type arguments (an error that is kind-of allowed in JS, but definitely results in an error if you copy it unmodified into TS)).

Fixes #57587


tests/baselines/reference/optionalMethods.js
```diff
@@ -138,12 +138,12 @@ interface Foo {
 }
 declare function test1(x: Foo): void;
 declare class Bar {
-    d?: number | undefined;
+    d?: number;
```
jakebailey, line 141: I am somewhat suspicious of the class properties just because of exactOptionalPropertyType; the other PR went out of its way to add `| undefined` back, but maybe this doesn't matter in practice given the code that generates it doesn't have the flag on so produces code for that case (sorta like if you compile without strictNullChecks... you get weird stuff).

weswigham: As does this PR... for required initialized parameters. Properties, however, retain their `?`, and only even "do the right thing" under `exactOptionalPropertyTypes` when you *don't* print an `undefined` unless the user wrote it (and, under any other mode, it doesn't matter). We've been super loose and just print whatever type we get back (for odd reasons), while now we go "oh, this is optional, a `number` here means the same thing as the `number | undefined` we got back from the checker, so we can just reuse the `number` annotation". Notably, "reuse the annotation" is our normal behavior everywhere else (because this is definitely annotated) - it's actually a little confusing why we weren't retaining it here before.

weswigham (commented):
Neglected to post this last night, but I have some explanations for some of the smaller sub-changes here, that are a bit apart from the core change.

src/compiler/checker.ts
```diff
@@ -8460,17 +8525,15 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
          * Unlike `typeToTypeNodeHelper`, this handles setting up the `AllowUniqueESSymbolType` flag
          * so a `unique symbol` is returned when appropriate for the input symbol, rather than `typeof sym`
          */
-        function serializeTypeForDeclaration(context: NodeBuilderContext, type: Type, symbol: Symbol, enclosingDeclaration: Node | undefined, includePrivateSymbol?: (s: Symbol) => void, bundled?: boolean) {
+        function serializeTypeForDeclaration(context: NodeBuilderContext, type: Type, symbol: Symbol, enclosingDeclaration: Node | undefined, includePrivateSymbol?: (s: Symbol) => void, bundled?: boolean, addUndefined?: boolean) {
```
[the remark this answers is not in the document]
DanielRosenwasser: > but #57465 will change this

Would it be good to proactively add a test case for that?

src/compiler/checker.ts
```diff
@@ -8481,17 +8544,20 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             ) {
                 context.flags |= NodeBuilderFlags.AllowUniqueESSymbolType;
             }
-            const result = typeToTypeNodeHelper(type, context);
+
+            const decl = symbol.valueDeclaration || symbol.declarations?.[0];
```
DanielRosenwasser, line 8548:
```suggestion
            const decl = symbol.valueDeclaration ?? symbol.declarations?.[0];
```

src/compiler/checker.ts
```diff
@@ -8542,6 +8608,10 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             return { introducesError, node };
         }
 
+        /**
+         * Do you mean to call this directly? You probably should use `tryReuseExistingTypeNode` instead,
+         * which performs sanity checking on the type before doing this.
+         */
         function serializeExistingTypeNode(context: NodeBuilderContext, existing: TypeNode, includePrivateSymbol?: (s: Symbol) => void, bundled?: boolean) {
```
DanielRosenwasser, line 8615: Looks like this is only called in one other place - maybe call it


```suggestion
        function trySerializeExistingTypeNodeHelper(context: NodeBuilderContext, existing: TypeNode, includePrivateSymbol?: (s: Symbol) => void, bundled?: boolean) {
```

src/compiler/checker.ts
```diff
@@ -6455,6 +6456,8 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     function createNodeBuilder() {
         return {
             typeToTypeNode: (type: Type, enclosingDeclaration?: Node, flags?: NodeBuilderFlags, tracker?: SymbolTracker) => withContext(enclosingDeclaration, flags, tracker, context => typeToTypeNodeHelper(type, context)),
+            expressionOrTypeToTypeNode: (type: Type, expr: Expression | JsxAttributeValue | undefined, addUndefined?: boolean, enclosingDeclaration?: Node, flags?: NodeBuilderFlags, tracker?: SymbolTracker) => withContext(enclosingDeclaration, flags, tracker, context => expressionOrTypeToTypeNode(context, expr, type, addUndefined)),
```
DanielRosenwasser, line 6459: The ordering of these parameters is a little strange. Should `expr` go first?

DanielRosenwasser (commented):
Seems reasonable - the naming is a little interesting for `expressionOrTypeToTypeNode`, but we could rename in the future.

src/compiler/checker.ts
weswigham: `expressionOrType` matches the priority of the generator, but the argument order matches the required-ness of the types (`type` being required, and `expr` being allowed to be `undefined`). I can swap the arguments to match the name if you think that consistency is more important.

src/compiler/checker.ts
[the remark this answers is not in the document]
weswigham: It adds its' own tests for it and since you can't cast into a type predicate like `null! as x is number`, I don't think this PR has much to add on top of that. But it should still make the refactor I mentioned above easier to do - though I don't really want to do it here, since it'll probably require fixing some more discrepancies between the declaration emitter's existing type node reuse logic and the node builder's inner type node reuse logic, as while the later is generalized to apply at the top level (using `serializeTypeForDeclaration` on the top-level type node if present), I've not yet done a comparison to see if the behaviors are identical (though it's already in use for JS declarations, ofc).

src/compiler/checker.ts
DanielRosenwasser: It's not really meant to be an optional argument, just a potentially undefined one - so I'd prefer the `expr, type` ordering.

DanielRosenwasser:
Other than my comments (and the current merge conflict) I think this is good. Any other thoughts @RyanCavanaugh @jakebailey ?

jakebailey:
I think I still feel uneasy about the behavior of parameter properties. Take `tests/cases/compiler/declarationEmitParameterProperty.ts`, which has:

```ts
export class Foo {
  constructor(public bar?: string) {
  }
}
```

Our JS emit is:

```
var Foo = /** @class */ (function () {
    function Foo(bar) {
        this.bar = bar;
    }
    return Foo;
}());
```

`bar` is always assigned. Right now, the declaration file we emit is:

```ts
export declare class Foo {
    bar?: string | undefined;
    constructor(bar?: string | undefined);
}
```

But after this PR, it's:

```ts
export declare class Foo {
    bar?: string;
    constructor(bar?: string);
}
```

I feel like maybe I'm not understanding https://github.com/microsoft/TypeScript/pull/57772#discussion_r1525165795 or something, because this behavior really does look like a regression, and one of the ones that I pointed out in #53463.

weswigham:
@jakebailey I've brought back the old emit for optional parameter properties - we can look into changing it later. It feels like they don't really play well with `exactOptionalPropertyTypes` as they are today.

jakebailey:
> @jakebailey I've brought back the old emit for optional parameter properties - we can look into changing it later. It feels like they don't really play well with `exactOptionalPropertyTypes` as they are today.

Thanks; I do wonder if we could simply preserve parameter properties in declaration emit, eliminating the need to rewrite the properties and making other emitters easier to write (just leave it as-is). But, not in this PR.

tests/cases/compiler/declarationEmitCastReusesTypeNode4.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export class C {
+    field = /** @type {P} */(something);
+    /** @optional */ optField = /** @type {P} */(something); // not a thing
+    /** @readonly */ roFiled = /** @type {P} */(something);
+    method(p = /** @type {P} */(something)) {}
+    /** @param {number} req */
+    methodWithRequiredDefault(p = /** @type {P} */(something), req) {}
+
+    constructor(ctorField = /** @type {P} */(something)) {}
```
jakebailey, line 31: This looks like a typo; should be `public ctorField =`..., right?

weswigham: This is javascript, so no - since parameter properties are a syntax error in JS files.

weswigham: (And there's no jsdoc equivalent)

jakebailey: Oh, whoops, didn't notice that.

jakebailey (approved):

weswigham:
> I do wonder if we could simply preserve parameter properties in declaration emit, eliminating the need to rewrite the properties and making other emitters easier to write (just leave it as-is).

Unfortunately, we issue `A parameter property is only allowed in a constructor implementation.` when we read a parameter property from a constructor in a declaration file. So such a change would be a backwards incompatible change to declaration emit (though we admittedly still check it correctly, despite the error, so maybe not the worst break in emit in the world).

PR #57778, opened 2024-03-14
fix type import check for default-import/re-export in js files

### LINKED ISSUE #57740: No auto-import from aliased symbol from ts files in js files
### 🔎 Search Terms

default export
auto-import
barrel export

- This changed between versions __5.3__ and __5.4__

### ⏯ Playground Link

Reproduction repository instead of playground because it needs at least 2 files. One in ts and one in js.
https://github.com/jasonlyu123/auto-import-aliased

### 💻 Code

```ts
// foo.ts

export default function foo() {}

function foo2() {}
export { foo2 }
```

```ts
// index.js
foo
// ^trigger completion here
```

### 🙁 Actual behavior

No auto-import to import the function `foo` and `foo2`

### 🙂 Expected behavior

Has auto-import like in 5.3.

This only happens in js files when trying to auto-import values from a ts file and the symbol has some alias or is a default export. This seems to have to do with #53619. In another check for type-only-location, the symbol flags are combined with flags from the aliased symbol and the symbol has flags like `class` and `function`.

jasonlyu123:
Fixes #57740. Pretty much copied from `shouldIncludeSymbol`. `getCombinedLocalAndExportSymbolFlags` for the default export and `skipAlias` for `export { someFunciton }` and `export { someFunction } from './module'`.

### DIFF
--- a/src/services/completions.ts
+++ b/src/services/completions.ts
@@ -2704,7 +2704,8 @@ export function getCompletionEntriesFromSymbols(
     }
 
     function symbolAppearsToBeTypeOnly(symbol: Symbol): boolean {
-        return !(symbol.flags & SymbolFlags.Value) && (!isInJSFile(symbol.declarations?.[0]) || !!(symbol.flags & SymbolFlags.Type));
+        const flags = getCombinedLocalAndExportSymbolFlags(skipAlias(symbol, typeChecker));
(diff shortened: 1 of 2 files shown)


andrewbranch (approved):

iisaduan (approved):

andrewbranch:
Thanks @jasonlyu123!

andrewbranch:

PR #57782, opened 2024-03-14
Require exactly 'true' for truthy reference directive prop

jakebailey:
This is for #57681; noticed that we just say "if it's present it's true" rather than actually requiring the string true. Seems a little weird. No breaks locally, and didn't see much in https://github.com/search?q=%2Fno-default-lib%3D%22%2F+language%3Atypescript&type=code either.


jakebailey:
@typescript-bot test top200
@typescript-bot run dt

RyanCavanaugh (approved):

PR #57784, opened 2024-03-14
Add 'configurable: true' to descriptors in __createBinding and __setModuleDefault

### LINKED ISSUE #43081: tslib: "TypeError: Cannot redefine property:" Bug Appears To Have Returned
# Bug Report

Version 1.13.0 fixed the bug referenced here:
https://github.com/microsoft/tslib/issues/102

But, in moving to either 2.0.0 or 2.1.0 - the bug is happening again.

### 🔎 Search Terms

tslib 
TypeError: Cannot redefine property:

This is fixed in 1.13.0 - but broke again in both version 2.0.0 and 2.1.0 

- This prevents us from using tsconfig aliased paths in conjuncture with jest spyOn(..).  We can work around this by providing the relative paths instead of the paths as determined by the tsconfig - but this is against our code policy.
- This changed between versions 1.13.0 and 2.0.0

### 💻 Code

internal/business functions/directories scrubbed from the following code snippet - but in case this makes it more clear

```
import * as Utils from '@some-platform/core/libs/utils/support';

describe( 'test tslib compatibility with aliased imports', () => {

	it( 'jest spyOn', async () => {

		// test that we don't get the following error (tslib 2.0 and 2.1 will cause this):
		// TypeError: Cannot redefine property: foobar
		// at Function.defineProperty (<anonymous>)
		const foobarSpy = jest.spyOn( Utils, 'foobar' );

		const imported = await Utils.foobar( );

		expect( foobarSpy ).toHaveBeenCalled();
	} );
} );

```

### 🙁 Actual behavior

We're forced to use relative paths when we use jest spyOn type functionality - instead of the aliased paths from tsconfig.

### 🙂 Expected behavior

We shouldn't have to update our paths to be relative paths.

rbuckton:
This adds `configurable: true` to the property descriptors used in the `__createBinding` and `__setModuleDefualt` helpers to be more compatible with Jest's mocking mechanism.

Fixes #43081

### DIFF
--- a/src/compiler/factory/emitHelpers.ts
+++ b/src/compiler/factory/emitHelpers.ts
@@ -1165,7 +1165,7 @@ export const createBindingHelper: UnscopedEmitHelper = {
                 if (k2 === undefined) k2 = k;
                 var desc = Object.getOwnPropertyDescriptor(m, k);
                 if (!desc || ("get" in desc ? !m.__esModule : desc.writable || desc.configurable)) {
-                  desc = { enumerable: true, get: function() { return m[k]; } };
+                  desc = { enumerable: true, configurable: true, get: function() { return m[k]; } };
                 }
                 Object.defineProperty(o, k2, desc);
             }) : (function(o, m, k, k2) {
@@ -1182,7 +1182,7 @@ export const setModuleDefaultHelper: UnscopedEmitHelper = {
     priority: 1,
     text: `
             var __setModuleDefault = (this && this.__setModuleDefault) || (Object.create ? (function(o, v) {
-                Object.defineProperty(o, "default", { enumerable: true, value: v });
+                Object.defineProperty(o, "default", { enumerable: true, configurable: true, value: v });
             }) : function(o, v) {
                 o["default"] = v;
             });`,
--- a/tests/baselines/reference/ambientShorthand_reExport.js
+++ b/tests/baselines/reference/ambientShorthand_reExport.js
@@ -28,7 +28,7 @@ var __createBinding = (this && this.__createBinding) || (Object.create ? (functi
     if (k2 === undefined) k2 = k;
     var desc = Object.getOwnPropertyDescriptor(m, k);
     if (!desc || ("get" in desc ? !m.__esModule : desc.writable || desc.configurable)) {
-      desc = { enumerable: true, get: function() { return m[k]; } };
+      desc = { enumerable: true, configurable: true, get: function() { return m[k]; } };
     }
     Object.defineProperty(o, k2, desc);
 }) : (function(o, m, k, k2) {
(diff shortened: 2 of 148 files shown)


jakebailey:
Interestingly I'm pretty sure that esbuild's equivalent to this does not make them configurable, in an attempt to more closely match the ESM output (where you can't really do anything to the imported thing). Does jest break with them?

(https://github.com/aelbore/esbuild-jest/issues/26, https://github.com/evanw/esbuild/issues/1079)

weswigham (approved):
The spec says these should be immutable on a real module, but I don't *really* care, since it's not like that helps anyone but engines, and engines aren't optimizing these downlevel modules like they are real ones - but overriding these a la jest will definitely break live bindings, and it definitely doesn't work with real esm. So long as we're okay with downlevel being looser than spec... it's fine, I guess.

rbuckton:
Jest's [ESM support](https://jestjs.io/docs/ecmascript-modules) is fairly minimal. Yes, users could write code that doesn't work as an ESM module, but they're still actually running in CJS and are likely using other features that Jest supports differently in ESM vs CommonJS. 

For instance, in ESM `import { foo } from "./bar"` doesn't trigger Jest's module mocking (you have to use `await import("./bar")` instead), but `import { foo } from "./bar"` in a TS file that is downleveled to CJS works just fine.

I think that without this, users are just in an ambiguous place as to what works and what doesn't work with jest. That said, I don't have a strong preference for having this vs. closing the issue as working as intended.

sandersn:
Does this need more discussion or is it ready to merge? Specifically, @jakebailey were your concerns addressed?

jakebailey:
I don't have a principled opinion besides looking at what other tools do (e.g. [esbuild makes these non-configurable by not saying `configurable: true`](https://jakebailey.dev/esbuild-playground/#v3=PTAEAEDMEsBsFMB2BDAtvAXKAxge0TAOYB0AVgM74BQA3laKAERIAuATgJ4AKu0iL5RlgDajYHwAm8AB7EB0xgF0ANPSaRcbVMhZCm2CoyoBfKlRAQYCFOiySZc8tKoyADppagAVKEhtcqEykAI4ArvCcjADcQA)) and the general feeling that people are slowly using these helper less and less (especially with the rise of "verbatim code" and ESM). If this change seems desirable then that's fine, I'm just not sure how much it'll help people these days or if it'll actually make things more consistent. But I just don't trust myself to really know 😅

rbuckton:
Following discussion in the Design Meeting, we've decided not to take this fix as we would be inconsistent with esbuild, babel, et al. There are already workarounds for this, such as using `jest.unstable_moduleMock`, and this seems like an issue Jest would need to address in their mock implementation so as to be compatible with other transpilers, such as replacing the module object, introducing a `Proxy`, etc.

PR #57800, opened 2024-03-15
Make builder not depend on information from dts emit about really needed modules

sheetalkamat:
Earlier we use to use information from d.ts emit about all the module symbols that the file used so we could only mark subset of actual import/references etc to invalidate semantic diagnostics and d.ts emit for. 
Given that d.ts emit is changing, we cant deterministically get this information any more so we have to invalidate it for all references

This removes use of exportMap from builder and buildInfo
8d3a02f42c5f70ea323bfce181953385d7062dc9 Is just test update that fixes the incorrect information about signature (marks as version use even though it computed signature if signature and version are same)
dcdc6548b3a8c144b29dfbfa2c800672b9b1f6e1 actual change of not depending on `exportedModulesFromDeclarationEmit` and baselines that only change buildInfo to remove `exportsMap` entry
7b29ccb3acc96524877d83315a20075391dd32c1 are all baselines that got updated because of using extra set of files to invalidate.

Precursor for  #57681

### DIFF
--- a/src/compiler/builder.ts
+++ b/src/compiler/builder.ts
@@ -73,6 +73,7 @@ import {
     returnUndefined,
     sameMap,
     SemanticDiagnosticsBuilderProgram,
+    SignatureInfo,
     skipTypeChecking,
     SourceFile,
     sourceFileMayBeEmitted,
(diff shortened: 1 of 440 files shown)


jakebailey:
Just so I understand, is this cleanup a net negative? Or is it fine?

sheetalkamat:
With this change we may refresh diagnostics and d.ts emit for more files depending on if dts emit uses subset of ts file imports or all of them. But with upcoming change of not writing references, type refs etc this is kind of needed if that goes in,

jakebailey (approved):
Seems alright to me, then. Thanks for looking into this.

PR #57829, opened 2024-03-18
Don't indirect through _namespaces in tsserver/typingsInstaller

jakebailey:
A follow-up to #55326 which directly pulls from the public API project rather than indirecting through `_namespaces`, which incurs an extra copy due to esbuild's helpers.

I expect the bundle size to get larger due to the name esbuild chooses for the module, but potentially faster to load as it's not recreating a public API object.

This also likely locks off using named imports for nested modules as asked about in https://github.com/microsoft/TypeScript/pull/55326#pullrequestreview-1940018989, but I have not found a way to do those within our current import syntax.

### DIFF
--- a/src/tsserver/_namespaces/ts.ts
+++ b/src/tsserver/_namespaces/ts.ts
@@ -1,1 +1,0 @@
-export * from "../../typescript/typescript";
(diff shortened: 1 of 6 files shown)


jakebailey:

jakebailey:
Like 1% on tsserver, not exactly amazing.

jakebailey:
But, probably worth it to not have `_namespaces` anymore.

DanielRosenwasser (approved):

PR #57830, opened 2024-03-18
Update to nightly TS after type predicate inference

jakebailey:
As you can see, this lets eslint flag a few unnecessary type assertions.

Can't really take this yet as the VS Code nightly extension hasn't published since 3/14.


src/compiler/commandLineParser.ts
```diff
@@ -3512,8 +3512,7 @@ export function convertJsonOption(
                 convertJsonOption(opt.element, value, basePath, errors, propertyAssignment, valueExpression, sourceFile);
         }
         else if (!isString(opt.type)) {
-            // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
-            return convertJsonOptionOfCustomType(opt as CommandLineOptionOfCustomType, value as string, errors, valueExpression, sourceFile);
+            return convertJsonOptionOfCustomType(opt, value as string, errors, valueExpression, sourceFile);
```
jakebailey, line 3515:
```suggestion
            return convertJsonOptionOfCustomType(opt as CommandLineOptionOfCustomType, value as string, errors, valueExpression, sourceFile);
```

PR #57837, opened 2024-03-18
Avoid infinite recursion with `inferReverseMappedType`

### LINKED ISSUE #57786: Maximum call stack size exceeded with two mapped types
### 🔎 Search Terms

"Maximum call stack size exceeded", "mapped types"

- This is a crash

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.4.2#code/C4TwDgpgBAYg9nAPANQHxQLxQN4CgoFQDaA0lAJYB2UA1hCHAGZTIC6AXLAiqa6gNy4AvrlyhIUAEIBDAE4p0WPIWJkqtekxYcWvKBAAewCJQAmAZyhwARgCsIAY2BQA-FLk8SfKJ3PBZVADmgiK4jACulE7kcNTGfgoAFABu0gA24RCc8EhoAJQ4+FAOseZwaRAAdGlwgSnpmXkhoiWUflDWcpwy8tKUIIo4QoJiEH6JnbJNuEA

### 💻 Code

```ts
type Foo<V> = {
    [K in keyof V]: Foo<V[K]>;
}

type Bar<V> = {
    [K in keyof V]: V[K] extends object ? Bar<V[K]> : string;
}

function test<V>(value: Foo<V>) {
  console.log(value);
}

const bar: Bar<any> = {};

test(bar);
```

### 🙁 Actual behavior

Throws an exception
> Uncaught (in promise) RangeError: Maximum call stack size exceeded

### 🙂 Expected behavior

Either passes type check or reports a problem

_No response_

gabritto:
Fixes #57786.

For reference, the test case reported in the issue loops until stack overflow with the following function calls repeating:

```
at resolveReverseMappedTypeMembers (src\compiler\checker.ts:13790:69)
at resolveStructuredTypeMembers (src\compiler\checker.ts:14103:21)
at C:\Users\gabrielaa\code\Typescript\src\compiler\checker.ts:17091:148
at some (src\compiler\core.ts:665:21)
at removeSubtypes (src\compiler\checker.ts:17091:50)
at getUnionTypeWorker (src\compiler\checker.ts:17314:27)
at getUnionType (src\compiler\checker.ts:17286:16)
at getTypeFromInference (src\compiler\checker.ts:25353:39)
at inferReverseMappedType (src\compiler\checker.ts:25307:16)
```


gabritto:
@typescript-bot run dt
@typescript-bot test top400

gabritto:

weswigham (commented):
The less-broadly used `inferTypeForHomomorphicMappedType` has caching/circularity prevention mechanisms already, but I guess since we use this lazily to generate the types of properties of reverse mapped types (directly, in more places than just the aforementioned function), it's gotta get dedicated circularity guards. I'm just wondering if the fail case should be `unknown`, or if it should be the constraint of the `target`?

gabritto:
> The less-broadly used `inferTypeForHomomorphicMappedType` has caching/circularity prevention mechanisms already, but I guess since we use this lazily to generate the types of properties of reverse mapped types (directly, in more places than just the aforementioned function), it's gotta get dedicated circularity guards.

What's the difference between calling `inferTypeForHomomorphicMappedType` and calling `inferReverseMappedType` directly?

> I'm just wondering if the fail case should be `unknown`, or if it should be the constraint of the `target`?

I also have this same question. I'm returning `unknown` just because that's what `inferReverseMappedType` was already returning when inference failed (`return getTypeFromInference(inference) || unknownType`), but I don't know how we usually deal with failed inferences.

weswigham:
>What's the difference between calling inferTypeForHomomorphicMappedType and calling inferReverseMappedType directly?

`inferTypeForHomomorphicMappedType` is used to create the outer object type with lazily inferred properties, while `inferReverseMappedType` is used to make the types for the properties of that type on demand.

> but I don't know how we usually deal with failed inferences.

`getInferredType` usually falls back to the constraint of the type parameter. I dunno if we wanna change the reverse mapped type fallback behavior to match. Maybe that should be a separate followup PR.

src/compiler/checker.ts
```diff
@@ -25299,14 +25302,33 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
-    function inferReverseMappedType(sourceType: Type, target: MappedType, constraint: IndexType): Type {
+    function inferReverseMappedTypeWorker(sourceType: Type, target: MappedType, constraint: IndexType): Type {
         const typeParameter = getIndexedAccessType(constraint.type, getTypeParameterFromMappedType(target)) as TypeParameter;
         const templateType = getTemplateTypeFromMappedType(target);
         const inference = createInferenceInfo(typeParameter);
         inferTypes([inference], sourceType, templateType);
         return getTypeFromInference(inference) || unknownType;
     }
 
+    function inferReverseMappedType(source: Type, target: MappedType, constraint: IndexType): Type {
```
weswigham, line 25313: Should we cache the results inside this, so we guarantee deeply nested results get cached? Of the callers of this function, only `getTypeOfReverseMappedSymbol` currently caches - the other uses don't (directly) cache the result, so could, in theory, witness different results for the same set of input types at different points. You can probably only easily trigger such a behavior with a reverse mapped type over an array...? Maybe?

weswigham (approved):
This seems fine - definitely an improvement. Just wondering if we should also be doing caching within this function, rather than relying on callers of it to cache the results somehow.

src/compiler/checker.ts
gabritto: I was adding the cache to `inferReverseMappedType` and looking at `inferTypeForHomomorphicMappedType`. I'm now wondering: could I do away with the recursion detection in `inferTypeForHomomorphicMappedType` (that uses `homomorphicMappedTypeInferenceStack`)?

gabritto: I'm pretty sure that we can, however this once again begs the question of what to return when we find a circularity. `inferTypeForHomomorphicMappedType` returns `undefined` in this case. I'll have to think some more about this, but I'm planning on removing that circularity detection code in `inferTypeForHomomorphicMappedType`.

sandersn:
I can't tell from the discussion if this is ready to merge. Do we want to investigate better caching before taking this? Do we need to figure out what happens in case of circularity?

Andarist:
> getInferredType usually falls back to the constraint of the type parameter. I dunno if we wanna change the reverse mapped type fallback behavior to match. Maybe that should be a separate followup PR.

I think that PR might already exist: https://github.com/microsoft/TypeScript/pull/56300 ;p I need to take another look at it to see what is left to do there though

gabritto:
@typescript-bot test it

gabritto:
Latest changes: I got rid of the circularity detection that existed in `inferTypeForHomomorphicMappedType`, as every circularity of that function is also a circularity of `inferReverseMappedType`. To do that but keep the existing behavior where `inferTypeForHomomorphicMappedType` returns `undefined` in cases of circularity, I changed `inferReverseMappedType` to also return `undefined` when a circularity is detected, and so I had to update the callers of `inferReverseMappedType` to handle the `undefined`.
We may be able to get rid of those `undefined`s in the future in something like #56300, where when inference fails due to circularity, we default to the constraint of the type.

weswigham (approved):

Issue #57827, opened 2024-03-18
`Extract` not narrowing type union in mapped type passed as generic


mikeauclair:
### 🔎 Search Terms

Extract, type union

### 🕗 Version & Regression Information

- This changed between versions 5.33 and 5.4

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.4.2#code/C4TwDgpgBAysBOBLAdgcwNIRAZwDwBUA+KAXigFEAPBAQwGNhcBrLAewDMoBvAWACgogqAG0AClBRQWIDlHxQa2OWIC6UCNQjIAJkuwIUqKAH4oogFxRkEAG4R4KyzWQh+AXwA0UfUjSF+-HSsyPpQAEYANqysYKRQBB6EABQAXpZwvhhYeEQAlKTEvAJQQSHAUOyWPoZxKe4BfCjA9uz00Ira3PyCAI4A7hBVBmj1fPwREOU06cNZOLgd-nxAA

### 💻 Code

```ts
type StringKeys<T> = Extract<keyof {
    [P in keyof T as T[P] extends string ? P: never]: any
}, string>

const bloop = <T,>(z: StringKeys<T>) => {
  const f: string = z
}

interface asd {
  qwe: string
}

let a: StringKeys<asd>

```

### 🙁 Actual behavior

`z` is not assignable to `string` in the `bloop` function

### 🙂 Expected behavior

`z` should be assignable to `string`, because we extract string-extending keys

### Additional information about the issue

This was working just fine on 5.3.3, and continues to work when the generic is concretized (hence the `let` at the bottom of the example)

Andarist:
Conditional types are tricky as they often stay deferred in generic contexts. This version could somewhat easily work though (and it doesn't):
```t
type StringKeys<T> = keyof {
  [P in keyof T as T[P] extends string ? P & string : never]: any;
};

const bloop = <T,>(z: StringKeys<T>) => {
  const f: string = z; // errors but shouldnt
};
```

As a workaround, you can use this version:
```ts
type StringKeys<T> = keyof {
  [P in keyof T as T[P] extends string ? P : never]: any;
} &
  string;

const bloop = <T,>(z: StringKeys<T>) => {
  const f: string = z;
};
```

mikeauclair:
That workaround works in my real codebase, so I'll cut over to that approach for now - is this a matter of the intersection (instead of `Extract`) causing non-deferred evaluation?

[...]

PR #57838, opened 2024-03-18
Fixed an issue with apparent mapped type keys

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/57827

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -22486,7 +22486,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 /*stringsOnly*/ false,
                 t => void mappedKeys.push(instantiateType(nameType, appendTypeMapping(targetType.mapper, getTypeParameterFromMappedType(targetType), t))),
             );
-            return getUnionType(mappedKeys);
+            return mappedKeys.length ? getUnionType(mappedKeys) : stringNumberSymbolType;
         }
 
         function structuredTypeRelatedToWorker(source: Type, target: Type, reportErrors: boolean, intersectionState: IntersectionState, saveErrorInfo: ReturnType<typeof captureErrorCalculationState>): Ternary {
--- a/tests/cases/compiler/keyRemappingKeyofResult.ts
+++ b/tests/cases/compiler/keyRemappingKeyofResult.ts
@@ -1,3 +1,4 @@
+// @strict: true, false
 // @target: es6
 const sym = Symbol("")
 type Orig = { [k: string]: any, str: any, [sym]: any }
@@ -67,6 +68,19 @@ function g<T>() {
     let x: Oops;
     x = sym;
     x = "str";
+}
+
+// https://github.com/microsoft/TypeScript/issues/57827
+
+type StringKeys<T> = Extract<
+  keyof {
+    [P in keyof T as T[P] extends string ? P : never]: any;
+  },
+  string
+>;
+
+function test_57827<T>(z: StringKeys<T>) {
+  const f: string = z;
 }
 
 export {};
--- a/tests/baselines/reference/keyRemappingKeyofResult(strict=false).js
+++ b/tests/baselines/reference/keyRemappingKeyofResult(strict=false).js
@@ -71,6 +71,19 @@ function g<T>() {
     x = "str";
 }
 
+// https://github.com/microsoft/TypeScript/issues/57827
+
+type StringKeys<T> = Extract<
+  keyof {
+    [P in keyof T as T[P] extends string ? P : never]: any;
+  },
+  string
+>;
+
+function test_57827<T>(z: StringKeys<T>) {
+  const f: string = z;
+}
+
 export {};
 
 //// [keyRemappingKeyofResult.js]
@@ -97,5 +110,8 @@ function g() {
     let x;
     x = sym;
     x = "str";
+}
+function test_57827(z) {
+    const f = z;
 }
 export {};
--- a/tests/baselines/reference/keyRemappingKeyofResult(strict=false).symbols
+++ b/tests/baselines/reference/keyRemappingKeyofResult(strict=false).symbols
@@ -192,4 +192,35 @@ function g<T>() {
 >x : Symbol(x, Decl(keyRemappingKeyofResult.ts, 65, 7))
 }
 
+// https://github.com/microsoft/TypeScript/issues/57827
+
+type StringKeys<T> = Extract<
+>StringKeys : Symbol(StringKeys, Decl(keyRemappingKeyofResult.ts, 68, 1))
+>T : Symbol(T, Decl(keyRemappingKeyofResult.ts, 72, 16))
+>Extract : Symbol(Extract, Decl(lib.es5.d.ts, --, --))
+
+  keyof {
+    [P in keyof T as T[P] extends string ? P : never]: any;
+>P : Symbol(P, Decl(keyRemappingKeyofResult.ts, 74, 5))
+>T : Symbol(T, Decl(keyRemappingKeyofResult.ts, 72, 16))
+>T : Symbol(T, Decl(keyRemappingKeyofResult.ts, 72, 16))
+>P : Symbol(P, Decl(keyRemappingKeyofResult.ts, 74, 5))
+>P : Symbol(P, Decl(keyRemappingKeyofResult.ts, 74, 5))
+
+  },
+  string
+>;
+
+function test_57827<T>(z: StringKeys<T>) {
+>test_57827 : Symbol(test_57827, Decl(keyRemappingKeyofResult.ts, 77, 2))
+>T : Symbol(T, Decl(keyRemappingKeyofResult.ts, 79, 20))
+>z : Symbol(z, Decl(keyRemappingKeyofResult.ts, 79, 23))
+>StringKeys : Symbol(StringKeys, Decl(keyRemappingKeyofResult.ts, 68, 1))
+>T : Symbol(T, Decl(keyRemappingKeyofResult.ts, 79, 20))
+
+  const f: string = z;
+>f : Symbol(f, Decl(keyRemappingKeyofResult.ts, 80, 7))
+>z : Symbol(z, Decl(keyRemappingKeyofResult.ts, 79, 23))
+}
+
 export {};
--- a/tests/baselines/reference/keyRemappingKeyofResult(strict=false).types
+++ b/tests/baselines/reference/keyRemappingKeyofResult(strict=false).types
@@ -172,4 +172,24 @@ function g<T>() {
 >"str" : "str"
 }
 
+// https://github.com/microsoft/TypeScript/issues/57827
+
+type StringKeys<T> = Extract<
+>StringKeys : StringKeys<T>
+
+  keyof {
+    [P in keyof T as T[P] extends string ? P : never]: any;
+  },
+  string
+>;
+
+function test_57827<T>(z: StringKeys<T>) {
+>test_57827 : <T>(z: StringKeys<T>) => void
+>z : Extract<keyof { [P in keyof T as T[P] extends string ? P : never]: any; }, string>
+
+  const f: string = z;
+>f : string
+>z : Extract<keyof { [P in keyof T as T[P] extends string ? P : never]: any; }, string>
+}
+
 export {};
--- a/tests/baselines/reference/keyRemappingKeyofResult(strict=true).js
+++ b/tests/baselines/reference/keyRemappingKeyofResult(strict=true).js
@@ -1,0 +1,117 @@
+//// [tests/cases/compiler/keyRemappingKeyofResult.ts] ////
+
+//// [keyRemappingKeyofResult.ts]
+const sym = Symbol("")
+type Orig = { [k: string]: any, str: any, [sym]: any }
+
+type Okay = Exclude<keyof Orig, never>
+// type Okay = string | number | typeof sym
+
+type Remapped = { [K in keyof Orig as {} extends Record<K, any> ? never : K]: any }
+/* type Remapped = {
+    str: any;
+    [sym]: any;
+} */
+// no string index signature, right?
+
+type Oops = Exclude<keyof Remapped, never>
+declare let x: Oops;
+x = sym;
+x = "str";
+// type Oops = typeof sym <-- what happened to "str"?
+
+// equivalently, with an unresolved generic (no `exclude` shenanigans, since conditions won't execute):
+function f<T>() {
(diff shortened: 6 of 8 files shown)


src/compiler/checker.ts
```diff
@@ -22486,7 +22486,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 /*stringsOnly*/ false,
                 t => void mappedKeys.push(instantiateType(nameType, appendTypeMapping(targetType.mapper, getTypeParameterFromMappedType(targetType), t))),
             );
-            return getUnionType(mappedKeys);
+            return mappedKeys.length ? getUnionType(mappedKeys) : stringNumberSymbolType;
```
Andarist, line 22489: On second thought, this likely should only happen when obtaining apparent mapped type keys on the source side of the relation check. I'll think about the test case for the target side that fails with this.

The problem here is that when `modifiersType` is `unknown` (or `{}`) then `keyof modifiersType` is `never`. Since `never` is assignable to everything, the checker assumed that `keyof { [P in keyof T as T[P] extends string ? P: never]: any }` is assignable to `string`. 

Then based on that information it managed to simplify the instantiated substitution type of `Extract` to just `keyof { [P in keyof T as T[P] extends string ? P: never]: any }` (its `newBaseType`). Then in turn `getNarrowableTypeForReference` concluded that it should substitute constraints. Those were substituted with `stringNumberSymbolType` (for the `keyof ...`) and that's not assignable to `string`.

Before https://github.com/microsoft/TypeScript/pull/56742 , it didn't mistakenly assume that `keyof { [P in keyof T as T[P] extends string ? P: never]: any }` is assignable to `string` so the substitution type was instantiated as `(keyof { [P in keyof T as T[P] extends string ? P: never]: any }) & string` and that is assignable to a string.

LukeAbby:
Hi! I appreciate that reviewers can be very busy but I discovered this PR when bisecting my own issue with a bit of an absurd [reproduction](https://www.typescriptlang.org/play/?noErrorTruncation=true#code/CYUwxgNghgTiAEYD2A7AzgF3lMGkwC54BJFFEGUgSwyqgioC8RgAVATwAcQAeAb3hp2mEAFsiHbgBEoGKADEqICMHgBfAHwBuAFA48MAHRCRo3ToxcEpcpRQ06DZmys9Fy4Gg3wAvPD468PAA9MHwAEognNBgVCgA5vAABnEAZhQk9hiSIEnwAO40ABbJUCjsealUAB4gaPAYRQgUMPiGgfAA2gDSIOzwcfAA1n1IqfDuKmgAukSTnj190-Ag1RggKJ7wMnLzPGkZ1NlWGh1BAPzY5WfwRGXsumrmltzwAMpUotEgALIU8bxWLAARgADTwADyADcWlRQN4-BDRDQeECYCDwSN2GNITCYDA4SBvAAyXGw0DmUCQWAIaloeoAfVI6xgKHoOR2UB+SFAEB4bzATVEUG8ASCcVsR0cTBYOQAomsNp4iDYKFL6DKXNx+YKxCKVorNvUAERQY3wS73W5XB46DqheAAQSwEBAUEw8AArPBOK1uDBaHVsHB4HAAI4AVyocGA7XFKBZbIgHNkXJ5ygACn6AIx3cq6eOJ9lWTnc3lZpCcABMedthYoSZTcjLmb9AGZawWBgmG8XpKmWxAK5wACydjpxIvJksD9NDv2e8dqO0vBCRURIGFSChUGHAHgAIXdIFYRSoaHBR7Qvznl+PnPB24Je85CI6SJRH4wh+Pp-PmNGcYrxATl4AAH2GQDth3F9Uw0ADsSA49BxJeBgMHZ4rHgP50RADNYFoehUXBABVBF4FYeBSTFLpum7SDEPgEjsHqBUMBgfQeG6cF8IDRwcIBXphGIpiNA0aZziXTDXkiSNoxYIS0FRciaJ6eisRxVhpgAWkkiiDXWI1-FDN1gFQCB+k6Bl6O6WYbXUC14DoohyDxR5Og08YtOkhBeMIiABJARSRLI3x-A6NTBk8pidL0ujVkMrZZKjGNgrIm5Lniw0kpAOTUr6JTWFOIISsch0-3qZBNgcVBsAgfIoGEFYoXoCNZCDPAGhgCMEAAIwjLAaEEWgIAgQQQDEepwwjAkWDjUqSqtBKlXqdiepuUrLlcigNpKog6IdZ14Fdd0sFQBAcUaOoECSbaYCSepRAjD1epu7okkMCiz3qc6CndRAiiQa8UHgV6wCgZ6EBoAByeo7q6nrQYG+AUCQeBhQwFkCjPQV4F+7EIx9KgwCGBomhWfF8AGVamjgebSpckA8RuRm3J0NQPKgkjpkpcBoBDKqPQZBlgM5IgI3sSMECEUReqQCBdCpfnaVQIWGSfXcWDF+AJaoKXBHYWX5fMScKFSHBrHQOQUDAXhRdTR8YK1uDwqCKyRfvVM7PtuQu3djXYLkOyA+d332ZXLCsyiGkQ+ATkKp-a8KrfIJgIqgyVsyTAylt-YUHSGA0M9uRwQOQvY9fDLsP+PCCMcHhdqCD4vldQKG4Whbyu+5Ivzyc8UfAOo0FgfpQDQTgaAQOANyhOJElRhi0HpjvIWRb9G4W9dNxAp39zT77wQloZUfyFA72vB9oOfUORVBDfSru+-4Hgp+fagDeX5Xq-Nbj1NduKvaA82Z2iVjSRA0B6QURnM2Oc7d3i6mFHfVOxcoBIO-oHVB9owjpwShxBoWFOCwCgKIEAWN+7TXkqofqg0sBmSDKjLARQoAwgaGjV6Otryxg6CRFAkNVB+HuHfbwy0jJMh7KyPsIFZy8h1EKfUqlhZvzsj8KAnBuBajtig7QEVhYVy9kQFRajZSuD0XIbRHRfTRzgKYqARAo6EOsbveO30eCNHPOYoIq5SxzmHLmG0XYvHSNbJWGs-iOiBJgeWdsS47QOlIAwewCQyb93oGgNGVRag0yhvSHq7RVwUXVE4YxrwfAdAgk2NMMiBDVHYIwVmGQ1DgkETaAB5ToGVOUPwbAvUwD1MLo0m0TTyjmK7v3fIHEjGF3ydvQuSROTzD7lNXKKUWDwFSFTNJJCTLulQO0Ok9QKnzAzkZeZSgVCokKZqHIoplw6FAQLCB9RTkeB4Jc5wOQwr3FFB0SGbyWB+L+Ro8CKMIyywoFoEIYQ3gTVEKtNGTCWGdXYVAHW9hdk-OvIC0JgKciPAjq8Qx6icjKTCjRUZ9R+5oE+NEKgVRVmdSuj6JAk48bjFenPbAoMhpQF6mgGaXC3Z0SilBSi-0lrZXqHRLaTMMh9LsvcR4uggA) that was erroring with a much more opaque error "Type instantiation is excessively deep and possibly infinite." so I was unable to find any workarounds before the process of bisecting.

I'd really like to see this PR merged because when I built it locally it makes my code work! I wouldn't want to ask this of anyone else without being willing to do it myself so I did also review it and I can't find anything objectionable, though of course my opinion doesn't weigh as much given I'm not a maintainer.

gabritto:
@typescript-bot pack this

gabritto:
@typescript-bot test it

src/compiler/checker.ts
```diff
@@ -23277,7 +23277,18 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     // allow assignments of index types of identical (or similar enough) mapped types.
                     // eg, `keyof {[X in keyof A]: Obj[X]}` should be assignable to `keyof {[Y in keyof A]: Tup[Y]}` because both map over the same set of keys (`keyof A`).
                     // Without this source-side breakdown, a `keyof {[X in keyof A]: Obj[X]}` style type won't be assignable to anything except itself, which is much too strict.
-                    const sourceMappedKeys = nameType && isMappedTypeWithKeyofConstraintDeclaration(mappedType) ? getApparentMappedTypeKeys(nameType, mappedType) : (nameType || getConstraintTypeFromMappedType(mappedType));
+                    let sourceMappedKeys: Type;
+                    if (nameType && isMappedTypeWithKeyofConstraintDeclaration(mappedType)) {
+                        sourceMappedKeys = getApparentMappedTypeKeys(nameType, mappedType);
+                        if (sourceMappedKeys.flags & TypeFlags.Never) {
```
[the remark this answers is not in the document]
Andarist: > keyof unknown is undefined.

If we think about `keyof T` as a set of types that can be safely used to access properties of `T` in `T[keyof T]` then `never` makes sense here. `never` is the only type that can be safely used as `keyof T` after we instantiate `T` as `unknown`. 

> the fix is looking at the result of getApparentMappedTypeKeys. I'd expect it to be inside getApparentMappedTypeKeys. Isn't there a case where getApparentMappedTypeKeys is never and sourceMappedKeys should stay as never?

That's what I started with but mapping `never` to `PropertyKey` would break the other call site where `getApparentMappedTypeKeys` is used. In there, it's used to get the *target* keys. So this would remove the error below:

```ts
type StringKeys2<T> = keyof {
  [P in keyof T as T[P] extends string ? P : never]: any;
};
function h<T>(z: StringKeys2<T>) {
  z = "foo"; // error
}
```

tests/cases/compiler/keyRemappingKeyofResult.ts
```diff
@@ -55,18 +56,40 @@ function g<T>() {
...
+  const f: string = z;
+  z = "foo"; // error
+}
+
+type StringKeys2<T> = keyof {
+  [P in keyof T as T[P] extends string ? P : never]: any;
+};
+
+function h<T>(z: StringKeys2<T>) {
+  z = "foo"; // error
```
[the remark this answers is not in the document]
Andarist: Shouldn't this be OK though? Only string keys are extracted by `StringKeys2` so `z` has a string constraint~ and is assignable to `string`. This is how it works today ([TS playground](https://www.typescriptlang.org/play/?ts=5.8.0-dev.20241130#code/C4TwDgpgBAysBOBLAdgcwNIRAZwEwB4AVAPigF4oBrLAewDMoBvAKCigG0AFKFK2hwlACG2KIS4BdKBAAewCMgAmo7AhSooAfijcAXFGQQAbhHgT9Q5CADczAL63mdAK7IAxsEQ1kUABZFiAAoAL304JDRMHAISAEomVihg8igAIjoaGlTrKAB6XOl4eBp4RLdvVSg6fVUIjQpgnPyoGkp7IA)) and this PR doesn't change that

src/compiler/checker.ts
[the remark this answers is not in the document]
gabritto: This all has me thinking that the original code that uses `getApparentMappedTypeKeys` on the source side is just... wrong?

gabritto: I still think it makes more sense to put this change into `getApparentMappedTypeKeys` and update that function with a parameter for whether we're calling it on a source or target type, since it's in theory possible that we get `never` as a result not because of `keyof unknown`, but I couldn't come up with such an example (it's already hard to hit this). But I think in general this fix is ok.

Andarist: @gabritto done

gabritto:
@typescript-bot test it

src/compiler/checker.ts
```diff
@@ -23464,16 +23464,22 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
             forEachMappedTypePropertyKeyTypeAndIndexSignatureKeyType(
                 modifiersType,
                 TypeFlags.StringOrNumberLiteralOrUnique,
                 /*stringsOnly*/ false,
-                t => void mappedKeys.push(instantiateType(nameType, appendTypeMapping(targetType.mapper, getTypeParameterFromMappedType(targetType), t))),
+                t => void mappedKeys.push(instantiateType(nameType, appendTypeMapping(mappedType.mapper, getTypeParameterFromMappedType(mappedType), t))),
             );
-            return getUnionType(mappedKeys);
+            const apparentKeys = getUnionType(mappedKeys);
+            if (forSource && apparentKeys.flags & TypeFlags.Never) {
```
gabritto: Sorry if I wasn't completely clear, I think moving this check to this function allows us to simply check if `modifiersType` is `unknown`, and then we return `string | number | symbol` instead of `never`. To me this would make more clear where the special case is coming from and when it should be applied. But maybe this should actually be applied whenever `never` is there result, regardless of what the modifiers type was? That's what I'm confused about still.

What I also find confusing about the existing use of `getApparentMappedTypeKeys` is that, if we're using it to compare a source type with a target type, I would imagine we want to be conservative and obtain the *largest* possible type the source type could be. But the logic of `getApparentMappedTypeKeys` seems to compute the *smallest* possible type the source type could be: if we have some generic mapped type M, its apparent type will be, say, `M1`, and `M <: M1` (`M``` is a subtype of `M1`). We then compute the `keyof M1`, but I think `keyof` is contravariant, and so `keyof M` :> keyof M1`, so we're effectively approximating `keyof M` by a subtype of it, whereas to be conservative on the side of a source type, we'd need to approximate `keyof M` by a supertype. @weswigham is my thinking wrong here?

If that all made sense, then this PR fixes the `keyof` approximation by inverting `never` (smallest type for a key) and `string | number | symbol` (largest type for a key), but then I still worry the cases *in between* could be wrong.

Andarist: You raise a good point here, I'll research this further and come back with some thoughts

src/compiler/checker.ts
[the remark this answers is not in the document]
Andarist: Yes. I think it's wrong:
```ts
type AcceptString<T extends string> = T;
type FilterByLiteralValue<T, V> = { [P in keyof T as T[P] extends V ? P : never]: any };

function test<T extends { a: 1; b: 2 }>(k: keyof FilterByLiteralValue<T, 1>) {
  const s: string = k; // errors correctly
  type Result = AcceptString<typeof k> // doesn't error but it should
}
```

src/compiler/checker.ts
Andarist: I think it's wrong:
```ts
type AcceptString<T extends string> = T;
type FilterByLiteralValue<T, V> = { [P in keyof T as T[P] extends V ? P : never]: any };

function test<T extends { a: 1; b: 2 }>(k: keyof FilterByLiteralValue<T, 1>) {
  const s: string = k; // errors correctly
  type Result = AcceptString<typeof k> // doesn't error but it should
}
```

So yes, you are right - obtaining apparent keys on the source side isn't the right approach at all.

PR #57839, opened 2024-03-19
fix54092: return replacement ranges for completions on unclosed strings

### LINKED ISSUE #54092: Completion for string property doesn't include replacementSpan if string is unclosed
monaco-editor version: 0.10.1
Browser: All
OS: All
Steps or JS usage snippet reproducing the issue:
Completions of fields with dot in their name is somewhat broken.
How to reproduce:

1. You have the following object: `let x = { "foo.bar": "test" };`
2. Typing in `x["fo]` and pressing ctrl+space, and then selecting `foo.bar` generates the expected result
3. Typing in `x["foo.b` and pressing ctrl+space, and selecting `foo.bar` generates `x["foo.foo.bar`.  

Basically if there is a dot in the field name the issue appears. It only happens in monaco editor. VScode does not have the same bug.

I would gladly fix this if you can point me in the right direction (monaco-typescript most likely, in the completions provider, but i don't really see what the issue is in the code).

iisaduan:
fixes #54092.

Replacement ranges for string completions will be returned unless the closed string is empty (a single unclosed quote).

### DIFF
--- a/tests/cases/fourslash/stringCompletionsUnterminated.ts
+++ b/tests/cases/fourslash/stringCompletionsUnterminated.ts
@@ -1,12 +1,28 @@
 ///<reference path="fourslash.ts" />
 
-//// const a = { "foo.bar": 1}
+// @Filename: file1.ts
+//// const a = { "foo.bar": 1 }
 //// a["[|foo.b/*0*/|]
 //// a["[|foo.b           /*1*/|]
 //// a[    "[|foo.b/*2*/|]
 //// a["[|foo.b/*3*/|]"
 //// a["[|foo./*4*/b|]
+//// a['[|foo./*5*/b|]
+//// a[`[|foo./*6*/b|]
 
+// @Filename: file2.ts
+//// const a = { "foo.bar": 1 }
+//// a[`/*7*/
+
+// @Filename: file3.ts
+//// const a = { "foo.bar": 1 }
+//// a["/*8*/
+
+// @Filename: file4.ts
+//// const a = { "foo.bar": 1 }
+//// a['/*9*/
+
+// tests 7-9 should return no replacementSpan
 for (let i = 0; i < test.markers().length; i++) {
     verify.completions({
         marker: test.markers()[i],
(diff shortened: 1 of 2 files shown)


tests/cases/fourslash/stringCompletionsUnterminated.ts
```diff
@@ -0,0 +1,36 @@
...
+
+// tests 7-9 should return no replacementSpan
+for (let i = 0; i < test.markers().length; i++) {
+    verify.completions({
+        marker: test.markers()[i],
+        includes: [{
+            "name": "foo.bar",
+            "kind": "property",
+            "kindModifiers": "",
+            "replacementSpan": test.ranges()[i],
```
DanielRosenwasser, line 33: Nit: I believe calling these over and over creates a copy.

```suggestion
const markers = test.markers();
const ranges = test.ranges();

for (let i = 0; i < markers.length; i++) {
    verify.completions({
        marker: markers[i],
        includes: [{
            "name": "foo.bar",
            "kind": "property",
            "kindModifiers": "",
            "replacementSpan": ranges[i],
```

tests/cases/fourslash/stringCompletionsUnterminated.ts
```diff
@@ -0,0 +1,36 @@
...
+
+// @Filename: file1.ts
+//// const a = { "foo.bar": 1 }
+//// a["[|foo.b/*0*/|]
+//// a["[|foo.b           /*1*/|]
+//// a[    "[|foo.b/*2*/|]
+//// a["[|foo.b/*3*/|]"
+//// a["[|foo./*4*/b|]
+//// a['[|foo./*5*/b|]
+//// a[`[|foo./*6*/b|]
```
DanielRosenwasser, line 11: Can you add test cases for when you have `foo.` (no contents following the dot) and `fo` (no dot or following contents)?

iisaduan: added! I also tested manually in vscode to see that the behavior didn't change for the previously working cases even though it does return a `replacementSpan` now.

gabritto (approved):

Issue #33143, opened 2019-08-29
Different instances of the same error give different error messages


AnyhowStep:
**TypeScript Version:**  3.5.1

**Search Terms:**

**Code**

```ts
interface ExprData {
  mapper : () => unknown,
  blah : string,
}
interface Expr<DataT extends ExprData> {
  mapper : DataT["mapper"],
  blah : DataT["blah"],

  extra : number,
}

declare function where (
  callback : () => Expr<{
    mapper : () => boolean,
    blah : string,
  }>
) : void;

declare function getInvalidWhere () : Expr<{
  mapper : () => (boolean|null),
  blah : string,
}>;

/*
Type 'Expr<{ mapper: () => boolean | null; blah: string; }>' is not assignable to type 'Expr<{ mapper: () => boolean; blah: string; }>'.
  Type '{ mapper: () => boolean | null; blah: string; }' is not assignable to type '{ mapper: () => boolean; blah: string; }'.
    Types of property 'mapper' are incompatible.
      Type '() => boolean | null' is not assignable to type '() => boolean'.
        Type 'boolean | null' is not assignable to type 'boolean'.
          Type 'null' is not assignable to type 'boolean'.
*/
where(() => getInvalidWhere())
/*
  Expected: Same error as above,
  Actual:
    Type 'Expr<{ mapper: () => boolean | null; blah: string; }>' is not assignable to type
    'Expr<{ mapper: () => boolean; blah: string; }>'.
*/
where(() => getInvalidWhere())
```

**Expected behavior:**

+ Both errors should have the same error message, if they're in the same file.

**Actual behavior:**

**Playground Link:** 

[...]

Issue #3276, opened 2015-05-27
Diagnostic elaboration is not consistently reported in language service


DanielRosenwasser:
Adapted from #3275:

``` TypeScript
class GenericThingamabob<T> {
    constructor(private entity: T) {}
    add(item: T) { }
}

class CouponInfo {
    private couponTag: {};
}

class Snake {
    private snakeTag: {};
}

var blah = new GenericThingamabob(new CouponInfo());

blah.add(new Snake());
//       ~~~~~~~~~~~
var x: CouponInfo = new Snake();
//  ~
```

On the first error, I get

```
Argument of type 'Snake' is not assignable to parameter of type 'CouponInfo'.
  Property 'couponTag' is missing in type 'Snake'
```

On the second, I get

```
Type 'Snake' is not assignable to type 'CouponInfo'.
```

Note that if you turned `blah.add(new Snake());` into a variable assignment like `var a: CouponInfo = new Snake()`, you'll get the same issue.

Andarist:
There is a complex interplay here between `headMessage`, `overrideNextErrorInfo `, and perhaps also `lastSkippedInfo`.

Calls coming from `checkVariableLikeDeclaration` don't have any `headMessage`. Based on that an early return might be chosen in `reportErrorResults` based on `!headMessage && maybeSuppress` (this early return also assigns to `lastSkippedInfo`). This basically happens when we only test `var x: CouponInfo = new Snake();` and in such a case the reported errors looks like this:
> Property 'couponTag' is missing in type 'Snake' but required in type 'CouponInfo'.(2741)

It's worth noting that in this scenario the default head message that would be computed ("Type 'Snake' is not assignable to type 'CouponInfo'.(2322)") is not reported exactly because it was skipped.

The elaboration is reported here though based on the `reportUnmatchedProperty` call that happened when checking this variable declaration.

When we put both a function call and a variable declaration in the file, like this
```ts
blah.add(new Snake());
var x: CouponInfo = new Snake();
```

Then this call never~ skips reporting the `headMessage` part because it comes with its own `headMessage`. The elaboration part is reported together with it because `reportUnmatchedProperty` gets called when checking that function call.

[...]

PR #57842, opened 2024-03-19
Region-based semantic diagnostics

### LINKED ISSUE #57393: Region-Prioritized Error Checking for Editors
One of the difficulties in working in larger files is that error updates have per-file granularity. This means that even if a viewport is focused on a region of 60 lines of code, the entire file must be type-checked. For a file that's more than 50,000 lines like TypeScript's `checker.ts`, it can be painfully slow to wait for errors to go away after typing something.

One idea might be to enrich TypeScript's existing errors requests with regions of interest. By specifying the viewport of the editor as a reigon, TypeScript can first prioritize gathering errors across statements that are visible to a user, and inform an editor of any diagnostic/error spans discovered in that region. Any prior diagnostics in that region can be considered invalidated, and a subsequent response with *all* the diagnostics can be returned afterwards.

# Implementation

Right now TypeScript has a `GeterrRequestArgs` that looks like this:

```ts
/**
 * Arguments for geterr messages.
 */
export interface GeterrRequestArgs {
    /**
     * List of file names for which to compute compiler errors.
     * The files will be checked in list order.
     */
    files: string[];

    /**
     * Delay in milliseconds to wait before starting to compute
     * errors for the files in the file list
     */
    delay: number;
}
```

Instead of `files` being `string[]`, we could switch over to each element being `string | FileSpan` or something similar.

When TypeScript receives a span, it can find a more appropriate span to start/end checking on. Errors are typically emitted as events that are defined through these types:

```ts
export interface DiagnosticEventBody {
    /**
     * The file for which diagnostic information is reported.
     */
    file: string;

    /**
     * An array of diagnostic information items.
     */
    diagnostics: Diagnostic[];
}

export type DiagnosticEventKind = "semanticDiag" | "syntaxDiag" | "suggestionDiag";

/**
 * Event message for DiagnosticEventKind event types.
 * These events provide syntactic and semantic errors for a file.
 */
export interface DiagnosticEvent extends Event {
    body?: DiagnosticEventBody;
    event: DiagnosticEventKind;
}
```

and these events types can be adjusted to indicate the following information:

* whether the event is a region-based diagnostic event, and
* what the adjusted region for a diagnostic request was

This information ***MUST*** be provided so that editors

* do not clear out *all* the diagnostics
* know what to adjust in case the user scrolls, or TypeScript has adjusted the span for whatever reason

# Risks

There are some risks associated with this approach.

## Checking Inconsistencies

First, out-of-order type-checking is already something that can cause inconsistencies between `tsc` and the editor. This could get worse. However, we removed the separate diagnostics-producing type-checker from the one used by the rest of the language service 2 years ago (#36747), so completions, quick info, etc. can already cause this. Since then, the problems have mostly smoothed out over time.

The other concern is that diagnostics in a region are not always caused by the code within that same region. Diagnostics can be placed wherever they choose. The risk here is that if type-checking the code within a viewport does *not* create the currently-visible errors, then a user will end up with flashing diagnostics as they type. Basically:

1. A user types some code
2. TypeScript comes back with diagnostics for the viewport
3. There are no diagnostics due to the current viewport, so the editor clears out all the diagnostics
4. TypeScript then comes back with diagnostics *for the entire file*, and the errors re-appear

This could get aggravating.

For the most part, we do not issue errors in such a non-local fashion, and strive to keep most errors close to the relevant code. The exception to this might be conflicting declarations that are duplicates. Otherwise, most non-local diagnostics are *related spans*. If there are any other errors that are not co-located with their code, we should probably try to relocate them to improve the experience.

## Chattiness

In some cases, it may be unnecessary to check only a specific region. For some files, we might be better off just checking the entire file. This is why TypeScript should reserve the right to decide on an appropriate span, or to even report diagnostics only for a specific span.

gabritto:
Fixes #57393.
Counterpart in vscode: https://github.com/microsoft/vscode/pull/208713.

## Existing behavior

Currently, the way vscode and tsserver communicate about diagnostics works like this:

1. vscode sends a `geterr` request to tsserver with a list of file names, corresponding to the typescript files open in editor, for which diagnostics should be computed.
2. tsserver receives the `geterr` request, and immediately sends back a response that is basically an "ack" and doesn't contain diagnostics.
3. tsserver starts computing diagnostics. For each file, it computes syntax diagnostics (generated by parser), then sends an event to vscode with the syntax diagnostics. Then tsserver computes semantic diagnostics (generated by binder, checker, and program), and again sends an event with those diagnostics to vscode. Lastly, tsserver computes suggestion diagnostics and sends them to vscode.

## Idea

Computing semantic diagnostics, in particular checker diagnostics, for the whole file can be slow, especially for large files like `checker.ts`. So the idea here is that, before we type check the entire file, we can type check just the parts of the file that are visible, send those requests to the editor so that they are updated, and then proceed to type check the entire file.

## Feature

The change I've implemented alters the diagnostics dynamic between vscode and tsserver in the following way:

1. vscode sends a `geterr` request to tsserver with a list whose elements are either a file name *or* a file name and ranges. The file names correspond to the typescript files open in editor, and if the ranges are specified, they correspond to the parts of the file that are visible in the editor.
2. Step 2 stays the same: tsserver receives the `geterr` request, and immediately sends back a response.
3. For each file, tsserver computes syntax diagnostics, then region semantic diagnostics (which I'll define below), then normal, whole-file semantic diagnostics, then suggestion diagnostics, each time sending the diagnostics back to vscode through an event.
 
## Region semantic diagnostics

A region semantic check is similar to a normal semantic check, but instead of type checking the whole file, we type check only the nodes that are roughly in the ranges specified by the `geterr` request.
So some semantic checks remain the same. Binder diagnostics are computed for the whole file, and program diagnostics like resolution errors are also done as normal. The big difference is in `checker.ts`, where instead of calling `checkSourceFileWorker` to check the whole file, we now call `checkSourceFileNodesWorker`, which does 

Some types of diagnostics are explicitly not computed in this region semantic check, because they require the whole file to be checked to produce accurate diagnostics. For instance, we don't check for unused identifiers, and declaration diagnostics are also not computed.

Note: according to Wesley, "the only reason some whole file export/duplicate checks require whole-file checking is because they rely on some traversal-collected data that... we could probably move up into the binder in the future with some work".

## Editor UX

On the editor side, what happens is that as the editor receives back the different kinds of diagnostics, it updates the diagnostics that are displayed. So, as vscode receives the syntax diagnostics, it replaces the previous syntax diagnostics with the new ones, and so on. So vscode will receive the region semantic diagnostics, and replace the previous semantic diagnostics, but only for the ranges that the region semantic diagnostics were computed. Then, later, vscode receives the normal semantic diagnostics, and replaces the previous semantic diagnostics (all of them) with the new ones.

Because of that updating behavior, a first goal is for the region diagnostics to be a subset of the normal semantic diagnostics.
So even though the region diagnostics may not contain all of the semantic diagnostics for a specific range of a file, it should not contain a spurious diagnostic, because that would cause an annoying user experience, where an error shows up and then goes away.

The normal semantic diagnostics that are computed after the region semantic diagnostics are supposed to be the source of truth for semantic diagnostics for a given file. So a second goal is that there shouldn't be any difference between the semantic diagnostics we eventually compute for a whole file when we compute region semantic diagnostics vs when we don't compute semantic diagnostics. In other words, using this feature shouldn't change the diagnostics a user sees in the editor.

## Test results

To test this feature, I did a randomized test on the top 200 repos that consisted of opening a file, deleting a non-whitespace character, requesting diagnostics without providing a range, adding back the character, requesting diagnostics without providing a range. Then repeat this, but this time providing a range when requesting diagnostics. The first iteration is the baseline, i.e. the current behavior of `geterr`, and the second is the `geterr` behavior using this feature.

### Performance

Total files tested: 1725
Average time:

| Baseline whole-file semantic check | Region semantic checks | Whole-file semantic check (after region) | Line count |
| ------------------------------------- | --------------------------- | -------------------------------------------- | ------------ |
| 207 ms                                           | 120 ms                             | 68 ms                                                        | 1398           |

If we take the average for the **150 files** whose baseline semantic check took at least 500 ms, the results are:

| Baseline whole-file semantic check | Region semantic checks | Whole-file semantic check (after region) | Line count |
| ------------------------------------- | --------------------------- | -------------------------------------------- | ------------ |
| 1311 ms                                         | 755 ms                             | 472 ms                                                     | 7066           |

### Test on `checker.ts`

I did the same random testing as above but on our `checker.ts` (the largest file we have). Results:

| Baseline whole-file semantic check | Region semantic checks | Whole-file semantic check (after region) |
| ------------------------------------- | --------------------------- | -------------------------------------------- |
| 3330 ms                                         | 143 ms                             | 3199 ms                                                     |

## Diagnostic inconsistencies

The diagnostics generated by the checker depend on the order in which nodes are checked, and the state of the checker (e.g. type comparison caches). Because of that, using region semantic diagnostics may cause the checker to generate different diagnostics than it would if it didn't use it.

During testing, I found the following inconsistencies:

### Different diagnostic position

Some diagnostics have as position the node where we first do something. For instance, error ts2307:

```ts
/** @jsxImportSource @emotion/react */
import { css } from "@emotion/react";

function Component1() {
    return (
        <input // error ts2307: Cannot find module '@emotion/react/jsx-runtime' or its corresponding type declarations.
            css={css`
        color: red;
      `}
        />
    );
}
function Component2() {
    return (
        <Component1>
            <input
                css={css`
           color: red;
       `}
            />
        </Component1>
    );
}
```
Error 2307 will show up on the first JSX element the checker tries to check. When using region checks, the first JSX element we check for the file may be a different one.

Error 2354, "This syntax requires an imported helper but module 'tslib' cannot be found.", also behaves the same way: its position is the first `async` function we check.

In some cases, it's even possible for those diagnostics to occur in a different file altogether, because we might go off and check things in a different file while checking the file we actually want diagnostics for.

These diagnostics moving around inside the file is, I think, acceptable for now: either the diagnostic is going to be moved to a position that is in the region we're checking, and therefore in the visible parts of the file, or it will move around to a "random" position in the file, like you already get today.

Moving to a different file altogether is a worse problem. But, again, this can already happen today in the editor...

### Different elaboration

When we compare two types, we cache the result of the comparison so that in the future we can skip work and just use the cached result. That means the first time we compare two types, we do detailed comparisons (e.g. we compare each property, each component type in a union, etc). While we're doing this detailed comparison, we produce elaboration for the diagnostic that indicates two types are not comparable/assignable/etc, so that we provide a more detailed error message saying why that type comparison failed.

The next time we compare those same two types, we don't do this detailed work, and instead produce a diagnostic with less information.
This can happen in two ways:
(1) The less detailed diagnostic has a different error code from the more detailed one
(2) The less detailed diagnostic has the same error code as the more detailed one, but less elaboration in its message.

This is also reported in #3276, #33143.

Examples:
(1) Different error code
```ts
interface ThroughStream {
    a: string;
}
interface ReadStream {
    f: string;
    g: number;
    h: boolean;
    i: BigInt;
    j: symbol;
}
function foo(): ReadStream {
    return undefined as any as ThroughStream; // ts2739: Type 'ThroughStream' is missing the following properties from type 'ReadStream': f, g, h, i, j
}
function bar(): ReadStream {
    return undefined as any as ThroughStream; // ts2322: Type 'ThroughStream' is not assignable to type 'ReadStream'.
}
```

(2) Different elaboration
```ts
type Wrap = {
    someProp: Bar<number>;
}
type OtherWrap = {
    someProp: Foo<string>;
}
type Foo<T> = {
    foo: { what: T };
}
type Bar<T> = {
    foo: { what: T };
} | boolean;

function fun(param: Wrap): void {}

declare let fooStr: Foo<string>;
declare let otherWrap: OtherWrap;

fun(otherWrap); // ts2345: Argument of type 'OtherWrap' is not assignable to parameter of type 'Wrap'.
  Types of property 'someProp' are incompatible.
    Type 'Foo<string>' is not assignable to type 'Bar<number>'.
      Type 'Foo<string>' is not assignable to type '{ foo: { what: number; }; }'.
        The types of 'foo.what' are incompatible between these types.

let a: Bar<number> = fooStr;
```
If you change the order of the last two statements (or, alternatively, if you first type check the last statement, and then type check `fun(otherWrap)`), you get instead: 
```ts
let a: Bar<number> = fooStr;

fun(otherWrap); // ts2345: Argument of type 'OtherWrap' is not assignable to parameter of type 'Wrap'.
  Types of property 'someProp' are incompatible.
    Type 'Foo<string>' is not assignable to type 'Bar<number>'.
```
[Playground link](https://www.typescriptlang.org/play?target=9&ts=5.5.0-dev.20240405#code/PTAEFEA8EMFsAcA2BTAXKAJgSwGY+QE7IB2ALqIQQPYGgDGVGyo0ipVA5sqQBaEBQ-LGUI5odZgBUe1AK4ceAZVJE4oAN79Q2lugDOK4RwDc-AL5CRBMRNAAlZNAzLVsDVp059h4iY-aOdGJZWAAjQlMdUB50UKoqFGhiSJ0sdAAhLA4ASTIU7QArfQBPMITTCxxZYjpSLCpiUBx4gAoASnQHJxdHN00oolJZAkbqphxhZAwWPRZiYpnQaTkFHrgK-iqauobQUOgCds7HZxVe9wHuYdHiccnp6FmkhcelmSp5JTP180EQCBgCBQ6GweEIJHIyEQ0DiBGgO2Ign4pGK8GYAHU4fBQABeC46PRUWDIAAK1HgGQOAB5gmFCAA+DYotGgADyvEImOg2Lx-QJRNJ5PQADF4lSDAQjIzfszmKKqFTJPTcfjtM0qOh1KAAO48eHoSSgMxM1HMdLUpUqvlq+KanV60gGo0VUAAHz28USyUEW1q9UaWxa8AOcHQXPgHVAADcqFhpuoLPwmHRoURQChyOqXCKxRKpaZk6nmBnQFQOQRw+h2XwK1jTJtqi0yzXw216yXoJSCDSQuECMq8VmVPX+EA)

This difference already exists today, so also seems like it's acceptable for now.

~Note: The thing to be extra careful about is diagnostic duplication. Because the diagnostics can be different, if we check a node in the region check and then re-check it during the whole-file check, we might produce two different diagnostics for it, and then we don't de-duplicate those because we think they are different diagnostics (due to having a different error code or different error message).~ Fixed, see #58207.

### Different suggestion

We have a hardcoded limit (at most 12) on how many times we provide "Did you mean ... ?" suggestions for "Cannot find ..." errors. This seems to be because of performance reasons: it's costly to try and find suggestions.
The result is that, depending on where you start checking your file, some of those "Cannot find ..." errors may or may not have suggestions.

I think this is an acceptable difference in diagnostics.


jakebailey:
I haven't gone through the whole thing yet, but one thing we may want to do (maybe just for this PR's development and not long term) is to try and run this request at random positions before running compiler tests or similar. The thinking being that it shouldn't affect the analysis to do that. We had a fuzzer like this in pyright and it found loads of bugs in the type evaluator / checker that we were only previously able to observe via people crashing or reporting bugs we couldn't figure out.

src/server/session.ts
```diff
@@ -299,9 +299,10 @@ export function formatDiagnosticToProtocol(diag: Diagnostic, includeFileName: bo
         : common;
 }
 
-export interface PendingErrorCheck {
+interface PendingErrorCheck {
```
gabritto, line 302: I have no idea why we were exporting this in the first place, since it's only used as a parameter type of a private method. But I can export it again if it's relevant.

src/server/session.ts
```diff
@@ -1265,57 +1276,123 @@ export class Session<TMessage = string> implements EventSender {
...
+        }
+        this.sendDiagnosticsEvent(file, project, diagnosticsResult.diagnostics, "regionSemanticDiag", diagnosticsResult.spans);
+        tracing?.pop();
+        return true;
+    }
+
+    // We should only do the region-based semantic check if we think it would be considerably faster than a whole-file semantic check
+    protected shouldDoRegionCheck(file: NormalizedPath): boolean {
+        const perf = this.semanticCheckPerformance.get(file);
+        return !!perf && perf > 1000;
```
gabritto, line 1308: This is tentative and should be discussed; I don't know what number to use as a threshold, or if I should also consider other information (e.g. region size vs file size...) in this heuristic.

gabritto:
So far I locally tested this on the top 200 TS repos (relying on ts-error-deltas). Nothing crashed, I tested normal semantic diagnostics vs region-based semantic diagnostics on 1690 random files with at least 400 lines, at random positions, before and after deleting a random character. Average time for whole file semantic diagnostics was 198ms vs 120ms for region-based semantic diagnostics with a single range of 200 lines.

I'm going to check that the diagnostics themselves didn't change unexpectedly when using region-based diagnostics vs not using it during these tests, and I'll post updates here.

src/compiler/checker.ts
```diff
@@ -47308,9 +47316,24 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             // Grammar checking
             checkGrammarSourceFile(file);
 
+            clear(potentialThisCollisions);
```
jakebailey, line 47319: Honestly I think this whole "one giant array kept until the end" is super weird and I've been wondering if we could refactor this to not use this pattern...

gabritto: That could make my life more convenient for this PR.

gabritto:

gabritto:
I found enough weird error inconsistencies that I have to investigate to feel suspicious about this PR, so, drafting it again.

uhyo:
👀

gabritto:
Update on diagnostics inconsistencies:
I'm working on fixing the position inconsistencies and the type comparison elaboration inconsistencies; they'll be separate, follow-up PRs.

src/compiler/checker.ts
```diff
@@ -1542,6 +1542,8 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         onSuccessfullyResolvedSymbol,
     });
 
+    var inFullCheckAfterPartial = false;
```
weswigham, line 1545: This seems redundant with the `NodeFlags.PartiallyTypeChecked` flag itself. Isn't checking that node flag sufficient?

src/testRunner/unittests/tsserver/regionDiagnostics.ts
```diff
@@ -0,0 +1,76 @@
...
+                file: file1.path,
+                ranges: [{ startLine: 6, startOffset: 1, endLine: 7, endOffset: 13 }],
+            }],
+        });
+
+        baselineTsserverLogs("regionDiagnostics", "diagnostics for select nodes", session);
+    });
+});
+
+interface VerifyGetErrRegionRequest {
```
sheetalkamat, line 41: Please update `VerifyGetErrRequest` in `src\testRunner\unittests\helpers\tsserver.ts` to handle this instead so we can later in time add repros that contain these type of error requests easily

- Change files property to `(string | File | { file, range} ) []` and `SkipErrors` to contain `region?: true` so that we know when the `region` diagnsotics are present and need to postpone semantic and suggestion check to later. (modify `checkAllErrors`) to handle that.

src/testRunner/unittests/tsserver/regionDiagnostics.ts
```diff
@@ -0,0 +1,76 @@
...
+    // Run suggestion diagnostics
+    session.host.runQueuedImmediateCallbacks();
+}
+
+class RegionTestSession extends TestSession {
+    constructor(optsOrHost: TestSessionConstructorOptions) {
+        super(optsOrHost);
+    }
+
+    protected override shouldDoRegionCheck(_file: ts.server.NormalizedPath): boolean {
```
sheetalkamat, line 73: Instead add this to `TestSession` so its always done.

src/server/session.ts
```diff
@@ -1264,57 +1275,123 @@ export class Session<TMessage = string> implements EventSender {
...
+            tracing?.pop();
+            return false;
+        }
+        this.sendDiagnosticsEvent(file, project, diagnosticsResult.diagnostics, "regionSemanticDiag", diagnosticsResult.spans);
+        tracing?.pop();
+        return true;
+    }
+
+    // We should only do the region-based semantic check if we think it would be considerably faster than a whole-file semantic check
+    protected shouldDoRegionCheck(file: NormalizedPath): boolean {
```
sheetalkamat, line 1305: Please make this internal method .. this should be overrriden only by "TestSession"

src/server/session.ts
```diff
@@ -1264,57 +1275,123 @@ export class Session<TMessage = string> implements EventSender {
...
+            );
         }
         catch (err) {
             this.logError(err, kind);
         }
     }
 
     /** It is the caller's responsibility to verify that `!this.suppressDiagnosticEvents`. */
-    private updateErrorCheck(next: NextStep, checkList: readonly string[] | readonly PendingErrorCheck[], ms: number, requireOpen = true) {
+    private updateErrorCheck(
```
sheetalkamat, line 1337: Please update updateErrorCheck to be more readable as:
```suggestion
private updateErrorCheck(next: NextStep, checkList: readonly string[] | readonly PendingErrorCheck[], ms: number, requireOpen = true) {
        Debug.assert(!this.suppressDiagnosticEvents); // Caller's responsibility
        const seq = this.changeSeq;
        const followMs = Math.min(ms, 200);
        let index = 0;
        const getFullFileIndex = () => index - checkList.length;
        let filesFullCheck: PendingErrorCheck[] | undefined;
        const goNext = () => {
            index++;
            if (checkList.length > index || (filesFullCheck?.length ?? 0) > getFullFileIndex()) {
                next.delay("checkOne", followMs, checkOne);
            }
        };
        const checkOne = () => {
            if (this.changeSeq !== seq) {
                return;
            }
            let item: string | PendingErrorCheck | undefined = checkList[index] ?? filesFullCheck![getFullFileIndex()];
            if (isString(item)) {
                // Find out project for the file name
                item = this.toPendingErrorCheck(item);
                if (!item) {
                    // Ignore file if there is no project for the file
                    goNext();
                    return;
                }
            }
            const { fileName, project, ranges, skipSyntax } = item;
            // Ensure the project is up to date before checking if this file is present in the project.
            updateProjectIfDirty(project);
            if (!project.containsFile(fileName, requireOpen)) {
                return;
            }
            if (!skipSyntax) {
                this.syntacticCheck(fileName, project);
                if (this.changeSeq !== seq) {
                    return;
                }
            }
            // Don't provide semantic diagnostics unless we're in full semantic mode.
            if (project.projectService.serverMode !== LanguageServiceMode.Semantic) {
                goNext();
                return;
            }
            if (ranges) {
                return next.immediate("regionSemanticCheck", () => {
                    const didRegionCheck = this.regionSemanticCheck(fileName, project, ranges);
                    if (didRegionCheck) {
                        (filesFullCheck ??= []).push({ fileName, project, skipSyntax: true });
                        goNext();
                        return;
                    }
                    // If we opted out of region check, do semantic check as normal in next step
                });
            }
            next.immediate("semanticCheck", () => {
                this.semanticCheck(fileName, project);
                if (this.changeSeq !== seq) {
                    return;
                }
                if (this.getPreferences(fileName).disableSuggestions) {
                    goNext();
                    return;
                }
                next.immediate("suggestionCheck", () => {
                    this.suggestionCheck(fileName, project);
                    goNext();
                });
            });
        };
        if ((checkList.length + (filesFullCheck?.length ?? 0)) > index && this.changeSeq === seq) {
            next.delay("checkOne", ms, checkOne);
        }
    }
```

src/server/protocol.ts
```diff
@@ -2383,8 +2363,9 @@ export interface GeterrRequestArgs {
     /**
      * List of file names for which to compute compiler errors.
      * The files will be checked in list order.
+     * Files with ranges specified will be checked first.
      */
-    files: string[];
+    files: (string | FileRangesRequestArgs)[];
```
sheetalkamat, line 2368: ` GeterrForProjectRequest` does this also need to enable region based .. we already create high priority list etc in there.

src/server/protocol.ts
```diff
@@ -2383,8 +2363,9 @@ export interface GeterrRequestArgs {
     /**
      * List of file names for which to compute compiler errors.
      * The files will be checked in list order.
+     * Files with ranges specified will be checked first.
```
sheetalkamat, line 2366: This says the files with region are checked first but per your updateErrorCheck logic, its checking files in order with delaying full checks when regions are specified after all the files order is done.,  Is this comment uptodate or intentional order  for file1 with region, file2 , file3 region, file4 is to check file1 with region, file3 with region, file2, file 4, file1 semantic and suggestion, file3 semantic and suggestion...

tests/baselines/reference/api/typescript.d.ts
```diff
@@ -5933,8 +5948,8 @@ declare namespace ts {
         getOptionsDiagnostics(cancellationToken?: CancellationToken): readonly Diagnostic[];
         getGlobalDiagnostics(cancellationToken?: CancellationToken): readonly Diagnostic[];
         getSyntacticDiagnostics(sourceFile?: SourceFile, cancellationToken?: CancellationToken): readonly DiagnosticWithLocation[];
-        /** The first time this is called, it will return global diagnostics (no location). */
-        getSemanticDiagnostics(sourceFile?: SourceFile, cancellationToken?: CancellationToken): readonly Diagnostic[];
+        /** The first time this is called without `nodesToCheck`, it will return global diagnostics (no location). */
+        getSemanticDiagnostics(sourceFile?: SourceFile, cancellationToken?: CancellationToken, nodesToCheck?: Node[]): readonly Diagnostic[];
```
sheetalkamat, line 5952: I think "nodesToCheck" part of the API needs to be internal?

tests/baselines/reference/api/typescript.d.ts
```diff
@@ -9967,6 +9989,10 @@ declare namespace ts {
          * @param fileName A path to the file you want semantic diagnostics for
          */
         getSemanticDiagnostics(fileName: string): Diagnostic[];
+        /**
+         * Similar to {@link getSemanticDiagnostics}, but only checks the specified ranges of the file for diagnostics.
+         */
+        getRegionSemanticDiagnostics(fileName: string, ranges: TextRange[]): RegionDiagnosticsResult | undefined;
```
sheetalkamat, line 9995: internal?

src/testRunner/unittests/tsserver/regionDiagnostics.ts
```diff
@@ -0,0 +1,76 @@
...
+
+        verifyGetErrRegionRequest({
+            session,
+            files: [{
+                file: file1.path,
+                ranges: [{ startLine: 6, startOffset: 1, endLine: 7, endOffset: 13 }],
+            }],
+        });
+
+        baselineTsserverLogs("regionDiagnostics", "diagnostics for select nodes", session);
```
sheetalkamat, line 37: Following tests are needed:
- Multiple files ,- some with region , some without mixed. in order
- Files where suggestion diagnostics are present, - overlaps with selected nodes and dont overlap with selected nodes.

sheetalkamat (changes requested):

src/server/protocol.ts
gabritto: I wasn't planning on enabling region-based diagnostics for the project diagnostics scenario at the moment. From my last talk with Matt on the vscode usage of this command, it's still marked as experimental on the vscode side and it has known issues, so I figured I wouldn't support it at first.

gabritto:
Heads up when reviewing: commit https://github.com/microsoft/TypeScript/pull/57842/commits/a3cdc74908d69ebc00e7405f1483d88f65e91d4a updates around 100 unit test logs with the new diagnostics event property `"duration"`, properly sanitized. It might be easier to review that one separately.

src/server/session.ts
```diff
@@ -1264,57 +1275,123 @@ export class Session<TMessage = string> implements EventSender {
...
+        return !!perf && perf > 1000;
+    }
+
+    private sendDiagnosticsEvent(file: NormalizedPath, project: Project, diagnostics: readonly Diagnostic[], kind: protocol.DiagnosticEventKind, spans?: TextSpan[]): void {
         try {
-            this.event<protocol.DiagnosticEventBody>({ file, diagnostics: diagnostics.map(diag => formatDiag(file, project, diag)) }, kind);
+            const scriptInfo = Debug.checkDefined(project.getScriptInfo(file));
+            const duration = hrTimeToMilliseconds(this.hrtime(this.diagnosticsTime));
+            if (kind === "semanticDiag") {
+                this.semanticCheckPerformance?.set(file, duration);
```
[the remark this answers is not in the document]
gabritto: If I move the `if (kind === "semanticDiag") { ... }`, I would need to also move the line above, `const duration = hrTimeToMilliseconds(this.hrtime(this.diagnosticsTime));`, and then that line would be duplicated for the `semanticCheck`, `suggestionCheck`, `regionSemanticCheck`, `syntacticCheck` methods, no?

src/testRunner/unittests/tsserver/regionDiagnostics.ts
gabritto: I did something slightly different, which is to detect whether region diagnostics should be skipped or not based on the `ranges` property on the `file`, inside `checkAllErrors`. Let me know if that seems good.

src/server/session.ts
gabritto: I'm a bit confused, what specifically about `updateErrorCheck` do you think could be more readable?

src/server/protocol.ts
gabritto: The comment is true not because of `updateErrorCheck`, but because of this line in a caller of `updateErrorCheck`: https://github.com/microsoft/TypeScript/pull/57842/files#diff-c9fde3d99427a4b56c409a7dec15c227b21119d0d56bcfab4dbfad23f83118b0R2588.

I did this intentionally, as I think if there are regions of a file visible in the editor, it means they should come before files that are open in the editor but not currently being looked at, if that makes sense.

src/server/session.ts
gabritto: Update: I changed the behavior of `updateErrorCheck` and as a consequence the code is simpler now, let me know if there are still any parts that are confusing or could be improved.

gabritto:
@typescript-bot test this

gabritto:
@typescript-bot test it

gabritto:

gabritto:

src/server/session.ts
```diff
@@ -1266,28 +1277,81 @@ export class Session<TMessage = string> implements EventSender {
...
+        }
+        this.sendDiagnosticsEvent(file, project, diagnosticsResult.diagnostics, "regionSemanticDiag", diagnosticsResult.spans);
+        tracing?.pop();
+        return true;
+    }
+
+    // We should only do the region-based semantic check if we think it would be considerably faster than a whole-file semantic check
+    /** @internal */
+    protected shouldDoRegionCheck(file: NormalizedPath): boolean {
+        const lineCount = this.projectService.getScriptInfoForNormalizedPath(file)?.getLineCount();
```
sheetalkamat, line 1309: Sine this is the only place we are using linecount, why not just directly do direct evaluation instead of adding method to scriptInfo. 
```suggestion
        const lineCount = this.projectService.getScriptInfoForNormalizedPath(file)?.getLineInfo().getLineCount();
```

src/server/scriptInfo.ts
```diff
@@ -734,6 +734,10 @@ export class ScriptInfo {
             this.sourceMapFilePath = undefined;
         }
     }
+
+    getLineCount(): number {
+        return this.textStorage.getLineInfo().getLineCount();
+    }
```
sheetalkamat, line 740:
```suggestion
```

src/server/session.ts
```diff
@@ -1266,28 +1277,81 @@ export class Session<TMessage = string> implements EventSender {
...
+            const duration = hrTimeToMilliseconds(this.hrtime(this.diagnosticsTime));
+            if (kind === "semanticDiag" && !this.shouldDoRegionCheck(file)) {
+                this.semanticCheckPerformance.set(file, duration);
+            }
+
+            const body: protocol.DiagnosticEventBody = {
+                file,
+                diagnostics: diagnostics.map(diag => formatDiag(file, project, diag)),
+                spans: spans?.map(span => toProtocolTextSpan(span, scriptInfo)),
+                duration,
```
sheetalkamat, line 1333: Instead of passing this here do we want to use existing machinery we use for performance events that is : 
- Do we need this on every event per file per duiagnostic kind or can it be : on diagnostic completion event .
- Name this similar to out `protocol.Respose` as `performanceData` with correct keys

src/server/session.ts
```diff
@@ -1334,21 +1407,23 @@ export class Session<TMessage = string> implements EventSender {
...
+                    const didRegionCheck = this.regionSemanticCheck(fileName, project, ranges);
+                    if (didRegionCheck) {
+                        if (this.changeSeq !== seq) {
+                            return;
+                        }
+                        next.immediate("semanticCheck", () => doSemanticCheck(fileName, project));
+                    }
+                    else {
+                        doSemanticCheck(fileName, project);
+                    }
```
sheetalkamat, line 1422: i think whether we did region check or not if "region diangostics" were expected, we should do semantic diagnostics checking on next tick

src/server/session.ts
```diff
@@ -2479,19 +2554,26 @@ export class Session<TMessage = string> implements EventSender {
...
+                if (!scriptInfo) {
+                    return undefined;
+                }
+                const ranges = arg.ranges.map(range => this.getRange({ file: arg.file, ...range }, scriptInfo));
+                return {
+                    fileName,
+                    ranges,
+                };
+            });
+            this.updateErrorCheck(next, [...filesWithRange, ...files], delay);
```
sheetalkamat, line 2576: I am not 100% sold on we modifying the order. Why is it on us? why is it not on client? 
I would much prefer us just going in order like you had it before. 
Checking scriptInfo here is not ideal because we dont want to check for scriptinfo, its project and updates before we actually start calculating diagnostics. (project could change in between in background or new open request arrives etc) so this sorting functionality seems unnecessary and should not be our responsibility

src/server/session.ts
```diff
@@ -3279,6 +3362,7 @@ export class Session<TMessage = string> implements EventSender {
                 })),
                 request.arguments.closedFiles,
             );
+            this.cleanUpSemanticCheckPerformance(request.arguments.closedFiles);
```
sheetalkamat, line 3365: You need this on `protocol.CommandTypes.Close` as well

src/services/types.ts
```diff
@@ -445,6 +445,11 @@ export const enum SemanticClassificationFormat {
     TwentyTwenty = "2020",
 }
 
+export interface RegionDiagnosticsResult {
```
sheetalkamat, line 448:
```suggestion
/** @internal*/
export interface RegionDiagnosticsResult {
```

src/testRunner/unittests/tsserver/regionDiagnostics.ts
```diff
@@ -0,0 +1,215 @@
...
+foo(10, 50);`,
+        };
+        const host = createServerHost([file1]);
+        const session = new TestSession(host);
+
+        session.executeCommandSeq<ts.server.protocol.OpenRequest>({
+            command: ts.server.protocol.CommandTypes.Open,
+            arguments: { file: file1.path },
+        });
+
```
sheetalkamat, line 28:
```suggestion
openFilesForSession([file1], session);
```

src/testRunner/unittests/tsserver/regionDiagnostics.ts
```diff
@@ -0,0 +1,215 @@
...
+        verifyGetErrRequest({
+            session,
+            files: [{
+                file: file1.path,
+                ranges: [{ startLine: 6, startOffset: 1, endLine: 7, endOffset: 13 }],
```
sheetalkamat, line 33: May be use `protocolLocationFromSubstring` so its clear which locations you are referring to?

src/testRunner/unittests/tsserver/regionDiagnostics.ts
```diff
@@ -0,0 +1,215 @@
...
+        const files = [file1, file2, file3, file4];
+        const host = createServerHost(files);
+        const session = new TestSession(host);
+
+        session.executeCommandSeq<ts.server.protocol.UpdateOpenRequest>({
+            command: ts.server.protocol.CommandTypes.UpdateOpen,
+            arguments: {
+                openFiles: files.map(f => ({ file: f.path, fileContent: f.content })),
+            },
+        });
```
sheetalkamat, line 98:
```suggestion
openFilesForSession([files], session);
```

src/testRunner/unittests/tsserver/regionDiagnostics.ts
```diff
@@ -0,0 +1,215 @@
...
+
+        verifyGetErrRequest({
+            session,
+            files: [
+                {
+                    file: file1.path,
+                    ranges: [{ startLine: 6, startOffset: 1, endLine: 6, endOffset: 13 }], // `add(10, 50);`
```
sheetalkamat, line 105: Same pls use `protocolLocationFromSubstring` for better readability

src/testRunner/unittests/tsserver/regionDiagnostics.ts
```diff
@@ -0,0 +1,215 @@
...
+    it("diagnostics for select nodes in a single file", () => {
+        const file1 = {
+            path: "/a/b/app.ts",
+            content: `function foo(x: number, y: string): number {
+    return x + y;
+}
+
+
+
+foo(10, 50);`,
```
sheetalkamat, line 19:
```suggestion
            content: dedent`
                function foo(x: number, y: string): number {
                    return x + y;
               }
               foo(10, 50);
            `
```

sheetalkamat: This and others - pls indent for better readability

src/testRunner/unittests/tsserver/regionDiagnostics.ts
```diff
@@ -0,0 +1,215 @@
...
+        };
+        const host = createServerHost([file1]);
```
sheetalkamat, line 21: Add `libFile`

sheetalkamat: And in other tests

src/testRunner/unittests/tsserver/regionDiagnostics.ts
```diff
@@ -0,0 +1,215 @@
...
+    describe("diagnostics for select nodes when files have suggestion diagnostics", () => {
+        const config = {
+            path: "/tsconfig.json",
+            content: `
+{
+    "compilerOptions": {
+        "allowSyntheticDefaultImports": true,
+    }
+}
+`,
```
sheetalkamat, line 134:
```suggestion
            content: jsonToReadableText({
                compilerOptions: {
                    allowSyntheticDefaultImports: true
                }
            }),
```

src/testRunner/unittests/helpers/tsserver.ts
```diff
@@ -154,6 +155,10 @@ export class TestSession extends ts.server.Session {
         request.type = "request";
         return this.executeCommand(request);
     }
+
+    protected override shouldDoRegionCheck(_file: ts.server.NormalizedPath): boolean {
+        return true;
```
sheetalkamat, line 160: I would also add option in `TestSessionOptions` to `disableRegionDiagnostics?: boolan` and use that here to return false if enabled. So you can add tests when region diagnsotics are enabled but we skip them

src/testRunner/unittests/tsserver/regionDiagnostics.ts
```diff
@@ -0,0 +1,215 @@
...
+                    },
+                ],
+                skip: [
+                    { regionSemantic: false },
+                ],
+            });
+
+            baselineTsserverLogs("regionDiagnostics", "region does not have suggestion", session);
+        });
+    });
```
sheetalkamat, line 214: Add one test for each these scenarios where test session is skipping region diagnostics..

tests/baselines/reference/api/typescript.d.ts
```diff
@@ -8434,12 +8451,15 @@ declare namespace ts {
     function textSpanIsEmpty(span: TextSpan): boolean;
     function textSpanContainsPosition(span: TextSpan, position: number): boolean;
     function textSpanContainsTextSpan(span: TextSpan, other: TextSpan): boolean;
+    function textSpanContainsTextRange(span: TextSpan, range: TextRange): boolean;
+    function textRangeContainsTextSpan(range: TextRange, span: TextSpan): boolean;
```
sheetalkamat, line 8455: why are these public

tests/baselines/reference/api/typescript.d.ts
```diff
@@ -8434,12 +8451,15 @@ declare namespace ts {
...
     function textSpanOverlapsWith(span: TextSpan, other: TextSpan): boolean;
     function textSpanOverlap(span1: TextSpan, span2: TextSpan): TextSpan | undefined;
     function textSpanIntersectsWithTextSpan(span: TextSpan, other: TextSpan): boolean;
     function textSpanIntersectsWith(span: TextSpan, start: number, length: number): boolean;
     function decodedTextSpanIntersectsWith(start1: number, length1: number, start2: number, length2: number): boolean;
     function textSpanIntersectsWithPosition(span: TextSpan, position: number): boolean;
+    function textRangeIntersectsWithTextSpan(range: TextRange, span: TextSpan): boolean;
```
sheetalkamat, line 8462: Same.. should be internal unless we really need them to be public

src/testRunner/unittests/helpers/tsserver.ts
```diff
@@ -355,31 +360,55 @@ export interface VerifyGetErrRequestBase {
...
-        if (!skip?.[i]?.suggestion) session.host.runQueuedImmediateCallbacks();
+        if (fileSkip?.regionSemantic === false) {
+            // Run region check
+            session.host.runQueuedImmediateCallbacks();
+        }
+        // Run semantic check
+        if (!fileSkip?.semantic) session.host.runQueuedImmediateCallbacks();
+        // Run suggestion check
+        if (!fileSkip?.suggestion) session.host.runQueuedImmediateCallbacks();
     }
```
sheetalkamat, line 412: Please also add the check after these to ensure that session contains keys only for open files in `semanticCheckPerformance`

src/testRunner/unittests/helpers/tsserver.ts
sheetalkamat: I would also write it as below to ensure we are running logic for the real session here as:
`return super.shouldDoRegionCheck(file) && !this.disableRegionDiagnostics;`

src/server/session.ts
sheetalkamat: Also pls check if VS should be ok with this extra data coming in the event. I have completely lost track of if and when they use the events .. i think they use that for project diagnostics (GeterrForProjectRequest)

sheetalkamat (changes requested):

src/server/session.ts
gabritto: I think having the duration per file and diagnostic kind will help analyze the data we will collect via vscode telemetry. Is there a significant downside to having this data per file per kind of diagnostics?

gabritto:
@typescript-bot pack this

gabritto:
@typescript-bot pack this

src/testRunner/unittests/helpers/tsserver.ts
gabritto: I think the point of the this is to bypass the actual logic, since the test files will not be large enough to make the actual session implementation do the region check. I'll add an option to explicitly enable this bypassing, always on region check.

package.json
```diff
@@ -2,7 +2,7 @@
     "name": "typescript",
     "author": "Microsoft Corp.",
     "homepage": "https://www.typescriptlang.org/",
-    "version": "5.5.0",
+    "version": "5.6.0",
```
jakebailey, line 5: Bad merge? Or maybe out of date?

src/services/services.ts
```diff
@@ -2044,6 +2059,152 @@ export function createLanguageService(
...
+        }
+        const checkedSpans = normalizeSpans(nodes.map(node => createTextSpanFromBounds(node.getFullStart(), node.getEnd())));
+        const semanticDiagnostics = program.getSemanticDiagnostics(sourceFile, cancellationToken, nodes);
+        return {
+            diagnostics: semanticDiagnostics.slice(),
+            spans: checkedSpans,
+        };
+    }
+
+    function getNodesForRanges(file: SourceFile, ranges: TextRange[]): Node[] | undefined {
```
jakebailey, line 2090: Maybe this is out of place in our repo, but I almost wonder if it would end up being more performant to make this a generator and pass that down as an interable, avoiding collecting a huge list of nodes that happen to be on screen. But maybe it's not that many.

gabritto: I think that's doable, but I don't know how much that impacts performance. How impactful do you think this change could be?

src/compiler/checker.ts
```diff
@@ -47975,13 +47990,49 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+            }
+
+            // Grammar checking
+            checkGrammarSourceFile(file);
+
+            clear(potentialThisCollisions);
+            clear(potentialNewTargetCollisions);
+            clear(potentialWeakMapSetCollisions);
+            clear(potentialReflectCollisions);
+            clear(potentialUnusedRenamedBindingElementsInTypes);
```
jakebailey, line 48007: Sort of wonder if we should be making one single object that we could drop all at once instead. We seem to keep adding more and more stuff like this into the checker, but I guess it's not so bad.

src/compiler/utilities.ts
```diff
@@ -10116,11 +10116,11 @@ export function skipTypeChecking(sourceFile: SourceFile, options: CompilerOption
         options.skipDefaultLibCheck && sourceFile.hasNoDefaultLib) ||
         options.noCheck ||
         host.isSourceOfProjectReferenceRedirect(sourceFile.fileName) ||
-        !canIncludeBindAndCheckDiagnsotics(sourceFile, options);
+        !canIncludeBindAndCheckDiagnostics(sourceFile, options);
 }
 
 /** @internal */
-export function canIncludeBindAndCheckDiagnsotics(sourceFile: SourceFile, options: CompilerOptions) {
+export function canIncludeBindAndCheckDiagnostics(sourceFile: SourceFile, options: CompilerOptions) {
```
jakebailey, line 10123: Nit, but I would suggest sending this typo fix to main just to get it out of this PR's diff.

src/compiler/checker.ts
```diff
@@ -47488,6 +47488,10 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     }
 
     function checkSourceElementWorker(node: Node): void {
```
jakebailey, line 47490: I think this function (and checkSourceElement) should probably only accept SourceElement types rather than Node.

src/compiler/utilitiesPublic.ts
```diff
@@ -366,30 +376,59 @@ export function textSpanOverlap(span1: TextSpan, span2: TextSpan): TextSpan | un
...
+}
+
 export function textSpanIntersection(span1: TextSpan, span2: TextSpan): TextSpan | undefined {
     const start = Math.max(span1.start, span2.start);
     const end = Math.min(textSpanEnd(span1), textSpanEnd(span2));
     return start <= end ? createTextSpanFromBounds(start, end) : undefined;
 }
 
+/** @internal */
+export function normalizeSpans(spans: readonly TextSpan[]): TextSpan[] {
```
jakebailey, line 409: Can you stick a JSDoc comment describing what "normalization" is, for reference?

src/services/services.ts
jakebailey: I think it depends on how big these arrays tend to be, and if we ever cancel the diagnostics request. I'm thinking of me looking at a particularly heavy part of the checker and scrolling, where it may turn out to be faster to not have produced a giant list and done a bunch of normalization or something. But, likely not required, just thinking aloud.

jakebailey (commented):
So far I feel like this is looking good; happy with how minimal the changes seem to be.

sheetalkamat (changes requested):
.

src/compiler/utilities.ts
gabritto: #58840

sheetalkamat: Sorry about the typo and extra work you had to do @gabritto

src/compiler/checker.ts
gabritto: I just tried that and it causes a million errors, some because if you write `SourceElement` as a union of specific node types, then `TypeNode` is not assignable to `SourceElement`, and some others because `forEachChild`'s type doesn't know the children of, say, `ClassDeclaration` are all `SourceElement`. So I'm not doing that here.

gabritto:

src/server/session.ts
```diff
@@ -993,6 +993,9 @@ export class Session<TMessage = string> implements EventSender {
     private eventHandler: ProjectServiceEventHandler | undefined;
     private readonly noGetErrOnBackgroundUpdate?: boolean;
 
+    // Minimum number of lines for attempting to use region diagnostics for a file.
+    protected regionDiagLineCountThreshold = 500;
```
sheetalkamat, line 997: internal pls.

sheetalkamat (approved):

src/server/session.ts
sheetalkamat: apart from this , it looks good

PR #57847, opened 2024-03-19
Control flow analysis for element access with variable index

### LINKED ISSUE #56389: Narrow object property when key is a variable

ahejlsberg:
With this PR we perform control flow analysis for element access expressions `obj[key]` where `key` is a `const` variable, or a `let` variable or parameter that is never targeted in an assignment. The intuition here is that even if the exact name of the property selected by `key` isn't known, any control flow analysis proof that involves `obj[key]` should hold in subsequent statements as long as `obj` and `key` aren't mutated,

Some examples:

```ts
function f1(obj: Record<string, unknown>, key: string) {
    if (typeof obj[key] === "string") {
        obj[key].toUpperCase();  // Now ok, previously was error
    }
}

type Thing = { a?: string, b?: number, c?: number };

function f2(obj: Thing, getKey: () => keyof Thing) {
    const key = getKey();
    if (obj[key] !== undefined) {
        if (typeof obj[key] === "string") {
            obj[key].toUpperCase();  // Now ok, previously was error
        }
        if (typeof obj[key] === "number") {
            obj[key].toFixed();  // Now ok, previously was error
        }
    }
}

function f3<K extends string>(obj: Record<K, string | undefined>, key: K) {
    if (obj[key]) {
        obj[key].toUpperCase();  // Now ok, previously was error
    }
}
```

Fixes #56389.

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -26611,7 +26611,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 if (isElementAccessExpression(source) && isElementAccessExpression(target) && isIdentifier(source.argumentExpression) && isIdentifier(target.argumentExpression)) {
                     const symbol = getResolvedSymbol(source.argumentExpression);
                     if (isConstantVariable(symbol) || isParameterOrMutableLocalVariable(symbol) && !isSymbolAssigned(symbol)) {
-                        return symbol === getResolvedSymbol(target.argumentExpression);
+                        return symbol === getResolvedSymbol(target.argumentExpression) && isMatchingReference(source.expression, target.expression);
                     }
                 }
                 break;
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -26653,7 +26653,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     }
 
     function tryGetNameFromEntityNameExpression(node: EntityNameOrEntityNameExpression) {
-        const symbol = isIdentifier(node) ? getResolvedSymbol(node) : resolveEntityName(node, SymbolFlags.Value, /*ignoreErrors*/ true);
+        const symbol = resolveEntityName(node, SymbolFlags.Value, /*ignoreErrors*/ true);
         if (!symbol || !(isConstantVariable(symbol) || (symbol.flags & SymbolFlags.EnumMember))) return undefined;
 
         const declaration = symbol.valueDeclaration;
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -26610,8 +26610,8 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 }
                 if (isElementAccessExpression(source) && isElementAccessExpression(target) && isIdentifier(source.argumentExpression) && isIdentifier(target.argumentExpression)) {
                     const symbol = getResolvedSymbol(source.argumentExpression);
-                    if (isConstantVariable(symbol) || isParameterOrMutableLocalVariable(symbol) && !isSymbolAssigned(symbol)) {
-                        return symbol === getResolvedSymbol(target.argumentExpression) && isMatchingReference(source.expression, target.expression);
+                    if (symbol === getResolvedSymbol(target.argumentExpression) && (isConstantVariable(symbol) || isParameterOrMutableLocalVariable(symbol) && !isSymbolAssigned(symbol))) {
+                        return isMatchingReference(source.expression, target.expression);
                     }
                 }
                 break;
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -26543,13 +26543,20 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 return getFlowCacheKey((node as NonNullExpression | ParenthesizedExpression).expression, declaredType, initialType, flowContainer);
             case SyntaxKind.QualifiedName:
                 const left = getFlowCacheKey((node as QualifiedName).left, declaredType, initialType, flowContainer);
-                return left && left + "." + (node as QualifiedName).right.escapedText;
+                return left && `${left}.${(node as QualifiedName).right.escapedText}`;
             case SyntaxKind.PropertyAccessExpression:
             case SyntaxKind.ElementAccessExpression:
                 const propName = getAccessedPropertyName(node as AccessExpression);
                 if (propName !== undefined) {
                     const key = getFlowCacheKey((node as AccessExpression).expression, declaredType, initialType, flowContainer);
-                    return key && key + "." + propName;
+                    return key && `${key}.${propName}`;
+                }
+                if (isElementAccessExpression(node) && isIdentifier(node.argumentExpression)) {
+                    const symbol = getResolvedSymbol(node.argumentExpression);
+                    if (isConstantVariable(symbol) || isParameterOrMutableLocalVariable(symbol) && !isSymbolAssigned(symbol)) {
+                        const key = getFlowCacheKey((node as AccessExpression).expression, declaredType, initialType, flowContainer);
+                        return key && `${key}.@${getSymbolId(symbol)}`;
+                    }
                 }
                 break;
             case SyntaxKind.ObjectBindingPattern:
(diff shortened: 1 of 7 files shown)


ahejlsberg:
@typescript-bot test top200
@typescript-bot run dt

jakebailey:
Weird; reading some of the breaks, there seems to be a bug in this iteration of the PR where unrelated accesses are getting narrowed too.

```ts
declare const key: string;

declare const obj1: Record<string, number | undefined>;
declare const obj2: Record<string, number | undefined>;

if (obj1[key]) {
    const x = obj1[key];
    //    ^?

    const y = obj2[key];
    //    ^?
}
```

image

ahejlsberg:
Argh, I see what the issue is. Will fix.

ahejlsberg:
@typescript-bot test top200
@typescript-bot run dt

ahejlsberg:

jakebailey:
@typescript-bot pack this

ahejlsberg:
@typescript-bot test top200

jakebailey:
@typescript-bot pack this

ahejlsberg:
Tests and performance look good. This is ready for a review.

src/compiler/checker.ts
```diff
@@ -26595,9 +26602,19 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
-                    isMatchingReference((source as AccessExpression).expression, (target as AccessExpression).expression);
+                if (sourcePropertyName !== undefined) {
+                    const targetPropertyName = isAccessExpression(target) ? getAccessedPropertyName(target) : undefined;
+                    if (targetPropertyName !== undefined) {
+                        return targetPropertyName === sourcePropertyName && isMatchingReference((source as AccessExpression).expression, (target as AccessExpression).expression);
+                    }
+                }
+                if (isElementAccessExpression(source) && isElementAccessExpression(target) && isIdentifier(source.argumentExpression) && isIdentifier(target.argumentExpression)) {
+                    const symbol = getResolvedSymbol(source.argumentExpression);
+                    if (symbol === getResolvedSymbol(target.argumentExpression) && (isConstantVariable(symbol) || isParameterOrMutableLocalVariable(symbol) && !isSymbolAssigned(symbol))) {
```
jakebailey, line 26613: Does it make more sense to do `isPastLastAssignment` using a more specific location? How does this PR interact with #56908?

ahejlsberg: Unfortunately not that simple. We would need to know that _all_ narrowing operations for `obj[key]` occur past the last assignment to `key` which would require an additional CFA graph walk. We don't have that issue in #56908 because the only location we care about is the location where the arrow function is created.

ahejlsberg:

src/compiler/checker.ts
jakebailey: Makes sense; I really need to prototype the idea I had to make that efficient enough for us to do, but I've lost that brainwave ☹️

Can you add a test in the vein of #56908 in terms of what closures do? Something like: [Playground Link](https://www.staging-typescript.org/play?ts=5.5.0-pr-57847-24#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXyk11QAowAjALnhIEp4BeAPngDccthbr3OBuALAAoYSnTY8SAIwkA1iACe1AM4YYWVAHN6Ab2HwD8CCAzwc5AFbUASuBwxgAHlXqtAGnguNm+AB94aKCIGiDALAzwOgC+gkKGZpYA2vIKALqM8ABEUORgmbHxYHiqbFIZ5hbJiqkFhgD0dfEAegD8+oaEEqR0jCw68EWoJQAe5Ukp6VG0tQYN8fMLzW1xhhVVaRmBIMGooTMDxaasAExjlRP7c4at7QadxCQ9zJEHQ6ajEWsT8FOXjYsA67LKLCYRzACSyk8WAAtlhoDB4BgcAA6NGgoRiIiSRDHHp6FYGYymCoqNTePwBVBBELATbU7a0-YVDLZXL5W6vEqsMqfSx-Jac+54R70Z79QYjM4-aacq6AgE3QkJCz0mm7YD7SVHU58iwCoFC7HdMV9LnvaW-OX-BWLJVRIA) (which does behave properly I think)

gabritto (approved):

src/compiler/checker.ts
ahejlsberg: > Can you add a test in the vein of...

Such a test wouldn't show anything. The new analysis in #56908 only affects references to local variables, not references to properties, so there are no changes across closure boundaries with this PR.

ipanasenko:
Hi. I'm not sure if it's okay to ask here or I should rather open new discussion, so sorry in advance.

Do you think the following type narrowing using `in` should work as well?
I have a `Record` with predefined keys in it, and I expected that if `in` returns true, then `animal` can only be of `Animal` type.
https://www.typescriptlang.org/play/?#code/C4TwDgpgBAggdgSwLYEMA2UC8UDkBjFYHKAH1wBMB7Ac2LJwCMEAncnAbgCg9K4BnYFBSJUaAAooA7nwCyKMAC4oAJQg9WAHnjJ0AGihwArkgYRmAPixQA3pyhQCwBQBZddqFWou39pqyUATG4AvlzcvAJQABaUknJwIBLSVgAUwjpoSgLMCHDUAJRKRiZmWJa29ggAZlBpIuhQuUL14lKy8vk27vbMEMCGzHDNGUntYADa6aIAulz2wZzuvf2DUACScFW5CKBcwUA

I think this might be related to this feature.

JoostK:
> Hi. I'm not sure if it's okay to ask here or I should rather open new discussion, so sorry in advance.
> 
> Do you think the following type narrowing using `in` should work as well? I have a `Record` with predefined keys in it, and I expected that if `in` returns true, then `animal` can only be of `Animal` type. https://www.typescriptlang.org/play/?#code/C4TwDgpgBAggdgSwLYEMA2UC8UDkBjFYHKAH1wBMB7Ac2LJwCMEAncnAbgCg9K4BnYFBSJUaAAooA7nwCyKMAC4oAJQg9WAHnjJ0AGihwArkgYRmAPixQA3pyhQCwBQBZddqFWou39pqyUATG4AvlzcvAJQABaUknJwIBLSVgAUwjpoSgLMCHDUAJRKRiZmWJa29ggAZlBpIuhQuUL14lKy8vk27vbMEMCGzHDNGUntYADa6aIAulz2wZzuvf2DUACScFW5CKBcwUA
> 
> I think this might be related to this feature.

That error is correct and desirable: the index signature only specifies that properties in `Animal` (i.e. `'cat' | 'dog' | 'bird'`) have a number value, it does _not_ specify a type for any other properties. Since TypeScript does not have strict types it's fine for a `Record<Animal, number>` to also have _other_ properties, [as demonstrated in this playground](https://www.typescriptlang.org/play/?#code/C4TwDgpgBAggdgSwLYEMA2UC8UDkBjFYHKAH1wBMB7Ac2LJwCMEAncnAbgCg0JgoVEqNAAUUAdwDOAWRRgAXFABKEPJVYAeeMnQAaKHACuSBhGYA+LFADenKFALA5AFh22oVas9d2mrBQCZXAF8uTlU4CT4xBGAACylKOABrCBBLGzsAOmyBbRFxaVlvKCRElJAFHABJfiR9Sj4UfSMTZig5AFoAShxg0NyhUUkZMEtouITk1NDwyKhYyjEZOBAhiUsACgH0BUjmBDhqLoVDY1MsCwyoBAAzKC3BdGu4fkf84dku6zc7Zl4DZgvbbvQpgADawIAulw7EFOG4-sAAS8qnAbgcYiAuHCwokJJQeJk0DQNgslgJVgUNjhSlMQDgul0gA). If the `animal` parameter is changed from type `string` to `Animal` then TypeScript no longer reports an error for the implementation of `howManyPaws`, while correctly preventing the monkey from requesting its number of paws.

ipanasenko:
I see. Big thank you for an explanation

geakstr:
Hi and thank you for your work! Trying to clarify for myself this statement _"The intuition here is that even if the exact name of the property selected by key isn't known, any control flow analysis proof that involves obj[key] should hold in subsequent statements as long as obj and key aren't mutated"_ from the PR description. 

Here I mutate `obj` and TS 5.5.2 doesn't throw errors: 

```
function f1(obj: Record<string, unknown>, key: string) {
  if (typeof obj[key] === "string") {
    delete obj[key];
    obj[key].toUpperCase(); // ok
  }
}
```

Is this "expected"?

PR #57848, opened 2024-03-19
Remove rich-navigation workflow

jakebailey:
This service has been discontinued: https://github.com/microsoft/RichCodeNavIndexer/blob/da2609c36cf846790334413645b0786ba8dab26a/src/main.ts#L4


DanielRosenwasser:
I don't get it, is it because every repo has it enabled now?

DanielRosenwasser (approved):

jakebailey:
I've found no info relating to this, but the action is just a no-op now and isn't doing anything.

PR #54637, opened 2023-06-13
fix(dts): add missing types to support resizable arraybuffer

indrajitbnikam:
Fixes #54636

I am opening this PR to fix the missing(not updated) type for ArrayBuffer. This PR is being raised due to this discussion in [denoland/deno](https://github.com/denoland/deno) issue [here](https://github.com/denoland/deno/issues/19191).

PR #57858, opened 2024-03-19
fix(dts): array buffer types (ES2024,ES2017)

### LINKED ISSUE #54636: fix(dts): Update type of `ArrayBuffer`
# lib Update Request

The problem is that proper types are not available when `Resizable ArrayBuffer` need to be created by passing `options` in the constructor.

## Configuration Check

My compilation *target* is `ESnext` and my *lib* is `the default`.

**`ArrayBuffer`** Constructor signature.

## Sample Code

```TypeScript
const buff = new ArrayBuffer(10, {
  maxByteLength: 100,
});

console.debug(buff.resizable);
console.debug(buff.byteLength);

buff.resize(100);
console.debug(buff.byteLength);
```

## Documentation Link

You can check the signature over here.

[ArrayBuffer() constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer/ArrayBuffer)

santifoo:
This should be an updated, and hopefully correct fix from the stale PR here https://github.com/microsoft/TypeScript/pull/54637.

In summary this PR is an extension of the previous PR with several modifications (to comments, interfaces, references etc.) The previous PR included ArrayBuffer-related interfaces that are part of ES2024.

It passes most of the tests with the exceptions of baselines from what I understand, so those are accepted and send with a commit.

Would be useful to get some feedback if this looks good in principle @petamoriken @sandersn, I've included most of the reviews you made in the stale PR that I linked earlier. 

Fixes #54636


santifoo:
I'm unsure what the licence implies about original code, part of this PR is based on a stale PR linked at the top, since it's impossible to rebase the original one (it's 800 commits behind.)

jakebailey:
Not a lawyer, but they gave the PR away in https://github.com/microsoft/TypeScript/pull/54637#issuecomment-1859019824, so I would just add `Co-authored-by: ...` to your commit to credit the original code.

santifoo:
> Not a lawyer, but they gave the PR away in https://github.com/microsoft/TypeScript/pull/54637#issuecomment-1859019824, so I would just add `Co-authored-by: ...` to your commit to credit the original code.

Thanks, sounds good. I can only do that revising and forcing a push I think. Is this fine for you?

jakebailey:
Yes, feel free to force push

santifoo:
> Yes, feel free to force push

done, it did work because the profile picture is correct in the co-authored commit.

santifoo:

santifoo:
Still needs some tests https://github.com/microsoft/TypeScript/pull/54637#pullrequestreview-1504255858

and also for the new def of `slice()` and the empty ArrayBuffer and such constructors

BlackAsLight:
Any update on this?

santifoo:
> Any update on this? 

I can't spend the time now. I think it needs some tests only. Feel free to take over the PR.

src/lib/esnext.arraybuffer.d.ts
```diff
@@ -0,0 +1,53 @@
...
+     */
+    get resizable(): boolean;
+
+    /**
+     * Resizes the ArrayBuffer to the specified size (in bytes).
+     * @param newByteLength The new length, in bytes, to resize the ArrayBuffer to.
+     */
+    resize(newByteLength: number): undefined;
+}
+
```
fwqaaq, line 17: I think there are other methods and properties that need to be added. Like this: <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer/detached>

petamoriken:
Oh, It appears that the account has been lost. I'll take it instead.

guest271314:
@petamoriken Status?

petamoriken:
@guest271314 Please see #58573

PR #57859, opened 2024-03-20
[Experiment] feat(51556): support satisfies operator on functions/classes

### LINKED ISSUE #51556: Support `satisfies` operator on functions
# Suggestion

## 🔍 Search Terms

satisfies function

## ✅ Viability Checklist

## ⭐ Suggestion

Support the `satisfies` operator to appear after a function

```ts
function foo () { } satisfies Bar
```

This was requested/mentioned in #47920 several times but never followed up on

## 📃 Motivating Example

`satisfies` is super handy to use, and you can already use it for arrow functions. Not having support for it on functions feels unintuitive and forces a certain coding style where people have to use arrow functions when they rather want to use functions.

## 💻 Use Cases

Typing a function while preserving its return type.

a-tarasyuk:
Fixes #51556


a-tarasyuk:
@alvis It's still a draft because there's no final consensus on certain cases. In addition, I think that version 5.6 won't include this feature...

I'm uncertain about the future of this feature, however, @RyanCavanaugh or @DanielRosenwasser might have more information.

louwers (commented):
Looking forward to this functionality. I think the satisfies operator is already supported on functions, it's just not supported on function declarations.

Does this also bring support for `@satisfies` in JSDoc? Could you add tests for it? See my comment [here](https://github.com/microsoft/TypeScript/issues/51556#issuecomment-2395424930
) for an example that is expected to work.

a-tarasyuk:
@louwers You're right — `satisfies` currently works only for expressions. I think it’s better to handle JSDoc support in a separate PR. This PR should focus on supporting `satisfies` for func declarations (and possibly class declarations https://github.com/microsoft/TypeScript/issues/60849?).

eczn:
> @louwers You're right — `satisfies` currently works only for expressions. I think it’s better to handle JSDoc support in a separate PR. This PR should focus on supporting `satisfies` for func declarations (and possibly class declarations #60849?).

``` ts
class Test satisfies TestStatic implements TestDynamic { ... } 
class Test implements TestDynamic satisfies TestStatic { ... } 

// equivalent to two-line: 
class Test implements TestDynamic { ... }
Test satisfies TestStatic;
```

i can't wait to write such code precent to type-checking for static member types on `Class Declaration`

PR #57867, opened 2024-03-20
Preserve resolution-mode as-written too

jakebailey:
I lost this change while working on #57681 and meant for it to be there; my understanding is that we want reference directives to be basically written verbatim. Conditionally emitting `resolution-mode` based on `impliedNodeFormat` breaks that rule.

Also, while here, put `const preserve = ...` after `const = resolutionMode`. Not sure why I added it out of order.


andrewbranch (approved):
I guess this should have been discussed, but it should be small potatoes compared to needing to opt into preservation in the first place.

RyanCavanaugh (approved):

sheetalkamat (approved):

PR #50267, opened 2022-08-10
Report aggregate statistics for solution as well as some solution perf numbers

sheetalkamat:
This change under `--extendedDiagnostics` aggregates the diagnostics from all projects built and reports it at the end. Apart from that it also outputs some measurements for work that happens in tsc --build like finding if projects are uptodate etc.
Also removes unnecessary node count per suggestion by @amcasey 

Different work per suggestion by @rbuckton in #49285

```
c:\TypeScript>node c:\TypeScript\built\local\tsc.js -b src --extendedDiagnostics -v
[4:35:05 PM] Projects in this build:
    * src/compiler/tsconfig.json
    * src/executeCommandLine/tsconfig.json
    * src/tsc/tsconfig.json
    * src/jsTyping/tsconfig.json
    * src/services/tsconfig.json
    * src/server/tsconfig.json
    * src/webServer/tsconfig.json
    * src/deprecatedCompat/tsconfig.json
    * src/tsserver/tsconfig.json
    * src/typingsInstallerCore/tsconfig.json
    * src/typingsInstaller/tsconfig.json
    * src/watchGuard/tsconfig.json
    * src/debug/tsconfig.json
    * src/cancellationToken/tsconfig.json
    * src/dynamicImportCompat/tsconfig.json
    * src/harness/tsconfig.json
    * src/loggedIO/tsconfig.json
    * src/testRunner/tsconfig.json
    * src/tsconfig.json

[4:35:05 PM] Project 'src/compiler/tsconfig.json' is out of date because output for it was generated with version '4.4.0-dev' that differs with current version '4.9.0-dev'

[4:35:05 PM] Building project 'c:/TypeScript/src/compiler/tsconfig.json'...

Files:                         163
Lines of Library:             9124
Lines of Definitions:        37786
Lines of TypeScript:        155407
Lines of JavaScript:             0
Lines of JSON:                   0
Lines of Other:                  0
Identifiers:                360048
Symbols:                    206708
Types:                       85416
Instantiations:             144470
Memory used:               569041K
Assignability cache size:    34575
Identity cache size:          2497
Subtype cache size:          15697
Strict subtype cache size:    3387
I/O Read time:               0.05s
Parse time:                  1.72s
ResolveModule time:          0.05s
ResolveTypeReference time:   0.00s
Program time:                1.91s
Bind time:                   1.07s
Check time:                 11.68s
transformTime time:          1.66s
Source Map time:             0.58s
commentTime time:            0.68s
printTime time:              7.81s
[...]

PR #42586, opened 2021-02-01
Switched 'ts.performance' to a mixed mode only uses native performance APIs when necessary

rbuckton:
@amcasey recently discovered that the Web Performance API in NodeJS adds quite a bit more overhead to the total compile time than what we were doing before. This changes our use of native performance APIs in the following ways:

- Revert back to calculating our own `mark` times and `measure` durations in `ts.performance` rather than using `PerformanceObserver`.
- Conditionally call the native `mark` and `measure` when _any_ of the following are true:
  - We are running in the browser (since a debugger can start a cpu profile at any time)
  - We are running in NodeJS with Node's `--cpu-prof` option specified (to capture user timing events in the profile)
  - We are running in NodeJS with V8's `--prof` option specified (to capture user timing events in the profile)
  - We are running in NodeJS with our own `--generateCpuProfile` option specified (to capture user timing events in the profile)
  - We are running in NodeJS and `sys.debugMode` is true (since a debugger can start a cpu profile at any time)

If none of those cases are true, we will not call the native `mark` and `measure` methods. I've filed https://github.com/nodejs/diagnostics/issues/464 to track the performance issue we encountered in NodeJS, and it sounds like they plan to have a patch available in the near future that improves their implementation.

PR #57875, opened 2024-03-21
Use performance.now when possible, re-disable unconditional perf marking on all Node versions

jakebailey:
Now that we're testing different OSs, Node 14 on Windows spuriously fails, like:

```
diff --git a/./tests/baselines/reference/tsserver/fourslashServer/nodeNextPathCompletions.js b/./tests/baselines/local/tsserver/fourslashServer/nodeNextPathCompletions.js
index 50e476c62..5a9909bcd 100644
--- a/./tests/baselines/reference/tsserver/fourslashServer/nodeNextPathCompletions.js
+++ b/./tests/baselines/local/tsserver/fourslashServer/nodeNextPathCompletions.js
@@ -1215,7 +1215,7 @@ Info seq  [hh:mm:ss:mss] request:
     }
Info seq  [hh:mm:ss:mss] Starting updateGraphWorker: Project: /tsconfig.json
Info seq  [hh:mm:ss:mss] FileWatcher:: Added:: WatchInfo: /node_modules/dependency/lib/package.json 2000 undefined Project: /tsconfig.json WatchType: File location affecting resolution
-Info seq  [hh:mm:ss:mss] Finishing updateGraphWorker: Project: /tsconfig.json projectStateVersion: 2 projectProgramVersion: 1 structureChanged: true structureIsReused:: SafeModules Elapsed:: *ms
+Info seq  [hh:mm:ss:mss] Finishing updateGraphWorker: Project: /tsconfig.json projectStateVersion: 2 projectProgramVersion: 1 structureChanged: true structureIsReused:: SafeModules Elapsed: -1ms
Info seq  [hh:mm:ss:mss] Project '/tsconfig.json' (Configured)
Info seq  [hh:mm:ss:mss]   Files (2)
  /node_modules/dependency/lib/index.d.ts Text-1 "export function fooFromIndex(): void;"
@@ -1242,7 +1242,7 @@ Info seq  [hh:mm:ss:mss] response:
       "request_seq": 28,
       "success": true,
       "performanceData": {
-        "updateGraphDurationMs": *
+        "updateGraphDurationMs": -1
       },
       "body": {
         "isGlobalCompletion": false,
Error: Process completed with exit code 1.
```

https://github.com/microsoft/TypeScript/actions/runs/8367516104/job/22910040628?pr=57870#step:6:1

Somehow, time went backwards???

My guess is that after #50267, native performance hooks only work in node 16.7+ since we wholesale disable native performance support when some parts of it are missing. This also disables the use of `performance.now`, meaning Node 14 is falling back to `Date.now()` for `timestamp()` even though `performance.now()` would have worked. Now, why `Date.now()` would ever go backwards a millisecond, I have no idea, but the flag isn't happening anywhere other than Node 14, and 14 -> 16 is where `timestamp()` differs too. Surely not a coincidence...

This PR restructures things such that we can still use native performance time functions when available, but still disable the other perf stuff when the mark/measure support is incomplete.

Also, while here, don't check for `PerformanceObserver`, which we haven't used since #42586.

Funnily, this has exposed the fact that we have been accidentally enabling native perf marking on Node 20+, because it implements all of the global perf hooks we need and so we "think" it's a browser.

This PR started out as slow because I removed the check for `PerformanceObserver`, which then made Node 18 _also_ satisfy the checks via globals, so we slowed down and that pointed out the problem.


src/compiler/performance.ts
```diff
@@ -180,13 +180,13 @@ export function enable(system: System = sys) {
     if (!enabled) {
         enabled = true;
         perfHooks ||= tryGetNativePerformanceHooks();
-        if (perfHooks) {
+        if (perfHooks?.performance) {
             timeorigin = perfHooks.performance.timeOrigin;
             // NodeJS's Web Performance API is currently slower than expected, but we'd still like
             // to be able to leverage native trace events when node is run with either `--cpu-prof`
             // or `--prof`, if we're running with our own `--generateCpuProfile` flag, or when
             // running in debug mode (since its possible to generate a cpu profile while debugging).
```
jakebailey, line 188: Node has a performance API, and its performance API is the same as the web API and is also exposed globally. I'm not sure that checking for the globals actually makes much sense.

image

But, I've kept in this PR the behavior of still enabling native perf when the global is present.

jakebailey:

RyanCavanaugh:
Many years ago I encountered similar behavior where the time reported by each *core* on the CPU was slightly different (this was C# code and the behavior didn't repro when I set the thread processor affinity). Unclear if this was a hardware bug or just not an invariant under Windows.

RyanCavanaugh (approved):

jakebailey:
Yeah, I'm pretty sure that `performance.now()` is a completely different clock that's monotomic and everything (hence why it has its own timeOrigin).

jakebailey:
Going to rerun this because that result does not seem right at all

jakebailey:
Okay this is definitely slower for some reason.

jakebailey:
I really have no clue what's going on here; I've tried to compare this and there's just no reason for this to be slower.

fatcerberus:
> Somehow, time went backwards???

Spectre mitigation, perhaps?

jakebailey:

jakebailey:
I understand what happened; PerformanceObserver is only available as a global on Node 20+. This meant that Node 18 was not getting performance hooks at all as the old code was all or nothing. Then I removed that check, which then re-enabled hooks on Node 18, which is what the faster benchmark uses.

Now the order is swapped and so it's no longer enabled in Node at all. Probably this means Node 20 is faster now too.

Will update the description and comments and stuff to reflect that.

@typescript-bot perf test this full

jakebailey:
Wow, literally 5% faster vscode build time

jakebailey:
@typescript-bot perf test bun

src/compiler/performance.ts
jakebailey: It turns out that having native perf enabled _is_ very slow. I've made the change to stop doing this on node again (it was probably an oversight) and that caused a very significant perf boost in Node 20.

src/compiler/performanceCore.ts
```diff
@@ -116,6 +98,4 @@ export function tryGetNativePerformanceHooks() {
  *
  * @internal
  */
-export const timestamp = nativePerformance ? () => nativePerformance.now() :
-    Date.now ? Date.now :
-    () => +(new Date());
```
jakebailey, line 121: While here, removing this branch. Date.now is most definitely defined.

src/compiler/performanceCore.ts
```diff
@@ -116,6 +98,4 @@ export function tryGetNativePerformanceHooks() {
  *
  * @internal
  */
-export const timestamp = nativePerformance ? () => nativePerformance.now() :
-    Date.now ? Date.now :
-    () => +(new Date());
+export const timestamp = nativePerformanceTime ? () => nativePerformanceTime.now() : Date.now;
```
jakebailey, line 101: FWIW this must be a closure because Node 19+ require `this` to be correct when calling, so we can't just rip off a method.

jakebailey:
And just to go back to the original Date.now thing, MDN says:

> JavaScript Date times are subject to system clock skew or adjustments. This means that the value of time may not always be monotonically increasing. The main purpose of Date objects is to display time and date information to the user and so many operating systems run a daemon which regularly synchronizes time. It might be that the clock is tweaked a few milliseconds several times per hour.

So, this that probably explains things. Thank you random chance for finding this perf win.

src/compiler/performanceCore.ts
rbuckton: It wasn't guaranteed to be when we targeted ES3, but since we've moved up our minimum target to run tsc this makes sense.

PR #57877, opened 2024-03-21
fix: correct mtime for useFsEventsOnParentDirectory events

### LINKED ISSUE #57792: tsc --watch via directory-level fsevents not working on macOS 14.4
### 🔎 Search Terms

fsevents tsc watch macOS 14.4 not watching working detecting changes found 0 errors

- This changed between versions 5.3.3 and 5.4.2 (also occurs in 5.5 dev latest)

### ⏯ Playground Link

https://gist.github.com/nfarina/d5538d16343830e41c9f3e031a65237e

### 💻 Code

Reproducible with just this `package.json` and an arbitrary `test.ts` file (see Gist link):

```json
{
  "name": "tsc-test",
  "version": "1.0.0",
  "description": "tsc watch problem repro",
  "license": "ISC",
  "scripts": {
    "start": "tsc test.ts --watch"
  },
  "dependencies": {
    "typescript": "^5.4.2"
  }
}
```

### 🙁 Actual behavior

This is confusing to me because it seems like _many_ people would be experiencing this right now, but I have tested quite a lot and, for me, TypeScript 5.4.2 broke folder watching completely on my system. (This is a brand new M3 MacBook Air running macOS 14.4).

This works:

```json
"watchFile": "useFsEvents"
```

And this doesn't:

```json
"watchFile": "useFsEventsOnParentDirectory"
```

On TypeScript 5.3.3 and below, both work as expected.

### 🙂 Expected behavior

`tsc --watch` should observe file changes using fsevents folder watching

I did some digging in the tsc codebase to find where fsevents is being used but was unsuccessful - I'm happy to investigate further if anyone can give me a lead.

llllvvuu:
Fixes #57792

## Problem
On macOS Sonoma, `getModifiedTime(fileOrDirectory)` will return the wrong mtime for file change events in `FileSystemEntryKind.Directory` watchers, since `foo/bar.ts` modification will not bump `foo/` mtime nor will the event itself specify an mtime. This will cause file modifications to not be picked up when running `tsc --watch --watchFile usefseventsonparentdirectory`.

## Solution
Change `getModifiedTime(fileOrDirectory)` to `getModifiedTime(filename)`, where `filename` is calculated by passing through `entryKind`.

## Testing
Provided regression test which fails on `main`. The macOS Sonoma behavior is modeled by the mock filesystem using a new option `modifyFileWithoutModifyDirectory`.

### DIFF
--- a/src/testRunner/unittests/tscWatch/watchEnvironment.ts
+++ b/src/testRunner/unittests/tscWatch/watchEnvironment.ts
@@ -725,7 +725,7 @@ describe("unittests:: tsc-watch:: watchEnvironment:: tsc-watch with different po
         verify(/*fsWatchWithTimestamp*/ false);
     });
 
-    describe("with fsWatch with fsWatchWithTimestamp with useFsEventsOnParentDirectory", () => {
+    describe("with fsWatch with fsWatchWithTimestamp with useFsEventsOnParentDirectory (GH#57877)", () => {
         verifyTscWatch({
             scenario,
             subScenario: "fsWatch/fsWatchWithTimestamp/useFsEventsOnParentDirectory",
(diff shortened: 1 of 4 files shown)


llllvvuu:

src/testRunner/unittests/tscWatch/watchEnvironment.ts
```diff
@@ -725,6 +725,38 @@ describe("unittests:: tsc-watch:: watchEnvironment:: tsc-watch with different po
         verify(/*fsWatchWithTimestamp*/ false);
     });
 
+    describe("with fsWatch with fsWatchWithTimestamp with useFsEventsOnParentDirectory", () => {
```
llllvvuu, line 728: It's actually quite tricky to make the regression test fail on main. In the test environment, the callback is receiving a defined currentModifiedTime, but in an actual project it receives an undefined currentModifiedTime.

sheetalkamat (changes requested):
This isnt the right fix and #57938 addresses the issue instead.

PR #57890, opened 2024-03-21
Apply same single-line style to copied nodes as manufactured ones

weswigham:
I brought this up before, but not applying these emit styles to copied nodes in the node builder makes reusing more input nodes a fairly noisy activity in the test baselines (which might itself be a good thing, but not intentional), and, moreover, makes the types print poorly in error messages (with lots of tabs - `.types` baselines use basically the same print settings errors do). I'd like to take this chance to correct that and apply the node builder's "house style" to any nodes we copy (which is to say: `SingleLine` everything - except object literal types when the node builder flag to print them multiline is set).


tests/baselines/reference/functionDeclarationWithArgumentOfTypeFunctionTypeArray.types
```diff
@@ -2,7 +2,7 @@
 
 === functionDeclarationWithArgumentOfTypeFunctionTypeArray.ts ===
 function foo(args: { (x): number }[]) {
->foo : (args: {    (x: any): number;}[]) => number
+>foo : (args: { (x): number; }[]) => number
```
RyanCavanaugh, line 5: Just commenting on the ones that aren't whitespace only for visiblity

tests/baselines/reference/keyofAndIndexedAccess2.types
```diff
@@ -201,7 +201,7 @@ function f3<K extends string>(a: { [P in K]: number }, b: { [key: string]: numbe
 }
 
 function f3b<K extends string>(a: { [P in K]: number }, b: { [P in string]: number }, k: K) {
->f3b : <K extends string>(a: {    [P in K]: number;}, b: {    [P_1 in string]: number;}, k: K) => void
+>f3b : <K extends string>(a: { [P in K]: number; }, b: { [P in string]: number; }, k: K) => void
```
RyanCavanaugh, line 204: 👀 Do we have a bug re: `P_1` here?

tests/baselines/reference/templateLiteralTypesPatterns.types
```diff
@@ -646,7 +646,7 @@ conversionTest("testDowncast");
 >"testDowncast" : "testDowncast"
 
 function conversionTest2(groupName: | "downcast" | "dataDowncast" | "editingDowncast" | {} & `${string}Downcast`) {}
->conversionTest2 : (groupName: "downcast" | "dataDowncast" | "editingDowncast" | {} & `${string}Downcast`) => void
+>conversionTest2 : (groupName: "downcast" | "dataDowncast" | "editingDowncast" | ({} & `${string}Downcast`)) => void
```
RyanCavanaugh, line 649: Just commenting on the ones that aren't whitespace only for visiblity

tests/baselines/reference/keyofAndIndexedAccess2.types
weswigham: Yes but no - the bug is that P_1 is used at all - both the written types use `P`. The node builder is currently a bit too conservative in it's renaming rules for type parameters. Something I'm working on fixing separately.

tests/baselines/reference/jsdocParseDotDotDotInJSDocFunction.types
```diff
@@ -4,7 +4,7 @@
 // from bcryptjs
 /** @param {function(...[*])} callback */
 function g(callback) {
->g : (callback: (...args: [any][]) => any) => void
+>g : (callback: (...args: [*][]) => any) => void
```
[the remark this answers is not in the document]
weswigham: Thanks, good catch. I had a dummy mistake in the style applying code that was dropping the nested transforms T.T

tests/baselines/reference/templateLiteralTypesPatterns.types
weswigham: I think this is because our parenthesizer when we clone the node is stricter than the syntax allows. I'll have to check.

jakebailey:
I feel like I want this, but the fact that we don't do this at the moment means that we can actually see node reuse changes in other PRs, so I'm sorta thinking that it'd be nice to see what a future reuse indicator might be... But I'm not sure what that will be yet.

weswigham:
> But I'm not sure what that will be yet.

The whitespace only worked as an indicator object literal types - if we want a similar restriction, we can instrument the printer to print `{| ... |}` and `[| ... |]` for object and tuple literal nodes that have non-synthetic positions and `.original` pointers, which'd be somewhat similar to the spans used in fourslash baselines. Another option would be... nearly doubling the size of the `.types` baselines to underline every preserved node? Or completely synthetic ones, whichever direction is preferred.

tests/baselines/reference/templateLiteralTypesPatterns.types
weswigham: Yeah, now that we make a new node to apply `SingleLine` to the `{}`, the node goes through the parenthesizer, and the relevant rule is
```ts
    // UnionType[Extends] :
    //     `|`? IntersectionType[?Extends]
    //     UnionType[?Extends] `|` IntersectionType[?Extends]
    //
    // - A union type constituent has the same precedence as the check type of a conditional type
    function parenthesizeConstituentTypeOfUnionType(type: TypeNode) {
        switch (type.kind) {
            case SyntaxKind.UnionType: // Not strictly necessary, but a union containing a union should have been flattened
            case SyntaxKind.IntersectionType: // Not strictly necessary, but makes generated output more readable and avoids breaks in DT tests
                return factory.createParenthesizedType(type);
        }
        return parenthesizeCheckTypeOfConditionalType(type);
    }
```
which acknowledges it's extraneous.

jakebailey: I do wonder if there's a still a reason to do it; the union case seems like overkill, and I recently changed ExpectType to normalize unions and could probably just make it always strip or add the parens.

weswigham:
@jakebailey I have a draft of a [followup PR](https://github.com/weswigham/TypeScript/pull/69) that explicitly underlines type nodes without positional mappings in the `.types` baselines, like so
```
var a14: (x: { a: string; b: number }) => Object;
>a14 : (x: { a: string; b: number; }) => Object
>    : ^^^^                         ^^^^^^^^^^^
```
that should do much nicer to actually track the desired information, without being too overbearing. The extra linebreaks, IMO, also make the type baselines a smidge more readable.

weswigham:
+600k line diff, though :)

tests/baselines/reference/templateLiteralTypesPatterns.types
jakebailey: I sent https://github.com/microsoft/TypeScript/pull/57900 which shows the affect of this; I do sorta think we should stop doing this as I don't think that the extra parens are that much of an improvement?

But, on its own I think this particular change is alright.

jakebailey (approved):

tests/baselines/reference/templateLiteralTypesPatterns.types
DanielRosenwasser: Personally I found it a bit easier to read with the parens, so consider that a soft vote for keeping the parens.

> I do wonder if there's a still a reason to do it; the union case seems like overkill

Are there cases of non-normalization to consider?

PR #57895, opened 2024-03-21
Infer from type arguments of the same alias in `inferFromObjectTypes`

Andarist:
I noticed that `inferFromObjectTypes` already does this "shortcircuit" between type references but it doesn't do the same in this other case. Both are handled by `inferFromTypes`. The logic related to type references in `inferFromObjectTypes` is *almost* the same as the one in `inferFromTypes` (the latter has one extra condition) - including exactly the same inline comment. Something tells me that, at some point, it was copy-pasted from the other place.

I verified that this added branch is sometimes hit by the existing test suite (not a lot though). I wonder if the perf tests would show anything particularly interesting here (I kinda doubt it though - based on the mentioned infrequest hits here).

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -26187,6 +26187,19 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         }
 
         function inferFromObjectTypes(source: Type, target: Type) {
+            if (source.aliasSymbol && source.aliasSymbol === target.aliasSymbol) {
+                if (source.aliasTypeArguments) {
+                    // Source and target are types originating in the same generic type alias declaration.
+                    // Simply infer from source type arguments to target type arguments, with defaults applied.
+                    const params = getSymbolLinks(source.aliasSymbol).typeParameters!;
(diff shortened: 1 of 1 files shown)


src/compiler/checker.ts
```diff
@@ -26187,6 +26187,19 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         }
 
         function inferFromObjectTypes(source: Type, target: Type) {
+            if (source.aliasSymbol && source.aliasSymbol === target.aliasSymbol) {
```
Andarist, line 26190: this is literally copied-over from `inferFromTypes`

jakebailey:
@typescript-bot test top200
@typescript-bot run dt

Andarist:
@jakebailey could you rerun this on more top repos? I'd love to analyze this failure and I hope that running it on more repos might reveal some test cases that won't use recursive mapped types (it should be easier to analyze such ;p)

jakebailey:
@typescript-bot test top800

PR #54788, opened 2023-06-26
Make `module` control `impliedNodeFormat` and `moduleResolution` control just module resolution

andrewbranch:
Follow-up to #54567. This helps establish a mental model that's easier to document, and gives us more room to fix stuff like #54102.

Issue #55221, opened 2023-07-31
Rethinking `module` for the present and the future


andrewbranch:
# What does `--module` actually mean?

Which of these is a better definition for the flag as it exists today? Which is a better fit for the future?

1. The output module syntax you want to be emitted
2. A declarative description of the module system that will process your emitted code at bundle-time or runtime

When the possoble values of `module` were limited to `amd`, `umd`, `commonjs`, `system`, and `es2015`, the former definition was perfectly fine. When `es2020` and `es2022` were added, which added syntax features like `import.meta` and top-level `await` that couldn’t be transformed into other module emit targets besides `system`, it started to feel like `module` described not just an output format, but the intrinsic capabilities of some external system. With `node16` and `nodenext`, the scope of the `module` flag suddenly expanded to include a new module format detection algorithm used by the target module system and special interop rules between module formats, while it _stopped_ directly controlling the output format, since the format of every output file would be fully determined by Node.js’s format detection algorithm.

The latter interpretation of `module`, one that fully describes the target module system, works well for `node16`/`nodenext`, but trying to project that definition onto the other, older `module` values makes them feel kind of incoherent.

## All the values except `node16`/`nodenext` are kind of weird

Some of the important characteristics of the module system described by `--module nodenext` are:

- Multiple module formats are supported (CJS/ESM)
- The module format of each file is determined via some algorithm
- Modules of different formats interact with each other in specific ways

If we try to infer from existing what the other `module` values say about these characteristics, the result is confusing. For example, you might expect that `--module esnext` means an ESM-only module system that must reject CommonJS/AMD/System modules—after all, you’re not allowed to write `import foo = require("./mod")` in that mode. But you _are_ allowed to import a dependency that _declares_ CommonJS constructs like that.

None of these `module` modes have any restriction on the kinds of modules that can be imported, nor do they particularly make any effort to detect what kind of module a dependency is. Essentially, type checking between modules proceeds as if _everything is CommonJS_, even when we’re explicitly emitting `esnext`. This can be observed direclty by writing a default import of a `.d.ts` file that only declares named exports:

```ts
// @module: esnext
// @esModuleInterop: true

// @Filename: /esm.d.ts
export const x: string;

// @Filename: /main.ts
import esm from "./esm";
esm.x; // string, no error, what??
```

[...]

Issue #54102, opened 2023-05-02
`moduleResolution: bundler` diverges from bundlers following node semantics in regards to `default` import


Andarist:
# Bug Report

### 🔎 Search Terms

moduleResolution bundler node webpack default namespace

### 🕗 Version & Regression Information

- This is the behavior in every version I tried: 5.0.4 (the only one with `moduleResolution: bundler`

### ⏯ Playground Link

Repro case can be found [here](https://github.com/Andarist/import-reexporting-cjs-bundlers/commit/c41f9256cf3153107226664456f667672fcd64ae)

### 🙁 Actual behavior

When resolving a `.d.mts` file that proxies to `.d.ts` in a CJS package with `moduleResolution: bundler`, TypeScript complains with an error like:
> `Property 'default' does not exist on type 'string'.ts(2339)`

The reason that it complains about it is that TypeScript is resolving `default` import as if the `.d.ts` in this CJS package could export the "real" default export. webpack actually follows semantics coined by node and it loads the **namespace** object as the `default` export of that "cjs file".

So the type-level reality diverges there from runtime. At runtime, with webpack - it works exactly like it would work in node. TypeScript thinks that the intention was to load `module.exports.default` there though.

### 🙂 Expected behavior

I'm not exactly sure what's the expected behavior here because "resolving" like a bundler is quite under-specified and bundlers are not quite consistent when it comes to this. Even when we check this very same repro, we can see that I had to use `defaultIsModuleExports: true` in Rollup for it to behave the same. By default, Rollup behaves in the same way as TypeScript does.

cc @andrewbranch

fatcerberus:
Isn’t this controlled by `allowSyntheticDefaultImports`?

andrewbranch:
Yeah, I’m aware of this, but of course it’s not a `moduleResolution` issue. Really, it’s something that `module` should handle. I _believe_ behavior differs between different bundlers, which is, in technical terms, a huge bummer. What’s really bad, IMO, is that last I checked, Webpack does the Node behavior for `.{m,c}js` files, but not the equivalent TS extensions, which violates a pretty core assumption we have about the substitutability of declaration files, TS files, and JS files. This is already on my radar to investigate soon.

andrewbranch:
[...]

PR #57896, opened 2024-03-21
Respect package.json `"type"` and module-format-specific file extensions in more `module` modes

### LINKED ISSUE #50647: Typescript [4.8.2] is adding invalid javascript for *.cjs files
# Bug Report

My project is set to emit ESNext modules. I have a variety of TS files, but I also have a single `.cjs` file, which needs to remain CommonJS to try and require legacy packages.

My input looks like this:

```js
module.exports = {
  tryRequire(modulePath) {
    return require(modulePath);
  },
};
```

I expect the output I get from TypeScript 4.7.4:

```js
module.exports = {
  tryRequire(modulePath) {
    return require(modulePath);
  },
};
```

When I upgrade to TypeScript 4.8.2, it now emits an extra `export {}` at the end of the .cjs file:

```js
module.exports = {
  tryRequire(modulePath) {
    return require(modulePath);
  },
};
export {}
```

This causes Node to fail parsing the `.cjs` file because `export` does not exist in the .cjs context.

Worked in 4.7.4, broken in 4.8.2, also broken in current nightly (4.9.0-dev.20220905).

### Repro here

https://github.com/dzearing/ts-repro-cjs

### LINKED ISSUE #54752: ESM-only packages come with synthesized default in `moduleResolution: bundler`
# Bug Report

### 🔎 Search Terms

module resolution bundler synthesized default

### ⏯ Playground Link

[Playground link with relevant code](https://www.typescriptlang.org/play?target=99&moduleResolution=100#code/PTAEAEFsHsBMFcA2BTASsgztR8AuBLaAOwC5QAjeI2FAJwCgQIYEUzMjkAPXRscXAENaAc2S52GTj3r18kAA7RauUJFiCMuADSgA3qFzQAyrlr4iI0AF9QAM1rRIoAETrNuALR58iT0c8tc0sXAG5QJlwATwVMMgBmADoAJkSABll3LUSjU2CRelyzCwL6AGNiLTUNLWTQAF5QQQB3QXxVeSUVAAo3Gq8fPwCgkpcASnDImLjQJNSM+izcVNhkO0EkXiWVtY3EXByTYstF-tSi-NluLtU9ayA)

### 💻 Code

```ts
// @moduleResolution: bundler
// @module: esnext
// @target: esnext

import mdast, { toString } from "mdast-util-to-string"; // types: 3.2.0

mdast.toString
toString

const mdast2 = await import("mdast-util-to-string"); // types: 3.2.0

mdast2.default
mdast2.default.toString
mdast2.toString

export {}
```

### 🙁 Actual behavior

`default` export is available here and points to the whole module. I don't think this is right, even with `moduleResolution: bundler` but I can be wrong about this since it's really hard to know what kind of assumptions and goals this mode has when it comes to subtle details

### 🙂 Expected behavior

I don't expect `default` to be available here since this ESM-only package **doesn't** have a default export. 

Note that even if this isn't a bug, `.default` available on that dynamically imported binding should behave in the same way in the IDE as the specifier coming from the import statement. Right now the hover info isn't available on it and go to (type?) definition doesn't work on it with cmd+click.

cc @andrewbranch

andrewbranch:
Fixes #54752
Fixes #50647
Closes #54788 
Related #55221
Prerequisite for #54102

Suggest reviewing individual commits:

- 96bffce005c4f7c293e8d36878b19221f4f5a0ef: source changes
- 09330b8d4c663035a1a5ceb350f4dc7a43548b43: tests showing new behavior
- c89137d45d4d1d3afa3da240e178b65d08efe595: noise due to always looking up package.json
- 11e57279fcce0313d1ed4b18a30c4fa2330c34ef: noise due to the same, plus impliedNodeFormat always existing in buildinfo
- f9f46fdbdda67978bde8683d8453ebd765754969: less noisy changes to existing test behavior, worth reviewing

This PR allows us to use module-format-specific file extensions and package.json `"type"` fields in all `module` modes—not just in `node16`/`nodenext` as today—but the way we use them can vary by `module`/`moduleResolution` mode. This has two noticeable effects:

1. Emit respects file extension / package.json `"type"` more often.

   - When `module` is `es2015`–`esnext`, we will now treat `.cts`/`.cjs` files, as well as `.ts`/`.js` files when the nearest package.json has `"type": "commonjs"`, as CommonJS and emit CommonJS outputs accordingly. Note that unlike in `--module nodenext`, a lack of package.json `"type"` does _not_ cause `.ts`/`.js` files to be treated as CommonJS. We only override the default behavior of the `module` setting when `"type"` is explicitly set to `"commonjs"`, or when the file extension is `.cts`/`.cjs`.
   - When `module` is `commonjs`, we will now treat `.mts`/`.mjs` files, as well as `.ts`/`.js` files when the nearest package.json has `"type": "module"`, as ESM and output ESM outputs accordingly.
   - When `module` is `preserve`, ESM import/export declarations are now forbidden in `.cts`/`.cjs` files, as well as in `.ts`/`.js` files when the nearest package.json has `"type": "commonjs"`. We want to ensure that `--module preserve` means the module syntax you write gets emitted without significant transformation, but we also want to prevent ESM syntax from being used in unambiguously CommonJS files, so this situation acts much like `verbatimModuleSyntax` is enabled.

2. Unambiguously ESM-mode declaration files do not get synthesized default exports

   - This is #54752. Now that we always collect information about the package.json `"type"` and file extension, we gain the ability to notice when a declaration file, likely in a node_modules dependency, must represent an ESM JavaScript file, which changes how a default import targeting that file works. Previously, under all `module` modes except `node16`/`nodenext`, this was allowed:

     ```ts
     // @Filename: node_modules/esm-dep/foo.d.mts
     export declare const x: any;
     
     // @Filename: index.ts
     import esm from "esm-dep/foo.mjs";
     esm.x;
     ```

     Since `foo.d.mts` has an unambiguously-ESM file extension, we should not allow a default import from a module that doesn’t declare a default export. However, it was allowed because we believed that `foo.d.mts` might actually represent a CommonJS module with a `module.exports.x` property, in which case a default import should be allowed and would link to the `module.exports` object. With this PR, when a declaration file has an ESM file extension or exists in a `"type": "module"` context, we use that to prevent us from letting a default import happen where it shouldn’t.

This work also unlocks a fix for #54102. Webpack and esbuild use module-format-specific file extensions and package.json `"type"` to vary their ESM/CJS interop rules, so if we want to provide correct types for projects using those bundlers, we will need to be able to leverage that module format info under `--module esnext` or `--module preserve`.

There is one test failing that represents an pre-existing bug that was not exposed until now because the test did not previously look up package.json files, and a particular sequence of edits causes a failure in combination with that. It has been isolated to a failure that doesn’t rely on this PR’s changes in #57757 so it can be investigated separately.


andrewbranch:
@typescript-bot test top400

jakebailey:
> * When `module` is `preserve`, ESM import/export declarations are now forbidden in `.cts`/`.cjs` files, as well as in `.ts`/`.js` files when the nearest package.json has `"type": "commonjs"`. We want to ensure that `--module preserve` means the module syntax you write gets emitted without significant transformation, but we also want to prevent ESM syntax from being used in unambiguously CommonJS files, so this situation acts much like `verbatimModuleSyntax` is enabled.

FWIW this does not match the behavior of esbuild or bun; both accept import statements with those file extensions:

```console
$ cat index.cts
import fs from "fs";

console.log("my code is:");
console.log(fs.readFileSync(__filename, "utf8"));

$ bun run index.cts
my code is:
import fs from "fs";

console.log("my code is:");
console.log(fs.readFileSync(__filename, "utf8"));

$ bunx esbuild --platform=node --bundle index.cts
var __create = Object.create;
var __defProp = Object.defineProperty;
var __getOwnPropDesc = Object.getOwnPropertyDescriptor;
var __getOwnPropNames = Object.getOwnPropertyNames;
var __getProtoOf = Object.getPrototypeOf;
var __hasOwnProp = Object.prototype.hasOwnProperty;
var __markAsModule = (target) => __defProp(target, "__esModule", { value: true });
var __reExport = (target, module2, copyDefault, desc) => {
  if (module2 && typeof module2 === "object" || typeof module2 === "function") {
    for (let key of __getOwnPropNames(module2))
      if (!__hasOwnProp.call(target, key) && (copyDefault || key !== "default"))
        __defProp(target, key, { get: () => module2[key], enumerable: !(desc = __getOwnPropDesc(module2, key)) || desc.enumerable });
  }
  return target;
};
var __toESM = (module2, isNodeMode) => {
  return __reExport(__markAsModule(__defProp(module2 != null ? __create(__getProtoOf(module2)) : {}, "default", !isNodeMode && module2 && module2.__esModule ? { get: () => module2.default, enumerable: true } : { value: module2, enumerable: true })), module2);
};

// index.cts
var import_fs = __toESM(require("fs"));
console.log("my code is:");
console.log(import_fs.default.readFileSync(__filename, "utf8"));
```

I'm not totally sure how widespread this is, though. But it does in general make me nervous about making files stricter based on `type=...` semantics (especially in the mode which vscode will be using for all users?)

andrewbranch:
@jakebailey but if you make those file extensions `.cjs` instead of `.cts`?

jakebailey:
Works the same:

```console
$ bun run index.cjs
my code is:
import fs from "fs";

console.log("my code is:");
console.log(fs.readFileSync(__filename, "utf8"));

$ bunx esbuild --platform=node --bundle index.cjs
var __create = Object.create;
var __defProp = Object.defineProperty;
var __getOwnPropDesc = Object.getOwnPropertyDescriptor;
var __getOwnPropNames = Object.getOwnPropertyNames;
var __getProtoOf = Object.getPrototypeOf;
var __hasOwnProp = Object.prototype.hasOwnProperty;
var __markAsModule = (target) => __defProp(target, "__esModule", { value: true });
var __reExport = (target, module2, copyDefault, desc) => {
  if (module2 && typeof module2 === "object" || typeof module2 === "function") {
    for (let key of __getOwnPropNames(module2))
      if (!__hasOwnProp.call(target, key) && (copyDefault || key !== "default"))
        __defProp(target, key, { get: () => module2[key], enumerable: !(desc = __getOwnPropDesc(module2, key)) || desc.enumerable });
  }
  return target;
};
var __toESM = (module2, isNodeMode) => {
  return __reExport(__markAsModule(__defProp(module2 != null ? __create(__getProtoOf(module2)) : {}, "default", !isNodeMode && module2 && module2.__esModule ? { get: () => module2.default, enumerable: true } : { value: module2, enumerable: true })), module2);
};

// index.cjs
var import_fs = __toESM(require("fs"));
console.log("my code is:");
console.log(import_fs.default.readFileSync(__filename, "utf8"));
```

andrewbranch:
I think it’s ok for us to be more restrictive and opinionated than any single bundler or runtime here. It’s in Bun’s best interest to make plausibly interpretable input code execute without errors. `tsc` outputs can and will often be used across a wide variety of compilers and runtimes, so we have an interest in making sure they’re not footguns. And I don’t see any real reason why someone would _want_ to emit an import into a `.cjs` file.

jakebailey:
Sure, the CJS thing isn't so bad; from my reading this PR doesn't prevent people from using the icky "set module=esnext to emit ESM" trick, so I don't think there are any extra cases I'm concerned about off the top of my head.

src/compiler/transformer.ts
```diff
@@ -77,17 +77,23 @@ import * as performance from "./_namespaces/ts.performance";
...
-            return transformNodeModule;
+        case ModuleKind.CommonJS:
+            // Wraps `transformModule` and `transformECMAScriptModule` and
+            // selects between them based on the `impliedNodeFormat` of the
+            // source file.
+            return transformImpliedNodeFormatDependentModule;
+        case ModuleKind.System:
+            return transformSystemModule;
         default:
             return transformModule;
```
weswigham, line 98: Should we just make this switch exhaustive? This last case is really only for `ModuleKind.None`, right?

src/compiler/utilities.ts
```diff
@@ -8568,17 +8589,76 @@ export function getSetExternalModuleIndicator(options: CompilerOptions): (file:
...
+ * The resolution mode to use for module resolution or module specifier resolution
+ * outside the context of an existing module reference, where
+ * `program.getModeForUsageLocation` should be used instead.
+ */
+export function getDefaultResolutionModeForFile(sourceFile: Pick<SourceFile, "fileName" | "impliedNodeFormat" | "packageJsonScope">, options: CompilerOptions): ResolutionMode {
+    return importSyntaxAffectsModuleResolution(options) ? impliedNodeFormatForEmit(sourceFile, options) : undefined;
+}
+
+/** @internal */
+export function impliedNodeFormatForEmit(sourceFile: Pick<SourceFile, "fileName" | "impliedNodeFormat" | "packageJsonScope">, options: CompilerOptions): ResolutionMode {
```
weswigham, line 8626:
```suggestion
export function getImpliedNodeFormatForEmit(sourceFile: Pick<SourceFile, "fileName" | "impliedNodeFormat" | "packageJsonScope">, options: CompilerOptions): ResolutionMode {

```
? Just by convention?

src/compiler/transformer.ts
andrewbranch: AMD/UMD too

jakebailey (approved):
With #57931 I think this is all good.

src/compiler/watch.ts
```diff
@@ -352,13 +353,14 @@ export function explainFiles(program: Program, write: (s: string) => void) {
     for (const file of program.getSourceFiles()) {
         write(`${toFileName(file, relativeFileName)}`);
         reasons.get(file.path)?.forEach(reason => write(`  ${fileIncludeReasonToDiagnostics(program, reason, relativeFileName).messageText}`));
-        explainIfFileIsRedirectAndImpliedFormat(file, relativeFileName)?.forEach(d => write(`  ${d.messageText}`));
+        explainIfFileIsRedirectAndImpliedFormat(file, program.getCompilerOptions(), relativeFileName)?.forEach(d => write(`  ${d.messageText}`));
```
sheetalkamat, line 356: This needs to be optionsForFile?

src/services/codefixes/importFixes.ts
```diff
@@ -1118,7 +1121,7 @@ function getUmdSymbol(token: Node, checker: TypeChecker): Symbol | undefined {
  * @internal
  */
 export function getImportKind(importingFile: SourceFile, exportKind: ExportKind, compilerOptions: CompilerOptions, forceImportKeyword?: boolean): ImportKind {
-    if (compilerOptions.verbatimModuleSyntax && (getEmitModuleKind(compilerOptions) === ModuleKind.CommonJS || importingFile.impliedNodeFormat === ModuleKind.CommonJS)) {
+    if (compilerOptions.verbatimModuleSyntax && getEmitModuleFormatOfFile(importingFile, compilerOptions) === ModuleKind.CommonJS) {
```
sheetalkamat, line 1124: All usage of options in this file as well need to be resolution options for file?

andrewbranch: I think you’re right, though not related to this PR; this is pretty much a direct translation

andrewbranch: I think I’d like to keep this for now—essentially none of the language service (and very little of the checker) ever looks up project reference options. It makes sense to me that we possibly should, but I think it would be a massive refactor and not related to this PR at all.

tests/baselines/reference/tsbuild/amdModulesWithOut/prepend-reports-deprecation-error.js
```diff
@@ -387,11 +387,46 @@ sourceFile:global.ts
...
-      "./file2.ts": "-13729954175-export const y = 20;",
-      "./global.ts": "1028229885-const globalConst = 10;"
+      "../../lib/lib.d.ts": {
+        "original": {
+          "version": "3858781397-/// <reference no-default-lib=\"true\"/>\ninterface Boolean {}\ninterface Function {}\ninterface CallableFunction {}\ninterface NewableFunction {}\ninterface IArguments {}\ninterface Number { toExponential: any; }\ninterface Object {}\ninterface RegExp {}\ninterface String { charAt: any; }\ninterface Array<T> { length: number; [n: number]: T; }\ninterface ReadonlyArray<T> {}\ndeclare const console: { log(msg: any): void; };",
+          "impliedFormat": 1
+        },
+        "version": "3858781397-/// <reference no-default-lib=\"true\"/>\ninterface Boolean {}\ninterface Function {}\ninterface CallableFunction {}\ninterface NewableFunction {}\ninterface IArguments {}\ninterface Number { toExponential: any; }\ninterface Object {}\ninterface RegExp {}\ninterface String { charAt: any; }\ninterface Array<T> { length: number; [n: number]: T; }\ninterface ReadonlyArray<T> {}\ndeclare const console: { log(msg: any): void; };",
+        "impliedFormat": "commonjs"
```
sheetalkamat, line 396: I think impliedFormat encoding here seems wrong. its not module? we need better logic in builder for checking things as well as not bloating buildinfo with this. It quickly increases when you have lots of files in project like "owa"

src/compiler/checker.ts
```diff
@@ -4076,27 +4080,36 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
 
     function canHaveSyntheticDefault(file: SourceFile | undefined, moduleSymbol: Symbol, dontResolveAlias: boolean, usage: Expression) {
-        const usageMode = file && getUsageModeForExpression(usage);
-        if (file && usageMode !== undefined && ModuleKind.Node16 <= moduleKind && moduleKind <= ModuleKind.NodeNext) {
-            const result = isESMFormatImportImportingCommonjsFormatFile(usageMode, file.impliedNodeFormat);
-            if (usageMode === ModuleKind.ESNext || result) {
-                return result;
+        const usageMode = file && getEmitSyntaxForModuleSpecifierExpression(usage);
+        if (file && usageMode !== undefined) {
+            const targetMode = getImpliedNodeFormatForEmit(file, compilerOptions);
```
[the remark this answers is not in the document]
andrewbranch: This function is supposed to return the module format of the corresponding JS file, whether that file is being emitted by the current TypeScript project/program/compilation. It can return an answer even for a file that isn't going to be part of the current emit. The usage of the word "emit" is intended to draw contrast with how module resolution in that file should work. I think it's fine to move this to the program/host though.

src/services/codefixes/importFixes.ts
andrewbranch: I'm almost through making these changes, but it did occur to me while going through this file that `importingFile` should always belong to the program/options being used... if it belonged to a referenced project, a different LanguageService would have been called.

src/compiler/program.ts
```diff
@@ -834,9 +838,11 @@ export function flattenDiagnosticMessageText(diag: string | DiagnosticMessageCha
  * @internal
  */
 export interface SourceFileImportsList {
-    /** @internal */ imports: SourceFile["imports"];
-    /** @internal */ moduleAugmentations: SourceFile["moduleAugmentations"];
+    imports: SourceFile["imports"];
+    moduleAugmentations: SourceFile["moduleAugmentations"];
     impliedNodeFormat?: ResolutionMode;
```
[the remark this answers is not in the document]
andrewbranch: The meaning of the field hasn't changed; just what we do with it has. So it's impossible to say for API users whether they need to change their usage.

tests/baselines/reference/tsbuild/amdModulesWithOut/prepend-reports-deprecation-error.js
andrewbranch: What would cause it to be `module`? What do you suggest for reducing the buildinfo size? We can keep `impliedNodeFormat` as `undefined` for AMD/UMD, but those are rarely used and we may decide to deprecate them in 6.0 anyway.

sheetalkamat: I just looked into this a bit more and realized that this is anyways encoded as object in tsbuildinfo as it has `affectsGlobalScope` so this should be ok as is.. i will later look into if i can modify how the fileInfo is encoded and see if we can do anything differently but nothing for you to block on.

sheetalkamat (approved):

Issue #57897, opened 2024-03-21
A shadowed variable is generated in a an async function with a default parameter declared as a static property of a class when targeting ES5.


iclanton:
### 🔎 Search Terms

"shadowed", "shadowed variable", "generator"

### 🕗 Version & Regression Information

- This changed in commit or PR #56296

### ⏯ Playground Link

https://www.typescriptlang.org/play?target=1&ts=5.4.3#code/MYGwhgzhAEAa0G8BQ1oAcCuAjEBLY0EALmEftGNALwUQCeAdgQBQQD2AtgKYBqYATgC5oWNmxBcwDatCL8MXAJTUAfIhSpo-LkQz9pYAO5hcROADoszRQG4NAXzsbMOcsVLksMyIxbKqasiaWjp60gBEAOIA8tEAIuF2qI5I9khIACZcoAJc0DlQ6PxswFyFQYQkZARcAB6mzMBsWcIMGBxYXPyKrVwAbl12aUiw5mDWGuZEABZcDI1sDOwS5iBsAOaKk8CkwNPM1qrqwU1L4lzmXcX8zOEAQgCCCbYaqGjFpVCX9UTMAIwvZIvIA

### 💻 Code

```ts
class X {
  public static a = async (someVar: boolean = true) => {
    return await X.b();
  };

  public static b = async () => {
    return "GOOD";
  };
}

declare class process {
  static exit(code: number): never;
}

X.a()
  .then(console.log)
  .catch(() => {
    console.error("BAD");
    process.exit(1);
  });
```

### 🙁 Actual behavior

Between TS v5.4.2 and TS v5.3.3, the generated code from the above code sample changes to shadow the `_a` variable in the outer closure. See diff below:

```diff
var X = /** @class */ (function () {
    function X() {
    }
    var _a;
    _a = X;
-    X.a = function (someVar) {
-        if (someVar === void 0) { someVar = true; }
-        return __awaiter(_a, void 0, void 0, function () {
-            return __generator(_a, function (_b) {
-                switch (_b.label) {
+    X.a = function () {
+        var args_1 = [];
+        for (var _i = 0; _i < arguments.length; _i++) {
+            args_1[_i] = arguments[_i];
+        }
+        return __awaiter(_a, __spreadArray([], args_1, true), void 0, function (someVar) {
+            if (someVar === void 0) { someVar = true; }
+            return __generator(_a, function (_a) {
+                switch (_a.label) {
[...]

PR #57898, opened 2024-03-21
Cherry pick PR 57887 into release-5.4

jakebailey:
See #57897

I have no idea why the baselines are so big in the cherry pick compared to main, but they're all improvements?

### DIFF
--- a/tests/baselines/reference/ParameterList5.types
+++ b/tests/baselines/reference/ParameterList5.types
@@ -2,6 +2,6 @@
 
 === ParameterList5.ts ===
 function A(): (public B) => C {
->A : () => (B: any) => C
+>A : () => (public B) => C
 >B : any
 }
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -6306,7 +6306,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         }
     }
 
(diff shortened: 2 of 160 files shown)


tests/baselines/reference/ParameterList5.types
```diff
@@ -2,6 +2,6 @@
 
 === ParameterList5.ts ===
 function A(): (public B) => C {
->A : () => (B: any) => C
+>A : () => (public B) => C
```
weswigham, line 5: Changes like this one aren't in `main` because #57772 added checks/transforms to prevent them. Here, for example, now we lookup `B` successfully and think "hey, I'll just reuse the node" (which also happens with the `isDeclarationName` fix added in #57772, for this specific case), but the node's got a modifier that needs deleting and a `: any` that needs adding that it doesn't add, while in #57772 I added a line to the visitor to add the missing `: any` and drop the `public` from existing nodes in the node builder. There's a couple changes like this in here. They're not that *bad* a change because it's not like this node's form nowhere - it's the input, so it's at least no worse than the input - but seeing an error in the input (like this) and just dumping the erroneous node into the output is something we usually try to avoid. We could also backport the changes to `visitExistingNodeTreeSymbols` from that PR if we want to minimize the diff.

jakebailey: Does that mean that this PR is wrong without that other PR? Or is this PR safe to include in a patch by itself?

Issue #57908, opened 2024-03-22
Contravariant inference gets picked when a a better covariant candidate exists


Andarist:
### 🔎 Search Terms

covariant covariance contravariant contravariance inference candidates subtyping assignable subtype

### 🕗 Version & Regression Information

- This is the behavior in every version I tried

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.5.0-dev.20240322#code/C4TwDgpgBAShCOBXCBnYAeAKgZQMYAsIBbAQyggA9gIA7AExSj0NID4oBeKAbwCgooSCACcQALig4CxEgG0ARENHyAugG5eAXw29QkJtNKceg5KID8ExDQDWNAPYB3GmqgAje3RCWo1u05oobV5eOghcABsSYWgAM2tcYABLe0Dhe0RqLGYZcipaBgMWElYACns3ACsJPgEwGIlSkgkpYoBKTnYAN3skug0BFENmyRzSAah8EnoIiEaY+Ak4ITRs4dYOjm7e-q02iR6+nVxUtCgukgi+kmpjUoB9GtMRbytbB2dXDy8fPw-AzSbdjcYK8dKZCClWpQepzc6Xa7UAA0-CgQ2KNVRAiU4ig8nkKIEmkJk2mdFm8wQQJ4WKgJxoZ2oaAkaGESRoAHNjAsAHQ4ibEvY6ShgezCYA8YJAA

### 💻 Code

```ts
type Request<TSchema extends Schema> = {
  query: TSchema["query"];
};

type Schema = { query?: unknown; body?: unknown };

declare function route<TSchema extends Schema>(obj: {
  pre: (a: TSchema) => void;
  schema: TSchema;
  handle: (req: Request<TSchema>) => void;
}): void;

const validate = (_: { query?: unknown; body?: unknown }) => {};

route({
  pre: validate,
  schema: {
    query: "",
  },
  handle: (req) => {
    const test: string = req.query;
  },
});

export {};
```

### 🙁 Actual behavior

assignment to `test` fails because `{ query?: unknown; body?: unknown; }` was inferred there

### 🙂 Expected behavior

I'd expect `{ query: string }` to be inferred here and the assignment to `test` to succeed

### Additional information about the issue

This is a reduced version of [this failure](https://github.com/microsoft/TypeScript/pull/54183#issuecomment-2008420338).

PR #54072, opened 2023-04-30
Add fallback when both co- and contra-variant inference candidates exist

ahejlsberg:
In cases where both co- and contra-variant inference candidates exist, this PR adds the ability to fall back to the secondary inference when the constraint check fails for the primary inference.

Fixes #54005.

PR #52180, opened 2023-01-10
Tweak co/contra inference logic from #52123 

jakebailey:
Fixes #52179

In the case of `coAndContraVariantInferences4`, we end up with `Modifier` as the covariant type, but that isn't a subtype of both `Modifier` and `Decorator`, so it shouldn't be chosen; we instead want `Node`, the contravariant inference. But, if you flip things and test for `Decorator` instead, `Modifier` is _still_ the covariant inference, but now it doesn't match `Decorator`, so we get the desired result, `Node`.

This tweaks the logic from #52123 by ensuring that even the _current_ inference's candidates are checked against the inferred covariant type; this effectively removes the problematic choice for the covariant type produced via union ordering.

I'm not an expert; maybe the better win is to change `getCovariantInference` somehow. I just tried this and was surprised that it fixed all of my examples and broke no other test.

PR #52123, opened 2023-01-06
Improve logic that chooses co- vs. contra-variant inferences

ahejlsberg:
Fixes #52111.

PR #57909, opened 2024-03-22
Improve logic that chooses co- vs. contra-variant inferences

Andarist:
To review this it might be helpful to see how this evolved over time:
https://github.com/microsoft/TypeScript/pull/27028
https://github.com/microsoft/TypeScript/pull/46392
https://github.com/microsoft/TypeScript/pull/52123
https://github.com/microsoft/TypeScript/pull/52180
https://github.com/microsoft/TypeScript/pull/54072

The reason why this inference fails today is that `isTypeSubtypeOf` leads to `requireOptionalProperties === true`. This has such inferences:
```ts
inferredCovariantType // { query: string }
inferredContravariantType // { query?: unknown; body?: unknown; }
```

So the covariant inference lacks `body` property and thus it fails the check and the contravariant inference gets chosen at the end.

fixes https://github.com/microsoft/TypeScript/issues/57908
fixes https://github.com/microsoft/TypeScript/issues/58468


jakebailey:
@typescript-bot test top800
@typescript-bot run dt

src/compiler/checker.ts
```diff
@@ -26465,10 +26465,10 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     // and has inferences that would conflict. Otherwise, we prefer the contra-variant inference.
                     const preferCovariantType = inferredCovariantType && (!inferredContravariantType ||
                         !(inferredCovariantType.flags & TypeFlags.Never) &&
-                            some(inference.contraCandidates, t => isTypeSubtypeOf(inferredCovariantType, t)) &&
+                            some(inference.contraCandidates, t => isTypeAssignableTo(inferredCovariantType, t)) &&
```
Andarist, line 26468: If this is not the correct change then at the very least a test case should be proposed that shows how the `isTypeSubtypeOf` is better.

When this check was originally introduced by @ahejlsberg [here](https://github.com/microsoft/TypeScript/pull/27028#discussion_r217112232) it was states that:
> Furthermore, knowing that an error will result when the co-variant inference is not a subtype of the contra-variant inference, we now prefer the contra-variant inference because it is likely to have come from an explicit type annotation on a function. This improves our error reporting.

An alternative idea to fix this example would be to keep `coAndContraRelationCheck` on `inferenceContext`. From what I understand, the `subtypeRelation` in this context is mainly used when dealing with multiple overloads - a single signature case uses `assignableRelation` with `chooseOverload`. So perhaps the used relation should determine what is being used here.

jakebailey:
@typescript-bot pack this

Andarist:
@jakebailey could you rerun top800, dt and perf suites? a new playground would also be appreciated :)

jakebailey:
@typescript-bot test top800
@typescript-bot run dt

@typescript-bot pack this

Andarist:
The only reported error [here](https://github.com/microsoft/TypeScript/pull/57909#issuecomment-2016530535) is actually desired! See the comment [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e1386bb7838c35504edfdb12876bf0d156d84a2a/types/styled-theming/styled-theming-tests.tsx#L71-L72). I didn't look into this one before because of that and because it just involves multiple libraries with complicated type. I plan to reduce it to a test case now though and add it here.

jakebailey:
@typescript-bot test it

Andarist:
Still just one error :p and I have commented on it already [here](https://github.com/microsoft/TypeScript/pull/57909#issuecomment-2016569072)

jakebailey:
Indeed, just rechecking since this is old and also we have new perf benchmarks / stats.

weswigham (approved):
Since the tests are good, I'm fine with this (nice cheeky PR title that copies the title given to two other PRs that have edited this), but I do want to take a moment to this about this function at a conceptual level. The comment in the code does a poor job of this - it's literally restating what the code below does, rather than really expounding on the *why* it's doing what it does.

We're trying to pick between the covariant position inference result and the contravariant position inference result. Strictly speaking, when the results differ, you can easily be in a situation where there simply is no correct inference. For example, if we're inferring from `(x: number) => string` to `(x: T) => T`. Picking either `number` or `string` instead is basically just a choice of which error to produce at the argument site - there is no single `T` that the input type will be assignable to (except `any`). However, if you're a bit lucky, one of the two results will work in both positions and produce a type for which the overall argument is assignable. For example, if you have `(x: number) => 0`, `0` will work for `T` in both positions. So will `number`. Ideally, we'd just note all the possible inferences and try them all to see if any choice results in a valid assignment of the argument, then pick that one. Instead we hem and haw a bit because we don't really want to do all that backtracking (it's [costly](https://github.com/microsoft/TypeScript/pull/57421)), and try to find a local heuristic to pick one or the other for each type argument. And, moreover, when multiple results are valid, determining a useful "ranking" of them for the "best" match is helpful, since users typically have an intent or expectation for how we bias our choices in situations like these.

As of this PR, that ranking is:
1. Covariant result if not never or any, assignable to all contravariant candidates, and the variable being inferred isn't referred to by another variable's constraint directly and every covariant inference candidate is assignable to the chosen covariant inference result
2. Contravariant result, if present
3. Covariant result

Why is this the ranking? Uh... I dunno. Mostly just empirical testing of "this gives good results". I really wish I could point to more rigor here. Certainly, I cannot point to an algebra from which this algorithm arises. Heck, I'd argue it's kinda wrong and bad - in the `(x: number) => 0` example above, it picks `0`, but `number` would be a more reasonable pick. By what metric? Feels. Literals are constraining, and I'd feel you should only pick them as a last resort.

But at least over our current implementation, assignability over subtyping makes sense, since arguments are ultimately compared via assignability and not subtyping (though using the `compareTypes` on the inference context may be better still, since that would allow the subtype overload pass to use subtypes for its heuristic as those arguments *are* compared via subtype, while the second pass can use assignment), and excepting `Any` in the same way as `Never` from the ranking also tracks (since they're basically the same on the source side of a relationship, as is compared in the first step here).

src/compiler/checker.ts
```diff
@@ -26339,12 +26339,14 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     // all co-variant inferences are subtypes of it (i.e. it isn't one of a conflicting set of candidates), it is
                     // a subtype of some contra-variant inference, and no other type parameter is constrained to this type parameter
                     // and has inferences that would conflict. Otherwise, we prefer the contra-variant inference.
```
jakebailey, line 26341: All of the subtype calls are gone; does this prose need to be updated to reflect that?

Andarist: fixed this, thanks for catching this

jakebailey (approved):

jakebailey:
Hm, baselines don't seem to be fully up to date.

Andarist:
@jakebailey fixed that :)

Andarist:
> though using the compareTypes on the inference context may be better still, since that would allow the subtype overload pass to use subtypes for its heuristic as those arguments are compared via subtype, while the second pass can use assignment

I see why you mention this but I failed to create a reasonable test case showing this is better in practice. I could apply this change blindly as it doesn't make a difference for the existing test suite. I also don't think it would surface anything in the extended test suite since that would essentially move the needle closer to the state before this PR for some cases. So I'm hesitant to make this change as I'd prefer to do it with a test case at hand.

At the same time, I see a risk in introducing this change (and funny enough I failed to create a test case showing this too, even a contrived one) since contextual parameter types (cached!) would be assigned based on the proposed subtyped check from the first overload pass. This could potentially yield worse results for them when the secondary pass (using `assignableRelation`) could still succeed.

trevorade:
Should this PR be mentioned in the 5.6 Announcements under Notable Behavioral Changes?

I realize the Beta and RC announcements have already gone out.

PR #29813, opened 2019-02-07
Build .tsbuildinfo file when building using tsc --b

sheetalkamat:
- When tsconfig specifies compiler options `incremental` file with `.tsbuildinfo` extension is emitted.
- Composite projects indirectly specify `incremental` as `true` and its error to specify `incremental` as `false` in the composite project
- If project specifies `tsBuildInfoFile` options, it uses the path specified by that to write the build information.
- If project specifies `out` or `outFile` option, the `outFileWithoutExtension.tsbuildinfo` file is written next to output js file.
- If project specified `outDir`, config files base file name with extension as `.tsbuildinfo` is written in `outDir`
- Otherswise config files base file name with extension as `.tsbuildinfo` is written next to the config file
- Generated file is json with type as `BuildInfo` shown below.
``` ts
    export interface SourceFileInfo {
        // List of helpers in own source files emitted if no prepend is present
        helpers?: string[];
        prologues?: SourceFilePrologueInfo[];
    }

    export interface BundleFileInfo {
        sections: BundleFileSection[];
        sources?: SourceFileInfo;
    }

    export interface BundleBuildInfo {
        js?: BundleFileInfo;
        dts?: BundleFileInfo;
        commonSourceDirectory: string;
        sourceFiles: ReadonlyArray<string>;
    }

    export interface ProgramBuildInfo {
        fileInfos: MapLike<BuilderState.FileInfo>;
        options: CompilerOptions;
        referencedMap?: MapLike<string[]>;
        exportedModulesMap?: MapLike<string[]>;
        semanticDiagnosticsPerFile?: string[];
    }

    export interface BuildInfo {
        bundle?: BundleBuildInfo;
        program?: ProgramBuildInfo;
        version: string;
    }
```
- Field `program` in buildInfo is Builder information so we can track the file changes since that version and build incrementally even without watch (that is emitting only affected files or getting errors only from changed/affected files). This should help us get perf similar to  [--w incremental builds](https://github.com/Microsoft/TypeScript/pull/29161) even on subsequent invokes of build (In short we are serializing information of builder program that we keep in memory during incremental builds in `--w` mode). Also pulled 6c5ae93 to store semantic diagnostics though in `--b` scenario it will always be just file name as emit happens only if there are no errors.
[...]

PR #22658, opened 2018-03-16
Declaration maps and transparent goto definition using them

weswigham:
Adds one command line option: `--declarationMaps`. When enabled alongside `--declaration`, it causes us to emit `.d.ts.map` files alongside the output `.d.ts` files. Services can also now understand these map files, and uses them to map declaration-file based definition locations to their original source, if possible.

Fixes #14479

There's a handful of TODOs that I want to talk about:
* Do we need to recur and map again if the target file is also a declaration file? We won't do this, but tools that process TS may emit sourcemaps that indicate that we should.
* Does the length of the textSpan for `getTargetOfMappedDeclarationFile` need to be mapped somehow?
* Does the textSpan for `getDefinitionAndBoundSpan` need to be mapped in some way?
* Do we need to report some kind of diagnostic (since this is in services, I'm not sure how we would) on invalid source map or silently fail (right now it silently fails)?
* Should we read/emit inline declaration sourcemaps? (right now we do not)
* Should we check a declaration file's `sourceMappingURL` comment? (probably, but right now we do not)
* Should we support sourcemaps represented with the `sections` property (which would likely be beneficial for @RyanCavanaugh 's work)? We currently neither emit nor recognize them.

Additionally, this could still use some more tests exercising more of the many sourcemap options we have with `declarationMaps`. Do we have a better harness for testing things like this (ie, language service features that depend on compilation output)? Fourslash works (see `declarationMapGoToDefinition.ts`), but it's really tedious (and fragile), since I need to include the sourcemap within the test itself (rather than build the map for the test). Should I write a new harness/update an existing one?

PR #57916, opened 2024-03-23
Remove sys.bufferFrom, duplicate Buffer declarations, etc

jakebailey:
Direct use of `Buffer` was added in #22658 for base64. Later, #25659 added `bufferFrom` to work around a bug in Node v5. #29813 added a duplicate declaration of `Buffer` to get the project to compile properly.

These days, none of this appears to be needed. We can just use the global `Buffer` directly and drop all of the duplicate declarations and wrapping.

`BufferEncoding` unfortunately remains due to its use in the tsserver `Session` interface.


jakebailey:

PR #57931, opened 2024-03-25
Fix incorrect use of "path" instead of "resolvedPath" when watching file's package json locations

sheetalkamat:
Fixes issue found at #57757


jakebailey (approved):

andrewbranch (approved):
Thank you! So let me see if I understand this correctly. In the failing test case, a referenced project’s output declaration file gets renamed due to a tsconfig edit, while the input source file keeps the same name. The source file’s `path` was the path to the input file and did not change during this edit, but its `resolvedPath` was the path to the declaration file and did change. So when the test was failing, `existing` had a length of `5`, containing all the package.json locations looked up for the old declaration file name. After your change, `existing` is `emptyArray`, causing all the watched locations for the new declaration file name to be updated or created.

jakebailey:
Seems correct to me; the string is just a map key so its value doesn't seem to matter for other resolution stuff (which seems good).

sheetalkamat:
@andrewbranch exactly what you said.. we were incorrectly using "index.ts" as the location and thats not where the lookups happen so we were incorrectly deciding its just update with last element instead of new locations to watch.

Issue #29651, opened 2019-01-30
Skip typechecking; only emit (support `--transpileOnly` in `tsc`, re-open of #4176)


dtinth:
## Search Terms

isolatedModules, incremental build slow, allowJs, transpileOnly. #4176

## Suggestion

Support a compilation mode where files are only transpiled without typechecking. This can greatly improve compilation speed. Similar to the `transpileOnly` flag in `ts-node` and `ts-loader`.

## Use Cases

At @taskworld, we are trying to migrate our project to TypeScript. We have 1400 source files.

As we try to get our `.js` files to be processed and transpiled by `tsc`, setting `"allowJs": true` makes `tsc` take a painfully long time (40 seconds) to complete, even in `--watch` mode. `tsc --diagnostics` shows that lots of time is spent in typechecking phase.

I have checked these issues:

- https://github.com/Microsoft/TypeScript/issues/7808
  No resolution.

- https://github.com/Microsoft/TypeScript/issues/10018
  Irrelevant — compiler speed has since been improved but that did not solve our problem.

- https://github.com/Microsoft/TypeScript/issues/21221
  I used `tsc --listFiles` to check the files included in the compilation unit. Looks normal. In that issue, they solved the problem by renaming all `.js` to `.ts`, which we can’t do yet without causing errors all over the place, due to our JS files are still CommonJS modules, [which is only recognized when using JS files](https://github.com/Microsoft/TypeScript/wiki/Type-Checking-JavaScript-Files#commonjs-modules-are-supported). I try to migrate with minimal code changes, so I avoid mass-converting CommonJS to ES modules for now.

- https://github.com/Microsoft/TypeScript/issues/13538
  People at Google solved this problem by separating their project into libraries and creates their own build system (based on Bazel). Since I tried to add the tooling with minimal code change, we don’t want to “split our code base into multiple units” right now.

- https://github.com/Microsoft/TypeScript/issues/10157
  OP solved this problem by using `transpileOnly` mode in `ts-loader`, which uses the “single-module transpilation mode” (the `ts.transpileModule` API). However, we aren’t using `webpack` and we’re directly using `tsc` to compile `.ts` files to `.js` files.

- https://github.com/Microsoft/TypeScript/issues/22953
  No resolution.

I tried to profile the `tsc` process, and found that a lot of time is spent in `resolveCallSignature`.

If we can skip the type-checking process, this compilation phase should be faster.

[...]

PR #57934, opened 2024-03-25
Add `noCheck` API option

weswigham:
This PR adds `noCheck`, a compiler option which instructs the compiler to skip the semantic diagnostics phase. This means the compiler skips right form program construction, parsing, and binding, to emit. Implementation-wise, the basically just considers every file the same way we do declaration files under `skipLibCheck`, or project reference redirects.

As-is, there are some limitations on this - declaration emit works wonderfully already - every test with declaration emit on also tests declaration emit with `noCheck`, while JS emit needs some functionality disentangled from the checker's full-tree-error-walk to consistently work with this. As such, `noCheck` requires `emitDeclarationOnly` for now. (This limitation could be lifted with some additional work to make a lazy version of the `.referenced` node links calculations in checker, and by making lazy versions of all the calculations implied by the `getNodeCheckFlags` calls in the JS transforms - which would be required to do #50699 and for this to fulfill #29651) Additionally, passing `noCheck` alongside `noEmit` is an error - there's nothing  _wrong_ with this combination of options, it just seems like it doesn't do anything.

`noCheck` is not currently exposed on the commandline or to tsconfig files, and is an internal API, as there was some concern at the design meeting that it might need to change a bit when we add support for JS emit to it.


src/compiler/checker.ts
```diff
@@ -9038,6 +9038,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             }
 
             function serializeSymbol(symbol: Symbol, isPrivate: boolean, propertyAsAlias: boolean): void {
+                void getPropertiesOfType(getTypeOfSymbol(symbol)); // resolve symbol's type and properties, which should trigger any required merges
```
weswigham, line 9041: The exhaustive testing in the harness revealed that some JS declaration emit cases were a bit broken with out-of-order declaration emit (and were emitting declarations twice) - these two changes fix them by ensuring the checker work to unify weird JS merges is always done before first we look at the symbol's identity (so the merge symbol should always exist if it will exist).

RyanCavanaugh:
````
skipDefaultLibCheck
skipLibCheck
skipCheck
```

weswigham:
But `//@ts-nocheck` and `--noEmit` 🤔

sheetalkamat:
This is not really `--no-check` its just dont report diagnostics right? Because `getEmitResolver` still checks sourceFile and ensures the info?

weswigham:
>This is not really --no-check its just dont report diagnostics right? Because getEmitResolver still checks sourceFile and ensures that file is typechecked

It doesn't request any diagnostics from the checker, so it skips the whole exhaustive check traversal - on our own codebase this skips *basically all* checker time - the bits of the checker declaration emit actually pulls on to produce output are minimal. In theory, yes, it could cascade into a complete program type check. In practice, it never does, for the same reason completions basically never results in a whole program typecheck - the checker's lazy.

weswigham:
Specifically: The `skipTypeChecking` function is used inside `checkSourceFile`, so even though `createEmitResolver` tries to call `getDiagnostics`, with `noCheck` on, it does nothing~

src/compiler/program.ts
```diff
@@ -4357,6 +4357,15 @@ export function createProgram(rootNamesOrOptions: readonly string[] | CreateProg
...
         }
 
+        if (options.noCheck) {
+            if (options.noEmit) {
+                createDiagnosticForOptionName(Diagnostics.Option_0_cannot_be_specified_with_option_1, "noCheck", "noEmit");
+            }
+            if (!options.emitDeclarationOnly) {
+                createDiagnosticForOptionName(Diagnostics.Option_0_cannot_be_specified_without_specifying_option_1, "noCheck", "emitDeclarationOnly");
+            }
+        }
```
sheetalkamat, line 4367: for now we should disable this for incremental as the build info generated will not handle this correctly.

weswigham: As in I should add an error for using `noCheck` alongside `incremental`?

weswigham: > or we need to make sure noCheck recalculates semantic diagnostics and stores the that flag into buildinfo

That's just setting `affectsSemanticDiagnostics` in the option descriptor when we add it back to being enabled for cli/tsconfig in the future, right?

sheetalkamat: "in option" yes but now it needs to be handled in builder or reported as error to have incremental and nocheck as buildinfo written will not be correct. It will show no errors and next time you run tsc it will not report errors . (without storing noCheck value in buildInfo and checking for that in builder)

weswigham: Is [051a1d1](https://github.com/microsoft/TypeScript/pull/57934/commits/051a1d16e87bb418e44d1dd56a1d689cd3982fe0) along the lines of what you were thinking? Since it's not in the `commonOptionsWithBuild`, `commandOptionsWithoutBuild`, or `optionDeclarations` lists, it's never included in `help` or valid on the CLI/in a config, but since it's built into the `OptionsNameMap`, all the normal option serializing/deserializing machinery works with it - build info included.

sheetalkamat: We need test cases: 

1. Run through API with `--no-check`  , followed by `tsc` .(with and without incremental). 
2.  Run through API with `--no-check` , followed by `tsc --b` (with and without incremental) - have a feeling that tsc - in non incremental is going to find that everything is uptodate (there is no buildinfo to check, only output file stamps) and will not report errors? 

You can add these testcases by using `verifyTsc` where the initial `fs` is generated through running API with `--no-check`

weswigham: > Run through API with --no-check

We.... don't have the harness for this right now. At least not a harness that then feeds into the existing `verifyTsc` stuff (obviously, all the `compiler`-style tests build against the API without using the `executeCommandLine` entrypoint). This is quite a lot to put together for something that's only supposed to be API-only temporarily... I started putting it together, and such a harness is looking like it's going to be more than the entire rest of the PR. 😵

So I'm not going to. This is gonna be a normal command line option, accepted on both the CLI and via `tsconfig`, except with a
```ts
        extraValidation(value) {
            return [Diagnostics.Unknown_compiler_option_0, "noCheck"];
        },
```
so anything that runs validation (API methods that take `CompilerOptions` obviously do not) reports that it doesn't exist, but otherwise, it uses all the usual pipelines for things. And then in the test where I'd like to consider it a real option, I'll override that validator with a `noop` temporarily, which I can do for the `tsc -b` style tests, to validate it will work correctly with `-b` once it's actually public and usable via.... anything other than the API.

tests/baselines/reference/deferredLookupTypeResolution.js
```diff
@@ -62,3 +62,29 @@ declare function f3(x: 'a' | 'b'): {
     b: any;
     x: any;
 };
+
+
+!!!! File deferredLookupTypeResolution.d.ts differs from original emit in noCheck emit
```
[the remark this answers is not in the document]
weswigham: Known differences - if you copy these into an actual before/after difftool, you see a union has shifted its' order in the output.

I can shift it to a diff; we already pull in `diff` for the diff-style diffs in the errors we print to the commandline for baseline diffs, so I can just use that to format this section, if you'd prefer. That'll probably break the syntax highlighting for this file even more mightily than the `!!!!` on this line already does, though.

src/compiler/commandLineParser.ts
```diff
@@ -1578,6 +1578,24 @@ export const optionDeclarations: CommandLineOption[] = [
...
+// These aren't available via CLI or config (yet), but we still want to handle them correctly
+// They'll be included in the result of `getOptionsNameMap()`, so they serialize and deserialize correctly,
+// but not in any of the category-specific lists or help menus.
+/** @internal */
+export const privateOptionDeclarations: CommandLineOption[] = [
+    {
+        name: "noCheck",
+        type: "boolean",
+        showInSimplifiedHelpView: true,
+        category: Diagnostics.Emit,
```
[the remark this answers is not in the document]
weswigham: I've actually changed it to a category we don't print in any `help`/`showConfig` output for now (`Compiler_diagnostics`), but yes.

tests/baselines/reference/deferredLookupTypeResolution.js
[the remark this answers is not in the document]
jakebailey: I guess the isolated declarations version of this used a `.diff` file separately instead, but maybe that's not desirable here. I would expect this to largely not happen very much though, so I would sorta prefer a diff even if it breaks things just at the end... but we could always just prefix them with `// `.

src/compiler/program.ts
```diff
@@ -4357,6 +4357,15 @@ export function createProgram(rootNamesOrOptions: readonly string[] | CreateProg
             }
         }
 
+        if (options.noCheck) {
+            if (options.noEmit) {
+                createDiagnosticForOptionName(Diagnostics.Option_0_cannot_be_specified_with_option_1, "noCheck", "noEmit");
+            }
+            if (!options.emitDeclarationOnly) {
+                createDiagnosticForOptionName(Diagnostics.Option_0_cannot_be_specified_without_specifying_option_1, "noCheck", "emitDeclarationOnly");
+            }
```
jakebailey, line 4366: Do you know how much more work it'll take to remove this restriction?

src/compiler/builder.ts
```diff
@@ -451,6 +451,11 @@ function createBuilderProgramState(newProgram: Program, oldState: Readonly<Reusa
                     pendingEmitKind;
             }
         }
+        else if (!canCopySemanticDiagnostics) {
+            // No emit being done, but semantic diagnostics must be recalculated - since these (and the options that change these)
+            // are in the build info file, we need to emit it to keep the info on-disk up-to-date
+            state.buildInfoEmitPending = true;
```
weswigham, line 457: @sheetalkamat When I added a test for `noCheck`, it had a `-inconsistencies` baseline for incremental mode, and the culprit seemed to just be that we didn't place an updated `.buildinfo` on disk after the first build that added semantic diagnostics (since we skipped emit, we also skipped updating the build info). This is probably an existing issue with other semantic-diagnostic-affecting-options, too (if they don't also affect emit), but since the fix seemed simple enough, I put it in here, rather than making a separate issue. This change is what's responsible for the handful of `tsbuild` baseline changes in this PR now - they write a `buildinfo` an extra time or two.

src/compiler/program.ts
weswigham: I have some in-progress work to remove it, so we can prooooobably remove the restriction post-beta (seeings as this is still getting reviewed right now, and I don't think we wanna delay the beta by like a week? Unless we do.). Mostly because reviewing the refactors to the checker to make it happen is going to be annoying, for sure. Making `.references` lazily-calculatable requires taking a lot of logic sprinkled throughout the checker and giving it all a single, central entrypoint (a la `collectLinkedAliases` for the declaration-emit-equivalent elision functionality), which is finicky in terms of preserving the _exact same_ behavior (fun fact: we have many transforms/elisions that apparently just don't apply if a parent node has a grammar error), or at least close enough behavior. Meanwhile, refactoring away from `getNodeCheckFlags` to a `hasNodeCheckFlag` is basically the same thing, but for a handful more uniquely calculated features (though most of them are just variants on "does this node have a child which is X", or "is this node an X within a Y"). A lot of the `NodeCheckFlag` feature-detections could actually be moved into the binder now that we have a name resolver available outside the checker, which is something I want to experiment with.

All this to say: It's definitely doable, and will get done and is being done, but the timeline isn't "one day", and "one week" is a maybe at best, given what's gotta get reviewed.

weswigham: Alternatively: Changing the restriction to "must target esnext and use `verbatimModuleSyntax`" also probably just works.

tests/baselines/reference/deferredLookupTypeResolution.js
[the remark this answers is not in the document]
weswigham: I have changed to diffs, should be easy to see the... diff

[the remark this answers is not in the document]
jakebailey: looking at the diff diff, the diff is diffently better to my eyes

tests/baselines/reference/tsbuild/commandLine/emitDeclarationOnly-false-on-commandline-with-declaration-and-incremental-discrepancies.js
```diff
@@ -115,127 +115,4 @@ IncrementalBuild:
     ]
   },
   "version": "FakeTSVersion"
-}
-TsBuild info text without affectedFilesPendingEmit:: /src/project2/src/tsconfig.tsbuildinfo.readable.baseline.txt::
```
weswigham, line 119: @sheetalkamat As pictured here, apparently the bug where we didn't update the `buildinfo` when we didn't emit but semantic diagnostics may have changed was already captured in a baseline. There's just other stuff here, so it's noisy.

tests/baselines/reference/variadicTuples1.js
```diff
@@ -833,3 +833,21 @@ type U3 = [...[string, number], boolean];
...
+//// [variadicTuples1.d.ts]
+===================================================================
+--- Expected	The full check baseline
++++ Actual	with noCheck set
+@@ -17,9 +17,9 @@
+ declare const tc2: [string, number];
+ declare const tc3: [number, number, number, ...string[]];
+ declare const tc4: [...string[], number, number, number];
+ declare function concat2<T extends readonly unknown[], U extends readonly unknown[]>(t: T, u: U): (T[number] | U[number])[];
+-declare const tc5: (2 | 4 | 1 | 3 | 6 | 5)[];
```
weswigham, line 848: Aside: I find it insane that the full typecheck ordered this `2 | 4 | 1 | 3 | 6 | 5` for _reasons_.

We should *really* revisit stable sorting of literal types at least at some point.

src/compiler/program.ts
weswigham: For context, here's the diffs for that: https://github.com/weswigham/TypeScript/commit/64dddfbc0d57e2d986bdeefa475916738267a9b9

Very easy to enable at least *some* js emit targets in that regard (and you can see one baseline actually does differ, but it's an error case for `verbatimModuleSyntax` that's a reexport of a type-only import, so it's what you'd expect to be broken, given the error). @dragomirtitian said this was probably all they wanted, since it's all you need to strip away TS syntax. 🤷‍♂️ 

I, personally, don't want such a limitation though, since the limitation is basically just tech debt on our part.

jakebailey: I think I'd rather skip JS emit in this PR than to have a more complicated set of flags to enable; so, all good as-is.

jakebailey (approved):

src/testRunner/unittests/tsbuild/noCheck.ts
```diff
@@ -0,0 +1,86 @@
...
+function verifyNoCheckFlag(variant: string) {
+    function verifyNoCheckWorker(subScenario: string, declAText: string, commandLineArgs: readonly string[]) {
+        verifyTsc({
+            scenario: variant,
+            subScenario,
+            fs: () =>
+                loadProjectFromFiles({
+                    "/src/a.ts": getATsContent(declAText),
+                    "/src/tsconfig.json": jsonToReadableText({
+                        compilerOptions: { noCheck: true, emitDeclarationOnly: true, declaration: true },
```
sheetalkamat, line 21: Having this in json and not reporting error is a bug because then it doesnt become API only option per design meeting.,

The scenario i am talking is missed here because:
modifying tsconfig.json will check that config file is newer than output and hence program will be rebuilt. The problem with doing this with API is that output will be updated but not inputs which means the build will report that everything is uptodate

Here is how to write this test case
1 load project files (without noCheck option) in your test case as is . No errors
2. As part of first edit: Do these steps to emulate API usage:  Change a.ts to introduce error, run API (by creating program and emitting with noCheck on, use `sys= new vfs.sys(fs))` to create compiler host  
3. This will fail because after edit, test framework will run `tsc --build` again and it will find that everything is UpToDate and not report error.  (test will not create discrepancy baseline because unfortunately we don't check discrepancies in error reporting just file emit validation)

sheetalkamat (changes requested):

src/testRunner/unittests/tsbuild/noCheck.ts
weswigham: > Having this in json and not reporting error is a bug because then it doesnt become API only option per design meeting.,

And it *does* _normally_ report an error, which is what the other test block that doesn't override the validator function verifies (and naturally, these collapse into one test once it's actually exposed via CLI)~

> The problem with doing this with API is that output will be updated but not inputs which means the build will report that everything is uptodate

If you build via API and pass options not in your on-disk config file (as you would need to, since `noCheck` is API only right now), your up-to-dateness checks for configs are going to *all* be invalidated anyway, since your API-fueled build will have different options applied than a commandline one. Precisely

> The problem with doing this with API is that output will be updated but not inputs which means the build will report that everything is uptodate

is in no way unique to this new flag - any time you do a API build with different flags than what your on-disk configs specify this'll happen, and I'm pretty sure how `-b` handles that is just entirely out of scope, unless we're talking specifically `incremental` output, which just comes down to the options serialized in the `.buildinfo` which is why:

> This will fail because after edit, test framework will run tsc --build again and it will find that everything is UpToDate and not report error. (test will not create discrepancy baseline because unfortunately we don't check discrepancies in error reporting just file emit validation)

It both will and won't. If `--incremental` isn't on, yeah, it won't build again because all the timestamps say it's newer, and `build` doesn't know that your last build was with different settings. Again, that's pretty out-of-scope and isn't going to get fixed, since there's no incremental artifact. Nothing new there, that issue has nothing to do with `noCheck` - you can replicate that with any semantic diagnostic affecting option. For `--incremental`, that *won't* be the case, because the `.buildinfo` emitted by the API build will have different compiler options than the new build (semantic diagnostic affecting ones), which'll force a recheck (though not reemit), and is _precisely_ what this test verifies without needing all the API-layering hoopla by temporarily enabling it in configs/cli (since an incremental build via config is going to produce the same buildinfo as one via api). AFAIK, the only critical thing to test is that `noCheck` is serialized into the `buildinfo` just like every other semantic diagnostic-affecting option, when it's allowed, which this test verifies.

As an aside "via API" is ambiguous, since we expose `executeCommandLine` now, but I've taken that to mean "only allowed where we take `CompilerOptions`-type objects" for now, which just means any API that's post-option-parsing/validating. Absolutely none of those APIs (eg, `createProgram`) are even going to emit a `.buildinfo` file on their own without extra instrumentation on behalf of the caller of the API, so it is quite literally up to the API user to keep things up-to-date, provided our `emitBuildInfo` function does serialize the option for them when they call it to do that, which, again, this test does verify.

src/compiler/program.ts
```diff
@@ -4357,6 +4357,15 @@ export function createProgram(rootNamesOrOptions: readonly string[] | CreateProg
             }
         }
 
+        if (options.noCheck) {
+            if (options.noEmit) {
```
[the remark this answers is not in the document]
weswigham: `noEmitOnError` blocks on call of `options`, `syntactic`, global`, `semantic`, and `declaration` diagnostics. `noCheck` only blocks `semantic` from being generated, so `noEmitOnError` does still have meaning alongside `noCheck`, though it certainly will block emit _less_.

src/compiler/builder.ts
sheetalkamat, line 457: This should be set only if (!outputFile)

src/compiler/emitter.ts
```diff
@@ -848,8 +848,8 @@ export function emitFiles(resolver: EmitResolver, host: EmitHost, targetSourceFi
         const filesForEmit = forceDtsEmit ? sourceFiles : filter(sourceFiles, isSourceFileNotJson);
         // Setup and perform the transformation to retrieve declarations from the input files
         const inputListOrBundle = compilerOptions.outFile ? [factory.createBundle(filesForEmit)] : filesForEmit;
-        if (emitOnly && !getEmitDeclarations(compilerOptions)) {
-            // Checker wont collect the linked aliases since thats only done when declaration is enabled.
+        if ((emitOnly && !getEmitDeclarations(compilerOptions)) || compilerOptions.noCheck) {
```
sheetalkamat, line 851: Also need to add forceDtsEmit check here? shouldnt signature generation behave like `noCheck`?

src/compiler/utilities.ts
```diff
@@ -9938,6 +9938,7 @@ export function skipTypeChecking(sourceFile: SourceFile, options: CompilerOption
     // '/// <reference no-default-lib="true"/>' directive.
     return (options.skipLibCheck && sourceFile.isDeclarationFile ||
         options.skipDefaultLibCheck && sourceFile.hasNoDefaultLib) ||
+        options.noCheck ||
```
sheetalkamat, line 9941: Adding this here also skips grammar checks .. May be i am remembering it incorrectly but i thought we wanted to do grammar checks ?

tests/baselines/reference/tsbuild/commandLine/emitDeclarationOnly-false-on-commandline-with-declaration-and-incremental-discrepancies.js
sheetalkamat: This is incorrectly emitting buildInfo because "emitDeclarationsOnly" does not affect semantic diagnostics. The condition in builder to check for this is: 
```ts
 const canCopySemanticDiagnostics = useOldState && oldState!.semanticDiagnosticsPerFile && !!state.semanticDiagnosticsPerFile &&
        !compilerOptionsAffectSemanticDiagnostics(compilerOptions, oldCompilerOptions!);
```

where as you want it to be `compilerOptionsAffectSemanticDiagnostics(compilerOptions, oldCompilerOptions!);` when emitting buildInfo again. 

I think you just want to check if "noCheck" is changed and then mark the buildInfo emit pending.

src/compiler/builder.ts
sheetalkamat: As mentioned somewhere below, you want this condition to be `useOldState && !!oldState!.compilerOptions.noCheck !== !!compilerOptions.noCheck`

src/testRunner/unittests/tsbuild/noCheck.ts
sheetalkamat: The problem with writing this test case this way is that it is modifying timestamp of tsconfig files and hence build will take place. To check this correctly, you want to use API emit as `edit` as mentioned above.

sheetalkamat (changes requested):

src/compiler/utilities.ts
weswigham: > Adding this here also skips grammar checks

Not... quite. More in a bit.

> May be i am remembering it incorrectly but i thought we wanted to do grammar checks ? 

I wanna say "no". We ruminated on it briefly, but "grammar checks" aren't a defined subset of the diagnostics we publicly recognize (and some things you'd think are in that category are emitted in the parser, binder, or checker passes over the AST - roll a d6 to find out where), so while are somewhat well-defined in ECMA-262, are incredibly poorly defined for us. Additionally, what diagnostics we skip vs emit isn't in any way what's important for this flag, what _is_ is that we skip the checker's full tree walk (and all the work it does) for diagnostics and jump right to emit (which is what inclusion in this function does). Doing a walk specifically for grammar errors would run pretty counter to that goal, though I admit if you're in a checker context and already traversing an AST most grammar-style errors are usually trivial to emit. Again, not terribly important for this, though - we just want to skip the big walk that does all the expensive work. (Usually so it can be done asynchronously.)

tests/baselines/reference/tsbuild/commandLine/emitDeclarationOnly-false-on-commandline-with-declaration-and-incremental-discrepancies.js
weswigham: > I think you just want to check if "noCheck" is changed and then mark the buildInfo emit pending.

How is `noCheck` any different than any other flag that reduces the number of errors produced? Like toggling `noImplicitAny` to `false`? I'm not seeing why anything should be `noCheck`-specific here.

src/testRunner/unittests/tsbuild/noCheck.ts
weswigham: Or I could allow `--noCheck` as a `build` option (why not? seems like a reasonable thing to want to turn on/off for the whole project graph) and pass it down from the top-level CLI invocation, rather than by editing the `tsconfig`?

src/compiler/builder.ts
weswigham: I have no clue why this would be `noCheck`-specific, though? `noCheck` seems to me to be no different from any other compiler option that varies the errors we produce.

src/compiler/emitter.ts
weswigham: AFAIK it already does - the `|| compilerOptions.noCheck` being tacked on here is so that `noCheck` builds do the same data collection signature generation runs already do.

weswigham:
Per our last design meeting discussion on this, this PR enables `noCheck` for `transpileDeclaration` by default. On this workstation, without `noCheck`, a simple `transpileDeclaration` on each file in our `src` directory (single thread, I/O time not included) takes 27144ms, with `noCheck`, that is cut to 8404ms. A *substantial* improvement, considering semantic diagnostics aren't even available via the `transpile` APIs!

weswigham:
@sheetalkamat Since the option isn't available on the CLI yet, nor does it have any special behavior W.R.T. build mode compared to any other semantic diagnostics, I'm not sure there's much to do in this PR. As we said at the last design meeting we likely want to clean up how build mode handles changing flags that aren't from tsconfig files before we make this available via CLI (since a primary usecase is toggling it for a whole project graph), but that's not something that's going to happen in this PR, nor likely for 5.5. I could revert the [builder.ts](https://github.com/microsoft/TypeScript/pull/57934/files#diff-75990196aa2d9dff50be349193d8df248ccc31e33fd36c39c8f1bb9b0e1412f8) change here just so we can keep all that cleanup together, if you'd like.

robpalme:
The new baselines show that enabling `noCheck` can re-order union items compared to checked declaration emit.  Do we know of any other declaration emit differences?

tests/baselines/reference/outModuleConcatES6.js
```diff
@@ -0,0 +1,22 @@
...
+//// [a.ts]
+export class A { }
+
+//// [b.ts]
+import {A} from "./ref/a";
+export class B extends A { }
+
+
+
+!!!! File all.d.ts missing from original emit, but present in noCheck emit
```
jakebailey, line 12: In general, how are we ending up with these messages? Are they all cases where we don't emit due to errors?

tests/baselines/reference/tsbuild/configFileErrors/reports-syntax-errors-in-config-file-discrepancies.js
```diff
@@ -1,85 +1,4 @@
...
-    "changeFileSet": [
-      "../lib/lib.d.ts",
-      "./a.ts",
-      "./b.ts"
-    ]
-  },
-  "version": "FakeTSVersion"
-}
\ No newline at end of file
+*** Supplied discrepancy explanation but didnt file any difference
```
jakebailey, line 4: This seems like it needs fixing?

jakebailey: Ah, there are other files with this (and other threads), forgot.

tests/baselines/reference/outModuleConcatES6.js
weswigham: Yea, a checker error blocks the .d.ts emit without `noCheck` on.

weswigham:
> The new baselines show that enabling noCheck can re-order union items compared to checked declaration emit. Do we know of any other declaration emit differences?

Yeah, this can happen in `incremental` mode, too. It's a pretty well-known issue that we haven't found a great way to fix, but usually doesn't have semantic implications. Other stuff that depends on check order is regarded as a bug, since it could make diagnostics in the editor inconsistent.

tests/baselines/reference/tsbuild/configFileErrors/reports-syntax-errors-in-config-file-discrepancies.js
weswigham: I've removed the small builder change in this PR - we can look into it separately. The issue there isn't related to the new  `noCheck` flag, but rather applies to any semantic-diagnostic-affecting option you build-via-api with a different setting than you build-via-cli with, which, as we discussed, needs some more extensive builder changes to properly support (and to support passing builder-wide semantic settings via CLI generally).

src/compiler/commandLineParser.ts
```diff
@@ -772,6 +772,20 @@ const commandOptionsWithoutBuild: CommandLineOption[] = [
...
+        type: "boolean",
+        showInSimplifiedHelpView: false,
+        category: Diagnostics.Compiler_Diagnostics,
+        description: Diagnostics.Disable_full_type_checking_only_critical_parse_and_emit_errors_will_be_reported,
+        transpileOptionValue: undefined,
+        defaultValueDescription: false,
+        affectsSemanticDiagnostics: true,
+        affectsBuildInfo: true,
+        extraValidation() {
+            return [Diagnostics.Unknown_compiler_option_0, "noCheck"];
```
weswigham, line 786: For anyone reading, this validation is what blocks this option from being passed on the CLI. This is a nicer way to do it than our usual method (just adding an entry to the `CompilerOptions` interface), since it still provides metadata for `.buildinfo` serialization and `showConfig`, plus makes it easier to test as though it were CLI-allowed, like is done in the `noCheck` unittest.

sheetalkamat:
@weswigham Will review soon probably tonight or tomrorow morning, but i think we need to make sure buildInfo is not written when "noCheck" is on accidently. This can be done by changing "getTsBuildInfoEmitOutputFilePath" to return undefined if noCheck is true.

weswigham:
> but i think we need to make sure buildInfo is not written when "noCheck" is on accidently. This can be done by changing "getTsBuildInfoEmitOutputFilePath" to return undefined if noCheck is true.

`noCheck` isn't any different from any other option that affects checker diagnostics, I don't see why we'd do a one-off fix for it that doesn't apply to every other semantic-diagnostic-affecting option. `noCheck` isn't in any way special, other than it *always* adding an option parsing diagnostic when you pass it through the command line helper.

weswigham:
For additional context, `allowNonTsExtensions` and `suppressOutputPathCheck` are other internal-only compiler option that *definitely* affect diagnostics, which we don't do any special handling for _at all_. I assume our approach for internal only options is "use at your own risk", since it's not like they can be passed through normal build/cli/config pipelines.

And `noCheck` shouldn't even be internal only for long. 🤷‍♂️

weswigham:
Per out-of-band discussions, I've opened #58336 to formally track future work with `--build` and `--noCheck` to help clarify the long-term goals for this. As-is, I'm going to merge this, and any other changes we want to make can be taken as follow-ups.

zanminkian:
Will this land in 5.5?

PR #57936, opened 2024-03-25
Test for tsc --watch to stub out actual time

sheetalkamat:
We already changed this in tsserver but missed for tsc --watch where time is printed.
Any change that affects how modified times are done needs to update lots of baseline just because time changes without this change .


jakebailey (approved):
LGTM but we'll definitely have to be careful merging newer PRs

PR #57938, opened 2024-03-25
Directories dont check modified time when sending "change" event

sheetalkamat:
When checking whether to send event invoked with `fs.watch` we shouldnt do that for directory. Directory can send `change` event for multiple reasons and its not right to filter that out.

a7a308d1d7dce2be48edb0f7688ab5c3fe89da1b Adds the failing test. It modifies out vfs sending incorrect timestamps, incorrect events etc. 
88c5588181e0340b41f8a1a1c87b27bada3de610 actual fix

Fixes https://github.com/microsoft/TypeScript/issues/57792

### DIFF
--- a/src/compiler/sys.ts
+++ b/src/compiler/sys.ts
@@ -378,16 +378,24 @@ function createDynamicPriorityPollingWatchFile(host: {
     }
 }
 
-function createUseFsEventsOnParentDirectoryWatchFile(fsWatch: FsWatch, useCaseSensitiveFileNames: boolean): HostWatchFile {
+function createUseFsEventsOnParentDirectoryWatchFile(
+    fsWatch: FsWatch,
+    useCaseSensitiveFileNames: boolean,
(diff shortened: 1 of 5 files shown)


jakebailey (approved):
Does this need a backport to 5.4?

sheetalkamat:
@DanielRosenwasser @RyanCavanaugh  do we want this for 5.4 ? If we need to port this, there will be test breaks unless we take #57936

jakebailey:
The cherry pick task will fix up the baselines, so that all would be handled (but we can still manually fix it)

NMinhNguyen:
Just to check - is this going to be backported to 5.4 or only 5.5?

jakebailey:
I think it's worth considering.

NMinhNguyen:
Apologies for the direct mention @jakebailey, but is a new 5.4 version still planned to go out at some point?

jakebailey:
This was in 5.4.4 which came out last week.

PR #57948, opened 2024-03-26
Correct the type of return value of FinalizationRegistry.unregister

### LINKED ISSUE #57919: FinalizationRegistry.unregister should return boolean instead of void
### ⚙ Compilation target

ES2021

### ⚙ Library

es2021.weakref.d.ts

`FinalizationRegistry.unregister` is now defined as:

```typescript
    unregister(unregisterToken: WeakKey): void;
```

But it should return a boolean value that is `true` if at least one cell was unregistered and `false` if no cell was unregistered.

### Sample Code

```TypeScript
const obj = {};
const registry = new FinalizationRegistry(() => {});
registry.register(obj, obj, obj);
const removed = registry.unregister(obj); // failed compilation here
```

### Documentation Link

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry/unregister

jameslan:
Fixes #57919

### DIFF
--- a/src/lib/es2021.weakref.d.ts
+++ b/src/lib/es2021.weakref.d.ts
@@ -42,7 +42,7 @@ interface FinalizationRegistry<T> {
      * @param unregisterToken The token that was used as the unregisterToken argument when calling
      * register to register the target value.
      */
-    unregister(unregisterToken: WeakKey): void;
+    unregister(unregisterToken: WeakKey): boolean;
 }
 
 interface FinalizationRegistryConstructor {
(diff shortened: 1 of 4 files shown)


jameslan:

jakebailey:
@typescript-bot test it

jameslan:
@jakebailey Merge conflicts were solved, please review. Thanks!

jakebailey (approved):
This seems correct to me, but curious if @sandersn can see anything that may break here. Probably nothing based on the testing.

sandersn:
1. I can't think of anything that would break here except, *maybe* passing unregister itself around.
2. I doubt unregister is used that widely, so it's even less likely to cause a break.

sandersn (approved):

PR #57950, opened 2024-03-26
Watch events enhancements

sheetalkamat:
- adds ignoreUpdate to directory watcher so we are not unnecessarily watching for update events if we dont care
- Change protocol to get batched watch change notifications

Vscode PR reference: https://github.com/microsoft/vscode/pull/193848

cc: @bpasero

### DIFF
--- a/src/server/editorServices.ts
+++ b/src/server/editorServices.ts
@@ -972,34 +972,32 @@ function createWatchFactoryHostUsingWatchEvents(service: ProjectService, canUseW
             },
         };
     }
-    function onWatchChange({ id, path, eventType }: protocol.WatchChangeRequestArgs) {
-        onFileWatcherCallback(id, path, eventType);
-        onDirectoryWatcherCallback(watchedDirectories, id, path);
-        onDirectoryWatcherCallback(watchedDirectoriesRecursive, id, path);
+    function onWatchChange(args: protocol.WatchChangeRequestArgs | readonly protocol.WatchChangeRequestArgs[]) {
+        if (isArray(args)) args.forEach(onWatchChangeRequestArgs);
+        else onWatchChangeRequestArgs(args);
     }
 
-    function onFileWatcherCallback(
-        id: number,
-        eventPath: string,
-        eventType: "create" | "delete" | "update",
-    ) {
-        watchedFiles.idToCallbacks.get(id)?.forEach(callback => {
-            const eventKind = eventType === "create" ?
-                FileWatcherEventKind.Created :
-                eventType === "delete" ?
-                FileWatcherEventKind.Deleted :
-                FileWatcherEventKind.Changed;
-            callback(eventPath, eventKind);
-        });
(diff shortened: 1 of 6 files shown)


src/server/editorServices.ts
```diff
@@ -933,7 +933,16 @@ function createWatchFactoryHostUsingWatchEvents(service: ProjectService, canUseW
...
             callback,
-            id => ({ eventName: CreateDirectoryWatcherEvent, data: { id, path, recursive: !!recursive } }),
+            id => ({
+                eventName: CreateDirectoryWatcherEvent,
+                data: {
+                    id,
+                    path,
+                    recursive: !!recursive,
+                    // Special case node_modules as we watch it for changes to closed script infos as well
+                    ignoreUpdate: !path.endsWith("/node_modules") ? true : undefined,
```
jakebailey, line 943: Can this path be `\node_modules` on Windows?

bpasero: Slightly related: I did notice another watch request for `node_modules/@types`, so I wonder if these could also `ignoreUpdate`?

bpasero: Hm, I think I misunderstood our discussion around changes in `node_modules` then. I had assumed TS only cares about added/deleted events in `node_modules` to rerun module resolution, but not for updates. So now with your change, folder watchers ignore updates unless its `node_modules` being watched?

sheetalkamat: Yes. we only care about `updates` inside `node_modules` and no other folder.

sheetalkamat: Reference to code here https://github.com/microsoft/TypeScript/blob/main/src/server/editorServices.ts#L3102 this normally works out because with fs.watch most of the time we get "rename" event instead of "change" when doing `npm i` but with parcel you get `update` so that cannot be ignored from vscode watcher usage perspective

bpasero: Yeah I understand. My idea was to collect events to `node_modules` in the TS extension and then send you a single event for the module. That can be a `added` event to emulate the `rename` you got from `fs.watch`. Its a bit of a hack and very specific to this case but would essentially reduce the event load on `node_modules`, which can be a huge source of file events.

sheetalkamat: What i meant by `ignoreUpdate` here is that we really dont care if there is file that was modified within that folder and thats true for any folder watched except `node_modules`. So from protocol wise this change looks good. 

On whether to send single `directory event` for `module` inside `node_modules` or all files can be totally customized irrespective of this protocol change. In general, the more specific you are about the `path` the better reuse happens between programs. We were anyways getting these events from `fs.watch` and we handle both cases since the fileName may or may not be present when the event comes through `fs.watch`.

jakebailey (approved):

DanielRosenwasser:

DanielRosenwasser (approved):

DanielRosenwasser (approved):

PR #57955, opened 2024-03-27
Refactor expression evaluator

dragomirtitian:
Refactor expression evaluator to sit outside of the checker.

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -109,6 +109,7 @@ import {
     createDiagnosticForNodeFromMessageChain,
     createDiagnosticMessageChainFromDiagnostic,
     createEmptyExports,
+    createEvaluator,
     createFileDiagnostic,
     createGetCanonicalFileName,
     createGetSymbolWalker,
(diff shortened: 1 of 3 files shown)


src/compiler/checker.ts
```diff
@@ -1474,6 +1475,11 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     var checkBinaryExpression = createCheckBinaryExpression();
     var emitResolver = createResolver();
     var nodeBuilder = createNodeBuilder();
+    var evaluate = createEvaluator({
+        evaluateElementAccessExpression,
+        evaluateEntityNameExpression,
+        onNumericLiteral: checkGrammarNumericLiteral,
```
jakebailey, line 1481: It feels strange to have this callback. This only needs the AST node and nothing else. Can we just walk the nodes like other grammar checks?

dragomirtitian: I was trying to introduce as little extra overhead as possible, I agree that callback is kinds of the odd one out, but without an extra walk I'm not sure how we could ensure that check is performed in the same way. 

The check does appear to happen in other places though. Enum members should go through `checkEnumMember` which will call `checkGrammarNumericLiteral` during expression checking. Similarly for template literals. So the check might be redundant for a full check. I'll test and see.

dragomirtitian: Just tested .. removing the check from here does not fail any test 😕. I do think it ends up being called through other paths.

weswigham (approved):
Just a philosophy question here: `utilities.ts` is a now 10,000 line long file of what is essentially uncategorized helper functions. Since we refer to it as "the expression evaluator", does that imply it's standalone enough that it should warrant its' own file, rather than getting tossed in alongside the other uncategorized helper functions?

DanielRosenwasser:
> does that imply it's standalone enough that it should warrant its' own file, rather than getting tossed in alongside the other uncategorized helper functions?

Given how small it is and how often new files get created, I think it's probably fine in `utilities.ts` - at least for now.

PR #57960, opened 2024-03-27
preserve declarations if @internal is mentioned in unrelated comment

### LINKED ISSUE #57352: Class member incorrectly stripped from .d.ts output if @internal is mentioned in unrelated comment
### 🔎 Search Terms

internal, .d.ts, stripInternal

This issue appears to be present in all version of TypeScript (tested down to 3.3.3).

### ⏯ Playground Link

https://www.typescriptlang.org/play?stripInternal=true&ts=5.3.3#code/KYDwDg9gTgLgBAYwDYEMDOa4DEITgbwCg44B6AKnOJPLgGUALCAVyQBM4AjYONGKAJZgwwNtTi0AAgIB2MYFBkok48qWpomrNgCFgdfkJEcAvHACMAbkLVSpOABUA8gBEnALjgBbFAE9u3igA1jwwDAKY0nIKSkgA-LaUqvRa7BIyEDC0AXyCwqKq6iSaLOwAcpl6BnnGcGYATNYAvkA

### 💻 Code

```ts
export class Foo {
  /**
   * Should be stripped
   * @internal
   */
  shouldBeStripped = 1;

  // TODO: maybe make this @internal?
  /**
   * Should *not* be stripped
   */
  shouldNotBeStripped = 2;
}
```

### 🙁 Actual behavior

The compiler should preserve the `Foo.shouldNotBeStripped` property while removing `Foo.shouldBeStripped`.

### 🙂 Expected behavior

Both `Foo.shouldNotBeStripped` and `Foo.shouldBeStripped` are removed.

If a class member has an unrelated comment mentioning `@internal` preceding it, TypeScript will drop that member from the .d.ts output, even though that's not the intent since there's another comment after it. Only the first comment should be checked for `@internal`.

Zzzen:
Fixes #57352


src/compiler/utilitiesPublic.ts
```diff
@@ -2596,31 +2594,8 @@ export function isRestParameter(node: ParameterDeclaration | JSDocParameterTag):
...
-                getLeadingCommentRanges(text, node.pos),
-            )
-            : getTrailingCommentRanges(text, skipTrivia(text, node.pos, /*stopAfterLineBreak*/ false, /*stopAtComments*/ true));
-        return some(commentRanges) && hasInternalAnnotation(last(commentRanges), sourceFile);
-    }
-    const leadingCommentRanges = parseTreeNode && getLeadingCommentRangesOfNode(parseTreeNode, sourceFile);
-    return !!forEach(leadingCommentRanges, range => {
-        return hasInternalAnnotation(range, sourceFile);
-    });
+    return parseTreeNode && !!getJSDocInternalTag(parseTreeNode);
```
jakebailey, line 2600:
```suggestion
    return !!parseTreeNode && !!getJSDocInternalTag(parseTreeNode);
```

This function is public and should continue to return just boolean.

jakebailey:
Merged main for you.

As far as I'm concerned, this PR is correct; I tested `dtsBundler` with it and it still correctly strips internal declarations via `isInternalDeclaration`.

But I know we're pretty iffy on whether or not `@internal` is even supported or not. Only recently did I make the function to check the tag public, and this PR formalizes it into our API.

I'm personally still for it, regardless, but I'll bring this one up in a design meeting just to double check.

sheetalkamat (approved):

src/compiler/parser.ts
```diff
@@ -9107,6 +9107,9 @@ namespace Parser {
                     case "satisfies":
                         tag = parseSatisfiesTag(start, tagName, margin, indentText);
                         break;
+                    case "internal":
+                        tag = parseSimpleTag(start, factory.createJSDocInternalTag, tagName, margin, indentText);
+                        break;
```
jakebailey, line 9112: I was initially concerned that this didn't match any of the other tags, but this is actually more like `deprecated` above which does take this form.

jakebailey (commented):
I have no idea how it isn't breaking the tests, but this PR will need to modify `scanner.ts` to update `jsDocSeeOrLink` to also treat `@internal` as a JSDoc comment that needs to be parsed in TS files and an update to the docs for `JSDocParsingMode` to say that `@internal` is also parsed. Along with testing in `jsDocParsing.ts`.

Probably need to make `ParseForTypeErrors` the default in the compiler tests... Really thought I did that already.

jakebailey (changes requested):
Oops, meant to mark as needing changes.

jakebailey:
Close/reopening to rerun CI and show the failure.

jakebailey:
Took the liberty to apply the described change.

sheetalkamat (approved):

jakebailey (approved):

jakebailey:
Just checking:

@typescript-bot test it

jakebailey (changes requested):
Those new DT errors are interesting; seems like a bug in the PR somewhere.

Zzzen:
The DT errors should have been fixed.

jakebailey:
@typescript-bot test it

jakebailey:
@typescript-bot test it

jakebailey:
Perf is unfortunately still greatly affected and needs to be looked into.

Zzzen:
Perf problem is likely caused by the regex change.

`const jsDocNeededInTypeScriptFile = /@(?:see|link|internal)/i;` is 6% slower than the original one.

[perf.link](https://perf.link/#eyJpZCI6ImhxMGlieWNkeGVkIiwidGl0bGUiOiJGaW5kaW5nIG51bWJlcnMgaW4gYW4gYXJyYXkgb2YgMTAwMCIsImJlZm9yZSI6ImZ1bmN0aW9uIG1ha2VpZChsZW5ndGgpIHtcbiAgICBsZXQgcmVzdWx0ID0gJyc7XG4gICAgY29uc3QgY2hhcmFjdGVycyA9ICdBQkNERUZHSElKS0xNTk9QUVJTVFVWV1hZWmFiY2RlZmdoaWprbG1ub3BxcnN0dXZ3eHl6MDEyMzQ1Njc4OSc7XG4gICAgY29uc3QgY2hhcmFjdGVyc0xlbmd0aCA9IGNoYXJhY3RlcnMubGVuZ3RoO1xuICAgIGxldCBjb3VudGVyID0gMDtcbiAgICB3aGlsZSAoY291bnRlciA8IGxlbmd0aCkge1xuICAgICAgcmVzdWx0ICs9IGNoYXJhY3RlcnMuY2hhckF0KE1hdGguZmxvb3IoTWF0aC5yYW5kb20oKSAqIGNoYXJhY3RlcnNMZW5ndGgpKTtcbiAgICAgIGNvdW50ZXIgKz0gMTtcbiAgICB9XG4gICAgcmV0dXJuIHJlc3VsdDtcbn1cblxuY29uc3Qgc3RyaW5ncyA9IEFycmF5LmZyb20oe2xlbmd0aDogMWU0fSkubWFwKCgpID0%2BIG1ha2VpZCg1MCkpIiwidGVzdHMiOlt7Im5hbWUiOiJqc0RvY1NlZU9yTGluayIsImNvZGUiOiJjb25zdCBqc0RvY1NlZU9yTGluayA9IC9AKD86c2VlfGxpbmspL2k7XG5zdHJpbmdzLmZvckVhY2goeCA9PiBqc0RvY1NlZU9yTGluay50ZXN0KHgpKSIsInJ1bnMiOlsxMDAwLDgyMywxMDAwLDk0MSw0NzAsODgyLDY0Nyw4MjMsODgyLDExMTcsMTAwMCw2NDcsMTA1OCw4ODIsMTExNyw3NjQsODgyLDExMTcsOTQxLDg4MiwxMDAwLDEwNTgsNzA1LDEwMDAsOTQxLDgyMyw1OCw0NzAsMTA1OCw1MjksNzY0LDQ3MCw0NzAsNjQ3LDU4OCwxMDAwLDUyOSwxMDU4LDc2NCwxMTc2LDc2NCw4MjMsMTA1OCw4ODIsNzA1LDgyMyw3NjQsNzY0LDgyMyw0MTEsNDcwLDU4LDEwNTgsMTExNyw5NDEsNzA1LDQxMSw5NDEsNTg4LDEwMDAsMTAwMCw4MjMsODgyLDk0MSw3MDUsNzA1LDExNzYsODIzLDQxMSwxMTcsOTQxLDk0MSw5NDEsMTAwMCw4ODIsNTg4LDExNzYsNzA1LDUyOSwxMTc2LDgyMywxMTE3LDc2NCw3MDUsNjQ3LDEwNTgsNzA1LDc2NCw3NjQsOTQxLDExMTcsMTA1OCw5NDEsNTI5LDcwNSw4MjMsNTg4LDEwNTgsMTAwMCwxMDU4XSwib3BzIjo4MTd9LHsibmFtZSI6ImpzRG9jTmVlZGVkSW5UeXBlU2NyaXB0RmlsZSIsImNvZGUiOiJjb25zdCBqc0RvY05lZWRlZEluVHlwZVNjcmlwdEZpbGUgPSAvQCg%2FOnNlZXxsaW5rfGludGVybmFsKS9pO1xuc3RyaW5ncy5mb3JFYWNoKHggPT4ganNEb2NOZWVkZWRJblR5cGVTY3JpcHRGaWxlLnRlc3QoeCkpIiwicnVucyI6Wzc2NCw1ODgsOTQxLDk0MSw3NjQsOTQxLDU4OCw3MDUsNzA1LDk0MSw4MjMsNzY0LDc2NCw4ODIsMTAwMCw5NDEsNTg4LDg4Miw3MDUsMTAwMCw5NDEsOTQxLDg4Miw3NjQsODgyLDcwNSw4ODIsMTAwMCwxMTcsOTQxLDEwMDAsNzY0LDQ3MCw2NDcsOTQxLDUyOSw2NDcsNjQ3LDEwMDAsODgyLDY0Nyw3NjQsODgyLDcwNSw3MDUsODIzLDcwNSw4ODIsODIzLDg4Miw1MjksMTAwMCw0NzAsMTAwMCw1MjksNDcwLDk0MSw3MDUsODIzLDg4Miw5NDEsNjQ3LDQxMSw4ODIsNzY0LDc2NCw5NDEsOTQxLDI5NCw0MTEsNTI5LDc2NCw5NDEsMTAwMCw4MjMsOTQxLDEwMDAsOTQxLDcwNSw3NjQsOTQxLDg4Miw3NjQsNjQ3LDU4OCw5NDEsODIzLDgyMywxMDU4LDI5NCw2NDcsNzY0LDg4Miw4ODIsNTI5LDcwNSw4MjMsNTg4LDUyOSw4ODJdLCJvcHMiOjc3M31dLCJ1cGRhdGVkIjoiMjAyNC0wNi0yOVQxNDoyMTowMy40NzJaIn0%3D)

jakebailey:
Funnily I clicked that link and reran it, and it told me that the new one was 6% faster; I think this particular result is just noisy. Given the benchmark is consistent with so few samples, a profile may provide more insight (which I plan to test).

Zzzen:
LOL. After I executed it many times, I also found that the output result is somewhat random.

sandersn:
Do the random perf results mean this is safe to merge? I'd guess not, unless it's a problem with the perf infrastructure.

jakebailey:
No, there's nothing wrong with the perf infra, there really is a slowdown here that needs to be profiled.

jakebailey:

I can't seem to get the slowdown locally...

PR #57963, opened 2024-03-27
Underline type nodes without position mappings in .types baselines

weswigham:
The underlines themselves, we've talked about in general already, but the underlining logic should get a proper look-over - I think it's pretty good, but I wrote it 😄


src/harness/typeWriter.ts
```diff
@@ -139,18 +271,26 @@ export class TypeWriterWalker {
...
+                if (ts.isIdentifier(node) && ts.isTypeAliasDeclaration(node.parent) && node.parent.name === node && ts.isIdentifier(typeNode) && ts.idText(typeNode) === ts.idText(node)) {
                     // for a complex type alias `type T = ...`, showing "T : T" isn't very helpful for type tests. When the type produced is the same as
                     // the name of the type alias, recreate the type string without reusing the alias name
-                    typeString = this.checker.typeToString(type, node.parent, typeFormatFlags | ts.TypeFormatFlags.InTypeAlias);
+                    typeNode = this.checker.typeToTypeNode(type, node.parent, ((typeFormatFlags | ts.TypeFormatFlags.InTypeAlias) & ts.TypeFormatFlags.NodeBuilderFlagsMask) | ts.NodeBuilderFlags.IgnoreErrors)!;
                 }
+
+                const { printer, writer, underliner, reset } = createSyntheticNodeUnderliningPrinter();
+                printer.writeNode(ts.EmitHint.Unspecified, typeNode, this.currentSourceFile, writer);
+                const result = writer.getText();
```
jakebailey, line 283: Why not jus `typeString = ` here?

weswigham: ...yeah, that'll remove a line.

src/harness/typeWriter.ts
```diff
@@ -139,18 +271,26 @@ export class TypeWriterWalker {
...
+                let typeNode = this.checker.typeToTypeNode(type, node.parent, (typeFormatFlags & ts.TypeFormatFlags.NodeBuilderFlagsMask) | ts.NodeBuilderFlags.IgnoreErrors)!;
+                if (ts.isIdentifier(node) && ts.isTypeAliasDeclaration(node.parent) && node.parent.name === node && ts.isIdentifier(typeNode) && ts.idText(typeNode) === ts.idText(node)) {
                     // for a complex type alias `type T = ...`, showing "T : T" isn't very helpful for type tests. When the type produced is the same as
                     // the name of the type alias, recreate the type string without reusing the alias name
-                    typeString = this.checker.typeToString(type, node.parent, typeFormatFlags | ts.TypeFormatFlags.InTypeAlias);
+                    typeNode = this.checker.typeToTypeNode(type, node.parent, ((typeFormatFlags | ts.TypeFormatFlags.InTypeAlias) & ts.TypeFormatFlags.NodeBuilderFlagsMask) | ts.NodeBuilderFlags.IgnoreErrors)!;
                 }
+
+                const { printer, writer, underliner, reset } = createSyntheticNodeUnderliningPrinter();
+                printer.writeNode(ts.EmitHint.Unspecified, typeNode, this.currentSourceFile, writer);
```
jakebailey, line 282: Rather than going in this direction, does `checker.writeType` work?

jakebailey: Maybe not, actually, given you want to avoid the thing above where we are calling string stuff twice.

weswigham: We need a custom `printer` to register an `onEmitNode` hook, and a custom `writer` to write both the underline and the actual text. `typeToString` already takes an optional `writer`, but not a `printer`. `writeType` is the same - it's just a full binding to the internal `typeToString` that doesn't hide the `writer` parameter.

jakebailey (approved):
Underling code seems fine to me; I can't think of a better way to do it.

Of course, the rest of the change is HugeMcLarge and we're going to have to fix all PRs, but I think we were okay with it just to get this coverage.

rbuckton:
Unfortunately, due to the number of baselines that changed there are too many files in the diff view and it's crashing the browser tab when I try to comment. I provided some feedback to @weswigham offline. 

Otherwise, LGTM.

src/harness/typeWriter.ts
```diff
@@ -48,7 +48,7 @@ const createSyntheticNodeUnderliningPrinter = memoize((): { printer: ts.Printer;
     let underlining = false;
     const printer = createPrinter({ removeComments: true }, {
         onEmitNode: (hint, node, cb) => {
-            if (ts.nodeIsSynthesized(node) === underlining) {
+            if (ts.nodeIsSynthesized(node) !== underlining) {
                 // either node is synthetic and underlining needs to be disabled, or node is not synthetic and
```
rbuckton, line 52: My bad, I had the conditions reversed. I would update the comment here to match.

rbuckton (changes requested):

rbuckton (approved):

PR #57966, opened 2024-03-27
Use evaluator for isolatedModules enum restrictions

andrewbranch:
With #57914, it becomes possible to enable `isolatedModules` in our own codebase. I tried this, but there were a few inconvenient changes required to enum declarations due to #56736. You can see them here: https://github.com/andrewbranch/TypeScript/compare/bug/37774...andrewbranch:TypeScript:enable-isolatedModules?expand=1

The checks added in #56736 seem to be a bit too restrictive. Specifically, references to same-file constants and other enum members were not recognized. I refactored the evaluator to return extra information that lets us produce the same errors with higher fidelity to the capabilities of the evaluator. It seems that transpilers (tested esbuild and Babel) have built their behavior around what our own evaluator can do; they’re just obviously limited to analyzing a single file (when not bundling). So, this PR removes `isolatedModules` errors from a lot of code that transpiles perfectly fine.

(cc @frigus02)


jakebailey:
The worst PR race; we just merged #57955 which moved this code.

andrewbranch:
@typescript-bot test it

jakebailey:
Uh oh, the last merge just broke the PR a lot

andrewbranch:
Yeah I was not on the branch I thought I was on 🤦🏻‍♂️

src/compiler/types.ts
```diff
@@ -5626,6 +5626,7 @@ export interface EmitResolver {
     isEntityNameVisible(entityName: EntityNameOrEntityNameExpression, enclosingDeclaration: Node): SymbolVisibilityResult;
     // Returns the constant value this property access resolves to, or 'undefined' for a non-constant
     getConstantValue(node: EnumMember | PropertyAccessExpression | ElementAccessExpression): string | number | undefined;
```
jakebailey, line 5628: Does this still need to have `EnumMember`?

jakebailey: Actually, `declarations.ts` still calls `getConstantValue` but only on an enum value, which makes me think that the resolver wouldn't need that func at all anymore. But, `ts.ts`'s `tryGetConstEnumValue` still calls it for property access / element accesses :(

jakebailey (approved):
This seems good to me, though I'm still hoping that we can reduce the EmitResolver API surface since there are so few calls to the old API left...

Maybe @weswigham @dragomirtitian have some feedback on that front just from needing to implement that resolver in a minimal fashion, but given this code all exists and works via the existing evaluator that was just pulled out, I doubt this is really a big deal.

PR #57973, opened 2024-03-27
Compare package.json paths with correct sensitivity in getLocalModuleSpecifier

### LINKED ISSUE #57802: "Inferred type cannot be named" error caused by adding `paths` to tsconfig
### 🔎 Search Terms

"inferred type cannot be named", "paths"

- This changed in PR #55015

### ⏯ Playground Link

https://github.com/haines/typescript-paths-inferred-type-cannot-be-named

### 💻 Code

Please see the linked repo (sorry, I couldn't reduce it any further).

### 🙁 Actual behavior

In a multiproject build where `app` depends on `lib`, setting `paths` in `lib`'s tsconfig breaks `app`'s build with

```
error TS2742: The inferred type of '...' cannot be named without a reference to '...'. This is likely not portable. A type annotation is necessary.
```

### 🙂 Expected behavior

Adding `paths` shouldn't break the dependent project.

Judging by the PR that I was able to bisect to, this seems like it could a bad interaction between `paths` in tsconfig.json and `exports` in package.json.

### LINKED ISSUE #57926: Compilation fails with TS6059 under Windows 10 but passes under Windows 11 or Linux
### 🔎 Search Terms

"TS6059", "Windows 10"

- This changed between versions 5.4.0-dev.20231221 and 5.4.0-dev.20231222
- This changed in commit or PR [Add auto-import for the package.json imports field](https://github.com/microsoft/TypeScript/pull/55015)

### ⏯ Playground Link

_No response_

### 💻 Code

_No response_

### 🙁 Actual behavior

When compiling under Ubuntu Linux (containerised or via WSL), or on Windows 11: compilation of all modules is successful.

When compiling under Windows 10: compilation fails with TS6059 errors.

### 🙂 Expected behavior

Compilation is successful on all operating systems.

We discovered this problem on a large existing Angular codebase a few days after updating to Angular 17.3 along with TypeScript 5.4.3. We didn't discover this immediately because our CI processes run in Linux, and the build passes quite happily under this OS.

To narrow down the issue I have tried retaining Angular 17.3 but downgrading to Typescript 5.3.3, and the build passes on Windows 10. I then tried various nightly builds and found that:
- The build is successful when using TypeScript 5.4.0-dev.20231221
- The build fails when using 5.4.0-dev.20231222

We subsequently tried building on a Windows 11 workstation (it passes), and via a WSL Ubuntu terminal (it also passes).

This leads me to believe that the changes under https://github.com/microsoft/TypeScript/pull/55015 might have had an unexpected impact, particularly with regards to case insensitive paths specifically on Windows 10. I don't believe that Angular 17.3 has caused the problem because the problem does not exist when compiling under different versions of TypeScript.

I have been unable to narrow down a minimal reproduction of the problem because, other than the different operating system, it isn't clear what the specific problem is in the code.

For what it's worth, the Angular project is broken down into multiple libraries, and each library has multiple secondary entry points. We see one of the library secondary entry point builds fail with TS6059, but only under Windows 10. Some upstream libraries and library secondary entry points build correctly.

### LINKED ISSUE #57956: Incorrect Path Resolution for Type Imports in TypeScript Declarations
### 🔎 Search Terms

"specifier", "relative path"

- This changed between versions 5.4.0-dev.20231221 and 5.4.0-dev.20231222
- This changed in commit or PR https://github.com/microsoft/TypeScript/pull/55015

### ⏯ Playground Link

https://github.com/xlianghang/ts-specifier-bug

### 💻 Code

```typescript
//src/main.ts
import { PartialType } from './utils';

class Common {}

export class Sub extends PartialType(Common) {
  id: string;
}

```

```typescript
// src/utils/index.ts
import { MyType, MyReturnType } from './type-helpers';

export function PartialType<T>(classRef: MyType<T>) {
  abstract class PartialClassType {
    constructor() {}
  }

  return PartialClassType as MyReturnType;
}
```

```typescript
// src/utils/type-helpers.ts
export type MyReturnType = {
  new (...args: any[]): any;
};

export interface MyType<T = any> extends Function {
  new (...args: any[]): T;
}

```

### 🙁 Actual behavior

image

The `dist` folder is not the root directory, so do not import from `src`

### 🙂 Expected behavior

image

Import from a relative path

Compilation fails in the MacOS, but is expected to behave in the https://stackblitz.com/~/github.com/xlianghang/ts-specifier-bug 

💻 MacOS 14.3，Apple M2 Max

jakebailey:
Fixes #57956
Fixes #57926
Fixes #57802
Closes #57961

Thanks @xlianghang for the test.

See the second commit for the change.

### DIFF
--- a/src/compiler/moduleSpecifiers.ts
+++ b/src/compiler/moduleSpecifiers.ts
@@ -548,11 +548,10 @@ function getLocalModuleSpecifier(moduleFileName: string, info: Info, compilerOpt
             return maybeNonRelative;
         }
 
-        let nearestTargetPackageJson = getNearestAncestorDirectoryWithPackageJson(host, getDirectoryPath(modulePath));
-        let nearestSourcePackageJson = getNearestAncestorDirectoryWithPackageJson(host, sourceDirectory);
-        nearestTargetPackageJson &&= toPath(nearestTargetPackageJson, projectDirectory, getCanonicalFileName);
-        nearestSourcePackageJson &&= toPath(nearestSourcePackageJson, projectDirectory, getCanonicalFileName);
-        if (nearestSourcePackageJson !== nearestTargetPackageJson) {
+        const nearestTargetPackageJson = getNearestAncestorDirectoryWithPackageJson(host, getDirectoryPath(modulePath));
+        const nearestSourcePackageJson = getNearestAncestorDirectoryWithPackageJson(host, sourceDirectory);
+        const ignoreCase = !hostUsesCaseSensitiveFileNames(host);
+        if (!packageJsonPathsAreEqual(nearestTargetPackageJson, nearestSourcePackageJson, ignoreCase)) {
             // 2. The importing and imported files are part of different packages.
             //
             //      packages/a/
@@ -570,6 +569,12 @@ function getLocalModuleSpecifier(moduleFileName: string, info: Info, compilerOpt
 
     // Prefer a relative import over a baseUrl import if it has fewer components.
     return isPathRelativeToParent(maybeNonRelative) || countPathComponents(relativePath) < countPathComponents(maybeNonRelative) ? relativePath : maybeNonRelative;
+}
+
+function packageJsonPathsAreEqual(a: string | undefined, b: string | undefined, ignoreCase?: boolean) {
+    if (a === b) return true;
+    if (a === undefined || b === undefined) return false;
+    return comparePaths(a, b, ignoreCase) === Comparison.EqualTo;
 }
 
 /** @internal */
--- a/src/compiler/moduleSpecifiers.ts
+++ b/src/compiler/moduleSpecifiers.ts
@@ -548,8 +548,10 @@ function getLocalModuleSpecifier(moduleFileName: string, info: Info, compilerOpt
             return maybeNonRelative;
         }
 
-        const nearestTargetPackageJson = getNearestAncestorDirectoryWithPackageJson(host, getDirectoryPath(modulePath));
-        const nearestSourcePackageJson = getNearestAncestorDirectoryWithPackageJson(host, sourceDirectory);
+        let nearestTargetPackageJson = getNearestAncestorDirectoryWithPackageJson(host, getDirectoryPath(modulePath));
+        let nearestSourcePackageJson = getNearestAncestorDirectoryWithPackageJson(host, sourceDirectory);
+        nearestTargetPackageJson &&= toPath(nearestTargetPackageJson, projectDirectory, getCanonicalFileName);
+        nearestSourcePackageJson &&= toPath(nearestSourcePackageJson, projectDirectory, getCanonicalFileName);
         if (nearestSourcePackageJson !== nearestTargetPackageJson) {
             // 2. The importing and imported files are part of different packages.
             //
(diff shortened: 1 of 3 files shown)


jakebailey:
@typescript-bot pack this

andrewbranch (approved):

src/compiler/moduleSpecifiers.ts
```diff
@@ -548,8 +548,10 @@ function getLocalModuleSpecifier(moduleFileName: string, info: Info, compilerOpt
             return maybeNonRelative;
         }
 
-        const nearestTargetPackageJson = getNearestAncestorDirectoryWithPackageJson(host, getDirectoryPath(modulePath));
-        const nearestSourcePackageJson = getNearestAncestorDirectoryWithPackageJson(host, sourceDirectory);
+        let nearestTargetPackageJson = getNearestAncestorDirectoryWithPackageJson(host, getDirectoryPath(modulePath));
+        let nearestSourcePackageJson = getNearestAncestorDirectoryWithPackageJson(host, sourceDirectory);
+        nearestTargetPackageJson &&= toPath(nearestTargetPackageJson, projectDirectory, getCanonicalFileName);
+        nearestSourcePackageJson &&= toPath(nearestSourcePackageJson, projectDirectory, getCanonicalFileName);
```
jakebailey, line 554: Not clear if `getNearestAncestorDirectoryWithPackageJson` should just always return Path.

jakebailey: Or if i should just `comparePaths`?

rbuckton: This is used elsewhere and a `toPath` might have other negative affects at those use sites.

jakebailey: If you mean my `Path` comment, it's only used when reading a file so shouldn't actually affect anything, but it sounds like the comparison one may just be better.

rbuckton: I'd say just use `comparePaths`. We use it throughout `tryGetModuleNameFromExportsOrImports` against a path that comes from `getNearestAncestorDirectoryWithPackageJson` as well, so it would be consistent.

jakebailey: Updated to use a helper; slightly different than the one you suggested but I was having trouble understanding what I was reading...

jakebailey:
@typescript-bot test it

RyanCavanaugh (approved):

jakebailey:

rbuckton (approved):

src/compiler/moduleSpecifiers.ts
```diff
@@ -570,6 +571,12 @@ function getLocalModuleSpecifier(moduleFileName: string, info: Info, compilerOpt
     return isPathRelativeToParent(maybeNonRelative) || countPathComponents(relativePath) < countPathComponents(maybeNonRelative) ? relativePath : maybeNonRelative;
 }
 
+function packageJsonPathsAreEqual(a: string | undefined, b: string | undefined, ignoreCase?: boolean) {
+    if (a === b) return true;
+    if (a === undefined || b === undefined) return false;
+    return comparePaths(a, b, ignoreCase) === Comparison.EqualTo;
```
sheetalkamat, line 577: dont you need to pass current directory for this to work correctly

jakebailey: These paths are already absolute, it seems, so nothing broke...

jakebailey: Checked the other use in this file and they don't pass in the working dir either (also for paths from this host method)

PR #57974, opened 2024-03-27
Refactor name resolution to separate component

dragomirtitian:


src/compiler/checker.ts
```diff
@@ -1504,7 +1503,30 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+        getSymbolOfDeclaration,
+        error,
+        getNodeLinks,
+        lookup: getSymbol,
+        onPropertyWithInvalidInitializer: checkAndReportErrorForInvalidInitializer,
+        onFailedToResolveSymbol,
+        onSuccessfullyResolvedSymbol,
+    });
 
+    const resolveNameForSymbolSuggestion = createNameResolver({
```
jakebailey, line 1520:
```suggestion
    var resolveNameForSymbolSuggestion = createNameResolver({
```

Quick nit

dragomirtitian: Ah yes. Reflex to use const. will change.

src/compiler/checker.ts
```diff
@@ -3006,637 +3024,171 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
-                                isolatedModulesLikeFlagName,
-                                `${unescapeLeadingUnderscores(getSymbolOfNode(location)!.escapedName)}.${unescapeLeadingUnderscores(name)}`,
+        nameNotFoundMessage: DiagnosticMessage,
+    ) {
+        const name = typeof nameArg === "string" ? nameArg : (nameArg as Identifier).escapedText;
+        addLazyDiagnostic(() => {
+            if (
+                !errorLocation ||
+                errorLocation.parent.kind !== SyntaxKind.JSDocLink &&
+                    !checkAndReportErrorForMissingPrefix(errorLocation, name, nameArg) && // TODO: GH#18217
```
jakebailey, line 3061:
```suggestion
                    !checkAndReportErrorForMissingPrefix(errorLocation, name, nameArg) &&
```

(If we remove a `!`, the TODO can go away.)

jakebailey:

jakebailey:

src/compiler/utilities.ts
```diff
@@ -10761,3 +10777,577 @@ export function createEvaluator({ evaluateElementAccessExpression, evaluateEntit
...
+    onSuccessfullyResolvedSymbol,
+}: {
+    compilerOptions: CompilerOptions;
+    getSymbolOfDeclaration: (node: Declaration) => Symbol;
+    error: (location: Node | undefined, message: DiagnosticMessage, ...args: DiagnosticArguments) => void;
+    globals: SymbolTable;
+    argumentsSymbol: Symbol;
+    requireSymbol: Symbol;
+    lookup: (symbols: SymbolTable, name: __String, meaning: SymbolFlags) => Symbol | undefined;
+    getNodeLinks: (node: Node) => { declarationRequiresScopeChange?: boolean; };
```
dragomirtitian, line 10818: This does expose NodeLinks in a way that we might not want. We could split this into `setDeclarationRequiresScopeChange` and `getDeclarationRequiresScopeChange` that can be passed in without actually caring how the flag is stored. 

Keeping the cache local would result in different instantiations not being able to share the cache.

src/compiler/checker.ts
jakebailey: Need like an "only var" rule for these blocks 😄

src/compiler/utilities.ts
weswigham: We could push the logic all the way into a handler you pass in, a `onPossibleVariableResult` which takes the `result` `Symbol`, `location`, and `lastLocation`, and that the checker would just pass in as `useOuterVariableScopeInParameter` - that could keep the links abstraction inside the checker.

weswigham: > We could split this into setDeclarationRequiresScopeChange and getDeclarationRequiresScopeChange that can be passed in without actually caring how the flag is stored.

...that's probably better? I suppose this logic is important for callers to do (just not the particulars of storing the result - or even if the result needs to be stored), so moving it into the checker would result in duplication at all use-sites.

DanielRosenwasser:

src/compiler/utilities.ts
```diff
@@ -10761,3 +10779,578 @@ export function createEvaluator({ evaluateElementAccessExpression, evaluateEntit
...
+            location = isJSDocTemplateTag(location) ? getEffectiveContainerForJSDocTemplateTag(location) || location.parent :
+                isJSDocParameterTag(location) || isJSDocReturnTag(location) ? getHostSignatureFromJSDoc(location) || location.parent :
+                location.parent;
+        }
+
+        // We just climbed up parents looking for the name, meaning that we started in a descendant node of `lastLocation`.
+        // If `result === lastSelfReferenceLocation.symbol`, that means that we are somewhere inside `lastSelfReferenceLocation` looking up a name, and resolving to `lastLocation` itself.
+        // That means that this is a self-reference of `lastLocation`, and shouldn't count this when considering whether `lastLocation` is used.
+        if (isUse && result && (!lastSelfReferenceLocation || result !== lastSelfReferenceLocation.symbol)) {
+            result.isReferenced! |= meaning;
```
DanielRosenwasser, line 11200: How safe is logic like this to exist outside of the checker/binder?

weswigham: Oh, this isn't safe *at all* in the first place. It's mutating a field directly on the `Symbol`, which will never be reset in incremental compilation in the language service, but whose value *should* change as the file does. I don't know how this crops up as a visible error in the language service, but it almost definitely does. But it also *has* for... forever? And we haven't run into it yet? So.... pass. Deal with it later?

weswigham: Of the 3 kinds of reference tracking we doing, this one is like number 1 most likely to be buggy, but it's only used for unused reference tracking, unlike `NodeLinks["isVisible"]` and `NodeLinks["referenced"]` both of which specifically track usages for imports in `.d.ts` and `.js` emit, respectively. So bugs in this one are probably rare to see reported, since usually they only result in missing suggestion diagnostics.

weswigham: ...But if you do happen to know of issues along the lines of "this isn't marked as unused after I delete the code using it in this way" it's because this field here isn't on `SymbolLinks`.

DanielRosenwasser: I'm going to guess that it doesn't crop up as a problem because within a file that changes, we always blow away the information; in a file that stays the same, the information is still accurate.

I'm going to guess there's a bug to do with symbol merging across files.

dragomirtitian: Should we maybe also move this to a `markUsage` callback? I don't think this currently introduces any new bugs, since this is not called from anywhere else, and the future use I have in mind for this in the minimalist resolver will not really care about the usages.

jakebailey:
I guess I don't really have any other comments here; on the whole, it sure seems like a lot of checker functions needing to be passed into... I don't really have a good vision of what the next change that requires this looks like, though, to understand how impactful this is. So, soft +1 from me, and I think others know better than me 😅

DanielRosenwasser (approved):

PR #57977, opened 2024-03-28
Monomorphic flow nodes

ahejlsberg:
Experiment to observe the performance effects of monomorphic flow nodes.


ahejlsberg:
@typescript-bot test it

ahejlsberg:

ahejlsberg:

ahejlsberg:
@typescript-bot test it

ahejlsberg:
@typescript-bot  perf test this faster

jakebailey:
> Note that this is an API breaking change due to the more uniform layout of flow nodes.

To be honest, I'm not sure why any of `FlowNode` or `FlowFlags` are public; they are not referenced by anything but themselves. Makes me wonder if we should just be removing these so we can make more changes at will.

But, the API break will probably break the playground's flow node viewer and ts-ast-viewer.com's, but we can probably update both. Realistically the former should have been calling into `ts.Debug` anyway, and the latter pins specific TS versions, I think.

src/compiler/types.ts
```diff
@@ -4101,19 +4102,28 @@ export type FlowNode =
 
 export interface FlowNodeBase {
     flags: FlowFlags;
-    id?: number; // Node id used by flow type cache in checker
+    id: number; // Node id used by flow type cache in checker
+    node: unknown; // Node or other data
+    antecedent: FlowNode | FlowNode[] | undefined;
```
jakebailey, line 4107: I'm kind of surprised that this turned out okay; I would have thought these two types would have different shapes, but I guess they're both pointers...

ahejlsberg: They're both heap objects so should get the same hidden class.

ahejlsberg:
@jakebailey Maybe we just fold the `@internal` change into this PR?

jakebailey:
I figured it may be better not to, just to make it clear that we intentionally removed them, but I'm not against it if we're good with removing them.

If #58036 is merged first, then your PR doesn't have any breaks (no baseline changes at all).

ahejlsberg:
> If https://github.com/microsoft/TypeScript/pull/58036 is merged first, then your PR doesn't have any breaks (no baseline changes at all).

Fine by me.

jakebailey:
Done ☺️

src/compiler/binder.ts
```diff
@@ -496,9 +501,8 @@ export const enum ContainerFlags {
     IsObjectLiteralOrClassExpressionMethodOrAccessor = 1 << 7,
 }
 
-function initFlowNode<T extends FlowNode>(node: T) {
-    Debug.attachFlowNodeDebugInfo(node);
-    return node;
+export function createFlowNode(flags: FlowFlags, node: unknown, antecedent: FlowNode | FlowNode[] | undefined): FlowNode {
```
jakebailey, line 504:
```suggestion
/** @internal */
export function createFlowNode(flags: FlowFlags, node: unknown, antecedent: FlowNode | FlowNode[] | undefined): FlowNode {
```

ahejlsberg: Done!

jakebailey (approved):

PR #57986, opened 2024-03-28
Isolate test failure from #57896

andrewbranch:

PR #57988, opened 2024-03-28
Ensure that we are checking if correct file with resolved path is present in the new program when removing the existing packageJson watching

sheetalkamat:
The file's resolved path might not be same if "referenced tsconfig" gets deleted. So we need to ensure that file we get from program has same resolvedPath as what we are looking for, if not remove the existing watched locations
Fixes https://github.com/microsoft/TypeScript/pull/57986


jakebailey (approved):
I was hoping the first commit would add a baseline for the second to have a diff on, but the first commit shows that there's a debug failure, darn

PR #57990, opened 2024-03-28
Node builder reuses type predicate nodes, more often reuses return position nodes

weswigham:
As you can see from the baselines now, this improves how often we reuse nodes by quite a lot.


tests/baselines/reference/trackedSymbolsNoCrash.js
```diff
@@ -743,204 +743,6 @@ export interface Node99 {
...
-}> | Extract<ast.Node95, {
-    kind: NodeType;
-}> | Extract<ast.Node96, {
-    kind: NodeType;
-}> | Extract<ast.Node97, {
-    kind: NodeType;
-}> | Extract<ast.Node98, {
-    kind: NodeType;
-}> | Extract<ast.Node99, {
+export declare const isNodeOfType: <NodeType extends ast.SyntaxKind>(nodeType: NodeType) => (node: ast.Node | null | undefined) => node is Extract<ast.Node, {
```
weswigham, line 746: This change is probably the biggest "these kinds of changes [to reuse more input nodes] are obviously improvements for everybody" demonstrator in our test suite.

src/compiler/checker.ts
```diff
@@ -7780,21 +7782,10 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             if (thisParameter) {
                 parameters.unshift(thisParameter);
             }
+            context.flags = flags;
```
jakebailey, line 7785: I think I read this code at some point and the fact that this was never restored was always so odd

weswigham: I think the logic went was always "Suppress implicit any only matters for the lop-level signature, so we don't need to bother" - and that's true.... if the quickfix that sets it (it's only set by a quickfix), is never retooled to use `symbolTableToDeclarationStatements` to do multiple statements at once. If it ever was, then there'd be multiple top-level statements in the context, and that assumption'd be wrong.

So I just made sure we were resetting the flag after changing it, so if we ever do that, we won't get bit.

tests/baselines/reference/trackedSymbolsNoCrash.js
jakebailey: Very nice, though I wonder if the crash from this test case was related to the printing or not...

tests/baselines/reference/baseClassImprovedMismatchErrors.errors.txt
```diff
@@ -17,7 +17,7 @@ baseClassImprovedMismatchErrors.ts(14,5): error TS2416: Property 'n' in type 'De
           Type 'string | number' is not assignable to type 'number'.
             Type 'string' is not assignable to type 'number'.
 baseClassImprovedMismatchErrors.ts(15,5): error TS2416: Property 'fn' in type 'DerivedInterface' is not assignable to the same property in base type 'Base'.
-  Type '() => string | number' is not assignable to type '() => number'.
+  Type '() => number | string' is not assignable to type '() => number'.
     Type 'string | number' is not assignable to type 'number'.
```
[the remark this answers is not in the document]
weswigham: Yeah, the signature has a declaration, so can reuse its' return type node, but the union type itself has no such luck. Now, it doesn't *have* to be that way (intersections remember their order forever because it matters if there are overloads, unions could do the same even if it only matters for looks) - but we'd produce more type identities.

src/compiler/checker.ts
```diff
@@ -8743,15 +8756,24 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                         node.isTypeOf,
                     );
                 }
-                if (isParameter(node)) {
-                    if (!node.type && !node.initializer) {
-                        return factory.updateParameterDeclaration(node, /*modifiers*/ undefined, node.dotDotDotToken, visitEachChild(node.name, visitExistingNodeTreeSymbols, /*context*/ undefined), node.questionToken, factory.createKeywordTypeNode(SyntaxKind.AnyKeyword), /*initializer*/ undefined);
-                    }
+                if (isNamedDeclaration(node) && node.name.kind === SyntaxKind.ComputedPropertyName && !isLateBindableName(node.name)) {
+                    return undefined;
```
jakebailey, line 8760: This is the only new return of undefined; is there a way to avoid it to prevent all of the extra `!`?

weswigham: ...for now. There will be more.

src/compiler/checker.ts
```diff
@@ -8743,15 +8756,24 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+                    || (isPropertySignature(node) && !node.type && !node.initializer)
+                    || (isParameter(node) && !node.type && !node.initializer)
+                ) {
+                    let visited = visitEachChild(node, visitExistingNodeTreeSymbols, /*context*/ undefined);
+                    if (visited === node) {
+                        visited = setTextRange(factory.cloneNode(node), node);
                     }
+                    (visited as Mutable<typeof visited>).type = factory.createKeywordTypeNode(SyntaxKind.AnyKeyword);
+                    if (isParameter(node)) {
+                        (visited as Mutable<ParameterDeclaration>).modifiers = undefined;
```
jakebailey, line 8774: Are these guaranteed to be copies? This is the first time we're pulling `Mutable` into the checker and it feels odd to have to do this.

src/compiler/checker.ts
weswigham: (Right now the node builder's existing node tree reuse is somewhat poor at eliding erroneous nodes from the input - this is something that the top-level declaration emitter currently does more consistently than this. Honestly, ideally this function and `visitDeclarationSubtree` in the declaration emitter would be largely shared (since they do the same thing: walk an existing node tree and make it fit to emit), but unifying them is going to take a bit - one needs to subsume the other's feature set entirely first, rather than them having overlap, but also unique parts)

src/compiler/checker.ts
weswigham: Yes - right above we say "if it's not a copy, copy it so we can add a `.type` safely".

Honestly, this is just the bare minimum needed to keep the tests looking correct while reusing the current set of input nodes. As I mentioned in the last comment, this scenario, of transforming the node to no longer have some kinds of error states in them (erroneous modifiers, missing required type nodes), this function is actually pretty lacking in compared to the declaration emitter's `visitDeclarationSubtree` which is its' dual. It needs a proper cleanup in that regard, but I don't think we're quite ready for that just yet - testing it won't be easy until we're actually reusing way more nodes via the node builder (which is the plan!).

jakebailey: Ah, I guess I didn't read that code that way; wasn't sure if one could return an entirely different but not synthesized node. But, makes sense.

jakebailey (approved):

src/compiler/checker.ts
weswigham: For additional context: `visitDeclarationSubtree` in `declarations.ts` and `tryReuseExistingTypeNodeHelper` in `checker.ts` have the same goal: take in a type node (or part thereof), and make it fit for emit, per our declaration file rules. Except `tryReuseExistingTypeNodeHelper` has only really seen extensive use for JS emit - so today it mostly just copies JSDoc type nodes and transforms them, while also doing some reference and visibility tracking.  Meanwhile `visitDeclarationSubtree` doesn't handle JSDoc *at all*, but beacuse it's been in use longer for declaration emit, handles more edge cases around correcting erroneous input patterns when building the output (and also does some of that same reference tracking).

Eventually we'll merge the two into one implementation used in both places (ideally), but until then, I'm just fixing the things `tryReuseExistingTypeNodeHelper` is obviously demonstrated as lacking when it starts returning the input node in our tests.

PR #57992, opened 2024-03-28
[experiment] Use Sets instead of lists in more places

jakebailey:
An old experiment inspired by #55224

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -29745,8 +29745,8 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                             const links = getNodeLinks(part);
                             links.flags |= NodeCheckFlags.ContainsCapturedBlockScopeBinding;
 
-                            const capturedBindings = links.capturedBlockScopeBindings || (links.capturedBlockScopeBindings = []);
-                            pushIfUnique(capturedBindings, symbol);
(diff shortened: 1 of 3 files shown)


jakebailey:

src/compiler/checker.ts
```diff
@@ -25983,20 +25987,20 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+                        (matchedSources ??= new Set()).add(s);
+                        (matchedTargets ??= new Set()).add(t);
                     }
                 }
             }
             return [
-                matchedSources ? filter(sources, t => !contains(matchedSources, t)) : sources,
-                matchedTargets ? filter(targets, t => !contains(matchedTargets, t)) : targets,
+                matchedSources ? filter(sources, t => !matchedSources!.has(t)) : sources,
+                matchedTargets ? filter(targets, t => !matchedTargets!.has(t)) : targets,
```
DanielRosenwasser, line 26003: This is so weird - why didn't these always just return `matchedSources ?? sources` and `matchedTargets ?? targets` before?

DanielRosenwasser: Oh, these are the ones that *don't match*.

jakebailey:
Dissapointing

PR #57993, opened 2024-03-28
Remove getAllAccessorDeclarations from the EmitResolver

weswigham:
This reduces the coupling between the checker and the transforms a bit. We actually had a `getAllAccessorDeclarations` in `utilities` already - the only difference being it required the container node, too (which was readily available at all callsites). Within the checker, the reverse is done - it no longer uses the `getAllAccessorDeclarations` `utilities` function, and now reliably uses the `symbol`-based lookup method the `EmitResolver` previously exported, and more reliably looks up the type nodes from accessors. (This does not change any baselines yet, because accessors currently disqualify node reuse - another followup will change that.)

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -235,7 +235,6 @@ import {
     GenericType,
     GetAccessorDeclaration,
     getAliasDeclarationFromName,
-    getAllAccessorDeclarations,
     getAllJSDocTags,
     getAllowSyntheticDefaultImports,
     getAncestor,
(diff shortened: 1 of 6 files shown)


weswigham:

src/compiler/checker.ts
```diff
@@ -48894,6 +48909,37 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+        const direct = getEffectiveTypeAnnotationNode(node);
+        if (direct) {
+            return direct;
+        }
+        if (node.kind === SyntaxKind.Parameter && node.parent.kind === SyntaxKind.SetAccessor) {
+            const other = getAllAccessorDeclarationsForDeclaration(node.parent as SetAccessorDeclaration).getAccessor;
+            if (other) {
+                return getEffectiveReturnTypeNode(other);
+            }
+        }
```
jakebailey, line 48922: Just to be clear, this code is a noop, right? Since we're not actually handling these and no baselines change? This is brand new code, so, checking.

jakebailey (approved):

src/compiler/checker.ts
weswigham: Yes, for now. My goal is to move the full "should this node be reused" logic check into the node builder, to minimize the `EmitResolver`'s API surface. Today, the node builder never has to emit nonlocal types for `get` and `set` accessors. If one is missing an annotation, it gets the type the other one has, and if both have the same type, it prints them as a property, rather than accessors. Meanwhile, declaration emit proper obviously leaves accessor declarations as-is nowadays, and uses logic like this to pull the type node from the other accessor when one is missing an annotation. So this is *slightly* future-looking, in that it enables the node builder to properly reuse nodes for `get`/`set` pairs that are missing one type node, even though the declaration emitter doesn't *yet* ask it to do that (and you can't get an inferred type with a `get`/`set` pair without them having differing types, and thus both annotated).

PR #57994, opened 2024-03-28
Remove some unneeded parameters from NodeBuilder APIs

weswigham:
The `includePrivateSymbol` being passed around became extraneous when the emitter started overriding the `SymbolTracker` in the context with once that calls its' active `includePrivateSymbol` hook. And `bundled` is just trivially calculable from the initial context information, and literally never changed as it was passed around.


tests/baselines/reference/jsDeclarationsFunctionClassesCjsExportAssignment.js
```diff
@@ -240,6 +233,13 @@ type Hook = import("./hook");
  * Imports
  */
 type HookHandler = import("./hook").HookHandler;
+/**
+ * Input type definition
+ */
+type Input = {
+    timer: Timer;
+    hook: Hook;
+};
```
weswigham, line 242: The order private symbols get added to the output by the symbol tracker alone is potentially slightly different than the one you got when we were using both the symbol tracker and the `includePrivateSymbol` callback being passed around directly. It's just an artifact of the order the functions were getting called in. All the required ones still get added.

jakebailey (approved):

PR #57995, opened 2024-03-29
Actually have a chance to reuse optional property signatures in the node builder

weswigham:
As you can see, I'd forgotten `PropertyDeclaration` and `PropertySignature` were not caught by the same check when I originally wrote this comparison, so optional property signatures (that didn't have an explicit `undefined` written out) never got reused in the node builder. And `PropertySignature`s are the ones actually within type nodes!

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -8557,7 +8557,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             if (typeFromTypeNode === type) {
                 return true;
             }
-            if (annotatedDeclaration && (isParameter(annotatedDeclaration) || isPropertyDeclaration(annotatedDeclaration)) && annotatedDeclaration.questionToken) {
+            if (annotatedDeclaration && (isParameter(annotatedDeclaration) || isPropertySignature(annotatedDeclaration) || isPropertyDeclaration(annotatedDeclaration)) && annotatedDeclaration.questionToken) {
                 return getTypeWithFacts(type, TypeFacts.NEUndefined) === typeFromTypeNode;
             }
             return false;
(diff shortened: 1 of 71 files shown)


jakebailey:
I feel nervous about this one in the same way I felt nervous about other reuses when `exactOptionalPropertyTypes` gets involved. Makes me wonder if all of these sorts of things have accidentally made the flag easier to use...

weswigham:
> Makes me wonder if all of these sorts of things have accidentally made the flag easier to use...

Hope not, since if you write a declaration at the top level, it's already appropriately kept in the fashion in which it was written. Here's a demonstration:

```ts repro
// @declaration: true
// @strict: true
export const a: { key?: { x?: number } } = {};
                 //^?


export const b = () => ([null as any as { key?: { x?: number } }][0]);
           //^?
```

[Workbench Repro](https://www.typescriptlang.org/dev/bug-workbench/?exactOptionalPropertyTypes=true&ts=5.5.0-dev.20240328#code/PTAEAEBMFMGMBsCGAnRAXAlgewHYC5Q1kBXaAWACgQIBnIjWNAo0y6ADwActk1RZcdUIgIBvUAGtoATwD8Y0O3mgcxALYAjaMlABfPaAC8oUboDclUFes3bVkAD1ZlFxQ7de-QXw1HQACgBKIwA+AIBtVXh4YRphHGlYk0kZZXElAlVNbQNdAF1wgAY8wIsKO1tHWSA)

Despite being declared in the _exact same way_, `a` lacks `| undefined` on its' optional properties in the declaration emit (because it was defined without them and we used those nodes), while `b` has `| undefined` in all its' optional types (and quickinfo for both has `| undefined`). The irony here is that when you set `exactOptionalPropertyTypes`, this is corrected, and the `| undefined` is removed from both because of the missing-type-removal calls inserted to handle it. (And yes, before you ask, the inline hint for the `//^?` has different text than the assertion result in the sidebar - I could not tell you why; maybe inline hints in the workbench get calculated without the compiler options set?)

We already respect user intent and keep their nodes as-is sometimes, all this does is make us do it more often - that probably makes this a (at least partial) bugfix, and not just an experience improvement 🤷‍♂️

jakebailey:
Yeah, I guess for isolated declarations we weren't going to throw `| undefined` on everything randomly...

jakebailey (approved):

PR #57996, opened 2024-03-29
Allow references to const enums when provably emitted via preserveConstEnums

### LINKED ISSUE #51530: Allow use of const enums in value space when preserveConstEnums=true

jakebailey:
This builds on #57914, allowing references to const enums in value space when we can "prove" that they will actually be emitted due to preserveConstEnums.

Fixes #51530


jakebailey:
Also, maybe this fixes #37774?

jakebailey:
@typescript-bot test it
@typescript-bot pack this

src/compiler/checker.ts
```diff
@@ -39815,18 +39816,33 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
-            const redirect = host.getRedirectReferenceForResolutionFromSourceOfProject(getSourceFileOfNode(constEnumDeclaration).resolvedPath);
-            if (constEnumDeclaration.flags & NodeFlags.Ambient && !isValidTypeOnlyAliasUseSite(node) && (!redirect || !shouldPreserveConstEnums(redirect.commandLine.options))) {
-                error(node, Diagnostics.Cannot_access_ambient_const_enums_when_0_is_enabled, isolatedModulesLikeFlagName);
+    function isPreservedConstEnumUse(use: Node, enumType: Type) {
+        Debug.assert(!!(enumType.symbol.flags & SymbolFlags.ConstEnum));
+        const constEnumDeclaration = enumType.symbol.valueDeclaration as EnumDeclaration;
+        const otherFile = getSourceFileOfNode(constEnumDeclaration);
+        if (!otherFile.isDeclarationFile) {
+            // This file can only have come from the current project.
+            if (constEnumDeclaration.flags & NodeFlags.Ambient && !isValidTypeOnlyAliasUseSite(use)) {
```
jakebailey, line 39839: After thinking harder about #57914, why is it okay for us to allow code to access ambient enums in other projects? Maybe I just totally misunderstood #57914... @andrewbranch

andrewbranch: Ok, first things first. An ambient declaration asserts the existence of a real one. The baseline expectation is that you can always access an ambient declaration in exactly the same way you can access a non-ambient one:

```ts
const x = 0;
declare const y: number;
console.log(x, y);
```

We would never error on the access of `y` on grounds that it’s ambient. All ambient declarations exist either because the user was declaring something that exists outside of the compilation, or as the direct result of a compilation (`.d.ts` output files).

So why would there be a rule that treats ambient and non-ambient const enums differently under `isolatedModules`? First, we have to establish why it’s _ever_ ok to access imported const enums under `isolatedModules`. The _only_ way that works is that `isolatedModules` implies `preserveConstEnums`, so an access that cannot be inlined by a single-file transpiler will still function as a normal property access. Under this logic, the ambient-ness of a const enum is used as a (pretty good) proxy for it coming from a different compilation (e.g. an external dependency), which means we can’t rely on `preserveConstEnums` having been set, which means a normal property access may fail as the declaration may not exist in JavaScript.

#57914 just tries to determine whether that ambient const enum was compiled by a project that sets `preserveConstEnums`, so we can know that access is safe. Maybe what you’re picking up on is the fact that ambient-ness is not a perfect proxy for “compiled from a non-ambient declaration in a `.ts` file,” since we could be seeing

```ts
// input.ts
export declare const enum Foo {
  A
}
```

But the way that’s unsafe in a cross-project scenario is also unsafe in the same project under `isolatedModules`, which is also similar to the way you can easily crash yourself by doing `console.log(x, y)` when you’ve declared `y` to exist but it doesn’t.

It’s worth noting that while this error about ambient const enums (over)protects against a pretty big footgun, it’s super inconsistent with nearly every other error we issue, where we pretty much naively assume that declaration files represent compilations that happened under the same options that we’re currently compiling with. 🤷

jakebailey: > Maybe what you’re picking up on is the fact that ambient-ness is not a perfect proxy for “compiled from a non-ambient declaration in a .ts file,” 

Yes, that's exactly it.

I guess my question is whether or not we should _always_ allow ambient const enum accesses, under the same conditions I'm using here for accessing non-ambient ones. Concretely, this comes down to `tests/cases/compiler/isolatedModulesAmbientConstEnum.ts` test stopping erroring:

```ts
// @isolatedModules: true
// @target: es6

// @filename: file1.ts

declare const enum E { X = 1}
export var y = E.X;
```

jakebailey: (Which involves just deleting the check I've highlighted in this comment thread.)

andrewbranch: Yes, I see no reason why that should error, based on the intended semantics of `declare`.

Issue #57999, opened 2024-03-29
No narrowing for "undefined" in switch default with exactOptionalPropertyTypes: true


Florent-Bouisset:
### 🔎 Search Terms

"switch default undefined", "switch exactOptionalPropertyTypes

### 🕗 Version & Regression Information

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about type guards.

### ⏯ Playground Link

https://www.typescriptlang.org/play?exactOptionalPropertyTypes=true#code/C4TwDgpgBA8gRgKwgY2AQQM4YJYHMB2UAvFKJAPYBmsiKwAdAIZZ74DcAUB9vsBAE6VGyaGigBvDlCjkwwbOXyMANgAV+sgPwAuKACIAFhGXLyezgF8uHSgFd8qBYTsOAFI365daAJQSp0lDIihjAEjJyTirqslAWxFAeuJwBGADu2MDIBq6y8orRGmB+4oFQAdLIzND2ACYQlDwQtdoVgfwQwLb8hHUNTbVcZVUY0IbGpnqtZdIdXT36RiZmKWX1QrbKwLpt0syj-MAAqvgdwgaMcMoQuZEFakU+uzNlAPSvUAB6wfihEflKB6yXR9Rr4ZrPF7Sd5fDAGcibWpQODQMgQXR6cEANwEegCVisLkcikSWAEx1OEHOl2urgA+rpsQIfIyIDj+P5pMADBo0lBwXyAKL8DT8Vx6E5nbI06BgRjc0iMADWEHweh8liAA

### 💻 Code

```ts

interface A {
  optionalProp?: "hello";
}

function func(arg: A) {
    const { optionalProp } = arg;

  switch(optionalProp) {    
    case undefined:
      return undefined

    case "hello":
      return "hello";

    default: 
      assertUnreachable(optionalProp)
              // ^const optionalProp: undefined
              // ^should be type: "never"
  }
}
function assertUnreachable(_: never): never {
  throw new Error("Unreachable path taken");
}
```

### 🙁 Actual behavior

`optionalProp` is of type `undefined` in the default case

### 🙂 Expected behavior

`optionalProp` should be of type `never` in the default case

### Additional information about the issue

Need to set the configuration option  `exactOptionalPropertyTypes` to `true`.
It's working as intended when set to `false`

PR #58001, opened 2024-03-29
Fixed an issue with `missingType` narrowing by `case undefined` in default cases

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/57999

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -28753,7 +28753,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             if (!hasDefaultClause) {
                 return caseType;
             }
-            const defaultType = filterType(type, t => !(isUnitLikeType(t) && contains(switchTypes, getRegularTypeOfLiteralType(extractUnitType(t)))));
+            const defaultType = filterType(type, t => !(isUnitLikeType(t) && contains(switchTypes, t.flags & TypeFlags.Undefined ? undefinedType : getRegularTypeOfLiteralType(extractUnitType(t)))));
             return caseType.flags & TypeFlags.Never ? defaultType : getUnionType([caseType, defaultType]);
         }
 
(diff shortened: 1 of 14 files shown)


RyanCavanaugh:
@typescript-bot test it

RyanCavanaugh:
I'm curious why we don't just put the regular `undefined` type into `optionalProp`; seems bad that it's so easy for this kind of mistake to happen. Wondering if there are others.

We should source out all the "compiler-inserted `undefined`s" in tests, e.g.
```ts
function func2(arg: A) {
  // Under noUncheckedIndexedAccess
  const optionalProp = ["hello" as const][Math.random()];
  switch (optionalProp) {
    case undefined:
    case "hello":
      return;
    default:
    // Should OK
      assertUnreachable(optionalProp)
  }
}
```
plus vary this test by all quadrants of NUIA / EOPT

RyanCavanaugh:
@typescript-bot pack this

Andarist:
> I'm curious why we don't just put the regular undefined type into optionalProp; seems bad that it's so easy for this kind of mistake to happen.

Yeah, same. I actually have thought that this missing type should be converted to undefined way earlier but I could not find evidence of that anywhere right now. 

> Wondering if there are others.

I have 2 other open PRs related to EOPT ;p so yes, this is very easy to miss and I wouldn't be surprised if more bugs like this are still out there.

src/compiler/checker.ts
```diff
@@ -28753,7 +28753,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             if (!hasDefaultClause) {
                 return caseType;
             }
-            const defaultType = filterType(type, t => !(isUnitLikeType(t) && contains(switchTypes, getRegularTypeOfLiteralType(extractUnitType(t)))));
+            const defaultType = filterType(type, t => !(isUnitLikeType(t) && contains(switchTypes, t.flags & TypeFlags.Undefined ? undefinedType : getRegularTypeOfLiteralType(extractUnitType(t)))));
```
Andarist, line 28756: a potential alternative to this could look like this:
```suggestion
            const switchTypesUnion = getUnionType(filter(switchTypes, isUnitType));
            const defaultType = filterType(type, t => !(isUnitLikeType(t) && areTypesComparable(t, switchTypesUnion)));
```

this is inspired by the logic used by `narrowTypeByEquality` and I've checked that this alternative would still pass the existing test cases.

RyanCavanaugh:
Behavior / implementation seem fine but I'd love to know if there's a more comprehensive fix

moataz2002 (commented):
Looks completed very well!

ahejlsberg (approved):

PR #58007, opened 2024-03-30
Deprecate module keyword for namespace declarations

### LINKED ISSUE #57913: Mark ~module~ `namespace` as deprecated via LS strikethrough
### 🔍 Search Terms

module strikethrough namespace internal deprecate

### ✅ Viability Checklist

### ⭐ Suggestion

See https://github.com/microsoft/TypeScript/issues/51825#issuecomment-2015163061

### 📃 Motivating Example

```ts
module notcool { }
```

### 💻 Use Cases

1. What do you want to use this for? Soft-deprecate internal `module`
2. What shortcomings exist with current approaches? Too breaky
3. What workarounds are you using in the meantime? Being sad

Zzzen:
Fixes #57913


jakebailey:
@typescript-bot test it

I'm going to make a second PR that uses this to issue a full error so we can gauge how bad things are at the moment (and verify that this is working).

jakebailey (approved):
This looks good to me, but deferring to @DanielRosenwasser / @RyanCavanaugh for final wording.

src/compiler/checker.ts
```diff
@@ -46365,6 +46366,14 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
 
             if (isIdentifier(node.name)) {
                 checkCollisionsForDeclarationName(node, node.name);
+                if (!(node.flags & (NodeFlags.Namespace | NodeFlags.GlobalAugmentation))) {
```
sheetalkamat, line 46369: This will only report errors in editor and there is no way to report these errors on command line. I think we need to handle "deprecation" version here instead to issue error

sheetalkamat (changes requested):

src/compiler/checker.ts
jakebailey: The intent is to not report the in tsc yet, per #57913

Zzzen:
Tests added.

DanielRosenwasser:
@typescript-bot pack this

DanielRosenwasser (approved):

Issue #43961, opened 2021-05-05
Type parameter can leak out of a function instead of being bound to specific type


jsnajdr:
# Bug Report

### 🔎 Search Terms

type parameter leaks out

### 🕗 Version & Regression Information

In the TS playground, the first version that shows the bug is 3.5.1. Version 3.3.3 also behaves incorrectly, but in a different way.

### ⏯ Playground Link

[Playground link with relevant code](https://www.typescriptlang.org/play?jsx=0&ts=4.3.0-beta#code/GYVwdgxgLglg9mABAdxlAFgBQIwB5GaIB8AFIgA4BcBiAlIgN4CwAUAJAQIDOUiAtogC8ifAEliZYACc4fauPqCiiMg0QA6TdNkAaDZvKIAvnQDcrNlICmUEFKR9zLI61acwPRAEMAJj5xCKGhY2KqIwHBw1NjGZm7cvL4+Vj4xwkk4YQBGXlLUAEyxtE7xHnAANlbq5XAA5mRJKdjqEXB6janqOVJxLKWeENZeUFYAKlJeHhFSAsJiegDyEohQeSrhMnKICkLKS4rKqyUsoJCwCEEYmPn4hKQU1IT0zOzunrMi28vaWztK62pNC1NnogYYTMULNZbPZEIMrMMxhMpnAZmQBJDnK4WG9En5roFUFd8mFWtEik5cd4-ClCul8STGIhugUKdi3hUqjV6tTkj58i1Iu0afyurkzEA)

### 💻 Code

```ts
function withP1<P>(p: P) {
  const m = <I>(from: I) => ({ ...from, ...p });
  return m;
}

const addP1 = withP1({ foo: 1 });
const added1 = addP1({ bar: 2 });

console.log(added1.foo, added1.bar);

const createTransform = <I, O>(tr: (from: I) => O) => tr;

function withP2<P>(p: P) {
  const m = <I>(from: I) => ({ ...from, ...p });
  return createTransform(m);
}

const addP2 = withP2({ foo: 1 });
const added2 = addP2({ bar: 2 });

console.log(added2.foo, added2.bar);
```

### 🙁 Actual behavior

The `added2.foo` expression reports a type error:
```
Property 'foo' does not exist on type '{ bar: number; } & P'.
```

The `index.d.ts` generated from this source has a declaration for `addP2` that contains an undefined identifier `P`:
```ts
declare const addP2: <I>(from: I) => I & P;
```

[...]

PR #58008, opened 2024-03-30
Fixed single signature type parameter leak

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/43961

cc @weswigham


src/compiler/checker.ts
```diff
@@ -19867,6 +19867,25 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     }
                 }
             }
+            if (type.objectFlags & ObjectFlags.SingleSignatureType) {
```
Andarist, line 19870: To be honest - this is very much "I don't know what I'm doing here". I "brute forced" this solution based on the changes from https://github.com/microsoft/TypeScript/pull/57403 and the investigation I was doing regarding this issue from before this PR was created. I kinda knew **where** a fix potentially could be applied and that's it.

I can move this code to a helper function etc. I just didn't bother because I'm not sure yet how good this fix is.

I'm not sure if this should be somehow cached (instantiations are cached in this function). And the type parameter that is now mapped (`P`) is still not in `clone.outerTypeParameters` - I have no idea if that's a problem or not.

weswigham: > I'm not sure if this should be somehow cached

100%. Should definitely be cached. And you should be able to get that by adding the missing bits of this into `instantiateAnonymousType`, rather than here. It already does the `clone.outerTypeParameters = singleSignatureType.outerTypeParameters;` bit - what it's missing is the "apply mapper to signature inside the type" bit. Ish. That's *supposed* to be implicit in the `result.mapper = mapper;` line already there... you'd assume, but alas - it looks like that only actually does something for types with a `.target` member. That may be the quickest fix - just set the `.target` of the type.

Andarist: I pushed out changes based on the stuff that you have mentioned here. No extra work had to be done in `instantiateAnonymousType` though. That already works (it sets the mapper and the target)... it just was never used in this case. 

The problem was that the computed `id` for this instantiation was the same as the id computed for the `target` in the branch that initializes `target.instantiations`. In other words, `target` was returned as that was grabbed from the `target.instantiations` and assigned as the existing `result`.

weswigham (approved):

PR #58010, opened 2024-03-31
Add types for `String#isWellFormed` and `String#toWellFormed`

### LINKED ISSUE #55543: Support `isWellFormed` and `toWellFormed` in lib.esnext.d.ts
### 🔍 Search Terms
NodeJS 20, isWellFormed, toWellFormed

### ✅ Viability Checklist

- [x] This wouldn't be a breaking change in existing TypeScript/JavaScript code

### ⭐ Suggestion

NodeJS 20 introduces new features that are not supported by the TypeScript compiler yet.

As an example node version 20 introduces `String.prototype.isWellFormed` and `String.prototype.toWellFormed`. Using these functions causes `error TS2339: Property 'isWellFormed' does not exist on type 'string'.`

Current environment:
- Node version: 20.5.1
- TypeScript version: 5.2.2

### 📃 Motivating Example

This update will enable developers to fully use all features available in the latest major version of NodeJS.

### 💻 Use Cases

This would allow the usage of ECMAScript features such as `isWellFormed` and `toWellFormed`.

The only current workaround is to use `// @ts-ignore` to let the project compile.

andreubotella:
These two methods are part of the [well-formed Unicode strings](https://github.com/tc39/proposal-is-usv-string) proposal, which reached stage 4 in May 2023 and will be part of ES2024.

Fixes #55543.


andreubotella:

src/lib/esnext.string.d.ts
```diff
@@ -0,0 +1,11 @@
+interface String {
+    /**
+     * Returns true if the string does not contain any lone UTF-16 surrogates.
```
sandersn, line 3: I'd rather use the description in the proposal:

```suggestion
     * Returns true if all leading surrogates and trailing surrogates appear paired and in order.
```

src/lib/esnext.string.d.ts
```diff
@@ -0,0 +1,11 @@
+interface String {
+    /**
+     * Returns true if the string does not contain any lone UTF-16 surrogates.
+     */
+    isWellFormed(): boolean;
+
+    /**
+     * Returns a string where all lone UTF-16 surrogates of this string have been replaced by the Unicode replacement character (U+FFFD).
```
sandersn, line 8: Same here: 

```suggestion
     * Returns a string where all lone or out-of-order UTF-16 surrogates have been replaced by the Unicode replacement character (U+FFFD).
```

I'm not sure whether UTF-16 needs to be mentioned or not.

sandersn (changes requested):
Structure looks good. I'd like to stay closer to the method descriptions in the proposal, though.

andrewbranch (approved):

DanielRosenwasser:
Thanks for the PR! To help users out a bit: in `utilities.ts` there's a function called `getScriptTargetFeatures` that recommends using the appropriate `lib`/`target` setting when someone accesses values that don't seem to exist. Can you update that?

andreubotella:
> Thanks for the PR! To help users out a bit: in `utilities.ts` there's a function called `getScriptTargetFeatures` that recommends using the appropriate `lib`/`target` setting when someone accesses values that don't seem to exist. Can you update that?

Done

DanielRosenwasser:
@sandersn are we good to merge?

DanielRosenwasser (approved):

PR #58013, opened 2024-03-31
Omit effects-free conditional constructs from control flow graph

ahejlsberg:
With this PR we optimize control flow graph construction to omit conditional constructs that have no control flow effects. Consider this example:

```ts
function test(x: A | B | C) {
  let str = x instanceof A ? doWithA(x) :
    x instanceof B ? doWithB(x) :
    doWithC(x) :
  foo(x);
  return str;
}
```

This currently creates the following control flow graph for the reference to `x` in the `foo(x)` call:

graphviz

With this PR, the effects-free conditional operator expression is removed from the control flow graph for references that follow it in the graph:

graphviz (1)

Intuitively, this optimization is possible because the observed control flow types at the bottom **BranchLabel** node are the same as the observed control flow types at the **Start** node for any variable, parameter, or property reference.

The shallower control flow graphs reduce check time up to 2% in several projects.


ahejlsberg:
@typescript-bot test it

ahejlsberg:
@typescript-bot test it

ahejlsberg:

ahejlsberg:
@jakebailey Any idea why we're seeing the test failure here https://github.com/microsoft/TypeScript/pull/58013#issuecomment-2030786044? I can't reproduce it locally.

jakebailey:
I was out yesterday; will take a look at this shortly.

jakebailey:
Cloning webpack, doing `yarn`, then running `tsc -p ./tsconfig.types.json` on main and with this PR does produce a difference:

```console
$ git clone git@github.com:webpack/webpack.git
$ cd webpack
$ yarn
$ node ~/work/TypeScript/built/local-old/tsc.js -p tsconfig.types.json |& tee ~/out-main.txt
$ node ~/work/TypeScript/built/local/tsc.js -p tsconfig.types.json |& tee ~/out-pr.txt
$ wc -l ~/out-main.txt ~/out-pr.txt
  3188 /home/jabaile/out-main.txt
  3187 /home/jabaile/out-pr.txt
  6375 total
$ gdiff ~/out-main.txt ~/out-pr.txt
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
renamed: home/jabaile/out-main.txt to home/jabaile/out-pr.txt
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
@ home/jabaile/out-pr.txt:2007 @ lib/optimize/SplitChunksPlugin.js(1011,30): error TS7006: Parameter 'key' implic
lib/optimize/SplitChunksPlugin.js(1022,37): error TS7006: Parameter 'key' implicitly has an 'any' type.
lib/optimize/SplitChunksPlugin.js(1095,35): error TS18048: 'cacheGroup.minChunks' is possibly 'undefined'.
lib/optimize/SplitChunksPlugin.js(1099,9): error TS2722: Cannot invoke an object which is possibly 'undefined'.
lib/optimize/SplitChunksPlugin.js(1099,9): error TS18048: 'cacheGroup.getName' is possibly 'undefined'.
lib/optimize/SplitChunksPlugin.js(1254,21): error TS18048: 'cacheGroup.minChunks' is possibly 'undefined'.
lib/optimize/SplitChunksPlugin.js(1687,37): error TS7006: Parameter 'value' implicitly has an 'any' type.
lib/optimize/SplitChunksPlugin.js(1687,44): error TS7006: Parameter 'key' implicitly has an 'any' type.
```

jakebailey:
Also, I modified this PR to force `hasFlowEffects` to always be true, and the error count went back to the one on main, so it's definitely some behavior change in the PR (somehow).

ahejlsberg:
@jakebailey Yeah, sorry, I see what's going on now. What confused me is that we used to issue two errors, and now we only issue one, but the errors have the exact same location. The missing error is an effect of optimizing away part of the control flow graph, causing us to not call `getEffectsSignature`, the method that resolves signatures in control flow analysis. For a variety of reasons we can't use the regular (and more complex) signature resolution in CFA, and the CFA resolution reports nullable errors slightly differently. Anyway, the error that went away is just a (differently worded) duplicate, so there's no problem here.

ahejlsberg:
Now that we've sorted out the missing error, I think this is ready for a review.

jakebailey (approved):

weswigham (approved):

Issue #20110, opened 2017-11-17
Feature request: allow user to merge extended arrays in tsconfig files


sveinburne:
**Scenario**: As a user, I would like to optionally merge extended arrays in `tsconfig` files. To do so, I would add a nested empty array `[]` to the property I want to merge. Here is an example: 
<details>
<summary><code>tsconfig-base.json</code></summary>
<pre>
{
  "exclude": ["**/__specs__/*"]
}
</pre>
</details> 

<details>
<summary><code>tsconfig-custom.json</code></summary>
<pre>
{
  "extends":  "./tsconfig-base.json",
  "exclude": [[], "lib"] // resolved to ["**/__specs__/*"; "lib"]
}
</pre>
</details> 
<br/>

**Alternative**: using a config `{}` object

<details>
<summary><code>tsconfig-custom.json</code></summary>
<pre>
{
  "extends":  "./tsconfig-base.json",
  "exclude": [{ "extends": true }, "lib"] // resolved to ["**/__specs__/*"; "lib"]
}
</pre>
</details>

weswigham:
Personally, I'd sooner like to see a `tsconfig.js` affordance (similar to a webpack config) to generally leverage JS syntax for complex configuration merging rather than introduce special-snowflake "syntax" into a json file with complex execution semantics.

mhegazy:
> Personally, I'd sooner like to see a tsconfig.js or tsconfig.ts affordance (similar to a webpack config) to generally leverage JS syntax for complex configuration merging rather than introduce special-snowflake "syntax" into a json file with complex layering and execution semantics. There's already a thing that exists to describe these kinds of procedural transforms, and it's called "code".

Noooooooooooooooo.

mhegazy:
We did talk about the OP when we first put in the configuration inheritance in place, and had a proposal of `extends` and `overwrites` to be two different behaviors for cases like these.. but in the end we chose to avoid complexity and just go with `extends` meaning `overwrite`. i think in hindsight that was a good discussion, and most users did not have the need for additional complexity. i would say for this request as well, the additional complexity (both in supporting the feature, and for users tsconfig.json files) is not worth the value you get out of it.

weswigham:
"Do not over-complicate the implementation with abstractions where a little copying will suffice" - Unattributed programming koan

sveinburne:
[...]

PR #58025, opened 2024-04-01
Proposal: Types as Configuration

weswigham:
Recently, we went over some ideas for new and interesting ways to [merge config files](https://github.com/microsoft/TypeScript/issues/57486). Ever since, or maybe, [since forever](https://github.com/microsoft/TypeScript/issues/20110#issuecomment-345384878), I've been thinking about the problem. The progression, give or take a step, for most command-line tools is this:
* First, there were command line arguments
* Second, there were response files, for when the command line arguments got too long (`@filename` on the CLI)
* Third, there is a structured data file in a known location (`.npmrc`, `.vscode/settings.json`, `eslint.json`, `tsconfig.json`)
* and Finally, there is just more Code as Configuration (`webpack.js`, `eslint.js`, `Herebyfile.mjs`)

We have the first three and have been reluctant to make the leap to the fourth. There are various reasons stated in different places, but mostly it comes down to not wanting or being able to execute untrusted code (especially in the editor). We're not in the business of sandboxing a dynamic, interpreted language, like javascript. Philosophically, there is little disagreement that a config file like this:

```ts
import * as ts from "typescript";

import base from "./tsconfig.base.js";
import mixin from "./tsconfig.mixin.js";

export default ts.config({
    compilerOptions: {
        ...base.compilerOptions,
        jsx: "react-jsx",
        paths: {
            "dep": ["./node_modules/typescript"],
            ...mixin.compilerOptions.paths
        },
        traceResolution: true
    },
    include: [ "./app/**/*.ts" ]
});
```
is fairly compelling - solving most of the issues we have with the static files of today (inflexible merging and imports, inability to create calculated fields generally) without needing to reinvent the wheel and build a custom procedural transform definition language in a static context.

But I couldn't help but wonder - sure, we have our reasons why we didn't like Code as Configuration for TypeScript - but what if we went _beyond_ that - what if we evolved our configuration into something even greater - more expressive, yet structured than code itself. What if we used _types as configuration_. After all, why not use what we have available? For us, checking is basically as easy as JSON parsing. And our types are, somehow, a sandboxed, dynamic, interpreted language unto themselves. [People do a lot of crazy things in our typesystem](https://github.com/microsoft/TypeScript/issues/14833), and it's only gotten more expressive primitives over time. Imagine with me, for a moment. You could pretty easily define the shape of a tsconfig as a type, eg:

```ts
type Config = {
    compilerOptions: { strict: true }
};
```
and then stick it in a known location
```ts
export default Config;
```
and that's that - just look the type pointed at by `default`, and transform it into a config object[^1].
Now that... could work... but composability? Something like
```ts
import ConfigA from "./tsconfig.base";
type Config = {
    compilerOptions: { strict: true } & Omit<NonNullable<ConfigA["compilerOptions"]>, "strict">
};
export default Config;
```
maybe? That could work... it is flexible and expressive enough, but it is a bit wordy, especially compared to options today. Maybe some aliases would help. How's completion support inside the type? Missing? Well, that's a downgrade. Guaranteeing schema conformance would be nice. So we also need a helper like
```ts
type Config<T extends RawTSConfig> = T;
```
so we can write
```ts
import { Config } from "typescript";
import ConfigA from "./tsconfig.base";
type MyConfig = {
    compilerOptions: { strict: true } & Omit<NonNullable<ConfigA["compilerOptions"]>, "strict">
};
export default Config<MyConfig>;
```
and that gives us safety... still no completions, though. Only way to get that would be.... in... an... inference... context...

And this is where you realize we've been going about this exercise all wrong - types are great - we let you write a type that can exactly describe the precise single value of a json object (mostly), but a lot of the greatness of our typesystem isn't in its' constructors, rather it is in how it maps to the host language - javascript. In fact, it's *more* powerful in the context of javascript expressions than what type constructors alone can provide. That's when you realize if you write a definition like
```ts
export function config<const T extends RawTSConfig>(config: T): T;
```
you can invoke it like so
```ts
import * as ts from "typescript";
export default ts.config({
    compilerOptions: { strict: true }
});
```
and get full completions and typechecking on the call, and *still* get exact literal types for the type of the expression, provided all operations done within the expression resolve to exact literal types. Which means we've come full circle - the idealized form of _types as configuration_ is actually
```ts
import * as ts from "typescript";

import base from "./tsconfig.base.js";
import mixin from "./tsconfig.mixin.js";

export default ts.config({
    compilerOptions: {
        ...base.compilerOptions,
        jsx: "react-jsx",
        paths: {
            "dep": ["./node_modules/typescript"],
            ...mixin.compilerOptions.paths
        },
        traceResolution: true
    },
    include: [ "./app/**/*.ts" ]
});
```
or... _visibly identical_ to code as configuration. The only difference being that [I can't believe we didn't use eval](https://github.com/microsoft/TypeScript/commit/e91450c25f5dba8dd249cdbbe59b651dbe501353) to get there[^2].

This PR is a working proof of concept - it will load a `tsconfig.d.ts`, `tsconfig.ts`, or `tsconfig.js` (or any other code extension), and look for the type of the `default` member of that module[^3], convert the type to a config object, and use that config object. For example, the above:
config result of the above, showing merged configs
this should work at a basic level both on the command line and in the editor, but this shouldn't be regarded as anything close to production ready - it'd need tests, a dedicated language service project for type config files, more tests, actual diagnostics for when the `default` type is, say, circularly referential or non-literal, proper integration with caching in the language service layer, a less layered approach to the API (type -> json -> compiler options is indirect - type -> compiler options directly probably yields better UX), and tests.

Still, this is potentially useful, even as incomplete as this - playing around with it has made it apparent, to me at least, just how much I didn't care about the file being executed, but really did just want the syntactic niceties like imports and spreads. And maybe a function call or two. Things not hard to represent purely at compile time in the type system we have today, but that are terribly painful in a JSON document.

[^1]: _Waves hand at the specifics of the transform in the cases of unions, intersections, optionals, generics, etc._ - Only literal, object, and tuple types matter here - you probably wanna error on types that contain anything else.
[^2]: Though you could `eval` it and get the same thing... Unless you void your warranty by casting.
[^3]: _More handwaving about how that module is loaded._ - Suffice to say, you just pick one set of compiler settings to load configs with (`nodenext`-y, `moduleDetection: "force"`) and stick with it.


JoostK:
Is this also targeting 5.5.555 or is this evil (eval?) enough to land in 6.6.666?

xiBread:
I can't even tell if this is an April Fool's joke or not 😭

PR #58028, opened 2024-04-01
Infer type arguments on instance check

### LINKED ISSUE #17473: Infer constrained generic parameters after instanceof check
**TypeScript Version:**  2.4.0

When a variable passes an `instanceof` check generic type with constraints on the type parameter, apply to constraints to the variable.  There is currently no clean way to handle the following without an explicit cast:

**Code**

```ts
interface Foo {
    foo: string;
}

class Bar<T extends Foo> {
    constructor(readonly bar: T) {}
}

let a: any;
if (a instanceof Bar) {
    a.bar; // <-- a.bar should be 'Foo' instead of 'any'
}
```

**Expected behavior:**

`a.bar` is inferred as an instance of `Foo`.

**Actual behavior:**

`a.bar` is `any`

This change would only impact type checking, and will not affect compilation output.

### LINKED ISSUE #46668: Private field check narrows generic class too far
# Bug Report

Discovered a potential issue with https://github.com/microsoft/TypeScript/pull/44648 in regards to how classes with type parameters are handled. The presence of the brand indicates that the object is an instance of this class, but doesn't say anything about the type arguments. For `instanceof` the current behavior is to pass `any` for all type arguments (see https://github.com/microsoft/TypeScript/issues/17473 for that issue) but the default here should be that the type matches the constraint for that corresponding type parameter.

### 🔎 Search Terms

`ts4.5-beta`, `private fields`, `brand checks`, `generic`, `type parameter`, `narrowing`

TypeScript Version: `ts4.5-beta`
- I was unable to test this on prior versions because the feature didn't exist yet.

### ⏯ Playground Link

[Playground link with relevant code](https://www.typescriptlang.org/play?ts=4.5.0-beta#code/C4TwDgpgBAwg9gOwM7AE4EMCWDhQLxQDeUAZnHAFxQqrYDmUAvlAD5FQBG6qVCArgFsOEVEwDcAKAkBjADbokSKAFkQMeYoA8AFSgQAHsAgIAJkvjI0WHAD4iEqI6gBiDhlP4oaPhElOoAG7osj5U2n5O0og0fNLAcKgAFEEhEGEAlPb+-sAAFphIAHQpPp4lvg5OjFL+UWAgAGrBPgBiqHACifpUcBwAVhBxmYSV2VCYJFCJru4m4whQ+sOjY04A9GuL1LlwfLJzwipqGkiaFjTWwDYrq175ReWe+sXNvlAbevqQcWHg0ADk5ys2GA-3GSgQcFwCiQmDoCHQHFk0HiXj+UH+2n+hRuVRW1WqMmiuAA7hgwJATC1yJ4EBASUd1DDEsQyJQoAAiNkcpjpSRRSxQMnoCkQEwAIW4tPpjJOLM43CoAEYAEwAZl5kmFoqp5EKdUarzaHUS2spktQfKJyDgyMKsjgdFN5Mp1LgL1ShTZfPem1AkE+32ASjZABpAq9wQrUBIgA)

### 💻 Code

```ts
type Constraint = { foo: string } | { bar: number };

class MyClass<T extends Constraint> {
    #brand = true;
    value: T;
    constructor(value: T) {
        this.value = value;
    }

    copyValueFrom(x: object) {
        if (#brand in x) {
            this.value = x.value; 
        }
    }
}

const wrappedFoo = new MyClass({ foo: "foo" });
const wrappedBar = new MyClass({ bar: 123 });
wrappedFoo.copyValueFrom(wrappedBar);
console.log(wrappedFoo.value.foo); // type expects foo, value is bar
```

### 🙁 Actual behavior

Assignment in `copyValueFrom` proceeded without error, results in mismatch between type and value.

### 🙂 Expected behavior

In `copyValueFrom`, expected brand to narrow `x` to `MyClass<Constraint>` and report error `Type 'Constraint' is not assignable to type 'T'.` when assigning `this.value = x.value`.

### LINKED ISSUE #57317: Make `instanceof` narrow better using an anonymous class returned from a function

molisani:
Fixes #17473 - Infer constrained generic parameters after instanceof check
Fixes #46668 - Private field check narrows generic class too far

This is an updated and ready-to-be-reviewed recreation of my previous #49863. Please see that PR for a longer description of the problem and the proposed solution, as I don't want to duplicate it all here unnecessarily. In short, when instantiating the type of a class for an instance check, TypeScript would populate all type arguments with `any` no matter the constraint on the type parameter or the strictness settings. This change factors in the compiler strictness as well as the constraints and variances of the type parameters to "infer" the type arguments.

This does *NOT* fix #57317, per @Andarist on #57323 but I have included that test case here as these changes do introduce new behavior to that case (`any` to `unknown`). I can remove if it feels irrelevant to these changes.

## Breaking Changes

From the last PR, I have resolved to use the compiler option `strictInstanceOfTypeParameters` to gate these otherwise breaking changes. However, there was some discussion about the possibility of not hiding these changes behind a compiler option. I can [test enabling these changes for all](https://github.com/microsoft/TypeScript/pull/49863#discussion_r1144030484), similar to what I did last time.


jakebailey:
@typescript-bot test it

fatcerberus:
Wow, it broke approximately everything 😅

jakebailey:
Not surprising if you've seen #49863.

Notably, this new PR adds a strict mode setting. That's what made DT clean, as DT does not set `strict=true`, but rather enables specific strict mode options. (Been on my todo list to look into change that for a while.)

molisani:
@jakebailey If you pack this I can make playground reproductions of some of these errors.

And let me know if you want to test turning it on by default 😅

jakebailey:
@typescript-bot pack this

nmattia:
Hi there @molisani, are there plans to resurrect this? Would be a great addition!

molisani:
I would definitely be interested in refreshing and landing this, but I don't know what the appetite is from the TypeScript team given that it would lead to a lot of breakage and the codebase is currently being migrated to the golang rewrite. Will defer to @jakebailey or @sandersn for a recommendation.

sandersn:
I'd guess it's worth keeping around but we are all still heads-down on porting at the moment. I think it's likely that we'd need to discuss it at a design meeting as well, to decide on the question of breaks vs flags.

PR #58036, opened 2024-04-02
Remove FlowNodes and FlowFlags from public API

jakebailey:
These types are not referenced anywhere in our public API. #57977 modifies the flow nodes, but notes that the change is necessarily breaking.

However, these types are not even used in our public API. To be able use them, you'd already need to be using TS internals to poke at properties that were marked `@internal`. This PR marks them as internal, and you can see that nothing breaks in the output declaration files.

This will break https://github.com/dsherret/ts-ast-viewer/blob/74d0ba30e8bce861279a79e8f7d09634be9034d6/src/components/FlowNodeGraph.tsx at compile time, but #57977 and other changes like it will break it at runtime too. Given one already has to poke at internals, I feel like they should need internals to even have the types too...

### DIFF
--- a/src/compiler/types.ts
+++ b/src/compiler/types.ts
@@ -4070,6 +4070,7 @@ export interface JSDocImportTag extends JSDocTag {
 
 // NOTE: Ensure this is up-to-date with src/debug/debug.ts
 // dprint-ignore
+/** @internal */
 export const enum FlowFlags {
     Unreachable    = 1 << 0,  // Unreachable code
     Start          = 1 << 1,  // Start of flow graph
@@ -4089,6 +4090,7 @@ export const enum FlowFlags {
     Condition = TrueCondition | FalseCondition,
 }
 
+/** @internal */
 export type FlowNode =
     | FlowStart
     | FlowLabel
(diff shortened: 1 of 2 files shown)


jakebailey:
`ts.FlowNode` has 11 hits: https://github.com/search?q=%2F%5Cbts%5C.FlowNode%5Cb%2F+language%3ATypeScript&type=code

- 6 are https://github.com/dsherret/ts-ast-viewer or derivative
- 4 are https://github.com/p42ai/js-assistant
- 1 is https://github.com/literalpie/ngrid/blob/6222f35a92dd03f57d0909cd31be3bafb64e4199/libs/docsi/webpack/src/lib/api-reference/plugins/utils.ts#L100 (5 years old)

I haven't been able to search properly for named imports.

sheetalkamat (approved):

jakebailey:
@dsherret I know that the flow node viewer was a recent and amazing addition to ts-ast-viewer; is a PR like this problematic for you? My impression is that in order to access the info, it already much reach into the internals, so it doesn't seem too far off to also need to do something special here for the types. We have other PRs in progress that will change the flow node API in a breaking way, so I suspect that the viewer will need to be updated in various ways anyway (and probably wasn't consistent in older releases either).

dsherret:
That's fine. I'll deal with whatever happens 😄

Andarist:
>  I'll deal with whatever happens 😄

Real 10x dev right here 😉

DanielRosenwasser (approved):

weswigham (approved):

PR #58040, opened 2024-04-02
[experiment] error on `module` when `namespace` would work

jakebailey:
This is just #58007 but with a hard error.


jakebailey:
@typescript-bot test top800
@typescript-bot run dt

PR #58042, opened 2024-04-02
Add template variable ${configDir} for substitution of config files directory path

### LINKED ISSUE #57485: Add template variable ${configDir} (name to be determined) for file path substitution
### 🔍 Search Terms

"extends", "compilerOptions", "paths", "outdir", "include", "exclude", "typeRoots", "tsconfig", "ability to make tsconfig paths relative to final config"

### ✅ Viability Checklist

### ⭐ Suggestion

Subset of #56436

Today when all the options that are file paths, are always relative to config they are specified in. But that also means that if you are extending tsconfig and you want eg say `outDir` to be same name but relative to your config, you have to re-specify it in the config. The proposal is to use template variable in base config which would mean that the resulting file paths are relative to final config directory. Name is to be determined.

Consider:
```jsonc
// @fileName: /temp/test/base.tsconfig.json
{
    "compilerOptions": {
        "typeRoots": ["node_modules/@types"],
        "outDir": "dist"
    }
}
```

```jsonc
// @fileName: /temp/test/project1/tsconfig.json
{
    "extends": "../base.tsconfig.json"
}
```

Today the resulting config of project1 is:

```jsonc
// @fileName: /temp/test/project1/computed.tsconfig.json
{
    "compilerOptions": {
        "typeRoots": ["../node_modules/@types"],
        "outDir": "../dist"
    }
}
```

Proposal is to write `base.tsconfig.json` as:
```jsonc
// @fileName: /temp/test/base.tsconfig.json
{
    "compilerOptions": {
        "typeRoots": ["${configDir}/node_modules/@types"],
        "outDir": "${configDir}/dist"
    }
}
```

and this should result in computed `tsconfig.json` as:
```jsonc
// @fileName: /temp/test/project1/computed.tsconfig.json
{
    "compilerOptions": {
        "typeRoots": ["./node_modules/@types"],
        "outDir": "./dist"
    }
}
```

Names that came up in design meeting:
- configLocation
- configFolder
- configDirectory
- projectLocation
- projectFolder
- projectDirectory

### 📃 Motivating Example

 A way to specify root config that can specify `include` and/or `outDir` and not needing to write than in each project.
Issues: #29172, #30163, #37227, #45050, #51213

### 💻 Use Cases

sheetalkamat:
- `${configDir}` on any paths is replced with config files directory instead of making it relative to file its declared in.
- This is allowed on all the `isFilePath` options, `paths`, `files`, `include` and `exclude`
- Not allowed to pass from command line

7b1a84825a836557619e100565d51d71d98b1494 is just tests change
9fcd0d6739066e98b8093d05431db7add7d80ee7 actual change

Fixes #57485

TODO:
Actual string to use for replacement from design meeting:
- configLocation
- configFolder
- configDirectory
- projectLocation
- projectFolder
- projectDirectory


tests/baselines/reference/config/convertCompilerOptionsFromJson/Allow trailing comments with jsonSourceFile api.js
```diff
@@ -12,7 +12,7 @@ Fs::
 configFileName:: tsconfig.json
 CompilerOptions::
 {
-  "configFilePath": "tsconfig.json"
+  "configFilePath": "/apath/tsconfig.json"
```
jakebailey, line 15: I've been trying to find the code that changed to make these baselines change; do you know why this changed like this?

sheetalkamat: Its from https://github.com/microsoft/TypeScript/pull/58042/commits/7b1a84825a836557619e100565d51d71d98b1494

src/compiler/commandLineParser.ts
```diff
@@ -3042,6 +3082,98 @@ function parseJsonConfigFileContentWorker(
     }
 }
 
+/** @internal */
+export function handleWatchOptionsConfigDirTemplateSubstitution(
```
weswigham, line 3086: Can we eliminate the `createCopyOnSubstitute` variable here, and have these *always* return a new `options` object? And then at callsites where they want to update the original object, they can explicitly merge the new object into the old one? It seems like every callsite sets this to `true` except one, so it feels like it'd be a bit cleaner that way.

weswigham (commented):
Does a `--showConfig` on a config with paths like these show the input `${configDir}` or the resolved disk paths?

weswigham (commented):
...Do we need to allow escaping `$` do you can actually use a `$` in a path, rather than as one of these templates? Like if I (somehow - the character is forbidden on NTFS) had an actual directory named `${configDir}` - how do you escape from the pattern back to a literal interpretation?

sheetalkamat:
> Does a `--showConfig` on a config with paths like these show the input `${configDir}` or the resolved disk paths?


https://github.com/microsoft/TypeScript/pull/58042/files#diff-dfa255ce269b3c80a416b7c133db82a20292076b8c53370637b1aeaff7b8e7f3R110 shows resolved path and i think its useful to show it like that.

sheetalkamat:
> ...Do we need to allow escaping `$` so you can actually use a `$` in a path, rather than as one of these templates? Like if I (somehow - the character is forbidden on NTFS) had an actual directory named `${configDir}` - how do you escape from the pattern back to a literal interpretation?


Not sure. i think we should keep this simple as otherwise we need to ensure all the paths are normalized to contain escaped `$` we can choose other template pattern as well its a TODO but we shouldnt have to do escaping

andrewbranch (approved):

src/compiler/commandLineParser.ts
```diff
@@ -1034,6 +1037,7 @@ const commandOptionsWithoutBuild: CommandLineOption[] = [
         name: "paths",
         type: "object",
         affectsModuleResolution: true,
+        allowConfigDirTemplateSubstitution: true,
```
jakebailey, line 1040: Long term, I do wonder if we'll need some other flag here given we'll introduce more substitutions, but thankfully this is internal!

jakebailey (approved):

zhangzhuang15:
@sheetalkamat It's a good feature of `${configDir}`, but it's not perfect. Using `${configDir}`, we can make `paths` more simple, avoid trapping into black hole of relative path, such as `"@root/*" : ["../../../../../a/b/c/*"]`. Unfortunately, we still trap into this black hole when we config `extends`. I suggest that we should bring a template variable or a method, let user set it a path, and user can apply it into tsconfig.json or tell typescript figure it out. For example:

`/a/b/tsconfig.variables.json`:
```json
{
  "root": "./"
}
```
now, `${root}` refers to `/a/b/`, and it can be shared in all tsconfig.json under `/a/b/`, such as `/a/b/**/tsconfig.json`;

`/a/b/c/tsconfig.json`:
```json 
{
  "compilerOptions": {
    "paths": {
        "@root/*": ["${root}/*"]
    }
  }
}
```

PR #51682, opened 2022-11-29
Reduce polymorphism resulting from unstable Node shapes

rbuckton:
We have a number of places in our codebase where we optionally attach properties to a `Node`. While a common practice in JS, this can sometimes have negative consequences with regards to runtime performance. Most modern JavaScript engines use Inline Caches (ICs) to optimize code paths, primarily when reading or writing to properties on objects, or to global variables.

Ideally, we want property ICs to be "monomorphic", such that they only ever see a single object shape (or "map" in V8 parlance) as "monomorphic" property ICs provide the fastest lookups. A "polymorphic" IC has slightly worse performance since it shifts from a one-to-one lookup to a short list of "map"-> property entries. If a new "map" is encountered once a polymorphic IC has around four entries, V8 will shift to a "megamporphic" IC. "Megamorphic" property ICs are the worst in terms of performance, as V8 essentially gives up on fast property access and relies on slow property lookups (there is still a cache, but it is cyclical: repeated property lookups for the same "map" will be fast but will age out as new "maps" are encountered).

Monomorphism isn't always possible, especially when accessing properties like `.kind`. However, in cases where we have branched on `.kind`, we should endeavor to ensure that further property accesses within that branch remain monomorphic, or at least polymorphic.

One way to achieve this is to ensure the shape of every `Node` is stable relative to its `kind`. This has the following implications:

- Properties of `Node` and its subtypes really shouldn't be "optional" and should be fully initialized when a `Node` is produced (even if only with `undefined`).
- Avoid casting nodes to things they aren't (there were several cases of `node as any as JSDocContainer` and `node as TypeNode & { <optional properties> }`).
- Move properties off of supertypes that belong on subtypes (such as `symbol`, `localSymbol`, `locals`, `nextContainer`, `flowNode`, etc.), as that results in unchecked `node.symbol` accesses which often result in "wrong map" deoptimizations.
- Avoid using "cover types" with a lot of optional properties in favor of union types where you must discriminate before access.
- Avoid accessing properties other than `kind` when the node is a supertype or union, unless the function is only accessing one or two properties (like `pos`/`end`) since a megamorphic lookup on "pos" will have the same cost as a megamorpic lookup on "kind".
- Where possible, a function or code path should try to target a single `Node` subtype, especially if it accesses multiple properties.

[...]

PR #57703, opened 2024-03-08
[experiment] make tsc use the public API

jakebailey:
Wacky test built on #55326 which makes `tsc` use the public API. Curious how bad the perf of this is.

The below represents effectively the smallest our package is going to get without splitting it into multiple packages.

## Overall package size

|  | Before | After | Diff | Diff (percent) |
| - | - | - | - | - |
| Packed | 3.98 MiB | 2.97 MiB | -1.01 MiB | -25.31% |
| Unpacked | 21.38 MiB | 15.59 MiB | -5.79 MiB | -27.08% |

## Files

|  | Before | After | Diff | Diff (percent) |
| - | - | - | - | - |
| `lib/tsc.js` | 5.79 MiB | 2.31 KiB | -5.79 MiB | -99.96% |

PR #58045, opened 2024-04-02
Monomorphize allocators for tsserver/public API, just like core compiler

jakebailey:
In #51682 and #51880, @rbuckton monomorphized our Nodes and Symbols for the compiler's allocators.

This PR naively copies the property order from the compiler over to the classes we define in `services`, netting similar performance wins for similar memory costs. This should significantly improve performance of tsserver (every editor) and public API users who do anything more than parse (ts-eslint, ts-loader, ts-jest, rush, nx, the list goes on...).

Here are the latest benchmarks when _forcing `tsc` to use these allocators_, which is a good proxy for how this will affect downstream users: https://github.com/microsoft/TypeScript/pull/58045#issuecomment-2036191127

Note that this is _not_ a performance improvement for `tsc` itself; that benefit already happened in TS 5.0!

image

The tsserver benchmarks remain, and should lead to significant improvements in editor performance:

image

A new benchmark tests "tsc via public API" explicitly:

image


jakebailey:

src/services/services.ts
```diff
@@ -534,25 +542,30 @@ function createSyntaxList(nodes: NodeArray<Node>, parent: Node): Node {
...
-    constructor(pos: number, end: number) {
+    constructor(kind: TKind, pos: number, end: number) {
         // Set properties in same order as NodeObject
+        // Note: must match Token and Identifier in src/compiler/utilities.ts
         this.pos = pos;
         this.end = end;
+        this.kind = kind;
+        this.id = 0;
         this.flags = NodeFlags.None;
-        this.modifierFlagsCache = ModifierFlags.None;
```
jakebailey, line 553: The compiler does not set this prop at all. I suspect I could just leave this and get the same effect without the baseline change, at a cost of memory.

jakebailey:
After the results come back, I'll try and rebase this PR such that it targets a branch that has the tsc change in place. The benchmark system is smart enough to compare a PR against its target branch so that will provide better results than comparing two negative results to see which one is less negative.

(I am liking what I'm seeing in the logs so far, though.)

jakebailey:

jakebailey:
Those results are _a lot_ better than https://github.com/microsoft/TypeScript/pull/57703#issuecomment-1986571844.

jakebailey:
Yessss

jakebailey:
Hm, that lost all of the emit speed, without changing memory usage. Not great.

jakebailey:
I... tested the wrong allocators.

jakebailey:
FWIW comparing https://github.com/microsoft/TypeScript/pull/51682#issuecomment-1340222198, this increase memory is not unexpected. Maybe it's fine; this benchmark is already extra strange because I'm emitting stuff, which won't happen in tsserver... normally anyway.

I feel like if https://github.com/microsoft/TypeScript/pull/51682#issuecomment-1340222198 was "good", then this PR is good.

jakebailey:

src/services/services.ts
```diff
@@ -343,8 +345,8 @@ function createNode<TKind extends SyntaxKind>(kind: TKind, pos: number, end: num
     return node;
 }
 
-class NodeObject implements Node {
-    public kind: SyntaxKind;
+class NodeObject<TKind extends SyntaxKind> implements Node {
```
weswigham, line 348: Rather than copy & pasting the constructor ordering, can this not just `extend (objectAllocator.getNodeConstructor())`? This file only overrides the allocator with the services one at the end, so calling it when we construct these classes should get the base ones.

jakebailey:
The `_children` change appears to have even more greatly improved tsserver performance. Awesome.

src/services/services.ts
rbuckton: We don't set it because tokens and identifiers can't have modifiers, so we shave off some memory by ignoring it.

src/services/services.ts
rbuckton: I agree with the sentiment, but I'd be happy to leave it as it is in this PR until we can move all of this to compiler/ and do away with `ObjectAllocator` entirely.

jakebailey: I just tried it and I think it'd work at runtime, but at compile time it gets really awkward because:

- Services adds methods to `Node`, but since `getNodeConstructor()` gives a constructor for `Node`, `NodeObject` needs overrides for everything
- The type arguments on the services constructors conflict with the compiler's allocators, so that all has to go away again
- We can't do the same for Token/Identifier/PrivateIdentifier, as they share methods provided by `TokenOrIdentifierObject` (and they're also different than NodeObject).

So, I'm sort of inclined to leave this "as-is" for the quick win until we can drop the service altogether...

tests/baselines/reference/skipJSDocParsing/deprecated-ParseNone-file.ts.diff
```diff
@@ -38,7 +38,6 @@
 -                                "pos": 6,
```
rbuckton, line 38: Does this change break JSDoc? Why are these missing?

tests/baselines/reference/skipJSDocParsing/see-ParseForTypeInfo-file.ts.diff
```diff
@@ -28,7 +28,6 @@
 -                                "pos": 9,
```
rbuckton, line 28: This seems unintended?

jakebailey: This is a misleading diff; the file is already a diff, so it's a diff of a diff when viewed on github. The only prop going away is `modifierFlagsCache` below.

src/services/services.ts
jakebailey: Yeah, that's what I assumed, though I was surprised given it means that the shape of `Node` is different. But I guess they just don't go down the same paths? Probably worth another investigation (unless you've already done it) as this PR is just trying to copy the compiler for now.

tests/baselines/reference/skipJSDocParsing/deprecated-ParseNone-file.ts.diff
jakebailey: See the other comment, but this is misleading as this is just a removed line in a diff. These tests compare the ASTs with and without JSDoc parsing, we just so happen to not set this prop anymore (like in the other baselines).

jakebailey:
> The `_children` change appears to have even more greatly improved tsserver performance. Awesome.

Unfortunately, it also came with an even greater memory cost. About 1% on top. I can't show the diff because the benchmarks ran on different machines so aren't comparable like that.

jakebailey:
I'm marking this as ready to review, but I'll need to retarget `main` beforehand once we decide what variant of this we think is right.

rbuckton:
I am curious if we end up with multiple maps for `NodeObject` due to how `_children` is allocated, since it starts as `[]` before nodes are added.

jakebailey:
When I look at the deopt explorer, it sure seems like there's still only one, ignoring the extra one that's from the SourceFileObject extension.

image

jakebailey:
Oh scratch that, I'm running it as tsc, which never even asks for the children, so I can't observe that.

jakebailey:

jakebailey:
Uh, that's not good. I guess the WeakMap thing makes things very bad, which is very confusing because tsc never asks for them?

jakebailey:
Oh, the emit always assigns it an undefined value, so every single object creation now includes a map write.

jakebailey:

jakebailey:
Oh, this perf result will be bogus as it's totally broken. Oops.

jakebailey:
It turns out that `_children` is referenced all over the place; simply making it a private prop or weak map is not good enough because it has to be reset and such all over. I'm going to rebase to do the "right" thing just to test, but it's quite invasive.

I have at least confirmed that the object size is increasing when I add `_children`. And using `append` for `createChildren` does not currently help: #58064  (but may after this PR?)

jakebailey:

jakebailey:

tests/baselines/reference/api/typescript.d.ts
```diff
@@ -8040,7 +8040,6 @@ declare namespace ts {
     }
     interface SyntaxList extends Node {
         kind: SyntaxKind.SyntaxList;
-        _children: Node[];
```
jakebailey, line 8043: It's not entirely clear to me why this is here, but it wouldn't be too hard code wise to keep it working. I haven't found any evidence that anyone's actually using it...

jakebailey: I'm going to guess that it's named this way because it's a special node, but still needs to work with `getChildren`.

src/compiler/parser.ts
```diff
@@ -10022,7 +10021,7 @@ namespace IncrementalParser {
         }
 
         function visitArray(array: IncrementalNodeArray) {
-            array._children = undefined;
+            setNodeChildren(array, undefined);
```
jakebailey: Setting children on arrays appears to be pointless; nothing would ever have been able to read them, so this is actually probably a perf improvement itself.

jakebailey: #58069

jakebailey:

tests/baselines/reference/api/typescript.d.ts
rbuckton: `SyntaxList` is created during `getChildren` to handle cases where we parsed a `NodeArray` like `modifiers` for a node that doesn't support them (and thus isn't visited by `forEachChild`). I'm not sure why it's in the public API, so it's probably an oversight.

src/services/services.ts
```diff
@@ -640,8 +660,21 @@ class SymbolObject implements Symbol {
...
+        this.id = 0;
+        this.mergeId = 0;
+        this.parent = undefined;
+        this.members = undefined;
+        this.exports = undefined;
+        this.exportSymbol = undefined;
+        this.constEnumOnlyModule = undefined;
+        this.isReferenced = undefined;
+        this.lastAssignmentPos = undefined;
+        (this as any).links = undefined; // used by TransientSymbol
```
rbuckton, line 677: Unlike the `Symbol` "constructor" in compiler, we can actually define `NodeLinks` on `SymbolObject` and remove the `as any`.

rbuckton (changes requested):

src/services/services.ts
jakebailey: Good point!

src/services/services.ts
```diff
@@ -428,7 +438,7 @@ class NodeObject implements Node {
 
     public getChildren(sourceFile?: SourceFileLike): Node[] {
         this.assertHasRealPosition("Node without a real position cannot be scanned and thus has no token nodes - use forEachChild and collect the result if that's fine");
-        return this._children || (this._children = createChildren(this, sourceFile));
+        return getOrSetNodeChildren(this, () => createChildren(this, sourceFile));
```
rbuckton, line 441: I'm not sure I'm a fan of the closure here. Could we modify `setNodeChildren` to return it's input and do this instead?

```suggestion
        return getNodeChildren(this) ?? setNodeChildren(this, createChildren(this, sourceFile));
```

rbuckton (changes requested):

src/services/services.ts
jakebailey: I can just do that, yeah. This is the only use.

jakebailey: Done.

rbuckton (approved):

jakebailey:
@typescript-bot pack this

jakebailey:
The above looks exactly as expected for this PR targeting main; no change to tsc (it doesn't use this code), and all benefit for tsserver.

jakebailey:
The benchmarking system now includes a benchmark for simulating tsc via the public API; testing:

jakebailey:
Now that the benchmarker can test the public API as tsc:

image

PR #58046, opened 2024-04-02
Update an error message to match TypeScript’s behaviour which allows spreading non objects in objects

### LINKED ISSUE #57957: All primitives can be spread in objects according to MDN but TypeScript prevents it
### 🔎 Search Terms

"spreading primitives in objects"

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about spread.

### ⏯ Playground Link

https://www.typescriptlang.org/play?#code/MYewdgzgLgBAHjAvDA3jAdJgjAGg5qAJwFcBTPTdAMwEMAbCc-dMYuuizYsAE1KoCWYUj07oA5OJgBfANxA

### 💻 Code

```ts
const x = { ...1, ...true, ...false, ...null, ...undefined, ...'' };
```

### 🙁 Actual behavior

I got this TypeScript error:

> error TS2698: Spread types may only be created from object types.

It is wrong because of this quote from MDN on [Spread Syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#description):

> All [primitives](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#primitive_values) can be spread in objects. Only strings have enumerable own properties, and spreading anything else doesn't create properties on the new object.
> 
> ```js
> const obj = { ...true, ..."test", ...10 };
> // { '0': 't', '1': 'e', '2': 's', '3': 't' }
> ```

### 🙂 Expected behavior

I expected no error and `x` to be equal to `{}`.

_No response_

geryogam:
Update this error message

> error TS2698: Spread types may only be created from object types.

to this one

> error TS2698: Spread appears unintentional because this value is not possibly an object.

Fixes #57957

### DIFF
--- a/tests/baselines/reference/spreadTypeVariable.errors.txt
+++ b/tests/baselines/reference/spreadTypeVariable.errors.txt
@@ -1,13 +1,13 @@
-spreadTypeVariable.ts(2,12): error TS2698: Spread types may only be created from object types.
-spreadTypeVariable.ts(10,12): error TS2698: Spread types may only be created from object types.
-spreadTypeVariable.ts(14,12): error TS2698: Spread types may only be created from object types.
+spreadTypeVariable.ts(2,12): error TS2698: Spread appears unintentional because this value is not possibly an object.
+spreadTypeVariable.ts(10,12): error TS2698: Spread appears unintentional because this value is not possibly an object.
+spreadTypeVariable.ts(14,12): error TS2698: Spread appears unintentional because this value is not possibly an object.
 
 
 ==== spreadTypeVariable.ts (3 errors) ====
     function f1<T extends number>(arg: T) {
       return { ...arg };
                ~~~~~~
-!!! error TS2698: Spread types may only be created from object types.
+!!! error TS2698: Spread appears unintentional because this value is not possibly an object.
     }
     
     function f2<T extends string[]>(arg: T) {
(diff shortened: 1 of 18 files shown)


geryogam:

tests/baselines/reference/spreadTypeVariable.errors.txt
```diff
@@ -17,13 +17,13 @@ spreadTypeVariable.ts(14,12): error TS2698: Spread types may only be created fro
     function f3<T extends number | string[]>(arg: T) {
       return { ...arg }
                ~~~~~~
-!!! error TS2698: Spread types may only be created from object types.
+!!! error TS2698: Spread appears unintentional because this value is not possibly an object.
```
RyanCavanaugh, line 20: Need better logic here - `T` is constrained to `string[]` which is an object

tests/baselines/reference/spreadTypeVariable.errors.txt
```diff
@@ -17,13 +17,13 @@ spreadTypeVariable.ts(14,12): error TS2698: Spread types may only be created fro
...
                ~~~~~~
-!!! error TS2698: Spread types may only be created from object types.
+!!! error TS2698: Spread appears unintentional because this value is not possibly an object.
     }
     
     function f4<T extends number | { [key: string]: any }>(arg: T) {
       return { ...arg }
                ~~~~~~
-!!! error TS2698: Spread types may only be created from object types.
+!!! error TS2698: Spread appears unintentional because this value is not possibly an object.
```
RyanCavanaugh, line 26: Incorrect logic here; `T` is not constrained to only primitives

RyanCavanaugh (changes requested):

tests/baselines/reference/spreadTypeVariable.errors.txt
geryogam: @RyanCavanaugh

RyanCavanaugh: @geryogam This is likely going to be quite difficult, and unfortunately I can't allocate time to help new contributors with a "You can send a PR if you want to"-tier issue.

geryogam: Okay, since you think it's going to be quite difficult and you're the development lead for the TypeScript team at Microsoft, I don't think I'll even try to go any further, as I don't have much time to devote to open source contributions at the moment. Thanks for the information.

sandersn:
To help with PR housekeeping, I'm going to close this PR since it's pretty old now.

PR #58047, opened 2024-04-02
[experiment] playing around with Node/Token/Identifier allocators

jakebailey:
Something I noticed on #58045. I have no proof having not yet tested it, but it sure feels weird that `Node` gets `modifierFlagsCache` and `original` props but `Token`/`Identifier` do not. I would have expected those two to have differing shapes such that we stop being monomorphic on some paths.

Technically I can add the props to the others, but, trying out just using Node anyway?


jakebailey:

jakebailey:
Ah, yes, memory.

jakebailey:

jakebailey:
aight

Issue #44659, opened 2021-06-18
Crash when using a combination of a switch statement, optional chaining, an implicit any array, and ts strict settings.


VincentGuinaudeau:
# Bug Report

I discovered this crash today, when trying to build my JS project with typescript (I want to start introducing some TS files).

Once I refined the reproduction scenario, I was able to make any version of typescript crash, on any environment (playground, CLI, LSP).

### 🔎 Search Terms

crash, strict, optional chaining, implicit any

### 🕗 Version & Regression Information

- This is a crash
- This occurs on versions 3.7.2 and onward (tried with @next)
- 3.6.5 shows diagnotics, but doesn't handle optional chaining

### ⏯ Playground Link

No diagnostic is shown, despite the array missing a type annotation.

The crash can be fixed by :
 * adding a type annotation to the array (even `any []`)
 * removing the optional chaining
 * removing the access to `foo[0]`
 * turning the `switch` into an `if`
 * turning off `strictNullChecks`
 * turning off `noImplicitAny`

https://www.typescriptlang.org/play?&strict=true#code/DYUwLgBAZg9jEF4IG0C6BYAUFgzgdwEswBjACwgApYYB+AOlADsBzMUgSggG8sBIYgIY4QEAIwAuPr2rIADBkwBfIA

### 💻 Code

I know full well how bad this code is, it has already been fixed.

Pinpointing the issue was hard though because typescript doesn't provide any human readable trace as to what code it was parsing when the error occurred.

`tsconfig.json`
```json
{
	"compilerOptions": {
		"outDir": "./dist",

		"strictNullChecks": true,
		"noImplicitAny": true
	},
	"include": ["index.ts"]
}
```

`index.ts`
```typescript
let foo = []

switch (foo?.length) {
	case 1:
		foo[0]
}
```

```
# with typescript 4.3.4
$ tsc
$HOME/.npm_global/lib/node_modules/typescript/lib/tsc.js:90672
                throw e;
                ^

TypeError: Cannot read property 'length' of undefined
    at isFunctionObjectType ($HOME/.npm_global/lib/node_modules/typescript/lib/tsc.js:54183:47)
    at getTypeFacts ($HOME/.npm_global/lib/node_modules/typescript/lib/tsc.js:54227:21)
    at $HOME/.npm_global/lib/node_modules/typescript/lib/tsc.js:54260:60
    at filterType ($HOME/.npm_global/lib/node_modules/typescript/lib/tsc.js:54485:43)
[...]

PR #58049, opened 2024-04-03
Bailout early from `isFunctionObjectType` for evolving arrays

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/44659


src/compiler/checker.ts
```diff
@@ -27044,6 +27044,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     }
 
     function isFunctionObjectType(type: ObjectType): boolean {
+        if (getObjectFlags(type) & ObjectFlags.EvolvingArray) {
```
Andarist, line 27047: `resolveStructuredTypeMembers` is still going to crash if an evolving array manages to reach it. I concluded that it's still better - at least in this case - to bail out early. IIUC, it avoid "fixing" this evolving type without a strong reason to do so - so the evolving array can continue evolving in certain situations.

jakebailey:
@typescript-bot test it

src/compiler/checker.ts
sandersn: Are there other paths whereby an evolving array could reach resolveStructuredTypeMembers? This seems like an important limitation that needs to be checked, either multiple places or maybe within resolveStructuredTypeMembers itself.

Andarist: > Are there other paths whereby an evolving array could reach resolveStructuredTypeMembers? 

None that we are aware of. Any call with an evolving array to that function would throw and I didn't see any other crashes reported there. It's of course not completely unlikely that there are other ways to end up with that crash.

> This seems like an important limitation that needs to be checked, either multiple places or maybe within resolveStructuredTypeMembers itself.

There is already an assert there. The only thing that potentially could be improved is to switch the assert to some other logic that would handle evolving arrays gracefully. I don't know what that logic would be and I think if this crash is a rare one then this fix is good enough.

andrewbranch (approved):

jakebailey (approved):

PR #58053, opened 2024-04-03
[experiment] error on comparisons where the only comparable type is undefined

Zzzen:
For more detailed information, checkout https://github.com/typescript-eslint/typescript-eslint/issues/5602
As far as I know, there are no typescript issues related. I'm curious to witness how it performs on actual code in real-world scenarios. Would anyone be available to help me by executing the "top800" test suite?

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -38875,7 +38875,14 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                         error(errorNode, Diagnostics.This_condition_will_always_return_0_since_JavaScript_compares_objects_by_reference_not_value, eqType ? "false" : "true");
                     }
                     checkNaNEquality(errorNode, operator, left, right);
-                    reportOperatorErrorUnless((left, right) => isTypeEqualityComparableTo(left, right) || isTypeEqualityComparableTo(right, left));
+                    const hasReported = reportOperatorErrorUnless((left, right) => isTypeEqualityComparableTo(left, right) || isTypeEqualityComparableTo(right, left));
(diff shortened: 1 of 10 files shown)


jakebailey:
@typescript-bot test top800
@typescript-bot run dt

jakebailey:
Very surprisingly few hits, but I think they are all good hits.

I feel like we've had a PR like this before (did I send one?), but I can't remember where it is.

RyanCavanaugh:
https://github.com/supabase/supabase/blob/e5b647878d3c9defba4d61db12bd3752b41bd31a/apps/studio/components/layouts/AccountLayout/AccountLayout.types.ts#L8 🙂

jakebailey:
Probably not, but:

RyanCavanaugh:
It seems super weird to have a carve-out to say that you're allowed to write `string | 0 === boolean | 0` but not `string | undefined === boolean | undefined`, especially when `string === undefined` is allowed.

A possible alternate strategy is to only remove `missingType` from both operands (and do so unconditionally rather than requiring two rounds like we have here). This would make this code error:
```ts
type Str = {
  foo?: string;
}
type Num = {
  bar?: number;
}

function fn(x: Str, y: Num) {
  return x.foo === y.bar;
}
```
which feels like the motivating case and easier to justify in terms of behavioral inconsistency.

Zzzen:
>A possible alternate strategy is to only remove missingType from both operands (and do so unconditionally rather than requiring two rounds like we have here). This would make this code error:

`missingType` won't work if exactOptionalPropertyTypes is off?

PR #58057, opened 2024-04-03
Export executeCommandLine from public API (internal for now)

jakebailey:
This extracts just the `executeCommandLine` and public API changes from #57703. My intent is that I can make a mini tsc using the public API such that we can long-term perf test changes like #58045 which only affect the public API, without writing all of the code to directly use Program/build mode.

~I've opted to export the `executeCommandLine` function in public with a simplified signature, but I could also just leave this all internal.~

Exporting this increases the package size by 40KB, not that bad.

## Overall package size

| - | - | - | - | - |
| Packed | 3.77 MiB | 3.77 MiB | +8.05 KiB | +0.21% |
| Unpacked | 20.38 MiB | 20.42 MiB | +40.52 KiB | +0.19% |

## Files

| - | - | - | - | - |
| `lib/typescript.d.ts` | 561.75 KiB | 561.97 KiB | +228.00 B | +0.04% |
| `lib/typescript.js` | 8.18 MiB | 8.22 MiB | +40.30 KiB | +0.48% |


tests/baselines/reference/api/typescript.d.ts
```diff
@@ -11220,5 +11220,7 @@ declare namespace ts {
      * @param compilerOptions Optional compiler options.
      */
     function transform<T extends Node>(source: T | T[], transformers: TransformerFactory<T>[], compilerOptions?: CompilerOptions): TransformationResult<T>;
+    function executeCommandLine(system: System, cb: ExecuteCommandLineCallbacks, commandLineArgs: readonly string[]): void;
+    type ExecuteCommandLineCallbacks = (program: Program | BuilderProgram | ParsedCommandLine) => void;
```
sheetalkamat, line 11224: Would prefer to keep this internal so that we dont have to maintain compatibility and stuff

src/typescript/_namespaces/ts.ts
```diff
@@ -4,5 +4,6 @@ export * from "../../compiler/_namespaces/ts";
 export * from "../../jsTyping/_namespaces/ts";
 export * from "../../services/_namespaces/ts";
 export * from "../../server/_namespaces/ts";
+export * from "../../executeCommandLine/_namespaces/ts";
```
sheetalkamat, line 7: Also if this is going to be in public API, we could just move this to compiler instead of in separate folder ?

tests/baselines/reference/api/typescript.d.ts
jakebailey: All of it? Or just the callback?

src/typescript/_namespaces/ts.ts
jakebailey: Yeah, it's just one file, so that would work too.

tests/baselines/reference/api/typescript.d.ts
sheetalkamat: All of it

weswigham (approved):

jakebailey:
I've made it internal again, but, I will note that some people do actually want to use this API in public, though. I don't think it's a big deal to leave it internal for the time being in any case.

sheetalkamat (approved):

jakebailey:

jakebailey:
Good, it's working.

image

PR #58058, opened 2024-04-03
Sort imports with dprint instead of eslint

jakebailey:
A few people have expressed that they'd prefer if the formatter did this rather than eslint. dprint can do this a lot faster without the linter errors; if you save on format it'll be extra good.

The downside is that we will lose automatic grouping, until maybe https://github.com/dprint/dprint-plugin-typescript/issues/493. That's something I was relying on in #51455 to ensure the namespace imports always came first, but that could be achieved another way.

I also discovered https://github.com/dprint/dprint-plugin-typescript/issues/620, but that's not a very bad bug.

I had to disable formatting in `_namespaces`, as those import orders matter and dprint does not have the ability to ignore regions: https://github.com/dprint/dprint-plugin-typescript/issues/411

### DIFF
--- a/.vscode/settings.template.json
+++ b/.vscode/settings.template.json
@@ -18,7 +18,7 @@
         ".git-blame-ignore-revs"
     ],
 
-    // Match eslint-plugin-simple-import-sort in organize/auto-imports.
+    // Match dprint in organize/auto-imports.
     "typescript.unstable": {
         "organizeImportsCollation": "unicode",
         "organizeImportsCaseFirst": "upper",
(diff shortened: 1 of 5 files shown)


andrewbranch (approved):

sheetalkamat (approved):

.vscode/settings.template.json
```diff
@@ -18,7 +18,7 @@
         ".git-blame-ignore-revs"
     ],
 
-    // Match eslint-plugin-simple-import-sort in organize/auto-imports.
+    // Match dprint in organize/auto-imports.
```
jakebailey, line 21: While here, I should probably also set the setting for inline type imports.

jakebailey: Updated; verified that it works too, nice.

PR #58061, opened 2024-04-03
Shorten error spans for errors reported on constructor declarations

Andarist:


src/compiler/utilities.ts
```diff
@@ -2268,6 +2268,16 @@ export function getErrorSpanForNode(sourceFile: SourceFile, node: Node): TextSpa
             const pos = skipTrivia(sourceFile.text, (node as JSDocSatisfiesTag).tagName.pos);
             return getSpanOfTokenAtPosition(sourceFile, pos);
         }
+        case SyntaxKind.Constructor: {
+            const constructorDeclaration = node as ConstructorDeclaration;
+            const start = skipTrivia(sourceFile.text, constructorDeclaration.pos);
+            const scanner = createScanner(sourceFile.languageVersion, /*skipTrivia*/ true, sourceFile.languageVariant, sourceFile.text, /*onError*/ undefined, start);
+            while (scanner.scan() !== SyntaxKind.ConstructorKeyword) {
```
jakebailey, line 2275: Seems pretty hefty to need an entire rescan to understand this; the info above doesn't seem to do this either. Is the position not just after the modifiers?

Andarist: > Seems pretty hefty to need an entire rescan to understand this;

this is actually what some of the other branches do here indirectly (see the calls to `getSpanOfTokenAtPosition`)

> Is the position not just after the modifiers?

that would roughly be the `start` position of the `constructor` keyword but for some of the errors it makes sense to highlight the modifiers and I think there is little value in differentiating this span between different error types (highlighting modifiers is fine since they usually are placed on the same line anyway). The bigger problem is when it comes to finding the `end` position of this span. I've tried to use `node.parameters.start` but it wasn't too precise. That's how I landed on this solution

DanielRosenwasser: I have a fear that some bad parse recovery mechanism will possibly make this code end up in an endless loop. Consider bailing on an EOF token or something similar.

Andarist: This should be highly-unlikely cause we know that we are already processing `ConstructorDeclaration` and I can't imagine a situation in which that would be created without the `ConstructorKeyword`. Being extra safe is probably for the better though - I applied the suggested change

jakebailey: Yeah, I didn't realize that `getSpanOfTokenAtPosition` creates a scanner each time. Definitely expensive but not new for this.

Why didn't `getSpanOfTokenAtPosition` work for this again? I guess I'm not sure why we'd want to highlight modifiers at all.

Andarist: I was worried about `classConstructorOverloadsAccessibility` but taking another look at it right now... it doesn't look bad if we only highlight the constructor keyword there:
<details>
<summary>git diff</summary>

```diff
diff --git a/src/compiler/utilities.ts b/src/compiler/utilities.ts
index f13946d637..5965e78154 100644
--- a/src/compiler/utilities.ts
+++ b/src/compiler/utilities.ts
@@ -2274,12 +2274,12 @@ export function getErrorSpanForNode(sourceFile: SourceFile, node: Node): TextSpa
         }
         case SyntaxKind.Constructor: {
             const constructorDeclaration = node as ConstructorDeclaration;
-            const start = skipTrivia(sourceFile.text, constructorDeclaration.pos);
-            const scanner = createScanner(sourceFile.languageVersion, /*skipTrivia*/ true, sourceFile.languageVariant, sourceFile.text, /*onError*/ undefined, start);
+            const scanner = createScanner(sourceFile.languageVersion, /*skipTrivia*/ true, sourceFile.languageVariant, sourceFile.text, /*onError*/ undefined, skipTrivia(sourceFile.text, constructorDeclaration.pos));
             let token = scanner.scan();
             while (token !== SyntaxKind.ConstructorKeyword && token !== SyntaxKind.EndOfFileToken) {
                 token = scanner.scan();
             }
+            const start = skipTrivia(sourceFile.text, scanner.getTokenStart());
             const end = scanner.getTokenEnd();
             return createTextSpanFromBounds(start, end);
         }
diff --git a/tests/baselines/reference/classConstructorOverloadsAccessibility.errors.txt b/tests/baselines/reference/classConstructorOverloadsAccessibility.errors.txt
index e939262b02..1b0723b506 100644
--- a/tests/baselines/reference/classConstructorOverloadsAccessibility.errors.txt
+++ b/tests/baselines/reference/classConstructorOverloadsAccessibility.errors.txt
@@ -1,15 +1,15 @@
-classConstructorOverloadsAccessibility.ts(2,2): error TS2385: Overload signatures must all be public, private or protected.
-classConstructorOverloadsAccessibility.ts(3,2): error TS2385: Overload signatures must all be public, private or protected.
-classConstructorOverloadsAccessibility.ts(11,2): error TS2385: Overload signatures must all be public, private or protected.
+classConstructorOverloadsAccessibility.ts(2,9): error TS2385: Overload signatures must all be public, private or protected.
+classConstructorOverloadsAccessibility.ts(3,12): error TS2385: Overload signatures must all be public, private or protected.
+classConstructorOverloadsAccessibility.ts(11,12): error TS2385: Overload signatures must all be public, private or protected.
 
 
 ==== classConstructorOverloadsAccessibility.ts (3 errors) ====
     class A {
     	public constructor(a: boolean) // error
-    	~~~~~~~~~~~~~~~~~~
+    	       ~~~~~~~~~~~
 !!! error TS2385: Overload signatures must all be public, private or protected.
     	protected constructor(a: number) // error
-    	~~~~~~~~~~~~~~~~~~~~~
+    	          ~~~~~~~~~~~
 !!! error TS2385: Overload signatures must all be public, private or protected.
     	private constructor(a: string)
     	private constructor() { 
@@ -19,7 +19,7 @@ classConstructorOverloadsAccessibility.ts(11,2): error TS2385: Overload signatur
     
     class B {
     	protected constructor(a: number) // error
-    	~~~~~~~~~~~~~~~~~~~~~
+    	          ~~~~~~~~~~~
 !!! error TS2385: Overload signatures must all be public, private or protected.
     	constructor(a: string)
     	constructor() { 
diff --git a/tests/baselines/reference/parserConstructorDeclaration8.errors.txt b/tests/baselines/reference/parserConstructorDeclaration8.errors.txt
index 8b37b60364..c798654628 100644
--- a/tests/baselines/reference/parserConstructorDeclaration8.errors.txt
+++ b/tests/baselines/reference/parserConstructorDeclaration8.errors.txt
@@ -1,4 +1,4 @@
-parserConstructorDeclaration8.ts(3,3): error TS2390: Constructor implementation is missing.
+parserConstructorDeclaration8.ts(3,10): error TS2390: Constructor implementation is missing.
 parserConstructorDeclaration8.ts(3,21): error TS1005: '(' expected.
 
 
@@ -6,7 +6,7 @@ parserConstructorDeclaration8.ts(3,21): error TS1005: '(' expected.
     class C {
       // Not a constructor
       public constructor;
-      ~~~~~~~~~~~~~~~~~~
+             ~~~~~~~~~~~
 !!! error TS2390: Constructor implementation is missing.
                         ~
 !!! error TS1005: '(' expected.

```
</details>

jakebailey: I see what you mean, yeah. Maybe it's best as-is. That or `TS2385` should have been on the modifiers the whole time?

Andarist: > That or TS2385 should have been on the modifiers the whole time?

I'm not sure what do u mean by "that" here :p TS2385 - that's the one I think might be worth keeping over modifiers too. I don't feel super strong about it though

jakebailey: I was just saying that we could either do:

- This PR as is, squiggle over modifiers
- Change `TS2385` to only squiggle on modifiers, change this PR to only squiggle on `constructor`.

But the latter seems like work.

@DanielRosenwasser probably should say what feels best here; I don't know if the existing change is a problem, really.

Andarist: > But the latter seems like work.

Yeah, this isn't 100% straightforward because right now those errors are reported on nodes and u'd like to raise over a span or nodes array or something. I'm not sure if there is an existing infra for this at the higher level (closer to the checker - where the errors are raised). I'd prefer not to investigate this right now 😉

jakebailey (approved):
FWIW I think this PR is fine as-is, and after merging main could probably safely go in.

Andarist:
@jakebailey it's synced with `main` now :)

PR #58064, opened 2024-04-03
Use append in createChildren

jakebailey:
Related: https://github.com/microsoft/TypeScript/pull/58045#issuecomment-2035569234

It's pretty dubious that `getChildren` returns a _mutable_ array of nodes. I could return `emptyArray` but that scares me.


jakebailey:

jakebailey:
No significant change, but maybe that's just because without #58045, services are not very optimized.

jakebailey:
Going to close this; didn't seem to help and now this is stored off in a weakmap anyway.

PR #58065, opened 2024-04-03
[Experiment] Do not watch failed looksup of resolution if we are not going to generate diagnostics

sheetalkamat:
Things that will improve:
- Not watching many resolutions so should be more performant
- Less watchers

Things that I am worried about:
- On linux npm ci will cause watching `node_modules` and then closing (assuming all resolutions succeed) - creating recursive watcher is not cheap and needs to iterate all folders
- dependency on options => getResolutionDiagnostics => Sharing resolutions in tsserver will need to handle and wont be able to share resolutions if there are more options (like noImplicitAny) etc need to depend on affectsModuleResolution
- Are there scenarios where file is resolved but new npm ci takes precedence and is preferred (eg. instead of `@types` , actual `node_modules` resolutions etc and should that be handled


sheetalkamat:
@typescript-bot pack this

sheetalkamat:
Closing this for now as testing reveals things go hay wire on linux with npm i and other program updates as we keep adding and removing watchers

PR #58067, opened 2024-04-03
Skip visits to child nodes of entity names in visitExistingNodeTreeSymbols

weswigham:
This just skips visiting some children of entity names where we'd just repeat the same symbol resolution repeatedly (if we lookup `A` in `A.B`, we don't need to also visit `A` and `B`, so long as we attach symbol info for `A` during that first lookup), or would never actually resolve (because it's the name of the declaration).

I'm not sure if this'd actually show up on any perf tests (but it should definitely be good for perf - we're just doing less repeated work), but what it does do is make any future users of `trackSymbol` (eg, `isolatedDeclarations` resolver/checker) need to handle fewer cases.

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -8678,15 +8678,28 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 else {
                     context.tracker.trackSymbol(sym, context.enclosingDeclaration, meaning);
                 }
-                if (isIdentifier(node)) {
-                    const type = getDeclaredTypeOfSymbol(sym);
-                    const name = sym.flags & SymbolFlags.TypeParameter ? typeParameterToName(type, context) : factory.cloneNode(node);
-                    name.symbol = sym; // for quickinfo, which uses identifier symbol information
-                    return { introducesError, node: setTextRange(setEmitFlags(setOriginalNode(name, node), EmitFlags.NoAsciiEscaping), node) };
-                }
+                return { introducesError, node: attachSymbolToLeftmostIdentifier(node) as T };
             }
 
             return { introducesError, node };
+
+            /**
+             * Attaches a `.symbol` member to an identifier, cloning it to do so, so symbol information
+             * is smuggled out for symbol display information.
+             */
+            function attachSymbolToLeftmostIdentifier(node: Node): Node {
+                if (node === leftmost) {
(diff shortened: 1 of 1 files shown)


src/compiler/checker.ts
```diff
@@ -8855,11 +8868,13 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
                 if (isEntityName(node) || isEntityNameExpression(node)) {
+                    if (isDeclarationName(node)) {
+                        return node;
+                    }
                     const { introducesError, node: result } = trackExistingEntityName(node, context);
                     hadError = hadError || introducesError;
-                    if (result !== node) {
-                        return result;
-                    }
+                    // We should not go to child nodes of the entity name, they will not be acesible
```
jakebailey, line 8876:
```suggestion
                    // We should not go to child nodes of the entity name, they will not be accessible
```

src/compiler/checker.ts
```diff
@@ -8678,15 +8678,28 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+                return { introducesError, node: attachSymbolToLeftmostIdentifier(node) as T };
             }
 
             return { introducesError, node };
+
+            /**
+             * Attaches a `.symbol` member to an identifier, cloning it to do so, so symbol information
+             * is smuggled out for symbol display information.
+             */
+            function attachSymbolToLeftmostIdentifier(node: Node): Node {
```
jakebailey, line 8690: Why not `<T extends Node>` here to avoid the cast above?

weswigham: Because that needs an `as T` cast in the `node === leftmost` branch anyway, so I'm just choosing which place of the two I'd rather cast at.

jakebailey: Ah, yeah.

weswigham:
@typescript-bot perf test this though I doubt this'll register, since it'll probably only affect the time it takes to issue some overcomplicated errors?

jakebailey:
Oh yeah you can't just stick random garbage onto the bot commands, they're passed through (I intend to have arguments to perf testing in the near future).

weswigham:
note to self: make it not fail if the next word isn't "faster"

jakebailey (approved):
Seems fine, though it's a shame there's not really a testable thing here...

PR #58069, opened 2024-04-03
Don't clear _children on NodeArrays

jakebailey:
Noticed this in #58045; `_children` is never read by anyone from arrays. There isn't a need to clear it. And clearing it probably causes worse perf as we're tacking yet another random prop on.

No doubt, we won't see any perf change in the tester since this only affected incremental mode, but it's probably real in the editor.


jakebailey:

jakebailey:
Not significant, but still an ok cleanup and one that will probably happen due to #58045 anyway...

rbuckton (approved):

PR #58075, opened 2024-04-04
Always consider parameters in scope visible to node builder

weswigham:
Parameter symbols in scope resolve, thanks to recent changes, but don't always have a visible declaration (if their host signature isn't directly exported) - this doesn't matter in the node builder, since we're cloning that host signature in some way anyway, so if it's a parameter and we're looking at it, it aughta be visible.


tests/baselines/reference/cyclicGenericTypeInstantiationInference.types
```diff
@@ -2,8 +2,8 @@
 
 === cyclicGenericTypeInstantiationInference.ts ===
 function foo<T>() {
->foo : <T>() => { y2: { y2: { y2: { y2: { y2: { y2: { y2: { y2: { y2: { y2: { y2: any; }; }; }; }; }; }; }; }; }; }; }
->    : ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+>foo : <T>() => { y2: typeof z; }
```
weswigham, line 5: At first you'd think this is good; but `z` isn't actually in scope...? And isn't a parameter?

src/compiler/checker.ts
```diff
@@ -8672,6 +8672,16 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             const meaning = getMeaningOfEntityNameReference(node);
             const sym = resolveEntityName(leftmost, meaning, /*ignoreErrors*/ true, /*dontResolveAlias*/ true);
             if (sym) {
+                // If a parameter is resolvable in the current context it is also visible, so no need to go to symbol accesibility
+                if (
+                    sym.flags & SymbolFlags.FunctionScopedVariable
+                    && sym.valueDeclaration
+                ) {
+                    const parameter = sym.valueDeclaration && isParameterDeclaration(sym.valueDeclaration);
```
dragomirtitian, line 8680: The problem is that this will not work for binding elements. The original check was mean to walk up the binding elements. I this i lost something when I copied it over.

weswigham: The confusingly-named `isParameterDeclaration` does not do what you think  it does - it _does_ walk binding elements.

jakebailey: Yeah, we should really rename that to `isPartOfParameterDeclaration`. See also #52283

jakebailey (approved):

PR #58079, opened 2024-04-04
Special case check for this identifiers to skip exhaustive scope traversal

weswigham:
This PR adds a branch to `trackExistingEntityName` to handle `this`. `isEntityNameVisible` has a similar branch as a fallback when `resolveName` fails - `resolvename` will _always_ fail for a `this` identifier, since `this`, being a keyword, isn't a bindable name. By doing this check *first*, we can save on doing the exhaustive traversal of scopes trying to resolve the name.

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -8671,6 +8671,13 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 return { introducesError, node };
             }
             const meaning = getMeaningOfEntityNameReference(node);
+            if (isThisIdentifier(leftmost)) {
+                if (isSymbolAccessible(getSymbolOfDeclaration(getThisContainer(leftmost, /*includeArrowFunctions*/ false, /*includeClassComputedPropertyName*/ false)), leftmost, meaning, /*shouldComputeAliasesToMakeVisible*/ false).accessibility !== SymbolAccessibility.Accessible) {
+                    introducesError = true;
+                    context.tracker.reportInaccessibleThisError();
+                }
+                return { introducesError, node };
+            }
             const sym = resolveEntityName(leftmost, meaning, /*ignoreErrors*/ true, /*dontResolveAlias*/ true);
             if (sym) {
                 // If a parameter is resolvable in the current context it is also visible, so no need to go to symbol accesibility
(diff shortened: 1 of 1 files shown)


dragomirtitian:
There are currently there is no actual test for this. The test that failed for me was `tests/cases/compiler/declarationEmitTypeofThisInClass.ts` but my branch also contained a change that always puts nodes through the node builder instead of going though the type node reuse logic in the declaration emit (with the ultimate goal of reducing the handling of types in declaration emit) A test that would actually exercise this would be one that reuses an assertion type:

```ts
// tests/cases/compiler/declarationEmitTypeofThisInClass.ts
class Foo {
    public foo!: string
    public bar!: typeof this.foo 
    public baz = null! as typeof this.foo; // new 
}
```

weswigham:
We discussed this a bit, and as-is, this is only a +perf change - right now, there's no way to witness a difference between `this` not resolving (and returning the input symbol), and `this` resolving (and returning the input symbol). However, there may be some cases where `this` *shouldn't* resolve, but is written (eg, we don't allow `this` in object literal types), where, if the node builder is used to reuse the node, it should report a private `this` error (which is also currently reported during checking, making the error during declaration emit a smidge redundant unless you skip checking).

jakebailey:
I guess I was expecting to see an improvement, but, oh well 😄

jakebailey (approved):
Seems fine to me but @sandersn should take a peek if he knows what's going on here.

sandersn:
This change seems plausible -- but maybe resolution isn't that slow? I don't remember any subtleties with `this` but I haven't looked too deeply at its resolution before.

weswigham:
> I guess I was expecting to see an improvement, but, oh well

I think you'd have to use `this` types a *lot* before it'd be enough to show up.

> I don't remember any subtleties with this but I haven't looked too deeply at its resolution before.

It isn't handled by `resolveName` - everything that is supposed to handle a `this` symbol seems to do a container lookup like this on the spot. Honestly, we should *probably* move the logic into `resolveName` at some point, so every place `resolveName` is used doesn't also have a fallback (fall-forward?) for `this` identifiers like this, seeings as `resolveName` already handles `arguments`, which is also a keyword with similar scoping behaviors.

PR #58080, opened 2024-04-04
[experiment] services monomorphization + tsc using public API

jakebailey:
Temp PR to perf test #58045 + #57703, i.e. "how much perf difference is there between tsc versus tsc+services allocators?"


jakebailey:
Pretty good, not too far off tsc anymore.

jakebailey:
Closing, since we'll be able to test this in a different way now without that patch.

PR #58083, opened 2024-04-04
Don't propagate partial union/intersection properties between caches

### LINKED ISSUE #58050: type assertions affect the type narrowing of the subsequent code
### 🔎 Search Terms

Type Assertion type Narrowing 

- This changed between versions 5.3.3 and 5.4.3

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.4.3#code/JYWwDg9gTgLgBAbzjAnmApnAwhcEB26+MANMmpgAoA2ArgObD5wC+cAZlLnAOQButdDwBQwgPRjyGbHAC82XJELEA2j3wBDEEIC6w9AA9IsKZhx5lMAOrAYACwCS+AM4wN1agB4AKgD45cN5wAGRwNAxM+kbQ8KjSNvZOru7U4YzM8kgA+mBQ6OzABgD8AFxwrlBM9Kyihsbw7LT4AMYwwARwAO62ji5uHp5YcIYwRAAmzgoWRKSBwwaj+BNwCb3JHmlMvgAUwnBwzYoEM2VDAD7YKjoke8gaUPToMKWBwgCUZRVViLeHfQdHSyrJL9agBbZuB5POBFIoA6bEN5wDSTcxKGbAvopHy+X4EVzw9HESbyACCUCgGhQADpgM5yZSUNtDgiYEi4SyifAyipOcdiHo8f8kJptKwAnzLM4VAAGHRwW7Adhwbai9BIhC3fZ5GC0KDMNW3FiibVPPXMABEFuExuEQA

### 💻 Code

```ts
import { type Component, type Plugin } from 'vue'

// type C = Component['name']
export type ComponentWithInstall<T> = T & Plugin
export type WithInstallPlugin = { _prefix?: string }

export function withInstall<C extends Component, T extends WithInstallPlugin>(
  component: C | C[],
  target?: T
): string {
  const componentWithInstall = (target ?? component) as ComponentWithInstall<T>
  const components = Array.isArray(component) ? component : [component]

  const { name } = components[0] 
  if (name) {
    return name
  }

  return ""
}

```

### 🙁 Actual behavior

image
the name property is underlined in red, indicating an error.
> Property 'name' does not exist on type 'Component'.(2339)

The `type C = Component['name']` is valid, as the `name` property exists within the `Component` type. When I remove `const componentWithInstall = (target ?? component) as ComponentWithInstall<T>`, everything returns to normal. Alternatively, if I uncomment `type C = Component['name']`, even though it's not used anywhere, the error disappears. Type assertions should not affect the type narrowing of subsequent code, but an error occurs. This code worked fine in version 5.3.3, but started to error in version 5.4.3.

### 🙂 Expected behavior

Ts don't report the error.

_No response_

ahejlsberg:
There's currently no repro in this PR since it is really hard to reproduce the conditions in a stand-alone manner.

Fixes #58050.

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -14900,8 +14900,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     // these partial properties when identifying discriminant properties, but otherwise they are filtered out
     // and do not appear to be present in the union type.
     function getUnionOrIntersectionProperty(type: UnionOrIntersectionType, name: __String, skipObjectFunctionPropertyAugment?: boolean): Symbol | undefined {
-        let property = type.propertyCacheWithoutObjectFunctionPropertyAugment?.get(name) ||
-                !skipObjectFunctionPropertyAugment ? type.propertyCache?.get(name) : undefined;
+        let property = skipObjectFunctionPropertyAugment ?
+            type.propertyCacheWithoutObjectFunctionPropertyAugment?.get(name) :
+            type.propertyCache?.get(name);
         if (!property) {
             property = createUnionOrIntersectionProperty(type, name, skipObjectFunctionPropertyAugment);
             if (property) {
@@ -14909,7 +14910,8 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     type.propertyCacheWithoutObjectFunctionPropertyAugment ||= createSymbolTable() :
                     type.propertyCache ||= createSymbolTable();
                 properties.set(name, property);
-                if (skipObjectFunctionPropertyAugment && !type.propertyCache?.get(name)) {
+                // Propagate an entry from the non-augmented cache to the augmented cache unless the property is partial.
+                if (skipObjectFunctionPropertyAugment && !(getCheckFlags(property) & CheckFlags.Partial) && !type.propertyCache?.get(name)) {
                     const properties = type.propertyCache ||= createSymbolTable();
                     properties.set(name, property);
                 }
--- a/tests/baselines/reference/checkJsxChildrenCanBeTupleType.types
+++ b/tests/baselines/reference/checkJsxChildrenCanBeTupleType.types
@@ -4,7 +4,7 @@
 Assignability cache: 2,200 / 2,200 (nearest 100)
 Type Count: 7,800 / 7,800 (nearest 100)
 Instantiation count: 90,000 / 90,000 (nearest 500)
-Symbol count: 66,500 / 66,500 (nearest 500)
+Symbol count: 67,000 / 67,000 (nearest 500)
 
 === checkJsxChildrenCanBeTupleType.tsx ===
 /// <reference path="react16.d.ts" />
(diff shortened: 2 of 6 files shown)


ahejlsberg:
@typescript-bot test it

Andarist:
minimal repro case
```ts
interface ComponentOptions<Props> {
  setup?: (props: Props) => void;
  name?: string;
}

interface FunctionalComponent<P> {
  (props: P): void;
}

type ConcreteComponent<Props> =
  | ComponentOptions<Props>
  | FunctionalComponent<Props>;

type Component<Props = {}> = ConcreteComponent<Props>;

type WithInstallPlugin = { _prefix?: string };

export function withInstall<C extends Component, T extends WithInstallPlugin>(
  component: C | C[],
  target?: T,
): string {
  const componentWithInstall = (target ?? component) as T;
  const components = Array.isArray(component) ? component : [component];

  const { name } = components[0];
  if (name) {
    return name;
  }

  return "";
}
```

Andarist:
I was experiencing what I would assume to be out-of-order checking in the IDE so at some point I pasted the already slimmed down repro into the playground and continued there. Since it was a single file I could test it out quite reliably there - I didn't even run this with `tsc` though so I can't say how it behaved for me in this context.

> Meanwhile, this even shorter example reproduces in the IDE, but not with the command-line compiler:

This is *very* strange 😅

DanielRosenwasser:
I think it would be good to cherry-pick this into 5.4.5.

jakebailey:
Here's a fourslash test based on https://github.com/microsoft/TypeScript/pull/58083#issuecomment-2039606741 which fails before this PR but passes after:

```ts
/// <reference path="fourslash.ts" />

// @strict: true
// @lib: esnext

//// interface ComponentOptions<Props> {
////   setup?: (props: Props) => void;
////   name?: string;
//// }
//// 
//// interface FunctionalComponent<P> {
////   (props: P): void;
//// }
//// 
//// type ConcreteComponent<Props> =
////   | ComponentOptions<Props>
////   | FunctionalComponent<Props>;
//// 
//// type Component<Props = {}> = ConcreteComponent<Props>;
//// 
//// type WithInstallPlugin = { _prefix?: string };
//// 
//// 
//// /**/
//// export function withInstall<C extends Component, T extends WithInstallPlugin>(
////   component: C | C[],
////   target?: T,
//// ): string {
////   const componentWithInstall = (target ?? component) as T;
////   const components = Array.isArray(component) ? component : [component];
//// 
////   const { name } = components[0];
////   if (name) {
////     return name;
////   }
//// 
////   return "";
//// }

verify.noErrors();

goTo.marker();
edit.insert("type C = Component['name']");

verify.noErrors();
```

I'm observing the opposite of what @ahejlsberg is seeing, in that the above code fails in the editor for me but there's no errors when made into a compiler test. But I think it matches the original issue well enough to be a test for this PR, though.

jakebailey:
@typescript-bot pack this

jakebailey:
Confirmed that this test works just like the original issue in the playground; will push here since we want the patch out very shortly.

jakebailey:

jakebailey (approved):

Issue #56513, opened 2023-11-22
Make Intl locales arrays readonly


Shraymonks:
### ⚙ Compilation target

Any

### ⚙ Library

intl.d.ts

### Missing / Incorrect Definition

The `locales` array should be typed as a readonly array in Intl constructors and methods.

Currently they typed as normal arrays which causes a type error when passing a readonly array such as `navigator.languages`.

Resulting in the following:
`Argument of type 'readonly string[]' is not assignable to parameter of type 'string | string[] | undefined'.
  The type 'readonly string[]' is 'readonly' and cannot be assigned to the mutable type 'string[]'.(2345)`

There was a previous PR https://github.com/microsoft/TypeScript/pull/50135 that fixed it for one particular call site. However there are still several instances that also have this issue:
https://github.com/microsoft/TypeScript/blob/d4fbc9b57d9aa7d02faac9b1e9bb7b37c687f6e9/src/lib/es2016.intl.d.ts#L12

https://github.com/microsoft/TypeScript/blob/d4fbc9b57d9aa7d02faac9b1e9bb7b37c687f6e9/src/lib/es2018.intl.d.ts#L33-L36

https://github.com/microsoft/TypeScript/blob/d4fbc9b57d9aa7d02faac9b1e9bb7b37c687f6e9/src/lib/es2020.intl.d.ts#L203
https://github.com/microsoft/TypeScript/blob/d4fbc9b57d9aa7d02faac9b1e9bb7b37c687f6e9/src/lib/es2020.intl.d.ts#L226

https://github.com/microsoft/TypeScript/blob/d4fbc9b57d9aa7d02faac9b1e9bb7b37c687f6e9/src/lib/es2021.intl.d.ts#L128
https://github.com/microsoft/TypeScript/blob/d4fbc9b57d9aa7d02faac9b1e9bb7b37c687f6e9/src/lib/es2021.intl.d.ts#L146

https://github.com/microsoft/TypeScript/blob/d4fbc9b57d9aa7d02faac9b1e9bb7b37c687f6e9/src/lib/es2022.intl.d.ts#L74
https://github.com/microsoft/TypeScript/blob/d4fbc9b57d9aa7d02faac9b1e9bb7b37c687f6e9/src/lib/es2022.intl.d.ts#L88

### Sample Code

```TypeScript
new Intl.DateTimeFormat(navigator.languages)
```

### Documentation Link

[...]

PR #58084, opened 2024-04-05
intl.d.ts: cleanup, add missing features, fix library discrepancies

Renegade334:
Recent work on the Intl library definitions revealed a lot of outstanding inaccuracies and inconsistencies, and so I've taken a deep dive through ECMA-402 and done some significant housekeeping.

This is a big PR, and I've itemised the changes for ease of review. The majority are straightforward and uncontroversial updates to bring things in line with the spec, and it seems sensible just to have them as one PR.

A couple of proposed changes have breaking potential or are related to the typing meta, and are marked 🚩 to indicate that they may need discussion.

#### Typings
- Where a literal union is used more than once, it is now declared as a type in order to avoid duplication, under the `{ConstructorName}Options{PropertyName}` naming convention. This practice started to be used here and there in later lib.intl updates, but is now consistent throughout.
- Changed the convention for types derived from "option registry" interfaces.
  - Before, these types were simply aliases to `keyof R`, which works fine for validation, but isn't very helpful in tsserver output.
    - `type Intl.NumberFormatOptionsStyle = keyof NumberFormatOptionsStyleRegistry` – so what are the valid options?
  - Now, these types are of the form `R[keyof R]`, which ensures that they're resolved to a union type.
    - `type Intl.NumberFormatOptionsStyle = "decimal" | "percent" | "currency"` – much more accessible.
  - This mirrors the convention already used in the core library (`WeakKey` etc.)
- 🚩 Renamed a small number of remaining option literal union types from `{ConstructorName}{PropertyName}` to `{ConstructorName}Options{PropertyName}`. These are exported types, so in theory this is breaking if anyone is importing those specific union types from the library; however, the maintenance and DX benefits of a universal namespace convention seem to support this change. (There are [no resulting errors](https://github.com/microsoft/TypeScript/pull/58084#issuecomment-2161948705) in the user test suite.)
  - There is the question of whether the old names would need keeping as deprecated aliases if this change were made.
- 🚩 Removed the type `UnicodeBCP47LanguageTag`, which is simply a primitive type alias to `string`. This gets optimised away by the compiler, and serves no useful purpose; there are no other examples of "primitive alias as usage descriptor" elsewhere in the library.
#### Documentation
- MDN links and copied documentation have been removed throughout. I'm guessing these were added due to a sense of mistaken equivalence with the DOM APIs, but they shouldn't be present on ECMA definitions.
- Added/updated docblocks for all methods and properties of Intl prototypes.
- Docblocks for each individual option interface property have been removed. This declutters the library significantly, for no significant impact on DX: for self-explanatory options, the description adds nothing, whereas for complex ones, a one-sentence summary doesn't eliminate the need to read the manual.
- Docblocks for constructors have been removed, in keeping with the rest of the library.
#### Tests
- Conformance tests are consolidated into a single testfile per lib.intl version.
- Many of the existing intl tests were implementation tests copied from MDN, rather than type validation tests, which has been addressed.

### Intl
- Moved the definition of `Intl.supportedValuesOf` from es2022 to es2023.
- Added override for `Intl.getCanonicalLocales` to es2020.
- Added missing `prototype` properties of Intl constructors.

### Intl.Collator
- 🚩 Removed string union of valid collations. This list is subject to change in CLDR, and exactly which collations are supported is implementation-dependent. It doesn't seem like the sort of list that should be hard-coded here.
  - A list of implementation-supported collations is now accessible via es2023's `Intl.supportedValuesOf()`, if runtime validation is desired.
- Moved `collation` option from es5 to es2021.
- The collator options `caseFirst` and `numeric` only exist if they are supported by the implementation and are absent if not, hence these are typed as optional.

### Intl.DateTimeFormat
- Removed `bigint` from argument types in format functions, as explicitly not supported.
- Narrowed `ResolvedDateTimeFormatOptions` property types in es5.
- Moved `hourCycle` option from es2020 to es2018.
- Moved `dayPeriod`, `dateStyle`, `timeStyle` options from es2020 to es2021.
- Progressive typing of `timeZoneName`, with values `{short,long}Offset` and `{short,long}Generic` moved to es2022.
- Added missing part types `relatedYear` and `yearName` to es2020.
- Removed duplicate interface properties.

### Intl.DisplayNames
- Moved the definition from es2020 to es2021.
- `DisplayNamesOptions` properties may be `undefined`.
- Progressive typing of `DisplayNamesOptionsType`, with values `calendar` and `dateTimeField` moved to es2022.
- Moved `languageDisplay` option to es2022.
- New `DisplayNameConstructor` interface.

### Intl.ListFormat
- Added `ListFormatPart` interface.
- Renamed some existing option literal union types from `ListFormat{PropertyName}` to `ListFormatOptions{PropertyName}`.
- New `ListFormatConstructor` interface.

### Intl.Locale
- Refactored, with the `Locale` prototype no longer being typed as a child interface of `LocaleOptions`.
- New `LocaleConstructor` interface.
- `LocaleOptions` properties may be `undefined`.
- Correctly type `Locale` properties as required/readonly.
- The properties `caseFirst` and `numeric` only exist if they are supported by the implementation and are absent if not, hence these are typed as optional.

### Intl.NumberFormat
- Removed `bigint` from the original definition of the `formatToParts` method in es2018 (prior to the language feature's existence), and added a new override to es2020.bigint.

### Intl.PluralRules
- Added the missing `selectRange` method to es2023.
- Added "NumberFormat v3" options to PluralRules in es2023.
- Under `ResolvedPluralRulesOptions`, `{minimum,maximum}FractionDigits` may be omitted from es2023 onwards, and these are therefore typed as optional throughout. This mirrors the equivalent change to `ResolvedNumberFormatOptions` in #56902.
- Constructor cannot be invoked without `new`.

### Intl.RelativeTimeFormat
- Added the missing `numberingSystem` option.
- Partitioned part and unit types for clarity.
- Renamed some existing option literal union types from `RelativeTimeFormat{PropertyName}` to `RelativeTimeFormatOptions{PropertyName}`.

### Intl.Segmenter
- `Segments#containing` may return `undefined` if the specified index does not exist.
- New `SegmenterConstructor` interface.

- Implement the second-generation overrides for `Array.prototype.toLocaleString` and `TypedArray.prototype.toLocaleString` under es2020.array, which were omitted from #57679.
- Move first-generation overrides `String.prototype.toLocale{Lower,Upper}Case` to es2015.core.
- Ensure the `locales` argument to all first-generation overrides accepts `readonly string[]`, as per #56513.
- Use `NumberFormatOptions` in the definition of `BigInt.prototype.toLocaleString` in es2020.bigint, and remove the duplicate NumberFormat definitions.
- `toLocaleString` method of `BigInt64Array` and `BigUint64Array` prototypes updated to the second-generation override signature.
- Update docblocks for locale methods, including es5 placeholder methods.
- 🚩 Widen the type of the `options` parameter of `Array.prototype.toLocaleString`.
  - The current type assumes that the only objects overriding `toLocaleString` will be builtin number-like/date-like objects, whose `toLocaleString` method will pass its parameters to either NumberFormat or DateTimeFormat.
  - However, the standard for `Array.prototype.toLocaleString` doesn't specify this; it will pass its parameters to _any_ object's `toLocaleString` method, including user-defined methods. All that ECMA-262 asks is that those parameters not be used for anything other than the standard `toLocaleString` parameter pattern.
  - Accordingly, this function should, by right, remain agnostic about its elements' `toLocaleString` implementations &ndash; providing they obey the standardised function signature.
  - To achieve this, the type of the options parameter could be widened to `object`. However, this comes at the expense of not type-checking any option properties, even for arrays of builtin elements which were covered just fine by the previous behaviour.
  - The alternative is to leave the parameter as-is, and accept the above limitations.
  - In an ideal world, we would conditionally infer the options parameter type from `T`, but [this leads to issues with infinite recursion](https://github.com/microsoft/TypeScript/pull/58084#issuecomment-2161658973), so doesn't look like it's an option.
- 🚩 Discussion on #57679 covered the question of whether the `locales` parameter of first- and second-generation overrides should be required or optional when an es5 placeholder already exists.
  - The verdict there was that the first parameter should be required, so that the new override and the es5 placeholder don't have overlapping signatures.
  - However, this disallows usage like the following:
    ```ts
    numberArray.toLocaleString(undefined, { minimumIntegerDigits: 3 });
    ```
    ...which is a valid way of selecting the default locale while also specifying custom format options. Keeping the first parameter optional is an easy way of allowing this usage case.
  - It's only the array methods added by the above PR that currently use the "mandatory first parameter" signature; all other existing first- and second-generation overrides have the locales parameter as optional.


RyanCavanaugh:
@typescript-bot test it

Renegade334:
One blooper thrown up by top400, in that I'd managed to partially revert #47720, which has now been restored. The remaining errors are consistent with the removal/renaming of type exports as detailed above.

Re-running top400 might be useful to clarify the lay of the land.

jakebailey:
@typescript-bot test it

src/lib/es2015.core.d.ts
```diff
@@ -43,7 +43,16 @@ interface Array<T> {
...
-    toLocaleString(locales: string | string[], options?: Intl.NumberFormatOptions & Intl.DateTimeFormatOptions): string;
+    /**
+     * Returns a string representation of an array.
+     * Each element is converted to a string using its `toLocaleString` method, then concatenated using the list separator of the current locale.
+     * @param locales Passed as the `locales` parameter to each array element's `toLocaleString` method.
+     * @param options Passed as the `options` parameter to each array element's `toLocaleString` method.
+     */
+    toLocaleString(
+        locales?: string | readonly string[],
+        options?: (T extends { toLocaleString(locales?: any, options?: infer O): string; } ? (options: O) => void : never) extends (options: infer P) => void ? P : unknown,
```
Renegade334, line 54: See notes towards the bottom of the PR description.

This appears to work well, including on all of the examples in the test suite (see baseline diffs under 31d8a82). The only edge case I can see is that it doesn't resolve nested arrays.

The simpler alternative is to leave the type as `Intl.NumberFormatOptions & Intl.DateTimeFormatOptions`, which covers the builtin `toLocaleString` methods just fine. I'm ambivalent towards the change, but thought it would be worth proposing.

src/lib/es2018.intl.d.ts
```diff
@@ -18,24 +17,37 @@ declare namespace Intl {
...
+         */
         select(n: number): LDMLPluralRule;
+
+        /** Returns the locale and options computed during initialization of this `PluralRules` instance. */
+        resolvedOptions(): ResolvedPluralRulesOptions;
     }
 
     interface PluralRulesConstructor {
         new (locales?: string | readonly string[], options?: PluralRulesOptions): PluralRules;
-        (locales?: string | readonly string[], options?: PluralRulesOptions): PluralRules;
```
Renegade334, line 34: Cannot be invoked without `new`.

src/lib/es2018.intl.d.ts
```diff
@@ -60,6 +71,21 @@ declare namespace Intl {
...
+    }
+
+    type DateTimeFormatOptionsHourCycle = "h11" | "h12" | "h23" | "h24";
+
+    interface DateTimeFormatOptions {
+        hourCycle?: DateTimeFormatOptionsHourCycle | undefined;
+    }
+
+    interface ResolvedDateTimeFormatOptions {
+        hourCycle?: DateTimeFormatOptionsHourCycle;
```
Renegade334, line 89: Intentionally optional.

src/lib/es2020.intl.d.ts
```diff
@@ -1,227 +1,125 @@
...
+        /** The character collation associated with the locale. */
+        readonly collation: string | undefined;
+        /** The hour cycle type associated with the locale. */
+        readonly hourCycle: DateTimeFormatOptionsHourCycle | undefined;
+        /** The primary language subtag associated with the locale. */
+        readonly language: string;
+        /** The number system used by the locale. */
+        readonly numberingSystem: string | undefined;
+        /** The numeric sorting rule associated with the locale. */
+        readonly numeric?: boolean | undefined;
```
Renegade334, line 38: Intentionally optional.

src/lib/es2020.intl.d.ts
```diff
@@ -1,227 +1,125 @@
...
+        minimize(): Locale;
+        /** Returns the locale's full identifier string. */
+        toString(): string;
+
+        /** The Unicode Language Identifier associated with the locale. */
+        readonly baseName: string;
+        /** The calendar system associated with the locale. */
+        readonly calendar: string | undefined;
+        /** The case sorting rule associated with the locale. */
+        readonly caseFirst?: CollatorOptionsCaseFirst | undefined;
```
Renegade334, line 28: Intentionally optional.

src/lib/es2021.intl.d.ts
```diff
@@ -1,148 +1,153 @@
...
+        timeStyle?: DateTimeFormatOptionsTimeStyle | undefined;
+        dayPeriod?: DateTimeFormatOptionsDayPeriod | undefined;
+        fractionalSecondDigits?: DateTimeFormatOptionsFractionalSecondDigits | undefined;
+    }
+
+    interface ResolvedDateTimeFormatOptions {
+        dateStyle?: DateTimeFormatOptionsDateStyle;
+        timeStyle?: DateTimeFormatOptionsTimeStyle;
+        dayPeriod?: DateTimeFormatOptionsDayPeriod;
+        fractionalSecondDigits?: DateTimeFormatOptionsFractionalSecondDigits;
```
Renegade334, line 26: Intentionally optional.

src/lib/es2021.intl.d.ts
```diff
@@ -1,148 +1,153 @@
...
+        region: never;
+        script: never;
+        currency: never;
+    }
+    type DisplayNamesOptionsType = keyof DisplayNamesOptionsTypeRegistry;
+
+    interface DisplayNamesOptions {
+        localeMatcher?: LocaleMatcherAlgorithm | undefined;
+        style?: DisplayNamesOptionsStyle | undefined;
+        type: DisplayNamesOptionsType;
```
Renegade334, line 64: Intentionally required.

src/lib/es2021.intl.d.ts
```diff
@@ -1,148 +1,153 @@
...
+         * @param code The code to convert.
+         */
+        of(code: string): string | undefined;
+
+        /** Returns the locale and options computed during initialization of this `DisplayNames` instance. */
+        resolvedOptions(): ResolvedDisplayNamesOptions;
+    }
+
+    interface DisplayNamesConstructor {
+        new (locales: LocalesArgument, options: DisplayNamesOptions): DisplayNames;
```
Renegade334, line 87: Parameters intentionally required.

src/lib/es2022.intl.d.ts
```diff
@@ -1,99 +1,78 @@
...
+    }
+
+    type DisplayNamesOptionsLanguageDisplay = "dialect" | "standard";
+
+    interface DisplayNamesOptions {
+        languageDisplay?: DisplayNamesOptionsLanguageDisplay | undefined;
+    }
+
+    interface ResolvedDisplayNamesOptions {
+        languageDisplay?: DisplayNamesOptionsLanguageDisplay;
```
Renegade334, line 21: Intentionally optional.

src/lib/es5.d.ts
```diff
@@ -4382,35 +4398,60 @@ declare var Float64Array: Float64ArrayConstructor;
...
     }
 
     interface ResolvedCollatorOptions {
         locale: string;
-        usage: string;
-        sensitivity: string;
+        usage: CollatorOptionsUsage;
+        sensitivity: CollatorOptionsSensitivity;
         ignorePunctuation: boolean;
         collation: string;
```
Renegade334, line 4425: The resolved `collation` property remains here, even though the option to override it in `CollatorOptions` has moved to es2021.

src/lib/es5.d.ts
```diff
@@ -4382,35 +4398,60 @@ declare var Float64Array: Float64ArrayConstructor;
...
-        caseFirst: string;
-        numeric: boolean;
+        caseFirst?: CollatorOptionsCaseFirst;
+        numeric?: boolean;
```
Renegade334, line 4427: Intentionally optional.

src/lib/es2018.intl.d.ts
```diff
@@ -18,24 +17,37 @@ declare namespace Intl {
         pluralCategories: LDMLPluralRule[];
         type: PluralRuleType;
         minimumIntegerDigits: number;
-        minimumFractionDigits: number;
-        maximumFractionDigits: number;
+        minimumFractionDigits?: number;
+        maximumFractionDigits?: number;
```
Renegade334, line 21: Intentionally optional.

src/lib/es2021.intl.d.ts
```diff
@@ -1,148 +1,153 @@
...
+        fractionalSecond: never;
     }
 
+    type DateTimeFormatOptionsDateStyle = "full" | "long" | "medium" | "short";
+    type DateTimeFormatOptionsTimeStyle = "full" | "long" | "medium" | "short";
+    type DateTimeFormatOptionsDayPeriod = "narrow" | "short" | "long";
+    type DateTimeFormatOptionsFractionalSecondDigits = 1 | 2 | 3;
+
     interface DateTimeFormatOptions {
-        formatMatcher?: "basic" | "best fit" | "best fit" | undefined;
```
Renegade334, line 7: Defined in es5, and doesn't appear as a resolved property.

src/lib/es2021.intl.d.ts
```diff
@@ -1,148 +1,153 @@
...
+    type DisplayNamesOptionsStyle = "narrow" | "short" | "long";
+    type DisplayNamesOptionsFallback = "code" | "none";
+
+    interface DisplayNamesOptionsTypeRegistry {
+        language: never;
+        region: never;
+        script: never;
+        currency: never;
+    }
+    type DisplayNamesOptionsType = keyof DisplayNamesOptionsTypeRegistry;
```
Renegade334, line 59: - es2021: language, region, script, currency
- es2022: calendar, dateTimeField

src/lib/es5.d.ts
```diff
@@ -4464,31 +4502,59 @@ declare namespace Intl {
...
+    type DateTimeFormatOptionsDay = "2-digit" | "numeric";
+    type DateTimeFormatOptionsHour = "2-digit" | "numeric";
+    type DateTimeFormatOptionsMinute = "2-digit" | "numeric";
+    type DateTimeFormatOptionsSecond = "2-digit" | "numeric";
+
+    interface DateTimeFormatOptionsTimeZoneNameRegistry {
+        short: never;
+        long: never;
+    }
+    type DateTimeFormatOptionsTimeZoneName = keyof DateTimeFormatOptionsTimeZoneNameRegistry;
```
Renegade334, line 4544: - es5: short, long
- es2022: shortOffset, longOffset, shortGeneric, longGeneric

Renegade334:
> Here are the results of running the top 400 repos comparing `main` and `refs/pull/58084/merge`:
> 

date-fns and formatjs have been patched to avoid referencing these types altogether, so these errors should no longer appear.

jupyterlab would just need to add the es2021.intl library in order to use the dateStyle and timeStyle options of DateTimeFormat.

src/lib/es2021.intl.d.ts
```diff
@@ -1,148 +1,153 @@
...
-     */
+    type DisplayNamesOptionsStyle = "narrow" | "short" | "long";
+    type DisplayNamesOptionsFallback = "code" | "none";
+
+    interface DisplayNamesOptionsTypeRegistry {
+        language: "language";
+        region: "region";
+        script: "script";
+        currency: "currency";
+    }
```
Renegade334, line 58: - es2021: language, region, script, currency
- es2022: calendar, dateTimeField

src/lib/es5.d.ts
```diff
@@ -4464,31 +4502,59 @@ declare namespace Intl {
...
+    type DateTimeFormatOptionsMonth = "2-digit" | "numeric" | "narrow" | "short" | "long";
+    type DateTimeFormatOptionsDay = "2-digit" | "numeric";
+    type DateTimeFormatOptionsHour = "2-digit" | "numeric";
+    type DateTimeFormatOptionsMinute = "2-digit" | "numeric";
+    type DateTimeFormatOptionsSecond = "2-digit" | "numeric";
+
+    interface DateTimeFormatOptionsTimeZoneNameRegistry {
+        short: "short";
+        long: "long";
+    }
```
Renegade334, line 4543: - es5: short, long
- es2022: shortOffset, longOffset, shortGeneric, longGeneric

jakebailey:
@typescript-bot test it

src/lib/es2015.core.d.ts
```diff
@@ -43,7 +43,13 @@ interface Array<T> {
...
     copyWithin(target: number, start: number, end?: number): this;
 
-    toLocaleString(locales: string | string[], options?: Intl.NumberFormatOptions & Intl.DateTimeFormatOptions): string;
+    /**
+     * Returns a string representation of an array.
+     * Each element is converted to a string using its `toLocaleString` method, then concatenated using the list separator of the current locale.
+     * @param locales Passed as the `locales` parameter to each array element's `toLocaleString` method.
+     * @param options Passed as the `options` parameter to each array element's `toLocaleString` method.
+     */
+    toLocaleString(locales?: string | readonly string[], options?: object): string;
```
Renegade334, line 52: See notes towards the bottom of the PR description for discussion regarding the type of `options` here.

I'm ambivalent towards the change, but thought it would be worth proposing.

jakebailey:
@typescript-bot test it

Renegade334:
> Starting jobs; this comment will be updated as builds start and complete.
> 
> | Command | Status | Results |
> | --- | --- | --- |
> | `test top400` | [✅ Started](https://typescript.visualstudio.com/cf7ac146-d525-443c-b23c-0d58337efebc/_build/results?buildId=162179) | [👀 Results](https://github.com/microsoft/TypeScript/pull/58084#issuecomment-2162049043) |
> | `user test this` | [✅ Started](https://typescript.visualstudio.com/cf7ac146-d525-443c-b23c-0d58337efebc/_build/results?buildId=162180) | [✅ Results](https://github.com/microsoft/TypeScript/pull/58084#issuecomment-2161985538) |
> | `run dt` | [✅ Started](https://typescript.visualstudio.com/cf7ac146-d525-443c-b23c-0d58337efebc/_build/results?buildId=162178) | [✅ Results](https://github.com/microsoft/TypeScript/pull/58084#issuecomment-2161977198) |
> | `perf test this faster` | [✅ Started](https://typescript.visualstudio.com/cf7ac146-d525-443c-b23c-0d58337efebc/_build/results?buildId=162181) | [👀 Results](https://github.com/microsoft/TypeScript/pull/58084#issuecomment-2161989798) |

#### perf
- Small increase in symbol counts, likely primarily due to new method overrides on `Array.prototype` and `String.prototype`
- No significant change in overall performance

#### top400
- jupyterlab:
  - DateTimeFormatOptions `dateStyle` and `timeStyle` properties no longer present in ES2020.Intl
    - fix: add ES2021.Intl to libs
- vscode:
  - ResolvedCollatorOptions `numeric` property may be `undefined`
    - fix: `?? false`
  - `UnicodeBCP47LocaleIdentifier` no longer exists
    - fix: replace all instances with `string`

jakebailey:
> `UnicodeBCP47LocaleIdentifier` no longer exists
> * fix: replace all instances with `string`

We definitely do not want to be deleting types from lib.d.ts.

Renegade334:
Keep and deprecate?

jakebailey:
Yes, or keep using (if it makes sense).

src/lib/es2020.intl.d.ts
```diff
@@ -1,456 +1,240 @@
 /// <reference lib="es2018.intl" />
 declare namespace Intl {
-    /**
-     * A string that is a valid [Unicode BCP 47 Locale Identifier](https://unicode.org/reports/tr35/#Unicode_locale_identifier).
-     *
-     * For example: "fa", "es-MX", "zh-Hant-TW".
-     *
-     * See [MDN - Intl - locales argument](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl#locales_argument).
-     */
-    type UnicodeBCP47LocaleIdentifier = string;
```
Renegade334, line 10: todo: either restore or add deprecated type export.

While this type doesn't do any harm, primitive aliasing feels like a paradigm that should either be used everywhere in the library, or not used at all. This is the only case of it throughout lib.d.ts.

wartab:
I don't know if this is related, but since TypeScript 5.5, we have some issue with changes in Number.ToLocaleString().

```ts
// Defined once to avoid copy pasting this everywhere.
const percentOptions = {
    minimumFractionDigits: 2,
    maximumFractionDigits: 2,
    style: "percent",
};

let output1 = number1.toLocaleString(undefined, percentOptions);
let output2 = number2.toLocaleString(undefined, percentOptions);
let output3 = number3.toLocaleString(undefined, percentOptions);
```

Typescript will shout about this, because "percent" is narrowed to string instead of the subset of valid style values.
Now, of course, we can write this, but nobody is going to remember this from memory and this is just TypeScript getting in the way.

```ts
const percentOptions = {
    minimumFractionDigits: 2,
    maximumFractionDigits: 2,
    style: "percent" as keyof Intl.NumberFormatOptionsStyleRegistry,
};
```

Sorry if this is the wrong place to mention this, but I wasn't sure if opening a new issue was the way to go.

Renegade334:
@wartab: This was intentionally changed previously to make the enum-style string option types consistent.

For your example, using `as const` would be simpler than `as keyof...`, or you could influence the shape of the options object as a whole by using a `satisfies` clause or explicitly typing the variable as an `Intl.NumberFormatOptions` object.

3ru:
Hi @Renegade334, I came across here while looking into a few pieces of missing Intl API surface in TypeScript’s builtin libs, and I wanted to say I think there’s still some valuable work in this PR that would be great to get upstreamed.

If it would be helpful, I’d also be happy to help split out a few smaller follow-up PRs

Renegade334:
All yours. 🫡

PR #58085, opened 2024-04-05
Replace addUndefined in serializeTypeForDeclaration with the actual declaration

weswigham:
Which, given the name of the function, makes sense. This also allows the emit pipeline to prefer a specific declaration to pull type nodes from (which matters when there are duplicate declarations, for example).

This has quite a big diff to baselines (largely in the "good" direction), as we start to copy the input declaration more reliably in many situations.

In addition to just changing the API, this also has to include a fix to apply `.singleQuote` preference when copying input nodes (to minimize fourslash changes, as we reuse more nodes - this change has no baseline diff on its own, so has to be here), and to strip position information (but not origin node) from copied nodes when those positions are not in the current file (to fix a quickfix crash related to reusing nodes from outside the current file, which doesn't occur in the tests (but is definitely still a bug which could crop up) until this API change is applied).


src/compiler/checker.ts
```diff
@@ -11072,9 +11120,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         if (parentAccess && canHaveFlowNode(parentAccess) && parentAccess.flowNode) {
             const propName = getDestructuringPropertyName(node);
             if (propName) {
-                const literal = setTextRange(parseNodeFactory.createStringLiteral(propName), node);
+                const literal = setTextRangeWorker(parseNodeFactory.createStringLiteral(propName), node);
```
jakebailey, line 11123: Seems weird to rename this elsewhere; in other files we just import the symbol twice, once with another name like `ts_setTextRange`; maybe we do that here to not have to change anything outside of the node builder?

tests/cases/fourslash/jsDocFunctionSignatures2.ts
```diff
@@ -8,4 +8,4 @@
 ////
 //// f6('', /**/false)
 
-verify.signatureHelp({ marker: "", text: "f6(arg0: string, arg1?: boolean): number" });
+verify.signatureHelp({ marker: "", text: "f6(arg0: string, arg1?: boolean | undefined): number" });
```
jakebailey, line 11: Is this extra `| undefined` a behavior change for jsdoc types? I'm not super familiar with the semantics of `=` in this parsing to know if this is a good or bad change.

src/compiler/transformers/declarations.ts
```diff
@@ -638,7 +638,7 @@ export function transformDeclarations(context: TransformationContext) {
             case SyntaxKind.PropertyDeclaration:
             case SyntaxKind.BindingElement:
             case SyntaxKind.VariableDeclaration:
-                typeNode = resolver.createTypeOfDeclaration(node, enclosingDeclaration, declarationEmitNodeBuilderFlags, symbolTracker, shouldAddImplicitUndefined);
+                typeNode = resolver.createTypeOfDeclaration(node, enclosingDeclaration, declarationEmitNodeBuilderFlags, symbolTracker);
```
jakebailey, line 641: Do we still need `shouldAddImplicitUndefined` at this point? I did a diff locally forcing it to false above and it was... almost clean. So I think the answer is "yes" until we fix parameter properties and hopefully our goofy rules about optionalish/requiredish parameters...

src/compiler/checker.ts
```diff
@@ -6487,6 +6487,28 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             symbolToNode: (symbol: Symbol, meaning: SymbolFlags, enclosingDeclaration?: Node, flags?: NodeBuilderFlags, tracker?: SymbolTracker) => withContext(enclosingDeclaration, flags, tracker, context => symbolToNode(symbol, context, meaning)),
         };
 
+        /**
+         * Unlike the utilities `setTextRange`, this checks if the `location` we're trying to set on `range` is within the
+         * same file as the active context. If not, the range is not applied. This prevents us from copying ranges across files,
+         * which will confuse the node printer (as it assumes all node ranges are within the current file).
```
andrewbranch, line 6493: The printer’s single-file context has been a minor nuisance for code fixes and refactors in the past. For example, in an “add missing members from an incomplete `implements`” fix, we might write method signatures originating in two different files. How much more trouble would it be to fix the underlying assumption that every Node’s `originalNode` has the same source file?

jakebailey (commented):
Seems good. It's sort of hopeless to review all of these baselines, though. Are there any specific ones I should try to look at? (Would be cool to have a way to filter out the diff for any changes that aren't on reuse lines...)

src/compiler/checker.ts
weswigham: It's not even an issue with `.original`, it's an issue with assumptions about the text range - two of them. One is that, for literal strings and identifiers, unless some conditions are met, the printer just copies the input range to the output for them. AFAIK, this is to save on the escaping and unescaping work... I think. @rbuckton would have stronger opinions on that. The other is the token/node range bounds checking/assertions done in many services helpers. Specifically, we often assert that the last child of a node has the end position of its' parent - even inside the formatter. It's not clear to me that this is actually a good thing to assert on constructed node trees (though it is obviously true for parsed ones), but it is obviously not a good thing to assert on a tree with node ranges from a mixed set of input files, so we'd definitely need to revisit a lot of assertions in a bunch of services methods.

src/compiler/checker.ts
weswigham: Eh, there were only something like 3 uses outside of the node builder and, if I'm honest, they *probably* need the same sanity checking the node builder impl does, but I didn't want to put in the effort to prove it yet. (Why do the synthetic nodes manufactured for control flow even need a text range set? I honestly have no clue what it's used for. Errors?)

Anyways, the diff from the import rename isn't that large, since it's not used much outside the node builder - we'd just be trading 3 lines of diff for one line added.

tests/cases/fourslash/jsDocFunctionSignatures2.ts
weswigham: Yes. A jsdoc `type=` always has `| undefined` mixed into its' type (this is how it's typechecked, regardless of where in the parameter list it lies, though it is also *usually* considered an optional function parameter). Since we reuse the input node here, we transform the `type=` into `type | undefined` to strip the jsdoc syntax away.

weswigham: Of note: For parameters, `p?: boolean` and `p?: boolean | undefined` are functionally identical (since `exactOptionalProperties` only affects properties). Hence why the generated output based on the type omitted the `| undefined` before (it's redundant), while now that we're cloning/transforming the input node, we're being explicit about the `| undefined`. (Since a jsdoc `=` can appear outside of a parameter type position, unlike a parameter `?`, it really is just identical to adding a `| undefined` but with the extra baggage of also making a parameter that uses it optional)

src/compiler/transformers/declarations.ts
weswigham: > Do we still need shouldAddImplicitUndefined at this point?

Ultimately, I would like to remove it and the `EmitResolver` API backing it - for me, that was what motivated basically this whole stack of changes (quite literally: I removed it, and then started seeing what needed to be done to minimize the diffs). There's a bit more work to be done to minimize that diff, though - mostly, from what I can tell, around handling errors and duplicated declarations (with structurally-identical-but-identity-unique types) in the node builder more like the top-level declaration emitter does today.

weswigham: > So I think the answer is "yes" until we fix parameter properties and hopefully our goofy rules about optionalish/requiredish parameters...

Ah, you want to lock it to `true`, not `false`, that way it always falls back to the logic within `createTypeOfDeclaration` (which handles optional properties), rather than the declaration emitter's `visitDeclarationSubtree` (which doesn't) - that'll be more representative of what needs to be fixed in the node builder to remove the API.

weswigham: > I meant false in the code above (I couldn't highlight it in the diff, thanks GitHub), such that type && !shouldAddImplicitUndefined is true and we just copy the nodes (so, more reuse).

Yeah, exactly, at no point have I changed this case in this or any recent PR, however. The `visitDeclarationSubtree` codepath _always_ copies nodes, without regard for type information that may indicate otherwise. That's right enough most of the time, without considering the `shouldAddImplicitUndefined` carve-out. What we care about, IMO, is if the fallback `createTypeOfDeclaration` calls _also_ always copy the input nodes when they can - because if they do, the entire 
```ts
if (type && !shouldAddImplicitUndefined) {
    return visitNode(type, visitDeclarationSubtree, context);
}
```
branch can be removed, as `createTypeOfDeclaration` will internally have all the same logic (and it needs to anyway, in order to handle copying subtrees of input nodes within the nodes it generates). After this change, the only issues blocking that, AFAIK, are that `createTypeOfDeclaration` over-invalidates in the presence of an error type/structurally identical types across multiple declarations, and discards input nodes it could probably safely reuse.

weswigham:
> Are there any specific ones I should try to look at?

The small handful of non-`.types` baselines are the easiest ones. As for `.types`, largely this just makes us copy more parameter name origins (since we use the input declaration for the signature, rather than the signature's declaration, which isn't always set), but in some places we also pick up on a type node we missed/changed (since we find the input declaration's type node, rather than an arbitrary declaration's type node).

src/compiler/checker.ts
jakebailey: Alright, that's fine too.

tests/cases/fourslash/jsDocFunctionSignatures2.ts
jakebailey: man, jsdoc is weird

jakebailey:

Just checkin

jakebailey (approved):
Seems good to me so long as we're not totally throwing off @dragomirtitian :D

Definitely merge main first, as there's no way this isn't already out of date from recent test additions.

PR #58087, opened 2024-04-05
Move incremental state to WeakSets

jakebailey:
It turns out that the extra props we shove on nodes in incremental parsing are just "did we previously incrementally parse this, if so assert" and "did we make an edit that means we have to reparse", both of which are only ever set to true, so they don't even need to be WeakMaps, they're just write-only WeakSets.


jakebailey:

Doubt this will do much of anything though, given we don't really have great perf tests for incremental parsing.

DanielRosenwasser:
It's a bit of a shame that we don't have any long-running perf tests that make edits. The closest thing here is the crawler which does a few random edits.

src/compiler/parser.ts
```diff
@@ -9844,6 +9843,26 @@ namespace Parser {
     }
 }
 
+const incrementalInfo = {
+    incrementallyParsed: new WeakSet<SourceFile>(),
+    intersectsChange: new WeakSet<Node | NodeArray<Node>>(),
+};
```
jakebailey, line 9849: Probably these two should just be top-level variables, not one object.

sandersn (approved):

jakebailey:

PR #58088, opened 2024-04-05
[experiment] moving some Node state into WeakMaps

jakebailey:
Just playing around to see how this goes.

Given we don't touch `original`, `transformFlags`, or `emitNode` until emit, maybe these could be somewhere else instead. After all, editor users and most API users won't ever need these props at all.

Starting for now with `original` as that's not that hard to move.

### DIFF
--- a/src/compiler/emitter.ts
+++ b/src/compiler/emitter.ts
@@ -409,6 +409,7 @@ import {
     TypeQueryNode,
     TypeReferenceNode,
     UnionTypeNode,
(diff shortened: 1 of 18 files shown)


jakebailey:

jakebailey:
Good memory, bad emit perf. Blah. Maybe another prop.

jakebailey:

jakebailey:

Issue #58060, opened 2024-04-03
Mapped type reported as incompatible when passed through another mapped type (5.4 regression)


dragomirtitian:
### 🔎 Search Terms

mapped type assignability regression  

### 🕗 Version & Regression Information

- This changed between versions 5.3 and 5.4

### ⏯ Playground Link

[Playground Link](https://www.typescriptlang.org/play?ts=5.4.3#code/C4TwDgpgBAaghgGwK4QCrmgXigZ2AJwEsA7AcwG4AoS0SKAIQHsAPAHlSgmeAmIBMcsRCnSQAfFGwBvKADcAXFA4BfKjQxKIedp268BQ5GgwBtALoTsqarWgBVYgCMWAQXz44IdpahTKUAKgTAGkoEigAawgQRgAzJTNFVBCzXR5+QSY2EliIfCgAJQkAfkLAxWIIWTyqVRsNAEk+XmBCUG9JXyDQ8KiY+NREpRSoOupmgGMEOHxoWKRiCdbGYihY4no4Ph0udIM3Dy8s1ngjUQgxS4AKADo7mdIcJIBKJK1gViaWtq8HZ2YDp5vJc1JNprM1gsloQVmtiAB5CI7PQZKCAo4sE7CYzia53G4PJ5KV6abR-VzuIGoEHUdR0clsamdDi7fSCdGsY45PKFS5QUocopQCpVGqUMEzOZQ5ardYAZUIAFtCODkXt2ZSMWxTiIMJcxLd7vhHi83tovsRWu0GcCxFQgA)

### 💻 Code

```ts
type ValueType = string;

type Box<T extends ValueType> = { v: T };

type Test<T extends ValueType[]> = T

type UnboxArray<T> = {
    [K in keyof T]: T[K] extends Box<infer R> ? R   : never;
};

type Identity<T> = { [K in keyof T]: T[K] };


declare function fnBad<T extends Array<Box<ValueType>>>(...args: T): Test<Identity<UnboxArray<T>>>;

declare function fnOk<T extends Array<Box<ValueType>>>(...args: T): Test<UnboxArray<T>>;


type Unbox<T> = T extends Array<Box<infer R>> ? Array<R> : never;
declare function fnSimilar<T extends Array<Box<ValueType>>>(...args: T): Test<Identity<Unbox<T>>>;

```

### 🙁 Actual behavior

[...]

PR #58091, opened 2024-04-05
Improve apparent type of mapped types with mapped type modifiers

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/58060

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -14794,7 +14794,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         return type.resolvedApparentType || (type.resolvedApparentType = getResolvedApparentTypeOfMappedType(type));
     }
 
-    function getResolvedApparentTypeOfMappedType(type: MappedType) {
+    function getResolvedApparentTypeOfMappedType(type: MappedType): Type {
         const target = (type.target ?? type) as MappedType;
         const typeVariable = getHomomorphicTypeVariable(target);
         if (typeVariable && !target.declaration.nameType) {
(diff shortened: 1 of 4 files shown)


src/compiler/checker.ts
```diff
@@ -14794,15 +14794,16 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
-    function getResolvedApparentTypeOfMappedType(type: MappedType) {
+    function getResolvedApparentTypeOfMappedType(type: MappedType): Type {
         const target = (type.target ?? type) as MappedType;
         const typeVariable = getHomomorphicTypeVariable(target);
         if (typeVariable && !target.declaration.nameType) {
             const constraint = getConstraintTypeFromMappedType(type);
             if (constraint.flags & TypeFlags.Index) {
                 const baseConstraint = getBaseConstraintOfType((constraint as IndexType).type);
                 if (baseConstraint && everyType(baseConstraint, t => isArrayOrTupleType(t) || isArrayOrTupleOrIntersection(t))) {
-                    return instantiateType(target, prependTypeMapping(typeVariable, baseConstraint, type.mapper));
```
Andarist, line 14805: The `type.mapper` here looks something like:
```ts
m1: T_id -> Box<string>[]
m2: m1: K -> K
    m2: m1: T_id -> UnboxArray<T_outer>
        m2: T_id -> UnboxArray<T_outer>
```

Since `typeVariable` is `T_id` TS ended up "skipping" over the inner mapping of `UnboxArray`. The `constraint` (of the `type`) is `keyof T_outer` with a `baseConstraint` of `Box<string>[]`.

In a sense, I'd like to map `T_outer` to `baseConstraint` (`Box<string>[]`), process that through `type.mapper` and then map `T_id` to that result (or smth along those lines). I couldn't find a clear way to do that given that we need to account for non-homomorphic instantiations of homomorphic mapped types here. Using the modifiers type does seem to do the trick but I'm not overly confident in this solution and I won't mind being proven that this is wrong 😉

jakebailey:
@typescript-bot test it

ahejlsberg:
@Andarist Definitely the right idea to use `getModifiersTypeFromMappedType`, but the remaining logic can also be simplified more. I've put up #58098 with what I think is the correct fix.

Andarist:
Damn, I was close! :P It was worth a shot :)

PR #58093, opened 2024-04-05
Expand auto-import API to work on non-existent files and symbols

andrewbranch:
Fixes #

### DIFF
--- a/src/compiler/types.ts
+++ b/src/compiler/types.ts
@@ -4174,6 +4174,18 @@ export interface SourceFileLike {
 }
 
 /** @internal */
+export interface FutureSourceFile {
+    readonly path: Path;
+    readonly fileName: string;
+    readonly impliedNodeFormat?: ResolutionMode;
+    readonly packageJsonScope?: PackageJsonInfo;
+    readonly externalModuleIndicator?: true | undefined;
+    readonly commonJsModuleIndicator?: true | undefined;
+    readonly statements: readonly never[];
+    readonly imports: readonly never[];
+}
+
+/** @internal */
 export interface RedirectInfo {
     /** Source file this redirects to. */
     readonly redirectTarget: SourceFile;
--- a/src/services/codefixes/importFixes.ts
+++ b/src/services/codefixes/importFixes.ts
@@ -3,6 +3,7 @@ import {
     AnyImportOrRequireStatement,
     AnyImportSyntax,
     arrayFrom,
+    type BindingElement,
     CancellationToken,
     cast,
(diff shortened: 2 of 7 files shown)


src/services/codefixes/importFixes.ts
```diff
@@ -250,8 +275,69 @@ function createImportAdderWorker(sourceFile: SourceFile, program: Program, useAu
...
+        const useRequire = shouldUseRequire(sourceFile, program);
+        if (exportingSourceFile && exportingSourceFile.symbol) {
+            const { fixes } = getImportFixes(
+                [{
+                    exportKind,
+                    isFromPackageJson: false,
+                    moduleFileName: exportingFileName,
+                    moduleSymbol: exportingSourceFile.symbol,
+                    targetFlags: exportedMeanings,
+                    symbol: undefined!,
```
andrewbranch, line 293: It would be good if there was a less sketchy way to do this; I'll look a little closer

src/compiler/types.ts
```diff
@@ -4173,6 +4173,18 @@ export interface SourceFileLike {
     getPositionOfLineAndCharacter?(line: number, character: number, allowEdits?: true): number;
 }
 
+/** @internal */
+export interface FutureSourceFile {
+    readonly path: Path;
+    readonly fileName: string;
```
sheetalkamat, line 4179: Does this need to have resolvedPath and originalFileName as well ? are they used anywhere.. it would be needed if this is from other project that is referenced and i am not sure if thats something is needed as future file or not.

andrewbranch: I added only fields that were necessary to add for the current usage, with the expectation that other optional/empty fields be added in the future if needed.

sheetalkamat (approved):
Handle the API updates.. Hopefully no API changes just missing internal tag

src/services/codefixes/importFixes.ts
```diff
@@ -3,6 +3,7 @@ import {
     AnyImportOrRequireStatement,
     AnyImportSyntax,
     arrayFrom,
+    type BindingElement,
```
jakebailey, line 6: Are the type modifiers intentional, or just leftover from an editor setting?

Shouldn't matter too much now given dprint sorts these and I forced it in the settings.json template, but it is _technically_ inconsistent 😄

jakebailey (approved):

src/services/codefixes/importFixes.ts
andrewbranch: Left over, will fix

PR #58094, opened 2024-04-05
[experiment] Trying to remove / rewrite EmitResolver methods

jakebailey:
Just screwing around.

- `hasGlobalName` appears to _do nothing_; removing it (so assuming it's never true) doesn't change anything.
- `isOptionalParameter` (to me) feels like it's doing too much and could be syntax only, but this causes breaks I'll run tests on.


jakebailey:
@typescript-bot test it

jakebailey:
Hm, weird errors. Maybe the first thing _wasn't_ a noop.

jakebailey:
@typescript-bot test top400

fatcerberus:
This isn't the only PR where that "output file was not created from source file" error is showing up, I spotted it here too: https://github.com/microsoft/TypeScript/pull/58084
Very weird error, either way.

jakebailey:
@typescript-bot test top400

jakebailey:
From https://github.com/microsoft/TypeScript/pull/58201#discussion_r1567219997:

```ts
function isOptionalParameter(parameter: ParameterDeclaration) {  
        const signature = parameter.parent;  
        const paramIndex = signature.parameters.indexOf(parameter);  
        Debug.assert(paramIndex !== -1);  
        if (parameter.questionToken) return true;  
        if (parameter.dotDotDotToken) return !!parameter.initializer;  

        for (let i = paramIndex; i < signature.parameters.length; i++) {  
            const p = signature.parameters[i];  
            if (!p.questionToken && !p.initializer && !p.dotDotDotToken) {  
                return false;  
            }  
        }  
        return true;  
    }  
```

PR #58095, opened 2024-04-05
Use built-in no-restricted-syntax to ban null instead of plugin, ban null type too

jakebailey:
Not sure how valuable this is, but there's not really a need to use a plugin to ban `null`; ESLint already provides `no-restricted-syntax` which can be used to ban specific matching nodes.

The upside is that with #57684, we won't need any plugins anymore.

Also, `no-null` only banned the _value_ null, which is inconsistent from our guideline of not using null unless strictly required.


jakebailey:
This would be helpful as a precursor to #57684, as this package is untyped.

sheetalkamat (approved):

PR #58102, opened 2024-04-07
Always set node-version for setup-node

jakebailey:
Not setting a node version here means you get the global install, and the global install does not always behave, e.g. https://github.com/actions/runner-images/issues/9644#issuecomment-2041554588

Ensure these are all set to something. This should fix nightly and other tasks.


jakebailey:
Bypassing merge requirements so that we don't continue to not have nightly builds.

PR #58120, opened 2024-04-08
Send configFileDiag event when presence of errors change on project.update

sheetalkamat:
With this change we send configFileDiag when its a new program.
The need to add "skip" for this is because we have heuristics on when and how to send the configFileDiag when we open file so those operations just call with `skip` and do there own thing..

The main issue when looking into wsl2 repro was that when file is deleted: 
- file Delete event for that specific file is observed. and marked
- file delete and add in the config file's folder is observed but on "linux' or "wsl2" it is delayed so we can send that event in batches.
- Project gets updated as part of editor sending request to open file for renamed file with the new name - at this point the project only observes that old file name is deleted but because the wild card directory watcher is not invoked which updates root file names the root file names still contain old file name, So we sends configFileDiag that translates into vscode showing error: "missing old file"
- The timer for updating directory watcher runs and sends event for updating root file names of the config file 
- We update the project and have no errors but without this fix we could never report that to vscode since its only reported when tsconfig reloads or new file opens and doesnt belong to project. Result stale error in the editor. With this change, at this project update we will send configFileDiag to vscode and this should clear the errors for the project. displayed

Fixes https://github.com/microsoft/TypeScript/issues/57340

### DIFF
--- a/src/server/editorServices.ts
+++ b/src/server/editorServices.ts
@@ -840,10 +840,13 @@ export function projectContainsInfoDirectly(project: Project, info: ScriptInfo) 
         !project.isSourceOfProjectReferenceRedirect(info.path);
 }
 
-/** @internal */
+/**
+ * returns true if project updated with new program
+ * @internal
+ */
 export function updateProjectIfDirty(project: Project) {
     project.invalidateResolutionsOfFailedLookupLocations();
-    return project.dirty && project.updateGraph();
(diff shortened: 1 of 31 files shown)


jakebailey (approved):
Super super happy to have this fixed, has bugged me forever 😄

andrewbranch (approved):

PR #58124, opened 2024-04-08
Fix self tail call return type inference in assigned anonymous functions

### LINKED ISSUE #58100: Inference of recursive arrow function's return type is any
### 🔎 Search Terms

recursive function arrow function

- This is a crash

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.4.4#code/PTAEGMHsDsGcBdQDNKQFygBQEpQF4A+UAQ2gE8AoKORFSfLXQ0AbwtFAEsksBZY+AAsAdACdSAE0gBbHKCIAGYQFZcbDh1EBTeAFdR0ZKhztQAX1Pa9B0AooWKIZLujh4nGKABGxUTgzQutJeWqIUSC5uHoY+fmqm3HwCIuLQUrK4iirxGqBW+jG+JhwWmjoFtvYUQA

### 💻 Code

```ts
// const foo: () => any
const foo = () => {
  if (Math.random() > 0.5) {
    return foo()
  }
  return 0
}

// function bar(): number
function bar() {
  if (Math.random() > 0.5) {
    return bar()
  }
  return 0
}
```

### 🙁 Actual behavior

const foo: () => any
'foo' implicitly has return type 'any' because it does not have a return type annotation and is referenced directly or indirectly in one of its return expressions.

### 🙂 Expected behavior

const foo: () => number

_No response_

andrewbranch:
Good old `getMergedSymbol`.

~~I briefly thought through a case where the arrow function is assigned to a `let` and reassigned somewhere, but for that reassignment to change the type that can be returned by the function, it would have had to have a type annotation in the first place, so this code would never run. I _think_ this is safe.~~ User test showed one failure from this, so latest commit limits this feature to constant references (and function declarations as before).

Fixes #58100

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -37746,7 +37746,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 if (
                     expr.kind === SyntaxKind.CallExpression &&
                     (expr as CallExpression).expression.kind === SyntaxKind.Identifier &&
-                    checkExpressionCached((expr as CallExpression).expression).symbol === func.symbol
+                    checkExpressionCached((expr as CallExpression).expression).symbol === getMergedSymbol(func.symbol)
                 ) {
                     hasReturnOfTypeNever = true;
                     return;
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -28036,7 +28036,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             case SyntaxKind.Identifier:
                 if (!isThisInTypeQuery(node)) {
                     const symbol = getResolvedSymbol(node as Identifier);
-                    return isConstantVariable(symbol) || isParameterOrMutableLocalVariable(symbol) && !isSymbolAssigned(symbol);
+                    return isConstantVariable(symbol)
+                        || isParameterOrMutableLocalVariable(symbol) && !isSymbolAssigned(symbol)
+                        || !!symbol.valueDeclaration && isFunctionExpression(symbol.valueDeclaration);
                 }
                 break;
             case SyntaxKind.PropertyAccessExpression:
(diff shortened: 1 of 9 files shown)


andrewbranch:
@typescript-bot test it

andrewbranch:
The missing error in Webpack is a good thing, showing how this is intended to work. The new error in uglify-js is an example of a reassignment of the outer function throwing this off. It’s allowed because `noImplicitAny` is off, though I guess an explicitly-`any` variable could be assigned to a function variable under `noImplicitAny`. I think I’m going to limit this to `const` variables.

jakebailey:
Can you extract the uglify-js case into a test, since presumably it didn't fail in CI?

I find it weird that we would want to special case const vs let here, though...

andrewbranch:
The uglify-js case is added in the latest commit. We could do definite assignment analysis, but I’m not sure that’s worth it? There are other narrowing features that are predicated on const-ness, IIRC.

jakebailey:
Ah, it's the var case. I was looking for some sort of error baseline but that's the opposite of right. CI is also failing, but not sure if that's related or not.

jakebailey:
@typescript-bot test it

jakebailey:
@typescript-bot run dt

jakebailey (approved):

gabritto:
@typescript-bot pack this

gabritto:
Is it desirable that the following example changes behavior from `() => void` to `() => any` (and errors) with this PR? https://www.staging-typescript.org/play?ts=5.5.0-pr-58124-21#code/DYUwLgBAZgdhC8EAUBKBA+CBvAsAKAmjkVQ2wgCdwBXCuAZgG4IBffQgSymQFkBDMAAsAdBT4wAJgHsAtqjS4ChSjTpFUjdq3wsgA

andrewbranch:
@gabritto yes. We weren’t quite getting it right before and we’re still not quite getting it right now, but I think it’s arguably slightly better now? Previously, we _did_ identify the `return fn()` as a recursive tail call, because `getMergedSymbol` was never needed in this example. (To be honest I didn’t dig into when/why `getMergedSymbol` was needed in some of the other cases; I just noticed that the comparison was happening between a symbol and its merged version.) So previously, the return inside the `if` was eliminated to `never`, and falling out the bottom of the function contributed `void`. Now, we recognize that `fn` is reassigned, and so `return fn()` cannot be assumed to contribute `never`. So we try to get its return type and hit the circularity.

gabritto:
Ok, that makes sense.

gabritto (approved):

PR #58129, opened 2024-04-09
Fix 'diagnosticMessages‘ Chinese translations

### LINKED ISSUE #58130: Incorrect Chinese Translation for 'initializer' and Incorrect Use of Chinese Quotation Marks
### 🔎 Search Terms

initializer, i18n, Chinese

- This changed in commit (https://github.com/microsoft/TypeScript/commit/15c9cc8a324e6197179be298d58bab2805992aa3)

### ⏯ Playground Link

_No response_

### 💻 Code

```ts
// Your code here
```

### 🙁 Actual behavior

The current Chinese translation for 'initializer' is incorrect, as it has been translated as '初始化表杰式', which is not grammatically correct in Chinese. Additionally, the translation for accessing the property ['0'] incorrectly uses Chinese quotation marks [“0”], which may cause confusion for Chinese users. These issues should be addressed to improve the clarity and accuracy of the Chinese translations.

### 🙂 Expected behavior

I expected the Chinese translation for 'initializer' to be corrected to a grammatically correct form. Additionally, I expected the use of Chinese quotation marks for accessing the property ['0'] to be corrected to standard English quotation marks. These corrections would improve the accuracy and readability of the Chinese translations.

_No response_

cheezone:
This PR fixes Chinese translations for the 'diagnosticMessages' file. It corrects translation errors to improve clarity and accuracy.

1. initializer has been incorrectly translated as `初始化表杰式`, which is grammatically incorrect in Chinese.
2. The translation for accessing the property ['0'] incorrectly uses Chinese quotation marks [“0”], causing confusion for Chinese users.

Fix #58130

### DIFF
--- a/src/loc/lcl/chs/diagnosticMessages/diagnosticMessages.generated.json.lcl
+++ b/src/loc/lcl/chs/diagnosticMessages/diagnosticMessages.generated.json.lcl
@@ -10369,7 +10369,7 @@
         <Str Cat="Text">
           <Val><![CDATA[Property '{0}' cannot have an initializer because it is marked abstract.]]></Val>
           <Tgt Cat="Text" Stat="Loc" Orig="New">
-            <Val><![CDATA[属性“{0}”不能具有初始化表杰式，因为它标记为摘要。]]></Val>
+            <Val><![CDATA[属性“{0}”不能具有初始化表达式，因为它标记为抽象。]]></Val>
           </Tgt>
         </Str>
         <Disp Icon="Str" />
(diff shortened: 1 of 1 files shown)


jakebailey (changes requested):
At the moment, we do not have the means to verify that these changes are correct, both textually and that the files can still be loaded into the internal translation team's tooling.

Jack-Works:
> At the moment, we do not have the means to verify that these changes are correct, both textually and that the files can still be loaded into the internal translation team's tooling.

Yes. It's correct, and the current shipping one is wrong. I have to say, that TypeScript's message translation quality is very low (so I never use it), it does not fit the language's conventions and uses strange nouns/verbs. I hope we can have community translation.

sandersn:
I agree that community translations are a good idea for something we don't have the budget for. Preferably user-selected. Since we don't have a way to accept these PRs right now, I think the right thing to do here is to close this PR for now and open an issue about community translation. Does that sound OK to you @cheezone ?

PR #58138, opened 2024-04-09
Old Program disposal (memory leak)

### LINKED ISSUE #58137: Memory Leak disposing oldProgram
### 🔎 Search Terms

memory leak

- This is a crash

### ⏯ Playground Link

_No response_

### 💻 Code

```ts
// Your code here
```

### 🙁 Actual behavior

In safari (Webkit/JavaScriptCore). Old Program never gets GCed because of reference on it.

Memory build up really quickly.

### 🙂 Expected behavior

Unreferenced and gc

_No response_

cplepage:
Fixes #58137


src/compiler/program.ts
```diff
@@ -1858,6 +1858,7 @@ export function createProgram(rootNamesOrOptions: readonly string[] | CreateProg
     }
 
     // unconditionally set oldProgram to undefined to prevent it from being captured in closure
+    createProgramOptions.oldProgram = undefined;
```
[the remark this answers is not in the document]
jakebailey: (i.e. do the same as the lines after)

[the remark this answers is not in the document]
sheetalkamat: I am also wondering why this should cause the leak as unlike oldProgram and other things, this isnt used in any other closures and only at the start of the functions so technically this shouldnt cause issue? Am i missing something?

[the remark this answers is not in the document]
jakebailey: Well, the other variables below this are also not used within closures (as far as I can tell), so there's some funkiness here for sure which may imply that "not using" is not enough to not close over it, but to fix that should mean to do exactly what this code already does to fix that problem.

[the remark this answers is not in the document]
sheetalkamat: they are .. they are used in functions like tryReuseProgram etc..

src/services/services.ts
```diff
@@ -1801,6 +1801,9 @@ export function createLanguageService(
         };
         program = createProgram(options);
 
+        // unconditionally set oldProgram to undefined to prevent it from being captured in closure
+        options.oldProgram = undefined;
```
cplepage, line 1805: @jakebailey @sheetalkamat What about this then? I admit the previous way to unreference oldProgram was confusing.

jakebailey: This makes less sense... The thing I suggested (doing the same thing that the existing code did where you last edited) seems better.

cplepage: Ok I get it.

cplepage: Ok I played around with different approaches and could never figure out a clean way to fix this leak. I feel like there's indeed some funkiness in JSC that I just don't comprehend...

I will keep my fork for now since I cannot go forward without this ugly fix, but will try to comeback here if ever an update seems to fix anything...

I'm open to any other suggestion

jakebailey: The most helpful thing would be a repro that we can run that shows this leak.

DanielRosenwasser: JSC is technically a conservative collector, so that may tie into things. That said, it's also possible there's a leak for a precise GC without the mutation.

cplepage: I will build a little something to make the issue reproducible and figure out what could go on. But I will need some time. Please allow me 1-2 week and I will come back here with a clear workable playground.

cplepage: I've managed to setup a repro. Takes a while to get up and running, but hopefully you'll find a bit of time to check it out and maybe find an elegant fix 🙏

https://github.com/microsoft/TypeScript/issues/58137?notification_referrer_id=NT_kwDOAe993rQxMDE5MzE4ODQ5ODozMjQ3MjU0Mg#issuecomment-2076348479

sandersn:
@sheetalkamat or @jakebailey do you think the fix for #58138 is likely to be close to this PR? If so, let's keep it open. If not, let's close it and wait until one of us can get the repro to work for us. (Either way, one of the reviewers needs to repro locally to find out if this PR works.)

jakebailey:
#58138 is this PR, did you mean something else?

jakebailey:
If we have a repro, I can imagine there being a fix, but this change itself still looks incorrect. I believe we are waiting for some kind of repro.

tmm1:
There's a repro available here now: https://github.com/microsoft/TypeScript/issues/58137#issuecomment-2604840198

jakebailey:
I posted #61034 which is this PR (I added a co-author for you), plus some changes to make me feel better about the idea of accepting this kind of hack.

jakebailey:
Actually, I found a better cause; #61034 now contains a fix for that, which is not too far off from this PR. Going to close this one.

PR #58154, opened 2024-04-11
Experiment: Infer type predicates for functions with multiple returns

danvk:
Follow-on to #57465

_⚠️ **Disclaimer**: I'm not sure this is a good idea, I'm posting this mostly to get a sense for perf impact and breakages._

The original PR restricted itself to functions with a single `return`. This PR lifts that restriction.

This allows us to infer predicates for constructs like this:

```ts
// infers x is string | Date
function ifElseIfPredicate(x: string | number | Date) {
  if (typeof x === 'string') {
    return true;
  } else if (x instanceof Date) {
    return true;
  }
  return false;
}
```

or

```ts
// infers str is "a" | "b" | "c"
function switchPredicate(str: string) {
  switch (str) {
    case 'a':
    case 'b':
    case 'c':
      return true
    default:
      return false
  }
}
```

## How this works

Type predicate inference works by rewriting `return` statements from:

```ts
function foo(p: InitType) {
  return expr_involving_p;
}
```

to:

```ts
function foo(p: InitType) {
  if (expr_involving_p) {
    p  // TrueType
  } else {
    p  // FalseType
  }
}
```

If `TrueType != InitType` then we have a candidate for a type predicate. To generalize this to multiple returns, we determine the TrueType at each return statement and union them. These are all the parameter types for which the function can return true.

To verify the "if and only if" semantics of type predicates, we feed the type predicate type back into the function and look at all the false cases. For a valid type predicate, these should all be `never`.

I've also special-cased `return true` and `return false`. These weren't especially relevant when `getTypePredicateFromBody` only allowed one `return`, but they're very relevant now.

## Why this might not be a good idea

There were only two functions in the existing baselines that became type predicates with this change. So this might not be a very common pattern.

Moreover, I got a new circularity error in `emitter.ts`. This would _also_ have been a new circularity error with #57465, but the type predicate inference code didn't run because one of the functions had two returns. This change increases the odds that this happens on existing code.

More generally, I imagine the perf impact of this will be non-trivial since it might run on many more functions.

But those are all just hunches. I'd be curious to see the numbers and breakages on test repos!

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -37856,53 +37856,82 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         const functionFlags = getFunctionFlags(func);
         if (functionFlags !== FunctionFlags.Normal) return undefined;
 
-        // Only attempt to infer a type predicate if there's exactly one return.
-        let singleReturn: Expression | undefined;
+        // Collect the returns; bail early if there's a non-boolean.
+        let returns: Expression[] = [];
         if (func.body && func.body.kind !== SyntaxKind.Block) {
-            singleReturn = func.body; // arrow function
+            returns = [func.body]; // arrow function
         }
         else {
             const bailedEarly = forEachReturnStatement(func.body as Block, returnStatement => {
-                if (singleReturn || !returnStatement.expression) return true;
-                singleReturn = returnStatement.expression;
+                if (!returnStatement.expression) return true;
+                const expr = skipParentheses(returnStatement.expression, /*excludeJSDocTypeAssertions*/ true);
+                const returnType = checkExpressionCached(expr);
+                if (!(returnType.flags & (TypeFlags.Boolean | TypeFlags.BooleanLiteral))) return true;
+                returns.push(expr);
             });
-            if (bailedEarly || !singleReturn || functionHasImplicitReturn(func)) return undefined;
+            if (bailedEarly || !returns.length || functionHasImplicitReturn(func)) return undefined;
         }
-        return checkIfExpressionRefinesAnyParameter(func, singleReturn);
-    }
 
-    function checkIfExpressionRefinesAnyParameter(func: FunctionLikeDeclaration, expr: Expression): TypePredicate | undefined {
-        expr = skipParentheses(expr, /*excludeJSDocTypeAssertions*/ true);
-        const returnType = checkExpressionCached(expr);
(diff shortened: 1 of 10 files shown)


jakebailey:
@typescript-bot test it

jakebailey:
@typescript-bot test top800

danvk:
I put up #58173, which reports an error whenever it infers a type predicate for a function with multiple `return` statements. That should clarify how common this pattern is in the wild.

danvk:
Results of the experiment:

- three new type predicates in TypeScript itself, one of which had an impact
- one new type predicate in pyright
- three new type predicates in the top400 suite

So not many functions match this pattern in the wild, but a few do and there's also not a big perf impact. Up to TS team whether this is a helpful generalization. See @jakebailey's comment https://github.com/microsoft/TypeScript/pull/58173#issuecomment-2052696449

anka-213:
I would love to have this, since it allows splitting up type predicates instead of having to write them as a single long expression. I'm surprised that things like

```typescript
function isArrayOfStrings(x: unknown): x is string[] {
  if (!x instanceof Array) return false;
  
  return x.every((y) => typeof y === 'string');
}
```

didn't occur more in your experiment, since I've seen it pretty often for more complicated type predicates.

danvk:
@anka-213 if there are examples in the wild that you can point to, that would be helpful. This makes me realize that my experiment (#58173) would not tell us about _existing_ type predicates with 2+ returns, since the predicate inference code doesn't run if there's an explicit return type annotation. In other words, it would not have detected your `isArrayOfStrings` function. Finding functions like this would be another interesting data point.

danvk:
I've merged upstream changes and brought the implementation more in line with `main` to reduce the diff. I've also updated the PR description in light of the experiments and performance numbers from April. @jakebailey it would be interesting to get new performance numbers to see if the "minimal impact" holds up.

I can open an issue like typescript-bot suggests if that would be a more appropriate way to discuss the merits of this extension.

jakebailey:
@typescript-bot test it

danvk:
Other than the compiler-unions emit time, this looks like ~0 impact. I can take a look at whether there's some giant type predicate being emitted that would account for that, but I'm skeptical since we [didn't see change](https://github.com/microsoft/TypeScript/pull/58154#issuecomment-2050652764) in this number back in April and the behavior of the PR shouldn't be different.

PR #58165, opened 2024-04-12
Add missing recursive vistor on copied `import` type nodes

### LINKED ISSUE #58159: [nightly][regression] Some emitted imports are syntactically invalid
### 🔎 Search Terms

import, declaration, empty

- Bisects to #58085 by @weswigham 

### ⏯ Playground Link

_No response_

### 💻 Code

```ts
n/a
```

### 🙁 Actual behavior

A type referenced in a declaration file is sometimes emitted as `import().SomeType` and sometimes as `import(<actual_import_path>).SomeType`

### 🙂 Expected behavior

The type is always emitted as `import(<actual_import_path>).SomeType`

It's not easy for me to provide a self-contained reproduction, as it involves the composition of a lot of our private code, but I figure since it's already known which PR introduced the regression, it should hopefully be easier for someone on the TS team to spot how an import could possibly be empty. I can patch speculative fixes or add some debug logs if requested.

weswigham:
And swap to using the cached (and much more robust) `getSpecifierForModuleSymbol` for generating new specifiers while we're at it (so we can more often know to reuse the input node when the specifier matches).

Fixes #58159


tests/baselines/reference/privateNamesUnique-2.types
```diff
@@ -65,12 +65,12 @@ const b = new B();
 a.copy(b); // error
 >a.copy(b) : void
 >          : ^^^^
->a.copy : (other: import(} fro).Foo) => void
->       : ^     ^^                 ^^^^^^^^^
+>a.copy : (other: import("b").Foo) => void
+>       : ^     ^^                     ^^^^^^^^^
 >a : A
 >  : ^
->copy : (other: import(} fro).Foo) => void
```
weswigham, line 72: That we already had a test demonstrating the issue is... yeah, reviewing big baseline diffs is hard - easy to miss the *one* test that's still off. And that's why we try and break it down where we can. 🤷‍♂️

tests/baselines/reference/jsDeclarationsClassImplementsGenericsSerialization.js
```diff
@@ -68,4 +68,4 @@ export class Encoder<T> implements IEncoder<T> {
      */
     encode(value: T): Uint8Array;
 }
-export type IEncoder<T> = import('./interface').Encoder<T>;
+export type IEncoder<T> = import("./interface").Encoder<T>;
```
weswigham, line 71: Since we're doing the recursive visit on the specifier now, the quote-style-conformity change we just made is now also being applied here~

tests/baselines/reference/declarationEmitCrossFileCopiedGeneratedImportType.js
```diff
@@ -0,0 +1,32 @@
...
+Object.defineProperty(exports, "__esModule", { value: true });
+exports.d = void 0;
+var projC_1 = require("../projC");
+exports.d = { e: projC_1.e };
+
+
+//// [index.d.ts]
+export declare const d: {
+    e: {
+        f: (foo: import("../projA").Foo) => boolean;
```
weswigham, line 30: If I didn't also swap to `getSpecifierForModuleSymbol` as part of this, this emits as `"../projA/index"` because, well, `getResolvedExternalModuleName` is not the _robust_ specifier generation solution in the checker that calls into `moduleSpecifiers.ts`, while `getSpecifierForModuleSymbol` is. `getResolvedExternalModuleName` was probably sufficient for just handling the unbundled -> bundled specifier case (which is all the `declarations.ts` version of `rewriteModuleSpecifier` needs to handle and what this was originally based on), but doesn't handle rewriting specifiers copied from other files, which is an additional case the node builder needs to handle (and as of this week, happens waaaaaaay more often). `getExternalModuleNameFromDeclaration` does.

src/compiler/checker.ts
```diff
@@ -8479,22 +8477,16 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+                        const targetFile = getExternalModuleFileFromDeclaration(parent);
+                        if (targetFile) {
+                            const newName = getSpecifierForModuleSymbol(targetFile.symbol, context);
+                            if (newName !== lit.text) {
+                                return setOriginalNode(factory.createStringLiteral(newName), lit);
                             }
                         }
                     }
-                    return lit;
+                    return visitNode(lit, visitExistingNodeTreeSymbols) as StringLiteral;
```
jakebailey, line 8489:
```suggestion
                    return visitNode(lit, visitExistingNodeTreeSymbols, isStringLiteral)!;
```

tests/baselines/reference/privateNamesUnique-2.types
jakebailey: Oops!

jakebailey (approved):

MichaelMitchell-at:
I can confirm that this fixes the issue in our codebase

Issue #58158, opened 2024-04-12
[nightly][regression] Wrong generic parameter inferred for constructed class instance


MichaelMitchell-at:
### 🔎 Search Terms

generic, class, infer

### 🕗 Version & Regression Information

- Bisects to https://github.com/Microsoft/TypeScript/pull/57403 by @weswigham 

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.5.0-dev.20240411#code/KYDwDg9gTgLgBDAnmYcCCBnDBLA5gOwB4AVAGjgFUA+OAXjgHkBbbGE8ga2EQgDNKaAMkoBuAFBiAxgBsAhljgAxCBBI0A3mLja4kiPgwwoAV0kxoACjDGARtOySEALjjEAlHHUBfCTvRY8fAtjF3UQFwBGLzcXZVVMHAJ2T3C4KKoNLT8dPQN4ADcXBMDksMivGnoGGwArYDMAOnlEoO9yGAALbAwGmHJjN3FsnShgGGMofDh8YAB3JRULfMGs4YB6NbgAPQB+VeyN12RUAHI4wmKky6IyTzhUiJE4CvJ1e8iniqoTuG7piHgzUCsjsqHMCGOcDOKguASStzeD0+GRODX2fkOs1kMH2Ph8QA

### 💻 Code

```ts
export type Assign<T, U> = Omit<T, keyof U> & U;

class Foo<T> {
    constructor(public t: T) {}

    assign(u: {x: 1}): Foo<Assign<T, {x: 1}>> {
        const v: Assign<T, {x: 1}> = Object.assign({}, this.t, u);
        return new Foo(v);
        // ^?
        // Type 'Foo<Assign<Assign<T, { x: 1; }>, { x: 1; }>>' is not assignable to type 'Foo<Assign<T, { x: 1; }>>'.
        // wat
    }
}
```

### 🙁 Actual behavior

Type error

### 🙂 Expected behavior

No type error

weswigham:
Duplicate of #58074 :)

PR #58168, opened 2024-04-12
Use better context scope for class constructor implementation signatures

### LINKED ISSUE #58074: Class infers wrong generic parameter in latest (regression?)
### 🔎 Search Terms

"infer"

- This changed in commit e1874f3ffed435b6367adaba3b81608e931dcf6c

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.5.0-dev.20240404#code/KYDwDg9gTgLgBAYwDYEMDOa4BVhpgHgEEAaOAIQD44BvAKDgcQgDs8oBXBGaOACjHYAjJAEsEcAG4okALjgk4A4WLgBbFGDAjmAczm8JcwgEo4AXiplT1AL616jGLgIBhCr0HaAJtr19D5KYWcC7Gcjh4+GSkbjQOjIwILHhwzMAA7gBq0uZwMAAWImgAdOqavrwFRcVSSMYA3PEJcFDAMOxQzKkZ2M68aVnSpJ7MProNTXZ2QA

### 💻 Code

```ts
export class Test<A, B> {
    constructor (public val: A, public mapping: (v: A) => B) {}

    test<C>(binding: (v: B) => C): Test<B, C> {
        const newVal = this.mapping(this.val);
        return new Test(newVal, binding);
    }
}
```

### 🙁 Actual behavior

The invocation of `new Test` infers generic parameters of `<C, C>`, even though `newVal` is `B` and `binding` is `(v: B) => C`, which should pretty clearly result in `<B, C>`. This incorrect inference causes an error on line 6:
```
Type 'Test<C, C>' is not assignable to type 'Test<B, C>'.
  The types returned by 'toVal()' are incompatible between these types.
    Type 'C' is not assignable to type 'B'.
      'B' could be instantiated with an arbitrary type which could be unrelated to 'C'.ts(2322)
```

Explicitly invoking `new Test<B, C>` compiles without error.

### 🙂 Expected behavior

Compiles successfully.

_No response_

weswigham:
Since those type parameters are reachable anywhere in the body of the class, not just the constructor itself.

Honestly, we could *always* do the `getImplementationSignature` call, it doesn't *have* to be conditional, but we like trying to cut down on the type parameter copying we do, where possible. And certainly, as long as you only ever use the signature externally to the scope within which its' type parameters are internally reachable, you don't *need* to copy it.

Fixes #58074

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -34806,7 +34806,8 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 if (candidate.typeParameters) {
                     // If we are *inside the body of candidate*, we need to create a clone of `candidate` with differing type parameter identities,
                     // so our inference results for this call doesn't pollute expression types referencing the outer type parameter!
-                    if (candidate.declaration && findAncestor(node, a => a === candidate.declaration)) {
+                    const candidateParameterContext = candidate.declaration && isConstructorDeclaration(candidate.declaration) ? candidate.declaration.parent : candidate.declaration;
+                    if (candidateParameterContext && findAncestor(node, a => a === candidateParameterContext)) {
                         candidate = getImplementationSignature(candidate);
                     }
                     let typeArgumentTypes: readonly Type[] | undefined;
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -34806,7 +34806,8 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 if (candidate.typeParameters) {
                     // If we are *inside the body of candidate*, we need to create a clone of `candidate` with differing type parameter identities,
                     // so our inference results for this call doesn't pollute expression types referencing the outer type parameter!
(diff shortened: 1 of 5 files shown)


RyanCavanaugh (approved):

jakebailey (approved):
@typescript-bot test it

MichaelMitchell-at:
Hm, the repro I provided in https://github.com/microsoft/TypeScript/issues/58158 is fixed now, but I'm still seeing the issue in our codebase. I'll have to update the repro to more closely resemble our actual code.

MichaelMitchell-at:
@weswigham here's another regressed case that doesn't yet pass
```ts
type Assign<T, U> = Omit<T, keyof U> & U;

abstract class Base<T >  {
    public constructor(public _schema: T) {}
}

export class Foo<T > extends Base<T> {
    public sensitivity(
        x: 1,
    ): Foo<Assign<T, {x: 1}>> {
        const v: Assign<T, {x: 1}> = Object.assign(this._schema, {x});
        return new Foo(v);
    }
}
```

weswigham:
@MichaelMitchell-at handled~

@typescript-bot test it

MichaelMitchell-at:
Nice, that did the trick, thank you

jakebailey (approved):

PR #58169, opened 2024-04-12
Don’t emit __esModule on CJS files that didn’t use ESM syntax

### LINKED ISSUE #58127: Problems with __esModule marker in JS/TS mixed node16 environemnt
### 🔎 Search Terms

__esModule, node16, module resolution

This caused a crash when I assumed synthetic default import would work on a cjs js file, but the resulting output from the cjs js file got the __esModule marker stopping the synthetic default import mechanism. The bug is that the file is certainly not a transpiled es module.

I could not make a convincing repro in playground, so I made a full repo instead:
https://github.com/meros/tsc-module-bug-repro

### ⏯ Playground Link

https://github.com/meros/tsc-module-bug-repro

### 💻 Code

In a node16 project, using a mix of js/ts files. CommonJS JS source files get the __esModule marker causing synthetic default imports to fail:

```ts
import syntheticDefaultImport from "./cjs-to-cjs-js.cjs";
import * as starImport from "./cjs-to-cjs-js.cjs";

// Will be undefined
console.log("syntheticDefaultImport", syntheticDefaultImport);

// Will be OK
console.log("starImport", starImport);
```

### 🙁 Actual behavior

synthetic default import fails due to misleading __esModule marker in output file

### 🙂 Expected behavior

a JS source that should be treated as, or is detected to be, a commonjs that is transpiled to a commonjs module should NOT get the __esModule marker.

This bug makes the transition to modules in node projects harder and more error prone. I spent at least a day trying to figure out if I misconfigured something or misunderstood the module system but I cannot see how this is not a bug in tsc related to node16 handling.

andrewbranch:
Fixes #58127

### DIFF
--- a/src/compiler/transformers/module/module.ts
+++ b/src/compiler/transformers/module/module.ts
@@ -56,6 +56,7 @@ import {
     getOriginalNodeId,
     getStrictOptionValue,
     getTextOfIdentifierOrLiteral,
+    hasJSFileExtension,
     hasJsonModuleEmitEnabled,
     hasSyntacticModifier,
     Identifier,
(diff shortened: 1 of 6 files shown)


weswigham (approved):

meros:
Thanks man! Didn't expect this to be picked up and fixed so quickly 🙏

PR #58173, opened 2024-04-12
Experiment: report functions with multiple returns that could be type predicates

danvk:
This is a side experiment for #58154. That PR extends type predicate inference to functions with multiple `return`s. It seems to have minimal perf impact and also minimal breakages. This makes you wonder… _are_ there any such functions?

This fork of the PR should answer that question by reporting errors on all functions with 2+ `return` statements where it can infer a type predicate. To be clear, these aren't real errors! They're just identifying multi-return predicates in the wild.

There are three such function in TypeScript itself:

1. `tryAddPropertyAssignment` in `nodeFactory.ts`:

```ts
function tryAddPropertyAssignment(properties: PropertyAssignment[], propertyName: string, expression: Expression | undefined) {
    if (expression) {
        properties.push(createPropertyAssignment(propertyName, expression));
        return true;
    }
    return false;
}
```

This gets an inferred return type of `expression is Expression`. This is accurate, though not consequential for any of the code that calls this function.

2. `charIsPunctuation` in `patternMatcher.ts`:

```ts
function charIsPunctuation(ch: number) {
    switch (ch) {
        case CharacterCodes.exclamation:
        case CharacterCodes.doubleQuote:
        // ...
        case CharacterCodes._:
        case CharacterCodes.openBrace:
        case CharacterCodes.closeBrace:
            return true;
    }

    return false;
}
```

This gets an inferred return type of `ch is CharacterCodes._ | CharacterCodes.ampersand | CharacterCodes.asterisk | ... | CharacterCodes.slash`. Again, this is accurate but not consequential for any of the calling code.

3. An anonymous function in `getLinkedEditingRangeAtPosition` in `services.ts`:

```ts
const tag = findAncestor(token.parent, n => {
    if (isJsxOpeningElement(n) || isJsxClosingElement(n)) {
        return true;
    }
    return false;
});
if (!tag) return undefined;
Debug.assert(isJsxOpeningElement(tag) || isJsxClosingElement(tag), "tag should be opening or closing element");
```

The arrow function becomes a type predicate with #58154 which changes the type of `tag`. This is a positive change that eliminates the need for the `Debug.assert` statement.

Of course, the function could have been written more succinctly as `n => isJsxOpeningElement(n) || isJsxClosingElement(n)` and the existing code would infer a type predicate.

I'm curious what other errors this reports on the user/top/etc suites. This will characterize whether this is a common pattern in the wild.

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -37874,7 +37874,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
 
         // We know it's all boolean returns. For each parameter, get the union of the true types.
         // Then feed that union back through to make sure that a false return value corresponds to never.
-        return forEach(func.parameters, (param, i) => {
+        const predicate = forEach(func.parameters, (param, i) => {
             const initType = getTypeOfSymbol(param.symbol);
             if (!initType || initType.flags & TypeFlags.Boolean || !isIdentifier(param.name) || isSymbolAssigned(param.symbol) || isRestParameter(param)) {
                 // Refining "x: boolean" to "x is true" or "x is false" isn't useful.
@@ -37932,6 +37932,10 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 return createTypePredicate(TypePredicateKind.Identifier, unescapeLeadingUnderscores(param.name.escapedText), i, paramTrueType);
             }
         });
+        if (predicate && returns.length > 1) {
+            error(func.name ?? func, Diagnostics.Function_with_multiple_returns_is_implicitly_a_type_predicate);
+        }
+        return predicate;
     }
 
     /**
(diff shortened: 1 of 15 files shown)


RyanCavanaugh:
@typescript-bot test it

jakebailey:
The pyright one is: https://github.com/microsoft/pyright/blob/92f2284d2459fed8a9fdaab023527d8598749fcc/packages/pyright-internal/src/analyzer/parseTreeUtils.ts#L2588

jakebailey:
I feel like it's telling that we're seeing so few of these, and that when we do, the code still compiles as it did before. That makes it seem like this isn't very valuable, and that the initial design was good enough.

danvk:
Closing this out since we've learned what we're going to from it.

jakebailey:
BTW thank you for sending these PRs in exactly the format we need to let the infra report something, super helpful.

Issue #42873, opened 2021-02-19
The inferred type of "X" cannot be named without a reference to "Y". This is likely not portable. A type annotation is necessary.


mistic:
# Bug Report

### 🔎 Search Terms

inferred type cannot be named, symlink node_modules

### 🕗 Version & Regression Information

I'm verifying the problem on the `typescript@4.1.3`. I've not tried older versions but at least is also reproducible on the `@next` version as of today.

It is probably a regression or a corner case related with other issues opened and already closed like:

- https://github.com/microsoft/TypeScript/issues/30858
- https://github.com/microsoft/TypeScript/issues/28689
- https://github.com/microsoft/TypeScript/issues/2338
- https://github.com/microsoft/TypeScript/issues/29221

### ⏯ Playground Link

[Link for a repo where the problem is being reproduced](https://github.com/mistic/reproduce-typescript-problem-when-symlinking-node_modules)

> NOTE: Just clone the repo and run `yarn tsc`

### 💻 Code

All the relevant code can be found at https://github.com/mistic/reproduce-typescript-problem-when-symlinking-node_modules

It is just reproducing a similar setup that I had on other project that was generating the problem:

- node_modules are a symlink to another location that is not a direct parent of the symlinked node_modules
- we are using types in the compilation from a library where those types are just exported from other one, like for example `withRouter` within `react-router-dom` that is just a plain export from the same type on `react-router`.

### 🙁 Actual behavior

I got the following error:

```
error TS2742: The inferred type of 'Nav' cannot be named without a reference to '../../deps/node_modules/@types/react-router'. This is likely not portable. A type annotation is necessary.

8 export const Nav = withRouter(({ history }: NavProps) => {
               ~~~


Found 1 error.
```
### 🙂 Expected behavior

I was expecting no error at all and that the typescript compiler was just able to find all the respective modules. I've tried everything that I thought was related like enabled the `preserveSymlinks`. The only thing that stops the error is importing the `withRouter` type directly from `react-router` and not from `react-router-dom` but that doesn't make very sense because I actually want to use the `react-router-dom` on a couple of places.

\cc @weswigham @sheetalkamat @andrewbranch because I saw you previously worked to try to solve similar issues.

weswigham:
[...]

PR #58176, opened 2024-04-12
Check nearest package.json dependencies for possible package names for specifier candidates

weswigham:
Implements [this comment](https://github.com/microsoft/TypeScript/issues/42873#issuecomment-781747185).

I won't say this "fixes" #42873, but it does prevent it from occurring for people in workspaces with appropriately set-up explicit dependencies in their package files, even when they haven't explicitly referenced all of those dependencies in their program. IMO, that's the upper limit of how much we can or should fix that issue - anything further would imply unsafe/unportable specifiers, which I don't want us to generate. I say that, but this could be improved upon: While this leverages the fact that we already lookup all `package.json` files to make this cheap, this does not look through those `package.json` `dependencies` to try and make a deep import, so we'll only discover top-level package imports this way right now. That's a potential improvement that could be built on this.


src/compiler/moduleSpecifiers.ts
```diff
@@ -1026,6 +1028,46 @@ function tryGetModuleNameFromPackageJsonImports(moduleFileName: string, sourceDi
...
+
+    const candidates = getOwnKeys(dependencies);
+    for (const name of candidates) {
+        const resolved = resolveModuleName(name, combinePaths(sourceDirectory, "dummy.ts"), options, host as (typeof host & { readFile: NonNullable<(typeof host)["readFile"]> }), cache, /*redirectedReference*/ undefined, importMode);
+        if (resolved && resolved.resolvedModule && resolved.resolvedModule) {
+            const resolvedPath = toPath(resolved.resolvedModule.resolvedFileName, sourceDirectory, getCanonicalFileName);
+            if (resolvedPath === moduleFileName) {
+                return name;
+            }
+            if (host.realpath && host.realpath(resolvedPath) === host.realpath(moduleFileName)) {
```
weswigham, line 1063: I didn't *want* to have to `realpath` here, but it seems to be the only way we can check if the two paths actually refer to the same thing through symlinks.

RyanCavanaugh:
I've been reading the comments on #42873 and I'd estimate anywhere from 30% to 80% are legitimate misconfigurations, or at least configurations that we can't plausibly disambiguate from incorrect ones. It's hard to tell.

This looks like a good fix but I'll also write up a FAQ entry on this which covers the basics. What's the most-plausible way someone would legitimately hit this error in a way that is also "fixable" without some restructuring of their package folder layout?

weswigham:
> This looks like a good fix but I'll also write up a FAQ entry on this which covers the basics. What's the most-plausible way someone would legitimately hit this error in a way that is also "fixable" without some restructuring of their package folder layout?

The test case in this PR is a good example - a workspace (`yarn`, `npm`, `pnpm`, take your pick) with explicit deps between workspace members with everything (sym)linked together from the workspace (some package managers may also link into the global cache first (or make similar structures with symlinks within the cache), but extra layers of link indirection shouldn't affect this), the only thing missing is an explicit in-code import from one package to one of its' dependents to make us recognize the valid specifier up-front.

RyanCavanaugh:
draft faq
----
Let's say you use a package manager with strict dependencies:
```
|-/my_package_1
|-- /node_modules
|---- /other_package <- direct dependency
|------ index.d.ts
|------ /node_modules
|--------- /sub_dep <- symlink!
|----------- index.d.ts
|-- /src
|---- tsconfig.json  <- project root
|---- foo.ts         <- a source file that imports the above file
```
Where `foo.ts` looks like this:
```ts
import { getMakeSubDep } from "other_package";

// The inferred type of p refers to a type defined
// inside node_modules/other_package/node_modules/sub_dep/index.d.ts
export const p = getMakeSubDep();
```
When TypeScript needs to emit `foo.d.ts`, it needs to write out a type for `p`:
```ts
export const p: ???
```

What should go in the `???` ?

One option would be to use a relative path:
```ts
import { subdep } from "../node_modules/other_package/node_modules/sub_dep";
export const p: subdep
```
This is *obviously* wrong: It's implausible that a consumer of `foo.d.ts` would have a folder layout that matches what we happened to have here.

Another option would be to use a subpath:
```ts
import { subdep } from "other_package/node_modules/sub_dep";
export const p: subdep
```
This is *also* obviously wrong: The semantics of `other_package` are not that it exposes `node_modules/sub_dep` as as sub-path. This probably won't work at runtime, and even if it did, it's not what you want

Another option would be to use a module name:
```ts
import { subdep } from "sub_dep";
export const p: subdep
```

Is *this* correct?

If `other_package` has a dependency on `sub_dep@2.0.0` and *your* package has a dependency on `sub_dep@3.0.0`, then these aren't the same type, and it's wrong.

If you don't have a declared dependency on `sub_dep` at all, then this code would also fail to load `sub_dep` when ingested in a downstream project.

These situations - the non-working ones - are the "non-portable identifiers" that TypeScript is complaining about. TS was put into a position where it had to reference the name of a module, but it couldn't find a name for that module that appeared to work. This is why this error occurs.

Now, if you *do* have a declared dependency on `sub_dep`, and it resolves to the "same" target as the one `other_package` did, then this is fine. The way this works (roughly) is that TS keeps a reverse map of the files it has loaded and what module specifiers it used to find them. So if you have `import { subdep } from "sub_dep"`, and it resolved to a correct .d.ts file, then TS will have it in the lookup table and can use `sub_dep` to refer to the same file that `other_package` did even though these two module specifiers *may have* referred to different files (but they didn't).

But! If you never referred to `sub_dep` in your compilation, then TypeScript has never "seen" `sub_dep` be referred to from the module resolution context that `foo.d.ts` will have, so it doesn't know whether or not `"sub_dep"` is a legal way to refer to `other_package/node_modules/sub_dep`, and ends up issuing this error.

The correct way to address this is generally to import the type that the `.d.ts` needs to refer to:
```ts
// in foo.ts
import { subdep } from "sub_dep";
import { getMakeSubDep } from "other_package";

// No additional type annotation needed
export const p = getMakeSubDep();
```
If *you* can't do this, then TypeScript can't either, and the only real alternative is to use a broader type annotation so that TS doesn't need to refer to types which are impossible to name:
```ts
// in foo.ts
import { getMakeSubDep } from "other_package";

// Annotate to anonymous version of whatever subdep is
export const p: { isSubDep: boolean } = getMakeSubDep();
```
Or restructure the input file in such a way that the referenced type is not exposed in the `.d.ts`.

weswigham:
Yep, excellent summary of the current state of affairs. That's basically all the productive bits of the years of discussion in https://github.com/microsoft/TypeScript/issues/42873 summed up.

robpalme:
For the FAQ, another option to offer up is to extract the type from the original import to avoid adding additional (fragile?) routes to the type.

```ts
export const p: ReturnType<getMakeSubDep> = getMakeSubDep();
```

(Even better would be if this could be auto-inferred during DTS emit to avoid the user needing to explicitly perform the extraction in the TS file.)

eps1lon:
Is there an option to enforce that whatever `getMakeSubDep` returns, is also exported from `other_package`? Then TypeScript could reference those exports instead. It should already do that today if you explicitly export the return type. But TypeScript favors `import('sub_dep').Options` over `import('other_package').Options` even if `other_package` explicitly exports `Options`.


> If other_package has a dependency on sub_dep@2.0.0 and your package has a dependency on sub_dep@3.0.0, then these aren't the same type, and it's wrong.

@RyanCavanaugh I don't see how this is wrong. `other_package` would use `sub_dep@2.0.0` at runtime and your package would use `sub_dep@3.0.0` at runtime so the types would correctly reflect that.

> For the FAQ, another option to offer up is to extract the type from the original import to avoid adding additional (fragile?) routes to the type.

@robpalme `ReturnType` works in this simple scenario but I'm not aware how to refine it to a specific overload that way. For generics, TypeScript could emit instantiation expressions at least.

fatcerberus:
> I don't see how this is wrong. `other_package` would use `sub_dep@2.0.0` at runtime and your package would use `sub_dep@3.0.0` at runtime so the types would correctly reflect that.

The point of this is that `getMakeSubDep()` might, _at runtime_, return a type `subdep` that isn't compatible with the same type `subdep` as exported from 3.0.  It would be incorrect for TS to synthesize an import for the return type that refers to the latter.

eps1lon:
> It would be incorrect for TS to synthesize an import for the return type that refers to the latter.

It wouldn't if it would emit imports that reflect runtime imports. Resolving the right package versions is the responsibility of package managers and bundlers, not TypeScript.

eps1lon:
Right, I misread the package name. 

What about 

> Is there an option to enforce that whatever getMakeSubDep returns, is also exported from other_package? Then TypeScript could reference those exports instead. It should already do that today if you explicitly export the return type. But TypeScript favors import('sub_dep').Options over import('other_package').Options even if other_package explicitly exports Options.

then?

RyanCavanaugh:
> Is there an option to enforce that whatever getMakeSubDep returns, is also exported from other_package? 

I don't understand the proposal. If the package fully conceals its dependencies, as it should, then you won't have a problem. If it doesn't, you may or may not have a problem depending on how things go. If you turn on this rule, you'll always have a problem. Having more problems doesn't seem like a solution?

fz6m:
I use an automated script as a workaround, help TypeScript find the type for non-flattened dependency directory structures ( like pnpm ), first locate the real position of the indirect dependencies:

```ts
  const otherDepPath = path.dirname(require.resolve('other_package/package.json'))
  const subDepPath = path.dirname(
    require.resolve(`sub_dep/package.json`, { paths: [otherDepPath] })
  )
```

then add `/// <reference />` to make the types work:

```ts
/// <reference types="${subDepPath}" />
```

if the `/// <reference />` doesn't work, generating it in `tsconfig.json` as an alias, and the types work as expected:

```ts
// tsconfig.json

{
  "compilerOptions": {
    "paths": {
      "sub_dep": "/Users/path/to/pkg/dir" // `${subDepPath}`
    }
  }
}
```

src/compiler/moduleSpecifiers.ts
```diff
@@ -1026,6 +1028,46 @@ function tryGetModuleNameFromPackageJsonImports(moduleFileName: string, sourceDi
...
+    }
+    const packageJsonContent = cachedPackageJson?.contents.packageJsonContent || tryParseJson(host.readFile(packageJsonPath)!);
+    const dependencies = packageJsonContent?.dependencies;
+    if (!dependencies) {
+        return undefined;
+    }
+
+    const candidates = getOwnKeys(dependencies);
+    for (const name of candidates) {
+        const resolved = resolveModuleName(name, combinePaths(sourceDirectory, "dummy.ts"), options, host as (typeof host & { readFile: NonNullable<(typeof host)["readFile"]>; }), cache, /*redirectedReference*/ undefined, importMode);
```
andrewbranch, line 1057: Should `ModuleSpecifierResolutionHost` have a required `readFile` / be a subtype of `ModuleResolutionHost`? N.B. these functions may be called with hosts from public API usage, though I think `readFile` is safe because it’s required on `LanguageServiceHost`.

andrewbranch (commented):
This currently only works for the entrypoint resolved by the bare package name itself—how much work would it take to make it work for subpath `exports`?

src/compiler/moduleSpecifiers.ts
weswigham: I mean, this funky-lookin' cast is only here because 1. it's minimal (I could do `as ModuleResolutionHost`, I just feel like this cast more precisely follows the check done in the function body) and 2. there's a `if (!host.readFile) { return }` at the top which doesn't narrow the type as passed in here. Potential future improvement in checking there. But as for

> Should ModuleSpecifierResolutionHost have a required readFile / be a subtype of ModuleResolutionHost?

we could... but I think having it optional reminds us to always prefer to check caches before we go crawling the disk, so we at least have that fallback available wherever possible. It's reasonable if we do want to break the API, I just don't think we need to overmuch, and as it is, it keeps us in good habits.

andrewbranch: I missed the check for `host.readFile` earlier—no issues with this then 👍

weswigham:
> how much work would it take to make it work for subpath exports?

Hm, maybe not too bad? If we use a not-direct-match result that has the same affecting `package.json` as the target file (which should be cached, so cheapish to check - though equality may necessitate some `realpath` or psuedo-`realpath`ing), we can probably use that to hint to the existing lookup-from-known-packages logic to use that package. Maybe I can get away with just adding another entry into the symlink cache when the `realpath`s we do here discover one? Basically inject the newly found symlink into the symlink cache, and restart back from `forEachFileNameOfModule`? I'll have to look into it. It should be doable, at least, so long as we're interested.

weswigham:
@typescript-bot test it

@andrewbranch slightly new approach here - now, rather than looking for a `package.json` and its' dependencies on-demand, I do it up-front when resolution is first requested within a file. 99% of the time, this is going to just find cached and/or pre-create cached resolutions we were going to use anyway (since who doesn't reference their deps), but that remaining 1% is what fixes this issue and allows us to discover the symlink for the unused dependency, where it goes right into the resolution cache (which we now directly crawl to build the symlink cache, rather than only indirectly via the resolved modules).

weswigham:

weswigham:
Interesting. I'm only precaching `dependencies` specifiers for things not-in-`node_modules` now (since we don't need extra specifiers for things we can't directly reference from an input file anyway), and `vscode` and `mui-docs` still take a minor hit (3-5%) to parse time for it (parse includes program construction/resolution time and IO, iirc, which is usually actually most of it - the parser itself is fast - `extendedDiagnostics` would break these out), though not enough to move overall times. Meanwhile, `xstate` is _slightly faster overall_, due to faster `checker` time - likely because the extra symlinks we find make specifier generation in errors/declaration emit proceed more smoothly. The rest all have basically no change.

I think this is probably worth it at this point - the perf cost scales with how many unused dependencies you list in your `package.json`, which are exactly what we want to capture to be able to generate desirable paths for. What this PR basically does at this point is find all the `package.json`s affecting your input files, and act like your program has an import for all the `dependencies` declared in them.

I have another change to try and minimize the perf cost a wee bit more (it may or may not help, depends on our cache structure), but this looks good on the fundamentals to me at this point.

To see if this helps/hurts any.

weswigham:

I think this'll help the `vscode` case perf, unsure about `mui-docs` - all the packages there have local configs and `package.json`s; there might just be that many unused dependencies; I'll have to check.

weswigham:
> This does not include devDependencies, right? I think based on the description and code those should not be included. 

It does not. `dependencies` only.

weswigham:

weswigham:
There we go, no perf impact. We still inject into the symlink cache, we just do it lazily as specifiers are requested, instead of upfront.

andrewbranch (approved):
Awesome!

sheetalkamat (changes requested):
Need test for tsc-watch and need to make sure the resolution cache doesn’t get in bad state

src/compiler/moduleSpecifiers.ts
```diff
@@ -708,6 +712,29 @@ function getAllModulePaths(
...
+        const state = getTemporaryModuleResolutionState(cache.getPackageJsonInfoCache(), host, {});
+        const packageJson = getPackageScopeForPath(info.importingSourceFileName, state);
+        if (packageJson) {
+            const deps = packageJson.contents.packageJsonContent.dependencies;
+            if (deps && typeof deps === "object") {
+                const toResolve = getOwnKeys(deps as MapLike<unknown>);
+                for (const depName of toResolve) {
+                    const cached = cache.getOrCreateCacheForNonRelativeName(depName, /*mode*/ undefined, /*redirectedReference*/ undefined).get(info.canonicalSourceDirectory);
+                    if (cached) continue;
+                    const resolved = resolveModuleName(depName, info.importingSourceFileName, {}, host, cache);
```
sheetalkamat, line 731: Do the options not matter? do we not need `resolutionMode`

sheetalkamat (changes requested):

src/compiler/moduleSpecifiers.ts
```diff
@@ -708,6 +712,29 @@ function getAllModulePaths(
...
+    const links = host.getSymlinkCache?.();
+    if (cache && links && host.readFile && !pathContainsNodeModules(info.importingSourceFileName)) {
+        Debug.type<ModuleResolutionHost>(host);
+        // Cache resolutions for all `dependencies` of the `package.json` context of the input file.
+        // This should populate all the relevant symlinks in the symlink cache, and most, if not all, of these resolutions
+        // should get (re)used.
+        const state = getTemporaryModuleResolutionState(cache.getPackageJsonInfoCache(), host, {});
+        const packageJson = getPackageScopeForPath(info.importingSourceFileName, state);
+        if (packageJson) {
+            const deps = packageJson.contents.packageJsonContent.dependencies;
```
sheetalkamat, line 725: Also do we need to consider peer dependencies ?

src/compiler/moduleSpecifiers.ts
```diff
@@ -708,6 +712,29 @@ function getAllModulePaths(
...
+            if (deps && typeof deps === "object") {
+                const toResolve = getOwnKeys(deps as MapLike<unknown>);
+                for (const depName of toResolve) {
+                    const cached = cache.getOrCreateCacheForNonRelativeName(depName, /*mode*/ undefined, /*redirectedReference*/ undefined).get(info.canonicalSourceDirectory);
```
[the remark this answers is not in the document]
weswigham: Yeah, thinking about it, I probably shouldn't check the cache here at all (except maybe to check that one exists), and should just get the resolution result to feed into the symlink cache. `resolveModuleName`'ll internally do cache checking as appropriate.

src/compiler/moduleSpecifiers.ts
weswigham: They do not - we just need any single resolution for the module name, if resolvable, under any options/settings (since we're just interested in discovering a symlink in its' folder).

src/compiler/moduleSpecifiers.ts
weswigham: We could also lookup `peerDependencies`, if we'd like, yeah.

src/compiler/moduleSpecifiers.ts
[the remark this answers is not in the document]
weswigham: > Also i think you need to get redirectedReference from importngSourceFileName

Unreachable from here. I can get an actual redirected path form the host for the `importngSourceFileName`, if there is one, but the actual redirected reference object isn't available/exposed via the host. But also, I shouldn't really need it. A resolution of a nonrelative name from the og filepath vs the redirected path should result in the same thing, barring funky edgecases that'll make our emit assumptions invalid anyway. I could rerun the resolution from both paths... but I could also just run the resolution from the `package.json` itself. I think I like that more, honestly, and will swap to that. This is all just a heuristic to discover symlinks we haven't found yet from program construction anyway.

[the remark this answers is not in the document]
sheetalkamat: The reason we have redirected references is to cache the resolutions depending on module resolution affecting options. between projects. `importngSourceFileName`  is going to be file that is going to be file that is going to be emitted (for d.ts) and from correct project (for auto import provider) so is it safe to use `options` here? If so we should pass those as otherwise we dont lookup cached result if things dont match up

src/compiler/moduleSpecifiers.ts
andrewbranch: Isn’t this sensitive to our default `moduleResolution` option?

weswigham:
@typescript-bot test it

weswigham:
@sheetalkamat extended tests are still good, and I think I've addressed all your concerns.

src/compiler/moduleSpecifiers.ts
```diff
@@ -700,14 +707,45 @@ function getAllModulePaths(
...
 
-function getAllModulePathsWorker(info: Info, importedFileName: string, host: ModuleSpecifierResolutionHost): readonly ModulePath[] {
+const runtimeDependencyFields = ["dependencies", "peerDependencies", "optionalDependencies"] as const;
+
+function getAllRuntimeDependencies(packageJson: PackageJsonPathFields) {
+    let result;
+    for (const field of runtimeDependencyFields) {
+        const deps = packageJson[field];
+        if (deps && typeof deps === "object") {
+            result = concatenate(result, getOwnKeys(deps));
```
jakebailey, line 724: You can probably just use `append` here, given `getOwnKeys` is going to give a fresh copy anyway.

jakebailey (approved):

src/compiler/moduleSpecifiers.ts
jakebailey: Eh, wrong argument type. Nevermind.

sheetalkamat (approved):

src/compiler/moduleSpecifiers.ts
weswigham: You'd want `addRange`, not `append`, since `append` doesn't handle an array for input or take a spread list of arguments. But even that's not perfect - `concatenate` unconditionally allocates a new output array if both inputs are present (which is actually going to be rare here, usually it'll just be `dependencies`) - meanwhile `addRange` unconditionally copies `from` if `to` is *missing* (so it guarantees returning either `to` to a new array). That means it'd copy the first array needlessly every time. We actually don't have a "use either array mutably if they're present, it's fine"-type array-merging helper.

Honestly, it probably doesn't matter, though. Resizing one temporary array to merge in another one is probably going to trigger the similar allocations in the VM that making a new array would anyway (if they're both reasonably large). It's all short-lived objects.

PR #58177, opened 2024-04-12
fix(58166): Class parameter property with initializer before required property emits non-nullable parameter for declaration emit

### LINKED ISSUE #58166: Class parameter property with initializer before required property emits non-nullable parameter for declaration emit
### 🔎 Search Terms

class parameter property initializer declaration emit 

- This is the behavior in every version I tried (3.5.1 -> nightly 20240412), and I reviewed the FAQ for entries about (I skipped reviewing faq to be honest)

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.5.0-dev.20240412&ssl=11&ssc=2&pln=1&pc=1#code/PTAEHsGsCgFMA8AO4BOAXUBjANgQwM76gAqs+GA3tKFuAHbkoCumaqAFAG67ZOwBcoOkwC2AI1goA2gF1QAXlCyANKG69YAJkHDxkgJSgqNAL7Qz0EEPAYocJKgw4CRUuQAKuFLhHuU4RCNqWgY0ZlYORCYxbABLTDUePh1RCWk5RRVEjW0hVIMg03NoIA

### 💻 Code

```ts
// ok
export class Test {
  constructor(value: number[] = [], value2: number) {
  }
}

// not ok
export class TestParamProp {
  constructor(public value: number[] = [], value2: number) {
  }
}
```

### 🙁 Actual behavior

```ts
export declare class Test {
    constructor(value: number[] | undefined, value2: number);
}
export declare class TestParamProp {
    value: number[];
    constructor(value: number[], value2: number);
}
```

### 🙂 Expected behavior

```ts
export declare class Test {
    constructor(value: number[] | undefined, value2: number);
}
export declare class TestParamProp {
    value: number[];
    constructor(value: number[] | undefined, value2: number);
}
```

I was just writing weird test cases and ran into this one. This does not affect any real world code that I know about.

a-tarasyuk:
Fixes #58166


tests/baselines/reference/parameterPropertyInConstructor4.js
```diff
@@ -0,0 +1,28 @@
...
+        this.a = a;
+    }
+    return C;
+}());
+exports.C = C;
+
+
+//// [parameterPropertyInConstructor4.d.ts]
+export declare class C {
+    a: number[] | undefined;
```
weswigham, line 26: This is incorrect
```suggestion
    a: number[];

```
since the `= []` ensures that the property is always assigned, so only the parameter type should be `| undefined`.

weswigham (changes requested):

tests/baselines/reference/parameterPropertyInConstructor4.js
a-tarasyuk: @weswigham thanks for your review. I thought that both parameter and property declarations should use the same type...

a-tarasyuk: @weswigham could you review the latest changes? thanks

a-tarasyuk: @weswigham could you review the latest changes? thanks

a-tarasyuk: @weswigham could you review the latest changes? thanks

a-tarasyuk: @weswigham could you review the latest changes? thanks

weswigham (approved):

tests/baselines/reference/parameterPropertyInConstructor4.js
weswigham: Yeah, you can see the difference in an example like
```ts
class C {
	constructor(public arr: number[] = [], t: number) {}
}

const a = new C(undefined, 0); // allowed

a.arr = undefined; // is an error
```
The parameter and the property actually do have different internal types. The printer *might* not always be getting that correct right now with some recent changes to how nodes are reused, but this is definitely a step in the right direction.

Issue #58150, opened 2024-04-11
TS7018 error without file and line number


ktrz:
### 🔎 Search Terms

filename, noImplicitAny, TS7018, 7018

### 🕗 Version & Regression Information

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about TS7018, line number, filename

### ⏯ Playground Link

https://www.typescriptlang.org/play?strict=false&strictNullChecks=false&jsx=0&ssl=9&ssc=14&pln=8&pc=7#code/GYVwdgxgLglg9mABAEzgZSgCwOoywBQEMAnQgWwGcAKABxPIoC5FwBrMOAdzAEpEBvAFCJEMYIipQAnjQCmccXVKVEAQgC86xAHI4AIwBWs6Nr5CRI4rKghiSfgF8A3MMQPBrqzbsDXIiMzawHBw2gA0fog0gXok4ZFMOrEAXvEizoLughAIFFCIscSIWuZRzGAgADaVESKJFdW1iAB0ragYOHiYRMrUpSLROpiy1aFNdYGccMSVyGluPBEZQA

### 💻 Code

```ts
function doSthWithParams(params: unknown) {
  if (typeof params !== 'object') {
    return {};
  }

  return {
    c: 'foo',
    p: 'bar',
    s: 'baz',
  };
}

const bar = {
  p: null,
  s: null,
  ...doSthWithParams({
    p: 'hello',
    s: 'world',
  }),
};

```

### 🙁 Actual behavior

We're trying to incrementally get our codebase toward the strict mode. When we've enabled `noImplicitAny` flag we got an error that was very hard to track as it didn't give a file and line number where the error happens. 

```
Object literal's property 'p' implicitly has an 'any' type.

Object literal's property 's' implicitly has an 'any' type.
```

### 🙂 Expected behavior

The TS compiler error should point to a file and line number where the error occurs

### Additional information about the issue

This happens if `noImplicitAny` is true and `strictNullChecks` is not yet enabled

fatcerberus:
This just looks like a bug - there shouldn't be an error in this code at all.

Andarist:
> This just looks like a bug - there shouldn't be an error in this code at all.

What would be the correct behavior here? We get no implicit any error here outside of `strictNullChecks` (which this issue is about):
```ts
const obj = {
  prop: null, // Object literal's property 'prop' implicitly has an 'any' type.(7018)
};
```

In the OP's example the computed type of `bar` is:
```ts
[...]

PR #58157, opened 2024-04-12
Fixed widening errors locations in union-normalized object literal types

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/58150

PR #58180, opened 2024-04-13
Fixed propagating object flags when dealing with spreads

Andarist:
spotted by @fatcerberus here: https://github.com/microsoft/TypeScript/issues/58150#issuecomment-2052517378

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -18985,7 +18985,10 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         const skippedPrivateMembers = new Set<__String>();
         const indexInfos = left === emptyObjectType ? getIndexInfosOfType(right) : getUnionIndexInfos([left, right]);
 
+        // propagating flags of the right properties should be preserved if they override the left properties
+        let finalObjectFlags = objectFlags & ~ObjectFlags.PropagatingFlags;
         for (const rightProp of getPropertiesOfType(right)) {
+            finalObjectFlags |= getObjectFlags(getTypeOfSymbol(rightProp)) & ObjectFlags.PropagatingFlags;
             if (getDeclarationModifierFlagsFromSymbol(rightProp) & (ModifierFlags.Private | ModifierFlags.Protected)) {
                 skippedPrivateMembers.add(rightProp.escapedName);
             }
@@ -19021,12 +19024,13 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 }
             }
             else {
+                finalObjectFlags |= getObjectFlags(getTypeOfSymbol(leftProp)) & ObjectFlags.PropagatingFlags;
                 members.set(leftProp.escapedName, getSpreadSymbol(leftProp, readonly));
             }
         }
 
         const spread = createAnonymousType(symbol, members, emptyArray, emptyArray, sameMap(indexInfos, info => getIndexInfoWithReadonly(info, readonly)));
-        spread.objectFlags |= ObjectFlags.ObjectLiteral | ObjectFlags.ContainsObjectOrArrayLiteral | ObjectFlags.ContainsSpread | objectFlags;
+        spread.objectFlags |= ObjectFlags.ObjectLiteral | ObjectFlags.ContainsObjectOrArrayLiteral | ObjectFlags.ContainsSpread | finalObjectFlags;
         return spread;
     }
 
--- a/tests/baselines/reference/noImplicitAnySpreads1.errors.txt
+++ b/tests/baselines/reference/noImplicitAnySpreads1.errors.txt
@@ -1,0 +1,22 @@
+noImplicitAnySpreads1.ts(13,3): error TS7018: Object literal's property 'other' implicitly has an 'any' type.
+
+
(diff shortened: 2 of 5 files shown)


tests/baselines/reference/noImplicitAnySpreads1.errors.txt
```diff
@@ -0,0 +1,22 @@
...
+      return {
+        c: "foo",
+        p: "bar",
+        s: "baz",
+      };
+    }
+    
+    const bar = {
+      p: null, // no error, this gets overriden
+      other: null, // error, this does not get overriden
```
sandersn, line 17: I am confused. This is the behaviour I get when I paste the test into the playground and turn off strictNullChecks. What does this PR fix about this test?

Also, this isn't the case mentioned in the linked comment, as far as I can tell.

sandersn (changes requested):

tests/baselines/reference/noImplicitAnySpreads1.errors.txt
Andarist: To be honest... I'm confused too. I've checked out previous commits, re-run this test and all and it doesn't make a lot of sense to me why I have added this here.

**Maybe** this is a test case validating that I didn't break this behavior with my changes and I forgot to add a test case from the referenced comment? That case from the comment **does** exhibit a change. I just added it to the test file here and u can compare it with the current non-sensical behavior [here](https://www.typescriptlang.org/play/?strictNullChecks=false&ts=5.6.0-dev.20240615#code/PTAEAEGcBcCcEsDG0ByBXANhgwgCwKaIDWkAXKAGYCGGk+AUCBAHYD2AkgLYAOGS80AILMAnuThoGFNM2TxWzUABNWAZWi4A6gNwAFKrCqdIACm4GjZUDKJsA7swCUoAN71QoWPmhpYitx4eiOQA5BSsrCEANO6B3KEARgbRsR5WIUkAXikeAL4A3PS59PSICjCgWaAAvK6x8aDMmBgxaeRNWK2gAHS9KupaOvqGxiYBcaEEWJFdbaAhdqywGEo5oLmOMQVAA)

src/compiler/checker.ts
```diff
@@ -19023,7 +19023,10 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         const skippedPrivateMembers = new Set<__String>();
         const indexInfos = left === emptyObjectType ? getIndexInfosOfType(right) : getUnionIndexInfos([left, right]);
 
+        // propagating flags of the right properties should be preserved if they override the left properties
```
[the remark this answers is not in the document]
Andarist: No, i dont think u are misunderstanding - this comment wasn't super clear. I pushed out a change to make it more specific, I hope this helps

tests/cases/compiler/noImplicitAnySpreads1.ts
```diff
@@ -0,0 +1,19 @@
+// @strictNullChecks: false
+// @noImplicitAny: true
```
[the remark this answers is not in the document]
Andarist: I default all the tests I add to `@strict: true, @noEmit: true`. This combination of flags is important to showcase the behavior change here.

I agree it would be great if we'd have some other test case that wouldn't require those. I'll try to come up with some, it might be the easiest with `ObjectFlags.ContainsObjectOrArrayLiteral`.

tests/baselines/reference/noImplicitAnyUnionNormalizedObjectLiteral1.errors.txt
Andarist, line 1: this "conflicted" with the recently merged https://github.com/microsoft/TypeScript/pull/58157

The resulting type of the variable here is `{ c?: undefined; p: any; s: any; } | { c: string; p: string; s: string; }`. I think that in this situation it should error as `bar.p` results in `any`. But at the same time, i think it will be easier/better to fix the union normalization in this case so the result of this would become `{ c?: undefined; p: string; s: string; } | { c: string; p: string; s: string; }`

tests/cases/compiler/noImplicitAnySpreads1.ts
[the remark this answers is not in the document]
Andarist: Hm, I have tried to come up with a test case that would rely on the change in how `ObjectFlags.ContainsObjectOrArrayLiteral` or `ObjectFlags.NonInferrableType` might get propagated but I failed.

tests/baselines/reference/noImplicitAnyUnionNormalizedObjectLiteral1.errors.txt
Andarist: After more thinking, I'm not sure if this would be the right solution. `@strictNullChecks: false, @noImplicitAnt: true` is really a confusing combination.

This behavior doesn't help me either when it comes to determining what's the right solution here:
```ts
// @strictNullChecks: false
// @noImplicitAny: true

let test1 = null; // no error 😱
let test2 = {
  prop: null, // Object literal's property 'prop' implicitly has an 'any' type.(7018)
};
```

It all depends on what kind of conceptual safety this combination of settings is supposed to give the user.

PR #58183, opened 2024-04-13
fix(58151): JSDoc tags get lost when inheriting from a grandparent class

### LINKED ISSUE #58151: JSDoc tags get lost when inheriting from a grandparent class
### 🔎 Search Terms

jsdoc tag inherit

### ⏯ Playground Link

https://www.typescriptlang.org/play?#code/MYGwhgzhAECC0G8CwAoa7oHoBU3qo3TwBEBTCYAJwEsAHAF2oHsA7aAC1MtIDoDC8-DHgACAN2qV6AVzAghRaNkwKAZkyYAKAJSJoAX1SGUqUJBgAhaKQAe9UiwAmMeMjQYmYrjUelo6rV0EAyNUU3AoaABhazsHZ2grN0JPb2pffw0dPWN9IA

### 💻 Code

```ts
class A {
    /** 
     * Description text here.
     * 
     * @virtual
     * */
    foo() { }
}

class B extends A {
    override foo() { }
}

class C extends B {
    override foo() { }
}
```

### 🙁 Actual behavior

Only the description text is provided for `C.foo`.

### 🙂 Expected behavior

Both the description text and the tag should be provided for `C.foo`, just as the behavior of `B.foo`.

I'm trying to provide a solution for https://github.com/microsoft/TypeScript/issues/1534 by using typescript-eslint + JSDoc tag. This bug makes `@virtual` tag disappear and I will not be able to override `C.foo` in "strict-virtual" mode.

Also note that `@inheritdoc` tag doesn't work here. Adding `@inheritdoc` to `B.foo` will only provide `@inheritdoc` itself to `C.foo`, instead of the `@virtual` tag, which is an unexpected and unuseful behavior.

a-tarasyuk:
Fixes #58151

### DIFF
--- a/src/services/services.ts
+++ b/src/services/services.ts
@@ -740,6 +740,7 @@ class SymbolObject implements Symbol {
 
     getJsDocTags(checker?: TypeChecker): JSDocTagInfo[] {
         if (this.tags === undefined) {
+            this.tags = emptyArray; // Set temporarily to avoid an infinite loop finding inherited tags
             this.tags = getJsDocTagsOfDeclarations(this.declarations, checker);
         }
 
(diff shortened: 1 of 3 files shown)


jakebailey:
@typescript-bot pack this

jakebailey (approved):
Looks to work, and makes sense to me.

PR #58187, opened 2024-04-14
Strip JSDoc from private properties in declaration emit

### LINKED ISSUE #58145: Declaration emit of private properties should strip jsdoc
### 🔎 Search Terms

jsdoc dts declaration emit private properties property

### ⏯ Playground Link

https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBAYwDYEMDOa4DEITgbwCg4S4B6AKgrgBUALAS0ybhTjAFcAjJBhOALbAYdCABMAdHAplipTjz7tuvBAFlhosQAoAlAQC+hOSUrV6LFmzBQGANxQxggzeIA0rAHZi4AdwZISHBczmgwtmBgwD4MnnAizmLAyChQjgwQcQBmAcBSMibstg5ORfaOwBoi4nqGhAZAA

### 💻 Code

```ts
export class Foo {
    /** This is a public method. */
    public publicMethod() {}

    /** This is a private method, and will be stripped in the declaration file. */
    private privateMethod() {}
}
```

### 🙁 Actual behavior

```ts
export declare class Foo {
    /** This is a public method. */
    publicMethod(): void;
    /** This is a private method, and will be stripped in the declaration file. */
    private privateMethod;
}
```

Note the jsdoc on the private method.

### 🙂 Expected behavior

```ts
export declare class Foo {
    /** This is a public method. */
    publicMethod(): void;
    private privateMethod;
}
```

It sure seems like this info should be stripped; after all, the type information is gone, leaving only a marker to inform child classes.

This causes TypeScript's own dts files to be filled with private method jsdoc. https://github.com/microsoft/TypeScript/pull/57492/files#r1559884318

Zzzen:
Fixes #58145

### DIFF
--- a/src/compiler/transformers/declarations.ts
+++ b/src/compiler/transformers/declarations.ts
@@ -12,6 +12,7 @@ import {
     canHaveModifiers,
     canProduceDiagnostics,
     ClassDeclaration,
+    ClassElement,
     compact,
     concatenate,
     ConditionalTypeNode,
@@ -764,6 +765,15 @@ export function transformDeclarations(context: TransformationContext) {
     function checkEntityNameVisibility(entityName: EntityNameOrEntityNameExpression, enclosingDeclaration: Node) {
         const visibilityResult = resolver.isEntityNameVisible(entityName, enclosingDeclaration);
         handleSymbolAccessibilityError(visibilityResult);
+    }
+
+    function removeCommentsFromPrivateClassMembers(node: NodeArray<ClassElement>) {
+        node.forEach(member => {
+            if (hasEffectiveModifier(member, ModifierFlags.Private)) {
+                removeAllComments(member);
+            }
+        });
+        return node;
     }
 
     function preserveJsDoc<T extends Node>(updated: T, original: Node): T {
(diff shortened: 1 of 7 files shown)


Zzzen:
It drives me crazy that it doesn't work on `typescript.d.ts`

jakebailey:
> It drives me crazy that it doesn't work on `typescript.d.ts`

You'd need to bump TS to another version of itself; we do that sometimes and maybe this would be a good reason to do so early before the beta cut.

jakebailey:
Alternatively we could make tsc always compile itself, bootstrap style. But that will definitely be odd to set up thanks to `--no-bundle`.

jakebailey:
I _think_ you'd be able to test with `hereby dts --built`... (I'm not at my machine to try)

jakebailey:
Yeah, `hereby tsc && hereby dts --built` works. But it also shows that the current PR strips away the `@internal` comments before that info can be read.

Now, maybe that's not actually a major problem, as their types are guaranteed to be stripped and therefore you won't end up with any unresolved types.

Here's the diff: https://gist.github.com/jakebailey/67b4aafa550361f169bbbfeb9f747794

jakebailey (changes requested):
Marking as needing review per the above; this needs some tests for the `@internal` thing and a fix so that TS's own declaration emit doesn't regress.

jakebailey:
Oh, hm, I'm brainfarting; the problem path is:

- After this PR, declaration emit strips comments from all private properties
- TypeScript doesn't use stripInternal, instead post-processing the output dts files to strip them
- Now, we see the private properties without `@internal`, so don't strip them in `dtsBundler.mjs`

So, I don't know that there's actually a solution to this... Hm...

jakebailey:
Oh, I forgot, I was working on modifying our lint rule to complain about `@internal` on `private`.

jakebailey:
I sent #58869 to try and handle this a little better for our case.

One gotcha is that this may negatively affect any tools which also post-process declaration files, as they won't see comments on private declarations anymore. However, not sure how much of a problem it is in practice.

Now, in the process of writing #58869, I recognized that there are more declarations that can be private than I thought. This PR tests properties and methods on classes, but index signatures can also be private, so I believe that will need a test as well.

PR #58188, opened 2024-04-14
lib.d.ts: add missing TypedArray-from-iterable signatures

### LINKED ISSUE #45198: Big(Int|Uint)64ArrayConstructor: missing from Iterable
# lib Update Request

## Configuration Check

My compilation *target* is `ESNext` and my *lib* is `the default`.

`BigInt64Array.from` is missing an iterable overload (same for BigUint64Array).

## Sample Code

```ts
console.log(
  Int32Array.from(new Set([1, 2, 3]), (value) => Number(value) ** 2).join(", ")
);
console.log(
  BigInt64Array.from(new Set([1n, 2n, 3n]), (value) => value ** 2n).join(", ")
);
```

```stderr
Argument of type 'Set<bigint>' is not assignable to parameter of type 'ArrayLike<unknown>'.
  Property 'length' is missing in type 'Set<bigint>' but required in type 'ArrayLike<unknown>'.
```

## Documentation Link

[TypedArray.from() - JavaScript | MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/from)

### LINKED ISSUE #45199: TypedArray constructors: missing from mapping iterable
# lib Update Request

## Configuration Check

My compilation *target* is `ESNext` and my *lib* is `the default`.

Typed array constructors (`Int8Array`, `Float32Array`, `BigInt64Array`, etc.) are missing an iterable overload that transforms from some non-`number`/`bigint` type to the desired `number`/`bigint` type.

## Sample Code

```ts
console.log(
  Int32Array.from(new Set(["1", "2", "3"]), (value) => Number(value) ** 2).join(", ")
);
```

```stderr
  Overload 1 of 3, '(arrayLike: Iterable<number>, mapfn?: ((v: number, k: number) => number) | undefined, thisArg?: any): Int32Array', gave the following error.
    Argument of type 'Set<string>' is not assignable to parameter of type 'Iterable<number>'.
      The types returned by '[Symbol.iterator]().next(...)' are incompatible between these types.
        Type 'IteratorResult<string, any>' is not assignable to type 'IteratorResult<number, any>'.
          Type 'IteratorYieldResult<string>' is not assignable to type 'IteratorResult<number, any>'.
            Type 'IteratorYieldResult<string>' is not assignable to type 'IteratorYieldResult<number>'.
  Overload 2 of 3, '(arrayLike: ArrayLike<unknown>, mapfn: (v: unknown, k: number) => number, thisArg?: any): Int32Array', gave the following error.
    Argument of type 'Set<string>' is not assignable to parameter of type 'ArrayLike<unknown>'.
      Property 'length' is missing in type 'Set<string>' but required in type 'ArrayLike<unknown>'.
```

## Documentation Link

[TypedArray.from() - JavaScript | MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/from)

Renegade334:
Resolves #45198
Resolves #45199

There are stale PRs linked to these issues, but those involved wider remodelling of the array library. This should be simple enough to merge independently.


jakebailey:
@typescript-bot test it

src/lib/es2020.bigint.d.ts
```diff
@@ -384,14 +384,19 @@ interface BigInt64ArrayConstructor {
      */
     of(...items: bigint[]): BigInt64Array;
 
+    /**
+     * Creates an array from an array-like or iterable object.
+     * @param arrayLike An array-like or iterable object to convert to an array.
+     */
+    from(arrayLike: ArrayLike<bigint> | Iterable<bigint>): BigInt64Array;
```
sandersn, line 391: why do these differ from the types in TypeArray.from? For that one, we just use `Iterable`.

Some guesses:
- es2020.bigint can be used without including es2015.iterable (or whatever file makes Arrays Iterable)
- Iterable is added here because it's in TypedArray.from, but ArrayLike stays for backward compatibility.

sandersn (changes requested):
Looks good except I have one question about addition of Iterable in es2020.bigint

src/lib/es2020.bigint.d.ts
Renegade334: Because in those cases, the `ArrayLike` overloads had already been defined in es5 prior to the introduction of iterables, so the `ArrayLike` signature and the `Iterable` signature exist as two separate overloads (_eg._ [here](https://github.com/microsoft/TypeScript/blob/v5.5-rc/src/lib/es5.d.ts#L2119-L2123) and [here](https://github.com/microsoft/TypeScript/blob/v5.5-rc/src/lib/es5.d.ts#L2119-L2123)). Since the bigint typed arrays are post-es2015, it's possible to have these as one single overload.

`ArrayLike` objects don't have to be iterable, so the method definitions need to accept both:
```ts
BigInt64Array.from(
  // ArrayLike<number> but not iterable
  {
    0: 12345678,
    1: 98765432
    length: 2
  },
  (n: number): bigint => BigInt(n) ** 2n,
) // BigInt64Array [ 152415765279684n, 9754610558146624n ]
```

Renegade334: On that subject, though, the bigint TypedArray constructors in es2020.bigint don't look like they have a `new (ArrayLike<bigint>)` signature, which probably also needs addressing.

tests/cases/compiler/bigintWithLib.ts
```diff
@@ -19,10 +19,18 @@ stringVal = bigintVal.toLocaleString('de-DE', { style: 'currency', currency: 'EU
...
+bigIntArray = new BigInt64Array({0: 1n, 1: 2n, 2: 3n, length: 3});
 bigIntArray = new BigInt64Array([1, 2, 3]); // should error
+bigIntArray = new BigInt64Array({0: 1, 1: 2, 2: 3, length: 3}); // should error
 bigIntArray = new BigInt64Array(new ArrayBuffer(80));
 bigIntArray = new BigInt64Array(new ArrayBuffer(80), 8);
 bigIntArray = new BigInt64Array(new ArrayBuffer(80), 8, 3);
+bigIntArray = BigInt64Array.from([1n, 2n, 3n]);
+bigIntArray = BigInt64Array.from([1n, 2n, 3n], n => n * 10n);
+bigIntArray = BigInt64Array.from({0: 1n, 1: 2n, 2: 3n, length: 3});
+bigIntArray = BigInt64Array.from({0: 1n, 1: 2n, 2: 3n, length: 3}, n => n * 10n);
```
Renegade334, line 31: re. [this conversation](https://github.com/microsoft/TypeScript/pull/58188#discussion_r1640355261), this is the case that necessitates keeping `ArrayLike` alongside `Iterable`.

Issue #44794, opened 2021-06-28
Symbol properties should not be exempt from excess property checks in presence of a string index signature


DanielRosenwasser:
In #44793, we found that a `symbol` index signature doesn't turn off excess property checks on `symbol`s; however, a *`string`* index signature turns these excess property checks entirely.

```ts
interface A {
    a?: string
    b?: number
    [key: string]: unknown
}

const symbolKey = Symbol('key')

const a: A = {
    [symbolKey]: '123', // works!
}
```

It's likely that some users relied on this behavior; however, those users will be able to use `symbol` index signatures in TypeScript 4.4.

andrewbranch:
> Should symbols be exempt from excess property checks in presence of a string index signature?

Nope

DanielRosenwasser:
Well

> It's likely that some users relied on this behavior

and from @ahejlsberg's comment at https://github.com/microsoft/TypeScript/issues/44793#issuecomment-869966487

> It's an easy fix, but the odd thing is that it ends up being a breaking change because we previously would validate symbol-named properties again string index signatures when performing excess property checking.

andrewbranch:
Well I assumed you’d weigh my “nope” against other opinions and data, but it seems like a good breaking change to me (one I’d vet in 4.5, unless we want to try to backport it into 4.4-beta).

DanielRosenwasser:
I would prefer not picking that into 4.4 at this point. My preference is to give people a step-wise upgrade pattern. Ship one version with the string indexer behavior, and then rip off the band-aid in 4.5.

PR #58190, opened 2024-04-14
Don't exempt late bound symbols from EPC in presence of string index signature

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/44794


jakebailey:
@typescript-bot test it

jakebailey:
Shocking

@typescript-bot test top800

Andarist:
Symbols are not that popular in the wild and often might not even be used directly so I'm not that shocked ;p

fatcerberus:
> Symbols are not that popular in the wild

Given that you either write something ugly like `foo[Symbol.someBuiltInSymbol] = whatever` or else `foo[localSymVar] = bar` which is syntactically indistinguishable from a string key access (and therefore potentially confusing), people shying away from symbols isn't that surprising.  IMO symbols were a cool concept but they really needed specialized syntax.

PR #58201, opened 2024-04-15
Isolated declarations errors

dragomirtitian:
Added command line flag for isolated declarations that will report errors on constructs that can't easily be inferred from local information.

Some examples (not an exhaustive list):

```ts
// ❌ Expressions must not use information outside the expression (we can't know what the call will return)
export const nBad = Math.random();
//                  ^^^^^^^^^^^^^
// Variable must have an explicit type annotation with --isolatedDeclarations.ts(9010)

export const nOk: number = Math.random();

export const lOk = 1;

// ❌ Expressions must not reference other variables
export const refConst = lOk;
//                      ^^^
// Variable must have an explicit type annotation with --isolatedDeclarations.

// ❌ Inferred return types are not supported
export function noReturn(a: number) { }
//              ^^^^^^^^
// Function must have an explicit return type annotation with --isolatedDeclarations.

export function explicitReturnOk(a: number): void { }

// ✔ Object literals are supported 
export const colors = {
    'blue': 0x0000FF,
    'red': 0xFF0000,
    'green': 0x00FF00,
} as const

// ✔ Mutable object literals are also supported
export const point = {
    x: 1,
    y: 1,
    dist(): number {
        return Math.sqrt(this.x * this.x + this.y * this.y);
    }
}

// ❌ Spread operators are not supported
export const newPoint = { ...point };
//                        ^^^^^^^^
// Objects that contain spread assignments can't be inferred with --isolatedDeclarations.

// ❌ Mutable arrays are not supported
export const values = [1, 2, 3]
//                    ^^^^^^^^^
// Only const arrays can be inferred with --isolatedDeclarations.ts(9017)

// ✔ Const arrays are supported
export const directions = ["UP", "DOWN", "LEFT", "RIGHT"] as const

// ✔ Function expressions are supported if all types are specified
export const component = (props: { name: string }): string => "";

```


src/compiler/program.ts
```diff
@@ -4229,6 +4229,16 @@ export function createProgram(rootNamesOrOptions: readonly string[] | CreateProg
             }
         }
 
+        if (options.isolatedDeclarations) {
+            if (options.out) {
+                createDiagnosticForOptionName(Diagnostics.Option_0_cannot_be_specified_with_option_1, "out", "isolatedDeclarations");
+            }
```
jakebailey, line 4235: This option is deprecated; I think you can safely drop it now.

src/compiler/checker.ts
```diff
@@ -1100,6 +1103,7 @@ import {
 } from "./_namespaces/ts";
 import * as moduleSpecifiers from "./_namespaces/ts.moduleSpecifiers";
 import * as performance from "./_namespaces/ts.performance";
+import { createSyntacticTypeNodeBuilder } from "./expressionToTypeNode";
```
jakebailey, line 1106: It's probably fine to start doing this if it works, but we generally aren't yet using direct imports and instead export stuff via the big module.

src/compiler/checker.ts
```diff
@@ -5980,9 +5985,27 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
-            serializeTypeForDeclaration: (declaration: Declaration, type: Type, symbol: Symbol, enclosingDeclaration?: Node, flags?: NodeBuilderFlags, tracker?: SymbolTracker) => withContext(enclosingDeclaration, flags, tracker, context => serializeTypeForDeclaration(context, declaration, type, symbol)),
-            serializeReturnTypeForSignature: (signature: Signature, enclosingDeclaration?: Node, flags?: NodeBuilderFlags, tracker?: SymbolTracker) => withContext(enclosingDeclaration, flags, tracker, context => serializeReturnTypeForSignature(context, signature)),
+            expressionOrTypeToTypeNode: (expr: Expression | JsxAttributeValue | undefined, type: Type, addUndefined?: boolean, enclosingDeclaration?: Node, flags?: NodeBuilderFlags, tracker?: SymbolTracker) =>
+                withContext(enclosingDeclaration, flags, tracker, context => {
+                    if (expr) {
+                        syntacticNodeBuilder.serializeTypeOfExpression(expr, context, addUndefined);
+                    }
+                    return expressionOrTypeToTypeNode(context, expr, type, addUndefined);
```
jakebailey, line 5993: This is extra logic that this code didn't have before; is there a reason to not have it within `expressionOrTypeToTypeNode` or something?

src/compiler/checker.ts
```diff
@@ -6112,6 +6135,18 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 approximateLength: 0,
                 trackedSymbols: undefined,
                 bundled: !!compilerOptions.outFile && !!enclosingDeclaration && isExternalOrCommonJsModule(getSourceFileOfNode(enclosingDeclaration)),
+                isEntityNameVisible: (entityName, shouldComputeAliasToMakeVisible) => isEntityNameVisible(entityName, context.enclosingDeclaration!, shouldComputeAliasToMakeVisible),
+                isExpandoFunctionDeclaration,
+                isOptionalParameter,
```
jakebailey, line 6140: At some point I feel like we need to do something better for this function; it depends on semantic info which I think is not so good for isolatedDeclarations, no? Or is it just here to error when it differs from a syntactic interpretation?

src/compiler/checker.ts
```diff
@@ -48243,12 +48283,24 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
 
     function isExpandoFunctionDeclaration(node: Declaration): boolean {
-        const declaration = getParseTreeNode(node, isFunctionDeclaration);
+        const declaration = getParseTreeNode(node, (n): n is FunctionDeclaration | VariableDeclaration => isFunctionDeclaration(n) || isVariableDeclaration(n));
         if (!declaration) {
             return false;
         }
-        const symbol = getSymbolOfDeclaration(declaration);
-        if (!symbol || !(symbol.flags & SymbolFlags.Function)) {
+        let symbol: Symbol;
```
jakebailey, line 48290:
```suggestion
        let symbol: Symbol | undefined;
```

I'm guessing, just based on the use below.

src/compiler/types.ts
```diff
@@ -10203,3 +10217,29 @@ export interface EvaluationResolver {
...
+    | PropertySignature
+    | ExportAssignment;
+
+/** @internal */
+export interface SyntacticTypeNodeBuilderContext {
+    tracker: Required<Pick<SymbolTracker, "reportInferenceFallback">>;
+    isUndefinedIdentifier(name: Identifier): boolean;
+    isLiteralComputedName(name: ComputedPropertyName): boolean;
+    isExpandoFunctionDeclaration(name: FunctionDeclaration | VariableDeclaration): boolean;
+    isOptionalParameter(name: ParameterDeclaration): boolean;
```
jakebailey, line 10240: This looks unused; maybe others are too?

src/compiler/checker.ts
```diff
@@ -5980,9 +5985,27 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
-            serializeTypeForDeclaration: (declaration: Declaration, type: Type, symbol: Symbol, enclosingDeclaration?: Node, flags?: NodeBuilderFlags, tracker?: SymbolTracker) => withContext(enclosingDeclaration, flags, tracker, context => serializeTypeForDeclaration(context, declaration, type, symbol)),
-            serializeReturnTypeForSignature: (signature: Signature, enclosingDeclaration?: Node, flags?: NodeBuilderFlags, tracker?: SymbolTracker) => withContext(enclosingDeclaration, flags, tracker, context => serializeReturnTypeForSignature(context, signature)),
+            expressionOrTypeToTypeNode: (expr: Expression | JsxAttributeValue | undefined, type: Type, addUndefined?: boolean, enclosingDeclaration?: Node, flags?: NodeBuilderFlags, tracker?: SymbolTracker) =>
+                withContext(enclosingDeclaration, flags, tracker, context => {
+                    if (expr) {
```
weswigham, line 5990: By calling these here instead of within `expressionOrTypeToTypeNode` directly, these aren't going to be executed during JS declaration emit (whose only top-level entrypoint is `symbolTableToDeclarationStatements`), since that calls the internal functions, not the external ones. As such, unless you're calling using `isolatedDeclarations` with `allowJs` an error (but why), you wanna shift these checks into the functions of the self-safe names, rather than here in the external API wrapper functions. These should really just be adapting the external arguments to the internal ones.

src/compiler/diagnosticMessages.json
```diff
@@ -6227,6 +6227,11 @@
         "category": "Message",
         "code": 6718
     },
+    "Ensure that each file can have declaration emit generated without type information": {
```
weswigham, line 6230:
```suggestion
    "Report errors on emitted declarations which rely excessively on type inference to produce the emit.": {

```
since
1. It doesn't actually "ensure" anything (since it's not adjusting the emit), it just adds errors (I'd have the same remark on the `isolatedModules` description - though that does sometimes affect emit, so it's less egregious). And
2. Type information (or something close enough to it - it's not like types are magical) is still being used, just only a subset of more local, simpler inferences (since it's not like we're issuing a diagnostic on *every* missing annotation, we do have a bunch of carve-outs that are "simple enough").

Does that reek of more ambiguity than the original message? Yes. But what this does is ambiguous - you really do need to know if it's reporting on what you want or not. For example, I doubt this is going to hit up everything you'd need to issue an error on for a JSR registry declaration emitter as it exists right now - the description they give is much more constrained than what we have here. This flag is basically a "set this if someone in your toolchain told you to" flag. Moreover, by keeping the description a bit more wishy-washy, when someone comes to us and says "Hey, `export const a = [1]`  is missing an annotation, why doesn't it have an error in `isolatedDeclarations`?" we can say "Well, the type's self-evident from the expression and we think any tool worth its' salt can synthesize a type node for it." without as much "oh, well, there's not _no_ inference going on because that'd be unusable in a laundry list of common examples" explaination. Plus, the ambiguity gives room for future changes, since we already *know* there's probably gonna be some.

src/compiler/commandLineParser.ts
```diff
@@ -827,6 +827,15 @@ const commandOptionsWithoutBuild: CommandLineOption[] = [
...
         defaultValueDescription: false,
     },
+    {
+        name: "isolatedDeclarations",
+        type: "boolean",
+        category: Diagnostics.Interop_Constraints,
+        description: Diagnostics.Ensure_that_each_file_can_have_declaration_emit_generated_without_type_information,
+        defaultValueDescription: false,
+        affectsBuildInfo: true,
+        affectsEmit: true,
```
weswigham, line 837: Also `affectsSemanticDiagnostics: true`, since this does have errors emitted in the checker. Honestly, `affectsEmit: false` may actually be more true as well, since afaik this does not actually affect our output.

src/compiler/expressionToTypeNode.ts
```diff
@@ -0,0 +1,504 @@
...
+                    if (unaryExpression.operand.kind === SyntaxKind.NumericLiteral) {
+                        return typeFromPrimitiveLiteral();
+                    }
+                }
+                break;
+            case SyntaxKind.NumericLiteral:
+                return typeFromPrimitiveLiteral();
+            case SyntaxKind.TemplateExpression:
+                if (!isConstContext && !preserveLiterals) {
+                    return true;
```
weswigham, line 278: ...is this right? Eg, for
```ts
const a = "a";
export const b = `b${a}`
```
we emit
`export declare const b = "ba";` but for
```ts
let a = "a";
export const b = `b${a}`
```
we emit `export declare const b: string;` - you can't know if a template literal type should be emitted as a literal type (or initializer) or `string` without doing evaluation, and if that evaluation crosses file bounds... 🤷‍♂️ I'm pretty sure this has been a consideration, given the `evaluator` changes in this PR, it just doesn't seem to be reflected in the diagnostic.

src/compiler/program.ts
```diff
@@ -4229,6 +4229,16 @@ export function createProgram(rootNamesOrOptions: readonly string[] | CreateProg
             }
         }
 
+        if (options.isolatedDeclarations) {
+            if (options.out) {
+                createDiagnosticForOptionName(Diagnostics.Option_0_cannot_be_specified_with_option_1, "out", "isolatedDeclarations");
```
weswigham, line 4234: ...Is there a reason for this? It doesn't seem like anything in the implementation particularly cares about whether the output is concatenated by tsc or not. Or did I miss an error on all script-mode files somewhere and the reason for it?

src/compiler/checker.ts
```diff
@@ -48243,12 +48283,24 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
-        const declaration = getParseTreeNode(node, isFunctionDeclaration);
+        const declaration = getParseTreeNode(node, (n): n is FunctionDeclaration | VariableDeclaration => isFunctionDeclaration(n) || isVariableDeclaration(n));
         if (!declaration) {
             return false;
         }
-        const symbol = getSymbolOfDeclaration(declaration);
-        if (!symbol || !(symbol.flags & SymbolFlags.Function)) {
+        let symbol: Symbol;
+        if (isVariableDeclaration(declaration)) {
+            if (declaration.type || !isVarConstLike(declaration)) {
```
weswigham, line 48292: This is only true in `TS` (where only declared-with-function-syntax things are expando) - in JS,
```ts
let a = function() {};
a.bar = 1;
```
is both valid, and expando.

cc @sandersn for eyes on this part of the change, since you probably have the most context on what is and is not "expando".

src/compiler/checker.ts
```diff
@@ -48585,6 +48637,28 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         }
         return false;
     }
+    function isLiteralComputedName(node: ComputedPropertyName) {
```
weswigham, line 48640: This feels like it overlaps with `hasBindableName` a _lot_. If I had to tell you at a glance how this differed in its' selectivity from that, and why, I'm not sure I could. Should this be `hasBindableName(node.parent)` with some extra addenda further restricting it?

src/compiler/checker.ts
```diff
@@ -6112,6 +6135,18 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+                isExpandoFunctionDeclaration,
+                isOptionalParameter,
+                isLiteralComputedName,
+                trackComputedName(accessExpression) {
+                    trackComputedName(accessExpression, context.enclosingDeclaration, context);
+                },
+                getAllAccessorDeclarations: getAllAccessorDeclarationsForDeclaration,
+                requiresAddingImplicitUndefined,
+                isUndefinedIdentifier(node: Identifier) {
+                    return getResolvedSymbol(node) === undefinedSymbol;
```
weswigham, line 6148: `getResolvedSymbol(node)` on arbitrary identifiers is spooky - it's only safe to call on specific kinds of expression-context identifiers (and if you call it on an `undefined` identifier in a type context, it's going to do wonky things, like resolve `undefined` in that scope as a value). `getSymbolAtLocation(node)` is the general-purpose version that figures out the right symbol lookup method from the context. Since you're only using this on expression-identifiers (and not type references or declaration names), maybe renaming to `isUndefinedIdentifierExpression` would be more clear? A `Debug.assert(isExpressionNode(node))` would not be unwarranted.

src/compiler/checker.ts
sandersn: In JS, besides arrows, function expressions and declarations, expando includes classes and empty object literals. I'll take a look at this part.

src/compiler/checker.ts
```diff
@@ -48243,12 +48283,24 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
             return false;
         }
-        const symbol = getSymbolOfDeclaration(declaration);
-        if (!symbol || !(symbol.flags & SymbolFlags.Function)) {
+        let symbol: Symbol;
+        if (isVariableDeclaration(declaration)) {
+            if (declaration.type || !isVarConstLike(declaration)) {
+                return false;
+            }
+            if (!(declaration.initializer && isFunctionExpressionOrArrowFunction(declaration.initializer))) {
```
sandersn, line 48295: might as well combine this `if` with the previous one

src/compiler/checker.ts
```diff
@@ -48243,12 +48283,24 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+                return false;
+            }
+            symbol = getSymbolOfDeclaration(declaration.initializer);
```
sandersn, line 48298: This doesn't seem right, but I'd have to do some investigation to make sure. My memory is that 

```
const f = function g() { g.ex1 = 1 }
f.ex2 = 2
```

has to work in JS, and that `const f` has the symbol for it, not `function g`. But I might be misremembering.
I also thought there were utilities that give the right node to get the symbol from. I'll see what I can turn up.

sandersn (commented):
A couple of comments--I haven't had time to go find the JSDoc utilities I'm thinking of, though.

src/compiler/checker.ts
sandersn: Oh, also *non*-empty object literals, when assigned to class prototypes, so

```
C.prototype = {
  method() { // ...
  }
}
C.prototype.othermethod = function() { }
```

I'm not sure that makes sense here, but I think it does.

src/compiler/checker.ts
sandersn: No, I'm wrong, according to `getExpandoSymbol` the initialiser is the one with all the members.

src/compiler/checker.ts
sandersn, line 48295: If you need to support JS expandos instead of just TS ones, `getDeclaredExpandoInitializer` will check variable declarations for expando initialisers, and then you can afterward require TS expandos to be functions initialising a const variable declaration.

src/compiler/program.ts
dragomirtitian: I think this was mostly for triple slash references, but we have since removed all of those anyway. Not sure what use case there is for this though, since the option is deprecated anyway and I don't think other tools what to implement the `out` behavior.

src/compiler/types.ts
dragomirtitian: You are right, also `trackComputedName` was unused. They will be used when we add the node creation part. Will remove for now.

src/compiler/checker.ts
dragomirtitian: Removed the check as it was not needed

src/compiler/checker.ts
dragomirtitian: You are right, `hasBindableName` does the same thing, the only check this function adds is the guard against type narrowing. 

```ts
       const symbol = getSymbolAtLocation(expression);
        if (!symbol) {
            return false;
        }
        // Ensure not type narrowing
        const declaredType = getTypeOfSymbol(symbol);
        return declaredType === type;        
```


Could we add a flag to `hasBindableName` maybe called `noNarrowing`? Or keep this function and use `hasBindableName` for the first part and just keep the guard against narrowing. I'll implement the second one for ow and we can discuss further.

src/compiler/checker.ts
dragomirtitian: I was going by what happens in `isNullOrUndefined` but that is indeed called in an expression context. Will rename it,

src/compiler/expressionToTypeNode.ts
```diff
@@ -0,0 +1,504 @@
...
+    TypeNode,
+    TypeParameterDeclaration,
+    UnionTypeNode,
+    VariableDeclaration,
+} from "./_namespaces/ts";
+
+export function createSyntacticTypeNodeBuilder(options: CompilerOptions) {
+    const strictNullChecks = getStrictOptionValue(options, "strictNullChecks");
+
+    return {
```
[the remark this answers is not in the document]
dragomirtitian: I have two questions with regard to this:
1. The true/false result is less important, the errors are generated through calling `reportInferenceFallback` on the actual node that causes the issue. Should this be caching the nodes (as there could be multiple) that caused the fallback to inference?

2. Currently we return true/false but the final end state is that this is integrated into the `NodeBuilder` in the checker and that it will actually generate new type nodes instead. I don't think caching the synthetic type nodes is something the current `NodeBuilder` does. Am I wrong about this?

src/compiler/commandLineParser.ts
dragomirtitian: I am unsure how these flags work, however the errors are still emitted in the declaration transform, not the checker. Even if they originate in checker functions, they aren't called in regular type checking, only in during declaration emit. If we set  `affectsSemanticDiagnostics` will it cause the declaration transformer to be called in order to trigger the errors?

src/compiler/expressionToTypeNode.ts
dragomirtitian: This is exactly the reason for the `!isConstContext && !preserveLiterals`. `preserveLiterals` is true for const vars and readonly properties. `isConstContext` is if we are in a const assertion. Both of these may result in preservation of template literal string types, so we don't even try since we can't determine if the type is preserved only based on the expression (non local type information is needed, as shown in your example)

So the only place where we can safely infer `string` is mutable variables and mutable property declarations.

```ts
const a = "a";
export const b = `b${a}` // Can't infer string based on local info
export let m = `b${a}` // Will be string 
export class Foo { 
    b = `b${a}` // Will be string 
    readonly ro = `b${a}` // Can't infer string based on local info
}
```

Template literals in const assertions are not really useful without non local information. I don't think doing something like `` `b${"a"}` `` or `` `b${"a" as "a" | "b"}` ``(which would only need local information to be evaluated) is common enough to be worth testing  for and that is the only  kind of template literal string we could get a template literal string type from.

src/compiler/program.ts
jakebailey: I think @weswigham's question was more about `out`/`outFile` together, not specifically on `out`. `out` has been removed from 5.5 (the option stays behind in types so we can detect and error on it), but `outFile` has not.

src/compiler/checker.ts
jakebailey: > `isOptionalParameter` is also used i the checker and I don't think the syntactic version is adequate i that case.

Yeah, this is the deeper thing I mean; I personally think the checker is way too forgiving here by using the types, when it could use something like the above (plus the IIFE check).

src/compiler/checker.ts
dragomirtitian: Done

src/compiler/checker.ts
weswigham: I'd say keep this function but use `hasBindableName` for the first part. That should at least ensure they stay in-sync.

src/compiler/commandLineParser.ts
weswigham: `affectsSemanticDiagnostics` basically proxies for `affectsDeclarationDiagnostics`, which isn't a standalone thing. The builder logic only copies declaration diagnostics if semantic diagnostics get copied (and `skipLibCheck` is unchanged). See `createBuilderProgramState` in `src/compiler/builder.ts` for the relevant logic. So, name aside, `affectsSemanticDiagnostics` is appropriate here, if ever-so-slightly over-invalidating.

weswigham: Honestly, if we're unsure what level of invalidation gets the desired behavior, a `builder` `verifyTsc`-style unittest is probably not unwarranted.

src/compiler/diagnosticMessages.json
weswigham: That's probably a right enough description, since the upper limit of the desired behavior is deferring all imported names as opaque types/namespaces and deferring all (access, arithmetic, construction, etc) operations on them). Minor nit, ~source~ - just `files`. No need for the jargon that means the same thing. Maybe,
```suggestion
    "Report errors for missing type annotations whose creation requires nonlocal information.": {

```
since `import {N} from "a"; export const a = null as N;` is okay - it's not referring to other files that's conceptually banned by the flag, but leaking data from other files into this one that's banned.

src/compiler/program.ts
weswigham: Yeah, that. It seems like the intent actually has nothing to do with `out`, and more script-mode globally scoped files. If that's the case, we should error on those, and not on `outFile` - amazingly, you can actually use `outFile` with `amd` modules. Eg,
```ts
// @declaration: true
// @strict: true
// @module: amd
// @outFile: out.js
// @filename: input1.ts
export class a {};

// @filename: input2.ts
import {a} from "./input1";
export const b = () => a;
```
and this actually produces declaration files that are widely useful, even if `amd` modules themselves are not.

tests/baselines/reference/tsc/incremental/change-to-type-that-gets-used-as-global-through-export-in-another-file-through-indirect-import.js
```diff
@@ -43,7 +43,7 @@ exitCode:: ExitStatus.Success
 
 
 //// [/src/project/class1.d.ts]
-declare const a = 1;
+declare const a: MagicNumber;
```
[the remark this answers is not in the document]
dragomirtitian: In `shouldPrintWithInitializer` I now test specifically if a type annotation already exists.

```ts
    function shouldPrintWithInitializer(node: Node): node is CanHaveLiteralInitializer & { initializer: Expression; } {
        return canHaveLiteralInitializer(node)
            && !node.type // <- New test 
            && !!node.initializer
            && resolver.isLiteralConstDeclaration(getParseTreeNode(node) as CanHaveLiteralInitializer); // TODO: Make safea
    }
```

src/compiler/checker.ts
dragomirtitian: Was not aware of it. Used it instead. Thank you.

src/compiler/types.ts
```diff
@@ -5394,7 +5404,7 @@ export const enum NodeBuilderFlags {
     AllowUniqueESSymbolType                 = 1 << 20,
     AllowEmptyIndexInfoType                 = 1 << 21,
     /** @internal */ WriteComputedProps      = 1 << 30, // { [E.A]: 1 }
-
+    /** @internal */ NoSyntacticPrinter     = 1 << 31,
```
dragomirtitian, line 5407: When the syntactic node builder will do more than just emit errors (hopefully very very soon) there will not be a need for this as the syntactic node builder should be the one building the types (and if it fails falls back to inference). But for now we need a way to avoid checking the nodes at multiple levels (ex: checking the whole object literal, then also checking the properties individually as they are added to the synthetic object type node)

src/compiler/checker.ts
dragomirtitian: Changed to use `getDeclaredExpandoInitializer` and to return true for JS functions as well.

src/compiler/checker.ts
dragomirtitian: Should I change it to the namespace import?

src/compiler/checker.ts
dragomirtitian: Changed so that these are also executed for JS declarations.

src/compiler/checker.ts
dragomirtitian: So no action needed?

src/compiler/checker.ts
jakebailey: I would, yes, just for consistency (and maybe for the people who aren't afraid to use TS internals)

tests/baselines/reference/isolatedDeclarationErrorsClassesExpressions.errors.txt
```diff
@@ -0,0 +1,42 @@
...
+        return cls;
+    }
+    
+    
+    export class Base {
+    
+    }
+    
+    export class Mix extends id(Base) {
+                             ~~~~~~~~
```
h-joo, line 27: The error here looks like a duplication, do you think this should be an issue? I personally feel that this is not good for UX.

src/compiler/diagnosticMessages.json
fatcerberus: Yeah, I was going to say "nonlocal" is ambiguous - someone could easily interpret "local" as meaning "not from an external dependency" when in fact it's more constrained than that.  The scope should be defined explicitly.

tests/baselines/reference/isolatedDeclarationErrorsClassesExpressions.errors.txt
dragomirtitian: Removed the redundant error

tests/baselines/reference/tsc/incremental/change-to-type-that-gets-used-as-global-through-export-in-another-file-through-indirect-import.js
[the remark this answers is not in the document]
weswigham: Whatever made us ignore the type annotation here in the first place is odd. I can only get it to repro with the same global-type-alias-referencing-a-default-expr-type structure as is in this test.

src/compiler/expressionToTypeNode.ts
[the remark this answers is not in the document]
weswigham: > The true/false result is less important, the errors are generated through calling reportInferenceFallback on the actual node that causes the issue. Should this be caching the nodes (as there could be multiple) that caused the fallback to inference?

You could, but probably don't need to. If you have a cached result, the diagnostics *should* have already been emitted, so you'll just issue duplicates that'll get deduped anyway.

> I don't think caching the synthetic type nodes is something the current NodeBuilder does. Am I wrong about this?

Yes, actually. `typeParameterToName` has caching, since the same parameter often gets serialized many times, though you're right that there's no top-level cache.

src/compiler/commandLineParser.ts
dragomirtitian: Changed to `affectsSemanticDiagnostics: true`. It functions like other diagnostics that come from declaration emit.

tests/baselines/reference/tsc/incremental/change-to-type-that-gets-used-as-global-through-export-in-another-file-through-indirect-import.js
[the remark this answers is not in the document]
dragomirtitian: I did look into it a very long time ago. I was something about the freshness of the literal type. But if the type annotation is there then there is no reason to look further and it is more in line with what an external tool would need  to do as well.

src/compiler/transformers/declarations.ts
```diff
@@ -265,9 +289,190 @@ export function transformDeclarations(context: TransformationContext) {
...
     return transformRoot;
 
+    function reportExpandoFunctionErrors(node: FunctionDeclaration | VariableDeclaration) {
+        resolver.getPropertiesOfContainerFunction(node).forEach(p => {
+            if (isExpandoPropertyDeclaration(p.valueDeclaration)) {
+                const errorTarget = isBinaryExpression(p.valueDeclaration) ?
+                    p.valueDeclaration.left :
+                    p.valueDeclaration;
+
+                context.addDiagnostic(createDiagnosticForNode(
```
weswigham, line 302: Likewise, this should be within `checkFunctionDeclarationDiagnostics` and `checkVariableDeclaration` in the checker instead of during emit, since it doesn't block emit (and isn't reliant on information only available during emit like node-builder triggered errors).

src/compiler/transformers/declarations.ts
```diff
@@ -265,9 +289,190 @@ export function transformDeclarations(context: TransformationContext) {
...
+            | ArrowFunction
+            | MethodDeclaration
+            | ConstructSignatureDeclaration
+            | BindingElement
+            | VariableDeclaration
+            | PropertyDeclaration
+            | ParameterDeclaration
+            | PropertyAssignment
+            | ClassExpression;
+        function getDiagnostic(node: Node) {
```
weswigham, line 343: Like the maps at the bottom of the file, this closure is probably best moved to `src\compiler\transformers\declarations\diagnostics.ts`, so all the error selection logic continues to live in the same place.

tests/baselines/reference/isolatedDeclarationErrorsObjects.errors.txt
```diff
@@ -0,0 +1,161 @@
...
+        a: 1,
+        b: ""
+    }
+    
+    export let oBad = {
+        a: Math.random(),
+           ~~~~~~~~~~~~~
+!!! error TS9013: Expression type can't be inferred with --isolatedDeclarations.
+!!! related TS9027 isolatedDeclarationErrorsObjects.ts:6:12: Add a type annotation to the variable oBad.
+!!! related TS9035 isolatedDeclarationErrorsObjects.ts:7:8: Add a type assertion to this expression to make type type explicit.
```
weswigham, line 30: Suggesting a cast to fix this feels pretty unsafe. `satisfies` would be safer, but because of how that works, you can't pull its' type node, so isn't useful. Like, nothing is going to tell you off for writing `Math.random() as 0`, since casts aren't type annotations - they're just a place the emitter can read a type node from. I don't think we should suggest adding casts as the right fix for arbitrary expressions - annotations or bust, otherwise we're suggestion you make your code unchecked and unsafe.

tests/baselines/reference/isolatedDeclarations.errors.txt
```diff
@@ -0,0 +1,8 @@
+error TS5053: Option 'outFile' cannot be specified with option 'isolatedDeclarations'.
+
+
+!!! error TS5053: Option 'outFile' cannot be specified with option 'isolatedDeclarations'.
```
weswigham, line 4: Without `outFile` set, you can still have global script input files - just don't write any exports and ensure `module` isn't `nodenext` or `node16`. Checking for this option, instead of issuing an error in `checkSourceFileWorker` on global non-ambient files, still feels like checking the wrong thing.

tests/baselines/reference/isolatedDeclarationsAllowJs.errors.txt
```diff
@@ -0,0 +1,12 @@
+error TS5053: Option 'allowJs' cannot be specified with option 'isolatedDeclarations'.
```
weswigham, line 1: 😔We'll definitely have to make a work item to remove this restriction. Other than type nodes needing to be looked up in JSDoc comments (and having some looser expando function inference rules), JS isn't particularly special, nor should it be. TS is, after all, mostly a superset.

weswigham (changes requested):
Largely this seems OK, but there are two larger classes of things to move - a lot of the `context.addDiagnostic` calls being made in the transform don't seem to depend on any emit-related information - all that can should be moved forward into the checker, probably (generally, we prefer being able to emit diagnostics without emitting actual output wherever possible - removing the existing cases in declaration emit that issue errors late is a long-standing backlog item). `isolatedModules` works that way already, anyway. Additionally, a lot of the new declaration diagnostic selection logic should get moved to `src\compiler\transformers\declarations\diagnostics.ts`, since that's where all the other declaration emit diagnostic selection helpers already are. Lastly, there are some other small issues not related to those two larger themes. I've mostly avoided nits for `expressionToTypeNode.ts`, since I know it's where a lot of follow-up changes are occurring.

tests/baselines/reference/tsc/incremental/change-to-type-that-gets-used-as-global-through-export-in-another-file-through-indirect-import.js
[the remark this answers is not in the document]
weswigham: We're probably failing to widen (and strip literal freshness) from `export default expr` types. Or the `typeof` type operation of them does, at least. Unsure. It's *probably* a bug.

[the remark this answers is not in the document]
jakebailey: IIRC we still haven't solved the problem of freshness in general when it comes to isolatedDeclarations, either...

tests/baselines/reference/isolatedDeclarationsAllowJs.errors.txt
dragomirtitian: I think we ca do the work as well. It's just that we have not tested this, and have not looked at this enough to know it works. Adding this now is risky.

tests/baselines/reference/isolatedDeclarationErrorsObjects.errors.txt
dragomirtitian: > Suggesting a cast to fix this feels pretty unsafe. `satisfies` would be safer, but because of how that works, you can't pull its' type node, so isn't useful. Like, nothing is going to tell you off for writing `Math.random() as 0`, since casts aren't type annotations - they're just a place the emitter can read a type node from. I don't think we should suggest adding casts as the right fix for arbitrary expressions - annotations or bust, otherwise we're suggestion you make your code unchecked and unsafe.

Can we suggest `satisfies T as T`?  From my experience having some sort of inline annotation is very useful for large object literals.

weswigham: > Can we suggest satisfies T as T? From my experience having some sort of inline annotation is very useful for large object literals.

That's safe enough that I'd be OK with it - it at least ensures you're not unsafely downcasting, since `Math.random() satisfies 0` errors, so if your expression type changes in some way, the cast will get an error if it becomes unsafe.

tests/baselines/reference/isolatedDeclarations.errors.txt
dragomirtitian: Removed the restriction

src/compiler/transformers/declarations.ts
```diff
@@ -1628,7 +1860,15 @@ export function transformDeclarations(context: TransformationContext) {
...
                         // Rewrite enum values to their constants, if available
-                        const constValue = resolver.getConstantValue(m);
+                        const enumValue = resolver.getEnumMemberValue(m);
+                        const constValue = enumValue?.value;
+                        if (
+                            isolatedDeclarations && m.initializer && (constValue === undefined || enumValue?.hasExternalReferences) &&
+                            // This will be its own compiler error instead, so don't report.
+                            !isComputedPropertyName(m.name)
+                        ) {
+                            context.addDiagnostic(createDiagnosticForNode(m, Diagnostics.Enum_member_initializers_must_be_computable_without_references_to_external_symbols_with_isolatedDeclarations));
```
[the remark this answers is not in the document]
dragomirtitian: We can't really move these as we don't know what enums will be included until the declaration emitter actually pulls in the needed symbols.

src/compiler/transformers/declarations.ts
```diff
@@ -945,6 +1160,19 @@ export function transformDeclarations(context: TransformationContext) {
...
+                    // Classes usually elide properties with computed names that are not of a literal type
+                    // In isolated declarations TSC needs to error on these as we don't know the type in a DTE.
+                    // The
+                    && isClassDeclaration(input.parent)
+                    && isEntityNameExpression(input.name.expression)
+                    // If the symbol is not accessible we get another TS error no need to add to that
+                    && resolver.isEntityNameVisible(input.name.expression, input.parent).accessibility === SymbolAccessibility.Accessible
+                    && !resolver.isNonNarrowedBindableName(input.name)
+                ) {
+                    context.addDiagnostic(createDiagnosticForNode(input, Diagnostics.Computed_properties_must_be_number_or_string_literals_variables_or_dotted_expressions_with_isolatedDeclarations));
```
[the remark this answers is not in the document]
dragomirtitian: Same, for these, we don't know they are needed for declaration emit until the mit actually happens

src/compiler/transformers/declarations.ts
dragomirtitian: Moved

src/compiler/transformers/declarations.ts
```diff
@@ -1903,3 +2143,38 @@ function isProcessedComponent(node: Node): node is ProcessedComponent {
     }
     return false;
 }
+
+const relatedSuggestionByDeclarationKind = {
```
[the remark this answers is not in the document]
dragomirtitian: Moved

src/compiler/program.ts
dragomirtitian: Removed the restriction.

tests/baselines/reference/isolatedDeclarationErrorsObjects.errors.txt
dragomirtitian: Changed suggestion. @h-joo Could you please change the code fix to also do this.

src/compiler/diagnosticMessages.json
dragomirtitian: Changed to `Require sufficient annotation on exports so other tools can trivially generate declaration files.` May still change later

tests/baselines/reference/tsc/incremental/change-to-type-that-gets-used-as-global-through-export-in-another-file-through-indirect-import.js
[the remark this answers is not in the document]
dragomirtitian: In what sense? `ensureNoInitializer` will emit an error if the initializer is not a primitive. It is true that if you add the annotation, the meaning will not be exactly the same though.

```ts
const Original = 1;
const Alias = Original; // adding :1 here will chnage the type of v
let v = Alias; // v is number if Alias is not annotated, or 1 if const Alias:1 
```

[Playground Link](https://www.typescriptlang.org/play?#code/MYewdgzgLgBA8gJwJYHMlgIYBsYF4YCMA3ALABQoksAglkhhHvMmplkTAPScwYAmfdChgAuAjAAWAUwRSYAdyRYcwCZhRyo0mFACeABzkgAZjABu5LFNhmmtehA7dzMJIzABXALYAjGa9N7Bld3EFgMMEiwjCgpPgAaGBAEQgCYSmgYIIgxGCA)

src/compiler/checker.ts
```diff
@@ -8177,6 +8207,10 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             let returnTypeNode: TypeNode | undefined;
             const returnType = getReturnTypeOfSignature(signature);
             if (returnType && !(suppressAny && isTypeAny(returnType))) {
+                if (signature.declaration && !(context.flags & NodeBuilderFlags.NoSyntacticPrinter)) {
+                    syntacticNodeBuilder.serializeReturnTypeForSignature(signature.declaration, context);
+                }
+                context.flags |= NodeBuilderFlags.NoSyntacticPrinter;
```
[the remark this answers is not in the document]
dragomirtitian: It is unapplied a bit lower `context.flags = flags;`. Flags were already being saved and restored and there are no other calls between.

src/compiler/checker.ts
```diff
@@ -8155,6 +8181,10 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             const decl = declaration ?? symbol.valueDeclaration ?? symbol.declarations?.[0];
             const expr = decl && isDeclarationWithPossibleInnerTypeNodeReuse(decl) ? getPossibleTypeNodeReuseExpression(decl) : undefined;
 
+            if (decl && hasInferredType(decl) && !(context.flags & NodeBuilderFlags.NoSyntacticPrinter)) {
+                syntacticNodeBuilder.serializeTypeOfDeclaration(decl, context);
+            }
+            context.flags |= NodeBuilderFlags.NoSyntacticPrinter;
```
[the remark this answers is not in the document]
dragomirtitian: It is unapplied a bit lower `context.flags = oldFlags;`. Flags were already being saved and restored and there are no other calls between.

tests/baselines/reference/tsc/incremental/change-to-type-that-gets-used-as-global-through-export-in-another-file-through-indirect-import.js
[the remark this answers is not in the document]
jakebailey: I was referring to https://github.com/microsoft/TypeScript/pull/55860 https://github.com/microsoft/TypeScript/issues/55439 etc.

Also, Symbol freshness; I tried my hand at fixing that a bit ago and was not successful.

src/compiler/transformers/declarations.ts
```diff
@@ -945,6 +1160,19 @@ export function transformDeclarations(context: TransformationContext) {
         if (isDeclaration(input)) {
             if (isDeclarationAndNotVisible(input)) return;
             if (hasDynamicName(input) && !resolver.isLateBound(getParseTreeNode(input) as Declaration)) {
+                if (
+                    isolatedDeclarations
+                    // Classes usually elide properties with computed names that are not of a literal type
+                    // In isolated declarations TSC needs to error on these as we don't know the type in a DTE.
+                    // The
```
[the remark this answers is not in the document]
dragomirtitian: Removed.

src/compiler/checker.ts
```diff
@@ -6018,6 +6022,19 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+        /**
+         * Same as expressionOrTypeToTypeNodeHelper, but also checks if the expression can be syntactically typed.
+         */
+        function expressionOrTypeToTypeNodeHelper(context: NodeBuilderContext, expr: Expression | JsxAttributeValue | undefined, type: Type, addUndefined?: boolean) {
+            const oldFlags = context.flags;
+            if (expr && !(context.flags & NodeBuilderFlags.NoSyntacticPrinter)) {
+                syntacticNodeBuilder.serializeTypeOfExpression(expr, context, addUndefined);
+            }
+            context.flags |= NodeBuilderFlags.NoSyntacticPrinter;
+            const result = expressionOrTypeToTypeNode(context, expr, type, addUndefined);
```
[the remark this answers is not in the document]
dragomirtitian: `expressionOrTypeToTypeNodeHelper` (after rename)  may call `typeToTypeNodeHelper` which in turn will call `serializeTypeForDeclaration` and `serializeReturnTypeForSignature` and these would perform the check again. I can refactor those into `serializeTypeForDeclarationHelper` and `serializeReturnTypeForSignatureHelper`  which do not perform the check and `serializeTypeForDeclaration` and `serializeReturnTypeForSignature`  which would only be called from the top level. This would ensure the code does not check multiple times the same expression, and would remove the need for the `NoSyntacticPrinter` flag. 

The flag will disappear in the next PRs. It's definitely not something that will remain. When the syntactic builder actually builds the types and falls back to the checked builder when it can't the problem of duplicate checks will go away.

Should I do the  `serializeTypeForDeclarationHelper` and `serializeReturnTypeForSignatureHelper` refactor for now? Or ca we live with this until after the beta?

dragomirtitian:
I think I addressed most comments. The ones that remain I don't feel are actionable:

1. Caching the result of the syntactic check. I don't think I can address this until tomorrow. I think it would be better to think about this once the syntactic builder actually produces nodes and see about caching those. [link](https://github.com/microsoft/TypeScript/pull/58201#discussion_r1566231093)

2. `shouldPrintWithInitializer` change - I can revert this. The old version sometimes returned true for constants that are annotated. This should not happen, am external tool would use the explicit annotation. This might be more in the emit alignment category, so I can open a PR with it after the beta. [link](https://github.com/microsoft/TypeScript/pull/58201#discussion_r1567835558)

3. remove `NoSyntacticPrinter` - Any changes I make around this would be temporary as this flag is not needed in the final version. I did make a suggestion that I could implement to make it better for now. [link](https://github.com/microsoft/TypeScript/pull/58201#discussion_r1570995658)

weswigham:
> Caching the result of the syntactic check. I don't think I can address this until tomorrow. I think it would be better to think about this once the syntactic builder actually produces nodes and see about caching those. https://github.com/microsoft/TypeScript/pull/58201#discussion_r1566231093

It's not critical, it'll probably just be good for perf in the normal command-line compile case.

> shouldPrintWithInitializer change - I can revert this. The old version sometimes returned true for constants that are annotated. This should not happen, am external tool would use the explicit annotation. This might be more in the emit alignment category, so I can open a PR with it after the beta. https://github.com/microsoft/TypeScript/pull/58201#discussion_r1567835558

I do think the existing emit is odd, but this is definitely something we should consider separately, as potentially weirdly breaking and not scoped to isolated declarations. Let's split that into a separate PR.

> remove NoSyntacticPrinter - Any changes I make around this would be temporary as this flag is not needed in the final version. I did make a suggestion that I could implement to make it better for now. https://github.com/microsoft/TypeScript/pull/58201#discussion_r1570995658

From here, for now, I'd say just apply [this change](https://github.com/weswigham/TypeScript/commit/f05f91c4a8c4a29d6676df0130a9de4d07aa34ee.patch), and we can revisit as functionality changes. That's all I was getting at before.

src/compiler/program.ts
```diff
@@ -4229,6 +4229,12 @@ export function createProgram(rootNamesOrOptions: readonly string[] | CreateProg
             }
         }
 
+        if (options.isolatedDeclarations) {
```
sheetalkamat, line 4232: Does `isolatedDeclarations` means its "declaration: true` if yes we need to update `createComputedCompilerOptions` to handle that... otherwise we need error here if this is specified with declarations

dragomirtitian: Ok. Will add the error.

src/compiler/program.ts
```diff
@@ -4229,6 +4229,15 @@ export function createProgram(rootNamesOrOptions: readonly string[] | CreateProg
             }
         }
 
+        if (options.isolatedDeclarations) {
+            if (getAllowJSCompilerOption(options)) {
+                createDiagnosticForOptionName(Diagnostics.Option_0_cannot_be_specified_with_option_1, "allowJs", "isolatedDeclarations");
+            }
+            if (!options.declaration) {
```
weswigham, line 4236:
```suggestion
            if (!getEmitDeclarations(options)) {

```
since `declaration` is implied by some other options.

src/compiler/program.ts
```diff
@@ -4233,6 +4233,9 @@ export function createProgram(rootNamesOrOptions: readonly string[] | CreateProg
             if (getAllowJSCompilerOption(options)) {
                 createDiagnosticForOptionName(Diagnostics.Option_0_cannot_be_specified_with_option_1, "allowJs", "isolatedDeclarations");
             }
+            if (!options.declaration) {
+                createDiagnosticForOptionName(Diagnostics.Option_0_cannot_be_specified_without_specifying_option_1, "isolatedDeclarations", "declaration");
```
sheetalkamat, line 4237:
```suggestion
                createDiagnosticForOptionName(Diagnostics.Option_0_cannot_be_specified_without_specifying_option_1_or_option_2, "isolatedDeclarations", "declaration", "composite");
```

src/compiler/checker.ts
```diff
@@ -6114,6 +6131,15 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
                 bundled: !!compilerOptions.outFile && !!enclosingDeclaration && isExternalOrCommonJsModule(getSourceFileOfNode(enclosingDeclaration)),
+                isEntityNameVisible: (entityName, shouldComputeAliasToMakeVisible) => isEntityNameVisible(entityName, context.enclosingDeclaration!, shouldComputeAliasToMakeVisible),
+                isExpandoFunctionDeclaration,
+                isNonNarrowedBindableName,
+                getAllAccessorDeclarations: getAllAccessorDeclarationsForDeclaration,
+                requiresAddingImplicitUndefined,
+                isUndefinedIdentifierExpression(node: Identifier) {
+                    Debug.assert(isExpressionNode(node));
+                    return getSymbolAtLocation(node) === undefinedSymbol;
+                },
```
jakebailey, line 6142: Every time we clone the context, we're copying these functions over and over again even though they never change. I'm guessing this is why memory usage in the benchmarks has increased. We should split these off onto another object. Probably even, construct the object (a host, I guess?) at the top of the checker.

This refactor also exposes what looks like a bug; `isEntityNameVisible` closes over `context`, but `context` can be cloned but `isEntityNameVisible` will still refer to the old object. We screw around with `context.enclosingDeclaration` all over the place, so this could potentially lead to a bug.

dragomirtitian: I was planning on doing exactly this in my follow on PRs. I can try to move it now if this is critical before the beta.

jakebailey: I have a change locally I could push up, nearly finished.

dragomirtitian: That would be great. Thank you.

jakebailey:
I left some unused code behind, sorry :(

Shouldn't affect benchmarks, so

jakebailey:

weswigham (approved):

jakebailey (approved):
:tada:

We think the memory increase is just the increased cache size from `NoSyntacticPrinter`, but that can be fixed in a follow-up (and the difference is small).

PR #58203, opened 2024-04-15
Allow signature help argument count to be equal to argument index

Andarist:
fixes the issue reported here https://github.com/microsoft/TypeScript/issues/58191#issuecomment-2054196481
a regression from https://github.com/microsoft/TypeScript/pull/56372 , not fixed by https://github.com/microsoft/TypeScript/pull/57637


src/services/signatureHelp.ts
```diff
@@ -287,9 +287,6 @@ function getArgumentOrParameterListInfo(node: Node, position: number, sourceFile
     const { list, argumentIndex } = info;
 
     const argumentCount = getArgumentCount(checker, list);
-    if (argumentIndex !== 0) {
-        Debug.assertLessThan(argumentIndex, argumentCount);
-    }
```
Andarist, line 292: While it might look weird that this allows what usually isn't allowed - it all depends on the definition of those values. I think it makes sense to treat them as follows:
- argument index - a min argument index at the position
- argument count - a minimum count of known arguments

A spread element doesn't always increase such an argument count but the argument index at its positions gets increased

DanielRosenwasser: But what should an editor do when the argument index exceeds the count? The index has to be capped, doesn't it?

jakebailey: Out of bounds indexes are just ignored; on Python I did this intentionally, too (previously used a negative number until some rust-based LSP clients started getting picky). See: https://github.com/microsoft/language-server-protocol/issues/1271

But, I'm more confused by how a spread element would change the number here; we're definitely indexing them correctly left to right, correct? Like if we have `(x, {y, z})`, we don't go out of bounds on `z` because it's "at index 2" but the number of real params is 2? I assume not.

Andarist: > But what should an editor do when the argument index exceeds the count? The index has to be capped, doesn't it?

The added baseline here shows how we can still highlight the parameter in the signature - at least in this case 😉 

> But, I'm more confused by how a spread element would change the number here

```ts
function test(...args: any[]) {}

const twoNums = [2, 3] as const
test(1, ...twoNums)
```

In the situation above the rest adds 2 to the count (and not 1). See the PR that changed this: https://github.com/microsoft/TypeScript/pull/56372

fatcerberus: Nitpick: That's a spread, not a rest.  Rest is what you do when destructuring, spread is what you do at construction ("rest" doesn't make sense in that context because you can have multiple).

jakebailey:
@typescript-bot pack this

jakebailey (approved):

PR #58204, opened 2024-04-15
Fix Debug Fail Go-to-def on JS class

armanio123:
Fixes https://github.com/microsoft/TypeScript/issues/57924#issuecomment-2016963512

### DIFF
--- a/src/services/goToDefinition.ts
+++ b/src/services/goToDefinition.ts
@@ -4,6 +4,7 @@ import {
     AssignmentOperatorToken,
     CallLikeExpression,
     canHaveSymbol,
+    ClassLikeDeclaration,
     concatenate,
     createTextSpan,
     createTextSpanFromBounds,
@@ -15,6 +16,7 @@ import {
     DefinitionInfoAndBoundSpan,
     emptyArray,
     every,
+    Expression,
     FileReference,
     filter,
     find,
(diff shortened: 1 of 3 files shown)


src/services/goToDefinition.ts
```diff
@@ -589,7 +593,7 @@ function getDefinitionFromSymbol(typeChecker: TypeChecker, symbol: Symbol, node:
         // Applicable only if we are in a new expression, or we are on a constructor declaration
         // and in either case the symbol has a construct signature definition, i.e. class
         if (symbol.flags & SymbolFlags.Class && !(symbol.flags & (SymbolFlags.Function | SymbolFlags.Variable)) && (isNewExpressionTarget(node) || node.kind === SyntaxKind.ConstructorKeyword)) {
-            const cls = find(filteredDeclarations, isClassLike) || Debug.fail("Expected declaration to have at least one class-like declaration");
+            const cls = find(filteredDeclarations, isClassLike) || getRightMostAssignedExpression(find(filteredDeclarations, isPropertyAccessExpression)?.parent as Expression) as ClassLikeDeclaration || Debug.fail("Expected declaration to have at least one class-like declaration");
```
Andarist, line 596: note that this change "competes" with https://github.com/microsoft/TypeScript/pull/57628

gabritto:
@sandersn do you have an opinion on whether we should accept this PR or https://github.com/microsoft/TypeScript/pull/57628?

src/services/goToDefinition.ts
armanio123: Thanks for noticing. I'll close this as I think yours makes more sense.

PR #42211, opened 2021-01-04
Fix leaked alias symbol independent type params in printback

weswigham:
When we measure the variance of an alias, an `Independent` type parameter is an unused type parameter - for whatever reason, it has no effect on the comparability of the resulting type. What this often means is that that type parameter also does not play into the construction of the the target type _at all_. That, in turn, means that due to interning, you can have multiple seemingly unique references to a type, which all point at a single underlying type - usually the one first manufactured within the type alias declaration itself. That underlying type, in turn, has an alias from when it was first created, and that alias's type arguments may contain type parameters only available within the scope of the original alias declaration. Normally, this would disqualify use of the alias as a reference external to that declaration, but we failed to detect such a case as our assumption is that if we are doing printback using a type parameter, that it will be made available in scope (otherwise, how did we get it?). Since this type parameter comes from far afield, we can't safely do printback using it, _however_, in this fix, rather than issuing an error in such a circumstance (which would be unfortunate and hard to fix for users), I detect when this situation arises by checking the variance and visibility of the alias symbol type arguments, and substitute an unreachable `Independent` type argument with `unknown`, as it will not affect the resulting type's shape, thus creating a functioning declaration file.

This approach required checking the variance of some type aliases where we previously had not bothered (usually simply because the test in question had not been written to require it), and in turn exposed a crash-y problem we had calculating variance while handling recursive references to unions and intersections within type aliases (and recursive normalization thereof - see the types in `recursivelyExpandingUnionNoStackoverflow` as an example of a type which we crashed when fetching the variance on it). To fix _that_, #37910 (which just got merged into `master`) mostly handles it, however the size limit it originally imposed has to be removed, as we need the deeply nested type guarding part of `recursiveTypeRelatedTo` even for the smallest unions, as recursive type references can create situations where these can infinitely expand upon normalization. In addition, I've added an identity for unions and intersections with aliases into `getRecursionIdentity` to improve how quickly we recognize these structures as deeply nested (the time it took to calculate variances in `recursiveTypeReferences2` proved that needed).

TL;DR: We now print
```ts
declare const nested: PublicWrap<unknown, {}>;
```
rather than the erroneous
```ts
declare const nested: PublicWrap<T, {}>;
```
where `PublicWrap`'s first type parameter is `Independent`.

Fixes #42079
Fixes #44727

PR #58208, opened 2024-04-15
Add test for fixed declaration emit type parameter leak

### LINKED ISSUE #44727: Incorrect transpilation of generic type parameter
# Bug Report

I have a type with an unused generic type parameter.\*
This type annotates an **optional parameter property** in the constructor of a class.
Starting with Version 4.2.X of TypeScript, this class is incorrectly transpiled, so that TS code using the transpiled sources (.d.ts) does not transpile anymore.
The error was:
```
error TS2304: Cannot find name 'T'
```

No wonder it cannot be found, it was not in the original code (please see the minimal example below).

During my investigation I found that this issue occurs, when these prerequisites are met:
1. The given type has an unused generic parameter.
1. The given type represents a union type.
1. You have a class with a constructor argument using the given type.
1. The constructor argument is a parameter property.
1. The constructor argument is optional.

*\* (Tried something. Didn't work. Postponed it, but left the parameter for later use.)*
### 🔎 Search Terms

generics, transpilation, missing parameter

- This changed between versions 4.1.X and 4.2.X

### ⏯ Playground Link

[Playground link with relevant code](https://www.typescriptlang.org/play?target=5#code/C4TwDgpgBAQgTgewNYQHYBVwQDzoHxQC8UA5AIYlQA+pARiQNwCwAUKwMYA2ZAzj1AFkQAYW58oAb1ZQZUdglQ9gcAK7tgCOAAo4EMgBMFnEFDJwA5gH4AXLEQoMWbGVQg8ASkkBfVj5ZA)

### 💻 Code

Before transpilation:
```ts
type BrokenType<T> = 'a' | 'b';

class MyClass {
    constructor(readonly arg?: BrokenType<any>) {}
}
```

After transpilation
```ts
declare type BrokenType<T> = 'a' | 'b';
declare class MyClass {
    readonly arg?: BrokenType<T> | undefined;
    constructor(arg?: BrokenType<T> | undefined); // <----- T should be any
}
```

### 🙁 Actual behavior

The code is transpiled to contain the generic parameter name, instead of the actual type that was passed as argument.

### 🙂 Expected behavior

I would expect the result to contain the original type as specified in the original code. The generic parameter is not known when consuming the type. This worked before version 4.2.X.

jakebailey:
Closes #44727, which was actually fixed by #57772.

### DIFF
--- a/tests/baselines/reference/declarationEmitNoTypeParamLeak1.js
+++ b/tests/baselines/reference/declarationEmitNoTypeParamLeak1.js
@@ -1,0 +1,25 @@
+//// [tests/cases/compiler/declarationEmitNoTypeParamLeak1.ts] ////
+
+//// [declarationEmitNoTypeParamLeak1.ts]
+type BrokenType<T> = 'a' | 'b';
+
+class MyClass {
+    constructor(readonly arg?: BrokenType<any>) {}
+}
+
+//// [declarationEmitNoTypeParamLeak1.js]
+"use strict";
+var MyClass = /** @class */ (function () {
+    function MyClass(arg) {
+        this.arg = arg;
+    }
+    return MyClass;
+}());
+
+
+//// [declarationEmitNoTypeParamLeak1.d.ts]
+type BrokenType<T> = 'a' | 'b';
+declare class MyClass {
+    readonly arg?: BrokenType<any> | undefined;
+    constructor(arg?: BrokenType<any> | undefined);
+}
--- a/tests/baselines/reference/declarationEmitNoTypeParamLeak1.symbols
+++ b/tests/baselines/reference/declarationEmitNoTypeParamLeak1.symbols
@@ -1,0 +1,14 @@
+//// [tests/cases/compiler/declarationEmitNoTypeParamLeak1.ts] ////
+
+=== declarationEmitNoTypeParamLeak1.ts ===
+type BrokenType<T> = 'a' | 'b';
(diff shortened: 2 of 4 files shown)


jakebailey:
That or it's _not_ fixed, if this type were to be inferred somehow?

weswigham:
> That or it's not fixed, if this type were to be inferred somehow?

If you add a `, prop = [null as any as BrokenType<any>][0]` does it repro? Looks like it does on the [playground](https://www.typescriptlang.org/play?ts=5.5.0-dev.20240415#code/C4TwDgpgBAQgTgewNYQHYBVwQDzoHxQC8UA5AIYlQA+pARiQNwCwAUKwMYA2ZAzj1AFkQAYW58oAb1ZQZUdglQ9gcAK7tgCOAAo4EMgBMFnEFDJwA5gH4AXLEQoMWbGVQg8AGihhEYIlADaqCqcnKb8Lia8dshomJDOrngAuv4ADEkAlJIAvqzZQA).

jakebailey:
Sure does, so, not fixed. Oh well!

weswigham:
If you want a fix, the first block in the [old related PR](https://github.com/microsoft/TypeScript/pull/42211/files#diff-d9ab6589e714c71e657f601cf30ff51dfc607fc98419bf72e04f6b0fa92cc4b8R4539) (replacing `Independent` type parameters in the emit with `unknown`) is probably still the way to go; I don't think the emit's changed all that much.

PR #58212, opened 2024-04-16
Logarithmish baseline counts

RyanCavanaugh:
Changes the baseline perf counters to round to the nearest 10/25/50/100-interval; suppresses numbers which are below the reporting threshold.


src/harness/harnessIO.ts
```diff
@@ -786,21 +786,20 @@ export namespace Compiler {
...
+
+                if (perfStats.some(([, threshold, , postValue]) => postValue >= threshold)) {
                     perfLines.push(`=== Performance Stats ===`);
-                    for (const [name, _, rounding, preValue, postValue] of perfStats) {
-                        const preDisplay = valueToString(preValue, rounding);
-                        if (preDisplay !== "0") {
-                            perfLines.push(`${name}: ${preDisplay} / ${valueToString(postValue, rounding)} (nearest ${rounding})`);
+                    for (const [name, threshold, preValue, postValue] of perfStats) {
+                        if (postValue >= threshold) {
+                            perfLines.push(`${name}: ${valueToString(preValue)} / ${valueToString(postValue)}`);
```
jakebailey, line 802: This is a nit but could you (while here) change this to `->` instead of `/`? I had to explain to a few people that this was a before/after and not a percentage or something.

DanielRosenwasser: Thought - if the post-value is identical, maybe don't show it at all. If it's different maybe just write out `(post-check: #####)`

RyanCavanaugh: Done and done

tests/baselines/reference/recursiveConditionalCrash3.types
```diff
@@ -1,10 +1,9 @@
 //// [tests/cases/compiler/recursiveConditionalCrash3.ts] ////
 
 === Performance Stats ===
-Assignability cache: 100 / 3,000 (nearest 100)
-Type Count: 500 / 18,400 (nearest 100)
-Instantiation count: 1,000 / 70,500 (nearest 500)
-Symbol count: 26,000 / 28,000 (nearest 500)
+Assignability cache: 100 -> 2,500
+Type Count: 500 -> 25,000
```
RyanCavanaugh, line 5: 👀

jakebailey (approved):

RyanCavanaugh:
fyi @ahejlsberg

Issue #58205, opened 2024-04-15
Debug Failure. False expression: Token end is child end


Andarist:
### 🔎 Search Terms

completions formatting class member snippets constructor

### 🕗 Version & Regression Information

- This is a crash

### ⏯ Playground Link

N/A

### 💻 Code

```ts
/// <reference path="fourslash.ts" />

// @strict: true

// @filename: KlassConstructor.ts

//// type GenericConstructor<T> = new (...args: any[]) => T;
//// export type KlassConstructor<Cls extends GenericConstructor<any>> =
////   GenericConstructor<InstanceType<Cls>> & { [k in keyof Cls]: Cls[k] };

// @filename: ElementNode.ts
//// import { KlassConstructor } from "./KlassConstructor";
////
//// export type NodeKey = string;
////
//// export class ElementNode {
////   ["constructor"]!: KlassConstructor<typeof ElementNode>;
////   constructor(key?: NodeKey) {}
//// }

// @filename: CollapsibleContainerNode.ts
//// import { ElementNode, NodeKey } from "./ElementNode";
////
//// export class CollapsibleContainerNode extends ElementNode {
////   __open: boolean;
////
////   constructor(open: boolean, key?: NodeKey) {
////     super(key);
////     this.__open = open;
////   }
////   /*1*/
//// }

format.setFormatOptions({
  insertSpaceAfterConstructor: false,
});

verify.completions({
    marker: "1",
    preferences: {
        includeCompletionsWithClassMemberSnippets: true,
        includeCompletionsWithInsertText: true,
    },
    // TODO: tweak this line after fixing the issue
    includes: ["constructor"]
});
```

### 🙁 Actual behavior

it crashes

### 🙂 Expected behavior

it shouldn't crash

### Additional information about the issue

repro based on https://github.com/microsoft/TypeScript/issues/58191#issuecomment-2054196459

PR #58216, opened 2024-04-16
Fixed crash when cross-file reusing nodes for class member snippet completions

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/58205


DanielRosenwasser:
Uhhhhh, kinda nervous about walking the tree on every literal.

DanielRosenwasser:
I think the right thing to do is do a clone of the node from wherever this comes from. Especially since there are other bugs that often come up like when a node is reused within the tree and the formatter tries to overwrite it.

gabritto:
> I think the right thing to do is do a clone of the node from wherever this comes from. Especially since there are other bugs that often come up like when a node is reused within the tree and the formatter tries to overwrite it.

Yep, if the underlying problem is that node re-use messes up the formatter calls in the services layer, we seem to have many occurrences of that. @armanio123 is working on a fix for https://github.com/microsoft/TypeScript/issues/57924#issuecomment-2016963508, which is an instance of this, although that crash is caused by re-use of a question token instead of a literal.

gabritto:

Andarist:
So I thought that the node reuse would be desired here. If I understand correctly you are saying that it should never happen here, right? Cloning the original node would likely make it synthesized or something and that would prevent reuse altogether.

gabritto:
> So I thought that the node reuse would be desired here. If I understand correctly you are saying that it should never happen here, right? Cloning the original node would likely make it synthesized or something and that would prevent reuse altogether.

The current state of the formatter is such that we should never pass it an AST that has node reuse, as that will cause a crash. Other parts of the compiler don't suffer from this problem. I think the emitter doesn't care about node reuse, for the most part.

I *think* the problem is that the formatter was written to handle actual parse trees, coming from actual files, but we've been using it to format trees we generate for completions.

DanielRosenwasser:
> I think the problem is that the formatter was written to handle actual parse trees, coming from actual files, but we've been using it to format trees we generate for completions.

Yeah, and I believe the formatter actually *mutates* the parse trees, so that complicates things too.

Andarist:
I didn't see (yet?) any mutation like that happening. `getSynthesizedDeepCloneWorker` (that gets already called!) has some specific logic to handle string literals:
https://github.com/microsoft/TypeScript/blob/21b5c9624c0765e6baea1f21a26531aac4824a41/src/services/utilities.ts#L3205-L3211

`createStringLiteralFromNode` sets `.textSourceNode` on the created node:
https://github.com/microsoft/TypeScript/blob/21b5c9624c0765e6baea1f21a26531aac4824a41/src/compiler/factory/nodeFactory.ts#L1266

and that gets used here by the emitter in `getLiteralTextOfNode`:
https://github.com/microsoft/TypeScript/blob/21b5c9624c0765e6baea1f21a26531aac4824a41/src/compiler/emitter.ts#L5150

Given all of that It still feels like using the correct source file is the easiest here - but I can change the location where it's looked for because only the code that uses `.textSourceNode` has to look for that original source file. I pushed out this change - please take a look and let me know what do you think about it. If you like this approach I'll look for more locations using `.textSourceNode` because I'm pretty sure that this doesn't cover everything right now.

gabritto (approved):

gabritto:

gabritto:
Some notes for the future:
1. This is not a bug related to node reuse. The bug is in how the emitter emits literals. A string literal node can be created from another node via `createStringLiteralFromNode`, and this newly string literal node will have a `textSourceNode` that points to the node it was created from. When the emitter tries to emit this string literal, it will attempt to get the text from this `textSourceNode`, because we prefer to get the original text as it was written, rather than the normalized version we store in the node's `text` property. But when it tries to get the text from this text source node, the emitter assumes the text source node's file is the current file, which is not necessarily true, and thus it gets the text from a different file, so the text is wrong.

2. In this PR, we'll only ever look up a node's source file when we are emitting a string literal that has the `textSourceNode` property, i.e., nodes created by transformers or services, so that doesn't seem very impactful for performance in general.

armanio123 (approved):

Issue #58215, opened 2024-04-16
Assigning a @type to an async function emits an error


jespertheend:
### 🔎 Search Terms

ts1065 jsdoc async return promise

### 🕗 Version & Regression Information

- This changed between versions `5.1.6` and `5.2.2`

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.2.2&filetype=js#code/PQKhAIAEBcE8AcCmATRAzcBvAFASnALwB84ACgE4D2AtgJYDOiAPAG6W3IkA+4bHAvuACC9WADsAxgDEx4EMABQC0BBgJEWEeOljB8hRMpj60cNVhbJMwuACGoyeDyESmfkA

### 💻 Code

JSDoc version:
```js
/** @typedef {() => Promise<void> | void} AsyncFn */

/** @type {AsyncFn} */
const myAsyncFn = async () => {}
```

TypeScript equivalent:
```ts
type AsyncFn = () => Promise<void> | void;

const myAsyncFn : AsyncFn = async () => {}
```

### 🙁 Actual behavior

JSDoc version emits an error, TypeScript equivalent does not.

### 🙂 Expected behavior

Neither should emit an error.

### Additional information about the issue

I'm fairly sure this is related to a change from a while back where a `new Promise` needs to be explicitly marked as returning `Promise<void>`, which was intentional. But I can't seem to find the related issue anymore so I'm not sure if this is intentional as well.

Either way, the fact that this error is only emitted in JavaScript files seems to suggest that it's a bug.

Andarist:
The problem is that `@type` can be a source of an effective return type node for an arrow function like this. I don't know all the rules but it's kinda like a mix of a contextual type and an explicit return type annotation. And an explicit return type annotation has to refer to `Promise`.

A quick experiment to ignore `@type` tags for this purpose here shows that it's not really a good idea:
<details>
<summary>git diff</summary>

```diff
diff --git a/src/compiler/checker.ts b/src/compiler/checker.ts
index d0a2843989..13407fdb86 100644
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -39900,7 +39900,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         function checkSignatureDeclarationDiagnostics() {
             checkCollisionWithArgumentsInGeneratedCode(node);
 
-            let returnTypeNode = getEffectiveReturnTypeNode(node);
[...]

PR #58217, opened 2024-04-16
Add an extra test case for an error on async function typed using `@type`

Andarist:
I broke this [here](https://github.com/microsoft/TypeScript/issues/58215#issuecomment-2059930788) so I'm somewhat sure that there is no other test case like this right now


jakebailey (approved):

DanielRosenwasser:
Thanks!

PR #58220, opened 2024-04-17
Improve diagnostics deduplication

### LINKED ISSUE #58207: Duplicate diagnostics
### 🔎 Search Terms

duplicate diagnostics, duplicate errors

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.5.0-dev.20240415#code/JYOwLgpgTgZghgYwgAgMIHsC2AHANsBOSAEwBUBPbCAITgGcUBvAKGTeQG06AuZOsKKADmAXV4BBWgwDczAL7NQkWIhQYc+QiQpUAItGAA3CMWQt2nHnwHCxycfsHHishUujwk9qU1bs4vPyCIEKuiuAeqvaORiZmfmwARoE2IWEIuPR0yD7xFjDo6ACEvOp4BEQmOjT0EOnoIPzIAB7IALzIGVnIMc7IEM2QIMTZuebsBcWlWOVaVZQQvSZhuBBgyOiJAFa8jMgAzCnBQshy7WYHvABEYAAWUBAQV6eym1sczSLSQA

### 💻 Code

```ts
interface ComplicatedTypeBase {
    [s: string]: ABase;
}
interface ComplicatedTypeDerived {
    [s: string]: ADerived;
}
interface ABase {
    a: string;
}
interface ADerived {
    b: string;
}
class Base {
    foo!: ComplicatedTypeBase;
}
const x = class Derived extends Base {
    foo!: ComplicatedTypeDerived;
}
let obj: { 3: string } = { 3: "three" };
obj[x];
```

### 🙁 Actual behavior

Two instances of the same error on line `foo!: ComplicatedTypeDerived;`:
- `Property 'foo' in type 'Derived' is not assignable to the same property in base type 'Base'.
  Type 'ComplicatedTypeDerived' is not assignable to type 'ComplicatedTypeBase'.(2416)`
- `Property 'foo' in type 'Derived' is not assignable to the same property in base type 'Base'.
  Type 'ComplicatedTypeDerived' is not assignable to type 'ComplicatedTypeBase'.
    'string' index signatures are incompatible.
      Property 'a' is missing in type 'ADerived' but required in type 'ABase'.(2416)`

### 🙂 Expected behavior

A single instance of the error

_No response_

gabritto:
Fixes #58207.

Before, we only considered two diagnostics the same if everything about them was equal, including the message chain and related information. With this PR, we now consider diagnostics at the same location and *with the same head message* as being the same, regardless of the rest of the message chain and related information.
That is necessary because, as the new test shows, we can generate the "same" diagnostic at different moments in type checking, and they will have different elaboration (different message chains and related information).

This PR also makes it so deduplication keeps the diagnostic with more elaboration between two diagnostics considered to be equal.

Note that this doesn't completely solve the issue, because not all diagnostics that should be considered equal have the same code and head message.
Edit: follow-up PR #58318 fixes this.

### DIFF
--- a/src/compiler/utilitiesPublic.ts
+++ b/src/compiler/utilitiesPublic.ts
@@ -49,6 +49,7 @@ import {
     Decorator,
     Diagnostic,
     Diagnostics,
+    diagnosticsEqualityComparer,
     ElementAccessChain,
     ElementAccessExpression,
     emptyArray,
@@ -304,7 +305,7 @@ export function isExternalModuleNameRelative(moduleName: string): boolean {
 }
 
 export function sortAndDeduplicateDiagnostics<T extends Diagnostic>(diagnostics: readonly T[]): SortedReadonlyArray<T> {
-    return sortAndDeduplicate<T>(diagnostics, compareDiagnostics);
+    return sortAndDeduplicate<T>(diagnostics, compareDiagnostics, diagnosticsEqualityComparer);
 }
 
 export function getDefaultLibFileName(options: CompilerOptions): string {
--- a/src/compiler/core.ts
+++ b/src/compiler/core.ts
@@ -808,7 +808,12 @@ export function createSortedArray<T>(): SortedArray<T> {
 }
 
 /** @internal */
-export function insertSorted<T>(array: SortedArray<T>, insert: T, compare: Comparer<T>, allowDuplicates?: boolean): boolean {
+export function insertSorted<T>(
+    array: SortedArray<T>,
+    insert: T,
+    compare: Comparer<T>,
+    equalityComparer?: EqualityComparer<T>,
+    allowDuplicates?: boolean): boolean {
     if (array.length === 0) {
         array.push(insert);
         return true;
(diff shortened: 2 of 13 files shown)


gabritto:
@typescript-bot test it

jakebailey:
@typescript-bot run dt

src/compiler/utilitiesPublic.ts
```diff
@@ -304,7 +305,7 @@ export function isExternalModuleNameRelative(moduleName: string): boolean {
 }
 
 export function sortAndDeduplicateDiagnostics<T extends Diagnostic>(diagnostics: readonly T[]): SortedReadonlyArray<T> {
-    return sortAndDeduplicate<T>(diagnostics, compareDiagnostics);
+    return sortAndDeduplicate<T>(diagnostics, compareDiagnostics, diagnosticsEqualityComparer);
```
weswigham, line 308: As far as I can tell, `DIagnosticCollection`'s `.add` method is still using `compareDiagnosticsSkipRelatedInformation`, so it's just first-in-wins (with the same root message) - should that also get updated to prefer the "lesser" message like this?

gabritto: `compareDiagnosticsSkipRelatedInformation` is still the same in terms of what it considers equal - two diagnostics with the same location and head message but different elaboration will still be considered different by that function, and so both will be added via `DiagnosticCollection.add`. Only later, when we call `sortAndDeduplicateDiagnostics`, will we use `diagnosticsEqualityComparer` and get rid of all but one of the diagnostics that we now consider equal.

gabritto: Now that I think about it, if we want to keep using functions like `sortAndDeduplicate` and `insertSorted`, I think we need to have two separate functions for comparison and equality: one that fully compares diagnostics for purposes of sorting (and sorting a diagnostic with more elaboration before one with less), and another for purposes of deduplication that is more permissive and only compares location and head message for equality. Otherwise we'd have to implement a function that somehow says "yes, those two diagnostics are equal, but one of them is preferred".

andrewbranch (approved):

weswigham (approved):

PR #58221, opened 2024-04-17
Fix miss-reported node reuse in types.

dragomirtitian:
Node reuse is not being reported correctly. When using a cached node builder result the original node was being set as a synthetic node (the node that was in the cache) rather than an original source node.

Ex:

```txt
var v = (a: ) => {
>v : (a: any) => void                         //<- correct, no node reuse
>  : ^ ^^^^^^^^^^^^^^
>(a: ) => {   } : (a: any) => void
>               :                        //<-  second time around nodes are reused?
```


dragomirtitian:
cc: @weswigham

src/compiler/checker.ts
```diff
@@ -6009,8 +6009,8 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             if (!nodeIsSynthesized(range) && !(range.flags & NodeFlags.Synthesized) && (!context.enclosingFile || context.enclosingFile !== getSourceFileOfNode(range))) {
                 range = factory.cloneNode(range);
             }
-            if (!location) {
-                return range;
+            if (!location || nodeIsSynthesized(location)) {
+                return setOriginalNode(range, location?.original);
```
weswigham, line 6013: If `range` already has a `.original` set, this is going to unset it if `location` is `undefined` or lacks a `.original` of its own, which, while is what fixes the issue you witness, doesn't seem desirable. Unsetting `.original` on a copy is probably not a good thing. If this issue is that when we clone a node from the cache, we `.original` it, we should update the underlining requirements in `nodeIsFullySynthetic` in `typeWriter.ts` to instead of just checking for a `.original`, specifically look for a `getParseTreeNode` instead. So,
```ts
function nodeIsFullySynthetic(node: ts.Node) {
    return ts.nodeIsSynthesized(node) && !ts.getParseTreeNode(node);
}
```

dragomirtitian: That also works. My thought was that the node you get should be the same regardless of if it comes from the cache or from the original operation.

dragomirtitian: Changed to using the parse tree node.

weswigham (approved):

weswigham (changes requested):
I don't think we need the new `nodeIsSynthesized` check, either - all that's going to do is block setting `.original` on the range when the `location` has no position data, but it may still have a `.original` of its' own that is a parse tree node, and we wanna keep that 3 node chain (and is what `getParseTreeNode` looks up)

weswigham:
🤦 I meant the one just added in this PR in `checker.ts` - though that one is *probably* redundant, too (since we generally set original ndoe now), I'd rather keep it, since a node made with a `pos`/`end` set, but no `.original` set to a parse node is probably still based enough on an input node, since it has concrete positions.

PR #58222, opened 2024-04-17
add types for iterator helpers proposal

### LINKED ISSUE #54481: PSA: potential lib breaking change after iterator-helpers proposal
# lib Update Request

## Configuration Check

My compilation *target* is `ESNext` and my *lib* is `the default`.

The TS lib defs for `Iterator`, `Iterable`, and `IterableIterator` treat them as if they are all fictitious interfaces (a.k.a. protocols). The lib def's inheritance structure looks like this:

```ts
interface Iterator<T, TReturn = any, TNext = undefined> {
    // NOTE: 'next' is defined using a tuple to ensure we report the correct assignability errors in all places.
    next(...args: [] | [TNext]): IteratorResult<T, TReturn>;
    return?(value?: TReturn): IteratorResult<T, TReturn>;
    throw?(e?: any): IteratorResult<T, TReturn>;
}

interface Iterable<T> {
    [Symbol.iterator](): Iterator<T>;
}

interface IterableIterator<T> extends Iterator<T> {
    [Symbol.iterator](): IterableIterator<T>;
}
```

However, while `Iterable` and arguably `IterableIterator` are protocols, `Iterator` is an actual JS entity. See MDN docs: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Iterator This has some practical issues, because what TS terms as `IterableIterator` is the actual JS `Iterator` class, but what TS terms as `Iterator` is what JS regards as "iterator-like" or "iterator protocol". For example, the def for `new Set().values()` says it returns an `IterableIterator`, while in fact it should return an instance of the JS `Iterator` instance.

This becomes more of an issue after the [iterator-helpers](https://github.com/tc39/proposal-iterator-helpers) proposal, which is going to expose `Iterator` as an actual global that people can use as `extends`. For example:

```js
class MyClass {
  static #MyClassIterator = class extends Iterator {
    next() { ... }
  };

  [Symbol.iterator]() {
    return new MyClass.#MyClassIterator();
  }
}
```

Under the current lib def, `#MyClassIterator` won't be seen as iterable by TS, because it extends `Iterator`, not `IterableIterator`.

On the other hand, if we change the definition of `Iterator` to say it has `[Symbol.iterator]` (mirroring what happens in JS), then it will break code of the following kind:

```ts
function stepIterator(it: Iterator) {
  return it.next();
}

stepIterator({ next() {} });
```

Of course you could assume that almost all userland iterators are built-in and therefore inherit from `Iterator` and have `@@iterator`, but it's a breaking change nonetheless.

The question thus arises that when iterator-helpers lands, where should we put the new definitions on. Right now, I'm using core-js to polyfill these methods, and when writing declarations, I chose to put them on `IterableIterator`, so that `new Set().values()` works:

```ts
declare global {
  interface IterableIterator<T> {
    filter(
      predicate: (value: T, index: number) => unknown,
    ): IterableIterator<T>;
    map<U>(mapper: (value: T, index: number) => U): IterableIterator<U>;
    toArray(): T[];
  }
}
```

...but this is suboptimal, for obvious reasons (it means the `Iterator` class doesn't actually have these methods).

bakkot:
Fixes #54481.

I have only minimal tests because almost all the types are copied directly from Array with very minimal changes (mostly dropping the third "array" parameter to callbacks and the "thisArg" parameter to callback-taking methods, and returning `NativeIterator<T>` instead of `T[]`). But I'm happy to make the tests more extensive if you'd like.

For reviewing:
- ignore bd51b0076f0e4d8fd039898c2ca8ac7a23e1b6e2 through 30d9bc258b8e2b72fe2fc03a2d7e05e1f72a12fb; the last one reverts the earlier ones so the next commits can be clearer
- f5393db657427d6afae9dbf11e7abfba0f615da5 and ebc4e71095f4c94b28872cb1f78afa2fd3753c61 add the new `BuiltinIterator` and `AsyncBuiltinIterator` types and adopt them everywhere except the `.generated` files. `AsyncBuiltinIterator` isn't necessary yet but I figured we might as well do this plumbing work now.
- c4a8a08e5b96365f63dcca1a576ddc9292cabe7b adds the new types following the strategy @rbuckton suggested [below](https://github.com/microsoft/TypeScript/pull/58222#issuecomment-2061741499)

The "rebaseline" commits are just running `npm run test -- --no-lint; npx hereby baseline-accept` and committing the result.

---

Current status (as of 2024-04-25):
- waiting on https://github.com/microsoft/TypeScript/pull/58243
- needs `CloseableBuiltinIterator` or something to mean "built-in iterator with a `return` method"
- needs a rebase
- needs reviews from the team

I'll address the outstanding todos one #58243 lands.


fatcerberus:
> I will need someone to tell me how to handle the generated DOM typings

The DOM typings are maintained at https://github.com/microsoft/TypeScript-DOM-lib-generator

bakkot:
More concretely, I'll need someone to tell me how to handle the change: changing the DOM typings to use `NativeIterator` would break anyone not importing `esnext.iterator`, but there's not (currently) an equivalent of `esnext` for DOM typings, so there's not anywhere else to do the update.

Renegade334:
> More concretely, I'll need someone to tell me how to handle the change: changing the DOM typings to use `NativeIterator` would break anyone not importing `esnext.iterator`, but there's not (currently) an equivalent of `esnext` for DOM typings, so there's not anywhere else to do the update.

Would the following not work?

1. Introduce a hollow `NativeIterator` interface in es2015.iterable.d.ts
2. Re-type all intrinsic appearances of `IterableIterator` in es2015.iterable.d.ts etc. to `NativeIterator`
3. Extend the `NativeIterator` interface in esnext.iterator.d.ts, add the constructor definition, etc.

This would allow what you're looking for: keeps compatibility for pre-esnext targets, while allowing the helper methods to be merged into `NativeIterator` in lib.esnext.

Sparse illustration:
```ts
// @target: esnext
// @lib: esnext

// @filename: lib.es2015.iterable.d.ts

interface Iterator<T, TReturn = any, TNext = undefined> {
    next(...args: [] | [TNext]): IteratorResult<T, TReturn>;
    return?(value?: TReturn): IteratorResult<T, TReturn>;
    throw?(e?: any): IteratorResult<T, TReturn>;
}

interface Iterable<T> {
    [Symbol.iterator](): Iterator<T>;
}

interface IterableIterator<T> extends Iterator<T> {
    [Symbol.iterator](): IterableIterator<T>;
}

interface NativeIterator<T, TReturn = void, TNext = undefined> extends Iterator<T, TReturn, TNext> {
    [Symbol.iterator](): NativeIterator<T, TReturn, TNext>;
}

interface Generator<T = unknown, TReturn = any, TNext = unknown> extends NativeIterator<T, TReturn, TNext> {
    // etc.
}

interface Array<T> /* etc. */ {
    // was: [Symbol.iterator](): IterableIterator<T>;
    [Symbol.iterator](): NativeIterator<T>;

    // was: entries(): IterableIterator<[number, T]>;
    entries(): NativeIterator<[number, T]>;

    // etc.
}


// @filename: lib.esnext.iterator.d.ts

interface NativeIterator<T, TReturn, TNext> {
    map<U>(callbackfn: (value: T, index: number) => U): NativeIterator<U>;

    // etc.
}

// @filename: test.ts

const mappedArrayIterator = ['a', 'b', 'c'].entries().map(([k, s]) => k * s.charCodeAt(0)); // NativeIterator<number>

const castToIterableIterator: IterableIterator<number> = mappedArrayIterator; // NativeIterator<T> remains assignable to IterableIterator<T>, for what it's worth
```

bakkot:
> Would the following not work?

Nice, that sounds like a good approach to me.

rbuckton:
@bakkot the approach I am experimenting with to support `abstract` on `NativeIterator` looks something like this:

```ts
// lib.esnext.iterator.d.ts
export {};

// Abstract type that allows us to mark `next` as `abstract`
declare abstract class Iterator<T> {
  abstract next(value?: undefined): IteratorResult<T, void>;
}

// Merge all members of `NativeIterator<T>` into `Iterator<T>`
interface Iterator<T> extends globalThis.NativeIterator<T> {}

// Capture the `Iterator` constructor in a type we can use in the `extends` clause of `IteratorConstructor`.
type NativeIteratorConstructor = typeof Iterator;

declare global {
  // Global `NativeIterator<T>` interface that can be augmented by polyfills
  interface NativeIterator<T> extends Iterator<T> {
    [Symbol.iterator](): NativeIterator<T>;
    // prototype elements
  }

  // Global `IteratorConstructor` interface that can be augmented by polyfills
  interface IteratorConstructor extends NativeIteratorConstructor {
    // static elements
  }

  var Iterator: IteratorConstructor;
}
```

And in use:

```ts
new Iterator<number>(); // ts(2511): Cannot create an instance of an abstract class.

class C extends Iterator<number> {} // ts(2515): Non-abstract class 'C' does not implement inherited
                                    //           abstract member next from class 'Iterator<number>'.
```

src/lib/esnext.iterator.d.ts
```diff
@@ -0,0 +1,133 @@
...
+     * Creates a native iterator from an iterator or iterable object.
+     * Returns its input if the input already inherits from the built-in Iterator class.
+     * @param value An iterator or iterable object to convert a native iterator.
+     */
+    from<T>(value: Iterator<T> | Iterable<T>): NativeIterator<T>;
+}
+
+declare var Iterator: (abstract new <T>() => NativeIterator<T>) & IteratorConstructor;
+
+// TODO BEFORE MERGING: update all existing IterableIterator-return methods to return NativeIterator
```
rbuckton, line 109: As suggested elsewhere, I would suggest you define `NativeIterator` as follows in the es2015 libs and then just update all of the `IterableIterator` references to `NativeIterator`:

```ts
interface NativeIterator<T, TReturn = void, TNext = undefined> extends Iterator<T, TReturn, TNext> {
  [Symbol.iterator](): NativeIterator<T, TReturn, TNext>;
}
```

src/lib/esnext.iterator.d.ts
```diff
@@ -0,0 +1,133 @@
+interface NativeIterator<T, TReturn = void, TNext = undefined> extends Iterator<T, TReturn, TNext> {
```
rbuckton, line 1: The `export {}` trick mentioned in my earlier comment can address the `abstract next()` method definition.

rbuckton (changes requested):

src/lib/esnext.iterator.d.ts
Jack-Works, line 1: The word choose "Native" is not so good. Do we have other options?

Renegade334:
@rbuckton does the interface merging make any existing "custom" Iterator objects fail type validation in target:esnext, since they won't implement the new Iterator.prototype methods?

Jack-Works:
> @rbuckton does the interface merging make any existing "custom" Iterator objects fail type validation in target:esnext, since they won't implement the new Iterator.prototype methods?

It does not merging the current Iterator interface, but create a new one called NativeIterator

src/lib/esnext.iterator.d.ts
bakkot: Should I split that into a separate PR, or do it here?

src/lib/esnext.iterator.d.ts
bakkot: @rbuckton had previously suggested [`Builtin`](https://github.com/microsoft/TypeScript/issues/54481#issuecomment-1572920703), which I'm also fine with.

rbuckton:
> @rbuckton does the interface merging make any existing "custom" Iterator objects fail type validation in target:esnext, since they won't implement the new Iterator.prototype methods?

@bakkot is correct. This should not affect anyone implementing a custom `Iterator`, or even `IterableIterator`.

After this is in we may want to add a shortcuts for it to `getIterationTypesOfIterableFast` and `getIterationTypesOfIteratorFast` in the checker.

bakkot:
Updated almost exactly following [this comment](https://github.com/microsoft/TypeScript/pull/58222#issuecomment-2061741499), except that I had BuiltinIterator's `Symbol.iterator` method pass all the type arguments to the return type; this is necessary so `Generator` can extend `BuiltinIterator` (otherwise the `Symbol.iterator` methods' types conflict).

---

A consequence of this change is that the type of `yield* []` is now `void` instead of `any`, because the default for `TReturn` on `BuiltinIterator` is `void` instead of `any` (as it is for `Iterator).

----

Also, passing `void` for the second argument in the return type in `next` in

```ts
declare abstract class Iterator<T> {
    abstract next(value?: undefined): IteratorResult<T, void>;
}
```
means that classes attempted to implement `Iterator` have some pain: you can't just `return { done: false, value: 0 }`  from your `next` because the `done` value there will be inferred as `boolean`, which means that the `value` field has to conform to `void` so that it can extend the `{ done: true, value: TReturn }` half of `IteratorResult<T, void>`, which it does not. Also, you can't make an iterator which actually does have a non-void `value` for the `done: true` case, which seems like an unnecessary constraint.

I have test cases illustrating the difficulties:

```
    class BadIterator1 extends Iterator<number> {
      next() {
      ~~~~
!!! error TS2416: Property 'next' in type 'BadIterator1' is not assignable to the same property in base type 'Iterator<number>'.
!!! error TS2416:   Type '() => { readonly done: false; readonly value: 0; } | { readonly done: true; readonly value: "a string"; }' is not assignable to type '(value?: undefined) => IteratorResult<number, void>'.
!!! error TS2416:     Type '{ readonly done: false; readonly value: 0; } | { readonly done: true; readonly value: "a string"; }' is not assignable to type 'IteratorResult<number, void>'.
!!! error TS2416:       Type '{ readonly done: true; readonly value: "a string"; }' is not assignable to type 'IteratorResult<number, void>'.
!!! error TS2416:         Type '{ readonly done: true; readonly value: "a string"; }' is not assignable to type 'IteratorReturnResult<void>'.
!!! error TS2416:           Types of property 'value' are incompatible.
!!! error TS2416:             Type 'string' is not assignable to type 'void'.
        if (Math.random() < .5) {
          return { done: false, value: 0 } as const;
        } else {
          return { done: true, value: "a string" } as const;
        }
      }
    }
    
    class BadIterator2 extends Iterator<number> {
      next() {
      ~~~~
!!! error TS2416: Property 'next' in type 'BadIterator2' is not assignable to the same property in base type 'Iterator<number>'.
!!! error TS2416:   Type '() => { done: boolean; value: number; }' is not assignable to type '(value?: undefined) => IteratorResult<number, void>'.
!!! error TS2416:     Type '{ done: boolean; value: number; }' is not assignable to type 'IteratorResult<number, void>'.
!!! error TS2416:       Type '{ done: boolean; value: number; }' is not assignable to type 'IteratorReturnResult<void>'.
!!! error TS2416:         Types of property 'done' are incompatible.
!!! error TS2416:           Type 'boolean' is not assignable to type 'true'.
        return { done: false, value: 0 };
      }
    }
    
    class BadIterator3 extends Iterator<number> {
      next() {
      ~~~~
!!! error TS2416: Property 'next' in type 'BadIterator3' is not assignable to the same property in base type 'Iterator<number>'.
!!! error TS2416:   Type '() => { done: boolean; value: number; } | { done: boolean; value: string; }' is not assignable to type '(value?: undefined) => IteratorResult<number, void>'.
!!! error TS2416:     Type '{ done: boolean; value: number; } | { done: boolean; value: string; }' is not assignable to type 'IteratorResult<number, void>'.
!!! error TS2416:       Type '{ done: boolean; value: number; }' is not assignable to type 'IteratorResult<number, void>'.
!!! error TS2416:         Type '{ done: boolean; value: number; }' is not assignable to type 'IteratorReturnResult<void>'.
!!! error TS2416:           Types of property 'done' are incompatible.
!!! error TS2416:             Type 'boolean' is not assignable to type 'true'.
        if (Math.random() < .5) {
          return { done: false, value: 0 };
        } else {
          return { done: true, value: "a string" };
        }
      }
    }
```


I see a few approaches to improving this last issue:
- pass `any` as the second parameter for `IteratorResult` in the `abstract class Iterator` (this fixes the `BadIterator1` and `BadIterator2` cases)
- pass `unknown` as the second parameter for `IteratorResult` in the above (this fixes the `BadIterator1` and `BadIterator2` cases)
- make `abstract class Iterator` take a second parameter for `TReturn`, defaulting to `void` (this fixes the `BadIterator1` and `BadIterator2` cases if the user explicitly specifies the second parameter in the classes)

bakkot:
One detail not captured by the types here is that the iterator returned by helpers like `map` unconditionally has a `return` method. I guess that's probably worth expressing?

rbuckton:
> means that classes attempted to implement `Iterator` have some pain: you can't just `return { done: false, value: 0 }` from your `next` because the `done` value there will be inferred as `boolean`, which means that the `value` field has to conform to `void` so that it can extend the `{ done: true, value: TReturn }` half of `IteratorResult<T, void>`, which it does not.

We could probably add `TReturn = void, TNext = undefined` type parameters to the built in `Iterator` class without issue, since it's up to the user to implement `next` anyways.

rbuckton:
> One detail not captured by the types here is that the iterator returned by helpers like `map` unconditionally has a `return` method. I guess that's probably worth expressing?

Do helpers like `map` also have a `throw` method? If so, you could just use `Generator` for those. If not, we'd have to add another `NativeIterator` subtype for those.

rbuckton:
Regarding the typing issues, I'm hoping #58243 might help with that.

bakkot:
> Do helpers like `map` also have a `throw` method? If so, you could just use `Generator` for those. If not, we'd have to add another `NativeIterator` subtype for those.

Nope, just `return`. `throw` is really a generator-specific thing.

Suggestions for the name? `CloseableBuiltinIterator`, maybe, since presence of `return` indicates that you can close the iterator?

felixfbecker:
Some more ideas:
- `HelpfulIterator`
- `HelperIterator`
- `TransformableIterator`
- `IteratorInstance`

Josh-Cena:
This "`new class extends Iterator` does not return something cross-compatible with `Iterator`" behavior still sounds confusing to me, IMO.

bakkot:
@Josh-Cena Not sure what you mean. It is cross-compatible as long as your class implements the `next` method, which you are required to do anyway because you're extending an abstract class. That is, this works:

```ts
class GoodIterator extends Iterator<number> {
  next() {
    return { done: false, value: 0 } as const;
  }
}

let x: Iterator<number> = new GoodIterator();
```

What's a concrete example of the case you're worried about?

Josh-Cena:
Sorry, I didn't say what I meant to say.

If we have the following:

```ts
declare const x: Iterator<number>;

const y = new Iterator();
```

Then we expect `x` and `y` to be mutually replaceable with each other. Now because here `SomeClass` is an abstract class, `new Iterator()` doesn't make sense, so let's use your `GoodIterator` instead. But, there should be no difference because `GoodIterator` merely overrides an abstract method but does not implement any additional behavior. However, I would be surprised to discover that `x` is not iterable, despite the fact that the abstract class `Iterator` provides `Symbol.iterator`.

```ts
declare const x: Iterator;

for (const val of x) {
  // Doesn't work
}
```

It's like a value of type `Map` and `new Map` doing different things.

Josh-Cena:
Actually, I see that in esnext `Iterator` would also have the methods of `BuiltInIterator`, I'm still slightly surprised as we didn't really "add" `Symbol.iterator` in esnext...

bakkot:
Ah, yeah. That's an unavoidable consequence of TS already having an `Iterator` interface, though - the only alternative is to change meaning of the the type-level `Iterator`, which would be a pretty large breaking change. I think this ends up being the best available option, despite the downside you point out.

Josh-Cena:
Um... Sure? There still are breaking changes and anyone using esnext might not be able to interop with old libraries, though. Instead of adding a new name which is virtually impossible to remove in the future (while its utility diminishes as people adopt the current esnext), I still feel it's better to accept a minor breaking change to `Iterator`, assuming people nearly always use generators anyway.

bakkot:
There still needs to be a distinction between "has a `next` method" and "has all the methods given by the built-in ES `Iterator` class", so we still need to have two named types - we can't just add more stuff to the existing TS `Iterator` interface and be done with it. I think it makes sense for the first to be named "Iterator" and the second to be named `BuiltinIterator`.

Renegade334:
> Um... I see the reason. There still are breaking changes and anyone using esnext might not be able to interop with old libraries, though. Instead of adding a new name which is virtually impossible to remove in the future (while its utility diminishes as people adopt the current esnext), I still feel it's better to accept a minor breaking change to `Iterator`, assuming people nearly always use generators anyway.

There is no breaking change; in fact, quite the opposite. The declarations in esnext.iterators.d.ts are defined under module scope, and then selectively exported into the global namespace in such a way that they don't interfere with the existing `Iterator` interface.

Referencing `Iterator` by _value_ will refer to the new global Iterator constructor. Referencing `Iterator` by _type_ will refer to the existing "an object that implements iterator methods" interface. The former allows users to refer to `Iterator.prototype` and extend the class if desired, while the latter ensures backward compatibility.

Josh-Cena:
I missed the `export {}` in the iterator def. Then my only qualm is the divergence between value-space and type-space `Iterator` structure.

conmaster2112:
Isn't it better to make breaking changes and keep up with the ECMAScript implementation? I know that the result would be disastrous, but sometimes it has to happen because the change to NativeIterator could be very confusing for newbies in a few years or until we are no longer here. TS would become very messy and I think keeping up with ES should be more important.

Are there any other reasons why not moving current Iterator to IteratorLike, except the breaking change?

GabenGar:
@conmaster2112 
> the change to NativeIterator could be very confusing for newbies in a few years or until we are no longer here

It's a general problem with made up typescript types living in a global space. I wish they would all be chucked into `typescript` module and require an explicit import like `import type { Record } from "typescript/types/util"` to clearly denote these are 3rd-party declarations not related to ecmascript/dom/html/workers. But that would require all JSDoc (more like TSDoc) users to import them like this too and realize they still depend on typescript even in explicitly "vanilla js without a build step complexity" projects. And we can't have that.

> Are there any other reasons why not moving current Iterator to IteratorLike, except the breaking change?

It wouldn't look good in a blogpost for a new version announcement.

rbuckton:
Assuming we land #58243 today or tomorrow, I'd like to try to get this PR updated and ready to merge tomorrow as well. In addition to updating from `main`, there are a few changes we will need to make to align with #58243 and `--strictBuiltinIteratorReturn`. I'll provide additional feedback once #58243 has merged, or if needed I can directly update this PR with any necessary changes.

bakkot:
@rbuckton I'm happy either way; if it's easier for you to just push the changes yourself rather than leaving feedback that's fine by me. Otherwise I'll try to get to your feedback asap, though I can't commit to having it done tomorrow.

I did the initial update on `main`, though of course it will need another after https://github.com/microsoft/TypeScript/pull/58243.

src/lib/es2015.iterable.d.ts
```diff
@@ -27,6 +27,10 @@ interface Iterator<T, TReturn = any, TNext = undefined> {
     throw?(e?: any): IteratorResult<T, TReturn>;
 }
 
+interface BuiltinIterator<T, TReturn = void, TNext = undefined> extends Iterator<T, TReturn, TNext> {
```
rbuckton, line 30: It's possible this may need to be
```ts
interface BuiltinIterator<T, TReturn = any, TNext = any> extends Iterator<T, TReturn, TNext> {
```
based on my findings in #58243, or `BuiltinIterator` shouldn't supply defaults at all.

src/lib/esnext.iterator.d.ts
```diff
@@ -0,0 +1,125 @@
+/// <reference lib="es2015.iterable" />
+export {};
+
+// Abstract type that allows us to mark `next` as `abstract`
+declare abstract class Iterator<T> {
+    abstract next(value?: undefined): IteratorResult<T, void>;
+}
+
+// Merge all members of `BuiltinIterator<T>` into `Iterator<T>`
+interface Iterator<T> extends globalThis.BuiltinIterator<T, void, undefined> {}
```
rbuckton, line 10: Should we add `TReturn` and `TNext` type parameters to `Iterator`?

We also probably shouldn't specify `TNext` as `undefined` since builtins generally don't actually care what the argument is. You could theoretically have a generator that takes `number | undefined` for `TNext`, but it wouldn't be able to delegate to `yield* [1, 2]` since `number | undefined` isn't assignable to `undefined`.

src/lib/esnext.iterator.d.ts
```diff
@@ -0,0 +1,125 @@
...
+         * Creates an iterator whose values are the values from this iterator after skipping the provided count.
+         * @param count The number of values to drop.
+         */
+        drop(count: number): BuiltinIterator<T>;
+
+        /**
+         * Creates an iterator whose values are the result of applying the callback to the values from this iterator and then flattening the resulting iterators or iterables.
+         * @param callback A function that accepts up to two arguments to be used to transform values from the underlying iterator into new iterators or iterables to be flattened into the result.
+         */
+        flatMap<U>(callback: (value: T, index: number) => Iterator<U> | Iterable<U>): BuiltinIterator<U>;
```
rbuckton, line 57: Should we restrict this to only allow `Iterator<U, unknown, undefined> | Iterable<U, unknown, undefined>`, since `flatMap` is going to call `next()` on the result without passing an argument?

src/lib/esnext.iterator.d.ts
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    // Global `IteratorConstructor` interface that can be augmented by polyfills
+    interface IteratorConstructor extends BuiltinIteratorConstructor {
+        /**
+         * Creates a native iterator from an iterator or iterable object.
+         * Returns its input if the input already inherits from the built-in Iterator class.
+         * @param value An iterator or iterable object to convert a native iterator.
+         */
+        from<T>(value: Iterator<T> | Iterable<T>): BuiltinIterator<T>;
```
rbuckton, line 121: Should we restrict this to only allow `Iterator<U, unknown, undefined> | Iterable<U, unknown, undefined>`, since `from` is going to call `next()` on the result without passing an argument?

rbuckton:
I've made updates to align with the changes in https://github.com/microsoft/TypeScript/pull/58243, though it required a few small compromises. If necessary, we can refine this somewhat after its in the beta.

src/lib/esnext.iterator.d.ts
rbuckton: I've added this restriction for now, since it seems useful, though we don't prevent you from using `.map` on a `Generator<string, boolean, number>`, even though the generator expects to receive `number` in calls to `next()`, and not `undefined`.

src/lib/esnext.iterator.d.ts
rbuckton: I've opted to leave `TReturn` and `TNext` off of the `Iterator<T>` class for now, though we may want to revisit that after we ship the beta.

src/lib/es2015.iterable.d.ts
rbuckton: For now I've opted to use the weaker defaults.

rbuckton (approved):

rbuckton:
I realize I neglected to run the user test suite before merging, though I expect it will only call out projects that were already called out by #58243. Running that suite now in case that assumption turns out to be false.

@typescript-bot run dt
@typescript-bot test top400

rbuckton:
Also running benchmarks as I expect I need to put up a follow-up PR that adds a `BuiltinIterator` optimization to the checker like we do for `Iterator`

@typescript-bot perf test

petamoriken:
Why `AsyncBuiltinIterator`? I think `BuiltinAsyncIterator` is a more consistent name.

petamoriken:
ref: #59388

PR #58227, opened 2024-04-17
Error on replacement character only in top-level scanning

### LINKED ISSUE #57930: Replacement character (`\uFFFD`) in JavaScript/TypeScript causes to breaks
### 🔎 Search Terms

* TS1490: File appears to be binary
* replacement character

#57008 is incorrect on some real files.

### ⏯ Playground Link

https://github.com/micromark/micromark/pull/166

### 💻 Code

Note the `�` in the first 256 characters.

```js
import assert from 'node:assert/strict'
import test from 'node:test'
import {micromark} from 'micromark'

test('nul', function () {
  assert.equal(
    micromark('asd\0asd'),
    '<p>asd�asd</p>',
    'should replace `\\0` w/ a replacement characters (`�`)'
  )
})
```

### 🙁 Actual behavior

*error TS1490: File appears to be binary*

### 🙂 Expected behavior

0xFFFD should be fine among other regular characters.

Related-to: #21136.
Related-to: #56516.
Related-to: #57008.

jakebailey:
This one came to me in the shower; everyone with this character "legally" in their files can only have put them into string literals or comments. If we just check this at the top level scan function, we'll never have a false positive because it'd definitely be a parse error at runtime. If any of the other scanner functions (e.g. `scanIdentifier`) see this character code, they're going to bail out anyway because it's an unexpected character, then this new case will hit.

I could also modify this to check only at the start of a file, but since we're now back in the switch-case where we process all of the other scanning, this isn't actually inefficient anymore.

Fixes #57930


jakebailey:
@typescript-bot test it

RyanCavanaugh (approved):

jakebailey:

sheetalkamat (approved):

jakebailey:
@typescript-bot run dt

wooorm:
Thanks Jake!

PR #58228, opened 2024-04-17
Remove deprecated originalKeywordKind and isInJSDocNamespace

### LINKED ISSUE #51496: Experiment with removing `originalKeywordKind` from every `Identifier`
In our factories, every `Identifier` has an unconditional assignment to `originalKeywordKind`. This feels wasteful, but it's at the expense of reducing polymorphism. As an experiment, we should see

1. Whether we can conditionally set it at no cost, or
2. Whether we can just remove it entirely, computing the necessary information on the fly.

jakebailey:
These were both deprecated in 5.0, started warning in 5.1, and now throw errors in 5.2+; I think it's safe to remove these. Probably helps perf to do so.

Closes #51496


jakebailey:

jakebailey:
@typescript-bot perf test this predictable

jakebailey:
noisy noisy...

DanielRosenwasser (approved):

rbuckton (approved):

PR #58233, opened 2024-04-17
Don't cache child lists for tokens

DanielRosenwasser:
This PR always returns a shared frozen array from `getNodeChildren` when given a token (e.g. an identifier or any other keyword, punctuation, etc.). It also marks all child-lists as `readonly Node[]` instead of just `Node[]`.

If we want, we can probably just use the internal `emptyArray`, but I don't think we've ever frozen that one.

### DIFF
--- a/src/compiler/factory/nodeChildren.ts
+++ b/src/compiler/factory/nodeChildren.ts
@@ -1,4 +1,7 @@
-import { Node, isToken } from "../_namespaces/ts";
+import {
+    isToken,
+    Node,
+} from "../_namespaces/ts";
 
 const nodeChildren = new WeakMap<Node, readonly Node[] | undefined>();
 
--- a/src/compiler/factory/nodeChildren.ts
+++ b/src/compiler/factory/nodeChildren.ts
@@ -1,6 +1,6 @@
 import {
     emptyArray,
-    isToken,
+    isNodeKind,
     Node,
 } from "../_namespaces/ts";
 
@@ -8,7 +8,8 @@ const nodeChildren = new WeakMap<Node, readonly Node[] | undefined>();
 
 /** @internal */
 export function getNodeChildren(node: Node): readonly Node[] | undefined {
-    if (isToken(node)) return emptyArray;
+    if (!isNodeKind(node.kind)) return emptyArray;
+
     return nodeChildren.get(node);
 }
 
--- a/src/compiler/factory/nodeChildren.ts
+++ b/src/compiler/factory/nodeChildren.ts
@@ -1,15 +1,14 @@
 import {
+    emptyArray,
     isToken,
     Node,
 } from "../_namespaces/ts";
 
 const nodeChildren = new WeakMap<Node, readonly Node[] | undefined>();
 
-const emptyChildren: readonly Node[] = Object.freeze([]);
-
 /** @internal */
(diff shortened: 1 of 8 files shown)


DanielRosenwasser:

DanielRosenwasser:
I guess an alternative path would be to have `IdentifierObject`, `PrivateIdentifierObject`, and `TokenObject` all have their own `getChildren`, but 🤷‍♂️.

src/compiler/factory/nodeChildren.ts
```diff
@@ -1,14 +1,20 @@
-import { Node } from "../_namespaces/ts";
+import {
+    isToken,
+    Node,
+} from "../_namespaces/ts";
 
-const nodeChildren = new WeakMap<Node, Node[] | undefined>();
+const nodeChildren = new WeakMap<Node, readonly Node[] | undefined>();
+
+const emptyChildren: readonly Node[] = Object.freeze([]);
```
rbuckton, line 8: Why not use `emptyArray`?

jakebailey: Confusingly, services defines its own mutable empty array (and none of them actually guard against modification)

jakebailey: Oh, this isn't even in services, yeah, emptyArray would be nice. (I used to have a PR that froze it but it didn't really seem to matter)

DanielRosenwasser: From the description:

> If we want, we can probably just use the internal `emptyArray`, but I don't think we've ever frozen that one.

Happy to switch over to that too, but Jake mentioned that he didn't want to make that change since `getChildren()` always was declared as a mutable array. Freezing this array (for runtime safety) and propagating the change (for static verification) helped validate that it's unlikely we mutate these arrays.

DanielRosenwasser:
@typescript-bot test top400

src/compiler/factory/nodeChildren.ts
rbuckton: Does freezing `emptyArray` affect our perf at all? If not, I'd just do that and use `emptyArray`.

jakebailey: I tried in https://github.com/microsoft/TypeScript/pull/55755 but I think that was actually more of a problem with trying to make immutable empty maps/sets. Sent #58240 to try just arrays quick.

jakebailey: it sure looks like freezing is bad: https://github.com/microsoft/TypeScript/pull/58240#issuecomment-2064751428

DanielRosenwasser:

DanielRosenwasser:
It turns out that we always made this optimization - the only difference with this PR is that now

- we no longer stick that `emptyArray` in a `WeakMap`
- every list is marked as `readonly`

I don't know if there is a difference practically between `!isNode(...)` and `isToken(...)`.

jakebailey (approved):
Seems reasonable to me

jakebailey:
> I don't know if there is a difference practically between `!isNode(...)` and `isToken(...)`.

They're equivalent in behavior, but `isNode` only makes one comparison so I'd just keep that.

rbuckton (approved):

PR #57093, opened 2024-01-19
No number constraint to mapped type with as clause

gabritto:
Fixes #57007.

PR #58237, opened 2024-04-18
Add numeric constraint to type parameter of mapped types with name type and array constraints (take 2)

### LINKED ISSUE #55762: Can't remap tuple keys to object keys in mapped types
### 🔎 Search Terms

mapped tuple, mapped types

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about mapped types

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.2.2#code/CYUwxgNghgTiAEkoGdnwCLgPahgHgBUA+eAbwCh4r5QwcQAKWGKATwCEBXAM25BgBc8AIIwWHHnxgBKIQQDc5AL7lyAF1YAHBADUoETiAJaQAeW6F4IAB5qQAO2BpMdXHij3WREgF54BK1sHJwxsNwBLeyl4ACUSAH5Y+CF7EAA3fkV1E3gAZTUYTjA1TGQwGHDNNSwYeD8YkChgLHsIVlFxPDgmlrb4AG0AaxBWIWQCyIBzABp4DW0hF3p8Dy8AXSIs+YR8wuLjbQAxGrxS8srq2ps7RzRdopKQMoqqmt8ySmp+gGl4SPgGMNWFhuKFnhcatJ4AAyeD2TgAWwARvx4CgwedXjAfmt+gAGNZrIR6AxGEzmU5PTGXHH9ACMG0USiySFQeQKDyWbjo9nGGJel0CNxC92KZwFbyFwWcYX4eFFagOIGO+HFEJg3g+1EQLXGe0uDFA4KxiypEpkWu11GQnG0MAY0kU2pUKnItGgcB1vLUfwAzAAmLn8U2uOXw5H8TZu8AehA8vnhABsABYg4JQqH8EjwpNImoo+R4z69ZzZbU-KkAO7s-Vphj9T5UfoAIigzdm4QDabW00bA2bSPbfxT3d7a0dqiL8BLxTq045YrLADpaPQGFWRGI2FxePwGHS8XjpBPC7qfWkUoiUeX5-ql1BFEA

### 💻 Code

```ts
declare class Decoder<T> {
    decode(arrayBuffer: ArrayBuffer): T;
}

type ValueTypeOf<T extends Decoder<any>> = T extends Decoder<infer R> ? R : never;

type StructDescriptor = ReadonlyArray<readonly [key: string, type: Decoder<any>]>;

type StructTypeFor<Descriptor extends StructDescriptor> = {
    [K in (keyof Descriptor) & number as Descriptor[K][0]]: ValueTypeOf<Descriptor[K][1]>;
};

class StructDecoder<const Descriptor extends StructDescriptor> extends Decoder<StructTypeFor<Descriptor>> {
    constructor(descriptor: Descriptor) {
        super();
    }
}

declare const i32Decoder: Decoder<number>;
declare const i64Decoder: Decoder<bigint>;

const structDecoder = new StructDecoder([
    ["a", i32Decoder],
    ["b", i64Decoder],
]);

const struct = structDecoder.decode(new ArrayBuffer(100));

// I expected this would work, but it does not
const v: number = struct.a;
```

### 🙁 Actual behavior

The type of `struct.a` has type `number | bigint`, however it should only have type `number`.

### 🙂 Expected behavior

`struct.a` should have type `number` (and similarly `struct.b` should have type `bigint`).

This does work when using [a record instead](https://www.typescriptlang.org/play?ts=5.2.2#code/CYUwxgNghgTiAEkoGdnwCLgPahgHgBUA+eAbwCh4r5QwcQAKWGKATwCEBXAM25BgBc8AIIwWHHnxgBKIQQDc5AL7lyAF1YAHBADUoETiAJaQAeW6F4IAB5qQAO2BpMdXHij3WREgF54BK1sHJwxsNwBLeyl4ACUSAH5Y+CF7EAA3fkV1E3gAZTUYTjA1TGQwGHDNNSwYeD8YkChgLHsIVjwGuhhgPHtOAFsAI34AGng4Jpa2+ABtAGsQViFkAsiAczGNbSEXenwPLwBdbyythHzC4uNtADEavFLyyuramztHNAuikpAyiqqar4yJRqDMANLwSLwBgLVhYbihP7PGrSeAAMngfSG-HgKERTwBMHBhxmAAZDochHoDEYTOYHr8CS9iTMAIzHRRKLJIVB5ArfXZuOj2Fb4-4vQLvEJfYqPcWAyXBZxhfh4GVqa4gO74OXImDeYHURAtFaXF4MUBIwk7RnymSGo3UZCcbQwBjSRRGlQqci0aBwY0itSQgDMACZBfwba5VVjhvrFH7YAhhaLwgA2AAskcEoRj+EG4TWkTURG5JuDpoFKtqflSAHc+WacwwKEbSUIZgAiKBdsbhcM5w4jEFUVmdruDPuQrNDkdKD2qVOV-nFOrwKuymsAOlo9AYDZEYjYXF4-AYrNJpOki-Iy-gaRSA3j683am3UEUQA), however this isn't as useful as I want the descriptor to still be an array so I can iterate over it.

Andarist:
fixes #55762

this was already approved by @gabritto and @weswigham and got merged here: https://github.com/microsoft/TypeScript/pull/55774 . It was reverted not because it was decided that the change is bad but because it conflicted with another work that was merged *before* it ( https://github.com/microsoft/TypeScript/pull/56727 ) - so it was only reverted because it was merged more recently to `main` than the other one. Later on that other PR was also reverted (it in fact was the one that was incorrect out of the two) and the issues that it was addressing were fixed by other PRs.


tests/cases/compiler/mappedTypeTupleConstraintTypeParameterInNameType.ts
```diff
@@ -0,0 +1,50 @@
...
+const v1_2: bigint = struct1.b;
+
+declare const descriptor2: [["a", Decoder<number>], ["b", Decoder<string>], ...["c", Decoder<bigint>][]]
+const structDecoder2 = new StructDecoder(descriptor2);
+
+const struct2 = structDecoder2.decode(new ArrayBuffer(100));
+
+const v2_1: number = struct2.a; // error, rest element expands to index signature access
+const v2_2: string = struct2.b; // error, rest element expands to index signature access
+const v2_3: bigint = struct2.c; // error, rest element expands to index signature access
```
Andarist, line 50: Note that this is a change from the original https://github.com/microsoft/TypeScript/pull/55774 . In the meantime this got merged in https://github.com/microsoft/TypeScript/pull/57031 so it only makes sense now to always "deoptimize" this mapping to instantiate~ `K` with `number` when any rest element is present in the source tuple.

This isn't very helpful in this situation and maybe it's a reason this PR shouldn't be picked up at all. I don't know - it feels that this suffers from a lack of array/tuple-dedicated syntax for iteration.

The original case with fixed-sized tuples stays improved here though which should be the most popular one anyway.

gabritto:
In the midst of fixing some of the problems with mapped types and arrays/tuples that we had, I think we decided that we'd only do the "actually, only iterate through the array/tuple element keys" special behavior for mapped types only if the modifier type is array-like *and* there's no `as` clause.
IIUC, that's because that's the scenario where we can be confident the output of the mapped type should be array-like. https://github.com/microsoft/TypeScript/pull/57122 improved the first aspect of that (detecting when the modifier type is array-like), and https://github.com/microsoft/TypeScript/pull/57093 fixed a problem in the second aspect of that (`[K in keyof SomeArray as SomeType]: OtherType<K>` no longer considers `K` constrained by `${number}` in the template type).

Note that https://github.com/microsoft/TypeScript/pull/57093 means the following has an additional error on the template type side now:
```ts
type StructTypeFor<Descriptor extends StructDescriptor> = {
    [K in keyof Descriptor as Descriptor[K][0]]: ValueTypeOf<Descriptor[K][1]>;
};
```

I don't know, I feel like this PR in particular is sort of like type inference changes: it does what you want in some cases but maybe not what you want in others?

Andarist:
@gabritto I agree. This doesn't feel good right now with the recent changes. I mostly re-opened this to get clarity that this is not something that is meant to work. What about the corresponding issue? I think it might be worth closing it now.

gabritto:
> @gabritto I agree. This doesn't feel good right now with the recent changes. I mostly re-opened this to get clarity that this is not something that is meant to work. What about the corresponding issue? I think it might be worth closing it now.

Issue's been updated. Thanks for pointing that out.

PR #58242, opened 2024-04-18
Try sorting and deduplicating events before comparing.

sheetalkamat:
Fixes #

### DIFF
--- a/src/testRunner/unittests/sys/symlinkWatching.ts
+++ b/src/testRunner/unittests/sys/symlinkWatching.ts
@@ -127,13 +127,9 @@ describe("unittests:: sys:: symlinkWatching::", () => {
     }
 
     function compareEventFileName(a: EventAndFileName["fileName"], b: EventAndFileName["fileName"]) {
-        return a === b ? ts.Comparison.EqualTo :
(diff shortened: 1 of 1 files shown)


rbuckton (changes requested):

src/testRunner/unittests/sys/symlinkWatching.ts
```diff
@@ -126,23 +126,34 @@ describe("unittests:: sys:: symlinkWatching::", () => {
         return deferred.promise;
     }
 
+    function compareEventFileName(a: EventAndFileName["fileName"], b: EventAndFileName["fileName"]) {
+        if (a === null) a = undefined; // eslint-disable-line no-restricted-syntax
+        if (b === null) b = undefined; // eslint-disable-line no-restricted-syntax
+        return ts.compareStringsCaseSensitive(a, b);
```
rbuckton, line 132: Ah, if that's all that's needed we can even remove the linter comments and just do this:
```suggestion
        return ts.compareStringsCaseSensitive(a ?? undefined, b ?? undefined);
```

rbuckton (changes requested):

rbuckton (approved):

PR #30790, opened 2019-04-06
Improve type checking and inference for Generators and Async Generators

rbuckton:
This PR changes the definitions for `IteratorResult`, `Generator`, and `AsyncGenerator` and makes a number of improvements in how we check and infer types for `yield` and `return` inside of the body of a generator or async generator.

## Type Definition Changes:

```ts
// es2015.iterable.d.ts
interface IteratorYieldResult<TYield> {
    done: false;
    value: TYield;
}

interface IteratorReturnResult<TReturn> {
    done: true;
    value: TReturn;
}

type IteratorResult<T, TReturn = any> = 
    | IteratorYieldResult<T>
    | IteratorReturnResult<TReturn>;

interface Iterator<T, TReturn = any, TNext = undefined> {
    // NOTE: 'next' is defined using a tuple to ensure we report the correct assignability errors
    // in all places.
    next(...args: [] | [TNext]): IteratorResult<T, TReturn>;
    return?(value?: TReturn): IteratorResult<T, TReturn>;
    throw?(e?: any): IteratorResult<T, TReturn>;
}

// es2015.generator.d.ts
interface Generator<T = unknown, TReturn = any, TNext = unknown> 
    extends Iterator<T, TReturn, TNext> {
    // NOTE: 'next' is defined using a tuple to ensure we report the correct assignability errors
    // in all places.
    next(...args: [] | [TNext]): IteratorResult<T, TReturn>;
    return(value: TReturn): IteratorResult<T, TReturn>;
    throw(e: any): IteratorResult<T, TReturn>;
    [Symbol.iterator](): Generator<T, TReturn, TNext>;
}

// es2018.asynciterable.d.ts
interface AsyncIterator<T, TReturn = any, TNext = undefined> {
    // NOTE: 'next' is defined using a tuple to ensure we report the correct assignability errors
    // in all places.
    next(...args: [] | [TNext]): Promise<IteratorResult<T, TReturn>>;
    return?(value?: TReturn): Promise<IteratorResult<T, TReturn>>;
    throw?(e?: any): Promise<IteratorResult<T, TReturn>>;
}

// es2018.asyncgenerator.d.ts
interface AsyncGenerator<T = unknown, TReturn = any, TNext = unknown> 
    extends AsyncIterator<T, TReturn, TNext> {
    // NOTE: 'next' is defined using a tuple to ensure we report the correct assignability errors
    // in all places.
    next(...args: [] | [TNext]): Promise<IteratorResult<T, TReturn>>;
[...]

PR #58243, opened 2024-04-18
Add TReturn/TNext to Iterable et al

### LINKED ISSUE #33353: IterableIterator<T>.next().value is now any instead of T
**TypeScript Version:**  3.6.2

**Search Terms:** iterableiterator iterator

**Code**

```ts
declare function getIterator(): IterableIterator<number>;
const iter = getIterator()
const { value } = iter.next()
```

**Expected behavior:**
`value: number` (TS 3.5)

**Actual behavior:**
`value: any`

**Playground Link:**

NOTE: Playground is running TS 3.5 and exhibits "Expected behavior"
http://www.typescriptlang.org/play/#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXwHMQMBJDEGKDHGACgEoAueMiqAIwhFcupgB5UyALbsKAPgDcAKDB4AzhnhZyMeAF5CxHlRoNZCpQG94ANygRkCAL4blqgHSoQADwz6gA

**Related Issues:**

### LINKED ISSUE #43750: set.values().next() has wrong type
# Bug Report

Happens in both versions 4.2.3 and 4.3.0-beta

### ⏯ Playground Link

[Playground link with relevant code](https://www.typescriptlang.org/play?#code/MYewdgzgLgBBMF4ZgKYHcYGUVQDzQCcBLMAcwD4AKASgG4AoeiAOgDcBDAGwFcUIbmqAB5QBHHigZA)

### 💻 Code

```ts
const s = new Set<string>();
s.values().next().value;
```

### 🙁 Actual behavior

Type of `value` is `any`.

### 🙂 Expected behavior

Type of `value` should probably be `string | undefined` or something.

### LINKED ISSUE #52998: Built-in Symbol.iterator methods return IterableIterator<T> extending Iterator<T>, not Iterator<T, void>

rbuckton:
We avoided making this change in the past as it was very breaky, but at some point we need to address this discrepancy. Having incorrect types here is also causing problems with properly typing the Iterator Helpers proposal.

This also adds a new `BuiltinIteratorReturn` intrinsic type whose actual type is determined by the state of a new `strictBuiltinIteratorReturn` compiler option:
- `"strictBuiltinIteratorReturn": false` - `any` (emulates current behavior for `IterableIterator`)
- `"strictBuiltinIteratorReturn": true` - `undefined`

The `strictBuiltinIteratorReturn` is a `strict` option flag, meaning that it is enabled automatically when you set `"strict": true` in your `tsconfig.json`.

The new `BuiltinIteratorReturn` type is passed as the `TReturn` type argument for built-ins using `IterableIterator` or `AsyncIterableIterator` to enable stricter checks against the result of calling `next()` on the iterator.

Enabling `strictBuiltinIteratorReturn` results in a more accurate and type-safe return type for the `next()` methods of iterators produced by built-ins like `Array`, `Set`, `Map`, etc.:

```ts
// @strictBuiltinIteratorReturn: false
// NOTE: matches current behavior

const set = new Set(["a"]);
const result = set.keys().next();
//    ^?  result: IteratorResult<string, any>
const value = result.value;
//    ^?  value: any

if (!result.done) {
  const value = result.value;
  //    ^?  value: string
} else {
  const value = result.value;
  //    ^?  value: any
}
```
vs
```ts
// @strictBuiltinIteratorReturn: true

const set = new Set(["a"]);
const result = set.keys().next();
//    ^?  result: IteratorResult<string, undefined>
const value = result.value;
//    ^?  value: string | undefined

if (!result.done) {
  const value = result.value;
  //    ^?  value: string
} else {
  const value = result.value;
  //    ^?  value: undefined
}
```

Since this is a `strict` flag, there is a fair amount of existing code that will likely produce new compilation errors as a result of this change:

```ts
// @strict: true

function only<T>(set: Set<T>): T {
  if (set.size !== 1) throw new TypeError();
  return set.keys().next().value; // worked previously since result was `any`, but is now an error
}
```

This now fails as there is no correlation between `set.size` and the iterator produced by `keys()`, so the compiler is unaware that this constraint has been validated. If you are certain iterator will always yield at least one value, you can use a non-null assertion operator to strip off the `| undefined`:

```ts
// @strict: true

function only<T>(set: Set<T>): T {
  if (set.size !== 1) throw new TypeError();
  return set.keys().next().value!;
}
```

A follow-on PR to TypeScript-DOM-lib-generator can be found here: https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/1713

DefinitelyTyped breaks will be addressed by https://github.com/DefinitelyTyped/DefinitelyTyped/pull/69632

`fp-ts` breaks will be addressed by https://github.com/gcanti/fp-ts/pull/1949

`webpack` breaks will be addressed by https://github.com/webpack/webpack/pull/18591

Fixes #33353
Fixes #52998
Fixes #43750
Closes #56517
Related #58222


rbuckton:
@typescript-bot test top400
@typescript-bot perf test

rbuckton:
@typescript-bot run dt

rbuckton:
Most of the DT breaks are expected. A few aren't, but are probably correct.

rbuckton:
Take two after fixing a few issues with assignability checks for generators.

@typescript-bot test top400
@typescript-bot perf test
@typescript-bot run dt

rbuckton:
@typescript-bot run dt

rbuckton:
> Hey @rbuckton, the results of running the DT tests are ready.
> 
> There were interesting changes:
> 
> Changes are too big to display here, please check [the log](https://typescript.visualstudio.com/TypeScript/_build/results?buildId=161328&view=logs&j=275f1d19-1bd8-5591-b06b-07d489ea915a&t=40b1ee41-44d6-5bba-aa04-4b76a5c732e5).
> 
> [You can check the log here](https://typescript.visualstudio.com/TypeScript/_build/index?buildId=161328&_a=summary).

@jakebailey: Am I missing something? The bot says there were interesting changes but it links to a clean pipeline result.

jakebailey:
The log I'm seeing is not clean; the first link is https://typescript.visualstudio.com/TypeScript/_build/results?buildId=161328&view=logs&j=275f1d19-1bd8-5591-b06b-07d489ea915a&t=40b1ee41-44d6-5bba-aa04-4b76a5c732e5 which has loads of errors

rbuckton:
> The log I'm seeing is not clean; the first link is https://typescript.visualstudio.com/TypeScript/_build/results?buildId=161328&view=logs&j=275f1d19-1bd8-5591-b06b-07d489ea915a&t=40b1ee41-44d6-5bba-aa04-4b76a5c732e5 which has loads of errors

Ah, thanks. With everything showing green it was hard to pin down which of the jobs to look at.

jakebailey:
Yeah, it's always the last one that's joining all of the jobs. The builds generally don't fail unless something has gone catastrophically wrong, and success is "I posted a comment"

rbuckton:
The last commit should also cleanup a chunk of the DT failures as well.

@typescript-bot run dt

rbuckton:
@typescript-bot test top400
@typescript-bot perf test

jakebailey:
@typescript-bot perf test

rbuckton:
DT Failures:
- consumable-stream
  - Returns `{ done: true, value: "dummy" }` which is widened to `{ done: boolean, value: string }` which is not assignable to `IteratorResult<string, void>`.
  - The test uses `done: true`, which should only type the `return` type as `string`, not the `yield` type, but the consuming side expects to iterate over `string` values. This indicates the test is incorrect.
  - [ ] Fix: Rewrite test to be properly typed.
- consumable-stream/v1
  - Same issue as consumable-stream
  - [ ] Fix: Rewrite test to be properly typed.
- consumable-stream/v2
  - Same issue as consumable-stream
  - [ ] Fix: Rewrite test to be properly typed.
- symbol-tree
  - It looks like this does also return `T`, so the `implements` clause needs to be changed from `IterableIterator<T>` to `Iterator<T, T>`, though we could also just leave off the `implements` clause since it is unneeded.
  - [ ] Fix: Remove `implements` clause
- find-package-json
  - Inferred type of `IteratorResult` changed.
  - `FindIterator` type inherits from `IterableIterator<PackageWithPath>`
  - [ ] Fix: Make `FindIterator` inherit from `Iterator<PackageWithPath, void>` and manually implement `[Symbol.iterator]()`, and update `$ExpectType`.
- async-iterable-stream
  - Same issue as consumable-stream
  - [ ] Fix: Rewrite test to be properly typed.
- win-ca
  - `api` method returns `AsyncGenerator<T> | Generator<T>` and does not specify `TReturn` type.
  - It looks like `TReturn` should be `void`
  - [ ] Fix: Change return type to `AsyncGenerator<T, void> | Generator<T, void>` and update `$ExpectType`
- bresenham
  - `generator` return type is `Generator<Point>` but should be `Generator<Point, void>`
  - [ ] Fix: Retype `generator` return type.
- get-intrinsic
  - Many of the entries in the `Intrinsics` interface are typed using `Generator<any>`, etc.
  - [ ] Fix: Fill in missing type arguments or update `$ExpectType`
- es-get-iterator
  - Expected types are different
  - [ ] Fix: Update `$ExpectType` checks
- node
  - functions in stream/consumers take `AsyncIterable<any>` which now is incompatible.
  - `isGeneratorObject` returns `object is Generator`
  - [ ] Fix: Retype `isGeneratorObject` return type to `object is Generator<unknown, any, unknown>` and change default for `AsyncIterable`'s `TReturn` to `unknown`.
- node/v16
  - `isGeneratorObject` returns `object is Generator`
  - [ ] Fix: Retype return type to `object is Generator<unknown, any, unknown>`
- node/v18
  - `isGeneratorObject` returns `object is Generator`
  - [ ] Fix: Retype return type to `object is Generator<unknown, any, unknown>`
- sql.js
  - `StatementIterator` incorrectly typed as `Iterator<Statement>`
  - `StatementIteratorResult` incorrectly typed as `{ done: boolean, value: Statement }`
  - [ ] Fix: Retype to `Iterator<Statement, void>` and replace `StatementIteratorResult` with `IteratorResult<Statement>`
- bluebird
  - `coroutine` overloads take in `IterableIterator<any>` which is incompatible with the default for `IterableIterator`.
  - [ ] Fix: Default for `IterableIterator`'s `TReturn` should be `unknown`, not `void`, and all references to it in the libs should be explicit.
- es-abstract
  - Same issue as get-intrinsic
  - [ ] Fix: Fill in missing type arguments or update `$ExpectType`
- google-protobuf
  - `Map.Iterator` is not sufficiently typed.
  - [ ] Fix: Update types for `Map.Iterator`
- posterus
  - `fiber` function takes in `IterableIterator<any>` which is incompatible with the default for `IterableIterator`.
  - [ ] Fix: Default for `IterableIterator`'s `TReturn` should be `unknown`, not `void`, and all references to it in the libs should be explicit.
- regenerator-runtime
  - Expected type differs
  - [ ] Fix: Update `$ExpectedType`
  
I think I need to change the default for `TReturn` to be `unknown`, though that means I need to change all use sites of `Iterable` et all in the libs to provide an explicit `TReturn` to still fix #33353.

rbuckton:
@typescript-bot run dt

rbuckton:
@typescript-bot perf test
@typescript-bot test top400

src/compiler/checker.ts
```diff
@@ -1793,6 +1793,14 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             if (type === emptyGenericType) return undefined;
             return type;
         },
+        createTypeReference: (target, typeArguments) => {
+            if (target !== emptyGenericType && getObjectFlags(target) & ObjectFlags.ClassOrInterface) {
+                const interfaceType = target as InterfaceType;
+                if (some(interfaceType.typeParameters)) {
+                    return createTypeReference(interfaceType as GenericType, typeArguments);
```
weswigham, line 1800: It's up to the caller of `createTypeReference` to validate that the `typeArguments` have an appropriate count, and to fill missing values with defaults - should we not have a `fillMissingTypeArguments` and some count checks in here?

src/lib/dom.iterable.d.ts
```diff
@@ -1,30 +1,30 @@
 /// <reference lib="dom" />
 
 interface DOMTokenList {
-    [Symbol.iterator](): IterableIterator<string>;
+    [Symbol.iterator](): IterableIterator<string, void>;
```
weswigham, line 4: Presumably this change is going to be reflected in [the dom lib generator](https://github.com/microsoft/TypeScript-DOM-lib-generator), too?

tests/baselines/reference/YieldStarExpression4_es6.types
```diff
@@ -2,11 +2,12 @@
 
 === YieldStarExpression4_es6.ts ===
 function *g() {
->g : () => Generator<any, void, undefined>
->  : ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+>g : () => Generator<any, void, unknown>
```
weswigham, line 5: Is it weird to think that it's almost better for the type for an untyped `yield` result to be `void` instead of `unknown`? I guess it'd be funky seeing the `void` in the parameter list of the `.next` call, but it feels like if you wrote `function foo*() { return yield val }`, having `foo`'s `TReturn` be `void` instead of `unknown` would be kinda nice.

I dunno, probably not terribly important.

tests/baselines/reference/generatorTypeCheck62.errors.txt
```diff
@@ -0,0 +1,67 @@
...
+    }
+    
+    export interface State extends StrategicState {
+        foo: number;
+    }
+    
+    export const Nothing1: Strategy<State> = strategy("Nothing", function*(state: State) {
+                                                                 ~~~~~~~~
+!!! error TS2345: Argument of type '(state: State) => Generator<never, State, unknown>' is not assignable to parameter of type '(a: State) => IterableIterator<State, void>'.
+!!! error TS2345:   Call signature return types 'Generator<never, State, unknown>' and 'IterableIterator<State, void>' are incompatible.
```
weswigham, line 43: Aside: We should special-case the error messages for comparing a generator/iterable/iterator - this is a really big stack just to say one of the return, next, or yield types don't match.

src/compiler/checker.ts
weswigham: Alternatively, maybe just exposing a `getAnyAsyncIterator` might be more purpose-built for the quickfix, and we wouldn't have to harden it against misuse, since that's the one thing it's used for.

src/lib/dom.iterable.d.ts
jakebailey: This file isn't one of the generated ones, actually, so I don't think the generator has to change unless one of the `*.generated.*.d.ts` files are edited.

rbuckton: Yes, I plan to do that as well.

rbuckton: The generator will need to change so that we have the correct types on any DOM collections produced by the generator.

rbuckton: That ends up being a bit tricky if we're also shipping DOM types in a standalone package since older TS wouldn't support `TReturn` and `TNext`.

tests/baselines/reference/YieldStarExpression4_es6.types
rbuckton: Having `TNext` default to either `void` or `undefined` ended up being an even larger breaking change.

Renegade334:
Is this the right place for a discussion regarding the arity of `next()`?

The function signature introduced in #30790 means that when interacting with iterators manually, the compiler will always consider it valid if `next()` is called with zero-arity, even if TNext is of a type that cannot be void/undefined. For example:

```ts
function *g(): Generator<null, string, number> {
  const a = yield null; // const a: number
  return a.toString(16);
}

const it = g();
it.next();
it.next(); // TypeError: a is undefined
```

Admittedly a bit niche &ndash; the compiler already checks to ensure that iterators with positively-defined `TNext` types can't be used with built-in iteration algorithms that don't pass a parameter to `next()` (for...of, spread syntax etc.), so this specific error case only arises when interacting with iterators by hand.

bakkot:
> The compiler already checks to ensure that iterators with positively-defined TNext types can't be used with built-in iteration algorithms that don't pass a parameter to `next()` (for...of, spread syntax etc.), so this specific error case only arises when interacting with iterators by hand.

It only does that check for syntax. Most consumers of iterables are APIs, like `new Set(iterable)` etc, and this error will still happen there.

In an ideal world the type for `next` would be given as a function type rather than a method type, so that under `strictFunctionTypes` it would be checked contravariantly rather than bivariantly, but that's probably too breaky. (And maybe has other problems I'm not immediately seeing.)

src/lib/es2018.asynciterable.d.ts
```diff
@@ -9,17 +9,17 @@ interface SymbolConstructor {
     readonly asyncIterator: unique symbol;
 }
 
-interface AsyncIterator<T, TReturn = any, TNext = undefined> {
+interface AsyncIterator<T, TReturn = any, TNext = unknown> {
     // NOTE: 'next' is defined using a tuple to ensure we report the correct assignability errors in all places.
     next(...args: [] | [TNext]): Promise<IteratorResult<T, TReturn>>;
```
Renegade334, line 14: Is it worth taking the opportunity to label the tuple element here (and in AsyncIterator/Generator) to avoid `args_0` appearing in the computed function signature?

Renegade334:
(For clarification, the above comment was a response to some musings about the behaviour of `next()`, which I've subsequently forked to a new issue.)

src/lib/es2018.asynciterable.d.ts
rbuckton: I just looked into doing this, but it causes the number of types we produce to skyrocket in `unionAndIntersectionInference3.ts`, so it's probably better to leave it as is for now.

image

rbuckton:
@typescript-bot perf test
@typescript-bot run dt
@typescript-bot test top400

src/lib/es2020.bigint.d.ts
```diff
@@ -153,7 +153,7 @@ interface BigInt64Array {
     copyWithin(target: number, start: number, end?: number): this;
 
     /** Yields index, value pairs for every entry in the array. */
-    entries(): IterableIterator<[number, bigint]>;
+    entries(): IterableIterator<[number, bigint], void>;
```
Renegade334, line 156:
```suggestion
    entries(): IterableIterator<[number, bigint], BuiltinIteratorReturn>;
```

src/lib/es2020.bigint.d.ts
```diff
@@ -238,7 +238,7 @@ interface BigInt64Array {
     join(separator?: string): string;
 
     /** Yields each index in the array. */
-    keys(): IterableIterator<number>;
+    keys(): IterableIterator<number, void>;
```
Renegade334, line 241:
```suggestion
    keys(): IterableIterator<number, BuiltinIteratorReturn>;
```

src/lib/es2020.bigint.d.ts
```diff
@@ -360,9 +360,9 @@ interface BigInt64Array {
     valueOf(): BigInt64Array;
 
     /** Yields each value in the array. */
-    values(): IterableIterator<bigint>;
+    values(): IterableIterator<bigint, void>;
```
Renegade334, line 363:
```suggestion
    values(): IterableIterator<bigint, BuiltinIteratorReturn>;
```

src/lib/es2020.bigint.d.ts
```diff
@@ -360,9 +360,9 @@ interface BigInt64Array {
     valueOf(): BigInt64Array;
 
     /** Yields each value in the array. */
-    values(): IterableIterator<bigint>;
+    values(): IterableIterator<bigint, void>;
 
-    [Symbol.iterator](): IterableIterator<bigint>;
+    [Symbol.iterator](): IterableIterator<bigint, void>;
```
Renegade334, line 365:
```suggestion
    [Symbol.iterator](): IterableIterator<bigint, BuiltinIteratorReturn>;
```

src/lib/es2020.bigint.d.ts
```diff
@@ -425,7 +425,7 @@ interface BigUint64Array {
     copyWithin(target: number, start: number, end?: number): this;
 
     /** Yields index, value pairs for every entry in the array. */
-    entries(): IterableIterator<[number, bigint]>;
+    entries(): IterableIterator<[number, bigint], void>;
```
Renegade334, line 428:
```suggestion
    entries(): IterableIterator<[number, bigint], BuiltinIteratorReturn>;
```

src/lib/es2020.bigint.d.ts
```diff
@@ -510,7 +510,7 @@ interface BigUint64Array {
     join(separator?: string): string;
 
     /** Yields each index in the array. */
-    keys(): IterableIterator<number>;
+    keys(): IterableIterator<number, void>;
```
Renegade334, line 513:
```suggestion
    keys(): IterableIterator<number, BuiltinIteratorReturn>;
```

src/lib/es2020.bigint.d.ts
```diff
@@ -632,9 +632,9 @@ interface BigUint64Array {
     valueOf(): BigUint64Array;
 
     /** Yields each value in the array. */
-    values(): IterableIterator<bigint>;
+    values(): IterableIterator<bigint, void>;
```
Renegade334, line 635:
```suggestion
    values(): IterableIterator<bigint, BuiltinIteratorReturn>;
```

src/lib/es2020.bigint.d.ts
```diff
@@ -632,9 +632,9 @@ interface BigUint64Array {
     valueOf(): BigUint64Array;
 
     /** Yields each value in the array. */
-    values(): IterableIterator<bigint>;
+    values(): IterableIterator<bigint, void>;
 
-    [Symbol.iterator](): IterableIterator<bigint>;
+    [Symbol.iterator](): IterableIterator<bigint, void>;
```
Renegade334, line 637:
```suggestion
    [Symbol.iterator](): IterableIterator<bigint, BuiltinIteratorReturn>;
```

src/lib/es2020.string.d.ts
```diff
@@ -6,7 +6,7 @@ interface String {
      * containing the results of that search.
      * @param regexp A variable name or string literal containing the regular expression pattern and flags.
      */
-    matchAll(regexp: RegExp): IterableIterator<RegExpExecArray>;
+    matchAll(regexp: RegExp): IterableIterator<RegExpExecArray, void>;
```
Renegade334, line 9:
```suggestion
    matchAll(regexp: RegExp): IterableIterator<RegExpExecArray, BuiltinIteratorReturn>;
```

src/lib/es2020.symbol.wellknown.d.ts
```diff
@@ -15,5 +15,5 @@ interface RegExp {
      * containing the results of that search.
      * @param string A string to search within.
      */
-    [Symbol.matchAll](str: string): IterableIterator<RegExpMatchArray>;
+    [Symbol.matchAll](str: string): IterableIterator<RegExpMatchArray, void>;
```
Renegade334, line 18:
```suggestion
    [Symbol.matchAll](str: string): IterableIterator<RegExpMatchArray, BuiltinIteratorReturn>;
```

src/lib/es2022.intl.d.ts
```diff
@@ -37,7 +37,7 @@ declare namespace Intl {
         containing(codeUnitIndex?: number): SegmentData;
 
         /** Returns an iterator to iterate over the segments. */
-        [Symbol.iterator](): IterableIterator<SegmentData>;
+        [Symbol.iterator](): IterableIterator<SegmentData, void>;
```
Renegade334, line 40:
```suggestion
        [Symbol.iterator](): IterableIterator<SegmentData, BuiltinIteratorReturn>;
```

rbuckton:
@typescript-bot: pack this

src/lib/dom.iterable.d.ts
rbuckton: > Presumably this change is going to be reflected in [the dom lib generator](https://github.com/microsoft/TypeScript-DOM-lib-generator), too?

https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/1713

rbuckton:
@typescript-bot run dt
@typescript-bot test top400

jakebailey:
The new flag is going to be interesting for DT where we don't actually have strict enabled (just null checks, implicit any, SFC). Maybe it's time that we just did that?

rbuckton:
DefinitelyTyped breaks will be addressed by https://github.com/DefinitelyTyped/DefinitelyTyped/pull/69632

rbuckton:
@typescript-bot pack this

rbuckton:
@typescript-bot run dt
@typescript-bot test top400

rbuckton:
@typescript-bot run dt
@typescript-bot test top400

rbuckton:
Based on the last two runs, the following is a summary of the differences between `undefined` vs `void` as `TReturn` for built-in iterators across DT, our user tests, and the top 400 repos:

1. `undefined`
   1. ✔️ Can be forcibly ignored using postfix-`!`
   2. ❌ Requires an argument to call `iter.return()`.
      - See `graphql/graphql-js` in https://github.com/microsoft/TypeScript/pull/58243#issuecomment-2121081910
   3. ✔️ No problem with assignability to `T | undefined`.
   4. ❌ Assignability errors in interface implementations and subclasses.
      - See `effect` in https://github.com/microsoft/TypeScript/pull/58243#issuecomment-2120971326
2. `void`
   1. ✔️ Can be forcibly ignored using postfix-`!`
   2. ✔️ Doesn't require an argument to call `iter.return()`.
   3. ❌ `T | void` is not assignable to `T | undefined`.
      - See `backstage/backstage` in https://github.com/microsoft/TypeScript/pull/58243#issuecomment-2123353315
      - See `ionic-team/ionic-framework` in https://github.com/microsoft/TypeScript/pull/58243#issuecomment-2123353315

I think the assignability issues in (1.iv) are far more problematic than those in (2.iii), as those in (2.iii) can be addressed with postfix-`!`, while those in (1.iv) cannot. This leads me to believe that continuing to use `void` over `undefined` as the return type for built-in iterators is probably the correct direction.

Renegade334:
Would it be possible to get a fresh packaged build?

rbuckton:
@typescript-bot: pack this

Renegade334:
For what it's worth:

- ii. is related specifically to the generator fallback return type, rather than BuiltinIteratorReturn.
  - The `return()` argument being mandatory for non-void `TReturn` types only affects Generators. It doesn't affect normal Iterators, builtin or otherwise, since the argument to `return()` in the Iterator interface is defined as optional.
- It doesn't look like the assignability error in iv. is genuine.
  - Digging through [the pipeline output](https://typescript.visualstudio.com/TypeScript/_build/results?buildId=161815&view=logs&j=98775127-73cc-5696-36a2-d2fef8930518&t=d9d396e3-92f1-565f-9165-fe54ddb28c95&l=381), the number of errors from the compilation hasn't changed. What's happened is that the mismatched types causing these errors have changed from `Effect<void, unknown, unknown>` to `Effect<undefined, unknown, unknown>` due to the change in generator fallback return type, and this is being reported by the workflow as a "new" error.
  - The errors in question are related to missing internal imports when compiling from that particular tsconfig, and aren't actually related to iterators/generators.
  - It doesn't seem like there are any other assignability errors raised for that test run, other than those associated with `iter.next().value` possibly being undefined.

With that in mind, I wonder if it's worth considering a combination of the two approaches: keeping the generator body parsing behaviour as it is in 5.4 (with fallback return type `void`), but having BuiltinIteratorReturn as `undefined`. This would avoid the void-to-undefined assignability issues associated with iii., while also avoiding breaky behaviour associated with changing the TReturn of non-returning generator functions from void to non-void.

rbuckton:
> * ii. is related specifically to the generator fallback return type, rather than BuiltinIteratorReturn.
>   
>   * The `return()` argument being mandatory for non-void `TReturn` types is specific to Generators. It doesn't affect normal Iterators, since the argument to `return()` in the Iterator interface is defined as optional.

I'm really not certain they should differ. `Generator` and `IterableIterator` are essentially the same, we only added `Generator` to avoid this specific break.

> I wonder if it's worth considering a combination of the two approaches: keeping the generator body parsing behaviour as it is in 5.4 (with fallback return type `void`), but having BuiltinIteratorReturn as `undefined`. This would avoid the void-to-undefined assignability issues for builtin iterators specifically, while also avoiding breaky behaviour associated with changing the TReturn of non-returning generator functions from void to non-void.

One of the reasons for this change is to support the Iterator helpers proposal. One of the capabilities in that proposal is allowing users to subclass `Iterator` to provide a custom `next()` while gaining the benefits of the helpers:

```js
class MyIterator extends Iterator {
  next() {
    ...
  }
}
```

Since custom iterators can have a meaningful `TReturn`, we should ensure we are consistent with generators in the base case.

rbuckton:
> It doesn't look like the assignability error in iv. is genuine.
> * Digging through [the pipeline output](https://typescript.visualstudio.com/TypeScript/_build/results?buildId=161815&view=logs&j=98775127-73cc-5696-36a2-d2fef8930518&t=d9d396e3-92f1-565f-9165-fe54ddb28c95&l=381), the number of errors from the compilation hasn't changed. What's happened is that the mismatched types causing these errors have changed from `Effect<void, unknown, unknown>` to `Effect<undefined, unknown, unknown>` due to the change in generator fallback return type, and this is being reported by the workflow as a "new" error.

It is not just `effect`, there are also assignability errors for fp-ts, puppeteer, and webpack.

rbuckton:
@typescript-bot run dt
@typescript-bot test top400

Renegade334:
It looks like top400 generated an empty repo list on the last run, so it didn't actually test anything.

jakebailey:
Hm, that's concerning.

@typescript-bot test top400

Renegade334:
> ## [angular/angular-cli](https://github.com/angular/angular-cli)
> 
> **4 of 21 projects failed to build with the old tsc and were ignored**
> ### [tests/legacy-cli/tsconfig.json](https://github.com/angular/angular-cli/blob/79821006b81a50fa983b33652a445e6debe0bf83/tests/legacy-cli/tsconfig.json)
> 
>     * `error TS2416: Property 'forEach' in type 'PartiallyOrderedSet<T>' is not assignable to the same property in base type 'Set<T>'.`
>       
>       * [packages/angular_devkit/core/src/utils/partially-ordered-set.ts#L42](https://github.com/angular/angular-cli/blob/79821006b81a50fa983b33652a445e6debe0bf83/packages/angular_devkit/core/src/utils/partially-ordered-set.ts#L42)
> 
>     * `error TS2416: Property '[Symbol.iterator]' in type 'PartiallyOrderedSet<T>' is not assignable to the same property in base type 'Set<T>'.`
>       
>       * [packages/angular_devkit/core/src/utils/partially-ordered-set.ts#L133](https://github.com/angular/angular-cli/blob/79821006b81a50fa983b33652a445e6debe0bf83/packages/angular_devkit/core/src/utils/partially-ordered-set.ts#L133)
`Generator<T, void>` not being assignable to `Iterator<T>` in strict mode is potentially quite a headache.

rbuckton:
@typescript-bot run dt

rbuckton:
> > ## [angular/angular-cli](https://github.com/angular/angular-cli)
> > **4 of 21 projects failed to build with the old tsc and were ignored**
> > ### [tests/legacy-cli/tsconfig.json](https://github.com/angular/angular-cli/blob/79821006b81a50fa983b33652a445e6debe0bf83/tests/legacy-cli/tsconfig.json)
> 
> > ```
> > * `error TS2416: Property '[Symbol.iterator]' in type 'PartiallyOrderedSet<T>' is not assignable to the same property in base type 'Set<T>'.`
> >   
> >   * [packages/angular_devkit/core/src/utils/partially-ordered-set.ts#L133](https://github.com/angular/angular-cli/blob/79821006b81a50fa983b33652a445e6debe0bf83/packages/angular_devkit/core/src/utils/partially-ordered-set.ts#L133)
> > ```
> 
> `Generator<T, void>` not being assignable to `Iterator<T, BuiltinIteratorReturn>` is potentially a headache.

Potentially, but maybe accurate. `void` is a bit of a mixed bag as, while we can generally assume a `void` function returns `undefined` for first-order application (e.g., `f()`), we cannot assume a `void` function returns `undefined` when it is passed as a callback (e.g., `f(cb)`) as TypeScript allows any function signature return type to be assignable to `void`:

```ts
// cannot return any result with a type that isn't `void`
function f(): void {
  // this would be a compile time error:
  // return 1;

  // but this is ok
  return; // ok
}

function g(cb: () => void): void {
  // and so is this:
  return cb(); // returns `void`, which is assignable to `void`
}

const result = g(() => 1); // ok, `() => number` is assignable to `() => void`
typeof result; // "number", not "undefined"
```

Therefore, a `Generator<T, void>` isn't actually assignable to an `Iterator<T, undefined>`, since an `Iterator<T, undefined>` explicitly states that it returns `undefined`, while a `Generator<T, void>` states that it could return _anything_, but you shouldn't use the result.

rbuckton:
@typescript-bot test top400

jakebailey:
The PR is large so I'm having trouble finding it, but does this still work out with `isolatedDeclarations`/`noCheck`? Specifically the part where we reprint things differently. We can't really rely on the checker for that stuff so hoping it's alright and not going to cause problems externally.

(Otherwise, I'm not 100% certain I'm the right person to approve this in whole)

rbuckton:
I've added some tests for `isolatedDeclarations` to verify the expected outputs. This mostly effects inferred types though, so I'm not sure `isolatedDeclarations` would be impacted.

src/testRunner/compilerRunner.ts
```diff
@@ -127,44 +127,32 @@ export class CompilerBaselineRunner extends RunnerBase {
...
-        "exactOptionalPropertyTypes",
-        "importHelpers",
-        "importHelpers",
-        "isolatedModules",
-        "jsx",
-        "module",
-        "moduleDetection",
-        "moduleResolution",
+        // implicit variations from defined options
+        ...ts.optionDeclarations
```
jakebailey, line 131: Thank you for doing this, I was getting really tired of adding new flags and then wondering why I couldn't vary tests on them

src/compiler/checker.ts
```diff
@@ -12953,6 +12984,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     links.instantiations = new Map<string, Type>();
                     links.instantiations.set(getTypeListId(typeParameters), type);
                 }
+                if (type === intrinsicMarkerType && symbol.escapedName === "BuiltinIteratorReturn") {
+                    type = strictBuiltinIteratorReturn ? undefinedType : anyType;
+                }
```
jakebailey, line 12989: Can this be done in the same way as `IntrinsicTypeKind` like `IntrinsicTypeKind.NoInfer`, rather than hardcoding the name in a couple of places?

rbuckton: No. `IntrinsicTypeKind.NoInfer` (and all of the other `IntrinsicTypeKind` entries) indicate a generic type mapping, e.g. `type NoInfer<T> = intrinsic` maps `T` to something else, which in that instance is a substitution type. Since `BuiltinIteratorReturn` is not generic it isn't picked up by `getTypeAliasInstantiation`, so never hits those code paths.

rbuckton: If anything, it would exist in its own map of non-generic `intrinsic` types, but since it would be the only entry there's no need for a map yet.

src/lib/es2018.asynciterable.d.ts
jakebailey: Just to close the loop on this one, #59045 fixed this yes?

jakebailey:
I believe this is what we decided on in the meeting and seems sound; I think that all that's left here is just to run perf on the final state (I don't think its been done in a bit, just the extended tests).

At least, once ADO is back online: https://status.dev.azure.com/

rbuckton:
@typescript-bot perf test

rbuckton:
The perf shouldn't have changed since the only real changes since it was last run were swapping around what type to use when the flag is enabled, but I'll try it again when the AzDO outage is over

rbuckton:
@typescript-bot perf test

rbuckton:
@jakebailey: It looks like the perf run was nominal.

src/harness/collectionsImpl.ts
```diff
@@ -132,6 +132,7 @@ export class SortedMap<K, V> {
                 this._copyOnWrite = false;
             }
         }
+        return undefined;
```
sandersn, line 135: will all generators need to `return undefined` now? do you have an estimate of how much code that breaks?

src/compiler/checker.ts
```diff
@@ -6877,7 +6876,40 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     }
                     let typeArgumentNodes: readonly TypeNode[] | undefined;
                     if (typeArguments.length > 0) {
-                        const typeParameterCount = (type.target.typeParameters || emptyArray).length;
+                        let typeParameterCount: number;
```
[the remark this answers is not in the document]
rbuckton: Is that necessary? This is definitely assigned in both branches of the `if`, so we are either potentially assigning it to `0` twice, or potentially assigning it to `0` and then to its actual value depending on which branch we take.

jakebailey:
I tested this on the most iterator-y codebase I know (effect) and it only caused one break (positively), so that's good.

src/compiler/checker.ts
```diff
@@ -6877,7 +6876,40 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     }
                     let typeArgumentNodes: readonly TypeNode[] | undefined;
                     if (typeArguments.length > 0) {
-                        const typeParameterCount = (type.target.typeParameters || emptyArray).length;
+                        let typeParameterCount: number;
+                        if (type.target.typeParameters) {
+                            typeParameterCount = Math.min(type.target.typeParameters.length, typeArguments.length);
+
+                            // Maybe we should do this for more types, but for now we only elide type arguments that are
```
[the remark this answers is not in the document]
rbuckton: Potentially any type since this simplifies the emit output to avoid repeating type arguments for type parameters with defaults, but rather than take that on in this PR I focused on this subset to improve back compat.

src/compiler/checker.ts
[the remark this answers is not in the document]
sandersn: I thought it was easier to read to initialise once and drop the `else` branch.

src/compiler/checker.ts
```diff
@@ -6877,7 +6876,40 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
                     if (typeArguments.length > 0) {
-                        const typeParameterCount = (type.target.typeParameters || emptyArray).length;
+                        let typeParameterCount: number;
+                        if (type.target.typeParameters) {
+                            typeParameterCount = Math.min(type.target.typeParameters.length, typeArguments.length);
+
+                            // Maybe we should do this for more types, but for now we only elide type arguments that are
+                            // identical to their associated type parameters' defaults for `Iterable`, `IterableIterator`,
+                            // `AsyncIterable`, and `AsyncIterableIterator` to provide backwards-compatible .d.ts emit due
+                            // to each now having three type parameters instead of only one.
```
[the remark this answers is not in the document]
jakebailey: Realistically, this is all about what the type "looks" like, right? Or are we really needing to depend on this emit still emitting the "old thing" to continue to work right for downstream users of d.ts files?

src/harness/collectionsImpl.ts
sandersn: from reading type baselines, it looks like their return type changes to `any` instead of `undefined`

src/compiler/checker.ts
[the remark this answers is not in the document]
rbuckton: Consider an application that is currently built using TS 5.5 that depends on a library that is also being built using TS 5.5. The library chooses to upgrade its version of TypeScript to 5.6, and the new emit produces `Iterable<T, any, any>` where it previously produced `Iterable<T>`. If the library makes no other code changes and publishes their new build, the application cannot take the update without *also* updating TypeScript because the expected arity of `Iterable` has changed.
The library can't just ship a forward declaration of `interface Iterable<T, TReturn = any, TNext = any> {}`, because our checker performs arity checks on built-ins, and the application might not be running with `skipLibCheck`, so one way or another they would have errors they cannot silence without possibly giving up the dependency verification that `skipLibChecks: false` offers.

src/compiler/checker.ts
```diff
@@ -6877,7 +6876,40 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+                                isReferenceToType(type, getGlobalIterableIteratorType(/*reportErrors*/ false)) ||
+                                isReferenceToType(type, getGlobalAsyncIterableType(/*reportErrors*/ false)) ||
+                                isReferenceToType(type, getGlobalAsyncIterableIteratorType(/*reportErrors*/ false))
+                            ) {
+                                if (
+                                    !type.node || !isTypeReferenceNode(type.node) || !type.node.typeArguments ||
+                                    type.node.typeArguments.length < typeParameterCount
+                                ) {
+                                    while (typeParameterCount > 0) {
+                                        const typeArgument = typeArguments[typeParameterCount - 1];
```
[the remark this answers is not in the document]
rbuckton: type argument inference

src/harness/collectionsImpl.ts
rbuckton: This is specifically because this is an implementation of the native `Map<K, V>` interface, which is now specified to return `IterableIterator<T, BuiltinIteratorReturn>` (which is `IterableIterator<T, undefined>` under the new flag).

rbuckton:
> I tested this on the most iterator-y codebase I know (effect) and it only caused one break (positively), so that's good.

In all of the PRs I've created so far for top libraries, it's usually only been one or two cases where they were (possibly unknowingly) depending on the result of `set.next().value` being typed as `any` by default.

sandersn (approved):
I still have a couple of background questions. I think this PR is worth trying in the beta to see if it's too breaky. But all the first-order breaks look good, and the higher-order ones are, I bet, rare. (like, just fpt-ts :)

PR #58252, opened 2024-04-19
Add try catch and defer.reject for checking sys tests

sheetalkamat:


src/testRunner/unittests/sys/symlinkWatching.ts
```diff
@@ -45,10 +45,15 @@ describe("unittests:: sys:: symlinkWatching::", () => {
...
-                            assert(actual === undefined || actual === modifiedTimeToString(sys.getModifiedTime!(file)));
+                            try {
+                                assert.equal(fileName, toWatch);
+                                assert.equal(eventKind, ts.FileWatcherEventKind.Changed);
+                                const actual = modifiedTimeToString(modifiedTime);
+                                assert(actual === undefined || actual === modifiedTimeToString(sys.getModifiedTime!(file)));
+                            }
+                            catch (e) {
+                                result.deferred.reject(e);
+                            }
```
jakebailey, line 56: I think this may need a return, otherwise it will fall through to the resolve, right?

jakebailey (approved):

src/testRunner/unittests/sys/symlinkWatching.ts
fatcerberus: (note: resolving a rejected promise is a noop)

Issue #58214, opened 2024-04-16
ModuleResolution Node16/NodeNext fails without extension starting on 5.5.0-dev.20240413


Noriller:
### 🔎 Search Terms

moduleResolution node16 nodeNext import extension

### 🕗 Version & Regression Information

- This changed between versions 5.5.0-dev.20240413 and 5.5.0-dev.20240415 (last one checked)

### 💻 Code

```ts
type Something1 = typeof import('@some/package/path/to/file');
type Something2 = typeof import('@some/package/path/to/file.js');
```

### 🙁 Actual behavior

Something1 resolves to `any`
Something2 resolves to `Something`

### 🙂 Expected behavior

Something1 resolves to `Something`
Something2 resolves to `Something`

### Additional information about the issue

This happens on the latest nightly versions and on moduleResolution Node16/NodeNext.

From what I could get, it started on 5.5.0-dev.20240413.

RyanCavanaugh:
We need a complete repro here; the package.json of `@some/package`, the importing package, and the tsconfig for the importing file all matter

Noriller:
https://stackblitz.com/~/github.com/Noriller/ts-issue-58214

I hope this one helps.

In the `test.d.ts` with the current `typescript@next` and same `jsconfig` (its a legacy project, using mostly jsdocs, but with the problem in a `d.ts` file)

If changing ts version with:

` npm i -D typescript@5.5.0-dev.20240412`

then both work

jakebailey:
This bisects to #57896. @andrewbranch

(I'm not an expert, but it seems strange that you'd need an extension when the d.ts file is effectively CJS, importing another CJS file, but importing via a JS file is already funky.)

andrewbranch:
Guessing I accidentally made ImportTypes look like an ESM-mode source location, when they were previously kept CJS-mode for backward compatibility reasons. Will take a look. Thanks!

andrewbranch:
Your jsconfig.json is invalid:

> Option 'module' must be set to 'Node16' when option 'moduleResolution' is set to 'Node16'

`module` is defaulting to es2020 because of `target`, which is causing everything to resolve under ESM rules (since everything will emit as ESM). I’d argue this is probably _more_ logically consistent than what we used to do, where the module resolution implementation made a lot more assumptions about module format and emit.

[...]

PR #58254, opened 2024-04-19
Report positionless tsconfig option errors on `compilerOptions` key

andrewbranch:
Any program error that is talking about a compiler option but can’t find an option name/value to report the error on now reports the error on the `compilerOptions` key (if present) instead of positionlessly in the file. I noticed while investigating #58214 that these positionless errors are very easy to miss in the editor.

Before:

image

After:

image

A common one that’s still positionless is “Cannot write file 'foo.js' because it would overwrite input file.” We could consider moving that to `compilerOptions` too since the correct fix is to add either `outDir` or `noEmit` 99% of the time, along with suggesting that in the error message. Anyone have opinions on that?

### DIFF
--- a/src/compiler/program.ts
+++ b/src/compiler/program.ts
@@ -264,6 +264,7 @@ import {
     projectReferenceIsEqualTo,
     PropertyAssignment,
     PropertyDeclaration,
+    PropertyName,
     ReferencedFile,
     removeFileExtension,
     removePrefix,
@@ -1618,6 +1619,7 @@ export function createProgram(rootNamesOrOptions: readonly string[] | CreateProg
     // Map storing if there is emit blocking diagnostics for given input
     const hasEmitBlockingDiagnostics = new Map<string, boolean>();
     let _compilerOptionsObjectLiteralSyntax: ObjectLiteralExpression | false | undefined;
+    let _compilerOptionsKeySyntax: PropertyName | false | undefined;
 
     let moduleResolutionCache: ModuleResolutionCache | undefined;
     let actualResolveModuleNamesWorker: (
@@ -4400,7 +4402,7 @@ export function createProgram(rootNamesOrOptions: readonly string[] | CreateProg
         }
 
         if (options.checkJs && !getAllowJSCompilerOption(options)) {
-            programDiagnostics.add(createCompilerDiagnostic(Diagnostics.Option_0_cannot_be_specified_without_specifying_option_1, "checkJs", "allowJs"));
+            createCompilerOptionsDiagnostic(Diagnostics.Option_0_cannot_be_specified_without_specifying_option_1, "checkJs", "allowJs");
         }
 
         if (options.emitDeclarationOnly) {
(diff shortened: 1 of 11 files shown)


tests/baselines/reference/resolutionCandidateFromPackageJsonField2(moduleresolution=bundler).errors.txt
```diff
@@ -1,11 +1,12 @@
...
 test.ts(1,19): error TS5097: An import path can only end with a '.ts' extension when 'allowImportingTsExtensions' is enabled.
 
 
-!!! error TS5095: Option 'bundler' can only be used when 'module' is set to 'preserve' or to 'es2015' or later.
-==== tsconfig.json (0 errors) ====
+==== tsconfig.json (1 errors) ====
     {
         "compilerOptions": {
+        ~~~~~~~~~~~~~~~~~
+!!! error TS5095: Option 'bundler' can only be used when 'module' is set to 'preserve' or to 'es2015' or later.
```
andrewbranch, line 9: Note that this test is simulating a combination of tsconfig options and CLI flags

src/compiler/program.ts
```diff
@@ -4400,7 +4402,7 @@ export function createProgram(rootNamesOrOptions: readonly string[] | CreateProg
         }
 
         if (options.checkJs && !getAllowJSCompilerOption(options)) {
-            programDiagnostics.add(createCompilerDiagnostic(Diagnostics.Option_0_cannot_be_specified_without_specifying_option_1, "checkJs", "allowJs"));
+            createCompilerOptionsDiagnostic(Diagnostics.Option_0_cannot_be_specified_without_specifying_option_1, "checkJs", "allowJs");
```
sheetalkamat, line 4405: This can go on location of "checkJs"  no?

tests/baselines/reference/resolutionCandidateFromPackageJsonField2(moduleresolution=bundler).errors.txt
jakebailey, line 9: This looks like something that would be better reported on the bundler option, except it's not here at all? I guess since it's passed in as a twoslash option, but this does make me wonder if we should be even more specific when possible... (But I do think this PR is better than the status quo anyway)

andrewbranch: See my comment above

jakebailey: Ohh

src/compiler/program.ts
```diff
@@ -4914,6 +4933,17 @@ export function createProgram(rootNamesOrOptions: readonly string[] | CreateProg
         return _compilerOptionsObjectLiteralSyntax || undefined;
     }
 
+    function getCompilerOptionsKeySyntax() {
```
sheetalkamat, line 4936: Can this method (so cached result) be used in other methods that get option of compiler Options  `getCompilerOptionsObjectLiteralSyntax`

sheetalkamat (changes requested):

sheetalkamat (approved):

RyanCavanaugh (approved):

PR #58260, opened 2024-04-19
Add a code fixer for --isolatedDeclarations errors

h-joo:
Add a code-fixer to fix type errors for --isolatedDeclarations

This is built on top of the errors added in PR #58201.

This code fixer fixes the following patterns.
    1. Adding missing types in places that type can be explicitly annotated.
    2. Adding satisfies with type assertion on a value.
    3. Create a namespace for expando functions and declare the property on the namespace.
    4. Factor out expressions in heritage clauses (class B extends <expression>), give them separate names and export them.


jakebailey:
Seems like this may need a rebase now that the base PR is in.

h-joo:
> @h-joo please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

@microsoft-github-policy-service agree company="Google"

src/services/codeFixProvider.ts
```diff
@@ -127,6 +127,7 @@ function getDiagnostics({ program, sourceFile, cancellationToken }: CodeFixConte
     return [
         ...program.getSemanticDiagnostics(sourceFile, cancellationToken),
         ...program.getSyntacticDiagnostics(sourceFile, cancellationToken),
+        ...program.getDeclarationDiagnostics(sourceFile, cancellationToken),
```
jakebailey, line 130: Curious how much this slows down quick fixes; hopefully not too much.

jakebailey: Could we conditionally call this based on the compilerOptions isolatedDeclarations setting?

DanielRosenwasser: I was going to suggest the same - we should probably only do this depending on settings.

jakebailey: That being said, I don't think it actually matters; I'm guessing this will be cached given you'd already have declarations enabled and see the diagnostics in the editor.

h-joo: applied the suggestion

jakebailey: I actually don't think you need to do that; it looks like the results are cached anyway. Though reading other places that call this function, it does make me wonder if we should just be merging the idea of a declaration error into semantic errors...

h-joo: Ah, okay. I'll wait for more feedback on your idea - or do you think it's rather something that's orthogonal to this PR?

jakebailey: Yeah, I'd say it's orthogonal; note how the LS does this: https://github.com/microsoft/TypeScript/blob/21b5c9624c0765e6baea1f21a26531aac4824a41/src/services/services.ts#L2038

The relevant thing is https://github.com/microsoft/TypeScript/blob/21b5c9624c0765e6baea1f21a26531aac4824a41/src/compiler/program.ts#L2884 and https://github.com/microsoft/TypeScript/blob/21b5c9624c0765e6baea1f21a26531aac4824a41/src/compiler/program.ts#L3281 given this implies that it's already cached. However, it looks like everyone conditionally calls this based on whether or not declaration emit is enabled, which seems to imply that leaving it conditional may be right (or conditionalize it on --declararions). But clearly we haven't had any past declaration emit quick fixes.

So... I think the state of the PR (checking ahead of time) is actually correct. But you can see why I'm thinking that conditionalizing everywhere is probably no longer super clear.

jakebailey:
I'm still going through the test cases, but one thing I'm not seeing are test cases for decorated classes/functions/etc. Even if that's not implemented yet, it may be worth adding some just to make sure they are covered.

jakebailey (commented):
Just finishing up reading the tests; I think overall I wonder if we need to be this complete, e.g. if it's worth handling cases like mixins, destructurings, and so on, which have very complicated conversions.

MichaelMitchell-at:
Is it expected for this PR to be sufficient to work with Quick Fixes in VSCode, or with [ts-fix](https://github.com/microsoft/ts-fix)? I've tried both and was not able to see any fix suggestions despite successfully getting isolatedDeclarations errors?

tests/cases/fourslash/codeFixMissingTypeAnnotationOnExports7.ts
```diff
@@ -0,0 +1,16 @@
...
+////export const c = { foo: foo() };
+
+verify.codeFix({
+    description: `Add annotation of type '{ foo: number[]; }'`,
+    index: 0,
+    newFileContent:
+`function foo(): number[] { return [42]; }
+export const c: {
+    foo: number[];
+} = { foo: foo() };`,
```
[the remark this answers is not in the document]
h-joo: > At some level, I wish we just came up with a named type for this, sorta like extract. This example is simple but I'm thinking a bit ahead towards people returning large objects and then hitting the quick fix and getting a one-off inline type.

I do agree with you, and I think we can offer that as an additional fix to extract that as a named type, but also I imagine naming a type is horrendously difficult for a code fixer to do, but the users can do a post-manual-fix after applying the fixer. e.g. `type cType= {foo:number[]} export const c: cType;`

> That does beg the question of how much code is being reused from other places. (Though I get the impression that using typeToAutoImportableTypeNode implies that we are already reusing what we can?)

I didn't quite understand what you mean, could you elaborate? Do you mean how much the code to create a type node is being used in other places?

tests/cases/fourslash/codeFixMissingTypeAnnotationOnExports8.ts
```diff
@@ -0,0 +1,18 @@
...
+verify.codeFixAvailable([
+    { description: "Add return type 'number'" },
+]);
+
+verify.codeFix({
+    description: "Add return type 'number'",
+    index: 0,
+    newFileContent:
+`function foo() {return 42;}
+export const g = function (): number { return foo(); };`,
```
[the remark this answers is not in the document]
h-joo: So we did add support for these AFAIK, this isn't an error under --isolatedDeclarations (I test it and it confirms that it isn't) . It errors by first looking  at the variable, if there's no type annotation - goes into the initializer and sees if it's a function type and check all places have appropriate type annotation.

tests/cases/fourslash/codeFixMissingTypeAnnotationOnExports14.ts
```diff
@@ -0,0 +1,21 @@
...
+    index: 0,
+    newFileContent:
+`function foo() {
+    return { x: 1, y: 1};
+}
+const dest = foo();
+export const x: number = dest.x;
+const temp = dest.y;
+export const y: number = temp === undefined ? 0 : dest.y;
+export const z = 42;`});
```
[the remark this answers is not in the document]
h-joo: I'm also not a great fan of this 😅, I guess this pattern is not very common as well. If I recall the conversation with Titian, this was also mainly for completeness.

tests/cases/fourslash/codeFixMissingTypeAnnotationOnExports18.ts
```diff
@@ -0,0 +1,22 @@
...
+    { description: "Add return type 'number'" },
+]);
+
+verify.codeFix({
+    description: "Add return type 'number'",
+    index: 0,
+    newFileContent:
+`function foo() { return 42; }
+export class A {
+    readonly a = (): number => foo();
```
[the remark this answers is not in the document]
h-joo: We can probably also do that . Please let me know if you think the same way after my comment on the [thread about initializers](https://github.com/microsoft/TypeScript/pull/58260#discussion_r1575323309)

tests/cases/fourslash/codeFixMissingTypeAnnotationOnExports21-params-and-return.ts
```diff
@@ -0,0 +1,52 @@
...
+    applyChanges: true,
+    newFileContent:
+`/**
+ * Test
+ */
+export function foo(): number { return 0; }
+/**
+* Docs
+*/
+export const bar = (a = foo()): number =>
```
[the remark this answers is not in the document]
h-joo: It does! please see the test case below on line 49.

tests/cases/fourslash/codeFixMissingTypeAnnotationOnExports23-heritage-formatting.ts
```diff
@@ -0,0 +1,41 @@
...
+// @isolatedDeclarations: true
+// @declaration: true
+////function mixin<T extends new (...a: any) => any>(ctor: T): T {
+////    return ctor;
+////}
+////class Point2D { x = 0; y = 0; }
+////interface I{}
+////export class Point3D extends
+////    /** Base class */
+////    mixin(Point2D)
```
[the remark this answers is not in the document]
h-joo: I just searched and it's quite rare in Google, I think @dragomirtitian wanted this for completeness. @dragomirtitian would you have an opinion here?

tests/cases/fourslash/codeFixMissingTypeAnnotationOnExports30-inline-import.ts
```diff
@@ -0,0 +1,28 @@
...
+
+goTo.file("/code.ts");
+
+verify.codeFix({
+    description: "Add satisfies and an inline type assertion with 'Person'",
+    index: 1,
+    newFileContent:
+`import { getPerson, Person } from "./person-code";
+export const exp = {
+  person: getPerson() satisfies Person as Person
```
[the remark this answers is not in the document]
h-joo: No, the code fixer provides multiple fixes for the same error and this is just one of them. This single test case is testing against the fix which adds inline type annotation next to the property.

tests/cases/fourslash/codeFixMissingTypeAnnotationOnExports32-inline-short-hand.ts
```diff
@@ -0,0 +1,29 @@
...
+};`
+});
+
+verify.codeFix({
+  description: "Add satisfies and an inline type assertion with 'typeof x'",
+  index: 2,
+  newFileContent:
+`const x = 1;
+export default {
+  x: x as typeof x
```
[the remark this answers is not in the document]
h-joo: Yes, it preserves the variable `x` in the declaration file.

h-joo:
> Is it expected for this PR to be sufficient to work with Quick Fixes in VSCode, or with [ts-fix](https://github.com/microsoft/ts-fix)? I've tried both with this branch and was not able to see any fix suggestions despite successfully getting isolatedDeclarations errors.

Good point, I tested this once with a locally built tsd, but I can't reproduce the result with the current state. We intend to support this in vscode as a quick fix and anywhere that uses tsd. Thank you for bringing this up, I'll fix it before submission.

tests/cases/fourslash/codeFixMissingTypeAnnotationOnExports7.ts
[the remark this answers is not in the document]
jakebailey: > I didn't quite understand what you mean, could you elaborate? Do you mean how much the code to create a type node is being used in other places?

I just meant that we have existing code fixes for inferring types and extracting code with specific names, and these new fixes are pretty similar.

tests/cases/fourslash/codeFixMissingTypeAnnotationOnExports8.ts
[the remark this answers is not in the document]
jakebailey: Hm, I guess I was going off of the list of unfinished things including initializers with functions...

jakebailey:
Theoretically ts-fix should "just work", so that's definitely odd.

h-joo:
> Is it expected for this PR to be sufficient to work with Quick Fixes in VSCode, or with [ts-fix](https://github.com/microsoft/ts-fix)? I've tried both with this branch and was not able to see any fix suggestions despite successfully getting isolatedDeclarations errors.

Actually, I'm seeing the code fixes. I'm doing development on a cloud machine so I'm using a browser to access vscode in my cloud machine, but it seems to work when I'm running this locally. 

- From the web interface (non-local)
image

- From a local vscode run
image

tests/cases/fourslash/codeFixMissingTypeAnnotationOnExports21-params-and-return.ts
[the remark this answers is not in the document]
jakebailey: Oh, huh. How hard would it be to offer only one annotation per declaration which does both?

tests/cases/fourslash/codeFixMissingTypeAnnotationOnExports36-conditional-releative.ts
```diff
@@ -0,0 +1,31 @@
...
+      "description": "Add satisfies and an inline type assertion with 'typeof A | typeof B'"
+    }
+])
+verify.codeFix({
+    description: "Add satisfies and an inline type assertion with 'typeof A | typeof B'" ,
+    index: 3,
+    newFileContent:
+`const A = "A"
+const B = "B"
+export const AB = (Math.random() ? A : B) satisfies typeof A | typeof B as typeof A | typeof B;`
```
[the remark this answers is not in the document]
h-joo: Actually [it changes the type](https://www.typescriptlang.org/play?ts=5.4.5#code/MYewdgzgLgBAgjAvDARHFAoUlYCEmq6YCmAHgA4gBOs208+yAFALICGUAFgHRVtgATEAFsmAShgB+eDABcMXBIgcAlhABmK4hBhQAnuWIh1MgD66DRk-jY79h42YsPrAbgxlKNGHVhxcAEwErBw8fIIi4lIy8oquQA), I guess this will break downstream code.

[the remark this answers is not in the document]
jakebailey: Yeah, which can be fine, though maybe that means the fix needs to offer both unwidened and widened types if they differ.

tests/cases/fourslash/codeFixMissingTypeAnnotationOnExports21-params-and-return.ts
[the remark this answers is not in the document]
h-joo: I guess it'll require quite a code change as these are two separate diagnostics.

src/services/codeFixProvider.ts
```diff
@@ -124,9 +124,15 @@ export function eachDiagnostic(context: CodeFixAllContext, errorCodes: readonly
 }
 
 function getDiagnostics({ program, sourceFile, cancellationToken }: CodeFixContextBase) {
-    return [
+    const diagnostics = [
```
MichaelMitchell-at, line 127: Yeah I'm not sure why I don't see codefixes. Adding some console logs, I'm surprised to see that this code path isn't reached when running with ts-fix.

tests/cases/fourslash/codeFixMissingTypeAnnotationOnExports23-heritage-formatting.ts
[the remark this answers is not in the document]
dragomirtitian: We don't use this either. But we were trying to cover as many places as possible. My basic idea is that if previously TS would produce a declaration for it we should be able to take that type and put it in the ts file.

tests/cases/fourslash/codeFixMissingTypeAnnotationOnExports32-inline-short-hand.ts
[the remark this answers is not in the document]
dragomirtitian: @jakebailey declaration emit was already spart enough for this fortunately :)

tests/cases/fourslash/codeFixMissingTypeAnnotationOnExports36-conditional-releative.ts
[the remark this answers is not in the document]
h-joo: Done. Although I have two issues that I need to note.

1. I needed to expose `TypeChecker.getWidenedLiteralType` as an internal API to imitate what the type checker is doing during d.ts emit.
2. In the code fixer it's querying for the type of the same node twice, and it seems to affect the freshness to that manner, and results in `isFreshLiteralType` returning `False` . For the reasons that I don't understand, `getWidenedLiteralType` is conditionally widening a literal type only if it's fresh. So the  updated test file 36-conditional-relative.ts  an inline annotation with `string` is not being added. Would you know what's the best approach to take here?

h-joo:
> I'm still going through the test cases, but one thing I'm not seeing are test cases for decorated classes/functions/etc. Even if that's not implemented yet, it may be worth adding some just to make sure they are covered.

Added test cases for experimental / non-experimental decorators, I think nothing should be affected as the decorators shoudn't change the types. It seems to work well from what I can see.

src/services/codeFixProvider.ts
jakebailey: It looks like ts-fix does this: https://github.com/microsoft/ts-fix/blob/main/src/index.ts#L341

So it's not using this path at all.

src/compiler/checker.ts
```diff
@@ -1605,6 +1605,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         getBaseTypes,
         getBaseTypeOfLiteralType,
         getWidenedType,
+        getWidenedLiteralType,
```
jakebailey, line 1608: I tried to see if this could be removed and just use `getWidenedType`, but that didn't quite work, which is a shame.

src/services/codefixes/fixMissingTypeAnnotationOnExports.ts
```diff
@@ -0,0 +1,1192 @@
...
+                ),
+            );
+        }
+        else {
+            Debug.assertNever(targetNode);
+        }
+        return [Diagnostics.Add_satisfies_and_an_inline_type_assertion_with_0, typeToStringForDiag(typeNode)];
+    }
+
+    function suggestVariableName(node: Node) {
```
andrewbranch, line 410: I'm a little skeptical of spending this many lines generating a name beyond a single level of valid-identifier property assignment when the user will almost certainly replace it after extracting. There's a one-liner best effort in `extractSymbol.ts` that falls back to `getUniqueName("newLocal")` that I think would suffice here. Maybe you could extract that to a common utility, improving it if needed (but only insofar as the improvement results in a high-confidence suggestion).

It would be breat if we could automatically invoke Rename after the codefix like we do in the Extract Symbol refactor, but I don't think our current protocol allows for that.

src/services/codefixes/fixMissingTypeAnnotationOnExports.ts
```diff
@@ -0,0 +1,1192 @@
...
+            if (isPropertyAssignment(node)) {
+                const propName = node.name;
+                addPropertyName(propName);
+            }
+            node = node.parent;
+        }
+        if ((isVariableDeclaration(node) || isPropertyDeclaration(node)) && !isBindingPattern(node.name)) {
+            addPropertyName(node.name);
+        }
+        return nameParts.filter(s => isIdentifierText(s, program.getCompilerOptions().target)).reverse().join("_");
```
andrewbranch, line 422: I might be missing other instances of this issue: anything that appears in `computedOptions` in `src/compiler/utilities.ts` needs to be accessed through a helper function:

```suggestion
        return nameParts.filter(s => isIdentifierText(s, getEmitScriptTarget(program.getCompilerOptions()))).reverse().join("_");
```

src/services/codefixes/fixMissingTypeAnnotationOnExports.ts
```diff
@@ -0,0 +1,1192 @@
...
+        }
+
+        const { typeNode } = inferType(defaultExport.expression);
+        if (!typeNode) return undefined;
+        changeTracker.replaceNodeWithNodes(sourceFile, defaultExport, [
+            factory.createVariableStatement(
+                /*modifiers*/ undefined,
+                factory.createVariableDeclarationList(
+                    [factory.createVariableDeclaration(
+                        "__default",
```
andrewbranch, line 631: Declaration emit uses a single underscore when we generate this (not amazing) name. Does it need `factory.createUniqueName` as well?

```suggestion
                        factory.createUniqueName("_default"),
```

src/services/codefixes/fixMissingTypeAnnotationOnExports.ts
```diff
@@ -0,0 +1,1192 @@
...
+    | NodeBuilderFlags.WriteComputedProps;
+
+enum TypePrintMode {
+    // Prints it's fully spelled out type
+    FULL,
+    // Prints a relative type i.e. typeof X
+    RELATIVE,
+    // Prints a widened type in case the expression is known to
+    // e.g. export const a = Math.random() ? "0" : "1"; the type will be `string` in d.ts files
+    WIDENED,
```
jakebailey, line 172: Nit, but we typically just use `PascalCase` for enum members, not UPPERCASE like enums in other languages

src/services/codefixes/fixMissingTypeAnnotationOnExports.ts
```diff
@@ -0,0 +1,1192 @@
...
+        element?: BindingElement;
+        parent?: ExpressionReverseChain;
+        expression: SubExpression;
+    }
+
+    const enum ExpressionType {
+        TEXT = 0,
+        COMPUTED = 1,
+        ARRAY_ACCESS = 2,
+        IDENTIFIER = 3,
```
jakebailey, line 707: Here toom (or any other enums; i'll stop commenting on them)

src/services/codefixes/fixMissingTypeAnnotationOnExports.ts
```diff
@@ -0,0 +1,1192 @@
...
+    | NodeBuilderFlags.WriteClassExpressionAsTypeLiteral
+    | NodeBuilderFlags.UseTypeOfFunction
+    | NodeBuilderFlags.UseStructuralFallback
+    | NodeBuilderFlags.AllowEmptyTuple
+    | NodeBuilderFlags.GenerateNamesForShadowedTypeParams
+    | NodeBuilderFlags.NoTruncation
+    | NodeBuilderFlags.WriteComputedProps;
+
+enum TypePrintMode {
+    // Prints it's fully spelled out type
```
jakebailey, line 166:
```suggestion
    // Prints its fully spelled out type
```

jakebailey (approved):
On the whole, this seems fine to me; I think we can get feedback on this early and see if people find the big literal ones "too much" or not. But, I'm guessing we won't end up with much feedback until the actual release.

src/services/codefixes/fixMissingTypeAnnotationOnExports.ts
```diff
@@ -0,0 +1,1192 @@
...
+                    chainedExpression,
+                    nextSubExpr.arrayIndex,
+                );
+            }
+        }
+        return chainedExpression;
+    }
+    interface InferenceResult {
+        typeNode?: TypeNode | undefined;
+        mutatedTarget: boolean;
```
andrewbranch, line 921: This looks spooky to me 😅

andrewbranch (commented):
Couple of nits, overall I think the implementation looks really good.

src/services/codeFixProvider.ts
MichaelMitchell-at: To follow-up, I pulled this branch and rebuilt from scratch as well as reinstalled ts-fix from scratch. I can see the quick fix in VSCode now, but ts-fix still doesn't report any diagnostics to fix.

jakebailey: Yeah, https://github.com/microsoft/ts-fix/blob/2898fac87de321fe5d455b594f874bbcafbab6bd/src/index.ts#L343 needs to be updated to also call `getDeclarationDiagnostics`. You should be able to have it do `[...project.program.getSemanticDiagnostics(file), ...project.program.getDeclarationDiagnostics(file)`.

MichaelMitchell-at: Aha, I found the issue. ts-fix is only calling `getSemanticDiagnostics` but isolatedDeclarations are from `getDeclarationDiagnostics`. I don't know if this is considered working as intended, but I submitted a pull request nonetheless https://github.com/microsoft/ts-fix/pull/30.

MichaelMitchell-at: @jakebailey ah I didn't see your comment (still had this window open while investigating and didn't refresh)

MichaelMitchell-at:
Some feedback after trying the code fixer out on part of our codebase.

Unexpected:
- Comments on object properties are duplicated in their respective types. This also seems to be a bug with the pre-existing "Infer function return type" refactoring option. When declaration files are emitted with inferred types, only `/** docstring comments */` get copied (demo: https://tsplay.dev/m3dy1W (click ".D.TS" tab)), so should I think the quickfix should be consistent with that, otherwise these duplicative comments just create noise.
  ```tsx
  export const x: {
      // this will be copied
      a: number;
  } = {
      // this will be copied
      a: 42,
  } satisfies {a: number};
  ```

Bug:
- Quick fix isn't available for `export const x = <div aria-label="hello" />` (under "jsx": "react-jsx") and ts-fix just crashes. It seems to be a problem parsing the `-` in `aria-label`. The crash happens here: https://github.com/microsoft/TypeScript/blob/c763c277ebc16bb39956d822294a1e00059a0b26/src/services/formatting/formatting.ts#L880
- ts-fix will sometimes insert types in the wrong places and produce syntactically invalid files. On the same files, using the "Add all missing type annotations" autofix *does not* have this problem. Submitting a reproduction will take some time, but the only commonality I've seen so far is that these types are complex and composed of multiple imported types, e.g `import('some_module').Foo<{x: import('some_module').Bar, ...}>`.

src/compiler/checker.ts
h-joo: Yeah I tried that as well, and I noticed that the dts emit/ Typechecker was explicitly calling getWidenedLiteralType as well so I made the fixer call it instead.

src/services/codeFixProvider.ts
```diff
@@ -124,9 +124,15 @@ export function eachDiagnostic(context: CodeFixAllContext, errorCodes: readonly
 }
 
 function getDiagnostics({ program, sourceFile, cancellationToken }: CodeFixContextBase) {
-    return [
+    const diagnostics = [
         ...program.getSemanticDiagnostics(sourceFile, cancellationToken),
         ...program.getSyntacticDiagnostics(sourceFile, cancellationToken),
         ...computeSuggestionDiagnostics(sourceFile, program, cancellationToken),
     ];
+    if (program.getCompilerOptions().isolatedDeclarations) {
```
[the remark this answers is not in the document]
jakebailey: (In reality, maybe we should have `getDeclarationDiagnostics` do that internally, to make callers simple?)

src/services/codefixes/fixMissingTypeAnnotationOnExports.ts
h-joo: Oh right, done.

src/services/codefixes/fixMissingTypeAnnotationOnExports.ts
h-joo: Replaced all with `PascalCase`

src/services/codefixes/fixMissingTypeAnnotationOnExports.ts
h-joo: Fixed

src/services/codefixes/fixMissingTypeAnnotationOnExports.ts
h-joo: Removed and replaced the usages with the factored-out function, does this look sense to you?

src/services/codefixes/fixMissingTypeAnnotationOnExports.ts
h-joo: Done

src/services/codefixes/fixMissingTypeAnnotationOnExports.ts
h-joo: Done

src/services/codefixes/fixMissingTypeAnnotationOnExports.ts
h-joo: I'm assuming you're talking about the naming or would it be the logic?

andrewbranch: I mean the strategy of having functions that conditionally mutate something and have to communicate whether they've done so... I didn't study it enough to fully understand why that was needed. Not necessarily a blocker, just a little eyebrow-raising

src/services/codeFixProvider.ts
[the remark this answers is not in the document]
h-joo: Incorporated the first comment.

src/services/codeFixProvider.ts
```diff
@@ -124,9 +125,15 @@ export function eachDiagnostic(context: CodeFixAllContext, errorCodes: readonly
 }
 
 function getDiagnostics({ program, sourceFile, cancellationToken }: CodeFixContextBase) {
-    return [
+    const diagnostics = [
         ...program.getSemanticDiagnostics(sourceFile, cancellationToken),
         ...program.getSyntacticDiagnostics(sourceFile, cancellationToken),
         ...computeSuggestionDiagnostics(sourceFile, program, cancellationToken),
     ];
+    if (getEmitDeclarations(program.getCompilerOptions())) {
```
sheetalkamat, line 133: If currently we are supporting providing code fix for errors only when `--isolatedDeclarations` is turned on may be should get declaration diagnostics only if its on to avoid perf hit

jakebailey: I had them change to this in https://github.com/microsoft/TypeScript/pull/58260#discussion_r1585002609, though; in the editor, we'll have already called this for diags, and it's cached.

jakebailey: We can flip it back to just isolatedDeclarations if you think it'll be a problem, but it seemed less error prone.

h-joo: Quoting Jake from the [other comment](https://github.com/microsoft/TypeScript/pull/58260/#discussion_r1585002609) 

> Actually, now that I think about it, this should probably just be getEmitDeclarations(program.getCompilerOptions()). It's possible that we could have quick fixes for declaration errors that are not in isolated declarations. And this call is cached so should be available for free anyway in codebases with declaration mode enabled.

Would you agree?

h-joo: Oh, my comment was slightly late 😅

sheetalkamat: The cache does help a little but not whole lot since every edit will have to throw out the errors and recalculate them

jakebailey: But, we call this function already and offer them up as diagnostics in the editor; won't the cost already have been paid?

sheetalkamat: Yes i forgot about the part where LS.getSemanticDiagnostics gets declaration diagnostics as well.

jakebailey: Yeah, for posterity:

```
    function getSemanticDiagnostics(fileName: string): Diagnostic[] {
        synchronizeHostData();

        const targetSourceFile = getValidSourceFile(fileName);

        // Only perform the action per file regardless of '-out' flag as LanguageServiceHost is expected to call this function per file.
        // Therefore only get diagnostics for given file.

        const semanticDiagnostics = program.getSemanticDiagnostics(targetSourceFile, cancellationToken);
        if (!getEmitDeclarations(program.getCompilerOptions())) {
            return semanticDiagnostics.slice();
        }

        // If '-d' is enabled, check for emitter error. One example of emitter error is export class implements non-export interface
        const declarationDiagnostics = program.getDeclarationDiagnostics(targetSourceFile, cancellationToken);
        return [...semanticDiagnostics, ...declarationDiagnostics];
    }

```

src/services/codeFixProvider.ts
jakebailey: Resolving this thread; we can talk about ts-fix on the other repo since that doesn't block this PR.

jakebailey:
I think all that's left for me is to just do the final format of the PR (one out of place import). Any other concerns?

jakebailey (approved):

MichaelMitchell-at:
Just FYI, the `aria-label` case I mentioned above isn't `ts-fix`-specific. It just manifests in a worse way with `ts-fix`.

PR #58263, opened 2024-04-20
Clone questionMark and dotdotdot token

armanio123:
Fixes https://github.com/microsoft/TypeScript/issues/57924#issuecomment-2016963508

`CompletionInfo` is failing due to a synthetic `propertySignature` using an incorrect `questionToken` position which cause the Assert to fail. 

The positions are incorrect due to the checker reusing the synthetic node for the `questionToken`. When assigning positions the value gets overwritten by the latest position update this generating a node that is "not contained" on the parent.

The PR fixes this by always cloning a synthetic `questionToken` instead of reusing it.

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -6689,7 +6689,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     if (!nodeIsSynthesized(node) && getParseTreeNode(node) === node) {
                         return node;
                     }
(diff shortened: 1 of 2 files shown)


gabritto:

armanio123:

jakebailey:
This makes sense; `getSynthesizedDeepCloneWorker` is similar and passes in the token visitor. Makes me wonder if we should try and make the visitor required and see where else we've missed this.

But, this does need a test; the linked issue should come with one that can be minimized.

jakebailey:

andrewbranch (approved):

jakebailey (approved):

jakebailey:
Tested, and the test does actually test the crash, nice.

PR #58264, opened 2024-04-20
Error on references to literal enum members in conditions

weswigham:
So when you write
```ts
enum NodeKind {
    FunctionDeclaration = 0,
    VariableDeclaration = 1,
    ExpressionStatement = 2,
}

function kindToString(kind: NodeKind) {
    Debug.assert(!isExpressionStatementKind(kind));
    return NodeKind.FunctionDeclaration ? "func" : "var";
}
```
we inform you in advance that you've failed to actually write `kind ===`.

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -38493,7 +38493,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                         parent = parent.parent;
                     }
                     if (operator === SyntaxKind.AmpersandAmpersandToken || isIfStatement(parent)) {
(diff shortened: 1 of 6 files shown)


weswigham:
@typescript-bot test it

fatcerberus:
Is this a common mistake?  I don't think I've ever done this.  I think I'd rather this be a more general check of "you're testing the value of a (pseudo-)constant, did you mean to do that?" which feels way more useful than trying to detect that I wrote `if (x)` instead of `if (x === y)`

gabritto:
@typescript-bot pack this

gabritto (approved):

PR #58289, opened 2024-04-22
Improve Recovery of Unterminated Regular Expressions

graphemecluster:
Per https://github.com/microsoft/TypeScript/pull/55600#discussion_r1571313988, to not let `scanRegularExpressionWorker` affect the end position of the regex, I separated the logic for unterminated regexes out from the well-terminated ones. This should make moving `scanRegularExpressionWorker` away in the future much easier. Note that unterminated regexes are no longer parsed.


src/compiler/scanner.ts
```diff
@@ -2389,7 +2390,8 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
     function reScanSlashToken(): SyntaxKind {
         if (token === SyntaxKind.SlashToken || token === SyntaxKind.SlashEqualsToken) {
             // Quickly get to the end of regex such that we know the flags
-            let p = tokenStart + 1;
```
graphemecluster, line 2392: This is eliminated for readability

tests/cases/fourslash/whiteSpaceTrimming4.ts
graphemecluster, line 1: I know this is bad. Are there any solutions that does not affect pressing enter?

tests/baselines/reference/parser645086_1.errors.txt
graphemecluster, line 1: Does this now become division?

tests/baselines/reference/tsxAttributeInvalidNames.errors.txt
```diff
@@ -49,5 +49,5 @@ file.tsx(11,20): error TS1161: Unterminated regular expression literal.
 !!! error TS2304: Cannot find name 'data'.
                 ~
 !!! error TS1005: ';' expected.
-                       
+                      ~~
```
graphemecluster, line 52: Should we include angle brackets (in addition to parens and braces) to the logic?

src/testRunner/unittests/regExpParserRecovery.ts
graphemecluster, line 1: I should have named it `regExpScanningRecovery.ts` (which is analogous to `jsonParserRecovery.ts`, I forgot the fact that the recovery part will still remain in the scanner even after the regex worker is moved out)

src/testRunner/unittests/regExpParserRecovery.ts
```diff
@@ -0,0 +1,81 @@
...
+        "/{\\}}",
+        "/({[\\]})]})",
+    ];
+    const whiteSpaceSequences = [
+        "",
+        "  ",
+        "\t\v\r\n",
+        "\u3000\u2028",
+    ];
+    it("stops parsing unterminated regexes at correct position", () => {
```
graphemecluster, line 53: I didn’t name each test case; it’s unrealistic to give each of the test cases a meaningful name. Perhaps they should be named by their indices?

graphemecluster:
Just in case, this might be of interest to @rbuckton and @jakebailey.

graphemecluster:
Comments marked “Outdated” are not actually outdated, they’re just due to the merge

graphemecluster:
Please inform me if you would like me to move the last commit to #58320 or a new PR.

jakebailey:
> Please inform me if you would like me to move the last commit to #58320 or a new PR.

It seems unrelated, right? So should probably be separate?

src/testRunner/unittests/regExpParserRecovery.ts
jakebailey: You could just loop over each and then dynamically name them via say `JSON.stringify(testCase)`.

FWIW you don't need to use `ts.forEach` if you don't want; just a plain loop will work fine.

rbuckton:
#58339 also addressed some of this. Can you update your PR with the latest changes from `main`?

graphemecluster:
> It seems unrelated, right? So should probably be separate?

OK, I’ll do so if you don’t mind the number of PRs.

src/compiler/scanner.ts
```diff
@@ -2469,55 +2468,92 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
...
+                // Whitespaces and semicolons at the end are not likely to be part of the regex
+                while (isWhiteSpaceLike(charCodeChecked(pos - 1)) || charCodeChecked(pos - 1) === CharacterCodes.semicolon) pos--;
+                error(Diagnostics.Unterminated_regular_expression_literal, tokenStart, pos - tokenStart);
+            }
+            else {
+                // Consume the slash character
+                pos++;
+                let regExpFlags = RegularExpressionFlags.None;
+                while (pos < end) {
+                    const ch = codePointUnchecked(pos);
```
rbuckton, line 2521:
```suggestion
                while (true) {
                    const ch = codePointChecked(pos);
                    if (ch === CharacterCodes.EOF) {
                      break;
                    }
```

src/compiler/scanner.ts
```diff
@@ -2469,55 +2468,92 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
...
+                    const size = charSize(ch);
+                    if (reportErrors) {
+                        const flag = characterToRegularExpressionFlag(utf16EncodeAsString(ch));
+                        if (flag === undefined) {
+                            error(Diagnostics.Unknown_regular_expression_flag, pos, size);
+                        }
+                        else if (regExpFlags & flag) {
+                            error(Diagnostics.Duplicate_regular_expression_flag, pos, size);
+                        }
+                        else if (((regExpFlags | flag) & RegularExpressionFlags.UnicodeMode) === RegularExpressionFlags.UnicodeMode) {
```
rbuckton, line 2534: Using `RegularExpressionFlags.UnicodeMode` in this way makes it easy to confuse with `RegularExpressionFlags.Unicode`. I would suggest renaming this flag to `AnyUnicodeMode` to make it less ambiguous.

tests/baselines/reference/tsxAttributeInvalidNames.errors.txt
rbuckton: With respect to treating them as unterminated? I don't think that's necessary. `(?<)` will still produce an error during the grammar check pass while still maintaining a proper bound for the RegExp body.

Also, Annex B does not treat `/\k</` as an error since the RegExp does not define a named capture group, which indicates that `<>` handling is not a lexical syntax error.

src/compiler/scanner.ts
```diff
@@ -2469,55 +2468,92 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
...
+                    else if (!characterClassDepth) {
+                        if (ch === CharacterCodes.openParen) {
+                            bracketStack.push(CharacterCodes.closeParen);
+                        }
+                        else if (ch === CharacterCodes.openBrace) {
+                            bracketStack.push(CharacterCodes.closeBrace);
+                        }
+                        else if (ch === lastOrUndefined(bracketStack)) {
+                            bracketStack.pop();
+                        }
```
rbuckton, line 2504: Parentheses and braces are not part of the [RegularExpressionBody](https://tc39.es/ecma262/#prod-RegularExpressionBody) lexical grammar, and thus unbalanced parentheses don't affect the extent of a [RegularExpressionLiteral](https://tc39.es/ecma262/#prod-RegularExpressionLiteral). If they are unbalanced, that doesn't mean the regexp is considered to be unterminated.

graphemecluster: But this section is all about recovering an unterminated regex. This section does not apply to well-terminated regexes, regardless of whether they’re balanced or unbalanced.
And, aren’t parentheses and braces [RegularExpressionNonTerminator](https://tc39.es/ecma262/#prod-RegularExpressionNonTerminator)s, which are in turn parts of RegularExpressionBody?

src/compiler/scanner.ts
graphemecluster: OK (Well, then this should’ve renamed in the spec too. It seems like the old parameter names (U, N) are even less confusing 😅)

graphemecluster: The change was did in https://github.com/tc39/ecma262/pull/2411, just for my note

src/compiler/scanner.ts
```diff
@@ -2469,55 +2468,92 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
...
-                const ch = charCodeUnchecked(p);
-                if (!isIdentifierPart(ch, languageVersion)) {
-                    break;
-                }
-                if (reportErrors) {
-                    const flag = characterToRegularExpressionFlag(String.fromCharCode(ch));
-                    if (flag === undefined) {
-                        error(Diagnostics.Unknown_regular_expression_flag, p, 1);
+            const endOfRegExpBody = pos;
+            if (tokenFlags & TokenFlags.Unterminated) {
```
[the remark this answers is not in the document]
graphemecluster: Oh sorry I should’ve explained this sooner and clearer. This section is the whole objective of the PR.
Since unterminated regexes are simply invalid that they don’t even form a complete RegularExpressionLiteral, we can do everything at our discretion as long as a syntax error is produced. But we should do all our best to parse the remaining parts with fewer subsequent syntax errors.

Instead of consuming all until the end of the line starting from the slash character, if we encounter an unbalanced bracket, which is likely a bracket balancing something before the regex instead of inside the regex, we stop scanning the regex here such as to increase the chance that a bracket forms a pair of syntax characters with the corresponding bracket before the regex.

src/compiler/scanner.ts
rbuckton: They are part of the body, but RegularExpressionBody itself doesn't care if they are balanced. That check occurs later as part of the _Pattern_ grammar, not the lexical grammar. That said, I can see why they need to be considered for recovery purposes.

src/compiler/scanner.ts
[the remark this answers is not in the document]
rbuckton: Thank you for the explanation. My primary concern is the allocation of the `bracketStack` array, as it introduces an array allocation per scanned RegExp, even when that RegExp contains no brackets, parens, or braces. 

It might be possible to write this in a way such that an allocation isn't necessary. Parentheses can nest, and square brackets can nest, but square brackets can't have nested parentheses, and curly braces cannot have nested parentheses, braces, or brackets.

Instead, you can count the number of left parentheses until you encounter either a left bracket, left brace, or right parenthesis. If you encounter a left bracket, you then count the number of left brackets until you encounter an equal number of right brackets. If you encounter a left brace, you scan for the next right brace. If you encounter a right parenthesis, you decrement the left parenthesis count. Instead of an array, you only really need two counters and a `boolean`.

rbuckton (changes requested):

src/compiler/scanner.ts
rbuckton: I think you may be running afoul of the same confusion I'm talking about. `RegularExpressionFlags.Unicode` is synonymous with the `Unicode` (old `U`) production parameter in the spec, while `RegularExpressionFlags.UnicodeSets` is synonymous with the `UnicodeSets` production parameter in the spec.

`RegularExpressionFlags.UnicodeMode` has more in common with the [HasEitherUnicodeFlag](https://tc39.es/ecma262/#sec-runtime-semantics-haseitherunicodeflag-abstract-operation) AO, which is why I'm suggesting we change the name of the `UnicodeMode` flag.

rbuckton: Yes, but the names in `RegularExpressionFlags` don't currently match the names in `Pattern`.

```ts
export const enum RegularExpressionFlags {
    None        = 0,
    HasIndices  = 1 << 0, // d
    Global      = 1 << 1, // g
    IgnoreCase  = 1 << 2, // i
    Multiline   = 1 << 3, // m
    DotAll      = 1 << 4, // s
    Unicode     = 1 << 5, // u
    UnicodeSets = 1 << 6, // v
    Sticky      = 1 << 7, // y
    UnicodeMode = Unicode | UnicodeSets,
    Modifiers   = IgnoreCase | Multiline | DotAll,
}
```

tests/baselines/reference/tsxAttributeInvalidNames.errors.txt
rbuckton: I'm not too terribly concerned about this case, to be honest. Parsing falls off the rails here because of malformed JSX, not a malformed RegExp.

src/compiler/scanner.ts
graphemecluster: `RegularExpressionFlags.Unicode` is not synonymous to the `UnicodeMode` parameter. `RegularExpressionFlags.UnicodeMode` is.
image

src/compiler/scanner.ts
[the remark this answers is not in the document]
graphemecluster: > curly braces cannot have nested parentheses, braces, or brackets

It is not the case in Annex B. That’s why I wrote like this.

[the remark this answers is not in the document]
rbuckton: > > curly braces cannot have nested parentheses, braces, or brackets
> 
> It is not the case in Annex B. That’s why I wrote like this.

Can you point to where? Per the Annex B grammar, the only places `{` is balanced are [Quantifier](https://tc39.es/ecma262/#prod-Quantifier) and [InvalidBracedQuantifier](https://tc39.es/ecma262/#prod-annexB-InvalidBracedQuantifier), both of which only allow [DecimalDigits](https://tc39.es/ecma262/#prod-DecimalDigits) and comma (`,`). Otherwise, it matches [ExtendedPatternCharacter](https://tc39.es/ecma262/#prod-annexB-ExtendedPatternCharacter), which isn't balanced.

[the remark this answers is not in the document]
graphemecluster: Well, what I mean is regexes like `/{{\s*(\w+)\s*}}` do exist in the wild.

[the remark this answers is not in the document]
rbuckton: Maybe so, but those braces don't need to be matched. If you encounter an open brace that doesn't match a quantifier, then you don't need to balance it.

graphemecluster:
@jakebailey I’ve moved the changes related to flags scanning to #58612. After either PR is merged I will resolve the conflict of the other one.

src/compiler/scanner.ts
```diff
@@ -2469,73 +2462,115 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
...
+                        }
+                        else if (!inDecimalQuantifier) {
+                            if (ch === CharacterCodes.openParen) {
+                                groupDepth++;
+                            }
+                            else if (ch === CharacterCodes.closeParen && groupDepth) {
+                                groupDepth--;
+                            }
+                            else if (ch === CharacterCodes.closeParen || ch === CharacterCodes.closeBracket || ch === CharacterCodes.closeBrace) {
+                                // We encountered an unbalanced bracket outside a character class. Treat this position as the end of regex.
```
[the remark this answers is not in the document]
graphemecluster: Yes, you are right. _Before_ `]` or `}`, actually. And now `[/][/` will produce two unterminated regex. If this isn’t ideal, may you kindly suggest what should I do?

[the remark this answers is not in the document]
rbuckton: It isn't an issue. I just wanted to be sure I understood the intent.

rbuckton:
@typescript-bot perf test
@typescript-bot run dt
@typescript-bot test top400

src/compiler/scanner.ts
```diff
@@ -2763,25 +2794,26 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
 
         function scanPatternModifiers(currFlags: RegularExpressionFlags): RegularExpressionFlags {
             while (true) {
-                const ch = charCodeChecked(pos);
+                const ch = codePointChecked(pos);
```
rbuckton, line 2797: Should this be moved to #58612 as well?

graphemecluster: Ah, yes. It was a mistake that I thought I did entirely reverted.

rbuckton (approved):

PR #58290, opened 2024-04-22
Report relation cache sizes in --diagnostics too

jakebailey:
This would be helpful for https://github.com/microsoft/typescript-benchmarking/pull/58, as passing `--extendedDiagnostics` turns on tracing which slows down the compiler (not great for benchmarking).

But, maybe this is too much info for the more generic `--diagnostics` printout.

### DIFF
--- a/src/compiler/executeCommandLine.ts
+++ b/src/compiler/executeCommandLine.ts
@@ -1178,6 +1178,11 @@ function reportStatistics(sys: System, program: Program, solutionPerformance: So
         if (memoryUsed >= 0) {
             reportStatisticalValue({ name: "Memory used", value: memoryUsed, type: StatisticType.memory }, /*aggregate*/ true);
         }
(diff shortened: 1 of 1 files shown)


DanielRosenwasser:
> But, maybe this is too much info for the more generic `--diagnostics` printout.

I might be wrong, but I think the only reason `--diagnostics` doesn't just do the same as `--extendedDiagnostics` is because we didn't want to break existing tools that parsed the output (and possibly that the latter measures things a little incorrectly?).

jakebailey:
Yeah, we talked about this in the meeting. It's now a lot easier to backfill data so we can change things, but the main problem is that `extendedDiagnostics` requires native performance events which slows down the benchmarks and makes them less realistic (since as we add more trace points, the perf gets worse with the flag on, but not for any actual user).

Issue #58266, opened 2024-04-20
`NoInfer` isn't erased and breaks type narrowing


patroza:
### 🔎 Search Terms

NoInfer

### 🕗 Version & Regression Information

- This changed between versions 5.3 and 5.4 (introduced)

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.5.0-dev.20240419#code/C4TwDgpgBAglC8UDeB9YBDA5gLigInTwF8AoUSKAIQWTS1zwCNiSzxoBVGuAHytYD0AqOWgA5APYBJAHYAzCACcATAB4YAPm6sAxhJkBnYFAC2uJIogGArgBtguSbIWLVHDUQDaAXRo-WctYyOsAAlvqm6GDqADRUGgAUocAQJga4lugAJvq2ILA+cXK4CeiO0vJK6hoAlAhalHVIUJbA1ooyUFDJqQYAdCZRCXJ1pEJQEDJZBlDWYFDoM1AJzZY29uXOVe4A3FBEUHwydrY1Pt2GKdlQEnLLq1Z2DrB7B0cnZ94kepdQBhImCDAAAWoRkmBogzACRMcQSKDq8C0KD6ayefToEPg2PwhCgAH5kFA+iSUHE0RsoCiKcYDrhjrZTiQgA

### 💻 Code

```ts
type A = {_tag: "a"}
type B = {_tag: "b"}

type U = A | B

// type NoInfer2<A> = A

const m: { result: NoInfer<U> }[] = []

function map<A, B>(items: readonly A[], f: (a: NoInfer<A>) => B) { return  items.map(f) }
// ends up as  ({ result: NoInfer<U>; } | null)[] instead of ({ result: A; } | null)[]
const something = map(m, (_) => _.result._tag === "a" ? { ..._, result: _.result } : null)
```

### 🙁 Actual behavior

ends up as  `({ result: NoInfer<U>; } | null)[]`

### 🙂 Expected behavior

`({ result: A; } | null)[]`

### Additional information about the issue

use the custom `NoInfer2` type instead, and it works as expected.

[...]

PR #58292, opened 2024-04-23
Unwrap `NoInfer` types when narrowing

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/58266


src/compiler/checker.ts
```diff
@@ -27681,6 +27681,8 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         if (!flowNode) {
             return declaredType;
         }
+        declaredType = isNoInferType(declaredType) ? (declaredType as SubstitutionType).baseType : declaredType;
+        initialType = isNoInferType(initialType) ? (initialType as SubstitutionType).baseType : initialType;
```
Andarist, line 27685: My first instinct was that I should re-wrap the result of narrowing with `NoInfer` but quoting [the original `NoInfer` PR](https://github.com/microsoft/TypeScript/pull/56794):
> Other than blocking inference, NoInfer<T> markers have no effect on T. Indeed, T and NoInfer<T> are considered identical types in all other contexts.

So I don't see a reason why that would be necessary. I also tested something like this ([TS playground](https://www.typescriptlang.org/play?ts=5.5.0-dev.20240422&ssl=9&ssc=4&pln=1&pc=1#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXy1VRBgB4AVAPgApYBzALnnIEonyBuAKFElgRTpseeDmQYSFGlHYAaeGABGTWjEbwAcjgCSqRJKot4AXkrwAbjizA2Fq8G5cxEmNQDeSHDiYAiRbB94AF95VTojU3g3Lnh4ehj4AHpE+AA9AH4EsDwAZwx4OBzkCHzjAiISMJZuWOTY2IyuIOquIA)) to see if this would be needed and I don't think it is:
```ts
declare function inner<T>(arg: T): T;
declare function outer<T>(a: T, cb: (arg: NoInfer<T>) => void): void;

outer({ foo: "bar" }, (arg) => {
  arg
  // ^? (parameter) arg: NoInfer<{ foo: string; }>
  const result = inner(arg); // inferred despite `arg` being a `NoInfer` type!
  //    ^? const result: NoInfer<{ foo: string; }>
});
```

ahejlsberg: I agree it's fine to erase `NoInfer` in variable, parameter, and property references, but I think it would be simpler and more consistent to do it first thing in the `getNarrowableTypeForReference` function. I think that's all you'd need.

ahejlsberg (changes requested):
See my comment above.

src/compiler/checker.ts
Andarist: moved it to `getNarrowableTypeForReference`, works like a charm - thanks!

ahejlsberg (approved):

Andarist:
@sandersn @RyanCavanaugh could this one land before the upcoming RC?

Issue #58287, opened 2024-04-22
Regular Expression finds


RyanCavanaugh:
### Acknowledgement

- [X] I acknowledge that issues using this template may be closed without further explanation at the maintainer's discretion.

### Comment

Note: I eventually gave up on capturing "Not available unless target is ESXXXX" errors since they're not really interesting to look at

Via #58275

[This character cannot be escaped in a regular expression.](https://github.com/AbdullahAlfaraj/Auto-Photoshop-StableDiffusion-Plugin/blob/6f6d4907adf0b6a6e562d084238f02ed6b7503c1/utility/html_manip.js#L461)
```ts
const image_path_escape = image_path.replace(/\o/g, '/o') //escape string "\o" in "\output"
```

[Named capturing groups are only available when targeting 'ES2018' or later](https://github.com/adobe/react-spectrum/blob/1d70153c126e84fa7550ff63e3e48fd92b095ca8/packages/@internationalized/date/src/string.ts#L26)
```ts
/^((?<negative>-)|\+)?P((?<years>\d*)Y)?((?<months>\d*)M)?((?<weeks>\d*)W)?((?<days>\d*)D)?((?<time>T)((?<hours>\d*[.,]?\d{1,9})H)?((?<minutes>\d*[.,]?\d{1,9})M)?((?<seconds>\d*[.,]?\d{1,9})S)?)?$/
```

[Named capturing groups are only available when targeting 'ES2018' or later.](https://github.com/angular/angularfire/blob/73b40da11f7f56884c86da9cbe1123f3d00bc866/src/schematics/update/v7/index.ts#L7)
```ts
const IMPORT_REGEX = /(?<key>import|export)\s+(?:(?<alias>[\w,{}\s*]+)\s+from)?\s*(?:(?<quote>["'])?(?<ref>[@\w\s\\/.-]+)\3?)\s*(?<term>[;\n])/g
```

[Named capturing groups are only available when targeting 'ES2018' or later](https://github.com/ariakit/ariakit/blob/c18f793aed1af7235b8b2a5ed2b3d6db3501dff6/website/build-pages/reference-utils.js#L181)
```ts
const match = text.match(/^(?<description>(.|\n)*)```(?<language>[^\n]+)\n(?<code>(.|\n)+)\n```$/m);
```

[...]

PR #58295, opened 2024-04-23
Report RegExp errors in grammar check, use Annex B grammar

rbuckton:
This moves extended regular expression syntax checks to a grammar check. This ensures that RegExp grammar checks don't prevent type checking and allows them to be overridden via `// @ts-ignore` when necessary.

This also weakens the grammar check to use the productions and semantics in [Annex B](https://tc39.es/ecma262/#sec-regular-expressions-patterns). This uses a `boolean` value to choose whether to use Annex B grammar in case we want to make this configurable, but for now the only place this is defined is guaranteed to be `true`.

Related #58287


jakebailey:
Theoretically this should net some "missing" errors

@typescript-bot run dt
@typescript-bot test top400

graphemecluster:
FYI the cause of the CI error was modified here:
https://github.com/microsoft/TypeScript/pull/55600/files#diff-fd7e6ee327c7bf4e89fe266a9f7cb6cc989a6c59562cc8bec3581da1d2fe4eed

src/compiler/scanner.ts
```diff
@@ -2444,44 +2448,57 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
...
         return token;
 
-        function scanRegularExpressionWorker(text: string, end: number, regExpFlags: RegularExpressionFlags, isUnterminated: boolean) {
-            /** Grammar parameter */
-            const unicodeMode = !!(regExpFlags & RegularExpressionFlags.UnicodeMode);
+        function scanRegularExpressionWorker(text: string, end: number, regExpFlags: RegularExpressionFlags, isUnterminated: boolean, annexB: boolean) {
             /** Grammar parameter */
             const unicodeSetsMode = !!(regExpFlags & RegularExpressionFlags.UnicodeSets);
+            /** Grammar parameter */
+            const unicodeMode = unicodeSetsMode || !!(regExpFlags & RegularExpressionFlags.UnicodeMode); // v always implies u
```
graphemecluster, line 2495: `RegularExpressionFlags.UnicodeMode` differs from `RegularExpressionFlags.Unicode`, the former is essentially `Unicode | UnicodeSets`.
If you prefer this way you may want to remove `UnicodeMode` from the enum.

src/compiler/scanner.ts
```diff
@@ -1484,7 +1486,7 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
     //     | [0-3] [0-7] [0-7]?
     //     | [4-7] [0-7]
     // NonOctalDecimalEscapeSequence ::= [89]
-    function scanEscapeSequence(shouldEmitInvalidEscapeError: boolean, isRegularExpression: boolean): string {
+    function scanEscapeSequence(shouldEmitInvalidEscapeError: boolean, isRegularExpression: boolean | "annex-b"): string {
```
graphemecluster, line 1489: It might be worthwhile to make `annexB` another parameter since we might choose to support Annex B also for non-regex (related #51837)

rbuckton: I considered that but decided to wait until we need it.

rbuckton:
In general, I think there might be a happy medium between general specification strictness and Annex B looseness. For example, we probably want to treat `\1`, `\2`, etc., as an error when they are either higher than the number of capturing left parens or are used in a character class as they are ambiguous. It's likely that the user expected `\2` to be treated as a backreference, and most of the cases where users have written `[^\2]` seem to be due to a misunderstanding about how character classes work. Erroring in these cases makes sense as it would point out the misuse.

Cases like `[\w-_]`, where a minus (`-`) appears in a character class before or after a character class escape like `\w` or `\d`, are usually the result of a user conflating `-` as a potential character to match vs `-` to denote a range. Annex B would treat `[\w-_]` the same as `[-\w_]`, so there's no ambiguity with that syntax. The only ambiguity here would be one that strict checking _wouldn't_ catch, such as `[a-_]` where the user may have intended to use `-` as a character to match against rather than a range. There may be value in reporting an error for the `[\w-_]` case to make users take more notice of the `[a-_]` case, but that's not going to be a reliable check.

Most of the other Annex B checks we may want to just leave out in favor of stricter parsing as they generally indicate invalid syntax that you're less likely to actually come across in practice.

The upside of doing this during the grammar check pass is that if a user _really_ wants to use that syntax then they have the option to annotate the line with `// @ts-ignore`.

There are a few other things I want to explore either in this PR or another follow up PR to cut down on the number `--target`-specific errors, either by just treating them as suggestion diagnostics for now, or introducing a compiler flag to give the users control over reporting such errors broadly as opposed to just using `// @ts-ignore` to silence them on a case by case basis.

jakebailey (approved):

rbuckton:
I'll discuss whether we should use stricter rules than Annex B in the next Design Meeting, with any changes coming in a follow-up PR.

rbuckton:
@typescript-bot perf test
@typescript-bot run dt

src/compiler/checker.ts
```diff
@@ -31353,6 +31356,42 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+            );
+            Debug.assert(scanner.scan() === SyntaxKind.SlashToken);
+            Debug.assert(scanner.reScanSlashToken(/*reportErrors*/ true) === SyntaxKind.RegularExpressionLiteral);
+            return !!lastError;
+        }
+        return false;
+    }
+
+    function checkRegularExpressionLiteral(node: RegularExpressionLiteral) {
+        checkGrammarRegularExpressionLiteral(node);
```
weswigham, line 31391:
```suggestion
        addLazyDiagnostic(() => checkGrammarRegularExpressionLiteral(node));

```
you also might want to cache the state of having performed the check, since, as an expression, we might call `check` on it a _lot_ - deferral will remove that cost for LS operations, but a cache will help make it so we only do this rescan once.

src/compiler/scanner.ts
```diff
@@ -2571,7 +2588,8 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
                                     case CharacterCodes.equals:
                                     case CharacterCodes.exclamation:
                                         pos++;
-                                        isPreviousTermQuantifiable = false;
+                                        // In Annex B, `(?=Disjunction)` and `(?!Disjunciton)` are quantifiable
```
DanielRosenwasser, line 2591:
```suggestion
                                        // In Annex B, `(?=Disjunction)` and `(?!Disjunction)` are quantifiable
```

src/testRunner/unittests/incrementalParser.ts
```diff
@@ -160,7 +160,7 @@ describe("unittests:: Incremental Parser", () => {
         const oldText = ts.ScriptSnapshot.fromString(source);
         const newTextAndChange = withInsert(oldText, semicolonIndex, "/");
 
-        compareTrees(oldText, newTextAndChange.text, newTextAndChange.textChangeRange, 4);
+        compareTrees(oldText, newTextAndChange.text, newTextAndChange.textChangeRange, 0);
```
DanielRosenwasser, line 163: what is this 4th parameter?

src/compiler/checker.ts
```diff
@@ -31353,6 +31356,42 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         }
     }
 
+    function checkGrammarRegularExpressionLiteral(node: RegularExpressionLiteral) {
+        const sourceFile = getSourceFileOfNode(node);
+        if (!hasParseDiagnostics(sourceFile)) {
+            let lastError: DiagnosticWithLocation | undefined;
+            const scanner = createScanner(
```
DanielRosenwasser, line 31363: It might be good to cache a single scanner here and reset the state instead.

src/testRunner/unittests/incrementalParser.ts
jakebailey: 0 means "all AST nodes were reused", IIRC; so it being a parse error before meant we weren't reusing the AST in incremental parsing.

src/compiler/checker.ts
DanielRosenwasser: What's the purpose of `addLazyDiagnostic` here? Does it cache?

rbuckton: > you also might want to cache the state of having performed the check, since, as an expression, we might call `check` on it a _lot_

I'm surprised we need an independent cache when we already have `checkExpressionCached`. It feels like we've significantly weakened caching over the years.

src/testRunner/unittests/incrementalParser.ts
rbuckton: The number of elements that could be reused. This was `0` prior to the original PR and only changed due to the additional diagnostics that were added during parse. Now that these diagnostics are produced during check, this needs to be reverted to the prior value.

jakebailey:
Hm, crash on the DT code, it looks. Definitely relevant.

src/compiler/checker.ts
rbuckton: Perhaps, though we create a lot of single-use scanners during check already due to `getSpanOfTokenAtPosition` and `scanTokenAtPosition`.

src/compiler/checker.ts
rbuckton: Are you suggesting to cache *and* to call `addLazyDiagnostic`, or is caching alone sufficient?

src/compiler/checker.ts
jakebailey: Yeah, I've been trying to figure out a good way to do this more generally, but this isn't new.

rbuckton:
> Hm, crash on the DT code, it looks. Definitely relevant.

Not source mapped, so that's less than helpful :/

jakebailey:
The assert is from:

```
Debug.assert(scanner.reScanSlashToken(/*reportErrors*/ true) === SyntaxKind.RegularExpressionLiteral);
```

```console
$ NODE_OPTIONS=--enable-source-maps pnpm dtslint --localTs ~/work/TypeScript/built/local types codemirror
dtslint@0.2.20
Error: Debug Failure. False expression.
Occurred while linting /home/jabaile/work/DefinitelyTyped/types/codemirror/test/addon/mode/simple.ts:1
Rule: "@definitelytyped/expect"
    at checkGrammarRegularExpressionLiteral (/home/jabaile/work/TypeScript/src/compiler/checker.ts:31382:23)
    at <anonymous> (/home/jabaile/work/TypeScript/src/compiler/checker.ts:31397:37)
    at addLazyDiagnostic (/home/jabaile/work/TypeScript/src/compiler/checker.ts:47223:35)
    at checkRegularExpressionLiteral (/home/jabaile/work/TypeScript/src/compiler/checker.ts:31397:13)
    at checkExpressionWorker (/home/jabaile/work/TypeScript/src/compiler/checker.ts:39712:24)
    at checkExpression (/home/jabaile/work/TypeScript/src/compiler/checker.ts:39621:36)
    at checkExpressionForMutableLocation (/home/jabaile/work/TypeScript/src/compiler/checker.ts:39357:22)
    at checkPropertyAssignment (/home/jabaile/work/TypeScript/src/compiler/checker.ts:39371:16)
    at checkObjectLiteral (/home/jabaile/work/TypeScript/src/compiler/checker.ts:31658:80)
    at checkExpressionWorker (/home/jabaile/work/TypeScript/src/compiler/checker.ts:39716:24)
```

Repro instructions are:

```console
$ git clone --filter=blob:none https://github.com/DefinitelyTyped/DefinitelyTyped.git
$ cd DefinitelyTyped
$ pnpm i --filter . --filter '...{./types/codemirror}...'
$ NODE_OPTIONS=--enable-source-maps pnpm dtslint --localTs ~/work/TypeScript/built/local types codemirror
```

rbuckton:
> The assert is from: [...]

I was able to repro it locally. I was lamenting the lack of source maps, but I should have tagged the assert with an appropriate message anyways.

The assert was overaggressive since we can also rescan `/=` as well. I removed it since the assert on the next line would catch any discrepancy anyways.

rbuckton:
@typescript-bot run dt

src/compiler/checker.ts
rbuckton: > What's the purpose of `addLazyDiagnostic` here? Does it cache?

It appears to defer some non-consequential checks until later so that some LS features can use the checker without eagerly checking things that produce diagnostics but also don't affect types.

rbuckton:
@weswigham, @DanielRosenwasser, any other comments?

src/compiler/checker.ts
weswigham: Exactly - `addLazyDiagnostic` is actually eager in the CLI/diagnostics request case and immediately invokes the passed function. In the LS, though, it saves off that work and only does it later when diagnostics actually get requested (or never, if the checker is discarded due to an edit, as is often the case).

> Are you suggesting to cache and to call addLazyDiagnostic, or is caching alone sufficient?

Caching alone alleviates most of my concerns (since `checkExpreessionWithContextualType` is going to check the same expression without caching a bunch), but you'll get slightly better LS perf by deferring it, too.

src/compiler/checker.ts
```diff
@@ -31353,6 +31358,48 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+                scanner.setText("");
+                scanner.setOnError(/*onError*/ undefined);
+            }
+        }
+        return false;
+    }
+
+    function checkRegularExpressionLiteral(node: RegularExpressionLiteral) {
+        const nodeLinks = getNodeLinks(node);
+        if (!nodeLinks.grammarChecked) {
```
weswigham, line 31396: I think you can actually reuse `NodeCheckFlags.TypeChecked` for this, if you'd like. It's how we do similar marking for getter/setter pairs, and nodes with a `checkNodeDeferred` component.

weswigham (approved):

DanielRosenwasser:
> The assert was overaggressive since we can also rescan /= as well. I removed it since the assert on the next line would catch any discrepancy anyways.

Does that mean we have no test with `/=` in it?

rbuckton:
> > The assert was overaggressive since we can also rescan /= as well. I removed it since the assert on the next line would catch any discrepancy anyways.
> 
> Does that mean we have no test with `/=` in it?

Apparently not. I'll add one.

DanielRosenwasser:
@weswigham why does it have to be a lazy diagnostic though? Where does the order-checking impact anything else here?

DanielRosenwasser (approved):

weswigham:
> why does it have to be a lazy diagnostic though? Where does the order-checking impact anything else here?

It allows us to completely skip the rescan when, eg, requesting completions in the file.

rbuckton:
> > why does it have to be a lazy diagnostic though? Where does the order-checking impact anything else here?
> 
> It allows us to completely skip the rescan when, eg, requesting completions in the file.

We will eventually want to always perform the scan if we want to leverage information like group names or capture group counting to infer stronger types for `RegExpExecArray`, but it makes sense to defer it for now.

DanielRosenwasser:
Okay, I guess don't have a strong opinion about it. It does feel a little premature since ideally a scan should be fast enough, but it's no big deal for now. Are we good with merging?

rbuckton:
> Okay, I guess don't have a strong opinion about it. It does feel a little premature since ideally a scan should be fast enough, but it's no big deal for now. Are we good with merging?

I'm mostly waiting on the macos tests to catch up, but they're lagging about 4hrs behind right now. Nothing in this PR should be OS dependent, so I'm fine with merging if we don't want to wait.

graphemecluster:
After this PR, [these lines](https://github.com/microsoft/TypeScript/pull/55600#discussion_r1571313988) no longer work and might happen to break something as the early exit causes the token span and value to be different depending on `reportErrors`, thus should be removed. (I did it in #58289)

PR #6532, opened 2016-01-19
Readonly properties and index signatures

ahejlsberg:
This PR adds support for readonly properties and index signatures as suggested in #12. Changes include:

* A new `readonly` modifier that can applied to property and index signature declarations.
* A property or index signature declared with the `readonly` modifier is considered read-only.
* Read-only properties may have initializers and may be assigned to in constructors within the same class declaration, but otherwise assignments to read-only properties are disallowed.
* Assignments to read-only index signatures are always disallowed.

Entities are *implicitly* read-only in several situations:

* A property declared with a `get` accessor and no `set` accessor is considered read-only.
* In the type of an enum object, enum members are considered read-only properties.
* In the type of a module object, exported `const` variables are considered read-only properties.
* An entity declared in an `import` statement is considered read-only.
* An entity accessed through an ES2015 namespace import is considered read-only (e.g. `foo.x` is read-only when `foo` is declared as `import * as foo from "foo"`).

Some examples:

```typescript
interface Point {
    readonly x: number;
    readonly y: number;
}

var p1: Point = { x: 10, y: 20 };
p1.x = 5;  // Error, p1.x is read-only

var p2 = { x: 1, y: 1 };
var p3: Point = p2;  // Ok, read-only alias for p2
p3.x = 5;  // Error, p3.x is read-only
p2.x = 5;  // Ok, but also changes p3.x because of aliasing
```
```typescript
class Foo {
    readonly a = 1;
    readonly b: string;
    constructor() {
        this.b = "hello";  // Assignment permitted in constructor
    }
}
```
```typescript
// Read-only view of Array<T> with all mutating methods removed
interface ReadonlyArray<T> {
    readonly [x: number]: T;
    readonly length: number;
    toString(): string;
    toLocaleString(): string;
    concat<U extends ReadonlyArray<T>>(...items: U[]): T[];
    concat(...items: T[]): T[];
    join(separator?: string): string;
    slice(start?: number, end?: number): T[];
    indexOf(searchElement: T, fromIndex?: number): number;
    lastIndexOf(searchElement: T, fromIndex?: number): number;
    every(callbackfn: (value: T, index: number) => boolean, thisArg?: any): boolean;
    some(callbackfn: (value: T, index: number) => boolean, thisArg?: any): boolean;
    forEach(callbackfn: (value: T, index: number) => void, thisArg?: any): void;
    map<U>(callbackfn: (value: T, index: number) => U, thisArg?: any): U[];
[...]

Issue #22315, opened 2018-03-04
Suggestion: readonly method


Conaclos:
**Problem:**

A method (function written with the [ES6 method syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Method_definitions)) can currently be replaced with another function by a simple assignment.

As an example consider the following snippet of code:

```ts
class Person {
    constructor(...) {...}
    displayName (): string {...}
}
const x: Person = new Person(...)
x.displayName = function(this: Person): string {...} // Assignment
```
Because the `readonly` modifier is not usable for methods, this is not possible to prevent this kind of assignments.
 
**Proposal:**

A method is always readonly.

The following codes are identical:
```ts
class Person {
    displayName (): string {...}
}
```
```ts
class Person {
    readonly displayName: (this: Person): string = function () {...}
}
```

**Compatibility:**

This is a breaking change. However, method syntax is recent and mostly used in classes. Codes which assign a function to a method are certainly rare.

**Temporary workaround:**

Do not use method syntax in your classes and interfaces. Note hat this leads to very verbose codes.

If you use an interface, the verbosity is acceptable. However you get also strict variance.
```ts
class PersonI {
    readonly displayName: (this: Person) => string
}
```

feeddageek:
I'd like to add the the Readonly mapped type currently work on method event if the readonly keyword it self is not allowed.
```typescript
class SomeClass {
    //public readonly SomeMethod() {} // 'readonly' modifier can only appear on a property declaration or index signature.
    public readonly SomeOtherMethod() {}
}
const someObject =  new SomeClass();
someObject.SomeOtherMethod = () => {}; // OK
const someOtherObject = new SomeClass() as Readobly<typeof SomeClass>
someOtherObject.SomeOtherMethod = () => {}; // Cannot assign to 'SomeOtherMethod ' because it is a constant or a read-only property.
```

A flag under the strict umbrella to make all method readonly by default and force the use of property function for "assignable method" would also be welcome.

patrykuszynski:
@feeddageek It's allowed when you store method as arrow function class property:
```js
class SomeClass {
    public readonly SomeOtherMethod = () => {};
}

const someObject =  new SomeClass();
someObject.SomeOtherMethod = () => { /* new impl */ }; //  error TS2540: Cannot assign to 'SomeOtherMethod' because it is a read-only property.
```

Conaclos:
[...]

Issue #9252, opened 2016-06-19
Suggestion: Upper-bound generic type constraints


GregRos:
This is a proposal for generic structural supertype constraints.

I'll first detail the proposal, and then get into a discussion about use-cases.
## Proposal

A supertype constraint is a constraint on a generic type parameter to be the structural _supertype_ of another type. It's like a subtype/inheritance constraint, but from the other direction. For example, let's say we have a square:

```
interface Square {width : number; height : number}
```

In this suggested syntax, you could write:

```
performAction<T extended by Square>(partialSquare : T) {}
```

This means the type `T` must be a structural supertype of square. So the potential candidates for `T` are:

```
{width : number}, {height : number}, Square, {}
```

It's an established kind of type constraint, not something I just made up. Although it's not exactly common, some languages do implement this feature, such as Scala. In Scala, you can write:

```
def performAction[T >: Square](obj : T) = { ... }
```

To express a supertype/upper bound constraint. In this case, of course, the constraint isn't structural -- `T` must declare that it implements `Square`. 
## Utility

In most languages, including Scala, this kind of constraint isn't very useful. It only comes up in certain specific situations. 

However, Javascript libraries often have this kind of API, where you're allowed to specify the partial properties of an object to modify it (the rest remain at their previous value).

In many cases, you can support this by having optional interface members, but isn't always possible or correct.

An important example is React, which defines a method called `setState`:

```
class Component {
    var state;

    setState(partialState) {
        //merges the properties of partialState with the current state.
    }
}
```

The right signature for this method should be:

```
class Component<Props, State> {
    setState<PartialState extended by State>(partialState : PartialState) {
        //merges the properties of partialState with the current state.
    }
}
```

Currently,the definition files state that it is:

```
setState(fullState : State) {
    //merges the properties of partialState with the current state.
}
```

Which doesn't fully capture the functionality of the method.
## Notes

The suggested syntax doesn't give us a nice way of combining both subtype and supertype constraints. One possibility is:

```
exampleMethod<T extends LowerBound and extended by UpperBound>
```

Strate:
Possible duplicate of
#4889
#371

GregRos:
@Strate, Those really aren't duplicates at all. They're very different, both conceptually and practically.

That said, I'm quite surprised this hasn't been proposed previously, so I suspected it was a duplicate, although I couldn't find anything.

Strate:
[...]

PR #58296, opened 2024-04-23
New `--enforceReadonly` compiler option to enforce read-only semantics in type relations

### LINKED ISSUE #13347: Interface with readonly property is assignable to interface with mutable property
**TypeScript Version:**  2.1.4

**Code**

```ts
interface MutableValue<T> {
    value: T;
}

interface ImmutableValue<T> {
    readonly value: T;
}

let i: ImmutableValue<string> = { value: "hi" };
i.value = "Excellent, I can't change it"; // compile-time error

let m: MutableValue<string> = i;
m.value = "Oh dear, I can change it";
```

**Expected behavior:**
The assignment of `i` to `m` would fail, to stop us accidentally allowing `value` to be modified.

**Actual behavior:**
The assignment is allowed.

This is presumably a deliberately choice, in which case this is a breaking change (or strict flag) feature request rather than a bug report!

The Handbook [has this snippet](https://www.typescriptlang.org/docs/handbook/interfaces.html):

```ts
let a: number[] = [1, 2, 3, 4];
let ro: ReadonlyArray<number> = a;
ro[0] = 12; // error!
ro.push(5); // error!
ro.length = 100; // error!
a = ro; // error!
```

It notes that `a = ro` being an error is helpful. But this happens because `ReadonlyArray` has no `push` method, making it incompatible with `Array`.

My example above seems "morally equivalent" to modelling the input/output flow of values with separate methods:

```ts
interface MutableValue<T> {
    getValue(): T;
    setValue(v: T): void;
}

interface ImmutableValue<T> {
    getValue(): T;
}

declare let i: ImmutableValue<string>;
i.setValue("Excellent, I can't change it"); // compile-time error

let m: MutableValue<string> = i;
m.setValue("Oh dear, I can change it");
```

And sure enough, this stops the assignment of `i` to `m`.

Would be great if mutable and readonly properties had the same relationship as if they were modelled by separate get/set methods (which of course they might actually be, via property getter/setters).

ahejlsberg:
This PR introduces a new `--enforceReadonly` compiler option to enforce read-only semantics in type relations. Currently, the `readonly` modifier prohibits assignments to properties so marked, but it still considers a `readonly` property to be a match for a mutable property in subtyping and assignability type relationships. With this PR, `readonly` properties are required to remain `readonly` across type relationships:

```ts
type Box<T> = { value: T };
type ImmutableBox<T> = { readonly value: T };

const immutable: ImmutableBox<string> = { value: "hello" };
const mutable: Box<string> = immutable;  // Error with --enforceReadonly
mutable.value = "ouch!";
```

When compiled with `--enforceReadonly` the assignment of `immutable` to `mutable` becomes an error because the `value` property is `readonly` in the source type but not in the target type.

The `--enforceReadonly` option also validates that derived classes and interfaces don't violate inherited `readonly` constraints. For example, an error is reported in the example below because it isn't possible for a derived class to remove mutability from an inherited property (a getter-only declaration is effectively `readonly`, yet assignments are allowed when treating an instance as the base type).

```ts
interface Base {
    x: number;
}

interface Derived extends Base {  // Error with --enforceReadonly
    get x(): number;
}
```

In type relationships involving generic mapped types, `--enforceReadonly` ensures that properties of the target type are not more mutable than properties of the source type:

```ts
type Mutable<T> = { -readonly [P in keyof T]: T[P] };

function foo<T>(mt: Mutable<T>, tt: T, rt: Readonly<T>) {
    mt = tt;  // Error with --enforceReadonly
    mt = rt;  // Error with --enforceReadonly
    tt = mt;
    tt = rt;  // Error with --enforceReadonly
    rt = mt;
    rt = tt;
}
```

The `--enforceReadonly` option slightly modifies the effect of `as const` assertions and `const` type parameters to mean "as const as possible" without violating constraints. For example, the following compiles successfully with `--enforceReadonly`:

```ts
const obj: { a: string, b: number } = { a: "hello", b: 42 } as const;
```

whereas the following does not:

```ts
const c = { a: "hello", b: 42 } as const;  // { readonly a: "hello", readonly b: 42 }
const obj: { a: string, b: number } = c;  // Error
```

Some examples using `const` type parameters:

```ts
declare function f10<T>(obj: T): T;
declare function f11<const T>(obj: T): T;
declare function f12<const T extends { a: string, b: number }>(obj: T): T;
declare function f13<const T extends { a: string, readonly b: number }>(obj: T): T;
declare function f14<const T extends Record<string, unknown>>(obj: T): T;
declare function f15<const T extends Readonly<Record<string, unknown>>>(obj: T): T;

f10({ a: "hello", b: 42 });  // { a: string; b: number; }
f11({ a: "hello", b: 42 });  // { readonly a: "hello"; readonly b: 42; }
f12({ a: "hello", b: 42 });  // { a: "hello"; b: 42; }
f13({ a: "hello", b: 42 });  // { a: "hello"; readonly b: 42; }
f14({ a: "hello", b: 42 });  // { a: "hello"; b: 42; }
f15({ a: "hello", b: 42 });  // { readonly a: "hello"; readonly b: 42; }
```

Stricter enforcement of `readonly` has been debated ever since the modifier was introduced eight years ago in #6532. Our rationale for the current design is outlined [here](https://github.com/Microsoft/TypeScript/pull/6532#issuecomment-174356151). Given the huge body of existing type definitions that were written without deeper consideration of read-only vs. read-write semantics, it would be a significant breaking change to strictly enforce `readonly` semantics across type relationships. For this reason, the `--enforceReadonly` compiler option defaults to `false`. However, by introducing the option now, it becomes possible to gradually update code bases to correct `readonly` semantics in anticipation of the option possibly becoming part of the `--strict` family in the future.

Fixes #13347.


ahejlsberg:
@typescript-bot test it

weswigham:
@typescript-bot pack this

jakebailey:
Can you update `tests/cases/compiler/APILibCheck.ts` and enable this flag? This would let us ensure our public API will work with this flag; we already do this for `exactOptionalPropertyTypes`.

tests/cases/compiler/enforceReadonly1.ts
Andarist, line 1: Mapped type relationships should also check this:
```ts
function test<T>(a: T) {
  let obj1: { [K in keyof T]: T[K] } = null as any;
  let obj2: { readonly [K in keyof T]: T[K] } = null as any;

  obj1 = obj2; // should error
  obj2 = obj1;
}
```

Currently both assignments are OK: [TS playground (using this PR)](https://www.typescriptlang.org/play?enforceReadonly=true&ts=5.5.0-pr-58296-8#code/PTAEAEFMDsDMHsBOBjSAlSBDAJvaAbATwC5QAXRAV0gCgbZLpkyBLPcyAZzIB4AVAHwAKTKT4BKUAG8aoUPkhlQ8AEYArAIykpoANoBpUC2igA1pELxYoPgF0xB26AC+oALyholfPlCZOftCEANyy8orK6gBM2qCIWLgEhHqGxmYWVjb2No4u7p7evv6BIXRyqpr5FVHBoCCgnAAW8N7YoJCIiEhh1VXqGqHONEA)

robpalme:
nit: In the PR description the second code example does not match the preceding wording - it doesn't use `readonly`

fatcerberus:
Any reason why this isn't called `strictReadonly`?

ahejlsberg:
> Any reason why this isn't called `strictReadonly`?

Same reason as `--exactOptionalPropertyTypes`. It's too breaky to be in the `--strict` family of options and not necessarily desired by everyone.

ahejlsberg:
> nit: In the PR description the second code example does not match the preceding wording - it doesn't use readonly

Right, but only having a `get` accessor implies that the property is `readonly`. But apparently that isn't obvious.

fatcerberus:
> It's too breaky to be in the `--strict` family of options and not necessarily desired by everyone.

Alright, that makes sense.  That being said (and I realize this is just :bike: :house: at this point) I'm still not a big fan of `enforceReadonly`, as the implied opposite of "--enforceReadonly" is "don't enforce readonly", which... isn't really accurate.

src/compiler/checker.ts
```diff
@@ -13874,27 +13879,28 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
-            const optionality = getCombinedMappedTypeOptionality((type as IntersectionType).types[0]);
-            return every((type as IntersectionType).types, (t, i) => i === 0 || getCombinedMappedTypeOptionality(t) === optionality) ? optionality : 0;
+            const modifiers = getEffectiveMappedTypeModifiers((type as IntersectionType).types[0], mask);
+            return every((type as IntersectionType).types, (t, i) => i === 0 || getEffectiveMappedTypeModifiers(t, mask) === modifiers) ? modifiers : 0;
         }
         return 0;
     }
 
+    // Return -1, 0, or 1, where -1 means optionality is stripped (i.e. -?), 0 means optionality is unchanged, and 1 means
+    // optionality is added (i.e. +?).
```
Andarist, line 13898: nit:
```suggestion
    // Return -1, 0, or 1, where -1 means modifier is stripped (i.e. -?), 0 means modifier is unchanged, and 1 means
    // modifier is added (i.e. +?).
```

RyanCavanaugh:
Tried this on a [personal project](https://github.com/SeaRyanC/firaga-io) (a few thousand LOC) I wrote in the 4.3 timeframe and only got one (1!!) error. Looking ahead with skipLibCheck off, there's some stuff to address in `@types/node`. I'll do a DT PR once the beta is out
```
node_modules/@types/node/globals.d.ts:85:13 - error TS2403: Subsequent variable declarations must have the same type.  Variable 'AbortSignal' must be of type '{ new (): AbortSignal; prototype: AbortSignal; abort(reason?: any): AbortSignal; any(signals: AbortSignal[]): AbortSignal; timeout(milliseconds: number): AbortSignal; }', but here has type '{ new (): AbortSignal; prototype: AbortSignal; }'.

85 declare var AbortSignal: {
               ~~~~~~~~~~~

  ../typescript/built/local/lib.dom.d.ts:2360:13
    2360 declare var AbortSignal: {
                     ~~~~~~~~~~~
    'AbortSignal' was also declared here.

node_modules/@types/node/http2.d.ts:61:22 - error TS2430: Interface 'Http2Stream' incorrectly extends interface 'Duplex'.
  Property 'destroyed' is 'readonly' in the source but not in the target.

61     export interface Http2Stream extends stream.Duplex {
                        ~~~~~~~~~~~

node_modules/@types/node/net.d.ts:75:11 - error TS2415: Class 'Socket' incorrectly extends base class 'Duplex'.
  Property 'destroyed' is 'readonly' in the source but not in the target.

75     class Socket extends stream.Duplex {
             ~~~~~~

node_modules/@types/node/stream.d.ts:480:15 - error TS2420: Class 'Writable' incorrectly implements interface 'WritableStream'.
  Property 'writable' is 'readonly' in the source but not in the target.

480         class Writable extends Stream implements NodeJS.WritableStream {
                  ~~~~~~~~

```

tests/baselines/reference/enforceReadonly1.errors.txt
```diff
@@ -58,6 +58,22 @@ enforceReadonly1.ts(97,5): error TS2322: Type 'ImmutableValue<string>' is not as
         rt = tt;
     }
     
+    // A read-only property is assignable to a property declared as a method
```
Andarist, line 61: it might be good to add rationale behind this in the PR's description

tests/baselines/reference/enforceReadonly1.errors.txt
```diff
@@ -99,9 +115,6 @@ enforceReadonly1.ts(97,5): error TS2322: Type 'ImmutableValue<string>' is not as
     }
     
     class D4 extends B4 {  // Error
```
Andarist, line 117: this `// Error` comment is now outdated

ahejlsberg:
Latest commits exclude methods from strict checking. A bit of rationale is in order. Consider scenarios like this:

```ts
interface Base {
    foo(): void;
}

interface Item extends Base {
    name: string;
}

interface ReadonlyItem extends Base {
    readonly name: string;
}

declare let item1: ReadonlyItem;
declare let item2: Readonly<Item>;

item1 = item2;  // Error?
```

It seems unreasonable to error on the assignment, yet the `foo` property in `Readonly<Item>` is a `readonly` function type property, whereas the `foo` property in `ReadonlyItem` is a method that isn't and can't be marked `readonly`. Previously this wasn't an issue because we never enforced `readonly` semantics across type relationships.

Methods have always been considered mutable, and we don't even allow `readonly` modifiers on method declarations. TBH, if we did, I think most users would read it as "this method doesn't mutate the object", which wouldn't at all be the case. Interestingly, we've had very few requests for `readonly` methods. It doesn't really seem to be a pain point, and mutation is commonly used in interception patterns, function prototype construction, etc.

We could consider having mapped types with `readonly` modifiers _not_ add `readonly` to properties that originate in methods, but that just doesn't seem right either. I mean, `Readonly<T>` is supposed to make every property `readonly`.

In my opinion, the best course of action is to simply exclude methods from strict `readonly` checking. Meaning that, in a type relationship, when a target property is declared as a method, we don't care about the `readonly` state of the source property.

Conaclos:
> we've had very few requests for readonly methods. It doesn't really seem to be a pain point, and mutation is commonly used in interception patterns, function prototype construction, etc.

There are some interest expressed by #22315
We could add another flag that extends the check of `enforceReadonly` to methods and make methods readonly by default.

c-vetter:
@ahejlsberg 
> It seems unreasonable to error on the assignment, [..]
> 
> Methods have always been considered mutable, and we don't even allow `readonly` modifiers on method declarations. TBH, if we did, I think most users would read it as "this method doesn't mutate the object", which wouldn't at all be the case.

I disagree here. Given `enforceReadonly`, I'd expect `item1 = item2;  // Error?` to error. Also, I don't see why methods shouldn't be allowed to be declared `readonly` – I can (and do) already do this, and it works as I expect:
```ts
export type Type = {
	readonly toString: () => string
}
```

I'd like to be able to do this, purely based on consistency (I don't really need two ways to describe the same thing, so ¯\\­\_(ツ)\_/¯):
```ts
export type Type = {
	readonly toString () : string
}
```

Wrapping `Readonly<>` around types with methods also gives us that behavior. So, from my point of view, disallowing readonly on `readonly toString () : string` is just an unnecessary hurdle because it merely hinders a feature that is available via different avenues anyway:
```ts
class X {
	readonly toString : () => string = () => ``
}

class Y {
	readonly toString () { // no reason why exactly this should error…
		return ``
	}
}

class Z {
	get toString () {
		return () => ``
	}
}

const x = new X()
const y = new Y()
const z = new Z()

x.toString = ()=>``
y.toString = ()=>`` // …or why exactly this should not
z.toString = ()=>``
```

Also, while I consider inline assignment dangerous enough, function parameters are even more dangerous because you usually do not see the code that works on your supposedly readonly data:
```ts
const readonlyArray = [] as const

mutateArray(readonlyArray) // The type 'readonly []' is 'readonly' and cannot be assigned to the mutable type 'unknown[]'.ts(2345)

function mutateArray (array : Array<unknown>) {
	array.shift()
}

const readonlyObject = { toString () { return `` } } as const

mutateObject(readonlyObject) // This should error

function mutateObject (mutable : { toString() : string }) {
	mutable.toString = () => `-`
}
```

`mutateObject(readonlyObject)` should error because otherwise, you have no guarantees that the function doesn't do something to your methods. This introduces an opportunity for runtime errors that seems preventable by making the behavior consistent.

And I don't count “interception patterns, function prototype construction, etc.” as strong counter-arguments, because:
1. [`Proxy`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy)
2. The prototype can be built as a mutable object and then assigned as a readonly prototype
3. For runtime-safe tasks, there are usually options that work with the type system

If any specific project absolutely cannot work with this, then I guess that `enforceReadonly` is generally unfit anyway.

@Conaclos 
> We could add another flag that extends the check of `enforceReadonly` to methods and make methods readonly by default.

I'm unsure about the second part – I'd like to make everything readonly by default 😅 and I fear that's out of scope here…
But a flag to treat methods like everything else would be good. After all, JS doesn't treat them differently, does it?

src/lib/dom.generated.d.ts
```diff
@@ -1275,8 +1275,8 @@ interface PushSubscriptionOptionsInit {
 }
 
 interface QueuingStrategy<T = any> {
-    highWaterMark?: number;
-    size?: QueuingStrategySize<T>;
+    readonly highWaterMark?: number;
+    readonly size?: QueuingStrategySize<T>;
```
jakebailey, line 1279: I'm not sure we can take this until the generator is updated to do this, right?

src/server/project.ts
```diff
@@ -392,8 +392,8 @@ export abstract class Project implements LanguageServiceHost, ModuleResolutionHo
 
     public languageServiceEnabled: boolean;
 
-    readonly trace?: (s: string) => void;
-    readonly realpath?: (path: string) => string;
+    trace?(s: string): void;
+    realpath?(path: string): string;
```
jakebailey, line 396: Were these changes required to make the flag pass on our codebase, or needed for the PR at all?

erights:
While I welcome this direction, please do not use the term "enforce" for things that are not actually enforced. TypeScript is unsound and erased, so it cannot and does not actually enforce anything. It is useful for catching some many accidents, which is great. But some accidents will slip through the unsoundness, and much malice will. The term "enforce" implies otherwise and will give some people a false sense of security, which is dangerous.

RyanCavanaugh:
@erights did you have a synonym in mind?

> But some accidents will slip through the unsoundness, and much malice will. 

No one should be using a language with `any` in an adversarial situation. What scenarios are you thinking of here?

erights:
`Object.freeze`, which is the fundamental enforcement mechanism for read-only-like stability, was introduced in ES5 specifically to be genuine enforcement which could be used in adversarial situations. Since then, `Object.freeze` has been used as an adversarial enforcement mechanism many times.

We are of course using Hardened JavaScript and `harden` in many adversarial situations. We also use TypeScript in various ways in that context. Better TypeScript readonly support is welcome for many things, including better typing of `harden`. However, we already find we need to continually remind users of TypeScript-on-Hardened-JS that the TypeScript typing portions of their code are unsound and erased, and therefore unenforced. When they want to enforce something, they need to rely on unerased runtime enforcement mechanisms. That's fine. 

But seeing the term "enforce" in the documentation of the unenforced TypeScript properties will only deepen the confusion and make the explanations harder. It is harmful and confusing.

turadg:
> a synonym in mind?

`checkReadonly` implies a check for type system violations without implying that they're enforced.

Another option is `strictReadonly` since some aspects of `readonly` are already checked and this is a more strict form.

src/lib/dom.generated.d.ts
ahejlsberg: Right, this ultimately needs to be done in the DOM generator. However, without this workaround it's pretty much impossible to compile anything because `lib.dom.d.ts` fails to check properly with `--enforceReadonly`:

```ts
../../ts/built/local/lib.dom.d.ts(3236,11): error TS2430: Interface 'ByteLengthQueuingStrategy' incorrectly extends interface 'QueuingStrategy<ArrayBufferView>'.
  Property 'highWaterMark' is 'readonly' in the source but not in the target.
../../ts/built/local/lib.dom.d.ts(5900,11): error TS2430: Interface 'CountQueuingStrategy' incorrectly extends interface 'QueuingStrategy<any>'.
  Property 'highWaterMark' is 'readonly' in the source but not in the target.
```

src/server/project.ts
ahejlsberg: Actually, doesn't look like this change is necessary anymore. I'll back it out.

tests/baselines/reference/api/typescript.d.ts
```diff
@@ -2821,8 +2821,8 @@ declare namespace ts {
             private hasAddedOrRemovedSymlinks;
             protected languageService: LanguageService;
             languageServiceEnabled: boolean;
-            readonly trace?: (s: string) => void;
-            readonly realpath?: (path: string) => string;
+            trace?(s: string): void;
+            realpath?(path: string): string;
```
jakebailey, line 2825: I think these need to be updated (noticed the change without baseline update).

(Not sure if any other ones can be reverted either)

ahejlsberg:
Regarding the name of the flag, we're perfectly happy to entertain suggestions for better options.

The issue with `--checkReadonly` is it would imply that no `readonly` checks occur when it is disabled, which isn't the case. We always disallow assignments to `readonly` properties.

The issue with `--strictReadonly` is that it implies being in the `--strict` family of flags. That's not the case because it would be too significant of a breaking change.

jakebailey:
Oh come on, they weren't supposed to comment that (must be the unlucky first PR, sorry)

erights:
> Regarding the name of the flag, we're perfectly happy to entertain suggestions for better options.

Glad to hear it! Before trying to generate more suggestions, I'd like to better understand the feature. 

What surprises me is that it is a compiler flag at all, rather than a stronger form of readonly within the language alongside the existing one. As a flag, it changes meaning of existing code. Let's call the existing semantics the "weaker" constraints, and with the flag, "stronger" constraints. (I am not suggesting these names. Just need terminology to ask questions.)

What happens if I have some modules or packages that are correctly written to the weaker constraints, and other modules or packages written correctly to the stronger constraints, and now I want to correctly compose them together into a larger system. Can I statically type check this composite system? Clearly, if I check the composite only with the stronger constraints, the modules correct only by the weaker constraints will fail. If I check the composite only with the weaker constraints, if all the inputs were correct on their own terms, the composite would still pass. But the code written to the stronger constraints could come to violate them without causing a static error.

Is all this correct?

mhofman:
@typescript-bot pack this

jakebailey:
@typescript-bot pack this

ahejlsberg:
> Before trying to generate more suggestions, I'd like to better understand the feature.

The rationale for the existing functionality of `readonly` is outlined [here](https://github.com/Microsoft/TypeScript/pull/6532#issuecomment-174356151). The modifier was introduced several years after the initial release of TypeScript, and even then it was too large of a breaking change to check it strictly. Therefore, we decided to just error on direct assignments to `readonly` properties, but otherwise ignore the modifier for purposes of relations between types containing `readonly` properties.

The `--enforceReadonly` option (or whatever we call it) implements that final step of checking that `readonly` isn't defeated by treating a read-only property as a mutable property through a conversion of the containing type. We know this will be a breaking change for a lot of code, but there is a subset of the community for whom updating their code is a worthwhile cost. Over time we may see enough code updated to the stricter semantics that we could consider including the option in the `--strict` family of flags.

> What surprises me is that it is a compiler flag at all, rather than a stronger form of readonly within the language...

The surprising thing really is that we _aren't_ checking `readonly` in type relations and `--enforceReadonly` at least puts us on a path to doing so. I think that introducing another stronger readonly-like modifier would just add to the confusion.

> What happens if I have some modules or packages that are correctly written to the weaker constraints...

What happens is that you'll see new errors pointing out places where your code potentially _indirectly_ violates `readonly` by treating read-only properties as mutable properties through type conversions. These errors may be corrected by adding or removing `readonly` modifiers as appropriate, or by adding type assertions that explicitly call out the unsound behavior. In all cases, code that is updated to compile with `--enforceReadonly` will continue to compile without the switch.

We aren't providing a way to control the checking on an individual module or type basis, mostly because it becomes very hard to reason about. For example, what does it mean to derive a "weak" type from a "strong" type, and, in type relations, what happens when you mix "weak" and "strong" types? We have multiple other options (e.g. the entire `--strict` family of flags) that are global in nature in the same manner.

erights:
> We have multiple other options (e.g. the entire --strict family of flags) that are global in nature in the same manner.

Can we just do that?

JoostK:
I do understand the concern around breakages this would cause if it falls under the `strict` family of flags, but I do feel it's rather unfortunate; enabling `strict`, to me, opts you into TypeScript in its strictest form, whatever that may be in any given version. Breakages can be somewhat easily addressed by disabling a newly introduced flag, or by actually fixing the newly-reported errors. Assuming folks opt into `strict` because they desire the least room for "unsound" code, having to adopt a new flag—that may even become deprecated if the flag is eventually promoted to the `strict` family—during an update to benefit from _better_ strictness seems undesirable to me.

I do have a concrete proposal: let `strict` be an option that takes a version string to be able to tell which strict flags have been opted into, preventing future breakages for additional new strict flags. I believe this is similar to the recent work on a deprecation flag. That does raise the question as to whether `true` should mean _everything_ or `"5.5"`, where I think the former sounds like the more desirable behavior but that would leave the concern that _this_ change would still break apps (and I understand how this would break the "top ###" tests and potentially the performance tests that TS itself relies on for testing).

JoostK:
Alternatively, perhaps this should just be an `"edition"` flag (or whatever name makes sense, I'm using the Rust terminology here) as a high level switch that determines how the options are interpreted.

RobertSandiford:
> I do understand the concern around breakages this would cause if it falls under the `strict` family of flags, but I do feel it's rather unfortunate; enabling `strict`, to me, opts you into TypeScript in its strictest form, whatever that may be in any given version.

This not the current effect of "strict". For example both "noUncheckedIndexedAccess" and "useDefineForClassFields" improve safety.

That said some sort of versioned "strict" does seem reasonable.

src/lib/es2015.collection.d.ts
```diff
@@ -30,7 +30,7 @@ interface Map<K, V> {
 interface MapConstructor {
     new (): Map<any, any>;
     new <K, V>(entries?: readonly (readonly [K, V])[] | null): Map<K, V>;
-    readonly prototype: Map<any, any>;
+    prototype: Map<any, any>;
```
rbuckton, line 33: I'm concerned with this change as `prototype` is actually read-only on `Map`, `Set`, `WeakMap`, `WeakSet`, etc.:

image

JustinBordage:
> The issue with `--strictReadonly` is that it implies being in the `--strict` family of flags. That's not the case because it would be too significant of a breaking change.

@ahejlsberg I'm confused, how does it's name imply it's significance or expected significance as a breaking change? Also, why would me implying that it's part of the `--strict` family of flags be a bad thing? The way I interpret `--strictReadonly` is that I'd be more constrained with the flag enabled than without it. Which I believe fits the linguistic meaning of "strict", would it not?

Note: I'm interpreting this from a more surface level understanding of the concepts, since this level of discussion is generally outside my realm of expertise. But I felt the need to contribute since, I just recently encountered a use-case where not having this flag could prove to be dangerous as I don't have a means of gracefully handling this risk other than hoping the dev reads the jsdocs on the function in question.

As for flag name suggestions, if `--strictReadonly` is not appropriate, then perhaps something along the lines of `--noReadonlyCoercion`?

RobertSandiford:
> Edit: Oh I think I may have figured it out, my initial interpretation was that `--strict` was just a prefix and nothing more. However, if I've understood correctly, `--strict` is actually it's own flag which includes the effects of all the other flags that are prefixed with `--strict`?

"strict" enables a variety of flags, some prefixed with strict. It doesn't enable all safety features.

petamoriken:
The `prototype` properties need to be treated specially.

Perhaps we might need `// @ts-ignore` for `interface DOMMatrix extends DOMMatrixReadOnly` in the DOM types.

MichaelMitchell-at:
@jakebailey would it be much effort to rebase and pack this branch? I briefly tried to resolve the conflicts but felt I didn't have enough context around the changes.

HansBrende:
The FAQ section of the wiki [states](https://github.com/microsoft/TypeScript/wiki/FAQ#enforce-readonly-in-subtyping--assignability):

> Enable --enforceReadonly (available in TypeScript 5.6; see [#58296](https://github.com/microsoft/TypeScript/pull/58296))

Seems like this functionality is not actually available yet, considering this PR is still open? Should the wiki be updated to say "coming soon" or something? Somewhat confusing as-is.

mkantor:
@HansBrende you could have proposed that edit yourself. I did so [here](https://github.com/microsoft/TypeScript-wiki/pull/344).

Ghabriel:
Is there something currently blocking this PR? This would be an amazing feature to have — I often have to refactor codebases that use a lot of mutability to try and reduce overall mutations, and sadly the current behavior of `readonly` doesn't give me nearly enough confidence that functions which take readonly arguments do in fact "keep their word" and leave them as-is.

zimmed:
Apparently I have gaslit myself. For years, I thought this was already enforced behavior and recently discovered it wasn't.

It's hard to believe that this is valid in typescript:

```typescript
interface ImmutablePoint {
  readonly x: number;
  readonly y: number;
}

class FixedPoint  {
  get x() { return 5; }
  get y() { return 10; }
}

function move(mutablePoint: { x: number, y: number }) {
  mutablePoint.x += 1;
  mutablePoint.y += 1;
}

const point: ImmutablePoint = { x: 1, y: 2 };

move(point);
move(new FixedPoint());
```

Let's get this merged, please.

AFatNiBBa:
I too came across this problem, are there any news on this?

lishaduck:
> I too came across this problem, are there any news on this?

TypeScript is in ~feature freeze until Corsa lands, so we've got to wait at least another year 😢

AFatNiBBa:
Quite unfortunate ☹️ 

In the meantime, would this pull request change the behaviour of this code?
```ts
function f<K extends PropertyKey, T extends { [k in K]: number }>(obj: T, k: K) {
    // Type 'number' is not assignable to type 'T[K]'.
    //     'number' is assignable to the constraint of type 'T[K]', but 'T[K]' could be instantiated with a different subtype of constraint 'number'.
    obj[k] = 1;
    return obj[k];
}

const k = f({ a: 1 }, "a");
//    ^? const k: number
```
Maybe I'm wrong, but I think that `T[K]` should technically be allowed to be a sub-type of `number` (Instead of exactly `number`) only if `K` was readonly, precisely because it can't be set otherwise.
If I'm not mistaken, the way of saying it technically would be that "It shouldn't assume covariance"

RobertSandiford:
> In the meantime, would this pull request change the behaviour of this code?

I don't see why it would. There are no `readonly` markers here to begin with.

> Maybe I'm wrong, but I think that `T[K]` should technically be allowed to be a sub-type of `number` (Instead of exactly `number`) only if `K` was readonly, precisely because it can't be set otherwise. If I'm not mistaken, the way of saying it technically would be that "It shouldn't assume covariance"

TypeScript's assignability rules act as if all objects are readonly, so uses covariance rules. This makes most mutable objects unsafe as reading does not support contravariance and writing does not support covariance, so only objects of exactly the same type should be allowed. E.g the follow code is broken:

```ts
const o = { a: 1 }
function set(o: { a: string | number }) {
    o.a = 'foo'
}
set(o)
console.log(o.a.toPrecision(2)) // err: o.a.toPrecision is not a function 
```

In your example it is analyzing a bit more deeply than normal and is actually behaving correctly. (Removing the generic T and just using `{ [k in K]: number }` as the type for `obj` removes the error but is unsafe, as with the code above.

To be safe your code needs to specify that `obj[K]` is exactly `number` which just isn't naturally supported. There's no syntax for specific exactness or marking an argument as accepting contravariant input instead of covariant. You can fudge it with a validator type that uses an extends check to check that `obj[K]` is not narrower than a number, but you'll need to use casting inside the function because now you have a conditional type for obj that can't be assigned to directly: [Playground](https://www.typescriptlang.org/play/?#code/C4TwDgpgBAogHgQwMbADYgHIFcC2AjCAJwB4BpKCOYCAOwBMBnKABUIHtJDRSIQAaKABUKVWoygBvKAG0A1lACWNKKQC6ALig1cBQlAC+APigBeLTqIjq9JoOlqoAfiFRNNCADciAWABQfgDMsGhQFNmUAsisxJlYOIm5eAWFKa3EpOUVlNTcLPSMACjY8ACtNeGQ0TDyyZMMBWU1SAEpJPygOqCLSqAQmDPklFQ1zfEt9ZrlVUygAJgBuds7CCGAsQmVikqnF330-PyRwhmAoNhmpBE0ARgND49P5MwCigQAiBDfm+agAel+oAB5UgHXxHGgnKBgC69G4GXpMcEne4Q06oGYvMDvT7fP4AgCChAA5rhaKc2AEoKBIFAAORSFYIOjhdCwqDXH76WmKJg0NinPoMBREmgIPCoaDAc5gBCEBA4VaWClU8DQWnuLyEWlAA)

The solution would be something along the lines of this: https://github.com/microsoft/TypeScript/issues/9252 (expanded to apply to object properties, and with support for exact match only as well as supertype.

AFatNiBBa:
What I meant was that it seems that TypeScript assignability rules are based on readonly checks not being strict.
With this flag, covariance ans contravariance could be enforced in the right places, since now there would be no need to assume every prop is readonly.
This said, I understand that a change of this magnitude could bring more harm than good

RobertSandiford:
> What I meant was that it seems that TypeScript assignability rules are based on readonly checks not being strict.
With this flag, covariance ans contravariance could be enforced in the right places, since now there would be no need to assume every prop is readonly.

TS _could_ move to exact match only for mutable objects, and require readonly props for covariant assignability. With a writeonly flag they could allow contravariant props. But this is separate to this change, and could happen without this. With that change and not this ticket you'd be able to assign `{ readonly a: T|U }` to `{ a: T|U }`, and wouldn't be able to assign `{ readonly a: T }` to `{ a: T|U }` (covariant). This is readonly breaking as already exists, but certainly not a blocker to going ahead with the other change.

The above though is unlikely to happen due to the issues it would cause in existing codebases. Calls to libraries that expect objects and only read, that rely on covariant types being allowed (mega common) would break, unless those libs were updated with readonly keywords. A more likely change would be a strict flag that made objects received by functions readonly by default, with a mutable flag to enable writing and stricter assignability, because this would move to issue upstream into the libraries with nothing breaking unless a lib identified that it was writing to an object and needed to add the mutable keyword, and of course downstream projects wouldn't break until they updated those libs. Plus breaks would only occur where writing was actually happening, rather than many false positives.

Issue #58011, opened 2024-03-31
Increased memory usage when creating a watch program that includes local JavaScript files in node_modules


reduckted:
### 🔎 Search Terms

Reached heap limit Allocation failed out of memory createWatchCompilerHost createWatchProgram

### 🕗 Version & Regression Information

- This changed in commit 385db449f5395ab4d8f24c5f531e85aebbe228c7  ; PR #56946

### ⏯ Playground Link

https://github.com/reduckted/repro-typescript-memory-usage

### 💻 Code

This cannot be reproduced with a few lines of code. It needs a large TypeScript project. Use these steps to see the increased memory usage:

1. Clone the repository `git clone https://github.com/reduckted/repro-typescript-memory-usage`
2. Run `npm i`
3. Run `node run.js` and see the memory usage (this will be for TypeScript 5.4.3).
4. Install TypeScript 5.3.3 `npm i typescript@5.3.3 -D`
5. Run `node run.js` and see the memory usage.

For the linked repository, I am seeing 525 MB used in 5.3.3 and 775 MB used in 5.4.3. The larger the TypeScript project, the worse the memory usage is.

### 🙁 Actual behavior

The memory consumed by TypeScript 5.4 is much larger compared to 5.3. The larger the TypeScript project, the worse the memory usage is. Large projects can cause Node.js to run out of memory.

### 🙂 Expected behavior

TypeScript should have equivalent memory usage to v5.3.

### Additional information about the issue

What I've been able to determine is that when you have a JavaScript project in the repository and install it as a node module, creating a "watch" program that includes files that `require` files from that project will consume significantly more memory in 5.4.3 compared to 5.3.3.

I encountered this problem when running ESLint with a config that uses the `@typescript-eslint/*` plugins. Using TypeScript 5.3.3, ESLint would run fine. After upgrading to 5.4.3, ESLint would crash with an out of memory error:

```
FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
```

The repository that this was occurring on is an internal, closed-source company project, so I can't share it and signing an NDA will not be possible. However, I have been able to see an increased memory usage with a similar project setup in other projects. The repository I have linked above is a clone of ionic-team/ionic-framework with a few additions to it.

This is roughly the setup in the closed-source repository:

- All TypeScript code is contained under a `source/` directory.
- We have some JavaScript projects that are for various build tools. One of them is for our Storybook configuration.
[...]

PR #58299, opened 2024-04-24
Don't reuse program structure if we errored on filenames differing by case

jakebailey:
This is speculation based on https://github.com/microsoft/TypeScript/issues/58011 and https://github.com/typescript-eslint/typescript-eslint/issues/1192. Not 100% certain it'll help, and can't test without a repro...


jakebailey:
@typescript-bot pack this

sheetalkamat (changes requested):
Dont think this is going to cause any change in memory.. Also not reusing because there is error is more perf impacting as well. We need concrete repro to actually identify the issue. The linked issue talks about problem only with programatic usage so i would say there is something there but without repro its hard to just speculate

Issue #57475, opened 2024-02-22
TypeScript 5.5 Iteration Plan


DanielRosenwasser:
This document outlines our focused tasks for TypeScript 5.4. It minimally indicates intent to investigate tasks or contribute to an implementation. Nothing is set in stone, but we will strive to complete these tasks in a reasonable timeframe.

Date           | Event
---------------|-------------------------
2024-03-05     | TypeScript 5.4 Release
2024-04-12     | Create 5.5 Beta (5.5.0) Build for Testing
2024-04-16     | **TypeScript 5.5 Beta Release**
2024-05-31     | Create 5.5 RC (5.5.1) Build for Testing
2024-06-04     | **TypeScript 5.5 RC Release**
2024-06-14     | Create 5.5 Final (5.5.2) Build for Testing
2024-06-18     | **TypeScript 5.5 Final Release** 🚀

```mermaid
gantt
    dateFormat  YYYY-MM-DD
    TypeScript 5.4 Stabilization Period : 2024-02-16, 2024-03-05
    TypeScript 5.5 Beta Development : 2024-02-16, 2024-04-12
    TypeScript 5.5 RC Development : 2024-04-13, 2024-05-31
    TypeScript 5.5 Stabilization Period : 2024-06-01, 2024-06-18
todayMarker stroke-width:5px,stroke:#0f0,opacity:0.5
```

# Compiler and Language

* [The `--isolatedDeclarations` Flag](https://github.com/microsoft/TypeScript/pull/53463)
* [Type-Only Imports in JSDoc](https://github.com/microsoft/TypeScript/pull/57207)
* [Non-Null Assertions in JSDoc](https://github.com/microsoft/TypeScript/pull/57207)
* [Regex Syntax Validation](https://github.com/microsoft/TypeScript/pull/55600)
* [Checks for Never-Initialized Variables](https://github.com/microsoft/TypeScript/pull/55887)
* [Allow Implementing Functions That Return Conditional Types](https://github.com/microsoft/TypeScript/issues/33014)
* [Improvements in Extending `tsconfig.json` Files](https://github.com/microsoft/TypeScript/issues/56436)
* [Investigate Recursive Object Type References](https://github.com/microsoft/TypeScript/issues/57034)
* [Investigate Speculation Infrastructure for Overload Resolution](https://github.com/microsoft/TypeScript/pull/57421)
* [Provide Better `default` Import Interop Configurability](https://github.com/microsoft/TypeScript/issues/54102)
[...]

PR #58306, opened 2024-04-24
Disable no-inferrable-types eslint rule

jakebailey:
This should fix the `create release-X.Y` task.

This was a stylistic rule introduced when we switched to using ts-eslint's presets, but we explicitly annotate some stuff that's inferable on purpose. Since this is just style, it seems okay to just disable it.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -127,6 +127,7 @@
                 "varsIgnorePattern": "^(_+$|_[^_])"
             }
         ],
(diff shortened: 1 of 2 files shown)


DanielRosenwasser (approved):

DanielRosenwasser:
Since users in #57475 got annoyed, I'll post it here.

@typescript-bot create release-5.5

Issue #57652, opened 2024-03-05
TypeScript fails to resolve dependency in monorepo given specific conditions


remcohaszing:
### 🔎 Search Terms

"composite references"

### 🕗 Version & Regression Information

- This changed between versions 4.6 and 4.7 (by introducing `module` option `nodenext`).

### ⏯ Playground Link

https://github.com/remcohaszing/typescript-bug

### 💻 Code

Since this bug is about resolving imports in a mono repo and composite project references, it requires more than one file.

`package.json`: defines workspaces.

```json
{
  "private": true,
  "workspaces": [
    "packages/*"
  ],
  "dependencies": {
    "typescript": "5.3.3"
  }
}
```

`packages/a/package.json` defines the dependant project

```json
{
  "name": "a",
  "version": "0.0.0",
  "type": "module",
  "exports": {
    ".": {
      "types": "./types/index.d.ts",
      "default": "./index.js"
    }
  }
}
```

`packages/a/tsconfig.json` declares the referenced TypeScript project. It uses types in JSDoc and emits the types definitions into the `types` directory.

```js
{
  "compilerOptions": {
    "checkJs": true,
    "composite": true,
    "declaration": true,
    "emitDeclarationOnly": true,
    "module": "nodenext",
    "outDir": "types"
  }
}
```

`packages/a/index.js` must exist. This is what can’t be resolved. The contents don’t matter.

```js
// Empty
```

`packages/a/test/index.js` imports `a`. This file must be in a subdirectory of the `a` project for the bug to appear.

```js
import 'a'
```

`packages/b/package.json` defines the project that depends on `a`.

```json
{
  "name": "b",
  "version": "0.0.0",
  "type": "module"
}
```

`packages/b/tsconfig.json` declares the TypeScript project that references `a`. It uses types in JSDoc and doesn’t emit (but that’t irrelevant). Important is the `noImplicitAny` option. Disabling it, makes the bug go away.

```json
{
  "references": [{"path": "../a"}],
  "compilerOptions": {
    "checkJs": true,
    "module": "nodenext",
    "noEmit": true,
    "noImplicitAny": true
  }
}
```

`packages/b/index.js` dynamically imports `a`. A static import won’t make the bug appear.

```js
import('a')
```

Now build project `b` by running `tsc --build packages/b`.

### 🙁 Actual behavior

TypeScript builds `a` as it’s a referenced by `b`. It then errors with the following output:

```
[...]

PR #58308, opened 2024-04-24
Treat the input file name from referenced project as "ts" rather than file extension it is when determining eliding file for program construction

sheetalkamat:
During `tsc -b` the resolutions are shared across projects if the setting match. This means that if you resolved your own package name before building the project, you will get the "input" file as the resolution. And when we go to project that references this package and config, the input file will be used as resolution. This normally works out because we have rule to redirect to d.ts file and report error. But the issue is that if the input file is js file, it has rules for eliding imports. So if the file is input file name from referenced project, treat it as d.ts file for purpose of deciding eliding of imports and whether to include file or not.

Fixes https://github.com/microsoft/TypeScript/issues/57652

### DIFF
--- a/src/compiler/program.ts
+++ b/src/compiler/program.ts
@@ -4106,7 +4106,8 @@ export function createProgram(rootNamesOrOptions: readonly string[] | CreateProg
                 }
 
                 const isFromNodeModulesSearch = resolution.isExternalLibraryImport;
-                const isJsFile = !resolutionExtensionIsTSOrJson(resolution.extension);
+                // If this is js file source of project reference, dont treat it as js file but as d.ts
+                const isJsFile = !resolutionExtensionIsTSOrJson(resolution.extension) && !getProjectReferenceRedirectProject(resolution.resolvedFileName);
                 const isJsFileFromNodeModules = isFromNodeModulesSearch && isJsFile && (!resolution.originalPath || pathContainsNodeModules(resolution.resolvedFileName));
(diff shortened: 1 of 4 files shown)


src/compiler/program.ts
```diff
@@ -4106,7 +4106,8 @@ export function createProgram(rootNamesOrOptions: readonly string[] | CreateProg
                 }
 
                 const isFromNodeModulesSearch = resolution.isExternalLibraryImport;
-                const isJsFile = !resolutionExtensionIsTSOrJson(resolution.extension);
+                // If this is js file source of project reference, dont treat it as js file but as d.ts
+                const isJsFile = !resolutionExtensionIsTSOrJson(resolution.extension) && !getProjectReferenceRedirectProject(resolution.resolvedFileName);
```
jakebailey, line 4110: I feel like I don't quite understand how this works; "is JS" seems like it should just be "has the JS extension", but this would say that if it came from a referenced project, we don't treat it as such? So, won't we misinterpret later checks regarding node module checks?

Is it the case that an equivalent fix would just be calling `getProjectReferenceRedirectProject` below in the `shouldAddFile` checks?

sheetalkamat: for all purposes we will be adding ".d.ts" file so we are treating it as such. That happens in findSourceFile where we map the "original file name" to output "d.ts" file name

jakebailey (approved):

PR #58310, opened 2024-04-24
Rephrase error messages on backrererences.

DanielRosenwasser:
Rephrases "decimal capturing group" as just a "backreference".

I would actually call out the backreference number, but the scanner's `ErrorCallback` only accepts one parameter, so that can be fixed up in the future.


jakebailey (approved):
LGTM but will probably race with the other regex PRs since that touches diags?

gabritto (approved):
I think the new error message is more clear.

PR #57111, opened 2024-01-21
Add lib declaration for RegExp#unicodeSets

Josh-Cena:
Fixes #57055

PR #58311, opened 2024-04-24
Move RegExp flag version mapping to LanguageFeatureMinimumTarget

jakebailey:
Not pressing, but this is a follow-up to #55600 to move the feature listing to the single enum we're now using to track features.


src/compiler/scanner.ts
```diff
@@ -2456,7 +2457,7 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
                 }
                 else {
                     regExpFlags |= flag;
-                    const availableFrom = regExpFlagToFirstAvailableLanguageVersion.get(flag)!;
+                    const availableFrom = regExpFlagToFirstAvailableLanguageVersion.get(flag)! as unknown as ScriptTarget;
```
jakebailey, line 2460: This is... awkward.

src/compiler/types.ts
```diff
@@ -8218,6 +8218,11 @@ export type EmitHelperUniqueNameCallback = (name: string) => string;
  * @internal
  */
 export const enum LanguageFeatureMinimumTarget {
+    // ES3 Features
+    RegularExpressionFlagsGlobal = ScriptTarget.ES3,
+    RegularExpressionFlagsIgnoreCase = ScriptTarget.ES3,
+    RegularExpressionFlagsMultiline = ScriptTarget.ES3,
```
jakebailey, line 8224: Probably these should have been unconditional; we should not be referencing this script target anywhere anymore but the original regex PR predated us removing ES3 so that's just how it is.

src/compiler/types.ts
```diff
@@ -8259,9 +8267,11 @@ export const enum LanguageFeatureMinimumTarget {
     TopLevelAwait = ScriptTarget.ES2022,
     ClassFields = ScriptTarget.ES2022,
     PrivateNamesAndClassStaticBlocks = ScriptTarget.ES2022, // `class C { static {} #x = y, #m() {} }`, `#x in y`
+    RegularExpressionFlagsHasIndices = ScriptTarget.ES2022,
 
     // ES2023 Features
     ShebangComments = ScriptTarget.ESNext,
+    RegularExpressionFlagsUnicodeSets = ScriptTarget.ESNext,
```
jakebailey, line 8274: I _think_ this is 2023 but I can't find the page where I read that.

Josh-Cena: This is ES2024! See https://tc39.es/ecma262/multipage/#sec-intro and https://github.com/microsoft/TypeScript/pull/57111

src/compiler/types.ts
sheetalkamat: lets make these unconditional and remove the target here given ES3 is deprecated.

src/compiler/scanner.ts
graphemecluster: Nit: the non-null assertion is not needed anymore

src/compiler/scanner.ts
```diff
@@ -293,17 +294,18 @@ const charToRegExpFlag = new Map(Object.entries({
...
-    [RegularExpressionFlags.Sticky, ScriptTarget.ES2015],
+const regExpFlagToFirstAvailableLanguageVersion = new Map<RegularExpressionFlags, LanguageFeatureMinimumTarget>([
+    [RegularExpressionFlags.HasIndices, LanguageFeatureMinimumTarget.RegularExpressionFlagsHasIndices],
+    [RegularExpressionFlags.DotAll, LanguageFeatureMinimumTarget.RegularExpressionFlagsDotAll],
+    [RegularExpressionFlags.Unicode, LanguageFeatureMinimumTarget.RegularExpressionFlagsUnicode],
+    [RegularExpressionFlags.UnicodeSets, LanguageFeatureMinimumTarget.RegularExpressionFlagsUnicodeSets],
+    [RegularExpressionFlags.Sticky, LanguageFeatureMinimumTarget.RegularExpressionFlagsSticky],
 ]);
 
+function getRegExpFlagToFirstAvailableLanguageVersion(flag: RegularExpressionFlags) {
```
sheetalkamat, line 305: Instead of this may be a change this function to do:
```ts
const availableFrom = getRegExpFlagToFirstAvailableLanguageVersion(flag);
                        if (availableFrom && languageVersion < availableFrom) {
                            error(Diagnostics.This_regular_expression_flag_is_only_available_when_targeting_0_or_later, p, 1, getNameOfScriptTarget(availableFrom));
                        }
```

sheetalkamat (approved):

rbuckton (approved):

Issue #51000, opened 2022-09-29
Flag Deprecation Plan


RyanCavanaugh:
## TypeScript Flag Deprecation Plan

> (This is a proposal, but written in terms of a blog post for clarity)

As we think about the long-term future of TypeScript and its place in the JavaScript ecosystem, it's clear that many things TypeScript 0.8 users in 2012 needed aren't things that TypeScript users in 2024 are going to need. For example, in 2012, Internet Explorer 8 (which only supports ECMAScript 3) had enough marketshare to warrant serious compatibility concern. Today, thankfully, it doesn't.

While we still consider long-term language stability to be an absolutely paramount concern, having a well-defined deprecation policy in place also makes it easier for us to be *slightly* less risk-averse when adding new features to the language.

To that end, starting with TypeScript 5.0, we will begin the process of removing certain flags, features, and/or behaviors which we think the language or ecosystem have largely evolved past the need for.

### Schedule

The deprecation schedule is based on our existing `version += 0.1` versioning cadence, aligned with ones-place-incrementing versions for clarity.

|Release / Phase|Description|
|----|------|
|TypeScript 5.0 - Phase 1|❔ Deprecation warnings introduced|
|TypeScript 5.5 - Phase 2|⚠ Flags start doing nothing|
|TypeScript 6.0 - Phase 3|❌ Flags removed entirely|

The cycle repeats again with TypeScript 6.0 also introducing whatever new deprecation warnings deemed appropriate at that time.

#### Phase 1: ❔ Deprecation Warnings

In phase 1 (e.g. TypeScript 5.0), using a deprecated flags will produce a commandline error:
```
TS9998: Flag 'keyofStringsOnly' is deprecated and will stop functioning in TypeScript 5.5. Specify 'ignoreDeprecations: "5.0"' to silence this error
```

As hinted to in the error, you can supply a new flag value to silence this error:
```
    ignoreDeprecations: "5.0"
```

All other behavior is unchanged.

This gives people an immediate warning of impending deprecation, but won't block them from upgrading to TypeScript 5.0.

#### Phase 2: ⚠ Flags Start Doing Nothing 

In phase 2 (e.g. TypeScript 5.5), deprecated flags can still be *specified* in tsconfig / commandline settings, but no longer have any effect. This allows users to have `tsconfig.json` files that work on both sides of the deprecation cycle (as you might have in a monorepo with a "shared" tsconfig settings file).

#### Phase 3: ❌ Flags Removed

[...]

PR #58313, opened 2024-04-25
5.5 deprecations are no-ops, not errors

RyanCavanaugh:
In #51000 we said

|Release / Phase|Description|
|----|------|
|TypeScript 5.0 - Phase 1|❔ Deprecation warnings introduced|
|TypeScript 5.5 - Phase 2|⚠ Flags start doing nothing|
|TypeScript 6.0 - Phase 3|❌ Flags removed entirely|

So the correct behavior under e.g.
```json
  "ignoreDeprecations": "5.0",
  "keyofStringsOnly": true
```
is to not error, but instead to pretend we didn't see `keyofStringsOnly`


src/compiler/program.ts
```diff
@@ -4593,10 +4594,10 @@ export function createProgram(rootNamesOrOptions: readonly string[] | CreateProg
                 }
                 else {
```
jakebailey, line 4595: Would probably be even better to add new diags in the middle here that says "is deprecated and has stopped functioning", since it'll be in future tense even when we're on like TS 5.6.

jakebailey:
Sorry about this; I think I made this stricter than you intended when I refactored things a bit back in 5.0.

tests/baselines/reference/alwaysStrictNoImplicitUseStrict.errors.txt
```diff
@@ -1,8 +1,8 @@
-error TS5102: Option 'noImplicitUseStrict' has been removed. Please remove it from your configuration.
+error TS5101: Option 'noImplicitUseStrict' is deprecated and will stop functioning in TypeScript 5.5. Specify compilerOption '"ignoreDeprecations": "5.0"' to silence this error.
```
DanielRosenwasser, line 1: Like Jake said,

```suggestion
error TS5101: Option 'noImplicitUseStrict' is deprecated and has stopped functioning in TypeScript 5.5. Specify compilerOption '"ignoreDeprecations": "5.0"' to silence this error.
```

would be better. I'll approve, but we should probably update this.

DanielRosenwasser (approved):

DanielRosenwasser:
Going to merge for the beta

PR #58318, opened 2024-04-25
Improve diagnostics deduplication 2

### LINKED ISSUE #58207: Duplicate diagnostics

gabritto:
Fixes #58207.
Follow-up to #58220.

This applies deduplication for the two scenarios I'm aware of where we can produce different diagnostics (with different message and code) that report the same problem at the same location, depending on the current state of the type checker:
(1) not found diagnostics that may or may not have suggestions, and (2) type comparison diagnostics that may or may not suppress the top part of a diagnostic chain depending on whether we have a more useful elaboration or not.

Case (1) depends on the state of the checker because we have a constant limit on how many times we try to look for suggestion, for performance reasons. Case (2) depends on the state of the checker because we don't provide a more detailed elaboration when we already have a cached result for the type comparison.

The general idea here is to make these different diagnostics that refer to the same problem point to the same canonical diagnostic, and this canonical diagnostic is used for comparison and deciding equivalence.
e.g. in case (1), we can produce both "Cannot find name 'x'." and "Cannot find name 'x'. Did you mean 'y'?", so now the diagnostic with message "Cannot find name 'x'. Did you mean 'y'?" will have a canonical diagnostic of "Cannot find name 'x'.", and will be considered equivalent to "Cannot find name 'x'." (but the diagnostic with the suggestion will be preferred).


gabritto:
@typescript-bot test it

gabritto:

andrewbranch (approved):
Makes sense to me, though I would wait for another review from someone who knows more about `captureErrorCalculationState`/`resetErrorInfo`

tests/cases/compiler/duplicateErrorNameNotFound.ts
```diff
@@ -0,0 +1,7 @@
+// @strict: true
+
+type RoomInterfae = {};
+
+export type {
+    RoomInterface
+}
```
andrewbranch, line 7: Did these tests actually produce duplicates, or were they just useful in getting near the code path for debugging?

tests/baselines/reference/mappedTypeRecursiveInference.types
```diff
@@ -4,7 +4,7 @@
 Assignability cache: 5,000
 Type Count: 10,000
 Instantiation count: 250,000
-Symbol count: 100,000
+Symbol count: 250,000
```
weswigham, line 7: ...That's a big bump. Is this just caused by the extra `reportRelationError` call expanding the recursive type way more? Still feels like a lot more symbols.

src/compiler/utilities.ts
```diff
@@ -2171,6 +2172,7 @@ export function createFileDiagnosticFromMessageChain(file: SourceFile, start: nu
         category: messageChain.category,
         messageText: messageChain.next ? messageChain : messageChain.messageText,
         relatedInformation,
+        canonicalHead: messageChain.canonicalHead,
```
sheetalkamat, line 2175: Does this need to be serialized in buildInfo when serializing diagnostic. Will we need this to deduplicate across multiple files?

tests/cases/compiler/duplicateErrorNameNotFound.ts
gabritto: They produced duplicates, yes.

tests/baselines/reference/mappedTypeRecursiveInference.types
weswigham: This isn't necessarily something that aughta get fixed in this PR, but understanding where _so many symbols_ popped in from could be useful. It might indicate that we're not caching something we aughta be.

weswigham (approved):

weswigham:

Juuuust in case

tests/baselines/reference/mappedTypeRecursiveInference.types
gabritto: > Is this just caused by the extra `reportRelationError` call expanding the recursive type way more?

I believe so. Since this test has an infinitely recurring type, it didn't seem like a great reason for concern to me, but I could be wrong.

src/compiler/utilities.ts
gabritto: > Will we need this to deduplicate across multiple files?

I don't think so, this should only be relevant for deduplicating diagnostics of a single file.

> Does this need to be serialized in buildInfo when serializing diagnostic?
What's the criterion for needing to serialize this in `buildInfo`?

weswigham:
Minus 2% time in vscode? Neat. They must have had quite a few duplicates.

jakebailey:
That result is right on the cusp of significance, and its before value is +/- 2%, so it's possible that it's not a real result, unfortunately. We're still p-hacking ourselves...

gabritto:
> That result is right on the cusp of significance, and its before value is +/- 2%, so it's possible that it's not a real result, unfortunately. We're still p-hacking ourselves...

Yep, I'd suspect it's just noise. I don't expect anything to get faster, and I think nothing changed on extended tests, so doesn't seem like vscode has duplicate diagnostics. On the upside, nothing got significantly worse either :)

weswigham:
> We're still p-hacking ourselves...

Should we un-fix `n` and instead fix `p`?

jakebailey:
Running the vscode benchmark already takes half an hour, I don't see us being able to change `n` that much.

src/compiler/utilities.ts
sheetalkamat: you would need this info to deduplicate with errors from other files or errors emitted for same file in other phase (which obviously isn't true - like say same error is reported by declaration emit and semantic diagnostic just giving hypothetical example here)

sheetalkamat: Given this doesnt get used beyond file boundaries this should be ok to be not serialized. 
That is checking files in any order will result in same error in that file (which we were anyways assuming till now) this is good.

sheetalkamat (approved):

PR #58320, opened 2024-04-25
Correct Regular Expressions Behavior Related to Annex B

graphemecluster:
Follow-up of #58295
This fixes issues like the `'}' expected` cases given at https://github.com/microsoft/TypeScript/issues/58275#issuecomment-2068174045.


src/compiler/scanner.ts
```diff
@@ -3390,7 +3400,7 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
                                 error(Diagnostics.Unicode_property_value_expressions_are_only_available_when_the_Unicode_u_flag_or_the_Unicode_Sets_v_flag_is_set, start, pos - start);
                             }
                         }
-                        else if (unicodeMode) {
+                        else if (!annexB) {
```
graphemecluster, line 3403: In Annex B, braces after `p` actually should not be parsed at all, but it does provide helpful errors like https://github.com/microsoft/TypeScript/issues/58275#issuecomment-2068174123

graphemecluster: Not really outdated

tests/baselines/reference/regularExpressionAnnexB.errors.txt
```diff
@@ -0,0 +1,272 @@
...
+    const regexes: RegExp[] = [
+      /\q\u\i\c\k\_\f\o\x\-\j\u\m\p\s/,
+           
+!!! error TS1125: Hexadecimal digit expected.
+               ~~
+!!! error TS1510: '\k' must be followed by a capturing group name enclosed in angle brackets.
+                         
+!!! error TS1125: Hexadecimal digit expected.
+                               
+!!! error TS1125: Hexadecimal digit expected.
```
graphemecluster, line 88: These are all valid in Annex B. Even `/\u{1/` is. In Annex B essentially all weird things are valid, you know. Currently (after #58295) `/[\1]/` is valid but `/[\8]/` isn’t. I don’t think it’s ideal.

graphemecluster:
Again, FWIW this might be of interest to @rbuckton and @jakebailey.

src/compiler/scanner.ts
jakebailey: The outdated label is just because the comment is on an old revision of the PR and GitHub can't figure out where the comment goes after.

jakebailey:
This will probably need to be rebased.

rbuckton:
@jakebailey is correct. #58339 also made some changes to this code, so a rebase or merge from `main` is necessary to resolve conflicts.

src/compiler/scanner.ts
```diff
@@ -2684,26 +2688,32 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
...
                                 }
                             }
-                            if (max && Number.parseInt(min) > Number.parseInt(max)) {
+                            else if (max && Number.parseInt(min) > Number.parseInt(max) && (!annexB || text.charCodeAt(pos) === CharacterCodes.closeBrace)) {
                                 error(Diagnostics.Numbers_out_of_order_in_quantifier, digitsStart, pos - digitsStart);
                             }
                         }
                         else if (!min) {
-                            if (unicodeMode) {
+                            if (!annexB) {
```
rbuckton, line 2701: Though it may be redundant, I think it might be better to still indicate `unicodeMode` here so that someone editing this code in the future doesn't mistakenly think this only applies to non-Annex B code. It may be better to use `unicodeMode || !annexB` and remove the `if (unicodeMode) { annexB = false; }` at the top of `scanRegularExpressionWorker`.

The same would go for other uses of `annexB` here.

src/compiler/scanner.ts
```diff
@@ -2801,7 +2811,10 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
                         scanGroupName(/*isReference*/ true);
                         scanExpectedChar(CharacterCodes.greaterThan);
                     }
-                    else if (unicodeMode) {
+                    else {
+                        // This is actually allowed in Annex B if there are no named capturing groups in the regex,
```
rbuckton, line 2815: Could we keep track of whether we encountered a `(?<` during `reScanSlashToken` and add an entry to the `RegularExpressionFlags` enum? The spec passes `NamedCaptureGroups` as a production parameter just as it does for `UnicodeMode` and `UnicodeSetsMode`, but only ever passes it as `~NamedCaptureGroups` in [Annex B](https://tc39.es/ecma262/#sec-parsepattern-annexb).

rbuckton (changes requested):

rbuckton:
@graphemecluster do you have the time to make changes or respond to comments, or would you like me to take over the final changes to get this merged?

src/compiler/scanner.ts
graphemecluster: Oh, didn’t think of this clever but dirty solution.
I can implement this but… the point is still about how much Annex B things we are going to respect. If we allow `\k` then should we allow `\u` and `\x` (and also `\8` and `\9` inside character classes) too? (https://github.com/microsoft/TypeScript/pull/58320#discussion_r1581764104)

src/compiler/scanner.ts
graphemecluster: Are you sure you are really fine with a dozen of occurrences of `unicodeMode || !annexB`?

src/compiler/scanner.ts
graphemecluster: Agreed. I could’ve removed the review and re-comment but finally chose not to.

src/compiler/scanner.ts
rbuckton: I think it's better to error on `\u`, `\x`, `\8`, `\9` because in those cases you are more likely to have actually meant something different. Writing `\k` when there are no named capture groups is far less ambiguous.

src/compiler/scanner.ts
rbuckton: No, but 
```ts
const anyUnicodeModeOrNonAnnexB = unicode Mode || !annexB;
```
would work.

src/compiler/scanner.ts
graphemecluster: If then I would personally lean towards linting also `\1`~`\7` inside character classes to prevent [this kind of mistakes](https://github.com/microsoft/TypeScript/issues/58275#issuecomment-2068174007). Or perhaps the opposite, only outside character classes. I don’t know. I understand that you don’t want the syntax checking to be too breaky, but someone must find this useful.

rbuckton: Every decimal escape in a character class I've ever seen has been a bug, so it makes sense to error for that case.

graphemecluster: Ah, but this would revert https://github.com/microsoft/TypeScript/pull/58295/files#diff-e19070e6b0ecc4de5f7d4bc6b42a9a71c8f23ce8ce155f7bd7cd48763170e506R1528-R1530. See if this is fine for you.

graphemecluster:
Wait, I will resolve the errors

src/compiler/scanner.ts
```diff
@@ -2887,7 +2903,7 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
                         return "\\";
                     }
                     pos--;
-                    return scanEscapeSequence(/*shouldEmitInvalidEscapeError*/ unicodeMode, /*isRegularExpression*/ annexB ? "annex-b" : true);
+                    return scanEscapeSequence(/*shouldEmitInvalidEscapeError*/ unicodeMode, /*isRegularExpression*/ anyUnicodeModeOrNonAnnexB || "annex-b");
```
rbuckton, line 2906: Since `"annexB"` is no longer used in `scanEscapeSequence`, this can just be `true`:
```suggestion
                    return scanEscapeSequence(/*shouldEmitInvalidEscapeError*/ unicodeMode, /*isRegularExpression*/ true);
```

rbuckton (changes requested):
A few more minor comments to address and this will be ready to merge.

src/compiler/scanner.ts
```diff
@@ -2469,6 +2468,15 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
                 else if (ch === CharacterCodes.closeBracket) {
                     inCharacterClass = false;
                 }
+                else if (
+                    ch === CharacterCodes.openParen
+                    && charCodeUnchecked(p + 1) === CharacterCodes.question
+                    && charCodeUnchecked(p + 2) === CharacterCodes.lessThan
+                    && charCodeUnchecked(p + 3) !== CharacterCodes.equals
+                    && charCodeUnchecked(p + 3) !== CharacterCodes.exclamation
```
rbuckton, line 2476: I missed this in my last review. These need to be range checked:
```suggestion
                    && charCodeChecked(p + 1) === CharacterCodes.question
                    && charCodeChecked(p + 2) === CharacterCodes.lessThan
                    && charCodeChecked(p + 3) !== CharacterCodes.equals
                    && charCodeChecked(p + 3) !== CharacterCodes.exclamation
```

rbuckton (changes requested):

src/compiler/scanner.ts
```diff
@@ -2684,26 +2694,32 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
...
-                                    if (unicodeMode) {
-                                        error(Diagnostics.Unexpected_0_Did_you_mean_to_escape_it_with_backslash, start, 1, String.fromCharCode(ch));
-                                    }
+                                    error(Diagnostics.Unexpected_0_Did_you_mean_to_escape_it_with_backslash, start, 1, String.fromCharCode(ch));
                                     isPreviousTermQuantifiable = true;
                                     break;
                                 }
                             }
-                            if (max && Number.parseInt(min) > Number.parseInt(max)) {
+                            else if (max && Number.parseInt(min) > Number.parseInt(max) && (anyUnicodeModeOrNonAnnexB || text.charCodeAt(pos) === CharacterCodes.closeBrace)) {
```
rbuckton, line 2702:
```suggestion
                            else if (max && Number.parseInt(min) > Number.parseInt(max) && (anyUnicodeModeOrNonAnnexB || charCodeChecked(pos) === CharacterCodes.closeBrace)) {
```

rbuckton (changes requested):
One more I missed.

src/compiler/scanner.ts
```diff
@@ -1556,9 +1556,7 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
                 tokenFlags |= TokenFlags.ContainsInvalidEscape;
                 if (isRegularExpression || shouldEmitInvalidEscapeError) {
                     const code = parseInt(text.substring(start + 1, pos), 8);
-                    if (isRegularExpression !== "annex-b") {
```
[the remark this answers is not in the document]
graphemecluster: I thought that too, but there is a `isRegularExpression === true` at the bottom

tests/baselines/reference/regularExpressionScanning(target=es5).errors.txt
```diff
@@ -261,6 +265,16 @@ regularExpressionScanning.ts(47,101): error TS1501: This regular expression flag
...
+    	                      ~~
+!!! error TS1487: Octal escape sequences are not allowed. Use the syntax '\x01'.
+    	                        ~~~
+!!! error TS1487: Octal escape sequences are not allowed. Use the syntax '\x01'.
+    	                           ~~~~
+!!! error TS1487: Octal escape sequences are not allowed. Use the syntax '\x53'.
+    	                                   ~~~
+!!! error TS1487: Octal escape sequences are not allowed. Use the syntax '\x03'.
+    	                                         ~~~~
+!!! error TS1487: Octal escape sequences are not allowed. Use the syntax '\x05'.
```
graphemecluster, line 277: I would like to confirm again why decimal escapes (that don’t refer to existent capturing groups) are worth not being errored? Are octal escapes that common?

src/compiler/scanner.ts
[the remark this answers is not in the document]
graphemecluster: How about swapping it for `"anyUnicodeModeOrNonAnnexB"`?

tests/baselines/reference/regularExpressionScanning(target=es5).errors.txt
rbuckton: We _should_ report an error for octal escapes. They are permitted in Annex B, but every use I've seen in the wild has actually been a bug, i.e., people mistakenly thinking `/(['"])[^\1]?\1/` matches a quoted string except `[^\1]` is not "every character except the one captured by the backreference `\1`", it's actually "every character except `\x01`".

src/compiler/scanner.ts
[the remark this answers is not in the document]
graphemecluster: This probably is worth being refactored, since `shouldEmitInvalidEscapeError` does not explain itself but is taken over to mean `anyUnicodeMode` for regular expressions. But this will be yet another PR I guess.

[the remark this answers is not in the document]
graphemecluster: Do you use an enum for such a case?

tests/baselines/reference/regularExpressionScanning(target=es5).errors.txt
rbuckton: > I mean that outside character classes. 

We should make it an error again in or out of character classes.

> If we are going to emit errors for all octal escapes, [this change](https://github.com/microsoft/TypeScript/pull/58295/files#diff-e19070e6b0ecc4de5f7d4bc6b42a9a71c8f23ce8ce155f7bd7cd48763170e506R3437-R3439), too, should not have been made.

Correct. That PR swung towards being far more permissive while we were discussing a happy medium we should land on. If it not handled in this PR, I was planning to follow up with that once this PR was merged.

rbuckton:
Can you resolve the conflict with scanner.ts from `main` now that I've merged #58289?

tests/baselines/reference/regularExpressionScanning(target=es2015).errors.txt
```diff
@@ -250,10 +258,28 @@ regularExpressionScanning.ts(47,101): error TS1501: This regular expression flag
     	// Capturing groups
     	/\0/,
     	/\1/,
+    	  ~
+!!! error TS1534: This backreference is invalid because the containing regular expression contains no capturing groups.
```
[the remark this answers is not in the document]
rbuckton: Or "This backreference points to an invalid capture group.", or something to that effect.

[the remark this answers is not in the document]
graphemecluster: Let’s match this with TS1533.

tests/baselines/reference/regularExpressionScanning(target=es2015).errors.txt
```diff
@@ -250,10 +258,28 @@ regularExpressionScanning.ts(47,101): error TS1501: This regular expression flag
...
     	/\2/,
+    	  ~
+!!! error TS1534: This backreference is invalid because the containing regular expression contains no capturing groups.
     	/(hi)\1/,
     	/(hi) (hello) \2/,
     	/\2()(\12)(foo)\1\0[\0\1\01\123](\3\03)\5\005/,
+    	       ~~
+!!! error TS1533: This backreference refers to a group that does not exist. There are only 4 capturing groups in this regular expression.
+    	                      ~~
+!!! error TS1487: Octal escape sequences are not allowed. Use the syntax '\x01'.
```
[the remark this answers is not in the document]
graphemecluster: Have to do the refactor on `ScanEscapeSequence` (https://github.com/microsoft/TypeScript/pull/58320#discussion_r1612257742) to achieve this (or there will be an unrelated parameter on `ScanEscapeSequence` for string literals).

src/compiler/scanner.ts
```diff
@@ -997,6 +997,19 @@ export function isIdentifierText(name: string, languageVersion: ScriptTarget | u
     return true;
 }
 
+const enum EscapeSequenceScanningFlags {
```
graphemecluster, line 1000: Actually, there are only 8 valid states:
```
String,
String | ReportErrors,
RegularExpression,
RegularExpression | AnnexB,
RegularExpression | AnyUnicodeMode,
RegularExpression | AtomEscape,
RegularExpression | AnnexB | AtomEscape,
RegularExpression | AnyUnicodeMode | AtomEscape,
```
which is supposed to fit in 3 booleans.

rbuckton: I think this works fine.

rbuckton (approved):

rbuckton:

I think this is the first time I've ever actually seen someone using an octal escape for its intended purpose in a character class, but it's still probably better if Puppeteer changed its two cases to use `\x20` instead. @DanielRosenwasser, any concerns with merging this one?

rbuckton:

- `puppeteer` - These are errors on what would otherwise be a correct usage of octal escapes, but should probably be rewritten to hex escapes.
- `dotansimha/graphql-code-generator` - These errors correctly illustrate a bug.
- `microsoft/vscode` - These are errors on what would otherwise be correct usage of octal escapes, but should probably be rewritten to hex escapes.

rbuckton (approved):

PR #58321, opened 2024-04-25
Experiment: Remove lower priority from contextual signature instantiation return type inference

andrewbranch:


andrewbranch:
@typescript-bot test top200

andrewbranch:
@typescript-bot pack this

andrewbranch:
@typescript-bot test top200
@typescript-bot pack this

tests/baselines/reference/mappedTypeInferenceFromApparentType.errors.txt
```diff
@@ -20,7 +18,5 @@ mappedTypeInferenceFromApparentType.ts(10,1): error TS2322: Type 'foo' is not as
     ~
 !!! error TS2322: Type 'foo' is not assignable to type 'bar'.
 !!! error TS2322:   Types of parameters 'target' and 'source' are incompatible.
-!!! error TS2322:     Type '{ [K in keyof U]: Obj[K]; }' is not assignable to type '{ [K in keyof U]: U[K]; }'.
-!!! error TS2322:       Type 'Obj[K]' is not assignable to type 'U[K]'.
-!!! error TS2322:         Type 'Obj' is not assignable to type 'U'.
-!!! error TS2322:           'U' could be instantiated with an arbitrary type which could be unrelated to 'Obj'.
\ No newline at end of file
+!!! error TS2322:     Type '{ [K in keyof U]: Obj[K]; }' is not assignable to type '{ [x: string]: number; }'.
```
gabritto, line 21: I wrote this test to exercise a very specific inference code path in the PR that introduced it, so it's not grounded in a real world scenario, but I don't think the new error is worse? Maybe the connection between the error and the actual code is more surprising now just because we end up inferring `T = { [x: string]: number }` instead of `T = U`.

Maybe that's a separate problem with signature comparison: it's not immediately clear to users that we've done inference in order to compare two signatures, and it's not immediately clear what the result of that inference was.

Issue #57585, opened 2024-02-28
TypeScript Language Server False Positive Error Reporting


DScheglov:
Does this issue occur when all extensions are disabled?: Yes

- VS Code Version: 1.87.0
- OS Version: Ubuntu 22.04

Steps to Reproduce:

1. Clone the project `git clone https://github.com/DScheglov/res-test.git`
2. Open the VS Code: `code res-test`
3. Install dependencies (in terminal): `npm i`
4. Open file `./src/index.ts`

The following errors are shown:
![image](https://github.com/microsoft/vscode/assets/9888941/aca72243-17aa-4cff-956d-e029bc5d4c38)

These errors are false positive:
1. project is configured to use the same typescript compiler as a project. See `.vscode/settings.json`
2. project is compiled successfully via terminal: `npx tsc` (no errors)
3. errors are gone after TS Language Server is restared (`Ctrl + Shift + P` , TypeScript: Restart TS Server)
4. errors go back after any touch of the file (just add a space).

There is not such problem in WebStorm
![image](https://github.com/microsoft/vscode/assets/9888941/f679d7dc-a661-4937-b2a6-235c419c8056)
(Sorry, I have to check)

Errors are not displayed in the TS Playground as well:
[...]

PR #58323, opened 2024-04-26
[experiment] force evaluation of generator function type when getTypeOfNode-ing nodes inside it

jakebailey:
For #57585, but almost certainly a bad idea.


jakebailey:
@typescript-bot pack this

Issue #53287, opened 2023-03-16
TypeError: Cannot read properties of undefined (reading 'get') at getTypeAliasInstantiation


Itrulia:
# Bug Report

### 🔎 Search Terms

getTypeAliasInstantiation,  (reading 'get'), TypeError: Cannot read properties of undefined, createMarkerType

### 🕗 Version & Regression Information

- This is a crash
- This changed between versions ~4.6.0 and ~4.7.0 and is still present in 4.9.5

We have only now figured out something went wrong as we used babel to transpile and [fork-ts-checker-webpack-plugin](https://www.npmjs.com/package/fork-ts-checker-webpack-plugin) to type check. This caused the type checking to silently fail, which has caused a few bugs lately.

### ⏯ Playground Link

I don't have a Playground Link as I have no clue what causes it. But the repository is public:

https://github.com/wepublish/wepublish/tree/8e4198788c2d60aa5f03869a6fdf9bda2566b45f
The command needed to reproduce it: `npm i && npx tsc -p ./apps/editor/tsconfig.app.json`

If typescript version is set to `~4.6.0` everything works. If typescript is set to 4.7 or above, it fails

### 💻 Code

```ts
// We can quickly address your report if:
//  - The code sample is short. Nearly all TypeScript bugs can be demonstrated in 20-30 lines of code!
//  - It doesn't use external libraries. These are often issues with the type definitions rather than TypeScript bugs.
//  - The incorrectness of the behavior is readily apparent from reading the sample.
// Reports are slower to investigate if:
//  - We have to pare too much extraneous code.
//  - We have to clone a large repo and validate that the problem isn't elsewhere.
//  - The sample is confusing or doesn't clearly demonstrate what's wrong.
```

I haven't been able to produce a minimal reproduction.

### 🙁 Actual behavior

TSC crashes with an error thrown:

```
❯ npx tsc -p ./apps/editor/tsconfig.app.json
/Users/itrulia/Documents/wepublish/node_modules/typescript/lib/tsc.js:98439
                throw e;
                ^

TypeError: Cannot read properties of undefined (reading 'get')
    at getTypeAliasInstantiation (/Users/itrulia/Documents/wepublish/node_modules/typescript/lib/tsc.js:50531:54)
    at createMarkerType (/Users/itrulia/Documents/wepublish/node_modules/typescript/lib/tsc.js:56512:17)
[...]

PR #58326, opened 2024-04-26
fix53287 `mergeSymbol` checks if the resolved target can merge with the source

iisaduan:
Fixes https://github.com/microsoft/TypeScript/issues/53287

While investigating the original issue, I found that re-exporting an alias would circumvent any checks to determine whether or not the source and the target are allowed to merge at all. 

This adds a check in `mergeSymbol` to check if the resolved target can merge with the source, and if they cannot merge, then we issue the same errors as if there was no reexporting.

Got lots of help from @sandersn looking over the new errors 🙂


iisaduan:
@typescript-bot test it

iisaduan:
@typescript-bot pack this

iisaduan:
Testing results analysis: The extra errors correctly identify wrong code.
- ✅@types/karma-chai-sinon: Sinon already contains `export as namespace sinon`, so `declare global { var Sinon }` is correctly identified as a duplicate identifier [on line 9](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/karma-chai-sinon/index.d.ts)

- ✅ @types/bandagedb__dbapi: similar-- lodash has `export as namespace _`, and `declare global { const _: typeof _; }` [on line 9](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/64828b61751341bfdb2bd5ba88ef61c1f1f77cb8/types/bandagedbd__bdapi/index.d.ts#L7C1-L10C1) is redundant

- ✅ @types/regenerator-runtime: `export as namespace regeneratorRuntime` makes `declare global { var regeneratorRuntime: typeof import("."); }` redundant [on lines 2-4](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/64828b61751341bfdb2bd5ba88ef61c1f1f77cb8/types/regenerator-runtime/index.d.ts#L2C1-L4C2)


- ✅ [actualbudget/actual](https://github.com/actualbudget/actual): this is similar to the issue in the original bug. This package imports both @rschedule/core and @rschedule/json-tools.
`json-tools` attempts to extend `core`. In `json-tools`, the declarations of `namespace Schedule`[<sup>14</sup>](https://github.com/jorroll/rschedule/blob/3f125976722137e7a9c621efde3c2f8f2c479ff3/packages/json-tools/src/Schedule/index.ts#L14) and `class Schedule<Data = any>`[<sup>27</sup>](https://github.com/jorroll/rschedule/blob/3f125976722137e7a9c621efde3c2f8f2c479ff3/packages/json-tools/src/Schedule/index.ts#L27) clash with the `export class Schedule<Data = any> extends ScheduleBase<Data>` [ <sup>23</sup>](https://github.com/jorroll/rschedule/blob/3f125976722137e7a9c621efde3c2f8f2c479ff3/packages/core/src/generators/schedule.ts#L23) in `core`

sandersn (approved):
The main way I can imagine this going wrong is a merge of an alias that should have happened that is now blocked. The fact that there weren't any surprises in the top400 tests is encouraging. But I'd like the other reviewers to see if they can come up with anything.

As for the likely breaks, I think we've observed those already: bad merges that no longer merge, so that half of the intended merge is missing now. Given the low frequency of breaks I think it's not a problem.

andrewbranch (approved):
I think this is the right approach. If we were worried about any of the specific merges that became errors, I think the thing to do would be to adjust the merge exclusion flags to allow the merge to happen (whether mediated by an alias or not), then issue errors during type checking for cases that look wrong. For example, we might decide that

```ts
declare global {
  const React: typeof import("./module");
}
export {};

// module.d.ts
export = React;
export as namespace React;
```

is fine because the global variable declaration has the same type as the value side of the UMD module declaration, which matches our rules for when two ambient variable declarations can merge. However, I don't think we have evidence that changing the rules is necessary right now, so I think we should go with this and adjust if we get compelling feedback.

PR #58333, opened 2024-04-26
Dont get declaration diagnostics for file from referenced project

### LINKED ISSUE #58328: isolatedDeclarations wrongly enabled in VSCode
### 🔎 Search Terms

isolatedDeclaration

5.5.0-beta

### ⏯ Playground Link

https://github.com/Jack-Works/reproduce/tree/main/typescript-5.5-isolatedDeclarations-vscode

### 💻 Code

```sh
git clone https://github.com/Jack-Works/reproduce/ --filter=blob:none
cd typescript-5.5-isolatedDeclarations-vscode
code . # start vscode at the current directory
```

Open `b/index.ts` in the editor. (you must not open `a/index.ts` otherwise the error won't appear)

### 🙁 Actual behavior

img

But A does not have `isolatedDeclarations` enabled.

图片

If you open `a/index.ts`, the error is gone, and if you close it, the error appears again.

### 🙂 Expected behavior

Works

_No response_

sheetalkamat:
Fixes  #58328

### DIFF
--- a/src/compiler/program.ts
+++ b/src/compiler/program.ts
@@ -2882,6 +2882,7 @@ export function createProgram(rootNamesOrOptions: readonly string[] | CreateProg
     }
 
     function getDeclarationDiagnostics(sourceFile?: SourceFile, cancellationToken?: CancellationToken): readonly DiagnosticWithLocation[] {
+        if (sourceFile && isSourceOfProjectReferenceRedirect(sourceFile.fileName)) return emptyArray;
         const options = program.getCompilerOptions();
         // collect diagnostics from the program only once if either no source file was specified or out/outFile is set (bundled emit)
         if (!sourceFile || options.outFile) {
--- a/src/compiler/program.ts
+++ b/src/compiler/program.ts
@@ -2882,7 +2882,6 @@ export function createProgram(rootNamesOrOptions: readonly string[] | CreateProg
     }
 
     function getDeclarationDiagnostics(sourceFile?: SourceFile, cancellationToken?: CancellationToken): readonly DiagnosticWithLocation[] {
-        if (sourceFile && isSourceOfProjectReferenceRedirect(sourceFile.fileName)) return emptyArray;
         const options = program.getCompilerOptions();
         // collect diagnostics from the program only once if either no source file was specified or out/outFile is set (bundled emit)
         if (!sourceFile || options.outFile) {
--- a/src/compiler/transformers/declarations.ts
+++ b/src/compiler/transformers/declarations.ts
@@ -215,12 +215,13 @@ import {
 /** @internal */
 export function getDeclarationDiagnostics(host: EmitHost, resolver: EmitResolver, file: SourceFile | undefined): DiagnosticWithLocation[] | undefined {
     const compilerOptions = host.getCompilerOptions();
+    const files = filter(getSourceFilesToEmit(host, file), isSourceFileNotJson);
     const result = transformNodes(
         resolver,
         host,
         factory,
         compilerOptions,
-        filter(getSourceFilesToEmit(host, file), isSourceFileNotJson),
+        file ? contains(files, file) ? [file] : emptyArray : files,
         [transformDeclarations],
         /*allowDtsFiles*/ false,
     );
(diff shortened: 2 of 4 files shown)


src/compiler/program.ts
```diff
@@ -2882,6 +2882,7 @@ export function createProgram(rootNamesOrOptions: readonly string[] | CreateProg
     }
 
     function getDeclarationDiagnostics(sourceFile?: SourceFile, cancellationToken?: CancellationToken): readonly DiagnosticWithLocation[] {
+        if (sourceFile && isSourceOfProjectReferenceRedirect(sourceFile.fileName)) return emptyArray;
```
sheetalkamat, line 2885: Actually i think getDeclarationDiagnostics need to filter files with "sourceFileMayBeEmitted"

weswigham: `sourceFileMayBeEmitted` makes sense here, will need to pass in an appropriate host, though. It also probably makes sense to do the filtering down in the `getDeclarationDiagnostics` core implementation in `declarations.ts`.

weswigham (approved):

src/compiler/transformers/declarations.ts
```diff
@@ -214,7 +215,16 @@ import {
 /** @internal */
 export function getDeclarationDiagnostics(host: EmitHost, resolver: EmitResolver, file: SourceFile | undefined): DiagnosticWithLocation[] | undefined {
     const compilerOptions = host.getCompilerOptions();
-    const result = transformNodes(resolver, host, factory, compilerOptions, file ? [file] : filter(host.getSourceFiles(), isSourceFileNotJson), [transformDeclarations], /*allowDtsFiles*/ false);
+    const files = filter(getSourceFilesToEmit(host, file), isSourceFileNotJson);
```
sheetalkamat, line 218: @weswigham i am not sure if i should have just file here in the transform and it is ok or do you need to have all the files and then filter results from only that file ? (this is because of --out when all files need to be emitted together in a bundle) we have a test that failed without this change `getDeclarations` in fourslash

jakebailey:
Just verifying the playground works with this

@typescript-bot pack this

src/compiler/transformers/declarations.ts
```diff
@@ -214,7 +215,16 @@ import {
...
 export function getDeclarationDiagnostics(host: EmitHost, resolver: EmitResolver, file: SourceFile | undefined): DiagnosticWithLocation[] | undefined {
     const compilerOptions = host.getCompilerOptions();
-    const result = transformNodes(resolver, host, factory, compilerOptions, file ? [file] : filter(host.getSourceFiles(), isSourceFileNotJson), [transformDeclarations], /*allowDtsFiles*/ false);
+    const files = filter(getSourceFilesToEmit(host, file), isSourceFileNotJson);
+    const result = transformNodes(
+        resolver,
+        host,
+        factory,
+        compilerOptions,
+        file ? contains(files, file) ? [file] : emptyArray : files,
```
jakebailey, line 224: Readability nit, but I have a real hard time trying to read this one; is there another way to put this?

```ts
let files = filter(getSourceFilesToEmit(host, file), isSourceFileNotJson);
if (file) {
   files = contains(files, file) ? [file] : emptyArray;
}
```

Maybe? If not, no worries, I just have problems reading ternaries.

src/compiler/transformers/declarations.ts
weswigham: `getDeclarationDiagnostics` already differs slightly from normal declaration emit's transform call when `--out` is supplied, so I wouldn't sweat it - the little differences historically haven't mattered, since declaration emit-added errors are all visibility related (plus isolatedDeclaration related now, which are all very context-free), and generally don't depend on the input file set. Specifically, `emitFiles` will put all the files into a `Bundle` when `--out` is specified, but this does not. That difference has yet to matter in practice, since we don't have any bundle-specific errors (that are not instead switched based on compiler options), so I'm pretty sure just sending in the interesting file even when `--out` is set is probably OK.

weswigham (approved):

src/compiler/transformers/declarations.ts
sheetalkamat: Sorry its hard to read for you. I prefer `consts` over let for this. Do you think having it multi line will be more readable

andrewbranch (approved):

src/compiler/transformers/declarations.ts
jakebailey: It's fine as it is, no worries.

PR #58337, opened 2024-04-26
Fix ghost errors resulting from out-of-order type checking

### LINKED ISSUE #57429: Incorrect ts7022 error emitted by tsserver (not by tsc)
### 🔎 Search Terms

incorrect ts7022, wrong ts7022, tsserver only error ts7022

This is the behavior in every version I tried, and I reviewed the FAQ for entries about 'Type System Behavior'.

Summary: `tsserver` emits incorrect TS7022 error, meanwhile `tsc` and `TS Playground` behave as expected.

Steps to Reproduce (NOTE: the error is NOT reproducible in TS Playground):

1. Clone [https://github.com/cleric-sh/repro/tree/phantom-ts7022-error-in-vscode](https://github.com/cleric-sh/repro/tree/phantom-ts7022-error-in-vscode)
2. Ensure switched to branch `phantom-ts7022-error-in-vscode`
3. `npm i`
4. `npm run compile` --> observe no compiler errors
5. Open `index.ts` (in VSCode or another editor utilizing `tsserver`)
6. Error may or may not appear on line 26 on `value` (depending on whether analysis is on first pass or not)
7. Make a change to the file (e.g. add a space) to trigger tsserver to re-analyse the file.
8. Error should definitely be visible now.
9. Restart the TS Language Server and error disappears. Make a change, and error re-appears.

The code in this repro emits an error that is only visible in VSCode whose intellisense is server by `tsserver`.

To compare, the same error is not emitted by `tsc` for the same source and configuration.

I have attempted to raise this issue at:
- https://github.com/microsoft/vscode/issues/195075, and
- https://github.com/typescript-language-server/typescript-language-server/issues/766

In `typescript-language-server` I was informed that the error is actually produced in `tsserver` and so I should raise this with the `typescript` team, so here I am!

### ⏯ Playground Link

https://tsplay.dev/WkqM2N

### 💻 Code

```ts
function Builder<I>(def: I) {

  return def;
}

interface IThing {
  doThing: (args: { value: object }) => string
  doAnotherThing: () => void
}

Builder<IThing>({

  doThing(args: { value: object }) {
    /**
     * ts7022 is shown on 'value' below.
     * 
     * Only shown when func param 'arg' type is declared. Removing the declaration resolves the error.
     * Only shown in vscode. tsc compiles without errors.
     * Only shown on second pass, after a change has been made to the file. Initial type check on load shows no errors.
     * Only shown when noImplicitAny: true
     * 
     * In a more complex file, declaring this function
     * after 'doAnotherThing' prevents the error being
     * displayed.
     */
    const { value } = this.args
    return `${value}`
  },

  doAnotherThing() { },  

})
```

### 🙁 Actual behavior

Error is shown on Line 26: 
```
'value' implicitly has type 'any' because it does not have a type annotation and is referenced directly or indirectly in its own initializer.ts(7022)
```

### 🙂 Expected behavior

Behaviour in `tsserver` to match `tsc` (which does not emit the error).

Not sure exactly what *should* happen.... I would imagine one of:
- `this` resolves to `any`, since `ThisType` isn't used, and thus `value` should also be `any`.
- `this` resolve to `IThing`.

But the `ts7022` error shouldn't be displayed.

The error appears with all extensions disabled.

The error appears with all versions of `typescript` (I have checked all versions since `5.2.2`), including `5.4.0-beta`.

The error does NOT appear in TS Playground: https://tsplay.dev/WkqM2N

I've noted that the error is:

- Only shown when func param 'arg' type is declared. Removing the declaration resolves the error.
- Only shown in vscode. tsc compiles without errors.
- Only shown on second pass, after a change has been made to the file. Initial type check on load shows no errors.
- Only shown when noImplicitAny: true

Also, in a more complex file, declaring this function after doAnotherThing prevents the error being displayed.

There is generally a lot of strange behaviour in this scenario of using 'this' in a non-class context. The repro is non-sensical, but is the simplest reproduction of the same error I could find. 

At least this repro demonstrates:
- the error doesn't seem to make sense. e.g. why wouldn't `value` just be `any` since `this` isn't typed?
- the error is inconsistent:
  - disappearing if (for example) the returned object from `Builder` is assigned to a variable.
  - not emitted by TSC or visible in TS Playground

### LINKED ISSUE #57585: TypeScript Language Server False Positive Error Reporting

ahejlsberg:
This PR introduces logic that suppresses ghost errors resulting from "inverted" type checking situations where, for example, an API client asks for the type of a symbol in the middle of a section of code, causing type resolution to occur in a different order than when the code is checked from top to bottom in a regular compilation.

Some context on the fix in this PR. In this example

```ts
function builder(def: IThing) {
  return def;
}

interface IThing {
  doThing: (args: { value: object }) => string
  args: { value: number }
}

builder({
  doThing(args: { value: object }) {
    const { value } = this.args;  // Ghost circularity error on "value"
    return `${value}`
  },
  args: { value: 42 }
});
```

a circularity error is reported in the VS Code IDE, but no error is reported with the command-line compiler. The issue is that the IDE language service requests type information on identifiers in the code (likely for semantic classification, but doesn't really matter) _before_ it requests the full list of diagnostics for the code. Specifically, a request is initially made for `getTypeOfSymbol` for the symbol of the `value` identifier. This causes resolution of `this.args`, and to resolve the type of `this`, a request is made for the contextual type of the object literal passed as an argument in the `builder({...})` call. That in turn means we need to resolve the signature for the call, which in turn means resolving the type of the arguments, which in turn means obtaining the type of the `{...}` object literal argument, which further requires resolving the return type of `doThing`, which gets us back to a second `getTypeOfSymbol` call for `value`. And we have a circularity.

However, in the regular top-to-bottom sweep to produce diagnostics, we request resolving the signature of the `builder({...})` call first, which eventually gets us to a `getTypeOfSymbol` for `value`, which in turn requests the contextual type for the object literal. We then see that we're in the process of resolving the signature of the containing `builder({...})` call, and therefore we simply report that there is no contextual type. And thus no circularity.

With the fix in this PR, we temporarily reset the resolution stack during `getResolvedSignature` calls. In the IDE scenario this means that the first call to `getTypeOfSymbol` is no longer reflected on the resolution stack, so resolution can proceed in the same manner as the regular case. Following that, the outer `getTypeOfSymbol` call notices that a resolved type has already been recorded, so it simply proceeds with that. No circularity, and all is well.

Fixes #57429.
Fixes #57585.

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -11741,7 +11741,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 }
                 type = anyType;
             }
-            links.type = type;
+            links.type ??= type;
         }
         return links.type;
     }
@@ -11763,7 +11763,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 writeType = anyType;
             }
             // Absent an explicit setter type annotation we use the read type of the accessor.
-            links.writeType = writeType || getTypeOfAccessors(symbol);
+            links.writeType ??= writeType || getTypeOfAccessors(symbol);
         }
         return links.writeType;
     }
@@ -11849,7 +11849,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             // type symbol, call getDeclaredTypeOfSymbol.
             // This check is important because without it, a call to getTypeOfSymbol could end
             // up recursively calling getTypeOfAlias, causing a stack overflow.
-            links.type = exportSymbol?.declarations && isDuplicatedCommonJSExport(exportSymbol.declarations) && symbol.declarations!.length ? getFlowTypeFromCommonJSExport(exportSymbol)
+            links.type ??= exportSymbol?.declarations && isDuplicatedCommonJSExport(exportSymbol.declarations) && symbol.declarations!.length ? getFlowTypeFromCommonJSExport(exportSymbol)
                 : isDuplicatedCommonJSExport(symbol.declarations) ? autoType
                 : declaredType ? declaredType
                 : getSymbolFlags(targetSymbol) & SymbolFlags.Value ? getTypeOfSymbol(targetSymbol)
@@ -11857,7 +11857,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
 
             if (!popTypeResolution()) {
                 reportCircularityError(exportSymbol ?? symbol);
-                return links.type = errorType;
+                return links.type ??= errorType;
             }
         }
         return links.type;
@@ -12199,7 +12199,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             }
             if (!popTypeResolution()) {
                 error(type.symbol.valueDeclaration, Diagnostics._0_is_referenced_directly_or_indirectly_in_its_own_base_expression, symbolToString(type.symbol));
-                return type.resolvedBaseConstructorType = errorType;
+                return type.resolvedBaseConstructorType ??= errorType;
             }
             if (!(baseConstructorType.flags & TypeFlags.Any) && baseConstructorType !== nullWideningType && !isConstructorType(baseConstructorType)) {
                 const err = error(baseTypeNode.expression, Diagnostics.Type_0_is_not_a_constructor_function_type, typeToString(baseConstructorType));
@@ -12216,9 +12216,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                         addRelatedInfo(err, createDiagnosticForNode(baseConstructorType.symbol.declarations[0], Diagnostics.Did_you_mean_for_0_to_be_constrained_to_type_new_args_Colon_any_1, symbolToString(baseConstructorType.symbol), typeToString(ctorReturn)));
                     }
                 }
-                return type.resolvedBaseConstructorType = errorType;
+                return type.resolvedBaseConstructorType ??= errorType;
             }
-            type.resolvedBaseConstructorType = baseConstructorType;
+            type.resolvedBaseConstructorType ??= baseConstructorType;
         }
         return type.resolvedBaseConstructorType;
     }
@@ -12500,7 +12500,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     error(isNamedDeclaration(declaration) ? declaration.name || declaration : declaration, Diagnostics.Type_alias_0_circularly_references_itself, symbolToString(symbol));
                 }
             }
-            links.declaredType = type;
+            links.declaredType ??= type;
         }
         return links.declaredType;
     }
@@ -13814,7 +13814,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 error(currentNode, Diagnostics.Type_of_property_0_circularly_references_itself_in_mapped_type_1, symbolToString(symbol), typeToString(mappedType));
                 type = errorType;
             }
-            symbol.links.type = type;
+            symbol.links.type ??= type;
         }
         return symbol.links.type;
     }
@@ -14266,7 +14266,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     }
                     result = circularConstraintType;
                 }
-                t.immediateBaseConstraint = result || noConstraintType;
+                t.immediateBaseConstraint ??= result || noConstraintType;
             }
             return t.immediateBaseConstraint;
         }
@@ -15824,10 +15824,10 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 node.kind === SyntaxKind.ArrayType ? [getTypeFromTypeNode(node.elementType)] :
                 map(node.elements, getTypeFromTypeNode);
             if (popTypeResolution()) {
-                type.resolvedTypeArguments = type.mapper ? instantiateTypes(typeArguments, type.mapper) : typeArguments;
+                type.resolvedTypeArguments ??= type.mapper ? instantiateTypes(typeArguments, type.mapper) : typeArguments;
             }
             else {
-                type.resolvedTypeArguments = type.target.localTypeParameters?.map(() => errorType) || emptyArray;
+                type.resolvedTypeArguments ??= type.target.localTypeParameters?.map(() => errorType) || emptyArray;
                 error(
                     type.node || currentNode,
                     type.target.symbol ? Diagnostics.Type_arguments_for_0_circularly_reference_themselves : Diagnostics.Tuple_type_arguments_circularly_reference_themselves,
@@ -29080,7 +29080,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 return true;
             }
 
-            links.parameterInitializerContainsUndefined = containsUndefined;
+            links.parameterInitializerContainsUndefined ??= containsUndefined;
         }
 
         return links.parameterInitializerContainsUndefined;
@@ -35749,7 +35749,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             return cached;
         }
         const saveResolutionStart = resolutionStart;
-        if (saveResolutionStart === 0) {
+        if (!cached) {
             // If we haven't already done so, temporarily reset the resolution stack. This allows us to
             // handle "inverted" situations where, for example, an API client asks for the type of a symbol
             // containined in a function call argument whose contextual type depends on the symbol itself
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -35748,8 +35748,19 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         if (cached && cached !== resolvingSignature && !candidatesOutArray) {
             return cached;
         }
+        const saveResolutionStart = resolutionStart;
+        if (saveResolutionStart === 0) {
+            // If we haven't already done so, temporarily reset the resolution stack. This allows us to
+            // handle "inverted" situations where, for example, an API client asks for the type of a symbol
+            // containined in a function call argument whose contextual type depends on the symbol itself
+            // through resolution of the containing function call. By resetting the resolution stack we'll
+            // retry the symbol type resolution with the resolvingSignature marker in place to suppress
+            // the contextual type circularity.
+            resolutionStart = resolutionTargets.length;
+        }
         links.resolvedSignature = resolvingSignature;
         let result = resolveSignature(node, candidatesOutArray, checkMode || CheckMode.Normal);
+        resolutionStart = saveResolutionStart;
         // When CheckMode.SkipGenericFunctions is set we use resolvingSignature to indicate that call
         // resolution should be deferred.
         if (result !== resolvingSignature) {
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -23761,6 +23761,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         if (!links.variances) {
             tracing?.push(tracing.Phase.CheckTypes, "getVariancesWorker", { arity: typeParameters.length, id: getTypeId(getDeclaredTypeOfSymbol(symbol)) });
             const oldVarianceComputation = inVarianceComputation;
+            const saveResolutionStart = resolutionStart;
             if (!inVarianceComputation) {
                 inVarianceComputation = true;
                 resolutionStart = resolutionTargets.length;
@@ -23805,7 +23806,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             }
             if (!oldVarianceComputation) {
                 inVarianceComputation = false;
-                resolutionStart = 0;
+                resolutionStart = saveResolutionStart;
             }
             links.variances = variances;
             tracing?.pop({ variances: variances.map(Debug.formatVariance) });
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -15392,7 +15392,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 }
                 type = anyType;
             }
-            signature.resolvedReturnType = type;
+            signature.resolvedReturnType ??= type;
         }
         return signature.resolvedReturnType;
     }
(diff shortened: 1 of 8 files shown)


ahejlsberg:
@typescript-bot test it

ahejlsberg:
@typescript-bot pack this

jakebailey:
You should be able to pull the tests from Andarist's comments, i.e. copy these ones: https://github.com/microsoft/TypeScript/pull/58323/files#diff-8814cff6472937bf39f151270b038e16552d0896cc67cb5c5382aa432c909acf

I tested locally by pulling this code into my branch and it seemed to work!

DanielRosenwasser:
Those webpack breaks are kind of weird.

@typescript-bot test top800

ahejlsberg:
@typescript-bot test it

ahejlsberg:
> Those webpack breaks are kind of weird

With the temporary resetting of the resolution stack, it is now possible for `getReturnTypeOfSignature` to resolve the same signature twice in a nested manner. When that happens we should always keep the result of the first resolution to complete, and that's now happening with the latest commit.

Andarist:
Could you also create a playground for this one?

jakebailey:
@typescript-bot pack this

ahejlsberg:
The new `redux` error is another example of a ghost error. Without this PR, the error doesn't show in command line builds, but comes and goes when viewing/editing the file in the IDE. With this PR, the error is consistently reported in command line builds and the IDE.

ahejlsberg:
The new `webpack` [error[(https://github.com/microsoft/TypeScript/pull/58337#issuecomment-2080879375) looks legit, the `internalApply` function has no type annotation and is indirectly recursively called in a `return` statement.

Andarist:
I verified that this fixes the other repro that I posted in #57585 (see [here](https://github.com/microsoft/TypeScript/issues/57585#issuecomment-1975323778)). To be honest, it's kinda the same repro as the one that you already have here as [tests/cases/fourslash/issue57585.ts](https://github.com/microsoft/TypeScript/pull/58337/files#diff-aab989aac4782ac6953ca0fc86402767155e06bcdb47cbafb4d87196b0a684e0)  - it just excercices a different path leading to a problem. 

I also checked `effect`'s case reported in #57585 and this **doesn't** fix that problem as far as I was able to verify it manually (cc @mikearnaldi ). So after all... it was worth digging into it :D (cc @jakebailey ). I'll reduce this to a minimal repro case over the weekend and post here.

mikearnaldi:
> I verified that this fixes the other repro that I posted in #57585 (see [here](https://github.com/microsoft/TypeScript/issues/57585#issuecomment-1975323778)). To be honest, it's kinda the same repro as the one that you already have here as [tests/cases/fourslash/issue57585.ts](https://github.com/microsoft/TypeScript/pull/58337/files#diff-aab989aac4782ac6953ca0fc86402767155e06bcdb47cbafb4d87196b0a684e0) - it just excercices a different path leading to a problem.
> 
> I also checked `effect`'s case reported in #57585 and this **doesn't** fix that problem as far as I was able to verify it manually (cc @mikearnaldi ). So after all... it was worth digging into it :D (cc @jakebailey ). I'll reduce this to a minimal repro case over the weekend and post here.

I don't seem to be able to reproduce the same issue anymore, but given it was sporadic I believe you :) hope you'll manage to get a repro which is minimal enough

ahejlsberg:
> I don't seem to be able to reproduce the same issue anymore

FWIW, I haven't been able to reproduce it either.

Andarist:
I could try to minimize this further if requested - I think it's pretty minimal already though (given the complexity of the original: Effect's codebase 😉 ):
```ts
/// <reference path="fourslash.ts" />

// @strict: true
// @target: esnext
// @lib: esnext

//// declare const EffectTypeId: unique symbol;
////
//// type Covariant<A> = (_: never) => A;
////
//// interface VarianceStruct<out A, out E, out R> {
////   readonly _V: string;
////   readonly _A: Covariant<A>;
////   readonly _E: Covariant<E>;
////   readonly _R: Covariant<R>;
//// }
////
//// interface Variance<out A, out E, out R> {
////   readonly [EffectTypeId]: VarianceStruct<A, E, R>;
//// }
////
//// type Success<T extends Effect<any, any, any>> = [T] extends [
////   Effect<infer _A, infer _E, infer _R>,
//// ]
////   ? _A
////   : never;
////
//// declare const YieldWrapTypeId: unique symbol;
////
//// class YieldWrap<T> {
////   readonly #value: T;
////   constructor(value: T) {
////     this.#value = value;
////   }
////   [YieldWrapTypeId](): T {
////     return this.#value;
////   }
//// }
////
//// interface EffectGenerator<T extends Effect<any, any, any>> {
////   next(...args: ReadonlyArray<any>): IteratorResult<YieldWrap<T>, Success<T>>;
//// }
////
//// interface Effect<out A, out E = never, out R = never>
////   extends Variance<A, E, R> {
////   [Symbol.iterator](): EffectGenerator<Effect<A, E, R>>;
//// }
////
//// declare const gen: {
////   <Eff extends YieldWrap<Effect<any, any, any>>, AEff>(
////     f: () => Generator<Eff, AEff, never>,
////   ): Effect<
////     AEff,
////     [Eff] extends [never]
////       ? never
////       : [Eff] extends [YieldWrap<Effect<infer _A, infer E, infer _R>>]
////       ? E
////       : never,
////     [Eff] extends [never]
////       ? never
////       : [Eff] extends [YieldWrap<Effect<infer _A, infer _E, infer R>>]
////       ? R
////       : never
////   >;
//// };
////
//// declare const succeed: <A>(value: A) => Effect<A>;
////
//// gen(function* () {
////   const a = yield* succeed(1);
////   const b/*1*/ = yield* succeed(2);
////   return a + b;
//// });

verify.quickInfoAt("1", "const b: number");
verify.getSemanticDiagnostics([]);
```

ahejlsberg:
I can confirm that @Andarist's Effect repro [here](https://github.com/microsoft/TypeScript/pull/58337#issuecomment-2081164425) reproduces for me and is fixed by this PR.

Andarist:
I'm confused. I posted this repro after testing it on this very branch - it doesn't pass:
```
  1) fourslash tests
       tests/cases/fourslash/issue57585-2.ts
         fourslash test issue57585-2.ts runs correctly:

      AssertionError: At marker '1': quick info text: expected 'const b: any' to equal 'const b: number'
      + expected - actual

      -const b: any
      +const b: number
      
      at _TestState.verifyQuickInfoString (src/harness/fourslashImpl.ts:1863:16)
      at _TestState.verifyQuickInfoAt (src/harness/fourslashImpl.ts:1835:14)
      at Verify.quickInfoAt (src/harness/fourslashInterfaceImpl.ts:272:20)
      at eval (issue57585-2.js:72:8)
      at runCode (src/harness/fourslashImpl.ts:4613:9)
      at runFourSlashTestContent (src/harness/fourslashImpl.ts:4571:5)
      at runFourSlashTest (src/harness/fourslashImpl.ts:4554:5)
      at Context.<anonymous> (src/testRunner/fourslashRunner.ts:59:39)
      at processImmediate (node:internal/timers:478:21)
```

I confirmed now that I am at the correct commit and all, rerun it and it still doesn't pass 😅 

```
> git rev-parse HEAD
abf40126d08dd8cdd1821273fdf0a258a2ec4cfe
```

mikearnaldi:
After fiddling for a while I can confirm this still occurs, more reliably when you specify a return type

Screenshot 2024-04-28 at 10 39 38

Andarist:
@ahejlsberg thanks ❤️

ahejlsberg:
Ok, _now_ it fixes the Effect repro. Sorry for the confusion.

ahejlsberg:
@typescript-bot pack this

ahejlsberg:
@typescript-bot test it

mikearnaldi:
> Ok, _now_ it fixes the Effect repro. Sorry for the confusion.

Thank you immensely!! This makes a lot of difference for Effect, I will test this tomorrow just to double check that in the large codebase nothing wierd happens.

adrian-gierakowski:
Will this land in 5.4.x or do we need to wait for 5.5 to benefit from it? Thanks!

Andarist:
I doubt it. TS usually only backports fixes for recent regressions or other critical things. 

You don’t have to wait till 5.5 though. Once this gets landed a nightly version will be released within 24h from the merge and you can start using it almost immediately

jakebailey:
Looking at the breaks, the webpack one does look like a preexisting circularity, but the redux example doesn't look like one. Any clue what's going on there?

weswigham (approved):
Looks like we're generalizing a mechanism we already used for this issue in variance annotations to signature resolution (and handling the nested calls better by using `??=` instead of `=` at the cache assignments). Seems OK, but maybe we should abstract this concept into a helper like `doOutsideOfCurrentResolutionContext` so it's easier to apply to more entrypoints as-needed? Part of me doubts that signature resolution and variance annotations are the only two places we'd ultimately like to start checking "from the top" regardless of entrypoint (class base type calculations? decorator resolution? apparent member lookup?).

ahejlsberg:
@jakebailey See my comment about the Redux change [here](https://github.com/microsoft/TypeScript/pull/58337#issuecomment-2081113290).

jakebailey:
Oops, missed both in the comment noise, though I can't say I see what the circularity is in that particular line...

ahejlsberg:
> Seems OK, but maybe we should abstract this concept into a helper like `doOutsideOfCurrentResolutionContext`...

Hmm, not so sure about that. We have just two places where we do it and all that needs to be saved/restored is the `resolutionStart` variable. Also, the saving/restoring is conditional, so doesn't lend itself all that well to a function with a callback.

ahejlsberg:
> Oops, missed both in the comment noise, though I can't say I see what the circularity is in that particular line...

I'm not sure, but it's really an orthogonal issue. It was an error before as well, just not consistently so. If there's something we can do to not make it an error, then that's probably a different feature request.

jakebailey (approved):

ssalbdivad:
So with some help from @Andarist and @jakebailey, it seems like it was this change that was responsible for the performance hit in https://github.com/arktypeio/arktype?

```ts
{
    "checkTime": 9.68,
    "types": 483415,
    "instantiations": 2629152
}

devDependencies:

{
    "checkTime": 12.2,
    "types": 514846,
    "instantiations": 2819487
}
```

Issue #52924, opened 2023-02-22
Using `var` in limited contexts to avoid runtime TDZ checks


DanielRosenwasser:
This issue aims to explain and track some optimization work within the TypeScript compiler and language service.

As of TypeScript 5.0, the project's output target was switched from `es5` to `es2018` as part of [a transition to ECMAScript modules](https://github.com/microsoft/TypeScript/pull/51387). This meant that TypeScript could rely on the emit for native (and often more-succinct) syntax supported between ES2015 and ES2018. One might expect that this would unconditionally make things faster, but surprise we encountered was a slowdown from using `let` and `const` natively!

Why is this? It's because `let` and `const` both really provide two features:

* **reasonable scoping rules** - `let`/`const` do not leak beyond the enclosing block scope.
* **definite-declaration-checks** - you cannot access a `let`/`const` variable prior to their declarations being evaluated

What's that second point look like?

```ts
function f() {
    x;
    let x = 10;
    return x;
}
```

Referring to `x` before `let x = 10` has run is supposed to be an error. The idea is that `x` really shouldn't exist as far as anyone knows until its declaration ran. In a sense, this is just enforcing something similar in spirit to the first point around scoping rules - these `let` and `const` bindings can't be referenced until the runtime runs them; but it's enforced in a different way. The binding does sort of exist, but it's specifically an error to access it.

Why is this so subtle? The previous example is "obvious" - `x` is clearly used before it's declared. It happens earlier in the function than the declaration.

It's because we can capture and use these variables in functions. For example.

```ts
function f() {
    let result = g();
    let x = 10;
    return result;

    function g() {
        return x;
    }
}
```

Here, `x` ends up being accessed before it's declared when we called `g()` - even though `g` was declared *after* `x`.

This period while a binding exists but can't be accessed is often called the "temporal dead zone" (or TDZ for short). So JavaScript runtimes need to track whether they've actually hit this declaration point, and this does impose a run time cost. Often, implementations *can* perform optimizations and remove these checks, but it can be tough, and there are limitations.

@jakebailey recently sent out a pull request (#52656) to experiment transforming only `let` and `const` to `var` by using a single Babel transformation. This does lose the TDZ checks, but we developed TypeScript like this for years without too many issues. We found some significant savings - [close to 8% of time reduced on some of our benchmarks](https://github.com/microsoft/TypeScript/pull/52656#issuecomment-1421249664). But we've been resistant to adding another build step.

[...]

PR #58339, opened 2024-04-27
RegExp syntax checking performance

rbuckton:
This makes a few changes to the RegExp syntax checking algorithm for possible performance improvements. This also fixes a small typo in one of the RegExp diagnostic messages.

One of the changes is to ensure we perform proper bounds checks when `pos` might advance past `end`. To accomplish this, I added the `charCodeChecked` and `charCodeUnchecked` functions. `charCodeChecked` performs boundary tests which both improves reliability and avoids potential deoptimizations due to reading past the end of a string. `charCodeUnchecked` is simply a wrapper for `text.charCodeAt` and exists primarily to make it easier to readily identify unchecked vs checked reads from `text`. `charCodeUnchecked` should only be used when we've already performed a bounds check before reading a character, such as in a `while (pos < end)` loop. I've also done the same for code points with `codePointChecked` and `codePointUnchecked`. Each of these functions is small, so ideally will be inlined by an optimizing compiler like the one used by V8.

I've primarily focused on optimizing `scanRegularExpressionWorker` for now. If this seems effective, I will investigate adopting `charCodeChecked` in more places in the scanner in a later PR.

I've also made a few other small perf improvements:
- We now delay allocations for capture group tracking until they are necessary.
- `scanRegularExpressionWorker` has been moved out of `reScanSlashToken` and updated to use the existing `text` and `end` values from the outer scope.
- A few frequently referenced variables from the RegExp scanner have been switched to `var` to avoid TDZ check overhead.

### DIFF
--- a/src/compiler/scanner.ts
+++ b/src/compiler/scanner.ts
@@ -1100,6 +1100,40 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
 
     return scanner;
 
+    /**
+     * Returns the code point for the character at the given position within `text`. This
+     * should only be used when pos is guaranteed to be within the bounds of `text` as this
+     * function does not perform bounds checks.
+     */
+    function codePointUnchecked(pos: number) {
+        return codePointAt(text, pos);
+    }
+
+    // /**
+    //  * Returns the code point for the character at the given position within `text`. If
+    //  * `pos` is outside the bounds set for `text`, `CharacterCodes.EOF` is returned instead.
+    //  */
+    // function codePointChecked(pos: number) {
+    //     return pos >= 0 && pos < end ? codePointUnchecked(pos) : CharacterCodes.EOF;
+    // }
+
+    /**
+     * Returns the char code for the character at the given position within `text`. This
+     * should only be used when pos is guaranteed to be within the bounds of `text` as this
+     * function does not perform bounds checks.
+     */
+    function charCodeUnchecked(pos: number) {
+        return text.charCodeAt(pos);
+    }
+
+    // /**
+    //  * Returns the char code for the character at the given position within `text`. If
+    //  * `pos` is outside the bounds set for `text`, `CharacterCodes.EOF` is returned instead.
+    //  */
+    // function charCodeChecked(pos: number) {
+    //     return pos >= 0 && pos < end ? charCodeUnchecked(pos) : CharacterCodes.EOF;
+    // }
+
     function error(message: DiagnosticMessage): void;
     function error(message: DiagnosticMessage, errPos: number, length: number, arg0?: any): void;
     function error(message: DiagnosticMessage, errPos: number = pos, length?: number, arg0?: any): void {
@@ -1117,7 +1151,7 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
         let isPreviousTokenSeparator = false;
         let result = "";
         while (true) {
-            const ch = text.charCodeAt(pos);
+            const ch = charCodeUnchecked(pos);
             if (ch === CharacterCodes._) {
                 tokenFlags |= TokenFlags.ContainsSeparator;
                 if (allowSeparator) {
@@ -1146,7 +1180,7 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
             }
             break;
         }
-        if (text.charCodeAt(pos - 1) === CharacterCodes._) {
+        if (charCodeUnchecked(pos - 1) === CharacterCodes._) {
             tokenFlags |= TokenFlags.ContainsInvalidSeparator;
             error(Diagnostics.Numeric_separators_are_not_allowed_here, pos - 1, 1);
         }
@@ -1176,9 +1210,9 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
     function scanNumber(): SyntaxKind {
         let start = pos;
         let mainFragment: string;
-        if (text.charCodeAt(pos) === CharacterCodes._0) {
+        if (charCodeUnchecked(pos) === CharacterCodes._0) {
             pos++;
-            if (text.charCodeAt(pos) === CharacterCodes._) {
+            if (charCodeUnchecked(pos) === CharacterCodes._) {
                 tokenFlags |= TokenFlags.ContainsSeparator | TokenFlags.ContainsInvalidSeparator;
                 error(Diagnostics.Numeric_separators_are_not_allowed_here, pos, 1);
                 // treat it as a normal number literal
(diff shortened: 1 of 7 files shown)


rbuckton:
@typescript-bot perf test

rbuckton:
@typescript-bot run dt
@typescript-bot test top400

src/compiler/scanner.ts
```diff
@@ -2464,986 +2497,1007 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
...
+        /** All named capturing groups defined in the regex. */
+        var groupSpecifiers: Set<string> | undefined;
+        /** All references to named capturing groups in the regex. */
+        var groupNameReferences: (TextRange & { name: string; })[] | undefined;
+        /** All numeric backreferences within the regex. */
+        var decimalEscapes: (TextRange & { value: number; })[] | undefined;
+        /** A stack of scopes for named capturing groups. @see {scanGroupName} */
+        var namedCapturingGroupsScopeStack: (Set<string> | undefined)[] = [];
+        var topNamedCapturingGroupsScope: Set<string> | undefined;
+
```
jakebailey, line 2549: Seems like this disable wasn't reenabled.

```suggestion
        /* eslint-enable no-var */
```

src/compiler/scanner.ts
```diff
@@ -1101,6 +1100,40 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
 
     return scanner;
 
+    /**
+     * Returns the code point for the character at the given position within `text`. This
+     * should only be used when pos is guaranteed to be within the bounds of `text` as this
+     * function does not perform bounds checks.
+     */
+    function codePointUnchecked(pos: number) {
+        return codePointAt(text, pos);
```
jakebailey, line 1109: `codePointAt` is also just a little wrapper that pretends that it's always in bounds; maybe that's something to fix up in the follow-up you mentioned.

jakebailey:
The perf seels really good. Is that all from avoiding out of bounds?

rbuckton:
> The perf seels really good. Is that all from avoiding out of bounds?

Yeah, if `charCodeAt` is called for an out of bounds index it returns `NaN`, which isn't an SMI. IIRC, accessing a string out of bounds also results in sub optional compilation as it requires a recompile to introduce additional checks in the compiled code.

src/compiler/scanner.ts
rbuckton: `text.codePointAt` returns `null` for an out of bounds index, which is just as bad as `NaN`. I can probably remove the intermediate call, but it's inlined anyways so it won't affect perf.

jakebailey: Yeah, it was the `null` return that I was thinking about; with this PR applied, debug asserting on out of bounds calls to `codePointAt` fails a few hundred tests I thought would be outside of this PR from other callers (like skipTrivia), though I was surprised to see some code paths which were in the new code:

```
  45) 
       
         conformance tests
           conformance tests for tests/cases/conformance/parser/ecmascript5/SkippedTokens/parserSkippedTokens18.ts
             "before all" hook for "Correct errors for tests/cases/conformance/parser/ecmascript5/SkippedTokens/parserSkippedTokens18.ts":
     Error: Debug Failure. False expression: Invalid index
      at codePointAt (src/compiler/scanner.ts:3922:11)
      at codePointUnchecked (src/compiler/scanner.ts:1109:16)
      at peekExtendedUnicodeEscape (src/compiler/scanner.ts:1730:13)
      at Object.scan (src/compiler/scanner.ts:2264:48)


  285) 
       
         unittests:: tsserver:: applyChangesToOpenFiles
           with updateOpen request:
     Error: Debug Failure. False expression: Invalid index
      at codePointAt (src/compiler/scanner.ts:3922:11)
      at codePointUnchecked (src/compiler/scanner.ts:1109:16)
      at checkForIdentifierStartAfterNumericLiteral (src/compiler/scanner.ts:1304:32)
      at scanNumber (src/compiler/scanner.ts:1298:13)
      at Object.scan (src/compiler/scanner.ts:2134:36)
```

Maybe fixing these would help perf some more too?

jakebailey:
```diff
diff --git a/src/compiler/scanner.ts b/src/compiler/scanner.ts
index 1fb0c63024..3446178dc9 100644
--- a/src/compiler/scanner.ts
+++ b/src/compiler/scanner.ts
@@ -983,6 +983,8 @@ export function isIdentifierPart(ch: number, languageVersion: ScriptTarget | und
 
 /** @internal */
 export function isIdentifierText(name: string, languageVersion: ScriptTarget | undefined, identifierVariant?: LanguageVariant): boolean {
+    if (name === "") return false;
+
     let ch = codePointAt(name, 0);
     if (!isIdentifierStart(ch, languageVersion)) {
         return false;
@@ -1301,7 +1303,7 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
     }
 
     function checkForIdentifierStartAfterNumericLiteral(numericStart: number, isScientific?: boolean) {
-        if (!isIdentifierStart(codePointUnchecked(pos), languageVersion)) {
+        if (!isIdentifierStart(codePointChecked(pos), languageVersion)) {
             return;
         }
 
@@ -1727,7 +1729,7 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
     }
 
     function peekExtendedUnicodeEscape(): number {
-        if (codePointUnchecked(pos + 1) === CharacterCodes.u && codePointUnchecked(pos + 2) === CharacterCodes.openBrace) {
+        if (codePointChecked(pos + 2) === CharacterCodes.openBrace && codePointUnchecked(pos + 1) === CharacterCodes.u) {
             const start = pos;
             pos += 3;
             const escapedValueString = scanMinimumNumberOfHexDigits(1, /*canHaveSeparators*/ false);
@@ -2945,7 +2947,7 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
                         error(Diagnostics.A_character_class_range_must_not_be_bounded_by_another_character_class, maxStart, pos - maxStart);
                         continue;
                     }
-                    if (!minCharacter) {
+                    if (!minCharacter || !maxCharacter) {
                         continue;
                     }
                     const minCharacterValue = codePointAt(minCharacter, 0);
@@ -3919,6 +3921,7 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
 
 function codePointAt(s: string, i: number): number {
     // TODO(jakebailey): this is wrong and should have ?? 0; but all users are okay with it
+    Debug.assert(i >= 0 && i < s.length, "Invalid index");
     return s.codePointAt(i)!;
 }
 
```

Is what I got that made this particular function pass.

rbuckton: If I do the same assert in both `codePointUnchecked` and `charCodeUnchecked` we end up with nearly 20k test failures, so we're definitely not performing accurate bounds checks during scan.

src/compiler/scanner.ts
```diff
@@ -2116,19 +2149,19 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
                         }
                     }
 
-                    if (text.charCodeAt(pos + 1) === CharacterCodes.lessThan) {
-                        if (text.charCodeAt(pos + 2) === CharacterCodes.equals) {
+                    if (charCodeUnchecked(pos + 1) === CharacterCodes.lessThan) {
+                        if (charCodeUnchecked(pos + 2) === CharacterCodes.equals) {
```
DanielRosenwasser, line 2153: Many of these should probably be switched over to the checked version, but just never trigger because it's rare to hit these tokens at the end of a file, right?

DanielRosenwasser:
Might be worth mentioning in the PR description that you also `var`-ified the regular expression parser (https://github.com/microsoft/TypeScript/issues/52924).

DanielRosenwasser:
Also other specific optimizations - like hoisting out helper functions and lazily initializing state for captures.

jakebailey (approved):
Oops, forgot my green check.

rbuckton:
I've been working on expanding this throughout scanner.ts and writing up a lint rule to ensure enforcement, so I'll probably need another review shortly.

src/compiler/scanner.ts
rbuckton: See #58362 for most of the rest of scanner.

Issue #58274, opened 2024-04-21
OOM typescript, mac m1+.


dvun0g:
### 🔎 Search Terms

I tried to find similar typescript OOM in issue github and on the Internet.

### 🕗 Version & Regression Information

typescript - 5.2.2
macos - macOS Ventura 13.6
chip - Apple M2

This problem appeared with typescript 5+, updating to the latest version of typescript does not prevent the error. Regressing to version 4 of typescript solves this problem. Also, this problem is reproduced only on macs with m1 processor and higher.

### ⏯ Playground Link

-

### 💻 Code

-

### 🙁 Actual behavior

Typescript is flowing with OOM errors.

### 🙂 Expected behavior

Working properly with memory.

### Additional information about the issue

I am running eslint on a specific area where the ts/ts files with the react application are located. As a result, the process kills with a memory error - that is, OOM occurs.

Below I provide two snapshots with memory, both to be made after using garbage collector. To see exactly which typescript objects are leaking, it is enough to compare these reports.

Below I attach a link to google drive with snapshot memory. It was not possible to attach this archive directly to the github issue due to the larger file size. - https://drive.google.com/file/d/1XHJCz79eAwBKiHDS2-vtpcR--7b_xF_Z/view?usp=drive_link

jakebailey:
If it used to work and now doesn't, consider trying to bisect the problem using https://www.npmjs.com/package/every-ts.

dimensi:
Hello @jakebailey , found a problematic commit with bisect. Everything crashes after it.

```bash
ff92ab0ac92b66397bd3b92ff5a3aca3c1ad8f43 is the first bad commit
commit ff92ab0ac92b66397bd3b92ff5a3aca3c1ad8f43
Author: Ryan Cavanaugh <RyanCavanaugh@users.noreply.github.com>
Date:   Fri Jan 20 17:20:26 2023 -0800

    Change default newLine, forceConsistentCasingInFileNames (#52298)
    
    Co-authored-by: Jake Bailey <5341706+jakebailey@users.noreply.github.com>
    Co-authored-by: Sheetal Nandi <shkamat@microsoft.com>
```

jakebailey:
This sounds like #58011 / https://github.com/typescript-eslint/typescript-eslint/issues/1192. Can you provide a codebase which exhibits this problem?

jakebailey:
[...]

PR #58352, opened 2024-04-29
Ensure equal FileIncludeReasons are not duplicitively added

jakebailey:
When we add file include reasons, they're effectively just stored per file path in a big list (a MultiMap). When we reuse a program, we keep using that same list, but may rediscover the same files over and over again, pushing more and more into that mapping.

Whenever we call `createDiagnosticExplainingFile`, we search through the file include reasons and include them in the related info. In the case of `forceConsistentCasingInFileNames`, we'll build a potentially massive diagnostic that references these other files.

Now, imagine the intersection of the two; I'm in eslint with a watch program, `forceConsistentCasingInFileNames` is the new default, but its diagnostics are never given to anyone. Maybe a file with the wrong case is queried over and over as the user types, it keeps getting added to a reused fileReasons map, then the `forceConsistentCasingInFileNames` just keeps building bigger and bigger diagnostics.

If we ensure that `fileReason` doesn't duplicitively include reasons, we avoid the blowup. It's still not great that there are bigish diagnostics, or that clients may query using a different path, etc, but I think this may handle the OOM. And the baseline change shows that duplication removes some errors.

It's also worth noting that `DiagnosticCollection` doesn't seem to handle deduplicating these diagnostics at all, probably due to the duplication or differing orders. But, it may after this PR? (Hard to observe.)

Relatd:

- https://github.com/microsoft/TypeScript/issues/58011
- https://github.com/typescript-eslint/typescript-eslint/issues/1192
- https://github.com/microsoft/TypeScript/issues/58274

### DIFF
--- a/src/compiler/program.ts
+++ b/src/compiler/program.ts
@@ -83,6 +83,7 @@ import {
     fileExtensionIsOneOf,
     FileIncludeKind,
     FileIncludeReason,
+    fileIncludeReasonIsEqual,
     fileIncludeReasonToDiagnostics,
     FilePreprocessingDiagnostics,
     FilePreprocessingDiagnosticsKind,
@@ -3797,7 +3798,11 @@ export function createProgram(rootNamesOrOptions: readonly string[] | CreateProg
     }
 
     function addFileIncludeReason(file: SourceFile | undefined, reason: FileIncludeReason) {
-        if (file) fileReasons.add(file.path, reason);
+        if (file) {
+            const existing = fileReasons.get(file.path);
+            if (some(existing, r => fileIncludeReasonIsEqual(r, reason))) return;
+            fileReasons.add(file.path, reason);
+        }
     }
 
     function addFileToFilesByName(file: SourceFile | undefined, path: Path, fileName: string, redirectedPath: Path | undefined) {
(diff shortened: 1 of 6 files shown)


jakebailey:
@typescript-bot test it
@typescript-bot pack this

jakebailey:
@typescript-bot test it
@typescript-bot pack this

sheetalkamat:
> When we reuse a program, we keep using that same list, but may rediscover the same files over and over again, pushing more and more into that mapping.

Thats not true.. Unless program is completely reused the `fileReasons` from old program are not reused so they should not change between programs.  
What you have fixed here seems like a issue where file(`a`) may be included by say `import` from one of the file(`b`) that was imported though different root file(`c`). Then if that `b` is also root file, we will process its imports again so file `c` is added with same reason and that is what this deduplication is handling.. 

If you look for `fileReasons` it does not change between programs. Its only changed through `findSourceFileWorker` and that is not called when we are trying to determine if program can be reused. (`tryReuseProgramStructure`)

jakebailey:
Gotcha; the deduping is somewhat nice though, but if it's not actually the problem then the deduping is probably just plain worse for perf and I should just ditch this PR.

jakebailey:
Closing in favor of #58398, which includes some aspects of this PR.

PR #58358, opened 2024-04-29
fixed types for Object.values and Object.entries

### LINKED ISSUE #26010: Object.values and Object.entries return type any when passing an object defined as having number keys
**TypeScript Version:**  3.1.0-dev.20180727

**Search Terms:** Object.values Object.entries wrong type inference any

**Code**

```ts
interface I {
  [x: number]: string | null;
}

const o: I = { 5: 'test', 8: null };
const v = Object.values(o);
```

**Expected behavior:** Type of `v` should be `(string | null)[]`.

**Actual behavior:** Type of `v` is `any[]`.

**Playground Link:** Typescript Playground doesn't support `Object.values()`.

**Related Issues:** https://github.com/Microsoft/TypeScript/issues/21089

This bug seems to occur due to the type definitions for ObjectConstructor, which used to include the possibility of number keys in their type signatures. I'm not sure exactly why this was changed to `ArrayLike` to resolve the related issue I linked ([this is the exact PR](https://github.com/Microsoft/TypeScript/pull/21129/files)), but it breaks the type inference on the results of `Object.values()` and `Object.entries()` when passing it a type defined as having `number` keys.

```ts
interface ObjectConstructor {
    /**
     * Returns an array of values of the enumerable properties of an object
     * @param o Object that contains the properties and methods. This can be an object that you created or an existing Document Object Model (DOM) object.
     */
    values<T>(o: { [s: string]: T } |  ArrayLike<T>): T[];

    // ...
}
```

This type signature used to be `values<T>(o: { [s: string]: T } | { [n: number]: T }): T[]`, and with this older signature, the example code properly types the result of `Object.values()`.

typeholes:
Fixes #26010

### DIFF
--- a/tests/cases/conformance/es2017/useObjectValuesAndEntries5.ts
+++ b/tests/cases/conformance/es2017/useObjectValuesAndEntries5.ts
@@ -1,0 +1,19 @@
+// @target: es6
+// @lib: es2017
+
+{
+enum X {A, B}
+
+const xsV = Object.values(X);
+const xsE = Object.values(X);
+
+interface I {
+  [x: number]: string | null;
+}
+
+const o: I = { 5: 'test', 8: null };
+const v = Object.values(o);
+const e = Object.entries(o);
+
+
+}
--- a/tests/cases/conformance/es2017/useObjectValuesAndEntries5.ts
+++ b/tests/cases/conformance/es2017/useObjectValuesAndEntries5.ts
@@ -1,5 +1,6 @@
 // @target: es6
 // @lib: es2017
+// @strictNullChecks: true
 
 {
 enum X {A, B}
--- a/src/lib/es2017.object.d.ts
+++ b/src/lib/es2017.object.d.ts
@@ -3,9 +3,9 @@ interface ObjectConstructor {
      * Returns an array of values of the enumerable properties of an object
      * @param o Object that contains the properties and methods. This can be an object that you created or an existing Document Object Model (DOM) object.
      */
-    values<T>(o: { [s: string]: T; } ): T[];
-    values<T>(o: { [s: number]: T; } ): T[];
-    values<T>(o: { [s: string|number]: T; } ): T[];
+    values<T>(o: { [s: string]: T; }): T[];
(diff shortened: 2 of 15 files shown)


jakebailey:
@typescript-bot test it

typeholes:
Once #58372 🙂 lands the overload should probably be replaced by a simple union.

typeholes:
fixes #58376

ehoogeveen-medweb:
Does this also need some sort of handling for objects with `symbol` type keys? Properties with such keys won't be returned, but `Object.values()` and `Object.entries()` should still be callable on such objects.

typeholes:
> Does this also need some sort of handling for objects with `symbol` type keys? Properties with such keys won't be returned, but `Object.values()` and `Object.entries()` should still accept such objects.

There is already an overload for `values` and `entries`, so they will be accepted.  
for example,  `values(o: {}): any[];`  Of course the `any[]` return type is not ideal, but I'm not sure anything else is workable for the fallback.

ehoogeveen-medweb:
Thanks, I didn't look beyond the context of the diff :)

Locally I use something like this:
```ts
type KeyToString<K extends number | string | symbol> = K extends string ? K : K extends number ? `${K}` : never;

keys<K extends number | string | symbol, V>(o: Record<K, V>): KeyToString<K>[];
entries<K extends number | string | symbol, V>(o: Record<K, V>): [KeyToString<K>, V][];
values<K extends number | string | symbol, V>(o: Record<K, V>): V[];
```
but aside from being too complex for the standard library, that probably opens up a whole can of worms if applied to everything (works well for my code though!).

tests/cases/conformance/es2017/useObjectValuesAndEntries5.ts
```diff
@@ -0,0 +1,20 @@
+// @target: es6
+// @lib: es2017
+// @strictNullChecks: true
+
+{
+enum X {A, B}
+
+const xsV = Object.values(X);
+const xsE = Object.values(X);
```
Renegade334, line 9:
```suggestion
const xsE = Object.entries(X);
```

PR #58362, opened 2024-04-29
Perform bounds-checking on string input in scanner

rbuckton:
Following on to #58339, this performs the same bounds checking throughout most of the rest of the scanner.

This also adds the `local/bounds-check` lint rule to warn if future changes incorrectly perform unchecked calls to `charCodeAt` or `codePointAt` on `text`. The lint rule performs some rudimentary analysis to determine whether a call to `codePointUnchecked` or `charCodeUnchecked` follows a valid boundary check to cut down on excess `// eslint-disable-line` comments for cases like these:

```ts
while (pos < end) {
  const ch = charCodeUnchecked(pos); // ok, checked on preceding line
  ...
}
```

```ts
if (pos + 1 < end && charCodeUnchecked(pos + 1) === CharacterCodes._) ...
```

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -144,7 +144,8 @@
         "local/no-in-operator": "error",
         "local/debug-assert": "error",
         "local/no-keywords": "error",
-        "local/jsdoc-format": "error"
+        "local/jsdoc-format": "error",
+        "local/bounds-check": "error"
     },
     "overrides": [
         // By default, the ESLint CLI only looks at .js files. But, it will also look at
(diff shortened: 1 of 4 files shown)


rbuckton:
@typescript-bot perf test

jakebailey:
@typescript-bot perf test

rbuckton:
Looks like there was no noticeable performance improvement with this change, so we should only take it if we think the bounds checking enforcement is worth adding purely for the sake of correctness and stability.

DanielRosenwasser:
It seems like parse is (only very slightly) worse - I wonder what happens if you switch that initial call in `scan` to the old unchecked version, but I'm guessing it's more death by a thousand cuts.

PR #58364, opened 2024-04-29
Enable JS emit for noCheck and noCheck for transpileModule

weswigham:
This PR enables `noCheck` for JS emit, and tests *all* js emit tests with `noCheck`, just as we do declaration emit tests with `noCheck`. It also enables `noCheck` for `transpileModule` by default.


weswigham:

tests/baselines/reference/autoAccessor1(target=es5).js
```diff
@@ -0,0 +1,61 @@
...
+    static accessor c: any;
+    static accessor d = 2;
+}
+
+
+
+
+!!!! File autoAccessor1.js missing from original emit, but present in noCheck emit
+//// [autoAccessor1.js]
+var __classPrivateFieldGet = (this && this.__classPrivateFieldGet) || function (receiver, state, kind, f) {
```
weswigham, line 16: The `target` error for auto-accessors, `Properties with the 'accessor' modifier are only available when targeting ECMAScript 2015 and higher.` is a checker error. So when it's suppressed, we happily do emit.

tests/baselines/reference/importDeclWithDeclareModifier.js
```diff
@@ -15,3 +15,16 @@ Object.defineProperty(exports, "__esModule", { value: true });
...
+!!!! File importDeclWithDeclareModifier.js differs from original emit in noCheck emit
+//// [importDeclWithDeclareModifier.js]
+===================================================================
+--- Expected	The full check baseline
++++ Actual	with noCheck set
+@@ -1,5 +1,3 @@
+ "use strict";
+ Object.defineProperty(exports, "__esModule", { value: true });
+-exports.a = void 0;
+-exports.a = x.c;
```
weswigham, line 29: I think this is a case where the `noCheck` output is an improvement. As a perf optimization, I made reference marking with an `Unspecified` hint skip all `Ambient` nodes - that causes this change. The `export declare import` statement in this file usually gets transformed into the output js, even though it has a `declare` modifier. That's likely because the `declare` modifier on an import is an error, so the alias marking is done unconditionally. In any case, since the input has a syntax/grammar type error, I think it's fine for the output to diverge a bit like this. I could bring it in-line (either by changing normal behavior to elide this, or `noCheck` to retain it), but I'm not sure it's worth the effort, given the error.

weswigham:
Much better, no more check regressions, only emit ones, and those aughta get fixed with proper application of `.calculatedFlags` during the checker walk (though it's admittedly pretty odd that only *some* projects with emit show the perf regression - maybe those projects have settings such that they don't really look at `NodeCheckFlags`?).

weswigham:

weswigham:

src/compiler/types.ts
```diff
@@ -5761,10 +5777,11 @@ export interface EmitResolver {
     isValueAliasDeclaration(node: Node): boolean;
     isReferencedAliasDeclaration(node: Node, checkChildren?: boolean): boolean;
     isTopLevelValueImportEqualsWithEntityName(node: ImportEqualsDeclaration): boolean;
-    getNodeCheckFlags(node: Node): NodeCheckFlags;
+    hasNodeCheckFlag(node: Node, flags: LazyNodeCheckFlags): boolean;
```
jakebailey, line 5780: All in all, I wonder how many other flag checks could be done like this (I did something similar for another flag which helped perf too)

weswigham:

jakebailey:
For fun

@typescript-bot perf test this predictable

jakebailey:
@typescript-bot perf test this predictable

jakebailey:
The perf results are still seemingly showing a regression in emit time on various projects. Not sure what's going on there but I'll try and profile...

weswigham:
I mentioned it in our meeting, but the only things not `noCheck`-flagged (aside from change the format of the API function from "get flags" to "has flag") at this point are [this range](https://github.com/microsoft/TypeScript/pull/58364/files#diff-d9ab6589e714c71e657f601cf30ff51dfc607fc98419bf72e04f6b0fa92cc4b8R48582-R48587) and [this range](https://github.com/microsoft/TypeScript/pull/58364/files#diff-d9ab6589e714c71e657f601cf30ff51dfc607fc98419bf72e04f6b0fa92cc4b8R4259-R4265), each of which solve a bug that only crops up when you do out of order checking (namely: cjs merge symbols not manifesting early enough and type only imports not being marked). You could `noCheck`-flag both of them if you really wanted, but I feel like it'd make using those API methods a bit footgun-y. But if one of them has outsized (though not *that* outsized) performance impact, I suppose we have little choice?

src/compiler/emitter.ts
```diff
@@ -924,6 +930,14 @@ export function emitFiles(resolver: EmitResolver, host: EmitHost, targetSourceFi
         forEachChild(node, collectLinkedAliases);
     }
 
+    function markLinkedReferences(file: SourceFile) {
+        ts.forEachChildRecursively(file, n => {
```
sheetalkamat, line 934: can this instead check if file doesnt have `TypeChecked` flag and do this work and then we can get rid of `getDiagnostics` all together from `getEmitResolver()`

tests/baselines/reference/autoAccessor3(target=es5).js
```diff
@@ -0,0 +1,61 @@
+//// [tests/cases/conformance/classes/propertyMemberDeclarations/autoAccessor3.ts] ////
```
jakebailey, line 1: Just so I understand, the reason all of the baselines change is because we've enabled `noCheck` for _all_ of them?

jakebailey:
I'm having a pretty hard time figuring out where the slowdown is; the vscode benchmark takes a long time so two runs are just not comparable, even in predictable mode on a perf locked core.

src/compiler/emitter.ts
weswigham: Not quite - the `TypeChecked` flag is, ironically, set on files whose type checking is skipped. Maybe it's a misnomer, but it really just means "Has `checkSourceFile` been called on this node?". I get your drift though and, yeah, we could probably do a bit of refactoring to use a node flag to track the diagnostic walk state of the containing file. I'll probably put that up as a separate cleanup, though. I also think I want to move around/refactor the API logic for `collectLinkedAliases` so the API is harder to misuse, and will probably change `markLinkedReferences` to match, so I'll put all those nonfunctional changes in a separate PR (which should have a 0-test diff).

src/compiler/checker.ts
```diff
@@ -48676,6 +48711,129 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         return nodeLinks[nodeId]?.flags || 0;
     }
 
+    function hasNodeCheckFlag(node: Node, flag: LazyNodeCheckFlags) {
+        calculateNodeCheckFlagWorker(node, flag);
+        return !!(getNodeCheckFlags(node) & flag);
+    }
+
+    function calculateNodeCheckFlagWorker(node: Node, flag: LazyNodeCheckFlags) {
```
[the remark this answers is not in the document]
weswigham: I also do reuse the actual flag-setting functions, just not the traversal to reach and invoke them. So the "should this flag be set" logic should be synced, at least.

DanielRosenwasser:
@typescript-bot pack this

jakebailey:
I thought I'd try this out on a ts-loader project, so tested out pyright's build before and after. Somehow, it's slower? I suppose this doesn't actually use transpileModule since they're using const enums and everything :(

```console
$ hyperfine 'node --max_old_space_size=8192 ./node_modules/webpack-cli/bin/cli.js --mode development'
Benchmark 1: node --max_old_space_size=8192 ./node_modules/webpack-cli/bin/cli.js --mode development
  Time (mean ± σ):     14.899 s ±  0.221 s    [User: 22.878 s, System: 1.591 s]
  Range (min … max):   14.379 s … 15.200 s    10 runs
 
$ hyperfine 'node --max_old_space_size=8192 ./node_modules/webpack-cli/bin/cli.js --mode development'
Benchmark 1: node --max_old_space_size=8192 ./node_modules/webpack-cli/bin/cli.js --mode development
  Time (mean ± σ):     15.110 s ±  0.164 s    [User: 23.219 s, System: 1.541 s]
  Range (min … max):   14.888 s … 15.395 s    10 runs
```

weswigham:
> I suppose this doesn't actually use transpileModule since they're using const enums and everything :(

Ya, `ts-loader` only uses `transpileModule` if you pass `transpileOnly`.

jakebailey:
In lieu of real-world benchmarks, I used `tinybench` and tested this before/after on files of various sizes in our codebase, calling:

```js
ts.transpileModule(input, {
    fileName,
    compilerOptions: { target: tsOld.ScriptTarget.ESNext, module: tsOld.ModuleKind.Preserve } },
);
```

```
Benchmarking src/typescript/typescript.ts (26 lines)...
┌─────────┬───────────────────────┬─────────┬───────────────────┬──────────┬─────────┐
│ (index) │ Task Name             │ ops/sec │ Average Time (ns) │ Margin   │ Samples │
├─────────┼───────────────────────┼─────────┼───────────────────┼──────────┼─────────┤
│ 0       │ 'old transpileModule' │ '1,092' │ 914970.5937785943 │ '±1.49%' │ 10930   │
│ 1       │ 'new transpileModule' │ '1,478' │ 676558.4366416298 │ '±1.69%' │ 14781   │
└─────────┴───────────────────────┴─────────┴───────────────────┴──────────┴─────────┘
35.24% faster, 1.35x faster, 26.06% less time than before
```
```
Benchmarking src/compiler/path.ts (1017 lines)...
┌─────────┬───────────────────────┬─────────┬────────────────────┬──────────┬─────────┐
│ (index) │ Task Name             │ ops/sec │ Average Time (ns)  │ Margin   │ Samples │
├─────────┼───────────────────────┼─────────┼────────────────────┼──────────┼─────────┤
│ 0       │ 'old transpileModule' │ '58'    │ 17076565.800341412 │ '±1.07%' │ 586     │
│ 1       │ 'new transpileModule' │ '126'   │ 7927894.958795462  │ '±1.06%' │ 1262    │
└─────────┴───────────────────────┴─────────┴────────────────────┴──────────┴─────────┘
115.40% faster, 2.15x faster, 53.57% less time than before
```
```
Benchmarking src/server/project.ts (3064 lines)...
┌─────────┬───────────────────────┬─────────┬───────────────────┬──────────┬─────────┐
│ (index) │ Task Name             │ ops/sec │ Average Time (ns) │ Margin   │ Samples │
├─────────┼───────────────────────┼─────────┼───────────────────┼──────────┼─────────┤
│ 0       │ 'old transpileModule' │ '12'    │ 79510811.21428606 │ '±1.14%' │ 126     │
│ 1       │ 'new transpileModule' │ '31'    │ 31801029.40000019 │ '±1.19%' │ 315     │
└─────────┴───────────────────────┴─────────┴───────────────────┴──────────┴─────────┘
150.03% faster, 2.50x faster, 60.00% less time than before
```
```
Benchmarking src/compiler/emitter.ts (6266 lines)...
┌─────────┬───────────────────────┬─────────┬───────────────────┬──────────┬─────────┐
│ (index) │ Task Name             │ ops/sec │ Average Time (ns) │ Margin   │ Samples │
├─────────┼───────────────────────┼─────────┼───────────────────┼──────────┼─────────┤
│ 0       │ 'old transpileModule' │ '7'     │ 135490544.418919  │ '±1.00%' │ 74      │
│ 1       │ 'new transpileModule' │ '15'    │ 65744860.17647067 │ '±1.03%' │ 153     │
└─────────┴───────────────────────┴─────────┴───────────────────┴──────────┴─────────┘
106.09% faster, 2.06x faster, 51.48% less time than before
```
```
Benchmarking src/compiler/checker.ts (51965 lines)...
┌─────────┬───────────────────────┬─────────┬────────────────────┬──────────┬─────────┐
│ (index) │ Task Name             │ ops/sec │ Average Time (ns)  │ Margin   │ Samples │
├─────────┼───────────────────────┼─────────┼────────────────────┼──────────┼─────────┤
│ 0       │ 'old transpileModule' │ '0'     │ 1719514998.7000027 │ '±5.94%' │ 10      │
│ 1       │ 'new transpileModule' │ '1'     │ 839784896.9166638  │ '±2.26%' │ 12      │
└─────────┴───────────────────────┴─────────┴────────────────────┴──────────┴─────────┘
104.76% faster, 2.05x faster, 51.16% less time than before
```

So, for files of a non-trivial size, it sure seems like a good 2x speed boost for `transpileModule` users.

I have not been able to pin down the perf regression, if there even is one. But I'm pretty sure we should just take this anyhow.

jakebailey (approved):
Looks good to me, though I'm not 100% thrilled with the increase in test time; it looks like it increases CI time from something like 7 minutes to 8.

sheetalkamat (approved):

PR #58366, opened 2024-04-29
Move all js emit alias marking behavior behind a single entrypoint

weswigham:
This is one of the two bigger checker refactors in #58364 - by validating its' perf separately, I can narrow down the perf regression there (and possibly simplify merging it by getting this merged separately).

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -29257,7 +29257,6 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         if (isThisIdentifier(left) || !isIdentifier(left)) {
             return;
         }
-        const right = isPropertyAccessExpression(location) ? location.name : location.right;
         const leftType = parentType || checkExpressionCached(left);
         const parentSymbol = getNodeLinks(left).resolvedSymbol;
         if (!parentSymbol || parentSymbol === unknownSymbol) {
@@ -29283,6 +29282,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         }
         let prop = propSymbol;
         if (!prop && !parentType) {
+            const right = isPropertyAccessExpression(location) ? location.name : location.right;
             const lexicallyScopedSymbol = isPrivateIdentifier(right) && lookupSymbolForPrivateIdentifierDeclaration(right.escapedText, right);
             const assignmentKind = getAssignmentTargetKind(location);
             const apparentType = getApparentType(assignmentKind !== AssignmentKind.None || isMethodAccessForCall(location) ? getWidenedType(leftType) : leftType);
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -1112,6 +1112,18 @@ import * as performance from "./_namespaces/ts.performance";
 const ambientModuleSymbolRegex = /^".+"$/;
 const anon = "(anonymous)" as __String & string;
 
+const enum ReferenceHint {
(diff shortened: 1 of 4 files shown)


weswigham:

weswigham:
Oof, okay, much to my surprise, it's this half that's most problematic as-is. I'm going to guess the closure allocations on each call (which are just to organize and prevent calling the inner functions) are no good. Let's remove those...

weswigham:

weswigham:

tests/baselines/reference/moduleImport.errors.txt
```diff
@@ -1,13 +1,10 @@
-moduleImport.ts(2,17): error TS2339: Property 'Y' does not exist on type 'typeof X'.
 moduleImport.ts(2,17): error TS2694: Namespace 'X' has no exported member 'Y'.
 
 
-==== moduleImport.ts (2 errors) ====
+==== moduleImport.ts (1 errors) ====
     module A.B.C {
     	import XYZ = X.Y.Z;
     	               ~
-!!! error TS2339: Property 'Y' does not exist on type 'typeof X'.
```
weswigham, line 9: We drop this (duplicate-ish) error because now, rather than relying on `checkExpressionCached` to paint the reference recursively for the `import =` statement, we directly use `markIdentifierAliasReferenced` on the leftmost identifier in the entity name. Which should be faster, and, as this test shows, shouldn't cause us to drop any real errors, since any missing members get reported when we actually resolve the name (with a more sensible error, even).

src/compiler/checker.ts
```diff
@@ -33082,28 +33307,13 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
-        if (
-            isIdentifier(left) && parentSymbol && (
-                getIsolatedModules(compilerOptions) ||
-                !(prop && (isConstEnumOrConstEnumOnlyModule(prop) || prop.flags & SymbolFlags.EnumMember && node.parent.kind === SyntaxKind.EnumMember)) ||
-                shouldPreserveConstEnums(compilerOptions) && isExportOrExportExpression(node)
-            )
-        ) {
-            markAliasReferenced(parentSymbol, node);
-        }
+        markLinkedReferences(node, prop, leftType);
```
weswigham, line 33316: I'm hoping reusing the `leftType`+`prop` here negates the performance impact this change had. Non-checker uses aren't expected to have done this work already, so it does still need to have the logic to lookup the `prop` itself, and is probably the costly part of alias marking. That you can't just say `A.B;` references `A` because `A.B` might be a `const enum` reference that gets inlined (and so you need to typecheck) is actually quite annoying from an encapsulation point of view - it's the _one place_ in this process where we have to resort to typechecking. I guess that's why it's not done in `isolatedModules`. 😄

weswigham:
Better - down to 1-2% check time, if any, but hopefully I can manage a bit more... The only thing now would seem to be that some of the checks on `location` (and weather it counts as a location whose reference we should mark) are repeated between caller and within the reference marker. I can probably do some hinting like the emitter does to avoid those at callsites where we know the input location's shape in advance.

weswigham:

weswigham:

weswigham:

jakebailey (approved):

andrewbranch (approved):
This has been sorely needing a refactor for a long time 👍 It's still wild to look at how complicated it is, but at least it's all in one place now.

PR #58372, opened 2024-04-30
Cache discriminated contextual types

ahejlsberg:
While researching another issue, @jakebailey noticed [this statement](https://github.com/microsoft/TypeScript/blob/main/src/compiler/transformers/jsx.ts#L681) takes an exceedingly long time to type check. With the caching added in this PR, the check time for the statement drops from 250ms to about zero on my machine.

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -30879,7 +30879,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 // That would evaluate mapped types with array or tuple type constraints too eagerly
                 // and thus it would prevent `getTypeOfPropertyOfContextualType` from obtaining per-position contextual type for elements of array literal expressions.
                 // Apparent type of other mapped types is already the mapped type itself so we can just avoid calling `getApparentType` here for all mapped types.
-                t => t.flags & TypeFlags.Nullable ? neverType : getObjectFlags(t) & ObjectFlags.Mapped ? t : getApparentType(t),
+                t => getObjectFlags(t) & ObjectFlags.Mapped ? t : getApparentType(t),
                 /*noReductions*/ true,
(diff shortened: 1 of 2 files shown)


ahejlsberg:
@typescript-bot test it

jakebailey:
@typescript-bot test it

Sorry, the new function app we had to deploy has some scaling issues being worked out.

jakebailey:
Wow, that's actually significant., 3% on our self test (expected), but also 1.5% in mui-docs.

jakebailey (approved):

ahejlsberg:
@typescript-bot test it

ahejlsberg:
Hmm, removing `null` and `undefined` from apparent contextual types seems to be a net negative. I'll get rid of that again.

ssalbdivad:
I bisected a perf regression typechecking https://github.com/arktypeio/arktype and found the following:

```ts
{
    "checkTime": 9.68,
    "types": 483415,
    "instantiations": 2629152
}

devDependencies:

{
    "checkTime": 12.2,
    "types": 514846,
    "instantiations": 2819487
}
```

It seems likely this change was somehow involved?

Andarist:
Guided by `package.json#gitHead` in both of those we can see the diff between those versions here:
https://github.com/microsoft/TypeScript/compare/3358157469d805df1deec01d26269c5d9d6b23ae...dc316afea4907087fcbac28a0fad95273926715e

This PR isn't part of that.

ssalbdivad:
My mistake, I likely jumped to conclusions when I saw this in the release notes as a likely candidate since I use quite a lot of unions internally.

@Andarist and I have been looking at which of the changes from that day could have caused such a large increase in type instantiations and check time and haven't been able to identify it yet.

PR #58377, opened 2024-04-30
Remove seemingly unused SymbolLink `constEnumReferenced`

weswigham:
This `SymbolLink` isn't referenced anywhere other than where it's set - presumably we stopped using it awhile ago and didn't notice we removed its' last reference.


weswigham:
Via @andrewbranch 
> Ah, it supported the error in
--importsNotUsedAsValues error
which we decided not to replace with non-deprecated options, so that code became dead when we recently removed the 5.0 deprecations

andrewbranch (approved):

Issue #58351, opened 2024-04-29
Strange runtime caching bug


JosXa:
### 🔎 Search Terms

`is:issue runtime caching as const `

### 🕗 Version & Regression Information

- This is a crash
- This is the behavior in every version I tried

### ⏯ Playground Link

https://www.typescriptlang.org/play/?target=99#code/C4TwDgpgBA8gRgKwgY2AMQE4HsC2BRAO2AwEsIBnAHgBUA+KAXimqggA9gICATcqDCAENuWAgBsQUANokCAMwgYoAaQiT2nHn3LFZAcygAfKAQCuOOIqNRyIC1jEAaKLIVKAaoLGmIAXSm+AFBQUAD8UADe0gDWai4EKmq+AFxQnt7QAL7BUKkEEABuigDcgYGgkIkg6T4ACoIkGFR0jJE5UsrxULEgWHLMKdLKztQdvr6lmVI9fQOlgexgWBjAUHKmBKgkolA4gmDwSKiEuhSUOciiOsysHFy8UFiIKMCOF1er1ACy+5Dct5oHlIdKQCAZjGYLFZjLZ7E4oBtogQsAB3Ai+N60AAUOSeCFS1DeIT2YEgGFSWKkgmccEGqmqXjqDSaNFoAEpGPRvr8INw3mzUocXphcCdSGduaTeQF6BEcgB6eUAAWA5AAtIsXhqMNglABBMQCYSSMDYMkSKAVXmWgAW2FMehttug5BIegIgmApgEOQEXowCSFqAAdHJsPgiOLyFig8Bg1xTtG8WzgySsSSyWy2ZMypcCNdONcmCTY2KyNGooJUgBGGmpABMUEyzkpbGcIF8HIY9CkbagHezZSAA

### 💻 Code

```ts
type ObjectFromEntries<T> = T extends readonly [infer Key extends string | number | symbol, infer Value][]
  ? { [key in Key]: Value }
  : never;

type KeyValuePairs<T> = {
  [K in keyof T]: [K, T[K]];
}[keyof T];

export function mapObjectEntries<
  const T extends object,
  const TMapped extends [string | number | symbol, unknown],
>(
  obj: T,
  mapper: ([a, b]: KeyValuePairs<T>) => TMapped,
): ObjectFromEntries<TMapped[]> {
  //@ts-expect-error Already properly typed through the signature
  return Object.fromEntries(Object.entries(obj).map(mapper));
}

[...]

PR #58378, opened 2024-04-30
Fixed a ghost error popping up after completions request caching a wrong contextual function type

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/58351


src/compiler/checker.ts
```diff
@@ -1923,32 +1923,42 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     }
 
     function runWithoutResolvedSignatureCaching<T>(node: Node | undefined, fn: () => T): T {
```
Andarist, line 1925: This becomes more and more complicated and error-prone. I'm definitely not a fan of this. It builds on top of the existing solution... so at least it shouldn't be significantly worse now. It's just that perhaps there is a completely better way to do it than the existing mechanism. I think that alternatives could be explored separately.

src/compiler/checker.ts
```diff
@@ -1923,32 +1923,42 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
-            for (const [symbolLinks, type] of cachedTypes) {
-                symbolLinks.type = type;
+        if (!node) {
+            return fn();
+        }
+        while (node) {
+            const nodeLinks = getNodeLinks(node);
+            cachedResolvedSignatures.push([nodeLinks, nodeLinks.resolvedSignature, nodeLinks.flags] as const);
+            nodeLinks.resolvedSignature = undefined;
+            nodeLinks.flags = NodeCheckFlags.None;
```
Andarist, line 1936: The key here is to reset `NodeCheckFlags.ContextChecked` so `contextuallyCheckFunctionExpressionOrObjectLiteralMethod` can correctly resolve (and cache) `signature.resolvedReturnType` multiple times (for requests with inference blocked and for requests without inference blocked). Otherwise, it has to be resolved later on... but without proper contextual information.

So I could just remove that bit and re-add it later if needed. I looked through all `NodeCheckFlags` and it might also be needed to do the same for `NodeCheckFlags.TypeChecked`. Although, I'm not 100% sure if that's really the case.

I just tried with a complete flags reset here and it does seem to work... but 🤷‍♂️ I'd probably be in favor of resetting solely what's really-really needed.

src/compiler/checker.ts
Andarist: I thought about using infra introduced in https://github.com/microsoft/TypeScript/pull/57421 but that PR is still in a draft. The problem space is also a little bit different - in this PR here we don't need to *revert* some recent work. In here, we need to temporarily evict some preexisting caches and restore them - we don't need to speculate per se. I assume that both are interested in the same lists of caches though.

Andarist:
@jakebailey could u build a playground for this one?

jakebailey:
@typescript-bot pack this

Andarist:
superseded by https://github.com/microsoft/TypeScript/pull/62184

PR #58388, opened 2024-05-01
Validate the options passed in are numbers instead of strings for our custom commandline options 

sheetalkamat:
cc: @RyanCavanaugh 

There are only 7 options that need to be checked so i think should be ok to check that irrespective of setting some symbol or something on our parsed option and not verify it.

### DIFF
--- a/src/compiler/program.ts
+++ b/src/compiler/program.ts
@@ -1553,7 +1553,9 @@ export function createProgram(rootNamesOrOptions: readonly string[] | CreateProg
     let { oldProgram } = createProgramOptions;
     for (const option of commandLineOptionOfCustomType) {
         if (hasProperty(options, option.name)) {
-            if (typeof options[option.name] === "string") throw new Error(`Found option: ${option.name} that is string and expected number/enum value.`);
+            if (typeof options[option.name] === "string") {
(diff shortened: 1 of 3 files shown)


src/compiler/program.ts
```diff
@@ -1550,6 +1551,11 @@ export function createProgram(rootNamesOrOptions: readonly string[] | CreateProg
     const createProgramOptions = isArray(rootNamesOrOptions) ? createCreateProgramOptions(rootNamesOrOptions, _options!, _host, _oldProgram, _configFileParsingDiagnostics) : rootNamesOrOptions; // TODO: GH#18217
     const { rootNames, options, configFileParsingDiagnostics, projectReferences, typeScriptVersion } = createProgramOptions;
     let { oldProgram } = createProgramOptions;
+    for (const option of commandLineOptionOfCustomType) {
+        if (hasProperty(options, option.name)) {
+            if (typeof options[option.name] === "string") throw new Error(`Found option: ${option.name} that is string and expected number/enum value.`);
```
RyanCavanaugh, line 1556:
```suggestion
            if (typeof options[option.name] === "string") {
                throw new Error(`${option.name} is a string value; tsconfig JSON must be parsed with zzzz before passing to createProgram`);
            }
```
Replace `zzzz` with whatever the right function is

sheetalkamat: we have so many ways thats why i didnt want to name the function name

jakebailey (approved):

andrewbranch (approved):

Issue #58391, opened 2024-05-01
Surprising circularity when parameter within base type expression refers to the class itself


Andarist:
### 🔎 Search Terms

circularity base expression class single call signature callback parameters

### 🕗 Version & Regression Information

- This is the behavior in every version I tried

### ⏯ Playground Link

https://www.typescriptlang.org/play/?ts=5.5.0-dev.20240501#code/JYOwLgpgTgZghgYwgAgApQmMBPAPAFQHsA+ZAbwChlkAKOALmSIEpGBnMKUAcwG4KAvhQqhIsRCgDKCABYQAtnFwBBUpWoY4AE0IgANtmQAHDFmwB+Rukw4VxfkJHho8JMgDCeuGzZ3yVZBAIAHdaVmRlB2EtCAQvDGQEXQ4PLx9GXEkIPRhiGgDgGPBgGGBodk4eABoKZmQAXlJcADEyvS02PIDS7I7GVt62Guo4EBBCMDgwYGTLZGk5RUzs3Jq6xtTvXwB5AFcwI33mqEJ5FraO4nthHCMUfGw7gj2D-dJ6-2oAfXwXw7BGL99v8HPwKLcUH8jiczvgBu1Og1PsgANoAaWQoGQAGsINhCDAmPCOgBdRgAIh+ULA5OQEAAHpAQB0cXiCUSLmx0SSAtRzBzBtyUZSga8aTzqNRGEEAG7QUHRWLxFAwXYgBDTXTIDhcEDcGjhB5PHU8a4UBlGQhQMCJNJsCJ0xkQZn2zxbOw0cnKcnMfLUMjIBjayp65ACYbI6gmGzYRh0dakABSkm2ADkAHQmvUlbDxiPh2rkARAA

### 💻 Code

```ts
interface Pretty<To> {
  (a: To): string;
}

interface Schema<A> {
  readonly pretty?: Pretty<A>;
}

interface Class<A> {
  new (): A;
}

declare const Class: <Self>(
  identifier: string,
) => <Fields>(
  fields: Fields,
  annotations?: Schema<Self>,
) => Class<OutputFrom<Fields>>;

type Type<TOutput> = {
  _TOutput: TOutput;
};

type OutputFrom<TFields> = {
  [K in keyof TFields]: "_TOutput" extends keyof TFields[K]
    ? TFields[K]["_TOutput"]
    : never;
};

declare function string(): Type<string>;

export class A extends Class<A>("A")(
[...]

PR #15104, opened 2017-04-10
Covariant checking for callback parameters

ahejlsberg:
In order to ensure that any generic type `Foo<T>` is at least co-variant with respect to `T` no matter how `Foo` uses `T`, TypeScript relates parameters bi-variantly (given that parameters are input positions, they naturally relate only contra-variantly). However, when source and target parameters both have function types with a single call signature, we known we are relating two callback parameters. In that case it is sufficient to only relate the parameters of the signatures co-variantly because, similar to return values, callback parameters are output positions. With this PR we introduce that change. This means that a `Promise<T>` or `Observable<T>`, where `T` is used only in callback parameter positions, will be co-variant (as opposed to bi-variant) with respect to `T`, which solves a commonly reported issue.

This change finds several issues in our real world code suites that all appear to be legitimate  inconsistencies.

Fixes #11022.
Fixes #14770.

PR #58392, opened 2024-05-01
Avoid getting single call signatures when parameter types are the same

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/58391

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -20769,10 +20769,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         for (let i = 0; i < paramCount; i++) {
             const sourceType = i === restIndex ? getRestOrAnyTypeAtPosition(source, i) : tryGetTypeAtPosition(source, i);
             const targetType = i === restIndex ? getRestOrAnyTypeAtPosition(target, i) : tryGetTypeAtPosition(target, i);
-            if (sourceType && targetType) {
-                if (sourceType === targetType) {
-                    continue;
-                }
+            if (sourceType && targetType && sourceType !== targetType) {
                 // In order to ensure that any generic type Foo<T> is at least co-variant with respect to T no matter
                 // how Foo uses T, we need to relate parameters bi-variantly (given that parameters are input positions,
                 // they naturally relate only contra-variantly). However, if the source and target parameters both have
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -20769,7 +20769,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         for (let i = 0; i < paramCount; i++) {
             const sourceType = i === restIndex ? getRestOrAnyTypeAtPosition(source, i) : tryGetTypeAtPosition(source, i);
             const targetType = i === restIndex ? getRestOrAnyTypeAtPosition(target, i) : tryGetTypeAtPosition(target, i);
-            if (sourceType && targetType && sourceType !== targetType) {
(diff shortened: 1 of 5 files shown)


src/compiler/checker.ts
```diff
@@ -20770,6 +20770,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             const sourceType = i === restIndex ? getRestOrAnyTypeAtPosition(source, i) : tryGetTypeAtPosition(source, i);
             const targetType = i === restIndex ? getRestOrAnyTypeAtPosition(target, i) : tryGetTypeAtPosition(target, i);
             if (sourceType && targetType) {
```
ahejlsberg, line 20772: Change to `sourceType && targetType && sourceType !== targetType` and get rid of the conditional `continue`.

ahejlsberg (approved):
Just curious, did you determine which part of the skipped logic is actually causing the circularity?

src/compiler/checker.ts
```diff
@@ -20770,6 +20770,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             const sourceType = i === restIndex ? getRestOrAnyTypeAtPosition(source, i) : tryGetTypeAtPosition(source, i);
             const targetType = i === restIndex ? getRestOrAnyTypeAtPosition(target, i) : tryGetTypeAtPosition(target, i);
             if (sourceType && targetType) {
+                if (sourceType === targetType) {
+                    continue;
+                }
```
Andarist, line 20775: I'm not super fond of this as a fix. It could be treated as a potential perf optimization~ - although I really doubt it would show anything in the perf runs, it's just an early bailout from this loop. It just happens to fix this particular circularity.

Similar bailouts are used in a couple of places in the code (even at the top of the containing function - `compareSignaturesRelated`). 

Ideally, I'd just like to recognize this type as one without call signatures so it could be typechecked using regular rules. At first, I thought that this code could just be simplified for the scenarios with `strictVariance` - after all, if we already have strict variance on why do we need to special case anything here? 

It turns out though that despite strict variance the callback parameters are still special - even if they are bivariant methods (!). This is somewhat surprising and I just learned about it today, while working on this but I guess it makes sense. Covariant type checking for callback parameters was introduced [here](https://github.com/microsoft/TypeScript/pull/15104). You can also see what changed when I tried to ignore this logic [here](https://github.com/microsoft/TypeScript/commit/4777d6eed461ae29d5e554df08b71ea709552e37)

The problem arises on `main` because both `getNonNullableType` and `getSingleCallSignature` resolve properties of the type and both are called to learn if the parameter is a callback parameter.

In this specific case, I could also experiment with:
- avoiding transitive `resolveStructuredTypeMembers` in `getTypeFactsWorker` based on `callerOnlyNeeds`. if the caller only needs to learn if the type is nullable or not... we don't need to check if it's a function type to compute those facts
- avoiding `resolveStructuredTypeMembers` in `getSingleSignature` when `SignatureKind.Call` is requested and the type's symbol has a `SymbolFlags.Class`. If the type is a class then it has a construct signature (and thus `getSingleCallSignature` can't return `true` in such a case) so the code could bail out early.

Some extra consideration for unions/intersections could be put into the above too.

Andarist: cc @ahejlsberg - i was elaborating on this down here when u asked the question about what exact code has caused this ;p

src/compiler/checker.ts
Andarist: done :) thanks for pointing out

mikearnaldi:
Any idea if this will land in 5.5?

jakebailey:
@typescript-bot test it

PR #58393, opened 2024-05-01
Remove unused SymbolLinks properties

JoostK:
I was going over some code over the weekend and hit various unused/unassigned `SymbolLinks` field declarations.

- The properties `outerTypeParameters`, `aliasSymbol` and `aliasTypeArguments` are stored on `Type`, not `SymbolLinks` (I'm guessing they used to be stored in `SymbolLinks`?)
- `exportsSomeValue` is entirely unused
- `enumKind` is used, but never assigned to so essentially all surrounding logic is dead.

Similar to #58377

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -778,7 +778,6 @@ import {
     JSDocAugmentsTag,
     JSDocCallbackTag,
     JSDocComment,
-    JSDocEnumTag,
     JSDocFunctionType,
     JSDocImplementsTag,
     JSDocImportTag,
@@ -1272,7 +1271,6 @@ const enum TypeSystemPropertyName {
     DeclaredType,
     ResolvedReturnType,
     ImmediateBaseConstraint,
-    EnumTagType,
     ResolvedTypeArguments,
     ResolvedBaseTypes,
     WriteType,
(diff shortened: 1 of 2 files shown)


JoostK:
Feel free to close if this introduces undesirable churn.

src/compiler/checker.ts
```diff
@@ -10486,8 +10484,6 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         switch (propertyName) {
             case TypeSystemPropertyName.Type:
                 return !!getSymbolLinks(target as Symbol).type;
-            case TypeSystemPropertyName.EnumTagType:
```
jakebailey, line 10489: Is this name also obsolete?

JoostK: Yes, the name has also been removed in the enum declaration.

jakebailey: If it has, you may not have pushed it, as I don't see that in the diff.

jakebailey: Oh oops, sorry, I though this was an import 
image

jakebailey (approved):
This seems reasonable to me, unless _not_ storing things here is an oversight and we shouldn't have moved them away.

src/compiler/checker.ts
JoostK: I had that _exact_ problem when going over the diff myself :)

PR #58396, opened 2024-05-01
Add synthetic TypeScriptSettings interface that exposes some compiler options to type system

### LINKED ISSUE #50196: Feature Request: Expose TS configuration and internals as Types
# Suggestion

## 🔍 Search Terms

expose CompilerOptions types 

## ✅ Viability Checklist

## ⭐ Suggestion

As library authors, we currently have to apply quite a number of hacks to support as many versions of TypeScript and as many different user configurations as possible.

Some possible requirements we face:

1. Slight behavioral changes between TS versions. For example, #49307 forced us to change a `{}` to `ACR[T]` - but only in TS versions above 4.8. The fix breaks in older versions.
2. Behavioral changes depending on `tsconfig.json` settings. Imagine a mapped type with functions.
  The user specifies `someMethod(action: PayloadAction<string>): void` and we map that over to a `someMethod(arg: string): void`. 
  In the case the user specifies `someMethod(action: PayloadAction<string | undefined>): void`, we map it over to an optional argument in the form `someMethod(arg?: string): void`
  Now assume the user has `strictNullChecks: false` in their tsconfig. Suddenly everything falls into the second case. 
3. A library might for example also behave differently if the user has `noUncheckedIndexedAccess: true` in their `tsconfig.json`

This leads to quite a few weird hacks in library types that might essentially break with every new release.

1. Could be solved with a `typesVersion`, but honestly: if we can avoid having two complete sets of types, we want to avoid that. We had it for a while for pre-4.1 types and post-4.1 types and maintaining it was a pain. Right now, I've published https://github.com/phryneas/ts-version/ which uses a monstrous `package.json` with typesVersions just to export the current TS Major and Minor.  
So we solve our problem with an additional dependency and 
```ts
import('@phryneas/ts-version').TSVersion.AtLeast<4, 8> extends true
      ? ACR[T]
      : {}
```
In the past we used hacks like
```ts
export type AtLeastTS35<True, False> = [True, False][IsUnknown<ReturnType<<T>() => T>, 0, 1>]
```
for that.
For a while we also had a subfolder with a `package.json` that would only use `typesVersions` on one import in that subfolder. At least we did not have to maintain two complete separate type definitions, but let's say it was not a great developer experience.
2. We check for `strictNullChecks` with 
```ts
type WithStrictNullChecks<True, False> = undefined extends boolean ? False : True;
```
3. We are not doing this yet but I'm sure we'd find a way to *somehow* detect `noUncheckedIndexedAccess` as well.

All that said: we have problems and we have solutions. But our solutions feel horrible and wrong. And also, having a package around with 120 different `typesVersions` entries just to detect the current TypeScript version can't really be good for performance.

So I want to suggest that TypeScript would add some global namespace that contains information about the current running setup:
* TS version (having some `AtLeast` type like I showcased above would be great too, but if we know the major and the minor we can also do that in userland)
* active `compilerOptions` flags - it would be amazing to have some string union `TypeScript.CompilerOptions.ActiveFlags` where we could just do 
```ts
'strictNullChecks' extends TypeScript.CompilerOptions.ActiveFlags ? SomeBehaviour : SomeLessSafeFallback
```
* the current `compilerOptions` in general - I don't know why someone would need access to `TypeScript.CompilerOptions.allowSyntheticDefaultImports` but it would feel like a waste not to expose that

## 📃 Motivating Example

```
type MessageFromTypeScript = AtLeast<[TypeScript.Version.Major, TypeScript.Version.Minor], [4, 8]> extends true ? "it's the future" : "you live in the past"

class MyRecord<T>{
  get(index: string): 'noUncheckedIndexedAccess' extends TypeScript.CompilerOptions.ActiveFlags ? (T | undefined) : T
}
```

## 💻 Use Cases

See "Suggestion" above.

rbuckton:
This adds a synthetic `TypeScriptSettings` interface to the global scope that reflects the state of _current compilation_'s compiler options. The synthetic interface looks something like the following:

```ts
interface TypeScriptSettings {
  version: string; // e.g., "5.5.0-dev", etc.
  versionMajorMinor: string; // e.g., "5.5", etc.
  locale: string | undefined; // e.g., "en-US", etc.
  target: string; // e.g., "es2023", etc.
  module: string; // e.g., "node16", etc.
  moduleResolution: string; // e.g., "node16", etc.
  customConditions: readonly string[] | undefined;
  exactOptionalPropertyTypes: boolean;
  noImplicitAny: boolean;
  noUncheckedIndexedAccess: boolean;
  strictBindCallApply: boolean;
  strictFunctionTypes: boolean;
  strictNullChecks: boolean;
  useDefineForClassFields: boolean;
}
```

Some compiler options can already be detected by the type system, such as `--strictNullChecks` or `--exactOptionalPropertyTypes`:

```ts
type StrictNullChecks = (1 | undefined extends 1 ? false : true);
type ExactOptionalPropertyTypes = { _?: true | undefined } extends { _?: true } ? false : true;
```
[TypeScript playground](https://www.typescriptlang.org/play/?exactOptionalPropertyTypes=true&ts=5.5.0-pr-58243-49#code/PTAEAEGcBcCcEsDG0ByBXANhgwgCwKaIDWkAXKHGvgLABQIE+AHgIbIDyADtPAPYB2LDAAVYvTvljQAngBVpEshVhU6dGRNABlOElSYcBYpFABeUAAoA2gEZQAH1Bp+AE3wAzeP3wuAuqGZofFcTW38AflB3IUh8UHJKfABKAG46BlBQAD1wtVoNOIBRVg5uPkERMQkpOQV8E3MAb1AAfXCElTjHZzdPbxdQAF8ApiCQ0Ga2jqoh0EjojFj45So0+jBMnKA)

But most compiler options are not so easily derived.

## Supporting stricter `Iterable`/`IterableIterator`

One of the key benefits of this approach is that types can be tailored for specific compiler options without the need to introduce new syntax or new `intrinsic` types. For example, we found that the changes in #58243, which introduces new type parameters for a stricter definition of `Iterable`, caused far too many breaks were we to ship those changes unflagged. When we shipped `--strictBindCallApply`, we were able to control strictness by swapping out the base type of a function or constructor type with something stricter than `Function`. Doing the same for all of the different JS built-ins that are iterable becomes far tricker, and is more likely to run afoul of users writing custom iterators, or would cause complications with the new `Iterator` constructor introduced by #58222.

Instead, we can leverage `TypeScriptSettings` to handle this case:

```ts
type BuiltinIteratorReturnType = TypeScriptSettings extends { noUncheckedIndexedAccess: true } ? undefined : any;

...

interface Array<T> {
  value(): IterableIterator<T, BuiltinIteratorReturnType>;
}
```

Here, we can use the existing `--noUncheckedIndexedAccess` flag to control the strictness of the `TReturn` type passed to `IterableIterator`. When the flag is unset, we fallback to the current behavior for `IterableIterator` (where `TReturn` is `any`). When the flag is set, we instead pass `undefined` for `TReturn`.

## Why use an interface and not an `intrinsic` type alias?

Rather than use `intrinsic`, the new `TypeScriptSettings` type is introduced as an `interface` to allow for declaration merging when a library needs to target both newer and older versions of TypeScript:

```ts
type __StrictNullAwareType<T> = ...;
type __NonStrictNullAwareType<T> = ...;
type __LegacyType<T> = ...;

export type NullAwareType<T> = 
  TypeScriptSettings extends { strictNullChecks: true } ? __StrictNullAwareType<T> :
  TypeScriptSettings extends { strictNullChecks: false} ? __NonStrictNullAwareType<T> :
  __LegacyType<T>;
  
// stub TypeScriptSettings to support older compilers:
declare global { interface TypeScriptSettings {} }
```

## TODO

While this PR is fully functional, we must still discuss which flags to include/exclude, as well as whether to continue with `TypeScriptSetings` or use a different name. I've only included options that could have an impact on types. Some options like `customConditions` have some interesting potential, but may end up being cut as they could be abused.

Fixes #50196
Related #58243

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -337,6 +337,9 @@ import {
     getNameFromIndexInfo,
     getNameOfDeclaration,
     getNameOfExpando,
+    getNameOfModuleKind,
+    getNameOfModuleResolutionKind,
+    getNameOfScriptTarget,
     getNamespaceDeclarationNode,
     getNewTargetContainer,
     getNonAugmentationDeclaration,
@@ -1089,6 +1092,8 @@ import {
     VariableLikeDeclaration,
     VariableStatement,
     VarianceFlags,
+    version,
+    versionMajorMinor,
     visitEachChild,
     visitNode,
     visitNodes,
@@ -1473,6 +1478,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     var compilerOptions = host.getCompilerOptions();
     var languageVersion = getEmitScriptTarget(compilerOptions);
     var moduleKind = getEmitModuleKind(compilerOptions);
+    var moduleResolutionKind = getEmitModuleResolutionKind(compilerOptions);
     var legacyDecorators = !!compilerOptions.experimentalDecorators;
     var useDefineForClassFields = getUseDefineForClassFields(compilerOptions);
     var emitStandardClassFields = getEmitStandardClassFields(compilerOptions);
@@ -1485,6 +1491,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     var noImplicitThis = getStrictOptionValue(compilerOptions, "noImplicitThis");
     var useUnknownInCatchVariables = getStrictOptionValue(compilerOptions, "useUnknownInCatchVariables");
     var exactOptionalPropertyTypes = compilerOptions.exactOptionalPropertyTypes;
+    var noUncheckedIndexedAccess = compilerOptions.noUncheckedIndexedAccess;
 
     var checkBinaryExpression = createCheckBinaryExpression();
     var emitResolver = createResolver();
@@ -2330,6 +2337,28 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         [".jsx", ".jsx"],
         [".json", ".json"],
     ];
+
+    var typeScriptSettingsSymbol = createSymbol(SymbolFlags.Interface, "TypeScriptSettings" as __String, CheckFlags.SyntheticInterface);
+    typeScriptSettingsSymbol.declarations = [];
+    typeScriptSettingsSymbol.members = createSymbolTable([
+        createProperty("version" as __String, getStringLiteralType(version), CheckFlags.Readonly),
+        createProperty("versionMajorMinor" as __String, getStringLiteralType(versionMajorMinor), CheckFlags.Readonly),
+        createProperty("locale" as __String, compilerOptions.locale ? getStringLiteralType(compilerOptions.locale) : undefinedType, CheckFlags.Readonly),
+        createProperty("target" as __String, getStringLiteralType(getNameOfScriptTarget(languageVersion) ?? ""), CheckFlags.Readonly),
+        createProperty("module" as __String, getStringLiteralType(getNameOfModuleKind(moduleKind) ?? ""), CheckFlags.Readonly),
+        createProperty("moduleResolution" as __String, getStringLiteralType(getNameOfModuleResolutionKind(moduleResolutionKind) ?? ""), CheckFlags.Readonly),
+        createProperty("customConditions" as __String, createTupleType(compilerOptions.customConditions?.map(cond => getStringLiteralType(cond)) ?? emptyArray, /*elementFlags*/ undefined, /*readonly*/ true), CheckFlags.Readonly),
+        createProperty("esModuleInterop" as __String, getESModuleInterop(compilerOptions) ? trueType : falseType, CheckFlags.Readonly),
+        createProperty("exactOptionalPropertyTypes" as __String, exactOptionalPropertyTypes ? trueType : falseType, CheckFlags.Readonly),
+        createProperty("noImplicitAny" as __String, noImplicitAny ? trueType : falseType, CheckFlags.Readonly),
+        createProperty("noUncheckedIndexedAccess" as __String, noUncheckedIndexedAccess ? trueType : falseType, CheckFlags.Readonly),
+        createProperty("strictBindCallApply" as __String, strictBindCallApply ? trueType : falseType, CheckFlags.Readonly),
+        createProperty("strictFunctionTypes" as __String, strictFunctionTypes ? trueType : falseType, CheckFlags.Readonly),
+        createProperty("strictNullChecks" as __String, strictNullChecks ? trueType : falseType, CheckFlags.Readonly),
+        createProperty("useDefineForClassFields" as __String, useDefineForClassFields ? trueType : falseType, CheckFlags.Readonly),
+    ]);
+    globals.set(typeScriptSettingsSymbol.escapedName, typeScriptSettingsSymbol);
+
     /* eslint-enable no-var */
 
     initializeTypeChecker();
@@ -2542,8 +2571,8 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         return symbol;
     }
 
-    function createProperty(name: __String, type: Type) {
-        const symbol = createSymbol(SymbolFlags.Property, name);
+    function createProperty(name: __String, type: Type, checkFlags?: CheckFlags) {
+        const symbol = createSymbol(SymbolFlags.Property, name, checkFlags);
         symbol.links.type = type;
         return symbol;
     }
@@ -12079,6 +12108,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
 
     // The outer type parameters are those defined by enclosing generic classes, methods, or functions.
     function getOuterTypeParametersOfClassOrInterface(symbol: Symbol): TypeParameter[] | undefined {
+        if (getCheckFlags(symbol) & CheckFlags.SyntheticInterface) {
+            return undefined;
+        }
         const declaration = (symbol.flags & SymbolFlags.Class || symbol.flags & SymbolFlags.Function)
             ? symbol.valueDeclaration
             : symbol.declarations?.find(decl => {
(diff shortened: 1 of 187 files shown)


rbuckton:
@typescript-bot: pack this

Andarist:
This would address (at least partially) https://github.com/microsoft/TypeScript/issues/50196 , cc @phryneas

phryneas:
This is great! ❤️

rbuckton:
> This would address (at least partially) #50196 , cc @phryneas

Per @RyanCavanaugh's [comment](https://github.com/microsoft/TypeScript/issues/50196#issuecomment-1206637336) in #50196, the biggest issue with future extensibility would be how people choose to depend on a type like this. It may be that the best we can hope for is that developers would code defensively against future changes, such as:

```ts
type Foo =
  TypeScriptSettings extends { noUncheckedIndexAccess: false } ? something :
  TypeScriptSettings extends { noUncheckedIndexAccess: true } ? somethingElse :
  never;
```

The other missing piece to #50196 is a way to conveniently perform range tests against `TypeScriptSettings["version"]`. It's possible to do with template literal types, but is by no means convenient:

```ts
import { type N, type A } from "ts-toolbelt";

type Eq<X extends number, Y extends number> = A.Is<X, Y, "equals">;

type VersionGreaterEq<A extends `${bigint}.${bigint}`, B extends `${bigint}.${bigint}`> =
    A extends `${infer AMajor extends number}.${infer AMinor extends number}` ?
        B extends `${infer BMajor extends number}.${infer BMinor extends number}` ?
            [Eq<AMajor, BMajor>, N.GreaterEq<AMinor, BMinor>] extends [1, 1] ? true :
            N.Greater<AMajor, BMajor> extends 1 ? true : false :
        false :
    false;

type TSVer = TypeScriptSettings["versionMajorMinor"];
//   ^? type V = "5.5"

type T1 = VersionGreaterEq<TSVer, "5.5">;
//   ^? type T1 = true

type T2 = VersionGreaterEq<TSVer, "5.6">;
//   ^? type T2 = false
```

[playground](https://www.typescriptlang.org/play/?ts=5.5.0-pr-58396-2#code/JYWwDg9gTgLgBAbzjAnmApnAcgGmWzAQTgF84AzKCEOAIhgGcBaGCCAGwCN12ZaBuALAAoEagxwAogEcAPAA046AB4x0AOwAmDOOoCuIblDwBNJao3bdBowD44AXjiEAdAEkGC03lrppegEN2BlpbIVFhcUwANXQoBmAIdQBxKHQAtSgZWWIVNS0dAAMAEgROYABzYHUYEhdS8qqakkK8ACFzfKsSssrq2vreptrC+wcROEnnTssi0uryOOcAWQCAK2gZgutDOLr59UWoFerNvNmdoxa4AH4JqYeO8+2ehaW21Y3j56t9Xah9gg3scPqdvhZtn8roVbvcHvC4ABtbKET7QdpoqC2PBYFypdKZFHLMEYsG2AC6WysiIAjHgaZSbsgoHpMAAuOEIqa4-EZOI5TEY9bQew-HQ027M1lwNkUIIMdmch7keWK4TK1XhMQEOAAFQAyrFjk5dQR9QBjKDAMAwfXoGAwaoVBiI2gANziCSSmOJ6mgtHJ4QA9EGpgA9O4RKJ6iVOI1elJpPlZOQGo0+ACsLgzoWDocmEZE2okuoATI44PHEomCXFsmm4pmXAA2XMiEPhm5AA)

rbuckton:
Another scenario I'd investigated recently was how a library could ensure that specific compiler options are set for it to be used correctly:

```ts
// @exactOptionalPropertyTypes: false

interface TypeScriptSettingsError<Message extends string> { error: Message }

type ExpectedSetting<K extends keyof TypeScriptSettings, V, Message extends string> =
    TypeScriptSettings extends { [P in K]: V } ? never :
    TypeScriptSettingsError<Message>;

type CheckSetting<_T extends never> = never;

type _ = CheckSetting<ExpectedSetting<"exactOptionalPropertyTypes", true, `To use this package you must set 'exactOptionalPropertyTypes' to 'true' in your tsconfig.json.`>>;
//                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~...
// error: Type 'TypeScriptSettingsError<"To use this package you must set 'exactOptionalPropertyTypes' to 'true' in your tsconfig.json.">' does not satisfy the constraint 'never'.
```

[TypeScript Playground](https://www.typescriptlang.org/play/?exactOptionalPropertyTypes=false&ts=5.5.0-pr-58396-2#code/PTAEAEFMA8EMGMAuB5ADoglgewHawDYAKATlqpMYgJ4AqV5AzgFygBmBDkAsAFC8Y5EFdvEig65AMrxiGdJMiJMOAOYNkANwqyAJpAagA3gF9+g4QjETI02fMXK1AUWKliAHgCy+hrBViYIRwdAwZEWVUAPiNQbSxiFm8GX39QUx5qclAAYQALSHgAawUlARV3AH0aWOggkNAcSC1iaIBeBqaKAG5eXkyxJ2hyJEgdEsd3AGkauoNCyCosVnF6Gxk5RHGyhgAaUAA1PaSUgNrIYNDwsrbeUDuVqXX7UtV1Zt19GfP6w1AAbUIoAEoEmAF0WAAiDAqHDxSAQtKgAD8HWaoCYt3u1lsGy2ry+FxiAKBOBB4IOiJRjTRGJ49weazsmwc2xcbi8Pj8kEiPT4GVWoAqoHaeQKxRZqncg2GQjGEvKmLuEJgCBQ6GweCIpHIlFoqwYEJ2itA4QArpAjXT7gADGhYUCmzgm3IYAyoBCFLmgRam0AAW0diFAnCDAHIVUg0JhcAQSGQKNRrAxQyb7aGzZAU8CfcQTQx4LhWNCAHQAKwYuGL4ntOldsAARvgxIgXQYAO6wYg4Mp7PRFxrO5sC0N6eD4TtiFT4LD1ggxARCYgiKyrHHPRxvbQYPQGX4RtXRzVxnWJ-WQ6Gw4jwxHGUPWyI83hAA)

Unfortunately, this check would defeated by `--skipLibCheck`, but could at least catch some incorrect usages of a library until such time as a more comprehensive mechanism for settings validation is adopted.

rbuckton:
We've opted not to pursue this mechanism at this time.

PR #58403, opened 2024-05-02
Disable constraint reduction in intersections created by constraint hoisting

### LINKED ISSUE #58370: Constraint in generic type forgotten in assignability check
### 🔎 Search Terms

generic type constraint not applied inside type

- This changed in commit or PR #57871

### ⏯ Playground Link

https://www.typescriptlang.org/play/?ts=5.4.5#code/C4TwDgpgBAggdiA0hEUC8U4FcC2AjCAJygB8oBnYQgSzgHNSKR8B7AGwG4AoL0SKAEoRgWQnAAq4CMhAsAZgB4A8ngBWUCAA9gEOABNyUFmogBjYAD50UFeq079h46rPAuUD1AD8UANoBrFHkbNQBdDW1dAz84CAA3IlD3TxSfWITCZJSPAC4oAG8-GShaKEDZORD1AENDShp6UIBaLzyACgBKdCtigF9feto6JJS89KJuXikoGQBJOQBlajo4apFCCCU5W1cFLKqIh2jnVwAafeL7KMN4JBRzlKF1iSkZQwwn0RfIGXllNQs5ysH2EX0kPxQhiujignW6MxQ3lhbThaB6KC60OinzE4OkkKRxTG8SIXWJGUmWjALEIwCgfGgQj0WFMED0AEZ-nZIjCTuZTgjUFibggZAKAGrVNhYCAC2xvQ7XMpBSq2azlYK2YEXRHCwXzJYrNaiTbbEzmLkCmQWfapQWKmHygkAemdsDYAHdqiBDFQZQA6W3ZDw+Qq+RAlOCC0J5SXS6C9IPB8lEJNQFOESauqAAdRp-kM1EqphYcHq1VodOohg2TVMAAszIE9JHyNQ9NAGYGqTS6QzBGyWWyAExch3Hc3AK26nnRW5iqBxmVytQKvUa1VqdUqqralKXWciu4gNM+A9HQxzRbLVbrU07C22acgG3B4NnmcXqoK7MwT3e31CADNM31DIpI2jWMpRlKBEzfeCMxAzxEIQzASUzLggA

### 💻 Code

```ts
type AnyKey = number | string | symbol;

type ReturnTypeKeyof<Obj extends object> = Obj extends object
    ? [keyof Obj] extends [never]
        ? never
        : { [Key in keyof Obj as string]-?: () => Key }[string]
    : never;

type KeyIfSignatureOfObject<
    Obj extends object,
    Key extends AnyKey,
    ReturnTypeKeys = ReturnTypeKeyof<Obj>,
> = ReturnTypeKeys extends () => Key ? ((() => Key) extends ReturnTypeKeys ? Key : never) : never;

export type Reduced<Obj extends object, Key extends AnyKey, Value, ObjKeys extends keyof Obj = keyof Obj> =
    Key extends KeyIfSignatureOfObject<Obj, Key>
        ? Key extends ObjKeys // Always true.
            ? { [K in Key]: Value }
            : never
        : never;
```

### 🙁 Actual behavior

Gives error `Type 'ObjKeys & KeyIfSignatureOfObject<Obj, Key, ReturnTypeKeyof<Obj>> & Key' is not assignable to type 'string | number | symbol'.`

### 🙂 Expected behavior

No error; `Key` was constrained to `AnyKey` and should be usable as a key.

I reduced this from a much larger example, so the code here doesn't make much sense (the names of the helper types might also not make sense anymore). It might be possible to reduce it further but I wasn't able to find anything obvious.

As the playground shows, simply wrapping the body of the type with a `Key extends AnyKey` check makes the error go away, which makes me think that this must be a bug (I guess it distributes the union, but the entire union should be valid as a key type). Checking `Key extends ObjKeys` *before* `KeyIfSignatureOfObject` also makes the error go away.

ahejlsberg:
Fixes #58370.

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -1104,6 +1104,7 @@ import {
     WideningContext,
     WithStatement,
     YieldExpression,
+    IntersectionFlags,
 } from "./_namespaces/ts";
 import * as moduleSpecifiers from "./_namespaces/ts.moduleSpecifiers";
 import * as performance from "./_namespaces/ts.performance";
@@ -14204,7 +14205,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 }
             }
             // The source types were normalized; ensure the result is normalized too.
-            return getNormalizedType(getIntersectionType(constraints), /*writing*/ false);
+            return getNormalizedType(getIntersectionType(constraints, IntersectionFlags.NoConstraintReduction), /*writing*/ false);
         }
         return undefined;
     }
@@ -17467,7 +17468,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     // a type alias of the form "type List<T> = T & { next: List<T> }" cannot be reduced during its declaration.
     // Also, unlike union types, the order of the constituent types is preserved in order that overload resolution
     // for intersections of types with signatures can be deterministic.
-    function getIntersectionType(types: readonly Type[], aliasSymbol?: Symbol, aliasTypeArguments?: readonly Type[], noSupertypeReduction?: boolean): Type {
+    function getIntersectionType(types: readonly Type[], flags = IntersectionFlags.None, aliasSymbol?: Symbol, aliasTypeArguments?: readonly Type[]): Type {
         const typeMembershipMap = new Map<string, Type>();
         const includes = addTypesToIntersection(typeMembershipMap, 0 as TypeFlags, types);
         const typeSet: Type[] = arrayFrom(typeMembershipMap.values());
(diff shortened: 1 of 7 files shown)


ahejlsberg:
@typescript-bot test it

RyanCavanaugh (approved):

src/compiler/checker.ts
```diff
@@ -17467,7 +17468,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     // a type alias of the form "type List<T> = T & { next: List<T> }" cannot be reduced during its declaration.
     // Also, unlike union types, the order of the constituent types is preserved in order that overload resolution
     // for intersections of types with signatures can be deterministic.
-    function getIntersectionType(types: readonly Type[], aliasSymbol?: Symbol, aliasTypeArguments?: readonly Type[], noSupertypeReduction?: boolean): Type {
+    function getIntersectionType(types: readonly Type[], flags = IntersectionFlags.None, aliasSymbol?: Symbol, aliasTypeArguments?: readonly Type[]): Type {
```
jakebailey, line 17471: Is this flag something that should be required, in case we forget to pass it? Happens all the time for the relation intersection state...

ahejlsberg: No, it's exceedingly rare to need any of these flags (only one place for each flag in entire code base), so I think it's preferable to leave it optional.

PR #58404, opened 2024-05-02
Ensure jsDocParsingMode is passed to LanguageService document registry

jakebailey:
Noticed this inconsistency when looking into an ESLint crash (but this wouldn't fix that).


sheetalkamat (approved):

jakebailey:
@typescript-bot pack this

PR #58409, opened 2024-05-02
Isolated declarations fix signature serialization scoping

### LINKED ISSUE #58407: Scope confusion with reused signatures
### 🔎 Search Terms

type aliasing isolated declarations 

- This changed between versions 5.4 and 5.5beta

### ⏯ Playground Link

[Playground](https://www.typescriptlang.org/dev/bug-workbench/?target=7&ts=5.4.5#code/PTAEAEGcBcCcEsDG0Bco4FcCmAoEEATLRAGwENYzp4B7AOzU1zzHADN4Ss6yBbLNGQB00SDiwAPAA41Y0UPDrQssNmURZQAQVABvUGTR0MvAEYrQAX3HTZ8xPRig2oALygAFBLRaAlG4A+UAkDSG0AbhwWCA4uHn40UxExeF4ZOT1nK2dYGl5QACIhYDICyMl0+UVlVXVNHX0aaAALFSMTc1grG0rQBzonAHM3Z3CgA)

### 💻 Code

```ts
// @strict: true
// @declaration: true

// @filename: a.ts
export interface A { a: number }
export const f = (x: A) => x as A;

// @filename: b.ts
import { f } from "./a";
export interface A { other: number }
export const g = f;
```

### 🙁 Actual behavior

`g` uses `A` as the parameter and return types

### 🙂 Expected behavior

`g` should use `A` from the module `a` instead

_No response_

### LINKED ISSUE #58408: Declaration references expanded type paramater
### 🔎 Search Terms

expanded parameters isolated declarations 

- This changed between versions 5.4 and 5.5 beta

### ⏯ Playground Link

[Playground Link](https://www.typescriptlang.org/play/?strict=false&allowJs=true&target=2&emitDeclarationOnly=true&ts=5.5.0-dev.20240502&ssl=10&ssc=1&pln=4&pc=1#code/PTAEAEBMFMGMBsCGAnRAXAlgewHYC5Q1kBXaAKBAjRQHNo0DoBnAJgAYBGAVjLOgA8ADlmRpQsXEzEA3DqAC8oABQA6NYgIBtfKABEOXQBpQGvYl0BdAJQEA3mVCPQwAFQvQyD8zGCUiALagLsAOTqZoAJ6C0FgAZiaGZAC+CgB8oPZOXmjEyDigOMTw8ACEyWRAA)

### 💻 Code

```ts
// @declaration: true
// @target: es2015

export const v1 = (...a: [n: "n", a: "a"]): {
    /** rest param */
    a: typeof a,
} => {
    return null!
}
```

### 🙁 Actual behavior

`typeof a` is emitted even though the expanded declaration no longer contains it

### 🙂 Expected behavior

`typeof a` should be inlined

_No response_

dragomirtitian:
Fixes #58408
Fixes #58407


tests/baselines/reference/assertionsAndNonReturningFunctions.types
```diff
@@ -55,7 +55,7 @@ function assertIsString(x) {
 */
 function assert2(check) {
 >assert2 : (check: boolean) => asserts check
->        : ^     ^^       ^^^^^             
+>        : ^     ^^       ^^^^^^^^^^^^^^^^^^
```
jakebailey, line 58: How did this one end up regressing?

tests/baselines/reference/assignmentCompatWithGenericCallSignaturesWithOptionalParameters.types
```diff
@@ -481,11 +481,11 @@ module GenericSignaturesInvalid {
 >b.a2 = t.a : () => T
 >           : ^^^^^^^
 >b.a2 : <T_1>(x?: T_1) => T_1
->     : ^   ^^ ^^^   ^^^^^^^^
+>     : ^^^^^^ ^^^^^^^^^^^^^^
```
jakebailey, line 484: Same here; tihs looks worse? Though, `T_1` wasn't in the original source, so I guess this is just a baseline fix?

tests/baselines/reference/asyncFunctionsAndStrictNullChecks.types
```diff
@@ -18,8 +18,8 @@ declare namespace Windows.Foundation {
...
->     : ^^^   ^^       ^^^                                 ^^     ^^^                             ^^        ^^^                       ^^^^^^^^^^^^^^^^^^^ ^^       ^^^                               ^^     ^^^                 ^^        ^^^                       ^^^           ^^^   ^^       ^^^                       ^^     ^^^                             ^^        ^^^                       ^^^^^^^^^^^^^^^^^^^   ^^       ^^^                       ^^     ^^^                   ^^        ^^^                       ^^^^^^^^^^^^^^^^^^^
+>then : { <U_1>(success?: ((value: TResult) => IPromise<U_1>) | undefined, error?: ((error: any) => IPromise<U_1>) | undefined, progress?: (progress: any) => void): IPromise<U_1>; <U>(success?: (value: TResult) => IPromise<U>, error?: (error: any) => U, progress?: (progress: any) => void): IPromise<U>; <U_1>(success?: ((value: TResult) => U_1) | undefined, error?: ((error: any) => IPromise<U_1>) | undefined, progress?: (progress: any) => void): IPromise<U_1>; <U_1>(success?: ((value: TResult) => U_1) | undefined, error?: ((error: any) => U_1) | undefined, progress?: (progress: any) => void): IPromise<U_1>; }
```
jakebailey, line 21: This also looks like a negative, given the original code did not have `| undefined`, right?

tests/baselines/reference/awaitedTypeJQuery.types
```diff
@@ -11,8 +11,8 @@ interface PromiseBase<TR, TJ, TN,
...
...gressFilter?: null): PromiseBase<ARD_1 | ARF_1, AJD_1 | AJF_1, AND_1 | ANF_1, BRD_1 | BRF_1, BJD_1 | BJF_1, BND_1 | BNF_1, CRD_1 | CRF_1, CJD_1 | CJF_1, CND_1 | CNF_1, RRD_1 | RRF_1, RJD_1 | RJF_1, RND_1 | RNF_1>; <ARF_1 = never, AJF_1 = never, ANF_1 = never, BRF_1 = never, BJF_1 = never, BNF_1 = never, CRF_1 = never, CJF_1 = never, CNF_1 = never, RRF_1 = never, RJF_1 = never, RNF_1 = never>(doneFilter: null, failFilter: (t: TJ, u: UJ, v: VJ, ...s: SJ[]) => ARF_1 | PromiseBase<ARF_1, AJF_1, ANF_1, BRF_1, BJF_1, BNF_1, CRF_1, CJF_1, CNF_1, RRF_1, RJF_1, RNF_1> | Thenable<ARF_1>, progressFilter?: null): PromiseBase<ARF_1, AJF_1, ANF_1, BRF_1, BJF_1, BNF_1, CRF_1, CJF_1, CNF_1, RRF_1, RJF_1, RNF_1>; <ARD_1 = never, AJD_1 = never, AND_1 = never, BRD_1 = never, BJD_1 = never, BND_1 = never, CRD_1 = never, CJD_1 = never, CND_1 = never, RRD_1 = never, RJD_1 = never, RND_1 = never>(doneFilter: (t: TR, u: UR, v: VR, ...s: SR[]) => ARD_1 | PromiseBase<ARD_1, AJD_1, AND_1, BRD_1, BJD_1, BND_1, CRD_1, CJD_1, CND_1, RRD_1, RJD_1, RND_1> | Thenable<ARD_1>, failFilter?: null, progressFilter?: null): PromiseBase<ARD_1, AJD_1, AND_1, BRD_1, BJD_1, BND_1, CRD_1, CJD_1, CND_1, RRD_1, RJD_1, RND_1>; }
```
jakebailey, line 14: This also looks like a regression (way off to the right)

src/compiler/checker.ts
```diff
@@ -7304,17 +7310,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
-            // JSDoc, which are always outside the function declaration, so are not in the parameter scope.
-            let cleanup: (() => void) | undefined;
-            if (
-                context.enclosingDeclaration
-                && declaration
-                && declaration !== context.enclosingDeclaration
-                && !isInJSFile(declaration)
-                && (some(expandedParams) || some(typeParameters))
-            ) {
+            let cleanupParams: (() => void) | undefined;
```
jakebailey, line 7313: I guess we're not skipping this in JS anymore? That was a perf optimization, so will have to try and see if webpack regresses...

tests/baselines/reference/bluebirdStaticThis.types
```diff
@@ -27,7 +27,7 @@ export declare class Promise<R> implements Promise.Thenable<R> {
 
     static try<R>(dit: typeof Promise, fn: () => Promise.Thenable<R>, args?: any[], ctx?: any): Promise<R>;
 >try : { <R>(dit: typeof Promise, fn: () => Promise.Thenable<R>, args?: any[], ctx?: any): Promise<R>; <R_1>(dit: typeof Promise, fn: () => R_1, args?: any[], ctx?: any): Promise<R_1>; }
->    : ^^^ ^^   ^^              ^^  ^^                         ^^    ^^^     ^^   ^^^   ^^^          ^^^   ^^   ^^              ^^  ^^         ^^    ^^^     ^^   ^^^   ^^^^^^^^^^^^^^^^^^
+>    : ^^^ ^^   ^^              ^^  ^^                         ^^    ^^^     ^^   ^^^   ^^^          ^^^^^^^^   ^^              ^^  ^^^^^^^^^^^^^    ^^^     ^^   ^^^   ^^^^^^^^^^^^^^^^^^
```
jakebailey, line 30: Another case where reuse is seemingly lost but probably wrong before (I'll stop commenting on this class of change).

jakebailey (commented):
I stopped reviewing at the letter C (there's a lot changing, of course)

jakebailey:

tests/baselines/reference/visibilityOfCrossModuleTypeUsage.types
```diff
@@ -18,16 +18,16 @@ function run(configuration: commands.IConfiguration) {
 >                      : ^^^^^^
 >configuration.workspace.toAbsolutePath(configuration.server) : string
 >                                                             : ^^^^^^
->configuration.workspace.toAbsolutePath : (server: IServer, workspaceRelativePath?: string) => string
->                                       : ^      ^^       ^^                     ^^^      ^^^^^^^^^^^
+>configuration.workspace.toAbsolutePath : (server: import("visibilityOfCrossModuleTypeUsage_server").IServer, workspaceRelativePath?: string) => string
```
jakebailey, line 21: Huh, interesting; I guess this was a bug before?

src/compiler/checker.ts
dragomirtitian: I thought part of the point of this was so that JS declaration emit can also benefit more from types written in source. Without proper scoping it's hard to do any reuse in JS files.

tests/baselines/reference/assertionsAndNonReturningFunctions.types
dragomirtitian: This is a very interesting scoping issue. In js doc param and return tags, when you use a symbol you are actually referring to the symbol in the outer scope, even if there is a symbol in the parameter list that matches. This is fine in most cases, except for type predicates, where a symbol should refer to the parameter. Here is just some of the strangeness I found:


```ts
// @checkJs: true
const check =  "outer"
/**
 * @param {typeof check} check // this typeof check reffers to the outer variable
 * @returns {asserts check} /// this should reffer to the parameter? Although it's a bit confusing
*/
function assert2(check) {
    // Declaration is wrong here in 5.4
    // function assert2(check: typeof check): asserts check;
    if (!check) throw new Error();
}

/**
 * @param {typeof check} check
 * @returns {asserts check is typeof check} // So the check in assert should reffer to the parameter but the check in typeof should reffer to the outer variable?
*/
function assert3(check) {
    if (!check) throw new Error();
}

var c;
/**
 * @param {typeof check} check // this typeof check reffers to the outer variable
 * @returns {asserts c} /// You can assert an outer variable too! No error and is pulls it into the d.ts
*/
function assert4(check) {
    // Declaration is wrong here in 5.4
    // function assert2(check: typeof check): asserts check;
    if (!check) throw new Error();
}

assert4("outer")

```


I think a lot of this previously worked because the type printer took it on faith that the symbols in the return type annotation were checked before and were safe to use. This just points out the issue. I already have a follow on PR to fix this and bring back the reuse for the assert clause in this case, but I didn't want to pile on too much in this one.

tests/baselines/reference/visibilityOfCrossModuleTypeUsage.types
dragomirtitian: `trackExistingEntityName` would previously check if the symbol could be resolved in the original context and if the symbol is accessible from the enclosing declaration. The answer to both of those is yes. The issue is that the symbol is not accessible in the *same way*. This can lead to the scope confusion I showed in one of the bugs. 

What I do now is I test if the symbol is also resolvable in the context it is being used in. And in this case `IServer` is not directly accessible so it falls back on `typeToTypeNodeHelper` which will print the import. 

I have a follow on PR that will try to find a more appropriate name in the enclosing declaration where this is printed (such as `commands.IServer` in this case) and will also try to use more of the original node even if sometimes this means replacing parts of the node with nodes printed from types.

src/compiler/checker.ts
jakebailey: Both, my original code actually made scopes a whole lot less, so it was meaningful then. But this fake scope thing has been expanded out far more than I ever expected and you've definitely shown cases where my original assumptions were wrong! Perf doesn't seem affected, thankfully.

tests/baselines/reference/assertionsAndNonReturningFunctions.types
jakebailey: Wow, great example. Definitely a case I didn't think about when I introduced fake scopes.

jakebailey:

tests/baselines/reference/assignmentCompatWithGenericCallSignaturesWithOptionalParameters.types
dragomirtitian: I found the issue. I was looking up the original parameter name `T` to make sure it is available in the scope the signature is used in. The problem is the node was already renamed to prevent shadowing (so I was looking for `T` instead of `T_1`). This caused the whole signature to be thrown asway and fall back to regular type printing.

jakebailey: I think I'm half confused because if it's generated, doesn't that mean none of that should be reused? I thought the point of the underlining is to show what does and doesn't come from the original file?

dragomirtitian: I guess it depends what you call reuse. Renaming type parameters to avoid shadowing is still considered reuse. It does use the same structure as the original source just with a different parameter name. That logic was already there maybe @weswigham  has some more thoughts.

src/compiler/checker.ts
```diff
@@ -8111,20 +8144,22 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
-            }
-            if (initial.typeParameterNamesByTextNextNameCount) {
-                initial.typeParameterNamesByTextNextNameCount = new Map(initial.typeParameterNamesByTextNextNameCount);
-            }
-            initial.tracker = new SymbolTrackerImpl(initial, initial.tracker.inner, initial.tracker.moduleResolverHost);
-            return initial;
+            const oldMustCreateTypeParameterSymbolList = context.mustCreateTypeParameterSymbolList;
+            const oldMustCreateTypeParametersNamesLookups = context.mustCreateTypeParametersNamesLookups;
+            context.mustCreateTypeParameterSymbolList = true;
+            context.mustCreateTypeParametersNamesLookups = true;
```
jakebailey, line 8150: Is all of this `mustCreate*` stuff an optimization? The logic behind it isn't super clear to me.

tests/baselines/reference/assignmentCompatWithGenericCallSignaturesWithOptionalParameters.types
jakebailey: After Wesley's explanation in the ID meeting this morning, this seems fine if the definition of reuse is not as strict as I expected.

src/compiler/checker.ts
dragomirtitian: Yes. It's an optimization. It's basically copy on write. When creating a new context for a signature it's not always necessary to also duplicate the lookups. This defers the copy until there is actually a reason copy (until there is mutation of the map). In both of these cases (mustCreateTypeParameterSymbolList  and mustCreateTypeParametersNamesLookups) the mutation of the map only occurs in one place so it's easy enough to enforce this. 

The optimizations in the last comit avoids duplicating the context, the symbol tracker as well as the type parameters lookup maps (if the signature is not generic).

src/compiler/checker.ts
```diff
@@ -8266,18 +8301,38 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+                introducesError = true;
+                return { introducesError, node, sym };
+            }
+            if (
+                context.enclosingDeclaration &&
+                (getNodeLinks(context.enclosingDeclaration).fakeScopeForSignatureDeclaration || !findAncestor(node, n => n === context.enclosingDeclaration)) &&
+                !(sym && sym.flags & SymbolFlags.TypeParameter)
+            ) {
+                const symAtLocation = resolveEntityName(leftmost, meaning, /*ignoreErrors*/ true, /*dontResolveAlias*/ true, context.enclosingDeclaration);
+                if ((symAtLocation !== sym && (!symAtLocation || !isTransientSymbol(symAtLocation) || symAtLocation.links.target !== sym)) || symAtLocation === unknownSymbol) {
```
weswigham, line 8318: The later half of this condition needs some explanation - both here and in a comment, because I'm not sure offhand why we're looking at the cached target (and why that's safe) and the transient-ness of the symbol to determine if the resolution is an error. I'd think any symbol that isn't "the same" (eg, `getSymbolIfSameReference`) would be an error.

Also: If these conditions are met, isn't this the only lookup we need to do? Doesn't matter if it resolves at the original location if it resolves in the new one, no? Not sure how that could happen, but we could cut down on name resolutions we don't need to do.

weswigham (commented):
I like it - injecting new `unknownSymbol` refs into the synthetic scope to block lookups we know won't work due to the transform we've done to the parameters is good.

src/compiler/utilities.ts
```diff
@@ -11068,7 +11068,8 @@ export function createNameResolver({
                         if (meaning & result.flags & SymbolFlags.Type && lastLocation.kind !== SyntaxKind.JSDoc) {
                             useResult = result.flags & SymbolFlags.TypeParameter
                                 // type parameters are visible in parameter list, return type and type parameter list
-                                ? lastLocation === (location as FunctionLikeDeclaration).type ||
+                                ? nodeIsSynthesized(lastLocation) || // Synthetic fake scopes are added for signatures so type parameters are accessible from them
```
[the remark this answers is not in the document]
dragomirtitian: I will switch over to `node.flags & NodeFlags.Synthesized` the actual intent is more `isFakeScopeNode`, but that would mea going through `getNodeLinks(node).fakeScopeForSignatureDeclaration`. Not sure if it's worth it or if the fact that the node is synthetic is a good enough proxy.

src/compiler/checker.ts
dragomirtitian: The symbol resolved in the current enclosing scope might be different from the declaration scope since we only really look at the name. In https://github.com/microsoft/TypeScript/issues/58407 we see how this could happen even today when we are a lot more conservative about what declarations we transplant at a new location.


The `isTransientSymbol` is indeed superfluous. I removed it and used `getSymbolIfSameReference` instead to test for symbol sameness. That revealed some instances where the symbol was indeed the same that I was missing before. Thank you. 

I was wondering, `getSymbolIfSameReference` does not check `s1 === s2` as a fast path. would that make sense?  I am assuming most of the symbols will be resolved so probably there is no actual lookup happening, just looking at cached values.

src/compiler/utilitiesPublic.ts
```diff
@@ -2320,7 +2320,8 @@ function isDeclarationKind(kind: SyntaxKind) {
         || kind === SyntaxKind.VariableDeclaration
         || kind === SyntaxKind.JSDocTypedefTag
         || kind === SyntaxKind.JSDocCallbackTag
-        || kind === SyntaxKind.JSDocPropertyTag;
+        || kind === SyntaxKind.JSDocPropertyTag
+        || kind === SyntaxKind.NamedTupleMember;
```
dragomirtitian, line 2324: `NamedTupleMember` seemed to be  missing here. This meant that tuple member names were being funneled through `trackExistingEntityName` even when other declaration names were not. 

```ts
                if (isEntityName(node) || isEntityNameExpression(node)) {
                    if (isDeclarationName(node)) {
                        return node;
                    }
                    const { introducesError, node: result } = trackExistingEntityName(node, context);
                    hadError = hadError || introducesError;
                    // We should not go to child nodes of the entity name, they will not be accessible
                    return result;
                }
```

src/compiler/checker.ts
```diff
@@ -8318,7 +8319,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     // If the symbol is not found, but was not found in the original scope either we probably have an error, don't reuse the node
                     (symAtLocation === undefined && sym !== undefined) ||
                     // If the symbol is found both in declaration scope and in current scope then it shoudl point to the same reference
-                    (symAtLocation && sym && !getSymbolIfSameReference(symAtLocation, sym))
+                    (symAtLocation && sym && !getSymbolIfSameReference(symAtLocation?.exportSymbol ?? symAtLocation, sym))
```
jakebailey, line 8322: `getSymbolIfSameReference` should already be doing the whole "get export symbol if present" thing, I believe.

src/compiler/checker.ts
```diff
@@ -8306,6 +8306,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 (getNodeLinks(context.enclosingDeclaration).fakeScopeForSignatureDeclaration || !findAncestor(node, n => n === context.enclosingDeclaration)) &&
                 !(sym && sym.flags & SymbolFlags.TypeParameter)
             ) {
+                sym = sym?.exportSymbol ?? sym;
```
jakebailey, line 8309: Should this be `getExportSymbolOfValueSymbolIfExported`?

src/compiler/checker.ts
dragomirtitian: It does merge and resolution of aliases as far as I can tell. I did not see anything about exports, and in my tests if one symbol resolved to the export and the other to the local then it returns false.

src/compiler/checker.ts
dragomirtitian: I believe that only works for value symbols not for type symbols since it checks `(symbol.flags & SymbolFlags.ExportValue)` and in the binder only value symbols get this flag:

```ts

const exportKind = symbolFlags & SymbolFlags.Value ? SymbolFlags.ExportValue : 0;
                const local = declareSymbol(container.locals, /*parent*/ undefined, node, exportKind, symbolExcludes);
                local.exportSymbol = declareSymbol(container.symbol.exports!, container.symbol, node, symbolFlags, symbolExcludes);

```

Or am I misunderstanding this?

jakebailey: I swapped  this line for `sym = getExportSymbolOfValueSymbolIfExported(sym);` and nothing happened.

Truthfully I don't really know what I'm doing here, other than remembering past bugs where we forgot to call this function 😄

dragomirtitian: Ah, I did some more debugging and understood it. The local symbol for types actually don't have any flags, so when you look them up you always get the export symbol. So there is never any need to look for the export symbol of a type. 

Made the change.

src/compiler/checker.ts
dragomirtitian: I checked again `getSymbolIfSameReference` does not take export symbols into account.

src/compiler/checker.ts
dragomirtitian: I added the explanation

jakebailey (approved):

Issue #41232, opened 2020-10-25
`asserts` keyword does not enhance hoisting function type, works on functions assigned to a variable


phryneas:
**TypeScript Version:**  4.0.2, 4.1.0-beta

**Search Terms:**
asserts function 

**Code**

```ts
interface LabelledFunction { label: string; }

function assignLabel(fn: () => unknown, label: string): asserts fn is LabelledFunction {
  Object.assign(fn, { label });
}

function a() { }
assignLabel(a, 'a');
console.info(a.label); // Property 'label' does not exist on type '() => void'.(2339)

const b = function () { };
assignLabel(b, 'b');
console.info(b.label);

const c = () => { };
assignLabel(c, 'c');
console.info(c.label);
```

**Expected behavior:**

At least in runtime code coming *after* `assignLabel(a, 'a');`, `a.label` should be accessible.

**Actual behavior:**

> Property 'label' does not exist on type '() => void'.(2339)

**Playground Link:** [link](https://www.typescriptlang.org/play?ts=4.1.0-beta#code/JYOwLgpgTgZghgYwgAgDJwEYQDbYgEwDEBXEBMYAexGQG9ltMcAuZAZzClAHMBuZAL4AoITFLkqNOGzbBuIdFmwAKGCFbKAlMgC8APmSkA1iEoB3EABoGTbKw5cQ3Ta2ltoYNsjXJgXxTh4ROIU1HRCyMgA8hgAVhDkAHRuciCqVnQ2SoKavELCoiGSyHBamcIp8gEqcNYA5HB1uUII1GyUeImgMJTKcImMSrnIAPQjyAAKUJQADh4Ansh1gzh1yPiUEF6mYMgQAB5+u2Fg83NLZfrIAG6UwPh1icoATADMrwCcmiKtIBzIGF03iKYTK9AEeUqClsygw9QwTTyv3anW6vQwA1szRabV2CCBlwM4MhMlS1WUCHqCEROL+HQgXRAPQpmKGvCAA)

robbiespeed:
Same thing happens to classes:
```ts
interface Labelled {
  label: string;
}

function assignLabel <T> (value: T, label: string): asserts value is (T & Labelled) {
  Object.assign(value, { label });
}

function a() { }
assignLabel(a, 'a');
console.info(a.label); // Property 'label' does not exist on type '() => void'.(2339)

const b = function b () { };
assignLabel(b, 'b');
console.info(b.label); // Works

class X {}
[...]

PR #58411, opened 2024-05-02
Compute flow types for function declarations

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/41232

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -29393,9 +29393,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
 
         const isAlias = localOrExportSymbol.flags & SymbolFlags.Alias;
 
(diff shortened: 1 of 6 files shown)


src/compiler/checker.ts
```diff
@@ -29393,9 +29393,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
 
         const isAlias = localOrExportSymbol.flags & SymbolFlags.Alias;
 
-        // We only narrow variables and parameters occurring in a non-assignment position. For all other
+        // We only narrow variables, parameters and functions occurring in a non-assignment position. For all other
         // entities we simply return the declared type.
```
Andarist, line 29397: tbf... if this is something the plan would like to pull in, we probably should enable narrowing on just anything. `asserts` make it possible to expand all non-nullable types

PR #58418, opened 2024-05-02
Fix name generation scoping

### LINKED ISSUE #57897: A shadowed variable is generated in a an async function with a default parameter declared as a static property of a class when targeting ES5.

rbuckton:
There are a number of small oddities in the emitter regarding name generation for temp variables and private names. Despite using `reservedNames`/`reservedPrivateNames` to reserve names permanently within a scope, we weren't checking all reservations in the respective stacks. We also weren't consistent about name generation in a few places, so I've combined the private and temp variable name scopes so that we enter and exit both consistently.

I also removed a place where we exited and then reentered a private name generation scope when visiting a computed property name since both temp variables and private names should be generated immediately as we enter a declaration, not on-demand as we descend into and out of computed property names.

Fixes #57897


DanielRosenwasser:

src/compiler/emitter.ts
```diff
@@ -5372,7 +5349,30 @@ export function createPrinter(printerOptions: PrinterOptions = {}, handlers: Pri
...
+            stack = reservedNamesStack;
+        }
+
+        if (set?.has(name)) {
+            return true;
+        }
+        for (let i = stack.length - 1; i >= 0; i--) {
+            if (set === stack[i]) {
+                continue;
+            }
```
DanielRosenwasser, line 5369: This is partially off-topic, but is there any reason the current scope can't be stored separate of the stack itself? Or if it is possible, when is it the case that the current is not the same as the last in the stack?

rbuckton: The current scope is stored separate from the stack. The way push/pop works is that we don't reset the current scope until we actually record something so that we defer the cost of reassignment until the moment we actually need to record a name. This check here is an example of that. If the set for the current scope is also in the stack, then we can skip it as nothing else was actually recorded in that scope.

rbuckton: In other words, the algorithm favors code that requires no generated names so that we can avoid paying the cost for more recent `--target`s.

andrewbranch (approved):

PhoebeSzmucer:
@rbuckton @andrewbranch this seems to be a regression in 5.4.x, and I've unfortunately hit this after upgrading a large codebase to 5.4.x. Is there any chance we can backport this fix, so that I don't have to revert the upgrade? I see the fix is on 5.5.x but that's not released yet.

jakebailey:
TS 5.5 is coming out this week, so I wouldn't think we'd be patching 5.4 anymore.

PhoebeSzmucer:
We have 5.5+ million lines of typescript code so it will probably take us a couple of weeks to get to 5.5 - if there is any chance this can be backported it would be a huge help.

No worries if not though!

PR #58419, opened 2024-05-02
Make TypeScript itself ESM-only, made possible by require(ESM)

jakebailey:
https://github.com/nodejs/node/pull/51977 (behind `--experimental-require-module` in Node 22) enables Node to require ESM so long as that ESM does not make use of top-level await. TypeScript would only need top-level await to construct `ts.sys` at startup, which must be defined when the environment is detected to be Node. Via https://github.com/nodejs/node/issues/52599 and https://github.com/nodejs/node/pull/51977, we can synchronously access Node's built-ins, and therefore can offer up a TLA-free public API, enabling TypeScript to ship as ESM-only without breaking CJS users, once `require(ESM)` is unflagged.

TODO:

- [ ] Fix public API dts
- [ ] Figure out `__esModule`
	- The real question is the context of `require(ESM)` having `__esModule`. But I suspect that exporting `default` fixes this.
- [x] Consider looking for `require` if `process.getBuiltinModule` is not found
- [ ] Export maps? Actually make the package importable? (does this work?)
- [x] ~Maybe still ship CJS behind a condition to get this out sooner?~
	- ~Dual package hazard is very scary here. Maybe acceptable with `module` condition/top-level prop?~
- [x] ~Stop bundling? Share code between tsc.js/typescript.js?~
- [ ] Is this even a good idea before we figure out a better public API than our current huge omnibus?
- [x] `@types/diff` is typed wrong

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,21 +29,15 @@ jobs:
           - windows-latest
           - macos-latest
         node-version:
+          - '23'
           - '22'
           - '20'
-          - '18'
-          - '16'
-          - '14'
         bundle:
           - 'true'
         include:
           - node-version: 'lts/*'
             bundle: false
             os: ubuntu-latest
-        exclude:
-          # No Node 14 on ARM macOS
-          - node-version: '14'
-            os: macos-latest
 
     runs-on: ${{ matrix.os }}
     name: Test Node ${{ matrix.node-version }} on ${{ matrix.os }}${{ (!matrix.bundle && ' with --no-bundle') || '' }}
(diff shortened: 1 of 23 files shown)


jakebailey:
This has been rebased now that we're using extensions on main, condensing this down to a pretty small change (though there's still stuff not working).

jakebailey:
There are still a bunch of TODOs, but this is actually green! Wow!

jakebailey:
https://github.com/nodejs/node/pull/52762 was merged a bit ago; I should be able to retarget this PR at nightly.

jakebailey:
@typescript-bot perf test this faster hosts=node@23.0.0

jakebailey:
Huh, where is the perf hit coming from here...?

jakebailey:
Given the Go port, I don't think it makes sense for me to attempt this anymore. I think it's only going to cause more churn in the meantime.

PR #58421, opened 2024-05-02
Use explicit extensions for imports within src

jakebailey:
This is extracted from #58419.

~Doing this now would reduce churn, but I'm not 100% certain how we can avoid adding the "wrong" kind of imports in the future before `"type": "module"`?~ Lint rule!


andrewbranch (approved):

andrewbranch:
Something weird with the unused baselines check?

jakebailey:
Weird, will take a look later.

jakebailey:
@andrewbranch Is there a way to force this to always be required, even though we're currently being treated as CommonJS?

andrewbranch:
Not in tsc. You can set `"importModuleSpecifierEnding": "js"` in VS Code preferences for auto imports, and otherwise use a lint rule to enforce.

andrewbranch:
I know eslint-plugin-import has a rule for that; not sure about others.

scripts/eslint/rules/js-extensions.cjs
```diff
@@ -0,0 +1,55 @@
...
+                    && typeof moduleReference.expression.value === "string"
+                ) {
+                    source = moduleReference.expression;
+                }
+            }
+            else {
+                source = node.source;
+            }
+
+            if (source?.value.startsWith(".") && !/\.[cm]?js$/.test(source.value)) {
```
andrewbranch, line 38: Technically this could be failing on a directory module rather than an extensionless filename, but we don't do that anywhere in our codebase, so I'm fine with not checking. Though maybe the rule should have a `DO NOT COPY THIS` comment similar to the declaration bundler.

andrewbranch (approved):

scripts/eslint/rules/js-extensions.cjs
jakebailey: Does that happen even for relative paths? I had figured the startsWith would be enough.

(Trying to not pull in a huge plugin, and a message saying "don't use this would be good yeah)

andrewbranch: Yes, `import "./foo"` could be referencing a directory with a package.json or just an index file.

jakebailey: Ah, I forgot about both cases.

jakebailey:
Sanity checked, and reverting all changes in src, running `hereby lint --fix` and `dprint` nets the right result, missing only the manual changes for using `export *` for the unit test reexport file (no runtime difference, but it makes the imports checked), and then two unit test loading hacks that are in place due to our goofy test runner.

PR #58428, opened 2024-05-03
Declaration emit should retain (instead of elide) unresolved computed names

### LINKED ISSUE #58423: [transpileDeclaration API][5.5] Type containing enum values is incorrectly emitted
### 🔎 Search Terms

transpile declaration api emit enum missing

### ⏯ Playground Link

https://github.com/MichaelMitchell-at/transpile_declaration_bug_repro

### 💻 Code

```ts repro
// @strict: true
// @declaration: true
// @showEmit
// @showEmittedFile: a.d.ts
// @showEmittedFile: b.d.ts

// @filename: a.ts
export enum Foo {
  A = 'a',
  B = 'b',
}

// @filename: b.ts
import {Foo} from './a';

export type Bar = {
  [Foo.A]: 1;
  [Foo.B]: 2;
}

```

[Workbench Repro](https://www.typescriptlang.org/dev/bug-workbench/?target=7&ts=5.5.0-dev.20240502#code/PTAEAEGcBcCcEsDG0Bco4FcCmAoEEATLRAGwENYzp4B7AOzU13ygAsaB3AUQFt5o8YNp179oWAgDF4JLGjIA6AguiRBESO259o4qTLmgARkpVr14AGYG6ZHocWqcWAB4AHGrGigsdDD1BJGhpQAG8cUFAAQVAAXlAAcjIEgBoI0AAhOMSjVJwAXxwLa1lbezQTJ3geDy8woJp80EtYGgCEhWBkgG4i11rvaABPNyxMimzwyIBtBoUogF00AEZembmMpdAAJl7CoA)

### 🙁 Actual behavior

Refer to this program which invokes the `ts.transpileDeclaration` API:
https://github.com/MichaelMitchell-at/transpile_declaration_bug_repro/blob/main/transpile.mjs

Basic instructions to invoke it can be found in
https://github.com/MichaelMitchell-at/transpile_declaration_bug_repro/blob/main/README.md

```ts
export type Bar = {};
```
```ts
import { Foo } from './a';
export type Bar = {
    [Foo.A]: 1;
    [Foo.B]: 2;
};
```

### 🙂 Expected behavior

Using the API should emit
```ts
import { Foo } from './a';
export type Bar = {
    [Foo.A]: 1;
    [Foo.B]: 2;
};
```

It's entirely possible that usage of the API is incorrect, though it doesn't seem to deviate from the patterns demonstrated in https://github.com/Microsoft/TypeScript/wiki/Using-the-Compiler-API

weswigham:
Historically, if a `[name]: whatever` didn't resolve, it was elided in a declaration file, even when you wrote it explicitly in a type. This PR changes that behavior to instead copy the name into the output even when it doesn't resolve. Since the input has a checker error in this case, our declaration output here is flexible, and this makes us align with expectation under `isolatedDeclarations`.

Fixes #58423


MichaelMitchell-at:
I found another issue and am not sure if it will be fixed by this or not, so I attempted to use this branch locally and ran into a weird error:
```
TypeError: Cannot read properties of undefined (reading 'length')
    at computeLineStarts (...typescript.js:11830:21)
    at getLineStarts (...typescript.js:11876:54)
    at getCurrentLineMap (...typescript.js:118605:48)
    at emitDetachedCommentsAndUpdateCommentsInfo (...typescript.js:122650:106)
    at emitBodyWithDetachedComments (...typescript.js:122468:7)
    at emitSourceFile (...typescript.js:121109:7)
    at pipelineEmitWithHintWorker (...typescript.js:118705:14)
    at pipelineEmitWithHint (...typescript.js:118692:7)
    at pipelineEmit (...typescript.js:118641:5)
    at print (...typescript.js:118556:5)
```

MichaelMitchell-at:
Ok, the error I was getting was some problem with my local clone/build. I opted to just apply this patch directly to a pre-installed `node_modules/typescript/lib/typescript.js` and verified it address the original issue. It does not address the new issue, so I'll go ahead and file one by EOD (in case you want to sneak in a fix for it too)

jakebailey:
This seems fine but I'm still not clear how this totally fixes the linked issue; shouldn't there be an error somewhere about this?

weswigham:
> shouldn't there be an error somewhere about this?

There is... if the file is typechecked. I believe this isn't an isolated declarations error since it's the expectation that usually people won't write nonfunctional/conflicting computed names, and it'll just copy the input nodes to the output, which this PR adjusts our node builder and declaration emitter to do.

jakebailey:
Ah. Right, we don't include check errors in this output (unless we happen on them)

jakebailey (approved):
But we probably need to warn about that somewhere... But it's not like transpileModule is much different. And a syntactic transpiler will probably just copy too.

Issue #58371, opened 2024-04-30
tsc crash with "TypeError: Cannot read properties of undefined (reading 'flags')"


mirek:
### 🔎 Search Terms

none

### 🕗 Version & Regression Information

- This is a crash

### 💻 Code

```ts
// Too large / internal
```

### 🙁 Actual behavior

tsc crashes

### 🙂 Expected behavior

tsc should not crash

### Additional information about the issue

```
 /home/runner/work/my-project/node_modules/.pnpm/typescript@5.4.3/node_modules/typescript/lib/tsc.js:118617
      throw e;
      ^

TypeError: Cannot read properties of undefined (reading 'flags')
    at getCheckFlags (/home/runner/work/my-project/node_modules/.pnpm/typescript@5.4.3/node_modules/typescript/lib/tsc.js:15928:17)
    at getTypeOfSymbol (/home/runner/work/my-project/node_modules/.pnpm/typescript@5.4.3/node_modules/typescript/lib/tsc.js:53259:24)
    at getParameterCount (/home/runner/work/my-project/node_modules/.pnpm/typescript@5.4.3/node_modules/typescript/lib/tsc.js:73669:24)
    at combineUnionParameters (/home/runner/work/my-project/node_modules/.pnpm/typescript@5.4.3/node_modules/typescript/lib/tsc.js:54395:23)
    at combineSignaturesOfUnionMembers (/home/runner/work/my-project/node_modules/.pnpm/typescript@5.4.3/node_modules/typescript/lib/tsc.js:54443:20)
    at /home/runner/work/my-project/node_modules/.pnpm/typescript@5.4.3/node_modules/typescript/lib/tsc.js:54359:204
    at map (/home/runner/work/my-project/node_modules/.pnpm/typescript@5.4.3/node_modules/typescript/lib/tsc.js:204:19)
    at getUnionSignatures (/home/runner/work/my-project/node_modules/.pnpm/typescript@5.4.3/node_modules/typescript/lib/tsc.js:54359:182)
    at resolveUnionTypeMembers (/home/runner/work/my-project/node_modules/.pnpm/typescript@5.4.3/node_modules/typescript/lib/tsc.js:54482:28)
    at resolveStructuredTypeMembers (/home/runner/work/my-project/node_modules/.pnpm/typescript@5.4.3/node_modules/typescript/lib/tsc.js:54946:9)

Node.js v20.9.0
 ELIFECYCLE  Command failed with exit code 1.
```

[...]

PR #58440, opened 2024-05-05
Fixed incorrect `SignatureFlags.HasRestParameter` propagation when combining signatures

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/58371


Andarist:
@jakebailey could you prepare a TS playground for this?

jakebailey:
@typescript-bot test it
@typescript-bot pack this

Andarist:
`hono` change: [TS playground](https://www.typescriptlang.org/play/?ts=5.5.0-pr-58440-3&ssl=11&ssc=3&pln=1&pc=1#code/JYOwLgpgTgZghgYwgAgMIHsQGd0BsUDeAUMsrugOYAUAthFlnBRAPwBcycIAngDTIA6IegAOYYJji4ACnChwaWDl24BtALoBKDgDd0wACYBuIgF8iRfGDKUlyLGCigKG5AF5kGk1eTonFUCkAGUoOMG4RCHQYZARMHHwBcgoTOOw8CCTKd2QqIQE5CixNdwA+ZGJSQqwTUgB6OuQAPRYSGyKBEQBXLAALPKFqzRNTEyA)

RyanCavanaugh:
Are you saying `hono` definitely isn't OK, or the behavior introduced by this PR is definitely not OK?

Andarist:
Oops, I can see how my comment was confusing 😅 

I meant that the behavior change with my PR was definitely not OK in the `hono`'s case. On the other hand, I think that the behavior change in `tiptap` was OK - that *should* report the error. We can compare those 2 playground:
- [with `RemoveThis`](https://www.typescriptlang.org/play/?ts=5.4.5#code/JYOwLgpgTgZghgYwgAgKIA9IgM7APYgDCBMwA5gDwDyADmPjsgLzJwgCeAfMgN4BQyZBEwQQAEwCycKAGsAyggAWEALZwA-AC4Bg5AB9kACkM7dusIuDZNvU2d0g4KiDexgooMgG4795HjoGa2RaegJsHz9BAF8AGl9dYSxcAhspWXi-AEpmbgAlCAQ8KDEKNw8QMljWDk4suwMQAFcAGxafaL4+MUKW6RQEPuxsNBEcBmpA8OYarltBMHYaF2Ryz0jkR2dXd3WdGn7wGwxkhn1N1vadJWAWsWOxlJBz5raNgLCcG1CgjaKQUhkB6nAjEAHkDpdHqDfrIGHDZAAOTwPUmnxGLDYc34CyWKzWlQ2W3xu0J+0OYBsyJ6L0uf0sdypKJQjTpOg+QW+UxwkL4oEgsEQKGpEDBgLRQRmWO4OKEY0k0nkSlUGm0ZgMxgSC0swVlUU2ThJFW8Wt0HPCXPRGz8cVNSVETyZPUy9hyTHyhWKpQJVVmdQaFzefE63V6sPhI3SMgl00xtXm-m5wR+4T+JHIaUVYohwa6-Og8CQyCj2coKcYcexOnt4ijCmUai0Ac1+osVhseqixJ2xutUXNXxCSb7Zlt+prjuLipdZjdHqKJTKpN90vq6sDVxDi2WyAAghxSzMTg6GIeDCKz1PZKWfNuUPv2MfxgQj48zufmecoz5QzCoANwjAIQbAfJ8nh-O9kAKFQ8AANwgAAVHUKAQ7gWAQuUsDEEZDAAOnw6QyGCLE51mHR1CMfDcMI4IAAVpENAVsBQupcigiAwCaKAQAQvEWJ0GwEJ-aE+n-ZAYCaEAEE+ZAyA4sCGAAMWACA7hQqVahMQQJwYECOAUggXVIVT7lWZd4iyGxoLgxDkNQn9-jcTDRAVWR6xVGY5LAAyQGUkyKAfUsAG0ACIa1cpUGzgEKAF1OC0oQXTC+U62VNQQos4SwzExygJ01Ir2jaUf2AGAjHC1KopyWUKsVdy1EMfKQCyLxkAAeja1ZFDwVoxCEKAoGKXM+GEGhiiAnhoiAA)
- [without `RemoveThis`](https://www.typescriptlang.org/play/?ts=5.4.5#code/JYOwLgpgTgZghgYwgAgKIA9IgM7APYgDCBMwA5gDwDyADmPjsgLzJwgCeAfMgN4BQyZBEwQQAEwCycKAGsAyggAWEALZwA-AC5kACh3CsuAtqmyAlM24AlCAjxQxFbGCigyAGlYdOFgD7IQAFcAG2CAbj4AXz4+MVtg6RQEBOxsNBEcBmo6BjSWNi5eAWQwdhoIbWdXEDIIwRA4FQrkKrc65BpE8G0MQwZkfyDQ9qVgYLEejKMQAYCQ8OK8HIJsbVp6FZGSckm+gmIQUlqomLjkxORz1OQAOTw47I3GfO8iwVLyypc24oamr+qx0EnSgojA2jucVmQwWglG4wh9xQg3m7SWT1WyHWuQi0T4oEgsEQKEhEAOR0euWYXkK-EEBlEkmk8iUqg02j0DMyxluSIsTGstnsjlaNU8BR80NRJ1i8QuVzSphklJW1Ilb2Q6Nya2WOC2hx2yCV5PIuJiBOg8CQRuZJso2NVL1pxS5TNkCmUai0un0UwYJmZ-MFdgcTm+YppkpRwxlHxQAEEOHbqb1RNNk-5SRmbbI7RE48hE+xU9yZiwS9NZqTZkqIrLzqDLiswEJtEWKww6wWAAquFTAegANxQLBhs0C4ggpBAEDEs1FZClKgARtBZsu8HhghA2PP2Cut+vyASu2UUFJ2KuACqKYDYbvSRoQQlXs8UK-ccvoZKBB5Xzy9sA-ZDhA3CumkOjFAAdDB0hkJiBR8EGNLFOoyA3neD5QE+L5vhgP5-gBfYDsAw6cJwxTaAUdZnAkjYwBOCBPMgZDPh2BAAGLABA4zvmq3iQfSfo8u2wkgO4xSkDxEwtOGHjFHY4AGN6VDARQF7Xre96Pk0uHlO+nCeAARCCYJGYZSHaFedaKc4QgZG6LKenA1KsWA7EgFx0kUEWdoANpGa6SoemyRkALqcIJQgSYIgUOcFrJqEZElmDRcqNrZLZctMAayBQEp1sAMC6EFzIhWoFh0vZWCOeVcC+nsICpcgAD0LUtIoeAhHO0BQPYMrCDQ9gtjwkQREAA)

We can see here, on the second one, that the error is reported. On the first one, we pass the same argument - but the function(s) with `this` parameters were processed by `RemoveThis`. The resulting functions are conceptually the same as the ones on the second playground.

On top of that, we can remove `Extension*` types from `AnyConfig` and `AnyExtension` to make them unions of 2 types each instead of unions of 3 types each to see the error being reported: [TS playground](https://www.typescriptlang.org/play/?ts=5.4.5#code/CYUwxgNghgTiAEkoGdnwHIHtQB4DyADgC4CWmAdmgLzxTkCeAfPAN4BQ88R9BIAXPGREYJcgHMA3B3jkoAW36DhoydIKwQ5IgKyh4AHxkBXCBCmcwACxIRgO7AkPkTZ6ZmJlKAwqQrIpAL5sbKJEIDAAZlBgCLogAMIUESRi+B5+8DR0TKzSIAAeYeTAALKwANYAylYgclAA-HzSnIYAFK3NnF1E1sgC7F2DXbIKAkIi4uZDg+6+XvA+nv6dgwEANCtdBUXInvagG9MAlJnMAErgmDDAOOMqa7QMjEedTi6BwaBIcIjQqPBlGDlNJzaiPHIDeCzJbedKUKZgJIpASA8qJcjJVRBEJacJRGIAirozEgpaZcHMSHbTSlCrVSy1BpNQZtDrTTg9Eh9XLs4byRR3SabGZw7mLPxTabrYXwamUPaEoGHIYnKjnS7XW7KcQPbLPV7GUwfNjcXjwACCDGJKXJcWtYgMirRSNUpoQlvoAFFCppdhRbQ5HaipGwvtAfojKERZQIPd6dp4Q274Bc5JgAG4gAAqvRwWeYNCzsp9xTQrQAdJXYGJudlVcxstJ6vAK1WYDWBAAFWD8sIwZB556nFMgIhGGDkLM8ECD6QCLMhsMaeARIzkMBzeBiUfx32eABiJBAtjz5L1bOLCYosYYu-lFGVyWPdiUEzEGyOAlTGezufzIcjIRLxpVF6UZcltyIO8-XIQ9nxwD17QAbQAIjlWkgTAuoUIAXUYC8QGVNCSwwqoamwj9F3AcMEEA6M5RglEKhwPUQxICIW3Q0DyKgE4qRI7iGTqVoGM8I4JHgAB6STBEsTATGAWUYBgK42GxAoCCuaMWACIA). It's worth noting here that the reported error is not completely OK either as it mentions "Argument of type '[Mark<any>]'" but it should mention "Argument of type 'Mark<any>'".

I pushed out a fix for `hono` and tests based on `tiptap`. Could we run extended tests on this and build a new playground?

jakebailey:
@typescript-bot test it
@typescript-bot pack this

src/compiler/checker.ts
```diff
@@ -31261,12 +31266,13 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             const paramSymbol = createSymbol(
                 SymbolFlags.FunctionScopedVariable | (isOptional && !isRestParam ? SymbolFlags.Optional : 0),
                 paramName || `arg${i}` as __String,
+                isRestParam ? CheckFlags.RestParameter : isOptional ? CheckFlags.OptionalParameter : 0,
```
Andarist, line 31269: adding `CheckFlags.RestParameter` here is based on the similar change from https://github.com/microsoft/TypeScript/pull/55625

src/compiler/checker.ts
```diff
@@ -13848,8 +13848,13 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             paramMapper = createTypeMapper(right.typeParameters, left.typeParameters);
             // We just use the type parameter defaults from the first signature
         }
+        let flags = (left.flags | right.flags) & (SignatureFlags.PropagatingFlags & ~SignatureFlags.HasRestParameter);
         const declaration = left.declaration;
         const params = combineUnionParameters(left, right, paramMapper);
+        const lastParam = lastOrUndefined(params);
```
jakebailey, line 13854: Perhaps this isn't worth it, but this same code is being done in both `combineSignaturesOfUnionMembers` and `combineSignaturesOfIntersectionMembers`.

- Is it worth making a helper out of this? Probably not, given...
- These two functions in general look really, really similar. Why aren't they the same function with some branching? 🤔

jakebailey (approved):
This _feels_ correct to me, and the `tiptap` thing looks like a correct error. Am I missing anything else here?

src/compiler/checker.ts
Andarist: ye, that's true - I don't think it's worth the effort unifying them right now though

jakebailey:
Just writing this down to make it clear, but tiptap has a variable like:

```
const extendMarkSchema: ((extension: Mark<any, any>) => Record<string, any>) | ((extension: Node<any, any>) => Record<string, any>) | ((extension: Mark<any, any>) => Record<...>) | null | undefined
```

Which it then calls with a value of type `Mark<any, any>`. In current TS, calling this union somehow leads to a valid call:

image

After this PR, it's:

image

`Mark` and `Node` are totally not the same:

```ts
export class Node<Options = any, Storage = any> {
  type = 'node'

  name = 'node'

  parent: Node | null = null

  child: Node | null = null
// ...
}

export class Mark<Options = any, Storage = any> {
  type = 'mark'

  name = 'mark'

  parent: Mark | null = null

  child: Mark | null = null
// ...
}
```

It seems pretty clear to me that the intersection is the right parameter type. Though, I don't think that it can actually ever be satisfied as both are classes (and you can't be both).

jakebailey:
I don't think this PR, your crash, tsx, or Node adding TypeScript support are related; I would suggest filing a new issue with your full stack trace.

PR #58461, opened 2024-05-07
feat(58447): JS object destructuring: allow to shorten expression when assignment is renamed back to original property value

### LINKED ISSUE #58447: JS object destructuring: allow to shorten expression when assignment is renamed back to original property value
**Current behavior**
In a restructuring assignment expression, when alias is renamed back to original property's name, it has the full form (`const { property: property } = testObject;` in the following example):
vscode-destructuring

**Proposed behavior**
Shorten form of assignment (`const { property } = testObject;` in this example) - either automatically, or enabled with configuration.

a-tarasyuk:
Fixes #58447

### DIFF
--- a/src/services/findAllReferences.ts
+++ b/src/services/findAllReferences.ts
@@ -711,7 +711,18 @@ function getDefinitionKindAndDisplayParts(symbol: Symbol, checker: TypeChecker, 
 
 /** @internal */
 export function toRenameLocation(entry: Entry, originalNode: Node, checker: TypeChecker, providePrefixAndSuffixText: boolean, quotePreference: QuotePreference): RenameLocation {
-    return { ...entryToDocumentSpan(entry), ...(providePrefixAndSuffixText && getPrefixAndSuffixText(entry, originalNode, checker, quotePreference)) };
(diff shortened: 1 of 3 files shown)


jakebailey (dismissed):
This looks great; been wanting this for ages...

Hopefully nobody actually wants the rename to duplicate the name? (I think linters actually fix this too in our favor.)

andrewbranch (changes requested):
What the baseline is showing is not the intended behavior. Here’s a real-world reproduction of the test:

https://github.com/user-attachments/assets/2ffc1874-b5c4-4105-8b9b-092103f05859

jakebailey:
Oh, eek, yeah, I forgot that we'd need to do some step after getting the _new_ text and then determine that they're the same, not do it early (which is what this PR does).

andrewbranch:
I don’t want to derail this feature since I don’t know how much more work it would be to tack this on, but when I initially saw the issue, I thought it was talking about property assignments, not just destructuring. I run into this a lot too:

```ts
const foo = 0;
const obj = {
  bar: foo // <-- make shorthand when renaming 'bar' to 'foo' OR vice versa?
};
```

DanielRosenwasser:
It feels like it would be confusing to have one but not the other to be honest.

PR #58462, opened 2024-05-07
Always send configFileDiagEvent and instead set triggerFile for more consistency

sheetalkamat:
We have following rules on when to send the `configFileDiagEvent`
- When file is opened, we want to send the event for all the config files that get loaded for the first time
- If the project was already loaded, we don't want to send the same `configFileDiagEvent`
- But then if the file will be added to inferred project, we want to send `configFileDiagEvent` event for all loaded/looked up projects so that we can figure out if there are errors in config file which is the reason why the file went into inferred project
- When there is change in config file, we always want to send `configFileDiagEvent`
- We also want to send `configFileDiagEvent` if the diagnostics changed any time program update happens
So this handles that ensuring we arent sending duplicate event

468d91c deletes the unnecessary comment about loading/reloading configured project
f738ccd actual change

### DIFF
--- a/src/server/editorServices.ts
+++ b/src/server/editorServices.ts
@@ -2885,22 +2885,20 @@ export class ProjectService {
      *
      * @internal
      */
-    reloadConfiguredProject(project: ConfiguredProject, reason: string, isInitialLoad: boolean, clearSemanticCache: boolean) {
+    reloadConfiguredProject(project: ConfiguredProject, reason: string, clearSemanticCache: boolean) {
         // At this point, there is no reason to not have configFile in the host
         const host = project.getCachedDirectoryStructureHost();
         if (clearSemanticCache) this.clearSemanticCache(project);
 
         // Clear the cache since we are reloading the project from disk
         host.clearCache();
-        const configFileName = project.getConfigFilePath();
-        this.logger.info(`${isInitialLoad ? "Loading" : "Reloading"} configured project ${configFileName}`);
 
         // Load project from the disk
         this.loadConfiguredProject(project, reason);
         project.skipConfigDiagEvent = true;
         project.updateGraph();
 
-        this.sendConfigFileDiagEvent(project, configFileName);
+        this.sendConfigFileDiagEvent(project, project.getConfigFilePath());
     }
 
     /** @internal */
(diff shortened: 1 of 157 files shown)


src/server/editorServices.ts
```diff
@@ -2549,25 +2559,10 @@ export class ProjectService {
...
-        this.loadConfiguredProject(project, reason);
-        return project;
-    }
-
-    /** @internal */
-    private createLoadAndUpdateConfiguredProject(configFileName: NormalizedPath, reason: string) {
-        const project = this.createAndLoadConfiguredProject(configFileName, reason);
-        project.skipConfigDiagEvent = true;
-        project.updateGraph();
+    private createLoadAndUpdateConfiguredProject(configFileName: NormalizedPath, reason: string, triggerFile: NormalizedPath | undefined) {
```
sheetalkamat, line 2562: Removed all the createConfiguredProjectAnd** methods except this one as now `createConfiguredProject` takes reason and sets the pending level to Full to correctly represent the state.

tests/baselines/reference/tsserver/externalProjects/can-handle-tsconfig-file-name-with-difference-casing-with-lazyConfiguredProjectsFromExternalProject.js
```diff
@@ -164,23 +163,6 @@ Info seq  [hh:mm:ss:mss] event:
...
-        "diagnostics": [
-          {
-            "text": "No inputs were found in config file '/A/B/tsconfig.json'. Specified 'include' paths were '[]' and 'exclude' paths were '[]'.",
-            "code": 18003,
-            "category": "error"
-          }
-        ]
-      }
-    }
 Info seq  [hh:mm:ss:mss] event:
```
sheetalkamat, line 166: This is example of duplicate `configFileDiagEvent`

jakebailey:
Is there a linked bug that this fixes?

sheetalkamat:
This came up while i was working on finding default project #57196 and is followup to #58120

jakebailey (approved):

PR #58475, opened 2024-05-08
Preserve type nodes which resolve to errors in declaration emit output

### LINKED ISSUE #58430: [transpileDeclaration API][5.5] Type-only import is missing when type is used as arrow function argument
### 🔎 Search Terms

transpile declaration api emit missing import

### ⏯ Playground Link

https://github.com/MichaelMitchell-at/transpile_declaration_bug_repro/tree/bug2

### 💻 Code

```ts repro
// @strict: true
// @declaration: true
// @showEmit
// @showEmittedFile: a.d.ts
// @showEmittedFile: b.d.ts

// @filename: a.ts
export interface Type {}

// @filename: b.ts
import { type Type } from "./a";

export const foo = (_: Type): void => {};
export const bar = (_: import("./a").Type): void => {};

```

[Workbench Repro](https://www.typescriptlang.org/dev/bug-workbench/?target=7&ts=5.5.0-dev.20240502#code/PTAEAEGcBcCcEsDG0Bco4FcCmAoEEATLRAGwENYzp4B7AOzU13ygAsaB3AUQFt5o8YNp179oWAgDF4JLGjIA6AguiRBESO259o4qTLmgARkpVr14AGYG6ZHocWqcWAB4AHGrGih4dcbEsyRCxQABUATzcQgG8AXxwLa1lbezQTJ3geDy9QaPRIkIio0FjQS1gaHlAAIgVgMmqAbgTXbO9EehgymhpQAF5QAAoAfTQirABKNAA3GngCfoA+XNjm1s92zu8jCn6h0Z8sjcHa+uqJhXGp0Fn5pZXmoA)

### 🙁 Actual behavior

Refer to this program which invokes the `ts.transpileDeclaration` API:
https://github.com/MichaelMitchell-at/transpile_declaration_bug_repro/blob/bug2/transpile.mjs

Basic instructions to invoke it can be found in
https://github.com/MichaelMitchell-at/transpile_declaration_bug_repro/blob/bug2/README.md

```ts
export declare const foo: (_: Type) => void;
export declare const bar: (_: any) => void;
```
```ts
import { type Type } from "./a";
export declare const foo: (_: Type) => void;
export declare const bar: (_: import("./a").Type) => void;
```

### 🙂 Expected behavior

Using the API should emit
```ts
import { type Type } from "./a";
export declare const foo: (_: Type) => void;
export declare const bar: (_: import("./a").Type) => void;
```

weswigham:
Fixes #58430

Like #58428, this takes a case where the checker usually issues an error and blocks declaration emit and adjusts the output to match what `isolatedDeclaration` users expect - retaining the input type node (note that I've flagged this behind the flag I added in #58428 (and renamed that flag to be a bit more general) so error messages will still print `any` instead of whatever node produced the error).

I actually had this change in one of my older omnibus node builder improvement PRs, but it got lost when I was breaking things up to be more reviewable.


tests/baselines/reference/templateInsideCallback.js
```diff
@@ -146,7 +146,7 @@ declare function flatMap(): any;
  * @template T
  * @type {Call<T>}
  */
-declare const identity: any;
+declare const identity: Call<T>;
```
weswigham, line 149: There's a `Type Call is not generic` error on the input here, so the output is fairly flexible. Preserving the erroneous input node is fine, since it's what `isolatedDeclarations` wants.

jakebailey (approved):

PR #58477, opened 2024-05-08
typescript: ignore *.tsbuildinfo

sxlijin:
TS 3.4 introduced the notion of an incremental compilation cache, stored in `.tsbuildinfo` and sometimes in `foo.tsbuildinfo` (if `out` is [configured](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#outfile))

Since this is purely build output data, committing this file is purely noise

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -64,3 +64,4 @@ package-lock.json
 .eslintcache
 *v8.log
 /lib/
+*.tsbuildinfo


jakebailey:
I'm not sure why you're sending this here; our configuration already puts these files into an ignored directory.

sxlijin:
Ah, sorry, I thought I was sending the PR into github/gitignore.

PR #58478, opened 2024-05-08
fix(58399): [isolatedDeclarations][5.5] Autofix does not work when JSX prop contains a dash

### LINKED ISSUE #58399: [isolatedDeclarations][5.5] Autofix does not work when JSX prop contains a dash `-`
### 🔎 Search Terms

isolated declarations, jsx, autofix

- This changed in commit or PR https://github.com/microsoft/TypeScript/pull/58260

### ⏯ Playground Link

_No response_

### 💻 Code

```ts
export const x = <div aria-label="hello" />
```

### 🙁 Actual behavior

No autofix suggestion

### 🙂 Expected behavior

Autofix to add explicit type annotation

This is with `"jsx": "react-jsx"` in tsconfig

a-tarasyuk:
Fixes #58399

### DIFF
--- a/src/services/codefixes/fixMissingTypeAnnotationOnExports.ts
+++ b/src/services/codefixes/fixMissingTypeAnnotationOnExports.ts
@@ -392,7 +392,7 @@ function withContext<T>(
                 sourceFile,
                 targetNode,
                 createSatisfiesAsExpression(
-                    getSynthesizedDeepClone(targetNode),
+                    targetNode,
                     typeNode,
                 ),
             );
--- a/src/services/codefixes/fixMissingTypeAnnotationOnExports.ts
+++ b/src/services/codefixes/fixMissingTypeAnnotationOnExports.ts
@@ -78,6 +78,7 @@ import {
     PropertyAccessExpression,
     PropertyDeclaration,
     setEmitFlags,
+    setParentRecursive,
     SignatureDeclaration,
     some,
     SourceFile,
@@ -392,7 +393,7 @@ function withContext<T>(
                 sourceFile,
                 targetNode,
                 createSatisfiesAsExpression(
-                    targetNode,
+                    setParentRecursive(getSynthesizedDeepClone(targetNode), /*incremental*/ false),
                     typeNode,
                 ),
             );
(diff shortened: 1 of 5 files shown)


src/services/codefixes/fixMissingTypeAnnotationOnExports.ts
```diff
@@ -392,7 +392,7 @@ function withContext<T>(
                 sourceFile,
                 targetNode,
                 createSatisfiesAsExpression(
-                    getSynthesizedDeepClone(targetNode),
```
a-tarasyuk, line 395: The `formattingScanner` requires the Identifier to have `parent` for scanning attributes with a dash (data-foo). When we clone the node, we lose the `parent` node, which causes issues. I think in this case, we can avoid deep cloning. Are there any cases where deep cloning is necessary? @jakebailey

jakebailey: I think deep cloning is always the correct thing to do here, because we're creating new AST stuff and don't want to reuse old nodes. Can we just `setParentRecursive`?

jakebailey: I actually just checked if we could add `return setParentRecursive(clone, false);` to `getSynthesizedDeepClone` itself and it acutally worked, actually _improving_ another test's output. Makes me wonder if that's a better fix (at the cost of us doing this more).

jakebailey: @andrewbranch Do you have any thoughts about this?

andrewbranch: Seems fine to me?

jakebailey: Ah, just noticed the PR was changed to set parents for all cloned nodes; that's what I was askng about, yeah. (though tests need to be updated)

jakebailey:

jakebailey (approved):
Assuming perf is fine, LGTM. Doubt it will say anything; the function very clearly says:

```ts
/**
 * Creates a deep, memberwise clone of a node with no source map location.
 *
 * WARNING: This is an expensive operation and is only intended to be used in refactorings
 * and code fixes (because those are triggered by explicit user actions).
 *
 * @internal
 */
export function getSynthesizedDeepClone<T extends Node | undefined>(node: T, includeTrivia = true): T {
```

PR #58481, opened 2024-05-09
Load unit tests more consistently (pulled from ESM branch)

jakebailey:
Back when I did the module transform, I did a weird hack to get test loading to work bundled and unbundled.

Now, having actually spent more time understanding what's going on here, I think this PR is a better way to do the right thing. We just want to do some custom setup, then pull in the tests, then start running them.

Pulled out of #58419.

### DIFF
--- a/src/testRunner/parallel/worker.ts
+++ b/src/testRunner/parallel/worker.ts
@@ -17,6 +17,9 @@ import {
 } from "../_namespaces/Harness.Parallel.js";
 
 export function start(importTests: () => Promise<unknown>) {
+    // This brings in the tests after we finish setting things up and yield to the event loop.
+    const importTestsPromise = importTests();
+
     function hookUncaughtExceptions() {
         if (!exceptionsHooked) {
             process.on("uncaughtException", handleUncaughtException);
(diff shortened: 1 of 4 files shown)


src/testRunner/parallel/worker.ts
```diff
@@ -338,4 +340,5 @@ export function start() {
     }
 
     process.on("message", processHostMessage);
+    const importTestsPromise = importTests();
```
RyanCavanaugh, line 343: This looks weird because it appears to be an unused `const`. Can we move these two lines above the function declarations?

RyanCavanaugh (approved):

src/testRunner/parallel/worker.ts
jakebailey: Yeah, I'll move this and then force an early return for clarity.

jakebailey: Oh, it's not so simple; there's code right above here...

jakebailey: The best I think I can do is to make a `let` var at the top and then silence the lint rule that says that there's only one assignment.

jakebailey: Came up with something better, PTAL

RyanCavanaugh (approved):

PR #58482, opened 2024-05-09
Infer from union combined single signatures

Andarist:
I treat this as an experiment right now. I think the logic is sound and it makes sense. However, I wonder:
- if it does impact top repos/DT in a negative way? are we missing some tests?
- what's the perf impact of this? that said, if there would be a meaningful perf impact here there still might be ways to improve this

### DIFF
--- a/tests/baselines/reference/jsDeclarationsGetterSetter.js
+++ b/tests/baselines/reference/jsDeclarationsGetterSetter.js
@@ -280,5 +280,5 @@ export class L {
     set x(value: any);
 }
 export class M {
-    set x(value: any);
+    set x(value: boolean);
 }
--- a/tests/baselines/reference/neverIntersectionNotCallable.types
+++ b/tests/baselines/reference/neverIntersectionNotCallable.types
@@ -12,8 +12,8 @@ declare const f: { (x: string): number, a: "" } & { a: number }
 >  : ^^^^^^
 
 f()
->f() : any
->    : ^^^
+>f() : number
+>    : ^^^^^^
 >f : never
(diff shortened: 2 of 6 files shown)


tests/baselines/reference/neverIntersectionNotCallable.types
```diff
@@ -12,8 +12,8 @@ declare const f: { (x: string): number, a: "" } & { a: number }
 >  : ^^^^^^
 
 f()
->f() : any
->    : ^^^
+>f() : number
```
Andarist, line 15: The function itself is still not callable here but that erroneous call shows a better return type now. This is caused by `getSingleSignature` handling structured types now (an intersection in this very case)

tests/baselines/reference/jsDeclarationsGetterSetter.js
```diff
@@ -280,5 +280,5 @@ export class L {
     set x(value: any);
 }
 export class M {
-    set x(value: any);
+    set x(value: boolean);
```
Andarist, line 283: This comes from:
```ts
/**
 * @param {number | boolean} v
 */
const mSetter1 = (v) => {}
/**
 * @param {string | boolean} v
 */
const mSetter2 = (v) => {}
export class M {}
Object.defineProperty(M.prototype, "x", {
    set: Math.random() ? mSetter1 : mSetter2
});
```

This feels like an improvement, those setter functions can be combined safely to a single signature with a `boolean` parameter

jakebailey:
@typescript-bot test it

PR #58486, opened 2024-05-09
Handle config file change and default project management

sheetalkamat:
This is everything from #57196 except there is no change in functionality for finding default project. So that we can get this for 5.5 but behavior change for 5.6

This change includes:
- Unless we have detected change in config file or we are reloading, once file is opened, its default config file name is never recalculated
- All the paths to find default project - opening a file, reloading a project, just searching what's default project for a file or change in config file, use singular path for finding default project
- When config file change is detected, we recalculate what is the default config name for open file and schedule update for project only if its new. If it exists, change in config should not cause the update to project just because file is open. The update happens to project only if it will be referencing config file or is changed for own config file
- `tryFindDefaultConfiguredProjectAndLoadAncestorsForOpenScriptInfo` finds, creates, or reloads the default project and ancestor tree (for find all project)
- `tryFindDefaultConfiguredProjectForOpenScriptInfo` finds, creates or reloads default project for the open file
- When finding orphan configured projects, instead of keeping ref and its login on configured project, we go through open files and all the default projects and projects seen for finding default project are retained. This ensures whatever situation, the "should the project be kept alive" answer never changes because of some updates missing or something like that.

### DIFF
--- a/tests/baselines/reference/tsserver/configFileSearch/tsconfig-for-the-file-exists.js
+++ b/tests/baselines/reference/tsserver/configFileSearch/tsconfig-for-the-file-exists.js
@@ -306,22 +306,21 @@ ScriptInfos::
 Info seq  [hh:mm:ss:mss] FileWatcher:: Triggered with /a/b/projects/project/tsconfig.json 0:: WatchInfo: /a/b/projects/project/tsconfig.json 2000 undefined Project: /a/b/projects/project/tsconfig.json WatchType: Config file
 Info seq  [hh:mm:ss:mss] Scheduled: /a/b/projects/project/tsconfig.json
 Info seq  [hh:mm:ss:mss] getConfigFileNameForFile:: File: /a/b/projects/project/src/index.ts ProjectRootPath: /a/b/projects/proj:: Result: /a/b/projects/project/tsconfig.json
-Info seq  [hh:mm:ss:mss] Scheduled: /a/b/projects/project/tsconfig.json, Cancelled earlier one
 Info seq  [hh:mm:ss:mss] Scheduled: *ensureProjectForOpenFiles*
 Info seq  [hh:mm:ss:mss] Elapsed:: *ms FileWatcher:: Triggered with /a/b/projects/project/tsconfig.json 0:: WatchInfo: /a/b/projects/project/tsconfig.json 2000 undefined Project: /a/b/projects/project/tsconfig.json WatchType: Config file
 Info seq  [hh:mm:ss:mss] DirectoryWatcher:: Triggered with /a/b/projects/project/tsconfig.json :: WatchInfo: /a/b/projects/project 1 undefined Config: /a/b/projects/project/tsconfig.json WatchType: Wild card directory
 Info seq  [hh:mm:ss:mss] Project: /a/b/projects/project/tsconfig.json Detected file add/remove of non supported extension: /a/b/projects/project/tsconfig.json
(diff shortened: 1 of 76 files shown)


tests/baselines/reference/tsserver/configuredProjects/add-and-then-remove-a-config-file-when-parent-folder-has-config-file-and-file-from-first-config-is-not-open.js
```diff
@@ -663,6 +663,24 @@ Info seq  [hh:mm:ss:mss] Elapsed:: *ms DirectoryWatcher:: Close:: WatchInfo: /us
 Info seq  [hh:mm:ss:mss] DirectoryWatcher:: Close:: WatchInfo: /user/username/projects/node_modules/@types 1 undefined Project: /user/username/projects/myproject/folder/tsconfig.json WatchType: Type roots
 Info seq  [hh:mm:ss:mss] Elapsed:: *ms DirectoryWatcher:: Close:: WatchInfo: /user/username/projects/node_modules/@types 1 undefined Project: /user/username/projects/myproject/folder/tsconfig.json WatchType: Type roots
 Info seq  [hh:mm:ss:mss] `remove Project::
+Info seq  [hh:mm:ss:mss] Project '/user/username/projects/myproject/tsconfig.json' (Configured)
```
sheetalkamat, line 666: This is exactly what happens when you open commonFile2 when it has tsconfig next to it that doesnt contain that file. This will be fixed to look for this project in #57196 but this makes the behaviour consistent and not confuse when file belongs to right project or not .

tests/baselines/reference/tsserver/configFileSearch/tsconfig-for-the-file-exists.js
```diff
@@ -306,22 +306,21 @@ ScriptInfos::
 Info seq  [hh:mm:ss:mss] FileWatcher:: Triggered with /a/b/projects/project/tsconfig.json 0:: WatchInfo: /a/b/projects/project/tsconfig.json 2000 undefined Project: /a/b/projects/project/tsconfig.json WatchType: Config file
 Info seq  [hh:mm:ss:mss] Scheduled: /a/b/projects/project/tsconfig.json
 Info seq  [hh:mm:ss:mss] getConfigFileNameForFile:: File: /a/b/projects/project/src/index.ts ProjectRootPath: /a/b/projects/proj:: Result: /a/b/projects/project/tsconfig.json
-Info seq  [hh:mm:ss:mss] Scheduled: /a/b/projects/project/tsconfig.json, Cancelled earlier one
```
sheetalkamat, line 309: This doesnt reschedule update for project if its already scheduled through set of updated projects

tests/baselines/reference/tsserver/configuredProjects/Open-ref-of-configured-project-when-open-file-gets-added-to-the-project-as-part-of-configured-file-update.js
```diff
@@ -495,20 +495,21 @@ ScriptInfos::
 
 Info seq  [hh:mm:ss:mss] FileWatcher:: Triggered with /a/b/tsconfig.json 1:: WatchInfo: /a/b/tsconfig.json 2000 undefined Project: /a/b/tsconfig.json WatchType: Config file
 Info seq  [hh:mm:ss:mss] Scheduled: /a/b/tsconfig.json
+Info seq  [hh:mm:ss:mss] getConfigFileNameForFile:: File: /a/b/src/file1.ts ProjectRootPath: undefined:: Result: /a/b/tsconfig.json
```
sheetalkamat, line 498: Earlier we use to skip if file was not part of inferred project, but change in configFile can have impact on whats the default project, so we upate the default config File Name for this file in cache and make sure that the project is present. Only if its new we schedule update otherwise the project doesnt need to be updated.

jakebailey (approved):

PR #58487, opened 2024-05-09
Re-export more referenced types from protocol.ts

andrewbranch:
Adds some re-exports missed in #57361


sheetalkamat:
baseline update missed?

sheetalkamat (approved):

Issue #58493, opened 2024-05-10
Regression in detecting unused parameters in 5.5.0-beta


danvk:
### 🔎 Search Terms

- 5.5 predicate
- 6133

### 🕗 Version & Regression Information

- This changed between versions 5.4.5 and 5.5.0-beta

### ⏯ Playground Link

https://www.typescriptlang.org/play/?ts=5.5.0-beta#code/GYVwdgxgLglg9mABABwE4FMAmB1GUAWAqmCAM5YAUAhgFyLgDWYcA7mAJSIDeAUIv4gxQQqJAEIxAWSoEAdKiphMcALYV2Abh4BfHkA

### 💻 Code

```ts
function predWithUnused(a: unknown) {
    return !!Math.random();
}
```

### 🙁 Actual behavior

TS 5.5.0-beta reports that `a` is used.

### 🙂 Expected behavior

It should be reported as unused.

### Additional information about the issue

Compare 5.4.5

<img width="751" alt="image" src="https://github.com/microsoft/TypeScript/assets/98301/1e7ddc0e-46a2-4494-ace5-126e953a6c6b">

vs. 5.5.0-beta:

<img width="509" alt="image" src="https://github.com/microsoft/TypeScript/assets/98301/caf6727f-1f9a-4b4c-855c-a0478e069010">

This was almost certainly broken by #57465. It adds a call to `getTypeOfSymbol` for every parameter of a candidate predicate function in `checkIfExpressionRefinesAnyParameter`. This marks each of those parameters as referenced, even though there's no reason to assume that they are, since this call is speculative.

This also affects the "Delete all unused declarations" quickfix.

Pulkit1822:
@danvk , I would like to fix this it.

Issue #51326, opened 2022-10-27
Type predicate assertion does not carry over when used in another function


yogiduzit:
# Bug Report

### 🔎 Search Terms
- predicate
- type predicate
- type predicate return

### 🕗 Version & Regression Information

- This is NOT a crash
- This changed between versions ______ and _______
- This is the behavior in every version I tried, and I reviewed the FAQ for entries about _________
- I was unable to test this on prior versions because _______

### ⏯ Playground Link

[Playground link with relevant code](https://www.typescriptlang.org/play?ts=4.6.4#code/JYOwLgpgTgZghgYwgAgIImAWzgG2QbwChlkQ5MIAuZAZzClAHMBuQgX0MNElkRQCFgUACbIIAD0ghhNNBmx4iJAO5MAygAc4IanQYgW7TgHoAVKeKm0NGtDCywACzhhk2t-NzJgsgPYxkMABPDRQAFW8AhF9pYDBgGO8HKABXFBhfKEDnV20sXAA6S2NCGBSQBHjEn3R8nAB5GDCQiAAeCIkpGTk6gD4ACjyFalqFABpkaNiqnWRBzxxqMIBKZABeXuQAI19fHAhtZeohrx9kCKVkKAgwFKgQSZjhOISQebq3WRXWDlLyyteSUEInewx6ClWl2ut3uSVGuEazVCrWBwgGJxwE36WyEwlWG2Q-F2+202NxBVUBk0h2WP04ZQqM2QjBuAHUmDQtG8cSJqKjIcQrjc7g8ecIKeouXTCFM6B46rI1sgANr4UjkKjIABEcC1yDYEzVZAo1B1WomlMYNFNAGYYGA9WwALqfcG4ZVO1h-RmAhCOCAIADW8JwNFBuGtbpwHoFJGhIvlChoBWwGnDeAJlxIsr2EAKOF8jHTtMFJGAAX6PlRxdjJDrOf2+cL-RZYHZBk5pIxyxLdf1grYyyMnCAA)

### 💻 Code

```ts
interface Animal {
  name: string;
}

interface Bird extends Animal {
  wingSpan: string;
}

/**
 * Asserts that an animal is of type T if condition is true for that animal.
 */
function isAnimalOfType<T extends Animal>(animal: Animal, condition: (animal: T) => boolean): animal is T {
  return condition(animal as T);
}

function isBird(animal: Animal) {
[...]

PR #58495, opened 2024-05-10
Infer AssertsIdentifier type predicates

danvk:
This extends the type predicate inference from #57465 to cover "asserts identifier" predicates as well. For example:

```ts
// function isNumber(x: unknown): asserts x is number
function isNumber(x: unknown) {
  if (typeof x !== 'number') {
    throw new Error(`expected a number, got ${x}`);
  }
}
```

or:

```ts
// function assertNonNullish<T>(x: T): asserts x is NonNullable<T>
function assertNonNullish<T>(x: T) {
  if (x != null) {
    return;
  }
  throw new Error();
}
```

The constraints for inference are:

- Function is not a constructor, getter, setter, or an arrow functions with an expression (not block) body.
- Function does not have an explicit return type annotation, but is inferred to return `void` or `undefined`.
- Function takes at least one parameter that it does not mutate.

The criterion winds up being a bit simpler than it was in #57465:

- Check the type of the parameter at each `return` site (including an implicit return at the end of the function).
- Union these types.
- If this is different than the initial type of the parameter, then we've got an assertion function.

I wound up implementing "different than" with an assignability check since branching constructs can sometimes change the representation of a parameter's type in superficial ways, e.g. `unknown` can become `{} | null | undefined`.

There's no equivalent of the "false check" that we needed to infer identifier type predicates. Assertion predicates are inherently one-sided. If they fail, your code will throw. This means that we can infer assertions even when the negative type can't be represented. For example:

```ts
// function assertIsShortString(x: unknown): asserts x is string
function assertIsShortString(x: unknown) {
  if (typeof x !== 'string') {
    throw new Error('Expected string');
  } else if (x.length > 10) {
    throw new Error('Expected short string');
  }
}
```

This wouldn't work as a type predicate (`x is string`) because you can't represent "all types except short strings." But it's fine as an imprecise assertion, since there's no way to observe the negative case.

A few other things to note:

- This allows assertions to flow (#51326).
- This only infers assertions with a type: `asserts x is T`, not `asserts x`. The latter would also be interesting, but that feels like a different problem.
- This only infers a non-returning code path via its effect on types. This feels a little strange/risky to me since it means this code runs on all sorts of functions that don't throw. But it does work, and this is a natural way to handle all the different patterns (early throw, early return).
- One of the fourslash tests found a bug with unused parameters getting marked as used. This turns out to be an issue with inferred type predicates as well, see #58493.

It's possible that the "asserted type" really is the same as the parameter's declared type, but TS isn't able to figure that out. This happened for one of the baselines. In this case we'll infer a useless type predicate (`function f(x: T): asserts x is T`). This is noisy but won't produce spurious errors. There's one example of this in the baselines.

I didn't observe any slowdowns from this change locally and there's some reason to hope that's the case. For inferred type predicates, the expensive part was the "false" checks (see  https://github.com/microsoft/TypeScript/pull/57465#issuecomment-1976704315). There's no need to do that for assertion functions, though, so we won't pay that price. On the other hand, there might be more implicitly `void`-returning functions than implicitly `boolean`-returning functions, so we'll see!


tests/baselines/reference/narrowingByTypeofInSwitch.types
```diff
@@ -1192,8 +1192,8 @@ function unknownNarrowing(x: unknown) {
 }
 
 function keyofNarrowing<S extends { [K in keyof S]: string }>(k: keyof S) {
->keyofNarrowing : <S extends { [K in keyof S]: string; }>(k: keyof S) => void
->               : ^ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ ^^       ^^^^^^^^^
+>keyofNarrowing : <S extends { [K in keyof S]: string; }>(k: keyof S) => asserts k is (keyof S & number) | (keyof S & symbol) | (keyof S & string)
```
danvk, line 1195: This is the one loss: the other changes are all valid predicates.

```ts
// asserts k is (keyof S & number) | (keyof S & symbol) | (keyof S & string)
function keyofNarrowing<S extends { [K in keyof S]: string }>(k: keyof S) { ... }
```

This should not be an assertion function because:

```
  (keyof S & number) | (keyof S & symbol) | (keyof S & string)
= keyof S & (number | symbol | string)
= keyof S & PropertyKey
= keyof S
```

TS isn't able to figure that out, however, so we get a meaningless assertion predicate.

tests/cases/fourslash/codeFixUnusedIdentifier_all_delete.ts
```diff
@@ -68,7 +68,7 @@ import { x } from "foo";
 import { x2, used2 } from "foo";
 used1; used2;
 
-function f() {
+function f(a, b) {
```
danvk, line 71: This diff is clearly wrong and a bug in my PR. There's a similar issue with the existing inferred type predicates change, see #58493. The fix for that should also apply here.

fatcerberus:
> If this is different than the initial type of the parameter, then we've got an assertion function.

This feels like it’s going to turn a lot of functions into assertions that were never intended to be, just because a parameter was narrowed somewhere in the function.

danvk:
> > If this is different than the initial type of the parameter, then we've got an assertion function.
> 
> This feels like it’s going to turn a lot of functions into assertions that were never intended to be, just because a parameter was narrowed somewhere in the function.

It's not enough for the parameter to be narrowed somewhere in the function. The function has to throw in a way that's tied to that narrowing.

We could do an experiment similar to #58173 to see how often this triggers.

fatcerberus:
Ah, okay - you didn’t mention that in the description in the OP, the algorithm you laid out was just that you check all the return points, not also the throwing points.

danvk:
> Ah, okay - you didn’t mention that in the description in the OP, the algorithm you laid out was just that you check all the return points, not also the throwing points.

The algorithm only looks at parameter types when you return, or (important!) at the implicit return at the end of a function. When you union those types, the only way I can think of that you'd get something distinct from the initial parameter type is by having a code path that throws or does the equivalent (calls another assertion function or `never`-returning function, or goes into an infinite loop).

Maybe an example would make it clearer how this works:

```ts
function assertABC(x: 'A' | 'B' | 'C' | 'D' | 'E') {
  if (x === 'A') {
    return;  // type of x here is 'A'
  } else if (x === 'B' || x === 'C') {
    throw new Error();
  }
  // implicit return; type of x here is 'D' | E'
}
```

Unioning the types of `x` gives `'A' | 'D' | 'E'`. Since `'A' | D' | 'E'` is distinct from `'A' | 'B' | 'C' | 'D' | 'E'`, we have a type assertion function.

If you take a look at the code for the algorithm, it's all about `return`. The association with `throw` is implied by the check on the types.

fatcerberus:
Ah, so it’s based on how TS re-merges control-flow branches by unioning the narrowed types back together, that’s clever.  However do note - that can sometimes create an observably type on the other side (see: subtype reduction, introducing intersections, etc.)

fatcerberus:
Forgive me if you’ve accounted for all this already - I’m just spouting the caveats that come to mind when I think about this.

danvk:
> Ah, so it’s based on how TS re-merges control-flow branches by unioning the narrowed types back together, that’s clever. However do note - that can sometimes create an observably different type on the other side (see: subtype reduction, introducing intersections, etc.)

I won't infer an assertion predicate so long as the initial parameter type is _assignable_ to the asserted type, so superficial differences shouldn't matter. But there are some cases where TS can't figure out the relationship. There's one example of this in the baselines (see my comments on the code). There will probably be others in the wild.

This change would feel less aggressive if I could detect whether there was a code path that throws in the function, but I couldn't find anything that quite did what I wanted.

jakebailey:
@typescript-bot test it
@typescript-bot pack this

for fun

danvk:
Perf impact looks pretty minimal… but what are those two new errors in VS Code?

fatcerberus:
> Assertions require every name in the call target to be declared with an explicit type annotation.

Ouch.  I forgot that was something the compiler checked for.  The existence of that error means you have to be *very* careful about exactly what you infer as an assertion…

danvk:
Summary of breaks (https://github.com/microsoft/TypeScript/pull/58495#issuecomment-2105358748):

- darkreader: we inferred that an arrow function was a type assertion predicate, which doesn't generally work.
- InvokeAI: same; `validateBaseCompatibility` is an arrow function. This PR correctly infers that it returns a type assertion predicate, but this results in TS2775. If it were a function statement, the inferred type assertion predicate works fine.
- vscode: In this case, we infer a type assertion predicate for a method. Again, this doesn't always work and can produce TS2776. But that's not what happens here. The `ExtHostTreeView.validateTreeItem` method becomes a type assertion predicate. This, in turn, exposes what I believe had been a latent TS bug: #58518

The fix for all of these is to not infer type assertion predicates for methods or arrow functions because this can produce a confusing error at the call site. See #33622 for why. This definitely limits the scope of this PR: the latest version only runs on plain old function statements.

I think the nicest win from this PR is that it lets type predicates naturally flow through assertion wrappers:

```ts
declare function isFoo(x: any): x is Foo;

// function inferFromType(x: unknown): asserts x is Foo
function inferFromType(x: unknown) {
  if (!isFoo(x)) {
    throw new Error();
  }
}
```

Andarist:
> I think the nicest win from this PR is that it lets type predicates naturally flow through assertion wrappers:

This is dope 🚀

danvk:
At first I thought the [minimatch break](https://github.com/microsoft/TypeScript/pull/58495#issuecomment-2108525914) had something with old-fashioned JS constructor function statements that were inferred to be type predicates. But it turned out to be more straightforward. The relevant code is [here](https://yarnpkg.com/package?name=minimatch&version=3.1.2&file=%2Fminimatch.js) and all the failures are calls to this function:

```js
var MAX_PATTERN_LENGTH = 1024 * 64
var assertValidPattern = function (pattern) {
  if (typeof pattern !== 'string') {
    throw new TypeError('invalid pattern')
  }

  if (pattern.length > MAX_PATTERN_LENGTH) {
    throw new TypeError('pattern is too long')
  }
}
```

So I also need to exclude function _expressions_ from inference, not just arrow functions.

tests/cases/fourslash/codeFixUnusedIdentifier_all_delete.ts
danvk: I confirmed that @Andarist's PR #58514 also fixes this.

fatcerberus:
Ironically, inferring `assertValidPattern` as an assertion predicate is actually correct!  But unfortunately it hits the limitation that requires them to be explicitly annotated...

danvk:
> Ironically, inferring `assertValidPattern` as an assertion predicate is actually correct! But unfortunately it hits the limitation that requires them to be explicitly annotated...

Right. So far as I'm aware, the algorithm is correct: it will never infer a type assertion predicate that's inaccurate. But there are many situations (methods, arrow functions, function expressions) where it will push you right into TS2775 / TS2776.

jakebailey:
@typescript-bot test it
@typescript-bot pack this

andriyor:
Is there any chance it will be merged?

PR #58514, opened 2024-05-13
Fixed a regression with reporting unused parameters in potential predicates

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/58493


jakebailey:
@typescript-bot test it

src/compiler/utilities.ts
```diff
@@ -11360,18 +11360,23 @@ export function createNameResolver({
                         location = root.parent;
                     }
                     break;
-                case SyntaxKind.Parameter:
+                case SyntaxKind.Parameter: {
+                    const param = location as ParameterDeclaration;
```
sandersn, line 11364: style question: is there a reason for all the casts in this code? Maybe it's better to drop this alias and use `location as ParameterDeclaration)` everywhere.

src/compiler/utilities.ts
```diff
@@ -11360,18 +11360,23 @@ export function createNameResolver({
                         location = root.parent;
                     }
                     break;
-                case SyntaxKind.Parameter:
+                case SyntaxKind.Parameter: {
+                    const param = location as ParameterDeclaration;
+                    if (lastLocation === param.name) {
+                        lastSelfReferenceLocation = param;
```
sandersn, line 11366: I'd almost prefer a new parameter and entry in `isSelfReferenceLocation` like

```ts
    function isSelfReferenceLocation(node: Node, lastLocation: Node): node is SelfReferenceLocation {
        switch (node.kind) {
            case SyntaxKind.Parameter: return lastLocation === (node as ParameterDeclaration).name
            // ...
```

src/compiler/utilities.ts
```diff
@@ -11360,18 +11360,23 @@ export function createNameResolver({
                         location = root.parent;
                     }
                     break;
-                case SyntaxKind.Parameter:
+                case SyntaxKind.Parameter: {
+                    const param = location as ParameterDeclaration;
+                    if (lastLocation === param.name) {
```
sandersn, line 11365: what does this do when param.name is a BindingPattern? Do we have tests to make sure that they still correctly get marked unused (and used when used)?

sandersn (changes requested):
@danvk guessed that the source of the problem was a speculative `getTypeOfSymbol`. But this PR fixes resolution for all parameters. If it applies to all parameters, why wasn't this observed before Dan's change?

1. Need tests of used/unused binding patterns. These probably exist already, but please double-check.
2. If this implementation is correct, I think it would be easier to read in `isSelfReferenceLocation`, as long as that doesn't make performance worse (from adding a second parameter).

Andarist:
> My (fuzzier) guess is that we haven't tried to resolve a parameter name from itself before and never observed that it mistakenly caused the parameter to be marked as used.

This is correct.

> But I'm not sure what in the new code is causing that.

It's `checkIfExpressionRefinesParameter` through the `getFlowTypeOfReference`. It leads to `narrowTypeByCallExpression` and thus to `isMatchingReference` and `getResolvedSymbol` that calls `resolveName`. It's this newly introduced "fake" control flow that changed this. By fake, I mean it's not created by the binder.

src/compiler/utilities.ts
Andarist: No specific reason, I just prefer to have a single cast than multiple ones that are the same for the same variable. I can change this if you prefer the other style.

src/compiler/utilities.ts
Andarist: I just pushed out the suggested change

src/compiler/utilities.ts
Andarist: yeah, there are tests for that already and the changes to `checkIfExpressionRefinesAnyParameter` that introduced this bug don't affect binding patterns

src/compiler/utilities.ts
```diff
@@ -11417,7 +11417,7 @@ export function createNameResolver({
                         location = root.parent;
                     }
                     break;
-                case SyntaxKind.Parameter:
+                case SyntaxKind.Parameter: {
```
jakebailey, line 11420: Nit: leftover formatting change?

src/compiler/utilities.ts
```diff
@@ -11429,6 +11429,7 @@ export function createNameResolver({
                         }
                     }
                     break;
+                }
```
Andarist, line 11432:
```suggestion
```

src/compiler/utilities.ts
Andarist:
```suggestion
                case SyntaxKind.Parameter:
```

jakebailey:
@typescript-bot test it
@typescript-bot pack this

jakebailey (approved):

sandersn (approved):

jakebailey:

PR #58516, opened 2024-05-13
Partially reuse type nodes

### LINKED ISSUE #58515: Reuse original type nodes even when part of the node is not usable
### 🔍 Search Terms

node reuse declarations

### ✅ Viability Checklist

### ⭐ Suggestion

When `tryReuseExistingTypeNodeHelper` encounters an node it can't use, it will throw away the whole original node. We can instead fall back on type printing for only the subnode that is not reusable. 

### 📃 Motivating Example

```ts
// @filename: a.ts
export const nImported = "nImported"
export const nNotImported = "nNotImported"
const nPrivate = "private"
export const o = (p1: typeof nImported, p2: typeof nNotImported, p3: typeof nPrivate) => null! as { foo: typeof nImported, bar: typeof nPrivate, baz: typeof nNotImported }

// @filename: b.ts
import { o, nImported } from "./a";
export const g = o
console.log(nImported);

// @filename: c.ts
import * as a from "./a";
export const g = a.o
```

[Playgorund link]()

```
// currently generates
// @filename:c.d.ts
import { nImported } from "./a";
export declare const g: (p1: typeof nImported, p2: "nNotImported", p3: "private") => {
    foo: "nImported";
    bar: "private";
    baz: "nNotImported";
};

/// proposed
// @filename: c.d.ts
export declare const g: (p1: typeof a.nImported, p2: typeof a.nNotImported, p3: "private") => {
    foo: typeof a.nImported;
    bar: "private";
    baz: typeof a.nNotImported;
};

```

### 💻 Use Cases

1. What do you want to use this for? - Improve declaration fidelity and speed
2. What shortcomings exist with current approaches? - Throwing away the type node can happen half way through resulting in duplicate work. 

dragomirtitian:
Fixes #58515


tests/baselines/reference/callbackTagNamespace.types
```diff
@@ -16,7 +16,7 @@ var x = 1;
 /** @type {NS.Nested.Inner} */
 function f(space, peace) {
 >f : (space: any, peace: any) => string | number
->  : ^     ^^   ^^     ^^   ^^^^^^^^^^^^^^^^^^^^
+>  : ^     ^^^^^^^     ^^^^^^^^^^^^^^^^^^^^^^^^^
```
jakebailey, line 19: How does reuse end up being lowered here?

src/compiler/checker.ts
```diff
@@ -8340,6 +8341,59 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+            const resolvedSymbol = symbol.flags & SymbolFlags.Alias ? resolveAlias(symbol) : symbol;
+            if (isSymbolAccessible(symbol, context.enclosingDeclaration, meaning, /*shouldComputeAliasesToMakeVisible*/ false).accessibility !== SymbolAccessibility.Accessible) return undefined;
+            return symbolToTypeNode(resolvedSymbol, context, meaning, typeArguments);
+        }
+
+        function canReuseTypeNode(context: NodeBuilderContext, existing: TypeNode) {
+            if (isInJSFile(existing)) {
+                if (isLiteralImportTypeNode(existing)) {
+                    getTypeFromImportTypeNode(existing);
+                    // This was existing code. Should we call getTypeFromImportTypeNode to make sure teh symbol is actually there ?
```
weswigham, line 8357:
```suggestion
                    // This was existing code. Should we call getTypeFromImportTypeNode to make sure the symbol is actually there ?

```

src/compiler/checker.ts
```diff
@@ -8340,6 +8341,59 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+            }
+            if (isTypeReferenceNode(existing)) {
+                const symbol = resolveTypeReferenceName(existing, SymbolFlags.Type, /*ignoreErrors*/ true);
+                if (symbol.flags & SymbolFlags.TypeParameter) {
+                    return true;
+                }
+                if (isInJSDoc(existing)) {
+                    return existingTypeNodeIsNotReferenceOrIsReferenceWithCompatibleTypeArgumentCount(existing, getTypeFromTypeNode(existing))
+                        && !getIntendedTypeFromJSDocTypeReference(existing) // We should probably allow the reuse of JSDoc reference types such as String Number etc
+                        && resolveTypeReferenceName(existing, SymbolFlags.Type, /*ignoreErrors*/ true) !== unknownSymbol; // JSDoc type annotations can reference values (meaning typeof value) as well as types. We only reuse type nodes
```
weswigham, line 8378:
```suggestion
                        && symbol !== unknownSymbol; // JSDoc type annotations can reference values (meaning typeof value) as well as types. We only reuse type nodes

```

weswigham: Or
```suggestion
                        && resolveTypeReferenceName(existing, SymbolFlags.Value, /*ignoreErrors*/ true) !== unknownSymbol; // JSDoc type annotations can reference values (meaning typeof value) as well as types. We only reuse type nodes

```
depending on intent

src/compiler/checker.ts
```diff
@@ -8340,6 +8341,59 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+                            // The import type resolved using jsdoc fallback logic
+                            (!existing.isTypeOf && !(nodeSymbol.flags & SymbolFlags.Type)) ||
+                            // The import type had type arguments autofilled by js fallback logic
+                            !(length(existing.typeArguments) >= getMinTypeArgumentCount(getLocalTypeParametersOfClassOrInterfaceOrTypeAlias(nodeSymbol)))
+                        )
+                    );
+                }
+            }
+            if (isTypeReferenceNode(existing)) {
+                const symbol = resolveTypeReferenceName(existing, SymbolFlags.Type, /*ignoreErrors*/ true);
```
weswigham, line 8371: Alternatively,
```suggestion
                void getTypeFromTypeReference(existing);
                const symbol = getNodeLinks(existing).resolvedSymbol;

```
to used the cached value (and ensure it's present).

src/compiler/checker.ts
```diff
@@ -8340,6 +8341,59 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             }
         }
 
+        function serializeTypeName(context: NodeBuilderContext, node: EntityName, isTypeOf?: boolean, typeArguments?: readonly TypeNode[]) {
+            const meaning = isTypeOf ? SymbolFlags.Value : SymbolFlags.Type;
+            const symbol = resolveEntityName(node, meaning, /*ignoreErrors*/ true);
+            if (!symbol) return undefined;
```
weswigham, line 8347: The default behavior here is to fall back to serializing the type node when the name doesn't resolve, and usually I'm inclined to approve of that approach, but the `transpileDeclaration` issues we've been getting have made me keenly aware that *most* names won't resolve when transpiled under that mode, and you probably want the default behavior to be copy-the-node when the name doesn't resolve rather than print-any. This is somewhat reasonable since name-that-does-not-resolve is just `any` with more steps to get there.

src/compiler/checker.ts
```diff
@@ -8340,6 +8341,59 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+                        !(
+                            // The import type resolved using jsdoc fallback logic
+                            (!existing.isTypeOf && !(nodeSymbol.flags & SymbolFlags.Type)) ||
+                            // The import type had type arguments autofilled by js fallback logic
+                            !(length(existing.typeArguments) >= getMinTypeArgumentCount(getLocalTypeParametersOfClassOrInterfaceOrTypeAlias(nodeSymbol)))
+                        )
+                    );
+                }
+            }
+            if (isTypeReferenceNode(existing)) {
```
weswigham, line 8370: The `const` in `as const` is a type reference, so you mat want to except/handle that specially here, rather than looking for a variable named `const`.

src/compiler/checker.ts
```diff
@@ -8353,6 +8407,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             if (hadError) {
                 return undefined;
             }
+            context.approximateLength += existing.end - existing.pos;
```
weswigham, line 8410: 👌

src/compiler/checker.ts
```diff
@@ -8503,15 +8577,48 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
-
-                if (isEntityName(node) || isEntityNameExpression(node)) {
-                    if (isDeclarationName(node)) {
-                        return node;
+                if (isTypeQueryNode(node)) {
+                    const { introducesError, node: exprName } = trackExistingEntityName(node.exprName, context);
+                    if (introducesError) {
+                        const serializedName = serializeTypeName(context, node.exprName, /*isTypeOf*/ true);
+                        if (serializedName) {
+                            serializeExistingTypeNode(context, node);
```
weswigham, line 8585: Serialize for side effects? Probably warrants a comment explaining why. And being inside a `void` expression (which is just a personal preference on my part).

src/compiler/checker.ts
```diff
@@ -5993,6 +5993,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
          * It also calls `setOriginalNode` to setup a `.original` pointer, since you basically *always* want these in the node builder.
          */
         function setTextRange<T extends Node>(context: NodeBuilderContext, range: T, location: Node | undefined): T {
+            if (range === location) return range;
```
weswigham, line 5996: Move this below the copy check below - that copy needs to happen even if the range and location are the same node, originally.

src/compiler/checker.ts
weswigham: Also, in response to the comment: Probably.

weswigham:
@dragomirtitian Do you also wanna copy your tests to the `tests/cases/transpile` folder to run them through the `transpileDeclaration` API endpoint, to ensure the output is actually what you want when global/other file context isn't present? That's particularly relevant for some of these examples that generate `import` type nodes (whose generation would seem to rely on the symbol actually resolving).

tests/baselines/reference/callbackTagNamespace.types
dragomirtitian: The original version for JsDoc did fall back to type printing, but it also set the original node on this synthetic type node counting it as a reuse. Falling back to type printing should not count the node as reused in my opinion. 

```ts
if (isTypeReferenceNode(node) && isInJSDoc(node) && (!existingTypeNodeIsNotReferenceOrIsReferenceWithCompatibleTypeArgumentCount(node, getTypeFromTypeNode(node)) || getIntendedTypeFromJSDocTypeReference(node) || unknownSymbol === resolveTypeReferenceName(node, SymbolFlags.Type, /*ignoreErrors*/ true))) {
                    return setOriginalNode(typeToTypeNodeHelper(getTypeFromTypeNode(node), context), node);
                }
                
```

src/compiler/checker.ts
dragomirtitian: Was left over from some debugging. removed.

src/compiler/checker.ts
dragomirtitian: Yeah. The double `resolveTypeReferenceName` bas left over from moving the code around. It can just be `symbol !== unknownSymbol`. Changed.

dragomirtitian: I switched to `symbol.flags & SymbolFlags.Type` the idea is that in JSDoc values can be used in a type position, and we want to avoid reusing types in this case.

dragomirtitian:
> @dragomirtitian Do you also wanna copy your tests to the `tests/cases/transpile` folder to run them through the `transpileDeclaration` API endpoint, to ensure the output is actually what you want when global/other file context isn't present? That's particularly relevant for some of these examples that generate `import` type nodes (whose generation would seem to rely on the symbol actually resolving).

I can copy them, but these tests are not actually isolated declarations compliant and I would not expect these changes to actually work with `transpileDeclaration`. The use of import types and symbol resolution during node reuse will only work with the full type checker. This will only really kick in when a declaration is reused in a new location, and this should never happen with isolated declarations as if the type is not specified you get an error in isolated declarations.

src/compiler/checker.ts
dragomirtitian: This actually isn't an issue the way this function is currently used. `trackExistingEntityName` will already not error if the symbol is not found and will just return the original name. `serializeTypeName` is always used as a fallback, when `trackExistingEntityName` fails so it will never be called if the symbol is not resolvable. 

I added a test that shows that the types are reused even if the symbol is not found.

weswigham:
> I can copy them, but these tests are not actually isolated declarations compliant and I would not expect these changes to actually work with transpileDeclaration.

Yeah, I would just suggest the test be added to ensure there's still an `--isolatedDeclarations` error issued by the declaration emitter for these cases, though, and that the error itself isn't dependent on symbol or type information being available to be issued.

weswigham (approved):
Thanks for the extra tests - from the issues we've been getting in thus far, it seems like it's probably a good idea going forward to test both the full typecheck behavior and the `transpileDeclaration` behavior for anything `isolatedDeclarations`-adjacent, so we know the two are consistent.

Issue #58518, opened 2024-05-13
Narrowing by type predicate fails to produce intersection type with weak type


danvk:
### 🔎 Search Terms

- predicate
- intersection
- weak

### 🕗 Version & Regression Information

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about _________

### ⏯ Playground Link

https://www.typescriptlang.org/play/?ts=5.5.0-beta&ssl=7&ssc=20&pln=7&pc=21#code/JYOwLgpgTgZghgYwgAgILIN4ChnIDZwBGEeA-AFzIDOYUoA5gNw7JwXW0OO4D0PyAWzgBrFFAgBHAK7BxAE2RgA9shLABoOJFVQoSqFgC+WLKEixEKAEKYWBYmUo06IJr37iBSgG4oqACzh5ZAAHPRDoMABPRRU1DRAtFGg9A1xCSkIlJTwIOBBmYywYKRAEMGAlEGRgKisACjB-BkolQgArCHKASkomhhqqZBtsXHEwKShqgHJCaZrq-tdCkxKyiqrFCBpG5tdKVG7bXCX6ADo4bmQ+ZCVhFmAYZHrahtPuo9HcRT3zy-cdKlSCxjMYgA

### 💻 Code

```ts
interface A {
  label?: string;
  a?: string;  // make required to eliminate error
}

interface B {
  label?: string;  // remove shared property to eliminate error
  b: boolean;
}

function isB(thing: object): thing is B {
  return 'b' in thing;
}

function test(thing: A) {
  thing.a;  // ok
  if (isB(thing)) {
    thing.a;  // error?
  }
}
```

### 🙁 Actual behavior

The type of `thing` inside the conditional is `B`.

### 🙂 Expected behavior

The type of `thing` inside the conditional should be `A & B`.

### Additional information about the issue

This only happens if:

1. `A` is a "weak" type (all optional properties).
2. `A` and `B` have at least one overlapping property.

If either of these isn't the case, then you get the expected intersection type:
<img width="602" alt="image" src="https://github.com/microsoft/TypeScript/assets/98301/3134b071-0472-4d25-bea9-18a3baf20119">

This came up in vscode while testing https://github.com/microsoft/TypeScript/pull/58495#issuecomment-2105358748. The issue is there already but is latent because control flow analysis can't see that a type predicate becomes a type assertion without that PR.

[...]

PR #58520, opened 2024-05-13
Narrow non-directly related weak types by intersecting them with candidates

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/58518

### DIFF
--- a/tests/baselines/reference/typeGuardIntersectionTypes.types
+++ b/tests/baselines/reference/typeGuardIntersectionTypes.types
@@ -241,16 +241,16 @@ function identifyBeast(beast: Beast) {
 >                : ^^^^^^^
 >hasWings : (x: Beast) => x is Winged
 >         : ^ ^^     ^^^^^^^^^^^^^^^^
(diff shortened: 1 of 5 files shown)


tests/baselines/reference/typeGuardIntersectionTypes.types
```diff
@@ -241,16 +241,16 @@ function identifyBeast(beast: Beast) {
 >                : ^^^^^^^
 >hasWings : (x: Beast) => x is Winged
 >         : ^ ^^     ^^^^^^^^^^^^^^^^
->beast : Legged
->      : ^^^^^^
+>beast : Beast & Legged
```
Andarist, line 244: This is essentially the same test case as the added one. Without the change, we can't access `beast.wings` here (even though it's permitted before narrowing). 

Regardless of that, I find it useful to include the @danvk's test case here as I could include the ticket reference in a clean manner above it.

jakebailey:
@typescript-bot test it

PR #58521, opened 2024-05-13
Update unicode identifier start/part with more recent unicode version

weswigham:
I glanced at the script folder and thought to myself "hey, when did we last run this?" and the answer is: 10 major node versions ago, which include 3 major unicode versions. We should probably run this whenever node major updates unicode, at least, so we properly support new character classes as identifiers (when they're classified as such).


jakebailey (approved):
I could have sworn I sent a PR with this, but apparently not...

Issue #9824, opened 2016-07-19
Use lowercase module name when looking up `@types` packages


mhegazy:
New NPM packages can not have uppercase letters in their names. the original package has been shipped already. May be we can make the package lookup in the compiler always look for the lower case name of the type package..

See https://github.com/DefinitelyTyped/DefinitelyTyped/issues/9716#issuecomment-233519331

andy-ms:
It appears to already work this way, even with `--forceConsistentCasingInFileNames`.

My setup:

`src/tsconfig.json`:

    {
      "compilerOptions": {
        "forceConsistentCasingInFileNames": true,
        "outDir": "../out"
      }
    }

`node_modules/@types/foo/index.d.ts`:

    declare const x: number;
    export default x;

`node_modules/@types/bar/index.d.ts`:

    declare namespace Bar {
        const x: number;
    }

`src/index.ts`:

    import foo from 'FoO';
    foo + 1;

`src/foo-user.ts`:
    
    import foo from 'FoO';
    foo + 1;

`src/bar-user.ts`:

    /// <reference types='BaR'/>
    Bar.x;

`src/CAPS.ts`:

    export default 0;

`src/caps-user.ts`:

    // This one really does fail.
    import c from 'caps';
    c + 1;

There's still a chance that this won't work on a real case-insensitive OS.

DanielRosenwasser:
The thing that flag does is yell at you if you use uppercase in one place and lowercase in another, so it's unrelated to this AFAIK.

andy-ms:
My bad. I guess I'll have to test this out on my linux box.

PR #58525, opened 2024-05-13
Do not lowercase typeReference directive name

### LINKED ISSUE #45096: lowercase filename in triple slash directives

sheetalkamat:
As part of https://github.com/microsoft/TypeScript/issues/9824 all type reference directives are lower cased so type="somePackage" results in looking for type reference of somepackage instead.

We also did not lowercase file names found as part of auto type references. So there was always some descripency with how we handled typeRef. Now with this change, it's as written.

I reverted that part and only kept the jsTypings to do the lowercasing to find the `@types` packages to install. 
This fixes issues with `forceConsistentCasingInFileNames` esp when using relative paths in `types` reference. 

#26948 becomes absolute with this 
Fixes #45096


sheetalkamat:
@typescript-bot test this
@typescript-bot test top100
@typescript-bot run dt
@typescript-bot pack this

tests/baselines/reference/tsserver/typeReferenceDirectives/when-typeReferenceDirective-contains-UpperCasePackage.js
```diff
@@ -117,7 +166,6 @@ Info seq  [hh:mm:ss:mss] 	Files (4)
 	  Matched by default include pattern '**/*'
 	../lib/@types/UpperCasePackage/index.d.ts
 	  Entry point for implicit type library 'UpperCasePackage'
-	  Type library referenced via 'UpperCasePackage' from file '../lib/@app/lib/index.d.ts'
```
sheetalkamat, line 120: This stops showing as this because of how this is processed. When the `UpperCasePackage` typings are already found and are primary then they are not processed again and hence the file reason for this is never added. Thats a separate issue and am looking into updating that code as well but its not related to this change. 

it just stopped showing because before this the package name by auto type ref was "UpperCasePackage" while by type Ref was "uppercasepackage" which made two distinct entries and hence processed again.

andrewbranch (approved):
This seems correct, but it also seems like a lot of people in the linked issue are misusing type reference directives. They’re not really supposed to resolve to files, right? They should be package directories in `typeRoots`?

sheetalkamat:
Yes they are but given that we dont disallow relative names in the `type` it just seems like something we shouldnt do. I mean we shouldnt be in buisness of "lowercasing". Please write what you mean. So i feel like this is better change. 

We started doing this because of "Jquery.min" etc files and some other auto type Aquisition issues. But if you write "types" in your file, write it with correct casing is logical i think.

PR #7775, opened 2016-04-01
Library directives support

vladima:
This PR continues the work started in #7549

PR #58528, opened 2024-05-13
Stop checking existing resolution now that we have concept of package ID

sheetalkamat:
This was added in #7775 when we didnt have deduplication logic So now just rely on the existing dedup logic for type Ref management. 
This now follows exact same logic of module resolution and let the deduping take care of it. It also solves issue of when and how the type ref was resolved when they are same etc. Note that module resolution falls back to typeRef resolution on failure so right now you could still have a types in your program from two different versions. So this just streamlines it.

The test cases dont have package json and hence those are not deduped


sheetalkamat:
@typescript-bot test this
@typescript-bot test top100
@typescript-bot run dt
@typescript-bot pack this

tests/baselines/reference/library-reference-4.symbols
```diff
@@ -12,10 +12,14 @@ declare var foo: any;
 
 === /node_modules/foo/node_modules/alpha/index.d.ts ===
 declare var alpha: any;
->alpha : Symbol(alpha, Decl(index.d.ts, 0, 11))
+>alpha : Symbol(alpha, Decl(index.d.ts, 0, 11), Decl(index.d.ts, 0, 11))
```
sheetalkamat, line 15: These didnt have package.json so didnt get deduped. But I added test case 4b and 5b where there is package.json

tests/baselines/reference/library-reference-5.errors.txt
```diff
@@ -15,16 +12,13 @@
...
-!!! error TS4090:     Type library referenced via 'alpha' from file '/node_modules/bar/index.d.ts'
-!!! related TS1404 /node_modules/foo/index.d.ts:1:23: File is included via type library reference here.
     declare var bar: any;
     
-==== /node_modules/bar/node_modules/alpha/index.d.ts (0 errors) ====
+==== /node_modules/bar/node_modules/alpha/index.d.ts (1 errors) ====
     declare var alpha: {};
+                ~~~~~
+!!! error TS2403: Subsequent variable declarations must have the same type.  Variable 'alpha' must be of type 'any', but here has type '{}'.
+!!! related TS6203 /node_modules/foo/node_modules/alpha/index.d.ts:1:13: 'alpha' was also declared here.
```
jakebailey, line 23: I was going to say "the old message seemed useful because it said where the dupe was", but I guess this related line would be enough. Though it's moderately a shame to lose a message that gives a potential solution.

jakebailey (approved):

tests/baselines/reference/library-reference-5.errors.txt
sheetalkamat: The things work much better where you could have two typings that dont add to global and used by two packages and hence two different versions. This handles that well. We have this issue with module resolution. The only message we gave was with typeRef and it wasnt correct checking either so i feel better removing this custom logic for typeRef only and allows packages to use typeRef of different versions if they dont add conflicting global info.

PR #58536, opened 2024-05-14
Declaration emit should avoid issuing errors on unresolved names

### LINKED ISSUE #58496: [transpileDeclaration API] What to do about "has or is using private name" diagnostics
### 🔍 Search Terms

isolated declarations transpile transpiledeclaration diagnostic private name

### ✅ Viability Checklist

### ⭐ Suggestion

Either make `ts.transpileDeclaration` not report "has or is using private name" diagnostics, or provide a method to easily check whether a diagnostic is related to private names.

### 📃 Motivating Example

Usage of ts.transpileDeclaration for any project with isolatedDeclarations

### 💻 Use Cases

`ts.transpileDeclaration` will report usage of global types such as `Pick`, `Omit`, `Promise`, or project-specific ones, as using private names. These are not really actionable, so I wonder if these diagnostics should just not be reported at all. What I've been doing is to just filter them out. The structured way of doing that is checking the diagnostic code, but there are so many codes to check that I decided to take a more unstructured approach by just checking `diagnostic.messageText.includes('private name')`. So at the very least, if these diagnostics will continue to be reported, it would be nice if there was some boolean method that says whether or not a given diagnostic is related to private names.

### LINKED ISSUE #58535: transpileDeclaration - Errors on unresolvable globals
### 🔎 Search Terms

transpileDeclaration globals

- This changed between versions 5.4 and 5.5 beta

### ⏯ Playground Link

[Playground](https://www.typescriptlang.org/dev/bug-workbench/?emitDeclarationOnly=true#code/PTAEAEBMFMGMBsCGAnRAXAlgewHYC41kBXaAWACgQJoBbDNAETiVU1wHkd4BPPUQkhQpVwAMwzxoAOUQ1ofAObwsAI0TwAzgDo0Gimm4AHaKACyGDRow4FAcWVr4AFSMmAvKABES1es8BuChgEFBNYXA00UDpLazsHdQA1dRI+TwBBAOEwMQlpWXlQaxgADx09SjA0AAsLfmhI0ERLIjkNfmroZBM6jSw5UB9HUBhxHHpsHHbRLGQzCysbe19nVyacSGiFuOXHZPgSDvQm7tAcLCjEADdECUQVSX4sflQpwzymENZJimgSw1mUXCUyiJT45liSwSq2MoA8OCI8HgAEJAuQ-gDkECIlFRDg4aAABSIcHbKErFzGACUpMh8QpazcAD4zoiUWiMYDQMDGniAEwE4m0xb0xyU6BUuEshFI5FNdoQkW7dTitG-f5cnmgvl8AzGLCiLZ05XwfaHeFs1HqzHYkGgPEAZkFJP4rgNRqV0LNEt1bsNMU9K29UtZso5Gqx3Jx9pwABZnb79f6yaKkikJSGZSj5a6kx6dl702ryNbNdHnnwAN4UUC10AAbQDBaD6YAunxIsg4hQAL4ErOooA)

### 💻 Code

```ts
// @declaration:true
// @emitDeclarationOnly: true

// @fileName: globals.ts
type MissingGlobalType = "global";
declare const missingGlobalValue: "A";
// @fileName: index.ts
// this test assumes there is some global definitions for MissingGlobalType and missingGlobalValue that are not available to transpileDeclaration
export const x: MissingGlobalType = null!;
export const fn = (a: MissingGlobalType): MissingGlobalType => null!;
export const fn2 = (a: MissingGlobalType) => null! as MissingGlobalType;

export const x2: typeof missingGlobalValue = null!;
export const fn3 = (a: typeof missingGlobalValue): typeof missingGlobalValue => null!;
export const fn4 = (a: typeof missingGlobalValue) => null! as typeof missingGlobalValue;


export const o : {
    [missingGlobalValue]: string
} = null!;
```

### 🙁 Actual behavior

`trasnpileDeclaration` will return errors since it can't find `MissingGlobalType` and `missingGlobalValue`

### 🙂 Expected behavior

`trasnpileDeclaration` should not error or missing symbols. It does not have enough information to know if that symbol is resolvable. 

_No response_

weswigham:
These already will get errors in the checker when the file is typechecked, so the declaration emitter doesn't need to double up and also error on them.

This basically allows the declaration emitter to copy inputs with unresolvable names into the output without the output being blocked by a declaration-emitter issued error.

Fixes #58496
Fixes #58535

### DIFF
--- a/tests/baselines/reference/declarationEmitInvalidExport.errors.txt
+++ b/tests/baselines/reference/declarationEmitInvalidExport.errors.txt
@@ -1,14 +1,11 @@
-declarationEmitInvalidExport.ts(4,30): error TS4081: Exported type alias 'MyClass' has or is using private name 'myClass'.
 declarationEmitInvalidExport.ts(5,1): error TS1128: Declaration or statement expected.
 
 
-==== declarationEmitInvalidExport.ts (2 errors) ====
+==== declarationEmitInvalidExport.ts (1 errors) ====
     if (false) {
       export var myClass = 0;
     }
     export type MyClass = typeof myClass;
-                                 ~~~~~~~
-!!! error TS4081: Exported type alias 'MyClass' has or is using private name 'myClass'.
     }
     ~
 !!! error TS1128: Declaration or statement expected.
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -5864,7 +5864,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
 
         // Verify if the symbol is accessible
         return (symbol && hasVisibleDeclarations(symbol, shouldComputeAliasToMakeVisible)) || {
-            accessibility: SymbolAccessibility.NotAccessible,
+            accessibility: SymbolAccessibility.NotResolved,
             errorSymbolName: getTextOfNode(firstIdentifier),
             errorNode: firstIdentifier,
         };
--- a/src/compiler/transformers/declarations.ts
+++ b/src/compiler/transformers/declarations.ts
@@ -324,7 +324,8 @@ export function transformDeclarations(context: TransformationContext) {
             }
             // TODO: Do all these accessibility checks inside/after the first pass in the checker when declarations are enabled, if possible
         }
-        else {
+        // The checker should issue errors on unresolvable names, skip the declaration emit error for using a private/unreachable name for those
+        else if (symbolAccessibilityResult.accessibility !== SymbolAccessibility.NotResolved) {
             // Report error
             const errorInfo = getSymbolAccessibilityDiagnostic(symbolAccessibilityResult);
             if (errorInfo) {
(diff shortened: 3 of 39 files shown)


tests/baselines/reference/declarationEmitInvalidExport.errors.txt
```diff
@@ -1,14 +1,11 @@
...
 
 
-==== declarationEmitInvalidExport.ts (2 errors) ====
+==== declarationEmitInvalidExport.ts (1 errors) ====
     if (false) {
       export var myClass = 0;
     }
     export type MyClass = typeof myClass;
-                                 ~~~~~~~
-!!! error TS4081: Exported type alias 'MyClass' has or is using private name 'myClass'.
```
weswigham, line 11: Declaration emit errors are emitted in the presence of parse errors (unlike checker errors), so this (pretty bogus, given the parse error) error was here before, and now it's gone. If checker errors were reported, there would be one on `myClass` here, but, again, blocked by a parse error anyway.

sheetalkamat (approved):

PR #52686, opened 2023-02-08
Fix completion details auto-imports crashes

andrewbranch:
Fixes #52661

The ExportInfoMap is a multimap that contains information about every way that every export is re-exported from different places. Usually, a single symbol that gets re-exported multiple times has a single key pointing to an array of each place it’s exported from. When auto-imports are displayed in completions, we show one completion per key, and we simply pick the “best” re-exporting location out of that array. This lets us prefer short import paths over deep imports. (When and whether this is actually a good heuristic is beside the point.)

The exception is when a single symbol is exported by an ambient module. Ambient modules are considered a stronger signal that the name of the module is significant. For example, when `"node:fs"` re-exports everything from `"fs"`, we don’t want to be forced to pick which module is the “better” place to import `readFile` from—we’d rather show a completion for both, even though the exported symbol is the same in both. So each of these get their own keys in the multimap.

When `completionEntryDetails` is called and we have to re-find the export we showed in the previous `completionInfo` request, we never used to check the module symbol when searching the ExportInfoMap—instead, we asserted on it later, issuing the message seen in the telemetry bug. The fix is simply to check that the module symbol matches as part of our search instead of returning the first export info where just the export symbol itself matches.

The fix sounds obvious, so why was this not _more_ broken before? Well, when `completionEntryDetails` is called for an import from an ambient module, we were able to take shortcut that skips this search through the ExportInfoMap containing the faulty code. So the only way for the bug to manifest is when we request details for a symbol that a _non-ambient_ module re-exports from an ambient module, _and_ the iteration order mattered—it could have been correct by coincidence sometimes.

PR #58537, opened 2024-05-14
fix55816: exclude files with re-exports if excluded by preferences.autoImportFileExcludePatterns

### LINKED ISSUE #55816: TS is offering no quick fix to implement interface
In this case:

```ts
/*---------------------------------------------------------------------------------------------
 *  Copyright (c) Microsoft Corporation. All rights reserved.
 *  Licensed under the MIT License. See License.txt in the project root for license information.
 *--------------------------------------------------------------------------------------------*/

import { IEditorGroupsService } from 'vs/workbench/services/editor/common/editorGroupsService';

export class EditorParts implements IEditorGroupsService {

}
```

Image

iisaduan:
Fixes #55816

This failure happened when a file `foo.ts` that re-exports some variable `T` is ignored by `"typescript.preferences.autoImportFileExcludePatterns"`. `foo.ts` was properly ignored by `exportMap`, but the autoImports still attempted to generate the import with `foo.ts` instead of the file that originally exported `T`, which lead to a failure. 

This PR allows the `moduleSymbol` found by autoImports to not match the `moduleSymbol` found by `exportInfoMap` if the moduleSymbol found by autoimports is excluded. Furthermore, if the import info still cannot be resolved, the quickfix will still be available and we will generate the implementation code without the import statement.


fatcerberus:
> Fixes https://github.com/microsoft/TypeScript/issues/55816

Is that the right issue?  It seems unrelated.

iisaduan:
Yes it is the correct issue, see the call stack [here](https://github.com/microsoft/TypeScript/issues/55816#issuecomment-1730143944). The quick fix is not offered because getting autoimports fails.

src/services/codefixes/importFixes.ts
```diff
@@ -864,9 +866,14 @@ function codeFixActionToCodeAction({ description, changes, commands }: CodeFixAc
...
     const getChecker = createGetChecker(program, host);
+    const isFileExcluded = preferences.autoImportFileExcludePatterns && getIsFileExcluded(host, preferences);
+    const moduleSymbolExcluded = isFileExcluded && moduleSymbol.declarations && isSourceFile(moduleSymbol.declarations[0]) && isFileExcluded(moduleSymbol.declarations[0]);
     return getExportInfoMap(importingFile, host, program, preferences, cancellationToken)
         .search(importingFile.path, preferCapitalized, name => name === symbolName, info => {
-            if (skipAlias(info[0].symbol, getChecker(info[0].isFromPackageJson)) === symbol && info.some(i => i.moduleSymbol === moduleSymbol || i.symbol.parent === moduleSymbol)) {
+            if (
+                skipAlias(info[0].symbol, getChecker(info[0].isFromPackageJson)) === symbol
+                && info.some(i => moduleSymbolExcluded || i.moduleSymbol === moduleSymbol || i.symbol.parent === moduleSymbol)
```
andrewbranch, line 875: Plot twist: all existing tests pass if you delete this entire `info.some(...)` part of the condition. That might be all that’s needed to fix the bug?

iisaduan: I tried that and wasn't sure if that was the best fix, since I tracked it back to [here](https://github.com/microsoft/TypeScript/pull/52686/files#r1103313654), where it looked like it was intentionally added

iisaduan: Actually, I remember something did break when I tried that-- `tests/cases/fourslash/server/autoImportReExportFromAmbientModule.ts` breaks with the `info.some` call removed. Since exporting ambient modules doesn't use filenames, I figured it would be alright to skip the check for ignored files

andrewbranch:
```suggestion
                && (moduleSymbolExcluded || info.some(i => i.moduleSymbol === moduleSymbol || i.symbol.parent === moduleSymbol))
```

src/services/codefixes/importFixes.ts
```diff
@@ -864,9 +866,14 @@ function codeFixActionToCodeAction({ description, changes, commands }: CodeFixAc
 
 function getAllExportInfoForSymbol(importingFile: SourceFile | FutureSourceFile, symbol: Symbol, symbolName: string, moduleSymbol: Symbol, preferCapitalized: boolean, program: Program, host: LanguageServiceHost, preferences: UserPreferences, cancellationToken: CancellationToken | undefined): readonly SymbolExportInfo[] | undefined {
     const getChecker = createGetChecker(program, host);
+    const isFileExcluded = preferences.autoImportFileExcludePatterns && getIsFileExcluded(host, preferences);
+    const moduleSymbolExcluded = isFileExcluded && moduleSymbol.declarations && isSourceFile(moduleSymbol.declarations[0]) && isFileExcluded(moduleSymbol.declarations[0]);
```
andrewbranch, line 870: I think this needs to be something like “every declaration of `moduleSymbol` is excluded, by filename if SourceFile, or by containing SourceFile filename if a module declaration.”

andrewbranch (changes requested):

src/services/codefixes/importFixes.ts
andrewbranch: I was wrong about this; you can ignore my comment. If a module source file is excluded, augmentations to it in other files aren’t enough to make the module un-excluded.

iisaduan:
@typescript-bot pack this

src/services/codefixes/importFixes.ts
```diff
@@ -286,8 +286,11 @@ function createImportAdderWorker(sourceFile: SourceFile | FutureSourceFile, prog
         const checker = program.getTypeChecker();
         const symbol = checker.getMergedSymbol(skipAlias(exportedSymbol, checker));
         const exportInfo = getAllExportInfoForSymbol(sourceFile, symbol, symbolName, moduleSymbol, /*preferCapitalized*/ false, program, host, preferences, cancellationToken);
+
+        // If no exportInfo is found, this means the export could not be resolved, so we should not generate an import
+        if (!exportInfo) return;
```
andrewbranch, line 291: Maybe we should at least ensure this never happens unless there are `preferences.autoImportFileExcludePatterns`

tests/cases/fourslash/autoImportFileExcludePatterns11.ts
```diff
@@ -0,0 +1,43 @@
...
+//// 	interface Event {
+//// 		c: string;
+//// 	}
+//// }
+
+
+verify.codeFix({
+    description: "Implement interface 'Parts'",
+    newFileContent: 
+`import { Event } from '../event/event';
```
andrewbranch, line 35: Huh, this one with the broken import ended up working? What’s different from when we debugged it yesterday and this was crashing? Is there a test case that exercises `if (!exportInfo) return;`?

iisaduan: Yes, the broken import that we fixed was in `thing,ts`, and the file that the codefix was triggered in is in a different part of the folder tree than `thing`

iisaduan: I was trying to make cases with other file structures, since I was trying to test if the codefix would still crash in situations where the import map resolved to a different preferred file to generate the import from

iisaduan:
I added a debug statement to only exit quietly if `preferences.autoImportFileExcludePatterns` is provided, like @andrewbranch mentioned. It seems that otherwise, if the import was unresolved, autoimports goes through other functions to try to generate imports.

src/services/codefixes/importFixes.ts
```diff
@@ -284,8 +286,14 @@ function createImportAdderWorker(sourceFile: SourceFile | FutureSourceFile, prog
         const checker = program.getTypeChecker();
         const symbol = checker.getMergedSymbol(skipAlias(exportedSymbol, checker));
         const exportInfo = getAllExportInfoForSymbol(sourceFile, symbol, symbolName, moduleSymbol, /*preferCapitalized*/ false, program, host, preferences, cancellationToken);
+        if (!exportInfo) {
+            // If no exportInfo is found, this means export could not be resolved when we have filtered for autoImportFileExcludePatterns,
+            //     so we should not generate an import.
+            Debug.assert(preferences.autoImportFileExcludePatterns!.length > 0);
```
andrewbranch, line 292:
```suggestion
            Debug.assert(preferences.autoImportFileExcludePatterns?.length);
```

andrewbranch (approved):

PR #58538, opened 2024-05-14
Make sure that jsdoc is always in range with parameter declaration

### LINKED ISSUE #56677: Diagnostics and semantic tokens get incorrectly highlighted after doing some specific actions in a line

- VS Code Version: 1.81.1
- OS Version: Windows 11 (10.0.22621)

Steps to Reproduce:

1. Make a new file and set the language to JavaScript or TypeScript.
2. Paste this code:
   ```js
   function bug(
       test // true = test test 123
   ) {}
   ```
3. Put your cursor next to the second slash of the comment.
4. Press backspace.
5. Type `*` twice. (This will automatically extend to `/**| */` as JSDoc comments are marked as auto closing pairs)
6. Move your cursor three characters to the right using the arrow keys (It should be after the closing slash).
7. Press backspace three times.  
   Your code should now look like this:
   ```js
   function bug(
       test /**| true = test test 123
   ) {}
   ```
   Where `|` is the cursor position.
8. Now press `End` (Windows) or `Cmd+RightArrow` (Mac) to go to the end of the line.
9. Type ` */`.
10. Enjoy the bug ;-)

If anyone can't reproduce it, this is the outcome for me:
image

sheetalkamat:
Fixes #56677

### DIFF
--- a/src/compiler/utilities.ts
+++ b/src/compiler/utilities.ts
@@ -2474,6 +2474,7 @@ export function getJSDocCommentRanges(node: Node, text: string) {
         getLeadingCommentRanges(text, node.pos);
     // True if the comment starts with '/**' but not if it is '/**/'
     return filter(commentRanges, comment =>
+        comment.end <= node.end && // Due to parse errors sometime empty parameter may get comments assigned to it that end up not in parameter range
         text.charCodeAt(comment.pos + 1) === CharacterCodes.asterisk &&
         text.charCodeAt(comment.pos + 2) === CharacterCodes.asterisk &&
         text.charCodeAt(comment.pos + 3) !== CharacterCodes.slash);
--- a/src/testRunner/unittests/incrementalParser.ts
+++ b/src/testRunner/unittests/incrementalParser.ts
@@ -120,7 +120,7 @@ function insertCode(source: string, index: number, toInsert: string) {
     }
 }
 
-describe("unittests:: Incremental Parser", () => {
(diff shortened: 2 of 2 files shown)


weswigham (approved):
Funky. I'm not sure why we even parse a comment range on the incremental reparse in this case, since we deleted the leading `/`. The comment range scanner function pretty clearly only matches comments on seeing a `//` or `/*` sequence. The single `/` left after the edit should be interpreted as a division operator or opening to a regex literal, I'd think. I get that the jsdoc nodes we make from these spans, and adjusting them in the incremental parser, is ultimately what triggers the assertion (which is why excluding them fixes it), just feels weird that we have a jsdoc node at all after the edit - I'd assume we'd have rescanned from in front of the jsdoc host node and thrown it out.

jakebailey:
I guess I would have though this were a bug somewhere in `updateTokenPositionsAndMarkElements` in the incremental parser, since we should be marking nodes as need to be fully reparsed...

sheetalkamat:
The issue was that when first source File was created with `function bug(\r\n    test /** */ true = test test 123\r\n) {}` it created a parameter declaration of 0 width just before the jsdoc and jsdoc was attached to that parameter and hence the assert in incremental parsing was failing where it adjusts the spans for nodes in the change range and keeps it around so if parser can reuse node, it will. ( in this case it obviously doesnt as removing that leading `/` makes it expression)

PR #58539, opened 2024-05-14
Attempt to reuse type parameter constraint nodes

### LINKED ISSUE #58511: [transpileDeclaration API][5.5] More missing type-only imports
### 🔎 Search Terms

transpiledeclaration api emit missing import

### ⏯ Playground Link

https://www.typescriptlang.org/dev/bug-workbench/?target=7&ts=5.5.0-dev.20240512#code/PTAEAEGcBcCcEsDG0Bco4FcCmAoEEATLRAGwENYzp4B7AOzU13ygAsaB3AUQFt5o8YNp179oWAgDF4JLGjIA6AguiRBESO259o4qTLmgARkpVr14AGYG6ZHocWqcWAB4AHGrGih4dcbEsyRCxQAEk6UABvHFBQAG0AaywATzQYBDoAcwBdNAw6BLpOOgBuHABfZ3dPb19-QODQAHkMb2jYxJS0uF8cvIKijlKKqo8vdGS3EIAhMkgQgF5QAHIyZdAAHxWjZbKLa1lbezQTJ3geMbaJqbC6ABprkJboB+hJmbmQ8tBLWBoeUAAIgUwDIgL2rkuoEQ9BgoBoRgAVsRvEt2qACDQACqsXoAHixoFc4joBEgoFm8wAfAAKAD6qFAWIedN8aHCcSx2QAlGhnpzslEYrFQLAsNAMLBhrFyncKmUgA

### 💻 Code

```ts
// @strict: true
// @declaration: true
// @showEmit
// @showEmittedFile: a.d.ts
// @showEmittedFile: b.d.ts

// @filename: a.ts
export interface In {
  [key: string]: unknown;
}
export interface Out {
  [key: string]: unknown;
}
export type Base = 'a' | 'b';

// @filename: b.ts
import { type In, type Out, type Base } from "./a";

export const object = {
  doThing<T extends Base>(_t: T, _in: In[T]): Out[T] {
    return;
  },
};
```

### 🙁 Actual behavior

```ts
import { type In } from "./a";
export declare const object: {
    doThing<T extends Base>(_t: T, _in: In[T]): Out;
};
```
```ts
import { type In, type Out, type Base } from "./a";
export declare const object: {
    doThing<T extends Base>(_t: T, _in: In[T]): Out[T];
};
```
(interestingly the Debug tab of the workbench link shows something that's partially wrong)

### 🙂 Expected behavior

Using the API should emit
```ts
import { type In, type Out, type Base } from "./a";
export declare const object: {
    doThing<T extends Base>(_t: T, _in: In[T]): Out[T];
};
```

_No response_

weswigham:
Fixes #58511

We should probably also do defaults, but one thing at a time.


src/compiler/checker.ts
```diff
@@ -8274,7 +8278,6 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             sym = resolveEntityName(leftmost, meaning, /*ignoreErrors*/ true, /*dontResolveAlias*/ true);
             if (
                 context.enclosingDeclaration &&
-                (getNodeLinks(context.enclosingDeclaration).fakeScopeForSignatureDeclaration || !findAncestor(node, n => n === context.enclosingDeclaration)) &&
```
weswigham, line 8277: A fourslash test was affected by the code change and made it clear this small change to https://github.com/microsoft/TypeScript/pull/58516 was needed - checking that the entity name resolves within the enclosing context is just always required to ensure the name's actually available, not just when there's a fake scope.

src/compiler/checker.ts
```diff
@@ -8431,6 +8434,10 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             }
 
             function visitExistingNodeTreeSymbolsWorker(node: Node): Node | undefined {
+                if (isJSDocTypeExpression(node)) {
```
weswigham, line 8437: The constraint type node for a type parameter declared in jsdoc is a `JSDocTypeExpression` which, as it turns out, is not *actually* a type node (according to `isTypeNode`, though the typesystem thinks it is - inconsistent much?), so it needs to get unwrapped; you also definitely don't want the `{}` from jsdoc copied into a `.d.ts` file in the first place! So the assertion triggering off the core of this change found this bug, which was nice.

jakebailey (approved):

PR #58546, opened 2024-05-15
Improve reuse of nodes in signatures with type mapping

dragomirtitian:
Improve reuse of nodes in signatures with type mapping


jakebailey:
@typescript-bot test it

jakebailey:
Used your tool to review all 201 significantly changed files. Some notes:

These are examples of notably good changes (good enough to point out how good they are):

- declarationEmitGlobalThisPreserved.ts
- declarationEmitMappedTypeTemplateTypeofSymbol.ts
- defaultDeclarationEmitShadowedNamedCorrectly.ts
- moduleAugmentationDuringSyntheticDefaultCheck.ts
- recursiveFunctionTypes.ts
- subtypingWithConstructSignatures.ts and co
- unionTypeWithRecursiveSubtypeReduction3.ts

Most changes seem like just good reuse updates, so that's good.

Most of the other changes seem to be a couple of categories:

- The preservation of `undefined` or `null`, where the test has `strictNullChecks` disabled (since it's not the default, ugh). Not totally sure how I feel about these, especially if they're going to show in tooltips? But who the heck has this flag off, why do we have this flag, agh
	- e.g. tests/cases/compiler/copyrightWithNewLine1.ts showing null as return for getElementById
- Objects losing `| undefined` on optional props. Continually worried about this but that's how it goes when copying...

Other stuff:

- `constraintSatisfactionWithAny.ts` - this looks like a bug in the old code, but why don't I see it in dts emit or hover?
- `promisePermutations.ts` and co - seems to be a very large instantiation count increase here.

dragomirtitian:
# `constraintSatisfactionWithAny` 

I investigated and the way the signature is printed is different. The types base-line prints the types using the `GenerateNamesForShadowedTypeParams` flag, while the quick info does not use this flag so no type parameters will be renamed. 

And it's not showing up in declarations, because the type reuse in that case is actually not done by the checker printer, for declarations it's actually done in the declaration transform, which has its own ideas as to how things should be done (I really think we should unify these a lot more than they are today).

Something that would show the bug is `const foo4 = null! as <T extends <T>(x: T) => void>(x: T) => T` ([Playground Link](https://www.typescriptlang.org/play/?ts=5.5.0-dev.20240522#code/MYewdgzgLgBAZiEAWGBeGYCuAbbBCGAQwhgB4AVGAUwA8oqwATEigPgAoaAuGcgSjSsYANxABLRh268BqIeSA)). In this case declaration emit falls back to type printing in the checker with `GenerateNamesForShadowedTypeParams` and we can see that the type parameters are not renamed appropriately.

On a related node. Why is renaming of shadowed parameters needed in this case? Since both `T` definitions come from source at the location, I don't really see how an inlined type  would ever need to reference the outer `T` since the outer `T` was never accessible in the inner scope to being with. Is this a possible improvement or am I missing some case? (cc: @weswigham)


# promisePermutations

I do think most of these instantiations actually hit the instantiation cache, but they get counted anyway. So I don't think the performance impact should be too big.

The reason this happens though is because  when printing the return type I need to instantiate the return type I extract from the declaration and check it against the resolved return type in the signature. This is because  signature return types might be different to the declaration return type (even after mapping). Ideally we should change the cases where these two diverge and then we can remove the check.  I found two cases where this is so:


1. Constructor inherited from the base class.

These constructors signatures actually have the declaration from their original declaration but the return type from their containing class. This is a problem, because if we look at the declaration we get a different return type than the signature. Try as I might I have not been able to find a way, just by looking at the signature, to determine if this is a constructor signature that is inherited (and thus unreliable) or not.

Ex:

```ts
type Constructor<T = {}> = new (...args: any[]) => T;

function Mixin<TBase extends Constructor>(Base: TBase) {
    return class extends Base {
    };
}
```

In the example the constructor signature of `Base` is actually tied to the declaration that comes from `new (...args: any[]) => T`, which will not give the correct return type.

2. Signatures that don't actually have a mapper that can get us from declaration to signature types. For example in

```ts
const createTransform = <I, O>(tr: (from: I) => O) => tr;

function withP2<P>(p: P) {
  const m = <I,>(from: I) => ({ ...from, ...p });
  return createTransform(m);
}

const addP2 = withP2({ foo: 1 });
```

The signature for `addP2`  actually comes from the declaration  `tr: (from: I) => O`. But the mapper for the signature only contains `I -> I, P -> { foo: number }`.  This mapper is incorrect, it does not actually have a mapping for `O` which it should for us to be able to correctly get the return type.

This is because in `instantiateSignature` when creating the new mapper we have:

```ts
freshTypeParameters = map(signature.typeParameters, cloneTypeParameter);
mapper = combineTypeMappers(createTypeMapper(signature.typeParameters, freshTypeParameters), mapper);
```

But `signature` already has it's own mapper which contains a mapping from `O -> I & P`. I tried adding the signature mapper in the mix, but that seems to break a bunch of other things and it's an unrelated change I don't feel comfortable commingling in this PR.

jakebailey:
> I investigated and the way the signature is printed is different. The types base-line prints the types using the `GenerateNamesForShadowedTypeParams` flag, while the quick info does not use this flag so no type parameters will be renamed.

Ah, right. I've wanted to make hover pass that flag, honestly. (related, #55821)

> On a related node. Why is renaming of shadowed parameters needed in this case? Since both `T` definitions come from source at the location, I don't really see how an inlined type would ever need to reference the outer `T` since the outer `T` was never accessible in the inner scope to being with. Is this a possible improvement or am I missing some case?

We've iterated on the type parameter renaming code so much that I'm not sure why it doesn't work, honestly. I definitely had fixed some cases like this before, but we undid some of that, then redid some of that. So it does sound like an improvement, but that may be challenging because even though we introduce a new name `T` into the scope, it's possible that the inside bit could still refer to the outer `T` indirectly and be printed back. At least, that is my guess.

weswigham:
> On a related node. Why is renaming of shadowed parameters needed in this case?

It's a bit more pessimistic than is strictly has to be sometimes, but the renaming is needed because cases like this
```ts
export const a = <T,>() => {
  type U = T;
  return <T extends U>() => null as any as T;
}
```
exist.

weswigham (approved):
I think I'm gonna merge this at this point so we have more time to work on top of it - I like the approach and the changes on the whole look good, and I think that if it doesn't fix some other issues reported with the new node reuse we're doing, it's a required stepping stone for them.

PR #58551, opened 2024-05-16
perf: improve regexp performance by using non-capturing groups

antfu:
This PR is a performance improvement to TS cli/compiler for RegExps of file path matching. Currently, it uses dynamically constructed regex with groups to match multiple path patterns. While all RegExp groups are capturing, but we only use `regex.test(file)` function, which doesn't consume those groups at all.

I am building a RegExp monitor [regex-doctor](https://github.com/antfu/regex-doctor) and found those regexes are taking quite some time to do the matches, even in a very small project. I imagine the cost of running those regexes could scale up with the package size.

For example, in my testing repo, this regex cost a total `8.48ms` to match against 17 paths (and crates a `RegexMatchArray` with 233 unconsumed groups in this case)

Screenshot 2024-05-16 at 17 03 09

After this PR, the time cost to match with the exact same inputs becomes `3.36ms` (60% faster on regex matching).

Screenshot 2024-05-16 at 17 01 48

Can you help to initiate the bot to run a proper benchmark and see how it would help the overall performance? Thanks.

### DIFF
--- a/src/compiler/utilities.ts
+++ b/src/compiler/utilities.ts
@@ -9147,7 +9147,7 @@ const wildcardCharCodes = [CharacterCodes.asterisk, CharacterCodes.question];
 /** @internal */
 export const commonPackageFolders: readonly string[] = ["node_modules", "bower_components", "jspm_packages"];
 
-const implicitExcludePathRegexPattern = `(?!(${commonPackageFolders.join("|")})(/|$))`;
+const implicitExcludePathRegexPattern = `(?!(?:${commonPackageFolders.join("|")})(?:/|$))`;
 
 interface WildcardMatcher {
     singleAsteriskRegexFragment: string;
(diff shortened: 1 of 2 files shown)


jakebailey:
Actually, this PR is perpendicular as it affects dynamically created regexes. Very nice!

jakebailey:
Perf looks unaffected, I'm afraid.

DanielRosenwasser:

antfu:
While I do not fully understand how to read the report, I see it mostly reduces time by a fraction, `~0.5%`, which I think is still valuable considering the change is straightforward. Meanwhile, if you decide that this is not significant to have, also feel free to close the PR, no worries at all

jakebailey:
> I see it mostly reduces time by a tiny fraction ~0.5%, which I think is still valuable considering the change is straightforward

That's for one benchmark out of many; we have so many benchmarks that we are guaranteed to have at least one or two spuriously show as statistically significant (we're p hacking ourselves).

That's not to say this PR doesn't have value, just that the difference may not actually be measurable for most people.

jakebailey (approved):
This kinda slipped through the cracks. I see no reason not to just do this.

PR #58560, opened 2024-05-17
Handle root files listed in project config from referenced project to be same as if they were included through import

### LINKED ISSUE #58156: Upstream file not correctly recognized as output of other project when using path mappings
### 🔎 Search Terms

"typescript not compiling single file unless in file"

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about building

Discovered / tested with 5.4.5.

### ⏯ Playground Link

https://github.com/TimUnderhay/ts-build-references-bug

### 💻 Code

This is a snippet from the minimal repro repo I've created (https://github.com/TimUnderhay/ts-build-references-bug).  The readme contains relevant code / reproduction details.  This really can't be condensed into a snippet or playground link, as far as I can tell.

This is a reproduction of issue https://github.com/microsoft/TypeScript/issues/44845.  When using references within a monorepo to build an app that depends on a shared library reference, certain files get omitted from the tsc output for no evident reason.

## The repo

Under `projects/`, there are two TS 'references', one a shared lib -- 'shared', and an app called 'server'.  Server contains a reference to the shared lib, and also a path alias called `:shared`, pointing to the same place.

`projects/server/tsconfig.json`:
```json
{
  "extends": "../../tsconfig.json",
  "compilerOptions": {
    "baseUrl": "./src",
    "rootDir": "..",
    "lib": ["ES2022"],
    "outDir": "./dist",
    "paths": {
      ":shared/*": [
        "../../shared/src/*"
      ]
    },
    "types": [
      "node"
    ]
  },
  "include": [
    "src/**/*.ts",
    "../shared/src/**/*.ts"
  ],
  "references": [
    {
      "path": "../shared"
    }
  ]
}
```

### Steps to reproduce:

* From repo root, run `npm install`.
* `cd projects/server`
* `npm run build` (is just `rm -rf dist/ && tsc --build -v`)
* Check `projects/server/dist/shared/src` output.  Notice that only the shared logging module was built.  'errors.ts' and 'global-type-overrides.ts' should be there based on the include pattern, but they aren't.

### 🙁 Actual behavior

Check `projects/server/dist/shared/src` output.  Notice that only the shared logging module was built.  'errors.ts' and 'global-type-overrides.ts' should be there based on the include pattern, but they aren't.

### 🙂 Expected behavior

All modules derived from globs specified in `include` should be built.

If one builds `projects/shared` on its own, all output files are present and accounted for in `projects/shared/dist`.  The issue appears to be related to references.  Building `projects/server` with  either`tsc` or `tsc --build -v` appear to yield the same end result -- missing files under `projects/server/dist/shared`.

## It Gets Even Weirder

In `projects/server/src/server.ts`, if you uncomment the line `// import { log } from ':shared/logging.js';`, the entire shared subdir gets omitted from `projects/server/dist`!!

Just. Shouldn't. Happen.

sheetalkamat:
- Always look if file looking is redirect to output of referenced project
- In buildInfo store if root file is not same as resolved path so `tsc --b` can use it to determine correct checks on input hash

Fixes #58156


src/compiler/builder.ts
```diff
@@ -1214,6 +1221,17 @@ function getBuildInfo(state: BuilderProgramState): BuildInfo {
         return root.length = root.length - 1;
     }
 
+    function toResolvedRoot(): ProgramBuildInfoResolvedRoot[] | undefined {
+        let result: ProgramBuildInfoResolvedRoot[] | undefined;
+        rootFileNames.forEach(path => {
+            const file = state.program!.getSourceFileByPath(path);
+            if (file && path !== file.resolvedPath) {
+                (result ??= []).push([toFileId(file.resolvedPath), toFileId(path)]);
```
jakebailey, line 1229: I don't think this pattern is new in the file, but it does seem like we could be using `append` across the board.

jakebailey (approved):
I'm less familiar with this but so take my review with a grain of salt :sweat_smile:

PR #51421, opened 2022-11-06
Fix 46457: Improve autocomplete for union of tuples

rynobax:
Fixes #46457

tbh I have no idea if this is the "right" way to go about fixing this, I used a lot of trial and error to get to these changes.  I'm assuming since all the tests pass it makes some sort of sense, but lmk if there's something fundamentally wrong with this approach.

Issue #55632, opened 2023-09-05
No type inferrence of callback arguments inside a tuple union type


nenikitov:
### 🔎 Search Terms

- discriminated union inference 
- union argument inference 

### 🕗 Version & Regression Information

- This is a compilation error in `tsserver 3.3.2`                                                                  

### ⏯ Playground Link

https://www.typescriptlang.org/play?#code/C4TwDgpgBAkgdgMwgJwILIOZQLwCgpQA+UA2gIwA0UAFAIYBcUcArgLYBGKAlDgHxQA3APYBLACYBdfEVIAmKtXaMAzsGQi4GHtn7DxEgNy4AxkLiqotTI3hI0mHKUo1a2-gG8oAX0O4gA

### 💻 Code

```ts
type InferArg =
  | [1, (a: number) => void]
  | [2, (b: string) => void];
const arg: InferArg = [1, (a) => { }];
```

### 🙁 Actual behavior

I get a compilation error:
> Parameter 'a' implicitly has an 'any' type.(7006)

### 🙂 Expected behavior

`a` should be inferred to be a `number` because I already passed `1` as the first tuple element. Type inference should be identical to this code:
```ts
type InferArg = [1, (a: number) => void];
const arg: InferArg = [1, (a) => { }];
```

jcalz:
I think this is one of the issues that prevents #42987 from working.  (discriminated unions of tuples don't seem to get contextual typing the way discriminated unions of non-tuple objects do)

Andarist:
I had this for a while on my wishlist already - might give it a try soon-ish. 

The problem stems from the fact that `getApparentTypeOfContextualType` doesn't do anything with array literal expressions:
https://github.com/microsoft/TypeScript/blob/3f17500cbf0adcdae388a6d161002ab73971f3bf/src/compiler/checker.ts#L29990-L29992

[...]

PR #58571, opened 2024-05-19
Improve autocomplete for union of tuples

### LINKED ISSUE #46457: Code hint not narrowed properly on union of literal tuple
# Bug Report

### 🔎 Search Terms
- code hint tuple
- code hint union

4.4.4

### ⏯ Playground Link

[Playground link with relevant code](https://www.typescriptlang.org/play?ts=4.4.4#code/DYUwLgBAbghsCuIBcEDaByG6A0F1QEZ0BdCAHzXQCMc8oAmEiAXkq130eIG4g)

### 💻 Code

```ts
let value: ['a', 'v1'] | ['b', 'v2'] = ['a', 'v']; // Type 'v' here and get code hint
```

### 🙁 Actual behavior

image

The first element is `'a'`, so the type of value should be narrowed to `['a', 'v1']`.
So that the code hint of the second element should only have `'v1'`, but actually got `'v1' | 'v2'`.

The compile result is correct, but it is not consistent with code hint.

### 🙂 Expected behavior

image

rynobax:
Fixes #46457 

Another attempt at [this pr](https://github.com/microsoft/TypeScript/pull/51421) from two years ago 😅 

This feels fairly wrong (not sure the proper way to narrow these types, I don't really understand what `getTypeArguments` does), but once again since the tests pass I feel like it might be close?  If anyone could give pointers I can make changes to clean this up, or feel free to modify if you know what you are doing.  Thanks!

The changes to `src/services/stringCompletions.ts` was just me making changes while finding my way around the code, can revert them if you prefer the other formatting.


src/services/utilities.ts
```diff
@@ -3412,6 +3414,39 @@ export function getContextualTypeFromParent(node: Expression, checker: TypeCheck
         }
         case SyntaxKind.CaseClause:
             return getSwitchedType(parent as CaseClause, checker);
+        case SyntaxKind.ArrayLiteralExpression:
```
Andarist, line 3417: I believe that the first step to improving how tuples are autocomplete would be to improve how they are discriminated. I'd start from looking into `discriminateContextualTypeByObjectMembers` and implementing a similar function, smth like `discriminateContextualTypeByElements`

src/compiler/checker.ts
```diff
@@ -31361,6 +31361,27 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
 
+    function discriminateContextualTypeByElements(node: ArrayLiteralExpression, contextualType: UnionType) {
+        const key = `D${getNodeId(node)},${getTypeId(contextualType)}`;
+        const cached = getCachedType(key);
+        if (cached) return cached;
+
+        // Create a discriminator for each element in the list
+        const discriminators = map(
+            node.elements,
+            (e, ndx) => [() => getContextFreeTypeOfExpression(e), "" + ndx as __String] as const,
```
rynobax, line 31372: In the object version of this function there is a check for `isPossiblyDiscriminantValue`.  Is that necessary here?

src/services/utilities.ts
rynobax: implemented this using discriminateContextualTypeByElements, it feels a lot less hacky!  There's some failing tests that I am looking into, but lmk if this seems off track in any way

src/compiler/checker.ts
Andarist: I'd expect both functions to be very similar so I'd say - yes. if `isPossiblyDiscriminantValue` is needed there - it's also needed here.

src/compiler/checker.ts
```diff
@@ -26900,6 +26900,20 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         return propType && getConstituentTypeForKeyType(unionType, propType);
     }
 
+    function getMatchingUnionConstituentForArrayLiteral(unionType: UnionType, node: ArrayLiteralExpression) {
+        const resolvedElements = node.elements.map(el => getContextFreeTypeOfExpression(el));
+        for (const type of unionType.types) {
+            if (!isTupleType(type)) continue;
```
Andarist, line 26906: q: shouldn't this exit the whole process? Or at least when `!isTupleLikeType(type)`?

src/compiler/checker.ts
```diff
@@ -31361,6 +31375,28 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         );
     }
 
+    function discriminateContextualTypeByElements(node: ArrayLiteralExpression, contextualType: UnionType) {
```
Andarist, line 31378: food for thought: `length` should also be treated as a potential discriminant **but** it probably should be ignored when requesting the contextual type for completions. You could try to use `ContextFlags.Completions` for that or you might have to introduce a new context flag for this purpose, smth like `ContextFlags.NoTupleBounds`.

src/compiler/checker.ts
```diff
@@ -26900,6 +26900,20 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         return propType && getConstituentTypeForKeyType(unionType, propType);
     }
 
+    function getMatchingUnionConstituentForArrayLiteral(unionType: UnionType, node: ArrayLiteralExpression) {
+        const resolvedElements = node.elements.map(el => getContextFreeTypeOfExpression(el));
+        for (const type of unionType.types) {
+            if (!isTupleType(type)) continue;
+            const unionElements = getElementTypes(type);
+            if (unionElements.length !== resolvedElements.length) continue;
```
Andarist, line 26908: This is not quite right. You need to experiment more with this and try to create multiple test cases mixing tuples with different fixed sizes and also the ones with variadic sizes. It would probably be the easiest to only reason about the *leading* fixed elements.

rynobax: https://github.com/microsoft/TypeScript/pull/58571/commits/79a4a79d1b3cb2567642d7e5bc40e276c1bfc23a

i didn't see an obvious way to determine the number of leading elements so i removed that early exit

src/compiler/checker.ts
rynobax: not sure I understand this question.  if the union constituent is not a tuple, it's not valid, but I don't think that means other union constituents couldn't be, so i think we should continue looping?

src/compiler/checker.ts
rynobax: removed the length check in `getMatchingUnionConstituentForArrayLiteral` so not sure if this is still relevant

src/compiler/checker.ts
Andarist: TS is structural and you can create some weird (acceptable) types like `[number, "foo"] | { "1": "bar" }`. This could lead to rejecting some types that should still be considered here. So the way to ignore the problem would be to just (for now?) exit this process.

`isTupleLikeType` point stands, I think. Types like `{ "0": "foo" }` should contribute to autocompletes

src/compiler/checker.ts
Andarist: it's relevant because this `arg` could contextually be typed as `number`
```ts
const tup: [(arg: number) => void] | [(arg: string) => void, number] = [
  (arg) => {},
];
```

src/compiler/checker.ts
rynobax: I was confused when I left my last comment, was thinking this function was something else.  I'm realizing I don't even understand the purpose of `getMatchingUnionConstituentForObjectLiteral` or `getMatchingUnionConstituentForArrayLiteral`.  Do you know what this function should be doing?  I'm just going to remove it from the PR for now, can add something back if it makes sense.

src/compiler/checker.ts
```diff
@@ -23842,7 +23842,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             let matched = false;
             for (let i = 0; i < types.length; i++) {
                 if (include[i]) {
-                    const targetType = getTypeOfPropertyOrIndexSignatureOfType(types[i], propertyName);
+                    const targetType = typeof propertyName === "number"
+                        ? getContextualTypeForElementExpression(types[i], propertyName)
+                        : getTypeOfPropertyOrIndexSignatureOfType(types[i], propertyName);
```
rynobax, line 23847: found that this was not working like I wanted with variadic tuple types, so for tuples im using `getContextualTypeForElementExpression`.  Not sure if there's a better way to do this, also considered making this function an argument

fwolff:
I just wrote a fix for a related [issue](https://github.com/microsoft/TypeScript/issues/55632): it still needs improvements (see @Andarist comments [here](https://github.com/fwolff/TypeScript/commit/9eecc3035f4fc82ed994fbb8ce92fa8b9c3cc075)) and, while it fixes a compilation error, I have no idea if it helps solving your [code completion issue](https://github.com/microsoft/TypeScript/issues/46457).

Is this PR going to be merged soon or later? Does it solves this compilation [error](https://github.com/microsoft/TypeScript/issues/55632) as well? Should I continue to work on a (possibly) concurrent PR or help you in any way?

PR #58592, opened 2024-05-20
Skip typechecking file when generating declaraiton to get d.ts signature for incremental build

sheetalkamat:
- Skips typechecking when emitting d.ts for determining if signature of the file has changed.
- Sets cancellation token for emitting files to ensure that it is cancellable esp since files may not be type checked


tests/baselines/reference/tsbuildWatch/demo/updates-with-bad-reference.js
```diff
@@ -615,11 +615,11 @@ Output::
       "../../animals/index.ts": {
         "original": {
           "version": "-7220553464-import Animal from './animal';\n\nexport default Animal;\nimport { createDog, Dog } from './dog';\nexport { createDog, Dog };\n",
-          "signature": "1096904574-import Animal from './animal';\nexport default Animal;\nimport { createDog, Dog } from './dog';\nexport { createDog, Dog };\n",
+          "signature": "2198076214-export default Animal;\nexport { createDog, Dog };\n",
```
sheetalkamat, line 618: This is first example of incorrect d.ts emit in the signature, Means when error is fixed, we will re-compile which is ok in this case i think. But would be great to check if we can preserve the behavior

tests/baselines/reference/tsc/incremental/change-to-modifier-of-class-expression-field-with-declaration-emit-enabled.js
```diff
@@ -241,11 +241,11 @@ exitCode:: ExitStatus.DiagnosticsPresent_OutputsSkipped
       "./messageableperson.ts": {
         "original": {
           "version": "3462418372-const Messageable = () => {\n    return class MessageableClass {\n        protected message = 'hello';\n    }\n};\nconst wrapper = () => Messageable();\ntype MessageablePerson = InstanceType<ReturnType<typeof wrapper>>;\nexport default MessageablePerson;",
-          "signature": "-21450256696-declare const wrapper: () => {\n    new (): {\n        message: string;\n    };\n};\ntype MessageablePerson = InstanceType<ReturnType<typeof wrapper>>;\nexport default MessageablePerson;\n(116,7)Error4094: Property 'message' of exported class expression may not be private or protected.",
+          "signature": "-2474549887-export default MessageablePerson;\n",
```
sheetalkamat, line 244: This is the main issue: This one doesnt report error so we are not able to know that there is no change in signature, and we would need to rebuild.

tests/baselines/reference/tsbuildWatch/demo/updates-with-bad-reference.js
sheetalkamat: @weswigham  this one too if possible.

sheetalkamat:
@weswigham i tried passing noCheck equivalent when running emit for dts emit and i see few tests failing. i have commented on each but if you can take a look and see what we can do for those scenarios that would be great. Thanks

tests/baselines/reference/tsbuildWatch/demo/updates-with-bad-reference.js
weswigham: Import links are collected by the `collectLinkedAliases` call in the emitter, so the condition on calling it needs to be toggled when you pass this into the emit resolver, as it is for `noCheck`.

weswigham: To elaborate, `emitDeclarationFileOrBundle` in `emitter.ts` needs to swap
``ts
        if ((emitOnly && !getEmitDeclarations(compilerOptions)) || compilerOptions.noCheck) {
``
to
```ts
        if ((emitOnly && !getEmitDeclarations(compilerOptions)) || compilerOptions.noCheck || forceDtsEmit) {
```

weswigham: It's pretty likely doing so will fix the other discrepancies you've seen here, since they all seem to do with alias elision.

sheetalkamat: That did it. Thanks

src/testRunner/unittests/tsc/cancellationToken.ts
```diff
@@ -17,7 +17,7 @@ import {
     libFile,
 } from "../helpers/virtualFileSystemWithWatch.js";
 
-describe("unittests:: tsc:: builder cancellationToken", () => {
+describe("unittests:: tsc:: builder cancellationToken:: sheetal", () => {
```
weswigham, line 20:
```suggestion
describe("unittests:: tsc:: builder cancellationToken", () => {

```

XD

sheetalkamat: Sorry that was for debugging and left over

jakebailey:
Is this the winning PR or is #58593 still something to consider? Just checking before I review...

src/compiler/checker.ts
```diff
@@ -47578,13 +47578,13 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         }
     }
 
-    function getDiagnostics(sourceFile: SourceFile, ct: CancellationToken): Diagnostic[] {
+    function getDiagnostics(sourceFile: SourceFile, ct: CancellationToken, forceDtsEmit?: boolean): Diagnostic[] {
```
weswigham, line 47581: Maybe rename the parameter to `skipDiagnostics` from here on in, since it's no longer really concerned with emit, per sey, at this point?

src/testRunner/unittests/tsc/cancellationToken.ts
```diff
@@ -17,7 +17,7 @@ import {
     libFile,
 } from "../helpers/virtualFileSystemWithWatch.js";
 
-describe("unittests:: tsc:: builder cancellationToken", () => {
+describe("unittests:: tsc:: builder cancellationToken::", () => {
```
weswigham, line 20:
```suggestion
describe("unittests:: tsc:: builder cancellationToken", () => {

```

weswigham (commented):
> Is this the winning PR or is https://github.com/microsoft/TypeScript/pull/58593 still something to consider? Just checking before I review...

They don't appear mutually exclusive to my eye. This one skips the diagnostic pass when calculating signatures. The other skips it when a `//@ts-nocheck` comment is in a file.

src/testRunner/unittests/tsc/cancellationToken.ts
sheetalkamat: Most of our unit tests end with :: so we can run them with that id easily .. so will keep this change

src/compiler/checker.ts
```diff
@@ -2428,10 +2428,10 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         return visitEachChild(node, markAsSynthetic, /*context*/ undefined);
     }
 
-    function getEmitResolver(sourceFile: SourceFile, cancellationToken: CancellationToken) {
+    function getEmitResolver(sourceFile: SourceFile, cancellationToken: CancellationToken, forceDtsEmit?: boolean) {
         // Ensure we have all the type information in place for this file so that all the
         // emitter questions of this resolver will return the right information.
-        getDiagnostics(sourceFile, cancellationToken);
+        if (!forceDtsEmit) getDiagnostics(sourceFile, cancellationToken);
```
weswigham, line 2434: 10/10 - last night I realized that having the check be dynamic with the `EmitResolver` but fed into `getDiagnostics` meant that we could cache the dummy `NodeFlags.TypeChecked` flag for files we'd force-skipped diagnostics for, but otherwise still need to check later, and unless you discarded the checker, that wasn't un-set-able. By checking this flag up here instead, we avoid setting the `TypeChecked` flag on a `SourceFile` we may still want to actually get diagnostics for later (but still skip the diagnostics walk, if not caching that skip), which is much better.

weswigham (approved):

weswigham (commented):
Oh, one caveat though - https://github.com/microsoft/TypeScript/pull/58364 isn't in yet, so JS emit is unsafe when you skip diagnostics like this. Is an emit resolver with `forceDtsEmit` set guaranteed to be *only* emitting `.d.ts` files?

sheetalkamat:
Yes it is guaranteed to be dts only emit as "forceDtsEmit" is internal on program.emit and used only by builder

jakebailey:
Isn't it public via `getEmitOutput` #34495? Or does that not matter?

sheetalkamat:
LS.. compile on save.. forgot about that. :( i will upate the condition for now to check if its emitOnlyDts and forceDtsEmit to be true which we can later  update in https://github.com/microsoft/TypeScript/pull/58364 to remove emitOnlyDts part.

Issue #56956, opened 2024-01-05
Incremental compile can spend time type checking files with `@ts-nocheck`


wnayes:
### 🔎 Search Terms

incremental, nocheck, `@ts-nocheck`, tsbuildinfo

### 🕗 Version & Regression Information

- This is the behavior in every version I tried (just latest 5.3.3 but I doubt it is new).

### 💻 Code

The TypeScript compiler, during a subsequent `--incremental` compile, can end up doing semantic analysis of source files that are marked with `@ts-nocheck`. This is unexpected and different from a non-incremental compile.

I have not observed error diagnostics from the nocheck file actually get printed out from a compile, however the behavior can be quite noticeable performance-wise if a significant amount of code has `@ts-nocheck` and undergoes incremental compile. 

Specifically, `checkSourceFileWorker(node)` processing can occur during incremental compiles for a source file that had `@ts-nocheck` (or technically, `checkJsDirective.enabled == false`). It seems this is unique to incremental compiles. It looks to me like a nocheck file can end up being iterated through via `getNextAffectedFile`, and when processing occurs on it, there is no `checkJsDirective.enabled` check. Only `getBindAndCheckDiagnosticsForFileNoCache` checks `checkJsDirective.enabled`, which doesn't appear to be a code path used for incremental checking.

It looks like `ts.skipTypeChecking` is checked in more places than `checkJsDirective.enabled`. I happened to confirm that the extra type checking work is avoided if `ts.skipTypeChecking` were to be edited to include such a check, so potentially that is relevant to a possible fix.

-----

I don't think I can create a Playground link for this, because the issue to observe is basically "extra processing only during incremental compile."

I created this repo with a small project and some example traces that demonstrate the issue:
https://github.com/wnayes/incremental-no-check

The gist of it is that the `large.ts` file, marked with `@ts-nocheck`, can end up having CPU time spent type checking it during incremental compiles. There are instructions to reproduce the issue, however the "proof" is basically this line in some traces that I committed:

https://github.com/wnayes/incremental-no-check/blob/main/traces-2/trace.json#L318

There (as well as on many following lines) we can see semantic analysis being done on large.ts, yet the workflow never involves removing `@ts-nocheck`.

### 🙁 Actual behavior

Incremental compiles can spend CPU time doing semantic analysis on `@ts-nocheck` files, which seems wasteful.

### 🙂 Expected behavior

Incremental compiles shouldn't do semantic analysis on `@ts-nocheck` files.

[...]

PR #58593, opened 2024-05-20
Mark file as skips typechecking if it contains ts-nocheck

sheetalkamat:
This is different take on #58592 to skip checking only if no-checkJs 
Frankly i dont know what all will fail because we dont have enough coverage for js + incremental as such from this perspective. 

cc: @weswigham 

From investigating #56956

Needs https://github.com/microsoft/TypeScript/pull/58364 to work correctly for js emit


src/compiler/checker.ts
```diff
@@ -47511,10 +47511,10 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     }
 
     // Fully type check a source file and collect the relevant diagnostics.
-    function checkSourceFileWorker(node: SourceFile) {
+    function checkSourceFileWorker(node: SourceFile, skipCheck: boolean | undefined) {
         const links = getNodeLinks(node);
         if (!(links.flags & NodeCheckFlags.TypeChecked)) {
-            if (skipTypeChecking(node, compilerOptions, host)) {
+            if (skipCheck || skipTypeChecking(node, compilerOptions, host)) {
```
weswigham, line 47517: Personally, for this one, I'd just modify the implementation of the `skipTypeChecking` function to look for the `ts-nocheck` comment, in addition to its current conditions, rather than passing through all this context.

sheetalkamat: But if i move that into skipTypeChecking, it will disable suggestion diagnostics for files with "ts-noCheck" may be thats ok but just making sure it is intended.

sheetalkamat:
@typescript-bot test this

sheetalkamat:
@typescript-bot test it

src/compiler/checker.ts
weswigham: Yeah, but being included in this condition would *also* skip generating suggestions, anyway. Which is, imo, the point - you either do the full diagnostic pass and get suggestions or you don't and... don't. Anyway, keeping it all inside that one `skipTypeChecking` function simplifies things when other places need to ask "hey, is this file typechecked?", which should honestly be the condition in `emitter.ts`, rather than things-which-approximate-that.

sheetalkamat:
Does this need to wait for https://github.com/microsoft/TypeScript/pull/58364 given the ts-nocheck files may be emitting js?

weswigham:
> Does this need to wait for https://github.com/microsoft/TypeScript/pull/58364 given the ts-nocheck files may be emitting js?

Either that or, like the other PR, only enabled when only `.d.ts` emit is performed by the checker in question.

sheetalkamat:
> Either that or, like the other PR, only enabled when only .d.ts emit is performed by the checker in question.

Thats tough I think and not work flowing information temporarily till we get the PR in so i will just wait for that PR to merge.

jakebailey:
#58364 is in, so is this safe to update and get in?

weswigham:
Yeah, just remember the condition in `calculateNodeCheckFlagWorker` in `checker.ts` and `emitJsFileOrBundle` in `emitter.ts` needs to swap from just `compilerOptions.noCheck` to something that can handle this, too. (probably just the `skipTypeChecking` util function this edits, unless you wanna do the piecemeal thing `emitDeclarationFileOrBundle` does - but that should probably just swap to `skipTypeChecking`, too)

sheetalkamat:
@weswigham pls review again. I have merged with main and added those checks

src/compiler/checker.ts
```diff
@@ -49029,7 +49030,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     }
 
     function calculateNodeCheckFlagWorker(node: Node, flag: LazyNodeCheckFlags) {
-        if (!compilerOptions.noCheck) {
+        if (!compilerOptions.noCheck || !canIncludeBindAndCheckDiagnsotics(getSourceFileOfNode(node), compilerOptions)) {
```
weswigham, line 49033:
```suggestion
        if (!compilerOptions.noCheck && !canIncludeBindAndCheckDiagnsotics(getSourceFileOfNode(node), compilerOptions)) {

```

This difference will only surface as a performance issue, since we'll calculate flags when we don't need to :)

src/compiler/emitter.ts
```diff
@@ -859,15 +865,19 @@ export function emitFiles(resolver: EmitResolver, host: EmitHost, targetSourceFi
...
-            // Do that here when emitting only dts files
-            filesForEmit.forEach(collectLinkedAliases);
-        }
+        // Checker wont collect the linked aliases since thats only done when declaration is enabled and checking is performed.
+        // Do that here when emitting only dts files
+        filesForEmit.forEach(sourceFile => {
+            if (
+                (emitOnly && !getEmitDeclarations(compilerOptions)) ||
+                compilerOptions.noCheck ||
+                emitResolverSkipsTypeChecking(emitOnly, forceDtsEmit) ||
```
weswigham, line 874: Aside, since it's not related to this PR, we probably wanna feed this to the JS emit-related conditions we just edited here if we ever intend for the `skipDiagnostics`/`forceDtsEmit` `EmitResolver` parameter to be usable with arbitrary JS emit modes, too. Don't think it matters right now, tho, since it's only used for .d.ts emit afaik.

src/compiler/utilities.ts
```diff
@@ -10082,7 +10082,28 @@ export function skipTypeChecking(sourceFile: SourceFile, options: CompilerOption
     return (options.skipLibCheck && sourceFile.isDeclarationFile ||
         options.skipDefaultLibCheck && sourceFile.hasNoDefaultLib) ||
         options.noCheck ||
-        host.isSourceOfProjectReferenceRedirect(sourceFile.fileName);
+        host.isSourceOfProjectReferenceRedirect(sourceFile.fileName) ||
+        !canIncludeBindAndCheckDiagnsotics(sourceFile, options);
+}
+
+/** @internal */
+export function canIncludeBindAndCheckDiagnsotics(sourceFile: SourceFile, options: CompilerOptions) {
```
weswigham, line 10090: Observation: it seems like every caller of `canIncludeBindAndCheckDiagnsotics` writes `!canIncludeBindAndCheckDiagnsotics(...)`. Perhaps it should be inverted and be `shouldExcludeBindAndCheckDiagnsotics`?

weswigham (changes requested):

src/compiler/utilities.ts
sheetalkamat: Not all any more so i will let this be.

src/compiler/emitter.ts
sheetalkamat: I didnt do this because dont you need to handle that at `calculateNodeCheckFlagWorker` and didnt want to store if calling from emitResolver etc there as there is no easy way to flow that information.

sheetalkamat:
@weswigham pls review so i can get this in for 5.5

weswigham (approved):

Issue #57310, opened 2024-02-06
incremental compilation does not handle const enum properly


godmiaozi:
### 🔎 Search Terms

incremental 
const enum

### 🕗 Version & Regression Information

4.9.5

### 💻 Code

step1, assume we have the following code
a.ts
```ts
import {A} from "./c"
let a = A.ONE
```
b.d.ts
```
export const enum A {
         ONE = 1
 }
```

c.ts
```
import {A} from "./b"
let b = A.ONE
export {A}
```
cmdline 
```
tsc --incremental a.ts --tsBuildInfoFile ./buildFile
```
output 
a.js
```
"use strict";
exports.__esModule = true;
var a = 1 /* A.ONE */;
```
c.js
```
"use strict";
exports.__esModule = true;
var b = 1 /* A.ONE */;
```
step2, then modify b.d.ts
```
export const enum A {
         ONE = 2
 }
```
run cmdline
```
tsc --incremental a.ts --tsBuildInfoFile ./buildFile
```
step3, modify b.d.ts again
```
export const enum A {
         ONE = 3
 }
```
run cmdline
```
tsc --incremental a.ts --tsBuildInfoFile ./buildFile
```
the output is not correct
output:
a.js
```
"use strict";                                                                              
exports.__esModule = true;  
var a = 2 /* A.ONE */;
```
c.js
```
"use strict";
exports.__esModule = true;
var b = 3 /* A.ONE */
```

### 🙁 Actual behavior

output of step3
a.js
```
"use strict";                                                                              
exports.__esModule = true;  
var a = 2 /* A.ONE */;
```
c.js
```
"use strict";
exports.__esModule = true;
var b = 3 /* A.ONE */
```

### 🙂 Expected behavior

output of step3
a.js
```
"use strict";                                                                              
exports.__esModule = true;  
var a = 3 /* A.ONE */;
```
c.js
```
"use strict";
exports.__esModule = true;
var b = 3 /* A.ONE */
```

RyanCavanaugh:
TypeScript 4.9 is 18 months old - can you please try on latest?

zhuoli72:
Based on the latest (86a16636cac9599c8752f8588f4b9518860b47ff) of master branch, it still has the same issue

godmiaozi:
> Based on the latest ([86a1663](https://github.com/microsoft/TypeScript/commit/86a16636cac9599c8752f8588f4b9518860b47ff)) of master branch, it still has the same issue

[...]

PR #58594, opened 2024-05-20
If module exports const enum, invalidate js files along with dts as it can impact js emit as well

sheetalkamat:
Fixes https://github.com/microsoft/TypeScript/issues/57310


src/compiler/builder.ts
```diff
@@ -867,14 +883,20 @@ function handleDtsMayChangeOfReferencingExportOfAffectedFile(
     }
 
     const seenFileAndExportsOfFile = new Set<string>();
+    // If exported const enum, we need to ensure that js files are emitted as well since the const enum value changed
+    const invalidateJsFiles = !!affectedFile.symbol?.exports && !!forEachEntry(
+        affectedFile.symbol.exports,
+        exported => (skipAlias(exported, state.program!.getTypeChecker()).flags & SymbolFlags.ConstEnum) !== 0,
```
andrewbranch, line 889: I see you’re skipping the alias to handle

```ts
declare const enum AWorker {
  ONE = 1
}
export { AWorker as A };
```

but this would also return true for

```ts
export { AWorker as A } from "./some-other-module";
```

Did you intend to return true even if the const enum declaration is in another file?

andrewbranch (approved):

Issue #58533, opened 2024-05-14
Isolated declarations uses type information when emitting computed object keys 


lucacasonato:
### 🔎 Search Terms

- isolated declarations
- computed keys
- object literals

### 🕗 Version & Regression Information

5.5.0-dev.20240514

### ⏯ Playground Link

https://www.typescriptlang.org/play/?isolatedDeclarations=true&ts=5.5.0-dev.20240514#code/KYDwDg9gTgLgBAYwgOwM7wNLAJ5wLxwBEMAFsFtoQNwBQNAJsAgDYCGUwcA5sxAEatmcAN4044uKEixEKdHADivAcwr4ipYEv6CKhMRKnR4SNPACqyAJYBHAK7AAytgC2fCMwBccO9fudUV3dmWglJcGNZMzgAORRLWwdnNw9vQJSQmgBfOiMZU3kAQXVRMIBtCgBdbxgoBwAaA3Ey5OCAOhgIR1qrZC4AFVYuargAM0FUYEbyhP9WjxHx5knpiTLtFSrvJZXsmiA

### 💻 Code

```ts
export const Key = "theKey";

declare global {
    export const GlobalKey = "theGlobalKey"
    export const UniqueSymbol: unique symbol;
    export const NonUniqueSymbol: symbol;
}

export const A = {
    [Key]: true,
    [Symbol.toStringTag]: false,
    [UniqueSymbol]: false,
    [GlobalKey]: false,
}
```

### 🙁 Actual behavior

TypeScript is emitting the computed property keys with type information that can not be determined without a type checker.

```ts
export declare const Key = "theKey";
declare global {
    export const GlobalKey = "theGlobalKey";
    export const UniqueSymbol: unique symbol;
    export const NonUniqueSymbol: symbol;
}
export declare const A: {
    theKey: boolean;
    [Symbol.toStringTag]: boolean;
    [UniqueSymbol]: boolean;
    theGlobalKey: boolean;
};
```

### 🙂 Expected behavior

```ts
export declare const Key = "theKey";
declare global {
    export const GlobalKey = "theGlobalKey";
    export const UniqueSymbol: unique symbol;
    export const NonUniqueSymbol: symbol;
}
export declare const A: {
    [Key]: true,
    [Symbol.toStringTag]: false,
    [UniqueSymbol]: false,
    [GlobalKey]: false,
};
```

weswigham:
[...]

PR #58596, opened 2024-05-20
Lock down computed names in object literals and classes under --isolatedDeclarations

### LINKED ISSUE #58490: [isolatedDeclarations][5.5] Disagreement between transpileDeclaration API and typechecker on Symbol.iterator as computed property
### 🔎 Search Terms

isolated declarations transpileDeclaration symbol iterator

- nightly

### ⏯ Playground Link

https://tsplay.dev/wXLvLw

### 💻 Code

```ts
export class Foo implements Iterable<number> {
  public [Symbol.iterator]() {
    return {
      next: () => ({ value: 42, done: false }),
    };
  }
}
```

### 🙁 Actual behavior

`ts.transpileDeclaration` has diagnostic: `Computed properties must be number or string literals, variables or dotted expressions with --isolatedDeclarations.`

`tsc --isolatedDeclarations` completes successfully

### 🙂 Expected behavior

I'm not sure which is more correct here. If we treat this as not an error, then that seems to imply that `Symbol.iterator` is treated specially, which runs into weird edge cases like `Symbol` being overridden in the current scope. But if we do treat this as an error, then the question becomes how does one add an explicit type annotation for this case?

_No response_

weswigham:
Fixes #58490 - an error is now *always* emitted on computed names in classes and object literals under `--isolatedDeclarations`. Workarounds and carve-outs for builtin symbols, some new syntactic form, or locally bound unique symbols are followups~

cc @dragomirtitian


src/compiler/diagnosticMessages.json
```diff
@@ -7014,6 +7014,10 @@
         "category": "Error",
         "code": 9037
     },
+    "Computed property names on class or object literals cannot be inferred with --isolatedDeclarations.": {
```
jakebailey, line 7017: Does this diag need to be added to the fixer?

weswigham: Since it can have the same related span attached, I suppose so!

src/compiler/expressionToTypeNode.ts
```diff
@@ -351,7 +351,7 @@ export function createSyntacticTypeNodeBuilder(options: CompilerOptions, resolve
             }
             else if (prop.name.kind === SyntaxKind.ComputedPropertyName) {
                 const expression = prop.name.expression;
-                if (!isPrimitiveLiteralValue(expression, /*includeBigInt*/ false) && !isEntityNameExpression(expression)) {
+                if (!isPrimitiveLiteralValue(expression, /*includeBigInt*/ false)) {
```
dragomirtitian, line 354: There is also code in `typeFromObjectLiteral` that is not effectively unused. `isNonNarrowedBindableName` also probably becomes useless since we can now syntactically determine if the computed property is valid so it can probably be removed from the resolver.

dragomirtitian:
@weswigham  Also fixes #58533

jakebailey (approved):

PR #58597, opened 2024-05-20
Respect `importModuleSpecifierPreference` in sort order between fixes for the same symbol from different files

### LINKED ISSUE #55863: Auto-import not using up TypeScript path alias when using barrel export (index.ts)
Type: <b>Bug</b>

`tsconfig.ts`:
```
{
  "compilerOptions": {
    "allowJs": false,
    "baseUrl": "./src",
    "esModuleInterop": true,
    "forceConsistentCasingInFileNames": true,
    "isolatedModules": true,
    "jsx": "react-jsx",
    "lib": ["dom", "dom.iterable", "esnext"],
    "module": "esnext",
    "moduleResolution": "node",
    "noEmit": true,
    "noFallthroughCasesInSwitch": true,
    "outDir": "dist",
    "paths": {
      "~/api": ["api/index.ts"],
      "~/assets/*": ["assets/*"],
      "~/components": ["components/index.ts"],
      "~/constants/*": ["constants/*"],
      "~/hooks": ["hooks/index.ts"],
      "~/pages/*": ["pages/*"],
      "~/router": ["router/index.tsx"],
      "~/schemas": ["schemas/index.ts"],
      "~/store": ["store/index.ts"],
      "~/store/*": ["store/*"],
      "~/styles/*": ["styles/*"],
      "~/translations/*": ["translations/*"],
      "~/types": ["types/index.ts"],
      "~/utils": ["utils/index.ts"]
    },
    "resolveJsonModule": true,
    "skipLibCheck": true,
    "strict": true,
    "target": "ESNEXT"
  }
}
```
1. When I type `useModal`, I get the import suggestion `hooks/useModal`. I would expect the suggestion `~/hooks` instead, as this function is exported from `hooks/index.ts` as specified in my tsconfig file.
2. When I type `EMAIL`, I get the expected import suggestion `~/constants/regex`. The difference in behavior seems to be caused by using barrel export files (re-exporting all submodules in index.ts).

The exports in all `index.ts` files are specified as e.g. `export * from './useModal'`. It seems to make no difference when I change it to `export { useModal } from './useModal'`. I have also tried enabling `"typescript.preferences.importModuleSpecifier": "non-relative"` which makes no difference.  
Disabling all extensions has no effect.

Is there a way to solve this problem? It's annoying currently to always update the import manually.

---

VS Code version: Code 1.82.0 (8b617bd08fd9e3fc94d14adb8d358b56e3f72314, 2023-09-06T22:08:29.670Z)
OS version: Darwin arm64 22.4.0
Modes:

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Apple M2 (8 x 24)|
|GPU Status|2d_canvas: enabled<br>canvas_oop_rasterization: enabled_on<br>direct_rendering_display_compositor: disabled_off_ok<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>opengl: enabled_on<br>rasterization: enabled<br>raw_draw: disabled_off_ok<br>video_decode: enabled<br>video_encode: enabled<br>vulkan: disabled_off<br>webgl: enabled<br>webgl2: enabled<br>webgpu: enabled|
|Load (avg)|6, 4, 4|
|Memory (System)|16.00GB (0.14GB free)|
|Process Argv|--crash-reporter-id 4debfa25-ee5c-42dd-bb8b-59d36f34e883|
|Screen Reader|no|
|VM|0%|
</details><details><summary>Extensions (20)</summary>

Extension|Author (truncated)|Version
---|---|---
TabOut|alb|0.2.2
increment-selection|alb|0.2.0
laravel-extra-intellisense|ami|0.6.3
vscode-intelephense-client|bme|1.9.5
vscode-tailwindcss|bra|0.10.0
laravel-goto-view|cod|1.3.9
vscode-eslint|dba|2.4.2
gitlens|eam|14.3.0
vscode-great-icons|emm|2.1.102
prettier-vscode|esb|10.1.0
find-unused-exports|iul|2.0.1
template-string-converter|meg|0.6.1
vscode-docker|ms-|1.26.0
playwright|ms-|1.0.15
laravel-blade|one|1.34.0
laravel-pint|ope|1.1.4
vscode-import-cost|wix|3.3.0
change-case|wma|1.0.0
pretty-ts-errors|yoa|0.5.2
markdown-all-in-one|yzh|3.5.1

(1 theme extensions excluded)

</details><details>
<summary>A/B Experiments</summary>

```
vsliv368cf:30146710
vsreu685:30147344
python383:30185418
vspor879:30202332
vspor708:30202333
vspor363:30204092
vslsvsres303:30308271
vserr242cf:30382550
pythontb:30283811
vsjup518:30340749
pythonptprofiler:30281270
vshan820:30294714
vstes263:30335439
vscoreces:30445986
vscod805:30301674
binariesv615:30325510
bridge0708:30335490
bridge0723:30353136
vsaa593cf:30376535
pythonvs932:30410667
py29gd2263:30792226
vsclangdc:30486549
c4g48928:30535728
dsvsc012:30540252
pynewext54:30695312
azure-dev_surveyone:30548225
vscccc:30803845
282f8724:30602487
89544117:30613380
a9j8j154:30646983
showlangstatbar:30737416
0bi6i642:30835152
a2ce3375:30757347
pythonfmttext:30731395
fixshowwlkth:30771522
showindicator:30805244
pythongtdpath:30769146
i26e3531:30792625
pythonnosmt12:30797651
pythonidxptcf:30805731
pythonnoceb:30805159
copilotsettingt:30839829
asynctok:30821568
dsvsc013:30795093
dsvsc014:30804076
diffeditorv2:30821572
dsvsc015cf:30829746

```

</details>

andrewbranch:
Fixes #55863

There were two bugs in play:

1. `importModuleSpecifierPreference` was only being considered in how to generate module specifiers between a given pair of files. But when computing auto-imports, there may be multiple files a missing symbol can be imported from, so the module specifier generation code is called multiple times, and multiple fixes are generated. Those fixes are then sorted, but the sort logic didn’t consider `importModuleSpecifierPreference`. In most cases, all the generated fixes either successfully satisfied the preference or didn’t, but it’s possible, as in the bug repro and test case, that e.g. the symbol being imported is available from one file that can only be reached by relative path and another file that can be reached by a non-relative `paths`/`baseUrl` module specifier.
2. In the import fix sorting, we also sort by whether the module specifier for the fix is a listed package.json dependency whenever that specifier is non-relative. But a non-relative specifier may be a node_modules package, or it may be generated from `paths`/`baseUrl`. We were incorrectly checking whether these `paths`/`baseUrl` specifiers were package.json dependencies, which would further deprioritize them under relative paths. (Writing this out makes me realize that I need to follow up with a separate test to ensure that package.json `imports` and self-name imports are not being deprioritized for the same reason.)

### DIFF
--- a/tests/cases/fourslash/server/autoImportProvider_importsMap1.ts
+++ b/tests/cases/fourslash/server/autoImportProvider_importsMap1.ts
@@ -29,4 +29,4 @@
 // @Filename: /src/a.ts
 //// isBrowser/**/
 
-verify.importFixModuleSpecifiers("", ["./env/browser.js", "#is-browser"]);
+verify.importFixModuleSpecifiers("", ["#is-browser", "./env/browser.js"]);
--- a/src/compiler/moduleSpecifiers.ts
+++ b/src/compiler/moduleSpecifiers.ts
@@ -107,6 +107,7 @@ import {
     removeTrailingDirectorySeparator,
     replaceFirstStar,
     ResolutionMode,
+    ResolvedModuleSpecifierInfo,
     resolveModuleName,
     resolvePath,
     ScriptKind,
@@ -287,14 +288,15 @@ export function tryGetModuleSpecifiersFromCache(
     host: ModuleSpecifierResolutionHost,
     userPreferences: UserPreferences,
     options: ModuleSpecifierOptions = {},
-): readonly string[] | undefined {
-    return tryGetModuleSpecifiersFromCacheWorker(
+): ModuleSpecifierResult | undefined {
+    const result = tryGetModuleSpecifiersFromCacheWorker(
         moduleSymbol,
         importingSourceFile,
         host,
         userPreferences,
         options,
-    )[0];
+    );
+    return result[1] && { kind: result[0], moduleSpecifiers: result[1], computedWithoutCache: false };
 }
 
 function tryGetModuleSpecifiersFromCacheWorker(
@@ -303,7 +305,7 @@ function tryGetModuleSpecifiersFromCacheWorker(
     host: ModuleSpecifierResolutionHost,
     userPreferences: UserPreferences,
     options: ModuleSpecifierOptions = {},
-): readonly [specifiers?: readonly string[], moduleFile?: SourceFile, modulePaths?: readonly ModulePath[], cache?: ModuleSpecifierCache] {
+): readonly [kind?: ModuleSpecifierResult["kind"], specifiers?: readonly string[], moduleFile?: SourceFile, modulePaths?: readonly ModulePath[], cache?: ModuleSpecifierCache] {
     const moduleSourceFile = getSourceFileOfModule(moduleSymbol);
     if (!moduleSourceFile) {
         return emptyArray as [];
@@ -311,7 +313,7 @@ function tryGetModuleSpecifiersFromCacheWorker(
 
     const cache = host.getModuleSpecifierCache?.();
     const cached = cache?.get(importingSourceFile.path, moduleSourceFile.path, userPreferences, options);
-    return [cached?.moduleSpecifiers, moduleSourceFile, cached?.modulePaths, cache];
+    return [cached?.kind, cached?.moduleSpecifiers, moduleSourceFile, cached?.modulePaths, cache];
 }
 
 /**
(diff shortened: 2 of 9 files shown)


tests/cases/fourslash/server/autoImportProvider_importsMap1.ts
```diff
@@ -29,4 +29,4 @@
 // @Filename: /src/a.ts
 //// isBrowser/**/
 
-verify.importFixModuleSpecifiers("", ["./env/browser.js", "#is-browser"]);
\ No newline at end of file
+verify.importFixModuleSpecifiers("", ["#is-browser", "./env/browser.js"]);
```
andrewbranch, line 32: It should have been in this order due to number of path segments, but `#is-browser` was being demoted because it was being checked against package.json dependencies, as I mentioned in the PR description.

jakebailey (approved):

sheetalkamat (approved):

PR #49297, opened 2022-05-29
Normative: Arbitrary module namespace identifier names

Jack-Works:
Implementing https://github.com/tc39/ecma262/pull/2154/

Issue #15096, opened 2017-04-10
Support for TC39 "BigInt: Arbitrary precision integers in JavaScript" proposal


tarcieri:
The ECMAScript Technical Committee 39 has put forward a concrete proposal for adding integers to JavaScript:

https://tc39.github.io/proposal-bigint/

The proposal adds a new syntax for integer literals:

```
42n // This is an integer literal
```

Implicit conversions to/from numbers and integers and expressions with mixed operands are expressly disallowed:

```
42n + 1 // Throws TypeError
```

Type constructors are provided that wrap at specified widths:

* `Integer.asUintN(width, Integer)`: Wrap an Integer between `0` and `2**width-1`
* `Integer.asIntN(width, Integer)`: Wrap an Integer between `-2**(width-1)` and `2**(width-1)-1`
* `Integer.parseInt(string[, radix])`: Analogous to `Number.parseInt`, to parse an Integer from a String in any base.

Though not stated in the spec, these constructors should theoretically hint to the VM when it's possible to use a native 64-bit integer type instead of a bignum, and could therefore possibly be used by TypeScript to implement `int64` and `uint64` types in addition to e.g. an `integer` type of arbitrary precision.

I know people have been asking for integers for quite some time (e.g. #195, #4639), but have been held back by lack of native support of an integer type in JavaScript itself. Now it seems this [~proposal is stage 2~](https://twitter.com/BrendanEich/status/851235756483858432) (scratch that, [stage 3](https://github.com/tc39/proposal-bigint/commit/1b4c7e6cbc2be7d0cc71cec08ba37916774ee157)!) and has multi-browser vendor backing, so perhaps the prerequisites are finally in place to make integers happen.

aluanhaddad:
🚲:house:
Why didn't they choose `Z` for the literal suffix?

tarcieri:
Looks like this may be landing in Babel soon:

https://github.com/babel/proposals/issues/2

tarcieri:
Initial (WIP) babel PR for this: https://github.com/babel/babel/pull/6015

Cixelyn:
Looks like the BigInt proposal has also hit stage 3 (tc39/proposal-bigint@1b4c7e6cbc2be7d0cc71cec08ba37916774ee157). Would be awesome to see TypeScript support.

raix:
Personally looking forward to getting the feature available - only concern is the ducktape syntax, my hope is that typescript can help keep the syntax clean.

[...]

PR #58608, opened 2024-05-21
fix35982: allow BigIntLiteral to parse as PropertyName for literal object and indices

### LINKED ISSUE #35982: LiteralPropertyName should allow BigInt
**TypeScript Version:**  3.8.0-dev.20200101

**Search Terms:** bigint, property name

**Code**

```ts
({1n: 123});
```

**Expected behavior:**
No error

**Actual behavior:**
Property Assignment Expected

**Playground Link:** https://www.typescriptlang.org/play/?ts=3.8.0-dev.20200101&ssl=1&ssc=1&pln=1&pc=15#code/BQbwBAjAdgXJBMBmMBfAlAbiA
**Related Issues:**

https://github.com/tc39/test262/pull/2457
https://bugs.chromium.org/p/v8/issues/detail?id=10083

https://github.com/microsoft/TypeScript/commit/ecd12453f42cd95503bfd0a56a67b732a0ea83d4 should be partially reverted.

iisaduan:
Fixes #35982

This PR allows more valid bigint uses in JS to parse and emit correctly. We will parse (and correctly emit) the following cases, but give the following errors:
```
{ ({1n: 123}); };
    ~~ 1538: A `bigint` literal may not be used as a property name.

[1,2,3][1n]; 
        ~~ 2538: Type `bigint` cannot be used as an index type.

const arr = [1,2,3];
const { 0n: f } = arr; 
       ~~ 2538: Type `bigint` cannot be used as an index type.

class K {
    4n = 0;
    ~~ 1538: A `bigint` literal may not be used as a property name.
}
```


iisaduan:
@typescript-bot test it

src/compiler/checker.ts
```diff
@@ -18365,14 +18365,15 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
-            if (indexType.flags & (TypeFlags.StringLiteral | TypeFlags.NumberLiteral)) {
+            if (indexNode.kind !== SyntaxKind.BigIntLiteral && indexType.flags & (TypeFlags.StringLiteral | TypeFlags.NumberLiteral)) {
                 error(indexNode, Diagnostics.Property_0_does_not_exist_on_type_1, "" + (indexType as StringLiteralType | NumberLiteralType).value, typeToString(objectType));
             }
             else if (indexType.flags & (TypeFlags.String | TypeFlags.Number)) {
                 error(indexNode, Diagnostics.Type_0_has_no_matching_index_signature_for_type_1, typeToString(objectType), typeToString(indexType));
             }
             else {
-                error(indexNode, Diagnostics.Type_0_cannot_be_used_as_an_index_type, typeToString(indexType));
+                const typeString = typeToString(indexNode.kind === SyntaxKind.BigIntLiteral ? getBaseTypeOfLiteralType(getContextFreeTypeOfExpression(indexNode)) : indexType);
```
DanielRosenwasser, line 18375: Why do you need `getContextFreeTypeOfExpression` here?

iisaduan: It gets a type from `indexNode`. Without this call/if I used `getBaseTypeOfLiteralType` on `indexType`, then the type returned is `string` which isn't right

tests/cases/compiler/bigintIndex.ts
```diff
@@ -30,3 +31,11 @@ typedArray[2] = 0xCC;
...
 const b = {[1n]: 456};
 const c = {[bigNum]: 789};
+{ ({1n: 123}); };
+
+// @filename: c.ts
+const crr = [1, 2, 3] as const;
+
+const { 0n: f } = crr; // bigint should give an index error
+const { 0: d } = crr;
+const { "0": e } = crr;
```
DanielRosenwasser, line 41: For these new tests, I would consider put them into a separate file because they have more to do with parsing/grammar checks than indexing into a value (and I would still add tests around trying to index with `bigint`s into those values). I'd probably put that somewhere under `conformance`.

In that file, consider also adding separate "inner" files, each with a test for an `interface`, a `class`, an `enum`, an an [arbitrary import name](https://github.com/microsoft/TypeScript/pull/49297) like

```ts
import { 0n as foo } from "foo";
```

src/compiler/diagnosticMessages.json
```diff
@@ -1801,6 +1801,10 @@
         "category": "Error",
         "code": 1537
     },
+    "A `bigint` literal may not be used as a property name.": {
```
jakebailey, line 1804: Do we typically use backticks in diagnostic messages? I'm surprised to see them, personally. I tried searching and seems like we do, though inconsistency...

src/compiler/checker.ts
jakebailey: This doesn't seem right; when I debug, the literal appears to have all of the info that would be needed for an error message:

image

jakebailey: Yeah, it seems like `pseudoBigIntToString` would be sufficient:

```ts
            const indexNode = getIndexNodeForAccessExpression(accessNode);
            if (indexType.flags & (TypeFlags.StringLiteral | TypeFlags.NumberLiteral)) {
                error(indexNode, Diagnostics.Property_0_does_not_exist_on_type_1, "" + (indexType as StringLiteralType | NumberLiteralType).value, typeToString(objectType));
            }
            else if (indexType.flags & TypeFlags.BigIntLiteral) {
                error(indexNode, Diagnostics.Property_0_does_not_exist_on_type_1, pseudoBigIntToString((indexType as BigIntLiteralType).value), typeToString(objectType));
            }
            else if (indexType.flags & (TypeFlags.String | TypeFlags.Number)) {
                error(indexNode, Diagnostics.Type_0_has_no_matching_index_signature_for_type_1, typeToString(objectType), typeToString(indexType));
            }
            else {
                const typeString = typeToString(indexType);
                error(indexNode, Diagnostics.Type_0_cannot_be_used_as_an_index_type, typeString);
            }
```

src/compiler/diagnosticMessages.json
iisaduan: I saw the same when looking through the messages :P

iisaduan:
@typescript-bot test it

iisaduan:

src/compiler/checker.ts
```diff
@@ -18707,14 +18707,15 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
```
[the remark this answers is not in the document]
rbuckton: We should either write the value, or `bigint` (all lowercase) as it's the underlying base type.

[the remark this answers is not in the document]
iisaduan: Sorry, I've been looking at the tests, and some of the test cases have been inconsistent with the caps. I'll fix that.

Would you prefer the "property does not exist on type" error? I was using the "bigint literal type cannot be used as an index" error because the [original design choice](https://github.com/microsoft/TypeScript/issues/15096#issuecomment-438751487) was the bigintliteral type cannot be used as an index, as you should use a string or number instead. "cannot be used as index" seemed more descriptive of what was happening than "property does not exist". 

I don't mind using the value in the error message (`1n cannot be used as an index type`) if you like that better, and it is what you would get with `typeToString(indexType)`. I changed it because `bigint` made the above more clear as well.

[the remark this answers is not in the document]
iisaduan: cc @rbuckton @jakebailey @DanielRosenwasser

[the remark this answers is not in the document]
jakebailey: I guess if you just swap it back to `bigint`, that should be sufficient. I see what you mean about the inconsistency; it's not that say `1n` doesn't exist on the type, but that bigint isn't supported at all?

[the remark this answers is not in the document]
iisaduan: Yes, it's valid syntactically in js to index with bigints, but we've disallowed them when type checking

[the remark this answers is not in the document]
DanielRosenwasser: It's `bigint` not `bigInt`

src/compiler/checker.ts
```diff
@@ -18753,14 +18753,15 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
-            if (indexType.flags & (TypeFlags.StringLiteral | TypeFlags.NumberLiteral)) {
+            if (indexNode.kind !== SyntaxKind.BigIntLiteral && indexType.flags & (TypeFlags.StringLiteral | TypeFlags.NumberLiteral)) {
                 error(indexNode, Diagnostics.Property_0_does_not_exist_on_type_1, "" + (indexType as StringLiteralType | NumberLiteralType).value, typeToString(objectType));
             }
             else if (indexType.flags & (TypeFlags.String | TypeFlags.Number)) {
                 error(indexNode, Diagnostics.Type_0_has_no_matching_index_signature_for_type_1, typeToString(objectType), typeToString(indexType));
             }
             else {
-                error(indexNode, Diagnostics.Type_0_cannot_be_used_as_an_index_type, typeToString(indexType));
+                const typeString = indexNode.kind === SyntaxKind.BigIntLiteral ? `bigint` : typeToString(indexType);
```
DanielRosenwasser, line 18763: Stay consistent with surrounding code and just use a plains string literal

```suggestion
                const typeString = indexNode.kind === SyntaxKind.BigIntLiteral ? "bigint" : typeToString(indexType);

```

src/compiler/diagnosticMessages.json
```diff
@@ -1809,6 +1809,10 @@
         "category": "Error",
         "code": 1538
     },
+    "A `bigint` literal may not be used as a property name.": {
```
DanielRosenwasser, line 1812: It's throwing me off that half the messages in the output use a different quote style. Let's stay consistent with the majority of uses.

DanielRosenwasser: Let's also stay consistent with "cannot" instead of "may not".

DanielRosenwasser (approved):
There are a few things that I didn't catch before that I want addressed, but I'm going to mark this as approved so you're not blocked on merging.

PR #58610, opened 2024-05-21
Propagate the error any type in union and intersection construction

### LINKED ISSUE #58598: [transpileDeclaration API][5.5] Self referencing generic type constraint with union gets emitted as `unknown`
### 🔎 Search Terms

transpiledeclaration type generic constraint unknown self referencing union

[5.5.0-dev.20240519](https://www.npmjs.com/package/typescript/v/5.5.0-dev.20240519)

### ⏯ Playground Link

https://www.typescriptlang.org/dev/bug-workbench/?target=7&ts=5.5.0-dev.20240519&ssl=1&ssc=1&pln=3&pc=1#code/PTAEAEGcBcCcEsDG0Bco4FcCmAoEEATLRAGwENYzp4B7AOzU13ygAsaB3AUQFt5o8YNp179oWAgDF4JLGjIA6AguiQcgiADMZWOmR5zQi1TiwAPAA41Y0UInoxQNAEYArYrYC8oAN45QoJo0NGgAPAAqoObidASQoADKWNARAHygAD6gANoAuqkAFACUaABuNPAEoJ7pPgC+ADQ4dQDcOEA

### 💻 Code

```ts
// @strict: true
// @declaration: true
// @showEmit
// @showEmittedFile: a.d.ts

// @filename: a.ts
export const object = {
  foo: <T extends Set<T> | []>(): void => {},
};
```

### 🙁 Actual behavior

When using the API, a.d.ts is emitted as
```ts
export declare const object: {
    foo: <T extends unknown>() => void;
};
```

while when using tsc it gets emitted as
```ts
export declare const object: {
    foo: <T extends Set<T> | []>() => void;
};
```

### 🙂 Expected behavior

Using the API should emit

```ts
export declare const object: {
    foo: <T extends Set<T> | []>() => void;
};
```

Weird example I know. Trust me that our actual code makes a bit more sense.

weswigham:
You might not think this is the obvious fix for the issue, but it is! Type parameter constraints, when they see a non-error `any`, remap it to `unknown` (there are historical reasons for this). So by swallowing the error-`any` (and replacing it with a normal `any`), union and intersection construction resulted in the constraint calculation logic not witnessing the error type, and thus remapping the constraint (and, thus, making the constraint from the type node not match the cached constraint on the type parameter itself!)

This should also just behave a bit better with respect to follow on errors in some other cases, too.

Fixes #58598


jakebailey:
Is this at all related to #56429?

weswigham:
It's related insofar as better `error`-type propagation often creates better behaviors for follow-on locations for us.

src/compiler/checker.ts
```diff
@@ -17303,7 +17304,8 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         if (unionReduction !== UnionReduction.None) {
             if (includes & TypeFlags.AnyOrUnknown) {
                 return includes & TypeFlags.Any ?
-                    includes & TypeFlags.IncludesWildcard ? wildcardType : anyType :
+                    includes & TypeFlags.IncludesWildcard ? wildcardType :
+                        includes & TypeFlags.IncludesError ? errorType : anyType :
                     unknownType;
```
sandersn, line 17309: These ternaries are officially off the chain now.

src/compiler/checker.ts
```diff
@@ -17637,7 +17640,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             return neverType;
         }
         if (includes & TypeFlags.Any) {
-            return includes & TypeFlags.IncludesWildcard ? wildcardType : anyType;
+            return includes & TypeFlags.IncludesWildcard ? wildcardType : includes & TypeFlags.IncludesError ? errorType : anyType;
```
sandersn, line 17643: it would be nice to have multiple lines here

src/compiler/types.ts
```diff
@@ -6185,6 +6185,8 @@ export const enum TypeFlags {
     StringMapping   = 1 << 28,  // Uppercase/Lowercase type
     /** @internal */
     Reserved1       = 1 << 29,  // Used by union/intersection type construction
+    /** @internal */
+    Reserved2       = 1 << 30,  // Used by union/intersection type construction
```
sandersn, line 6189: remind me, do we have 31 bits available?

sandersn (approved):
Looks good, some quibbles with formatting. And we should see how DT/top400 tests look.

src/compiler/types.ts
jakebailey: In modern Node, yep, we do.

jakebailey:
@typescript-bot test it

src/compiler/checker.ts
weswigham: 🤷‍♀️ I just write the ternaries, I let the formatter figure out how it wants to indent them.

weswigham:
@typescript-bot pack this

weswigham:

The `effect` errors I can't seem to repro locally...? I think they're along the lines of what'd happen if something like something in the base type chain for the classes in question didn't resolve, but they all do...?

weswigham:
Something is funky with `Effect-TS` in the user harness - the runner shows thousands of errors in the log (`Error counts for https://github.com/Effect-TS/effect/blob/e3766411b60ebb45d31e9c9d94efa099121d4d58/tsconfig.json: new = 4804, old = 4775`). When you download, install, and build it locally (the same version linked in the log), even with this PR, it's not *nearly* that bad - 15 errors across all projects, none of them are the ones listed here.

@jakebailey is the runner silently failing to install/link dependencies correctly for that user test or something? That's the only thing I could think of, which is why I reran the test. And it would make sense - these kinda of errors are what I'd expect to see if some intersections in class base types contained unresolved types that now get propagated out (instead of quietly `any`'d). Error types in class base types make the class's base types get set to nothing, rather than `any`.

jakebailey:
The test thing infers packages and tsconfigs and attempts to install them and run on each of them, so it's possible that it's just doing things "wrong" for that repo. We can convert it to a script test or something where it more clearly tries to run things, but we'll lose the inference of configs and such. I'm not sure I have the time to try and fix the root problem, though.

weswigham:
The project *does* have one diff in the real output - a new error - but it's a follow-on caused by what I said above - a class has an error type intersected into its' base types, so all its' base types get erased now, rather than `any`'d. And the root error is creeping in because the project has `skipLibCheck` on, so isn't aware that the types for one of its' transitive deps is just straight up missing. (Specifically, `packages/sql-mysql2` depends on an external `mysql2`, which depends on `events` and `stream`, which are presumably supposed to be provided by `@types/node`, but aren't because the project doesn't include `@types/node`!)

Big downside of using `skipLibCheck` - you don't get any errors for a missing environment! Though I can see why they enable it - they have a metric ton of straight up broken deps.

weswigham:
Useful investigation regardless, even if there wasn't really much related to this issue, disabling `skipLibCheck` in it pointed out [this regression](https://github.com/microsoft/TypeScript/issues/58627) that I've opened an issue for.

PR #58612, opened 2024-05-22
Correct regular expression flags scanning for non-BMP characters

graphemecluster:
Per the [IsValidRegularExpressionLiteral](https://tc39.es/ecma262/#sec-isvalidregularexpressionliteral) static semantics in the ECMA262 specification, [FlagText](https://tc39.es/ecma262/#sec-static-semantics-flagtext) should be interpreted in code points, not code units.

Cherry picked from [#58289@`e67692a`](https://github.com/microsoft/TypeScript/pull/58289/commits/e67692acb3fdd068b4e577dc9ad9fa350f2e4ca8).


tests/cases/compiler/regularExpressionWithNonBMPFlags.ts
```diff
@@ -0,0 +1,3 @@
+// @target: esnext
+
+const 𝘳𝘦𝘨𝘦𝘹 = /(?𝘴𝘪-𝘮:^𝘧𝘰𝘰.)/𝘨𝘮𝘶;
```
rbuckton, line 3: It would be helpful for future readers if you added a comment indicating that the characters below are Unicode variant characters and not ASCII characters, e.g.

```suggestion
// The characters in the following regular expression are ASCII-lookalike characters found in Unicode, including:
// - 𝘴 (U+1D634 Mathematical Sans-Serif Italic Small S)
// - 𝘪 (U+1D62A Mathematical Sans-Serif Italic Small I)
// - 𝘮 (U+1D62E Mathematical Sans-Serif Italic Small M)
//
// See https://en.wikipedia.org/wiki/Mathematical_Alphanumeric_Symbols
const 𝘳𝘦𝘨𝘦𝘹 = /(?𝘴𝘪-𝘮:^𝘧𝘰𝘰.)/𝘨𝘮𝘶;
```

rbuckton (changes requested):

rbuckton (approved):

rbuckton:
Overall, this looks good. It looks like there are some failing tests that still need to be resolved. Also, since this isn't a bug fix, per se, I expect we will want to hold off from putting this in the 5.5 RC and will wait to merge after `main` is open for TypeScript 5.6 work.

rbuckton (changes requested):
Switching from "Approve" to "Request Changes" until the test issues are resolved.

graphemecluster:
IMO it is a bug fix especially since I forgot to revert the change on line 2561 in #58289. And it’s important to align the scanner behavior to the spec.
#58615 is a bug fix, too, that is even more critical since the bug breaks real-world codebases.
#58613 is not the case, but (as an excuse) it was supposed to be included in #55600.

rbuckton:
> IMO it is a bug fix especially since I forgot to revert the change on line 2561 in #58289. And it’s important to align the scanner behavior to the spec. 

What change do you mean? If it's not related to code points then I could see taking a small fix for that. Properly handling non-BMP characters doesn't come up in practice and delaying this specific functionality will only result in a slightly less informative error since none of the valid flag/modifier characters will ever be equivalent to the first byte of a multi-byte code point. 

> #58615 is a bug fix, too, that is even more critical since the bug breaks real-world codebases. #58613 is not the case, but (as an excuse) it was supposed to be included in #55600.

#58613 can wait till 5.6, but I'll review #58615 shortly.

graphemecluster:
> What change do you mean? If it's not related to code points then I could see taking a small fix for that.

[This line](https://github.com/microsoft/TypeScript/pull/58289/files#diff-e19070e6b0ecc4de5f7d4bc6b42a9a71c8f23ce8ce155f7bd7cd48763170e506R2521), which causes the `String.fromCharCode` below to convert incorrectly (as the method only consider the 16 least significant bits).

rbuckton:
> > What change do you mean? If it's not related to code points then I could see taking a small fix for that.
> 
> [This line](https://github.com/microsoft/TypeScript/pull/58289/files#diff-e19070e6b0ecc4de5f7d4bc6b42a9a71c8f23ce8ce155f7bd7cd48763170e506R2521), which causes the `String.fromCharCode` below to convert incorrectly (as the method only consider the 16 least significant bits).

I've just put up a minimal fix for that line in #58727.

rbuckton (approved):

PR #58618, opened 2024-05-22
feat(lib): allow `NonNullable<T, K>` to select specific properties

ctjlewis:
What are our thoughts on this pattern? It seems like a lot to ask people to find the ideal syntax to flag properties as non-nullable:

```ts
type NonNullableProperties<T, K extends keyof T> = Omit<T, K> & {
  [P in K]-?: NonNullable<T[P]>;
};
```

It seems like it makes sense to just update `NonNullable` and retain the fast path over adding a new lib type. I expect there will just be pushback on this and it will be closed, but just a suggestion.


ctjlewis:
Fast path does not work as expected actually, closing this.

PR #58620, opened 2024-05-22
Error if type node uses inaccessible type in isolated declarations

### LINKED ISSUE #58558: Isolated declarations allows return types that reference local expr or type
### 🔎 Search Terms

- isolated declarations
- typeof
- local types
- function return type

5.5.0-dev.20240517

### ⏯ Playground Link

https://www.typescriptlang.org/play/?target=99&isolatedDeclarations=true&ts=5.5.0-dev.20240517#code/MYewdgzgLgBAHjAvDARCg3AKAKZwA4gBOsAZgK5jBQCW4M42AFAJQwDemMXMh2UZhMOwC+MAIYQYUAJ55sIEvCzDMOfEVIUqtIVADuIFu07dQkWNKSoMJrr36CR4yTLkKY05atwFiMcpQ0dFAAFrxMrBzcUrLYMABaVtCE1GAA5ljR9gJCbKISCV5AA

### 💻 Code

```ts
const x = "";
export function one() {
    return {} as typeof x;
}

export function two() {
    const y = "";
    return {} as typeof y;
}

export function three() {
    type Z = string;
    return {} as Z;
}
```

### 🙁 Actual behavior

No diagnostics are raised.

### 🙂 Expected behavior

An isolated declarations error is raised for function `two` and function `three` (missing explicit return type), because even though both return an "explicit" type (using `as` casting), neither can be emitted without using a type checker because the type / expression they reference are internal to the function.

I was looking for a case where this can happen top level, but I just ran into a different TypeScript error then: https://github.com/microsoft/TypeScript/issues/58557

cc @dragomirtitian

dragomirtitian:
Fixes #58558


tests/baselines/reference/transpile/declarationNotInScopeTypes.d.ts
```diff
@@ -0,0 +1,47 @@
...
+        return {} as typeof y;
+                            ~
+!!! error TS9038: Type containing private name 'y' can't be used with --isolatedDeclarations.
+    }
+    
+    export function three() {
+        type Z = string;
+        return {} as Z;
+                     ~
+!!! error TS9038: Type containing private name 'Z' can't be used with --isolatedDeclarations.
```
weswigham, line 46: Q: Is there a reason this shouldn't get a "annotate this function return" related span?

weswigham (approved):

tests/baselines/reference/transpile/declarationNotInScopeTypes.d.ts
dragomirtitian: It should. I'll add it

weswigham (approved):

PR #58626, opened 2024-05-22
Always write tsbuildInfo when running tsc -b

sheetalkamat:
For now it only contains `version` but with plan to add more info about program for tsc -b uptodateness.

In future this can be used to:
- Remove necessity that `composite` must list all files (probably)
- Remove noEmitOnError mandate when running `tsc -b` 
- Experiment with building downstream projects irrespective of errors (or some type of errors)
- Handle `noCheck` on commandLine
- Add extra files to check for uptodate (that are not listed in program like packageJson or additional program files)


sheetalkamat:
@typescript-bot test it

src/compiler/tsbuildPublic.ts
```diff
@@ -328,6 +326,7 @@ function getCompilerOptionsOfBuildOptions(buildOptions: BuildOptions): CompilerO
     commonOptionsWithBuild.forEach(option => {
         if (hasProperty(buildOptions, option.name)) result[option.name] = buildOptions[option.name];
     });
+    result.tscBuild = true;
```
jakebailey, line 329: How will this flag affect people using our APIs to implement build mode like tsc? Will it be set for them? Or can we just assume it's always true?

jakebailey: Alrighty; just verifying so people are not surprised.

jakebailey (approved):

jakebailey (approved):

andrewbranch (approved):

PR #58640, opened 2024-05-24
Implement "Arbitrary Module Namespace Identifiers"

### LINKED ISSUE #40594: Support “Arbitrary module namespace identifier names”
## Search Terms

- Arbitrary module namespace identifier names
- Arbitrary module namespace identifiers
- Arbitrary module identifier names
- Arbitrary module identifiers
- Arbitrary module identifier
- Arbitrary export names
- Arbitrary export identifiers
- string export names
- string exports

## Suggestion

Implement support for <https://github.com/tc39/ecma262/pull/2154>.

## Use Cases

This is necessary for **WASM** module integration.

This would also allow typing [`@types/webidl‑conversions`](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/webidl-conversions) without needing to use `export =`.

---

For transpiled modules, this can already be supported, since it’s just creation of arbitrary properties on the **CommonJS** `exports` object, but native modules will require `module: "ESNext"`.

## Examples

See <https://github.com/tc39/ecma262/pull/2154>

```ts repro
// @showEmit
// @filename: module.d.ts
export const foo: unknown;

declare const some_imports: unknown;
export { some_imports as "some imports" };

declare const another_imports: unknown;
export { another_imports as "another imports" };

declare const extra_imports: unknown;
export { extra_imports as "extra imports" };

declare const rest_exports: unknown;
export { rest_exports as "rest exports" };

// @filename: index.ts
import { "some imports" as foo } from "./module.js";

export { foo as "some exports" } from "./module.js";
export { "another imports" as "another exports" } from "./module.js";
export { "rest exports" } from "./module.js";
export { "extra imports" as extra_imports } from "./module.js";

export * as "star exports" from "./module.js";
```

[Workbench Repro](https://www.typescriptlang.org/dev/bug-workbench/?#code/PTAEAEGcAsHsHcCiBbAlgFwLACgQQGaoA2ApgHYCGyJAXKMrACYCupAdI2+pDiQB4AHWACd0oAMawykMfliw6zMgGsyCMgG4cORiXFEKwkhKkzQkWNQD6qZENGRFKtfE29BIsQG9zlkjbtPSFAKYIAiC2pQW3tuMNAAXy1sHT0DIxNpMQo1dGgSYQDYx1AlVXVk-ljQHxzYPIKioJDwuobhaMCHeKTtbF19Q2NJLNB+dGEKJocnctdKj1Easb4JqZjm0NAw8cnO4p7k1MGMkbMjGSsqoNmXN2xrpZ8L9CvF7hbtl5WDxKPcMDgQikSjUOioMi6PhcHjYDZPbaRYzwuKfOSwRKgfDCSzbNjABgsdgAK0gYX+j28WPknwifh+QR6WJxyDxBKYrBIbFJ5Pc1R8YTa+Q6KLJtKFBQZ3Ux2NxYXxhM53LJC35XxIZkpYoSzLlCo5JJVfM8yx2qz2oviW126y6Hx1stZ8vZRK5PIp71AACpaTJDFLUY62YrDbzsEA)

## Checklist

## Related issues

- <https://github.com/microsoft/TypeScript/issues/43870>
- <https://github.com/microsoft/TypeScript/pull/49297>

### LINKED ISSUE #43870: Allow Unicode string imports and exports
https://github.com/tc39/ecma262/pull/2154

```ts
import { "😊" as yay } from "...";

yay();

export { yay as "🦈" };
```

The above code should be supported.

The minimum viable product should have tests for:

* `esnext` module output
* `es2015` or similar module output just to get an error
* `--declaration` emit
  * Specific declaration emit where the compiler might generate `import()` type syntax.
* Quick info, and error message tests where we might generate `import()` type syntax
* Completions in `import {}` and `export {}` lists
* Go-to-definition

It is likely that we would never try to emit our own `import()` type syntax.

### LINKED ISSUE #49147: ImportSpecifier.propertyName should allow string literals
# Bug Report

### 🔎 Search Terms

import specifier, property name, ModuleExportName, named import

### ⏯ Playground Link

[Playground link with relevant code](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgIgK4GcCmBlGBDGTZOPdODTAUQDNrMBjeAXzmqghBSkz0eQG4AUEA)

### 💻 Code

```ts
import { "useState" as useEffect } from "react";
```

### 🙁 Actual behavior

Parser error

### 🙂 Expected behavior

This should be parsed as valid code and properly type-checked. According to [ES spec](https://tc39.es/ecma262/multipage/ecmascript-language-scripts-and-modules.html#prod-ImportSpecifier), `ImportSpecifier` can be `ModuleExportName as ImportedBinding`, where `ModuleExportName` includes `StringLiteral`.

[Babel parses this correctly.](https://babeljs.io/repl#?browsers=chrome%2030&build=&builtIns=false&corejs=3.21&spec=true&loose=false&code_lz=JYWwDg9gTgLgBAbzgIgK4GcCmBlGBDGTZOPdODTAUQDNrMBjeAXzmqghBSkz0eQG4AUIPoQAduggAbTADopEAOYAKCjTqMAlEKA&debug=false&forceAllTransforms=false&shippedProposals=true&circleciRepo=&evaluate=true&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Creact&prettier=true&targets=&version=7.17.12&externalPlugins=babel-plugin-dynamic-import-node%402.3.3&assumptions=%7B%7D)

Accordingly, this should be also allowed:

```ts
const a = 1;
export { a as "a-b" };
```

### LINKED ISSUE #52092: Support arbitrary module namespace identifier names
# Suggestion

The new  ES2022 standard allows to use string literals in module exports and imports:

```javascript
import {"\0 any unicode" as foo} from "";
export {foo as "\0 any unicode"};
```
Ref: https://github.com/tc39/ecma262/pull/2154

This feature is supported in NodeJS since v16(2021) and I have been using this feature but typescript complains about invalid syntax and I don't see this feature in the roadmap?.

## 🔍 Search Terms
String literals import and export

## ✅ Viability Checklist

### LINKED ISSUE #54584: `export { a as "hello world" }` is valid JavaScript but invalid TypeScript
# Bug Report

### 🔎 Search Terms

"es2020"
"export as string"
"import as string export as string"

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about "exports"

### ⏯ Playground Link

[Playground link with relevant code](https://www.typescriptlang.org/play?target=99&ts=5.1.3#code/MYewdgzgLgBAhjAvDAzAFgNwCgCmAPABxACdYBveeCGAIgAscAbRkGAdxMYBMaYBfDEA)

### 💻 Code

```ts
// This typescript code...
const a = 34;
export { a as "hello world" };
```

### 🙁 Actual behavior

```js
// ...generates this invalid JavaScript code.
const a = 34;
export { a as  } from "hello world";
;
```

The input was **valid JavaScript** so I would _expect_ (intuitively) for it to be valid TypeScript. This is currently not the case.

### 🙂 Expected behavior

```js
// I expected this JavaScript code to be generted:
const a = 34;
export { a as "hello world" };
```

Current behavior in browsers and Node.js works with this syntax!
```js
// But running this in the DevTools or Node.js REPL works just fine.
await import("data:text/javascript," + encodeURIComponent(`
const a = 34;
export { a as "hello world" };
`))
```
image
image

JavaScript seems to allow strings as import/export names:
```js
import { "hello world" as a } from "./hello.js"
export { a as "hello world" };
```
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#syntax
https://developer.mozilla.org/en-US/docs/web/javascript/reference/statements/export#syntax

### LINKED ISSUE #54733: Support ECMA-262 arbitrary non-identifier module export names
# Suggestion

## 🔍 Search Terms

string exports, arbitrary identifier namespace names

## ✅ Viability Checklist

## 📃 Motivating Example

Example valid ECMA262 code that fails:

```
import { "some name" as impt } from 'external';
export { impt as "other name" }
```

See:

https://www.typescriptlang.org/play?ts=5.2.0-dev.20230621#code/KYDwDg9gTgLgBAbzmOBDAznARAC2AG3wjgHdp8ATLOAXyA

## 💻 Use Cases

Allow TypeScript to support non-identifier export names per ECMA-262 support (https://github.com/tc39/ecma262/pull/2154).

### LINKED ISSUE #56200: Does not support `export as "string literal"`
### 🔎 Search Terms

"export as", "string literal"

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about [syntax]

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.3.0-dev.20231023#code/MYewdgzgLgBAbgQwDYFcCmMC8MCMBuAKDQA8AHEAJ1gG95l0YEIYAiBMGE8qlmAXzxA

### 💻 Code

```ts
const value = 1;
export { value as "an export" };
```

### 🙁 Actual behavior

```
Cannot find module 'an export' or its corresponding type declarations.(2307)
```

### 🙂 Expected behavior

It should compile correctly and emit a parseable `.d.ts` if configured in the project's settings.

See plain JavaScript example:

lib.mjs 
```js
const value = 1;
export { value as "an export" };
```

main.mjs 
```js
import * as lib from "./lib.mjs";
console.log(lib);
```

node main.mjs 
```
[Module: null prototype] { 'an export': 1 }
```

This fails the "TypeScript is a superset of JavaScript" assertion in the project's description.

Relevant syntax documentation can be found here:
https://developer.mozilla.org/en-US/docs/web/javascript/reference/statements/export#syntax

And in the language specification, 16.2 Modules:
https://tc39.es/ecma262/#prod-ModuleExportName
```
[...]
ModuleExportName:
  IdentifierName
  StringLiteral
```

It seems like an exotic request but has benefits for CSS modules which export classes as named exports.

### LINKED ISSUE #56375: Typescript does not accept `"string name"` aliases
### 🔎 Search Terms

export alias string names
import alias string names
string alias for imports/exports

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about string names as import/export aliases

### ⏯ Playground Link

https://www.typescriptlang.org/play?module=1&ssl=3&ssc=17&pln=3&pc=20#code/MYewdgzgLgBAZiEMC8MCMBuAUFgpgDwAcQAnWAbwSQEMIYAiAI2pPoF9sg

### 💻 Code

```ts
const foo = 1;

export {foo as "bar"};
```

### 🙁 Actual behavior

This fails to compile because of the `as "bar"` in the `export` statement.

### 🙂 Expected behavior

This should compile and allow string name aliases. 

String name import/export aliases should be allowed by the ts compiler. They are valid javascript (see the [syntax section of the MDN page](https://developer.mozilla.org/en-US/docs/web/javascript/reference/statements/export#syntax) in `exports`).

### LINKED ISSUE #57017: Support string literal export names
### 🔍 Search Terms

export string name literal

### ✅ Viability Checklist

### ⭐ Suggestion

JavaScript allows exporting and importing names that are _string literals_ and not just identifiers. This is for compatibility with future non-JS modules types that might not have the same restrictions on export names, like WASM, CSS, HTML, etc.

See https://tc39.es/ecma262/multipage/ecmascript-language-scripts-and-modules.html#prod-ModuleExportName

### 📃 Motivating Example

```ts
import {'product-card' as productCard} from './product.html` with {type: 'html'};
import {'fast-inverse-square-root' as fastInverseSquareRoot} from 'carmack.wasm';
```

### 💻 Use Cases

  Importing from non-JS modules and defining types for non-JS modules

  You can't do either of the above

  None?

evanw:
This PR implements [Arbitrary Module Namespace Identifiers](https://github.com/tc39/ecma262/pull/2154), which is a JavaScript syntax feature that was added to JavaScript over two years ago (in ES2022) but that TypeScript never added support for. Here's an example of what this syntax looks like:

```js
import { "Foo::new" as Foo_new } from "./foo.wasm"

const foo = Foo_new()

export { Foo_new as "Foo::new" }
```

**What this PR does:**

* The syntax can be parsed and emitted
* It works with type checking and formatting
* Seems to work with declaration file generation
* Analogous syntax also works for type imports+exports
* Converts to various module formats (`ESNext`, `CommonJS`, `System`, `AMD`, `UMD`)
* Errors when the module format is `ES2015` or `ES2020`
* Can be used with "go to definition" and "find all references"
* Has basic support for renaming quoted export names across files
* Works with the "convert named imports to namespace import" refactoring
* Provides completions in `import {}` and `export {}` lists
* Includes test coverage for all of the above

This seemed to me like a pretty good MVP for this feature. It could certainly be more polished, but it's already very usable.

_Note that this PR involves changing the TypeScript AST._ This is unavoidable as the JavaScript language has added new syntax.

**What this PR does not do:**

* Syntax highlighting (a separate PR for that is [here](https://github.com/microsoft/TypeScript-TmLanguage/pull/1030))
* Does not work with the "convert namespace import to named imports" refactoring (introducing this syntax when it's absent may not be a good idea)

**My rationale for this PR:**

It's especially unfortunate that TypeScript hasn't implemented this JavaScript feature yet because of the popularity of VSCode. The TypeScript language server is also used for JavaScript, so the lack of support for this makes valid JavaScript files that use this feature look invalid in VSCode. I would not be surprised if TypeScript's lack of support is the main reason why I haven't seen any real usage of this feature yet. I'm hopeful that this PR will be able to get this fixed!

**Real-world example:**

This syntax feature is used by esbuild to configure expressions to be replaced at build-time. The [inject](https://esbuild.github.io/api/#inject) setting takes a file and considers each named export to be a potential expression to replace. For example ([see it live](https://esbuild.github.io/try/#YgAwLjIxLjMALS1idW5kbGUgLS1pbmplY3Q6Li9pbmplY3QuanMgLS1mb3JtYXQ9Y2pzIC0tbWluaWZ5LWlkZW50aWZpZXJzAGUAZW50cnkudHMAdHlwZSBNYXRyaXggPSBbCiAgW251bWJlciwgbnVtYmVyLCBudW1iZXJdLAogIFtudW1iZXIsIG51bWJlciwgbnVtYmVyXSwKICBbbnVtYmVyLCBudW1iZXIsIG51bWJlcl0sCl0KCmV4cG9ydHMucm90YXRlWCA9IChhbmdsZTogbnVtYmVyKTogTWF0cml4ID0+IHsKICBjb25zdCBzaW4gPSBNYXRoLnNpbihhbmdsZSksIGNvcyA9IE1hdGguY29zKGFuZ2xlKQogIHJldHVybiBbWzEsIDAsIDBdLCBbMCwgY29zLCAtc2luXSwgWzAsIHNpbiwgY29zXV0KfQoKZXhwb3J0cy5yb3RhdGVZID0gKGFuZ2xlOiBudW1iZXIpOiBNYXRyaXggPT4gewogIGNvbnN0IHNpbiA9IE1hdGguc2luKGFuZ2xlKSwgY29zID0gTWF0aC5jb3MoYW5nbGUpCiAgcmV0dXJuIFtbY29zLCAwLCBzaW5dLCBbMCwgMSwgMF0sIFstc2luLCAwLCBjb3NdXQp9CgpleHBvcnRzLnJvdGF0ZVogPSAoYW5nbGU6IG51bWJlcik6IE1hdHJpeCA9PiB7CiAgY29uc3Qgc2luID0gTWF0aC5zaW4oYW5nbGUpLCBjb3MgPSBNYXRoLmNvcyhhbmdsZSkKICByZXR1cm4gW1tjb3MsIC1zaW4sIDBdLCBbc2luLCBjb3MsIDBdLCBbMCwgMCwgMV1dCn0KCmV4cG9ydHMuaGFzaCA9IChkYXRhOiBVaW50OEFycmF5KTogbnVtYmVyID0+IHsKICBsZXQgaGFzaCA9IDB4ODExYzlkYzUKICBmb3IgKGxldCBpID0gMDsgaSA8IGRhdGEubGVuZ3RoOyBpKyspIGhhc2ggPSBNYXRoLmltdWwoaGFzaCBeIGRhdGFbaV0sIDB4MDEwMDAxOTMpCiAgcmV0dXJuIGhhc2gKfQAAaW5qZWN0LmpzAC8vIFRoaXMgZmlsZSBpcyBwYXNzZWQgdG8gZXNidWlsZCB1c2luZyAiLS1pbmplY3Q6Li9pbmplY3QuanMiCgovLyBNaW5pZnkgY2FsbHMgdG8gbWF0aCBmdW5jdGlvbnMKY29uc3Qgc2luID0gTWF0aC5zaW4sIGNvcyA9IE1hdGguY29zLCB0YW4gPSBNYXRoLnRhbgpleHBvcnQgeyBzaW4gYXMgJ01hdGguc2luJywgY29zIGFzICdNYXRoLmNvcycsIHRhbiBhcyAnTWF0aC50YW4nIH0KCi8vIFBvbHlmaWxsIDMyLWJpdCBpbnRlZ2VyIG11bHRpcGx5IGlmIG5lZWRlZApjb25zdCBpbXVsID0gTWF0aC5pbXVsIHx8ICgoeCwgeSkgPT4gewogIGNvbnN0IHhsbyA9IHggJiAweGZmZmYsIHlsbyA9IHkgJiAweGZmZmYKICByZXR1cm4geGxvICogeWxvICsgKCgoeCA+Pj4gMTYpICogeWxvICsgKHkgPj4+IDE2KSAqIHhsbykgPDwgMTYgPj4+IDApIHwgMAp9KQpleHBvcnQgeyBpbXVsIGFzICdNYXRoLmltdWwnIH0)):

```js
// This file is passed to esbuild using "--inject:./inject.js"

// Minify calls to math functions
const sin = Math.sin, cos = Math.cos, tan = Math.tan
export { sin as 'Math.sin', cos as 'Math.cos', tan as 'Math.tan' }

// Polyfill 32-bit integer multiply if needed
const imul = Math.imul || ((x, y) => {
  const xlo = x & 0xffff, ylo = y & 0xffff
  return xlo * ylo + (((x >>> 16) * ylo + (y >>> 16) * xlo) << 16 >>> 0) | 0
})
export { imul as 'Math.imul' }
```

This syntax feature is used to convey the expressions to replace because it's both concise and tree-shaking friendly. But writing such a file using VSCode is awkward as the TypeScript language service doesn't fully understand JavaScript syntax. It's also unfortunate that such a file has to be written in JavaScript instead of TypeScript and can't take advantage of type checking.

**See also:**

* https://github.com/microsoft/TypeScript/pull/49297

Fixes #40594
Fixes #43870
Fixes #49147
Fixes #52092
Fixes #54584
Fixes #54733
Fixes #56200
Fixes #56375
Fixes #57017


evanw:

jakebailey:
@typescript-bot test it

fatcerberus:
> I would not be surprised if TypeScript's lack of support is the main reason why I haven't seen any real usage of this feature yet.

I don't know.  I feel like, if I, as a library writer in control of my own API, have the choice between telling my users either:

1. You have to write `import { "Foo.new" as Foo_new } from "foo"`  (a renaming import) for each thing you want to import from this library, -OR-
2. You can just write `import { Foo_new } from "foo"` if you want.

Then I'm going to choose to tell them 2) every time.  If I know a plurality of my consumers are importing from JavaScript then I'm only going to tell them 1) if there's absolutely no way around it.

evanw:
Yes, if your API would be improved by not using this feature than you obviously shouldn't be using it. My point was that if you have a use case for which this feature is the most natural fit (see the real-world example I posted), you're still effectively prevented from using it because most people use VSCode these days and the TypeScript language service hasn't added support for it yet.

csvn:
> Then I'm going to choose to tell them 2) every time. If I know a plurality of my consumers are importing from JavaScript then I'm only going to tell them 1) if there's absolutely no way around it.

I don't think that's the point. One of the issues related to this PR, #40594, mentions this feature is for WASM support, since WASM can export unicode exports.

Typescript also prides itself on being a superset of Javascript, so since Javascript supports this type of imports, Typescript should too. This PR does not encourage exporting library authors to make their API more inconvenient to use, but there are valid use cases for it.

jakebailey:
Ignoring what Daniel already said, I can't really find a fault with the PR as-is.

Is this something we're trying to get into 5.5 RC, or wait for 5.6?

Jack-Works:
> Feel free to add that in this PR, but I wouldn't gate the PR on that at all.

Yes, I strongly suggest this PR should be done without fine-tuning everything. From my experience of developing #49297, I tried to make all the language service stuff great at once, and there was too much work to do and I drained.

I had a checklist for my PR, that's everything I can think of about this feature.

- Code fix: unnecessary string names `export { a as "b" }` => `export { a as b }`
- Special case: `export { a as "export=" }` reports `An export assignment cannot be used in a module with other exported elements.ts(2309)`
- Internal refactor: `InternalSymbolName`?, `CompletionDetailsFull`?, `SignatureHelp`
- Completion
  - `type z = typeof q.^`
  - `{ V as '' }` (empty name does not show up)
  - `export { V as 'export=' }` (this name does not show up)
- Find reference: `ns['b c']`
- Rename
  - `export { ^a }` (identifier to string)
  - `export { a as ^b }` (identifier to string)
  - `export { a as "b" }` (remove unnecessary quotes)
  - `export * as "b c"` (string rename)
  - `export * as x` (identifier to string)
  - `import { a }` (identifier to string)
  - `import { ^c as b }` (identifier to string)
  - `import { "^c" as b }` (string rename)
  - `ns['a b'] => ns['d']` (rename to identifier not removing quotes)
  - `ns.b` (identifier to string)
- Hover: generates invalid code

evanw:
Thanks for the heads up about the `export=` edge case—I wasn't aware of that one. Arbitrary module namespace identifiers causing name collisions with internal symbols is something I had to deal with in esbuild's implementation too (I was [using `"*"` as a special name](https://github.com/evanw/esbuild/commit/24811f1338ae48a4371e1b94b9e376d11c41925a) before I had to change it). It is a strange edge case, but luckily it doesn't block usage of this feature other than for that specific export alias.

It looks like the way to deal with that in the TypeScript code is to add double underscores before the internal symbol name (so maybe `__export=`?). That seems to work but it touches a lot of files (mostly test baselines) and has other consequences (e.g. potentially affecting error messages) so maybe it's best done in a follow-up PR? I'm happy to put a PR up for that too if that's a change that the TypeScript team is open to.

evanw:
> Spitballing, but these should be
> 
> * supported as string literal completions
> * converted into string literal completions if writing an identifier that matches and the client supports insertText
>     * insert an `as` if the name is strictly being used for a local `import` (not re-exporting)
> 
> Feel free to add that in this PR, but I wouldn't gate the PR on that at all.

I added support for the first one (completion within an existing string literal). I'm not sure whether or not I have all of the necessary context about the code to add support for the second one.

evanw:
Ok now I've also added support for the second one (complete identifiers into matching string literals, using `as` when appropriate). But I'm less sure about this change. Is it ok to always return a non-identifier alias as a completion like that? I'm happy to roll that change back and leave it for future work if the correct approach is indeed more complicated than what my solution does.

src/compiler/diagnosticMessages.json
```diff
@@ -8317,5 +8317,9 @@
     "Enum member following a non-literal numeric member must have an initializer when 'isolatedModules' is enabled.": {
         "category": "Error",
         "code": 18056
+    },
+    "Arbitrary module namespace identifiers are not supported when the '--module' flag is set to 'es2015' or 'es2020'.": {
```
rbuckton, line 8321: "Arbitrary module namespace identifiers" is a bit too much inside baseball for most developers. It might be worthwhile to report an error like this instead:

> String literal import and export names are not supported when the '--module' flag is ...

src/compiler/transformers/module/module.ts
```diff
@@ -1613,16 +1625,27 @@ export function transformModule(context: TransformationContext): (x: SourceFile
             for (const specifier of node.exportClause.elements) {
                 const exportNeedsImportDefault = !!getESModuleInterop(compilerOptions) &&
                     !(getInternalEmitFlags(node) & InternalEmitFlags.NeverApplyImportHelper) &&
-                    idText(specifier.propertyName || specifier.name) === "default";
-                const exportedValue = factory.createPropertyAccessExpression(
-                    exportNeedsImportDefault ? emitHelpers().createImportDefaultHelper(generatedName) : generatedName,
-                    specifier.propertyName || specifier.name,
-                );
+                    moduleExportNameIsDefault(specifier.propertyName || specifier.name);
+                const specifierName = specifier.propertyName || specifier.name;
```
rbuckton, line 1629: I would reorder these lines so that `const specifierName` comes first. Then you can reuse `specifierName` in the call to `moduleExportNameIsDefault`.

src/compiler/transformers/module/module.ts
```diff
@@ -1613,16 +1625,27 @@ export function transformModule(context: TransformationContext): (x: SourceFile
...
-                    idText(specifier.propertyName || specifier.name) === "default";
-                const exportedValue = factory.createPropertyAccessExpression(
-                    exportNeedsImportDefault ? emitHelpers().createImportDefaultHelper(generatedName) : generatedName,
-                    specifier.propertyName || specifier.name,
-                );
+                    moduleExportNameIsDefault(specifier.propertyName || specifier.name);
+                const specifierName = specifier.propertyName || specifier.name;
+                const exportedValue = specifierName.kind === SyntaxKind.StringLiteral
+                    ? factory.createElementAccessExpression(
+                        exportNeedsImportDefault ? emitHelpers().createImportDefaultHelper(generatedName) : generatedName,
```
rbuckton, line 1632: Since this is the same as the expression on L1636, if you extract this line out to its own variable and use it in both branches then this statement won't require quite as much nesting.

src/compiler/transformers/module/module.ts
```diff
@@ -2343,10 +2375,15 @@ export function transformModule(context: TransformationContext): (x: SourceFile
                 else if (isImportSpecifier(importDeclaration)) {
                     const name = importDeclaration.propertyName || importDeclaration.name;
                     return setTextRange(
-                        factory.createPropertyAccessExpression(
-                            factory.getGeneratedNameForNode(importDeclaration.parent?.parent?.parent || importDeclaration),
-                            factory.cloneNode(name),
-                        ),
+                        name.kind === SyntaxKind.StringLiteral
+                            ? factory.createElementAccessExpression(
+                                factory.getGeneratedNameForNode(importDeclaration.parent?.parent?.parent || importDeclaration),
```
rbuckton, line 2380: This line can be extracted into an independent variable and reused on L2384 to cut down on nesting.

src/compiler/transformers/module/system.ts
```diff
@@ -1853,13 +1856,19 @@ export function transformSystemModule(context: TransformationContext): (x: Sourc
...
                     return setTextRange(
                         factory.createPropertyAssignment(
                             factory.cloneNode(name),
-                            factory.createPropertyAccessExpression(
-                                factory.getGeneratedNameForNode(importDeclaration.parent?.parent?.parent || importDeclaration),
-                                factory.cloneNode(importDeclaration.propertyName || importDeclaration.name),
-                            ),
+                            importedName.kind === SyntaxKind.StringLiteral
+                                ? factory.createElementAccessExpression(
+                                    factory.getGeneratedNameForNode(importDeclaration.parent?.parent?.parent || importDeclaration),
```
rbuckton, line 1865: This can be extracted to an independent variable and reused on L1869.

src/compiler/transformers/module/system.ts
```diff
@@ -1921,11 +1930,17 @@ export function transformSystemModule(context: TransformationContext): (x: Sourc
...
                 else if (isImportSpecifier(importDeclaration)) {
+                    const importedName = importDeclaration.propertyName || importDeclaration.name;
                     return setTextRange(
-                        factory.createPropertyAccessExpression(
-                            factory.getGeneratedNameForNode(importDeclaration.parent?.parent?.parent || importDeclaration),
-                            factory.cloneNode(importDeclaration.propertyName || importDeclaration.name),
-                        ),
+                        importedName.kind === SyntaxKind.StringLiteral
+                            ? factory.createElementAccessExpression(
+                                factory.getGeneratedNameForNode(importDeclaration.parent?.parent?.parent || importDeclaration),
```
rbuckton, line 1937: This can be extracted to an independent variable and reused on L1941.

tests/baselines/reference/arbitraryModuleNamespaceIdentifiers_syntax.errors.txt
```diff
@@ -0,0 +1,70 @@
...
+    export type { foo as "valid 3" } from "./types1";
+    export type * as "valid 4" from "./types1";
+    
+    // Invalid
+    import type { "invalid 1" } from "./types1";
+                  ~~~~~~~~~~~
+!!! error TS2305: Module '"./types1"' has no exported member '"invalid 1"'.
+                              ~
+!!! error TS1005: 'as' expected.
+    import type { foo as "invalid 2" } from "./types1";
```
rbuckton, line 47: This should be an error.

tests/baselines/reference/arbitraryModuleNamespaceIdentifiers_syntax.errors.txt
```diff
@@ -0,0 +1,70 @@
...
+    export { foo as "valid 3" } from "./values";
+    export * as "valid 4" from "./values";
+    
+    // Invalid
+    import { "invalid 1" } from "./values";
+             ~~~~~~~~~~~
+!!! error TS2305: Module '"./values"' has no exported member '"invalid 1"'.
+                         ~
+!!! error TS1005: 'as' expected.
+    import { foo as "invalid 2" } from "./values";
```
rbuckton, line 27: This should be an error.

tests/baselines/reference/arbitraryModuleNamespaceIdentifiers_syntax.errors.txt
```diff
@@ -0,0 +1,70 @@
...
+    export type foo = 123;
+    export { type foo as "valid 1" };
+    import { type "valid 1" as bar } from "./types2";
+    export { type "valid 1" as "valid 2" } from "./types2";
+    export { type foo as "valid 3" } from "./types2";
+    
+    // Invalid
+    import { type "invalid 1" } from "./types2";
+                  ~~~~~~~~~~~
+!!! error TS2305: Module '"./types2"' has no exported member '"invalid 1"'.
```
rbuckton, line 63: This should be a grammar error, not an error referencing a missing export. You can't import a string literal name without an identifier binding.

tests/baselines/reference/arbitraryModuleNamespaceIdentifiers_syntax.errors.txt
```diff
@@ -0,0 +1,70 @@
...
+                              ~
+!!! error TS1005: 'as' expected.
+    import { type foo as "invalid 2" } from "./types2";
```
rbuckton, line 66: This should be an error.

tests/cases/conformance/es2022/arbitraryModuleNamespaceIdentifiers/arbitraryModuleNamespaceIdentifiers_system.ts
```diff
@@ -0,0 +1,33 @@
+//@module: System
```
rbuckton, line 1: You don't need individual tests for each module type. You can just use comma-separated variations in the twoslash comment and we'll run the test once for each variation:

For all possible values:
```js
// @module: *
```

For specific values:
```js
// @module: amd,commonjs,system,umd,es2015
```

For all values excluding specific values:
```js
// @module: -esnext,-node16
```

rbuckton (changes requested):

evanw (commented):
Many thanks for the review!

Edit: All comments should now be addressed.

tests/baselines/reference/arbitraryModuleNamespaceIdentifiers_syntax.errors.txt
rbuckton: Ah, I did miss the `as` error. Disregard this one.

rbuckton (approved):

rbuckton:
This looks good but may need to wait to merge until after we switch `main` to target 5.6. I'm going to modify the PR description to use `Fixes #<issue>` for each issue instead of the bulleted list since GH's issue linking and auto-close functionality doesn't understand bulleted lists.

rbuckton:
@typescript-bot perf test
@typescript-bot run dt
@typescript-bot test top400

PR #58641, opened 2024-05-24
Compact the tsbuild info by encoding info differently for some of the situations

sheetalkamat:
- For any diagnostics per file records, dont record file name if it is same as the key - file for which the diagnostics is recorded 8df8f4bdcd93d5571193c90bae2e757eedb78c7e
   So now the reusable diagnostics will have key as 
    - undefined if its same as file for which the diagnostics is
    - false if it was global error - though we dont have anything like this
    - filePath if the diagnostic or related information in the diagnostics is in different file
- Change recording of semantic diagnostics to reduce size in normal scenarios:
   It will now be recorded for each file as: 666f49f058e087246ece03fcf65d2aaa1586c769
   - If emptyArray - dont record 
   - If diagnsotics - record as [ fileId diagnostics]
    - If not cached for the file thats not in changeSet - record as fileId
- Dont record empty referenceMap since we know when to create referenceMap 8fc726898a79f426c94941d546a8232e7bafe265


lazytype:
Would older versions of TypeScript gracefully handle tsbuild info files created with the new TypeScript version? In a large, shared code base it may be common for developers to switch between branches which are on different TS versions.

sheetalkamat:
The buildInfos are invalidated between version change which means they arent reused and are regenerated if building with different version

jakebailey (approved):
Nice!

PR #58654, opened 2024-05-24
Check resolution of tslib per file

gabritto:
This issue was raised by @andrewbranch during design meeting.

When the compiler option [`importHelpers`](https://www.typescriptlang.org/tsconfig/#importHelpers) is on, when we emit a module file, we will insert an appropriate import from `tslib` when an emit helper is needed.
To make sure this emitted import will be valid, during type checking, we try to resolve `tslib` whenever we detect we will need an emit helper, and error if we can't resolve it.
However, that resolution and respective error is currently not done per every file that will need the emit helper during emit, it is only done once per type checker and the resolution result is cached. This means that we could be missing errors in situations where for the first file we check `tslib` resolution, it resolves, but for a future file it would not resolve.

The newly added test case `tslibNotFoundDifferentModules` illustrates this: we have two packages in a project, one of them (`package1`) that has `tslib` in its `node_modules/`, and the other (`package2`) that doesn't. We first check the file in `project1`, and there we are able to resolve `tslib`, and so when we go and check the file in `project2`, we don't error saying `tslib` is missing, but we should.

This PR fixes this problem by making this `tslib` resolution check for every file an emit helper will be used.

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -1462,8 +1462,6 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     var cancellationToken: CancellationToken | undefined;
 
     var requestedExternalEmitHelperNames = new Set<string>();
-    var requestedExternalEmitHelpers: ExternalEmitHelpers;
-    var externalHelpersModule: Symbol;
     var scanner: Scanner | undefined;
 
     var Symbol = objectAllocator.getSymbolConstructor();
@@ -49508,12 +49506,14 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     }
 
     function checkExternalEmitHelpers(location: Node, helpers: ExternalEmitHelpers) {
-        if ((requestedExternalEmitHelpers & helpers) !== helpers && compilerOptions.importHelpers) {
-            const sourceFile = getSourceFileOfNode(location);
-            if (isEffectiveExternalModule(sourceFile, compilerOptions) && !(location.flags & NodeFlags.Ambient)) {
+        const sourceFile = getSourceFileOfNode(location);
+        if (isEffectiveExternalModule(sourceFile, compilerOptions) && !(location.flags & NodeFlags.Ambient)) {
+            const links = getNodeLinks(sourceFile);
+            links.requestedExternalEmitHelpers ??= 0 as ExternalEmitHelpers;
+            if ((links.requestedExternalEmitHelpers & helpers) !== helpers && compilerOptions.importHelpers) {
                 const helpersModule = resolveHelpersModule(sourceFile, location);
                 if (helpersModule !== unknownSymbol) {
-                    const uncheckedHelpers = helpers & ~requestedExternalEmitHelpers;
+                    const uncheckedHelpers = helpers & ~links.requestedExternalEmitHelpers;
                     for (let helper = ExternalEmitHelpers.FirstEmitHelper; helper <= ExternalEmitHelpers.LastEmitHelper; helper <<= 1) {
                         if (uncheckedHelpers & helper) {
                             for (const name of getHelperNames(helper)) {
@@ -49543,7 +49543,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                         }
                     }
                 }
-                requestedExternalEmitHelpers |= helpers;
+                links.requestedExternalEmitHelpers |= helpers;
             }
         }
     }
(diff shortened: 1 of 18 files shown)


gabritto:
@typescript-bot test it

src/compiler/checker.ts
```diff
@@ -49508,12 +49506,14 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+        const sourceFile = getSourceFileOfNode(location);
+        if (isEffectiveExternalModule(sourceFile, compilerOptions) && !(location.flags & NodeFlags.Ambient)) {
+            const links = getNodeLinks(sourceFile);
+            links.requestedExternalEmitHelpers ??= 0 as ExternalEmitHelpers;
+            if ((links.requestedExternalEmitHelpers & helpers) !== helpers && compilerOptions.importHelpers) {
                 const helpersModule = resolveHelpersModule(sourceFile, location);
                 if (helpersModule !== unknownSymbol) {
-                    const uncheckedHelpers = helpers & ~requestedExternalEmitHelpers;
+                    const uncheckedHelpers = helpers & ~links.requestedExternalEmitHelpers;
                     for (let helper = ExternalEmitHelpers.FirstEmitHelper; helper <= ExternalEmitHelpers.LastEmitHelper; helper <<= 1) {
```
andrewbranch, line 49517: If I’m understanding this correctly, this checking is now happening once for every source file that needs to access helpers, but the checking itself only depends on the `helpersModule`. Maybe `requestedExternalEmitHelpers` could be stored on `getSymbolLinks(helpersModule)` instead of `getNodeLinks(sourceFile)` and only happen once per instance of helpers.

However, the _error reporting_ does reference the location being passed in, and maybe that matters for your region-based diagnostics work? Although, we’re still only doing this once per requested helper function per source file, so the location of the errors can move 🤔

gabritto: > If I’m understanding this correctly, this checking is now happening once for every source file that needs to access helpers, but the checking itself only depends on the `helpersModule`. Maybe `requestedExternalEmitHelpers` could be stored on `getSymbolLinks(helpersModule)` instead of `getNodeLinks(sourceFile)` and only happen once per instance of helpers.

Good point. So `requestedExternalEmitHelperNames` should also be stored via symbol links, instead of also being a global, because now `helpersModule` could be different symbols for the same type checker.

> However, the _error reporting_ does reference the location being passed in, and maybe that matters for your region-based diagnostics work? Although, we’re still only doing this once per requested helper function per source file, so the location of the errors could be check order dependent both before and after this PR, as it currently is 🤔

Yeah, I'm planning on dealing with the error location varying in a future PR. But with this PR, we already address the problem of this error possibly jumping to a different, unopened file and never showing up for the user in the editor.

andrewbranch: > So then `requestedExternalEmitHelperNames` would be irrelevant.

Hmm, would it? A given instance of a helpers module might have checked `__extends` in one call, and then the same instance needs to check `__assign` in a later call, and ought to be able to skip checking `__extends` again.

gabritto: Ah, I mean `requestedExternalEmitHelperNames` and `requestedExternalEmitHelpers` should represent the same set, if `requestedExternalEmitHelpers` is now stored at the symbol level, so we only need one or the other.

gabritto: Ok, so I think I confused myself again.

Say we have two different typescript files that need the emit helper called `__awaiter`. Say those two files share the same `tslib`. Should we have an error on both typescript files saying their `tslib` is missing the emit helper called `__awaiter`? Or only on one of the files?

andrewbranch: I think it would be annoying on the CLI, or in any list of project-wide diagnostics, to get tslib-related errors duplicated across many/all files. I’m not sure how we might want to handle that in an editor scenario, though.

rbuckton: > I think it would be annoying on the CLI, or in any list of project-wide diagnostics, to get tslib-related errors duplicated across many/all files. 

That, at least, matches what we currently do (i.e., only error on the first occurrence).

src/compiler/checker.ts
```diff
@@ -49508,12 +49506,14 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     }
 
     function checkExternalEmitHelpers(location: Node, helpers: ExternalEmitHelpers) {
-        if ((requestedExternalEmitHelpers & helpers) !== helpers && compilerOptions.importHelpers) {
-            const sourceFile = getSourceFileOfNode(location);
-            if (isEffectiveExternalModule(sourceFile, compilerOptions) && !(location.flags & NodeFlags.Ambient)) {
+        const sourceFile = getSourceFileOfNode(location);
```
rbuckton, line 49509: Could we preserve the `if (compilerOptions.importHelpers)` condition here? By moving the check to line 49513 you're doing a lot of extra work, including needlessly increasing the size of the node links table and defining the `requestedExternalEmitHelpers` property.

rbuckton (changes requested):

src/compiler/checker.ts
gabritto: I added test  `tslibMultipleMissingHelper.ts` that shows the specific helper checking behavior.

gabritto:
@typescript-bot test it

tests/cases/compiler/tslibMultipleMissingHelper.ts
```diff
@@ -0,0 +1,44 @@
...
+        "target": "ES2016",
+        "importHelpers": true,
+        "module": "commonjs",
+    }
+}
+
+// @filename: /package1/index.ts
+export {};
+async function foo(): Promise<void> {}
+async function bar(): Promise<void> {}
```
andrewbranch, line 14: Maybe put another file that needs a helper inside `package1` so we can see (if I understand correctly) that the error doesn’t duplicate across _files_ that resolve to the same copy of `tslib`?

andrewbranch (approved):

src/compiler/checker.ts
rbuckton: I still think this should move up before calling `getSourceFileOfNode` to avoid the unnecessary walk up the spine.

rbuckton (changes requested):

src/compiler/checker.ts
gabritto: Done.

rbuckton (approved):

Issue #58653, opened 2024-05-24
JSDoc `@private` on constructor without parameters is omitted in declaration files


Gerrit0:
### 🔎 Search Terms

jsdoc private constructor

### 🕗 Version & Regression Information

- This is the behavior in every version I tried back through TS 3.8 where JSDoc support was added, and I reviewed the FAQ for entries about JSDoc

### ⏯ Playground Link

https://www.typescriptlang.org/play/?target=99&module=1&ts=5.5.0-beta&ssl=8&ssc=10&pln=1&pc=1&filetype=js#code/MYGwhgzhAEBiD29oG8BQ0PQPQCofs2h2gAEAHAJwEsA3MAFwFMDMcsXph4A7CeigK7B68CgAoAlCgC+qWam6MA7nESSgA

### 💻 Code

```js
class Foo {
    /**
     * @private
     */
    constructor() {}
}

new Foo()
// Constructor of class 'Foo' is private and only accessible within the class declaration.(2673)
```

### 🙁 Actual behavior

Emits the declaration file:
```ts
declare class Foo {
}
```

### 🙂 Expected behavior

Should emit the declaration file:
```ts
declare class Foo {
    /**
     * @private
     */
    private constructor();
}
```

### Additional information about the issue

I ran into this as it was requested that TypeDoc add a new JSDoc tag to work around this issue. This seems wrong since the declaration file doesn't accurately describe the types within the file, but it's also very unfortunate that it results in the loss of any documentation added to the constructor.

https://github.com/TypeStrong/typedoc/issues/2577

PR #58655, opened 2024-05-24
Make sure that private constructors are always emitted in declarations

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/58653

### DIFF
--- a/tests/baselines/reference/privateConstructorJsDocDeclarationEmit.js
+++ b/tests/baselines/reference/privateConstructorJsDocDeclarationEmit.js
@@ -1,0 +1,41 @@
+//// [tests/cases/conformance/salsa/privateConstructorJsDocDeclarationEmit.ts] ////
+
+//// [privateConstructorJsDocDeclarationEmit.js]
+class Foo {
+    /**
+     * @private
+     */
+    constructor() { }
+}
+
+class Bar {
+    constructor() { }
+}
+
+class Baz extends Bar {
+    /**
+     * @private
+     */
+    constructor() {
+        super()
+    }
+}
+
+
+
+
+//// [privateConstructorJsDocDeclarationEmit.d.ts]
+declare class Foo {
+    /**
+     * @private
+     */
+    private constructor();
+}
+declare class Bar {
+}
+declare class Baz extends Bar {
+    /**
+     * @private
+     */
+    private constructor();
+}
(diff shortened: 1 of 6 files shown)


DanielRosenwasser:
@typescript-bot pack this

DanielRosenwasser:
It arguably has broken behavior but can you also add the following test too?

```ts
class C {
    /**
     * @private
     * @overload
     * @param x {string}
     * @param y {string}
     */

    /**
     * @private
     * @overload
     * @param x {number}
     * @param y {number}
     */

    /**
     * @private
     * @param {...any} args 
     */
    constructor(...args) {

    }
}
```

Gerrit0:
It'd be kind of nice to emit constructors even if they're public if they have a documentation comment, presumably the user wrote things there for a reason, and it'd be nice for downstream users to see it

tests/baselines/reference/privateConstructorJsDocDeclarationEmit.js
```diff
@@ -0,0 +1,82 @@
...
+    private constructor();
+}
+declare class C {
+    /**
+     * @private
+     * @overload
+     * @param x {string}
+     * @param y {string}
+     */
+    constructor(x: string, y: string);
```
sheetalkamat, line 74: This is not correct because we shouldnt be emitting types for private - this could reference private type

Andarist: yes, this is invalid but it's a different bug - this test case has been suggested by @DanielRosenwasser and I'm investigating now (separately) what can be done to fix this.

The problem is not about what gets emitted here but about the fact that it's not recognized as private at all. Notice how there is no error here: [TS playground](https://www.typescriptlang.org/play/?ts=5.5.0-dev.20240525&filetype=js#code/MYGwhgzhAEDC0G8BQ1XQPQCpMrdT0AAgA4BOAlgG5gAuAprmgYQPaV2kgtgAmjqzYmFJgAttAAeiCDQoA7AOYBffviJCR4gJ7TZ5RSrz50SflhxHBFavVXM2HLrzvrhYyYjkBXUQCMOhniCbtqePv6kgUwmZtguJNa0DJaumogAdJlgclpK0MIKMHYmeMAscjKkXsA0LKQAFJnpBRAAlIimeCoqSHJ0AO5w9QCMADTQw62m06ZAA)

PR #58662, opened 2024-05-26
Fixed an issue with private overloaded JSDoc constructors being callable

Andarist:
fixes what @DanielRosenwasser found [here](https://github.com/microsoft/TypeScript/pull/58655#issuecomment-2130552519)

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -589,6 +589,7 @@ import {
     isIntrinsicJsxName,
     isInTypeQuery,
     isIterationStatement,
+    isJSDoc,
     isJSDocAllType,
     isJSDocAugmentsTag,
     isJSDocCallbackTag,
@@ -1105,7 +1106,6 @@ import {
     WideningContext,
     WithStatement,
     YieldExpression,
-    isJSDoc,
 } from "./_namespaces/ts.js";
 import * as moduleSpecifiers from "./_namespaces/ts.moduleSpecifiers.js";
 import * as performance from "./_namespaces/ts.performance.js";


tests/baselines/reference/jsFileAlternativeUseOfOverloadTag.js
```diff
@@ -175,4 +175,17 @@ declare namespace example3 {
  * function (error, result)
  *     If callback is provided it will be called with evaluation result
  */
-type callback = (error: any, result: any) => any;
```
Andarist, line 178: This change isn't great but tbh the behavior for a scenario like this looks underspecified to me. We can even read this at the top of the relevant test file:
```ts
// These are a few examples of existing alternative uses of @overload tag.
// They will not work as expected with our implementation, but we are
// trying to make sure that our changes do not result in any crashes here.
```

And even here we can't type this implicit~ `callback` type in full as it complains about "'return' tag already specified.(1223)"
```ts
const example4 = {
  /**
   *   @param options {Record<string, any>}
   *   @param callback {callback}
   *   @return {string}
   *   @callback callback
   *     @param error {Error}
   *     @param result {string}
   *     @return {number}
   */
  evaluate: function evaluate(options, callback) {
    return ''
  },
};
```

tests/baselines/reference/quickInfoJsDocTags13.baseline
```diff
@@ -25,12 +25,14 @@
 // | ----------------------------------------------------------------------
 // | f(**a: number**): void
 // | First overload
+// | @overload
```
Andarist, line 28: often "redundant" `@param` (and more are rendered) - is this any different? I think it's fine to leave it as-is right now and redundant tags could be tackled separately if that's a desired feature.

tests/baselines/reference/signatureHelpJSDocCallbackTag.baseline
```diff
@@ -25,17 +25,17 @@
 //   ^
 // | ----------------------------------------------------------------------
 // | t(**eventName: string**, eventName2: number | string, eventName3: any): number
-// | @type {FooHandler} callback
+// | - A kind of magic
```
Andarist, line 28: this feels like an (unforeseen) improvement :)

Issue #58600, opened 2024-05-21
tsc crash with `Error: start < 0` in a specific config


yume-chan:
### 🔎 Search Terms

crash createTextSpan

### 🕗 Version & Regression Information

- This is a crash

### 💻 Code

https://github.com/yume-chan/ts-crash

This is a minimal repro with only 15 lines across 3 files.

### 🙁 Actual behavior

```sh
pnpm i
npx tsc
```

Output:

```
# npx tsc
/root/ts-crash/node_modules/.pnpm/typescript@5.5.0-dev.20240520/node_modules/typescript/lib/tsc.js:11034
    throw new Error("start < 0");
    ^

Error: start < 0
    at createTextSpan (/root/ts-crash/node_modules/.pnpm/typescript@5.5.0-dev.20240520/node_modules/typescript/lib/tsc.js:11034:11)
    at createTextSpanFromBounds (/root/ts-crash/node_modules/.pnpm/typescript@5.5.0-dev.20240520/node_modules/typescript/lib/tsc.js:11042:10)
    at getErrorSpanForNode (/root/ts-crash/node_modules/.pnpm/typescript@5.5.0-dev.20240520/node_modules/typescript/lib/tsc.js:13589:10)
    at createDiagnosticForNodeFromMessageChain (/root/ts-crash/node_modules/.pnpm/typescript@5.5.0-dev.20240520/node_modules/typescript/lib/tsc.js:13419:16)
    at resolveExternalModule (/root/ts-crash/node_modules/.pnpm/typescript@5.5.0-dev.20240520/node_modules/typescript/lib/tsc.js:48013:31)
    at resolveExternalModuleNameWorker (/root/ts-crash/node_modules/.pnpm/typescript@5.5.0-dev.20240520/node_modules/typescript/lib/tsc.js:47907:61)
    at resolveExternalModuleName (/root/ts-crash/node_modules/.pnpm/typescript@5.5.0-dev.20240520/node_modules/typescript/lib/tsc.js:47904:12)
    at getSymbolAtLocation (/root/ts-crash/node_modules/.pnpm/typescript@5.5.0-dev.20240520/node_modules/typescript/lib/tsc.js:84201:18)
    at Object.getSymbolAtLocation (/root/ts-crash/node_modules/.pnpm/typescript@5.5.0-dev.20240520/node_modules/typescript/lib/tsc.js:45191:21)
[...]

Issue #50394, opened 2022-08-21
`Node16` module resolution doesn't check side effect imports


alecmev:
# Bug Report

### 🕗 Version & Regression Information

This is the behavior in every version I tried, and I reviewed the FAQ for entries about this.

### ⏯ Playground Link

[Playground link with relevant code](https://www.typescriptlang.org/play?target=99&module=100#code/JYWwDg9gTgLgBAIgHQHoBmwA2BTBBuAWACgg)

### 💻 Code

```ts
// @module: Node16
import "./file";
```

### 🙁 Actual behavior

TypeScript says nothing.

### 🙂 Expected behavior

I want TypeScript to tell me that `.js` is missing, the same way it does in any other type of import. I think I understand why it doesn't right now, because side effects have nothing to do with static analysis, so the incorrectness of the import doesn't prevent TypeScript from doing its job in any way, but:

1. I haven't been able to set this up in the bug workbench, but what happens if the file is being imported from `node_modules` and has a corresponding `.d.ts` with a `declare ...` in it? Will that `declare` get applied?
2. Wouldn't this be useful to have either way? I guess ESLint could fill this void, but [seeing how `eslint-plugin-import` is dragging its feet on `import/extensions`](https://github.com/import-js/eslint-plugin-import/issues/2111), there's little hope for that.

Apologies if this has been addressed before, couldn't find.

DanielRosenwasser:
I believe that this has come up with @andrewbranch recently. I think that Node16 users probably want to get an error under this mode, but it's hard to say - plus those users need an alternative option if they are trying to get export maps working in conjunction with a bundler.

fatcerberus:
> I think that Node16 users probably want to get an error under this mode, but it's hard to say - plus those users need an alternative option if they are trying to get export maps working in conjunction with a bundler.

wait, I’m confused - why is this a concern only for side-effect imports?

andrewbranch:
I also don’t understand that sentence 🤔

andrewbranch:
The reason why this isn’t a no-brainer is that it’s common to do

```
import "./styles.css"
```

with certain bundler configurations in order to add globally-applicable CSS to the bundle. We don’t check existence of files with arbitrary extensions, so in order to make this resolve, you’d have to add a pattern ambient module for `".*css"` or a file at `styles.d.css.ts` via #51435. But both of these feel a bit pointless for a side-effect import, since the types written in that declaration will never be observed.

[...]

PR #28558, opened 2018-11-15
Dont emit module resolution errors when looking up specifiers for container symbols

weswigham:
Fixes #28538

Although I really do think it's strange we don't error on these kinds of imports.

PR #58668, opened 2024-05-27
Consistently avoid module resolution errors when using `getSymbolAtLocation`

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/58600


src/compiler/checker.ts
```diff
@@ -4521,17 +4521,17 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         const errorMessage = isClassic ?
             Diagnostics.Cannot_find_module_0_Did_you_mean_to_set_the_moduleResolution_option_to_nodenext_or_to_add_aliases_to_the_paths_option
             : Diagnostics.Cannot_find_module_0_or_its_corresponding_type_declarations;
-        return resolveExternalModuleNameWorker(location, moduleReferenceExpression, ignoreErrors ? undefined : errorMessage);
+        return resolveExternalModuleNameWorker(location, moduleReferenceExpression, ignoreErrors ? undefined : errorMessage, ignoreErrors);
```
Andarist, line 4524: This is the core of the fix. It propagates the `ignoreErrors` and the intention to ignore them is highlighted in the comment here:
https://github.dev/microsoft/TypeScript/blob/af3a61fe4487a92d59f9479aa4249d897b91af14/src/compiler/checker.ts#L1654-L1658

src/compiler/checker.ts
```diff
@@ -46808,6 +46812,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         }
         if (checkExternalImportOrExportDeclaration(node)) {
             const importClause = node.importClause;
+            const moduleExisted = resolveExternalModuleName(node, node.moduleSpecifier);
```
Andarist, line 46815: After the fix, some errors went away. Upon further inspection, I found that this code path was not resolving module names for specifier-less imports like:
```ts
import 'stuff'
```

So to bring back the errors that were already there I fixed this. One thing to notice about this is that those errors were reported **inconsistently**. They were reported for incremental/composite builds as that led to resolving everything to build the program's state and its `referenceMap`. They were not reported for normal builds.

However, with many baseline changes here we can see that this could be considered a breaking change. I think it would be a good "breaking change" here. The fact that the errors were reported inconsistently was weird and I doubt it was desired.

To quote @weswigham from https://github.com/microsoft/TypeScript/pull/28558 :
> Although I really do think it's strange we don't error on these kinds of imports.

src/testRunner/unittests/programApi.ts
```diff
@@ -216,14 +216,17 @@ describe("unittests:: programApi:: Program.getTypeChecker / Program.getSemanticD
         const sourceFile = program.getSourceFile("main.ts")!;
         const typeChecker = program.getTypeChecker();
         typeChecker.getSymbolAtLocation((sourceFile.statements[0] as ts.ImportDeclaration).moduleSpecifier);
-        assert.isEmpty(program.getSemanticDiagnostics());
+        const diagnostics = program.getSemanticDiagnostics()
+        assert.equal(diagnostics.length, 1);
+        assert.equal(diagnostics[0].code, ts.Diagnostics.File_0_is_not_a_module.code);
+        assert.equal(diagnostics[0].messageText, "File '/module.d.ts' is not a module.");
```
Andarist, line 222: I think, this stays true to the intention of the test case:
> getSymbolAtLocation does not cause additional error to be added on module resolution failure

src/testRunner/unittests/tsc/composite.ts
```diff
@@ -114,4 +114,58 @@ describe("unittests:: tsc:: composite::", () => {
             },
         ],
     });
+
+    verifyTsc({
```
Andarist, line 118: those are tests related to the referenced issue

tests/baselines/reference/moduleAugmentationGlobal5.js
```diff
@@ -32,3 +32,32 @@ require("B");
...
+f3.d.ts(2,8): error TS2307: Cannot find module 'B' or its corresponding type declarations.
+
+
+==== f3.d.ts (2 errors) ====
+    import "A";
+           ~~~
+!!! error TS2307: Cannot find module 'A' or its corresponding type declarations.
+    import "B";
+           ~~~
+!!! error TS2307: Cannot find module 'B' or its corresponding type declarations.
```
Andarist, line 50: I think this shows an extra bug that was here before this PR but one that slipped through the cracks as those errors were not surfaced. 

Those modules are declared by other files and the original `.ts` file here referenced them using tripleslash directives. Those directives were pruned from the resulting `.d.ts` so the declaration set can't locate those modules now.

tests/baselines/reference/tscWatch/resolveJsonModule/incremental-always-prefers-declaration-file-over-document.js
```diff
@@ -103,6 +108,11 @@ var x = data_json_1.default;
         "./main.ts"
       ]
     ],
+    "referencedMap": {
+      "./main.ts": [
+        "./data.d.json.ts"
+      ]
+    },
```
Andarist, line 115: This feels like an improvement - even though there is a resolution error here, it still depends on this file so it should be reflected here. IIRC, there are other instances of something like this - when despite a resolution errors requested modules are listed here.

src/compiler/checker.ts
Andarist: Yee, I have some Babel background and those are specifiers there: https://babeljs.io/docs/babel-types#importdeclaration ;p

> Not erroring on side-effect-only imports that don't resolve is intentional and issues are indeed reported about it from time to time. Explanation of the behavior is here: https://github.com/microsoft/TypeScript/issues/50394#issuecomment-1381014487

Thanks for the pointers. I think that, at the very least, we should issue errors on `import './foo.ts'` with `allowImportingTsExtensions: false`. That is what originally prompted me to introduce this change - such an error was removed from the composite test suite.

In addition to that, it feels like `import 'foo'` is also questionable... as the user, I would really like to get an error on this if I forgot to install `foo`. In some sense, I feel like this could only allow such side-effects imports for unknown extensions.

Also, this current behavior feels weird:
```ts
import 'foo' // no error
import {} from 'foo' // error
```

fatcerberus: Huh, Babel is being weird here because in ECMAScript the thing called "specifier" is very specifically (no pun intended) the module name:

ehoogeveen-medweb: I don't know what the initial source of this terminology is, but all ASTs seem to use `source` for the thing being imported from and `specifiers` for the thing(s) being imported (e.g. check `import foo from 'bar'` on https://astexplorer.net/)

jakebailey: Not responding these types of imports is a designed behavior, fwiw. I don't like it and want to change it too, but we'd break everyone importing css files and so on. So I don't think this is in scope for a bugfix PR

Andarist: I reverted those changes from here

tests/baselines/reference/tscWatch/programUpdates/when-changing-`allowImportingTsExtensions`-of-config-file.js
```diff
@@ -47,12 +47,7 @@ DirectoryWatcher:: Added:: WatchInfo: /user/username/projects/myproject/node_mod
 Elapsed:: *ms DirectoryWatcher:: Added:: WatchInfo: /user/username/projects/myproject/node_modules/@types 1 undefined Type roots
 DirectoryWatcher:: Added:: WatchInfo: /user/username/projects/node_modules/@types 1 undefined Type roots
 Elapsed:: *ms DirectoryWatcher:: Added:: WatchInfo: /user/username/projects/node_modules/@types 1 undefined Type roots
-[96mb.ts[0m:[93m1[0m:[93m8[0m - [91merror[0m[90m TS5097: [0mAn import path can only end with a '.ts' extension when 'allowImportingTsExtensions' is enabled.
-
-[7m1[0m import "./a.ts";
-[7m [0m [91m       ~~~~~~~~[0m
-
-[[90mHH:MM:SS AM[0m] Found 1 error. Watching for file changes.
```
Andarist, line 55: I think this error should be reported but currently it was reported here *accidentally*. It's not consistent with a regular - non-incremental/composite executions. I raised an issue about it here: https://github.com/microsoft/TypeScript/issues/58725

jakebailey:
@typescript-bot test it
@typescript-bot pack this

tests/baselines/reference/tscWatch/programUpdates/when-changing-`allowImportingTsExtensions`-of-config-file.js
jakebailey: FWIW resolving these would be https://github.com/microsoft/TypeScript/issues/50394

jakebailey (approved):

jakebailey:
This appears to be out of date with main.

tests/baselines/reference/tscWatch/programUpdates/when-changing-`allowImportingTsExtensions`-of-config-file.js
```diff
@@ -47,12 +47,7 @@ DirectoryWatcher:: Added:: WatchInfo: /user/username/projects/myproject/node_mod
 Elapsed:: *ms DirectoryWatcher:: Added:: WatchInfo: /user/username/projects/myproject/node_modules/@types 1 undefined Type roots
 DirectoryWatcher:: Added:: WatchInfo: /user/username/projects/node_modules/@types 1 undefined Type roots
 Elapsed:: *ms DirectoryWatcher:: Added:: WatchInfo: /user/username/projects/node_modules/@types 1 undefined Type roots
-[96mb.ts[0m:[93m1[0m:[93m8[0m - [91merror[0m[90m TS5097: [0mAn import path can only end with a '.ts' extension when 'allowImportingTsExtensions' is enabled.
```
sheetalkamat, line 50: This test case needs to change to import * as a from "./a.ts" so the errors are shown

sheetalkamat (changes requested):

tests/baselines/reference/tscWatch/programUpdates/when-changing-`allowImportingTsExtensions`-of-config-file.js
jakebailey: Why? Isn't this effectively for #58791 about side effect imports, which are not resolved today without #50394/#58941?

sheetalkamat: This was testing "allowTsExtension" has effect on change and i would prefer if there are errors before and after so that its clear

jakebailey: Aha, yes, gotcha.

jakebailey: In that case I would also actually duplicate the test and keep the side-effect import, since this did seem to show a bug.

Andarist: I added the duplicated test case that verifies the error goes away with a change from `allowImportingTsExtensions: false` to `allowImportingTsExtensions: true`

sheetalkamat (approved):

PR #58669, opened 2024-05-27
Fixed an accidental `undefined` leak into `getAccessibleSymbolChain`'s cache key

Andarist:


src/compiler/checker.ts
```diff
@@ -5473,7 +5473,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         const cache = (links.accessibleChainCache ||= new Map());
         // Go from enclosingDeclaration to the first scope we check, so the cache is keyed off the scope and thus shared more
         const firstRelevantLocation = forEachSymbolTableInScope(enclosingDeclaration, (_, __, ___, node) => node);
-        const key = `${useOnlyExternalAliasing ? 0 : 1}|${firstRelevantLocation && getNodeId(firstRelevantLocation)}|${meaning}`;
+        const key = `${useOnlyExternalAliasing ? 0 : 1}|${firstRelevantLocation ? getNodeId(firstRelevantLocation) : 0}|${meaning}`;
```
Andarist, line 5476: I noticed that those keys could look smth like `'0|undefined|788968'` (I spotted an instance of that in the debugger). It bugged me enough to create this PR ;p 

0 should be a safe fallback here as the lowest possible node id is 1

jakebailey:
@typescript-bot test it

Double checking this isn't depended on accidentally.

jakebailey (approved):

PR #58679, opened 2024-05-28
Handling statements from a known source file

navya9singh:
While testing copy/paste functionality, it was found that imports were not getting adding from a file. The reason behind this was because the right statements were not getting extracted from the `copiedFrom` file. Whenever there was a newline, the line number and position would cause a problem in extracting the statements. This pr fixes that problem.

### DIFF
--- a/src/services/pasteEdits.ts
+++ b/src/services/pasteEdits.ts
@@ -1,4 +1,7 @@
-import { addRange } from "../compiler/core.js";
+import {
+    addRange,
+    findIndex,
+} from "../compiler/core.js";
 import {
     CancellationToken,
     Program,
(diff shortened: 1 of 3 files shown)


src/services/pasteEdits.ts
```diff
@@ -77,7 +76,11 @@ function pasteEdits(
         if (copiedFrom?.range) {
             Debug.assert(copiedFrom.range.length === pastedText.length);
             copiedFrom.range.forEach(copy => {
-                addRange(statements, copiedFrom.file.statements, getLineOfLocalPosition(copiedFrom.file, copy.pos), getLineOfLocalPosition(copiedFrom.file, copy.end) + 1);
+                const statementsInSourceFile = copiedFrom.file.statements;
+                const startNodeIndex = findIndex(statementsInSourceFile, s => s.end > copy.pos);
+                if (startNodeIndex === -1) return undefined;
+                const endNodeIndex = findIndex(statementsInSourceFile, s => s.end >= copy.end, startNodeIndex);
```
andrewbranch, line 82: This includes the next statement if the end of the selection includes any leading trivia of the next statement:

```ts
// both statements get added to `statements`:
[|console.log(a);
|]
console.log(b);
```

src/services/pasteEdits.ts
```diff
@@ -77,7 +76,19 @@ function pasteEdits(
...
+                if (startNodeIndex === -1) return undefined;
+                let endNodeIndex = findIndex(statementsInSourceFile, s => s.end >= copy.end, startNodeIndex);
+                /**
+                 * [|console.log(a);
+                 * |]
+                 * console.log(b);
+                 */
+                if (endNodeIndex !== -1 && copy.end <= statementsInSourceFile[endNodeIndex].getStart()) {
+                    endNodeIndex--;
+                }
```
andrewbranch, line 90: I think this could be baked into the `findIndex` predicate, but this is alright.

andrewbranch (approved):

Issue #58687, opened 2024-05-29
Template literal type reduction is inconsistent (regression in 5.5-beta)


ahejlsberg:
### 🔎 Search Terms

Template literal type reduction

### 🕗 Version & Regression Information

This changed in 5.5-beta and is due to #58685.

### ⏯ Playground Link

https://www.typescriptlang.org/play/?ts=5.5.0-beta&ssl=7&ssc=1&pln=1&pc=1#code/MYewdgzgLgBApmArgWxgURgbxgQxgXhgCIciAaGAIwOMqJgF8BuAKBagE8AHOGAZSgAnAJZgA5sIBmHADwAVeAA8oCACYQY0EeIB8NAAYASTHIb7WLADZxYALwCMALhhHMaMzRJEmMX398A9AEwAPIA1lY2MLYATM4C2hLSMmh6hF4+MEHogoIgggCELEA

### 💻 Code

```ts
const enum E { a = "a", b = "b" };

type Stringify<T extends string> = `${T}`;

let z1: `${E}` = "a";        // Ok
let z2: Stringify<E> = "a";  // Error!
```

### 🙁 Actual behavior

Error as indicated.

### 🙂 Expected behavior

No error.

### Additional information about the issue

This is regression from 5.4.5.

ahejlsberg:
Previously, `` `${T}` `` would always strip enum-ness. Following #55371, when `T extends string`, `` `${T}` `` is simplified to just `T`, and enum-ness is no longer stripped. That's technically a breaking change. Furthermore, #55371 is inconsistent in that it _doesn't_ simplify `` `${E}` `` to just `E`, where `E` is an enum type. One way or the other, we should be consistent here. But I'm not really a fan of the change in #55371.

ahejlsberg:
See my comment [here](https://github.com/microsoft/TypeScript/pull/55371#issuecomment-2137812061). My recommendation is that we revert #55371.

PR #58703, opened 2024-05-29
Back out from reducing template literal types with no extra texts too eagerly

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/58687


src/compiler/checker.ts
```diff
@@ -26217,6 +26218,10 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         }
 
         function inferToTemplateLiteralType(source: Type, target: TemplateLiteralType) {
+            if (isTemplateLiteralTypeWithSinglePlaceholderAndNoExtraTexts(source) && isTemplateLiteralTypeWithSinglePlaceholderAndNoExtraTexts(target)) {
```
Andarist, line 26221: I'm still considering if this shouldn't have some extra rules on top of this to be completely safe

tests/baselines/reference/templateLiteralTypes5.errors.txt
```diff
@@ -0,0 +1,50 @@
...
+    // https://github.com/microsoft/TypeScript/issues/55364
+    interface TypeMap {
+      a: "A";
+      b: "B";
+    }
+    declare const f: <T0 extends "a" | "b">(x: `${T0}`) => TypeMap[T0];
+    type F1 = <T1 extends "a" | "b">(x: `${T1}`) => TypeMap[T1];
+    const f1: F1 = f; 
+    type F2 = <T2 extends 'a' | 'b'>(x: `${T2}`) => TypeMap[`${T2}`]
+    const f2: F2 = f
```
Andarist, line 27: before https://github.com/microsoft/TypeScript/pull/55371 this was OK but the assignment above was not. The one above being an error was really spooky as both functions (source and the target there) were syntactically identical.

This line here doesn't come from the user report per se but rather it was created in the process of analyzing the error reported previously for that first assignment (since the error mentioned template literal type at that changed position in the indexed access).

I think it still might be possible to fix this case here, string like index types in the indexed accesses are stringified:
```ts
enum E { a = "a" }
type Foo = { [E.a]: 1 } // type Foo = { a: 1; }
```

So it feels like this fact could be leveraged to simplify such indexed access types like here

src/compiler/checker.ts
```diff
@@ -18087,14 +18087,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
         }
-        if (
-            texts.length === 2 && texts[0] === "" && texts[1] === ""
-            // literals (including string enums) are stringified below
-            && !(types[0].flags & TypeFlags.Literal)
-            // infer T extends StringLike can't be unwrapped eagerly
-            && !types[0].symbol?.declarations?.some(d => d.parent.kind === SyntaxKind.InferType)
-            && isTypeAssignableTo(types[0], stringType)
-        ) {
+        if (texts.length === 2 && texts[0] === "" && texts[1] === "" && (types[0].flags & TypeFlags.StringMapping)) {
```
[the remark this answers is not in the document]
Andarist: no, it just keeps the behavior from https://github.com/microsoft/TypeScript/pull/55371 as this one seems like a safe reduction and i thought that it might be worth avoiding a new type identity in a case like this

src/compiler/checker.ts
ahejlsberg: I'm not understanding why this needs to be here. After reverting #55371 the following works just fine:

```ts
declare function bar<T extends string>(x: `${T}`): T;

function baz<U extends string>(x: `${U}`) {
    return bar(x);  // Infers U

}
```

Andarist: Hmm, I confused myself which one of those errored after reverting #55371:
```ts
interface TypeMap {
    a: 'A'
    b: 'B'
}

declare const f: <T extends 'a' | 'b'>(x: `${T}`) => TypeMap[T];

type F1 = <T extends 'a' | 'b'>(x: `${T}`) => TypeMap[T];
const f1: F1 = f;  // Ok

type F2 = <T extends 'a' | 'b'>(x: `${T}`) => TypeMap[`${T}`];
const f2: F2 = f;  // Error, T is not assignable to `${T}`
```

Then I just knew where I could try to improve inference for this as I knew that in here `` `${T}` `` was being inferred for `T` based on `` `${T}` `` source and `` `${T}` `` target as that originally created an issue further down the road.

i agree, this PR is completely redundant 🤦‍♂️ this is what I get for jumping too quickly between work, this, and the kids

PR #58705, opened 2024-05-29
Fix baselines after PR 58621

jakebailey:
This PR was not up to date with main and had not been recently run.


merceyz:
Perhaps the merge queue feature could be used to avoid this problem?

jakebailey:
Our tests take a while to run, so it probably will slow things down a bit while removing the ability to figure out which commit was the one that actually failed (instead of just the merge group), so I'm not sure if it's all good. But technically yes.

PR #58706, opened 2024-05-29
Clone node to remove location even when it has been modified if needed

### LINKED ISSUE #58698: TS5.5-beta injects a random `@ts-expect-error` comment in the generated `.d.ts` files
### 🔎 Search Terms

comment 5.5 invalid injected inlined

- This changed in version 5.5-beta

### ⏯ Playground Link

_No response_

### 💻 Code

1. Clone https://github.com/nicolo-ribaudo/babel/tree/ts-5.5-random-comment
2. Run `yarn` to install all deps
3. Run `make generate-tsconfig` to generate all the `tsconfig.json` files
4. Run `yarn tsc -b ./packages/babel-helper-plugin-utils/tsconfig.json`

If you ever need to reset the TS state (e.g. delete all `.tsbuildinfo` and `.d.ts` files), run `make clean-dts`.

### 🙁 Actual behavior

`dts/packages/babel-parser/src/plugins/typescript/index.d.ts`, line 540 contains a random `@ts-expect-error` comment:
```ts
        parseExprListItem(this: Parser, allowEmpty?: boolean, refExpressionErrors?: ExpressionErrors | null, allowPlaceholder?: boolean | // @ts-expect-error todo(flow->ts) 0121
        null): N.Expression | null;
```

That `parseExprListItem` is inlined there from another file, that doesn't have that comment:
https://github.com/nicolo-ribaudo/babel/blob/e5726bde94899149d60706698a0f0a968208bf60/packages/babel-parser/src/parser/expression.ts#L2711-L2728

Instead, that comment is coming from
https://github.com/nicolo-ribaudo/babel/blob/e5726bde94899149d60706698a0f0a968208bf60/packages/babel-parser/src/plugins/typescript/index.ts#L2637-L2642

### 🙂 Expected behavior

The comment shouldn't be there

weswigham:
Because when a node is updated because its' children changed, location data is, by default, copied by the node update functions. In addition, I also handle deleting location data from node _arrays_ from other files as well, as I noticed that while debugging.
 
Additionally, one of the comment range copies in the test was coming from an unchecked `setCommentRange` outside the node copying code, so that now has builtin checks now, too.

Fixes #58698


jakebailey:
Wow, this is a big baseline change, but maybe it's just going to have to happen given #58698 is a bad bug?

jakebailey:
Would run perf on this but I think the merge conflict will make the bot mad.

jakebailey:

weswigham:
@jakebailey I've minimized the diff (and swapped them all to only improvements) by adding a small change to explicitly add positions for type parameter names. I think the original change caused us to stop attaching positions to them spuriously, and now we actually do it explicitly, so there's a smaller diff, but there *is* a diff of parameter names where we now actually have locations. Also a fourslash test update that shows a new comment getting preserved.

tests/cases/compiler/declarationEmitTopLevelNodeFromCrossFile.ts
```diff
@@ -0,0 +1,16 @@
...
+import {fn} from "./a";
+export const m = {                        
+    /**
+    * leading doc for prop
+    */ 
+    prop: 1
+}
+
+
+export const x = { p: fn };
```
jakebailey, line 16: This test doesn't seem like it tests #58698; am I missing something?

jakebailey: Hm, maybe I'm just wrong given the issue was about pulling comments from other files, yeah.

jakebailey:

jakebailey (approved):
LGTM assuming perf remains unchanged.

Issue #58697, opened 2024-05-29
Spread of two generic array types is misreported as having length 2


dragomirtitian:
### 🔎 Search Terms

tuple spread any generic

### 🕗 Version & Regression Information

- This is the behavior in every version I tried, and I reviewed the FAQ

### ⏯ Playground Link

[Playground Link](https://www.typescriptlang.org/play/?ts=4.5.5#code/GYVwdgxgLglg9mABAKzjMAeAKogpgDylzABMBnRAQzAE8BtAXQBpEAFPQ48q2xgPgAUASgBciOgDopWFlImsGiAN4AoROsQAbXFET4x1eooC84hgG41GgE46Q1pPnOIA9C8RZK1gOY7EtgEcQGFsKACY8bQBbYigBMiFEACMQXTI4ewhcRCjKGkQAC0oAN2zgXAB3XGtEQBlyQHVyQGVyFQBfFRVQSFgEFDQwMOwOIlIKQ0YWdgJh7jGGQVFxORlEOQVlK3VtXX1xQxZDE0QwEE1NAEIN-zsHPWc3RABlDOssxAAHazhimBJcCjA4DlKFAIAVEMA4DVil4YJQSDAIJFcDEwLpge84GQYN0kAAGRDoRBQLy+XSAUHJWiogA)

### 💻 Code

```ts
function join<T extends any[], P extends any[]>(): [...T, ...P] {
    let x: any[] = [];
    return x; // Target requires 2 element(s) but source may have fewer ❌❗❓
}

function join2<T extends any[], P extends any[]>(): [...T, ...P] {
    let x: [any, any] = null!
    return x; // Source provides no match for variadic element at position 0 in target ✅
}

```

### 🙁 Actual behavior

`join` errors on return with `Target requires 2 element`. There is nothing in the target type that suggests it requires two elements.

### 🙂 Expected behavior

Should error with a message similar to `join2`: 
> Source provides no match for variadic element at position 0 in target

PR #58708, opened 2024-05-30
Fixed reported errors for variadic element mismatches

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/58697


jakebailey (approved):
This seems correct to me? @weswigham

src/compiler/checker.ts
```diff
@@ -23411,7 +23411,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     const sourceArity = getTypeReferenceArity(source);
                     const targetArity = getTypeReferenceArity(target);
                     const sourceRestFlag = isTupleType(source) ? source.target.combinedFlags & ElementFlags.Rest : ElementFlags.Rest;
-                    const targetRestFlag = target.target.combinedFlags & ElementFlags.Rest;
+                    const targetHasRestElement = target.target.hasRestElement;
```
weswigham, line 23414: `type.hasRestElement = !!(combinedFlags & ElementFlags.Variable);`, and `ElementFlags.Variable` is `ElementFlags.Rest | ElementFlags.Variadic` so... yeah, I guess this works to expand the check from just rest to rest and variadic elements! I read this like 3 times before checking how `hasRestElement` was defined and realizing that, though.

We should slap a `@deprecated` on `hasRestElement`, really.  The element flags are really the new way to check that.

weswigham (approved):

weswigham:
Looks like this one needs a small resync with `main` before it can be merged, though~

Andarist:
I synced this with main, it should now be ready to merge :)

src/compiler/checker.ts
Andarist: > We should slap a @deprecated on hasRestElement, really. The element flags are really the new way to check that.

https://github.com/microsoft/TypeScript/pull/58864 😉

PR #58711, opened 2024-05-30
feat: make Object.keys type strict

martinmunillas:
Makes Object.keys type strict, while still allowing devs to opt out if they want to.

Currently the following code
```typescript
interface User {
    email: string
    name: string
}

const martin: User = {name: martin, email: martin@gmail.com}

const keys = Object.keys(martin)
```
would end up in keys being of type `string[]` instead of `('email' | 'name')[]`, which usually leads to unnecessary type castings afterward.

This solution sill allows people to keep the previous behaviour by explicitly setting the types
```typescript
const keys = Object.keys<User, string>(martin) // string[]
```


martinmunillas:
just realised of tests/cases/compiler/noObjectKeysToKeyofT.ts

PR #58712, opened 2024-05-30
(type | interface) as part of an assertion should not trigger an parsing error.

### LINKED ISSUE #58248: `satisfies` does not work on a variable called `type`
### 🔎 Search Terms

satisfies, type, ts1005

This happens from 3.3.3333 to nightly.

`satisfies` was introduced in 4.9.

### ⏯ Playground Link

https://www.typescriptlang.org/play?#code/MYewdgzgLgBFCeAHApjAvDA5AD0wKDwRRggEMoBLCAMwuQhKgCcKwBzPIA

### 💻 Code

```ts
const type = 'x'

type satisfies string
```

### 🙁 Actual behavior

Gives an `'=' expected.(1005)` error.

### 🙂 Expected behavior

For it to work like with any other variable of any other name, specially since it's a common name.

This also happens with other TypeScript keywords like `interface`.

jean-michelet:
Will fixes #58248 

I'm not sure if there are any other cases I've missed. I'll take a closer look this week, but feel free to give feedbacks :pray: 

This should be ok:
```ts
const type = 'x';

type satisfies string;
type as string;

const interface = 'x';

interface satisfies string;
interface as string;
```

### DIFF
--- a/src/compiler/parser.ts
+++ b/src/compiler/parser.ts
@@ -2320,7 +2320,6 @@ namespace Parser {
         if (token() === SyntaxKind.AwaitKeyword && inAwaitContext()) {
             return false;
         }
-
         return token() > SyntaxKind.LastReservedWord;
     }
 
@@ -5140,7 +5139,24 @@ namespace Parser {
 
     function nextTokenIsIdentifierOnSameLine() {
         nextToken();
-        return !scanner.hasPrecedingLineBreak() && isIdentifier();
+        if (scanner.hasPrecedingLineBreak()) {
+            return false
+        }
+
+        switch(token()) {
+            case SyntaxKind.AsKeyword:
+            case SyntaxKind.SatisfiesKeyword:{
+                let truthy = false
+                lookAhead(() => {
+                    nextToken()
+                    truthy = token() === SyntaxKind.EqualsToken || token() === SyntaxKind.OpenBraceToken
+                })
+                return truthy
+            }
+            
+            default:
+                return isIdentifier();
+        }
     }
 
     function parseYieldExpression(): YieldExpression {
--- a/src/compiler/parser.ts
+++ b/src/compiler/parser.ts
@@ -2320,6 +2320,7 @@ namespace Parser {
         if (token() === SyntaxKind.AwaitKeyword && inAwaitContext()) {
             return false;
         }
+        
         return token() > SyntaxKind.LastReservedWord;
     }
 
--- a/tests/baselines/reference/typeOrInterfaceIdentifierSatisfiesExpression.js
+++ b/tests/baselines/reference/typeOrInterfaceIdentifierSatisfiesExpression.js
@@ -1,0 +1,20 @@
+//// [tests/cases/compiler/typeOrInterfaceIdentifierSatisfiesExpression.ts] ////
+
+//// [typeOrInterfaceIdentifierSatisfiesExpression.ts]
+const type = 'x';
+
+type satisfies string;
+type as string;
+
+const interface = 'x';
+
+interface satisfies string;
+interface as string;
+
+//// [typeOrInterfaceIdentifierSatisfiesExpression.js]
+var type = 'x';
+type;
+type;
+var interface = 'x';
+interface;
+interface;
--- a/tests/baselines/reference/typeOrInterfaceIdentifierSatisfiesExpression.symbols
+++ b/tests/baselines/reference/typeOrInterfaceIdentifierSatisfiesExpression.symbols
@@ -1,0 +1,21 @@
+//// [tests/cases/compiler/typeOrInterfaceIdentifierSatisfiesExpression.ts] ////
+
+=== typeOrInterfaceIdentifierSatisfiesExpression.ts ===
+const type = 'x';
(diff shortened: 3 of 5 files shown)


src/compiler/parser.ts
```diff
@@ -5139,8 +5139,26 @@ namespace Parser {
...
+        if (scanner.hasPrecedingLineBreak()) {
+            return false;
+        }
+
+        if (!isTypeKeyword) {
+            return isIdentifier();
+        }
+
+        switch (token()) {
+            case SyntaxKind.AsKeyword:
```
DanielRosenwasser, line 5154: It really feels like this function should have been left as-is, and the look-ahead should have been done after calling this function specifically for the `type` keyword.

~~It also might be the case that you don't actually need to special-case `as` and `satisfies`, you can just do an unconditional look-ahead. But I might have to think about that.~~

src/compiler/parser.ts
```diff
@@ -5139,8 +5139,26 @@ namespace Parser {
...
+            case SyntaxKind.SatisfiesKeyword: {
+                return lookAhead(() => nextToken() === SyntaxKind.EqualsToken);
```
DanielRosenwasser, line 5156: ~~The name of a type alias isn't always immediately followed by an `=`, so I don't think this specific look-ahead is enough.~~

See the comment I'm leaving in my review.

```ts
type satisfies<T> = T;
```

DanielRosenwasser (changes requested):
The more I think about it, I realized this is a bit of a tricky problem because the look-ahead is unbounded. You can write

```ts
type satisfies <T>() => T
```

And from the parser's point of view, when it sees

```ts
type satisfies <
```

it needs to make a call of whether this is going to be a type alias or a `satisfies` expression.

So what I would do is keep the special-casing for `as` and `satisfies` and make sure we have tests for things like

```ts
// These were written with the expectation of parse errors.
// We expect the parser to try constructing a statement expression
// with 'satisfies' and 'as' expressions.

type satisfies<T> = T;
type as<T> = T;
```

and

```ts
// These should be parsed as statements
// with 'satisfies' and 'as' expressions.

type satisfies <T>() => T;
type as <T>() => T;
```

DanielRosenwasser:
Also, like @whzx5byb mentioned, you should add an `interface satisfies`/`interface as` test to ensure there's no regressions

jean-michelet (commented):
I hope I understand what you have in mind.
I've left comments to make it clear what I understand needs to be done.

src/compiler/parser.ts
```diff
@@ -7168,8 +7169,19 @@ namespace Parser {
                 //
                 // could be legal, it would add complexity for very little gain.
                 case SyntaxKind.InterfaceKeyword:
-                case SyntaxKind.TypeKeyword:
                     return nextTokenIsIdentifierOnSameLine();
+                case SyntaxKind.TypeKeyword:
+                    const isIdentifierOnSameLine = nextTokenIsIdentifierOnSameLine();
+                    if (isIdentifierOnSameLine && token() === SyntaxKind.AsKeyword || token() === SyntaxKind.SatisfiesKeyword) {
```
jean-michelet, line 7175: Do you think this part should be in a proper function?
We could return something like this:
```ts
return nextTokenIsIdentifierOnSameLine() && someNameToExplainsWhatIsDoneRegardingSatisfiesAndAs()
```

src/compiler/parser.ts
```diff
@@ -7168,8 +7169,19 @@ namespace Parser {
...
                 // could be legal, it would add complexity for very little gain.
                 case SyntaxKind.InterfaceKeyword:
-                case SyntaxKind.TypeKeyword:
                     return nextTokenIsIdentifierOnSameLine();
+                case SyntaxKind.TypeKeyword:
+                    const isIdentifierOnSameLine = nextTokenIsIdentifierOnSameLine();
+                    if (isIdentifierOnSameLine && token() === SyntaxKind.AsKeyword || token() === SyntaxKind.SatisfiesKeyword) {
+                        return lookAhead(() => {
+                            nextToken();
+                            parseTypeParameters();
```
jakebailey, line 7178: This feels like it would be pretty wasteful, but I suppose that people can't write this code today anyway and it'd only be paid for `as` and `satisfies`.

tests/baselines/reference/interfaceIdentifierAssertionExpressionWithParseError.errors.txt
```diff
@@ -0,0 +1,21 @@
...
+    interface satisfies string;
+                        ~~~~~~
+!!! error TS1005: '{' expected.
+                        ~~~~~~
+!!! error TS2693: 'string' only refers to a type, but is being used as a value here.
+    interface as string;
+                 ~~~~~~
+!!! error TS1005: '{' expected.
+                 ~~~~~~
+!!! error TS2693: 'string' only refers to a type, but is being used as a value here.
```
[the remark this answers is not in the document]
jakebailey: Can you explicitly test `interface satisfies { }`?

jean-michelet:
Is there any interest for this PR?

PR #58715, opened 2024-05-30
Do not expand type references in keyof and index access

### LINKED ISSUE #58713: [ts-5.5] Private aliases expanded in index and keyof types - 5.5 regression
### 🔎 Search Terms

type alias inlined regression 5.5

- This changed between versions 5.5.0-dev.20240514 and 5.5.0-dev.20240515

### ⏯ Playground Link

[Workbench](https://www.typescriptlang.org/dev/bug-workbench/?ts=5.5.0-dev.20240515#code/PTAEAEBMFMGMBsCGAnRAXAlgewHYC5Q1kBXaAKDJAgDMN5odEBbaAxAOjQGcy0BPAA7RQAeVABeUAG8yoOaAHIsAglyIYcAc1nzFygEyr1WsgF8K-IaACSE6Trl6VoNcg3bzZaAA8BWZGigsLhqoNQ4dgAUAG4EIgDaAOROiQC6ADSgWHFMGGgAPCKZANbQfFjUNgB8AJQSVdKmANwUVOC09IwsBABGnDxkGEx+AdLhpmFKTKCJ7MCIiV6+-oHBOKGVkuFNQA)

### 💻 Code

```ts
// @declaration: true

// @filename: a.ts
type O = {
    prop: string
    prop2: string
}

type I = {
    prop: string
}

export const fn = (v: O['prop'], o: Omit<O, keyof I>) => {};

// @filename: b.ts

import {fn} from './a'
export const f = fn;
```

### 🙁 Actual behavior

`O['prop']` is written in declaration file as `{ prop: string; prop2: string; }[prop"]`. And `Omit<O, keyof I>` is printed as `Omit<{
    prop: string;
    prop2: string;
}, keyof {
    prop: string;
}>` 

This is problematic especially if the type is a bigger as it balloon the size of the printed type and it may pull in references that were not previously used (causing new errors)

### 🙂 Expected behavior

`O['prop']` is written in declaration file as `string` as `O` and `Omit<O, keyof I>` is printed as `Omit<O, 'prop'>` as in 5.4  since `O` and `I` can't be named

_No response_

dragomirtitian:
Fixes #58713


tests/baselines/reference/declarationEmitAliasInlineing.js
```diff
@@ -0,0 +1,85 @@
...
+export type O = {
+    prop: string;
+    prop2: string;
+};
+export type I = {
+    prop: string;
+};
+export declare const fnExp: (v: O["prop"], p: Omit<O, "prop">, key: keyof O, p2: Omit<O, keyof I>) => void;
+//// [b.d.ts]
+export declare const f: (v: string, p: Omit<{
```
rbuckton, line 75: I'm a bit surprised we don't emit this as
```ts
export declare const f: typeof import("./a").fn;
```
since we're willing to emit the `import("./aExp").O["prop"]` et al.

src/compiler/checker.ts
```diff
@@ -8590,27 +8612,17 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+                    const objectType = tryVisitTypeReference(node.objectType);
+                    if (!objectType) {
+                        return serializeExistingTypeNode(context, node);
                     }
-                    return serializeExistingTypeNode(context, node);
+                    return factory.updateIndexedAccessTypeNode(node, objectType, visitNode(node.indexType, visitExistingNodeTreeSymbols, isTypeNode)!);
+                }
+
+                if (isTypeReferenceNode(node)) {
+                    return tryVisitTypeReference(node) ?? serializeExistingTypeNode(context, node);
```
weswigham, line 8625: Are the explicit fallbacks needed with the automatic fallback added in https://github.com/microsoft/TypeScript/pull/58720?

tests/baselines/reference/declarationEmitAliasInlineing.js
dragomirtitian: `typeof` is currently only used for function declarations. I think leaning into `typeof` for variables is an options that should be considered. I'm not sure why only function declarations use `typeof` (although I haven't really though too deeply about it)

src/compiler/checker.ts
dragomirtitian: Yeah. You are right. Better to lean on the new fallback mechanism. As it currently is we could end up trying to print the type two times (when we fail). Pushed the change

tests/cases/fourslash/quickInfoForContextuallyTypedParameters.ts
```diff
@@ -34,7 +34,7 @@ verify.quickInfos({
     func?: Function;
 }, obj: T, key: "name"): "name"`,
     4: `function foof<Error, "name">(settings: (row: Error) => {
-    value: Error["name"];
+    value: string;
```
weswigham, line 37: Weird - why's this one change? Isn't `Error` in scope?

dragomirtitian: It's not about `Error`. Before when we saw a type parameter that needed mapping, I would fall back to type serialiation. Then type serialization would map `T` to `Error` and then print `Error`. Now when I see the `T` in `T["name"]` and that it can't be reused without falling back to serialization, I fall back to serializaing `T["name"]` insead which gets us `string`.

weswigham (approved):

PR #58719, opened 2024-05-30
Added NoTruncation flag to completions

armanio123:
Fixes https://github.com/microsoft/TypeScript/issues/58666#issuecomment-2132408261

The crash is generated due to completions not being able to format a node. The node happens to be synthetized node created with text `... 1 more ....` I believe formatting is not working because is invalid syntax.

### DIFF
--- a/src/services/codefixes/helpers.ts
+++ b/src/services/codefixes/helpers.ts
@@ -219,7 +219,8 @@ export function addNewNodeForMemberSymbol(
     switch (kind) {
         case SyntaxKind.PropertySignature:
         case SyntaxKind.PropertyDeclaration:
-            const flags = quotePreference === QuotePreference.Single ? NodeBuilderFlags.UseSingleQuotesForStringLiteralType : undefined;
+            let flags = NodeBuilderFlags.NoTruncation;
+            flags |= quotePreference === QuotePreference.Single ? NodeBuilderFlags.UseSingleQuotesForStringLiteralType : 0;
             let typeNode = checker.typeToTypeNode(type, enclosingDeclaration, flags, getNoopSymbolTrackerWithResolver(context));
             if (importAdder) {
                 const importableReference = tryGetAutoImportableReferenceFromTypeNode(typeNode, scriptTarget);


src/services/codefixes/helpers.ts
```diff
@@ -219,7 +219,8 @@ export function addNewNodeForMemberSymbol(
     switch (kind) {
         case SyntaxKind.PropertySignature:
         case SyntaxKind.PropertyDeclaration:
-            const flags = quotePreference === QuotePreference.Single ? NodeBuilderFlags.UseSingleQuotesForStringLiteralType : undefined;
+            let flags = NodeBuilderFlags.NoTruncation;
+            flags |= quotePreference === QuotePreference.Single ? NodeBuilderFlags.UseSingleQuotesForStringLiteralType : 0;
             let typeNode = checker.typeToTypeNode(type, enclosingDeclaration, flags, getNoopSymbolTrackerWithResolver(context));
```
jakebailey, line 224: Searching for `checker.typeToTypeNode`, it looks like there are a few other calls that are missing that flag. Some shouldn't have it, but I see quite a few that definitely should. May be good to audit all instances of that call in `services`.

jakebailey (approved):
Of course auditing the other ones shouldn't prevent this one from going in. Would be nice to have a test, though, but that may be hard.

DanielRosenwasser:

gabritto:
> While you might want to set this, even with `NoTruncation`, you'll still get invalid syntax back if the result is greater than around 1,000,000 characters (instead of 160) and forced truncation happens anyway. Just less often. It may still be worth considering how to handle it.
> 
> It's also somewhat questionable if you wanna complete something that's >160 characters, too.

Somewhat over 160 characters is still way more reasonable than 1 million. I think at a million characters it's very likely either generated code or maybe something went wrong in the compiler.

DanielRosenwasser:
Yeah, pragmatically this is the right fix. Long-term, I want to get rid of all the garbage crashes from TSServer reports. And what that means over time is probably making the parser and/or formatter more aware of these synthetic elision constructs.

armanio123:
> Of course auditing the other ones shouldn't prevent this one from going in. Would be nice to have a test, though, but that may be hard.

I'll look into the other services and send a PR if I find anything. About the test, I've been looking into a code that repros but I have not found anything. I'll merge as is and spent a little more time trying to get one.

PR #58722, opened 2024-05-30
Improve the performance of isolatedDeclarations quickfix 

### LINKED ISSUE #58426: Issues with `isolatedDeclarations` and the associated fixes in editor
### 🔎 Search Terms

isolatedDeclarations, quickfix, lag, editor, VSCode, 5.5

- This is specific to new 5.5 features

### ⏯ Playground Link

_No response_

### 💻 Code

_

### 🙁 Actual behavior

I've been testing out `isolatedDeclarations` in VSCode Insiders on WSL and wanted to collect a few of the issues I noticed in general, especially with some of the associated quick fixes as discussed with @jakebailey in the TS Discord.

1. The most egregious issue I encountered was a huge lag spike when calculating quick fixes for isolatedDeclarations, even for simple types.

Here you can see I can hover this.to?.traverseApply which is trivially inferred as a simple type from the base class. But then after I hover outValidator and it has to generate a quick fix, it's like it gets stuck recalculating it:

https://github.com/microsoft/TypeScript/assets/10645823/88f3dbed-a99f-459e-92d7-fe47dea0f3fe

2. I also notice when I'm annotating a value with an isolated declarations error, sometimes auto import suggestions can take several seconds to pop up, so it seems like a general performance issue with some of these fixes.

More broadly, I also notice the total check time for [ArkType's repo](https://github.com/arktypeio/arktype) has increased from 7.5 seconds in 5.4 to 9.5 seconds in 5.5, so there may be some underlying change to e.g. calculate additional predicates, but these fixes associated with isolatedDelcarations seem to be particularly problematic.

Another pattern I noticed was lots of inline imports when applying a quick-fix which is rarely desirable. It would be great if by default (e.g. if there are no naming conflicts), the import would be added to the top of the file and the type would be referenced as normal:

https://github.com/microsoft/TypeScript/assets/10645823/6c5c1c43-25e9-421e-868d-7abad2109db6

3. I use the pattern `type SomeType<parameter = SomeDefault>` frequently, and I notice whenever a quickfix is applied it adds the parameter even if it's the default.

This wouldn't be a huge deal, but since if `SomeDefault` is a type, it will also be expanded out to its structural form and potentially inline-imported, it can add a ton of visual clutter (sometimes an order of magnitude more than the type itself).

4. I don't understand well most of the underlying rules but I think my biggest wish for `isolatedDeclarations` is more constants like those strings initialized as class props or stuff like this could somehow not require duplicating the entire object structure at a type-level to replicate what `typeof someConstant` could do in the past:

image

If that's fundamentally in conflict with the goals of `--isolatedDeclarations`, feel free to disregard, it just feels frustrating to have no way to derive the type and value from a single source without repeating them anymore even for simple cases.

Sorry for the haphazard format of this issue. I'm very excited about this feature as some of these kinks are ironed out I think the fixes will also add a lot of value, I just wanted to make sure there was a record of these things, particularly since Jake requested I submit something.  

### 🙂 Expected behavior

_

_No response_

blickly:
It turns out that `typeChecker.getEmitResolver()` can be expensive, so avoid calling it unless necessary.

Hopefully helps #58426, in particularly the performance issue (bullet 1)

### DIFF
--- a/src/services/codefixes/fixMissingTypeAnnotationOnExports.ts
+++ b/src/services/codefixes/fixMissingTypeAnnotationOnExports.ts
@@ -183,7 +183,7 @@ registerCodeFix({
 
         addCodeAction(addInlineTypeAssertion, fixes, context, TypePrintMode.Full, f => f.addInlineAssertion(context.span));
         addCodeAction(addInlineTypeAssertion, fixes, context, TypePrintMode.Relative, f => f.addInlineAssertion(context.span));
-        addCodeAction(addAnnotationFix, fixes, context, TypePrintMode.Widened, f => f.addInlineAssertion(context.span));
+        addCodeAction(addInlineTypeAssertion, fixes, context, TypePrintMode.Widened, f => f.addInlineAssertion(context.span));
 
         addCodeAction(extractExpression, fixes, context, TypePrintMode.Full, f => f.extractAsVariable(context.span));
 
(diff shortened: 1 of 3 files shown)


blickly:
@microsoft-github-policy-service agree [company="Google"]

jakebailey:
Weird, how does this end up being expensive? Is it recalculating diagnostics? Something else? Do you have a profile before/after?

You'll want to resend that CLA comment again, I don't think it understands edits. I also modified your PR description as the language it contained would have closed the linked issue (when it shouldn't in full yet).

blickly:
Yeah, I assume it's recalculating diagnostics, as the method called inside `getEmitResolver` is called `getDiagnostics2`.

The full before profile is [vscode-profile-2024-05-29-16-18-13.cpuprofile](https://github.com/user-attachments/files/15506565/vscode-profile-2024-05-29-16-18-13.cpuprofile)

After this fix, the quickfix in my example projects didn't have perceptible latency, so I didn't try to grab an after profile.

@microsoft-github-policy-service agree company="Google"

src/services/codefixes/fixMissingTypeAnnotationOnExports.ts
```diff
@@ -887,7 +886,7 @@ function withContext<T>(
             type = widenedType;
         }
 
-        if (isParameter(node) && emitResolver.requiresAddingImplicitUndefined(node)) {
+        if (isParameter(node) && typeChecker.getEmitResolver().requiresAddingImplicitUndefined(node)) {
```
jakebailey, line 889: It turns out the emit resolver is needed _only_ for `requiresAddingImplicitUndefined`. Can we just export this from the checker directly and avoid the resolver entirely? That should save a bunch more time.

blickly:
Oops, I think I uploaded the wrong before profile. Darn reverse chronological.  The useful ones are these three:

* [vscode-profile-2024-05-30-13-47-48.cpuprofile](https://github.com/user-attachments/files/15506727/vscode-profile-2024-05-30-13-47-48.cpuprofile)
* [vscode-profile-2024-05-30-13-41-49.cpuprofile](https://github.com/user-attachments/files/15506728/vscode-profile-2024-05-30-13-41-49.cpuprofile)
* [vscode-profile-2024-05-30-13-31-47.cpuprofile](https://github.com/user-attachments/files/15506729/vscode-profile-2024-05-30-13-31-47.cpuprofile)

src/services/codefixes/fixMissingTypeAnnotationOnExports.ts
blickly: Right, that seems like a clear improvement.  Done.

jakebailey (approved):

robpalme:
Thank you @blickly for jumping on this 👍

PR #58728, opened 2024-05-31
[WIP] Scanner Refactoring & Optimization

graphemecluster:
~This involves only the scanner, and includes refactoring like preferring `charCodeChecked` and `codePointChecked` over bound checks plus `charCodeUnchecked` and `codePointUnchecked`, avoiding the use of string literals and string index property accessing, and preferring `String#substring` over `String#slice` due to performance.~

_Part of the above is already done in #58362._

Originally I would like to eliminate all regexes inside the scanner, but they don’t seem to affect performance at all, so I ended up abandoning the related changes.

This PR shouldn’t affect anything other than performance.

### DIFF
--- a/src/compiler/scanner.ts
+++ b/src/compiler/scanner.ts
@@ -704,10 +704,13 @@ export function skipTrivia(text: string, pos: number, stopAfterLineBreak?: boole
                 break;
 
             case CharacterCodes.hash:
-                if (pos === 0 && isShebangTrivia(text, pos)) {
-                    pos = scanShebangTrivia(text, pos);
-                    canConsumeStar = false;
-                    continue;
+                if (pos === 0) {
+                    const shebang = getShebang(text);
+                    if (shebang) {
+                        pos += shebang.length;
+                        canConsumeStar = false;
+                        continue;
+                    }
                 }
                 break;
 
@@ -720,7 +723,7 @@ export function skipTrivia(text: string, pos: number, stopAfterLineBreak?: boole
                 break;
 
             default:
-                if (ch > CharacterCodes.maxAsciiCharacter && (isWhiteSpaceLike(ch))) {
+                if (ch > CharacterCodes.maxAsciiCharacter && isWhiteSpaceLike(ch)) {
                     pos++;
                     continue;
                 }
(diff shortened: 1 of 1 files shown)


src/compiler/scanner.ts
```diff
@@ -1127,7 +1116,7 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
      * `pos` is outside the bounds set for `text`, `CharacterCodes.EOF` is returned instead.
      */
     function codePointChecked(pos: number) {
-        return pos >= 0 && pos < end ? codePointUnchecked(pos) : CharacterCodes.EOF;
+        return pos < end ? codePointUnchecked(pos) : CharacterCodes.EOF;
```
graphemecluster, line 1119: I carefully examined all expressions involving `pos--` and `pos -= number` and determined that the `pos >= 0` check is unnecessary.

jakebailey:
Is this the same as #58362?

src/compiler/scanner.ts
```diff
@@ -2559,7 +2550,7 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
                 let regExpFlags = RegularExpressionFlags.None;
                 while (true) {
                     const ch = codePointChecked(pos);
-                    if (ch === CharacterCodes.EOF || !isIdentifierPart(ch, languageVersion)) {
+                    if (!isIdentifierPart(ch, languageVersion)) {
```
graphemecluster, line 2553: This check is unnecessary since `isIdentifierPart(CharacterCodes.EOF, languageVersion)` must return false (because `CharacterCodes.EOF <= CharacterCodes.maxAsciiCharacter`).

src/compiler/scanner.ts
```diff
@@ -2997,7 +2988,7 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
         }
 
         function isClassContentExit(ch: number) {
-            return ch === CharacterCodes.closeBracket || ch === CharacterCodes.EOF || pos >= end;
+            return ch === CharacterCodes.closeBracket || ch === CharacterCodes.EOF;
```
graphemecluster, line 2991: This is unnecessary as all `isClassContentExit` calls are directly after `charCodeChecked`.

graphemecluster:
> Is this the same as #58362?

Ah, I overlooked this. Let me omit all bound-checking related changes and leave the rest then.

graphemecluster:
I guess I need to get into the habit of searching for relevant PR, not just issues, in a search engine.

graphemecluster:
I guess I need to get into the habit of searching for relevant PRs, not just issues, in a search engine, before open up one.

graphemecluster:
Fine, let me close this first since the branch cannot be renamed.

PR #58729, opened 2024-05-31
Analyze control flow effects of lambdas passed as arguments

### LINKED ISSUE #11498: Annotate immediately-invoked functions for inlining flow control analysis
Per #9998 and its many offshoots, we have problems analyzing code like this

```ts
let x: string | number = "OK";
mystery(() => {
  x = 10;
});
if (x === 10) { // Error, we believe this to be impossible  
}
```

The problem is that we don't know if `mystery` invokes its argument "now" or "later".

Proposal is to allow some form of indicating that the function is invoked immediately, e.g.
```ts
declare function mystery(arg: immediately () => void): void;
```
This would "inline" the function body for the purposes of flow control analysis

### LINKED ISSUE #15380: Variable type incorrectly narrowed to null when a callback is passed to a function
**TypeScript Version:**  2.2.2
Compile with StrictNullChecks enabled.

**Code**

```ts
class Foo {
    public bar: string = "";
}

function test() {
    let foo: Foo | null = null;

    [1].forEach((item) => {
        foo = new Foo();
    });

    // Here Typescript compiler incorrectly thinks that 'foo' has null type 
    // although it will always be an instance of Foo class.
    if (foo) {
        // The next line does not compile because 'foo' variable has never type here.
        foo.bar;
    }
}
```

**Expected behavior:**
Should compile. 
**Actual behavior:**
Does not compile. The compiler incorrectly narrows the type of foo variable to null type as it seems to ignore the callback given to the forEach function call, which actually will always be called.

Notes:
The code compiles without StrictNullChecks just fine.
The below code works as expected, and I believe the same would be the expected behavior for the example code as well:
```ts
class Foo {
    public bar: string = "";
}

function test() {
    let foo: Foo | null = null;

    // When calling a simple lambda explicitly this works as expected
    (() => {
        foo = new Foo();
    })();
    if (foo) {
        foo.bar;
    }
}
```

### LINKED ISSUE #57880: Type narrowed too much
### 🔎 Search Terms

type narrowing local variable

I don't see anything related in the FAQ. But I acknowledge that this seem too common to be unknown. Every version seem affected however.

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.4.2#code/MYewdgzgLgBMCGAbRMC8MAUAzAXJglGgHwwBuIAlgCaGolYb4DcAUC6JLFAKbRoHEYAbxYwxMRN1jw8AVzBVuWCmG5UYAHxhhZAWwBG3AE795i5aqqtxMAPS2YAFQCeAB17AjFV7AjduuhAwUCAwhnAguq6SPIjOMPLwAO7wRtzBABbwXBnppEiy3KLi9jAgWDAABvCVcPBgcFlgAObpKpnpWCDIIEkqzXXIAHTFYgjIGIyCQjDw-ACMTDAAvsyjdg4AggodaQA0MBS+UBTImSoA1pnZVTV1DeBx66WuIBAQFPpxYemVZkoqNSVEY2CgVDBzACEqHQ-wsakIIhsNlKUAyFCCiEBMFasDSACtuMAeFR1jZ4EMQgBlKBeFqMaziZYsZasIA

### 💻 Code

```ts
const call = (f: () => void) => f();

const test = () => {
    let a: undefined | number = undefined;
    // Typescript seems to be completely unaware that the value
    // of `a` can change in the following call.
    call(() => { a = 1; });
    // And there, it still think that `a` can only
    // possibly be `undefined`.
    if (a !== undefined) {
        // this line get rejected
        a.toString();
    }
};
```

### 🙁 Actual behavior

It seems that Typescript is not considering a function call to affect local variable. It proceed to narrow the type from this limitation, and thus reject valid code.

Not assigning `undefined` to `a` causes the error to disappear *BUT* this is still a problem, because I might want a default value here (could be a number initially, that later get changed to undefined for example).

For example, and this is worse:

```ts
const call = (f: () => void) => f();

const test = () => {
    let a: undefined | number = 42;
    call(() => {
        a = undefined;
    });
    a.toString();
};
```
This will compile fine, but it will fail at runtime! 

### 🙂 Expected behavior

The type should not be narrowed in this case.

Typescript should detect that a closure is referencing a local variable, and thus, should avoid narrowing its type.

_No response_

ahejlsberg:
This PR introduces a new `immediate` modifier that can be applied to callback parameters to indicate that control flow analysis should assume that function expressions or arrow functions (in the following called lambda expressions) passed as arguments in a function call _may_ be invoked synchronously during that call. For example:

```ts
declare function mystery(immediate cb: () => void): void;

let x: string | number = "OK";
x;  // string
mystery(() => {
    x = 10;
});
x;  // string | number
if (x === 10) {  // Error without 'immediate' modifier
}
```

Without the `immediate` modifier, control flow analysis will assume that the lambda expression is never invoked synchronously and thus will consider the type of `x` to be `string` following the call. With the `immediate` modifier, control flow analysis assumes that the lambda expression _may_ have been invoked synchronously and thus considers the type of `x` to be `string | number` following the call.

It an error to apply the `immediate` modifier to anything but a parameter with a type that permits functions. Async arrow functions, async function expressions, and generator function expressions are always assumed to be deferred, regardless of whether their corresponding parameter includes an `immediate` modifier.

In JavaScript files, an immediate callback parameter can be declared using an `/** @immediate */` JSDoc annotation:

```ts
/**
 * @param {() => void} cb
 */
function callNow(/** @immediate */ cb) { ... }
```

Since the `immediate` modifier is considered an error by earlier versions of TypeScript, it isn't possible to add `immediate` modifiers to frameworks and libraries without also requiring the latest version of the compiler. The compiler therefore also recognizes `/** @immediate */` JSDoc comments in TypeScript files such that they can be used in cases where backwards compatibility matters. However, use of `immediate` modifiers is recommended whenever possible.

NOTE: This PR initially changed control flow analysis to assume possibly-immediately-called semantics by default and implemented a `deferred` modifier to indicate never-immediately-called semantics. While technically more sound (absent knowledge of whether or when a callback is called, it is safest and most conservative to assume it _may_ have been called), this turned out to require too much remediation in existing code. The PR now preserves the default never-immediately-called semantics and allows code to opt into possibly-immediately-called analysis using the `immediate` modifier.

Fixes #11498.
Fixes #15380.
Fixes #57880.


ahejlsberg:
@typescript-bot test it

fatcerberus:
I feel like this is just going to flip the problem on its head in the end - instead of people complaining because their immediately-invoked callbacks didn't affect narrowing, an equal number of people will probably now complain instead that callbacks they *know* will be called out-of-band *do* affect narrowing.  The goalposts don't even move really - the other team just gets possession is all 😄

fatcerberus:
https://github.com/electron-react-boilerplate/electron-react-boilerplate/blob/83b826fa9d810e26da7b5f17b4ed543e5f52bb17/src/main/main.ts#L101 is a good example

ahejlsberg:
I think it's clear that there's no one-size-fits-all answer to whether effects of executing lambda arguments should be reflected in CFA types. Our current assumption is that lambdas are _never_ executed synchronously. This PR experiments with the assumption they're _possibly_ executed, which, given no additional information about the function to which the lambda arguments are passed, is definitely a more sound assumption. Also, the PR validates an implementation strategy and gives us data on the performance cost of analyzing lambda effects.

I would nice to avoid modifiers (like `sync`, `immediate`, `deferred`) on callback parameters, but in the end it may be unavoidable. Modifiers come with the added complication that CFA will need to resolve the function's type, which may in turn lead to circularities.

ahejlsberg:
@typescript-bot test it

ethanresnick:
@ahejlsberg If I'm understanding correctly, this would apply to arguments that are declared as arrow functions or function expressions, but not function declarations. Is that right? If so, why exclude function declarations?

ahejlsberg:
@ethanresnick It only applies to arrow functions and function expressions passed directly as arguments. It does not apply to functions referenced through identifiers or other indirect expression constructs.

ahejlsberg:
@typescript-bot test it

ahejlsberg:
@typescript-bot test it

ahejlsberg:
@Igorbek For now we're going to focus on:

* Switching to a default assumption that lambda arguments _may_ have been invoked synchronously (because that is the soundest and most conservative assumption).
* Providing a `Default<T>` intrinsic type to mark callback parameters that are _never_ invoked synchronously.
* Introducing a new `--strictCallbackAnalysis` flag that enables this behavior (and is part of the `--strict` family of flags).

I don't think further modifiers are merited at this point. There cases where it is known that a callback is _always_ invoked, but they're rare and more complicated to analyze because ordering now starts to matter when there are multiple callback arguments.

ahejlsberg:
@typescript-bot test it

ahejlsberg:
@typescript-bot test it

AlQubaiisi:
في سبت، 8 يونيو، 2024 في 5:51 م، كتب TypeScript Bot <
***@***.***>:

ahejlsberg:
@typescript-bot test it

ethanresnick:
> @ethanresnick It only applies to arrow functions and function expressions passed directly as arguments. It does not apply to functions referenced through identifiers or other indirect expression constructs.

Thanks @ahejlsberg. Out of curiosity, what's the reason behind this restriction (i.e., only applying to functions written inline)? Obviously, TS needs the function's source code (not merely a type declaration) to apply this new aproach, but it could still apply to lots of standalone functions that modify some global state, and are passed as callbacks in various places, for which the source is available.

So, is this restriction motivated by performance (i.e., the current approach means there's a lot fewer functions to analyze)? Or, is it based on an observation that applying it to functions referenced by identifier etc. would lead to more false positives? Or, is there some implementation difficulty (e.g., the compiler internals make it inconvenient to know whether the source code is available for a given identifier)? Something else?

ahejlsberg:
@ethanresnick Reasons this only applies to inline lambda arguments and not standalone functions (or, generally, expressions that have a function type):

* It is easy to explain and understand: Inline lambdas are analyzed, other ways of referencing functions are not.
* With inline lambdas it is 100% decidable where a call to the argument goes. With standalone functions it is really only decidable for `function` declarations and function expressions or arrow functions assigned to `const` variables. It isn't decidable for function type variable declared with `let` or `var`, method declarations in classes or object literals, or general expressions that have a function type. So functionality could surprisingly disappear because of seemingly unrelated changes to declarations.
* The effects we're mostly interested in analyzing are mutations of captured outer local variables in callback functions. The predominant way that is expressed is through lambda arguments. It is less common to use nested declared functions.
* Performance considerations. Inline lambdas are generally smaller than full function declarations, and there are fewer of them.

None of this is set in stone, but currently what's in the PR seems like the sweet spot.

ethanresnick:
> - It is easy to explain and understand: Inline lambdas are analyzed, other ways of referencing functions are not.
>
> - With inline lambdas it is 100% decidable where a call to the argument goes. With standalone functions it is really only decidable for function declarations and function expressions or arrow functions assigned to const variables. It isn't decidable for function type variable declared with let or var, method declarations in classes or object literals, or general expressions that have a function type. So functionality could surprisingly disappear because of seemingly unrelated changes to declarations.

Hmm, ok. Part of the reason I was asking is because I totally share the concerns you're outlining (about wanting the rules to be easy to explain and understand, and not be effected by seemingly arbitrary declaration-style changes). But my initial reaction was that the current rules actually are a little weird along those dimensions.

Specifically, if I author a function inline and then find that I need to use the same function a second time (e.g., to handle another event), so I pull it out into a `const` above both usage sites, it'd be pretty surprising to have the CFA results suddenly disappear. Similarly, if my codebase happens to use the style of nested function declarations, I would lose this analysis; I agree that that style is less common, but it still feels surprising that it's the sort of thing that would influence CFA.

In terms of explainability, an ideal rule _might_ be something like: "This analysis applies whenever TS can unambiguously find the source code of the function being passed in", but I can see how that's probably impractical to implement. Like, recognizing that this function's source is fixed is probably too much:

```ts
const x: { readonly a: () => void } = { a() { y = true; } }
myFunction(x.a)
```

So, maybe a rule like: "This analysis applies to all standalone functions (i.e., not inside a class/object literal/other container) whose source code is fixed"? That rule would include `const x = () => { /* ... */ }`, and the other cases you mentioned as decidable. I agree it still has some cliffs that turn off CFA somewhat unexpectedly, like switching from `const` to `let`. But there's already plenty of precedent for TS reasoning about mutable and immutable things differently, and I do think it'd be nice (and less surprising) for the analysis not to suddenly turn off when users DRY up a repeated callback.

Whether that change is worth the performance hit though, idk.

gabritto:
@typescript-bot pack this

jakebailey:
@typescript-bot perf test faster host=node@22.3.0

jakebailey:
@typescript-bot perf test faster host=node@22.3.0

(this time it'll work)

andrewbranch:
@typescript-bot pack this

g-plane:
Are there any reasons about using `deferred` keyword instead of `Deferred<T>`? Also, what about move the `deferred` before function type but after parameter name like this:

```typescript
function setTimeout(callback: deferred (args: void) => void, ms?: number): NodeJS.Timeout;
```

magic-akari:
```TypeScript
class Foo {
    constructor(deferred public a: () => void) { }
}

class Bar {
    constructor(public deferred a: () => void) { }
}
```

It seems that both orders are acceptable, is this expected?

ahejlsberg:
> Are there any reasons about using `deferred` keyword instead of `Deferred<T>`?

One reason is that the "deferred" nature of a parameter is an attribute of the parameter itself, not the parameter's type. Another reason is that a `Deferred<T>` marker type could appear in a number of situations where it would have no effect (for example as the type of a local variable or property), which would be confusing.

>  Also, what about move the `deferred` before function type but after parameter

That would make `deferred` an attribute of the type, which we want to avoid for the same reasons as `Deferred<T>` above.

ahejlsberg:
> It seems that both orders are acceptable, is this expected?

No, we should verify that `deferred` appears as the first modifier.

yume-chan:
Function parameters can also be declared using tuples, would the `deferred` modifier be allowed there?

```ts
function a(...args: [deferred () => void]): void {} // The tuple element can have no name, so this looks very differently
function b(...args: [deferred callback: () => void]): void {}
function c(...args: Parameters<typeof a>): void {} // Still deferred?
```
---

Would TypeScript check the function body to ensure a non-deferred callback parameter is actually (or at least might) synchronously called? 

```ts
function a(callback: () => void) { } // Error: `callback` is marked as synchronous but not synchronously invoked

function b(callback: () => void) { // Error: `callback` is marked as synchronous but not synchronously invoked
    setTimeout(callback, 1000);
}

function c(callback: () => void) { // OK
    if (Math.random() > 0.5) {
        callback();
    }
}
```

---

I think the `Deferred<T>` approach supports the tuple case with minimal change.

It might also be beneficial to allow the call site to specify whether an argument will be invoked synchronously or not, especially when library authors has not adopted the change into their typings, so there will be many callbacks mistakenly considered synchronous.

```ts
// library.d.ts
declare function doThingsAsync(callback: () => void): void; // Not updated

// src/index.ts
let x: string | undefined;
doThingsAsync((() => { x = "foo" }) as Deferred<() => void>);
x; // Still `string | undefined`
```

ahejlsberg:
> Function parameters can also be declared using tuples, would the `deferred` modifier be allowed there?

No. The `deferred` modifier can only be applied to parameters (and not tuple elements), which reflects the reality of how control flow analysis analyzes deferred callback parameters: It just looks at the modifiers associated with the particular position of the parameter in the parameter list, avoiding the numerous circularities that full type resolution of the call signature would otherwise cause (another reason `Deferred<T>` isn't really an option).

> Would TypeScript check the function body to ensure a non-deferred callback parameter is actually (or at least might) synchronously called?

We discussed doing this, but it's not clear how useful it would be, and there are several complex issues. We could potentially check that no calls are made directly through the deferred parameter in the body of the function, but that is quickly defeated by us not analyzing aliasing and calls made to other functions. And attempting to track it in types such as `Deferred<T>` is even more daunting. It's not enough to know that a function reference is deferred, you'd also need to know with respect to what and somehow track that.

> It might also be beneficial to allow the call site to specify whether an argument will be invoked synchronously or not

The recommendation here is to use a simple wrapper function that just returns its argument:

```ts
function deferred<T extends (...args: any) => any>(deferred cb: T) { return cb }

declare function doThingsAsync(callback: () => void): void; // Not updated

let xx: string | undefined = undefined;
doThingsAsync(deferred(() => { xx = "foo" }));
xx; // Still undefined
```

DanielRosenwasser:
@typescript-bot test this
@typescript-bot pack this

DanielRosenwasser:
@typescript-bot test this

DanielRosenwasser:
@typescript-bot test it

DanielRosenwasser:
I'm having a hard time constructing a simple use case for where this might come up, but I have a slight concern around how the `deferred` modifier doesn't travel along in any sort of higher order - for example, when inferring the parameter list of a signature that uses `deferred` ([playground link](https://www.typescriptlang.org/play/?ts=5.6.0-pr-58729-75#code/KYDwDg9gTgLgBAE2AYwDYEMrDgMwK4B2yMAlhAYhAMox444AUSOwUWCuAXHAwJRwBeAHxwAbhBIJe3cZIDcAWABQypGkzZ8RUuThZkWdDGAAeAIJQA5gGc4oYwQS3CAawIQA7gQDaAXQA0cABKwLRQBEIMONxMwCxswBwAdCmYNtwWNvzCwaF44dI8KUlp1hlW1tkiIWEEiipKahhYuITEZBT6hsYAInGs7OYVdiAOTnCu7l5+gTX5EVExzAOJcMWl5VmC1XkFMesVm5XbubX1ygDeynA3cKihcCDc1jBQJASWcAA+cAR4ALYAI1Y9Vuj0EcAAjAAmADM5yUYIQ1Fo9AYfBOV0RYNuIAhACIABYkfGg24AX14CLBIHqAHo6TcAHoAfmU5OUnKUWLB93gAE9nq93p8fn8gSDrrd+RCYfCuWCusAjMAmCi6DheOiqnAeTibjKBHAiSSyTdKdTpfTGXBWeyuZcpTc+XAAF5Ct4fb6-AHAqBmt2yuGWm5KlV9eLsNU0DVajE5PX610E4mkp1wC0K26u63MtlKDlKIA)).

```ts
export declare function doStuff(deferred f: () => void): void;

declare function recreate<Args extends unknown[], Return>(f: (deferred ...args: Args) => Return): (...args: Args) => Return;

declare function recreateDeferred<Args extends unknown[], Return>(f: (deferred ...args: Args) => Return): (...args: Args) => Return;

{
    let x: string | number;
    x = 123;

    doStuff(() => {
        x = "hi";
    });

    x;
//  ^? let x: number
}


{
    let y: string | number;
    y = 123;

    recreate(doStuff)(() => {
        y = "hi";
    });

    y;
//  ^? let y: string | number
}


{
    let z: string | number;
    z = 123;

    recreateDeferred(doStuff)(() => {
        z = "hi";
    });

    z;
//  ^? let z: string | number
}
```

At the very least, this would be a good test case to add.

ahejlsberg:
Not sure what your intent is, but did you mean to have `recreateDeferred` reintroduce the `deferred` modifier?

```ts
declare function recreateDeferred<Args extends unknown[], Return>(f: (...args: Args) => Return): (deferred ...args: Args) => Return;
```

Certainly that works as expected.

ilogico:
Thinking on the "standard library" functions, I can only think of the Promise constructor as the one that executes the callback synchronously exactly once.
Most callbacks execute either asynchronously (callback!) or zero or more times (like in the array methods).
So I think the default should be deferred.
Otherwise, deferred needs to be added everywhere.

whzx5byb:
@ilogico Note that the OP says "a function call **may** be invoked synchronously during that call" so this would apply to array methods which will be called zero or more times.

tests/cases/compiler/deferredCallbacks.ts
```diff
@@ -0,0 +1,238 @@
...
+            a;  // (string | boolean)[]
+        }
+    );
+    a;  // string[]
+}
+
+// deferred modifier must precede public/private/protected/readonly
+
+class CC {
+    constructor(deferred public readonly x: () => void) {}
```
sandersn, line 102: Just to make sure, it would be good to have tests of a construct *call*. I didn't see any but I might have missed them.

rbuckton:
Is being permissive (treating unannotated callbacks as immediate invocations) vs being restrictive (treating unannotated callbacks as deferred invocations, e.g., the current behavior) going to cause us problems down the road?

If we were to choose the restrictive approach and use an `immediate` modifier for callbacks that are immediately invoked, we could theoretically infer `immediate` from the body of a function based on usage, e.g., if the parameter is immediately invoked or passed as an argument to an `immediate` callback parameter before an `await` or `yield`, then we could infer `immediate` for the parameter on behalf of the user.

By choosing the permissive approach, we will assume many callbacks are immediately invoked that may not be and would have a far harder time inferring `deferred` given the wealth of existing unannotated code, so we may never be able to offer that capability.

tests/cases/compiler/deferredCallbacks.ts
DanielRosenwasser: Yeah; testing a variety of types with annotated construct signatures would be good too, like

```ts
interface Ctor1 {
  new(deferred cb: () => void): {}
}

type Ctor2 = new (deferred cb: () => void) => void;
```

DanielRosenwasser:
> Not sure what your intent is, but did you mean to have `recreateDeferred` reintroduce the `deferred` modifier?

Probably, yes - but I see that you added test cases for all of them in `recreate`, `recreateDeferred1`, and `recreateDeferred2`. Thanks!

tests/cases/compiler/deferredCallbacks.ts
```diff
@@ -0,0 +1,238 @@
...
+async function run() {
+    let result: boolean = true;
+    await execute(() => {
+        result = false;
+    });
+    if (result === false) {
+        console.log("error");
+    }
+    return result;
+}
```
sandersn, line 238: Can you add a test like

```ts
class C {
  constructor(options?: {
    /** @deferred */ fn?: Function | undefined
  }) { }
}
```

I failed to get an error in this case in [my node PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/70084)

bluepnume:
> Async arrow functions, async function expressions, and generator function expressions are always assumed to be deferred, regardless of whether their corresponding parameter includes a deferred modifier.

Is that a sound assumption? It's entirely possible to call an async function synchronously right?

```typescript
console.log('1');

[ 1, 2, 3 ].forEach(async () => {
    console.log('2')
});

console.log('3');
```

andrewbranch (dismissed):
There’s a slight issue with not considering `deferred` (or `immediate`, if we go that direction) in signature relations. I know we briefly discussed this having no effect on assignability, and I think that’s probably for the best, but I think it’s worth fixing subtype / strict subtype relations for union reduction. The current approach leads to a type ID order dependence bug [[PR playground](https://www.typescriptlang.org/play/?ts=5.6.0-pr-58729-75#code/PTAEGUHcEMAdQC4AsCmoD2AnAJizGAzRVAZzQUnVAGNoAbOkgKAIBEUC9MVsAKXgJSgAvAD5QAbwC+AgNwsAkgFslPAJbQEKfkLGSZ8prmp1o3UAQCuAO2oI16axfadM3PrlfuaAIwBcoIIi4gBu6GrYAgFhEfLGpuZWtvaOFsqq2BpavNT+gbqh4ZHRRYZ0KAgWIqAAsppIAHSY0NbY6EpB4gAMDQCsoAD8zhxcPKABBOnqmihlFaAAHgEkCJhq1gDmoAA+oNaWSj541QBEAPIA0ieGBDrBkkygi9UAjF3yBkyPTwvyIKAAPQGTCAA)]:

```ts
// Swap the order of these two calls
fDeferred(() => {});
fImmediate(() => {});

declare function fDeferred(deferred cb: () => void): void;
declare function fImmediate(cb: () => void): void;

let f = Math.random() > 0.5 ? fDeferred : fImmediate;

let x: string | number = "OK";

f(() => {
  x = 10;
});

   x;
// ^?
```

The type of `x` is dependent on the type of `f` which is dependent on the order of the first two calls, since the signatures of `fDeferred` and `fImmediate` are indistinguishable in the strict subtype relation.

ahejlsberg:
@typescript-bot test it

ahejlsberg:

jakebailey:
@typescript-bot perf test this faster predictable=true

ahejlsberg:
@jakebailey Asking for predictable seems to have very unpredictable effects. 😮

ahejlsberg:
@jakebailey Also, the issue with switching to actual `immediate` modifiers in the library files is that it requires a new LKG compiler.

ahejlsberg:
@jakebailey The issue with switching to actual `immediate` modifiers is that we need an updated ESLint.

jakebailey:
Even with the keyword, https://github.com/microsoft/TypeScript/pull/59580#issuecomment-2278672235 shows roughly the same perf regression in bind. Still not sure what it could be reading [`5ae48a2` (#58729)](https://github.com/microsoft/TypeScript/pull/58729/commits/5ae48a223af04b060175760cf24fdc5a81827b4b).

dgreensp:
I'm just a passer-by (who writes a lot of TypeScript), who got excited by the prospect of a new keyword that would improve CFA or soundness, but I'm not sure I understand the motivation here, relative to the magnitude of the change (new syntax).

I guess it started as an effort to improve soundness by considering that called functions might have side effects, and now it basically helps correct for overzealous CFA.

Here is another case (not related to this feature) where CFA thinks it knows everything and narrows the type annotation of a local variable:

```ts
function foo(a: number[], i: number) {
    // I'm serious, Typescript, I'm saying this could be out-of-bounds!
    const n: number | undefined = a[i];
    const n2 = n; // : number
}
```

In both cases, you have to occasionally override CFA, e.g. one could write:

```ts
declare function mystery(cb: () => void): void;

let x: string | number = "OK" as string | number;
mystery(() => {
    x = 10;
});
if (x === 10) { // works
}
```

In terms of the larger issue of side effects, I think it should at least be acknowledged that passing any closure to any function could lead to functions having side effects that affect local variables; it doesn't really depend on whether a callback is invoked "immediately" or "deferred."

For example:

```ts
const foo = new Foo(() => { /* ... set a local variable */ });
foo.execute();
```

Here, the callback is not called immediately, but it is called during the course of the function.  The point is, the question of whether a callback might be called immediately or not by a function is only loosely connected to the question of whether a function call has side effects that might affect the analysis of local variables.

I think an annotation that says "this callback is called exactly once, synchronously" or "this callback is called at least once, synchronously" might even be more useful?  It would serve a slightly different purpose.  If this keyword opens the door to other keywords, I'm for it.  There are a lot of keywords I would want.

SimonSchick:
Is this going to make the 5.7 release? I see it is in the iteration plan and I'm curious what work is remaining here 👀

jakebailey:
No, 5.7 is cut, and its release candidate is already out. https://devblogs.microsoft.com/typescript/announcing-typescript-5-7-rc/

Something being in the iteration plan does not guarantee its completion / inclusion in a release.

PR #58735, opened 2024-06-02
Output Config Option Values in Canonical Case

graphemecluster:
This aligns the letter cases of the values in `tsconfig.json`s generated by `tsc --init` and some diagnostics (e.g. TS1501, TS6046) with the autocomplete values provided to `tsconfig.json` in VSCode and other diagnostics with hard-coded target versions.
This should not affect anything related to value parsing.

Addresses https://github.com/microsoft/TypeScript/pull/55600#discussion_r1571605851.


jakebailey:
For ecmascript versions I can sort of understand it, but I'm not totally convinced we consider "Node16" more canonical than "node16", "Preserve" etc, https://www.typescriptlang.org/tsconfig/#module

And so on.

graphemecluster:
I don’t know, I just completely sticked to the autocompletions provided by VSCode (except for a few subtle inconsistencies of `lib` values, which should be fixed in VSCode). So perhaps changes to the values should be made on the VSCode side synchronically.

jakebailey:
The autocompletions in VS Code are just the ones from schemastore. Not sure any of us like those either 😄

graphemecluster:
Oh, I see. Wonder why we didn’t maintain one here or in the VSCode repo.
Anyway, may you give me a list of what should be kept and what should be discarded once there is a consensus on the letter cases? Do you prefer camel case to Pascal case?

PR #58753, opened 2024-06-03
Bump version to 5.6.

DanielRosenwasser:


jakebailey (approved):
LGTM if we're not going to merge main back into release-5.5 again. Probably not enough fixes to warrant that.

PR #58755, opened 2024-06-03
Check cancellation token in isolatedDeclarations codefix

jakebailey:
These are the main three entrypoints into the fixer. It's possible that there's some common hot function we could check instead (`inferTypes`?) but the call tree didn't make one super clear.


DanielRosenwasser (approved):

DanielRosenwasser:

PR #58760, opened 2024-06-03
Write buildInfo even if we dont emit js or dts for --outFile scenarios

sheetalkamat:
This makes `--outFile` behave similar to not specifying it.
Previously we use to store the bundle section information in the `outFileBuildInfo` which means we could not emit buildInfo without emitting the other output files and had to be always in sync. But now that we are not storing that information with no support for `prepend`, we can write buildInfo even if the other outputs like js or dts were not emitted. This helps with storing semantic diagnostics information and dts errors information in the buildInfo so we can re-use it just like multiFile scenario.
The changes are:
- Cache semantic diagnostics in `--outFile` scenario
- Ensuring all places that checked for outFile not present when emitting buildInfo explicitly to disable that
- Ensuring emitter can emit only outFile's js, dts, map etc without emitting buildInfo so that buildInfo can be synchronized correctly with the builder state for changeFileSet, dts diagnostics etc (this is similar to multiFileScenario where we emit all affected files one by one and then emit build info. In --outFile we have to emit program without emitting buildInfo and then emit buildInfo in later stage)

### DIFF
--- a/src/compiler/tsbuildPublic.ts
+++ b/src/compiler/tsbuildPublic.ts
@@ -1452,7 +1452,6 @@ function buildErrors<T extends BuilderProgram>(
     buildResult: BuildResultFlags,
     errorType: string,
 ) {
-    // Since buildinfo has changeset and diagnostics when doing multi file emit, only --out cannot emit buildinfo if it has errors
     reportAndStoreErrors(state, resolvedPath, diagnostics);
     state.projectStatus.set(resolvedPath, { type: UpToDateStatusType.Unbuildable, reason: `${errorType} errors` });
     return { buildResult, step: BuildStep.EmitBuildInfo };
--- a/src/compiler/tsbuildPublic.ts
+++ b/src/compiler/tsbuildPublic.ts
@@ -90,6 +90,7 @@ import {
     PollingInterval,
     Program,
     ProgramBuildInfo,
+    ProgramBundleEmitBuildInfo,
     ProgramHost,
     ProgramMultiFileEmitBuildInfo,
     ProgramUpdateLevel,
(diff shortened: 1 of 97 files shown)


src/compiler/tsbuildPublic.ts
```diff
@@ -1453,20 +1448,14 @@ function afterProgramDone<T extends BuilderProgram>(
 function buildErrors<T extends BuilderProgram>(
     state: SolutionBuilderState<T>,
     resolvedPath: ResolvedConfigFilePath,
-    program: T | undefined,
-    config: ParsedCommandLine,
     diagnostics: readonly Diagnostic[],
     buildResult: BuildResultFlags,
     errorType: string,
 ) {
     // Since buildinfo has changeset and diagnostics when doing multi file emit, only --out cannot emit buildinfo if it has errors
```
jakebailey, line 1455: Is this comment still relevant?

sheetalkamat:
@typescript-bot test this
@typescript-bot test top100
@typescript-bot run dt
@typescript-bot pack this

jakebailey (approved):

AlQubaiisi:
في ثلاثاء، 4 يونيو، 2024 في 4:15 ص، كتب TypeScript Bot <
***@***.***>:

jakebailey:
Uh, why does that perf run look like that?

sheetalkamat:
To me it seems more like a noise. https://github.com/microsoft/TypeScript/pull/57196 didnt change anything in compiler and still shows increase in symbols etc 
Also `Req 3 - references` increase makes no sense, no change in tsserver at all

jakebailey:
I was referring to the bit where ts-pre-modules stopped compiling

PR #58771, opened 2024-06-04
Allow references to the global Symbol in computed property names under isolatedDeclarations

weswigham:
Specifically, they must look like `Symbol.something` or `globalThis.Symbol.something`, and in the first case, that first identifier must refer exactly to the global `Symbol` value (or not resolve at all, in which case it's the missing global by default). In the second, `globalThis` must exactly refer to the global `globalThis`.

This should allow the common cases like `Symbol.iterator` members and the like, without allowing complicated ones that might not be right like imports that refer to arbitrary maybe-symbol names.

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -1511,6 +1511,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             Debug.assert(isExpressionNode(node));
             return getSymbolAtLocation(node) === undefinedSymbol;
         },
+        isDefinitelyReferenceToGlobalSymbolObject,
     });
     var evaluate = createEvaluator({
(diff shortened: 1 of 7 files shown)


jakebailey (approved):

jakebailey:
Is this something we should backport to 5.5?

MichaelMitchell-at:
sogood

MichaelMitchell-at:
> Is this something we should backport to 5.5?

Perhaps in 5.5.4? 

Also we just stumbled on some code today that could really use something like https://github.com/microsoft/TypeScript/pull/58829, since it uses custom created `unique symbols`

DanielRosenwasser:
I'll discuss with others whether it's a good idea to backport, but in the meantime let's just kick it off.

PR #58782, opened 2024-06-05
Some refactoring so we arent checking undefined on program or asserting it so much

sheetalkamat:


src/compiler/watch.ts
```diff
@@ -329,8 +329,8 @@ function createTabularErrorsDisplay(filesInError: (ReportFileInError | undefined
 }
 
 /** @internal */
-export function isBuilderProgram(program: Program | BuilderProgram): program is BuilderProgram {
-    return !!(program as BuilderProgram).getState;
+export function isBuilderProgram<T extends BuilderProgram>(program: Program | BuilderProgram): program is T {
```
jakebailey, line 332: This seems unsafe; this could conform to anything, right?

jakebailey: I guess it's copied from somewhere else...

jakebailey (approved):
Reminds me that we should see if we can make `Debug.checkDefined` or `Debug.assert` reject always-defined or never-falsy values (statically or at lint time), in case there are more of these that could be removed.

src/compiler/watch.ts
sheetalkamat: Yep

PR #58783, opened 2024-06-06
Store node children WeakMaps in vars and bind its methods as locals.

DanielRosenwasser:


DanielRosenwasser:

src/compiler/factory/nodeChildren.ts
```diff
@@ -4,18 +4,24 @@ import {
     Node,
 } from "../_namespaces/ts.js";
 
-const nodeChildren = new WeakMap<Node, readonly Node[] | undefined>();
```
jakebailey, line 7: No need to var at the top level, esbuild already does this.

Issue #58781, opened 2024-06-05
[compiler crash] TypeError: Cannot read properties of undefined (reading 'kind')


MichaelMitchell-at:
### 🔎 Search Terms

TypeError: Cannot read properties of undefined (reading 'kind')

### 🕗 Version & Regression Information

- This is a crash
- This changed between versions 5.5.0-dev.20240523 and 5.5.0-dev.20240524
- Bisects to https://github.com/microsoft/TypeScript/commit/2b4e7e34e7a40fe0d886aec6f799954420f2ac0c (https://github.com/microsoft/TypeScript/pull/58546) cc @dragomirtitian @weswigham 

### ⏯ Playground Link

https://www.typescriptlang.org/dev/bug-workbench/?target=7&emitDeclarationOnly=true&ts=5.5.0-dev.20240524#code/PTAEAEBMFMGMBsCGAnRAXAlgewHYC5Q1kBXaAKBAmgFsM0AROJVTXAeR3gE8CjSKw4AM4ALLAHcAorTQCIoidLppokAGIZ40XlwAO0AHSQDaIXOFipMles3bQGHDAAeRk2fMAzOzkTV7aHqGpmTQzrpYyGig0DjE1KCScQkA3mSgoACCoAC8oABEiPkANGQAvmSh4ZHRgfqgACpBuaApoM4A-ARtANpJ8QaZALoEAAygZRMA3JWU4N5avv4Eji7uZBjUEVGthM1N9ZOeyFgJ+QbAddD5M1Xb0bC4QtGeWFgtbQZfABQpk4hCRpBACU0zIQA

### 💻 Code

```ts
// @declaration: true
// @emitDeclarationOnly: true
// @showEmit
// @showEmittedFile: type.d.ts
// @showEmittedFile: index.d.ts

// @filename: type.ts
export enum Enum {
  A = "a",
}

export type Type = { x?: { [Enum.A]: 0 } };

// @filename: index.ts
import { type Type } from "./type";

export const foo = { ...({} as Type) };
```

### 🙁 Actual behavior

Compiler crash
```
TypeError: Cannot read properties of undefined (reading 'kind')
    at visitExistingNodeTreeSymbolsWorker (https://playgroundcdn.typescriptlang.org/cdn/5.5.0-dev.20240524/monaco/dev/vs/language/typescript/tsWorker.js:61261:29)
[...]

PR #58786, opened 2024-06-06
Fixed declaration emit crash related to enum entity name expressions

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/58781

cc @weswigham @dragomirtitian


src/compiler/checker.ts
```diff
@@ -8856,8 +8856,18 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
                         const computedPropertyNameType = typeToTypeNodeHelper(type, context);
-                        Debug.assertNode(computedPropertyNameType, isLiteralTypeNode);
-                        const literal = computedPropertyNameType.literal;
+                        let literal;
+                        if (isLiteralTypeNode(computedPropertyNameType)) {
+                            literal = computedPropertyNameType.literal;
+                        }
+                        else {
+                            const evaluated = evaluateEntityNameExpression(node.expression);
+                            const literalNode = typeof evaluated.value === 'string' ? factory.createStringLiteral(evaluated.value, /*isSingleQuote*/ undefined) :
```
Andarist, line 8865: creating those objects is wasteful here but it feels like it keeps the code cleaner, I can change this if you request that change

src/compiler/checker.ts
```diff
@@ -8856,8 +8856,18 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
                         const type = getWidenedType(getRegularTypeOfExpression(node.expression));
                         const computedPropertyNameType = typeToTypeNodeHelper(type, context);
-                        Debug.assertNode(computedPropertyNameType, isLiteralTypeNode);
-                        const literal = computedPropertyNameType.literal;
+                        let literal;
+                        if (isLiteralTypeNode(computedPropertyNameType)) {
+                            literal = computedPropertyNameType.literal;
+                        }
+                        else {
+                            const evaluated = evaluateEntityNameExpression(node.expression);
```
Andarist, line 8864: the result of this matches the 5.4 behavior, that's why I've reached for evaluating this but perhaps there are reasons why this is wrong

src/compiler/checker.ts
```diff
@@ -8856,8 +8856,18 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     else {
                         const type = getWidenedType(getRegularTypeOfExpression(node.expression));
                         const computedPropertyNameType = typeToTypeNodeHelper(type, context);
```
Andarist, line 8858: this can return `ImportType` in the enum's case, that's the reason why it crashed - the code here assumed that a literal type node would always be returned but that wasn't true

src/compiler/checker.ts
```diff
@@ -8856,8 +8856,18 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+                        let literal;
+                        if (isLiteralTypeNode(computedPropertyNameType)) {
+                            literal = computedPropertyNameType.literal;
+                        }
+                        else {
+                            const evaluated = evaluateEntityNameExpression(node.expression);
+                            const literalNode = typeof evaluated.value === "string" ? factory.createStringLiteral(evaluated.value, /*isSingleQuote*/ undefined) :
+                                typeof evaluated.value === "number" ? factory.createNumericLiteral(evaluated.value, /*numericLiteralFlags*/ 0) :
+                                undefined;
+                            Debug.assert(literalNode);
```
weswigham, line 8868: I don't think this assertion is legit - unless you make assumptions about the computed property name being "correct" in some way, its' expression could resolve to *any* type (as the issue prompting this showed). Notably, `evaluateEntityNameExpression` will just return `undefined` if the entity name doesn't resolve, which'll in turn trigger this assertion. In such a case, we *probably* want to retain the whole computed name as-is, but mark an error?

Andarist: I couldn't cause a crash using invalid computed property names but I managed to find a crashing case with symbols. Could you take another look?

tests/baselines/reference/declarationEmitComputedPropertyNameSymbol1.js
```diff
@@ -0,0 +1,61 @@
...
+    
+==== index.d.ts (2 errors) ====
+    export declare const foo: {
+        x?: {
+            [Foo.sym]: 0;
+            ~~~~~~~~~
+!!! error TS1170: A computed property name in a type literal must refer to an expression whose type is a literal type or a 'unique symbol' type.
+             ~~~
+!!! error TS2552: Cannot find name 'Foo'. Did you mean 'foo'?
+        };
```
Andarist, line 59: I had an issue in the test file - after fixing it, I see now that this one is wrong. I'm looking into fixing it.

andrewbranch:
Not sure how many of these repos have declaration emit, but

@typescript-bot test top400

weswigham (approved):
No longer incorrectly asserts, has fallback behaviors when each attempt fails to produce something usable here, seems good to me~

jakebailey:

Issue #58794, opened 2024-06-07
False positive ts(2373)


worstpractice:
### 🔎 Search Terms

Parameter cannot reference identifier declared after it, ts(2373)

### 🕗 Version & Regression Information

This is the behavior in every version I tried, and I reviewed the FAQ for entries about anything related (and found nothing).

### ⏯ Playground Link

https://www.typescriptlang.org/play/?noUncheckedIndexedAccess=true&noUnusedLocals=true&noUnusedParameters=true&esModuleInterop=false&declaration=false&target=99&jsx=0&noErrorTruncation=true&pretty=true&isolatedDeclarations=true&verbatimModuleSyntax=true&useUnknownInCatchVariables=true&exactOptionalPropertyTypes=true&noFallthroughCasesInSwitch=true&noImplicitOverride=true&noPropertyAccessFromIndexSignature=true&ts=5.6.0-dev.20240606&ssl=18&ssc=1&pln=1&pc=1#code/PQKhCgAIUgVALAlgZ0gMwK4DsDGAXRAey0j3gCdCB3VAAwCUBTNR8x3RgUXMvIC5IAYQCGWLITyRhOHI2SoA5ABt2AczILIAI2aE2kRFkQFhSxAC9hBYrSmTy2AgFtGAOigRocAMqQcetnwlAE9INgAHPTw6aIAKACYAZgB2RIBKWldIQF4NwFkdj2BwRgAPSPJJfyxkSTwHRgAFQmRjRAA3RkgAXkgAHkrquAA+WIBvSABtFSx1eEgAWkgARgBdATRDUwBJPEYnABpIKZnIAF8BNmEAE2IQuHHltIFYSAAfSGxL5kNGS67ByBGUDCjDwGHIJHWWC2OycAG5wCd4eBQJ4YAgUOhsPgiCQqHoANaoSEdKxhRyIFzuaAeHzQECGMD+HiMIKhCJRGLIBIpdKZHKAOR2CkVSlE-MQBmhTMgGk0Wu0ur1+pJYMMxkcyAdJmoyPMlqt0BslNtdqdzowrjdQrB7o84K93lhPsTfp1-oDIMDQeCDVCjTD4YjwEA

### 💻 Code

```ts
/**
 * This function throws `ReferenceError: Cannot access 'length' before initialization` at runtime.
 *
 * TS correctly reports `ts(2373)`. 👍
 */
[...]

PR #58797, opened 2024-06-07
Fixed a false positive 2373 with computed property names in binding elements

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/58794

### DIFF
--- a/src/compiler/utilities.ts
+++ b/src/compiler/utilities.ts
@@ -11433,7 +11433,8 @@ export function createNameResolver({
                     if (
                         lastLocation && (
                             lastLocation === (location as BindingElement).initializer ||
-                            lastLocation === (location as BindingElement).name && isBindingPattern(lastLocation)
+                            lastLocation === (location as BindingElement).name && isBindingPattern(lastLocation) ||
+                            lastLocation === (location as BindingElement).propertyName
                         )
                     ) {
                         if (isPartOfParameterDeclaration(location as BindingElement) && !associatedDeclarationForContainingInitializerOrBindingName) {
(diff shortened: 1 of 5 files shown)


tests/baselines/reference/classWithStaticFieldInParameterBindingPattern.2(target=es2015).errors.txt
```diff
@@ -7,7 +7,7 @@ classWithStaticFieldInParameterBindingPattern.2.ts(6,57): error TS2373: Paramete
     class C {}
     (({ [class extends C { static x = 1 }.x]: b = "" }) => { var C; })();
                        ~
-!!! error TS2373: Parameter '{ [class extends C { static x = 1 }.x]: b = "" }' cannot reference identifier 'C' declared after it.
+!!! error TS2373: Parameter 'b' cannot reference identifier 'C' declared after it.
```
jakebailey, line 10: All of these new error changes don't feel right, no? The problem is that you can't alias the class and the var since it'll be downleveled, but now we're blaming `b` but the error position is still in the old place?

Andarist: I agree that it doesn't completely feel right. I think it's a separate problem though. This is how TS reports errors referring to binding elements in those kinds of positions today:
```ts
// Parameter 'foo' cannot reference identifier 'bar' declared after it.
function test({ foo = bar, b: bar }: any) {}
```

This is a little bit different because this time it's the `.name` that contains the problem... but was the old "parameter span" that much better here? I could bring back the original "parameter span" in those errors but based on the example above I concluded that the new one is fine. Happy to put some extra work into this if you feel this deserves it.

I see what you are saying about downleveling and all but this one errors with all targets. That would make it yet another issue as this (beautiful) piece of code shouldn't be flagged as a problem in es2015+

Issue #58807, opened 2024-06-08
[nightly] Syntactically invalid declaration files emitted under extremely specific conditions


MichaelMitchell-at:
### 🔎 Search Terms

declaration syntax invalid comment docstring 

### 🕗 Version & Regression Information

- Bisected this to https://github.com/microsoft/TypeScript/commit/2b4e7e34e7a40fe0d886aec6f799954420f2ac0c cc @dragomirtitian @weswigham 

### ⏯ Playground Link

https://www.typescriptlang.org/dev/bug-workbench/?target=7&composite=true&ts=5.6.0-dev.20240608#code/PTAEAEGMHsFsAdoGcCWAXApgLlGgTgK4YBQIES+KkaO+RpY4AJhpADYCGeHaK0AdrUIky4JAAtoAdwCisdA3KTZ8tJiYAxFG2ygOAOib60SYovAAzbRn4dYugyeIoE0PGlABvUACNoADwwmACEA0ABfUAs8OFAAIn1gP0CQgLiAbjMMf0R3UBh+ClAAfVAAXl8AoND-TMUAQQ8UfhZ-UAAOAGZSACoe4lAe0CkeAZ7gM1ErHVt7HGTqgMNjU2zcjxZ2Lgx8gSKF1P8cF3WACgSktIBKfRqAHk8B0GeD49d3c8TkuJv7-gJYD4MHgAHyZcJg55Q6Ew2HQsgAPQAKuIUEhQGiMS1sh1OljcKj0dMSOZibNdMllk41m4NqxOHgdlskOj7kiQV4ns8AG4cNhEHBIzLQgoUQjUNynXn83RIq7g4g0vKbBk7CwEfjUPj8Sr+O7sqV8gWgOU4Nlg4hAA

### 💻 Code

```ts
// @composite: true
// @strict: true
// @declaration: true
// @showEmit
// @showEmittedFile: a.d.ts

// @filename: a.ts
import { boxedBox } from "./boxedBox";

export const _ = boxedBox;

// At index 83
/**
 * wat
 */

// @filename: boxedBox.d.ts
export declare const boxedBox: import("./box").Box<{
    boxed: import("./box").Box<number>;
}>;                        // ^This is index 83 in this file

// @filename: box.d.ts
export declare class Box<T> {
    value: T;
    constructor(value: T);
}
export declare function box<T>(value: T): Box<T>;
```

### 🙁 Actual behavior

`a.d.ts` is emitted as

```ts
[...]

PR #58810, opened 2024-06-09
Fixed declaration emit issue related to a qualifier being reused cross-file

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/58807

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -8840,7 +8840,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                         node,
                         factory.updateLiteralTypeNode(node.argument, rewriteModuleSpecifier(node, node.argument.literal)),
                         node.attributes,
-                        node.qualifier,
+                        visitNode(node.qualifier, visitExistingNodeTreeSymbols, isEntityName),
                         visitNodes(node.typeArguments, visitExistingNodeTreeSymbols, isTypeNode),
                         node.isTypeOf,
                     );
--- a/tests/baselines/reference/declarationEmitTopLevelNodeFromCrossFile2.js
+++ b/tests/baselines/reference/declarationEmitTopLevelNodeFromCrossFile2.js
@@ -1,0 +1,43 @@
+//// [tests/cases/compiler/declarationEmitTopLevelNodeFromCrossFile2.ts] ////
+
+//// [a.ts]
+import { boxedBox } from "./boxedBox";
+
+export const _ = boxedBox;
+
+// At index 83
+/**
+ * wat
+ */
+
+//// [boxedBox.d.ts]
+export declare const boxedBox: import("./box").Box<{
+    boxed: import("./box").Box<number>;
+}>;                        // ^This is index 83 in this file
+
+//// [box.d.ts]
+export declare class Box<T> {
+    value: T;
+    constructor(value: T);
+}
+export declare function box<T>(value: T): Box<T>;
+
+//// [a.js]
+"use strict";
+Object.defineProperty(exports, "__esModule", { value: true });
+exports._ = void 0;
+var boxedBox_1 = require("./boxedBox");
+exports._ = boxedBox_1.boxedBox;
+// At index 83
+/**
+ * wat
+ */
+
+
+//// [a.d.ts]
+export declare const _: import("./box").Box<{
+    boxed: import("./box").Box<number>;
+}>;
+/**
+ * wat
+ */
--- a/tests/baselines/reference/declarationEmitTopLevelNodeFromCrossFile2.symbols
+++ b/tests/baselines/reference/declarationEmitTopLevelNodeFromCrossFile2.symbols
@@ -1,0 +1,47 @@
+//// [tests/cases/compiler/declarationEmitTopLevelNodeFromCrossFile2.ts] ////
+
+=== a.ts ===
+import { boxedBox } from "./boxedBox";
+>boxedBox : Symbol(boxedBox, Decl(a.ts, 0, 8))
+
+export const _ = boxedBox;
+>_ : Symbol(_, Decl(a.ts, 2, 12))
+>boxedBox : Symbol(boxedBox, Decl(a.ts, 0, 8))
+
+// At index 83
+/**
+ * wat
+ */
+
+=== boxedBox.d.ts ===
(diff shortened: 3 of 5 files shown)


src/compiler/checker.ts
```diff
@@ -8840,7 +8840,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                         node,
                         factory.updateLiteralTypeNode(node.argument, rewriteModuleSpecifier(node, node.argument.literal)),
                         node.attributes,
-                        node.qualifier,
+                        visitNode(node.qualifier, visitExistingNodeTreeSymbols, isEntityName),
```
MichaelMitchell-at, line 8843: I definitely don't have enough context on the workings of the checker to give any informed feedback, but I can at least ask at a high level do we have some confidence that whether anything more needs to be done to avoid this general class of problem?

Andarist: Tbh, I don't know this part of TS too well either so we need to wait for @weswigham or @dragomirtitian to give some feedback on this. I'd expect that other things in this function (like `node.attributes` here) should be wrapped like this too (if only this is the right fix).

jakebailey: I would definitely suspect a lot of this needs this same fix.

weswigham: 100% - any child node we copy should have its' children visited - any missing visitor calls on manually remapped nodes are an oversight, presumably.

weswigham (approved):
@jakebailey do we just want to take this as-is, or should we use this opportunity to audit all the factory calls in the visitor?

jakebailey:
I think it'd be fine to merge this since it clearly helps code in the wild, but I do think we should audit effectively everything and backport what we find.

Andarist:
I can make this audit as part of this PR (or separately), if you are OK with waiting till tomorrow.

Andarist:
aye aye, captain

src/compiler/checker.ts
```diff
@@ -8765,7 +8765,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                                     /*modifiers*/ undefined,
                                     getEffectiveDotDotDotForParameter(p),
                                     setTextRange(context, factory.createIdentifier(getNameForJSDocFunctionParameter(p, i)), p),
-                                    p.questionToken,
+                                    p.questionToken ? factory.createToken(SyntaxKind.QuestionToken) : undefined,
```
jakebailey, line 8768: Does `factory.cloneNode` not work on tokens?

Andarist, line 8768: Some other code paths in this function do a similar "cloning" of the `.questionToken` so I assume that it's the right thing to do here and in other places that I touched

src/compiler/checker.ts
```diff
@@ -8798,7 +8798,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 if (isTypeParameterDeclaration(node)) {
                     return factory.updateTypeParameterDeclaration(
                         node,
-                        node.modifiers,
+                        visitNodes(node.modifiers, visitExistingNodeTreeSymbols, isModifier),
```
Andarist, line 8801: those modifiers are just tokens so i guess i could just map over them and clone them?

src/compiler/checker.ts
```diff
@@ -8839,8 +8839,8 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     return factory.updateImportTypeNode(
                         node,
                         factory.updateLiteralTypeNode(node.argument, rewriteModuleSpecifier(node, node.argument.literal)),
-                        node.attributes,
-                        node.qualifier,
+                        visitNode(node.attributes, visitExistingNodeTreeSymbols, isImportAttributes),
```
Andarist, line 8842: those are not jus flat tokens so I wasn't quite sure what to do about those, it seemed like maybe i should deep clone them?

src/compiler/checker.ts
```diff
@@ -8904,9 +8904,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
                     }
                     else {
-                        parameterName = node.parameterName;
+                        parameterName = factory.cloneNode(node.parameterName);
                     }
-                    return factory.updateTypePredicateNode(node, node.assertsModifier, parameterName, visitNode(node.type, visitExistingNodeTreeSymbols, isTypeNode));
+                    return factory.updateTypePredicateNode(node, node.assertsModifier ? factory.createToken(SyntaxKind.AssertsKeyword) : undefined, parameterName, visitNode(node.type, visitExistingNodeTreeSymbols, isTypeNode));
                 }
 
                 if (isTupleTypeNode(node) || isTypeLiteralNode(node) || isMappedTypeNode(node)) {
```
Andarist, line 8912: there are also places in this function that do smth like:
```ts
if (visited === node) {
  visited = factory.cloneNode(node)
}
```

this pattern is not applied consistently across the whole function, should it be? 🤔

src/compiler/checker.ts
Andarist: It does but this short circuits some of its logic, we know exactly what we want here. Unless those should have `.original` attached to them and such - if that's the case then it would be easier to use `cloneNode`. I went this route since some other code path here was already doing it like this, IIRC

jakebailey: Being consistent is fine with other code around it (I didn't look around)

src/compiler/checker.ts
jakebailey: I suspect yes.... Sure seems like though that we should make some sort of wrapper for the visitor func that would check that. Though I feel like I've seen that exact code in some visitor transformer code already (maybe we're not going through that)

Andarist: I think this might be a bigger change so I'd leave it out of this PR. I just want to flag this as a potential problem.

src/compiler/checker.ts
Andarist: Those 2 call sites related to `.questionToken` are also related to JSDoc. I wonder if this cloning even matters in this context.

jakebailey (approved):
LGTM, likely need to be cherry picking this to 5.5.

jakebailey:

PR #58818, opened 2024-06-10
refactor: improve string export name completions

Jack-Works:
continue of #49297 and #58640

### DIFF
--- a/src/services/completions.ts
+++ b/src/services/completions.ts
@@ -336,6 +336,7 @@ import {
     rangeContainsPosition,
     rangeContainsPositionExclusive,
     rangeIsOnSingleLine,
+    scanner,
     ScriptElementKind,
     ScriptElementKindModifier,
     ScriptTarget,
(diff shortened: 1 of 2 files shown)


src/services/completions.ts
```diff
@@ -1855,6 +1863,13 @@ function createCompletionEntry(
         }
     }
 
+    // TODO: the output code is still invalid (QualifiedName current cannot have form in namespace["invalid text name"] and but is no other way to access it)
+    // generating namespace["invalid text name"] with a type error is better than generating 'namespace.invalid text name' and have a syntax error.
+    if (needsConvertPropertyAccess && contextToken?.kind === SyntaxKind.DotToken && contextToken.parent.kind === SyntaxKind.QualifiedName) {
+        replacementSpan = createTextSpanFromNode(contextToken, sourceFile);
+        insertText = `[${quotePropertyName(sourceFile, preferences, name)}]`;
+    }
```
jakebailey, line 1871: Is it worth doing this if it's guaranteed to be wrong, compared to just not offering this?

Jack-Works: Yes, and I hope QualifiedName can contain `["stringLiteral"]` in the future, since now a type name may not be an identifier.

andrewbranch: What does the error message say when you land in this position? IMO it’s very annoying and confusing for completions to be invalid. It might be ok in this edge case if the diagnostic explains what’s going on and how to fix it, but if it’s just `Module 'foo' has no export "non identifier"`, then I think we need to do better somehow, and I’d lean toward not offering the completion in the meantime.

Jack-Works: Code:

```ts
type z = x.
```

Before this PR:

```ts
type z = x.a b
// Namespace '...' has no exported member 'a'.
// ';' expected.
// Cannot find name 'b'.
```

After this PR:

```ts
type z = x['a b']
// Cannot use namespace 'x' as a type.
```

I think the error message becomes better

andrewbranch: But is that better than not offering the broken completion at all? I’m not sure.

Jack-Works:
ping @andrewbranch please take a look again thanks!

andrewbranch:
My latest feedback stands—I don’t think we should offer invalid completions.

Jack-Works:
> My latest feedback stands—I don’t think we should offer invalid completions.

I still think the completion I gave was better, but I removed that to satisfy the review requirement. Can you review this now, thanks!

andrewbranch (approved):

jakebailey (approved):

andrewbranch:
Thanks @Jack-Works!

PR #58824, opened 2024-06-10
Only use resolutionStart when pushing resolution stack

### LINKED ISSUE #58805: Typescript 5.5 performance regression on arktype
It seems likely it was #58337 that was responsible for the performance hit in https://github.com/arktypeio/arktype

```ts
{
    "checkTime": 9.68,
    "types": 483415,
    "instantiations": 2629152
}

devDependencies:

{
    "checkTime": 12.2,
    "types": 514846,
    "instantiations": 2819487
}
```

(previously incorrectly attributed it to https://github.com/microsoft/TypeScript/pull/58372)

_Originally posted by @ssalbdivad in https://github.com/microsoft/TypeScript/issues/58337#issuecomment-2154832204_

jakebailey:
Fixes #58805

#58337 resets the stack by moving its start up temporarily. However, there are other parts of the checker other than the push/pop circular error code which rely on the stack. It looks like for #58805, `isResolvingReturnTypeOfSignature` in particular stops knowing that a particular return type is being resolved, causing us to start doing extra work that won't go anywhere.

"Fix" this by always using the full stack for these non-circularity-checking stack users.

To make this more clear, the second commit refactors the code a bit into clear functions that show which do and don't use the full stack.

Perhaps this isn't the right fix; this passes all tests, but maybe we shouldn't be using the stack for anything but circularity checking in push/pop anyway. That's a more invasive change.


jakebailey:
@typescript-bot test it
@typescript-bot pack this

DanielRosenwasser:
The perf tests seem to have a decent amount of variance - can you give a clue on what sorts of savings you're seeing specifically on arktype (the report in #58805)?

jakebailey:
Just finished running this:

```
Benchmark 1: 5.4.5
  Time (mean ± σ):     18.920 s ±  0.045 s    [User: 18.532 s, System: 0.300 s]
  Range (min … max):   18.834 s … 18.980 s    20 runs

Benchmark 2: main
  Time (mean ± σ):     20.088 s ±  0.031 s    [User: 19.679 s, System: 0.320 s]
  Range (min … max):   20.043 s … 20.146 s    20 runs

Benchmark 3: PR
  Time (mean ± σ):     19.012 s ±  0.056 s    [User: 18.625 s, System: 0.299 s]
  Range (min … max):   18.822 s … 19.081 s    20 runs

Summary
  5.4.5 ran
    1.00 ± 0.00 times faster than PR
    1.06 ± 0.00 times faster than main
```

DanielRosenwasser (approved):

ssalbdivad:
@jakebailey @DanielRosenwasser Awesome, this seems to totally address the regression on my end:

```ts
5.4.5:
{
    "checkTime": 8.74,
    "types": 479804,
    "instantiations": 2612248
}


5.5.1-rc
{
    "checkTime": 9.92,
    "types": 514330,
    "instantiations": 2812921
}

fix-58805
{
    "checkTime": 8.66,
    "types": 483064,
    "instantiations": 2622889
}
```

Thanks for taking this so seriously. It's ridiculously impressive how well your team integrates real-world perf feedback from across the ecosystem into your dev process.

ahejlsberg:
Hmm, I'm not 100% sure about this one. The whole idea with `resolutionStart` is that we can temporarily suspend some ongoing work and make it appear it isn't actually in process. But with this change it will appear to be in process _sometimes_. We're basically letting a subset of potential ghost errors back in, specifically in cases that depend on `isInResolutionStack`. Those cases aren't too common, but this will definitely make them more unpredictable.

ssalbdivad:
@jakebailey Do we know what led to this change impacting ArkType more than the other repos that were tested?

If it's some obscure pattern that is unlikely to exist in very many places, then identifying it will (hopefully) give me a chance to refactor the affected code.

If it stems from something more ubiquitous, it may be worth weighing the more consistent errors @ahejlsberg mentioned against the perf cost if there is no way to provide both.

jakebailey:
I didn't have time to debug further, but my impression is that there's something very specific to signatures here; the new stack resetting happens as a part of resolving a signature, but this check is checking for a resolving signature, more or less. Makes me think that we're pushing onto the stack then shortly after, resetting it such that we forget that we're resolving. (Which is probably clear given the fact that it works once we look past the new start.)

So, it sort of feels it may be the case that there's some way to avoid that.

jakebailey:
FWIW this difference is measurable with ts-perf (not surprising):

```
┌─────────────┬────────────────────┬────────────────────┬────────────────────┬──────────┬──────────┬─────────────┐
│ Project     │           Baseline │            Current │              Delta │     Best │    Worst │     p-value │
╞═════════════╧════════════════════╧════════════════════╧════════════════════╧══════════╧══════════╧═════════════╡
│ arktype - node (v18.15.0, x64)                                                                                 │
├─────────────┬────────────────────┬────────────────────┬────────────────────┬──────────┬──────────┬─────────────┤
│ Errors      │                  0 │                  0 │                  ~ │        ~ │        ~ │ p=1.000 n=6 │
│ Symbols     │          1,033,514 │          1,058,345 │  +24,831 (+ 2.40%) │        ~ │        ~ │ p=0.001 n=6 │
│ Types       │            477,728 │            504,614 │  +26,886 (+ 5.63%) │        ~ │        ~ │ p=0.001 n=6 │
│ Memory used │ 816,991k (± 0.00%) │ 847,235k (± 0.02%) │ +30,243k (+ 3.70%) │ 847,106k │ 847,534k │ p=0.005 n=6 │
├─────────────┼────────────────────┼────────────────────┼────────────────────┼──────────┼──────────┼─────────────┤
│ Parse Time  │    1.99s (± 0.20%) │    1.96s (± 0.32%) │   -0.03s (- 1.59%) │    1.95s │    1.97s │ p=0.003 n=6 │
│ Bind Time   │    0.82s (± 1.20%) │    0.81s (± 0.50%) │                  ~ │    0.81s │    0.82s │ p=0.213 n=6 │
│ Check Time  │   22.02s (± 0.31%) │   23.43s (± 0.36%) │   +1.40s (+ 6.37%) │   23.34s │   23.58s │ p=0.005 n=6 │
│ Emit Time   │              0.00s │              0.00s │                  ~ │        ~ │        ~ │ p=1.000 n=6 │
├─────────────┼────────────────────┼────────────────────┼────────────────────┼──────────┼──────────┼─────────────┤
│ Total Time  │   24.83s (± 0.29%) │   26.20s (± 0.32%) │   +1.37s (+ 5.50%) │   26.10s │   26.35s │ p=0.004 n=6 │
└─────────────┴────────────────────┴────────────────────┴────────────────────┴──────────┴──────────┴─────────────┘
```

(via https://github.com/microsoft/typescript-benchmarking/pull/73)

ssalbdivad:
Any additional thoughts on either potential plans to address this or ideas about what the culprit may be in [the ArkType repo](https://github.com/arktypeio/arktype) that's causing the performance issues? Is there maybe a broader suite of repos that can be checked for perf changes stemming from this change?

As mentioned, I'm happy to refactor some internal types if need be to accommodate these changes, but I don't have very good intuitions about what code is most likely responsible for the delta.

ssalbdivad:
@DanielRosenwasser Would getting some more data on the broader perf impact of that change be valuable so that it's clear the regression that occurred in ArkType is a clear outlier?

From what I understood, the core fix was already an edge case and Jake's PR (which had negligible overhead) apparently still addresses the majority of those cases. I understand wanting to guarantee ideal behavior as often as possible, but a very noticeable performance hit that could potentially occur in other repos as well seems like a hefty price to pay in this case for that extra bit of certainty.

jakebailey:
I am continuing to look at this; doing some extra tracing of the code involved here to try and compare before/after.

jakebailey:
I've instrumented some basic tracing of the push/pop of signatures and `isResolvingReturnTypeOfSignature`, and it has pretty much confirmed my suspicion; the stack resetting is perfectly placed to just barely cut off the initial push that we're resolving a particular signature.

Here's an example (not "the" example, but one that can fit in one screenshot):

image

On the left is `main`, on the right is with #58337 effectively reverted.

We can see that we `pushTypeResolution () => hasDomain({}, "b")` in `getReturnTypeOfSignature` for the signature for this node. But then, we drop down into `getResolvedSignature`, which resets the stack, cutting off that entry. Later calls to `isResolvingReturnTypeOfSignature` now say that the current signature is _not_ being processed, even though it really is.

This PR works because it allows `isResolvingReturnTypeOfSignature` to still see that resolution is in progress for this signature.

If this PR isn't acceptable, then I'm not totally sure what the right fix will be. I was thinking that maybe the stack could only be reset back up until the signature being resolved, but we're resetting in `getResolvedSignature`, which means that we have a call expression and not the signature (of course), so we can't actually go find the right entry to know where to move the stack start up to.

I also think that using the stack like this for non push-pop circularity checking reasons may itself be the bad thing, but changing that seems challenging.

Maybe the resetting needs to be somewhere other than `getResolvedSignature`?

Any ideas, people smarter than me? 😄  @ahejlsberg

jakebailey:
Here is the trace: https://gist.github.com/jakebailey/6718e643e3399316962bd8edf53f4558

If you're looking for "expensive" calls, I would search the "main.txt" for `\dms$`; these are calls which took more than 50ms, and there are 21 of them on `main`, but only 8 with the revert. That may help find what specific calls could be changed in arktype itself, maybe.

jakebailey:
I am however able to recover maybe 50% of the loss by doing what we do for the variance reset and only doing it once; the comment in the code for the `getResolvedSignature` reset implies that it happens only once, but I don't think it ended up implemented that way.

mikearnaldi:
> @DanielRosenwasser Would getting some more data on the broader perf impact of that change be valuable so that it's clear the regression that occurred in ArkType is a clear outlier?
> 
> From what I understood, the core fix was already an edge case and Jake's PR (which had negligible overhead) apparently still addresses the majority of those cases. I understand wanting to guarantee ideal behavior as often as possible, but a very noticeable performance hit that could potentially occur in other repos as well seems like a hefty price to pay in this case for that extra bit of certainty.

Unfortunately that's not the case, only Anders PR addressed the bug, we tested the other proposed fixes and they did not work except for small isolated cases

Issue #58663, opened 2024-05-26
[5.5 beta] The `moduleDetection` option is ignored


nicolo-ribaudo:
### 🔎 Search Terms

moduleDetection force legacy

### 🕗 Version & Regression Information

- This changed in version 5.5-beta

### 💻 Code

package.json:
```json
{
  "name": "ts-test",
  "version": "1.0.0",
  "description": "",
  "type": "commonjs",
  "devDependencies": {
    "typescript": "^5.5.0-beta"
  }
}
```
tsconfig.json:
```ts
{
  "include": ["src/index.ts"],
  "compilerOptions": {
    "module": "Preserve",
    "verbatimModuleSyntax": true,
    "moduleDetection": "force",
    "emitDeclarationOnly": true,
    "declaration": true,
    "declarationDir": "./dts"
  }
}
```
src/index.ts:
```ts
export let a = 1;
```

### 🙁 Actual behavior

```
src/index.ts:1:1 - error TS1287: A top-level 'export' modifier cannot be used on value declarations in a CommonJS module when 'verbatimModuleSyntax' is enabled.

1 export let a = 1;
```

### 🙂 Expected behavior

Given that I'm using `"moduleDetection": "force"`, that file is an ES module (regardless of what `package.json#type` says) and thus I shouldn't see that error. This was the behavior in 5.4.

jakebailey:
That option doesn't control anything to do with ESM versus CJS, it's controlling "module" (uses some form of exports, declarations are local) versus "script" (all globals, like a script tag).

robpalme:
This sounds similar (but not the same) as the 5.5 issue we hit of ES modules being emitted as CommonJS despite all `tsconfig` options conveying that input and output should all be ESM, e.g. `target: esnext`.

The root cause seems to be [the intentional decision](https://github.com/microsoft/TypeScript/pull/57896) to respect `package.json`'s `"type": "commonjs"` so that it takes precedence over `tsconfig`, i.e. input and output will be assumed to be CJS if the source file is `*.js` and there's no way in `tsconfig` to stop this.

In our case we have a build system that up to now intentionally ignored `package.json` so this coupling was unwelcome, but hey, maybe we just have an odd use-case.  @nicolo-ribaudo what is your use-case for wanting this decoupling of `tsconfig` and `package.json` semantics?

nicolo-ribaudo:
[...]

PR #58825, opened 2024-06-10
Only look up package.json type if module is node16/nodenext or file is in node_modules

### LINKED ISSUE #50647: Typescript [4.8.2] is adding invalid javascript for *.cjs files

### LINKED ISSUE #54752: ESM-only packages come with synthesized default in `moduleResolution: bundler`

andrewbranch:
This walks back the heuristics added to non-Node.js module modes in #57896 a bit, based mostly on feedback in #58663. In 5.5-beta, an explicit package.json `"type"` or module-format-specific file extension started overriding `"module": "commonjs"` or `"module": "esnext"` in tsconfig.json. This impacted both type checking (relevant primarily for imports of `node_modules` dependencies, but with some effects on local files, e.g. with `verbatimModuleSyntax`) and emit (relevant for local project source files, if emit was enabled). The feedback in #58663 was that the effects on local files were unwelcome because the emitted files (whether from tsc or another tool) were going to be placed in some other context tsc didn’t know about, so the package.json tsc was finding during compilation shouldn’t be considered during the build.

With this PR, we only look up package.json `"type"` if `--module` is `node16`/`nodenext` (as we’ve always done) _or_ if the file is inside `node_modules`.

~~Some testing and experimentation is still needed to ensure this doesn’t create inconsistencies with symlinked monorepos.~~ This doesn’t seem to cause inconsistencies with workspaces-style monorepos, because module resolution takes the realpath of files in node_modules before we see if we need to look up a package.json `"type"`. So in a project references example like

```
.
└── packages/
    ├── app/
    │   ├── node_modules/shared ↪︎
    │   ├── tsconfig.json
    │   ├── package.json
    │   └── index.ts
    └── shared/
        ├── dist/index.d.ts
        ├── src/index.ts
        ├── tsconfig.json
        └── package.json
```

When we build `shared/tsconfig.json`, we decide that `shared/src/index.ts` is **not inside node_modules**. Then, when we build `app/tsconfig.json`, even though we resolve `app/node_modules/shared/dist/index.d.ts`, we realpath it to `shared/dist/index.d.ts` and count it as **not inside node_modules**. Having this consistency between input and output files regardless of how they’re resolved is what matters most. The caveat is that the realpath and project references logic can cause the type checking of `app/tsconfig.json` during development to behave differently than it would after publishing/deploying into a situation where `app/node_modules/shared` is no longer a symlink, but a real node_modules package.

Fixes #54752
Fixes #50647


andrewbranch:
@typescript-bot test top400

andrewbranch:
@typescript-bot pack this

andrewbranch:
Perf should be unchanged or better, but

weswigham (commented):
We don't need these new `ModuleSpecifierResolutionHost` members to be public because we automatically attach them or have fallbacks, right? Because these'll be missing from a user-provided `host` object. We probably wannt mark 'em as optional and handle appropriately if they're non-public.

sheetalkamat (approved):

andrewbranch:
> We don't need these new `ModuleSpecifierResolutionHost` members to be public

@weswigham `ModuleSpecifierResolutionHost` itself is internal.

src/compiler/checker.ts
```diff
@@ -47012,10 +47028,22 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+                    moduleKind === ModuleKind.Preserve &&
+                    node.kind !== SyntaxKind.ImportEqualsDeclaration &&
+                    node.kind !== SyntaxKind.VariableDeclaration &&
+                    host.getEmitModuleFormatOfFile(getSourceFileOfNode(node)) === ModuleKind.CommonJS
+                ) {
+                    // In `--module preserve`, ESM input syntax emits ESM output syntax, but there will be times
+                    // when we look at the `impliedNodeFormat` of this file and decide it's CommonJS. To avoid
+                    // that inconsistency, we disallow ESM syntax in files that are unambiguously CommonJS in
+                    // this mode.
+                    error(node, Diagnostics.ESM_syntax_is_not_allowed_in_a_CommonJS_module_when_module_is_set_to_preserve);
```
jakebailey, line 47045: Just to be clera here, this is explicitly about `.cjs` files containing ESM syntax? Or will `.js` files error too when `type=commonjs` is set? The latter seems like it'd be a problem for bun/tsc misc scripts, etc, or projects that set that explicitly for completeness but then bundle.

src/compiler/utilities.ts
```diff
@@ -41,6 +40,7 @@ import {
     canHaveDecorators,
     canHaveLocals,
     canHaveModifiers,
+    type CanHaveModuleSpecifier,
```
jakebailey, line 43:
```suggestion
    CanHaveModuleSpecifier,
```

src/compiler/utilities.ts
```diff
@@ -8685,11 +8705,13 @@ function isFileModuleFromUsingJSXTag(file: SourceFile): Node | undefined {
...
  * in SourceFile construction.
  */
-function isFileForcedToBeModuleByFormat(file: SourceFile): true | undefined {
+function isFileForcedToBeModuleByFormat(file: SourceFile, options: CompilerOptions): true | undefined {
     // Excludes declaration files - they still require an explicit `export {}` or the like
     // for back compat purposes. The only non-declaration files _not_ forced to be a module are `.js` files
     // that aren't esm-mode (meaning not in a `type: module` scope).
-    return (file.impliedNodeFormat === ModuleKind.ESNext || (fileExtensionIsOneOf(file.fileName, [Extension.Cjs, Extension.Cts, Extension.Mjs, Extension.Mts]))) && !file.isDeclarationFile ? true : undefined;
+    //
+    // TODO: extension check never considered compilerOptions; should impliedNodeFormat?
```
jakebailey, line 8713: Is this TODO still relevant?

src/compiler/checker.ts
andrewbranch: It’s now just about `.cjs` files.

jakebailey: Thanks for clarifying!

andrewbranch:
There’s a knip failure but it’s public API?

weswigham:
> ModuleSpecifierResolutionHost itself is internal.

The specifics are internal, yes, but we expose and consume public objects and interfaces that we use as and expect to be `ModuleSpecifierResolutionHost`s which cannot be correctly implemented without these being made public in some way. There's a definite reason almost everything on `ModuleSpecifierResolutionHost` is optional. For example, `Program` has a second (`@internal`) declaration declaring it extends `ModuleSpecifierResolutionHost`. If there are any 3rd-party `Program`s (like our LS or builder construct internally to orchestrate builds, but external), they won't know they need these new methods on their `Program` to satisfy `ModuleSpecifierResolutionHost` uses, and since they're *required*, we'll just crash at runtime when they're inevitably missing.

andrewbranch:
There are 43 non-optional internal methods and properties declared on public `Program` interface declaration—we haven’t made it possible to implement a third-party `Program` up from its public interface.

andrewbranch:
@weswigham any other feedback on the API (or any of the PR)? I’d like to get this in soon.

PR #58829, opened 2024-06-11
Allow `satisfies keyof` assertions in computed property names

### LINKED ISSUE #58800: [isolatedDeclarations] Add a syntactic form of computed property name which is always emitted as a computed property name
### 🔍 Search Terms

isolatedDeclarations transpileModule computed property name

### ✅ Viability Checklist

### ⭐ Suggestion

# Background

Computed property names under `isolatedDeclarations` are very limited right now. Today, you can write `{[Symbol.iterator]: ...}` and that's about it. This restriction is in place because for an arbitrary `{[expression]: ...}` we don't know if the type should be `{[expression]: something}`, `{[expression: string]: something}` or even `{}` (or a future `{f1: something} | {f2: something}`). Computed property names in types today have to exactly be a single late bindable name - nothing more, nothing less - meanwhile computed property names in object expressions (and class declarations) are much more flexible in what we allow.

Thus far, this has worked pretty well for TS users, since we basically pre-solve and cache whatever the expression computed name resolves to into our declaration files. Unfortunately, for `isolatedDeclarations` users, this poses a problem, since the `expression` in the computed property name may be from or rely on type information from another file. In such a case, it's impossible to know how to emit the type for the expression. You could optimistically emit `{[expression]: something}`, but if `expression` ends up evaluating to `string` or `any` in a whole-program context, the declaration file will produce an error and incorrect type information.

# Proposal
What we could use in such a scenario is a syntactic opt-in to guaranteeing the preservation of a computed property name in the calculated type for an expression. A form of computed property name that, when you see it, always ensures a computed property name appears in the output, and issues checker errors if the types when checked cannot produce a valid computed property name in a declaration file.

I propose we reuse some existing syntax with a bit of a new meaning to accomplish this - a `satisfies keyof` postfix assertion, only valid in computed property name positions, and only on dotted entity name expressions. This would mean you could write
```ts
export const a = {
  [something satisfies keyof]: () => {}
}
```
and we would always emit
```ts
export const a: {
  [something]: () => void;
};
```
and issue an error on `something satisfies keyof` if `something` isn't exactly a single unique symbol, string literal, or number literal type (as is valid in the type position computed property name).

# Compatibility

Only `isolatedDeclarations`-concerned authors really need to think about this feature - it's erased from declaration files, since they already check this constraint, so library consumers will never see it. People not using `isolatedDeclarations` will never be driven to use it, since they will always be able to produce a declaration type without an assertion. This is pretty easy to integrate into the `isolatedDeclarations` quickfixer. This doesn't conflict with existing `satisfies keyof T` assertions, since they require a type argument for `keyof`. There is also the possibility of allowing `satisfies keyof` in other locations and on arbitrary expression kinds in the future to check the same invariant - that the expression is exactly a single literal key type - if we think such a check has use in broader contexts than just computed property names.

# Addenda: Making error cases better

Once we have `{[expression satsifies keyof]: ...}` in place, we know that that computed property name should always produce _exactly one object key_, even if `expression` _doesn't_ produce a valid key type (and thus an error). In such a scenario, it could be beneficial to override the type of `expression` with a property key unique to the `expression` symbol, and then fallback to using such a symbol whenever later `obj[expression]` lookups fail. In this way, we can preserve as much user intent as possible, without rapidly reverting to an unchecked `any` state. This is neat (I have a working prototype), especially in the context of single-file checking modes like what our language service does when loading the full program in the background, but isn't really necessary for the feature. The open questions I have for this are just
1. Is it worth supporting this scenario with a special case? and
2. Should the `keyof` result of a type containing one of the fallback property keys be adjusted to be `string | number | symbol`? Should the fallback error property just be filtered from `keyof` entirely?

weswigham:
Fixes #58800

This implements the proposal in the linked issue, along with the addendum about invalid `satisfies keyof` property types still getting unique, singular property names (which ensures declaration file output always includes the computed names, even when they don't resolve).

We can still bikeshed the syntax if anyone *really* wants, but I think this probably fits the bill without actually introducing anything too far off of what we already allow. Though yes, 15 characters is a lot - that's `satisfies` for you.

The open question I think worth answering is if `satisfies keyof` should be allowed in all expression positions and not just in computed property names. It's certainly at least useful as a shorthand for `satisfies (string | number | symbol)` (though the check should be a wee bit stronger than that, since it also checks unit-ness), which has a similar effect on literals to `as const`, without readonly-ing or const-ing array/object types (and instead erroring on them). It's probably worth noting, while generics are *generally* forbidden, given `T extends "a" | "b"`, `T & "a"` simplifies to `"a"` under current rules, so presents as a unit type - making it the *only* generic form, afaik, that can pass the unit-type-key check.


lazytype:
I think logically `satisfies keyof` makes more sense, but aesthetically I think I'd be happy with re-fitting the `deferred` prefix operator introduced in https://github.com/microsoft/TypeScript/pull/58729 if it goes through.
```ts
```

PR #58830, opened 2024-06-11
[experiment] Check resolution stack equality before checking property itself

jakebailey:
Maybe faster?


jakebailey:

jakebailey:
Nope, I guess.

PR #58837, opened 2024-06-12
[DO NOT MERGE UNTIL 5.6] Fix re-exported defaults in ExportInfoMap

### LINKED ISSUE #58590: Wrong import suggestions for JavaScript default exports

- VS Code Version: 1.90.0-insider (a48f464a3e01aad384703ec964018299b14bb7cf)
- OS Version: macOS Sonoma 14.4.1

If an npm package re-exports type definitions from nested paths, VS Code suggests wrong import paths for default exports. I noticed the issue when I tried to use packages from Atlassian's design system (e.g., [`@atlaskit/button`](https://www.npmjs.com/package/@atlaskit/button)), but I was able to reproduce it with a minimal example (see below).

Steps to Reproduce:

1. Clone this issue reproduction repository: https://github.com/klaussner/vs-code-import-issue
2. Follow the steps in `README.md`.

The video below shows how VS Code suggests the correct path for the named export, but the wrong path for the default export. The expected result is `import defaultExport, { namedExport } from "example";`.

https://github.com/microsoft/vscode/assets/1312807/7a85fb0e-9cfb-4d35-a05b-48752b24fa2d

andrewbranch:
Fixes #58590

### DIFF
--- a/tests/baselines/reference/tsserver/fourslashServer/completionsImport_jsModuleExportsAssignment.js
+++ b/tests/baselines/reference/tsserver/fourslashServer/completionsImport_jsModuleExportsAssignment.js
@@ -380,7 +380,7 @@ Info seq  [hh:mm:ss:mss] getCompletionData: Is inside comment: *
 Info seq  [hh:mm:ss:mss] getCompletionData: Get previous token: *
 Info seq  [hh:mm:ss:mss] getExportInfoMap: cache miss or empty; calculating new results
 Info seq  [hh:mm:ss:mss] getExportInfoMap: done in * ms
-Info seq  [hh:mm:ss:mss] collectAutoImports: resolved 0 module specifiers, plus 0 ambient and 3 from cache
+Info seq  [hh:mm:ss:mss] collectAutoImports: resolved 0 module specifiers, plus 0 ambient and 4 from cache
 Info seq  [hh:mm:ss:mss] collectAutoImports: response is incomplete
 Info seq  [hh:mm:ss:mss] collectAutoImports: *
 Info seq  [hh:mm:ss:mss] getCompletionData: Semantic work: *
@@ -1055,6 +1055,19 @@ Info seq  [hh:mm:ss:mss] response:
           },
           {
             "name": "defaults",
+            "kind": "property",
+            "kindModifiers": "",
+            "sortText": "16",
+            "hasAction": true,
+            "source": "/third_party/marked/src/defaults",
+            "data": {
+              "exportName": "export=",
+              "exportMapKey": "8 * defaults ",
+              "fileName": "/third_party/marked/src/defaults.js"
+            }
+          },
+          {
+            "name": "defaults",
             "kind": "alias",
             "kindModifiers": "",
             "sortText": "16",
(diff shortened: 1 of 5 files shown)


tests/cases/fourslash/completionsImport_reExportDefault2.ts
```diff
@@ -0,0 +1,39 @@
...
+
+// @Filename: /node_modules/example/dist/nested/module.d.ts
+//// declare const defaultExport: () => void;
+//// declare const namedExport: () => void;
+//// 
+//// export default defaultExport;
+//// export { namedExport };
+
+// @Filename: /node_modules/example/dist/index.d.ts
+//// export { default, namedExport } from "./nested/module";
```
andrewbranch, line 17: The ExportInfoMap stores chains of re-exports of a given symbol for as long as the symbol has “the same name” in every link in the chain. And by “name,” we mean ones that the user would want to reference a symbol by, not `default` or `export=`. So when we see `default` or `export=`, we try to figure out a better name to offer as the auto-import, by trying to resolve the default export to another named declaration, or making a name out of the filename as a last resort. We do this (among other places) as we’re building the chains of re-exports for the ExportInfoMap, with the goal of being able to answer the question “Give me all the modules from which I can import a given symbol by a given name.” For this test to work, we need to understand that we can import `defaultExport` declared in `example/dist/nested/module.d.ts` both from there and from this re-export right here ☝️. But before this PR, when processing _this_ re-exported `default` symbol, we didn’t search its target file for local names. Instead, we just decided to call it `dist`, after `dist/index.d.ts`, and so it created a new chain in the map instead of being joined to `defaultExport` from `example/dist/nested/module.d.ts`.

tests/baselines/reference/tsserver/fourslashServer/completionsImport_jsModuleExportsAssignment.js
```diff
@@ -1943,6 +1956,19 @@ Info seq  [hh:mm:ss:mss] response:
...
+            "kind": "property",
+            "kindModifiers": "",
+            "sortText": "16",
+            "hasAction": true,
+            "source": "/third_party/marked/src/defaults",
+            "data": {
+              "exportName": "export=",
+              "exportMapKey": "8 * defaults ",
+              "fileName": "/third_party/marked/src/defaults.js"
+            }
```
andrewbranch, line 1970: This was being filtered out by the `!(symbol.flags & SymbolFlags.Transient)` condition in utilities.ts that I removed here. That code path was no longer exercised by the name generation code in the ExportInfoMap construction, but I was curious why it was there, and traced it back to https://github.com/andrewbranch/TypeScript/commit/258be217a6244efcd8055d4096b50585427c5dce to see if it was going to be a problem. So much has changed around this in the last three years that it seems like the crash is no longer an issue, so I removed the transient symbol filtering from the other function where it was happening.

andrewbranch:
@typescript-bot perf test

jakebailey (approved):

sheetalkamat (approved):

andrewbranch:
I don’t think this is very risky, but the RC is already out and the prevalence of the issue seems very low, so I’m going to play it safe and hold off for merging until after the 5.5 stable branch is cut.

andrewbranch:
I’ve just been informed that main is already 5.6!

klaussner:
Thank you, @andrewbranch!

PR #58838, opened 2024-06-12
`tsc -b` removes silent `noEmitOnError` and emits files even if there are errors

sheetalkamat:
On top of #58626

Add hasErrors property to tsbuildInfo so we can build again if there are errors

Also add roots property to non incremental build to check if root files went missing.


sheetalkamat:
@typescript-bot test it

jakebailey:
Oh, I think those test results are actually good, because we're seeing compilation proceed where it couldn't before, right?

jakebailey:
Or, maybe not; the puppeteer case has two tsconfigs for cjs/esm but maybe are getting buildinfo put into the same place? Or something?

sheetalkamat:
> Or, maybe not; the puppeteer case has two tsconfigs for cjs/esm but maybe are getting buildinfo put into the same place? Or something?

I ran this locally and the errors are reported because the dependencies are not installed (i think it just clones and reports errors) but i tried this on main as well as this branch and number of errors are same "110" so i am not sure what bot is reporting?

jakebailey:
The runner definitely does an install; see: https://typescript.visualstudio.com/TypeScript/_build/results?buildId=162186&view=logs&j=e514356d-f816-5eb0-c695-cd1c85d38e8d&t=87388b4c-e53f-56cd-6000-1001b7a7d293 (starting at line 388)

The only weird bit is just the order in which projects are built as far as I can tell.

sheetalkamat:
All right so i reproed it and looked at the diff and it turns out its just order. This is better and correct behavior as the errors will be sorted and reported just like they were with `tsc -p` rather than order of program files. Before this change we jsut got the errors and reported them but with this change we use same machinery as `tsc -p` for reporting errors and all the errors are sorted.

weswigham (approved):
Now that there's a `.buildinfo` present, we can use it to store the error reporting state and don't have to rely solely on the output timestamps to ensure following builds report an error, allowing us to actually emit even when there are errors. Nice. This is probably the last thing that was blocking allowing all CLI options alongside `-b` at the top level, right? Just need to adjust the options parser to allow them as overrides alongside the explicit `-b` options and pass them through to the component projects.

src/compiler/builder.ts
```diff
@@ -1039,7 +983,7 @@ function handleDtsMayChangeOfFileAndExportsOfFile(
 function getSemanticDiagnosticsOfFile(
     state: BuilderProgramStateWithDefinedProgram,
     sourceFile: SourceFile,
-    cancellationToken: CancellationToken | undefined,
+    cancellationToken?: CancellationToken,
```
jakebailey, line 986: Is this intentional? IIRC the token is usually required as to not forget to pass it around.

sheetalkamat:
> Now that there's a `.buildinfo` present, we can use it to store the error reporting state and don't have to rely solely on the output timestamps to ensure following builds report an error, allowing us to actually emit even when there are errors. Nice. This is probably the last thing that was blocking allowing all CLI options alongside `-b` at the top level, right? Just need to adjust the options parser to allow them as overrides alongside the explicit `-b` options and pass them through to the component projects.

Yes thats a next todo on going through them and selecting which ones make sense.

src/compiler/builder.ts
sheetalkamat: Good catch.. that got left out from WIP and fiddling where and how to get the errors

jakebailey (approved):

sheetalkamat:
Just a note that even with this change project is still not built if upstream has errors and i am looking into relaxing that so we can always generate js files and report errors

PR #58845, opened 2024-06-12
Report updateGraph time in async requests (that is getErr) as well as part of request completed event

sheetalkamat:
I think this was over site when we added this to Respose and forgot about "getErr" being async request


sheetalkamat:
@mjbvz I am not sure if we need this or not because as I looked into vscode code, currently this is logged into telemetry for all requests but just for completions.
Not sure if this is interesting and useful or not. This came up when  @gabritto and i were discussing reporting duration of diagnostic event.

gabritto (approved):

PR #58850, opened 2024-06-13
Use monocart for coverage reports, enable codecov

jakebailey:
This is an experimental addition to `c8`, but works really well.

The v8 HTML page it outputs can load checker.ts without batting an eye, and has sub-expression fidelity and can see uncovered elses:

image

image


.c8rc.json
```diff
@@ -1,5 +1,5 @@
 {
-    "reporter": ["lcovonly", "cobertura"],
+    "reporter": ["lcovonly", "cobertura", "v8"],
```
jakebailey, line 2: I don't know how slow it is to always generate this; if it's expensive maybe we should just have a flag to customize the reporters we ask for through c8.

jakebailey:
I'm going to revert the CI and codecov change until we can get faster runners to not double our CI time. The local thing is fine, though, so I'll leave that working. I'll also do a benchmark quick to see how the report collection changes.

jakebailey:
The new runner pool is fast enough to give us coverage without slowing down CI.

.c8rc.json
```diff
@@ -1,5 +1,5 @@
 {
-    "reporter": ["lcovonly", "cobertura"],
+    "reporter": ["lcovonly", "cobertura", "v8", "codecov"],
```
jakebailey, line 2: Checked, and adding another reporter here does not affect the performance of the `report` stage, so we can keep this.

jakebailey:
Marked this PR as ready for review. Now that the CI perf is mitigated, should we just go ahead and enable codecov?

jakebailey:
A caveat is that c8 with monocart takes ~4x longer to run than c8 alone; reported https://github.com/cenfun/monocart-coverage-reports/issues/47 and the profile is somewhat promising for improving this.

RyanCavanaugh (approved):

PR #58854, opened 2024-06-13
Always build project irrespective of errors in dependency with tsc -b

sheetalkamat:


sheetalkamat:
@typescript-bot test it

jakebailey:
This is going to be awesome for benchmarking and extended testing, since we'll actually be able to build build-mode projects without them short circuiting, right?

sheetalkamat:
> This is going to be awesome for benchmarking and extended testing, since we'll actually be able to build build-mode projects without them short circuiting, right?

Yes. While debugging you could just build without having to block on errors

sheetalkamat:
https://github.com/microsoft/TypeScript/pull/58854#issuecomment-2166816302  shows how more packages are being built and errors are reported

weswigham (approved):
Since we don't do project concatenation anymore, there's not even a reason to block when `noEmitOnError` is set, neat.

jakebailey (approved):
So excited for this one.

Are there any issues that this closes? I assume that docs will have to be updated too (surely we document the short circuiting somewhere...).

I also assume that nobody's hard depending on this behavior and we don't need some sort of flag to ask build mode to bail early on failures?

sheetalkamat:
I am hoping we dont need a flag. If the need arises we should evaluate adding the flag at that point.

DanielRosenwasser:
Could that flag be `--noEmitOnError`?

sheetalkamat:
@DanielRosenwasser when i was working on this, i tried that option already and it has no effect, because "d.ts" files from referenced projects will not have errors and blocking project build just because its dependency had error means nothing really. its really whether d.ts file from referenced project has changed or not.

stavalfi:
@sheetalkamat Does this PR means that if project1 depends on project2, if project 2 has an error, I would eventually see compilation errors in both projects?
if yes, this is such an awasome PR!!!!!

sheetalkamat:
> @sheetalkamat Does this PR means that if project1 depends on project2, if project 2 has an error, I would eventually see compilation errors in both projects? if yes, this is such an awasome PR!!!!!

Yes

PR #58831, opened 2024-06-11
Disable #57896 (Respect package.json "type" and module-format-specific file extensions in more module modes)

andrewbranch:
This disables the behavior changes introduced in #57896, based on feedback in https://github.com/microsoft/TypeScript/issues/58663 and #58523 and feedback from @bpasero about effects on VS Code’s file watchers. The hope is to merge this for 5.5 stable, then merge #58825 (pending review), which will reintroduce the behavior in the original PR in a more limited way, at the beginning of the 5.6 period.

PR #58857, opened 2024-06-13
Revert #57896

andrewbranch:
My attempt (#58831) to just disable #57896 was incomplete, so this is the full revert instead.


jakebailey:
Does this also undo Wesley's change? Or does that still work?

andrewbranch:
The only interaction between my work and #58176 is how often package.jsons that one of us is doing a lookup for is already cached because of a lookup that the other is doing.

andrewbranch:

jakebailey (approved):
😢

PR #58859, opened 2024-06-13
When reporting errors, compute diagnostics even if type relation has cached result

### LINKED ISSUE #3276: Diagnostic elaboration is not consistently reported in language service

gabritto:
Updated version of #55234.
Fixes #3276.

Motivated by #57842.


gabritto:
@typescript-bot test it

gabritto:
It doesn't seem like the case of reporting the same assignability error more than once is that frequent, since there were only 3 changes in user tests, none in top 400, and no perf impact.

Two things I want to do, though:
- Special case type comparison "overflow" cases so that we don't do the comparison again.
- Maybe only have this consistent assignability elaboration when we know we are in an editor scenario, if that's possible.

gabritto:
@typescript-bot test top800

gabritto:
@typescript-bot test this

gabritto:

weswigham (approved):
Perf still looks fine, baselines look good, and I already approved the old version of the PR :)

PR #58862, opened 2024-06-14
[DO NOT MERGE] Test reverting 58337, to check extended test

jakebailey:
Not proposing this at all, but want to see if our user test suite catches this as a regression given Effect is in there.


jakebailey:

jakebailey:
Dang, no Effect repro. I guess that makes sense given it was an LS inconsistency.

PR #58864, opened 2024-06-14
Deprecate `.hasRestElement` in favor of `combinedFlags`

Andarist:
Just putting this up for discussion, @weswigham has expressed an interest in this [here](https://github.com/microsoft/TypeScript/pull/58708#discussion_r1640172609) 😉

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -18565,7 +18565,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             }
             if (everyType(objectType, isTupleType) && isNumericLiteralName(propName)) {
                 const index = +propName;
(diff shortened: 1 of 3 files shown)


weswigham (approved):
I'm down for this deprecation (since I proposed it) - @jakebailey ? @ahejlsberg ? @sandersn ? Y'all seem like you'd have the most interest in the tuple element API.

jakebailey:
I think it's fine, though the name being "Variable" makes me quite confused.

I would also say that we should drop the prop entirely and then strap it back on with a getter in the deprecation helpers, but that sounds pretty hard to do for Types.

sandersn:
> though the name being "Variable" makes me quite confused.
The enum is full of so many overlapping concepts that `Variable` is a pretty good choice. The only change I can think of is 'VariadicOrRest', which isn't really an improvement.

sandersn (approved):

sandersn:
@typescript-bot test it

sandersn:
The only perf test diff is in the binder for mui-docs, which seems unrelated.

Issue #58319, opened 2024-04-25
Watcher recursively watches irrelevant directories


dmichon-msft:
### 🔎 Search Terms

watch, excess, files, directories

### 🕗 Version & Regression Information

- This changed between versions 5.0.4 and 5.1.3, still present in typescript@next

### 💻 Code

[Repro](https://github.com/dmichon-msft/typescript-watch-repo)

Steps:
```
npm install -g pnpm
pnpm -r install
pnpm -r build
cd b/2/b-impl/b
./node_modules/.bin/tsc --watch --extendedDiagnostics
```
Once the build finishes, create a file `tmp.txt` in `a/2/logs` and observe that TypeScript acts on the file change.

### 🙁 Actual behavior

If a folder is added to TypeScript's list of "failed lookup locations", any file change at any level within said folder will trigger a rebuild, even if the change is in a nested folder that was not probed during module resolution.

### 🙂 Expected behavior

TypeScript only watches contents of directories that contain files that were referenced in a tsconfig or probed during module resolution.

### Additional information about the issue

Environment: GitHub Codespaces
OS: #19~22.04.1-Ubuntu SMP Wed Jan 10 22:57:03 UTC 2024
node 18.20.2

dmichon-msft:
Looking further, this seems to be triggered by TypeScript unexpectedly trying to resolve an import of `./something` from an `index.d.ts` file to a `./something.ts` file instead of going directly to `./something.d.ts`.

dmichon-msft:
When the offending call to recursively watch a common root directory occurs, the `resolution` object contains a valid `resolvedModule` entry with a `.d.ts` extension. It also contains entries under `failedLookupLocations` for the same file with a `.ts` and a `.tsx` extension.

dmichon-msft:
Issue updated with minimal repro.

sheetalkamat:
This is intentional as otherwise many watchers are created. for different resolutions and we want to avoid that. The watcher is smart to use lot of stuff so even if it triggers rebuild it will not do anything major.

RyanCavanaugh:
@dmichon-msft do you have a reproduction of the infinite build loop you mentioned on Teams?

dmichon-msft:
> This is intentional as otherwise many watchers are created. for different resolutions and we want to avoid that. The watcher is smart to use lot of stuff so even if it triggers rebuild it will not do anything major.
[...]

PR #58866, opened 2024-06-14
On linux or editor with canUseEvents to prefer immediate directory if its not in root or node_modules

sheetalkamat:
Experiment with what to watch with `preferNonRecursiveWatch` option on the host

Helps with #58319


sheetalkamat:
cc: @dmichon-msft

sheetalkamat:
@typescript-bot pack this

dmichon-msft:
I'm working on pulling this down to test in my environment. I'll note that I find it useful to also have a toggle to disable watching any path that contains a `node_modules` element, since modifying the set of installed packages during a watch mode session is already not allowed by the processes in my repository (and I think this is fairly common for monorepos in general). I can approximate the latter with the new `${configDir}` functionality though it gets tedious since the relative path depends on the project depth.

dmichon-msft:
Ah, right, I can't test this directly in my target repository, need to experiment with consuming it from  heft-typescript-plugin and passing the necessary config option first. That'll take a bit longer to set up.

dmichon-msft:
Happy to see this, but it looks like I need to do some work to be compatible with the various changes that shipped in TypeScript 5.5 before I'll be able to test it E2E.

sheetalkamat:
@typescript-bot test it

tests/baselines/reference/api/typescript.d.ts
```diff
@@ -2651,6 +2651,7 @@ declare namespace ts {
         interface ServerHost extends System {
             watchFile(path: string, callback: FileWatcherCallback, pollingInterval?: number, options?: WatchOptions): FileWatcher;
             watchDirectory(path: string, callback: DirectoryWatcherCallback, recursive?: boolean, options?: WatchOptions): FileWatcher;
+            preferNonRecursiveWatch?: boolean;
```
jakebailey, line 2654: Is the intent that downstream users configure this? Do we need to wait for David to check that this helps?

sheetalkamat: I was not sure whether to leave this public on ServerHost or not but TscWatch yes because i think last i checked david was using the API to patch tsc watch.  

I am looking to get this in early and revert or reiterate sooner than later as we dont get more feedback without this being in nightly esp since this depends on OS and vscode setting to use their watchers

sheetalkamat: i can make these internal on tsc-watch and tsserver internal and make them public in next PR if they are needed if that approach is better

jakebailey: Understood; just checking; I think you have more context than me on this.

jakebailey (approved):

Issue #58863, opened 2024-06-14
[nightly][performance] 97% of time spent reading type dependencies when --noCheck and --isolatedDeclarations is enabled


MichaelMitchell-at:
### 🔎 Search Terms

noCheck isolatedDeclarations

### 🕗 Version & Regression Information

- --noCheck was just added in https://github.com/microsoft/TypeScript/pull/58839 Thanks @sheetalkamat!

### 💻 Code

```ts
// @strict: true
// @declaration: true
// @isolatedDeclarations: true
// @allowJs: false
// @noCheck: true

// @filename: index.ts
export { type Falsey } from "lodash";
```

### 🙁 Actual behavior

Ran the above code using `tsc --generateTrace trace_dir` which produced the following trace.
[trace.json](https://github.com/user-attachments/files/15842388/trace.json)

If you load this trace into https://speedscope.app you can see that out of 271ms, about 97% of the time is spent parsing and binding `@types/lodash` and `lib.d.ts` and only 6ms is spent on the actual emit.

which seems superfluous and a huge optimization opportunity if we can skip these steps.  cc @weswigham @jakebailey @dragomirtitian 

### 🙂 Expected behavior

No time is spent parsing and binding and emit takes up most of the time

jakebailey:
Well, it's `noCheck`, not `noBind` or `noParse`; unless using one of the transpile functions (which never even hits the FS), all of the same up-front work is going to be done to build the Program before anything else is done.

Maybe `noCheck` could effectively disable any and all program construction besides what's listed explicitly in `includes`/`files` (so only actually emittable files will be parsed and bound), but I don't know how that would affect `buildinfo`'s contents.

MichaelMitchell-at:
I don't think I even have a mental model around how `noCheck` ought to behave in an editor context, but regardless of that I don't see why not parsing "@types/lodash" would break the editor.

jakebailey:
`noCheck` means "don't ask the checker to walk the file and collect diagnostics", but enabling it notably _doesn't_ break the editor for completion, hovers, nor the emitter for type-using emit (i.e. code that does not pass isolatedModules or isolatedDeclarations). Notably, you can have files locally that depend packages in `node_modules` for emit, and they need to be able to emit as expected too.

Doing better here would of course be nice, just needs some thinking! For code that passes `isolatedModules` and `isolatedDeclarations`, skipping a full project load _should_ be safe?

sheetalkamat:
[...]

Issue #57524, opened 2024-02-24
lib.dom.d.ts seems to not be referenced when any file contains no-default-lib=true triple slash


JakeYallop:
### 🔎 Search Terms

no-default-lib lib dom ignored cannot resolve HTMLElement

### 🕗 Version & Regression Information

This is an unintuitive error that took me a long time to pin down.

### 💻 Code

Given the following 3 files:

worker.d.ts
```typescript
/// <reference no-default-lib="true"/>
/// <reference lib="webworker" />
/// <reference lib="esnext" />

export declare class MyWorker {}
```

library.d.ts
```typescript
import type { MyWorker } from "./worker";

export interface Type1 {
  worker: MyWorker;
}
```

app.d.ts
```typescript
interface App {
  prop: HTMLElement;
}
```

tsconfig.json
```json
{
  "compilerOptions": {
    "lib": [ "ES2020", "DOM" ],
  },
}
```

### 🙁 Actual behavior

An error occurs in the app.d.ts file
```
Cannot find name 'HTMLElement'.ts(2304)
```

### 🙂 Expected behavior

Any of the following:
* A better error message inside the app.d.ts file
* an error in the tsconfig
* a better error message when using `--traceResolution`

Suggesting a reason for why `lib: [ "DOM" ]` in the tsconfig file is ignored.

Its also possible this is incorrect behaviour, however [the docs say](https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html) document that the use case for no-default-lib is for files that do not include the default library, rather than those (like in my case) that wanted to ignore the default libraries included via the tsconfig.json.

### Additional information about the issue

This bug is really here as a helpful hint in case anyone else runs into this same issue. Its particularly difficult to debug the issue, as app.d.ts has no references to anything. In my real project, the situation is a bit more complicated, and it took me a long time to figure out that this was the issue.

As a side note, what is the best way to do typings for a web worker file within an existing project? At the moment, I'll be adding a separate tsconfig file for the specific web worker, and exclude that file from the main tsconfig,

RyanCavanaugh:
I don't think it was intentional that this directive causes an explicitly-set `lib` setting to be ignored. @sheetalkamat ?

sheetalkamat:
I think this was intentional per https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html#-reference-no-default-libtrue  means it is as if you passed `noLib` atleast per documentation.

JakeYallop:
[...]

PR #58867, opened 2024-06-14
Fix incorrect condition of noLib

sheetalkamat:


MichaelMitchell-at:
Verified this fixes the issue described in https://github.com/microsoft/TypeScript/issues/58863#issuecomment-2168784178 ! Thank you!

jakebailey (approved):

jakebailey:
We should have run the extended tests on this; it this causes a crash in the vscode codebase, and also causes some 100 errors there too.

jakebailey:
I had to revert this in #58935; it caused compilation of vscode to crash when `--diagnostics` was passed, and some 100+ missing lib errors.

sheetalkamat:
> I had to revert this in #58935; it caused compilation of vscode to crash when `--diagnostics` was passed, and some 100+ missing lib errors.

The errors dont seem to be because of this PR. I ran `tsc -p src` with and without fix and the number of errors didnt change.
I do see the crash and looking into whats going on. But this fix is partially correct.

jakebailey:
Oops, you're totally right. I think I was incorrectly comparing against another baseline. That being said, I don't know why they have the errors at all, https://github.com/microsoft/vscode/blob/main/src/tsconfig.base.json#L23 says ES2022 so we should also be pulling in ES2021.

sheetalkamat:
Its because one of their file `c:/temp/vscode/src/vs/workbench/contrib/webview/browser/pre/service-worker.js` has `no-default-lib` is true. I will revisit #57524 and come up with some proposal but thats the cause.

PR #58869, opened 2024-06-14
Reject internal tag on private decls, strip comments from private decls in dtsBundler

jakebailey:
As exposed by #58187, private declarations do not compose with `@internal` when `stripInternal` is done via as separate tool.

Keeping the comments from private properties is undesirable as it leaks info to the declaration files that is meaningless without type info.

But, if we remove comments from private properties, and those private properties are marked as `@internal`, that info is removed too. So, anyone post-processing dts files to do stripping similar to `stripInternal` will break.

It seems to me that we just shouldn't tag anything already private as `@internal`. That will mean that we include more private properties in our declaration files, but this is nothing new, it's just that a small number of props got both `private` and `@internal`, which feels unintended.

So, this PR also uses `removeAllComments` in `dtsBundler.mjs` to drop all private comments too, at least until #58187 is out and has the same effect.

### DIFF
--- a/scripts/dtsBundler.mjs
+++ b/scripts/dtsBundler.mjs
@@ -40,6 +40,13 @@ const newLineKind = ts.NewLineKind.LineFeed;
 const newLine = newLineKind === ts.NewLineKind.LineFeed ? "\n" : "\r\n";
 
 /**
+ * @param {ts.Node} node
+ */
+function removeAllComments(node) {
+    /** @type {any} */ (ts).removeAllComments(node);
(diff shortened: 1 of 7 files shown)


sheetalkamat (approved):
This is ok but i think we probably should change bot to see if the diff with API is only private addition its not really protocol or Public API change to reduce the noise

jakebailey:
Yeah, outputting a baseline-only dts file which omits private props sounds like a good idea.

PR #58870, opened 2024-06-14
Specialize the message on JSX tags looking for the JSX factory namespace

DanielRosenwasser:


gabritto:
Should we do something similar for the error message in https://github.dev/microsoft/TypeScript/blob/b1c52c53cc6c8cf35b19accfc3e29916489c821e/src/compiler/checker.ts#L33073?

DanielRosenwasser:
Yes, probably!

DanielRosenwasser:
@gabritto done!

src/compiler/diagnosticMessages.json
```diff
@@ -3907,6 +3907,14 @@
         "category": "Error",
         "code": 2868
     },
+    "This JSX tag requires '{0}' to be in scope, but it could not be found.": {
+        "category": "Error",
+        "code": 2869
+    },
+    "This JSX tag requires the module path '{0}' to exist, but none could be found. Make sure you have the appropriate package installed.": {
```
weswigham, line 3914:
```suggestion
    "This JSX tag requires the module path '{0}' to exist, but none could be found. Make sure you have types for the appropriate package installed.": {

```
? Since it's perfectly possible (common, even) to have `react` installed, but `@types/react` is still missing. Or is that a different error?

DanielRosenwasser: Whether there's a separate error, I'm not sure about - but good call on specializing the message.

gabritto (approved):

PR #58873, opened 2024-06-15
Use more explicit operations in core helpers (and other nits)

DanielRosenwasser:
This PR takes care of a few nits in the codebase, including:

* Using `??` over `||` when possible.
* Using numeric `for` loops in `core.ts` for all iteration over arrays.
* Using explicit checks for `undefined` on `core.ts` helpers.
* Use compound assignment operators when possible.

Many of these are rooted in the need for core utilities to be faster, and therefore use fewer implicit operations like iterator usage, truthiness, etc.

There are some other things I noticed, but this seemed like a reasonable first pass.

### DIFF
--- a/src/compiler/core.ts
+++ b/src/compiler/core.ts
@@ -839,7 +839,7 @@ export function arrayIsSorted<T>(array: readonly T[], comparer: Comparer<T>) {
 }
 
 /** @internal */
-export function arrayIsEqualTo<T>(array1: readonly T[] | undefined, array2: readonly T[] | undefined, equalityComparer: (a: T, b: T, index: number) => boolean = equateValues): boolean {
+export function arraysEqual<T>(array1: readonly T[] | undefined, array2: readonly T[] | undefined, equalityComparer: (a: T, b: T, index: number) => boolean = equateValues): boolean {
     if (array1 === undefined || array2 === undefined) {
         return array1 === array2;
     }
--- a/src/compiler/core.ts
+++ b/src/compiler/core.ts
@@ -538,30 +538,6 @@ export function* mapDefinedIterator<T, U>(iter: Iterable<T>, mapFn: (x: T) => U 
 }
 
 /** @internal */
-export function mapDefinedEntries<K1, V1, K2, V2>(map: ReadonlyMap<K1, V1>, f: (key: K1, value: V1) => readonly [K2, V2] | undefined): Map<K2, V2>;
-/** @internal */
-export function mapDefinedEntries<K1, V1, K2, V2>(map: ReadonlyMap<K1, V1> | undefined, f: (key: K1, value: V1) => readonly [K2 | undefined, V2 | undefined] | undefined): Map<K2, V2> | undefined;
-/** @internal */
-export function mapDefinedEntries<K1, V1, K2, V2>(map: ReadonlyMap<K1, V1> | undefined, f: (key: K1, value: V1) => readonly [K2 | undefined, V2 | undefined] | undefined): Map<K2, V2> | undefined {
-    if (!map) {
-        return undefined;
-    }
-
-    const result = new Map<K2, V2>();
-    map.forEach((value, key) => {
-        const entry = f(key, value);
-        if (entry !== undefined) {
-            const [newKey, newValue] = entry;
-            if (newKey !== undefined && newValue !== undefined) {
-                result.set(newKey, newValue);
-            }
-        }
-    });
-
-    return result;
-}
-
-/** @internal */
 export function getOrUpdate<K, V>(map: Map<K, V>, key: K, callback: () => V) {
     if (map.has(key)) {
         return map.get(key)!;
--- a/src/compiler/core.ts
+++ b/src/compiler/core.ts
@@ -2044,7 +2044,13 @@ export function compose<T>(...args: ((t: T) => T)[]): (t: T) => T;
 /** @internal */
 export function compose<T>(a: (t: T) => T, b: (t: T) => T, c: (t: T) => T, d: (t: T) => T, e: (t: T) => T): (t: T) => T {
     if (!!e) {
-        return t => e(d(c(b(a(t)))));
+        const args: ((t: T) => T)[] = [];
+        for (let i = 0; i < arguments.length; i++) {
+            // eslint-disable-next-line prefer-rest-params
+            args[i] = arguments[i];
+        }
+
+        return t => reduceLeft(args, (u, f) => f(u), t);
     }
     else if (d) {
         return t => d(c(b(a(t))));
--- a/src/compiler/core.ts
+++ b/src/compiler/core.ts
@@ -682,7 +682,7 @@ export function concatenate<T>(array1: readonly T[], array2: readonly T[]): read
 export function concatenate<T>(array1: T[], array2: T[] | undefined): T[]; // eslint-disable-line @typescript-eslint/unified-signatures
 /** @internal */
 export function concatenate<T>(array1: T[] | undefined, array2: T[]): T[]; // eslint-disable-line @typescript-eslint/unified-signatures
-/** @internal *
+/** @internal */
 export function concatenate<T>(array1: readonly T[], array2: readonly T[] | undefined): readonly T[]; // eslint-disable-line @typescript-eslint/unified-signatures
 /** @internal */
 export function concatenate<T>(array1: readonly T[] | undefined, array2: readonly T[]): readonly T[]; // eslint-disable-line @typescript-eslint/unified-signatures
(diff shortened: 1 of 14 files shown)


DanielRosenwasser:

jakebailey:
@typescript-bot perf test this faster predictable host=node@22.3.0

src/compiler/core.ts
```diff
@@ -534,30 +537,6 @@ export function* mapDefinedIterator<T, U>(iter: Iterable<T>, mapFn: (x: T) => U
     }
 }
 
-/** @internal */
-export function mapDefinedEntries<K1, V1, K2, V2>(map: ReadonlyMap<K1, V1>, f: (key: K1, value: V1) => readonly [K2, V2] | undefined): Map<K2, V2>;
```
jakebailey, line 538: Any reason to delete this unused func but not other unused ones? Not that it much matters.

src/compiler/core.ts
```diff
@@ -703,7 +682,7 @@ export function concatenate<T>(array1: readonly T[], array2: readonly T[]): read
 export function concatenate<T>(array1: T[], array2: T[] | undefined): T[]; // eslint-disable-line @typescript-eslint/unified-signatures
 /** @internal */
 export function concatenate<T>(array1: T[] | undefined, array2: T[]): T[]; // eslint-disable-line @typescript-eslint/unified-signatures
-/** @internal */
+/** @internal *
```
jakebailey, line 685: The heck is happening here

DanielRosenwasser: I dunno how that happened, but it is pretty funny that this is never used.

src/compiler/core.ts
DanielRosenwasser: What other unused functions are there? I saw that it only has a single use-site and could easily be substituted with `new Set(mapDefinedIterator(...))` where it was used.

DanielRosenwasser:

src/compiler/core.ts
jakebailey: `compose` is unused IIRC. It has gone away and come back though.

src/compiler/core.ts
```diff
@@ -1779,6 +1752,8 @@ export function createSet<TElement, THash = number>(getHashCode: (element: TElem
             return size;
         },
         forEach(action: (value: TElement, key: TElement, set: Set<TElement>) => void): void {
+            // NOTE: arrayFrom means that if the callback mutates the underlying collection,
+            //       we won't have an accurate set of values
```
jakebailey, line 1756:
```suggestion
            // NOTE: arrayFrom prevents `action` from modifying the map while we iterate
```

src/compiler/debug.ts
```diff
@@ -1126,7 +1126,7 @@ m2: ${(this.mapper2 as unknown as DebugTypeMapper).__debugToString().split("\n")
 
         function renderGraph() {
             const columnCount = columnWidths.length;
-            const laneCount = nodes.reduce((x, n) => Math.max(x, n.lane), 0) + 1;
+            const laneCount = ts.maxBy(nodes, 0, n => n.lane) + 1;
```
jakebailey, line 1129: Do the import here rather than using `ts`; it's only in scope because https://github.com/microsoft/TypeScript/issues/51530

src/compiler/core.ts
DanielRosenwasser: To be honest, I'm not sure if the copy was intentional.

jakebailey: Nothing fails when it's removed, huh

src/compiler/core.ts
```diff
@@ -2145,6 +2120,14 @@ export function compareTextSpans(a: Partial<TextSpan> | undefined, b: Partial<Te
     return compareValues(a?.start, b?.start) || compareValues(a?.length, b?.length);
 }
 
+/** @internal */
+export function maxBy<T>(arr: readonly T[], init: number, mapper: (x: T) => number): number {
```
DanielRosenwasser, line 2124: I guess this could always be a `reduceMapped` instead.

src/compiler/core.ts
DanielRosenwasser: Neither behavior is really easy to make guarantees about without extra book-keeping.

Right now if you add a new value and it's added into an array under the hood, you *will* witness a new value if you're in the middle of iterating, but not if you've already finished iterating over all values with a given key.

On the other hand, if the multimap stores a single value, you'll never see any added values as you iterate.

jakebailey (approved):

src/compiler/core.ts
```diff
@@ -234,24 +236,19 @@ export function findMap<T, U>(array: readonly T[], callback: (element: T, index:
...
     return false;
 }
 
-/** @internal */
-export function arraysEqual<T>(a: readonly T[], b: readonly T[], equalityComparer: EqualityComparer<T> = equateValues): boolean {
-    return a.length === b.length && a.every((x, i) => equalityComparer(x, b[i]));
-}
-
 /** @internal */
 export function indexOfAnyCharCode(text: string, charCodes: readonly number[], start?: number): number {
```
dmichon-msft, line 250: If this function is ever called in a hot path it would probably be more efficient to convert `charCodes` to a `Set` first to reduce the `O(text*codes)` to `O(text)`

src/compiler/core.ts
```diff
@@ -713,8 +687,8 @@ export function concatenate<T>(array1: T[] | undefined, array2: T[] | undefined)
 export function concatenate<T>(array1: readonly T[] | undefined, array2: readonly T[] | undefined): readonly T[] | undefined;
 /** @internal */
 export function concatenate<T>(array1: readonly T[] | undefined, array2: readonly T[] | undefined): readonly T[] | undefined {
-    if (!some(array2)) return array1;
-    if (!some(array1)) return array2;
+    if (array2 === undefined || array2.length === 0) return array1;
+    if (array1 === undefined || array1.length === 0) return array2;
     return [...array1, ...array2];
```
dmichon-msft, line 692: if indexed array access is significantly faster via avoiding iterator semantics, using `array1.concat(array2)` is probably faster for the same reason.

src/compiler/core.ts
```diff
@@ -899,20 +873,20 @@ export function compact<T>(array: readonly T[]): readonly T[]; // eslint-disable
...
-                    result = array.slice(0, i);
-                }
+            // Either the result has been initialized (and is looking to collect truthy values separately),
+            // or we've hit our first falsy value and need to copy over the current stretch of truthy values.
+            if (result ?? !v) {
+                result ??= array.slice(0, i);
                 if (v) {
                     result.push(v);
                 }
             }
```
dmichon-msft, line 886: This might be more readable as:
```suggestion
            if (v) {
              // If result has been initialized it is looking to collect truthy values separately.
              // Otherwise it will be initialized when we hit our first falsy value.
              result?.push(v);
            } else {
              // If this is our first falsy value, copy over the current stretch of truthy values.
              result ??= array.slice(0, i);
            }
```

src/compiler/core.ts
```diff
@@ -1705,14 +1683,9 @@ export function createSet<TElement, THash = number>(getHashCode: (element: TElem
             const hash = getHashCode(element);
             if (!multiMap.has(hash)) return false;
             const candidates = multiMap.get(hash)!;
```
dmichon-msft, line 1685: Does this utility function expressly support `undefined` being a meaningful value in the set? Otherwise this would be more efficient to call `get` and do a check against `undefined` instead of calling into the map twice.

src/compiler/watch.ts
```diff
@@ -305,7 +306,7 @@ function createTabularErrorsDisplay(filesInError: (ReportFileInError | undefined
 
     const numberLength = (num: number) => Math.log(num) * Math.LOG10E + 1;
     const fileToErrorCount = distinctFiles.map(file => ([file, countWhere(filesInError, fileInError => fileInError!.fileName === file!.fileName)] as const));
-    const maxErrors = fileToErrorCount.reduce((acc, value) => Math.max(acc, value[1] || 0), 0);
+    const maxErrors = maxBy(fileToErrorCount, 0, value => value[1]);
 
     const headerRow = Diagnostics.Errors_Files.message;
     const leftColumnHeadingLength = headerRow.split(" ")[0].length;
```
dmichon-msft, line 312:
```suggestion
const leftColumnFirstSpace = headerRow.indexOf(" ");
const leftColumnHeadingLength = leftColumnFirstSpace < 0 ? headerRow.length : leftColumnFirstSpace;
```

Though I assume this line isn't a hot path.

src/compiler/core.ts
jakebailey: #58946

Issue #58833, opened 2024-06-12
Nested generic calls don't inherit outer inferences when their return types are intersections containing functions


Andarist:
### 🔎 Search Terms

generic call inference intersection return type

### 🕗 Version & Regression Information

- This is the behavior in every version I tried

### ⏯ Playground Link

https://www.typescriptlang.org/play/?ts=5.6.0-dev.20240611#code/JYOwLgpgTgZghgYwgAgMIHsQ2AcwDwAqAfMgN4CwAUMsgppAB5gBcyBA3FTaAG7oDWEVhWo1kACgasCASlZ9gAE06iaEBhAStJ0mcgC8JBcq7IAvirNUqizQBs4UFDACuIBGGCZaTuJEJE4nRYuKwYIfjEcsjGKjb2js5uHl4gyAC2AJ4AkiB8ggBCUAIQIAHiwMLI6praUmx6hjHoSubRIjQ6DfItJpRmyABkyGCZAA4Q6DDIwHGUCL6Q4h209OosZKtuGwDM5gA0prwlrFm5+RBFJSDLpmoaWhIMjSQrYsgMd2IA9N-IAHoAfi+NAYADo6NswWB0AAxYAaRTiABMMhU7zMh1EZhkWJxc1sCAcTmQrncnm8ZzyJXKlU2NUeXVkBiMvTawlMTOisVMDLquhZzSUljmCwgfggt1EwUYG1IW3ArD2mKO1MEpxyaslb2qDzqLzIII+Rt+AOBqjE4Mh4GhcIRECRqPRYhV2NxVHxVCAA

### 💻 Code

```ts
interface Config<T> {
  context: T;
  invoke: {
    (x: T): void;
    exec: (x: T) => void;
  };
}

declare function create<T>(config: Config<T>): void;

declare function myInvokeBroken<T>(i: { exec: (x: T) => void }): {
  (x: T): void;
} & typeof i;

create({
  context: { count: 3 },
  invoke: myInvokeBroken({
    exec: (x) => {
      x
      // ^?
      x.count.toFixed(2);
    },
  }),
});

declare function myInvoke<T>(i: { exec: (x: T) => void }): {
  (x: T): void;
  exec: (x: T) => void;
};

create({
  context: { count: 3 },
  invoke: myInvoke({
    exec: (x) => {
      x
      // ^?
      x.count.toFixed(2);
    },
  }),
});
```

### 🙁 Actual behavior

[...]

PR #58875, opened 2024-06-15
Defer processing generic functions returning intersected functions

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/58833

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -36090,7 +36090,11 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     }
 
     function isGenericFunctionReturningFunction(signature: Signature) {
-        return !!(signature.typeParameters && isFunctionType(getReturnTypeOfSignature(signature)));
+        if (!signature.typeParameters) {
+            return false;
+        }
+        const returnType = getReturnTypeOfSignature(signature);
+        return returnType.flags & TypeFlags.Intersection ? some((returnType as IntersectionType).types, isFunctionType) : isFunctionType(returnType);
     }
 
     /**
--- a/tests/baselines/reference/inferFromGenericFunctionReturnTypes4.symbols
+++ b/tests/baselines/reference/inferFromGenericFunctionReturnTypes4.symbols
@@ -1,0 +1,121 @@
+//// [tests/cases/compiler/inferFromGenericFunctionReturnTypes4.ts] ////
+
+=== inferFromGenericFunctionReturnTypes4.ts ===
+// https://github.com/microsoft/TypeScript/issues/58833
+
+interface Config<T> {
+>Config : Symbol(Config, Decl(inferFromGenericFunctionReturnTypes4.ts, 0, 0))
+>T : Symbol(T, Decl(inferFromGenericFunctionReturnTypes4.ts, 2, 17))
+
+  context: T;
+>context : Symbol(Config.context, Decl(inferFromGenericFunctionReturnTypes4.ts, 2, 21))
+>T : Symbol(T, Decl(inferFromGenericFunctionReturnTypes4.ts, 2, 17))
+
+  invoke: {
+>invoke : Symbol(Config.invoke, Decl(inferFromGenericFunctionReturnTypes4.ts, 3, 13))
+
+    (x: T): void;
+>x : Symbol(x, Decl(inferFromGenericFunctionReturnTypes4.ts, 5, 5))
+>T : Symbol(T, Decl(inferFromGenericFunctionReturnTypes4.ts, 2, 17))
+
+    exec: (x: T) => void;
+>exec : Symbol(exec, Decl(inferFromGenericFunctionReturnTypes4.ts, 5, 17))
+>x : Symbol(x, Decl(inferFromGenericFunctionReturnTypes4.ts, 6, 11))
+>T : Symbol(T, Decl(inferFromGenericFunctionReturnTypes4.ts, 2, 17))
+
+  };
+}
+
+declare function create<T>(config: Config<T>): void;
+>create : Symbol(create, Decl(inferFromGenericFunctionReturnTypes4.ts, 8, 1))
+>T : Symbol(T, Decl(inferFromGenericFunctionReturnTypes4.ts, 10, 24))
+>config : Symbol(config, Decl(inferFromGenericFunctionReturnTypes4.ts, 10, 27))
+>Config : Symbol(Config, Decl(inferFromGenericFunctionReturnTypes4.ts, 0, 0))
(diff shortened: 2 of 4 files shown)


src/compiler/checker.ts
```diff
@@ -36090,7 +36090,11 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     }
 
     function isGenericFunctionReturningFunction(signature: Signature) {
-        return !!(signature.typeParameters && isFunctionType(getReturnTypeOfSignature(signature)));
+        if (!signature.typeParameters) {
+            return false;
+        }
+        const returnType = getReturnTypeOfSignature(signature);
+        return returnType.flags & TypeFlags.Intersection ? some((returnType as IntersectionType).types, isFunctionType) : isFunctionType(returnType);
```
jakebailey, line 36097: Not just `someType`?

jakebailey: (or whatever that helper is called that walks intersections and maybe unions)

Andarist: `someType` handles "single" types and unions, not intersections. There is `everyContainedType` that works for unions and intersections but I'd need `someContainedType` here. It can be done here, I just didn't bother.

jakebailey: Maybe not in this PR, I guess, but `isUnitLikeType`, `isTypeDerivedFrom`, `isGenericTypeWithUnionConstraint`, `isGenericTypeWithoutNullableConstraint` all do this same thing. (`isEmptyObjectType` copies/pastes the `every` variant` too.)

Andarist: Just pushed out a change introducing `someContainedType`

src/compiler/checker.ts
```diff
@@ -36090,7 +36094,11 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     }
 
     function isGenericFunctionReturningFunction(signature: Signature) {
-        return !!(signature.typeParameters && isFunctionType(getReturnTypeOfSignature(signature)));
+        if (!signature.typeParameters) {
+            return false;
+        }
+        const returnType = getReturnTypeOfSignature(signature);
+        return returnType.flags & TypeFlags.Intersection ? someContainedType(returnType, isFunctionType) : isFunctionType(returnType);
```
weswigham, line 36101: Isn't the point of the `someContainedType` helper that this can just be
```suggestion
        return someContainedType(returnType, isFunctionType);

```
?

Andarist: I intentionally dont want to run this for unions here. Not because I know we shouldnt but because I dont know if we should. Changing the behavior for unions would increase the likelihood of this breaking something for some project

weswigham:
Guess that means we should
@typescript-bot test it
but honestly, if that comes back clean, we should probably try handling the union case, too.

Andarist:
The above is just a single break (a single type-level test case from `reselect`). It can be simplified to [this playground](https://www.typescriptlang.org/play/?ts=5.5.2#code/C4TwDgpgBAggdiAYgVzgY2ASwPZygXigAoA6MgQwCcBzAZwC4pyEBtAXQEoCA+JhAbgBQg0JCgBlCABsIGbJQA8gqBODlg0QsxAAaZVABKEWsinACUVAGs42AO5w9KgApVyAW1pQIADw1wAEy9KCHIA3CkQPhB2C212PV5CIlo1DUZxNIgdKDISMDdPRldKD1oufF4jEzMhEXBoSRk5ShhKUpAFTPVNaKSoELCIqKbZYHkurO52OtFoAFE-UowjYGRKOAAVBoUUdCxcWjaO7z8IQODQ8LhI2AQ9jBw4dn6Ab30WAElA3yhMPCsEBA2AAZlAHgc4Ed2uQQGxGBCntCOl8fj42Kd-EFwahHodjrCWHBkO4AEYQShsfQqAD8hggaw220gu1xkORhO+AV8bG41KgjDgEAAbhShABfWYNCTSMbyaqmYAEzqjFpeXxYryq8atGEgJL6RbAZbAVbrLY7bXyWjcKViADC2DJ-wpCm+YGQwCtlHVZwuMuaOuVOQVZn6RH0eRCNWADzadAyspaoaVerdcA9XqTOptggqVWMirt0AA8p7M97Jj0Q4WwxYiCkshksvn6TGuAAyKDvFRoGRUe3kNAACwgRA4jGF2EwAQldQA9POoHMoGXgBXsxNuhoazG3oJFypiKkes2ehO20WD0ve-3KIOR2OL1OZ0JD5LhNy+1RoCC2U8oDQQYNErfRt2yfR3U9b1fU1AM5V1Doqw0bgnEvMxEgjFQ8iA0IQM3XUEygFh+Tyf4N0Da1GCgrNKJ9NDeydUkXUoRhHWdIVFBomDd0VVD9CpC81wohDkOydDgFtYQV3AixXigEFsGwRhUkof5qH4KBSSoQUSXJShNO0gAvRhSSU-s8A-XwwHkcw0EOcxaAIixcJ6b0sMA4CIHc-lG1PVRzx4KATw0EhFOwBjjybAKNFbEKIBIbTKEiohwpyJLWyIeS0q0qgoHFDg0MK-Q-PSGKIDirJEvIIy0KyhSlPSvLxSaozWx7I97KhcxwoARhU411IscKhCPKBDwAAWAWgAFprLGOb2nkfkutSBrsAAJl0skKWGpS6jG1bzCS-qoGJHbKAsJLRqPSbprmnxIAwRbKGWw6HNyygtuCwa4GoK6qAOzqPuM06zOwCyAaMm6VDu2b5ueilXsoFaQZq77VKGwhjJulq8yEIA)

The break is related to [this comment](https://github.dev/microsoft/TypeScript/blob/5d70bf894efe9014d24d4ba439807311e5b33fb8/src/compiler/checker.ts#L36071-L36082). There is no way to know if deferring will lead to better or worse results. It turns out that in this specific case, it leads to worse.

This test case **lucked out** heavily though. `OutputSelector` is a function type, it happens to have some extra properties on it but they are not essential to `reselect`'s API. If this would be a plain function then this test case wouldn't pass.

Similarly, it just happens to use an intersection to add extra properties. And that (without this PR) prevents this deferred processing. It similarly wouldn't work if it would be declared as a single type.

I think this break is acceptable. It's not great that it breaks something in the wild but it doesn't make sense to have functions with extra properties to behave differently based on how those extra properties are declared. 

There is a simple workaround for this case, one can move the nested `createSelector` call, assign its result to variable, and use that in place of the original nested call. This fixes the problem because the problem is related to how nested calls are resolved. Moving to a variable sidesteps the problem as the variable's type will settle before the compiler get to computing  "outer" call.

Issue #41486, opened 2020-11-11
Preserve property comment and insert @default value, if provided


jedmao:
## Search Terms

default initial initializer value preserve comment tsdoc type options option bag

## Suggestion

As the title suggests, insert a `@default` comment to the TSDoc AST if an initializer is provided, preserving the original comment.

## Use Cases

I have a default value for an option and I want the consumer of the library to know what the default value is.

## Examples

```ts
interface Options {
  /**
   * The server host.
   */
  host?: string;
  port?: number;
}

function serve({ host = 'localhost', port }: Options) {
  return { host, port };
}
```

[Playground](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgPIAczAPYgM7IDeAUMsgPQBUlpZlyAKgBYp7QBu0yT2eYAdLWSVytHnwD8ALmR8ooAOYBuWumxQw05CACuAWwBG0FQF9ixGDpAIsuWRwgAKQt15hkAXmQByADbYEOF9xMG8AGmQ1DWQTGQxbfABKIlooCDAdKBAiVz4IqPcTU2IgA)

Hovering over `host` in the `return` statement shows the following in vscode:

![Screen Shot 2020-11-10 at 6 39 31 PM](https://user-images.githubusercontent.com/1058243/98750631-15016400-2384-11eb-8cec-a67e28809eba.png)

Not only is the TSDoc comment lost for this property, but it could furthermore be improved if `@default "localhost"` were added to it, seeing as that information is available in this context.

Much ❤️ for the TS team!

## Checklist

My suggestion meets these guidelines:

* [x] This wouldn't be a breaking change in existing TypeScript/JavaScript code
* [x] This wouldn't change the runtime behavior of existing JavaScript code
* [x] This could be implemented without emitting different JS based on the types of the expressions
* [x] This isn't a runtime feature (e.g. library functionality, non-ECMAScript syntax with JavaScript output, etc.)
* [x] This feature would agree with the rest of [TypeScript's Design Goals](https://github.com/Microsoft/TypeScript/wiki/TypeScript-Design-Goals).

MartinJohns:
I think this is a duplicate of #32392.

jedmao:
> I think this is a duplicate of #32392.

It’s related, but not the same, because that issue says nothing about `@default` comments.

jedmao:
@RyanCavanaugh I've provided a detailed example of the scenarios that are missing tooltip information below 👇🏻 
[...]

Issue #17146, opened 2017-07-12
Transforming type information into JSDoc comments


samvv:
So I am trying to create a new documentation generator for TypeScript together with @BurtHarris. Currently it does not do much; I've been mainly tinkering a bit with the compiler API to see what it is capable of. You will be able to find it [here](https://github.com/TypeForce/tsdoc) when it's finished.

One of the things we thought about was the ability to 'target' JSDoc comments using the type information provided by TypeScript. I've been investigating how to tackle the issue. Ideally, it would use the transformers the compiler API has to offer but of all the functions provided in `factory.ts`, the ones that came closest were `addSyntheticLeadingComment`, `setSyntheticLeadingComments`, `setSyntheticTrailingComments` and `addSyntheticTrailingComment`; none of which provide the means to create complete synthetic JSDoc tags and comments.

So with this issue I'm actually asking the status-quo of this feature. Is it still possible to get this on the [roadmap](https://github.com/Microsoft/TypeScript/wiki/Roadmap)? If so, can I help? If not, is there a possibility of a patch that makes it possible to implement this as a third-party module?

### Related issues

 - [Generate doc comments in emitted JS](https://github.com/Microsoft/TypeScript/issues/10)
 - [Cannot add or modify JSDoc with custom transformers](https://github.com/Microsoft/TypeScript/issues/15639)
 - [Expose public API for transformation](https://github.com/Microsoft/TypeScript/pull/13940)
 - [Replace emitter with syntax tree transformations and a simplified node emitter](https://github.com/Microsoft/TypeScript/issues/5595)

**PS** I'll be away tomorrow so it won't be strange if it takes a while to answer; I just wanted to get the debate going.

samvv:
Also related: http://ts2jsdoc.js.org/. It hasn't been updated in a while, however.

samvv:
I think we've found a way around it by using `ts.createSynthesizedNode` directly. Closing this issue for now.

samvv:
I'm re-opening this as I would like to know if someone of the TypeScript team would accept pull requests for this feature. I've tried implementing it as a third-party module, but I need access to some private functions and, more generally, this seems rather a good fit for being a built-in transformer.

rbuckton:
We are discussing what changes we could make to comment emit to allow this. We already scan comments and often parse JSDoc comments for nodes. One option we are considering is reifying all comments in the compiler and storing them on `Node`, which we could then make accessible to transformers.

[...]

PR #58879, opened 2024-06-16
feat(transformers): Add beforeDeclarations transformers

Danielku15:
Relates to #17146 and #41486 (not fully fixes them)

The custom transformers were lacking a `beforeDeclarations` allowing devs to do a AST transformation before the built-in transformers. If you wanted to use information from erased AST nodes (visibility, initializers, etc.) you were lost. With this change devs can improve their JSDoc comments or use other transformations specific for declarations. 

Sorry for the PR without `Backlog` milestone issue. I still hope this makes it through.

### DIFF
--- a/src/compiler/transformer.ts
+++ b/src/compiler/transformer.ts
@@ -183,7 +183,11 @@ function getScriptTransformers(compilerOptions: CompilerOptions, customTransform
 
 function getDeclarationTransformers(customTransformers?: CustomTransformers) {
     const transformers: TransformerFactory<SourceFile | Bundle>[] = [];
+
+    addRange(transformers, customTransformers && map(customTransformers.beforeDeclarations, wrapDeclarationTransformerFactory));
+
     transformers.push(transformDeclarations);
+
     addRange(transformers, customTransformers && map(customTransformers.afterDeclarations, wrapDeclarationTransformerFactory));
     return transformers;
 }
--- a/src/compiler/types.ts
+++ b/src/compiler/types.ts
@@ -4867,6 +4867,8 @@ export interface CustomTransformers {
     before?: (TransformerFactory<SourceFile> | CustomTransformerFactory)[];
     /** Custom transformers to evaluate after built-in .js transformations. */
     after?: (TransformerFactory<SourceFile> | CustomTransformerFactory)[];
+    /** Custom transformers to evaluate before built-in .d.ts transformations. */
+    beforeDeclarations?: (TransformerFactory<Bundle | SourceFile> | CustomTransformerFactory)[];
     /** Custom transformers to evaluate after built-in .d.ts transformations. */
     afterDeclarations?: (TransformerFactory<Bundle | SourceFile> | CustomTransformerFactory)[];
 }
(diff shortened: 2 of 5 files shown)


sandersn:
To help with PR housekeeping, I'd prefer not to have PRs for bugs that aren't accepted yet. Discussion usually brings up new implementation requirements, and the codebase can change underneath the PR.

rbuckton:
Unfortunately, I would not classify this as "low-hanging fruit". I'm very wary about adding a `beforeDeclaration` transformer given that the current declaration transformer expects to receive the original parse tree, not a transformed tree, and may be making many assumptions about the provenance of nodes in the tree. Adding a "before" parser to the transforms required a fair amount of consideration and rethinking of the `ts` transformer to ensure it worked properly with a transformed tree. I'd like to see far more test coverage of various potential tree mutations and their effects on declaration output before I'd feel confident about making this change.

rbuckton:
> may be making many assumptions about the provenance of nodes in the tree.

For example, *src/compiler/transformers/declarations.ts* contains *at least* 38 references to `.parent`, but a transformed source tree does not have reliable `.parent` pointers, and that's not taking into account imported functions that might also access `.parent`.

Danielku15:
I understand your concern, but how is `beforeDeclarations` much different to normal `before` transformers? There are also plenty of built-in `before` transformers which run after the custom transformers:

https://github.com/microsoft/TypeScript/blob/dc535a7be8f40fde10110b0fc29a106499327118/src/compiler/transformer.ts#L133-L139

With your argumentation `before` transformers should be removed because devs might create transformers producing a nodes incompatible with the built-in ones. Is your concern that you are held responsible for custom transformers breaking the built-in ones causing too much support?

rbuckton:
> I understand your concern, but how is `beforeDeclarations` much different to normal `before` transformers? 

> With your argumentation `before` transformers should be removed because devs might create transformers producing a nodes incompatible with the built-in ones.

You misunderstand my concern. Prior to adding custom `before` transforms, the `ts` transform *also* had numerous places where it expected to only receive an original parse tree. When we added custom `before` transforms, the `ts` transform had to be updated to be more resilient to custom transforms by avoiding `.parent` pointers, making calls to `getOriginalNode` as needed, etc. 

As far as the other JavaScript-emitting transformers, they were all initially written with the expectation that they were receiving a transformed tree, so they made no assumptions about provenance and avoided these pitfalls.

If we want to add a `beforeDeclarations` transform, the `declarations.ts` transform must *also* be made resilient.

Danielku15:
Thanks a lot for clarfiying, makes totally sense. Once I find some time for this topic again, I might dive a bit deeper into the transformer codebase to see how they are currently organized compared to the normal `before` runners. 

Are there already some good tests for `before` transformers which try to break the built-in transformers I could look at?

rbuckton:
> Are there already some good tests for `before` transformers which try to break the built-in transformers I could look at?

We have a number of tests in src\testRunner\unittests\transform.ts and src\testRunner\unittests\customTransforms.ts that exercise custom `before` transforms.

PR #58884, opened 2024-06-16
feat(58561): Allow leading underscore for types to bypass noUnusedLocals warning

### LINKED ISSUE #58561: Allow leading underscore for types to bypass noUnusedLocals warning
### 🔍 Search Terms

"ts-ignore", "ts-expect-error", "noUnusedLocals", "noUnusedLocals underscore", "noUnusedLocals prefix"

### ✅ Viability Checklist

### ⭐ Suggestion

Add support for leading underscore in type definition name to disable tsc noUnusedLocals warning `ts(6196)`, like it does for other things.

### 📃 Motivating Example

Allows for easier testing of types by providing a way to enabled `noUnusedLocals` for your project, and selectively opting out for local types used for testing. Prior to this, you had a couple of less than idea solutions:

1. live with the warnings
2. disable `noUnusedLabels` for test code
3. use `@ts-ignore`

Depending on the method you're using for testing your types, you might not be able to use `@ts-ignore`. For example, if you were relying on usage of `@ts-expect-error` to assert invalid usages of your types in tests, `@ts-ignore` would also cause typescript not to flag the type error you're trying to assert with `@ts-expect-error`.

For example, a test like this:
```ts
type StringKeyOf<T extends Record<string, unknown>> = T extends T
  ? Extract<keyof T, string>
  : never;

type TestType = { s: unknown; 1: unknown; };

// NOTE: `expectFails` type will get flagged with unused warning, when `noUnusedLocals` is enabled,
//     and `@ts-ignore` is no help because it also suppresses the compile error we want to validate

// @ts-expect-error - validate type error will occur
type expectFails = Expect<Equals<StringKeyOf<TestType>, 's' | 1>>;
```

Previously, prefixing the local type with underscore would have no affect on the warning, but now you can update the code to include a leading underscore for the type name, and the warning will go away.
```ts
// @ts-expect-error
type _expectFails = Expect<Equals<StringKeyOf<TestType>, 's' | 1>>;
```

### 💻 Use Cases

    - I'd like to use to in my type validation tests.
3. What shortcomings exist with current approaches?
    - The current approaches do not provide a very good way to avoid ending up with a lot of warnings in tests, without disabling `noUnusedLocals`
5. What workarounds are you using in the meantime?
    - I'm currently disabling `noUnusedLocals` in my project's base _tsconfig.json_ and enabling in _tsconfig.build.json_. This choice was made so the errors would not be present for developers working on test related code, and since the build config does not include tests, it applies the rule against production code. It results in a less than ideal developer experience though, because developers will see errors in build that don't present when viewing the source code, because vscode is using _tsconfig.json_.

a-tarasyuk:
Fixes #58561

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -43174,6 +43174,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             }
         }
     }
+
     function isTypeParameterUnused(typeParameter: TypeParameterDeclaration): boolean {
         return !(getMergedSymbol(typeParameter.symbol).isReferenced! & SymbolFlags.TypeParameter) && !isIdentifierThatStartsWithUnderscore(typeParameter.name);
     }
@@ -43194,6 +43195,13 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     }
 
     function isValidUnusedLocalDeclaration(declaration: Declaration): boolean {
+        if (isTypeAliasDeclaration(declaration)) {
+            /**
+             * ignore starts with underscore names _
+             * type _T = number;
+             */
+            return isIdentifierThatStartsWithUnderscore(declaration.name);
+        }
         if (isBindingElement(declaration)) {
             if (isObjectBindingPattern(declaration.parent)) {
                 /**
--- a/tests/baselines/reference/unusedTypeDeclarations.errors.txt
+++ b/tests/baselines/reference/unusedTypeDeclarations.errors.txt
@@ -1,0 +1,11 @@
+unusedTypeDeclarations.ts(1,6): error TS6196: 'T1' is declared but never used.
+
+
+==== unusedTypeDeclarations.ts (1 errors) ====
+    type T1 = number;  // error
+         ~~
+!!! error TS6196: 'T1' is declared but never used.
+    type _T2 = number; // ok
+    
+    export {};
+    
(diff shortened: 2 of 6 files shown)


src/compiler/checker.ts
```diff
@@ -43194,6 +43195,13 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     }
 
     function isValidUnusedLocalDeclaration(declaration: Declaration): boolean {
+        if (isTypeAliasDeclaration(declaration)) {
```
jakebailey, line 43198: At some level, I wonder if we should just be checking `isIdentifierThatStartsWithUnderscore` for all declarations, rather than special casing a few kinds. What declarations are we actually missing here?

a-tarasyuk: I think this is open to discussion. The fix shouldn't require significant work... What do you think, @RyanCavanaugh?

jakebailey: I guess it's sort of nonsensical to do this for enum members and properties and so on; after all, their names are important.

a-tarasyuk: @jakebailey What are your thoughts on the following list? Should we add or exclude any additional declarations?

1. Interface Declaration
2. Type Alias Declaration
3. Enum Declaration
4. Class Declaration

jakebailey: I would say any declaration that can be a child of a `Block` or `SourceFile` or `Module`, etc, if that makes sense. So basically _not_ enum members or properties. Not sure we have a list of that variety, though...

PR #58895, opened 2024-06-17
Fix global when typescript.js loaded as script

### LINKED ISSUE #58893: `ts` object has no properties when loading typescript.js in a browser with v5.5-rc
### Acknowledgement

### Comment

This came up after sending the initial Google v5.5-beta feedback in https://github.com/Microsoft/TypeScript/issues/58685 since I was just looking at build failures for that feedback.

Apologies if this isn't an officially supported way of using the TypeScript compiler or not - most of the docs I can find are specific to running the compiler via Node.

Google has some internal web apps that load the bundled TSC from https://github.com/microsoft/TypeScript/blob/v5.5-rc/lib/typescript.js file in a browser. One example use case is a tool similar to https://www.typescriptlang.org/play/.

In TS 5.5 these apps break at runtime because they call methods on the global `ts` object like `ts.createLanguageService`, but `ts` is now just an empty object with no properties. I narrowed this down to the bundling changes in https://github.com/microsoft/TypeScript/pull/57133. Here's typescript.js before & after:

[v5.4.2](https://raw.githubusercontent.com/microsoft/TypeScript/v5.4.2/lib/typescript.js):

``` js
var ts = (() => {
  // lots and lots of code
  return require_typescript();
})();

if (typeof module !== "undefined" && module.exports) { module.exports = ts; }
```

[v5.5-rc](https://raw.githubusercontent.com/microsoft/TypeScript/v5.5-rc/lib/typescript.js):

``` js
var ts = {}; ((module) => {
// lots and lots of code
module.exports = __toCommonJS(typescript_exports);
// even more code
})(typeof module !== "undefined" && module.exports ? module : { exports: ts });
if (typeof module !== "undefined" && module.exports) { ts = module.exports; }
```

I'd assume that in v5.5-rc, `ts` was intended to be instantiated via passing `{exports : ts }` as the `module` argument to this IIFE, causing the exported functions that are put on `module.exports` to also be defined as properties on `ts`. Instead the `module.exports = __toCommonJS(typescript_exports);` assignment just overwrites `ts` in `{exports: ts}` and `ts` is just an empty object.

jakebailey:
Fixes #58893

Not sure why I thought this worked, but I probably just assumed that the tests passed therefore it was okay. The tests didn't actually test anything. Oops.

getters/setters have been supported ~forever, so this method is fine. We have tests which verify everything now.

### DIFF
--- a/Herebyfile.mjs
+++ b/Herebyfile.mjs
@@ -216,7 +216,7 @@ function createBundler(entrypoint, outfile, taskOptions = {}) {
             // Monaco bundles us as ESM by wrapping our code with something that defines module.exports
             // but then does not use it, instead using the `ts` variable. Ensure that if we think we're CJS
             // that we still set `ts` to the module.exports object.
-            options.footer = { js: `})(typeof module !== "undefined" && module.exports ? module : { exports: ts });\nif (typeof module !== "undefined" && module.exports) { ts = module.exports; }` };
+            options.footer = { js: `})({ get exports() { return ts }, set exports(v) { ts = v; if (typeof module !== "undefined" && module.exports) { module.exports = v } } })` };
 
             // esbuild converts calls to "require" to "__require"; this function
             // calls the real require if it exists, or throws if it does not (rather than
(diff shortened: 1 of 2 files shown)


jakebailey:

jakebailey:

Checking startup perf, should not actually change much.

andrewbranch (approved):

jakebailey:

jakebailey:
@typescript-bot pack this

dunno if this works on merged PRs...

jakebailey:
nope

PR #58903, opened 2024-06-17
`organizeImports` makes no changes if there are parse errors in the sourceFile

iisaduan:
When called from our services, `organizeImports` returns no changes if there are parse errors in the `sourceFile`.

fixes https://github.com/microsoft/TypeScript/issues/56252 / https://github.com/microsoft/TypeScript/issues/56252#issuecomment-1784231506

The original crash happened because the parser was unable to parse a .js file, and as a result, the failed parse generated two duplicate/overlapping modules. Organize imports then tried to remove the same imports twice, leading to a change tracker crash.

### DIFF
--- a/src/services/services.ts
+++ b/src/services/services.ts
@@ -2508,8 +2508,9 @@ export function createLanguageService(
         synchronizeHostData();
         Debug.assert(args.type === "file");
         const sourceFile = getValidSourceFile(args.fileName);
-        const formatContext = formatting.getFormatContext(formatOptions, host);
+        if (containsParseError(sourceFile)) return [];
 
+        const formatContext = formatting.getFormatContext(formatOptions, host);
         const mode = args.mode ?? (args.skipDestructiveCodeActions ? OrganizeImportsMode.SortAndCombine : OrganizeImportsMode.All);
         return OrganizeImports.organizeImports(sourceFile, formatContext, host, program, preferences, mode);
     }
--- a/src/services/services.ts
+++ b/src/services/services.ts
@@ -2508,7 +2508,7 @@ export function createLanguageService(
         synchronizeHostData();
         Debug.assert(args.type === "file");
         const sourceFile = getValidSourceFile(args.fileName);
-        if (containsParseError(sourceFile)) return [];
+        if (containsParseError(sourceFile)) return emptyArray;
 
         const formatContext = formatting.getFormatContext(formatOptions, host);
         const mode = args.mode ?? (args.skipDestructiveCodeActions ? OrganizeImportsMode.SortAndCombine : OrganizeImportsMode.All);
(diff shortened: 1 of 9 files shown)


andrewbranch (approved):
This seems very reasonable to me, given commands like Go To Definition do nothing when there are parse errors. However, I will note that there were existing tests that showed the feature working when a parse error occurred somewhere else in the file. I think the crash this fixes happened when there was a parse error on a node that organize imports was actually going to operate on / analyze, so it might be possible to make a narrower fix and ignore nodes with parse errors deeper in the organize imports logic. I’m skeptical that’s worth it, but might look for input from other reviewers.

jakebailey (approved):
I feel like this is fine; it's not like eslint or dprint or prettier will operate on malformed files to organize imports either.

DanielRosenwasser:
> This seems very reasonable to me, given commands like Go To Definition do nothing when there are parse errors.

Is that true? Do we have an example?

DanielRosenwasser:
The thing I would suggest is that when a parse error is present in a file:

* Organize Imports will never remove unused imports (because the file might be borked and semantics won't work).
* Organize Imports only sorts imports when no parse error occurs by any import/export statement.
* Similarly, Sort Imports allows you to sort imports when no parse error occurs.

iisaduan:
> > This seems very reasonable to me, given commands like Go To Definition do nothing when there are parse errors.
> 
> Is that true? Do we have an example?

Quick example from the test case that I had to change (it had a parse error since `case` is a keyword)

```
import { case/*1*/, Sensitive/*2*/ } from './a'
````
```
//a.ts
export const case = 1;
export const Sensitive = 2;
````
GoToDefinition works fine at `2`, but does nothing at `1`, so it has more specific behavior than just "no definitions if there's any parse issues anywhere in the file"

File: src/services/services.ts [hunk 1]
```diff
@@ -2508,8 +2508,9 @@ export function createLanguageService(
         synchronizeHostData();
         Debug.assert(args.type === "file");
         const sourceFile = getValidSourceFile(args.fileName);
-        const formatContext = formatting.getFormatContext(formatOptions, host);
+        if (containsParseError(sourceFile)) return [];
```
sheetalkamat, line 2511: Return emptyArray

andrewbranch:
> commands like Go To Definition do nothing when there are parse errors

> GoToDefinition works fine at `2`

I should have said Go To Definition does nothing when there are parse errors _in the file where it’s triggered_

DanielRosenwasser:
@andrewbranch it still works in part, it is just isn't guaranteed to give you correct results. The language service is generally pretty resilient within a file even in the presence of a parse error.

andrewbranch:
You’re right—I based that assertion off a random single experiment, but putting another one together, I see it’s not disabled.

jakebailey:
Is there anything left to do here?

iisaduan:
I guess it doesn't have to be included in this PR. I wanted to bring up at sync if VS code has support for an error, instead of failing silently.

Issue #58542, opened 2024-05-15
Bad error when using `import()` type within JSDoc tag `@implements`


jaydenseric:
### 🔎 Search Terms

- `import`
- `@implements`
- `2304`

### 🕗 Version & Regression Information

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about the JSDoc tag [`@implements`](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#implements)

### ⏯ Playground Link

https://www.typescriptlang.org/play/?ts=5.5.0-beta&filetype=js#code/PQKhAIAEEsFsAcA2BTWyB2AXAzuA3nPAPYBOmAFAEQCGlAlAHQBCAvuCMAFADGi12uAIL4WQA

### 💻 Code

```js
/** @implements {import("a").B} */
class A {}
```

### 🙁 Actual behavior

Two TypeScript errors:

1. `Cannot find name 'import'.` (2304)
2. `'}' expected.` (2304)

<img width="469" alt="Screenshot 2024-05-15 at 5 59 00 PM" src="https://github.com/microsoft/TypeScript/assets/1754873/0cd66650-482b-477b-a1c0-dc326ac5d077">

### 🙂 Expected behavior

TypeScript should allow using `import()` types within the type of the JSDoc tag [`@implements`](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#implements).

### Additional information about the issue

A workaround is to use a JSDoc `@typedef` to import the type under an alias, and then use that alias within the JSDoc tag `@implements` type, e.g:

```js
/** @typedef {import("a").B} B */

/** @implements {B} */
class A {}
```

sandersn:
`@implements` uses the same grammar as Typescript implements. This is an error in Typescript:

```ts
class A implements import("a").B {}
```

However, the error is informative in typescript: "A class can only implement an identifier..." and not in Javascript. Javascript should give the same error.

With the new `@import` tag, you should be able to `/** @import { B } from 'a' */` instead of using `@typedef`.

jaydenseric:
See also https://github.com/microsoft/TypeScript/issues/49905 .

PR #58904, opened 2024-06-17
Improve `@implements` error

sandersn:
Make `@implements/@extends` parse its expression with `parseLeftHandSideExpressionOrHigher` instead of `parsePropertyAccessEntityNameExpression`. This turns many more parse errors into grammar errors, just like in TS. I kept `parseExpressionWithTypeArgumentsForJSDoc` as a separate function in order to keep parsing asterisks -- and `@implements` still only supports a single interface, unlike TS which supports a list of them.

This changes the public API, making the `expression` property a supertype of what it used to be. The one change it required in our codebase is in checker.ts

Fixes https://github.com/microsoft/TypeScript/issues/58542

### DIFF
--- a/src/compiler/types.ts
+++ b/src/compiler/types.ts
@@ -3949,12 +3949,12 @@ export interface JSDocUnknownTag extends JSDocTag {
  */
 export interface JSDocAugmentsTag extends JSDocTag {
     readonly kind: SyntaxKind.JSDocAugmentsTag;
-    readonly class: ExpressionWithTypeArguments & { readonly expression: Identifier | PropertyAccessEntityNameExpression; };
+    readonly class: ExpressionWithTypeArguments;
 }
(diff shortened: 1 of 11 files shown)


src/compiler/types.ts
```diff
@@ -3949,12 +3949,12 @@ export interface JSDocUnknownTag extends JSDocTag {
  */
 export interface JSDocAugmentsTag extends JSDocTag {
     readonly kind: SyntaxKind.JSDocAugmentsTag;
-    readonly class: ExpressionWithTypeArguments & { readonly expression: Identifier | PropertyAccessEntityNameExpression; };
+    readonly class: ExpressionWithTypeArguments;
```
jakebailey, line 3952: Are these types wrong and that's why they're changing?

PR #58910, opened 2024-06-18
Filter return type inferences by constraint applicability

### LINKED ISSUE #41461: Generic constrained to function: parameter is incorrectly inferred as `any`
**TypeScript Version:**  4.0.2

**Search Terms:**

**Code**

```ts
declare function useCallback<T extends (...args: any[]) => any>(
    callback: T,
): T;
const test: ((x: string) => void) | undefined = useCallback((x) => {
    x; // ❌ type is inferred as `any`, expected `string`
});
```

For context, this typing of `useCallback` is extracted from the React types.

**Expected behavior:** see above

**Actual behavior:** see above

**Playground Link:** https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEAzArgOzAFwJYHtX2QGcQBhKCCAIyjAGsAeAFXhAA8MRVhD4AKAOkGwA5oQBc8KKgCeAbQC6ASngBeAHySZa3gCh4++GHJUatCYwA0OxeYDcOsHkIZ4HZxN69WE5zCyphZXV4ADccLGBlAB8CLhBEfxBgVQJiMgpqOk9WII0Abz0DVlt4AHpS+EAZcldpAAcELB5-RBAYOGSoHgADKWkuixZWeswk+C7ff2EunQBfRVsgA

**Related Issues:** 

- https://github.com/microsoft/TypeScript/issues/31146
- https://github.com/DefinitelyTyped/DefinitelyTyped/pull/30057#discussion_r232123748

### LINKED ISSUE #50719: Function parameter does not get inferred through target typing by a union through a layer of indirection
# Bug Report

### 🔎 Search Terms

- optional function parameter inference 
- optional method parameter type inference
- optional method inference indirection 
- nullable method inference
- target type method union inference 
- target type generic function argument 

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about type inference

### ⏯ Playground Link

[Playground link with relevant code](https://www.typescriptlang.org/play?ts=4.9.0-dev.20220911#code/PTAEBkEsGsFNUgO1AJVgQwMYBcB0BXAZ1gGF0AbcgIy2gCgATWTc9AJ3gDN9EdIB7ZEVIVqtADwAVULAAe2WIgaFQAMR59BAPgAUnRAC5QkgJRHJAbjqNmrDqG69sA5HICMOzKJqZoRnbJGiPgAtlSwbCagALxaoABu-JAMZglJDFY2LOxcGs6CMrIATJ7etAD8-oGgwWERUbFpyamJyZlM2faOmq6yAMyllD5+oDoBQaHhkTFxrSmgAD6gAKwt6ZnuATOgAN4AvlEgoADyANK4oNW1U3SbwmRDtFuN+yYmFqBHZxdXkxHWchKsm2r0+YG+lwmdTYt2KOnuZV8zzir3eYNAAHYAAxYgBso0Q-AQIQADuRIJhINhQOhEABPEwA-rI3YHdEQ37Q2EDBGPJHAl4HNFHbF4glEyCk8mU6m0hl0IA)

### 💻 Code

```ts
// Like in React.useCallback
declare function useCallback<T extends Function>(fn: T): T;

declare function ex1(callback: (x: number) => void): void;

declare function ex2(callback?: (x: number) => void): void;

declare function ex3(callback: ((x: number) => void) | 5): void;

ex1(x => {}) // OK. x: number
ex1(useCallback(x => {})); // OK. x: number

ex2(x => {}) // OK. x: number
ex2(useCallback(x => {})); // 7006 (no implicit any)

ex3(x => {}) // OK. x: number
ex3(useCallback(x => {})); // 7006 (no implicit any)
```

### 🙁 Actual behavior

The type of `x` in the code example above is not inferred when `callback` is a union or optional.

### 🙂 Expected behavior

`x` should be inferred as `number`.

---
Edit: simplified example

### LINKED ISSUE #57905: Sync/Async mix return types with Generics not allowed
### 🔎 Search Terms

generics, promise

- Tested with 5.3 and 5.4.2

### ⏯ Playground Link

https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBAQwEYGcZQQY3pgNglFOAZRmgQHNg4BvAKDkcVXS3ipgB4AVOUGYADsAJkTRQAloIoA+ABQBKAFxxeAHzgAFKBAC2ElMB4yA3HQC+dOqEixm4tnDwEixAJ6DMpclT4gBIq5kGL70TPas2HAQAG7AUJLC1Bw8fgGicOJSsooq3GbmQA

### 💻 Code

```ts
export abstract class Storage {
    abstract get<T extends string>(): T | Promise<T>;
}

export abstract class SyncStorage extends Storage {
    abstract override get<T extends string>(): T;
}
```

### 🙁 Actual behavior

the overridden "get" method gives a type error:

> Property 'get' in type 'SyncStorage' is not assignable to the same property in base type 'Storage'.
  Type '<T extends string>() => T' is not assignable to type '<T extends string>() => T | Promise<T>'.
    Type 'string' is not assignable to type 'T | Promise<T>'.(2416)
> (method) SyncStorage.get<T extends string>(): T

### 🙂 Expected behavior

it detects that it is an allowd return type of the parent class and allows it

As soon as I do this without generics all works as expected, just with generics it does not

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/50719
fixes https://github.com/microsoft/TypeScript/issues/41461
fixes https://github.com/microsoft/TypeScript/issues/57905


src/compiler/checker.ts
```diff
@@ -26852,14 +26852,31 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
                 const instantiatedConstraint = instantiateType(constraint, context.nonFixingMapper);
-                if (!inferredType || !context.compareTypes(inferredType, getTypeWithThisArgument(instantiatedConstraint, inferredType))) {
-                    // If the fallback type satisfies the constraint, we pick it. Otherwise, we pick the constraint.
-                    inference.inferredType = fallbackType && context.compareTypes(fallbackType, getTypeWithThisArgument(instantiatedConstraint, fallbackType)) ? fallbackType : instantiatedConstraint;
+                if (!inferredType) {
+                    inference.inferredType = instantiatedConstraint;
+                }
+                else {
+                    if (inference.priority! & InferencePriority.ReturnType) {
+                        inference.inferredType = filterContextualInferredType(inferredType, instantiatedConstraint) ?? (fallbackType && filterContextualInferredType(fallbackType, instantiatedConstraint)) ?? instantiatedConstraint;
```
weswigham, line 26860: We only need to bother with this filtering if the constraint check for the whole union fails, right? (Since if it passes, no filtering will occur, as every union member must also pass.) This can probably be *inside* the comparison branch below?

Andarist: That’s right but it felt cleaner to have this in a separate branch. Shouldnt filterType already optimize for array/type allocations?

jakebailey:
@typescript-bot test it

gabritto:
This had a ton of breaks on the extended tests, so seems like not a good idea?

Andarist:
yep, might not be a good idea after all - I didn't yet look at those breaks though, at the very least I'd like to get some new test cases based on them so we could add them to `main`

Andarist:
Slimmed down repro based on `headlessui`:
```ts
type Query = (
  container: HTMLElement,
  ...args: any[]
) =>
  | Error
  | HTMLElement
  | HTMLElement[]
  | Promise<HTMLElement[]>
  | Promise<HTMLElement>
  | null;

interface Queries {
  [T: string]: Query;
}

type FindByText<T extends HTMLElement = HTMLElement> = (
  container: HTMLElement,
  text: string,
) => Promise<T>;

declare function findByLabelText<T extends HTMLElement = HTMLElement>(
  ...args: Parameters<FindByText<T>>
): ReturnType<FindByText<T>>;

const queries = {
  findByLabelText,
};

type MapQueries<Q extends Queries = typeof queries> = {
  [P in keyof Q]: Q[P];
};
```

I didn't quite expect a change in any code like this... I'll continue investigating this.

jakebailey:
@typescript-bot test it

jakebailey:
@typescript-bot test it

Andarist:
@gabritto I made a very silly mistake initially. It was somewhat a coincidence that the tests I had passed 😅 This was fixed now and I added extra test cases. The PR is now in a good state and can be reviewed :)

Above we can see some missing implicit anys in webpack. I added a test case based on that, this is a clear improvement.

ehoogeveen-medweb:
Aside from the new return type branch, it seems like this also changes the `!inferredType` case to always use `instantiatedConstraint` instead of checking whether `fallbackType` satisfies the constraint. Is that an intended part of this change?

Andarist:
No, that wasn't intentional. Thanks for pointing that out - I corrected the PR but I doubt it will make a difference for any of the extended tests.

Andarist:
@gabritto @weswigham friendly 🏓 :)

jakebailey:
@typescript-bot test it
@typescript-bot pack this

jakebailey:
Just to note it, this code now matches the fix that Anders applied separately in the Go codebase, so this should be good as-is now, yay.

jakebailey (approved):
Cross comparing typescript-go, and this is matching.

PR #58912, opened 2024-06-18
Fix redundant `export {}` in declaration file

### LINKED ISSUE #51338: Emmited Type Declarations generate empty export
# Bug Report

When generating type declaration files if a type that is only used inside the typescript code however it is not exported results in an empty export at the end of the file e.g. `export {};`

### 🔎 Search Terms

* declaration empty export
* declaration "export {}"

When i started inspecting the generated typescript declarations generated.
I see the same error/behaviour when either `typescript@4.8` and `typescript@next`

This is the behavior in every version I tried, and I reviewed the FAQ for entries about emmited code and private fields and methods.

### ⏯ Playground Link

* When some of the types are not exported this happens playground example [here](https://www.typescriptlang.org/play?experimentalDecorators=false&emitDecoratorMetadata=false&target=99&stripInternal=true#code/C4TwDgpgBAKghgZwNY3NAvFA5AeQHICiWUAPtjAOo7FlYwASASgUQNwBQ7A9F1AO4B7AK4AbACZQAdgOBQIAD0gBjYBAnABUABZwAbtDiS5AWzCg5igQCdZ3XgEsjwLdAjH7wVerRQxEJSJwVnDA9gJGSgJ+2hBW0PYIUM4JUJFWcSoA-OwKYNayjqpWAGZwStDwyFAA3uxQFv5CqpVIABSgkABcLaiQAJSdugL2YhwAvuxAA)
* When all of the types are exported there is no empty export playground example [here](https://www.typescriptlang.org/play?experimentalDecorators=false&emitDecoratorMetadata=false&target=99&stripInternal=true#code/KYDwDg9gTgLgBDAnmYcAqBDAzgazc1AXjgHIB5AOQFES4AfUtAdTNoZLQAkAlKmgbgBQggPQi4AdwgBXADYATOADsI8UCgDGMYIpgQ4ACwwA3VBiVxgAWzBJL4aPFHiAlhZgHU1lzG26CcPLAGrIYUBgwLhAWGhBBhsBQqC5YCAYpcLFQSVoA-ILqjnBu2lAAZhgaqJi4cADegnD2wdLaNTgAFEgoAFzt+CgAlD3GEC7yQgC+gkA

#### Empty Export Code
```ts
type TaskType = 'ONE' | 'TWO' | 'THREE';

// would not expected to have an empty export 
// in the emitted type declaration code here is this correct?
export interface Task {
  executeTask(type:TaskType):void;
}
```

#### No Empty Export
```ts
export type TaskType = 'ONE' | 'TWO' | 'THREE';

// would not expected to have an empty export 
// in the emitted type declaration code here is this correct?
export interface Task {
  executeTask(type:TaskType):void;
}
```

### 🙁 Actual behavior

Empty export for some reason…

### 🙂 Expected behavior

No default empty export.

Dunqing:
Fixes #51338

### DIFF
--- a/src/compiler/transformers/declarations.ts
+++ b/src/compiler/transformers/declarations.ts
@@ -1718,6 +1718,8 @@ export function transformDeclarations(context: TransformationContext) {
                 ));
             }
         }
+        // Anything left unhandled is an error, so this should be unreachable
+        return Debug.assertNever(input, `Unhandled top-level node in declaration emit: ${Debug.formatSyntaxKind((input as Node).kind)}`);
 
         function cleanup<T extends Node>(node: T | undefined): T | undefined {
             if (isEnclosingDeclaration(input)) {
--- a/src/compiler/transformers/declarations.ts
+++ b/src/compiler/transformers/declarations.ts
@@ -507,7 +507,7 @@ export function transformDeclarations(context: TransformationContext) {
         else {
             const statements = visitNodes(node.statements, visitDeclarationStatements, isStatement);
             combinedStatements = setTextRange(factory.createNodeArray(transformAndReplaceLatePaintedStatements(statements)), node.statements);
-            if (isExternalModule(node) && (!resultHasExternalModuleIndicator || (needsScopeFixMarker && !resultHasScopeMarker))) {
+            if (isExternalModule(node) && !resultHasExternalModuleIndicator) {
                 combinedStatements = setTextRange(factory.createNodeArray([...combinedStatements, createEmptyExports(factory)]), combinedStatements);
             }
         }
@@ -1529,8 +1529,6 @@ export function transformDeclarations(context: TransformationContext) {
                 needsDeclare = false;
                 const inner = input.body;
                 if (inner && inner.kind === SyntaxKind.ModuleBlock) {
-                    const oldNeedsScopeFix = needsScopeFixMarker;
-                    const oldHasScopeFix = resultHasScopeMarker;
                     resultHasScopeMarker = false;
                     needsScopeFixMarker = false;
                     const statements = visitNodes(inner.statements, visitDeclarationStatements, isStatement);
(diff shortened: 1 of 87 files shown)


weswigham (changes requested):
This is not sufficient to fix the issue. I'm not sure @RyanCavanaugh remembered in the linked issue, but we emit a `export {}` not just to ensure module-ness, but to actually make privately scoped names work *at all* - unless a declaration file has a export declaration (not an export modifier - an export declaration) _all names are implicitly exported_. If we stop emitting this marker, that behavior _has to change_, so privates remain private. Breaking that will be a breaking change, though! Declaration files like
```ts
type A = 1;
export interface B {}
```
allow importing `A` today, so forbidding that is a break - arguably a good one, but it'll be a break with a declaration file behavior we've had since approximately the origin of declaration files.

fatcerberus:
And here I always assumed declaration files worked exactly like source files w.r.t. what’s exported - before today I would have confidently told anyone who asked that `A` there was internal to the file and not importable, just like it would be if that were a `.ts` file

jakebailey:
Declaration files have really, really weird export semantics. Been trying to augment ts-eslint to handle that in https://github.com/typescript-eslint/typescript-eslint/pull/8611 but it's definitely a battle, so not surprised to see this be wrong.

weswigham:
It hasn't really needed to be something we've needed to think about since our declaration emitter started emitting privates and `export {}` automatically to work around the issue; it's weird, but it does "just work", back to ancient versions of TS.

Dunqing:
This is weird. I am implementing the `Isolated Declarations` transformation in [oxc](https://github.com/oxc-project/oxc/tree/main/crates/oxc_isolated_declarations). When I came across this empty export I thought it was a bug. it doesn't always occur, only non-exports referenced in the export statement generate `export {}`. so I tried searching in the TypeScript issues to confirm my suspicion.

So I want to ask, when a third-party tool wants to implement `Isolated Declarations`, does it also have to port this behavior?

Boshen:
For future implementors of isolated declarations, you need port this "bug" as a feature 😁

We are already using the `transpileDeclaration` API to guide our implementation, but due to https://github.com/microsoft/TypeScript/issues/51338, we assumed it is a bug and hence this PR.

jakebailey:
Yeah; if you do find anything weird, please do report it so we can at least make sure it isn't actually a bug. e.g. https://github.com/oxc-project/oxc/pull/3798 is likely a bug on our end.

Not quite sure where to document `export {}`, as this has been this way for ages and also applies to DT/hand-written types; I could have sworn this was documented on the website somewhere...

PR #58920, opened 2024-06-18
Cache the length of arrays in core utilities.

DanielRosenwasser:
Follow-up to #58873 to measure effects of caching the length on all arrays passed into our utilities.


DanielRosenwasser:

PR #58928, opened 2024-06-19
Experiment - monomorphic  node/type/signature

dragomirtitian:
Experiment to make several objects monomorphic by keeping the common properties in a single object shape and moving all other properties to a separate object and using accessors to preserve compatibility with existing code.

~~Local improvements seem promising, building TSC/compiler: -20% in time with +10% in memory~~ (See bellow for results)

Code is rough around the edges, just testing out as a proof of concept.

Note: This is the result of me and @acutmore bouncing ideas about improving access to `Node.kind`

## Update

On Node,  the improvements with everything in this PR are: 

- a total time win of 16.0% (ranging 10.5% to 23.7%)
- a checker time win of 20.8% (ranging 16.6% to 26.8%)
- a memory cost of 2.7% (ranging 1.9% to 4.4%)

Which is an exchange rate of 5.8 on total time, and 7.6 on checker time.

Breaking down where the performance win comes from (from the separate PRs):

AST Node monomorphism: 
- total time win: 3.3% - 9%
- Check Time win: 3.7% - 11%
- Mem cost: ~0-1%
	
	
Types  monomorphism: 
- total time win: 4.1% - 20%
- Check Time win: 5.3% - 22.5%
- Mem cost: 0.75%-1.2%
	
Signatures  monomorphism:
- total time win: 0.45% - 1.72%
- Check Time win: 0.56% - 2.04%
- Mem cost: ~0%

The AST nodes win is smaller than I remember in the first experiments. This is probably due to the reduction in memory which meant some relatively common properties are now in `data` not in `Node`.


jakebailey:

src/services/transpile.ts
```diff
@@ -109,10 +110,14 @@ interface Symbol {
     readonly [Symbol.toStringTag]: string;
 }`;
 const barebonesLibName = "lib.d.ts";
-const barebonesLibSourceFile = createSourceFile(barebonesLibName, barebonesLibContent, { languageVersion: ScriptTarget.Latest });
```
jakebailey, line 112: We should totally fix this on main regardless of this change; didn't realize this was unconditionally created.

jakebailey: https://github.com/microsoft/TypeScript/pull/58931

Andarist:
This is the most green thing I've seen this week. Nice!!

src/services/transpile.ts
dragomirtitian: Stumbled into it  accidentally while looking at the megamorphism in tests, and could not for the life of me understand why there were still two node shapes, turns out this was created before the node allocator was swapped for the service one.

jakebailey: Oh dang, that's actually an even worse regression.

MichaelMitchell-at:
Unfortunately I didn't see a noticeable perf improvement in the Airtable code base, but did see a 5-10% increase in peak memory usage.

dragomirtitian:
> Unfortunately I didn't see a noticeable perf improvement in the Airtable code base, but did see a 5-10% increase in peak memory usage.

@MichaelMitchell-at i find this a bit surprising. While the % of difference varies, we've generally noticed improvements across the board. Is this a public repo? If it not, could you run the tsc on this branch through dexnode to get a report for the inline caches?  

Also how did you measure the times? What is the output with - - diagnostics?

MichaelMitchell-at:
> > Unfortunately I didn't see a noticeable perf improvement in the Airtable code base, but did see a 5-10% increase in peak memory usage.
> 
> @MichaelMitchell-at i find this a bit surprising. While the % of difference varies, we've generally noticed improvements across the board. Is this a public repo? If it not, could you run the tsc on this branch through dexnode to get a report for the inline caches?
> 
> Also how did you measure the times? What is the output with - - diagnostics?

Actually there seems to be an error with my methodology so throw out the previous result I mentioned. I can't seem to apply the patch cleanly at the moment. I'll just have to wait for this to reach nightly.

MichaelMitchell-at:
Ok, I looked into it a bit more. Our custom runner uses the compiler API and that code path hasn't been updated to support the new node structure yet.
```
> const ts = require('./lib/typescript.js')
> ts.readConfigFile('src/tsconfig.json', ts.sys.readFile)
Uncaught Error: Debug Failure. False expression.
    at parseUpdateExpression (TypeScript/lib/typescript.js:37704:11)
    at parseUnaryExpressionOrHigher (/TypeScript/lib/typescript.js:37622:32)
    at parseBinaryExpressionOrHigher (TypeScript/lib/typescript.js:37544:25)
    at parseAssignmentExpressionOrHigher (TypeScript/lib/typescript.js:37215:18)
```

I should be able to get past this if I can swap out this part with a lib that can parse JSON with comments and trailing commas.

jakebailey:

dragomirtitian:
Latest perf results seemed to have clawed back most of the parser perf loss.

jakebailey:
Half a gig of memory for vscode is a bit scary, though, given how close that is to the max memory limit...

dragomirtitian:
@jakebailey I did some extra work. I was too eager to make everything monomorphic, a bit of polymorphism is fine. There should e 3 shapes for nodes now: Token, Identifier and Node. There is a 4th one Token with original that i might look into removing, but it doesn't seem to have much impact for now.

With these changes, at least locally the increase in mem usage goes form 13% to about 5%, while keeping the same performance profile.

jakebailey:

jakebailey:
Much better, though emit is now much slower on certain benchmarks. Maybe a case or two aren't monomorphic for those benchmarks?

dragomirtitian:
For the extra parse time I have a pretty good idea. `TextRange` has too many shapes. Beyond the 3 AST nodes, we also have at least 2 if not more that come from `createNodeArray` making setting of positions much more expensive in some cases. 

Will look into it a bit more.

jakebailey:

jakebailey:
@typescript-bot perf test this faster host=bun@1.1.16

jakebailey:
perf test this faster host=bun@1.1.16,commits=aa249c094b9b96c86025e964469ed4c803190cca...977cbc5c8a01330fb3e691d6d7601ecdcd478f38

jakebailey:
@typescript-bot perf test this faster host=bun@1.1.16,commits=aa249c094b9b96c86025e964469ed4c803190cca...977cbc5c8a01330fb3e691d6d7601ecdcd478f38

jakebailey:
@typescript-bot perf test this faster host=bun@1.1.16 commits=aa249c094b9b96c86025e964469ed4c803190cca...977cbc5c8a01330fb3e691d6d7601ecdcd478f38

dragomirtitian:
I really am not sure what to do with this particular line:

> **self-build-src - bun (v1.1.16, x64)**
>Memory used | 7,591,188k (±34.53%) | 10,215,912k (± 3.11%) | 🔻+2,624,724k (+34.58%) | 9,662,812k | 10,529,743k | p=0.005 n=6
>-- | -- | -- | -- | -- | -- | --


The original usage already seems pathological as the same tests in node takes 2.3GB not 7.5GB, and this PR makes it much worse, but I don't really think this is truly an issue with the PR seems like an issue with bun IMO. Still 9% faster though 😅

Jarred-Sumner:
Yeah 7 GB is quite high

We should add some more environment variables to test things, but I wonder if passing `--smol` or running `bun build --compile` on tsc first does it reduce memory usage meaningfully? I’m wondering if the transpiler which is run on every executed file is doing something silly or if its sourcemaps (or otherwise, a runtime issue)

jakebailey:
> I’m wondering if the transpiler which is run on every executed file is doing something silly or if it’s sourcemaps (or otherwise, a runtime issue).

The benchmark in question is equivalent to running `tsc -b ./src` in our codebase, so only `tsc.js` is ever loaded (true of all benchmarks here, besides `self-build-src-public-api` which loads `typescript.js`). So something to do with transpilation or source mapping would be very surprising.

dragomirtitian:
I reverted back to the fully monomorphic version and tried to reduce the memory usage on that. The critical insight came when I realized that most of the extra memory allocated was actually not in the new objects themselves, but rather in a property array that v8 uses to store dynamic properties for objects. Creating the data object with all the properties the node might have right in the node factory means that v8 will no longer create the property array thus reducing memory consumption.

The same can be applied to Type to further reduce memory, but that seems a bit more difficult since type creation is not as centralized, but I will in investigate further if the `Node` improvements are visible in the perf tests (as they have been locally)

@jakebailey When you get a chance can you please run the tests again?

jakebailey:
@typescript-bot perf test this faster host=bun@1.1.17

dragomirtitian:
For `node` things look really really good:  Increase of 🔻 5-7.5% in memory for a perf win of 🟩14-23%. 


For `bun` things are more complicated:

-  self-build-src-public-api - bun performs much worse in terms of memory usage, and on those this PR makes things worse ().
- vscode - The best time is 87s which is very fast, but that result is not consistent

Excluding the outliers we get: Increase of 🔻 3-15% in memory for a perf win of 🟩 3-16%. A much more mixed result.

dragomirtitian:
I moved some stuff around between node and node data. This seems to reduce some more of the extra memory that was being used. 

I also experimented with reducing the size of Types, but don't think there are any easy wins there.

@jakebailey When you get a chance, can you run the tests again please?

jakebailey:
@typescript-bot perf test this faster host=bun@1.1.17

robpalme:
Good stuff. On Node the mean averages for `tsc` are:

- a total time win of **15.1%** (ranging 9.6% to 23.7%)
- a memory cost of **3.6%** (ranging 2.5% to 5.6%)

...which is an exchange rate of 4.2

MichaelMitchell-at:
> Good stuff. On Node the mean averages for `tsc` are:
> 
> * a total time win of **15.1%** (ranging 9.6% to 23.7%)
> * a memory cost of **3.6%** (ranging 2.5% to 5.6%)
> 
> ...which is an exchange rate of 4.2

It's worth keeping in mind that if TSServer requires more than 4096 MB of memory then VSCode can no longer use its prebuilt version of Node with pointer compression, which has a significant effect on interactivity performance. So it might be worth re-measuring that exchange rate on projects that are close to reaching that threshold. I plan to do this on the Airtable monorepo (which at this point has already blown past the 4GB memory limit, but we hope to eventually bring it down), once the PR is stable.

dragomirtitian:
@jakebailey Can you please run the perf tests one more time? 

I hope this latest version will bring the mem usage a bit more, either way I think this is at the limit of what I can do to reduce memory usage. 

I can cleanup the PR if there is interest in merging this based on these perf metrics. 

Beside performance, we should also consider the issue of this being a breaking change for API consumers. Anyone manually iterating over the properties of nodes (either with `Object.keys` or `for .. in`)  will observe the difference. 

Possible future improvements: Actually use `data` instead of the accessors (both in the compiler and transition API consumers as well). While the accessors seems to perform well, the fact that a lot of them end up being megamorphic hints at the fact that direct access would be faster. Direct access would be monomorphic/polymorphic  in many places (although probably not all). More experimentation would be needed to decide if this change would be worth the DX pain.

jakebailey:
I'll rerun it, though I do think there are still concerns about this particular approach.

@typescript-bot perf test this faster host=bun@1.1.17

robpalme:
Wow that's even better.  On Node the mean averages for `tsc` are now:

- a total time win of **16.0%** (ranging 10.5% to 23.7%)
- a checker time win of **20.8%** (ranging 16.6% to 26.8%)
- a memory cost of **2.7%** (ranging 1.9% to 4.4%)

...which is an exchange rate of 5.8 on total time, and 7.6 on checker time.

src/compiler/utilities.ts
```diff
@@ -8201,60 +8203,2241 @@ function Symbol(this: Symbol, flags: SymbolFlags, name: __String) {
...
+            this.data = { original: value };
+        }
+        else {
+            this.data.original = value;
+        }
+    }
+
+    get sourceText() {
+        return this.data?.sourceText;
+    }
```
dmichon-msft, line 9146: It might look a bit silly, but if any of these accessors are a particular hot path, you can make an accessor monomorphic by having a switch/case on `this.kind` and having several branches that all read e.g. `this.data.sourceText` but behind different `case` headers.

dragomirtitian: I tried that. It did not improve performance in my experiment. And I tried several different versions.. switch, if - linear search, if-binary search. All performed worse.

PR #58929, opened 2024-06-19
Implement deterministic collapse of 'await' in 'await using'

### LINKED ISSUE #58876: Explicit Resource Management normative changes for April 2024
In the June 2024 plenary session, TC39 reached consensus on https://github.com/tc39/proposal-explicit-resource-management/pull/219, which also needs to be applied in TypeScript.

rbuckton:
This updates the `__disposeResources` helper to align with a recent change to the `DisposeResources` algorithm that reached consensus in the June TC39 meeting. It has the effect of reducing extraneous Await operations for cases like the following:

```ts
{
  await using x = null, y = null, z = null;
  HAPPENS_BEFORE();
}

HAPPENS_AFTER();
```

It is a requirement that an `await using` declaration must trigger an Await when exiting the block such that `HAPPENS_BEFORE()` and `HAPPENS_AFTER()` occur in different turns, even if the resource recorded was `null`/`undefined`. Prior to this change, the `DisposeResources` algorithm would Await for each `null`/`undefined` in an `await using`, which could introduce significant delays to function execution. Rather than Await-ing three times in the above example, we can collapse these down to a single Await instead to preserve the aforementioned requirement.

I should note that our downlevel emit is not as efficient as a native implementation since we must `await` the result of `__disposeResources`, which introduces 1-2 turn overhead as it adopts the state of the resulting `Promise`. The only way we could make that more efficient would be to inline the behavior `__disposeResources` into the `finally` block, which would dramatically increase emit size. 

For more context as to the effects of the DisposeResources algorithm as a result of this change, see https://github.com/tc39/proposal-explicit-resource-management/pull/219#issuecomment-2048223310

As this changes our helpers, you can find a companion PR for tslib here: https://github.com/microsoft/tslib/pull/262

Fixes #58876

### DIFF
--- a/src/testRunner/unittests/evaluation/usingDeclarations.ts
+++ b/src/testRunner/unittests/evaluation/usingDeclarations.ts
@@ -1849,9 +1849,11 @@ describe("unittests:: evaluation:: usingDeclarations", () => {
             0,
             1,
             2,
-            4, // This really should be 2, but our transpile introduces an extra `await` by necessity to observe the
-               // result of DisposeResources. The process of adopting the result of DisposeResource ends up taking two
-               // turns of the microtask queue.
+
+            // This really should be 2, but our transpile introduces an extra `await` by necessity to observe the
+            // result of __disposeResources. The process of adopting the result ends up taking two turns of the
+            // microtask queue.
+            4,
         ]);
     });
 });
--- a/src/testRunner/unittests/evaluation/usingDeclarations.ts
+++ b/src/testRunner/unittests/evaluation/usingDeclarations.ts
@@ -1810,4 +1810,48 @@ describe("unittests:: evaluation:: usingDeclarations", () => {
             "exit",
         ]);
     });
+
+    it("deterministic collapse of Await", async () => {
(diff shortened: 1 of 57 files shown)


src/testRunner/unittests/evaluation/usingDeclarations.ts
```diff
@@ -1810,4 +1810,48 @@ describe("unittests:: evaluation:: usingDeclarations", () => {
...
+
+        await main();
+
+        assert.deepEqual(output, [
+            0,
+            1,
+            2,
+            4, // This really should be 2, but our transpile introduces an extra `await` by necessity to observe the
+               // result of DisposeResources. The process of adopting the result of DisposeResource ends up taking two
+               // turns of the microtask queue.
```
jakebailey, line 1854:
```suggestion
            // This really should be 2, but our transpile introduces an extra `await` by necessity to observe the
            // result of DisposeResources. The process of adopting the result of DisposeResource ends up taking two
            // turns of the microtask queue.
            4,
```

Formatter's not a fan of this one being offset, it seems.

iisaduan (approved):

PR #58932, opened 2024-06-19
Fix crash in isSourceFileDefaultLibrary for vscode benchmark

jakebailey:
I am not 100% certain how to test this, but the benchmark should show that this fixes things.


jakebailey:

jakebailey:
Now that the bench is running, this points out a bunch of errors that show up in vscode that did not before.

sheetalkamat (changes requested):
#58945 is correct change

PR #58935, opened 2024-06-19
Revert "Fix incorrect condition of noLib" (PR 58867)

jakebailey:
It looks like #58867 caused some regressions; vscode benchmark crashes on `--diagnostics`, and even if that crash is fixed, it fails to resolve lib and gets more errors.


jakebailey:
@typescript-bot test it

jakebailey:
Going to merge this to unblock benchmarking (per DM); I have a change locally that would also fix the original crash the PR solved, but I think there's more going on here.

Issue #58725, opened 2024-05-31
Imports without import clause don't report that the specifier can't end with `.ts`


Andarist:
### 🔎 Search Terms

import clause ts extension resolution

### 🕗 Version & Regression Information

- This is the behavior in every version I tried

### ⏯ Playground Link

https://www.typescriptlang.org/dev/bug-workbench/?module=1#code/PTAEAEDMEsBsFMB2BDAtvAXKZA6ALgM4BQR8AHgA4D2ATnqAMZWIH3KgC8oA5MtySAgwEKdFgBG+YkWipqdHjmC5C-WfPoBvAL6hINKqkXKp-IA

### 💻 Code

```ts
// @filename: a.ts

export const a = 'a'

// @filename: b.ts

import './a.ts'
import {} from './a.ts'
```

### 🙁 Actual behavior

We get one error in `b.ts`

### 🙂 Expected behavior

We should get errors on both imports in `b.ts`

jakebailey:
These imports are canonically unchecked by design. I personally would love to see this change, but IIRC this is currently intentional behavior, so people can write `import "./App.css";`.

Andarist:
I understand this is something that works as expected for a class of imports. I'm specifically reporting this as an issue with `*.ts` imports. Those have special meaning for TS and it feels weird that this works inconsistently for them here.

PR #58941, opened 2024-06-20
Add new option "noUncheckedSideEffectImports"

### LINKED ISSUE #50394: `Node16` module resolution doesn't check side effect imports

jakebailey:
When `noUncheckedSideEffectImports` is enabled, imports like:

```ts
import "something":
```

Are now resolved and can produce resolution errors.

Closes #50394
See also https://github.com/microsoft/TypeScript/issues/58725#issuecomment-2153030770

TODO:

- [x] Rename to `checkSideEffectImports` or something
- [x] Silence noImplicitAny error
- [x] Don't complain about importing scripts?

Future:

- [ ] Be okay with resolution if a file exists on disk, even if it's not JS/TS?


jakebailey:
@typescript-bot test it

jakebailey:
The user/top tests are pretty much as expected; css/sass etc. But what's more interesting are the DT results. There's a load of stuff there that is probably just plain broken on DT.

jakebailey:
100% this should not be called "bare", as I believe that actually means imports like "foo" versus "./blah" or something. Oops.

jakebailey:
@typescript-bot test top400
@typescript-bot run dt

jakebailey:
Rerunning this; I think the error deltas repos are _not_ resilient to the PR changing while jobs are still being queued. (Perf suite is)

@typescript-bot test top400
@typescript-bot run dt

jakebailey:
> Rerunning this; I think the error deltas repos are _not_ resilient to the PR changing while jobs are still being queued. (Perf suite is)

Nope, wasn't it. Back to figuring out why I can't make tests comparable to the top400.

jakebailey:
Apparently I've been searching the wrong results; the latest ones have lost the implicit any errors entirely, as expected.

jakebailey:
I am opening this up for review; we still need to bikeshed the name.

Honestly, I'm still still thinking `resolveSideEffectImports` is the right thing here. But `checkSideEffectImports` feels odd to me, but I also don't quite know the right thing either.

src/compiler/commandLineParser.ts
```diff
@@ -1192,6 +1192,14 @@ const commandOptionsWithoutBuild: CommandLineOption[] = [
         category: Diagnostics.Modules,
         description: Diagnostics.Conditions_to_set_in_addition_to_the_resolver_specific_defaults_when_resolving_imports,
     },
+    {
+        name: "resolveSideEffectImports",
+        type: "boolean",
+        affectsModuleResolution: true,
```
sheetalkamat, line 1198: i dont think you want `affecsModuleResolution` as resolution isnt changing - that is whether and how we resolve that import doesnt change? Its `affectsSemanticDiagnostics` ?

jakebailey: Ah, true; I'm not 100% certain what each of them "mean" in total. I think I just repeated the same options as for `json` but that's not the same situation at all.

jakebailey: Changed it, though no test changed, odd.

sheetalkamat: You will see that only if you add test in tsc-watch or tsserver where the option changes between edits. not till then.

jakebailey:
Good point, I should add that test.

src/compiler/commandLineParser.ts
```diff
@@ -1201,6 +1201,15 @@ const commandOptionsWithoutBuild: CommandLineOption[] = [
         category: Diagnostics.Modules,
         description: Diagnostics.Conditions_to_set_in_addition_to_the_resolver_specific_defaults_when_resolving_imports,
     },
+    {
+        name: "resolveSideEffectImports",
+        type: "boolean",
+        affectsSemanticDiagnostics: true,
+        affectsBuildInfo: true,
+        category: Diagnostics.Modules,
+        description: Diagnostics.Resolve_side_effect_imports,
```
sheetalkamat, line 1210: Resolve is very misleading here. We always resolve these irrespective of this flag right. We only "check" with this flag.

jakebailey: Sort of; we never actually called `resolveExternalModuleName` on them before.

What would you suggest instead? checkSideEffectImports? noUncheckedSideEffectImports?

sheetalkamat: But resolveExternalModuleName is just part of checking. We did resolve those imports and add the found files into the program so from resolution point of view nothing changed. 

vote for `noUncheckedSideEffectImports` because we have more of `noUnchecked` options .. but `check` would be ok too i guess ..

jakebailey:
Flag name has been updated to `noUncheckedSideEffectImports`.

src/compiler/checker.ts
```diff
@@ -4621,6 +4623,10 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
 
             if (sourceFile.symbol) {
                 if (errorNode && resolvedModule.isExternalLibraryImport && !resolutionExtensionIsTSOrJson(resolvedModule.extension)) {
+                    if (isSideEffectImport) {
+                        // This a side-effect import; ignore the implicit any.
+                        return undefined;
+                    }
                     errorOnImplicitAnyModule(/*isError*/ false, errorNode, currentSourceFile, mode, resolvedModule, moduleReference);
```
sheetalkamat, line 4630: may be let `errorOnImplicitAnyModule` take isSideEffectImport and return early instead?

src/compiler/checker.ts
```diff
@@ -4656,6 +4662,10 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 return getMergedSymbol(sourceFile.symbol);
             }
             if (errorNode && moduleNotFoundError) {
+                if (isSideEffectImport) {
+                    // This a side-effect import; ignore the fact that it's not a module.
+                    return undefined;
+                }
```
sheetalkamat, line 4668:
```suggestion
            if (errorNode && moduleNotFoundError && !isSideEffectImport) {
```

src/compiler/diagnosticMessages.json
```diff
@@ -6388,6 +6388,10 @@
         "category": "Message",
         "code": 6805
     },
+    "Resolve side effect imports.": {
```
sheetalkamat, line 6391: Fix this to make it have text "check"?

src/testRunner/unittests/tsbuildWatch/programUpdates.ts
```diff
@@ -796,4 +796,40 @@ createSomeObject().message;`,
             }),
         commandLineArgs: ["--b", "src/project", "-i", "-w"],
     });
+
+    verifyTscWatch({
```
sheetalkamat, line 800: I think `tsc --watch` test is good enough ? we can remove this one.

sheetalkamat (changes requested):

andrewbranch (approved):
Good after resolving Sheetal's comment about the command line description

src/testRunner/unittests/tsbuildWatch/programUpdates.ts
jakebailey: Sure, I'll delete it.

src/compiler/diagnosticMessages.json
jakebailey: Oops, yes!

src/compiler/checker.ts
jakebailey: You mean return early unconditionally from this branch? Right now, it falls through on purpose. Returning early breaks a lot of tests.

Maybe I shouldn't be returning `undefined` at all, just silencing errors?

jakebailey: I'm almost thinking that's true; can't users call this from the API?

src/compiler/checker.ts
jakebailey: Did this (more or less); fixed a case where the public API would misbehave with this.

RyanCavanaugh (approved):

src/compiler/checker.ts
sheetalkamat: That’s what I meant .. thanks

alecmev:
Thank you! 👍

PR #58946, opened 2024-06-20
Use concat instead of spread in concatenate

jakebailey:
Per https://github.com/microsoft/TypeScript/pull/58873#discussion_r1646769566 (Thanks @dmichon-msft)

Benchmarked these, and yeah concat is a lot faster.


jakebailey:

jakebailey:
nodejs.org outage of all things; need to look into retrying those requests.

dmichon-msft (approved):

jakebailey:
Jeeze, everything's breaking today.

dmichon-msft:
That's a remarkably large RAM reduction for such a simple change.

jakebailey:
Yeah, I'm just not sure how to rationalize the 2% slowdown in xstate (which was in both benchmark results).

jakebailey:
Hm, that one's the only statistically significant one, but the other check benchmarks also seem slightly worse.

DanielRosenwasser:
For my own curiosity, can you try a possibly-silly experiment where you effectively reimplement `concat`?

```ts
let result = [];
for (let i = 0; i < array1.length; i++) result.push(array1[i]);
for (let i = 0; i < array2.length; i++) result.push(array2[i]);
return result;
```

Andarist:
could preallocating the array's size be helpful in the suggestion above?

DanielRosenwasser:
To be honest, if you call `new Array(length)`, I don't know if the true array length is ever really respected.

jakebailey:
> could preallocating the array's size be helpful in the suggestion above?

Nope, preallocation actually doesn't help at all. When you do something like `new Array({ length: ... })` you actually get the least optimal V8 array kind. There's a patch series to V8 that would change this but my experimentation showed that it didn't actually help much.

I can try the manual push.

dmichon-msft:
From some microbenchmarking on node 18, the scaling varies wildly with array size (on smaller arrays concat shows a stronger lead than for large arrays, but I wasn't able to find any regression). Also worth pointing out that there is significant savings when concatenating more than 2 arrays to do them all at once instead of a pair at a time.

i.e.
```ts
a1.concat(a2,a3);
// or, faster still
[].concat(a1, a2, a3);
```
is much faster than
```ts
a1.concat(a2).concat(a3);
```

jakebailey:

jakebailey:
> Also worth pointing out that there is significant savings when concatenating more than 2 arrays to do them all at once instead of a pair at a time.

I tried to find an example where we concat'd more than once and didn't seem to find any, unfortunately.

dmichon-msft:
> I tried to find an example where we concat'd more than once and didn't seem to find any, unfortunately.

I find three matches for `concatenate(concatenate(` in the TypeScript codebase when searching from VSCode online.:

https://github.com/microsoft/TypeScript/blob/beb375a9cac75ba29abb707e6736e2d645ddb2f0/src/compiler/binder.ts#L1667

https://github.com/microsoft/TypeScript/blob/beb375a9cac75ba29abb707e6736e2d645ddb2f0/src/compiler/checker.ts#L5228

https://github.com/microsoft/TypeScript/blob/beb375a9cac75ba29abb707e6736e2d645ddb2f0/src/compiler/transformers/declarations.ts#L1646

Not sure if any of those callers are the hot paths for concatenate, though. That's a question for the profiler.

jakebailey:
Hm, yeah, those all may be hot.

dmichon-msft:
My largest project (90s build time) only spends 37ms in concatenate according to the profiler, so I really want to know how the perf test hit a >1% regression from this.

jakebailey:
The test case is just https://github.com/statelyai/xstate, if you're interested.

jakebailey:
Once again, 2% in xstate... Going to try the append version instead.

jakebailey:

jakebailey:
FWIW this change is totally wrong because `append` discards `undefined`... Oof

dmichon-msft:
Just for the sake of noise, can you test what happens if you make this change entirely cosmetic (e.g. just add a comment to the previous implementation) and rerun the perf test?

jakebailey:

jakebailey:
Yeah, the 2 percent was real.

ehoogeveen-medweb:
Spitballing, but could the change be causing an extra garbage collection during the test? i.e. something that would eventually happen anyway (and maybe take *more* time) but happens to trigger earlier with this pattern.

jakebailey:
That doesn't feel right, just because these all have reduced memory usage, which implies that GC is actually doing less work overall... This may take some profiling to figure out.

jakebailey:

jakebailey:
Hm, forgot about xstate being bad with the original change :(

PR #58952, opened 2024-06-20
Change reporting of diagnostics duration

gabritto:


src/server/session.ts
```diff
@@ -1319,23 +1329,40 @@ export class Session<TMessage = string> implements EventSender {
...
+        file: NormalizedPath,
+        kind: protocol.DiagnosticEventKind,
+        duration: number,
+        requestId: number,
+    ): void {
+        if (!this.requestPerformanceData.has(requestId)) {
+            this.requestPerformanceData.set(requestId, new Map());
+        }
+        const requestData = this.requestPerformanceData.get(requestId)!;
+        const fileData = requestData.get(file) || {};
```
DanielRosenwasser, line 1355:
```suggestion
        const fileData = requestData.get(file) ?? {};
```

src/server/session.ts
```diff
@@ -1411,7 +1438,12 @@ export class Session<TMessage = string> implements EventSender {
                 return next.immediate("regionSemanticCheck", () => {
                     const scriptInfo = this.projectService.getScriptInfoForNormalizedPath(fileName);
                     if (scriptInfo) {
-                        this.regionSemanticCheck(fileName, project, ranges.map(range => this.getRange({ file: fileName, ...range }, scriptInfo)));
+                        this.regionSemanticCheck(
+                            fileName,
+                            project,
+                            ranges.map(range => this.getRange({ file: fileName, ...range }, scriptInfo)),
```
DanielRosenwasser, line 1444: Call me superstitious but

```suggestion
                            ranges.map(range => this.getRange({ ...range, file: fileName }, scriptInfo)),
```

src/server/session.ts
```diff
@@ -1319,23 +1329,40 @@ export class Session<TMessage = string> implements EventSender {
                 file,
                 diagnostics: diagnostics.map(diag => formatDiag(file, project, diag)),
                 spans: spans?.map(span => toProtocolTextSpan(span, scriptInfo)),
-                duration,
             };
             this.event<protocol.DiagnosticEventBody>(
                 body,
                 kind,
             );
+            if (requestId) this.addDiagnosticsPerformanceData(file, kind, duration, requestId);
```
DanielRosenwasser, line 1337:
```suggestion
            if (requestId === undefined) {
                this.addDiagnosticsPerformanceData(file, kind, duration, requestId);
            }
```

src/server/session.ts
```diff
@@ -1068,6 +1071,12 @@ export class Session<TMessage = string> implements EventSender {
     }
 
     private sendRequestCompletedEvent(requestId: number): void {
+        if (this.requestPerformanceData.has(requestId)) {
+            const diagnosticsDuration = arrayFrom(this.requestPerformanceData.get(requestId)!.entries()).map(([file, data]) => ({ file, ...data }));
```
DanielRosenwasser, line 1075: Two things

1. `arrayFrom` takes a mapper I believe
2. I might put the spread earlier to ensure stuff is never overwritten.

```suggestion
            const diagnosticsDuration = arrayFrom(this.requestPerformanceData.get(requestId)!.entries()), ([file, data]) => ({ ...data, file }));
```

DanielRosenwasser (changes requested):

src/server/session.ts
```diff
@@ -1068,6 +1070,12 @@ export class Session<TMessage = string> implements EventSender {
     }
 
     private sendRequestCompletedEvent(requestId: number): void {
+        if (this.requestPerformanceData.has(requestId)) {
```
sheetalkamat, line 1073: I dont think you need request id in the map at all.. There cannot be another active "err" request before cancelling this one so you should be able to just use this as is just like `performanceData` field on session. 
We should just add the data in that one instead of creating new field and sending it across. so we are also sending data about update graph etc. #58845

gabritto: That makes sense. I wasn't sure if it was possible or not to end up having more than one error computation at the same time, because of it being async and cancellation, but that makes sense. I'll review #58845 and make my changes on top of that.

gabritto:
I redid this on #59122 which uses #58968 (thanks @sheetalkamat).

PR #58953, opened 2024-06-21
Update es2017.object.d.ts to indicate that Object.values/entries looks only at own properties

### LINKED ISSUE #58954: update Object.values and Object.entries docs to mention they look only at own properties
### ⚙ Compilation target

n/a

### ⚙ Library

n/a

Object.values, Object.entries (only the jsdoc descriptions)

### Sample Code

```TypeScript
n/a
```

### Documentation Link

https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_objects/Object/values

https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_objects/Object/entries

trusktr:
Closes #58954

This makes it clear that `Object.values` and `Object.entries` look for _own_ properties only.

An update is also needed for `Object.keys`.

### DIFF
--- a/src/lib/es2017.object.d.ts
+++ b/src/lib/es2017.object.d.ts
@@ -1,24 +1,24 @@
 interface ObjectConstructor {
     /**
-     * Returns an array of values of the enumerable properties of an object
+     * Returns an array of values of the enumerable own properties of an object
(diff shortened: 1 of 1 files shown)


trusktr:
Closes #58954

RyanCavanaugh (approved):

Issue #58948, opened 2024-06-20
Isolated declarations quick fix suggests `boolean` return type for predicates


bgenia:
### 🔎 Search Terms

isolated declarations, quick fix, type predicates

### 🕗 Version & Regression Information

TS 5.5+

### ⏯ Playground Link

https://www.typescriptlang.org/play/?isolatedDeclarations=true#code/GYVwdgxgLglg9mABDAzgZSgJxmA5gCgDcBDAGxAFMAuRcAazDgHcwBKRAbwChFfFMKUEJiRQAngAcKcYIhLkKiALwrEAIhRYcuNVwC+XIA

### 💻 Code

```ts
function isString(value: unknown) {
    return typeof value === "string"
}
```

### 🙁 Actual behavior

Quick fix suggests 'boolean' return type annotation

### 🙂 Expected behavior

Quick fix should suggest the inferred 'value is string` predicate annotation

RyanCavanaugh:
cc @danvk

Ankita-lal:
> ### 🔎 Search Terms
> isolated declarations, quick fix, type predicates
> 
> ### 🕗 Version & Regression Information
> TS 5.5+
> 
> ### ⏯ Playground Link
> https://www.typescriptlang.org/play/?isolatedDeclarations=true#code/GYVwdgxgLglg9mABDAzgZSgJxmA5gCgDcBDAGxAFMAuRcAazDgHcwBKRAbwChFfFMKUEJiRQAngAcKcYIhLkKiALwrEAIhRYcuNVwC+XIA
> 
> ### 💻 Code
> ```ts
> function isString(value: unknown) {
>     return typeof value === "string"
> }
> ```
> 
> ### 🙁 Actual behavior
> Quick fix suggests 'boolean' return type annotation
> 
> ### 🙂 Expected behavior
> Quick fix should suggest the inferred 'value is string` predicate annotation
> 
> ### Additional information about the issue
> _No response_

We can write the code this way to make it work,
function isString(value: unknown): boolean {
    return typeof value === "string";
}

PR #58958, opened 2024-06-21
Fixed quick fixes for inferred type predicates

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/58948

### DIFF
--- a/src/services/codefixes/fixMissingTypeAnnotationOnExports.ts
+++ b/src/services/codefixes/fixMissingTypeAnnotationOnExports.ts
@@ -874,7 +874,7 @@ function withContext<T>(
             return relativeType(node);
         }
 
-        let type: Type | undefined
+        let type: Type | undefined;
 
         if (isValueSignatureDeclaration(node)) {
             const signature = typeChecker.getSignatureFromDeclaration(node);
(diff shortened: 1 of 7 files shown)


src/services/codefixes/fixMissingTypeAnnotationOnExports.ts
```diff
@@ -1084,11 +1108,18 @@ function withContext<T>(
...
+        let isTruncated = false;
+        const result = typePredicateToAutoImportableTypeNode(typeChecker, importAdder, typePredicate, enclosingDeclaration, scriptTarget, declarationEmitNodeBuilderFlags | flags, {
+            moduleResolverHost: program,
+            trackSymbol() {
+                return true;
+            },
+            reportTruncationError() {
+                isTruncated = true;
+            },
+        });
```
jakebailey, line 1121: This looks pretty much the same as `typeToTypeNode`; why is it that we need a second set of helpers for type predicates? Why can't `typeToAutoImportableTypeNode` "just work" for these cases?

Andarist: Because type predicates are neither types nor type nodes :cry:

jakebailey: 😭

jakebailey (approved):

PR #58962, opened 2024-06-21
Make canUseWatchEvents test framework more generic so we can add more tests easily

sheetalkamat:


src/testRunner/unittests/helpers/tsserver.ts
```diff
@@ -67,6 +68,130 @@ export function patchHostTimeouts(
...
+            recordChange =>
+                session.host.watchUtils.pollingWatch(
+                    session.host.toNormalizedAbsolutePath(event.path),
+                    {
+                        cb: (fileName, eventKind) =>
+                            recordChange(
+                                event.id,
+                                session.host.windowsStyleRoot ?
+                                    path.win32.resolve(fileName) :
+                                    path.posix.resolve(fileName),
```
jakebailey, line 104: Is this path just used for debug logging? Surprised to see node's real path lib being used here, but tests do pass...

sheetalkamat: Ye this is testing part - where clients like vscode will send win32 like paths instead of normalized so this is testing only

jakebailey: Ah, okay; I think I was just worried about this resolving via a real path on disk like cwd.

jakebailey (approved):

Issue #58907, opened 2024-06-18
Module augmentation seems to break entirely a class definition/body IntelliSense autocomplete. Debug failure (exception on executing command "completionInfo").


synulux:
### 🔎 Search Terms

"debug failure completionInfo", "module augmentation breaks autocomplete classes"

### 🕗 Version & Regression Information

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about `completionInfo` debug failure, or classes IntelliSense autocomplete.
- I was unable to test this on prior versions because `@sapphire/framework` module types seems to not work below TypeScript 4.

### ⏯ Playground Link

https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAIgCbAM4GNpIHQCtUIDcAUKJLHAN5wDCEIIAhgHZJwC+cAZlPYgAKpGYMAAtgUAKYB6HoxCSA7tADWRYsXQAbRqlRwACsGYBzOgxZtJADxiTW+801ZVicd3GnTDUvXADaNAAqAEoAMnAA1HAAygYAgjQAogC6cDAQcKiSknCMAK4ZmOBaknbYcACS8NmSIPoZcMpQKnAQzFoAnnDGcEGdYJIx6FDAYPBgOp0mvPmsADRwAEaFPVxwnRD56VDdMOL6QgpZ6PaMo5m9AGoxdEiS824eXumi9j01OfXpmcwQNaIIIo8sxuggnJYEHBtLp9Ap9hAkPoAGRwMC8QawYCSVCLFbwYAArZaHBPdzEdhAA

### 💻 Code

```ts
import "discord.js";
import { Command } from "@sapphire/framework";

class PingCommand extends Command {
    // Press [CTRL + SPACE] to see autocomplete. It seems to work only in TypeScript playground, 
    // but if you try this same scenario in VSCode, then it seems to not show any "Command" class 
    // methods & properties, but it should.
    
}
```

- This seems to be a TypeScript issue and not the `@sapphire/framework` package, will explain why in additional information about the issue.

### 🙁 Actual behavior

In VSCode, pressing `CTRL + SPACE` inside the class definition doesn't show the extended class methods & properties, but it should, as it works with any other extended class.

- This happens on VSCode Insiders too.
- Tried also installing `typescript@next` and this issue still happens.

---

- **TypeScript version:** 5.4.5
- **VSCode TypeScript version:** Same as project (VSCode uses the `node_module` lib).
[...]

Issue #58756, opened 2024-06-03
Can't augment function overload is function is exported via `export {}`


eps1lon:
### 🔎 Search Terms

TS2383, function overload, module augmentation

### 🕗 Version & Regression Information

- This is a crash
- Prior to 3.2.4 you'd get "TS2484: Export declaration conflicts with exported declaration of 'get'."

### ⏯ Playground Link

https://github.com/eps1lon/tsc-augment-overload

### 💻 Code

```ts
// library.d.ts
declare function get(): string;
declare function set(value: string): void;

export { get, set };

// app.d.ts
import "library";

declare module "library" {
  function get(): string | null;
}

// index.ts
import { get, set } from "library";

// module augmentation did work
const value = get();
//    ^? string | null
// @ts-expect-error Argument of type 'string | null' is not assignable to parameter of type 'string'.
set(value);
```

### 🙁 Actual behavior

```bash
$ yarn tsc --skipLibCheck false
app.d.ts:4:12 - error TS2383: Overload signatures must all be exported or non-exported.

4   function get(): string | null;
             ~~~


Found 1 error in app.d.ts:4
```

### 🙂 Expected behavior

No error since the module augmentation seems to have worked.

### Additional information about the issue

As a workaround, we can switch to 
```diff
-function get(): string
+export function get(): string
-export {get}
```
though this seems like TypeScript is enforcing a certain code style. Some teams prefer writing a single `export {}` statement.

fatcerberus:
I’m guessing the argument is that this should work because the original library wrote `export { get }` instead of exporting functions individually.  But that’s an implementation detail - it shouldn’t be able to affect how the augmenting library is allowed to be written.

eps1lon:
>I’m guessing the argument is that this should work because the original library wrote export { get } instead of exporting functions individually. But that’s an implementation detail - it shouldn’t be able to affect how the augmenting file is allowed to be written.

I'd be fine with the constraint if we could actually write `export {}` in module augmentations but we can't. We get these two errors instead:
1. Exports and export assignments are not permitted in module augmentations.(2666)
1. Export declaration conflicts with exported declaration of 'X'.(2484)

```tsx
import * as React from 'react'

declare module 'react' {
  function useId(): string | null

  export {useId}
}
```

[...]

PR #58965, opened 2024-06-21
Fixed crash in import fixes for augmented modules

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/58907

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -2626,7 +2626,8 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
      * Note: if target is transient, then it is mutable, and mergeSymbol with both mutate and return it.
      * If target is not transient, mergeSymbol will produce a transient clone, mutate that and return it.
      */
-    function mergeSymbol(target: Symbol, source: Symbol, unidirectional = false): Symbol {
+    function mergeSymbol(target: Symbol, source: Symbol, unidirectional = false, isExports = false): Symbol {
+        const originalTarget = target;
         if (
             !(target.flags & getExcludedSymbolFlags(source.flags)) ||
             (source.flags | target.flags) & SymbolFlags.Assignment
@@ -2646,6 +2647,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     (source.flags | resolvedTarget.flags) & SymbolFlags.Assignment
                 ) {
                     target = cloneSymbol(resolvedTarget);
+                    if (isExports) {
+                        target.parent ??= originalTarget.parent;
+                    }
                 }
                 else {
                     reportMergeSymbolError(target, source);
@@ -2668,7 +2672,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             }
             if (source.exports) {
                 if (!target.exports) target.exports = createSymbolTable();
-                mergeSymbolTable(target.exports, source.exports, unidirectional);
+                mergeSymbolTable(target.exports, source.exports, unidirectional, /*isExports*/ true);
             }
             if (!unidirectional) {
                 recordMergedSymbol(target, source);
@@ -2757,10 +2761,10 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         return combined;
     }
 
-    function mergeSymbolTable(target: SymbolTable, source: SymbolTable, unidirectional = false) {
+    function mergeSymbolTable(target: SymbolTable, source: SymbolTable, unidirectional = false, isExports = false) {
         source.forEach((sourceSymbol, id) => {
             const targetSymbol = target.get(id);
-            target.set(id, targetSymbol ? mergeSymbol(targetSymbol, sourceSymbol, unidirectional) : getMergedSymbol(sourceSymbol));
+            target.set(id, targetSymbol ? mergeSymbol(targetSymbol, sourceSymbol, unidirectional, isExports) : getMergedSymbol(sourceSymbol));
         });
     }
 
@@ -5283,20 +5287,10 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         const exports = getExportsOfSymbol(container);
         const quick = exports.get(symbol.escapedName);
         if (quick && getSymbolIfSameReference(quick, symbol)) {
-            if (!quick.parent) {
-                const clone = cloneSymbol(quick);
-                clone.parent = container;
-                return clone;
-            }
             return quick;
         }
         return forEachEntry(exports, exported => {
             if (getSymbolIfSameReference(exported, symbol)) {
-                if (!exported.parent) {
-                    const clone = cloneSymbol(exported);
-                    clone.parent = container;
-                    return clone;
-                }
                 return exported;
             }
         });
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -5283,10 +5283,20 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         const exports = getExportsOfSymbol(container);
         const quick = exports.get(symbol.escapedName);
         if (quick && getSymbolIfSameReference(quick, symbol)) {
+            if (!quick.parent) {
+                const clone = cloneSymbol(quick);
+                clone.parent = container;
+                return clone;
+            }
             return quick;
         }
         return forEachEntry(exports, exported => {
             if (getSymbolIfSameReference(exported, symbol)) {
+                if (!exported.parent) {
+                    const clone = cloneSymbol(exported);
+                    clone.parent = container;
+                    return clone;
+                }
                 return exported;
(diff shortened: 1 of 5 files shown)


iisaduan:
@typescript-bot test it

Andarist:
The new error might be a positive change but I'm not entirely sure. Module augmentation is a weird beast.

Test case:
```ts
// @strict: true
// @module: nodenext
// @moduleResolution: nodenext
// @noEmit: true
// @skipLibCheck: true

// @filename: node_modules/urql/index.d.ts
type AnyVariables =
  | {
      [prop: string]: any;
    }
  | void
  | undefined;

declare function useQuery<Data = any, Variables extends AnyVariables = AnyVariables>(
  args: Variables
): Data;

export { AnyVariables, useQuery };

// @filename: src/index.ts
import { AnyVariables, useQuery } from 'urql';

declare module 'urql' {
  export function useQuery<Data = any, Variables extends AnyVariables = AnyVariables>(
    name: string,
    args: Variables
  ): Data;
}
```

The important ingredient here is `skipLibCheck: true`. The reported error gets reported with the current TS version only with `skipLibCheck: false`. With this branch it errors "consistently" with `skipLibCheck` and without it. Since the error is reported in both scenario in a `.ts` file... that's perhaps an improvement?

That said, I'm probably somewhat surprised that this is an error at all. I guess it's different to export a declaration using a modifier or using a separate export declaration but... it really feels like a detail that the user shouldn't be concerned with.

iisaduan:
> That said, I'm probably somewhat surprised that this is an error at all. I guess it's different to export a declaration using a modifier or using a separate export declaration but... it really feels like a detail that the user shouldn't be concerned with.

I agree that it's a detail that users shouldn't need to worry about. I haven't had time to fully review the PR yet, but I was able to strip your original test case even further (below). It's interesting that the completion works if you remove `Container` out of the export list in `index.d.ts`, and instead export the declaration: `export interface Container`

```
// @module: nodenext
// @Filename: ./node_modules/@sapphire/pieces/index.d.ts
//// interface Container {
////   stores: unknown;
//// }
////
//// declare class Piece {
////   container: Container;
//// }
////
//// export { Piece, type Container };

// @FileName: ./augmentation.ts
//// declare module "@sapphire/pieces" {
////   interface Container {
////     client: unknown;
////   }
//// }

// @Filename: ./index.ts
//// import { Piece } from "@sapphire/pieces";
//// class FullPiece extends Piece {
////   /*1*/
//// }
```

Andarist:
> I haven't had time to fully review the PR yet, but I was able to strip your original test case even further (below). 

Thanks. I just stopped reducing at the point that test could be easily debugged and proceeded to fixing it 😅 I added this now as an extra test case here.

> It's interesting that the completion works if in index.d.ts, you remove Container out of the export list, and instead export the declaration: export interface Container

That's likely related to the directly exported `Container` having `.exportSymbol`. The local symbol that gets exported through a separate export declaration doesn't have it.

synulux:
Is there any progress on this? Like, a dev-release that can be installed through NPM with a fix, or maybe a workaround?

gabritto:
> Is there any progress on this? Like, a dev-release that can be installed through NPM with a fix, or maybe a workaround?

Once this or another bug fix PR is merged, the [nightly build](https://www.typescriptlang.org/docs/handbook/nightly-builds.htm) of that day will contain the fix.

gabritto:
> The new error might be a positive change but I'm not entirely sure. Module augmentation is a weird beast.
> 
> Test case:
> 
> ```ts
> // @strict: true
> // @module: nodenext
> // @moduleResolution: nodenext
> // @noEmit: true
> // @skipLibCheck: true
> 
> // @filename: node_modules/urql/index.d.ts
> type AnyVariables =
>   | {
>       [prop: string]: any;
>     }
>   | void
>   | undefined;
> 
> declare function useQuery<Data = any, Variables extends AnyVariables = AnyVariables>(
>   args: Variables
> ): Data;
> 
> export { AnyVariables, useQuery };
> 
> // @filename: src/index.ts
> import { AnyVariables, useQuery } from 'urql';
> 
> declare module 'urql' {
>   export function useQuery<Data = any, Variables extends AnyVariables = AnyVariables>(
>     name: string,
>     args: Variables
>   ): Data;
> }
> ```
> 
> The important ingredient here is `skipLibCheck: true`. The reported error gets reported with the current TS version only with `skipLibCheck: false`. With this branch it errors "consistently" with `skipLibCheck` and without it. Since the error is reported in both scenario in a `.ts` file... that's perhaps an improvement?
> 
> That said, I'm probably somewhat surprised that this is an error at all. I guess it's different to export a declaration using a modifier or using a separate export declaration but... it really feels like a detail that the user shouldn't be concerned with.
> 
> I'm also not sure how to even augment this correctly since [this is not allowed](https://www.typescriptlang.org/play/?ts=5.6.0-dev.20240627#code/JYWwDg9gTgLgBAbzgQQHYE8BqBDKxsBGANgKYDOANHAKpkkCKAriVOslAOaU11MvoAlcpFR04AXzgAzKBBBwARIygBHIgoDcAWABQugCYkAxkVwk4ICPsalFytQsS640xqiMxgEVHEa9mrAA8ACLYMNhwALxw2BhUOHiEpGRwJAAeMCSo+iloWLj4xORRKBgJhckAfAAUzi4xnGQAXDwMAWyNgeVJ5FSh4ZUUdQCULbRt-EJkInQhYdjxBT1kldp6Oi7pkLCIvv78EmviukA)



> The new error might be a positive change but I'm not entirely sure. Module augmentation is a weird beast.
> 
> Test case:
> 
> ```ts
> // @strict: true
> // @module: nodenext
> // @moduleResolution: nodenext
> // @noEmit: true
> // @skipLibCheck: true
> 
> // @filename: node_modules/urql/index.d.ts
> type AnyVariables =
>   | {
>       [prop: string]: any;
>     }
>   | void
>   | undefined;
> 
> declare function useQuery<Data = any, Variables extends AnyVariables = AnyVariables>(
>   args: Variables
> ): Data;
> 
> export { AnyVariables, useQuery };
> 
> // @filename: src/index.ts
> import { AnyVariables, useQuery } from 'urql';
> 
> declare module 'urql' {
>   export function useQuery<Data = any, Variables extends AnyVariables = AnyVariables>(
>     name: string,
>     args: Variables
>   ): Data;
> }
> ```
> 
> The important ingredient here is `skipLibCheck: true`. The reported error gets reported with the current TS version only with `skipLibCheck: false`. With this branch it errors "consistently" with `skipLibCheck` and without it. Since the error is reported in both scenario in a `.ts` file... that's perhaps an improvement?
> 
> That said, I'm probably somewhat surprised that this is an error at all. I guess it's different to export a declaration using a modifier or using a separate export declaration but... it really feels like a detail that the user shouldn't be concerned with.
> 
> I'm also not sure how to even augment this correctly since [this is not allowed](https://www.typescriptlang.org/play/?ts=5.6.0-dev.20240627#code/JYWwDg9gTgLgBAbzgQQHYE8BqBDKxsBGANgKYDOANHAKpkkCKAriVOslAOaU11MvoAlcpFR04AXzgAzKBBBwARIygBHIgoDcAWABQugCYkAxkVwk4ICPsalFytQsS640xqiMxgEVHEa9mrAA8ACLYMNhwALxw2BhUOHiEpGRwJAAeMCSo+iloWLj4xORRKBgJhckAfAAUzi4xnGQAXDwMAWyNgeVJ5FSh4ZUUdQCULbRt-EJkInQhYdjxBT1kldp6Oi7pkLCIvv78EmviukA)

This new error seems "correct" in that those two `useQuery` declarations should be considered overloads of the same function, but the error itself is of course wrong, because they are both exported, even though TypeScript doesn't think so. This last part is a bug tracked by #58756.

gabritto (approved):

afern247:
having the same issue, is this ready to merge?

andrewbranch (changes requested):
Isabel and I are investigating this, but I just wanted to block this from merging for now as it’s not correct. In Isabel’s test case, under this PR, we can see the merged `Container` interface symbol has a parent of a non-merged external module symbol, when (if anything), the parent of the merged export should be the merged module. In general, when merging two symbols, we don’t want to just arbitrarily pick the parent of one of the constituents to be the sole parent of the new symbol, which is what this PR is doing.

iisaduan (commented):
I don't think this is the correct fix, for a couple reasons: 
1. If you alias the exported symbol that will be augmented (`Container` in this case), the crash still happens
2. The parent that is being assigned here is incorrect, as the merged symbol is not a descendent of the unmerged module/augmentation

I spent some time testing the module augmentation behavior and it does work correctly with the merged symbols not having parents. As a result, I don't think it makes sense to generate parents for all symbols of this type if they're not needed in typechecking.

I will open another PR that fixes autoimports directly. (#59582)

Andarist:
superseded by https://github.com/microsoft/TypeScript/pull/59582

PR #58981, opened 2024-06-23
Provide User-Friendly Message for Extended Unicode Escapes in Regular Expressions in Non-Unicode Modes

graphemecluster:
This replaces the error message of `/\u{10000}/` from `Hexadecimal digit expected.` to `Unicode escape sequences are only available when the Unicode (u) flag or the Unicode Sets (v) flag is set.` (message analogous to TS1530), as @k1tikurisu pointed out in [his article <sup>[in Japanese]</sup>](https://zenn.dev/cybozu_frontend/articles/349e9a778dfe0b#%E3%81%9D%E3%81%AE%E4%BB%96%E6%8C%99%E5%8B%95%E3%81%AE%E9%81%95%E3%81%84) that it’s hard to understand the intention of the message.


DanielRosenwasser (approved):

src/compiler/scanner.ts
```diff
@@ -1609,13 +1609,15 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
...
-                    pos < end && charCodeUnchecked(pos) === CharacterCodes.openBrace
-                ) {
+                if (pos < end && charCodeUnchecked(pos) === CharacterCodes.openBrace) {
                     // '\u{DDDDDD}'
                     pos -= 2;
-                    return scanExtendedUnicodeEscape(!!(flags & EscapeSequenceScanningFlags.ReportInvalidEscapeErrors));
+                    const result = scanExtendedUnicodeEscape(!!(flags & EscapeSequenceScanningFlags.ReportInvalidEscapeErrors));
+                    if (!(flags & EscapeSequenceScanningFlags.AllowExtendedUnicodeEscape)) {
+                        Debug.assert(flags & EscapeSequenceScanningFlags.ReportInvalidEscapeErrors);
+                        error(Diagnostics.Unicode_escape_sequences_are_only_available_when_the_Unicode_u_flag_or_the_Unicode_Sets_v_flag_is_set, start, pos - start);
```
rbuckton, line 1618:
```suggestion
                        tokenFlags |= TokenFlags.ContainsInvalidEscape;
                        if (flags & EscapeSequenceScanningFlags.ReportInvalidEscapeErrors) {
                            error(Diagnostics.Unicode_escape_sequences_are_only_available_when_the_Unicode_u_flag_or_the_Unicode_Sets_v_flag_is_set, start, pos - start);
                        }
```

Previously we would have skipped over this, set `ContainsInvalidEscape`, and returned, so I don't think the `assert` is warranted.

rbuckton (changes requested):

src/compiler/scanner.ts
graphemecluster: Well, currently in a regular expression, all errors are reported and all token flags set are ignored and reverted. Still, it seems worthwhile to apply your suggestion to prevent bugs in possible scenarios in the future.

graphemecluster: Well, currently in a regular expression, errors are always reported and all token flags set are ignored and reverted. Still, it seems worthwhile to apply your suggestion to prevent bugs in possible scenarios in the future.

rbuckton: Odd, it looks like that caused the new tests to fail. I wonder if setting `ContainsInvalidEscape` is affecting whether we actually report the error.

graphemecluster: The new test errors look correct. It’s because you merged #58982 but the previous check was done before the merge.

rbuckton (approved):

graphemecluster:
I’ve reordered the commits and did the test before and after the changes to ensure everything works as expected.
(But I don‘t understand why there’s an unrelated assertion error before the rerun.)

rbuckton (approved):

Issue #48126, opened 2022-03-04
source.organizeImports causing extra blank line


clabough:
*TS Template added by @mjbvz*

**TypeScript Version**: 4.6.2

**Search Terms**
- organize imports
- source action 

---

Issue Type: <b>Bug</b>

source.organizeImports is adding an extra blank line at the end when the typescript file only has exports. If there are any other lines (e.g. code) after the exports, then the extra line does not get added.

VS Code version: Code 1.65.0 (b5205cc8eb4fbaa726835538cd82372cc0222d43, 2022-03-02T11:12:08.962Z)
OS version: Windows_NT x64 10.0.19044
Restricted Mode: No

Extensions: none

In this project, I've created a new default Angular project and added an index.ts to show the issue.

![VSCode-OrganizeImports](https://user-images.githubusercontent.com/5217044/156839159-e81376a1-cc33-4eeb-a88a-5cd026865b91.gif)

mjbvz:
Minimal repo file:

```js
import 'fs';
```

This returns the following code edit: 

```
[Trace  - 23:49:58.180] <semantic> Response received: organizeImports (276). Request took 2 ms. Success: true 
Result: [
    {
        "fileName": "/Users/matb/projects/san/index.tsx",
        "textChanges": [
            {
                "start": {
                    "line": 1,
                    "offset": 1
                },
                "end": {
                    "line": 1,
                    "offset": 13
                },
                "newText": "import 'fs';\n"
            }
        ]
    }
]
```

RyanCavanaugh:
It seems to top out at 2 trailing newlines, so the process at least reaches a fixed point. This doesn't seem like a priority.

clabough:
If you're using the linting rule 'no-multiple-empty-lines', then saving causes a violation of that rule, and without knowing about the 'File:Save without formatting', you end up in an endless loop.

clabough:
I just found this suggestion. 

https://github.com/microsoft/TypeScript/issues/41494

Like @RyanCavanaugh said, ignoring one line would be challenging, but I think something like `// organize-imports-disable` at the top to disable for the whole file would solve 95% of the issues.

TheDevMinerTV:
Having the same issue here, it's causing us to fail our formatting checking...

Would love for this to get fixed

lortonx:
bump, my eslint conflicts with this

pflannery:
[...]

PR #58986, opened 2024-06-24
prevented organizeImports appending an extra new line to declarations when end of a file

pflannery:
relates #48126

A visual of what this PR fixes:

organizeImportBug

I've updated existing tests that were forced to add an extra line due to this bug and added three new harness fourslash tests.
The new `organizeImports 24-26` tests in this PR prove that no extra EOLs are being appended by the server.

There is still one additional issue left after this fix is applied. It occurs when having an EOL after an export declaration and executing the `organizeImport` command in vscode causes another EOL to be appended.  (double EOL as visualized in #48126)

I couldn't recreate this additional issue using the harness fourslash test engine. I can only recreate the double blank EOL when attaching to a vscode instance so my guess is that there is a bug in the vscode extension which is appending these other extra EOLs.


iisaduan (commented):
I like this change to organizeImports. I would like to test it out more, since altering the changeTracker would change the behavior of other services. @navya9singh would this affect moveToFile?

iisaduan (changes requested):
Since `textChanges` are sorted by position and non-overlapping, and we are only trying to prevent the text change at the very end of the file from extra newlines, we should be able to only perform this check on the last change.

pflannery:
> Since `textChanges` are sorted by position and non-overlapping, and we are only trying to prevent the text change at the very end of the file from extra newlines, we should be able to only perform this check on the last change.

I did try to do that originally but it's not always the last change entry. I noticed that there can be blank text entries in the change list which instruct removal of additional lines from the original source. So I ended up checking each change to be sure it would work.

iisaduan:
> I noticed that there can be blank text entries in the change list which instruct removal of additional lines from the original source. 

Do these extra removal lines only happen with organizeImports?

pflannery:
> > I noticed that there can be blank text entries in the change list which instruct removal of additional lines from the original source.
> 
> Do these extra removal lines only happen with organizeImports?

I'm not sure.

Issue #53904, opened 2023-04-18
Current type for `Array.prototype.includes` may reject valid use cases


yw662:
# lib Update Request

## Configuration Check

My compilation *target* is `ESNEXT` and my *lib* is `ESNEXT`.

## Missing / Incorrect Definition

This is `lib.es2016.array`
```ts
interface Array<T> {
    includes(searchElement: T, fromIndex?: number): boolean;
}
```

### Problem may relevant:
```ts
const a = [1, 2, 3, 'a', 'b', 'c'] as const
const b = [1, 2, 3, 4, 5, 6] as const

const n = b[Math.floor(Math.random()*6)]
a.includes(n) // type error but should pass 
```

### Suggested fix:
```ts
interface Array<T> {
    includes<I = T, E = T>(this: readonly I[], v: E, from?: number): v is I & E
}
```
And for ReadonlyArray<T> as well.

## Documentation Link
`Array.prototype.includes`, but irrelevant.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

RyanCavanaugh:
Please search for duplicates before logging new issues

fatcerberus:
@RyanCavanaugh All the dupes of this tend to get closed and duped to https://github.com/microsoft/TypeScript/issues/14520; anyone searching for open issues about this is unlikely to find that one organically.

yw662:
And I have no idea why this is considered a dup of #14520. That one has nothing to do with this since `1 | 2 | 3 | 4 | 5 | 6` is not a subtype of `1 | 2 | 3 | "a" | "b" | "c"` or the opposite. 
It is like, even if the array type `T` has no intersection with the search element type `E`, it does not always mean an error. It will always be a false but it should be like "a constant condition may indicate something wrong", not the `Array.prototype.includes` part.

RyanCavanaugh:
There are five duplicate issues with `Array.prototype.includes` in the title.

The missing type system feature to represent this would be #14520.

yw662:
I am not requesting a type system feature, I am requesting a change to the signature.

RyanCavanaugh:
The proposed signature allows *any* argument, which we consider to be too loose:
```ts
a.includes({ foo: 23 }) // should fail, but doesn't
```

In general most feedback we get is that TS should be *strict by default* and people should have to "do something" to get looser behavior. The current `includes` behavior follows that line of thinking.

[...]

PR #58988, opened 2024-06-24
Implement type narrowing for Array.prototype.includes

jun-sheaf:
This PR implements type narrowing for Array.prototype.includes.

Issue: https://github.com/microsoft/TypeScript/issues/58989


MartinJohns:
This approach has been rejected before already: https://github.com/microsoft/TypeScript/issues/53904#issuecomment-1513910169

> The proposed signature allows *any* argument, which we consider to be too loose

Issue #58345, opened 2024-04-28
Error when accessing properties on `globalThis` with the same name of a global variable


nicolo-ribaudo:
### 🔎 Search Terms

globalThis redefined error script property does not exist

### 🕗 Version & Regression Information

- This is the behavior in every version I tried

### ⏯ Playground Link

https://www.typescriptlang.org/play/?noImplicitAny=false&ts=5.5.0-beta#code/DYUwLgBAlgdgbgQ2FAJgRggXggc2AewCMkAVACygGcA6WRZdAbgChRIBPLXA44cq2vCSo0LNhABeXPEVIUadYSgBMLIA

### 💻 Code

```ts
let invalid1 = globalThis.invalid1;
let y = globalThis.invalid1;
let z = globalThis.invalid2;
```

### 🙁 Actual behavior

`Property 'invalid1' does not exist on type 'typeof globalThis'.`

### 🙂 Expected behavior

It should either complain about both `invalid1` and `invalid2`, or neither

### Additional information about the issue

It only happens in scripts and not in modules (if I add `export {}`, there are no reported errors).

RyanCavanaugh:
I think what you should actually be seeing here is a circularity error

nicolo-ribaudo:
`let invalid1 = 1` in a module does not define a `globalThis.invalid1` property, so there is no circularity going on.

nicolo-ribaudo:
`let invalid1 = 1` doesn't define `globalThis.invalid1`, so there is no circularity going on (it only happens with `var`).

Strapazzon:
Hello @nicolo-ribaudo. 

Attempting to use the variable before declaration will result in a ReferenceError. This is known as the "temporal dead zone," where the variable exists in the scope but cannot be accessed.

But, When you declare a variable with `var`, the variable declaration is "hoisted" (or "raised") to the top of your current scope. This means that, regardless of where you declared the variable in the code, JavaScript treats it as if you had declared it at the beginning of the scope. However, only the declaration is hoisted, not the initialization. Therefore, if you try to access the variable before initializing it, the value will be `undefined`, but it will not cause an error.

On the other hand, when you declare a variable with `let`, the declaration is also hoisted, but JavaScript puts these variables in a "temporal dead zone" from the start of the block until the declaration is processed. During this "dead zone", if you try to access the variable, you will receive a ReferenceError.

[...]

PR #58990, opened 2024-06-24
Don't report an error when accessing a name of a global block-scoped variable on `globalThis` outside of `noImplicitAny`

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/58345


sandersn:
Even though the bug is labelled Help Wanted, I didn't see a consensus on the change we should make here. So that's probably the first task.

Also, will this fix the bug for the person who filed it? I'm not sure whether restricting this change to `noImplicitAny: false` makes sense -- since it really is a circularity error. I personally think it should remain one, but you can make the case that removing it here is useful. (and in JS there is special-case code to ensure that this pattern *does* work).

Andarist:
It’s not a circularity. Block-scoped top-level variables define global variables but not `globalThis` properties (this is different from `var`)

andrewbranch (changes requested):
I’ve spent over a day playing with our global handling code to make more sense of this, and have decided it’s not worth it. I tried making two separate symbol tables for block-scoped globals (which don’t appear on `globalThis`) and the rest. The problem with that is that when you have a block-scoped global _value_ that merges with a global _type_ (such as any global class declaration), the type meaning is still expected to be accessible on `globalThis`, and this is sometimes the only way the type can be named:

```ts
// source
class A {}
const f = <A>(x: A) => new A();

// declaration emit
declare class A {}
declare const f: <A>(x: A) => globalThis.A;
```

So, we can’t move global class declarations out of the `globalThis` exports symbol table without seriously reworking type-space `globalThis` access, which would essentially be a hack equivalent to what we’re already doing by filtering value-space accesses by whether the symbols are block-scoped.

However, I’m a little skeptical of the fix in this PR. I’m wondering if we should limit it to `--target es2015` and later. If a user writes:

```ts
// @target: es5
// @noImplicitAny: false
let x = 0;
globalThis.x = "oops";
console.log(x.toFixed(2));
```

This will be error-free after this PR, but will crash in the browser due to downleveling the variable declaration to `var x = 0` which makes the `globalThis.x` assignment overwrite it. That’s more than just an implicit any issue.

(Separately and less importantly, I think it’s weird that the two cases here have different error messages, and that `noImplicitAny` might introduce an error that _doesn’t_ mention an implicit `any`, but I don’t see a great way to resolve those things given how weirdly tangled all this already is.)

Issue #59001, opened 2024-06-24
Warn when regular expression is used in conditional


mjbvz:
### 🔍 Search Terms

- conditional
- regular expression
-  regex

### ✅ Viability Checklist

- [x] This wouldn't be a breaking change in existing TypeScript/JavaScript code
- [x] This wouldn't change the runtime behavior of existing JavaScript code
- [x] This could be implemented without emitting different JS based on the types of the expressions
- [x] This isn't a runtime feature (e.g. library functionality, non-ECMAScript syntax with JavaScript output, new syntax sugar for JS, etc.)
- [x] This isn't a request to add a new utility type: https://github.com/microsoft/TypeScript/wiki/No-New-Utility-Types
- [x] This feature would agree with the rest of our Design Goals: https://github.com/Microsoft/TypeScript/wiki/TypeScript-Design-Goals

### ⭐ Suggestion

When a regular expression literal appears in a conditional, such as:

```ts
if (/abc?/) {
    foo()
} else {
    bar()
}
```

It would be helpful warn that `bar()` is unreachable. The programmer likely meant to call `/abc?/.test(...)`

TypeScript could help catch this easy to make mistake

### 📃 Motivating Example

We recently were bitten by this in VS Code. It slipped through code review too

### 💻 Use Cases

Help programmers catch an easy to make mistake

PR #59003, opened 2024-06-24
Disallow truthiness checks on certain literal kinds

RyanCavanaugh:
Seeing what a potential fix for #59001 turns up

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -43948,6 +43948,19 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
 
         function helper(condExpr: Expression, body: Expression | Statement | undefined) {
             const location = isLogicalOrCoalescingBinaryExpression(condExpr) ? skipParentheses(condExpr.right) : condExpr;
+
+            switch (condExpr.kind) {
+                case SyntaxKind.RegularExpressionLiteral:
+                case SyntaxKind.FunctionExpression:
+                case SyntaxKind.ObjectLiteralExpression:
+                case SyntaxKind.ClassExpression:
+                case SyntaxKind.ArrayLiteralExpression:
+                case SyntaxKind.JsxElement:
+                case SyntaxKind.JsxSelfClosingElement:
+                    error(location, Diagnostics.Truthiness_test_of_this_kind_of_expression_appears_unintentional);
+                    return;
+            }
+
             if (isModuleExportsAccessExpression(location)) {
                 return;
             }
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -43949,18 +43949,6 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         function helper(condExpr: Expression, body: Expression | Statement | undefined) {
             const location = isLogicalOrCoalescingBinaryExpression(condExpr) ? skipParentheses(condExpr.right) : condExpr;
 
-            switch (condExpr.kind) {
-                case SyntaxKind.RegularExpressionLiteral:
-                case SyntaxKind.FunctionExpression:
-                case SyntaxKind.ObjectLiteralExpression:
-                case SyntaxKind.ClassExpression:
-                case SyntaxKind.ArrayLiteralExpression:
-                case SyntaxKind.JsxElement:
-                case SyntaxKind.JsxSelfClosingElement:
-                    error(location, Diagnostics.Truthiness_test_of_this_kind_of_expression_appears_unintentional);
-                    return;
-            }
-
             if (isModuleExportsAccessExpression(location)) {
                 return;
             }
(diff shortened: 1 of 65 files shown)


RyanCavanaugh:
@typescript-bot test this

jakebailey:
@typescript-bot test it

I guess I should make that an alias, as two people have done that today...

src/compiler/checker.ts
```diff
@@ -43948,6 +43948,19 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
 
         function helper(condExpr: Expression, body: Expression | Statement | undefined) {
             const location = isLogicalOrCoalescingBinaryExpression(condExpr) ? skipParentheses(condExpr.right) : condExpr;
+
+            switch (condExpr.kind) {
+                case SyntaxKind.RegularExpressionLiteral:
+                case SyntaxKind.FunctionExpression:
```
Renegade334, line 43954: Aware this is still a draft, but function expressions will almost certainly be a mistyped IIFE, so it might be helpful to separate out a specific error message for these &ndash; maybe some sort of "Did you mean to call..." message, à la 2774?

DanielRosenwasser:
Does this test for the negated forms too?

RyanCavanaugh:
@typescript-bot pack this

src/compiler/checker.ts
Renegade334: `ArrowFunction` should probably also be in this list.

RyanCavanaugh:
`!expr` isn't considered a truthiness position today. Maybe for this check in particular it would make sense to generalize it to all boolean coercions

RyanCavanaugh:
Gonna find all the bad code now, I think

@typescript-bot test it

jakebailey:
6 new errors in vscode based on the benchmarks; it's confusing why the user tests did not say anything there, but maybe top400 will say something.

jakebailey:
Tested manually:

```
src/vs/base/common/path.ts:335:4 - error TS2869: Truthiness test of this kind of expression appears unintentional

335    `${resolvedDevice}${resolvedTail}` || '.';
       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

src/vs/editor/standalone/test/browser/monarch.test.ts:236:24 - error TS2869: Truthiness test of this kind of expression appears unintentional

236    uselessReplaceKey5: '@ham' || '',
                           ~~~~~~

src/vs/platform/files/test/browser/indexedDBFileService.integrationTest.ts:192:16 - error TS2869: Truthiness test of this kind of expression appears unintentional

192     assert.ok(!'Unexpected value ' + basename(value.resource));
                   ~~~~~~~~~~~~~~~~~~~

src/vs/platform/files/test/node/diskFileService.integrationTest.ts:276:16 - error TS2869: Truthiness test of this kind of expression appears unintentional

276     assert.ok(!'Unexpected value ' + basename(value.resource.fsPath));
                   ~~~~~~~~~~~~~~~~~~~

src/vs/platform/files/test/node/diskFileService.integrationTest.ts:320:16 - error TS2869: Truthiness test of this kind of expression appears unintentional

320     assert.ok(!'Unexpected value ' + basename(value.resource.fsPath));
                   ~~~~~~~~~~~~~~~~~~~

src/vs/workbench/contrib/terminal/browser/terminalInstance.ts:1615:14 - error TS2869: Truthiness test of this kind of expression appears unintentional

1615    tooltip: (`${exitMessage} ` ?? '') + nls.localize('launchFailed.exitCodeOnlyShellIntegration', 'Disabling shell integration in user settings might help.'),
                  ~~~~~~~~~~~~~~~~~


Found 6 errors in 5 files.

Errors  Files
     1  src/vs/base/common/path.ts:335
     1  src/vs/editor/standalone/test/browser/monarch.test.ts:236
     1  src/vs/platform/files/test/browser/indexedDBFileService.integrationTest.ts:192
     2  src/vs/platform/files/test/node/diskFileService.integrationTest.ts:276
     1  src/vs/workbench/contrib/terminal/browser/terminalInstance.ts:1615
```

RyanCavanaugh:
Collated breaks

Clearly intentional, albeit weird:
```ts
  optimizeNodes(): this | undefined {
    return `${this.code}` ? this : undefined
  }
```

Very good find -- catching an operator precedence bug. The checking to allow e.g. `while (0)` means this isn't found for the common case of `a > b ?? 0`; I'll put up a second PR
```ts
if (historyMsgLength > modelConfig?.max_tokens ?? 4000) 
```

Good find, maybe someone thought `{...null}` was `null` (it isn't)? Two instances of this in the same project
```ts
 setTheme({ ...theme_setting } || {});

 {...prevStyles} || {}
```

Very good find: This project idiomatically does `t("some str") || "fallback"` but misplaced the parens in two places:
```ts
<QuestionTip ml={1} label={publishT('QPM Tips' || '')}></QuestionTip>
// and
              label={t('support.outlink.share.Response Quote tips' || '')}
```

??. Asked about it
```ts
        <Image
          src={`https://github.com/${profile?.username}.png` || ''}
```

??. Asked about it.
```ts
    {
      layout: 'vertical' || 'flex',
      labelLayout: undefined,
      flex: false,
      className: 'flex flex-row gap-2 justify-between',
    },
```

DanielRosenwasser:
For `` `${this.code}` ? this : undefined` ``, you can just check if any TemplateExpression has a non-empty literal span.

fatcerberus:
The webpack one above looks like a bona fide bug:

```js
if (typeof value === "string" && /^[+-]?\d*(\.\d*)[eE]\d+$/)
```

This was almost certainly meant to actually run that regex on something.

jakebailey:
Yep, though given the full code is:

```js
		case "number":
			if (typeof value === "number") return value;
			if (typeof value === "string" && /^[+-]?\d*(\.\d*)[eE]\d+$/) {
				const n = +value;
				if (!isNaN(n)) return n;
			}
```

I would personally just remove the regex altogether; after all, this block says a number is being parsed, and there's already a saving `isNaN` that made this code not broken.

Issue #58740, opened 2024-06-03
Debug failure on paste with imports


mjbvz:
### 🔎 Search Terms

- paste with imports

### 🕗 Version & Regression Information

5.5.0-dev.20240602

### 💻 Code

Don't have a minimal repo yet but hitting this pretty consistently:

1. In https://github.com/microsoft/vscode-markdown-languageservice with VS Code
2. `"typescript.experimental.updateImportsOnPaste": true` with latest nightly TS
3. Copy the constructor here https://github.com/microsoft/vscode-markdown-languageservice/blob/7156aba6bdffef89b9210def0e0dda785569d1ab/src/languageFeatures/folding.ts#L26C1-L34C3
4. Paste it into a new file

### 🙁 Actual behavior

```
copyPasteController.ts:510 Error: <semantic> TypeScript Server Error (5.5.0-dev.20240602)
Debug Failure.
Error: Debug Failure.
    at Object.addImportFromExportedSymbol (/Users/matb/projects/vscode-markdown-languageservice/node_modules/typescript/lib/typescript.js:154212:32)
    at /Users/matb/projects/vscode-markdown-languageservice/node_modules/typescript/lib/typescript.js:144257:19
    at Map.forEach (<anonymous>)
    at addTargetFileImports (/Users/matb/projects/vscode-markdown-languageservice/node_modules/typescript/lib/typescript.js:144251:17)
    at /Users/matb/projects/vscode-markdown-languageservice/node_modules/typescript/lib/typescript.js:178190:7
    at ConfiguredProject2.runWithTemporaryFileUpdate (/Users/matb/projects/vscode-markdown-languageservice/node_modules/typescript/lib/typescript.js:183844:5)
    at pasteEdits (/Users/matb/projects/vscode-markdown-languageservice/node_modules/typescript/lib/typescript.js:178172:55)
    at /Users/matb/projects/vscode-markdown-languageservice/node_modules/typescript/lib/typescript.js:178158:118
    at _ChangeTracker.with (/Users/matb/projects/vscode-markdown-languageservice/node_modules/typescript/lib/typescript.js:174292:5)
    at Object.pasteEditsProvider (/Users/matb/projects/vscode-markdown-languageservice/node_modules/typescript/lib/typescript.js:178158:60)
    at Object.getPasteEdits (/Users/matb/projects/vscode-markdown-languageservice/node_modules/typescript/lib/typescript.js:149221:34)
[...]

PR #59004, opened 2024-06-24
Fixing namespace import debug failure

navya9singh:
Fixes https://github.com/microsoft/TypeScript/issues/58740
This debug failure happened when trying to resolve a namespace import. As the identifier for the import did not have a parent, an error was thrown.


andrewbranch (changes requested):
> the identifier for the import did not have a parent

This isn’t quite right—the _symbol_ the identifier for the import resolved to doesn’t have a parent, because the symbol _is the module symbol_, not an export of a module. That’s what namespace imports are—imports that bind an identifier to an entire module namespace record. So while your draft implementation looks pretty good, there are two problems: (1) `addImportFromExportedSymbol` is not the right function to contain it, since the symbol being passed in is _not_ an export, and (2) there are other ways an import/require can bind to a module symbol that are still broken. For example, I just changed in `pasteEdits_namespaceImport.ts`

```diff
- import * as test from "./a";
+ import test = require("./a");
```

and replicated the same crash. The same thing can also happen with a JavaScript require, as well as with a default import under certain conditions.[^1] The detection you need to determine if the symbol you have is for a module symbol is the `isExternalModuleSymbol` function instead of relying on looking at the syntax of `referenceImport`. However, I think you should maybe move your new code into a new ImportAdder function instead of expanding `addImportForExportedSymbol`.

[^1]: <details><summary>Test case</summary>
      
      ```ts
      /// <reference path="../fourslash.ts" />
      
      // @Filename: /folder/c.ts
      //// [||]
      
      
      // @Filename: /a.ts
      //// const abc = 10;
      //// const def = 20;
      //// export interface testInterface {
      ////     abc: number;
      ////     def: number;
      //// }
      
      // @Filename: /b.mts
      //// import test from "./a.js";
      ////
      //// [|function foo(abc: test.testInterface, def: test.testInterface) {
      ////    console.log(abc);
      ////    console.log(def);
      //// }|]
      //// 
      
      // @Filename: /tsconfig.json
      ////{ "compilerOptions": { "module": "nodenext" }, "files": ["/folder/c.ts", "a.ts", "b.mts"] }
      
      const range = test.ranges();
      verify.pasteEdits({
          args: {
              pastedText: [`function foo(abc: test.abc, def: test.def) {
      console.log(abc);
      console.log(def);
      }`],
              pasteLocations: [range[0]],
              copiedFrom: { file: "b.mts", range: [range[1]] },
          },
          newFileContents: {
              "/folder/c.ts":
      `import * as test from "../a";
      
      function foo(abc: test.abc, def: test.def) {
      console.log(abc);
      console.log(def);
      }`
          }
      });
      
      ```
      </details>

src/services/refactors/moveToFile.ts
```diff
@@ -1131,14 +1131,22 @@ export function getExistingLocals(sourceFile: SourceFile, statements: readonly S
         const declaration = importFromModuleSpecifier(moduleSpecifier);
         if (
             isImportDeclaration(declaration) && declaration.importClause &&
-            declaration.importClause.namedBindings && isNamedImports(declaration.importClause.namedBindings)
+            declaration.importClause.namedBindings
         ) {
-            for (const e of declaration.importClause.namedBindings.elements) {
-                const symbol = checker.getSymbolAtLocation(e.propertyName || e.name);
+            if (isNamespaceImport(declaration.importClause.namedBindings)) {
+                const symbol = checker.getSymbolAtLocation(declaration.importClause.namedBindings.name);
```
andrewbranch, line 1137: I think there’s no need for `checker.getSymbolAtLocation` anywhere in this function. Declarations always have a `.symbol`; you can just do `declaration.importClause.namedBindings.symbol` here, and `.symbol` on each `ImportSpecifier` below. Even the existing variable declaration / binding pattern stuff further down can lose the `getSymbolAtLocation`.

src/services/codefixes/importFixes.ts
```diff
@@ -356,6 +357,24 @@ function createImportAdderWorker(sourceFile: SourceFile | FutureSourceFile, prog
         }
     }
 
+    function addImportForExternalModuleSymbol(symbol: Symbol, originalSymbol: Symbol, isValidTypeOnlyUseSite: boolean) {
```
andrewbranch, line 360: `symbol`/`originalSymbol` are confusing names here. It looks like `originalSymbol` is the actual module symbol, and `symbol` is just an existing alias for the module that’s only being used to see what name to use. Instead, try just passing in the module symbol and the name you want to use?

src/services/codefixes/importFixes.ts
```diff
@@ -356,6 +357,24 @@ function createImportAdderWorker(sourceFile: SourceFile | FutureSourceFile, prog
         }
     }
 
+    function addImportForExternalModuleSymbol(symbol: Symbol, originalSymbol: Symbol, isValidTypeOnlyUseSite: boolean) {
+        const useRequire = shouldUseRequire(sourceFile, program);
+        if (originalSymbol.valueDeclaration) {
```
andrewbranch, line 362: What is this check for? When does an external module symbol not have a valueDeclaration?

src/services/codefixes/importFixes.ts
```diff
@@ -356,6 +357,24 @@ function createImportAdderWorker(sourceFile: SourceFile | FutureSourceFile, prog
...
+        const useRequire = shouldUseRequire(sourceFile, program);
+        if (originalSymbol.valueDeclaration) {
+            const moduleSpecifier = moduleSpecifiers.getLocalModuleSpecifierBetweenFileNames(
+                sourceFile,
+                originalSymbol.valueDeclaration.getSourceFile().fileName,
+                compilerOptions,
+                createModuleSpecifierResolutionHost(program, host),
+            );
+            const info: FixInfo = {
+                fix: { kind: ImportFixKind.AddNew, importKind: ImportKind.Namespace, addAsTypeOnly: isValidTypeOnlyUseSite ? AddAsTypeOnly.Allowed : AddAsTypeOnly.NotAllowed, useRequire, moduleSpecifierKind: undefined, moduleSpecifier },
```
andrewbranch, line 370: Since there may be multiple different forms of import that can reference a whole module, you were asking about whether it’s ok to unconditionally coerce them into a namespace import. That might be fine, but it seems like it would make sense for this function to take an optional `referenceImport` like `addImportFromExportedSymbol`, and copy its kind and type-only status where appropriate.

andrewbranch (changes requested):

src/services/refactors/moveToFile.ts
```diff
@@ -964,7 +964,7 @@ function inferNewFileName(importsFromNewFile: Map<Symbol, unknown>, movedSymbols
 function forEachReference(node: Node, checker: TypeChecker, enclosingRange: TextRange | undefined, onReference: (s: Symbol, isValidTypeOnlyUseSite: boolean) => void) {
     node.forEachChild(function cb(node) {
         if (isIdentifier(node) && !isDeclarationName(node)) {
-            if (enclosingRange && !rangeContainsRange(enclosingRange, node)) {
+            if (enclosingRange && !rangeContainsRange(enclosingRange, { pos: node.getStart(), end: node.getEnd() })) {
```
andrewbranch, line 967:
```suggestion
            if (enclosingRange && !rangeContainsRange(enclosingRange, createTextRangeFromNode(node, sourceFile))) {

```

And declare `sourceFile` outside the recursive function so it doesn’t have to be retrieved with a parent walk each time

src/services/codefixes/importFixes.ts
```diff
@@ -356,6 +358,25 @@ function createImportAdderWorker(sourceFile: SourceFile | FutureSourceFile, prog
...
+            compilerOptions,
+            createModuleSpecifierResolutionHost(program, host),
+        );
+        let info: FixInfo = {
+            fix: { kind: ImportFixKind.AddNew, importKind: ImportKind.Namespace, addAsTypeOnly: isValidTypeOnlyUseSite ? AddAsTypeOnly.Allowed : AddAsTypeOnly.NotAllowed, useRequire, moduleSpecifierKind: undefined, moduleSpecifier },
+            symbolName,
+            errorIdentifierText: undefined,
+        };
+        if (referenceImport && isImportClause(referenceImport)) {
+            info = { ...info, fix: { ...info.fix, importKind: ImportKind.Default, addAsTypeOnly: isTypeOnlyImportDeclaration(referenceImport) ? AddAsTypeOnly.Required : AddAsTypeOnly.NotAllowed } };
```
andrewbranch, line 375: I think `isValidTypeOnlyUseSite` should be consulted here in addition to the reference import. Also, what if the reference import is a type-only namespace import? It looks like we're only considering the type-onlyness of the import if it's a default import.

andrewbranch (commented):
Looks good except for the one comment

src/services/codefixes/importFixes.ts
andrewbranch: Can you explain what the rules are supposed to be for setting `addAsTypeOnly`? I don't quite understand the last commit.

navya9singh: I believe when `isValidTypeOnlyUseSite ` is true, then if `isTypeOnlyImportDeclaration(referenceImport)` is true then it should be `AddAsTypeOnly.Required` and if not, then it should be `AddAsTypeOnly.Allowed`.

andrewbranch: Yeah, and if `isValidTypeOnlyUseSite` is false, then it has to be `AddAsTypeOnly.NotAllowed`. That makes sense to me!

src/services/codefixes/importFixes.ts
```diff
@@ -356,6 +358,35 @@ function createImportAdderWorker(sourceFile: SourceFile | FutureSourceFile, prog
...
+            fix: { kind: ImportFixKind.AddNew, importKind: ImportKind.Namespace, addAsTypeOnly: isValidTypeOnlyUseSite ? AddAsTypeOnly.Allowed : AddAsTypeOnly.NotAllowed, useRequire, moduleSpecifierKind: undefined, moduleSpecifier },
+            symbolName,
+            errorIdentifierText: undefined,
+        };
+        if (referenceImport && isValidTypeOnlyUseSite && isImportClause(referenceImport)) {
+            info = {
+                ...info,
+                fix: {
+                    kind: ImportFixKind.AddNew,
+                    importKind: isNamespaceImport(referenceImport) ? ImportKind.Namespace : ImportKind.Default,
```
andrewbranch, line 379: This condition an never be true based on the if statement it's in. Also, it looks like we don't use the `referenceImport` to decide between default/namespace unless `isValidTypeOnlyUseSite` is true, which is an unrelated piece of information. I think all this code can be reduced to a single assignment to `info` with conditional properties, which would help untangle the two separate things you're trying to control.

PR #59014, opened 2024-06-25
Lowercase Microsoft github org links

weswigham:
Since ostensibly npm provenance statements are case-sensitive and will probably be unhappy with the redirect from the upper-case form eventually.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,11 +13,11 @@
         "javascript"
     ],
     "bugs": {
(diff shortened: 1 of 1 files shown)


jakebailey (approved):
This also bugged me on some other site too (can't remember which); thanks!

DanielRosenwasser (approved):

PR #59017, opened 2024-06-25
Use regular imports instead of require where possible

jakebailey:
This is pulled out of #58419, since I think this is generally better and we can do it now.

### DIFF
--- a/src/tsserver/nodeServer.ts
+++ b/src/tsserver/nodeServer.ts
@@ -182,7 +182,7 @@ export function initializeNodeSystem(): StartInput {
             if (this.fd >= 0) {
                 const buf = Buffer.from(s);
                 // eslint-disable-next-line no-restricted-syntax
-                fs.writeSync(this.fd, buf, 0, buf.length, /*position*/ null); // TODO: GH#18217
+                fs.writeSync(this.fd, buf, 0, buf.length, /*position*/ null);
             }
             if (this.traceToConsole) {
                 console.warn(s);
--- a/src/tsserver/nodeServer.ts
+++ b/src/tsserver/nodeServer.ts
@@ -1,3 +1,8 @@
+import childProcess from "child_process";
+import fs from "fs";
+import net from "net";
+import os from "os";
+import readline from "readline";
 import {
     CharacterCodes,
     combinePaths,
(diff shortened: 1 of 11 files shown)


sheetalkamat (approved):

jakebailey:

src/tsserver/nodeServer.ts
```diff
@@ -1,3 +1,8 @@
+import childProcess from "child_process";
+import fs from "fs";
+import net from "net";
+import os from "os";
+import readline from "readline";
```
DanielRosenwasser, line 5: Is there any risk of these becoming unconditional imports?

jakebailey: Nope, this is the tsserver project which only runs in node.

src/tsserver/nodeServer.ts
```diff
@@ -326,7 +277,7 @@ export function initializeNodeSystem(): StartInput {
 
     let cancellationToken: ts.server.ServerCancellationToken;
     try {
-        const factory = require("./cancellationToken");
+        const factory = require("./cancellationToken.js");
```
mmkal, line 280: Hi, I came across this line running [dependency-cruiser](https://www.npmjs.com/package/dependency-cruiser) on the typescript codebase. After building, it's the only require/import that can't be resolved. I notice that this file, `nodeServer`, does not have a `cancellationToken` sibling file - is it possible that this is always throwing (seems like it could be, since the catch block silently catches the error and uses `ts.server.nullCancellationToken`)

jakebailey: This file should exist as a sibling. https://unpkg.com/browse/typescript@5.6.3/lib/cancellationToken.js

Are you running this tool on the built code, or the source?

jakebailey: (realistically, it's not clear to me why this is a separate file at all anyway)

mmkal: Got it - thanks very much!

jakebailey: https://github.com/microsoft/TypeScript/pull/60250 you've inspired me

PR #59025, opened 2024-06-25
Option to skipSysTests since they dont change branch to branch and ca…

sheetalkamat:


src/testRunner/unittests/sys/symlinkWatching.ts
```diff
@@ -12,799 +15,801 @@ import {
     osFlavorToString,
     TestServerHostOsFlavor,
 } from "../helpers/virtualFileSystemWithWatch.js";
-describe("unittests:: sys:: symlinkWatching::", () => {
```
jakebailey, line 15: You could instead just return at the top before any of the other code runs, if you want the diff to be simpler.

sheetalkamat: Cant return from module at top level

jakebailey: Not the module, I mean the `describe` callback, i.e. never call a test function.

jakebailey (approved):

PR #59026, opened 2024-06-25
fix(59011): TypeScript generates invalid types if @import tags are spread over multiple lines

### LINKED ISSUE #59011: TypeScript generates invalid types if `@import` tags are spread over multiple lines
### 🔎 Search Terms

_“import”_

- I was unable to test this on prior versions because this feature was introduced in TypeScript 5.5

### ⏯ Playground Link

https://github.com/remcohaszing/typescript-jsdoc-import-bug

### 💻 Code

```js
/**
 * @import {
 *   Program
 * } from 'estree'
 */

/**
 * @param {Program} program
 */
export function compile(program) {}
```

### 🙁 Actual behavior

TypeScript emits the following type invalid definition:

```ts
/**
 * @import {
 *   Program
 * } from 'estree'
 */
/**
 * @param {Program} program
 */
export function compile(program: *   Program): void;
import type { Program } from 'estree';
```

### 🙂 Expected behavior

TypeScript emits the following valid type definition:

```ts
/**
 * @import {
 *   Program
 * } from 'estree'
 */
/**
 * @param {Program} program
 */
export function compile(program: Program): void;
import type { Program } from 'estree';
```

This bug does not appear in the playground, only when running `tsc` locally.

a-tarasyuk:
Fixes #59011


DanielRosenwasser:

DanielRosenwasser (approved):

DanielRosenwasser:

src/compiler/parser.ts
```diff
@@ -2635,7 +2635,7 @@ namespace Parser {
     function createIdentifier(isIdentifier: boolean, diagnosticMessage?: DiagnosticMessage, privateIdentifierDiagnosticMessage?: DiagnosticMessage): Identifier {
         if (isIdentifier) {
             identifierCount++;
-            const pos = getNodePos();
+            const pos = scanner.hasLeadingAsterisks() ? scanner.getTokenStart() : getNodePos();
```
jakebailey, line 2638: I feel like this is odd; why is the asterisk a part of the full start of node in the first place?

a-tarasyuk: @jakebailey Thanks for the feedback.

For similar JsDoc cases, the parser uses the following https://github.com/microsoft/TypeScript/blob/15f67e0b482faf9f6a3ab9965f3c11196bf3e99b/src/compiler/parser.ts#L9848-L9861 which references the start token position instead of the full start.

We could try changing `fullStartPos` to advance to token start directly in the `scanner`, however, this might cause breaking changes in certain scenarios - need to check...

The `import` tag parsing reuses as much logic as possible from import declarations to avoid duplication and accommodate differences. Essentially, it only requires a change in context for the scanner to skip leading asterisks and avoid deep args passing https://github.com/microsoft/TypeScript/blob/15f67e0b482faf9f6a3ab9965f3c11196bf3e99b/src/compiler/parser.ts#L8461-L8463

Since the `import specifier` uses `parseIdentifierName`, which calls `createIdentifier` and sets pos as the `fullStartPos` of the identifier, it causes an issue with the identifier range. We could switch to using `parseJSDocIdentifierName` instead of https://github.com/microsoft/TypeScript/blob/15f67e0b482faf9f6a3ab9965f3c11196bf3e99b/src/compiler/parser.ts#L8558, or modify `createIdentifier` directly. What do you think would work best for this case?

src/compiler/scanner.ts
```diff
@@ -4012,6 +4016,10 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
     function setSkipJsDocLeadingAsterisks(skip: boolean) {
```
sandersn, line 4016: why doesn't setSkipJsDocLeadingAsterisks suffice? Looking briefly at the parser code, it looks like it's set correctly. Is the position returned by getNodePos incorrect when it's set? In that case, probably the fix needs to be there instead of createIdentifier.

Edit: that's probably it; getNodePos returns scanner.getTokenFullStart; this PR switches to getTokenStart. But if skipJsDocLeadingAsterisks was set, it seems to me that getTokenFullStart should also skip the leading asterisk -- and return the same position as getTokenStart.

sandersn (changes requested):

src/compiler/parser.ts
jakebailey: I would think advancing it when there's an ignored asterisk is correct (what nathan said i think)

src/compiler/scanner.ts
a-tarasyuk: @sandersn Thanks for the detailed feedback 👍. I'll look into how we can add these changes to the scanner. If I remember correctly, the end position relies on `getTokenFullStart`. If we adjust `fullStartPos` after `*`, we might run into situations where `*` appears after names. or am I missing something here?

```ts
/**
 * @import {    
 *    Foo       // node pos + end (fullStartPos after *) "Foo * " ?
 * } from "./a"
 */

File: tests/cases/conformance/jsdoc/importTag18.ts [hunk 3]
```diff
@@ -0,0 +1,19 @@
...
+// @allowJs: true
+
+// @filename: a.ts
+export interface Foo {}
+
+// @filename: b.js
+/**
+ * @import {
+ *    Foo
+ * } from "./a"
```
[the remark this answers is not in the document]
a-tarasyuk: I've added tests to covers these two cases

tests/baselines/reference/importTag7.types
```diff
@@ -24,10 +24,10 @@ export interface B {
  * @param { B } b
  */
 function f(a, b) {}
->f : (a: * A, b: * B) => void
```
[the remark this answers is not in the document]
a-tarasyuk: I've added a signatureHelp test to confirm, and it seems to be working as expected.

DanielRosenwasser:
I think we will schedule this for a follow-up patch. I want to get 5.5.3 out today.

src/compiler/scanner.ts
sandersn: No, your memory is probably better than mine on this front. Still, thanks for looking into it.

sandersn:
I think this is good enough to merge once @a-tarasyuk has a chance to confirm that the scanner changes are infeasible (or, if they're feasible, to make the changes). Can you check whether adjusting fullStartPos after `*` does actually cause `*` appears after names?

a-tarasyuk:
Modifying `fullStartPos` in the scanner might lead to a `*` being added after the name or `\n*` being included in the node range.

```ts
/*HIGHLIGHTS*/<|[|{| kind: "reference" |}foo|]: string;
//  *   |>
```

The first issue can be fixed by adjusting `fullStartPos` for identifiers. However, the second case won't stop at the `;` token because `\n*` is considered a valid leading asterisk within the node range...

sandersn (approved):
Given the complexity of a better fix, let's take this with the idea of improving the scanner's ability to skip `*` in the future.

(Based on discussion between me and @jakebailey .)

@a-tarasyuk would you mind bringing this branch up to date with main?

a-tarasyuk:
@sandersn Sure. One sec I'll update

PR #59038, opened 2024-06-26
Error if the left operand of ?? is syntactically provable to not be nullish

RyanCavanaugh:
Follow-on to #59003 based on a hit that PR had

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -39377,7 +39377,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 return state;
             }
 
-            checkGrammarNullishCoalesceWithLogicalExpression(node);
+            checkNullishCoalesceOperand(node);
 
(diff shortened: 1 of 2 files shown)


RyanCavanaugh:
@typescript-bot test top400

RyanCavanaugh:
headlessui. Huh?
```ts
let isParentDisabled = parent?.getAttribute('disabled') === '' ?? false
```

supabase. Again, huh?
```ts
  const isAcknowledged =
    typeof window !== 'undefined' ? localStorage?.getItem(key) === 'true' ?? false : false
```

vscode. Looks like missing parens?
```ts
    return this.viewModel?.welcomeExperience === WelcomeExperience.ForWorkspace ?? true;
```

ChatGPT-Next-Web. Missing parens as found in prior PR
```ts
    if (historyMsgLength > modelConfig?.max_tokens ?? 4000)
```

Issue #59030, opened 2024-06-26
Unable to use top-level for-await loops with module `preserve`


clemyan:
### 🔎 Search Terms

"top-level await", "for-await", "preserve", "ts1432"

### 🕗 Version & Regression Information

- This changed between versions 5.3 and 5.4 (for `module: preserve`)

### ⏯ Playground Link

https://www.typescriptlang.org/play/?target=99&module=200&ts=5.6.0-dev.20240625#code/GYewTgBAhg7lCWAXCAKAxiAdgZ2QDwhGAgG0BdASggG8BfCAegYgFEwxwAoWBZABQ4BbeNgCmAOjCjsIADYA3USgoBuRswDyAaU6dReAA7hkdFUA

### 💻 Code

```ts
for await (const x of []) {} // Error
await Promise.resolve(); // OK

export {};
```

### 🙁 Actual behavior

The code errors with `ts1432` under `module: preserve` (regardless of `target`)

### 🙂 Expected behavior

The code compiles without error under `module: preserve` and `target: ESNext`

PR #59042, opened 2024-06-26
Fixed an issue with top-level for-await loops not being allowed with `--module preserve`

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/59030
this was missed in https://github.com/microsoft/TypeScript/pull/56785 , cc @andrewbranch


andrewbranch (approved):
Thank you!

PR #59048, opened 2024-06-27
Optimize path mapping lookups

DanielRosenwasser:
When checking if a specifier satisfies a pattern (e.g. for a path mapping entry in `paths`), it seems like TypeScript needs to check every single available pattern to see if the specifier is matched.

Here, I've slightly modified the code to create a `Set` of exact specifiers, and an array of all remaining `Pattern`s. The `Set` becomes a fast look-up, and the fallback becomes a walk across the array. A new difference is that once we've found a match, we don't bother checking any other patterns unless they have a longer prefix than the current match. That can help quickly dismiss irrelevant matches.

This also adds a caches so that we don't end up re-creating these objects over and over.

In earlier iterations of this PR, the array of patterns was sorted and lookups on patterns could optimistically be done via a binary search; however, there's some nuance there with respect to how `paths` differs from `exports`, and we don't have enough test coverage there (nothing fails if you only sort by prefix specifier length!). So for now, I've stuck with a linear search.


DanielRosenwasser:
I don't know if our perf suites have repos with lots of path mappings, but...

DanielRosenwasser:
@typescript-bot pack this

DanielRosenwasser:
In some experiments on internal codebases, this seems to shave off maybe 200-300ms depending on the project (they are using pretty big lists), but it looks like it ends up in a situation where the function gets called with two different arrays in an interleaving way. So instead of a single-slot cache, I added a `WeakMap`.

@typescript-bot test top400
@typescript-bot pack this

src/compiler/core.ts
```diff
@@ -2532,12 +2532,12 @@ export function matchedText(pattern: Pattern, candidate: string): string {
...
+export function findBestPatternMatch<T>(values: readonly T[], getPattern: (value: T) => Pattern, candidate: string, endIndex: number = values.length): T | undefined {
     let matchedValue: T | undefined;
     // use length of prefix as betterness criteria
     let longestMatchPrefixLength = -1;
 
-    for (let i = 0; i < values.length; i++) {
+    for (let i = 0; i < endIndex; i++) {
         const v = values[i];
         const pattern = getPattern(v);
         if (isPatternMatch(pattern, candidate) && pattern.prefix.length > longestMatchPrefixLength) {
```
dmichon-msft, line 2543: Wouldn't it be better to run these conditions in the reverse order? No point in running the cost of pattern matching if the pattern to be matched is a worse fit than the current best.

src/compiler/utilities.ts
```diff
@@ -10069,18 +10078,95 @@ export const emptyFileSystemEntries: FileSystemEntries = {
...
+    }
+
+    if (matchableStringSet.has(candidate)) {
+        return candidate;
+    }
+    if (sortedPatterns.length === 0) {
+        return undefined;
+    }
+
+    let index = binarySearchKey(sortedPatterns, candidate, getPatternPrefix, compareStringsCaseSensitive);
```
dmichon-msft, line 10116: In `Rush` we have to efficiently match a large array of Git file paths to project prefixes as part of hashing our input states; for that we use this class which you might find useful for this application: https://github.com/microsoft/rushstack/blob/main/libraries/rush-lib/src/logic/LookupByPath.ts

dmichon-msft: For some numbers, that tool matches 142936 unique file paths to 1279 unique path prefixes in 62 milliseconds on my codespace machine (single threaded).

DanielRosenwasser:
I think the latest change helped a bit; however, it seems like frequently this function ends up with an (effectively) identical list every single time. So I've tried adding an earlier cache. I'll see if that helps.

@typescript-bot pack this

DanielRosenwasser:
For whatever reason, I'm still seeing at least one occurrence of an identical path array being passed in; but I think this is pretty ready for review.

DanielRosenwasser:
@typescript-bot test top400

src/compiler/utilities.ts
```diff
@@ -10069,18 +10078,95 @@ export const emptyFileSystemEntries: FileSystemEntries = {
...
+            }
+        }
+
+        sortedPatterns.sort((a, b) => compareStringsCaseSensitive(a.prefix, b.prefix));
+
+        patternOrStringsCache.set(patternOrStrings, {
+            matchableStringSet,
+            sortedPatterns,
+        });
+    }
```
jakebailey, line 10107: I feel like this could be extracted out into a helper above.

DanielRosenwasser:
Something for me to look into here is that I sort by longest prefix; but I believe for export maps the longest prefix has to account for the total length of the pattern as a tie-breaker (https://github.com/nodejs/node/blob/8f39f51cbbd3b2de14b9ee896e26421cc5b20121/lib/internal/modules/esm/resolve.js#L722). I think that is subtly different from TypeScript's `paths`.

DanielRosenwasser:
@typescript-bot pack this

src/compiler/core.ts
DanielRosenwasser: In practice, what will probably happen is we'll have an empty prefix, that'll match, and then everything after will still need to be checked. It probably makes sense to walk the list backwards looking for a match in case a long match is found, and then bail out as soon as an entry with a shorter prefix length is encountered.

DanielRosenwasser:
@typescript-bot pack this

DanielRosenwasser:
The current state of this just always caches the objects and divides the entries into exact specifiers (as a `Set`) and an array of patterns as written. This still helps the internal codebase in question since about half of the entries in the huge list are non-pattern specifiers.

I'm open to re-introducing the binary search. It probably still makes sense to sort the array and iterate backwards to find a match and bail earlier when possible. But apparently we don't seem to have a test where suffix length matters in the case of `exports`, and I'd need someone to make sure I get the details there right.

DanielRosenwasser:

src/compiler/utilities.ts
```diff
@@ -10068,16 +10111,15 @@ export const emptyFileSystemEntries: FileSystemEntries = {
...
  * @internal
  */
-export function matchPatternOrExact(patternOrStrings: readonly (string | Pattern)[], candidate: string): string | Pattern | undefined {
-    const patterns: Pattern[] = [];
-    for (const patternOrString of patternOrStrings) {
-        if (patternOrString === candidate) {
-            return candidate;
-        }
+export function matchPatternOrExact(patternOrStrings: ParsedPatterns, candidate: string): string | Pattern | undefined {
+    const { matchableStringSet, patterns } = patternOrStrings;
```
weswigham, line 10115:
```suggestion
    const { matchableStringSet, patterns } = parsedPatterns;

```
just because, well, it's not an array of pattern or string anymore. The comment above this function also needs updating.

weswigham (approved):

PR #56652, opened 2023-12-03
Defer `keyof (T & U)` when constituents are generic

ahejlsberg:
With this PR we defer normalization of `keyof (T & U)` where one or both are generic types. Normalization into `keyof T | keyof U` now occurs in `getSimplifiedType` (where we similarly normalize indexed access and conditional types).

```ts
type Foo = { x: string } | undefined;
type Bar = { y: string };

type FooAndBar = Foo & Bar;

type Keys<T, U> = keyof (T & U);  // Preserved as keyof (T & U), previously normalized to keyof T | keyof U

type K1 = keyof Foo;  // never
type K2 = keyof Bar;  // "y"
type K3 = Keys<Foo, Bar>;  // "x" | "y", but previously was just "y" due to normalization
type K4 = keyof (Foo & Bar);  // "x" | "y"
```

Fixes #51331.

PR #59059, opened 2024-06-27
unconstrained type parameters consistently check for `{} | null | undefined` in `strictNullChecks`

### LINKED ISSUE #50603: No error on unconstrained type parameter in `>` comparison
# Bug Report

### 🔎 Search Terms

### ⏯ Playground Link

### 💻 Code

```ts
    let result: T | null = null;
    for (const v of it) {
            result = v;
        }
    }
    return result;
}
```
```

                               ~~~~~~
```

### 🙁 Actual behavior

I get a type error.

### 🙂 Expected behavior

No error.

iisaduan:
fixes #50603

todo: this description

### DIFF
--- a/tests/baselines/reference/voidUndefinedReduction.types
+++ b/tests/baselines/reference/voidUndefinedReduction.types
@@ -20,8 +20,8 @@ function isDefined<T>(value: T | undefined | null | void): value is T {
 >          : ^^^^^^^^^
 >value !== null : boolean
 >               : ^^^^^^^
->value : (T & {}) | null
->      : ^^^^^^^^^^^^^^^
+>value : (T & ({} | null)) | null
+>      : ^^^^^^^^^^^^^^^^^^^^^^^^
 }
 
 declare const foo: string | undefined;
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -27330,7 +27330,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     function getTypeFactsWorker(type: Type, callerOnlyNeeds: TypeFacts): TypeFacts {
         if (type.flags & (TypeFlags.Intersection | TypeFlags.Instantiable)) {
             const constraintType = getBaseConstraintOfType(type) || unknownType;
-            if (type.flags & TypeFlags.Instantiable && constraintType === unknownType) return callerOnlyNeeds;
+            if (strictNullChecks && type.flags & TypeFlags.Instantiable && constraintType === unknownType) return callerOnlyNeeds;
 
             type = constraintType
         }
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -27329,7 +27329,10 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
 
     function getTypeFactsWorker(type: Type, callerOnlyNeeds: TypeFacts): TypeFacts {
         if (type.flags & (TypeFlags.Intersection | TypeFlags.Instantiable)) {
-            type = getBaseConstraintOfType(type) || unknownType;
+            const constraintType = getBaseConstraintOfType(type) || unknownType;
+            if (type.flags & TypeFlags.Instantiable && constraintType === unknownType) return callerOnlyNeeds;
+
+            type = constraintType
         }
         const flags = type.flags;
         if (flags & (TypeFlags.String | TypeFlags.StringMapping)) {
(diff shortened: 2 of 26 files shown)


iisaduan:
@typescript-bot pack this

tests/baselines/reference/voidUndefinedReduction.types
```diff
@@ -20,8 +20,8 @@ function isDefined<T>(value: T | undefined | null | void): value is T {
 >          : ^^^^^^^^^
 >value !== null : boolean
 >               : ^^^^^^^
->value : (T & {}) | null
->      : ^^^^^^^^^^^^^^^
+>value : (T & ({} | null)) | null
```
iisaduan, line 23: this one seems weird, so I am still investigating.

iisaduan:
@typescript-bot test it

iisaduan:
With this PR, this lint check *is* necessary, as it was what was causing the self-check to fail earlier https://github.com/microsoft/TypeScript/actions/runs/9751008069/job/26911678446

DanielRosenwasser:
So one of the things that no longer works is [the following](https://www.typescriptlang.org/play/?#code/GYVwdgxgLglg9mABMAPAFQDSINKIKYAeUeYAJgM6IDWeAnnMImgHwAUcARgFYBcTWNWn2wBKRAG8AsAChEcxACc8UEAqScuAbUEBdANwyAvjJnBW4UnmAwweUlgBEACzwAbV3AciD0gPS-5QMCAP1Cw4Jl-fAUFOAUAQkQAQQUAcxAAWxIoRAZEKFoABzxEAHJyKAUbVNLEGEowOByAQ3JyGFSwZo5XEqg4RELmhWas4gVcxgListsANzwFUoA6IA):

```ts
function f<T, K extends keyof T>(obj: T, key: K) {
    return obj[key];
}
```

One of the reasons we've typically said that this is okay is because `keyof undefined` or `keyof null` is `never`; so even if it's kind of "unsafe", it's supposed to be impossible to call `f` in a reasonable way.

```ts
f(undefined, "hello");
//           ~~~~~~~
// error! Argument of type 'string' is not assignable to parameter of type 'never'.
```

I know @ahejlsberg was considering changing this for other reasons around the time that #56652 was out. Maybe he has some thoughts here or can give some pointers.

I think we can https://github.com/microsoft/TypeScript/issues/50603 well enough without disrupting the behaviors in place.

iisaduan:
> I think we can fix #50603 well enough without disrupting the behaviors in place, but I'd have to take a deeper look to help out there.

Thanks for the insight! I think so too, I will fix that. Those test cases looked alright at first, but after you brought the above up, I was reminded that the following is similarish and fine
```typescript
function f<T>(obj: T) {
    for (const key in obj) {
        obj[key]; // no errors
    }
}
```

iisaduan:
@typescript-bot test it

iisaduan:
@DanielRosenwasser interestingly, this is also a current issue (5.5.3, [playground](https://www.typescriptlang.org/play/?ts=5.5.3#code/GYVwdgxgLglg9mABMAPAFUQUwB5U2AEwGdEBvAX0QB9EwQAbe6xcAzYGMTAgGkQGksufMUQBrTAE84wRGgB8ACjgAjAFYAuOXwmSt-AJRkAUIjOIATpiggLSVWoDaugLoBuY+WPHQkWAkQAc3Q+QRw8QhJdGTklBy00HSl9I1JTcysbO0QHZyl3TyA)) when `T` explicitly `extends {} | null | undefined`: 
``` typescript
    function f<T extends {} | null | undefined , K extends keyof T>(obj: T, key: K) {
        return obj[key];
               ~~~
!!! error TS18049: 'obj' is possibly 'null' or 'undefined'.
               ~~~~~~~~
!!! error TS2536: Type 'K' cannot be used to index type '{}'.
    }
```

iisaduan:
@typescript-bot test top400

ahejlsberg:
Looks like #50603 is being addressed in #59352 (and I'll comment on that separately there). Meanwhile, in the example

```ts
function f<T, K extends keyof T>(obj: T, key: K) {
    return obj[key];  // Ok
}

function g<T extends {} | null | undefined , K extends keyof T>(obj: T, key: K) {
    return obj[key];  // Error, 'obj' is possibly 'null' or 'undefined'.
}
```

the second error really shouldn't be there. As @DanielRosenwasser points out above, for any `T` that includes `null` or `undefined`, `keyof T` is `never`, so it isn't possible to have a value of type `K` that would cause the code to fail. In fact, in an `obj[key]`, when `key` is a valid property name for `obj`, it must be because `obj` is of a non-nullable type. That said, the non-nullability check produces nicer and more understandable error messages, so we want to keep it when `obj` is of a non-generic type. I will put up a PR that fixes that.

PR #59068, opened 2024-06-28
Switch to JSDoc imports in scripts

jakebailey:
This makes things a whole lot cleaner.

### DIFF
--- a/scripts/eslint/rules/no-in-operator.cjs
+++ b/scripts/eslint/rules/no-in-operator.cjs
@@ -1,5 +1,8 @@
 const { createRule } = require("./utils.cjs");
 
+/** @import { TSESTree } from "@typescript-eslint/utils" */
+void 0;
+
 module.exports = createRule({
     name: "no-in-operator",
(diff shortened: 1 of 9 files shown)


scripts/eslint/rules/no-in-operator.cjs
```diff
@@ -1,5 +1,8 @@
 const { createRule } = require("./utils.cjs");
 
+/** @import { TSESTree } from "@typescript-eslint/utils" */
+void 0;
```
DanielRosenwasser, line 4: It's super annoying that we need this.

DanielRosenwasser (approved):
Weird question - is it fine that these all appear in CJS files? Does it do the "right" lookup strategy?

scripts/eslint/rules/no-in-operator.cjs
jakebailey: Yep; I could leave it floating but then it'd be attached to some other random declaration.

PR #58398, opened 2024-05-01
Lazy calculation of expensive file explaining diagnsotics and some caching to be used to share the diagnostic data

sheetalkamat:
- Store information about file explaining diagnostics to be generated only when program diagnostics are asked. We store this already if it's the diagnostic while we are creating program graph, we now store other compiler options errors that explain file include reasons as well as lazy diagnostics to be generated and then actually populate it when program diagnostics are queried.
- In diagnostic explaining the reason why file is in program, we are caching details as well as related info that can be reused when there are multiple diagnostics for same file. The cache is cleared after population of program diagnostics is complete so doesn't stay beyond the computation of these diagnostics.

Probably fixes
- https://github.com/microsoft/TypeScript/issues/58011
- https://github.com/typescript-eslint/typescript-eslint/issues/1192
- https://github.com/microsoft/TypeScript/issues/58274

cc: @jakebailey 
As discussed offline few days ago the real issue is that `typescript-eslint` is passing in canonical file path of tsconfig and that generates lot of errors in program because of `forceConsistentCasingInFileNames` being defaulting to `true` in later versions of typescript. I have https://github.com/typescript-eslint/typescript-eslint/pull/9042 for tsconfig issue we discussed on wednesday but i dont know if there are any other eg source files or anything that are having these issues as i didnt get to look into code that deeply

Here are some perf numbers with the repro in https://github.com/microsoft/TypeScript/issues/58011#issuecomment-2092659730 where API is used with and without config file name lower casing. Special thanks to @reduckted to give us this repro so we can actually verify this. 

When diagnostics are queried (runs `emitFilesAndReportErrors`) after creating watch program:
  | tsc      (5.4.5) | eslint     (5.4.5) | delta     (5.4.5) | tsc     (fix) | eslint     (fix) | delta     (fix) | tsc     (delta with fix) | eslint     (delta with fix)
-- | -- | -- | -- | -- | -- | -- | -- | --
Files: | 4843 | 4843 |   | 4843 | 4843 |   |   |  
Memory used: | 313744K | 1366218K | 335.46% | 334185K | 363474K | 8.76% | 6.52% | -73.40%
Program time: | 9.38s | 31.98s | 240.94% | 9.16s | 9.18s | 0.22% | -2.35% | -71.29%
Total time: | 11.22s | 33.40s | 197.68% | 10.50s | 10.72s | 2.10% | -6.42% | -67.90%

<details>
<summary>Detailed Extended Diagnostics</summary>

[...]

PR #59070, opened 2024-06-28
Delay the calculation of common source root if it would be needed when calculation dts files

sheetalkamat:
Also report error if it would be used (`declarationDir` is specified) just like we do with `outDir`

Fixes https://github.com/microsoft/TypeScript/pull/58398#issuecomment-2195913116

### DIFF
--- a/src/testRunner/unittests/tsserver/projectErrors.ts
+++ b/src/testRunner/unittests/tsserver/projectErrors.ts
@@ -1,5 +1,6 @@
 import * as ts from "../../_namespaces/ts.js";
 import { jsonToReadableText } from "../helpers.js";
+import { libContent } from "../helpers/contents.js";
 import {
     baselineTsserverLogs,
     closeFilesForSession,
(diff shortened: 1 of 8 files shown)


tests/baselines/reference/tsserver/projects/synchronizeProjectList-provides-updates-to-redirect-info-when-requested.js
```diff
@@ -717,31 +717,7 @@ Info seq  [hh:mm:ss:mss] event:
       "body": {
         "triggerFile": "/users/username/projects/project/A/tsconfig.json",
         "configFile": "/users/username/projects/project/A/tsconfig.json",
-        "diagnostics": [
```
sheetalkamat, line 720: This shows how this error was unnecessarily reported when it wont be used as the files are emitted next  to source files

src/testRunner/unittests/tsserver/projectErrors.ts
```diff
@@ -833,3 +834,29 @@ describe("unittests:: tsserver:: projectErrors:: with file rename on wsl2::", ()
...
+describe("unittests:: tsserver:: projectErrors:: dts errors when files dont belong to common root", () => {
+    [undefined, "decls"].forEach(declarationDir => {
+        it(`dts errors when files dont belong to common root${declarationDir ? " with declarationDir" : ""}`, () => {
+            const host = createServerHost({
+                "/home/src/projects/project/src/file.ts": `import { a } from "../a";`,
+                "/home/src/projects/project/a.ts": `export const a = 10;`,
+                "/home/src/projects/project/src/tsconfig.json": jsonToReadableText({
+                    compilerOptions: {
+                        composite: true,
+                        noEmit: true,
```
sheetalkamat, line 847: We use to calculate common source dir and report error if `OutDir`, `rootDir`, `mapRoot` or `sourceRoot` were specified or files would be emitted. So if you had `noEmit` we wouldnt cache this and then when getting dts diagnostics it would crash because it was over eager (or didnt cache for `declarationDir`

jakebailey (approved):

sheetalkamat:

Issue #59074, opened 2024-06-28
TS7022 (circular reference) on simple assignments in a loop


Ghabriel:
### 🔎 Search Terms

TS7022, implicitly has type 'any' because it does not have a type annotation and is referenced directly or indirectly in its own initializer, circular reference, circular type

### 🕗 Version & Regression Information

- This is the behavior in every version (3.3.3+) I tried, and I reviewed the FAQ for entries about circular references

### ⏯ Playground Link

https://www.typescriptlang.org/play/?ts=5.5.2#code/CYUwxgNghgTiAEYD2A7AzgF3hAlikAXPJjHgOYDcAUBCFhiJkSefAD7woCuEE8AvJx4RqVAGZIY8ABS0sOAfAAMFeAoA82PCAB0tFGQwALVTgDUZgJTwA3lXiJUmREdiLc+ANo4AutQc4YjIMzvxhLrDWNgC+9o7oWFAQaEgAwq5SgmAZ-vAhWIJJKemw1LFAA

### 💻 Code

```ts
declare const line: string;
let test: string | null = null;

for (let i = 0; i < line.length; i++) {
  const char = line[i];
  if (test === char) {}
  const alsoChar = char;
  test = alsoChar;
}
```

### 🙁 Actual behavior

The following error is triggered:
`'alsoChar' implicitly has type 'any' because it does not have a type annotation and is referenced directly or indirectly in its own initializer.(7022)`

### 🙂 Expected behavior

No errors. `char` is clearly a string and `alsoChar` is just a copy of it, so there shouldn't be any type confusion and/or circularity. `test` isn't even read from, except in the empty if statement.

### Additional information about the issue

* Removing the empty `if` on line 6 fixes the error.
* Replacing the `if` on line 6 by `if (test !== char) {}` also produces the error, but `if (test! < char) {}` doesn't.
* Replacing line 2 by `let test: string | undefined = undefined;` also produces the error, but `let test: string | undefined;` doesn't.

This version also reproduces the issue:
```ts
declare const line: string;
let test: boolean | null = null;

for (let i = 0; i < line.length; i++) {
  const char = line[i];
  if (test !== char) {}
  const alsoChar = char === ""; // <- using a simple comparison
  test = alsoChar;
}
```

Same for this one:
```ts
declare const line: string;
let test: string | null = null;

for (let i = 0; i < line.length; i++) {
  const char = line[i];
  if (test !== char) {}
[...]

PR #54380, opened 2023-05-24
Suppress error caused by intermediate types in `getTypeOfExpression`

ahejlsberg:
Fixes #46475.

PR #59075, opened 2024-06-29
Don't avoid caching variable types resolved using `CheckMode.TypeOnly`

Andarist:
Fixes the weird behavior observed in https://github.com/microsoft/TypeScript/issues/59074

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -11985,7 +11985,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         return false;
     }
 
-    function getTypeOfVariableOrParameterOrProperty(symbol: Symbol, checkMode?: CheckMode): Type {
+    function getTypeOfVariableOrParameterOrProperty(symbol: Symbol, checkMode = CheckMode.Normal): Type {
         const links = getSymbolLinks(symbol);
         if (!links.type) {
             const type = getTypeOfVariableOrParameterOrPropertyWorker(symbol, checkMode);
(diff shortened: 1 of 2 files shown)


src/compiler/checker.ts
```diff
@@ -11994,7 +11994,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             // to preserve this type. In fact, we need to _prefer_ that type, but it won't
             // be assigned until contextual typing is complete, so we need to defer in
             // cases where contextual typing may take place.
-            if (!links.type && !isParameterOfContextSensitiveSignature(symbol) && !checkMode) {
+            if (!links.type && !isParameterOfContextSensitiveSignature(symbol) && !(checkMode & ~CheckMode.TypeOnly)) {
```
Andarist, line 11997: this cache suppression was introduced in https://github.com/microsoft/TypeScript/pull/56753 and its original PR https://github.com/microsoft/TypeScript/pull/50586 was first created before `CheckMode.TypeOnly` was even a thing (that was introduced in https://github.com/microsoft/TypeScript/pull/54380 )

It would be an easy thing to miss either way but partially the bug was born because that PR was long-lived and a new concept was introduced to the checker in the meantime.

jakebailey:
@typescript-bot test it

JoostK:
Related: https://github.com/microsoft/TypeScript/issues/59177#issuecomment-2214761255

ahejlsberg:
The proper fix here is #59177. It reverts the changes in #56753 which is the cause of the inconsistent tooltips.

PR #59080, opened 2024-06-30
fix: undefined type id getTypeListId #59062

### LINKED ISSUE #59062: `getTypeListId` crashes with `TypeError: Cannot read properties of undefined (reading 'id')`
### 🔎 Search Terms

getTypeListId, circular reference, conditional types

- This is a crash
- This changed between versions (confirmed using the Playground)
    - ~3.6.3: does not crash
    - 3.7.5 ~ 3.8.3: throws `Debug Failure. False expression.`
    - 3.9.7 ~ 5.6.0-dev.20240628: throws `TypeError: Cannot read properties of undefined (reading 'id')`
- ~~This is the behavior in every version I tried~~, and I reviewed the FAQ for entries about "Circularity Errors May Occur In The Presence of Circularities"

### ⏯ Playground Link

https://www.typescriptlang.org/play/?#code/GYVwdgxgLglg9mABMAPAfQCoBpFoMoB8AFAJSIDeAsAFCJ2IxhQCmATsAIYTOIByIAWwCCrVhwCeKDImYAPFmAAmAZ0RhBAIzYEZ85ktUixkjDvIBfGvURRxABx4ANRAF4+goxJTO5ClRXNEAH41TTZEAC5QgS1WAgBuGktqIA

### 💻 Code

```ts
function f<_T, _S>() { // The number of type parameters need to be more than 2 to reproduce, strangely
    // Both have to be inside f
    interface NumArray<T extends number> extends Array<T> {}
    type X = NumArray<X extends {} ? number : number>;
}
```

### 🙁 Actual behavior

TypeScript throws the following error when type-checking the code above
```
      throw e;
      ^

TypeError: Cannot read properties of undefined (reading 'id')
    at getTypeListId (/Users/wasabi/workspace/typescript/crash/node_modules/typescript/lib/tsc.js:58078:34)
    at createTypeReference (/Users/wasabi/workspace/typescript/crash/node_modules/typescript/lib/tsc.js:58114:16)
    at createNormalizedTypeReference (/Users/wasabi/workspace/typescript/crash/node_modules/typescript/lib/tsc.js:59115:100)
    at instantiateTypeWorker (/Users/wasabi/workspace/typescript/crash/node_modules/typescript/lib/tsc.js:61599:63)
    at instantiateTypeWithAlias (/Users/wasabi/workspace/typescript/crash/node_modules/typescript/lib/tsc.js:61584:20)
    at instantiateType (/Users/wasabi/workspace/typescript/crash/node_modules/typescript/lib/tsc.js:61562:29)
    at resolveObjectTypeMembers (/Users/wasabi/workspace/typescript/crash/node_modules/typescript/lib/tsc.js:55820:77)
    at resolveTypeReferenceMembers (/Users/wasabi/workspace/typescript/crash/node_modules/typescript/lib/tsc.js:55843:5)
    at resolveStructuredTypeMembers (/Users/wasabi/workspace/typescript/crash/node_modules/typescript/lib/tsc.js:56666:11)
    at getSignaturesOfStructuredType (/Users/wasabi/workspace/typescript/crash/node_modules/typescript/lib/tsc.js:57363:24)
```

### 🙂 Expected behavior

For TypeScript not to crash and to display some appropriate error message (such as TS4109).

- The `number` parts in the code are probably irrelevant to the issue: we still get the same crash after replacing `number` to `string` for example.

----

Thank you.

gdavidetanca:
Fixes #59062

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -16215,9 +16215,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             const length = types.length;
             let i = 0;
             while (i < length) {
-                const startId = types[i].id;
+                const startId = types[i]?.id;
                 let count = 1;
-                while (i + count < length && types[i + count].id === startId + count) {
+                while (i + count < length && types[i + count]?.id === startId + count) {
                     count++;
                 }
                 if (result.length) {
--- a/src/compiler/utilities.ts
+++ b/src/compiler/utilities.ts
@@ -7989,7 +7989,7 @@ export function getClassLikeDeclarationOfSymbol(symbol: Symbol): ClassLikeDeclar
 
 /** @internal */
 export function getObjectFlags(type: Type): ObjectFlags {
-    return type.flags & TypeFlags.ObjectFlagsType ? (type as ObjectFlagsType).objectFlags : 0;
+    return type?.flags & TypeFlags.ObjectFlagsType ? (type as ObjectFlagsType).objectFlags : 0;
 }
 
 /** @internal */
--- a/tests/baselines/reference/getTypeListId.errors.txt
+++ b/tests/baselines/reference/getTypeListId.errors.txt
@@ -1,0 +1,11 @@
+getTypeListId.ts(3,11): error TS4109: Type arguments for 'NumArray' circularly reference themselves.
+
+
+==== getTypeListId.ts (1 errors) ====
+    function f<_T, _S>() {
+    	interface NumArray<T extends number> extends Array<T> {}
+    	type X = NumArray<X extends {} ? number : number>;
+    	         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+!!! error TS4109: Type arguments for 'NumArray' circularly reference themselves.
+    }
+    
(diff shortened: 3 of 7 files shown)


src/compiler/checker.ts
```diff
@@ -16215,9 +16215,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             const length = types.length;
             let i = 0;
             while (i < length) {
-                const startId = types[i].id;
+                const startId = types[i]?.id;
```
jakebailey, line 16218: This doesn't seem right; this array should never contain undefined. I would look for where the undefined entry is added as that would be the big, not this function which just iterates over the array of types.

src/compiler/utilities.ts
```diff
@@ -7989,7 +7989,7 @@ export function getClassLikeDeclarationOfSymbol(symbol: Symbol): ClassLikeDeclar
 
 /** @internal */
 export function getObjectFlags(type: Type): ObjectFlags {
-    return type.flags & TypeFlags.ObjectFlagsType ? (type as ObjectFlagsType).objectFlags : 0;
+    return type?.flags & TypeFlags.ObjectFlagsType ? (type as ObjectFlagsType).objectFlags : 0;
```
jakebailey, line 7992: type should not be undefined here, per the type definition; you should see where the undefined is coming from as that'd be the bug.

PR #59089, opened 2024-07-01
Fixed crash on circular local type arguments when outer ones are present too

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/59062

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -16298,7 +16298,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     function getTypeArguments(type: TypeReference): readonly Type[] {
         if (!type.resolvedTypeArguments) {
             if (!pushTypeResolution(type, TypeSystemPropertyName.ResolvedTypeArguments)) {
-                return type.target.localTypeParameters?.map(() => errorType) || emptyArray;
+                return concatenate(type.target.outerTypeParameters, type.target.localTypeParameters?.map(() => errorType) || emptyArray);
             }
             const node = type.node;
             const typeArguments = !node ? emptyArray :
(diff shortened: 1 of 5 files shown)


andrewbranch:
@typescript-bot test it

andrewbranch (approved):

src/compiler/checker.ts
```diff
@@ -16298,7 +16298,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     function getTypeArguments(type: TypeReference): readonly Type[] {
         if (!type.resolvedTypeArguments) {
             if (!pushTypeResolution(type, TypeSystemPropertyName.ResolvedTypeArguments)) {
-                return type.target.localTypeParameters?.map(() => errorType) || emptyArray;
+                return concatenate(type.target.outerTypeParameters, type.target.localTypeParameters?.map(() => errorType) || emptyArray);
```
jakebailey, line 16301: No need for the `emptyArray` bit given the parameter accepts undefined.

DanielRosenwasser (approved):

DanielRosenwasser:
Approved, though I'd also like the nits from Jake fixed up.

src/compiler/checker.ts
Andarist: it's actually kinda needed so `concatenate` doesn't return `undefined`. If I remove it get type errors here

jakebailey: Could stick it on the outside, maybe...

Andarist: I'm not sure what do you mean

jakebailey: As in `return concatenate(a, b) ?? emptyArray`

But maybe that's worse. Can't recall if this helper will make a spurious copy on an empty input array or not

Andarist:
thanks @gabritto !

PR #59106, opened 2024-07-02
Optimize `scan` with a lookup table to reduce branching

DanielRosenwasser:
This change introduces two lookup tables to the lexical scanner/tokenizer. One is a simple array-based lookup for entities in the ASCII range, and the other is a `Map` for wider codepoints.

These tables assign categories to codepoints to indicate how they should be handled. For example, all line breaks, single-line whitespace, and single-character tokens can be handled generically in their own respective fashion. Doing so allows us to perform fewer individual checks to jump into or skip over a given branch.

One additional optimization is the way in which simple single-character tokens are handled is that the token itself is encoded in the lower 8 bits the category value. This allows us to avoid any branching for such cases.

This all comes at the expense of some memory overhead, along with some overhead from fetching from the table. To avoid hitting the table too often, spaces, tabs, and carriage returns, and line feeds are currently handled separately.

### DIFF
--- a/src/compiler/scanner.ts
+++ b/src/compiler/scanner.ts
@@ -2104,7 +2104,7 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
                 return token = scanNumber();
             }
 
-            Debug.assert(tokenInfo & TokenInfo.RecognizedMisc);
+            if (!(tokenInfo & TokenInfo.RecognizedMisc)) Debug.fail(`Unhandled token category ${tokenInfo}`);
             switch (ch) {
                 case CharacterCodes.exclamation:
                     if (charCodeUnchecked(pos + 1) === CharacterCodes.equals) {
--- a/src/compiler/scanner.ts
+++ b/src/compiler/scanner.ts
@@ -2025,421 +2025,422 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
             const tokenInfo = ch < charcodeToTokenInfoCommon.length ?
                 charcodeToTokenInfoCommon[ch] :
                 charcodeToTokenInfoUncommon.get(ch) ?? TokenInfo.None;
-            if (tokenInfo !== TokenInfo.None) {
-                if (tokenInfo & TokenInfo.Whitespace) {
-                    if (skipTrivia) {
+
+            if (tokenInfo === TokenInfo.None) {
+                const identifierKind = scanIdentifier(ch, languageVersion);
+                if (identifierKind) {
+                    return token = identifierKind;
+                }
+                else if (isWhiteSpaceSingleLine(ch)) {
+                    pos += charSize(ch);
+                    continue;
+                }
+                else if (isLineBreak(ch)) {
+                    tokenFlags |= TokenFlags.PrecedingLineBreak;
+                    pos += charSize(ch);
+                    continue;
+                }
+                const size = charSize(ch);
+                error(Diagnostics.Invalid_character, pos, size);
+                pos += size;
+                return token = SyntaxKind.Unknown;
+            }
+
+            if (tokenInfo & TokenInfo.Whitespace) {
+                if (skipTrivia) {
+                    pos++;
+                    continue;
+                }
+                else {
+                    while (pos < end && isWhiteSpaceSingleLine(charCodeUnchecked(pos))) {
                         pos++;
+                    }
+                    return token = SyntaxKind.WhitespaceTrivia;
+                }
+            }
+
+            if (tokenInfo & TokenInfo.SimpleToken) {
+                pos++;
+                return token = tokenInfo & TokenInfo.SimpleTokenMask;
+            }
+
+            if (tokenInfo & TokenInfo.Digit) {
+                if (ch === CharacterCodes._0) {
+                    if (pos + 2 < end && (charCodeUnchecked(pos + 1) === CharacterCodes.X || charCodeUnchecked(pos + 1) === CharacterCodes.x)) {
+                        pos += 2;
+                        tokenValue = scanMinimumNumberOfHexDigits(1, /*canHaveSeparators*/ true);
+                        if (!tokenValue) {
+                            error(Diagnostics.Hexadecimal_digit_expected);
+                            tokenValue = "0";
+                        }
+                        tokenValue = "0x" + tokenValue;
+                        tokenFlags |= TokenFlags.HexSpecifier;
+                        return token = checkBigIntSuffix();
+                    }
+                    else if (pos + 2 < end && (charCodeUnchecked(pos + 1) === CharacterCodes.B || charCodeUnchecked(pos + 1) === CharacterCodes.b)) {
+                        pos += 2;
+                        tokenValue = scanBinaryOrOctalDigits(/* base */ 2);
+                        if (!tokenValue) {
+                            error(Diagnostics.Binary_digit_expected);
+                            tokenValue = "0";
+                        }
+                        tokenValue = "0b" + tokenValue;
+                        tokenFlags |= TokenFlags.BinarySpecifier;
+                        return token = checkBigIntSuffix();
+                    }
+                    else if (pos + 2 < end && (charCodeUnchecked(pos + 1) === CharacterCodes.O || charCodeUnchecked(pos + 1) === CharacterCodes.o)) {
+                        pos += 2;
+                        tokenValue = scanBinaryOrOctalDigits(/* base */ 8);
+                        if (!tokenValue) {
+                            error(Diagnostics.Octal_digit_expected);
+                            tokenValue = "0";
+                        }
+                        tokenValue = "0o" + tokenValue;
+                        tokenFlags |= TokenFlags.OctalSpecifier;
+                        return token = checkBigIntSuffix();
+                    }
+                }
+
+                return token = scanNumber();
+            }
+
+            Debug.assert(tokenInfo & TokenInfo.RecognizedMisc);
+            switch (ch) {
+                case CharacterCodes.exclamation:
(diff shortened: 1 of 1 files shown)


DanielRosenwasser:
It's probably faster to actually check the common case of spaces, tabs, and newlines, but the current draft always does the lookup.

DanielRosenwasser:

DanielRosenwasser:

DanielRosenwasser:

DanielRosenwasser:
I'm actually not sure if that's much better or not, it could just be a better run.

@typescript-bot pack this

DanielRosenwasser:
@typescript test it
@typescript-bot pack this

DanielRosenwasser:
@typescript-bot test top400

DanielRosenwasser:
@dragomirtitian maybe this balances out the parse-time hit from #58928. 😄

DanielRosenwasser:

src/compiler/scanner.ts
```diff
@@ -1882,69 +1996,115 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
...
+                        tokenFlags |= TokenFlags.OctalSpecifier;
+                        return token = checkBigIntSuffix();
                     }
+                }
+
+                return token = scanNumber();
+            }
+
+            if (!(tokenCategory & TokenCategory.RecognizedMisc)) Debug.fail(`Unhandled token category ${tokenCategory}`);
+            switch (ch) {
```
rbuckton, line 2107: I wonder if the `switch` here is optimized in any way or if it is doing case-by-case comparisons against each character code. If the latter, maybe it would be worthwhile to analyze what that most common character codes in JS are and change the order of the cases so that the more frequent cases are earlier in the list. I expect the perf improvement here actually comes from handling digit and identifier characters earlier than the `switch`, since those are the most frequent characters in a given file. If that's the case, some of the complexity and memory overhead could be reduced by just optimizing for those two cases.

src/compiler/scanner.ts
```diff
@@ -1882,69 +1996,115 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
...
+
+            if (tokenCategory & TokenCategory.LineBreak) {
+                tokenFlags |= TokenFlags.PrecedingLineBreak;
+                pos += charSize(ch);
+                continue;
+            }
+
+            if (tokenCategory & TokenCategory.SimpleToken) {
+                pos++;
+                return token = tokenCategory & TokenCategory.SimpleTokenMask;
```
rbuckton, line 2063: A comment here would be good for future readers as it isn't obvious at first glance how a `TokenCategory` becomes a token.

src/compiler/scanner.ts
DanielRosenwasser: I think it is typically case-by-case comparisons. You're absolutely right that doing the checks earlier on for common classes; however, I think that checking for all the identifier starts isn't practical (or requires the same sort of fast-path/slow-path separation where we check for ranges), so that's why the token categorization is helpful here. By dismissing all the other types of categories, we can immediately launch into the fast path.

src/compiler/scanner.ts
```diff
@@ -301,6 +301,120 @@ const regExpFlagToFirstAvailableLanguageVersion = new Map<RegularExpressionFlags
...
+        [CharacterCodes.greaterThan, TokenCategory.RecognizedMisc],
+        [CharacterCodes.question, TokenCategory.RecognizedMisc],
+        [CharacterCodes.caret, TokenCategory.RecognizedMisc],
+        [CharacterCodes.bar, TokenCategory.RecognizedMisc],
+        [CharacterCodes.backslash, TokenCategory.RecognizedMisc],
+        [CharacterCodes.hash, TokenCategory.RecognizedMisc],
+        [CharacterCodes.replacementCharacter, TokenCategory.RecognizedMisc],
+    ]
+) {
+    if (key < tokenCategoryLookup.length) {
```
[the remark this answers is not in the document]
DanielRosenwasser: Can you be more specific? In the first two commits, I split this out from a map to an array and map to avoid the map lookup.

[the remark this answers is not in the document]
rbuckton: I mean that, since we know ahead of time which characters belong in the `Map`, we can just construct the map with those character codes rather than looping and comparing. It's not that important, and if it makes more sense to have them visually grouped together, that's perfectly fine.

src/compiler/scanner.ts
```diff
@@ -1882,69 +1996,115 @@ export function createScanner(languageVersion: ScriptTarget, skipTrivia: boolean
...
+                        if (!tokenValue) {
+                            error(Diagnostics.Octal_digit_expected);
+                            tokenValue = "0";
                         }
-                        return token = SyntaxKind.WhitespaceTrivia;
+                        tokenValue = "0o" + tokenValue;
+                        tokenFlags |= TokenFlags.OctalSpecifier;
+                        return token = checkBigIntSuffix();
                     }
+                }
```
dmichon-msft, line 2101:
```suggestion
                if (ch === CharacterCodes._0 && pos + 2 < end) {
                    switch (charCodeUnchecked(pos + 1)) {
                    case CharacterCodes.X:
                    case CharacterCodes.x:
                        pos += 2;
                        tokenValue = scanMinimumNumberOfHexDigits(1, /*canHaveSeparators*/ true);
                        if (!tokenValue) {
                            error(Diagnostics.Hexadecimal_digit_expected);
                            tokenValue = "0";
                        }
                        tokenValue = "0x" + tokenValue;
                        tokenFlags |= TokenFlags.HexSpecifier;
                        return token = checkBigIntSuffix();
                    case CharacterCodes.B:
                    case CharacterCodes.b:
                        pos += 2;
                        tokenValue = scanBinaryOrOctalDigits(/* base */ 2);
                        if (!tokenValue) {
                            error(Diagnostics.Binary_digit_expected);
                            tokenValue = "0";
                        }
                        tokenValue = "0b" + tokenValue;
                        tokenFlags |= TokenFlags.BinarySpecifier;
                        return token = checkBigIntSuffix();
                    case CharacterCodes.O:
                    case CharacterCodes.o:
                        pos += 2;
                        tokenValue = scanBinaryOrOctalDigits(/* base */ 8);
                        if (!tokenValue) {
                            error(Diagnostics.Octal_digit_expected);
                            tokenValue = "0";
                        }
                        tokenValue = "0o" + tokenValue;
                        tokenFlags |= TokenFlags.OctalSpecifier;
                        return token = checkBigIntSuffix();
                }
```

src/compiler/scanner.ts
dmichon-msft: Assuming `CharacterCodes` is a `const enum`, the VM should be able to optimize the switch case quite heavily as a branch table, but that's a question for V8.

rbuckton: If the majority of identifier/keyword characters are ASCII characters, then it's fairly trivial to check for those major cases without an array or `Map` using a simple comparison:

```ts
if (
    (ch | 32) >= CharacterCodes.a && (ch | 32) <= CharacterCodes.z ||
    ch === CharacterCodes._ ||
    ch === CharacterCodes.$
) { /* [A-Za-z_$] */ }
```

If the array/map lookups are faster than this for the identifier case, then by all means keep it as is. I only point this out as you explicitly avoid hitting the table for line terminators and spaces. If that was an important optimization, then it might make sense to do the same kind of optimization for the most frequent (and easy to optimize for) characters in a source file, e.g., ASCII letters.

rbuckton: > Assuming `CharacterCodes` is a `const enum`, the VM should be able to optimize the switch case quite heavily as a branch table, but that's a question for V8.

We've generally found that not to be the case, which is why we've had to implement our own jump tables for functions like `forEachChild`. I'm not sure what heuristics V8 uses for `switch` optimization, if any.

src/compiler/scanner.ts
DanielRosenwasser: It's hard to parse out what the suggestion is - can you explain it?

src/compiler/scanner.ts
DanielRosenwasser: > If the majority of identifier/keyword characters are ASCII characters, then it's fairly trivial to check for those major cases without an array or Map using a simple comparison:

Fair, and this is what @ahejlsberg also suggested. The same way whitespace is prioritized, ASCII range is also very likely and worth prioritizing as a quicker up-front case instead of hitting the lookup table. It would mean there's some code duplication but maybe it's fine.

src/compiler/scanner.ts
dmichon-msft: Ah, that's fair, I don't know why GitHub insists on showing the old code in the diff rather than diffing against your PR version of the code.

Suggestion is to promote the `pos + 2 < end` check to the same `if` that checks for `CharacterCodes._0`, since all branches beneath it use that condition, then replace the nested if statements with a single `switch (charCodeUnchecked(pos + 1))` with grouped case statements for each of the radix markers.

src/compiler/scanner.ts
DanielRosenwasser: >> Assuming `CharacterCodes` is a `const enum`, the VM should be able to optimize the switch case quite heavily as a branch table, but that's a question for V8.

> We've generally found that not to be the case, which is why we've had to implement our own jump tables for functions like `forEachChild`. I'm not sure what heuristics V8 uses for `switch` optimization, if any.

Especially in the case where these character codes are so sparse - I am very doubtful that V8 would do such an optimization, and at best I'd guess we'd end up with something like a binary search.

DanielRosenwasser:
#59244 tries to just check for identifiers up-front, and moves checks for `space` earlier. It is faster, but not as fast as the lookup approach here.

PR #59120, opened 2024-07-02
fix(58772): Duplicate exports.* = assignments in CommonJS output in some cases

### LINKED ISSUE #58772: Duplicate exports.* = assignments in CommonJS output in some cases
### 🔎 Search Terms

commonjs, exports, function

- This changed in https://github.com/microsoft/TypeScript/pull/58489.

### ⏯ Playground Link

Thanks for merging   I'm still seeing an odd behavior I think is related: here's the minimal repro I could find - [bug workbench link](https://www.typescriptlang.org/dev/bug-workbench/?target=1&module=1#code/PTAEAEGcAsHsHcCiBbAlgF3QUwCYDFUAbLALlGQENUA7AOnUgCgQJlYcBXYsgY1mTbUAVk2ZhwAMyJZqFZKXJU6DRqmQAHWACd0oAN6g8sWKAC+oCVv6gARLWBTiAWngZoTrAA9NOm4y8+uhIc1DzoqLDUoDxaWBTYRrAAFACUZIn6jKDZoLHoHFpReqYA3Iym-t7aunoxcQnGoBSQ0bHxWIkAagBMpYxiEI4ycgpDLm4eVTr0TAHVoDTYWhIUPFiGjcX9QA)  

### 💻 Code

```ts
// @showEmittedFile: main.ts
// @module: commonjs

// @filename: main.ts
import { Foo } from "./file-with-export"
export function createFoo(): Foo {
    return {};
}
export {createFoo as createFooV2};

// @filename: file-with-export.ts
export interface Foo {}
```

### 🙁 Actual behavior

main.js has two exports for `createFoo` and `createFooV2`:

``` ts
"use strict";
exports.createFoo = createFoo;
exports.createFooV2 = createFoo;
exports.createFoo = createFoo;
exports.createFooV2 = createFoo;
function createFoo() {
    return {};
}
```

### 🙂 Expected behavior

``` ts
"use strict";
exports.createFoo = createFoo;
exports.createFooV2 = createFoo;
function createFoo() {
    return {};
}
```

I believe this is a result of the fix for https://github.com/microsoft/TypeScript/issues/58473. (I still appreciate having these duplicate exports rather than the compiler dropping the exports entirely 😃 )

In the debugger, I see that this `exportedFunctions` (https://github.com/microsoft/TypeScript/blob/506f3e26c3c8af7e4de38996656c947050e68567/src/compiler/transformers/utilities.ts#L108) set contains two separate `createFoo` nodes.

If I remove the `export {createFoo as createFooV2}` from the repro, or remove the declared return type `: Foo`, I get the expected behavior.

This was uncovered by testing the latest TS version in Google (https://github.com/Microsoft/TypeScript/issues/58685)

a-tarasyuk:
Fixes #58772

### DIFF
--- a/src/compiler/transformers/utilities.ts
+++ b/src/compiler/transformers/utilities.ts
@@ -47,6 +47,7 @@ import {
     isClassStaticBlockDeclaration,
     isDefaultImport,
     isExpressionStatement,
+    isFunctionDeclaration,
     isGeneratedIdentifier,
     isGeneratedPrivateIdentifier,
     isIdentifier,
@@ -321,7 +322,7 @@ export function collectExternalModuleInfo(context: TransformationContext, source
     }
 
     function addExportedFunctionDeclaration(node: FunctionDeclaration, name: ModuleExportName | undefined, isDefault: boolean) {
-        exportedFunctions.add(node);
+        exportedFunctions.add(getOriginalNode(node, isFunctionDeclaration));
         if (isDefault) {
             // export default function() { }
             // function x() { } + export { x as default };
--- a/tests/baselines/reference/exportsAndImports5.js
+++ b/tests/baselines/reference/exportsAndImports5.js
@@ -1,0 +1,24 @@
+//// [tests/cases/conformance/es6/modules/exportsAndImports5.ts] ////
+
+//// [a.ts]
+export interface A { }
+
+//// [b.ts]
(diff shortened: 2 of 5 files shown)


jakebailey (approved):
I believe this is right, especially given `exportedBindings` also works in terms of original nodes.

Issue #59108, opened 2024-07-02
Regression in types after v5.5


Titozzz:
### 🔎 Search Terms

cannot be used to index type
Generics
forwardRef

### 🕗 Version & Regression Information

- This changed in [commit](https://github.com/microsoft/TypeScript/commit/e0755dc7c12e442db0e6278dbebc29f30a90866f)

### ⏯ Playground Link

https://www.typescriptlang.org/play/?ts=5.5.2#code/CYUwxgNghgTiAEYD2A7AzgF3gMwFzwB4AVAPgAo94yAPfIgSngF4T4BXFAaxSQHcVGLKrXgNmrDtz4oAUAHo58ALZROASxQBzeBgAWatPCjw4UCPDWgUGNRgCeODmBup4ICGqUaoGEIb0IIDAwSDC48oqgkLAIyOhYlMTkImJCRDIyUdBwiKiY8Jr4xG7UvijAhpI8-Mn4AN44SEh08AC+ghJc1bIKJiBKSABuGtoBBSAoQWpgufEwUBpY7p7evv66gcGh4b1ZMbP5hYSkNPWNzaJtHexd0hlx+brMOGSa9BElAA4eYLYQDrwYLYRvAkGwsGMQNQoM4dHZPgg0LppBYUHCEajsEg3B4vCgfH4dBs3Fswh89jkHlgoEUiCUyhUblIaqd4A0sRc6e1xEzuhl7BiAGJqGCYAAKsCgShAvhgSWedKhDMMrI02CC8DF1yqKIA-Jr4PgdfwProoMMtETAhA0AgAEbzFBgJ5sW3wAAGk0GQXdOJW+LWVpJITJvQFCGFoowEvm0tl8qYlyVE0ZqpQ6pgmu1t348H1YsN8C9QX58IQAA1npHxZK40ECOGkNh4LoSABtADkHI7AF0ZEA

### 💻 Code

```ts
declare const f: <T>(f: (x: T) => unknown) => (x: T) => unknown
// making this a real identity function eliminates the error:
// declare const f: <T>(x: T) => T

declare const g: <T extends unknown>(x: { foo: T }) => unknown
// removing the generic constraint eliminates the error:
// declare const g: <T>(x: { foo: T }) => unknown

const h = f(g)
// explicitly writing out the exact type shown in type info eliminates the error:
// declare const a: <T extends unknown>(x: { foo: T }) => unknown

[...]

PR #59121, opened 2024-07-02
Fixed regression in signature instantiation

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/59108
cc @weswigham

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -15958,7 +15958,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     }
 
     function createSignatureTypeMapper(signature: Signature, typeArguments: readonly Type[] | undefined): TypeMapper {
-        return createTypeMapper(signature.typeParameters!, typeArguments);
+        return createTypeMapper(map(signature.typeParameters!, tp => tp.mapper ? instantiateType(tp, tp.mapper) : tp), typeArguments);
     }
 
     function getErasedSignature(signature: Signature): Signature {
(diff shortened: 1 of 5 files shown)


src/compiler/checker.ts
```diff
@@ -15958,7 +15958,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     }
 
     function createSignatureTypeMapper(signature: Signature, typeArguments: readonly Type[] | undefined): TypeMapper {
-        return createTypeMapper(signature.typeParameters!, typeArguments);
+        return createTypeMapper(map(signature.typeParameters!, tp => tp.mapper ? instantiateType(tp, tp.mapper) : tp), typeArguments);
```
Andarist, line 15961: The problem is that the `signature` might already be a product of `instantiateSignature` that was called without `eraseTypeParameters` and which thus cloned type parameters. Those cloned parameters are instantiated there in the arguments list using the available `mapper` but the same mapper wasn't later used when erasing type parameters from that result. This led to the mapper-instantiated cloned type parameter being left non-instantiated after erasing type parameters.

andrewbranch:
@typescript-bot test it

src/compiler/checker.ts
jakebailey: Not 100% certain, but it may be better to do `sameMap` here just because that would avoid extra arrays when nothing actually needs to be mapped (but maybe it doesn't happen that often).

Andarist: sure, I pushed out a commit to use `sameMap` here

weswigham (approved):

andrewbranch:

PR #30089, opened 2019-02-25
Convert to named parameters

gabritto:
Fixes #23552
# Convert to "named parameters" refactor

A function-like declaration with parameters is refactored to receive a single parameter object with properties corresponding to its old parameters. Thus a function's arguments can be named when calling it. This applies to functions, methods, constructors, arrow functions and function expressions.

Example:
```ts
function add(a: number, b: number, c: number): number {
    return a + b + c;
}
```
would be refactored to
```ts
function add({ a, b, c }: { a: number, b: number, c: number }): number {
    return a + b + c;
}
```

## Refactor availability
The refactor will be listed as available on:
* functions
* methods
* constructors for class declarations that have names or for class expressions which are assigned to a const variable without a type annotation
* arrow functions and function expressions that initalize a const variable without type annotation
The refactor will not be available on:
* overloads
* functions with less than 2 parameters
* functions having binding patterns in its parameter declarations
* constructors with parameters that have properties, e.g.
```ts
class Foo {
  constructor(private a: string) {}
}
```

## References

This refactor changes a function's signature, so it affects the way a function should be called either directly and indirectly. The refactor also changes assignability to and from a function's type, since its type will change.
Because of that, there are several situations where performing the refactor could cause a breaking change that would in the best scenario result in type errors or, in the worst scenario, runtime errors and wrong behavior.  
The current solution to this problem is to adopt a conservative approach: the refactor will perform changes only if those changes are guaranteed not to break the code. Otherwise, it will not perform any change.  
The refactor has to look at all the references of a function (through `findAllReferences`) to determine if changing it is safe. If a constructor is being refactored, the refactor also looks at all the references of the parent class declaration or class expression.
However, finding all references is too slow to be performed when the user is checking what are the available refactors. So the refactor will only look at all references if the user tries to apply it.
Therefore, the refactor can fail to perform changes even if it was listed as available and the user tried to applied it. Currently there is no way to inform a user why a refactor has failed.

* **Future Work**: create a mechanism for informing the user why a refactor failed.
Related issue: #28410
 
#### Direct calls

The refactor can change direct calls to functions to conform to the new function signature. This is done by taking the old argument list and replacing it with an argument object whose properties have values corresponding to the old arguments.
Example:  
1. 
```ts
function foo(a: number, b: number) { }
foo(1, 2);
```
becomes
```ts
function foo({ a, b }: { a: number; b: number; }) { }
foo({ a: 1, b: 2 });
```
2. 
```ts
[...]

PR #59140, opened 2024-07-04
allow more cases to be restructured

### LINKED ISSUE #58930: Convert params to destructed object: Cannot apply refactoring
Repro:

1. Go to buffer.ts
2. Put cursor in `static wrap`
3. ctrl+.
4. Select convert params code action

Recording 2024-06-19 at 06 48 54

iisaduan:
fixes #58930

^ was purposefully disabled in https://github.com/microsoft/TypeScript/pull/30089#issue-414340630: 

>This refactor changes a function's signature, so it affects the way a function should be called either directly or indirectly. The refactor also changes assignability to and from a function's type, since its type will change. Because of that, there are several situations where performing the refactor could cause a breaking change...

>The current solution to this problem is to adopt a conservative approach: the refactor will perform changes only if those changes are guaranteed not to break the code...

>Therefore, the refactor can fail to perform changes even if it was listed as available and the user tried to applied it. Currently there is no way to inform a user why a refactor has failed.

TODO: update description more.

### DIFF
--- a/src/services/refactors/convertParamsToDestructuredObject.ts
+++ b/src/services/refactors/convertParamsToDestructuredObject.ts
@@ -44,6 +44,7 @@ import {
     getTouchingToken,
     getTypeNodeIfAccessible,
     Identifier,
+    isCallExpression,
     isCallOrNewExpression,
     isClassDeclaration,
     isConstructorDeclaration,
@@ -239,8 +240,10 @@ function getGroupedReferences(functionDeclaration: ValidFunctionDeclaration, pro
                     continue;
                 }
                 const call = entryToFunctionCall(entry);
-                if (call) {
-                    groupedReferences.functionCalls.push(call);
+                if (call !== undefined) {
+                    if (call !== true) {
+                        groupedReferences.functionCalls.push(call);
+                    }
                     continue;
                 }
             }
@@ -248,8 +251,10 @@ function getGroupedReferences(functionDeclaration: ValidFunctionDeclaration, pro
             const contextualSymbol = getSymbolForContextualType(entry.node, checker);
             if (contextualSymbol && contains(contextualSymbols, contextualSymbol)) {
                 const decl = entryToDeclaration(entry);
-                if (decl) {
-                    groupedReferences.declarations.push(decl);
+                if (decl !== undefined) {
+                    if (decl !== true) {
+                        groupedReferences.declarations.push(decl);
+                    }
                     continue;
                 }
             }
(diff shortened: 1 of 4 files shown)


gabritto (approved):
Left minor comments, but in general I think this looks ok to me. One thing I'm wondering, though, is that if we are willing to allow the refactor to be applied even when we know it will introduce errors, we shouldn't simply allow all cases (i.e. more than the ones in this PR), and rely on the fact that there will be type errors to alert the user to the places where the code needs to be manually updated.

src/services/refactors/convertParamsToDestructuredObject.ts
```diff
@@ -355,47 +366,63 @@ function entryToImportOrExport(entry: FindAllReferences.NodeEntry): Node | undef
     return undefined;
 }
 
-function entryToDeclaration(entry: FindAllReferences.NodeEntry): Node | undefined {
+function entryToDeclaration(entry: FindAllReferences.NodeEntry): Node | true | undefined {
+    // returns a callOrNewExpression if the reference needs to be updated during the refactor
+    // returns 'true' if the reference will not be updated by the refactor, but should not block the refactor
+    // returns 'undefined' if the reference should block the refactor
     if (isDeclaration(entry.node.parent)) {
         return entry.node;
```
gabritto, line 374: I think this also needs to be modified to return `true` in some cases. For instance, the following test case that is similar to the method assignment one, results in the refactor failing to be applied:
```ts
function /*a*/foo/*b*/(actual: number) {
    return actual;
}
const x = foo;
```

That's because `const x = foo` is a declaration, so that `foo` reference will be added to the grouped reference's `declarations`, but it will later be considered invalid [here](https://github.com/microsoft/TypeScript/blob/dacc5d2537a20d95e9c84ab635d974ae0bb05d3f/src/services/refactors/convertParamsToDestructuredObject.ts#L212).

gabritto (changes requested):
Thought of a test case that's not currently covered immediately after approving 😅

src/services/refactors/convertParamsToDestructuredObject.ts
iisaduan: I don't know what I did, but I thought I had tested that case out and it wasn't blocked! I'll add it.

iisaduan:
> One thing I'm wondering, though, is that if we are willing to allow the refactor to be applied even when we know it will introduce errors, we shouldn't simply allow all cases (i.e. more than the ones in this PR), and rely on the fact that there will be type errors to alert the user to the places where the code needs to be manually updated.

I was thinking the same (hence why the draft)-- the original bug report provided a case where there was just one broken reference in a test case, but I can see how it could get bad fast if there are a lot of broken cases. And if we allow this refactor to break this case, why not others?

PR #59141, opened 2024-07-04
Send updated sourceFile to getLiteralText

armanio123:
Fixes https://github.com/microsoft/TypeScript/issues/59085#issuecomment-2198774412

When the node's sourceFile is not the same as the `currentSourceFile` the method returns an incorrect position. In example, in the code below, the position will hit the comment text in `foo.ts` instead of `bar.ts`.
```typescript
// @Filename: foo.ts
// Some comment that is going to be used instead on the correct ['constructor'] text.
import { Bar } from "./bar";

export class Foo extends Bar {
    // | <-- trigger completions here should show ['constructor'] from bar.ts. Instead, crashes.
}

// @Filename: bar.ts
export class ElementNode {
    ['constructor']!: any;
}
```

Note that I unsuccessfully tried creating the repro as a fourslash test but for some reason the completions api treats the whole thing as a single file avoiding the issue. Also, I don't know how to avoid the test that changed.


armanio123:
Closed in favor of https://github.com/microsoft/TypeScript/pull/58216

Issue #56998, opened 2024-01-09
Binary expression in an `if` condition can't narrow identifier reassigned within expression of the optional chain


Andarist:
### 🔎 Search Terms

binary condition refinement predicate optional chain undefined

### 🕗 Version & Regression Information

- This is the behavior in every version I tried

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=5.4.0-dev.20240109#code/C4TwDgpgBAKu0F4oG8BQUoEsAmAuKAdgK4C2ARhAE4DcqAvraqJFAJIEBmVEBAxhOw4B7KEjQZmEAAoBDSjJIRgVfHEi0MmEmAA2mCNgCClTKAD8+YuSq0GqVNgi8dc6ByJ9gmIQSgBzJUFuPgFOIQAxIUo1CAAKSVV4AEp8IMoefkERAB8oD0cOTAIDRndPb18OAgBGeOrEyAAaKGAAJgaIJJR0KB0lLDDRf0DOYMywyOj4OqSNLA4oWKLhMwA6LV19IxNQKABCBCR8iELi7C7xDAHhde09A2NTEDm6KAgdAGdoTAXYpcGkAFgGkMqFhJMYvFWkkkmsNvdtk99oc8gQCkUDBcepowrdNg8ds8enR6PYyrwvD4oFVWnUOs02h0sRg+sBriJASMuOkQlkIdNgNVZj0fotlkI4Xcto9dgcjmiThjzt0ruy8QiZUSMK93l95ot-sIhkCQbyJlFIW0YZL8YjQMyruL1dLCS9SUA

### 💻 Code

```ts
type Type = {
  id: number;
};

type InferenceInfo = {
  typeParameter: Type;
  impliedArity?: number;
};

declare function getInferenceInfoForType(type: Type): InferenceInfo | undefined;

function fn1(t1: Type, t2: Type) {
  let info = getInferenceInfoForType(t1);
  if (info?.impliedArity !== undefined) {
    info.impliedArity; // ok
  } else if ((info = getInferenceInfoForType(t2))?.impliedArity !== undefined) {
    info.impliedArity; // 'info' is possibly 'undefined'.(18048)
  }
}

function fn2(t1: Type, t2: Type) {
  let info = getInferenceInfoForType(t1);
  if (info?.impliedArity !== undefined) {
    info.impliedArity; // ok
  } else if ((info = getInferenceInfoForType(t2))?.impliedArity) {
[...]

PR #59144, opened 2024-07-05
Fixed detection of optional chains containing a reference

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/56998

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -27143,7 +27143,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     function optionalChainContainsReference(source: Node, target: Node) {
         while (isOptionalChain(source)) {
             source = source.expression;
-            if (isMatchingReference(source, target)) {
+            if (isMatchingReference(target, source)) {
                 return true;
             }
         }
--- a/tests/baselines/reference/controlFlowOptionalChain4.symbols
+++ b/tests/baselines/reference/controlFlowOptionalChain4.symbols
@@ -1,0 +1,106 @@
+//// [tests/cases/conformance/controlFlow/controlFlowOptionalChain4.ts] ////
+
+=== controlFlowOptionalChain4.ts ===
+// https://github.com/microsoft/TypeScript/issues/56998
+
+type Type = {
+>Type : Symbol(Type, Decl(controlFlowOptionalChain4.ts, 0, 0))
+
+  id: number;
+>id : Symbol(id, Decl(controlFlowOptionalChain4.ts, 2, 13))
+
+};
+
+type InferenceInfo = {
+>InferenceInfo : Symbol(InferenceInfo, Decl(controlFlowOptionalChain4.ts, 4, 2))
+
+  typeParameter: Type;
+>typeParameter : Symbol(typeParameter, Decl(controlFlowOptionalChain4.ts, 6, 22))
+>Type : Symbol(Type, Decl(controlFlowOptionalChain4.ts, 0, 0))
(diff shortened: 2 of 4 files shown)


src/compiler/checker.ts
```diff
@@ -27143,7 +27143,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     function optionalChainContainsReference(source: Node, target: Node) {
         while (isOptionalChain(source)) {
             source = source.expression;
-            if (isMatchingReference(source, target)) {
+            if (isMatchingReference(target, source)) {
```
Andarist, line 27146: I think an extra refactor should be made here. 

`isMatchingReference` accepts reference as the first parameter but `optionalChainContainsReference` accepts it as the second one (!). That caused the issue here as both of them refer to both of the parameters as source and target - so it wasn't apparent that the reference (called a `target` here) is passed to `isMatchingReference` at the position that is used for candidate expressions (also called a `target` by that function).

I think it would make sense to:
1. rename those parameters to something like `expression` and `reference` or maybe `candidate` and `reference`. `isMatchingReference` sometimes calls recursively itself with adjusted `source` (what I propose to call a `reference` here) but I think that doesn't impact the proposed naming
2. use consistent order of parameters. I think what `optionalChainContainsReference` uses is better (`reference` passed as the second argument to it) but `isMatchingReference` is likely an older function so maybe the team would prefer that.

jakebailey:
@typescript-bot test it

src/compiler/checker.ts
rbuckton: This function is nearly identical to `containsMatchingReference` aside from only supporting optional chaining. I'm not sure argument order is the issue here.

rbuckton: `isMatchingReference` walks both `source` and `target` to different extents. It may just be that some check on one side or the other is not exhaustive enough.

rbuckton (changes requested):
Argument order is likely not the culprit here.

src/compiler/checker.ts
Andarist: > This function is nearly identical to containsMatchingReference aside from only supporting optional chaining. 

Yeah, it is almost identical - and notice that `source` and `target` are passed down to `isMatchingReference` by them at different parameter positions. That only makes me believe more that the argument order is the issue here.

> isMatchingReference walks both source and target to different extents. It may just be that some check on one side or the other is not exhaustive enough.

It *may* be that. But I'm not sure to what extent that function should be walking both so I'm hesitant to make changes there as previous authors haven't embed that logic related to optional chains there. 

---

I re-read the code now and analyzed this again and I'm only more convinced that the argument order is incorrect here. Fixing this in `isMatchingReference` (without changing the adjusted argument order) would require fixing this at its end (in the switch/case that checks `source`):
```ts
            case SyntaxKind.BinaryExpression:
                return (isBinaryExpression(source) && source.operatorToken.kind === SyntaxKind.CommaToken && isMatchingReference(source.right, target));
```

The case here doesn't fit what is handled here at all, it *does* fit what is handled on the target side though:
```ts
            case SyntaxKind.BinaryExpression:
                return (isAssignmentExpression(target) && isMatchingReference(source, target.left)) ||
                    (isBinaryExpression(target) && target.operatorToken.kind === SyntaxKind.CommaToken && isMatchingReference(source, target.right));
```

So given that and the fact that arguments really seem to be semantically flipped when judging by other calls to `isMatchingReference`, I'm inclined to argue the current fix is the correct one.

Issue #59145, opened 2024-07-05
Aliased optional chain calls don't narrow like their inline equivalents


Andarist:
### 🔎 Search Terms

aliased conditions control flow analysis cfa optional chain call

### 🕗 Version & Regression Information

- This is the behavior in every version I tried

### ⏯ Playground Link

https://www.typescriptlang.org/play/?ts=5.6.0-dev.20240704#code/GYVwdgxgLglg9mABAJwKYGc4BsBuqCMAFDACYBci6UyMYA5ogD6LgmrC2okCUiA3gChEiCAiqIY6AEKssXRAF4JJAPwA6WhCwg26QgCIA+gCNZXfdwDcQicETFpZnvxvDRYcVWQUvtBktJrYQBfAVCBUEhYBBQMbDwAJmJySmo-JhYwNg4wLl5BYRg7ZPVNbV0DEycLfNcRMShU7yb0gJIgxFDgoA

### 💻 Code

```ts
function resolve1(id: string | undefined) {
  const isBundled = id?.includes("_bundled");
  if (isBundled) {
    const str: string = id; // error
  }
}

function resolve2(id: string | undefined) {
  if (id?.includes("_bundled")) {
    const str: string = id; // ok
  }
}
```

### 🙁 Actual behavior

`resolve1` errors within the if block

### 🙂 Expected behavior

I'd expect it to be OK. Just like `resolve2` is OK with it here - the condition is exactly the same and TS supports [Control Flow Analysis of Aliased Conditions and Discriminant](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-4.html#control-flow-analysis-of-aliased-conditions-and-discriminants) since 4.4

PR #59153, opened 2024-07-05
Fixed narrowing types by aliased optional chain call expressions

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/59145

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -29297,6 +29297,11 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
 
         function narrowTypeByCallExpression(type: Type, callExpression: CallExpression, assumeTrue: boolean): Type {
             if (hasMatchingArgument(callExpression, reference)) {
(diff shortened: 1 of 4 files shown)


src/compiler/checker.ts
```diff
@@ -29297,6 +29297,11 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
 
         function narrowTypeByCallExpression(type: Type, callExpression: CallExpression, assumeTrue: boolean): Type {
             if (hasMatchingArgument(callExpression, reference)) {
+                // Ordinarily we don't need to to this in control flow analysis because the Binder breaks this down.
+                // However, we may encounter the need to narrow this down here when analyzing aliased conditional expressions.
```
Andarist, line 29301: A need to do some extra narrowings when dealing with aliased conditions isn't new, see: https://github.dev/microsoft/TypeScript/blob/3163fe7e3898c1f48cd9bc097b96e3426cd2a453/src/compiler/checker.ts#L28886-L28888

sandersn:
#59144 is related, at least in the general area it fixes.

PR #59154, opened 2024-07-05
Create a `SourceFile`-level indirection on children maps, store `SyntaxList` children directly on nodes.

### LINKED ISSUE #59101: perf: `Node.getChildren` speed regression in ts 5.5 when used at very large scale
### 🔎 Search Terms

typescript 5.5 performance, slowdown, typescript-eslint

- This changed in commit or PR https://github.com/microsoft/TypeScript/pull/58045 (verified by running our type-aware lints while running `git bisect` on a local typescript repo, it slows down specifically on this commit)

### ⏯ Playground Link

_No response_

### 💻 Code

I've put together a repro in a forked version of the sentry monorepo here: https://github.com/walkerdb/sentry/pull/1. It's using the sentry repo only because it's a large public repo that uses typescript-eslint.

### 🙁 Actual behavior

When running typescript-eslint's type-aware lint rules in a large monorepo with typescript 5.5, we observe lint times between 1.3x to 3x worse than when running the exact same rules and the same config under typescript 5.4. The slowdown seems to be entirely coming from the monomorphized objects change in the ts 5.5 release.

### 🙂 Expected behavior

There should be no performance impact when moving from typescript 5.4 to 5.5.

We've submitted a bug report to typescript-eslint but we're also reporting here since the slowdown can be pinpointed to a specific typescript change.

* typescript-eslint issue: https://github.com/typescript-eslint/typescript-eslint/issues/9474
* the PR that introduces the slowdown: https://github.com/microsoft/TypeScript/pull/58045

Also attaching some before/after update perf traces via 0x that show the extra time is coming largely from typescript internals. The highlighted boxes show the specific part of the process that has ballooned in runtime. I'm happy to share the actual traces with maintainers if they'd be useful in debugging.

| ts 5.4 | ts 5.5 |
| --- | --- |
| Screenshot 2024-07-01 at 5 48 38 PM | Screenshot 2024-07-01 at 6 04 48 PM |

DanielRosenwasser:
Fixes #59101

### DIFF
--- a/src/compiler/factory/nodeChildren.ts
+++ b/src/compiler/factory/nodeChildren.ts
@@ -26,19 +26,17 @@ export function getNodeChildren(node: Node, sourceFile: SourceFileLike): readonl
 /** @internal */
 export function setNodeChildren(node: Node, sourceFile: SourceFileLike, children: readonly Node[]): readonly Node[] {
     if (node.kind === SyntaxKind.SyntaxList) {
-        // Store the children directly on the syntax list node.
-        // This makes orphaned syntax lists easier to clean up,
-        // and avoids the need to track all of them in a WeakMap.
-        (node as SyntaxList)._children = children;
+        // SyntaxList children are always eagerly created in the process of
+        // creating their parent's `children` list. We shouldn't need to set them here.
+        Debug.fail("Should not need to re-set the children of a SyntaxList.");
     }
-    else {
-        let map = sourceFileToNodeChildren.get(sourceFile);
-        if (map === undefined) {
-            map = new WeakMap();
-            sourceFileToNodeChildren.set(sourceFile, map);
-        }
-        map.set(node, children);
+
+    let map = sourceFileToNodeChildren.get(sourceFile);
+    if (map === undefined) {
+        map = new WeakMap();
+        sourceFileToNodeChildren.set(sourceFile, map);
     }
+    map.set(node, children);
     return children;
 }
 
--- a/src/compiler/factory/nodeChildren.ts
+++ b/src/compiler/factory/nodeChildren.ts
@@ -50,3 +50,12 @@ export function unsetNodeChildren(node: Node, origSourceFile: SourceFileLike) {
     }
     sourceFileToNodeChildren.get(origSourceFile)?.delete(node);
 }
+
+/** @internal */
+export function transferSourceFileChildren(sourceFile: SourceFileLike, targetSourceFile: SourceFileLike) {
+    const map = sourceFileToNodeChildren.get(sourceFile);
+    if (map !== undefined) {
+        sourceFileToNodeChildren.delete(sourceFile);
+        sourceFileToNodeChildren.set(targetSourceFile, map);
+    }
+}
--- a/src/compiler/factory/nodeChildren.ts
+++ b/src/compiler/factory/nodeChildren.ts
@@ -1,25 +1,54 @@
 import {
+    Debug,
     emptyArray,
     isNodeKind,
     Node,
+    SourceFileLike,
+    SyntaxKind,
+    SyntaxList,
 } from "../_namespaces/ts.js";
 
-const nodeChildren = new WeakMap<Node, readonly Node[] | undefined>();
+const sourceFileToNodeChildren = new WeakMap<SourceFileLike, WeakMap<Node, readonly Node[] | undefined>>();
 
 /** @internal */
-export function getNodeChildren(node: Node): readonly Node[] | undefined {
-    if (!isNodeKind(node.kind)) return emptyArray;
+export function getNodeChildren(node: Node, sourceFile: SourceFileLike): readonly Node[] | undefined {
+    const kind = node.kind;
+    if (!isNodeKind(kind)) {
+        return emptyArray;
+    }
+    if (kind === SyntaxKind.SyntaxList) {
+        return (node as SyntaxList)._children;
+    }
 
-    return nodeChildren.get(node);
+    return sourceFileToNodeChildren.get(sourceFile)?.get(node);
 }
 
 /** @internal */
-export function setNodeChildren(node: Node, children: readonly Node[]): readonly Node[] {
-    nodeChildren.set(node, children);
+export function setNodeChildren(node: Node, sourceFile: SourceFileLike, children: readonly Node[]): readonly Node[] {
+    if (node.kind === SyntaxKind.SyntaxList) {
+        // Store the children directly on the syntax list node.
+        // This makes orphaned syntax lists easier to clean up,
+        // and avoids the need to track all of them in a WeakMap.
+        (node as SyntaxList)._children = children;
+    }
+    else {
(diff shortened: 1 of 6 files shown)


DanielRosenwasser:
@typescript-bot pack this
@typescript-bot test tsserver top800

DanielRosenwasser:
I am marking this ready for review with cautious optimism. Unfortunately, our LS/TSServer benchmarks don't capture memory usage throughout the operations (nor the behavior as we edit). But I don't really see a noticeable difference in the current benchmarks, and @walkerdb has noted that it removes much of the memory pressure on each individual map.

I also think that moving the children onto `SyntaxList`s is a win, since the entire point of these objects is to hold onto child nodes and it's going to be at most the same amount of space to hold a new property as it is [to create a hash solely for the purposes of storing in a `WeakMap`](https://v8.dev/blog/hash-code).

jakebailey:
The main concern is deopts due to the shape of the syntax list being out of sorts with other nodes, especially if we are not setting that prop consistently in its constructors, which may negatively affect editor performance.

src/compiler/factory/nodeChildren.ts
```diff
@@ -1,25 +1,54 @@
...
+    SyntaxList,
 } from "../_namespaces/ts.js";
 
-const nodeChildren = new WeakMap<Node, readonly Node[] | undefined>();
+const sourceFileToNodeChildren = new WeakMap<SourceFileLike, WeakMap<Node, readonly Node[] | undefined>>();
 
 /** @internal */
-export function getNodeChildren(node: Node): readonly Node[] | undefined {
-    if (!isNodeKind(node.kind)) return emptyArray;
+export function getNodeChildren(node: Node, sourceFile: SourceFileLike): readonly Node[] | undefined {
```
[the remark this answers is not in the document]
jakebailey: Hm, I see now, though it's in one of the `SyntaxKind.SyntaxList` branches where it's questionable whether or not this file should or shouldn't touch it.

DanielRosenwasser:
@typescript-bot test tsserver top800

src/compiler/factory/nodeChildren.ts
```diff
@@ -1,25 +1,54 @@
...
 
 /** @internal */
-export function setNodeChildren(node: Node, children: readonly Node[]): readonly Node[] {
-    nodeChildren.set(node, children);
+export function setNodeChildren(node: Node, sourceFile: SourceFileLike, children: readonly Node[]): readonly Node[] {
+    if (node.kind === SyntaxKind.SyntaxList) {
+        // Store the children directly on the syntax list node.
+        // This makes orphaned syntax lists easier to clean up,
+        // and avoids the need to track all of them in a WeakMap.
+        (node as SyntaxList)._children = children;
```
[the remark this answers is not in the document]
DanielRosenwasser: You're right, it doesn't. I've turned it into an assertion failure. Running with light mode off didn't trigger it, and I'm running the top 800 just to be safe even if it's trivial.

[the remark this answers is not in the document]
jakebailey: Is it actually worth setting explicitly in those other cases rather than just making these helpers use the prop? That seems less error-prone.

DanielRosenwasser:
I just realized that one new behavior here is that child nodes now get thrown away entirely per edit.

DanielRosenwasser:
@typescript-bot test tsserver top800

src/compiler/factory/nodeChildren.ts
```diff
@@ -50,3 +50,12 @@ export function unsetNodeChildren(node: Node, origSourceFile: SourceFileLike) {
     }
     sourceFileToNodeChildren.get(origSourceFile)?.delete(node);
 }
+
+/** @internal */
+export function transferSourceFileChildren(sourceFile: SourceFileLike, targetSourceFile: SourceFileLike) {
+    const map = sourceFileToNodeChildren.get(sourceFile);
+    if (map !== undefined) {
+        sourceFileToNodeChildren.delete(sourceFile);
```
jakebailey, line 58: Is it harmful to leave the old entry behind? Theoretically it will get GC'd once the old one goes away, but I guess maybe it's bad to have two people able to write to this? (Theoretically their keys would be distinct anyway?)

src/compiler/utilities.ts
```diff
@@ -1199,14 +1199,15 @@ export function getTokenPosOfNode(node: Node, sourceFile?: SourceFileLike, inclu
     // trivia for the list, we may have skipped the JSDocComment as well. So we should process its
     // first child to determine the actual position of its first token.
     if (node.kind === SyntaxKind.SyntaxList) {
-        const first = firstOrUndefined(getNodeChildren(node));
+        sourceFile ??= getSourceFileOfNode(node);
```
dmichon-msft, line 1202: Isn't this wasted work when the `getNodeChildren` function special cases `SyntaxKind.SyntaxList` to not use `sourceFile`?

dmichon-msft (commented):
Since this has two changes there is a question of how often the node being queried was a SyntaxList vs. other node types in the performance regression.

src/compiler/factory/nodeChildren.ts
DanielRosenwasser: That's not a bad reason, but I guess the main thing I was thinking about is avoiding a `set` resizing the table if possible. It's probably a small cost, and not bullet-proof, but there's no reason for the two `SourceFile`s to take up space in the map if they're not being used.

Technically you could argue we never need to call `unsetNodeChildren` either since the node will never be reused.

DanielRosenwasser:
> Since this has two changes there is a question of how often the node being queried was a SyntaxList vs. other node types in the performance regression.

This is true, but it was also the simplest way of implementing the feature since the "smart selection" feature creates arbitrary `SyntaxList`s with no parent.

src/compiler/utilities.ts
DanielRosenwasser: Apparently not - this is a recursive function, and at almost every other terminal branch the source file is consulted for something. So I can remove the check, but in practice we'll pull on `getSourceFileOfNode` at some point lower in the tree.

Happy to remove it if you all think it's cleaner to just pass `children` along.

jakebailey (approved):
LGTM but I think it's changed, so:

I assume this is for backporting?

DanielRosenwasser:
Yeah, I do want to back-port after we get some usage in nightly builds, and some feedback from people who filed the issue.

@typescript-bot pack this

src/compiler/factory/nodeChildren.ts
dmichon-msft: Since WeakMap is not iterable and therefore doesn't have a defined order of keys, I'm unclear on whether or not deleting entries from a WeakMap is a trivial operation. It's an interesting question which way performance goes on trimming a WeakMap vs. letting the values get GC'd, but there is something to be said for cleanliness.

jakebailey:

I think we're going to let this sit for a little bit, but making the cherry-pick ahead of time.

Issue #59159, opened 2024-07-06
TypeError: Cannot read properties of undefined (reading 'flags') at needsNameFromDeclaration at getNameForExportedSymbol


Andarist:
### 🔎 Search Terms

import type node class member completion parameter method

### 🕗 Version & Regression Information

- This is a crash

### ⏯ Playground Link

N/A

### 💻 Code

```ts
/// <reference path="fourslash.ts" />

// @module: nodenext

// @Filename: /generation.d.ts
//// export type GenerationConfigType = { max_length?: number };

// @FileName: /index.d.ts
//// export declare class PreTrainedModel {
////   _get_generation_config(
////     param: import("./generation.js").GenerationConfigType,
////   ): import("./generation.js").GenerationConfigType;
//// }
////
//// export declare class BlenderbotSmallPreTrainedModel extends PreTrainedModel {
////   /*1*/
//// }

verify.completions({
  marker: "1",
  includes: [
    {
      name: "_get_generation_config",
      insertText: `_get_generation_config(param: import("./generation.js").GenerationConfigType): import("./generation.js").GenerationConfigType;`,
      filterText: "_get_generation_config",
    },
  ],
  preferences: {
    includeCompletionsWithClassMemberSnippets: true,
    includeCompletionsWithInsertText: true,
  },
  isNewIdentifierLocation: true,
});
```

### 🙁 Actual behavior

crash

### 🙂 Expected behavior

no crash

### Additional information about the issue

repro based on https://github.com/microsoft/TypeScript/issues/59085#issuecomment-2198774440:
```
Req #6451 - completionInfo
    at needsNameFromDeclaration (/typescript-5.6.0-dev.20240630/lib/typescript.js:138216:19)
    at getNameForExportedSymbol (/typescript-5.6.0-dev.20240630/lib/typescript.js:138210:7)
    at visit (/typescript-5.6.0-dev.20240630/lib/typescript.js:161488:20)
    at visitNode (/typescript-5.6.0-dev.20240630/lib/typescript.js:92400:19)
    at tryGetAutoImportableReferenceFromTypeNode (/typescript-5.6.0-dev.20240630/lib/typescript.js:161481:20)
    at /typescript-5.6.0-dev.20240630/lib/typescript.js:161118:37
    at sameMap (/typescript-5.6.0-dev.20240630/lib/typescript.js:2470:22)
[...]

PR #59160, opened 2024-07-06
Fixed crash on authored import type nodes when serializing for declarations

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/59159

### DIFF
--- a/src/services/codefixes/helpers.ts
+++ b/src/services/codefixes/helpers.ts
@@ -899,8 +899,14 @@ export function tryGetAutoImportableReferenceFromTypeNode(importTypeNode: TypeNo
         if (isLiteralImportTypeNode(node) && node.qualifier) {
             // Symbol for the left-most thing after the dot
             const firstIdentifier = getFirstIdentifier(node.qualifier);
-            const name = getNameForExportedSymbol(firstIdentifier.symbol, scriptTarget);
-            const qualifier = name !== firstIdentifier.text
+            if (!firstIdentifier.symbol) {
+                // if symbol is missing then this doesn't come from a synthesized import type node
+                // it has to be an import type node authored by the user and thus it has to be valid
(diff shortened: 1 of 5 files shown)


src/services/codefixes/helpers.ts
```diff
@@ -899,8 +899,14 @@ export function tryGetAutoImportableReferenceFromTypeNode(importTypeNode: TypeNo
         if (isLiteralImportTypeNode(node) && node.qualifier) {
             // Symbol for the left-most thing after the dot
             const firstIdentifier = getFirstIdentifier(node.qualifier);
-            const name = getNameForExportedSymbol(firstIdentifier.symbol, scriptTarget);
-            const qualifier = name !== firstIdentifier.text
+            if (!firstIdentifier.symbol) {
```
Andarist, line 902: Value-space identifiers get their `.symbol` in the binder but type-level identifiers in import type nodes don't.

Alternatively, this could be fixed in `visitExistingNodeTreeSymbolsWorker` but it's more tricky there.

tests/cases/fourslash/completionsClassMemberImportTypeNodeParameter2.ts
```diff
@@ -0,0 +1,31 @@
...
+//// export declare class Derived extends Cls {
+////   /*1*/
+//// }
+
+verify.completions({
+  marker: "1",
+  includes: [
+    {
+      name: "method",
+      insertText: `method(param: import("./doesntexist.js").Foo);`,
```
Andarist, line 21: This is weird but not anything new. Return types nodes with errors are omitted based on `NodeBuilderFlags.SuppressAnyReturnType`. No mechanism like this for parameters is used.

andrewbranch (approved):

andrewbranch:

PR #59164, opened 2024-07-07
fix: allow BindableStaticNameExpression as late-bindable name

### LINKED ISSUE #25083: Enum keys not accepted as computed properties if their name is not a valid identifier
**TypeScript Version:**  2.9.2

**Search Terms:** computed property name

**Code**

```ts
enum Type {
  Foo = 'foo',
  '3x14' = '3x14'
}

type TypeMap = {
  [Type.Foo]: any
  [Type['3x14']]: any
}
```

**Expected behavior:**
`Type['3x14']` to be usable as a computed key in a type definition.

**Actual behavior:**
`A computed property name in a type literal must refer to an expression whose type is a literal type or a 'unique symbol' type.`

This is the case even for `Type.Foo` if it's written as `Type['Foo']` instead; so it seems there is some early bailout when the bracket operator is involved. For certain key names it may not be possible to write them without the brackets.

**Playground Link:** [Link](https://www.typescriptlang.org/play/#src=enum%20Type%20%7B%0A%20%20Foo%20%3D%20'foo'%2C%0A%20%20'3x14'%20%3D%20'3.14'%0A%7D%0A%0Atype%20TypeMap%20%3D%20%7B%0A%20%20%5BType.Foo%5D%3A%20any%0A%20%20%5BType%5B'3x14'%5D%5D%3A%20any%0A%7D%0A)

jeengbe:
Fixes #25083

Most of the stuff was already in place, and I mostly aligned the functions with the types.

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -13216,17 +13216,16 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     /**
      * Indicates whether a declaration name is definitely late-bindable.
      * A declaration name is only late-bindable if:
-     * - It is a `ComputedPropertyName`.
-     * - Its expression is an `Identifier` or either a `PropertyAccessExpression` an
-     * `ElementAccessExpression` consisting only of these same three types of nodes.
-     * - The type of its expression is a string or numeric literal type, or is a `unique symbol` type.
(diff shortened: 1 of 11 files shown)


jeengbe:
I forgot to update the types, and now that I did, there are a ton of errors. Closing until I can figure it out.

PR #59166, opened 2024-07-07
add missing interfaces for sub-classes of PerformanceEntry

### LINKED ISSUE #58644: TS 5.4.5: Return type for performance.getEntriesByType is inaccurate
### ⚙ Compilation target

ESNexr

### ⚙ Library

lib.dom.d.ts

The return type for `getEntriesByType` is a union of many different object types. A list can be seen in the spec, or at https://developer.mozilla.org/en-US/docs/Web/API/PerformanceEntry ("The PerformanceEntry instances will always be one of the following subclasses:: _[List of possibel sub-classes]_")

The return type can - and I think should - be hard-coded for the respective string parameter to `getEntriesByType`, since there is a limited short list of possible string constants, and each corresponds exactly to one possible return type. Maybe an overloaded definition for this function?

Right now the return type is set to `PerformanceEntryList`, which is `PerformanceEntry[]`, and that type is an interface with only the minimal number of entries.

The effect is that when I use, for example, `performance.getEntriesByType('navigation')`, I get a TypeScript error trying to access the properties on the `PerformanceNavigationTiming` array I get from that query.

### Sample Code

[Playground Link](https://www.typescriptlang.org/play/?#code/GYVwdgxgLglg9mABABwIYHMCmB1VBnAJU1QBs5UATTCgCgEoAuRAIzjhOKQG8AoRfxACdMUEIKTJMg4HEEBbVJEx8BqgHRYoAUTBRBMTHgBCATwAqJyTQDkYVADcY6VLATW6K1fwD03xFsFBWSYzAGUAJgBmSIBOJgAFIMlBKBNEVMlECjhDRDA4KERMAA8YPEKEdMtMRHipGXlFCEwdPRNPL19EMwALMsR+gHcgsHQAGhZMCFQQPBqoPrwBpdRa+tkFJQA5BycXeDAzGDkYUcQ4ZgArKagO9Tw4OUwaTF1BNIBeAD4it5M1DI1D7AxDWYRkSjuADcPAAvkA)

```TypeScript
function pageWasRealoaded(): boolean {
    return performance
        .getEntriesByType('navigation')
        // Error: TS2339: Property type does not exist on type PerformanceEntry
        // This is wrong, because this is a PerformanceNavigationTiming object
        .some(entry => entry.type === 'reload');
}
```

### Documentation Link

https://developer.mozilla.org/en-US/docs/Web/API/PerformanceEntry

heysujal:
Fixes #58644

The interface were missing for subclasses of `PerformanceEntry` which causes an error when we try to run below code snippet [Playground Link](https://www.typescriptlang.org/play/?#code/GYVwdgxgLglg9mABABwIYHMCmB1VBnAJU1QBs5UATTCgCgEoAuRAIzjhOKQG8AoRfxACdMUEIKTJMg4HEEBbVJEx8BqgHRYoAUTBRBMTHgBCATwAqJyTQDkYVADcY6VLATW6K1fwD03xFsFBWSYzAGUAJgBmSIBOJgAFIMlBKBNEVMlECjhDRDA4KERMAA8YPEKEdMtMRHipGXlFCEwdPRNPL19EMwALMsR+gHcgsHQAGhZMCFQQPBqoPrwBpdRa+tkFJQA5BycXeDAzGDkYUcQ4ZgArKagO9Tw4OUwaTF1BNIBeAD4it5M1DI1D7AxDWYRkSjuADcPAAvkA)

```ts
function pageWasRealoaded(): boolean {
    return performance
        .getEntriesByType('navigation')
        // Error: TS2339: Property type does not exist on type PerformanceEntry
        // This is wrong, because this is a PerformanceNavigationTiming object
        .some(entry => entry.type === 'reload');
}
```
I have added missing interfaces for different subclasses for every return type which can be found [here](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceEntry/entryType). All of them inherit from `PerformanceEntry` interface.

I found that some interfaces were missing some properties. For example [PerformanceResourceTiming.initiatorType](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming/initiatorType)

I think there are still changes that needs to be done. Thanks @lll000111 for guiding me.
 Please provide feedback.


heysujal:

heysujal:
I have now found that I might need to add interfaces in webworker.generated.d.ts because the issue is not resolved yet.
Earlier, I compiled the code to JavaScript and ran that compiled code. I missed that TypeScript compiler is still throwing the same error.

heysujal:
I just found that some entry types are not supported by `getEntriesByType()`. [See here](https://developer.mozilla.org/en-US/docs/Web/API/Performance/getEntriesByType). I added interfaces for them too, earlier. So, I will have to remove them in the next commit.

heysujal:
@RyanCavanaugh The issue is still not getting solved, can you please help me with this?

PR #59170, opened 2024-07-08
Consider enclosing declaration when serializing inferred return types 

### LINKED ISSUE #58937: Some long function return expressions no longer evaluated in 5.5
### 🔎 Search Terms

function return type inferred isolatedDeclarations regression

This changed occurred as a result of `2b4e7e34e7a40fe0d886aec6f799954420f2ac0c` (https://github.com/microsoft/TypeScript/pull/58546)

### ⏯ Playground Link

https://tsplay.dev/WY7ZbN

### 💻 Code

```ts
/** This is the minimal version which seems somewhat innocuous */

export type inferPipe<t, pipe> =
    pipe extends (In: t) => unknown ? (In: t) => ReturnType<pipe> : never

interface Type<t> {
    pipe<fn extends (In: t) => unknown>(fn: fn): Type<inferPipe<t, fn>>
}

declare const t: Type<string>

// Before: Type<(In: string) => number>
// 2b4e7e34e7a40fe0d886aec6f799954420f2ac0c
// After: Type<(In: string) => ReturnType<(s: string) => number>>

const out = t.pipe(s => parseInt(s))


/** In ArkType and in other cases where these expressions are used as part of a deeper type,
 * it can lead to hovers growing exponentially in size as expressions are chained 
 * instead of being evaluted at each step and staying the same.
 * 
 * Here's an example of what occurs with a longer expression
 *  */
 
export type inferPipe2<t, pipe> =
	pipe extends (In: t) => unknown ?
		(In: t) => ReturnType<pipe> extends infer n extends number ? n
		: ReturnType<pipe> extends infer s extends string ? s
		: ReturnType<pipe> extends infer b extends boolean ? b
		: never
	:	never

interface Type2<t> {
	pipe<fn extends (In: t) => unknown>(fn: fn): Type2<inferPipe2<t, fn>>
}

declare const t2: Type2<string>

// Before: Type2<(In: string) => number>
// 2b4e7e34e7a40fe0d886aec6f799954420f2ac0c
// After: Type2<(In: string) => ReturnType<(s: string) => number> extends infer n extends number ? n : ReturnType<(s: string) => number> extends infer s extends string ? s : ReturnType<(s: string) => number> extends infer b extends boolean ? b : never>

const out2 = t2.pipe(s => parseInt(s))
```

### 🙁 Actual behavior

Trivially evaluated `ReturnType` not evaluated on hover.

### 🙂 Expected behavior

It should be evaluated

The example from the playground can worked around by inferring a variable in the position of `unknown` to get the return type directly instead of using the `ReturnType` alias.

This causes a massive regression in the quality of hover displays for many types in ArkType, which I'm still working to determine if can be avoided as-is.

If the team does decide it wants to address this, it would be a huge relief to me to know so I can avoid trying to find workarounds that may involve changing significant parts of the code base to have hovers that were previously a couple dozen characters long be remotely readable after 5.5.

jakebailey:
In playing around with the repros from #58937 and https://github.com/Effect-TS/effect/issues/3171, I tried to go through the changes in #58546 to narrow things down to what might have caused the problem with inferred/unannotated return types.

One thread to pull on is that removing the `context.mapper` stuff seemed to fix the problem, but cause other undesirable changes.

Another area of note was `serializeReturnTypeForSignatureWorker`; he doc for `tryReuseExistingTypeNodeHelper` says:

```
/**
 * Do you mean to call this directly? You probably should use `tryReuseExistingTypeNode` instead,
 * which performs sanity checking on the type before doing this.
 */
```

`serializeReturnTypeForSignatureWorker` suspiciously calls `tryReuseExistingTypeNodeHelper` directly. I just swapped `tryReuseExistingTypeNode` in with parameters that seemed to make sense, and the repro cases appeared to snap back to what they used to be. For example, the declaration emit in https://github.com/Effect-TS/effect/issues/3171's original post looked like this in 5.4:

```ts
export declare const traverse: {
    <A, R, O, E, B>(f: (a: A) => import("effect/Option").Option<B>): <TR, TO, TE>(self: import("effect/Record").ReadonlyRecord<string, A>) => import("effect/Option").Option<import("effect/Record").ReadonlyRecord<string, B>>;
    <TR_1, TO_1, TE_1, A_1, R_1, O_1, E_1, B_1>(self: import("effect/Record").ReadonlyRecord<string, A_1>, f: (a: A_1) => import("effect/Option").Option<B_1>): import("effect/Option").Option<import("effect/Record").ReadonlyRecord<string, B_1>>;
};
```

But in 5.5 is:

```ts
export declare const traverse: {
    <A, R, O, E, B>(f: (a: A) => import("effect/HKT").Kind<import("effect/Option").OptionTypeLambda, R, O, E, B>): <TR, TO, TE>(self: import("effect/HKT").Kind<import("effect/Record").ReadonlyRecordTypeLambda<string>, TR, TO, TE, A>) => import("effect/HKT").Kind<import("effect/Option").OptionTypeLambda, R, O, E, import("effect/HKT").Kind<import("effect/Record").ReadonlyRecordTypeLambda<string>, TR, TO, TE, B>>;
    <TR, TO, TE, A, R, O, E, B>(self: import("effect/Record").ReadonlyRecord<string, A>, f: (a: A) => import("effect/HKT").Kind<import("effect/Option").OptionTypeLambda, R, O, E, B>): import("effect/HKT").Kind<import("effect/Option").OptionTypeLambda, R, O, E, import("effect/HKT").Kind<import("effect/Record").ReadonlyRecordTypeLambda<string>, TR, TO, TE, B>>;
};
```

Note that we copies `Kind` out of the call signature of `traverse`. With this PR, we go to:

```ts
export declare const traverse: {
    <A, R, O, E, B>(f: (a: A) => import("effect/Option").Option<B>): <TR, TO, TE>(self: import("effect/Record").ReadonlyRecord<string, A>) => import("effect/Option").Option<import("effect/Record").ReadonlyRecord<string, B>>;
    <TR, TO, TE, A, R, O, E, B>(self: import("effect/Record").ReadonlyRecord<string, A>, f: (a: A) => import("effect/Option").Option<B>): import("effect/Option").Option<import("effect/Record").ReadonlyRecord<string, B>>;
};
```

Which is the best of both, in that `Kind` has gone away, and we still kept the better type parameter renaming behavior.

A number of other baselines "regressed" in that they seem to have less node reuse, but some appear to be more reasonable in some ways. Notably, no dts emit in our tests changed, which is good, but indicates that we need more testing here.

This is extremely unscientific. If this PR looks wrong, it should at least provide someone else with some info that could track the issue down better. ~I'm not going to say that this is the fix to #58937 without a bit more consensus.~

Fixes #58937

cc @weswigham @dragomirtitian


jakebailey:
@typescript-bot test it
@typescript-bot pack this

tests/baselines/reference/chainedCallsWithTypeParameterConstrainedToOtherTypeParameter.types
```diff
@@ -79,7 +79,7 @@ class C extends B {
 >A : typeof A
 >  : ^^^^^^^^
 >then : <S extends A>(cb: (x: A) => S) => Chain<S>
```
jakebailey, line 81: In examples like these, I think it's incredibly weird that the old behavior had us not reusing effectively anything _but_ the return type annotation.

tests/baselines/reference/declarationEmitQualifiedAliasTypeArgument.types
```diff
@@ -56,17 +56,17 @@ import { create } from "./bbb";
 
 export const fun = create<T>();
 >fun : () => import("bbb").INode<T>
->    : ^^^^^^       ^^^^^  ^^^^^ ^ 
+>    : ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
jakebailey, line 59: The old reuse here seems dubious to me; `create` is imported, so how are we reusing nodes from another file? Similarly to a lot of what changes in this PR.

tests/cases/compiler/issue58937.ts
jakebailey, line 1: I need to rename this test, I just usually don't start my work with a nice test name given the whole point is that I don't know what's wrong until I start testing...

jakebailey:
@typescript-bot test it
@typescript-bot pack this

src/compiler/checker.ts
```diff
@@ -8322,11 +8322,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             const typePredicate = getTypePredicateOfSignature(signature);
             const type = getReturnTypeOfSignature(signature);
             if (context.enclosingDeclaration && (!isErrorType(type) || (context.flags & NodeBuilderFlags.AllowUnresolvedNames)) && signature.declaration && !nodeIsSynthesized(signature.declaration)) {
-                const annotation = signature.declaration && getNonlocalEffectiveReturnTypeAnnotationNode(signature.declaration);
-                // Default constructor signatures inherited from base classes return the derived class but have the base class declaration
-                // To ensure we don't serialize the wrong type we check that that return type of the signature corresponds to the declaration return type signature
-                if (annotation && getTypeFromTypeNode(context, annotation) === type) {
-                    const result = tryReuseExistingTypeNodeHelper(context, annotation);
+                const annotation = getNonlocalEffectiveReturnTypeAnnotationNode(signature.declaration);
```
dragomirtitian, line 8325: Are we sure removing this check `getTypeFromTypeNode(context, annotation) === type` is a good idea? I think there are cases where the type of the annotation might differ from the type of the signature.

weswigham: That's what `tryReuseExistingTypeNode` does internally and is how it differs from `tryReuseExistingTypeNodeHelper`

jakebailey: Right; removing this didn't change anything except reducing some type instantiations. Sorry, I rebased that particular thing away because I felt pretty confident that I could remove the equality here.

jakebailey:
Interestingly, this is not the fix for the type display issue from https://github.com/DefinitelyTyped/DefinitelyTyped/pull/69838.

gcanti:
I can confirm that installing the npm package referenced here https://github.com/microsoft/TypeScript/pull/59170#issuecomment-2212956524 fixes the type display issue reported in https://github.com/Effect-TS/effect/issues/3171

jakebailey:
Thanks for confirming; I'm going to make one last attempt to make a repro that exhibits the "cannot be named" error before merging/backporting.

jakebailey:
Okay, I give up trying to produce a test case for the cannot be named error.

jakebailey:

jakebailey:
Okay, guess I'll do it manually

PR #55968, opened 2023-10-03
[wip] Shared resolutions in tsserver 

sheetalkamat:
Still working and creating PR to do testing as i push commits
Rework of #55067

PR #59181, opened 2024-07-08
In project, instead of iterating over program files to determine if file needs to be detached, do it through existing mechanism of releasing oldSourceFile

sheetalkamat:
Another refactor found as part of working on #55968


src/server/project.ts
```diff
@@ -1393,6 +1393,24 @@ export abstract class Project implements LanguageServiceHost, ModuleResolutionHo
         this.hasAddedOrRemovedSymlinks = true;
     }
 
+    /** @internal */
+    onReleaseOldSourceFile(
+        oldSourceFile: SourceFile,
+        _oldOptions: CompilerOptions,
```
jakebailey, line 1399: Is this parameter intentionally unused?

jakebailey (approved):

src/server/project.ts
sheetalkamat: yeah its used in services and is API so its unused here

PR #59183, opened 2024-07-08
Fix circularity errors in intra-binding-pattern references

### LINKED ISSUE #49989: Using variable defined by array destructuring assignment in default value of variables defined afterwards.

### LINKED ISSUE #59177: Ghost errors and inconsistencies in destructuring patterns
### 🔎 Search Terms

Destructuring

This changed in #56753

### ⏯ Playground Link

https://www.typescriptlang.org/play/?ssl=7&ssc=1&pln=1&pc=1#code/GYVwdgxgLglg9mABMOcAUBKRBvRAnAUyhDyQENEBfAKGogQGcpEBtMgGkQCNEBeRCgGpEARgC6fVgBYATGIDciRAHpliAKJ48cPIgAmMBmQAOxgmTwNEAdwAWBJCjiJDiegFt3DqAT2I4IFC09GBMrAAeIpwAniKSkRL8LCIyAMwKSqqIAHLOBFo6dIzMLOEyMTLx5YgAXpX80XKSyWkZKmoAqmAE4WbQvoj52rr2hNRAA

### 💻 Code

```ts
function foo() { return a }

const [a, b = a + 1] = [42];  // Error disappears when foo is commented out

const [x1, y1 = x1] = [123];  // No error
const [x2, y2 = x2, z2 = y2] = [123];  // Unexpected error here
```

### 🙁 Actual behavior

Unexpected errors.

### 🙂 Expected behavior

No errors.

_No response_

ahejlsberg:
This PR reverts #56753 and implements a new fix for #49989.

Consider the example:

```ts
const [a, b = a + 1] = [42];
```

The array literal `[42]` is contextually typed by the implied type of the binding pattern. To compute the implied type, we substitute `any` for for the first element (because it has no initializer) and the type of `a + 1` for the second element. However, to compute the type of `a + 1` we need the type of `a`, which in turn means we need the type of `[42][0]`, which means we need to know the type of the array literal `[42]`, which is contextually typed by the implied type of the binding pattern... and we have a circularity.

In #56753, logic was introduced that silently squelches circularity errors, but that isn't feasible because it is highly dependent on resolution order. In this PR we instead introduce logic that detects intra-binding-pattern references (such as the reference to `a` in `a + 1`) and types them as `any` when computing implied binding pattern types.

Though strictly not necessary for the fix, the PR also revises the manner in which we "pad" types of initializer expressions for object destructuring patterns. This used to happen in `checkObjectLiteral` but now happens in `checkDeclarationInitializer`, aligning with the way we pad types of initializer expressions for array destructuring patterns. This change is responsible for the "Initializer provides no value for this binding element and the binding element has no default value" errors changing to messages like "Property 'a' does not exist on type '{}'" in the baselines.

Fixes #49989.
Fixes #59177.


ahejlsberg:
@typescript-bot test it

ahejlsberg:
@typescript-bot test it

ahejlsberg:
Tests are all clean and performance is unaffected. I think this is good to merge.

tests/baselines/reference/circularReferenceInReturnType2.types
```diff
@@ -104,8 +104,8 @@ type Something = { foo: number };
 
 // inference fails here, but ideally should not
 const A = object<Something>()({
->A : ObjectType<Something>
->  : ^^^^^^^^^^^^^^^^^^^^^
+>A : any
```
Andarist, line 107: The previous type was better here but i guess maybe that relied on broken assumptions from the PR that got reverted

babakks:
@ahejlsberg Thanks for this fix, and sorry my PR caused inconsistencies. ❤️

tests/baselines/reference/circularReferenceInReturnType2.types
ahejlsberg: Right, when we report circularity errors we also assign the error type, but the reverted PR sometimes didn't do that.

RyanCavanaugh (approved):
We should follow up on the change in `effect`. It looks correct on first glance but there might be some interplay with the outer contextual type in that example

ahejlsberg:
@RyanCavanaugh The `effect` change is just a switch to a better error message, similar to what you see in the compiler test suites.

Issue #58934, opened 2024-06-19
getPasteEdits returns edits even when no imports are added


mjbvz:
### 🔎 Search Terms

- getPasteEdits
- Paste with imports

### 🕗 Version & Regression Information

5.6.0-dev.20240619

### 💻 Code

`a.ts`

```ts
export interface Foo { }

export const foo: Foo = {}
```

1. Copy `export` or `const`

Paste into `b.ts`

### 🙁 Actual behavior

A paste edit is still returned. This makes VS Code show paste UI even though applying the TS paste does nothing

### 🙂 Expected behavior

No paste edits are returned unless imports are added

PR #59189, opened 2024-07-08
pasteEdits returns no edit when there are no imports needed

navya9singh:
This pr fixes https://github.com/microsoft/TypeScript/issues/58934. It makes getPasteEdits not return any edits if there are no imports added as a fix. Instead, the normal vscode's pasting logic would kick in to only paste text without imports.


src/services/pasteEdits.ts
```diff
@@ -71,6 +71,7 @@ function pasteEdits(
         newText = actualPastedText ? newText.slice(0, pos) + actualPastedText[0] + newText.slice(end) : newText.slice(0, pos) + pastedText[i] + newText.slice(end);
     }
 
+    let returnImportEdits = true;
```
andrewbranch, line 74: Just move the `importAdder` declaration outside the `runWithTemporaryFileUpdate` callback and then you won't need a new variable

andrewbranch (approved):

PR #59192, opened 2024-07-08
Make Signature monomorphic

dragomirtitian:
Part of #58928 - Only make signatures monomorphic.


jakebailey:

PR #59193, opened 2024-07-08
Mark `jsxFactorySymbol` as referenced for noUnusedLocals even in verbatimModuleSyntax

### LINKED ISSUE #59117: Regression issue with React import when using `"jsx": "react"` compiler
### 🔎 Search Terms

"verbatimModuleSyntax", "ts6133, "react", "jsx"

- This changed between versions 5.4.5 and 5.5.3

### ⏯ Playground Link

https://www.typescriptlang.org/play/?noUnusedLocals=true&noUnusedParameters=true&target=7&verbatimModuleSyntax=true&isolatedModules=false&allowSyntheticDefaultImports=false#code/JYWwDg9gTgLgBAJQKYEMDG8BmUIjgcilQ3wG4AocpAD0ljkwFcA7DYCZuAYV0maWYwAjAAoAlHADe5OHCIxGUTgB4AJsABucAPQA+CgF9yQA

### 💻 Code

```ts
import React from 'react';

export function Component1() {
  return <div />;
}

```

### 🙁 Actual behavior

When using `"jsx": "react"` in the compiler option, we are still required to import React in our components.  Now with the version 5.5 of typescript doing such import in conjunction with `verbatimModuleSyntax` returns an unused import error because the code doesn't directly refer to the React import before its transpiled.

### 🙂 Expected behavior

This kind of import should not raise any errors as this is required for the code to work at runtime.

_No response_

andrewbranch:
Fixes #59117 

`markJsxAliasReferenced` itself checks `canCollectSymbolAccessibilityInfo` before marking the symbol as `referenced` for _elision_ purposes (which is not necessary in `verbatimModuleSyntax`), but marks the symbol as `isReferenced` for _unused locals checking_ purposes beforehand. That needs to happen unconditionally, but `markLinkedReferences` bails immediately if `canCollectSymbolAccessibilityInfo` is false.

I re-read all the changes in https://github.com/microsoft/TypeScript/pull/58366 and tried and failed to break other features in combination with `verbatimModuleSyntax`.

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -33371,7 +33371,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
 
         checkJsxPreconditions(node);
 
-        markLinkedReferences(node, ReferenceHint.Jsx);
+        markJsxAliasReferenced(node);
 
         if (isNodeOpeningLikeElement) {
             const jsxOpeningLikeNode = node;
--- a/tests/cases/compiler/verbatimModuleSyntaxReactReference.ts
+++ b/tests/cases/compiler/verbatimModuleSyntaxReactReference.ts
@@ -1,0 +1,14 @@
+// @module: preserve
+// @verbatimModuleSyntax: true
+// @jsx: react
+// @noEmit: true
+// @noUnusedLocals: true
+// @noTypesAndSymbols: true
+
+// @Filename: react.d.ts
+declare module 'react';
+
+// @Filename: index.tsx
+import React from 'react';
+
+export const build = <div>hello </div>;


andrewbranch:

jakebailey (commented):
I think this is okay, though it does break the abstraction if `markLinkedReferences`...

When I was looking at this, what I had found was that the PR that refactored this also modified the behavior to return early from `markLinkedReferences`, but the old code did not for JSX and for others.

Maybe that early return is what needs to be modified, like https://github.com/jakebailey/TypeScript/commit/29869a917124a5b8ea4bad4cd5dee207c85d797f ?

weswigham:
> I think this is okay, though it does break the abstraction of markLinkedReferences...

Yeah, personally, if we think the `isReferenced` setting behavior is needed independently of the `.referenced` setting behavior (because it's used for errors and not emit), we should probably pull it out of the whole `markLinkedReferences` tree of functions and just do it inline.

weswigham:
> (That and all of these functions could be moved inside of markLinkedReferences to not be directly callable)

That had measurably worse performance last I tried, since they all become closures :(

PR #58451, opened 2024-05-06
Fix tslib resolutions

andrewbranch:
Fixes #57688
Also discovered and fixed #58450 in the process

PR #59195, opened 2024-07-09
Added affectsSourceFile to importHelpers and jsxImportSource

armanio123:
Fixes https://github.com/microsoft/TypeScript/issues/59168#issuecomment-2212598315

The issue happens in a project containing nested projects. One of them defines `"importHelpers": true`, while the root config doesn't.

In this scenario, depending on the order the files are opened an incorrect checker is used causing to hit the assert: `Debug Failure. False expression: Expected sourceFile.imports[0] to be the synthesized tslib import`.


sheetalkamat (commented):
This is good but test case would be better

DanielRosenwasser:
Is this a regression from #58654? Does it need to be cherry-picked to `release-5.5`? (CC @gabritto)

andrewbranch:
The inconsistency was always there, but the debug assert is new in #58451, so yes, a cherry-pick would be good

gabritto:
> Is this due to #58654? Does the change need to be cherry-picked to `release-5.5`? (CC @gabritto)

Adding to what Andrew said, #58654 makes us call the function with the failed assertion for potentially every file in a project instead of once per checker.

andrewbranch:
Also FWIW, the reason we're not seeing user reports about this is you need a project references scenario where two projects contain the same source file and differ in compiler options _only_ in `importHelpers`, and it's dependent on project creation order (the one without `importHelpers` has to create the shared source file).

andrewbranch (approved):

sheetalkamat (approved):
Pls fix formatting

andrewbranch:

PR #59208, opened 2024-07-09
Write non-missing undefined on mapped type results into output

### LINKED ISSUE #59056: Declaration emit change in 5.5 with `exactOptionalPropertyTypes`
### 🔎 Search Terms

mapped type, narrow, optional, undefined, union

This changed between versions 5.4 and 5.5. This can be reproduced in the playground.

### ⏯ Playground Link

https://www.typescriptlang.org/play/?ts=5.5.2#code/C4TwDgpgBAkgdhAHgQwMbAPJmASwPZzIA2AzgDwCCAfFALxQDeAUFFANoDSUOcUA1hBB4AZlApRkJKAFc4AEwjCeEOVCTAI8qRU4BdKAH4oXAFxQEANwgAnXQbOyFShKvWa52vS1aGxeqAA+MvKKynLerGY6HLoA3EwAvlAAZIzenNy8AkKi4pLBTmFqiBpafjG+ljZQZjFRevEJ8UygkLC89Myswnh49lAkwNY8AObxrABGyNZmcNIAthM241BTAF4OIc4qjUxMAPT7UACiiJDoKmYMUD19ZoPDcCOBBaEusavTswtL1h-r-Ucbx2UASLXA0Aw0mAdHaSDQmGw+EIpDI8CoTCAA

### 💻 Code

```ts
type InexactOptionals<A> = {
  [K in keyof A as undefined extends A[K] ? K : never]?: undefined extends A[K]
    ? A[K] | undefined
    : A[K];
} & {
  [K in keyof A as undefined extends A[K] ? never : K]: A[K];
};

type In = {
  foo?: string;
  bar: number;
  baz: undefined;
}

// Expected: { foo?: string | undefined; bar: number; baz?: undefined; }
type Out = InexactOptionals<In>
```

### 🙁 Actual behavior

The `A[K] | undefined` in the mapped type ternary is narrowed to `A[K]`, meaning that `foo` in the example is `foo?: string`.

### 🙂 Expected behavior

The `A[K] | undefined` in the mapped type ternary is not narrowed, so we see `foo?: string | undefined` as expected.

My use case is cheaply migrating a codebase to support `exactOptionalPropertyTypes`. It can also be useful for better React DX.

weswigham:
Fixes #59056


weswigham:
This is a bit of a compromise fix - when we see a property that came from a mapped type (or reverse mapped type) instantiation where the input and output are optional, if the type contains an actual `undefined` (and not `missing`) type, we now always write that `| undefined` to the output. It's not strictly required outside of `exactOptionalPropertyTypes` (since the optional prop and the optional prop with an `| undefined` are functionally identical), but keeps the output consistent between the two modes. More relevantly, under `exactOptionalPropertyTypes`, we *only* add the `| undefined` to the preserved node *if there's a real, explicit undefined in the type*. By limiting this to only the results of mapped type instantiations, I think I make it so this has no impact on `isolatedDeclarations` behaviors (since those can't resolve a mapped type).

src/compiler/checker.ts
```diff
@@ -6113,11 +6113,12 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
             const originalType = type;
             if (addUndefined) {
-                type = getOptionalType(type);
+                type = getOptionalType(type, !isParameter(host));
             }
             const clone = tryReuseExistingNonParameterTypeNode(context, typeNode, type, host);
             if (clone) {
-                if (addUndefined && !someType(getTypeFromTypeNode(context, typeNode), t => !!(t.flags & TypeFlags.Undefined))) {
+                // explicitly add `| undefined` if it's missing from the input type nodes and the type contains `undefined` (and not the missing type)
+                if (addUndefined && someType(type, t => !!(t.flags & TypeFlags.Undefined) && t !== missingType) && !someType(getTypeFromTypeNode(context, typeNode), t => !!(t.flags & TypeFlags.Undefined))) {
```
jakebailey, line 6121:
```
containsUndefinedType(type) && (type as UnionType).types[0] !== missingType
```

Here too?

src/compiler/checker.ts
```diff
@@ -6113,11 +6113,12 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
             const originalType = type;
             if (addUndefined) {
-                type = getOptionalType(type);
+                type = getOptionalType(type, !isParameter(host));
             }
             const clone = tryReuseExistingNonParameterTypeNode(context, typeNode, type, host);
             if (clone) {
-                if (addUndefined && !someType(getTypeFromTypeNode(context, typeNode), t => !!(t.flags & TypeFlags.Undefined))) {
+                // explicitly add `| undefined` if it's missing from the input type nodes and the type contains `undefined` (and not the missing type)
+                if (addUndefined && containsUndefinedType(type) && (type as UnionType).types[0] !== missingType && !someType(getTypeFromTypeNode(context, typeNode), t => !!(t.flags & TypeFlags.Undefined))) {
```
jakebailey, line 6121: Hm, rethinking, couldn't `type` itself be undefined in these cases? Is it safe to cast to `UnionType`?

weswigham: `containsUndefinedType` can only return `true` if `type` is a union, so the cast is safe.

weswigham: Wait, no, I'm mistaken, you're correct

src/compiler/checker.ts
```diff
@@ -21425,6 +21428,12 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         return !!((type.flags & TypeFlags.Union ? (type as UnionType).types[0] : type).flags & TypeFlags.Undefined);
     }
 
+
+    function containsNonMissingUndefinedType(type: Type) {
+        const candidate = type.flags & TypeFlags.Union ? (type as UnionType).types[0] : type;
+        return !!(candidate.flags & TypeFlags.Undefined) && candidate !== missingType;
```
jakebailey, line 21434: Does `missingType` always precede `undefinedType`? I actually thought the order was flipped.

weswigham: You can't have both in a union - if non-missing `undefined` is added to a union, `missing` gets removed.

jakebailey (approved):

jakebailey:
I think we want this?

PR #59217, opened 2024-07-10
Disallow truthiness/nullishness checks on syntax that never varies on it

RyanCavanaugh:
Inspired by @mjbvz's suggestion to disallow `if (/regex/)`, this PR implements a series of checks that mostly overlap with ESLint's `no-constant-binary-expression` rule, which [found some great stuff in their v8.14](https://eslint.org/blog/2022/07/interesting-bugs-caught-by-no-constant-binary-expression/).

The rule here is that it's an error to check for the truthiness of an expression which has "static truthy semantics", that is, always evaluates to the same truthiness. Examples:
 * `if ([someArr]) {` is illegal (array literals are always truthy)
 * `const foo = { ...someObj } || { }` is illegal (object literals are always truthy, even those only spread in `null`)

The expressions `false`, `true`, `0`, and `1` are excluded from this check, since code like `while (true) {`, `while(1) {`, `do { } while (false)`, `if (true || i_am_just_debugging) {` are common and unlikely to represent programmer errors.

The same rule is applied to nullishness (the thing `??` checks for). Because `??`'s precedence can be surprising, this finds a huge number of "missing parens" bugs.

This check is entirely syntactic, because with `noUncheckedIndexAccess` off, "hidden" nulls/undefined are idiomatically observed in a variety of places:
```ts
function idx(arr: string[], index: number) {
    return arr[index] ?? "out of bounds";
}
```

A breakdown of all the user test hits can be found at https://github.com/microsoft/TypeScript/pull/59217#issuecomment-2221459919 (part 1) and https://github.com/microsoft/TypeScript/pull/59217#issuecomment-2222063345 (part 2).

There are a *lot*, but 100% of these appear to be legitimate bugs in code, so I don't think it's necessary to add a flag for this.


RyanCavanaugh:
@typescript-bot test it
@typescript-bot test top800

jakebailey:
Wow, not entirely related but the new "build mode continues on error" makes the self test super noisy, it's like some files aren't getting emitted from our repo or something and that really breaks analysis. I assume because we use noEmitOnError (probably we shouldn't?)

jakebailey:
I think the `??` checks are dubious; from the self check, this code is marked as bad:

```
const end = (endMarkName !== undefined ? marks.get(endMarkName) : undefined) ?? timestamp();
```

I don't see a bug in this case; is this just a bug in the rule?

RyanCavanaugh:
Yeah, I forgot ternaries, plus some other bugs. A lot of good finds in there though.

@typescript-bot test top800

src/compiler/types.ts
```diff
@@ -920,6 +920,14 @@ export const enum RelationComparisonResult {
     ReportsMask         = ReportsUnmeasurable | ReportsUnreliable,
 }
 
+/** @internal */
+export const enum PredicateSemantics {
```
jakebailey, line 924: I think `Ternary` would actually cover this, if you didn't want to come up with something new.

tests/baselines/reference/api/typescript.d.ts
```diff
@@ -8705,6 +8705,7 @@ declare namespace ts {
      * ```
      */
     function getJSDocCommentsAndTags(hostNode: Node): readonly (JSDoc | JSDocTag)[];
+    function isBinaryLogicalOperator(token: SyntaxKind): boolean;
```
jakebailey, line 8708: Do we want `/** @internal */` on this? (I know it's a draft but before I forget)

RyanCavanaugh:
Not done looking at all of these yet, but so far a 100% correctness rate...

AbdullahAlfaraj/Auto-Photoshop-StableDiffusion-Plugin

Definitely a bug:
```ts
<div key={`${node_id}_${name}__${index}` ?? void 0}>
```

---

ariakit/ariakit

Definitely a bug:
```ts
horizontal = itemObject.orientation === "horizontal" ?? horizontal;
```

---

bitwarden/clients

Definitely a bug:
```ts
map((options) => options?.trustedDeviceOption != null ?? false),
```

---

ChatGPTNextWeb/ChatGPT-Next-Web

Definitely a bug:
```ts
if (historyMsgLength > modelConfig?.max_tokens ?? 4000) {
```

---

clash-verge-rev/clash-verge-rev

Definitely a bug:
```ts
setTheme({ ...theme_setting } || {});
```

---

dream-num/univer

Definitely bugs (4!):
```ts
// Two instances of this:
undos: [] || undoMutations,

let fontFamily = localeService.t(`fontFamily.${(`${value}` ?? '').replace(/\s/g, '')}`);

        <span className={styles.uiPluginSheetsFontFamilyItem} style={{ fontFamily: value }}>
            {localeService.t(`fontFamily.${(`${value}` ?? '').replace(/\s/g, '')}`)}
        </span>
```

---

ethers-io/ethers.js

Definitely a subtle parenthesization bug:
```ts
                    if (isError(error, "CANCELLED") || isError(error, "BAD_DATA") ||
                        isError(error, "NETWORK_ERROR" || isError(error, "UNSUPPORTED_OPERATION"))) {
```

---

facebook/lexical

Definitely a bug:
```ts
   {...prevStyles} || {},
```

---

gpbl/react-day-picker

Definitely a bug:
```ts
if (monthsDiff < numberOfMonths ?? 1) {
```

---

graphile/crystal

Definitely a bug:
```ts
      const nonNullableAttribute = this.codec.attributes
        ? Object.entries(this.codec.attributes).find(
            ([_attributeName, spec]) =>
              !spec.via && !spec.expression && spec.notNull,
          )?.[0]
        : null ?? pk?.attributes[0];
```

---

labring/FastGPT

Definitely three bugs:
```ts
<QuestionTip ml={1} label={publishT('QPM Tips' || '')}></QuestionTip>
// Second instance of above
<QuestionTip ml={1} label={publishT('QPM Tips' || '')}></QuestionTip>

            <QuestionTip
              ml={1}
              label={t('support.outlink.share.Response Quote tips' || '')}
            ></QuestionTip>
```
---

Licoy/ChatGPT-Midjourney

Definitely a bug:
```ts
if (historyMsgLength > modelConfig?.max_tokens ?? 4000) {
```

---

material-components/material-web

Definitely a bug:
```ts
    methodsTable.addRow([
      `\`${rowObj.name}\``,
      rowObj.parameters.map((p) => `\`${p.name}\``).join(', ') || '_None_',
      `\`${rowObj.returns}\`` ?? '`void`',
      rowObj.description ?? '',
    ]);
```

---

material-shell/material-shell

Definitely a bug:
```ts
    const msWorkspaceIsInFloatLayout =
        metaWindow.msWindow?.msWorkspace.layout.state.key === 'float' ?? false;
```

---

microsoft/azuredatastudio

Definitely two bugs:
```ts
this._logger.warn('Failed to delete token from server.' + e.message ?? e);
// and
const verbose = !!options.verbose ?? (!!process.env['CI'] || !!process.env['BUILD_ARTIFACTSTAGINGDIRECTORY']);
```

---

microsoft/vscode

Many bugs!
```ts
uselessReplaceKey5: '@ham' || '',

// 3 instances of this
assert.ok(!'Unexpected value ' + basename(value.resource.fsPath));

const contentHeight = (editor._getViewModel()?.getLineCount()! * lineHeight) ?? editor.getContentHeight();

singlePerResource: () => !this.getCustomEditorCapabilities(contributedEditor.id)?.supportsMultipleEditorsPerDocument ?? true

tooltip: (`${exitMessage} ` ?? '') + nls.localize('launchFailed.exitCodeOnlyShellIntegration', 'Disabling shell integration in user settings might help.')

return !!this._chatWidget?.rawValue?.hasFocus() ?? false

return this.viewModel?.welcomeExperience === WelcomeExperience.ForWorkspace ?? true;

return !item?.handle.startsWith('vscode-command:') ?? false;

this._logger.warn('Failed to delete token from server.' + e.message ?? e);

const verbose = !!options.verbose ?? (!!process.env['CI'] || !!process.env['BUILD_ARTIFACTSTAGINGDIRECTORY']);
```

---

motion-canvas/motion-canvas

Definitely a bug:
```ts
throw 'Unexpected Status: ' + result.statusCode ?? 'NO_STATUS';
```

---

naver/billboard.js

Definitely a bug, but super unclear what they meant here:
```ts
if (isArcishData && "id") {
```
twofer:
```ts
function getScrollPosition(node: HTMLElement) {
	return {
		x: (window.pageXOffset ?? window.scrollX ?? 0) + node.scrollLeft ?? 0,
		y: (window.pageYOffset ?? window.scrollY ?? 0) + node.scrollTop ?? 0
	};
}
```

---

openblocks-dev/openblocks

```ts
  width: ${(props) =>
    props.placement === "right"
      ? "calc(100% - 96px)"
      : "bottom"
      ? "calc(100% - 112px)"
      : "calc(100% - 136px"};
  flex-grow: 1;
```
and
```ts
const LabelWrapper = styled.div<{ placement: ControlPlacement }>`
  flex-shrink: 0;
  width: ${(props) => (props.placement === "right" ? "96px" : "bottom" ? "112px" : "136px")};
`;
```

RyanCavanaugh:
(wiping sweat from brow) They're literally all bugs

portainer/portainer

Definitely a bug:
```ts
Name: `${node.metadata?.name}${isApi ? 'api' : ''}` ?? ''
```

---

react-page/react-page

Definitely a bug:
```ts
  return useNodeProps(nodeId, (node) =>
    isRow(node)
      ? node.cells?.length > 0 ?? false
      : (node?.rows?.length ?? 0) > 0 ?? false
  );
```

---

reactjs/react.dev

Definitely a bug (what exactly were they trying to do here?):
```ts
    throw new Error(
      'Expected name, title, permalink, and children for ' + name ??
        title ??
        permalink ??
        'unknown'
    );
```

---

rollingscopes/rsschool-app

Definitely a bug:
```ts
 await this.courseRepository.query(
      `select "name", "id", 'event' as "type"  from "event" where "id" = ANY($1)
        union
       select "name", "id", 'task' as "type"  from "task" where "id" = ANY($2)`,
      [fetchEvents.size > 0 ? [...fetchEvents] : null, [...fetchTasks] || null],
    );
```

---

Shopify/react-native-skia

Definitely a bug:
```ts
    ps.textAlign =
      value.textAlign !== undefined
        ? { value: value.textAlign }
        : undefined ?? ps.textAlign;
```

---

supabase/supabase

Definitely 6 bugs here:
```ts
const isNotOrgWithPartnerBilling = !subscription?.billing_via_partner ?? true;
// and
const isAcknowledged = typeof window !== 'undefined' ? localStorage?.getItem(key) === 'true' ?? false : false
// and
function checkFeature(feature: Feature, features?: Feature[]) {
  return !features?.includes(feature) ?? true
}
// and 2 instances of this:
layout: 'vertical' || 'flex',
// and
    <img
       src={`${BASE_PATH}` + menu.icon ?? `/img/icons/menu/${id}.svg`}
       className="w-5 rounded"
    />
```

---

tailwindlabs/headlessui

Definitely a bug:
```ts
let isParentDisabled = parent?.getAttribute('disabled') === '' ?? false
```

---

tremorlabs/tremor

Two bugs with eight manifestations here:
```ts
disabled={!hasScroll?.left ?? true}
disabled={!hasScroll?.right ?? true}
// six instances of this exact line:
const prev = previousValues ? previousValues[dataKey] : undefined;
const percentage = ((value - prev) / prev) * 100 ?? undefined;
```
---

Uniswap/web3-react

This appears to be less of a mechanical error and more of just a wat. This isn't how React actually works; unclear if there's relevant syntactic processing going on here or what.

https://github.com/Uniswap/web3-react/blob/bd267025f06b5808fd6ccc4e18dde4d400c4076a/packages/core/src/hooks.ts#L360
```ts
// to ensure connectors remain fresh, we condition re-renders on loaded, isActive and chainId
void loaded && isActive && chainId
```

---

Vendicated/Vencord

Definitely a bug:
```ts
    wrapSort(comparator: Function, row: any) {
        return row.type === 5
            ? -UserAffinitiesStore.getUserAffinity(row.user.id)?.affinity ?? 0
            : comparator(row);
    },
```

---

vercel/ai-chatbot

Definitely a bug, though I don't know what they really meant here
```ts
const fileName = window.prompt('Enter file name' || '', suggestedFileName)
```

---

withfig/autocomplete

Definitely a bug:
```ts
const name = ("FileSystemArn" ? elm["FileSystemArn"] : elm) as string;
```

---

zu1k/bs-core

Definitely a bug:
```ts
                  <Description name={`${t('book.md5') ?? 'MD5'}: `}>
                    {(
                      <Button
                        colorScheme="gray"
                        variant="ghost"
                        size="xs"
                        leftIcon={<LinkIcon />}
                        onClick={() => {
                          navigator.clipboard.writeText(md5?.toLowerCase() ?? 'Unknown');
                          setMd5Copied(true);
                          setTimeout(() => {
                            setMd5Copied(false);
                          }, 2000);
                        }}
                      >
                        {md5Copied ? t('copied') : md5}
                      </Button>
                    ) ||
                      t('book.unknown') ||
                      'Unknown'}
                  </Description>
```

RyanCavanaugh:
Coallating the errors that people seem to be making here. 100.0% of this code is extremely indicative of a logic error of some sort.

 * Likely brainfarts of `||` or `&&` using `??` with a boolean left operand,
   * Example: `isReady = msg.kind === "ready" || isReady`
   * ariakit/ariakit
   * microsoft/azuredatastudio
 * "Tunnel vision" where the left operand of `??` being `!`ed isn't apparent, or they forgot parens
   * Example: `const p = !someSettingName ?? theDefault`
   * supabase/supabase
   * tremorlabs/tremor
   * Vendicated/Vencord
 * Forgot to fully process the left operand
   * Example: `const name = ("FileSystemArn" ? elm["FileSystemArn"] : elm)`
   * withfig/autocomplete
 * Subtle parens misplacement
   * Example: `if (isError(a, "foo") || isError(c, "bar" || isError(d, "baz")) {`
   * labring/FastGPT
   * ethers-io/ethers.js
   * zu1k/bs-core `||`'d a JSX element
 * Parsing precedence confusion on how `a + b ?? c` parses
   * It's `(a + b) ?? c`, not `a + (b ?? c)`
   * microsoft/azuredatastudio
   * motion-canvas/motion-canvas
   * naver/billboard.js
   * supabase/supabase
 * Parsing precedence confusion, using `??` on a binary op whose right operand is possibly-nullish
   * Difficult to hit without `any` since `number > undefined` is not legal
   * Example: `if (historyMsgLength > modelConfig?.max_tokens ?? 4000)`
   * ChatGPTNextWeb/ChatGPT-Next-Web
   * gpbl/react-day-picker
   * Licoy/ChatGPT-Midjourney
 * Parsing precedense confusion on `? :`
   * Example: `const val = cond ? arr[idx] : undefined ?? "fallback"`
     * This is a coalesce on `undefined ?? "fallback"`, not a coalesce on the ternary
   * These are so subtle it's likely people will think TypeScript is wrong, but it's not
   * graphile/crystal
   * Shopify/react-native-skia
 * Code that is trying to handle impossible conditions, likely due to underconfidence in language semantics?
   * Example: `const msg = \`foo_${bar}\` ?? "empty"
   * Example: `const isReady = options?.ready != null ?? false`
   * Example: `const isNeat = a > b ?? false`
   * People seem to think comparisons involving `NaN` might be nullish? Hard to tell
   * AbdullahAlfaraj/Auto-Photoshop-StableDiffusion-Plugin
   * material-shell/material-shell
   * portainer/portainer
   * react-page/react-page
   * supabase/supabase
   * tailwindlabs/headlessui
   * tremorlabs/tremor
 * Possible left/right typos?
   * Example: `processArray([] || someArray);`
   * dream-num/univer
 * Likely misapprehension that `{ ...null }` is `null`, not `{ }`, same for `\`${value}\``, same for `[...someArray]`
   * Example: `const opts = { ...someOpts } || { }`
   * clash-verge-rev/clash-verge-rev
   * dream-num/univer
   * facebook/lexical
   * material-components/material-web
   * rollingscopes/rsschool-app
 * Critical ternary overflow confused a programmer
   * Example: `const test = a ? b : c ? d : "e" ? f : g;`
   * openblocks-dev/openblocks x2
   * Somehow none in our codebase.
 * No idea how this went wrong or what they intended to do
   * naver/billboard.js: `if (isArcishData && "id") {`
   * reactjs/react.dev: `throw new Error("Expected a, b, c, d : + a ?? b ?? c ?? d ?? "unknown")`
   * supabase/supabase has `"vertical" || "flex"` and didn't seem to understand what I was pointing out
   * vercel/ai-chatbot has `window.prompt("Enter file name" || "")
   * Uniswap/web3-react maybe uses a syntactic postprocessor? `void foo && bar && baz`, ?

src/compiler/diagnosticMessages.json
```diff
@@ -3907,6 +3907,18 @@
         "category": "Error",
         "code": 2868
     },
+    "This expression is always truthy. Did you mean to test something else?": {
```
sandersn, line 3910: nit: I don't like using 'Did you mean' unless we have something specific to suggest.
I'd prefer either just "This expression is always truthy" or "Using XX on this expression appears unintentional because it always evaluates the left-hand side".

Probably the former, because the squiggly should be enough to point out the error to people -- more than a longer message, at least.

RyanCavanaugh:
@typescript-bot pack this

DanielRosenwasser:

RyanCavanaugh:
```ts
// Right operand of ?? is unreachable because the left operand is never nullish
const a = {} ?? null;

declare let id: number | undefined;
// This binary expression is never nullish. Are you missing parentheses?
const e = 0 > id ?? null;

// This expression is always nullish
const c = undefined ?? null;

// This kind of expression is always truthy
const b = [] || false;
```

jakebailey (approved):
LGTM but needs baseline updates and fmt.

src/compiler/checker.ts
```diff
@@ -39571,7 +39577,63 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+                return PredicateSemantics.Always;
+            case SyntaxKind.TypeAssertionExpression:
+            case SyntaxKind.AsExpression:
+            case SyntaxKind.ParenthesizedExpression:
+                return getSyntacticNullishnessSemantics((node as OuterExpression).expression);
+            case SyntaxKind.ConditionalExpression:
+                return getSyntacticNullishnessSemantics((node as ConditionalExpression).whenTrue) | getSyntacticNullishnessSemantics((node as ConditionalExpression).whenFalse);
+            case SyntaxKind.Identifier:
+                if ((node as Identifier).escapedText === "undefined") {
+                    return PredicateSemantics.Always;
```
DanielRosenwasser, line 39632: Put a dollar in the "not a keyword" jar.

```ts
function foo(undefined: string) {
  undefined.toUpperCase();
}
```

sandersn:
I've barely skimmed the code, but I looked at the error messages and tests and both are in a good place.

src/compiler/checker.ts
```diff
@@ -39571,7 +39577,63 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+                }
+                return PredicateSemantics.Sometimes;
         }
+        return PredicateSemantics.Never;
```
sandersn, line 39636: I would advocate making this `Sometimes` juuust in case &mdash; except that all the errors reported by this appear to be good.

sandersn (approved):

tests/baselines/reference/computedPropertyNames48_ES5.types
```diff
@@ -17,6 +17,7 @@ enum E { x }
 
 var a: any;
 >a : any
+>  : ^^^
```
DanielRosenwasser, line 20: What happened that updated the types baselines like this?

src/compiler/checker.ts
DanielRosenwasser: (specifically, I think the right thing to do is check to see if it resolves to the global `undefined`, or doesn't resolve to anything)

tests/baselines/reference/computedPropertyNames48_ES5.types
jakebailey: This is a goofy quirk of the types baselines; in `typeWriter.ts`:

```ts
// Distinguish `errorType`s from `any`s; but only if the file has no errors.
```

So this types file changed because `computedPropertyNames48_ES5.errors.txt` was emitted.

jakebailey: (Honestly, I wonder if it's worth the special case; I noticed it separately while working on something else)

src/compiler/checker.ts
```diff
@@ -39571,7 +39577,63 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+                    case SyntaxKind.BarBarEqualsToken:
+                    case SyntaxKind.AmpersandAmpersandToken:
+                    case SyntaxKind.AmpersandAmpersandEqualsToken:
+                        return PredicateSemantics.Sometimes;
+                }
+                return PredicateSemantics.Never;
+            case SyntaxKind.TypeAssertionExpression:
+            case SyntaxKind.AsExpression:
+            case SyntaxKind.ParenthesizedExpression:
+                return getSyntacticNullishnessSemantics((node as OuterExpression).expression);
```
DanielRosenwasser, line 39625: Missed `satisfies` expressions. Technically you could move this to the top of the function with `skipOuterExpressiosn`

src/compiler/checker.ts
```diff
@@ -44206,9 +44268,60 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+            case SyntaxKind.VoidExpression:
+            case SyntaxKind.NullKeyword:
+                return PredicateSemantics.Never;
+            case SyntaxKind.NoSubstitutionTemplateLiteral:
+            case SyntaxKind.StringLiteral:
+                return !!(node as StringLiteral | NoSubstitutionTemplateLiteral).text ? PredicateSemantics.Always : PredicateSemantics.Never;
+            case SyntaxKind.TypeAssertionExpression:
+            case SyntaxKind.AsExpression:
+            case SyntaxKind.ParenthesizedExpression:
+                return getSyntacticTruthySemantics((node as OuterExpression).expression);
```
DanielRosenwasser, line 44313: Same here about `satisfies` and `skipOuterExpressions`.

DanielRosenwasser (approved):
Minor nits, but I would actually be okay with merging as-is.

DanielRosenwasser:
Actually - if you could make sure all of the outer expression cases at least have tests that'd be ideal.

DanielRosenwasser:
@typescript-bot pack this

spenceryue:
Could we have a tsconfig option to turn this "feature" off?

Here are some debugging/placeholder patterns I've used that no longer work after this change:
- `"hardcoded value" ?? realValue()`
- `"TODO fix this" && { not: "fully working yet" }`
- `"force 'A' until TICKET-123 resolves things" || condition() ? "A" : "B"`
- `{ width: 123 ?? /* Tried but doesn't work for some reason */ "100%" }`

I know I can use comments instead, but then I have to quote the code in the inactive branch to separate it from the comment itself like:
- ``"hardcoded value" /* `realValue()` */``
- `/* TODO fix this */ { not: "fully working yet" }`
- `/* force 'A' until TICKET-123 resolves things */ true || condition() ? "A" : "B"`
- ``{ width: 123 /* Tried but doesn't work for some reason `"100%"` */ }``

There are two problems introduced
- It becomes more difficult to quickly toggle the two branches (e.g. delete `123 ??` to try `"100%"`)
- Prettier often moves the comments around, so that they get detached from their context

ESLint's `no-constant-binary-expression` has the advantage over this feature of being able to configure the severity ("warn" or "error"). I personally prefer rendering "possible logic mistakes" as 🟡 warnings and reserving "possible runtime errors" for 🔴 errors.

I might use this feature if it were restricted to the `if (condition)`. The point is I'd like to be able to configure it and see what works best. If adding configurability increases scope too much, then at least please allow the feature to be toggled on/off like many other existing flags (e.g. [`allowUnreachableCode`](https://www.typescriptlang.org/tsconfig/#allowUnreachableCode), [`allowUnusedLabels`](https://www.typescriptlang.org/tsconfig/#allowUnusedLabels), etc.).

> _Loosely related: I also wish [TS2742 "This is likely not portable."](https://github.com/microsoft/TypeScript/issues/42873) could be disabled or at least silenced with `@ts-expect-error`, because it's triggering false positives in my repo, where I know the types are portable due to `package.json#peerDependences`._

RyanCavanaugh:
`// @ts-ignore` will suppress this. I can't imagine having so many of these patterns that you need to globally re-enable a syntactic trapdoor -- we didn't find any codebases like that.

PR #59232, opened 2024-07-10
Fixed regression in reverse mapped type inference caused by cache leak

Andarist:
This PR fixes regression from https://github.com/microsoft/TypeScript/pull/57837 that was discovered here: https://github.com/reduxjs/redux-toolkit/pull/4500

Unfortunately, it also conflicts with a PR that landed later on: https://github.com/microsoft/TypeScript/pull/58318 . With this fix [`tests/cases/compiler/mappedTypeRecursiveInference.ts`](https://github.com/microsoft/TypeScript/blob/972e9a70c661140ed5df6d780b977acc38237bd4/tests/cases/compiler/mappedTypeRecursiveInference.ts) runs into:
> RangeError: Maximum call stack size exceeded

This is specifically caused by the newly introduced `reportRelationError` here: https://github.com/microsoft/TypeScript/pull/58318/files#diff-d9ab6589e714c71e657f601cf30ff51dfc607fc98419bf72e04f6b0fa92cc4b8R21710

This gets called by `isPropertySymbolTypeRelated` in an already deep stack created by this infinitely expanding mapped type. Then that type (or its slice) gets converted into a string eagerly within this function by `getTypeNamesForErrorDisplay` and this recurses quite a bit too. 

I think that this isn't a new problem. Nothing is fundamentally wrong with the PR that added this `reportRelationError`. It's just that two big stacks call were accidentally "concatenated" in a way that they were not before. 

I'm looking for some guidance as to what could be a potential fix for this. My current best idea is to lazily convert types to strings when actually needed/when handing over to the TS Server as a final diagnostic. This would affect a bunch of places though and I don't want to follow that rabbit hole before getting initial thoughts from the team about this.

cc @gabritto


src/compiler/checker.ts
```diff
@@ -25163,7 +25163,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
      * variable T[P] (i.e. we treat the type T[P] as the type variable we're inferring for).
      */
     function inferTypeForHomomorphicMappedType(source: Type, target: MappedType, constraint: IndexType): Type | undefined {
-        const cacheKey = source.id + "," + target.id + "," + constraint.id;
+        const cacheKey = "H" + source.id + "," + target.id + "," + constraint.id;
```
Andarist, line 25166: The same keys (and results) were shared by `inferTypeForHomomorphicMappedType` and `inferReverseMappedType` and that caused a regression. This means that each of those could get a separate cache as they are not meant to be shared anyway. I didn't bother doing that here but if the team thinks that's a good idea I could do it.

jakebailey:
This looks good but the tests are out of date, it seems!

src/compiler/checker.ts
gabritto: I would recommend using a different cache, since there's no reason for them to be the same.

gabritto:
> This looks good but the tests are out of date, it seems!

The fix is correct but causes test [tests/cases/compiler/mappedTypeRecursiveInference.ts](https://github.com/microsoft/TypeScript/blob/972e9a70c661140ed5df6d780b977acc38237bd4/tests/cases/compiler/mappedTypeRecursiveInference.ts) to stack overflow.

Andarist:
Yep, i mention that in the description of the PR. What do you think about the idea to make type-printing lazy in the diagnostics?

src/compiler/checker.ts
Andarist: done

gabritto:
> I think that my theory about 2 deep stacks being stacked on top of each other here might be correct.
> 
> This `reportRelationError` tries to relate `globalThis` source to what I think is `Deep<typeof globalThis>` target (or some anonymous type derived from it). We get there by descending roughly through this property chain:
> 
> ```
> XMLHttpRequest > responseXML > body > offsetParent > shadowRoot > adoptedStyleSheets > (relateVariances) > CSSStyleSheet > ownerRule > parentRule > parentStyleSheet > ownerNode > (unionOrIntersectionRelatedTo) > Element > ownerDocument > defaultView > frames > self > globalThis
> ```
> 
> I captured this target and printed it _after_ the stack was cleared and it **succeeded**. The result isn't pretty but it did "work":
> 
> typeToString(target)
> This isn't particularly useful display... so ye, definitely something could be done to optimize how this gets printed. That said, maybe making the type print lazy in diagnostics would be something worth exploring _either way_? If it's only a feasible idea it could improve the performance of some of those pathological cases as some of the created error messages end up being discarded.
> 
> Interestingly, this doesn't print the same expansion of this type in the error message:
> 
> ```ts
> // @strict: false
> // @strictNullChecks: false
> type Deep<T> = { [K in keyof T]: Deep<T[K]> };
> 
> const test: {
>   globalThis: Deep<typeof globalThis>;
> } = { globalThis };
> ```
> 
> I don't know where the difference is between those 2.

Adding my explanation of things to this:
1. `typeToString(type)` succeeds for the infinitely deep reverse mapped type if given enough stack because we have a 100k truncation limit for `typeToString`, so eventually we do stop going deeper and deeper into the reverse mapped type because we reach that character limit. Also, we only go so deep during `typeToString` in the tests because the tests set `noErrorTruncation` to `true`. If this option is not set, we start truncating much earlier.

2. I believe the stack overflow didn't happen before 5.5 because we reached the 100k truncation limit before blowing up the stack.

3. My incorrect 5.5 PR #57837 meant that, upon digging deeper into the infinitely deep reverse mapped type, we would call `inferReverseMappedType` and incorrectly return the previous result cached by `inferTypeForHomomorphicMappedType`. This meant that, when digging deeper into the reverse mapped type, we would produce the same mapped type, thus breaking the infinity of the type. So `typeToString` didn't dig infinitely into this type before due to this bug, because it detects that it is printing the same reverse mapped type and stops.

4. With my other 5.5 PR #58318, the stack is already bigger when we call `typeToString` with the infinitely deep type, so there isn't enough stack space to print the type until reaching the 100k truncation limit.

Putting all of those things together means that this PR causes the test to stack overflow.

Andarist:
thanks @gabritto for fixing that printing issue ❤️

gabritto:
@typescript-bot test it

gabritto (approved):
My review counts just for the parts of the PR I didn't write, but the original fix makes sense.

weswigham (approved):

gabritto:

PR #59233, opened 2024-07-11
Experiments with literal type creation/inference

DanielRosenwasser:
Just an experiment.


DanielRosenwasser:

jakebailey:
The benchmark results appear positive because tsc crashed; the benchmarker tries to be resilient to non-zero exit codes since that's what happens on a failed compile, but due to the way we emit `--diagnostics`, what's happening is that in build mode it's managing to build a few projects in the repo, crashes, and then the benchmarker parses whichever project built last as the overall stats.

Likely we need to update the benchmarker to be able to ignore crashes like this and not report false stats.

DanielRosenwasser:

Issue #59234, opened 2024-07-11
Grammar allows infer statement on the left side in a nested conditional


DavidHancu:
### 🔎 Search Terms

infer on the left side, nested conditional

### 🕗 Version & Regression Information

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about common bugs.

### ⏯ Playground Link

https://www.typescriptlang.org/play/?#code/C4TwDgpgBAygPAQQHxQLxQVCAPYEB2AJgM5QAUAlvgGYQBOUAQlrgSVPgK4C2ARvVAD8UYsDpUA5lABcHCADd6ASiFMZUagEMANsQgBuIA

### 💻 Code

```ts
type S<A> = A extends (infer B extends number ? string : never) ? B : false;
```

### 🙁 Actual behavior

It was allowed.

### 🙂 Expected behavior

This seems off since the constraint on A changes based on the inference of B, but this would create a cyclic type.

Andarist:
You might find the information contained [here](https://github.com/microsoft/TypeScript/pull/59235/files#r1673602168) interesting

That-Guy977:
[...]

PR #22764, opened 2018-03-21
Fix missing 'infer T' in declaration files

ahejlsberg:
With this PR we properly emit `infer T` declarations in representations of instantiated conditional types.

~Additionally, we report errors on `infer T` declarations in invalid locations within conditional types nested within other conditional types~. Additionally, we now properly locate the owning conditional type for an `infer T` declaration: It is the innermost containing conditional type in whose `extends` type the `infer T` declaration occurs.

Fixes #22755.

PR #59235, opened 2024-07-11
Raise grammar errors on infer type nodes in nested conditional types

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/59234

### DIFF
--- a/tests/baselines/reference/inferTypes2.js
+++ b/tests/baselines/reference/inferTypes2.js
@@ -52,3 +52,29 @@ export declare function foo2<T>(obj: T): T extends {
 export declare function bar2<T>(obj: T): T extends {
     x: infer P extends number ? infer P : string;
 } ? P : never;
+
+
+//// [DtsFileErrors]
+
+
+inferTypes2.d.ts(10,8): error TS1338: 'infer' declarations are only permitted in the 'extends' clause of a conditional type.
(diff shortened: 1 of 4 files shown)


tests/baselines/reference/inferTypes2.js
```diff
@@ -52,3 +52,29 @@ export declare function foo2<T>(obj: T): T extends {
 export declare function bar2<T>(obj: T): T extends {
     x: infer P extends number ? infer P : string;
 } ? P : never;
+
+
+//// [DtsFileErrors]
```
Andarist, line 57: There was a discussion about this thing when this test case was introduced, see [here](https://github.com/microsoft/TypeScript/pull/22764#pullrequestreview-105846102). Does it meant that this PR is obsolete and the current behavior is OK? It certainly feels like a bug to the reporter and it felt like one to me when I've read the issue.

It turns out though that @ahejlsberg explicitly reverted a similar strict containment rule in that old PR [here](https://github.com/microsoft/TypeScript/pull/22764/commits/a75449cabc70de2ee74d7d9e4d37847d0e357f1f). Would the issue resolution be different in 2024 than it was in 2018? 😅 

cc @weswigham

Andarist:
As pointed out [here](https://github.com/microsoft/TypeScript/issues/59234#issuecomment-2222649867) this would be a breaking change.

PR #59243, opened 2024-07-11
Do not reuse ambient module name resolution from other files while determining if resolution can be reused

sheetalkamat:
When trying to reuse resolutions for a program, we were trying to decide we can just not resolve a module name if it is same name as one of the unmodified source files. 
But that is not correct heuristics as shown in the test added. If we resolve the resolution and we find it and that file augments the ambient module or some other ambient module, we wont do that when using program incrementally (so tsc -watch or tsserver scenarios). Removed that heuristic as the only reliable and same between first time program creation and program with partial structure reuse can tell that certainly. (The added test case shows that incremental program structure does not match the initial program - fails our incremental verifier) 
Found while working on #55968

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -1794,11 +1794,6 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         tryGetMemberInModuleExports: (name, symbol) => tryGetMemberInModuleExports(escapeLeadingUnderscores(name), symbol),
         tryGetMemberInModuleExportsAndProperties: (name, symbol) => tryGetMemberInModuleExportsAndProperties(escapeLeadingUnderscores(name), symbol),
         tryFindAmbientModule: moduleName => tryFindAmbientModule(moduleName, /*withAugmentations*/ true),
-        tryFindAmbientModuleWithoutAugmentations: moduleName => {
-            // we deliberately exclude augmentations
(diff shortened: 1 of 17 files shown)


sheetalkamat:
@typescript-bot test it

sheetalkamat:

src/compiler/checker.ts
```diff
@@ -1794,11 +1794,6 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         tryGetMemberInModuleExports: (name, symbol) => tryGetMemberInModuleExports(escapeLeadingUnderscores(name), symbol),
         tryGetMemberInModuleExportsAndProperties: (name, symbol) => tryGetMemberInModuleExportsAndProperties(escapeLeadingUnderscores(name), symbol),
         tryFindAmbientModule: moduleName => tryFindAmbientModule(moduleName, /*withAugmentations*/ true),
-        tryFindAmbientModuleWithoutAugmentations: moduleName => {
```
jakebailey, line 1797: In another case of someone doing it wrong, https://github.com/angular/tsickle/blob/2129d363a0562b7ca1acdf55945c22b7d16539a3/src/googmodule.ts#L536

I think it's fine remove this; they'll crash and should just use `tryFindAmbientModule`, but moreso they should file an issue that asks for what they need!

jakebailey (approved):

PR #59244, opened 2024-07-11
Introduce scanning fast paths

DanielRosenwasser:
This change is an alternative PR to #59106.


DanielRosenwasser:

PR #59245, opened 2024-07-11
When resolving from @types resolve node instead of core node modules

sheetalkamat:
On top of #59243 
- Resolves `@types/node` for core node module names if resolution failed. 
- This helps with picking up node without having to put it in types when they are specified. 
- Only single change that seems like desirable

### DIFF
--- a/src/testRunner/unittests/helpers.ts
+++ b/src/testRunner/unittests/helpers.ts
@@ -118,23 +118,23 @@ export function createTestCompilerHost(texts: readonly NamedSourceText[], target
     });
     if (useCaseSensitiveFileNames === undefined) useCaseSensitiveFileNames = ts.sys && ts.sys.useCaseSensitiveFileNames;
     const getCanonicalFileName = ts.createGetCanonicalFileName(useCaseSensitiveFileNames);
-    const filesByPath = ts.mapEntries(files, (fileName, file) => [ts.toPath(fileName, "", getCanonicalFileName), file]);
+    const filesByPath = ts.mapEntries(files, (fileName, file) => [ts.toPath(fileName, "/", getCanonicalFileName), file]);
     const trace: string[] = [];
     const result: TestCompilerHost = {
         trace: s => trace.push(s),
         getTrace: () => trace,
         clearTrace: () => trace.length = 0,
-        getSourceFile: fileName => filesByPath.get(ts.toPath(fileName, "", getCanonicalFileName)),
+        getSourceFile: fileName => filesByPath.get(ts.toPath(fileName, "/", getCanonicalFileName)),
         getDefaultLibFileName: () => "lib.d.ts",
         writeFile: ts.notImplemented,
-        getCurrentDirectory: () => "",
+        getCurrentDirectory: () => "/",
         getDirectories: () => [],
         getCanonicalFileName,
         useCaseSensitiveFileNames: () => useCaseSensitiveFileNames,
         getNewLine: () => ts.sys ? ts.sys.newLine : newLine,
-        fileExists: fileName => filesByPath.has(ts.toPath(fileName, "", getCanonicalFileName)),
+        fileExists: fileName => filesByPath.has(ts.toPath(fileName, "/", getCanonicalFileName)),
         readFile: fileName => {
-            const file = filesByPath.get(ts.toPath(fileName, "", getCanonicalFileName));
+            const file = filesByPath.get(ts.toPath(fileName, "/", getCanonicalFileName));
             return file && file.text;
         },
     };
--- a/src/compiler/diagnosticMessages.json
+++ b/src/compiler/diagnosticMessages.json
@@ -5237,6 +5237,10 @@
         "category": "Message",
         "code": 6184
     },
+    "Found core node module '{0}', looking in 'node'": {
+        "category": "Message",
+        "code": 6185
+    },
     "Enable strict checking of function types.": {
         "category": "Message",
         "code": 6186
--- a/src/compiler/moduleNameResolver.ts
+++ b/src/compiler/moduleNameResolver.ts
@@ -520,7 +520,10 @@ function getOriginalAndResolvedFileName(fileName: string, host: ModuleResolution
 
 function getCandidateFromTypeRoot(typeRoot: string, typeReferenceDirectiveName: string, moduleResolutionState: ModuleResolutionState) {
     const nameForLookup = endsWith(typeRoot, "/node_modules/@types") || endsWith(typeRoot, "/node_modules/@types/") ?
-        mangleScopedPackageNameWithTrace(typeReferenceDirectiveName, moduleResolutionState) :
+        mangleScopedPackageNameWithTrace(
+            nonRelativeModuleNameForTypingCacheWithTrace(typeReferenceDirectiveName, moduleResolutionState),
+            moduleResolutionState,
+        ) :
         typeReferenceDirectiveName;
     return combinePaths(typeRoot, nameForLookup);
 }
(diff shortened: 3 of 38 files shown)


sheetalkamat:
@typescript-bot test this

sheetalkamat:
@typescript-bot test it

sheetalkamat:
> @sheetalkamat Here are the results of running the user tests with tsc comparing `main` and `refs/pull/59245/merge`:
> 
> 
> Details
> ## [axios-src](https://github.com/axios/axios.git)
> ### [tsconfig.json](https://github.com/axios/axios/blob/0e4f9fa29077ebee4499facea6be1492b42e8a26/tsconfig.json)
> * [MISSING] `error TS2591: Cannot find name 'require'. Do you need to install type definitions for node? Try `npm i --save-dev @types/node` and then add 'node' to the types field in your tsconfig.`
>   
>   * [test/module/ts-require-default/index.ts#L1](https://github.com/axios/axios/blob/0e4f9fa29077ebee4499facea6be1492b42e8a26/test/module/ts-require-default/index.ts#L1)
>   * [test/module/ts-require-default/index.ts#L2](https://github.com/axios/axios/blob/0e4f9fa29077ebee4499facea6be1492b42e8a26/test/module/ts-require-default/index.ts#L2)
>   * [test/module/ts-require/index.ts#L1](https://github.com/axios/axios/blob/0e4f9fa29077ebee4499facea6be1492b42e8a26/test/module/ts-require/index.ts#L1)
>   * [test/module/ts-require/index.ts#L2](https://github.com/axios/axios/blob/0e4f9fa29077ebee4499facea6be1492b42e8a26/test/module/ts-require/index.ts#L2)
> * [MISSING] `error TS2307: Cannot find module 'assert' or its corresponding type declarations.`
>   
>   * [test/module/ts/index.ts#L1](https://github.com/axios/axios/blob/0e4f9fa29077ebee4499facea6be1492b42e8a26/test/module/ts/index.ts#L1)

From comment : https://github.com/microsoft/TypeScript/pull/59245#issuecomment-2224243898

This seems like good change, Earlier without this change we didnt include `@types/node` because of `compileroptions.types = []` and gave the errors

```
test/module/ts-require-default/index.ts(1,16): error TS2591: Cannot find name 'require'. Do you need to install type definitions for node? Try `npm i --save-dev @types/node` and then add 'node' to the types field in your tsconfig.
test/module/ts-require-default/index.ts(2,15): error TS2591: Cannot find name 'require'. Do you need to install type definitions for node? Try `npm i --save-dev @types/node` and then add 'node' to the types field in your tsconfig.
test/module/ts-require/index.ts(1,16): error TS2591: Cannot find name 'require'. Do you need to install type definitions for node? Try `npm i --save-dev @types/node` and then add 'node' to the types field in your tsconfig.
test/module/ts-require/index.ts(2,15): error TS2591: Cannot find name 'require'. Do you need to install type definitions for node? Try `npm i --save-dev @types/node` and then add 'node' to the types field in your tsconfig.
test/module/ts/index.ts(1,20): error TS2307: Cannot find module 'assert' or its corresponding type declarations.
```

With this change we will resolve the `assert` and `require` names to `@types/node` (They are present just werent picked up before) and no errors any more. Build is error free now.

jakebailey:
The title still marks this as an experiment, is this ready for review?

jakebailey:
> This helps with picking up node without having to put it in types when they are specified.

Is this desirable? Shouldn't I only load node types if I asked for it or imported a lib which referenced it? Or am I misunderstanding?

sheetalkamat:
> The title still marks this as an experiment, is this ready for review?

Yes. Sorry i forgot to remove., Also had to rebase because the base PR merged and conflicts were not resolved automatically.

sheetalkamat:
> > This helps with picking up node without having to put it in types when they are specified.
> 
> Is this desirable? Shouldn't I only load node types if I asked for it or imported a lib which referenced it? Or am I misunderstanding?

It is loading node types because there is import for name `assert` and `require`. If they were not there the program would not include node types. 
Without this PR we already give error - to install node types (which the repro has, and add to compileroptions.types - this repro didnt but it the point is it should not be necessary)
We already do this in few places - while reporting error we report to install node types.
For type aquisition and resolving unresolved imports from type aquisition cache - we do transform these names to "node"
So this seems like desirable behaviour.

jakebailey:
Oh, I see, they're already requiring `assert`. But, I don't know why just that would be desirable? If I set `[]`, imports to `assert` shouldn't work either since that's ambient, right?

That being said, I'm also confused because https://github.com/axios/axios/blob/0e4f9fa29077ebee4499facea6be1492b42e8a26/test/module/ts-require-default/tsconfig.json doesn't contain `"types": []`...

sheetalkamat:
The ambient definition comes from node types. If it was in same file, it would not be resolved.

src/testRunner/unittests/helpers.ts
```diff
@@ -118,23 +118,23 @@ export function createTestCompilerHost(texts: readonly NamedSourceText[], target
     });
     if (useCaseSensitiveFileNames === undefined) useCaseSensitiveFileNames = ts.sys && ts.sys.useCaseSensitiveFileNames;
     const getCanonicalFileName = ts.createGetCanonicalFileName(useCaseSensitiveFileNames);
-    const filesByPath = ts.mapEntries(files, (fileName, file) => [ts.toPath(fileName, "", getCanonicalFileName), file]);
+    const filesByPath = ts.mapEntries(files, (fileName, file) => [ts.toPath(fileName, "/", getCanonicalFileName), file]);
```
sheetalkamat, line 121: This is just making sure the current directory and file paths are operating correctly and we are testing right things. (Kind of unrelated but needed to ensure things are working correctly for this PR)

sheetalkamat:
@jakebailey @andrewbranch can you please review so we can get this before beta

jakebailey:
> The ambient definition comes from node types. If it was in same file, it would not be resolved.

I feel like I still don't understand this; if I have `types = []`, shouldn't that mean that the node types are just never automatically loaded without something specifically including them? I'm not sure what "the same file" means in this context...

andrewbranch:
What’s the motivation for special-casing Node.js like this? Not being able to resolve ambient modules without first loading the types via `"types"` or otherwise getting the definitions into the program is expected behavior, and we already have some special error messages for @types/node (which we could improve/expand, if we wanted to detect that `"types"` needs to be modified). This seems like a problem if someone wanted to install or declare their own stub types for some of the Node.js built-ins. If someone adds

```
"types": ["my-node-like-env"]
```

where `my-node-like-env` declares a module named `"assert"`, it would be really frustrating if an import of `"assert"` always pulls in `@types/node` and there’s no way to turn it off.

jakebailey:
> where `my-node-like-env` declares a module named `"assert"`, it would be really frustrating if an import of `"assert"` always pulls in `@types/node` and there’s no way to turn it off.

That's a very real situation, since non-node users may set `types: []`, install `@types/assert` (an for the `assert` npm package), and then import that.

sheetalkamat:
The main reason was that there is discrepancy in how this works in editor vs command line. With type Aquisition enabled, if locally the module is unresolved, we try to resolve the module using `node` if its name is `assert` or the other core modules and otherwise the module name from global cache. So i was wondering why we wont do same thing locally as well. 

If you had assert types installed they will always be preferred. That part hasnt changed, Its when the module resolution fails, and you have node types installed, with this change, we will resolve to node.

jakebailey:
>  Its when the module resolution fails, and you have node types installed, with this change, we will resolve to node.

To me, that's the part that I think is undesirable; if I set `types=[]`, I under no circumstance want the node types to load unless some other file explicitly referenced them. Otherwise places like DT will start including node's types in extra places, I'll get access to things I didn't expect, etc.

I think for ATA in JS code, that's an okay tradeoff for UX (in JS files!), but I don't think that the heuristic should apply to regular tsc compiles.

armano2:
> > Its when the module resolution fails, and you have node types installed, with this change, we will resolve to node.
> 
> To me, that's the part that I think is undesirable; if I set `types=[]`, I under no circumstance want the node types to load unless some other file explicitly referenced them. Otherwise places like DT will start including node's types in extra places, I'll get access to things I didn't expect, etc.
> 
> I think for ATA in JS code, that's an okay tradeoff for UX (in JS files!), but I don't think that the heuristic should apply to regular tsc compiles.

but in this case you are actually referencing package, eg "assert" or "fs", issue is that those packages do not follow "normal" pattern for `@types`, its simmilar to importing `yarn`, if you install it and there `@types/yarn` you want it to be loaded

we can either do something simmilar to what is done in this PR, we could actually limit it to only `node:* fs` or expose redirects as "@types/fs" -> `@types/node`

jakebailey:
That isn't the way I'm thinking about it; unlike other packages, `@types/node` defines an environment, which just so happens to provide imports.

If I wanted to write the types for a "mostly node compatible" runtime (e.g. what `bun-types` used to do for its subset of the Node builtins), I would not want to see TypeScript forcibly loading `@types/node` just because it's a well-known package.

PR #59246, opened 2024-07-11
Use set instead of array for late visible statements

jakebailey:
Probably not going to be perf critical but I'm looking at places we use arrays and then scan over them repeatedly.


jakebailey:

jakebailey:

PR #59259, opened 2024-07-12
DOM update 2024-07-12

sandersn:
I'm doing this early in preparation for adding `/** @deferred */` comments on callbacks for 5.6.

Changes:
- new interface GetHTMLOptions
- add ImportMeta.resolve
- new interface MediaKeysPolicy
- new interface PointerLockOptions
- delete empty interface RTCRtpCodecCapability, replace with its base type RTCRtpCodec.
- many MDN links are deleted or added
- delete interface InnerHTML, inline its property innerHTML where it was used
- add TextEvent, with associated event properties
- a few new properties on:
   - CSSStyleDeclaration
   - CanvasState
   - ClipboardItem (+ static side)
   - Document
   - MediaKeys
   - MediaSource
   - PopStateEvent
- add Element..getHTML, .insertAdjacentHTML, .requestPointerLock
- Two new events:
   - contextlost
   - contextrestored
- add HTMLTemplateElement.shadowRootClonable, .shadowRootDelegatesFocus, .shadowRootSerializable
- new class MediaSourceHandle
- remove OffscreenCanvasRenderingContext2D.commit
- add ShadowRoot.serializable, .getHTML
- URL.canParse's second parameter adds URL to its type; now `string | URL`
- URL adds parse methods
- new class ViewTransition
- new interface UpdateCallback
- type Transferable adds MediaSourceHandle to union

Test baseline changes:
- Nothing substantive except the addition of URL.parse and URL.canParse's second parameter now allowing URL.


sandersn:
@typescript-bot test this

sandersn:
@typescript-bot test it

src/lib/dom.generated.d.ts
```diff
@@ -12103,13 +11737,8 @@ declare var HTMLObjectElement: {
  * [MDN Reference](https://developer.mozilla.org/docs/Web/API/HTMLOptGroupElement)
  */
 interface HTMLOptGroupElement extends HTMLElement {
-    /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/HTMLOptGroupElement/disabled) */
     disabled: boolean;
-    /**
-     * Sets or retrieves a value that you can use to implement your own label functionality for the object.
-     *
-     * [MDN Reference](https://developer.mozilla.org/docs/Web/API/HTMLOptGroupElement/label)
-     */
```
jakebailey, line 12112: Why did so many props lose their MDN links?

jakebailey:
This is scary:

```
/mnt/vss/_work/1/DefinitelyTyped/types/node/v18/module.d.ts
  290:13  error  TypeScript@local tsconfig.dom.json compile error: 
Overload signatures must all be optional or required  @definitelytyped/expect
```

sandersn:
This PR adds ImportMeta.resolve as a non-optional method:

```ts
interface ImportMeta {
    url: string;
    resolve(specifier: string): string;
}
```

But `resolve` is still experimental-only in node 16 and 18. Node versions really need an associated DOM version instead of all claiming to work with the newest.
(and of course node technically shouldn't depend on the DOM at all.)

sandersn:
After discussing with @andrewbranch we decided to make ImportMeta.resolve required in node 16 and 18 as well. Forcing a null check before usage isn't the right way to handle experimental features in the first place. I'd prefer to delete `resolve`, but that will break existing dependents of node 16/18.

The correct way would have been to have a separate experimental package, probably.

sandersn:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/70054 is merged now, so I'm going to merge this after resolving the baseline conflict.

PR #59264, opened 2024-07-13
Extend Iterator and AsyncIterator with Symbol.dispose.

### LINKED ISSUE #59263: Iterator and AsyncIterator missing Symbol.dispose method in esnext.disposable
### ⚙ Compilation target

ESNext

### ⚙ Library

esnext.disposable

Iterator and AsyncIterator are missing types for Symbol.dispose method.

### Sample Code

```TypeScript
function* myIterable() {
  using stack = new DisposableStack();
  stack.defer(console.log('cleanup'));

  yield 1;
  yield 2;
  yield 3;
}

{
  using iter = myIterable();
  //           ^^^^^^^^^^ error
  iter.next();
}
```

```
The initializer of a 'using' declaration must be either an object with a '[Symbol.dispose]()' method, or be 'null' or 'undefined'.
  Property '[Symbol.dispose]' is missing in type 'Generator<number, void, void>' but required in type 'Disposable'.ts(2850)
lib.esnext.disposable.d.ts(34, 5): '[Symbol.dispose]' is declared here.
```

### Documentation Link

The proposal lists these extensions:

- https://github.com/tc39/proposal-explicit-resource-management?tab=readme-ov-file#built-in-disposables

core-js implementation of the proposal includes the new built-ins:

- https://github.com/zloirock/core-js/pull/1151
- https://github.com/zloirock/core-js/blob/master/packages/core-js/modules/esnext.iterator.dispose.js
- https://github.com/zloirock/core-js/blob/master/packages/core-js/modules/esnext.async-iterator.async-dispose.js

jsonnull:
Fixes #59263


jsonnull:

Josh-Cena:
Iterator and AsyncIterator aren't what they mean in JS yet. https://github.com/microsoft/TypeScript/pull/58222

jsonnull:
> Iterator and AsyncIterator aren't what they mean in JS yet. #58222

I think I know what you mean, I suppose I assumed that using `ESNext` lib opts you into the assumption of a compliant runtime.

Josh-Cena:
No; the problem is more awkward than any other ESNext lib type. Perhaps https://github.com/microsoft/TypeScript/issues/54481 summarizes it better?

In essence, `Iterator` in the lib def actually means the iterator _protocol_, and the built-in `Iterator` prototype is actually `IterableIterator` by convention.

jsonnull:
I'll take a look at revisiting this when some of the iterator changes required have moved forward.

PR #36654, opened 2020-02-06
feat(36266): async function needs Promise<T> return type quick fix

a-tarasyuk:
Fixes #36266

PR #59268, opened 2024-07-14
Expose TypeChecker. getAwaitedType to public

### LINKED ISSUE #59256: Expose TypeChecker.getAwaitedType for public api
### 🔍 Search Terms

getAwaitedType
public api

### ✅ Viability Checklist

### ⭐ Suggestion

`TypeChecker.getAwaitedType` is currently an Internal API. I propose to expose it as a public API.https://github.com/microsoft/TypeScript/blob/46410044add2e9f53cea58e445de18dcda53443f/src/compiler/types.ts#L4971-L4972

### 📃 Motivating Example

In `typescript-eslint`, I'm working on improving the `no-unsafe-return` rule to check for types that returns `Promise<any>`.
- https://github.com/typescript-eslint/typescript-eslint/issues/8674
- https://github.com/typescript-eslint/typescript-eslint/pull/8693

To implement this feature, we need to get the awaited type of the promise. And it looks like this functionality is already implemented in `TypeChecker getAwaitedType`.
If getAwaitedType is provided as public, I think we can avoid duplicating the same logic in multiple tools that need similar functionality.

### 💻 Use Cases

1. What do you want to use this for? Use it in the typescript-eslint.

4. What workarounds are you using in the meantime? There is a way to get the first type argument of a Promise generic. But it doesn't catch some cases. (https://github.com/typescript-eslint/typescript-eslint/pull/8693#discussion_r1672039318)

yeonjuan:
Fixes #59256

### DIFF
--- a/src/compiler/types.ts
+++ b/src/compiler/types.ts
@@ -4968,7 +4968,6 @@ export interface TypeChecker {
     getWidenedLiteralType(type: Type): Type;
     /** @internal */
     getPromisedTypeOfPromise(promise: Type, errorNode?: Node): Type | undefined;
-    /** @internal */
     getAwaitedType(type: Type): Type | undefined;
     /** @internal */
     isEmptyAnonymousObjectType(type: Type): boolean;
--- a/tests/baselines/reference/api/typescript.d.ts
+++ b/tests/baselines/reference/api/typescript.d.ts
@@ -6070,6 +6070,7 @@ declare namespace ts {
         getBaseTypes(type: InterfaceType): BaseType[];
         getBaseTypeOfLiteralType(type: Type): Type;
         getWidenedType(type: Type): Type;
+        getAwaitedType(type: Type): Type | undefined;
         getReturnTypeOfSignature(signature: Signature): Type;
         getNullableType(type: Type, flags: TypeFlags): Type;
         getNonNullableType(type: Type): Type;


yeonjuan:

src/compiler/types.ts
```diff
@@ -4968,7 +4968,6 @@ export interface TypeChecker {
     getWidenedLiteralType(type: Type): Type;
     /** @internal */
     getPromisedTypeOfPromise(promise: Type, errorNode?: Node): Type | undefined;
-    /** @internal */
     getAwaitedType(type: Type): Type | undefined;
```
jakebailey, line 4971: Can you copy the docstring from `getAwaitedType` in the checker here?

yeonjuan: @jakebailey  sure! https://github.com/microsoft/TypeScript/pull/59268/commits/0232440d8d0883bf828c7b8fe7f8f230405bbc5b

jakebailey:
I think my only question left is whether or not `getPromisedTypeOfPromise` is the one that should be exported; it was used in #59316 in favor of `getAwaitedType`. Not 100% certain this case is the same.

rbuckton:
`getPromisedTypeOfPromise` does not account for generics, e.g., the "promised type" of `T` is `T`, even though `T` could be instantiated with something like a `Promise<number>`. `getAwaitedType` accounts for generics, as it will also get the "promised type" but will wrap the result in `Awaited<T>` if the "promised type" is generic.

rbuckton:
That said, if we expose `getAwaitedType` we might want to consider also exposing `unwrapAwaitedType` in the event you actually want the `T`.

jakebailey:
Ah, gotcha. I think people are mainly wanting to simulate the result of the `await` operator, so in that sense this function seems correct (and I have some PRs to ts-eslint that would use it).

src/compiler/types.ts
```diff
@@ -4968,7 +4968,16 @@ export interface TypeChecker {
...
+    /**
+     * Gets the "awaited type" of a type.
+     *
+     * The "awaited type" of an expression is its "promised type" if the expression is a
+     * Promise-like type; otherwise, it is the type of the expression. If the "promised
+     * type" is itself a Promise-like, the "promised type" is recursively unwrapped until a
+     * non-promise type is found.
+     *
+     * This is used to reflect the runtime behavior of the `await` keyword.
+     */
```
rbuckton, line 4980: The original comment contains a bit of "inside baseball" terminology that isn't explained elsewhere in the API documentation, such as what a "promised type" is. I would amend this comment to clarify:

```suggestion
    /**
     * Gets the "awaited type" of a type.
     *
     * If an expression has a Promise-like type, the "awaited type" of the expression is
     * derived from the type of the first argument of the fulfillment callback for that
     * Promise's `then` method. If the "awaited type" is itself a Promise-like, it is
     * recursively unwrapped until a non-promise type is found.
     *
     * If an expression does not have Promise-like type, its "awaited type" is the type
     * of the expression.
     *
     * In both cases, if the resulting  "awaited type" is a generic object type, then it
     * is wrapped in an `Awaited<T>`.
     *
     * This is used to reflect the runtime behavior of the `await` keyword.
     */
```

rbuckton (changes requested):

src/compiler/types.ts
jakebailey: When does `getAwaitedType` return `undefined`?

jakebailey: (I think only in erroring cases but I think it'd would be worth documenting.)

rbuckton: It only returns `undefined` in two cases: A promise that recursively unwraps to itself, or a non-Promise "thenable", which is an object with a callable `then()` method that doesn't accept a callback as its first argument.

rbuckton: I've amended my suggestion above to explain the failure cases.

jakebailey:
Updated the PR; I'm hoping to sneak this in to 5.6, given this function's been exported for a while now so this is really just a doc change... @DanielRosenwasser do you think it's too late for this sort of change?

jakebailey (approved):

jakebailey:
For reference, this function was first added to `TypeChecker` in #36654, so has been available since TS 4.0.

rbuckton (approved):

Issue #59271, opened 2024-07-14
[Regression] Circular reference error when passing a class expression with non-primitive static properties to a generic function


whzx5byb:
### 🔎 Search Terms

class expression static type

### 🕗 Version & Regression Information

- This changed between versions 4.0.5 and 4.1.5

### ⏯ Playground Link

https://www.typescriptlang.org/play/?ts=4.1.5#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXzg2RlQElUAHZDAHgBUA+ACi0uoC546BKTugbgBQgsHgDOGeFHgBeAiCIlyVDE0hQxY+AEF4Ab0EBICVGxgkOHLP0BfQTe5A

### 💻 Code

```ts
declare function returnInput<T>(input: T): T;

const a = returnInput(class A {
	static foo = {}
//              ^?
})
```

### 🙁 Actual behavior

`A.foo: any`
Error: 'foo' implicitly has type 'any' because it does not have a type annotation and is referenced directly or indirectly in its own initializer.(7022)

And also note that this bug only happens when the type of `foo` is **not** a primitive type.
### 🙂 Expected behavior

`A.foo: {}`, as it works in 4.0.5

### Additional information about the issue

To define a React class component, it is a common practice to include its "sub-components" as private static properties. I also use a "observer" function to make the state change auto observed. So the real-world code may look like:
```ts
import { observer } from 'mobx-react';
import { styled } from 'styled-components';

@observer
class ListView extends React.Component {
    // render method here...
    private static readonly Item = observer(class Item extends React.Component {
        private static readonly Header = styled.h4` // <- Cannot infer the type of Header
        // CSS style here...
        `;
    })
}
```

A workaround is to use static block to initialize the static property but it's cumbersome:
```ts
const a = returnInput(class A {
	static foo
//              ^? OK
	static { this.foo = {}; }
})
```

PR #59274, opened 2024-07-14
Avoid circularity when contextual parent type of static property is its class

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/59271

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -31166,7 +31166,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
 
     function getContextualTypeForStaticPropertyDeclaration(declaration: PropertyDeclaration, contextFlags: ContextFlags | undefined): Type | undefined {
         const parentType = isExpression(declaration.parent) && getContextualType(declaration.parent, contextFlags);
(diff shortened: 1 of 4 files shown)


src/compiler/checker.ts
```diff
@@ -31166,7 +31166,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
 
     function getContextualTypeForStaticPropertyDeclaration(declaration: PropertyDeclaration, contextFlags: ContextFlags | undefined): Type | undefined {
         const parentType = isExpression(declaration.parent) && getContextualType(declaration.parent, contextFlags);
-        if (!parentType) return undefined;
+        if (!parentType || getSymbolOfDeclaration(declaration.parent) === parentType.symbol) return undefined;
```
Andarist, line 31169: The idea behind the fix is very similar to the one [here](https://github.com/microsoft/TypeScript/pull/58392). We can't get any new information from a situation like this so we can avoid resolving members of this class at this point.

RyanCavanaugh:
@typescript-bot test this

jakebailey:
@typescript-bot test it

PR #59276, opened 2024-07-14
Simplify statSync

jakebailey:
Not all of our stat calls went through the "faster" error stack path, so I fixed that. But, since we support only Node 14.17, we don't even need it since `throwIfNoEntry: false` is available.


jakebailey:

src/compiler/sys.ts
```diff
@@ -1626,14 +1623,10 @@ export let sys: System = (() => {
...
-         * `throwIfNoEntry` was added so recently that it's not in the node types.
-         * This helper encapsulates the mitigating usage of `any`.
-         * See https://github.com/nodejs/node/pull/33716
-         */
+        /** Calls fs.statSync, returning undefined if any errors are thrown */
         function statSync(path: string): import("fs").Stats | undefined {
-            // throwIfNoEntry will be ignored by older versions of node
-            return (_fs as any).statSync(path, { throwIfNoEntry: false });
+            // throwIfNoEntry is available in Node 14.17 and above, which matches our supported range.
+            return _fs.statSync(path, { throwIfNoEntry: false });
```
jakebailey, line 1629: Potentially I can leave a try/catch here just in case someone uses something before 14.17; it wouldn't be faster but probably best overall since all supported users will stop needing to modify the stack lengh limit.

sheetalkamat (approved):

jakebailey:

src/compiler/sys.ts
```diff
@@ -1626,14 +1625,17 @@ export let sys: System = (() => {
...
-         * `throwIfNoEntry` was added so recently that it's not in the node types.
-         * This helper encapsulates the mitigating usage of `any`.
-         * See https://github.com/nodejs/node/pull/33716
-         */
+        /** Calls fs.statSync, returning undefined if any errors are thrown */
         function statSync(path: string): import("fs").Stats | undefined {
-            // throwIfNoEntry will be ignored by older versions of node
-            return (_fs as any).statSync(path, { throwIfNoEntry: false });
+            // throwIfNoEntry is available in Node 14.17 and above, which matches our supported range.
+            try {
```
sheetalkamat, line 1631: Don’t you then need to set the stack size to make it better like it was there in getModifiedTime

jakebailey: Yes, but the main point is to avoid doing that for Node and runtimes which properly implement `throwIfNoEntry`, but still not crash if the runtime doesn't, at the expense of that particular maybe being slower than it used to be.

jakebailey: Of course, maybe none of this is worth it, since it's "just" a property write.

jakebailey:

jakebailey:
Not really helping, honestly. I don't know that this is worth it besides the cleanup...

sheetalkamat:
> Not really helping, honestly. I don't know that this is worth it besides the cleanup...

If we are keeping the `try catch` around `stat` calls then i wouldn't change anything otherwise change before you added `try catch` block is desirable.

jakebailey:
> If we are keeping the `try catch` around `stat` calls then i wouldn't change anything

You mean, you think this PR isn't worth it, or, that this PR is okay in the current state?

sheetalkamat:
i am saying this PR is not needed if we arent getting rid of try and catch block.
If we need to make any change: change till ba4e6dbd4f605615f4fa39bfe60e718e6c879883 should be done That is not including 508530b

jakebailey:
Hm, maybe I'm not understanding the problem; this at least consolidates the try catching. If we're super confident that no polyfill will get this wrong, I can just revert the try catch. Mainly I was just trying to eliminate the writes to `Error.stackTraceLimit`, since they wouldn't be needed on a well-behaved system anymore.

jakebailey:
I've also read a bit and it seems like try/catch is free as long as any error is never thrown, which it won't be unless someone has a bad platform implementation.

sheetalkamat (approved):

PR #59277, opened 2024-07-15
add more excess property checks 

iisaduan:
todo: description, etc. This is an experimental PR for testing.

Fixes #39998

PR #59279, opened 2024-07-15
[draft followup]

iisaduan:
todo: description, etc. This is an experimental PR for testing.

attempts to do the same thing as #59277 but removes excess property checking on jsx


iisaduan:
@typescript-bot test it

PR #59280, opened 2024-07-15
Fixed crash in `classFields` transform related to broken bodyless constructors

Andarist:
It's just a crash that I noticed in a console when typing out the code

### DIFF
--- a/src/compiler/transformers/classFields.ts
+++ b/src/compiler/transformers/classFields.ts
@@ -2420,7 +2420,7 @@ export function transformClassFields(context: TransformationContext): (x: Source
                 ),
                 multiLine,
             ),
-            /*location*/ constructor?.body
+            /*location*/ constructor?.body,
         );
     }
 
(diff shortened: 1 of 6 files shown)


DanielRosenwasser (approved):

src/compiler/transformers/classFields.ts
```diff
@@ -2416,11 +2416,11 @@ export function transformClassFields(context: TransformationContext): (x: Source
             factory.createBlock(
                 setTextRange(
                     factory.createNodeArray(statements),
-                    /*location*/ constructor ? constructor.body!.statements : node.members,
+                    /*location*/ constructor?.body ? constructor.body.statements : node.members,
```
jakebailey, line 2419:
```suggestion
                    /*location*/ constructor?.body?.statements ?? node.members,
```

This is also valid, right?

Andarist: ye, sure

jakebailey (approved):

Issue #58330, opened 2024-04-26
isolatedDeclarations should not have warning for functions that have no return statements


Jack-Works:
### 🔎 Search Terms

isolatedDeclaration

### 🕗 Version & Regression Information

5.5.0

### ⏯ Playground Link

https://www.typescriptlang.org/play/?isolatedDeclarations=true&ts=5.5.0-dev.20240426#code/KYDwDg9gTgLgBAMwK4DsDGMCWEWIBQCUcA3gL5A

### 💻 Code

```ts
export function f() {}
```

### 🙁 Actual behavior

Function must have an explicit return type annotation with --isolatedDeclarations. (9007)

### 🙂 Expected behavior

No error

### Additional information about the issue

Functions with no return statements can be easily inferred as type `void`, so there is no need to give this warning.

jakebailey:
Not doing this was intentional; a function without a return may not always be `void`.

```ts
declare function fail(): never;
export function f() {
	fail();
}
export const g = () => { fail(); }
```

Is:

```ts
export declare function f(): void;
export declare const g: () => never;
```


[Playground Link](https://www.typescriptlang.org/play/?isolatedDeclarations=true&ts=5.5.0-dev.20240426#code/CYUwxgNghgTiAEAzArgOzAFwJYHtVKiwgAoBKALnlRADcQYBuAWACgQAPABxxgyTUy58iMvADerAJCJCJUsxYBfVh2694YPAGc+Ac3gBeeKIMA+cQSJkG8RUA)

RyanCavanaugh:
Correct error for function expressions, but not function declarations - declarations won't infer `never`

jakebailey:
#58331 is indiscriminate in what it affects; if we want to allow this for function declarations only, that may be possible, but it sure seems like an extra inconsistency just to avoid 5 characters...

Jack-Works:
hmm, I didn't think of `never`, so the function expression part is wrong. But I find it annoying to write `: void` for obviously clear function declarations.

RyanCavanaugh:
Discussed and we don't want to make it untenable to, in the future, change the rules around function declaration return type inference. For example, it's maybe preferable that
```ts
function dofail() {
  Debug.fail("oops");
}
```
have an inferred return type of `never` (as it would if written as a function expression) rather than `void`.

PR #59282, opened 2024-07-15
Extract node type printer

dragomirtitian:
Extract the syntactic type reuse both from type annotations and from assertions in order to enable a type printer Independent of the type checker.


tests/baselines/reference/ambientConstLiterals.js
```diff
@@ -74,5 +74,5 @@ declare const c9: {
 declare const c10: number[];
 declare const c11: string;
 declare const c12: number;
-declare const c13: string;
-declare const c14: number;
+declare const c13: "abc" | "def";
+declare const c14: 123 | 456;
```
jakebailey, line 78: This seems to reintroduce the widening problem we've previously discussed.

dragomirtitian: I did notice this, but I think this is actually a bug. Hovering over those variables in the [playground](https://www.typescriptlang.org/play/?#code/MYewdgzgLgBMCMBmGBeGBZAhlAFgOgCdMwATEAWwAoBKGAPhgAY8BWGAfhgCJMAjYLjABc3EgFMAZlwDcAWABQoSLAQAWVBmz4ipCjXpNWHGPABMyEapYA2aUA) shows the union on hover while the declarations show `string` and `number`.

DanielRosenwasser:
@typescript-bot test tsserver top300

tests/baselines/reference/jsFileCompilationDuplicateVariableErrorReported.js
```diff
@@ -14,4 +14,4 @@ var x = 10; // Error reported
 
 //// [out.d.ts]
 declare var x: string;
-declare var x: string;
```
weswigham, line 17: ...? Feels like we need to talk about the behavior of multiple declarations of the same thing :(

I don't think the result determined by the syntactic printer for the second declaration (`number`) jives with the inferred type of the declaration (`string`). When a value has multiple declarations, I think it's only safe to apply the node reuse logic on the first one.

tests/baselines/reference/objectSpreadSetonlyAccessor.types
```diff
@@ -26,14 +26,14 @@ const o2: { foo: undefined } = { foo: 1, ... { set foo(_v: number) { } } }
 >   : ^^^^^^^         ^^^
 >foo : undefined
 >    : ^^^^^^^^^
->{ foo: 1, ... { set foo(_v: number) { } } } : { foo: undefined; }
```
weswigham, line 29: ...this just seems wrong? The `set` in the spread 100% overrides the `1`, so pulling the type from the `1` leads to an incorrect declaration. Another case where there being multiple declarations for the symbol maybe implies it's not safe to just reuse the type from the initializer!

tests/baselines/reference/optionalMethods.js
```diff
@@ -142,7 +142,7 @@ declare class Bar {
     e: number;
     a: number;
     b?: number;
-    c?: number | undefined;
```
weswigham, line 145: Unsure if this counts as a break or not. the declaration is `c? = 2`. Seems like the `?` is redundant if it can't explicitly be assigned `undefined`, so this is supposed to get an explicit `| undefined` added. I think.

tests/baselines/reference/promiseTypeStrictNull.types
```diff
@@ -1,7 +1,7 @@
 //// [tests/cases/compiler/promiseTypeStrictNull.ts] ////
 
 === Performance Stats ===
-Instantiation count: 2,500 -> 5,000
```
weswigham, line 4: Another big increase.

tests/baselines/reference/strictFunctionTypes1.js
```diff
@@ -82,7 +82,7 @@ declare function fs(x: string): void;
 declare function fx(f: (x: "def") => void): void;
 declare const x1: (x: string) => void;
 declare const x2 = "abc";
-declare const x3: string;
+declare const x3: "def" | "abc";
```
weswigham, line 85: Seems like another case of differing widening in the declaration emit?

tests/baselines/reference/stringLiteralTypesOverloads02.js
```diff
@@ -104,9 +104,9 @@ declare namespace Consts1 {
...
 declare const number = "number";
 declare const boolean = "boolean";
-declare const stringOrNumber: string;
-declare const stringOrBoolean: string;
-declare const booleanOrNumber: string;
-declare const stringOrBooleanOrNumber: string;
+declare const stringOrNumber: "string" | "number";
+declare const stringOrBoolean: "string" | "boolean";
+declare const booleanOrNumber: "number" | "boolean";
+declare const stringOrBooleanOrNumber: "string" | "number" | "boolean";
```
weswigham, line 110: More `string`s that have become literal unions.

tests/baselines/reference/typeFromJSInitializer.types
```diff
@@ -215,8 +215,8 @@ var n;
 
 // should get any on parameter initialisers
 function f(a = null, b = n, l = []) {
->f : (a?: null, b?: number | undefined, l?: any[]) => void
->  : ^ ^^^^^^^^^ ^^^^^^^^^^^^^^^^^^^^^^^ ^^^^^^^^^^^^^^^^^
+>f : (a?: null, b?: number | undefined, l?: never[]) => void
```
weswigham, line 218: `any[]` -> `never[]`? Seems like a big different in the inferred type between the syntactic builder and the inferred type. An empty `[]` is more often than not inferred as an `any[]` rather than a `never[]`. Usually for back compat.

weswigham (changes requested):
The crashes in the extended suites are obviously concerning, too. Probably shouldn't stack overflow when printing language service stuff. 😆

tests/baselines/reference/strictFunctionTypes1.js
dragomirtitian: This seems like it is wrong in the current emit. Hovering over the type will show `"def" | "abc"`. This seems because the type was widened in `createTypeOfDeclaration` regardless of if it was actually needed or not.

```ts

    function createTypeOfDeclaration(declarationIn: AccessorDeclaration | VariableLikeDeclaration | PropertyAccessExpression, enclosingDeclaration: Node, flags: NodeBuilderFlags, internalFlags: InternalNodeBuilderFlags, tracker: SymbolTracker) {
        const declaration = getParseTreeNode(declarationIn, isVariableLikeOrAccessor);
        if (!declaration) {
            return factory.createToken(SyntaxKind.AnyKeyword) as KeywordTypeNode;
        }
        // Get type of the symbol if this is the valid symbol otherwise get type at location
        const symbol = getSymbolOfDeclaration(declaration);
        const type = symbol && !(symbol.flags & (SymbolFlags.TypeLiteral | SymbolFlags.Signature))
            ? getWidenedLiteralType(getTypeOfSymbol(symbol))
            : errorType;

        return nodeBuilder.serializeTypeForDeclaration(declaration, type, symbol, enclosingDeclaration, flags | NodeBuilderFlags.MultilineObjectLiterals, internalFlags, tracker);
    }
    
    ```

File: tests/baselines/reference/jsFileCompilationDuplicateVariableErrorReported.js (hunk 2 above)
Comment (dragomirtitian) [reply tests/baselines/reference/jsFileCompilationDuplicateVariableErrorReported.js:17]: Isn't this always an error case (this case is an error case). If so is it really a problem that the type gets printed is different?

File: tests/baselines/reference/typeFromJSInitializer.types (hunk 8 above)
Comment (dragomirtitian) [reply tests/baselines/reference/typeFromJSInitializer.types]: Reverted to old behavior.

File: tests/baselines/reference/optionalMethods.js (hunk 4 above)
Comment (dragomirtitian) [reply tests/baselines/reference/optionalMethods.js]: Reverted to original behavior.

File: tests/baselines/reference/objectSpreadSetonlyAccessor.types (hunk 3 above)
Comment (dragomirtitian) [reply tests/baselines/reference/objectSpreadSetonlyAccessor.types]: Reverted to old behavior.

File: tests/baselines/reference/jsDeclarationEmitDoesNotRenameImport.js [hunk 9]
```diff
@@ -46,7 +46,7 @@ declare class Test {
 //// [index.d.ts]
 export default X;
 export type Options = {
-    test?: typeof import("./Test.js").default | undefined;
```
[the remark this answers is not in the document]
dragomirtitian: Reverted to old behavior.

tests/baselines/reference/declarationsWithRecursiveInternalTypesProduceUniqueTypeParams.types
```diff
@@ -1,7 +1,7 @@
 //// [tests/cases/compiler/declarationsWithRecursiveInternalTypesProduceUniqueTypeParams.ts] ////
 
 === Performance Stats ===
-Instantiation count: 500 -> 2,500
+Instantiation count: 2,500 -> 10,000
```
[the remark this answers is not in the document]
dragomirtitian: Reduced instantiations in this case.

tests/baselines/reference/declarationEmitExactOptionalPropertyTypesNodeNotReused(exactoptionalpropertytypes=false).js
```diff
@@ -32,7 +32,7 @@ exports.baddts = foo();
 
 //// [declarationEmitExactOptionalPropertyTypesNodeNotReused.d.ts]
 export declare const baddts: (x: {
-    foo?: string | undefined;
+    foo?: string;
```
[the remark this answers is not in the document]
dragomirtitian: Reverted to old behavior.

tests/baselines/reference/promiseTypeStrictNull.types
dragomirtitian: Now decreased compared to main.

tests/baselines/reference/nodeModulesImportTypeModeDeclarationEmitErrors1(module=node16).js
```diff
@@ -121,16 +121,16 @@ exports.b = null;
 
 //// [index.d.ts]
 export type LocalInterface = import("pkg", { assert: { "resolution-mode": "foobar" } }).RequireInterface & import("pkg", { assert: { "resolution-mode": "import" } }).ImportInterface;
-export declare const a: import("pkg").RequireInterface;
-export declare const b: import("pkg", { with: { "resolution-mode": "import" } }).ImportInterface;
+export declare const a: import("pkg", { assert: { "resolution-mode": "foobar" } }).RequireInterface;
+export declare const b: import("pkg", { assert: { "resolution-mode": "import" } }).ImportInterface;
```
jakebailey, line 125: I think we're trying to avoid emitting `assert`; is this desirable?

tests/baselines/reference/isolatedDeclarationErrors.errors.txt
```diff
@@ -15,6 +16,10 @@ isolatedDeclarationErrors.ts(8,1): error TS9023: Assigning properties to functio
 !!! error TS9023: Assigning properties to functions without declaring them is not supported with --isolatedDeclarations. Add an explicit declaration for the properties assigned to this function.
     
     const errorOnMissingReturn = () => {}
+                                 ~~~~~~~~
+!!! error TS9007: Function must have an explicit return type annotation with --isolatedDeclarations.
```
jakebailey, line 20: Was this not an error before for some reason? I can't recall. Also failing on the self check for the same reason.

dragomirtitian: I fixed the self check. This was a bug previously, we were not catching a case that should have been an isolated declarations error (the function had a return with an assertion, but the return type could also be `undefined`). I will have a look at this one tomorrow, but it probably should have been an error? I need to think through our reasoning for not inferring return types on function expressions that have no return again.

jakebailey: The overarching reason was that having no return did not imply a `void` return, since CFA internally could instead say it's `never` due to asserts or similar.

For empty funcs, see https://github.com/microsoft/TypeScript/issues/58330, and I think consistency between arrow funcs and func declarations is desirable.

But, this makes me wonder if we should be extracting this fix out separately since it seems like a bug fix that would be good to have without this PR.

dragomirtitian: I am not sure we should ever consider functions without a return type as ok for isolated declarations. Right now we use the type from an assertion in a return. But this is only valid if the return is at the top level. I can make this change in a separate PR,  but are we sure at that point it's still worth it?

tests/baselines/reference/nodeModulesImportTypeModeDeclarationEmitErrors1(module=node16).js
dragomirtitian: I could change the reuse logic to never reuse types that have `assert` in them and use the printer instead.

tests/baselines/reference/isolatedDeclarationErrorsClasses.errors.txt
```diff
@@ -59,6 +60,9 @@ isolatedDeclarationErrorsClasses.ts(56,5): error TS9013: Expression type can't b
 !!! error TS7032: Property 'setOnly' implicitly has type 'any', because its set accessor lacks a parameter type annotation.
                     ~~~~~
 !!! error TS7006: Parameter 'value' implicitly has an 'any' type.
+                    ~~~~~
+!!! error TS9009: At least one accessor must have an explicit return type annotation with --isolatedDeclarations.
```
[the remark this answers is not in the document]
dragomirtitian: Changed the error message to not include return to say `At least one accessor must have an explicit type annotation with --isolatedDeclarations.`

jakebailey:
@typescript-bot test it

jakebailey:
This turns out to change DT printing greatly, it seems.

dragomirtitian:
@jakebailey The test failures in definitely type are due to the fact that this PR enabled syntactic printing even when no enclosing declaration was present. I reverted this, so hopefully there should not be any failures. Can you run the tests again please?

jakebailey:
@typescript-bot test it

dragomirtitian:
@jakebailey Thanks for the CI run :). Performance looks good. DT looks good. There are some small differences in some error messages, but they don't seem significant (just minor changes in type printing) 

@weswigham Can you please have a look at the PR again, and my comments above?

tests/baselines/reference/declFileObjectLiteralWithOnlySetter.types
```diff
@@ -3,13 +3,13 @@
 === declFileObjectLiteralWithOnlySetter.ts ===
 function /*1*/makePoint(x: number) { 
 >makePoint : (x: number) => { b: number; x: number; }
->          : ^ ^^      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+>          : ^ ^^      ^^^^^^^^^^^^^^^^^^^^^      ^^^
```
jakebailey, line 6: Why is this getting marked as reused? Presumably the return value is inside the function, so wouldn't it not be accessible?

tests/baselines/reference/isolatedDeclarationErrorsClasses.errors.txt
```diff
@@ -52,13 +53,16 @@ isolatedDeclarationErrorsClasses.ts(56,5): error TS9013: Expression type can't b
...
+!!! error TS9009: At least one accessor must have an explicit type annotation with --isolatedDeclarations.
 !!! related TS9032 isolatedDeclarationErrorsClasses.ts:11:9: Add a return type to the get accessor declaration.
         set setOnly(value) { }
             ~~~~~~~
 !!! error TS7032: Property 'setOnly' implicitly has type 'any', because its set accessor lacks a parameter type annotation.
                     ~~~~~
 !!! error TS7006: Parameter 'value' implicitly has an 'any' type.
+                    ~~~~~
+!!! error TS9009: At least one accessor must have an explicit type annotation with --isolatedDeclarations.
+!!! related TS9033 isolatedDeclarationErrorsClasses.ts:12:9: Add a type to parameter of the set accessor declaration.
```
jakebailey, line 65: Is there a way to silence this by chance? Usually we try to avoid duplicate errors but maybe it's fine...

tests/baselines/reference/declFileObjectLiteralWithOnlySetter.types
dragomirtitian: The way the syntactic node printer is now hooked up, it will get a chance to look for a type for any symbol whose types is serialized. So when printing the return type, the syntactic printer will be asked for a type for `x`, it then sees the accessor `set x` and the finds the `number` annotation, and since it's the same as the inferred type, it will be reused.

tests/baselines/reference/nodeModulesImportTypeModeDeclarationEmitErrors1(module=node16).js
dragomirtitian: Changed to not reuse import types with assert keyword.

tests/baselines/reference/isolatedDeclarationErrorsClasses.errors.txt
dragomirtitian: This is also an issue for parameters ([Playground Link](https://www.typescriptlang.org/play/?noImplicitAny=false&isolatedDeclarations=true#code/MYGwhgzhAEAa0G8BQ1ougWwBQAcCUAXNAG4D2AlgCaJLoC+SdQA)). I think we can suppress these errors if the user also has `noImplicitAny` turned on. But I think we should do that in a separate PR so we don't overload this PR with other changes.

tests/baselines/reference/jsFileCompilationDuplicateVariableErrorReported.js
weswigham: In the context of `//@ts-ignore` existing at all, yes - it might not need to be ID-compatible, since if you suppress an ID error all bets are off with respect to the output being ID-similar, but we still want accurate-to-the-best-of-our-analysis declarations for incremental compilation.

tests/baselines/reference/ambientConstLiterals.js
weswigham: It's a case where the declaration file is going to be inaccurate whichever it picks. In reality, only something like `declare const c13 = var ? "abc" : "def"` would actually get the correct widening behavior, but we don't support that in declaration files, so we compromise and fall on the side of emitting the widened type. Is it a bug? Yes. Is this the fix? No. Is the fix in the scope of this PR? Definitely not.

weswigham (changes requested):

src/compiler/checker.ts
```diff
@@ -6656,13 +6745,23 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             function createAnonymousTypeNode(type: ObjectType): TypeNode {
                 const typeId = type.id;
                 const symbol = type.symbol;
+                // if (checkTruncationLength(context)) {
```
weswigham, line 6748: Commented out code?

tests/baselines/reference/strictFunctionTypes1.js
dragomirtitian: Removed the change.

tests/baselines/reference/ambientConstLiterals.js
dragomirtitian: Removed the change.

src/compiler/checker.ts
dragomirtitian: Wops. Removed

tests/baselines/reference/jsFileCompilationDuplicateVariableErrorReported.js
dragomirtitian: Reverted to old emit.

jakebailey:
Hm, self check is now failing...

dragomirtitian:
> Hm, self check is now failing...

Fixed.

tests/baselines/reference/parseImportAttributesError.js
```diff
@@ -47,5 +47,5 @@ ImportInterface;
 
 //// [index.d.ts]
 export type LocalInterface = import("pkg", { with: {} });
-export declare const a: any;
-export declare const b: any;
+export declare const a: import("pkg", { with: {} });
+export declare const b: import("pkg", { with: {} });
```
jakebailey, line 51: Are there any issues with preserving these attributes? I assume not, and that it's probably good to keep them.

tests/baselines/reference/nodeModulesImportAttributesTypeModeDeclarationEmit(module=node16).js
```diff
@@ -32,5 +32,5 @@ exports.b = null;
 
 //// [index.d.ts]
 export type LocalInterface = import("pkg", { with: { "resolution-mode": "require" } }).RequireInterface & import("pkg", { with: { "resolution-mode": "import" } }).ImportInterface;
-export declare const a: import("pkg").RequireInterface;
+export declare const a: import("pkg", { with: { "resolution-mode": "require" } }).RequireInterface;
```
jakebailey, line 35: This is a behavior change, but, I think an okay one since we're trying to make it so that isolatedDeclaration doesn't have to reason about this at all.

tests/baselines/reference/transpile/declarationRestParameters.d.ts
```diff
@@ -31,17 +48,30 @@ export declare const v2: (n: "n", a: "a") => {
...
+!!! related TS9027 v2.ts:2:14: Add a type annotation to the variable v2.
         /** module var */
         n: typeof n,
+        ~
+!!! error TS9013: Expression type can't be inferred with --isolatedDeclarations.
+!!! related TS9027 v2.ts:2:14: Add a type annotation to the variable v2.
+!!! related TS9035 v2.ts:6:5: Add satisfies and a type assertion to this expression (satisfies T as T) to make the type explicit.
+                  ~
+!!! error TS9039: Type containing private name 'n' can't be used with --isolatedDeclarations.
+!!! related TS9027 v2.ts:2:14: Add a type annotation to the variable v2.
```
jakebailey, line 74: This is all new an interesting but I think still valid.

jakebailey (approved):
I went through all of the baselines; this all seems fine, I think?

tests/baselines/reference/nodeModulesImportAttributesTypeModeDeclarationEmit(module=node16).js
weswigham: We usually elide them if they weren't needed because they matched the default mode for the file, yeah, but there's no harm (other than ascetic) in including them if they were written explicitly.

tests/baselines/reference/mappedTypeWithAsClauseAndLateBoundProperty2.js
```diff
@@ -9,101 +9,4 @@ export const thing = null;
...
-        keys?: boolean;
-        values?: boolean;
-        includes?: boolean;
-        flatMap?: boolean;
-        flat?: boolean;
-        [Symbol.iterator]?: boolean;
-        readonly [Symbol.unscopables]?: boolean;
-    };
-};
+export declare const thing: { [K in keyof number[] as Exclude<K, "length">]: (number[])[K]; };
```
weswigham, line 12: There is a distinct possibility that improvements to node reuse in this PR have made this test not test what was intended anymore (knowing if the late bound property is included in the remapped type). If we can bail out of the node reuse use via an indirection, that'd probably be a good thing.

tests/baselines/reference/isolatedDeclarationErrorsExpandoFunctions.errors.txt
```diff
@@ -16,18 +21,33 @@ isolatedDeclarationErrorsExpandoFunctions.ts(8,1): error TS9023: Assigning prope
...
     foo.call = ()=> {}
     ~~~~~~~~
 !!! error TS9023: Assigning properties to functions without declaring them is not supported with --isolatedDeclarations. Add an explicit declaration for the properties assigned to this function.
+               ~~~~~~~
+!!! error TS9007: Function must have an explicit return type annotation with --isolatedDeclarations.
+!!! related TS9030 isolatedDeclarationErrorsExpandoFunctions.ts:4:12: Add a return type to the function expression.
     foo.bind = ()=> {}
     ~~~~~~~~
 !!! error TS9023: Assigning properties to functions without declaring them is not supported with --isolatedDeclarations. Add an explicit declaration for the properties assigned to this function.
+               ~~~~~~~
```
weswigham, line 36: Isn't this error redundant? If `foo.bind` is typed (somehow - interface merge maybe), you don't need types on this lambda.

tests/baselines/reference/declarationEmitNoNonRequiredParens.js
```diff
@@ -29,4 +29,4 @@ export declare enum Test {
     C = 2
 }
 export type TestType = typeof Test;
-export declare const bar: Test[];
+export declare const bar: TestType[Extract<keyof TestType, string>][];
```
weswigham, line 32: Hm, this test probably needs some kind of extra indirection to prevent node reuse - it definitely wanted to baseline the result of this operation, not copy this operation.

tests/baselines/reference/parseImportAttributesError.js
weswigham: Now that they're not nightly-only, it should be fine to keep them around - you're more likely to randomly start breaking your consumers on older TS that you incidentally hadn't before, but... 🤷

tests/baselines/reference/mappedTypeWithAsClauseAndLateBoundProperty2.js
dragomirtitian: Added `satisfies any` which the syntactic printer will not look past

tests/baselines/reference/declarationEmitNoNonRequiredParens.js
dragomirtitian: Added `satisfies any` which the syntactic printer will not look past

tests/baselines/reference/isolatedDeclarationErrorsExpandoFunctions.errors.txt
dragomirtitian: You are right. Removed it.

tests/baselines/reference/transpile/declarationRestParameters.d.ts
dragomirtitian: I think these errors are redundant. I'll suppress them

weswigham (approved):

tests/baselines/reference/transpile/declarationRestParameters.d.ts
jakebailey: I don't think it's a requirement; the errors seem correct to me?

dragomirtitian: Yes. It's correct. It's pointing out the exact parts that are causing the issue.

jakebailey: Yeah, I wouldn't worry about suppressing them; my original comment wasn't meant to be a complaint! 😅

PR #59285, opened 2024-07-15
Fix captured shorthand properties in ES2015 loops

### LINKED ISSUE #59262: Incorrect ES5 code emitted for closures in loops when noCheck is enabled
### 🔎 Search Terms

ES5, noCheck, loop, closures, transpilation, transpileModule

This broke in either https://github.com/microsoft/TypeScript/pull/57934 or https://github.com/microsoft/TypeScript/pull/58364

### ⏯ Playground Link

https://www.typescriptlang.org/dev/bug-workbench/?target=1&noCheck=true#code/PTAEAEDsHsGEAsCmBjA1gLlAFwE4FdEAoECLAQxwHNEtNEBnAVmLHHvmgHcBRAWwEsshQsmiR6WUADNxoALygA2gF0A3ISnQcoABSjxkgG5kANgVDQpSgIwAaUACZ7AZnsAWZQEpQAb0KgA6XEAOgAHPHYdHW85AD5dH1BjM0RQAF9PT3U0kTEJUBwGPBNJBRl6YN4yUJ0ZeXiZaKzc8WgTRGCTaEodQvpirE8gA

### 💻 Code

```ts
// @noCheck: true
// @target: es5
// @showEmit

const fns = [];
for (const value of [1, 2, 3]) {
    fns.push(() => ({ value }));
}
const result = fns.map(fn => fn());
console.log(result)
```

### 🙁 Actual behavior

The following incorrect code is produced:
```ts
"use strict";
var fns = [];
for (var _i = 0, _a = [1, 2, 3]; _i < _a.length; _i++) {
    var value = _a[_i];
    fns.push(function () { return ({ value: value }); });
}
var result = fns.map(function (fn) { return fn(); });
console.log(result);
```

Which evaluates to this incorrect value: `[{ value: 3 }, { value: 3 }, { value: 3 }]`

### 🙂 Expected behavior

The following should be produced:
```ts
"use strict";
var fns = [];
var _loop_1 = function (value) {
    fns.push(function () { return ({ value: value }); });
};
for (var _i = 0, _a = [1, 2, 3]; _i < _a.length; _i++) {
    var value = _a[_i];
    _loop_1(value);
}
var result = fns.map(function (fn) { return fn(); });
console.log(result);
```

Which would evaluate to: `[{ value: 1 }, { value: 2 }, { value: 3 }]`

`transpileModule(ts, { compilerOptions: { target: ScriptTarget.ES5 } })` started producing incorrect code in TypeScript 5.5, since I believe `noCheck: true` is the default there.

jakebailey:
Fixes #59262

Two problems:

`isInExpressionContext` (and by extension `isExpressionNode`) didn't consider the case of "name only" object prop shorthand. This meant we weren't even walking into the shorthand in this example. ~Fixing this caused knock-on effects in some baselines, but I'm not sure any are actually a problem.~ They are ☹️ 

Then, we were using `getSymbolAtLocation` to get the node for the shorthand prop, but that ended up returning the wrong symbol; we want the one for the "value side" of the prop. Switching to `getResolvedSymbol` (same as the other caller of `checkIdentifierCalculateNodeCheckFlags`) seems to fix that.

Not 100% certain if this makes perf bad, though. I assume not given we already had to call the same symbol stuff via `getSymbolAtLocation`.


weswigham (approved):
I'm not particularly worried about the perf - I think I only even used `getSymbolAtLocation` over `getResolvedSymbol` because a previous WIP version passed in non-identifier nodes (or I was explicitly looking for the `ignoreErrors` flag because at the time of it being very WIP, I needed it to cut down on the error diff).

The `isExpressionNode` change is the one that's a bit more iffy, since it'll affect other callsites (like the type baselines), but if that ends up having weird knock-on effects we can't predict, we can always undo that part and put an exception into the condition in `checkSingleIdentifier` specifically (since that's the place we definitely need it to behave like this). But @rbuckton might have more opinions on that, and weather the `{ name }` in a shorthand property should truly count as an expression or not.

jakebailey:
@typescript-bot test it
@typescript-bot pack this

jakebailey:
Preemptively sending this just so we don't forget:

tests/baselines/reference/shorthandOfExportedEntity02_targetES5_CommonJS.types
```diff
@@ -16,7 +16,7 @@ export function foo () {
 >  : ^^^^^^^^^^^^^^^^^
 >{ test } : { test: string; }
 >         : ^^^^^^^^^^^^^^^^^
->test : string
+>test : "test"
```
rbuckton, line 19: I assume this was an expected change? Previously the type writer was writing the type for the property name, but now it's writing the type of the identifier reference.

rbuckton (commented):
There are quite a few types baseline changes that seem like an unintentional consequence of considering the `name` of a shorthand assignment as an expression.

tests/baselines/reference/shorthandOfExportedEntity02_targetES5_CommonJS.types
jakebailey: Seems to have no affect on hover; this test case:

image

That being said, I went through the walker, and this _does_ change the type returned by `getTypeAtLocation` for this node. So, I think I'm going to have to undo this and put in some sort of hack for this case for safety.

jakebailey: Will send a revert for the cherry-pick now.

src/compiler/checker.ts
```diff
@@ -49373,11 +49373,17 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             forEachNodeRecursively(node, checkIdentifiers);
         }
 
+        function isExpressionNodeOrShorthandPropertyAssignmentName(node: Identifier) {
+            // TODO(jakebailey): Just use isExpressionNode once that considers these identifiers to be expressions.
+            return isExpressionNode(node)
+                || isShorthandPropertyAssignment(node.parent) && (node.parent.objectAssignmentInitializer ?? node.parent.name) === node;
```
jakebailey, line 49379: Here's the targeted fix.

rbuckton (approved):

src/compiler/utilities.ts
```diff
@@ -3550,6 +3550,7 @@ export function isInExpressionContext(node: Node): boolean {
         case SyntaxKind.ExpressionWithTypeArguments:
             return (parent as ExpressionWithTypeArguments).expression === node && !isPartOfTypeNode(parent);
         case SyntaxKind.ShorthandPropertyAssignment:
+            // TODO(jakebailey): it's possible that node could be the name, too
```
jakebailey, line 3553: I plan to follow-up on this after merging; hoping that it's straightforwad to fix getTypeAtLocation to retain its old behavior.

jakebailey:

Issue #59222, opened 2024-07-10
Completions fail to be provided for type alias arguments before all required type arguments are provided


Andarist:
### 🔎 Search Terms

string completions type alias 

### 🕗 Version & Regression Information

- This is the behavior in every version I tried

### ⏯ Playground Link

https://www.typescriptlang.org/play/?ts=5.6.0-dev.20240710#code/C4TwDgpgBA4hwGED2A7AxgJ3hAQiA0hCADwAqANFKYSFBAB7AQoAmAzlANZFIBmVlUgDUAhgBsArtAZNWHUgG1qRALoA+KAF4qUAGQAoKFABKENEgwsyNQaMkQ1Abn37QkKADERASzEdtcIiomNh4NMQA3lBuEABcUABEvEhICVAAvlAAPlBRMfEJAEYiGGnplAkJTq7g0ADqFpz+sPDI6FhMYUSR0bUFyakZ2bm9kAXFpRkVCRUDVc76DGAWwLnpzkA

### 💻 Code

```ts
type GetConcreteByKey<T, TKey extends keyof T, TValue extends T[TKey]> = T &
  Record<TKey, TValue>;

type Fails = GetConcreteByKey<{ type: "foo" } | { type: "bar" }, "">;
type Works = GetConcreteByKey<{ type: "foo" } | { type: "bar" }, "", "foo">;

export {};
```

### 🙁 Actual behavior

Completions are not provided for the second type argument in `Fails`

### 🙂 Expected behavior

I'd expect this to work the same way as it works in `Works`

RyanCavanaugh:
Closing language service bugs related to the 6.0 implementation. For more information, see #62827

PR #20232, opened 2017-11-22
Always instantiate the extends clause, even in the presence of an error

weswigham:
This prevents extra downstream errors from polluting the output in `checkJs`, and also leaves the component as flexibly consumable in `allowJs`.

PR #59297, opened 2024-07-16
Keep type aliases on error types created when some type arguments are missing

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/59222


src/compiler/checker.ts
```diff
@@ -16367,7 +16367,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 error(node, diag, typeStr, minTypeArgumentCount, typeParameters.length);
                 if (!isJs) {
                     // TODO: Adopt same permissive behavior in TS as in JS to reduce follow-on editing experience failures (requires editing fillMissingTypeArguments)
```
Andarist, line 16369: This is a pretty old TODO (see https://github.com/microsoft/TypeScript/pull/20232/files#r152707677 ). Based on a quick local test removing this branch altogether and just following through with the logic would return very similar results display-wise. But then all of those would became "valid" type references etc and not error types. 

I think this is something that requires input from the team. I think though that this TODO should either be removed or implemented in this PR - I'm just not sure what's the preference of the team.

tests/baselines/reference/arrayLiteralAndArrayConstructorEquivalence1.types
```diff
@@ -20,8 +20,8 @@ var myCars3 = new Array({});
 >   : ^^
 
 var myCars4: Array; // error
->myCars4 : any
->        : ^^^
+>myCars4 : Array<any>
```
Andarist, line 23: The first iteration of this PR didn't fill missing type arguments (that were otherwise required) with error types. Personally, I like this variant that fills them with error types better: https://github.com/microsoft/TypeScript/pull/59297/commits/e13df6c1c70e55859ff4ac37b0d25cd996097bf1

tests/baselines/reference/generic.types
```diff
@@ -49,8 +49,8 @@ let c: A<boolean> = {};
 >   : ^^
 
 let d: B = { a: "" };
->d : any
->  : ^^^
+>d : "/a".A<any>
```
Andarist, line 52: this is a surprising print and it should be fixed before landing this PR

tests/baselines/reference/genericArrayWithoutTypeAnnotation.types
```diff
@@ -10,8 +10,8 @@ class Bar {
     public getBar(foo: IFoo[]) {
 >getBar : (foo: IFoo[]) => void
 >       : ^   ^^      ^^^^^^^^^
->foo : any[]
->    : ^^^^^
+>foo : IFoo<any>[]
```
Andarist, line 13: I feel like those kind of changes are especially nice here - when the error type appears as a type argument somewhere.

src/compiler/checker.ts
```diff
@@ -34619,12 +34629,12 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         return checkIndexedAccessIndexType(getFlowTypeOfAccessExpression(node, getNodeLinks(node).resolvedSymbol, indexedAccessType, indexExpression, checkMode), node);
     }
 
-    function callLikeExpressionMayHaveTypeArguments(node: CallLikeExpression): node is CallExpression | NewExpression | TaggedTemplateExpression | JsxOpeningElement {
+    function isCallLikeExpressionMayHaveTypeArguments(node: Node): node is CallExpression | NewExpression | TaggedTemplateExpression | JsxOpeningElement {
```
DanielRosenwasser, line 34632:
```suggestion
    function isCallLikeExpressionThatSupportsTypeArguments(node: Node): node is CallExpression | NewExpression | TaggedTemplateExpression | JsxOpeningElement {
```

Issue #46049, opened 2021-09-24
Can't call `protected` method from union type (but can call `private` method)


antross:
# Bug Report

### 🔎 Search Terms

union this protected

### 🕗 Version & Regression Information

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about `protected`

This appears to have always failed for `protected` methods. It used to also fail for `private` methods but started working for `private` in TS 3.9.7. Might be related to #9974 or #29337 (both of which were filed before `private` started working).

### ⏯ Playground Link

[Playground link with relevant code](https://www.typescriptlang.org/play?ts=4.4.2#code/KYDwDg9gTgLgBAYwDYEMDOa4EE4G8CwAUHHGFAJYBuKMwcAZhBABQCUAXHJROQCZ4BfIiTIRaCWvwBGKKG07c+g4XBkAvNnhUkEEAHZp4MAJ5hgAJWD1OOALxwYAC3JoA3NoemLVgHSMWrK5wAPTBcADu0ADWaB4mZpb0PjJygSFhkVAxKh66BkbOaImcTi5w9qVucYWJfkxsQaER0bHEJA41vikN6c1Zrbn6hnAArnrk+sXYAD6V5R0u7m0kYxN6tf49TZnZy6Pjk12yW2EAolBQ0ACEKkKEd0A)

### 💻 Code

```ts
export class A {
  private foo(): void {}
  protected bar(): void {}
  baz() {
    const typeRef: A = this;
    typeRef.foo(); // works
    typeRef.bar(); // works
  
    const thisRef: this = this;
    thisRef.foo(); // works
    thisRef.bar(); // works

    const unionRef: A|this = this;
    unionRef.foo(); // works
    unionRef.bar(); // Error!
  }
}
```

### 🙁 Actual behavior

Error when trying to call `protected` method from a union type (but `private` method works)

### 🙂 Expected behavior

Call to `protected` method to be allowed from the union type just like the call to the `private` method.

andrewbranch:
I’m not 100% sure that this is actually a bug, but the reasons not to copy the property outlined in #9974 don’t seem to apply when all union constituents have the same property declaration... can’t guarantee a PR would be accepted here though; we’d have to apply some extra scrutiny.

antross:
[...]

Issue #59289, opened 2024-07-16
Constructor cannot access own protected instance when typed as union


JacobLey:
### 🔎 Search Terms

protected constructor union accessible

### 🕗 Version & Regression Information

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about protected constructor accessibility

### ⏯ Playground Link

https://www.typescriptlang.org/play/?ts=5.5.3#code/MYGwhgzhAEBiD29oG8CwAoaXoAcCuARiAJbC6EBc0AdngLYECmATgNwba7PwAujwfACZd4VWgxbtM2HM2IA3MHy4Kx9Jmw7YAxAAtIutRM3otWYPGoQezPAPjMAFADNEAfioIkAH2gAhMGYAShQzTmhiZ2gXRBC0aXDwnl1iCAA6fAJoAF5oV3gMwmgAamgARilExIB6augAeWoQAE9oZNToRmobVuZGCDwQHhhiak7mbmYAQjCqtpT02SRc-IzuEvLKuex2xbl5HLzENYUNitmq3bS9A0PVm4hdM63EgF9OkAhGUITtq8zDgBZJS6NLMMDUQTwOiOIIvP4LNbLaDA5JgiFQmFwi6Jf77IEg9GQ6Gw+FzK4PJ65VGg8HErFk7CvMLM9CsjCgSAwALMToADz4kJgXhQryAA

### 💻 Code

```ts
class Foo {
    public pub: number;
    protected pro: number;
    private priv: number;
    #hash: number;

    constructor(foo?: Foo | Bar) {
        if (foo) {
            this.pub = foo.pub + 1;
            // Only this entry results in error!
            this.pro = foo.pro + 1;
            this.priv = foo.priv + 1;
            this.#hash = foo.#hash + 1;
        } else {
            this.pub = Math.random();
            this.pro = Math.random();
            this.priv = Math.random();
            this.#hash = Math.random();
        }
    }
}

class Bar extends Foo {}
```

### 🙁 Actual behavior

The line
```ts
this.pro = foo.pro + 1;
```

Throws an error
> Property 'pro' is protected and only accessible through an instance of class 'Foo'. This is an instance of class 'Foo | Bar'.

### 🙂 Expected behavior

No error.

Neither the private or hash field experience the same issue, despite having (supposedly) stricter requirements.

Supplying the type as `Foo` or `Bar` alone (opposed to `Foo | Bar`) removes this issue.

### Additional information about the issue

[...]

PR #59302, opened 2024-07-16
Allow protected property to be accessed on union of types with the correct base type

Andarist:
https://github.com/microsoft/TypeScript/issues/59289


JacobLey:
I believe this also fixes https://github.com/microsoft/TypeScript/issues/46049

Andarist:
It doesn’t 😉

Andarist:
but now it does 😉

jakebailey:
@typescript-bot test it

PR #59306, opened 2024-07-16
fix(59304): Convert to ESM uses template strings instead of string literals

### LINKED ISSUE #59304: Convert to ESM uses template strings instead of string literals
when you convert a require ``const swc = require(`@swc/core`)`` to esm with the quickfix, its gonna result in a syntax error because you can't have backticks in import statements

- VS Code Version: 1.90.0
- OS Version: Linux 6.5.0-41-generic x86_64 x86_64
image

Steps to Reproduce:

1. ``const swc = require(`@swc/core`)``
2. apply the "'require' call may be converted to an import." fix

a-tarasyuk:
Fixes #59304

### DIFF
--- a/src/services/codefixes/requireInTs.ts
+++ b/src/services/codefixes/requireInTs.ts
@@ -10,10 +10,12 @@ import {
     factory,
     first,
     getAllowSyntheticDefaultImports,
+    getQuotePreference,
     getTokenAtPosition,
     Identifier,
     ImportSpecifier,
     isIdentifier,
+    isNoSubstitutionTemplateLiteral,
     isObjectBindingPattern,
     isRequireCall,
     isVariableDeclaration,
@@ -21,10 +23,12 @@ import {
     NamedImports,
     ObjectBindingPattern,
     Program,
+    QuotePreference,
     SourceFile,
     StringLiteralLike,
     textChanges,
     tryCast,
+    UserPreferences,
     VariableStatement,
 } from "../_namespaces/ts.js";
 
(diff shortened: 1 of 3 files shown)


src/services/codefixes/requireInTs.ts
```diff
@@ -66,25 +70,27 @@ interface Info {
...
     const namedImports = isObjectBindingPattern(decl.name) ? tryCreateNamedImportsFromObjectBindingPattern(decl.name) : undefined;
     if (defaultImportName || namedImports) {
+        const moduleSpecifier = first(parent.arguments);
         return {
             allowSyntheticDefaults: getAllowSyntheticDefaultImports(program.getCompilerOptions()),
             defaultImportName,
             namedImports,
             statement: cast(decl.parent.parent, isVariableStatement),
-            required: first(parent.arguments),
+            moduleSpecifier: isNoSubstitutionTemplateLiteral(moduleSpecifier) ? factory.createStringLiteral(moduleSpecifier.text, quotePreference === QuotePreference.Single) : moduleSpecifier,
```
jakebailey, line 93: Maybe this isn't a new bug, but we reject dynamic requires like `require(`node:${a}`)`, yes?

a-tarasyuk: There is test that covers this case  https://github.com/microsoft/TypeScript/blob/ce1c83ebc38b2ebf7a0a9806f78ee5e6bfab69d9/tests/cases/fourslash/codeFixRequireInTs5.ts

DanielRosenwasser (approved):

jakebailey (approved):

PR #59310, opened 2024-07-16
Allow `this` when it appears in `this is T` positions

### LINKED ISSUE #59252: Incorrect report of `Signature '(): false' must be a type predicate` since TS 5.5
### 🔎 Search Terms

type predicate this inference object

### ⏯ Playground Link

https://www.typescriptlang.org/play/?ts=5.5.3#code/KYDwDg9gTgLgBASwHY2FAZgQwMbDgeTBgQiUwBsAeAFQD44BvAKDlbgDcKBXYAfgC441ANws2AC0wBnAGrdgACgCUgmOIRTEmwsVIUa9AGSMO8wdTgBfUZabZSU+JgBGU4CkE6SZKki4BbZzR6AF5GMVZOch5BLiQAE2B0ZGB4gBoIuElZeWVwtgK4KGAYLigkOCxyN1ECywzrJiA

### 💻 Code

```ts
export interface Optional<T> {
    value?: T;
    hasValue(): this is Optional<T> & { value: T };
}
const absent: Optional<number> = {
    value: undefined,
    hasValue() {
        return false;
    },
};
```

### 🙁 Actual behavior

```
Type '() => false' is not assignable to type '() => this is Optional<number> & { value: number; }'.
  Signature '(): false' must be a type predicate.deno-ts(2322)
file.ts(11, 5): The expected type comes from property 'hasValue' which is declared here on type 'Optional<number>'
```

### 🙂 Expected behavior

Code compiles just fine, just like it did prior to 5.5.

It seems like it is impossible to install type predicates for `this` on POJOs. That is unexpected because the code works at runtime, and it compiled with TS 5.4. At runtime, `this` simply refers to the object which the function is installed on, and it is expected that the compiler can figure this out.

Note that this code does not even work when using a class as compilation fails for
```ts
class Absent implements Optional<number> {
    value: undefined;
    hasValue() {
        return false;
    }
}
```
with a related error:

```
Property 'hasValue' in type 'Absent' is not assignable to the same property in base type 'Optional<number>'.
  Type '() => boolean' is not assignable to type '() => this is Optional<number> & { value: number; }'.
    Signature '(): boolean' must be a type predicate.
```

> This seems like a closely related issue to me, but please let me know if you'd like me to open a separate issue for it.

You can, however, add the return type fot the type predicate explicitly and do
```ts
class Absent implements Optional<number> {
    value: undefined;
    hasValue(): this is Absent & { value: number } {
        return false;
    }
}
```
which compiles just fine, but that obviously requires us to use
```ts
const absent = new Absent();
```
which was not the original goal.

Adding the same explicit type annotation to the POJO is naturally not possible because `A 'this' type is available only in a non-static member of a class or interface.`

RyanCavanaugh:
Fixes #59252


Andarist:
This has quite some commits that likely shouldn't be included in this PR :P

src/compiler/checker.ts
```diff
@@ -19673,7 +19673,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         if (isJSConstructor(container) && isNodeDescendantOf(node, container.body)) {
             return getDeclaredTypeOfClassOrInterface(getSymbolOfDeclaration(container)).thisType!;
         }
-        error(node, Diagnostics.A_this_type_is_available_only_in_a_non_static_member_of_a_class_or_interface);
+        if (node.parent.kind !== SyntaxKind.TypePredicate) {
+            error(node, Diagnostics.A_this_type_is_available_only_in_a_non_static_member_of_a_class_or_interface);
+        }
```
jakebailey, line 19678: I think we'll have to check that it's on the left side, yes? otherwise I could write `x is this`?

jakebailey (dismissed):

src/compiler/checker.ts
```diff
@@ -19673,7 +19673,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         if (isJSConstructor(container) && isNodeDescendantOf(node, container.body)) {
             return getDeclaredTypeOfClassOrInterface(getSymbolOfDeclaration(container)).thisType!;
         }
-        error(node, Diagnostics.A_this_type_is_available_only_in_a_non_static_member_of_a_class_or_interface);
+        if (!(node.parent.kind === SyntaxKind.TypePredicate && node === (node.parent as TypePredicateNode).parameterName)) {
+            error(node, Diagnostics.A_this_type_is_available_only_in_a_non_static_member_of_a_class_or_interface);
+        }
         return errorType;
```
DanielRosenwasser, line 19679: The change here avoids the error message, but it's a little weird that in this case, the type would be `errorType`.

You can see from the current behavior that if you hover over `this`, you'll see `any`, which is weird if `this` has an explicit type in a case like:

```ts
let x = {
    isYes(this: typeof x): this is Object {
        //                 ^?
        return true;
    }
}
```

Ideally none of this would go through `getTypeFromThisTypeNode` in the first place.

RyanCavanaugh: Yeah, it looks like the inbound call from `getTypeFromThisTypeNode` is bogus.

RyanCavanaugh:
@typescript-bot test this

RyanCavanaugh:
@typescript-bot test this

jakebailey (approved):

Issue #59313, opened 2024-07-17
[transpileDeclaration API] ts.transpileDeclaration always returns empty sourceMapText


MichaelMitchell-at:
### 🔎 Search Terms

transpileDeclaration source map declarationmap

### 🕗 Version & Regression Information

- This is the behavior in every version I tried

### 💻 Code

```ts
const {config} = ts.readConfigFile(project, ts.sys.readFile);

const basePath = path.resolve(path.dirname(project));

const {
    options,
    fileNames: rootNames,
} = ts.parseJsonConfigFileContent(config, ts.sys, basePath);

const program = ts.createProgram({
    rootNames,
    options,
    host: ts.createCompilerHost(options),
});

const sourceFiles = program.getSourceFiles();

for (const sourceFile of sourceFiles) {
    const declarationFile = ts.transpileDeclaration(sourceFile.text, {
        compilerOptions: options,
        fileName: sourceFile.fileName,
    });

    console.log(declarationFile.sourceMapText);
}
```

### 🙁 Actual behavior

If the TSConfig contains `"declarationMap": true`, `sourceFileText` is `undefined`.

### 🙂 Expected behavior

If the TSConfig contains `"declarationMap": true`, `sourceFileText` contains the declaration source map.

PR #59314, opened 2024-07-17
Fix ts.transpileDeclaration having empty sourceMapText when declarationMap is true

MichaelMitchell-at:
Fixes https://github.com/microsoft/TypeScript/issues/59313

### DIFF
--- a/src/services/transpile.ts
+++ b/src/services/transpile.ts
@@ -205,7 +205,7 @@ function transpileWorker(input: string, transpileOptions: TranspileOptions, decl
         addRange(/*to*/ diagnostics, /*from*/ program.getOptionsDiagnostics());
     }
     // Emit
-    const result = program.emit(/*targetSourceFile*/ undefined, /*writeFile*/ undefined, /*cancellationToken*/ undefined, /*emitOnlyDtsFiles*/ undefined, transpileOptions.transformers, /*forceDtsEmit*/ undefined);
+    const result = program.emit(/*targetSourceFile*/ undefined, /*writeFile*/ undefined, /*cancellationToken*/ undefined, /*emitOnly*/ undefined, transpileOptions.transformers, /*forceDtsEmit*/ undefined);
 
     addRange(/*to*/ diagnostics, /*from*/ result.diagnostics);
 
--- a/src/services/transpile.ts
+++ b/src/services/transpile.ts
@@ -205,7 +205,7 @@ function transpileWorker(input: string, transpileOptions: TranspileOptions, decl
         addRange(/*to*/ diagnostics, /*from*/ program.getOptionsDiagnostics());
     }
     // Emit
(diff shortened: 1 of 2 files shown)


src/services/transpile.ts
```diff
@@ -205,7 +205,7 @@ function transpileWorker(input: string, transpileOptions: TranspileOptions, decl
         addRange(/*to*/ diagnostics, /*from*/ program.getOptionsDiagnostics());
     }
     // Emit
-    const result = program.emit(/*targetSourceFile*/ undefined, /*writeFile*/ undefined, /*cancellationToken*/ undefined, /*emitOnlyDtsFiles*/ declaration, transpileOptions.transformers, /*forceDtsEmit*/ declaration);
+    const result = program.emit(/*targetSourceFile*/ undefined, /*writeFile*/ undefined, /*cancellationToken*/ undefined, /*emitOnlyDtsFiles*/ undefined, transpileOptions.transformers, /*forceDtsEmit*/ undefined);
```
MichaelMitchell-at, line 208: Before we were passing `true` to `forceDtsEmit`, which caused this to evaluate to false:
https://github.com/microsoft/TypeScript/blob/f37482cd16118409284e015409deaf14b416b4a0/src/compiler/emitter.ts#L907
though I'm not sure if this is actually the correct fix.

Changing `emitOnlyDtsFiles` to `undefined` actually doesn't seem to matter, at least with the configuration I tested with.

MichaelMitchell-at (commented):
I've verified this fixes the issue for my use case, but I'm not sure how to write a test for this. I didn't find any existing tests that I could use as a reference.

cc @weswigham

src/services/transpile.ts
```diff
@@ -205,7 +205,7 @@ function transpileWorker(input: string, transpileOptions: TranspileOptions, decl
         addRange(/*to*/ diagnostics, /*from*/ program.getOptionsDiagnostics());
     }
     // Emit
-    const result = program.emit(/*targetSourceFile*/ undefined, /*writeFile*/ undefined, /*cancellationToken*/ undefined, /*emitOnlyDtsFiles*/ declaration, transpileOptions.transformers, /*forceDtsEmit*/ declaration);
+    const result = program.emit(/*targetSourceFile*/ undefined, /*writeFile*/ undefined, /*cancellationToken*/ undefined, /*emitOnly*/ undefined, transpileOptions.transformers, /*forceDtsEmit*/ undefined);
```
MichaelMitchell-at, line 208: Before we were passing `true` to `forceDtsEmit`, which caused this to evaluate to false:
https://github.com/microsoft/TypeScript/blob/f37482cd16118409284e015409deaf14b416b4a0/src/compiler/emitter.ts#L907
though I'm not sure if this is actually the correct fix. Perhaps it would be to just remove `!forceDtsEmit && ` from `emitter.ts`.

Changing `emitOnlyDtsFiles` to `undefined` actually doesn't seem to matter, at least with the configuration I tested with.

MichaelMitchell-at: Well, no permutation of what I said results in all the tests passing, so clearly not as easy of a fix as I hoped.

sheetalkamat:
Closing in favor of https://github.com/microsoft/TypeScript/pull/59337

PR #57701, opened 2024-03-08
Use more universal ANSI sequence for 'clear screen and clear buffer'

colinta:
Fixes #57700
Fixes #29829
Fixes #57894
Closes #59315

Tiny tiny patch to bring just a little more terminal support to `tsc --watch`. In my case, iTerm2 needs this `ESC [ 3J` sequence in order to clear screen _and remove scroll history_.

If there's a setting in iTerm2 to change this, I would happily do that, but I didn't see anything.

```
  92049 passing (6m)

Finished do-runtests-parallel in 6m 22.9s
Completed runtests-parallel in 6m 57.3s
✨  Done in 417.81s.
```

PR #59315, opened 2024-07-17
Fix: tsc -w is losing working directory in macOS Terminal

### LINKED ISSUE #57894: `tsc --watch` changes shell's current directory to ~ in Terminal.app while running
### 🔎 Search Terms

watch current directory

- I was unable to test this on prior versions because it's too much work

### ⏯ Playground Link

_No response_

### 💻 Code

```sh
tsc --watch -p ./
```

### 🙁 Actual behavior

When running the above command in macOS's Terminal.app and pressing Cmd+t to open a new shell it opens a new shell at my home directory

### 🙂 Expected behavior

Terminal.app should open the new shell in the same directory that I ran the command from.

```
webpack --watch
```

works the way I expect.

Dremora:
Fixes #57894.

See [this comment](https://github.com/microsoft/TypeScript/issues/57894#issuecomment-2227449140) for the detailed explanation of the fix.

The escape sequences are taken from [Jest](https://github.com/jestjs/jest/blob/cce44d767e5e1d5c62eaaf5f93fb73748c25266b/packages/jest-util/src/specialChars.ts#L18-L20) that doesn't have the same issue.

### DIFF
--- a/src/compiler/sys.ts
+++ b/src/compiler/sys.ts
@@ -1604,7 +1604,11 @@ export let sys: System = (() => {
             setTimeout,
             clearTimeout,
             clearScreen: () => {
-                process.stdout.write("\x1Bc");
+                process.stdout.write(
+                    platform === "win32"
+                        ? "\u001B[2J\u001B[0f"
+                        : "\u001B[2J\u001B[3J\u001B[H",
+                );
             },
             setBlocking: () => {
                 const handle = (process.stdout as any)?._handle as { setBlocking?: (value: boolean) => void; };


src/compiler/sys.ts
```diff
@@ -1604,7 +1604,11 @@ export let sys: System = (() => {
             setTimeout,
             clearTimeout,
             clearScreen: () => {
-                process.stdout.write("\x1Bc");
+                process.stdout.write(
+                    platform === "win32"
```
jakebailey, line 1608: Like https://github.com/microsoft/TypeScript/pull/57701 this will be a behavior change; we'll have to check to see how it behaves and make sure we're okay with the change.

jakebailey (changes requested):
I posted a big list of tests on https://github.com/microsoft/TypeScript/pull/57701#issuecomment-2243607183; I believe that other PR to be better (but both fix your issue).

PR #59316, opened 2024-07-17
fix: filtering promise properties in object literal completion

### LINKED ISSUE #59298: Avoid providing promise methods as completions at return positions in non-async functions
### 🔎 Search Terms

completions promise return expression type async members

### ⏯ Playground Link

https://www.typescriptlang.org/play/?ts=5.6.0-dev.20240716#code/C4TwDgpgBAsiAq5oF4oG8CwAoKUBmA9gQFxQDOwATgJYB2A5gNzYC+zW2eArrQMbDUCtKMAgUAFLwBGpcQEooyAHywESKAB8oABUoEAttTIQAPHESQlCtC2zZRE+YpXjMOKAHoAVAEYvH1jk5diA

### 💻 Code

```ts
type MyType = {
  foo: string;
};

function test(cb: () => MyType | Promise<MyType>) {}

test(() => ({
  /*1*/
}));
```

### 🙁 Actual behavior

Completions provided at the marker position include `catch`, `foo` and `then`

### 🙂 Expected behavior

It would be great if in a situation like this promise methods would get at least deprioritized and if they would appear at the bottom of the list.

It's technically okay to return thenables compatible with `Promise` interface but it should be a very rare situation. I think a DX win here would justify hiding those members completely.

syi0808:
### screenshot
스크린샷 2024-07-17 오후 4 08 23

~~In the actual development environment, it was possible to reproduce what was solved.~~
~~But I can't find a case to write a test. Can someone give me an idea?~~

Fixes #59298

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -42301,7 +42301,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
      *
      * This is used to reflect the runtime behavior of the `await` keyword.
      */
-    function  getAwaitedType(type: Type, errorNode?: Node, diagnosticMessage?: DiagnosticMessage, ...args: DiagnosticArguments): Type | undefined {
+    function getAwaitedType(type: Type, errorNode?: Node, diagnosticMessage?: DiagnosticMessage, ...args: DiagnosticArguments): Type | undefined {
         const awaitedType = getAwaitedTypeNoAlias(type, errorNode, diagnosticMessage, ...args);
         return awaitedType && createAwaitedTypeIfNeeded(awaitedType);
     }
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -1637,6 +1637,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         getParameterIdentifierInfoAtPosition,
         getPromisedTypeOfPromise,
         getAwaitedType: type => getAwaitedType(type),
+        getAwaitedTypeOfPromise: type => getAwaitedTypeOfPromise(type),
         getReturnTypeOfSignature,
         isNullableType,
         getNullableType,
@@ -1861,6 +1862,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             const node = getParseTreeNode(nodeIn);
             return node && tryGetThisTypeAt(node, includeGlobalThis, container);
         },
+        filterType,
         getTypeArgumentConstraint: nodeIn => {
             const node = getParseTreeNode(nodeIn, isTypeNode);
             return node && getTypeArgumentConstraint(node);
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -1637,7 +1637,6 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         getParameterIdentifierInfoAtPosition,
         getPromisedTypeOfPromise,
         getAwaitedType: type => getAwaitedType(type),
-        getAwaitedTypeOfPromise: type => getAwaitedTypeOfPromise(type),
         getReturnTypeOfSignature,
         isNullableType,
         getNullableType,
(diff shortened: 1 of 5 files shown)


Andarist:
Tests for completions can be written like this: 
https://github.com/microsoft/TypeScript/blob/f37482cd16118409284e015409deaf14b416b4a0/tests/cases/fourslash/completionsPropertiesPriorities.ts

src/compiler/checker.ts
```diff
@@ -31176,6 +31176,14 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     return contextualAwaitedType && getUnionType([contextualAwaitedType, createPromiseLikeType(contextualAwaitedType)]);
                 }
 
+                const symbol = getSymbolForExpression(node);
+
+                if (symbol && (symbol.flags & SymbolFlags.ObjectLiteral)) {
+                    return filterType(contextualReturnType, t => {
+                        return !getAwaitedTypeOfPromise(t);
+                    });
```
Andarist, line 31184: Personally, I don't think this is the right fix. TS is a structural type at heart so this doesn't seem to be right - even if it's not overly practical for such types to contribute to contextual typing here... there is no type system rule that forbids it.

Completions are a separate thing and that could potentially be handled elsewhere (https://github.com/microsoft/TypeScript/blob/f37482cd16118409284e015409deaf14b416b4a0/src/services/completions.ts)

syi0808: Thank you very much for taking the time to review it. Oh There was a scope called "completion", i didn't know sorry.
I think this is related to autocomplete, so I think it's right to revise that part. I'll turn this PR to draft and research it more.

syi0808: I modified this inside the completion.

I'd appreciate it if you could give me your opinion when you have time!

src/compiler/types.ts
```diff
@@ -4973,6 +4973,7 @@ export interface TypeChecker {
     getPromisedTypeOfPromise(promise: Type, errorNode?: Node): Type | undefined;
     /** @internal */
     getAwaitedType(type: Type): Type | undefined;
+    getAwaitedTypeOfPromise(type: Type): Type | undefined;
```
jakebailey, line 4976: This should be internal if it's actually needed.

src/compiler/types.ts
```diff
@@ -5288,6 +5289,8 @@ export interface TypeChecker {
     tryGetThisTypeAt(node: Node, includeGlobalThis?: boolean, container?: ThisContainer): Type | undefined;
     /** @internal */ getTypeArgumentConstraint(node: TypeNode): Type | undefined;
 
+    filterType(type: Type, f: (t: Type) => boolean): Type;
```
jakebailey, line 5292: Eek, definitely not sure about putting this in the exported interface, let alone the public API...

syi0808: Okay, I will find a way to solve the completion internally without creating more public api.

jakebailey:
This looks good to me, though #59316 was accepted without determining whether or not the items should be prioritized or filtered, and this PR filters them out entirely.

I assume that ordering them is going to be a challenge, so are we okay with not providing promise-like method completions here?

andrewbranch:
@typescript-bot pack this

src/services/completions.ts
```diff
@@ -5472,9 +5472,10 @@ function getJsDocTagAtPosition(node: Node, position: number): JSDocTag | undefin
 /** @internal */
 export function getPropertiesForObjectExpression(contextualType: Type, completionsType: Type | undefined, obj: ObjectLiteralExpression | JsxAttributes, checker: TypeChecker): Symbol[] {
     const hasCompletionsType = completionsType && completionsType !== contextualType;
+    const promiseFilteredContextualType = checker.getUnionType(filter(contextualType.flags & TypeFlags.Union ? (contextualType as UnionType).types : [contextualType], t => !checker.getAwaitedType(t) || !checker.getPromisedTypeOfPromise(t)));
```
andrewbranch, line 5475: Why are both `getAwaitedType` and `getPromisedTypeOfPromise` needed?

syi0808: Looking at it again, I don't think I need the `getAwaitedType`. I deleted it. Thank you.

andrewbranch (approved):

jakebailey (approved):

PR #59324, opened 2024-07-17
Add environment variable to disable recursive watchDirectory tests

rbuckton:
One of the `symlinkWatching` tests does not appear to be reliable on Windows 11. This changes the tests to be run conditionally based on the presence or absence of a `NO_RECURSIVE_WATCHDIRECTORY_TESTS` environment variable until such time as we can determine the cause.

When `NO_RECURSIVE_WATCHDIRECTORY_TESTS` is set, it marks the test suite as "skipped" so that the related tests are reported as "pending" in the results.

### DIFF
--- a/src/testRunner/unittests/sys/symlinkWatching.ts
+++ b/src/testRunner/unittests/sys/symlinkWatching.ts
@@ -713,7 +713,7 @@ describe("unittests:: sys:: symlinkWatching::", () => {
         });
 
         if (osFlavor !== TestServerHostOsFlavor.Linux) {
-            describe("recursive watchDirectory using fsEvents", () => {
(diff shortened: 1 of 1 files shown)


jakebailey:
I think you can already use #59025 for skipping these.

rbuckton:
This may not be necessary as I believe we've isolated the issue. On Windows, it seems that file deletions against a path observed recursively within a directory junction will not trigger the `"change"` event for the containing directory unless some other file operation occurs subsequently on that directory. Therefore, if any process is also actively observing the junction (such as antivirus software), the `"change"` event will subsequently be triggered.

rbuckton:
I've filed an issue on NodeJS for the inconsistency: https://github.com/nodejs/node/issues/53903

PR #59325, opened 2024-07-17
Don't skip markLinkedReferences on ambient properties

### LINKED ISSUE #59051: TS 5.5 - Imports used by decorator output are removed
### 🔎 Search Terms

remove import compilation decorator

### ⏯ Playground Link

https://github.com/RomainLanz/typescript-output-issue

```sh
npm install
npm run build
```

### 💻 Code

```ts
import { DateTime } from "luxon";
import { column } from "@adonisjs/lucid/orm";

class User {
  @column.dateTime()
  declare createdAt: DateTime;
}
```

### 🙁 Actual behavior

When compiled the `DateTime` import is being removed (because it is only used as a type in our file), but the compilation of `@column.dateTime()` will use the import.

```ts
// ...

import { column } from "@adonisjs/lucid/orm";
class User {
}
__decorate([
    column.dateTime(),
    __metadata("design:type", typeof (_a = typeof DateTime !== "undefined" && DateTime) === "function" ? _a : Object)
], User.prototype, "createdAt", void 0);
```

### 🙂 Expected behavior

The import should be kept like in TypeScript 5.4.

```ts
import { DateTime } from "luxon";
import { column } from "@adonisjs/lucid/orm";
class User {
}
__decorate([
    column.dateTime(),
    __metadata("design:type", typeof (_a = typeof DateTime !== "undefined" && DateTime) === "function" ? _a : Object)
], User.prototype, "createdAt", void 0);
```

_No response_

jakebailey:
Fixes #59051

Alternatively, I can also only walk ambients when `emitDecoratorMetadata` is on, or only when the hint is `Unspecified` or `Decorator`.

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -29679,9 +29679,6 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         if (!canCollectSymbolAliasAccessabilityData) {
             return;
         }
-        if (location.flags & NodeFlags.Ambient) {
-            return; // References within types and declaration files are never going to contribute to retaining a JS import
-        }
         switch (hint) {
             case ReferenceHint.Identifier:
                 return markIdentifierAliasReferenced(location as Identifier);
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -29679,6 +29679,11 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         if (!canCollectSymbolAliasAccessabilityData) {
             return;
         }
+        if (location.flags & NodeFlags.Ambient && !isPropertySignature(location) && !isPropertyDeclaration(location)) {
+            // References within types and declaration files are never going to contribute to retaining a JS import,
+            // except for properties (which can be decorated).
+            return;
+        }
         switch (hint) {
             case ReferenceHint.Identifier:
                 return markIdentifierAliasReferenced(location as Identifier);
--- a/tests/baselines/reference/decoratorMetadataElidedImportOnDeclare(module=commonjs).js
+++ b/tests/baselines/reference/decoratorMetadataElidedImportOnDeclare(module=commonjs).js
@@ -1,0 +1,35 @@
+//// [tests/cases/compiler/decoratorMetadataElidedImportOnDeclare.ts] ////
+
+//// [observable.d.ts]
+export declare class Observable<T> {}
+
+//// [index.ts]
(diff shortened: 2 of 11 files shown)


jakebailey:
@typescript-bot test it

jakebailey:
Everything seems fine in the above; @weswigham is there a major downside to removing this bit?

rbuckton:
I think the only `declare` that effects runtime emit are `declare` fields with decorators. Should we preserve all ambients, or should this fix be more narrowly focused?

jakebailey:
> I think the only `declare` that effects runtime emit are `declare` fields with decorators. Should we preserve all ambients, or should this fix be more narrowly focused?

I can scope it down, the trouble is just that there's an "unspecified" hint which is hit from `noCheck` as it's not actually walking things the same way as we do the checker, so I have to disable it for that category as well. I wasn't sure how bad that'd be, but it did work fine too.

We just didn't have this optimization at all in the original code before #58366 so reverting it seemed okayish.

src/compiler/checker.ts
```diff
@@ -29679,9 +29679,6 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         if (!canCollectSymbolAliasAccessabilityData) {
             return;
         }
-        if (location.flags & NodeFlags.Ambient) {
```
weswigham, line 29682: Yeah, I wouldn't want to remove this outright - perf tests won't show it, but this is pretty important to keep this fast from my testing. I'd just except ambient property signatures from the early skip, since they're like... not really ambient? It looks like when we have something like
```ts
class A {
  @decorator(class B {})
  declare prop: string;
}
```
the `prop` is ambient marked, but the decorator is not, so we should just be able to except the prop, like @rbuckton says.

```
if (location.flags & NodeFlags.Ambient && !isPropertySignature(location) && !isPropertyDeclaration(location)) {
```
seems like it aughta work?

jakebailey:
Updated; PTAL

weswigham (approved):
Funky to me that we allow a supposedly ambient prop to be decorated at all - seems like that kinda forces it to be non-ambient 🤷‍♂️

jakebailey:

src/compiler/checker.ts
rbuckton: `declare` fields aren't *quite* as ambient as other declarations. They were added as a compatiblity mechanism due to native class fields using Define semantics, with legacy decorators being one of the main motivations.

PR #59326, opened 2024-07-17
Enable `--enforceReadonly` in TypeScript code base

ahejlsberg:
This PR enables `--enforceReadonly` in the code base and fixes inconsistencies uncovered by the stricter checking. Note that this requires an updated version of the compiler that includes #58296. Until then, it is possible to compile this branch as follows:

```ts
hereby clean
hereby tsc --no-typecheck
hereby build-src --built
```

A few of the type assertions introduced in this PR are due to inconsistences in `Node.d.ts`. Once `Node.d.ts` is updated to properly compile with `--enforceReadonly`, those assertions can be reverted.


tests/baselines/reference/api/typescript.d.ts
```diff
@@ -2821,8 +2821,8 @@ declare namespace ts {
             private hasAddedOrRemovedSymlinks;
             protected languageService: LanguageService;
             languageServiceEnabled: boolean;
-            readonly trace?: (s: string) => void;
-            readonly realpath?: (path: string) => string;
+            trace?(s: string): void;
+            realpath?(path: string): string;
```
DanielRosenwasser, line 2825: So why did these need to switch away from `readonly` (and to methods, presumably to exempt some code from `readonly` checks)?

Issue #54305, opened 2023-05-18
Need an option for ignore error when emit declaration


XHighIntell:
# Suggestion

## 🔍 noEmitOnErrorDeclaration, noEmitOnError for declaration

[noEmitOnError](https://www.typescriptlang.org/tsconfig#noEmitOnError) do not ignore error when generating declaration files (d.ts)

## ✅ Viability Checklist

My suggestion meets these guidelines:

* [x] This wouldn't be a breaking change in existing TypeScript/JavaScript code
* [x] This wouldn't change the runtime behavior of existing JavaScript code
* [x] This could be implemented without emitting different JS based on the types of the expressions
* [x] This isn't a runtime feature (e.g. library functionality, non-ECMAScript syntax with JavaScript output, new syntax sugar for JS, etc.)
* [x] This feature would agree with the rest of [TypeScript's Design Goals](https://github.com/Microsoft/TypeScript/wiki/TypeScript-Design-Goals).

## ⭐ Suggestion

I wonder if this should be a bug or a suggestion. 

`tsconfig.json` has an option `noEmitOnError`, when set to `false` compiler generates javascript files and sourcemaps even if some types are missing.

`tsconfig.json` also has a `declaration` option for generating d.ts files, when set to `true` we get errors while generating output files even when `noEmitOnError` equals `false`.

## 📃 Motivating Example

Let's look at the simplest example. A website project that contains only 2 files `file.ts` and `tsconfig.json`

- files.ts
```ts
function fn(): Coordinate {
    return { x: 0, y: 0 };
}   
```
- tsconfig.json
```json
{
    "compilerOptions": {
        "noImplicitAny": false,
        "noEmitOnError": false,
        "removeComments": false,
        "declaration": true,
        "sourceMap": true,
        "target": "es5"
    },
    "compileOnSave": true
}
```

The `Coordinate` is missing type.  We can't generate output files.
- (TS4060) Return type of exported function has or is using private name 'Coordinate'
If we turn off (declaration = false), we can generate output files.

## 💻 Use Cases

We already have an option to emitOnError for JavaScript files, we should have emitOnError for declaration.

Why did I code and build typescript missing types? 
- When I develop, I use Visual Studio with a thousand packages from npm.
[...]

PR #59332, opened 2024-07-17
Added typeToTypeNode with truncation

armanio123:
Fixes https://github.com/microsoft/TypeScript/issues/59168#issuecomment-2212598317

Make completions and formatting handle truncation gracefully. In this PR, when truncation is reported instead tsserver will replace the type with `any` type. This should also solved the issue mentioned by @weswigham in this comment https://github.com/microsoft/TypeScript/pull/58719#issuecomment-2142734097 about force truncation after 1,000,000 characters.


src/services/codefixes/helpers.ts
```diff
@@ -777,10 +796,40 @@ function createMethodImplementingSignatures(
     );
 }
 
+function getParameters(parameters: NodeArray<ParameterDeclaration>, signature: Signature, checker: TypeChecker, enclosingDeclaration: Node | undefined, flags: NodeBuilderFlags, context: TypeConstructionContext) {
+    return parameters.map((parameter, i) =>
+        parameter.type
+            && (!signature.thisParameter && i === 0)
+            && typeToTypeNode(checker, checker.getParameterType(signature, i - (signature.thisParameter ? 1 : 0)), enclosingDeclaration, flags, context).isTruncated
```
andrewbranch, line 803: I don't think I understand why we need to round-trip from a TypeNode (`parameter.type`) to a Type back to a TypeNode 🤔

armanio123: The only method I have found to make the checker re-serialize is by using `checker.typeToTypeNode`. Since `parameter.type` is `TypeNode`, I first get the type, then the TypeNode after identifying truncation.

Hope this makes sense.

andrewbranch: But why do we need to reserialize?

gabritto: I think the problem is that we're calling `checker.signatureToSignatureDeclaration` on this line: https://github.com/microsoft/TypeScript/pull/59332/files#diff-880eafae481c815c3af893123092cc0d93305284d8e6d2200638e7d546123b6cR414.

That means that the types get converted to type nodes as part of that call to `signatureToSignatureDeclaration`, and if we don't have direct control here of where the type nodes are being created, we can't replace the truncated ones.

I think a better approach to avoid serializing parameter types twice (and any other types that show up in a signature) would be to directly change the truncation behavior of the checker functions that convert things to nodes, e.g. `typeToTypeNode`, etc. So instead of inserting `...` and similar, it could insert `any`, for instance.
I think we could have a new context flag to control this, and set this new flag everywhere the relevant checker functions are being called in this file and others modified by this PR.

armanio123: You're correct @gabritto. The first time serializing is during `checker.signatureToSignatureDeclaration`. Replacing truncation there is going to be complicated becuase reporting truncation happens on high level. 

I do agree that is better to replace the `NoTruncation` behavior with using an `any` node. The challenge I see there is that by the time we have identified truncation, we have already serialized a bunch of nodes. We must then replace all of the with a single `any` node. Otherwise, we'll end up with a huge union with with an `any` in the second to last node.

I have not analyzed if that's feasible, but I can certainly look into it if it's promising.

andrewbranch: I was also wondering if the node builder could be configured to set a flag/property on node hierarchies that contain truncation.

armanio123: As per @gabritto comment, I've looked into adding a new flag that replaces the truncation behavior for replacing it with an `any` keyword. So far, the change looks good and very promising. The only problem is that we no longer have space for the NodeBuilderFlag to add a flag; I'm starting to look into that.

DanielRosenwasser: I'm slightly concerned around `any` showing up in some of these places in case a user ever asks "where did this `any` come from?"

One alternative might be to create a node like `{ /*...*/ }` which retains that comment, or even `[...elided]`.

Another approach might be to introduce a special `...` type node that is recognized by the parser. It could still be an error, but gracefully parsed and interpreted as an `errorType` in the checker.

Maybe the concern with the `any` isn't as big a deal as I'm making it out to be, but I'm wondering what others think.

andrewbranch: I don’t think we should insert a syntax error. I think adding an inline comment like `/* elided */` or `/* error */` to an `any` is fine, maybe an improvement over a silently introduced `any`. I think a new special error type node is an interesting idea and I’m not against it; however, remember that this bug comes from fuzzing, not user reports or even telemetry, so I’m not sure how much effort should go into polishing what seems to be a rare edge case.

armanio123: I kind of like the idea of prepending a comment like `/* elided */` or something similar just to indicate to the user that this node got truncated and has been transformed to `any`. Arguably this might be confusing to the user, but at least something to search.

I have not search in telemetry to figure out the recurrence of this bug in release and I don't think we will get many user reports either because this causes completions to not be displayed. Maybe users didn't notice or care to report. Nonetheless, the error-deltas recurrently finds these errors.

tests/baselines/reference/emitClassExpressionInDeclarationFile.js
```diff
@@ -107,12 +107,12 @@ export declare var simpleExample: {
...
     new (): {
-        tags(c: any): any;
+        tags(c: /*elided*/ any): /*elided*/ any;
     };
     getTags(c: {
-        tags(c: any): any;
+        tags(c: /*elided*/ any): /*elided*/ any;
     }): {
-        tags(c: any): any;
+        tags(c: /*elided*/ /*elided*/ any): /*elided*/ /*elided*/ any;
```
jakebailey, line 115: Uh oh

src/services/codefixes/helpers.ts
armanio123: I have updated the PR to instead elide in case of truncation. This doesn't add a new flag as we discussed, because I think it would be unnecessary. Plus. we already handle it like that when creating declarations, so this is consistent with that behavior.

Ping @weswigham as there's some discussion about it privately.

tests/baselines/reference/emitClassExpressionInDeclarationFile.js
armanio123: oh no! good eye. Taking a look.

weswigham (commented):
I think `createTypeNodesFromResolvedType` and the later half of `mapToTypeNodes` have references to `checkTruncationLength` that don't have a style switch like this (since they don't go through `createElidedInformationPlaceholder`), still. They need to be a bit more creative than an `any` to be grammatical, though, since they're not type node positions.

`typeToString` also has an emergency hard cutoff that slices the output and adds a trailing `...` in the event the smart truncation fails - not sure if you want to adjust that, or if you even can without introducing a hang (you definitely *have* to have a cutoff for `typeToString` at some point to avoid a hang from sending 5MB of type text across the protocol). I don't think `typeToString`'s output ever gets formatted, so it having non-grammatical output for that function specifically is probably fine.

Otherwise, this seems OK (sans duplicated comment in that one test - but that should be avoidable) - dunno if an `/*elided*/` comment is sufficient feedback for users, but if it works 🤷‍♀️

src/compiler/factory/nodeFactory.ts
```diff
@@ -7449,13 +7448,17 @@ function mergeEmitNode(sourceEmitNode: EmitNode, destEmitNode: EmitNode | undefi
     // `leadingComments` are concatenated with any existing leading comments on the destination
     if (leadingComments) {
         // We use `.slice()` in case `destEmitNode.leadingComments` is pushed to later
-        destEmitNode.leadingComments = addRange(leadingComments.slice(), destEmitNode.leadingComments);
+        for (const comment of leadingComments.slice()) {
+            destEmitNode.leadingComments = appendIfUnique(destEmitNode.leadingComments, comment);
+        }
```
jakebailey, line 7453: This doesn't seem like the right fix; what if someone is intentionally duplicating a comment?

I would expect the fix to take place somewhere other than where we add comments, rather something in the emitter.

armanio123: I understand but looks suspicious to me that we want to append the same comment. This is also consistent with the on the rest of the properties, i.e `helpers`.

I can check for another solution if you still think this is not correct.

armanio123:
@weswigham I don't understand how to fix the `createTypeNodesFromResolvedType`. We want to report an error when creating a declaration for a type correct?

i.e the code from test hugeDeclarationOutputGetsTruncatedWithError.ts shows the error `The inferred type of this node exceeds the maximum length the compiler will serialize. An explicit type annotation is needed.`
```ts
// @declaration: true
type props = "a" | "b" | "c" | "d" | "e" | "f" | "g" | "h" | "i" | "j" | "k" | "l" | "m" | "n" | "o" | "p" | "q" | "r" | "s" | "t" | "u" | "v" | "w" | "x" | "y" | "z";

type manyprops = `${props}${props}`;

export const c = [null as any as {[K in manyprops]: {[K2 in manyprops]: `${K}.${K2}`}}][0];
```
because `c` resolves to:
```ts
>c : { aa: { aa: "aa.aa"; ... 527 more ...; zz: { ...; }; }
```
we can instead resolve `c` to something like `any` or `{ [key: string]: any }` but seems to me like this might be incorrect

src/compiler/factory/nodeFactory.ts
weswigham: `addSyntheticLeadingComment` and friends are public APs, so this would be a breaking change to the API's behavior (we'd stop emitting comments as given, and start doing this editorializing) - we really should check this only at a higher level, while we're building the node ourselves.

jakebailey: Helpers are not really comparable here; those are special functions emitted at the top of the file where you only ever want one of them at a time. They're not unique to any specific node.

But, I guess here, it's `appendIfUnique` where unique is by object identity, so it's less dangerous, but it still sort of feels like a bandaid and that there's a bug somewhere else which is causing the same comment nodes to appear in two different places.

jakebailey: Sent https://github.com/microsoft/TypeScript/pull/59613 after our quick call, really wonder why the function was written this way originally?

rbuckton: Correction, it looks like the issue might stem from either `deepCloneOrReuseNode` or `setTextRange` in `createNodeBuilder()`. Both `range` and `location` already have leading comments set before `setOriginalNode` is called.

weswigham:
> We want to report an error when creating a declaration for a type correct?

Truncations already report a truncation error (via `reportTruncationError`) which the caller may/may not ignore - if the caller is ignoring it, but has indicated they need syntactically valid output by setting `NoTruncation`, we should probably emit something other than `... 527 more ...`, yes. I wouldn't go with an index signature, though - maybe... something like `/*... 527 more ...*/` as a trailing comment to the prior member?

src/compiler/factory/nodeFactory.ts
rbuckton: It looks like the issue is that `range` already has an `original` node pointing to `location` when we enter `setTextRange` (and thus already has the leading synthetic comments), but then we clone the range which produces a fresh copy of the node that already has its `original` node set to the previous range we cloned. 

We then end up calling `setOriginalNode` *again* to overwrite the `original` node for the `range` to point it at `location`.

The question is whether we need the call to `setOriginalNode` in `setTextRange` on L6062 of checker.ts, or if we should at least guard against calling it if we cloned it on L6060, or if it already has an `original` pointer set.

rbuckton: See #59647 for the fix. We can either take that and rebase, or you can apply that change here instead.

armanio123:
>  yes. I wouldn't go with an index signature, though - maybe... something like /*... 527 more ...*/ as a trailing comment to the prior member?

Do you find feasible to do `c: /* elided */ any` instead of commenting the additional ones? I think that we can check for truncation happening on `createTypeNodeFromObjectType` and use the same "elided" logic. I actually prefer to have the same behavior when NoTruncation is happening.

armanio123:
To your comment about `typeToString`, doesn't seem to me like we need to fix this method. Unless I'm understanding it wrong, the result of `typeToString` is a string to be used in errors, display parts, or other places that servers as documentation. Having being truncated by using `...` sounds more usefull.

weswigham:
> I actually prefer to have the same behavior when NoTruncation is happening.

Yeah, I generally do, too, that's why usually it all goes through the `createElidedInformationPlaceholder` helper usually - but that location has some extra formatting to include the number of skipped elements in the output (since that location isn't skipping a type node, but rather is skipping some number of property declarations), so doesn't. Seems like we aughta retain that? I think it looks a bit nicer to have the count in that position anyway, and it's not like a random type node is syntactically valid there, so you need some kind of special handling. To be clear, what I'm saying is that rather than
```ts
>c : { aa: { aa: "aa.aa"; ... 527 more ...; zz: { ...; }; }
```
when `NoTruncation` is set, we could emit
```ts
>c : { aa: { aa: "aa.aa"; /*... 527 more ...*/ zz: { /*elided*/ }; }
```
- attaching the elision text as a comment to the last successful node, rather than inserting the syntactically invalid node or a bogus property/signature.

armanio123:
Changed as requested. I still have to figure out how to handle the `zz: { /*elided*/ }` scenario.

.vscode/tasks.json
```diff
@@ -36,7 +41,10 @@
             "label": "npm: build:compiler",
             "type": "npm",
             "script": "build:compiler",
-            "group": "build",
+            "group": {
+                "kind": "build",
+                "isDefault": true
```
jakebailey, line 46: Can you undo this unrelated change?

src/compiler/factory/nodeFactory.ts
```diff
@@ -7446,16 +7445,16 @@ function mergeEmitNode(sourceEmitNode: EmitNode, destEmitNode: EmitNode | undefi
 
     // `annotatedNodes` are not merged as they should only present on the parse tree node of a `SourceFile`.
 
-    // `leadingComments` are concatenated with any existing leading comments on the destination
+    // `leadingComments` overwrites the destination
     if (leadingComments) {
         // We use `.slice()` in case `destEmitNode.leadingComments` is pushed to later
-        destEmitNode.leadingComments = addRange(leadingComments.slice(), destEmitNode.leadingComments);
+        destEmitNode.leadingComments = leadingComments.slice();
```
jakebailey, line 7451: This code should be able to be removed now that main has the merging fixed.

armanio123: Done.

.vscode/tasks.json
armanio123: Right this was never intended to be merged. Reverted.

src/compiler/checker.ts
```diff
@@ -7278,6 +7291,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 for (const type of types) {
                     i++;
                     if (checkTruncationLength(context) && (i + 2 < types.length - 1)) {
+                        if (context.flags & NodeBuilderFlags.NoTruncation) {
```
jakebailey, line 7294: Am I confused or is this truncating things _more_ when the flag is enabled?

armanio123: It is not more per se, but it is replacing the previous values with an `any`

We kinda have 3 states: when `NoTruncation` is disabled, we truncate up to `defaultMaximumTruncationLength`, if `NoTruncation` is enabled it truncates up to `noTruncationMaximumTruncationLength`, anything after that is an `any`.

src/compiler/checker.ts
```diff
@@ -7278,6 +7291,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 for (const type of types) {
                     i++;
                     if (checkTruncationLength(context) && (i + 2 < types.length - 1)) {
+                        if (context.flags & NodeBuilderFlags.NoTruncation) {
+                            return [addSyntheticLeadingComment(factory.createKeywordTypeNode(SyntaxKind.AnyKeyword), SyntaxKind.MultiLineCommentTrivia, "elided")];
```
weswigham, line 7295: Hm, this is discarding all the parts we already actually got to before hitting the truncation limit - shouldn't this all be something more like
```ts
result.push(context.flags & NodeBuilderFlags.NoTruncation ? addSyntheticLeadingComment(factory.createKeywordTypeNode(SyntaxKind.AnyKeyword), SyntaxKind.MultiLineCommentTrivia, `... ${types.length - i} more elided ...`)  : factory.createTypeReferenceNode(`... ${types.length - i} more ...`, /*typeArguments*/ undefined));
```
merged into the line below? This way we still get something like
```
A | B | C | /* ... 6 more elided ... */ any | undefined
```
out? (Part of the reason elision is a bit special here is because it endeavors to retain the last element, since elision only really makes sense in the middle.)
As written, this is just going to output
```
/* elided */ any
```
for the whole union, even if there's enough allotted space for a lot of the union.

armanio123: Including part of types then truncating with an `any` seems to me like it would be incorrect and unnecessary, even for the last type. Instead, I would argue that if we are truncating and putting `any` anyway, we just as well remove the whole thing and replace it for an any.

Arguably, adding the types might serve as documentation, but it will be incorrect as we're not including all of the types, just some of them.

armanio123:
> It'd also be nice to add a test case that actually hits the `createNotEmittedTypeElement`

I don't seem to find a way to create a declaration file with an error like truncation, so I've been testing it locally by removing the error itself. There's a thread about this ongoing here https://github.com/microsoft/TypeScript/issues/54305

src/compiler/checker.ts
armanio123: If we decide to remove all of the types and just have an `any` it will not make sense to have this message as the code would like like:
```
 foo(bar: /*... 516 more elided ...*/ any): /*elided*/ any {
    }
    get bar(): /*... 516 more elided ...*/ any {
    }
    set bar(value: /*... 516 more elided ...*/ any) {
    }
    baz<V extends /*... 516 more elided ...*/ any>(value: V): V {
    }
```

We can instead count all of the types that were elided and put that:
```
 foo(bar: /*... 800 elided ...*/ any): /*... 800 elided ...*/ any {
 }
```

src/compiler/checker.ts
weswigham: Unnecessary for the type arithmetic? Yes. `A | B | any` is just any. Unnecessary for the human that has to go and fill in the blank after the quickfix runs? Probably not! The more context we retain, the easier it is for the human that has to go and fix this fix's output what's supposed to got there!

armanio123: I changed as requested but I have some concerns.

I understand your point about retaining context, however, I respectfully disagree because the amount of code generated might become too large to be useful. I fear even the comment might go unnoticed. If this happens, including partial types and then truncating with any can lead to confusion and inaccuracies. 

I feel is more efficient to replace the entire section with `any` to maintain clarity and correctness. This way, we avoid the risk of misinterpretation and ensure the code remains manageable.

i.e take a look at the test `codeFixClassImplementInterfaceNoTruncation.ts` in the latest change, finding where the parameter type ends is already complicated. A user noticing the comment, or the introduction of `any` seems likely to be missed.

weswigham (approved):
👍 I like the consistency between quick info and the fix output now.

cc @rbuckton just to be aware of the `NotEmittedTypeElement` node kind and it being public, in case he can think of anything we should do for a new node kind that we've just forgotten (surely we'd have some kind of typesafe thing yelling at us, though, right? 😅).

> i.e take a look at the test codeFixClassImplementInterfaceNoTruncation.ts in the latest change, finding where the parameter type ends is already complicated.

Yeah, IMO, that's the point of setting `NoTruncation` - you get the full fat, complex output - as much as we can give. It's possible we may want a middleground in the future, where we want the syntactic validity but not the extra long truncation length the flag also implies, but we can consider that separately. We may just wanna make the truncation length easily configurable in the future (though we need to remember that will affect node caching).

> I fear even the comment might go unnoticed. 

Actually, that's a good idea for another new feature/followup - maybe we should issue suggestions/errors on explicitly written `any` (or `unknown`) in unions with other members, since we know they trivially collapse the union to nothing! What could you *possibly* intend when you write `A | B | any`, anyway?

> I don't seem to find a way to create a declaration file with an error like truncation, so I've been testing it locally by removing the error itself

I was thinking of triggering it via a quickfix, since that's what the new test was doing - you have to tune the length of the expansion-generated nodes pretty precisely to get it to truncate inside the explicit object you want truncated, though (you need just enough length to be below the limit, and then the object literal we want to truncate within).

src/compiler/checker.ts
```diff
@@ -7054,6 +7055,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
 
             function createTypeNodesFromResolvedType(resolvedType: ResolvedType): TypeElement[] | undefined {
                 if (checkTruncationLength(context)) {
+                    if (context.flags & NodeBuilderFlags.NoTruncation) {
```
rbuckton, line 7058: Is there a test that shows what this looks like?

armanio123: Now that you mentioned, there's no test for this scenario. I removed this check and non-regressed. I'm not sure how to test it but if it's utterly necessary I can address that another PR.

jakebailey: https://app.codecov.io/gh/microsoft/TypeScript/pull/59332?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=microsoft implies that this code is covered _somewhere_ in a test. Maybe add an assert and reverse engineer a specific test from that?

armanio123: After deeper checks, the file `hugeDeclarationOutputGetsTruncatedWithError.types` access this code.

Check line 18, for the last property `zz`. It changes from `"zz": { ...; }` to `zz: {  };`. This is a "type" tests, emit the "elided" comment gets added: `zz: { /*elided*/ };`

jakebailey: Why don't we add `elided` in type tests?

armanio123: I was working on it. Added.

jakebailey: Just to confirm, you're saying that in `*.types` files we will write `"zz": { ...; }` but in emit we will write `zz: { /*elided*/ };`? The thing I was interested in is whether or not we should do the latter for both, but maybe that's not what you were describing?

armanio123: No, I mean, before this change files in `*.types` files and during emit we write `"zz": { ...; }`.

After this change, `*.types` files and emit write `"zz": { }`, and `.d.ts` files `"zz": { /*elided*/ }` respectively.

See the test `codeFixClassImplementInterfaceNoTruncationProperties.ts` for the emit example. `hugeDeclarationOutputGetsTruncatedWithError.types` shows the `*.types` change.

jakebailey (approved):

rbuckton (approved):

PR #59339, opened 2024-07-18
Add commit characters to protocol

### LINKED ISSUE #27623: TS Server API proposal: Commit Characters
## Problem
VS Code (and Visual Studio) have the concept of commit characters for suggestions. This is a set of characters that, when typed, accept the current suggestion. For example, a suggestion for a function may treat `(` as a commit character.

VS Code currently computes these commit characters itself using the metadata that typescript returns for each completion entry. This has two problems:

- This logic is VS Code specific and may be out of sync with how VS and other editors handle commit characters.

- There are some cases where it is difficult to determine what the commit characters should be. A classic example is:

```js
const a = {
    b: 1,
    .|
}
```

Where the user has just typed `.`. Currently we show a suggestion for `b` when the user types the `.`. If the user then types another `.` as part of a spread, we incorrectly end up accepting that completion, which leaves the code as:

```js
const a = {
    b: 1,
    .b.
}
```

TypeScript, with its access to the ast, should be able to determine that `.` should not be a commit character in this case

## Proposal
Introduce the concept of a commit characters to the TypeScript server aAPI. This would be an optional set of characters returned on each `CompletionEntry`:

```ts
 interface CompletionEntry {
     ...

     commitCharacters?: ReadonlyArray<string>;
}
```

Return these commit characters on completions. Here's an approximation of how VS Code computes commit characters:

* `.` and `;` are commit characters for most completion types except for keywords and a few others.
* Additionally, `(` and `,` are commit characters for variables.
* There are no commit characters if `isNewIdentifierLocation` is set.

gabritto:
Fixes #27623.

For now, this mostly emulates the current behavior of vscode:
https://github.com/microsoft/vscode/blob/6656043f47671918a6d539c0834a31b1d84bbaf0/extensions/typescript-language-features/src/languageFeatures/completions.ts#L503
I'm not including the behavior for adding `(` as a commit character because that's currently controlled by a vscode setting.

The idea is to have follow-up PRs for computing more specific commit characters instead of e.g. simply disabling all of them whenever `isNewIdentifierLocation` is true.

Both new properties have a corresponding one in LSP: `defaultCommitCharacters` in a completion response are similar to [`itemDefaults.commitCharacters` in `CompletionList`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionList), and `commitCharacters` in a completion entry is similar to [`commitCharacters` in `CompletionItem`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItem). They are intended to work in the same way as specified by LSP: `defaultCommitCharacters` provides commit characters for all of the completion entries in the response, and `commitCharacters` provides commit characters for a specific completion entry, overriding the `defaultCommitCharacters`.


gabritto:

gabritto:
I'll try running the extended tests again when ADO comes back. 🙁

gabritto:

src/services/completions.ts
```diff
@@ -1630,7 +1661,13 @@ function completionNameForLiteral(sourceFile: SourceFile, preferences: UserPrefe
 }
 
 function createCompletionEntryForLiteral(sourceFile: SourceFile, preferences: UserPreferences, literal: string | number | PseudoBigInt): CompletionEntry {
-    return { name: completionNameForLiteral(sourceFile, preferences, literal), kind: ScriptElementKind.string, kindModifiers: ScriptElementKindModifier.none, sortText: SortText.LocationPriority };
+    return {
+        name: completionNameForLiteral(sourceFile, preferences, literal),
+        kind: ScriptElementKind.string,
+        kindModifiers: ScriptElementKindModifier.none,
+        sortText: SortText.LocationPriority,
+        commitCharacters: [],
```
andrewbranch, line 1669: Why do literals not get commit characters?

andrewbranch (approved):

PR #59347, opened 2024-07-18
Stop using latest Node in CI

jakebailey:
Per https://github.com/nodejs/node/issues/53902, Node 22.5.0 is broken; update everything to either use some version of LTS, or pin back to 22.4.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
           - windows-latest
           - macos-14
         node-version:
-          - '22'
+          - '22.4.1'
           - '20'
           - '18'
           - '16'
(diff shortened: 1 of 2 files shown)


sheetalkamat (approved):

.github/workflows/ci.yml
```diff
@@ -83,8 +83,7 @@ jobs:
       - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
       - uses: actions/setup-node@1e60f620b9541d16bece96c5465dc8ee9832be0b # v4.0.3
         with:
-          node-version: '*'
-          check-latest: true
+          node-version: 'lts/*'
```
jakebailey, line 86: FWIW these were all set to latest because they would benefit from being as fast as possible, but, I guess it's too risky.

Issue #59338, opened 2024-07-17
⚡ Performance: Project service doesn't cache all fs.statSync


JoshuaKGoldberg:
### Acknowledgement

- [X] I acknowledge that issues using this template may be closed without further explanation at the maintainer's discretion.

### 🔎 Search Terms

project service fs stat statSync

### 🙁 Actual behavior

When using [typescript-eslint's `parserOptions.projectService`](https://typescript-eslint.io/blog/announcing-typescript-eslint-v8-beta#project-service), type checking APIs switch from the traditional manual TypeScript `ts.Program` approach to the editor-style `ts.ProjectService`. We're observing excess calls to the `ts.sys.statSync` function on some paths in `node_modules/` - up to a few dozen for some paths (!).

### 🙂 Expected behavior

There should be no uncached `statSync` calls, I'd think? Even if in a persistent session, I'd expect them to be debounced in some way.

As a draft, I added a basic caching `Map` to `statSync` and ran a before & after comparison with [hyperfine](https://github.com/sharkdp/hyperfine). The results showed a ~7-12% improvement in lint time:

| Variant  | Measurement       | User Time |
| -------- | ----------------- | --------- |
| Baseline | 3.112 s ± 0.033 s | 4.382     |
| Caching  | 2.740 s ± 0.030 s | 4.032     |

<details>
<summary><code>diff</code> patch to switch to the <em>Caching</em> variant...</summary>

```diff
diff --git a/node_modules/typescript/lib/typescript.js b/node_modules/typescript/lib/typescript.js
index 4baad59..44639d5 100644
--- a/node_modules/typescript/lib/typescript.js
+++ b/node_modules/typescript/lib/typescript.js
@@ -8546,9 +8546,15 @@ var sys = (() => {
         }
       }
     };
+    const statCache = new Map();
     return nodeSystem;
     function statSync(path) {
-      return _fs.statSync(path, { throwIfNoEntry: false });
+      if (statCache.has(path)) {
+        return statCache.get(path);
+      }
+      const result = _fs.statSync(path, { throwIfNoEntry: false });
+      statCache.set(path, result);
+      return result;
     }
     function enableCPUProfiler(path, cb) {
       if (activeSession) {
```

</details>

### Additional information about the issue

On the typescript-eslint side:

* https://github.com/typescript-eslint/performance is the performance repository showing how to create these profiles.
[...]

PR #59351, opened 2024-07-18
Dont filter type aquisition when its not enabled

sheetalkamat:
This helps getting auto type directives if they wont be needed

Helps with #59338

### DIFF
--- a/src/server/project.ts
+++ b/src/server/project.ts
@@ -1629,7 +1629,7 @@ export abstract class Project implements LanguageServiceHost, ModuleResolutionHo
     protected removeExistingTypings(include: string[]): string[] {
         if (!include.length) return include;
         const existing = getAutomaticTypeDirectiveNames(this.getCompilerOptions(), this.directoryStructureHost);
-        return include.filter(i => !existing.includes(i));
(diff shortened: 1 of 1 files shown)


src/server/project.ts
```diff
@@ -1627,6 +1627,7 @@ export abstract class Project implements LanguageServiceHost, ModuleResolutionHo
     }
 
     protected removeExistingTypings(include: string[]): string[] {
+        if (!include.length) return include;
```
jakebailey, line 1630: Should this return a copy? Or is it fine to share?

sheetalkamat: Not needed since the array is just used to set the value as new include option in type aquisition. 
I am wondering if i should make this private .. not idea why this was protected.

jakebailey (approved):

src/server/project.ts
jakebailey: Could it just be `readonly string[]` at that point? (But, fine as-is)

sheetalkamat: it wont match the TypeAcquisitions input property and thats why i left it like that.

PR #59352, opened 2024-07-19
consistently check for >, >=, <, <= for unconstrained types with strictNullChecks

### LINKED ISSUE #50603: No error on unconstrained type parameter in `>` comparison

iisaduan:
fixes #50603, which bisected to #49119

Previously: unconstrained type parameters were inconsistently checked in comparisons, as shown below (5.5.3 [playground](https://www.typescriptlang.org/play/?ts=5.5.3#code/GYVwdgxgLglg9mABMOcA8AVAfACgIYBciGANIgEZEYCUiA3gFCLOIBOAplCK0nluQG4GAXwZjQkWAmSoATJkTsAHlHZgAJgGdE4ANZg4AdzC5CxMpWK1GLNp269+Q0ePDR4SFHADMC5ao1tOmFEAB9EMBAAGyiwnQ12YBgwdnVTKgsqayYWDi4eRD5BESA)): 
```typescript
function f<T>(a: T, b: T) {
   return a > b; \\ okay
}

function g<T extends {} | undefined | null>(a: T, b: T) {
   return a > b; \\ error!! 
      \\ 'a' is possibly 'null' or 'undefined'.ts(18049)
      \\ 'b' is possibly 'null' or 'undefined'.ts(18049)
}
```
That would allow us to write `f(undefined, undefined)` 

Now:

```typescript
    function f1<T>(a: T, b: T): boolean {
        return a > b;
               ~
!!! error TS18046: 'a' is of type 'unknown'.
                   ~
!!! error TS18046: 'b' is of type 'unknown'.
    }
    
    function f2<T extends {} | undefined | null>(a: T, b: T): boolean {
        return a > b;
               ~
!!! error TS18049: 'a' is possibly 'null' or 'undefined'.
                   ~
!!! error TS18049: 'b' is possibly 'null' or 'undefined'.
    }    
```
This PR does not affect the checking of unconstrained types in other locations.

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -27493,6 +27493,11 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         return type === unknownUnionType ? unknownType : type;
     }
 
+    // use to determine if a parameter may be undefined or null (or is unknown/unconstrained)
+    function getUnknownIfMaybeUnknown(type: Type) {
+        return (strictNullChecks && type.flags & TypeFlags.Instantiable) ? getBaseConstraintOfType(type) || unknownType : type;
+    }
+
     function getTypeWithDefault(type: Type, defaultExpression: Expression) {
         return defaultExpression ?
             getUnionType([getNonUndefinedType(type), getTypeOfExpression(defaultExpression)]) :
@@ -39677,8 +39682,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             case SyntaxKind.LessThanEqualsToken:
             case SyntaxKind.GreaterThanEqualsToken:
                 if (checkForDisallowedESSymbolOperand(operator)) {
-                    leftType = getBaseTypeOfLiteralTypeForComparison(checkNonNullType(leftType, left));
-                    rightType = getBaseTypeOfLiteralTypeForComparison(checkNonNullType(rightType, right));
+
+                    leftType = getBaseTypeOfLiteralTypeForComparison(checkNonNullType(getUnknownIfMaybeUnknown(leftType), left));
+                    rightType = getBaseTypeOfLiteralTypeForComparison(checkNonNullType(getUnknownIfMaybeUnknown(rightType), right));
                     reportOperatorErrorUnless((left, right) => {
                         if (isTypeAny(left) || isTypeAny(right)) {
                             return true;
--- a/tests/baselines/reference/unconstrainedTypeComparison.errors.txt
+++ b/tests/baselines/reference/unconstrainedTypeComparison.errors.txt
@@ -1,0 +1,109 @@
+unconstrainedTypeComparison.ts(2,12): error TS18046: 'a' is of type 'unknown'.
+unconstrainedTypeComparison.ts(2,16): error TS18046: 'b' is of type 'unknown'.
+unconstrainedTypeComparison.ts(6,12): error TS18049: 'a' is possibly 'null' or 'undefined'.
+unconstrainedTypeComparison.ts(6,16): error TS18049: 'b' is possibly 'null' or 'undefined'.
+unconstrainedTypeComparison.ts(10,12): error TS18046: 'a' is of type 'unknown'.
+unconstrainedTypeComparison.ts(10,16): error TS18046: 'b' is of type 'unknown'.
+unconstrainedTypeComparison.ts(14,12): error TS18046: 'a' is of type 'unknown'.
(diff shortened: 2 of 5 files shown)


iisaduan:
@typescript-bot test it 
@typescript-bot pack this

tests/baselines/reference/unconstrainedTypeComparison.errors.txt
```diff
@@ -0,0 +1,109 @@
...
+unconstrainedTypeComparison.ts(34,16): error TS18049: 'b' is possibly 'null' or 'undefined'.
+unconstrainedTypeComparison.ts(45,12): error TS18047: 'a' is possibly 'null'.
+unconstrainedTypeComparison.ts(45,16): error TS18048: 'b' is possibly 'undefined'.
+
+
+==== unconstrainedTypeComparison.ts (20 errors) ====
+    function f1<T>(a: T, b: T): boolean {
+        return a > b;
+               ~
+!!! error TS18046: 'a' is of type 'unknown'.
```
andrewbranch, line 27: This error message isn't right, though I'm not sure what. `possilby 'null' or 'undefined'` might be alright, since those are the values we're concerned about entering a comparison? And then it's consistent with a constraint of `{} | null | undefined`.

iisaduan: `possibly 'null' or 'undefined'` is the one that comes up with the other PR

src/compiler/checker.ts
```diff
@@ -27493,6 +27493,11 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         return type === unknownUnionType ? unknownType : type;
     }
 
+    // use to determine if a parameter may be undefined or null (or is unknown/unconstrained)
+    function getUnknownIfMaybeUnknown(type: Type) {
+        return (strictNullChecks && type.flags & TypeFlags.Instantiable) ? getBaseConstraintOfType(type) || unknownType : type;
```
gabritto, line 27498: I'm assuming from #59059 that making this change to `getTypeFacts` breaks a lot of existing code, is that right?

Also, does making this change to `checkNonNullType` instead also break a lot of code?

iisaduan: Yes, I tried that and putting this change direclty into `checkNonNullType` gets a very similar result to #59059 in terms of breaking object index checking

iisaduan: A lot of the cases that break relate to this type of index access https://github.com/microsoft/TypeScript/pull/59059#issuecomment-2230097569

gabritto (approved):

DanielRosenwasser:
Checking this

src/compiler/checker.ts
DanielRosenwasser: I don't know if this it's noticeably different in semantics, but istead of doing this, you might want to try to modify `getBaseTypeOfLiteralTypeForComparison` and rename it `getBaseTypeForComparison`. Then pass the result of that into `checkNonNullType`.

DanielRosenwasser: I would also consider renaming this to `getBaseConstraintOrUnknown`, use `??` instead of `||`, and just get rid of the `strictNullChecks` check.

DanielRosenwasser: Also - any clue if this works on unions of unconstrained type parameters? Can you add a test?

```ts
function f<T, U>(x: T | U, y: T | U) {
  return x < y;
}
```

DanielRosenwasser (changes requested):

iisaduan:
@DanielRosenwasser Thanks for the suggestions! Good catch--it does work for unions but not intersections, so I added those cases as well.

weswigham (approved):

gabritto:
@typescript-bot test it

ahejlsberg:
I have just put up #59437 which I believe is the right fix for #50603 plus the inconsistencies related to unconstrained type parameters vs. type parameters constrained to `{} | null | undefined`. There are an awful lot of interacting parts here. In particular, it only becomes feasible to check that unconstrained generic types are possibly nullable in `--strictNullChecks` mode by first fixing the fact that `obj` shouldn't be null-checked in `obj[key]` when `obj` and `key` both have generic types.

PR #59355, opened 2024-07-19
Disallow number type from being assignable to enums with known literal members

Andarist:
@jakebailey mentioned this type hole to me and this is an experiment to make assignability rules stricter here. I didn't find a compelling reason in the existing test suite that this rule is particularly important today and I'd be interested in learning what user tests think about that ;p


Andarist:
So the self-check isn't particularly happy with this one 😂 it's surprising to me that the test suite didn't have any tests like this with bitmasks

PR #59359, opened 2024-07-19
Unpin Node 22 now that 22.5.1 is out

jakebailey:
22.5.1 fixes the regression from 22.5.0.


jakebailey:
Okay, well check-latest isn't actually pulling 22.5.1 yet for some reason

PR #59360, opened 2024-07-19
Use tuple name inference for `Iterator.next` et al

rbuckton:
This updates our definitions of `Iterator`, `Generator`, `AsyncIterator`, and `AsyncGenerator` to leverage the tuple name inference change from #59045, as requested in https://github.com/microsoft/TypeScript/pull/58243#discussion_r1586225461


jakebailey:
Hm, the thread there seems to imply that there should be some inferred named changed, but none show in the baselines?

rbuckton:
The change is primarily observed in quickinfo/signature help, as is shown in the tests in #59045.

rbuckton:
I can add a fourslash test to illustrate this

sandersn (approved):

PR #59366, opened 2024-07-19
Manual update of generated DOM and Worker libs

rbuckton:
This is a manual insertion of updated types for `dom.generated.d.ts` et al built from https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/1713


jakebailey (approved):
Seems right; does this need an extended test? I guess not, we're all-in...

sandersn (approved):

Issue #58936, opened 2024-06-19
getPasteEdits returns wrong range for primary edit


mjbvz:
### 🔎 Search Terms

- `getPasteEdits`
- Paste with imports

### 🕗 Version & Regression Information

5.6.0-dev.20240619

### 💻 Code

`a.ts`

```ts
export interface Foo { }

export const foo: Foo = {}
```

1. On line 3, copy `Foo`
2. Paste into a file `b.ts` at the `|` (notice the blank lines before and after):

```ts

|

```

### 🙁 Actual behavior

The returned edits are at the same location:

```
    {"seq":0,"type":"response","command":"getPasteEdits","request_seq":166,"success":true,"performanceData":{"updateGraphDurationMs":1.7999577522277832},"body":{"edits":[{"fileName":"/Users/matb/projects/sandbox/b.ts","textChanges":[{"start":{"line":1,"offset":1},"end":{"line":1,"offset":1},"newText":"import { Foo } from \"./a\";\n"},{"start":{"line":1,"offset":2},"end":{"line":1,"offset":2},"newText":"Foo"}]}],"fixId":"providePostPasteEdits"}}
Info 567  [13:03:44.720] request:
    {
      "seq": 167,
      "type": "request",
      "command": "updateOpen",
      "arguments": {
        "changedFiles": [
          {
            "fileName": "/Users/matb/projects/sandbox/b.ts",
            "textChanges": [
              {
                "newText": "import { Foo } from \"./a\";\n",
                "start": {
                  "line": 1,
                  "offset": 1
                },
                "end": {
                  "line": 1,
                  "offset": 1
                }
              },
              {
                "newText": "Foo",
                "start": {
                  "line": 1,
                  "offset": 1
                },
                "end": {
                  "line": 1,
                  "offset": 1
                }
              }
            ]
          }
        ],
        "closedFiles": [],
        "openFiles": []
      }
    }
```

This causes VS Code to place the cursor incorrectly after applying the edit

### 🙂 Expected behavior

There should be two edits: the import one on line `1` and the normal paste on line `2`

PR #59369, opened 2024-07-19
Fixing range for primary edit

navya9singh:
Fixes https://github.com/microsoft/TypeScript/issues/58936 
Due to `getText()` and not `getFullText()`, new line characters were getting ignored when trying to perform a temporary file update which later made the paste at the wrong positions.

### DIFF
--- a/src/server/project.ts
+++ b/src/server/project.ts
@@ -2318,7 +2318,7 @@ export abstract class Project implements LanguageServiceHost, ModuleResolutionHo
     runWithTemporaryFileUpdate(rootFile: string, updatedText: string, cb: (updatedProgram: Program, originalProgram: Program | undefined, updatedFile: SourceFile) => void) {
         const originalProgram = this.program;
         const rootSourceFile = Debug.checkDefined(this.program?.getSourceFile(rootFile), "Expected file to be part of program");
-        const originalText = Debug.checkDefined(rootSourceFile.getText());
+        const originalText = Debug.checkDefined(rootSourceFile.getFullText());
 
         this.getScriptInfo(rootFile)?.editContent(0, originalText.length, updatedText);
         this.updateGraph();
--- a/src/server/project.ts
+++ b/src/server/project.ts
@@ -2326,7 +2326,7 @@ export abstract class Project implements LanguageServiceHost, ModuleResolutionHo
             cb(this.program!, originalProgram, (this.program?.getSourceFile(rootFile))!);
         }
         finally {
-            this.getScriptInfo(rootFile)?.editContent(0, this.program!.getSourceFile(rootFile)!.getText().length, originalText);
+            this.getScriptInfo(rootFile)?.editContent(0, this.program!.getSourceFile(rootFile)!.getFullText().length, originalText);
         }
     }
 
--- a/src/server/project.ts
+++ b/src/server/project.ts
@@ -2326,7 +2326,7 @@ export abstract class Project implements LanguageServiceHost, ModuleResolutionHo
             cb(this.program!, originalProgram, (this.program?.getSourceFile(rootFile))!);
         }
         finally {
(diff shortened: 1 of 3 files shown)


DanielRosenwasser (changes requested):
There is another call to `getText()` below that is incorrectly used. It's arguable whether or not that should just reuse `originalText`, but I honestly am not familiar enough with the code.

But something must be wrong when copying/pasting trivia over, and the new test must not cover it.

navya9singh:
> There is another call to `getText()` below that is incorrectly used. It's arguable whether or not that should just reuse `originalText`, but I honestly am not familiar enough with the code.
> 
> But something must be wrong when copying/pasting trivia over, and the new test must not cover it.

I haven't found a case where the next `getText()` would be a problem, but yes I think it's better to have that instead. 
This part of the code takes the targetFile, and adds the new, 'pasted' text to it, then performs the callback function to deal with imports. In the finally block, it adds the original text to the target file which was there before the paste took place.

sheetalkamat (approved):

src/server/project.ts
```diff
@@ -2318,15 +2318,15 @@ export abstract class Project implements LanguageServiceHost, ModuleResolutionHo
...
+        const originalText = Debug.checkDefined(rootSourceFile.getFullText());
 
         this.getScriptInfo(rootFile)?.editContent(0, originalText.length, updatedText);
         this.updateGraph();
         try {
             cb(this.program!, originalProgram, (this.program?.getSourceFile(rootFile))!);
         }
         finally {
-            this.getScriptInfo(rootFile)?.editContent(0, this.program!.getSourceFile(rootFile)!.getText().length, originalText);
+            this.getScriptInfo(rootFile)?.editContent(0, this.program!.getSourceFile(rootFile)!.getFullText().length, originalText);
```
sheetalkamat, line 2329:
```suggestion
 this.getScriptInfo(rootFile)?.editContent(0, updatedText.length, originalText);
```

sheetalkamat (changes requested):

sheetalkamat (approved):

src/server/project.ts
```diff
@@ -2318,15 +2318,15 @@ export abstract class Project implements LanguageServiceHost, ModuleResolutionHo
...
+        const originalText = Debug.checkDefined(rootSourceFile.getFullText());
 
         this.getScriptInfo(rootFile)?.editContent(0, originalText.length, updatedText);
         this.updateGraph();
         try {
             cb(this.program!, originalProgram, (this.program?.getSourceFile(rootFile))!);
         }
         finally {
-            this.getScriptInfo(rootFile)?.editContent(0, this.program!.getSourceFile(rootFile)!.getText().length, originalText);
+            this.getScriptInfo(rootFile)?.editContent(0, updatedText.length, originalText);
```
DanielRosenwasser, line 2329: Hmm... looking at it again, is it really safe to assume that `cb` doesn't itself change the contents of the file? I'm sure the way we use this function is used in our codebase, it's fine, but I just want to check. What do you think @sheetalkamat?

sheetalkamat: yes we should assume cb doesnt change the contents of file since file is never "muted" without change in script info

DanielRosenwasser: (assuming you mean mutated)

What I meant was that `cb` could itself run `editContent` somehow. If that happens, the `ScriptInfo` object may not actually contain the same text anymore.

I'm okay with assuming that for now, I just wanted to run it by you.

DanielRosenwasser (approved):

src/server/project.ts
DanielRosenwasser: The alternative is just to replace the text contents with the script info's own text contents. It should all be available anyway.

sheetalkamat: Yes I did mean mutated ☺️ Currently editing script info from callback is not possible as it’s just trying to calculate import fixes.. when that changes we can revisit this  ..

Issue #59372, opened 2024-07-20
Debug Failure caused by Invalid syntax(null)


Roise-yue:
### 🔎 Search Terms

Debug Failure, null,syntax

### 🕗 Version & Regression Information

- This is a crash happens in version 5.5.3

### 💻 Code

This program was obtained by mutating a seed program.
```ts
export const box: string
null subTitle: 
export const title: string
```

### 🙁 Actual behavior

```ts
/usr/lib/node_modules/typescript/lib/tsc.js:120127
      throw e;
      ^

Error: Debug Failure.
    at visitLabeledStatement (/usr/lib/node_modules/typescript/lib/tsc.js:107374:13)
    at topLevelNestedVisitor (/usr/lib/node_modules/typescript/lib/tsc.js:107140:16)
    at topLevelVisitor (/usr/lib/node_modules/typescript/lib/tsc.js:107114:16)
    at visitArrayWorker (/usr/lib/node_modules/typescript/lib/tsc.js:87678:49)
    at visitNodes2 (/usr/lib/node_modules/typescript/lib/tsc.js:87649:19)
    at transformCommonJSModule (/usr/lib/node_modules/typescript/lib/tsc.js:106739:26)
    at transformSourceFile (/usr/lib/node_modules/typescript/lib/tsc.js:106697:21)
    at transformSourceFileOrBundle (/usr/lib/node_modules/typescript/lib/tsc.js:89291:49)
    at transformation (/usr/lib/node_modules/typescript/lib/tsc.js:112511:14)
    at transformRoot (/usr/lib/node_modules/typescript/lib/tsc.js:112534:71)

Node.js v20.14.0
```

### 🙂 Expected behavior

TSC report syntax error.Debugging failure should be caused by the second line of the code. Removing the second line can correctly report syntax errors

### Additional information about the issue

tsconfig.json
```json
{
  "compilerOptions": {
    "target": "ESNext",
    "module": "CommonJS",
    "moduleResolution": "Node",
    "esModuleInterop": true,
    "skipLibCheck": true,
    "forceConsistentCasingInFileNames": true,
    "lib": ["es2023","dom"]
  },
  "include": [
    "TScorpus/*"
  ],
  "exclude": [
    "node_modules"
  ]
}
```

Roise-yue:
@Andarist 
(the same as https://github.com/microsoft/TypeScript/issues/59373)
[...]

PR #59374, opened 2024-07-20
Fixed a crash when transforming modules with top-level labels

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/59372

### DIFF
--- a/src/compiler/transformers/module/module.ts
+++ b/src/compiler/transformers/module/module.ts
@@ -993,7 +993,7 @@ export function transformModule(context: TransformationContext): (x: SourceFile 
         return factory.updateLabeledStatement(
             node,
             node.label,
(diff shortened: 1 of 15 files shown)


src/compiler/transformers/module/module.ts
```diff
@@ -1786,6 +1786,7 @@ export function transformModule(context: TransformationContext): (x: SourceFile
 
             // If we're exporting these variables, then these just become assignments to 'exports.x'.
             for (const variable of node.declarationList.declarations) {
+                // andarist
```
jakebailey, line 1789: andarist

Andarist: oops, my breadcrumb slipped through :P

jakebailey (approved):

Issue #59373, opened 2024-07-20
Debug failure caused by incorrect identifier and undefined type


Roise-yue:
### 🔎 Search Terms

Debug failure, identifier, reserved keyword, undefined type

### 🕗 Version & Regression Information

- This is a crash happened in version 5.5.3
- This changed between versions ______ and _______
- This changed in commit or PR _______
- This is the behavior in every version I tried, and I reviewed the FAQ for entries about _________
- I was unable to test this on prior versions because _______

### 💻 Code

This program was obtained by mutating a seed program.
```ts
interface if {
  'container': string;
  'hotkeys': string;
  'key': string;
  'keys': string;
  'list': string;
  'text': string;
  'title': string;
}
export const cssExports: CssExports;
export default cssExports;
```

### 🙁 Actual behavior

In this code, there are syntax errors in both the interface and export parts. I tried to reduce the code and found that only when both parts are present, tsc will crash, otherwise it will report syntax errors correctly
```ts
/usr/lib/node_modules/typescript/lib/tsc.js:120127
      throw e;
      ^

Error: Debug Failure.
    at visitIfStatement (/usr/lib/node_modules/typescript/lib/tsc.js:107388:13)
    at topLevelNestedVisitor (/usr/lib/node_modules/typescript/lib/tsc.js:107144:16)
    at topLevelVisitor (/usr/lib/node_modules/typescript/lib/tsc.js:107114:16)
    at visitArrayWorker (/usr/lib/node_modules/typescript/lib/tsc.js:87678:49)
    at visitNodes2 (/usr/lib/node_modules/typescript/lib/tsc.js:87649:19)
    at transformCommonJSModule (/usr/lib/node_modules/typescript/lib/tsc.js:106739:26)
    at transformSourceFile (/usr/lib/node_modules/typescript/lib/tsc.js:106697:21)
    at transformSourceFileOrBundle (/usr/lib/node_modules/typescript/lib/tsc.js:89291:49)
    at transformation (/usr/lib/node_modules/typescript/lib/tsc.js:112511:14)
    at transformRoot (/usr/lib/node_modules/typescript/lib/tsc.js:112534:71)

Node.js v20.14.0
```

### 🙂 Expected behavior

TSC shouldn't crash.

### Additional information about the issue

tsconfig.json
```json
{
  "compilerOptions": {
    "target": "ESNext",
    "module": "CommonJS",
    "moduleResolution": "Node",
    "esModuleInterop": true,
[...]

PR #59375, opened 2024-07-20
Fixed crash when transforming modules with top-level if statements incorrectly containining export statements

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/59373

### DIFF
--- a/src/compiler/transformers/module/module.ts
+++ b/src/compiler/transformers/module/module.ts
@@ -1019,7 +1019,7 @@ export function transformModule(context: TransformationContext): (x: SourceFile 
         return factory.updateIfStatement(
             node,
             visitNode(node.expression, visitor, isExpression),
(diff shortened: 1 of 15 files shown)


src/compiler/transformers/module/module.ts
```diff
@@ -1019,7 +1019,7 @@ export function transformModule(context: TransformationContext): (x: SourceFile
         return factory.updateIfStatement(
             node,
             visitNode(node.expression, visitor, isExpression),
-            Debug.checkDefined(visitNode(node.thenStatement, topLevelNestedVisitor, isStatement, factory.liftToBlock)),
+            visitNode(node.thenStatement, topLevelNestedVisitor, isStatement, factory.liftToBlock) ?? factory.createBlock([]),
```
Andarist, line 1022: alternatively, this could use `factory.createIdentifier("")` - like https://github.com/microsoft/TypeScript/pull/59374 . Producing an empty block seemed slightly better to me here but maybe the goal should be not to alter the input code too much

jakebailey (approved):

PR #59398, opened 2024-07-23
Skip markLinkedReferences import elision walk entirely in some common cases

weswigham:
This should just improve performance of js->js emit with `noCheck` on.

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -49579,11 +49579,14 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
 
         function checkSingleIdentifier(node: Node) {
             const nodeLinks = getNodeLinks(node);
-            nodeLinks.calculatedFlags |= NodeCheckFlags.ConstructorReference | NodeCheckFlags.CapturedBlockScopedBinding | NodeCheckFlags.BlockScopedBindingInLoop;
-            if (isIdentifier(node) && isExpressionNodeOrShorthandPropertyAssignmentName(node) && !(isPropertyAccessExpression(node.parent) && node.parent.name === node)) {
-                const s = getResolvedSymbol(node);
-                if (s && s !== unknownSymbol) {
-                    checkIdentifierCalculateNodeCheckFlags(node, s);
+            nodeLinks.calculatedFlags |= NodeCheckFlags.ConstructorReference;
+            if (isIdentifier(node)) {
+                nodeLinks.calculatedFlags |= NodeCheckFlags.BlockScopedBindingInLoop | NodeCheckFlags.CapturedBlockScopedBinding; // Can't set on all arbitrary nodes (these nodes have this flag set by `checkSingleBlockScopeBinding` only)
+                if (isExpressionNodeOrShorthandPropertyAssignmentName(node) && !(isPropertyAccessExpression(node.parent) && node.parent.name === node)) {
+                    const s = getResolvedSymbol(node);
+                    if (s && s !== unknownSymbol) {
+                        checkIdentifierCalculateNodeCheckFlags(node, s);
+                    }
                 }
             }
         }
--- a/src/compiler/emitter.ts
+++ b/src/compiler/emitter.ts
@@ -961,8 +961,7 @@ export function emitFiles(
     }
 
     function markLinkedReferences(file: SourceFile) {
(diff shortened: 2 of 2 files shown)


weswigham:
@typescript-bot test this

src/compiler/checker.ts
```diff
@@ -49579,11 +49579,14 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
         function checkSingleIdentifier(node: Node) {
             const nodeLinks = getNodeLinks(node);
-            nodeLinks.calculatedFlags |= NodeCheckFlags.ConstructorReference | NodeCheckFlags.CapturedBlockScopedBinding | NodeCheckFlags.BlockScopedBindingInLoop;
-            if (isIdentifier(node) && isExpressionNodeOrShorthandPropertyAssignmentName(node) && !(isPropertyAccessExpression(node.parent) && node.parent.name === node)) {
-                const s = getResolvedSymbol(node);
-                if (s && s !== unknownSymbol) {
-                    checkIdentifierCalculateNodeCheckFlags(node, s);
+            nodeLinks.calculatedFlags |= NodeCheckFlags.ConstructorReference;
+            if (isIdentifier(node)) {
+                nodeLinks.calculatedFlags |= NodeCheckFlags.BlockScopedBindingInLoop | NodeCheckFlags.CapturedBlockScopedBinding; // Can't set on all arbitrary nodes (these nodes have this flag set by `checkSingleBlockScopeBinding` only)
```
weswigham, line 49584: The references walk was implicitly calling `checkComputedPropertyName` on the computed properties, which was calculating the `CapturedBlockScopedBinding` and `BlockScopedBindingInLoop` flag on them, which was hiding the fact that preemptively setting these computed flags on *all* nodes we walk while walking only identifiers would block us from calculating them properly for computed and private names. So that's fixed here, too.

src/compiler/emitter.ts
```diff
@@ -961,6 +961,7 @@ export function emitFiles(
     }
 
     function markLinkedReferences(file: SourceFile) {
+        if (ts.isSourceFileJS(file)) return; // JS files don't use reference calculations as they don't do import ellision, no need to calculate it
```
jakebailey, line 964: Hopefully we remember to remove this if/when decorator metadata becomes a thing again with non-experimental decorators, since preserving imports used for those types I think will end up mattering even in JS files (cross this issue with that other bug I fixed).

Actually, does this work for JSX files? or does this trigger that _other_ bug where we don't realize that `import "react"` is going to happen?

jakebailey: e.g. should we copy the test from https://github.com/microsoft/TypeScript/pull/59193 and make it a .jsx file?

weswigham: In JSX files the `import "react"` we add to the output is injected by the emitter based on compiler options and doesn't depend on the refrenced-ness of the synthetic import thing in the checker (you can check pretty easily by seeing there's no `resolver.isReferencedAliasDeclaration` type calls in the jsx transform - only in the ts transform, and only if the file isn't JS).

weswigham: But yeah, should circumstances change and `shouldEmitAliasDeclaration` in `ts.ts` changes, this'd need to change to match.

weswigham: Fun fact: I tried skipping this walk for files with nothing in their `.imports` member, too (so no references to external modules), but tests showed that we *still* needed import elision calculation done even in files with no external imports, because local elision on `export import a = b` statements is done using the same logic and relies on this running to work.  And there's no simple way to know if those are present without traversing the tree anyway. 🤷‍♂️ I could have done `if (!length(file.imports) && !(file.transformFlags & TransformFlags.TypeScript)) return;` (since `export import =` is TS syntax), but how many `.ts` files don't contain any typescript syntax - hardly seems worth it.

jakebailey: Ah, I guess it doesn't matter because we don't care about exact errors in this case, yeah.

jakebailey (approved):

weswigham:
@typescript-bot cherry-pick this into release-5.5

jakebailey:
5.5.4 already went out, are we doing a 5.5.5?

DanielRosenwasser:
Yeah, I think we should try to unbreak that repro.

PR #59403, opened 2024-07-23
Add generics for getElementById etc.?

patrickjh:
I was coding today and noticed that I could cast querySelector() with generics, like querySelector<HTMLInputElement>('.my-input-class')!.value

This was nicer to do inline compared to (querySelector('.my-input-class') as HTMLInputElement).value

I tried to do the same with getElementById() but noticed that did not work

I am wondering if it is possible to make the getElementById() etc. methods generic like querySelector() so you can use generics inline like getElementById<HTMLInputElement>('someid')!.value instead of (getElementById('someid') as HTMLInputElement).value.

I just like the way it reads better.

Fixes #


jakebailey:
FWIW this is also an unsafe change we would probably not take (even if it were done via the DOM lib project); this allows you to put whatever you want in as the type argument as it's only observed in the return type.

PR #59407, opened 2024-07-24
Make typed arrays generic over ArrayBufferLike

rbuckton:
This is an experiment to make typed arrays generic over `ArrayBufferLike`, as well as to more closely align with the _TypedArray_ projections in the ECMAScript specification.


rbuckton:
Generic `TypedArray` and `TypedArrayConstructor` would need HKTs to be typed correctly. The HKT workaround introduced in this PR is not a good long-term solution.

Closing in favor of #59417

CASABECI (commented):
starting with parsley

Issue #59377, opened 2024-07-21
TS2416 nonsensical fix suggestion (and possible false positive?) in a generic class implementing an interface verbatim


ankhzet:
### 🔎 Search Terms

TS2416

### 🕗 Version & Regression Information

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about covariance in method params

### ⏯ Playground Link

https://www.typescriptlang.org/play/?ts=5.5.3#code/JYOwLgpgTgZghgYwgAgKoB4AqBGANMzAJgD5kBvAWAChlbkAHQgLgMIG5kB6T5KCAWwD2AN1ABzZGAAWwAM7JBUXgJHiC2BnChx+yGFEG7pKUJFiIUMYAA8I848mgGo1AL7Vqp6PCTIAkgBigoJYpJQ0dJCyYOh+RPgACo7WkCAAJvIYmPhxJMQAFGDM-vEMLAkAlCzCgsBpbNR0XDy0AHrICIb0wAA20MiyAK5iYnZg9oLIg7IoAAa5yakZJYSzyFLQEPgA7jIIUsj8cADWdsggkzMgM24eVAg9cLLyQSGYpMD89H38EODygWCoXIjUiYywhESiz+yyy+CIxAKRRYpXo5SqyBqdRBEVo3CaBMJROJyHaszgIAAnmsNnw9DYzg4nIpQXieO0EgZ6NAwJTkAByKJgfnIUCSSncgWvUIiuTnQRgZBPWTAMQgOAAIz6kkmDlkOhQ9C5PL5Yo1TxQvMl-MBb2I-IAdKzmrRMBKUPyIVCICkYZksPC8oViqj0cgALykLFpWXyC6K5Wq9Vay2690C2KlJI+pb+7IrRHBli5fBo5CVCNR2oxp24l10N3c+SCGCabS-Mzyfn0EUUtICntKumgTpfOBgYAp2tNfEExsehKx+UJ55JzXasBp61wguO52zwnzgU73L20VxhVK1dq9ep8XbgO7h0EbaTNLAGAwTbge9nbbAaRJBkOMDWSOQwHwDVBkVYw+S0FBBhAPhHkgNJpzoA8iSPG0iCXeMrxVG8Ux1X8BVyR0XzfD8vz4H8rT-ACDmkOV1V+MDokg6CgIgOC6UQ5DxwgNDnXcKhXCAA

### 💻 Code

```ts
interface U<T1, T2> {
[...]

PR #59409, opened 2024-07-24
Fixed canonical signature instantiations with type parameters depending on replaced type parameters

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/59377

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -16018,9 +16018,43 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     }
 
     function createCanonicalSignature(signature: Signature) {
+        // Create an instantiation of the signature where each unconstrained type parameter is replaced with
+        // its original. When a generic class or interface is instantiated, each generic method in the class or
+        // interface is instantiated with a fresh set of cloned type parameters (which we need to handle scenarios
+        // where different generations of the same type parameter are in scope). This leads to a lot of new type
+        // identities, and potentially a lot of work comparing those identities, so here we create an instantiation
+        // that uses the original type identities for all unconstrained type parameters.
+        //
+        // Whenever a type parameter is replaced with its original other type parameters must be cloned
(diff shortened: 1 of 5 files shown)


src/compiler/checker.ts
```diff
@@ -16024,9 +16024,37 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         // where different generations of the same type parameter are in scope). This leads to a lot of new type
         // identities, and potentially a lot of work comparing those identities, so here we create an instantiation
         // that uses the original type identities for all unconstrained type parameters.
+        //
+        // Whenever a type parameter is replaced with its original other type parameters must be cloned
+        // and appropriate mapper must be associated with them. This is necessary to ensure that their constraints
+        // refer to those replacement type parameters.
```
Andarist, line 16030: A simpler fix is possible (see [the first commit in this PR](https://github.com/microsoft/TypeScript/pull/59409/commits/33915949aa3a456b4659bfdb0e5e6b7c2b855fbe)) but it makes the type instantiation to go up in one of the tests.

But since it turns out that some type parameters have to be cloned after some type parameters are replaced... it feels like balancing what cases are more common etc. Effectively this tries to reduce some type identitities but some new ones have to be created. Intuitively it feels like this still is net positive though.

src/compiler/checker.ts
```diff
@@ -16024,9 +16024,37 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         // where different generations of the same type parameter are in scope). This leads to a lot of new type
         // identities, and potentially a lot of work comparing those identities, so here we create an instantiation
         // that uses the original type identities for all unconstrained type parameters.
+        //
```
DanielRosenwasser, line 16027: Maybe just create a break in the comment so it's clear the new stuff refers to the code that immediately follows.

```suggestion

```

Andarist: I changed the comment location, I hope this helps.

PR #59410, opened 2024-07-24
Fix completion entry conversion to protocol format

gabritto:
Fixes an oversight in #59339.


andrewbranch (approved):
Should we just use object rest/spread to prevent us from needing future updates here?

DanielRosenwasser:
> Member

I think the usual issue is excess stuff getting through unnecessarily, plus incorrectly spreading at the end vs. at the beginning.

PR #59412, opened 2024-07-24
Use spread when converting completion entry to protocol format

gabritto:
This avoids a problem that's happened in the past where we add a new property to the internal completion entry type, but forget to set it when converting it to the protocol completion entry format.

For instance, this happened in #59339 and was then fixed in #59410, and also in #36063 for property `isFromUncheckedFile` which I'm now fixing in this PR.

My hope is that by using object spread in the conversion of completion entries, we don't repeat those problems. The other side to this is that now we have to explicitly drop undesired properties when converting to protocol entry, but I believe that's easier to notice because if you add a new property to the internal completion entry type, it is likely that it'll show up in existing server test baselines. And this seems like a far more uncommon case than adding new properties that should be present in the protocol entry.

### DIFF
--- a/src/server/session.ts
+++ b/src/server/session.ts
@@ -2452,45 +2452,13 @@ export class Session<TMessage = string> implements EventSender {
         const prefix = args.prefix || "";
         const entries = mapDefined<CompletionEntry, protocol.CompletionEntry>(completions.entries, entry => {
             if (completions.isMemberCompletion || startsWith(entry.name.toLowerCase(), prefix.toLowerCase())) {
-                const {
-                    name,
-                    kind,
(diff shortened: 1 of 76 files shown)


gabritto:
Majority of test changes are due to different order of property `hasAction` in the test logs. Other is that now we have `isFromUncheckedFile` present where we weren't setting it before, e.g. in baseline `tests/baselines/reference/tsserver/fourslashServer/completionEntryDetailAcrossFiles01.js`

gabritto:

andrewbranch (approved):

sheetalkamat (approved):

PR #59416, opened 2024-07-24
Don’t enforce export/declare overload modifier consistency across module augmentations

### LINKED ISSUE #58756: Can't augment function overload is function is exported via `export {}`

andrewbranch:
Fixes #58756

These rules are only there because it _looks_ confusing to have overload declarations like

```ts
// what does this mean??
declare function foo(): any;
function foo(s: number): string;
export declare function foo(s: string): number;
```

But if the overloads came from a module augmentation, it's reasonable for them to have different modifiers (see code comment and tests for the differences that are allowed).

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -40731,8 +40731,8 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             // deviations, we XOR someOverloadFlags with allOverloadFlags
             const someButNotAllOverloadFlags = someOverloadFlags ^ allOverloadFlags;
             if (someButNotAllOverloadFlags !== 0) {
+                const canonicalFlags = getEffectiveDeclarationFlags(getCanonicalOverload(overloads, implementation), flagsToCheck);
                 group(overloads, o => getSourceFileOfNode(o).fileName).forEach(overloadsInFile => {
-                    const canonicalFlags = getEffectiveDeclarationFlags(getCanonicalOverload(overloads, implementation), flagsToCheck);
                     const canonicalFlagsForFile = getEffectiveDeclarationFlags(getCanonicalOverload(overloadsInFile, implementation), flagsToCheck);
                     for (const o of overloadsInFile) {
                         const deviation = getEffectiveDeclarationFlags(o, flagsToCheck) ^ canonicalFlags;
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -40716,7 +40716,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     }
 
     function checkFunctionOrConstructorSymbolWorker(symbol: Symbol): void {
-        function getCanonicalOverload(overloads: Declaration[], implementation: FunctionLikeDeclaration | undefined): Declaration {
+        function getCanonicalOverload(overloads: readonly Declaration[], implementation: FunctionLikeDeclaration | undefined): Declaration {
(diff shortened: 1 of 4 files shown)


src/compiler/checker.ts
```diff
@@ -40731,20 +40731,36 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
-                    else if (deviation & ModifierFlags.Ambient) {
-                        error(getNameOfDeclaration(o), Diagnostics.Overload_signatures_must_all_be_ambient_or_non_ambient);
-                    }
-                    else if (deviation & (ModifierFlags.Private | ModifierFlags.Protected)) {
-                        error(getNameOfDeclaration(o) || o, Diagnostics.Overload_signatures_must_all_be_public_private_or_protected);
-                    }
-                    else if (deviation & ModifierFlags.Abstract) {
-                        error(getNameOfDeclaration(o), Diagnostics.Overload_signatures_must_all_be_abstract_or_non_abstract);
+                group(overloads, o => getSourceFileOfNode(o).fileName).forEach(overloadsInFile => {
+                    const canonicalFlags = getEffectiveDeclarationFlags(getCanonicalOverload(overloads, implementation), flagsToCheck);
```
gabritto, line 40735: This computation can go outside the `forEach` since it doesn't change, right?

andrewbranch:
@typescript-bot test it

gabritto:
I feel like I don't have a good understanding of the old behavior to be able to assess the new one. Why was the old behavior in place? What are the bad scenarios we're trying to avoid, exactly?

andrewbranch:
Stating my understanding, but I might have some of the details wrong. AFAIK, overloads are additional declarations of the same method/function symbol. Since there’s ultimately only one merged symbol, all these modifiers necessarily apply uniformly to the symbol. We don’t really have mechanisms in place to reason about what it would mean for some signatures to be exported while others aren’t. And the presence or absence of `declare` on an overload signature doesn’t really have any meaning at all, since overloads aren’t a runtime/emitting construct anyway. So I think the rules are in place because the mixing of modifiers like

```ts
export function foo(x: T1): T2;
export function foo(x: T3): T4;
       function foo(x: T5): T6;
export function foo(x: any): any {
  // ...
}
```

_looks_ like it might make the third overload file-local, but there’s actually nothing in place to make that happen, so we want to enforce consistency in the modifiers.

gabritto (approved):

PR #59205, opened 2024-07-09
feat(lib): define TypedArray interface

mfulton26:
Each concrete typed array type and constructor share common interfaces.

Library types can be defined extending base interfaces:

1. higher code reuse
2. less duplication
3. easier maintenance (less error prone)

Closes #15402
Fixes #45198

PR #58573, opened 2024-05-19
Introduce ES2024 target and fix some types

petamoriken:
Fixes #54636
Past PRs #54637, #57858
Refs https://github.com/denoland/deno/issues/19191, https://github.com/denoland/deno/issues/24238

This PR includes the following changes:

## Introduce ES2024 target

* Introduce ES2024 target
* Move some ESNext types to ES2024
* **BREAKING**: Move `Atomics.waitAsync` types from ES2022 to ES2024
* Add Resizable/Transfer ArrayBuffer types to ES2024

## Fix some types

* Fix `ArrayBuffer.prototype.slice` types in ES5
* Fix `SharedArrayBuffer.prototype.slice` types in ES2017
* Add empty arguments of  `ArrayBuffer`/`SharedArrayBuffer` constructor types to ES2017

PR #59417, opened 2024-07-24
Make typed arrays generic over `ArrayBufferLike`

rbuckton:
Prior to ES2024, the interface for `SharedArrayBuffer` was assignment-compatible with `ArrayBuffer`, which meant that APIs that declared that they accepted an `ArrayBuffer` would blindly accept a `SharedArrayBuffer` without complaining, despite `SharedArrayBuffer` being an invalid argument at runtime. 

Starting with ES2024, both `ArrayBuffer` and `SharedArrayBuffer` diverge significantly due to `ArrayBuffer` now being resizable, and `SharedArrayBuffer` being growable:

```ts
// es2024.arraybuffer.d.ts
interface ArrayBuffer {
    get maxByteLength(): number;
    get resizable(): boolean;
    get detached(): boolean;
    resize(newByteLength?: number): void;
    transfer(newByteLength?: number): ArrayBuffer;
    transferToFixedLength(newByteLength?: number): ArrayBuffer;
}

// es2024.sharedmemory.d.ts
interface SharedArrayBuffer {
    get maxByteLength(): number;
    get growable(): number;
    grow(newByteLength?: number): void;
}
```

Where this becomes a problem is that many projects are using typed arrays, such as `Uint8Array`, and passing along the `.buffer` property to an API that only accepts `ArrayBuffer`:

```js
const u8array = new Uint8Array(bytes);
const buffer = u8array.buffer; // ArrayBuffer | SharedArrayBuffer
const hash = crypto.subtle.digest("SHA-512", buffer);
```

With these two interfaces diverging, existing code will report errors due to the union on `.buffer`.

To address this, we propose introducing a `Buffer` type parameter on all typed array types (as well as `DataView`), as follows:

```ts
interface Uint8Array<Buffer extends ArrayBufferLike = ArrayBufferLike> {
  ...
  readonly buffer: Buffer;
  filter(predicate: (value: number, index: number, array: Uint8Array<Buffer>) => any, thisArg?: any): Uint8Array<ArrayBuffer>;
  ...
  subarray(begin?: number, end?: number): Uint8Array<Buffer>;
  ...
}

interface Uint8ArrayConstructor {
    readonly prototype: Uint8Array<ArrayBufferLike>;
    readonly BYTES_PER_ELEMENT: number;

    new (length: number): Uint8Array<ArrayBuffer>;
    new <Buffer extends ArrayBufferLike>(buffer: Buffer, byteOffset?: number, length?: number): Uint8Array<Buffer>;
    new (array: ArrayLike<number> | ArrayBuffer): Uint8Array<ArrayBuffer>;
    new (array: ArrayLike<number> | ArrayBufferLike): Uint8Array<ArrayBufferLike>;

    of(...items: number[]): Uint8Array<ArrayBuffer>;
    from(arrayLike: ArrayLike<number>): Uint8Array<ArrayBuffer>;
    from<T>(arrayLike: ArrayLike<T>, mapfn: (v: T, k: number) => number, thisArg?: any): Uint8Array<ArrayBuffer>;
}
...
```

This has the following benefits:
- `.buffer` becomes generic over the underlying `ArrayBufferLike`
- Simple construction of a TypedArray produces an instance backed by an `ArrayBuffer`, as per spec:
  ```ts
  const u8array = new Uint8Array(bytes); // Uint8Array<ArrayBuffer>
  ```
- Constructing a TypedArray with a `SharedArrayBuffer` is properly generic:
  ```ts
  const u8array = new Uint8Array(new SharedArrayBuffer(16)); // Uint8Array<SharedArrayBuffer>
  ```
- TypedArray methods which produce fresh arrays are properly backed by `ArrayBuffer`, as per spec:
  ```ts
  const u8array = new Uint8Array(new SharedArrayBuffer(16)); // Uint8Array<SharedArrayBuffer>
  const result = u8array.map(x => x * 2); // Uint8Array<ArrayBuffer>
  ```
- Sub-array views over the same `SharedArrayBuffer` are properly generic:
  ```ts
  const u8array = new Uint8Array(new SharedArrayBuffer(16)); // Uint8Array<SharedArrayBuffer>
  const result = u8array.subarray(0, 8); // Uint8Array<SharedArrayBuffer>
  ```
- Resizing or growing an `ArrayBuffer` or `SharedArrayBuffer` can be accomplished through the `.buffer` property without the need for a type guard:
  ```ts
  const u8array = new Uint8Array(new ArrayBuffer(16), { maxByteLength: 1024 }); // Uint8Array<ArrayBuffer>
  u8array.buffer.resize(512);
  u8array.buffer.resize(256);
  
  const su8array = new Uint8Array(new SharedArrayBuffer(16), { maxByteLength: 1024 }); // Uint8Array<SharedArrayBuffer>
  su8array.buffer.grow(512);
  ```
- Existing APIs that accept non-generic TypedArray instances will fall back to `ArrayBufferLike` by default, which should be compatible with prior versions TypeScript.

Please note that the changes in this PR are necessary to unblock the introduction of the `ES2024` target added by #58573.

Also note that the introduction of the `Buffer` type parameter would preclude our ability to define common `TypedArray` and `TypedArrayConstructor` interfaces, such as is proposed in #59205. For `TypedArrayConstructor` and `TypedArray` to be properly generic over `Buffer` would require either Higher-Kinded Types or a [complex HKT workaround](https://github.com/microsoft/TypeScript/pull/59407/files?show-viewed-files=true&file-filters%5B%5D=#diff-bc5657cc7554ca0c944ccc883a1334694829bce1293fd4fc5f2059cada6c003fR1855-R1870) to be able to properly type construction. HKTs are not likely to be addressed in a near enough time frame to block the introduction of an ES2024 target, and there is little appetite to ship an HKT workaround in our default libs purely to address the TypedArray use case. As such, this PR must close #59205 until such time as an acceptable solution to the HKT concern is presented.

Closes #59205
Closes #59407
Related #58573


rbuckton:
@typescript-bot perf test
@typescript-bot run dt
@typescript-bot test top400

src/lib/es2015.iterable.d.ts
```diff
@@ -497,13 +553,20 @@ interface Float64Array {
...
      */
-    from(arrayLike: Iterable<number>, mapfn?: (v: number, k: number) => number, thisArg?: any): Float64Array;
+    from(arrayLike: Iterable<number>): Float64Array<ArrayBuffer>;
+    /**
+     * Creates an array from an array-like or iterable object.
+     * @param arrayLike An array-like or iterable object to convert to an array.
+     * @param mapfn A mapping function to call on every element of the array.
+     * @param thisArg Value of 'this' used to invoke the mapfn.
+     */
+    from<T>(arrayLike: Iterable<T>, mapfn?: (v: T, k: number) => number, thisArg?: any): Float64Array<ArrayBuffer>;
```
gabritto, line 571: How breaky is this change in particular? I'm guessing if people are calling e.g.  `int32array.from(iterable, (...) => { ... })`, we'd pretty much always be able to infer something good for `T`, but I'm no inference expert.

gabritto:
@typescript-bot pack this

gabritto:
> Hey @rbuckton, the results of running the DT tests are ready.
> 
> There were interesting changes:
> 
> Changes are too big to display here, please check [the log](https://typescript.visualstudio.com/TypeScript/_build/results?buildId=163421&view=logs&j=275f1d19-1bd8-5591-b06b-07d489ea915a&t=40b1ee41-44d6-5bba-aa04-4b76a5c732e5).
> 
> [You can check the log here](https://typescript.visualstudio.com/TypeScript/_build/index?buildId=163421&_a=summary).

Most errors seem to be due to ExpectType changes, but at least one [in uuid](https://typescript.visualstudio.com/TypeScript/_build/results?buildId=163421&view=logs&j=275f1d19-1bd8-5591-b06b-07d489ea915a&t=40b1ee41-44d6-5bba-aa04-4b76a5c732e5&l=360) was not, and instead was due to `class CustomBuffer extends Uint8Array {}`.

rbuckton:
The "Base constructors must all have the same return type" error is slightly problematic as we don't currently have a good way for a class to declare custom return types in constructor overloads.

I can address this in the DefinitelyTyped tests by changing this:
```ts
class CustomBuffer extends Uint8Array {}
```
to this:
```ts
interface CustomBuffer extends Uint8Array {}
interface CustomBufferConstructor extends Uint8ArrayConstructor {}
declare var CustomBuffer: CustomBufferConstructor;
```
which is compatible with all TS versions.

Code that is targeting TS 5.7 could change to any of these instead:
```ts
// if only ArrayBuffer is supported
class CustomBuffer extends Uint8Array<ArrayBuffer> {
}

// if either are supported
class CustomBuffer<Buffer extends ArrayBufferLike = ArrayBufferLike> extends Uint8Array<Buffer> {
}
```

src/lib/es2015.iterable.d.ts
rbuckton: This shouldn't break any existing code. Any place that previously would have passed an iterable or arraylike for `number` will still get `number` values. This is actually more flexible than what we had, and was already being used on BigInt64Array.

rbuckton:
I think I have a way to satisfy the "Base constructors must all have the same return type" error. I'll have an update shortly.

rbuckton:
@typescript-bot run dt
@typescript-bot test top400

rbuckton:

rbuckton:

rbuckton:
@typescript-bot: pack this

rbuckton:
It looks like `effect` is running into an issue with the NodeJS types because `Buffer` has a few members that make it incompatible with `Uint8Array<ArrayBufferLike>` when comparing structurally:
- in `Buffer`, `slice()` returns `Buffer`, while in `Uint8Array` it returns `Uint8Array<ArrayBuffer>`
- in `Buffer`, `subarray()` returns `Buffer`, while in `Uint8Array` it returns `Uint8Array<TArrayBuffer>`
- The `array`/`obj` arguments in various `Uint8Array` callbacks like `map`/`filter`/etc. specify `Uint8Array<TArrayBuffer>`, which triggers assignability errors from other methods like `slice()` and `subarray()`.

The `array`/`obj` argument issue can be addressed by using `this` types. For `slice()` and `subarray()` this can either be addressed by either making `Buffer` generic over `TArrayBuffer`, or changing the return types of `slice()` and `subarray()`. Making `Buffer` generic is the more reliable option, but would require reintroducing a `"typesVersions"` entry to the NodeJS types to handle the inconsistency. A less invasive and more backwards compatible fix would be a minor change to the return types of `slice()` and `subarray()` in `Buffer`:

```ts
interface WithArrayBufferLike<TArrayBuffer extends ArrayBufferLike> {
    readonly buffer: TArrayBuffer;
}
interface Buffer extends Uint8Array {
    ...
    slice(start?: number, end?: number): Buffer & WithArrayBufferLike<ArrayBuffer>;
    subarray(start?: number, end?: number): Buffer & WithArrayBufferLike<this["buffer"]>;
    ...
}
```

I've tested both options with `effect` and either would resolve the issues.

rbuckton:
A fix for both the failing DefinitelyTyped tests and the necessary changes to `Buffer` to address the errors reported in `effect` can be found here: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/70390

jakebailey:
Based on the design meeting, is it the case that we just don't have any better options?

If there's a future where we gain HKT, would the type parameters on these declarations change? Once we add them, I would think we're stuck with them for good.

rbuckton:
> Based on the design meeting, is it the case that we just don't have any better options?

IIRC, the only real alternative we discussed in the design meeting was leaving the typed arrays as-is and making the DOM and NodeJS types more lenient, but that doesn't help `google-webfonts-helper`, `mapbox-gl-js`, `pixijs`, or [any other project](https://github.com/microsoft/TypeScript/pull/58573#issuecomment-2246391435) that depends on a 3rd party package that specifically expects `ArrayBuffer`. This change results in a far more accurate type that addresses DOM, NodeJS, and 3rd party package types.

> If there's a future where we gain HKT, would the type parameters on these declarations change? Once we add them, I would think we're stuck with them for good.

I investigated how HKTs might affect typed arrays and we would still have this type parameter in this position for the actual typed array instance types, so HKTs wouldn't change anything here, they would only make the declarations less repetitive.

jakebailey:
@typescript-bot test it

jakebailey:
I assume https://github.com/DefinitelyTyped/DefinitelyTyped/pull/70390 fixes everything above, right?

jakebailey (approved):
Approving because I believe this matches the outcome of the design meeting and I know you want to get this all through.

rbuckton:
> I assume [DefinitelyTyped/DefinitelyTyped#70390](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/70390) fixes everything above, right?

I believe so, yes.

psnet:
@gabritto @jakebailey @DanielRosenwasser @weswigham @rbuckton please read https://github.com/microsoft/TypeScript/issues/60745 carefully. Who tested this improvements?

Do you imagine how wide changes need to do across multiple projects to fix one old caveat with union types and this new one that combined together are real mess?

meszaros-lajos-gyorgy:
I think I'm finally at the correct place to find info. So this was working before (5.6.3): `class X extends Dataview {}`, but now(5.7.2) it says: `error TS2508: No base constructor has the specified number of type arguments.`

What is expected here, I don't understand the message. Is DataView now a generic?

PR #59418, opened 2024-07-25
Use open file to determine file existence

### LINKED ISSUE #59119: Renaming files imported with alias/absolute paths does not update the path

- VS Code Version: 1.90.2
- OS Version: macOS 14.1.1 (23B81)

Steps to Reproduce:

1. Create a project with alias imports (e.g. [setting up a new Nextjs](https://nextjs.org/learn-pages-router/basics/create-nextjs-app/setup) project and configure to use @/ alias imports)
2. Create a new typescript file
3. Import this file, and ensure to import it using your `@/` alias
4. Rename the created typescript file
5. The error is that the reference in the import is not updated, and now vscode is complaining it cannot find the file

See this short video:

https://github.com/microsoft/vscode/assets/3786627/8fb2ffd2-af46-4664-a44a-675649b17543

tsconfig:
--
```json
{
  "compilerOptions": {
    "lib": [
      "dom",
      "dom.iterable",
      "esnext"
    ],
    "allowJs": true,
    "skipLibCheck": true,
    "strict": true,
    "noEmit": true,
    "esModuleInterop": true,
    "module": "esnext",
    "moduleResolution": "bundler",
    "resolveJsonModule": true,
    "isolatedModules": true,
    "jsx": "preserve",
    "incremental": true,
    "plugins": [
      {
        "name": "next"
      }
    ],
    "baseUrl": "./src",
    "paths": {
      "@/*": [
        "./*"
      ]
    },
  },
  "include": [
    "next-env.d.ts",
    "**/*.ts",
    "**/*.tsx",
    ".next/types/**/*.ts"
  ],
  "exclude": [
    "node_modules"
  ]
}
```

Possibly related:
- https://github.com/microsoft/TypeScript/issues/27959
- https://github.com/microsoft/vscode/issues/97135

sheetalkamat:
- File exists if it is open
- Invoke wild card watchers if the info was not present before opening - so the project can update with new file existence if it didn't account for that earlier

Fixes #59119

### DIFF
--- a/src/compiler/watchUtilities.ts
+++ b/src/compiler/watchUtilities.ts
@@ -486,8 +486,8 @@ export function updateMissingFilePathsWatch(
 }
 
 /** @internal */
-export interface WildcardDirectoryWatcher {
-    watcher: FileWatcher;
+export interface WildcardDirectoryWatcher<T extends FileWatcher = FileWatcher> {
+    watcher: T;
     flags: WatchDirectoryFlags;
 }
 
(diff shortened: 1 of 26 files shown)


jakebailey (approved):

DanielRosenwasser:
If open files determines whether a file exists, does that mean that closing a file may require invalidating some program state? I'm thinking about the case where:

1. A file is opened.
2. The file is deleted on disk, but it is open in the editor.
3. The file is closed

sheetalkamat:
We already dp that even before this change. That is we do invalidate projects that file is part of if file does not exist.

andrewbranch (changes requested):
This doesn’t fix #59119 for me. Sending you the log via Teams.

jakebailey:
LGTM but if Andrew has manual tests I'd defer to him.

andrewbranch (approved):

Issue #59241, opened 2024-07-10
Add setting to exclude constructors from definition list


EwenDC:
I am writing a Playwright test suite in TypeScript using VSCode. We are following a standard pattern of using classes wrapped around a Playwright page object that abstract common workflows. For convenience we have created an extremely basic base class that all other classes are extending, which allows us to do inline property definition and omit constructors from the other classes entirely.
```typescript
export class PageWrapper {
  constructor(protected readonly page: Page) {}
}
``` 
```typescript
export class LoginPage extends PageWrapper {
  readonly email = this.page.getByLabel('Email')
  readonly password =  this.page.getByLabel('Password')
  /* More code here */
}
```

However, when trying to jump to the class from where we're defining it within test code (like below), VSCode prioritises jumping to the dummy constructor of `PageWrapper` instead of the actual class of `LoginPage` (though will still show both in the definitions list).
```typescript
const loginPage = new LoginPage(page)
```

I understand the logic behind this behaviour (since this is technically an invocation of the `PageWrapper` constructor function), but in an ES6+ codebase I think the constructor definition is less important than the class definition. After all, if we needed to see the constructor code in this instance, we can follow the class dependency chain down to the invoked constructor.

I propose a new setting that allows constructors from being excluded from the definitions list when jumping to definition from a constructor invocation. I don't know whether a setting like this is necessary for languages other than JavaScript, or if this behaviour is just a byproduct of how JavaScript internally treats constructors.

mjbvz:
Possibly could be handled by `go to type definition` or `go to declaration` (which VS Code supports but TS doesn't implement)

RyanCavanaugh:
I don't even think this needs to be a setting. Going to `new Foo` should take you to `class Foo {` if it has an implicit constructor; from there you can jump to the base class if needed. @mjbvz agree?

mjbvz:
Yes I don't think we should add a setting. For an explicit constructor, my first feeling was that `go to type definition` should jump to the class. Not as sure about implicit/inherited ctors. Let's discuss this at our next sync

RyanCavanaugh:
Agreed in the VS Code sync to show both the constructor (as the first result) and the class definition (as the second result) when the constructor definition isn't a child of the class definition

iisaduan:
[...]

PR #59421, opened 2024-07-25
also return classes when caling goToDef on a constructor call

iisaduan:
When requesting definition on a non-extended class,
```typescript
class Foo {
    /*
     * possibly a lot of code
     */
    constructor()
}
const foo = new /*goToDefinition*/Foo()
```
we now also return the class definition, in addition to the constructor.

Described here/generated by/inspired by https://github.com/microsoft/TypeScript/issues/59241#issuecomment-2248737599 (related #59241)

### DIFF
--- a/src/services/goToDefinition.ts
+++ b/src/services/goToDefinition.ts
@@ -229,16 +229,18 @@ export function getDefinitionAtPosition(program: Program, sourceFile: SourceFile
     // Don't go to the component constructor definition for a JSX element, just go to the component definition.
     if (calledDeclaration && !(isJsxOpeningLikeElement(node.parent) && isJsxConstructorLike(calledDeclaration))) {
         const sigInfo = createDefinitionFromSignatureDeclaration(typeChecker, calledDeclaration, failedAliasResolution);
+
         // For a function, if this is the original function definition, return just sigInfo.
         // If this is the original constructor definition, parent is the class.
-        let includeDeclarations: ((d: Declaration) => boolean) | undefined;
+        // Here, we filter declarations to not duplicate returned definitions.
+        let declarationFilter: (d: Declaration) => boolean = d => d !== calledDeclaration;
         if (typeChecker.getRootSymbols(symbol).some(s => symbolMatchesSignature(s, calledDeclaration))) {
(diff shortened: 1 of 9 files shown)


iisaduan:
@typescript-bot pack this

src/services/goToDefinition.ts
```diff
@@ -230,14 +231,16 @@ export function getDefinitionAtPosition(program: Program, sourceFile: SourceFile
         const sigInfo = createDefinitionFromSignatureDeclaration(typeChecker, calledDeclaration, failedAliasResolution);
         // For a function, if this is the original function definition, return just sigInfo.
         // If this is the original constructor definition, parent is the class.
+        let includeDeclarations: ((d: Declaration) => boolean) | undefined;
```
RyanCavanaugh, line 234: This is named like it's a boolean; I would suggest something like `declarationFilter`

src/services/goToDefinition.ts
```diff
@@ -584,9 +587,10 @@ function isExpandoDeclaration(node: Declaration): boolean {
     return !!containingAssignment && getAssignmentDeclarationKind(containingAssignment) === AssignmentDeclarationKind.Property;
 }
 
-function getDefinitionFromSymbol(typeChecker: TypeChecker, symbol: Symbol, node: Node, failedAliasResolution?: boolean, excludeDeclaration?: Node): DefinitionInfo[] | undefined {
-    const filteredDeclarations = filter(symbol.declarations, d => d !== excludeDeclaration);
-    const signatureDefinition = getConstructSignatureDefinition() || getCallSignatureDefinition();
+function getDefinitionFromSymbol(typeChecker: TypeChecker, symbol: Symbol, node: Node, failedAliasResolution?: boolean, excludeDeclaration?: Node, includeDeclarations?: (d: Declaration) => boolean): DefinitionInfo[] | undefined {
```
RyanCavanaugh, line 590: There's only one inbound call to this function that uses `excludeDeclaration`; let's just simplify this to take a predicate and then that call can pass `e => e !== node`

RyanCavanaugh (approved):
LGTM after a couple nits

PR #59425, opened 2024-07-25
Expand intersection reduction division strategy down to 3-member intersections

### LINKED ISSUE #58898: Unable to use `Pick` on union with data-${string} index signature
Based on https://github.com/microsoft/fluentui/issues/27302, and more specifically, this comment https://github.com/microsoft/fluentui/issues/27302#issuecomment-2158761470

[Playground Link](https://www.typescriptlang.org/play/?ts=5.6.0-dev.20240617#code/JYWwDg9gTgLgBAKjgQwM5wEoFNkGN4BmUEIcA5FDvmQFA3AB2MWUBeWcAEgCoCyAMgBFkMZAEEYMKMABGAV2boA3jThq4AbQAmI5AC44AAx2iAtABIlqKYwDmAX0MBdA3IYBrBhADuDGvbpGZlZ2ODEGXAALaABlABsIGHQsAA9mBi10HgFhUQkbeUUAGkwqGAA6cKjobP586UKsVAAeWqroqABROKwQLCYAPjglAPomFjZcDgAhBRgIBnjE5LT+zK4+IV162QUmkuw8CtnJBbbJBr2W2pP5hm7e-pghkcDxkKm4QWAANyWkuCpdLrWq5cQXXbFUpHcrnApXVqbb4-B59QbDUa4BbWOApAwABWAuHczXasQSAIAPnBbgt-uhqcj6SUyMAtGQhgBeYZsgys9mjAD0wuFcDewUmMzQHBUal5cGs0gYtgA3Ko4GBiGAAAwARgMirsarUmogOoATAabMrjRqtdqAMxWpWq9WmnUAFmdRrd9oArN6bb6zdqAGyB10m+0Adgjtvd2oAHHHgzqAJwpqNm3XazN27P6hXWyP5sC6y1Fl3xrW6p2Vn1Zste+tBxu6gMtkvu3XhzvV7Oxvup3XJodtjNj0vm3OT93mwuG1tTiuLrta8111f9sDm5tb1Pmjv7xvm3vHqeD89z0dX9cT29mh0zh9gB0L4vbh0rj+ph2bn+Ng6e4AaWDpHiB7oOmeEFag6l4-qMQQTKEYiAqsGToNM0olKC2wQo0yjqsg75VkR36kWoyD-hRKDATRyDgfR0H0fB9E3gh4rIZ80xocCmHYRsOR4fCijDOqMgkQ2cAyORUkyNRcl0XJjFycxcmsXJ7GkYh7ySnAADCvFrPxqBYDhSLCZcomynAuCSUuuCyQ5CkOUpDkqQ5akORpDlaT6mLYvAKQAJ4AF4EkSJKodSPHUvpLJshycDcko8r8mQARAA)

```ts
import * as React from 'react'

interface HTMLDataAttributes {
    [data: `data-${string}`]: unknown
}

interface AnchorSlots extends HTMLDataAttributes, React.AnchorHTMLAttributes<HTMLAnchorElement> {}

interface ButtonSlots extends HTMLDataAttributes, React.ButtonHTMLAttributes<HTMLButtonElement> {}

interface DivSlots extends HTMLDataAttributes, React.HTMLAttributes<HTMLDivElement> {}

const x: Pick<AnchorSlots | ButtonSlots | DivSlots, 'id'> = {id: 'id'}
//       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
// Expression produces a union type that is too complex to represent.
```

Simplified

```ts
interface Base {
  id: string;
  prop01: string;
  prop02: string;
  prop03: string;
  prop04: string;
  prop05: string;
  prop06: string;
  prop07: string;
  prop08: string;
  prop09: string;
  prop10: string;
  prop11: string;
  prop12: string;
  prop13: string;
  prop14: string;
  prop15: string;
  prop16: string;
  prop17: string;
  prop18: string;
  prop19: string;
  prop20: string;
  prop21: string;
  prop22: string;
  prop23: string;
  prop24: string;
  prop25: string;
  prop26: string;
  prop27: string;
  prop28: string;
  prop29: string;
  prop30: string;
  prop31: string;
  prop32: string;
  prop33: string;
  prop34: string;
  prop35: string;
  prop36: string;
  prop37: string;
}

interface A extends Base, HTMLDataAttributes {
  a1: string;
  a2: string;
  a3: string;
  a4: string;
  a5: string;
  a6: string;
  a7: string;
  a8: string;
}

interface B extends Base, HTMLDataAttributes {
  b1: string;
  b2: string;
  b3: string;
  b4: string;
  b5: string;
  b6: string;
  b7: string;
  b8: string;
}

interface C extends Base, HTMLDataAttributes {
  c1: string;
  c2: string;
  c3: string;
  c4: string;
  c5: string;
  c6: string;
  c7: string;
  c8: string;
}

const xyz: Pick<A | B | C, 'id'> = {id: 'id'}
//         ~~~~~~~~~~~~~~~~~~~~~
// Expression produces a union type that is too complex to represent.
```

The workaround I am advising is to use a different helper type: `DistPick`, which first distributes over each type in the union, and *then* applies a `Pick` if possible. ([Playground Link](https://www.typescriptlang.org/play/?ts=5.4.5#code/JYWwDg9gTgLgBAKjgQwM5wEoFNkGN4BmUEIcA5FDvmQFA3AB2MWUBeWcAEgCoCyAMgBFkMZAEEYMKMABGAV2boA3jThq4AbQAmI5AC44AAx2iAtABIlqKYwDmAX0MBdA8gYBPGvbox3YDoLA1gAKwLgA1gA83AA0cADSAHxwALyqcNxwWAAezAxa6IwELHAAqnAA-HAAFPFZuVj56OFY7hAEZZVwoRHRcUlwBgxYAG4sAJSD6cNjUADcdIzMrOxwAJJM0gyoYQDKADYQMMrpAPSnXHz86fqXAsKiEjbyinAAZOlq2HgwAHTBxDAqAA6sAYAALbAESLffC-QRYUTAfZYLQ8AQAiBAmFUP5iBi4cHQdH8J7SF5YVCREn4wnQACiKJAjRgiTiNIJRKgjKwzKYiUSCzULxgEAYBhJD3EknJCkp70+mFx-0BILBkKw0NhfwRSJRaKumOx2t+ACEFKKGDSZbI5VSSebJGKeXzWeyro7LS6WQKhXAtMARhKrlKybbXh91EqfiqsWqIVCcTHdchkaiSUaqSbrc87dSQ4Hvfz3fdC0yfYKvD4-BxPWKAPJQWlc1LrTaMHa4A5HVAaMgisVkJxwAA+bZs2z2h2OfeQQ9H462ne7M7IAZGQ7ouDF1jg2QM2AAjnJgJQtJFAiEwlE6wxG83oHEyMAtGQBa2lC+DM-X-YgA)):

```ts
type DistPick<T, K> =
  T extends infer _ ? (K extends keyof T ? Pick<T, K> : never) :
  never;
```

or even this (*maybe* less efficient?) version:

```ts
type DistPick<T, K> =
  T extends infer _ ? Pick<T, K & keyof T> :
  never;
```

weswigham:
So instead of 50^3 tripping a breaker, we instead at least try the 50^2 intersection first and see if it reduces any (and it does). The same error can still occur this way, but the objects need to have 1000 members instead of 50 now - a much more generous limit.

I played around with some other heuristics to find and simplify intersections of unions like this earlier (almost-primitive unions), but they ended up being overly costly for how narrow they were - pattern literals throw a wrench into our usual intersection fast path for literal types, and we *could* preserve it in cases like this one where there's a single pattern literal type common to all otherwise-primitive unions in the intersection (intersections of the pattern with other literals either produce never or that literal), but that all goes away once there are two or more pattern literals, since then you need to worry about myriad ways the patterns can combine (which is to say: potentially as many ways as the power set indicates), so it didn't seem worth it for the complexity. Even `react` has *two* of these pattern index signatures (aria- and data-), not one, so such a heuristic won't help the common case anyway. I think there is still *some* splitting and pre-simplification we can do among the literal members, probably, even when there are non-primitive members, but I haven't been able to nail down a sufficiently general algorithm yet.

Fixes #58898

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -18117,7 +18117,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     removeFromEach(typeSet, TypeFlags.Null);
                     result = getUnionType([getIntersectionType(typeSet, flags), nullType], UnionReduction.Literal, aliasSymbol, aliasTypeArguments);
                 }
-                else if (typeSet.length >= 4) {
+                else if (typeSet.length >= 3) {
                     // When we have four or more constituents, some of which are unions, we employ a "divide and conquer" strategy
                     // where A & B & C & D is processed as (A & B) & (C & D). Since intersections of unions often produce far smaller
                     // unions of intersections than the full cartesian product (due to some intersections becoming `never`), this can
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -18118,7 +18118,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     result = getUnionType([getIntersectionType(typeSet, flags), nullType], UnionReduction.Literal, aliasSymbol, aliasTypeArguments);
                 }
                 else if (typeSet.length >= 3 && types.length > 2) {
-                    // When we have four or more constituents, some of which are unions, we employ a "divide and conquer" strategy
+                    // When we have three or more constituents, more than two inputs (to head off infinite reexpansion), some of which are unions, we employ a "divide and conquer" strategy
                     // where A & B & C & D is processed as (A & B) & (C & D). Since intersections of unions often produce far smaller
                     // unions of intersections than the full cartesian product (due to some intersections becoming `never`), this can
                     // dramatically reduce the overall work.
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -18117,7 +18117,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     removeFromEach(typeSet, TypeFlags.Null);
                     result = getUnionType([getIntersectionType(typeSet, flags), nullType], UnionReduction.Literal, aliasSymbol, aliasTypeArguments);
                 }
-                else if (typeSet.length >= 3) {
+                else if (typeSet.length >= 3 && types.length > 2) {
                     // When we have four or more constituents, some of which are unions, we employ a "divide and conquer" strategy
                     // where A & B & C & D is processed as (A & B) & (C & D). Since intersections of unions often produce far smaller
                     // unions of intersections than the full cartesian product (due to some intersections becoming `never`), this can
--- a/tests/baselines/reference/pickOfLargeObjectUnionWorks.js
+++ b/tests/baselines/reference/pickOfLargeObjectUnionWorks.js
@@ -1,0 +1,85 @@
+//// [tests/cases/compiler/pickOfLargeObjectUnionWorks.ts] ////
+
+//// [pickOfLargeObjectUnionWorks.ts]
+interface HTMLDataAttributes {
+    [data: `data-${string}`]: unknown
+}
+
+interface Base {
+    id: string;
+    prop01: string;
+    prop02: string;
+    prop03: string;
+    prop04: string;
+    prop05: string;
+    prop06: string;
+    prop07: string;
+    prop08: string;
+    prop09: string;
+    prop10: string;
+    prop11: string;
+    prop12: string;
+    prop13: string;
+    prop14: string;
+    prop15: string;
+    prop16: string;
+    prop17: string;
+    prop18: string;
+    prop19: string;
+    prop20: string;
+    prop21: string;
+    prop22: string;
+    prop23: string;
+    prop24: string;
+    prop25: string;
+    prop26: string;
+    prop27: string;
+    prop28: string;
+    prop29: string;
+    prop30: string;
+    prop31: string;
+    prop32: string;
+    prop33: string;
+    prop34: string;
+    prop35: string;
+    prop36: string;
+    prop37: string;
+  }
+  
+  interface A extends Base, HTMLDataAttributes {
+    a1: string;
+    a2: string;
+    a3: string;
+    a4: string;
+    a5: string;
+    a6: string;
+    a7: string;
+    a8: string;
+  }
+  
+  interface B extends Base, HTMLDataAttributes {
+    b1: string;
+    b2: string;
+    b3: string;
+    b4: string;
+    b5: string;
+    b6: string;
+    b7: string;
+    b8: string;
+  }
+  
+  interface C extends Base, HTMLDataAttributes {
+    c1: string;
+    c2: string;
+    c3: string;
+    c4: string;
+    c5: string;
+    c6: string;
+    c7: string;
+    c8: string;
+  }
+  
+  const xyz: Pick<A | B | C, 'id'> = {id: 'id'}
+
+//// [pickOfLargeObjectUnionWorks.js]
+var xyz = { id: 'id' };
--- a/tests/baselines/reference/pickOfLargeObjectUnionWorks.symbols
+++ b/tests/baselines/reference/pickOfLargeObjectUnionWorks.symbols
@@ -1,0 +1,226 @@
+//// [tests/cases/compiler/pickOfLargeObjectUnionWorks.ts] ////
+
+=== pickOfLargeObjectUnionWorks.ts ===
+interface HTMLDataAttributes {
+>HTMLDataAttributes : Symbol(HTMLDataAttributes, Decl(pickOfLargeObjectUnionWorks.ts, 0, 0))
+
+    [data: `data-${string}`]: unknown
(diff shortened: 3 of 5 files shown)


weswigham:
@typescript-bot test this

src/compiler/checker.ts
```diff
@@ -18117,7 +18117,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     removeFromEach(typeSet, TypeFlags.Null);
                     result = getUnionType([getIntersectionType(typeSet, flags), nullType], UnionReduction.Literal, aliasSymbol, aliasTypeArguments);
                 }
-                else if (typeSet.length >= 4) {
+                else if (typeSet.length >= 3 && types.length > 2) {
                     // When we have four or more constituents, some of which are unions, we employ a "divide and conquer" strategy
```
jsep, line 18121: Should this comment be updated to accurately describe the new conditions (typeSet.length >= 3 and types.length > 2)?

tests/baselines/reference/pickOfLargeObjectUnionWorks.js
```diff
@@ -0,0 +1,85 @@
...
+    c2: string;
+    c3: string;
+    c4: string;
+    c5: string;
+    c6: string;
+    c7: string;
+    c8: string;
+  }
+  
+  const xyz: Pick<A | B | C, 'id'> = {id: 'id'}
```
sandersn, line 82: the heuristic doesn't apply for `A | B`, right?

src/compiler/checker.ts
```diff
@@ -18117,8 +18117,8 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     removeFromEach(typeSet, TypeFlags.Null);
                     result = getUnionType([getIntersectionType(typeSet, flags), nullType], UnionReduction.Literal, aliasSymbol, aliasTypeArguments);
                 }
-                else if (typeSet.length >= 4) {
-                    // When we have four or more constituents, some of which are unions, we employ a "divide and conquer" strategy
+                else if (typeSet.length >= 3 && types.length > 2) {
+                    // When we have three or more constituents, more than two inputs (to head off infinite reexpansion), some of which are unions, we employ a "divide and conquer" strategy
```
sandersn, line 18121: can you give an example of the infinite re-expansion?
Does it occur because `typeSet.slice(middle)` might recur right back to this case, repeatedly when `typeSet.slice(middle).length === 1`?

sandersn (commented):
Couple of questions to see if I understand what's going on at all.

src/compiler/checker.ts
weswigham: > Does it occur because typeSet.slice(middle) might recur right back to this case, repeatedly when typeSet.slice(middle).length === 1?

Yep. If the input is one type, but that one type expands into multiple typeSet members when inspected (because it is itself an intersection containing a union), we can end up back here as we ping-pong between trying to distribute intersections to within unions while also maximally breaking down these intersection powerset to pairwise combinations. Just bailing on this optimization if we only got a single type as input is a very low-overhead way to short-circuit that loop, since our splitting will (eventually) split down to, at the worst, a pair of types on one side plus a single type on the other.

tests/baselines/reference/pickOfLargeObjectUnionWorks.js
weswigham: So this heuristic is for _intersections_ - the way it applies here is to the key types within the mapped type. Before normalization, we construct `keyof A & keyof B & keyof C` which are all unions of many things - and this lets us construct it as `(keyof A & keyof B) & keyof C` instead, so when we see that `keyof A & keyof B` just becomes `keyof Base` (essentially), we know we can do `keyof Base & keyof C` without exponential blowup exceeding our limits.

Definitely doesn't apply to two element intersections, though - there's no set of comparisons to divide up.

A notable thing to say is that we attempt to do literal reduction of the intersection *before* the pairwise breakdown, since we can do literal-only intersection combinations in linear time (we just look for literals common to every intersected union, since them intersected with one another is always `never`), so this more complete division is only critical for `keyof` types that contain pattern literal index signatures, like in this test.

sandersn (approved):
Seems like it's worth a try.

(The root problem of *very* popular libraries using using *very* complicated types needs to be addressed too; someday, somehow.)

Issue #59422, opened 2024-07-25
Debug Failure. Expected 55 <= 54


Roise-yue:
### 🔎 Search Terms

Debug Failure,Expected 55 <= 54

### 🕗 Version & Regression Information

- This is a crash in version 5.5.4

### 💻 Code

```ts
function identity<T>(arg: T: T {
    return arg;
}
```

### 🙁 Actual behavior

```ts
/usr/lib/node_modules/typescript/lib/tsc.js:120196
      throw e;
      ^

Error: Debug Failure. Expected 55 <= 54
    at assertDiagnosticLocation (/usr/lib/node_modules/typescript/lib/tsc.js:13507:9)
    at createFileDiagnostic (/usr/lib/node_modules/typescript/lib/tsc.js:17266:3)
    at checkPotentialUncheckedRenamedBindingElementsInTypes (/usr/lib/node_modules/typescript/lib/tsc.js:80540:13)
    at /usr/lib/node_modules/typescript/lib/tsc.js:84135:11
    at addLazyDiagnostic (/usr/lib/node_modules/typescript/lib/tsc.js:84177:33)
    at checkSourceFileWorker (/usr/lib/node_modules/typescript/lib/tsc.js:84126:7)
    at checkSourceFile (/usr/lib/node_modules/typescript/lib/tsc.js:84087:5)
    at checkSourceFileWithEagerDiagnostics (/usr/lib/node_modules/typescript/lib/tsc.js:84178:5)
    at forEach (/usr/lib/node_modules/typescript/lib/tsc.js:32:22)
    at getDiagnosticsWorker (/usr/lib/node_modules/typescript/lib/tsc.js:84197:5)
```

### 🙂 Expected behavior

TSC should report a syntax error instead of crashing

### Additional information about the issue

tsconfig.json
```json
{
  "compilerOptions": {
    "target": "ESNext",
    "module": "CommonJS",
    "moduleResolution": "Node",
    "esModuleInterop": true,
    "skipLibCheck": true,
    "forceConsistentCasingInFileNames": true,
    "lib": ["es2023","dom"],
    "noEmitOnError":true
  },
  "include": [
    "TScorpus/*"
  ],
  "exclude": [
    "node_modules"
  ]
}
```

jakebailey:
> This is a crash in version 5.5.4

I know you've been posting a bunch of crashes, but can you please use one of the options that the template asks for?

- ~This is a crash~ (we should delete this option, honestly)
[...]

Issue #59353, opened 2024-07-19
Debug Failure caused by missing right parenthesis


Roise-yue:
### 🔎 Search Terms

Debug Failure，Paren

### 🕗 Version & Regression Information

- This is a crash
- This is the behavior in every version I tried, from version 5.4.5 to version 5.5.3

### 💻 Code

```ts
export default function getThing( { return 'thing'; }
```

### 🙁 Actual behavior

```ts
/usr/lib/node_modules/typescript/lib/tsc.js:120127
      throw e;
      ^

Error: Debug Failure. Expected 54 <= 53
    at assertDiagnosticLocation (/usr/lib/node_modules/typescript/lib/tsc.js:13499:9)
    at createFileDiagnostic (/usr/lib/node_modules/typescript/lib/tsc.js:17258:3)
    at checkPotentialUncheckedRenamedBindingElementsInTypes (/usr/lib/node_modules/typescript/lib/tsc.js:80475:13)
    at /usr/lib/node_modules/typescript/lib/tsc.js:84070:11
    at addLazyDiagnostic (/usr/lib/node_modules/typescript/lib/tsc.js:84112:33)
    at checkSourceFileWorker (/usr/lib/node_modules/typescript/lib/tsc.js:84061:7)
    at checkSourceFile (/usr/lib/node_modules/typescript/lib/tsc.js:84022:5)
    at checkSourceFileWithEagerDiagnostics (/usr/lib/node_modules/typescript/lib/tsc.js:84113:5)
    at forEach (/usr/lib/node_modules/typescript/lib/tsc.js:32:22)
    at getDiagnosticsWorker (/usr/lib/node_modules/typescript/lib/tsc.js:84132:5)

Node.js v20.14.0
```

### 🙂 Expected behavior

TSC report syntax error（like: error TS1005: ')' expected.）

### Additional information about the issue

tsconfig.json
```json
{
  "compilerOptions": {
    "target": "ESNext",
    "module": "CommonJS",
    "moduleResolution": "Node",
    "esModuleInterop": true,
    "skipLibCheck": true,
    "forceConsistentCasingInFileNames": true,
    "lib": ["es2023","dom"]
  },
  "include": [
    "TScorpus/*"
  ],
  "exclude": [
    "node_modules"
  ]
}
```

Roise-yue:
When I updated the typescript version to 5.5.4, this crash issue disappeared. But I'm not entirely sure how this issue was fixed.

PR #59428, opened 2024-07-26
Fixed crash related to creating file diagnostics outside of the source file range in `checkPotentialUncheckedRenamedBindingElementsInTypes`

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/59422
fixes https://github.com/microsoft/TypeScript/issues/59353
fixes https://github.com/microsoft/TypeScript/issues/59449


tests/cases/compiler/parametersSyntaxErrorNoCrash1.ts
```diff
@@ -0,0 +1,6 @@
+// @strict: true
+// @noEmit: true
+
+function identity<T>(arg: T: T {
+    return arg;
+}
```
Andarist, line 6: It's pretty important that we don't have any extra characters here (no newline). For that reason, it might be better to move this test somewhere else where I'd have improved control over source text. I'm not sure where such tests are kept - I could use a hint :)

src/compiler/checker.ts
```diff
@@ -43532,7 +43532,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     // entire parameter does not have type annotation, suggest adding an annotation
                     addRelatedInfo(
                         diagnostic,
-                        createFileDiagnostic(getSourceFileOfNode(wrappingDeclaration), wrappingDeclaration.end, 1, Diagnostics.We_can_only_write_a_type_for_0_by_adding_a_type_for_the_entire_parameter_here, declarationNameToString(node.propertyName)),
+                        createFileDiagnostic(getSourceFileOfNode(wrappingDeclaration), wrappingDeclaration.end, 0, Diagnostics.We_can_only_write_a_type_for_0_by_adding_a_type_for_the_entire_parameter_here, declarationNameToString(node.propertyName)),
```
Andarist, line 43535: There are other file diagnostics with `length: 0` so this isn't something new. I'm not sure what exact impact this might have on IDEs though 🤷‍♂️ If necessary I could detect a situation when this would exceed the source file's length and adjust the start position and the length but that would produce a worse location for this (edge) case. So it feels like this is actually better right now.

Alternatively, this could be node diagnostic with an adjusted message or smth? But then the start location of this diagnostic would still be way earlier than the location at which this error suggests that a type annotation should be inserted.

There are also other places that might be affected by similar bugs and they could be investigated. Especially some calls to `grammarErrorAtPos` might suffer from this. That's a separate issue though - I don't have any test cases for them at hand

tests/cases/compiler/parametersSyntaxErrorNoCrash1.ts
jakebailey: Leaving this is fine IMO; we don't format tests. The only hazard is if someone has the editor flag enabled to add one (or uses one of the many editors that adds it, correctly tbh). Could leave a note, though.

Alternatively, you could write it as a unit test; it wouldn't be a bad idea to make a new file for these kinds of crashes in the unittest folder.

src/compiler/checker.ts
jakebailey: Length 1 and length 0 in my experience show up the same in editors; as a single character squiggle. So, it should be okay.

This makes me wonder what other length 1s we have that are problematic.

jakebailey (approved):
Checking other calls to createFileDiagnostic in the checker, the other ones that hardcode a value all do it with length 0, so this seems fine to me. I definitely fixed a bug like this before though I can't remember which PR that was anymore.

PR #59433, opened 2024-07-26
Add option --stopBuildOnErrors to tsbuild to get previous behavior of not building downstream projects if upstream has errors

sheetalkamat:
#58854 removed the "error" as condition to block the downstream project build on errors.
Adds option `--noDownstreamOnError` to block the downstream project build on errors in upstream projects


jakebailey:
I was going to say "why not just noEmitOnError", but I guess it makes some sense to be able to emit all files no matter the type checking errors. But, I guess if `noDownstreamOnError` is enabled, you won't get the emit anyway... Not sure how I feel about that... You sort of want to emit everything but then only show the base errors?

I'm not sure what the right answer is here.

DanielRosenwasser:
I think "downstream" might be hard for developers to build an intuition around. I'd just call it something like

* `haltOnProjectErrors`
* `stopOnProjectErrors`
* `haltBuildOnErrors`
* `stopBuildOnErrors`

or something similar.

jakebailey:
The cut is early next week; what are we doing here? If I had to choose, I'd probably say `--stopBuildOnErrors`.

sheetalkamat:
But it’s not halt on build error .. it’s halt building dependencies .. so may be there are other projects that get built that are not in the tree ..

jakebailey:
That's true, yeah.

DanielRosenwasser:
I don't think the distinction is so important. `--stopBuildOnErrors` implies stopping *something* about `--build` as soon as an error is encountered.

DanielRosenwasser:
Also, to be pedantic, the cut is today. RC goes out on Tuesday.

jakebailey (approved):
LGTM code wise; if `stopBuildOnErrors` is the name we're doing with, the PR title should be updated just to ensure we don't confuse ourselves later.

Issue #59345, opened 2024-07-18
Debugging failure caused by lack of backtick（`）


Roise-yue:
### 🔎 Search Terms

backtick, Debug Failure 

### 🕗 Version & Regression Information

- This is a crash
- This happened in version 5.5.3 and version 5.4.5

### 💻 Code

This program was obtained by mutating a seed program. 
The specific error location is at the end of the first for statement under the parseButton function, which is the following statement:

```ts
this.classFormat(
                    `.cxd-Button--${fontType}:not(:disabled):not(.is-disabled):hover:active`,
                    `${style('active')} // here ,missing a quotation mark(`)
                );
```
the whole program is:
```ts
import type {PlainObject, ThemeDefinition} from './declares';

export class ParseThemeData {
    style: string[] = [];
    class: string[] = [];
    data: ThemeDefinition;
    scope: string[];
    theme: string;
    constructor(data: ThemeDefinition, scope: string[]) {
        this.data = data;
        this.scope = scope;
        this.theme = data.config.key;
    }

    generator() {
        const {global, component} = this.data;
        const {colors, fonts, borders, sizes, shadows} = global;
        this.parseColor(colors);
        this.parseFont(fonts);
        this.parseGlobalCommon(borders);
        this.parseGlobalCommon(sizes);
        this.parseSizesBase(sizes);
        this.parseShadows(shadows);
        for (let key in component) {
            if (key === 'button1') {
                this.parseButton(component['button1']);
            } else if (key === 'inputRating') {
                this.parseInputRating(component['inputRating']);
            } else {
                this.parseComponentCommon(component[key]);
            }
        }
    }

    getStyle() {
        return this.getCssVariable() + this.getCustomClass();
    }

    getCssVariable() {
        return `${this.scope.join(', ')}{${this.style.join(';')};}\n`;
    }

    getCustomClass() {
        return `${this.class.join('\n')}`;
    }

    /**
     * ????????????less?scss?????
     */
    getCustomStyle() {
        const customStyle = this.data?.customStyle?.style || '';
        return customStyle;
    }

    /**
     * ??css??
     */
    cssFormat(key: string, value: string) {
        if (!value) {
            return;
        }
        this.style.push(`${key}: ${value}`);
    }

    /**
     * ??class
     */
    classFormat(classname: string, value: string) {
        // ??????????????
        this.class.push(`${classname}{${value}}`);
    }

    /**
     * ??????
     */
    parseColor(colors: any) {
[...]

PR #59434, opened 2024-07-26
Fixed crashed related to emptied labeled statements in converted loop bodies

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/59345

### DIFF
--- a/src/compiler/transformers/es2015.ts
+++ b/src/compiler/transformers/es2015.ts
@@ -2731,11 +2731,12 @@ export function transformES2015(context: TransformationContext): (x: SourceFile 
 
     function visitVariableStatement(node: VariableStatement): Statement {
         const ancestorFacts = enterSubtree(HierarchyFacts.None, hasSyntacticModifier(node, ModifierFlags.Export) ? HierarchyFacts.ExportedVariableStatement : HierarchyFacts.None);
-        let updated: Statement = node;
+        let updated: Statement;
         if (convertedLoopState && (node.declarationList.flags & NodeFlags.BlockScoped) === 0 && !isVariableStatementOfTypeScriptClassWrapper(node)) {
+            const declarations = node.declarationList.declarations;
             // we are inside a converted loop - hoist variable declarations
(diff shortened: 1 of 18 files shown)


src/compiler/transformers/es2015.ts
```diff
@@ -2758,10 +2758,6 @@ export function transformES2015(context: TransformationContext): (x: SourceFile
             if (assignments) {
                 updated = setTextRange(factory.createExpressionStatement(factory.inlineExpressions(assignments)), node);
             }
-            else {
-                // none of declarations has initializer - the entire variable statement can be deleted
-                updated = undefined;
```
Andarist, line 2763: alternatively, emptied labeled statements could be removed by [this function](https://github.com/microsoft/TypeScript/blob/451388cfd5b0d086c5f69b9ff5675b650feef623/src/compiler/transformers/es2015.ts#L2942-L2950). This would perhaps make sense for `labeledStatementDeclarationListInLoopNoCrash1` and `labeledStatementDeclarationListInLoopNoCrash2`. When it comes to the other ones... they are so malformed that I'm not so sure. I think that TS responsibility isn't necessarily to minify the output and preserving nodes structure in cases like this is just safer

tests/baselines/reference/capturedLetConstInLoop3.js
```diff
@@ -273,6 +273,7 @@ function foo2(x) {
 function foo3(x) {
     var _loop_5 = function () {
         var x_5;
+        var v;
```
jakebailey, line 276: The original code was:

```
function foo3(x) {
    do {
        let x;
        var v;
        (function() { return x + v });
        (() => x + v);
    } while (1 === 1);
    
    use(v);
}
```

But this just declared an all new `v`?

Andarist: shoot, I missed that those were hoisted to the outer scope already here so this created a bug. I added an extra test case for that bug and fixed this.

jakebailey (approved):

PR #59415, opened 2024-07-24
Fix inconsistent generic indexed access

ahejlsberg:
Fixes the issue reported [here](https://github.com/microsoft/TypeScript/pull/59059#issuecomment-2230097569). Rationale for fix in comment [here](https://github.com/microsoft/TypeScript/pull/59059#issuecomment-2243269330).

PR #59437, opened 2024-07-26
Fix generic nullable checks

### LINKED ISSUE #50603: No error on unconstrained type parameter in `>` comparison

ahejlsberg:
This PR builds on #59415 to implement the following in aggregate:

* In an indexed access expression `obj[key]`, when `obj` and `key` both have generic types we no longer require the constraint of `obj` to be a non-nullable type. The rationale for this change is that when `key` is a valid property name for `obj`, it must be because `obj` is of a non-nullable type, as discussed [here](https://github.com/microsoft/TypeScript/pull/59059#issuecomment-2243269330).
* We now consider unconstrained generic types and generic types constrained to `unknown` to possibly be `null` or `undefined` in `--strictNullChecks` mode. Previously, we would only consider generic types with constraints including `null` or `undefined` to possibly be nullable.

Fixes #50603.

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -1226,8 +1226,11 @@ export const enum TypeFacts {
     Falsy = 1 << 23,              // !x
     IsUndefined = 1 << 24,        // Contains undefined or intersection with undefined
     IsNull = 1 << 25,             // Contains null or intersection with null
+    IsPossiblyUndefined = 1 << 26, // Possibly undefined
+    IsPossiblyNull = 1 << 27,     // Possibly null
     IsUndefinedOrNull = IsUndefined | IsNull,
-    All = (1 << 27) - 1,
+    IsPossiblyUndefinedOrNull = IsPossiblyUndefined | IsPossiblyNull,
+    All = (1 << 29) - 1,
     // The following members encode facts about particular kinds of types for use in the getTypeFacts function.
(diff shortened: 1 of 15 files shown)


ahejlsberg:
@typescript-bot test it

ahejlsberg:
The new errors in the top 400 repos tests are all the same: Functions that compare values of unconstrained generic types. Our existing behavior is to permit such comparisons--we only error when the types have constraints that explicitly include `null` or `undefined` (which obviously is inconsistent).

We should decide if we think erroring on operands that are possibly `null` or `undefined` is the right thing to do since it basically amounts to a breaking change. The semantics of comparisons involving `null` and `undefined` are well defined (they act like `NaN`) and don't cause exceptions, yet it's probably rare for such comparisons to be intended.

ahejlsberg:
We're going to wait for 5.7 on this one since it is technically a breaking change.

jakebailey:
5.7 happened, and we're now at 5.8... should we do this? 😄

PR #59438, opened 2024-07-26
Move ambient const enum error from use site to import in verbatimModuleSyntax

### LINKED ISSUE #48040: verbatimModuleSyntax + import of ambient const enum missing error
# Bug Report

[`preserveValueImports`](https://www.typescriptlang.org/tsconfig#preserveValueImports) doesn't error when you import an ambient const enum with [`isolatedModules`](https://www.typescriptlang.org/tsconfig#isolatedModules) *on*. Instead you get an error at runtime.

### :adhesive_bandage: Proposed Solution

If the named import were a type vs. an ambient const enum, you'd get a compile-time error instead.

https://github.com/microsoft/TypeScript/pull/47817 raises that error on ambient const enums, by changing its predicate to include ambient const enums.

### :mag: Search Terms

`preserveValueImports` ambient const enum

### :clock8: Version & Regression Information

4.5.0-dev.20210909 - 4.7.0-dev.20220225

### :play_or_pause_button: Playground Link

[Workbench Repro](https://www.typescriptlang.org/dev/bug-workbench/?preserveValueImports=true&isolatedModules=true#code/PTAEAEAcCcFMGdbQG6wGoEMA2BXWBJAW0gHtoAXeAKBAgEt4SsNzYATAWRLZywSrrEy5UAG9QAJRI4AdmzoyA5lzaxQAX1AAzaCUKgARACM6igHQAreAYDcVGmHBa6fGRkKwAXKBndYAfUJuXgRgE3MrYAVVAA8zNjNKKlgY0gpQAGMSGXgRWBkcfSlZeSUVNVF1IA)

### :technologist: Code

```ts repro
// @preserveValueImports
// @isolatedModules
import { RoundingMode } from "big.js";

// @filename: node_modules/big.js/index.d.ts
export const enum RoundingMode {}
```

### :slightly_frowning_face: Actual Behavior

Runtime error:
```
import { RoundingMode } from "big.js";
         ^^^^^^^^^^^^
SyntaxError: Named export 'RoundingMode' not found. The requested module 'big.js' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'big.js';
const { RoundingMode } = pkg;
```

### :slightly_smiling_face: Expected Behavior

Compile-time error (with https://github.com/microsoft/TypeScript/pull/47817):
```
index.ts:1:10 - error TS1446: 'RoundingMode' resolves to a type-only declaration and must be imported using a type-only import when 'preserveValueImports' and 'isolatedModules' are both enabled.

1 import { RoundingMode } from "big.js";
           ~~~~~~~~~~~~
```

andrewbranch:
Fixes #48040

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -39520,7 +39520,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             error(node, Diagnostics.const_enums_can_only_be_used_in_property_or_index_access_expressions_or_the_right_hand_side_of_an_import_declaration_or_export_assignment_or_type_query);
         }
 
-        if (getIsolatedModules(compilerOptions)) {
+        // Do not check for --verbatimModuleSyntax here because imports of ambient const enums
+        // will already be checked for the same error
+        if (compilerOptions.isolatedModules) {
             Debug.assert(!!(type.symbol.flags & SymbolFlags.ConstEnum));
             const constEnumDeclaration = type.symbol.valueDeclaration as EnumDeclaration;
             const redirect = host.getRedirectReferenceForResolutionFromSourceOfProject(getSourceFileOfNode(constEnumDeclaration).resolvedPath);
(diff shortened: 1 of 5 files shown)


src/compiler/checker.ts
```diff
@@ -39520,7 +39520,12 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             error(node, Diagnostics.const_enums_can_only_be_used_in_property_or_index_access_expressions_or_the_right_hand_side_of_an_import_declaration_or_export_assignment_or_type_query);
         }
 
-        if (getIsolatedModules(compilerOptions)) {
+        // Do not check for --verbatimModuleSyntax here because imports of ambient const enums
```
jakebailey, line 39523: I don't understand this comment; isn't `getIsolatedModules` just `compilerOptions.isolatedModules || compilerOptions.verbatimModuleSyntax`?

andrewbranch: the `verbatimModuleSyntax` condition is `&&`'d with “resolves to a global”; dprint would not let me group them together

jakebailey:
```
compilerOptions.isolatedModules
 || compilerOptions.verbatimModuleSyntax && ok && !(resolveName(node, getFirstIdentifier(node as EntityNameOrEntityNameExp ...
```

Seems to work.

jakebailey (approved):

gabritto (approved):

PR #59440, opened 2024-07-27
Refactored node builder flags and tests

armanio123:
The `NodeBuilderFlags` enum has shifted more than the allowed signed int32 limit.

~~This refactors experiments with creating a type and constant using `BigInt` instead.~~ This approach was discarded becuase of the performance of BigInt, and the limitations the type has.

Added an additional flag `InternalNodeBuilderFlags`, that releases a couple of bits on `NodeBuilderFlag` by moving all the internal flags.

Reference to this PR https://github.com/microsoft/TypeScript/pull/58475.


armanio123:
@typescript-bot test this

jakebailey:
#59282 gives back a node flag, can we not just wait for that?

src/compiler/types.ts
```diff
@@ -5427,52 +5427,60 @@ export const enum ContextFlags {
     SkipBindingPatterns = 1 << 3, // Ignore contextual types applied by binding patterns
 }
 
+export type NodeBuilderFlags = bigint;
+
 // NOTE: If modifying this enum, must modify `TypeFormatFlags` too!
 // dprint-ignore
-export const enum NodeBuilderFlags {
-    None                                    = 0,
+export const NodeBuilderFlags = {
```
jakebailey, line 5434: This is most definitely a breaking change; it will cause anyone using the reverse mapping to break (including our own debugging helpers), and also not be inlinable by our build. Not sure I like this approach...

armanio123: I agree, this is a breaking change, but I feel like any refactor on NodeBuilderFlags is going to be a breaking change.

Could you point me to the debugging helpers? I want to see how is the enum used.

The other approaches I'm considering for refactor are:
- Divide the flags into multiple enums/Add a new enum. (i.e. `NodeBuilderOptionFlags`, `NodeBuilderErrorFlags`, `NodeBuilderStateFlags`). This keeps the reverse mapping but requires more changes on the typescript repo. Plus, I can sense there's going to be some issues when combining flags from different enums.
- Replace the enum with an options interface. Using `&` and `|` would be gone but instead we can define the options. It will also have the same limitation as using BigInt:
 ```js
const NodeBuilderFlags = {
    NoTruncation: true,
    SuppressAnyReturnType: true
    ...
}
```

Do you have any suggestion of preferred approach?

**Comment by armanio123** [comment]
> #59282 gives back a node flag, can we not just wait for that?

I was planning on adding a new flag to it, that's how I noticed this issue. So even if we reduce the number of flags, we're going to be dealing with this limitation.

File: src/compiler/types.ts (hunk 1 above)
Comment (jakebailey) [reply src/compiler/types.ts:5434]: We could have two enums, then convert existing APIs that take those flags to instead take an object with the two enums as props (leaving old callers still valid with the old values, while new callers who want new flags have to pass the object in).

Comment (jakebailey) [reply src/compiler/types.ts:5434]: e.g. more or less like the createSourceFile options.

Comment (armanio123) [reply src/compiler/types.ts]: Changed as suggested. I decided to use the `flags` on the new interface and move the `AllowUnresolvedNames` as a property instead. The idea being using properties as options instead of starting to populate another `enum`.

Let me know if you prefer this change.

File: src/compiler/transformers/declarations.ts [hunk 2]
```diff
@@ -1471,7 +1474,7 @@ export function transformDeclarations(context: TransformationContext) {
                             return undefined; // unique symbol or non-identifier name - omit, since there's no syntax that can preserve it
                         }
                         getSymbolAccessibilityDiagnostic = createGetSymbolAccessibilityDiagnosticForNode(p.valueDeclaration);
-                        const type = resolver.createTypeOfDeclaration(p.valueDeclaration, fakespace, declarationEmitNodeBuilderFlags | NodeBuilderFlags.NoSyntacticPrinter, symbolTracker);
+                        const type = resolver.createTypeOfDeclaration(p.valueDeclaration, fakespace, declarationEmitNodeBuilderOptions.flags | NodeBuilderFlags.NoSyntacticPrinter, symbolTracker);
```
jakebailey, line 1477: This is the sort of thing where this approach breaks without some helpers; this call site, for example, loses the other node builder options because it pulls off just the flags.

armanio123: Yeah, that's definitely a bug I have missed. This could be easily fixed by doing `{ ...declarationEmitNodeBuilderOptions, flags: declarationEmitNodeBuilderOptions.flags | NodeBuilderFlags.NoSyntacticPrinter }`

armanio123:
Refactored as suggested during the design meeting. There's a couple of tests that looks suspicious to me so I appreciate extra set of eyes on them.

jakebailey:
Yeah, I suspect there's a bug lurking in there; most of the baseline changes are unexpected.

src/compiler/checker.ts
```diff
@@ -6192,15 +6195,27 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             return context.encounteredError ? undefined : resultingNode;
         }
 
+        function saveFlags(context: NodeBuilderContext) {
+            const flags = context.flags;
+            const internalFlags = context.internalFlags;
+
+            return { restore };
+
+            function restore() {
```
jakebailey, line 6204: This is a good idea, but could probably just return the restore function itself to avoid the object.

jakebailey:
Skimming the code locally, though, I can't find the place where things may have gone wrong... It all seems correct to me.

src/compiler/types.ts
```diff
@@ -5459,12 +5459,9 @@ export const enum NodeBuilderFlags {
     AllowEmptyTuple                         = 1 << 19,
     AllowUniqueESSymbolType                 = 1 << 20,
     AllowEmptyIndexInfoType                 = 1 << 21,
-    /** @internal */ WriteComputedProps      = 1 << 30, // { [E.A]: 1 }
-    /** @internal */ NoSyntacticPrinter     = 1 << 31,
     // Errors (cont.)
     AllowNodeModulesRelativePaths           = 1 << 26,
-    /** @internal */ DoNotIncludeSymbolChain = 1 << 27,    // Skip looking up and printing an accessible symbol chain
-    /** @internal */ AllowUnresolvedNames = 1 << 32,
```
jakebailey, line 5467: Given `AllowUnresolvedNames` and `NoTruncation` are the same, I think we need to go vet the places we passed in `NoTruncaction` before to see if they require `AllowUnresolvedNames`. That may be the source of the types baseline changes.

tests/cases/fourslash/completionsClassMemberImportTypeNodeParameter2.ts
```diff
@@ -18,7 +18,7 @@ verify.completions({
   includes: [
     {
       name: "method",
-      insertText: `method(param: import("./doesntexist.js").Foo);`,
+      insertText: `method(param: any);`,
```
jakebailey, line 21: This seems like a case where we would have allowed unresolved names before, so maybe we also need to vet everyone who passed NoTruncation to see if they need AllowUnresolvedNames?

src/compiler/checker.ts
armanio123: Done. I changed the name to saveRestoreFlags as it makes more sense to me.

src/compiler/types.ts
armanio123: I think the latest commit fixed this by just adding the `AllowUnresolvedNames` flag it two spots. No tests baselines are changed after this change.

jakebailey: Awesome!

src/services/codefixes/helpers.ts
```diff
@@ -222,7 +223,7 @@ export function addNewNodeForMemberSymbol(
         case SyntaxKind.PropertyDeclaration:
             let flags = NodeBuilderFlags.NoTruncation;
```
jakebailey, line 224: This doesn't need `AllowUnresolvedNames` like the code below?

Sort of wondering if all old users of `NoTruncation` need that flag, but maybe we shouldn't add ones where we don't think it should matter except those where we would have set it manually...

src/services/codefixes/helpers.ts
```diff
@@ -506,6 +507,7 @@ export function createSignatureDeclarationFromCallExpression(
         contextNode,
         scriptTarget,
         NodeBuilderFlags.NoTruncation,
+        /*internalFlags*/ undefined,
```
jakebailey, line 510: This makes me wonder if we are missing a test with an unresolved name for those who call this code. (Specifically fixAddMissingMember)

jakebailey (changes requested):
Overall this now seems good, though the title needs to be updated to not say it's an experiment, and I do think we may need to be adding the unresolved name thing in a couple of places that maybe we just don't have tests for?

I think that the tests that fail without the change to `createSignatureDeclarationFromSignature` may need to be recreated as tests which make the other cases where `NoTruncation` failed also fail without `AllowUnresolvedNames`.

src/services/codefixes/helpers.ts
armanio123: Currently is not needed to pass the test cases. Thought we might be missing some, because adding it doesn't change any of them.

jakebailey: Right, our test suite is just not always exhaustive, so I suspect that there are some lurking bugs without this flag for these auto-import / codefix cases, given those seem to in other cases depend on this flag

src/services/codefixes/helpers.ts
armanio123: I agree, that's the same thing I just responded to your last comment.

armanio123:
If I understand the `AllowUnresolvedNames` flag correctly this tests cases should not use `any` correct? If so, there's no reason to add the flag as we don't handle it.

```ts
interface A {
    foo: import("./doesntexist.js").Foo;
    get bar(): import("./doesntexist.js").Foo;
}

class B implements A {
| <-- Trigger completion here
}
```
Completion returns:
```ts
foo: any;           // This should have been  `foo: import("./doesntexist.js").Foo;`
get bar(): any {    // and this `get bar(): import("./doesntexist.js").Foo;`
}
```

jakebailey:
I would expect those to not spit out `any`, no, though I'm confused because it sounds like you're saying it spits out `any` when the flag _is_ set?

armanio123:
That's correct, we spit out `any` regardless of the flag. Most likely we don't handle the flag on this code path.

jakebailey:
If you don't want to write a test for those other branches, can we just put that flag back on the places that used to have `NoTruncation` to maintain the existing behavior?

armanio123:
Yeah, I was going to propose the same as I don't know how to create all of the necessary tests. I'm going to put the flag in everyplace I find `NoTruncation`.

src/compiler/checker.ts
```diff
@@ -5962,7 +5964,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
 
     function typeToString(type: Type, enclosingDeclaration?: Node, flags: TypeFormatFlags = TypeFormatFlags.AllowUniqueESSymbolType | TypeFormatFlags.UseAliasDefinedOutsideCurrentScope, writer: EmitTextWriter = createTextWriter("")): string {
         const noTruncation = compilerOptions.noErrorTruncation || flags & TypeFormatFlags.NoTruncation;
-        const typeNode = nodeBuilder.typeToTypeNode(type, enclosingDeclaration, toNodeBuilderFlags(flags) | NodeBuilderFlags.IgnoreErrors | (noTruncation ? NodeBuilderFlags.NoTruncation : 0));
+        const typeNode = nodeBuilder.typeToTypeNode(type, enclosingDeclaration, toNodeBuilderFlags(flags) | NodeBuilderFlags.IgnoreErrors | (noTruncation ? NodeBuilderFlags.NoTruncation : NodeBuilderFlags.None), noTruncation ? InternalNodeBuilderFlags.AllowUnresolvedNames : InternalNodeBuilderFlags.None);
```
armanio123, line 5967: This scenario seems weird to me as it depends on `NoTruncation` being set in `TypeFormatFlags`. Nonetheless, this is the existing behavior, so I decided to change it but let me know if you disagree and I can rollback.

jakebailey: Hm, I think it's a little weird, though I guess it's also true that nobody would have ever _intentionally_ set `AllowUnresolvedNames` here because that was never a valid `TypeFormatFlag`.

Sort of leaning towards _not_ doing this one and seeing what happens? I assume no test changes here either.

armanio123: They don't change. Actually, the latest commit doesn't change any of the tests. I'll rollback this line.

jakebailey (approved):
I think this seems alright now, though it would be nice if we could figure out the rest.

Given we just added a flag everywhere, curious what @weswigham thinks, but it should at least maintain our old behavior which I feel is pretty safe.

PR #59441, opened 2024-07-27
Fixed inlay hints for inferred type predicates

Andarist:


tests/baselines/reference/inlayHintsInferredTypePredicate1.baseline
```diff
@@ -0,0 +1,9 @@
+// === Inlay Hints ===
+function test(x: unknown) {
+                         ^
+{
+  "text": ": x is number",
+  "position": 25,
+  "kind": "Type",
```
Andarist, line 7: do I need to introduce a new `TypePredicate` kind for this?

jakebailey: Nope: https://code.visualstudio.com/api/references/vscode-api#InlayHintKind

jakebailey (approved):

src/services/inlayHints.ts
```diff
@@ -463,28 +471,33 @@ export function provideInlayHints(context: InlayHintsContext): InlayHint[] {
         return typeToInlayHintParts(signatureParamType);
     }
 
-    function printTypeInSingleLine(type: Type) {
+    function typeToInlayHintParts(type: Type): InlayHintDisplayPart[] | string {
         const flags = NodeBuilderFlags.IgnoreErrors | NodeBuilderFlags.AllowUniqueESSymbolType | NodeBuilderFlags.UseAliasDefinedOutsideCurrentScope;
-        const printer = createPrinterWithRemoveComments();
 
-        return usingSingleLineStringWriter(writer => {
```
jakebailey, line 470: Why is this bit being removed? Don't we need to ensure that these are single lines?

jakebailey (changes requested):

src/services/inlayHints.ts
Andarist: IIRC, I concluded that this didn't have anything logic inside it that would not already be covered by `typeToString`. When types are written in error diagnostics etc they are single line. I could miss something so if you think this is wrong - I could revisit this.

MariaSolOs: I think that even if the logic is already being done in `typeToString` it's better to be explicit here and keep the current code (also because we don't want to rely on `typeToString`s behavior).

Andarist: I reverted this change 👍

jakebailey (approved):

PR #59446, opened 2024-07-27
fix(59292): cover usage of isolatedDeclarations with optional parameter typed as imported generic type

### LINKED ISSUE #59292: [isolatedDeclarations] Optional parameter in arrow function typed as imported generic type gets emitted as `any`
### 🔎 Search Terms

isolated declarations transpiledeclarations any arrow 

### ⏯ Playground Link

https://www.typescriptlang.org/play/?noCheck=true&isolatedDeclarations=true#code/JYWwDg9gTgLgBAbxgTzAUzgMQhAvnAMyghDgHICcyBuAKFrQA9JY4BjCAOwGd4AjAIZQ4AXjgAeACoAaAHwAKShAD8ALiw4psgJTqAbhGAATUbMS46QA

### 💻 Code

```ts
import {type Foo} from 'foo';

export const bar = <T,>(foo?: Foo<T>): void => {};
```

### 🙁 Actual behavior

With `"isolatedDeclarations": true` this gets emitted as

```ts
export declare const bar: <T>(foo?: any) => void;
```

### 🙂 Expected behavior

This gets emitted as
```ts
import {type Foo} from 'foo';

export declare const bar: <T>(foo?: Foo<T>) => void;
```

_No response_

a-tarasyuk:
Fixes #59292

### DIFF
--- a/tests/baselines/reference/isolatedDeclarationsOptionalParameter.js
+++ b/tests/baselines/reference/isolatedDeclarationsOptionalParameter.js
@@ -1,0 +1,25 @@
+//// [tests/cases/compiler/isolatedDeclarationsOptionalParameter.ts] ////
+
+//// [package.json]
+{
+    "name": "foo",
+    "types": "index.d.ts"
+}
+
+//// [index.d.ts]
+export interface Foo<T> {
+    a: T;
+}
+
+//// [bar.ts]
+import { type Foo } from "foo";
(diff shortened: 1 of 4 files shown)


jakebailey:
I'm going to close this; #60195 was what fixed the issue, and I believe its tests are sufficient.

a-tarasyuk:
@jakebailey np. thanks

Issue #59367, opened 2024-07-19
TS2590 for simple ternary expression with large discriminated union


wctaiwan:
### 🔎 Search Terms

TS2590 ternary conditional

### 🕗 Version & Regression Information

- This changed between versions 3.3.3 and 3.5.1

### ⏯ Playground Link

[...]

PR #59448, opened 2024-07-27
Avoid union subtype reduction when union gets unionized with another type that is exclusively nullable

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/59367


src/compiler/checker.ts
```diff
@@ -17673,6 +17673,10 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         }
         // We optimize for the common case of unioning a union type with some other type (such as `undefined`).
         if (types.length === 2 && !origin && (types[0].flags & TypeFlags.Union || types[1].flags & TypeFlags.Union)) {
+            // avoid subtype reduction when one of the types has only types that can't have subtypes
+            if (unionReduction === UnionReduction.Subtype && some(types, t => everyType(t, t => !!(t.flags & TypeFlags.Nullable)))) {
```
Andarist, line 17677: I'm not sure if this is a great idea, this should be a somewhat common case though. This check could be expanded to cover more types that can't be subtypes (like symbols) but I wouldn't expect those cases to be that common so perhaps it's not worth it.

PR #39266, opened 2020-06-25
update JSON.parse argument signature

probablyup:
Testing with the Node REPL, `JSON.parse()` accepts all copy-by-value primitives and returns them directly.

PR #59458, opened 2024-07-29
Update JSON.parse signature

### LINKED ISSUE #59457: JSON.parse also works with null, not just a string
### ⚙ Compilation target

esnext

### ⚙ Library

typescript

`JSON.parse` works if first argument is null. Current type definition doesn't allow null, only a string.

### Sample Code

```TypeScript
JSON.parse(null)
```

### Documentation Link

_No response_

ic-768:
Fixes #59457


MartinJohns:
https://github.com/microsoft/TypeScript/pull/39266#issuecomment-649869084

PR #59459, opened 2024-07-29
Fix switch-true case-true narrowing to behave equivalently to default case

### LINKED ISSUE #59133: switch-true case-true narrowing not-equivalent to default case.
### 🔎 Search Terms

`"switch true"` `"switch-true"` `switch true narrowing`.  Most issues that turned up were from 2023 which I think precedes the release of the switch-true type narrowing feature?

- This is a gap

### ⏯ Playground Link

https://www.typescriptlang.org/play/?ts=5.6.0-dev.20240703&ssl=22&ssc=1&pln=12&pc=1#code/GYVwdgxgLglg9mABHA1gQwJ4AoZgA4hQBciAzlAE64DmAlIiQEQCCjAPowELuMDCjiAN4AoRGLIB3GFAgALRFkogApvRHiNiCGlLLEAejT6AdFGXkc+QrSKjN9isqggKSFowDcd+9t0GARiZmFrgEUDbe9uKOzq6IXJ6RGgAmysBoIAA2xElRMS5u-F4aAL7CZcKgkLAIiCCycLKWYSTkVGB0DPGsHNwc-EKRpFIy8ooUKmq5Yr56hkHmUM3WtlGa+XHuxVGzAQshVuGra9FOBfHc2z46ekrKxyeIG4WJpeXCQA

### 💻 Code

```ts
function uhoh(input: string) : "A"|"B"|"C" {
    switch (true) {
        case /a/.test(input):
            return "A";
        case /b/.test(input):
            return "B";
        case true: // <--- default: works fine here
            return "C";
    }
}
```

### 🙁 Actual behavior

Complains that the result of `uhoh` could be undefined.

### 🙂 Expected behavior

`switch(true) { case true: }` should behave like `switch(true) { default: }`, probably?

_No response_

Zzzen:
… case

Fixes #59133

### DIFF
--- a/tests/cases/compiler/narrowByClauseExpressionInSwitchTrue11.ts
+++ b/tests/cases/compiler/narrowByClauseExpressionInSwitchTrue11.ts
@@ -1,0 +1,24 @@
+// @strict: true
+// @noEmit: true
+
+function uhoh(input: string) : "A"|"B"|"C" {
+    switch (true) {
+        case /a/.test(input):
+            return "A";
+        case /b/.test(input):
+            return "B";
+        case true:
+            return "C";
+    }
+}
+
+function uhoh1(input: string) : "A"|"B"|"C" {
+    switch ((true)) {
+        case /a/.test(input):
+            return "A";
+        case /b/.test(input):
+            return "B";
+        case (true):
+            return "C";
+    }
+}
--- a/src/compiler/binder.ts
+++ b/src/compiler/binder.ts
@@ -1702,7 +1702,8 @@ function createBinder(): (file: SourceFile, options: CompilerOptions) => void {
         preSwitchCaseFlow = currentFlow;
         bind(node.caseBlock);
         addAntecedent(postSwitchLabel, currentFlow);
-        const hasDefault = forEach(node.caseBlock.clauses, c => c.kind === SyntaxKind.DefaultClause);
+        const hasDefault = forEach(node.caseBlock.clauses, c => c.kind === SyntaxKind.DefaultClause) ||
+            (skipParentheses(node.expression).kind === SyntaxKind.TrueKeyword && node.caseBlock.clauses.some(clause => clause.kind === SyntaxKind.CaseClause && skipParentheses(clause.expression).kind === SyntaxKind.TrueKeyword));
         // We mark a switch statement as possibly exhaustive if it has no default clause and if all
         // case clauses have unreachable end points (e.g. they all return). Note, we no longer need
         // this property in control flow analysis, it's there only for backwards compatibility.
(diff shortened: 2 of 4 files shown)


jakebailey:
@typescript-bot test it

Firsh:
Would this also fix this function's return type being inferred as `number | undefined` instead of just `number`?
```ts
type Circle = {
  kind: "circle";
  radius: number;
};

type Square = {
  kind: "square";
  sideLength: number;
};

type Shape = Circle | Square;

function calculateArea(shape: Shape) {
  switch (true) {
    case shape.kind === "circle": {
      return Math.PI * shape.radius * shape.radius;
    }
    case shape.kind === "square": {
      return shape.sideLength * shape.sideLength;
    }
  }
}
```

tests/cases/compiler/narrowByClauseExpressionInSwitchTrue11.ts
```diff
@@ -0,0 +1,24 @@
+// @strict: true
+// @noEmit: true
+
+function uhoh(input: string) : "A"|"B"|"C" {
+    switch (true) {
```
jakebailey, line 5: Awful, but, what about `switch (false)` + `case false:`?

Zzzen: To my surprise, it seems some folks are writing `switch(false)`. No `case false:` found though. Guess it wouldn't hurt to support it too! https://github.com/search?q=switch%28false%29%7B+language%3Atypescript&type=code&p=1

Zzzen: However, after some research, I don’t think we want to support it since switch(false) isn’t something we generally support.

[playground](https://www.typescriptlang.org/play/?#code/C4TwDgpgBAglC8UDeVSQFxQEQy1AvgNwBQa0AQgsquBJluXkabbAPYBOlicAPlORLEAxmwB2AZ2BQAlhLiIAFAA9MMTuQCUmZbImwEAPijKAdGQTxEOLCVGTpc7lBVqN2k3oFGT51letGIQAzAFcxYWAZcVQIKQBGV3YuTWRiKCgJAHcZYGEACxdggEMAGwkIVKR0jKhhYoq9GBVtGtqoAHoO2TEoYqhIDmCISKgszlKAEwAaT3HQqah84oA3aAtyNtrlEnaMgCMOCGKAa13a+sanFvQtjK6evoGIIZHpcY4p2d15xeW1miQWB3EzndqHY5nGr4Ygw4hhCJRGLAOLAABMSXUKTSGWyuQKLmAHFClRxFwa0DkzWUrT2oJBENOYLqFL05BuIJ2DKOTOhsKAA)


```ts
type A = { type: "A" };
type B = { type: "B" };
type AorB = A | B;

const isA = (x: AorB): x is A => x.type === "A";
const isB = (x: AorB): x is B => x.type === "B";

function test1(x: AorB) {
  switch (false) {
    case isA(x):
      // in a perfect world, x would have type B
      x;
      break;
    case isB(x):
      // in a perfect world, x would have type A
      x;
      break;
  }
}

function test2(x: AorB) {
  switch (true) {
    case isA(x):
      x;
      break;
    case isB(x):
      x;
      break;
  }
}
```

Zzzen:
> Would this also fix this function's return type being inferred as `number | undefined` instead of just `number`?

@Firsh no, you probably want to create a new issue for this.

PR #59466, opened 2024-07-29
Update deps

jakebailey:
Mainly for:

- New monocart with proper file sorting in reports
- Update `globals`; we don't support linting in older versions so this is okay and gives us proper globals for newer node.
- Removes an `npm audit`


package-lock.json
```diff
@@ -649,6 +649,21 @@
                 "concat-map": "0.0.1"
             }
         },
+        "node_modules/@eslint/eslintrc/node_modules/globals": {
+            "version": "13.24.0",
```
jakebailey, line 653: This being duplicated was one reason I avoided upgrading this, but it's not harmful and eventually we'll get ESLint v9 and this all will drop away.

RyanCavanaugh (approved):

Issue #59450, opened 2024-07-28
Conditional type evaluation not deferred when it depends on an implicitly-present type parameter


m-shaka:
### Acknowledgement

- [X] I acknowledge that issues using this template may be closed without further explanation at the maintainer's discretion.

### Comment

I'm not sure if it's a bug, but it seems like it's inconsistency of generic type.

```typescript
declare function f(fun: <T>(t: T) => void): void
f((t) => {
    type T = typeof t

    type IsAny = T extends any ? true: false // deferred
    type IsAnyArray = T[] extends any[] ? true: false // return true

    type IsString = T extends string ? true : false // deferred
    type IsStringArray = T[] extends string[] ? true : false // return false
})

function g<T>(t: T) {
    type IsAny = T extends any ? true: false // deferred
    type IsAnyArray = T[] extends any[] ? true: false // return true

    type IsString = T extends string ? true : false // deferred
    type IsStringArray = T[] extends string[] ? true : false // deferred
}
```

I expected `IsStringArray` is deferred in the function passed to `f` like the. body of `g`, but evaluated as `false`

TS version: 5.5.3
Reproduction code: [playground](https://www.typescriptlang.org/play/?ssl=18&ssc=2&pln=1&pc=1#code/CYUwxgNghgTiAEAzArgOzAFwJYHtVIAoVUAueAHgBUA+AjMygSngF5r4A3HLYRsrngChEBOszbwA3oPiz4GAJ4AHBJVbzlIHInmCZcxSvgBJAM4BBVAvVqQADwwhUwU-ChX4AfnkxkIMohQEKYIAPSh8KCIIDBwwPqyhghmlgrmsVDWLPCUANoAuvD2js6u7goFXj5+AUEh8OHwcBjIMPgYviB6chpGZgDKHVioAOY2RQ5OLvCmQ6NVHX7wtcFhEVExcQm9yaaDMMMj6TCZNpXFU66zB6OV3osIK-WNza34gauCAL6MesSYuHwIyotHoOWY0h6SRMFg82Vsk1Kbg8906TzWkRA0ViIHiUM0MNSx1O8POiOm5Tu1X8SDqGNebWp3QMBIGczG8ImJWm10OC06y1pqwa6yxm1x22hbJuRwyWRyZO5V3ZVIego+z1F2K2XyAA)

[...]

PR #59481, opened 2024-07-31
Fixed conditional type deferral when using deferred type references with type parameters from contextual signature

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/59450


ahejlsberg:
This isn't quite the right fix. See my comment [here](https://github.com/microsoft/TypeScript/issues/59450#issuecomment-2265677804).

Andarist:
superseded by https://github.com/microsoft/TypeScript/pull/59516

Issue #44442, opened 2021-06-04
Design Meeting Notes, 6/4/2021


DanielRosenwasser:
# ES Modules in Node

https://gist.github.com/weswigham/22a064ffa961d5921077132ae2f8da78

* Last time we left off talking about `.tsx` syntax and how it affected JS module support.
    * `.cjs` and `.mjs` need a corresponding input format for them.
    * `.cts` and `.mts`
    * `.jsx` -> `.tsx`
    * `.mjsx` -> `.mtsx`?
        * Can potentially get away with just putting JSX in `.mts`.
        * Is `.mjsx` a thing?
        * No, probably an assumption that `.mjs` can contain JSX.
        * JSX is a superset of JS, right?
            * Yes but not a superset of TS.
            * In `.js` files we always parse JSX anyway.
                * We were trying to be "descriptive", not "prescriptive" - not something we're in love with.
* `.cjs` and `.mjs`
    * Have to start recognizing these as inputs under `allowJs`
    * Idea: in the new Node mode, we "do the right thing" - otherwise, we just assume they should be handled as regular `.js` files.
    * No `.cjsx`, `.mjsx`.
    * Do we conditionally parse JSX on whether the JSX flag is specified?
        * No, wouldn't start now either.
        * We just complain in grammar checks if you use JSX syntax without the flag.
* What is the distinction between `.js` and `.jsx` in today's implementation?
    * None.
* `.cts` and `.mts` extensions
    * These include TypeScript syntax and correspond exactly to `.cjs` and `.mjs` respectively.
    * That way we don't have to add a flag to say "emit as `.cjs`" and "emit as `.mjs`" from `.ts` input.
* How do we avoid `.ctsx` and `.mtsx`?
    * Can just assume `.cts` includes JSX, `.mts` includes JSX.
    * Would have to get rid of
        * Angle bracket type assertion syntax (`<Type>expression`)
            * Workaround: `expression as Type`
        * Arrow functions with single type parameters (`<T>(x: T) => x`)
            * Workaround: `<T,>(x: T) => x` or `<T extends unknown>(x: T) => x`
        * Infinite lookahead?
            * Already do infinite look-ahead in other places.
            * Could try parsing an arrow function following specific JSX opening tags.
                * Then error if you can successfully parse.
                * Feels like we can do the disambiguation.
                * But the `>` in the arrow is not illegal.
                    * "Could've sworn that JSX rejected certain characters."
                    * "Both major compilers just allowed them." 😕
[...]

PR #59485, opened 2024-07-31
Fix parsing of generic type parameters in arrow functions with JSX syntax

### LINKED ISSUE #15713: Generic usage reported as JSX Error
**TypeScript Version:**  2.2.2 & 2.3.2

**Code**

```ts
// Basic example
const paramArray = <T>(param1: T, param2: T) => [param1, param2];

// React specific example
type Component<T> = React.ComponentClass<T> | React.StatelessComponent<T>;
const decorator = <T>(Component: Component<T>) : Component<T> => (props) => <Component {...props} />
```

**Expected behavior:**

Both examples should work in a `.tsx` file in addition to the basic example working in a `.ts` file.

**Actual behavior:**

The usage of `<T>` prior to the function braces causes a JSX error within `.tsx` files: "JSX element <T> has no corresponding closing tag.". Basic example works as expected in a `.ts` file.

Specifying an unused type variable seems to be a workaround to avoid the error:

```ts
const paramArray = <T1, T2>(param1: T1, param2: T1) => [param1, param2];
```

poteboy:
Potentially fixes #15713 

This PR addresses an issue where the TypeScript parser incorrectly interprets generic type parameters in arrow functions as JSX elements. Specifically, the code `const a = <T>(b: T) => b` was being parsed as a JSX element instead of a generic type parameter. 

### Testing:
Added a new test case `jsxGenericArrowFunction.tsx` to verify the fix. Before the change, the following errors were observed:

```js
jsxGenericArrowFunction.tsx(1,12): error TS2304: Cannot find name 'T'.
jsxGenericArrowFunction.tsx(1,12): error TS17008: JSX element 'T' has no corresponding closing tag.
jsxGenericArrowFunction.tsx(1,22): error TS1382: Unexpected token. Did you mean `{'>'}` or `&gt;`?
jsxGenericArrowFunction.tsx(1,25): error TS1005: '</' expected.
```

After the change, these errors no longer occur.

This change ensures that generic type parameters in arrow functions are correctly parsed, improving TypeScript's type-checking and syntax parsing capabilities.

Additionally, as noted in the [Design Meeting Notes](https://github.com/microsoft/TypeScript/issues/44442), there has been anticipation for a fix to this issue from the TypeScript community

### DIFF
--- a/src/compiler/parser.ts
+++ b/src/compiler/parser.ts
@@ -5352,6 +5352,15 @@ namespace Parser {
                     return Tristate.True;
                 }
 
+                const isNextTokenGreater = lookAhead(function () {
+                    var third = nextToken();
+                    return third === SyntaxKind.GreaterThanToken;
+                });
+
+                if (isNextTokenGreater) {
+                    return Tristate.Unknown;
+                }
+
                 return Tristate.False;
             }
 
--- a/tests/baselines/reference/jsxGenericArrowFunction.symbols
+++ b/tests/baselines/reference/jsxGenericArrowFunction.symbols
@@ -1,0 +1,10 @@
+//// [tests/cases/compiler/jsxGenericArrowFunction.tsx] ////
+
+=== jsxGenericArrowFunction.tsx ===
+const a = <T>(b: T) => b
+>a : Symbol(a, Decl(jsxGenericArrowFunction.tsx, 0, 5))
+>T : Symbol(T, Decl(jsxGenericArrowFunction.tsx, 0, 11))
+>b : Symbol(b, Decl(jsxGenericArrowFunction.tsx, 0, 14))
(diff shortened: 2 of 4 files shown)


poteboy:

DanielRosenwasser:
This would technically be a breaking change for if someone wrote something like

```tsx
const a = <Formatted>(b: T) => b</Formatted>
```

I'm surprised that we don't have a test like that at all though.

poteboy:
@DanielRosenwasser 
You're right, I didn't take this case into account until the CI failed. Parsing this case as an arrow function and backtracking when we hit the JSX closing element would solve the issue, but it could potentially have exponential time complexity. If that's acceptable, it could be a viable solution.

PR #39199, opened 2020-06-22
Add emit support for jsx/jsxs experimental jsx runtime api

weswigham:
This adds two new options for the `jsx` compiler option - `react-jsx`, and `react-jsxdev`. They use the `jsx` and `jsxDEV` constructors and the `react/jsx-runtime` and `react/jsx-dev-runtime` entrypoints, respectively. The primary differences, as noted in the linked issue, are that `children` are passed as part of the `props` object, and the `key` is passed as a separate argument. (The `jsxDEV` constructor further takes some debug data.) In addition, this adds a `jsxImportSource` compiler option (and `@jsxImportSource` file pragma) to control the root specifier the jsx runtime is imported from (which defaults to `react`). In it's current state, I believe this is usable, but there are some known flaws:

* The checker/program need to load the `jsxImportSource` module and use it as the source for the JSX namespace automatically. Presently you need to include it in the program yourself (ie, with a `types` directive or compiler option). This is moreso a TODO for myself, since I don't think there's any question to what we _should_ do (implicitly include the package or `@types` version in the program). This is more interesting for the `jsxImportSource` pragma, which, for all intents and purposes, is an `import` statement. We have a lot of services for _real_ import statements - should they be applied to the pragma, where possible?
* Babel additionally supports a `jsxRuntime` pragma (with values `classic` or `automatic`) to swap between the old emit and the new one. This is supportable; however currently I only key our emit off the `jsx` compiler option, and the presence of the `jsxImportSource` pragma.
* In some cases, the new transform still falls back to `createElement` (when a `key` prop follows a spread) - this is a kind of soft deprecation; however it still implicitly imports the jsx runtime and pulls in `createElement`. This PR does not yet do this, as, awkwardly enough, the module `createElement` is supposed to come from is _different_ from the `jsx-runtime` module entrypoint. I could make this work, but I think this is something we should provide some feedback on - so long as there's still a `createElement` fallback required in the API, why can't it be exported by the same entrypoint exporting `jsx` or `jsxDEV`?
* Without an import or export, a file is not marked as a module, and will not have an import automatically added. Should this issue an error? Or should, under these `jsx` modes, the presence of a jsx tag imply module-ness?

Fixes #34547

This is currently experimental in babel, and AFAIK not yet supported in a stable version of `react`; as such, we should probably continue to experiment with it for awhile before merging it into a stable release.

# TL;DR
When `jsx: react-jsx` (and `target: esnext`) is set,
```ts
[...]

PR #59500, opened 2024-08-01
Add support for the @jsxruntime pragma

### LINKED ISSUE #59340: jsxRuntime pragma does not work
### 🔎 Search Terms

jsxRuntime

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about jsxRuntime

### ⏯ Playground Link

https://www.typescriptlang.org/play/?jsx=4&ts=5.6.0-dev.20240717#code/PQKgBAAgVgzgHgJQK4DsAuBLAtgUzAYwBsBDGGDfMEYAKBzgAcB7AJzQKZRnYAkdDCTAOqtCAEzABeMAAoAlFIB8YADwALAIyK+ApmADuosSuCbFAbiA

### 💻 Code

```tsx
/* @jsxRuntime classic */
export const HelloWorld = () => <h1>Hello world</h1>;
```

### 🙁 Actual behavior

The `jsxRuntime` pragma is not respected and function emit follows the `jsx` option in the TSConfig.

For example, with `"jsx": "react-jsx"`, the function is emitted as

```js
export const HelloWorld = () => _jsx("h1", { children: "Hello world" })
```

### 🙂 Expected behavior

The `jsxRuntime` pragma is respected and the function is emitted as

```js
export const HelloWorld = () => React.createElement("h1", null, "Hello world")
```

The code above is copied verbatim from the [TSConfig Reference](https://www.typescriptlang.org/tsconfig/#jsx)

weswigham:
Not sure who updated [the docs](https://www.typescriptlang.org/tsconfig/#jsx) to imply we supported it, but [we didn't, and haven't since our initial implementation of the new jsx transform](https://github.com/microsoft/TypeScript/pull/39199). As I said in that PR, supporting it, if we'd like to, is fairly easy. Since it seems like it was assumed we did, here we are.

Fixes #59340

### DIFF
--- a/src/compiler/utilities.ts
+++ b/src/compiler/utilities.ts
@@ -9164,10 +9164,16 @@ export function getJSXTransformEnabled(options: CompilerOptions): boolean {
 export function getJSXImplicitImportBase(compilerOptions: CompilerOptions, file?: SourceFile): string | undefined {
     const jsxImportSourcePragmas = file?.pragmas.get("jsximportsource");
     const jsxImportSourcePragma = isArray(jsxImportSourcePragmas) ? jsxImportSourcePragmas[jsxImportSourcePragmas.length - 1] : jsxImportSourcePragmas;
+    const jsxRuntimePragmas = file?.pragmas.get("jsxruntime");
+    const jsxRuntimePragma = isArray(jsxRuntimePragmas) ? jsxRuntimePragmas[jsxRuntimePragmas.length - 1] : jsxRuntimePragmas;
+    if (jsxRuntimePragma?.arguments.factory === "classic") {
+        return undefined;
+    }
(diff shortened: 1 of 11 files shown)


jakebailey:
That was changed in https://github.com/microsoft/TypeScript-Website/pull/3064/files; we could just remove this line. 

Not sure if something like esbuild supports this either?

andrewbranch:
Where do the values "classic" and "automatic" come from?

jakebailey:
I actually think I just missed a typo in the docs; shouldn't they have been `@jsxRuntime react` and `jsxRuntime react-jsx` to match the `jsx` config option?

andrewbranch (approved):

PR #59506, opened 2024-08-01
Disambiguate `BuiltinIterator`/`BuiltinIteratorReturn`

### LINKED ISSUE #59444: TS 5.6 Beta `BuiltinIteratorReturn`: can this be the default type argument of `BuiltinIterator`'s `TReturn`?
### Acknowledgement

### Comment

#58243 introduced the `BuiltinIteratorReturn` type and used it everywhere in type definition.

Then #58222 introduced the `BuiltinIterator` type and replaced some `IterableIterator` usage with `BuiltinIterator`.

I wonder if `BuiltinIteratorReturn` could be set to the default type argument of the `TReturn` type parameter of `BuiltinIterator`, e.g.:

```ts
// lib.es2015.iterable.d.ts
interface BuiltinIterator<T, TReturn = BuiltinIteratorReturn, TNext = any> extends Iterator<T, TReturn, TNext> {
    [Symbol.iterator](): BuiltinIterator<T, TReturn, TNext>;
}
```

#58463 mentions that `BuiltinIteratorReturn` cannot be used for `IterableIterator` because `IterableIterator` is a general purpose type not only for built-in iterators. However, this doesn't apply to `BuiltinIterator`. I think we can safely use `BuiltinIteratorReturn` specifically for `BuiltinIterator`'s default type argument.

The downside of current approach (specify `BuiltinIteratorReturn` everywhere) is that it is too easy to disable the benefit of the new compiler option by manually using the `BuiltinIterator` type:

```ts
const nums = new Set([1, 2, 3, 4, 5]);

// oops! This is BuiltinIterator<number, any, any> regardless of compiler option
const values: BuiltinIterator<number> = nums.values();
```

rbuckton:
In #58243 we introduced the `--strictBuiltinIteratorReturn` flag and the `BuiltinIteratorReturn` type to improve type checking for iterator results produced by the iterators of built-ins like `Array`, `Map`, `Set`, etc. While in #58222 we separately introduced the type `BuiltinIterator` to identify the shape of the new `Iterator.prototype` used by built-ins like `Array`, `Map`, and `Set`, but also for Generators. Where these two features touch, you will generally see a type reference like `BuiltinIterator<T, BuiltinIteratorReturn>`.

Despite the overlap between these two features, there is a subtle distinction between the two mechanisms that are causing some confusion for users. In #59444, it was requested that we make `BuiltinIteratorReturn` the default for the `TReturn` type parameter of `BuiltinIterator`, but this is problematic since `Generator` now inherits from `BuiltinIterator`, but the default inference we make for the `TReturn` of a `Generator` is `void`, which is not assignable to `BuiltinIteratorReturn` when the `strictBuiltinIteratorReturn` flag is set. On its own, this is not terribly problematic since a `Generator` instantiation would properly instantiate `BuiltinIterator` with the correct return type, it does mean that the following code sample would produce an error:
```ts
function* f() { yield ""; }
const it: BuiltinIterator<string> = f(); // error: `void` is not assignable to `undefined`
```

In order to avoid this ambiguity, this PR seeks to make a few small changes to clean up the ambiguity:

- What is currently `BuiltinIterator<T, TReturn, TNext>` will be renamed to `IteratorObject<T, TReturn, TNext>`. This interface is what will describe the global `Iterator.prototype` along with the iterator helpers added in #58222.
- As `IteratorObject` is a new interface, we can use `unknown` instead of `any` as the defaults for `TReturn` and `TNext`, which will be far more type-safe.
- `Generator<T, TReturn, TNext>` will inherit from `IteratorObject<T, TReturn, TNext>`.
- Individual interfaces (`ArrayIterator<T>`, `MapIterator<T>`, etc.) will be added as necessary for each built-in what will inherit from `IteratorObject<T, BuiltinIteratorReturn, unknown>`. Since the iterators produced by these built-ins will always return `undefined`, the `TReturn` type parameter is explicitly set to `BuiltinIteratorReturn`.

The individual built-in interfaces we will add are based on the unique implementations defined within ECMA-262, ECMA-404 (`Intl`), and the DOM:
- [`ArrayIterator<T>`](https://tc39.es/ecma262/#sec-array-iterator-objects) - `Array`, `Uint8Array`, etc.
- [`MapIterator<T>`](https://tc39.es/ecma262/#sec-map-iterator-objects) - `Map`
- [`SetIterator<T>`](https://tc39.es/ecma262/#sec-set-iterator-objects) - `Set`
- [`StringIterator<T>`](https://tc39.es/ecma262/#sec-string-iterator-objects) - `String`
- [`RegExpStringIterator<T>`](https://tc39.es/ecma262/#sec-regexp-string-iterator-objects) - `String.prototype[Symbol.matchAll]`/`RegExp.prototype[Symbol.matchAll]`
- [`SegmentIterator<T>`](https://tc39.es/ecma402/#sec-segment-iterator-objects) - `Intl.Segments`

Per the WebIDL spec, lists that can be accessed by index may have "[value iterators](https://webidl.spec.whatwg.org/#dfn-value-iterator)" which are defined to return `ArrayIterator`.

Per the WebIDL spec, lists that can be accessed by key may have "[pair iterators](https://webidl.spec.whatwg.org/#dfn-pair-iterator)" which results in a unique prototype per each declaring interface. These include:
- [`FormDataIterator<T>`](https://xhr.spec.whatwg.org/#interface-formdata) - `FormData`
- [`HeadersIterator<T>`](https://fetch.spec.whatwg.org/#headers-class) - `Headers`
- [`StylePropertyMapReadOnlyIterator<T>`](https://www.w3.org/TR/css-typed-om-1/#the-stylepropertymap) - `StylePropertyMapReadOnly`
- [`URLSearchParamsIterator<T>`](https://url.spec.whatwg.org/#interface-urlsearchparams) - `URLSearchParams`

Per the WebIDL spec, [async iterable declarations](https://webidl.spec.whatwg.org/#idl-async-iterable) each have a unique prototype per each declaring interface. These include:
- [`FileSystemDirectoryHandleAsyncIterator<T>`](https://fs.spec.whatwg.org/#api-filesystemdirectoryhandle) - `FileSystemDirectoryHandle`
- [`ReadableStreamAsyncIterator<T>`](https://streams.spec.whatwg.org/#rs-class-definition) - `ReadableStream`

This results in changes to our declarations similar to the following:

**es2015.iterable.d.ts**
```ts
...
// User-defined `Iterable` + `Iterator`. Does not inherit from `Iterator.prototype`
interface IterableIterator<T, TReturn = any, TNext = any> extends Iterator<T, TReturn, TNext> {
    [Symbol.iterator](): IterableIterator<T, TReturn, TNext>;
}

// Prototype for built-in iterators and generators. Interface where iterator helpers will be defined.
// NOTE: Was previously named `BuiltinIterator<T, TReturn, TNext>`
interface IteratorObject<T, TReturn, TNext> extends Iterator<T, TReturn, TNext> {
    [Symbol.iterator](): IteratorObject<T, TReturn, TNext>;
}

// Type driven by `--strictBuiltinIteratorReturn`
type BuiltinIteratorReturn = intrinsic;

interface ArrayIterator<T> extends IteratorObject<T, BuiltinIteratorReturn, unknown> {
    [Symbol.iterator](): ArrayIterator<T>;
}

interface Array<T> {
    // NOTE: previously returned `BuiltinIterator<T, BuiltinIteratorReturn>`
    [Symbol.iterator](): ArrayIterator<T>;
    entries(): ArrayIterator<[number, T]>;
    keys(): ArrayIterator<number>;
    values(): ArrayIterator<T>;
}

interface MapIterator<T> extends IteratorObject<T, BuiltinIteratorReturn, unknown> {
    [Symbol.iterator](): MapIterator<T>;
}

// etc.

...
```

**esnext.iterator.d.ts**
```ts
...
// Iterator helpers, as defined on global `Iterator.prototype`
interface IteratorObject<T, TReturn, TNext> {
    map<U>(callbackfn: (value: T, index: number) => U): IteratorObject<U, undefined, unknown>;
    filter(predicate: (value: T, index: number) => unknown): IteratorObject<T, undefined, unknown>;
    ...
}
...
```

**Examples**

```ts
function* f() { yield ""; } 
// f: () => Generator<string, void, unknown>

// ok: `Generator<string, void, unknown>` is assignable to `IteratorObject<string, unknown, unknown>`
const it1: IteratorObject<string, unknown, unknown> = f();

const it2 = [1, 2, 3].values();
// it2: ArrayIterator<number>

// ok
const it3: IteratorObject<number, unknown, unknown> = it2;

class RepeatIterator<T> extends Iterator<T> {
  #value: T;
  constructor(value: T) { this.#value = value; }
  next(value?: unknown): IteratorResult<T, undefined> {
    return { done: false, value: this.#value };
  }
}

// ok
const it4: IteratorObject<number, unknown, unknown> = new RepeatIterator(0);
it4.take(4).toArray(); // [0, 0, 0, 0]
```

Fixes #59444


rbuckton:
Note that this is going to require updates to the DOM lib generator once this is in nightly.

rbuckton:
@typescript-bot perf test
@typescript-bot run dt
@typescript-bot test top400

rbuckton:
Instead of continuing to use `BuiltinIterator`, we could instead properly define `ArrayIterator<T>`, `MapIterator<T>`, and `SetIterator<T>`, since those do have runtime counterparts (if not constructors), e.g.:

**lib.es2015.d.ts**
```ts
...
interface ArrayIterator<T> extends IteratorObject<T, BuiltinIteratorReturn, unknown> {
}

interface Array<T> {
    [Symbol.iterator](): ArrayIterator<T>;
    entries(): ArrayIterator<[number, T]>;
    keys(): ArrayIterator<number>;
    values(): ArrayIterator<T>;
}

interface MapIterator<T> extends IteratorObject<T, BuiltinIteratorReturn, unknown> {
}
...

interface SetIterator<T> extends IteratorObject<T, BuiltinIteratorReturn, unknown> {
}
...
```

**examples**
```ts
let ArrayIteratorPrototype = Object.getPrototypeOf([].values());
ArrayIteratorPrototype; // [object Array Iterator]
ArrayIteratorPrototype.x; // undefined
ArrayIteratorPrototype.x = 1;
[].keys().x; // 1
new Uint8Array()[Symbol.iterator]().x; // 1

let MapIteratorPrototype = Object.getPrototypeOf(new Map().entries());
MapIteratorPrototype; // [object Map Iterator]
MapIteratorPrototype.x; // undefined
MapIteratorPrototype.x = 2;
new Map().keys().x; // 2

let SetIteratorPrototype = Object.getPrototypeOf(new Set().keys());
SetIteratorPrototype; // [object Set Iterator]
SetIteratorPrototype; // [object Map Iterator]
SetIteratorPrototype.x; // undefined
SetIteratorPrototype.x = 3;
new Set().entries().x; // 3
```

rbuckton:
Regarding the failing DT/user tests: These all look to be expected errors resulting from this change. I will follow up with relevant PRs for DT.

As there are a few failing user tests due to DT dependence, it may be that we need to a transitional (deprecated) `TReturn` type parameter for `BuiltinIterator`, i.e.:

```ts
interface BuiltinIterator<
    T,
    /** @deprecated */ TReturn extends BuiltinIteratorReturn = BuiltinIteratorReturn
> extends IteratorObject<T, TReturn, unknown> {
    [Symbol.iterator](): BuiltinIterator<T, TReturn>;
}
```

If we decide to add one, we should be able to drop the deprecated `TReturn` by release.

rbuckton:
@typescript-bot: pack this

rbuckton:
@typescript-bot perf test
@typescript-bot run dt
@typescript-bot test top400

rbuckton:
@typescript-bot perf test
@typescript-bot run dt
@typescript-bot test top400

jakebailey (approved):

ExE-Boss:
Maybe `IteratorPrototype` might’ve been a better name instead of `IteratorObject`.

Issue #59490, opened 2024-07-31
React's `ComponentProps` type issues in TypeScript 5.5.4


jakubmazanec:
### 🔎 Search Terms

react, forwardRef, component props, ComponentProps

### 🕗 Version & Regression Information

- This changed between versions 5.4.5 and 5.5.5

### ⏯ Playground Link

[...]

PR #59510, opened 2024-08-02
Fixed regression in generic signature instantiation with default type args

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/59490

complements the fix from https://github.com/microsoft/TypeScript/pull/59121 so the correct cloned type parameter is used as the inference target

cc @weswigham


jakebailey:
@typescript-bot test it

weswigham (approved):
Needs a merge conflict resolved before it can be merged - but I'm starting to think it'd be nice to pull `sameMap(signature.typeParameters!, tp => tp.mapper ? instantiateType(tp, tp.mapper) : tp)` into a helper function like `getTypeParametersForMapper(signature)` for the two locations we use it (an inference context is a type mapper).

Andarist:
done and done :)

jakebailey (approved):

jakebailey:

This is the same fix as #59121 which we also backported to 5.5.

Bongdautay:
quaroibn1@gmail.com

PR #59514, opened 2024-08-02
Use global `Iterator.prototype` for down-level generators

### LINKED ISSUE #59513: Generators compiled with `--target es5 --lib esnext` do not support iterator helpers
### 🔎 Search Terms

generator iterator helper map esnext prototype

- This changed in commit or PR #58222

### ⏯ Playground Link

https://www.typescriptlang.org/play/?target=1&lib=lib.esnext.d.ts&ts=5.6.0-beta#code/PTAEAEBcEMCcHMCmkBcpEGcCsAoEEAbASwCM1MA7RAD0hxwGMB7CjSUOUAXlAG0BGADSgATMIDMwgCwBdANw44AOgBu0AgFdMACgCUSgLbQADtqLcAfKHMAqUbrmh8vMaCnCAbMIAcM+gDMNCgZIIhZQO389UABvHFBQAE8iRAIAEwi+IVEJaXkcAF8cKP0jU3MuK1t7R3xYINCDRHRYWCZYNAByKKVe0pNO6wxQCiZ2aFBA4NCWIA

### 💻 Code

```ts
// @target: es5
// @lib: esnext

const ar = [1, 2, 3, 4];
ar.values().map(i => i * 2); // [2, 4, 6, 8]

function * f() {
  yield * [1, 2, 3, 4];
}
f().map(i => i * 2); // error: 'f(...).map' is not a function
```

### 🙁 Actual behavior

The last line above throws even when `Iterator.prototype` is present, be it from the host or a polyfill.

### 🙂 Expected behavior

The last line above should not throw when the global `Iterator.prototype` is present and can be used as a prototype.

_No response_

rbuckton:
If you are compiling generators with `--target ES5 --lib esnext` the types of those generators will also include the iterator helpers methods like `map`, `filter`, etc., though our down-level generator emit will not have those methods even when running with newer editions of V8 that do support them:

```ts
// @target: es5
// @lib: esnext

const ar = [1, 2, 3, 4];
ar.values().map(i => i * 2); // [2, 4, 6, 8]

function * f() {
  yield * [1, 2, 3, 4];
}
f().map(i => i * 2); // error: 'f(...).map' is not a function
```

Even when not running in latest V8, iterator helpers can be polyfilled, e.g.:
```js
if (typeof Iterator === "undefined") {
    let Iterator = function () {};
    Iterator.prototype = Object.getPrototypeOf(Object.getPrototypeOf([].values()));
    Iterator.prototype.map = ...;
    Iterator.prototype.filter = ...;
    globalThis.Iterator = Iterator;
}
```

To better support both scenarios, this makes a small change to our `__generator` helper to use the global `Iterator.prototype` object as the prototype for down-level generators, if it is present.

This also does the same for our `__asyncGenerator` helper when a global `AsyncIterator.prototype` object is present, even though async iterator helpers are still at Stage 2. Support for `AsyncIterator.prototype` seems harmless enough to add at this time but can be postponed to a later PR if necessary.

Please note that this will also require the same change in `tslib`.

Fixes #59513

### DIFF
--- a/src/compiler/factory/emitHelpers.ts
+++ b/src/compiler/factory/emitHelpers.ts
@@ -836,7 +836,7 @@ const asyncGeneratorHelper: UnscopedEmitHelper = {
         var __asyncGenerator = (this && this.__asyncGenerator) || function (thisArg, _arguments, generator) {
             if (!Symbol.asyncIterator) throw new TypeError("Symbol.asyncIterator is not defined.");
             var g = generator.apply(thisArg, _arguments || []), i, q = [];
-            return i = {}, verb("next"), verb("throw"), verb("return", awaitReturn), i[Symbol.asyncIterator] = function () { return this; }, i;
+            return i = Object.create((typeof AsyncIterator === "function" ? AsyncIterator : Object).prototype), verb("next"), verb("throw"), verb("return", awaitReturn), i[Symbol.asyncIterator] = function () { return this; }, i;
             function awaitReturn(f) { return function (v) { return Promise.resolve(v).then(f, reject); }; }
             function verb(n, f) { if (g[n]) { i[n] = function (v) { return new Promise(function (a, b) { q.push([n, v, a, b]) > 1 || resume(n, v); }); }; if (f) i[n] = f(i[n]); } }
             function resume(n, v) { try { step(g[n](v)); } catch (e) { settle(q[0][3], e); } }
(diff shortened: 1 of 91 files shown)


rbuckton:
@typescript-bot: pack this

src/compiler/factory/emitHelpers.ts
```diff
@@ -1102,8 +1102,8 @@ const generatorHelper: UnscopedEmitHelper = {
     priority: 6,
     text: `
             var __generator = (this && this.__generator) || function (thisArg, body) {
-                var _ = { label: 0, sent: function() { if (t[0] & 1) throw t[1]; return t[1]; }, trys: [], ops: [] }, f, y, t, g;
-                return g = { next: verb(0), "throw": verb(1), "return": verb(2) }, typeof Symbol === "function" && (g[Symbol.iterator] = function() { return this; }), g;
+                var _ = { label: 0, sent: function() { if (t[0] & 1) throw t[1]; return t[1]; }, trys: [], ops: [] }, f, y, t, g = Object.create((typeof Iterator === "function" ? Iterator : Object).prototype);
```
jakebailey, line 1105: I assume it's not a problem if someone redefines the global `Iterator`, given we don't care if someone does that for other global symbols?

jakebailey (approved):
This needs a `tslib` update too, right?

src/compiler/factory/emitHelpers.ts
rbuckton: Correct

rbuckton:
> This needs a `tslib` update too, right?

Correct. The tslib PR is https://github.com/microsoft/tslib/pull/267

PR #59516, opened 2024-08-02
Properly account for type parameters introduced by contextual types

### LINKED ISSUE #59450: Conditional type evaluation not deferred when it depends on an implicitly-present type parameter

ahejlsberg:
Fixes #59450.

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -12541,7 +12541,8 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             if (!node) {
                 return undefined;
             }
-            switch (node.kind) {
+            const kind = node.kind;
+            switch (kind) {
                 case SyntaxKind.ClassDeclaration:
                 case SyntaxKind.ClassExpression:
                 case SyntaxKind.InterfaceDeclaration:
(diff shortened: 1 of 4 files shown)


ahejlsberg:
@typescript-bot test it

ahejlsberg:
Performance is unaffected. One [new error](https://github.com/honojs/hono/blob/1fafc7aec87f1afe68b37506df58a66acd9f29aa/src/types.test.ts#L2260) in `honojs/hono`, a top 400 repo. The new error is correct, but wasn't previously detected because instantiations of the type of `{ data }` object literal weren't correctly resolved for the contextually injected `TData` type parameter. So, I think this is good to merge.

m-shaka:
Yeah, it's me who added this test of hono.js and I expected the failure. You can go forward

src/compiler/checker.ts
```diff
@@ -12563,15 +12564,21 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 case SyntaxKind.MappedType:
                 case SyntaxKind.ConditionalType: {
                     const outerTypeParameters = getOuterTypeParameters(node, includeThisTypes);
-                    if (node.kind === SyntaxKind.MappedType) {
+                    if ((kind === SyntaxKind.FunctionExpression || kind === SyntaxKind.ArrowFunction || isObjectLiteralMethod(node)) && isContextSensitive(node as Expression | MethodDeclaration)) {
+                        const signature = firstOrUndefined(getSignaturesOfType(getTypeOfSymbol(getSymbolOfDeclaration(node as FunctionLikeDeclaration)), SignatureKind.Call));
+                        if (signature && signature.typeParameters) {
+                            return [...(outerTypeParameters || emptyArray), ...signature.typeParameters];
```
jakebailey, line 12570: Should this be concatenate like the other cases?

ahejlsberg: We are required to return a mutable array, and `concatenate` only returns a mutable array when both arguments are mutable arrays. In this case, `signature.typeParameters` is read-only.

Can't get rid of the empty array, spreading `undefined` causes an exception.

jakebailey: I meant more that `concatenate` can accept `undefined` so you don't have to `||` things, but, yeah, it doesn't return a copy.

Though it's sort of confusing given all of the surrounding code uses `append` / `concatenate`

jakebailey (approved):

PR #59518, opened 2024-08-02
Update to eslint v9, typescript-eslint v8

jakebailey:
Blocked on:

- https://github.com/a-tarasyuk/eslint-formatter-autolinkable-stylish/issues/33 (or we remove the formatter; VS Code parses ESLint's output now too sorta)
	- Worked around with an override, good enough for now.
- https://github.com/typescript-eslint/typescript-eslint/pull/9694 to be released


jakebailey:
@webpro Do you happen to know why knip would be complaining about `@typescript-eslint/rule-tester` being unused, even though debug logs seem to imply that `scripts/eslint/tests/support/RuleTester.cjs` is analyzed?

webpro:
> @webpro Do you happen to know why knip would be complaining about `@typescript-eslint/rule-tester` being unused, even though debug logs seem to imply that `scripts/eslint/tests/support/RuleTester.cjs` is analyzed?

Not entirely sure, but some relevant bits include:

- Knip does not automatically split projects based on `tsconfig.json` files. This repo is a single workspace to Knip and all files are in a single `ts.Program` with a single custom module resolver function.
- No root `tsconfig.json` means no value for `compilerOptions.moduleResolution` when `ts.resolveModuleName()` is invoked. That could explain why `@typescript-eslint/utils` was resolved (by `@import` in JSDoc tag) but not `@typescript-eslint/rule-tester` (by `require` call). 
- In that custom function Knip uses an additional module resolver (prior to `ts.resolveModuleName`). Coincidentally last week the `resolve` package that this is based on has been replaced with `enhanced-resolve` and that seems to do the trick here. Not exactly sure why, though. I expect this fix to be released this week.

jakebailey:
Updated knip and yeah, that works, thanks!

sheetalkamat (approved):

PR #59519, opened 2024-08-02
Suggest simplifications for overzealous shifts

weswigham:
This PR adds a suggestion on bitwise shifts where the right side operand is evaluable (which is implicitly non-`bigint`) to be outside of the range `-31` to `31` to simplify the shift down into that range. Doing so should make it a bit more obvious when you've accidentally exceeded the maximum shift available for a number. 😄 

This suggestion is escalated to an error in the context of `enum` members, since that's where it's most likely an overshift may result in an unintentionally duplicated `enum` member that is otherwise silent.

So this:
```ts
1 << 32;
```
results in
```
suggestion TS6807: This operation can be simplified. This shift is identical to `1 << 0`.
```
while this:
```ts
enum A {
  M = 1 << 32,
}
```
results in
```
error TS6807: This operation can be simplified. This shift is identical to `1 << 0`.
```
This logic is applied for all the shift operators - `<<`, `>>`, `>>>`, `<<=`, `>>=`, and `>>>=`.


jakebailey:
Can this me made a regular error temporarily to run user/topXYZ?

(I also sort of feel like we should make it a regular error in general...)

weswigham:
Oh boy a failing self-check build :)

Always error user/top600 results rolling in [here](https://github.com/microsoft/TypeScript/pull/59520#issuecomment-2266116210).

gabritto:
Are there other common ways people can end up with two enum members of the same value by accident?

src/compiler/checker.ts
```diff
@@ -39752,6 +39752,28 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
                         checkAssignmentOperator(resultType);
+                        switch (operator) {
+                            case SyntaxKind.LessThanLessThanToken:
+                            case SyntaxKind.LessThanLessThanEqualsToken:
+                            case SyntaxKind.GreaterThanGreaterThanToken:
+                            case SyntaxKind.GreaterThanGreaterThanEqualsToken:
+                            case SyntaxKind.GreaterThanGreaterThanGreaterThanToken:
+                            case SyntaxKind.GreaterThanGreaterThanGreaterThanEqualsToken:
+                                const rhsEval = evaluate(right);
+                                if (typeof rhsEval.value === "number" && Math.abs(rhsEval.value) >= 32) {
```
jakebailey, line 39763: Should this convert the value to an integer or something?

ehoogeveen-medweb: The `Math.abs` also seems unnecessary - any value below 0 seems wrong and `1 << -1 == 1 << 31`, `1 << -2 == 1 << 30` etc. so `Math.abs` "reverses" it.

Interestingly, `1 << -0.9 == 1 << 0` so I guess you want `Math.trunc` (or `| 0`) rather than `Math.floor`.

weswigham: > any value below 0 seems wrong

It's a useful shorthand. Similar to negative slice indices in python. Not doing an `abs` on the evaluated result here, either - just using it to measure the magnitude of the number to issue the error or not.

> Should this convert the value to an integer or something?

Eh. Don't really need to. Fractional parts just get dropped anyway. Issuing an error on *any* fractional literal in a shift position would probably be fine (`0.5` is the same as `0` for this operator after all), but I don't think it's particularly related to this range error.

weswigham:
> Are there other common ways people can end up with two enum members of the same value by accident?

Other arithmetic errors in calculated enum values, perhaps, but it may also be intentional. Especially with flags enums and bitwise arithmetic - you may end up deriving two sets of flags that just so happen to resolve to the same thing via different means.

gabritto:
@typescript-bot pack this

gabritto:
@typescript-bot test it

gabritto (approved):

gabritto:

gabritto:
@weswigham there seems to be a slight increase in check times for our codebase but also for vscode, webpack and xstate. Any ideas of why that could be? Do we now issue more suggestions on those repos? Is it just the extra work of evaluating operators?

weswigham:
Yeah, just a little bit of extra work evaluating the shifts, I imagine. `vscode` has... 470 left shifts, and a handful of right shifts and rotates. I don't think any actually trigger the suggestion.

PR #59520, opened 2024-08-02
Always error on overshifts

weswigham:
To see if making it always an error instead of a suggestion is viable.


weswigham:
@typescript-bot test top600

weswigham:
Hm, unsure if those are legitimate errors in webpack or funky
AssemblyScript-isms that hint how to handle the numbers to the
runtime/compiler.

RyanCavanaugh:
Seems idiomatic in asm? It's declared
```js
let result: u64;
```
so presumably you can 32-shift it

weswigham:
Well, not breaking AssemblyScript `u64` idioms is probably as good a reason as any to leave it as a suggestion only outside enums.

PR #59523, opened 2024-08-02
Enable existing commit characters in certain locations where `isNewIdentifier` is true

### LINKED ISSUE #25691: isNewIdentifierLocation is always set to true inside function call
**TypeScript Version:**  3.0.0-dev.20180713

**Search Terms:** 
- isNewIdentifierLocation
- completions
- suggest

**Bug**
The `completionInfo` request seems to always return `isNewIdentifierLocation` when inside a function call such as:

```ts
console.log(c|)
```

**Expected behavior:**
I believe that `isNewIdentifierLocation` should only be returned when the user is starting to write a arrow function with a parenthesized parameter list, such as:

```ts
(c|)
```

or

```ts
console.log((c|)
```

### LINKED ISSUE #31244: Suggestions not accepted when using commit characters
- VSCode Version: 1.33.1
- OS Version: 10.14.4

Steps to Reproduce:

1. `mkdir test`
2. `code --disable-extensions --user-data-dir temp test`
3. Create new JS file

```
const mySpecialVar = 1;
const foo = mySpec<cursor>
```

When suggestions appear, press `;` (commit character).

Expected: suggestion is accepted + `;` is inserted.
Actual: `;` is inserted.

OTOH, this does work as expected:

```
const mySpecialVar = 1;
const foo = 1 + mySpec<cursor>
```

gabritto:
Fixes #31244.
Fixes #25691.

As I said in #59339, our existing behavior, implemented based on vscode's behavior, is that whenever `isNewIdentifierLocation` is `true`, we disable all commit characters, otherwise the commit characters are `.`, `,`, `;`.

However, there have been issues opened over time that this approach is too conservative.
This PR enables some of the usual commit characters in some places where `isNewIdentifierLocation` is `true` and that previously we'd enable no commit characters.

Mainly, this PR enables all the three commit characters in every position where an expression is expected. In some of those positions, where we are inside an open parenthesis, the user could be typing a parameter list for an arrow expression, so we disable `,` as a commit character.

I considered adding space as a commit character too, but we would have to disable it pretty much everywhere where `isNewIdentifierLocation` is `true` because space could validly follow a new identifier in all of those places, so I figured users wouldn't be able to use it reliably as a commit character and it didn't seem like a good idea.

This PR should be testable in vscode insiders once it picks up https://github.com/microsoft/vscode/pull/223541.


gabritto:
@typescript-bot pack this

andrewbranch (approved):

gabritto:
This should be testable using vscode insiders and setting TS to this PR (see [this](https://github.com/microsoft/TypeScript/pull/59523#issuecomment-2266283972)).

PR #59526, opened 2024-08-03
Add custom eslint rule 'no-array-mutating-method-expressions'

jakebailey:
Motivated by the bug found in #59523.

Needs tests.

### DIFF
--- a/scripts/eslint/rules/no-array-mutation-method-returns.cjs
+++ b/scripts/eslint/rules/no-array-mutation-method-returns.cjs
@@ -30,7 +30,12 @@ module.exports = createRule({
 
         const checker = services.program.getTypeChecker();
 
-        /** @type {(callee: TSESTree.MemberExpression) => boolean} */
+        /**
+         * This is a heuristic to ignore cases where the mutating method appears to be
+         * operating on a "fresh" array.
+         *
+         * @type {(callee: TSESTree.MemberExpression) => boolean}
+         */
         const isFreshArray = callee => {
             const object = callee.object;
 
--- a/scripts/eslint/rules/no-array-mutation-method-returns.cjs
+++ b/scripts/eslint/rules/no-array-mutation-method-returns.cjs
@@ -89,13 +89,11 @@ module.exports = createRule({
 
         /** @type {(callee: TSESTree.MemberExpression & { parent: TSESTree.CallExpression; }, method: string, toMethod: string | undefined) => void} */
         const check = (callee, method, toMethod) => {
+            if (callee.parent.parent.type === "ExpressionStatement") return;
             if (isFreshArray(callee)) return;
 
             const calleeObjType = getConstrainedTypeAtLocation(services, callee.object);
-
             if (!isTypeArrayTypeOrUnionOfArrayTypes(calleeObjType, checker)) return;
-
-            if (callee.parent.parent.type === "ExpressionStatement") return;
 
             if (toMethod) {
                 context.report({ node: callee.property, messageId: "noSideEffectUseToMethod", data: { method, toMethod } });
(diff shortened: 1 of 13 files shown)


RyanCavanaugh:
We could instead add an earlier overload via declaration merging to return `void`. Might be less work and/or better, haven't tried it

jakebailey:
I'm not totally sure how we'd add an overload before the lib.d.ts, but I guess it must be possible because ts-reset or whatever can do it?

But, that wouldn't allow us to ignore the calls that happen after we're sure the array has been cloned (which is what this PR does); we don't have that many calls to sort but most of them are actually ignored by this rule.

RyanCavanaugh:
Overloads from declaration merging to in "before" the overloads they merge with

RyanCavanaugh (approved):

scripts/eslint/rules/no-array-mutation-method-returns.cjs
```diff
@@ -0,0 +1,146 @@
...
+void 0;
+
+module.exports = createRule({
+    name: "no-array-mutation-method-returns",
+    meta: {
+        docs: {
+            description: ``,
+        },
+        messages: {
+            noSideEffectUse: `This call to {{method}} appears to be unintentional as it appears in an expression position. Sort the array in a separate statement, or use slice/{{toMethod}} to ensure a copy is made.`,
```
jakebailey, line 17: This message with `toMethod` is slightly untrue because we only have `toSorted`, not `toReversed` and `toSpliced` (without defining polyfills for ourselves).

src/harness/projectServiceStateLogger.ts
```diff
@@ -93,7 +94,7 @@ export function patchServiceForStateBaseline(service: ProjectService) {
     function sendLogsToLogger(title: string, logs: StateItemLog[] | undefined) {
         if (!logs) return;
         logger.log(title);
-        logs.sort((a, b) => compareStringsCaseSensitive(a[0], b[0]))
+        toSorted(logs, (a, b) => compareStringsCaseSensitive(a[0], b[0]))
```
jakebailey, line 97: This is a bug.

src/harness/fourslashImpl.ts
```diff
@@ -4058,7 +4058,7 @@ export class TestState {
     public verifyRefactorKindsAvailable(kind: string, expected: string[], preferences = ts.emptyOptions) {
         const refactors = this.getApplicableRefactorsAtSelection("invoked", kind, preferences);
         const availableKinds = ts.flatMap(refactors, refactor => refactor.actions).map(action => action.kind);
-        assert.deepEqual(availableKinds.sort(), expected.sort(), `Expected kinds to be equal`);
+        assert.deepEqual(availableKinds.slice().sort(), expected.slice().sort(), `Expected kinds to be equal`);
```
jakebailey, line 4061: This is a bug, more or less.

src/testRunner/unittests/services/extract/ranges.ts
```diff
@@ -12,7 +12,7 @@ function testExtractRangeFailed(caption: string, s: string, expectedErrors: stri
         const result = ts.refactor.extractSymbol.getRangeToExtract(file, ts.createTextSpanFromRange(selectionRange), /*invoked*/ false);
         assert(result.targetRange === undefined, "failure expected");
         const sortedErrors = result.errors.map(e => e.messageText as string).sort();
-        assert.deepEqual(sortedErrors, expectedErrors.sort(), "unexpected errors");
+        assert.deepEqual(sortedErrors, expectedErrors.slice().sort(), "unexpected errors");
```
jakebailey, line 15: This is a bug.

Issue #58030, opened 2024-04-02
Uncaught Error: Debug Failure. at getThisContainer (while executing getCompletionsAtPosition n JavaScript file)


Andarist:
### 🔎 Search Terms

crash get this container get completions at position javascript example

### 🕗 Version & Regression Information

- This is the behavior in every version I tried

### ⏯ Playground Link

https://www.typescriptlang.org/play?target=1&ts=5.5.0-dev.20240401&filetype=js#code/PQKhCgAIUgBBTAHgQwLYAcA28qQDwBmAltpAHZrwC8ARAOaYCe6AFkQMYD2ZAzgHTsePGgD5ckOEQycATgBdIAVxmYAFMGBl4cgCYU+AI06c5POTOTp2egZ1TAjJsxfTAAzHwAMX4IJ4PjU3NLAFoGZjYuXgEhAEoAblw8YGJsMQk8JDQseEhUTh1FbFoyOgBBMilkOXgaSB14dB5aZFKi5BkQ1qqavgArYUhu1Gqibmaac0Va9Il8VNyKVGoaYdHxmOFZucg+NZqQnhZOAHc+UpC2Bq6dHXO6S6Jr5Fuu9jkiADd4ABpxOb2lRGByOp3uj2uMng+W+4Ku8BCUJhCOQ7y+uQA3v8dsFeEQPtwAFzITCYSCYIhaDqQbwAVh4iR2EgAvv9kgtZskshg0lAQMBwJ9qaUAMrHE4AESIULR30gVEgAG0AOQAEn28GVP0gBEUZDR3FU6qB1XgsUgGOZAF1EuAgA

### 💻 Code

```ts
/**
 * @example
  <file name="glyphicons.css">
    @import url(//netdna.bootstrapcdn.com/bootstrap/3.0.0/css/bootstrap-glyphicons.css);
  </file>
  <example module="ngAnimate" deps="angular-animate.js" animations="true">
    <file name="animations.css">
      .animate-show.ng-hide-add.ng-hide-add-active,
      .animate-show.ng-hide-remove.ng-hide-remove-active {
        transition:all linear 0.5s;
      }
    </file>
  </example>
 */
var ngShowDirective = ['$animate', function($animate) {}];
```

### 🙁 Actual behavior

Requesting completions here `transition:all linear 0./*1*/5s;` leads to an error (with `includeCompletionsWithInsertText`).

### 🙂 Expected behavior

It shouldn't error

### Additional information about the issue

[...]

PR #59527, opened 2024-08-03
Fixed completion crashes related to `JSDocImportTag`

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/58030

### DIFF
--- a/src/compiler/parser.ts
+++ b/src/compiler/parser.ts
@@ -8801,7 +8801,6 @@ namespace Parser {
 
             const comment = doInsideOfContext(NodeFlags.JSDoc, () => parseJSDocCommentWorker(start, length));
             setParent(comment, parent);
-            setParentRecursive(comment, /*incremental*/ false);
 
             if (contextFlags & NodeFlags.JavaScriptFile) {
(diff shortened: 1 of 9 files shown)


src/compiler/parser.ts
```diff
@@ -8801,6 +8801,7 @@ namespace Parser {
 
             const comment = doInsideOfContext(NodeFlags.JSDoc, () => parseJSDocCommentWorker(start, length));
             setParent(comment, parent);
+            setParentRecursive(comment, /*incremental*/ false);
```
Andarist, line 8804: I know it's an added cost but I'm also hesitant to add `setParent` all over the place. Maybe this should be moved somewhere within services? but then it would require a full crawl of the program to find JSDoc comments so nodes within them could get their `.parent` set.

Regular nodes receive their `.parent` from the binder when the typechecker gets created but that's not the case for JSDoc

jakebailey (approved):

sandersn (approved):

Bongdautay:
quaroibn1@gmail.com

PR #50820, opened 2022-09-17
Implement the Stage 3 Decorators Proposal

rbuckton:
This implements support for the Stage 3 Decorators proposal targeting `ESNext` through `ES5` (except where it depends on functionality not available in a specific target, such as WeakMaps for down-level private names).

The following items are *not* currently supported:
- `--emitDecoratorMetadata`, as metadata is currently under discussion in https://github.com/tc39/proposal-decorator-metadata and has not yet reached Stage 3.
- Decorators on a `declare` field.
- Parameter decorators will not be supported until a follow-on proposal has been adopted and has advanced to Stage 3.
  - See https://github.com/tc39/proposal-decorators/issues/47 for additional information.
- Decorators may not change the type of the member or class they decorate. If we decide to allow this capability, we will do so in a later PR. This is under discussion in the following two issues:
  - #50159
  - #49229

With that out of the way, the following items are what *is* supported, or is new or changed for Decorators support in the Stage 3 proposal:
- The `--experimentalDecorators` flag will continue to opt-in to the legacy decorator support (which still continues to support `--emitDecoratorMetadata` and parameter decorators).
- ES Decorators are now supported _without_ the `--experimentalDecorators` flag.
- 🆕 ES Decorators will be transformed when the target is less than `ESNext` (or at least, until such time as the proposal reaches Stage 4).
- 🆕 ES Decorators now accept exactly two arguments: `target` and `context`:
  - `target` &mdash; A value representing the element being decorated:
    - Classes, Methods, `get` accessors, and `set` accessors: This will be the function for that element.
    - Auto-Accessor fields (i.e., `accessor x`): This will be an object with `get` and `set` properties.
    - Fields: This will always be `undefined`.
  - `context` &mdash; An object containing additional context information about the decorated element such as:
    - `kind` - The kind of element (`"class"`, `"method"`, `"getter"`, `"setter"`, `"field"`, `"accessor"`).
    - `name` - The name of the element (either a `string` or `symbol`).
    - `private` - Whether the element has a private name.
    - `static` - Whether the element was declared `static`.
    - `access` - An object with either a `get` property, a `set` property, or both, that is used to read and write to the underlying value on an object.
    - `addInitializer` - A function that can be called to register a callback that is evaluated either when the class is defined or when an instance is created:
      - For static member decorators, initializers run after class decorators have been applied but before static fields are initialized.
      - For Class Decorators, initializers run after all static initializers.
      - For non-static member decorators, initializers run in the constructor before all field initializers are evaluated.
[...]

PR #59538, opened 2024-08-05
Don't treat an instantiation expression as an assertion in skipOuterExpressions

### LINKED ISSUE #59535: Instantiation expressions shouldn't be allowed in write positions
### 🔎 Search Terms

instantiation expression write read positions type arguments functions

### ⏯ Playground Link

https://www.typescriptlang.org/play/?ts=5.6.0-dev.20240805#code/JYOwLgpgTgZghgYwgAgKoGdrIN4ChnIhwC2EAXMumFKAOYDc+yct5hArsQEbSMC+uXKEixEKAPJcAVjiYwQAHgAqAPgAUASgpL+uACYQEAGzhQUCAPYgqyC9IqSpjfYZNnkl62GTtMUALIWCADWFJrIALwqaH7OAPRxyGAAFsDolMkW7EZ6IADk3jzMRkYWAO4Qerh2UgB08goY0NERPn6BIc4GxqYoRhDe8hTK6hqR0UqCCUmp6eiZ2bkFyEVwJeWVyBDAKdC4DU1QLW3QHcG4QA

### 💻 Code

```ts
interface User {
  name: string;
  age: number;
}

interface Obj {
  fn<T>(): T;
}
declare const obj: Obj;

declare const userMock: () => User;

// this shouldn't be allowed
obj.fn<User> = userMock;

declare let fn: <T>() => T

// this shouldn't be allowed either
fn<User> = userMock
```

### 🙁 Actual behavior

Those instantiation expressions are allowed

### 🙂 Expected behavior

They shouldn't be

it has been "discovered" here: https://x.com/kettanaito/status/1820456054724993191 😅

jakebailey:
#50820 added `ExpressionWithTypeArguments` to `isOuterExpression` but treated it as an assertion, which then caused the checker to skip past it in reference checking. Split that out so we can skip past it in transforms like we did previously, which feels better because an instantiation is not really an assertion.

Also, while here, fix the fact that other `OuterExpression` related code was not `ExpressionWithTypeArguments` either.

Fixes #59535

### DIFF
--- a/src/compiler/factory/nodeFactory.ts
+++ b/src/compiler/factory/nodeFactory.ts
@@ -6553,6 +6553,8 @@ export function createNodeFactory(flags: NodeFactoryFlags, baseFactory: BaseNode
                 return updateSatisfiesExpression(outerExpression, expression, outerExpression.type);
             case SyntaxKind.NonNullExpression:
                 return updateNonNullExpression(outerExpression, expression);
+            case SyntaxKind.ExpressionWithTypeArguments:
+                return updateExpressionWithTypeArguments(outerExpression, expression, outerExpression.typeArguments);
             case SyntaxKind.PartiallyEmittedExpression:
                 return updatePartiallyEmittedExpression(outerExpression, expression);
         }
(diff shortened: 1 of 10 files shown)


jakebailey:
@typescript-bot test it

src/compiler/factory/nodeFactory.ts
```diff
@@ -6553,6 +6553,8 @@ export function createNodeFactory(flags: NodeFactoryFlags, baseFactory: BaseNode
                 return updateSatisfiesExpression(outerExpression, expression, outerExpression.type);
             case SyntaxKind.NonNullExpression:
                 return updateNonNullExpression(outerExpression, expression);
+            case SyntaxKind.ExpressionWithTypeArguments:
+                return updateExpressionWithTypeArguments(outerExpression, expression, outerExpression.typeArguments);
```
jakebailey, line 6557: I will note that most of the cases here are uncovered by tests: 
image

weswigham (approved):

PR #59542, opened 2024-08-07
Fixing delay caused in vscode due to pasteEdits

navya9singh:
This is a wip, to get some feedback on how the fix looks like.


src/services/pasteEdits.ts
```diff
@@ -1,4 +1,6 @@
-import { findIndex } from "../compiler/core.js";
```
jakebailey, line 1: Woah, is this line in main? We should fix this, this kind of import is most likely broken...

jakebailey: There are others wrong than just this one, working on a PR quick...

src/services/pasteEdits.ts
```diff
@@ -106,12 +108,34 @@ function pasteEdits(
...
-                }
-                node.forEachChild(cb);
+
+            /**
+             * `updatedRanges` represent the new ranges that account for the offset changes caused by pasting new text and
+             * `offset` represents by how much the starting position of `pasteLocations` needs to be changed.
+             *
+             * We iterate over each updated range to get the node that wholly encloses the updated range. For each child of that node, it is checked if the
+             * identifier lies within the updated range and if it is not resolved, we try resolving it.
+             */
```
DanielRosenwasser, line 118: Since this comment isn't strictly about `updatedRanges`, I'd move away from JSDoc.

```suggestion
             // `updatedRanges` represents the new ranges that account for the offset changes caused by pasting new text and
             // `offset` represents how much the starting position of each item in `pasteLocations` needs to be changed.
             //
             // We iterate over each updated range to get the node that wholly encloses the updated range. For each child of that node, it is checked if the
             // identifier lies within the updated range and if it is not resolved, we try resolving it.
```

src/services/pasteEdits.ts
```diff
@@ -106,12 +108,34 @@ function pasteEdits(
...
+                const startPos = location.pos - offset;
+                const endPos = startPos + textToBePasted.length - 1;
+                updatedRanges.push({ pos: startPos, end: endPos });
+                offset += deletionNeeded - textToBePasted.length;
+            });
+
+            updatedRanges.forEach(range => {
+                const enclosingNode = findAncestor(getTokenAtPosition(context.sourceFile, range.pos), cb => rangeContainsPosition({ pos: cb.pos, end: cb.end }, range.end));
+                if (!enclosingNode) return;
+                forEachChild(enclosingNode, function cb(node) {
```
DanielRosenwasser, line 133: I'd try to give `cb` a more descriptive name if you can think of one that conveys the information.

src/services/pasteEdits.ts
```diff
@@ -106,12 +108,34 @@ function pasteEdits(
...
+                    if (isIdentifier(node) && rangeContainsPosition(range, node.getStart()) && !originalProgram?.getTypeChecker().resolveName(node.text, node, SymbolFlags.All, /*excludeGlobals*/ false)) {
```
DanielRosenwasser, line 134: If you really wanted `node.getStart`, pass in `sourceFile` here.

DanielRosenwasser:
```suggestion
                    if (isIdentifier(node) && rangeContainsPosition(range, node.getStart(sourceFile)) && !originalProgram?.getTypeChecker().resolveName(node.text, node, SymbolFlags.All, /*excludeGlobals*/ false)) {
```

src/services/pasteEdits.ts
```diff
@@ -106,12 +108,34 @@ function pasteEdits(
...
+                const deletionNeeded = location.pos === location.end ? 0 : location.end - location.pos + 1;
+                const textToBePasted = actualPastedText ? actualPastedText[0] : pastedText[i];
+                const startPos = location.pos - offset;
+                const endPos = startPos + textToBePasted.length - 1;
+                updatedRanges.push({ pos: startPos, end: endPos });
+                offset += deletionNeeded - textToBePasted.length;
+            });
+
+            updatedRanges.forEach(range => {
+                const enclosingNode = findAncestor(getTokenAtPosition(context.sourceFile, range.pos), cb => rangeContainsPosition({ pos: cb.pos, end: cb.end }, range.end));
```
DanielRosenwasser, line 131: I think this would be better to write and avoids an allocation.

```suggestion
                const enclosingNode = findAncestor(getTokenAtPosition(context.sourceFile, range.pos), ancestorNode => rangeContainsPosition(ancestorNode, range.end));

```

src/services/pasteEdits.ts
```diff
@@ -106,12 +108,34 @@ function pasteEdits(
...
+             * `updatedRanges` represent the new ranges that account for the offset changes caused by pasting new text and
+             * `offset` represents by how much the starting position of `pasteLocations` needs to be changed.
+             *
+             * We iterate over each updated range to get the node that wholly encloses the updated range. For each child of that node, it is checked if the
+             * identifier lies within the updated range and if it is not resolved, we try resolving it.
+             */
+            const updatedRanges: TextRange[] = [];
+            let offset = 0;
+            pasteLocations.forEach((location, i) => {
+                const deletionNeeded = location.pos === location.end ? 0 : location.end - location.pos + 1;
```
DanielRosenwasser, line 122: What's the purpose of the `+ 1` here and `- 1` that you're doing to adjust for the `endPos`?

src/services/pasteEdits.ts
jakebailey: https://github.com/microsoft/TypeScript/pull/59544

src/services/pasteEdits.ts
andrewbranch: I also don’t understand the purpose of this. Why would we delete 2 characters when 1 is selected to be pasted over?

navya9singh: Actually, I checked it again and it does need a +1 and -1. For example, if the text that would be replaced is `apple` with `{pos: 0, end: 4}` and the text replacing it is `range`, then the deletion needed should be 5, and not 4 (as 5 characters will be deleted), and the new position for `range` should again come out to be `{pos: 0, end: 4}` and not `{pos:0, end:5}`

DanielRosenwasser (changes requested):
Create some tests around the specific questions I mentioned if they don't already exist.

Also, I think you need a few more tests for when the paste replacement text items are (compared to their target ranges):

1. consistently equal in size
2. consistently smaller in size
3. consistently larger in size
4. vary by growing in size
5. vary by shrinking in size

src/services/pasteEdits.ts
andrewbranch: > if the text that would be replaced is `apple` with `{pos: 0, end: 4}`

This is a contradiction. If `apple` is selected at the beginning of the file, the location sent should be `{ pos: 0, end: 5 }`. If the location is `{ pos: 0, end: 4 }`, then the text that should be replaced is `appl`.

DanielRosenwasser: It might help to remember that the case of a 0-length range is actually not special.

Like Andrew said, `{ pos: 0, end: 1 }` is a paste with 1 character selected, and replaces 1 character.

Similarly, `{ pos: 0, end: 0 }` is a paste with no characters selected. It replaces 0 characters.

If you're finding that you need to adjust for some reason (and it's somehow working) then it's really important to know why and create a counter-example test where the current logic fails.

navya9singh: You're right, I misunderstood the `pos` and `end` values. With the right values, there is no need for +1 and -1 anywhere.

src/services/pasteEdits.ts
```diff
@@ -104,12 +106,33 @@ function pasteEdits(
...
+                const deletionNeeded = location.pos === location.end ? 0 : location.end - location.pos;
+                const textToBePasted = actualPastedText ? actualPastedText[0] : pastedText[i];
+                const startPos = location.pos - offset;
+                const endPos = startPos + textToBePasted.length;
+                updatedRanges.push({ pos: startPos, end: endPos });
+                offset += deletionNeeded - textToBePasted.length;
+            });
+
+            updatedRanges.forEach(range => {
+                const enclosingNode = findAncestor(getTokenAtPosition(context.sourceFile, range.pos), ancestorNode => rangeContainsPosition(ancestorNode, range.end));
```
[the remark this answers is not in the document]
navya9singh: Yes, it works in that case. For example, in test pasteEdits_multiplePastesGrowingInSize.ts

src/services/pasteEdits.ts
```diff
@@ -104,12 +106,33 @@ function pasteEdits(
...
+
+            const updatedRanges: TextRange[] = [];
+            let offset = 0;
+            pasteLocations.forEach((location, i) => {
+                const deletionNeeded = location.pos === location.end ? 0 : location.end - location.pos;
+                const textToBePasted = actualPastedText ? actualPastedText[0] : pastedText[i];
+                const startPos = location.pos - offset;
+                const endPos = startPos + textToBePasted.length;
+                updatedRanges.push({ pos: startPos, end: endPos });
+                offset += deletionNeeded - textToBePasted.length;
```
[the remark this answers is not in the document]
navya9singh: There wasn't a very specific reason to make `offset` shrink for longer content, but yeah, making it larger for longer content and smaller for shorter makes it easier to understand. I've reversed the order here now for calculating `offset` and `startPos`.

src/services/pasteEdits.ts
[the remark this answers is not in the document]
DanielRosenwasser: Okay, I see that `allowPositionInLeadingTrivia` is set by `getTokenAtPosition` (but not in `getTouchingToken`).

src/services/pasteEdits.ts
```diff
@@ -104,12 +106,32 @@ function pasteEdits(
...
+                const oldTextLength = location.end - location.pos;
+                const textToBePasted = actualPastedText ? actualPastedText[0] : pastedText[i];
+                const startPos = location.pos + offset;
+                const endPos = startPos + textToBePasted.length;
+                updatedRanges.push({ pos: startPos, end: endPos });
+                offset += textToBePasted.length - oldTextLength;
+            });
+
+            updatedRanges.forEach(range => {
+                const enclosingNode = findAncestor(getTokenAtPosition(context.sourceFile, range.pos), ancestorNode => rangeContainsPosition(ancestorNode, range.end));
```
DanielRosenwasser, line 127: Aside: I almost wanted to suggest `rangeContainsRange` but I guess this just works because you've found the token at the start position, but you're trying to find the parent that also contains the end.

src/services/pasteEdits.ts
```diff
@@ -104,12 +106,32 @@ function pasteEdits(
...
+                if (!enclosingNode) return;
+                forEachChild(enclosingNode, function cb(node) {
+                    if (isIdentifier(node) && rangeContainsPosition(range, node.getStart(updatedFile)) && !originalProgram?.getTypeChecker().resolveName(node.text, node, SymbolFlags.All, /*excludeGlobals*/ false)) {
+                        importAdder.addImportForUnresolvedIdentifier(context, node, /*useAutoImportProvider*/ true);
+                    }
+                    node.forEachChild(cb);
```
DanielRosenwasser, line 133:
```suggestion
                forEachChild(enclosingNode, function importUnresolvedIdentifiers(node) {
                    if (isIdentifier(node) && rangeContainsPosition(range, node.getStart(updatedFile)) && !originalProgram?.getTypeChecker().resolveName(node.text, node, SymbolFlags.All, /*excludeGlobals*/ false)) {
                        importAdder.addImportForUnresolvedIdentifier(context, node, /*useAutoImportProvider*/ true);
                        return;
                    }
                    node.forEachChild(importUnresolvedIdentifiers);
```

tests/cases/fourslash/server/pasteEdits_multiplePastesConsistentlyLargerInSize.ts
```diff
@@ -0,0 +1,78 @@
...
+
+const range = test.ranges();
+verify.pasteEdits({
+    args: {
+        pastedText: [ `const t = 1 + juice + p;`,`function avacado() { return sauce; }`,
+            `fig + kiwi`,
+`function k() {
+            const cherry = 3 + tomato + cucumber;
+        }`
+],
```
DanielRosenwasser, line 46:
```suggestion
        pastedText: [
            `const t = 1 + juice + p;`,`function avacado() { return sauce; }`,
            `fig + kiwi`,
            `function k() {
            const cherry = 3 + tomato + cucumber;
        }`
    ],
```

tests/cases/fourslash/server/pasteEdits_multiplePastesConsistentlyLargerInSize.ts
```diff
@@ -0,0 +1,78 @@
...
+        pasteLocations: [range[0], range[1], range[2], range[3]],
+    },
+    newFileContents: {
```
DanielRosenwasser, line 49: Is there no way to do baselining instead? I guess this is fine

tests/cases/fourslash/server/pasteEdits_multiplePastesConsistentlyLargerInSize.ts
```diff
@@ -0,0 +1,78 @@
...
+const range = test.ranges();
+verify.pasteEdits({
+    args: {
+        pastedText: [ `const t = 1 + juice + p;`,`function avacado() { return sauce; }`,
+            `fig + kiwi`,
+`function k() {
+            const cherry = 3 + tomato + cucumber;
+        }`
+],
+        pasteLocations: [range[0], range[1], range[2], range[3]],
```
DanielRosenwasser, line 47:
```suggestion
        pasteLocations: test.ranges(),
```

?

DanielRosenwasser (changes requested):
Looking a lot better! I left some sugestions. I'd look over the other tests to see if there's similar feedback, but there's a cat on my right arm.

src/services/pasteEdits.ts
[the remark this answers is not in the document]
andrewbranch: I think `offset` is negative what it should be and `startPos` should be `location.pos + offset`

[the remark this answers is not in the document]
navya9singh: I don't think I understand that

src/services/pasteEdits.ts
```diff
@@ -104,12 +107,33 @@ function pasteEdits(
...
+            const updatedRanges: TextRange[] = [];
+            let offset = 0;
+            pasteLocations.forEach((location, i) => {
+                const oldTextLength = location.end - location.pos;
+                const textToBePasted = actualPastedText ? actualPastedText[0] : pastedText[i];
+                const startPos = location.pos + offset;
+                const endPos = startPos + textToBePasted.length;
+                updatedRanges.push({ pos: startPos, end: endPos });
+                offset += textToBePasted.length - oldTextLength;
+            });
```
sheetalkamat, line 125: Why is it that this need to be separate pass and not inline where we are iterating over nodes in the the updated range?

src/services/pasteEdits.ts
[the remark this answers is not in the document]
andrewbranch:
```ts
const startPos = location.pos + offset;
const endPos = startPos + textToBePasted.length;
updatedRanges.push({ pos: startPos, end: endPos });
offset += textToBePasted.length - deletionNeeded;
```

This is mathematically equivalent but I agree with Daniel that it’s less brain-breaking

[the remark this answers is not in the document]
navya9singh: Ah, yes, made that change👍

src/services/pasteEdits.ts
navya9singh: In the updatedFile, we don't know the exact range of where the pasted text exists, so doing this pass helps us get the right location of the pasted text. Then, it becomes easy to just go over the updated ranges and check for unresolved nodes. Is that what you were asking about?

sheetalkamat: Isnt it same as :
```ts
let offset = 0;
            pasteLocations.forEach((location, i) => {
                const oldTextLength = location.end - location.pos;
                const textToBePasted = actualPastedText ? actualPastedText[0] : pastedText[i];
                const startPos = location.pos + offset;
                const endPos = startPos + textToBePasted.length;
                const range = { pos: startPos, end: endPos } satisfies TextRange;
                offset += textToBePasted.length - oldTextLength;
                const enclosingNode = findAncestor(
                    getTokenAtPosition(context.sourceFile, range.pos),
                    ancestorNode => rangeContainsRange(ancestorNode, range),
                );
                if (!enclosingNode) return;
                forEachChild(enclosingNode, function importUnresolvedIdentifiers(node) {
                    if (
                        isIdentifier(node) && rangeContainsPosition(
                            range,
                            node.getStart(updatedFile),
                        ) && !originalProgram?.getTypeChecker().resolveName(
                            node.text,
                            node,
                            SymbolFlags.All,
                            /*excludeGlobals*/ false,
                        )
                    ) {
                        return importAdder.addImportForUnresolvedIdentifier(
                            context,
                            node,
                            /*useAutoImportProvider*/ true,
                        );
                    }
                    node.forEachChild(importUnresolvedIdentifiers);
                });
            });
```

src/services/pasteEdits.ts
```diff
@@ -104,12 +107,45 @@ function pasteEdits(
...
+            // identifier lies within the updated range and if it is not resolved, we try resolving it.
+            let offset = 0;
+            pasteLocations.forEach((location, i) => {
+                const oldTextLength = location.end - location.pos;
+                const textToBePasted = actualPastedText ? actualPastedText[0] : pastedText[i];
+                const startPos = location.pos + offset;
+                const endPos = startPos + textToBePasted.length;
+                const range = { pos: startPos, end: endPos } satisfies TextRange;
+                offset += textToBePasted.length - oldTextLength;
+                const enclosingNode = findAncestor(
```
DanielRosenwasser, line 124:
```suggestion

                const enclosingNode = findAncestor(
```

src/services/pasteEdits.ts
```diff
@@ -104,12 +107,45 @@ function pasteEdits(
...
+            //
+            // We iterate over each updated range to get the node that wholly encloses the updated range. For each child of that node, it is checked if the
+            // identifier lies within the updated range and if it is not resolved, we try resolving it.
+            let offset = 0;
+            pasteLocations.forEach((location, i) => {
+                const oldTextLength = location.end - location.pos;
+                const textToBePasted = actualPastedText ? actualPastedText[0] : pastedText[i];
+                const startPos = location.pos + offset;
+                const endPos = startPos + textToBePasted.length;
+                const range = { pos: startPos, end: endPos } satisfies TextRange;
```
DanielRosenwasser, line 122:
```suggestion
                const range: TextRange = { pos: startPos, end: endPos };
```

src/services/pasteEdits.ts
```diff
@@ -104,12 +107,45 @@ function pasteEdits(
...
+                offset += textToBePasted.length - oldTextLength;
+                const enclosingNode = findAncestor(
+                    getTokenAtPosition(context.sourceFile, range.pos),
+                    ancestorNode => rangeContainsRange(ancestorNode, range),
+                );
+                if (!enclosingNode) return;
+                forEachChild(enclosingNode, function importUnresolvedIdentifiers(node) {
```
DanielRosenwasser, line 129:
```suggestion

                forEachChild(enclosingNode, function importUnresolvedIdentifiers(node) {
```

src/services/pasteEdits.ts
```diff
@@ -104,12 +107,45 @@ function pasteEdits(
...
+                    if (
+                        isIdentifier(node) && rangeContainsPosition(
+                            range,
+                            node.getStart(updatedFile),
```
DanielRosenwasser, line 133: Do you need to run `node.getStart()`? Can you just use `node.pos`?

src/services/pasteEdits.ts
```diff
@@ -104,12 +107,45 @@ function pasteEdits(
...
+                const range = { pos: startPos, end: endPos } satisfies TextRange;
+                offset += textToBePasted.length - oldTextLength;
+                const enclosingNode = findAncestor(
+                    getTokenAtPosition(context.sourceFile, range.pos),
+                    ancestorNode => rangeContainsRange(ancestorNode, range),
+                );
+                if (!enclosingNode) return;
+                forEachChild(enclosingNode, function importUnresolvedIdentifiers(node) {
+                    if (
+                        isIdentifier(node) && rangeContainsPosition(
```
DanielRosenwasser, line 131: I would just hoist the entire expression out of the `if` with trailing `&&`s, and break each call into a single line if possible.

```ts
const isImportCandidate =
    isIdentifier(node) &&
    rangeContainsPosition(...) &&
    ...
```

src/services/pasteEdits.ts
```diff
@@ -104,12 +107,45 @@ function pasteEdits(
...
-                    // generate imports
-                    importAdder.addImportForUnresolvedIdentifier(context, node, /*useAutoImportProvider*/ true);
-                }
-                node.forEachChild(cb);
+
+            // `updatedRanges` represent the new ranges that account for the offset changes caused by pasting new text and
+            // `offset` represents by how much the starting position of `pasteLocations` needs to be changed.
+            //
+            // We iterate over each updated range to get the node that wholly encloses the updated range. For each child of that node, it is checked if the
+            // identifier lies within the updated range and if it is not resolved, we try resolving it.
```
DanielRosenwasser, line 115:
```suggestion
            // We iterate over each updated range to get the node that wholly encloses the updated range.
            // For each child of that node, we checked for unresolved identifiers
            // within the updated range and try importing it.
```

tests/cases/fourslash/server/pasteEdits_multiplePastesEqualInSize.ts
```diff
@@ -0,0 +1,76 @@
...
+
+verify.pasteEdits({
+    args: {
+        pastedText: [ 
+            `console.log(juice);`,`function kl() { return sauce; }`,
+            `apple`,
+            `function k() {
+       const cherry = 3 + tomato + cucumber;
+        }`
+],
```
DanielRosenwasser, line 45:
```suggestion
        ],
```

tests/cases/fourslash/server/pasteEdits_multiplePastesGrowingInSize.ts
```diff
@@ -0,0 +1,77 @@
...
+
+verify.pasteEdits({
+    args: {
+        pastedText: [ 
+            `const t = figs;`,`apples : number`,
+            `             console.log(sauce + tomato); `,
+            `//function k(i:string) {
+         const cherry = 3 + juices + cucumber;
+//      }`
+],
```
DanielRosenwasser, line 46:
```suggestion
        ],
```

src/services/pasteEdits.ts
```diff
@@ -104,12 +107,45 @@ function pasteEdits(
...
+            let offset = 0;
+            pasteLocations.forEach((location, i) => {
+                const oldTextLength = location.end - location.pos;
+                const textToBePasted = actualPastedText ? actualPastedText[0] : pastedText[i];
```
DanielRosenwasser, line 119: Is it ever possible that the count of paste locations exceeds the number of items in `pastedText`?

DanielRosenwasser (changes requested):

src/services/pasteEdits.ts
sheetalkamat: This is done at top of the function to make sure its always right

sheetalkamat (approved):
Apart from daniel's code formatting comments, looks good.

src/services/pasteEdits.ts
navya9singh: Yes, `node.pos` gives the position with the whitespaces. So, for example in pasteEdits_multiplePastesConsistentlyLargerInSize.ts for the 3rd paste location (`const test = [|1 + 2|] + 3;` ) , `node.pos` will give the position of the whitespace before 1, but the updated range of the pasted text starts from 1. As `node.pos` will not lie in this range, it will skip checking for the pasted text here. `node.Start()` avoids this by giving the exact position of 1

src/services/pasteEdits.ts
navya9singh: Yes, in that case all `pastedText` is joined together with a newline character and that is pasted at all locations. Line 64 does this.

src/services/pasteEdits.ts
```diff
@@ -104,12 +107,46 @@ function pasteEdits(
...
+                const enclosingNode = findAncestor(
+                    getTokenAtPosition(context.sourceFile, range.pos),
+                    ancestorNode => rangeContainsRange(ancestorNode, range),
+                );
+                if (!enclosingNode) return;
+
+                forEachChild(enclosingNode, function importUnresolvedIdentifiers(node) {
+                    const isImportCandidate = isIdentifier(node) &&
+                        rangeContainsPosition(range, node.getStart(updatedFile)) &&
+                        !originalProgram?.getTypeChecker().resolveName(
```
andrewbranch, line 135: Shouldn’t this use the updated program? For example, if I have a single paste of

```ts
const b = 0;
console.log(b);
```

won’t this try to resolve `b` against the original text which didn’t have a declaration for `b`, so we’ll try to auto-import it unnecessarily?

andrewbranch (changes requested):

src/services/pasteEdits.ts
sheetalkamat: I didnt see this before, but originalProgram's typechecker usage is sketchy as well.. The reuse of sourceFile and parent pointer changing makes it non-viable to use.

andrewbranch: Can you add a test that fails when using `originalProgram`?

src/services/pasteEdits.ts
DanielRosenwasser: Gotcha - while you're here, can you fix that so it uses the appropriate newline character instead of `\n`?

DanielRosenwasser: Also, `actualPastedText` doesn't have to be an array if you only ever use it for `actualPastedText[0]`. Just use `actualPastedText ?? pastedText[i]`.

navya9singh: Do you mean using `getNewLineOrDefaultFromHost()` for the appropriate newline character?

andrewbranch (approved):

tests/cases/fourslash/server/pasteEdits_noImportNeededInUpdatedProgram.ts
```diff
@@ -0,0 +1,21 @@
...
+// @Filename: /tsconfig.json
+////{ "files": ["a.ts", "b.ts"] }
+
+const range = test.ranges();
+verify.pasteEdits({
+    args: {
+        pastedText: [
+            `const b = 1;
+console.log(b);`],
+        pasteLocations: [range[0]],
```
DanielRosenwasser, line 18: Two things

1. Make this `test.ranges()` like everywhere else.
2. In the future, if you have an array, use the plural version (`ranges` instead of `range`).

DanielRosenwasser (approved):

DanielRosenwasser:

PR #59543, opened 2024-08-07
Add `autoImportSpecifierExcludeRegexes` preference

### LINKED ISSUE #55092: Allow ambient module names to be filtered out of auto-imports or something
# Suggestion

## 🔍 Search Terms

autoImportFileExcludePatterns, exclude auto import suggestions 

## ✅ Viability Checklist

* [X] This wouldn't be a breaking change in existing TypeScript/JavaScript code
* [X] This wouldn't change the runtime behavior of existing JavaScript code
* [X] This could be implemented without emitting different JS based on the types of the expressions
* [X] This isn't a runtime feature (e.g. library functionality, non-ECMAScript syntax with JavaScript output, new syntax sugar for JS, etc.)
* [X] This feature would agree with the rest of [TypeScript's Design Goals](https://github.com/Microsoft/TypeScript/wiki/TypeScript-Design-Goals).

## ⭐ Suggestion

The option `autoImportFileExcludePatterns` currently works on the package/module (?) level. I can prevent a whole package from being suggested in auto-imports, but I can't exclude only some of the imports in certain subdirectories of that package. This was already mentioned as a caveat in the original implementation of the option (https://github.com/microsoft/TypeScript/pull/49578#issuecomment-1200091246).

## 📃 Motivating Example

As in the comment mentioned above, my issue also comes from `date-fns`, but for a slightly different reason. The package has a `locale` module that includes exports for many different two-letter country codes. Those names also happen to conflict with variable names quite often. When typing one of the country codes, I get tons of suggestions from `date-fns` which appear before the one I actually want:

image

It's a minor annoyance, but I'd like to exclude those suggestions.

## 💻 Use Cases

Currently, I can add `"**/node_modules/date-fns"` or `"date-fns"` to the `autoImportFileExcludePatterns` option, which completely removes `date-fns` suggestions. But then I also lose the suggestions I actually want from the library.

I tried adding `"date-fns/locale"`, `"date-fns/locale/**"`, `"date-fns/locale/*.*"`, but none of those work and result in no suggestions being excluded at all.

andrewbranch:
Closes #55092 

This has the limitation that it doesn't compose very intelligently with the `importModuleSpecifierEnding` preference&mdash;you can set that to `"js"` and then write a regex that excludes `\.js$` and the result will be that you get no (relative) auto imports at all. But I think I'd like to try this and get feedback.

### DIFF
--- a/src/compiler/moduleSpecifiers.ts
+++ b/src/compiler/moduleSpecifiers.ts
@@ -130,7 +130,26 @@ import {
 
 const stringToRegex = memoizeOne((pattern: string) => {
     try {
-        return new RegExp(pattern);
+        let slash = pattern.indexOf("/");
+        if (slash !== 0) {
+            // No leading slash, treat as a pattern
+            return new RegExp(pattern);
+        }
+        const lastSlash = pattern.lastIndexOf("/");
+        if (slash === lastSlash) {
+            // Only one slash, treat as a pattern
+            return new RegExp(pattern);
+        }
+        while ((slash = pattern.indexOf("/", slash + 1)) !== lastSlash) {
+            if (pattern[slash - 1] !== "\\") {
+                // Unescaped middle slash, treat as a pattern
+                return new RegExp(pattern);
+            }
+        }
+        // Only case-insensitive and unicode flags make sense
+        const flags = pattern.substring(lastSlash + 1).replace(/[^iu]/g, "");
+        pattern = pattern.substring(1, lastSlash);
+        return new RegExp(pattern, flags);
     }
     catch {
         return undefined;
--- a/src/compiler/moduleSpecifiers.ts
+++ b/src/compiler/moduleSpecifiers.ts
@@ -83,6 +83,7 @@ import {
     mapDefined,
     MapLike,
     matchPatternOrExact,
+    memoizeOne,
     min,
     ModuleDeclaration,
     ModuleKind,
@@ -127,6 +128,13 @@ import {
     UserPreferences,
 } from "./_namespaces/ts.js";
 
+const stringToRegex = memoizeOne((pattern: string) => {
+    try {
+        return new RegExp(pattern);
+    }
+    catch {}
+});
+
 // Used by importFixes, getEditsForFileRename, and declaration emit to synthesize import module specifiers.
 
 /** @internal */
@@ -144,11 +152,12 @@ export interface ModuleSpecifierPreferences {
      * @param syntaxImpliedNodeFormat Used when the import syntax implies ESM or CJS irrespective of the mode of the file.
      */
     getAllowedEndingsInPreferredOrder(syntaxImpliedNodeFormat?: ResolutionMode): ModuleSpecifierEnding[];
+    readonly excludeRegexes?: readonly string[];
 }
 
 /** @internal */
 export function getModuleSpecifierPreferences(
-    { importModuleSpecifierPreference, importModuleSpecifierEnding }: UserPreferences,
+    { importModuleSpecifierPreference, importModuleSpecifierEnding, autoImportSpecifierExcludeRegexes }: UserPreferences,
     host: Pick<ModuleSpecifierResolutionHost, "getDefaultResolutionModeForFile">,
     compilerOptions: CompilerOptions,
     importingSourceFile: Pick<SourceFile, "fileName" | "impliedNodeFormat">,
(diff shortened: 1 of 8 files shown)


src/compiler/moduleSpecifiers.ts
```diff
@@ -512,7 +532,11 @@ function computeModuleSpecifiers(
     return pathsSpecifiers?.length ? { kind: "paths", moduleSpecifiers: pathsSpecifiers, computedWithoutCache: true } :
         redirectPathsSpecifiers?.length ? { kind: "redirect", moduleSpecifiers: redirectPathsSpecifiers, computedWithoutCache: true } :
         nodeModulesSpecifiers?.length ? { kind: "node_modules", moduleSpecifiers: nodeModulesSpecifiers, computedWithoutCache: true } :
-        { kind: "relative", moduleSpecifiers: Debug.checkDefined(relativeSpecifiers), computedWithoutCache: true };
+        { kind: "relative", moduleSpecifiers: relativeSpecifiers ?? emptyArray, computedWithoutCache: true };
+}
+
+function isExcludedByRegex(moduleSpecifier: string, excludeRegexes: readonly string[] | undefined): boolean {
+    return some(excludeRegexes, pattern => !!stringToRegex(pattern)?.test(moduleSpecifier));
```
sheetalkamat, line 539: wont this keep memoized map forever in tsserver - think you would want to memoize this and keep lifetime same as excludeRegex or something like that ?

andrewbranch: Yes, but I think normally the preference won’t change (or won’t change much) during the life of the server. It might be nice not to have to redo this on every request. Theoretically this could waste memory, but it will only have an impact if the user is changing between thousands of unique strings...

sheetalkamat (approved):

src/compiler/moduleSpecifiers.ts
```diff
@@ -127,6 +128,15 @@ import {
     UserPreferences,
 } from "./_namespaces/ts.js";
 
+const stringToRegex = memoizeOne((pattern: string) => {
+    try {
+        return new RegExp(pattern);
```
jakebailey, line 133: Do we have any other settings which are controlled by a regex like this? Trying to decide if people expect to be able to set it to a literal like `"/foo/i"` or something, but maybe that's a bit much.

That and, should this use `getRegexFromPattern` and handle case insensitive paths?

andrewbranch: We don’t, but C Spell does seem to allow slashes and flags. I’m not opposed to implementing that. Without the ability to provide flags, I figured case sensitive was the best option.

jakebailey: Yeah, I think people can already do like `(?:i)` at the start of their patterns too?

I don't know how stable this option is or if you're just experimenting, though. That and I sort of bet people will want to write patterns that start with `/`...

gabritto (approved):

sandersn (approved):
Looks good, with the caveat that I don't know this area very well.

One question: since `autoImportFileExcludePatterns` already exists, do these preferences always show up in a way that makes people aware that both exist? Eg in a graphical config UI next to each other, or in completions for the prefix "autoImport|". If that's not the case, I'd be worried that people would discover one but not the other.

andrewbranch:
@sandersn yes, they will, after the corresponding VS Code PR.

src/compiler/moduleSpecifiers.ts
mjbvz: `(?:i)` isn't supported in js yet afaik so maybe better to require the whole `/.../`. Requires a little extra parsing but should be more explicit that this is a regex too

jakebailey: Ah, yeah, I'm thinking of re2/Go syntax.

Other places  that take regexes are `git.branchValidationRegex` in vscode, which just passes it into `new RegExp` https://github.com/microsoft/vscode/blob/8f88d203d76dee8c92e491f8a1e1331d74792b42/extensions/git/src/commands.ts#L2679

Knagis:
not sure if it is worth opening separate issue as it might already been discussed.

Would it be an option to improve this feature to allow specifying for which files the exclusion apply?

In other words, as an example, while in `shared-package`, i want auto imports for all files, however, while in `another-package`, i only want to allow `shared-package/index.ts` and nothing else from there.

andrewbranch:
I’m assuming you’re using relative imports within `shared-package`, so you can just exclude `^shared-package/` everywhere.

PR #59544, opened 2024-08-07
Fix accidental relative imports from non namespace barrels

jakebailey:
Noticed in #59542; no file in our codebase should be bypassing the namespace barrels except those in `src/compiler` (in rare cases) or test code.

The only reason things didn't break is because these imports were done within services, which is only ever consumed by other entrypoints; if this had been done from say, `tsserver`, the magic import rewriting done in our build would have caused "double bundling" where we'd include the same code twice.

However, the fact that it _did_ work implies that we could get away with direct imports except in our entrypoints, which is great because I thought I had unintentionally made direct imports impossible by making our package smaller. Turns out, no, we still can do it for anything _without_ that magic build step.


andrewbranch (approved):

sheetalkamat (approved):
Would prefer if this was some kind of warning if this is not desired. If we would do direct imports that would be ideal.

jakebailey:
Yeah, I'll write a lint rule quick that prevents this; given it seems like we can actually do direct imports, but only in _some_ project, we'd need that lint rule anyway, just applied to a different set of packages.

sheetalkamat (approved):

Issue #11920, opened 2016-10-28
disallow comparing to null and undefined unless they are valid cases in strict null mode


aleksey-bykov:
```typescript
declare const value: number;
if (value !== null) { // <-- somehow valid, expected to be invalid, since number doesn't have null as a possible value
}
if (value !== '') { // <-- invalid as expected 
}
```

RyanCavanaugh:
We intentionally allow this for the sake of defensive programming (i.e. defending against missing inputs from non-TS code). If there's enough demand we could add a flag or something.

Spongman:
i would suggest to make type safety the default, and have interop require explicit casting.

DanielRosenwasser:
Internal implementation detail: right now we use the comparability relationship for type assertions (casting) - if we did this, we'd need to have a separate type relationship (or ad-hoc check) for allowing you to cast `undefined`/`null` to any other type.

wizarrc:
https://github.com/Microsoft/TypeScript/issues/17896#issuecomment-323403042 

> As a quick fix for now there is a lint rule: https://palantir.github.io/tslint/rules/strict-type-predicates/

This seems to work good enough for me right now, but in the long run I would like to see this baked into the language.

devoto13:
I would also like to have it implemented for the same reason as https://github.com/microsoft/TypeScript/issues/14764: vulnerable refactoring. This proposal is a bit more broad, then https://github.com/microsoft/TypeScript/issues/14764. Maybe they should be merged?

In the recent refactoring I've changed `public key: string | null;` to `public key: string;`. `null` was used to mark a special case and now it was decided to use particular string constant for that special case instead. Deep in the codebase there was a bunch of checks like:

```
if (myObject.key == null) {
  // perform some logic for the special case
}
```

This silently stopped working and took quite a bit of effort to debug. It was also surprising that this didn't fail type checking. Maybe do it as `strictNullComparisons` flag or something, so it is `true` by default under `strict` mode and then consumers willing to do defensive programming can opt out? This way people will get consistent (and expected) behaviour out of the box and those, who need to defend their code from JavaScript can do so as well 😄

jfirebaugh:
[...]

PR #59559, opened 2024-08-08
Don't special case nullable types in `isTypeEqualityComparableTo`

Andarist:
closes https://github.com/microsoft/TypeScript/issues/11920

Potential downsides of this PR:
- it makes it harder to program defensively against potentially missing browser APIs and similar. This isn't quite a new problem though (`if (window.matchMedia) {}` is already an error)
- it also makes it harder to consume potentially missing keys from index signature types outside of `noUncheckedIndexedAccess`. This can easily be mitigated by including `| undefined` in those index signatures - if the user wants to check against missing values then it makes sense to include those nullable types in the index signatures themselves

Advantages of this PR:
- more cohesive behavior around `===` that is easier to understand for people. The current logic is a special case and many people don't know that it is one
- it helps to catch legitimate issues as it's not unlikely to have logical errors because of this behavior

@RyanCavanaugh [mentions](https://github.com/microsoft/TypeScript/issues/11920#issuecomment-256961196):
> We intentionally allow this for the sake of defensive programming (i.e. defending against missing inputs from non-TS code). If there's enough demand we could add a flag or something.

I'm not sold, personally, on introducing a new flag to enable this behavior. Would it even be a `strict` flag? or one of the other opt-ins like `noUncheckedIndexedAccess`? 

@DanielRosenwasser [mentions](https://github.com/microsoft/TypeScript/issues/11920#issuecomment-266106279):
> Internal implementation detail: right now we use the comparability relationship for type assertions (casting) - if we did this, we'd need to have a separate type relationship (or ad-hoc check) for allowing you to cast undefined/null to any other type.

I'm also not sure if this is needed. Note that both comments are pretty old at this point and the TS landscape and user expectations changed a lot since then. 

I don't mind implementing any of the above. This PR is meant as a conversation starter - I'd expect the final decision about this will require a design meeting either way and I can only address the async feedback received based on that.

I also expect user tests to reveal this to be quite breaking and at the same time, it might be hard to judge if those are good breaks or not. I could categorize them and count them by those categories to help aid the final decision though. Regardless, it will be interesting to see how breaking it turns out to be in its current form.

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -39487,8 +39487,8 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         }
     }
 
-    function isTypeEqualityComparableTo(source: Type, target: Type) {
-        return (target.flags & TypeFlags.Nullable) !== 0 || isTypeComparableTo(source, target);
+    function isTypeEqualityComparableTo(source: Type, target: Type, strictEq: boolean) {
+        return !strictEq && (target.flags & TypeFlags.Nullable) !== 0 || isTypeComparableTo(source, target);
     }
 
     function createCheckBinaryExpression() {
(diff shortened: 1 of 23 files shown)


kof:
Really excited about this one. This has been costing us real bugs uncaught by TS. When we checked undefined but didn't check for null and undefined was literally not even a potential value.

Andarist:
heh, at least it has been proven to be *quite breaking* 😅 so I guess now the question is if you'd be willing to pick this up as a new opt-in flag outside of the regular `strict` mode ;p

kof:
I briefly looked over the issues the bot has found and it seems to me that it found loads of code that shouldn't be there?

jakebailey:
The first link I opened led to https://github.com/Effect-TS/effect/pull/3435, so that's a good sign, but wow that's a lotta damage

jakebailey:
For javascript-obfuscator, since it was interesting:

- This is good, but I actually think this is something that our new syntactic-nullishness could test for? [Playground Link](https://www.typescriptlang.org/play/?ts=5.6.0-dev.20240808#code/GYVwdgxgLglg9mABDAzgOTgEwKYAo4BGAVttAFyIDyxpUiAZIgN6JQCeADtgPwUpQAnGGADmiAL4BKCoRLRkKZuOYAoROsQDsUEAKSzaDRgEID0AHTsuiYwF5bicDmDDsmANwrxKoA),  https://github.com/javascript-obfuscator/javascript-obfuscator/blob/828a190cf80a86227ef77be38e99aad9838aed70/src/node/NodeGuards.ts#L352
- This is obviously never null but probably a guard https://github.com/javascript-obfuscator/javascript-obfuscator/blob/828a190cf80a86227ef77be38e99aad9838aed70/src/node/NodeUtils.ts#L121
- These two are just because they don't use noUncheckedIndexAccess https://github.com/javascript-obfuscator/javascript-obfuscator/blob/828a190cf80a86227ef77be38e99aad9838aed70/src/utils/AbstractTransformerNamesGroupsBuilder.ts#L91 https://github.com/javascript-obfuscator/javascript-obfuscator/blob/828a190cf80a86227ef77be38e99aad9838aed70/src/utils/LevelledTopologicalSorter.ts#L49

Andarist:
> These two are just because they don't use noUncheckedIndexAccess

"just" doesn't give this credit 😅 most people don't use this (including the compiler's code ;p). This kind of break will be the most painful but it can usually be easily refactored to `typeof x === 'undefined'`

jakebailey:
Yeah "just" was doing a lot of work in that sentence 😅

Issue #59097, opened 2024-07-01
Investigate Adopting `--isolatedDeclarations` in TypeScript's Codebase


DanielRosenwasser:
We would like to enable the `isolatedDeclarations` flag on the TypeScript codebase. One reason for this is simple dog-fooding (i.e. making sure we understand what users are experiencing). Another is to enable faster builds, whether that be through parallelized declaration emit, or through handing declaration emit off to a faster build tool.

First thing's first though - enable the flag, see what this feels like, and understand the pain points.

escaton:
I came across this by accident, but have a suggestion to at least try `oxc-transform` for `d.ts` generation. I found it to be the fastest and closest to what tsc emits on my project (20k TS files).

PR #59562, opened 2024-08-08
update generated diagnostic messages for isolated declarations

iisaduan:
In investigating adopting isolated declarations (#59097), I found that the generated `Diagnostics` needs a type annotation. This updates the script so we're able to provide a type annotation for it. 

(this PR does not turn on the flag but once I make the PR that turns on the flag, I'll link it here: ____)

### DIFF
--- a/scripts/processDiagnosticMessages.mjs
+++ b/scripts/processDiagnosticMessages.mjs
@@ -93,20 +93,23 @@ function buildInfoFileOutput(messageTable, inputFilePathRel) {
         "}",
         "",
         "/** @internal */",
-        "export const Diagnostics = {",
+        "export const Diagnostics: Diagnostics = {",
     ];
+    const interfaceResult = ["\r\ninterface Diagnostics {"];
     messageTable.forEach(({ code, category, reportsUnnecessary, elidedInCompatabilityPyramid, reportsDeprecated }, name) => {
         const propName = convertPropertyName(name);
         const argReportsUnnecessary = reportsUnnecessary ? `, /*reportsUnnecessary*/ ${reportsUnnecessary}` : "";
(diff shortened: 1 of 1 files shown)


scripts/processDiagnosticMessages.mjs
```diff
@@ -93,20 +93,23 @@ function buildInfoFileOutput(messageTable, inputFilePathRel) {
...
 
+        interfaceResult.push(`    ${propName}: DiagnosticMessage;`);
         result.push(`    ${propName}: diag(${code}, DiagnosticCategory.${category}, "${createKey(propName, code)}", ${JSON.stringify(name)}${argReportsUnnecessary}${argElidedInCompatabilityPyramid}${argReportsDeprecated}),`);
     });
 
+    interfaceResult.push("}");
     result.push("};");
 
-    return result.join("\r\n");
+    return result.join("\r\n") + interfaceResult.join("\r\n");
```
jakebailey, line 112: Can you put the interface first? It doesn't matter in practice, but seems better to declare the type before using it.

scripts/processDiagnosticMessages.mjs
```diff
@@ -93,20 +93,23 @@ function buildInfoFileOutput(messageTable, inputFilePathRel) {
         "}",
         "",
         "/** @internal */",
-        "export const Diagnostics = {",
+        "export const Diagnostics: Diagnostics = {",
     ];
+    const interfaceResult = ["\r\ninterface Diagnostics {"];
```
jakebailey, line 98: This will have to be exported / internal.

jakebailey:
The duplication is unfortunate; I do wonder if there's a better way but I think this is the minimally invasive way.

The only other thing I can think of is like:

```ts
export const Diagnostics = {
    Foo: diag() as Diagnostic,
}
```

But not sure if ID accepts that.

iisaduan:
closed in https://github.com/microsoft/TypeScript/pull/59635

PR #59563, opened 2024-08-08
fix(59558): Implement Interface Quick Fix generates duplicate declarations

### LINKED ISSUE #59558: Implement Interface Quick Fix generates duplicate declarations
### 🔎 Search Terms

quick fix implement duplicate

TypeScript Version: 5.6.0-beta

### ⏯ Playground Link

https://github.com/Soarex16/ts-bug-disposable-impl

### 💻 Code

First, you should install `@types/node` and add `ESNext` in `compilerOptions/lib` in `tsconfig.json`. Then try to invoke 'Implement Interface' on the following class:
```ts
class Foo implements Disposable {

}
```

### 🙁 Actual behavior
```ts
class Foo implements Disposable {
    [Symbol.dispose](): void;
    [Symbol.dispose](): void;
    [Symbol.dispose](): void {
    }

}
```

### 🙂 Expected behavior
```ts
class Foo implements Disposable {
    [Symbol.dispose](): void {
    }

}
```

This happens because `Disposable` interface comes from both `@types/node` and `ESNext` lib.

a-tarasyuk:
Fixes #59558


src/services/codefixes/helpers.ts
```diff
@@ -300,6 +300,9 @@ export function addNewNodeForMemberSymbol(
             }
 
             for (const signature of signatures) {
+                if (signature.declaration && (signature.declaration.flags & NodeFlags.Ambient)) {
```
DanielRosenwasser, line 303: Is this correct? Does this work for the following two cases?

```ts
declare class Base {
  method(): void;
}

class Derived implements Base {
}
```


```ts
declare abstract class Base {
  abstract method(): void;
}

class Derived extends Base {
}
```

iisaduan (approved):

PR #59564, opened 2024-08-08
Fix auto import file extensions with package.json imports wildcards

### LINKED ISSUE #59553: Auto import with `.js` extension instead of `.ts` with `"allowImportingTsExtensions": true` and `package.json`, imports fields
### 🔎 Search Terms

"auto imports", "allowImportingTsExtensions", "package.json imports", ".js extension instead of .ts", "moduleResolution": "NodeNext"

TypeScript v5.5.4

### ⏯ Playground Link

_No response_

### 💻 Code

`tsconfig.json`:
```json
{
  "compilerOptions": {
    "module": "NodeNext",
    "moduleResolution": "NodeNext",
    "allowImportingTsExtensions": true
  }
}
```
`package.json`:
```json
{
  "type": "module",
  "imports": {
    "#*": "./src/*"
  }
}
```
`src/lib/add.ts` (not really important, the content of the file, path is more important):
```ts
export const add = (a: number, b: number): number => {
  return a + b
}
```
`src/index.ts`:
```ts
import { add } from "#lib/add.js" // WRONG AUTO IMPORT, using .js, instead of .ts
console.log(add(2, 3))
```

### 🙁 Actual behavior

Auto import with `#lib/add.js"`

### 🙂 Expected behavior

Auto import with `#lib/add.ts"` because I'm using the `"allowImportingTsExtensions": true` config option.

This is all the more relevant since the newly added CLI Flag `--experimental-strip-types` in Node.js, because it enforces imports to use the `.ts` extension.
Ref: https://nodejs.org/docs/latest/api/typescript.html

Related issue: https://github.com/microsoft/TypeScript/issues/52167

andrewbranch:
Fixes #59553

Also discovered and fixed a related bug where auto-imports could give you a `.d.ts` extension when resolving through `"imports"`.


tests/cases/fourslash/autoImportAllowImportingTsExtensionsPackageJsonImports1.ts
```diff
@@ -0,0 +1,44 @@
...
+
+// @Filename: /src/add.ts
+//// export function add(a: number, b: number) {}
+
+// @Filename: /src/index.ts
+//// add/*imports*/;
+//// external/*exports*/;
+
+verify.importFixModuleSpecifiers("imports", ["#add.ts"]);
+verify.importFixModuleSpecifiers("exports", ["pkg/external.js"]);
```
andrewbranch, line 41: Using `.ts` extensions where possible is limited to `"imports"`, and doesn’t take effect for `"exports"`. We may have to revisit this in the coming years if npm fills up with TS-only packages. We do not currently offer auto-imports for self-name imports (which are resolved through your own package.json `"exports"`), which is a place where it might make sense to respect `allowImportingTsExtensions`, so this may need to be revisited in the future as well.

tests/cases/fourslash/autoImportAllowImportingTsExtensionsPackageJsonImports2.ts
```diff
@@ -0,0 +1,34 @@
...
+//// }
+
+// @Filename: /src/add.ts
+//// export function add(a: number, b: number) {}
+
+// @Filename: /src/index.ts
+//// add/*imports*/;
+//// external/*exports*/;
+
+verify.importFixModuleSpecifiers("imports", ["#add.js"]);
```
andrewbranch, line 34: This was `"#add.d.ts"` before this PR 😱

src/compiler/moduleSpecifiers.ts
```diff
@@ -980,19 +1007,23 @@ function tryGetModuleNameFromExportsOrImports(options: CompilerOptions, host: Mo
                     return { moduleFileToTry: combinePaths(packageName, fragment) };
                 }
                 if (declarationFile && containsPath(pathOrPattern, declarationFile, ignoreCase)) {
-                    const fragment = getRelativePathFromDirectory(pathOrPattern, declarationFile, /*ignoreCase*/ false);
+                    const fragment = changeFullExtension(getRelativePathFromDirectory(pathOrPattern, declarationFile, /*ignoreCase*/ false), getJSExtensionForFile(declarationFile, options));
```
sheetalkamat, line 1010: shouldnt this be dependent of file case sensitvity?

andrewbranch: Sure looks like it!

andrewbranch: Either that, or the `containsPath` call should be hard-coded to `false`. `pathOrPattern` is a concatenation of a path and an `"exports"` string key, and Node’s resolution algorithm is case sensitive according to my reading. We may be doing this wrong all around.

sheetalkamat: paths and case-sensitivity .. ugh..

andrewbranch: Oh wait, it’s an `"exports"` value, not a key. Yeah, I think the whole thing should use the FS case sensitivity.

andrewbranch: Oh, this possible bug only occurs under `MatchingMode.Directory`, which was deprecated in Node.js 14 and EOL in 17, so I’m not going to touch this.

sheetalkamat: vfs is case insensitive by default. Dont you need to test it on "case sensitive" that its not offered or doesnt do some wierd case replacement.

andrewbranch: The test I added shows that a difference in case between the VFS and the paths written in package.json doesn’t prevent auto-imports from working, which is what’s most important IMO. Also, this PR doesn’t touch/change anything about case sensitivity. This `ignoreCase` argument was already there.

PR #59569, opened 2024-08-09
Account for right operands & fix a weird error message for leftmost nullish literals in checkNullishCoalesceOperands

### LINKED ISSUE #59546: "This binary expression is never nullish" errors due to the way I format my code
### 🔎 Search Terms

This binary expression is never nullish

- (Pretty confident that) this changed in commit or PR #59217 

### ⏯ Playground Link

https://www.typescriptlang.org/play/?ts=5.6.0-beta#code/KYOwrgtgBACghgJwC4HkBmAROBPKBvAWACgBIAOQEsBzACySQomCgF4oAmKAKi6gAYANMRJYA7iFYduvAIxDSAZTAgEFAM7M2nHh3kicDJpO28AzHqUgNSY9KgAWPRjBqA1rZ0BWecICCAG38sbENNKDEJAB8oS1UNKGjg0ISY5Ws9AP9KWnpGMOc3FOy6UIz-URw1SUykvJTM4tymeQBfYmIAE2AAY39EZm6AeysbKgRBsAAHAC4oACUewYQOgB54ZHRggShwCAAjYAQAbQBdU5TlLrQKEGAOgD4Abk6evoQB4bUbScRUTBxZus-sFnkRiEMRlBrFU2JdgNdbh1hAB6ZFQJAIXBjODgN5QH7IKCDNBQDo4KDXBBfYQAfhpUDGE0mRwJwJwJxRaKQNFAUG6PO67goJO5zGxuMQ+N+RJJZNw6igONwrJlO2oJTy2xxHSgwqhg3RmMVgTVOShP3Ealp9IAFEDNjgAHQNdVNZgAMilG3+uHpjKmR3tPudgUaoROUFmcIRdwAlJz0TyJPyekKRTyGQgcWA8SriaTyQqlV6bPm5aVFSAdXq1AaMbg4Ca5ea4JbrVA7b8HdgQ0EDHVPazu1A-eMA0Hgr3akwI1Gq-CbnGE2hG-49nBBeiDauC7g1BarO3-cyJ06AhVsGoTqCgA

### 💻 Code
```ts
const sets = undefined
	// try granular part of day first
	?? group[partOfDay]
	// then check if the granular part of day is any part of nighttime, and if so try all night spawns
	?? (PartOfDay.AllNighttime & partOfDay ? group[PartOfDay.AllNighttime] : undefined)
	// then check if the granular part of day is any part of daytime, and if so try all day spawns
	?? (PartOfDay.AllDaytime & partOfDay ? group[PartOfDay.AllDaytime] : undefined)
	// fallback to all day spawns
	?? group[PartOfDay.Always];
```

### 🙁 Actual behavior
My codebases have a number of instances of things similar to the above — `undefined` is used to put the actual potential results each on their own line, all formatted the same way. However this no longer works in TS 5.6.0 due to it throwing an error. I think this is an over-eager error message.

### 🙂 Expected behavior

I should be able to format the code how I want to still

<details>
<summary>Other examples of code in my codebases broken by this</summary>

```ts
let plugs: PlugRaw[] = undefined
	?? (this.state ? refresh.plugs ?? [] : undefined)
	?? this.plugs?.slice()
	?? (!this.plugSetHash ? undefined : (await DestinyPlugSetDefinition.get(this.plugSetHash))?.reusablePlugItems)
	?? (this.plugSetHash ? undefined : (await DeepsightSocketExtendedDefinition.get(this.item?.definition.hash))?.sockets[this.index]?.rewardPlugItems)
	?? [];
```

```ts
Component.create()
	.classes.add(ItemTooltipSourceClasses.ActivityName)
	.text.set(undefined
		?? Display.name(lostSectorDisplay)
		?? Display.name(source.dropTable.displayProperties)
		?? Display.name(activity))
	.appendTo(activityComponent);
```

```ts
.setRefreshMethod(() => undefined
	?? (status === UiStatusType.Cut ? this.human?.getStatusLevel(StatusType.Bleeding) === BleedLevel.Minor : undefined)
	?? (status === StatusType.Bleeding ? this.human?.getStatusLevel(StatusType.Bleeding) === BleedLevel.Major : undefined)
	?? !!this.human?.hasStatus(status as StatusType))
```
</details>

Also, this still works for `true &&` conditions and `false ||` conditions — would be nice if it stayed consistent:
```ts
const inRange = true
	&& gatherTile.x - treasureRange <= treasure.x && gatherTile.x + treasureRange >= treasure.x
	&& gatherTile.y - treasureRange <= treasure.y && gatherTile.y + treasureRange >= treasure.y
	&& executor.z === map.position.z;
```

```ts
const needsAuthView = false
	|| (viewRequiredAuth === "required" && !ProfileManager.isAuthenticated())
	|| (viewRequiredAuth === "spy" && !ProfileManager.get());
```

ChiriVulpes:
Fixes #59546 

This PR:
1. Removes the error for nullish literals as the leftmost operand of an expression of all nullish coalescing operators.
2. Performs the same checks for right-side operands that are not the last operand in an expression of all nullish coalescing operators.

Examples:
```ts
declare let opt: number | undefined;

const p01 = null ?? 1; // fixed error message, says "always nullish" instead of "never nullish"
const p02 = (opt ? null : undefined) ?? 1; // previously error, still error
const p03 = 1 ?? 2; // previously error, still error

const p10 = opt ?? null ?? 1; // previously OK, now errors on `null` as "always nullish"
const p11 = opt ?? (opt ? null : undefined) ?? 1; // previously OK, now errors on `opt ? null : undefined` as "always nullish"
const p12 = opt ?? (opt ? 1 : 2) ?? 3; // previously OK, now errors on `opt ? 1 : 2` as "never nullish"
```

The tests in `predicateSemantics.ts` have more cases to verify this works in all situations, but it's possible I've missed some and this needs more coverage. 

For more information, see #59546 

Note: this overview has been edited to reflect the current PR


ChiriVulpes:

MartinJohns:
> const p01 = null ?? 1; // previously error, now OK

According to #59546 this should be an error.

ChiriVulpes:
No? That was the whole point of me making this PR 😛 Unless I’m misunderstanding what you mean? The PR accomplishes two things, it prevents erroring in the way I use it while preserving all other errors, and it adds new errors to right side operands that were previously not checked

jakebailey:
I'm pretty sure Ryan was saying that the error message needed to be fixed, and that an unconditional nullish value on the left side was dubious code that should still error. I'm not sure how https://github.com/microsoft/TypeScript/issues/59546#issuecomment-2274296823 could he read any other way.

ChiriVulpes:
The error message was already fixed by another commit since the release of 5.6.0-beta, it was irrelevant to my issue

tests/baselines/reference/predicateSemantics.errors.txt
```diff
@@ -41,21 +49,50 @@ predicateSemantics.ts(36,8): error TS2872: This kind of expression is always tru
...
+    const p02 = 0 > 1 ?? null;
+                ~~~~~
 !!! error TS2869: Right operand of ?? is unreachable because the left operand is never nullish.
-    const p7 = null ?? null;
-               ~~~~
-!!! error TS2871: This expression is always nullish.
-    const p8 = (class foo { }) && null;
-               ~~~~~~~~~~~~~~~
+    const p03 = null ?? 1;
+    const p04 = null ?? null;
```
RyanCavanaugh, line 59: Missing error here

RyanCavanaugh (changes requested):
Most of the diff looks good but the special case for `null ?? null` etc is not an approved change.

ChiriVulpes:
Okay so I've addressed the line you pointed out and some similar lines, unfortunately it increased the surface area of the PR a bit. I tried to limit it as much as I could.

List of changes:
- The `PredicateSemantics` of binary expressions in `getSyntacticNullishnessSemantics` are now actually calculated like the `ConditionalExpression`, which results in situations like `null ?? null ?? null` actually erroring. As a result, the right operand check for always nullish is no longer necessary, so it was removed
- Since nullish coalesce errors were all previously added to the operands of a nullish coalesce expression, `null ?? null` was not able to be caught, so now the semantics of the left and right operands are used to add a new error if both operands are always nullish
- Added a couple more test lines

Note that this does result in more overlapping errors in the baselines, but I *think* most cases in real-world code won't get much or any of that.

Hopefully this is what you wanted? Up to revert or do this differently if you prefer.

ChiriVulpes:
Actually maybe it would be more useful for me to explain the changes with two variations.

So the current code I just pushed results in errors like this:
image

The alternative would be reverting the change to start resolving `PredicateSemantics` of binary expresions and keeping the right side operand never nullish check, which looks like this:
image

I thought the former was probably closer to what you actually wanted though so that's what I went with. If you wanted something *between* the two of them, IE preferring the second version of `p22` and `p23`, but wanting to error on more than just the middle `null` of `null ?? null ?? null`, I think it would result in the surface area of this PR getting even bigger than it already is

MartinJohns:
@RyanCavanaugh Can you clarify if I missed something? In the issue #59546 you said something like `undefined ?? value` should be an error, but this PR will allow `null ?? value`. Is `null` treated differently here, did you change your mind, or should `null ?? value` be an error?

tests/baselines/reference/predicateSemantics.errors.txt
```diff
@@ -41,21 +63,82 @@ predicateSemantics.ts(36,8): error TS2872: This kind of expression is always tru
...
+                ~~~~~~~~~~~~~~~
 !!! error TS2872: This kind of expression is always truthy.
+    const p07 = null ?? null ?? null;
+                ~~~~~~~~~~~~
+!!! error TS2871: This expression is always nullish.
+                ~~~~~~~~~~~~~~~~~~~~
+!!! error TS2871: This expression is always nullish.
+                        ~~~~
+!!! error TS2871: This expression is always nullish.
+    const p08 = null ?? opt ?? null;
```
RyanCavanaugh, line 89: Missing error here too. See comment below

RyanCavanaugh (changes requested):
I don't feel like we're getting a "meeting of the minds" here and the PR is still trying to create a carve-out for the behavior described in the original bug. The only change approved from that bug is to tweak the error message to not incorrectly talk about binary expressions when one exists. It might be cleaner to start fresh on a new PR that only makes that change?

ChiriVulpes:
As I've already stated, the error message being confusing was already fixed due to other changes since the initial beta release, I just forgot to test nightly when I made my initial bug report. I wanted to make a PR that addressed my issue in whatever way was most likely to actually be considered and pulled into the project. If it won't be, then that's fine, I guess. It's bad for me, because it means that if I want to write code that's still just as comprehensible I'll have to be maintaining a TS fork or a TS patch indefinitely, but I have to do what I have to do and I don't fault anyone for having different opinions. I do think the eagerness of the error makes it a better fit for linting than an unconfigurable TS level error, but w/e

ChiriVulpes:
This PR no longer addresses my issue, and instead only fixes the error message and adds in missing checks for the right operands. I don't think this will be controversial anymore

RyanCavanaugh (approved):
Thanks!

jakebailey:
This was approved but I think just not merged. @ChiriVulpes do you mind pulling in main? If you don't have time, I'm happy to do it for you.

ChiriVulpes:
@jakebailey I think it should be good now, it's been months since I touched the TS codebase so I might have done something wrong, though. Let me know if it needs tweaks

jakebailey:
@typescript-bot test it

jakebailey:
Both things above are bugs. In hoarder:

```ts
  feedData.items.forEach((item) => {
    item.guid = item.guid ?? `${item.id}` ?? item.link;
  });
```

`${item.id}` is definitely never nullish.

In vscode:

```ts
const id = `${claims.tid}/${(claims.oid ?? (claims.altsecid ?? '' + claims.ipd ?? ''))}`;
```

This parses like:

```ts
const id = `${claims.tid}/${(claims.oid ?? (claims.altsecid ?? ('' + claims.ipd) ?? ''))}`;
```

Clearly they meant:

```ts
const id = `${claims.tid}/${(claims.oid ?? ((claims.altsecid ?? '') + (claims.ipd ?? '')))}`;
```

PR #59582, opened 2024-08-09
autoimports checks for merged parent if symbol parent is not found

### LINKED ISSUE #58907: Module augmentation seems to break entirely a class definition/body IntelliSense autocomplete. Debug failure (exception on executing command "completionInfo").

iisaduan:
Fixes #58907

### DIFF
--- a/src/services/codefixes/importFixes.ts
+++ b/src/services/codefixes/importFixes.ts
@@ -40,6 +40,7 @@ import {
     firstDefined,
     flatMap,
     flatMapIterator,
+    forEach,
     forEachExternalModuleToImportFrom,
     forEachNameOfDefaultExport,
     formatting,
@@ -92,9 +93,11 @@ import {
     isJsxOpeningFragment,
     isJsxOpeningLikeElement,
     isJSXTagName,
+    isModuleDeclaration,
     isNamedImports,
     isNamespaceImport,
     isRequireVariableStatement,
+    isSourceFile,
     isSourceFileJS,
     isStringLiteral,
     isStringLiteralLike,
@@ -122,6 +125,7 @@ import {
     NodeFlags,
     nodeIsMissing,
     ObjectBindingPattern,
+    or,
     OrganizeImports,
     PackageJsonImportFilter,
     Path,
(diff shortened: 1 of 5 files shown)


src/services/codefixes/importFixes.ts
```diff
@@ -274,10 +278,20 @@ function createImportAdderWorker(sourceFile: SourceFile | FutureSourceFile, prog
         addImport(first(info));
     }
 
+    function getExportedSymbolParent(symbol: Symbol, checker: TypeChecker): Symbol {
```
Andarist, line 281: I wonder if this is something that will only be relevant for import fixes. Could any other code crash similarly due to a missing parent?

andrewbranch: It would be nice/convenient if the checker set the parent, but I thought it was better not to touch the checker and add a bunch of parent pointers to everyone’s program if `tsc` never needs them. We could try it and see if it shows up in the perf run. At least, we should have a comment here explaining that this function is making up for an arguable checker deficiency.

andrewbranch: Tried in #59609

iisaduan:
closed by https://github.com/microsoft/TypeScript/pull/59609

PR #59604, opened 2024-08-12
Allow existing imports in file to supersede package.json filter

### LINKED ISSUE #59487: Auto-import from transitive dependency should work when existing import is present
### 🔎 Search Terms

import suggest resolve package.json subfolder paths workspace

- I tested this in multiple versions of typescript and the behavior always seems to be the same. I search google a lot. I found information about package.json imports property, and also types, typeRoots and paths property of tsconfig compilerOptions. I tried several things and nothing quite seems to fit.

### ⏯ Playground Link

_No response_

### 💻 Code

```ts
import React from "react"

useEffect() // ctrl-dot on this does not trigger the correct suggestion, if the relevant package.json is in a subfolder. even if the npm package exists in the root node_modules folder!
```

### 🙁 Actual behavior

pressing ctrl-dot on "useEffect" does **not** trigger a suggestion to import useEffect from the react package

### 🙂 Expected behavior

I expect the import suggestion to be shown, since npm indeed installs all package into the root folder, even if the package.json is in a subfolder. And it does compile and run OK if I write the import manually. But I am tired of writing imports manually.

* Why I need this: I have a  submodule that I include as npm workspace in each of my repositories. The dependecies are only declared in package.json in my submodule. This avoids the need to declare the same dependencies in each of my projects
* if this cannot be solved any other way, may I ask if there is a way to write some sort of typescript plugin that will recognise the package.json in my submodule to determine relevant import suggestions?

andrewbranch:
Fixes #59487

FWIW, I thought we had said no to this request in the past, but I think that was the wrong call if so 🤷‍♂️


andrewbranch:
@typescript-bot perf test

andrewbranch:
Interesting that xstate apparently contains imports of transitive dependencies—I wonder what they are?

src/compiler/types.ts
```diff
@@ -5461,7 +5461,7 @@ export const enum NodeBuilderFlags {
     AllowEmptyIndexInfoType                 = 1 << 21,
     // Errors (cont.)
     AllowNodeModulesRelativePaths           = 1 << 26,
-    
+
```
jakebailey, line 5464: Nothing you did I don't think but this and Dimitri's PR both had this weird diff; makes me wonder if the formatter is broken?

jakebailey (approved):

src/compiler/types.ts
andrewbranch: I was surprised these spaces were here before—maybe the formatter improved?

jakebailey: Maybe, but we enforce that all files are formatted in CI, so it shouldn't have been possible to have them at all...

sheetalkamat (approved):

Issue #59095, opened 2024-07-01
parseCommandLine fails with --build and --force


CraigMacomber:
### 🔎 Search Terms

parseCommandLine build

### 🕗 Version & Regression Information

- This is the behavior in every version I tried (5.1 and 5.4)

### 💻 Code

When using the typescript package programatically, the function `parseCommandLine` is exported. Implementation: https://github.com/microsoft/TypeScript/blob/main/src/compiler/commandLineParser.ts#L1999

As far as I can tell, this function is public but undocumented. It seems like you should be able to give it command line arguments and have it parse them, and it usually works. However, if you give it `--build` and --force` it errors with `"Compiler option '--force' may only be used with '--build'."` which is confusing since it is used with --build.

### 🙁 Actual behavior

`"Compiler option '--force' may only be used with '--build'."` which is confusing since it is used with --build.

### 🙂 Expected behavior

One of:
- Document that `parseCommandLine` does not work for the `--build` case in its doc comment, and have the error reported when `--build` is encountered. Maybe also expose `isBuild` and `parseBuildCommand`.
- Support `--build`, and only emit `"Compiler option '--force' may only be used with '--build'."` when --build is not provided.

### Additional information about the issue

I believe when TypeScript is parsing its command line inputs it does not use this function for the "build" case, based on https://github.com/microsoft/TypeScript/blob/main/src/compiler/executeCommandLine.ts#L752, which is how it works when running tsc, but not in `parseCommandLine`. In that case it uses `parseBuildCommand`, which is `@internal`.

jakebailey:
This isn't wholly unexpected as `parseCommandLine` returns `ParsedCommandLine` which is explicitly just one tsconfig, not for use with build mode. Exposing something to be able to parse the whole thing doesn't seem too unreasonable to me. @sheetalkamat what do you think?

Zakronindustry:
->  **Documentation Improvement**: Add documentation to parseCommandLine to clarify that it does not support --build options. This would help users understand the limitations of the function.
->  **Function Updates**: Modify the parseCommandLine function or provide a separate function for handling --build scenarios if it’s a common use case. You might want to use the internal parseBuildCommand function if possible or expose functionality to handle build-specific options.

sheetalkamat:
We have `parseBuildCommand` which could be made public.

PR #59605, opened 2024-08-12
Fixed handle --build and --force in parseCommandLine (#59095)

codewithsupra:
Fixes #


sheetalkamat (changes requested):
This is incorrect change and that api cannot be used to parse build mode

Issue #52892, opened 2023-02-21
Instantiation expression inside nested classes produces unexpected circular reference error


whzx5byb:
# Bug Report

### 🔎 Search Terms

instantiation expression nested class

### 🕗 Version & Regression Information

- This is the behavior in every version I tried

### ⏯ Playground Link

[Playground link with relevant code](https://www.typescriptlang.org/play?jsx=0&ts=4.9.5#code/MYGwhgzhAECCA8AVaBeaA7ArgWwEYFMAnAPmgG8Aoaa6AelogBcxGBLYc6RgTwAd9oAfSTR8AD0b50AExixUcUmgAM0AL5Ua9AHrQA7qxAh9Ae0IBrfQAsp0TOmAns2KYwD8FTdQBuYEJnwAQgAuaEQAbi9oYCtDaRDoAEl0JjAHfEQ+fHgefhMAMzgAOgAheAREYirPGmhUtg4ShVBIGDLkcUkZOQVYUkpa2t4wQlcEiKi1aAoNCl9CaDAFdHw9OAowIpi4ouHR9EYi3398CnpBi5ptNyA)

### 💻 Code

```ts
class A<T = number> {
    //static { type _<T extends A = A> = 0 }
    //^ will work when uncomment it?

    value!: T;
    child!: InstanceType<typeof A.B<A<T>>>

    static B = class B<T extends A = A> {
        parent!: T;
    } 
}

var a = new A
a.child.parent.value
//              ^?
```

### 🙁 Actual behavior

Shows an error `'child' is referenced directly or indirectly in its own type annotation.`.

However, when uncommenting `// static {...}` it will work only in playground. 
In VSCode it still shows the error above.

### 🙂 Expected behavior

`a.child.parent.value` is `number`

### Workaround

[...]

Issue #59606, opened 2024-08-12
Non-first occurrence of circular class instantiation is always silently allowed


Andarist:
### 🔎 Search Terms

variance class expression instantiation reference

### 🕗 Version & Regression Information

- This is the behavior in every version I tried

### ⏯ Playground Link

https://www.typescriptlang.org/play/?ts=5.6.0-dev.20240812#code/MYGwhgzhAECCA8AVaBeaA7ArgWwEYFMAnAPmgG8AoaaANzBE3wEIAuaRAbiumAAsBLEABNW0AC4BPAA74A9gDM4AOgBC8BImLEO0APS7oRQrMIRufQUIBMoyTIXK1GrTv3QIYbPmiQfuWTTeuJhi0EKy+BDoAOShRiYU3BBiYGL8wNAqqDzgUJlIhgAeYvjoQjCw2bCklNTUUmCEpWKinNwAvlztiaCQFVYFaFh4RDXcdAzMbG3UFsK20nKKsFaq6gPDBCTa5gLCNmx2S3CrTgOa2noGHl4+FUpzQtAAFGBlyo9WAJTQwaHhkRicUIxkIiWoyVS6Uy2V6eTUyHwxVK5ROVSsYzq0AaTXQLWmXGonQo3SAA

### 💻 Code

```ts
class A<T = number> {
  value!: T;
  child!: typeof A.B<A<T>>; // errors
  child2!: typeof A.B<A<T>>; // same as above but doesn't error

  static B = class B<T extends A = A> {
    parent!: T;
  };
}

class A2<T = number> {
  value!: T;
  child!: typeof A2.B<A2<number>>;
  child2!: typeof A2.B<A2<T>>; // same as A.child (and A.child2) but doesn't error

  static B = class B<T extends A2 = A2> {
    parent!: T;
  };
}
```

### 🙁 Actual behavior

`child2` in both classes is free of errors

### 🙂 Expected behavior

both `child2` properties should contain errors (like `child` in `A`)

### Additional information about the issue

This is related to the fact that `child2` reenters `getVariancesWorker` for the same symbol, so it returns with `emptyArray` that was eagerly set as `links.variances` on the first entry of this function. This, in turn, is a signal for `structuredTypeRelatedToWorker` to return `Ternary.Unknown`.

Since this is truly a circular situation, I think this is effectively a design limitation. I'm only reporting this for documentation purposes.

PR #53549, opened 2023-03-27
Reset type resolution stack when starting to compute variance

gabritto:
Fixes #53470

PR #59607, opened 2024-08-12
Defer checking of class static blocks

Andarist:
closes https://github.com/microsoft/TypeScript/issues/52892

This only helps with a specific situation when a type alias defined in a class static block triggered the first variance check in those circular situations. Then all properties referencing `A` again were reentering `getVariancesWorker` and thus it was assumed that they were OK. My comment at the bottom of this post explains more about this kind of situation: https://github.com/microsoft/TypeScript/issues/59606#issue-2461841606

In other words, this only mitigates the problem in this specific situation but it's still possible to run into problems in similar circular situations.

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -41494,7 +41494,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         }
     }
 
-    function checkClassStaticBlockDeclaration(node: ClassStaticBlockDeclaration) {
+    function checkClassStaticBlockDeclarationDeferred(node: ClassStaticBlockDeclaration) {
         checkGrammarModifiers(node);
 
         forEachChild(node, checkSourceElement);
@@ -47937,7 +47937,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             case SyntaxKind.MethodSignature:
                 return checkMethodDeclaration(node as MethodDeclaration | MethodSignature);
             case SyntaxKind.ClassStaticBlockDeclaration:
-                return checkClassStaticBlockDeclaration(node as ClassStaticBlockDeclaration);
+                return checkNodeDeferred(node);
             case SyntaxKind.Constructor:
                 return checkConstructorDeclaration(node as ConstructorDeclaration);
             case SyntaxKind.GetAccessor:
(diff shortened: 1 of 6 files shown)


tests/cases/fourslash/quickInfoAvoidsClassVarianceCircularity1.ts
Andarist, line 1: this one was already fixed by https://github.com/microsoft/TypeScript/pull/53549 , this is just an extra test case

RyanCavanaugh:
@typescript-bot test it

src/compiler/checker.ts
```diff
@@ -47937,7 +47937,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             case SyntaxKind.MethodSignature:
                 return checkMethodDeclaration(node as MethodDeclaration | MethodSignature);
             case SyntaxKind.ClassStaticBlockDeclaration:
-                return checkClassStaticBlockDeclaration(node as ClassStaticBlockDeclaration);
+                return checkNodeDeferred(node);
```
jakebailey: Maybe a stupid question, but why is it that this matters here, yet MethodSignatures are not deferred?

jakebailey: As in, can't we write that same local type declaration in a method and it breaks? Or is that fine because the _body_ gets deferred?

Andarist: My understanding is that:
1. The same local type declaration in a method shouldn't suppress the error either.
2. Object literal methods are deferred but class methods are not. If I try to defer class methods... a ton of tests break. So there has to be a reason behind why they are not deferred. Perhaps it's important for the variance worker?
3. Class static blocks conceptually execute after the class declaration and they shouldn't affect the variance measurement and shouldn't trigger circularities in a way class methods/properties can. So it should be OK to defer them. They should behave in an almost same way as statements executed after the class. They only get a special treatment when it comes to class type parameters and `this`. But it feels like it shouldn't break anything if those get simply resolved in a deferred manner - after the class declaration gets processed.
4. Right now it looks like this PR doesn't do anything because a bug has been introduced in the meantime: https://github.com/microsoft/TypeScript/issues/61982

PR #59608, opened 2024-08-12
fix(59587): permit trailing commas in type parameter printing

### LINKED ISSUE #59587: Printer introduces syntax error into .tsx files
### 🔎 Search Terms

tsx printer printing syntax error generic type parameter

- This is the behavior in every version I tried.
- Bug occurs here in the TypeScript source code: https://github.com/microsoft/TypeScript/blob/4b12d82b6bb4c8e2ad7df3219ae7282e360e62f9/src/compiler/emitter.ts#L4507-L4512

### ⏯ Playground Link

https://github.com/maxpatiiuk/typescript-printer-tsx-syntax-error

### 💻 Code

```js
import ts from 'typescript';

const sourceCode = 'export const id = <T,>(id: T): T => id';
const sourceFile = ts.createSourceFile(
  'index.tsx',
  sourceCode,
  ts.ScriptTarget.Latest,
  undefined,
  ts.ScriptKind.TSX
);

const printer = ts.createPrinter();
const printed = printer.printFile(sourceFile);

console.log('Original:');
console.log(sourceCode); // 'export const id = <T,>(id: T): T => id;'
console.log('Printed:');
console.log(printed); // 'export const id = <T>(id: T): T => id;'

```

### 🙁 Actual behavior

The printer removes trailing comma from the generic type parameters list.
That trailing comma is crucial to correctly parse the `.tsx` files. See how without it TypeScript (and ESBuilt and others) are reporting a syntax error: https://www.typescriptlang.org/play/?target=99#code/KYDwDg9gTgLgBAYwgOwM7wEbAGbWABSgEtl4BeOAHgBUAaAPgAoiATALjmoEoPq4z6cVgChQkWIhTo4AQ2wxgUQiXJVqTVrx6d+g1kA

### 🙂 Expected behavior

When generic type parameters are used in `.tsx` files, the list must end with `,` to disambiguate the syntax from jsx opening tag

See also https://github.com/microsoft/TypeScript/issues/15713

a-tarasyuk:
Fixes #59587

### DIFF
--- a/src/compiler/types.ts
+++ b/src/compiler/types.ts
@@ -10013,7 +10013,7 @@ export const enum ListFormat {
     SourceFileStatements = MultiLine | NoTrailingNewLine,
     Decorators = MultiLine | Optional | SpaceAfterList,
     TypeArguments = CommaDelimited | SpaceBetweenSiblings | SingleLine | AngleBrackets | Optional,
-    TypeParameters = CommaDelimited | SpaceBetweenSiblings | SingleLine | AngleBrackets | Optional,
+    TypeParameters = CommaDelimited | SpaceBetweenSiblings | SingleLine | AngleBrackets | Optional | AllowTrailingComma,
     Parameters = CommaDelimited | SpaceBetweenSiblings | SingleLine | Parenthesis,
     IndexSignatureParameters = CommaDelimited | SpaceBetweenSiblings | SingleLine | Indented | SquareBrackets,
     JSDocComment = MultiLine | AsteriskDelimited,
--- a/src/testRunner/unittests/printer.ts
+++ b/src/testRunner/unittests/printer.ts
@@ -100,6 +100,16 @@ describe("unittests:: PrinterAPI", () => {
                 ts.ScriptKind.TSX,
             ));
         });
+
+        printsCorrectly("typeParameters", {}, printer => {
+            return printer.printFile(ts.createSourceFile(
+                "source.tsx",
+                String.raw`export const id = <T,>(id: T): T => id;`,
+                ts.ScriptTarget.Latest,
+                /*setParentNodes*/ undefined,
+                ts.ScriptKind.TSX,
+            ));
+        });
     });
 
     describe("No duplicate ref directives when emiting .d.ts->.d.ts", () => {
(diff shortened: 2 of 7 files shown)


src/compiler/types.ts
```diff
@@ -10013,7 +10013,7 @@ export const enum ListFormat {
     SourceFileStatements = MultiLine | NoTrailingNewLine,
     Decorators = MultiLine | Optional | SpaceAfterList,
     TypeArguments = CommaDelimited | SpaceBetweenSiblings | SingleLine | AngleBrackets | Optional,
```
jakebailey, line 10015: Can this happen with type arguments within JSX too?

src/compiler/types.ts
```diff
@@ -10013,7 +10013,7 @@ export const enum ListFormat {
     SourceFileStatements = MultiLine | NoTrailingNewLine,
     Decorators = MultiLine | Optional | SpaceAfterList,
     TypeArguments = CommaDelimited | SpaceBetweenSiblings | SingleLine | AngleBrackets | Optional,
-    TypeParameters = CommaDelimited | SpaceBetweenSiblings | SingleLine | AngleBrackets | Optional,
+    TypeParameters = CommaDelimited | SpaceBetweenSiblings | SingleLine | AngleBrackets | Optional | AllowTrailingComma,
```
a-tarasyuk, line 10016: @weswigham is it acceptable to allow a trailing comma in all type parameters if it exists, or is it better to exclude it in certain cases?

src/compiler/types.ts
weswigham: No - it's a quirk of lambda parsing that they require a `,` , an `extends`, or a `=` clause to disambiguate them from a jsx tag.

src/compiler/types.ts
weswigham: https://github.com/microsoft/TypeScript/pull/59612 - definitely only want it for arrow functions. The trailing comma looks odd everywhere, but is only syntactically required for some arrows.

a-tarasyuk: @weswigham ok

PR #59610, opened 2024-08-12
Avoid checking class expressions more than once

Andarist:
When investigating https://github.com/microsoft/TypeScript/issues/52892 I noticed that a class expression in the circular situation could be checked more than once. 

I couldn't recognize any issue being *fixed* by this. I manually confirmed that various diagnostics could be created by this wrapped code more than once... but `DiagnosticCollection['add']` uses a deduplication mechanism so this can't be observed.


DanielRosenwasser:
@typescript-bot test it

PR #59613, opened 2024-08-12
Simplify mergeEmitNode

jakebailey:
@armanio123 and I noticed in #59332 that simply copying `leadingComments`/`trailingComments` and force overwriting (rather than concatenating) fixes the comment duplication problem from #59332.

Reading things further, it the function already force copies over the other node's information (flags, comment ranges, etc). Out of interest, I tried removing some of the other "complicated" merging operations and found that they also worked fine when just regular slices.

Maybe this extra simplification is valid?

### DIFF
--- a/src/compiler/factory/nodeFactory.ts
+++ b/src/compiler/factory/nodeFactory.ts
@@ -7449,13 +7449,13 @@ function mergeEmitNode(sourceEmitNode: EmitNode, destEmitNode: EmitNode | undefi
     // `leadingComments` are concatenated with any existing leading comments on the destination
     if (leadingComments) {
         // We use `.slice()` in case `destEmitNode.leadingComments` is pushed to later
-        destEmitNode.leadingComments = addRange(leadingComments.slice(), destEmitNode.leadingComments);
+        destEmitNode.leadingComments = leadingComments.slice();
     }
 
     // `trailingComments` are concatenated with any existing trailing comments on the destination
     if (trailingComments) {
         // We use `.slice()` in case `destEmitNode.trailingComments` is pushed to later
-        destEmitNode.trailingComments = addRange(trailingComments.slice(), destEmitNode.trailingComments);
+        destEmitNode.trailingComments = trailingComments.slice();
     }
 
     // `commentRange` overwrites the destination
@@ -7470,7 +7470,7 @@ function mergeEmitNode(sourceEmitNode: EmitNode, destEmitNode: EmitNode | undefi
 
     // `tokenSourceMapRanges` are merged with the destination
     if (tokenSourceMapRanges) {
-        destEmitNode.tokenSourceMapRanges = mergeTokenSourceMapRanges(tokenSourceMapRanges, destEmitNode.tokenSourceMapRanges!);
+        destEmitNode.tokenSourceMapRanges = tokenSourceMapRanges.slice();
     }
 
     // `constantValue` overwrites the destination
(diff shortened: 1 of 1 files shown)


jakebailey:

src/compiler/factory/nodeFactory.ts
```diff
@@ -7517,11 +7515,3 @@ function mergeEmitNode(sourceEmitNode: EmitNode, destEmitNode: EmitNode | undefi
...
     return destEmitNode;
 }
-
-function mergeTokenSourceMapRanges(sourceRanges: (TextRange | undefined)[], destRanges: (TextRange | undefined)[]) {
-    if (!destRanges) destRanges = [];
-    for (const key in sourceRanges) {
-        destRanges[key] = sourceRanges[key];
-    }
-    return destRanges;
-}
```
jakebailey, line 7527: I read this function and it sort of scared me; what if the test has more elements than the source? Then I replaced it and nothing changed, so I kept going...

rbuckton:
A lot of the complexity in `mergeEmitNode` is an attempt to be defensive against someone creating a synthetic node, setting various emit related properties on it (via `emitNode`), and *then* calling `setOriginalNode`. We likely don't have good test coverage for this, so it's possible that this could break existing custom transformers somewhere.

src/compiler/factory/nodeFactory.ts
rbuckton: `sourceRanges` and `destRanges` are holey arrays, so `length` doesn't actually matter. It probably should just be a `Map<number, TextRange>`, to be honest.

src/compiler/factory/nodeFactory.ts
```diff
@@ -7449,13 +7446,13 @@ function mergeEmitNode(sourceEmitNode: EmitNode, destEmitNode: EmitNode | undefi
     // `leadingComments` are concatenated with any existing leading comments on the destination
     if (leadingComments) {
         // We use `.slice()` in case `destEmitNode.leadingComments` is pushed to later
-        destEmitNode.leadingComments = addRange(leadingComments.slice(), destEmitNode.leadingComments);
+        destEmitNode.leadingComments = leadingComments.slice();
```
rbuckton, line 7449: I worry this results in a different inconsistency now. Before, we would attempt to merge the emit information for the old node *into* anything set on the new node prior to calling `setOriginalNode`. Now we will conditionally overwrite some properties but not others.

```ts
const oldNode = ...;
addSyntheticLeadingComment(oldNode, SyntaxKind.SingleLineCommentTrivia, "foo");

const newNode = ...;
addSyntheticLeadingComment(oldNode, SyntaxKind.SingleLineCommentTrivia, "bar");
addSyntheticTrailingComment(oldNode, SyntaxKind.SingleLineCommentTrivia, "baz");

setOriginalNode(newNode, oldNode);

// before
getSyntheticLeadingComments(newNode).length; // 2 (foo and bar)
getSyntheticTrailingComments(newNode).length; // 1 (baz)

// after
getSyntheticLeadingComments(newNode).length; // 1 (foo)
getSyntheticTrailingComments(newNode).length; // 1 (baz)
```

After this change we end up with `oldNode`'s leading comments but `newNodes` trailing comments, which seems more confusing.

jakebailey: I certainly am not confident in this PR, but we do need _something_ for #59332.

armanio123: I originally considered using `appendIfUnique` to avoid duplicating but also to concatenate. Would you think that's a good fix?

rbuckton:
The issue that drove this PR is not so much an issue with `mergeEmitNode` but with the implementation in #59332. I don't think we should take this change.

rbuckton:
I found the actual underlying problem in #59332 in `setTextRange()` in checker.ts, per https://github.com/microsoft/TypeScript/pull/59332#discussion_r1718920733. I'm working up a PR for it now.

Issue #59600, opened 2024-08-12
[5.5 regression] "Maximum call stack size exceeded" when serializing nested self-referencing class expressions


Andarist:
### 🔎 Search Terms

dts declaration serializing class expression self-referencing nested

### 🕗 Version & Regression Information

- This changed between versions 5.4.x and 5.5.x

### ⏯ Playground Link

https://www.typescriptlang.org/play/?ts=5.6.0-dev.20240812#code/MYewdgzgLgBAgjAvDYAbAhhC8A8AVAPhgG8AoGFACwEtUATAQgC4YBJSKdMYAUzwE8ADjxxQhPEADN4AOgBC+AgQDc5GNHRRqwGHKQoMWXYpJqKg9ACceYKMxh5VFAL6rXpIA

### 💻 Code

```ts
const A = class A<T> {
  child!: InstanceType<typeof A.B<T>>;
  static B = class B<T> {
    parent!: T;
  };
};
```

### 🙁 Actual behavior

it crashes

### 🙂 Expected behavior

it shouldn't crash 😉

PR #59619, opened 2024-08-13
Fixed crash when reusing nodes referring to symbols inaccessible after emit

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/59600
I bisected the regression to https://github.com/microsoft/TypeScript/pull/58066 , cc @weswigham

### DIFF
--- a/tests/baselines/reference/classExpressionInClassInstantiationEmit1.js
+++ b/tests/baselines/reference/classExpressionInClassInstantiationEmit1.js
@@ -1,0 +1,252 @@
+//// [tests/cases/compiler/classExpressionInClassInstantiationEmit1.ts] ////
+
+//// [classExpressionInClassInstantiationEmit1.ts]
+class A1<T> {
+  child!: InstanceType<typeof A1.B<T>>;
+  static B = class B<T> {
+    parent!: T;
+  };
+};
+
+const A2 = class<T> {
+  child!: InstanceType<typeof A2.B<T>>;
+  static B = class B<T> {
+    parent!: T;
+  };
+};
+
+const A3 = class A3<T> {
+  child!: InstanceType<typeof A3.B<T>>;
+  static B = class B<T> {
+    parent!: T;
+  };
+};
+
+const A4 = class C<T> {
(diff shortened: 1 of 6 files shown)


tests/baselines/reference/classExpressionInClassInstantiationEmit1.js
```diff
@@ -0,0 +1,252 @@
...
+        function B() {
+        }
+        return B;
+    }()),
+    _f);
+
+
+//// [classExpressionInClassInstantiationEmit1.d.ts]
+declare class A1<T> {
+    child: InstanceType<typeof A1.B<T>>;
```
Andarist, line 162: Note how here `InstanceType` is still reused (unlike in 5.4). That's what made me pursue the current solution instead of trying to tweak the function changed by https://github.com/microsoft/TypeScript/pull/58066 (even if the change to this function introduced this regression)

weswigham (commented):
Hm. Not 100% sure this tracks fully. I get the idea - if we already tried to reuse a node, but reuse for that node failed, don't try and use it again elsewhere, since we may have already issued the relevant diagnostic and may fail to repeat the error (though... shouldn't we reissue the error?). But the structure we're holding onto here to ensure that seems to be not quite accurate - I would think a lot of errors that end up marking a node as not reusable are dependent not just on the node in question, but the `enclosingDeclaration`+`enclosingFile` active in the context and the `flags` in the context; if those differ, I'm not sure that you could actually guarantee that you shouldn't reuse the node. This seems like it's going to be much more pessimistic than it strictly needs to be because of that.

Andarist:
I agree with your point. With those constraints... it becomes much harder to track the correct nodes that can't be reused. Not impossible but much harder. `context` is manipulated in many places and I don't have enough context (😉) on this part of the codebase, it would be rather easy for me to miss some nuance when working on the more correct node tracking.

That makes me question the solution I chose here - could this solution be replaced with something else? My initial idea was to tweak `existingTypeNodeIsNotReferenceOrIsReferenceWithCompatibleTypeArgumentCount` but I couldn't figure out a proper change there.

PR #59622, opened 2024-08-13
Always treat `.d.*.ts` files as commonjs-format modules for the purposes of import interop

### LINKED ISSUE #57229: .d.json.ts file (allowArbitraryExtensions) not working correctly when importing from ESM file with node16/nodenext module setting
### 🔎 Search Terms

allowArbitraryExtensions, import, JSON, d.json.ts, nodenext, node16, mjs, mts, "type": "module"

- This is the behavior in every version I tried, and I reviewed the FAQ for entries

### ⏯ Playground Link

[https://codesandbox.io/p/devbox/json-import-bug-frj89p](https://codesandbox.io/p/devbox/json-import-bug-frj89p?file=%2Findex.ts%3A5%2C43&layout=%257B%2522sidebarPanel%2522%253A%2522EXPLORER%2522%252C%2522rootPanelGroup%2522%253A%257B%2522direction%2522%253A%2522horizontal%2522%252C%2522contentType%2522%253A%2522UNKNOWN%2522%252C%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522id%2522%253A%2522ROOT_LAYOUT%2522%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522contentType%2522%253A%2522UNKNOWN%2522%252C%2522direction%2522%253A%2522vertical%2522%252C%2522id%2522%253A%2522clrztojnh0006356gw3xtxjhi%2522%252C%2522sizes%2522%253A%255B73.71660859465737%252C26.283391405342627%255D%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522contentType%2522%253A%2522EDITOR%2522%252C%2522direction%2522%253A%2522horizontal%2522%252C%2522id%2522%253A%2522EDITOR%2522%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL%2522%252C%2522contentType%2522%253A%2522EDITOR%2522%252C%2522id%2522%253A%2522clrztojng0002356gevchi9el%2522%257D%255D%257D%252C%257B%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522contentType%2522%253A%2522SHELLS%2522%252C%2522direction%2522%253A%2522horizontal%2522%252C%2522id%2522%253A%2522SHELLS%2522%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL%2522%252C%2522contentType%2522%253A%2522SHELLS%2522%252C%2522id%2522%253A%2522clrztojng0003356gfm9fcqxz%2522%257D%255D%252C%2522sizes%2522%253A%255B100%255D%257D%255D%257D%252C%257B%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522contentType%2522%253A%2522DEVTOOLS%2522%252C%2522direction%2522%253A%2522vertical%2522%252C%2522id%2522%253A%2522DEVTOOLS%2522%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL%2522%252C%2522contentType%2522%253A%2522DEVTOOLS%2522%252C%2522id%2522%253A%2522clrztojng0005356g41wfg52x%2522%257D%255D%252C%2522sizes%2522%253A%255B100%255D%257D%255D%252C%2522sizes%2522%253A%255B100%252C0%255D%257D%252C%2522tabbedPanels%2522%253A%257B%2522clrztojng0002356gevchi9el%2522%253A%257B%2522id%2522%253A%2522clrztojng0002356gevchi9el%2522%252C%2522tabs%2522%253A%255B%255D%257D%252C%2522clrztojng0005356g41wfg52x%2522%253A%257B%2522id%2522%253A%2522clrztojng0005356g41wfg52x%2522%252C%2522tabs%2522%253A%255B%255D%257D%252C%2522clrztojng0003356gfm9fcqxz%2522%253A%257B%2522id%2522%253A%2522clrztojng0003356gfm9fcqxz%2522%252C%2522tabs%2522%253A%255B%255D%257D%257D%252C%2522showDevtools%2522%253Afalse%252C%2522showShells%2522%253Atrue%252C%2522showSidebar%2522%253Atrue%252C%2522sidebarPanelSize%2522%253A15%257D)

### 💻 Code

tsconfig.json
```json
{
  "compilerOptions": {
    "target": "esnext",
    "module": "nodenext",
    "resolveJsonModule": true,
    "allowArbitraryExtensions": true,
    "noImplicitAny": true
  }
}
```

foo.json
```json
["foo", "bar", "baz"]
```

foo.d.json.ts
```ts
// In this contrived example we're declaring it as a tuple since TS otherwise infers it as string[]
declare const data: ["foo", "bar", "baz"];
export default data;
```

index.mts, or index.ts with `"type": "module"` in package.json
```ts
import data from "./foo.json" with { type: "json" };

// Apparently not an array, these all error!
const x = data satisfies any[];
data.forEach((item) => console.log(item));
type Item = (typeof data)[number];

// TS thinks it lives on .default
const x2 = data.default satisfies any[];
data.default.forEach((item) => console.log(item));
type Item2 = (typeof data.default)[number];
```

### 🙁 Actual behavior

The compiler/intellisense thinks the JSON value lives on the `.default` property, as if the import were a CommonJS/"synthetic" ES module.

### 🙂 Expected behavior

It should know that the "base" (actual default) import holds the JSON value, which is the behaviour when there isn't a .d.json.ts file present and most importantly the runtime behaviour.

I tried renaming to *.d.json.mts but then the file simply doesn't get picked up.

The motivation for using this feature in the first place is because I'm importing a big JSON array of objects that causes a long delay every time autocompletion is invoked or something is hovered and generally makes things unbearably sluggish, and fails to infer the type anyway. Some people reporting the same thing in this issue: #48364

Side note: if you're messing around in the CodeSandbox sometimes the intellisense won't update unless you do something like modify the import path to trigger it. Also if you use the terminal you should use `npx tsc` because the globally installed version (at the time of writing) is 5.2 which complains about import attribute syntax. So I recommend testing locally.

weswigham:
Fixes #57229

The way `allowSyntheticDefaultImports`/`esModuleInterop` and `module: nodenext`'s `"type": "module"` detection interplay in the context of `.d.*.ts` files right now is... confusing. Since they *are* `.ts` files, they do _currently_ get assigned a `package.json`-`type`-dependent module kind, which determines if the module is made available as a `default` or not on an esm import, and if a cjs file can import it _at all_, and furthermore we have heuristics involving the presence of a `default` in the declaration file which can then disable the synthetic `default` creation for interop'd imports (import statements in cjs format files under `module: nodenext`).

This menagerie of behaviors means that to write a declaration file for a json file similar to `["foo", "bar"]`, so _is no single way to write it that works for all callers, in all package type contexts, even though json imports (or any non-js import) are not package type dependent_.

On investigation, I find this state of affairs confusing.

This PR changes how we interpret `.d.*.ts` file module formats to _always_ be cjs-like. This means the correct way to write the declaration file for `["foo", "bar"]` is always
```ts
declare const _export: ["foo", "bar"];
export = _export;
```
and registers as importable by all import and require constructs in all package type contexts.

### DIFF
--- a/src/compiler/program.ts
+++ b/src/compiler/program.ts
@@ -1396,8 +1396,8 @@ export function getImpliedNodeFormatForFileWorker(
     const shouldLookupFromPackageJson = ModuleResolutionKind.Node16 <= moduleResolution && moduleResolution <= ModuleResolutionKind.NodeNext
         || pathContainsNodeModules(fileName);
     return fileExtensionIsOneOf(fileName, [Extension.Dmts, Extension.Mts, Extension.Mjs]) ? ModuleKind.ESNext :
-        fileExtensionIsOneOf(fileName, [Extension.Dcts, Extension.Cts, Extension.Cjs]) ? ModuleKind.CommonJS :
-        shouldLookupFromPackageJson && (fileExtensionIsOneOf(fileName, [Extension.Dts, Extension.Ts, Extension.Tsx, Extension.Js, Extension.Jsx]) || isDeclarationFileName(fileName)) ? lookupFromPackageJson() :
+        fileExtensionIsOneOf(fileName, [Extension.Dcts, Extension.Cts, Extension.Cjs]) || (isDeclarationFileName(fileName) && !fileExtensionIs(fileName, Extension.Dts)) ? ModuleKind.CommonJS :
+        shouldLookupFromPackageJson && fileExtensionIsOneOf(fileName, [Extension.Dts, Extension.Ts, Extension.Tsx, Extension.Js, Extension.Jsx]) ? lookupFromPackageJson() :
         undefined; // other extensions, like `json` or `tsbuildinfo`, are set as `undefined` here but they should never be fed through the transformer pipeline
 
     function lookupFromPackageJson(): Partial<CreateSourceFileOptions> {
--- a/src/compiler/program.ts
+++ b/src/compiler/program.ts
@@ -1397,7 +1397,7 @@ export function getImpliedNodeFormatForFileWorker(
         || pathContainsNodeModules(fileName);
     return fileExtensionIsOneOf(fileName, [Extension.Dmts, Extension.Mts, Extension.Mjs]) ? ModuleKind.ESNext :
         fileExtensionIsOneOf(fileName, [Extension.Dcts, Extension.Cts, Extension.Cjs]) ? ModuleKind.CommonJS :
-        shouldLookupFromPackageJson && fileExtensionIsOneOf(fileName, [Extension.Dts, Extension.Ts, Extension.Tsx, Extension.Js, Extension.Jsx]) ? lookupFromPackageJson() :
+        shouldLookupFromPackageJson && (fileExtensionIsOneOf(fileName, [Extension.Dts, Extension.Ts, Extension.Tsx, Extension.Js, Extension.Jsx]) || isDeclarationFileName(fileName)) ? lookupFromPackageJson() :
         undefined; // other extensions, like `json` or `tsbuildinfo`, are set as `undefined` here but they should never be fed through the transformer pipeline
 
     function lookupFromPackageJson(): Partial<CreateSourceFileOptions> {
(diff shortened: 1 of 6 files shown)


andrewbranch:
With `--allowJsonModule`, we leave the `impliedNodeFormat` of JSON source files as `undefined`. It feels like that might be the right move here? I expect the results to be the same, but it would leave us the ability to futz with things later if needed. I'm imagining a scenario where, say, a bundler/runtime loader converts a CSS file to an ES module, and the declaration file for it looks like

```ts
export const class1: string;
export const class2: string;
```

With this change, it becomes legal to import that file like

```ts
import css from "./styles.css";
```

when in fact no default import will be synthesized by the runtime since the loader exposed it as a real module. If we had any precedent for doing something like this, maybe it would be nice to assign the module kind based on module syntax (CJS for `export =`, ESM otherwise).

I think this will work alright and is better than the status quo, but I can see us needing to revisit it if some third-party framework begins making really heavy use of `.d.*.ts` files or something.

PR #59623, opened 2024-08-13
Experiment: Always report `useDefineForClassFields`-related errors even when disabled

rbuckton:
This is an experiment to establish a rough estimate of how frequently projects in the wild depend on the initializer order for TypeScript "Parameter Property Initializers" feature when used in conjunction with `useDefineForClassFields: false`.

Note that this will only catch cases where projects have an ordering dependency that would be reported as a use-before-definition error by the compiler when `useDefineForClassFields` is enabled.


rbuckton:
@typescript-bot test top1000

trevorade:
FYI, this commit is most likely missing a change to make `isBlockScopedNameDeclaredBeforeUse` ignore `emitStandardClassFields` when called from `checkPropertyNotUsedBeforeDeclaration`. I needed to also update that code to make `TS2729` universally be emitted.

PR #59633, opened 2024-08-14
built-in iterators should be disposable

### LINKED ISSUE #59263: Iterator and AsyncIterator missing Symbol.dispose method in esnext.disposable

rbuckton:
This updates the definitions for `IteratorObject` and `AsyncIteratorObject` to be disposable, per the spec:

- https://arai-a.github.io/ecma262-compare/?pr=3000&id=sec-%25iteratorprototype%25-%40%40dispose
- https://arai-a.github.io/ecma262-compare/?pr=3000&id=sec-%25asynciteratorprototype%25-%40%40asyncdispose

It should be noted that `%IteratorPrototype%` is synonymous with `Iterator.prototype` per the Iterator Helpers proposal and is represented by `IteratorObject` in TypeScript, and that `%AsyncIteratorPrototype%` is synonymous with `AsyncIterator.prototype` per the Async Iterator Helpers proposal and is represented by `AsyncIteratorObject` in TypeScript.

Fixes #59263

### DIFF
--- a/src/testRunner/unittests/evaluation/awaitUsingDeclarations.ts
+++ b/src/testRunner/unittests/evaluation/awaitUsingDeclarations.ts
@@ -1924,11 +1924,16 @@ describe("unittests:: evaluation:: awaitUsingDeclarations", () => {
     it("'await using' with downlevel generators", async () => {
         abstract class Iterator {
             return?(): void;
-            [evaluator.FakeSymbol.iterator]() { return this; }
-            [evaluator.FakeSymbol.dispose]() { this.return?.(); }
+            [evaluator.FakeSymbol.iterator]() {
+                return this;
+            }
+            [evaluator.FakeSymbol.dispose]() {
+                this.return?.();
+            }
         }
 
-        const { main } = evaluator.evaluateTypeScript(`
+        const { main } = evaluator.evaluateTypeScript(
+            `
             let exited = false;
 
             function * f() {
@@ -1948,11 +1953,14 @@ describe("unittests:: evaluation:: awaitUsingDeclarations", () => {
 
                 return exited;
             }
-        `, {
-            target: ts.ScriptTarget.ES5,
-        }, {
-            Iterator
-        });
+        `,
+            {
+                target: ts.ScriptTarget.ES5,
+            },
+            {
+                Iterator,
+            },
+        );
 
         const exited = await main();
         assert.isTrue(exited, "Expected 'await using' to dispose generator");
@@ -1961,11 +1969,16 @@ describe("unittests:: evaluation:: awaitUsingDeclarations", () => {
     it("'await using' with downlevel async generators", async () => {
         abstract class AsyncIterator {
             return?(): PromiseLike<void>;
-            [evaluator.FakeSymbol.asyncIterator]() { return this; }
-            async [evaluator.FakeSymbol.asyncDispose]() { await this.return?.(); }
+            [evaluator.FakeSymbol.asyncIterator]() {
+                return this;
+            }
+            async [evaluator.FakeSymbol.asyncDispose]() {
+                await this.return?.();
+            }
         }
 
-        const { main } = evaluator.evaluateTypeScript(`
+        const { main } = evaluator.evaluateTypeScript(
+            `
             let exited = false;
 
             async function * f() {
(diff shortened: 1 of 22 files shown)


src/testRunner/unittests/evaluation/awaitUsingDeclarations.ts
```diff
@@ -1874,4 +1874,140 @@ describe("unittests:: evaluation:: awaitUsingDeclarations", () => {
             "catch",
         ]);
     });
+
+    it("deterministic collapse of Await", async () => {
+        const { main, output } = evaluator.evaluateTypeScript(
```
sandersn, line 1879: Are these tests of existing functionality? I don't see how changing the type of [Async]IteratorObject would change its runtime behaviour.

sandersn (approved):
I like the code change, but I had a couple of questions about the tests.

src/testRunner/unittests/evaluation/awaitUsingDeclarations.ts
rbuckton: This test was in the wrong test suite. I merely moved while adding other tests.

tests/baselines/reference/usingDeclarationsWithIteratorObject.errors.txt
```diff
@@ -0,0 +1,31 @@
...
+    
+    function f() {
+        // should pass
+        using it0 = io;
+        using it1 = g;
+        using it2 = Iterator.from(i)
+        using it3 = new MyIterator();
+        using it4 = [].values();
+        using it5 = new Map<string, string>().entries();
+        using it6 = new Set<string>().keys();
```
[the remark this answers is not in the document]
rbuckton: `it7` is illustrating that the `Iterator` interface (which only describes a minimal Iterator compatible with `for..of`) does not implement `Symbol.dispose`. Only iterators that inherit from the `IteratorObject` interface (which represents the global `Iterator.prototype` value) will have a `Symbol.dispose` by default.

Iterators returned from `values()`/`keys()`/`entries()` on `Map`/`Set`/`Array`/etc. will inherit from `IteratorObject` and thus are disposable.

trevorade:
I'm seeing some build errors that are related to this:

image

Rather than,

```ts
interface IteratorObject<T, TReturn, TNext> extends Disposable {
}
```

Should it have been:

```ts
interface Iterator<T, TReturn, TNext> extends Disposable {
}
```

`IteratorObject` extends `Iterator` so the mismatch of `IteratorObject` being disposable but `Iterator` _not_ being disposable causes compiler errors in other locations.

jakebailey:
No, it being on the interface isn't correct, I think you just need to update your `@types/node` to pick up fixes we made.

trevorade:
Thx for the guidance!

For anyone else looking into this, the referenced PR is:
https://github.com/DefinitelyTyped/DefinitelyTyped/commit/8c36e5365e2c6ace4c00e8f21c7a10e8bbcabac3

PR #59635, opened 2024-08-14
Enable `--isolatedDeclarations` on TS codebase

### LINKED ISSUE #59097: Investigate Adopting `--isolatedDeclarations` in TypeScript's Codebase

iisaduan:
fixes #59097 : enables `--isolatedDeclarations` on our codebase


src/compiler/core.ts
```diff
@@ -329,7 +329,7 @@ export function map<T, U>(array: readonly T[] | undefined, f: (x: T, i: number)
 }
 
 /** @internal */
-export function* mapIterator<T, U>(iter: Iterable<T>, mapFn: (x: T) => U) {
+export function* mapIterator<T, U>(iter: Iterable<T>, mapFn: (x: T) => U): Generator<U, void, unknown> {
```
iisaduan, line 332: @rbuckton

src/compiler/core.ts
```diff
@@ -434,7 +434,7 @@ export function flatMapToMutable<T, U>(array: readonly T[] | undefined, mapfn: (
 }
 
 /** @internal */
-export function* flatMapIterator<T, U>(iter: Iterable<T>, mapfn: (x: T) => readonly U[] | Iterable<U> | undefined) {
+export function* flatMapIterator<T, U>(iter: Iterable<T>, mapfn: (x: T) => readonly U[] | Iterable<U> | undefined): Generator<U, void, any> {
```
iisaduan, line 437: @rbuckton 
This is a case where I changed it to `any` to allow `TNext` of `Iterable<U>` to accept values

src/compiler/core.ts
```diff
@@ -505,17 +505,17 @@ export function mapDefined<T, U>(array: readonly T[] | undefined, mapFn: (x: T,
 }
 
 /** @internal */
-export function* mapDefinedIterator<T, U>(iter: Iterable<T>, mapFn: (x: T) => U | undefined) {
+export function* mapDefinedIterator<T, U>(iter: Iterable<T>, mapFn: (x: T) => U | undefined): Generator<NonNullable<U>, void, unknown> {
```
iisaduan, line 508: @rbuckton 
I had to also add a null check for the fixed nullchecks behavior

src/compiler/core.ts
```diff
@@ -1039,14 +1039,14 @@ export function toSorted<T>(array: readonly T[], comparer?: Comparer<T>): Sorted
 }
 
 /** @internal */
-export function* arrayReverseIterator<T>(array: readonly T[]) {
+export function* arrayReverseIterator<T>(array: readonly T[]): Generator<T, void, unknown> {
```
iisaduan, line 1042: @rbuckton

src/harness/collectionsImpl.ts
```diff
@@ -158,7 +158,7 @@ export class SortedMap<K, V> {
         return undefined;
     }
 
-    public *entries() {
+    public *entries(): Generator<[K, V], undefined, unknown> {
```
iisaduan, line 161: @rbuckton

src/harness/collectionsImpl.ts
```diff
@@ -184,7 +184,7 @@ export class SortedMap<K, V> {
         return undefined;
     }
 
-    public [Symbol.iterator]() {
+    public [Symbol.iterator](): Generator<[K, V], undefined, unknown> {
```
iisaduan, line 187: @rbuckton

iisaduan (commented):
I had to use `any` in these cases since `unknown` is not assignable to `undefined` (and the inferred `TNext` type in these cases is `undefined`

src/harness/vfsUtil.ts
```diff
@@ -1367,9 +1368,10 @@ export const IOErrorMessages = Object.freeze({
     ENOTEMPTY: "directory not empty",
     EPERM: "operation not permitted",
     EROFS: "file system is read-only",
-});
+} as const;
+export const IOErrorMessages: typeof TemplateIOErrorMessages = Object.freeze(TemplateIOErrorMessages);
```
iisaduan, line 1372: @rbuckton @weswigham

src/compiler/types.ts
```diff
@@ -10055,7 +10074,7 @@ export interface PragmaDefinition<T1 extends string = string, T2 extends string
 // While not strictly a type, this is here because `PragmaMap` needs to be here to be used with `SourceFile`, and we don't
 //  fancy effectively defining it twice, once in value-space and once in type-space
 /** @internal */
-export const commentPragmas = {
+export const commentPragmas: ConcretePragmaSpecs = {
```
iisaduan, line 10077: @weswigham

src/compiler/types.ts
```diff
@@ -7541,6 +7541,7 @@ export const enum ScriptKind {
 // NOTE: We must reevaluate the target for upcoming features when each successive TC39 edition is ratified in
 //       June of each year. This includes changes to `LanguageFeatureMinimumTarget`, `ScriptTarget`,
 //       transformers/esnext.ts, commandLineParser.ts, and the contents of each lib/esnext.*.d.ts file.
+// IF UPDATING 'ScriptTarget', ALSO UPDATE 'LanguageFeatureMinimumTarget' (if needed)
```
iisaduan, line 7544: @rbuckton

src/compiler/types.ts
```diff
@@ -8306,62 +8307,80 @@ export type EmitHelperUniqueNameCallback = (name: string) => string;
  * @internal
  */
 export const enum LanguageFeatureMinimumTarget {
+    // These are copied from 'ScriptTarget' since under isolatedDeclarations, enums cannot have references to external symbols.
```
iisaduan, line 8310: @rbuckton

src/compiler/types.ts
```diff
@@ -10151,7 +10170,78 @@ export type PragmaArgumentType<KPrag extends keyof ConcretePragmaSpecs> = Concre
     : never;
 
 /** @internal */
-export type ConcretePragmaSpecs = typeof commentPragmas;
+export interface ConcretePragmaSpecs {
```
iisaduan, line 10173: @weswigham

src/compiler/core.ts
rbuckton: I think typing this as `undefined` will work for both 5.5 and 5.6:
```suggestion
export function* flatMapIterator<T, U>(iter: Iterable<T>, mapfn: (x: T) => readonly U[] | Iterable<U> | undefined): Generator<U, void, undefined> {
```

src/compiler/core.ts
rbuckton: Instead of `NonNullable<U>` you can use `U & ({} | null)` and you wouldn't need to change the check on L511. It's not as pretty but it matches the type of `value`:

```suggestion
export function* mapDefinedIterator<T, U>(iter: Iterable<T>, mapFn: (x: T) => U | undefined): Generator<U & ({} | null), void, unknown> {
```

src/compiler/utilitiesPublic.ts
```diff
@@ -669,7 +669,7 @@ function getNodeFlags(node: Node) {
 }
 
 /** @internal */
-export const supportedLocaleDirectories = ["cs", "de", "es", "fr", "it", "ja", "ko", "pl", "pt-br", "ru", "tr", "zh-cn", "zh-tw"];
+export const supportedLocaleDirectories = ["cs", "de", "es", "fr", "it", "ja", "ko", "pl", "pt-br", "ru", "tr", "zh-cn", "zh-tw"] as const;
```
iisaduan, line 672: `as const`

rbuckton (changes requested):

scripts/processDiagnosticMessages.mjs
```diff
@@ -92,6 +92,7 @@ function buildInfoFileOutput(messageTable, inputFilePathRel) {
         "    return { code, category, key, message, reportsUnnecessary, elidedInCompatabilityPyramid, reportsDeprecated };",
         "}",
         "",
+        "/* eslint-disable @typescript-eslint/no-unnecessary-type-assertion*/", // type assertions are needed for isolatedDeclarations
```
iisaduan, line 95: disabled lint rule since the assertions provide the same type as the inferred type

scripts/processDiagnosticMessages.mjs
```diff
@@ -101,7 +102,7 @@ function buildInfoFileOutput(messageTable, inputFilePathRel) {
         const argElidedInCompatabilityPyramid = elidedInCompatabilityPyramid ? `${!reportsUnnecessary ? ", /*reportsUnnecessary*/ undefined" : ""}, /*elidedInCompatabilityPyramid*/ ${elidedInCompatabilityPyramid}` : "";
         const argReportsDeprecated = reportsDeprecated ? `${!argElidedInCompatabilityPyramid ? ", /*reportsUnnecessary*/ undefined, /*elidedInCompatabilityPyramid*/ undefined" : ""}, /*reportsDeprecated*/ ${reportsDeprecated}` : "";
 
-        result.push(`    ${propName}: diag(${code}, DiagnosticCategory.${category}, "${createKey(propName, code)}", ${JSON.stringify(name)}${argReportsUnnecessary}${argElidedInCompatabilityPyramid}${argReportsDeprecated}),`);
+        result.push(`    ${propName}: diag(${code}, DiagnosticCategory.${category}, "${createKey(propName, code)}", ${JSON.stringify(name)}${argReportsUnnecessary}${argElidedInCompatabilityPyramid}${argReportsDeprecated}) as DiagnosticMessage,`);
```
iisaduan, line 105: every member in `Diagnostics` is defined with a function, and isolated declarations does not like that (and adding `as const` does not fix it). 

I chose to add `as DiagnosticMessage` to every line in order reduce the _amount_ of code, but another option would be to generate an interface for this (which I opened in #59562). Since it's a generated file, and the types are all the same, it doesn't seem like it would harm type safety for this object.

iisaduan (commented):
Here are updates to `as const` objects (and other related kinds of objects)

and this https://github.com/microsoft/TypeScript/pull/59635#discussion_r1717602292

src/compiler/core.ts
iisaduan: This did not build for me 😢

rbuckton: Then I suppose `any` is fine for now.

src/compiler/types.ts
weswigham, line 10173: 🥹

This here is why you should only enable `isolatedDeclarations` if you _really need to_, and definitely not as a default - this type is automatic and you don't have to think about it normally, since it's just derived from the expressions, but once you're forced to write it out, it's huge (despite its' simplicity!). The worst part is that now the expressions in the `commentPragmas` const and this type can accidentally diverge subtly, since we don't have "exact types". It is a degradation in UX, _and_ in type safety. Very unfortunate, but so long as we want to self-test `isolatedDeclarations`, there's no avoiding it. 🙍

src/compiler/builder.ts
```diff
@@ -2395,7 +2395,10 @@ export function getBuildInfoFileVersionMap(
     program: IncrementalBuildInfo,
     buildInfoPath: string,
     host: Pick<ReadBuildProgramHost, "useCaseSensitiveFileNames" | "getCurrentDirectory">,
-) {
+): {
+    fileInfos: Map<Path, string>;
+    roots: Map<Path, Path | undefined>;
```
jakebailey, line 2400: For cases like these where we've returned an object, I do wonder if we should be naming the type.

src/compiler/core.ts
jakebailey: Honestly I think we should just be writing `T extends {}, U extends {}` for these.

src/compiler/builder.ts
iisaduan: It certainly would make things look better in some locations, but there are lots of places where the return type is just used as that return

src/compiler/debug.ts
```diff
@@ -365,7 +365,7 @@ export namespace Debug {
     export function type<T>(value: unknown): asserts value is T;
     export function type(_value: unknown) {}
 
-    export function getFunctionName(func: AnyFunction) {
+    export function getFunctionName(func: AnyFunction): any {
```
jakebailey, line 368: This should be `string`, I think?

src/compiler/types.ts
weswigham: Actually, since `commentPragmas` is already an `as const` tree of exclusively literal expressions, why does it even have an ID error? The types should be trivially inferrable.

src/compiler/sourcemap.ts
```diff
@@ -360,12 +360,12 @@ export function createSourceMapGenerator(host: EmitHost, file: string, sourceRoo
 
 // Sometimes tools can see the following line as a source mapping url comment, so we mangle it a bit (the [M])
 /** @internal */
-export const sourceMapCommentRegExpDontCareLineStart = /\/\/[@#] source[M]appingURL=(.+)\r?\n?$/;
+export const sourceMapCommentRegExpDontCareLineStart: RegExp = /\/\/[@#] source[M]appingURL=(.+)\r?\n?$/;
```
jakebailey, line 363: This is unfortunate given it can _only_ be `RegExp` but I guess this is what we decided to avoid closing the door on better regex types later.

src/compiler/sys.ts
```diff
@@ -128,7 +128,11 @@ function createPollingIntervalBasedLevels(levels: Levels) {
 const defaultChunkLevels: Levels = { Low: 32, Medium: 64, High: 256 };
 let pollingChunkSize = createPollingIntervalBasedLevels(defaultChunkLevels);
 /** @internal */
-export let unchangedPollThresholds = createPollingIntervalBasedLevels(defaultChunkLevels);
+export let unchangedPollThresholds: {
+    250: number;
```
jakebailey, line 132: This seems dubious; can we derive this type from something else?

src/compiler/transformers/declarations.ts
```diff
@@ -251,7 +251,11 @@ const declarationEmitNodeBuilderFlags = NodeBuilderFlags.MultilineObjectLiterals
  *
  * @internal
  */
-export function transformDeclarations(context: TransformationContext) {
+export function transformDeclarations(context: TransformationContext): {
+    (node: Bundle): Bundle;
```
jakebailey, line 255: This seems weird; do we not have a type for this somewhere? I think this is a regular transformer? I'd expect this to have the same signature as other transformers...

src/compiler/transformers/module/impliedNodeFormatDependent.ts
```diff
@@ -14,7 +14,7 @@ import {
 } from "../../_namespaces/ts.js";
 
 /** @internal */
-export function transformImpliedNodeFormatDependentModule(context: TransformationContext) {
+export function transformImpliedNodeFormatDependentModule(context: TransformationContext): (node: SourceFile | Bundle) => SourceFile | Bundle {
```
jakebailey, line 17: I _think_ we have some sort of type that names this kind of transform?

src/compiler/transformers/ts.ts
```diff
@@ -231,7 +231,7 @@ const enum ClassFacts {
 }
 
 /** @internal */
-export function transformTypeScript(context: TransformationContext) {
+export function transformTypeScript(context: TransformationContext): (node: SourceFile | Bundle) => SourceFile | Bundle {
```
jakebailey, line 234: Similarly here; I think we have a type to use for this.

src/compiler/types.ts
```diff
@@ -8306,62 +8307,80 @@ export type EmitHelperUniqueNameCallback = (name: string) => string;
  * @internal
  */
 export const enum LanguageFeatureMinimumTarget {
+    // These are copied from 'ScriptTarget' since under isolatedDeclarations, enums cannot have references to external symbols.
+    // Make sure these are updated alongside any changes to 'ScriptTarget'.
+    /**@deprecated */
+    ES3 = 0,
```
jakebailey, line 8313: This is odd; can we really not infer these?

src/compiler/types.ts
jakebailey: Yeah, I think that we should find a way to derive this from some other type.

src/compiler/utilities.ts
```diff
@@ -4284,25 +4284,22 @@ export function forEachImportClauseDeclaration<T>(node: ImportClause, action: (d
 }
 
 /** @internal */
-export function hasQuestionToken(node: Node) {
-    if (node) {
```
jakebailey, line 4288: Is this behavior change intentional? It's not annotated to take `Node`, but I would verify we didn't put this here to handle some silly crash. (I would think we should do this change though)

src/compiler/utilities.ts
```diff
@@ -8814,7 +8814,108 @@ function createComputedCompilerOptions<T extends Record<string, CompilerOptionKe
 }
 
 /** @internal */
-export const computedOptions = createComputedCompilerOptions({
+export const computedOptions: {
```
jakebailey, line 8817: This is really really unfortunate. I'm not sure what to do about this one.

src/compiler/utilities.ts
```diff
@@ -11199,7 +11303,14 @@ export function createNameResolver({
         associatedDeclarationForContainingInitializerOrBindingName: ParameterDeclaration | BindingElement | undefined,
         withinDeferredContext: boolean,
     ) => void;
-}) {
+}): (
+    location: Node | undefined,
```
jakebailey, line 11307: This seems like it should be a named type.

src/server/editorServices.ts
```diff
@@ -189,9 +189,9 @@ import {
 } from "./_namespaces/ts.server.js";
 import * as protocol from "./protocol.js";
 
-export const maxProgramSizeForNonTsFiles = 20 * 1024 * 1024;
+export const maxProgramSizeForNonTsFiles: number = 20 * 1024 * 1024;
```
jakebailey, line 192: Is this a bug in isolatedDeclarations? Why is this not inferred?

jakebailey (commented):
Pretty hard to review the whole thing because the diff is so big, but I do wonder if we should even bother doing this for harness/testRunner given those don't actually have a public API...

In general though I do think we should try and name some of the types introduced here. That and there are some really unfortunate losses of inference here :(

src/compiler/types.ts
iisaduan: It doesn't know what to do with the `PragmaKindFlags`

src/compiler/utilities.ts
iisaduan: yes, I intentionally made this change. For all current references to `hasQuestionToken`, a `Node` is always passed in and can't be undefined. I wasn't able to trace the history of changes for this function, but I'm guessing that other things in the function were deleted, and this check was accidentally left in.

src/compiler/utilities.ts
iisaduan: Oops, I missed it in my pass through of const objects :,). Very unfortunate indeed

src/compiler/utilities.ts
```diff
@@ -7598,7 +7598,7 @@ export function readJson(path: string, host: { readFile(fileName: string): strin
 }
 
 /** @internal */
-export function tryParseJson(text: string) {
+export function tryParseJson(text: string): any {
```
iisaduan, line 7601: any

scripts/processDiagnosticMessages.mjs
Andarist: since `as` is a potentially unsafe cast you could make it safer with this beauty: `satisfies Diagnostic as Diagnostic`

src/compiler/core.ts
iisaduan: Hmm, it seems reasonable to me that we could have iterators through values that could be not defined.

iisaduan:

iisaduan:

iisaduan:

src/testRunner/unittests/helpers/contents.ts
```diff
@@ -33,6 +33,9 @@ interface Symbol {
 }
 `;
 
-export function getProjectConfigWithNodeNext(withNodeNext: boolean | undefined) {
+export function getProjectConfigWithNodeNext(withNodeNext: boolean | undefined): {
+    module: string;
+    target: string;
+} | undefined {
```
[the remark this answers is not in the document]
iisaduan: Why does this need to be less specific?

iisaduan:

src/compiler/builder.ts
```diff
@@ -89,6 +89,7 @@ import {
     WriteFileCallback,
     WriteFileCallbackData,
 } from "./_namespaces/ts.js";
+import { BuildInfoFileVersionMap } from "./types.js";
```
jakebailey, line 92: This should be an import from the TS namespace

src/compiler/moduleNameResolver.ts
```diff
@@ -3332,8 +3332,8 @@ function resolveFromTypeRoot(moduleName: string, state: ModuleResolutionState) {
 // Program errors validate that `noEmit` or `emitDeclarationOnly` is also set,
 // so this function doesn't check them to avoid propagating errors.
 /** @internal */
-export function shouldAllowImportingTsExtension(compilerOptions: CompilerOptions, fromFileName?: string) {
-    return !!compilerOptions.allowImportingTsExtensions || fromFileName && isDeclarationFileName(fromFileName);
+export function shouldAllowImportingTsExtension(compilerOptions: CompilerOptions, fromFileName?: string): boolean | undefined {
+    return !!compilerOptions.allowImportingTsExtensions || (fromFileName !== undefined) && isDeclarationFileName(fromFileName);
```
jakebailey, line 3336: This is technically not semantically equivalent, no? Should be `!!fromFileName`?

src/compiler/scanner.ts
```diff
@@ -295,7 +295,7 @@ const charCodeToRegExpFlag = new Map<CharacterCodes, RegularExpressionFlags>([
     [CharacterCodes.y, RegularExpressionFlags.Sticky],
 ]);
 
-const regExpFlagToFirstAvailableLanguageVersion = new Map<RegularExpressionFlags, LanguageFeatureMinimumTarget>([
+const regExpFlagToFirstAvailableLanguageVersion = new Map<RegularExpressionFlags, ScriptTarget>([
```
jakebailey, line 298: Can't be `LanguageFeatureMinimumTarget` anymore?

src/compiler/transformers/declarations.ts
```diff
@@ -213,6 +213,7 @@ import {
     visitNodes,
     VisitResult,
 } from "../_namespaces/ts.js";
+import { DeclarationTransformer } from "../types.js";
```
jakebailey, line 216: This should be imported another way.

src/compiler/types.ts
```diff
@@ -4966,6 +4966,14 @@ export interface EmitTransformers {
     declarationTransformers: readonly TransformerFactory<SourceFile | Bundle>[];
 }
 
+/** @internal */
+
+export interface DeclarationTransformer {
```
jakebailey, line 4971: I'm confused why we need this; is `Transformer<SourceFile | Bundle>` sufficient, like our other transforms?

src/compiler/types.ts
```diff
@@ -10410,3 +10524,11 @@ export interface SyntacticTypeNodeBuilderResolver {
     requiresAddingImplicitUndefined(parameter: ParameterDeclaration | JSDocParameterTag, enclosingDeclaration: Node | undefined): boolean;
     isDefinitelyReferenceToGlobalSymbolObject(node: Node): boolean;
 }
+/** @internal */
+
```
jakebailey, line 10528: Stray empty line here; I've seen a couple of these in the diff.

src/compiler/types.ts
```diff
@@ -4966,6 +4966,14 @@ export interface EmitTransformers {
     declarationTransformers: readonly TransformerFactory<SourceFile | Bundle>[];
 }
 
+/** @internal */
+
```
jakebailey, line 4970: Misc empty line

src/services/exportInfoMap.ts
```diff
@@ -527,7 +527,7 @@ function getIsExcluded(excludePatterns: readonly RegExp[], host: LanguageService
 }
 
 /** @internal */
-export function getIsFileExcluded(host: LanguageServiceHost, preferences: UserPreferences) {
+export function getIsFileExcluded(host: LanguageServiceHost, preferences: UserPreferences): ({ fileName, path }: SourceFile) => boolean {
```
jakebailey, line 530:
```suggestion
export function getIsFileExcluded(host: LanguageServiceHost, preferences: UserPreferences): (sourceFile: SourceFile) => boolean {
```

src/compiler/utilities.ts
jakebailey: I still hope there's something less complicated we can do here; I was hoping for a bunch of overloads on a single function instead, but `convertToTSConfig` iterates over this...

src/compiler/types.ts
iisaduan: These and the incorrect imports are moveToFile bugs. I noticed some as I was making the edits, but I didnt manage to fix all of the extra spaces and imports :,(

src/compiler/types.ts
iisaduan: Someone who is more familiar with these transforms should weigh in here

It didn't seem to me that `Transformer<SourceFile | Bundle>` would be sufficient, because it seemed to be important here that we get the same and narrowest type out of the transform as the type put in. At least some of the other `Transformer<SourceFile | Bundle>` could/should be typed conditionally (if we supported it) so that the input is the same type as the output, and I don't think it's possible to currently improve on those function types without also adding overloads.

src/compiler/utilities.ts
iisaduan: @andrewbranch do you have suggestions?

src/compiler/types.ts
jakebailey: I guess the existing function already has the overload; honestly I'd just move the type to the file where it's used. But, I changed the signature to:

```
function transformRoot(node: SourceFile | Bundle): SourceFile | Bundle {
```

And got no compiler errors, so I don't think it's actually critical.

src/compiler/program.ts
```diff
@@ -982,7 +993,7 @@ function getEmitSyntaxForUsageLocationWorker(file: Pick<SourceFile, "fileName" |
 }
 
 /** @internal */
-export function getResolutionModeOverride(node: ImportAttributes | undefined, grammarErrorOnNode?: (node: Node, diagnostic: DiagnosticMessage) => void) {
+export function getResolutionModeOverride(node: ImportAttributes | undefined, grammarErrorOnNode?: (node: Node, diagnostic: DiagnosticMessage) => void): ModuleKind.CommonJS | ModuleKind.ESNext | undefined {
```
sheetalkamat, line 996:
```suggestion
export function getResolutionModeOverride(node: ImportAttributes | undefined, grammarErrorOnNode?: (node: Node, diagnostic: DiagnosticMessage) => void): ResolutionMode | undefined {
```

src/compiler/program.ts
```diff
@@ -1393,7 +1404,7 @@ export function getImpliedNodeFormatForFileWorker(
     packageJsonInfoCache: PackageJsonInfoCache | undefined,
     host: ModuleResolutionHost,
     options: CompilerOptions,
-) {
+): ModuleKind.CommonJS | ModuleKind.ESNext | Partial<CreateSourceFileOptions> | undefined {
```
sheetalkamat, line 1407:
```suggestion
): ResolutionMode | Partial<CreateSourceFileOptions> | undefined {
```

src/compiler/sys.ts
```diff
@@ -559,17 +559,17 @@ export function getFileWatcherEventKind(oldTime: number, newTime: number) {
 }
 
 /** @internal */
-export const ignoredPaths = ["/node_modules/.", "/.git", "/.#"];
+export const ignoredPaths: string[] = ["/node_modules/.", "/.git", "/.#"];
```
sheetalkamat, line 562:
```suggestion
export const ignoredPaths: readonly string[] = ["/node_modules/.", "/.git", "/.#"];
```

src/compiler/utilities.ts
sheetalkamat: Can this be value on "CommandLineOption" in commandLineParser so its accessible via that?

src/compiler/sys.ts
iisaduan: should I make this `as const` as well?

src/compiler/utilities.ts
andrewbranch: https://github.com/microsoft/TypeScript/commit/17efd55bbc1d2e12f29f2807d2c57a5112bbae5e

src/compiler/commandLineParser.ts
```diff
@@ -248,15 +248,15 @@ const libEntries: [string, string][] = [
...
+export const libs: string[] = libEntries.map(entry => entry[0]);
 
 /**
  * A map of lib names to lib files. This map is used both for parsing the "lib" command line
  * option as well as for resolving lib reference directives.
  *
  * @internal
  */
-export const libMap = new Map(libEntries);
+export const libMap: Map<string, string> = new Map<string, string>(libEntries);
```
jakebailey, line 259: I would recommend making these readonly, while we're here.

rbuckton (approved):

src/compiler/types.ts
```diff
@@ -10189,7 +10224,78 @@ export type PragmaArgumentType<KPrag extends keyof ConcretePragmaSpecs> = Concre
     : never;
 
 /** @internal */
-export type ConcretePragmaSpecs = typeof commentPragmas;
+export interface ConcretePragmaSpecs {
+    readonly "reference": {
```
jakebailey, line 10228: Still wish we could do something simpler here, but we also don't really touch this that much. (I'm personally okay with it becoming a little less safe just to not have this duplication)

src/server/editorServices.ts
jakebailey: I still sort of feel like this is an oops in ID, like, we can const evaluate left shifts, why not multiplication?

jakebailey (approved):

src/server/editorServices.ts
iisaduan: Yup, I'm filing an issue on this

rbuckton: left and right shifts can't overflow `Number.MAX_SAFE_INTEGER`

Zzzen:
Does isolated declarations improve the performance of the language service?

iisaduan:
@Zzzen it shouldn't have any effect on our implementation of the language service. It's possible enabling it in your own code could make some parts of the LS faster in your projects, but we have not benchmarked anything or made ID specific performance improvements for the LS.

PR #59636, opened 2024-08-15
allow `generateTrace` in tsconfig and compiler API

### LINKED ISSUE #59617: add `generateTrace` to tsconfig's `compilerOptions`
### 🔍 Search Terms

generateTrace analyze-trace @typescript/analyze-trace

### ✅ Viability Checklist

### ⭐ Suggestion

Would a PR be accepted to add `generateTrace` ([the existing CLI option](https://www.typescriptlang.org/docs/handbook/compiler-options.html#compiler-options)) to the `tsconfig.json`'s `compilerOptions`?  [Similar to `generateCpuProfile`](https://www.typescriptlang.org/tsconfig/#generateCpuProfile) being in both the CLI and the config.

### 📃 Motivating Example

I am making a monorepo tool that will attempt to generate traces and leverage the [analyze-trace](https://github.com/microsoft/typescript-analyze-trace) tool to give users information about hotspots in their monorepo.  Adding this to the tsconfig makes what this tool is trying to do a lot easier: monorepos may have hundreds of `"typecheck": "tsc"` package.json scripts, and it may not be feasible or desirable to modify them all.

### 💻 Use Cases

metaprogramming for monorepo library diagnositcs

Adding the CLI flag requires more complexity to support different package managers (yarn, pnpm, npm, etc.), whereas if a compilerOption existed, it'd be easier.

In a repo I'm working with that has 500+ packages, I'm using find-and-replace to modify the `typecheck` commands in each package.

dimitropoulos:
Fixes #59617

> [!NOTE]
> _If it's at all possible to fit this in to 5.6 I'd be happy to immediate do whatever is required to make that happen: docs, more tests, diagrams, release notes, a full comedy routine about distributive conditionals, you-just-name-it._

I noticed there was actually already [a baseline test](https://github.com/microsoft/TypeScript/blob/main/tests/baselines/reference/config/showConfig/Shows%20tsconfig%20for%20single%20option/generateTrace/tsconfig.json) for this (despite that it fails config parse validation in the latest release).

### DIFF
--- a/src/compiler/commandLineParser.ts
+++ b/src/compiler/commandLineParser.ts
@@ -437,7 +437,6 @@ export const commonOptionsWithBuild: CommandLineOption[] = [
         name: "generateTrace",
         type: "string",
         isFilePath: true,
-        isCommandLineOnly: true,
         paramType: Diagnostics.DIRECTORY,
         category: Diagnostics.Compiler_Diagnostics,
         description: Diagnostics.Generates_an_event_trace_and_a_list_of_types,
--- a/src/compiler/tsbuildPublic.ts
+++ b/src/compiler/tsbuildPublic.ts
@@ -158,7 +158,7 @@ export interface BuildOptions {
     /** @internal */ extendedDiagnostics?: boolean;
     /** @internal */ locale?: string;
     /** @internal */ generateCpuProfile?: string;
-    generateTrace?: string;
+    /** @internal */ generateTrace?: string;
 
     [option: string]: CompilerOptionsValue | undefined;
 }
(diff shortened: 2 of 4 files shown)


src/compiler/commandLineParser.ts
```diff
@@ -437,7 +437,6 @@ export const commonOptionsWithBuild: CommandLineOption[] = [
         name: "generateTrace",
         type: "string",
         isFilePath: true,
-        isCommandLineOnly: true,
```
sheetalkamat, line 440: Please revert all changes except this since we still don’t want the option to be in api but allowed ok to be in tsconfig

dimitropoulos: no problem, done.

dimitropoulos:
@microsoft-github-policy-service agree company="Vercel"

sheetalkamat (approved):

dimitropoulos:
thanks @sheetalkamat!

Issue #59335, opened 2024-07-17
⚡ Performance: Project service spends excess time cleaning client files when called synchronously


JoshuaKGoldberg:
### Acknowledgement

- [X] I acknowledge that issues using this template may be closed without further explanation at the maintainer's discretion.

### 🔎 Search Terms

project service cleanupProjectsAndScriptInfos openClientFileWithNormalizedPath openClientFile

### 🙁 Actual behavior

TypeScript project services open a file with [`service.openClientFile(filePathAbsolute, ...)`](https://github.com/typescript-eslint/typescript-eslint/blob/d1ddbc21ef2786dd63628a694e18ebaf17db5e4b/packages/typescript-estree/src/useProgramFromProjectService.ts#L67). That internally calls to a [`cleanupProjectsAndScriptInfos`](https://github.com/microsoft/TypeScript/blob/752135eb4046fc820006b41b2f621242f77b6c3e/src/server/editorServices.ts#L4408) to clean up projects, such as removing orphans. There's no way to skip that cleanup.

When using [typescript-eslint's `parserOptions.projectService`](https://typescript-eslint.io/blog/announcing-typescript-eslint-v8-beta#project-service), that cleanup adds (seemingly?) unnecessary overhead per file being linted. 

### 🙂 Expected behavior

When I tried commenting out its contents of `cleanupProjectsAndScriptInfos` locally, the time for a lint run of ~1024 files with the project service speed up by ~15-20%, from ~3.2 seconds to ~2.5 seconds. ⚡ 

**Request:** could we add an option to the `ProjectService` class to skip cleanups, please?

Alternate request: failing that, maybe the cleanups can be debounced and/or delayed in some way? E.g. added to a queue that only executes if the project service is still active?

### Additional information about the issue

On the TypeScript side:

* Point of comparison: https://github.com/JoshuaKGoldberg/TypeScript/commit/2ae3568cf0e425ef50d2986ef8044667b5609f28 would be one way of doing this.
* #58486 is the PR that introduced `cleanupProjectsAndScriptInfos`.

On the typescript-eslint side:

* https://github.com/typescript-eslint/performance is the performance repository showing how to create these profiles.
  * https://github.com/typescript-eslint/performance/blob/main/README.md#comparison-project-service-client-file-cleanups has the measurements for this specific investigation.
* https://github.com/typescript-eslint/typescript-eslint/pull/6754 is the PR that first added usage of the project service in typescript-eslint.
[...]

PR #59645, opened 2024-08-15
Use ProjectService for linting

jakebailey:
This simplifies things, and lets us dogfood the experience.

```console
$  hyperfine -w 3 -n 'before' -p 'git switch main' 'node ./node_modules/eslint/bin/eslint.js .' -n 'after' -p 'git switch eslint-project-service' 'node ./node_modules/eslint/bin/eslint.js .'
Benchmark 1: before
  Time (mean ± σ):     26.980 s ±  0.523 s    [User: 42.924 s, System: 1.743 s]
  Range (min … max):   26.235 s … 28.133 s    10 runs

Benchmark 2: after
  Time (mean ± σ):     27.924 s ±  0.322 s    [User: 44.788 s, System: 1.784 s]
  Range (min … max):   27.451 s … 28.619 s    10 runs

Summary
  'before' ran
    1.04 ± 0.02 times faster than 'after'
```

Slower, but we also have technically enabled typed linting on a handful of new files?

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -52,6 +52,7 @@ export default tseslint.config(
             parserOptions: {
                 warnOnUnsupportedTypeScriptVersion: false,
                 projectService: true,
+                tsconfigRootDir: __dirname,
                 defaultProject: "./scripts/tsconfig.json",
                 allowDefaultProject: ["*.js", "*.cjs", "*.mjs"],
             },
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -51,10 +51,11 @@ export default tseslint.config(
         languageOptions: {
             parserOptions: {
                 warnOnUnsupportedTypeScriptVersion: false,
-                projectService: true,
                 tsconfigRootDir: __dirname,
-                defaultProject: "./scripts/tsconfig.json",
-                allowDefaultProject: ["*.js", "*.cjs", "*.mjs"],
+                projectService: {
+                    defaultProject: "./scripts/tsconfig.json",
+                    allowDefaultProject: ["*.js", "*.cjs", "*.mjs"],
+                },
             },
             globals: globals.node,
         },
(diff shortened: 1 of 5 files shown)


eslint.config.mjs
```diff
@@ -51,6 +51,9 @@ export default tseslint.config(
         languageOptions: {
             parserOptions: {
                 warnOnUnsupportedTypeScriptVersion: false,
+                projectService: true,
+                defaultProject: "./scripts/tsconfig.json",
```
jakebailey, line 55: This file does not cover our misc config files at the root (maybe it should?), but using it here is okay given it's a config with reasonable settings / noEmit.

eslint.config.mjs
```diff
@@ -51,6 +51,11 @@ export default tseslint.config(
         languageOptions: {
             parserOptions: {
                 warnOnUnsupportedTypeScriptVersion: false,
+                tsconfigRootDir: __dirname,
+                projectService: {
+                    defaultProject: path.join(__dirname, "scripts", "tsconfig.json"),
```
jakebailey, line 56: ts-eslint internally is calling `getParsedCommandLineOfConfigFile`, on say `"./scripts/tsconfig.json"`, but then passes in a host which sets the cwd to `path.dirname("./scripts/tsconfig.json")`, which isn't quite right.

Passing in an absolute path avoids that, but I believe there's a bug here in ts-eslint.

jakebailey: (cc @JoshuaKGoldberg)

JoshuaKGoldberg: Nice find! I also suspect there's an additional improvement we should make in reporting a more informative complaint.

DanielRosenwasser:
> Slower, but we also have technically enabled typed linting on a handful of new files?

What would a more apples-to-apples comparison look like?

jakebailey:
I would just ignore the three root files altogether, then compare before and after.

But, those files are small, so I would think the culprit here is actually #59335.

PR #59646, opened 2024-08-15
Enable useDefineForClassFields

jakebailey:
Been meaning to send this one for a bit. Unfortunately, our target is still too low to get this without runtime helpers.

I also had to add `declare` to our object allocators to prevent some very confusing crashes. I suspect that it'd be better to merge the allocators together and then just use defined properties in a specific order to make the shape consistent, which this PR does not do.


DanielRosenwasser:

jakebailey:
This is the diff in `typescript.js`: https://gist.github.com/jakebailey/3626c2be38a6c4f7b77525b78eb70c6b

Because our target doesn't support define fields, esbuild is downleveling them with helpers. That doesn't seem so great.

However, if we instead target _just_ Node 14.17 and not `es2020`, then we do get the syntax.

jakebailey:
Honestly, just saying our minimum is Node 14.17 would probably be acceptable; all browsers can handle this, so it's really just Node, and we already declare `>=14.17`.

PR #59661, opened 2024-08-16
Report implict any error when widening null/undefined in presence of …

### LINKED ISSUE #44913: Promise.catch `null` vs `null as null`
# Bug Report

- This is a crash

### ⏯ Playground Link

### 💻 Code

> Version 4.4.0-dev.20210705

image

```ts
export function f1()
{
	return Promise.resolve(1).catch(e => null)
}

export function f2()
{
	return Promise.resolve(1).catch(e => null as null)
}

// @ts-expect-error
f1().then(v => v.split('')) // <====== should is `number` but `any`

// @ts-expect-error
f2().then(v => v.split(''))

```

```
{
  "compilerOptions": {
    "module": "commonjs",
    "target": "es2019",
    "jsx": "preserve",
    "moduleResolution": "node",
    "allowJs": false,
    "allowUmdGlobalAccess": false,
    "sourceMap": true,
    "inlineSourceMap": false,
    "inlineSources": true,
    "declaration": true,
    "experimentalDecorators": true,
    "emitDecoratorMetadata": true,
    "resolveJsonModule": true,
    "strict": false,
    "strictBindCallApply": true,
    "strictNullChecks": false,
    "strictFunctionTypes": true,
    "noImplicitThis": true,
    "noImplicitOverride": true,
    "preserveConstEnums": true,
    "forceConsistentCasingInFileNames": true,
    "incremental": true,
    "noErrorTruncation": true,
    "removeComments": true,
    "traceResolution": false,
    "newLine": "lf",
    "esModuleInterop": true,
    "importHelpers": true
  }
}

```

### 🙁 Actual behavior

```
// @ts-expect-error
f1().then(v => v.split('')) // <====== should is `number` but `any`
```

### 🙂 Expected behavior

```
// @ts-expect-error
f1().then(v => v.split('')) // <====== should is `number`
```

rbuckton:
This adds a missing implicit-any error when we would infer `any` from the body of a function as the result of widening when that function has a contextual signature whose return type is generic.

This is intended to catch cases like the following:

```ts
// @strictNullChecks: false
// @noImplicitAny: true
declare var p: Promise<number>;
const p2 = p.catch(() => null);
//    ^? p2: Promise<any>;
```

Fixes #44913

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -25513,7 +25513,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             case SyntaxKind.ArrowFunction:
                 if (noImplicitAny && !(declaration as NamedDeclaration).name) {
                     if (wideningKind === WideningKind.GeneratorYield) {
-                        error(declaration, Diagnostics.Generator_implicitly_has_yield_type_0_because_it_does_not_yield_any_values_Consider_supplying_a_return_type_annotation, typeAsString);
+                        error(declaration, Diagnostics.Generator_implicitly_has_yield_type_0_Consider_supplying_a_return_type_annotation, typeAsString);
                     }
                     else {
                         error(declaration, Diagnostics.Function_expression_which_lacks_return_type_annotation_implicitly_has_an_0_return_type, typeAsString);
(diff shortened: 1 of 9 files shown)


rbuckton:
@typescript-bot perf test
@typescript-bot run dt
@typescript-bot test top400

rbuckton:
The other option here would be to not widen `null` or `undefined` in cases like `promise.catch(() => null)`, such that the result is `T | null`.

rbuckton:
> The other option here would be to not widen `null` or `undefined` in cases like `promise.catch(() => null)`, such that the result is `T | null`.

While this seems reasonable, it would make widening inconsistent compared to something like 

```ts
declare function f<T>(value: T): T;
f(null);
```
which, while generic, still widens `null` to `any`.

As a result, I think the approach proposed in this PR is the most reasonable solution.

rbuckton:
@weswigham, @andrewbranch: can you take a look?

src/compiler/checker.ts
```diff
@@ -25535,12 +25535,40 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         errorOrSuggestion(noImplicitAny, declaration, diagnostic, declarationNameToString(getNameOfDeclaration(declaration)), typeAsString);
     }
 
+    function reportErrorsFromWideningWithContextualSignature(declaration: FunctionLikeDeclaration, wideningKind: WideningKind) {
```
jakebailey, line 25538: Rename to `shouldReportErrorsFromWideningWithContextualSignature` given this returns a boolean, it doesn't actually report?

jakebailey (commented):
The linked issue and PR description mention the`.then`/`.catch` case but there doesn't seem to be a test for it; is that intentional?

sandersn (commented):
Looks good. I agree with Jake's naming suggestion.

rbuckton:
> The linked issue and PR description mention the`.then`/`.catch` case but there doesn't seem to be a test for it; is that intentional?

The fix applies regardless as to whether it is a function or method, so the tests cover the smallest general case. I can add a test for `.catch()` though.

jakebailey (approved):

vadzim:
```ts
declare var p: Promise<number>;
const p2 = p.catch(() => null);
//    ^? p2: Promise<any>;
```
but this is not `Promise<any>`, this is `Promise<number | null>`...

jakebailey:
@vadzim

Compare: https://www.typescriptlang.org/play/?strictNullChecks=false#code/PTAEAEGcBcCcEsDG0ByBXANhgwgCwKaIDWkAXKAGYCGGk+AsAFAgQB2A9gJIC2ADhknjQAgqwCe5OGgaMmAE0IYqsfKABuy0L3IAFWO27w6AHlZpuAI3ywAfAG4midqxhaATKAC8WgHSIq0Ii4ABTBAJReNqBmWGEOzGCgSQB6APxAA

To: https://www.typescriptlang.org/play/?strictNullChecks=false&ts=5.8.0-dev.20241119#code/PTAEAEGcBcCcEsDG0ByBXANhgwgCwKaIDWkAXKAGYCGGk+AsAFAgQB2A9gJIC2ADhknjQAgqwCe5OGgaMmAE0IYqsfKABuy0L3IAFWO27w6AHlZpuAI3ywAfAG4midqxhaATKAC8WgHSIq0Ii4ABTBAJReNqBmWGEOzGCgSQB6APxAA

Both are any, and this PR issues the error in the latter.

jeremymeng:
is this error new in 5.7.2 related to this PR? https://www.typescriptlang.org/play/?strictNullChecks=false&ts=5.7.2#code/PTAEAEGcBcCcEsDG0ByBXANhgwgCwKaIDWkAXKAGYCGGk+AsAFAgQB2A9gJIC2ADhknjQAgqwCe5OGgaMm8VtHyxqifKABCoAN5NQe0PIAm+AB74yoVmm4AjJQG0AukwC+TJonasYoG1WiIBBbqTqAAvKD2AIwANABMjgB03FS8ABQA+vDhAHzauvqw+NBosKz5jPpVBqzGZhZOBXpujC4AlEA

jakebailey:
Very unlikely; consider using https://www.npmjs.com/package/every-ts to bisect the change.

jeremymeng:
Thanks @jakebailey TIL

jeremymeng:
Turns out that this is the commit that starting report the error for
```ts
// @strictNullChecks: false
// @noImplicitAny: true

interface B {
  indexes: number[]
}

const batches: B[] = [1,2].map(_i => {
  return {
    // Object literal's property 'indexes' implicitly has an 'any[]' type.(7018)
    indexes: []
  }
})
```


```
25ab551a123fb90ea27b877afc15d0689c5cf0a3 is the first bad commit
commit 25ab551a123fb90ea27b877afc15d0689c5cf0a3
Author: Ron Buckton <ron.buckton@microsoft.com>
Date:   Thu Sep 26 17:06:57 2024 -0400

    Report implict any error when widening null/undefined in presence of … (#59661)
```

jakebailey:
Please file a new issue, then. Thanks.

PR #59665, opened 2024-08-17
Drop unnecessary type arguments in the isolated declarations quick fix

### LINKED ISSUE #58449: Isolated declarations quick fix explicitly prints out default type arguments
```ts
// @declaration: true
// @isolatedDeclarations: true

export interface Foo<T, U = T[]> {
}

export function /**/foo(x: Foo<string>) {
    return x;
}
```

Request a quick fix at `/**/`.

**Current**:

```
Add return type 'Foo<string, string[]>'.
```

**Expected**:

```
Add return type 'Foo<string>'.
```

blickly:
Drops unnecessary type arguments in the isolated declarations quick fix by repeatedly trying to resolve a resulting reference type with a prefix of the full type arguments until finding the minimal number of arguments that work.

Fixes #58449

(I'm sure that the code at present isn't using the proper APIs, but I'd appreciate feedback)


src/compiler/types.ts
```diff
@@ -5031,6 +5031,9 @@ export interface TypeCheckerHost extends ModuleSpecifierResolutionHost {
 }
 
 export interface TypeChecker {
+    
+    /** @internal */
+    fillMissingTypeArguments(typeArguments: readonly Type[], typeParameters: readonly TypeParameter[] | undefined, minTypeArgumentCount: number, isJavaScriptImplicitAny: boolean): Type[];
```
jakebailey, line 5036: Please stick this at the bottom of the interface.

src/services/codefixes/helpers.ts
```diff
@@ -608,6 +619,43 @@ export function typeToAutoImportableTypeNode(checker: TypeChecker, importAdder:
     return getSynthesizedDeepClone(typeNode);
 }
 
+function endOfRequiredTypeParameters(checker: TypeChecker, type: GenericType, fullTypeArguments: readonly Type[]) : number {
+    if (fullTypeArguments !== type.typeArguments!) {
+        throw new Error('fullTypeArguments should be set')
```
jakebailey, line 624: Instead of throwing, use `Debug.fail`. Or maybe `Debug.assertNotEqual` (if that variant exists).

src/services/codefixes/helpers.ts
```diff
@@ -608,6 +619,43 @@ export function typeToAutoImportableTypeNode(checker: TypeChecker, importAdder:
...
+    if (!typeNode) {
+        return undefined;
+    }
+    if (isTypeReferenceNode(typeNode)) {
+        const genericType = type as GenericType;
+        if (genericType.typeArguments) {
+            const cutoff = endOfRequiredTypeParameters(checker, genericType, genericType.typeArguments);
+            if (cutoff !== undefined && typeNode.typeArguments) {
+                // Looks like the wrong way to do this. What APIs should I use here?
+                (typeNode as any).typeArguments = typeNode.typeArguments.slice(0, cutoff);
```
jakebailey, line 651: An `as any` doesn't sound right, no. Is this just so you can mutate?

src/compiler/types.ts
```diff
@@ -5031,6 +5031,7 @@ export interface TypeCheckerHost extends ModuleSpecifierResolutionHost {
 }
 
 export interface TypeChecker {
+
```
jakebailey, line 5034:
```suggestion
```

Style nit 😄

blickly: Fixed

src/services/codefixes/helpers.ts
blickly: Right, I thought it seemed wasteful to throw away the newly created TypeNode, but I'm happy to change if there's a better recommendation here.

src/compiler/types.ts
blickly: Done.

src/services/codefixes/helpers.ts
blickly: I simplified the signature here so it can just use Debug.assert

src/services/codefixes/helpers.ts
jakebailey: I would probably just update the node via the factory like we normally do in fixes.

blickly: Makes sense. Done

blickly:
OK, I switched to the more idiomatic way of creating a new AST node, rather than trying to mutate anything. PTAL

src/services/codefixes/helpers.ts
```diff
@@ -608,6 +618,42 @@ export function typeToAutoImportableTypeNode(checker: TypeChecker, importAdder:
...
+export function typeToMinimizedReferenceType(checker: TypeChecker, type: Type, contextNode: Node | undefined, flags?: NodeBuilderFlags, internalFlags?: InternalNodeBuilderFlags, tracker?: SymbolTracker): TypeNode | undefined {
+    let typeNode = checker.typeToTypeNode(type, contextNode, flags, internalFlags, tracker);
+    if (!typeNode) {
+        return undefined;
+    }
+    if (isTypeReferenceNode(typeNode)) {
+        const genericType = type as GenericType;
+        if (genericType.typeArguments) {
+            const cutoff = endOfRequiredTypeParameters(checker, genericType);
+            if (cutoff !== undefined && typeNode.typeArguments) {
```
jakebailey, line 648: `cutoff` can't be undefined; should this instead be a check that skips this block if the length is different than the existing node?

tests/cases/fourslash/codeFixMissingTypeAnnotationOnExports50-generics-with-default.ts
```diff
@@ -0,0 +1,17 @@
...
+// @file lib.d.ts
+////interface MyIterator<T, TReturn = any, TNext = undefined> {}
+////let x: MyIterator<number>;
+////export const y = x;
+
+verify.codeFix({
+    description: "Add annotation of type 'MyIterator<number>'",
+    index: 0,
+    newFileContent:
+`interface MyIterator<T, TReturn = any, TNext = undefined> {}
```
jakebailey, line 14: Do we have tests for generators elsewhere? I seem to recall us auto-adding type long type annotations for those...

src/services/codefixes/helpers.ts
blickly: Good catch. Done.

tests/cases/fourslash/codeFixMissingTypeAnnotationOnExports50-generics-with-default.ts
blickly: Sure, I can add a case with Generator.

For reasons I don't fully understand, `Generator<T>` and `Iterator<T>` get generated with the full 3 type arguments, whereas `Iterable<T>` and `IterableIterator<T>` are special cased inside the TypeChecker to omit the extra arguments when unnecessary.

The special casing inside `typeReferenceToTypeNode`:
```
                            // Maybe we should do this for more types, but for now we only elide type arguments that are
                            // identical to their associated type parameters' defaults for `Iterable`, `IterableIterator`,
                            // `AsyncIterable`, and `AsyncIterableIterator` to provide backwards-compatible .d.ts emit due
                            // to each now having three type parameters instead of only one.
                            if (
                                isReferenceToType(type, getGlobalIterableType(/*reportErrors*/ false)) ||
                                isReferenceToType(type, getGlobalIterableIteratorType(/*reportErrors*/ false)) ||
                                isReferenceToType(type, getGlobalAsyncIterableType(/*reportErrors*/ false)) ||
                                isReferenceToType(type, getGlobalAsyncIterableIteratorType(/*reportErrors*/ false))
                            ) {
                                if (
                                    !type.node || !isTypeReferenceNode(type.node) || !type.node.typeArguments ||
                                    type.node.typeArguments.length < typeParameterCount
                                ) {
                                    while (typeParameterCount > 0) {
                                        const typeArgument = typeArguments[typeParameterCount - 1];
                                        const typeParameter = type.target.typeParameters[typeParameterCount - 1];
                                        const defaultType = getDefaultFromTypeParameter(typeParameter);
                                        if (!defaultType || !isTypeIdenticalTo(typeArgument, defaultType)) {
                                            break;
                                        }
                                        typeParameterCount--;
                                    }
                                }
                            }
```

jakebailey: That was a part of #58243.

tests/cases/fourslash/codeFixMissingTypeAnnotationOnExports54-generator-generics.ts
```diff
@@ -0,0 +1,17 @@
...
+// @lib: es2015
+//// export function foo(x: Generator<number>) {
+////     return x;
+//// }
+
+verify.codeFix({
+    description: "Add return type 'Generator<number>'",
+    index: 0,
+    newFileContent:
+`export function foo(x: Generator<number>): Generator<number> {
```
jakebailey, line 14: To be clear I was meaning to write `function *foo() { }` here and then see what happens.

blickly: Ah, OK, added.

It's not technically related to the rest of this PR since it gets inferred a precise return that needs all 3 type arguments, but it's a good case to include in the test suite anyway.

tests/cases/fourslash/codeFixMissingTypeAnnotationOnExports50-generics-with-default.ts
blickly: Makes sense. Thanks for the context.

jakebailey (approved):
LGTM I think, but would be good to get a couple of other eyes on it.

blickly:
Any advice on who would make a good secondary reviewer?

src/services/codefixes/helpers.ts
```diff
@@ -608,6 +618,42 @@ export function typeToAutoImportableTypeNode(checker: TypeChecker, importAdder:
...
+    const fullTypeArguments = type.typeArguments;
+    const target = type.target;
+    next_cutoff: for (let cutoff = 0; cutoff < fullTypeArguments.length; cutoff++) {
+        const typeArguments = fullTypeArguments.slice(0, cutoff);
+        const filledIn = checker.fillMissingTypeArguments(typeArguments, target.typeParameters, cutoff, /*isJavaScriptImplicitAny*/ false);
+        for (let i = 0; i < filledIn.length; i++) {
+            // If they don't match, then we haven't yet reached the right cutoff
+            if (filledIn[i] !== fullTypeArguments[i]) continue next_cutoff;
+        }
+        return cutoff;
```
sandersn, line 632:
```suggestion
        if (filledIn.every((fill, i) => fill === fullTypeArguments[i])) {
            return cutoff;
        }
```

and then you can drop the `next_cutoff` label

src/services/codefixes/helpers.ts
```diff
@@ -608,6 +618,42 @@ export function typeToAutoImportableTypeNode(checker: TypeChecker, importAdder:
...
+
+/** @internal */
+export function typeToMinimizedReferenceType(checker: TypeChecker, type: Type, contextNode: Node | undefined, flags?: NodeBuilderFlags, internalFlags?: InternalNodeBuilderFlags, tracker?: SymbolTracker): TypeNode | undefined {
+    let typeNode = checker.typeToTypeNode(type, contextNode, flags, internalFlags, tracker);
+    if (!typeNode) {
+        return undefined;
+    }
+    if (isTypeReferenceNode(typeNode)) {
+        const genericType = type as GenericType;
+        if (genericType.typeArguments && typeNode.typeArguments) {
```
sandersn, line 646: I don't understand why both of these need to be checked. If genericType.typeArguments is set, then isn't typeNode.typeArgument guaranteed to be filled in completely? I thought that was the bug. Or did I misunderstand, and sometimes typeToTypeNode doesn't fill out the arguments completely?

tests/cases/fourslash/codeFixMissingTypeAnnotationOnExports52-gerenrics-oversimplification.ts
```diff
@@ -0,0 +1,23 @@
+/// <reference path='fourslash.ts'/>
```
sandersn, line 1: typo in filename: gerenrics

tests/cases/fourslash/codeFixMissingTypeAnnotationOnExports52-gerenrics-oversimplification.ts
```diff
@@ -0,0 +1,23 @@
...
+
+// In the abstract, we might prefer the inferred return type annotation to
+// be identical to the parameter type (with 2 type parameters).
+// Our current heursitic to avoid overly complex types in this case creates
+// "overly simple" types, but this tradeoff seems reasonable.
```
sandersn, line 6: :+1:

tests/cases/fourslash/codeFixMissingTypeAnnotationOnExports52-gerenrics-oversimplification.ts
```diff
@@ -0,0 +1,23 @@
+/// <reference path='fourslash.ts'/>
+
+// In the abstract, we might prefer the inferred return type annotation to
+// be identical to the parameter type (with 2 type parameters).
+// Our current heursitic to avoid overly complex types in this case creates
```
sandersn, line 5:
```suggestion
// Our current heuristic to avoid overly complex types in this case creates
```

sandersn (changes requested):

tests/cases/fourslash/codeFixMissingTypeAnnotationOnExports52-gerenrics-oversimplification.ts
blickly: Good catch. Fixed

src/services/codefixes/helpers.ts
blickly: No, you're right.  If `genericType.typeArguments` is defined then `typeNode.typeArguments` will be as well.  I think I'm just using the check to signal to the type system that fact (so I don't need to use `typeNode.?typeArguments` later), but wasn't sure if there is a way to do assertions in the TS codebase that might fail faster and more clearly.

jakebailey:
Tested locally with main and it merges and passes cleanly.

PR #59666, opened 2024-08-17
Add support for "specialize" tag

### LINKED ISSUE #27387: Allow to explicitly pass parameter types via JSDoc
## Search Terms

jsdoc generics constraints

## Suggestion

It seems like it is not possible via JSDoc to explicitly tell compiler which type parameters to pass to a generic function.

## Use Cases

In TypeScript it is possible to explicitly pass type parameters such as `mongoose.model<Model, Schema>('modelName', Schema)` while I could not find a way to do same with JSDoc.

## Examples

`mongoose.model` has two signatures, first one with one type parameter and the other with two. To make use of the second signature we must pass types explicitly.

```ts
export function model<T extends Document>(name: string, schema?: Schema, collection?: string,
    skipInit?: boolean): Model<T>;
  export function model<T extends Document, U extends Model<T>>(
    name: string,
    schema?: Schema,
    collection?: string,
    skipInit?: boolean
  ): U;
```
```ts
//  Choose second signature in typescript.
const model = mongoose.model<Model, Schema>('modelName', Schema);

// With JSDoc, compiler always chooses first signature and we receive a type mismatch error.
/** @type {Schema & mongoose.Model<Model>} */
const model = mongoose.model('modelName', Schema);

// But something like this would be great.
const model = mongoose.model/**<Model, Schema>*/('modelName', Schema);
```

My apologies if this is already possible, but I've spend almost a week battling this.

Related: https://github.com/Microsoft/TypeScript-Node-Starter/issues/101

## Checklist

* [?] This wouldn't be a breaking change in existing TypeScript / JavaScript code
* [?] This wouldn't change the runtime behavior of existing JavaScript code
* [?] This isn't a runtime feature (e.g. new expression-level syntax)

apendua:
Fixes #27387 

This is an implementation proposal for `@specialize` JSDoc tag, which will allow specifying type arguments in JS files. For example if `f` has a type parameter `T`, the following JSDoc syntax:
```javascript
let x = /** @specialize {number} */(f());
```
will be equivalent to TS:
```typescript
let x = f<number>();
```
The tag can be also attached to any other parent node, e.g. assignment statement would also do:
```javascript
/** @specialize {number} */
let x = f();
```
Currently, it is also possible to wrap type arguments with `<...>` as in `/** @specialize <number> */`,
if that's the preference. I couldn't make up my mind so I allowed both options for now.

The following expressions are also supported:
- tagged template expression
- "new" expression
- JSX element and self-closing tag

I am still planning to add more unit tests, e.g.
- check if "find all references" works
- test JSX syntax

Having said that, I've decided to open this PR early (as draft), because there are some open questions where I seek your advice:
- Is `@specialize` a good name for this tag? I was also considering `@typeArgs`, which I think would be a little bit cleaner in communicating its purpose, but I don't like the fact that it consists of two words, as so far there has been no precedence.
- One issue with `@specialize` is the fact that some people would prefer to spell it "specialise" and I am worried that it may cause confusion and typing errors in the future.


Gerrit0:
> Currently, it is also possible to wrap type arguments with <...> as in `/** @specialize <number> */`,
if that's the preference. I couldn't make up my mind so I allowed both options for now.

Please no. That isn't consistent with how any other type comments are parsed, the extra edge case doesn't just affect TS. `@specialize {number}` is consistent with `@type`

apendua:
I've resolved the conflicts and reverted the support for `<`, `>` delimiters.

apendua:
@DanielRosenwasser Do you think there's a chance this will be re-considered at some point?

PR #59672, opened 2024-08-18
Fix Callback argument type not inferred for union of interfaces #59309

### LINKED ISSUE #59309: Callback argument type not inferred for union of interfaces
### 🔎 Search Terms

`"implicitly has an" inference`

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about "inference"

### ⏯ Playground Link

https://www.typescriptlang.org/play/?target=5&ts=5.5.3#code/KYDwDg9gTgLgBASwHY2FAZgQwMbDgQQGcBPJbAJWEIFcAbGAYU1toCMcBrAHgBUAaOAFE4AXiFQo0AHxwA3gFgAUHBVwAFGigB+AFxC4AHzhI6tAVCp0YuuDwCUegG4QEAEwDcSgL5KloSLCIKGhYuAQkZJQ09ACSqFCYMNC8AuQCwmKCEtJySqrqCKgAtnr8cNjMbJx6RKQUlvRMLOzY3GlCUg5wzm6eij6K-tDwyPGheLWRDTBxaInQAAqSRQiEwClw5DIKyqpqhcAltl1LECtrXFt9A0quwNi0mBZw6NRkMAgQSHBFmGAAMggVjANu0MuJJFApGo8qonlBSgBtAC6fFhKloQMKehMRVYaDRuxUBwSSQRKkm9WiM3i8ygp3O6zKW0M4TqUSss1JyWZ6SkhJOy1W63IKKkfSUvwBWJgakRAEY+AAmPgAZlRcFVAkwEWw6iQdlE23RcAsMGoUG+3wAVNa4ErrnZ3EA

### 💻 Code

```ts
export interface AsyncResultCallback<T, E = Error> {
    (err?: E | null, result?: T): void;
}

export interface AsyncResultIterator<T, R, E = Error> {
    (item: T, callback: AsyncResultCallback<R, E>): void;
}
export interface AsyncResultIteratorPromise<T, R> {
    (item: T): Promise<R>;
}

declare function mapLimit<T, R, E = Error>(
    arr: T[],
    limit: number,
    iterator:  AsyncResultIteratorPromise<T, R> | AsyncResultIterator<T, R, E>,
): Promise<R[]>;

mapLimit([1,2,3], 3, async (n) => {
    return n ** 2;
});
```

### 🙁 Actual behavior

The type of `n` in the `mapLimit` callback is inferred as `any`:

```
Parameter 'n' implicitly has an 'any' type. (7006)
```

### 🙂 Expected behavior

I would expect the type of `n` to be inferred as `number`.

This example is based on the types for the `async` package:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/b9820410f85f259cbad6065d8cb6f23f403109ac/types/async/index.d.ts#L321-L325

Note that if the type for `iterator` in `mapLimit` is changed to either just `AsyncResultIteratorPromise<T, R>` or just `AsyncResultIterator<T, R, E>` (that is, no longer a union type), things work as expected and `n` has type `number`.

I found a very old discussion of what looks to be the same issue in the `DefinitelyTyped` repo: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/24297. The issue described there is exactly the one I'm facing. One commenter claimed that things worked correctly, and another commenter claimed that changing the way that `async` was imported had some effect? Neither of those sounds plausible to me.

I found some issues that may describe the same or a similar problem:

- https://github.com/microsoft/TypeScript/issues/56881
- https://github.com/microsoft/TypeScript/issues/55632
- https://github.com/microsoft/TypeScript/issues/52047
- https://github.com/microsoft/TypeScript/issues/47599

However, my knowledge of TypeScript internals/terminology isn't strong enough to say for sure that this isn't a novel issue I'm facing. If this is in fact a duplicate, please close it as such and I'll add this example to the relevant issue. Thank you!

MichalMarsalek:
Please verify that:
* [x] There is an associated issue in the `Backlog` milestone (**required**)
* [x] Code is up-to-date with the `main` branch
* [x] You've successfully run `hereby runtests` locally
* [x] There are new or updated unit tests validating the change

Fixes #59309


src/compiler/checker.ts
```diff
@@ -32423,22 +32423,23 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             return getContextualCallSignature(type, node);
         }
         let signatureList: Signature[] | undefined;
-        const types = (type as UnionType).types;
+        const types = (type as UnionType).types
+            .map(type => ({ type, signature: getContextualCallSignature(type, node) }))
+            .filter(type => type.signature)
```
Andarist, line 32428: Do you need to pass through `type` here? If I'm not mistaken, you could simplify this to:
```ts
const signatures = mapDefined((type as UnionType).types, t => getContextualCallSignature(t, node))
if (!signatures.length) {
  return;
}
const signatureList = signatures[0];
for (let i = 1; i < signatures.length; i++) {
  // loop body
}
```

src/compiler/checker.ts
```diff
@@ -32423,22 +32423,23 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
-                else {
-                    // Use this signature for contextual union signature
-                    signatureList.push(signature);
-                }
+            const signature = current.signature!;
+            if (!signatureList) {
+                // This signature will contribute to contextual union signature
+                signatureList = [signature];
+            }
+            else if (!compareSignaturesIdentical(signatureList[0], signature, /*partialMatch*/ true, /*ignoreThisTypes*/ true, /*ignoreReturnTypes*/ true, compareTypesSubtypeOf)) {
```
Andarist, line 32436: If you are loosening here how parameter types are meant to be compared I'd probably expect here:
```suggestion
            else if (!compareSignaturesIdentical(signatureList[0], signature, /*partialMatch*/ true, /*ignoreThisTypes*/ true, /*ignoreReturnTypes*/ true, compareTypesAssignable)) {
```

Unless there is already a reason why the subtype relationship was used here.

src/compiler/checker.ts
```diff
@@ -32423,22 +32423,23 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             return getContextualCallSignature(type, node);
         }
         let signatureList: Signature[] | undefined;
-        const types = (type as UnionType).types;
+        const types = (type as UnionType).types
+            .map(type => ({ type, signature: getContextualCallSignature(type, node) }))
+            .filter(type => type.signature)
+            .sort((t1, t2) => t1.signature!.parameters.length - t2.signature!.parameters.length);
```
Andarist, line 32429: You need to account for rest parameters here so it would be more appropriate to use `getParameterCount` here over directly checking the `.parameters.length`

src/compiler/checker.ts
MichalMarsalek: You are right, I was able to simplify, thank you.

src/compiler/checker.ts
MichalMarsalek: I'm not really sure when each relation should be used. I used `compareTypesSubtypeOf` because that's what is used in a call to `compareSignaturesIdentical` with `partialMatch === true` in the function `findMatchingSignatures`.

src/compiler/checker.ts
MichalMarsalek: Thank you. I think that actually what I need is `getMinArgumentCount`.
For the case `((a: string, b?: string) => void) | ((a: string, ...rest: string[]) => void)` I don't think it matters which one is first.

jakebailey:
@typescript-bot test it

Andarist:
The reported error in `webpack` is correct. Right now it errors earlier (which is not mentioned by this comment): [TS playground](https://www.typescriptlang.org/play/?ts=5.6.0-dev.20240818&filetype=ts#code/JYWwDg9gTgLgBAbzgYQgOwGbAOYFcoCGMw6cAvnBlBCHAEQDuApgEZgEDGA1nQNwBQ-DugDO8YZhwAuFOix5CxUgF5E-OHAhhiIYAC8iJNDITqNcCRyJM01gLIQAJrgA2TETJhRcTADRmyfw0mAA8YJihbFw84AG0zDQIRAE80DjgACiQJcLDfOCgmAEcfMXzsJhgAJXcIFwA3JnIASjhlAD41c3NgDEzCkvd4AEJlVTpQ8MiCFzpWwph8NEoZkSYBbo0JMQLahqbVCuq9xuGM5o3N7fhCkTrGxza4AgYCYBuTpgycyfyB0pgFwS5gWSzgAAN6gQoHAACQIABSAGUAPIAOQAdGIoMA0NheskMrd7kxHM0yODLhpAmYALr+Mi8IA). It's enough to add an extra unused parameter to this function to get the same error as the one reported by the bot.

That said, maybe you could also filter out here signatures that can't be applicable in the first place. Like, if the function is `async` then only functions that return `Promise<any>` are valid as contextual signatures for it. Note that the contextual signature could still return a union containing a type assignable to `Promise<any>`. The same principle applies to generators (and async generators). I've done similar filtering on contextual return types [here](https://github.com/microsoft/TypeScript/pull/51196). That patch would still be relevant, I think, but some contextual signatures could even be discarded earlier

tests/cases/compiler/contextualSignatureWithExtraParameters.ts
```diff
@@ -0,0 +1,55 @@
+// https://github.com/microsoft/TypeScript/issues/59309
```
Andarist, line 1: tip: I almost always include `// @strict: true` at the top of the test file unless I want to test non-strict behavior specifically. I already lost too many hours hunting down mismatched actual results from test files just to learn that the test was OK, it just wasn't executed with `strictNullChecks` or another strict option ;p

SImilarly, I usually use `@noEmit: true` to avoid `.js` baselines being generated redundantly if I only want to test type-checking behavior and not js/dts emit.

src/compiler/checker.ts
Andarist: I'm not sure if it will make any difference but to exercise even more funky test cases you could add test cases with signatures of this shape:
```ts
((a: string, ...rest: [string, number?] | [number, boolean?]) => void) | ((a: string, ...rest: [string, number] | [number, boolean]) => void)
```

Note that I didn't think through what types you should use in those rest parameters, just that they can be unions of tuples and that one of the signatures might have some of the elements there required and some might have them optional etc.

src/compiler/checker.ts
Andarist: `findMatchingSignatures` gets only called by `getUnionSignatures` so I agree that this probably makes sense

MichalMarsalek:
> Like, if the function is `async` then only functions that return `Promise<any>` are valid as contextual signatures for it.

That's smart, I added that.

src/compiler/checker.ts
```diff
@@ -32422,29 +32422,24 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
-                    // Signatures aren't identical, do not use
-                    return undefined;
-                }
-                else {
-                    // Use this signature for contextual union signature
-                    signatureList.push(signature);
-                }
+
+        let signatures = mapDefined((type as UnionType).types, t => getContextualCallSignature(t, node));
+        if (hasSyntacticModifier(node, ModifierFlags.Async)) {
```
Andarist, line 32427: similar treatment should be applied to generators

src/compiler/checker.ts
```diff
@@ -32422,29 +32422,24 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+            signatures = signatures.filter(s => !!getAwaitedTypeOfPromise(getReturnTypeOfSignature(s)));
```
Andarist, line 32428: this might not be correct for async generators since they have `ModifierFlags.Async` but are not expected to return a type assignable to `Promise<any>`

Andarist, line 32428: you should prefer using `filter` import - it avoids redundant array allocations when possible

src/compiler/checker.ts
MichalMarsalek: Done. Thank you for the link to you PR, I'd have zero idea how to do this otherwise.

tests/cases/compiler/contextualSignatureWithExtraParameters.ts
```diff
@@ -0,0 +1,77 @@
...
+function f5(
+    cb: ((a: string, b: string) => void) | ((...rest: string[]) => void)
+  ) {}
+  
+f5((a) => {});
+f5((a, b) => {});
+
+function f6(
+  cb: ((a: string) => string) | ((a: number) => Promise<number>)
+) {}
```
Andarist, line 62: in a case like this, `cb` is not callable inside the function's body (it expects an argument of type `never`). That doesn't have to prevent contextual parameter typing since the manually-annotated parameter is still OK:
```ts
function f6(cb: ((a: string) => string) | ((a: number) => Promise<number>)) {
  cb(1)
}

f6(async (a: number) => {
  return 42;
});
```

(I'm just trying to defend (from myself) my previous suggestion to discriminate those contextual signatures based on the function's modifiers etc 😉 )

MichalMarsalek: Yeah the fact that it's not callable without further narrowing doesn't mean it's illegal to pass such a function.

Issue #59668, opened 2024-08-17
Signatures with less parameters aren't assignable to compatible targets with more when their rest param is an instantiated `NoInfer`


Andarist:
### 🔎 Search Terms

NoInfer signature parameters list rest

### 🕗 Version & Regression Information

- This is the behavior in every version I tried

### ⏯ Playground Link

https://www.typescriptlang.org/play/?ts=5.6.0-dev.20240817#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXzCgggB4BBeEADwxFWAGd44pg8IBPeNAa1RwDuqANoBdAHwAKAFDx4sAOYAueJIB0GxQxUA5HAElUiEDHLiAlPAC84+ADccWYABpZ8DWq0qy0894Dc0mB4DBjwWEYmcMDWBEQQkpJQKqjIALYARiaWNvAA3gC+zvAAjMUATOaBAPTVcvUNjfAAegD80tKgkLAIKOjYeHHE5eSUNHSMzCCs7Fy8-EJiUm6KKuqaMAra8GQ5tg5OrnIeXju+AUEhYRHGMNHlsYTDJMKpmSbFb1kwEonJ8F9stZbIVimV4JUanV6m1pEA

### 💻 Code

```ts
declare function call<A extends readonly unknown[]>(
  arg: (...args: NoInfer<A>) => void,
  ...args: A
): A;

// Argument of type '(a: number) => void' is not assignable to parameter of type '(...args: NoInfer<[number, number]>) => void'.
//   Types of parameters 'a' and 'args' are incompatible.
//     Type '[number, number]' is not assignable to type '[a: number]'.
//       Source has 2 element(s) but target allows only 1.(2345)
const inferred = call((a: number) => {}, 1, 2);
//               ^? function call<[number, number]>(arg: (...args: NoInfer<[number, number]>) => void, args_0: number, args_1: number): [number, number]


declare function call2<A extends readonly unknown[]>(
  arg: (...args: A) => void,
  ...args: A
): A;
const inferred2 = call2<[number, number]>((a: number) => {}, 1, 2);
//    ^? const inferred2: [number, number]
```

### 🙁 Actual behavior

The first call fails to typecheck

### 🙂 Expected behavior

[...]

PR #59682, opened 2024-08-19
Normalize `NoInfer`red tuple types

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/59668

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -12192,14 +12192,19 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             type = tryGetTypeFromEffectiveTypeNode(declaration) || checkObjectLiteralMethod(declaration, CheckMode.Normal);
         }
         else if (
-            isParameter(declaration)
-            || isPropertyDeclaration(declaration)
+            isPropertyDeclaration(declaration)
             || isPropertySignature(declaration)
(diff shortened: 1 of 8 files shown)


src/compiler/checker.ts
```diff
@@ -12192,15 +12192,20 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
             || isVariableDeclaration(declaration)
             || isBindingElement(declaration)
             || isJSDocPropertyLikeTag(declaration)
         ) {
             type = getWidenedTypeForVariableLikeDeclaration(declaration, /*reportErrors*/ true);
         }
+        else if (isParameter(declaration)) {
+            type = getWidenedTypeForVariableLikeDeclaration(declaration, /*reportErrors*/ true);
+            if (declaration.dotDotDotToken) {
+                type = normalizeNoInferSpread(type);
```
jakebailey, line 12206: Maybe I'm thinking about this wrong, but shouldn't this actually be a part of `expandParameters` or something? IIRC that's where we normalize these out. Compared to sprinkling it around elsewhere.

Andarist: `getExpandedParameters` is only used by `signatureToSignatureDeclarationHelper` and node builder. This change has behavioral changes for inference etc so having this `getExpandedParameters` wouldn't work

PR #59685, opened 2024-08-19
[experiment] Never treat properties of `export=` type as module exports for auto-imports

andrewbranch:


andrewbranch:
@typescript-bot perf test tsserver-only

andrewbranch:
@typescript-bot perf test tsserver-only

PR #59683, opened 2024-08-19
Try not allowing string iteration

RyanCavanaugh:
Trial for #59676 to see what happens in real world code

PR #59686, opened 2024-08-19
Fix Set of one string literal bug in openExternalProject

jakebailey:
No test changed, which is... interesting.

Noted in https://github.com/microsoft/TypeScript/pull/59683#issuecomment-2297157267


sheetalkamat (approved):

sheetalkamat:
> No test changed, which is... interesting. 

this is just is fast path to not iterate over projects to keep for this external project. Without this change, we will iterate and retain configured projects for this external project so no change as such in baseline

PR #59692, opened 2024-08-19
Don’t try to recreate AutoImportProvider after giving up because of too many dependencies until program structure changes

andrewbranch:
I noticed the following sequence in a TS Server log:

```
Info 1254 [11:44:16.297] getCompletionsAtPosition: isCompletionListBlocker: 0.037459000002854737
Info 1255 [11:44:16.328] AutoImportProviderProject: attempted to add more than 10 dependencies. Aborting.
Info 1256 [11:44:16.328] getExportInfoMap: cache miss or empty; calculating new results
Info 1257 [11:44:16.399] AutoImportProviderProject: attempted to add more than 10 dependencies. Aborting.
Info 1258 [11:44:16.399] getExportInfoMap: done in 97.71966699999757 ms
Info 1259 [11:44:16.403] AutoImportProviderProject: attempted to add more than 10 dependencies. Aborting.
```

`host.getPackageJsonAutoImportProvider()` gets called multiple times during completions. When an AutoImportProvider program exists, it’s cached and updated appropriately. But when one fails to be created because either it finds no dependencies or too many dependencies, the result of that failure wasn’t being stored; instead, we tried to recreate it every time we attempt to get it from the host.

View commit d44feecd776c54ac42c83f2cad6e47abe8c51139 in isolation to see the desired effect in the server logs.


src/server/project.ts
```diff
@@ -1376,8 +1376,11 @@ export abstract class Project implements LanguageServiceHost, ModuleResolutionHo
     /** @internal */
     onPackageJsonChange() {
         this.moduleSpecifierCache.clear();
-        if (this.autoImportProviderHost) {
-            this.autoImportProviderHost.markAsDirty();
+        if (this.autoImportProviderHost === false) {
+            this.autoImportProviderHost = undefined;
+        }
+        else {
+            this.autoImportProviderHost?.markAsDirty();
```
sheetalkamat, line 1383:
```suggestion
        this.markAutoImportProviderAsDirty()
```

sheetalkamat (approved):

sheetalkamat (approved):

PR #45080, opened 2021-07-16
Detect preference for `node:`-prefixed node core modules

andrewbranch:
Fixes #43842

1. It was actually really easy and cheap to detect the presence of `node:` module specifiers as we build the program, so this _does_ use other files as a backup heuristic, despite my hesitance about that in the design meeting. It currently prefers the `node:` versions if it sees any imports/re-exports of that flavor in non-node_modules, non-declaration files. It would be trivial to change that to a more conservative strategy of preferring the `node:` flavor if it finds some `node:`-prefixed imports/re-exports **and no unprefixed ones**, if we’re more comfortable with that.
2. The _first_ recognized node core module import in a file sets the preferred style for that file. Other files are only considered for files that have no node core module imports yet.
3. For **completions**, only the preferred style  is shown. For **codefixes**, both styles are shown, but the preferred style is sorted higher. (There is a precedent for codefixes offering a more granular and complete set of modules to import from.)

PR #59702, opened 2024-08-20
Simplify handling of `node:`-prefixed modules in auto-imports

andrewbranch:
Our previous behavior is described in https://github.com/microsoft/TypeScript/pull/45080. The new behavior implemented here is simpler:

- Preference inference: any `node:`-prefixed import that is not one of the built-ins only available with the prefix (e.g. `node:test`) indicates a preference for using the prefix. If no built-ins are referenced besides the ones only available with the prefix, no preference is set. Otherwise, the preference is to avoid the prefix. When a preference can be inferred from the imports in a single file, that will be used for auto-imports in that file. Otherwise, the same algorithm is used for all imports in non-node_modules, non-declaration files.
- Both codefixes and completions only offer auto-imports matching the inferred preference. If no preference could be inferred, both variants are offered. (Generally, this means new projects will see both options the first time they auto-import a built-in module, and then never again.)


andrewbranch:
@typescript-bot perf test

sheetalkamat (approved):
Change looks good to me but pls fix the knip issue

Issue #59655, opened 2024-08-16
Long running encodedSemanticClassifications-full request


roryscot:
### 🔎 Search Terms

@RyanCavanaugh [requested](https://github.com/microsoft/TypeScript/issues/54459#issuecomment-2293800187) that I create a new issue

### 🕗 Version & Regression Information

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about 5.5.4

### 💻 Code

- Clone [example repo](https://github.com/roryscot/tsbugexample/tree/main)
- run `npm i`
- navigate to [line 26](https://github.com/roryscot/tsbugexample/blob/63a1f479c8f91d70d41147b3d6d4d7e542e0e8f4/src/components/UserProfile/UserProfile.tsx#L26) in VSCode
- try to use auto-suggestion menu (`Command + .`) 

The warning `React Hook useEffect has a missing dependency: 'getProfile'. Either include it or remove the dependency array.eslint[react-hooks/exhaustive-deps](https://github.com/facebook/react/issues/14920)` should appear on [line 26](https://github.com/roryscot/tsbugexample/blob/63a1f479c8f91d70d41147b3d6d4d7e542e0e8f4/src/components/UserProfile/UserProfile.tsx#L26)

### 🙁 Actual behavior

[...]

PR #59705, opened 2024-08-21
[experiment] Bail out early in union production if too many literals/templates

jakebailey:
Motivated by #59655 and similar ☹️


jakebailey:
@typescript-bot test it

Issue #59656, opened 2024-08-16
5.6 regression: Incorrect param type inference for type with all optional props


trevorade:
### 🔎 Search Terms

inference, param

### 🕗 Version & Regression Information

- This changed between versions 5.5.4 and 5.6.0

### ⏯ Playground Link

https://www.typescriptlang.org/play/?ts=5.6.0-dev.20240816#code/JYOwLgpgTgZghgYwgAgMoHsC2EAqBPABxQG8AoZC5AeiuQHkRkBWAOgDYWAGAGmQgA8iUYNnDIARnmSY4Aa1ABzZGAAWwAM7ICUdAWRQIARwCuwAwBMW5SsZDATEAAo6CAfgBcydWGEgFAblIAX1JScwgEABs4A2QEdBBvZHRxdU86VOgANzhxSIgAHgxsfCIAPkDSGFsEMGAE5BiFAHEIMHUAdR0-UogACgBKdMyoHLzC4iCy5DJKfTbjKEYU9RYCYCI+sDgCPqbkAF5p2bnKJsDTyhoAPVdrS5pkAFFBCMhzT2LcQhRQZhZWAAWe6nR4AQVqxjgkU8FEmyD+WVYHE4AFpxG04Mp0P8UajwlkWAAmThEwGcAAcAEY2Mg9mBkPk4N4BiCggMBoEQlUanUGk1Wu0AMLoKAGWq9QYzEEGMCLZapNYbfrbXb7I7Sy4Uc4guY3O5a6i0MGRADucDwmniYrebI5XNCoEgsEQKAy6myuXyBRwxxB6yIBTBZT6unSQjgYFFADFefUQD7eMGhvQRmNvcGHU7oPAkMgAKogGJ4WMgWrxxPIABKxzp6nQiyQnhwKar-mQIWzLrzdAjUagpfLCUrNb4-EgIHMmkLxcHfIT7s94x9ZV4i9GXsKNdrnfAOddyAAsgl0L1e9BIzG48PfWOJ1P6H2r2X55Xb8QO2EItFYtUX-HlB2Fc+hAAQwE8PockiYwIGbAZDmmLJ0GAcwU2PEBTx+c8oEvAdrwTX1AiAA

### 💻 Code

```ts
interface SomeType {
    // On 5.6.0, experiment by making this prop required.
    uniqueProp?: string;
}

declare const obs: Observable<SomeType>;

function argGetsWrongType(): Observable<{}> {
[...]

PR #59709, opened 2024-08-21
Improve logic that chooses co- vs. contra-variant return type inferences

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/59656
cc @weswigham as the reviewer of https://github.com/microsoft/TypeScript/pull/57909 by which this got affected

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -26982,18 +26982,38 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 const inferredCovariantType = inference.candidates ? getCovariantInference(inference, context.signature) : undefined;
                 const inferredContravariantType = inference.contraCandidates ? getContravariantInference(inference) : undefined;
                 if (inferredCovariantType || inferredContravariantType) {
-                    // If we have both co- and contra-variant inferences, we prefer the co-variant inference if it is not 'never',
-                    // all co-variant inferences are assignable to it (i.e. it isn't one of a conflicting set of candidates), it is
-                    // assignable to some contra-variant inference, and no other type parameter is constrained to this type parameter
-                    // and has inferences that would conflict. Otherwise, we prefer the contra-variant inference.
-                    // Similarly ignore co-variant `any` inference when both are available as almost everything is assignable to it
-                    // and it would spoil the overall inference.
+                    // ideally all inferences would be tried out in a ranked order but that's too costly
+                    // a local heuristic (created based on empirical testing) is used to pick the best candidate here:
+                    // 1. covariant `never` and `any` are ignored as they are assignable to most types and choosing them would spoil the overall result
+                    // 2. covariant inference has to be compatible with at least some contravariant inferences
+                    // 3. the variable being inferred isn't referred to by another variable's constraint directly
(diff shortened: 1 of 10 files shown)


src/compiler/checker.ts
```diff
@@ -26982,18 +26982,38 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 const inferredCovariantType = inference.candidates ? getCovariantInference(inference, context.signature) : undefined;
                 const inferredContravariantType = inference.contraCandidates ? getContravariantInference(inference) : undefined;
                 if (inferredCovariantType || inferredContravariantType) {
-                    // If we have both co- and contra-variant inferences, we prefer the co-variant inference if it is not 'never',
-                    // all co-variant inferences are assignable to it (i.e. it isn't one of a conflicting set of candidates), it is
-                    // assignable to some contra-variant inference, and no other type parameter is constrained to this type parameter
-                    // and has inferences that would conflict. Otherwise, we prefer the contra-variant inference.
-                    // Similarly ignore co-variant `any` inference when both are available as almost everything is assignable to it
-                    // and it would spoil the overall inference.
+                    // ideally all inferences would be tried out in a ranked order but that's too costly
```
Andarist, line 26985: I rewritten this comment based on the excellent comment by @weswigham [here](https://github.com/microsoft/TypeScript/pull/57909#pullrequestreview-2119045380). I tried to be concise though as this is already pretty long and Wes' comment was even longer 😉 Feel free to suggest any improvement to this to make it better. I write better in TS than in English.

src/compiler/checker.ts
```diff
@@ -26982,18 +26982,38 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+                    //     query: ''
+                    //   },
+                    // });
+                    //
+                    // however, for return types inferences subtyping is used. In those situations, the contravariant inference often comes from the argument of mapping function
+                    // and the outer covariant requirement can still be satisfied by the return position of the return type when it's a subtype of that required return type.
+                    // this helps to provide better contextual parameter types in scenarios like this:
+                    //
+                    // declare const obs: Observable<{ a?: string; b?: number }>;
+                    // const test = (): Observable<{ a?: string }> => obs.pipe(tap((arg) => {}));
```
Andarist, line 27009: This part of the comment tries to explain the idea behind the fix. 

An alternative idea I had was to experiment with a new inference priority to differentiate between covariant inferences made from return positions. However, that's a bigger experiment and what I propose here is much safer at this stage. It's essentially reverting *part* of https://github.com/microsoft/TypeScript/pull/57909 to 5.5 behavior. So it shouldn't behave worse in any case. 

But also given that return positions could themselves be within nested functions, I don't have a good intuition on how it could play out. At the moment, there is no concept of "depth" in the inference and the core of this idea conceptually is that a return position might not be more important than a contravariant inference made from the (any?) *containing* function.

This would have some potential to solve Wes' concern raised [here](https://github.com/microsoft/TypeScript/pull/57909) related to inferring from `(x: number) => 0` to `(x: T) => T`.

jakebailey:
@typescript-bot test it

jakebailey:
@typescript-bot pack this

DanielRosenwasser:
@typescript-bot test top999

PR #59714, opened 2024-08-22
Add missing test case for assertion functions with `PropertySignature` declarations

Andarist:
I was looking into `isDeclarationWithExplicitTypeAnnotation` to assess if it's not missing `MethodSignature` handling. I couldn't come up with a test case that would prove it is. I think that case is always handled by an early check for `SymbolFlags.Method` in `getExplicitTypeOfSymbol` (and that's the only caller of `isDeclarationWithExplicitTypeAnnotation`).

In the process, I found out that there was no test case for the `isPropertySignature` check in `isDeclarationWithExplicitTypeAnnotation`. So here it is (plus some extras) 😉


gabritto (approved):

Bongdautay:
quaroibn1@gmail.com

Issue #59464, opened 2024-07-29
Paste with imports should not add imports when pasting into a string


mjbvz:
### 🔎 Search Terms

- Paste with imports

### 🕗 Version & Regression Information

5.6.0-dev.20240729

### 💻 Code

For a project

```ts
// abc.ts
export const abc =123;

console.log(abc);
```

Copy and paste `abc` into a file:

```ts
const str = "|"
```

Where `|` is the cursor position 

### 🙁 Actual behavior

An import is added:

```ts
import { abc } from "./abc";

const a = 'abc'
```

### 🙂 Expected behavior

No import added

mjbvz:
Other cases that are also broken:

- Paste into comment: `// |` or `/**  | */`
- Paste into regex: `/x|/g`
- Pasting into JSX body: `<b>|</b>`
- Past into JSX attribute: `<b className="|"></b>`

IllusionMH:
> Pasting into JSX body: `<b>|</b>`

But what if you paste components and expressions? They might require imports

mjbvz:
Yes for any of these cases, we also have to consider what the file will look like post paste. So for the string case (`'|'`), even though imports should not be added when pasting `abc`, they should be added when pasting `'+ abc + '` for example

RyanCavanaugh:
Closing language service bugs related to the 6.0 implementation. For more information, see #62827

PR #59719, opened 2024-08-22
Skip adding imports at some locations

navya9singh:
This pr fixes https://github.com/microsoft/TypeScript/issues/59464. If the file the text was copied from is known, then pasteEdits adds imports only for symbols that were exported from that file or for imports that need to be copied over to the target file.


src/services/pasteEdits.ts
```diff
@@ -66,18 +67,19 @@ function pasteEdits(
...
 
-    const statements: Statement[] = [];
     let newText = targetFile.text;
     for (let i = pasteLocations.length - 1; i >= 0; i--) {
         const { pos, end } = pasteLocations[i];
         newText = actualPastedText ? newText.slice(0, pos) + actualPastedText + newText.slice(end) : newText.slice(0, pos) + pastedText[i] + newText.slice(end);
     }
 
     let importAdder: codefix.ImportAdder;
+    const symbolsUsageCopiedFrom = new Map<string, [Symbol, boolean, codefix.ImportOrRequireAliasDeclaration | undefined]>();
```
DanielRosenwasser, line 77: Can you use a (preferably named) object type instead of this tuple?

src/services/pasteEdits.ts
```diff
@@ -94,62 +96,54 @@ function pasteEdits(
...
             });
             const usage = getUsageInfo(copiedFrom.file, statements, originalProgram!.getTypeChecker(), getExistingLocals(updatedFile, statements, originalProgram!.getTypeChecker()), { pos: copiedFrom.range[0].pos, end: copiedFrom.range[copiedFrom.range.length - 1].end });
-            Debug.assertIsDefined(originalProgram);
-            const useEsModuleSyntax = !fileShouldUseJavaScriptRequire(targetFile.fileName, originalProgram, host, !!copiedFrom.file.commonJsModuleIndicator);
-            addExportsInOldFile(copiedFrom.file, usage.targetFileImportsFromOldFile, changes, useEsModuleSyntax);
-            addTargetFileImports(copiedFrom.file, usage.oldImportsNeededByTargetFile, usage.targetFileImportsFromOldFile, originalProgram.getTypeChecker(), updatedProgram, importAdder);
+            // Map created with `symbol.name` as the key to retrieve symbols declared in the old file that need to be resolved in the target file (only the one's that were exported).
+            for (const [symbol, [isValidTypeOnlyUseSite, declaration]] of usage.oldImportsNeededByTargetFile) {
```
DanielRosenwasser, line 100: Should `oldImportsNeededByTargetFile` just be a bunch of `SymbolInfo`s as well? If you did that, you could just copy all the same objects over.

src/services/pasteEdits.ts
```diff
@@ -167,3 +161,49 @@ function pasteEdits(
...
+    updatedProgram: Program,
+    importAdder: ImportAdder,
+    host: LanguageServiceHost,
+    preferences: UserPreferences,
+    formatContext: formatting.FormatContext,
+    cancellationToken: CancellationToken,
+    copiedFrom?: { file: SourceFile; range: TextRange[]; } | undefined,
+    symbolsUsageCopiedFrom?: Map<string, SymbolInfo>,
+) {
+    if (copiedFrom) {
```
DanielRosenwasser, line 178: You technically only need to pass in `symbolsUsageCopiedFrom`. That avoids these two parameters from drifting apart in any intended meaning.

PR #59720, opened 2024-08-22
Enable Node.js/v8 compile caching via entrypoint shims

jakebailey:
This makes use of https://github.com/nodejs/node/pull/54501 by converting our larger entrypoints (`tsc`, `tsserver`, `typingsInstaller`) into shims which call `require("node:module").enableCompileCache()`, then `require` the actual code. Our entrypoints are assumed to always be run in Node (or a compatible runtime), so this is safe.

This gives roughly a 2.5x startup time boost.

```
Benchmark 1: node ./built/local/_tsc.js --version
  Time (mean ± σ):     122.2 ms ±   1.5 ms    [User: 101.7 ms, System: 13.0 ms]
  Range (min … max):   119.3 ms … 132.3 ms    200 runs
 
Benchmark 2: node ./built/local/tsc.js --version
  Time (mean ± σ):      48.4 ms ±   1.0 ms    [User: 34.0 ms, System: 11.1 ms]
  Range (min … max):    45.7 ms …  52.8 ms    200 runs
 
Summary
  node ./built/local/tsc.js --version ran
    2.52 ± 0.06 times faster than node ./built/local/_tsc.js --version
```

```
Benchmark 1 (83 runs): node ./built/local/_tsc.js --version
  measurement          mean ± σ            min … max           outliers         delta
  wall_time           120ms ± 1.43ms     118ms …  128ms          2 ( 2%)        0%
  peak_rss           75.1MB ±  648KB    74.6MB … 79.3MB          2 ( 2%)        0%
  cpu_cycles          486M  ± 3.27M      479M  …  499M           2 ( 2%)        0%
  instructions        658M  ±  265K      658M  …  659M           0 ( 0%)        0%
  cache_references   10.0M  ±  192K     9.45M  … 10.7M           2 ( 2%)        0%
  cache_misses       1.19M  ±  122K     1.00M  … 1.64M           1 ( 1%)        0%
  branch_misses      3.05M  ± 8.76K     3.01M  … 3.07M           1 ( 1%)        0%
Benchmark 2 (208 runs): node ./built/local/tsc.js --version
  measurement          mean ± σ            min … max           outliers         delta
  wall_time          48.1ms ± 1.23ms    45.5ms … 52.7ms          5 ( 2%)        ⚡- 60.0% ±  0.3%
  peak_rss           64.7MB ±  448KB    63.1MB … 66.9MB         20 (10%)        ⚡- 13.9% ±  0.2%
  cpu_cycles          166M  ± 3.95M      159M  …  183M           4 ( 2%)        ⚡- 65.8% ±  0.2%
  instructions        231M  ±  597K      230M  …  233M           2 ( 1%)        ⚡- 64.9% ±  0.0%
  cache_references   7.37M  ±  151K     7.08M  … 7.77M           0 ( 0%)        ⚡- 26.4% ±  0.4%
  cache_misses        666K  ±  133K      465K  … 1.13M           2 ( 1%)        ⚡- 43.8% ±  2.8%
  branch_misses       949K  ± 9.10K      930K  …  978K           2 ( 1%)        ⚡- 68.9% ±  0.1%
```

```
Benchmark 1: echo '{"seq": 1, "command": "status"}' | node ./built/local/_tsserver.js --disableAutomaticTypingAcquisition
  Time (mean ± σ):     206.0 ms ±   8.5 ms    [User: 176.2 ms, System: 24.4 ms]
  Range (min … max):   194.8 ms … 250.7 ms    200 runs
 
Benchmark 2: echo '{"seq": 1, "command": "status"}' | node ./built/local/tsserver.js --disableAutomaticTypingAcquisition
  Time (mean ± σ):      83.8 ms ±   1.8 ms    [User: 66.9 ms, System: 18.1 ms]
  Range (min … max):    80.9 ms …  92.4 ms    200 runs
 
Summary
  echo '{"seq": 1, "command": "status"}' | node ./built/local/tsserver.js --disableAutomaticTypingAcquisition ran
    2.46 ± 0.12 times faster than echo '{"seq": 1, "command": "status"}' | node ./built/local/_tsserver.js --disableAutomaticTypingAcquisition
```

Gotchas:

- VS Code or other clients who consume us but trim unused files will need to be updated to ensure to include `_tsc.js`, `_tsserver.js`, `_typingsInstaller.js`
- People who patch our package will break until their packages are updated (yarn).
- I have not added this to our libraries. It would be weird for that to start caching someone else's process, and it's not clear how one would even do it with an ESM based API.


jakebailey:
https://github.com/nodejs/node/pull/54501 was merged into Node, so this API should end up available in a later version. I've marked this PR as ready for review, if we want to think about whether or not it's worth the extra shim / indirection.

jakebailey:
Based on the design meeting, this seems like a good idea. Before we do this we'll need to make sure downstream users (@mjbvz, @zkat) know these new files will exist and cannot be deleted.

Herebyfile.mjs
```diff
@@ -313,7 +314,34 @@ function entrypointBuildTask(options) {
         run: () => buildProject(options.project),
     });
 
-    const bundler = createBundler(options.srcEntrypoint, options.output, options.bundlerOptions);
+    const mainDeps = options.mainDeps?.slice(0) ?? [];
+
+    let output = options.output;
+    if (options.enableCompileCache) {
+        const originalOutput = output;
+        output = path.join(path.dirname(output), "_" + path.basename(output));
```
jakebailey, line 322: This was just an expedient name choice; I can also remove the extension and then make it like `tsc_.js` which may sort better.

jakebailey:
@typescript-bot test startup-only hosts=node@20.17.0,node@22.8.0

jakebailey:
@typescript-bot perf test startup-only hosts=node@20.17.0,node@22.8.0

jakebailey:
As a note to self, if/when we switch to ESM when `require(ESM)` is stable, to ensure `require("node_modules/typescript/lib/tsc.js")` still works (is synchronous), we'd need to write this:

```ts
const { createRequire, enableCompileCache } = process.getBuiltinModule("node:module");
enableCompileCache();
createRequire(import.meta.url)("./_tsc.js");
```

Herebyfile.mjs
sandersn: I was going to say that `_tsc` tab-completes better on the command-line, though, because it doesn't interfere with `types<TAB>` -- but there's so much stuff in built/local that it doesn't actually make a difference.

sandersn (approved):

jakebailey:
@typescript-bot perf test this faster hosts=node@20.17.0,node@22.8.0

anishkumar127 (commented):
nice

PR #59736, opened 2024-08-24
Do not indent semicolons on new lines

### LINKED ISSUE #48571: Suggestion: Formatting: Disable inserting spaces around leading semicolon
# Suggestion

## 🔍 Search Terms

semicolon at the start of a line
formatting spaces around semi

## ✅ Viability Checklist

## ⭐ Suggestion

In setups where semicolons are not used and the code relies on ASI, there are certain statements which can confuse both the developer and the engine.
In such cases, some people prefer to use semicolon at the start of the particular statement that could potentially cause such behavior regardless of what is actually on the preceding line.

## 📃 Motivating Example

This adjustment will fix formatting issue and make it consistent.

## 💻 Use Cases

With vscode formatting setting
```
"typescript.format.semicolons": "remove"
```
and eslint rules,
```
"semi-style": ["error", "first"],
"semi-spacing": "error"
"indent": ["error", 2],
```

I get this automatic `eslint --fix`,

```ts
// before
console.log('hi');
(something ??= {_default: 0})[key] = 1

// after 
console.log('hi')
;(something ??= {_default: 0})[key] = 1
^ ..... this is desired
```

which is then auto-formatted by vscode to this,
```ts
console.log('hi')
  ; (something ??= {_default: 0})[key] = 1
^^ ^ ..... spaces inserted, undesired
```

which in turn raises this eslint error.
```
Expected indentation of 14 spaces but found 16. eslint(indent)
```

ChiriVulpes:
Fixes #48571


DanielRosenwasser:
@typescript-bot pack this

DanielRosenwasser:
I'm surprised there's no test for this, but there's a change where the following

```ts
type X =
    | 1
    | 2
    | 3
    ;
```

is formatted into:

```ts
type X =
    | 1
    | 2
    | 3
;
```

and I don't think that that's desirable. Not sure if others have opinions here.

jakebailey:
Yeah, that would probably be a regression. Should definitely have a test for it...

ChiriVulpes:
Hmm. That's going to require scanning forward to see if the `;` is alone on the line (or just it and comments?) 

Since that adds probably too much complexity, maybe it would be best to only do what this commit does if the `semicolons` format option is set to `remove`?

PR #59761, opened 2024-08-26
`this` can be nullish

### LINKED ISSUE #59755: 5.6.1-rc Incorrectly assumes "this" is not nullish even when declared as such
### 🔎 Search Terms

ts2869
Right operand of ?? is unreachable because the left operand is never nullish.

Feature introduced in 5.6
reproduced in 5.6.1-rc and v5.7.0-dev.20240826 

### ⏯ Playground Link

https://www.typescriptlang.org/play/?ts=5.6.1-rc#code/GYVwdgxgLglg9mABMOcAUUAWMDOAuRAeQCMArAU2kQB9FwATc4GMc+gSkQG8AoRfxACdyUEIKRZciAPzTEABgDcPAL48gA

### 💻 Code

```ts
function foo(this: Object | undefined) {
    return this ?? 0;
}
```

### 🙁 Actual behavior

Complains that `this` is never nullish.

image

### 🙂 Expected behavior

Since `this` is explicitly declared as `| undefined`, the error should not be shown.

In non-strict mode, `this` indeed could not be nullish, but in strict mode it can.

```
new Function(`"use strict"
function foo() {
    console.log(this);
}
foo.apply(undefined)`)()
```

_No response_

RyanCavanaugh:
`this` keyword wasn't categorized as possibly nullish, but it is.

Fixes #59755

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -39818,6 +39818,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             case SyntaxKind.NewExpression:
             case SyntaxKind.PropertyAccessExpression:
             case SyntaxKind.YieldExpression:
+            case SyntaxKind.ThisKeyword:
                 return PredicateSemantics.Sometimes;
             case SyntaxKind.BinaryExpression:
                 // List of operators that can produce null/undefined:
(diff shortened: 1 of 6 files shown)


jakebailey (approved):

RyanCavanaugh:

PR #59767, opened 2024-08-26
Rewrite relative import extensions with flag

### LINKED ISSUE #59597: Rewrite imports with ".ts" extension to ".js" to support Nodejs's "--experimental-strip-types"
### 🔍 Search Terms

https://github.com/microsoft/TypeScript/issues/55346

### ✅ Viability Checklist

- [ ] This wouldn't change the runtime behavior of existing JavaScript code
- [ ] This could be implemented without emitting different JS based on the types of the expressions
- [ ] This feature would agree with the rest of our Design Goals: https://github.com/Microsoft/TypeScript/wiki/TypeScript-Design-Goals

### ⭐ Context

Nodejs is shipping built-in TypeScript support for local & development projects, where distributing for production still need to compile typescript sources to javascript for them to work. 

This can be used in version `>=22.6.0` behind the `--experimental-strip-types` flag.

A point of contention is that the Nodejs TypeScript transpiler relies on importing TypeScript sources using the `.ts` file extension.

Example;

```typescript
// node --experimental-strip-types ./index.ts
// index.ts
import * as foo from './foo.ts'
```

While we can use the new Nodejs flag to run local TypeScript sources during development with `.ts` file extensions and `tsconfig.json#compilerOptions.allowImportingTsExtensions: true`; The TypeScript compiler is no longer capable of producing a valid output to run in Nodejs using this source code.

### ⭐ Discussion

In previous issues, members of the TypeScript team have stated that rewriting imports is out of scope for TypeScript and will not be implemented. 

Has this stance changed in light of the newly added TypeScript support in Nodejs?

I am aware that the support status is experimental so Nodejs will make changes before it rolls out generally but it looks like this particular decision is not likely to change

andrewbranch:
This PR adds `--rewriteRelativeImportExtensions` to support transforming module specifiers from e.g. `"./utils.ts"` to `"./utils.js"`. A module specifier will be statically rewritten or shimmed during JS emit if it meets all of these criteria:

- it begins with `./` or `../`
- it does not end in a declaration file name (`.d.ts`, `.d.mts`, `.d.cts`, `.d.*.ts`)
- it ends in `.ts`, `.tsx`, `.mts`, or `.cts`

## Error checking

Errors are issued in an attempt to catch common mistakes:

- A module specifier **will** get rewritten but will break after rewriting:
  - `"./foo.ts"` actually resolves to `./foo.ts.ts` or `foo.ts/index.ts`
  - `"../other-project/src/index.ts"` belongs to another project where outputs go to a different outDir
- A module specifier **will not** get rewritten but will break after emitting as-is:
  - `"#blah/foo.ts"` resolves to `foo.ts` where the `.ts` matches through a wildcard

## Syntax support

| Scenario            | Example input       | Operation | Example output                             |
|---------------------|---------------------|-----------|--------------------------------------------|
| Import declarations | `import "./foo.ts"` | Rewrite   | `import "./foo.js"`, `require("./foo.js")` |
| Export declarations | `export * from "./foo.ts"` | Rewrite | `export * from "./foo.js"` |
| Import equals       | `import foo = require("./foo.ts")` | Rewrite | `import foo = require("./foo.js")` |
| Import calls with qualifying literal argument | `import("./foo.ts")` | Rewrite | `import("./foo.js")` |
| Require calls in JS with qualifying literal argument | `require("./foo.ts")` | Rewrite | `require("./foo.js")` |
| Import calls with non-literal argument | `import(getPath())` | Shim | `import(__rewriteRelativeImportExtension(getPath()))` |
| Require calls in JS with non-literal argument | `require(getPath())` | Shim | `require(__rewriteRelativeImportExtension(getPath()))` |
| Import calls with non-qualifying literal argument | `import("foo")` | **None** | `import("foo")` |
| Require calls with non-qualifying literal argument | `require("foo")` | **None** | `require("foo")` |
| Require calls in TS | `require("./foo.ts")` | **None** | `require("./foo.ts")` |
| Indirected require calls | `(require)("./foo.ts")` | **None** | `(require)("./foo.ts")` |

## Limitations and future work

This PR doesn’t change how module resolution works, so it’s still possible to reference a `.ts` file by a `.js` module specifier:

```ts
// @Filename: a.ts
export {};

// @Filename: b.ts
import {} from "./a.js"; // Always ok
```

We discussed that it’s probably desirable to turn this functionality off under `--rewriteRelativeImportExtensions` (and `--allowImportingTsExtensions`), since these options are a strong indicator that the user is going to run the input TS files in a TS-native runtime like Bun or `node --experimental-strip-types`. That would be a breaking change and isn’t included in this PR.

tslib PR: https://github.com/microsoft/tslib/pull/270
Fixes #59597


andrewbranch:
@typescript-bot pack this

andrewbranch:
@typescript-bot pack this

andrewbranch:
@typescript-bot pack this

src/compiler/factory/emitHelpers.ts
```diff
@@ -1422,6 +1438,29 @@ const disposeResourcesHelper: UnscopedEmitHelper = {
...
 };
 
+const rewriteRelativeImportExtensionsHelper: UnscopedEmitHelper = {
+    name: "typescript:rewriteRelativeImportExtensions",
+    importName: "__rewriteRelativeImportExtension",
+    scoped: false,
+    text: `
+        var __rewriteRelativeImportExtension = (this && this.__rewriteRelativeImportExtension) || function (path, preserveJsx) {
+            if (typeof path === "string" && path[0] === "." && (path[1] === "/" || path[1] === "." && path[2] === "/")) {
+                if (path.substring(path.length - 4) === ".tsx") {
```
weswigham, line 1448: I'm going to regret asking this, but does this (and the rest of the extension comparisons) need to be case-insensitive? At least on windows, `.TSX` and `.tsx` files are going to be the same - I dunno how loose we are on case-sensitive filesystems.

tests/baselines/reference/config/showConfig/Shows tsconfig for single option/rewriteRelativeImportExtensions/tsconfig.json
```diff
@@ -0,0 +1,6 @@
+{
+    "compilerOptions": {
+        "rewriteRelativeImportExtensions": true,
+        "allowImportingTsExtensions": true
```
weswigham, line 4: I think something's wrong for `showConfig` - it's supposed to be smart enough to know `allowImportingTsExtensions` is derived from `rewriteRelativeImportExtensions` and elide it, right?

tests/baselines/reference/emit(jsx=preserve).js
```diff
@@ -0,0 +1,131 @@
...
+};
+// Rewrite
+import {} from "./foo.js";
+import {} from "../foo.mjs";
+import {} from "../../foo.cjs";
+import {} from "./foo.jsx";
+import "./foo.js";
+export * from "./foo.js";
+// Shim
+import(__rewriteRelativeImportExtension("./foo.ts", true));
```
weswigham, line 109: Minor potential UX optimization: You think it'd be worth building in knowledge in the checker/binder that `import(__rewriteRelativeImportExtension("./foo.ts"))` can be bound/checked the same as `import("./foo.ts")`, just so LS operations on emitted files work a bit better?

tests/baselines/reference/config/showConfig/Shows tsconfig for single option/rewriteRelativeImportExtensions/tsconfig.json
andrewbranch: No, since a couple releases ago, `--showConfig` shows all implied options (if they evaluate to something other than their default value)

tests/baselines/reference/emit(jsx=preserve).js
jakebailey: Yeah, I guess I thought we'd not emit the helper when we know statically that it's a literal, right?

src/compiler/transformers/module/module.ts
```diff
@@ -797,8 +809,15 @@ export function transformModule(context: TransformationContext): (x: SourceFile
             case SyntaxKind.PartiallyEmittedExpression:
                 return visitPartiallyEmittedExpression(node as PartiallyEmittedExpression, valueIsDiscarded);
             case SyntaxKind.CallExpression:
+                const needsShim = node === importsAndRequiresToShim?.[0];
```
rbuckton, line 812: We should perform a length check here to avoid a deopt due to out of bound indexed access. I'd suggest using `firstOrUndefined`

rbuckton (changes requested):

tests/baselines/reference/emit(jsx=preserve).js
andrewbranch: > Yeah, I guess I thought we'd not emit the helper when we know statically that it's a literal, right?

It was suggested that import/require calls be shimmed unconditionally so someone debugging the emitted code can understand what happened

andrewbranch: But, _not_ taking that suggestion would be a much easier way of making the emitted code work in the LS. I’m a bit concerned about how far the increased complexity of changing what imports/requires can be considered a module reference would reach into the LS.

weswigham: 🤷 Maybe one day we'll change `require`/`import` inference logic to operate on resolved types and not just literals, so declaring `declare function __rewriteRelativeImportExtension<T extends string>(arg: T): T;` makes an `import`/`resolve` of the call on the literal still work (since the function is defined as a pass-thru, type-wise) and we shouldn't bother special-casing it now.

andrewbranch: I’m going to push a commit that changes the literal case from shimming to rewriting—I think losing the explicitness of what happened in the emit is the least bad of the tradeoffs discussed here. We can always emit a synthesized comment if we want to make it more clear that something happened in the emit. (Might not be a bad idea for all the rewrites.) I won’t have time to get to that before I leave, but can come back to it for the RC if that’s something we want to do.

src/compiler/factory/emitHelpers.ts
```diff
@@ -1422,6 +1438,21 @@ const disposeResourcesHelper: UnscopedEmitHelper = {
         });`,
 };
 
+const rewriteRelativeImportExtensionsHelper: UnscopedEmitHelper = {
+    name: "typescript:rewriteRelativeImportExtensions",
+    importName: "__rewriteRelativeImportExtension",
+    scoped: false,
+    text: `
+        var __rewriteRelativeImportExtension = (this && this.__rewriteRelativeImportExtension) || function (path, preserveJsx) {
```
rbuckton, line 1446:
```suggestion
    var __rewriteRelativeImportExtension = (this && this.__rewriteRelativeImportExtension) || function (path, preserveJsx) {

```

tests/baselines/reference/emitModuleCommonJS(module=commonjs).js
```diff
@@ -0,0 +1,45 @@
...
+
+//// [b.ts]
+{
+  import("" + "./foo.ts");
+  import("./foo.ts");
+}
+
+
+//// [a.js]
+    var __rewriteRelativeImportExtension = (this && this.__rewriteRelativeImportExtension) || function (path, preserveJsx) {
```
rbuckton, line 19: First line here is indented too far. See my suggestion above.

rbuckton (changes requested):

rbuckton (approved):
Barring feedback from others, LGTM.

jakebailey (approved):
LGTM as designed, though I am curious about whether or not we need to be adding errors for people who wrote `.js` but should have written `.ts` in this mode (since that will cause people running source directly to error).

bakkot:
The [original issue](https://github.com/microsoft/TypeScript/issues/49083) was locked; it would be nice for someone with permissions to comment there so that people who were following there get this update.

jlmart88:
@RyanCavanaugh I happened upon your final post here: https://github.com/microsoft/TypeScript/issues/49083#issuecomment-1435399267 minutes before you commented and referenced this PR. 

I'm curious what changed from your reasoning a year ago as to why this would never be a feature?

RyanCavanaugh:
NodeJS shipping `--experimental-strip-types` means there's now an actual runtime scenario where you need *both* JS and TS extensions

AlttiRi:
I have tested TS `5.7.0-dev`. 

**It does not rewrite `.d.ts` files.**

My library preset on `5.6.2`:
```jsonc
    /* Library preset */
    "module": "NodeNext",
    "moduleResolution": "NodeNext",
    "noEmit": false,
    "declaration": true,
```
and new one on `5.7.0-dev`:
```jsonc
    /* Library preset */
    "module": "NodeNext",
    "moduleResolution": "NodeNext",
    "allowImportingTsExtensions": true,      // <- new
    "rewriteRelativeImportExtensions": true, // <- new
    "noEmit": false,
    "declaration": true,
```

It works, I can use `tsc` _(after changing `.js` to `.ts` in imports in all files)_.

Also, now I can run TS files with `deno` without using `--unstable-sloppy-imports` which prints annoying warnings.

**However, imports inside `.d.ts` files have `.ts` ending now. There were no rewriting.**

I expect that both configs should produce absolutely the same result (`.js` and `.d.ts` files) after using `tsc`.

RyanCavanaugh:
Imports in `.d.ts` files are allowed to use `.ts` extensions regardless of config, so I wouldn't really consider that a bug.

AlttiRi:
It's a bug. Not rewritten imports in `d.ts.` breaks (partially) the code assistance in an IDE.

It's OK:

Screenshot
(`npm install @alttiri/util-js@1.14.0-20240821`)


After using [`rewriteRelativeImportExtensions`](https://github.com/AlttiRi/util-js/commit/b61ce2645a5ec0e8da4bc58aa0f3113c98a37a25):

Screenshot_1
(`npm install @alttiri/util-js@1.15.0-dev.20241002 --registry=https://npm.pkg.github.com`)

In this case, the IDE no longer highlights the function names, nor does it show the function signature.

**The imports in `.d.ts` also should be rewritten.**

AlttiRi:
Also, it currently (?) does not support aliases which resolve to _a relative path_.

```json
{
  "extends": "../tsconfig.json",
  "compilerOptions": {
    "baseUrl": ".",
    "paths": {
      "@/*": ["../*"]
    }
  }
}
```

I get the follow error for `import {sleep} from "@/index.ts";`:
```
TS2877: This import uses a '.ts' extension to resolve to an input TypeScript file, but will not be rewritten during emit because it is not a relative path.
```

RyanCavanaugh:
> Also, it currently (?) does not support aliases which resolve to a relative path.

This is 100% intentional

AlttiRi:
Is it a problem to resolve the alias first and only after that check if it is a relative path or not?

It doesn't seem like something that is difficult to implement. Just only one extra simple action.

RyanCavanaugh:
I don't think the setup you have there works under `--experimental-strip-types`. https://github.com/microsoft/TypeScript/pull/59767#issuecomment-2383844756

AlttiRi:
Okay, I have checked it with the latest Node.js (v22.9.0) too _(I tested it with Deno before)_.

It works (as expected).

Screenshot

RyanCavanaugh:
I can't replicate what you're doing there
```ts
// with
// import * as bar from "./bar.ts";

node --experimental-strip-types foo.ts
hi
```
vs
```ts
// with
// import * as bar from "@/bar.ts";
Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@/bar.ts' imported from D:\Throwaway\est-test\foo.ts
```

The purpose of the flag is to rewrite paths which are syntactically relative, not "semantically" relative (whatever that would even mean).

nicolo-ribaudo:
Hey, I was reading the table in the PR description and I'm getting confused by these two lines:

> | Scenario | Example input | Operation | Example output |
> |-|-|-|-|
> | Require calls in JS with qualifying literal argument | `require("./foo.ts")` | Rewrite | `require("./foo.js")` |
> | Require calls in TS | `require("./foo.ts")` | **None** | `require("./foo.ts")` |

Does it mean that `require("./foo.ts")` will be rewritten if it's inside a JS file, but not if it's inside a TS file?

andrewbranch:
Yes. Plain `require` calls aren’t recognized as special in TS files. You’ll notice in JS that `require("./foo")` performs module resolution (the compiler knows the type to return), but in TS, `require` isn’t even allowed unless it’s declared as a global function, in which case that function’s return type is used (usually `any`).

jremy42:
Just out of curiosity, I’m facing an issue where my .d.ts files don’t have an extension in the path. For example: export * as v1alpha2 from './v1alpha2/index.gen';. Why is the .ts extension mandatory to transform it into .js? Thank you.

nyan-left:
@RyanCavanaugh 
> Imports in `.d.ts` files are allowed to use `.ts` extensions regardless of config, so I wouldn't really consider that a bug.

While `.d.ts` files may allow `.ts` extensions in imports without configuration, wouldn't this break backwards compatibility for libraries targeting TypeScript 4 support since `allowImportingTsExtensions` wasn't introduced until TypeScript 5?

src/compiler/factory/emitHelpers.ts
```diff
@@ -1422,6 +1438,21 @@ const disposeResourcesHelper: UnscopedEmitHelper = {
...
 };
 
+const rewriteRelativeImportExtensionsHelper: UnscopedEmitHelper = {
+    name: "typescript:rewriteRelativeImportExtensions",
+    importName: "__rewriteRelativeImportExtension",
+    scoped: false,
+    text: `
+        var __rewriteRelativeImportExtension = (this && this.__rewriteRelativeImportExtension) || function (path, preserveJsx) {
+            if (typeof path === "string" && /^\\.\\.?\\//.test(path)) {
+                return path.replace(/\\.(tsx)$|((?:\\.d)?)((?:\\.[^./]+?)?)\\.([cm]?)ts$/i, function (m, tsx, d, ext, cm) {
```
sant123, line 1448: If a path contains `.jsx` it is preserved even with the `preserveJsx` argument. 

Should the regex be `/\.([jt]sx)$|((?:\.d)?)((?:\.[^./]+?)?)\.([cm]?)ts$/i`?

This now works:

```ts
console.log(rewriteRelativeImportExtension("./main.tsx"));
console.log(rewriteRelativeImportExtension("./main.ts"));
console.log(rewriteRelativeImportExtension("./main.jsx"));
console.log(rewriteRelativeImportExtension("./main.jsx", true));
console.log(rewriteRelativeImportExtension("./main.js"));
```

```
./main.js
./main.js
./main.js
./main.jsx
./main.js
```

andrewbranch: Can you clarify your actual behavior and expected behavior?

sant123: Nevermind, I'm writing a plugin that is based on `__rewriteRelativeImportExtension`. For Typescript is fine, however for esbuild the `.jsx` can the transformed to `.js`.

nwidynski:
@RyanCavanaugh Shouldn't calls to `import.meta.resolve()`/`require.resolve()` also be shimmed?

Issue #59765, opened 2024-08-26
5.6 regression: Inference with mixed in type intersection no longer compiles


trevorade:
### 🔎 Search Terms

Inference

### 🕗 Version & Regression Information

- This changed between versions 5.5.4 and 5.6.0-beta (still happening in 5.6.1-rc)

### ⏯ Playground Link

https://www.typescriptlang.org/play/?ts=5.6.1-rc#code/C4TwDgpgBAYgrgOwMbAJYHsEGF0FsyYQLAA8AClALxQCGCIAfFVABRgBO6YAzgFxRkAlFSYBRADYRcRYFAA+UBHHHiA3ACh1qYhHYAzGkmgSpMgILBg7VACM4wCNygBvdVCioAJngDSEEAD8-NxW2gDm8orK4pFKuDa6Gu5IABao4p7sREFQiADWCOgA7ghJUNx5qGA5NujoknQaAL6anhBI4jRZUHqIKBgIUBCS0sTkDCxuUMA0YfzwyGiYOPiEYxQAZFAmo8AWoXYO3AwANFM0ltaHjvyb2yPml7b2jmeC-DsyGuptHV3QvUWA3uADkaNI2JwePxnAhwRAciFrAgwk13vdTMRvsNMcAWBIwdITi40RoAPRkqAAEXQjkU6FkSDwYHS0G0UAArAA6ABsXKgAGUUuhlJ4oEz8KyueogA

### 💻 Code

```ts
type FunctionComponent<P = any> = (props: P) => Element | null;

interface ElementAttributes {
  idomKey?: string | null | number;
  children?: unknown;
  skip?: boolean;
}

declare function element<P>(
  tag: FunctionComponent<P & ElementAttributes>,
  attributes: P & ElementAttributes,
): Element;

declare function ElName(props: {name?: string}): Element;

element(ElName, {});
// Does not compile in 5.6. Should compile.
```

### 🙁 Actual behavior

Last line errors with:

```
Argument of type '(props: { name?: string | undefined; }) => Element' is not assignable to parameter of type 'FunctionComponent<ElementAttributes>'.
  Types of parameters 'props' and 'props' are incompatible.
    Type 'ElementAttributes' has no properties in common with type '{ name?: string | undefined; }'.(2345)
```

### 🙂 Expected behavior

Compiles as it did in previous versions

### Additional information about the issue

Confirmed that https://github.com/microsoft/TypeScript/pull/59709 does not resolve this issue.

[...]

PR #59772, opened 2024-08-27
Tweak logic that chooses co- vs. contra-variant inferences when covariant inference is an empty object type

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/59765
cc @weswigham as the reviewer of https://github.com/microsoft/TypeScript/pull/57909 by which this got affected

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -26989,7 +26989,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     // Similarly ignore co-variant `any` inference when both are available as almost everything is assignable to it
                     // and it would spoil the overall inference.
                     const preferCovariantType = inferredCovariantType && (!inferredContravariantType ||
-                        !(inferredCovariantType.flags & (TypeFlags.Never | TypeFlags.Any)) &&
+                        !(inferredCovariantType.flags & (TypeFlags.Never | TypeFlags.Any) || isEmptyAnonymousObjectType(inferredCovariantType) && isWeakType(inferredContravariantType)) &&
                             some(inference.contraCandidates, t => isTypeAssignableTo(inferredCovariantType, t)) &&
                             every(context.inferences, other =>
                                 other !== inference && getConstraintOfTypeParameter(other.typeParameter) !== inference.typeParameter ||
--- a/tests/baselines/reference/coAndContraVariantInferences12.errors.txt
+++ b/tests/baselines/reference/coAndContraVariantInferences12.errors.txt
@@ -1,0 +1,39 @@
+coAndContraVariantInferences12.ts(18,19): error TS2353: Object literal may only specify known properties, and 'age' does not exist in type '{ name?: string | undefined; } & ElementAttributes'.
+coAndContraVariantInferences12.ts(21,18): error TS2345: Argument of type '{}' is not assignable to parameter of type '{ name: string; } & ElementAttributes'.
+  Property 'name' is missing in type '{}' but required in type '{ name: string; }'.
+
+
+==== coAndContraVariantInferences12.ts (2 errors) ====
+    // https://github.com/microsoft/TypeScript/issues/59765
+    
+    type FunctionComponent<P = any> = (props: P) => Element | null;
+    
+    interface ElementAttributes {
+      idomKey?: string | null | number;
+      children?: unknown;
+      skip?: boolean;
+    }
+    
+    declare function element<P>(
+      tag: FunctionComponent<P & ElementAttributes>,
+      attributes: P & ElementAttributes,
+    ): Element;
+    
+    declare function ElName(props: { name?: string }): Element;
+    element(ElName, {}); // ok
+    element(ElName, { age: 42 }); // error
+                      ~~~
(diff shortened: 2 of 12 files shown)


src/compiler/checker.ts
```diff
@@ -26989,7 +26989,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     // Similarly ignore co-variant `any` inference when both are available as almost everything is assignable to it
                     // and it would spoil the overall inference.
                     const preferCovariantType = inferredCovariantType && (!inferredContravariantType ||
-                        !(inferredCovariantType.flags & (TypeFlags.Never | TypeFlags.Any)) &&
+                        !(inferredCovariantType.flags & (TypeFlags.Never | TypeFlags.Any) || isEmptyAnonymousObjectType(inferredCovariantType) && isWeakType(inferredContravariantType)) &&
```
Andarist, line 26992: This is a conservative change that aims to address https://github.com/microsoft/TypeScript/issues/59765 . 

I could imagine this might need some further tweaking to produce good error locations when both of the inferred types are weak or something. However, so far so good. This already produces good results and good error locations as far as I can tell - perhaps part of that can be attributed to the further checks like `some(inference.contraCandidates, t => isTypeAssignableTo(inferredCovariantType, t))`.

Andarist: Actually, there is an open question here as to what the inferred type here should be (what should be the `result`'s type):
```ts
declare function call<P>(tag: (props: P) => void, attributes: P): P;

declare function consume(p: { a?: string; b?: number; c?: boolean }): void;
declare const obj: { a?: string };

const result = call(consume, obj);
```

There is no clear answer to this - both satisfy the constraint and both of the arguments. I think that, in general, a covariant inference is safer in a situation like this. It's less likely to introduce problems like this one:
```ts
declare function call<P>(tag: (props: P) => void, attributes: P): P;

declare function consume(p: { a?: string; b?: number; c?: boolean }): void;
const obj: { a?: string; } = { a: 'foo', b: 'bar' }

const result = call(consume, obj);
obj.b // it's `number` in 5.5 but at runtime it could easily be a `string`
```

Nowadays the user can use `NoInfer` to guide what should be used as the inference source if they don't get the desired inferences for their use case - it's all a little bit situational after all.

jakebailey:
@typescript-bot test it

src/compiler/checker.ts
Andarist: I also experimented with an alternative patch:
```diff
diff --git a/src/compiler/checker.ts b/src/compiler/checker.ts
index 954e45cdeb..f5e3d12778 100644
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -26989,11 +26989,12 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     // Similarly ignore co-variant `any` inference when both are available as almost everything is assignable to it
                     // and it would spoil the overall inference.
                     const preferCovariantType = inferredCovariantType && (!inferredContravariantType ||
-                        !(inferredCovariantType.flags & (TypeFlags.Never | TypeFlags.Any) || isEmptyAnonymousObjectType(inferredCovariantType) && isWeakType(inferredContravariantType)) &&
+                        !(inferredCovariantType.flags & (TypeFlags.Never | TypeFlags.Any)) &&
                             some(inference.contraCandidates, t => isTypeAssignableTo(inferredCovariantType, t)) &&
                             every(context.inferences, other =>
                                 other !== inference && getConstraintOfTypeParameter(other.typeParameter) !== inference.typeParameter ||
-                                every(other.candidates, t => isTypeAssignableTo(t, inferredCovariantType))));
+                                every(other.candidates, t => isTypeAssignableTo(t, inferredCovariantType))) &&
+                            (some(inference.candidates, t => !isObjectOrArrayLiteralType(t)) || isTypeAssignableTo(inferredCovariantType, inferredContravariantType) && !isTypeAssignableTo(inferredContravariantType, inferredCovariantType)));
                     inferredType = preferCovariantType ? inferredCovariantType : inferredContravariantType;
                     fallbackType = preferCovariantType ? inferredContravariantType : inferredCovariantType;
                 }
```

All tests pass with it with no changes whatsoever but then it makes those 2 to behave differently:
```ts
declare function call<P>(tag: (props: P) => void, attributes: P): P;
declare function consume(p: { a?: string; b?: number; c?: boolean }): void;

declare const obj1: { a?: string };
const result1 = call(consume, obj1);
const result4 = call(consume, { ...obj1 });
```

And all in all, I'm just not sure what's right here. It all boils down to choosing some tradeoffs.

RyanCavanaugh:
I don't feel like this change is well-motivated enough to merge as-is. It seems like it's just detecting the repro in the linked issue and doing something else when it sees that, rather than a principled approach of what to do if all inference sites for a type parameter are intersections. Given the call on the linked issue it sounds like we're in agreement to not take this, but I'm definitely willing to have a discussion from first principles (just a log a new bug and we can take it from there).

Andarist:
I agree. I think that *maybe* a patch closer to [this one](https://github.com/microsoft/TypeScript/pull/59772#discussion_r1735378361) could improve some scenarios but I'm also not sure. It's hard for me to know which tradeoffs should be preferred by the overall TS design in cases like this.

PR #59774, opened 2024-08-27
 Don't generalize when assigment target is never (fix: #41707) 

### LINKED ISSUE #41707: Union of string literals is incorrectly widened in error message when compared to never
**TypeScript Version:**  4.1.2, 4.0.5 (worked better in 3.9.7)

**Search Terms:** "string is not assignable to type never" (all 15 issues I found look slightly different I think)

**Code**

```ts
function assertUnreachable(_: never): void {}

type ActionTypes = 'a' | 'b'

function reducer(state: {}, action: {type: ActionTypes}) {
  switch(action.type) {
    case 'a':
      return state
    default: {
      assertUnreachable(action.type) // <-- incorrect widening on this line
      return state
    }
  }
}
```

**Expected behavior:** like in 3.9.7, the error message should be `Argument of type '"b"' is not assignable to parameter of type 'never'`

**Actual behavior:** `Argument of type 'string' is not assignable to parameter of type 'never'` - which is too wide to be actually helpful :(

**Playground Link:** https://www.typescriptlang.org/play?ts=4.1.2#code/GYVwdgxgLglg9mABAQwM6oKYCcoFUxYbIQAWyARgDYYAUA+gFyJgYBu2AlE63DACaIA3gF8AUKID0ExCFTIA5hlFQAngAcMiAILR4YACrqMqRAF5EAcmQXEAH0vkL40JFgJEhPiAjYaqKMhQGEwiADQougghqhpMOm4GRqjCHEKiiIioAO4wUKQ0xAkAdDEYqYLpGYgQaJpWFgyVVR4YUCBYSP6BSs18GMDIIJRQIU1VaJg4+ITEZFS0hXolRhxjGYRtHZkBQWNiGWJiohAIqHDURZRw8jSe3r5hQqVMFo4pHEA (change the version number to see the previous error)

**Related Issues:** ?

hyzyla:
Fixes #41707

Added a condition that when the expected type is "never," the checker should not generalize the literal type. It might not be the best solution, but it's definitely the easiest one. Another idea that I've considered was to check if the "source" type is narrowed and skip generalization for narrowed types, but I'm not sure how to do it properly and if it OK solution.

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -22056,7 +22056,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             const [sourceType, targetType] = getTypeNamesForErrorDisplay(source, target);
             let generalizedSource = source;
             let generalizedSourceType = sourceType;
-
+            
+            // Don't generalize on 'never' - we really want the original type
+            // to be displayed for use-cases like 'assertNever'.
             if (!(target.flags & TypeFlags.Never) && isLiteralType(source) && !typeCouldHaveTopLevelSingletonTypes(target)) {
                 generalizedSource = getBaseTypeOfLiteralType(source);
                 Debug.assert(!isTypeAssignableTo(generalizedSource, target), "generalized source shouldn't be assignable");
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -22057,7 +22057,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             let generalizedSource = source;
             let generalizedSourceType = sourceType;
 
-            if (isLiteralType(source) && !typeCouldHaveTopLevelSingletonTypes(target)) {
+            if (isLiteralType(source) && (!typeCouldHaveTopLevelSingletonTypes(target) && !(target.flags & TypeFlags.Never))) {
                 generalizedSource = getBaseTypeOfLiteralType(source);
                 Debug.assert(!isTypeAssignableTo(generalizedSource, target), "generalized source shouldn't be assignable");
                 generalizedSourceType = getTypeNameForErrorDisplay(generalizedSource);
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -22057,7 +22057,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             let generalizedSource = source;
             let generalizedSourceType = sourceType;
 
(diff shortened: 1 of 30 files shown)


hyzyla:

src/compiler/checker.ts
```diff
@@ -22057,7 +22057,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             let generalizedSource = source;
             let generalizedSourceType = sourceType;
 
-            if (isLiteralType(source) && !typeCouldHaveTopLevelSingletonTypes(target)) {
+            if (isLiteralType(source) && (!typeCouldHaveTopLevelSingletonTypes(target) && !(target.flags & TypeFlags.Never))) {
```
RyanCavanaugh, line 22060:
```suggestion
            if (!(target.flags & TypeFlags.Never) && isLiteralType(source) && !typeCouldHaveTopLevelSingletonTypes(target)) {

```

RyanCavanaugh (changes requested):

RyanCavanaugh:
I think this is a good fix; un-narrowing the source is going to make things more confusing. I can't commit the suggestion to the PR branch for some reason - please merge it in. Behavior changes LGTM.

src/compiler/checker.ts
hyzyla: I've added your suggestion manually, because I also cann't commit the suggestion using github ui.

RyanCavanaugh (approved):

src/compiler/checker.ts
```diff
@@ -22057,7 +22057,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             let generalizedSource = source;
             let generalizedSourceType = sourceType;
 
-            if (isLiteralType(source) && !typeCouldHaveTopLevelSingletonTypes(target)) {
+            if (!(target.flags & TypeFlags.Never) && isLiteralType(source) && !typeCouldHaveTopLevelSingletonTypes(target)) {
```
DanielRosenwasser, line 22060:
```suggestion
            // Don't generalize on 'never' - we really want the original type
            // to be displayed for use cases like 'assertNever'.
            if (!(target.flags & TypeFlags.Never) && isLiteralType(source) && !typeCouldHaveTopLevelSingletonTypes(target)) {
```

hyzyla: I've added suggestion manually

jakebailey:
This was merged skipping checks; not sure that was intended.

@hyzyla I would recommend sending PRs from branches on your fork; that's why we couldn't modify it.

Issue #59777, opened 2024-08-27
Multi-line top-level `await` causes duplicate declaration error


sethp:
### 🔎 Search Terms

typescript top-level await export class "duplicate identifier"

### 🕗 Version & Regression Information

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about _await, identifiers, modules_

### ⏯ Playground Link

https://www.typescriptlang.org/play/?ts=5.7.0-dev.20240827#code/AYKAFAtg9gJgrgGwKYgJBgGZwHYGMAEYSAHgA5QBOALvgEQCWAbACwB05pu2VtAlIaQCGFQRHxNm-MBSQBnRDXoBmAEy8Q+TfgRRcghKwDmSGgAYNWieyiduFzcpWsA7iNIB9Cb3WgA9L-wAN3pBfAALKipSWQAuf2ckACNBWVkkCESEAE8jeiowuETWeihfZ0FEql8YdNLyqhVy2QhfEFwobFkaJrEAXkJBcry0AHUkgEFU9Myc+k6qQW4QqiQAZSoZUTnDMDRUDBNcMLBaGEEFmMFSUgR6PSoS7DKUiABuZLSWABpxgHEAMQAXokAIoAUXGkPGACFoYZ-oZxgANZjQjDOACyAGFxiDxgAJAAihjBAGlBAAVACapgAMthoYFcL8AHJwGBQrGGQywxGI-E4LHOcak6G-ACsYUSIwAqjjnBDIb1enwvnsAN4AX28rxAbQ6XXE80WuCQ+H6PWKxrwSF1JHI1Hw7Xm+A47RwNH6cy6JqQrHtlCosgA2gByKxu7ihgC6+BShECMXwiXohjmVH4vQAfPhsHAMkgKLqQAHHbgEClZPhoXlViZ8Or7PgAMSGChQZxgXhakCavX+fDQeDIfA4doQCBIbjuSfuKhQdwYejEBuaoA

### 💻 Code

```ts
`
(module
	(func (export "i64.popcnt") (param i64) (result i32)
    local.get 0
    i64.popcnt
    i32.wrap_i64))
`
// via https://webassembly.github.io/wabt/demo/wat2wasm/
const wasm = (await
	WebAssembly.instantiateStreaming(
[...]

PR #59788, opened 2024-08-28
Fixed top level await reparsing for multiline await expressions

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/59777

### DIFF
--- a/src/compiler/parser.ts
+++ b/src/compiler/parser.ts
@@ -1891,7 +1891,15 @@ namespace Parser {
                     }
 
                     if (pos >= 0) {
-                        const nonAwaitStatement = sourceFile.statements[pos];
+                        let nonAwaitStatement = sourceFile.statements[pos];
+                        // skip all statements in range already completely consumed by the reparsed statement
+                        for (let i = pos + 1; i < sourceFile.statements.length; i++) {
+                            if (nonAwaitStatement.end >= statement.end) {
+                                break;
+                            }
+                            pos = i;
+                            nonAwaitStatement = sourceFile.statements[pos];
+                        }
                         if (statement.end === nonAwaitStatement.pos) {
                             // done reparsing this section
                             break;
--- a/tests/baselines/reference/topLevelAwait.4.symbols
+++ b/tests/baselines/reference/topLevelAwait.4.symbols
@@ -1,0 +1,13 @@
+//// [tests/cases/conformance/externalModules/topLevelAwait.4.ts] ////
+
+=== topLevelAwait.4.ts ===
+(await
+ WebAssembly.instantiateStreaming(fetch("")));
+>WebAssembly.instantiateStreaming : Symbol(WebAssembly.instantiateStreaming, Decl(lib.dom.d.ts, --, --))
+>WebAssembly : Symbol(WebAssembly, Decl(lib.dom.d.ts, --, --))
+>instantiateStreaming : Symbol(WebAssembly.instantiateStreaming, Decl(lib.dom.d.ts, --, --))
+>fetch : Symbol(fetch, Decl(lib.dom.d.ts, --, --))
+
+export class BitSet {}
+>BitSet : Symbol(BitSet, Decl(topLevelAwait.4.ts, 1, 46))
+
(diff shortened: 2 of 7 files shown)


src/compiler/parser.ts
```diff
@@ -1891,7 +1891,15 @@ namespace Parser {
                     }
 
                     if (pos >= 0) {
-                        const nonAwaitStatement = sourceFile.statements[pos];
+                        let nonAwaitStatement = sourceFile.statements[pos];
+                        // skip all statements in range already completely consumed by the reparsed statement
+                        for (let i = pos + 1; i < sourceFile.statements.length; i++) {
+                            if (nonAwaitStatement.end >= statement.end) {
+                                break;
+                            }
```
sandersn, line 1899:
```suggestion
                        for (let i = pos + 1; i < sourceFile.statements.length && sourceFile.statements[i].end < statement.end; i++) {
```

but it would actually be better to have a loop that increments pos as far as it needs to go and ends with `const nonAwaitStatement = sourceFile.statements[pos]`.

src/compiler/parser.ts
```diff
@@ -1891,7 +1891,15 @@ namespace Parser {
...
-                        const nonAwaitStatement = sourceFile.statements[pos];
+                        let nonAwaitStatement = sourceFile.statements[pos];
+                        // skip all statements in range already completely consumed by the reparsed statement
+                        for (let i = pos + 1; i < sourceFile.statements.length; i++) {
+                            if (nonAwaitStatement.end >= statement.end) {
+                                break;
+                            }
+                            pos = i;
+                            nonAwaitStatement = sourceFile.statements[pos];
+                        }
```
sandersn, line 1902: I think this code is equivalent except that it runs the check one time before incrementing `pos`:
```suggestion
                        // skip all statements in range already completely consumed by the reparsed statement
                        while (sourceFile.statements[pos] < statement.end) {
                            pos++;
                        }
                        const nonAwaitStatement = sourceFile.statements[pos];
```

sandersn (changes requested):
I haven't read this function before, so I really need to sit down with an editor and check how it's used. But given the local context, the fix looks right. I had a suggestion to make the loop easier to read, though.

Andarist:
@sandersn good shout, I simplified this loop as suggested - could you take another look?

src/compiler/parser.ts
```diff
@@ -1891,6 +1891,10 @@ namespace Parser {
                     }
 
                     if (pos >= 0) {
+                        // skip all statements in range already completely consumed by the reparsed statement
+                        while (sourceFile.statements[pos].end < statement.end) {
```
sandersn, line 1895: I missed this before, but this no longer checks that `pos < sourceFile.statements.length`, but it still should

src/compiler/parser.ts
```diff
@@ -1891,6 +1891,10 @@ namespace Parser {
                     }
 
                     if (pos >= 0) {
+                        // skip all statements in range already completely consumed by the reparsed statement
+                        while (sourceFile.statements[pos].end < statement.end) {
+                            pos++;
+                        }
                         const nonAwaitStatement = sourceFile.statements[pos];
                         if (statement.end === nonAwaitStatement.pos) {
```
sandersn, line 1899: I would like to exploit the parallelism between the `while`, `if` and `else if` to make the mathematical properties more obvious. Specifically, drop `nonAwaitStatement` and flip the order of comparisons in the two `if`s so that `sourceFile.statements[pos]` is always first. That way you can see the `while` skipping over sourceFiles.statements, then the two ifs checking whether the pos is exactly equal to the end, or has gone beyond it.

(It, uh, also points out what a bad name `pos` and `start` are, since they're so similar to the different meetings of pos/end in TextRange.)

```suggestion
                        if (sourceFile.statements[pos].pos === statement.end) {
```

sandersn (changes requested):
I looked at the whole function and I think this fix makes sense. One style request and one additional check.

rbuckton (approved):

rbuckton:
@typescript-bot perf test
@typescript-bot test top400

Issue #59790, opened 2024-08-29
Variance annotations in type aliases for interfaces do not affect assignability


bgenia:
### 🔎 Search Terms

variance annotations, type aliases, interfaces, assignability

### 🕗 Version & Regression Information

Previously noticed [here](https://github.com/microsoft/TypeScript/pull/56418#issuecomment-1814082905) for #56390 #56418, it still seems very strange. Is this behavior really intended?

This behavior exists since #56418 and looks like a bug.

### ⏯ Playground Link

[...]

PR #56418, opened 2023-11-15
Allow variance annotations on generic references

JoshuaKGoldberg:
Fixes #56390

Adds `ObjectFlags.Reference` to the allowed object flags, to encompass generic interfaces. Also adds tests to ensure that multi-layered `type` references to object or primitive (`number`) shapes still error only when they should.

Issue #56390, opened 2023-11-14
Variance annotations are not allowed in type aliases for interfaces


bgenia:
### 🔎 Search Terms

variance annotations

### 🕗 Version & Regression Information

- This seems to be a bug
- This is the behavior in every version I tried since variance annotations came out

### ⏯ Playground Link

https://www.typescriptlang.org/play?#code/C4TwDgpgBAKgYgewQHgGoEMA2BXCA+KAXigG8oA3LXALigx2gF8AoUSWAIXQCdkBLAHZQE2YHSr4isRCnq4CAegVQA8gGlmzQcAjcAZugDG0AJIy0EgmUoNacploE79R0116Dho8QwIQAHjoCACYAzlBmSBa+pIxQSqoarODQ7vxCImL2BMQmadnxygCi3NwI3MxAA

### 💻 Code

```ts
type TFoo<Value> = { value: Value }
type TBar<in out Value> = TFoo<Value> // OK

interface IFoo<Value> { value: Value }
interface IBar<in out Value> extends IFoo<Value> {} // OK

type Bar<in out Value> = IBar<Value> // Error
```

### 🙁 Actual behavior

The error says that variance annotations are only allowed for object types, but `IBar` is an object type.

### 🙂 Expected behavior

There should be no errors

RyanCavanaugh:
This is a correct error because it prevents a footgun with variance annotations.

So let's say you have
```ts
type TypeRef<T> = TypeTarget<T>;
type TypeTarget<T> = TypeFinal<T>;
type TypeFinal<T> = { value: T };
```

Given an assignability check from `TypeRef<Source>` to `TypeRef<Target>`, we could in theory do any of these three operations equivalently (since they all eventually end up at the same place):
* `Source` to `Target` (if `TypeRef` is covariant on `T`)
* `TypeTarget<Source>` to `TypeTarget<Target>`
* `TypeFinal<Source>` to `TypeFinal<Target>`
* `Source` to `Target` (via the `value` property)

For performance reasons, it's best to start as low in the stack as possible, so when there's an immediate type alias, we instantly go to the second bullet point and do the comparison starting there. This means any variance annotations present on `TypeRef` are ignored.

[...]

PR #59793, opened 2024-08-29
Revert "Allow variance annotations on generic references (#56418)"

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/59790

see the comment [here](https://github.com/microsoft/TypeScript/issues/59790#issuecomment-2316883138), cc @JoshuaKGoldberg @gabritto


tests/baselines/reference/varianceReferences.errors.txt
Andarist, line 1: the new assignability errors (for newly added assignments) are here the same as on `main` - that's why those variance annotations that were allowed shouldn't be allowed (because they are completely ignored anyway)

gabritto:
@typescript-bot test it

bgenia:
Will this reopen #56390?

Andarist:
If @RyanCavanaugh keeps the opinion that this is a bug then yes, it should be reopened. 

I could alternatively try to fix the assignability issue but it wasn’t clear to me if that should be fixed. So as the first step I decided that it’s better to have the checker rules in a cohesive state by reverting this PR and then hash out the desired behavior for this

gabritto:
@typescript-bot test top800

gabritto (approved):

Bongdautay:
quaroibn1@gmail.com

Bongdautay:
quaroibn1@gmail.com

PR #59794, opened 2024-08-29
Completions from template literal types

### LINKED ISSUE #59698: Suboptimal completion list for template literal type member
### 🔎 Search Terms

"completion list", "template literal", "prefix", "unexpected items"

- This is the behavior in every version I tried, and I reviewed the FAQ for entries.

### ⏯ Playground Link

https://www.typescriptlang.org/play/?ts=5.7.0-dev.20240820#code/C4TwDgpgBAKlC8UAGASA3gIjAJwgMwEsAPARgwB8tdCiAmDAX3QGdhsCA7AcwaQG4AUAIA2EYFCIAuWAigYMQA

### 💻 Code

```ts
type T = `${"prefix1"|"prefix2"}${string}`;

let x: T = ""
```

### 🙁 Actual behavior

When requesting a completion inside the `""`, I get `"$"`, `"let"`, `"prefix1"`, `"prefix2"`, `"string"`, `"T"`, `"type"`, `"x"`.

### 🙂 Expected behavior

I expect to get just `"prefix1"`, `"prefix2"`.

The current completions include strings that the value cannot legally start with.

MichalMarsalek:
Implements completions from types of the form
```
type T = `a${string}b${string}c...y${string}z`
```
, that is, from template literal types where `string` is the only placeholder. For the other cases (`prefix${number}`), it's much less clear to me how they should work.

The completions implemented in this PR suggest appending the "next" (the first that is not already included) text part of the template literal type. For example
```
const x: T = "a_/**/
```
suggests `"a_b`.

Closes #59698 

Please verify that:
* [x] There is an associated issue in the `Backlog` milestone (**required**)
* [x] Code is up-to-date with the `main` branch
* [x] You've successfully run `hereby runtests` locally
* [x] There are new or updated unit tests validating the change


src/services/stringCompletions.ts
```diff
@@ -297,6 +310,22 @@ function convertStringLiteralCompletions(
     }
 }
 
+function getStringLiteralCompletionFromTemplateLiteralTypeAndTextOfNode(type: TemplateLiteralType, tokenTextContent: string, quoteChar: CharacterCodes.backtick | CharacterCodes.singleQuote | CharacterCodes.doubleQuote) {
```
Andarist, line 313: There is an existing algorithm for matching strings against template literal types that you can find in `isTypeMatchedByTemplateLiteralType`. Instead of trying to implement a custom one, I'd recommend looking into how you could reuse that one. You can expose *some* checker functionality on the `TypeChecker` using new methods annotated with `/** @internal */`.

Perhaps you will have to focus on the algorithm in `inferFromLiteralPartsToTemplateLiteral`. It could return a number for the last segment matched successfully instead of `undefined`. Using this number you could look at the text after it and suggest that as autocompletion. 

Note that this is just a rough idea, I don't know how feasible it will be in practice 😉 

It could be nice to use the same algorithm to autocomplete properties coming from index signatures with template literal keys. You don't have to implement this as part of this PR though. One thing at a time 😅

MichalMarsalek: Hmm, this matching works in the escaped string (source code) space, while the checker's works in raw string content space. Shouldn't be the biggest issue though, instead of escaping the type.text, we would need to unescape token.content. Especially for future generalisations, this is probably a good direction to take.

MichalMarsalek: Also, `inferFromLiteralPartsToTemplateLiteral` starts by first checking that both the starts and the ends of the source and the target are compatible, which is not what we need here.

MichalMarsalek: Actually, I don't think that function is very useful for matching of the partial string. I thing that this requires a different algorithm by nature.

Andarist: Could u expand on that? At the very least the rules there could guide u how to handle number interpolations and more

MichalMarsalek: Sure, there are functions like `isValidTypeForTemplateLiteralPlaceholder` that could certainly prove usefule, but I'm not sure the overall algorithm that finds the target text parts in the source type is that reuseable. And

> it's much less clear to me how they should work

meant less "I don't know how to do it" and more "I don't know what it should be doing". Like in

 ```ts
const x: `a${number}` = "a/**/
```
what is the completion? I guess nothing? How about
 ```ts
const x: `a${number}` = "ab/**/
```
Maybe replacement of `"ab` with `"a` ?

PR #47050, opened 2021-12-07
Allow nongeneric string mapping types to exist

weswigham:
In the same way that `` `${number}` `` represents the set of all number strings, this lets `Uppercase<string>` represent the set of all strings unaffected by uppercasing (implying they're already uppercase), and the same for `Lowercase`, `Capitalize`, and `Uncapitalize`. As part of this, all of these string mapping types also are actually applied to template literal types, rather than... not. So ``Lowercase<`A${string}`>`` is now `` `a${Lowercase<string>}` ``, rather than the very clearly wrong `` `A${string}` ``. `Capitalize` and `Uncapitalize` only affect the first segment of a pattern template literal, reflecting their implementation, so ``Capitalize<`a${string}`>`` is just `` `A${string}` ``, while ``Capitalize<`${string}-${string}`>`` is `` `${Capitalize<string>}-${string}` ``.

Technically, this is a breaking change, since now `Uppercase<string>` isn't just `string`.

PR #59803, opened 2024-08-29
Remove redundant `getStringLiteralType` call

MichalMarsalek:
Is there a reason for this extra call instead of just using the source type directly? The tests pass the same without it. This was introduced in #47050.


jakebailey (approved):
I'm pretty sure this is correct.

jakebailey:
@typescript-bot test it

Issue #59804, opened 2024-08-29
TS2454 Variable is used before being assigned with strictNullChecks when variable is clearly assigned but can be undefined


raythurnvoid:
### 🔎 Search Terms

"Variable is used before being assigned" TS2454 TS(2454) undefined strictNullChecks

### 🕗 Version & Regression Information

- This happens at lest since version 3.7.5 (tested in the playground)
- This is the behavior in every version I tried, and I reviewed the FAQ

### ⏯ Playground Link

https://www.typescriptlang.org/play/?#code/DYUwLgBGIM6QvBArgOwCYgGYEsUjQNwBQR0cEiA5JRAIYwRwBOuA5hAD7Lpa77GlYYAPwA6UClZgAFgSA

### 💻 Code

```ts
let test = undefined;

test = '' as string | undefined;

test?.length;
```

### 🙁 Actual behavior

Typescript throws an error complaining about `test` being used before being assigned but it's clearly assigned to undefined and then to ''.

Mine is a very simplicistic problem, but the same issues happens if a replace `''` with a call to a function.

### 🙂 Expected behavior

Typescript should assign the type `string | undefined` to the variable `test` and let me use it.

andersk:
The same occurs with a value that doesn’t even have a `length` property.

```ts
let test = undefined;
test = 4 as number | undefined;
test?.length; // error TS2454: Variable 'test' is used before being assigned.
```

The expected `error TS2339: Property 'length' does not exist on type 'number'` is entirely missing. But this seems to be specific to the property `length` in particular—if I access any other nonexistent property like `width`, I get only TS2339 as expected, and if I access an existent property like `toFixed`, I get no errors as expected.

PR #59807, opened 2024-08-30
Fixed incorrect "used before assigned" errors for auto types

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/59804

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -30539,19 +30539,17 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         // A variable is considered uninitialized when it is possible to analyze the entire control flow graph
         // from declaration to use, and when the variable's declared type doesn't include undefined but the
         // control flow based type does include undefined.
-        if (!isEvolvingArrayOperationTarget(node) && (type === autoType || type === autoArrayType)) {
-            if (flowType === autoType || flowType === autoArrayType) {
-                if (noImplicitAny) {
-                    error(getNameOfDeclaration(declaration), Diagnostics.Variable_0_implicitly_has_type_1_in_some_locations_where_its_type_cannot_be_determined, symbolToString(symbol), typeToString(flowType));
-                    error(node, Diagnostics.Variable_0_implicitly_has_an_1_type, symbolToString(symbol), typeToString(flowType));
-                }
-                return convertAutoToAny(flowType);
-            }
-        }
-        else if (!assumeInitialized && type !== autoType && !containsUndefinedType(type) && containsUndefinedType(flowType)) {
+        if (!typeIsAutomatic && !assumeInitialized && !containsUndefinedType(type) && containsUndefinedType(flowType)) {
             error(node, Diagnostics.Variable_0_is_used_before_being_assigned, symbolToString(symbol));
             // Return the declared type to reduce follow-on errors
             return type;
+        }
+        if (flowType === autoType || flowType === autoArrayType && !isEvolvingArrayOperationTarget(node)) {
+            if (noImplicitAny) {
+                error(getNameOfDeclaration(declaration), Diagnostics.Variable_0_implicitly_has_type_1_in_some_locations_where_its_type_cannot_be_determined, symbolToString(symbol), typeToString(flowType));
+                error(node, Diagnostics.Variable_0_implicitly_has_an_1_type, symbolToString(symbol), typeToString(flowType));
+            }
+            return convertAutoToAny(flowType);
         }
         return assignmentKind ? getBaseTypeOfLiteralType(flowType) : flowType;
     }
--- a/tests/cases/compiler/useBeforeAssignedAutoType1.ts
+++ b/tests/cases/compiler/useBeforeAssignedAutoType1.ts
@@ -4,15 +4,15 @@
 // https://github.com/microsoft/TypeScript/issues/59804
 
 let test1 = undefined;
-test1 = '' as string | undefined;
+test1 = "" as string | undefined;
 test1?.length;
 
 let test2 = null;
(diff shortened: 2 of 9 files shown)


tests/cases/compiler/useBeforeAssignedAutoType1.ts
```diff
@@ -0,0 +1,37 @@
...
+let test7 = undefined;
+test7 = undefined;
+let other4 = test7;
+
+let test8 = null;
+test8 = undefined;
+let other5 = test8;
+
+let test9;
+let other6 = test9;
```
Andarist, line 37: I think this one could error but it's not easy to distinguish this case from others.

`flowType` is `undefinedType` in this case. In theory, I could probe this with an extra `getFlowTypeOfReference` call by supplying `autoType` as `initialType` and then check if `type !== probedFlowType` but I'm not really convinced it's worth the extra CFA call here.

RyanCavanaugh (approved):

RyanCavanaugh:

DanielRosenwasser:
@typescript-bot test it

RyanCavanaugh:
image

DanielRosenwasser:
Trying to figure out [why the following differed between these two variables](https://www.typescriptlang.org/play/?ts=5.7.0-dev.20240830#code/PTAEDkHtQUwJzpOA6AUAGxgF1AD37qALygCuAdgCYwBmAluTJQNyoGEkCyAhlgBbI43KpAC2ACgCUoAPygALACZQALjJVaDJqwDGkcgGdImZOkgBzcQAMAwqQQxyOAG7d0pGGoAkAb3YzkV3cYAHkaKQBfK0lWVBAQVDiwAAluA1gEJBkMbFAALwK84nVqekYWVEKirl4BIREJaTkAIiNRGFADLDgGc2bVEs1y3X0jEzNLW3s4Rxc3D28fKoDMcnN+KJjEoA):

```ts
// No error.
let xxxx = undefined;
xxxx = Math.random() ? 42 : undefined;
console.log(`Current value: ${xxxx?.valueOf()}`);

////

// Has error?
let zzzz = undefined;
zzzz = Math.random() ? "some string" : undefined;
console.log(`Current value: ${zzzz?.length}`);
```

Andarist:
@DanielRosenwasser `.length` satisfies `isEvolvingArrayOperationTarget`. I'll have to take another look at this PR tomorrow cause I missed the fact that this particular repro was hitting a somewhat specific branch when it comes to auto types (as usually, they go into the `if` above the one I touched).

src/compiler/checker.ts
```diff
@@ -30539,20 +30539,18 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         // A variable is considered uninitialized when it is possible to analyze the entire control flow graph
         // from declaration to use, and when the variable's declared type doesn't include undefined but the
         // control flow based type does include undefined.
-        if (!isEvolvingArrayOperationTarget(node) && (type === autoType || type === autoArrayType)) {
```
Andarist, line 30542: This block was moved below, this allows the most relevant code to the comment above this to be now directly below this comment. 

Other than that this condition got a new `!typeIsAutomatic &&` (coming from the first commit in this PR: https://github.com/microsoft/TypeScript/pull/59807/commits/3046304c125984581db05acee2b6c9989a83f80c and already approved by Ryan)

src/compiler/checker.ts
```diff
@@ -30539,20 +30539,18 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
-                return convertAutoToAny(flowType);
-            }
-        }
-        else if (!assumeInitialized && !containsUndefinedType(type) && containsUndefinedType(flowType)) {
+        if (!typeIsAutomatic && !assumeInitialized && !containsUndefinedType(type) && containsUndefinedType(flowType)) {
             error(node, Diagnostics.Variable_0_is_used_before_being_assigned, symbolToString(symbol));
             // Return the declared type to reduce follow-on errors
             return type;
         }
+        if (flowType === autoType || flowType === autoArrayType && !isEvolvingArrayOperationTarget(node)) {
```
Andarist, line 30547: I simplified this a little bit. We don't need to check if `type` is an auto type - it's sufficient to check `flowType`s. They can't be auto types if the `type` wasn't auto in the first place.

On top of that, `isEvolvingArrayOperationTarget` is only checked now when `flowType === autoArrayType`. This fixes an extra bug ([TS playground](https://www.typescriptlang.org/play/?ts=5.7.0-dev.20240830&ssl=11&ssc=2&pln=1&pc=1#code/PTAEDsHtQUwJzpOBnUheDcIx7AoANjALqPjMvgIwDMA3FgGYCu4AxvgJaTii3hkAUAlKADeWUKDgF6cTkNB5wAc3wALAFxES5CgDp5S5aAC+NQ1iwhYCJMgCE6QLI7uAqEgr4lGg2ZsOXcACYBYVFxSWlhLkhIdVdldx1aKKMTLCA)):
```ts
// no errors 😱
let test13;
function fn1() {
  return { length: test13.length };
}

// errors! 👍
let other13;
function fn2() {
  return { foo: other13.foo };
}
```

Andarist:
@DanielRosenwasser niiice catch! This highlighted an extra bug that I fixed now as part of this PR, see my comment [here](https://github.com/microsoft/TypeScript/pull/59807#discussion_r1739627859)

re uglify break~: this is a changed error which is kind of related to my comment [here](https://github.com/microsoft/TypeScript/pull/59807#discussion_r1738317079). Since, for example, `cur` is not initialized its `flowType` is `undefined`. So with this change TS stopped reporting "used before being assigned" and instead reported that the variable is potentially `undefined`. I think it's an acceptable change.

PR #59815, opened 2024-08-31
Provide string literal completions for flow properties

Andarist:
I noticed that even though this type checks OK:
```ts
type Test1<T> = "foo" extends keyof T ? T["foo"] : never;
```

completions do not suggest the `"foo"` property. This PR fixes that :)

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -769,6 +769,7 @@ import {
     isTypeReferenceType,
     isTypeUsableAsPropertyName,
     isUMDExportSymbol,
+    isUnaryTupleTypeNode,
     isValidBigIntString,
     isValidESSymbolDeclaration,
     isValidTypeOnlyAliasUseSite,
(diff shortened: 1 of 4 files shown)


src/compiler/checker.ts
```diff
@@ -16709,17 +16710,13 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
-    }
-
     function getImpliedConstraint(type: Type, checkNode: TypeNode, extendsNode: TypeNode): Type | undefined {
         return isUnaryTupleTypeNode(checkNode) && isUnaryTupleTypeNode(extendsNode) ? getImpliedConstraint(type, (checkNode as TupleTypeNode).elements[0], (extendsNode as TupleTypeNode).elements[0]) :
             getActualTypeVariable(getTypeFromTypeNode(checkNode)) === getActualTypeVariable(type) ? getTypeFromTypeNode(extendsNode) :
             undefined;
     }
 
-    function getConditionalFlowTypeOfType(type: Type, node: Node) {
+    function getFlowTypeOfType(type: Type, node: Node) {
```
jakebailey, line 16719: Why the rename?

Andarist: Originally this was handling flow types in conditional types alone. Later on, this function was expanded with some logic for mapped types but the function's name was never updated. I was on the fence about renaming it but it felt like it's too specific today when perhaps it shouldn't mention conditional types given its expanded scope

jakebailey:
Seems like this PR is conflicting.

Andarist:
> Seems like this PR is conflicting.

not anymore! :p

Issue #59402, opened 2024-07-23
Hovering @param comments duplicates the comment in the hover


Tyriar:
Testing microsoft/vscode#223154

<img width="1136" alt="Screenshot 2024-07-23 at 2 07 23 PM" src="https://github.com/user-attachments/assets/0d7b65b4-ecbc-4a3c-8937-bca3421c6d00">

<img width="771" alt="Screenshot 2024-07-23 at 2 07 57 PM" src="https://github.com/user-attachments/assets/026428f6-eb04-4bd8-9199-f4b837cd73b7">

mjbvz:
Minimal repo:

```ts
/**
 * @param x Does the thing
 */
function foo(x) {}
```

Example response after hovering over `x` in `@param x` showing the duplicated `tags` section:

```
Perf 621  [14:55:38.646] 204::quickinfo: elapsed time (in milliseconds) 5.1846
Info 622  [14:55:38.646] response:
    {"seq":0,"type":"response","command":"quickinfo","request_seq":204,"success":true,"body":{"kind":"parameter","kindModifiers":"","start":{"line":3,"offset":11},"end":{"line":3,"offset":12},"displayString":"(parameter) x: any","documentation":[{"text":"Does the thing","kind":"text"}],"tags":[{"name":"param","text":[{"text":"x","kind":"parameterName"},{"text":" ","kind":"space"},{"text":"Does the thing","kind":"text"}]}]}}
```

a-tarasyuk:
> duplicated tags

@mjbvz It looks like there are no duplicate tags, but there is a duplicate comment from the `param` tag, which is included in both the `documentation` and the `tags` info. Should comments related to params tags be included exclusively in the tags?


/cc @sandersn

mjbvz:
Yes, I could have been more clear: the entire `tags` section here should be omitted because we are already hovering over a specific tag. Instead we should return just tag content in the `documentation ` and leave `tags` empty

PR #59816, opened 2024-09-01
Don't return `tags` when returning quick info for JSDoc nodes

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/59402

### DIFF
--- a/src/services/symbolDisplay.ts
+++ b/src/services/symbolDisplay.ts
@@ -13,6 +13,7 @@ import {
     EnumMember,
     ExportAssignment,
     find,
+    findAncestor,
     first,
     firstDefined,
     forEach,
@@ -50,6 +51,7 @@ import {
     isFunctionLike,
     isIdentifier,
     isInExpressionContext,
+    isJSDoc,
     isJsxOpeningLikeElement,
     isLet,
     isModuleWithStringLiteralName,
@@ -730,7 +732,7 @@ function getSymbolDisplayPartsDocumentationAndSymbolKindWorker(typeChecker: Type
         }
     }
 
-    if (tags.length === 0 && !hasMultipleSignatures) {
+    if (tags.length === 0 && !hasMultipleSignatures && !findAncestor(location, isJSDoc)) {
         tags = symbol.getContextualJsDocTags(enclosingDeclaration, typeChecker);
     }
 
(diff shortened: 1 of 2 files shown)


tests/cases/fourslash/jsDocContextualTagsInJsDoc1.ts
```diff
@@ -0,0 +1,9 @@
+/// <reference path="fourslash.ts" />
+
+//// /**
+////  * @param x/*1*/ Does the thing
+////  */
+//// function foo(x/*2*/) {}
+
+verify.quickInfoAt("1", "(parameter) x: any", "Does the thing", undefined);
```
Andarist, line 8: This implements the change described by @mjbvz [here](https://github.com/microsoft/TypeScript/issues/59402#issuecomment-2246456638). Maybe *some* JSDoc tags should be returned when requesting quick info for a JSDoc node? I don't know when it would be desirable though. It seems that they always describe an associated node and not itself. The "itself" description gets still returned as part of the documentation string

tests/cases/fourslash/jsDocContextualTagsInJsDoc1.ts
```diff
@@ -0,0 +1,9 @@
...
+verify.quickInfoAt("2", "(parameter) x: any", "Does the thing", [{ name: "param", text: "x Does the thing" }]);
```
Andarist, line 9: I think that this isn't particularly desirable result either. It still contains duplicate information. I think this should be fixed as part of https://github.com/microsoft/TypeScript/issues/57819 though and not here

src/services/symbolDisplay.ts
```diff
@@ -730,7 +732,7 @@ function getSymbolDisplayPartsDocumentationAndSymbolKindWorker(typeChecker: Type
         }
     }
 
-    if (tags.length === 0 && !hasMultipleSignatures) {
+    if (tags.length === 0 && !hasMultipleSignatures && !findAncestor(location, isJSDoc)) {
```
jakebailey: Seems like this could just be:

```suggestion
    if (tags.length === 0 && !hasMultipleSignatures && !isInJSDoc(location)) {
```

So we don't walk all the way up when we already have a flag for it on the node?

Andarist: 👍 this does seem to do the trick too, I pushed out this suggested change

jakebailey (approved):

Issue #58943, opened 2024-06-20
JSDoc: property access within static initialization block of @extends annotated class from different file


Daniel-Knights:
### 🔎 Search Terms

"jsdoc class @extends", "static initialization block", "static properties/methods", "class extends", "class is referenced directly or indirectly in its own base expression"

### 🕗 Version & Regression Information

- Static initialization blocks don't seem to be supported on versions `<= 4.3.5`
- From `4.4.2` to `4.6.4` everything seems to work fine - types are correct and no linting errors
- Every version from `4.7.2` onwards has the behaviour described in this issue

### 💻 Code

`a.js`:
```js
export class A {
  static a() {}
}
```

`b.js`:
```js
// @ts-check
import { A } from "./a";

/**
 * @extends A
 */
export class B extends A { // 'B' is referenced directly or indirectly in its own base expression.ts(2506)
  static {
    B.a(); // Property 'a' does not exist on type 'typeof B'. ts(2339)
  }

  constructor() {
    super();

    B.a(); // Property 'a' does not exist on type 'typeof B'. ts(2339)
  }

  b() {
    B.a(); // Property 'a' does not exist on type 'typeof B'. ts(2339)
  }

  c = () => {
    B.a(); // Property 'a' does not exist on type 'typeof B'. ts(2339)
  };

  static d() {
    B.a(); // Property 'a' does not exist on type 'typeof B'. ts(2339)
  }

  get e() {
    B.a(); // Property 'a' does not exist on type 'typeof B'. ts(2339)

    return;
  }

  set e(value) {
    B.a(); // Property 'a' does not exist on type 'typeof B'. ts(2339)
  }
}
```

### 🙁 Actual behavior

- Linting error under extending class's name: `'B' is referenced directly or indirectly in its own base expression.`.
- Linting error/unknown type for any use of `B.a()`.

### 🙂 Expected behavior

`B.a()` type is correct across files with no linting errors.

### Additional information about the issue

- Removing `@extends` fixes the issue
- Removing `B.a()` from the `static` block fixes the issue
- Defining `A` in the same file fixes the issue

Possibly related to #51786.

Andarist:
bisects to https://github.com/microsoft/TypeScript/pull/44969

[...]

PR #59817, opened 2024-09-01
Fixed `.flowNode` binding on JSDoc nodes

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/58943

quoting [my comment](https://github.com/microsoft/TypeScript/issues/58943#issuecomment-2323222122) there:
> JSDoc is bound as a child of its associated declaration and static block declarations are bound as immediately invoked when binding the declaration itself. So this @extends has a flow node pointing to the body of the static block. So when the class' type is requested it tries to resolve the flow node of the super class it received, which leads to resolving the type of the class' in the static block (this bit is circular here).

### DIFF
--- a/src/compiler/binder.ts
+++ b/src/compiler/binder.ts
@@ -534,6 +534,7 @@ function createBinder(): (file: SourceFile, options: CompilerOptions) => void {
 
     // state used by control flow analysis
     var currentFlow: FlowNode;
+    var currentFlowJsDoc: FlowNode;
     var currentBreakTarget: FlowLabel | undefined;
     var currentContinueTarget: FlowLabel | undefined;
     var currentReturnTarget: FlowLabel | undefined;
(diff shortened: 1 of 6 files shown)


src/compiler/binder.ts
```diff
@@ -2791,14 +2794,18 @@ function createBinder(): (file: SourceFile, options: CompilerOptions) => void {
...
         }
         inStrictMode = saveInStrictMode;
+        currentFlowJsDoc = saveCurrentFlowJsDoc;
     }
 
     function bindJSDoc(node: Node) {
         if (hasJSDocNodes(node)) {
             if (isInJSFile(node)) {
+                const saveCurrentFlow = currentFlow;
+                currentFlow = currentFlowJsDoc;
```
Andarist, line 2804: I have also considered moving `bindJSDoc` outside of `bindChildren` and calling it earlier - before `bindContainer` and some other function. This seemed like more work and *maybe* there is a reason why it shouldn't be done - I hope that a reviewer might have some intuition for this 😅

PR #56552, opened 2023-11-26
Fixed symbol lookup for binding expando properties in blocks

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/56538

Issue #56538, opened 2023-11-25
Member mistakenly attached to expando function 


dragomirtitian:
### 🔎 Search Terms

expando functions members shadowed

### 🕗 Version & Regression Information

- This is the behavior in every version I tried

### ⏯ Playground Link

[Playground Link](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABADQBQEpEG9EF8CwAUDMKgLICGUAFgHQBOFYAJnALYaZZGK+IQIAzlBQAubIigBTYQH5xTAJ55EAXmwFCfFLWnC1iAIxFNQA)

### 💻 Code

```ts
function X() { }
if(Math.random()) {
    const X: { test?: any } = {}
    X.test = 1
}
```

### 🙁 Actual behavior

`test` appears in declarations as a member of `X`, even though the function `X` is shadowed by a local variable

### 🙂 Expected behavior

`test` should not appear on `X` as it does not belong to it.

### Additional information about the issue

Found while trying to detect expando functions in isolated declarations

PR #59828, opened 2024-09-02
Fixed symbol lookup for binding expando properties in nested blocks

Andarist:
fixes more issues like https://github.com/microsoft/TypeScript/issues/56538
I realized that my PR ( https://github.com/microsoft/TypeScript/pull/56552 ) that fixed that issue missed quite a bit of cases, cc @weswigham (as the reviewer of that previous PR)

1. This PR fixes lookup in nested blocks - when the expando is meant to be bound to a symbol in some outer block, between the current block and its enclosing container
2. It also fixes the a block/container lookup in more cases related to special properties, expando prototypes, auto this properties and `Object.defineProperty` expando members

### DIFF
--- a/src/compiler/binder.ts
+++ b/src/compiler/binder.ts
@@ -3391,7 +3391,7 @@ function createBinder(): (file: SourceFile, options: CompilerOptions) => void {
 
     function bindSpecialPropertyAssignment(node: BindablePropertyAssignmentExpression) {
         // Class declarations in Typescript do not allow property declarations
-        const parentSymbol = lookupSymbolForPropertyAccess(node.left.expression, blockScopeContainer) || lookupSymbolForPropertyAccess(node.left.expression, container);
+        const parentSymbol = lookupSymbolForPropertyAccess(node.left.expression);
         if (!isInJSFile(node) && !isFunctionSymbol(parentSymbol)) {
             return;
         }
@@ -3442,8 +3442,7 @@ function createBinder(): (file: SourceFile, options: CompilerOptions) => void {
                     return symbol;
                 }
                 else {
-                    const table = parent ? parent.exports! :
-                        file.jsGlobalAugmentations || (file.jsGlobalAugmentations = createSymbolTable());
+                    const table = parent ? parent.exports! : (file.jsGlobalAugmentations ??= createSymbolTable());
                     return declareSymbol(table, parent, id, flags, excludeFlags);
                 }
             });
@@ -3467,7 +3466,7 @@ function createBinder(): (file: SourceFile, options: CompilerOptions) => void {
         let includes = SymbolFlags.None;
         let excludes = SymbolFlags.None;
         // Method-like
-        if (isFunctionLikeDeclaration(getAssignedExpandoInitializer(declaration)!)) {
+        if (isFunctionLikeDeclaration(getAssignedExpandoInitializer(declaration))) {
             includes = SymbolFlags.Method;
             excludes = SymbolFlags.MethodExcludes;
         }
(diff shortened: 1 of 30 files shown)


src/compiler/binder.ts
```diff
@@ -3391,7 +3391,7 @@ function createBinder(): (file: SourceFile, options: CompilerOptions) => void {
 
     function bindSpecialPropertyAssignment(node: BindablePropertyAssignmentExpression) {
         // Class declarations in Typescript do not allow property declarations
-        const parentSymbol = lookupSymbolForPropertyAccess(node.left.expression, blockScopeContainer) || lookupSymbolForPropertyAccess(node.left.expression, container);
```
Andarist, line 3394: now this recursive lookup is handled solely by `lookupSymbolForPropertyAccess`, so all its callers can benefit from the same logic

src/compiler/binder.ts
```diff
@@ -3467,7 +3466,7 @@ function createBinder(): (file: SourceFile, options: CompilerOptions) => void {
         let includes = SymbolFlags.None;
         let excludes = SymbolFlags.None;
         // Method-like
-        if (isFunctionLikeDeclaration(getAssignedExpandoInitializer(declaration)!)) {
```
Andarist, line 3470: I added `| undefined` to `isFunctionLikeDeclaration` as that's a possible runtime value so now the non-null assertion isn't needed

src/compiler/binder.ts
```diff
@@ -3554,14 +3553,27 @@ function createBinder(): (file: SourceFile, options: CompilerOptions) => void {
...
+                const symbol = lookupSymbolForName(parent, node.escapedText);
+                if (symbol) {
+                    return symbol;
+                }
+            }
+            if (containerFlags & ContainerFlags.IsContainer) {
+                return;
+            }
+            parent = parent.parent;
+        }
```
Andarist, line 3576: The core of the fix is here

src/compiler/utilities.ts
```diff
@@ -1966,7 +1966,7 @@ function isShorthandAmbientModule(node: Node | undefined): boolean {
 }
 
 /** @internal */
-export function isBlockScopedContainerTopLevel(node: Node): boolean {
+export function isBlockScopedOrTopLevelContainer(node: Node): boolean {
```
Andarist, line 1969: I thought I'd have to use this, I didn't end up using it but I think the new name is better so I left this change here

src/compiler/utilitiesPublic.ts
```diff
@@ -1657,13 +1657,13 @@ export function isFunctionLike(node: Node | undefined): node is SignatureDeclara
 }
 
 /** @internal */
-export function isFunctionLikeOrClassStaticBlockDeclaration(node: Node | undefined): node is SignatureDeclaration | ClassStaticBlockDeclaration {
-    return !!node && (isFunctionLikeKind(node.kind) || isClassStaticBlockDeclaration(node));
+export function isFunctionLikeOrClassStaticBlockDeclaration(node: Node | undefined): node is FunctionLikeDeclaration | ClassStaticBlockDeclaration {
```
Andarist, line 1660: `isFunctionLikeDeclaration` handles value-space declaration whereas `isFunctionLikeKind` handles value and type space declarations. `isFunctionLikeOrClassStaticBlockDeclaration` always works within the value space so I think it's more appropriate to use `isFunctionLikeDeclaration` here

src/services/refactors/moveToFile.ts
```diff
@@ -1104,7 +1104,7 @@ function moveStatementsToTargetFile(changes: textChanges.ChangeTracker, program:
     }
 }
 
-function getOverloadRangeToMove(sourceFile: SourceFile, statement: Statement) {
+function getOverloadRangeToMove(sourceFile: SourceFile, statement: Statement | undefined) {
```
Andarist, line 1107: `statement === undefined` gets passed here in the test run so I adjusted this to match the reality more closely

Issue #59826, opened 2024-09-02
NoInfer doesn't work in string template types


sparecycles:
### 🔎 Search Terms

NoInfer string template

### 🕗 Version & Regression Information

- This changed between versions ______ and _______
- This changed in commit or PR _______
- This is the behavior in every version I tried, and I reviewed the FAQ for entries about _________
- I was unable to test this on prior versions because _______

### 💻 Code

```ts
function f<A>(a: NoInfer<A>, b: A extends string ? `rel:${A}` : never) {}
f("5", "rel:7"); // error - correct, type of A is "7"
```

but 

```ts
function g<A>(a: A, b: A extends string ? `rel:${NoInfer<A>}` : never) {}
g("5", "rel:7"); // no error - surprising - type of A is inferred to be ("5" | "7")
```

### 🙁 Actual behavior

no type error on `g()` above, because the `A` type is being inferred via the string template in spite of the `NoInfer` marker.

### 🙂 Expected behavior

Both of the example calls should be recognized as wrong?

### Additional information about the issue

Also tried 

```ts
function g<A>(a: A, b: A extends string ? NoInfer<`rel:${A}`> : never) {}
function g<A>(a: A, b: NoInfer<A extends string ? `rel:${A}` : never>) {}
```

no luck.

Andarist:
It's not about template literal types but rather about `NoInfer` used in a conditional type. This already doesn't work correctly:
```ts
function h<const A>(a: A, b: A extends string ? NoInfer<A> : never) {}
h("foo", "bar");
```

sparecycles:
[...]

PR #59834, opened 2024-09-03
Fixed substition flow types of `NoInfer` types

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/59826

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -16773,7 +16773,14 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             }
             node = parent;
         }
-        return constraints ? getSubstitutionType(type, getIntersectionType(constraints)) : type;
+        if (!constraints) {
+            return type;
+        }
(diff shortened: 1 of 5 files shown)


src/compiler/checker.ts
```diff
@@ -16773,7 +16773,14 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
             node = parent;
         }
-        return constraints ? getSubstitutionType(type, getIntersectionType(constraints)) : type;
+        if (!constraints) {
+            return type;
+        }
+        if (isNoInferType(type)) {
+            constraints = append(constraints, (type as SubstitutionType).baseType);
+            return getNoInferType(getIntersectionType(constraints));
+        }
```
Andarist, line 16782: Alternatively, I was thinking about handling the substitution types created here today (with `NoInfer` base types and non-unknown constraints) in `isNoInferType` (or at least handling such at the top of `inferFromTypes`). I concluded this is better and less error-prone though

Issue #59831, opened 2024-09-02
macOS: TypeScript installs recursive file watchers on `~/Library` leading to access prompts


bpasero:
### 🔎 Search Terms

file watcher

### 🕗 Version & Regression Information

- This changed between versions 5.5 and 5.4 with the introduction of TS using VS Code for file watching.

### 🙁 Actual behavior

We have seen a user in https://github.com/microsoft/vscode/issues/208105 report that TypeScript installed a recursive file watcher on `~/Library`. This can trigger permission dialogs on macOS starting from 14.x.

![image](https://github.com/user-attachments/assets/6d675d08-5605-4fbd-a3eb-460a26c89a9e)

```
2024-09-01 16:19:25.676 [trace] MainThreadFileSystemEventService#$watch(): 
request to start watching correlated (extension: vscode.typescript-language-features, 
path: file:///Users/tmm1/Library, recursive: true, session: 0.7154528361377308)
```

### 🙂 Expected behavior

TypeScript should not install recursive file watchers so broadly.

//cc @DanielRosenwasser @sheetalkamat maybe you could comment how such a path can be used for watching? Are there global installations of TypeScript in `~/Library`?

tmm1:
Here is a quick and dirty patch:

```diff
diff --git a/extensions/typescript-language-features/src/tsServer/fileWatchingManager.ts b/extensions/typescript-language-features/src/tsServer/fileWatchingManager.ts
index 27b9ec3d7f6c..d9a4e02ee7db 100644
--- a/extensions/typescript-language-features/src/tsServer/fileWatchingManager.ts
+++ b/extensions/typescript-language-features/src/tsServer/fileWatchingManager.ts
@@ -60,6 +60,12 @@ export class FileWatcherManager implements IDisposable {
 		if (watchParentDirs && uri.scheme !== Schemes.untitled) {
 			// We need to watch the parent directories too for when these are deleted / created
 			for (let dirUri = Utils.dirname(uri); dirUri.path.length > 1; dirUri = Utils.dirname(dirUri)) {
+				// on macOS, avoid turning `~/Library/Caches/typescript/5.5/...` into a watcher on `~/Library`
+				// this prevents security alerts, due to a parcel-bundler/watcher bug trying to open(~/Library/Containers)
+				// https://github.com/microsoft/vscode/issues/208105#issuecomment-2323462416
+				if (dirUri.path.endsWith('/Library/Caches')) {
+					break;
+				}
[...]

PR #59835, opened 2024-09-03
On macOS prevent inadvertent watchDirectory(~/Library, WatchType.FailedLookupLocations) due to globalTypingsCacheLocation

tmm1:
Fixes https://github.com/microsoft/TypeScript/issues/59831


sheetalkamat (commented):
We need repro tsserver code so we can add test and ensure this is the right way to fix. I dont think this is the right fix.

PR #59844, opened 2024-09-03
Refactor tsserver, tsc and fourslash-server tests so that paths are always watchable

sheetalkamat:
This is precursor to increase the coverage for tsserver resolution sharing:

- This ensures that in tsserver, tsc tests the libs are always present.
- Current directory for tsserver  and fourslash-server tests is "vscode" like path thats watchable: `/home/src/vscode/projects/bin`
- Current directory for tsc tests needs to be specified all the time (ensures its watchable)
- Ensured globaltypings location is always watchable: `/home/src/typinginstaller/globalcache/data`
- Libs is separate location and is watchable: `/home/src/tslibs/ts/lib`

This discovered issues:
- Typing installer test failing because of using incorrect "current directory": Fixed with 324fb89 with which test called "external project - deduplicate from local @types packages" starts passing again.
- inferred projects use current directory as "vscode" (which is known but need to investigate again) - This is also probably reason we never catch  #59831 since its setting directory watcher in vscode directory which in tests framework use to be "/"
- Some of the semantic diagnostics use - full paths rather than relative paths - and this will cause issues with "tsbuildinfo" portability #59851
- This also found that some of the tests used incorrect "lib" in fourslash-server tests eg `tests/cases/fourslash/server/autoImportProvider_exportMap3.ts` because it was matched by tsconfig at "/". I left it like that where it fails to resolve lib and updated test to match "includes" and "excludes" instead of exact: "globalPlus" + "include" as globals for esnext i didnt want to encode. Also i remembered some conversation about trying to use default lib unless test absolutely needs it so left it like that.


src/harness/harnessLanguageService.ts
```diff
@@ -249,14 +249,16 @@ export abstract class LanguageServiceAdapterHost {
...
+export const harnessTypingInstallerCacheLocation = "/home/src/typinginstaller/globalcache/data";
 /// Native adapter
 class NativeLanguageServiceHost extends LanguageServiceAdapterHost implements ts.LanguageServiceHost, LanguageServiceAdapterHost {
     isKnownTypesPackageName(name: string): boolean {
         return !!this.typesRegistry && this.typesRegistry.has(name);
     }
 
     getGlobalTypingsCacheLocation() {
-        return "/Library/Caches/typescript";
+        return harnessTypingInstallerCacheLocation;
```
andrewbranch, line 261: Are we missing any coverage by having the new values all lowercase?

tests/cases/fourslash/server/autoImportFileExcludePatterns1.ts
```diff
@@ -2,19 +2,19 @@
 
 // @module: commonjs
 
-// @Filename: /project/node_modules/aws-sdk/package.json
+// @Filename: /home/src/workspaces/project/node_modules/aws-sdk/package.json
```
andrewbranch, line 5: I can understand if we don’t want to test files in the disk root like `/tsconfig.json`, but in this test, everything was inside `/project/*`. Did this test actually need to change, or was this just a bulk find and replace? Is the `/home/src/workspaces` prefix significant?

sheetalkamat: This does need to change because we don’t watch directories with /soemthing/anotherdir/

src/harness/harnessLanguageService.ts
sheetalkamat: Good point I will make it with upper case

tests/cases/fourslash/server/autoImportFileExcludePatterns1.ts
sheetalkamat: Just as an example if you did not start file name with "/" they are relative to "/tests/case/fourslash" and i didnt need to change those

andrewbranch: Directories have to be 2 or more levels deep?

sheetalkamat: Yes. https://github.com/microsoft/TypeScript/blob/main/src/compiler/resolutionCache.ts#L307

sheetalkamat:
@andrewbranch ready for re-review. Thanks

andrewbranch (approved):

PR #59853, opened 2024-09-04
Include type-meaning 'extends' clause in auto-import fix list

### LINKED ISSUE #59842: Cannot import Disposable
* in the vscode repo, open `src/vs/base/common/strings.ts`
* type:
```
class X extends Disposable {
}
```
* do cmd+.
* observe that the import Disposable from lifecycle.ts quick fix is not offerred 

Image

RyanCavanaugh:
If you have something like
```ts
// In global or local scope
interface Foo {
}

// User is editing here
class Bar extends Foo { }
//                ~~~
// Cannot extend interface, did you mean implements?

// In another file
export class Foo { }
```
This error is potentially fixable with an auto-import, so should appear in the list of errors that trigger autoimports

Checked locally
image

Fixes #59842


DanielRosenwasser:
Seems right, can you just add a quick test case?

jakebailey (approved):

PR #59857, opened 2024-09-04
Clarify logging for "No Project" when hitting memory limit

### LINKED ISSUE #59856: Clarify logging for "No Project" when hitting memory limit
### Acknowledgement

### Comment

typescript fails with "No Project" and no other error in its log when hitting default memory limit. Improve this by providing better logging output.

cbarber:
typescript fails with "No Project" and no other error in its log when hitting default memory limit. Improve this by providing better logging output.

Fixes #59856

### DIFF
--- a/src/server/project.ts
+++ b/src/server/project.ts
@@ -1053,7 +1053,7 @@ export abstract class Project implements LanguageServiceHost, ModuleResolutionHo
         }
     }
 
-    disableLanguageService(lastFileExceededProgramSize?: string) {
+    disableLanguageService(lastFileExceededProgramSize: string) {
         if (!this.languageServiceEnabled) {
             return;
         }
--- a/src/server/project.ts
+++ b/src/server/project.ts
@@ -1071,6 +1071,8 @@ export abstract class Project implements LanguageServiceHost, ModuleResolutionHo
         this.clearGeneratedFileWatch();
         this.projectService.verifyDocumentRegistry();
         this.projectService.onUpdateLanguageServiceStateForProject(this, /*languageServiceEnabled*/ false);
+
+        this.error("Project service was disabled for exceeding memory limit. See disableSizeLimit config option.");
     }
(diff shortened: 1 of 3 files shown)


src/server/project.ts
```diff
@@ -1071,6 +1071,8 @@ export abstract class Project implements LanguageServiceHost, ModuleResolutionHo
         this.clearGeneratedFileWatch();
         this.projectService.verifyDocumentRegistry();
         this.projectService.onUpdateLanguageServiceStateForProject(this, /*languageServiceEnabled*/ false);
+
+        this.error("Project service was disabled for exceeding memory limit. See disableSizeLimit config option.");
```
sheetalkamat, line 1075: This is already logged. And shouldnt be logged as error.

src/server/session.ts
```diff
@@ -2089,7 +2089,12 @@ export class Session<TMessage = string> implements EventSender {
         projects = filter(projects, p => p.languageServiceEnabled && !p.isOrphan());
         if (!ignoreNoProjectError && (!projects || !projects.length) && !symLinkedProjects) {
             this.projectService.logErrorForScriptInfoNotFound(args.file ?? args.projectFileName);
-            return Errors.ThrowNoProject();
+            if (some(projects, p => !p.languageServiceEnabled)) {
+                Errors.ThrowProjectLanguageServiceDisabled();
```
sheetalkamat, line 2093: Can you show case where this happens. I dont think we ever check language service disabled when getting project

sheetalkamat (changes requested):
This is incorrect change

src/server/project.ts
cbarber: Where is it logged?

sheetalkamat: getFilenameForExceededTotalSizeLimitForNonTsFiles

PR #59860, opened 2024-09-04
Support interpreting non-literal computed properties in classes as implicit index signatures

weswigham:
Instead of silently dropping them (if they're declared by methods) as we do today.

This still has some minor issues and open questions, but is far enough along to share and talk about. Today, when you have a class like
```ts
declare const a: symbol;
export class A {
    [a]() { return 1 };
}
```
that's just fine - no errors anywhere. But if you add a
```ts
declare const e1: A[typeof a];
```
you get an error that `A` has no symbol index signature. What's up with that? Well, we dropped the `[a]` member entirely, silently. There are historical reasons for this, but they don't really hold much weight nowadays beyond "for back compat". Under non-strict, this can actually be insidious, as when you use an expression, the `(new A())[a]` expression can silently be `any`, since under non-strict mode, non-existing element accesses resolve to `any` without error. With this PR, `A[typeof a]` instead resolves to `() => number` (the method's type) as we've created an implicit index signature for the computed name, just like we do for object literals.

Technically, this involves late-binding the `__index` member of the class like we do other computed literal names (so you could call these "late bound index signatures" as I do in the code), but that's not so important compared to the effect. This is pretty different from how these kinds of indexes are built for object literals, but that's because object literals are built somewhat on-demand from their constituent declarations, rather than having a big table of declared metadata built up-front like classes and interfaces do.

The big questions I'm still considering are:
* Should all the indexes for the same key types get combined? Right now for objects they do, but in the implementation, the first one decides the signature type and errors are issued on any others that don't match. (Is our behavior for objects good?)
* Should the error on non-literal computed properties, `A computed property name in a class property declaration must have a simple literal type or a 'unique symbol' type.(1166)` be removed, too? Methods don't emit an error, which is what motivated me here, but all this logic equally applies to properties, so the error doesn't make too much sense for them now.
* Should this require opt-in via `noImplicitAny` or another (strict?) flag? (preferably not - this being flagged somewhat defeats the point, imo)

And, ofc, this needs more tests, covering other key types (unions, pattern literals) and modifiers (static, readonly) on the computed names (though I've informally handled them in the code, I think) before this can actually be merged (though these probably have some coverage already from documenting our current behavior - they could stand to be more comprehensive).

This is, IMO, a worthwhile step towards unifying all of our computed property name behaviors. In an ideal world, a computed property name in a class, object, in a source or declaration file, could all be interpreted the same way (and thus, it could always be safe to copy one from one place or file kind into the other). Unfortunately, today we have some behavioral carve-outs from the time before we even meaningfully supported literal types, or even had `--strict`, that get in the way of that, like this.

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -583,6 +583,7 @@ import {
     isImportTypeNode,
     isInCompoundLikeAssignment,
     isIndexedAccessTypeNode,
+    isIndexSignatureDeclaration,
     isInExpressionContext,
     isInfinityOrNaNString,
     isInitializedProperty,
@@ -878,6 +879,7 @@ import {
     ModuleResolutionKind,
     ModuleSpecifierResolutionHost,
     Mutable,
+    MutableNodeArray,
     NamedDeclaration,
     NamedExports,
     NamedImportsOrExports,
@@ -1121,8 +1123,6 @@ import {
     WideningContext,
     WithStatement,
     YieldExpression,
-    isIndexSignatureDeclaration,
-    MutableNodeArray,
 } from "./_namespaces/ts.js";
 import * as moduleSpecifiers from "./_namespaces/ts.moduleSpecifiers.js";
 import * as performance from "./_namespaces/ts.performance.js";
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -1121,6 +1121,7 @@ import {
     WideningContext,
     WithStatement,
     YieldExpression,
+    isIndexSignatureDeclaration,
 } from "./_namespaces/ts.js";
 import * as moduleSpecifiers from "./_namespaces/ts.moduleSpecifiers.js";
 import * as performance from "./_namespaces/ts.performance.js";
@@ -13344,12 +13345,25 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
      * - The type of its expression is a string or numeric literal type, or is a `unique symbol` type.
      */
     function isLateBindableName(node: DeclarationName): node is LateBoundName {
+        return isLateBindableAST(node)
+            && isTypeUsableAsPropertyName(isComputedPropertyName(node) ? checkComputedPropertyName(node) : checkExpressionCached((node as ElementAccessExpression).argumentExpression));
+    }
+
+    function isLateBindableIndexSignature(node: DeclarationName): node is LateBoundName {
+        return isLateBindableAST(node)
+            && isTypeUsableAsIndexSignature(isComputedPropertyName(node) ? checkComputedPropertyName(node) : checkExpressionCached((node as ElementAccessExpression).argumentExpression));
+    }
+
+    function isLateBindableAST(node: DeclarationName) {
         if (!isComputedPropertyName(node) && !isElementAccessExpression(node)) {
             return false;
         }
         const expr = isComputedPropertyName(node) ? node.expression : node.argumentExpression;
-        return isEntityNameExpression(expr)
-            && isTypeUsableAsPropertyName(isComputedPropertyName(node) ? checkComputedPropertyName(node) : checkExpressionCached(expr));
+        return isEntityNameExpression(expr);
+    }
+
+    function isTypeUsableAsIndexSignature(type: Type): boolean {
+        return isTypeAssignableTo(type, stringNumberSymbolType);
     }
 
     function isLateBoundName(name: __String): boolean {
@@ -13364,6 +13378,11 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
     function hasLateBindableName(node: Declaration): node is LateBoundDeclaration | LateBoundBinaryExpressionDeclaration {
         const name = getNameOfDeclaration(node);
         return !!name && isLateBindableName(name);
+    }
+
+    function hasLateBindableIndexSignature(node: Declaration) {
+        const name = getNameOfDeclaration(node);
+        return !!name && isLateBindableIndexSignature(name);
     }
 
     /**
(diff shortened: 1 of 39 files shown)


weswigham:
@typescript-bot pack this

weswigham:
@typescript-bot test it

jakebailey:
Seems like the angular codebase is gaining new errors?

weswigham:
@typescript-bot test it

Now I'm using *exactly* the same index signature creation logic we use for object literal expressions, with all of it's flaws (like making a lot of `number` index signatures you may not expect). Let's see how this goes.

weswigham:
Hm. Still some new errors in angular, but everything else looks fine now. Guess I'll have to look at what those are.

weswigham:
@typescript-bot test it

weswigham:
There we go, no more new errors in angular - now just to see if top400 comes back cleaner, too. (I think it should, the most recent fix is squarely aimed at those errors, since they're the same cause as the ones in angular - forgetting to include the explicit prop types in the implied index signature type)

src/compiler/checker.ts
```diff
@@ -16204,24 +16252,67 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+                        }
+                        computedPropertySymbols.push(declaration.symbol);
                     }
                 }
             }
+            const allPropertySymbols = concatenate(computedPropertySymbols, filter(siblingSymbols, s => s !== indexSymbol));
+            // aggregate similar index infos implied to be the same key to the same combined index info
+            if (hasComputedStringProperty && !findIndexInfo(indexInfos, stringType)) indexInfos.push(getObjectLiteralIndexInfo(readonlyComputedStringProperty, 0, allPropertySymbols, stringType));
+            if (hasComputedNumberProperty && !findIndexInfo(indexInfos, numberType)) indexInfos.push(getObjectLiteralIndexInfo(readonlyComputedNumberProperty, 0, allPropertySymbols, numberType));
+            if (hasComputedSymbolProperty && !findIndexInfo(indexInfos, esSymbolType)) indexInfos.push(getObjectLiteralIndexInfo(readonlyComputedSymbolProperty, 0, allPropertySymbols, esSymbolType));
```
jakebailey, line 16315: Are all of the `findIndexInfo` calls likely to cost a bunch or is `indexInfos` just always short?

jakebailey (approved):

src/compiler/checker.ts
weswigham: In practice, it's always short (there's usually just `string` or `number` indexes) - in theory, a user could explicitly define hundreds of pattern literal index infos and it'd be bad, however, and you'd prefer you did the up-front work to make a map keyed on key type (if you had multiple kinds of computed names). But that's not code anyone's realistically writing.

pfumagalli:
Hmm... I hit a bit of a snag here. I have a class like this:

```typescript
import util from 'node:util'
const inspect = util.inspect.custom

export class MyMap extends Map<MyObjectA, MyObjectB> {
  [inspect](depth: number, options: util.InspectOptions): string {
    // my fancy code for making this look pretty...
  }
  // ... and here goes all the extra stuff I want
}
```

Now this compiles out as a `.d.ts` as:

```typescript
export declare class MyMap extends Map<MyObjectA, MyObjectB> {
  [x: symbol]: ((depth: number, options: util.InspectOptions) => string) 
             | (() => MapIterator<[string, string]>);
  // ... all the other niceties
}
```

... so far so good (maybe???) but when I try to import my map in some other project I get an error:

```
Property '[Symbol.toStringTag]' of type 'string' is not assignable to 'symbol' index type '((depth: number, options: InspectOptions) => string) | (() => MapIterator<[string, string]>)'. [TS2411]
    [x: symbol]: ((depth: number, options: util.InspectOptions) => string) | (() => MapIterator<[string, string]>);
```

Why? Because `` declares `Map` as follows, which is obviously incompatible:

```typescript
interface Map<K, V> {
    readonly [Symbol.toStringTag]: string;
}
```

Maybe the typescript compiler should merge all symbol declarations and write something like:

```typescript
export declare class MyMap extends Map<MyObjectA, MyObjectB> {
  [x: symbol]: string 
             | ((depth: number, options: util.InspectOptions) => string) 
             | (() => MapIterator<[string, string]>);
  // ... all the other niceties
}
```

I don't know what'd be the _right_ solution in these cases.

For my code, I can simply remove that assignment `const inspect = util.inspect.custom` and use it directly in my code:

```typescript
import util from 'node:util'

export class MyMap extends Map<MyObjectA, MyObjectB> {
  [util.inspect.custom](depth: number, options: util.InspectOptions): string {
    // my fancy code for making this look pretty...
  }
  // ... and here goes all the extra stuff I want
}
```

produces a more correct `.d.ts` as follows (which makes dependant libraries work):

```typescript
import util from 'node:util';

export declare class MyMap extends Map<MyObjectA, MyObjectB> {
  [util.inspect.custom]: ((depth: number, options: util.InspectOptions) => string) 
                       | (() => MapIterator<[string, string]>);
  // ... all the other niceties
}
```

But I noticed that anytime I import a `unique symbol` (such as in the case of `utils.inspect.custom`) assign it locally, and use it as a key in a class, the _uniqueness_ of the symbol is somehow lost and the whole thing collapses...

farfromrefug:
@weswigham i have an issue with that change.
I have a class like this:
```ts
class Property {
    readonly getDefault: symbol;
    constructor(propertyName:string) {
        const getDefault = Symbol(propertyName + ':getDefault');
        this.getDefault = getDefault;
    }
}
```
And in another class i do this:
```ts
const myProp =  new Property()
class A {
    [myProp.getDefault](): boolean {
        return true;
    }
}
```
Before TS 5.7 it was working correctly. Now because of this PR it reports an error:
```
Property '[myProperty.getDefault]' of type '() => boolean' is not assignable to 'symbol' index type '() => string'.ts(2411)
```
And i really dont see how to fix it. Can you help?

PR #59874, opened 2024-09-05
Updates to --build parsing on command line

### LINKED ISSUE #59095: parseCommandLine fails with --build and --force

sheetalkamat:
- Moves "--build" from compiler options to "options for build"-
- Made `isBuildCommand`, `parseBuildCommand` and `ParsedBuildCommand` public
Fixes #59095


sheetalkamat:
@jakebailey @navya9singh i am hoping this does not affect website in some way. Let me know

jakebailey:
Checked, and I don't think this breaks the website's tsconfig generation script; build is still listed as an option there when I build this PR and then pull it into the website.

Issue #58342, opened 2024-04-27
Watch cannot handle large amount of file changes


2767mr:
### 🔎 Search Terms

watch, tsc, tsserver, files, Cannot find module, or its corresponding type declarations

### 🕗 Version & Regression Information

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about (tsc v5.4.5)

### ⏯ Playground Link

https://github.com/2767mr/tsbug

### 🙁 Actual behavior

After deleting and regenerating a large amount of auto-generated files, typescript reports an error as seen below when using `--watch`. This also affects the VS Code tsserver.
```
src/main.ts:1:32 - error TS2307: Cannot find module './app/services/generated' or its corresponding type declarations.

1 import { TranslationDto } from "./app/services/generated";
``` 

### 🙂 Expected behavior

Typescript should pick up new files even when there are a lot of them.

### Additional information about the issue

Reproduction:
1. Clone https://github.com/2767mr/tsbug
2. `npm install`
3. Open one terminal with `npm run watch`
4. After a few seconds, open another terminal with `npm run copy`

Here is the tsserver log that is created when doing this: [tsserver.log](https://github.com/microsoft/TypeScript/files/15138615/tsserver.log)

Possibly related issues:
#57330
#22710
#37514
#54511 (same use case as we have - but on our end it does not work at all instead of being slow)

2767mr:
Note that this bug does not happen if you reduce the amount of files that will be copied by `npm run copy`

sheetalkamat:
Thank you for repro as thats the most helpful and crucial in investigating such issues.

PR #59882, opened 2024-09-06
We may not always get file notifications for file creation, so always check for failed lookups starting with the path as well as exact path

sheetalkamat:
Fixes https://github.com/microsoft/TypeScript/issues/58342


jakebailey (approved):

jakebailey:
Another spurious sys test error; been happening fairly often everywhere in CI (on all OSs I think)

sheetalkamat:
> Another spurious sys test error; been happening fairly often everywhere in CI (on all OSs I think)

unfortunately it does not repro on my windows or linux machine ever: This one is in "fs" polling watch api

PR #59884, opened 2024-09-07
fix59799

### LINKED ISSUE #59799: Import named `Disposable` is not moved when moving code to another file
```ts
// @filename: src/_producer.ts
export class Disposable {}
```

```ts
// @filename: src/_consumer.ts
import { Disposable } from './_producer';

class Something extends Disposable {
}
```
Doing move to a new file on `Something` generates the following 🐛 (note how `Disposable` is not imported from `_producer.ts`):

```ts
// @filename: src/Something.ts
class Something extends Disposable {
}

```

iisaduan:
related #59799

### DIFF
--- a/tests/cases/fourslash/moveToNewFile_disposable.ts
+++ b/tests/cases/fourslash/moveToNewFile_disposable.ts
@@ -1,0 +1,26 @@
+
+/// <reference path='fourslash.ts' />
+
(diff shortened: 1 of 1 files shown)


iisaduan:
My bad, I forgot I already had a branch/PR for this fix when I made my other branch.

PR #59885, opened 2024-09-07
add support for autoimports/moveToFile to generate aliased named imports

### LINKED ISSUE #59798: Alias import is moved incorrectly when moving code to another file
```ts
// @filename: src/_producer.ts
export function doit() {}
```

```ts
// @filename: src/_consumer.ts
import { doit as doit2 } from "./_producer";

class Another {}

class Consumer {
    constructor() {
        doit2();
    }
}
```
Doing move to a new file on Consumer generates the following 🐛 (note how doit2 does not exist in `_producer.ts`):

```ts
// @filename: src/Consumer.ts
import { doit2 } from "./_producer";

class Consumer {
    constructor() {
        doit2();
    }
}

```

iisaduan:
Fixes #59798

### DIFF
--- a/src/services/codefixes/importFixes.ts
+++ b/src/services/codefixes/importFixes.ts
@@ -241,7 +241,7 @@ export function createImportAdder(sourceFile: SourceFile | FutureSourceFile, pro
 interface AddToExistingState {
     readonly importClauseOrBindingPattern: ImportClause | ObjectBindingPattern;
     defaultImport: Import | undefined;
-    readonly namedImports: Map<string, AddAsTypeOnly>;
+    readonly namedImports: Map<string, [AddAsTypeOnly, /*propertyName*/ string | undefined]>;
 }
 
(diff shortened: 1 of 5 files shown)


src/services/codefixes/importFixes.ts
```diff
@@ -241,7 +241,7 @@ export function createImportAdder(sourceFile: SourceFile | FutureSourceFile, pro
 interface AddToExistingState {
     readonly importClauseOrBindingPattern: ImportClause | ObjectBindingPattern;
     defaultImport: Import | undefined;
-    readonly namedImports: Map<string, AddAsTypeOnly>;
+    readonly namedImports: Map<string, [AddAsTypeOnly, /*propertyName*/ string | undefined]>;
```
andrewbranch, line 244: Nit: tuples support real labels! Although, maybe an object type would be better at that point.

andrewbranch (approved):

Issue #59868, opened 2024-09-05
Constraints of nested conditional types applied to constrained type variables are incorrect


Andarist:
### 🔎 Search Terms

constraints conditional types type variables nested

### 🕗 Version & Regression Information

- This is the behavior in every version I tried

### ⏯ Playground Link

https://www.typescriptlang.org/play/?ts=5.7.0-dev.20240904#code/C4TwDgpgBAkgzgWQIbAMYAsCWA7A5gZWACcdcB5bCAGQgDcIAbAHgBUoIAPYCbAEzihxipAHxQAvFDadufAQAMkAEgDeQkngC+8gFBQoAfijEArhD1QAXFABmSBnAgBuHTpsnsqYJgD22WwBMTPjsXDz8gsJ4IgAUHNbwyGhYeIQa5JQ09Mz4IgCUUCoWDBDAxtam0JIcTlAA9HVQAKJERD5ExaW21nYOVVA19Y0tbR2arqCQsIgoGKRppCwA7j5ZjHCsobIR6qISUlvhCspqUbjaFkbSYXJQ8qq7WgBeuvr6RpUW+j32jhY-fRcbg8Xl8-hsAGZgodbo9cLF4tMknNUmdlqs6OtgvlCp0ysAKkQzPtBg1mq12njurZfv1SY0yABrKAAIxMZTg6B8JgYvFZ0CQ-ggFLGOiAA

### 💻 Code

```ts
type IsMatchingStringOneLevel<T extends string> = T extends `a${string}`
  ? true
  : false;

function f2<S extends string>(x: IsMatchingStringOneLevel<S>) {
  let t: true = x; // Error
  let f: false = x; // Error
}

type IsMatchingStringTwoLevels<T extends string> = T extends `a${string}`
  ? T extends `${string}z`
    ? true
    : false
  : false;

function f3<S extends string>(x: IsMatchingStringTwoLevels<S>) {
  let t: true = x; // Error
  let f: false = x; // Ok but should be an error
}
```

### 🙁 Actual behavior

It doesn't error on the fourth assignment

### 🙂 Expected behavior

It should error

### Additional information about the issue

This is just a variant of what was fixed for one-level conditionals in https://github.com/microsoft/TypeScript/pull/56004

PR #59886, opened 2024-09-07
Improve constraints of nested conditional types applied to constrained type variables

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/59868

this is an extension of https://github.com/microsoft/TypeScript/pull/56004


src/compiler/checker.ts
```diff
@@ -19327,8 +19332,8 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     // there are possible values of the check type that are also possible values of the extends type.
                     // We use a reverse assignability check as it is less expensive than the comparable relationship
                     // and avoids false positives of a non-empty intersection check.
-                    if (checkType.flags & TypeFlags.Any || forConstraint && !(inferredExtendsType.flags & TypeFlags.Never) && someType(getPermissiveInstantiation(inferredExtendsType), t => isTypeAssignableTo(t, getPermissiveInstantiation(checkType)))) {
-                        (extraTypes || (extraTypes = [])).push(instantiateType(getTypeFromTypeNode(root.node.trueType), combinedMapper || mapper));
+                    if (checkType.flags & TypeFlags.Any || distributiveConditionalConstraintIdentityStack.length !== 0 && !(inferredExtendsType.flags & TypeFlags.Never) && someType(getPermissiveInstantiation(inferredExtendsType), t => isTypeAssignableTo(t, getPermissiveInstantiation(checkType)))) {
+                        (extraTypes ??= []).push(instantiateType(getTypeFromTypeNode(root.node.trueType), combinedMapper || mapper));
```
Andarist, line 19336: this is the core of the fix, previously only the first level of nested conditional type (in non tail-recursive scenarios) was including the instantiated true type in its `extraTypes` (the reasoning behind that is outlined in the comment above and in the https://github.com/microsoft/TypeScript/pull/56004 's description)

So now we are including it at all nested levels. In other words - previously the `forConstraint` was only passed to the direct call of `getConditionalTypeInstantiation` from `getConstraintOfDistributiveConditionalType` but now this contextual~ information is used for all nested calls.

I've used a global variable to track this as it's not really feasible today to thread this through `instantiateType` here.

tests/cases/compiler/distributiveConditionalTypeConstraints.ts
```diff
@@ -90,3 +90,61 @@ function test4<T extends A>(y: T extends C ? string : number) {
...
+
+type IsMatchingStringInfiniteRecursionInFalseType2<T extends string> =
+  T extends `a${string}`
+    ? 1
+    : (2 | IsMatchingStringInfiniteRecursionInFalseType2<T>);
+
+function f9<S extends string>(
+  x: IsMatchingStringInfiniteRecursionInFalseType2<S>,
+) {
+  let t1: 1 = x; // Error
```
Andarist, line 148: The error here changed from "Type instantiation is excessively deep and possibly infinite" to "Type '1 | 2' is not assignable to type '1'". 

It feels fine *here* but perhaps it's not fine in general case and I should only return `neverType` for recursive calls when the compiler is instantiating for the true type. This change comes from the fact that I'm letting the `neverType` to be returned for any recursive call and this one comes from the false type.

gabritto:
@typescript-bot test it

src/compiler/checker.ts
gabritto: Why is it not feasible to thread `forConstraint` through `instantiateType`?

gabritto:
@Andarist seems like a bunch of things broke because some types became `never`. Maybe we're being overeager in detecting circularities?

src/compiler/checker.ts
Andarist: Conditional types could appear just at any deeper level of the instantiated type. It's doable to introduce some kind of `InstantiationFlags` and pass it down to all called functions etc but it felt more complicated to attempt doing it. It would also be fairly easy to forget to pass that down somewhere.

Andarist:
I have a hunch that this might be related to my comment [here](https://github.com/microsoft/TypeScript/pull/59886#discussion_r1748003141). I'll investigate and report back with the findings and new tests.

Issue #59877, opened 2024-09-06
"const" was transformed to "var" when target is "esnext"


Roise-yue:
### 🔎 Search Terms

transform, const, var, target, ECMA, esnext

### 🕗 Version & Regression Information

- This is the behavior in every version I tried from Version 5.4 to Version 5.7.0-dev.20240904

### 💻 Code

```ts
export const cilBlurLinear : string [ ] = [  ,  ] 
const [ ,  ] = cilBlurLinear;
```

### 🙁 Actual behavior

The JS code generated by tsc is as follows:
```js
"use strict";
Object.defineProperty(exports, "__esModule", { value: true });
exports.cilBlurLinear = void 0;
exports.cilBlurLinear = [,];
var ;
```

### 🙂 Expected behavior

“const” should be translated as "const" instead of "var", as the target in my configuration file is "esnext".

### Additional information about the issue

tsconfig.json:
```json
{
  "compilerOptions": {
    "target": "ESNext",
    "module": "esnext",
    "moduleResolution": "Node",
    "esModuleInterop": true,
    "skipLibCheck": true,
    "forceConsistentCasingInFileNames": true,
    "lib": ["es2023","dom"],
    "noEmitOnError":true,
    "force":true,
    "strict":true
  },
  "include": [
    "TScorpus/*"
  ],
  "exclude": [
    "node_modules"
  ]
}

```

MartinJohns:
You are using the compiler wrong. You're calling the compiler by passing a list of filenames, in which case your local `tsconfig.json` is not being used and the compiler uses the default values for the compiler options.

If you call the compiler correctly, your code would not even compiler due to two errors:

- `force` is not a valid value for your `tsconfig.json`, it can only be used on the CLI together with `--build`.
- `const cilBlurLinear : string [ ] = [  ,  ]` is not valid with `strictNullChecks` enabled, because `[,]` contains `undefined`.

Roise-yue:
> You are using the compiler wrong. You're calling the compiler by passing a list of filenames, in which case your local `tsconfig.json` is not being used and the compiler uses the default values for the compiler options.
> 
> If you call the compiler correctly, your code would not even compiler due to two errors:
> 
> * `force` is not a valid value for your `tsconfig.json`, it can only be used on the CLI together with `--build`.
> * `const cilBlurLinear : string [ ] = [  ,  ]` is not valid with `strictNullChecks` enabled, because `[,]` contains `undefined`.

[...]

PR #59893, opened 2024-09-08
Drop variable statements when the declaration list gets transformed to an empty list

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/59877


tests/baselines/reference/emptyArrayBindingPatternParameter05.js
```diff
@@ -0,0 +1,26 @@
...
+
+
+//// [emptyArrayBindingPatternParameter05.js]
+"use strict";
+Object.defineProperty(exports, "__esModule", { value: true });
+exports.cilBlurLinear = void 0;
+exports.cilBlurLinear = [[]];
+var _a = exports.cilBlurLinear[0];
+var _b = exports.cilBlurLinear[0];
+var _c = exports.cilBlurLinear[0];
```
Andarist, line 26: we can see here that `const [[]] = ...` isn't as "optimized" as `const [,] = ...`. I think it's fine-ish. All this code assumes `downlevelIteration: false` so it's OK to drop those element access expressions but it's also kinda OK to keep them. It would be cool to use a consistent logic here but that's a separate issue in my eyes

tests/baselines/reference/NonInitializedExportInInternalModule.js
```diff
@@ -41,7 +41,6 @@ var Inner;
 (function (Inner) {
     var ;
     let;
-    var ;
```
Andarist, line 44: All those inputs are incorrect in the first place. So I'm not quite bothered that they produce different kind of an output output. For an invalid input it's OK to emit an invalid output. It just happened that fixing a case for a valid input changed this here - but I think that's fine too. In a similar vein, I think it's OK to drop invalid statements from the invalid input.

tests/cases/conformance/es6/destructuring/emptyArrayBindingPatternParameter05.ts
DanielRosenwasser, line 1: nit on the file name - there are no parameters in this file

tests/baselines/reference/emptyArrayBindingPatternParameter05.js
```diff
@@ -0,0 +1,26 @@
+//// [tests/cases/conformance/es6/destructuring/emptyArrayBindingPatternParameter05.ts] ////
+
+//// [emptyArrayBindingPatternParameter05.ts]
+export const cilBlurLinear: string[][] = [[]];
```
DanielRosenwasser, line 4: It's a little hard to tell whats happening in the output of this test. Maybe introduce 3 different variables so it's clearer where and how they're being destructured.

tests/baselines/reference/emptyArrayBindingPatternParameter05.js
DanielRosenwasser: I'd prefer we didn't just entirely elide the destructuring given proxies/getters - but I'll defer to others. @RyanCavanaugh @rbuckton

DanielRosenwasser (changes requested):

tests/baselines/reference/downlevelLetConst1.js
```diff
@@ -4,4 +4,3 @@
 const
 
 //// [downlevelLetConst1.js]
-var ;
```
jakebailey, line 7: Spooky given this test is named in such a way that implies that it's testing exactly this transformation?

tests/cases/conformance/es6/destructuring/emptyArrayBindingPatternParameter05.ts
Andarist: ah, good catch - that was a bad reuse of one of the other test names 😅

tests/baselines/reference/emptyArrayBindingPatternParameter05.js
Andarist: done

tests/baselines/reference/downlevelLetConst1.js
Andarist: see https://github.com/microsoft/TypeScript/pull/59893/files#r1749158145 , i think the same principle would apply here

this particular test case is testing `const`, the next one in line is testing `const a` and the next one is `const a = 1`. So they just exercise different kinds of inputs and now the question is if it's ok to drop the variable declaration that was invalid in the first place

Issue #56391, opened 2023-11-14
Misleading EPC error on deferred generic type when assigning any object literal, but empty object is also not assignable


fatcerberus:
### 🔎 Search Terms

N/A (bug report forked from #56388)

### 🕗 Version & Regression Information

- This is the behavior in every version I tried

### ⏯ Playground Link

https://www.typescriptlang.org/play?#code/C4TwDgpgBAylC8UDeAPAXFA5AQwEYGNMBfAKADMBXAO32AEsB7KqMgHgBUoIVgIqATAM6wAfAAoAlMhJQo+JoOBR0yANoBrKHWbtVmFJgC6hjFQoBbXBABORBMjz4MARjsB6N1ACiABQDCXNbWDNYANCwhUNhUIFrm2ADm2ngANtDqECAycgpKIBhIGlo6egbGphZWtvZI7p7yQRC0UKCQgcHWJKRAA

### 💻 Code

```ts
type S = {x: 'abc'}
function f<T extends S>() {
  const x: {[k in T['x']]: number} = {abc: 1} // EPC error, for any imaginable key
  const y: {[k in T['x']]: number} = {} // ...but this is also a (correct!) type error
}
```

### 🙁 Actual behavior

EPC diagnostic on any non-empty object literal, but `{}` is also not assignable.

### 🙂 Expected behavior

No EPC diagnostic; should just be a straight assignability failure.

### Additional information about the issue

Producing an EPC (excess property check) error for any non-empty object literal here is misleading since that ultimately implies `{}` would be a legal value (it is not).  5.3 exacerbates this because it removes the "X is not assignable to Y" text from EPC diagnostics, which further implies that the value would be legal if assigned indirectly (also not true; the type is generic and might have other properties).

fatcerberus:
btw w.r.t. the issue template: Can there be a Summary field above Search Terms, so people stop trying to put the bug description into Search Terms? (tbh I almost made the same mistake--it's a bit confusing when you click the search terms box and it expands to a full-featured multiline field, especially since there's no designated place for a summary)

PR #59901, opened 2024-09-09
Exempt generic mapped type targets from being EPC check targets

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/56391

### DIFF
--- a/tests/baselines/reference/genericMappedTypeAsClause.errors.txt
+++ b/tests/baselines/reference/genericMappedTypeAsClause.errors.txt
@@ -3,7 +3,7 @@ genericMappedTypeAsClause.ts(14,11): error TS2322: Type 'number' is not assignab
 genericMappedTypeAsClause.ts(15,11): error TS2322: Type 'string' is not assignable to type 'MappedModel<T>'.
 genericMappedTypeAsClause.ts(16,11): error TS2322: Type 'number[]' is not assignable to type 'MappedModel<T>'.
 genericMappedTypeAsClause.ts(17,11): error TS2322: Type 'boolean' is not assignable to type 'MappedModel<T>'.
-genericMappedTypeAsClause.ts(18,34): error TS2353: Object literal may only specify known properties, and 'a' does not exist in type 'MappedModel<T>'.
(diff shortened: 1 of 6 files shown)


tests/baselines/reference/genericMappedTypeAsClause.errors.txt
```diff
@@ -38,8 +38,8 @@ genericMappedTypeAsClause.ts(25,60): error TS2312: An interface can only extend
               ~~
 !!! error TS2322: Type 'boolean' is not assignable to type 'MappedModel<T>'.
         const x5: MappedModel<T> = { a: 'bar', b: 42 };  // Error
-                                     ~
-!!! error TS2353: Object literal may only specify known properties, and 'a' does not exist in type 'MappedModel<T>'.
```
Andarist, line 42: this was already confusing, `{ a: string, b: number }` is a valid instantiation of `MappedModel<T>`:
```ts
type Model = {
  a: string;
  b: number;
};

type MappedModel<Suffix extends string> = {
  [K in keyof Model as `${K}${Suffix}`]: Model[K];
};

type Compute<T> = { [K in keyof T]: T[K] } & unknown;
declare function f1<T extends string>(): Compute<MappedModel<T>>;

const result = f1<"">();
//    ^? const result: { a: string; b: number; }
```

tests/baselines/reference/noExcessPropertyCheckWithGenericMappedType1.errors.txt
```diff
@@ -0,0 +1,58 @@
...
+    
+      const x: Target = { abc: 1 };
+      const y: Target = {};
+            ~
+!!! error TS2322: Type '{}' is not assignable to type 'Target'.
+      const z: Target = { abc: "" };
+            ~
+!!! error TS2322: Type '{ abc: string; }' is not assignable to type 'Target'.
+    
+      const a: Target = { abc: 1, other: 2 };
```
Andarist, line 28: The `getConstraintOfType(getConstraintTypeFromMappedType(type))` is a string literal. So this one (just like `c` here) could, technically, report an EPC. It feels to me that implementing it might not be worth it, but I can certainly do it if requested.

PR #59907, opened 2024-09-09
Update deps, TS 5.6, remove node-fetch

jakebailey:
This updates our deps, including pulling in 5.6.2. I've also removed `node-fetch` as we only ever run the script in versions of node that have it built in already (so we don't need to keep compat with Node 14).


sheetalkamat (approved):

Machine391:
Continue script main Completed

Machine391 (commented):
jakebailey:dep-redo

Issue #59902, opened 2024-09-09
Removing optional modifier in homomorphic mapped types does not work in generic contexts since 5.5.x


Fullfungo:
### 🔎 Search Terms

optional modifier, required fields, generic, NonNullable, strictNullChecks, homomorphic mapped types

### 🕗 Version & Regression Information

- This changed between versions 5.4.5 and 5.5.2
- This changed in commit or PR e418f8d12c5f1b6c10fc3127764f34dad44d4586 (as reported by `every-ts`)

### ⏯ Playground Link

https://www.typescriptlang.org/play/?ts=5.5.4#code/MYewdgzgLgBApgDwIYFsAOAbOMC8MA8AKvAlHGACYQzQBOAlmAOYB8AFAJQBcMSYAnrhYwA3gCgYMAPRSYAJTiYkwRkxgB3elAAWvATTRwVAM3rAaAVwBGAWij9DMEMZgADOqtcwUSANZxqHWx7Q2pQWlojWCQmJEYAOglpWRDsYjwPZgBuMSTUmABlEBQ4AHk0KHpwJAxcURgAbX9BRhhCAF0Afh4ARhgAXxzJPIdsAEEMDAUARwt6SIo6kUbmmFbm50LisoqqsBr2m26YPsHcyXyZuYWANRqLALqJqbhZ+bgKJrh+Teer94o7RySRkMAAwsUlIxqK0AKzxAAs8QQABoYFYLLBTGBsHD4vCEElQJBYAgeP9bvdHngLJQ4NiPucYJEoBZaGAYAgcoMgA

### 💻 Code

```ts
const example = <T extends string>(): any => {
  // Replacing with any specific sub-type of `string` makes the types correct again.
  // type T = string;

  type SomeOptional = { [key in T]?: 1 };
  
  type AllRequired = { [key in keyof SomeOptional]-?: 1 };

  type RequiredValues = AllRequired[keyof AllRequired];

  // Complains in 5.4.x, but fine in 5.5.x
  const x: RequiredValues = undefined

  return x;
};
```

### 🙁 Actual behavior

Using `-?` leaves the fields marked as optional, and taking the values of the resulting type gives a union with `undefined`.

### 🙂 Expected behavior

Using `-?` makes all fields required, and taking the values of the resulting type gives a union of field types.

### Additional information about the issue

Requires a `--strictNullChecks`  flag.

[...]

PR #59921, opened 2024-09-10
Fixed an issue with optionality not being stripped from generic mapped types on indexed access

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/59902

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -19103,9 +19103,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         const mapper = createTypeMapper([getTypeParameterFromMappedType(objectType)], [index]);
         const templateMapper = combineTypeMappers(objectType.mapper, mapper);
         const instantiatedTemplateType = instantiateType(getTemplateTypeFromMappedType(objectType.target as MappedType || objectType), templateMapper);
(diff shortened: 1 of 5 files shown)


src/compiler/checker.ts
```diff
@@ -19103,9 +19103,9 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         const mapper = createTypeMapper([getTypeParameterFromMappedType(objectType)], [index]);
         const templateMapper = combineTypeMappers(objectType.mapper, mapper);
         const instantiatedTemplateType = instantiateType(getTemplateTypeFromMappedType(objectType.target as MappedType || objectType), templateMapper);
-        const isOptional = getMappedTypeOptionality(objectType) > 0 || (isGenericType(objectType) ?
```
Andarist, line 19106: In the given repro the immediate modifiers type has `-?` for which `getMappedTypeOptionality` returns `-1`. So the first condition was not satisfied and the compiler proceeded to compute `getCombinedMappedTypeOptionality`. This time it was called with the modifiers type and not the original object type. Since that was a mapped type too it got its modifiers type etc.

Essentially this was skipping the outermost mapped type modifiers in this scenario - the ones that were supposed strip the optionality.

Issue #59920, opened 2024-09-10
function object parameter destructure field without default value will be ignored


tjx666:
### 🔎 Search Terms

js, function, parameter, destructure, required

### 🕗 Version & Regression Information

- This changed between versions 5.6 and latest nightly
- We use ts 5.5 no this issue, after upgrade this issue occur.

### ⏯ Playground Link

- [ts playground](https://www.typescriptlang.org/dev/bug-workbench/?allowJs=true&ts=5.7.0-dev.20240904#code/PTAEAEEMBtoewO4CkDOAuUAXATgVwKYBQhIEAZgJbT4B2kAtvhrplSgHQBWKx+AHgAc42TKADGcGilEBzfJgCqKfNgAK2OJWoAhAJ5KVdRqAC8oABQBvQqFC5l2I-gA0N0AgqYAFgFF6kKlNQMhhlV1sPbwA5OFZKMUhWSQBleVYaGRQgkOgwwgBfIMt8gEpTAD5Qa1tseVxHKvyAbgKWkjBwLVoGJlB-Chp2TB5CCnohESrQOUUHdU0qfD0DRx7QQrINelAAcnZgFjYdtpmV+a7lhycrN3tDHowd6F0d11KmoA)
- [stackblitz](https://stackblitz.com/edit/vitejs-vite-usm5mx
)
### 💻 Code

```js
// utils.js

export const getUserProfileByUsername = ({
  username,
  withEmail = false,
  withNotificationSettings = false,
} = {}) => {
  return {};
};
```

```ts
// main.ts
import { getUserProfileByUsername } from './utils';

getUserProfileByUsername({
  username: 'ly',
});
```

### 🙁 Actual behavior

```node
src/main.ts:4:3 - error TS2353: Object literal may only specify known properties, and 'username' does not exist in type '{ withEmail?: boolean | undefined; withNotificationSettings?: boolean | undefined; }'.

4   username: 'ly',
    ~~~~~~~~


Found 1 error in src/main.ts:4
```

### 🙂 Expected behavior

no ts validate error

### Additional information about the issue

ts 5.x not this issue, must be introduced by 5.6

SmolPatat:
https://stackblitz.com/edit/vitejs-vite-usm5mx?file=package.json is not a playground link.
Is there a reason you cannot replicate the issue on https://www.typescriptlang.org/play?

Andarist:
[...]

PR #4598, opened 2015-09-01
Improved checking of destructuring with literal initializers

ahejlsberg:
Fixes #2469.

This PR makes the checking of destructuring patterns with an object literal or array literal initializers more flexible.

When an object literal is contextually typed by the implied type of an object binding pattern:
* Properties with default values in the object binding pattern become optional in the object literal.
* Properties in the object binding pattern that have no match in the object literal are required to have a default value in the object binding pattern and are automatically added to the object literal type.
* Properties in the object literal that have no match in the object binding pattern are an error.

When an array literal is contextually typed by the implied type of an array binding pattern:

* Elements in the array binding pattern that have no match in the array literal are required to have a default value in the array binding pattern and are automatically added to the array literal type.

Some examples:

```typescript
// Type of f1 is (arg?: { x?: number, y?: number }) => void
function f1({ x = 0, y = 0 } = {}) { }
f1();
f1({});
f1({ x: 1 });
f1({ y: 1 });
f1({ x: 1, y: 1 });

// Type of f2 is (arg?: (x: number, y?: number) => void
function f2({ x, y = 0 } = { x: 0 }) { }
f2();
f2({});        // Error, x not optional
f2({ x: 1 });
f2({ y: 1 });  // Error, x not optional
f2({ x: 1, y: 1 });

// Type of f3 is (arg?: [number, number]) => void
function f3([x = 0, y = 0] = []) { }
f3();
f3([]);   // Error, [] not assignable to [number, number]
f3([,]);  // Error, [undefined] not assignable to [number, number]
f3([,,]);
f3([1, 2]);
f3([1, 2, 3]);
```

In the `f3` example, note that the initializer for the parameter binding pattern is allowed to omit elements (and, in particular, is allowed to be an empty array literal). This is a common pattern that we want to support. However, we still require tuples to specify all elements in other cases, as is demonstrated by the calls to `f3`.

UPDATE 1: A question remains about how far we go in allowing tuple literals to omit elements. Possible rules are:

* Allow omitted elements when contextually typed by a binding pattern that specifies default values for the elements, e.g. `function foo([x = 0, y = 0] = [])`. Since that pattern is common, this is the minimum bar.
[...]

PR #59924, opened 2024-09-10
Pad object literal types with implicit anys based on binding pattern elements

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/59920 (a regression)


Andarist:
@jakebailey, I have made local WIP improvements for nested patterns. However, there are some pre-existing inconsistencies between array and binding patterns. It would be great if we could establish the desired behavior as that heavily impacts how I should implement this. Let's take a look at this ([TS 5.5 playground](https://www.typescriptlang.org/play/?allowJs=true&ts=5.5.4#code/GYVwdgxgLglg9mABFApgZygeQEYCsAUA3ogIYA0i2FEiAvIgIyIC+dixJAXIgEwsCU7ZgChhqDDgIduAZgEBuUaEiwEydFACCAJ234A2uUrU2DALpt9PM4MIixGnXv0ybioA)):
```ts
function testObj({ a, b, c = 1 } = { a: 2 }) {} // implicit any on `b`, ok

testObj({ a: 3 }); // error, missing `b`

function testArr([a, b, c = 1] = [2]) {} // implicit any on `b`, ok

testArr([3]); // no error
```

As we can see above, the optionality is different for those members with inferred implicit anys

src/compiler/types.ts
```diff
@@ -6010,7 +6010,6 @@ export interface SymbolLinks {
     exportsChecked?: boolean;                   // True if exports of external module have been checked
     typeParametersChecked?: boolean;            // True if type parameters of merged class and interface declarations have been checked.
     isDeclarationWithCollidingName?: boolean;   // True if symbol is block scoped redeclaration
-    bindingElement?: BindingElement;            // Binding element associated with property symbol
```
Andarist, line 6013: This one has been introduced in https://github.com/microsoft/TypeScript/pull/4598 but its usage was removed in https://github.com/microsoft/TypeScript/pull/59183

tests/baselines/reference/api/typescript.d.ts
```diff
@@ -8750,7 +8750,6 @@ declare namespace ts {
     function isJSDocCommentContainingNode(node: Node): boolean;
     function isSetAccessor(node: Node): node is SetAccessorDeclaration;
     function isGetAccessor(node: Node): node is GetAccessorDeclaration;
-    /** True if has initializer node attached to it. */
```
Andarist, line 8753: this one was copied over from an internal function declared above it, I don't think this comment is true as it made much more sense for the other function

Andarist:
I pushed out changes to make those members with inferred implicit anys optional. Today they must be optional because the **initializer's** type is no longer "padded" (since https://github.com/microsoft/TypeScript/pull/59183 ). 

Given this example:
```ts
// @strict: true
function test({ a } = {}) {}
```

We get `{}` as the initializer's type and the initializer's type is checked if it's assignable to the parameter type. So if we infer `{ a: any }` we get 2 errors:
- an implicit any is reported on the `a` binding element
- we get an error that `{}` is missing a required `a` property

This doesn't feel right. So I've decided to make those implicit anys optional. This allows the non-padded initializer's type to stay assignable to the padded parameter type.

Based on my previous comment I concluded that array and binding patterns were inconsistent and that it would be nice ot unify their behaviors. However, in the process of working on this, I discovered https://github.com/microsoft/TypeScript/pull/4598 and the description of this PR indicates that they were not even meant to behave identically. This was in 2015 though, perhaps some adjustments could be made today.

Either way, to fix the reported problem we can:
- make those implicit anys optional
- pad the initializer type based on the padded parameter type (I dont like this at all ;p)
- bring back padding in `checkObjectLiteral` so the initializer's type would have those properties. This would essentially revert a part of https://github.com/microsoft/TypeScript/pull/59183 . Note that in this situation... the initializer's type here would be `{ a: any }` (and it was that in 5.5!), despite not having the `a` property
- adjust how the initializer is checked against the padded parameter type. I'm not sure how exactly this would work since it's just using a regular assignability check today, so this would have to be a more involved change.

jakebailey:
PR was changed, so rerun

@typescript-bot test it

Andarist:
I believe [the latest commit in my other PR](https://github.com/microsoft/TypeScript/pull/59975/commits/909bac282e05fafa4b0961f4465ee08b642c2450) addresses the issues above. So that other one would have to land before this one here

src/compiler/checker.ts
```diff
@@ -32208,11 +32209,35 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
+        }
+        return undefined;
+    }
+
     function pushCachedContextualType(node: Expression) {
         pushContextualType(node, getContextualType(node, /*contextFlags*/ undefined), /*isCache*/ true);
     }
 
-    function pushContextualType(node: Expression, type: Type | undefined, isCache: boolean) {
+    function pushContextualType(node: Node, type: Type | undefined, isCache: boolean) {
```
jakebailey, line 32240: Can this be a narrower type than just Node?

jakebailey:
> As we can see above, the optionality is different for those members with inferred implicit anys

It sure seems weird for that tuple case to pass, other than that maybe the type is inferred as `number[]`?

@typescript-bot pack this

src/compiler/checker.ts
Andarist: With this implementation, it could be `Expression | BindingPattern | BindingElement` if you'd prefer that.

2 extra notes that come to my mind now:
- maybe I shouldn't be reusing the same `contextualTypes` stack for this and I should just introduce a new one? It seemed fine here to avoid creating a new but then reusing `getContextualType` wasn't possible so it's slightly weird now that I'm only using part of the existing pattern . Perhaps with a new context flag it would be possible but introducing `getContextualPaddingType` seemed easier
- if this gets approved, it might be a good idea to explore expanding `getContextualPaddingType` with support for array patterns and using `pushContextualType`/`popContextualType` in `padTupleType`. Since this PR tries to address a regression in a backportable manner I refrained from doing this to avoid any new unforeseen behavioral changes

Andarist:
> It sure seems weird for that tuple case to pass, other than that maybe the type is inferred as number[]?

I'm not sure which one is `number[]` there, this is the inferred signature I see there:
```ts
function testArr([a, b, c]?: [number, any?, (number | undefined)?]): void
```

jakebailey:
Oh, I guess it could be undefined, yeah.

Andarist:
Closing as per https://github.com/microsoft/TypeScript/issues/59920#issuecomment-2379478733

PR #59927, opened 2024-09-10
Improve generic type signature of Iterator.from()

### LINKED ISSUE #59926: Improve Iterator Helper Type Signatures
### ⚙ Compilation target

ESNext

### ⚙ Library

esnext.iterator.d.ts

Iterator.from

### Sample Code

```TypeScript
https://github.com/nikolaybotev/iteratorhelpersdemo
```

### Documentation Link

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Iterator

and

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols

nikolaybotev:
Fixes #59926

## Background

`Iterator.from` returns an iterator that passes along the return value of its wrapped iterator, but does NOT pass along the argument to its `next` method to the wrapped iterator. This behavior is observed in node.js 22 using the following scripts and I assume is the standard behavior:

* https://github.com/nikolaybotev/iteratorhelpersdemo/blob/main/helper-propagates-next.ts
* https://github.com/nikolaybotev/iteratorhelpersdemo/blob/main/helper-propagates-return.ts

## Problem

The current type signature for `Iterator.from` basically ignores the `TReturn` and `TNext` type arguments of the underlying `Iterator` protocol and make very hard assumptions about their values, which restrict certain valid code from passing typescript compilation, such as these two examples:

* https://github.com/nikolaybotev/iteratorhelpersdemo/blob/main/problems/from/helper-propagates-next.ts
* https://github.com/nikolaybotev/iteratorhelpersdemo/blob/main/problems/from/helper-propagates-return.ts

## Solution

Add the `TReturn` type argument to `Iterator.from` and reflect the observed behavior in the type signature of the argument and return value.

## Tests

Tested manually using the scripts linked above under the *Problem* section.

### DIFF
--- a/src/lib/esnext.iterator.d.ts
+++ b/src/lib/esnext.iterator.d.ts
@@ -123,7 +123,7 @@ declare global {
          * Returns its input if the input already inherits from the built-in Iterator class.
          * @param value An iterator or iterable object to convert a native iterator.
          */
-        from<T>(value: Iterator<T, unknown, undefined> | Iterable<T, unknown, undefined>): IteratorObject<T, undefined, unknown>;
+        from<T, TReturn>(value: Iterator<T, TReturn, any> | Iterable<T, TReturn, any>): IteratorObject<T, TReturn, unknown>;
     }
 
     var Iterator: IteratorConstructor;


nikolaybotev:

nikolaybotev:
Done. I have updated the baselines.

nikolaybotev:
```
```

nikolaybotev:
Tests added.

nikolaybotev:
This PR is now ready for review.

nikolaybotev:
I am no realizing that the passing along of the wrapped iterator's return value by the object returned by `Iterator.from` is very likely an optimization side-effect and not a behavior by-design.

NOT passing along the next() argument is achieved by removing one argument to one function call.

NOT passing along the return() value requires checking and unwrapping the IteratorResult object if done... adding unnecessary overhead.

Iterator Helpers really operate on one-way streams and the one-way stream aspect of Iterators and Generators is really best captured by their `Iterator<T, undefined, undefined>` specialization,  and as such this is what the baseline `Iterator.from` signature reflects - the design intent, not any underlying implementation side effects, which can (and will be 99.99% of the time) be ignored by users.

The `Iterator<T, unknown, undefined>` and `Iterator<T, undefined, unknown>` are variations that make the protocol more permissive for the user: the first one is used in the argument position to tell the client - give me an Iterator that takes no arguments in its `next` method (because I am not going to send any!) and can return whatever `IteratorReturnResult` it wants because I am going to ignore it anyway. Conversely, the other variant is used in the return position to say - I am giving you an Iterator object, that you can pass anything you want to its `next` function as I am going to ignore it, and you are never going to get any value in `IteratorReturnResult` back (well you might, but that's an implementation optimization side-effect and you can ignore it and should not expect it or rely on it).

Can someone confirm my reading is correct?

mattmcnally:
@nikolaybotev Your original analysis appears to be correct - both `next` and `return` results should be directly forwarded from the wrapped iterator: see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Iterator/from#description.

For added detail, see the ES specs for [next](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-%wrapforvaliditeratorprototype%.next) and [return](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-%wrapforvaliditeratorprototype%.return). Also here is the V8 implementation of `return` (https://github.com/v8/v8/blob/6f1700ffbce61d4a3ff397b388e5e8446128a5e2/src/builtins/iterator-from.tq#L108C34-L108C69) which just forwards the result as you said (as per the spec).

Based on this, the types for `Iterator.from` appear incorrect - `TReturn` should be included.

PR #59932, opened 2024-09-10
Sanitize version string for syntactically invalid buildinfo baselines

weswigham:
Nightlies haven't been publishing for [a few days](https://github.com/microsoft/TypeScript/actions/workflows/nightly.yaml) because some new baselines inadvertently captured the current ts version in them. The fancy automatic `.buildinfo` version stripping logic done by the builder baselines only applies to *syntactically valid* `.buildinfo` files, so these tests with _slightly invalid_ ones were inadvertently capturing the current version in the test baseline. Do note, with how the version removal machinery works, this extra ad-hoc sanitization is edit-order-dependent - if you try and make this edit *before* making the file syntactically invalid, the version-stripping-aware-sys will just add the current version right back to the file when it's re-read for the 2nd edit. 😅

### DIFF
--- a/src/testRunner/unittests/tsbuild/sample.ts
+++ b/src/testRunner/unittests/tsbuild/sample.ts
@@ -1,7 +1,7 @@
 import { Baseline } from "../../_namespaces/Harness.js";
 import * as ts from "../../_namespaces/ts.js";
 import { jsonToReadableText } from "../helpers.js";
-import { patchHostForBuildInfoReadWrite } from "../helpers/baseline.js";
+import { fakeTsVersion, patchHostForBuildInfoReadWrite } from "../helpers/baseline.js";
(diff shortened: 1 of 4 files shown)


jakebailey:
This applies the fix in the couple of tests; there's no way to do it more generally for syntactically invalid cases?

sheetalkamat (approved):

jakebailey (approved):
In any case, worth unblocking nightly

weswigham:
> This applies the fix in the couple of tests; there's no way to do it more generally for syntactically invalid cases?

Probably not - regex searching for "probable version strings" or the like in test baseline data is probably best avoided, since it could result in some unintuitive failures; I think we can probably handle sanitizing it manually in the tests where we know the automatic system is going to fail. Probably.

PR #28403, opened 2018-11-07
Convert node and symbol links into checker-local sparse arrays with per-property getters and setters

weswigham:
Was talking about this with @ahejlsberg the other day. Still to need rerun perf tests to verify that this still gets us some perf gains (I put up #19163 a long time ago, but we hesitated on it because the API was eh, despite the perf gains); but this API (thanks for the suggestions @ahejlsberg) is actually nicer than our older symbol links API (in addition to being monomorphic), since adding new ones is easy and doesn't require synchronizing types. I'll have some perf testing results up soon.

As an aside, while refactoring I discovered we actually had two _unused_ `NodeLinks` members - `resolvedIndexInfo` and `resolvedJsxElementAllAttributesType`. They are simply gone, now.

PR #59933, opened 2024-09-10
add jsx fragments to callLikeExpression

### LINKED ISSUE #50429: JSX Fragments are in-properly typed when not using `React.Fragment`
# Bug Report

### 🔎 Search Terms

React Fragment

I believe this is a new inconsistency due to changes in React.

### ⏯ Playground Link

https://github.com/gkemp94/test-typescript-fragment

### Code
```
import React from "react";

const test = () => "asd";

// No Errors
const jsxWithJsxFragment = <>{test}</>;

// Type '() => string' is not assignable to type 'ReactNode'.
const jsxWithReactFragment = <React.Fragment>{test}</React.Fragment>;

```

### 🙁 Actual behavior

When using `<>` vs `<React.Fragment>` the children are not properly type checked

### 🙂 Expected behavior

When jsx: react is specified, to type check `<>` the same as `React.Fragment`

iisaduan:
This PR adds type checking to the children of JSX fragments. Fixes #50429.

- Adds `JsxOpeningFragment` to `CallLikeExpression` by defining `JsxCallLike`.
  - `JsxOpeningFragment` is not added to `JsxOpenlingLikeFragment` because it requires many more changes than necessary for this typecheck, and causes crashes in expression checking. More details [here](https://github.com/microsoft/TypeScript/pull/59933/#discussion_r1777695626).
- Updates to `NodeLinks`:
  - Adds `jsxFragmentType` to store the type of the fragment at the `SourceFile` level, since a different fragment factory can be referenced per file.
  - Removes `resolvedJsxElementAllAttributesType` and `resolvedIndexInfo` (since both have been unused [since at least 2018??](https://github.com/microsoft/TypeScript/pull/28403#issue-378512706))
- Fixes `isCallLikeExpression` to account for `InstanceOfExpression`
- New error `TS2876: Using JSX fragments requires fragment factory {0} to be in scope, but it could not be found.` We previously did not issue an error if a fragment factory was set through a pragma or compiler option, and the specified factory could not be resolved.


iisaduan:
@typescript-bot test it

src/compiler/types.ts
```diff
@@ -3187,6 +3187,10 @@ export type JsxOpeningLikeElement =
     | JsxSelfClosingElement
     | JsxOpeningElement;
 
+export type JsxCallLike =
+    | JsxOpeningLikeElement
+    | JsxOpeningFragment;
```
Andarist, line 3192: wouldnt it make sense to include `JsxOpeningFragment` as part of `JsxOpeningLikeElement`?

src/compiler/utilitiesPublic.ts
Andarist, line 1: This file has also `isCallLikeExpression` utility. Since its return type is `node is CallLikeExpression` and `JsxCallLike` is now a `CallLikeExpression`, its implementation should be adjusted. Note that this function wasn't completely correct even before this change... it doesn't handle correctly the `InstanceofExpression`/`BinaryExpression` case.

iisaduan: Good catch, thanks!

src/compiler/types.ts
DanielRosenwasser: @iisaduan can you address that suggestion?

src/compiler/checker.ts
```diff
@@ -35500,10 +35525,12 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             }
 
             if (reportErrors) {
-                const diag = createDiagnosticForNode(node.tagName, Diagnostics.Tag_0_expects_at_least_1_arguments_but_the_JSX_factory_2_provides_at_most_3, entityNameToString(node.tagName), absoluteMinArgCount, entityNameToString(factory), maxParamCount);
-                const tagNameDeclaration = getSymbolAtLocation(node.tagName)?.valueDeclaration;
+                const errorNode = isJsxOpeningFragment(node) ? node : node.tagName;
+                const errorEntityName = isJsxOpeningFragment(node) ? "Fragment" : entityNameToString(node.tagName);
+                const diag = createDiagnosticForNode(errorNode, Diagnostics.Tag_0_expects_at_least_1_arguments_but_the_JSX_factory_2_provides_at_most_3, errorEntityName, absoluteMinArgCount, entityNameToString(factory), maxParamCount);
+                const tagNameDeclaration = isJsxOpeningFragment(node) ? getJSXFragmentType(node).symbol.valueDeclaration : getSymbolAtLocation(node.tagName)?.valueDeclaration;
```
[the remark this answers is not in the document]
iisaduan: I did some testing, and even with proper checking of fragments in the rest of `checkTagNameDoesNotExpectTooManyArguments`, it doesn't seem possible to hit this error message with a fragment, so I reverted this change. Hitting this error message requires attributes to be passed in through the tag (example `<TagName x={/*stuff*/}>`). For just an opening element, if there are no attributes in the tag (`<TagName>`), we won't ever hit this error because we issue missing property or not assignable errors instead. So even if `Fragment` is defined to always take multiple arguments, we won't ever get this error

src/compiler/types.ts
iisaduan: Apologies for the delay in response. It's a good question and I wanted something detailed for PR description. 

TLDR keeping a type that has only `opening element` and `self closing element`, is still beneficial, and adding `JsxOpeningFragment` to `JsxOpeningLikeElement` requires a much bigger change than is needed to add this fragment type check. The biggest reason is because this add causes crashes in expression checking. We would either need to rework what fragments are, in a way that is making them almost identical to an `JsxOpeningElement`, or we would need to update the majority of places `JsxOpeningLikeElement` is referenced to ignore fragments anyway (and if we don't want to ignore them, we'd need to handle fragments specially, which could cause us to do redundant/unnecessary work). 

In my first iteration of this PR, I did try to add fragments to `openingLike`, and the biggest issue was probably that opening fragments are not `Expression`s, and we run into crashes during expression checking. Surprisingly, `CallLikeExpression`s are not actually used in many places that require `Expression`s, so we don't run into this issue when only adding fragments to `CallLikeExpression`. Since now the fragment check fully added, I tried again to fix these crashes over the last day or so, but I still ran into layers of them, so it just ends up being much more work than is necessary for this type check.

Also regarding the use of `JsxOpeningLikeElement`-- they have many more checks than what we need to check fragments, so adding fragments would require us to then ignore fragments in many uses of openinglike. For example, after this PR, just in `checker`, there are ~30 functions that would still error because of fragment incompatibilty with what `JsxOpeningLikeElement` is currently. These functions would almost always be updated to ignore fragments anyways, since ~20 are related to checking `tagName` or `attributes`. While the rest may have checks applicable to fragments, they aren't performing any different checks than what we already do (so we should probably ignore fragments because we don't need to do the work for them again). This is without taking into account other parts of the compiler/services, and anyone else using `JsxOpeningLikeElement`

weswigham (approved):

PR #59940, opened 2024-09-11
Proposed expandable hover API

gabritto:
VSCode counterpart: https://github.com/microsoft/vscode/pull/228255

This is an incomplete prototype of expandable hovers using the [current vscode proposed expandable hover API](https://github.com/microsoft/vscode/issues/210904).

Mainly the only scenario implemented so far is for hovering over **variables**, e.g.:

```ts
type A = number | string;
type B = symbol | boolean;
type SomeType<T, U> = {
    propA: T,
    propB: U,
};

let a: A; // Hover on `a` will be expandable
let b: B; // Hover on `b` will be expandable
let someVar: SomeType<A, B>; // Hover on `someVar` will be expandable
```


gabritto:
@typescript-bot test it

gabritto:

RyanCavanaugh (approved):

jasonkuhrt:
Hey @gabritto I'm interested in this for two reasons as a library author:

1. Manual efforts I make trying to apply `Simplify` etc. utility types to my complex type interfaces for a readable final type experience for users.
2. The above exacerbating cases where I'm hitting "instantiation depth" TS limit.

Other than reading the code/try out the feature is there a complete explanation I could review to understand what the new behaviour will be? If it is just code and "try it" that's understandable, but just checking.

Also, IIUC, the type display changes depend on the user being a VSCode user? Or, could it be any editor that uses the TS LSP?

gabritto:
The feature isn't finished as of yet, but we're hoping to soon put out an initial version of it. The basic idea is that users would be able to expand levels on their types when they hover over it in the editor, and as they expand a type, we replace type aliases with their definitions, e.g. going from `Person` to `{ name: string, address: Address }` to `{ name: string, address: { street: string, number: number }}`.

The feature will be available via our TSServer API. Any editor using that API will have access to it. Note though that TSServer API is not LSP compatible.

jasonkuhrt:
@gabritto thanks for the info I see it's not quite what I thought, but still very cool!

Issue #59948, opened 2024-09-12
Inconsistent behaviour of `-?`


m-shaka:
### 🔎 Search Terms

optional mapped type

### 🕗 Version & Regression Information

- This changed between versions 5.4.5 and 5.5.4

### ⏯ Playground Link

https://www.typescriptlang.org/play/?exactOptionalPropertyTypes=true#code/C4TwDgpgBA8mwEsD2A7AhgGwCpIKouRQHUFgALfAEwgDMEUJKAeLAPigF4oBvAKCgFQA2gGko9KAGsIIJDShYAugC4FUCAA9gEFJQDOUAEoQAxkgBOzEQBoFoxa36DnAfjsjFT5wNVZ7UAB8oAFddWnpGXgBfXl5QSCgASRRtcwBbBEoENG1OWHhCTBx8QhJyKnCGZm4oNBdVPWBzegBzKCj2AHpOnlr6qEbmlDag0Oo6Kva48GgAGVIIc0w8mrqGptbAkLCJxnaoboG0NOh46DQDZNSMrJyIWLOjCD0ARjzjAEdghHNGJivFjdstouj1Vushm0Yo9jHoAEzvCBfH5-eapTCg3poCGbUY7CKUdpAA

### 💻 Code

```ts
type OptionalToUnionWithUndefined<T> = {
    [K in keyof T]: T extends Record<K, T[K]>
        ? T[K]
        : T[K] | undefined
}

type Intermidiate = OptionalToUnionWithUndefined<{ a?: string }> // { a?: string | undefined }
type Literal = { a?: string | undefined } // same type as Intermidiate

type Res1 = Required<Intermidiate> // { a: string }
type Res2 = Required<Literal> // { a: string | undefined }
```

### 🙁 Actual behavior

`Res1` and `Res2` are different types

### 🙂 Expected behavior

they should be the same because the inputs are

### Additional information about the issue

This issue may share the cause with #59902

SmolPatat:
This seems to be the expected behaviour from the `--exactOptionalPropertyTypes` flag.

SmolPatat:
I think TS is tripping balls and hallucinating here. Both `Res1` and `Res2` are supposed to be `{a: string | undefined}`, and they **are**!
The following code compiles:
```ts
type Res1a = Res1['a']; // string | undefined
let r: Res1 = {a: undefined}; // Ok
```

RyanCavanaugh:
[...]

PR #59957, opened 2024-09-13
Fixed an issue with mapped property symbol not displaying added `| undefined` when its origin symbol was optional

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/59948

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -8378,8 +8378,8 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 if (declWithExistingAnnotation && !isFunctionLikeDeclaration(declWithExistingAnnotation) && !isGetAccessorDeclaration(declWithExistingAnnotation)) {
                     // try to reuse the existing annotation
                     const existing = getNonlocalEffectiveTypeAnnotationNode(declWithExistingAnnotation)!;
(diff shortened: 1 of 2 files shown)


mkantor:
Over in #59948 I shared [a reduced reproduction](https://www.typescriptlang.org/play/?ts=5.5.4#code/C4TwDgpgBAglC8UDeUDaBpKBLAdlA1hCAPYBmyUAhgPwBcUAzsAE64DmUAvgLoC0djFuy4AoAPRioUgHrURoSFABCCChmx5CJcjG70mrHBwA+UAK44AJhFK4Il0RKlRZIkQGNiOJlABG9FUQUSnoLa1scey4oJxxiKAhmZmJmABooawYwLGBoBWhcUnjKX2IANwggA) of this issue which doesn't require `exactOptionalPropertyTypes`. Could be nice to include as a regression test here too? I tested locally and your fix does seem to address it.

```ts
// @strict: true

//// type A/*1*/ = { [K in keyof { a?: string }]-?: string }
//// type B/*2*/ = { [K in keyof A]: string | undefined }

verify.quickInfoAt("1", `type A = {
    a: string;
}`);
verify.quickInfoAt("2", `type B = {
    a: string | undefined;
}`);
```

(I could open a pull request targeting your branch if you'd like; didn't want to cause confusion.)

PR #59963, opened 2024-09-13
Add CopilotRelated command

sandersn:
Add a command to provide information about a file for copilot.

Associated VSCode PR: https://github.com/microsoft/vscode/pull/228610

- I'm not sure what to name this yet. I expect this will be a catch-all command to provide information about a file.
- I'm also not 100% sure this is going to get all the imports. @andrewbranch I expect you are the best person to ask about that.

@genlu you were interested in seeing this.


src/services/services.ts
```diff
@@ -3345,6 +3345,12 @@ export function createLanguageService(
         );
     }
 
+    function getImports(fileName: string): readonly string[] {
+        synchronizeHostData();
```
sheetalkamat, line 3349: If you are using synchronize host data this is not syntax operation. Need to add this to `invalidOperationsInPartialSemanticMode`

Need to add protocol command to `invalidPartialSemanticModeCommands`

src/server/session.ts
```diff
@@ -2034,6 +2033,14 @@ export class Session<TMessage = string> implements EventSender {
         return this.mapTextChangesToCodeEdits(changes);
     }
 
+    private getCopilotRelatedInfo(args: protocol.FileRequestArgs): protocol.CopilotRelatedItems {
+        const { file, languageService } = this.getFileAndLanguageServiceForSyntacticOperation(args);
```
sheetalkamat, line 2037: This doesnt look like syntax operation : need to use `const { file, project } = this.getFileAndProject(args);`

sheetalkamat (changes requested):

src/services/services.ts
sandersn: I don't know if I need to synchronize host data to call getResolvedModuleFromModuleSpecifier. How can I tell?
(I'm guessing the answer is yes, though.)

sheetalkamat: If you need to use more than open file , it needs to synchronize data .. that is anything that uses program needs to synchronize host data

src/services/services.ts
```diff
@@ -3345,6 +3346,12 @@ export function createLanguageService(
         );
     }
 
+    function getImports(fileName: string): readonly string[] {
+        synchronizeHostData();
+        const file = getValidSourceFile(fileName)
+        return mapDefined(file.imports, i => program.getResolvedModuleFromModuleSpecifier(i, file)?.resolvedModule?.resolvedFileName);
```
andrewbranch, line 3352: This will get

- imports written by the user (normal, async, type-only)
- re-exports (`export * from "./foo"`) written by the user
- the implicit tslib and/or JSX factory import added during emit
- import types (`typeof import("./foo")`)
- requires in JS

and it is possible it will contain duplicates. In other words, this is collecting the maximal amount of stuff you could mean by "imports in a file." Not sure what you intend, but I'm guessing you may at least want to deduplicate, and maybe consider dropping the tslib/JSX one?

src/harness/fourslashInterfaceImpl.ts
```diff
@@ -2039,3 +2043,8 @@ export interface RenameOptions {
     readonly providePrefixAndSuffixTextForRename?: boolean;
     readonly quotePreference?: "auto" | "double" | "single";
 }
+
+export interface VerifyGetImportsOptions {
+    filename: string;
```
andrewbranch, line 2048: Is this set in stone on the VS Code side? All other properties in the protocol use `fileName`

andrewbranch (approved):

src/harness/fourslashInterfaceImpl.ts
sandersn: Nope! I just made it up. I'll change it.

src/server/protocol.ts
```diff
@@ -2391,6 +2392,19 @@ export interface MapCodeResponse extends Response {
...
 }
 
+export interface CopilotRelatedRequest extends FileRequest {
+    command: CommandTypes.CopilotRelated;
+    arguments: FileRequestArgs;
+}
+
+export interface CopilotRelatedItems {
+    relatedFiles: readonly string[];
+    traits: { name: string; value: string; }[];
```
sheetalkamat, line 2402: what is this supposed to be?

src/services/services.ts
```diff
@@ -3345,6 +3347,20 @@ export function createLanguageService(
...
 
+    function getImports(fileName: string): readonly string[] {
+        synchronizeHostData();
+        const file = getValidSourceFile(fileName);
+        const imports: Set<string> = new Set();
+        let specifiers = file.imports;
+        specifiers = specifiers.slice(specifiers.findIndex(nodeIsSynthesized) + 1);
+        if (specifiers.length === 0) return [];
+        for (const specifier of specifiers) {
+            const name = program.getResolvedModuleFromModuleSpecifier(specifier, file)?.resolvedModule?.resolvedFileName;
```
sheetalkamat, line 3358: it may happen that we have resolvedFileName = say "module.js" but its not part of program. Is this ok to send back that file name ?

src/services/services.ts
```diff
@@ -3345,6 +3347,20 @@ export function createLanguageService(
...
+            if (name && !imports.has(name)) imports.add(name);
+        }
+        return Array.from(imports);
```
sheetalkamat, line 3361: use `arrayFrom` helper instead ?

src/services/services.ts
```diff
@@ -3345,6 +3347,20 @@ export function createLanguageService(
...
+    function getImports(fileName: string): readonly string[] {
+        synchronizeHostData();
+        const file = getValidSourceFile(fileName);
+        const imports: Set<string> = new Set();
+        let specifiers = file.imports;
+        specifiers = specifiers.slice(specifiers.findIndex(nodeIsSynthesized) + 1);
+        if (specifiers.length === 0) return [];
+        for (const specifier of specifiers) {
+            const name = program.getResolvedModuleFromModuleSpecifier(specifier, file)?.resolvedModule?.resolvedFileName;
+            if (name && !imports.has(name)) imports.add(name);
```
sheetalkamat, line 3359:
```suggestion
if (name) imports.add(name);
```
checking for presence and then adding is not required ?

src/services/services.ts
```diff
@@ -3345,6 +3347,20 @@ export function createLanguageService(
         );
     }
 
+    function getImports(fileName: string): readonly string[] {
+        synchronizeHostData();
+        const file = getValidSourceFile(fileName);
+        const imports: Set<string> = new Set();
+        let specifiers = file.imports;
+        specifiers = specifiers.slice(specifiers.findIndex(nodeIsSynthesized) + 1);
+        if (specifiers.length === 0) return [];
```
sheetalkamat, line 3356: return emptyArray

src/services/services.ts
```diff
@@ -3345,6 +3347,20 @@ export function createLanguageService(
         );
     }
 
+    function getImports(fileName: string): readonly string[] {
+        synchronizeHostData();
+        const file = getValidSourceFile(fileName);
+        const imports: Set<string> = new Set();
+        let specifiers = file.imports;
+        specifiers = specifiers.slice(specifiers.findIndex(nodeIsSynthesized) + 1);
```
sheetalkamat, line 3355: instead of splicing array just skip this when iterating ?

src/services/services.ts
```diff
@@ -3345,6 +3347,20 @@ export function createLanguageService(
         );
     }
 
+    function getImports(fileName: string): readonly string[] {
+        synchronizeHostData();
+        const file = getValidSourceFile(fileName);
+        const imports: Set<string> = new Set();
```
sheetalkamat, line 3353: let imports: Set<string> | undefined and allocated when needed?

sheetalkamat (changes requested):

src/server/protocol.ts
sandersn: Project-related information for copilot like module, lib, target, etc; not implemented in this PR, but coming soon.

src/services/services.ts
sandersn: I thought emptyArray was slower than `[]`. Also, what happens if somebody mutates the array returned from getImports? Is that a concern for services?

src/services/services.ts
sandersn: Why? It looks worse than the built-in, since it's an overloaded function that checks at each iteration whether a mapping function is defined.

src/services/services.ts
sandersn: Efficiency isn't important in this code. It's called once every 3 minutes at most.

src/services/services.ts
sandersn: Yeah, that's simpler.

src/services/services.ts
sandersn: Yes, files that are not part of the program are fine; we don't expect to get information from Typescript about them -- they are fine if they exist on disk.

src/server/protocol.ts
sheetalkamat: But do we know how the structure will look like. May be add this when we actually set?

src/services/services.ts
```diff
@@ -3345,6 +3348,21 @@ export function createLanguageService(
...
+        let imports: Set<string> | undefined;
+        for (const specifier of file.imports) {
+            if (nodeIsSynthesized(specifier)) continue;
+            const name = program.getResolvedModuleFromModuleSpecifier(specifier, file)?.resolvedModule?.resolvedFileName;
+            if (name) {
+                if (!imports) imports = new Set();
+                imports.add(name);
+            }
+        }
+        return imports ? arrayFrom(imports) : [];
```
sheetalkamat, line 3363:
```suggestion
        return imports ? arrayFrom(imports) : emptyArray;
```

src/services/services.ts
```diff
@@ -3345,6 +3348,21 @@ export function createLanguageService(
...
+        synchronizeHostData();
+        const file = getValidSourceFile(fileName);
+        let imports: Set<string> | undefined;
+        for (const specifier of file.imports) {
+            if (nodeIsSynthesized(specifier)) continue;
+            const name = program.getResolvedModuleFromModuleSpecifier(specifier, file)?.resolvedModule?.resolvedFileName;
+            if (name) {
+                if (!imports) imports = new Set();
+                imports.add(name);
+            }
```
sheetalkamat, line 3361:
```suggestion
           if (name) (imports ??= new Set()).add(name);
```

src/server/protocol.ts
sandersn: It's not going to change substantially, although it may add some optional properties -- I believe there's a PR from the C++ team to do so. I'm going to start returning some project information as traits quite soon and C# is already doing so. So I'd like to add it now, and I can add any optional properties that we use then.

sheetalkamat (approved):

Issue #59937, opened 2024-09-10
React's ComponentProps type issues in TypeScript 5.6.2


jakubmazanec:
### 🔎 Search Terms

react, forwardRef, component props, ComponentProps

### 🕗 Version & Regression Information

- This changed between versions 5.5.4 and 5.6.2 - but in version 5.5.4 there are other related bugs 🤷‍♂️

### ⏯ Playground Link

[...]

PR #59972, opened 2024-09-15
Fixed single signature type instantiations leaking between calls with different mappers

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/59937

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -20218,7 +20218,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         const links = getNodeLinks(declaration);
         const target = type.objectFlags & ObjectFlags.Reference ? links.resolvedType! as DeferredTypeReference :
             type.objectFlags & ObjectFlags.Instantiated ? type.target! : type;
-        let typeParameters = type.objectFlags & ObjectFlags.SingleSignatureType ? (type as SingleSignatureType).outerTypeParameters : links.outerTypeParameters;
+        let typeParameters = links.outerTypeParameters;
         if (!typeParameters) {
             // The first time an anonymous type is instantiated we compute and store a list of the type
             // parameters that are in scope (and therefore potentially referenced). For type literals that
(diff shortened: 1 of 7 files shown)


tests/cases/compiler/genericCallInferenceInConditionalTypes1.ts
```diff
@@ -0,0 +1,59 @@
...
+type Test<T> = T extends { component?: infer Component }
+  ? Component extends FunctionComponent<any>
+    ? ComponentProps<Component>
+    : never
+  : never;
+
+// the first one here has a chance to pollute the cache
+type Result1 = ComponentProps<typeof ComponentWithForwardRef>;
+// that could be incorrectly reused by this one
+type Result2 = Test<{ component: typeof ComponentWithForwardRef }>; // no `T` leak
```
Andarist, line 46: in the process of resolving the first conditional type here a result with a conditional type's restrictive mapper was cached, then when resolving the second conditional type that cached result was reused incorrectly, failing to produce a type with its correct mapper applied to it

src/compiler/checker.ts
```diff
@@ -20244,18 +20244,13 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
...
                 target.instantiations = new Map<string, Type>();
                 target.instantiations.set(getTypeListId(typeParameters) + getAliasId(target.aliasSymbol, target.aliasTypeArguments), target);
             }
             let result = target.instantiations.get(id);
             if (!result) {
-                if (type.objectFlags & ObjectFlags.SingleSignatureType) {
-                    result = instantiateAnonymousType(type, mapper);
-                    target.instantiations.set(id, result);
-                    return result;
-                }
```
Andarist, line 20258: It might feel like a downgrade that those are not cached right now. `getObjectTypeInstantiation` (the containing function) is only called from a single place and all the functions called from the branch leading to `getObjectTypeInstantiation` are implementing instantiation caching (`createNormalizedTypeReference`, `instantiateReverseMappedType` and `getObjectTypeInstantiation`).

But now not the branch I added - the one that calls `instantiateAnonymousType` for single signature types. I have no idea what I would use to create the cache key there. The only unique bits there are the results of the `mapper` but `mapper` isn't exactly called eagerly there so it doesn't have access to those.

However, I have a feeling that perhaps caching is already - sort of - implemented for those single signature types. It's just at a different level, like in the `getReturnTypeOfSignature`.

jakebailey:
@typescript-bot test it

src/compiler/checker.ts
jakebailey: My gut is making me worried that we're dropping some perf here due to not caching quickly, but maybe it doesn't matter at all?

Andarist: I was definitely worried about this when proposing this change but it feels that the `getReturnTypeOfSignature` caching should already cover most of this. It would still be best if @weswigham could take a look at this though.

I have just pushed out a new test case to this PR because I narrowed down another issue caused by those cache leaks here.

Andarist:
@weswigham friendly 🏓

Issue #59936, opened 2024-09-10
Binding element incorrectly reported as having `any` type in JS with JSDoc starting with 5.6.2


burtek:
### 🔎 Search Terms

Binding element implicitly has an any type 7031

### 🕗 Version & Regression Information

- This changed between versions 5.5.4 and 5.6.2

### ⏯ Playground Link

https://www.typescriptlang.org/play/?strictNullChecks=true&target=9&suppressImplicitAnyIndexErrors=true&ts=5.6.2&filetype=js#code/PQKhCgAIUgBAHAhgJ0QW0gbwPICMBWApgMYAuAvpANrED2AdgGYCWA5gLpQwIrpYAKKUs0QAbADwAlErWQATcQHJ8AZwaLIAH0jK19Yhu26GAVkUAaSCtLJm9VlXYA+J5RoMWrAHSI5c5sIMYgBizKKEKpzQwOCEAB7wsqSQjACu+oH0kPDIhEi5AMIebAAUmFCQlb7+mSFhEQBcWBWVrZCqDJAAvNRRrZQ9mOTgA1jkAJTNbXT0auFeorSsJR304yNAA

### 💻 Code

```js
/**
 * @param {Object} [config]
 * @param {Partial<Record<'json' | 'jsonc' | 'json5', string[]>>} [config.additionalFiles]
 */
export function prepareConfig({
    additionalFiles: {
        json = []
    } = {}
} = {}) {
    console.log(json)
}
```

### 🙁 Actual behavior

`json` is errored as `Binding element 'json' implicitly has an 'any[]' type.(7031)` even though it's typed as `string[]`

![image](https://github.com/user-attachments/assets/29b07884-02c7-4e10-bf6f-ee9a2b9b1967)

### 🙂 Expected behavior

No error expected, as `json` is typed as `string[]`

### Additional information about the issue

This worked well in 5.5.4 but broke with upgrade to 5.6.2. It works correctly in .ts file, but breaks in .js file with JSDoc.

Andarist:
I'll fix this as part of https://github.com/microsoft/TypeScript/pull/59924

burtek:
@Andarist ok, thanks. I guess it's gonna go out as 5.6 patch bump?

Andarist:
@burtek in the end I decided to create a separate PR for this one since I was able to fix this independently and the other PR raises some design questions. The fix is available here: https://github.com/microsoft/TypeScript/pull/59975

[...]

PR #59975, opened 2024-09-15
Only pad initializer types for parameters without annotations and contextual signatures

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/59936

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -40553,7 +40553,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             checkExpressionCached(initializer, checkMode));
 
         const rootDeclaration = getRootDeclaration(declaration);
-        if (isParameter(rootDeclaration) && !getEffectiveTypeAnnotationNode(rootDeclaration)) {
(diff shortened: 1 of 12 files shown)


jakebailey:
@typescript-bot test it

jakebailey:
@typescript-bot test it

jakebailey:
@typescript-bot pack this

Andarist:
> Maybe I'm misunderstanding it, but isn't this sorta kinda a revert of https://github.com/microsoft/TypeScript/pull/59183 but only in the parameter case? Is this too targeted? Or am I misunderstanding?

It's not a revert. Previously object literal types were contextually-typed by their corresponding patterns. 

So in a situation like this:
```ts
function test({ a = 1 } = {}) {}
```

The type of this `{}` ended up to be `{ a?: number }`. Since #59183 that's no longer the case - its type stays as `{}`. This fact/difference isn't particularly observable to the user but the signature computed based on it can be quite different. The signature is meant to be `(_: { a?: number }) => void`. 

Since the type of the initializer expressions changed in #59183, Anders introduced "object padding" for those initializer types - a way to go from the initializer type to a parameter type using the available binding pattern information. Such mechanism was already used for tuple types. So this was a way to unify the algorithm for both object and array patterns in parameter positions.

That new object padding introduced regressions but only because it didn't cover some cases (see https://github.com/microsoft/TypeScript/pull/59924 ) and because it's too eager in some other cases (like this one here). 

TLDR: it's not a revert, it's just an extension of those changes

Issue #59989, opened 2024-09-17
Cannot access protected field but private field inside nested class


jrieken:
### 🔎 Search Terms

private, protected

### 🕗 Version & Regression Information

Version 5.6.2

### ⏯ Playground Link

https://www.typescriptlang.org/play/?#code/KYDwDg9gTgLgBAYwDYEMDOa4DEITgbwChC5S4woIZgFqATOGACwEs0BJNABUutuAYBeOAEYA3CTIUWANxTVGrDtyiz5wOMPHEycAEYooACgCUALhx4iu3QggA7NPGbzNithJtlJXqMBgArlD2cPbAAO6IqBhwoNT2dJiWBDpeNmgQALb+rPYA5qYpacVwfoHBivIAdMxsnDxqCgDUlTA1SvW8NPSeJaQAvj42g7qD-UA

### 💻 Code

```ts
export class Foo {

    protected thisIsProtected = 1;
    private thisIsPrivate = 1;

    bar():Foo {
        const that = this;
        
        return new class extends Foo {

            something() {
                return that.thisIsPrivate + that.thisIsProtected;
            }
        }
    }
}
```

### 🙁 Actual behavior

I get a compile error for `that.thisIsProtected`, but not for `that.thisIsPrivate`

### 🙂 Expected behavior

I can access the protected field, like the private

schart:
### Firstly playground link: https://www.typescriptlang.org/play/?#code/KYDwDg9gTgLgBAYwDYEMDOa4DEITgbwCg4S4woIZgEqATOGACwEs0BJNABQqpuHoC8cAIwBuYqXLMAbiioMW7LlBlzgcIWMISSAIxRQAFAEoAXNlwEdpElGAwArlAB2cZ8ADuiVBjigqzrSYOHhENuGkaBAAtvYszgDmJlYRqaR2ji4KrADaAORMrBzcqlR5ALpwANTZaAB0hUrclNR04mk2AL7WXe2k3d2ECBDOaPAAZrgA+sjomELuXiEm4sOj8LTAKtLMiTM+83CTEPtzdfpGxqsjUUjAdUgQSZvbuwmnGFeEQA
 
### Error: **Since the private variable is accessed through "that" variable, TypeScript considers this as a rule violation and gives a compilation error. Because private variables cannot be accessed directly from outside the class in which they are defined.**

### **Code:**

    export class Foo {

            protected thisIsProtected = 1;
            private thisIsPrivate = 1;
              
[...]

PR #47078, opened 2021-12-09
Explicitly Omit unspreadable properties from rest type in the generic case

jakebailey:
Fixes #46967

In the non-generic case, we'd produce a concrete type that consisted only of the spreadable properties, excluding already destructured and unspreadable properties. This is good.

However, in the generic case (including `this`, as it's an implicit type parameter), we instead produced an `Omit` type, but only omitted the already destructured keys. This left a type where all of the unspreadable properties (e.g. properties with getters, methods) where left accessible.

Make the generic case match the behavior of the non-generic case by also adding the unspreadable items to the omitted set of keys.

This might get pretty hairy for types with lots of getters/setters/methods, though, but the alternative (making a concrete type instead if it looks like `Omit` can't handle it) appears to break a few tests in even more undesirable ways. Hopefully destructuring `this` with a rest is not all that common (the user tests imply it is not).

Additionally, I don't know if the behavior on non-public properties is correct; it seems weird that destructuring `this` within the class would only keep the public properties, even though at runtime they are there. They're dropped in the `T extends SomeClass` case because that's a use where those symbols are not intended to be visible, but internally, maybe it's not right to drop them.

[...]

PR #60001, opened 2024-09-18
Allow protected members to be accessed within the enclosing class

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/59989


src/compiler/checker.ts
```diff
@@ -33868,7 +33868,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             // get the original type -- represented as the type constraint of the 'this' type
             containingType = (containingType as TypeParameter).isThisType ? getConstraintOfTypeParameter(containingType as TypeParameter)! : getBaseConstraintOfType(containingType as TypeParameter)!; // TODO: GH#18217 Use a different variable that's allowed to be undefined
         }
-        if (!containingType || !hasBaseType(containingType, enclosingClass)) {
+        if (!containingType || !hasBaseType(containingType, enclosingClass) && !isNodeWithinClass(location, getClassLikeDeclarationOfSymbol(getParentOfSymbol(prop)!)!)) {
```
Andarist, line 33871: the added check is the same one as for private properties a couple of lines above this (in this function)

jakebailey:
@typescript-bot test it

jakebailey:
In a way this feels wrong; I remember on my first checker bugfix that we didn't want to let one spread out private props which feels similar: https://github.com/microsoft/TypeScript/pull/47078#issuecomment-990132945

But, allowing private and not protected seems the oddest of all.

Andarist:
Honestly, I don't really know what are the expectations around private/protected when it comes to closures etc. Private properties are allowed in those positions quite explicitly by the code (and IIRC there are tests for it) so this merely matches the behavior of protected properties with that.

I don't feel anything in the linked comment says strongly this might be wrong:
> access to private/protected members is only supposed to be allowed through references of the actual class type from within the class

Those references to this class here are **within the class** ;p I feel like the linked comment is mainly about object types created by spreads than about other cases like this here.

src/compiler/checker.ts
jakebailey: It does besides the null checking; is it really safe to `!` all of these? The other case doesn't.

Andarist: The other case does exactly the same. I copy-pasted this :p

```ts
        // Private property is accessible if the property is within the declaring class
        if (flags & ModifierFlags.Private) {
            const declaringClassDeclaration = getClassLikeDeclarationOfSymbol(getParentOfSymbol(prop)!)!;
            if (!isNodeWithinClass(location, declaringClassDeclaration)) {
                if (errorNode) {
                    error(errorNode, Diagnostics.Property_0_is_private_and_only_accessible_within_class_1, symbolToString(prop), typeToString(getDeclaringClass(prop)!));
                }
                return false;
            }
            return true;
        }
```

jakebailey: Oh, yeah, I can't read

jakebailey: Why don't we just do this check unconditionally at the top? It sure seems like the result of this code is that "you can access any property of the current class as long as you're within that class", such that we don't need to do all of this runaround.

Issue #60004, opened 2024-09-19
Source mappings are missing for serialized properties


Andarist:
### 🔎 Search Terms

source map declaration map properties symbols navigation definition goto

### 🕗 Version & Regression Information

- This is the behavior in every version I tried

### ⏯ Playground Link

N/A

### 💻 Code

```ts
// api.ts
type ValidateShape<T> = {
  [K in keyof T]: T[K];
};

function test<T>(arg: ValidateShape<T>) {
  function createCaller<T>(arg: T): () => {
    [K in keyof T]: () => T[K];
  } {
    return null as any;
  }
  return {
    createCaller: createCaller(arg),
  };
}

const api = test({
  foo/*target*/: 1,
  bar: "",
});

export const createCaller = api.createCaller;

// main.ts

import { createCaller } from "./api";
const caller = createCaller();
caller.foo/*source*/;
```

### 🙁 Actual behavior

Please consider the 2 files above as 2 separate projects.

When those files are in the same project we can successfully go from the source marker to the target marker with "go to definition" and others like it. The same works with project references on.

However, when `api.ts` gets compiled with declaration maps and consumed through that then this navigation no longer works correctly.

### 🙂 Expected behavior

I'd expect this navigation to continue to work

PR #59675, opened 2024-08-18
fix(59397): JsDoc is missing/duplicated in declarations for overloads declared in classes declared in functions

a-tarasyuk:
Fixes #59397

PR #60005, opened 2024-09-19
Add source mappings for serialized properties with available declaration

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/60004


src/testRunner/unittests/helpers/virtualFileSystemWithWatch.ts
```diff
@@ -1158,7 +1158,12 @@ export class TestServerHost implements server.ServerHost, FormatDiagnosticsHost,
 
         // base folder has to be present
         const base = getDirectoryPath(file.path);
-        const folder = Debug.checkDefined(this.getRealFolder(base));
+        const folder = this.getRealFolder(base);
+        // this line can throw on purpose, don't use Debug.assert to avoid redundant debugger hits
+        // eslint-disable-next-line no-restricted-syntax
+        if (folder === null || folder === undefined) {
+            throw new Error(`Directory not found: ${base}`);
+        }
```
jakebailey, line 1166: Do you mind sending this as a separate change; I've been wanting to do this for a while so would be nice to get in regardless of the status of this PR.

No need to leave the comment, and I suspect that a plain truthiness check is fine too.

Andarist: https://github.com/microsoft/TypeScript/pull/60009

src/compiler/checker.ts
```diff
@@ -985,6 +985,7 @@ import {
     setNodeFlags,
     setOriginalNode,
     setParent,
+    setSourceMapRange,
```
jakebailey, line 988: I feel like it's suspicious that the checker did not already import this but presumably gets things right for other nodes...

src/testRunner/unittests/tsserver/projectReferencesSourcemap.ts
```diff
@@ -549,4 +550,51 @@ fn5();
...
+                    const caller = createCaller();
+                    caller.foo;
+                `,
+            };
+            const files = [dependencyTs, dependencyConfig, mainTs, mainConfig];
+            const session = createSession(type, files);
+            openFilesForSession([mainTs], session);
+            session.executeCommandSeq<ts.server.protocol.DefinitionAndBoundSpanRequest>({
+                command: ts.server.protocol.CommandTypes.DefinitionAndBoundSpan,
+                arguments: { file: mainTs.path, line: 3, offset: "caller.foo".length - 2 },
```
sheetalkamat, line 595: Pls use `protocolFileLocationFromSubstring` to get line and offset

src/compiler/checker.ts
Andarist: Heh, ye - definitely. I was quite surprised that this was the first time this had to be imported here.

Usually, this gets called by *transforms* but this case is different than that use case. There might also be other issues like this one. In general, declaration maps bugs are less likely to be reported than typechecking bugs etc so 🤷

src/testRunner/unittests/tsserver/projectReferencesSourcemap.ts
Andarist: done

sheetalkamat (approved):

src/compiler/checker.ts
```diff
@@ -7174,8 +7175,12 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                     context.tracker.reportNonSerializableProperty(symbolToString(propertySymbol));
                 }
             }
-            context.enclosingDeclaration = propertySymbol.valueDeclaration || propertySymbol.declarations?.[0] || saveEnclosingDeclaration;
+            const propertyDeclaration = propertySymbol.valueDeclaration || propertySymbol.declarations?.[0];
+            context.enclosingDeclaration = propertyDeclaration || saveEnclosingDeclaration;
             const propertyName = getPropertyNameNodeForSymbol(propertySymbol, context);
+            if (propertyDeclaration && (isPropertyAssignment(propertyDeclaration) || isShorthandPropertyAssignment(propertyDeclaration) || isMethodDeclaration(propertyDeclaration) || isMethodSignature(propertyDeclaration) || isPropertySignature(propertyDeclaration) || isPropertyDeclaration(propertyDeclaration) || isGetOrSetAccessorDeclaration(propertyDeclaration))) {
+                setSourceMapRange(propertyName, propertyDeclaration.name);
```
jakebailey, line 7182: I still feel at least a little uncomfortable with this; why isn't it declaration emit which can do this? The node builder produces lots of nodes, why doesn't this happen elsewhere?

jakebailey: Also, can this just be done in `getPropertyNameNodeForSymbol`?

(In any case, this PR is safe, since `getPropertyNameNodeForSymbol` is returning a new node.)

Andarist: > I still feel at least a little uncomfortable with this; why isn't it declaration emit which can do this?

the required information is on the symbol, emit works based on the produced nodes and we need to pass somehow the information that it needs from here as this is the place where this information is being lost. This just leaves a breadcrumb that can be picked up by the emitter


> Also, can this just be done in getPropertyNameNodeForSymbol?

maybe, im not sure sure ;p I'll try to analyze this

weswigham, line 7182: For declaration emit nodes, just use `setTextRange` rather than `setSourceMapRange` - there's a helper in scope that takes care of a bunch of sanity checking for you that's absent here. But yeah, you should also be able to move this into `getPropertyNameNodeForSymbol` so it happens for all callers.

weswigham (changes requested):

src/compiler/checker.ts
Andarist: Using `setTextRange` doesn't cut it and the results are as on the `main` branch. Am I using it wrong or missing something?

I have moved the fix to `getPropertyNameNodeForSymbol` though

jakebailey: Yeah, it doesn't seem to help. I don't understand why this isn't working:

```ts
propertyNameNode = setTextRange(context, propertyNameNode, declaration.name);
```

So I would expect the original node stuff to "just work". `setTextRange` doesn't call `setSourceMapRange` at all, though, so...

jakebailey: If you stick `setSourceMapRange(range, location);` into `setTextRange` after `setOriginalNode`, that actually also changes a whole bunch of other stuff kinda for the better? But then breaks some other JSDoc related code (which I think was another recent bugfix that may conflict?) Dunno if that's a good idea or not, I don't have a good mental model of this.

jakebailey: (maybe #59675?)

Eigilak:
Status?

Jacse:
@Andarist @jakebailey is this progressing?

mamlzy:
sorry for tagging @Andarist, is this pr going to be merged?

prometx11:
Is there any chance for this to be merged soon? 
This is blocking go-to navigation in monorepos, and degrades DX quite a bit.

jakebailey:
This needs main merged into it, for one thing.

The other thing I'm wondering is how trivial this is to port over to the Go codebase, just to make sure that this sticks.

dbrxnds:
Should there be an issue opened in the native repo to track this for the go port? We recently switched to go but would like for this to stick.

Andarist:
@jakebailey I synced this with `main`. I also don't see any reason why this would be hard to port to the Go codebase. I could prepare a porting PR there but I don't know of the top of my head if you already have test infra there that would be able to test it.

jakebailey:
We have declaration map tests and go to def baselining, so theoretically it should all be there.

jakebailey:
But, I don't know what the equivalent to projectReferencesSourcemap.ts is in the Go codebase; maybe @sheetalkamat knows.

sheetalkamat:
you should be able to add fourslash test. https://github.com/microsoft/typescript-go/blob/main/internal/fourslash/tests/statedeclarationmaps_test.go

RahimGuerfi:
...

PR #60009, opened 2024-09-19
Replace `Debug.checkDefined` to avoid redundant debugger stops

Andarist:
cc @jakebailey


jakebailey (approved):

jakebailey:
Adding @sheetalkamat to verify but I've personally had issues debugging locally due to this assert firing when it was okay to do so, so I believe this is an okay change.

sheetalkamat (approved):

PR #60011, opened 2024-09-19
Fix JSDoc @import resolution in `--moduleResolution bundler`

### LINKED ISSUE #60007: JSDoc `@import` of "tslib" fails under `--moduleResolution bundler`
```ts
// @noEmit: true
// @checkJs: true
// @allowJs: true
// @module: preserve

// @Filename: node_modules/tslib/package.json
{
    "name": "tslib",
    "exports": {
        ".": {
            "module": {
                "types": "./modules/index.d.ts",
                "default": "./tslib.es6.mjs"
            },
            "import": {
                "node": "./modules/index.js",
                "default": {
                    "types": "./modules/index.d.ts",
                    "default": "./tslib.es6.mjs"
                }
            },
            "default": "./tslib.js"
        },
        "./*": "./*",
        "./": "./"
    }
}

// @Filename: node_modules/tslib/modules/index.d.ts
export {};

// @Filename: node_modules/tslib/tslib.d.ts
export {};

// @Filename: test.js
/** @import * as tslib from "tslib" */
/** @type {typeof tslib} T */
```

## Expected behavior

Import resolves

## Actual behavior

```
Cannot find module 'tslib' or its corresponding type declarations.ts(2307)
```

Even though quick info also shows

```ts
module "/Users/andrew/Developer/microsoft/eg/js/node_modules/tslib/modules/index"
```

andrewbranch:
Fixes #60007

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -4617,12 +4617,10 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             ? location
             : (isModuleDeclaration(location) ? location : location.parent && isModuleDeclaration(location.parent) && location.parent.name === location ? location.parent : undefined)?.name ||
                 (isLiteralImportTypeNode(location) ? location : undefined)?.argument.literal ||
-                (isInJSFile(location) && isJSDocImportTag(location) ? location.moduleSpecifier : undefined) ||
                 (isVariableDeclaration(location) && location.initializer && isRequireCall(location.initializer, /*requireStringLiteralLikeArgument*/ true) ? location.initializer.arguments[0] : undefined) ||
                 findAncestor(location, isImportCall)?.arguments[0] ||
-                findAncestor(location, isImportDeclaration)?.moduleSpecifier ||
-                findAncestor(location, isExternalModuleImportEqualsDeclaration)?.moduleReference.expression ||
-                findAncestor(location, isExportDeclaration)?.moduleSpecifier;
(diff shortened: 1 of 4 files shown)


jakebailey:
@typescript-bot test it

andrewbranch:
I think this can go in after the beta cut if needed, but either way, @sandersn feel free to merge for me upon approval.

src/compiler/checker.ts
```diff
@@ -4617,12 +4617,10 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
             ? location
             : (isModuleDeclaration(location) ? location : location.parent && isModuleDeclaration(location.parent) && location.parent.name === location ? location.parent : undefined)?.name ||
                 (isLiteralImportTypeNode(location) ? location : undefined)?.argument.literal ||
-                (isInJSFile(location) && isJSDocImportTag(location) ? location.moduleSpecifier : undefined) ||
```
sandersn, line 4620: Does it matter that we aren't checking `isInJsFile` any more? Would we resolve an `@import` inside a ts file now?

sandersn (changes requested):
I'd like to know whether this will make us start resolving `@import` in TS files.

src/compiler/checker.ts
sandersn: Confirmed by changing the test to .ts instead. I'll add that as importTag22.ts and merge the PR.

sandersn (approved):

PR #60019, opened 2024-09-20
Validate JSON imports into ESM in `--module nodenext`

### LINKED ISSUE #59684: Module resolution: Named imports not supported with import attributes
### Demo Repo

https://github.com/mdmower/with-json-tsc

See bottom of post for information.

### Which of the following problems are you reporting?

The module specifier resolves at build time, but not at run time

### Demonstrate the defect described above with a code sample.

import {compilerOptions} from './tsconfig.json' with {type: 'json'};

### Run `tsc --showConfig` and paste its output here
```json
{
    "compilerOptions": {
        "target": "es2023",
        "lib": [
            "es2023"
        ],
        "module": "nodenext",
        "moduleResolution": "nodenext",
        "resolveJsonModule": true,
        "strict": true,
        "moduleDetection": "force",
        "esModuleInterop": true,
        "resolvePackageJsonExports": true,
        "resolvePackageJsonImports": true,
        "useDefineForClassFields": true,
        "noImplicitAny": true,
        "noImplicitThis": true,
        "strictNullChecks": true,
        "strictFunctionTypes": true,
        "strictBindCallApply": true,
        "strictPropertyInitialization": true,
        "alwaysStrict": true,
        "useUnknownInCatchVariables": true
    },
    "files": [
        "./index.ts"
    ]
}
```

### Run `tsc --traceResolution` and paste its output here

Found 'package.json' at 'C:/Users/matt.mower/source/with-json-tsc/package.json'.

> There was an error creating your Issue: body is too long, body is too long (maximum is 65536 characters).

### Paste the `package.json` of the *importing* module, if it exists
```json
{
  "type": "module",
  "scripts": {
    "build": "tsc",
    "start": "node index.js"
  },
  "devDependencies": {
    "@types/node": "^22",
    "typescript": "^5.5"
  }
}
```

### Paste the `package.json` of the *target* module, if it exists

N/A

### Any other comments can go here

TypeScript 5.3 added support for [import attributes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-3.html#import-attributes). `tsc` does not warn or error when attempting to use named imports for these kinds of imports. When named imports are used, they are preserved in the produced `.js` and there is no environment where they are supported.

For example, in the linked demo repository:
```ts
import {compilerOptions} from './tsconfig.json' with {type: 'json'};
console.log(compilerOptions);
```
compiles to:
```js
import { compilerOptions } from './tsconfig.json' with { type: 'json' };
console.log(compilerOptions);
```

which does not run in Node.js 22 or Chrome 127.

Given that `--resolveJsonModule` has historically allowed named imports and produces valid CommonJS, it's likely that developers will assume TypeScript can handle named imports when using import attributes as well.

For example, if the linked demo repository is set to output CommonJS instead of ESM, then:
```ts
import {compilerOptions} from "./tsconfig.json";
console.log(compilerOptions);
```
compiles to:
```js
"use strict";
const tsconfig_json_1 = require("./tsconfig.json");
console.log(tsconfig_json_1.compilerOptions);
```

It seems that TypeScript should disallow named imports when import attributes are used, or should compile so that a default import is used, similar to handling in CommonJS.

andrewbranch:
Fixes #59684 

This does two main things:

1. Enforces use of import attributes on JSON-file imports in `--module nodenext`, since Node.js crashes without them
2. Errors on named imports of JSON-file imports in `--module nodenext`, since those don't work in Node.js. We already sort of considered this information when building up the type of a namespace import of a JSON file, but named imports bypassed that check.

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -3967,7 +3967,11 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
                 const symbol = symbolFromModule && symbolFromVariable && symbolFromModule !== symbolFromVariable ?
                     combineValueAndTypeSymbols(symbolFromVariable, symbolFromModule) :
                     symbolFromModule || symbolFromVariable;
-                if (!symbol) {
+
+                if (isImportOrExportSpecifier(specifier) && isOnlyImportableAsDefault(moduleSpecifier) && nameText !== InternalSymbolName.Default) {
+                    error(name, Diagnostics.Named_imports_from_a_JSON_file_into_an_ECMAScript_module_are_not_allowed_when_module_is_set_to_0, ModuleKind[moduleKind]);
+                }
+                else if (!symbol) {
                     errorNoModuleMemberSymbol(moduleSymbol, targetSymbol, node, name);
                 }
                 return symbol;
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -163,7 +163,6 @@ import {
     EmitResolver,
     EmitTextWriter,
     emptyArray,
-    endsWith,
     EntityName,
     EntityNameExpression,
     EntityNameOrEntityNameExpression,
@@ -265,6 +264,7 @@ import {
     getContainingClassStaticBlock,
     getContainingFunction,
     getContainingFunctionOrClassStaticBlock,
+    getDeclarationFileExtension,
     getDeclarationModifierFlagsFromSymbol,
     getDeclarationOfKind,
     getDeclarationsOfKind,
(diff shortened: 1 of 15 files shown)


src/compiler/checker.ts
```diff
@@ -3700,11 +3700,15 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         return usageMode === ModuleKind.ESNext && targetMode === ModuleKind.CommonJS;
     }
 
-    function isOnlyImportableAsDefault(usage: Expression) {
+    function isOnlyImportableAsDefault(usage: Expression, resolvedModule?: Symbol) {
         // In Node.js, JSON modules don't get named exports
         if (ModuleKind.Node16 <= moduleKind && moduleKind <= ModuleKind.NodeNext) {
             const usageMode = getEmitSyntaxForModuleSpecifierExpression(usage);
-            return usageMode === ModuleKind.ESNext && endsWith((usage as StringLiteralLike).text, Extension.Json);
```
andrewbranch, line 3707: The text of the module specifier is not sufficient to know whether we're importing a JSON file; this produced both false positives and false negatives (see imports of `"actually-json"` and `"not.json"` in the test)

src/compiler/parser.ts
```diff
@@ -10517,9 +10517,10 @@ export function getDeclarationFileExtension(fileName: string): string | undefine
         return standardExtension;
     }
     if (fileExtensionIs(fileName, Extension.Ts)) {
-        const index = getBaseFileName(fileName).lastIndexOf(".d.");
+        const baseName = getBaseFileName(fileName);
+        const index = baseName.lastIndexOf(".d.");
         if (index >= 0) {
-            return fileName.substring(index);
+            return baseName.substring(index);
```
andrewbranch, line 10523: This was extremely bad; I guess we were only really using this to determine whether something _was_ a declaration file name, not to actually use the extension.

weswigham (approved):

Issue #59946, opened 2024-09-12
Fail to infer Indexed access to the intersection type when it includes an union derived with a generic input


n-ryu:
### 🔎 Search Terms

"indexed access on union", "generic index"

### 🕗 Version & Regression Information

5.6.2, and it seems never worked for every ts versions.

### ⏯ Playground Link

https://www.typescriptlang.org/play/?ts=5.6.2#code/C4TwDgpgBAKhDOwA8MoQB7AgOwCbykQCcBLbAcwD4oBeKAbwG0BpKM2AXQC4GBrCED2YBfYYxgcoAMgbAAhkXIRgPYmXLCAUJoDGAe2yIoAMz17aUFGkw58hYKQqUAFAvI84iFJQCUtavSaUMFQ+obAUABucgA2AK4QHgjIMNR0bgDcQSFhRtHxEABMSV6pjADk-CDlkumKFVU1WSGhBnmxCQDMJSmUFfKKyjUWbv1uQxwZUAD001AAokREekSawkA

### 💻 Code

```ts
type Test<T extends string> = {[K in T]: {key: K}}[T] & {target: string}

const foo = <T extends string>(arg: Test<T>) => {
    const value: Test<T> = arg; 
    const value2: Test<T>['key'] = arg['key']; 
    const value3: Test<T>['target'] = arg['target']; // Error
}
```

### 🙁 Actual behavior

The declaration of `value` and `value2` works well, but ts emit an error saying `Type 'string' is not assignable to type '{ [K in T]: { key: K; }; }[T]["target"] & string'` for `value3`. 

### 🙂 Expected behavior

The `Test<T>['target']` should equal to `string` and assume the type of `'{ [K in T]: { key: K; }; }[T]["target"] ` as `any` before intersection, as it does when I give non-generic type in the same place. It seems even weird because it well infers the type of `Test<T>['key']`.

### Additional information about the issue

With additional experiment, I found that it always happens when I generate an union type with generic input. For instance, 
```ts
type Test<T> = (T extends string ? {key: number} : {key: string}) & {target: string}

const foo = <T extends string>(arg: Test<T>) => {
    const value: Test<T> = arg;
    const value2: Test<T>['key'] = arg['key'];
    const value3: Test<T>['target'] = arg['target']; // Error
}
```

it also does not work in the case above

PR #60021, opened 2024-09-21
Simplify object types when distributing index over them

Andarist:
fixes https://github.com/microsoft/TypeScript/issues/59946

### DIFF
--- a/src/compiler/checker.ts
+++ b/src/compiler/checker.ts
@@ -19148,6 +19148,20 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         // In noUncheckedIndexedAccess mode, indexed access operations that occur in an expression in a read position and resolve to
         // an index signature have 'undefined' included in their type.
         if (compilerOptions.noUncheckedIndexedAccess && accessFlags & AccessFlags.ExpressionPosition) accessFlags |= AccessFlags.IncludeUndefined;
+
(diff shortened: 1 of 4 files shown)


src/compiler/checker.ts
```diff
@@ -18997,7 +18997,7 @@ export function createTypeChecker(host: TypeCheckerHost): TypeChecker {
         // (T | U)[K] -> T[K] & U[K] (writing)
         // (T & U)[K] -> T[K] & U[K]
         if (objectType.flags & TypeFlags.Union || objectType.flags & TypeFlags.Intersection && !shouldDeferIndexType(objectType)) {
-            const types = map((objectType as UnionOrIntersectionType).types, t => getSimplifiedType(getIndexedAccessType(t, indexType), writing));
+            const types = map((objectType as UnionOrIntersectionType).types, t => getSimplifiedType(getIndexedAccessType(getSimplifiedType(t, writing), indexType), writing));
```
Andarist, line 19000: Tt's important to simplify the `t` (a constituent of the `objectType`) here "eagerly". At this stage `getIndexedAccessType` has a chance to return `unknownType` (thus effectively eliminating this constituent from the overall simplified indexed access type on an intersection, since `T & unknown` is `T`).

Without it, the compiler comes back to `getSimplifiedIndexedAccessType` with the non-simplified indexed access. It simplifies the object type there, but it also decides there is nothing to simplify further so it returns the original non-simplified `type`. That leads to a relationship error.

jakebailey:
@typescript-bot test it

